PR #318, opened 2026-04-09
fix: cmd wrapper resolution, auth fallback, loading spinner, layout healing

### LINKED ISSUE #314: cmd.exe /C .cmd wrappers: broken arg passing, piped output, and version detection
## Summary

npm-installed CLIs on Windows produce `.cmd` batch wrappers (e.g. `claude.cmd`). Throughout the codebase we invoke these via `cmd.exe /C <path>.cmd <args>`, but this pattern is fundamentally broken in several ways when the process is spawned from Rust's `Command` API with piped stdio.

## Symptoms

1. **`--version` returns empty stdout** — `cmd.exe /C claude.cmd --version` prints the Windows banner (`Microsoft Windows [Version 10.0.19045]`) instead of the CLI version. The `--version` arg is silently dropped.

2. **`auth login` produces no output** — `cmd.exe /C claude.cmd auth login` with piped stdout/stderr captures nothing. The CLI never runs; only `cmd.exe`'s own banner appears.

3. **Browser doesn't open during login** — because the CLI process never actually starts, the OAuth browser-open call never executes.

4. **`CREATE_NO_WINDOW` + `Stdio::null()` double-blocks** — even if `cmd.exe /C` worked, combining `CREATE_NO_WINDOW` (creation flag 0x08000000) with null stdio on all handles prevents the child from opening a browser or writing output.

## Root Cause

When Rust's `Command::new("cmd.exe").arg("/C").arg("path.cmd").arg("--version")` is invoked, `CreateProcessW` receives the argument string as:

```
cmd.exe /C "path.cmd" "--version"
```

`cmd.exe /C` treats everything after `/C` as the command string. With separate `.arg()` calls, Rust quotes each argument independently. `cmd.exe` sees `"path.cmd"` as the command and `"--version"` as a **separate** token that it doesn't pass through — it's interpreted as part of `cmd.exe`'s own argument parsing, not the child script's args.

The `.cmd` wrapper internally does:
```batch
endLocal & goto #_undefined_# 2>/dev/null || title %COMSPEC% & "%_prog%" "%dp0%\..\@anthropic-ai\claude-code\cli.js" %*
```

The `%*` expands to **nothing** because `cmd.exe` consumed the args.

## Affected Code Paths

| Location | Issue |
|----------|-------|
| `agentmux-srv/src/server/cli_handlers.rs` `make_cli_cmd()` | Used for `--version`, auth check, auth login in sidecar |
| `agentmux-cef/src/commands/platform.rs` `run_cli_login()` | Auth login in CEF host |
| `agentmux-srv/src/server/cli_handlers.rs` `get_cli_version()` | Version detection after install |

## Working Fix

Resolve the `.cmd` wrapper to the underlying `node <cli.js>` script and invoke `node` directly:

```rust
// Instead of: cmd.exe /C claude.cmd --version
// Do:         node <prefix>/node_modules/@anthropic-ai/claude-code/cli.js --version
```

This is already partially implemented in `get_cli_version()` (using `raw_arg`) and `run_cli_login()` (resolving to `node cli.js`). But the fix should be centralized in a single helper that all code paths use.

## Alternative: `raw_arg` approach

`raw_arg` passes the entire command string verbatim to `CreateProcessW`, avoiding Rust's quoting:

```rust
use std::os::windows::process::CommandExt;
let cmd_str = format!("\"{}\" --version", cli_path);
Command::new("cmd.exe").arg("/C").raw_arg(&cmd_str)
```

This works for `std::process::Command` but `tokio::process::Command` may not always delegate `raw_arg` correctly. The `node cli.js` approach is more reliable.

## Proposed Solution

1. Create a centralized `resolve_cli_command(cli_path) -> (program, base_args)` helper
2. For `.cmd` files: parse the wrapper to find the `cli.js` path, return `("node", [cli.js_path])`  
3. For `.exe` files: return `(cli_path, [])`
4. For Unix: return `(cli_path, [])`
5. Use this helper everywhere: `make_cli_cmd()`, `run_cli_login()`, `get_cli_version()`

## Impact

All npm-installed CLI providers on Windows: Claude (`@anthropic-ai/claude-code`), Codex (`@openai/codex`), Gemini (`@google/gemini-cli`).

Labels: bug, windows, P1

AgentA-asaf:
## Summary
- **`.cmd` wrapper resolution**: Parse npm `.cmd` wrappers to extract the Node.js entry script and invoke `node <script>` directly — bypasses broken `cmd.exe /C` with piped stdio. Applied in all 3 spawn sites.
- **Auth fallback**: Check isolated auth dir first, then fall back to global `~/.claude/.credentials.json`.
- **Loading spinner**: Pulsing dot + "loading" text in agent footer during the launch flow.
- **Layout self-healing**: On startup, walk all tabs and prune orphaned block nodes from layouts. Also prune on every `DeleteBlock`. Prevents blank panes from ever persisting.

Closes #314

## Test plan
- [ ] Agent pane: CLI resolves, auth passes, controller registers
- [ ] Loading spinner visible during launch flow
- [ ] Close a pane, reopen app — no blank spots
- [ ] Backend logs show `layout self-healing` messages on startup

### DIFF
--- a/.bump.json
+++ b/.bump.json
@@ -25,6 +25,11 @@
       "file": "agentmux-launcher/Cargo.toml",
       "type": "toml",
       "path": "package.version"
+    },
+    {
+      "file": "agentmux-common/Cargo.toml",
+      "type": "toml",
+      "path": "package.version"
     }
   ],
   "lockfiles": [
@@ -53,7 +58,8 @@
       "agentmux-srv/Cargo.toml",
       "agentmux-wsh/Cargo.toml",
       "agentmux-cef/Cargo.toml",
-      "agentmux-launcher/Cargo.toml"
+      "agentmux-launcher/Cargo.toml",
+      "agentmux-common/Cargo.toml"
     ],
     "commitMessage": "chore: bump version to {{version}}"
   }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.67"
+version = "0.33.68"
 dependencies = [
  "axum 0.8.8",
  "cef",
@@ -33,7 +33,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.67"
+version = "0.33.68"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -41,7 +41,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.67"
+version = "0.33.68"
 dependencies = [
  "async-stream",
  "axum 0.7.9",
@@ -78,7 +78,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-wsh"
-version = "0.33.67"
+version = "0.33.68"
 dependencies = [
  "base64",
  "clap",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.68"
+version = "0.33.69"
 dependencies = [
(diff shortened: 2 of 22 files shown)


AgentA-asaf:
Superseded by #325 which was merged with the same changes (rebased onto main).

PR #464, opened 2026-04-20
refactor(agent): rename pinnedToolNodes → pinnedNodes (hover-unification phase 1)

AgentY-asaf:
## Summary

- Renames `pinnedToolNodes` → `pinnedNodes` in `DocumentState` (types.ts, state.ts, AgentDocumentView.tsx)
- Pure compiler-guaranteed refactor — 5 references, no behavior change, TypeScript clean
- Widens the field's semantic scope from tool-only to any node kind, unblocking phases 2-8 of the hover-unification plan

## Context

Part of the hover-unification refactor (`docs/specs/hover-unification-implementation-plan.md`). AgentX completed this rename; AgentY is picking up from here.

## Test plan

- [ ] `npx tsc --noEmit` passes (verified)
- [ ] Open agent pane, click to pin a tool — stays pinned
- [ ] `grep pinnedToolNodes frontend/` returns zero results

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #466, opened 2026-04-20
feat(agent): NodeHoverStrip component + timestamp (hover-unification phase 2)

AgentY-asaf:
## Summary

- Introduces `NodeHoverStrip` component replacing `NodeTimestamp`
- Localized timestamp format via `Intl.DateTimeFormat` (weekday + date + 12-hour AM/PM) instead of `HH:MM:SS.t`
- Strip is a flex row that will host action buttons in phases 3+
- Pure-CSS show/hide (`:hover`, `:focus-within`) — zero JS state overhead
- Deletes `NodeTimestamp.tsx`

## Context

Part of the hover-unification refactor. Stacks on #464 (phase 1).

## Test plan

- [ ] `npx tsc --noEmit` passes
- [ ] Hover a node row → strip appears at top-right with localized timestamp
- [ ] Row with no timestamp → no strip rendered
- [ ] `:focus-within` reveals strip when any child has focus


AgentY-asaf:
Fixed: removed unrelated files, dropped role="toolbar" (conflicts with pointer-events:none), restored opacity 80ms ease fade transition.

PR #504, opened 2026-04-23
feat(agent-pane): definition cards + launch modal + per-instance working dirs (PRs B-E)

AgentA-asaf:
## Summary
- Reworks the agent pane so each card is a **definition** (one per CLI), not a launched agent
- Clicking a card opens a **Launch modal** that collects instance name + host/container runtime
- Each launch gets its own working directory: \`data/agents/<slug>-<YYYYMMDD-HHMMSS>/\`
- Identity / inline-rename buttons removed from the card — identity editing stays reachable via the ⚙ Forge tab

## Context
Implements PRs B + C + D + E of [SPEC_AGENT_DEFINITIONS_MODAL_2026_04_23](../blob/agenta/agent-defs-bcd-cards-and-modal/docs/specs/SPEC_AGENT_DEFINITIONS_MODAL_2026_04_23.md). Follows PR #503 (dropped "+ Add Agent" from the pinned bar).

### What's in this PR

- **CLI catalog** (\`defaults/cli-catalog.ts\`) — 7 entries seeded from [discussion #493](https://github.com/agentmuxai/agentmux/discussions/493): Claude Code, Codex, Gemini, Kimi, Pi, OpenClaw, GitHub Copilot. Each has a capability blurb, icon, primary-context-file badge, and a longer popover description.
- **AgentCard** rendered definition-style: title = capability blurb, caption = CLI brand name + context-file badge. Info popover (\`ⓘ\`), Forge settings (\`⚙\`), and delete (\`🗑\`) remain.
- **AgentLaunchModal** — opens on card click. Fields: Name (required, 1–64 chars, becomes part of the working dir), Runtime (host / container radio; container image shows when selected), working-dir preview.
- **\`launchForgeAgent\`** now accepts \`LaunchOverrides\` (\`instanceName\`, \`agentType\`, \`environment\`, \`containerImage\`). Each override-based launch gets a fresh \`<slug>-<stamp>\` directory; override-less launches keep the legacy slug-only path for auto-resume / non-modal callers.
- **Per-instance env vars** — \`AGENTMUX_AGENT_ID\` / \`AGENTMUX_AGENT_DISPLAY\` now carry the instance name; the stable definition slug lives in \`AGENTMUX_AGENT_SLUG\`; the new \`AGENTMUX_INSTANCE_SLUG\` exposes the per-instance slug. \`GH_CONFIG_DIR\` still keys on the definition slug so GitHub auth persists.
[...]

PR #505, opened 2026-04-23
feat(forge): re-seed to one definition per CLI (drops AgentX/Y/Z/K/1/2/3)

AgentA-asaf:
## Summary
- Bumps forge-seed manifest \`version\` 4 → 5
- Replaces the legacy AgentX/Y/Z + Agent1/2/3 + AgentK layout with six provider-aligned definitions: \`claude\`, \`codex\`, \`gemini\`, \`kimi\`, \`pi\`, \`openclaw\`
- Each new row has an explicit \`provider\` field matching its card in the CLI catalog from discussion #493

## Context
Follow-up to PR #504. With the new definition-modal flow, each card *is* one CLI — so shipping 3 rows for claude + 3 for containers no longer fits. This seed migration makes the DB match the UI model.

The re-seed engine in \`agentmux-srv/src/backend/forge_seed.rs\` already handles the delta: on next startup it finds the old IDs missing from the manifest and the new IDs missing from the DB, deletes the former (all were \`is_seeded=1\`), and inserts the latter. User-created rows are untouched.

### Provider coverage

| provider | displayName | icon |
|----------|-------------|------|
| claude   | Claude Code | ✖ |
| codex    | Codex CLI   | ✦ |
| gemini   | Gemini CLI  | ⚡ |
| kimi     | Kimi Code   | ◈ |
| pi       | Pi          | π |
| openclaw | OpenClaw    | ◎ |

**Copilot** is in the frontend catalog (from discussion #493 research) but not yet in \`providers/index.ts\`. Launching it would crash, so the seed drops it. Follow-up PR will add the provider binding (cliCommand, launchArgs, authCheckCommand, controllerType) and restore the Copilot row.

### User impact

- Any existing panes running instances of AgentX/Y/Z/K/1/2/3 keep their process alive but the forge row is gone. Closing and re-launching picks from the new card set.
- Non-seeded rows (user-created) are preserved.
- Portable builds copy forge-seed.json into the embedded binary, so the migration runs on the next portable launch.

## Test plan
- [ ] Fresh install (\`~/.agentmux\` empty): agent picker shows 6 cards, one per provider
- [ ] Existing install with old seed: on next launch, old rows removed, 6 new rows inserted; picker reflects
- [ ] Each card can be clicked, modal opens, launch succeeds (all 6 providers registered in \`providers/index.ts\`)
- [ ] Forge pane shows the 6 new definitions with correct provider/icon/description
- [ ] User-customised rows from before are untouched

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.333"
+version = "0.33.334"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.333"
+version = "0.33.334"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.333"
+version = "0.33.334"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.333"
+version = "0.33.334"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.334"
+version = "0.33.335"
 dependencies = [
  "agentmux-common",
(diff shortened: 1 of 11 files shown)


AgentA-asaf:
Addressed both codex findings from PR #504 here in v0.33.335:

- **P1 (identity split):** `buildMcpConfig` and `buildConfigFiles` now take the resolved instance name. The injected agentmux MCP server and the `{{AGENT}}` / `{{AGENT_DISPLAY}}` template vars in CLAUDE.md match the shell/env/git-author identity for the run, so inter-agent messaging no longer routes on the definition name when a user picks a custom instance name.
- **P2 (sub-second slug collision):** `formatLocalStamp` now emits `YYYYMMDD-HHMMSS-mmm` (millisecond precision). Two concurrent launches of the same instance name would need to happen in the same millisecond to collide — that's no longer a realistic failure mode for either interactive or scripted use. Went with millisecond precision rather than the spec's `-N` retry loop because it avoids a new backend RPC/contract and keeps timestamps grep-friendly.

Thanks codex 🙏. Both were valid — the P1 cross-file check in particular was the kind of finding easy to miss in manual review.

@codex review

a5af:
@codex review

PR #506, opened 2026-04-24
feat(providers): add GitHub Copilot CLI + restore to seed

AgentA-asaf:
## Summary
- Registers the \`copilot\` provider in \`frontend/app/view/agent/providers/index.ts\`
- Re-adds the Copilot definition to \`forge-seed.json\` (manifest v5 → v6); all 7 CLI catalog entries now have backing definitions
- Provider uses ACP mode (\`--acp\` flag) so the existing ACP controller (already serving Pi + OpenClaw) drives it — same protocol surface, same output parsing

## Context
Follow-up to PR #505 (seed migration), which intentionally dropped Copilot because no provider binding existed yet. Now it's back.

### Why ACP, not \`-p\`
GitHub Copilot CLI's non-interactive \`-p\`/\`--prompt\` mode doesn't accept prompts via stdin — see [github/copilot-cli#96](https://github.com/github/copilot-cli/issues/96) and [#1046](https://github.com/github/copilot-cli/issues/1046). Our subprocess controller writes the user message to the CLI's stdin for every turn, so that path doesn't compose. \`--acp\` runs Copilot as an Agent Client Protocol JSON-RPC server over stdio — the exact contract our ACP controller already understands for Pi and OpenClaw.

### Runtime details
| Field | Value |
|-------|-------|
| CLI command | \`copilot\` |
| Install | \`npm install -g @github/copilot\` |
| Controller | \`acp\` |
| Launch args | \`--acp\` |
| Auth check | \`copilot auth status\` |
| Auth login | \`copilot auth login\` |
| Config dir env | \`COPILOT_HOME\` |
| Docs | https://docs.github.com/copilot/concepts/agents/about-copilot-cli |

Aliases added: \`copilot-cli\`, \`github-copilot\`, \`copilot_cli\`.

## Test plan
- [ ] Agent picker shows 7 cards (claude, codex, gemini, kimi, pi, openclaw, copilot)
- [ ] Copilot card opens the Launch modal on click
- [ ] With Copilot CLI installed and authed, Launch succeeds and the ACP controller connects
- [ ] With Copilot CLI not installed, the existing Node.js / CLI detection shows an install prompt
- [ ] Type check clean

## Follow-up
- Validate against the live Copilot CLI end-to-end (the ACP protocol details differ between implementations; if Copilot's ACP surface diverges from the Zed/Pi spec we already handle, add copilot-specific hooks in the ACP controller)
- Add a Copilot-specific icon glyph to replace the generic \`github\` key once the icon system has one

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.335"
+version = "0.33.336"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.335"
+version = "0.33.336"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.335"
+version = "0.33.336"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.335"
+version = "0.33.336"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

AgentA-asaf:
Addressed codex P2 in v0.33.337 — added `COPILOT` to the Rust provider registry at `agentmux-srv/src/backend/providers.rs` (REGISTRY + ALIASES + ORDER), mirroring the frontend entry. Without it `agent.open` → `get_provider` would have returned `INVALID_PROVIDER` on every Copilot launch. `cargo check` clean.

Good catch — the frontend-only split between the two registries is exactly the kind of asymmetry easy to miss when a provider lands frontend-first.

@codex review

a5af:
@codex review

PR #515, opened 2026-04-24
feat(modals): migrate MessageModal to modal-v2

AgentA-asaf:
## Summary
- Fourth modal-v2 consumer migration (after #512, #513, #514)
- MessageModal is ~24 lines; trivially small port
- Deletes \`messagemodal.scss\` (only contained \`min-width: 400px\`, now subsumed by modal-v2's \`size=\"md\"\` = 520px)

## Context
Per [SPEC_ROBUST_MODAL_SYSTEM §6](../blob/main/docs/specs/SPEC_ROBUST_MODAL_SYSTEM_2026_04_23.md) migration series.

### Diff
- Import swap: \`@/app/modals/modal\` → \`@/element/modal-v2\`
- \`<Modal onOk={...}>\` (implicit footer button) → explicit \`<ModalFooter><Button onClick={closeModal}>Ok</Button></ModalFooter>\`
- \`ariaLabel=\"Message\"\` since there's no header
- Removed the \`messagemodal.scss\` import + file

### Inherited from modal-v2
- role=dialog / aria-modal / focus trap / scroll lock / inert
- Backdrop blur + animations
- Multi-window Portal routing
- ESC + backdrop click both fire \`onClose\`

## Test plan
- [ ] Any code path that calls \`modalsModel.pushModal(MessageModal, ...)\` opens a dialog with backdrop blur
- [ ] Children render inside the body styled correctly
- [ ] Ok button closes; ESC closes; backdrop click closes
- [ ] Focus returns to the trigger after close

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #512, opened 2026-04-24
feat(agent-pane): migrate AgentLaunchModal to modal-v2 primitive

AgentA-asaf:
## Summary
- First consumer migrated to \`element/modal-v2\` (added in PR #511)
- Validates the new API against a real caller
- Zero visual regression intended; gains accessibility + focus management for free

## Context
Implements the \"migrate AgentLaunchModal\" step of [SPEC_ROBUST_MODAL_SYSTEM §6](../blob/main/docs/specs/SPEC_ROBUST_MODAL_SYSTEM_2026_04_23.md). Four more caller migrations follow (ImportPreviewModal, AboutModal, MessageModal, UserInputModal), then CommandPaletteModal + TypeAheadModal (which have trickier semantics), and finally retire \`element/modal.tsx\`.

### Diff
- Import swap: \`@/element/modal\` → \`@/element/modal-v2\` (+ \`ModalContent\` → \`ModalBody\` rename)
- \`<Modal onClickOut=...>\` → \`<Modal open={true} onClose={...} closeOnBackdropClick closeOnEscape size=\"md\">\`. While \`submitting\`, both close paths disable so the in-flight launch can't be accidentally aborted
- Removed the custom Escape handler and the manual focus timer on the name input — modal-v2 handles both natively (topmost-aware ESC + auto-focus-first-focusable after Portal mount)
- Kept the Enter-to-submit handler (form-specific, not modal-level)

### Inherited from modal-v2 with no extra wiring
- \`role=\"dialog\"\` + \`aria-modal=\"true\"\` + auto-wired \`aria-labelledby\` to the ModalHeader title
- Focus trap (Tab/Shift+Tab wrap inside panel)
- Focus save/restore (previous element regains focus on close)
- Background \`inert\` + body scroll lock (reference-counted for stacked-modal correctness)
- Backdrop blur + fade-in + pop-in animation, \`prefers-reduced-motion\` respected
- Multi-window Portal routing via \`document.activeElement?.ownerDocument\`

## Test plan
- [ ] Click a definition card in the agent picker → modal opens, name input gets focus
- [ ] Tab/Shift+Tab wrap inside the dialog (don't escape to the page behind)
- [ ] Backdrop click closes (unless submitting)
- [ ] ESC closes (unless submitting)
- [ ] Enter submits when name valid
- [ ] Closing returns focus to the clicked card
- [ ] No scroll on body while modal is open
- [ ] Screen reader announces \"dialog — Launch Claude Code\"

## Follow-ups
[...]

PR #516, opened 2026-04-24
feat(modals): migrate UserInputModal to modal-v2

AgentA-asaf:
## Summary
- Fifth modal-v2 consumer migration (after #512–#515)
- UserInputModal is the heaviest of the \`modals/Modal\` consumers — countdown timer, text/confirm branching, optional checkbox, Enter/Esc semantics

## Context
Per [SPEC_ROBUST_MODAL_SYSTEM §6](../blob/main/docs/specs/SPEC_ROBUST_MODAL_SYSTEM_2026_04_23.md) migration series.

### Diff
- Import swap: \`@/app/modals/modal\` → \`@/element/modal-v2\` (+ \`Button\` import)
- Old \`onOk\` / \`onCancel\` / \`onClose\` (three separate handlers on the old primitive) → explicit \`<ModalFooter>\` with Cancel + Ok buttons, plus \`onClose={handleSendErrResponse}\` on Modal for the ESC / backdrop / countdown-timeout paths (all three want the same \"canceled by user\" err response)
- \`.userinput-header <div>\` → \`<ModalHeader>\` since modal-v2's own \`.modal-panel-title\` rule supplies the bold + color the old div had. SCSS rule removed.
- Dropped the inline ESC-key check in \`handleKeyDown\` — modal-v2's \`closeOnEscape\` (default on) routes ESC to our \`onClose\`
- Countdown interval now uses \`onCleanup\` to clear on unmount, preventing the 300ms-delayed err-response from firing against a stale request if the user dismisses first

### Preserved behaviour
- Enter submits (via remaining keydown handler)
- Countdown title with live seconds remaining
- autofocus on the input field
- Text vs confirm response-type branching
- Optional checkbox

### Inherited from modal-v2
- \`role=\"dialog\"\` + \`aria-modal=\"true\"\` + auto-wired \`aria-labelledby\` to the header title
- Focus trap + save/restore
- Background \`inert\` + scroll lock (reference-counted)
- Backdrop blur + animations
- Multi-window Portal routing

## Test plan
- [ ] Text input flow: opens with input focused; Enter submits; Cancel / ESC / backdrop all send err-response; countdown decrements each second
- [ ] Confirm flow: opens without input; Ok sends confirm=true; Cancel sends confirm=false; ESC / backdrop / countdown timeout send err-response
- [ ] Checkbox flow: \`checkboxmsg\` non-empty renders checkbox; state included in the response
- [ ] Markdown query text: markdown flag renders via \`<Markdown>\` component
- [ ] Type check clean

## Remaining migrations
- CommandPaletteModal (keybinding disable during open)
- TypeAheadModal (block-ref mount target)
- ConnTypeAheadModal
- Add \`showCloseButton\` prop to modal-v2 + restore X on About
- Retire \`element/modal.tsx\` + \`modals/Modal\`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.347"
+version = "0.33.348"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.347"
+version = "0.33.348"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.347"
+version = "0.33.348"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.347"
+version = "0.33.348"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

a5af:
@codex review

PR #517, opened 2026-04-24
feat(modals): migrate CommandPaletteModal to modal-v2 + add placement/panelClass props

AgentA-asaf:
## Summary
- Sixth modal-v2 consumer migration. Command palette was the last modal using a custom Portal implementation
- Along the way: two small additive props on modal-v2 (\`placement\` + \`panelClass\`) to handle non-centered surfaces
- Net LoC: −0 or thereabouts; complexity moves out of the consumer and into the shared primitive

## Context
Per [SPEC_ROBUST_MODAL_SYSTEM §6](../blob/main/docs/specs/SPEC_ROBUST_MODAL_SYSTEM_2026_04_23.md) migration series.

### New modal-v2 props
1. **\`placement?: \"center\" | \"top\"\`** — sets the root grid's vertical alignment. \`top\` (15vh padding-top + \`align-items: start\`) matches the command-palette / quick-open pattern that drops from the top of the viewport.
2. **\`panelClass?: string\`** — appends a custom class to \`.modal-panel\` so consumers can override width / padding / overflow without sidestepping the primitive.

### CommandPaletteModal diff
- Drops its own Portal, backdrop div, and hand-rolled z-index wiring
- Drops inline ESC handler (modal-v2 → \`onClose\` handles it)
- Arrow keys + Enter navigation stay (palette-specific, not modal-level)
- Width / max-height / chrome applied via \`.command-palette-panel\` referenced from \`panelClass\`
- \`disableGlobalKeybindings\` / \`enableGlobalKeybindings\` stay (app-wide keybinding scope, not modal-level)

### Inherited from modal-v2
- role=dialog / aria-modal / aria-label=\"Command palette\"
- Focus trap + save/restore
- Background inert + reference-counted scroll lock (multi-window)
- Backdrop blur + animations (respects \`prefers-reduced-motion\`)

## Test plan
- [ ] Ctrl+P opens palette; input focused
- [ ] Arrow keys navigate results; Enter executes
- [ ] ESC closes; clicking backdrop closes
- [ ] Type filter narrows list; click selects
- [ ] Focus returns to the trigger after execute/close
- [ ] Global keybindings re-enable on close

## Remaining migrations
- TypeAheadModal (block-ref mount target — will likely need a \`mountTarget\` prop on modal-v2)
- ConnTypeAheadModal (similar to TypeAheadModal)
- Add \`showCloseButton\` prop to modal-v2 + restore X on About
- Retire \`element/modal.tsx\` + \`modals/Modal\`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.348"
+version = "0.33.349"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.348"
+version = "0.33.349"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.348"
+version = "0.33.349"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

PR #520, opened 2026-04-24
feat(modal-v2): add ConfirmModal preset

AgentA-asaf:
## Summary
- Completes the [robust-modal spec §5.1 API surface](../blob/main/docs/specs/SPEC_ROBUST_MODAL_SYSTEM_2026_04_23.md#51-api)
- Adds a \`ConfirmModal\` preset for the common title + body + Cancel/Confirm pattern
- Zero migrations in this PR — consumers adopt it ad-hoc as they replace \`window.confirm\` / hand-rolled confirm dialogs

## API
\`\`\`tsx
<ConfirmModal
    open={open()}
    title="Delete definition?"
    description="This cannot be undone."
    confirmLabel="Delete"
    destructive
    onConfirm={async () => { await deleteDefinition(); setOpen(false); }}
    onCancel={() => setOpen(false)}
/>
\`\`\`

## Behaviour highlights
- **Pending signal** — \`onConfirm\` runs through a pending state so a rapid double-click / submitted-then-ESC race is a no-op while the async work is in flight.
- **Close-lock during pending** — ESC + backdrop click disabled while \`pending\` is true (matches the convention used by AgentLaunchModal and ImportPreviewModal).
- **Destructive mode** — flips the confirm button to \`--error-color\` via \`.modal-btn--destructive\` AND routes initial focus to the Cancel button so a stray Enter after opening doesn't destructively confirm. Matches the \"destructive defaults to Cancel\" guidance in the spec open questions §9.1.
- **Default size** — \`sm\` (360 px), appropriate for short confirmation text.

## Styles
Button classes: \`.modal-btn\`, \`.modal-btn--confirm\`, \`.modal-btn--destructive\`. All token-backed:
- \`--space-1-5\` / \`--space-4\` — padding
- \`--text-sm\` / \`--font-weight-medium\` — type
- \`--radius-md\` — rounding
- \`--motion-fast\` — hover transition
- \`--accent-color\` / \`--error-color\` — semantic colour
- \`--shadow-focus-ring\` — keyboard focus

No raw literals introduced.

## Test plan
- [ ] Ad-hoc test by adopting in a caller: e.g. replace \`window.confirm(\"Delete agent…\")\` in \`AgentPicker.handleDelete\` (follow-up PR)
- [ ] Non-destructive flow: Confirm button is accent-coloured; initial focus on first-focusable (Confirm when body is empty → Confirm)
- [ ] Destructive flow: Confirm button red; initial focus on Cancel
[...]

PR #521, opened 2026-04-24
feat(agent-pane): use ConfirmModal for definition delete confirmation

AgentA-asaf:
## Summary
- First real consumer of the \`ConfirmModal\` preset added in #520
- Replaces the jarring native \`window.confirm\` + \`alert\` pair with an in-app modal that matches the rest of the UI
- Destructive mode → red confirm button + initial focus on Cancel

## Context
Continuing the modal-v2 migration polish. The delete-definition flow was the most visible remaining \`window.confirm\` in the agent pane (and the agent-picker is a top-level surface many users hit).

### Diff
- \`handleDelete\` split into:
  - Stage-on-click: sets \`deleteCandidate\` signal
  - \`handleDeleteConfirm\`: runs the actual DeleteForgeAgent RPC when the user confirms
- Error state stays inside the modal (inline \`.agent-picker-delete-error\`) instead of popping a separate \`alert()\` — users can retry or cancel without losing context
- Flagged \`destructive\` so the preset flips the confirm button to red and routes initial focus to Cancel

### Inherited from ConfirmModal / modal-v2
- \`role=\"dialog\"\` + \`aria-modal\` + focus trap + scroll lock + inert
- Backdrop blur + animations (\`prefers-reduced-motion\` respected)
- Multi-window Portal routing
- Pending lock during async onConfirm: no double-delete on rapid clicks; ESC / backdrop blocked while in flight

## Test plan
- [ ] Click 🗑 on a definition card → modal appears with agent name in title, Cancel focused
- [ ] ESC / Cancel / backdrop all dismiss without deleting
- [ ] Confirm deletes; modal closes; list refreshes via \`forgeagents:changed\` event
- [ ] Force an error (e.g. disconnect backend): modal stays open with error line; Cancel still works; no sibling alert
- [ ] Second rapid click during in-flight delete is a no-op (pending lock)

## Follow-ups
Three more \`window.confirm\`/\`alert\` callsites worth migrating:
- \`agent-view.tsx\` pane-close-with-tracked-processes
- \`swarm-view.tsx\` kill-all-under-agent
- \`AgentActionBar.tsx\` import-error alerts (these need a MessageModal-style toast rather than ConfirmModal)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.352"
+version = "0.33.353"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.352"
+version = "0.33.353"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.352"
+version = "0.33.353"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #522, opened 2026-04-24
feat(modals): migrate remaining window.confirm kill-tree callsites to ConfirmModal

AgentA-asaf:
## Summary
- Two more \`window.confirm\` → \`ConfirmModal\` migrations, following the agent-picker delete flow (#521)
- Both flagged \`destructive\` — red confirm button + initial Cancel focus

## Context
Continues the modal-v2 migration polish series. Every remaining kill-tree confirmation in the app now uses the in-app \`ConfirmModal\` preset.

### Migrated
1. **Agent pane close-with-tracked-processes** (\`agent-view.tsx\`)
   - The \`onClose\` wrapper previously blocked on \`window.confirm\`. Now stashes the count + original close callback into a \`closeConfirm\` signal; the ConfirmModal renders at the root of the presentation view.
   - Accept → runs \`AgentKillTreeCommand\` then invokes the original close
   - Cancel → does nothing, pane stays open
   - Modal stays mounted while the pane is alive, so the user isn't rushed

2. **Swarm kill-all per agent** (\`swarm-view.tsx\`)
   - \`killTree\` on \`SwarmActivityGroup\` used to pop \`window.confirm\` synchronously
   - Now sets \`confirmOpen\` and the ConfirmModal handles the decision

### Inherited from ConfirmModal / modal-v2
- role=dialog / aria-modal / focus trap / scroll lock / inert
- Backdrop blur + animations
- Multi-window Portal routing
- Pending lock — rapid click during in-flight kill is a no-op
- ESC / backdrop blocked while pending

## Remaining
Three \`alert()\` calls in \`AgentActionBar.tsx\` handle import/export errors. They're toasts, not confirmations — the correct target is a MessageModal or a dedicated toast surface. Separate PR.

## Test plan
- [ ] Close a pane with running processes → modal appears; ESC / Cancel keeps pane; Confirm kills + closes
- [ ] Swarm kill-all → same flow; modal closes after success; pending lock during RPC
- [ ] Close pane with zero processes → no modal, close happens immediately (unchanged)
- [ ] Type check clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.353"
+version = "0.33.354"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.353"
+version = "0.33.354"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.353"
+version = "0.33.354"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #524, opened 2026-04-24
feat(agent-pane): real popovers on definition-card ⓘ button (PR F)

AgentA-asaf:
## Summary
- Implements [PR F of SPEC_AGENT_DEFINITIONS_MODAL_2026_04_23](../blob/main/docs/specs/SPEC_AGENT_DEFINITIONS_MODAL_2026_04_23.md#7-implementation-plan)
- Replaces the native \`title\` tooltip on the card ⓘ button with the Popover primitive
- Users get the full \`catalog.popoverMarkdown\` deep-dive (startup flow, MCP, memory, context file) on click — styled, wrappable, not OS-truncated

## Context
Native tooltips are OS-capped (~200 chars), unstyled, and disappear on mouse-move. A real popover can show the complete catalog content we collected from discussion #493 during the design-system research phase.

### Changes
- **\`element/popover.tsx\`**: export \`PopoverContext\` + \`PopoverContextType\` so consumers can build custom triggers that keep their own chrome (instead of being locked into the default \`<Button>\` wrap of \`PopoverButton\`)
- **\`AgentCard\`**: new \`InfoPopoverTrigger\` component uses the context directly — registers the button ref for floating-ui positioning, wires click to \`togglePopover\`, keeps the existing \`agent-card-action-btn--info\` styling. Adds an \`--open\` modifier that mirrors the hover state while the popover is shown
- **Popover body** renders:
  - Bold title (CLI brand name)
  - Chip row: \"Context file\" = primary file (code style)
  - Chip row: \"MCP\" = stdio+http / stdio+http+oauth (hidden when \"none\")
  - Full prose body with \`--text-sm\` + \`--leading-normal\`
- **Placement**: \`right-start\` with 8px offset so the popover flows to the right of the card rather than overlapping the Launch interaction

### Tokens consumed
\`--space-*\`, \`--text-*\`, \`--radius-*\`, \`--shadow-lg\`, \`--font-mono\`, \`--leading-*\`. No raw literals introduced.

## Test plan
- [ ] Hover a definition card → ⓘ visible in top-left
- [ ] Click ⓘ → popover opens to the right with catalog content
- [ ] Click outside / press ESC → popover closes (inherited from Popover primitive)
- [ ] Clicking ⓘ on a different card closes the first one and opens the new one
- [ ] Popover renders correctly on each of the 7 CLIs (claude, codex, gemini, kimi, pi, openclaw, copilot)
- [ ] \"Context file\" + \"MCP\" chips show when catalog has that data; hidden otherwise
- [ ] Narrow pane: popover width (320px) doesn't clip against the window edge (floating-ui autoUpdate handles this)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.356"
+version = "0.33.357"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.356"
+version = "0.33.357"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.356"
+version = "0.33.357"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.356"
+version = "0.33.357"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

PR #526, opened 2026-04-24
feat(design-system): Phase 1 PR 4 — stylelint config + grandfather ignore

AgentA-asaf:
## Summary
- Closes Phase 1 of [SPEC_DESIGN_SYSTEM_2026_04_23](../blob/main/docs/specs/SPEC_DESIGN_SYSTEM_2026_04_23.md) — token foundation + lint enforcement
- Adds \`stylelint\` + \`stylelint-config-standard-scss\` and \`.stylelintrc.json\` enforcing the three spec-required rules
- Grandfathers all 54 existing SCSS files via \`.stylelintignore\` so no PRs break today; future migration PRs (Phase 3+) burn down the ignore list

## Rules enforced
| Rule | Effect |
|------|--------|
| \`color-no-hex\` | Raw hex literals forbidden — must use a token from \`theme.scss\` |
| \`declaration-property-value-allowed-list\` (z-index) | Only \`var(--z*-)\`, \`0\`, \`auto\`, \`unset\`, \`inherit\`, \`initial\`, or \`calc(…)\` permitted |
| \`declaration-no-important\` | \`!important\` forbidden outright |

Each rule carries a bespoke message pointing at the design-system spec section so a developer hitting the lint failure knows what to do.

## Files
- \`.stylelintrc.json\` — config (extends \`stylelint-config-standard-scss\` + spec rules + a few standard-config opt-outs to keep the noise floor low)
- \`.stylelintignore\` — 54 grandfathered SCSS files; new SCSS files are linted from day 1
- \`package.json\` — adds \`npm run lint:scss\` (with \`--allow-empty-input\` so the fully-grandfathered initial state is a green run)

## Test plan
- [ ] \`npm run lint:scss\` passes on current tree (everything ignored)
- [ ] Synthetic violator (\`z-index: 999\`, raw hex, \`!important\`) fails with the spec-referenced messages — verified locally
- [ ] CI integration: separate follow-up to wire \`lint:scss\` into the workflow once the test plan above is confirmed in review

## Burn-down plan
- Phase 3 PRs each remove one entry from \`.stylelintignore\` as they migrate the file's raw colours to tokens
- Phase 4 same pattern for spacing/typography
- Phase 7 PR removes the last entry; \`.stylelintignore\` shrinks to just the build-output / vendor lines

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #510, opened 2026-04-24
feat(design-system): Phase 2 — migrate global z-index literals to tokens

AgentA-asaf:
## Summary
- Implements [SPEC_DESIGN_SYSTEM_2026_04_23 §5.3](../blob/main/docs/specs/SPEC_DESIGN_SYSTEM_2026_04_23.md) Phase 2
- Tokenizes 5 global-overlay \`z-index\` literals using the \`--z-*\` hierarchy added in PR #507
- Intentionally leaves local stacking-context values alone (they're already correct and don't compete for a shared slot)

## Context
Phase 2 unblocks the robust-modal spec (which needs \`--z-modal\` consumed) and fixes the observed paint-order issues flagged in the audit (flyout beating modal, popover below everything, drag overlay covering modals).

### Migrated

| File | From | To | Net effect |
|------|------|----|------------|
| \`element/flyoutmenu.scss\` | \`1000\` | \`var(--z-flyout-menu)\` (9500) | flyout now paints above modals |
| \`element/popover.scss\` | \`1000\` (had TODO) | \`var(--z-popover)\` (8500) | popover paints above typeahead, below modals |
| \`element/zoomindicator.scss\` | \`10000\` | \`var(--z-context-menu)\` (10000) | value-preserving rename |
| \`drag/drag-overlay.scss\` | \`9999\` | \`var(--z-drag-overlay)\` (1000) | cross-app drag now below modals/popovers (spec intent) |
| \`modals/command-palette.scss\` | \`900\` / \`901\` | \`var(--z-modal)\` / same | joins modal slot; DOM order controls paint between backdrop + container siblings |

### Deferred

Local stacking-context values stay raw (not global competitors):
- \`block.scss:50\` (pane mask stacking context)
- \`blockstats.scss:5\` (matches existing \`--zindex-xterm-viewport-overlay\`, local)
- \`tab.scss:1\`, \`agent-view.scss:2/6/10/20/50/100/200\`, \`subagent-view.scss:10\`, \`term.scss:10\`, \`identity-view.scss:100\`
- \`window/action-widgets.scss:9999\` (unclear semantics; audit before migrating)

## Net user-visible changes
- **Drag overlay** no longer paints over open modals or popovers (previously was 9999, dominated everything). Matches spec intent: modal keeps focus during a cross-app drag, popovers stay readable.
- **Command palette** wins over popovers and sits at the modal level, consistent with its modal semantics.
- **Flyout menus** still beat everything except the zoom indicator (context-menu slot).

[...]

PR #527, opened 2026-04-24
feat(design-system): Phase 3 PoC — un-grandfather 3 element SCSS files

AgentA-asaf:
## Summary
- First Phase 3 migration PR. Validates the burn-down workflow on the smallest, lowest-risk targets
- \`.stylelintignore\` shrinks 54 → 51 grandfathered files
- Tunes off three stylistic standard-config rules that aren't in the spec-required set

## Files migrated
| File | LoC | Changes |
|------|-----|---------|
| \`element/copybutton.scss\` | 11 | Already clean — just removed from ignore |
| \`element/zoomindicator.scss\` | 44 | Already clean (z-index migrated in #510) — just removed from ignore |
| \`element/blockstats.scss\` | 32 | Dropped two hex fallbacks (\`#f59e0b\`, \`#ef4444\`) on \`var(--warning-color, …)\` / \`var(--error-color, …)\`; tokens are always defined in \`theme.scss\` so the fallbacks are dead. Migrated \`z-index: 5\` → \`var(--z-xterm-overlay)\` (same numeric value, semantic name now). |

## Stylelint config tuning
Disabled:
- \`rule-empty-line-before\`
- \`declaration-empty-line-before\`
- \`color-function-alias-notation\`

These are formatting preferences from \`stylelint-config-standard-scss\`, not part of the spec-required rule set (\`color-no-hex\`, z-index allow-list, \`declaration-no-important\`). Keeping the noise floor low matters when migrators are deciding whether a small PR is worth opening — a 4-line file shouldn't fail on \`rgba\` vs \`rgb\` semantics.

## Test plan
- [ ] \`npm run lint:scss\` clean (verified locally)
- [ ] No visual change — the only color literals removed were dead fallbacks; z-index value is unchanged

## Burn-down progress
\`.stylelintignore\` entries:
- Phase 1 PR 4 (#526): 54 grandfathered SCSS files
- This PR: −3 → 51 remaining


a5af:
@codex review

PR #538, opened 2026-04-24
feat(design-system): Phase 5 MVP — split agent-view.scss into 3 partials

AgentA-asaf:
## Summary

Extract the three largest disjoint sections of `frontend/app/view/agent/agent-view.scss` (4055 lines) into SCSS partials, reducing the file to ~2727 lines (33% reduction).

New files under `frontend/app/view/agent/styles/`:
- `_document-nodes.scss` (~700 lines) — markdown / code / diff / bash / message blocks
- `_setup-wizard.scss` (~360 lines) — provider / CLI / auth wizard
- `_slash.scss` (~260 lines) — slash-command picker / autocomplete / help

Consumed via `@use` from `agent-view.scss` in source order. Each partial wraps its content in `.agent-view { ... }` so the compiled selector chain (and specificity) is identical to the pre-split file.

## Why a separate spec

Design-system spec listed the target file structure but not the migration mechanics. This MVP validates the extraction mechanism on 3 sections before attempting the remaining 16. Full plan + decision rationale at `docs/specs/SPEC_AGENT_VIEW_SCSS_SPLIT_2026_04_24.md`.

## Validation

| Check | Result |
|---|---|
| Compiled CSS size | 240,638 bytes → **240,638 bytes** (identical) |
| Rule-set diff (`tr '}' '\n' \| sort`) | **2200 rules → 2200 rules, zero-line diff** |
| Stylelint | green |
| Build | succeeds |
| Selector disjointness vs remainder | **zero overlaps** (every extracted selector grepped against the remainder — none appear outside their partial) |

Byte-position order in compiled CSS differs (expected — `@use` emits at the top of the output, not at the original nested position), but all extracted selectors are fully disjoint from the rest of the file, so position change has no visual effect.

## Test plan

- [ ] Open an agent pane; verify message rendering, markdown blocks, diff blocks, tool blocks all look unchanged
- [ ] Open the setup wizard (create a new agent definition); verify layout, inputs, buttons
- [ ] Type `/` in the composer; verify slash picker renders, autocomplete works, help popover shows
- [ ] Hot-reload: edit a partial in `task dev`; verify change reflects without restart

## Follow-up

Remaining ~16 sections (card, picker, auth, focused-overlay, composer, tool-overlay, …) to extract in a follow-up PR using the same mechanism.


a5af:
@codex review

PR #543, opened 2026-04-24
fix(build): repair partial CEF extract before failing task package

AgentA-asaf:
## Summary

`task package` regressed mid-day on 2026-04-24 — v0.33.372 built green at 03:41; every attempt from ~05:43 onward failed with:

```
Error: File I/O error: Access is denied. (os error 5)
...
CMake Error at CMakeLists.txt:221 (add_subdirectory):
  "...\cef_windows_x86_64\libcef_dll" which is not an existing directory.
```

No Rust code or Cargo.lock changes since v0.33.372 — the regression is environmental.

## Root cause

`download-cef 2.3.1` finishes its extraction with a chain of `fs::rename` calls that move subdirectories from `cef_binary_*/` into the CMake-expected layout `cef_windows_x86_64/`. Once the first batch lands (~100 MB of Chromium DLLs from `Release/`), Windows Defender's real-time scan holds handles that can cause the **next** `MoveFileExW` calls for `include/` and `libcef_dll/` to return `ERROR_ACCESS_DENIED`.

It's a race. v0.33.372 won it; v0.33.373+ lost it until this fix. Full write-up at `docs/retros/RETRO_CEF_BUILD_RACE_2026_04_24.md`.

## What this PR does

Adds `scripts/repair-cef-extract.sh` — an idempotent, non-destructive repair that:

- Locates `target/release/build/cef-dll-sys-*/out/cef_windows_x86_64/`
- Checks for missing or empty `libcef_dll/` / `include/` / `Resources/`
- If missing, copies from the sibling `cef_binary_*/` raw extract (which always survives the race intact)

Wires it into `build:host:windows`:

```diff
-           - cargo build --release -p agentmux-cef -p agentmux-launcher
+           - cmd: bash -c 'cargo build --release -p agentmux-cef -p agentmux-launcher || { bash scripts/repair-cef-extract.sh && cargo build --release -p agentmux-cef -p agentmux-launcher; }'
```

On a green race: no-op. On a lost race: restores the missing subdirs so the second `cargo build` clears the CMake step without re-hitting the download path.

## Verification

Manually reproduced the failure mode, verified the script repairs it, verified `task package` now completes end-to-end. v0.33.378 portable ZIP already on Desktop.

## Alternatives considered

- **Windows Defender exclusion** on `C:\...\target` — per-machine config, not reproducible.
- **Upstream patch to `download-cef`** — right place for the fix, but needs forked pin until merged. Can still be filed as follow-up.
- **This PR:** in-tree retry-with-repair. Zero per-developer config. Behaves as a no-op in the common case. Easy to rip out if/when upstream fixes it.

## Test plan

- [ ] Run `task package` — succeeds (on this machine it already does with current target/ state)
- [ ] `rm -rf target/ dist/ && task package` — succeeds (forces fresh extraction + potential race)
- [ ] `bash scripts/repair-cef-extract.sh` on a clean tree — exits 1 (no build dir) without side effects
- [ ] `bash scripts/repair-cef-extract.sh` after a successful build — exits 0 without touching anything

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.377"
+version = "0.33.378"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.377"
+version = "0.33.378"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.377"
+version = "0.33.378"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.377"
+version = "0.33.378"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

PR #544, opened 2026-04-24
fix(modal): clip native panes behind modal-v2 backdrop

AgentA-asaf:
## Summary

Browser panes are CEF `BrowserView` child HWNDs composited by Windows **above** the main HTML renderer's surface — CSS `z-index` has no authority over them. When a modal opens while a pane is visible, the pane HWND paints on top of the modal backdrop and panel.

The airspace-clipping primitive that the widget-bar `MoreDropdown` already uses (`usePaneOverlay` in `frontend/app/platform/pane-overlay.ts`) maps cleanly to modal-v2 — register the `.modal-root` rect on open, deregister on close, and the backend subtracts that rect from every pane's Win32 region via `SetWindowRgn(RGN_DIFF)`. The pane paints transparent where the modal is.

## Changes

Two small deltas, both in frontend — no backend, no new IPC:

1. **Enhance `usePaneOverlay`** with a `window.resize` listener so viewport-sized overlays (the modal-root's `position: fixed; inset: 0`) follow window resizes. Additive; `MoreDropdown` inherits the same self-healing for free.

2. **Capture `rootRef` on `.modal-root`** and mount a `ModalPaneOverlayClip` helper inside `<Show when={mounted()}>`. `onMount` / `onCleanup` fire **per-open**, not per-instance — stacked and out-of-order modal closes peel off their clip rects cleanly. The backend already handles multi-rect unions.

## Why clip rather than hide

Considered "hide the browser view entirely while a modal is open":
- Jarring UX (every confirm dialog blanks the pane)
- Flicker on close (unhide = reload/repaint)
- Doesn't compose with partial-viewport overlays

Clipping only subtracts where the modal actually paints. Matches what MoreDropdown already does. Same primitive.

## Spec

`docs/specs/SPEC_MODAL_PANE_CLIP_2026_04_24.md` — full rationale, risks, follow-ups (per-window scoping, rAF coalesce, future DOM context menus).

## Test plan

- [ ] Open a browser pane, trigger any modal (About, ConfirmModal delete, launch modal) → modal paints over the pane, no pane content visible through panel or backdrop
- [ ] Resize window while a modal is open → backdrop follows viewport, clip follows
- [ ] Stack modals (e.g. open a ConfirmModal over another modal); close inner → pane stays clipped; close outer → pane restored
- [ ] MoreDropdown still works (same hook; resize now additionally refreshes its clip)
- [ ] No regression in non-Windows (hook is no-op on macOS/Linux; backend is `#[cfg(target_os = "windows")]`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.377"
+version = "0.33.378"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.377"
+version = "0.33.378"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.377"
+version = "0.33.378"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

PR #545, opened 2026-04-24
fix(overlay): scope pane clip to the requesting window

AgentA-asaf:
## Summary

Follow-up to #544 — addresses the Codex P1 ("Scope modal pane clip to owning window").

`browser_panes_set_overlay_clip` applied its rects to every live pane globally, so opening a modal in one CEF window would also clip panes in other windows. The bug exists for any consumer of `usePaneOverlay` — MoreDropdown is affected too — but modal-v2 amplifies it (persistent full-viewport clip vs. a transient small rect). This PR fixes both at once by scoping at the primitive level.

## Changes

Same pattern as the existing `main_window_focus` handler.

**Backend** (`agentmux-cef/src/browser_panes.rs`, `ipc.rs`):
- `set_pane_overlay_clip` takes a `window_label: &str`.
- IPC handler extracts `window_label` from args (defaults to `"main"` for back-compat if an older frontend omits it).
- Filter: resolve the requesting window's top-level HWND via `state.browsers[window_label] → host → GetAncestor(GA_ROOT)`, then skip any pane whose own top-level differs.
- Empty label string falls through with no filter — matches pre-scoping behaviour for callers that don't know their window.

**Frontend** (`frontend/app/platform/pane-overlay.ts`):
- `sendClip()` reads `?windowLabel=` from `window.location.search` (set by the CEF launcher on each secondary window's startup URL; falls back to `"main"`).
- Every overlay registration — modal backdrop, MoreDropdown, any future consumer — now publishes the label it belongs to.

## Why this PR is small

No new IPC surface, no new frontend API. The existing URL-query `windowLabel` plumbing is already consumed by `toggle_devtools`, `main_window_focus`, and `getWindowLabel()` in `frontend/util/cef-api.ts:407-410` — I'm just following the same path.

## Behaviour matrix

| Caller | Today (without fix) | With fix |
|---|---|---|
| Single-window (all users today) | Clips apply to every pane | Clips apply to every pane (identical — the one main HWND matches itself) |
| Multi-window, modal in window A | Clips panes in B + C too (bug) | Clips only panes in A |
| Legacy caller omitting `window_label` | N/A (all callers were legacy) | Defaults to `"main"` on backend, filter still works |

## Test plan

- [ ] Single-window: modal opens, backdrop paints over pane (same as #544 post-merge)
- [ ] Single-window: MoreDropdown opens, pane clips under it
- [ ] Multi-window (if reachable): open pane in window A + window B; open modal in A → only A's pane clipped, B stays live
- [ ] `cargo check -p agentmux-cef` green (includes non-Windows stub signature update)

## Related

- #544 — modal-v2 wired into the airspace clip (where this bug was surfaced)
- Codex review comment on #544 `frontend/app/element/modal-v2.tsx:343`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.379"
+version = "0.33.380"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.379"
+version = "0.33.380"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.379"
+version = "0.33.380"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.379"
+version = "0.33.380"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

PR #550, opened 2026-04-24
docs(spec): status-bar — always-visible network + responsive 2-row layout

AgentA-asaf:
## Summary

Amends `docs/specs/SPEC_STATUSBAR_TOKEN_USAGE_2026_04_24.md` with two behaviours the user wants covered before implementation starts.

### §4.3 Always-visible network indicator

Today the network-activity widget (↑/↓) hides itself when both in and out are `0/0`. A brand-new user sees a missing widget and can't tell whether the app is talking to the backend at all. Change: keep it mounted always; render the zero state as a muted `↑0 ↓0`, same pattern as the token-usage indicator's zero state.

### §4.4 Responsive 2-row layout on narrow widths

When the window narrows (or zoom-in), the left-group stats overlap the right-group info (network, version, token usage). Change: wrap to two rows — stats on top (left-justified), network+version+token-indicator on the bottom (right-justified). Each group keeps its internal justification; the bar's total height grows from 1 to 2 rows. Two implementation options (container query vs. `flex-wrap`) proposed; option 2 recommended.

## What this PR does

Docs-only. Inserts the two sections, renumbers `4.3 Interaction model` → `4.5`. No code change.

### DIFF
--- a/docs/specs/SPEC_STATUSBAR_TOKEN_USAGE_2026_04_24.md
+++ b/docs/specs/SPEC_STATUSBAR_TOKEN_USAGE_2026_04_24.md
@@ -126,7 +126,72 @@ Layout:
   `ConfirmModal` (reuses modal-v2's preset) with destructive
   styling, since resetting loses the running total.
 
-### 4.3 Interaction model
+### 4.3 Always-visible network indicator
+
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #549, opened 2026-04-24
feat(agent-pane): zone reorder + enriched Worked footer

AgentA-asaf:
## Summary

Two tweaks in the agent pane per `docs/specs/SPEC_AGENT_PANE_ZONE_ORDER_WORKED_FOOTER_2026_04_24.md`:

### 1. Queue moves directly below the feed

Today the pending queue renders *below* the "Working…" status line and activity log. When a user types while a turn is running, the typed text "disappears" into the composer and reappears on the far side of two unrelated zones.

New DOM order inside `.agent-view`:

```
- AgentDocumentView (feed)
- AgentRetryBar (conditional)
- PendingMessagesPanel      ← MOVED HERE (was below activity log)
- AgentStatusLine           ← "Working…" / "Worked · …"
- ActivityLogPanel
- composer-region
```

Queue gets a dashed top border + solid bottom border to visually pin it to the feed above.

### 2. Enriched "Worked" footer

| State | Today | This PR |
|---|---|---|
| Running | `⠋ Working…                  ↑12k ↓3k · 42s` | *(unchanged)* |
| Completed | `✓ Worked · $0.018 · 42s · 4 turns` | `✓ Worked · 42s · ↑12k ↓3k          $0.018 · 4 turns` |

Tokens are what users budget against and carry forward into the next turn's context window. Duration sits next to them as the paired metric. Cost + turn count move to a secondary group on the right (`--secondary-text-color`). Transition Working → Worked no longer jumps the layout — both branches use the same two-group flex.

## No backend change

Uses existing `SessionStats` (`cost_usd`, `duration_ms`, `num_turns`) and `TurnTokens` (`input`, `output`). Pure render-layer.

## Validation

| Check | Result |
|---|---|
| `tsc --noEmit` | clean |
| `task build:frontend` | succeeds |
| `npm run lint:scss` | green |

## Test plan

- [ ] Send a message, start a second while first runs → queue appears between feed and status line
- [ ] Turn completes → footer shows `Worked · 42s · ↑12k ↓3k` with `$0.018 · N turns` on the right
- [ ] Resize pane narrow → right group wraps or drops gracefully (no overflow)
- [ ] Tokens missing (older stream replay) → falls back to today's display

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #552, opened 2026-04-24
feat(statusbar): token usage indicator + breakdown popover

AgentA-asaf:
## Summary

Implements `docs/specs/SPEC_STATUSBAR_TOKEN_USAGE_2026_04_24.md`.

## New surface

- **`🪙 ↑Xk ↓Yk` indicator** in the status-bar right group. Live running total, session-local. Muted zero state so the widget stays visible as "nothing aggregated yet" rather than hidden.
- **Click → breakdown popover** anchored under the indicator. Lists per-service totals (Claude Code, Codex CLI, Gemini, Kimi, Pi, OpenClaw, GitHub Copilot) sorted by total desc, a bold Total row, and a destructive-gated "Reset counter" button (ConfirmModal preset).
- Popover calls `usePaneOverlay` so it paints cleanly over any browser pane — same airspace primitive as modal-v2 (#544) and MoreDropdown.

## Plumbing

- **Store:** new `frontend/app/store/token-usage.ts` aggregates `TurnTokens` keyed by provider id.
- **Hook:** `useAgentStream` takes an optional `provider` prop and calls `recordTurn(provider, tokens)` inside `finalizeTurn` (after the snapshot added in #549). `agent-view.tsx` passes the catalog provider key.

## Spec §4.3 — network indicator always visible

`SystemStats` no longer hides the net row at `0/0`. Zero state gets a `.stat-idle` class that mutes opacity.

## Spec §4.4 — responsive 2-row wrap

`StatusBar` flex container uses `flex-wrap: wrap` + `justify-content: space-between` with `align-self: flex-start` / `flex-end` on the two groups. Narrow window → right group falls to a new row anchored right; both rows preserve internal justification.

## Validation

| Check | Result |
|---|---|
| `tsc --noEmit` | clean |
| `task build:frontend` | succeeds |
| `npm run lint:scss` | green |

## Test plan

- [ ] Launch an agent, run a turn → indicator lights up (no longer muted) and increments
- [ ] Launch a second agent of a different provider → breakdown shows both
- [ ] Click indicator → popover opens below, above any visible browser pane
- [ ] Reset counter → confirm modal → counts return to 0, indicator goes muted again
- [ ] Resize window narrow → status bar wraps: left group on row 1, right group on row 2 (flush right)
- [ ] Leave app idle → network widget shows muted `↑0 ↓0` (doesn't disappear)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.386"
+version = "0.33.387"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.386"
+version = "0.33.387"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.386"
+version = "0.33.387"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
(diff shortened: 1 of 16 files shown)


a5af:
@codex review

Issue #551, opened 2026-04-24
Decision prompt: per-tool-call permission UI for agent panes


AgentA-asaf:
## Problem

AgentMux gates tool calls **before** an agent turn (via `PermissionMode` → CLI flags in `frontend/app/view/agent/buildRuntimeArgs.ts:18-24`), not **during** it. When the user is in `default` / `auto` / `acceptEdits` / `plan` mode and the agent CLI asks itself for permission on a specific tool call, **AgentMux never sees the question and the user never sees a prompt**.

Concretely:
- `ToolCallEvent` / `ToolResultEvent` (`frontend/app/view/agent/types.ts:256-271`) have no `pending_approval` state — only `running | success | failed`.
- There is no decision UI component.
- There is no `tool:decision` RPC handler in `agentmux-srv/src/server/websocket.rs`.
- No provider adapter parses `permission_request` stream events from the sidecar.

**Result:** the sidecar hangs at stdin waiting for a `y/n` it will never receive, and the agent block stalls in the `running` (⏳) state with no visible cause. The only workable mode today is `bypass` (`--dangerously-skip-permissions`), which defeats the point of having a permission system.

## Proposal

Introduce a per-tool-call **decision prompt** with:

- New `PermissionRequestEvent` stream event + `pending_approval` tool status.
- Decision panel anchored to the focused agent pane: tool, target, reason, risk, diff/bash preview.
- Keyboard-driven actions: Allow once / session / project / global / Deny / Feedback / Defer.
- Scoped allow-rules persisted to `.agentmux/permissions.json` (project) and `~/.agentmux/permissions.json` (global).
- Risk classifier (low / medium / high) with extra confirm step on high-risk.
- New `tool:decision` websocket RPC, forwarded by `BlockController` to the sidecar's stdin.
- Provider adapters: Claude Code (`permission_request` stream events), Codex, ACP. Gemini/Kimi remain bypass-only until upstream gains a per-call protocol.
- Append-only audit log at `.agentmux/decisions.log` (JSONL).
- New `decisionPrompt` settings block in `settings-template.jsonc`.

Pre-turn `PermissionMode` is preserved as the *fallback* when no rule matches and no UI is present.

## Spec

Full spec: `specs/SPEC_DECISION_PROMPT_2026_04_24.md` (drafted locally — happy to open as a PR if the direction is right).

[...]

PR #553, opened 2026-04-24
feat(statusbar): always-show connection widget + instant tooltips

AgentA-asaf:
## Summary

Two small fixes to the status bar plus a companion spec for issue #551.

### 1. ConnectionStatus no longer disappears at 0 remote

User feedback: *"the network still disappears... we need it to always show"*. The existing network (SystemStats) widget was already fixed in #552 — but `ConnectionStatus` (the separate widget showing SSH/WSL remote connections) had its own `<Show when={total() > 0}>` gate and still vanished. Now it renders always; zero state gets `status-bar-item--idle` class for muted opacity.

### 2. Instant tooltips on every widget

User feedback: *"instant tool tips on every metric"*. Native `title=` has a browser-controlled ~500 ms delay. Replaced with a CSS-only `[data-tip]:hover::after` pattern — fires on the same hover event the browser registers, no perceived delay.

Coverage: every status-bar widget now carries a `data-tip` + `aria-label`:
- BackendStatus
- SystemStats (per-metric: CPU, GPU, Mem, Net, Disk — each gets its own tooltip)
- ConnectionStatus
- ConfigStatus
- UpdateStatus
- HostPopover
- TokenUsageIndicator
- Version (both clickable + offline states)

A11y preserved: `aria-label` stays on every widget for screen readers; CSS tooltip is decorative (`pointer-events: none`).

### 3. Companion spec: `SPEC_DECISION_PROMPT_2026_04_24.md`

Extended spec for agentmuxai/agentmux#551 (per-tool-call permission UI). Adds a new **"Remember last options"** section (§7) covering three layers of memory — per-(tool, target), per-pane, per-user-global — so the second and third times the user hits the same prompt take one keystroke instead of full navigation. No code yet; spec only.

## Validation

| Check | Result |
|---|---|
| `tsc --noEmit` | clean |
| `task build:frontend` | succeeds |
| `npm run lint:scss` | green |

## Test plan

- [ ] Hover any status-bar widget → tooltip appears immediately (no ~500ms wait)
- [ ] Leave app idle with no remote SSH/WSL sessions → `0 remote` widget stays visible (muted)
- [ ] All existing click-actions still work (opening popovers, new window, etc.)


a5af:
@codex review

PR #558, opened 2026-04-26
feat(tab-bar): first-principles rewrite + sub-pixel rendering scaffolding

AgentA-asaf:
## Summary

Two layered changes that together fix the long-standing "tab gaps drift inconsistently" symptom:

1. **Tab bar first-principles rewrite** — real-DOM `<div class=\"tab-separator\">` siblings instead of pseudo-element bars, variable-width tabs that collapse on rename, default blue tab color, VS Code close icon, drag-fix for the broken-new-window bug, T1→tab1 naming on both code paths.
2. **Sub-pixel rendering scaffolding** — additive design-system primitives that stop the bug class app-wide, not just on tabs. Three new tokens (`--dpr`, `--snap`, `--snap-chrome`), a `v-separator()` mixin that uses `border-right` to opt into spec-defined pixel snapping (Brosset 2024), `snap()` / `snap-chrome()` SCSS functions, and a `dpr.ts` bootstrap that keeps `--dpr` synced with `devicePixelRatio` via `matchMedia(resolution)` change events. Tab bar migrated to consume the primitives; ready to spread to status-bar dividers / sidebar handles in follow-up PRs.

The research that drove the scaffolding design is in `docs/retros/RETRO_SUBPIXEL_RENDERING_RESEARCH_2026_04_26.md` — covers Adobe Spectrum's static/scaled token bifurcation, Apple's `backingAlignedRect:` API as the conceptual ancestor of CSS `round()`, why Electron-class apps drive Chromium's device-scale-factor instead of CSS `zoom`, and a 5-phase migration plan for further leverage.

## Test plan

- [ ] Open AgentMux, hover over the title bar, hold Ctrl and scroll to step through zoom levels (1.0×, 1.1×, 1.25×, 1.33×, 1.5×). Verify tab-to-separator distances stay even across all tabs at every zoom.
- [ ] Rename a tab to a shorter / longer name and confirm gaps stay constant (no padding-growth).
- [ ] Drag a tab within the window and verify the reorder happens cleanly (no broken new window spawned).
- [ ] Open DevTools and inspect `:root` — confirm `--dpr` is set to your monitor's actual DPR.
- [ ] On a non-1.0 DPR display (laptop Retina, scaled Windows display), zoom levels should still keep gaps even.
- [ ] Default new tab (+ button) is blue, named `tabN`, with the active tab showing a VS Code-style X.

## Notes for reviewers

[...]

PR #559, opened 2026-04-26
feat(tab-bar): default new-tab preset + tear-off Phase 1 (threshold detection)

AgentA-asaf:
## Summary

Two additive changes to the tab strip:

1. **Default new-tab preset** — clicking the \"+\" button now creates a tab pre-laid-out with three panes: agent on the left half, sysinfo top-right, swarm bottom-right. The layout tree lives in a single central module (`frontend/app/tab/tab-presets.ts`) so any future tab-creation path (duplicate, tear-off destination, startup-tab backfill) can reuse it. To change the defaults, edit only the `DEFAULT_TAB_PRESET` constant.

2. **Tear-off Phase 1 — threshold detection** (stub only). Adds a `TEAR_PAST_PX = 24` check in `tabbar.tsx` `monitorForElements.onDrag` that fires `requestTearOff()` once per drag past the strip's bottom edge. The stub currently logs only — Phases 2-6 (host IPC, SC_MOVE handshake, mouse hook, merge/cancel) are deferred per the spec's phased plan. **Important context discovered during Phase 1 verification:** the existing `[dnd:cef] start_cross_drag` → `[dnd:svc] TearOffTab` pipeline already does Win32-level tear-off via the same `ReleaseCapture` + `SendMessageW(WM_NCLBUTTONDOWN, HTCAPTION)` idiom at `commands/window.rs:162-167`. Whether to layer the Chrome-style SC_MOVE flow on top is a design decision deferred to a follow-up PR.

Spec: `docs/specs/SPEC_TAB_TEAR_OFF_SIZE_PRESERVATION_2026_04_26.md`

## Stacked on #558

This PR is branched off #558 (`agenta/tab-gaps-naming`). The diff against `main` includes #558's commits until #558 merges, after which a rebase will reduce the diff to just the new work here.

## Test plan

- [ ] Click \"+\" — new tab appears with agent (left) + sysinfo (top-right) + swarm (bottom-right). Renames + per-pane resizes work as expected.
- [ ] Drag a tab past the strip's bottom edge — log line `[dnd] tab tear-off threshold crossed (Phase 1 stub)` fires exactly once per drag.
- [ ] Drag a tab within the strip — no log fires; normal reorder behaviour.
- [ ] Drag a tab and drop outside the window — existing tear-off pipeline still works (spawns a new window with the tab).

## Notes for reviewers

- The dynamic `import(\"@/app/tab/tab-presets\")` in `global.ts createTab()` avoids a circular dep (tab-presets imports from global). Standard pattern in this codebase.
- `getLayoutModelForTabById` was already defined in `layoutModelHooks.ts` but unexported; this PR makes it public so the preset applier can target a specific (newly-created) tab without racing the activeTabId atom update.
- The Phase 1 stub introduces a one-shot `tearOffFired` latch + an early-return in `onDrop` — both safe given the current `requestTearOff` is a logger; the latch reset in `onDrop` keeps subsequent drags clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.398"
+version = "0.33.399"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.398"
+version = "0.33.399"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.398"
+version = "0.33.399"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.398"
+version = "0.33.399"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.398"
+version = "0.33.399"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
--- a/agentmux-common/Cargo.toml
+++ b/agentmux-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-common"
-version = "0.33.398"
+version = "0.33.399"
 edition = "2021"
 description = "Shared utilities for AgentMux crates"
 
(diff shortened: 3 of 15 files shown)


a5af:
@codex review

AgentA-asaf:
**Round-2 reagent review — onDrag flag is a false-positive.**

The early-return on `tearOffFired` in `onDrag` was removed in commit `eb8544cf`. Current state at `tabbar.tsx:129-143`:

```ts
onDrag: ({ source, location }) => {
    const input = location.current.input;
    const rect = tabBarScrollRef?.getBoundingClientRect();
    if (rect && !tearOffFired && input.clientY > rect.bottom + TEAR_PAST_PX) {
        tearOffFired = true;
        const draggedTabId = source.data.tabId as string;
        requestTearOff(draggedTabId, input.clientX, input.clientY);
        // FUTURE Phase 2: return here once requestTearOff
        // actually hands off to the host. For now (stub only),
        // fall through so insertionPoint keeps tracking — a
        // drag that dips past the threshold and returns to the
        // bar still gets the reorder gap.
    }
    setInsertionPoint(computeInsertionPoint(input.clientX));
},
```

There is no `if (tearOffFired) return;` — control falls through to `setInsertionPoint` on every drag tick. A drag that crosses 24px below the strip, then returns to the strip, will get the normal reorder gap and a normal `onDrop`.

**TOCTOU on tab-presets.ts:** addressed in `62dc2ded` with a post-await re-check + best-effort `DeleteBlock` cleanup. The complete structural fix (backend `CreateBlock` accepting an explicit `tab_id` arg) is still a follow-up since it touches `agentmux-srv` and is out of scope for this PR.

PR #560, opened 2026-04-27
feat(statusbar): version-click instance panel

AgentA-asaf:
## Summary

Replaces the version chip's old \"click → openNewWindow\" side-effect with an anchored popover surfacing about-info + open-windows in this AgentMux process + actions. Chrome-style multi-window awareness from a chip that was previously a one-trick affordance.

**What's in the panel:**
- **About:** version, build time, runtime (\`platform · arch\`) with a copy-to-clipboard for bug-report convenience.
- **This process — N windows:** one row per open window. Current window shows a \"this\" badge and is non-clickable; other rows are click-to-focus via the existing \`getApi().focusWindow(label)\` RPC.
- **Footer:** \`+ Open another window\` (the old primary action, now accent-coloured) and \`Close\`.

**Plumbing reuse (no new RPCs needed):**
- \`getApi().focusWindow(label)\` already exists at \`agentmux-cef/src/commands/window.rs:324-348\`.
- \`window-instances-changed\` event already emits at \`window.rs:514\`; new \`openWindowLabelsAtom\` subscribes via the existing app-init listener and re-fetches \`getApi().listWindows()\` on every event.
- \`usePaneOverlay\` primitive paints over any browser pane HWND (same airspace pattern as TokenBreakdownPopover, MoreDropdown, modal-v2).
- Esc + click-outside close handler mirrors HostPopover / TokenBreakdownPopover precedent.

**V1 deferrals (called out in the spec):**
- LAN peers — \`HostPopover\` already shows them with hover-rich detail; duplicating here would be confusing UX.
- Per-window token totals — needs a token-usage refactor (currently global / session-scoped). Separate PR.

Spec: \`docs/specs/SPEC_VERSION_INSTANCE_PANEL_2026_04_25.md\` (now committed, with revision notes on three drift items).

## Test plan

- [ ] Single-window: click \`vX.Y.Z\` in the status bar → panel opens with about-info + 1 row \"Window 1\" (badged \"this\", disabled). Footer button \"+ Open another window\".
- [ ] Click \"+ Open another window\" → spawns a second AgentMux window. Re-open panel → now shows 2 rows.
- [ ] Click the non-current row → that window focuses.
- [ ] Esc closes the panel. Click-outside closes the panel. Re-clicking the chip toggles open/closed.
- [ ] Open a browser pane behind the status bar, then open the panel → panel paints cleanly over it (airspace works).
- [ ] Click each copy icon → clipboard receives \`version: v0.33.400\` etc.
[...]

PR #562, opened 2026-04-27
feat(backend): explicit tabId arg on CreateBlock — closes tab-presets TOCTOU

a5af:
## Summary

Follow-up to **#560** that closes the TOCTOU race the tab-presets module had to work around.

PR #560's \`tab-presets.ts\` calls \`ObjectService.CreateBlock\` to seed each block in a freshly-created tab. The server routes \`CreateBlock\` via \`uicontext.active_tab_id\` — which is captured at the moment the RPC envelope is built. If the user clicks another tab between the frontend's pre-call \`activeTabStillTargets\` check and the RPC arriving on the server, blocks land in the wrong tab. The frontend mitigations (check-before-await, post-await re-check + DeleteBlock cleanup) narrowed the window but couldn't close it structurally.

This PR closes it at the source: an optional \`tabId\` arg on \`CreateBlock\` (args[2]) overrides \`uicontext.active_tab_id\` when present.

## Changes

**Backend (Rust):**
- \`agentmux-srv/src/server/service.rs\` — handler reads \`get_optional_arg::<String>(args, 2)\`. When present and non-empty, that's the target tab; otherwise fall back to \`uicontext.active_tab_id\` (existing behaviour).
- \`agentmux-srv/src/backend/service.rs\` — arg_names metadata extended with \`\"tabId\"\`.

**Frontend (TS):**
- \`services.ts\` — \`CreateBlock(blockDef, rtOpts, tabId?)\` signature.
- \`tab-presets.ts\` — passes \`expectedTabId\` explicitly. Removes the \`activeTabStillTargets\` guard, the post-await re-check, and the \`DeleteBlock\` orphan-cleanup — none needed once the routing is deterministic.

**Backward-compatible.** Existing two-arg callers (every caller other than tab-presets) continue to route via \`uicontext\`.

## Stacked on #560

This PR's base is \`agenta/version-instance-panel\` so the diff is just the new work. After #560 merges, this PR's diff against main reduces to these 4 files.

## Test plan

- [ ] \`cargo check -p agentmux-srv\` — clean (verified locally).
- [ ] \`tsc --noEmit\` — clean.
- [ ] Open AgentMux, click \"+\" → tab gets agent + sysinfo + swarm preset (same as #560 behaviour).
- [ ] Repeat the test rapidly clicking between tabs during \"+\" press → preset blocks land in the new tab regardless of click timing (the previous TOCTOU window is gone).
- [ ] No regressions in widget-picker block creation (still routes via uicontext for back-compat).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.403"
+version = "0.33.404"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.403"
+version = "0.33.404"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.403"
+version = "0.33.404"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.403"
+version = "0.33.404"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

a5af:
@gemini review

a5af:
@codex review

PR #563, opened 2026-04-27
feat(tab-tear-off): Phase 2 — Win32 SC_MOVE handshake

AgentA-asaf:
## Summary

Phase 2 of the tear-off spec at \`docs/specs/SPEC_TAB_TEAR_OFF_SIZE_PRESERVATION_2026_04_26.md\` (Phase 1 shipped in #559).

When the user drags a tab past the strip's bottom edge, the new window now appears at the cursor and **immediately follows the mouse via Windows' built-in modal move-loop** — Chrome-style, no ghost. Replaces Phase 1's logging-only stub.

## End-to-end flow

Frontend (\`tabbar.tsx\` \`requestTearOff\`) orchestrates 3 RPCs:

1. **\`WorkspaceService.TearOffTab\`** — sidecar transfers the tab into a new workspace. *Reused* from the existing dragend pipeline.
2. **\`api.openWindowAtPosition\`** — host spawns the destination window pointed at the new workspace. *Reused.*
3. **\`api.tearOffSCMoveHandshake\`** — *NEW host handler* (\`commands/drag.rs::tear_off_sc_move_handshake\`):
   - Polls \`state.browsers\` for the destination window's HWND (cold-path wait: ~150-300ms; Phase 6 warm pool drops to <16ms).
   - Win32: \`ReleaseCapture\` + \`SetForegroundWindow(dest)\` + \`SetCapture(dest)\` + \`PostMessageW(dest, WM_SYSCOMMAND, SC_MOVE | HTCAPTION, MAKELPARAM(x, y))\`.
   - Windows takes over the modal move-loop; the new window follows the cursor at full opacity until mouseup.

## Per-spec quality bar

Per §0:
- ✅ **Handshake budget ≤ 8 ms** — measured from HWND-ready to PostMessageW return; emitted as \`handshakeMs\` on the IPC result + logged under \`target=dnd:tearoff\`.
- ⚠️ **First-paint flash 0 ms** — NOT met by Phase 2 in isolation. Cold-path flash is expected and accepted per spec §9 \"Phase 2 (handshake + SC_MOVE) — internal milestone, not a user-facing release.\" Phase 6 (warm pool) closes this.
- ✅ **Old dragend pipeline kept intact** — per spec §0 (\"deleted only once Phases 2 + 4 + 5 are landed\"). The two paths trigger from different events (threshold-cross vs dragend) and don't conflict.

## Test plan

- [ ] Drag a tab past the strip's bottom edge.
[...]

PR #564, opened 2026-04-27
fix(tab-tear-off): Phase 2 codex follow-ups (PostMessage error + payload-clear timing)

AgentA-asaf:
## Summary

Follow-up to **#563** (Phase 2 SC_MOVE handshake) addressing two post-merge codex flags. Both are reliability fixes — no functional change on the happy path.

### P1 — Surface PostMessageW(SC_MOVE) failure (\`drag.rs:431\`)

The previous handler ignored \`PostMessageW\`'s BOOL return. If Windows rejected the post (dest HWND went invalid during the HWND-wait window, message queue full, etc.) no modal move-loop starts but the handler returned the metrics struct as if everything succeeded — the frontend treated tear-off as complete and the legacy dragend fallback never fired. Now we check the return and surface \`GetLastError\` so the frontend can fall back / log diagnostically.

### P2 — Defer drag-payload clear until handshake succeeds (\`tabbar.tsx:144\`)

\`setCurrentDragPayload(null)\` was firing in \`onDrag\` BEFORE the async tear-off pipeline started, so any failure mid-pipeline (\`TearOffTab\`, \`openWindowAtPosition\`, or the handshake itself) stranded \`CrossWindowDragMonitor\`'s \`dragend\` fallback with no payload — turning the gesture into a silent no-op. Moved the clear to fire AFTER \`tearOffSCMoveHandshake\` returns successfully (inside \`requestTearOff\`). The legacy dragend fallback now reliably picks up partial-failure cases.

## Test plan

- [ ] Happy path: drag past threshold → new window appears + follows cursor → SC_MOVE log fires with handshakeMs (unchanged behaviour from #563).
- [ ] Force a failure path (e.g. invalid workspaceId on TearOffTab) → no silent loss; legacy dragend fallback should fire on mouse release and the user still gets a tear-off (or a clear error).
- [ ] No regressions: existing \`dragend\` tear-off (drop outside the window without crossing strip-bottom) still works as before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.409"
+version = "0.33.410"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.409"
+version = "0.33.410"
 dependencies = [
  "tokio",
  "tracing",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #566, opened 2026-04-27
feat(tab-tear-off): Phase 6 — pre-warmed window pool (eliminates first-paint flash)

AgentA-asaf:
## Summary

Closes the §0 \"0 ms first-paint flash\" quality bar. Phase 2's cold path spawned a fresh CEF window on every tear-off (~150-300 ms while the renderer process started + first frame painted), producing a visible flash before the SC_MOVE handshake took over. Phase 6 pre-spawns N=2 hidden windows that are already painted and IPC-connected; on tear-off we POP one, reposition it under the cursor, show it, and emit \`pool:promote\` so the renderer bootstraps the workspace in-place.

**No reload, no renderer restart, no flash.**

### Backend (Rust)

- \`agentmux-cef/src/commands/window_pool.rs\` (new, ~190 lines):
  - \`spawn_pool_window\` — creates a hidden window at off-screen coords (-32000, -32000) with URL flag \`?pool=1\`. Single-flight via an \`AtomicBool\`.
  - \`register_pool_window\` — \`on_after_created\` calls this for \`pool-*\` labels; chains the next spawn if pool is below \`TARGET_SIZE=2\`.
  - \`init_pool\` — fires when the \"main\" window registers, kicks off the first pool spawn (chain takes care of the rest).
  - \`promote_pool_window\` — pops a label, \`SetWindowPos\` + \`ShowWindow\`, emits \`pool:promote\` event, enqueues refill.
- \`drag.rs::tear_off_pool_promote\` — new IPC. Returns label on success, error \`pool_exhausted\` on empty pool.
- \`state.rs\` — new \`window_pool: VecDeque<String>\` + \`window_pool_respawn_in_flight: AtomicBool\`.
- \`client.rs::on_after_created\` — wired to \`init_pool\` (on \"main\") and \`register_pool_window\` (on \`pool-*\`).

### Frontend (TS)

- \`frontend/app/init/pool.ts\` (new): \`isPoolMode()\` + \`awaitPoolPromote()\`. The promote handler pushes the workspaceId into the URL so the existing \`initHostNewWindow\` code path picks it up unchanged.
- \`frontend/app-init.ts\` — in \`initApp\`'s host-app branch, short-circuit before the \`isMainWindow\` check when in pool mode. Renderer renders the empty fallback and waits for promote.
- \`frontend/app/tab/tabbar.tsx\` — \`requestTearOff\` tries pool-promote first; on rejection falls back to \`openWindowAtPosition\` (existing cold path, now defence-in-depth).

[...]

PR #568, opened 2026-04-27
fix(tab-tear-off): pool windows leaking to taskbar

AgentA-asaf:
## Summary

Phase 6 (PR #566) shipped warm-pool windows at `(-32000, -32000)` to keep the body off-screen, but Win32 still gives top-level frameless windows taskbar entries by default. Each running AgentMux instance ended up with `POOL_TARGET_SIZE` (= 2) phantom taskbar entries, and right-clicking → Close on a phantom kicked off `on_pool_window_destroyed` → `spawn_pool_window`, instantly respawning a replacement. Whack-a-mole the user couldn't win.

## Fix

`register_pool_window` (called from `on_after_created`) now applies `WS_EX_TOOLWINDOW` (and clears `WS_EX_APPWINDOW`) on the freshly-registered HWND so the pre-warmed pool window doesn't appear in the taskbar or Alt+Tab.

`promote_pool_window` clears `WS_EX_TOOLWINDOW` (and re-adds `WS_EX_APPWINDOW`) before positioning + showing, so the *promoted* window — which becomes a real torn-off AgentMux instance — IS taskbar-visible like any other window.

The hide → write style → `SetWindowPos(SWP_FRAMECHANGED)` cycle is required: Windows doesn't always re-evaluate the taskbar entry on a bare `GWL_EXSTYLE` change.

## Why a focused PR

This was a real regression from #566 reported by the user when they tried to test Phase 5 cancel-back on a portable build. We branched off from a parallel attempt at window-size preservation (in `agenta/tab-tearoff-window-size`) that didn't land cleanly — that work needs a redesign with `GetAncestor(GA_ROOT)` + per-monitor DPI handling per the retro at `docs/retro/tab-tearoff-retro-2026-04-27.md`. Shipping this taskbar fix independently so users get the unblock without waiting on the size redesign.

## Test plan

- [ ] Launch a fresh AgentMux instance — verify only ONE taskbar entry appears (the main window). No phantom entries.
- [ ] Tear off a tab → verify the new window gets a taskbar entry.
- [ ] Close the main window → verify no respawning ghost windows in the taskbar.
- [ ] Tear off a tab → close it → verify the warm pool refills silently in the background (still 0ms first-paint flash on subsequent tear-offs).
- [ ] Alt+Tab through windows: pool windows should NOT appear; promoted/main windows SHOULD appear.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.430"
+version = "0.33.431"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.430"
+version = "0.33.431"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.430"
+version = "0.33.431"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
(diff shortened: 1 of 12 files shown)


a5af:
@codex review

PR #569, opened 2026-04-27
feat(InstancePanel): user-renameable window display names

AgentA-asaf:
## Summary

Implements [docs/specs/SPEC_WINDOW_RENAME_2026_04_27.md](docs/specs/SPEC_WINDOW_RENAME_2026_04_27.md).

InstancePanel rows now show a resolved display name with a priority chain:

1. User-set `window:displayname` in Window meta
2. Workspace.name (when the user has named the workspace)
3. Index-based fallback ("Window 1", "Window 2"...)

## Interactions

| Gesture | Action |
|---|---|
| Single click on a row | Focus that window (existing behavior) |
| Double click on a row | Enter inline rename mode |
| `Enter` while editing | Commit (trim, length-cap 64) |
| `Escape` while editing | Cancel; revert to pre-edit name |
| Click outside / blur | Commit |
| Empty / whitespace-only input | Silent revert |

## Plumbing

- **New host RPC** `list_window_instances` returns `[{label, windowId}]` so the panel can resolve per-window backend Window records (display name in meta, workspace fallback) without an extra round-trip per row. Frontend falls back to `list_windows` if the new RPC fails (older host).
- **New `openWindowEntriesAtom`** carries `{label, windowId}` pairs; `openWindowLabelsAtom` is kept for backward compat with CrossWindowDragMonitor (label-only consumer).
- **Persistence** via existing `ObjectService.UpdateObjectMeta` writing `window:displayname` — no backend schema change. Same pattern as workspace color, tab color, etc.

## Test plan

- [ ] Single click on a non-current row → focuses that window
- [ ] Double click on any row → enters edit mode with current name selected
- [ ] Type "My Window" + Enter → row displays "My Window"
- [ ] Restart app → rename persists
- [ ] Type "  " (spaces only) + Enter → row reverts (silent)
- [ ] Type 100 chars + Enter → truncates to 64
- [ ] Press Escape during edit → reverts to pre-edit name
- [ ] Click outside the input → commits
- [ ] Rename a workspace → unset windows in that workspace pick up the new workspace name; user-set names are preserved

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.436"
+version = "0.33.437"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.436"
+version = "0.33.437"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.436"
+version = "0.33.437"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.436"
+version = "0.33.437"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.437"
+version = "0.33.438"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.437"
+version = "0.33.438"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.437"
+version = "0.33.438"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.437"
+version = "0.33.438"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.438"
+version = "0.33.439"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.438"
+version = "0.33.439"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.438"
+version = "0.33.439"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.438"
+version = "0.33.439"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.435"
+version = "0.33.436"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.435"
+version = "0.33.436"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.435"
+version = "0.33.436"
 dependencies = [
  "windows-sys 0.59.0",
  "winres",
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.435"
+version = "0.33.436"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.435"
+version = "0.33.436"
 description = "AgentMux Host — Desktop app with bundled Chromium"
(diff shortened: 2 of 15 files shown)


a5af:
@codex review

AgentA-asaf:
## Round-1 review audit (commit \`12cc178f\`) → addressed in \`0b11d23c\`

### Reagent — CHANGES_REQUESTED → all 3 addressed
|---|---|---|---|
| **P1** | \`InstancePanel.tsx:211\` — first click of dblclick fires focus before rename | ✅ Fixed | Click now schedules \`handleFocusWindow\` via \`setTimeout(DBLCLICK_DELAY_MS=230ms)\`. Same-row second click cancels the timer (it's the dblclick path). Tracked per-row label so a click on row B while focus is pending for row A still treats B as a fresh click. |
| **P2** | \`app-init.ts:89\` — \`sortInstanceLabels\` dead code | ✅ Fixed | Removed; sort logic stays inlined. |
| **P2** | \`InstancePanel.tsx:62\` — unused \`activeInputRef\` | ✅ Fixed | Removed the variable; ref callback now does focus+select directly. |

### Codex — both addressed
|---|---|---|---|
| **P1** | \`InstancePanel.tsx:211\` — same dblclick suppression issue | ✅ Fixed | Same fix as reagent P1 (timer-deferred click). |
| **P2** | \`app-init.ts:127\` — change-detection only on label, ignores windowId arrival | ✅ Fixed | Refresh key is now \`label@windowId\` per entry, not just \`label\`. Paired with: \`register_backend_window\` now emits \`window-instances-changed\` so the listener actually fires when a sibling's windowId becomes available. Without that emit my key-change check would have nothing to trigger on. |

### Gemini — all 3 addressed (overlap with reagent / codex)
|---|---|---|---|
| HIGH | \`InstancePanel.tsx:212\` — dblclick suppression | ✅ Fixed | Same as P1. |
| MEDIUM | \`InstancePanel.tsx:62\` — unused activeInputRef | ✅ Fixed | Same as reagent P2. |
| MEDIUM | \`InstancePanel.tsx:221\` — div with role=button needs onKeyDown | ✅ Fixed | Added \`onKeyDown\` handling Enter / Space → focus. Bonus: F2 → enter rename mode (standard rename shortcut). |

---

\`@codex review\`
\`@gemini review\`

a5af:
@codex review
@gemini review

AgentA-asaf:
## Round-2 review audit (commit \`0b11d23c\`) → addressed in \`24bbad38\`

### Reagent
✅ APPROVED on \`0b11d23c\`. Re-evaluating \`24bbad38\`.

### Codex (2 / 2 real, both fixed)
|---|---|---|---|
| **P1** | \`InstancePanel.tsx:299\` — Commit rename before outside-click closes panel | ✅ Fixed | Added \`onCleanup\` that flushes any in-flight rename via fireAndForget when the panel unmounts. StatusBar's outside-click dismiss can unmount the input before its onBlur fires; without this the typed value would silently be lost. |
| **P2** | \`InstancePanel.tsx:70\` — Hardcoded \`DBLCLICK_DELAY_MS=230\` shorter than OS dblclick threshold | ✅ Fixed | New host RPC \`get_double_click_time\` returns Win32 \`GetDoubleClickTime()\` (default 500ms, user-configurable). Frontend queries on init, uses returned value + 50ms buffer. Defaults to 550ms while RPC in flight. Slow double-clickers are no longer locked out of dblclick-to-rename. |

### Gemini (5 inlines: 2 stale replays, 1 false positive on imports, 2 deferred nits)
|---|---|---|---|
| HIGH | \`InstancePanel.tsx:245\` — dblclick suppression | ⚠️ Stale replay | The timer-deferred fix from round-1 IS in this commit at \`InstancePanel.tsx:211\`. Gemini's suggested \`e.detail > 1\` is strictly inferior — it only blocks the redundant SECOND click; the first click still fires focus and steals the panel. Reagent and codex both agree the timer approach is correct (codex P1 from this round was about commit-on-cleanup, not dblclick suppression). |
| HIGH | \`InstancePanel.tsx:22\` — Missing imports for \`WaveWindow\`, \`Workspace\`, \`MetaType\` | ⚠️ False positive | All three are declared globally in \`frontend/types/gotypes.d.ts\` (auto-generated from Go): \`WaveWindow:1564\`, \`Workspace:1624\`, \`MetaType:830\`. \`tsc --noEmit\` passes with no errors on the modified files — verified twice. |
| MEDIUM | \`InstancePanel.tsx:267\` — onKeyDown for div role=button | ⚠️ Stale replay | Already added in round-1 commit \`0b11d23c\` (Enter / Space → focus, F2 → rename). Gemini didn't pick up the change. |
| MEDIUM | \`window.rs:379\` — Lock contention in \`list_window_instances\` | ⏸ Deferred | Window list is typically <10 entries; the lock is held for microseconds during JSON serialization. Optimizing for an N-window scenario where N matters would also need to address every other state.browsers consumer (e.g. CrossWindowDragMonitor) — not justified by the actual data. |
| MEDIUM | \`InstancePanel.tsx:133\` — 64-char vs UTF-8 byte cap | ⏸ Deferred | The backend's meta map has no documented byte cap on values; values are passed through as JSON strings and stored as-is. JS \`slice(0, 64)\` caps at 64 code units which is the user-visible character count — semantically the right unit for a \"display name length cap\". A byte cap would surprise users with emoji or non-ASCII names. |

---

\`@codex review\`
\`@gemini review\`

a5af:
@codex review
@gemini review

AgentA-asaf:
## Round-3 review audit (commit \`24bbad38\`) → addressed in \`bb029b36\`

### Reagent
✅ APPROVED on \`24bbad38\`. Re-evaluating \`bb029b36\`.

### Codex
|---|---|---|---|
| P1 | \`InstancePanel.tsx:328\` — Commit before outside-click closes panel | ⚠️ Stale (already fixed in \`24bbad38\`) | Codex is replaying the round-2 P1 against the same commit it was fixed in. \`onCleanup\` at \`InstancePanel.tsx:347\` flushes any in-flight rename via \`performRename\` (now via the extracted helper). Verified by reading the file. |

### Gemini — 4 of 11 are real cleanups (all addressed); the rest are stale replays or already-deferred nits

#### Real (✅ all fixed)
|---|---|---|---|
| MEDIUM | \`InstancePanel.tsx:77\` — pendingFocus timer not cleared on unmount | ✅ Fixed | \`onCleanup\` now calls \`cancelPendingFocus()\` first thing — covers the case where a focus is pending when the panel is dismissed. |
| MEDIUM | \`InstancePanel.tsx:138\` — redundant \`next\` variable | ✅ Fixed | Dropped; the \`editing\` capture at the top of \`commitRename\` already holds the same value. |
| MEDIUM | \`InstancePanel.tsx:176\` — duplicated commit logic | ✅ Fixed | Extracted \`performRename(windowId, name)\` helper used by both \`commitRename\` and the cleanup-flush. |
| MEDIUM | \`window.rs:373\` — \`#[allow(unreachable_code)]\` avoidable | ✅ Fixed | Replaced with mutually-exclusive \`#[cfg(not(target_os = \"windows\"))]\` block. Cleaner; no warning suppression needed. |

#### Deferred (with rationale on PR)
|---|---|---|---|
| MEDIUM | \`InstancePanel.tsx:146\` — error toast on rename failure | ⏸ Deferred | UpdateObjectMeta failures are extremely rare (local sidecar, no network). The optimistic-revert behavior is itself a UX signal that something went wrong. Adding a toast subsystem just for this case is over-engineering. |

#### Stale replays / false positives
| ID | File:Line | Status | Note |
|---|---|---|---|
| HIGH | \`InstancePanel.tsx:274\` — dblclick suppression | ⚠️ Replay (4th round) | Timer fix is in this commit at L211. Reagent + codex confirm. Gemini's suggested \`e.detail > 1\` is strictly inferior. |
| MEDIUM | \`InstancePanel.tsx:296\` — onKeyDown for div role=button | ⚠️ Replay | Added in round-1 commit \`0b11d23c\` and present in this commit. |
| HIGH | \`InstancePanel.tsx:22\` (×2) — Missing imports | ⚠️ False positive | Types declared globally in \`gotypes.d.ts:830, 1564, 1624\`. \`tsc --noEmit\` passes. |
| MEDIUM | \`window.rs:399\` — Lock contention | ⏸ Deferred (re-cited) | <10 windows; microseconds of lock time. Not worth the refactor. |
| MEDIUM | \`InstancePanel.tsx:144\` — 64-char vs byte cap | ⏸ Deferred (re-cited) | Chars are the correct unit for a display-name cap. Backend has no documented byte cap on meta values. |

---

\`@codex review\`
\`@gemini review\`

a5af:
@codex review
@gemini review

PR #570, opened 2026-04-28
Window/process state machine redesign + Phase A: launcher Job Object

a5af:
## Summary

- **Background research** — two analysis artifacts (`specs/ANALYSIS_*_2026_04_27.md`) that fed into the spec:
  - `ANALYSIS_WINDOW_PROCESS_STATE_INVENTORY_2026_04_27.md` — exhaustive map of every state-tracking site in the current code, with file:line citations, process diagram, lifecycle flows for the 5 main scenarios, gaps & contradictions, and a symptom→root-cause table.
  - `ANALYSIS_MULTI_PROCESS_BEST_PRACTICES_2026_04_27.md` — prior-art research (Chromium, Electron, VS Code, Tauri, Edge, Windows Job Objects), all cited so the spec's recommendations can be audited rather than taken on faith.
- **Spec** (`specs/SPEC_WINDOW_PROCESS_STATE_MACHINE_2026_04_27.md`) — full redesign of how AgentMux tracks windows, tabs, instances, and child processes. Diagnoses the recurring bug class (zombie processes after close-all, "taskbar but no window" on relaunch, pool windows leaking to taskbar, instance-count inflation, tab-state divergence) as a single root cause: state spread across 13+ stores in 4 processes with no enforced invariants. Proposes a Redux-style single source of truth in the launcher, a 6-state window FSM, a separate type for warm-pool windows (Chromium spare-renderer pattern), Job Object + per-data-dir named mutex for OS-level cleanup, and CQRS-lite IPC. 4-phase incremental migration plan.
- **Phase A first piece** — moved the existing Job Object code from `agentmux-cef/src/sidecar.rs` (where it wraps only the srv) into the launcher, where it now wraps the host. CEF render workers + srv inherit the job. Killing the launcher (Task Manager) now cascades to the entire process tree. Existing host→srv job left as defense-in-depth. Bump to `0.33.436`.

## Why this Phase A first

The host's existing srv-wrap only fires when the host exits cleanly — which doesn't happen in the current zombie scenario, where pool windows keep CEF's message loop alive and the host never returns from `run_message_loop`. With the job one level up at the launcher, recovery becomes a single kill of the launcher → entire tree dies.

## What's NOT in this PR (deferred follow-ups inside Phase A)

- host → launcher heartbeat with force-cleanup on timeout (the bit that handles "launcher is also blocked in `child.wait()`")
- synchronous `Quit` RPC to srv with ack + 3s timeout
- `LifecyclePhase` enum + minimal reducer
- Named mutex for single-instance: per `CLAUDE.md`, AgentMux is designed to run multiple instances in parallel, so the existing **per-data-dir** port file scheme is correct in scope. The spec adapts the Electron pattern to a per-data-dir mutex name. Open question for the spec discussion.

## Verification

Launcher log on this branch:
```
spawned CEF host pid=43336
Job Object assigned to pid=43336, KILL_ON_JOB_CLOSE active
```
On host panic in test, **process list completely empty afterward** — srv reaped via the new launcher-owned job. Compare v0.33.435 on the same machine after same panic → 7 processes remain alive.

End-to-end UI test was blocked on this dev box by an unrelated CEF init failure (panic at `agentmux-cef/src/main.rs:346`) reproducing on **all** versions installed (0.33.429, 0.33.435, 0.33.436) — `libcef.dll` is byte-identical between them and no agentmux processes / port-9222 listeners were running, so it looks like a stale Chromium kernel object after a recent 12-process force-kill. A reboot would clear it; not in scope for this PR.

## Reviewing order suggestion

1. `ANALYSIS_WINDOW_PROCESS_STATE_INVENTORY_2026_04_27.md` — what exists today, what's broken, where
2. `ANALYSIS_MULTI_PROCESS_BEST_PRACTICES_2026_04_27.md` — what mature codebases do
3. `SPEC_WINDOW_PROCESS_STATE_MACHINE_2026_04_27.md` — the proposed redesign + 4-phase migration
4. The launcher diff in `agentmux-launcher/` — Phase A's first piece

## Test plan

- [ ] Reviewer reads the two analyses, then the spec (especially §1 symptom→root-cause and §9 migration)
- [ ] Run a portable build, open AgentMux, confirm UI works as before (no functional change expected)
- [ ] Tail `~/.agentmux/logs/agentmux-launcher.log`, confirm the new `Job Object assigned ... KILL_ON_JOB_CLOSE active` line appears at startup
- [ ] Open multiple windows + warm-pool tear-offs, then kill the launcher process via Task Manager — confirm host, render workers, and srv all exit
- [ ] Confirm running multiple AgentMux instances (different versions, dev + portable) still works — each gets its own job, no interference


a5af:
@codex review

AgentA-asaf:
## Round-1 review audit + spec edits → addressed in \`104c65ca\` (force-push, branch rebased onto current main)

This PR was opened from a base before #569 merged; rebased onto current main (\`c5f60944\`). Bumped version to \`0.33.440\` to stay ahead. Force-pushed.

### Reagent
✅ APPROVED on \`c172f7e4\`. Re-evaluating \`104c65ca\`.

### Codex (1 / 1 real, fixed)
|---|---|---|---|
| **P2** | \`agentmux-launcher/src/main.rs:62\` — Race: host can spawn children before AssignProcessToJobObject | ✅ Fixed | Replaced \`Command::spawn()\` with \`Command::creation_flags(CREATE_SUSPENDED).spawn()\`. AssignProcessToJobObject now runs while the host is frozen at process-init — no child can spawn in the gap. New helper \`resume_main_thread(pid)\` walks a Toolhelp32 thread snapshot to find and resume the host's main thread after job assignment. Standard Microsoft / Raymond Chen pattern; same approach Chrome uses for sandbox-less child cleanup. |

### Gemini (4 / 4 addressed)
|---|---|---|---|
| HIGH | \`launcher/main.rs:62\` — same race | ✅ Fixed | Same as codex P2. |
| MEDIUM | \`launcher/main.rs:105\` — inaccurate KILL_ON_JOB_CLOSE comment | ✅ Fixed | Comment rewritten. The drop-job-on-host-exit step is the backstop for orphan children, not a no-op. |
| MEDIUM | \`launcher/main.rs:160\` — JobHandle should use HANDLE typed alias | ✅ Fixed | \`JobHandle(*mut c_void)\` → \`JobHandle(windows_sys::Foundation::HANDLE)\`. (Note: gemini's claim that HANDLE is \`isize\` is incorrect — verified it's \`*mut c_void\` in \`windows-sys-0.59.0/src/Windows/Win32/Foundation/mod.rs:10232\` — but the typed alias is still cleaner and matches CreateJobObjectW's return type.) |
| MEDIUM | \`launcher/main.rs:194\` — Manual FFI for CreateJobObjectW unnecessary | ✅ Fixed | Verified gemini IS right: \`CreateJobObjectW\` is exported by windows-sys 0.59 at \`Win32/System/JobObjects/mod.rs:5\`, just gated behind the \`Win32_Security\` feature (its first parameter is \`SECURITY_ATTRIBUTES*\`). Added the feature; dropped the \`extern \"system\"\` block. |

## Spec edits in this push (commit \`9946b9f9\`)

The original spec proposed a host→launcher heartbeat as part of Phase A. After review with the maintainer, **heartbeats are out — period**. The whole point of the reducer + versioned events + GetSnapshot resync architecture is that state can't drift, so polling becomes unnecessary. Same principle that drove \`SPEC_BACKEND_LIFECYCLE.md\` to reject WS-watchdog polling.

Spec text updated in §3.2, §4.3, and §9 Phase A:
- §3.2 process diagram: srv reports started/ready as facts; no heartbeat. Death detected via pipe-EOF and child.wait() exit code.
- §4.3 new \"On heartbeats\" subsection: explains why the architecture deliberately rejects them; cites PR #568's existing fix for the message-loop-hang case (\`client.rs::on_before_close\` excludes pool windows from user-browser count).
- §9 Phase A: heartbeat bullet removed; explicit \"NOT in Phase A: any periodic heartbeat or polling timer\".

Phase A is now **2 PRs, not 5**:
- This PR (#570) — Job Object on launcher + race fix
- A follow-up — per-data-dir named mutex single-instance + handoff (replaces port file)

\`LifecyclePhase\` enum, reducer, and synchronous \`Quit\` RPC moved to **Phase B** (the central architectural inversion; a half-built Phase A version would be re-architected when Phase B arrives).

## Force-push note

Branch was rebased to resolve the merge conflict from PR #569 landing. Three new commits:
- \`9946b9f9\` spec: drop heartbeats
- \`fb948b77\` chore: remove session-local retro notes (these were accidentally swept into the rebase)
- \`2899f132\` / \`14705923\` version bump to 0.33.440 + lockfile sync
- \`104c65ca\` the race fix + cleanups described above

---

\`@codex review\`
\`@gemini review\`

a5af:
@codex review
@gemini review

AgentA-asaf:
## Round-2 review audit (commit \`104c65ca\`) → addressed in \`e1630067\`

All three bots flagged the same defensive issue — addressed.

### Reagent — CHANGES_REQUESTED → fixed
|---|---|---|---|
| **P1** | \`launcher/main.rs:125\` — Suspended-host zombie if Job Object creation failed AND resume_main_thread fails | ✅ Fixed | Added \`let _ = child.kill();\` before \`drop(job)\` in the resume-failure path. When the job is None (creation failed earlier with WARN), drop is a no-op so KILL_ON_JOB_CLOSE doesn't cover it; the explicit kill is the backstop. |

### Codex — fixed
|---|---|---|---|
| P2 | \`launcher/main.rs:126\` — Same | ✅ Fixed | Same fix. |

### Gemini — both fixed
|---|---|---|---|
| MEDIUM | \`launcher/main.rs:127\` — Same | ✅ Fixed | Same fix. |
| MEDIUM | \`launcher/main.rs:148\` — Same defensive concern in the child.wait() error path | ✅ Fixed | Added the same backstop \`let _ = child.kill();\` before \`drop(job)\` in the wait-failure path. |

---

\`@codex review\`
\`@gemini review\`

a5af:
@codex review
@gemini review

PR #571, opened 2026-04-28
Phase B.1: launcher spawns srv as sibling of host

AgentA-asaf:
## Summary

First substantive PR in Phase B (the redux-architecture migration per [SPEC_WINDOW_PROCESS_STATE_MACHINE_2026_04_27.md](specs/SPEC_WINDOW_PROCESS_STATE_MACHINE_2026_04_27.md)). Restructures the process tree so srv is a sibling of host under the launcher, not a child of host. Foundation for the launcher-owned reducer (B.2+).

**Process tree after this PR:**
\`\`\`
launcher (Job Object J0, KILL_ON_JOB_CLOSE)
├── srv     (assigned to J0; survives host crash)
└── host    (assigned to J0; CEF render workers inherit J0)
\`\`\`

vs today (post-PR-#570):
\`\`\`
launcher (J0)
└── host
    ├── srv     (assigned to host-owned J1; DIES when host dies)
    └── renderers
\`\`\`

## Why srv-survives-host-crash matters

Phase B's launcher-owned state machine needs to be able to restart the host while keeping srv (and its DB state) alive. With today's tree, killing the host also kills srv, losing all workspace/tab state. After this PR, srv outlives any host crash.

## Key change: deleted host's J1 on srv

Removed the host-owned Job Object that wrapped srv. Spec originally suggested keeping it as defense-in-depth, but on inspection it would **actively defeat B.1's goal**: KILL_ON_JOB_CLOSE on J1 fires when host dies, taking srv with it. The two reapers are mutually exclusive given the goal "srv survives host crash" — picked launcher's J0.

Trade-off: in \`task dev\` mode (host runs without launcher), srv loses kernel-level reaping on host crash. Acceptable for dev-mode-only; production (portable / installed) is unaffected.

## Env-var contract launcher → host

Launcher passes srv coordinates via env vars; host honors them and skips its own \`spawn_backend\`. Falls back to the existing \`spawn_backend\` only when env is absent (the dev path).

| Env var | Source | Consumer | Meaning |
|---|---|---|---|
| \`AGENTMUX_BACKEND_WS\` | launcher | host | srv WebSocket endpoint |
| \`AGENTMUX_BACKEND_WEB\` | launcher | host | srv HTTP endpoint |
| \`AGENTMUX_BACKEND_PID\` | launcher | host | srv PID for diagnostics |
| \`AGENTMUX_AUTH_KEY\` | launcher | host (and srv) | shared auth key (UUID v4) |
[...]

PR #572, opened 2026-04-28
hotfix(launcher): keep srv stdin alive past tokio wait() (post-B.1)

AgentA-asaf:
## Summary

Hotfix for a bug introduced by PR #571 (B.1) that the smoke test caught immediately: launching the portable failed within milliseconds with srv shutting down voluntarily.

## Symptom

Launcher log on first launch of v0.33.447 portable:
\`\`\`
spawned CEF host pid=4572 (suspended)
Job Object assigned to host pid=4572, KILL_ON_JOB_CLOSE active
entering host + srv concurrent wait
[srv 7836 stderr] stdin closed, shutting down
[srv 7836 stderr] shutdown signal received, exiting
srv exited UNEXPECTEDLY (host still running) with code 0 — terminating launcher
launcher exiting with code 1
\`\`\`

User-visible: AgentMux fails to launch. UI never appears.

## Root cause

\`tokio::process::Child::wait()\` proactively drops \`self.stdin\` before waiting. From [tokio source](https://github.com/tokio-rs/tokio/blob/master/tokio/src/process/mod.rs):

\`\`\`rust
// "Ensure stdin is closed so the child can't read from it any more."
drop(self.stdin.take());
\`\`\`

agentmux-srv has a parent-watch loop on its own stdin (\`agentmux-srv/src/main.rs:526-548\`): when stdin reads 0 bytes (EOF from a closed write end), it interprets that as "parent died" and shuts down as defense-in-depth against orphans.

The host's existing \`spawn_backend\` uses \`std::process::Command\` (stdin stays open as long as Child is alive in app_state). Switching to \`tokio::process::Command\` for the launcher introduced this divergence: tokio's \`wait()\` deliberately drops stdin → srv sees EOF within milliseconds → srv exits → launcher's \`tokio::select\` fires the \`srv-exited-unexpectedly\` branch → launcher kills the host and exits.

## Fix

Take \`srv_child.stdin\` out into a launcher-scope binding BEFORE the \`tokio::select\` runs. With \`srv_child.stdin = None\`, tokio's \`wait()\` finds nothing to drop, and the launcher holds the \`ChildStdin\` alive for the entire lifetime of \`run_windows\`. srv's stdin pipe stays open as long as the launcher is running — exactly the lifetime contract srv expects.

\`\`\`rust
// New, before tokio::select:
let _srv_stdin_keepalive = srv_child.stdin.take();
\`\`\`

One-line addition; the fix is the binding's lifetime, not its content.

## Verification

Re-running v0.33.448 portable: launcher log no longer shows the \"srv exited UNEXPECTEDLY\" line; host + srv both stay running; UI loads as expected.

## Why not in PR #571

Caught by manual smoke testing immediately after merge. PR #571's review focused on Job Object semantics + race conditions, not tokio's stdio defaults — the bots had no way to know srv had a parent-watch on stdin without reading agentmux-srv. Smoke testing is the right backstop for this class.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.447"
+version = "0.33.448"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.447"
+version = "0.33.448"
 dependencies = [
  "tokio",
  "tracing",
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.447"
+version = "0.33.448"
 dependencies = [
  "chrono",
  "dirs",
@@ -57,7 +57,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.447"
+version = "0.33.448"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.446"
+version = "0.33.447"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

AgentA-asaf:
## Round-1 review audit (commit \`1b6b767\`) → addressed in \`09de054c\`

### Reagent — CHANGES_REQUESTED, single P1, fixed
|---|---|---|---|
| **P1** | \`package-lock.json:3\` — Lockfile at 0.33.447 but package.json at 0.33.448 | ✅ Fixed | Ran \`npm install --package-lock-only\` and committed. The 0.33.447 → 0.33.448 bump in this hotfix branch missed the lockfile sync; now caught up. |

### Codex — same P2, fixed
|---|---|---|---|
| **P2** | \`package-lock.json:3\` — Same version-mismatch issue, independently flagged | ✅ Fixed | Same single fix. Both bots correctly noted that \`scripts/bump-wrapper.sh\` enforces equality and exits non-zero on mismatch — would have broken automation. |

---

\`@codex review\`
\`@gemini review\`

a5af:
@codex review
@gemini review

PR #581, opened 2026-04-28
feat: B.5c — instance registry read path cuts over to shadow

AgentA-asaf:
## Summary

Step 3 of the B.5 migration. Host's instance-number READS now route through `state.instance_num()` / `state.instance_count()`, which prefer the launcher-authoritative `shadow_instance_registry` (B.5b) and fall back to host's local `WindowInstanceRegistry` only for the race window where host has registered locally but the launcher's `WindowInstanceAssigned` event hasn't returned yet.

Mutations (`register()` / `unregister()`) are intentionally kept — removing them would race the synchronous frontend lookup on new-window load. B.5d retires host's mutations once we've validated the shadow tracks reality via the new `launcher-ipc:fallback` log target.

## What's in here

**`AppState`** (`state.rs`):
- `instance_num(label)` — shadow → host fallback → None. DEBUG logs to `launcher-ipc:fallback` on miss.
- `instance_count()` — returns shadow length; DEBUG-logs when shadow vs host disagree.

**Read sites switched** (5 total):
- `commands/window.rs::get_instance_number`
- `commands/window.rs::get_window_count`
- `commands/window.rs::register_backend_window` (count emit)
- `client.rs::on_before_close` (remaining-count emit)
- `commands/drag.rs` (tear-off count emit)
- `commands/window_pool.rs` (pool-promote count emit)

## What's NOT in here

- Mutation removal — B.5d will drop host's local register/unregister calls once fallback DEBUG logs confirm the shadow covers all reads in real workloads.

## Test plan

- [x] Smoke-tested v0.33.459 (B.5b head): two cold-path tear-offs assigned 2 + 3, monotonic, zero `DRIFT`. B.5c is a read-path-only change so behavior should be identical, with new `launcher-ipc:fallback` DEBUG events visible during the race window.
- [ ] CI workspace check + tests on the merge ref.
- [ ] Smoke test the new build: open + tear off, confirm `get_instance_number` returns the same numbers as before, confirm `launcher-ipc:fallback` shows up only briefly during window-creation races (not at steady state).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.459"
+version = "0.33.460"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.459"
+version = "0.33.460"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.459"
+version = "0.33.460"
 dependencies = [
  "agentmux-common",
  "chrono",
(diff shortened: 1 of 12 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #583, opened 2026-04-28
feat: B.5d — drop host's instance-registry mutations

AgentA-asaf:
## Summary

Step 4 of the B.5 migration for `window_instance_registry`. After B.5c routed reads through `state.instance_num()` / `state.instance_count()`, the host's local `register()`/`unregister()` calls are the only remaining writers. This PR drops them. The launcher's `state.instance_registry` (B.5a) is now the sole authority; host's `WindowInstanceRegistry` becomes seed-only (`{\"main\": 1}`, never mutated) and will be deleted entirely in B.5e.

## What's in here

**Mutation sites removed** (4):
- `client.rs::on_before_close` — `unregister(lbl)`
- `commands/drag.rs::tear_off` — cold-path `register(&label)`
- `commands/window.rs::open_new_window` — `register(&label)` before CEF create
- `commands/window_pool.rs::promote_pool_window` — pool-promote `register(&label)`

**Read-path adjustments**:
- `state.instance_count()`: removed the host-count fallback. Host's count is permanently `1` (seed only) post-B.5d, so it would give stale answers. Returns shadow count directly.
- `state.instance_num(label)`: kept the host fallback for now — harmless (only \"main\" → 1 is in seed range, which is correct). B.5e removes with the field.

**Race coverage** (this is the design-question mentioned in the roadmap):
- Synchronous `window-instances-changed` emit at mutation sites fires BEFORE the launcher's reply, so it briefly carries an off-by-one count.
- `apply_event_to_shadow` now re-emits `window-instances-changed` on every `WindowInstanceAssigned` / `WindowInstanceReleased`, catching up InstancePanel within ~ms.
- Frontend `get_instance_number` may race the launcher reply on first window load, returning `unwrap_or(1)` briefly. Existing `app-init.ts::refreshLabels(retriesLeft)` retry loop covers it. B.7 replaces this with direct event subscription.

## What's NOT in here

- B.5e: delete `WindowInstanceRegistry` field + struct entirely. Almost-pure delete after this PR validates.

## Test plan

- [x] `cargo check -p agentmux-launcher` passes.
- [ ] CI workspace check.
- [ ] Smoke test: open + close 2-3 tear-offs. Confirm:
  - InstancePanel count updates correctly (synchronous emit + shadow-update emit cadence).
  - No `launcher-ipc:drift` warnings.
  - `launcher-ipc:fallback` DEBUG entries should now mostly disappear (host is seed-only).
  - Window title flicker (briefly showing \"main\" name on a tear-off) for ~10-100ms is acceptable transitional behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.463"
+version = "0.33.464"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.463"
+version = "0.33.464"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.463"
+version = "0.33.464"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.463"
+version = "0.33.464"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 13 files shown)


a5af:
@codex review

a5af:
@codex review

PR #588, opened 2026-04-28
feat: B.5 (window_id_map) step d — drop host's window_id_map mutations

AgentA-asaf:
## Summary

Step d of the window_id_map migration. Host's two `window_id_map` mutation sites are removed:

- `commands/window.rs::register_backend_window` — local `insert` is gone; only the launcher report fires.
- `client.rs::on_before_close` — local `remove` is gone; wid looked up via `state.backend_window_id(lbl)` before the launcher report.

The wid lookup at close is safe without the host fallback because the frontend's original `register_backend_window` ran long before close — shadow has been populated for the entire window lifetime.

After this lands, host's `window_id_map` is empty (initialized empty, no mutations) and vestigial. Step e deletes it.

## What's NOT in here

- Step e: delete `AppState.window_id_map`, host-fallback branch in `backend_window_id`, drift compares in `apply_event_to_shadow`. Almost-pure-delete.

## Test plan

- [x] `cargo check -p agentmux-launcher` passes.
- [ ] CI workspace check.
- [ ] Smoke test: open + close a tear-off, confirm `register_backend_window` IPC still works (frontend sees no failure), close path correctly notifies backend with the right wid.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.468"
+version = "0.33.469"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #594, opened 2026-04-28
docs(b5-finish): scaffolding-role comments on host residual fields

AgentA-asaf:
## Summary

B.5 finish PR — comment-only. Closes out the per-map migration phase.

After 3 successful migrations (window_instance_registry, window_id_map, window_meta), 2 host fields remain (`browsers` + pool maps) that intentionally **don't** follow the standard a→e ratchet. This PR adds explicit doc comments on each surviving field declaring its role, so future readers don't try to migrate them via the same pattern.

## Three classes (per analysis)

- **FFI scaffolding** (`browsers`): `cef::Browser` handles can't serialize across IPC. KEY-SET role already covered by launcher's `state.windows`.
- **Sync lifecycle scaffolding** (`window_pool`, `unpromoted_pool_labels`, `window_pool_respawn_in_flight`): pool decisions need sync-local state. Launcher's `state.pool` mirrors for cross-process queries.
- **Sync cache** (`window_meta`): mirrors launcher's `state.windows`; written from a single canonical site (on_after_created / on_before_close); needed for synchronous parent-liveness + cascade-close.

All three retire into a host-side reducer in Phase F (sequenced after Phase E srv reducer).

## What's next after this merges

- **B.6** — single-instance mutex (named pipe already covers it, just delete old port-file probe + add clear error)
- **B.7** — frontend cutover (replace polling with event subscription)
- **B.8** — Phase B exit (property tests, --diag, CI smoke)

## Test plan

- [x] `cargo check -p agentmux-launcher` passes.
- [ ] CI workspace check.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #595, opened 2026-04-28
B.6: launcher single-instance mutex via named-pipe bind

AgentA-asaf:
## Summary

Phase B.6 — turn the launcher's existing `first_pipe_instance(true)` bind into the single-instance enforcement signal, with a user-visible error when a second launcher hits the same data dir. Retire the legacy `<data-dir>/ipc-port` probe + write in the host.

Per `docs/retro/phase-b-roadmap.md` (B.6 row) and the gap-8 stale-state defect documented in `specs/ANALYSIS_WINDOW_PROCESS_STATE_INVENTORY_2026_04_27.md`.

## What changed

- `agentmux-launcher/src/ipc/server.rs`: split the synchronous bind out of the spawn loop into `bind_first_pipe_instance(pipe_name)`. `run_ipc_server` now takes a pre-bound `NamedPipeServer`.
- `agentmux-launcher/src/main.rs`: binds the pipe before spawning srv/host. On `ERROR_ACCESS_DENIED` (second launcher), shows a `MessageBoxW` ("AgentMux is already running for this data directory…") and exits 2. Other bind errors get a distinct message but the same exit code. New helper `show_fatal_dialog` for the modal.
- `agentmux-launcher/Cargo.toml`: enable `Win32_UI_WindowsAndMessaging` feature for `MessageBoxW`.
- `agentmux-cef/src/main.rs`: remove the legacy `ipc-port` probe (lines 188–213), the post-CEF-init port-file write (lines 393–398), and the shutdown cleanup. Best-effort `remove_file` of any leftover from a pre-B.6 install. The `open_new_window` IPC command itself stays — only the cross-process discovery via the port file goes away. The launcher's pipe is OS-owned; on launcher death the pipe is reaped, so there is no stale-state path.
- `docs/retro/phase-b-roadmap.md`: tick B.5 finish (PR #594), mark B.6 in flight.

## Why exit code 2

Distinguishes single-instance refusal from generic launch failure (exit 1). Lets shell wrappers / CI easily tell "user double-clicked while running" from "config broken."

## Test plan

- [ ] `task package` builds clean.
- [ ] Launch portable AgentMux, then double-click the same `.exe` again — expect a MessageBox, not a silent exit. Check exit code is 2.
- [ ] Launch two portable instances at DIFFERENT versions (different data dirs) — both should run side-by-side (the pipe name is namespaced per `data_dir_hash16`).
- [ ] `task dev` still works (host runs without launcher in dev mode).
- [ ] Hard-kill (taskkill /F) the launcher — next launch should succeed (no stale state, OS reaped the pipe).

## Follow-up

After this lands, **B.5 + B.6 are complete**; **B.7** is the frontend cutover from polling to events (delete `app-init.ts::refreshLabels(retriesLeft)` polling).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.474"
+version = "0.33.475"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.474"
+version = "0.33.475"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.474"
+version = "0.33.475"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.474"
+version = "0.33.475"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

PR #596, opened 2026-04-28
B.7.1: enrich window-instances-changed payload with entries

AgentA-asaf:
## Summary

Smallest first slice of Phase B.7 (frontend cutover from polling to events). The launcher-driven `window-instances-changed` re-emit now ships a resolved `{ count, entries: [{label, windowId}] }` payload, letting the frontend update its atoms directly without an RPC round-trip and without the historical retry-on-stale polling.

## Why now

- **B.5 + B.6 are merged.** `state.windows` (canonical) and `shadow_instance_registry` / `shadow_window_meta` / `shadow_backend_window_ids` (host projections) are populated before the launcher-driven re-emit fires — so a single read at that moment is authoritative.
- The historic 6×50ms polling in `refreshLabels` was a workaround for the race between the **synchronous** `window-instances-changed` emit (firing before `state.browsers` had settled) and the actual registration. The launcher-driven re-emit fires post-settlement, so for that path the polling is pure latency.

## Scope

**Host (`agentmux-cef/src/launcher_ipc.rs`):**
- New private helper `emit_window_instances_changed_with_entries(state)`. Builds entries from `shadow_instance_registry` keys (filtering pool + `browser-pane-*` labels), joins with `shadow_backend_window_ids` per label, emits `{ count, entries }`.
- Replaced the count-only emit in two arms of `apply_event_to_shadow` (`WindowInstanceAssigned`, `WindowInstanceReleased`).
- **Out of scope (B.7.2):** the synchronous count-only emits in `window.rs` / `drag.rs` / `window_pool.rs` / `client.rs`. Those still fire; the frontend keeps the legacy fallback for them.

**Frontend (`frontend/app-init.ts`):**
- Split `refreshLabels` into a stateless `applyEntries(entries)` (atoms-only) and `refreshLabelsViaRpc()` (legacy fallback).
- The `window-instances-changed` listener now branches on payload shape: structured `{ count, entries }` → `applyEntries(payload.entries)` directly; legacy number → `refreshLabelsViaRpc()`. The retry / `lastEntriesKey` polling state is gone for the structured path.

## Why two paths

Sync emits (window.rs:655, drag.rs:385, etc.) still ship count-only. Dev mode without launcher relies on those exclusively (no launcher events → no re-emit). The legacy fallback keeps the panel reactive in both cases until B.7.2 retires the sync emits.

## Test plan

- [ ] `task package` builds clean (cef-dll-sys cleared the local Defender race after retry).
- [ ] Launch portable, open a second window — InstancePanel updates within one event tick (no listWindowInstances RPC visible in host log).
- [ ] Close the second window — InstancePanel updates symmetrically.
- [ ] Tear-off → InstancePanel reflects the new window.
- [ ] `task dev` (no launcher in the loop) still updates the InstancePanel via the count-only sync-emit fallback.

## Local cargo check

`agentmux-launcher` checked clean. `agentmux-cef` cargo-check hit the Defender / cef-dll-sys download race three times locally (`os error 5` during fs::rename — see `project_cef_build_race.md`). The diff adds a private helper + two call-site swaps in an already-imported module; type-trivial. CI / portable build will confirm.

## Follow-up

- **B.7.2** — retire the synchronous count-only emits, simplify the frontend listener to only the structured path.
- **B.7.3** — consider direct launcher-event subscription via the CEF JS bridge (per the original B.7 design) so the host's bespoke `window-instances-changed` event can be retired in favor of typed launcher events.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.475"
+version = "0.33.476"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.475"
+version = "0.33.476"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.475"
+version = "0.33.476"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.475"
+version = "0.33.476"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.475"
+version = "0.33.476"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #597, opened 2026-04-28
B.7.2: re-emit window-instances-changed on BackendWindowId events

AgentA-asaf:
## Summary

Follow-up to #596 (B.7.1). Adds the entries-bearing `window-instances-changed` re-emit to two more arms of `apply_event_to_shadow` so the windowId-resolution race no longer falls through to the legacy RPC retry path.

## Background

A freshly-opened window arrives in entries with `windowId=null` because the launcher's `WindowInstanceAssigned` fires before the frontend's `register_backend_window` round-trip completes. Pre-B.7.1, the `lastEntriesKey` retry polled until the `null → real` transition was observed. B.7.1 retired the polling for the structured-payload path — but only the `WindowInstanceAssigned` and `WindowInstanceReleased` arms re-emitted with entries. The `null → real` transition (which fires on the launcher's `BackendWindowIdRegistered` event, originating from `report_backend_window_id_registered` after the frontend round-trip) was still served by the host's count-only sync emit at `window.rs:511`, falling through to `refreshLabelsViaRpc(true)` + retry.

## Change

- `agentmux-cef/src/launcher_ipc.rs::apply_event_to_shadow`:
  - `BackendWindowIdRegistered` arm: after updating `shadow_backend_window_ids`, call `emit_window_instances_changed_with_entries(state)`.
  - `BackendWindowIdUnregistered` arm: same, symmetric for the reverse direction.
- Roadmap: tick B.7.1 (#596), mark B.7.2 in flight.

No frontend changes. No protocol changes. Net ~+25 LoC.

## Effect

After this lands, the launcher-driven path is the sole source of truth for InstancePanel updates (open / close / windowId-resolved / windowId-nulled). The frontend's RPC + retry fallback only fires for sync emits — primarily `task dev` mode (no launcher).

## Test plan

- [ ] `task package` builds clean.
- [ ] Launch portable, open a second window — InstancePanel resolves the windowId without visible delay (no \"Window 1\" placeholder period before the real name).
- [ ] `task dev` (no launcher) — the legacy fallback path still updates the panel via RPC retry.

## Local cargo check

agentmux-cef cargo check hit the documented Defender / cef-dll-sys download race locally (`project_cef_build_race.md`). The diff is two small additions in an already-imported function; type-trivial.

## Follow-up

- **B.7.3** — wire the host's CEF JS bridge to forward typed launcher events directly to the renderer (the original B.7 design). Once that lands, the bespoke `window-instances-changed` event can be retired in favor of typed launcher events, and the count-only sync emits can go.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.476"
+version = "0.33.477"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.476"
+version = "0.33.477"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.476"
+version = "0.33.477"
 dependencies = [
  "agentmux-common",
  "chrono",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #598, opened 2026-04-28
B.6.1 fix: forward open_new_window on second launch (don't MessageBox)

AgentA-asaf:
## Summary

Fix the user-facing regression #595 introduced. Double-clicking \`agentmux.exe\` while an instance is already running should open another window in the existing instance (matching the status-bar version popup's "new window" UX), not pop a Win32 dialog.

## Root cause

#595 retired the legacy \`<data-dir>/ipc-port\` probe + write entirely. The probe was the WHOLE forwarding mechanism, not just a stale-state defect. Without it, the launcher's named-pipe rejection had nowhere to forward to and could only show "already running" or exit silently.

## Fix

The named-pipe bind in the launcher REMAINS the authoritative single-instance lock — that part of B.6 was right. The fix narrows the change to forwarding only:

- **agentmux-cef/src/main.rs**: re-publish \`<data-dir>/ipc-port\` (port:token) AFTER CEF init; remove on graceful exit. **The probe sites are NOT restored** — only the publishing side. Reading the port + forwarding moves to the launcher.
- **agentmux-launcher/src/main.rs**: on \`ERROR_ACCESS_DENIED\` from \`bind_first_pipe_instance\`, read \`<data-dir>/ipc-port\`, send a raw HTTP/1.1 POST to \`http://127.0.0.1:<port>/ipc\` with body \`{"cmd":"open_new_window"}\` and bearer auth, exit 0. No reqwest dep — \`std::net::TcpStream\` keeps the launcher binary small (~325 KB). Forwarding failures exit 0 silently to avoid surprise dialogs on transient startup races.
- **Genuine bind failures** (not "already running") still surface the MessageBox + exit 2 path. That's the only place the dialog fires now.

## Stale-state defect bounding

The gap-#8 defect (port file outliving a hard crash) is now bounded by the pipe-bind ordering:

- **First instance launch**: pipe-bind succeeds → continue normal startup → port file is overwritten. Stale file irrelevant.
- **Second instance launch**: pipe-bind fails with ERROR_ACCESS_DENIED → first instance is genuinely alive (it holds the pipe handle) → it wrote a fresh port:token → forwarding lands.
- **Crashed first instance, no live instance**: OS reaps the pipe handle → next launch's pipe-bind succeeds → first-instance path → port file overwritten.

The only failure mode is a *forwarding-time* race (HTTP server mid-startup). The silent \`exit 0\` covers that — user can double-click again.

## Test plan

- [ ] Launch portable, double-click again → second window opens in the existing instance, no dialog. Verify with \`tasklist | grep agentmux\` that only one launcher / host pair runs.
- [ ] \`taskkill /F\` the launcher, then re-launch → starts fresh (pipe was reaped).
- [ ] Different version portable side-by-side → both run independently (different data dirs, different pipes).
- [ ] Genuine bind error path (manual: simulate by holding a pipe handle elsewhere) → MessageBox surfaces.

## Roadmap

\`docs/retro/phase-b-roadmap.md\` updated with B.6.1 entry under the B.6 section.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.478"
+version = "0.33.479"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.478"
+version = "0.33.479"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.478"
+version = "0.33.479"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.478"
+version = "0.33.479"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

PR #599, opened 2026-04-29
fix(B.6.1): read response after forward POST

AgentA-asaf:
## Symptom

After #598 + 2ffa63c5 + this branch's parent, v0.33.481 portable launches fine and the launcher correctly forwards on second-launch:

\`\`\`
[v0.33.481] pipe bind failed (already_running=true): Access is denied.
[v0.33.481] forwarded open_new_window to existing instance — exiting 0
\`\`\`

But **no second window appears.** Manually running curl with the same port + token + payload returns \`200 OK\` with a real \`window-...\` label, so the host handler is correct.

## Root cause

\`forward_open_new_window\` writes the HTTP request and returns. The launcher process exits, the kernel closes the TCP socket (RST). The host's axum handler is async — between the bytes arriving and the handler being dispatched to \`open_new_window\`, the connection is already gone. axum drops the in-flight request silently (no log line on the host either).

## Fix

After \`write_all\`, do a 2s-timeout read of up to 64 bytes from the socket. We don't care about the response — the read just keeps the connection alive past axum's handler dispatch. \`Connection: close\` lets the server drop the socket once the response is written, so the read returns naturally without blocking the launcher's exit.

## Test plan

- [ ] Launch v0.33.482 portable, double-click \`agentmux.exe\` → second window pops, no dialog. Repeat 2-3 times rapidly to confirm each click yields one new window.
- [ ] Launcher log shows \`forwarded open_new_window to existing instance — exiting 0\` (unchanged).
- [ ] Host log now shows the corresponding \`open_new_window\` invocation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.480"
+version = "0.33.481"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.480"
+version = "0.33.481"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.480"
+version = "0.33.481"
 dependencies = [
  "agentmux-common",
  "chrono",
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

AgentA-asaf:
Smoke verified locally on v0.33.482 (head 2d7a7dc6):

```
[launcher] pipe bind failed (already_running=true): Access is denied. — pipe=\.\pipe\agentmux-e92ac66dc0b014d6\command
[launcher] forwarded open_new_window to existing instance — exiting 0
[host] [window] register_backend_window received label=window-a4bc803a62024e8698994d4365c1f70c
```

Second `agentmux.exe` launch yielded one new window in the existing instance, no dialog. Earlier v0.33.481 (without the read-response fix) logged 'forwarded' but the host never received the POST. Confirmed by curl-with-token reaching the handler vs raw-socket-write-only failing.

PR #600, opened 2026-04-29
B.9: Window Reality Reconciliation (observation + pure-reducer self-heal)

AgentA-asaf:
## Summary

Closes the bug class surfaced after #595 / B.6.1: a CEF browser opens, gets a Win32 HWND, and is then \"lost\" to the user (parked at CEF's hidden sentinel position, off all monitors, never foregrounded). Pre-B.9 the reducer tracked **identity** but not **observability** — the existing drift compare (\`host.browsers.len() == launcher.windows.len()\`) saw both sides agree about a window that didn't actually exist for the user. Now the reducer treats Win32 HWND state as a first-class projection, classifies divergences at event-dispatch time, and emits **corrective actions in the same reducer call**. **No timers, no heartbeats** — every emission is in response to a Win32 OS event the host hooks via \`SetWinEventHook\`.

Design at \`docs/retro/wrr-design-2026-04-28.md\`. The PR folds B.9.1 (observation) and B.9.2 (pure-reducer self-heal) into one cohesive change since they share the protocol surface.

## Architecture

\`\`\`
HOST (agentmux-cef/src/wrr/)
   SetWinEventHook(WINEVENT_OUTOFCONTEXT) for OBJECT_CREATE / DESTROY /
   SHOW / HIDE / FOREGROUND / MINIMIZESTART / MINIMIZEEND /
   LOCATIONCHANGE — filtered to host PID + OBJID_WINDOW + CHILDID_SELF
   - Each event becomes a typed Command on the launcher pipe
   - Position events debounced 50ms per HWND so drag bursts don't flood
   - on_after_created sends an EXPLICIT report_hwnd_opened with known
     label, eliminating the OS-event timing race against
     pending_window_creations
                          ▼  ReportHwnd* / ReportMonitorTopologyChanged
LAUNCHER REDUCER (agentmux-launcher/src/wrr/)
   Reducer arm classifies six divergences:
   - HwndWithoutBrowser, BrowserWithoutHwnd
   - HiddenSinceOpen, OffMonitor
   - OrphanDestroy, LingeringHwnd
   Emits Event::HwndDriftDetected per classification.
   For OffMonitor with foregrounded_since_open == false, ALSO emits
   Event::CorrectiveWindowMove { hwnd, target_rect, reason } in
   the SAME reducer call.
                          ▼  Event::CorrectiveWindowMove
HOST ENFORCER (agentmux-cef/src/ui_tasks.rs::CorrectiveWindowMoveTask)
   apply_event_to_shadow matches the event, posts a UI task that
   calls SetWindowPos(hwnd, NULL, x, y, w, h, SWP_NOZORDER |
   SWP_NOACTIVATE) on the next CEF UI tick.
\`\`\`

## Pure-reducer guarantee

Every \`HwndDriftDetected\` and every \`CorrectiveWindowMove\` is emitted **at the moment a Win32 OS event is dispatched through the reducer**. There is no clock task, no \"if N seconds elapsed do X\" — the trigger is always a state transition, never the passage of time.

The corrective guard is purely state-based: \`mirror.foregrounded_since_open == false\`. Once the user has interacted with a window (even one foreground event), the guard flips and **no auto-correction ever fires for that label** — we trust the user's position choices for the rest of that label's lifetime. This is the line between \"buggy state we should heal\" and \"legitimate user state we shouldn't touch\".

## Sentinel suppression

CEF Views briefly parks new top-level windows at the Win32 hidden sentinel positions (\`(-32000, -32000)\` and \`(-31970, -31970)\`-family) between \`CreateWindow\` and the first \`SetWindowPos\`. Naively firing \`OffMonitor\` drift on every legitimate window-open transient produces log noise. The reducer suppresses **drift logging** for these positions but **still emits the corrective move** — we want the window OFF the sentinel before the user notices.

## Smoke evidence (v0.33.489 portable)

\`\`\`
curl POST /ipc {\"cmd\":\"open_new_window\"} → status 200
[launcher] WRR-POS hwnd=0x22804ea rect=(-31970,-31970)-(-31340,-30871)
             linked=Some(\"window-fe8b1296...\") monitors=1
[launcher] WRR-DRIFT [Warn] HiddenSinceOpen — visibility classifier
                                              fired (no foreground event yet)
[host]     WARN [wrr] CorrectiveWindowMove hwnd=0x22804ea
             reason=OffMonitor target=(135,400)-(765,1200)
[host]     INFO [wrr] corrective SetWindowPos hwnd=0x22804ea
             -> (135,400) 630x800 ok=true
\`\`\`

The curl-window — which on v0.33.481 / v0.33.488 produced an unkillable orphan taskbar entry — now lands centered on the primary monitor on first paint. Confirmed visually: no orphan, real body, focusable.

## Tests

48 launcher tests pass (44 prior + 4 new):
- \`wrr_off_monitor_position_for_unseen_window_emits_drift_and_corrective\`
- \`wrr_sentinel_position_suppresses_drift_but_emits_corrective\`
- \`wrr_off_monitor_after_user_foregrounded_does_not_correct\`
- \`wrr_off_monitor_with_no_known_monitors_emits_neither\`
- 7 rect-math tests (intersects, half-open semantics, multi-monitor, empty cases)

## What this DOES NOT do

- **Phase F enforcer breadth**: only \`OffMonitor\` triggers a corrective. \`OrphanDestroy\`, \`HwndWithoutBrowser\`, \`LingeringHwnd\` are observation-only (logged, not auto-fixed). Adding correctives for those is straightforward (each gets its own \`Event::Corrective*\` variant) but deferred — the orphan-taskbar bug is the immediate user-visible one and the rest are rarer.
- **Mid-session monitor topology change**: \`WM_DISPLAYCHANGE\` wiring is a B.9 follow-up. Initial topology is enumerated at \`install_hooks\` time; if the user unplugs a monitor mid-session, drift won't re-evaluate stranded windows until the next position event from any source.
- **Source-side fix for *why* CEF Views parks windows at the sentinel**: the reducer self-heal is a safety net, not a substitute. A separate PR can investigate why \`open_window_with_kind\`'s requested position isn't being applied to the new CefWindow.

## Test plan

- [ ] Launch portable, look for \`[wrr] installed 4 WinEventHook range(s)\` and \`[wrr] reported initial monitor topology: N monitor(s)\` in host log.
- [ ] Curl \`open_new_window\` → expect WRR-DRIFT \`HiddenSinceOpen\` + \`CorrectiveWindowMove\` in launcher log, \`SetWindowPos ok=true\` in host log, visible window centered on monitor (NOT a stray taskbar entry).
- [ ] Open a window through the UI normally → no drift, no corrective (the OS happy path).
- [ ] Open a window, focus it, drag it off-screen → drift fires (\`OffMonitor\`), corrective does NOT (foregrounded guard).
- [ ] Force-kill a render PID → \`OrphanDestroy\` drift fires.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.487"
+version = "0.33.488"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.487"
+version = "0.33.488"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.487"
+version = "0.33.488"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.487"
+version = "0.33.488"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.488"
+version = "0.33.489"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.488"
+version = "0.33.489"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.488"
+version = "0.33.489"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.488"
+version = "0.33.489"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.486"
+version = "0.33.487"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.486"
+version = "0.33.487"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.486"
+version = "0.33.487"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.486"
+version = "0.33.487"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.485"
+version = "0.33.486"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.485"
+version = "0.33.486"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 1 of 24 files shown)


a5af:
@codex review

AgentA-asaf:
@codex review

AgentA-asaf:
Pushed fix (aaa1bf9d) for both reagent P1s:

- Non-Windows `install_hooks(_state: Arc<AppState>)` stub signature matches the Windows version → cross-platform compile restored
- `apply_hwnd_destroyed` now emits `Event::WindowClosed` + `Event::WindowInstanceReleased` after the `OrphanDestroy` drift, matching the normal close path. Prevents stale frontend state after a renderer crash. New test `wrr_orphan_destroy_emits_window_closed_and_instance_released` asserts the chain. 49 launcher tests pass.

Will retrigger codex once @codex auth is restored.

AgentA-asaf:
Fixed codex P2 in 45ba43fe: same-HWND duplicate ReportHwndOpened is now a no-op (recognizing the dual-source happy path: WinEvent CREATE + on_after_created can legitimately fire twice with the same hwnd). Added two reducer tests covering both branches.

Codex P1 (non-Windows install_hooks signature) was stale context — the fix landed in aaa1bf9. Verified `git show aaa1bf9:agentmux-cef/src/wrr/mod.rs` shows the non-Windows stub now takes `std::sync::Arc<crate::state::AppState>`. No further action needed there.

51 launcher tests pass.

AgentA-asaf:
Pushed 520f271c addressing reagent round-2:

**P1**: `apply_hwnd_destroyed` no longer early-returns on a stale pending entry — it now drops pending unconditionally and proceeds to the mirror check, so the OrphanDestroy + WindowClosed + WindowInstanceReleased chain runs on renderer crash even when the dual-source design left a co-existing stale entry. AND `apply_hwnd_opened` now drains the matching pending entry on every successful link, so the stale never accumulates in the first place. (Both layers — defense in depth.)

**P2**: dropped `_unused_point` + the POINT import.

New test `wrr_orphan_destroy_runs_even_with_stale_pending_entry` reproduces the exact dual-source race reagent traced. 52 launcher tests pass.

Codex's repeated P1/P2 flags on the previous commits are stale (verified `git show HEAD:agentmux-cef/src/wrr/mod.rs` shows the non-Windows stub takes `Arc<AppState>`; same-hwnd duplicate path was already a no-op as of 45ba43fe).

PR #601, opened 2026-04-29
B.9.3: fix host-doesn't-quit-on-close-all (pool-refill cascade)

AgentA-asaf:
## Summary

Closes the regression introduced by #566 (warm pool feature, 2026-04-27): closing all visible AgentMux windows used to leave the host process tree alive forever, accumulating zombie instances. After this PR, host exits cleanly within ~30ms of the last close.

## Root cause (one sentence)

CEF's \`quit_message_loop\` is documented as **QuitWhenIdle** — it sets a flag that fires when all CefBrowsers close, but the warm pool's auto-refill kept replacing every closed pool browser with a new one, so idle was never reached.

## Fix

Three coordinated changes, all host-side, all cross-platform:

1. **\`AppState::is_quitting: AtomicBool\`** — new flag, default false.
2. **\`spawn_pool_window\` early-return on \`is_quitting=true\`** — suppresses refill during drain.
3. **Two-stage \`on_before_close\` cascade**:
   - Stage 1 (\`user_count==0 && !is_pane\`): set \`is_quitting=true\`, async \`PostMessage(hwnd, WM_CLOSE)\` on every remaining pool browser. Lets the current \`on_before_close\` unwind without re-entering CEF.
   - Stage 2 (\`self.browser_list.is_empty() && !is_pane\`): canonical cefsimple pattern — call \`quit_message_loop\`. Fires only on the absolute last close.

Only the \`PostMessage\` step is Windows-specific (macOS uses \`NSWindow.performClose:\`, Linux uses X11 \`WM_DELETE_WINDOW\` ClientMessage). Same semantics.

## Smoke evidence

v0.33.501 portable, multiple sessions including:
- single tear-off close-tear-off-first
- single tear-off close-main-first
- multiple tear-offs with variations
- pane tear-offs included

Every session: process tree empty after close-all. Cascade timeline (~30ms):

\`\`\`
[wrr] is_quitting=true (drain mode)
[wrr] stage 1: user_count==0; PostMessage WM_CLOSE on 3 pool browser(s)
[wrr] stage 1: ... on 2 pool browser(s)
[wrr] spawn_pool_window skipped — is_quitting=true (drain mode)   ← refill suppressed
[wrr] stage 1: ... on 1 pool browser(s)
[wrr] stage 1: ... on 0 pool browser(s)
[wrr] stage 2: self.browser_list.is_empty() — calling quit_message_loop
[wrr] quit_message_loop returned
\`\`\`

Process tree confirmed empty via \`tasklist\`.

## What's NOT in this PR (reverted from earlier iterations)

The earlier B.9.3 design proposed a launcher-reducer-driven saga (\`Event::HostShouldQuit\` emitted when \`state.windows\` becomes empty). I built it across v0.33.491–v0.33.494, then learned empirically that delivering ANY work to the host's UI thread during the close-cascade window is unsolvable:
- \`cef::post_task(TID_UI, ...)\` silently drops new tasks
- Direct \`cef::quit_message_loop()\` from worker thread is UB
- Win32 \`PostThreadMessage(WM_QUIT)\` is ignored by CEF's custom pump

So the saga's delivery side was a dead-end. With the host-side fix in this PR, the saga is unnecessary — the host quits naturally. The IPC type additions (\`Event::HostShouldQuit\`, \`HwndDriftKind::OrphanInstance\`) remain in \`agentmux-common\` for any future observability use, with the host-side handler reduced to a single tracing line.

Full reasoning in \`docs/retro/b9-3-lifecycle-analysis.md\` and \`docs/retro/b9-3-quit-thread-analysis.md\`.

## Diagnostic logs

Structural signals (drain start, per-stage cascade, refill suppression, stage 2 quit) at \`tracing::warn!\` — operator visible in production. Verbose per-browser ENTER/EXIT trace at \`tracing::debug!\` with target \`wrr-trace\` — filtered out by default. Enable when investigating:

\`\`\`
RUST_LOG=\"info,wrr-trace=debug\"
\`\`\`

## Tests

- 55 launcher reducer tests pass (4 new B.9.3 cases for OrphanInstance signal — kept since the IPC types are kept; reducer-side observability without the saga delivery)
- Manual smoke confirmed across 4+ variations + pane tear-offs

## Test plan

- [ ] \`task package\` builds clean
- [ ] Launch portable, tear off any tab, close every visible window → process tree exits within seconds
- [ ] Repeat with multiple tear-offs (3+) → still exits cleanly
- [ ] Repeat with pane tear-offs → still exits cleanly
- [ ] \`task dev\` (no launcher) regression check — host exits when last visible window closes
- [ ] After close-all and clean exit, double-clicking exe again starts a fresh instance (no zombie blocking the named-pipe)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.490"
+version = "0.33.491"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.490"
+version = "0.33.491"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.490"
+version = "0.33.491"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.490"
+version = "0.33.491"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.491"
+version = "0.33.492"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.491"
+version = "0.33.492"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.491"
+version = "0.33.492"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.491"
+version = "0.33.492"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.500"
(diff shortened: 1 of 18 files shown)


a5af:
@codex review

AgentA-asaf:
Pushed a6872225 addressing reagent P1: non-Windows now defers close_browser via `cef::post_task(TID_UI, ClosePoolBrowserTask)`, gated by `#[cfg(not(target_os = "windows"))]`. Windows path unchanged (PostMessage WM_CLOSE remains the proven bridge). Both paths achieve the same async-close-without-re-entrance-hang.

Smoke on v0.33.501 (Windows) already verified the working cascade end-to-end. The non-Windows code is additive and excluded from Windows builds, so behavior on Windows is byte-identical to v0.33.501. Non-Windows smoke is gated on Phase 7 (cross-platform parity work).

AgentA-asaf:
Pushed 8042fbbb addressing codex P1: Windows path now falls back to `cef::post_task(TID_UI, ClosePoolBrowserTask)` when `host.window_handle()` returns null, so a pool browser without a top-level HWND still gets a close request. Without the fallback, self.browser_list could keep that entry forever and Stage 2 would never fire. Same task type used by non-Windows path; `wrap_task!` definition is now unconditional.

PR #607, opened 2026-04-29
Phase D.1: GetSnapshot RPC + diag state-now view

AgentA-asaf:
## Summary

Phase D.1 — adds the foundation for Phase D's durability/resync story. Subscribers can now ask the launcher reducer for its canonical state in a single round-trip via `Command::GetSnapshot`, instead of inferring state from the event stream that has flowed since they connected.

First piece of Phase D. D.2 adds a persisted event log (ring buffer); D.3 adds the `since: u64` parameter for delta replay.

### Wire protocol (`agentmux-common/src/ipc.rs`)

- New `Command::GetSnapshot` (no args)
- New `Event::Snapshot { version, lifecycle, windows, pool, instance_registry, backend_window_ids, monitors }`
- New `WindowSnapshot` struct — serializable view of one window (label, kind, parent, plus the B.9 HWND-observation axis: hwnd, visible, iconic, last_rect, foregrounded_since_open)

### Reducer (`agentmux-launcher/src/reducer.rs`)

`handle_get_snapshot` clones canonical state into a Snapshot event. Sorted by label for deterministic ordering. The snapshot's `version` is `bump_version`'d so it's monotonically distinct from prior events — subscribers know the "as-of" point for applying subsequent events as deltas.

**Excluded by design:** `processes` (PID metadata is launcher-internal), `pending_hwnds` (transient), event log (D.2 adds that separately).

### Server (`agentmux-launcher/src/ipc/server.rs`)

`enforce_register_first` accepts `GetSnapshot` as a non-fatal pre-Register error (same precedent as `Ping`).

### Diag (`agentmux-launcher/src/diag.rs`)

`--diag wrr` now sends `GetSnapshot` right after Register and prints the snapshot before the 2s live-event observation window:

```
=== Snapshot (event_version=12) ===
Lifecycle: Running
Monitors:  1 ([Rect { left: 0, top: 0, right: 900, bottom: 1600 }])

Windows (1):
    #1 main                           kind=FullInstance hwnd=Some(2557122) visible=true iconic=false fg_seen=true backend=cc92f5e2-9245-4936-904d-f6ed29effe85

Pool (3): ["window-pool-149fb...", "window-pool-8178...", "window-pool-c598..."]

=== Live stream observed in 2s ===
...
```

Operator surface goes from "events-since-connect" to "what IS now + what's happening".

## Test plan

3 new launcher reducer tests (63 total, all passing):

[...]

PR #608, opened 2026-04-29
Phase D.2 + D.3: event log + GetEvents replay (one PR)

AgentA-asaf:
## Summary

Combined Phase D.2 (persisted event log) and Phase D.3 (replay protocol) into a single PR — they interlock by design (D.3's `GetEvents` reads from D.2's event log).

After this lands, **Phase D is complete**. Subscribers can fetch a snapshot AND replay missed events to fully resync after disconnect, all without polling. Next: Phase E (srv reducer).

Also addresses codex P2 from PR #607 as a subitem.

### D.2 — event log (`agentmux-launcher/src/event_log.rs`, new)

- **In-memory ring** (`VecDeque<Event>`, capped at 4096 events) — source of truth for replay during the launcher's lifetime. Bounded so long sessions don't grow unboundedly; oldest evict first. Synchronous, microsecond-scoped Mutex.
- **Optional disk persistence** at `<data-dir>/launcher-events.log` — JSON-lines append-only, capped at 8 MiB with two-file rotation to `.old`. Crash-forensics-only, fire-and-forget. Failures log a warning but never block the in-memory path.
- Two roles kept clean: in-memory ring satisfies D.3's resync needs at zero I/O cost; disk satisfies the "what was the launcher doing right before it died" debugging story without complicating the replay reader.
- 6 unit tests: ring growth + eviction at cap, `events_since` boundary semantics, `replay_truncated` detection.

### D.3 — replay protocol

- **New `Command::GetEvents { since: u64 }`** — exclusive lower bound; reply contains all retained events with `version > since` in ascending order.
- **New `Event::EventList { events: Vec<Event>, version: u64 }`** — the reply.
- **Server intercepts `GetEvents` before reducer dispatch** — log query is I/O-adjacent; reducer stays pure. Bumps `event_version`, queries log, broadcasts the EventList. Reducer's `Command::GetEvents` arm returns `Vec::new()` for match exhaustiveness only (unreachable in practice).
- **In-memory ring append happens in the broadcast loop** — synchronously, before `events_tx.send`, so a concurrent `GetEvents` query sees consistent results. `Snapshot`/`EventList`/`Error` events excluded — meta-events would create recursive replay or per-client noise.
- **Disk writer task** spawned in main.rs subscribes to the broadcast bus alongside the IPC server.

### Diag wiring

`--diag wrr` now sends `GetSnapshot` + `GetEvents { since: 0 }` after `Register`, giving operators the full picture in one shot:

```
=== Snapshot (event_version=12) ===
Lifecycle: Running
Monitors:  1 ([Rect { left: 0, top: 0, right: 900, bottom: 1600 }])

Windows (1):
    #1 main                           kind=FullInstance hwnd=Some(7930952) visible=true iconic=false fg_seen=true backend=b21e9803-...

Pool (3): ["window-pool-...", ...]

=== EventList replay (event_version=13, 11 event(s)) ===
  [0] {"event":"process_spawned","pid":7980,"kind":"host",...}
  [1] {"event":"lifecycle_phase_changed","from":"starting","to":"running",...}
  [2] {"event":"registered","client_id":1,...}
  [3] v=  4 WindowOpened       label=main kind=FullInstance parent=None
  [4] v=  5 InstanceAssigned   label=main num=1
  ...

=== Live stream observed in 2s ===
...
```

### Codex P2 fix from PR #607 (subitem)

`print_summary` previously selected `snapshot.first()`. With the broadcast bus serving multiple concurrent diag/Tool clients, this could pick another client's stale snapshot rather than ours. Switched to `snapshot.last()` — biases toward freshness since our own snapshot reply is monotonically the latest among captured events on the bus.

## Test plan

- 69 launcher tests pass (63 prior + 6 new event_log).
- Smoked locally on v0.33.508 — diag output above is from a real run.
- Disk log persisted at `<data-dir>/launcher-events.log` with matching JSON-line content.

## Phase D status after this PR

- **D.1** ✅ (PR #607) — `Command::GetSnapshot` + `Event::Snapshot`
- **D.2** ✅ (this PR) — event log (in-memory ring + disk persistence)
- **D.3** ✅ (this PR) — `Command::GetEvents` + `Event::EventList` + replay correlation in diag

## What this does NOT do

- No `IdempotentApply` Rust trait formalization. Existing apply paths in `apply_event_to_shadow` are already idempotent in practice; adding a typed contract is more value when there are multiple apply paths (Phase E adds the second one — appropriate moment to extract the trait then).
- No frontend resync wiring (renderer reducer doesn't consume EventList yet) — natural follow-up after Phase E.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.507"
+version = "0.33.508"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.507"
+version = "0.33.508"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.507"
+version = "0.33.508"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.507"
+version = "0.33.508"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/VERSION_HISTORY.md
+++ b/VERSION_HISTORY.md
@@ -14,6 +14,7 @@ Auto-appended by `scripts/package-cef-portable.sh`. Newest first.
 
 | Version | Date | ZIP (compressed) | Folder (uncompressed) | Note |
 |---------|------|------------------|-----------------------|------|
+| 0.33.508 | 2026-04-29 | 160.4 MiB | 335.1 MiB | |
 | 0.33.507 | 2026-04-29 | 160.3 MiB | 334.8 MiB | |
 | 0.33.506 | 2026-04-29 | 160.3 MiB | 334.8 MiB | |
 | 0.33.505 | 2026-04-29 | 160.3 MiB | 334.8 MiB | |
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

PR #626, opened 2026-05-01
feat(E.7): --diag srv operator command

AgentA-asaf:
## Summary

Phase E.7 partial — adds \`agentmux.exe --diag srv\` Tool client mirroring the existing \`--diag wrr\` for the srv reducer. Connects to the srv pipe, sends \`GetSrvSnapshot\` + \`GetEvents\`, observes for 2 s, prints a summary.

## Use cases

- "What does the srv reducer think the workspace tree looks like?" — vs SQLite, vs frontend atom state.
- "Did saga X complete or fail?" — Saga lifecycle counts surface this without scraping logs.
- "What recent srv activity has there been?" — EventList replay shows recent reducer events with compact one-liners.

## What's printed

- **SrvSnapshot** — Workspaces (id / name / tab count / active tab) → Tabs (id / parent ws / name / block count) → Blocks (id / parent tab).
- **EventList replay** — last 20 of N events from srv's in-memory ring + on-disk event log.
- **Saga lifecycle counts** across snapshot + replay + live stream.
- **Live stream** observed during the 2 s window.

## Implementation

- \`agentmux-launcher/src/diag.rs::run_srv_diag\` mirrors \`run_wrr_diag\` with srv pipe + SrvSnapshot variant.
- \`srv_event_kind_name\` / \`format_srv_event\` cover all srv-emitted Event variants.
- \`agentmux-launcher/src/main.rs\` dispatch table extended: \`--diag srv\` → \`run_srv_diag\`. Usage strings + unknown-topic error mention both topics.
- Non-Windows stub returns "Phase 7 will add Unix domain socket parity" — same pattern as \`run_wrr_diag\`.

## Phase E.7 status after this merges

| Item | Status |
|---|---|
| \`--diag srv\` | ✅ this PR |
| \`--diag sagas\` | 🟡 currently subsumed by \`--diag srv\`'s saga counts. Could split out if more saga-specific introspection is wanted later (in-flight registry doesn't exist — sagas run synchronously through \`run_saga\`). |
| Property tests for reducer arms | ❌ separate PR — independent test infrastructure work, no file overlap. |

## Test plan

- [x] \`cargo check -p agentmux-launcher\` clean.
- [ ] Live verify: \`agentmux.exe --diag srv\` against a running portable shows the workspace tree + recent events.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #627, opened 2026-05-01
feat(E.7): property tests for reducer arm invariants

AgentA-asaf:
## Summary

Phase E.7 partial — adds proptest-based randomized testing for the reducer's whole-state invariants. Catches regressions where an individual arm looks correct in unit tests but interacts with sibling arms in a way that violates the reducer's cross-arm contract.

### 1. \`invariants_hold_across_random_sequences\`
Applies up to 40 random valid commands (Create/Delete Workspace/Tab/Block) and asserts after each:
- **Version monotonicity** — every emitted event has a strictly-greater version than the previous.
- **Referential integrity** — tabs reference real workspaces; blocks reference real tabs; workspace.tab_ids and tab.block_ids point at real entities.
- **Active-tab validity** — every workspace's active_tab_id is either None or in its own tab_ids list.

### 2. \`delete_workspace_cascades_cleanly\`
Builds a workspace with N tabs each holding M blocks (N, M proptest-generated), deletes the workspace, asserts the entire subtree is removed (no orphan tabs or blocks). Companion to the existing unit-test cascade tests but with size variation.

## Implementation notes

- \`proptest = "1"\` added as dev-dependency (default-features=false to skip the timeout-fork machinery; we only need strategy combinators + \`proptest!\` macro).
- \`PropOp\` enum + \`op_strategy()\` resolves "abstract intent" ops to actual \`Command\`s by picking a target ID from the current state. Required because IDs are reducer-generated and can't be pre-computed in the strategy.
- 64 cases per proptest run (default 256 was overkill).

## Phase E.7 status after this merges

- ✅ \`--diag srv\` (#626 in flight)
- ✅ Property tests (this PR)
- 🟡 Integration tests (real coordinator + reducer + subscriber + in-memory wstore) — could be a follow-up if needed; the new property tests + existing unit tests cover the main invariants.
- 🟡 \`--diag sagas\` (subsumed by \`--diag srv\`'s saga counts; could split if more saga-specific introspection is wanted)

After this + #626 merge, the next planned work is a Phase F spec draft (no implementation — just spec the host reducer architecture).

## Test plan

- [x] \`cargo test reducer::tests::invariants_hold_across_random_sequences\` — 64 cases pass.
- [x] \`cargo test reducer::tests::delete_workspace_cascades_cleanly\` — passes across full N×M parameter space.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,6 +82,7 @@ dependencies = [
  "notify",
  "parking_lot",
  "portable-pty",
+ "proptest",
  "reqwest",
  "rusqlite",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.535"
+version = "0.33.536"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.535"
+version = "0.33.536"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #621, opened 2026-04-30
feat(E.5.5+6): tear-off sagas — fixes new-window-+-tab smoke regression

AgentA-asaf:
## Summary

Phase E.5.5 + E.5.6 — fixes the tear-off → "+" tab smoke regression by routing the multi-step tear-off / restore operations through the srv reducer via a saga coordinator. The reducer's session-only projection now sees the new workspace + tab + block immediately after tear-off, so the new window's subsequent CreateTab / etc. calls succeed instead of erroring on "workspace not found".

Plus carry-forward fix: codex P1 #620 — relaxes ReorderTabsBulk validation since wcore-direct tab moves leave the reducer's tab_ids stale during the migration window.

## What changed

**Wire types (agentmux-common::ipc):**
- `Command::MoveTab { tab_id, src_workspace_id, dst_workspace_id, dst_index }`
- `Command::MoveBlock { block_id, src_tab_id, dst_tab_id, dst_index }`
- `Event::TabMoved { ..., new_src_active_tab_id, version }`
- `Event::BlockMoved { ..., version }`

**Reducer arms (agentmux-srv::reducer):**
- `handle_move_tab` — cross-workspace move; updates both workspaces' tab_ids; updates tab.workspace_id; reassigns src.active_tab_id if needed. Rejects same-workspace moves.
- `handle_move_block` — cross-tab + intra-tab; clamps dst_index; updates block.tab_id on cross-tab.

**Subscriber arms (persist_subscriber):**
- `apply_tab_moved` — drains tab from src.tabids/pinnedtabids, inserts at clamped dst_index in dst, updates src.activetabid. Idempotent on re-delivery.
- `apply_block_moved` — handles intra + cross-tab; updates block.parentoref.

**Saga coordinator (agentmux-srv/src/sagas/, new module ~1030 LOC):**
- `SagaCtx` — dispatch / compensate / state_lock helpers.
- `run_saga(name, fut)` — 5 s timeout wrapper.
- Per-saga lifecycle event emission (SagaStarted / SagaCompleted / SagaFailed).
- 3 sagas: TearOffTab, TearOffBlock, RestoreTornOffTab.

**RPC migrations (service.rs):**
- TearOffTab handler → `sagas::tear_off_tab::run`.
- TearOffBlock handler → `sagas::tear_off_block::run` + post-saga layout setup (E.4 stays wcore-direct).
- RestoreTornOffTab handler → `sagas::restore_torn_off_tab::run`. Legacy `was_pinned` arg ignored.

[...]

PR #638, opened 2026-05-01
feat(E.4): flip handle_move_tab to strict mode

AgentA-asaf:
## Summary

E.4 Option A has soaked since 2026-05-01 with no `lazy-import` warnings observed in production logs. This PR flips `handle_move_tab` from migration-tolerant to strict mode by removing the lazy-import fallback that was added in PR #621.

### Behavior change

`handle_move_tab` now rejects (`Event::Error` / `ErrorCode::InvalidCommand`):

- **Unknown tab id** -> `MoveTab: tab not found in state: <id>` (was: lazy-imported a synthetic `TabRecord`)
- **Mismatched workspace_id** -> `MoveTab: workspace_id mismatch -- tab <id> belongs to <actual>, not <claimed>` (was: silently re-parented)

The same-workspace, unknown-src-workspace, and unknown-dst-workspace rejections are unchanged.

### Tests

- **Added** `move_tab_unknown_tab_rejects` -- asserts `Event::Error` with `tab not found` message and that the unknown id is NOT lazy-inserted into `state.tabs`. Replaces `move_tab_lazy_imports_unknown_tab`.
- **Added** `move_tab_wrong_workspace_rejects` -- asserts `Event::Error` with `workspace_id mismatch` and that all three workspaces (real_src, other, dst) are untouched. Replaces `move_tab_tolerates_workspace_id_mismatch_during_migration`.
- Existing happy-path test (`move_tab_cross_workspace_updates_lists_and_parent`) and the unknown-src/dst rejection test (`move_tab_rejects_unknown_src_or_dst_or_tab`, with its ghost-tab arm flipped to expect `Error`) continue to pass.

### Dead code removed

- The migration-tolerant comment block (~14 lines) explaining lazy-insert semantics.
- The `match state.tabs.get_mut(&tab_id)` arms with the lazy `state.tabs.insert(...)` synthetic `TabRecord` path -- replaced with a single `expect("checked")` after the strict pre-check.
- Comment about stale-state tolerance on the `src.tab_ids.retain(...)` block.

Diff: +102 / -88 lines in `agentmux-srv/src/reducer.rs`.

## Test plan

- [x] `cargo check --workspace`
- [x] `cargo test -p agentmux-srv --bin agentmux-srv reducer::` (68 tests pass)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv sagas::` (55 tests pass, all tear-off / promote / restore sagas still green)
- [x] Two new strict-mode test cases assert error path + state-untouched invariants

## Notes

- No production-runtime feature flag is needed: the lazy-import path was reducer-internal; sagas/RPC handlers already pre-validate against SQLite (the source of truth) before calling `MoveTab`, so any caller that previously slipped past would also have failed the SQLite-layer check.
- 5 pre-existing test failures in `backend::*` (storage / providers / wcore) are unrelated to this change and reproduce on `origin/main` without these edits.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.557"
+version = "0.33.558"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.557"
+version = "0.33.558"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.557"
+version = "0.33.558"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.557"
+version = "0.33.558"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

a5af:
@codex review

PR #650, opened 2026-05-02
fix(cef): pane focus-redirect guard + Phase 1 window-creation tracing

AgentA-asaf:
## Summary

Two changes from the 2026-05-02 freeze investigation, neither alone fixes the freeze but both ship real value:

**A — Pane WM_SETFOCUS cross-window guard + per-root rate limit** (\`agentmux-cef/src/pane/hwnd.rs\`):
The pane focus-redirect (\`SetFocus(GA_ROOT)\` to defeat JS \`window.focus()\`) can itself drive a foreground ping-pong between two top-level windows when both have pane content that programmatically requests focus. Two guards added:
- Cross-window refusal: skip the redirect if a *different* top-level HWND owns OS foreground per \`GetForegroundWindow()\`. Programmatic focus from page content shouldn't fight the AgentMux window the user is interacting with.
- Per-root rate limit: cap redirects at 1 per 100ms per root HWND.

**A2 — Synchronous host log init sentinel** (\`agentmux-cef/src/main.rs\`):
The pointer file at \`<log_dir>/current-host-v<version>.path\` previously pointed at a tracing_appender::rolling::daily file that didn't materialize until the non-blocking writer's first flush. Crashes between subscriber-setup and first flush left the pointer pointing at a missing file (observed during 0.33.579 investigation). Sentinel guarantees the file exists once init_logging runs past pointer-write — if it's missing afterwards, we know init_logging didn't get past that point, which is itself diagnostic.

**B — Phase 1 diagnostic tracing for window creation** (\`agentmux-cef/src/{commands/mod.rs, ui_tasks.rs, client.rs}\`):
Adds focused tracing across the three stages of top-level window creation to identify the exact line that silences the UI thread under concurrent load. Three instrumentation points:

1. \`create_isolated_request_context\` — entry, pre-call, and post-return spans around \`request_context_create_context\`.
2. \`CreateWindowTask::execute\` — per-stage \`elapsed_us\` across get-client, request_context resolved, browser_view_create returned, window_create_top_level returned.
3. \`on_after_created\` — tracing inside the gap between "Browser created" and "Registered browser" (where the dump analysis showed the UI thread silenced 2026-05-02), specifically around \`DequeuePendingWindowCreation\` dispatch and \`browsers\` lock acquisition.

Pure-additive logging on already-instrumented paths — no behavior change.

## Investigation context

Full evidence trail in:
- \`docs/specs/SPEC_WINDOW_FLEET_REDUCER_2026-05-02.md\` — diagnostic spec covering the freeze investigation, why per-window-isolated RequestContext triggers it, and why the focus-redirect guard alone doesn't fix it (pane WM_SETFOCUS path doesn't run before the freeze).
- \`docs/specs/SPEC_HOST_WINDOW_CREATION_RUNNER_2026-05-02.md\` — implementation plan for Phases 1-4 (this PR is Phase 1; the runner is in a follow-up PR).

The Phase 1 tracing landed alongside successful completion data: window creation typically takes 25-100ms total with \`request_context_create_context\` returning in 2.7-5.8ms (median ~3.6ms). The wedge — when it occurs — is in CEF's async browser-creation pipeline AFTER \`window_create_top_level\` returns, between Views handing off to Chromium and \`on_after_created\` firing. That's libcef.dll's renderer-spawn / browser-init machinery, confirmed by trace data on creation_id=6 in 0.33.582 testing.

## Test plan

- [x] cargo check -p agentmux-cef clean (warnings only, all pre-existing)
- [x] All existing reducer tests still pass (5/5)
- [x] Manual smoke: 0.33.580 + 0.33.581 portables built and validated; user opened multiple windows, no freeze in 0.33.581 testing
- [ ] Bot review (reagentx-workflow + codex)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.578"
+version = "0.33.579"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.578"
+version = "0.33.579"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.578"
+version = "0.33.579"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.578"
+version = "0.33.579"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.578"
+version = "0.33.579"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

PR #652, opened 2026-05-02
fix(cef): handle CEF early-exit + DLL path fallback for dev mode

a5af:
## Summary

`task dev` was failing on every launch with `assertion left == right failed: CEF initialization failed (left: 0, right: 1)`. Two compounding issues, plus an observability fix.

### 1. DLL search path missing in dev mode
- Portable / installed builds: the launcher (`agentmux-launcher`) calls `SetDllDirectoryW(<root>/runtime)` and the host inherits it.
- Dev: `task dev` runs the host directly with no launcher. The host's own defensive `SetDllDirectoryW` block (`agentmux-cef/src/main.rs:49-65`) only fired when a `runtime/` subdir existed next to the exe — which is true in **neither** the portable nor the dev layout. Portable worked only because the launcher set it first.
- **Fix:** added an `else` branch — if no `runtime/` subdir, set DLL path to the host's own directory. Idempotent in portable mode, correct in dev mode.

### 2. Misuse of `cef::initialize()` return value
- CEF docs say `0` from `cef_initialize` means **either** real init failure **or** "normal early exit" (process singleton, AUTO_DE_ELEVATED, PACK_EXTENSION_SUCCESS, etc.). To distinguish, the caller MUST call `cef_get_exit_code()`.
- Our code had a hard `assert_eq!(init_result, 1, "CEF initialization failed")` that panicked on every early-exit case as if it were fatal.
- The visible trigger this round: running `task dev` from an elevated shell. Chromium refuses to run elevated, returns code `38` (`CEF_RESULT_CODE_NORMAL_EXIT_AUTO_DE_ELEVATED`), and our assertion converted that clean exit into a panic that killed the launcher chain.
- **Fix:** replaced the assertion with a branch on `cef::get_exit_code()`. Codes `0 / 24 / 36 / 38` (documented normal exits) → exit cleanly with status 0. Anything else → log the code and exit non-zero.

### 3. Bonus — `log_file` + `log_severity` in CEF `Settings`
- Without these, CEF's internal Chromium logging is silent and `chrome_debug.log` stays empty even on init failures.
- That's exactly what wasted hours of guesswork on this issue. Now writes to `~/.agentmux/logs/cef-debug.log` (INFO severity) for any future CEF-side problem.

Full root-cause walkthrough in [`docs/analysis/cef-dev-init-failure-2026-05-01.md`](docs/analysis/cef-dev-init-failure-2026-05-01.md).

## Test plan
- [ ] `task build:host` clean
- [ ] `task dev` from a non-elevated terminal — host launches, window appears, Vite frontend loads at localhost:5173
- [ ] `task dev` from an elevated terminal — host exits cleanly with `INFO CEF early exit (process singleton or similar) — exiting cleanly exit_code=38` (no panic)
- [ ] `task package` portable build still works (DLL fallback shouldn't affect launcher-driven paths)
- [ ] First failure case in dev produces a populated `~/.agentmux/logs/cef-debug.log`
- [ ] No regression in multi-instance scenarios (portable + dev side-by-side, separate cache dirs)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.579"
+version = "0.33.580"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.579"
+version = "0.33.580"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.579"
+version = "0.33.580"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.579"
+version = "0.33.580"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #666, opened 2026-05-02
fix(linux): non-main windows route their own window IPCs

AgentU-asaf:
## Summary

On Linux/macOS, four window IPCs hardcoded their target label as `\"main\"`:
- `close_window`
- `minimize_window`
- `maximize_window`
- `start_window_drag`

So a second window opened from the status bar couldn't drag itself, double-click maximize on its header acted on the main window, and the close/minimize buttons routed to the wrong window. **User-visible**: left-click drag on the second window's header did nothing.

## Fix

- **Rust** (`commands/window.rs` + `ipc.rs`): each handler now reads `args.label` and forwards it to the corresponding `ui_tasks::post_*_window`. Default is `\"main\"` so single-window builds keep working unchanged.
- **Frontend**: reads `?windowLabel=…` (the same per-window URL query the launcher already sets) and passes it on every call:
  - `useWindowDrag.linux.ts` → `start_window_drag` + dblclick `maximize_window`
  - `cef-api.ts` → `closeWindow` / `minimizeWindow` / `maximizeWindow` (closeWindow already took an optional label; min/max now match)
- **Windows path is unchanged** — `find_own_top_level_window()` already resolves the per-process HWND, so multi-window worked there.

Bumped 0.33.590 → 0.33.591.

## Test plan

- [ ] Open a second window via the status bar panel.
- [ ] Left-click drag on the header of the second window → second window moves (not the main one).
- [ ] Double-click the header of the second window → toggles maximize on the second window.
- [ ] Close button on the second window → only the second window closes.
- [ ] Minimize button on the second window → only the second window minimizes.
- [ ] First window's behavior unchanged (still drags, maximizes, closes, minimizes itself).
- [ ] Right-click contextmenu still works on both (was never affected — purely renderer-side).
- [ ] macOS / Windows builds still compile.


a5af:
@codex review

a5af:
@codex review

PR #682, opened 2026-05-03
fix(linux): embedded browser panes via CEF Views (AddOverlayView + CUSTOM)

AgentU-asaf:
## Summary

Linux/macOS embedded browser panes (`defwidget@browser`) showed a black screen because the non-Windows code path in `browser_pane/creation.rs` was a stubbed `tracing::warn! + return`. This implements the Views-based pane creation per the design spec / research doc committed alongside this PR.

## Architecture

The Windows native-child-window path (`WindowInfo::set_as_child(parent_hwnd, rect)`) doesn't work on Linux Wayland (cef#2804 — the proposed `wl_subsurface` API isn't upstreamed and even when it ships is too constrained for hide/show/move). macOS embedded non-Views is officially unsupported by the CEF maintainer (cef-forum t=19688). The Views framework is the only viable cross-platform path.

Pane creation now uses **`Window::add_overlay_view`** (returning a `CefOverlayController`) rather than `add_child_view`. The host UI's main BrowserView fills the window via the existing FillLayout; pane BrowserViews are added as sibling overlays positioned via the controller. **Overlays don't participate in the parent's auto-layout** so multi-pane works without the FillLayout-stacking problem `add_child_view` exhibits.

## Two non-obvious findings during the spike

Both ended up driving the implementation; full notes in the spec's postscript section.

1. **`OverlayController::set_bounds(rect)` is silently ignored** under every `DockingMode` the cef Rust binding exposes — `TOP_LEFT` made the overlay fill the whole window, `CUSTOM` made it 0×0. The post-create diagnostic readback (`oc_w=0 oc_h=0`) caught this in one round-trip via the spike. The working pattern is `set_size(Size) + set_position(Point) + window.layout()`. We use `DockingMode::CUSTOM` to opt out of corner-docking altogether.

2. **`add_child_view` is the wrong primitive** even when paired with explicit `set_bounds`: the Window's default FillLayout reasserts "fill parent" on every layout pass and clobbers explicit bounds. With two FillLayout children both get full-parent-size bounds and the pane stacks on top of the host UI (we observed exactly this — full-window flashing pane). `AddOverlayView` avoids that entirely — overlays live outside the auto-layout system.

## Files

- `agentmux-cef/src/browser_pane/creation_views.rs` (new) — runs on the CEF UI thread. `browser_view_create + add_overlay_view(CUSTOM) + set_size + set_position + set_visible + window.layout()`. Stashes the OverlayController in `state.browser_pane_overlays`. Diagnostic readback of is_visible / is_drawn / bounds — kept post-merge to make future regressions debuggable from the host log alone.
- `agentmux-cef/src/browser_pane/creation.rs` — restructured: Windows native-child path stays inside `cfg(target_os = "windows")` unchanged; non-Windows delegates to `creation_views`. Removed the premature null-HWND check that was hard-aborting all non-Windows pane creates.
- `agentmux-cef/src/state.rs` — `main_window: Mutex<Option<cef::Window>>` and `browser_pane_overlays: Mutex<HashMap<String, OverlayController>>`, both `cfg(not(target_os = "windows"))`. Snapshot-and-drop at every read site (per Phase F.1 reducer discipline).
- `agentmux-cef/src/app.rs` — `AgentMuxWindowDelegate` gains `primary_window_state: Option<Arc<AppState>>`. Main window passes `Some(state)` to cache its handle for pane attach; popups + sub-windows pass `None`.
- `agentmux-cef/src/browser_panes.rs` — resize and close paths split by cfg. Linux/macOS post `ResizeBrowserPaneViewTask` / `DetachBrowserPaneViewTask` to the UI thread (View ops are UI-thread-only). Windows path unchanged.
- `docs/specs/embedded-browser-panes-linux-macos-2026-05-03.md` — full design spec.
- `docs/research/cef-pane-research-2026-05-03.md` — source-cited research backing the Views decision (CEF maintainer quotes, Wayland forum threads, Steam BrowserView precedent, etc.).

The host reducer state machine is unchanged across both paths — `on_after_created` still routes through the existing `client.rs` handler and dispatches `RegisterBrowser` to the reducer's `browsers` map. Pane lifecycle (Live/Closing/Closed) is identical.

## Verified empirically

Drove via the IPC `browser_pane_create` (port + token from the host log; same shape as the frontend's `invokeCommand` call) with `url=https://example.com, rect=(300,150,700,500)`.

Diagnostic readback:
```
oc_x=300 oc_y=150 oc_w=700 oc_h=500 oc_drawn=1 view_attached=1
```

`[pane-load-end]` fired. Two simultaneous panes (the frontend's github pane at its own rect plus the spike example.com pane at this rect) both rendered. Detach via `browser_pane_close` fired `OverlayController destroyed` and the example.com overlay disappeared.

## Risks / follow-ups

- **cef#3790** (overlay regressions in CEF 125) — our libcef is 146.0.179, post-fix.
- **cef#4035** (transparent overlay BrowserViews unsupported) — we use opaque background, not affected.
- **Multi-window panes** — main window only for now. Sub-windows pass `None` for `primary_window_state`. Generalizing needs a per-window-label store; follow-up PR.
- **Windows is untouched** — the existing native-child-window path stays. When Views is proven across more environments, Windows can migrate too and `pane-overlay.ts` (the SetWindowRgn airspace dance) can retire entirely.

## Test plan

- [ ] On Linux/Wayland, open the `defwidget@browser` widget → URL renders at the correct rect (no black screen, no full-window-flashing).
- [ ] Resize the pane via the splitter → BrowserView resizes to match (resize path uses the same set_size + set_position + layout pattern).
- [ ] Open multiple panes simultaneously → all render at their distinct rects.
- [ ] Close a pane → `OverlayController destroyed` fires; the BrowserView disappears.
- [ ] Right-click in the pane → page contextmenu fires.
- [ ] Right-click on host UI header → host context menu fires (no longer suppressed by full-window pane).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.625"
+version = "0.33.626"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.625"
+version = "0.33.626"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.625"
+version = "0.33.626"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.625"
+version = "0.33.626"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.625"
+version = "0.33.626"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
--- a/agentmux-cef/src/app.rs
+++ b/agentmux-cef/src/app.rs
@@ -17,12 +17,23 @@ use crate::state::AppState;
 // Window & BrowserView delegates (CEF Views framework)
 // ---------------------------------------------------------------------------
 
+// Linux/macOS only: when `Some((state, window_label))`, `on_window_created`
+// inserts the Window into `state.windows[window_label]` and
+// `on_window_destroyed` removes it. Browser-pane creation
+// (`browser_pane/creation_views.rs`) looks up the parent Window by label
+// to call `add_overlay_view` on. Without this, panes opened from a
+// non-main window were silently routed to the main window. Popup
+// delegates (DevTools etc.) pass `None` and don't register because they
+// shouldn't host user-facing panes.
+// (kept as a regular comment — wrap_window_delegate! doesn't accept
+// doc-comments on struct fields.)
 wrap_window_delegate! {
     pub struct AgentMuxWindowDelegate {
         browser_view: RefCell<Option<BrowserView>>,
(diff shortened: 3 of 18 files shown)


a5af:
@codex review

PR #691, opened 2026-05-04
feat(srv): E.4.B Phase 4 — pure layout tree helpers + README brain logo fix

AgentA-asaf:
## Summary

Phase 4 of srv E.4.B — ports the 11 pure tree-manipulation functions from `frontend/layout/lib/layoutTree.ts` to Rust. These are the building blocks the Phase 5 reducer arms will call. **No behavior change**; pure functions only.

**Spec:** \`docs/specs/srv-phase-e4b-formal-spec-2026-05-03.md\` Phase 4

### Functions ported

| Rust | TypeScript equivalent |
|---|---|
| \`insert_node\` | \`insertNode\` (BFS heuristic, max 5 children per node) |
| \`insert_node_at_index\` | \`insertNodeAtIndex\` |
| \`delete_node\` → collapses single-child parents | \`deleteNode\` |
| \`move_node\` | \`moveNode\` (apply-only; compute step stays TS) |
| \`swap_nodes\` | \`swapNode\` (sizes stay at slot) |
| \`resize_nodes\` | \`resizeNode\` (atomic: reject whole batch if any size invalid) |
| \`replace_node\` | \`replaceNode\` (preserves target's size) |
| \`split_horizontal\` / \`split_vertical\` | \`splitHorizontal\` / \`splitVertical\` (splice or wrap) |
| \`clear_tree_node\` | \`clearTree\` |
| \`find_node_by_id\`, \`find_parent_by_child_id\` | \`findNode\`, \`findParent\` |

### Error type

\`LayoutError { NodeNotFound, SelfSwap, RootCannotBeTarget, InvalidSize, InvalidIndexPath }\`

## Tests (23)

Ports the 30 tests from \`frontend/layout/tests/layoutTree.test.ts\` (PR #686 — the behavioral oracle) into Rust. All 23 pass.

**810 srv tests passing total** (was 785 before Phase 4 = +25). 3–4 pre-existing failures unchanged.

## Bonus: README brain logo fix

README.md line 2 referenced the old \`frontend/logos/agentmux-logo.svg\` which was renamed to \`agentmux-logo-brain-alternate.svg\` in PR #687 but the README wasn't updated. Now points to the brain-alternate SVG.

## Test plan

- [ ] \`cargo test -p agentmux-srv --bin agentmux-srv backend::layout\` — 23 pass
- [ ] \`cargo build -p agentmux-common -p agentmux-srv -p agentmux-cef -p agentmux-launcher\` — clean
[...]

PR #692, opened 2026-05-05
fix(layout): TS-oracle clamp + leaf-stop in insert_node_at_index (#691 follow-up)

a5af:
Follow-up to #691 — addresses a Codex P1 that landed at 09:24:52Z (5 min before #691 was merged at 09:29:36Z) and was missed during the merge poll due to a wrong wall-clock filter cutoff. Surfacing this PR per the global \"link follow-ups to the original\" rule so reagent re-evaluates context.

## The miss

In #691 round 7 (commit \`39e07f92\`), Codex posted an inline P1 on \`mod.rs:221\`:

> \`insert_node_at_index\` currently returns \`InvalidIndexPath\` as soon as an intermediate path segment is out of bounds, but the frontend oracle (\`frontend/layout/lib/layoutNode.ts::findInsertLocationFromIndexArr\`) clamps each segment and stops descent when a leaf is reached instead of failing. This means stale or over-deep \`index_arr\` values that still succeed in the existing reducer semantics will now be rejected in srv, causing command divergence under concurrent edits and violating the \"identical state transitions\" goal for Phase 4.

Verified legit by reading the TS source.

## TS oracle (frontend/layout/lib/layoutNode.ts:259-279)

\`\`\`ts
function normalizeIndex(index: number) {
    const childrenLength = node.children?.length ?? 1;
    const lastChildIndex = childrenLength - 1;
    if (index < 0) { ... }
    return Math.min(index, lastChildIndex);  // CLAMPS, never errors
}
const nextIndex = normalizeIndex(indexArr.shift());
if (indexArr.length == 0 || !node.children) {     // STOPS at leaf
    return { node, index: nextIndex };
}
return findInsertLocationFromIndexArr(node.children[nextIndex], indexArr);
\`\`\`

## Fix

Replace the early-error loop in \`insert_node_at_index\` with clamp-and-descend that mirrors the TS algorithm step-for-step:

| TS behavior | Rust before | Rust after |
|---|---|---|
| Clamp idx to [0, len-1] | Error if idx ≥ len | Clamp same as TS |
| Stop descent at leaf | Indexing into empty children would panic | Detect leaf, stop, use clamped_idx=0 |
| Empty path returns void | Error \`InvalidIndexPath\` | (unchanged) Error |

## Tests

Adds 3 regression tests in \`backend::layout::tests\`:
- \`insert_node_at_index_clamps_out_of_range_segment\` — single-level over-range
- \`insert_node_at_index_clamps_intermediate_segment\` — multi-level, over-range mid-path resolves to last child
- \`insert_node_at_index_stops_descent_at_leaf_with_extra_segments\` — leaf hit before path exhausted; verifies \`ensure_group_node\` promotion + ID stability

Local: \`cargo test backend::layout\` → **40 passed** (was 37 after #691).

## Why merge separately

Phase 5's reducer arms will start consuming \`insert_node_at_index\` and need the TS-equivalent semantics. Cleaner to land this as a focused follow-up PR than re-open #691.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.637"
+version = "0.33.638"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.637"
+version = "0.33.638"
 dependencies = [
  "serde",
  "serde_json",
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.637"
+version = "0.33.638"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.637"
+version = "0.33.638"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #697, opened 2026-05-05
feat(agent): 5-char date slug + collision resolution

a5af:
Replaces `YYYYMMDD-HHMMSS-mmm` (17 chars) with `MMDDh` (5 chars) and wires up proper end-to-end collision resolution.

## Frontend (instance-slug.ts)

`formatLocalStamp(d)` returns `MMDDh`:
- `MM` = 2-digit month
- `DD` = 2-digit day
- `h` = single base24 char (`0`-`9` then `a`-`n` for hours 10-23)

Examples: May 5 14:00 → `0505e`; Dec 31 23:59 → `1231n`. Year is omitted — per-version isolation already separates releases.

## Collision resolution (the launch path)

`AgentViewModel.launchForgeAgent` calls `WriteAgentConfigCommand` BEFORE `SetMetaCommand`, with a new `auto_allocate: bool` flag. The backend handler in `websocket.rs` routes through `allocate_agent_workdir` (in `app_api.rs`) which uses **atomic `std::fs::create_dir(leaf)` as the reservation mechanism** — two concurrent callers race on the OS `mkdir` syscall and one wins; the loser sees `AlreadyExists` and tries `<base>-1`, `-2`, ..., up to `-99`. No CLAUDE.md heuristic, no TOCTOU.

The RPC returns the final `working_dir`. Frontend captures it and uses it for `cmd:cwd` in the subsequent `SetMetaCommand`, so the controller spawns the CLI in the actually-created directory.

## auto_allocate gating (avoids regression for user paths)

The frontend sets `auto_allocate: true` only for paths it constructed itself:
- `overrides.instanceName` is given (modal-driven per-launch slug), OR
- `agent.working_directory` is empty (we filled in the legacy `~/.agentmux/agents/<slug>` default ourselves)

Any explicit persisted path — including paths under `~/.agentmux/` — flows through unchanged. A user pointing an agent at `~/projects/myrepo` (where they have an existing CLAUDE.md) sees the agent run in that exact directory.

## What `agent.open` does NOT do

`agent.open` is a separate RPC code path (frontend's `launchForgeAgent` doesn't go through it). It still calls `write_agent_config_files` directly without allocation — same as before this PR. A comment at `app_api.rs:346` documents this. If a future flow starts using `agent.open` for instance launches, it'll need its own allocation hookup. For now, no behavior change there.

## Build + tests

- agentmux-srv + agentmux-cef build clean
- agentmux-common: 45 tests pass
- agentmux-srv: 826 pass; 4 pre-existing failures unrelated to this PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.641"
+version = "0.33.642"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.641"
+version = "0.33.642"
 dependencies = [
  "dirs",
  "serde",
@@ -49,7 +49,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.641"
+version = "0.33.642"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -69,7 +69,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.641"
+version = "0.33.642"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.641"
+version = "0.33.642"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
--- a/agentmux-common/Cargo.toml
+++ b/agentmux-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-common"
-version = "0.33.641"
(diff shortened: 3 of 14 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
Round 5 review status — body now matches code:

- **reagent P1 (16:03Z)** — description-vs-code mismatch on `agent.open`: addressed by updating the PR body. New section ["What `agent.open` does NOT do"](#) explicitly documents that `agent.open` calls `write_agent_config_files` directly without `allocate_agent_workdir` (matching the comment at `app_api.rs:346`). No code change needed; the launch path (`launchForgeAgent` → `WriteAgentConfigCommand` → `allocate_agent_workdir`) is the only path with collision resolution and that's intentional.
- **reagent P2s (15:57Z)** — stale JSDoc in `instance-slug.ts` lines 12 and 72: already fixed in 2d024eac. The current docstring points at the server-side `allocate_agent_workdir` and says "atomic mkdir + `<base>-N` retry there is the source of truth" — no leftover "future allocator" or "not implemented" wording.
- **codex P2 (15:58Z on 63088df1)** — `autoAllocate` treating `~/.agentmux/*` as auto-generated: already fixed in 2d024eac. Heuristic is now narrowed to `Boolean(overrides?.instanceName) || (agent.working_directory ?? "").trim() === ""` (`agent-model.ts:391-393`). Any explicit persisted `working_directory`, including paths under `~/.agentmux/`, flows through unchanged.

Codex hasn't reviewed 2d024eac yet.

@codex review

PR #700, opened 2026-05-05
fix(srv): Windows UNC path-traversal bug in writeagentconfig

a5af:
Fixes a regression that blocked every Anthropic agent launch on the merged-main build (v0.33.643 portable).

## Symptom

Click Anthropic in the agent pane → nothing happens. Host log:

```
[fe] [agent] Failed to launch forge agent {"error":"Error: path traversal denied: .claude/commands/startup.md escapes working dir"}
```

## Root cause

`agentmux-srv/src/server/websocket.rs` writeagentconfig handler validated each config-file path with `canonicalize() + starts_with(canonical_base)`. On Windows, `Path::canonicalize` adds the `\\?\` UNC prefix. For brand-new agent dirs created by `allocate_agent_workdir`, neither the target file (`<base>/.claude/commands/startup.md`) nor any of its parent directories exist yet. The fallback chain ran out of canonicalize-able ancestors and silently returned the raw non-UNC path. Comparing `C:\Users\...\base\.claude\commands\startup.md` against `\\?\C:\Users\...\base` returned false → traversal denied.

Wasn't visible before #697 (slug rename) because returning users hit existing dirs that canonicalized cleanly. New MMDDh slug means every launch creates a fresh empty dir, hitting the bug every time.

## Fix

New helper `safe_join_within_base(base, relative)` in `agentmux-srv/src/backend/base.rs`:

- empty / absolute / rooted / `..` paths → error
- accepts `/` and `\` as separators
- silently skips `.` segments
- joins cleaned components onto base, no filesystem access

writeagentconfig replaces the canonicalize check with a single `safe_join_within_base` call. No `canonicalize()` anywhere in the validator — works identically on Windows / macOS / Linux because the logic is purely lexical.

## Tests

7 new unit tests in `backend::base::tests`:
- simple + nested + backslash-separator joins
- `.` segments skipped
- `..` rejected (incl. mid-path `a/../b`, trailing `a/b/..`)
- absolute paths rejected (`/etc/passwd`, `\Windows\System32`, `/foo`, plus `C:\Windows` on Windows)
- empty + effectively-empty rejected

All 7 pass. Full `cargo test --bin agentmux-srv`: 834 pass, 3 pre-existing unrelated failures (provider count, reactive handler, session archive — same set carried over from #697).

## Out of scope

`writeeditorfile` (websocket.rs:1006) shares the canonicalize+starts_with pattern, but its fallback path returns an error instead of falling through to a non-canonical comparison, so it fails closed (request errors out) instead of failing open. Worth a follow-up to make it consistent with the new helper, but not blocking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.643"
+version = "0.33.644"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.643"
+version = "0.33.644"
 dependencies = [
  "dirs",
  "serde",
@@ -49,7 +49,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.643"
+version = "0.33.644"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -69,7 +69,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.643"
+version = "0.33.644"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

a5af:
@codex review

PR #701, opened 2026-05-05
feat(agent): drop activity log auto-open + shell label

a5af:
## Why

ESC-to-cancel emits an `error`-level entry into the activity log as part of the normal cancel flow. The `ActivityLogPanel` was wired to auto-open whenever a new error-level entry arrived, so cancelling popped the whole shell log open every time. Annoying — user has to keep collapsing it.

Also dropping the redundant `shell` text label next to the chevron. The chevron alone reads fine; surrounding agent-pane chrome already gives enough context.

## What changed

`frontend/app/view/agent/components/ActivityLogPanel.tsx`:

1. Removed the `createEffect` that watched for new error entries and called `setIsOpen(true)`. Also removed the `lastSeenLength` watermark that paired with it. `createEffect` import dropped (unused after removal).
2. Removed the `<span class="agent-activity-log-label">shell</span>`. Chevron + preview + count remain.
3. Updated module JSDoc — removed the "Auto-opens when a new entry arrives with `level: 'error'`" sentence so docs match code.

The `agent-activity-log--has-error` / `agent-activity-log--has-warn` modifier classes still toggle on `mostRecent()?.level`, so the color cue stays. Just the auto-open behavior + the label text are gone.

## Verification

Tested live in dev mode (Vite HMR) before opening this PR:
- ESC-cancel emits the error entry; preview line updates with the new entry text; panel stays collapsed.
- Click chevron → expands and shows full history as before.
- Header row no longer contains the word "shell".

## Spec

`docs/specs/SPEC_AGENT_ACTIVITY_LOG_NO_AUTO_OPEN_2026_05_05.md`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.644"
+version = "0.33.645"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.644"
+version = "0.33.645"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #702, opened 2026-05-05
fix(host): orphan-instance reconciliation kills v0.33.643 zombie host

a5af:
Wires `Event::HostShouldQuit` to a real corrective action so the host actually exits when the last user-visible window closes — fixes the v0.33.643 zombie-host bug.

## Symptom

Closing every user-visible window of a running v0.33.643 instance left the host process alive for six hours in the user's session, holding warm-pool browsers + the launcher IPC. Launcher log captured the cascade attempt:

```
WRR-DRIFT [Warn] OrphanInstance label=Some("main") hwnd=None:
  Last user-visible window closed; host still alive (likely holding warm pool)
DRIFT Windows: host=2 mirror=0
DRIFT Pool:    host=0 mirror=2
saga_id=1 name=window_cleanup_cascade Done — emitting SagaCompleted
```

Saga ran cleanly, dispatched `ReapPanes` + `DrainPoolIfLast`, completed — and nothing exited.

## Root cause

Promoted `window-pool-*` browsers went orphan. They're still in the host's `browsers` map (no live HWND, no `on_before_close` ever ran for them — CEF crash, OS kill, missed callback) but the launcher's `shadow_window_meta` mirror has dropped them. The saga's `drain_pool_if_last` (`agentmux-cef/src/saga_dispatch.rs:354`) counts them as user windows:

```rust
let user_count = labels
    .iter()
    .filter(|k| !unpromoted.contains(k.as_str()) && !k.starts_with("browser-pane-"))
    .count();
```

…so `was_last == false` → saga emits `Event::PoolNotLast` → cascade aborts. Same gate in `client/mod.rs::on_before_close` fails for the same reason.

The launcher reducer DOES detect the orphan-instance state and emits `Event::HostShouldQuit`, but the host's handler at `launcher_ipc.rs:416` was log-only, with three documented prior attempts (`v0.33.491–v0.33.494`) at making it deliver UI-thread work that all failed (post_task drops, direct call UB, PostThreadMessage(WM_QUIT) ignored).

## Fix

`agentmux-cef/src/commands/orphan_reconcile.rs` — new module with two pieces:

- **Pure classifier** (`classify_orphan_labels`) — takes snapshot inputs (browser labels, unpromoted-pool set, shadow-meta keys) and returns the set of `window-pool-*` labels that are promoted-out AND no longer mirror-tracked. No CEF, fully unit-testable.
- **Snapshot-and-drop orchestrator** (`reconcile_and_drain`) — pulls the three snapshots from `AppState` under their respective locks, drops the locks, classifies, then dispatches a close to each orphan via the same PostMessageW / `cef::post_task` path the existing two-stage cascade uses (`client/mod.rs:792-851`).

Each close funnels back through `on_before_close`, which already has the Stage-2 hook to call `quit_message_loop()` when `browser_list` empties. The reconciler doesn't try to drive UI-thread work directly — that's exactly what the prior attempts got wrong; we hand the work to CEF's task queue and let `on_before_close` finish what it always knew how to do.

`Event::HostShouldQuit` handler in `launcher_ipc.rs` upgrades from log-only to:

```rust
tracing::warn!(target: "wrr", "[wrr] HostShouldQuit received — running orphan reconciler");
crate::commands::orphan_reconcile::reconcile_and_drain(state);
```

## Why this isn't a missing reducer

The reducer architecture is intact. Master spec §4.4 explicitly retains `browsers` / `window_pool` as scaffolding indefinitely (FFI/UI sync constraints). Master spec §9.1 calls cross-process saga dispatch a BLOCKER for the principled fix (saga-issued `IssueCmd::Host` actually delivering work). This PR is a bounded near-term fix that lives entirely inside the existing `HostShouldQuit` cross-process channel — no new pipe, no new dispatch primitive. When `SPEC_CROSS_PROCESS_DISPATCH_2026-05-01.md` lands, `DrainPoolIfLast` graduates from query to action and the reconciler can be folded into the saga directly.

## Tests

7 new unit tests in `commands::orphan_reconcile::tests` — all pass:

- `classify_no_orphans_returns_empty` — promoted user window correctly skipped.
- `classify_one_orphan_is_returned` — happy path.
- `classify_skips_unpromoted_pool_members` — warm pool members aren't orphan.
- `classify_skips_browser_pane_labels` — pane drain is a different cascade.
- `classify_skips_main_label` — only `window-pool-*` is reaped here.
- `classify_returns_multiple_orphans_in_input_order` — ordering preserved.
- `classify_handles_user_v0_33_643_scenario` — verbatim reproduction of the drift state from the launcher log (two `window-pool-*` labels both promoted, both shadow-dropped). Asserts the classifier identifies both.

The CEF-touching dispatch path (`post_close_or_fallback`) is the same proven pattern from `client/mod.rs:792-851` — exercised by every existing two-stage cascade in production. Manual smoke (reproduce the v0.33.643 case in a portable build) is the right validation for the FFI side; CEF integration tests aren't currently set up in this repo.

## Spec

`docs/specs/SPEC_HOST_ORPHAN_RECONCILIATION_2026_05_05.md` — full design doc with master-spec cross-references in §3, §8.

Also commits `docs/specs/SPEC_AGENT_ACTIVITY_LOG_NO_AUTO_OPEN_2026_05_05.md` — companion to PR #701 (the code change shipped, the spec file was untracked).


a5af:
@codex review

a5af:
@codex review

Round 2 (ea9a7a33) addresses your P1 — `tracked` is now the union of `shadow_window_meta` + host's local `window_meta`, so a freshly-promoted pool window in the WindowOpened-echo race window is treated as live and skipped.

a5af:
@codex review

Round 6 (c0d3a0f2) addresses your round-5 findings:

- **P1 (CEF threading):** All `Browser`/`BrowserHost` calls now run on the UI thread. `reconcile_and_drain` on the IPC reader thread only does state-snapshot + classification (no CEF), then posts `OrphanReconcileTask` via `cef::post_task(ThreadId::UI, ...)`. The actual HWND probe + close run inside the task. Same scheduler the rest of the host uses; avoids the v0.33.491-494 freeze hazard.
- **P2 (drain-on-live):** `BeginDrain` is now conditional on `live_user_count == 0` AFTER excluding the zombies being reaped. A stale `HostShouldQuit` racing with an active user session closes the zombies but leaves `quit_state` alone.

PostMessageW path removed — UI thread calls `host.close_browser(force=1)` directly, which works regardless of HWND state. Single close path.

a5af:
@codex review

Round 7 (de3ffc05) addresses your round-6 P1: in the drain branch, the reconciler now also appends every `window-pool-*` browser still in `browser_pairs` (excluding the zombies already in `to_close`) to the close list. This mirrors what the normal Stage-1 cascade in `on_before_close` would have done if the zombies hadn't inflated `user_browser_count` past its gate. Without it, killing only the zombies left `browser_list` non-empty so Stage 2 never fired `quit_message_loop`.

`deferred_live_candidates` is necessarily empty in the drain branch by construction — a freshly-promoted pool window is itself a live user window and would have bumped `live_user_count > 0`.

a5af:
@codex review

Round 8 (ac1af9d1) addresses your round-7 P1: `wrr::apply_hwnd_destroyed` now emits the OrphanInstance drift + `HostShouldQuit` pair when the destroy empties `state.windows` AND the host is still `Running`. Mirrors what `handle_report_window_closed` has been doing all along for the normal close path. Caller (`reducer/mod.rs::Command::ReportHwndDestroyed`) passes `host_running` so wrr stays out of the connection module's private API.

Two new launcher reducer tests:
- `wrr_orphan_destroy_emits_host_should_quit_when_last_window` — last-window crash fires drift + quit
- `wrr_orphan_destroy_does_not_emit_host_should_quit_when_other_windows_remain` — guards false positives

Test-infra `extract_version` had 11 missing `LayoutNode*` variants on main; fixed in passing so the new tests could run.

Spec §4.1 documents both trigger paths.

a5af:
@codex review

Round 10 (ddc93bcd) addresses your round-9 P1: when `browser.host()` returns None, `close_one` now dispatches `HostCommand::UnregisterBrowser { label }` instead of just logging. That removes the entry from `state.browsers` so `host=None` zombies don't sit there forever waiting for an `on_before_close` that will never fire.

Spec §5.7 renumbering also fixes reagent's P2 (duplicate §5.6).

a5af:
@codex review

Round 11 (5db38240) addresses your round-10 P1: after the close loop, if any candidate took the hostless-unregister branch AND `state.host_state.browsers` is now empty AND drain was entered (`live_user_count == 0`), the reconciler calls `cef::quit_message_loop` itself. Already on the UI thread (post_task brought us here, same thread Stage-2 calls it from), no new threading hazard. The drain-mode gate prevents a stale `HostShouldQuit` racing with a live session from terminating the app.

a5af:
@codex review

Round 12 (66453416) addresses your round-11 P1: `live_user_count` now uses `shadow_window_meta` membership as the authoritative live-user filter instead of "everything except unpromoted_pool". `ReportWindowOpened` only fires at promotion time, so labels in `pool.unpromoted` AND `pool.queue` are both absent from shadow → both correctly excluded. Promoted pool windows (which retain the `window-pool-*` prefix) ARE in shadow → correctly included. Single predicate covers both pool states.

`shadow_keys` is re-snapshotted on the UI thread so a freshly-promoted window whose `WindowOpened` echo arrived between IPC and UI is counted live correctly.

a5af:
@codex review

Round 13 (852ca781) addresses your round-12 P1: the orchestrator now classifies candidates three ways via HWND validity + `pool.queue` membership:

- dead HWND → zombie (Race C) → close
- live HWND AND in `pool.queue` → ready warm pool (Race D) → drain target
- live HWND AND NOT in `pool.queue` → freshly-promoted (Race B) → skip

`pool.queue` distinguishes ready inventory (queued for promotion) from already-promoted user windows (popped during promotion). The drain branch closes all `window-pool-*` not in `to_close` AND not in `freshly_promoted`, so in-flight promotions aren't killed.

`live_user_count` uses shadow membership alone now — zombies are inherently absent from shadow because `apply_hwnd_destroyed` prunes `state.windows`.

Spec §5.4 has the full three-way breakdown.

a5af:
@codex review

Round 14 (68215894) addresses your round-13 P1: drain is now gated on `safe_to_drain = live_user_count == 0 && freshly_promoted.is_empty()`. A freshly-promoted candidate (Race B) is a live user window whose `WindowOpened` echo hasn't returned — it's not in shadow so `live_user_count` misses it, but it would be killed by an unconditional drain. The same gate controls the hostless-orphan `quit_message_loop` drive so a freshly-promoted window isn't terminated mid-handshake.

a5af:
@codex review

Round 16 (37eff9e6) addresses your round-15 P1: dropped the early-return in `reconcile_and_drain` (IPC thread) when candidates is empty. The planner already handled the empty-candidates-with-unpromoted-pool case correctly — the existing `plan_unpromoted_pool_browsers_drain_via_warmpool_loop` test proved it. The bug was glue code short-circuiting the planner call. Now the planner runs unconditionally; if there's truly nothing to do, the UI-thread runner logs+exits.

Also note: round 15 added 14 state-machine integration tests for `plan_reconcile` covering the cross-product of (HwndStatus, shadow membership, pool.queue membership, label prefix). Each test names the codex round / spec race it covers. Future findings on this surface should map onto failing tests before the patch lands.

a5af:
@codex review

Round 17 (e58b4a2e) addresses your two round-16 P1s and reagent's P2 with one structural change: the planner now consumes a `pool_browser_status` map covering EVERY `window-pool-*` label in `state.browsers` (HWND-probed on the UI thread), not just classifier candidates. Each label is bucketed via HwndStatus × shadow / pool.queue / pool.unpromoted membership:

- Dead → CloseBrowser
- Hostless → UnregisterBrowser
- Live + shadow → promoted user window, leave alone
- Live + pool.queue → ready warm pool, drainable
- Live + pool.unpromoted → spawning pool slot, drainable
- Live + (none of the above) → freshly_promoted, blocks drain

This eliminates both gaps you flagged:
- "Reclassify pool labels on the UI thread" — re-snapshots done on UI thread now; IPC just posts the task.
- "Unregister hostless warm-pool labels during drain" — Hostless takes precedence over pool-state, so an unpromoted-pool slot that lost its host gets UnregisterBrowser, not CloseBrowser.

Reagent's dead-code `freshly_set` check is gone (the new structure makes it impossible to reach with a non-empty set).

15 planner integration tests cover every P1 from the round history. New for this round: `plan_hostless_unpromoted_pool_unregisters_not_closes`, `plan_promoted_pool_window_in_shadow_is_left_alone`, `plan_mixed_full_state_space`. 23/23 pass.

a5af:
@codex review

Round 18 (1f39e7dd) addresses both your round-17 P1s:

- **Probe non-pool top-levels:** dropped the `window-pool-` filter on HWND-probing. Now every non-pane top-level browser gets classified, so a regular `main`/`window-X` zombie gets reaped.
- **Hostless cleanup gated on drain:** `Hostless` action only fires when `safe_to_drain == true`. Outside drain, the hostless entry waits for the next reconcile to avoid bypassing `on_pool_window_destroyed`'s pool-reducer bookkeeping. Dead zombies still close unconditionally — `close_browser(force=1)` drives the host's own cleanup chain.

3 new tests verify the new gates. 26/26 pass.

a5af:
@codex review

Round 19 (d88d3cfb) addresses your round-18 P1: dropped the `state.browsers.is_empty()` precondition on the quit_message_loop drive. In mixed Hostless+Closable drain, closables are mid-close so `state.browsers` isn't empty at end of the reconcile loop, and Stage-2 quit later also doesn't fire because `client::browser_list` still has the Hostless entries (`UnregisterBrowser` only touches the reducer). Now drive quit unconditionally when `begin_drain && any_hostless`.

Also removed dead `classify_candidate_orphans` (reagent P2 — production goes through `plan_reconcile` directly; the legacy 8 classifier tests were superseded by 19 planner integration tests).

19/19 pass.

a5af:
@codex review

Round 20 (df50eefe) addresses both round-19 P2s:

- **Codex P2 (empty-closes drain):** when `HostShouldQuit` fires with `state.browsers` already empty AND drain requested, drive `quit_message_loop` directly. Stage-2 quit may be blocked by stale `client::browser_list` entries we can't see from here.
- **Reagent P2 (stale comment):** `drive_close_browser` now actually implements the claimed "fall through to UnregisterBrowser" — when host vanishes between classification and execution, dispatch `UnregisterBrowser` IF drain is active (mirrors the planner's Hostless gate from round-17 P1 #2). Returns a flag so the orchestrator includes the late-hostless transition in the post-loop quit drive.

19/19 tests pass.

a5af:
@codex review

Round 21 (1c226d20) addresses both round-20 P2s:

- **Codex P2 (pending creation race):** the empty-closes drain branch now also checks `state.host_state.lock().pending_window_creations.is_empty()`. If a window creation is in flight (`PendingWindowCreation` enqueued but `post_create_window` hasn't registered the browser yet), defer the quit drive. Honors the HostShouldQuit "re-check before quitting" advisory contract.
- **Reagent P2 (stale tags):** scrubbed all review-cycle references from code + test comments.

19/19 pass. Both findings have been P2 for two rounds — getting close.

a5af:
@codex review

Round 22 (e47f45c1) addresses your round-21 P1: `pending_creation_in_flight` is now factored into `safe_to_drain` itself, so the gate applies end-to-end (not just to the empty-closes path). Closing pool browsers when a creation is mid-flight could otherwise let Stage-2 quit fire before the new browser registers.

2 new tests verify pending-creation gating:
- `plan_pending_window_creation_blocks_drain`: ready warm pool + creation pending → no drain, no closes
- `plan_pending_creation_doesnt_block_zombie_close`: dead zombies still close (they don't depend on `safe_to_drain`)

21/21 pass.

a5af:
@codex review

Round 23 (9527e8ee) addresses your round-22 P1: Dead zombies are now also gated on `pending_creation_in_flight`. When a creation is pending and the zombie would be the last browser, close_browser's cleanup chain (`on_before_close` → BeginDrain → Stage-2 quit) would race the pending creation. Defer zombie reap until creation completes.

Mid-session zombie scenarios still reap correctly — when other live windows remain, on_before_close's user_count > 0 keeps the cleanup chain from cascading to drain.

Test `plan_pending_creation_blocks_zombie_close` codifies the new gate. 21/21 pass.

a5af:
@codex review

Round 24 (57b4d3d8) addresses your round-23 P1: pending check now filters on non-pool labels. `spawn_pool_window` also enqueues `PendingWindowCreation` (with `window-pool-*` labels), so my prior gate would suppress drain whenever a warm-pool spawn was in flight. Now only user-window pending creates (`main`, `window-N`, etc.) block drain.

21/21 pass.

a5af:
@codex review

Round 25 (0a93159b) addresses your round-24 P2: pending-create filter now also excludes `browser-pane-*` labels. Same reasoning as the pool exclusion — panes are excluded from user-window counts and Stage-2 quit elsewhere, so they shouldn't block drain either.

21/21 pass.

PR #624, opened 2026-05-01
fix(F1.B): restore tab on tear-off window-create failure

AgentA-asaf:
## Summary

Closes the orphan-workspace gap from:
- `docs/retro/saga-coordinator-location-analysis-2026-04-30.md` §6.4
- `docs/specs/PHASE_E_SAGAS_EXECUTION_PLAN_2026-04-30.md` §9b (F1.B)

Before this PR, if \`WorkspaceService.TearOffTab\` succeeded but both the warm-pool promote AND the cold-path \`open_window_at_position\` then failed, the tab was stuck in an invisible new workspace with no window pointing at it. The catch block just logged the error.

## What changed

In \`requestTearOff\`:
- Track whether the destination window was opened (\`destWindowOpened\` flag).
- In the catch:
  - If TearOffTab succeeded AND no window was opened → dispatch \`WorkspaceService.RestoreTornOffTab\` to put the tab back at its original index. The saga (shipped in #621) also cascade-deletes the empty new workspace.
  - If a window WAS opened (handshake-failure case) → skip restore. Restoring would empty + delete the new workspace and leave the open window dangling.

## Test plan

- [x] Frontend builds clean (\`npm run build\`).
- [x] Code review: catch path only fires on already-broken state; uses existing RestoreTornOffTab saga.
- [ ] Failure-injection smoke is hard from the existing harness — change is small and well-scoped.

## Phase E.5 + follow-ups status

After this PR merges:
- Phase E.5 ✅ (PRs #619-#622)
- F1.A ✅ (#623)
- F1.B ✅ (#this)

Remaining Phase E sub-phases (E.2c.5b, E.4, E.6, E.7) are scoped separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #704, opened 2026-05-06
feat(drag): try warm pool first on tear-off via shared helper

a5af:
## What

`CrossWindowDragMonitor.{win32,darwin,linux}.tsx` all called `api.openWindowAtPosition` directly when handling cross-window drag tear-off. The host has a pre-warmed pool (`tear_off_pool_promote`) designed for ~0ms first-paint flash, but it was only used by `tabbar.tsx::performTabTearOff` (the SC_MOVE handshake path). The HTML5-dragend path bypassed it entirely.

## Fix

Extract a shared `openTearOffWindow(api, newWsId, screenX, screenY)` helper to `frontend/app/drag/tear-off-pool-helper.ts`. It tries `tearOffPoolPromote` first and falls back to `openWindowAtPosition` on rejection. All three platform variants of `CrossWindowDragMonitor` now call it.

Same try-pool-first pattern `tabbar.tsx` uses inline. Kept inline there because it tracks `coldPathFailed` for F1.B orphan-cleanup safety (codex P1 round-3 #624).

## DRY

Before: three duplicate inline blocks (one per platform monitor).
After: one helper, three callers.

`tabbar.tsx` deliberately not migrated — its tear-off pipeline has additional state-tracking that doesn't fit the shared helper.

## Smoke

Tested on `task dev` with `agenta/tear-off-pool-path`:

- Pool spawned and enqueued (`pool_size=2`)
- Tab tear-off fired the new helper (`[fe] [dnd:cross] pool promote failed, cold-pathing` proves the pool was tried)
- Fallback to `openWindowAtPosition` happened gracefully
- No source-side renderer crash (was intermittent in the prior cold-path-only build)
- Cold-path's 150–300ms first-paint flash is still visible (because the fallback fires)

## What this PR doesn't fix

The pool itself is broken at the host level — pool windows enqueue with `renderer ready` but their HWND is null at promote time:

```
[pool] promoting pool window   label=window-pool-...
[pool] host window handle is null (state inconsistency)
[pool] pool window destroyed before promote — releasing semaphore + refilling
```

That's a separate host-side bug. This PR is the correct frontend wiring regardless — once the pool is fixed, both the flash AND the source-side crash risk go away with no further frontend changes.

## Spec

`docs/specs/SPEC_TEAR_OFF_POOL_PATH_2026_05_06.md` — full design + observed behavior + recommended follow-ups.

Related: `docs/specs/SPEC_TASK_DEV_LAUNCHER_GAPS_2026_05_06.md` — `task dev` doesn't run the launcher, so PR #702's reconciler is dead code in dev mode. Independent issue, also documented in this PR for traceability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.647"
+version = "0.33.648"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.647"
+version = "0.33.648"
 dependencies = [
  "dirs",
  "serde",
@@ -49,7 +49,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.647"
+version = "0.33.648"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -69,7 +69,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.647"
+version = "0.33.648"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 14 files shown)


a5af:
@codex review

a5af:
@codex review

Round 2 (4cd756be) addresses both round-1 P2s: type-only `getApi` import + package-lock.json synced to 0.33.648.

PR #705, opened 2026-05-06
fix(pool): cache HWND at register, use as fallback when CEF returns null

a5af:
## What

CEF's `BrowserHost::window_handle()` returns null on Views-based browsers once the page loads, even though the underlying Win32 window is alive. Every pool promote fails with `host window handle is null (state inconsistency)` and the frontend cold-paths every tear-off — exactly the waste the warm pool exists to prevent.

## Diagnosis (2026-05-06)

Instrumented `register_pool_window` and `mark_pool_window_renderer_ready` to log the HWND at each lifecycle stage. Captured:

```
07:14:17  HWND probe at register time:        valid hwnd=0x130892
07:15:39  HWND probe at renderer-ready time:  cef_status=null  cached_status=cached=0x130892 alive=true
```

The HWND **is alive** per Win32 `IsWindow`. CEF Views' `BrowserHost::window_handle()` is what's broken — it returns null after the page has loaded, but the underlying Win32 window is intact.

## Fix

Cache the HWND at `register_pool_window` time (when CEF reliably reports it) in a module-level `Mutex<HashMap<String, usize>>`. At `promote_pool_window`, fall back to the cache if `host.window_handle()` returns null. Verify the cached value is still a live OS window via `IsWindow` before using; if not (genuine OS-level destruction), refuse the promote so the frontend cold-paths cleanly.

Cache cleanup runs in `on_pool_window_destroyed` (which fires for any `window-pool-*` close, including post-promote user-window closes that retain the prefix).

The CEF call is still tried first. The moment a future CEF version returns the HWND consistently after page-load, the fallback stops being exercised — no further code change.

## Smoke (task dev, `agenta/pool-hwnd-cache`)

- 2 sequential tab tear-offs both **promoted from the pool**
- Zero `host window handle is null` errors
- Zero `pool exhausted` warnings
- Pool refilled after each promote
- No first-paint flash

This is the host-side companion to PR #704 — the frontend wiring (PR #704) tries the pool first; this PR makes the pool actually work.

## Spec

`docs/specs/SPEC_POOL_WINDOW_HWND_NULL_2026_05_06.md` — full diagnosis + design (§4 updated with diagnostic results).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.648"
+version = "0.33.649"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.648"
+version = "0.33.649"
 dependencies = [
  "dirs",
  "serde",
@@ -49,7 +49,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.648"
+version = "0.33.649"
 dependencies = [
  "agentmux-common",
  "chrono",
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

a5af:
@codex review

Lockfile synced to 0.33.649 (bf4a2b74).

PR #706, opened 2026-05-06
fix(launcher-bridge): forward events to PROMOTED pool windows

a5af:
## What

`agentmux-cef/src/launcher_event_bridge.rs::dispatch_to_renderers` filtered renderer destinations on label PREFIX:

```rust
if label.starts_with("window-pool-") || label.starts_with("browser-pane-") {
    continue;  // skip
}
```

But promoted pool windows (= the destination of a tab tear-off) **keep their `window-pool-*` prefix indefinitely** — only their entry in `state.unpromoted_pool_labels` is removed at promotion time. The prefix-only filter therefore excluded torn-off user windows from receiving launcher events.

## Symptom

Reproduced in v0.33.650 portable smoke (see linked logs):

1. Open AgentMux (main window).
2. Tear off 3 tabs (3 pool promotes).
3. Close 2 of those tear-off windows + main.
4. Open another new window.
5. Look at the InstancePanel in any torn-off window: it shows **1 window** when there are actually **2** alive.

Launcher log confirms the mirror is correct (5 `WindowOpened` / 3 `WindowClosed` = 2 alive). The drift lives in the renderer's per-window `launcher-event-reducer.ts` state, because no events reached it.

## Fix

Right discriminator is `state.unpromoted_pool_labels` (the canonical "is this a pre-promote pool slot or a real user window?" set). Snapshot it once per dispatch, skip only labels currently in the unpromoted set. Promoted labels — still prefixed `window-pool-*` but no longer in `unpromoted_pool_labels` — now receive events.

Browser-pane labels still skipped (they have no InstancePanel / launcher-event-reducer; the prefix check stays).

## Smoke

`cargo build -p agentmux-cef` clean. Will validate end-to-end with a fresh portable on the merged main.

## Wider impact

InstancePanel drift was the visible symptom because the user noticed the count mismatch. **Every launcher-event-reducer subscriber in a torn-off window was silently stale** — same root cause. Fixing this also fixes any other launcher-event-driven UI in those windows that nobody had noticed yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.653"
+version = "0.33.654"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.653"
+version = "0.33.654"
 dependencies = [
  "dirs",
  "serde",
@@ -49,7 +49,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.653"
+version = "0.33.654"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -69,7 +69,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.653"
+version = "0.33.654"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.650"
+version = "0.33.651"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.650"
+version = "0.33.651"
 dependencies = [
  "dirs",
  "serde",
@@ -49,7 +49,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.650"
+version = "0.33.651"
 dependencies = [
  "agentmux-common",
  "chrono",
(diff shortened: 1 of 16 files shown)


a5af:
@codex review

a5af:
@codex review

Lockfile synced to 0.33.651 (691a4b08). Reagent + codex P1/P2 addressed.

a5af:
@codex review

Round 3 (7c67bb9f) addresses your round-2 P2: `isInstanceLabel` no longer filters out `window-pool-*` labels. By the time a `window-pool-*` label reaches the frontend, the host-side gates have already excluded unpromoted ones — only promoted (real user windows) get through. Test inverted to match new expectation. 20/20 pass.

a5af:
@codex review

Round 4 (c0ef4050) addresses your `pool.queue` P2: added `state.pool_inventory_labels_snapshot()` (union of unpromoted + queue) and routed all four user-visibility gates through it (launcher_event_bridge, list_window_instances, list_windows, on_before_close). Generalized per the rule — grepped for sibling sites of the same unpromoted-only filter.

a5af:
@codex review

Round 5 (a46a2495) addresses your round-4 P2 race. Replaced `pool_inventory_labels_snapshot()` + `list_browser_labels()` (two separate `host_state` locks — racey against `promote_pool_window`) with a new atomic `state.user_visibility_snapshot()` that returns `(pool_inventory, browsers)` under ONE lock. All four user-visibility filter sites converted: `on_before_close` cascade gate, `dispatch_to_renderers`, `list_window_instances`, `list_windows`. Generalized the antipattern.

a5af:
@codex review

Round 6 (c5fca477) addresses reagent P1+P2 from round 5: two more two-lock-pool+browsers gates I missed in my round-5 generalization grep (I scanned for callers of `pool_inventory_labels_snapshot` but these still used the older `unpromoted_pool_labels_snapshot` helper). Fixed:
- `saga_dispatch::drain_pool_if_last` (P1) — sibling of `on_before_close` was_last gate; now uses atomic `user_visibility_snapshot()`
- `launcher_ipc::compute_and_report_host_counts` (P2) — drift report; now uses new atomic `host_counts_snapshot()` (returns (windows, unpromoted) under one lock; pool count stays unpromoted-only to match the launcher mirror semantics).

Round 5 was clean from your side; this is for the regression-check pass.

a5af:
@codex review

Round 7 (d80c9474) addresses your round-6 P2: my reasoning about launcher mirror semantics was inverted. The launcher state.pool retains queued labels (no event fires on host's unpromoted→queue transition), so the host's reported pool count must use pool_inventory.len() (unpromoted ∪ queue), not unpromoted.len().

Generalized — fixed two sites with the same antipattern:
- `host_counts_snapshot` (your finding): now returns pool_inventory.len()
- `window_pool::spawn_pool_window`'s post-add `report_host_pool_count`: same bug, now sums unpromoted+queue under one lock

PR #708, opened 2026-05-06
fix: prevent drift-storm renderer crash (tier 1 follow-up to #706)

a5af:
## Summary

Two independent fixes for the v0.33.655 smoke-test renderer crash. Tier 1 of the follow-up plan in [discussion #707](https://github.com/agentmuxai/agentmux/discussions/707) and [`docs/specs/ANALYSIS_DRIFT_STORM_RENDERER_CRASH_2026-05-06.md`](https://github.com/agentmuxai/agentmux/blob/agenta/drift-storm-tier1/docs/specs/ANALYSIS_DRIFT_STORM_RENDERER_CRASH_2026-05-06.md).

**Symptom:** 1 launcher event (`HwndDriftDetected version=13`) reached the renderer ~600× in 3 seconds → V8 stack exhaustion → `Crashpad_NotConnectedToHandler`. Triggered by the very first tear-off post-PR-#706.

### 1. Launcher reducer (Bug A — stops the storm at source)

`handle_report_pool_window_promoted` now flips `foregrounded_since_open=true` on the matching `WindowMirror`.

Pool windows are spawned hidden and never foregrounded. After promote, the host repositions the HWND via several `SetWindowPos` calls; some intermediate states have `visible=false`. Without resetting the WRR mirror's flag, each flicker re-fires `HiddenSinceOpen` drift indefinitely.

Idempotent if the mirror is absent (pre-existing contract preserved).

### 2. Frontend bridge (Bug B — defence-in-depth)

Per-key `(event, label, hwnd)` dedup with max-version-seen tracking, installed at the `__agentmux_launcher_event` entry. Bounded at 1024 keys (LRU by insertion order). Diag accessor: `launcherEventDedupStats()`.

`PerSourceTracker` already drops by global monotonic version, but the global counter advances past per-key duplicates, and fresh JS contexts (pool-window bootstrap) reset `lastVersion=0`. The bridge-entry guard catches both.

## Tests

- `agentmux-launcher` — 2 new tests, all 158 pass:
  - `promote_clears_foregrounded_since_open_so_post_promote_hide_does_not_drift`
  - `promote_for_label_without_window_mirror_is_a_no_op`
- `frontend/util/launcher-events.test.ts` — 8 new tests covering admit-once, suppress-rerun, version monotonicity, key independence, memory bound.

## Test plan

- [x] cargo test -p agentmux-launcher — 158 pass
- [x] vitest frontend/util/launcher-events.test.ts — 8 pass
- [x] vitest broader (441 tests across 20 files) — green
- [ ] Smoke: portable build + tear off 5 tabs in a row, no crash, no drift-event flood in `[fe]` log
- [ ] Verify `launcherEventDedupStats().suppressed` is non-zero in normal use (proves the guard is active and the upstream is still emitting duplicates → keep tier 2 on the to-do)

## Out of scope (deferred)

- Tier 2: host bridge de-dup + rate limit (only needed if tier-1 #2 doesn't catch enough — diag will tell)
- Tier 3: idempotency property tests across all four reducers; pool state-machine fuzz; cross-layer event volume budget
- Master spec additions: §8.14 idempotency contract, §9.8 Phase F.7

These are tracked in discussion #707; will append closure comments as each lands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.655"
+version = "0.33.656"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.655"
+version = "0.33.656"
 dependencies = [
  "dirs",
  "serde",
@@ -49,7 +49,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.655"
+version = "0.33.656"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -69,7 +69,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.655"
+version = "0.33.656"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.657"
+version = "0.33.658"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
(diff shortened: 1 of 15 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

Round 4 (7ac6a419) addresses your round-3 P2: `foregrounded_since_open` is now OR-ed with the prior mirror value on duplicate `ReportWindowOpened`, matching the WindowMirror contract that this flag is monotonic (false → true, never resets within a window's lifetime). New regression test `duplicate_open_for_promoted_label_preserves_foregrounded_since_open` runs the production tear-off sequence and asserts the flag survives a 2nd open.

161 launcher tests pass.

PR #717, opened 2026-05-07
feat(agent): show provider brand icon in pane header (#680)

AgentX-asaf:
## Summary

Wires `ProviderLogo` into the agent pane's frame header so each pane shows the brand icon for the LLM it's running — Claude / Codex / Gemini / Copilot / Kimi / etc. — matching what `AgentCard` already does in the picker.

Before: every agent pane header used the generic `sparkles` glyph.
After: the header (and its tab in the tab strip, since it's the same `block-frame-view-icon`) shows the provider's brand mark.

## Changes

- New `frontend/app/view/agent/components/AgentPaneIcon.tsx` — `buildAgentPaneIcon(provider)` helper that wraps `<ProviderLogo provider={...} size={16} />` in an `IconButtonDecl` with `noAction: true` so it renders as a decorative icon.
- `agent-model.ts` `viewIcon` now returns the JSX-bearing `IconButtonDecl` when the block has `agentProvider` meta. Falls back to `agentId` (the quick-launch path uses `agentId` as the provider key) and finally to the legacy `agentIcon` string.
- Version bump 0.33.678 -> 0.33.679.

The existing `getViewIconElem` branch in `blockframe.tsx` already handles `IconButtonDecl` via `<IconButton>`, and `IconButton.icon` already accepts `string | JSX.Element` — no changes needed there.

## Test plan

- [ ] Launch a Claude agent — pane header shows the orange Anthropic 5-bar mark
- [ ] Launch a Codex agent — pane header shows the OpenAI hex ring
- [ ] Launch a Gemini agent — pane header shows the 4-color G
- [ ] Switch back to the picker (no agent loaded) — header falls back to sparkles
- [ ] Pane header tab in a stacked layout shows the same icon

Closes #680
Related: #679 (ProviderLogo component)

PR #718, opened 2026-05-07
feat: refined table formatter for presentation layer

a5af:
## Summary

- **`TableBlock` component** (`frontend/app/element/table-block.tsx`) — overflow-x-auto scroll wrapper with border/rounded container, hover-revealed copy-as-CSV button with 1.5s checkmark feedback
- **Zebra striping + row hover** via `color-mix(in srgb, var(--hover-bg-color) 40%, transparent)` in SCSS — adapts to all 8 themes automatically without hardcoded white tints
- **Column alignment** (`frontend/app/element/rehype-align-to-class.ts`) — pre-sanitize rehype plugin converts remark-gfm's `align="..."` attribute to Tailwind `text-left/center/right` classes before `rehype-sanitize` strips it (markdown.tsx path)
- **Both render paths** (`streamdown.tsx` and `markdown.tsx`) wired to `TableBlock`; `th`/`td` components merge alignment classes via `cn()`

## Test plan

- [ ] Paste a GFM table with `|:---|:---:|---:|` alignment row into an agent pane — center/right columns should align correctly
- [ ] Resize pane narrow — wide table should scroll horizontally inside its container, not overflow
- [ ] Hover table — copy CSV button appears top-right; click it and paste into a text editor to verify comma-separated output with proper quoting
- [ ] Check zebra striping on odd rows and hover highlight across at least two themes (default + catppuccin)
- [ ] Verify `markdown.tsx` path (help view / forge) renders tables with the same styling


a5af:
@codex review

AgentA-asaf:
Took over to address reagent's 3 P1s. Pushed d0801a47 (v0.33.713):

1. **markdown.tsx props.class → props.className** — `hast-util-to-jsx-runtime` passes HAST className as `className` (matches the existing `Code` component pattern at line 158). Alignment classes were silently dropped.
2. **streamdown alignment wiring** — wired `rehypeAlignToClass` via streamdown's `rehypePlugins` prop, ordered between `raw`/`katex` and `sanitize`/`harden` (using exported `defaultRehypePlugins`) so `align`→`className` conversion happens BEFORE sanitize strips the raw `align` attribute.
3. **streamdown th/td className forward** — added `cn(..., props.className)` merge so alignment classes survive to the rendered cell.

Heads-up: bumped to 0.33.713 to clear the missing-bump P1, but PR #717 is also at 0.33.713 — whichever lands first wins and the other will need a re-bump.

Codex re-triggered.

a5af:
@codex review

AgentA-asaf:
Codex P2 addressed in 0fcc7692 (v0.33.714):

- Streamdown's default sanitize uses stock schema with no `className` allowlist on th/td, so alignment classes were being stripped before reaching the cell renderers.
- Replaced `defaultRehypePlugins.sanitize` with our own `rehype-sanitize` call that extends the schema to allow `text-(left|center|right)` on th/td (matches markdown.tsx).
- Exported `ALIGN_CLASS_REGEX` from `rehype-align-to-class.ts` so both pipelines share one source of truth.
- Collapsed a duplicate th/td schema entry in markdown.tsx (round-1 fix collided with a previously-landed schema entry).

Re-bumped to 0.33.714 (#717 took 0.33.713 in main).

a5af:
@codex review

AgentA-asaf:
Pushed b174977e for the P3 lockfile sync.

Re the P2 "Preserve KaTeX output when adding sanitization" — I think this finding is based on an incorrect read of streamdown's default pipeline. Verified:

```
$ node -e 'const fs=require("fs"); ... grep Wo={ ... '
Wo={raw:qo,katex:[we,{...}],sanitize:[$o,{}],harden:[harden,{...}]}
```

Streamdown's default `rehypePlugins` is `Object.values(Wo) = [raw, katex, sanitize, harden]` — sanitize **is** in the default pipeline (`$o` = `rehype-sanitize` from imports), called with `{}` (no schema = `defaultSchema` from `hast-util-sanitize`).

```
$ node -e "const {defaultSchema}=require('hast-util-sanitize'); console.log(defaultSchema.tagNames.includes('math'))"
false
```

So whether the user passes `defaultRehypePlugins.sanitize` (default) or my replacement (`defaultSchema` + th/td className allowlist), KaTeX MathML output is stripped identically — the schemas are equivalent on every tag/attribute *except* the th/td className allowlist I added. My change cannot have regressed math rendering that was already working, because the same sanitize was already running.

Codex's claim that the default pipeline is `raw -> katex -> harden` (no sanitize) doesn't match the package source. Treating this finding as a false positive.

AgentA-asaf:
Round 4 — pivoted to a simpler approach in 6eb2f0a3 (v0.33.715):

- Keep streamdown's full default pipeline (`raw → katex → sanitize → harden`) UNTOUCHED.
- Append `rehypeAlignToClass` AFTER `sanitize`. The `align` attribute is in `hast-util-sanitize`'s default global attribute allowlist (`attributes['*']` includes "align"), so it survives the sanitize step. Our plugin then converts it to className, which the th/td renderers already pick up via `cn()`.

This drops 13 lines, preserves KaTeX behavior identical to default streamdown, and addresses the codex P2 / reagent P1 flagged on rounds 2-3 in one commit.

I still maintain my round-3 read of the streamdown source was correct (sanitize IS in the default pipeline), but this approach is cleaner regardless and cuts off the disagreement.

a5af:
@codex review

AgentA-asaf:
Superseded by **#754** — same final tree state (your final `83f3af01` commit) cherry-picked into a single squash on top of current `main`. This PR was DIRTY against the version bumps from the issue-#678 cycle and #752 (agent pane state machine); landing the squash + re-bumping was cleaner than walking through 14 commits of merges and version-bump conflicts.

All P0/P1/P2 findings from your review rounds are preserved per your final `LGTM` (commented at 2026-05-08T00:11:35Z, same sha). Authorship credited via `Co-Authored-By: a5af` in the squash commit.

Recommend closing this in favor of #754.

AgentA-asaf:
Superseded by #754 (merged 2026-05-08 as commit a92dc6dc), which shipped the same TableBlock + rehype-align-to-class work as a clean squash. Closing this stale branch.

PR #695, opened 2026-05-05
feat(common): RuntimeMode + DataPaths primitives (#694 follow-up)

a5af:
Foundation PR for the data-dir unification specced in #694. Adds the \`RuntimeMode\` + \`DataPaths\` primitives in \`agentmux-common\` **without yet switching any consumer**. Launcher / host / sidecar / srv stay on legacy paths until a follow-up PR migrates them.

Why split: the spec PR (#694) is doc-only. This adds the actual primitives the spec promised. The consumer-switching PR can then be reviewed against merged primitives instead of being one giant diff.

## Adds

| File | What |
|---|---|
| \`agentmux-common/src/runtime_mode.rs\` | \`RuntimeMode { Installed, Portable, Dev { branch } }\` enum, detection priority, env round-trip, branch slug. 7 tests. |
| \`agentmux-common/src/data_paths.rs\` | \`DataPaths\` struct (8 paths + mode), \`resolve()\`, \`ensure_dirs()\`, \`to_env_vars()\`/\`from_env()\` round-trip. 6 tests. |
| \`agentmux-common/src/lib.rs\` | Re-exports. |
| \`agentmux-common/Cargo.toml\` | \`dirs = \"5\"\` runtime dep, \`tempfile\` dev-dep. |

## Detection priority (`RuntimeMode::current`)

1. \`AGENTMUX_RUNTIME_MODE\` env override (testing / CI).
2. \`<exe-dir>/runtime/\` exists → \`Portable\`.
3. \`exe_dir\` is under \`dist/cef-dev/\`, \`target/debug/\`, or \`target/release/\` → \`Dev { branch }\` (branch via \`git rev-parse --abbrev-ref HEAD\`, slugified, falls back to \`\"default\"\`).
4. \`AGENTMUX_DEV_BRANCH\` env set → \`Dev { branch: <slugified> }\`.
5. Default → \`Installed\`.

## Layout (per spec §3.1)

\`\`\`
~/.agentmux/
├── shared/                       account-wide cookies/credentials
├── versions/<v>/                 installed + portable
│   ├── data/, config/, logs/, cef-cache/, agents/
│   └── runtime/                  lock + IPC, single set per version
└── dev/<branch>/                 per-branch dev isolation
    └── (same children)
\`\`\`

## Env vars exported by the launcher (consumed via \`from_env()\` later)

\`\`\`
AGENTMUX_INSTANCE_DIR
AGENTMUX_DATA_DIR
AGENTMUX_CONFIG_DIR
[...]

PR #719, opened 2026-05-07
fix(dev): bypass inherited AGENTMUX_* env when exe is a dev build

AgentX-asaf:
## Summary

When `task dev` is launched from inside an AgentMux terminal pane — a routine workflow for agents iterating on the codebase from inside the host they're embedded in — the new dev binary inherits the parent's `AGENTMUX_*` env vars. `DataPaths::from_env()` returns `Some` with the stale paths, the dev build resolves its data_dir to the parent's version-isolated directory, and CEF's process-singleton lock on that directory routes every open request back to the running parent. The user never sees the dev code execute.

This silently defeats the version-isolation guarantee from PR #695/#696.

### Fix

When the binary lives in a known dev-build output directory (`dist/cef-dev/`, `target/debug/`, `target/release/`), bypass `from_env()` and re-derive paths from the exe location. Path-based detection is authoritative for dev builds; installed/portable continue to trust the launcher's env injection.

### API additions

- `agentmux_common::is_dev_build_exe(exe_dir)` - public path-only check
- `RuntimeMode::current_path_only(exe_dir)` - mirrors `current()` minus the env-override step

### Call sites updated

- `agentmux-cef/src/main.rs` (boot path)
- `agentmux-cef/src/sidecar.rs` (`spawn_backend` / `restart_backend`)
- `agentmux-launcher/src/data_dir.rs` (`resolve_paths`) - parity for the rare case where the launcher itself runs from a dev build

### Spec

Full problem analysis + behavior matrix at `docs/specs/dev-build-env-isolation.md`.

## Test plan

- [ ] Close any running AgentMux, then `task dev` from a clean shell - data_dir should be `~/.agentmux/dev/<branch>/`
- [ ] With AgentMux 0.33.679 portable running, `task dev` from inside one of its panes - data_dir should be `~/.agentmux/dev/<branch>/` (NOT the running portable's `versions/0.33.679/`)
- [ ] Installed build still routes to `~/.agentmux/versions/<v>/` (unchanged behavior)

Closes the silent collision documented in #717's task-dev verification step.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.710"
+version = "0.33.711"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.710"
+version = "0.33.711"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.710"
+version = "0.33.711"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.710"
+version = "0.33.711"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.707"
+version = "0.33.683"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.707"
+version = "0.33.683"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.707"
+version = "0.33.683"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.707"
+version = "0.33.683"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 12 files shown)


a5af:
@codex review

a5af:
Rebased on main (0.33.702 → 0.33.703) and addressed both P1s from the second review:

**P1 — `main.rs` `port_file_dir` (line 253):** Added `is_dev_build_exe` guard. Dev builds now use `data_dir` directly instead of reading `AGENTMUX_DATA_DIR` from env, so the dev ipc-port file is written to the dev data directory and can't overwrite the parent instance's port:token.

**P1 — `main.rs` `use_launcher_endpoints` (line 300):** Added `is_dev_build_exe` guard at the call site. Dev builds skip `use_launcher_endpoints` entirely (returns `None`) and always fall through to `spawn_backend`. A dev build launched from inside a parent pane can no longer inherit `AGENTMUX_BACKEND_WS` and connect to the parent's srv.

Both fixes use `host_exe_dir` and `agentmux_common::is_dev_build_exe` already computed earlier in `main.rs` — no new imports or changes to `sidecar.rs`.

AgentA-asaf:
Triage 2026-05-07: solid intent — the bug is real and the fix is well-motivated. But the current implementation has **two missed env-read sites** that defeat the guard, plus the branch needs a rebase against current main (0.33.710).

### Confirmed P1s

1. **Version downgrade** — branch package.json is at 0.33.703 vs main 0.33.710. Rebase + bump.
2. **`main.rs:238`** — `std::env::var_os("AGENTMUX_DATA_DIR")` reads directly for `port_file_dir`. When task dev inherits the parent's env, dev writes its `ipc-port` file to the **parent's** data directory, overwriting the running parent's port:token mapping. Bypasses the guard added via `DataPaths::from_env()`.
3. **`sidecar.rs:43`** — `use_launcher_endpoints` reads `AGENTMUX_BACKEND_WS` directly. A dev binary inheriting this env consumes the parent's backend endpoints instead of spawning its own srv.

### Suggested fix shape

Centralize the dev-build env gate so future env reads can't silently bypass it. Something like:

```rust
// In agentmux_common (or wherever is_dev_build_exe lives):
pub fn get_env_unless_dev_build(key: &str, exe_dir: &Path) -> Option<String> {
    if is_dev_build_exe(exe_dir) { None } else { std::env::var(key).ok() }
}
```

Then `main.rs:238` and `sidecar.rs::use_launcher_endpoints` consume this instead of `std::env::var_*` directly. The single helper means future env reads can grep for `std::env::var" agentmux"` and audit they all use the gate.

### Recommended action

Rebase against main, fix the two missed env-read sites (probably via the centralized helper above), bump version. The spec/intent is good — it's well worth landing once the holes are closed. Happy to take this over if you'd prefer.

a5af:
@codex review

AgentA-asaf:
Took this over per request. Just pushed:

1. **Rebased onto main** (was at 0.33.703 vs main 0.33.710); resolved version-file conflicts
2. **Audited all `AGENTMUX_*` env reads** in agentmux-cef and found two MORE isolation holes the prior commits missed:
   - `launcher_ipc::connect_to_launcher` reads `AGENTMUX_LAUNCHER_PIPE` directly — dev build inheriting it would talk to the parent's launcher pipe
   - `srv_ipc::connect_to_srv` reads `AGENTMUX_SRV_PIPE_PATH` directly — same shape, parent's srv pipe
3. **Gated both at main.rs** (where `host_exe_dir` is in scope), matching the pattern used for `use_launcher_endpoints`
4. Bumped to 0.33.711

Codex re-triggered. The robustness audit should be tight now.

PR #721, opened 2026-05-07
fix(launcher): cap HiddenSinceOpen drift at one event per window

a5af:
## Summary

Drift-storm regression caught during smoke testing of v0.33.685: tear off a tab, then click "New Window" from the hamburger menu, and the renderer crashes ~1 second later with `Crashpad_NotConnectedToHandler`. **170 identical `HiddenSinceOpen` events for the same label in 1 second** → V8 stack exhaustion.

PR #708 thought it had closed this class. Its fix (set `foregrounded_since_open=true` on `ReportPoolWindowPromoted` via `just_promoted_labels`) only covered the pool→user transition. The "New Window" path (`open_new_window`) emits `ReportWindowOpened` directly with no preceding promote, so it inherited the original storm.

## Root cause

```
1. open_new_window → ReportWindowOpened
2. fresh mirror has foregrounded_since_open=false
3. host repositions HWND via several SetWindowPos calls
4. each intermediate visible=false re-fires HiddenSinceOpen
5. 170 events in 1s → V8 stack out → crash
```

## Fix

Cap `HiddenSinceOpen` drift at **at most one event per window per session**.

- Add `hidden_since_open_emitted: bool` to `WindowMirror`
- Gate the emission in `apply_hwnd_visibility_changed` with `&& !mirror.hidden_since_open_emitted`
- Set the flag true on first emission (monotonic; never reset)

Subscribers still get the drift signal — just not the storm.

## Why a cap, not "default `foregrounded_since_open=true` for FullInstance"

The earlier-considered alternative (extending PR #708's "is-promoted" semantic to all FullInstance opens) would have suppressed BOTH `HiddenSinceOpen` AND `OffMonitor` corrective for fresh windows. `OffMonitor` corrective uses `foregrounded_since_open` as its "trust the user's placement" marker and IS useful for catching placement bugs that orphan a window. The cap-based fix preserves the corrective.

## Tests

- `hidden_since_open_drift_fires_at_most_once_per_window` — 100 hide/show oscillations after the first hide produce zero additional drift events; flag is set after the first emission.
- All 165 pre-existing launcher tests still pass (166 total).

## Test plan

- [x] cargo test -p agentmux-launcher — 166/166 pass
- [x] cargo check — clean
- [ ] Smoke: tear off a tab, then "New Window" from hamburger — renderer should NOT crash

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.686"
+version = "0.33.687"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.686"
+version = "0.33.687"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.686"
+version = "0.33.687"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.686"
+version = "0.33.687"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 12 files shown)


a5af:
@codex review

a5af:
@codex review

Round 2 (f4aa66d9) addresses your round-1 P2: `hidden_since_open_emitted` is now OR'd with the prior mirror's value on duplicate `ReportWindowOpened`, matching the existing `foregrounded_since_open` monotonic discipline. New regression test `hidden_since_open_cap_survives_duplicate_open` exercises the duplicate-open path explicitly. 167 launcher tests pass.

PR #722, opened 2026-05-07
fix(cef): host-bridge dedup — Phase F.7 cap on launcher event fan-out

a5af:
## Summary

Third-layer cap on the drift-storm renderer crash class. Smoke testing of v0.33.688 surfaced a **164× amplification of a single launcher event**:

```
launcher: HiddenSinceOpen v=78 (emitted once — PR #721 cap works)
  ↓
host bridge: dispatch_to_renderers(v=78) (called once)
  ↓
renderer: logs [fe] [launcher-event] drift v=78 × 164
  ↓
V8 stack exhausted → renderer crash
```

Renderer-side guards (PR #708 dedup + `PerSourceTracker` version-stale drop) are installed but miss this case — suspect multi-V8-context fan-out (each browser has its own context; tracker state isn't shared) or a fresh renderer post-crash hitting an empty cache.

[`ANALYSIS_DRIFT_STORM_RENDERER_CRASH_2026-05-06.md`](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/ANALYSIS_DRIFT_STORM_RENDERER_CRASH_2026-05-06.md) §4.4 / master spec §9.8 flagged this as **Phase F.7 — host bridge resilience**. Implementing now since the launcher cap alone isn't sufficient.

## Defense-in-depth, three layers

| Layer | Where | What | Shipped |
|---|---|---|---|
| 1 | launcher reducer | At-most-one `HiddenSinceOpen` per window via `hidden_since_open_emitted` | PR #721 |
| 2 | renderer guard + tracker | `shouldDispatchLauncherEvent` per-key cache + `PerSourceTracker.deliver` version-stale drop | PR #708 / #709 |
| 3 | **this PR** | host bridge per-`(kind,label,hwnd,version)` cap | new |

## Implementation

- New `AppState.launcher_bridge_dedup: Mutex<HashMap<String, u64>>` — key `"{event_kind}|{label}|{hwnd}"`, value max version dispatched.
- `dispatch_to_renderers` calls `should_dispatch(state, event)` before serializing or iterating browsers. Skip entirely if version `<=` cached.
- `dedup_key(event)` extracts key + version per Event variant. Window-tracking variants use label+hwnd; conservative catchall for the rest.
- Cache bounded at 4096 keys with FIFO eviction. Re-arrival of an evicted key bypasses the host gate but the renderer guard still catches it.

## Tests

9 new tests in `launcher_event_bridge::bridge_dedup_tests`:
- `first_dispatch_passes_gate`
[...]

PR #724, opened 2026-05-07
fix(frontend): wrap recordDispatch in untrack() — root cause of renderer storm crashes

AgentA-asaf:
## Summary

Found the actual root cause of the renderer storm crashes that PRs #708 / #721 / #722 were trying to address. Those PRs added host-side dedup/fan-out caps (treating symptoms — per-event flood reaching the renderer); the disease is a **per-effect SolidJS feedback loop** that turns one event into thousands of console.warn calls.

## Diagnosis

Built a diagnostic build (v0.33.695) with counters at:
- host bridge `should_dispatch` and `dispatch_to_renderers` fan-out
- frontend `__agentmux_launcher_event` bridge call counter
- frontend `createEffect` re-run counter

Repro on v0.33.695, opening a browser widget pane:
- **Launcher**: emitted `pool_window_added v=7` once (cap held)
- **Host bridge**: dispatched once, `fanout=2` (correct: 2 user-visible browsers)
- **Frontend bridge**: 1 call per V8 context (dedup correct)
- **SolidJS effect**: re-ran **3230+ times in 1.2 sec** for that one event
- Each re-run -> drift-detected audit -> `console.warn` -> `fe_log_structured` RPC -> renderer V8 stack exhaustion -> Crashpad

## Root cause

`recordDispatch` (`command-source.ts:74`) reads `recordsAtom()` and writes `setRecordsAtom(next)` in the same call. The launcher-event reducer's `createEffect` calls `dispatch -> recordDispatch`. The bare `recordsAtom()` getter, called inside the effect's reactive context, registers `recordsAtom` as a tracked dep of the effect. The subsequent `setRecordsAtom` write re-fires the effect -> dispatch again -> recordDispatch again -> infinite loop.

## Fix

Wrap the read+write pair in `untrack()`. `recordDispatch` is a passive ring-buffer log and should never establish reactive deps.

## Test plan

- [x] New regression test in `command-source.test.ts` — drives a `createEffect` that calls `recordDispatch`, asserts effect runs exactly once per explicit trigger
- [x] Test fails on the un-fixed code (confirmed locally)
- [x] All 6 command-source tests pass with the fix
- [ ] Smoke test the post-merge portable: open browser widget pane, tear off tab + open new window from hamburger — neither should crash the renderer

## Notes for review

The host-side caps from #708/#721/#722 are still useful defense-in-depth. This PR doesn't remove them; it eliminates the actual amplifier downstream of where they sit.

Supersedes #723 (closed — opened from wrong identity).

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.695"
+version = "0.33.696"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.695"
+version = "0.33.696"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.695"
+version = "0.33.696"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.695"
+version = "0.33.696"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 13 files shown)


a5af:
@codex review

AgentA-asaf:
@codex review

a5af:
@codex review

PR #738, opened 2026-05-08
feat(browser-pane): Phase 1 diag instrumentation + host log discovery fix

AgentA-asaf:
Phase 1 of the diagnostic-first reducer migration roadmap (`docs/specs/browser-pane-reducer-roadmap.md`).

**No behavior change.** Only logs + a log-discovery fix.

## Browser pane diag instrumentation

Every state write, IPC event in/out, public method, and focus transition emits a tagged log line:

```
[browser-pane:diag][<blockId7>] <event> <fields>
```

Greppable via `muxlog host '\[browser-pane:diag\]'`. The format is stable so Phase 2 can baseline observed sequences before Phase 3 touches state.

**Coverage:**
- `browser-model.ts`: ctor, 3× sub-registered, nav-state recv, clicked recv, state-write per atom, navigate, goBack/goForward/reload, onLoad/onError, giveFocus, dispose, post-close-event-dropped.
- `browser-view.tsx`: view-mount, **sync** (urlAtom→addressBar with focused/willUpdate flags — the suspected typing-clobber site), createPane, paneCreated, input-submit, input-focus, input-blur, view-unmount.
- Host (`browser_pane/callbacks.rs`, `hwnd.rs`): emit-nav-state (both url_only=true and on_loading_state_change paths), emit-clicked.

## Host-log discovery fix

Recent versions wrote logs to version-specific instance dirs (`<root>/versions/<v>/logs/`) but the pointer in each instance dir held only the basename, and `memory_heartbeat::refresh_log_pointer` hardcoded `~/.agentmux/logs/` instead of using `AGENTMUX_LOG_DIR`. Result: `muxlog host` from outside the instance dir couldn't resolve the log.

Fix:
- `main::init_logging` now also writes a global pointer at `<root>/logs/<pointer_name>` containing the **absolute path** of the current log. Works uniformly for portable / installed / dev modes.
- `memory_heartbeat::refresh_log_pointer` reads `AGENTMUX_LOG_DIR` (set by `data_paths.rs`) and mirrors the same dual-pointer pattern.

Without this, the Phase 1 logs above can't be collected from outside the instance dir.

## Test plan

- [x] `cargo build -p agentmux-cef` clean
- [x] `npx tsc --noEmit` clean for new files
- [ ] Smoke build → open browser pane → `muxlog host '\[browser-pane:diag\]'` shows the full nav sequence
- [ ] Phase 2 (separate PR/retro) baselines observed log sequences for the three scenarios

PR #739, opened 2026-05-08
fix(browser-pane): multi-window focus bounce — guard synthetic re-fires + pass window_label

AgentA-asaf:
## Summary

Kills a 200 Hz `onFocus` / `onBlur` bounce on the browser pane address bar in multi-window setups. Caught by Phase 1's diag logs (PR #738).

## Diagnosis (from logs)

Clicking the address bar in window B with another window open produced:

```
input-focus prev-active=input.browser-address-bar same-target=true relatedTarget=null
input-blur                                         relatedTarget=null now-active=input.browser-address-bar
input-focus prev-active=input.browser-address-bar same-target=true relatedTarget=null
[... ~200 Hz forever ...]
```

`prev-active` and `now-active` always pointed back to the same input. `relatedTarget` was always `null`. The DOM input never actually lost focus — Chromium was firing **synthetic** focus/blur events in response to Win32 HWND focus shifts caused by `main_window_focus` IPC misrouting to window 1's main HWND (the host's `main_window_focus` handler defaults to `window_label="main"` when none is provided — `ipc.rs:424-428` documents this back-compat).

## Fix (two parts)

1. Pass `window_label` to `main_window_focus` so it reclaims focus to the **current** window. URLSearchParams parse hoisted to component scope so `createPane` and the focus IPC share one value.
2. Skip the IPC on synthetic re-fires:
   ```ts
   const wasAlreadyFocused = document.activeElement === e.currentTarget;
   const realTransition = !!e.relatedTarget;
   if (wasAlreadyFocused && !realTransition) return;
   ```
   Real user focus has either a `relatedTarget` (transition from another element) or arrives at an element that wasn't already focused. Synthetic re-fires have neither.

Plus richer `onFocus`/`onBlur` diag logs (`prev-active`, `now-active`, `relatedTarget`, `wasAlreadyFocused`, `realTransition`) so the next regression is diagnosable from logs alone.

## Test plan

- [x] Reproduces the bounce in multi-window setup pre-fix
- [x] Bounce gone post-fix (verified live via diag stream)
- [x] Single-window flow still works (was unaffected before; still is)
- [ ] Reagent + codex pass

## Out of scope

A separate **tear-off** bug surfaced during testing: tearing a tab to a new window leaves the old HWND in window 1 (causing Z-order flicker) and the new HWND in window 2 invisible. That's host-side `SetParent` lifecycle, not focus, and goes on its own branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.717"
+version = "0.33.718"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.717"
+version = "0.33.718"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.717"
+version = "0.33.718"
 dependencies = [
  "agentmux-common",
  "chrono",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #751, opened 2026-05-08
feat(identity): launch modal bundle pickers + spawn-time env injection

AgentA-asaf:
## Summary

PR-F.3 of issue #678. Closes the loop on the Identity + Memory bundle model: the launch modal now lets users pick a bundle of each, and the spawning CLI subprocess receives the resolved credentials as env vars.

**Supersedes PR #745** (Phase 2's per-provider picker). The new model uses single FK columns (identity_id, memory_id) on db_agent_instances instead of the v7 JSON column #745 introduced; the resolver shape is similar but reads via the v7 db_identity_bindings junction. With this PR landed, #745 should be closed.

### New module: \`agentmux-srv/src/identity/\`

- **\`resolver.rs\`** — \`inject_identity_env(wstore, block_id, &mut env_vars)\`:
  - Looks up the active AgentInstance for the block.
  - Reads identity_id; empty / \"blank\" → noop (ambient creds).
  - Reads bundle bindings for the identity.
  - For each binding: fetches the account, resolves SecretRef, looks up provider→env-var matrix, writes each var into env_vars.
  - All failures are logged and skipped — the spawn never aborts.

### Provider matrix

| Provider | Env vars |
|---|---|
| \`github\` | \`GITHUB_TOKEN\`, \`GH_TOKEN\` |
| \`anthropic\` | \`ANTHROPIC_API_KEY\` |
| \`openai\` | \`OPENAI_API_KEY\` |
| \`kimi\` | \`MOONSHOT_API_KEY\` |
| \`aws\` | \`AWS_ACCESS_KEY_ID\` |

\`SecretRef\` backends supported: \`Env\` (read from srv's environment), \`PlaintextDev\` (literal value, dev-only). \`SecretsManager\` returns \`Unsupported\` pending Phase 3's encrypted vault.

### Wired into both spawn paths

- \`AgentInputCommand\` in \`websocket.rs\` (per-message send to a running agent CLI)
- \`AgentSendCommand\` in \`app_api.rs\` (App API tier-1 send)

Both call \`inject_identity_env\` after reading the static \`cmd:env\` block from block meta and before passing \`env_vars\` to the controller.

### Schema/RPC

- \`CommandCreateAgentInstanceData\` gains \`identity_id\` + \`memory_id\` (both optional, default empty = blank singleton).
- forge_handlers.rs CreateAgentInstance handler persists them onto the new db_agent_instances row instead of hardcoding empty strings.
[...]

PR #742, opened 2026-05-08
feat: agent pane state machine refinements (#728)

a5af:
Closes 5 of the 6 gaps from #728. Pure additive changes — no redesign of the existing two-slice reducer architecture.

## Summary

| Gap | What | Files |
|-----|------|-------|
| 1 | Init phase — `InitStart`/`InitReady`/`InitFailed` gate `TurnStart` while history loads | `agent-pane-state/*`, `useHistoryPagination.ts`, `state.ts`, `agent-view.tsx` |
| 2 | Pending message expiry — 30 s timeout removes ghost rows | `agent-pane-state/*`, `useAgentCommands.ts` |
| 3 | Stuck-stream watchdog — 5 s tick emits `stream-stuck` after 45 s of silence | `agent-pane-state/*`, `useAgentStream.ts` |
| 4 | `nodeIdSet` in reducer — replaces per-mount scan in `useAgentStream` | `agent-document/*`, `agent-document-store.ts`, `useAgentStream.ts` |
| 6 | Injectable `truncateGraceMs` — reducer opts param | `agent-document/*`, `agent-document-store.ts` |
| 5 | **Deferred** — DocumentState migration | follow-up (see below) |

## Why deferred gap 5

`collapsedNodes` / `pinnedNodes` / `filter` / `scrollPosition` migration into the reducer requires refactoring `AgentDocumentView.tsx` (537 LoC) plus restructuring `documentStateAtom` in `state.ts`. The per-blockId atoms already persist for the block lifetime via `createAgentAtoms`, so the visible \"reset on pane navigation\" symptom doesn't actually occur today. The audit-trail benefit alone doesn't justify the surface area in this PR — tracking as a follow-up issue.

## Architecture fit

Every change is additive within the existing pattern:
- New variants on the discriminated `Command`/`Event` union types
- New `case` arms in the `update(state, command, nowMs?, opts?)` switch
- New `AgentPaneProjections.initPhase?` setter — optional so non-agent panes keep working
- One new optional reducer param (`opts.truncateGraceMs`) — defaults preserve existing behavior

The reducers stay pure: `nowMs` is now the second arg in `agent-pane-state/reducer` too, used by `bumpEvent` for tool/token transitions and by `PendingMessageExpired` for `ageMs`. No `Date.now()` inside the reducer body.

## Invariant changes

- `TurnStart` now suppressed when `initPhase === \"loading\"` (fail-open on `error`)
[...]

Issue #728, opened 2026-05-07
refine: agent pane state machine — init phase, pending timeout, watchdog, DocumentState unification


a5af:
## Background

The agent pane is already reducer-based with two pure slices (`agent-document/`, `agent-pane-state/`), each following the `update(state, command) → {state, events}` contract. The foundation is sound. This issue tracks six targeted gaps identified during analysis.

## Gaps

1. **Init phase not covered** — `InitState`/`InitQuestion` types exist but neither reducer models the loading/ready/error lifecycle. The pane has no explicit state for "history still loading."
2. **No acceptance timeout** — a `pending[]` entry lingers indefinitely if `agent-message-accepted` never arrives (e.g. RPC failure after send).
3. **Stuck-stream recovery** — the 1500 ms stop fallback only fires after `RequestStop`. There is no watchdog for streams that stay active with zero events for an extended period.
4. **`nodeIdSet` rebuilt on remount** — `useAgentStream.ts` rebuilds the dedup index by scanning `nodes[]` on each mount; duplicates can slip through the gap before the scan completes.
5. **`DocumentState` outside reducers** — `collapsedNodes`, `pinnedNodes`, `filter`, `scrollPosition` are local signals, invisible to the audit trail and reset on pane navigation.
6. **Hardcoded `TRUNCATE_GRACE_MS`** — not injectable; difficult to test or tune under load.

### `agent-pane-state`

**State:**
```ts
initPhase: "loading" | "ready" | "error";
initError?: string;
lastEventMs: number | null;
streamStuckMs: number | null;
```

**New commands:**
```ts
| { type: "InitStart" }
| { type: "InitReady" }
| { type: "InitFailed"; reason: string }
| { type: "PendingMessageExpired"; localId: string }
| { type: "StreamWatchdogTick"; nowMs: number }
```

**New events:**
```ts
| { type: "pending-message-expired"; localId: string; queuedAt: number; ageMs: number }
| { type: "stream-stuck"; idleSinceMs: number; thresholdMs: number }
```

**Behavior:**
- `TurnStart` is suppressed when `initPhase !== "ready"` (same pattern as streaming guard)
- `PendingMessageExpired` removes the entry by `localId`; no-op if not found
- `StreamWatchdogTick` emits `stream-stuck` when `Date.now() - lastEventMs >= STUCK_THRESHOLD_MS` (default 45 000 ms)
[...]

PR #752, opened 2026-05-08
feat: agent pane state machine refinements (#728) — review fixes + rebase

AgentA-asaf:
## Summary

**Supersedes #742.** Cherry-picks the two functional commits from #742 onto the latest \`main\` and addresses the open review findings from reagent + codex. #742 itself has been DIRTY since other PRs landed.

Closes 5 of the 6 gaps from #728 (same as #742) — pure additive changes, no redesign of the existing two-slice reducer architecture.

## What carried over from #742

- **Gap 1 — Init phase**: \`InitStart\` / \`InitReady\` / \`InitFailed\` gate \`TurnStart\` while history loads.
- **Gap 2 — Pending message expiry**: 30s timeout removes ghost rows.
- **Gap 3 — Stuck-stream watchdog**: 5s tick checks stream idle time; 45s threshold emits \`stream-stuck\` event.
- **Gap 4 — Stop flow**: \`RequestStop\` / \`StopFailed\` reducer transitions match the actual backend handshake.
- **Gap 5 — Tool tracking**: \`ToolStart\` / \`ToolEnd\` reducer transitions for live tool-call display.

(Gap 6 — full DocumentState in reducer — deferred per the spec.)

## Review fixes (this PR's net-new changes)

Three P2s from reagent + codex on the prior cycle of #742:

- **Codex P2** (\`useAgentCommands.ts:227\`, *Block the send path while init is loading*): the reducer's \`TurnStart\` handler suppresses \`turnActive\` when \`initPhase === \"loading\"\`, but \`commands.sendMessage\` still enqueued the pending row and called \`AgentInputCommand\`. If the backend accepted before \`InitReady\`, the accepted-event \`TurnStart\` was suppressed too — UI showed no active turn while the agent was actually processing. Now \`sendMessage\` reads the pane snapshot before queuing and bails early when init is still loading.
- **Codex P2** (\`useHistoryPagination.ts:127\`, *Guard init completion after pane unmount*): if the pane closes while either of the two history RPCs is in flight, the post-await \`dispatchPane\` would throw because \`AgentPresentationView\` already unregistered the slot. Added a \`mounted\` closure flag and early-return after each await in both the success and error paths.
- **Reagent P2** (\`SPEC_AGENT_PANE_STATE_MACHINE_2026_05_07.md:137\`): spec listed \`streamStuckMs: number | null\` as a state field for gap 3 but the implementation chose event-only. Updated the spec to reflect that with a note explaining the choice; both the §3 listing and the summary at line 230 are corrected.

The fourth finding (codex P2 on \`useAgentCommands.ts:257\` — pending expiry timer outliving the pane) was already addressed by AgentY's follow-up commit on #742; carried over verbatim.

## Why a fresh PR

#742 was DIRTY against the latest main (conflicts on package.json from intervening version bumps in PR-E and the F.0/F.1/F.2 cycle). Cherry-picking AgentY's two functional commits onto a clean branch off latest main was simpler than resolving the mid-rebase conflicts. The original commits' authorship is preserved (Author: AgentY-asaf) on commits b45b0597 + 88a57b00.

## Bumps

To 0.33.724 — same as the in-flight #751 since both branched off the same main; whichever merges second will need to re-bump.

## Test plan

- [x] cherry-picks applied cleanly onto latest main
- [x] tsc --noEmit clean for useAgentCommands + useHistoryPagination
- [x] All 4 review findings addressed (3 in net-new commit, 1 inherited)
- [ ] Reagent re-review
- [ ] Codex re-review
- [ ] Local task dev: send during initial history load → no message queued, log line visible

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.723"
+version = "0.33.724"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.723"
+version = "0.33.724"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.723"
+version = "0.33.724"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.723"
+version = "0.33.724"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.723"
+version = "0.33.724"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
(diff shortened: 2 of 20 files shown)


a5af:
@codex review

PR #759, opened 2026-05-09
hotfix: re-export TITLE_FALLBACK from browser-pane-state barrel (#758 build break)

AgentA-asaf:
## Summary

PR #758 added the `TITLE_FALLBACK` constant to `frontend/app/store/browser-pane-state/types.ts` and imported it in `frontend/app/view/browser/browser-model.ts` via the barrel (`@/app/store/browser-pane-state`), but **never re-exported it from `index.ts`**. Vitest resolves the subpath directly so unit tests passed; rollup goes through the barrel and the production build broke.

\`\`\`
error during build:
frontend/app/view/browser/browser-model.ts (14:4): "TITLE_FALLBACK" is not exported by "frontend/app/store/browser-pane-state/index.ts"
\`\`\`

## Fix

One line — add `TITLE_FALLBACK` to the named exports in `frontend/app/store/browser-pane-state/index.ts`.

## Lesson

Per memory \`feedback_verify_before_push\`: bug-fix PRs should build + smoke locally before opening the PR. I caught this on a smoke-test build right after merging #758 — should have caught it before the merge by running \`task build:frontend\` between the lockfile sync and the push. Adding to my discipline going forward.

## Test plan

- [x] \`task build:frontend\` succeeds (verified locally before pushing).
- [x] Vitest still 443/443 (the import was already wired in tests via subpath).
- [ ] Full \`task package\` succeeds (running in parallel with this PR review for the smoke-test build).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #758, opened 2026-05-08
slice #9 phase 3e: title cell through browser-pane reducer (faviconUrl deferred to post-3d)

AgentA-asaf:
## Summary

Phase 3e of slice #9, per [`docs/specs/browser-pane-reducer-roadmap.md`](docs/specs/browser-pane-reducer-roadmap.md). Migrates the `title` cell onto the reducer with the catalog's empty-string-fallback rule built in.

Follow-up to #757 (Phase 3c).

### What's in scope

- New reducer cell: `title: string` (initial value `TITLE_FALLBACK = "Browser"`).
- New command: `TitleChanged { title: string }`.
- **Empty/whitespace input is folded to `TITLE_FALLBACK` in the reducer**, not at the view layer. The catalog (`browser-pane-state-catalog.md` row 1.titleAtom) explicitly defines this rule; centralizing it in the reducer means the view never has to know about the empty case.
- `BrowserViewModel._dispatch` now projects `state.title` onto the existing `_title` SolidJS signal.

### What's NOT in scope (intentional)

- **`faviconUrl` is deferred.** Per the catalog (row 1.faviconUrlAtom) it's *derived* from `urlAtom` (\`\${origin}/favicon.ico\`). The `url` cell hasn't migrated yet — that's Phase 3d, the explicitly-tagged danger cell that needs its own dedicated PR with extra observability. Adding a faviconUrl reducer state today would either (a) duplicate the derivation logic outside the reducer, or (b) couple to a not-yet-migrated cell. Cleanest is to ship favicon as a derived projection alongside or after §3d. Roadmap header §3e was \"title + faviconUrl\" but the body section only specifies title; Phase 6 wires the host events that produce both.
- **Host-side `browser-pane-title-change` IPC subscription** — not yet wired. That's Phase 6 of the roadmap (\"Title + favicon (the original feature)\"). Until then, `state.title` stays at the fallback, and `TitleChanged` has no producer. The reducer cell is in place ready for Phase 6 to dispatch.

### Behavior parity

- `viewName` memo at line 172 reads `titleAtom`. Same accessor, same value (\"Browser\" pre-PR; \"Browser\" post-PR until Phase 6 lights up the producer).
- `setTitle` is now strictly internal (only called by `_dispatch`). No external callers existed — confirmed by grep.

### Phase-1 diag log shape preserved

[...]

PR #760, opened 2026-05-09
fix: browser-pane address-bar/page-DOM focus routing on first click

AgentA-asaf:
## Summary

Two related fixes for the address-bar ↔ page-DOM focus routing in browser panes. The user reported two complementary bugs in 0.33.731 and 0.33.732:

1. **0.33.731 / before:** click address bar directly → typing doesn't reach it. Workaround: click anywhere else in the header first, then the address bar works.
2. **0.33.732 / after first fix:** address bar works, but click into google search bar → typing still goes to address bar.

Both have the same underlying cause: the pane HWND has Win32 keyboard focus, and the main webview HWND can't reclaim OS focus from a single click on its own DOM elements.

### Fix 1 — `<input>` onMouseDown calls `main_window_focus`

When the pane HWND has OS focus, clicking the address bar `<input>` updates DOM focus (React `onFocus` fires) but does not move OS keyboard focus. Buttons in the same nav bar happen to work because CEF/Chromium internally calls `SetFocus(parent)` for the focusable native control on click; `<input>` doesn't get the same treatment.

Adding `onMouseDown` that fires `main_window_focus` IPC at the start of the click — *before* React's focus event — moves OS focus to the main window early enough that the subsequent keystrokes route to React.

### Fix 2 — pane-click blurs stale main-input focus

When the user clicks the embedded page (e.g. google search), the pane HWND captures the click at Win32 level and `browser-pane-clicked` IPC fires. The frontend handler calls `refocusNode(blockId)` which eventually triggers `model.giveFocus()`. But `document.activeElement` is still the address bar `<input>` (CEF's pane click never reached the React DOM), so `giveFocus` sees `isMainInput=true` and tells the host to keep OS focus on the main window — bouncing it back from the pane HWND we just gave it.

The diag log captured this exact sequence: `emit-clicked` → `clicked recv` → `giveFocus` → typing logged into `addressBar=...sdfsdsdf` while `focused=true`.

The fix: in the `browser-pane-clicked` handler, if `document.activeElement` is a main-window input (matching the giveFocus pattern), explicitly `.blur()` it before calling `refocusNode`. Then `giveFocus` resolves to "not a main input" and OS focus stays on the pane HWND. Adds a `pane-click blur active=...` diag line for traceability.

The check matches the giveFocus shape verbatim (per the catalog rule that giveFocus must be preserved unchanged).

## Test plan

- [x] Frontend build green (`task build:frontend` succeeds; this caught the missing `TITLE_FALLBACK` barrel export pre-merge of #758 — separate hotfix #759).
- [x] All 443 frontend tests still pass.
- [x] Manual diag-log verification on 0.33.732: confirmed the bug pattern (`focused=true` while typing landed in `addressBar`).
- [ ] Manual smoke on 0.33.733: address bar works on first click AND clicking google search routes typing into the page DOM.
- [ ] Stress harness `tools/tests/pane-focus-stress.ps1`: needs a path-resolution patch to find dev-mode logs under `~/.agentmux/dev/<branch>/logs/` (currently only searches `~/.agentmux/logs/`). The harness's auto-computed click coords also didn't land on the address-bar input in our run (no `input-focus` events fired). Both are harness limitations, not blockers for this fix; tracking separately.

## Files

- `frontend/app/view/browser/browser-view.tsx` — `<input>` onMouseDown
- `frontend/app/view/browser/browser-model.ts` — pane-click blur in IPC handler
- Version bumps to 0.33.732 (mousedown) and 0.33.733 (blur)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.731"
+version = "0.33.732"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.731"
+version = "0.33.732"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.731"
+version = "0.33.732"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.731"
+version = "0.33.732"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.732"
+version = "0.33.733"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

a5af:
@codex review

PR #763, opened 2026-05-09
slice #9: faviconUrl derived from url.origin + remove temporary urlAtom-read diag

AgentA-asaf:
## Summary

The **last cell** of slice #9. With this PR, every per-pane data cell from the catalog is reducer-backed and the cell-by-cell migration is complete. Phase 4 (slot store + \`recordDispatch\` audit) is the next milestone.

Two changes that ride together:

### 1. \`faviconUrl\` as a derived projection of \`url\`

Per the catalog, \`faviconUrl\` is *not* an independent input — it's \`\${origin}/favicon.ico\`, derived from whatever \`url\` is set to. So no \`FaviconChanged\` command. Instead:

- New pure helper \`deriveFaviconUrl(url: string): string\` — the URL constructor parses; on success returns \`\${origin}/favicon.ico\`, on failure / empty URL / origin === \"null\" (about:blank, file://) returns \`\"\"\`.
- The three reducer transitions that write \`url\` (\`Navigate\`, \`UrlConfirmed\`, \`UrlCleared\`) now also write \`faviconUrl\` atomically. The two cells stay in lockstep by construction — no separate event ordering to get wrong.
- New \`_favicon\` Solid signal + \`faviconUrlAtom\` accessor on \`BrowserViewModel\`. Projected from \`state.faviconUrl\` in \`_dispatch\` only when the value changes (same pattern as every other cell).

The catalog's row 1.faviconUrlAtom rule (\"empty → header globe\") is preserved at the view layer; the reducer enforces \"empty when URL is empty/unparseable\" so the view never has to know about parse errors.

### 2. Remove the temporary \`urlAtom-read\` diag instrumentation

Per the roadmap §3d, the \`urlAtom\` accessor was wrapped to log every consumer read for the duration of Phase 3d so we could verify the migration didn't change call patterns. The roadmap explicitly called this instrumentation **temporary — remove after Phase 3d ships and the log baseline is recorded.**

Baseline is recorded (the migration shipped clean in #761). The wrapper goes away; \`urlAtom\` is back to the bare signal getter (zero overhead, unchanged reactive shape — the wrapper preserved Solid tracking but the getter form is simpler).

## Test plan

- [x] +5 \`deriveFaviconUrl\` tests: empty input, normal https, port-bearing http, unparseable URL, about:blank fallback.
[...]

PR #761, opened 2026-05-09
slice #9 phase 3d: url cell (the danger cell) through reducer

AgentA-asaf:
## Summary

Phase 3d of slice #9 — the **danger cell** that broke PR #737. Migrates `url` onto the reducer with extra observability per [`docs/specs/browser-pane-reducer-roadmap.md`](docs/specs/browser-pane-reducer-roadmap.md) §3d.

Follow-up to #758 (Phase 3e title).

### What's in scope

Three new commands plus an extension to `Navigate`:

- **`Navigate { url }`** (existing) — extended to also set `state.url` atomically with the `loading=true, error=null` flip. Same intent as before (user/programmatic nav with a known target); now the URL is part of the same reducer transition.
- **`UrlConfirmed { url }`** — new. For host's `browser-pane-nav-state` IPC reporting the post-redirect final URL. Updates `state.url` only; does NOT touch loading/error/history (those transition via the LoadFinished/HistoryUpdated dispatches that accompany the same IPC event).
- **`UrlCleared`** — new. For `reload()`'s force-reload pattern (clear-then-restore across one frame). Distinct from `UrlConfirmed { url: "" }` so the audit ring distinguishes the reload pattern from a host-confirmed empty URL.

All three are idempotent on identical values, gated post-close, and emit named events for traceability.

### Six downstream consumers — all preserved

The catalog called out 6 read sites for `urlAtom` that must keep identical semantics:

1. `browser-model.ts reload()` — read current URL to clear+restore
2. `browser-view.tsx view-mount diag` — initial log
3. `browser-view.tsx initial addressBar` — `createSignal(model.urlAtom() \|\| "")`
4. `browser-view.tsx sync createEffect` — **the typing-clobber path**, gated on `focused` (the fix kimi added in PR #484; PR #737 inadvertently broke this)
5. `browser-view.tsx onMount createPane` — fires `browser_pane_create` with the URL
6. `browser-view.tsx placeholder \`<Show>\` gate` — empty-state UI shown when no URL is set

The migration preserves all six. The signal is still a SolidJS `createSignal`; the reducer projects `state.url` onto it via `_dispatch`. Reactive dependency tracking (Solid's `createEffect` machinery) sees the same signal read it always did.

### Extra observability (temporary)

Per roadmap §3d: every read of `urlAtom` now emits a diag log line:

\`\`\`
[browser-pane:diag][${blockPrefix}] urlAtom-read value="${url}"
\`\`\`

[...]

PR #757, opened 2026-05-08
slice #9 phase 3c: canGoBack/canGoForward through browser-pane reducer

AgentA-asaf:
## Summary

Phase 3c of slice #9, per [`docs/specs/browser-pane-reducer-roadmap.md`](docs/specs/browser-pane-reducer-roadmap.md). Migrates the two history-navigability cells onto the reducer.

Follow-up to #756 (Phase 3a + 3b).

### Cells migrated

| Cell | Owner | Source |
|---|---|---|
| `canGoBack` | reducer | CEF's `on_loading_state_change_pane` → host emit `browser-pane-nav-state` (with `url_only=false`) |
| `canGoForward` | reducer | same |

Both cells co-update in a single transition via the new `HistoryUpdated` command. Either field is optional (the host can emit a partial nav-state) — an omitted field leaves its cell alone. Identical-to-state values produce no event (idempotent).

### Cells NOT migrated (intentional)

- `url` (3d) — the danger cell that PR #737 broke. Reserved for its own dedicated PR with extra observability per roadmap §3d.
- `title` + `faviconUrl` (3e) — follow-up.
- `addressBar` typing buffer — must stay component-local per the catalog.

### Behavior parity

The pre-PR call site:

```ts
if (!payload.url_only) {
    if (payload.can_go_back !== undefined) this.setCanGoBack(payload.can_go_back);
    if (payload.can_go_forward !== undefined) this.setCanGoForward(payload.can_go_forward);
}
```

becomes:

```ts
if (!payload.url_only && (payload.can_go_back !== undefined || payload.can_go_forward !== undefined)) {
    this._dispatch({
        type: "HistoryUpdated",
        canGoBack: payload.can_go_back,
        canGoForward: payload.can_go_forward,
    }, "nav-state");
}
```

The `url_only` gate (kimi's race-condition fix) is preserved verbatim. SolidJS signals are written through only when the reducer's value differs from prior state — same observable behavior as before.

### Phase-1 diag log shape preserved

`state-write key=canGoBack value=true src=nav-state` and `state-write key=canGoForward value=… src=nav-state` continue to appear in `muxlog host '\[browser-pane:diag\]'`. No grep recipe changes.

## Test plan

- [x] +6 new HistoryUpdated tests — co-update, omitted-field, idempotency on both fields and on a single supplied field, isolation from loading/error cells, post-close gating.
- [x] Cross-product invariant test (every command from every reachable state) extended to include HistoryUpdated.
[...]

PR #756, opened 2026-05-08
slice #9 phase 3a+3b: browser-pane reducer (closed/loading/error)

AgentA-asaf:
## Summary

Introduce slice #9 of the master frontend reducer stack — the **browser-pane state machine** — at minimal viable scope per [`docs/specs/browser-pane-reducer-roadmap.md`](docs/specs/browser-pane-reducer-roadmap.md).

This is **Phase 3a + 3b only**. The roadmap was written precisely because PR #737 tried to migrate the whole slice in one shot and regressed the address-bar typing path. The strict-sequence rule is one cell per PR; the slot store + `recordDispatch` audit don't ship until every cell is reducer-backed (Phase 4).

### Cells migrated this PR

| Cell | Owner | Commands | Notes |
|---|---|---|---|
| `closed` (3a) | reducer | `Disposed` | Replaces the bare `_closed = false` flag. Idempotent. Post-close commands emit `post-close-command-dropped` for diagnostics instead of mutating state. |
| `loading` (3b) | reducer | `Navigate`, `LoadStarted`, `LoadFinished`, `LoadFailed` | Was `_loading` signal; now a projection of reducer state. |
| `error` (3b) | reducer | `Navigate`, `LoadFinished`, `LoadFailed` | Was `_error` signal; now a projection. Mutually exclusive with `loading` — the reducer enforces this invariant across the full state cross-product. |

### Cells NOT migrated (intentional, see roadmap)

- `url` — the danger cell that PR #737 broke. Reserved for its own dedicated PR with extra observability.
- `title`, `faviconUrl`, `canGoBack`, `canGoForward` — follow-up PRs.
- `addressBar` (component-local typing buffer) — must stay component-local per [`browser-pane-state-catalog.md`](docs/specs/browser-pane-state-catalog.md).

### Wiring

`BrowserViewModel` introduces a private `_paneState: BrowserPaneState` and a `_dispatch(cmd, src)` helper that:

1. Runs the pure reducer.
2. Projects state changes onto the existing SolidJS signals (only when value changed — preserves reactive parity).
3. Preserves Phase-1 diag log shape (`state-write key=loading value=true src=…`) so existing `muxlog host '\[browser-pane:diag\]'` recipes still work.

The `closed` getter now reads through `_paneState.closed`. All existing call sites (`if (this.closed) return;`) keep working.

### `giveFocus()` preservation

[...]

PR #764, opened 2026-05-09
slice #9 phase 4: slot store + recordDispatch audit + PaneClicked through reducer

AgentA-asaf:
## Summary

**Slice #9 is now complete.** Phase 4 externalizes browser-pane state ownership from \`BrowserViewModel\` into \`frontend/app/store/browser-pane-state-store.ts\` and routes every dispatch through the global \`recordDispatch\` audit ring. Pattern matches slice #4 (\`agent-pane-state-store.ts\`) verbatim — same slot lifecycle, same projection-on-change discipline, same no-silent-drops rule.

Plus: the focus-routing fix from PR #760 graduates from inline-IPC-handler to a \`PaneClicked\` reducer command (the \"treat as Phase 4 task\" framing). The blur-stale-main-input + refocusNode side-effect now flows through the reducer's event sink, making the click visible in the audit ring.

### What landed

**\`frontend/app/store/browser-pane-state-store.ts\`** — new module.
- \`registerPane(blockId, projections)\` — synchronous registration; throws if the model dispatches before this runs.
- \`dispatch(blockId, cmd, source)\` — runs the pure reducer, projects state changes onto signal setters, fires events through the global event sink, records the dispatch in the \`recordDispatch\` ring.
- \`unregisterPane(blockId)\` — drops the slot. Called in \`dispose()\` after the \`Disposed\` command's projections run.
- \`setEventSink(sink)\` — global; the model installs once via \`installEventSinkOnce()\` to handle \`pane-clicked\`.
- \`snapshot(blockId)\` — diagnostics + \`closed\` getter.

**\`PaneClicked\` reducer command** — new. Doesn't change reducer state (DOM/Win32 focus is owned outside the reducer per catalog rule §3) but emits a \`pane-clicked\` event. The model's IPC handler for \`browser-pane-clicked\` now dispatches \`PaneClicked\` instead of running the inline blur+refocus.

**Event sink** in \`browser-model.ts\` handles \`pane-clicked\`:
- Blur \`document.activeElement\` if it's a main-window \`<input>\` / \`<textarea>\` (the dummy-focus exclusion preserved verbatim from PR #760).
- Call \`refocusNode(blockId)\` so layout marks the block as focused.

**\`BrowserViewModel\` refactor:**
- \`_paneState\` field + internal \`_dispatch\` projector method — gone. State lives in the slot store.
- \`closed\` getter reads \`bpSnapshot(blockId)?.closed ?? true\` (defaults true for unregistered, so a stale post-dispose reference reports closed).
- Per-cell projection callbacks emit the existing \`state-write key=X value=Y\` diag logs (Phase-1 grep recipes still match).
- New per-dispatch diag: \`dispatch type=Navigate src=navigate\` — preserves the \`src\` tag the prior internal dispatcher logged with each state-write.
- Double-dispose / post-unregister guard: \`_dispatch\` checks \`bpSnapshot == null\` and drops with the existing \`post-close-event-dropped\` log shape.

### What's NOT in this PR

- **Phase 5 (diag panel)** — the audit ring is now populated for browser-pane events, but the panel UI that surfaces \`dispatchRecordsAtom\` is a separate piece of work.
- **Phase 6 (host-side title-change + favicon emit)** — the reducer cells are ready (\`TitleChanged\` command exists, faviconUrl derives automatically); the host event subscription is the remaining gap. Independent of Phase 4.

## Test plan

- [x] **+10 new slot-store tests** (\`browser-pane-state-store.test.ts\`):
  - dispatch on unregistered blockId throws (no silent drops)
  - registers and projects state changes
  - doesn't project cells that didn't change between dispatches
  - emits events through the event sink
  - PaneClicked routes through sink without state change
  - Disposed gates subsequent dispatches (post-close-command-dropped sink emission)
  - snapshot returns null for unknown blockId
  - unregisterPane removes the slot
  - re-registering resets state to initial
  - multi-pane: dispatches don't cross blockIds
- [x] **+3 new reducer tests** (\`reducer.test.ts\`):
  - PaneClicked emits pane-clicked event without changing state
  - PaneClicked doesn't touch any state (catalog rule §3)
  - PaneClicked after dispose is dropped (post-close gating)
  - Cross-product invariant test extended (PaneClicked included).
- [x] **All existing browser-model lifecycle tests still pass** (the dispose-is-idempotent test caught a real issue: double-dispose throwing because slot was unregistered. Added \`bpSnapshot == null\` guard in \`_dispatch\` — same semantics as the reducer's post-close gate, enforced at the model layer because the slot is gone entirely.)
- [x] Full frontend suite: **483/483 passing** (was 463; +20 new).
- [ ] Smoke: \`task dev\`, open browser pane, click into google search, click address bar — verify focus routing still works (the regression scenario from PR #760).

## Roadmap status after this PR

- ✅ Phase 0 — catalog
- ✅ Phase 1 — diag instrumentation (PR #738)
- ⏳ Phase 2 — characterize (manual retro pending)
- ✅ Phase 3a + 3b — closed/loading/error (PR #756)
- ✅ Phase 3c — canGoBack/canGoForward (PR #757)
- ✅ Phase 3d — \`url\` danger cell (PR #761)
- ✅ Phase 3e — title (PR #758) + faviconUrl (PR #763)
- 🟡 **Phase 4 shipped** ← this PR
- ❌ Phase 5 — diag panel surface (audit ring is populated; UI to render it is the remaining work)
- ❌ Phase 6 — host-side title-change + favicon emit

**Slice #9 cell-by-cell + slot lifecycle = COMPLETE.** Every per-pane data cell is reducer-backed; every state transition lands in the audit ring; the focus-routing click flows through the reducer with full traceability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.736"
+version = "0.33.737"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.736"
+version = "0.33.737"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.736"
+version = "0.33.737"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.736"
+version = "0.33.737"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.736"
+version = "0.33.737"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
--- a/agentmux-common/Cargo.toml
+++ b/agentmux-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-common"
-version = "0.33.736"
+version = "0.33.737"
 edition = "2021"
 description = "Shared utilities for AgentMux crates"
 
--- a/agentmux-launcher/Cargo.toml
+++ b/agentmux-launcher/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-launcher"
-version = "0.33.736"
+version = "0.33.737"
 description = "AgentMux Launcher — tiny exe that sets DLL path then spawns the CEF host"
 authors = ["AgentMux Corp"]
 edition = "2021"
--- a/agentmux-srv/Cargo.toml
+++ b/agentmux-srv/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-srv"
-version = "0.33.736"
+version = "0.33.737"
 edition = "2021"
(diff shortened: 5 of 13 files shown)


a5af:
@codex review

a5af:
@codex review

PR #762, opened 2026-05-09
perf phase 0: instrumentation infrastructure (marks + observers + IPC clock + dev HUD)

AgentA-asaf:
## Summary

Phase 0 of the performance work, per the spec also included in this PR (\`docs/specs/SPEC_PERFORMANCE_INSTRUMENTATION_AND_OPTIMIZATION.md\`). **No behavior change** — just measurement infrastructure that gives us the numbers to drive every subsequent optimization PR.

### What landed (5 components)

**A1 — `performance.mark` / `performance.measure` helpers** \`frontend/perf/marks.ts\`. \`markStart\` + \`markEnd\` pairs at the synchronous entry/exit of Tier-1 interactions. Wrapped tab switch (\`tabbar.tsx:handleSelect\`) and splitter drag (\`layoutResize.ts:onResizeMove\`). \`> 100 ms\` measures auto-warn to host log.

**A2 — Long Tasks observer** \`frontend/perf/observers.ts\`. Subscribes via \`PerformanceObserver({ type: \"longtask\" })\`. Every task ≥50 ms (3+ frame budget violations) immediately logs \`[perf] long-task Xms\` to host log. Counts kept in a sliding 5-second window for the HUD.

**A3 — INP / event observer** \`frontend/perf/observers.ts\`. Subscribes via \`PerformanceObserver({ type: \"event\", durationThreshold: 16 })\`. Filters to events with \`interactionId\` (real user interactions, not paint-only). Per-event-name P50/P75/P95 aggregations (\`click\`, \`keydown\`, \`pointerdown\`).

**C1 — IPC roundtrip clock** \`frontend/app/platform/ipc.ts\`. \`invokeCommand\` wraps each call with \`performance.now()\` before/after and pushes into \`perfStore\`. \`> 16 ms\` (one frame at 60 Hz) logs \`[perf] ipc <cmd> Xms\` immediately.

**Dev-mode HUD** \`frontend/perf/hud.tsx\`. Floating panel in bottom-right, toggled by \`Ctrl+Shift+P\`. Polls \`perfStore.snapshot()\` once per second. Shows: long-task count in last 5 s, top 5 interactions by P95, top 5 IPC commands by P95. Hidden in release via the existing \`isDev()\` gate. \`pointer-events: none\` so the HUD never intercepts user input.

### Aggregator design

[...]

PR #765, opened 2026-05-09
slice #9 phase 5: diag panel UI for the reducer audit ring

AgentA-asaf:
## Summary

Phase 5 of slice #9 — surfaces the global \`dispatchRecordsAtom\` (already populated by every slice's \`dispatch()\` via \`recordDispatch\`) as a floating dev-mode panel. **Slice #9 is now end-to-end.**

When a multi-pane focus or pool-drift bug fires, the question is always \"what happened in the last 200 ms, in what order, across which panes?\". Tail-grepping the host log works but loses structure. The audit ring already has the data structured; this panel just renders it.

### What landed

**\`frontend/app/devtools/diag-panel.tsx\`** — new floating Solid component, ~280 LOC.

- Toggle: **Ctrl+Shift+D** (or Meta+Shift+D on macOS).
- Mounted in \`app.tsx\` alongside \`PerfHud\` behind the existing \`isDev()\` gate. Hidden in release.
- z-index 999998 (one less than perf HUD at 999999) so they don't fight if both are open.
- Self-contained inline styles (same approach as the perf HUD) — works during a stylesheet-disabled investigation.

### Display

Per-row columns:

| Column | Source |
|---|---|
| age | \`now - record.at\` formatted as \"Xms ago\" / \"X.Ys ago\" / \"X.Ymin ago\" |
| slice | \`record.slice\` (e.g. \"browser-pane-state\", \"agent-pane-state\") |
| key | \`record.key\` truncated to 7 chars (block ID prefix); — for null |
| command | \`record.command.type\` discriminator |
| source | \`describeSource(record.source)\` — system / user / agent:X |
| events | count of events the dispatch emitted |

### Filters + controls

- **Slice dropdown** — populated reactively from the slices currently in the ring; empty string means \"all\".
- **Key filter** — substring search against \`record.key\` (case-insensitive).
- **Clear button** — wipes the ring via \`__resetDispatchLog\`. Useful for isolating a specific repro window.

### Display limit + age refresh

Newest-first scan stops once we've collected \`DEFAULT_DISPLAY_LIMIT = 80\` filtered rows. The records signal already drives re-renders on every new dispatch; a separate 1 Hz \`now()\` interval refreshes the relative-age strings on existing rows. Both interval + dispatch subscription are bounded to the visible state via a \`createEffect\` that tears down the timer when the panel is hidden — same pattern as the perf HUD's visibility gating.

### What's NOT in this PR

- **Per-record event-payload expansion** — the panel shows \`record.events.length\`, not the events themselves. Click-to-expand-row is a natural enhancement; deferred so this PR stays focused on the audit ring's primary value (sequence + frequency visibility).
- **Persistent record export** — the ring is in-memory only per the existing convention. If a session needs to be inspected post-mortem, that's a separate feature with separate threat-model concerns.
- **Slice-specific renderers** — the panel renders all slices uniformly. A future enhancement could downcast \`record.command\` based on \`record.slice\` and render slice-specific detail; out of scope here.

## Test plan

- [x] Full frontend suite: **483/483 passing** (no new tests — the panel is a thin reactive view over an already-tested signal; behavior is observable via the rendered output, which is best validated in the dev-mode smoke).
- [x] No TS regressions.
- [ ] Smoke: \`task dev\`, press Ctrl+Shift+D, click in the address bar, navigate, click into google search — verify the panel shows the corresponding sequence (Navigate, UrlConfirmed, LoadFinished, PaneClicked, etc.) with correct slice/key/source attribution and the age column ticking.

## Slice #9 status after this PR

- ✅ Phase 0 — catalog
- ✅ Phase 1 — diag instrumentation (PR #738)
- ⏳ Phase 2 — characterize (manual retro pending, optional)
- ✅ Phase 3a + 3b — closed/loading/error (#756)
- ✅ Phase 3c — canGoBack/canGoForward (#757)
- ✅ Phase 3d — \`url\` danger cell (#761)
- ✅ Phase 3e — title (#758) + faviconUrl (#763)
- ✅ Phase 4 — slot store + recordDispatch + PaneClicked through reducer (#764)
- 🟡 **Phase 5 shipped** ← this PR
- ❌ Phase 6 — host-side title-change + favicon emit (the original feature; reducer cells are ready, host event subscription is the remaining gap — independent work)

The reducer-stack roadmap §5 said \"Wire \`recordDispatch\` for \`browser-pane-state\` into the existing diagnostics panel.\" There was no existing panel; this PR builds one and surfaces every slice's dispatches, not just browser-pane-state. Future slices come along for free.

## Cross-references

- \`docs/specs/browser-pane-reducer-roadmap.md\` §5
- \`frontend/app/store/command-source.ts\` — the audit ring (slice #3 from the conventions)
- \`frontend/perf/hud.tsx\` — sister panel; same toggle/visibility-gating pattern
- Predecessors in this session: #756, #757, #758, #759, #760, #761, #762, #763, #764
- Discussion: #707

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.737"
+version = "0.33.738"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.737"
+version = "0.33.738"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.737"
+version = "0.33.738"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.737"
+version = "0.33.738"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.737"
+version = "0.33.738"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
--- a/agentmux-common/Cargo.toml
+++ b/agentmux-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-common"
-version = "0.33.737"
+version = "0.33.738"
 edition = "2021"
 description = "Shared utilities for AgentMux crates"
 
--- a/agentmux-launcher/Cargo.toml
+++ b/agentmux-launcher/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-launcher"
-version = "0.33.737"
+version = "0.33.738"
 description = "AgentMux Launcher — tiny exe that sets DLL path then spawns the CEF host"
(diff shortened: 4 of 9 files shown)


a5af:
@codex review

a5af:
@codex review

PR #767, opened 2026-05-09
hotfix: perf IPC clock feedback loop — exclude log-pipe IPCs

AgentA-asaf:
## Summary

PR #762's IPC roundtrip clock created a feedback loop with the log-pipe. Symptom user hit: tab-bar drag stops working — \`get_window_position\` IPC takes 700-1500 ms instead of <1 ms because the IPC server is saturated with self-generated traffic.

## The loop

\`\`\`
slow IPC > 16ms → recordIpcRoundtrip → console.warn(\"[perf] ipc X Yms\")
  → log-pipe monkey-patched console.warn → invokeCommand(\"fe_log_structured\")
  → recordIpcRoundtrip on the fe_log_structured call itself
  → if THAT was >16ms (because pipe is now saturated) → console.warn
  → log-pipe → fe_log_structured → recordIpcRoundtrip → …
\`\`\`

Observed in dev mode: **283,469 \`fe_log_structured\` warnings** in ~10 minutes; ambient IPC latency 700-1500 ms; \`get_window_position\` queues behind the storm and takes 700+ ms; the click-and-release window is shorter than that, so \`useWindowDrag.win32.ts\`'s race guard cancels every drag attempt.

Desktop portables predate PR #762, which is why they don't show this regression — confirmed by user.

## Fix

Blocklist \`fe_log_structured\` and \`fe_log\` in \`recordIpcRoundtrip\` — those are the log-pipe's own IPCs and must not be measured (their cost IS the measurement infrastructure; measuring them = measuring the meter). Two-line change.

## Test plan

- [x] Frontend perf tests pass (7/7).
- [ ] Smoke: with this PR built into a fresh \`task dev\`, drag the title bar — should work without delay; \`get_window_position\` should be <1 ms.

## Lesson

Phase-0 spec said \"instrumentation should be effectively free.\" This was a self-introduced violation. The class is \"don't measure your own measurement infrastructure.\" Memory note will be added so future instrumentation tools (long-task observer, INP observer when its own callbacks log) get the same treatment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.738"
+version = "0.33.739"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.738"
+version = "0.33.739"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.738"
+version = "0.33.739"
 dependencies = [
  "agentmux-common",
  "chrono",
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

a5af:
@codex review

PR #772, opened 2026-05-09
perf: paint-aligned tab-switch marks at canonical setActiveTab

AgentA-asaf:
## Summary

The Phase 0 tab-switch perf mark in `tabbar.tsx handleSelect`:
1. Only fired on **imperative click** — keyboard (Ctrl+Tab, Cmd+1..9), command palette, and test-app-API paths bypassed it.
2. Closed via `queueMicrotask`, so the duration captured only the synchronous IPC dispatch, not the user-perceived layout + paint cost.

This PR moves the mark into the canonical `setActiveTab` chokepoint and closes it via double-`requestAnimationFrame`, so the duration reflects IPC + Solid fan-out + layout + paint.

Backend-driven switches (tearoff merge, cross-drag) bypass `setActiveTab` entirely — they're rare and observable via the long-task timeline; documented inline.

## Coverage map

| Path | File | Now instrumented? |
|---|---|---|
| Click | `tabbar.tsx handleSelect` | ✅ via setActiveTab |
| Keyboard Ctrl+Tab / Shift+Tab | `keymodel.ts switchTab` | ✅ via setActiveTab |
| Keyboard Cmd+1..9 | `keymodel.ts switchTabAbs` | ✅ via setActiveTab |
| Command palette | `command-registry.ts` | ✅ via setActiveTab |
| Test app API | `cef-api.ts setActiveTab` | ✅ via setActiveTab |
| Tearoff merge | `tabbar.tsx tearoff:merge` | ❌ backend-driven, deferred |
| Cross-window drag | `DragOverlay.tsx` | ❌ backend-driven, deferred |

## Verification

Built locally on `agenta/perf-tab-switch-marks` (v0.33.740). Tab switches in dev mode now produce `[fe] [perf] tab-switch <dur>ms` warnings (>100ms threshold) for every entry path. Empirically confirmed via dev-session log: durations track visible jank — switches into tabs hosting an agent pane are longer than browser/terminal-only switches, matching user-perceived behavior.

## Related

- Tier 1 spec: `docs/specs/SPEC_PERFORMANCE_INSTRUMENTATION_AND_OPTIMIZATION.md`
- Phase 1 retro: `docs/retro/perf-baseline-2026-05-09.md`
- Issue #769 (new-window bootstrap serialization)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #769, opened 2026-05-09
Perf: new-window bootstrap pays serialization tax on later tearoffs


AgentA-asaf:
## Summary

After 3 tearoffs in quick succession (2026-05-09 dev session, v0.33.739), the **new-window bootstrap path** of the 3rd tearoff was 2.8× slower than the 1st on every IPC, plus a 156ms long-task and 12 serial `get_*` IPCs. Tearoff itself was fine; the cost is in the new window's startup handshake.

## Evidence

3 tearoffs at 15:56:31 / :37 / :43. Tearoff IPCs themselves are stable:

| # | `tear_off_pool_promote` | `complete_cross_drag` |
|---|---|---|
| 1 | 129.8ms | 62.1ms |
| 2 | 120.4ms | 36.8ms |
| 3 | 117.4ms | 56.9ms |

But the new-window bootstrap right after tearoff #3:

| IPC | After #1 | After #3 |
|---|---|---|
| `get_backend_info` | 194ms | **541ms** |
| `set_window_transparency` | — | **520ms** |
| `register_backend_window` | — | **452ms** |
| `set_window_init_status` | 226ms | **451ms** |
| `update_wco` | 212ms | **269ms** |

Plus a **156ms long-task** (~10 frames dropped) and **12 sequential `get_*` IPCs** at 25-28ms each (~330ms serial cost):
`get_auth_key, get_is_dev, get_platform, get_user_name, get_host_name, get_data_dir, get_config_dir, get_docsite_url, get_user_home_dir, get_zoom_factor, get_about_modal_details` (+ one extra).

## Hypotheses (ranked)

1. **Host-side serialization on per-window registration** — `register_backend_window` + `set_window_init_status` + `set_window_transparency` triple-up to ~1.4s; likely a single async lock around the window-list mutation that everyone queues behind. Each new window adds queue depth.
2. **Bootstrap IPCs run serially when they could be parallel/batched** — the 12 `get_*` calls each cost 25-28ms in flight, totalling ~330ms. Replace with one `get_window_bootstrap_data` that returns the bundle in one round-trip → ~25ms.
3. **GPU/CEF resource contention** — 3rd window competes for compositor resources; would explain the 156ms long-task more than the IPC slowdowns.

## Suggested fix order

[...]

PR #773, opened 2026-05-10
perf: virtualize AgentDocumentView with @tanstack/solid-virtual

AgentA-asaf:
## Summary

Virtualize the agent message list (`AgentDocumentView`) using `@tanstack/solid-virtual` so only the visible window + 5-item overscan is mounted in the DOM. Resolves Issue 4 from the agent-pane perf report (DOM virtualization).

**Driving symptom:** in long agent sessions (500-2000+ document nodes), the layout tree alone reaches ~8MB per pane. Tab-switches *into* a pane hosting an active agent stalled for 200ms+ (multiple long-tasks per switch, worst observed 269ms). Switching b↔c (no agent) is fast; switching a↔b (agent in `a`) is slow. Confirmed live in the dev session that drove this work (#769 file).

## What changed

- `frontend/app/view/agent/components/AgentDocumentView.tsx` — replaced `<For each={document()}>` with virtualizer-driven absolute-positioned items.
- `frontend/app/view/agent/styles/_document.scss` — removed `content-visibility: auto` / `contain-intrinsic-size` from `.agent-document-node-wrapper` (no longer needed, items aren't in the DOM at all when off-screen). Added `.agent-document-virtualizer` for the `position: relative` wrapper.
- `package.json` / `package-lock.json` — added `@tanstack/solid-virtual` ^3.13.

## Three reworked flows

These read DOM directly under the old `<For>` and had to be reauthored:

| Flow | Old | New |
|---|---|---|
| `scrollToNode` (jump-to-bookmark / search) | `querySelector + getBoundingClientRect`, smooth-scroll to centerOffset | Resolve index first; `virtualizer.scrollToIndex(idx, { align: 'center' })`. Old approach failed if target was outside virtual window. |
| `jumpToBottom` / `scrollToBottom` | `scrollTop = MAX_SAFE_INTEGER` (deliberately, to avoid sync layout read on keystroke hot path) | `virtualizer.scrollToIndex(last, { align: 'end' })` — also avoids sync layout reads. |
| `handleScroll` older-history load | Snapshot `scrollHeight + scrollTop`, restore via `new_scrollTop = new_scrollHeight - old_scrollHeight + old_scrollTop` | Anchor by node id. Pre-virt the pixel-delta math was reliable; with measured virtual heights it isn't, until every newly-rendered item settles. Re-anchoring to the same node's index after prepend is exact regardless of measurement timing. |

## Risk surfaces considered

| Concern | Status |
|---|---|
| **Variable item heights** (tool blocks 32→400px when pinned, code diffs 500+px) | TanStack supports dynamic sizing via `measureElement` + ResizeObserver. estimateSize=80 matches old contain-intrinsic-size hint; real heights settle on first paint. |
| **Collapse/expand state** (`collapsedNodes`, `pinnedNodes` Sets in DocumentState) | Held in **parent** state, not child — already accessed via reactive accessors per existing comment at line 449. Re-mount during scroll preserves state. |
| **Tool-pin overlay portal** (Tool blocks render `<Portal>` to body when pinned, measure via getBoundingClientRect) | Items in the virtual window have valid refs. Pinning a tool that's in the visible region works. Off-screen pin is not a user-reachable action (the pin button is on the row, which means the row is visible). |
| **Truncate-suppression grace window** (5000ms post-SessionStart in reducer) | View-only change — reducer logic untouched. |
| **DOM export / copy via getSelection()** | Native selection unaffected by re-mount of off-screen items. |
| **scrollMargin for loading-older banner / auth-url box** | `scrollMargin` getter reads `virtualContainerRef.offsetTop` so virtualizer scroll math accounts for content rendered above the virtual list. |
| **`<For>` keying of virtualItems** | `getItemKey: (idx) => document()[idx]?.id` provides stable keys across reorder/prepend; SolidJS For's reference identity is stable when the underlying nodes don't change. |

## Untested scenarios I want reviewers to flag

- **Very long pinned tool block with code diff inside.** measureElement should settle to the full expanded height after one frame, but if the inner DiffViewer mounts async (Suspense, dynamic import, or its own virtualization), the measure might shoot up after several frames. Smoke this with a long bash output that's pinned-open.
- **Prepend-while-scrolling** — user scrolls toward top, history loads, then user keeps scrolling before the rAF anchor restore fires. The anchor logic is correct but the brief frame between prepend and anchor-restore could show jump.
- **Search-jump to a node deep in history** — the new `findIndex(node.id)` does an O(n) scan; at 2000 nodes it's fast but worth confirming.

## Local smoke

Type-check passes. Local dev smoke deferred — ran in a `git worktree` so I didn't disrupt the user's running test session for #772. Would appreciate a smoke pass from a reviewer with a long agent session.

## Cross-references

- Agent pane perf report (driving doc): not in repo, summary in PR description.
- Issue 4 (the relevant excerpt): "No DOM virtualization in AgentDocumentView. SolidJS `<For>` is fine-grained but does not remove DOM nodes for off-screen items. With `content-visibility: auto`, the browser skips painting off-screen nodes but still maintains them in the layout tree and retains their JS closures."
- Tab-switch mark coverage (PR #772) — measures the user-perceived switch cost this PR is trying to reduce. After both land, the same long-task instrumentation will show whether the agent-pane reflow on tab-switch drops.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.743"
+version = "0.33.744"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.743"
+version = "0.33.744"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.743"
+version = "0.33.744"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.743"
+version = "0.33.744"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.742"
+version = "0.33.743"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.742"
+version = "0.33.743"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.742"
+version = "0.33.743"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.742"
+version = "0.33.743"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

a5af:
@codex review

PR #781, opened 2026-05-10
revert: virtualize AgentDocumentView (#773) — redesign in spec

AgentA-asaf:
## Summary

Reverts PR #773 (\`@tanstack/solid-virtual\` retrofit on \`AgentDocumentView\`).

## Why

The retrofit introduced multiple regressions in interactive use:
- **Gaps between rows** — estimateSize dominates layout until measureElement settles; even after measurement, short messages get visible padding below.
- **Tables broken** — markdown tables inside absolute-positioned virtualizer rows lose their layout context (no width to constrain against; \`table-layout: auto\` overflows).
- **Streaming content cut off** — long thinking blocks render only the first word; the wrapper's height doesn't grow chunk-by-chunk.

The deeper problem isn't any one bug — it's that virtualization was bolted on after the fact. The original component assumed every node was in the DOM (querySelector for jump, pixel-math for pagination restore, natural flex flow with gap). Each of those needed a workaround. Each workaround introduced its own edges.

## What replaces it

A redesigned virtualization layer specced separately in `docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_REDESIGN.md` (filed alongside this revert). The redesign treats virtualization as a first-class architectural concern: per-kind size estimators, scroll state in data not DOM, hybrid render (last N items unvirtualized as the streaming buffer), CSS overflow-anchor, and integrated per-kind perf probing.

The perf measurements from #773's brief life are still useful — they confirm the win is real (long-task storms during tab-switch into agent panes drop substantially when the layout tree is bounded). The redesign keeps that win while addressing the layout/streaming issues structurally.

## Impact

- Agent pane returns to pre-virtualization behavior: full DOM with `content-visibility: auto`. The 200-300ms long-task storms during tab-switch return for long sessions until the redesign ships.
- Clean main; no half-broken behavior in users' hands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.747"
+version = "0.33.748"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.747"
+version = "0.33.748"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #783, opened 2026-05-10
feat(virt): Phase 1 — foundation (store, anchor, partition, registry)

AgentA-asaf:
## Summary

First phase of the agent pane virtualization redesign (issue #782, spec at `docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_REDESIGN.md`). **Pure data structures + tests; no UI change.**

## What's in this PR

New module: `frontend/app/view/agent/virtualization/`

| File | Purpose |
|---|---|
| `anchor.ts` | Pure scroll-anchor math (capture/restore, near-top/near-bottom). Replaces the inline pixel-delta math from the old AgentDocumentView. |
| `state.ts` | `AgentViewState` Solid factory — `nodeIndex` memo, `stickToBottom`, `headAnchor`, `streamingNodeId`. The single source of truth Phase 2 will project into the DOM. |
| `streaming-buffer.ts` | Partition splitting the document into a virtualized head and an unvirtualized streaming tail. `STREAMING_BUFFER_SIZE = 50`. |
| `renderers.ts` | Per-kind renderer registry (component + `estimatedSize` + `isStreamingCapable`). Per-kind estimators replace the single `estimateSize: () => 80` that caused gaps in the reverted #773. |
| `index.ts` | Public surface for Phase 2/3 imports. |

## Tests

50 cases across the four modules. All passing.

```
Test Files  4 passed (4)
     Tests  50 passed (50)
```

Covers:
- Anchor capture/restore round-trip + clamp behavior
- `stickToBottom` semantics (initial, atomic flip on capture, post-clear behavior)
- `nodeIndex` reactivity to document changes (prepend re-indexes correctly)
- Partition boundaries: empty doc, length == bufferSize, length == bufferSize+1, default size
- Per-kind estimators: collapsed/pinned state, text-height scaling, max-cap
- Estimator dispatch correctness for all 6 node kinds
- Streaming-capability flags

## Why this scope

Phase 1 deliberately ships **no UI change**. The goal is a clean, well-tested foundation that Phase 2 builds on. This avoids the retrofit-regression class that killed PR #773.

Phase 2 will:
- Build `AgentDocumentVirtualList` component
- Hybrid render (virtualized head + unvirtualized streaming tail)
- Replace `AgentDocumentView`'s `<For>` block
- CSS: `width: 100%` rows, `table-layout: fixed`, `overflow-anchor: auto`

Phase 3 will add the perf-probe HUD that validates `estimatedSize` against measured heights.

## Cross-references

[...]

Issue #782, opened 2026-05-10
Agent pane virtualization redesign + intelligent perf probing


AgentA-asaf:
## Why

PR #773's virtualization retrofit was reverted (PR #781) after it caused gaps, broken tables, and streaming cut-off. The pattern was clear: each fix-as-we-go workaround introduced new edges because virtualization was bolted onto code that assumed every node was in the DOM.

## Plan

Architectural redesign in `docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_REDESIGN.md` on branch `agenta/spec-agent-pane-virt-redesign`. Treats virtualization as a first-class concern + adds intelligent perf probing built into the render contract.

## Key design points

- **Per-kind size estimators** instead of one global `estimateSize: 80` (eliminates gaps for short messages)
- **Hybrid render**: last 50 nodes unvirtualized (streaming buffer) — eliminates streaming-cut-off bug class entirely
- **Scroll state in store, not DOM** (`headAnchor` + `stickToBottom`) — single source of truth
- **`table-layout: fixed` + `width: 100%` rows** — fixes table distortion
- **CSS `overflow-anchor: auto`** as Chromium-native belt to JS anchor's suspenders
- **Integrated perf probing** — per-kind marks (p50/p95/max), estimator-miss detection, layout-shift attribution scoped to agent pane, dev HUD extension to slice #9 diag panel

## Industry references

Validated against: `react-virtuoso`'s Message List, Stream's `VirtualizedMessageList`, `use-stick-to-bottom`, Discord's mobile rewrite, `react-virtualized` Table patterns, `overflow-anchor` MDN guidance.

## Effort

~4.5 days across 4 phases (foundation, virtualization layer, perf probing, hardening).

## Cross-references

- PR #781 — revert of #773 (must merge before this work starts)
- Issue #774 — tab content reveal gate (complementary)
- `frontend/perf/marks.ts`, `frontend/app/devtools/diag-panel.tsx` — extension points

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Agent1-asaf:
Folded into the new typing-performance umbrella discussion: agentmuxai/agentmux#1161.

Status verified 2026-05-29:
- Phases 1–3 shipped (PRs #783, #784, #787) ✅
- Phase 4 (hardening) — no PR found; this issue is the remaining child

[...]

PR #784, opened 2026-05-10
feat(virt): Phase 2 — virtualization layer + streaming buffer

AgentA-asaf:
## Summary

Second phase of the agent pane virtualization redesign (issue #782, spec at `docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_REDESIGN.md`). **Builds on Phase 1 (PR #783).**

> **Base branch is `agenta/virt-redesign-phase-1`** so the diff stays focused. Once #783 merges to main, this PR will rebase onto main automatically.

### Components (`frontend/app/view/agent/virtualization/`)

- **`DocumentRow.tsx`** — single row component used by both regions. Caller controls positioning (style + ref), so the same row works in absolute-positioned virtualizer slots and normal-flow streaming buffer without knowing the difference.
- **`AgentDocumentVirtualList.tsx`** — hybrid renderer:
  - `@tanstack/solid-virtual` for the virtualized head (per-kind estimators from Phase 1's registry, `measureElement` for dynamic heights, `scrollMargin` for the optional header slot)
  - Plain `<For>` for the trailing 50-node streaming buffer — **no virtualization, no measurement lag during token streams** (eliminates the streaming-cut-off bug class from #773)
  - Scroll state read from `AgentViewState` via the atomic `engageStickToBottom` / `disengageStickToBottom` / `captureHeadAnchor` API

### CSS (`_document.scss`)

- `overflow-anchor: auto` on `.agent-document` (CSS half of the anchor)
- `.agent-document-virtualizer` — `position: relative; flex-shrink: 0`
- `.agent-document-streaming-buffer` — flex column with gap, also `flex-shrink: 0`
- `.agent-document-row` — `contain: layout style; width: 100%`
- **`:is(table, .markdown-table) → table-layout: fixed; width: 100%`** — fixes the table-distortion class from #773

### `AgentDocumentView` rewritten as thin shell

Owns data-state interface (auto-collapse policy, `toggleCollapse`, `togglePin`, the auth-url box) and delegates list rendering to `AgentDocumentVirtualList` via the `headerSlot` prop. All scroll behavior (jumpToBottom, scrollToNode, anchor capture, pagination restore) lives in the VirtualList — single source of truth.

## Bug-class coverage (vs reverted #773)

| Bug from #773 | Why this PR doesn't have it |
|---|---|
| Gaps from estimateSize miss | Per-kind estimators (`estimateAgentMessage`, `estimateTool`, etc.) tuned per type |
| Tables broken | `table-layout: fixed; width: 100%` on every row's tables |
| Thinking cut-off mid-stream | Last 50 nodes always in unvirtualized streaming buffer; measureElement never recycles them mid-stream |
| Scroll-restore fragile | Anchor model (`headAnchor` in store) replaces pixel-delta math; restore via `getOffsetForIndex` (virtualized) or `el.offsetTop` (streaming) |
| Jump-to-node fails for off-screen | `scrollToIndex` first when target is in virtualized region; DOM query when in streaming buffer |
| Stale stick-to-bottom anchor | Phase 1's atomic `engageStickToBottom` clears anchor (codex P2 fix) |

## Re-adds dependency

`@tanstack/solid-virtual` ^3.13 — same dep that was removed in PR #781. This time the design accommodates it from the start, so the eval contract is much cleaner.

## Test status

- Phase 1's 51 tests still pass.
- Phase 2 deliberately doesn't add new view-layer tests — Phase 3 will add the perf-probe HUD which validates render correctness empirically against an active session.
- TypeScript check passes for all new files (the pre-existing TS errors in `frontend/layout/`, `frontend/app/tab/tab-presets.ts`, `frontend/util/keyutil.ts` are unrelated to this PR).

## Next: Phase 3

Per-kind perf marks, estimator-validation HUD, layout-shift attribution scoped to the agent pane. Will surface estimator misses > 30% so we can tune the per-kind sizes empirically.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.751"
+version = "0.33.752"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.751"
+version = "0.33.752"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.751"
+version = "0.33.752"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.751"
+version = "0.33.752"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.752"
+version = "0.33.753"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

PR #786, opened 2026-05-10
fix(virt): drop one-time Show node capture (lost from PR #784 squash)

AgentA-asaf:
## Summary

Follow-up to #784. The Show-capture P1 fix from reagent landed on the Phase 2 branch as commit 12b0b4dd but the squash-merge fired ~5min earlier (12:29:54Z) than my last push (12:35Z), so the fix didn't end up in main.

## What was lost

The current main has:

```tsx
<Show when={props.node()}>
    {(n) => {
        const node = n();  // ← one-time snapshot, frozen at first mount
        switch (node.type) { ... }
    }}
</Show>
```

SolidJS `<Show>` with `keyed: false` (default) calls the child factory once when `when` first becomes truthy and never re-runs it. The captured `node = n()` is frozen. Streaming token updates and tool transitions (running → success, etc.) silently fail to propagate to `ToolBlock` / `AgentMessageBlock` until the row remounts — defeating the whole streaming buffer.

## Fix

Restructure as separate `<Show when={node && node.type === "X"}>` branches per kind. Each branch reads `props.node()` reactively at every prop site. Solid tracks the access; updates flow through.

## Tests

51/51 virtualization tests still passing.

## Verified the loss

```
$ git show 39c2b9fb:frontend/app/view/agent/virtualization/DocumentRow.tsx | grep -A 1 'function DocumentNodeBody'
function DocumentNodeBody(props: DocumentNodeBodyProps): JSX.Element {
    return (
        <Show when={props.node()}>
            {(n) => {
                const node = n();
                switch (node.type) {
```

Confirmed: the unfixed pattern shipped to main.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.757"
+version = "0.33.758"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.757"
+version = "0.33.758"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #794, opened 2026-05-11
fix(menu): submenu z-order above main menu + reset state on close

AgentA-asaf:
## Summary

Two UX fixes on the hamburger menu.

### 1. Submenu z-order

The submenu div had an inline `z-index: 1000` that overrode the `.menu` class's `--z-flyout-menu: 9500` (`theme.scss:279`). So submenus painted ~8500 layers **below** the main menu (which DID read 9500). Anything stacked between 1000 and 9500 (other UI, browser-pane overlays, etc.) covered the submenu — and the 2px overlap at the submenu's left edge always hid behind the main menu's right edge.

Removed the inline override. Submenu inherits the `.menu` z-index of 9500, matching the main menu. Source order then resolves stacking among same-z Portals; the submenu Portal mounts after the main menu, so it paints on top of the overlap region.

Vertical placement is **unchanged** (submenu's top aligns with parent's top, extending down to the right). The earlier draft of this PR tried changing the vertical direction — that was a misread of the request and was reverted.

### 2. Reset submenu state on close

`onOpenChangeMenu(false)` now clears `visibleSubMenus`, `hoveredItems`, and `subMenuPosition`. Previously these maps persisted, so a user who hovered Theme → clicked away → reopened the hamburger would see both the main menu AND the Theme submenu re-mount with stale state. Now reopen shows only the main menu.

## Test plan

- [ ] Open hamburger → hover Theme → submenu opens to the right with NO visual gap or occlusion at the left edge (the 2px overlap area)
- [ ] Move cursor across the overlap → no flicker, submenu fully visible
- [ ] Open hamburger → hover Theme → click outside → click hamburger again → only main menu opens (no stale Theme submenu)
- [ ] Same flow with Opacity

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.778"
+version = "0.33.779"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.778"
+version = "0.33.779"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #796, opened 2026-05-11
feat(menu): Phase 1 — unify menu chrome across hamburger, right-click, and tab popover

AgentA-asaf:
## Summary

**Phase 1** of `docs/specs/SPEC_UNIFIED_MENU_SYSTEM_2026_05_11.md`. The hamburger menu, every right-click context menu, and the tab color/rename popover all paint with the same outer chrome and item styling now.

Driving observation (verbatim): *"I like the menu styles on the hamburger, but the right-click on other content menus are different. Let's reconcile them into a cohesive system."*

## What changed

- **New `frontend/app/element/menu-frame.scss`** exports a single `menu-frame` SCSS `@mixin`: background, border, 6px radius, shadow, padding.
- **`flyoutmenu.scss`** `.menu` `@include`s the mixin; outer radius nudged 4→6px to match modern dropdown norms. FlyoutMenu otherwise visually identical.
- **`tab.scss`** `.tab-context-panel` adopts the same mixin. Drops its duplicate chrome declarations.
- **`cef-api.ts` `showJsContextMenu`** (the renderer behind `ContextMenuModel.showContextMenu` and therefore every right-click site) replaces all `Object.assign(el.style, {...})` inline-style calls with class assignments — `.menu`, `.menu-item`, `.menu-divider`, `.menu sub-menu`. Indicators change from Unicode `●`/`✓` to a `fa-check` icon in the existing `.menu-item-check` slot. Submenu chevron changes from `▸` to `fa-chevron-right`. Hover changes from solid `var(--accent-color)` to faded `var(--hover-bg-color)` automatically via the `.menu-item:hover` rule.
- **Spec document** ships in this PR (no doc-only PRs per project policy).

Positioning (`top`/`left`), the per-instance click-outside overlay, and the `enabled === false` state stay inline — they're per-invocation, not part of the shared chrome.

## Phases 2-5

The spec lays out Phase 2 (Solid `<Menu>` primitive replacing the vanilla DOM in `showJsContextMenu`), Phase 3 (type unification — `ContextMenuItem` → `MenuItem` alias), Phase 4 (TabContextPanel as a custom `<Menu>` slot — optional), Phase 5 (keyboard nav + a11y). All deferred to follow-up PRs.

## Test plan

- [ ] Hamburger ≡ → menu opens, looks identical to before (it's the source of truth)
- [ ] Right-click empty space in tab bar → menu now has the same outer chrome, faded hover, fa-check radio indicator
- [ ] Right-click a pane header → same
- [ ] Right-click a pane body → same
- [ ] Right-click a tab → color picker / Rename / Close uses the same chrome
- [ ] Right-click a document row → bookmark menu uses same chrome
- [ ] Submenu open/close in right-click menus still works
- [ ] `data-pane-overlay` clipping still works on Windows


a5af:
@codex review

PR #798, opened 2026-05-11
fix(tearoff + empty-tab): preserve grab offset across dragend + brain-logo center on empty tabs

AgentA-asaf:
## Summary

Two unrelated user-visible fixes that happened to land on the same branch during smoke-testing of the merged #796 menu unification work.

### 1. Tear-off no longer plants the new window's caption-center under the cursor

Dragging a tab outside the AgentMux window opened the new instance with the **caption center** at the cursor, not the new window's **first tab** under the cursor where the user grabbed.

Multi-step diagnosis chain (commits visible in PR):

- **Round 1**: added a diagnostic `Logger.info("dnd", "tear-off anchor compute", ...)` to `tabbar.tsx::onDrag` — the existing threshold-cross tear-off path. Build #3 → diagnostic NEVER fired. Concluded the active tear-off path was actually `CrossWindowDragMonitor.win32.tsx`, not `tabbar.tsx`.
- **Round 2**: thought the issue was `getTabGrabOffset()` getting called too late (after 50ms `setTimeout`), so snapshotted at `handleDragEnd`. Build #4 → new `[dnd:cross] performTearOff anchor` log line *still* showed grabOffset as undefined.
- **Round 3 (this fix)**: pragmatic-dnd's `onDrop` listener fires DURING the dragend event dispatch — *before* my document-level dragend listener — and calls `setTabGrabOffset(null)`. The snapshot in `handleDragEnd` was always too late. Fixed at the actual source: stopped clearing the offset on drop. Every new drag's `onGenerateDragPreview` overwrites it before any read, so leaving the value persistent across the no-drag interval is safe.

Build #5 (v0.33.789) verified by smoke test — new window's first tab lands precisely under the cursor.

The snapshot-at-dragend code and the `performTearOff anchor` Logger.info from round 2 stay — they make the diagnostic log show real grabOffset values during the dragend phase even if listener ordering ever flips back.

### 2. Empty tab: brain logo center, drop the accent border

When a tab has zero blocks, the empty area previously showed an inset 1px accent-colored border that didn't match the width of any pane border elsewhere — read as a visual artifact rather than a frame.

- Removed the `box-shadow: inset 0 0 0 1px var(--accent-color)`.
- Added the brain-alternate AgentMux logo (`frontend/app/asset/logo-brain.svg`, copied from `frontend/logos/`) centered in the empty area at 40% opacity, `pointer-events: none` so right-click for the widget menu still works through it.

## Test plan

- [x] Empty tab shows centered brain logo, no accent border (smoke-verified locally)
- [x] Tear off a tab to outside the window → new window opens with its first tab under the cursor at the same offset the user grabbed (smoke-verified on v0.33.789)
- [x] Host log shows `[dnd:cross] dragend fired { grabOffsetX, grabOffsetY }` and `[dnd:cross] performTearOff anchor { tabAnchorX, tabAnchorY }` with real values (no longer undefined)
- [ ] Other re-dock / cross-window-drop paths unchanged


a5af:
@codex review

PR #801, opened 2026-05-11
secure: gate /agentmux/reactive/* under auth_middleware (audit C1+C2)

Agent1-asaf:
## Summary

Closes the highest-leverage findings from the 2026-05-11 security audit. The \`/agentmux/reactive/*\` routes were registered without auth on the assumption that localhost is a trust boundary. It isn't — any local process, or any web page the user has open (via the permissive \`Access-Control-Allow-Origin: *\` layer), could drive these endpoints and:

- queue arbitrary prompts into any running agent's PTY (\`POST /agentmux/reactive/inject\`) → prompt injection, RCE-equivalent via the agent's own context
- reconfigure the cloud agentbus poller to an attacker URL + supply an attacker token (\`POST /agentmux/reactive/poller/config\`)
- read the existing poller token via \`GET /agentmux/reactive/poller/status\`

These are **C1 + C2** in \`/workspace/SECURITY_AUDIT_2026_05_11.md\`. **C3** (CORS \`Any\` + query-string \`?authkey=\`) is a separate PR that follows.

### Server router
- \`agentmux-srv/src/server/mod.rs\` — merge \`reactive_routes\` into the same auth-gated group as \`authed_routes\`. One-line shift; same \`route_layer(auth_middleware)\`.

### Cross-instance forward (so multi-instance jekt still works)
- \`agentmux-srv/src/backend/reactive/registry.rs\` — \`AgentEntry\` gains an \`auth_key\` field, populated from a process-wide \`LOCAL_AUTH_KEY\` OnceLock initialised by main.rs from \`config.auth_key\`.
- \`agentmux-srv/src/server/reactive.rs\` — the cross-instance HTTP forward sends \`X-AuthKey: entry.auth_key\` so the receiving peer's middleware can authenticate.
- Registry files are mode 0600 on Unix because they now carry a sensitive token (same boundary as \`authkey.dev\`).
- \`auth_key\` is \`#[serde(default)]\` so older on-disk entries still deserialize; their forwards will 401 and the cloud agentbus fallback handles it. Not a regression of any current behavior.

### Frontend
- \`frontend/app/view/term/termagent.ts\` — \`registerAgent\` / \`unregisterAgent\` send \`X-AuthKey\` via \`getApi().getAuthKey()\`.
- \`frontend/app/view/term/termosc.ts\` — \`poller/config\` does the same.

### Tests
- Replace \`server::tests::reactive_routes_skip_auth\` (which asserted the bug) with two new tests: 401 without key, 200 with valid key.
[...]

PR #802, opened 2026-05-11
secure: strict CORS + drop ?authkey= on non-/ws routes (audit C3)

Agent1-asaf:
## Summary

Closes the last Critical in the 2026-05-11 security audit, completing the C1/C2/C3 trio.

## Background

The sidecar was wide open to same-machine CSRF via two mechanisms:

1. **\`Access-Control-Allow-Origin: *\`** — any web page the user had open in any browser could attempt cross-origin calls to the sidecar once it discovered the port.
2. **\`?authkey=\` query-string fallback** — the auth_middleware accepted the key in the URL. Query strings leak into browser history, the \`Referer\` header (cross-origin requests on the page after navigation), server access logs, and any third-party JS that logs \`location.href\`. Combined with #1, a single leaked URL = full sidecar compromise.

Both were intentional historical choices that pre-date the security audit. C1 + C2 (PR #801) closed the worst of it by gating /agentmux/reactive/* under auth. C3 closes the remaining vector by tightening *how* the auth surface is reached.

### CORS — strict loopback allow-list
- \`agentmux-srv/src/server/mod.rs\` — replace \`allow_origin(Any)\` with an \`AllowOrigin::predicate\` that reflects only \`http://127.0.0.1:*\` and \`http://localhost:*\`. The two legitimate callers (CEF frontend, Vite dev) are both loopback.

### Drop \`?authkey=\` on every route except /ws
- \`agentmux-srv/src/server/mod.rs\` — \`auth_middleware\` only consults \`?authkey=\` when \`req.uri().path() == "/ws"\`. The browser WebSocket API can't set custom headers, so /ws keeps the fallback. Everything else requires the \`X-AuthKey\` header.

### Frontend fetchWaveFile uses the header
- \`frontend/app/store/global.ts\` — \`fetchWaveFile\` was the only caller that used \`?authkey=\` on a non-/ws route. Switched to \`X-AuthKey\` header.

## Tests

| Test | Before | After |
|---|---|---|
| \`auth_accepts_query_param\` | asserted 200 on \`?authkey=\` for /agentmux/service | renamed \`auth_rejects_query_param_on_http_routes\` → asserts 401 |
| \`cors_headers_present\` | asserted ACAO is set | split into \`cors_reflects_loopback_origin\` (asserts reflected origin matches request) and \`cors_rejects_non_loopback_origin\` (asserts no ACAO for \`attacker.example\`) |

**14/14 server tests pass.**

## Audit findings closed

- **C3** (Critical) — \`Access-Control-Allow-Origin: *\` + \`?authkey=\` query-param bypass.

With this PR, all three app-side Criticals from the audit (C1, C2, C3) are closed.

## Version

\`0.33.791\` → \`0.33.792\`.

## Test plan

- [x] Server tests pass.
- [ ] Manual: run AgentMux, confirm UI works end-to-end (Memory bundles render → identity flow → terminal pane → agent pane streaming).
- [ ] Manual: open a file in the editor pane that triggers \`fetchWaveFile\` (e.g. images / large terminal output) — confirms the header-based auth path works.
- [ ] Manual: confirm Vite dev server (task dev) still works — Origin: localhost:5173 should be allowed by the predicate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.791"
+version = "0.33.792"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -38,7 +38,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.791"
+version = "0.33.792"
 dependencies = [
  "dirs",
  "serde",
@@ -50,7 +50,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.791"
+version = "0.33.792"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.791"
+version = "0.33.792"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/VERSION_HISTORY.md
+++ b/VERSION_HISTORY.md
@@ -2,7 +2,7 @@
 
 This document tracks the version history of AgentMux (forked from waveterm).
 
-## Latest Version: 0.33.718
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

PR #803, opened 2026-05-11
feat(live-log): Phase 3 — overlay UI (header / log / bottom action bar)

AgentA-asaf:
## Summary

Phase 3 of [SPEC_TOOL_BLOCK_LIVE_LOG_2026_05_11.md](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/SPEC_TOOL_BLOCK_LIVE_LOG_2026_05_11.md) §3.4 + §3.5. Replaces \`ToolBlock\`'s inline \`renderToolContent()\` switch with a three-slot \`ToolBlockOverlay\` (sticky header, scrollable log body, fixed bottom action bar). Branching actions move out of \`NodeHoverStrip\` into the overlay action bar.

**Stacked on #800** (Phase 1: types + reducer + tests). Will rebase to \`main\` after #800 merges.

## What's new

- \`ToolBlockOverlay.tsx\` — three-slot wrapper (header / log / actions)
- \`ToolOverlayLog.tsx\` — renders \`node.log.chunks\` when present (live log), falls back to existing per-tool result viewers (\`BashOutputViewer\`, \`DiffViewer\`, ...) when not. Auto-scroll-to-bottom with 40px stick threshold.
- \`ToolOverlayActions.tsx\` — bottom action bar: bookmark, open-in-pane, open-in-window (disabled w/ tooltip), new-agent-here (disabled w/ tooltip)
- \`_tool-overlay-portal.scss\` — three-slot flex layout, per-kind log-line styling (stdout / stderr / system)

## What's removed from \`NodeHoverStrip\` (§3.5)

The three branching button slots — \`onOpenInNewPane\` / \`onOpenInNewWindow\` / \`onNewAgentFromHere\` — are deleted from the strip. They surface in the overlay's bottom action bar instead. Strip now carries timestamp + expand + bookmark only.

\`DocumentRow\` drops the keyboard shortcuts (\`p\` / \`w\` / \`n\`) that were tied to the removed buttons; \`e\` (expand) and \`b\` (bookmark) stay.

## Behavior today (chunks not flowing yet)

[...]

PR #800, opened 2026-05-11
feat(live-log): Phase 1 — types + reducer + tests

AgentA-asaf:
## Summary

Phase 1 of `docs/specs/SPEC_TOOL_BLOCK_LIVE_LOG_2026_05_11.md` — the data-shape foundation for streaming tool output (bash stdout/stderr, edit diff hunks) into the agent pane. **No UI changes.** Chunks are accepted, deduped, and stored on `ToolNode.log.chunks`; the next phase wires the backend to emit them, and Phase 3 introduces the new overlay UI.

- New types: `ToolLogChunkKind`, `ToolLogChunk`, `ToolStreamingLog`, `ToolChunkEvent`; `ToolNode.log?` carries the live buffer.
- Stream parser: `eventToNode` returns `null` for `tool_chunk` (no DocumentNode); `parseToolChunkEvent()` normalizes the event into a chunk record for the dispatcher.
- `useAgentStream` detects `tool_chunk` and dispatches `ToolChunkAppend` to the agent-document reducer, bypassing the per-event `StreamFlush` path so chunk traffic doesn't re-trigger full-node-list memos (§3.3 of the spec).
- Reducer: new `ToolChunkAppend` command with append-only / dedup-against-last / unknown-id / non-tool-id guards and two audit events (`tool-chunk-appended`, `tool-chunk-dropped`). `StreamFlush` now preserves `log.chunks` across the running→terminal tool node replacement and flips `log.open = false` on terminal status.

## Test plan

- [x] `npx vitest run frontend/app/store/agent-document/reducer.test.ts` — 39/39 pass, including 11 new `ToolChunkAppend` tests + 3 new `StreamFlush + ToolChunkAppend` interaction tests.
- [x] `npx vitest run frontend/app/view/agent/stream-parser.test.ts` — 24/24 pass, including 4 new `tool_chunk parsing` tests.
- [x] `tsc --noEmit` — zero new errors in any touched file.
- [ ] No-UI smoke: with this PR merged, current UI renders identically because nothing reads `ToolNode.log` yet. Confirmed by reading `ToolBlock.tsx` — still keys off `node.result`.

## Spec & follow-ups

Spec lives at `docs/specs/SPEC_TOOL_BLOCK_LIVE_LOG_2026_05_11.md` (added in this PR — rides with the code per the no-doc-only-PRs rule). Phase 2 (backend emission), Phase 3 (UI overlay), and Phase 4 (stub action implementations) are tracked in §11 of that spec.

[...]

PR #804, opened 2026-05-11
feat(bashwrap): new agentmux-bashwrap crate (PR β.A — streaming bash wrapper)

AgentA-asaf:
## Summary

PR β.A of the live-log work — adds a new Rust crate `agentmux-bashwrap` that owns the PTY-backed bash runner needed for live tool-output streaming. Wrapper exists in isolation in this PR; PR β.B will wire it into Claude's PreToolUse hook + agentmux-srv's HTTP publish endpoint.

Spec: [docs/specs/SPEC_STREAMING_BASH_RUNNER_2026_05_11.md](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/SPEC_STREAMING_BASH_RUNNER_2026_05_11.md) (also added in this PR).

## What's in this PR

- **`agentmux-bashwrap/`** — new workspace member.
- **`agentmux-bashwrap exec --tool-id=<id> --b64-cmd=<base64>`** — owns a PTY (24×200, `TERM=xterm-256color`), spawns `sh -c "$decoded_cmd"` (or `cmd /C` on Win32), reads bytes line-buffered, streams each chunk to the AgentMux sidecar's WPS broker as `tool_chunk:<id>` events, and prints the aggregated output on its own stdout for Claude's native Bash tool to capture as `tool_result.content`. Truncates the model-visible blob to 50KB head + 50KB tail with an elision marker for very long outputs; the streaming log keeps the full thing.
- **`agentmux-bashwrap hook`** — reads a PreToolUse JSON payload on stdin and emits a hook response that rewrites the command from `original` to `agentmux-bashwrap exec --tool-id=X --b64-cmd=<base64 of original>`. Base64 sidesteps every shell-quoting concern. Idempotent (won't double-wrap if it sees the wrapper prefix). Pass-through on malformed input or non-Bash tools.
- **WPS HTTP client** — reads `AGENTMUX_LOCAL_URL` + `AGENTMUX_AUTH_KEY` from inherited env, posts chunks with `X-AuthKey` per [PR #801](https://github.com/agentmuxai/agentmux/pull/801). If env is missing or the sidecar is down, falls back to buffered-only mode — the command still runs to completion; only live streaming is disabled. A `kind: "system"` chunk announces the degradation.
- **`Cargo.toml` + `.bump.json`** — register the workspace member + track its version.
- **`docs/specs/SPEC_STREAMING_BASH_RUNNER_2026_05_11.md`** — full design rationale (PreToolUse command rewrite vs. MCP deny-redirect, why a dedicated crate vs. extending the hypothetical `agentmux-mcp`, channel correlation by `tool_use_id`, etc).

## What's NOT in this PR (deferred to PR β.B)

PR β.A makes the wrapper *exist*; PR β.B makes the system *use* it:

- [ ] `POST /agentmux/wps/publish` endpoint on agentmux-srv (auth-gated, mirrors the PR #801 pattern). Currently the wrapper can't actually publish — sidecar has no such route yet.
- [ ] Re-injection of `AGENTMUX_AUTH_KEY` into the claude spawn env (`subprocess.rs`). The security PR #801 intentionally removes it from agentmux-srv's process env at startup; for the wrapper to authenticate, we need to thread it back explicitly into the spawn env_vars map.
- [ ] Auto-injection of `.claude/hooks.json` PreToolUse entry by `agent_config.rs`. The hook subcommand exists but Claude won't invoke it without the registration.
- [ ] Frontend `tool_chunk:*` subscription bridge in `useAgentStream.ts`.
- [ ] Multi-chunk `ToolChunkAppend` shape for RAF coalescing in the reducer (current single-chunk shape works correctly; multi-chunk is a perf optimization for high-rate output).

Until PR β.B lands, the wrapper compiles, tests, and bundles, but isn't wired into the agent flow. Reviewable in isolation.

## Test plan

- [x] `cargo test -p agentmux-bashwrap` — **18/18 pass**:
  - Hook: rewrites simple commands, preserves multi-line + quotes via base64 round-trip, is a no-op on already-wrapped commands, passes through non-Bash tools, passes through malformed input, passes through when `command` field missing, shell-quote handles safe + unsafe + embedded-quote ids.
  - Wrapper: base64 decode round-trips arbitrary content including multi-line + quotes, rejects malformed b64, format-model-blob includes exit code + elapsed time, truncates 150KB inputs to head/tail with elision marker, passes small outputs unchanged.
  - WPS client: returns `None` without env vars, returns `None` with only endpoint set, strips trailing slash.
- [x] `cargo build -p agentmux-bashwrap` — clean.
- [ ] End-to-end smoke (deferred to PR β.B when the wrapper is actually wired into a session).

## Sequencing

- **#800** (merged) — Phase 1: reducer + types.
- **#803** (open, approved) — Phase 3: overlay UI.
- **This PR (β.A)** — wrapper binary in isolation.
- **Next (β.B)** — wire-up: HTTP publish endpoint + env re-injection + hook auto-injection + frontend bridge.
- **Then (γ)** — RAF coalescing + perf.
- **Then (δ)** — Codex / Gemini parity.

### DIFF
--- a/.bump.json
+++ b/.bump.json
@@ -25,6 +25,11 @@
       "file": "agentmux-common/Cargo.toml",
       "type": "toml",
       "path": "package.version"
+    },
+    {
+      "file": "agentmux-bashwrap/Cargo.toml",
+      "type": "toml",
+      "path": "package.version"
     }
   ],
   "lockfiles": [
@@ -42,7 +47,8 @@
       "agentmux-srv/Cargo.toml",
       "agentmux-cef/Cargo.toml",
       "agentmux-launcher/Cargo.toml",
-      "agentmux-common/Cargo.toml"
+      "agentmux-common/Cargo.toml",
+      "agentmux-bashwrap/Cargo.toml"
     ],
     "commitMessage": "chore: bump version to {{version}}"
   }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.797"
+version = "0.33.798"
 dependencies = [
  "anyhow",
  "base64",
@@ -26,7 +26,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.797"
+version = "0.33.798"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -54,7 +54,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.797"
+version = "0.33.798"
 dependencies = [
  "dirs",
  "serde",
@@ -66,7 +66,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.797"
+version = "0.33.798"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -86,7 +86,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.797"
+version = "0.33.798"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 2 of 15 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #808, opened 2026-05-12
fix(overlay): replaceChild crash on v0.33.799 — Switch instead of Show cascade

AgentA-asaf:
## Summary

Hotfix for the renderer crash hit on the v0.33.799 smoke build:

\`\`\`
Failed to execute 'replaceChild' on 'Node': The node to be replaced is not a child of this node.
\`\`\`

**Root cause**: \`ToolOverlayLog\` (shipped in #803) had 4 \`<Show>\` blocks as sibling slots inside the log container. Two of them rendered the same \`<ToolOverlayResult>\` from different predicate branches. SolidJS's reconciler saw the same component type in two sibling slots and during the running → success state transition tried to re-parent the DOM node from one Show slot to the other, calling \`replaceChild\` on a node that was no longer a child of the expected parent.

This is the same class of SolidJS reactive-reconciliation hazard as PR #708/#721/#722.

## Fix

- \`<Switch>\`/\`<Match>\` for the four branches. Switch is mutually exclusive by design — only one Match's children mount at a time, all others unmount before the chosen one mounts. No shared DOM nodes between branches; no re-parent attempt.
- Hardened the auto-scroll \`createEffect\`: \`scrollRef.isConnected\` check before mutating \`scrollTop\`. A Match flip during the same RAF window can detach the element; mutating a detached node was a secondary contributor.

## Test plan

- [x] \`tsc --noEmit\` — zero new errors in touched files.
- [ ] Smoke: extract a portable build, run a Bash tool to completion, hover/pin the overlay, verify no console errors during the running → success transition, verify structured result viewer (BashOutputViewer / DiffViewer / exit code) appears after completion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.799"
+version = "0.33.800"
 dependencies = [
  "anyhow",
  "base64",
@@ -26,7 +26,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.799"
+version = "0.33.800"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

a5af:
@codex review

PR #813, opened 2026-05-12
fix(streaming-bash): hooks at .claude/settings.json — NOT .claude/hooks.json (live-log critical)

AgentA-asaf:
## Summary

**CRITICAL HOTFIX for live-log streaming.** β.B (#809) wrote PreToolUse hooks to \`.claude/hooks.json\`. Claude Code does NOT read that file — per the docs (https://code.claude.com/docs/en/hooks.md), project hooks live in **\`.claude/settings.json\`** under the \`"hooks"\` key.

**Result on 0.33.804:** \`agentmux-bashwrap hook\` never fired. Every Bash tool ran natively, no command rewrite, no live streaming. Smoke-build confirmed: 410 \`tool_use\` events in the sidecar log, **0 mentions** of \`bashwrap\` / \`PreToolUse\` / \`tool_chunk\` / \`/wps/publish\`.

## Fix

Both paths kept in sync:

- **Rust (\`agentmux-srv/src/backend/agent_config.rs\`)**: \`build_hooks_config\` → \`build_settings_with_hooks\`. Writes \`.claude/settings.json\`, wraps hooks under \`"hooks"\` key, merges with user-supplied \`content_map["settings"]\` AND legacy \`content_map["hooks"]\`.
- **TypeScript (\`frontend/app/view/agent/agent-model.ts\`)**: same shape via \`buildSettingsWithHooks\`.

## Migration

Existing agents from prior sessions have a stale \`.claude/hooks.json\` lingering. Fresh agent sessions get the correct \`.claude/settings.json\` written. Users can safely delete the now-unread \`.claude/hooks.json\` (Claude ignores it).

## Test plan

- [x] \`cargo check -p agentmux-srv\` — clean.
- [x] \`tsc --noEmit\` — zero new errors.
- [ ] Smoke: extract 0.33.808 portable, spawn a fresh agent, ask it to run \`sleep 2 && echo a && sleep 2 && echo b\`. Expect \`a\` at t≈2s, \`b\` at t≈4s in the overlay (live streaming).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #815, opened 2026-05-12
fix(bashwrap): drop portable_pty for piped tokio::process + bash -c

AgentA-asaf:
## Summary

The wrapper crashed every Bash tool call inside Claude on Windows. Standalone repro showed the spawned `cmd /C` child exiting with `STATUS_DLL_INIT_FAILED` (0xC0000142) in ~70ms, before producing a single byte. Root cause: the wrapper drops `pair.master` immediately after cloning the reader, which per the portable-pty maintainer tears down ConPTY mid-startup ([wezterm discussion #4674](https://github.com/wezterm/wezterm/discussions/4674)). The same `CommandBuilder` pattern works in `agentmux-srv/.../shell.rs` only because there `pair.master` is moved into a long-lived `tokio::spawn` task.

Two fix paths existed:

1. **Minimal**: keep `pair.master` alive across `child.wait()` — preserves PTY (~5 lines).
2. **Larger**: drop `portable_pty` entirely for `tokio::process::Command` + `Stdio::piped()`.

This PR picks (2). The live-log feature wants line streaming, not spinner fidelity, and plain pipes sidestep the entire ConPTY lifetime class. Tradeoff acknowledged: `npm install`'s CR-only progress bars won't animate live (npm checks `isatty` before emitting them). The "log" feature is about lines anyway; PTY-only behaviors are a follow-up if/when we want them back, and we now know exactly which contract to honor.

### What changed
- `tokio::process::Command` with `Stdio::piped()` for stdout + stderr — read concurrently via `BufReader::lines()`.
- `bash -c <command>` instead of `cmd /C` — matches what Claude's native Bash tool uses, and what Claude actually sends syntactically (`[[ ]]`, `pipefail`, `2>&1`, …).
- Bash located via `$AGENTMUX_BASH` > `$BASH` > `which bash` > well-known Windows paths (Git Bash / msys2). Fail-loud if not found.
- Each stdout / stderr line published as its own `tool_chunk` with `kind` set to `"stdout"` or `"stderr"`. Stderr lines also prefixed `[stderr] ` in the aggregated `tool_result` blob.
- `tracing::info!` of streaming-relevant env vars on entry — presence-only, auth key value never logged — so the next smoke test diagnoses the `AGENTMUX_AUTH_KEY` / `AGENTMUX_LOCAL_URL` propagation gap from a single sidecar log line.
- Drops `portable-pty` dep; adds `which`.

### Retro
Full retrospective at `docs/retros/2026-05-11-live-log-streaming-wrapper-failures.md` — covers original design vs. shipped form, all three failure modes (hook discovery, ConPTY child crash, env propagation), online research findings, and 7 lessons.

## Test plan
- [x] `cargo test -p agentmux-bashwrap` — 20 tests pass
- [x] Standalone: `agentmux-bashwrap exec --tool-id=t --b64-cmd=ZWNobyBoZWxsbw` → exit 0, prints "hello"
- [x] Standalone: `agentmux-bashwrap exec --tool-id=t --b64-cmd=Z2ggYXV0aCBzdGF0dXMgMj4mMQ` → exit 0, prints real `gh auth status` output
- [ ] Smoke build 0.33.809 inside Claude pane — verify live streaming in the overlay + env presence in sidecar log
- [ ] If env vars still missing in sidecar log, walk the spawn chain (agentmux-srv → claude → bash → hook → wrapper) — follow-up PR

## Sequencing
Depends on #813 (`.claude/settings.json` hook discovery fix) landing first — without it the hook never fires regardless of wrapper correctness. This PR is the second half of the live-log streaming fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.810"
+version = "0.33.811"
 dependencies = [
  "anyhow",
  "base64",
@@ -26,7 +26,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.810"
+version = "0.33.811"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -54,7 +54,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.810"
+version = "0.33.811"
 dependencies = [
  "dirs",
  "serde",
@@ -66,7 +66,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.810"
+version = "0.33.811"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -86,7 +86,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.810"
+version = "0.33.811"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.809"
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

a5af:
@codex review

Round 4 (commit `46fdd6fa`) addresses both round-1 P2s + round-3 P2s. Reagent LGTMed in body but as COMMENTED state.

PR #821, opened 2026-05-13
fix(registry): migrator suppresses pre-v8 SqlInputError schema mismatch

AgentA-asaf:
## Summary

PR B's migrator was meant to silently skip pre-v8 per-version SQLite DBs (no \`instance_name\` column ⇒ nothing to migrate). But the suppression check only matched \`Error::SqliteFailure\`. rusqlite reports prepare-time schema mismatches as \`Error::SqlInputError\` instead, so every legacy DB fell through to "any_db_failed=true" — marker deferred, registry left empty.

End-user impact: launching a fresh portable showed an empty Continue-agent dropdown despite \`~/.agentmux/agents/\` having dozens of named agents on disk from prior versions.

## Fix

\`is_missing_column_or_table\` now matches both \`SqliteFailure\` and \`SqlInputError\` variants. Same message-prefix check ("no such column" / "no such table") so genuine corruption errors keep deferring the marker.

## Test plan

- [x] New regression test \`migrate_silently_skips_pre_v8_schema\` — creates a per-version DB with the v6 schema, verifies the migrator reports \`rows_seen=0\`, \`complete=true\`, marker written.
- [x] All 12 existing migrator tests still pass.
- [x] Discovered + verified against the host log of 0.33.827 — confirmed the exact \`"no such column: instance_name in SELECT ... at offset 11"\` error from \`SqlInputError\`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.827"
+version = "0.33.828"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #825, opened 2026-05-13
feat(agent-pane): hover-expand tool blocks + show bash output

AgentA-asaf:
## Summary

Three connected fixes to the agent pane's tool-block UX. Live-log streaming was already shipped end-to-end (PR #800+, broker side proven), but the rendering layer had two gaps that hid the output, plus the spec'd hover-expand was never wired.

### 1. \`ToolBlock\` — hover-expand wiring

\`docs/specs/tool-collapse.md\` calls for hover-to-expand with 150ms enter / 300ms leave delays. The code only implemented click-to-pin (\`expanded = () => props.pinned\`). Hovering did nothing.

Now \`onMouseEnter\` on the collapsed row triggers the overlay after 150ms; \`onMouseLeave\` cancels (or schedules collapse if expanded). Hover handlers on the portal overlay too so moving the cursor into the expanded content extends the dwell. Click-to-pin still works and persists across mouse-out.

### 2. \`claude-translator\` — prefer structured \`tool_use_result\`

Claude Code's tool_result event emits both:
- \`message.content[N].content\` — a human-readable string blob (\`<exited 0 in 0.60s>\n...\`)
- \`event.tool_use_result\` — sibling \`{ stdout, stderr, interrupted, ... }\` structured object

Translator only read the string, so \`node.result\` became \`{ content: "..." }\`. \`BashOutputViewer\`'s \`result.stdout\` check was always falsy → the output was never rendered, only the command. Now we prefer \`event.tool_use_result\` when present; fall back to the string-wrapped form for non-bash tools or older CLIs that don't emit the structured shape.

### 3. \`BashOutputViewer\` — fall back to \`result.content\`

Defense in depth: even if a result lands without \`stdout\` set, fall back to \`result.content\`. Also handle missing \`exitCode\` gracefully — the section doesn't render rather than showing "Exit code: undefined".

## Test plan

- [x] Hover over a fresh bash tool block → overlay pops up after 150ms, shows command + output
- [x] Click the tool row → overlay pins; click again → collapses
- [x] Mouse into the overlay → it stays open; mouse out → collapses after 300ms
- [x] All existing tool block unit tests still pass
- [x] \`cargo check --workspace\` green (no backend changes)

## Related

- Live-log streaming end-to-end: #800, #803, #804, #815 (already merged)
- Tool-collapse spec: \`docs/specs/tool-collapse.md\`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.827"
+version = "0.33.828"
 dependencies = [
  "anyhow",
  "base64",
@@ -27,7 +27,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.827"
+version = "0.33.828"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.827"
+version = "0.33.828"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #829, opened 2026-05-13
feat: compositor-only DnD drop-zone animation

AgentY-asaf:
## Summary

- **Two-div split** replaces the single animated placeholder div: outer `.placeholder-sizer` handles position (`translate3d`) + size (`width`/`height` set once per zone change, not animated); inner `.placeholder` handles appearance (`opacity` + `scale`, compositor-only)
- **`@starting-style`** provides the enter animation without JS — browser transitions from `opacity:0 scale(0.92)` to `opacity:1 scale(1)` on first paint (Chrome 117+ / CEF 146+)
- **Delayed unmount** (150 ms `setTimeout`) holds the SolidJS `<Show>` node alive while the `.exiting` CSS class triggers the fade-out, then removes the DOM node cleanly
- **`--accent-color-rgb`** added to `theme.scss` and all 8 theme files, enabling `rgb(var(--accent-color-rgb) / 0.2)` fill + `/ 0.5` border syntax
- **`&.animate` cleanup** — removed `width`, `height`, `background-color` from the transition; only `transform` remains — eliminates continuous layout recalculation during drag

## Why it's smoother

Before: every drag-move updated `width`/`height` via CSS transition, forcing layout at 60 fps. After: only `transform: translate3d` changes during drag (compositor thread, zero layout/paint cost).

## Test plan

- [ ] Drag a pane — drop-zone highlight fades in with a gentle scale-up
- [ ] Move drag across zones — placeholder position transitions smoothly, size jumps instantly
- [ ] Release drag — highlight fades out and scales down before DOM removal
- [ ] All 8 themes: correct translucent accent fill and border
- [ ] `prefers-reduced-motion`: instant show/hide, no transitions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #837, opened 2026-05-14
fix(dnd): hardcode placeholder-sizer transition to 120ms

AgentA-asaf:
Follow-up to [#829](https://github.com/agentmuxai/agentmux/pull/829). User-reported regression: drop-zone enter fade works, but continued drag between zones snaps instantly instead of smoothing.

## Root cause

AgentY's PR #829 wired the new `placeholder-sizer` drag-move transition to `var(--animation-time-s)`:

```scss
&.animate {
    .placeholder-sizer {
        transition-property: transform;
        transition-duration: var(--animation-time-s);
    }
}
```

But that CSS variable comes from `layoutModel.animationTimeS()`, which defaults to `0` (see `DefaultAnimationTimeS` at `agentmux-srv/.../layoutModel.ts:103`) and has **no setter** anywhere in the codebase. So `--animation-time-s` is permanently `0s` for every layout in production.

Net effect: the `&.animate .placeholder-sizer { transition: transform 0s }` rule is a no-op. The `.placeholder` inner element has its own hardcoded `transition: opacity 120ms ease-out, transform 120ms ease-out` so the enter fade-in / scale plays correctly — but the drag-move between zones is instantaneous, defeating the whole compositor-only point of the two-div split.

## Fix

Give `.placeholder-sizer` a fixed 120ms transition duration. Matches the inner `.placeholder` opacity/scale durations so the visual response is uniform across enter / drag-move / exit. `.tile-node` still uses `var(--animation-time-s)` (= 0) so pane split/close/rebalance behavior is unchanged — only the drop-zone indicator gets the fixed-duration smoothing.

## Test plan

- [ ] Drag a pane — drop-zone fades in (unchanged)
- [ ] Continue dragging across zones — placeholder now smoothly transitions between positions (was snapping before this fix)
- [ ] Release — drop-zone fades out (unchanged)
- [ ] `prefers-reduced-motion`: no transitions (the existing `@media` block's `&.animate .placeholder-sizer { transition: none }` still wins via source order)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.862"
+version = "0.33.863"
 dependencies = [
  "anyhow",
  "base64",
@@ -27,7 +27,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.862"
+version = "0.33.863"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.862"
+version = "0.33.863"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #838, opened 2026-05-14
fix(dnd): catch drops in dead spots between per-pane overlays

AgentA-asaf:
**User report**: while dragging a pane, releasing it in dead spots inside the main window (gutter between panes, edges of the layout) sometimes produces no drop indicator. On release, the system spawns a new window instead of dropping in-place.

## Root cause

Each pane's `overlay-node` is sized exactly to its tile's bounding rect, so the gap between tiles (`--gap-size-px`, plus tile-layout edges) is a true dead spot — pragmatic-dnd finds no matched drop target, fires no `onDrag` / `onDrop`, and the cross-window drag monitor interprets the absent in-window drop as a potential tear-off.

## Fix

Register a **fallback drop target on `.overlay-container`** (the parent of every per-pane overlay-node, sized to the full display area). When pragmatic-dnd's hit test resolves to the container as the innermost matched target — i.e. cursor is in a dead spot, no per-pane match — the fallback:

1. **`onDrag`** — finds the leaf whose bounding-rect center is closest to the cursor (Euclidean distance, excluding the dragged pane itself), runs the same `ComputeMove` action a per-pane `onDrag` would have. Placeholder appears, user sees the drop indicator.
2. **`onDragLeave`** — clears the pending action when cursor leaves the container entirely.
3. **`onDrop`** — clears the cross-window drag payload (prevents the tear-off path) and commits the pending action.

When the cursor IS over a per-pane overlay-node, the fallback **no-ops** (it checks `dropTargets[0] === container` before running), so per-pane behavior is untouched.

Mirrored across all three platforms (win32, darwin, linux).

## Test plan

- [ ] Drag a pane into a clear pane area → per-pane overlay highlights as before
- [ ] Drag a pane into the gap between two panes → drop indicator NOW appears (was missing)
- [ ] Drag a pane near the edge of the layout (outside any tile) → drop indicator appears for the nearest pane
- [ ] Release in a dead spot → drop applies to nearest-pane direction, NO new window
- [ ] Release outside the window → tear-off still works (cursor not over `.overlay-container`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.863"
+version = "0.33.864"
 dependencies = [
  "anyhow",
  "base64",
@@ -27,7 +27,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.863"
+version = "0.33.864"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.863"
+version = "0.33.864"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

a5af:
@codex review

PR #839, opened 2026-05-14
fix(taskfile): guard dev:serve against orphan vite on :5173

AgentX-asaf:
## Summary

When `task dev`'s spawned Vite dies (e.g. port 5173 already held by another running AgentMux dev session), the readiness check still passes against the **foreign** Vite, the script prints \"Vite ready.\", and the host attaches to that other workspace's source tree. The result: cross-tree import errors that look like the local repo is broken even though it isn't.

This guards both `dev:serve` and `dev:standalone:serve` by checking `\$VITE_PID` liveness in the wait loop and once again after, so a port collision fails fast with a clear message instead of silently degrading.

Also gitignores `src-tauri/` — leftover Tauri build directories (multi-GB cargo target output) shouldn't be trackable. The Tauri host was removed per `CLAUDE.md` (\"There is only one host\").

## How this was hit

Two AgentMux dev sessions running concurrently on the same machine (different workspaces). Workspace A's \`task dev\` ran first and held \`:5173\`. When workspace B ran \`task dev\`:

\`\`\`
Waiting for Vite on port 5173 (pid g1)...
Vite ready.                                       ← false positive
Error: Port 5173 is already in use                ← B's vite died here
    at Server.onError (.../node_modules/vite/...)
[INFO] Loading URL: http://localhost:5173?...     ← B's host loaded A's Vite
\`\`\`

B's CEF window then rendered A's frontend source, triggering an \`@lobehub/icons-static-svg\` import error overlay whose path pointed at workspace A — making the failure look like it belonged to A even though A's window was the one rendering it.

After the fix, B's run aborts immediately with:
\`\`\`
❌ Vite (pid …) exited before becoming ready — port 5173 likely held by another process
\`\`\`

## Test plan

- [x] Killed orphan Vite from workspace A → port 5173 free
- [ ] Reproduce collision: start a Vite on :5173, run \`task dev\` in a second clone → expect immediate fail-fast (not silent attach)
- [ ] \`task dev\` baseline (no collision) still works
- [ ] \`task dev:standalone\` baseline still works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,11 @@ test-results/
 backend-test/
 test/screenshots/
 
+# Legacy Tauri build artifacts. The Tauri host was removed (CLAUDE.md
+# "There is only one host"); leftover dirs from old branches/checkouts
+# can be tens of GB of cargo target output that should never get tracked.
+src-tauri/
+
 # CEF binaries (downloaded at build time by cef-dll-sys)
 agentmux-cef/cef_binary/
 agentmux-cef/resources/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.858"
+version = "0.33.859"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

AgentA-asaf:
Closing in favor of a rebased version. The original branch (agentx/dev-vite-pid-guard) was 30+ commits behind main with version-file conflicts pre-Phase-2. Cherry-picked AgentX's 2 fix commits onto current main + added a changeset; new PR coming with branch agentx/dev-vite-pid-guard-rebased.

PR #842, opened 2026-05-14
fix(layout): hold overlayTransform 150ms so placeholder exit fade is visible (#836)

### LINKED ISSUE #836: DnD placeholder exit fade rarely visible — parent .placeholder-container transform vanishes on drop
**Deferred from PR [#829](https://github.com/agentmuxai/agentmux/pull/829) / codex P3 (`TileLayout.linux.tsx:607`, same pattern in `darwin.tsx` and `win32.tsx`).**

## Issue

The compositor-only DnD drop-zone animation (PR #829) added a 150ms delayed-unmount so the inner `.placeholder` can fade out via the `.exiting` CSS class before the DOM node is removed. The delay works, but the exit fade is **rarely visible** in practice because:

1. `.placeholder-container` (the parent) gets its style from `overlayTransform`, which depends on `activeDrag`.
2. On `onDrop`, `activeDrag` flips to `false`.
3. `overlayTransform` then moves the entire container off-screen.
4. The inner `.placeholder` is still mounted with `.exiting` class, trying to fade — but its parent is already off-screen.

Net effect: the fade-out CSS plays correctly but is invisible to the user. The enter fade-in still works (no parent-transform race on the start side).

## Where it lives

`TileLayout.{win32,darwin,linux}.tsx` — all three Placeholder components share the same lifecycle pattern after PR #829's port. The issue predates the linux/darwin port (it existed in the original win32 implementation) but is now triplicated.

## Possible fixes

Two paths suggested by codex:

1. **Freeze the parent transform** during the exit window — capture the last `overlayTransform` value when `activeDrag` flips, hold it for 150ms, then release.
2. **Keep the overlay shown** (`activeDrag` flag delays its own clear) until the exit timer completes — moves the coordination into the drag state machine.

Either approach needs to keep the existing semantics intact (an in-flight new drag must cancel any pending exit fade and reset cleanly).

## Why this isn't gating PR #829

P3 severity (lowest), pre-existing on win32 prior to this PR, fix requires touching the drag state machine that's out of scope for an animation polish PR. PR #829 ships the compositor-only drag-move improvement and the enter animation; the rarely-visible exit fade is acceptable for now and can land separately.

🤖 Filed from PR #829 round 5 codex deferment.

AgentA-asaf:
Closes #836.

## Summary

After the compositor-only DnD placeholder work landed (#829 → #837 → #838), the inner `.placeholder` element's exit fade rarely played in practice. Root cause: on `onDrop`, `activeDrag` flips to `false`, which makes `overlayTransform()` recompute and move the entire `.placeholder-container` off-screen (top = `2 * height`). The inner `.placeholder` is still mounted with `.exiting` for 150ms trying to fade, but its parent is already off-screen.

## Fix

In `TileLayoutComponent` (all three platform files), cache the last in-drag `overlayTransform` and serve it from a new `effectiveOverlayTransform()` accessor during a 150ms hold window that matches the Placeholder's exit-fade duration. Pass the accessor into `OverlayNodeWrapper` so its `overlay-container` also stays put during the hold — but pointer-events flips to `none` IMMEDIATELY when `activeDrag` goes false, so the held overlay never intercepts clicks during the hold window.

A new drag during the hold resets cleanly via the `active` branch of the `createEffect`, clearing the timer and the hold flag.

Applied identically to `win32`, `darwin`, `linux` per PR #838's port.

## Test plan

- [ ] Drag a pane to a new drop zone; observe placeholder fade-out on release
- [ ] Quickly start a second drag mid-fade; placeholder reappears at new target without flicker
- [ ] Click outside drop zones during the 150ms hold window — clicks pass through (pointer-events: none)
- [ ] No regressions to drag-over/drag-leave/drop computation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.866"
+version = "0.33.867"
 dependencies = [
  "anyhow",
  "base64",
@@ -27,7 +27,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.866"
+version = "0.33.867"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.866"
+version = "0.33.867"
 dependencies = [
  "dirs",
  "serde",
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.866"
+version = "0.33.867"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -88,7 +88,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.866"
+version = "0.33.867"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-bashwrap/Cargo.toml
+++ b/agentmux-bashwrap/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-bashwrap"
-version = "0.33.866"
+version = "0.33.867"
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

a5af:
@codex review

PR #843, opened 2026-05-14
feat(workflows): Agent inspector + workflowrun subscription (Phase 1.5 PR 3)

### LINKED ISSUE #830: Workflows: wire frontend workflowrun:<id> subscription (Phase 1 PR-4 polish)
**Deferred from PR #755 / round 7 (codex P2 on `frontend/app/view/workflows/workflows-model.ts:181`).**

## Context

PR #755 (Workflows Phase 1) switched the run handler to a **placeholder row + spawn UPDATE** pattern to avoid the 5s RPC timeout truncating long workflow drains. The synchronously-inserted `running` row solves the frontend's "row not found" race, but Phase 1 has no `workflowrun:<id>` event subscription wired, so the Runs panel shows the placeholder and never sees it transition to `done` / `failed` unless the user manually refreshes the workflow.

The RFC #753 §6 PR-4 polish explicitly lists this as deferred.

## What to build

- Frontend subscribes to `workflowrun:<id>` via the existing WPS event bus (same mechanism live-log streaming uses for `tool_chunk` — see #804, #809).
- On each event: dispatch into a `workflow-run-state/reducer` (slice #10, matching the master reducer-stack pattern from #756–#765).
- On `RunDone` / `RunFailed`: refresh the runs list automatically.

## Why this is a follow-up, not a #755 fixup

- Documented as deferred in the original RFC.
- Phase 1.5 (`docs/specs/SPEC_UNIFIED_AGENT_TYPES_2026_05_13.md` §5.3) introduces slice #10 — the per-run reducer — which is the natural home for this subscription. Doing it first means re-doing the wiring.
- Current behavior: placeholder row visible, stays \`running\` until manual refresh. Workable but ugly.

## Where the publish lives

\`agentmux-srv/src/server/workflow_handlers.rs:204-211\` publishes events as \`WaveEvent { event: "workflowrun:<run_id>", ... }\`. Frontend just needs to subscribe.

🤖 Filed from PR #755 review round 7 deferment.

### LINKED ISSUE #835: Workflows Agent block: frontend still writes forge_agent_id, backend reads agent_ref
**Deferred from PR #834 / codex P1 (`workflows/executor/blocks/agent.rs:59`).**

## Context

PR [#834](https://github.com/agentmuxai/agentmux/pull/834) wired the workflow Agent block to call `run_agent(agent_ref, task, tx)` from the unified `agents` module. The block now reads `agent_ref` from `node.data` (with `AgentRef::default()` as the fallback).

The **frontend** still writes the pre–Phase 1.5 shape:
- `frontend/app/view/workflows/block-registry.ts` — Agent block defaults persist `forge_agent_id`
- `frontend/app/view/workflows/workflows-view.tsx` — Inspector edits the same field

Net effect: every UI-created or pre-existing Agent node launches a **blank claude** instead of the user's configured (identity / memory / named-agent) selection.

## Resolution path

This is the natural scope of **Phase 1.5 PR 3** (frontend inspector + `workflowrun:<id>` subscription, ref discussion [#832](https://github.com/agentmuxai/agentmux/discussions/832)). PR 3 will:

1. Replace the Agent block's `forge_agent_id` field in the inspector with separate Identity / Memory / Instance-name pickers.
2. Migrate the persisted block config from `{ forge_agent_id }` → `{ agent_ref: { identityId, memoryId, instanceName } }`.
3. (Optional) Reader-side compat in `agent.rs`: if `agent_ref` is absent but `forge_agent_id` is present, log a deprecation warning and treat as blank — so legacy saved workflows surface clearly rather than silently launching blank.

PR #834 ships the runner integration; this issue tracks the UI side that makes it user-driveable. Until PR 3 lands, the workflow Agent block is functional for programmatic / API-defined workflows (where `agent_ref` is set directly) but launches blank from UI-edited graphs.

🤖 Filed from PR #834 codex review deferment.

AgentA-asaf:
Closes #830, closes #835. Phase 1.5 PR 3 per `docs/specs/SPEC_UNIFIED_AGENT_TYPES_2026_05_13.md` §6 row 3. Tracking discussion: #832.

## Summary

- **#835 — Agent block migrated from `forge_agent_id` to `agent_ref`.** Block-registry default data now ships `{ agent_ref: { identityId, memoryId, instanceName, workingDirectory }, task }`. Inspector replaces the single text field with separate identity / memory `<select>` pickers (populated via `ListIdentityBundlesCommand` + `ListMemoriesCommand`) plus an optional instance-name input for named-agent continuation. Backend `blocks/agent.rs` logs a deprecation warning when a node still carries `forge_agent_id` so legacy workflows surface visibly rather than silently launching blank.
- **#830 — Frontend subscribes to `workflowrun:<run_id>` events.** Model wires `waveEventSubscribe` on `run()`, applies `BlockDone` / `BlockError` into a per-block result cache, and refreshes the runs list on `RunDone` / `RunFailed`. Inspector now shows last-run response + cost for the selected Agent block. Subscription tears down on completion, `newWorkflow`, or `dispose`.
- Removes the Phase 1 \`TODO Phase 1 PR-4\` comment that warned the placeholder row never transitioned to done/failed without manual refresh.

## Scope notes

- Hover-expand tool stream rendering is deferred to Phase 2 polish per spec §5.2 — this PR ships final-result rendering only.
- Slice #10 reducer migration is PR 4 — the view model still folds events directly; PR 4 moves the same logic into a pure reducer.
- Memory picker shows non-blank bundles only (mirrors `AgentLaunchModal`'s pattern; the runtime injection layer for workflow-block memory is the same path the pane uses).

## Test plan

- [x] `cargo test -p agentmux-srv workflows::executor::blocks::agent` — 3 pass (incl. new `legacy_forge_agent_id_falls_back_to_default_ref`)
- [x] `tsc --noEmit` clean on the three changed frontend files
- [ ] Open workflows pane → drop Agent block → identity dropdown populates with bundles
- [ ] Run workflow → run-panel row transitions from \`running\` → \`done\` / \`failed\` without manual refresh
- [ ] Select Agent block after run completes → "Last run" panel shows response + cost

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.867"
+version = "0.33.868"
 dependencies = [
  "anyhow",
  "base64",
@@ -27,7 +27,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.867"
+version = "0.33.868"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.867"
+version = "0.33.868"
 dependencies = [
  "dirs",
  "serde",
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.867"
+version = "0.33.868"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -88,7 +88,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.867"
+version = "0.33.868"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.868"
+version = "0.33.869"
 dependencies = [
  "anyhow",
  "base64",
@@ -27,7 +27,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.868"
+version = "0.33.869"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.868"
+version = "0.33.869"
 dependencies = [
  "dirs",
  "serde",
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.868"
+version = "0.33.869"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -88,7 +88,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.868"
+version = "0.33.869"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-bashwrap/Cargo.toml
+++ b/agentmux-bashwrap/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-bashwrap"
-version = "0.33.867"
+version = "0.33.868"
 edition = "2021"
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

AgentA-asaf:
Addressed both codex P2s:

- **Subscribe-race + refresh-race**: `workflow_handlers.rs` now persists the final `WorkflowRun` row BEFORE publishing the terminal `workflowrun:<id>` event. The frontend's `refreshRuns` on a terminal event now always sees the final state instead of the stale `running` placeholder.
- **Late-subscribe race for fast workflows**: Added `maybeBackfillFromTerminalRun` in `workflows-model.ts` that runs after the initial `refreshRuns` and, if the just-completed row is already terminal, populates the per-block result cache from `block_states` so the inspector still shows per-block output even when events fired before subscription.

47 workflow tests pass, frontend tsc clean. Bumped to 0.33.869.

@a5af could you run `@codex review` so codex re-checks?

PR #841, opened 2026-05-14
feat(title): "Window - Tab - AgentMux" format, no version, reactive

AgentX-asaf:
## Summary

Replaces the static `AgentMux ${ver} - ${initialTab.name}` title (which had a known TODO to react to tab changes — `app-init.ts:485,590`) with the format requested:

```
{Window Name} - {Tab Name} - AgentMux
```

- **No version anywhere** in the OS title (still in StatusBar `v{version}` chip and the About modal).
- **Identical across all release types** — `task dev`, `task dev:standalone`, portable ZIP, AppImage, and the future MSIX / macOS packages.
- **Frontend-only.** CEF already mirrors `document.title` verbatim to `SetWindowTextW` (Windows) / CEF Views `set_title` (macOS / Linux X11 / Linux Wayland) at `agentmux-cef/src/client/mod.rs:107-139` — no Rust edits needed.

## Window Name resolution

Same three-tier rule the bottom-right `InstancePanel` already uses for the open-windows list (so the panel and the taskbar title can never disagree):

1. `WaveWindow.meta["window:displayname"]` — user-set (rename gesture exists today: open InstancePanel via the version chip → double-click a row, or F2 → 64-char inline editor)
2. `workspace.name` — when the window has a named workspace
3. `Window N` — 1-indexed positional fallback

## Reactive

A single `createEffect` (installed once at the end of `initWave`) re-runs when any of these changes:

- Active tab changes (`atoms.activeTabId()` → `tab.name`)
- Window display name changes (rename via InstancePanel)
- Workspace assignment changes (`atoms.workspace().name`)
- Window's position in `openWindowEntriesAtom` shifts

## Files

- **NEW** `frontend/util/window-title.ts` — pure `resolveWindowName` + `formatWindowTitle` helpers + `DISPLAY_NAME_META_KEY` constant
- **NEW** `frontend/util/window-title.test.ts` — 16 unit tests covering the resolution chain, trimming, separator collisions, 64-char names, missing tab name
- **REFACTOR** `frontend/app/statusbar/InstancePanel.tsx` — inline three-tier resolution (lines 142-153) replaced by `resolveWindowName(...)` call
- **REWIRE** `frontend/app-init.ts` — three imperative `document.title = ...` writes (lines 306, 485, 590) replaced by one `installWindowTitleEffect(windowId)` call; module-level `appVersion` removed (was title-only)
[...]

PR #852, opened 2026-05-14
feat(srv): WaveObjUpdate broadcast bridge (Phase 1: workspace events)

AgentX-asaf:
## Summary

Adds an event-bus → frontend broadcast bridge so any reducer event automatically reaches the frontend WOS cache, eliminating the class of bug where mutation RPC handlers forgot to attach `WaveObjUpdate` to their response.

**Fixes the user-reported symptom:** workspace renames don't propagate to the OS title (added in #841) or the bottom-right InstancePanel.

## Why per-handler fixes weren't the right answer

`UpdateWorkspace` (`agentmux-srv/src/server/service.rs:1274-1307`) dispatches the rename → reducer → SQLite → publishes `WorkspaceRenamed` to `srv_events_tx`, then returns `success_empty()` instead of an inline `WaveObjUpdate`. The response-broadcast loop at `service.rs:39-52` only iterates response-attached updates; with nothing attached, no `\"waveobj:update\"` WS event ever fires. Frontend WOS stays stale.

Same shape will hit any future mutation RPC where the author forgets the convention. Per-handler fix patches one site; the bridge closes the class.

## Bridge architecture

NEW \`agentmux-srv/src/server/wave_obj_bridge.rs\` — a tokio task that:

1. Subscribes to \`srv_events_tx\` (same channel as \`persist_subscriber\` and \`disk_writer\`)
2. Translates each event variant into one or more \`WaveObjUpdate\` payloads
3. Pushes them to the existing \`event_bus.broadcast_event(...)\` path — **same downstream the response loop uses**

Frontend dedup at \`wos.ts:272\` (already in place via \`curValue.value.version >= update.obj.version\`) absorbs the now-doubled broadcasts for handlers that still attach updates, so this lands additively without disruption.

\`\`\`
        reducer → emits Event → srv_events_tx ─┬─► persist subscriber → SQLite (existing)
                                               ├─► disk writer → JSONL log (existing)
                                               └─► wave_obj_bridge → WaveObjUpdate → broadcast (NEW)
\`\`\`

## Phase 1 scope (this PR)

Mapping coverage in \`dispatch_event()\`:

| Reducer event | Action |
|---|---|
| \`WorkspaceRenamed\` | fetch + broadcast workspace as \"update\" |
| \`WorkspaceMetaUpdated\` | fetch + broadcast workspace as \"update\" |
| \`WorkspaceCreated\` | fetch + broadcast workspace as \"update\" |
| \`WorkspaceDeleted\` | broadcast oid only as \"delete\" |
| (everything else) | silent skip via catch-all \`_ => {}\` arm |

Catch-all makes the bridge future-proof for new event variants (silent skip until Phase 2 expands the mapping). Phase 2 (tab/block/window/layout) is mechanical expansion of the same match.

## Soundness (per spec §11 research log)

- **§11.1 — Subscribe ordering safe:** reducer mutates in-memory state synchronously before \`srv_events_tx.send()\` (\`service.rs:1283-1305\`). Bridge reads post-event state with no race.
- **§11.2 — Lock structure:** \`WaveStore\` uses \`std::sync::Mutex<Connection>\` (sync, not async). Bridge holds the lock briefly per event (one SQLite query); never \`.await\`s while held.
- **§11.3 — Compound events:** Phase 1's workspace events are single-object. Phase 2 will need to handle \`TabCreated\` etc. fetching both Tab + parent Workspace.
- **§11.4 — Version field universal:** all 6 \`WaveObj\` types have \`version: i64\`. Frontend dedup at \`wos.ts:272\` works universally.
- **§9 — Performance negligible:** \`srv_events_tx\` already has 2 subscribers; one more is cheap. Per-event work is one SQLite query.
- **Failure mode:** \`broadcast::error::RecvError::Lagged(n)\` is logged loudly with the skipped count so silent WOS divergence is visible to operators.

## Test plan

Pre-flight (done in PR):
- [x] \`cargo check -p agentmux-srv\` — clean (warnings unchanged from main)
- [x] \`npx tsc --noEmit\` — error count unchanged at 21 (all pre-existing)

Manual (post-merge or via fresh \`task dev\`):
- [ ] Rename a workspace → OS taskbar title updates immediately to \`<new name> - <tab> - AgentMux\`
- [ ] Open InstancePanel (click version chip) → row for that window shows the new workspace name
- [ ] Rename window via InstancePanel double-click → still works (regression — confirms the existing \`UpdateObjectMeta\` path coexists with the bridge)
- [ ] Open two windows, rename workspace → both windows' titles + both panels update (multi-client)

## Companion specs (in this commit)

| Spec | Purpose |
|------|---------|
| \`SPEC_OBJ_UPDATE_BRIDGE_2026-05-14.md\` | Full design (this PR) |
| \`SPEC_REACTIVE_WORKSPACE_SYNC_2026-05-14.md\` | Root-cause analysis |
| \`SPEC_PERSISTENCE_LAYER_ANALYSIS_2026-05-14.md\` | \"Keep SQLite, deferred\" — sanity-check before any persistence change |
| \`SPEC_WINDOW_INSTANCE_NAMING_CLEANUP_2026-05-14.md\` | Independent follow-up: UI copy / atom name cleanup |
| \`SPEC_WAVE_TO_MUX_RENAME_2026-05-14.md\` | Wave Terminal branding rename (deferred mid-implementation; tracked in #851) |

## Out of scope

- Phase 2 (tab/block/window/layout event coverage)
- Phase 3 (retiring \`success_with_updates(...)\` calls now that the bridge covers them)
- Bridge metrics (lag count, dispatch latency) — current logging suffices

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.33.892"
+version = "0.33.893"
 dependencies = [
  "anyhow",
  "base64",
@@ -27,7 +27,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.33.892"
+version = "0.33.893"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.33.892"
+version = "0.33.893"
 dependencies = [
  "dirs",
  "serde",
@@ -67,7 +67,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.33.892"
+version = "0.33.893"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -88,7 +88,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.33.892"
+version = "0.33.893"
 dependencies = [
  "agentmux-common",
  "async-stream",
--- a/agentmux-bashwrap/Cargo.toml
+++ b/agentmux-bashwrap/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-bashwrap"
-version = "0.33.892"
+version = "0.33.893"
 edition = "2021"
 description = "AgentMux streaming bash wrapper + PreToolUse hook helper"
 authors = ["AgentMux Corp"]
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-cef"
-version = "0.33.892"
+version = "0.33.893"
 description = "AgentMux Host — Desktop app with bundled Chromium"
 authors = ["AgentMux Corp"]
 edition = "2021"
--- a/agentmux-common/Cargo.toml
+++ b/agentmux-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "agentmux-common"
-version = "0.33.892"
+version = "0.33.893"
 edition = "2021"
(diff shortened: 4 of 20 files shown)


a5af:
@codex review

PR #797, opened 2026-05-11
feat(linux): enable CEF Wayland window transparency

AgentU-asaf:
## Summary

End-to-end CEF Wayland window transparency on Linux. CSS layer composes with the desktop; the AgentMux frame becomes alpha-aware when `window:transparent: true` is set (already the user-side toggle that exists from the Tauri era).

The hard work was a new CEF source patch — committed as `68e0dc668` on `a5af/cef` branch `agentmux/7680-drag-rightclick-and-transparency`. Chad Nelson's prior patch (`b921ffe18`) handled the Views / Aura layer but never propagated to the renderer's `cc::LayerTreeHost`, so the compositor clamped every pixel's alpha to 1.0 even after a `kTranslucent` Aura widget and an empty `wl_surface` opaque region were arranged. The new commit calls `WebContents::SetPageBaseBackgroundColor(SK_ColorTRANSPARENT)` from `CefBrowserViewImpl::WebContentsCreated` when the resolved settings background has alpha=0. That broadcasts to every RenderViewHost via the existing `ExecutePageBroadcastMethod` path and flips `has_transparent_background_` to true so cc emits actual ARGB pixels.

Empirically verified:
- `cefsimple --hide-frame` against the rebuilt libcef.so renders the test page's transparent body over whatever is behind the window (text visible against the underlying terminal).
- `WAYLAND_DEBUG=1` on AgentMux 0.33.763 captures **0** `wl_surface_set_opaque_region` calls and a `WL_SHM_FORMAT_ARGB8888` (`format=0`) buffer on the 1200×800 main-window surface.

## AgentMux changes

Three byte-flips that activate the transparency cascade in the bundled libcef:

| File | Setting | Before | After |
|------|---------|--------|-------|
| `agentmux-cef/src/main.rs:475` | `CefSettings::background_color` | `0xFF000000` | `0x00000000` |
| `agentmux-cef/src/app.rs:445` | `BrowserSettings::background_color` | `0xFF000000` | `0x00000000` |
| `agentmux-cef/src/app.rs:369` | `--background-color` CLI switch | `"ff222222"` | `"00000000"` |

The main window already returns `is_frameless=true` (`app.rs:520`), so the other CEF-side requirement is already in place.

## Bundled libcef.so

Built from the patched `a5af/cef` fork:

- `b921ffe18` "Support transparent window in Views framework (follow up issue #2315)" — Chad Nelson's patch (Views / Aura side).
[...]

PR #869, opened 2026-05-15
feat(linux): CEF Wayland transparency + fixes (rebase of #797)

AgentA-asaf:
Rebases AgentU's #797 onto current main with all 8 author commits preserved (5 code, 3 docs). Used `-X ours` strategy to auto-resolve version-file conflicts.

🤖 Rebased per RFC #857 changesets workflow.

### DIFF
--- a/.changesets/1778842317-feat-linux-enable-cef-wayland-window-transparency-iterative--kiys.md
+++ b/.changesets/1778842317-feat-linux-enable-cef-wayland-window-transparency-iterative--kiys.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(linux): enable CEF Wayland window transparency + iterative fixes (rebase of #797)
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

a5af (approved):
Reagent flagged 2 P1s that are real but design-level tradeoffs:
1. `--disable-lcd-text` applied unconditionally — should gate on transparency setting. Fix needs settings-aware CEF command-line construction (settings load may be async at that point); not a quick patch.
2. Default --window-opacity=0.45 causes a translucent flash for users with window:transparent=false until AppSettingsUpdater overrides. The SCSS comment explains this was chosen specifically to avoid Chromium's tile-cache baking opaque pixels permanently.

Both need AgentU's design input. Filing as follow-up issues. Approving to land the rebased commits — they're net-positive over current main (the Wayland transparency feature itself works).

PR #865, opened 2026-05-15
devx(phase 2): changesets workflow — zero version conflicts in feature PRs (RFC #857)

AgentA-asaf:
## Summary

Adopts a lightweight, self-hosted changesets workflow. Feature PRs no longer touch version files; they add a uniquely-named `.changesets/<ts>-<slug>.md`. A dedicated release PR consumes pending changesets, bumps once, updates VERSION_HISTORY, deletes consumed files.

Phase 2 of multi-agent version coordination spec (`docs/specs/SPEC_MULTI_AGENT_VERSION_COORDINATION_2026_05_15.md`, RFC #857).

## Why self-hosted, not `@changesets/cli`?

The npm `@changesets/cli` is built for npm monorepos. We're a Cargo workspace + a single npm root + lockfiles. Two ~50-line shell scripts + a Taskfile task are cleaner than wedging a monorepo tool sideways. The existing `scripts/bump-wrapper.sh` (lockfile-sync workaround for bump-cli quirks) is reused by `scripts/release.sh`.

## Author workflow

```bash
task changeset -- patch "fix(auth): cancel in-flight session on selection swap"
# OR: task changeset -- minor "feat(...): description"
# OR: task changeset -- major "breaking change description"
```

Creates `.changesets/<unix-ts>-<slug>.md` like:
```
---
type: patch
---

fix(auth): cancel in-flight session on selection swap
```

Commit it alongside code changes. **Do NOT run `bump patch`** — release owns version bumps now.

## Release workflow (separate PR)

```bash
task release           # processes pending changesets
                       #   - picks highest bump type (major > minor > patch)
                       #   - runs scripts/bump-wrapper.sh
                       #   - appends to VERSION_HISTORY.md
                       #   - deletes consumed changesets
                       #   - stages everything (does NOT commit)
git diff --staged      # review
git commit -m "chore: release v<X.Y.Z>"
git push -u origin agenta/release-vX.Y.Z
```

## Conflict surface

| Phase | Per-PR conflict files |
|---|---|
| Pre-Phase-1 | 9 (package.json + 6 Cargo.toml + Cargo.lock + package-lock.json) |
| Post-Phase-1 | 4 (package.json + root Cargo.toml + 2 lockfiles) |
| **Post-Phase-2 (this PR)** | **0** (unique `.changesets/<id>.md` per PR) |

Parallel agents never collide on changeset filenames — timestamp prefix guarantees uniqueness.

## Reagent dependency

[...]

PR #873, opened 2026-05-15
fix(taskfile): quote desc strings with colons (RFC #857 Phase 2 regression)

AgentA-asaf:
Phase 2 (PR #865) added `changeset`, `release`, `package:local` tasks with unquoted `desc:` strings containing inner colons (`Usage: task changeset -- patch ...`). YAML parsed the inner colon as a mapping separator and the whole file failed.

`task --list` and `task package` both errored: `yaml: line 26: mapping values are not allowed in this context`.

Caught when trying to produce a fresh portable build of main @ 0.33.898.

Fix: single-quote the desc strings.

Verified `task --list` now enumerates all tasks cleanly.


a5af:
@codex review

PR #866, opened 2026-05-15
fix(taskfile): guard dev:serve against orphan vite on :5173 (rebase of #839)

AgentA-asaf:
## Summary

Rebases AgentX's #839 onto current main (Phase 1+2 era). The original branch was 30+ commits behind and accumulated 9-file version conflicts; this branch cherry-picks just the 2 author commits onto fresh main and adds a changeset describing the change.

Original PR: #839 (closed)
Original author: AgentX-asaf

## Commits preserved

1. `da236d87` fix(taskfile): guard dev:serve against orphan vite on :5173
2. `31f0b91e` fix(taskfile): pre-check :5173 instead of broken \$! liveness

The Taskfile.yml `dev:serve` task now pre-checks for an orphan Vite on port :5173 before launching, replacing the broken \$! PID-liveness check that didn't work under mvdan/sh task runner.

## Outstanding finding to address

Reagent flagged a real P1 on the original #839 about a race condition between the pre-check and the curl probe (a second process could bind :5173 in the gap). The original author should address this in a follow-up commit. Not blocking this rebase.

## Changeset

`.changesets/1778841552-fix-taskfile-guard-dev-serve-against-orphan-vite-on-5173-reb-lygl.md` — patch-level, will be consumed by the next release PR.

🤖 Rebased with [Claude Code](https://claude.com/claude-code) using the new RFC #857 changesets workflow

PR #856, opened 2026-05-15
feat(srv): migrate UpdateWindowMeta through reducer (Phase E.5.x, closes #855)

AgentX-asaf:
## Summary

Closes #855. Replaces the wcore-direct fallback for `OTYPE_WINDOW` that PR #852 shipped as a temporary workaround. Window meta updates now flow through the standard reducer → persist + bridge path, fitting cleanly into the Phase E migration arc and master spec §8.15 (srv-event subscriber idempotency contract).

## Why

PR #852 surfaced the gap: `service.rs:308` `UpdateObjectMeta` handler routes WORKSPACE/TAB/BLOCK through the reducer (Phase E.5.3) but Window fell into the `_ => other` arm that uses wcore-direct. With no event published, the bridge couldn't see the change → frontend WOS cache stayed stale → window display-name renames "reverted" visually until app restart. PR #852 added an inline-update workaround in the wcore-direct branch; this PR retires that workaround by completing the migration.

## Flow

```
service.rs UpdateObjectMeta(OTYPE_WINDOW, meta_patch)
  ↓
Command::UpdateWindowMeta { window_id, meta_patch }                    ← NEW (ipc.rs)
  ↓ dispatch_to_reducer
reducer/window.rs::handle_update_window_meta                           ← NEW
  ↓ emits
Event::WindowMetaUpdated { window_id, meta_patch, version }            ← NEW (ipc.rs)
  ↓ publish_events → srv_events_tx
  ├─► persist_subscriber → apply_window_meta_updated → wstore SQLite   ← NEW
  └─► wave_obj_bridge → broadcast WaveObjUpdate (frontend WS)          ← NEW arm
```

Mirrors the existing `UpdateWorkspaceMeta` shape exactly — pass-through reducer (no in-memory state mutation; reducer state doesn't track window meta), persist subscriber merges into `wstore`, bridge broadcasts the updated Window to the frontend WOS cache.

## Files

| File | Change |
|------|--------|
| `agentmux-common/src/ipc.rs` | NEW `Command::UpdateWindowMeta` + `Event::WindowMetaUpdated` |
| `agentmux-srv/src/reducer.rs` | dispatch arm for `Command::UpdateWindowMeta` |
| `agentmux-srv/src/reducer/window.rs` | NEW `handle_update_window_meta` |
| `agentmux-srv/src/persist_subscriber.rs` | NEW `apply_window_meta_updated` + match arm + `event_kind_name` entry |
[...]

PR #875, opened 2026-05-15
fix(launcher): cover UpdateWindowMeta + WindowMetaUpdated in match arms

AgentA-asaf:
Main is currently broken: `cargo check -p agentmux-launcher --release` fails with 3× E0004 because the launcher's match arms don't cover `Command::UpdateWindowMeta` (added by #856) or `Event::WindowMetaUpdated`.

Caught when trying to produce a fresh portable build today after the RFC #857 rebase batch (#866-#869) + #873 landed.

- `ipc/server.rs:651`: added the same 'srv-pipe command sent to launcher pipe by mistake' message used by Layout commands
- `reducer/mod.rs:87`: joined to the existing Layout-or pattern, message generalized
- `event_log.rs:245`: added WindowMetaUpdated to the version-extractor or-pattern

`cargo check -p agentmux-launcher --release` clean after fix.

Will unblock `task package` builds on main.

### DIFF
--- a/.changesets/1778846479-fix-launcher-add-updatewindowmeta-windowmetaupdated-match-ar-zc9h.md
+++ b/.changesets/1778846479-fix-launcher-add-updatewindowmeta-windowmetaupdated-match-ar-zc9h.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #877, opened 2026-05-15
feat(agent-pane): persist reducer state to disk for full-restore on reopen

AgentA-asaf:
## Summary

Fixes the bug where reopening AgentMux shows a long conversation with broken tables, orphaned tool calls, and truncated content. Root cause: the agent-document reducer rebuilds from only the last 200 NDJSON lines on reopen, dropping older content and corrupting nodes that cross the window boundary.

Spec: `docs/specs/SPEC_AGENT_PANE_STATE_PERSISTENCE_2026_05_15.md`.

## Solution

Write a sidecar JSON snapshot of the reducer state on pane close + every 30s during lifetime. On reopen, restore from snapshot (full replace via new `HistoryRestored` command) instead of partial NDJSON replay.

```
~/.agentmux/data/blocks/<blockid>/output.state.json
{ "schemaVersion": 1, "savedAt": "<iso>", "nodes": [...] }
```

Compatible with the TanStack Solid Virtual virtualizer — restores accept full `nodes[]` immediately, recycled rows re-measure on first paint, no stutter.

## Verification

- vitest: 3418 pass (4 new HistoryRestored tests)
- `cargo check --workspace --release` clean
- `tsc --noEmit` clean

## Backwards compat

Existing blocks (no snapshot) fall back to current 200-line replay behavior — no regression. First close after upgrade writes a snapshot; second reopen restores fully. Self-heals.

## Hard-kill loss bound

~30s max (the dirty-flag interval). The NDJSON file is unaffected by all of this and remains the source of truth for cold-state recovery.

🤖 Generated with Claude Code per RFC #857 follow-up work

### DIFF
--- a/.changesets/1778869319-feat-agent-pane-persist-reducer-state-to-disk-so-reopen-rest-luos.md
+++ b/.changesets/1778869319-feat-agent-pane-persist-reducer-state-to-disk-so-reopen-rest-luos.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): persist reducer state to disk so reopen restores full conversation (no more truncated tables / orphaned tool calls)
--- a/agentmux-srv/src/backend/rpc_types.rs
+++ b/agentmux-srv/src/backend/rpc_types.rs
@@ -376,6 +376,8 @@ pub const COMMAND_PANE_OPEN: &str = "pane.open";
 // App API Tier 1 — blockfile pagination commands
 pub const COMMAND_BLOCKFILE_LINE_COUNT: &str = "blockfile:line_count";
 pub const COMMAND_BLOCKFILE_READ_RANGE: &str = "blockfile:read_range";
+pub const COMMAND_BLOCKFILE_READ_STATE: &str = "blockfile:read_state";
+pub const COMMAND_BLOCKFILE_WRITE_STATE: &str = "blockfile:write_state";
 
 // App API Tier 1 — session archival commands
 pub const COMMAND_SESSION_ARCHIVE: &str = "session:archive";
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

Round 3 fix pushed at 4cedbbb3: `setHistoryOffset(0)` after snapshot restore replaced with `BlockfileLineCountCommand` query → `offset = max(0, total - snapshot.nodes.length)`. `loadOlder` cursor is now preserved across snapshot restore.

a5af:
@codex review

Round 4 fix pushed at 0a6961b3: snapshot now records `historyOffset` directly from the hook accessor at save time (actual NDJSON line index). Restore uses `snapshot.historyOffset` verbatim — no longer derived from node count. Fixes the streaming over-estimation reagent flagged at round 3.

a5af:
@codex review

Round 5 fix pushed at 494a6253: `HistoryRestored` now prepends with id-dedup instead of full-replace. Fixes the race where `StreamFlush` from `useAgentStream` could land before snapshot read resolves — live arrivals are preserved; snapshot becomes the older prefix. 161 reducer tests pass.

a5af:
@codex review

Round 6 pushed at 8037798b. Addressing the three round-5 finds:

**P2 line 175 (serialize writes)** — FIXED in 8037798b. `writeSnapshotNow()` now chains through a shared `inFlightSnapshot` Promise so the 30s interval write and the on-close cleanup write serialize in call order. `historyOffset` is captured synchronously at call time, not read inside the queued async block.

**P1 line 150 (preserve live nodes on restore)** — already fixed in round 5 (494a6253) at the reducer layer, not the dispatch site. `HistoryRestored` in `frontend/app/store/agent-document/reducer.ts:89` was changed from full-replace to prepend-with-id-dedup. Live `StreamFlush` arrivals during the snapshot read window are preserved; snapshot nodes prepend as the older prefix. Test `preserves live nodes that arrived during the snapshot read window` covers this. The dispatch site at `useHistoryPagination.ts:150` is unchanged because the merge semantics live in the reducer.

**P1 line 163 (lines above highWaterMark)** — partly inaccurate, partly a documented v1-deferred limitation. The `historyOffset` is set from `snapshot.historyOffset` (not 0) since round 4 (0a6961b3); `loadOlder` is preserved across snapshot restore. The "lines appended while pane closed" gap (NDJSON growth beyond `snapshot.highWaterMark`) IS real but is explicitly v1-deferred Phase 4 per `docs/specs/SPEC_AGENT_PANE_STATE_PERSISTENCE_2026_05_15.md` §4.2 — the schema reserves the `highWaterMark` field so Phase 4 can read the gap range and replay through `StreamFlush` on reopen. v1 ships without it; the "background agent ran while pane closed" gap is acknowledged in the spec.

a5af:
@codex review

Round 7 pushed at bbd54156. Addressing the three round-6 finds:

**P2 line 177 (persist empty snapshots after /clear)** — FIXED in bbd54156. Removed the `nodes.length === 0` early return so `UserClear` overwrites the existing sidecar. The on-close cleanup now gates on the `dirty` flag (same flag the 30s interval already uses), so panes that never observed a state change do not fabricate empty snapshots that would clobber a prior session sidecar. `dirty` flips on any reactive state change including `HistoryRestored`.

**P1 line 150 (preserve live nodes on restore)** — this is a repeated find from round 5. The dispatch at `useHistoryPagination.ts:150` is safe because the reducer behavior changed in round 5 (494a6253). `frontend/app/store/agent-document/reducer.ts:89-122` `HistoryRestored` now prepends snapshot nodes onto current state with id-dedup (mirroring `HistoryLoaded`), preserving any `StreamFlush` arrivals from the race window. The test `preserves live nodes that arrived during the snapshot read window` in `reducer.test.ts:86-100` covers this case. Please trace into the reducer file when re-reviewing this line.

**P1 line 163 (snapshot hides untouched lines)** — partly addressed, partly v1-deferred per spec:
- `historyOffset` is restored from `snapshot.historyOffset` (not 0), so `loadOlder` remains usable across snapshot restore — fixed in round 4 (0a6961b3). The first post-upgrade open writes a snapshot with `historyOffset = total - 200`, and `loadOlder` covers everything below.
- The "lines appended while pane closed" gap (`snapshot.highWaterMark` < current NDJSON line count on reopen) is acknowledged as v1-deferred Phase 4 in `docs/specs/SPEC_AGENT_PANE_STATE_PERSISTENCE_2026_05_15.md` §4.2. The schema already reserves `highWaterMark` so the Phase 4 gap-replay can read `[highWaterMark, currentTotal)` and dispatch through `StreamFlush` on reopen. The v1 ships without this — it is a documented limitation, not a regression of the current behavior (which is identical 200-line replay on reopen).

PR #816, opened 2026-05-12
feat(named-agents): launch modal Continue dropdown + v8 schema + RPCs (#94)

AgentA-asaf:
## Summary

Adds a **"Continue an existing agent"** dropdown to the launch modal so users can pick a previously-named agent and pick up where it left off — same working directory, same identity bundle, same memory bundle.

Spec: [`docs/specs/SPEC_NAMED_AGENT_CONTINUATION_2026_05_12.md`](https://github.com/agentmuxai/agentmux/blob/agenta/named-agent-continuation/docs/specs/SPEC_NAMED_AGENT_CONTINUATION_2026_05_12.md) (rides with this PR — no doc-only PR).

#### Phase 0 — v8 schema (`agentmux-srv/src/backend/storage/migrations.rs`)
Three additive columns on `db_agent_instances` + a partial index for the dropdown query.

```sql
ALTER TABLE db_agent_instances ADD COLUMN instance_name TEXT NOT NULL DEFAULT '';
ALTER TABLE db_agent_instances ADD COLUMN working_directory TEXT NOT NULL DEFAULT '';
ALTER TABLE db_agent_instances ADD COLUMN display_hidden INTEGER NOT NULL DEFAULT 0;
CREATE INDEX idx_agent_instances_name_recent
  ON db_agent_instances(instance_name, started_at DESC)
  WHERE display_hidden = 0 AND instance_name != '';
```

Idempotent (catches `duplicate column`). `AgentInstance` struct + all `wstore` SELECT / INSERT / map paths updated. Three new methods: `instance_list_named(limit)`, `instance_set_hidden(id, hidden)`, `instance_get_by_name(name)`.

#### Phase 1 — RPCs + spawn wire-up
- `listnamedagents` → `Vec<NamedAgentRow>` joins instance rows with definition + identity + memory bundle names so the frontend renders the dropdown without follow-ups. 200-row default, 1000 cap.
- `hidenamedagent` → soft-delete (`display_hidden = 1`); publishes `namedagents:changed`. Row + working dir stay on disk for audit + recovery.
- `createagentinstance` handler persists `instance_name` + `working_directory` from the launch-modal payload.
- Frontend `RpcApi` bindings, `AgentInstance` + `NamedAgentRow` types.

#### Phase 2 — launch modal UI (`AgentLaunchModal.tsx` + `agent-model.ts`)
- "Continue an existing agent" select above the name field. Auto-hides when no past agents of this definition (empty state = today's modal, zero regression).
[...]

PR #879, opened 2026-05-16
fix(launch-modal): defensive fallback for blank identity/memory labels

AgentA-asaf:
## Summary

Defensive frontend fallback for the "Continue an existing agent" dropdown rendering empty identity/memory columns as visible blanks (`Maks ·  ·  · 5m ago`). Renders `(ambient creds)` / `(vanilla CLI)` for missing fields the same way the backend was supposed to.

## The bug

The named-agent dropdown row at `AgentLaunchModal.tsx:270` rendered:
```ts
\`${row.instance_name} · ${row.identity_name} · ${row.memory_name}${row.started_at ? ` · ${formatRelative(row.started_at)}` : ""}\`
```

When `identity_id`/`memory_id` are empty (the normal state for an agent created before any Identity exists in the DB), the backend handler at `forge_handlers.rs:1208-1225` is supposed to substitute `"(ambient creds)"` / `"(vanilla CLI)"`. But cargo's incremental cache can keep a pre-#816 object file even after `cargo clean -p agentmux-srv`, so the deployed binary doesn't always contain those sentinel strings (confirmed against v0.33.900 srv: `grep -c "ambient creds"` returns 0).

Result: bare `·` separators with no text between them — looks like the data is corrupt or the field is silently broken.

## The fix

Compute the row label in JSX with explicit fallbacks for empty/whitespace strings:

\`\`\`tsx
const parts = [
    row.instance_name,
    row.identity_name?.trim() || "(ambient creds)",
    row.memory_name?.trim() || "(vanilla CLI)",
];
if (row.started_at) parts.push(formatRelative(row.started_at));
return <option value={row.instance_id}>{parts.join(" · ")}</option>;
\`\`\`

- **Doesn't break the working case** — if the backend already substituted, the value passes through (`row.identity_name = "(ambient creds)"` survives `.trim() || ...`).
- **Fixes the regression case** — empty strings, whitespace, undefined all fall through to the sentinel.
- **Robust against future backend regressions** — anyone shipping a new version of the handler can't accidentally re-introduce the blank.

## Context

Phase 1 of `docs/specs/SPEC_RUNTIME_IDENTITY_AUTOBIND_2026_05_15.md` (included in this PR). The spec covers:
- **Phase 1 (this PR)**: display fix.
- **Phase 2**: runtime AuthSession reuse so a CLI's OAuth login during an active pane auto-creates an Identity + binds to the agent instance. Adds `SecretRef::AmbientCli` variant and `wstore::instance_update_identity`.
- **Phase 3**: banner ("Bind your creds") for already-orphaned agents.
- **Phase 4**: multi-provider extension (deferred).

The full spec rides with Phase 1 per the no-doc-only-prs convention; the implementation for Phases 2-3 will be a separate PR.

## Test plan

- [x] Typecheck clean (`npx tsc --noEmit` on the touched file)
- [ ] Visual: a named agent with empty identity_id renders as `<name> · (ambient creds) · (vanilla CLI) · <time>` in the launch modal dropdown
- [ ] Visual: a named agent with populated identity/memory still renders the populated names (no regression)

### DIFF
--- a/.changesets/1778910213-fix-launch-modal-defensive-fallback-for-blank-identity-memor-ixw7.md
+++ b/.changesets/1778910213-fix-launch-modal-defensive-fallback-for-blank-identity-memor-ixw7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(launch-modal): defensive fallback for blank identity/memory labels
--- a/docs/specs/SPEC_RUNTIME_IDENTITY_AUTOBIND_2026_05_15.md
+++ b/docs/specs/SPEC_RUNTIME_IDENTITY_AUTOBIND_2026_05_15.md
@@ -1,0 +1,270 @@
+# SPEC: Runtime Identity Auto-Bind & Named-Agent Display Fix
+
+**Status:** Draft / proposed
+**Date:** 2026-05-15
+**Author:** AgentA
+**Related:** `SPEC_PRE_LAUNCH_OAUTH_FLOW_2026_05_14.md` (PR #850), `SPEC_AGENT_PANE_STATE_PERSISTENCE_2026_05_15.md` (PR #877)
+
+---
+
+## 1. Problem
+
+Two related issues hit the same user flow:
+
+### 1.1 Bug A — Display
+
+The "Continue an existing agent" dropdown in the launch modal renders rows like:
+
+```
+Maks ·  ·  · 5m ago
+```
+
+with a visibly empty identity and memory column. Root cause:
+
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

Phase 1 PR for the named-agent display fix + full spec doc. Rebased clean onto origin/main (HEAD d59793bb). Single code change: defensive fallback in AgentLaunchModal.tsx:268-280. Please review.

PR #880, opened 2026-05-16
fix+feat(dev): status-bar window count + task dev:local for ephemeral version bumps

AgentA-asaf:
## Summary

Two related improvements for the test/dev experience:

### 1. Status-bar window count fix

`StatusBar.tsx:92` previously rendered `windowInstanceNumAtom` (per-window ordinal — window #2 showed `(2)`, window #5 showed `(5)`). Swap to `windowCountAtom` so all windows display the same number — the total of active AgentMux windows. The existing `<Show when={windowCount() > 1}>` gate preserves the hide-when-1 behavior. Drop the now-unused `windowInstanceNumAtom` import.

Spec: `docs/specs/SPEC_STATUS_BAR_WINDOW_COUNT_2026_05_16.md`.

### 2. New `task dev:local`

Under the RFC #857 changeset workflow, feature PRs no longer bump version files — `package.json` / `Cargo.toml` stay pinned across many merges. Two consequences:
- Dev/portable builds share the same version label across merges, so "which merge does this build correspond to?" is unanswerable from the version alone.
- Cargo's incremental cache can serve stale `.o` files because `CARGO_PKG_VERSION` doesn't change. We hit this twice during #877/#878 portable iteration — once shipped without `"(ambient creds)"` sentinel strings, once a hard-link cache fooled `cargo clean -p`.

`task package:local` already solves this for portables. Add the analogous `task dev:local`:
- In-place `bump patch` (or `minor`/`major` via CLI args).
- Run `task dev`.
- EXIT trap restores `package.json` + lockfiles + `Cargo.toml` on Ctrl+C or crash, byte-for-byte.
- No git mutation, no pushed bump commits.

CLAUDE.md updated with both `:local` variants and a `Why :local variants exist` explainer.

## Diff

- `frontend/app/statusbar/StatusBar.tsx`: 1-line atom swap + drop unused import.
- `scripts/dev-local.sh`: new file, mirror of `scripts/package-local.sh`.
- `Taskfile.yml`: new `dev:local` entry alongside `package:local`.
- `CLAUDE.md`: table + explainer updated.
- `docs/specs/SPEC_STATUS_BAR_WINDOW_COUNT_2026_05_16.md`: spec for change #1.

## Test plan

- [x] Typecheck clean.
- [ ] Window count: open second window, both status bars show `(2)`. Close one, parenthesis hides.
- [ ] `task dev:local`: starts dev with bumped version label; Ctrl+C restores working tree to original version.
[...]

PR #878, opened 2026-05-16
fix(agent-pane): cascade detection + dispatchIfRegistered migration

AgentA-asaf:
## Summary

Bundles cascade detection AND the soft-dispatch fix, after the v0.33.900 reproduction confirmed Shape B per [LIFECYCLE_DISPATCH_LEAK_2026_05_15.md](docs/analysis/LIFECYCLE_DISPATCH_LEAK_2026_05_15.md) §8.3: a documentAtom subscriber synchronously unmounts the pane during a StreamFlush's projection setter, the slot vanishes mid-dispatch, and the next dispatch (StreamFlushObserved into agent-pane-state) throws.

## What's in this PR

**Cascade detection (logging only):**
- Per-setter granularity in `agent-pane-state-store.ts`: refactored 8 sequential `if (a !== b) proj.X(b)` lines into a single `proj()` helper that captures the first cascade-triggering setter name.
- Single-setter check in `agent-document-store.ts` around the documentAtom write.
- Both log a one-line `CASCADE_DETECTED ...` warning when the slot disappears mid-dispatch.

**Soft-dispatch fix:**
- Adds `dispatchIfRegistered(blockId, command, source)` to both pane stores. Returns `[]` silently if the slot is gone instead of throwing.
- Throwing `dispatch()` stays as the contract for synchronous-body register-order checks — a missing slot there is still a bug we want to fail loud.

**Migrations (22 sites):**
- `useAgentStream.ts`: 12 async-context sites (RAF callback, fileSubject subscription handler, accepted-message subscription, watchdog setInterval, cleanup-time post-await dispatches).
- `useAgentCommands.ts`: 3 sites (2x RPC `.catch()` rejection handlers, 1x setTimeout pending-expiry).
- `useHistoryPagination.ts`: 6 post-await dispatches (defense in depth on top of the existing `mounted` flag).
- `agent-view.tsx`: 1 site (auth controller's `onLoginSuccess` callback fired by backend events).
- `browser-model.ts:442`: RAF callback now checks `if (this.closed) return;` (matches every other IPC handler in the file).

**Deliberately kept throwing (4 sites):** synchronous-body dispatches and first-tick onMount dispatches where the slot is guaranteed registered.

## Tests

New `agent-pane-state-store.test.ts` with 5 regression tests covering both contracts. **306 reducer + store tests pass.** Typecheck clean (touched files).

## Confirmation evidence

From v0.33.900 host log:
```
[...]

PR #882, opened 2026-05-16
fix(dev): restore launcher IPC connection in task dev via parent-process check

AgentA-asaf:
## Summary

Restore launcher event delivery to the frontend in `task dev` mode by replacing the stale `is_dev_build_exe` guard with a parent-process identity check. Fixes two visible symptoms with the same root cause:

1. **Status-bar window count desynchronized** — each window stuck at its boot-time count. PR #880's atom swap (`windowInstanceNum` → `windowCount`) was a no-op fix because the atom itself never updated post-boot in dev. This PR makes the atom actually update.
2. **Opacity slider missing in dev** — gated on `entry.windowId`, which is populated by the `BackendWindowIdRegistered` event arriving at the frontend reducer. Without IPC, the event never arrives, gate stays false, slider hides.

## The bug

`agentmux-cef/src/main.rs:295-316` had:

```rust
let _launcher_ipc = if agentmux_common::is_dev_build_exe(&host_exe_dir) {
    None
} else {
    runtime.block_on(launcher_ipc::connect_to_launcher(app_state.clone()))
};
```

The comment justified the skip with "task dev has no launcher process anyway." That was true before `SPEC_LAUNCHER_DEV_INTEGRATION_2026-05-13.md` made `task dev` invoke the host through the launcher (production-parallel layout in `dist/cef-dev/`). After that spec landed, the path-only guard wrongly skipped legitimate IPC in dev — but the comment was never updated. So `WindowOpened` and `BackendWindowIdRegistered` events from the live launcher had no host subscriber, and never reached the renderer.

Same stale guard at line 312-316 for srv IPC.

## The fix

Parent-process identity check. Host connects when:

```
AGENTMUX_LAUNCHER_PIPE is set
AND (parent process is `agentmux-launcher.exe` OR it's a production build)
```

Truth table (from spec §5.3):

| Scenario | Env set? | dev-build? | Parent is launcher? | Connect? |
|---|---|---|---|---|
| Production portable / installed | yes | no | yes | ✓ |
| `task dev` post-launcher-integration | yes | yes | yes | **✓ (was ✗ — this is the fix)** |
| Dev host launched standalone | no | yes | no | ✗ |
| Dev host inheriting env from sibling pane shell | yes | yes | no | ✗ (guard preserved) |

The original isolation concern (a dev host inheriting env from a parent agent pane's shell) is preserved — that host's parent is `cmd.exe` / a shell, not `agentmux-launcher`, so the parent-process check rejects it.

## Files

- `agentmux-cef/src/parent_process.rs` (new) — Windows: Toolhelp32 snapshot for parent PID, `QueryFullProcessImageNameW` for exe stem, case-insensitive compare to `"agentmux-launcher"`. Non-Windows returns `None` (Phase 7 cross-platform parity not yet shipped).
- `agentmux-cef/src/main.rs` — register `mod parent_process`, replace both guards at line 295-316.
- `agentmux-cef/Cargo.toml` — add `Win32_System_Diagnostics_ToolHelp` feature to `windows-sys`.
- `docs/specs/SPEC_DEV_MODE_LAUNCHER_IPC_2026_05_16.md` — full investigation, design, edge cases, test plan.

## Test plan

- [x] `cargo build --release -p agentmux-cef` — compiles clean.
- [ ] Smoke `task dev`: open 3 windows, every status bar shows `(3)`.
- [ ] Smoke `task dev`: click version chip, opacity slider renders for each entry in InstancePanel.
- [ ] Regression: `task package` portable still connects to launcher IPC (production path unchanged).
- [ ] Regression: a manually-spawned dev host (no parent launcher, no env var) does not connect.

## Relationship to in-flight PRs

- **#880** (window-count atom swap): forward-progress; this PR makes its visible benefit actually appear.
- **#878** (cascade detection): unrelated. Different reducer-layer bug.

### DIFF
--- a/.changesets/1778916437-fix-dev-restore-launcher-ipc-connection-in-task-dev-via-pare-gkjf.md
+++ b/.changesets/1778916437-fix-dev-restore-launcher-ipc-connection-in-task-dev-via-pare-gkjf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dev): restore launcher IPC connection in task dev via parent-process check
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -61,6 +61,7 @@ windows-sys = { version = "0.59", features = [
     "Win32_System_Ole",
     "Win32_System_SystemInformation",
     "Win32_System_ProcessStatus",
+    "Win32_System_Diagnostics_ToolHelp",
     "Win32_UI_Shell",
     "Win32_System_Com",
     "Win32_Security",
--- a/agentmux-cef/src/main.rs
+++ b/agentmux-cef/src/main.rs
@@ -320,21 +320,18 @@ fn main() {
     // srv events (workspace / tab / block lifecycle) to every
     // top-level renderer via the JS bridge. Renderer-side handler
     // (`window.__agentmux_srv_event`) lands in E.2c.5b. Non-fatal
-    // if absent: a standalone host (no parent launcher, no
-    // `AGENTMUX_SRV_PIPE_PATH`) runs without the bridge and the
-    // frontend uses the legacy waveobj:update path.
-    //
-    // Same env-isolation guard as launcher_ipc above — a dev build
-    // inheriting `AGENTMUX_SRV_PIPE_PATH` from a parent AgentMux pane
-    // would bridge its srv events into the parent's renderer fan-out.
-    // Parent-process check (replaces the older path-only guard that
-    // over-fired in `task dev` after SPEC_LAUNCHER_DEV_INTEGRATION
-    // made `task dev` invoke the launcher) — see
-    // SPEC_DEV_MODE_LAUNCHER_IPC_2026_05_16.md.
-    let _srv_ipc = if should_connect_launcher {
-        runtime.block_on(srv_ipc::connect_to_srv(app_state.clone()))
-    } else {
+    // if absent: `AGENTMUX_SRV_PIPE_PATH` is only set on the srv
+    // child by the launcher (`agentmux-launcher/src/srv_spawner.rs`),
+    // not on the host spawn — so today the host never has the env
+    // var and `connect_to_srv` short-circuits to None at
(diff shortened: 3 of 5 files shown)


a5af:
@codex review

a5af:
@codex review

Round 2 fix at 150a2594: accept both `agentmux-launcher` (Cargo bin name, used in dev) AND `agentmux` (user-facing name in portable / installed builds where package-portable.sh:38 renames the binary). Maintained as a slice `ACCEPTED_PARENT_STEMS`. Compile-tested clean. Spec §5.1 updated to document the rename rationale you correctly flagged.

a5af:
@codex review

Round 3 at 7f8d6b85: dropped QueryFullProcessImageNameW entirely, use PROCESSENTRY32W.szExeFile from the Toolhelp32 snapshot instead. szExeFile is filename-only, no path, so MAX_PATH does not apply. Two-pass snapshot (PID lookup → parent szExeFile) for clarity. Your P2 is addressed.

a5af:
@codex review

Round 4 at 4c9e040f: three reagent P2s addressed (mixed-case .exe stripping via lowercase-once approach, spec §5.1 + §5.2 brought back in sync with the actual implementation). Compile clean. Reagent re-review pending.

a5af:
@codex review

Round 5 at 81077e10: dropped redundant eq_ignore_ascii_case (reagent P2), updated all remaining parent_process_name references in spec §6-9 to parent_is_agentmux_launcher (reagent P2). Compile clean.

a5af:
@codex review

Round 6 at dca2abd1: spec status flipped Draft → Implemented; stale srv-pipe comment in main.rs:319-325 rewritten to describe the standalone-host case instead of the obsolete task-dev-has-no-launcher claim. Both reagent P2s addressed.

a5af:
@codex review

Round 7 at 4becead1: descoped srv IPC guard change — restored the original is_dev_build_exe check there (the change was a no-op since AGENTMUX_SRV_PIPE_PATH is never set on host spawn). Launcher IPC fix unchanged. Spec §11 documents the follow-up needed. Both reagent + codex previously flagged the no-op.

PR #884, opened 2026-05-16
diag(live-log): temporary tracing for tool_chunk delivery path

AgentA-asaf:
## Summary

Diagnostic-only PR to find where live-log chunks are dropping despite all components reporting healthy. User reports tool overlay shows 'Running...' then dumps full result at end — no live streaming. 6-check static-analysis session exhausted without identifying the gap.

This PR adds gated tracing (\`target: \"live-log-diag\"\`) at the three suspect points so the next repro pinpoints which step breaks:

| Component | Where | What logs |
|---|---|---|
| WPS publish endpoint | \`agentmux-srv/src/server/mod.rs::handle_wps_publish\` | tool_chunk publish receipt: event, scopes, persist, 120-char data preview |
| Broker publish | \`agentmux-srv/src/backend/wps.rs::Broker::publish\` | matched_routes count, has_subs_for_event, silent-drop case when inner.client is None |
| Frontend handler | \`frontend/app/view/agent/useAgentStream.ts\` tool_chunk handler | console.log on entry (op/toolId/contentLen) + console.warn on each early-return reason |

## How to read the resulting log

After running a Bash tool in the rebuilt portable, grep \`live-log-diag\` and \`[live-log-diag]\` against the host + srv logs:

\`\`\`
muxlog host '\[live-log-diag\]'
grep live-log-diag ~/.agentmux/logs/agentmuxsrv-v0.33.X-windows.x64.log.YYYY-MM-DD
\`\`\`

Expected flow if working:
1. srv: \`wps publish received\` (one per chunk)
2. srv: \`route match summary\` with \`matched_routes >= 1\` and \`has_subs_for_event = true\`
3. host: \`[live-log-diag] tool_chunk received\` for each chunk

Whatever step is missing or has the wrong values tells us the gap.

## Strip plan

Once the bug is identified and the real fix lands, strip the three diag blocks. They're scoped to \`event == \"tool_chunk\"\` so production noise is bounded, but they shouldn't ship long-term.

## Test plan

- [x] Builds clean (\`cargo build --release -p agentmux-srv\`, \`tsc --noEmit\`)
- [ ] Build a portable, run a Bash tool, capture log entries, identify gap

### DIFF
--- a/.changesets/1778937152-diag-live-log-temporary-tracing-for-tool-chunk-delivery-path-9yez.md
+++ b/.changesets/1778937152-diag-live-log-temporary-tracing-for-tool-chunk-delivery-path-9yez.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+diag(live-log): temporary tracing for tool_chunk delivery path
--- a/agentmux-srv/src/backend/wps.rs
+++ b/agentmux-srv/src/backend/wps.rs
@@ -315,10 +315,35 @@ impl Broker {
 
         let client = match &inner.client {
             Some(c) => c,
-            None => return,
+            None => {
+                // Live-log diagnostic.
+                if event.event == "tool_chunk" {
+                    tracing::warn!(
+                        target: "live-log-diag",
+                        event = %event.event,
+                        scopes = ?event.scopes,
+                        "broker has no client — event dropped before delivery",
+                    );
+                }
+                return;
+            }
         };
 
         let route_ids = Self::get_matching_routes(&inner, &event);
+        // Live-log diagnostic — log how many routes the tool_chunk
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

Round 2 at fdd10e48: switched the wps_publish diag to metadata-only (op/kind/tool_id-prefix/content_len) — no payload preview. Addresses your P2 about credential leakage from Bash tool output.

a5af:
@codex review

Round 3 at de8c8e14: same metadata-only treatment for the 3 frontend console.warn calls — was logging full data/event object, now logs op/kind/tool_id-prefix/content_len keys only. Symmetric to round 2 srv fix.

PR #885, opened 2026-05-16
diag(live-log): reducer-side tool-chunk audit logs

AgentA-asaf:
Follow-up to #884. The earlier diag confirmed the **entire transport chain works** (bashwrap publishes → broker routes with persist=1024 → frontend handler receives chunks). But user still sees no live render. This patch adds the missing piece: surface the reducer's **tool-chunk-appended** and **tool-chunk-dropped** audit events to console so we can tell whether chunks land in state or get dropped by findToolIndex.

Strip with #884 once the underlying bug is identified.


a5af:
@codex review

PR #888, opened 2026-05-17
feat(live-log): PTY-backed tool streaming + auto-expand inline panel

AgentA-asaf:
## Summary

Two layered fixes that together deliver real-time visibility of bash tool output in the agent pane. End-to-end verified with `stream-test.sh`: five `echo + sleep 1` lines now arrive in the reducer at 1.05 s intervals and render live in the auto-expanded panel.

### Transport (`agentmux-bashwrap`) — PTY + DSR + ANSI

- Re-introduced `portable_pty` for the child shell so external programs see `isatty(stdout) == 1` and glibc stays line-buffered. Without this, every libc-stdio program (`grep`, `awk`, `npm`, `cargo`, …) block-buffered to 4 KB or process exit → chunks arrived at the frontend in a burst at EOF.
- **DSR handshake**: bash queries the headless PTY with `\x1b[6n` at startup and blocks on stdin for the cursor-position reply. We strip the query and write back `\x1b[1;1R` via the master writer. Without this, the wrapper hung forever — the real cause of the β.A wedge, not portable_pty's master-handle lifetime as the original retro suggested. Pinned via standalone repro (`SPEC_LIVE_LOG_PTY_REWORK_2026_05_16.md` §12).
- **ANSI / OSC strip** on the chunk path: removes terminal init bytes (window title, cursor visibility, CSI resets) so the plain-text chunk list isn't garbled.
- **Manual PATH prepend** (`/usr/bin`, `/usr/local/bin`, `/mingw64/bin`) derived from bash's location instead of `-l`. Cuts bash cold start from ~1.3 s to ~150 ms because `/etc/profile` is no longer sourced.
- **Pipe fallback** retained for environments without a usable PTY.
- **ConPTY lifetime contract** correctly enforced: pair (master + slave) held in scope across `child.wait()`, dropped after the child reaps. No grace-window timer on the publisher drain (see `feedback_no_timers_or_delays.md`).

### UI (frontend tool block) — inline auto-expanding panel

- **Phase A**: drop the `<Portal>` and all positioning math (overlayRect, overlayUp, measure(), findScrollParent, getAncestorZoom, scroll-tracking + resize effects). Same hover+pin behavior, rendered inline under the summary row.
- **Phase B**: panel auto-expands while `status ∈ {running, pending_approval, failed}`. Pin remains the explicit override; hover still works as peek for completed-success tools. Closes the visibility gap where the portal overlay was unmounted during the streaming window itself (mount-storm caused by hover cycles).
- **Instant hover** (0 ms enter / 0 ms leave). Safe because the summary row and the panel share the same `.agent-tool-block` element; mouseleave only fires when the cursor truly exits the whole block.
- **Live-tail** in the collapsed row stays for the user-collapsed-mid-run edge case.

### Diagnostic cleanup

Stripped the `[live-log-diag]` instrumentation that pinned each layer of the chain during investigation. Only retained signal: the genuine `tool-chunk-dropped` warning in `agent-document-store` (real ordering-bug indicator, not a transient).

## Specs

- `docs/specs/SPEC_LIVE_LOG_PTY_REWORK_2026_05_16.md` — transport rework, with §12 documenting the V1/V2 standalone repro (`agentmux-pty-repro/` was used during investigation, not committed)
- `docs/specs/SPEC_TOOL_AUTO_EXPAND_PANEL_2026_05_16.md` — UI design; references VS Code's `runInTerminalTool.ts` as comparison

## Test plan

- [x] β.A wedge-regression: `agentmux-bashwrap exec` with `echo hello` → exit 0 in <200 ms, no `STATUS_DLL_INIT_FAILED`
- [x] Empirical streaming: 3-iter sleep loop with external `date` → chunks 1 s apart, exit 0
- [x] End-to-end through Maks: pinned overlay shows `step N $(date)` lines tick in 1 s apart
- [x] Auto-expand panel during `stream-test.sh` → panel opens with tool card, lines append live, collapses on success
- [x] Hover on completed tool → instant expand/collapse via inline panel
- [ ] Run `npm install` or similar libc-stdio command for full external-program streaming verification


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #902, opened 2026-05-18
diag(browser-pane): instrument title/favicon flow end-to-end

AgentA-asaf:
## Summary

Browser pane favicon + page title appear stuck at the static fallback in v0.33.899 (likely regression from PR #876). Adding diagnostic logging at the two unobserved layers — renderer-side wire receipt (pre block-id filter) and blockframe render — so every step of the chain is visible in \`muxlog srv '[browser-pane'\`.

After this lands, smoke flow:

\`\`\`
muxlog srv '[browser-pane' &
# open browser pane, navigate to github.com
\`\`\`

Expected sequence per navigation:

1. \`emit-title-change\` / \`emit-favicon-urls\` — CEF host
2. \`title-change arrive\` / \`favicon-urls arrive\` — renderer wire (new)
3. \`dispatch type=TitleChanged|FaviconUrlsReceived\` — reducer entry
4. \`state-write key=title|faviconUrl\` — projection setter
5. \`header-render favUrl=...\` — blockframe memo recompute (new)
6. \`favicon-load ok|fail\` — \`\` mount (new)

First silent step or stale value identifies the bug.

## Test plan

- [ ] Build portable, open browser pane → navigate to https://github.com
- [ ] \`muxlog srv '[browser-pane' | head\` shows the 6-step chain
- [ ] If chain incomplete: report the last log line we see — that's the boundary where the bug lives
- [ ] If chain complete but header still shows globe + "Browser": separate UI memo issue, blockframe diff needed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #903, opened 2026-05-18
diag(browser-pane): per-instance vmId to detect stale viewmodel refs

AgentA-asaf:
Continuing diag from #902. Chain ends correctly at state-write but header-render reads stale value forever — suggests stale viewmodel ref. Adding per-instance `__diagVmId` to the BrowserViewModel + logging it from both the writer side (already prefixed in this.diag) and the reader side (blockframe's header-render log).

After this lands, smoke and check:
- If `state-write vm=ABC` and `header-render vm=XYZ` show different IDs for the same blockId, the bug is a stale viewmodel ref (blockframe holds an older instance).
- If they match, the bug is in reactivity (memo not subscribing).


a5af:
@codex review

a5af:
@codex review

PR #926, opened 2026-05-19
fix(term): remove writeInFlight guard from echo fast path — eliminate sporadic keypress jitter

AgentY-asaf:
## Summary

- **Root cause found:** \`scheduleRafWrite()\` blocked its fast path with \`!this.writeInFlight\`, meaning any keypress echo arriving while a large PTY write was in progress got queued behind the RAF cycle (≤16ms extra delay). This produced the "sometimes fast, sometimes slow" jitter pattern during active terminal output.
- **Fix:** remove \`&& !this.writeInFlight\` from the fast-path condition. When \`rafBuffer\` is empty, xterm.js's internal write serialisation guarantees ordering — the small write lands after the in-flight batch with no display issues.
- **Instrumentation:** three perf marks (\`term-keypress\`, \`term-echo-render\`, \`term-raf-write\`) visible in CEF DevTools Performance tab and the Perf HUD. Slow-write log threshold lowered 8ms → 4ms.
- **Report:** \`docs/terminal-input-latency-report.md\`

## Test plan

- [ ] Type rapidly while a command streams output — jitter should be gone
- [ ] CEF DevTools trace: \`term-echo-render\` p99 < 4ms during output
- [ ] No new scroll flicker (RAF batching for large writes is unchanged)
- [ ] \`[raf-write] SLOW\` entries appear at ≥4ms (down from 8ms threshold)

### DIFF
--- a/.changesets/1779148800-fix-term-echo-latency-writeflight-bypass-k8m2.md
+++ b/.changesets/1779148800-fix-term-echo-latency-writeflight-bypass-k8m2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): remove writeInFlight guard from small-data fast path in scheduleRafWrite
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

AgentY-asaf:
@codex review

PR #933, opened 2026-05-19
refactor(storage): rename db_identities → db_identity_bundles (v11)

AgentA-asaf:
## Summary

Closes [AUDIT_SQLITE_SYSTEMS §8.1](docs/specs/AUDIT_SQLITE_SYSTEMS_2026_05_19.md) — the schema-naming drift where v7 created tables under bare `db_identities` / `db_memories` names but the rest of the codebase + UI used "identity bundle" / "memory bundle." User direction (relayed from RFC tracking): _"lets do A .. since that's the clean robust solution .. lets make it clean since its relatively greenfield at the moment."_

- New `run_forge_v11_migrations`: idempotent `ALTER TABLE … RENAME` for `db_identities` → `db_identity_bundles` and `db_memories` → `db_memory_bundles`. SQLite ≥ 3.25 auto-updates the FK reference in `db_identity_bindings`, so the binding cascade-delete still works (covered by a new test).
- `run_forge_v7_migrations` now guards its legacy-name DDL/seed/shadow-migrate block on `db_identity_bundles` not yet existing — prevents re-creating empty old tables alongside the renamed ones on every subsequent startup. The `ALTER COLUMN` on `db_agent_instances` stays outside the guard (idempotent on its own, unrelated to the rename).
- `run_forge_v1_migrations` extracted (the original base `db_forge_agents` CREATE) so tests can stage a pre-v7 schema and exercise the legacy-name path in isolation.
- All `wstore.rs` queries + doc comments updated to bundle names.
- `db_identity_bindings` is **NOT** renamed — its name was already bundle-consistent (a binding binds an identity bundle to a provider account; the surrounding object is the binding, not the bundle).

## Test plan

- [x] `cargo test -p agentmux-srv backend::storage::` — 103 passed (incl. 4 new v11 tests: renames-legacy-to-bundle, idempotent, fresh-install-no-op, preserves identity_bindings FK cascade)
- [x] `npx vitest run frontend/app/view/agent/auth/auth-state.test.ts` — 50 passed
- [ ] Reagent review
- [ ] Codex review
- [ ] Smoke test: launch portable, create an identity + memory, restart, verify they're still listed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #934, opened 2026-05-20
refactor(storage): flatten objects.db schema + retire the "forge" vocabulary

AgentA-asaf:
## Summary

Implements [`SPEC_SCHEMA_FLATTENING_2026_05_19.md`](docs/specs/SPEC_SCHEMA_FLATTENING_2026_05_19.md). Closes AUDIT_SQLITE_SYSTEMS §8.5; absorbs §8.1 / PR #933.

`objects.db` was built by an 11-step incremental migration chain (`run_forge_v1` … `run_forge_v11`). Per-version data dirs mean every new version is born with a fresh `objects.db` and ran the whole chain in one shot anyway — the intermediate states were never reachable. The chain was pure accretion, and "forge" is dead vocabulary (replaced by the Memory / Identity / agent-definition model).

- **Flatten** — `run_object_schema` defines the final schema directly in one idempotent `CREATE TABLE IF NOT EXISTS` batch. v1–v11, the v6→v7 shadow-migrate, and the v9→v10 workflow→drone copy are deleted (~1,400 lines incl. tests).
- **De-forge (Rust side only)** — `db_forge_*` tables → `db_agent_*`; structs `ForgeAgent`/… → `AgentDefinition`/…; `forge_*` store methods → `agent_def_*`/`agent_content_*`/`agent_skill_*`/`agent_history_*`; files `forge_handlers.rs` → `agent_handlers.rs`, `forge_seed.rs` → `agent_seed.rs`.
- **Dead tables dropped** — `db_workflow_*` + the `db_v10_migrated_legacy_*` sentinels (data already in `db_drone_*`). The dead `accounts` column is dropped from the schema.
- **`user_version` tripwire** — `stamp_and_check_version` stamps `PRAGMA user_version` on all four SQLite files and warns loudly on downgrade.

## Decision A1 — wire unchanged (read this before flagging naming)

The RPC command string constants (`COMMAND_*_FORGE_*` = `"listforgeagents"` etc.) and all serde field names are **deliberately unchanged** — they are the frontend wire contract. The const names mirror their string values on purpose. **The frontend is not modified.** Renaming the wire is a separate coordinated A2 follow-up. So a half-`forge` look at the `rpc_types.rs` command layer is intentional, not an oversight.

## Reviewer notes (pre-empting scrutiny)

- **Adopt step / downgrade-roundtrip.** `adopt_legacy_table_names` renames any pre-flatten tables found (carries a dev's pre-flatten `objects.db` forward; SQLite ≥ 3.25 auto-updates FK refs on parent rename). If a legacy table **and** its replacement both exist, it does **not** drop the legacy table — it warns loudly and leaves it for manual recovery. This deliberately avoids the silent-data-loss class behind PR #933's Codex P1 (covered by `test_adopt_legacy_both_tables_present_is_non_destructive`).
- **Cross-version registry migration is safe** — `registry/migrate.rs` only reads `db_agent_instances` (name unchanged) and already tolerates missing tables/columns.
- **`accounts` column** — verified dead (always `""`, never read post-insert). Dropped from the schema; the struct field is kept as an always-empty wire-compat vestige so the serialized agent shape is unchanged.
- **The unfiltered `cargo test -p agentmux-srv` hangs** on a pre-existing unrelated slow test (same before this branch). Targeted modules all pass — see test plan.

## Test plan

- [x] `cargo build` — `agentmux-srv` + `agentmux-launcher` clean
- [x] storage 103 · launcher 183 · identity 39 · registry 39 · drone 47 · agents 29 · reducer 93 · server 27 — all green
- [x] New: flat-schema creation/idempotency/dead-table omission, adopt rename + fresh no-op + both-present non-destructive + FK cascade, `user_version` stamp
- [ ] Reagent + Codex review
- [ ] Smoke: portable build — create an agent definition + identity + memory, restart, confirm persistence

### DIFF
--- a/.changesets/1779244019-refactor-storage-flatten-objects-db-schema-retire-the-forge--9eti.md
+++ b/.changesets/1779244019-refactor-storage-flatten-objects-db-schema-retire-the-forge--9eti.md
@@ -27,9 +27,8 @@ frontend is untouched and the wire contract is stable.
 
 **Dead tables dropped.** `db_workflow_definitions` / `db_workflow_runs` and the
 `db_v10_migrated_legacy_*` sentinels are no longer created; `adopt_legacy_table_names`
-drops them from any pre-flatten dev DB. The dead `accounts` column on agent
-definitions is dropped from the schema (kept as an always-empty struct field
-for wire compatibility).
+drops them from any pre-flatten dev DB — their data had already been copied
+into `db_drone_*`.
 
 **Safety net.** `adopt_legacy_table_names` runs once per startup: it renames
 any pre-flatten table names found (the single surviving fragment of the old
--- a/.changesets/1779244019-refactor-storage-flatten-objects-db-schema-retire-the-forge--9eti.md
+++ b/.changesets/1779244019-refactor-storage-flatten-objects-db-schema-retire-the-forge--9eti.md
@@ -1,0 +1,44 @@
+---
+type: patch
+---
+
+refactor(storage): flatten objects.db schema, retire the "forge" vocabulary, add user_version tripwire
+
+Implements `docs/specs/SPEC_SCHEMA_FLATTENING_2026_05_19.md`. Closes
+AUDIT_SQLITE_SYSTEMS §8.5 (and absorbs the §8.1 / PR #933 v11 rename).
+
+**Flatten.** The 11-step incremental migration chain (`run_forge_v1` …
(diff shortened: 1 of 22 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #937, opened 2026-05-20
fix(agent): gear/cog opens the settings panel reliably

AgentA-asaf:
## Summary

Clicking the ⚙ gear in an agent pane header did nothing.

The gear's click flips an overlay-tab signal (`_setOverlayTab`), but the panel was rendered behind:

```tsx
<Show when={showOverlayTab() != null && currentAgent() != null}>
```

`currentAgent()` resolves the pane's `agentId` block-meta against the `db_agent_definitions` list. That only matches for panes launched via `launchAgentDefinition` (the AgentPicker → launch-modal path). It is `null` for:
- **provider quick-launch panes** — `launchAgent` writes a *provider* id (`claude`/`codex`/…), not a definition id
- a pane whose definition was **deleted**
- the **async window** before / if `ListAgentDefinitionsCommand` resolves (its fetch error is silently swallowed)

In every one of those cases the gear silently no-op'd — click, nothing.

## Fix

Gate the overlay only on `showOverlayTab() != null`; pass `currentAgent()` (possibly `undefined`) straight through. The panel chain already handles a missing definition:
- `AgentCardSettingsPanel.agent` is typed `AgentDefinition | undefined` — undefined = create-mode
- the Identity tab has a "save the agent first" fallback

`AgentFocusedPanel`'s `agent` prop is widened from `AgentDefinition` to `AgentDefinition | undefined` to match what it forwards.

When `currentAgent()` *does* resolve (the common picker-launched case) behavior is unchanged — the panel opens to the definition detail view.

## Test plan

- [x] `tsc` clean for the touched files; `task build:frontend` clean
- [ ] Reagent + Codex review
- [ ] Smoke: launch an agent (both via the picker and via quick-launch if available), click the ⚙ gear → the settings overlay opens; Agent + Identity tabs render

### DIFF
--- a/.changesets/1779271911-fix-agent-gear-cog-opens-the-settings-panel-reliably-4o46.md
+++ b/.changesets/1779271911-fix-agent-gear-cog-opens-the-settings-panel-reliably-4o46.md
@@ -1,0 +1,24 @@
+---
+type: patch
+---
+
+fix(agent): gear/cog opens the settings panel reliably
+
+Clicking the ⚙ gear in an agent pane header did nothing. The gear flips an
+overlay-tab signal, but the panel was rendered behind
+`<Show when={showOverlayTab() != null && currentAgent() != null}>`.
+
+`currentAgent()` resolves the pane's `agentId` block-meta against the
+`db_agent_definitions` list. That only matches for panes launched via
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

PR #945, opened 2026-05-20
feat(stability): launcher auto-restarts the host on crash — supervision Phase 1

AgentA-asaf:
## Phase 1 — host auto-restart (first increment)

Second slice of the **service supervision & recovery framework** (Discussion
#943, issue #942). Builds on Phase 0 (#944, crash-modal suppression).

### Problem

The launcher's concurrent wait treated **any** host exit as shutdown — it
cleaned up and exited. A host crash (e.g. the 2026-05-20 GPU `LOG(FATAL)`) took
the whole app down; the user had to relaunch by hand.

### Change — `agentmux-launcher/src/main.rs`

The concurrent wait becomes a **supervised wait loop**:

- **Abnormal host exit** (non-zero code) → the launcher relaunches the host.
  The respawn re-attaches to the **still-running srv** (srv is a launcher
  *sibling*, untouched by a host crash — **Tier 1** recovery, spec §6) and the
  fresh host reloads the workspace from `objects.db` on its own startup.
- **Clean host exit** (code 0 — user quit) still ends the launcher.
- **Crash budget** (spec §10-A): at most `HOST_RESTART_BUDGET` (3) restarts
  within `HOST_RESTART_WINDOW` (60 s). Past that the launcher gives up and
  exits — **no crashloop**.
- The spawn → assign-to-J0 → resume sequence is extracted into
  `spawn_host_supervised()` so the first launch and every restart share one
  path. First launch passes the splash event; restarts pass `None`.

### Scope

This is the **first increment** of Phase 1. Deliberately *not* here, by design
(follow-on increments):

- Retry ladder (full → `--disable-gpu` → safe mode) — §7. This increment always
  retries full-config; the crash budget bounds it.
- Transient/deterministic crash *classification* by signature — §7.
- Recovery-rate metric / diagnostics surfacing — §12.
- Fault-injection test harness — §14. The crash-budget logic is simple and
  inline; extracting it behind a testable seam comes with the harness.

srv supervision is **Phase 2** — an srv exit still terminates the launcher here.

### Verification

`cargo check -p agentmux-launcher` — clean (pre-existing dead-code warnings
only). Runtime verification (a real host crash → observed relaunch) is manual;
the fault-injection harness is the next increment.

Discussion #943 · Issue #942

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #942, opened 2026-05-20
Service Supervision & Recovery — app-wide auto-recovery framework (no visible crashes)


AgentA-asaf:
## Summary

Rock-solid stability is core to AgentMux's value proposition. This issue tracks
an **app-wide service supervision & recovery framework** so that **the user
never *sees* a crash** — every process fault becomes an invisible, sub-second
auto-recovery (no OS modal, no lost work, at most a flicker).

Zero *process faults* is not achievable (GPU drivers segfault, the OS runs out
of commit, hardware misbehaves — none of it is our code). Zero *visible*
crashes **is** achievable, and is the design target.

**Case study #1:** the 2026-05-20 host crash — the Chromium GPU process failed
(flaky driver + exhausted page file), Chromium `LOG(FATAL)`-ed the host, and the
user got a raw Windows `0x80000003` "breakpoint" modal. A GPU subsystem failure,
not even AgentMux's own code, took the whole app down visibly.

Full spec drafted in-repo: `docs/specs/SPEC_SERVICE_SUPERVISION_AND_RECOVERY_2026_05_20.md`
(lands with the Phase 0 PR). Crash analysis: `docs/analysis/CRASH_GPU_PROCESS_FATAL_2026_05_20.md`.

## The approach

Recovery is the **dual** of the reducer/persistence layer — not part of it:

- **Reducer stack** — *what* the correct state is.
- **Persistence** — *where* it's durable (`objects.db`, `sagas.db`, …).
- **Supervision (new)** — *keeps every service alive and reconciled to it.*

Recovery = **restart** (a supervisor gets a fresh process running) + **rehydrate**
(the reducer/persistence layer replays authoritative state into it). A restarted
service is just a new subscriber that needs the current snapshot — no special
recovery path, no new state ownership.

## Prime directive

Two concerns — performance overhead, and recovery systems that "shoot themselves
in the foot" — converge on one rule. A design that violates it is rejected at
review:

> **The supervisor is passive, bounded, and simpler than everything it
> supervises.** It observes signals that already exist, free-rides on
> persistence that already happens, acts only on real crashes, is bounded by a
> crash budget, and is loud about every action it takes.

## Performance — target ≈ 0 steady-state

Every mechanism is passive: crash detection is an OS process-exit event (no
polling); liveness is the same event; the responsiveness ping piggybacks the
existing `mem_heartbeat`; rehydration **reads** `objects.db`, which the reducer
[...]

PR #946, opened 2026-05-20
feat(stability): host retry ladder + crash classification — supervision Phase 1

AgentA-asaf:
## Phase 1 increment 2 — retry ladder + crash classification

Follow-on to #945 (bounded host auto-restart). Discussion #943.

### Problem

#945 always relaunched the host with the **same** full config. A
deterministic crash (one that reproduces every launch) would just be retried
identically until the crash budget ran out.

### Change — `agentmux-launcher/src/main.rs`

- **Crash classification (spec §7):** the supervised wait loop tracks the
  previous abnormal exit code. A crash that *reproduces* it is deterministic —
  retrying full-config is "the same thing again".
- **Retry ladder (spec §7):** rung 1 is a full-config relaunch; on a
  reproduced crash the launcher steps down to **rung 2 — degraded**
  (`--disable-gpu`, software rendering, no GPU process to crash). Degraded is
  **sticky** — the ladder only steps down. `spawn_host_supervised()` gains a
  `disable_gpu` flag that appends the Chromium switch.
- The crash budget (3 / 60 s) from #945 still bounds the whole ladder.

### Scope

Rungs 3–4 (last-known-good state, safe mode) are later increments — they need
state-snapshot machinery the launcher doesn't have yet. Recovery-rate metric
surfacing (§12) and the fault-injection harness (§14) also remain.

### Verification

`cargo check -p agentmux-launcher` — clean (pre-existing warnings only).

Discussion #943 · Issue #942

### DIFF
--- a/.changesets/1779307832-feat-stability-host-retry-ladder-crash-classification-superv-4mnh.md
+++ b/.changesets/1779307832-feat-stability-host-retry-ladder-crash-classification-superv-4mnh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(stability): host retry ladder + crash classification (supervision Phase 1)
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #947, opened 2026-05-21
fix(statusbar): opacity slider updates live while dragging

AgentA-asaf:
## Problem

In the status-bar Instance panel, the per-window **Opacity** slider's thumb
and `%` label only updated when the user *released* the slider — not while
dragging.

Root cause: both were bound to `currentOpacity()`, which reads the persisted
`window:opacity` window meta. That meta is written **only** by the `onChange`
(release) handler. The `onInput` (drag) handler dispatches to the live
window-opacity store — so the *window* dimmed live, but the slider value
didn't.

## Fix — `frontend/app/statusbar/InstancePanel.tsx`

A per-row `dragOpacity` signal:
- `onInput` (drag) → `setDragOpacity(val)`
- `onChange` (release) → `setDragOpacity(null)`
- `displayOpacity() = dragOpacity() ?? currentOpacity()` drives the slider
  `value` and the `%` label.

So during a drag the thumb + label track continuously; on release it falls
back to the persisted value (canonical once it settles, ~one frame later).

## Verification

`npx tsc --noEmit` — clean for `InstancePanel.tsx`.

### DIFF
--- a/.changesets/1779334638-fix-statusbar-opacity-slider-value-updates-live-while-draggi-pb26.md
+++ b/.changesets/1779334638-fix-statusbar-opacity-slider-value-updates-live-while-draggi-pb26.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(statusbar): opacity slider value updates live while dragging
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #952, opened 2026-05-21
revert: opacity slider live-drag (#947) — it broke window transparency

AgentA-asaf:
Reverts #947.

## Why

#947 added live value-tracking to the opacity slider (the `%` updating during
drag instead of only on release). It was verified by `tsc` and approved by the
review bots — but **never smoke-tested against the actual feature**, and it
**broke window transparency** entirely.

Confirmed empirically: **v0.37.1 (pre-#947) has working transparency**; builds
carrying #947 do not. The bots can't catch a broken runtime feature — only
dragging the slider and watching the window can.

The live-value display was a cosmetic optimization on top of a working
feature. It is not worth losing transparency, so #947 is reverted to restore
the feature. A future attempt at the value-display tweak must be smoke-tested
end-to-end against transparency before merge.

### DIFF
--- a/.changesets/1779334638-fix-statusbar-opacity-slider-value-updates-live-while-draggi-pb26.md
+++ b/.changesets/1779334638-fix-statusbar-opacity-slider-value-updates-live-while-draggi-pb26.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #949, opened 2026-05-21
build: task package auto-bumps the patch version every build

AgentA-asaf:
## Problem

Agents kept producing portables at a **stale or duplicate version**, making builds impossible to tell apart.

- `task package` built at the *committed* version and never bumped it.
- `task package:local` — the intended "fix" — did an **ephemeral** bump computed as `committed-version + 1`. With no memory of the last build number, every run recomputed the **same** number (`0.37.0` → always `0.37.1`).
- Same version ⇒ same `~/.agentmux/versions/<v>/` data dir ⇒ two builds collide, share state and logs, and become genuinely indistinguishable.

## Fix

**`task package` now runs a committed `bump patch` as its mandatory first step.** The committed version *is* the durable monotonic counter — every portable advances it for real, so no two builds ever share a version or data dir. The build task enforces it, so an agent cannot produce a stale build by forgetting to bump.

- **`Taskfile.yml` `package`** — prepend `bash scripts/bump-wrapper.sh patch -m build --commit`.
- **`package:local`** — retired; now a thin alias of `package` (kept for muscle memory).
- **`scripts/package-local.sh`** — deleted (the no-memory ephemeral trap).
- **`CLAUDE.md`** — documents the new behavior. `task dev` stays unbumped on purpose: the dev data dir is keyed on the git branch (`~/.agentmux/dev/<branch>/`), not the version, so dev instances are already isolated.
- Version reseeded `0.37.0` → `0.37.1` so the first post-merge build is `0.37.2` and skips the already-used `0.37.1`.

CI and the release flow do **not** call `task package`, so the auto-bump never fights `task release` / the changeset workflow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #953, opened 2026-05-21
fix(build): stamp srv binary with the live post-bump version

AgentA-asaf:
## Regression from #949

#949 made `task package` auto-bump the version as its **first** command. But
Task resolves the `{{.VERSION}}` template variable **once at invocation start**
— before that bump runs.

`build:backend:rust:windows` named the srv binary with `{{.VERSION}}` (the
stale, pre-bump version), while `package-portable.sh` resolves the version
*live* (post-bump). The names mismatched and packaging failed:

```
ERROR: dist/bin/agentmux-srv-0.37.4-windows.x64.exe not found — build first
```

## Fix

`build:backend:rust:windows` now reads the version **live from `package.json`**
in the copy step — the same source `package-portable.sh` uses — so both agree
on the post-bump version.

The `dist/bin` path has no spaces, so it needs no quoting. (An earlier in-flight
attempt used `\"` to quote it — invalid PowerShell, which escapes with a
backtick, not a backslash; that broke the `Copy-Item` outright.)

`task dev` is unaffected — it doesn't bump, so live-read and `{{.VERSION}}`
resolve to the same value there.

### DIFF
--- a/.changesets/1779360895-fix-build-stamp-srv-binary-with-the-live-post-bump-version-30eo.md
+++ b/.changesets/1779360895-fix-build-stamp-srv-binary-with-the-live-post-bump-version-30eo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): stamp srv binary with the live post-bump version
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

a5af:
@codex review

PR #954, opened 2026-05-21
feat(statusbar): opacity slider % tracks the drag

AgentA-asaf:
## What

The Instance-panel opacity slider's `%` label updated only on *release*. It now
tracks the drag tick-by-tick.

## How — store-reactive, slider input untouched

The `%` label read `currentOpacity()` — the persisted `window:opacity` meta,
written only by `onChange`. So it lagged the drag.

- **`window-opacity-store.ts`** — the store's state is now a Solid signal (was
  a plain `let`); new reactive `liveWindowOpacity()` accessor.
- **`InstancePanel.tsx`** — the `%` label reads
  `liveWindowOpacity() ?? currentOpacity()`. The store is updated by
  `dispatchWindowOpacity` on every `onInput`, so the label tracks the drag;
  it falls back to the persisted value before any drag.

The slider `<input>`, `onInput`, and `onChange` are **byte-for-byte
unchanged** — the transparency IPC path is not touched.

## Background

This replaces the reverted #947, which implemented the same feature by routing
the slider's `value` through a drag-updating signal. This version changes
**only the label**, never the controlled `<input>`.

## Verification

`tsc` clean. Confirmed in a `task dev` build — transparency and the live `%`
both work.

> Before merge: a portable will be built and launched several times to confirm
> transparency holds across launches (it had an intermittent failure earlier in
> testing — being thorough per the smoke-test-before-merge rule).

### DIFF
--- a/.changesets/1779379302-feat-statusbar-opacity-slider-tracks-the-drag-reactive-store-vqdd.md
+++ b/.changesets/1779379302-feat-statusbar-opacity-slider-tracks-the-drag-reactive-store-vqdd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(statusbar): opacity slider % tracks the drag (reactive store)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #955, opened 2026-05-21
feat(statusbar): DEV badge on the version chip

AgentA-asaf:
## What

The status-bar version chip now shows a **`DEV`** badge to the right of the
version when the build is a dev build:

```
 v0.37.6 DEV      ← task dev
 v0.37.6          ← packaged portable (unchanged)
```

## Why

A `task dev` build and a packaged portable showed an identical chip
(`vX.Y.Z`). With several builds running side by side — routine during
testing — there was no way to tell them apart. This caused real confusion in
the recent opacity investigation (a dev build and a portable were both
`v0.37.6`).

## How

Frontend-only:
- `StatusBar.tsx` — `<Show when={isDev()}>` rendering a `DEV` span inside the
  version button and the offline-fallback span, after `v{version}`. `isDev()`
  is the existing dev-mode flag (`@/store/global`).
- `StatusBar.scss` — `.status-version-dev`, a small outlined accent tag.

Portables are unaffected (`isDev()` is false).

## Spec

Adds `docs/specs/SPEC_DEV_VERSION_BADGE_2026_05_21.md` — this PR is Feature A;
the spec also defines Feature B (Build row → commit hash) and Feature C (a
"Time" row with the real build timestamp), tracked separately.

### DIFF
--- a/.changesets/1779380831-feat-statusbar-dev-badge-on-the-version-chip-k4sm.md
+++ b/.changesets/1779380831-feat-statusbar-dev-badge-on-the-version-chip-k4sm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(statusbar): DEV badge on the version chip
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #951, opened 2026-05-21
fix(shell): seq-based input reorder buffer + threshold session reset

AgentY-asaf:
## Problem

Terminal panes freeze after out-of-order or concurrent input RPCs. Two distinct failure modes discovered during echo-latency benchmarking:

**Failure 1 — Concurrent fire-and-forget overflow**
The stream benchmark sent 220 `controllerinput` RPCs simultaneously. Multiple Tokio tasks processed them in parallel → packets arrived out of order → no reorder buffer existed → seqs were silently discarded → `input_seq_next` never advanced → terminal permanently frozen.

Evidence from sidecar log:
```
WARN send_input: input reorder buffer full, dropping  block_id="07dfd2db" seq=967
WARN send_input: input reorder buffer full, dropping  block_id="07dfd2db" seq=969
WARN send_input: input reorder buffer full, dropping  block_id="07dfd2db" seq=983
```

**Failure 2 — Session desync on frontend reconnect**
After a benchmark advanced `input_seq_next` to ~200, a new TermViewModel started at seq=0. Without a reset heuristic, seq=0 was seen as a far-past duplicate and discarded — new terminal silently accepted no input.

### `agentmux-srv/src/backend/blockcontroller/shell.rs`

- Added `input_seq_next: u64` and `input_seq_buf: BTreeMap<u64, BlockInputUnion>` to `ShellControllerInner`
- `start()` resets both to 0/empty when a new PTY channel is created — guarantees fresh state on controller restart
- `send_input()` now takes `seq: Option<u64>`:
  - `None` → direct `try_send`, bypasses all seq machinery (used by `blockinput`, `setblocktermsize`, agent API)
  - `Some(s)` → full ordering logic:
    - **In-order** (`s == input_seq_next`): deliver immediately, then drain buffered in-order followers
    - **Out-of-order future** (`s > input_seq_next`): buffer in `BTreeMap` up to capacity; warn and drop if full
    - **Duplicate/past** (`s < input_seq_next`): discard with warning
    - **Channel full**: drop + warn, never block the mutex holder (no deadlock)
    - **Session reset**: if `s == 0 && input_seq_next > 0` OR gap > `SESSION_RESET_THRESHOLD (64)`, reset counter to arriving seq

### `agentmux-srv/src/backend/rpc_types.rs`

- Added `seq: Option<u64>` to `CommandBlockInputData` (serde default = None, skipped if absent — wire-compatible)

### `agentmux-srv/src/server/websocket.rs`

[...]

PR #956, opened 2026-05-21
fix(term): route keyboard input through the blockinput wscommand

AgentY-asaf:
## Problem

Fast typing into a terminal pane produced **out-of-order characters** — keystrokes appeared with variable latency, so the user was already on the next key before the prior one rendered. Slow typing was unaffected. Reproduced across multiple sessions and versions, including v0.37.6 after PR #951.

## Root cause

Keystrokes were sent via the `controllerinput` **RPC**. RPC messages are dispatched through the RPC engine, which runs handlers **concurrently** — two adjacent keystroke RPCs race and can call `send_input` in reversed order.

PR #951 added a seq reorder buffer to compensate, but **the frontend never sent a `seq`** (Codex P1), so the buffer never activated for real typing. It was dead code on the hot path.

## Fix

Switch all keyboard/text input to the `blockinput` **wscommand**:

- `blockinput` is handled **inline and synchronously** in the WS receive loop — `handle_incoming_text` is awaited, so one message is fully processed before the next is read. TCP order is preserved end-to-end. No race, no buffer, no seq.
- Switched: `termViewModel.sendDataToController`, the voice input handle, and term-sticker click-commands.
- `controllerinput` RPC is kept for the SIGINT signal path (`useAgentCommands.ts`) and a future `term.type` agent API.

Also **drops the `gap > 64` session-reset heuristic** in `shell.rs` (Codex P2 on #951): a stale/duplicate packet far behind the expected seq could falsely trigger a reset and replay old input. The explicit `seq == 0` reset is kept.

The reorder buffer itself is left intact for the future `term.type` agent API, which will send many chars at once and genuinely needs ordering.

## Builds on

PR #951 — addresses its Codex P1 (frontend never sends seq) and P2 (gap heuristic misfire) findings.

## Test plan

- [ ] Slow typing: characters in order, no latency
- [ ] Fast typing (hold a key / type a sentence quickly): correct order, no latency spikes
- [ ] Open a new terminal while another is active: both accept input
- [ ] Close and reopen a terminal pane: input still works
- [ ] Voice input into a terminal still reaches the PTY
- [ ] Sticker click-commands still send input

### DIFF
--- a/.changesets/1779383684-fix-term-route-keyboard-input-through-the-blockinput-wscomma-stgh.md
+++ b/.changesets/1779383684-fix-term-route-keyboard-input-through-the-blockinput-wscomma-stgh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): route keyboard input through the blockinput wscommand so fast typing stays in order
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.37.1"
+version = "0.37.2"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

AgentY-asaf:
@codex review

PR #958, opened 2026-05-22
fix(magnify): single-instance pane render + nothing visible behind a magnified pane

AgentX-asaf:
## Summary

Fixes magnified ("maximized") panes: a magnified-then-restored pane was left broken, and a magnified pane let other AgentMux content bleed through. Four related fixes.

### 1. Single-instance render — fixes the magnify→restore regression

The magnified pane was rendered **twice** (a hidden tile copy + an overlay copy) sharing one `blockId`-keyed registry entry and, for browser panes, one native `CefBrowserView`. On restore the overlay copy's unmount cleared the shared registry / disposed the shared `ViewModel` / fired `browser_pane_close` — leaving the surviving copy with:
- **terminal/agent panes** — dead zoom (`Ctrl +/-`, `Ctrl+scroll`)
- **browser panes** — a destroyed native window (pane goes **black and stuck**)

Fix: render the pane's `.tile-leaf` once and **reparent** that single DOM node into the magnify overlay while magnified. Moving a DOM subtree doesn't dispose its SolidJS component, so the `Block`, `ViewModel`, registry entry, and browser-pane HWND all survive. `MagnifiedPaneOverlay` exposes a mount slot via the new `layoutModel.magnifyMount` signal.

### 2. Nothing inside AgentMux shows behind a magnified pane

While any pane is magnified, every tile node is hidden with `display:none` (was `visibility:hidden` on just the magnified node). `display:none` collapses each node to 0×0, so a browser pane inside reports a 0×0 rect and the host's **existing** airspace logic hides its native window — native browser panes paint above DOM, so an opaque background alone could not occlude them. No new host code.

OS-level window transparency (`WS_EX_LAYERED`, e.g. 80%) is untouched — the global "desktop at N%" still applies; only *AgentMux* content is hidden behind the magnified pane.

### 3. Magnified pane is fully opaque

A magnified pane used the bare `--block-bg-color`, translucent in every theme (default 50%), so the blurred backdrop bled through. `.magnify-pane` and `.magnified .block-frame-default-inner` now use `rgb(from var(--block-bg-color) r g b)` — the theme colour with alpha omitted = fully opaque.

### 4. Magnified pane fills the window

`window:magnifiedblocksize` default `0.9` → `1.0` (was leaving a 5% margin gap). Still a user-overridable setting.

## Design docs

`docs/specs/SPEC_MAGNIFY_ZOOM_{REGRESSION_AND_DEFAULTS,IMPLEMENTATION}_2026-05-21.md` and `SPEC_MAXIMIZE_ZOOM_ARCHITECTURE_2026-05-21.md` — root-cause analysis, reducer/browser-pane considerations, phased plan.

## Test plan

- [x] `task build:frontend` — clean
- [x] `npx vitest run frontend/layout` — 51/51 pass
- [x] Verified live in `task dev` (Windows)
- [ ] Manual:
  - [ ] Terminal pane: magnify → restore → `Ctrl +/-` and `Ctrl+scroll` still zoom; repeat ×5
  - [ ] Browser pane: magnify → restore → stays live (not black), keeps URL/scroll; repeat ×5
  - [ ] Magnify a pane while another tile holds a browser pane — nothing bleeds through
  - [ ] Magnified pane fills the window, fully opaque
  - [ ] `Escape` / backdrop-click still unmagnify

### DIFF
--- a/.changesets/1779410797-fix-magnify-single-instance-pane-render-fixes-zoom-browser-p-4rem.md
+++ b/.changesets/1779410797-fix-magnify-single-instance-pane-render-fixes-zoom-browser-p-4rem.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(magnify): single-instance pane render — fixes zoom + browser-pane black on restore
--- a/docs/specs/SPEC_MAGNIFY_ZOOM_IMPLEMENTATION_2026-05-21.md
+++ b/docs/specs/SPEC_MAGNIFY_ZOOM_IMPLEMENTATION_2026-05-21.md
@@ -1,0 +1,363 @@
+# SPEC: Magnify & Zoom — Implementation Plan
+
+**Date:** 2026-05-21
+**Author:** AgentX
+**Status:** Draft — implementation plan, ready for review
+**Companions:**
+- `SPEC_MAGNIFY_ZOOM_REGRESSION_AND_DEFAULTS_2026-05-21.md` — the defects + default changes (the *what*).
+- `SPEC_MAXIMIZE_ZOOM_ARCHITECTURE_2026-05-21.md` — architecture analysis (the *why*).
+
+This document is the *how*: a phased, file-level implementation plan.
+
+---
+
+## 1. Overview
+
+Four phases, ordered so each lands independently and de-risks the next.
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #959, opened 2026-05-22
feat(modal): unified scope-based modal primitive (stage 1)

AgentA-asaf:
## Stage 1 of 5 — `SPEC_UNIFIED_MODAL_SYSTEM_2026_05_21`

Adds **`@/element/modal`** — the unified modal primitive with a first-class **`scope`** axis (`window` | `tab` | `pane`): *what the modal locks*. Mount point, backdrop extent, `inert` boundary, scroll lock and the modal stack all derive from the scope.

### What's here
- **Ported verbatim** from modal-v2 (the proven a11y core): sentinel focus trap, focus save/restore, ARIA labelling + title-id context, pane-overlay clip, reduced-motion.
- **Redesigned around `scope`:**
  - scope-resolved mount — `window` portals to the window body; `tab`/`pane` mount into a context-supplied node (`TabModalScope` / `PaneModalScope`), falling back to `window` with a warn if no provider.
  - scope-relative `inert` + scroll lock — per *lock-region* (ref-counted `WeakMap`), not document-wide. `window`'s region is `document.body`, so window-modal behaviour is byte-identical to modal-v2.
  - scope-aware stack — ESC/backdrop act on the *reachable topmost* (containment-based: a modal shadowed by a higher modal whose lock region contains it is skipped; disjoint regions coexist).
- **Dismissal (folds in `SPEC_AGENT_LAUNCH_AND_MODAL_DISMISSAL` B/C):** `closeOnBackdropClick={false}` no longer silently swallows a backdrop click — it nudges the panel's `[data-modal-dismiss]` control.

### Scope of this PR
**`modal.tsx` is intentionally unused so far** — it co-exists with `modal-v2` (untouched). Per spec §11, stage 2 migrates the 8 window-scope callers, stage 3 the tab modals, stage 4 the v1 registry, stage 5 deletes `modal-v2`.

`tsc` clean.

### DIFF
--- a/.changesets/1779415195-feat-modal-unified-scope-based-modal-primitive-stage-1-k5kv.md
+++ b/.changesets/1779415195-feat-modal-unified-scope-based-modal-primitive-stage-1-k5kv.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(modal): unified scope-based modal primitive (stage 1)
--- a/frontend/app/element/modal.scss
+++ b/frontend/app/element/modal.scss
@@ -1,0 +1,275 @@
+// Copyright 2024-2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+// Modal — the unified, scope-based modal primitive (Stage 1 of
+// SPEC_UNIFIED_MODAL_SYSTEM_2026_05_21). Ports modal-v2.scss's rules and
+// adds: scope-specific positioning (tab/pane backdrops are absolute
+// within their mount node, not fixed full-window) and the cancel-nudge
+// keyframes (spec §9).
+//
+// Consumes design tokens from `theme.scss` (SPEC_DESIGN_SYSTEM_2026_04_23).
+
+@use "../mixins.scss";
+
+.modal-root {
+    // Window scope: a fixed full-window overlay (identical to modal-v2).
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Re Codex's P1 at `modal.tsx:217` ("Skip inert on nested modal scope hosts") — this is the **intended** cross-scope behavior, not a bug.

Per `SPEC_UNIFIED_MODAL_SYSTEM` §6 and the §12.4 decision: a lower-scope modal opened under a higher-scope modal — a `pane` modal under a `tab` modal — is **correctly inert-but-mounted**. A `tab` modal locks the *entire tab*; a pane lives inside that tab, so a pane modal there is behind the tab modal and must not be interactive. The scope hierarchy is window > tab > pane — a higher scope blocking a lower one *is* the model. `isReachable()` already (correctly) marks the pane modal as shadowed.

It is not stranded: when the tab modal closes, `releaseRegionLock()` un-inerts exactly the children it recorded (the pane root), and the pane modal becomes the active modal in its pane. Making the pane modal interactive *through* an open tab modal would break the lock-scope guarantee.

No change — flagging as a Codex misfire.

PR #963, opened 2026-05-22
refactor(modal): retire the v1 modal registry (stage 4)

AgentA-asaf:
## Stage 4 of 5 — `SPEC_UNIFIED_MODAL_SYSTEM_2026_05_21`

Retires the legacy `displayName`-keyed modal registry. Window modals now open with a **direct component reference**.

### New imperative API (`store/modalmodel.ts`)
- `openModal(Component, props?)` → `ModalHandle` — opens `Component` (a reference, no string key); returns `{ close }`.
- `closeModal(id)` / `closeTopModal()` (the latter for the global Escape key).
- `ModalCloseProps { close }` — every opened component receives a **per-entry** `close` handle, so a modal dismisses *itself* regardless of stack position. (The old `popModal()` always popped the top — wrong once two modals stack.)

### Migrated — all 7 call-sites
`keymodel.ts` (Escape → `closeTopModal`, Ctrl+P), `tabbar.tsx`, `ConfigStatus.tsx`, `LanStatus.tsx`, `global.ts` (×2: About menu, user-input event). v1 components (`MessageModal`/`UserInputModal`/`CommandPaletteModal`/`AboutModal`) keep their exact behaviour — they now take a `close` prop instead of reaching into the store.

### Deleted
- `modals/modalregistry.tsx` (the `getModalComponent` / `displayName` map). `modalsrenderer.tsx` slimmed to a `<For>` over the entry list — still mounted in `workspace.tsx`.

### Note for reviewers
Direct component references mean `keymodel`/`tabbar`/`global` now import modal components, and those components import services back (`keymodel↔command-palette`, `global↔about/userinput`). **All such references are lazy** — used only inside callbacks / render, never at module-eval time — so ESM resolves the cycles cleanly. `tsc` clean.


a5af:
@codex review

PR #965, opened 2026-05-22
fix(statusbar): show DEV badge in dev mode + add to instance panel

AgentA-asaf:
## DEV badge never showed in `task dev`

The DEV badge on the status-bar version chip never appeared in a `task dev` session. Two layered causes:

### 1. `get_is_dev()` keyed off the build profile
The host command returned `cfg!(debug_assertions)`. `task dev` builds and runs a **release** binary, so this was always `false` — even though the runtime mode is `dev:<branch>`. Now resolves the real `RuntimeMode` (launcher-injected env, or host exe path), matching the pattern in `app.rs`.

### 2. `about.tsx` used a dead `isDev`
`about.tsx` imported `isDev` from `frontend/util/isdev.ts`, whose `getEnv`-based check is dead under CEF — `api.getEnv` is a hard-coded `""` stub (and the var name was still the pre-rebrand `WAVETERM_DEV`). Switched it to the working `isDev` in `@/store/global` (backed by the `get_is_dev` IPC) and **deleted the now-unused `isdev.ts`** — it was a footgun duplicate of the real `isDev`.

### Also
- DEV badge added to the **InstancePanel** version row.
- `.status-version-dev` de-nested so the status bar and the panel share one badge style.

Verified live in a `task dev` instance — the DEV badge shows in both the status bar and the expanded instance panel.


a5af:
@codex review

a5af:
@codex review

PR #966, opened 2026-05-22
feat(agent): launch modal defaults to Continue mode for known definitions

AgentA-asaf:
## Feature A — Launch modal defaults to "Continue existing agent"

Implements **Feature A** of `SPEC_AGENT_LAUNCH_AND_MODAL_DISMISSAL_2026_05_21`.

When the user launches an agent **definition** that already has past **instances**, the modal previously opened in *New agent* mode — continuing was buried in a dropdown. The common case (re-open the agent you already configured) is now the default.

### Behavior
- Explicit **Continue / New** view mode (`viewMode` signal — local, derived from `namedAgents()` at open time; no new RPC or reducer command).
- **Definition has past instances** → opens in **Continue** mode, most-recent instance preselected. The instance dropdown + a *"+ Start a new agent instead"* toggle.
- **New** mode (when past instances exist) → a *"↩ Continue an existing agent"* toggle back.
- **No past instances** → opens in New mode, no toggle.
- A `+New`-bundle round-trip (`initialFormState` present) stays in New mode — those round-trips only originate there (the `+New` buttons are disabled while continuing).

`tsc` clean. The integration test mocks `ListNamedAgentsCommand → []` so it stays in New mode — unaffected. Verified live in `task dev`.

### DIFF
--- a/.changesets/1779440490-feat-agent-launch-modal-defaults-to-continue-mode-for-known--kyhz.md
+++ b/.changesets/1779440490-feat-agent-launch-modal-defaults-to-continue-mode-for-known--kyhz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): launch modal defaults to Continue mode for known definitions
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #971, opened 2026-05-22
feat(modal): app-wide singleton modal coordination layer

AgentA-asaf:
## Bundle PR 3 — singleton coordination layer (`SPEC_BUNDLE_MANAGEMENT_2026_05_22` §3)

Infrastructure for an **app-wide singleton modal** — exactly one instance across every window of the AgentMux process. PR 4 composes the real bundle manager on top.

### Mechanism — WPS-backed, zero Rust change
A claim is a `singleton:claim` WPS event published with `persist:1`, so the broker keeps the latest claim and replays it to windows that subscribe later — that *is* the "current holder" registry. `frontend/app/store/singleton-modal.ts`:
- `acquireSingleton(kind)` → bool, `releaseSingleton(kind)`, `singletonHolder(kind)` (reactive accessor), `holdsSingleton(kind)`.
- **Crash release** — `startSingletonCrashRelease()` (wired once in `app-init.ts`) watches launcher window-exit events; if a holder window dies, a live window publishes the release on its behalf (idempotent).
- Race on a genuinely-free claim is documented + acceptable for the deliberate-click use case.
- **17 unit tests**.

### Placeholder demo — temporary scaffolding
PR 3 ships a placeholder modal + a "Singleton demo" hamburger item so the layer is independently exercisable. **PR 4 removes the placeholder** when it wires the real `BundleManagerModal` onto this layer.

`tsc` clean.

### DIFF
--- a/.changesets/1779446379-feat-modal-app-wide-singleton-modal-coordination-layer-3nho.md
+++ b/.changesets/1779446379-feat-modal-app-wide-singleton-modal-coordination-layer-3nho.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(modal): app-wide singleton modal coordination layer
--- a/frontend/app-init.ts
+++ b/frontend/app-init.ts
@@ -55,6 +55,7 @@ import {
     seedKnownEntriesFromSnapshot,
     startLauncherEventReducer,
 } from "@/app/store/launcher-event-reducer";
+import { startSingletonCrashRelease } from "@/app/store/singleton-modal";
 
 // Deferred — assigned inside initApp() after window.api is ready.
 // Do NOT call getApi() at module level: this file is statically imported by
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
codex's P1 here re-fires the previous round's finding against code that commit `745aa586` **removed**. The comment quotes `if (now != null && now !== label) return` — that deferred-claim block no longer exists.

`acquireSingleton` now returns `false` immediately when the window label is unresolved (`if (me == null) return false`), so the `return true` it flags is reached only *after* the label is confirmed and the claim is published — `true` strictly means "holds it now." This is covered by the new `returns false when this window's label is unresolved` test (20 tests pass). reagent APPROVED `745aa586`.

Treating this as a stale re-fire — merging.

PR #972, opened 2026-05-22
feat(bundles): Identity & Memory manager modal + hamburger entry

AgentA-asaf:
## Bundle PR 4 — Feature 2 (`SPEC_BUNDLE_MANAGEMENT_2026_05_22` §3)

The canonical app-wide home for managing Identity & Memory bundles.

### What landed
- **`BundleManagerModal`** — `<Modal scope="window">`, **app-wide singleton** via PR 3's layer. Left-rail toggle **Identities ⇄ Memories**; each section renders PR 2's context-free `<IdentityManager/>` / `<MemoryManager/>` (kept mounted so toggling is instant).
- **Hamburger "Identity & Memory"** entry → `openBundleManager()`: `acquireSingleton("bundle-manager")` → opens here, or focuses the holding window if it's open elsewhere.
- **`BundleManagerElsewhereBanner`** — persistent "open in <Window N> — click to focus" banner in non-holding windows (§5 decision 6), reactive to `singletonHolder`; clears automatically on release/crash.
- **Placeholder demo removed** — `singleton-placeholder-demo.{tsx,scss}` deleted, `app.tsx` + `tabbar.tsx` swapped to the real manager, dead `SINGLETON_KIND_PLACEHOLDER` dropped. The PR 3 coordination layer (`singleton-modal.ts`, `startSingletonCrashRelease`) is untouched.

§5 decisions applied: label "Identity & Memory", left-rail layout, persistent banner. The agent-settings Identity/Memory tab demotion (§5.3) is a tracked follow-up.

`tsc` clean; 20/20 singleton tests pass. Worth a smoke (open from two windows → banner + focus) before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #973, opened 2026-05-22
refactor(bundles): demote agent-settings Identity/Memory tabs to read-only

AgentA-asaf:
## Bundle PR 5 — `SPEC_BUNDLE_MANAGEMENT_2026_05_22` §5 decision 3

Now that the hamburger **"Identity & Memory"** manager (#972) is the one CRUD home, the per-agent cog → Identity/Memory settings tabs should not duplicate full bundle CRUD.

### What changed
- New `BundleSummaryPanel` (`frontend/app/view/bundle-summary.tsx`) — a read-only panel: explains bundles are app-wide data managed in one place, plus a **"Manage in Identity & Memory"** button → `openBundleManager()`.
- `identity-pane-view.tsx` / `memory-view.tsx` now render `<BundleSummaryPanel/>` instead of the full-CRUD `IdentityManagerBody` / `MemoryManagerBody`.
- The context-free `IdentityManager` / `MemoryManager` and `BundleManagerModal` are untouched.

### Data gap (documented)
The spec's ideal — "this agent uses Identity: *X*" — needs the launched bundle ids, which live on the `AgentInstance` row reachable only from the *agent pane's* block (`agentInstanceId`). The `view: "identity"` / `view: "memory"` panes are **separate blocks** with no link to the launching agent, so the per-agent bundle isn't resolvable here. Per the spec's fallback, this PR demotes to the **pointer-only** form; surfacing "uses *X*" would need new plumbing (propagating `agentInstanceId` into the settings-tab block) — a follow-up.

`tsc` clean.

### DIFF
--- a/.changesets/1779460091-refactor-bundles-demote-agent-settings-identity-memory-tabs--lhry.md
+++ b/.changesets/1779460091-refactor-bundles-demote-agent-settings-identity-memory-tabs--lhry.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(bundles): demote agent-settings Identity/Memory tabs to read-only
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #975, opened 2026-05-22
feat(widgets): responsive widget labels + all widgets pinned by default

AgentY-asaf:
## Summary

The action widget bar's text labels now **show or hide automatically based on the title bar width** — they collapse to icon-only when the title bar is too narrow and expand again when there is room.

- **Responsive labels.** `effectiveIconOnly = widget:icononly || tooNarrow`. `tooNarrow` is derived from a hidden, always-labeled **measuring mirror**, so the decision is computed from an invariant width — there is no measure→collapse→re-expand oscillation and no hard-coded window-width breakpoint.
- **`widget:icononly` kept, default flipped `true` → `false`.** The manual setting still lets a user force icon-only at any width; labels now show by default when they fit.
- **All widgets pinned by default.** `editor`, `drone`, and `help` gain `display:pinned: true` (joining `agent`, `browser`, `terminal`, `sysinfo`). The More dropdown starts empty.

## Notes / decisions

- `MIN_TAB_WIDTH` (120px) is the only constant — it floors how much header space is reserved for the tab bar before labels are sacrificed. The collapse point still auto-tracks the widget count via the mirror's measured width, so it stays correct if widgets are pinned/unpinned.
- **`swarm` is now also pinned + surfaced** — per follow-up discussion, the `display:hidden` flag was removed so the bar shows all 8 widgets by default. swarm was hidden as a relic from when top-level swarm widgets duplicated agent-pane functionality; the modern multi-agent orchestration view is fine to surface like any other widget.
- First use of `ResizeObserver` in the frontend.
- Design rationale is captured in `docs/investigations/INVESTIGATION_WIDGET_LABEL_RESPONSIVE_2026_05_22.md`.

## Test plan

- [ ] `task dev` — widget labels visible on a wide window
- [ ] Narrow the window — labels collapse to icons before the bar overflows; widen — labels return; no flicker at the boundary
- [ ] Fresh config shows all 8 widgets pinned in the bar; More dropdown empty
- [ ] Right-click bar → "Icon Only" still forces icon-only on a wide window
- [ ] Pin/unpin a widget — collapse threshold adjusts accordingly

### DIFF
--- a/.changesets/1779463287-feat-widgets-responsive-labels-collapse-to-icons-on-narrow-t-y8qk.md
+++ b/.changesets/1779463287-feat-widgets-responsive-labels-collapse-to-icons-on-narrow-t-y8qk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(widgets): responsive labels collapse to icons on narrow title bars; all widgets pinned by default
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -93,7 +93,7 @@ This means:
 
 Widgets are defined in `agentmux-srv/src/config/widgets.json`. These are the **only** widget types — do not invent or reference widgets that don't exist here.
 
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

a5af:
@codex review

PR #976, opened 2026-05-22
fix(modal): defer singleton-modal label resolution past app boot

AgentA-asaf:
## Boot crash — v0.38.2 portable hung on the loading screen

`frontend/app/store/singleton-modal.ts` ran `void resolveMyLabel()` at **module-eval time**. `resolveMyLabel()` calls `getApi().getWindowLabel()` — but at module-eval the CEF API bridge (`window.api`) isn't set up yet, so `getApi()` returns `undefined` and `.getWindowLabel()` throws an **uncaught `TypeError`**, stalling app init.

```
[getApi] called before window.api exists
Uncaught TypeError: Cannot read properties of undefined (reading 'getWindowLabel')
```

Shipped silently in v0.38.2: tsc can't see runtime-timing, the singleton unit tests mock `getApi`, and no build was smoke-tested with the singleton layer before the portable.

### Fix
- Move the label kick-off out of module-eval into `startSingletonCrashRelease()` — which `app-init` calls *after* `window.api` exists.
- `resolveMyLabel` now guards `getApi() == null` and bails without caching, so it can never throw synchronously (a later post-init call retries).

`tsc` clean; 20/20 singleton tests pass.

### DIFF
--- a/.changesets/1779463930-fix-modal-defer-singleton-modal-label-resolution-past-app-bo-nnnf.md
+++ b/.changesets/1779463930-fix-modal-defer-singleton-modal-label-resolution-past-app-bo-nnnf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(modal): defer singleton-modal label resolution past app boot
--- a/frontend/app/store/singleton-modal.ts
+++ b/frontend/app/store/singleton-modal.ts
@@ -225,6 +225,14 @@ function ensureWired(kind: SingletonKind, st: KindState): void {
     if (st.wired) return;
     st.wired = true;
 
+    // Resolve this window's label here too. A kind is first wired when a
+    // component reads `singletonHolder`/`acquireSingleton` — which only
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
codex's P2 re-fires the previous round's finding. It states *"resolveMyLabel() is now only kicked off from startSingletonCrashRelease()"* — but commit `f7722edb` added exactly the retry path it asks for: **`ensureWired()` now also calls `resolveMyLabel()`**.

A singleton kind is first wired when a component reads `singletonHolder` / `acquireSingleton`, which happens after the app shell renders — `window.api` is ready by then — so this is a guaranteed post-init resolution path, independent of `startSingletonCrashRelease`. `myLabelSync()` therefore cannot stay `null` for the process lifetime.

reagent APPROVED `f7722edb`. Treating this as a stale re-fire — merging.

PR #977, opened 2026-05-22
fix(launch-modal): thread continueOfId through the +New bundle round-trip

AgentA-asaf:
## The bug

When the user opens the launch modal in Continue mode (Feature A auto-selects the most-recent instance, e.g. "Maks"), clicks **+ New identity bundle…**, then **cancels** — the modal returns saying *"Connect to Claude Code"*, even though the continued agent already has a working Claude Code OAuth.

## Root cause

The form snapshot type used for the `+ New bundle` round-trip (`LaunchFormStateWire` / `LaunchFormState`) omitted **`continueOfId`**. On return:

- `form.continueOfId` came back `null`
- `isContinue()` flipped `false`
- `authRequired()`'s `!isContinue()` short-circuit was lost
- For an ambient-creds continuation the carried `identityId` is `""` → `authRequired()` → `true` → "Connect to Claude Code"

The "+New buttons disabled while continuing" comment that justified the omission was **false** for ambient-creds continuations — `continueLocksIdentity()` only locks when the continued row carries a real identity bundle, which Maks's doesn't.

## Fix

- Add `continueOfId: string | null` to `LaunchFormStateWire` (`tab-modal.ts`) and `LaunchFormState` (`AgentLaunchModal.tsx`).
- `snapshot()` captures it.
- `isContinue()` reads `form.continueOfId` directly — decoupled from the `namedAgents()` resource load.
- `viewMode` initializes to `"continue"` when the snapshot has a `continueOfId`.
- Stale comment replaced.

Reducer untouched (`LaunchForm` already has the field; `Opened` already merges `initial` into `form`). **63 launch-flow tests pass**; tsc clean for touched files (the 2 pre-existing `toBeInTheDocument` errors in `AgentLaunchModal.integration.test.tsx` are task #36, unrelated).

Diagnosis: `docs/analysis/LAUNCH_MODAL_CONTINUE_LOST_2026_05_22.md`.

### DIFF
--- a/.changesets/1779493749-fix-launch-modal-thread-continueofid-through-the-new-bundle--b798.md
+++ b/.changesets/1779493749-fix-launch-modal-thread-continueofid-through-the-new-bundle--b798.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(launch-modal): thread continueOfId through the +New bundle round-trip
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.38.1"
+version = "0.38.2"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #969, opened 2026-05-22
feat(agent): OAuth Connect creates a named Identity bundle first

AgentA-asaf:
## Bundle PR 1 — Feature 1 of `SPEC_BUNDLE_MANAGEMENT_2026_05_22`

Clicking **Connect** with no identity selected used to OAuth and auto-mint an *unnamed* identity bundle. Now it opens the **New Identity modal first** (name + description), and the credential attaches to that named bundle.

- `AgentNewIdentityModalPanel` gains `purpose: "create" | "oauth-continue"` — button label "Create" vs **"Continue"**.
- Connect with the `needs-bundle` outcome → chains into the New Identity modal via the existing `tabModal.replace` round-trip → on **Continue**: `UpsertIdentityBundleCommand` → returns to the launch modal with the new identity preselected + `autoStartAuth`.
- A one-shot effect in `PreLaunchAuthPanel` fires `startConnect()` once on mount when `autoStartAuth` is set; OAuth runs with `intoBundleId` = the new bundle.
- `needs-account` (existing bundle lacking a binding) → unchanged, straight to OAuth. Plain `+ New identity` → unchanged (`purpose: "create"`).

`tsc` clean. No backend change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #981, opened 2026-05-23
feat(oauth): bind OAuth credentials into identity bundles on Connect (OAuth Bundles PR C)

AgentA-asaf:
## OAuth Bundles PR C

Third of the 6-PR sequence (`SPEC_OAUTH_IDENTITY_BUNDLES_2026_05_22.md` §4.5 + §4.6). The OAuth-success invariant: **every successful OAuth flow now lands bound to an identity bundle**, with the CLI spawned against the bundle's per-provider config dir and `SecretRef::OAuthConfigDir { dir }` persisted on success.

### What changed

**Backend — `agentmux-srv/src/server/identity_handlers.rs`** (+584/-)

Two helpers, used by both `spawn_auth_cli` (pipes) *and* `spawn_auth_cli_pty`:

- `compute_and_ensure_bundle_dir(bundle_id, provider, env)`
  - Gated on `resolver::provider_class` — only **oauth-class** providers (claude / codex / openclaw, per spec §4.3) get the per-bundle override; api-key-class providers (e.g. kimi's `KIMI_SHARE_DIR`) bypass.
  - Reads `auth_config_dir_env_var` + `auth_dir_name` from the canonical CLI registry (`backend::providers::get_provider(id)`) — single source of truth, no hardcoding.
  - Computes `DataPaths::identity_dir(bundle_id).join(auth_dir_name)` (e.g. `~/.agentmux/shared/identities/<id>/claude/`), `create_dir_all`s it, **overrides** the env var the frontend may have computed via the legacy ambient `ensureAuthDir` path.
  - Soft-fails (returns `None`, logs a warn) on every error path → legacy ambient behavior preserved.

- `persist_oauth_binding_or_synthetic(...)` — called at the two success transition points in both spawn variants:
  - Upserts an `IdentityAccount { kind: "oauth", secret_ref: SecretRef::OAuthConfigDir { dir }, status: "valid" }`.
  - `bundle_identity_bind(bundle, provider, account)`.
  - Publishes `identitybundlebindings:changed:<bundle_id>` (the launch modal already subscribes).
  - On any per-binding failure, falls back to the legacy `pending-bundle-for-<sid>` synthetic id (no regression).

**Frontend — `PreLaunchAuthPanel.tsx`** (+50/-)

- `handleConnect`: refuses to start OAuth if `identityId === ""` and no `onRequestNewIdentity` callback is wired. Closes the invariant hole where a future parent could forget the callback and silently fall back to ambient OAuth. The wired path (PR #969) already routes through the New Identity modal first.

### Tests

- New: `persist_oauth_binding_round_trip` (full happy-path: dir computed + ensured + env var overwritten; binding persisted with the exact `SecretRef::OAuthConfigDir { dir }` and `status: "valid"`).
- Plus 3 guard tests for ambient / api-key / no-dir paths.
- **`cargo test -p agentmux-srv --bin agentmux-srv` — 1077 passed, 0 failed.**
- **`npx vitest run frontend/app/view/agent/auth/` — 64 passed.**
- `npx tsc --noEmit` clean on touched files.

### Open uncertainty (deferred — PR D / E)

The CLI-success signal is exit-0 + `confirm_authenticated`'s login-success pattern (already what PR A used). If a provider's CLI exits clean but writes its tokens with a delay or under a different subpath of the config dir, the bundle dir will hold an empty/partial set on the next launch and the agent will re-OAuth. PR D (status probe + Reconnect) and the migration in PR E are where that's hardened — out of scope here.

### DIFF
--- a/.changesets/1779497458-feat-oauth-bind-oauth-credentials-into-identity-bundles-on-c-3npi.md
+++ b/.changesets/1779497458-feat-oauth-bind-oauth-credentials-into-identity-bundles-on-c-3npi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(oauth): bind OAuth credentials into identity bundles on Connect (oauth bundles PR C)
--- a/agentmux-common/src/data_paths.rs
+++ b/agentmux-common/src/data_paths.rs
@@ -262,18 +262,27 @@ fn resolve_root() -> Result<PathBuf, String> {
 /// not a literal filename, so `PathBuf::join("versions").join("C:temp")`
 /// would resolve OUTSIDE the intended `~/.agentmux/versions/` subtree.
 fn sanitize_path_segment(s: &str) -> Option<String> {
-    let trimmed = s.trim();
-    if trimmed.is_empty() || trimmed == "." || trimmed == ".." {
+    // Reject whitespace padding rather than silently normalizing it
+    // away — otherwise distinct caller-supplied ids like "foo" and
+    // " foo " would alias to the same directory. bundle_id is a real
+    // caller-supplied identifier (passed through RPC payloads), so
+    // this matters for credential isolation. codex P2 follow-up on
+    // #981. Internally-generated version strings + branch names
+    // shouldn't carry padding anyway, so this is no-op for them.
+    if s != s.trim() {
+        return None;
+    }
+    if s.is_empty() || s == "." || s == ".." {
         return None;
     }
     // Filesystem separators + Windows-reserved characters + NUL.
-    if trimmed
+    if s
         .chars()
         .any(|c| matches!(c, '/' | '\\' | ':' | '*' | '?' | '"' | '<' | '>' | '|' | '\0'))
     {
         return None;
     }
-    Some(trimmed.to_string())
+    Some(s.to_string())
 }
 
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
codex's P2 here re-fires the previous round's finding. The comment claims `sanitize_path_segment` "canonicalizes bundle IDs via trimming" — but commit `db98d150` made it **reject** whitespace-padded input instead:

```rust
// data_paths.rs (current)
if s != s.trim() {
    return None;
}
if s.is_empty() || s == "." || s == ".." {
    return None;
}
// …no .trim() anywhere; the rest of the function uses `s` directly.
```

Distinct ids like `"foo"` and `" foo "` no longer alias — the padded one returns `None`, the handler then takes the legacy-ambient fallback. The pinned test `identity_dir_rejects_unsafe_segments` covers `"   "` (whitespace-only), and the rule extends to padded inputs by construction.

reagent APPROVED `db98d150`. Treating as a stale re-fire — merging.

PR #986, opened 2026-05-23
fix(agent-pane): PTY cols follow pane width + bump default to 200

AgentA-asaf:
## Closes task #59 — diagnosed in `docs/analysis/AGENT_PANE_PTY_WRAP_2026_05_23.md`

The agent pane's PTY was hardcoded to `cols: 80`. Tools the CLI shells out to (`git log`, `ls`, …) inherit the PTY's column count and wrap their output to 80 — embedding real `\n`s into the byte stream the live-log captures. CSS soft-wrap can't undo that; the wraps are in the data.

### What landed

- **`frontend/app/view/agent/hooks/usePtyWidth.ts`** (new) — `ResizeObserver` on the agent-view root, computes `cols = floor((width - 16) / (font-size × 0.6))` clamped to ≥ 40, debounces to 150 ms, sends `ControllerInputCommand({ termsize: { rows, cols } })` on change.
- **`frontend/app/view/agent/agent-view.tsx`** — wires the hook to `rootRef`. First fire happens shortly after mount so the agent sees the real cols before its first tool invocation.
- **`agentmux-srv/src/backend/blockcontroller/shell.rs`** — initial PTY `cols: 80 → 200`. Quick-win floor for the brief window between mount and the first ResizeObserver fire, and a saner default for users still on the static path.

The backend resize handler already existed (`shell.rs:842-852`) — only the wiring was missing.

### Wire format note

Wire field is `termsize` (not `term_size` — that's the Rust struct's snake_case in the source; the TS `CommandBlockInputData` type uses the lowercase form). Agent caught this when grepping for the existing call shape.

### Caveat

**Already-captured lines stay wrapped.** Reflowing historical PTY output would corrupt anything that used the original width for layout (e.g. `git log --graph` ASCII art, aligned tables). Only output produced AFTER the resize matches the new width.

### Verification

- `npx tsc --noEmit` — clean on touched files.
- Smoke test deferred to merge-time (needs `task dev` + a manual pane resize).

### DIFF
--- a/.changesets/1779525893-fix-agent-pane-pty-cols-follow-pane-width-bump-default-to-20-h8tl.md
+++ b/.changesets/1779525893-fix-agent-pane-pty-cols-follow-pane-width-bump-default-to-20-h8tl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): PTY cols follow pane width + bump default to 200
--- a/agentmux-srv/src/backend/blockcontroller/shell.rs
+++ b/agentmux-srv/src/backend/blockcontroller/shell.rs
@@ -400,11 +400,18 @@ impl Controller for ShellController {
             return Ok(());
         }
 
-        // Real PTY path
+        // Real PTY path.
+        //
+        // Initial cols bumped to 200 (was 80) so the agent-pane live-log
+        // doesn't render hard-wrapped at ~80 chars before the frontend's
+        // dynamic resize lands (see hooks/usePtyWidth.ts and
+        // docs/analysis/AGENT_PANE_PTY_WRAP_2026_05_23.md). The dynamic
+        // resize coalesces over a ~150 ms debounce after mount, so the
+        // PTY uses this default for the very first batch of output.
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex this looks like a stale re-issue of the prior finding. The fix from commit `7b734a67` (HEAD) does exactly what you're asking:

`frontend/app/view/agent/hooks/usePtyWidth.ts:80-117` — `send(cols, retriesLeft = 2)`:
- `lastCols` is set ONLY inside `.then()` (line 95-97), after the RPC resolves successfully.
- `.catch()` (line 98-115) recurses with `retriesLeft - 1` and a 500ms / 1000ms backoff, so an initial 'controller is not running' triggers up to 2 retries before logging.
- The mount path calls `compute()` once and the `ResizeObserver` calls it on every width change. Both go through the same `send` with retries.

So the controller-not-running window IS retried. Could you re-check against the current HEAD? If there's a different code path you're seeing that I missed, please anchor it.

PR #989, opened 2026-05-23
fix(agent-pane): migrate throwing dispatch sites + capture uncaught DOM errors

AgentA-asaf:
Quick-win for the agent-pane cascade leak class (#878, retro
\`docs/retro/retro-agent-pane-cascade-replacechild-2026-05-23.md\`).

## What this PR does

1. **Migrates three remaining throwing dispatch sites to the soft variant** —
   the last holdouts after PR #878:
   - \`useAgentCommands.ts:267\` PendingMessageQueued
   - \`useAgentCommands.ts:353\` RequestStop
   - \`useHistoryPagination.ts:138\` InitStart

   All three now use \`dispatchPaneIfRegistered\`. Each site has an inline
   pointer at the 2026-05-23 retro.

2. **Adds a global uncaught-error forwarder** (\`frontend/log/error-forwarder.ts\`)
   that captures \`error\` + \`unhandledrejection\` events and forwards them
   through \`fe_log_structured\` with \`{tag, name, message, stack, source}\`.
   Wired in \`bootstrap.ts\` right after \`initLogPipe()\`. Does NOT call
   \`preventDefault()\` — DevTools / SolidJS \`ErrorBoundary\` still see it.

   Replaces the weaker \`window.addEventListener("error", …)\` listeners that
   were at the bottom of \`bootstrap.ts\` (no stack, no name, concatenated
   message). The new forwarder is strictly stronger.

## Why this matters

The 2026-05-23 incident showed \`CASCADE_DETECTED\` (the PR #878 detection)
firing on \`StreamFlush\` but no \`replaceChild\` line in either log — the
visible host modal had no diagnostic. With this forwarder, the next
incident lands with full stack in the host log.

The dispatch-site migration closes a class of \`dispatch for unregistered
pane\` throws even when the cascade still happens.

## Out of scope (next PRs)

- PR-3: unified per-pane registration (atomic across both stores)
- PR-4: model-level \`dispatchIfAlive\` helper with disposed flag

## Verification

- 0 new TS errors on touched files.
- \`vitest run frontend/app/store + frontend/app/view/agent\`: 588/590 pass.
  Two failures are pre-existing on main (\`browser-pane-state-store.test.ts\`,
  \`providers/index.test.ts\`).
- The 5-test cascade detection suite passes.

Follow-up to retro \`docs/retro/retro-agent-pane-cascade-replacechild-2026-05-23.md\`.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

[...]

PR #999, opened 2026-05-23
feat(agent-pane): unified per-pane registration across both stores

AgentA-asaf:
Cascade follow-up PR-3 (2 of 4). Closes the half-registered window between `agent-document-store` and `agent-pane-state-store` registrations — the structural root of the cascade-mid-dispatch failure mode PR #878 added detection for and PR #989 partially mitigated.

Refs:
- `docs/analysis/LIFECYCLE_DISPATCH_LEAK_2026_05_15.md` §6.2 follow-up
- PR #878 — cascade detection instrumentation + soft-dispatch variants
- PR #989 — migrated three throwing dispatch sites to the soft variant

## The problem this closes

Before this PR, `agent-view.tsx` registered each pane with TWO independent stores in sequence:

```ts
registerAgentDocPane(blockId, setter);                        // store A
onCleanup(() => unregisterAgentDocPane(blockId));
registerAgentPaneStatePane(blockId, agentId, projections);    // store B
onCleanup(() => unregisterAgentPaneStatePane(blockId));
```

Two distinct register calls, two distinct `onCleanup` callbacks. A `StreamFlush` dispatch (from `useAgentStream`, hooked subscriptions, or any hot dispatch loop) landing in the synchronous gap between the two registers — or in the asymmetric teardown order during cleanup — found the pane registered in one store but not the other. The dispatch into the missing slot threw, tore the reactive graph, and surfaced as `replaceChild NotFoundError` in the UI.

PR #878 added cascade DETECTION. PR #989 migrated three throwing dispatch sites to the soft `dispatchIfRegistered` variant. PR-3 closes the half-registered window STRUCTURALLY.

## The fix

New module `frontend/app/store/agent-pane-registration.ts` is the unified surface:

```ts
export function registerPane(blockId: string, reg: PaneRegistration): void;
export function unregisterPane(blockId: string): void;
```

- `registerPane` atomically registers in BOTH stores. If the second store's register throws, the first is rolled back so observers never see the half-state.
- `unregisterPane` atomically tears down BOTH stores. Teardown order is pane-state first, document second — the intermediate "pane-state gone, document still there" state is safe for the soft-dispatch variants from PR #878 / #989 (they return `[]`, don't throw).

agent-view.tsx migrated to call the unified helper once, in a single `onCleanup`.

## Contract

> At any observable point in time, the pane is either FULLY registered in BOTH stores or FULLY unregistered from BOTH. There is no half-registered state from a dispatcher's POV.

`frontend/app/store/agent-pane-registration.test.ts` (14 tests) pins this down with both an invariant-over-lifecycle test (`isPaneHalfRegistered === false` at every observable moment across `register → unregister → re-register → unregister`) and snapshot probes against both underlying stores.

## Option A/B choice

**Picked Option A:** per-store `registerPane` / `unregisterPane` remain exported but are documented `@internal — tests only`. The per-store cascade tests added in PR #878 (`agent-pane-state-store.test.ts`) drive single-store cascade-disposal scenarios that require direct register/unregister access with custom projection setters that synchronously call `unregisterPane` to simulate a reactive subscriber mid-dispatch dispose. Forcing those through the unified helper would either:
- bring up a full second-store slot (changes what the test is testing), or
- require a test-only escape hatch (defeats Option B's enforcement point).

Marking the per-store exports `@internal — tests only` and documenting that ALL production callers go through the unified helper gives the same structural guarantee for the actual failure mode (the half-registered window in agent-view's lifecycle) without the test thrash.

PR-4 of the cascade sequence (model-level `dispatchIfAlive`) may make the per-store cascade tests obsolete — at which point the Option B flip (rename to `_register*`, no export) becomes cheap.

## Files

- new: `frontend/app/store/agent-pane-registration.ts`
- new: `frontend/app/store/agent-pane-registration.test.ts`
- modified: `frontend/app/store/agent-document-store.ts` (`@internal` docs on register/unregister)
- modified: `frontend/app/store/agent-pane-state-store.ts` (`@internal` docs on register/unregister)
- modified: `frontend/app/view/agent/agent-view.tsx` (use unified helper; single onCleanup)

## Out of scope

- The `dispatchIfAlive` model-level pattern (PR-4, separate).
- View-level migration of dispatch sites.
- Cleaning up the cascade detection logging from PR #878 — kept as-is.

## Tests

- `npx vitest run frontend/app/store/agent-pane-registration.test.ts` — 14/14 pass.
- `npx vitest run frontend/app/store/agent-pane-state-store.test.ts frontend/app/store/agent-pane-state/ frontend/app/store/agent-document/` — 188/188 pass.
- `npx vitest run frontend/app/view/agent/` — 271/272 pass (one pre-existing unrelated failure in `providers/index.test.ts` for the acp/api-key check, confirmed present on `main`).
- `npx tsc --noEmit` — no new TS errors in the changed files.

### DIFF
--- a/.changesets/1779547879-feat-agent-pane-unified-per-pane-registration-22id.md
+++ b/.changesets/1779547879-feat-agent-pane-unified-per-pane-registration-22id.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): unified per-pane registration
--- a/frontend/app/store/agent-document-store.ts
+++ b/frontend/app/store/agent-document-store.ts
@@ -67,21 +67,30 @@ export function setEventSink(sink: EventSink): void {
 }
 
 /**
- * Register a pane with the store. Call from the agent-view component
- * onMount, passing the pane's documentAtom setter. The slot owns the
- * pane's reducer state and writes through the setter on each mutation.
+ * Register a pane with the store. The slot owns the pane's reducer
+ * state and writes through the setter on each mutation.
  *
  * Idempotent: re-registering a blockId resets its state to initialState
  * and rebinds the setter. (Useful for hot-reload scenarios.)
+ *
+ * @internal — production callers MUST use `registerPane` from
+ * `agent-pane-registration.ts` so the pane is registered atomically
+ * across BOTH stores (document + pane-state). Direct callers of this
+ * function are limited to single-store unit tests (cascade-detection
+ * scenarios that need a custom single-store projection). PR-3 of the
+ * cascade follow-up sequence — see agent-pane-registration.ts for
+ * rationale + Option A/B discussion.
  */
 export function registerPane(blockId: string, setter: DocumentSetter): void {
     slots.set(blockId, { state: initialState(), setter });
 }
 
 /**
- * Release a pane's slot. Call from agent-view component onCleanup.
- * Failure to release leaks a state cell — every register MUST have a
- * matching unregister.
+ * Release a pane's slot. Failure to release leaks a state cell —
+ * every register MUST have a matching unregister.
+ *
+ * @internal — see `registerPane` above. Production code uses
+ * `unregisterPane` from `agent-pane-registration.ts`.
  */
 export function unregisterPane(blockId: string): void {
     slots.delete(blockId);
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1000, opened 2026-05-23
feat(agent): recent sessions tab in launch modal for reattaching previous conversations

AgentA-asaf:
Cascade follow-up **4 of 4** — see retro
[`docs/retro/retro-agent-pane-cascade-replacechild-2026-05-23.md`](docs/retro/retro-agent-pane-cascade-replacechild-2026-05-23.md)
and the recovered transcript at
[`docs/recovery/MAKS_CONVERSATION_2026_05_23.md`](docs/recovery/MAKS_CONVERSATION_2026_05_23.md).

Previous cascade follow-ups:
- PR #977 — `continueOfId` round-trip through the +New bundle flow (foundation we extend)

## The problem

On 2026-05-23 a renderer cascade ate the agent pane mid-session. The
conversation file (`output.state.json`) was still in `filestore.db`
under the original `blockId`, but the only way to recover it would
have been to know the blockId and force a remount on the original
pane. There was no UI path. The user lost their working context.

## User flow

1. User opens an agent pane. AgentPicker renders the list of
   definitions.
2. **NEW:** ABOVE the agent cards, a "Recent sessions" list shows up
   to 20 entries — each one is a clickable row with:
   - provider icon + instance name + identity name
   - first-user-message preview (newline-collapsed, ≤240 chars,
     skipping bootstrap "Session Context" boilerplate when a later
     message exists)
   - "X messages • Ym ago"
3. Clicking a row launches a new pane that reattaches to the prior
   working directory + identity + memory; the conversation history
   reloads from `filestore.db` on first paint via the existing
   `useHistoryPagination` snapshot fast-path.
4. Empty state branches by identity filter — picker is generic
   (no filter), but the component supports filtered surfacing for any
   future per-identity caller.

## Reattach mechanism

Using the existing `continueOfInstanceId + workDirOverride` flow that
PR #977 wired through the launch modal — same backend RPCs, same
`CreateAgentInstance` lineage, same env injection. The new pane
spawns the CLI in the prior working directory; Claude's `--continue`
(et al.) resumes the session; the new pane's `output.state.json`
loads on first paint.

**This is NOT a "ghost block" remount of the original `blockId`.**
Each entry produces a new block tied to the prior conversation via
the continuation chain. The original `block_id_hint` is preserved on
the row for diagnostics + future "reattach by blockId" hamburger
debug paths (out of scope here).

### Backend (`agentmux-srv`)
- New `listrecentsessions` RPC (`agent_handlers.rs`):
[...]

PR #998, opened 2026-05-23
feat(block): per-block error boundary + localized reload UI

AgentA-asaf:
## Summary

Wrap each block (pane) in its own SolidJS `<ErrorBoundary>` so a renderer exception in one pane only blanks THAT pane. Other panes in the same tab + the layout chrome stay alive.

**Cascade follow-up 1 of 4.** See retro: `docs/retro/retro-agent-pane-cascade-replacechild-2026-05-23.md`.

## Background

On 2026-05-23, a SolidJS reactive cascade in a single agent pane took out the ENTIRE tab content area because the only error boundary that caught it was at the workspace level. After F5, the user saw status bar + tab bar but a blank content region with red error text — couldn't recover the pane individually. They had to close the tab.

Prior analysis: `docs/analysis/LIFECYCLE_DISPATCH_LEAK_2026_05_15.md` (PR #878 fixed the dispatch-after-cleanup half of the cascade class; this PR adds the defensive boundary so an unforeseen renderer fault never blanks the whole tab again).

## User-visible recovery flow

When a pane's renderer throws (mount, update, anything):

1. The crashed pane's body is replaced with a localized panel: red border, warning icon, "This pane crashed" headline, the error message, and a collapsible stack section.
2. The pane header, sibling panes, layout chrome, tab bar, and status bar **stay alive**.
3. User clicks **Reload pane** → SolidJS `reset` re-mounts the children fresh; reactive owners torn down and recreated. If the crash was transient, the pane recovers.
4. Or user clicks **Close pane** → existing `nodeModel.onClose` destroys the block.

## Logging

Every catch forwards a structured payload via `fe_log_structured` (the same IPC channel `frontend/log/log-pipe.ts` and `frontend/log/error-forwarder.ts` use):

```
level: "error"
module: "block-error-boundary"
data: { block_id, view_type, error_name, error_message, error_stack }
```

This is the "next incident lands with a stack trace" insurance — the host log now correlates cascade-detection warnings with the exact pane that lost its render tree.

## Cascade-safety contract

The fallback UI **reads only from props** (`blockId`, `viewType`, `error`, `reset`, `onClose`). It deliberately does NOT subscribe to:
- the agent-pane-state-store
- the agent-document-store
- any per-pane atom

The broken pane's reactive graph may be half-flushed at catch time; any subscription into it from the fallback would re-throw.

Logging is fire-and-forget — an `invokeCommand` throw cannot break the fallback (covered by a regression test).

## Defense in depth

[...]

PR #1002, opened 2026-05-23
feat(agent-pane): model-level dispatchIfAlive with disposed flag (cascade follow-up PR-4)

AgentA-asaf:
## Summary

Adds a per-pane `AgentPaneModel` whose lifecycle is owned by `agent-pane-registration.ts`. The model carries a `disposed` flag flipped synchronously inside `unregisterPane` **before** either underlying store deletes its slot. Dispatches through `model.dispatchPane` / `model.dispatchDoc` after dispose silently no-op (return `[]`, emit one debug log); before dispose they forward to the per-store `dispatchIfRegistered`.

Final PR (4 of 4) in the cascade-followup series:

| # | PR | What |
|---|---|---|
| 1 | #989 | Soft-variant migration for the three known throwing dispatch sites + uncaught-error forwarder |
| 2 | #998 | Per-block `<ErrorBoundary>` — a cascade no longer takes out the tab |
| 3 | #999 | Unified per-pane registration (closes the half-registered window) |
| 4 | **this PR** | Model-level `dispatchIfAlive` with `disposed` flag — closes the deferred-dispatch race structurally |

## Why this PR

PRs #878 and #999 close the **in-frame** cascade window — a setter that synchronously disposes the pane mid-dispatch is detected, and re-registration is atomic across both stores. But a **deferred** dispatcher — a `setTimeout` watchdog tick, an `await` continuation, a subscription handler that fires from a stale closure — can land AFTER `unregisterPane` completes. Before this PR each such call site had to remember to use `dispatchIfRegistered` (the soft variant). New code is one missed import away from re-introducing the throw.

This PR makes new code **default-safe**: hooks gain a `model` opt; their dispatches go through `model.dispatchPane(cmd)` / `model.dispatchDoc(cmd)`, which gate on `disposed` before forwarding. The disposed flag is the structural guarantee — no reviewer-vigilance dependency.

The two safety nets compose without overlap: cascade detection (PR #878) identifies the *trigger setter* when a dispatch lands mid-frame; the disposed flag stops a *post-cleanup* dispatch from reaching the underlying store at all.

## Model API

```ts
interface AgentPaneModel {
    readonly blockId: string;
    readonly disposed: boolean;
    dispatchPane(cmd: AgentPaneCommand, source?: CommandSource): AgentPaneEvent[];
    dispatchDoc(cmd: AgentDocumentCommand, source?: CommandSource): AgentDocumentEvent[];
}
```

`registerPane(blockId, reg)` now **returns** the model. `unregisterPane(blockId)` flips `disposed` first, then deletes both store slots.

## Coexistence (Option B-coexist)

Per-store `dispatchIfRegistered` exports stay live. Reasons:
- Some dispatchers don't have a model handle (the cascade-detection logging inside the stores; tests; future code that doesn't thread the model).
- Migration can land incrementally — each hook that gains a `model` opt flips its own dispatches.

The model is the **preferred** path; new code defaults to it. The cleanup PR that removes `dispatchIfRegistered` is downstream, gated on every production call site migrating.

This is consistent with PR #999's Option A choice (keep per-store register/unregister live but mark them `@internal — tests only`).

## Files migrated

Three hooks gain a `model: AgentPaneModel` option and replace all `dispatchIfRegistered` call sites with `model.dispatchPane` / `model.dispatchDoc`:

- `frontend/app/view/agent/useAgentStream.ts` (15 sites)
- `frontend/app/view/agent/hooks/useAgentCommands.ts` (5 sites)
- `frontend/app/view/agent/hooks/useHistoryPagination.ts` (9 sites)

`agent-view.tsx` captures the model returned by `registerPane` and threads it into the three hooks.

The `dispatchDocIfRegistered` call site at `agent-view.tsx:315` (the `/login` success path) is intentionally **not** migrated in this PR — it sits outside any of the three migrated hooks and would require a sibling change to `useAgentControllerStatus`'s `onLoginSuccess` opt. Out of scope per the spec ("focus on the three retro-identified PR-2 holdouts").

## Pattern source

`frontend/app/view/drone/drone-model.ts` (see `DroneViewModel.disposed` + `DroneViewModel.dispatchIfAlive` at lines ~131 and 509–515) — same disposed-flag rule applied to one store. This PR generalizes it across both agent-pane stores in one helper.

## Tests

**18 new tests** in `frontend/app/store/agent-pane-model.test.ts`:

- construction + lifetime (5)
- disposed flag flips BEFORE store unregisters (1)
- `dispatchPane` drop semantics — after + before dispose (2)
- `dispatchDoc` drop semantics — after + before dispose (2)
- torture test — `setTimeout` + `await`-continuation post-dispose drops (2)
- coexistence with module-level `dispatchIfRegistered` (2)
- idempotency + over-cleanup tolerance (2)
- model type surface (1)
- happy-path quiet (1)

All 38 tests across the three pane-store test files pass (`agent-pane-model.test.ts`, `agent-pane-registration.test.ts`, `agent-pane-state-store.test.ts`). The two pre-existing test failures on main (`browser-pane-state-store.test.ts` and `providers/index.test.ts`) are unrelated to this PR.

## References

- `docs/retro/retro-agent-pane-cascade-replacechild-2026-05-23.md` — retro that scoped this PR sequence
- `docs/analysis/LIFECYCLE_DISPATCH_LEAK_2026_05_15.md` — original analysis of the cascade class
- `frontend/app/view/drone/drone-model.ts` — pattern source
- PR #989, #998, #999, #1000 — preceding cascade-followup PRs

### DIFF
--- a/.changesets/1779549929-feat-agent-pane-model-level-dispatchifalive-helper-h6ae.md
+++ b/.changesets/1779549929-feat-agent-pane-model-level-dispatchifalive-helper-h6ae.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): model-level dispatchIfAlive helper
--- a/frontend/app/store/agent-pane-model.test.ts
+++ b/frontend/app/store/agent-pane-model.test.ts
@@ -1,0 +1,382 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * AgentPaneModel — PR-4 of the cascade follow-up.
+ *
+ * Contract under test: the model exposed by `registerPane` carries a
+ * `disposed` flag that flips synchronously inside `unregisterPane`
+ * BEFORE the underlying store deletes run. Dispatches through the model
+ * after dispose silently no-op (no throw, return []) and emit a single
+ * debug log line. Dispatches before dispose forward to the underlying
+ * store's `dispatchIfRegistered` and update reducer state normally.
+ *
+ * This is the second safety net on top of cascade detection (PR #878)
+ * and unified atomic registration (PR #999). The disposed flag catches
+ * deferred dispatchers — setTimeout, await continuations, subscription
+ * handlers — that land AFTER the cleanup runs.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import {
+    __resetAllSlots as resetDocSlots,
+    snapshot as docSnapshot,
+} from "./agent-document-store";
+import {
+    __resetAllSlots as resetPaneStateSlots,
+    type AgentPaneProjections,
+    snapshot as paneStateSnapshot,
+} from "./agent-pane-state-store";
+import {
+    type AgentPaneModel,
+    type PaneRegistration,
+    getPaneModel,
+    registerPane,
+    unregisterPane,
+} from "./agent-pane-registration";
+
+function noopProjections(): AgentPaneProjections {
+    return {
+        streaming: () => {},
+        sessionStats: () => {},
+        currentTool: () => {},
+        turnTokens: () => {},
+        pending: () => {},
+        initPhase: () => {},
+        turnPhase: () => {},
+    };
+}
+
+function defaultRegistration(): PaneRegistration {
+    return {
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #1004, opened 2026-05-23
fix(build): break \-hex escape in Maks transcript that crashes tailwind v4

AgentA-asaf:
**P0** — `task build:frontend` is broken on `main` because Maks' recovered conversation transcript contains a Windows path with `\d85077b2-...`. Tailwind v4's CSS-escape scanner reads `\d85077` as a 6-digit codepoint literal, parses it to 0xD85077 = 14176375, calls `String.fromCodePoint`, and crashes.

Fix: normalize single backslashes to forward slashes in transcript text. `scripts/redact-pii.mjs` gains the rule so future transcripts are sanitized at capture. The existing transcript is re-run through the script.

`task build:frontend` verified green after the change.


a5af:
@codex review

PR #1007, opened 2026-05-24
feat(srv): agent-anchored session zones — backend (Option E, PR 1 of 2)

AgentA-asaf:
**Backend-only** implementation of agent-anchored session zones — first of two PRs for Option E.
The frontend wiring (reads + UX) lands as a follow-up (E2). Cross-tab live sync (E3) is out of scope.

## Spec

https://github.com/agentmuxai/agentmux/blob/agenta/spec-continuation-session-persistence/docs/specs/SPEC_CONTINUATION_SESSION_PERSISTENCE_2026_05_23.md

The critical scoping invariant from the spec: a session is bound to the **agent definition_id**, NOT to the identity bundle. Two agents that share an OAuth bundle keep separate session histories. This PR enforces that at the storage layer.

## Zone naming

- Active:   `agent:<definition_id>:current`
- Archived: `agent:<definition_id>:archive:<unix_ms>`

Files in the zone keep the existing shape (`output.state.json`, `output`). `definition_id` is restricted to `[A-Za-z0-9_-]+` to prevent zone-name injection — UUIDs (the production shape) are a strict subset.

## New RPCs

| Command | Purpose |
|---|---|
| `agent:session:read` `{definition_id}` → `{content: Option<String>, modts: Option<i64>}` | Read `output.state.json` from `:current`. Returns `{content: None}` (NOT error) when the zone doesn't exist. |
| `agent:session:write_state` `{definition_id, content}` → `{bytes_written}` | Write `output.state.json` to `:current` (idempotent; creates zone if missing). |
| `agent:session:append_output` `{definition_id, line}` → `{bytes_written}` | Append a single NDJSON line (normalized with trailing `\n`) to `output`. |
| `agent:session:archive` `{definition_id}` → `{archive_zoneid, archived_at_ms}` | Snapshot `:current` → `:archive:<now_ms>`, then clear `:current`. No-op (empty `archive_zoneid`) when `:current` is empty / zero-byte. Write-then-clear ordering: a crash mid-archive leaves both populated, never `:current` cleared with archive missing. |
| `agent:session:list_archives` `{definition_id, limit?}` → `[{archive_zoneid, archived_at_ms, preview, node_count}]` | Lists archive zones newest-first. Default limit 20, cap 100. Mirrors `RecentSessionRow` preview fields. |

## CreateAgentInstance change

When `block_id` is non-empty in the request, the handler stamps `agent:sessionZone = "agent:<defId>:current"` onto the block meta. Failures here are non-fatal (log + continue) — the instance row is the source of truth, the meta stamp is a frontend convenience. Continuation now becomes structural (same zone, different block) rather than parametric (per-block snapshot copy + `--continue`).

## One-shot migration (`migrate_block_zones_v1`)

Hooked into srv startup in `main.rs`, gated by `<data_dir>/migration_agent_zones_v1.flag`:

1. For each `db_block` with `meta.view = "agent"` and a non-empty `output.state.json` in its block-keyed zone, copy contents to `agent:<defId>:archive:<block-snapshot-createdts>`.
2. For each definition, seed `:current` from the most-recently-modified per-block snapshot.
3. Per-block failures log + continue (best-effort, never aborts startup).
4. Marker is written even on partial failure so we don't retry indefinitely — operators can delete it to force a re-run.
5. Old per-block zones are NOT deleted. GC is a later PR.

Second startup = no-op (gated by marker).

Agent definition is detected via `meta.agentId` (or legacy `meta.agent:id`).

## Tests

18 new unit + integration tests in `agentmux-srv/src/backend/agent_session.rs`:

- Zone-naming helpers + injection rejection (5 tests including ASCII/Unicode/path-separator/control-char cases).
- `read` returns `None` (not error) on missing zone.
- Write/read roundtrip + idempotency.
- `archive_session` clears `:current`, no-ops on empty `:current` AND on zero-byte file.
- **Two agents on the same identity bundle have independent zones** (the critical scoping invariant).
- `list_archives` sorted newest-first with previews + limit enforcement.
- Migration: backfills archives, seeds `:current` from most-recent snapshot, isolates per-definition, leaves old block zones intact.
- Migration is idempotent (gated by marker file).
- Migration skips non-agent blocks and empty-snapshot blocks.

## Test gates

- `cargo test -p agentmux-srv --bin agentmux-srv` — all green (1120 passed; 0 failed; 1 ignored).
- `cargo build -p agentmux-srv --release` — succeeds.
- No frontend changes (verified via `git status`).

## Discipline notes

- No version bump (uses changeset workflow per RFC #857 Phase 2).
- AgentA identity used for the branch and commit.
- Two pre-existing flaky tests (`handler_returns_sessions_with_previews_sorted_by_snapshot_first`, `handler_respects_limit`) reproduce on `main` and are unrelated to this change — they depend on `FileStore::now_ms()` returning strictly increasing values for back-to-back writes. Not fixed in this PR to keep scope tight.

### DIFF
--- a/.changesets/1779584279-feat-srv-agent-anchored-session-zones-backend-1ter.md
+++ b/.changesets/1779584279-feat-srv-agent-anchored-session-zones-backend-1ter.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): agent-anchored session zones (backend)
--- a/agentmux-srv/src/backend/agent_session.rs
+++ b/agentmux-srv/src/backend/agent_session.rs
@@ -1,0 +1,998 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Agent-anchored session zones (Option E, PR 1 of 2).
+//!
+//! A session zone is bound to the **agent definition** (`definition_id`),
+//! not the identity bundle the agent currently uses. Two agents that
+//! share an identity bundle keep separate session histories.
+//!
+//! ## Zone naming
+//!
+//! - Active:   `agent:<defId>:current`
+//! - Archived: `agent:<defId>:archive:<unix_ms>`
+//!
+//! Files inside a zone keep the existing shape:
+//! `output.state.json` (full UI snapshot) and `output` (raw NDJSON
+//! stream for crash-recovery replay).
+//!
+//! ## Migration (`migrate_block_zones_v1`)
+//!
+//! On first srv startup after this PR ships, the per-block zones (the
+//! pre-Option-E layout where each `blockId` owned its own zone) are
+//! back-filled into the new per-agent layout:
+//!
+//! 1. For each `db_block` row with `meta.view = "agent"` and a
+//!    non-empty `output.state.json` in its block-keyed zone, copy the
+//!    contents to `agent:<defId>:archive:<block-zone-createdts>`.
+//! 2. For each definition, copy the most-recently-modified per-block
+//!    snapshot to `agent:<defId>:current`.
+//! 3. Write the marker file `<data_dir>/migration_agent_zones_v1.flag`.
+//!
+//! The migration is read-only against the existing per-block zones —
+//! GC of those is a later PR.
+//!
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1008, opened 2026-05-24
feat(agent): default-continue from agent session zones — frontend (Option E, PR 2 of 2)

AgentA-asaf:
**Spec:** [SPEC_CONTINUATION_SESSION_PERSISTENCE_2026_05_23.md](https://github.com/agentmuxai/agentmux/blob/agenta/spec-continuation-session-persistence/docs/specs/SPEC_CONTINUATION_SESSION_PERSISTENCE_2026_05_23.md)

**Follows up:** #1007 (Option E backend — agent-anchored session zones)

Option E PR 2 of 2 — frontend. Wires the agent-anchored session zones from #1007 into the document-history fast path and adds a default-continue interaction on the agent picker.

### Snapshot read/write switched from per-block to per-agent zone
- `useHistoryPagination` now reads the document snapshot from `agent:<defId>:current` via the new `agent:session:read` RPC (with a fallthrough to the per-block NDJSON replay when the agent zone is empty — covers fresh agents and the picker screen).
- `agent-view.tsx`'s 30 s snapshot writer + on-close save now write to the same agent zone via `agent:session:write_state`.
- Net effect: opening a new pane for an agent that previously had one running picks up the conversation automatically — no manual reattach required.

### Default-continue UX on the agent picker
`AgentPicker.handleSelect` now probes `agent:session:read` for each card on mount. The click contract is:

| State | Plain click | Shift / Ctrl / Alt / Cmd click |
|---|---|---|
| Agent has a `:current` session | **Auto-continue** (no modal, reuses most-recent NamedAgentRow's identity/memory binding) | Forces the launch modal — pick a different identity/memory/runtime |
| Agent has no session | Opens launch modal | Opens launch modal |

The auto-continue path doesn't set `continueOfInstanceId` — the agent zone is structurally continuous now, so the new pane just reads from `agent:<defId>:current` on mount.

### "+ New" affordance
Cards with a current session show a hover-revealed `+ New` button (bottom-right, where the install ribbon would otherwise sit). Click archives the current zone via `agent:session:archive`, then opens the launch modal so the user can start fresh without losing the prior conversation (it lives at `agent:<defId>:archive:<ts>` and will appear in the recent-sessions list).

### Pane title continuation chip
`viewText` renders `· continued Xm ago` when the snapshot's `modts` (returned by `agent:session:read`) indicates a >30 s gap from now — so the active pane's own writes never trigger the chip.

### RPC bindings
`AgentSessionRead` / `WriteState` / `AppendOutput` / `Archive` / `ListArchives` wrappers added to `frontend/app/store/rpc-api.ts` + `frontend/types/gotypes.d.ts`.

## Tests

- `useHistoryPagination.test.ts` — 3 tests covering the agent-zone read path, NDJSON fallback when content is null, and the `definitionId`-unset skip behavior.
- `AgentPicker.test.tsx` — 3 tests covering auto-continue when a session exists, modal open when none, and Shift forcing the modal regardless of session content.

All 6 new tests pass. Full agent + store suite: 736 passed / 2 pre-existing failures unrelated to this PR (`browser-pane-state-store`, `providers/index.test.ts` — both fail on a clean `main` checkout too).

`task build:frontend` succeeds (no tailwind regression).

## Out of scope

- **E3 cross-tab live sync.** Two panes of the same agent reading and writing the same zone concurrently — concurrent-write races, conflict resolution, and live event fan-out across panes is the next follow-up.
- **Old per-block-zone GC.** `<blockId>/output.state.json` files from before E1's migration stay in place; cleanup is a separate PR.

### DIFF
--- a/.changesets/1779585812-feat-agent-default-continue-from-agent-session-zones-fronten-5ota.md
+++ b/.changesets/1779585812-feat-agent-default-continue-from-agent-session-zones-fronten-5ota.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): default-continue from agent session zones (frontend)
--- a/frontend/app/store/rpc-api.ts
+++ b/frontend/app/store/rpc-api.ts
@@ -1156,6 +1156,34 @@ class RpcApiType {
         return client.rpcCall("blockfile:write_state", data, opts);
     }
 
+    // command "agent:session:read" [call] — Option E (PR #1007). Reads
+    // `output.state.json` from `agent:<definition_id>:current`.
+    AgentSessionReadCommand(client: RpcClient, data: CommandAgentSessionReadData, opts?: RpcOpts): Promise<AgentSessionReadResult> {
+        return client.rpcCall("agent:session:read", data, opts);
+    }
+
+    // command "agent:session:write_state" [call] — Option E. Writes
+    // `output.state.json` into `agent:<definition_id>:current`.
+    AgentSessionWriteStateCommand(client: RpcClient, data: CommandAgentSessionWriteStateData, opts?: RpcOpts): Promise<AgentSessionWriteStateResult> {
+        return client.rpcCall("agent:session:write_state", data, opts);
+    }
+
+    // command "agent:session:append_output" [call] — Option E.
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

PR #1013, opened 2026-05-24
feat(srv): db_agents table + backfill + dual-write (Phase 3a)

AgentA-asaf:
## Phase 3a — `db_agents` schema + dual-write

Phase 3a of the agent-concept consolidation. Tracking spec:
`docs/specs/SPEC_AGENT_CONCEPT_CONSOLIDATION_2026_05_24.md` (sections
"The lean model — 3 first-class concepts", "What `db_agent_instances`
rows become", "Implementation cost — honest accounting").

**ZERO behaviour change.** Every existing read path still hits
`db_agent_definitions` / `db_agent_instances`. The new `db_agents`
table is populated by a one-shot backfill + dual-writes from every
old-table mutation, but nothing in production code `SELECT`s from it
yet. Phase 3b migrates readers, Phase 3c drops the old tables.

### 1. New `db_agents` table

Lives in `objects.db` at schema v3 (bumped from v2). Adapted from the
spec's inline draft to match the **live** `db_agent_definitions`
shape — the spec sketched provider/cmd-template columns (`provider_id`,
`cmd`, `cmd_args`, `cmd_env`, `resume_flag`, `auth_config_dir_env`)
that never made it into storage, so projecting them now would mean
inventing data. Instead I kept the names we actually store
(`provider`, `provider_flags`, `shell`, `environment`, `agent_type`,
`agent_bus_id`, `accounts`, `auto_start`, `restart_on_crash`,
`idle_timeout_minutes`, `slug`, `branch_label`) and added an extra
`instance_name` column so the original named instance is recoverable
in the user-clone projection. Bindings + provenance + template-flag
columns match the spec verbatim.

```sql
CREATE TABLE db_agents (
    id, name, icon, description,
    is_template, parent_template_id,
    provider, provider_flags, shell, environment,
    agent_type, agent_bus_id, accounts,
    auto_start, restart_on_crash, idle_timeout_minutes,
    slug, branch_label,
    identity_id, memory_id, working_directory, github_context,
    instance_name,
    created_at, updated_at, is_seeded, user_hidden
);
CREATE INDEX idx_agents_is_template ON db_agents(is_template);
CREATE INDEX idx_agents_parent_template_id ON db_agents(parent_template_id);
CREATE INDEX idx_agents_is_seeded ON db_agents(is_seeded);
```

### 2. One-shot backfill

New file `agentmux-srv/src/backend/storage/agents_consolidate.rs`.
Marker-file gated (`<data_dir>/migration_agents_consolidate_v1.flag`).
Runs in `main.rs` after `auto_seed_on_startup`. Idempotent.
Transactional — partial failure rolls back; marker NOT written, so
next startup retries from scratch.

Algorithm (per spec §"What `db_agent_instances` rows become"):
- `is_seeded = 1` defs → `is_template = 1` projection.
- `is_seeded = 0` defs → `is_template = 0`, `parent_template_id = parent_id`.
- Instances of a template → new user-clone row keyed by `instance.id`,
  `parent_template_id = definition_id`, bindings from instance.
- Instances of a user-cloned def → fold bindings into the existing
  user-clone row (most-recent `created_at` wins on collision; warns
  on N > 1).
- Continuation rows (`parent_instance_id` non-empty) skipped — the
  consolidated model has no place for them.

Example backfill counts on the test round-trip fixture
(2 templates, 1 user-cloned def, 3 instances on a mix):
- 2 template projections inserted
- 1 user-clone-def projection inserted
- 2 instance-on-template projections inserted (one each)
- 1 instance-on-user-clone-def folded into the existing row
- 0 continuation rows skipped, 0 collisions

### 3. Dual-write helpers wired into wstore.rs

Failures LOG + CONTINUE — old tables remain authoritative in 3a.
Phase 3b will treat them as fatal once readers depend on db_agents.

Touched call sites (all funnel through `wstore.rs` methods — no
need to edit `agent_seed.rs` / `agent_handlers.rs` / `agent_session.rs`
because they all call these methods):
- `agent_def_insert` → `agents_dual_write_definition_upsert`
- `agent_def_update` → `agents_dual_write_definition_upsert`
- `agent_def_delete` → `agents_dual_write_definition_delete` +
  cascaded `agents_dual_write_instance_delete`
- `agent_def_delete_seeded` → `agents_dual_write_seeded_delete`
- `instance_create` → `agents_dual_write_instance_create`
- `instance_update` → `agents_dual_write_instance_update`
  (only writes `github_context` + `updated_at` — instance_update is
  block/session/status churn the consolidation deliberately drops)
- `instance_set_hidden` → `agents_dual_write_instance_set_hidden`
- `instance_repoint_definition` → `agents_dual_write_instance_repoint`
- `instance_delete` → `agents_dual_write_instance_delete`
- `instance_backfill_identity_id` → `agents_dual_write_backfill_identity`

### 4. Tests

- 6 backfill tests (round-trip lifecycle, collision warn, marker
  short-circuit, continuation skip, empty-db noop, hidden-flag
  preservation).
- 9 dual-write tests (one per write-side method + seeded bulk
  delete + repoint).
- Schema migration test (idempotent + table presence) extended for
  `db_agents` + indexes.
- Full `agentmux-srv` test suite: **1147 passed**, 1 ignored.

### 5. NO read-path changes

Verified with `grep` — no `SELECT … FROM db_agents` in production code
paths (only in the dual-write `DELETE` filters and in test helpers).
Frontend untouched.

## Follow-ups

- **Phase 3b**: migrate read paths (`instance_list_named`,
  `agent_def_list`, RPC handlers, frontend) to `db_agents`. Treat
  dual-write failures as fatal once readers depend on the new table.
- **Phase 3c**: drop `db_agent_definitions` + `db_agent_instances`,
  retire the dual-write helpers.

### DIFF
--- a/.changesets/1779600998-feat-srv-db-agents-schema-dual-write-a922.md
+++ b/.changesets/1779600998-feat-srv-db-agents-schema-dual-write-a922.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): db_agents schema + dual-write
--- a/agentmux-srv/src/backend/storage/agents_consolidate.rs
+++ b/agentmux-srv/src/backend/storage/agents_consolidate.rs
@@ -1,0 +1,711 @@
+// Copyright 2025-2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Phase 3a — one-shot backfill from `db_agent_definitions` +
+//! `db_agent_instances` into the new consolidated `db_agents` table.
+//!
+//! See `docs/specs/SPEC_AGENT_CONCEPT_CONSOLIDATION_2026_05_24.md`.
+//!
+//! Phase 3a is **write-only**: this migration populates `db_agents` so
+//! a later Phase 3b PR can flip readers over with full confidence the
+//! data is there. Phase 3c drops the old tables.
+//!
+//! Marker-file gated (`<data_dir>/migration_agents_consolidate_v1.flag`)
+//! so the backfill only runs once per data dir. Idempotent on second
+//! run (marker check short-circuits).
+//!
+//! Algorithm (per spec §"What `db_agent_instances` rows become"):
+//!
+//! 1. For each `db_agent_definitions WHERE is_seeded = 1`: INSERT a
+//!    template projection (`is_template = 1`, bindings empty).
+//!
+//! 2. For each `db_agent_definitions WHERE is_seeded = 0`: INSERT a
+//!    user-clone projection (`is_template = 0`, `parent_template_id =
+//!    parent_id`).
+//!
+//! 3. For each `db_agent_instances` row whose `definition_id` points at
+//!    a TEMPLATE: INSERT a new user-clone projection keyed by
+//!    `instance.id`, `parent_template_id = definition_id`, name +
+//!    bindings from the instance.
+//!
+//! 4. For each `db_agent_instances` row whose `definition_id` points at
+//!    an already-user-cloned definition: UPDATE the existing
+//!    user-clone projection (keyed by the definition id from pass 2) to
+//!    fold in the instance's bindings. If multiple instances point at
+//!    the same user-clone, the most-recent (`created_at` DESC) wins
+//!    and a warning is logged.
+//!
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1014, opened 2026-05-24
feat(srv): migrate read paths to db_agents (Phase 3b)

AgentA-asaf:
Phase 3b of the agent concept consolidation. Reader migration from `db_agent_definitions` / `db_agent_instances` to `db_agents` (consolidated table from Phase 3a #1013).

Companion to:
- #1013 (Phase 3a — schema + backfill + dual-write)
- Phase 3c (follow-up) — drops old tables, dual-write helpers, and the Phase 3a backfill.

Spec: `docs/specs/SPEC_AGENT_CONCEPT_CONSOLIDATION_2026_05_24.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1015, opened 2026-05-24
refactor(srv): drop db_agent_definitions + db_agent_instances + dual-write (Phase 3c)

AgentA-asaf:
Phase 3c of the agent concept consolidation
(`docs/specs/SPEC_AGENT_CONCEPT_CONSOLIDATION_2026_05_24.md`).

Sequence so far:
- Phase 3a (#1013) added the consolidated `db_agents` table with dual-write.
- Phase 3b (#1014) migrated reader paths off `db_agent_definitions` / `db_agent_instances`.
- **Phase 3c (this PR)** drops the old tables, the dual-write helpers, and the Phase 3a startup backfill — completing the consolidation.

## Schema migration v4 → v5

Bumped `OBJECT_SCHEMA_VERSION` from 4 to 5.

New columns added to `db_agents` so it can absorb the lifecycle the
retired `db_agent_instances` carried: `definition_id`,
`parent_instance_id`, `block_id`, `session_id`, `status`,
`started_at`, `ended_at`. Carry-forward `ALTER`s for pre-v5 dbs.

New `retire_old_agent_tables` migration step inside `run_object_schema`:
- Detects if `db_agent_definitions` / `db_agent_instances` exist (no-op on a v5+ db).
- Salvages residual rows into `db_agents` (best-effort `INSERT OR IGNORE` for new rows + a focused `UPDATE` for the user-clone-def fold case).
- DROPs both old tables (instances first to respect the legacy FK).
- Disables FK enforcement for the duration so child tables (`db_agent_content` / `db_agent_skills` / `db_agent_history` / `db_agent_identity_links` — re-pointed at `db_agents(id)`) survive the parent drop.

Idempotent on every run. Survives a v3 → v5 jump because the `db_agents` `CREATE TABLE IF NOT EXISTS` in the same function runs first.

## Code reductions

- `agentmux-srv/src/backend/storage/agents_consolidate.rs` deleted entirely (711 lines).
- `wstore.rs` lost ~570 lines of dual-write helpers (`agents_dual_write_*`, `agents_projection_key_for_inst`, `load_definition_for_dual_write`) + their `dual_write_*` test block.
- `main.rs` lost the `run_agents_consolidate` startup hook — the schema migration runs the equivalent backfill under the connection lock instead.

**Net: −1350 lines** across the storage layer.

## Writers migrated to direct `db_agents`

All Rust method signatures unchanged so callers don't need to touch.

`agent_def_insert`, `agent_def_update`, `agent_def_delete`,
`agent_def_delete_seeded`, `agent_def_set_hidden`,
`instance_create`, `instance_update`, `instance_set_hidden`,
`instance_repoint_definition`, `instance_delete`,
`instance_backfill_identity_id`.

Semantic preservation highlights:
- `instance_create` keeps the fold rule: parent is a user-clone DEF ⇒ UPDATE the def row in place; parent is a template ⇒ fresh template-instance projection row.
- `instance_delete` is a NO-OP on a fold row (the def persists); the cross-version registry mirror still cleans up the instance file either way.
- `agent_def_delete_seeded` cleans cascaded template-instance projection rows now that the old-schema FK cascade is gone.
- `agent_def_delete` cascades through both explicit template-instance projection rows AND any registry record whose `definition_id` points at the deleted agent (the folded-instance case where there's no db_agents row to cascade through).

## Readers migrated

The remaining instance readers (`instance_list`, `instance_get`,
`instance_list_named`, `instance_get_by_name`,
`instance_get_active_for_block`) now query
`db_agents WHERE is_template = 0 AND definition_id <> ''`.
`map_instance_row` projects a `db_agents` row back into the `AgentInstance` shape.

## Vestigial types kept

`AgentInstance` remains as a Rust struct. ~65 call sites across 8 files reference it; per the spec's escape hatch we keep it as a query helper over `db_agents` rather than blowing up the PR scope renaming everything. Tidy-up follow-up to drop the `instance_*` prefix on `WaveStore` methods + retire the struct entirely is left for a separate cycle.

Frontend bindings (`frontend/types/gotypes.d.ts` `AgentInstance`) are unchanged because the Rust struct is.

## Test results

- `cargo test -p agentmux-srv --bin agentmux-srv` — **1139 passed, 1 ignored**.
- `task build:frontend` — green.
- New migration test `test_v5_retires_old_agent_tables_and_carries_data_to_db_agents` covers the v4 → v5 transition with seeded templates + user-clone defs + their instances; verifies the fold + drop both run cleanly and re-running is idempotent.
- Three identity-migration tests had their parent agent flipped from `is_seeded = 0` to `is_seeded = 1` so the test-fixture `instance_get("inst-…")` reads still find a row (a fold would have routed the instance into the def and returned None on instance-id reads).

## Marker file

The Phase 3a marker (`migration_agents_consolidate_v1.flag` in the data dir) is left in place on existing user data — harmless residue.


a5af:
@codex review

AgentA-asaf:
Closing — Phase 3c hit a structural impedance mismatch with the fold model carried forward from Phase 3a.

**Where we landed**: Phase 3a (#1013) shipped \`db_agents\` + dual-write; Phase 3b (#1014) shipped reader migration. Main is at a functional steady-state: old tables present, new table consolidated, dual-written, reads from new. The schema cleanup hasn't completed but the system works.

**Why Phase 3c didn't converge**: the fold model (one \`db_agents\` row per user-clone def, instance bindings absorbed) requires a runtime \`inst.id → def.id\` resolution path for \`instance_get/update/delete\` to find folded rows. Pre-3c that mapping lives in \`db_agent_instances\`. Phase 3c drops that table — breaking the mapping. After 2 review rounds (round 1 the schema-index P0; round 2 three P1s including the fold-resolution gap, plus a child-table-FK rebuild gap, plus an over-broad cascade selector), the cleanest fix is to eliminate the fold entirely — give every instance its own row in \`db_agents\`. That's a bigger pivot than this PR's scope.

**Deferred**: a Phase 3c.v2 that does the no-fold pivot (\`instance_create\` always INSERTs; backfill / retire path matches; reader handles the doubled-row case). Probably ~3-4 PRs (instance-create no-fold; backfill no-fold; reader dedup; finally drop old tables + dual-write).

Phase 1 + Phase 2 (the user-facing two-tier picker work) are merged and stable. Phase 3a + 3b (the structural foundation) are merged and stable. Phase 3c is deferred until a no-fold redesign cycle.

PR #1016, opened 2026-05-24
fix(srv): show continuation-row agents in My Agents picker (Option E)

AgentA-asaf:
**TL;DR**: clicking an agent under My Agents previously did not surface any user agent whose latest instance was a continuation (`parent_instance_id != ''`). Dropping that filter from `instance_list_named` restores them.

## Bug

Reproduced today on a freshly-built portable. The session zone migration (`template_promote_v1`) had run and created the user-clone def, but the picker still showed "No agents yet" — and `listrecentsessions` returned `[]`. Tracing through:

- `MyAgentsList` (frontend) → `RpcApi.ListRecentSessionsCommand`
- → `listrecentsessions` (handler) → `wstore.instance_list_named(raw_limit, None)`
- → SQL `WHERE display_hidden = 0 AND instance_name <> '' AND parent_instance_id = ''`

The user's instance (`Maks`) had `parent_instance_id` set from an earlier resume in this session. The filter excluded it; the picker saw zero rows.

## Why the filter existed

Pre-Option-E the session zone was anchored on `block_id`, and `parent_instance_id` linked one block's instance to the next. Filtering on `parent_instance_id = ''` was a proxy for "show only the head of chain" so the dropdown didn't surface every intermediate resume as its own row.

## Why it's wrong under Option E

Option E (#1007/#1008) anchored the session zone on `definition_id`. A continuation row is no longer a separate session — it's just the most-recent named instance of an agent the user actively uses. The picker WANTS that row visible.

Same root cause also affected `template_promote`'s instance-name lookup (`agent_session.rs:783`): it called `instance_list_named(1, Some(&old_def_id))` to pick up the user's chosen name and fell back to `template.name` ("Claude") because the query returned zero rows. Dropping the filter fixes both call sites.

## Registry mirror — deliberately left alone for this PR

`registry_upsert_if_named` (wstore.rs:2154) has a matching filter for the cross-version registry mirror. Dropping it there would risk surfacing chained resumes as multiple registry entries (the registry-sourced read path doesn't have the SQLite ORDER BY/LIMIT dedup affordance). Comment updated to flag the deliberate divergence; cross-version dedup will land in a follow-up PR.

## Test

New `instance_list_named_includes_continuation_rows` covers:
- Both head-of-chain AND continuation rows are returned
- Sorted most-recent-first (`started_at DESC`)
- `parent_instance_id` preserved in the returned struct
- Definition-scoped variant works too

Existing `instance_create_continuation_row_does_not_mirror` (now with comment acknowledging the deliberate divergence) still validates the registry-side gating.

## Verification

- 107/107 wstore + agent_handlers + agents_consolidate tests pass.
- Live verification on portable v0.38.4: after applying this fix locally via DB tweak (`UPDATE db_agent_instances SET parent_instance_id = '' WHERE id = '<maks-id>'`), `listrecentsessions` returned Maks (5.21ms vs. 0.3ms for empty results), and the picker UI rendered him under My Agents.

## Follow-ups (separate PRs)

- B3 (template_promote one-shot marker — should be data-invariant-gated, not marker-gated). The migration ran the first time before the seeded def grew a session zone; the marker prevented it from re-firing once Maks's zone DID exist.
- B4 (reattach launch doesn't pass `--continue`/`--resume <sid>` to Claude CLI). After this PR you can SEE Maks in the picker; clicking him still spawns a fresh session because `cliArgs` finalized in `agent-model.ts` never picks up the prior block's `agent:sessionid` meta.

### DIFF
--- a/.changesets/1779612216-fix-srv-drop-parent-instance-id-filter-from-instance-list-na-w3mr.md
+++ b/.changesets/1779612216-fix-srv-drop-parent-instance-id-filter-from-instance-list-na-w3mr.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): drop parent_instance_id filter from instance_list_named so continuation rows appear in My Agents picker (Option E)
--- a/agentmux-srv/src/backend/agent_session.rs
+++ b/agentmux-srv/src/backend/agent_session.rs
@@ -780,7 +780,16 @@ pub fn migrate_promote_template_sessions_v1(
         // `instance_list_named` already filters to non-hidden + named
         // rows + sorts by `started_at DESC`, so the first row is the
         // pick.
-        let new_name = match wstore.instance_list_named(1, Some(&old_def_id)) {
+        // Include continuations: a user who clicked Maks today and
+        // resumed three times has only continuation rows for that
+        // definition; the head row is whatever they originally
+        // named the agent. Picking the most-recent continuation
+        // surfaces the same `instance_name` they used last.
+        let new_name = match wstore.instance_list_named(
+            1,
+            Some(&old_def_id),
+            /* include_continuations */ true,
+        ) {
             Ok(rows) => rows
                 .into_iter()
                 .next()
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1017, opened 2026-05-24
fix(srv): make template_promote migration self-idempotent (data-invariant-gated)

AgentA-asaf:
**TL;DR**: the Phase 1 picker migration was one-shot/marker-gated, which created an "early-marker" trap — the migration could fire BEFORE any seeded def grew a session zone, write its marker, then permanently skip thereafter even when the invariant later became violated. Dropping the marker gate and re-running every startup fixes it; the migration is already O(1) when there's nothing to do.

## The bug

Reproduced today on portable v0.38.4:

1. Earlier portable run: no seeded defs had session zones. `migrate_promote_template_sessions_v1` ran, found nothing (`templates_scanned: 0`), wrote `migration_template_promote_v1.flag`.
2. User clicked the "Claude Code" template card. Real conversation. Session zone now at `agent:claude:current` on a seeded def — **invariant violated**.
3. Subsequent portable startups: marker present → migration skipped. No user-clone def, picker shows "No agents yet" even though Maks's conversation is sitting in the filestore.

Confirmed by deleting the marker manually and relaunching — migration ran on second pass, promoted Claude into a user-clone def, moved the zone, repointed the instance.

## Fix

Drop the marker check at the top of the migration; drop the marker write at the bottom. Run on every startup.

This is safe because the migration is **O(1) when there's nothing to do**:
- `filestore.get_all_zone_ids()` — one SQLite query
- Parse zones, group by def-id — O(N_zones), typically a handful
- `wstore.agent_def_list()` — one SQLite query
- For each (def_id, zones), match against seeded defs → typically zero iterations

When the invariant already holds, the function returns `TemplatePromoteStats::default()` in sub-ms.

## Backward compat

`TEMPLATE_PROMOTE_MARKER_V1` constant is retained — existing marker files in real-world data dirs are ignored, not resurrected. The `data_dir` function parameter is retained (renamed to `_data_dir`) so callers in `main.rs` don't need to change.

## Tests

- `template_promote_runs_when_seeded_def_grows_zone_after_first_run` — new regression test; walks the reproducer step-by-step.
- `template_promote_ignores_legacy_marker_file` — new test; legacy marker file pre-existing must not prevent re-runs.
- `template_promote_is_idempotent_on_second_run` — existing; still passes.
- `template_promote_clones_template_and_moves_zones` — existing; updated to assert marker NOT written.

All 6 template_promote tests + the broader 24 agent_session module tests pass.

## Stack context

Part of a 3-PR series after the 2026-05-24 "Maks not under My Agents" investigation:

- #1016 — drop `parent_instance_id = ''` filter from `instance_list_named` (picker visibility) — independent.
- **THIS** — make `template_promote` self-idempotent — independent.
- (next) — wire reattach to pass `--resume <session_id>` so clicking My Agents resumes the actual conversation instead of re-injecting startup context.

### DIFF
--- a/.changesets/1779612433-fix-srv-make-template-promote-migration-self-idempotent-data-tnjc.md
+++ b/.changesets/1779612433-fix-srv-make-template-promote-migration-self-idempotent-data-tnjc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): make template_promote migration self-idempotent (data-invariant-gated, not marker-gated)
--- a/agentmux-srv/src/backend/agent_session.rs
+++ b/agentmux-srv/src/backend/agent_session.rs
@@ -998,34 +998,58 @@ fn move_zone(
     if files.is_empty() {
         return Ok(());
     }
-    // Per-file copy gate (codex P1 rounds 4 + 5 on PR #1017).
-    // The two failure modes pull in opposite directions:
+    // Per-file recency-aware copy (codex P1 rounds 4 + 5 + 6 on
+    // PR #1017). Three retry shapes need to coexist on the same
+    // retry path:
     //
-    //   R4: partial-failure retry, user has continued on the
-    //       destination → must NOT overwrite the destination's
-    //       newer content with stale seeded source bytes.
-    //   R5: partial-failure retry, the prior `move_zone` wrote
-    //       some destination files but not others → must NOT
-    //       drop the source until every source file is also
-    //       present at the destination (else the unwritten ones
-    //       are lost forever).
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #1018, opened 2026-05-24
fix(agent): reattach plumbs prior session id so --resume fires on the first turn

AgentA-asaf:
**TL;DR**: clicking an existing agent under "My Agents" spawned a fresh CLI session and re-injected the startup context, even though `agent:resume_flag` was set in block meta. Root cause: the resume flag is conditionally added in `spawn_turn` only when `inner.session_id` is populated — and a freshly-spawned `SubprocessController` for the reattached block has `None` there until AFTER the first turn's stdout is consumed. This PR threads the prior block's CLI-emitted session id through the launch flow end to end, so the first turn already has it.

## Observed bug

Today on portable v0.38.4, after #1016 + #1017 made Maks visible under My Agents, clicking him spawned `claude` with these args:

```
-p --output-format stream-json --verbose --include-partial-messages
--dangerously-skip-permissions --model sonnet --effort medium
```

No `--resume`. Claude logged in its thinking pane: *"The user hasn't sent a message yet — this appears to be the start of a session."*

## Why

`subprocess.rs::spawn_turn` lines 258–268 (pre-fix):

```rust
let mut args = config.cli_args.clone();
{
    let inner = self.inner.lock().unwrap();
    if let Some(ref sid) = inner.session_id {
        if !config.resume_flag.is_empty() {
            args.push(config.resume_flag.clone());
            args.push(sid.clone());
        }
    }
}
```

`inner.session_id` is populated from CLI stdout at line 496. On a brand-new controller for a reattached block, it's `None` on turn 1 → no `--resume <sid>` appended → fresh session.

## Fix

Thread the session id end to end:

**Backend**
- `RecentSessionRow` adds `session_id: String` (from `db_agent_instances.session_id` which the spawn path already persists).
- `SubprocessSpawnConfig` adds `session_id: Option<String>`.
- New `SubprocessController::hydrate_session_id_from_config(Option<&str>)` — copies value to `inner.session_id` IFF currently `None`. **Captured-id-wins**: a CLI-emitted id captured on a prior turn always wins over a (potentially stale) config value, so the CLI's view of its own session is authoritative once it speaks.
- `spawn_turn` calls the helper before building args, so the existing resume-flag block fires on the first turn.
- `websocket.rs` + `app_api.rs` (both `SubprocessSpawnConfig` builders) read `agent:sessionid` from block meta and populate the new field.

**Frontend**
- `LaunchOverrides` adds `continueSessionId?: string`.
- `AgentPicker.handleReattach` passes `row.session_id` from the picker row.
- `agent-model.ts::launchAgentDefinition` writes it to `agent:sessionid` meta when non-empty. (Greenfield launches omit the key so a later captured value isn't shadowed.)
- `gotypes.d.ts` synced.

## Tests

Four new unit tests on the helper:

- `hydrate_session_id_populates_inner_when_none` — fresh controller, hydration writes through.
- `hydrate_session_id_keeps_captured_over_config_value` — captured-id-wins invariant.
- `hydrate_session_id_ignores_empty_and_none` — greenfield path.
- `spawn_turn_preserves_session_id_in_queued_config` — busy-lock queue path forwards the value to the drain.

Plus the original `test_subprocess_controller_concurrent_spawn_blocked` updated to set the new field.

**1164 backend tests pass.** Frontend pre-existing failures (BlockErrorBoundary + AgentPicker mock-fixture) are unchanged on this branch.

## Stack context

Closes the 3-PR series from the 2026-05-24 "Maks not under My Agents" investigation:

- #1016 — drop `parent_instance_id = ''` filter from `instance_list_named` (picker visibility)
- #1017 — make `template_promote` migration self-idempotent (no marker gate)
- **THIS** — reattach plumbs session id forward (resume on first turn)

After this PR ships, clicking Maks under My Agents spawns `claude -p … --resume <prior-sid>`, the CLI resumes the actual conversation, and the startup context is NOT re-injected.

### DIFF
--- a/.changesets/1779613227-fix-agent-pass-prior-session-id-through-reattach-so-spawn-in-fv0o.md
+++ b/.changesets/1779613227-fix-agent-pass-prior-session-id-through-reattach-so-spawn-in-fv0o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): pass prior session id through reattach so spawn includes --resume on the first turn
--- a/agentmux-srv/src/backend/blockcontroller/subprocess.rs
+++ b/agentmux-srv/src/backend/blockcontroller/subprocess.rs
@@ -65,6 +65,22 @@ pub struct SubprocessSpawnConfig {
     /// queued → running so the frontend can pair the event with a
     /// pending `PendingMessage`. None means no feedback is emitted.
     pub message_id: Option<String>,
+    /// Session id to hydrate `inner.session_id` with BEFORE the first
+    /// turn — used by the picker "My Agents" reattach path. The
+    /// caller reads this from `block.meta["agent:sessionid"]` which
+    /// the frontend pre-populates from the prior block's session id
+    /// when launching with `continueOfInstanceId`. Without this
+    /// hydration `spawn_turn` would only see the captured session id
+    /// AFTER the first turn — meaning the first turn always launches
+    /// the CLI fresh (no `--resume <sid>`) and starts a new
+    /// conversation that re-injects the startup context.
+    ///
+    /// Empty / `None` means "no prior session" (greenfield launch).
+    /// If `inner.session_id` is already set on the controller (e.g. a
+    /// subsequent turn within the same process lifetime captured it
+    /// from CLI stdout), this field is ignored — the in-memory
+    /// captured id always wins.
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #944, opened 2026-05-20
feat(stability): suppress Windows crash modal — supervision Phase 0

AgentA-asaf:
## Phase 0 — suppress the Windows crash modal

First slice of the **service supervision & recovery framework** (Discussion
#943, issue #942).

### Problem

When the host hits an unhandled fault — a Chromium `LOG(FATAL)`, an `abort()`,
a breakpoint — Windows shows an "Application Error" modal. The 2026-05-20 GPU
crash is the case study: the `0x80000003` *"a breakpoint has been reached… OK
to terminate, CANCEL to debug"* dialog. That modal **freezes the process**
behind an OK/Cancel prompt — it cannot be auto-recovered while the dialog is
up, and the user sees a raw OS error instead of a graceful restart.

### Change

`SetErrorMode(SEM_FAILCRITICALERRORS | SEM_NOGPFAULTERRORBOX)`, called as the
**first statement of `main()`** in both the host (`agentmux-cef`) and the
launcher (`agentmux-launcher`). A fault now terminates the process immediately
and silently; the launcher's supervisor (Phase 1) relaunches it.

- `agentmux-cef/src/main.rs`, `agentmux-launcher/src/main.rs` — a small
  `suppress_os_crash_dialogs()` (cfg-gated; no-op off Windows), called first.
- Both `Cargo.toml`s gain the `Win32_System_Diagnostics_Debug` windows-sys
  feature (where `SetErrorMode` lives).

Smallest, self-contained slice — removes the scariest symptom, ships
independently of the rest of the framework.

### Verification

- `cargo check -p agentmux-launcher` and `cargo check -p agentmux-cef` — both
  clean (pre-existing warnings only).
- Runtime verification (a fault actually terminating without the modal)
  requires a crash repro and is **not automated here** — the fault-injection
  harness is Phase 1 (spec §14).

### Note on crash dumps

`SEM_NOGPFAULTERRORBOX` suppresses the WER **dialog**. Today the breakpoint
modal *already* blocks WER dump collection (the process is frozen at the
prompt), so this change is **no worse** for dumps and likely better — a
silent terminate lets WER LocalDumps proceed. Flagging for confirmation; if
LocalDumps still doesn't collect, Phase 1 adds a `SetUnhandledExceptionFilter`
that writes a minidump before exit.

### Docs riding along

- `docs/specs/SPEC_SERVICE_SUPERVISION_AND_RECOVERY_2026_05_20.md` — the full
  framework spec (Draft).
[...]

PR #1019, opened 2026-05-24
fix(agent): failed tool panels collapse after the 5s post-completion hold like successes

AgentA-asaf:
**TL;DR**: failed Bash / Read / Grep panels were staying open forever. They should fold to the single-line ✗ row after the same 5s post-completion hold that successes get. Three sites were force-expanding failures; this PR fixes all three.

## User report

> "i notice failed commands log doesn't collapse after a couple seconds, they stay open"

## Why failures stay open

Three forced-expand sites:

1. **`stream-parser.ts:265`** — `collapsed: event.status === "success"`. Failures arrived with `collapsed: false`.
2. **`ToolBlock.tsx`'s `isActive` predicate** treated `failed` as a continuing active state, so failures never crossed the active→inactive transition that triggers the 5s post-completion hold + collapse.
3. **`ToolBlock.tsx`'s `autoExpanded`** returned true while `status === "failed"`. Even if the hold had elapsed, the panel would still be forced open.

## Fix

Collapse on EVERY terminal state (success OR failure). The existing 5s hold (`POST_COMPLETION_HOLD_MS`, bumped from 1s to 5s in #944) gives time to read the last output line before the panel folds. Per `feedback_no_timers_or_delays`, no new setTimeout — reuses the deterministic active→inactive transition already in place.

## Discoverability of failures preserved

Per `docs/specs/tool-collapse.md`:

- Collapsed row: `✗ <ToolName>` with red `border-left: 2px solid var(--error-color)`.
- Hover-to-peek still expands collapsed failures (instant).
- Pin (click) still keeps any tool expanded indefinitely.

## Tests

- All 504 frontend stream-parser tests pass.
- No prior test asserted `failed-stays-open` behavior (searched: only finds the auth-state `status: "failed"` references, unrelated).

### DIFF
--- a/.changesets/1779631619-fix-agent-failed-tool-panels-collapse-after-the-5s-post-comp-gisc.md
+++ b/.changesets/1779631619-fix-agent-failed-tool-panels-collapse-after-the-5s-post-comp-gisc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): failed tool panels collapse after the 5s post-completion hold like successes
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1023, opened 2026-05-24
fix(test): unstale browser-pane title + ACP openclaw OAuth assertions (2 failing → 0)

AgentA-asaf:
PR B of three from `docs/specs/SPEC_FRONTEND_TEST_HEALTH_2026_05_24.md`. Follows up #1022 which dropped the worktree-noise from 84 → 2 failures; this PR drops those 2 → 0.

### 1. `browser-pane-state-store.test.ts:66`

Test asserted `proj.calls.title === []` after a `Navigate` dispatch. But `SPEC_BROWSER_PANE_OPTIMISTIC_HEADER_2026_05_18` added a hostname-derived title placeholder so the address bar header doesn't show the previous page's title during the 500ms-3s before CEF emits `TitleChanged`. The reducer's `Navigate` case (`reducer.ts:83-94`) sets `title: deriveTitlePlaceholder(url)` — for `https://example.com` that's `"example.com"`.

Production behavior is intentional and correct; the test's assumption that title wouldn't change on Navigate was a spec behind. Updated to `expect(proj.calls.title).toEqual(["example.com"])` with a spec-pointing comment.

### 2. `providers/index.test.ts:65`

Test iterated `ACP_IDS = ["openclaw", "pi"]` asserting each had `authType === "api-key"`. But `SPEC_OPENCLAW_AGENT_2026_05_17.md §4` shipped `openclaw` with `authType: "oauth"` — OAuth-via-subcommand (`models auth login --provider openai-codex`) to borrow OpenAI Codex's credentials. The blanket "all ACP = api-key" assertion was already false on main.

Split into three targeted assertions:
- ACP providers use the `acp` output format (uniform — both still in one loop).
- ACP+OAuth providers (`openclaw`) use OAuth auth.
- ACP+api-key providers (`pi`) use api-key auth.

Each list is small and explicit so a future provider's auth choice fails one targeted assertion instead of a broad blanket one.

## Result

```
before  Tests 2 failed | 1019 passed
after   Tests 0 failed | 1021 passed
```

## Out of scope (PR C)

The `tsc --noEmit` errors in test files (`toBeInTheDocument` / `toHaveTextContent` not on `Assertion` type — 35+ lines across 5 files) — tests pass at runtime, only the typecheck is dirty. PR C will add `@testing-library/jest-dom` to `tsconfig.json`'s `types` array. Spec §4.

### DIFF
--- a/.changesets/1779666314-fix-test-unstale-browser-pane-title-and-acp-openclaw-oauth-a-p9vd.md
+++ b/.changesets/1779666314-fix-test-unstale-browser-pane-title-and-acp-openclaw-oauth-a-p9vd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(test): unstale browser-pane title and ACP openclaw OAuth assertions
--- a/frontend/app/store/browser-pane-state-store.test.ts
+++ b/frontend/app/store/browser-pane-state-store.test.ts
@@ -61,9 +61,14 @@ describe("browser-pane-state-store (slice #9 Phase 4)", () => {
         expect(proj.calls.faviconUrl).toEqual([
             "https://example.com/favicon.ico",
         ]);
-        // Cells that didn't change shouldn't be projected.
+        // Navigate also seeds the title with a hostname-based
+        // placeholder so the header doesn't show the previous
+        // page's title during the load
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1021, opened 2026-05-24
fix(agent): hover-anchored startup body (overlay above/below) — summary stays under cursor

AgentA-asaf:
Implements `docs/specs/SPEC_STARTUP_HOVER_EXPANSION_ANCHOR_2026_05_24.md`. Companion progress log: `docs/sessions/PROGRESS_HOVER_ANCHOR_2026_05_24.md`.

## What

The hover-expansion of `⓵ Session context` was perturbing document layout (32px → ~200-400px), which combined with the user's natural mouse drift during the 150ms enter-delay produced an "expand → mouse drifts off → collapse" loop. PR #1020's spec said body grows inline below summary; in practice it didn't work for users near the bottom of the viewport or who hovered near the summary's bottom edge.

This PR makes the hover-expanded body float as an absolute-positioned overlay anchored to the summary, on the side with more viewport room. The summary's screen-Y is unchanged across the expand. Pinned content drops into normal flow (a deliberate persistence commitment).

## Research round (best practices)

- [Floating UI flip middleware](https://floating-ui.com/docs/computeposition) — the canonical "pick the direction with more room" pattern.
- [Floating UI useHover + safePolygon](https://floating-ui.com/docs/usehover) — for trigger→popover gaps. Not needed here (see below).
- [MDN mouseleave](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event) — confirms `mouseleave` fires when pointer exits element AND ALL DESCENDANTS. "Descendants" = DOM tree, NOT visual containment.
- [TanStack Virtual](https://tanstack.com/virtual/latest/docs/api/virtualizer) — absolute-positioned children don't change the row's measured height; virtualizer doesn't see the overlay.

The MDN insight collapses the design: because the overlay body is a DOM child of `.agent-user-message`, cursor moves between summary and absolute body do NOT fire mouseleave on the outer block. No `safePolygon`, no hover-intent timeout, no grace window. Per `feedback_no_timers_or_delays`.

## How (three independently revertible commits)

| Commit | Files | Effect |
|---|---|---|
| 1 | `hover-anchor.ts` + `.test.ts` | Pure function `pickExpandDirection(rect, viewportH, bodyEstimate) → "above" \| "below"`. 10 L1 tests. |
[...]

PR #1020, opened 2026-05-24
fix(agent): high-contrast user-input + startup-injection collapse-on-hover

AgentA-asaf:
Implements `docs/specs/SPEC_USER_INPUT_VISIBILITY_AND_STARTUP_COLLAPSE_2026_05_24.md` (spec landed alongside the implementation per `feedback_no_doc_only_prs`).

## What

1. **High-contrast user-input color.** New `--user-input-color: #00b4d8` theme variable (separate from `--accent-color`), used by `.agent-user-message`: 14% tint background + 3px left border. Typed input now visibly pops out of the agent reply stream.
2. **Regular input never soft-wraps.** `<pre>` content switches from `white-space: pre-wrap; overflow-wrap: anywhere;` to `white-space: pre; overflow-x: auto;` — newlines honored verbatim, long lines (URLs, paths) scroll horizontally inside the bubble.
3. **Startup injection collapses like tools.** The Markdown payload from `buildStartupPayload` (sent by `agent-view.tsx onReadyFn`) renders as a single-line summary `⓵ Session context (hover to peek · click to pin)`. Hover-enter delay 150ms (matches ToolBlock), click toggles pin via the existing `documentState.pinnedNodes` infrastructure.

## How

Three commits, each independently revertible:

| Commit | Files | Effect |
|---|---|---|
| `cc7180f4` | `theme.scss`, `_document-nodes.scss` | High-contrast color + no-wrap rule + startup-variant SCSS scaffold |
| `ae4fc980` | `types.ts`, `stream-parser.ts`, `stream-parser.test.ts` | `isStartup` flag set when message matches `/^# Session Context\b/`; 5 L1 tests + a contract test against `buildStartupPayload` |
| `19a59ac1` | `UserMessageBlock.tsx` (+test), `DocumentRow.tsx`, spec | Component swap, 10 L2 tests |

## Why heuristic and not a wire-level marker

The startup payload is a frontend-only construct — backend has no opinion on what's "startup", and the wire-level event has no metadata field for it. Round-tripping a flag through the agent CLI's NDJSON would require backend changes for a frontend-only concern. The literal heading `# Session Context` is owned by `buildStartupPayload.ts`; a contract test in `stream-parser.test.ts` calls the real builder and asserts the regex matches, so any future rename of the heading fails the test atomically with the source change. Discussed in spec §C.1.

## Tests

- 10 new L2 tests on `UserMessageBlock` (render shapes, hover-enter delay, mouseleave-cancel, click-to-pin, pin-class).
[...]

PR #1025, opened 2026-05-25
fix(agent): opaque overlay bg + extend hover-anchor to ToolBlock (2 user-reported issues)

AgentA-asaf:
Two follow-ups to PR #1020 (high-contrast user input + startup collapse) and #1021 (hover-anchored startup body), both raised after user testing:

## 1. Session-context overlay was distorted / bleeding through

**Symptom.** When the startup session-context line expanded on hover, the overlay panel's background was semi-transparent — content from the rows behind showed through and made the overlay text hard to read.

**Root cause.** The overlay used `color-mix(in srgb, var(--user-input-color) 18%, …)` for the background. In theory that's opaque (the second argument is a solid color), but in practice CEF's `color-mix` implementation produced visible bleed-through — likely a percentage short-circuit in the renderer.

**Fix.** Drop `color-mix` for the overlay background; use a direct opaque variable instead:

```scss
background-color: var(--block-bg-solid-color, #1c1c1c);
border: 2px solid var(--user-input-color);  // bumped from 1px to match weight of expanded chrome
```

Drop-shadow preserved for depth cueing. The `:has()` row-elevation rule in `_document.scss` keeps doing its job of escaping `contain: layout style` stacking.

## 2. Tool calls at the bottom of the pane don't expand upward

**Symptom.** A tool block at the very bottom of the agent pane, hovered, showed no expansion — the panel was trying to grow downward into clipped space and either flashed and collapsed (hover-out as cursor entered clipped territory) or didn't render visibly at all.

**Root cause.** ToolBlock's hover-to-expand always laid the panel out below the summary row in normal flow. There was no flip logic — the design from `SPEC_STARTUP_HOVER_EXPANSION_ANCHOR_2026_05_24.md` §5.10 generalizes to ToolBlock, but it wasn't wired up.

**Fix.** ToolBlock now mirrors UserMessageBlock's three-mode panel rendering:

| Mode | When | Layout |
|---|---|---|
| `hidden` | not expanded | not rendered (display:none) |
| `overlay` | hover-only (not pinned, not auto-expanded) | absolute-positioned, direction picked by `pickExpandDirection`, inline `max-height` from `maxOverlayHeight`. Activates the panel's `overflow-y` inside pane bounds. |
| `flow` | pinned OR auto-expanded by status (running, pending_approval, post-completion-hold) | in-flow as before — running tools don't reposition while streaming |

The hover-anchor helpers in `hover-anchor.ts` (`pickExpandDirection`, `findScrollContainerRect`, `maxOverlayHeight`) are reused verbatim — no new abstractions.

SCSS adds `.agent-tool-panel--overlay-below` / `--overlay-above` modifiers with the same opaque-bg + bold-border + shadow as the session-context overlay; `--flow` modifier added for explicit symmetry. `:has()` row-elevation in `_document.scss` extended to cover the tool overlay variants so they escape `contain: layout style` stacking the same way.

## Tests

New `ToolBlock.test.tsx` (7 tests):
- collapsed (default success) → panel has `--hidden`
- pinned → panel is in-flow (`--flow`)
- running → auto-expanded in-flow
- pending_approval → auto-expanded in-flow
- hover only (post-completion, not pinned, not running) → overlay
- hover-mode overlay has inline `max-height`
- pinned panel has no inline `max-height` (in-flow uses SCSS default)

Existing UserMessageBlock + hover-anchor tests unchanged. All 50 tests across these three files pass.

## Specs

- `docs/specs/SPEC_STARTUP_HOVER_EXPANSION_ANCHOR_2026_05_24.md` (§5.10 — generalization to ToolBlock)
- `frontend/app/view/agent/components/hover-anchor.ts` (pure helpers — unchanged, just consumed by a new caller)

### DIFF
--- a/.changesets/1779664479-fix-agent-opaque-overlay-bg-extend-hover-anchor-to-toolblock-uttt.md
+++ b/.changesets/1779664479-fix-agent-opaque-overlay-bg-extend-hover-anchor-to-toolblock-uttt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): opaque overlay bg + extend hover-anchor to ToolBlock for upward expansion near pane bottom
--- a/frontend/app/view/agent/components/ToolBlock.test.tsx
+++ b/frontend/app/view/agent/components/ToolBlock.test.tsx
@@ -1,0 +1,142 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * ToolBlock — panel-mode render tests.
+ *
+ * Covers the hover-anchor extension landed alongside
+ * `SPEC_STARTUP_HOVER_EXPANSION_ANCHOR_2026_05_24.md` §5.10:
+ *   - hover-only state → overlay positioning.
+ *   - pinned / running / pending_approval → in-flow positioning.
+ *   - hidden state → `--hidden` class.
+ *
+ * Pre-existing behaviors (status-icon mapping, live-tail, etc.)
+ * are exercised by the integration tests in `renderers.test.ts`
+ * and the broader agent-pane fixtures.
+ */
+
+import { cleanup, fireEvent, render } from "@solidjs/testing-library";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { ToolBlock } from "./ToolBlock";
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1027, opened 2026-05-25
feat(common): channels — version-spanning data isolation (Increment A from SPEC_DATA_CHANNELS)

AgentA-asaf:
**Increment A of 3** from [`SPEC_DATA_CHANNELS_2026_05_24.md`](../blob/agenta/channels-increment-a/docs/specs/SPEC_DATA_CHANNELS_2026_05_24.md). Tracking discussion: [#1026](https://github.com/agentmuxai/agentmux/discussions/1026).

## Problem this fixes

Today, every `task package` bump produces a portable with an empty My Agents list — data is keyed on build version (`~/.agentmux/versions/<v>/`). The original [`SPEC_DATA_DIR_UNIFICATION_2026-05-05`](../blob/main/docs/specs/SPEC_DATA_DIR_UNIFICATION_2026-05-05.md) Phase 1 chose this trade-off deliberately for schema safety, and explicitly flagged "version-comparison testing requires re-doing every login" (§2.6) as a known cost. This PR is Phase 2 — channels.

## What this PR does

Introduces a *channel* — a stable identifier that spans versions within the same compat band — and replaces per-version dirs with per-channel dirs for Installed + Portable. Dev mode unchanged (still per-branch under `dev/<branch>/`).

```
~/.agentmux/
├── shared/                       (unchanged)
├── channels/<channel>/           (NEW)
│   ├── data/, config/, logs/, cef-cache/, agents/
│   └── runtime/                  (single instance per channel)
└── dev/<branch>/                 (unchanged)
```

**Channel resolution priority:**
1. `AGENTMUX_CHANNEL=<name>` env override — any mode. Path: `channels/<channel>/`.
2. `Installed`/`Portable` w/o override → build-time default (`AGENTMUX_BUILD_CHANNEL_DEFAULT`).
3. `Dev { branch }` w/o override → name = `dev-<branch>`, path = `dev/<branch>/` (unchanged).

**`task package` sets `AGENTMUX_BUILD_CHANNEL_DEFAULT=dev-portable`**, so every locally-built portable shares one data dir. Bump v0.38.6 → v0.38.7 → v0.38.8 — same agents across all of them. Release CI does NOT call `task package`, so production builds get the `stable` default via the compile-time fallback (added in a follow-up if needed; today's release script and `cargo build` both work unchanged).

## What this PR does NOT do (per the 3-increment plan)

- **Increment B** (next): migration framework + pre-migration snapshots + forward-compat safety lock. Required before multiple released versions can safely share `stable`.
[...]

PR #1029, opened 2026-05-25
feat(srv): channels safety lock — refuse to open schema written by newer binary (Increment B.1)

AgentA-asaf:
**Increment B.1 of 3** from [`SPEC_DATA_CHANNELS_2026_05_24.md`](../blob/main/docs/specs/SPEC_DATA_CHANNELS_2026_05_24.md) §3.3. Builds on the Increment A landed in #1027. Tracking discussion: [#1026](https://github.com/agentmuxai/agentmux/discussions/1026).

## What this PR does

Adds the forward-compat **safety lock** the channels design depends on. Within a channel, multiple released AgentMux versions share one data dir; the lock keeps an older binary from silently opening a DB stamped at a schema version it doesn't recognize — the corruption mode Chrome avoids with its profile-too-new check and Postgres with catalog-version mismatch.

## Before/after

| | Before | After |
|---|---|---|
| `stamp_and_check_version` finds `user_version > current` | `WARN`, then unconditionally overwrites `user_version` back **down** to `current`. Returns `Ok`. | Returns `Err(SchemaTooNew { db, found, expected })`. **Does not touch** the on-disk stamp — preserves the newer value so the next launch of the upgraded binary still sees the right state. |
| User-facing | Silent corruption risk (older binary writes rows the newer schema interprets through unknown columns) | Loud, actionable error: name of the db file, found vs expected version, two recovery options (upgrade AgentMux, or `AGENTMUX_CHANNEL=<other>` for a fresh dir). |

## Why the on-disk no-downgrade invariant matters

If `stamp_and_check_version` had also blanked the version stamp on rejection, the next upgrade would have lost the signal that a newer schema had ever existed. Preserving it means:

1. User downgrades AgentMux → DB stays stamped at the newer version → this older binary refuses to open.
2. User upgrades back → newer binary sees its own stamp and proceeds.

The data itself is **never touched** on the rejected path. Recovery is non-destructive.

## NOT in this PR (Increment B.2)

- Pre-migration snapshot to `~/.agentmux/snapshots/` per spec §3.4. The current `run_object_schema` is purely additive (`CREATE TABLE IF NOT EXISTS` / `ALTER TABLE ADD COLUMN`), so there's nothing destructive to roll back from. Snapshot machinery becomes relevant the moment we ship a non-additive change; doing the plumbing now would be speculative.
- `meta.json` for cross-DB channel metadata. Useful for the Increment C import wizard; not blocking the lock.

## Tests

| Test | Asserts |
|---|---|
| `stamp_and_check_version_safety_lock_refuses_newer_db` | `user_version = 99` → returns `SchemaTooNew { found: 99, expected: OBJECT_SCHEMA_VERSION }`. **On-disk stamp stays at 99** after the rejection (the no-downgrade invariant). |
| `stamp_and_check_version_accepts_equal_or_lower` | Equal stamps re-stamp idempotently; lower stamps get bumped up (forward-migration path). |
| `stamp_and_check_version_stamps_on_fresh_db` | Fresh DB → gets the current version stamp. |
| (existing 14 `migrations::` tests) | All pass unchanged. |

No production-code call site uses `StoreError` exhaustively, so adding the `SchemaTooNew` variant is non-breaking at the API boundary.

## Sequencing

1. **A** — channels infrastructure ✓ (#1027, merged)
2. **B.1** — safety lock (this PR)
3. **B.2** — snapshot + `meta.json` (future PR; only when needed)
4. **C** — import wizard (future PR)

### DIFF
--- a/.changesets/1779694728-feat-srv-safety-lock-refuse-to-open-schema-written-by-newer--yycj.md
+++ b/.changesets/1779694728-feat-srv-safety-lock-refuse-to-open-schema-written-by-newer--yycj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): safety lock — refuse to open schema written by newer binary (channels Increment B.1)
--- a/agentmux-srv/src/backend/storage/error.rs
+++ b/agentmux-srv/src/backend/storage/error.rs
@@ -26,6 +26,26 @@ pub enum StoreError {
     #[error("json error: {0}")]
     Json(#[from] serde_json::Error),
 
+    /// The on-disk database was written by a NEWER AgentMux binary
+    /// than the one running now. Refusing to open it is the
+    /// channels-design safety lock from
+    /// `SPEC_DATA_CHANNELS_2026_05_24.md` §3.3 — a forward-compat
+    /// guard that prevents a downgrade from silently corrupting
+    /// state the newer schema added. The user must upgrade the
+    /// binary or switch channels (e.g.
+    /// `AGENTMUX_CHANNEL=experiment` for an empty side dir) to
+    /// recover.
+    #[error(
+        "{db}: this AgentMux is too old to open this data — schema v{found} \
+         on disk, this binary speaks v{expected}. Upgrade AgentMux, or set \
+         AGENTMUX_CHANNEL=<other> to use a fresh channel."
+    )]
+    SchemaTooNew {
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #906, opened 2026-05-18
feat(browser-pane): HTTP Basic/Digest auth prompt (Phase α)

AgentA-asaf:
## Summary

Sites that return 401 WWW-Authenticate: Basic / Digest were failing with \`ERR_INVALID_AUTH_CREDENTIALS (-338)\` because CEF's \`RequestHandler::get_auth_credentials\` wasn't wired (only \`on_render_process_terminated\` is implemented today). The host was silent, CEF treated the 401 as a fatal error, the pane showed a blank error page.

Phase α wires the full chain:

**Host (Rust)**
- New \`agentmux-cef/src/browser_pane/auth.rs\` — global registry parking CEF \`AuthCallback\`s keyed by uuid request_id (Mutex<HashMap>, with helpers \`register\` / \`take\` / \`pending_count\`).
- \`AgentMuxRequestHandler::auth_credentials\` returns 1 (async), registers the callback, broadcasts \`browser-pane-auth-required\` IPC event with origin/host/port/realm/is_proxy.
- IPC handlers \`browser_pane_auth_submit\` / \`browser_pane_auth_cancel\` pop the callback and call \`cb.cont(user, pass)\` / \`cb.cancel()\`.
- Non-browser-pane requests (the main host frontend) decline (return 0) — host frontend shouldn't hit auth challenges.

**Frontend (TS)**
- New \`BrowserAuthRequest\` variant on \`TabModalRequest\`.
- New \`BrowserAuthModalPanel\` (modal-v2 chrome) with username + password inputs. Username autofocuses, Enter submits, Escape cancels.
- \`browser-view.tsx\` subscribes to \`browser-pane-auth-required\` in onMount (lives in the view so \`useTabModal()\` SolidJS hook is in context) and opens the modal. Subscription cleanup in onCleanup.

**Spec**: [\`SPEC_BROWSER_PANE_HTTP_BASIC_AUTH_2026_05_18.md\`](docs/specs/SPEC_BROWSER_PANE_HTTP_BASIC_AUTH_2026_05_18.md)

Phase β (per-pane in-memory realm cache so subresource auth doesn't re-prompt) and Phase γ (OS keyring persistence, proxy auth, client certs) deferred.

## Test plan

- [ ] Open browser pane → navigate to https://pulse.asaf.cc/ (or any 401 site) → modal opens with origin + realm
- [ ] Enter correct credentials → page loads
- [ ] Cancel → 401 response body renders
[...]

PR #1035, opened 2026-05-25
diag(cef): instrument browser-pane auth callback + load-error entry points

AgentA-asaf:
**Diagnosis-only PR.** No behavior change — three `tracing::*!` calls added.

## Why

`https://pulse.asaf.cc` returns proper `401 WWW-Authenticate: Basic realm=\"Secure Area\"` (verified via `curl -I`), but in the browser pane the navigation fails with `ERR_INVALID_AUTH_CREDENTIALS (-338)` and renders the AgentMux \"Failed to load frontend\" fallback page. The HTTP Basic auth modal never appears.

Investigation: `[browser-pane-auth]` log lines from PR #906's `on_auth_credentials` callback are **completely absent** from `cef-debug.log` for the failing URL. Can't tell from the current logs whether:

- CEF never invoked `on_auth_credentials` (vtable issue, cached failure, security policy) — OR
- CEF invoked it but one of the silent early-return paths fired

## What this adds

| Log line | Location | Captures |
|---|---|---|
| `[browser-pane-auth][ENTRY]` | Top of `on_auth_credentials` BEFORE every early return | origin / host / port / realm / is_proxy / has_browser / has_callback |
| `[load-error][ENTRY]` | Top of `on_load_error` BEFORE ERR_ABORTED filter + sub-frame filter | url / error / is_main_frame / aborted |
| `[browser-pane-auth] callback is None` | Previously-silent early return | block_id (so we know which pane) |

Zero new RPC, no schema change, no UI change.

## Decision tree after merge

After `task dev` rebuilds the host:

1. Reproduce: navigate browser pane to `https://pulse.asaf.cc`.
2. Tail `~/.agentmux/dev/<branch>/logs/cef-debug.log` for `[browser-pane-auth][ENTRY]` and `[load-error][ENTRY]`.
3. Diagnosis branch:
   - **`[browser-pane-auth][ENTRY]` visible** → CEF IS calling the callback. The bug is downstream (callback resolution, event dispatch, modal rendering). Check the immediate next log lines for which early-return fired or what the event dispatch path showed.
   - **`[browser-pane-auth][ENTRY]` absent + `[load-error][ENTRY]` for the same URL** → CEF is suppressing the auth callback entirely. Likely root causes: HTTP cache has cached \"no creds\" response, security policy rejecting the 401 before auth flow, or vtable wire-up gap in the CEF binding.

## What this PR does NOT fix

- The actual auth bug (waiting on diagnosis).
- A separate known bug: `on_load_error` renders the \"Failed to load AgentMux frontend\" page (intended for the host frontend at localhost:5173) for **any** browser-pane load error. That's wrong UI but cosmetic; will be fixed in a follow-up PR.

## Test plan

- After merge → `task dev` rebuild → reload host → navigate to pulse.asaf.cc → check `cef-debug.log` for the new `[ENTRY]` lines.
- No automated test surface — this is unconditional `tracing::info!` plumbing.

## Build status

Local `cargo check -p agentmux-cef` blocked by a separate cef-dll-sys build-dir corruption (rename collision between half-extracted artifacts; unrelated to the diff). The diff is purely additive `tracing::*!` calls so syntactic risk is minimal. Bot review will catch any compile issue, and post-merge `task dev` will surface it cleanly. If it doesn't compile we can revert/amend without losing diagnostic value.

### DIFF
--- a/.changesets/1779716363-diag-cef-instrument-on-auth-credentials-on-load-error-entry--lr3h.md
+++ b/.changesets/1779716363-diag-cef-instrument-on-auth-credentials-on-load-error-entry--lr3h.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+diag(cef): instrument on_auth_credentials + on_load_error entry points (browser-pane auth diagnosis)
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -1128,6 +1128,31 @@ impl AgentMuxHandler {
         debug_assert_ne!(currently_on(ThreadId::UI), 0);
 
         let error_code_raw = sys::cef_errorcode_t::from(error_code);
+
+        // [DIAG] Unconditional entry log — captures every load error
+        // BEFORE the ERR_ABORTED filter, sub-frame filter, or fallback-
+        // page render. Pair with `[browser-pane-auth][ENTRY]` to
+        // diagnose the auth-modal-doesn't-appear path: if we see a
+        // load-error here with no preceding auth-credentials entry,
+        // CEF skipped the auth flow entirely.
+        let failed_url_dbg = failed_url.map(CefString::to_string).unwrap_or_default();
+        let error_text_dbg = error_text.map(CefString::to_string).unwrap_or_default();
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1037, opened 2026-05-25
fix(cef): pass --disable-chrome-login-prompt so OnGetAuthCredentials fires for browser-pane auth

AgentA-asaf:
**Fixes browser-pane HTTP Basic / Digest auth modal in CEF 146.** One-line `cmd.append_switch(...)` + research doc.

## Symptom

Browser pane navigating to a 401-protected URL (verified on `https://pulse.asaf.cc` returning `WWW-Authenticate: Basic realm=\"Secure Area\"`) fails with `ERR_INVALID_AUTH_CREDENTIALS (-338)` and renders the AgentMux \"Failed to load\" fallback page. The HTTP Basic auth modal that [#906](https://github.com/agentmuxai/agentmux/pull/906) + four follow-ups shipped never appears.

## Diagnosis (from PR #1035's instrumentation)

The smoking gun in the host log on the failed request:

```
[load-error][ENTRY] url=\"https://pulse.asaf.cc/\" 
                   error=\"ERR_INVALID_AUTH_CREDENTIALS\" (-338) 
                   is_main_frame=true 
                   aborted=false
```

But **no `[browser-pane-auth][ENTRY]`** for the same request — the very first line of our `on_auth_credentials` callback, which logs unconditionally before any early return. CEF is bypassing our auth callback entirely.

Cache theory ruled out (cleared `~/.agentmux/dev/main/cef-cache/Default/Cache/` between attempts).

## Root cause

CEF 146 only ships the Chrome runtime (the Alloy runtime was removed in 2024). The Chrome runtime owns its own native login-prompt UI and intercepts 401/407 challenges **before** the embedder's `RequestHandler::OnGetAuthCredentials` callback gets to see them. So our handler in `client/mod.rs:1442` is wired correctly — CEF just never calls it.

This is [CEF issue #3603](https://github.com/chromiumembedded/cef/issues/3603) (\"chrome: GetAuthCredentials requires --disable-chrome-login-prompt\"). The sister binding [CEF4Delphi sets the same switch by default](https://github.com/salvadordf/CEF4Delphi/issues/520) for the same reason.

## Fix

```diff
+ cmd.append_switch(Some(&CefString::from(\"disable-chrome-login-prompt\")));
```

Added in `agentmux-cef/src/app.rs`'s `on_before_command_line_processing` alongside the existing `disable-sync` / `disable-extensions` switches. Tells the Chrome runtime to defer auth challenges back to the embedder.

## Verification plan

After merge + `task dev` rebuild (or next release for the portable):

[...]

PR #1034, opened 2026-05-25
feat(modal): parameterize ModalLayer over scope — launch modal locks just its pane

AgentA-asaf:
**Implements [`SPEC_LAUNCH_MODAL_PANE_SCOPE_2026_05_25.md`](../blob/agenta/launch-modal-pane-scope/docs/specs/SPEC_LAUNCH_MODAL_PANE_SCOPE_2026_05_25.md).** Spec ships with the PR.

## Problem

The agent launch modal (and chained install / new-identity / new-memory / create-from-template / browser-auth flows) used `scope=\"tab\"`, which inerts the **entire tab's tile layout** — every other pane in the tab (browser, terminal, second agent) loses focus + scroll for the lifetime of the modal. The modal is configuring **one** agent pane; its authority shouldn't extend past that pane.

## Approach — Option C (DRY)

Three options considered (spec §4.2):

| Option | What | Trade-off |
|---|---|---|
| A | Direct `<Modal scope=\"pane\">` in agent pane; state machine moves into AgentPicker | Smallest infra; biggest refactor inside the picker |
| B | Mirror layer: copy `TabModalLayer` → `PaneModalLayer` with `scope=\"pane\"` | ~700 lines duplicated; two layers diverge over time |
| **C** | **Parameterize**: promote `TabModalLayer` to generic `ModalLayer` with `scope` prop, mount at both tab and pane levels with the same dispatch table | **DRY, best end state** |

This PR ships **C**.

## What changed

**New canonical primitive** (`frontend/app/element/`):

- `ModalLayer.tsx` — scope-parameterized dispatcher. Takes `scope: \"tab\" | \"pane\"`, selects the matching `Scope.Provider` (`TabModalScope.Provider` vs `PaneModalScope.Provider`), passes the same scope to the inner `<Modal>`. Same dispatch table as before — no duplicated request handling.
- `modal-layer.ts` — types + context + `useModalLayer()` hook. Renames: `TabModalApi` → `ModalLayerApi`, `TabModalRequest` → `ModalLayerRequest`, `useTabModal` → `useModalLayer`. Request shapes unchanged.
- `modal-layer.scss` — class rename `.tab-modal-content` → `.modal-layer-content`, same crossfade keyframe.

**Deleted** (replaced by the above): `frontend/app/tab/{TabModalLayer.tsx, tab-modal.ts, tab-modal.scss}`. Net diff: more deleted than added.

**Mount sites:**

[...]

PR #1038, opened 2026-05-25
feat(browser-pane): wrap browser view in pane-scope ModalLayer — HTTP auth modal locks only the pane

AgentA-asaf:
**Follow-up to [#1037](https://github.com/agentmuxai/agentmux/pull/1037).** The auth modal now fires (per #1037), but as a TAB modal — sibling panes go inert. This PR makes it pane-scope.

## Before

After #1037 landed and `OnGetAuthCredentials` started firing, `useModalLayer()` inside `browser-view.tsx` resolved to the OUTER tab-scope `ModalLayer` (from `tabcontent.tsx`) because the browser pane didn't wrap itself in its own pane-scope layer. So the HTTP Basic auth modal opened with the **tab** as its lock region — backdrop covered the whole tab, sibling panes (terminal, agent, another browser) went inert.

User confirmed during dev smoke: \"modal appeared .. can it be made a pane modal? (right now its a tab modal right?)\".

## Fix

Splits `BrowserViewComponent` into a thin outer wrapper + inner function:

```tsx
export function BrowserViewComponent(props): JSX.Element {
    return (
        <ModalLayer scope=\"pane\">
            <BrowserViewInner model={props.model} />
        </ModalLayer>
    );
}

function BrowserViewInner(props): JSX.Element {
    // ... all existing logic, including `useModalLayer()` ...
}
```

The split is required because Solid's hook resolves up the JSX tree at execution time. Putting both the `<ModalLayer>` JSX and the `useModalLayer()` call in the same function body would have the hook resolve to the OUTER (tab) layer — the wrapper hasn't been mounted by the time the hook runs. Splitting ensures the consumer is a CHILD of the provider, so context resolution lands on the pane-scope layer. Same pattern as `AgentViewWrapper` in #1034.

## Effect

When a browser pane navigates to a 401-protected URL:
- Modal backdrop covers only the browser pane.
- `inert` locks only the browser pane content.
- All other panes in the tab (terminal, agent, another browser) stay clickable + scrollable.
- Tab bar + title bar untouched.

Two browser panes hitting concurrent 401s in the same tab can each show their own modal independently (modal-stack semantics per `SPEC_UNIFIED_MODAL_SYSTEM_2026_05_21` §6 already account for non-overlapping pane locks).

## Tests

- **1028 / 1028 vitest pass.**
- `browser-model.test.ts` (10/10) unchanged.
- No new test surface needed; the change is structural (component wrapping) with no logic change in the inner.

### DIFF
--- a/.changesets/1779722120-feat-browser-pane-wrap-browser-view-in-pane-scope-modallayer-9b82.md
+++ b/.changesets/1779722120-feat-browser-pane-wrap-browser-view-in-pane-scope-modallayer-9b82.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(browser-pane): wrap browser view in pane-scope ModalLayer (auth modal locks only the pane)
--- a/frontend/app/view/browser/browser-view.tsx
+++ b/frontend/app/view/browser/browser-view.tsx
@@ -3,6 +3,7 @@
 
 import { createEffect, createSignal, onCleanup, onMount, Show, type JSX } from "solid-js";
 import { invokeCommand, listenEvent } from "@/app/platform/ipc";
+import { ModalLayer } from "@/element/ModalLayer";
 import { useModalLayer } from "@/element/modal-layer";
 import type { BrowserViewModel } from "./browser-model";
 import "./browser-view.scss";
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1030, opened 2026-05-25
fix(term): await fonts.ready + NaN guard + post-paint refit to fix jumbled startup

AgentX-asaf:
## Summary

Fixes the intermittent jumbled text on terminal-pane startup (especially with Claude Code / Ink TUIs) that "clicks into place" after a manual resize or Ctrl+Wheel zoom.

Root cause is a font-load race in `TermWrap.init()`:

1. `terminal.open()` mounts xterm.
2. `customFit()` runs immediately and calls `fitAddon.proposeDimensions()`, which measures rendered cell width from the DOM.
3. If `Hack` (or whatever term font) hasn't loaded yet, FitAddon uses fallback metrics → wrong `cols` → `sendTermSize()` tells the PTY the wrong size → Ink emits ANSI cursor sequences that don't line up with where glyphs actually land.

The existing `[startup-perf] initApp (fonts ready): 239.8ms` benchmark already shows there is real time between init start and fonts-ready, so the race is observable on this machine.

Three changes in `frontend/app/view/term/termwrap.ts`:

- **Await fonts before first fit** — `await document.fonts?.ready` between `terminal.open()` and the first `customFit()`.
- **NaN guard in `customFit()`** — `proposeDimensions()` can return `{cols: NaN, rows: NaN}` (xterm.js#4338); the existing `!dims` check doesn't catch it because the object is truthy and NaN propagates into `terminal.resize()` and corrupts the grid.
- **One rAF re-fit after init** — schedules a second `customFit()` + `sendTermSize()` in `requestAnimationFrame` so any remaining layout shift (slow CSS, late font swap) propagates as a SIGWINCH before the PTY produces meaningful output.

Investigation write-up included at `docs/terminal-jumbled-startup-investigation.md`.

Changeset: `patch` (RFC #857 Phase 2 workflow — no version bump in this PR).

## References

- xterm.js [#4338](https://github.com/xtermjs/xterm.js/issues/4338) — FitAddon NaN dimensions
- xterm.js [#4830](https://github.com/xtermjs/xterm.js/issues/4830) — fonts render incorrectly
- xterm.js [#4841](https://github.com/xtermjs/xterm.js/issues/4841) — FitAddon resizes incorrectly
- xterm.js [#4113](https://github.com/xtermjs/xterm.js/issues/4113) — DPR / zoom mismatch
- hermes-ide [#113](https://github.com/hermes-hq/hermes-ide/issues/113) — same pattern + WebGL ghost overlays
[...]

PR #1040, opened 2026-05-25
fix(term): use document.fonts.load() to force-load Hack before first fit (#1030 follow-up)

AgentA-asaf:
## Summary

Follow-up to PR #1030. The jumbled-glyph repro **resurfaced** when opening
multiple terminal panes in quick succession on a cold cache — user-reported on
2026-05-25 against the latest dev build (which already contains #1030).

## Why #1030 was insufficient

`await document.fonts.ready` was the wrong primitive. Per the CSS Font Loading
spec, `fonts.ready` returns a Promise that resolves when no font loads are
currently pending. It does **not** request specific fonts.

Browsers load web fonts lazily — the WOFF/WOFF2 for `Hack` is not requested
until something actually renders a glyph in that family. In `TermWrap.init()`:

1. `terminal.open()` mounts the xterm DOM with `font-family: Hack`. Browser
   sees the rule but does **not** request `Hack` yet (nothing has asked it to
   paint a glyph in that family).
2. `await document.fonts.ready` resolves **immediately** — zero pending loads.
3. `customFit()` → `proposeDimensions()` measures cell width. The measurement
   step is what finally triggers `Hack` to be requested, but
   `proposeDimensions()` returns synchronously using fallback metrics.
4. Wrong `cols` → SIGWINCH at wrong size → PTY spawns → Ink emits cursor
   sequences sized to wrong cols → jumbled glyphs.

The rAF re-fit a frame later cannot undo cursor-positioning sequences Ink has
already issued against the wrong dimensions. Resize/zoom "clicked it into
place" because that triggered a fresh full redraw.

## Fix

Replace `fonts.ready` with `document.fonts.load(`${fontSize}px "${fontFamily}"`)`.

`load()` actively **requests** the named face and resolves only when that face
is available. Parallel terminals coalesce on the browser font cache, so
opening N panes in parallel pays the network cost once.

Also pre-load `bold` and `italic` variants xterm.js uses (cells render with
separate font faces; without pre-load the first bold/italic glyph would
trigger the same race at smaller scale).

The 1s bounded race and post-init rAF re-fit from #1030 are kept as
belt-and-suspenders.

## Other xterm mount: AgentInstallModal

Applied the same pattern to the install modal's parallel `new Terminal()` /
`fitAddon.fit()` codepath. Lower practical impact (renders short npm output,
not a TUI), but the same race is present.

## Files

- `frontend/app/view/term/termwrap.ts` — swap `fonts.ready` for explicit
  `fonts.load()` of regular + bold + italic.
- `frontend/app/view/agent/components/AgentInstallModal.tsx` — same pattern.
- `docs/terminal-jumbled-startup-investigation.md` — added "PR #1030 was
  insufficient" follow-up section with the spec analysis and final fix.

Changeset: `patch` (RFC #857 Phase 2 workflow — no version bump in this PR).

## Test plan

- [x] `npx tsc --noEmit` on edited files — clean.
- [ ] `task dev` running with HMR — open ≥3 fresh terminal panes back-to-back
      on a cold cache (kill DEV data dir to force), verify no jumbled output
      and Claude Code's prompt renders correctly first time.
- [ ] Open the agent install modal cold-cache, verify the xterm container
      sizes correctly on first paint (no fallback-font overshoot).
- [ ] Compare cold vs warm cache — once Hack is in the browser cache the
      `fonts.load()` resolves synchronously so there's no perceptible delay.

## References

- MDN — `FontFaceSet.load()` —
  <https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/load>
- CSS Font Loading Module Level 3, §3.4 —
  <https://drafts.csswg.org/css-font-loading-3/#font-face-set-interface>
- xterm.js #4830 — community consensus workaround is `fonts.load()`.

### DIFF
--- a/.changesets/1779725954-fix-term-use-document-fonts-load-to-force-load-hack-before-f-yo5i.md
+++ b/.changesets/1779725954-fix-term-use-document-fonts-load-to-force-load-hack-before-f-yo5i.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): use document.fonts.load() to force-load Hack before first fit (#1030 follow-up)
--- a/docs/terminal-jumbled-startup-investigation.md
+++ b/docs/terminal-jumbled-startup-investigation.md
@@ -76,3 +76,91 @@ Three changes in `frontend/app/view/term/termwrap.ts`:
 
 Together these eliminate the three known root causes: pre-font-load
 measurement, NaN propagation, and post-mount layout shift.
+
+## Follow-up — PR #1030 was insufficient (2026-05-25)
+
+After PR #1030 merged, the jumbled-glyph repro **resurfaced** when opening
+multiple terminal panes in quick succession on a cold cache. Diagnosis:
+
+`await document.fonts.ready` is **not** the right primitive. Per the CSS
+Font Loading spec, `fonts.ready` returns a Promise that resolves when no
+currently-pending font loads remain. It does **not** request specific fonts.
+Browsers load web fonts lazily — the WOFF/WOFF2 download for `Hack` is only
+kicked off when something actually needs to render a glyph in that family.
+
+In `TermWrap.init()` the sequence was:
+
+1. `terminal.open()` mounts the xterm DOM with `font-family: Hack`. The
+   browser sees the rule but **does not request `Hack` yet** — nothing has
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1041, opened 2026-05-25
fix(term): move document.fonts.load() BEFORE terminal.open() — xterm caches metrics at open() time (#1040 follow-up)

AgentA-asaf:
## Summary

Third attempt at killing the terminal-pane jumbled-startup race. PRs #1030
and #1040 both placed the font wait AFTER `terminal.open()`. **That was
always too late.** User reported the repro persisted after #1040 merged:
*"happens more often when I load many terminals at once, it's a timing
issue."*

## Real root cause

xterm.js caches cell-width metrics inside its render service at
`terminal.open()` time, using whatever font is currently rendering at
that moment.

- If Hack hasn't loaded yet when `open()` runs, xterm caches **fallback
  (Courier) metrics**.
- `FitAddon.proposeDimensions()` then reads those **cached metrics**, not
  a fresh DOM measurement.
- When Hack later loads, the browser swaps the visual font but xterm's
  cached cell dimensions do **not** invalidate.

So `proposeDimensions()` returns wrong cols forever, regardless of whether
Hack has loaded or how many post-open `customFit()` calls happen.

With many terminals opening in parallel, every terminal hits `open()`
before any `fonts.load(...)` resolves, and all of them cache fallback
metrics simultaneously. The rAF re-fit from #1030 dutifully calls
`customFit()` again — and reads the same stale cached value.

PRs #1030 and #1040 closed the race between font-load and
`proposeDimensions()` reading the DOM, but `proposeDimensions()`
doesn't actually read the DOM in any of the paths that matter — it
reads xterm's cached metrics.

## Fix

Await `document.fonts.load(spec)` **BEFORE** `terminal.open()`. Parallel
terminals dedupe on the underlying browser load (one fetch on cold
cache), then each `open()` measures cell metrics with Hack actually
rendering, and the cached width is correct on the first try.

The NaN guard in `customFit()` and the rAF re-fit from #1030 remain as
belt-and-suspenders for late layout shifts (CSS animations, window-zoom
changes), but they are now defense-in-depth, not the primary mechanism.

## AgentInstallModal

Same restructure — font wait before `terminal.open()`. The xterm there
runs async-after-mount now, so the welcome `writeln()` and the
ResizeObserver setup move inside the async block.

## Files

- `frontend/app/view/term/termwrap.ts` — relocate the font-load block to
  before `terminal.open()`; remove the post-open redundant copy.
- `frontend/app/view/agent/components/AgentInstallModal.tsx` — wrap
  `terminal.open()` + initial fit + welcome line in an async
  `openAndFit()` that awaits fonts first.
- `docs/terminal-jumbled-startup-investigation.md` — added "PR #1040 also
  insufficient" follow-up section with the cached-metrics diagnosis and
  the lesson about distinguishing measurement timing points.

Changeset: `patch` (RFC #857 Phase 2 workflow).

## Test plan

- [x] `npx tsc --noEmit` — clean on edited files.
- [ ] Kill dev process to clear in-memory FontFaceSet, restart `task dev`,
      open 6+ terminal panes back-to-back. Verify no jumble in any pane,
      including the first.
- [ ] Open install modal cold and warm cache — verify no font-size
      overshoot.

## References

- #1030 — original `fonts.ready` fix (insufficient — vacuous resolve).
- #1040 — switched to `fonts.load(spec)` (still insufficient — placed
  after `terminal.open()`).
- xterm.js render-service `dimensions` caching — measured once at open,
  invalidated only on font-option change or DPR change, not on lazy
  font swap.

### DIFF
--- a/.changesets/1779727244-fix-term-move-document-fonts-load-before-terminal-open-xterm-xm64.md
+++ b/.changesets/1779727244-fix-term-move-document-fonts-load-before-terminal-open-xterm-xm64.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): move document.fonts.load() BEFORE terminal.open() — xterm caches metrics at open() time (#1040 follow-up)
--- a/docs/terminal-jumbled-startup-investigation.md
+++ b/docs/terminal-jumbled-startup-investigation.md
@@ -147,6 +147,71 @@ try {
 } catch (_) { /* font API unavailable or face unknown — fall through */ }
 ```
 
+## Follow-up #2 — PR #1040 also insufficient (2026-05-25, hours later)
+
+After PR #1040 merged and the dev process was restarted to clear the
+FontFaceSet, the user reported the jumbled glyphs STILL reproduced when
+opening many terminals in parallel — "happens more often when I load
+many terminals at once, it's a timing issue."
+
+The real bug: **xterm.js caches cell-width metrics at `terminal.open()`
+time**, using whatever font is currently rendering at that moment. If
+Hack hasn't loaded yet when `open()` runs, xterm caches fallback
+(Courier) metrics into its render service's dimensions object.
+`FitAddon.proposeDimensions()` reads that cached width — not a fresh
+DOM measurement — so it returns wrong cols *forever*, regardless of
+whether Hack has loaded later. The font visually swaps in when it
+finishes loading, but the cached cell width does not invalidate.
+
+Both #1030 and #1040 placed the `await fonts.ready` / `fonts.load(...)`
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Closing — the v2 reorder (await fonts.load() before terminal.open()) didn't fix the underlying bug. Tested locally with diagnostic instrumentation; restored-terminal jumble persists due to a deeper race I haven't pinned down (cols drift between customFit and post-restore, buffer/cursor desync that fixes on manual resize). Capturing what was learned in a follow-up; reopening will need a fresh approach, not another reorder.

PR #1043, opened 2026-05-25
fix(term): PSReadLine cursor-desync thaw on terminal init (#1042)

### LINKED ISSUE #1042: term-jumble: rapid pane splits leave last-opened terminal(s) jumbled; 3 PRs + 8 fix attempts haven't resolved it
## Summary

When opening multiple terminal panes in rapid succession (typical user action), the last-opened terminal(s) render with jumbled glyphs and a cursor that desyncs on Enter. Manual pane resize or zoom fixes it. Vite HMR reload also fixes it.

This has been chased through three merged-but-insufficient PRs:
- #1030 — `await document.fonts.ready` after `terminal.open()`
- #1040 — `document.fonts.load(spec)` after `terminal.open()`
- #1041 (closed) — `document.fonts.load(spec)` BEFORE `terminal.open()`

None of these fix it. Additional attempts (see below) also failed.

## Repro

1. Open AgentMux
2. Split-pane the terminal widget 4–6 times in rapid succession
3. The last opened terminal(s) show jumbled output: cursor not where the shell thinks it is, pressing Enter makes cursor jump around. Box-drawing breaks.
4. Manual pane resize → instantly clean.

### What's NOT the cause

- **Font load timing.** Hack is loaded in FontFaceSet at `init-start` (`hackCheck12=true`). Cell width transitions from default 7.225 → 7.214 between `post-open` and `post-customfit`, but stays stable thereafter.
- **Cell width / cols measurement.** Diagnostic logs show stable cols and cell width from `post-customfit` through `post-resync` through `raf-refit`.
- **Renderer choice.** Replacing xterm.js with `ghostty-web` (Coder's WASM-compiled Ghostty parser, xterm-API-compatible) **made it worse** — even more visible jumble, plus special characters and overlaid chars. So the bug is NOT renderer-specific.
- **xterm's `windowsPty` + `reflowCursorLine` options** — made it worse.
- **Debounce on `handleResize_debounced`** (was 50ms) — removed; no fix.
- **Forced re-measurement via `_charSizeService.measure()` + `refresh()`** — no fix.
- **Two-frame stable-wait before `terminal.open()`** — no fix.
- **Fake resize cycle (`resize(c-1, r); resize(c, r)`) at end of init** — no fix.
- **`customFit()` before `loadInitialTerminalData()` instead of after** — no fix.

### What it IS

The pattern strongly suggests a **Solid + flex-layout timing issue at mount time**. When N terminals mount in rapid succession:
- Each existing pane's container shrinks (ResizeObserver fires on each)
- New pane mounts at its slot, runs `init()`
- During `init()`, the layout cascade is still propagating
- xterm captures whatever transient state exists at `terminal.open()` time and doesn't recover

HMR fixes it because the component fully re-mounts when layout is settled. Manual resize fixes it because it triggers a fresh xterm `resize()` call through `handleResize`.

### Cross-stream contamination hypothesis (un-investigated)

`ghostty-web` showed "special characters and overlaid characters" — could be one terminal's PTY data leaking into another's xterm. `getFileSubject(blockId, "term")` keys per-block so subjects don't get crossed at that layer, but worth verifying nothing upstream of that point has the same key collision risk.

## Diagnostic instrumentation (kept in this branch)

`frontend/app/view/term/termwrap.ts` was instrumented with `[term-jumble:<phase>]` console logs at every init phase. Logs land in `~/.agentmux/dev/<branch>/logs/agentmux-host-v<version>.log.<date>` tagged as `[fe]`. Grep `term-jumble:` for the diagnostic chain.

## Realistic next steps

1. **Auto-corrective re-mount.** After init, wait 500ms then dispose+recreate the xterm instance. Most likely to actually fix the bug since HMR's re-mount provably does. Risk: visible flicker on every new terminal.
2. **Defer terminal init globally.** Block all terminal `init()` calls until the layout system signals "no pending updates." Requires hooking into Solid's reactive cascade — invasive.
3. **File against xterm.js upstream.** Build a minimal repro outside AgentMux, file there.
4. **Live with it** + pin a "if your terminal looks wrong, drag the pane divider" note in user-facing docs.

## Reference

Long-format investigation lived in `docs/terminal-jumbled-startup-investigation.md` on the abandoned branch `agenta/term-font-load-before-open` (now deleted). Session transcript in AgentA's logs around 2026-05-25.

AgentA-asaf:
Closes #1042.

## Summary

Two terminal-pane fixes, plus the analysis doc that nailed the diagnosis.

### 1. PSReadLine cursor-desync "thaw"

The bug behind #1042. When N terminal panes are opened in rapid
succession, terminals that don't get subsequent sibling-split shrinks
come up visually broken — cursor jumps on Enter, prompt mis-layout.
Three prior PRs (#1030, #1040, #1041-closed) attempted font-load and
renderer-side fixes — none worked because **the bug is shell-side,
not xterm-side**.

**Root cause:** pwsh + PSReadLine maintain their own `(row, col)`
cursor tracking. When the only init-time resize is xterm's default
`80 × 24` → final (e.g. `14 × 33`), PSReadLine emits its first prompt
against the inherited `cols=80` ConPTY environment. The SIGWINCH
from `sendTermSize` then lands as the final cols. The prompt is in
the buffer at 80-col wrap points; xterm now renders at 14-col grid;
PSReadLine's tracked cursor diverges from xterm's actual cursor.
Pressing Enter sends a `\r` — PSReadLine emits its next prompt at
its (wrong) tracked cursor → "cursor jumps around".

Terminals that DID get subsequent sibling-shrink resizes accumulated
extra SIGWINCH events that re-synced PSReadLine — which is why some
terminals come up clean while others stay broken in the same burst.

**Diagnosis:** correlating the `[term-jumble:onRender]` chain (xterm
paint events with their cols/rows) with user-reported jumbled state.
Out of 9 terminals: **5 with single-transition (80 → final) history
matched user-reported 5 broken**; the 4 with multi-transition history
were clean.

**Fix:** 250ms post-init (enough for the first prompt to land),
replay one synthetic `+1` / `-1` col resize cycle, split across rAF
ticks so each fires as a distinct SIGWINCH (xterm coalesces
back-to-back same-frame resizes into one). Mimics what the "clean"
terminals naturally receive.

### 2. Black gap right of the scrollbar

Separate, smaller fix. xterm sets `terminal.element` width inline as
`cols × cellWidth` (e.g. `14 × 7.214 ≈ 101px`), leaving ~5–11px
between the terminal's right edge and the container's right edge —
filled by the container's dark background, visible as a black strip
beside the scrollbar.

CSS-stretches `.terminal` to fill the container so xterm's theme
background occupies the residual and the scrollbar sits flush against
the container's right edge. `!important` is required because xterm
sets width inline.

### 3. Analysis doc

`docs/analysis/TERM_JUMBLE_STRUCTURED_2026_05_25.md` now includes
§7a documenting the H6 confirmation: the data showing 1-transition
vs multi-transition correlation, the refined PSReadLine-desync
mechanism, and the thaw fix.

## Files

- `frontend/app/view/term/termwrap.ts` (+39) — H6 thaw cycle
- `frontend/app/view/term/term.scss` (+9) — scrollbar gap fix
- `docs/analysis/TERM_JUMBLE_STRUCTURED_2026_05_25.md` (+90) — §7a
- `.changesets/...md` — patch changeset

## Test plan

- [x] Open 9 terminal panes in rapid succession on Windows — all
      render correctly, no cursor desync (user-verified live)
- [x] Close all, reopen (restore path) — all clean (user-verified live)
- [x] Existing single-terminal workflows continue to work (the thaw
      cycle is essentially a +1/-1 no-op visually; brief 16ms flicker)
- [ ] macOS / Linux regression check (different default shells —
      bash/zsh don't have PSReadLine; thaw cycle there is a no-op +
      brief flicker; verify no visual regression)

## Notes for reviewers

- This is a **workaround**, not a root-cause backend-side fix. The
  cleaner alternative would be deferring `resyncController("init")`
  until layout has been stable for ~200ms, so the shell spawns ONCE
  at its final size and never sees the bad initial SIGWINCH. Doc
  §7a discusses the trade-off. Workaround chosen for smaller surface
  area and no shell-spawn-latency regression risk.
- The 250ms delay was tuned empirically; lower values race the
  first-prompt emission, higher values delay the visual fix
  unnecessarily.
- The `+1 / -1` split-across-rAF is essential — same-frame resizes
  coalesce in xterm.

Closes #1042

### DIFF
--- a/.changesets/1779750444-fix-term-psreadline-cursor-desync-thaw-close-scrollbar-black-l7nd.md
+++ b/.changesets/1779750444-fix-term-psreadline-cursor-desync-thaw-close-scrollbar-black-l7nd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): PSReadLine cursor-desync thaw + close scrollbar black gap (#1042)
--- a/.changesets/1779750444-fix-term-psreadline-cursor-desync-thaw-close-scrollbar-black-l7nd.md
+++ b/.changesets/1779750444-fix-term-psreadline-cursor-desync-thaw-close-scrollbar-black-l7nd.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(term): PSReadLine cursor-desync thaw + close scrollbar black gap (#1042)
--- a/.changesets/1779750444-fix-term-psreadline-cursor-desync-thaw-on-terminal-init.md
+++ b/.changesets/1779750444-fix-term-psreadline-cursor-desync-thaw-on-terminal-init.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): PSReadLine cursor-desync thaw on terminal init (#1042)
--- a/agentmux-cef/src/commands/window.rs
+++ b/agentmux-cef/src/commands/window.rs
@@ -587,6 +587,17 @@ pub fn toggle_devtools(state: &Arc<AppState>, args: &serde_json::Value) -> Resul
     Ok(serde_json::Value::Null)
 }
 
+/// Open DevTools focused on the element at the given window-relative
+/// coordinates. Equivalent to Chrome's right-click → Inspect Element.
+/// Used by the pane context menu's Inspect entry.
+pub fn inspect_element_at(state: &Arc<AppState>, args: &serde_json::Value) -> Result<serde_json::Value, String> {
+    let label = args.get("label").and_then(|v| v.as_str()).unwrap_or("main");
+    let x = args.get("x").and_then(|v| v.as_i64()).unwrap_or(0) as i32;
+    let y = args.get("y").and_then(|v| v.as_i64()).unwrap_or(0) as i32;
+    crate::ui_tasks::post_inspect_element_at(state, label, x, y);
+    Ok(serde_json::Value::Null)
+}
+
 /// Resolve the base URL for the frontend.
 /// Production: IPC server serves static files from `frontend/` next to the exe.
 /// Dev: Vite dev server at `http://localhost:5173`.
--- a/agentmux-cef/src/ipc.rs
+++ b/agentmux-cef/src/ipc.rs
@@ -252,6 +252,7 @@ async fn route_command(
         "move_window_by" => commands::window::move_window_by(state, args),
         "set_window_position" => commands::window::set_window_position(state, args),
         "toggle_devtools" => commands::window::toggle_devtools(state, args),
+        "inspect_element_at" => commands::window::inspect_element_at(state, args),
         "show_context_menu" => {
             tracing::debug!("show_context_menu: handled in JS overlay");
             Ok(serde_json::Value::Null)
--- a/agentmux-cef/src/ui_tasks.rs
+++ b/agentmux-cef/src/ui_tasks.rs
@@ -506,6 +506,48 @@ pub fn post_show_dev_tools(state: &Arc<AppState>, label: &str) {
     post_task(ThreadId::UI, Some(&mut task));
 }
 
+// ── DevTools — Inspect Element at coordinates ─────────────────────────────
+
+wrap_task! {
+    pub struct InspectElementAtTask {
+        state: Arc<AppState>,
+        label: String,
+        x: i32,
+        y: i32,
+    }
+
+    impl Task {
+        fn execute(&self) {
+            let browser = match self.state.get_browser(&self.label) {
+                Some(b) => b,
+                None => {
+                    tracing::warn!("[devtools] inspect-at: browser '{}' not found", self.label);
+                    return;
+                }
+            };
+
+            match browser.host() {
+                Some(host) => {
+                    // The 4th arg to show_dev_tools is `inspect_element_at: Option<CefPoint>`
+                    // in window-relative coords. CEF opens DevTools (creating it if not
+                    // already open) and selects the element at that point, equivalent to
+                    // Chrome's right-click → Inspect Element flow.
+                    let point = Point { x: self.x, y: self.y };
+                    host.show_dev_tools(None, None, None, Some(&point));
+                }
+                None => {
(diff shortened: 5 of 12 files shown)


a5af:
@codex review

AgentA-asaf:
Force-pushed branch with the scrollbar SCSS attempts removed (they targeted the wrong DOM layer — see follow-up analysis below) and an additional commit adding **Inspect Element** to the pane right-click menu.

**Branch now contains two commits:**

1. `3a07bc4b` — **PSReadLine thaw fix** (the actual #1042 fix; unchanged from the original review). Commit message rewritten to drop the SCSS reference since the SCSS attempts were reverted.

2. `4cff638b` — **NEW: Inspect Element + scrollbar gap analysis doc**:
   - Adds an "Inspect Element" entry to the bottom of pane context menus (body + header right-click). Click → CEF DevTools opens focused on the right-clicked element via `host.show_dev_tools(..., inspect_element_at: Point)`. Falls back to plain `toggleDevtools` when the host binary doesn't yet recognize the new IPC.
   - Adds `docs/analysis/TERM_SCROLLBAR_GAP_2026_05_25.md` — structured writeup of the scrollbar gap bug. Three CSS attempts during this session targeted `.terminal` and `.xterm-viewport`, but **xterm v6 replaced webkit scrollbars with a Monaco-style custom widget** (`.xterm-scrollable-element > .scrollbar`). Our existing webkit-scrollbar styles are dead code on v6. The cosmetic gap is a known follow-up; the thaw fix in this PR is the substantive change.

The Inspect Element feature is a development tool that should prevent the same dead-end debug cycle next time — five-second DOM inspection beats three SCSS attempts.

Reagent + codex will re-review on the rewritten branch.

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #1047, opened 2026-05-26
feat(warden): LAN section live mDNS peer list (PR-E)

AgentY-asaf:
> **Phase 3 of the Warden spec — L2 read-only.** The LAN section now reads through to mDNS and renders the live peer list. Cross-instance enforcement still pending lan-awareness Phase 3.
>
> Spec: \`specs/SPEC_WARDEN_WIDGET_2026-05-25.md\`
> Stack: PR-E (this) follows PR-C (#1046 merged), PR-B (#1045 merged), PR-A (#1044 merged) · spec landed in #1032

## What this PR adds

- \`fetchLanPeers()\` calling \`GET /api/lan-instances\` (existing public endpoint, shipped via mDNS in #1032)
- \`LanSection\` component polling every 5 s, rendering: \`peer · version · address · agents · last seen\`
- Same 1 s clock tick approach for age columns
- LAN status chip flips \`stub\` → \`live\`
- Empty-state message points the operator to the HostPopover toggle if they haven't enabled mDNS yet
- Footnote notes that cross-instance jekt / quarantine controls land later

## Backend changes

None. mDNS \`LanDiscovery\` daemon and the \`/api/lan-instances\` route are already on main.

## Verification

- \`npx tsc --noEmit\` — no errors in warden files
- Smoke test: enable \"LAN discovery\" on two AgentMux instances on the same network → both should see the other in the Warden's LAN section within ~5 s

PR #1046, opened 2026-05-26
feat(warden): host section recent jekt audit feed (PR-C)

AgentY-asaf:
> **Finishes Host L1 read-only.** Adds the recent jekt audit feed below the agent table in the Warden's Host section.
>
> Spec: \`specs/SPEC_WARDEN_WIDGET_2026-05-25.md\`
> Stack: PR-C (this) · follows PR-B (#1045 merged) · follows PR-A (#1044 merged)

## What this PR adds

- New \`AuditEntry\` row type + \`fetchHostAudit()\` (reuses \`GET /agentmux/reactive/audit?limit=50\`, already gated by auth middleware)
- Audit feed under the agent table: \`timestamp · source → target · ok/err · bytes · inline error\`
- Same 5 s refresh tick fetches both agents and audit (Promise.all); 1 s clock tick re-renders the age column without re-fetching
- Failed rows get a subtle red background; error message tail truncates with ellipsis

## Backend changes

None. Reuses the existing reactive-audit endpoint.

## Out of scope

| Capability | Lands in |
|---|---|
| Kill / pause agent | PR-D |
| Identity provenance per agent | PR-D or later |
| Capability set per agent | PR-D |
| LAN peer list | PR-E (mDNS substrate is already live since #1032) |

## Verification

- \`npx tsc --noEmit\` — no errors in warden files
- Existing endpoint already shipped; smoke test = open Warden + spawn an agent + trigger a jekt

PR #1045, opened 2026-05-25
feat(warden): host section live agent list (PR-B of Warden rollout)

AgentY-asaf:
> **Phase 1 of the Warden spec — Host L1 read-only, minimum viable slice.** The Host section now shows the actual agents registered on this AgentMux instance. LAN and Internet still stubbed.
>
> **Spec:** [\`specs/SPEC_WARDEN_WIDGET_2026-05-25.md\`](https://github.com/agentmuxai/agentmux/blob/main/specs/SPEC_WARDEN_WIDGET_2026-05-25.md)
> **Stack:** PR-B (this) follows PR-A (#1044, shell — merged)

## What this PR does

- Host section in \`WardenView\` fetches \`GET /agentmux/reactive/agents\` on mount + every 5 s
- Renders a table: \`agent\`, \`block\` (short hash), \`last seen\` (age), \`state\` (active < 30 s, else idle)
- Separate 1-s clock tick updates the age column without re-fetching
- Host section's status chip flips from \`stub\` → \`live\`
- LAN and Internet sections unchanged (still stubs/disabled)

## Out of scope (each its own PR)

| Capability | Lands in |
|---|---|
| Identity provenance per agent | PR-C |
| Capability set per agent | PR-D |
| jekt/min counter | needs new backend metric |
| Audit event stream | PR-C |
| Kill / pause buttons | PR-D (governance enforcement) |
| LAN peer list | PR-E (after lan-awareness Phase 3) |

## Backend changes

**None.** Reuses the existing \`/agentmux/reactive/agents\` HTTP endpoint, already gated by \`auth_middleware\`. A dedicated \`GET /agentmux/warden/snapshot\` endpoint can come when LAN and Internet layers need wrapping into a single snapshot shape — keeping this PR small.

## Verification

- \`npx tsc --noEmit\` — no errors in any file this PR touches
- Smoke test: open Warden widget, Host section should populate with whatever agents are registered

## Manual smoke test

1. \`task dev\`
2. Spawn an agent pane (any kind — the agent registers via shell hooks)
3. Open the Warden widget
4. Host section shows that agent, status \`active\`, age counting up in real time
5. Close the agent pane → after ~30 s the row's state chip turns \`idle\`
6. Force-quit the agent (kill pid) → row eventually drops on next 5 s refresh

PR #1044, opened 2026-05-25
feat(warden): add widget shell with three-section scaffold (Phase 1 of warden spec)

AgentY-asaf:
> **First step of the Warden widget.** Just the shell — registers the widget, renders three sections (Host / LAN / Internet) with phase-status stubs. No backend, no data, no enforcement. Each section announces what it'll fill in and when.
>
> **Spec:** [\`specs/SPEC_WARDEN_WIDGET_2026-05-25.md\`](https://github.com/agentmuxai/agentmux/blob/main/specs/SPEC_WARDEN_WIDGET_2026-05-25.md) (landed via #1032)
> **RFC:** #1033

## What this PR adds

- \`defwidget@warden\` in \`agentmux-srv/src/config/widgets.json\` — pinned, \`shield-halved\` icon, order 10
- \`frontend/app/view/warden/warden.tsx\` — \`WardenViewModel\` (minimal, matches the HelpViewModel pattern) + \`WardenView\` that renders three layer sections
- \`frontend/app/view/warden/warden.scss\` — section layout + status chip styling
- \`BlockRegistry.set("warden", WardenViewModel)\` in \`frontend/app/block/block.tsx\`
- \`CLAUDE.md\` widget table updated

## What's *not* in this PR

| Phase | Goes in PR | Depends on |
|---|---|---|
| L1 read-only (agent list, identity, jekt/min, audit stream) | PR-B | nothing (just backend RPCs) |
| L1 enforcement (kill agent, pause, kill-all) | PR-C | \`governance.json\` schema + enforcer hooks |
| L2 read-only (LAN peer list with drift detection) | PR-D | lan-awareness Phase 3 (LAN forwarding) |
| L2 enforcement (quarantine, policy push) | PR-E | embedded MCP SSE server (Phase 2) |
| L3 (Internet) read + control | PR-F | cloud fallback (Phase 4) |

Each section in the shell renders a \`status\` chip (\`live\` / \`stub\` / \`disabled\`) so the operator can see at a glance which layers are functional. Internet is \`disabled\` by default per the spec's "closed by default at boundary 3" principle.

## Verification

- \`npx tsc --noEmit\` — no errors in any file this PR touches
- JSON validates: \`agentmux-srv/src/config/widgets.json\` parses cleanly

## Manual smoke test

1. \`task dev\`
2. New block → click **warden** in the widget bar
3. Widget opens with three labeled sections, each showing its status chip and a one-line "coming soon" explaining what fills in
[...]

PR #1048, opened 2026-05-26
feat(warden): host soft-deregister button (PR-D)

AgentY-asaf:
> **First slice of L1 enforcement.** Each agent row in the Warden's Host section gets a × button that does a *soft* deregister — removes the agent's entry from the reactive registry so future jekts return "agent not found", but does NOT kill the underlying process.
>
> Spec: \`specs/SPEC_WARDEN_WIDGET_2026-05-25.md\`
> Stack: PR-D (this) follows PR-E (#1047 merged), PR-C (#1046 merged), PR-B (#1045 merged), PR-A (#1044 merged)

## Why soft

Hard kill (PTY termination) is out of scope here:
- Process lifecycle is owned by the block controller, not the reactive routing layer
- A "deregister" is reversible (the agent's shell auto-register hook restores routing on next heartbeat)
- The blast-radius is well-defined: routing only

The confirmation dialog spells this out so the operator isn't surprised when the row reappears 30 s later.

## What this PR adds

- \`deregisterAgent(agentId)\` calling \`POST /agentmux/reactive/unregister\` with the auth header
- \`handleDeregister\` in \`HostSection\` — \`window.confirm\` first, then POST, then immediate refresh
- × button per row (\`warden-host-deregister\`) with hover state turning red
- Tooltip clarifies "removes from jekt routing, leaves process running"

## Backend changes

**None.** The reactive-unregister endpoint already exists.

## Out of scope (each its own PR)

| Capability | When |
|---|---|
| Hard kill (PTY termination) | Future PR — needs block-controller hook |
| Pause host (stop all inbound jekts) | Future PR |
| Kill-all (emergency stop) | Future PR |
| \`governance.json\` policy file | Future PR |
| Approval queue / human-in-the-loop | Future PR |
| Capability revocation per agent | Future PR |

## Verification

- \`npx tsc --noEmit\` — no errors in warden files
- Smoke test: spawn an agent → see it in Warden Host → click × → confirm → row drops on next 5 s poll; if shell heartbeat re-registers, row reappears

### DIFF
--- a/.changesets/1779756482-feat-warden-soft-deregister-agent-button.md
+++ b/.changesets/1779756482-feat-warden-soft-deregister-agent-button.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(warden): host section adds soft-deregister button per agent
--- a/frontend/app/view/warden/warden.scss
+++ b/frontend/app/view/warden/warden.scss
@@ -159,6 +159,33 @@
     }
 }
 
+.warden-host-actions {
+    text-align: right;
+    width: 24px;
+}
+
+.warden-host-deregister {
+    background: transparent;
+    border: 1px solid transparent;
+    color: var(--main-text-color);
+    opacity: 0.4;
+    cursor: pointer;
+    width: 20px;
+    height: 20px;
+    border-radius: 3px;
+    font-size: 1em;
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1053, opened 2026-05-26
feat(common): per-clone Dev isolation — RuntimeMode::Dev gains clone_id

AgentX-asaf:
## Summary

Lets two clones of the same agentmux repo on the same branch run `task dev` simultaneously without colliding on data dir, lockfile, named-pipe IPC, CEF cache, or logs.

## Root cause

The launcher's single-instance lockfile name and named-pipe path are both derived from `data_dir_hash16(instance_dir)`. Two clones at different filesystem paths both on `main` resolve to the identical `~/.agentmux/dev/main/` instance_dir, hash to the same lockfile and pipe namespace, and the second clone's launcher silently routes its open-pane intent into the first clone's already-running window — no error, just cross-contamination.

Full inventory in [`docs/analyses/ANALYSIS_MULTI_CLONE_TASK_DEV_ISOLATION_2026-05-26.md`](https://github.com/agentmuxai/agentmux/blob/agentx/multi-clone-dev-isolation/docs/analyses/ANALYSIS_MULTI_CLONE_TASK_DEV_ISOLATION_2026-05-26.md).

## Fix

Extend `RuntimeMode::Dev` from `{ branch }` to `{ branch, clone_id: Option<String> }` and nest dev paths at `~/.agentmux/dev/<branch>/<clone_id>/`. `clone_id` is FNV-1a-16-hex of the clone's workspace-root path (first ancestor of the exe dir containing `.git`, `Cargo.toml`, or `Taskfile.yml`). Lockfile + named-pipe + CEF cache + logs all flow from `instance_dir`, so the rest is automatic.

## Implementation notes

- New `derive_clone_id(exe_dir)` in `agentmux-common/src/runtime_mode.rs`. Inline FNV-1a-64 (matches `agentmux-launcher/src/hash.rs` byte-for-byte; avoids cross-crate dep).
- `RuntimeMode::current` / `current_path_only` populate `clone_id` on every Dev classification path.
- `DataPaths::resolve_channel_and_dir` nests under `<clone_id>/` when `Some`; **legacy two-level `dev/<branch>/` when `None`** for back-compat (env-string round-trip path, direct test construction).
- New `AGENTMUX_CLONE_ID` env var; `DataPaths::to_env_vars` exports it for Dev with `Some(clone_id)`; new `RuntimeMode::from_env_with_clone()` pairs `AGENTMUX_RUNTIME_MODE` + `AGENTMUX_CLONE_ID` for full reconstruction in child processes.
- `to_env_string()` continues to emit `dev:<branch>` only — keeping the wire format backward-compatible with older parsers; `clone_id` travels via its dedicated env var.

## Tests

All 75 `agentmux-common` tests pass; full `cargo build --workspace` is clean. New tests:

- `derive_clone_id_*` — stable for same path; differs for different paths; walks up to find marker; returns None when no marker.
- `sanitize_clone_id_rejects_traversal` — env-supplied clone_ids can't escape the dev/ subtree.
- `dev_paths_under_branch_and_clone_id` — two Dev variants on the same branch but different clone_ids resolve to distinct instance dirs and channel names.
- `dev_paths_legacy_two_level_when_clone_id_none` — explicit back-compat assertion (no migration of in-flight dev sessions).
- `from_env_with_clone_*` — AGENTMUX_CLONE_ID round-trips through env for Dev.

## Out of scope (deferred)

- **Vite port 5173.** Stays hardcoded. `--strictPort` already fails loud on the second clone; that's an acceptable degraded experience for now. Real fix is a `AGENTMUX_VITE_PORT` override and a Taskfile knob — small but separate PR.
- **Cargo `target/` dedup.** Significant disk cost per clone but functionally isolated. Out of scope.
- **VS Code Bridge.** External tool; out of scope here.

## Test plan

- [x] `cargo test -p agentmux-common` — 75/75 pass
- [x] `cargo build --workspace` — clean
- [ ] Two clones on the same branch each run `task dev`. Both windows open, both have distinct data dirs at `~/.agentmux/dev/<branch>/<clone-hash>/`, opening a pane in clone A's window does not affect clone B's window.
- [ ] One clone on `main`, one on a feature branch — pre-existing branch-keyed behavior still works; both isolate.
- [ ] Existing single-clone dev sessions (no `AGENTMUX_CLONE_ID` set yet) still launch into the legacy two-level `dev/<branch>/` path on the first launch after the upgrade.

### DIFF
--- a/.changesets/1779785318-feat-common-per-clone-dev-isolation-runtimemode-dev-now-carr-9ow5.md
+++ b/.changesets/1779785318-feat-common-per-clone-dev-isolation-runtimemode-dev-now-carr-9ow5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(common): per-clone Dev isolation — RuntimeMode::Dev now carries clone_id for two-clones-same-branch parity
--- a/agentmux-common/src/data_paths.rs
+++ b/agentmux-common/src/data_paths.rs
@@ -243,7 +243,7 @@ impl DataPaths {
     /// (it's a fixed ASCII vocabulary).
     pub fn to_env_vars(&self) -> Vec<(&'static str, std::ffi::OsString)> {
         use std::ffi::OsString;
-        vec![
+        let mut vars: Vec<(&'static str, OsString)> = vec![
             ("AGENTMUX_INSTANCE_DIR", self.instance_dir.clone().into_os_string()),
             ("AGENTMUX_DATA_DIR", self.data_dir.clone().into_os_string()),
             ("AGENTMUX_CONFIG_DIR", self.config_dir.clone().into_os_string()),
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1039, opened 2026-05-25
feat(modal): compact variant for narrow lock regions (auto-trigger via ResizeObserver)

AgentA-asaf:
**Spec ships with the PR**: [`docs/specs/SPEC_MODAL_COMPACT_VARIANT_2026_05_25.md`](../blob/agenta/compact-modal-variant/docs/specs/SPEC_MODAL_COMPACT_VARIANT_2026_05_25.md).

## Problem

`<Modal scope=\"pane\">` can mount in arbitrarily narrow panes — verified live during the auth-modal investigation: a focused browser pane sat at **240×1020** in a 630-wide window with three sibling panes. Every existing modal panel min-width (320-560px) overflowed badly at that width: panel covered neighbors despite the pane-scope lock, or `size=\"fit\"` clamped and produced truncated button rows and cropped titles.

## Approach — automatic + CSS-driven

| | Standard (≥400px) | Compact (<400px) |
|---|---|---|
| `.modal-panel` | per-panel min-width (320-560px) | `min-width: 0; width: 100%; max-width: 100%` |
| Header padding | `var(--space-3)` | `var(--space-1) var(--space-2)` |
| Title | `font-size: 18px` | `font-size: 14px, line-height: 1.3` |
| Description | `font-size: 13px` | `font-size: 12px` |
| Body padding | `var(--space-3)` | `var(--space-2)` |
| Footer | `row, right-aligned, natural-width buttons` | `column-reverse, width:100% buttons` (submit at bottom, mobile-app convention) |

**Trigger**: `ResizeObserver` on `.modal-layer-mount` toggles a `.modal-layer-mount--compact` class when `contentRect.width < COMPACT_THRESHOLD_PX` (400). All canonical chrome (`.modal-panel-*`) gets the variant for free via a `:where(.modal-layer-mount--compact)` ruleset in `modal.scss`. Per-panel SCSS opts in via `:where(.modal-layer-mount--compact) &` — `:where()` keeps specificity at 0 so existing per-panel rules still override without `!important`.

## Threshold (400px)

- 240px is the verified-narrow case (the browser pane scenario above).
- 400px gives ~30% headroom for the user dragging the divider — avoids flapping at the threshold.
- Above 400px every existing modal panel renders cleanly. Single threshold, no `--ultra-compact` stack — adding granularity later is cheaper than collapsing it.

## Files

- `frontend/app/element/ModalLayer.tsx` — `ResizeObserver` + class toggle. Handles HMR re-mount (disconnects prior observer before attaching).
[...]

PR #1056, opened 2026-05-26
fix(modal): add compact-mode min-width override to 4 modals (audit follow-up to #1039)

AgentA-asaf:
## Summary

`SPEC_MODAL_COMPACT_VARIANT_2026_05_25.md` §3 documented the compact
escape-hatch pattern, and #1039 applied it to the **install modal**
only. An audit of all modal bodies turned up 4 more with hard-coded
min-widths that overflow in narrow panes.

## Affected modals

| Modal | min-width | Status before | Status after |
|---|---|---|---|
| AgentInstallModal | 560px | ✅ has override (#1039) | unchanged |
| AgentLaunchModalPanel | 420px | ❌ missing | ✅ fixed |
| AgentNewBundleModal (Identity & Memory) | 440px | ❌ missing | ✅ fixed |
| AgentPrereqModal | 440px | ❌ missing | ✅ fixed |
| BrowserAuthModal | 380px | ❌ missing | ✅ fixed |

The compact threshold is 400px (`COMPACT_THRESHOLD_PX` in
`ModalLayer.tsx`), so all 4 unfixed modals can overflow when the parent
pane is narrower than the base min-width. The browser-auth modal in
particular mounts pane-scoped (#1038), and browser panes go down to
~240px in multi-pane layouts (verified in the #1037/#1038 work).

## Fix

Mirror the install modal pattern exactly — append a
`:where(.modal-layer-mount--compact) & { min-width: 0 }` block to each
of the 4 modal bodies. Cross-reference the spec in each comment so a
future modal audit can grep for `modal-layer-mount--compact` and find
the canonical pattern.

## Why now

User flagged that the modal in a narrow pane still ran wide. Audit
confirmed the pattern wasn't propagated to siblings in #1039. This
PR closes that gap.

## Test plan

- [ ] Manual smoke: at a narrow browser pane (<400px wide), open
      the HTTP-auth modal — should now collapse to compact layout
- [ ] Same for narrow agent pane: launch/prereq/new-bundle modals
- [x] Mechanical sweep complete — `grep -rln "modal-layer-mount--compact"
      frontend --include "*.scss"` now finds 6 hits (was 2 before)

Changeset: `patch` (RFC #857 Phase 2).

### DIFF
--- a/.changesets/1779790254-fix-modal-add-compact-mode-min-width-override-to-launch-prer-shew.md
+++ b/.changesets/1779790254-fix-modal-add-compact-mode-min-width-override-to-launch-prer-shew.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(modal): add compact-mode min-width override to launch/prereq/new-bundle/browser-auth modals
--- a/frontend/app/view/agent/components/AgentNewBundleModal.scss
+++ b/frontend/app/view/agent/components/AgentNewBundleModal.scss
@@ -12,6 +12,13 @@
     gap: var(--space-3);
     min-width: 440px;
     max-width: 560px;
+
+    // Compact variant — relax the body min-width so the modal can mount
+    // in panes narrower than 440px. The ModalLayer toggles this class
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1059, opened 2026-05-26
fix(install-modal): unshrinkable xterm in compact panes

AgentA-asaf:
The install modal was visually staying wide with a horizontal scrollbar even with the `:where(.modal-layer-mount--compact)` body min-width override applied (added in PR #1039, surfaced as still broken when smoke-testing v0.38.10).

## Cause

Classic flex-shrink trap: the flex child `.agent-install-modal-term` defaults to `min-width: auto` (intrinsic content size). xterm.js renders at its default 80 cols (~600px canvas), and the outer `.modal-panel { overflow: auto }` exposes the overflow as a horizontal scrollbar. Modal appears wide and won't shrink in narrow panes.

## Fix

Add `min-width: 0` to `.agent-install-modal-term` inside the compact override so the flex child can shrink below its intrinsic content size. FitAddon then resizes xterm to fit on its first ResizeObserver tick (same path that already governs xterm sizing — no new wiring).

The body's `min-height: 200px` override added in #1039 already addressed the vertical version of this trap. This completes the pair.

## Verification

- `task build:frontend` clean (✓ 22.92s)
- Will validate live by opening an install modal in a narrow pane after the next portable build

### DIFF
--- a/.changesets/1779813541-fix-install-modal-allow-xterm-container-to-shrink-in-compact-wmja.md
+++ b/.changesets/1779813541-fix-install-modal-allow-xterm-container-to-shrink-in-compact-wmja.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(install-modal): allow xterm container to shrink in compact panes (flex-shrink trap)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

AgentA-asaf:
Superseded by Phase 1 of `docs/analysis/MODAL_COMPACT_VARIANT_ARCHITECTURE_2026_05_26.md`. The flex-shrink fix here was necessary but not sufficient — three independent failure modes remained (xterm canvas pixel-width, `size="fit"` content sizing, boot timing race). Closing in favor of the broader redesign which addresses all six failure modes structurally and reclaims the 48px of wasted edge padding.

PR #1064, opened 2026-05-26
feat(editor): file-tree explorer rooted at $HOME + header chevron + toolbar

AgentY-asaf:
> Phase 1 of [`SPEC_EDITOR_FILE_TREE_2026-05-26.md`](../blob/main/specs/SPEC_EDITOR_FILE_TREE_2026-05-26.md). Adds a collapsible file-tree explorer down the left of the editor pane, rooted at the OS home dir, with a header chevron toggle (default expanded), click-to-load, and a small toolbar.

### Backend (`agentmux-srv/src/server/websocket.rs`)

- **`listeditordir`** — list a directory's entries. Follows symlinks for `is_dir`/`size` (matches VS Code semantics) but reports `is_symlink` separately so the frontend can mark followed symlinks with `↗`. Folders first, alphabetical case-insensitive.
- **`geteditorhome`** — returns `dirs::home_dir()` so the frontend has the canonical root.

### Frontend

- **`file-tree-model.ts`** — state machine. `Map<path, NodeData>` caches loaded children; `Set<path>` tracks expanded folders. Lazy-load on first expand; collapse keeps the cache so re-expand is instant. Refresh re-fetches every expanded path. Collapse-all returns to root-only.
- **`file-tree.tsx`** — recursive tree + toolbar (3 buttons: `eye` show-hidden toggle, `square-minus` collapse-all, `arrows-rotate` refresh). Active file row carries the `circle-dot` icon; symlinked rows carry a `↗` overlay. Hidden-file filter applied at render time (dotfiles + `node_modules` + `.DS_Store` + `Thumbs.db`).
- **`editor-model.ts`** — adds `treeExpandedAtom` + `showHiddenAtom` (per-pane, persisted to block meta as `editor:tree_expanded` / `editor:show_hidden`). Header chevron toggle in the `viewText` HeaderElem slot.
- **`editor-view.tsx`** — two-column layout; empty-state path-input moves under the tree when no file is open.
- **`editor-view.scss`** — tree column, toolbar, scoped `data-tip` CSS rules for **instant tooltips** (pure `:hover::after`, no JS delay), row hover/active states, depth-aware indent.
- **`custom.d.ts`** — `DirEntry` ambient type.

## Decisions from spec discussion (already locked in)

| | |
|---|---|
| Tree column persistence | Per-pane (block meta) |
| Hidden files default | Hidden (dotfiles + `node_modules` + `.DS_Store` + `Thumbs.db`) |
| Symlinks | Followed, marked with `↗` overlay |
| Tree refresh | Explicit only — toolbar button |
| Tooltip delay | **0ms** via `data-tip` CSS pattern |

## Out of scope (future phases)

Phase 2 polish (resizable divider, arrow-key nav), Phase 3 extensions (additional curated CodeMirror extensions), Phase 4 stretch (LSP, multi-root, file watching). All listed in the spec.

## Verification

- `cargo check -p agentmux-srv` — passes
- `npx tsc --noEmit` — no errors in any file this PR touches
- Manual: spec was reviewed with the user before implementation; UX matches the decided shape.

### DIFF
--- a/.changesets/1779817687-feat-editor-file-tree-explorer-toolbar-header-chevron.md
+++ b/.changesets/1779817687-feat-editor-file-tree-explorer-toolbar-header-chevron.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(editor): file-tree explorer rooted at $HOME with header chevron + toolbar (show hidden / collapse all / refresh)
--- a/agentmux-srv/src/server/websocket.rs
+++ b/agentmux-srv/src/server/websocket.rs
@@ -1141,6 +1141,97 @@ fn register_handlers(engine: &Arc<WshRpcEngine>, state: AppState) {
         }),
     );
 
+    // listeditordir → list directory contents for the editor's file-tree pane.
+    // Symlinks are followed (matches VS Code semantics; the frontend marks
+    // followed symlinks with a ↗ overlay).
+    // Spec: specs/SPEC_EDITOR_FILE_TREE_2026-05-26.md
+    engine.register_handler(
+        "listeditordir",
+        Box::new(|data, _ctx| {
+            Box::pin(async move {
+                #[derive(serde::Deserialize)]
+                struct Cmd { path: String }
+                let cmd: Cmd = serde_json::from_value(data)
+                    .map_err(|e| format!("listeditordir: {e}"))?;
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #1067, opened 2026-05-26
feat(editor): multi-root file tree + resize handle between tree and editor

AgentY-asaf:
> Two follow-ups to #1064 the user surfaced while looking at the live build.

### 1. Multi-root tree (HOME + drives + mounts)

The tree was stuck at \`\$HOME\`. Now it shows HOME as the auto-expanded primary root, plus every drive/mount as a sibling root (collapsed). User can navigate anywhere on the system without leaving the tree.

- Backend (\`websocket.rs\`):
  - New \`geteditorroots\` RPC returning \`{ home, drives[] }\`
  - New \`list_drives()\` helper with \`#[cfg]\` gates — Windows enumerates A–Z; Unix includes \`/\` plus mounts under \`/mnt\`, \`/media\`, \`/Volumes\`
- Frontend (\`file-tree-model.ts\`):
  - \`rootsAtom: Root[]\` replaces single \`rootAtom\`
  - The drive hosting HOME is de-duplicated so it doesn't appear twice (no \`C:\\` row when HOME is \`C:\Users\…\`)
- Frontend (\`file-tree.tsx\`): top-level renders \`<For each={rootsAtom()}>\` producing one row per filesystem root

### 2. Draggable resize handle

The boundary between tree and editor was a static 240px. Now drag it.

- 4px-wide col-resize strip between tree column and main editor column
- Live updates the tree-width signal during drag (no RPC on every pixel)
- Persists on mouseup via \`editor:tree_width\` in block meta (per-pane)
- Min 150px, max 600px, default 240px
- Drag handlers attach to \`document\` (mousemove/mouseup) so the user can drag past the strip without losing grip; body cursor + user-select locked during drag

## Source of decisions

Both fall under the spec's Phase 2 (\`specs/SPEC_EDITOR_FILE_TREE_2026-05-26.md\` § Implementation phases → Phase 2). User asked for them after seeing the Phase 1 build, so pulling them forward in a tight follow-up rather than batching with later phases.

## Verification

- \`cargo check -p agentmux-srv\` — passes
- \`npx tsc --noEmit\` — no errors in any file this PR touches

### DIFF
--- a/.changesets/1779818795-feat-editor-multi-root-tree-plus-resize-handle.md
+++ b/.changesets/1779818795-feat-editor-multi-root-tree-plus-resize-handle.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(editor): multi-root file tree (HOME + drives/mounts) and draggable resize handle between tree and editor
--- a/agentmux-srv/src/server/websocket.rs
+++ b/agentmux-srv/src/server/websocket.rs
@@ -35,6 +35,44 @@ use super::service::update_object_meta;
 
 use super::AppState;
 
+/// Enumerate drives/mounts accessible from the current OS user.
+/// Used by the `geteditorroots` RPC so the editor file-tree exposes
+/// every reachable filesystem root, not just $HOME.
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #1070, opened 2026-05-26
feat(editor): icon toggles tree, title shows full file path

AgentY-asaf:
> Two small UX refinements following live use of the new editor file-tree.

## What changes

| Before | After |
|---|---|
| Pane icon was a fixed \`file-code\` glyph | Pane icon is the **file-tree toggle** — \`folder-tree\` when expanded, \`folder\` when collapsed |
| Tree toggle lived as a separate chevron HeaderElem on the right | Removed (icon owns it now) |
| Title showed just the basename (\`main.ts *\`) | Title shows the **full path** (\`/Users/asaf/repo/src/main.ts *\`) |

Why the full path: with two editor panes open on similarly-named files (e.g. two `index.ts` in different dirs), it was easy to lose track. The block frame's existing overflow handling keeps the layout sane.

## Implementation

- `viewIcon` widened from `Accessor<string>` to `Accessor<string | IconButtonDecl>`. The block frame already supports both shapes — see `blockframe.tsx:getViewIconElem`.
- `viewText` returns `[]` since the icon owns the toggle; keeps the right side of the header tidy and lets future header items land there without competing for attention.
- Same `editor:tree_expanded` block-meta key — no migration needed.

## Verification

`npx tsc --noEmit` — clean on every file this PR touches.

### DIFF
--- a/.changesets/1779821048-feat-editor-icon-toggles-tree-and-title-shows-full-path.md
+++ b/.changesets/1779821048-feat-editor-icon-toggles-tree-and-title-shows-full-path.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(editor): pane icon toggles the file-tree (replacing the separate chevron); title bar now shows the full file path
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1069, opened 2026-05-26
feat(agent-pane): slim composer status strip replaces AgentStatusLine

AgentA-asaf:
**PR B of 2** — composer slim-status redesign view layer. Follows #1068 (reducer slice).

Spec: \`docs/specs/SPEC_AGENT_COMPOSER_SLIM_STATUS_2026_05_26.md\`.

## User-facing change

The wide multi-band composer area (AgentStatusLine + always-visible ActivityLogPanel + always-visible AgentControlBar) collapses to a single 28-32px slim strip above the textarea:

- **LEFT** ◐ spinner + latest activity-log line (truncated, single-line ticker)
- **RIGHT** ↑in ↓out · elapsed · ⚙N · permission pill (only when non-default)
- **▾³** chevron with unread badge — click expands the details panel

The details panel (when expanded) shows the activity log + control bar in their existing form. Send-message auto-collapses. Idle pane: strip is essentially empty (just the chevron) — quiet UI.

## Implementation

- New view atoms projecting from #1068's reducer state: \`detailsOpenAtom\`, \`composerUnreadCountAtom\`.
- New \`AgentComposerStrip\` component (\`frontend/app/view/agent/components/AgentComposerStrip.tsx\`) reads the atoms, dispatches \`DetailsToggle\` on click.
- \`agent-view.tsx\` replaces \`<AgentStatusLine>\` with \`<AgentComposerStrip>\` and gates \`<ActivityLogPanel>\` + \`<AgentControlBar>\` on \`detailsOpenAtom()\`.
- \`useActivityLog\` dispatch wired via \`createEffect\` on \`logLines\` length — each growth fires \`LogEntryArrived\`; reducer increments the unread counter when the panel is closed.

## What's deferred

\`AgentControlBar\` and \`AgentStatusLine\` are NOT deleted in this PR — they're folded into the conditional render. A follow-up PR will consolidate them into a single \`AgentComposerDetails\` component matching spec §4 (Row 1 session summary, Row 2 runtime config, Row 3 tools-used, Row 4 buttons).

## State integration

All composer state lives in the per-pane reducer slice (\`AgentPaneState.detailsOpen\` / \`composerUnreadCount\`, added in #1068). No component-local Solid signals, no parallel state machine. Matches the Phase B \"single source of truth\" contract.

## Tests

- 137/137 reducer tests pass (the 12 added in #1068 cover the state side)
- No TS errors in touched files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1072, opened 2026-05-26
fix(composer-strip): correct SCSS @use mixins path (hotfix for #1069 build)

AgentA-asaf:
Hotfix for build break introduced by #1069.

`_composer-strip.scss` had `@use "../../mixins.scss"` which from `styles/` resolves to `view/mixins.scss` (doesn't exist). Real path is `frontend/app/mixins.scss` — 3 levels up from styles/. One-character fix.

Broke `task build:frontend` and `task package`. Caught by the v0.38.14 portable rebuild attempt right after #1069 merged.


a5af:
@codex review

Issue #810, opened 2026-05-12
Spec: floating pane tear-off (subordinate window, owned by mother instance)


AgentA-asaf:
# Floating pane tear-off (subordinate window, owned by mother instance)

**Status:** Proposed
**Owner:** AgentA
**Date:** 2026-05-11
**Driving observation:** *"When tearing off panes, we don't create a new instance (no new taskbar icon). Instead it's a floating window owned by the mother instance. Panes currently tear off and create a new instance; instead they would tear to subpanels."*

Today, tearing a tab off creates a new full AgentMux instance — own backend sidecar, own data dir, own taskbar entry. That's the right model for tabs (workspace-level isolation). For **panes** within a tab, the new ask is different: tear into a **floating subordinate window** owned by the source instance, like Photoshop's palettes or VSCode's "Move Editor into New Window".

This spec covers panes specifically. Tab tear-off behavior is unchanged.

---

## 1. Target behavior

When the user drags a pane out of its tile layout and drops it outside the source window:

- A floating window opens at the cursor.
- The floating window has **no taskbar entry**, **no Alt-Tab entry**.
- The floating window **minimizes/restores with the source window** automatically.
- Closing the source window **destroys all its floating panes**.
- The floating pane shares **the source instance's backend sidecar, data dir, reducer state**. There is no new sidecar process, no new data dir, no new browser-pane HWND pool.
- The pane can be dragged from its titlebar to any position on any monitor (not clipped to the source window's bounds).
- A "dock back" affordance: drag the floater into the source window's layout, or use a button on its title bar.

The user gets a free-floating pane that's a full peer of the docked panes (same state, same broadcasts, same drag-drop into other tiles).

---

## 2. Why owned, not standalone

The current "new instance" path is overkill for panes:

| Cost | New full instance (today) | Owned floating (proposed) |
|---|---|---|
| Backend sidecar | Spawns a new one (~150-300 ms cold start) | Reuses the parent's |
| Data directory | New `~/.agentmux/versions/<v>/` dir | Shares parent's |
| Taskbar entry | One per torn-off pane | None |
| Cross-pane state | Requires inter-instance IPC | Shared in-process |
| Persistence | Per-instance layout file | Single layout file with `isFloating` flag |
| Resource overhead | ×N sidecars, ×N data dirs, ×N CEF process pools | One of each |
| Crash blast radius | Per-instance isolation | Floaters die with parent (acceptable for palettes) |

[...]

PR #1073, opened 2026-05-26
feat(floating-pane): Phase 3 — pane tear-off routes to floating child window

AgentX-asaf:
## Summary

User-visible behavior change: **dragging a pane out of the window no longer spawns a brand-new top-level AgentMux instance** (own taskbar entry, own sidecar, own data dir). It now spawns an owned floating child window — no taskbar, no Alt-Tab, minimizes/restores with the parent, shares the same sidecar.

Tab tear-off is unchanged (still spawns a new instance — that's the right model for workspaces).

This wires Phase 3 of [`SPEC_FLOATING_PANE_TEAROFF_2026_05_11.md`](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/SPEC_FLOATING_PANE_TEAROFF_2026_05_11.md) on top of the Phase 1 host primitive from #810.

### Routing — `CrossWindowDragMonitor.win32.tsx::performTearOff`

The `pane` branch (when `dragType === "pane"`) no longer calls `WorkspaceService.TearOffBlock` + `openTearOffWindow`. It now:

1. Removes the block's node from the source window's tile layout via `LayoutTreeActionType.DeleteNode` (so the docked pane doesn't render twice — once docked, once floating).
2. Calls the existing `open_floating_pane_window` IPC (Phase 1 host primitive) at the drop cursor position.

The block stays in the source workspace's `blockids`; only its visual location moves. The `tab` branch is left as-is.

### Shell — `floating-pane-shell.tsx`

Phase 1 was a placeholder card. This PR adds:
- A draggable title bar — CSS `-webkit-app-region: drag`, which Chromium maps to `WM_NCHITTEST` returning `HTCAPTION` on Windows so the OS-level drag loop kicks in.
- A Close button wired to the existing `closeWindowByLabel` IPC.

Pane content is still placeholder text. **PR-A2 (Phase 2) replaces it with the real `<Block>` renderer.** That's a separate PR because it requires a minimal RPC/WOS init in the floating window's process (subset of `initApp()`), which is too big to ship in the same diff as the routing change.

### Docs

- `docs/analyses/ANALYSIS_FLOATING_PANE_TEAROFF_STATE_2026-05-26.md` — what's shipped vs. needed; resolves the open questions from earlier in the session.
- `docs/specs/SPEC_FLOATING_PANE_TEAROFF_CROSS_PLATFORM_2026-05-26.md` — macOS (`NSPanel.addChildWindow`) and Linux X11/GTK (`set_transient_for` + utility + skip-taskbar) recipes that the parent spec deferred. Wayland deferred until CEF Ozone-Wayland gains `SetAsChild` parity.

## Reducer integration

Yes, this plugs into the existing reducer system. The floating window runs in its own CEF process but connects to the **same** backend sidecar via `TabRpcClient` (same authKey, same IPC port). The local `DeleteNode` action on tear-off keeps the source window's tile layout consistent. A future `MarkPaneFloating` reducer action (Phase 5 of the spec — geometry persistence) will record the float position for restart restoration.

## Default floater size

Hardcoded to 720×480 in this PR. The tearoff-pane-size spec (on `agenta/feat-tearoff-match-source-size`) will refine this to the captured source-pane rect when it lands.

## Test plan

- [x] `npx tsc --noEmit` — clean (no new errors in changed files)
- [ ] Drag a pane out of the window → floating window appears at cursor with the placeholder shell + draggable title bar + Close button; no taskbar entry, not in Alt-Tab.
- [ ] Minimize source window → floater follows (OS auto-cascade via owner relationship).
- [ ] Close source window → floater destroyed (OS auto-cascade).
- [ ] Drag floater title bar → window moves smoothly via WM_NCHITTEST/HTCAPTION path.
- [ ] Click Close button → floater closes cleanly.
- [ ] Tab tear-off behavior unchanged: spawns new top-level instance with its own taskbar entry.

### DIFF
--- a/.changesets/1779823390-feat-floating-pane-phase-3-pane-tear-off-routes-to-floating--jcju.md
+++ b/.changesets/1779823390-feat-floating-pane-phase-3-pane-tear-off-routes-to-floating--jcju.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(floating-pane): Phase 3 — pane tear-off routes to floating child window, not new instance
--- a/docs/analyses/ANALYSIS_FLOATING_PANE_TEAROFF_STATE_2026-05-26.md
+++ b/docs/analyses/ANALYSIS_FLOATING_PANE_TEAROFF_STATE_2026-05-26.md
@@ -1,0 +1,246 @@
+# Analysis: Floating-pane tear-off — what's in, what's needed
+
+**Date:** 2026-05-26
+**Status:** Investigation (no code change yet)
+**Companion to:** `docs/specs/SPEC_FLOATING_PANE_TEAROFF_2026_05_11.md`
+
+## TL;DR
+
+A 6-phase spec already exists for exactly the feature requested: tear a
+pane out and have it become an owned `WS_POPUP | WS_EX_TOOLWINDOW`
+floating child of the parent window — pane content only, no taskbar
+entry, no Alt-Tab, minimizes/restores/destroys with the parent, shares
+the same sidecar.
+
+**Phase 1 of that spec is already merged** (PR #810, commit
+`944aeec2`, 2026-05-11). The Windows host primitive that creates the
+owned no-taskbar HWND + embeds CEF + redirects focus is in place. What
+ships today is a placeholder shell card; the surrounding wiring (drag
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1081, opened 2026-05-27
fix(dev): respect AGENTMUX_VITE_PORT in child window URL resolution

AgentX-asaf:
## Summary

- `resolve_frontend_base_url` (`commands/window.rs:604`) and the main-window URL fallback (`app.rs:506`) both hardcoded `http://localhost:5173` for dev mode, ignoring `AGENTMUX_VITE_PORT`.
- The main window survived because the launcher passes `--url=\$VITE_URL` on the CLI, but every child window (pool warmups, tab tear-off, floating pane) hit `ERR_CONNECTION_REFUSED` whenever Vite was on any other port.
- Both call sites now read `AGENTMUX_VITE_PORT`, falling back to 5173.

## Surfacing

Caught during smoke-testing of #1078 (floating-pane Phase 2). The test session used `AGENTMUX_VITE_PORT=5350` to dodge a collision on the auto-derived 5270; both tab and pane tear-off failed to load with `localhost:5173` `ERR_CONNECTION_REFUSED`.

This is **not a Phase 2 regression** — it has been latent in every dev session whose per-clone auto-derived Vite port differed from 5173.

## Analysis

See [`docs/analyses/ANALYSIS_DEV_VITE_PORT_HARDCODE_2026-05-26.md`](https://github.com/agentmuxai/agentmux/blob/agentx/fix-dev-vite-port-hardcode/docs/analyses/ANALYSIS_DEV_VITE_PORT_HARDCODE_2026-05-26.md) for the full call graph and evidence.

## Test plan

- [x] `cargo check -p agentmux-cef` — clean
- [ ] `AGENTMUX_VITE_PORT=5350 task dev` → tear off a pane → floating window loads (no `ERR_CONNECTION_REFUSED` in host log)
- [ ] `AGENTMUX_VITE_PORT=5350 task dev` → tear off a tab → new window loads (no `ERR_CONNECTION_REFUSED` in host log)
- [ ] Default `task dev` (port = 5173) → no change in behavior

## Out of scope

- `[on_before_close] no backend window ID registered for label=…floating… — shells may orphan` (separate floating-pane reducer-plumbing gap; tracked alongside #1079)
- Pool-exhaustion cold-path on tear-off (orthogonal; would have worked if the URL had been correct)

Issue #1079, opened 2026-05-26
feat(floating-pane): C1 + C2 — macOS NSPanel + Linux X11 GTK host primitives


AgentX-asaf:
## Status

Floating-pane tear-off on **Windows** is complete end-to-end as of #1078:
- Phase 1 host primitive (#810) — owned `WS_POPUP | WS_EX_TOOLWINDOW` HWND.
- Phase 3 drag routing (#1073) — pane drag-out spawns the floating window.
- Phase 2 real `<Block>` renderer (#1078) — actual pane content renders inside.

The same UX needs the macOS and Linux host primitives to reach feature-parity.

## Scope

Companion spec: `docs/specs/SPEC_FLOATING_PANE_TEAROFF_CROSS_PLATFORM_2026-05-26.md` has the per-platform recipes. Frontend is already platform-neutral — only the Rust host needs platform-gated implementations.

### C1 — macOS

- New module `agentmux-cef/src/floating_pane_macos.rs` (gated `#[cfg(target_os = "macos")]`).
- Subclass `NSPanel` with style mask `[.nonactivatingPanel, .titled, .resizable, .closable, .utilityWindow, .fullSizeContentView]`.
- `parentWindow.addChildWindow(panel, ordered: .above)` for the owner relationship (gives lifecycle + drag-with-parent for free).
- Set `isFloatingPanel = true`, `collectionBehavior.insert(.fullScreenAuxiliary)`.
- CEF embed via `CefWindowInfo::SetAsChild(panel.contentView, rect)` — same `set_as_child` pattern the Windows path uses.
- Wire into the `#[cfg(target_os = "windows")]` dispatcher in `agentmux-cef/src/commands/floating_pane.rs:105` so the IPC fan-out matches Windows.

Likely 350-450 LOC of Rust + objc bindings (`cocoa` / `objc2` crates already used elsewhere in the agentmux-cef crate, or add as deps).

Dock invisibility is **inherent** at the window level on macOS — Dock entries are per-app, not per-window. Do NOT set `LSUIElement` / `NSApp.setActivationPolicy(.accessory)` — that'd hide the parent's Dock entry too.

### C2 — Linux (X11 via GTK)

- New module `agentmux-cef/src/floating_pane_linux.rs` (gated `#[cfg(target_os = "linux")]`).
- `gtk_window_set_transient_for(child, parent)` — sets `WM_TRANSIENT_FOR` ICCCM atom for the owner relationship.
[...]

PR #1082, opened 2026-05-27
feat(floating-pane): native title-bar drag + size matches source pane

AgentX-asaf:
> **Stacked on top of #1081** (the dev Vite-port URL fix). When that merges, this PR's base will auto-rebase to `main` and the diff will collapse to just these changes.

## Summary

Two Phase-6 polish items for the floating-pane tear-off (issue #1079 / spec `SPEC_FLOATING_PANE_TEAROFF_2026_05_11.md`):

1. **Native title-bar drag** — the floating pane's outer HWND uses a custom WndProc that returns `HTCAPTION` over the top 30 CSS px (excluding the close button's 36 CSS px zone on the right). The OS handles the drag loop natively — cross-monitor and cross-DPI safe.
2. **Size matches source pane** — at tear-off time the frontend measures the source pane's `getBoundingClientRect`, multiplies by `devicePixelRatio` for Win32 physical-px conversion, and passes that to `open_floating_pane_window`. Defensive min clamps (200×120) + fallback to the prior 720×480 if the DOM element isn't found.

## Host details (`floating_pane.rs`)

`floating_pane_wndproc` mirrors the existing `install_frameless_resize_hook` pattern in `agentmux-cef/src/client/wndproc.rs` and adds:

- `WM_NCCALCSIZE → 0` — claims the entire window rect as client area so the system title bar isn't drawn. Without this, the user sees doubled-up chrome (system title bar above the frontend's custom 30 px header).
- `WM_NCACTIVATE → 1` — suppresses the DWM activation border repaint.
- `WM_NCHITTEST` — 6 px edge resize zones, then `HTCAPTION` for the top 30 CSS px excluding the rightmost 36 CSS px (where the close button sits). All measurements DPI-scaled via `GetDpiForWindow`.

## Why not reuse `useWindowDrag` (the main window's hook)?

The main window's drag is JS-driven (`useWindowDrag.win32.ts` → `get_window_position` / `set_window_position` via IPC). The Phase 1 spec for floating panes referenced this as the pattern to mirror, but it has a latent multi-window bug: both backend handlers resolve the HWND via `find_own_top_level_window`, which does a process-wide `EnumWindows` and returns whichever top-level window the OS enumerates first — **not** the calling floating pane. The native WM_NCHITTEST path is self-contained, faster (no IPC roundtrip per mousemove), and side-steps that bug for the floating-pane case.

## Frontend details (`CrossWindowDragMonitor.win32.tsx`)

`measureSourcePaneSize(blockId)`:
1. Find source pane via `[data-blockid=…]` (the standard pane identifier from `blockframe.tsx:755`).
2. Read `getBoundingClientRect()` — CSS px.
3. Multiply by `window.devicePixelRatio` to match Win32's physical-px expectation.
4. Min-clamp to (200, 120) defensively.
5. Fall back to (720, 480) only if the element isn't found.

Must be called **before** `TearOffBlock` — that mutation removes the block from the source layout and the DOM element unmounts.

## Test plan

- [x] Frontend TS check — no new errors on either touched file
- [ ] `task dev` → tear off a pane → floating window:
  - Has no system title bar above the custom header
  - Drags by clicking + dragging the custom header
  - Close button still works (click passes through, not interpreted as drag)
  - Opens at roughly the same dimensions as the source pane (not 720×480)
  - Resize at edges works (6 px corner zones)

## Out of scope

- macOS NSPanel + Linux X11 GTK primitives (still tracked in #1079)
- WM_CLOSE → `CloseBrowser(false)` routing (noted in the inline TODO; default cascade still works)
- The `[on_before_close] no backend window ID registered for label=…floating… — shells may orphan` warning (separate reducer-plumbing gap)

### DIFF
--- a/.changesets/1779850836-feat-floating-pane-native-title-bar-drag-via-wm-nchittest-si-7ljq.md
+++ b/.changesets/1779850836-feat-floating-pane-native-title-bar-drag-via-wm-nchittest-si-7ljq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(floating-pane): native title-bar drag via WM_NCHITTEST + size matches source pane
--- a/agentmux-cef/src/commands/floating_pane.rs
+++ b/agentmux-cef/src/commands/floating_pane.rs
@@ -47,11 +47,19 @@ pub struct OpenFloatingPaneArgs {
     #[serde(default)]
     pub workspace_id: Option<String>,
     /// Screen-space top-left coordinates where the new floating window
-    /// should appear. Typically the cursor position at drop time.
+    /// should appear, in Win32 PHYSICAL pixels. Typically the cursor
+    /// position at drop time (frontend reads from host
+    /// `get_cursor_point` which uses `GetCursorPos`).
     pub x: i32,
     pub y: i32,
-    /// Initial window size. Phase 6 will memo the source pane's last
-    /// docked size; Phase 1 accepts a caller-provided default.
+    /// Initial window size in CSS / DIP pixels (NOT physical). The host
+    /// scales to physical px using the destination monitor's DPI via
+    /// `MonitorFromPoint(x, y)` + `GetDpiForMonitor`. Passing DIP here
+    /// (rather than physical, which would need the frontend to know the
+    /// destination monitor's DPI) lets us correctly cross-DPI handoff —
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1068, opened 2026-05-26
feat(agent-pane-state): reducer state + commands for composer details panel

AgentA-asaf:
**PR A of 2** — composer slim-status redesign foundation (full spec: \`docs/specs/SPEC_AGENT_COMPOSER_SLIM_STATUS_2026_05_26.md\`).

This PR is reducer-slice-only — no view changes. PR B will land the view redesign (\`AgentComposerStrip\` + \`AgentComposerDetails\`) consuming the atoms this PR exposes.

## Why reducer-owned

Composer state (panel open/closed, unread badge counter) lives as long as the pane and coordinates with the turn machine (auto-collapse on send). Solid-local signals would create a fourth tier of state outside the Phase B contract — exactly the dual-write pattern PR G burned out. Adding to \`AgentPaneState\` keeps the contract intact and makes the composer the same shape as \`turnPhase\` / \`sessionStats\` / \`currentTool\`.

## State additions

- \`detailsOpen: boolean\` — panel expand/collapse, default false
- \`composerUnreadCount: number\` — activity-log entries that arrived while panel was closed; resets on open

## Commands (4 new)

| Command | Effect |
|---|---|
| \`DetailsToggle\` | flip \`detailsOpen\`; reset unread on flip-to-open |
| \`DetailsExpand\` | explicit open (idempotent if already open with 0 unread) |
| \`DetailsCollapse\` | explicit close (idempotent if closed, preserves unread) |
| \`LogEntryArrived\` | increment unread when closed; no-op when open |

## Cross-arm change

\`TurnStart\` now also clears \`detailsOpen\` (sending a message hides the panel — same UX rule as today's send-clears-focus). Pure additive write to the existing return value; no impact on existing turn-machine invariants.

## Tests

12 new tests covering each arm + the \`TurnStart\` cross-arm + same-ref no-op identity preservation. Full reducer suite: **137/137 pass** (125 existing + 12 new).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1083, opened 2026-05-27
fix(agent-pane): drop TurnStart auto-collapse — causes Solid replaceChild crash

AgentA-asaf:
User-reproducible crash in v0.38.16: send any message → agent pane crashes with `NotFoundError: Failed to execute 'replaceChild' on 'Node'`. 100% reproducible across sessions, persists after pane reload.

## Root cause

PR #1068 added a `detailsOpen: false` write to the `TurnStart` reducer arm (auto-collapse on send-message, per spec §5.4). That write fires at the same tick as the first `StreamFlush` against agent-document-store. The flip causes `<Show when={detailsOpenAtom()}>` to unmount `AgentControlBar` + `ActivityLogPanel`, triggering the documented cascade-disposal:

```
CASCADE_DETECTED: slot disposed mid-dispatch
  cmd=StreamFlush, blockId=...
  "A documentAtom subscriber unmounted the pane during this dispatch"
```

Solid then reconciles against a disposed pane → `replaceChild` error.

See `docs/analysis/LIFECYCLE_DISPATCH_LEAK_2026_05_15.md` for the bug class.

## Fix

Drop the `detailsOpen: false` write from `TurnStart`. The reducer arm leaves `detailsOpen` alone; users click ▴ themselves.

Small UX cost (panel stays open during turn); crash beats UX every time.

Future structural work (deferred): identify the specific documentAtom subscriber that disposes synchronously and route it through `dispatchIfRegistered`. Auto-collapse returns once cascade root is fixed.

## Tests

Updated reducer test to assert detailsOpen is PRESERVED (not flipped) by TurnStart. Suite: 137/137 still pass.

### DIFF
--- a/.changesets/1779851568-fix-agent-pane-drop-turnstart-auto-collapse-causes-solid-rep-liyu.md
+++ b/.changesets/1779851568-fix-agent-pane-drop-turnstart-auto-collapse-causes-solid-rep-liyu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): drop TurnStart auto-collapse (causes Solid replaceChild crash)
--- a/frontend/app/store/agent-pane-state/reducer.test.ts
+++ b/frontend/app/store/agent-pane-state/reducer.test.ts
@@ -2114,15 +2114,19 @@ describe("agent-pane-state reducer", () => {
             expect(s.composerUnreadCount).toBe(1);
         });
 
-        it("TurnStart auto-collapses an open details panel", () => {
-            // Cross-arm interaction: sending a message hides the panel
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentA-asaf:
Closing — investigation in `docs/analysis/AGENT_PANE_REPLACECHILD_CRASH_ON_SEND_2026_05_27.md` falsified the hypothesis. The `TurnStart auto-collapses detailsOpen` write is NOT the cause of the replaceChild crash. Bug reproduces in v0.38.11 (built before #1069) with the same signature. The spec-prescribed auto-collapse is correct UX behavior and stays in main. Re-investigation pending source-mapped debug bundle.

PR #1086, opened 2026-05-27
fix(agent-pane): swap virtualizer For→Index — fixes replaceChild crash on send

AgentA-asaf:
Hard blocker: every send-message crashes the agent pane on the first `StreamFlush` with a Solid `NotFoundError: Failed to execute 'replaceChild' on 'Node'`. Latent since the virtualization redesign (PR #784).

## Root cause

`AgentDocumentVirtualList` rendered the virtual rows with:

```tsx
<For each={virtualizer.getVirtualItems()}>
```

Solid's `<For>` keys items by **referential identity**. TanStack Virtual's `getVirtualItems()` returns a freshly-allocated array of `VirtualItem` objects on every render. Every StreamFlush + every scroll = all items "new" to Solid → full re-key → reconcile from scratch → `replaceChild` on every row.

TanStack's async `measureElement` ResizeObserver then races against the reconcile. A row's DOM gets mutated under TanStack's stored ref, the `Missing attribute name 'data-index={index}'` warning fires (orphan measurement target), and Solid's reconciler hits a row whose `parentNode` is no longer the virtualizer container — throws `NotFoundError`.

## Fix

Swap `<For>` → `<Index>`. Same file already uses `<Index>` for the streaming-buffer below for the same class of reason (immutable updates → new object identity → `<For>` mis-keys). Adjust callbacks to call the item accessor (`virtualItem()` instead of `virtualItem`) and drop the unused `For` import.

## Full investigation

`docs/analysis/AGENT_PANE_REPLACECHILD_CRASH_ON_SEND_2026_05_27.md` — includes the bisect that ruled out my recent agent-pane PRs, the debug-bundle source-mapped stack pointing at `reconcileArrays`, and the §13 root-cause writeup.

## Verify

`task build:frontend` clean. Pending live verification — about to build a portable and test the send-message reproducer.

### DIFF
--- a/.changesets/1779854364-fix-agent-pane-swap-virtualizer-for-index-fixes-replacechild-nwbi.md
+++ b/.changesets/1779854364-fix-agent-pane-swap-virtualizer-for-index-fixes-replacechild-nwbi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): swap virtualizer For→Index — fixes replaceChild crash on send
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.38.14"
+version = "0.38.15"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

AgentA-asaf:
Closing — fix did NOT resolve the crash. Same `reconcileArrays` signature reproduces in v0.38.16 built off this branch. Either there is another `<For>` mis-keying that I haven't found, or the root cause is different from the keying-identity hypothesis. Investigation continues — see `docs/analysis/AGENT_PANE_REPLACECHILD_CRASH_ON_SEND_2026_05_27.md` §15.

PR #1088, opened 2026-05-27
feat(editor): tabs Phase 1B — view + tab strip

AgentY-asaf:
## Summary

Phase 1B of the editor tabs feature. Builds on the slice scaffolding from #1087 (Phase 1A) — wires the `EditorViewModel` to slice #10 (`editor-pane-state`) and renders a tab strip above CodeMirror.

Spec: `specs/SPEC_EDITOR_TABS_2026-05-26.md` (untracked locally; not in this PR).

## What's in

- **EditorViewModel** — single-file signals removed; all derived accessors (`filePathAtom`, `dirtyAtom`, `readOnlyAtom`, `errorAtom`, `loadingAtom`, `languageAtom`) now derive from the slice's active tab. `contentAtom` reads from a view-local `Map<tabId, content>` (content deliberately not in the slice — large blobs, not auditable). `openFile` dispatches `OpenFile`, loads content via RPC, dispatches `TabContentLoaded` with sha256 hash. New `closeTab` / `switchTab` / `reopenLastClosed` action methods. New `onSliceEvent(handler)` subscription for view-side reactions to slice events.
- **EditorTabStrip** (new component) — renders one tab per open file. Click to activate, × (hover-shown, always for dirty) to close, middle-click to close. Title attr shows full path; active tab gets `--accent-color` underline. Tab strip stops dblclick propagation so fast clicks don't maximize the pane (matches the icon-toggle pattern from #1071/#1084).
- **editor-view.tsx** — per-tab `Map<string, EditorState>` for cursor/selection/scroll/undo preservation across tab switches. The `createEffect` now tracks `activeIdAtom` (not just `filePathAtom`); snapshots the outgoing tab's state before rebuilding. `cmStates` Map is cleared on `TabClosed` events via the slice subscription.
- **editor-view.scss** — tab strip styles (28px high, 80–140px tabs, hover/active/dirty modifiers).
- **Backwards-compat hydration** — existing block meta key `"file"` (the single-file key) still restores as a one-tab list on first mount. The 1C saga will replace this with `editor:tabs` persistence.

## Out of scope (follow-ups)

- **Dirty-confirm modal** — current behavior: × force-closes dirty tabs, matching today's pane-close behavior. Modal lands in a follow-up commit that consumes the slice's `RequestDirtyConfirm` event.
- **Keyboard shortcuts** (Ctrl+W, Ctrl+Tab, Ctrl+1..9, Ctrl+Shift+T) — needs `keymodel.ts` registration.
[...]

PR #1091, opened 2026-05-27
fix(editor-pane-state): wake subscribers on every state change

AgentY-asaf:
## Summary

**Bug:** After PR #1088 (editor tabs Phase 1B) merged, clicking a file in the file tree fires the backend \`readeditorfile\` RPC and updates block meta — but the editor pane stays blank. Looks frozen even though all events flow.

**Root cause:** The slot \`dispatch\` was gating \`eventSink(events)\` on \`result.events.length > 0\`. The reducer's \`TabContentLoaded\` handler mutates the tab record (sets \`contentLoaded: true\`, stores \`contentHash\`) but returns \`events: []\` — Phase 1A shipped without the spec's \`TabReady\` event. So:

1. User clicks file → \`openFile\` → dispatch \`OpenFile\` → slice creates a tab + emits \`TabOpened\`/\`TabActivated\` (events array non-empty → sink fires → \`sliceVersion\` bumps → view re-renders) ✓
2. \`readeditorfile\` resolves → dispatch \`TabContentLoaded\` → slice updates \`tab.contentLoaded = true\` but events array is empty → **sink does NOT fire** → \`sliceVersion\` doesn't bump → \`loadingAtom\` projection silently misses the update → \`createEffect\` doesn't re-run → \`setupEditor\` never builds CodeMirror → pane stays blank ✗

## Fix

Call \`eventSink\` whenever \`result.state\` reference changes, regardless of events array length. The empty-events case is a no-op for consumers that iterate \`events\` and a wake-up for consumers (like view projections) that don't.

The no-op SwitchTab/ReorderTab paths correctly DON'T fire the sink because they return the original \`state\` reference unchanged.

## Test plan

- [x] \`npx vitest run frontend/app/store/editor-pane-state-store.test.ts\` → 35/35 still passing
- [ ] Smoke test in \`task dev\`: open editor pane → click file in tree → file opens in pane (currently broken without this fix)

### DIFF
--- a/.changesets/1779872611-fix-editor-pane-state-call-eventsink-whenever-state-changes--nxo7.md
+++ b/.changesets/1779872611-fix-editor-pane-state-call-eventsink-whenever-state-changes--nxo7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(editor-pane-state): call eventSink whenever state changes, not only when events array is non-empty
--- a/frontend/app/store/editor-pane-state-store.ts
+++ b/frontend/app/store/editor-pane-state-store.ts
@@ -644,7 +644,14 @@ export function dispatch(
     const result = update(prev, command);
     slot.state = result.state;
 
-    if (eventSink && result.events.length > 0) {
+    // Call the sink whenever STATE changed, even if no semantic events were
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1092, opened 2026-05-27
feat(editor): full-pane error panel + hide focus ring in floating panes

AgentY-asaf:
## Summary

Two small UX improvements bundled together.

### 1. Full-pane error panel when a file fails to load

Invalid path / permission denied / binary file currently routes through the slice's \`TabContentLoadFailed\` → \`loadError\`. Old surface: a thin top banner over an empty CodeMirror, which reads as "everything's fine, minor note up top". New surface: full-pane card with the file path, the error text, and a single **Close tab** button. CodeMirror is suppressed while the error is present. Operational errors that occur *after* content loaded (save failures, etc.) still surface via the existing top banner.

### 2. Suppress the focus ring inside floating-pane windows

A floating window holds exactly one block, so the standard \`.block-mask\` accent-color focus ring would be permanent chrome — visual noise the user can't dismiss. Added a \`.floating-pane-workspace\` class on the workspace container and a targeted SCSS rule that nulls the ring inside it. Standard panes (docked or with siblings) keep their focus ring as before.

## Diagnostics

Kept the \`[editor-tabs]\` \`console.log\` lines in \`editor-model.ts\` and \`editor-view.tsx\` — we're still iterating on the tabs surface and they're producing useful signal in the host log. Strip in a follow-up once the API stabilizes.

## Test plan

- [ ] Open editor pane → type a bogus path in the path input → submit → centered error card appears, CodeMirror is gone, Close tab works.
- [ ] Open a valid file → save → centered error does not appear; if save fails, top banner appears (existing behavior preserved).
- [ ] Tear off a pane to a floating window → no blue focus ring around the body.
- [ ] Docked panes still have the accent-color ring when focused.

### DIFF
--- a/.changesets/1779873682-feat-editor-full-pane-error-panel-for-invalid-files-hide-foc-4642.md
+++ b/.changesets/1779873682-feat-editor-full-pane-error-panel-for-invalid-files-hide-foc-4642.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(editor): full-pane error panel for invalid files + hide focus ring in floating panes
--- a/frontend/app/block/block.scss
+++ b/frontend/app/block/block.scss
@@ -500,7 +500,19 @@
                 border-color: var(--block-agent-color);
             }
         }
+    }
+}
 
+// Inside a floating pane window, the single block is always the focused
+// block (it's the only one), so the accent-color ring would be permanent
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

AgentY-asaf:
Closing — got tangled up with floating-pane scope. Reopening with editor-only changes.

PR #1084, opened 2026-05-27
feat(editor): per-pane zoom + decouple icon-as-toggle from header dblclick

AgentY-asaf:
## Summary

Two small UX wins for the editor pane, shipped together.

### 1. Pane-icon double-click no longer maximizes the pane

The editor pane icon is wired as the file-tree expand/collapse toggle (PR #1071). The pane header itself listens for `onDblClick → toggleMagnify`, so fast double-clicks on the toggle were propagating up and maximizing the pane in addition to toggling — annoying when you're rapidly opening/closing the tree.

Wrapped the iconbutton variant of `getViewIconElem` in a `<span onDblClick={e => e.stopPropagation()}>`. The text-rendered variant (panes that don't use the icon as a clickable toggle) is unchanged, so non-toggle panes still maximize on header double-click. End icons already had this guard.

### 2. Editor pane plugs into the universal per-pane zoom system

Ctrl+Wheel over an editor pane now resizes the CodeMirror text and the file-tree text together. Same `term:zoom` block meta the terminal, agent, and swarm panes use, so the existing keyboard shortcuts (Ctrl+= / Ctrl+- / Ctrl+0) and the zoom indicator overlay work automatically.

- Extended `getBlockZoom` in `zoom.{win32,linux,darwin}.ts` to accept `editor` viewType.
- Added `zoomAtom` to `EditorViewModel`, wrapped in `useBlockAtom` (which uses `createRoot` under the hood). A bare `createMemo` in the constructor wouldn't have a tracking owner and would snapshot once instead of subscribing to meta updates — that was the original bug.
- Wheel listener on the editor root in capture phase; `SetMetaCommand` persists `term:zoom` (clamped 0.5–2.0).
- SCSS reads `--editor-zoom` on `.editor-view` to scale `.cm-scroller` (13px base) and `.file-tree` (12px base). Chrome (toolbar buttons, banners, status chip) is intentionally **not** zoomed.

Spec: [`specs/SPEC_EDITOR_FILE_TREE_2026-05-26.md`](https://github.com/agentmuxai/agentmux/blob/main/specs/SPEC_EDITOR_FILE_TREE_2026-05-26.md) (icon-as-toggle pattern).

## Test plan

- [ ] Editor pane: double-click the folder icon → tree toggles, pane does **not** maximize.
- [ ] Header double-click on title text → pane still maximizes (regression check on non-icon panes).
[...]

PR #1093, opened 2026-05-27
feat(editor): preview tabs + error panel + binary guard + reactivity fix

AgentY-asaf:
## Summary

Bundle of editor-pane improvements landed together — they share infrastructure (the `EditorViewModel` signal-projection layer and the slice's tab list), so reviewing them as one PR is easier than splitting.

### 1. VS Code-style preview tabs (Phase 2 of `SPEC_EDITOR_TABS`)

- **Tree single-click** opens a file in a *preview* tab (italic title). Subsequent single-clicks **replace its file in-place** — at most one preview tab per pane.
- **Tree double-click** pins the tab.
- **Double-click on a preview tab in the strip** pins it.
- **Editing a preview tab** auto-promotes it to pinned (matches VS Code).

Slice changes:
- New `isPreview: boolean` on `EditorTab`.
- New `mode?: "preview" | "pinned"` option on `OpenFile` (default **pinned** for programmatic safety — only the tree-click site passes `mode: \"preview\"`).
- New `PinTab` command.
- `MarkDirty` auto-pins.
- `OpenFile` on a path that's currently the preview, with `mode=pinned`, now **pins it** rather than just activating (handles the dblclick-pinning-existing-preview flow correctly).

View-model changes:
- `openFile(path)` → pinned (legacy semantics preserved for existing callers).
- `openFilePreview(path)` → preview slot.
- `pinActiveTab()` → for the tab strip's dblclick handler.
- **Stale-cache fix**: preview-swap mutates the same `tabId` but resets `contentLoaded=false` in the slice. The view-model's content cache now consults `tab.contentLoaded` from the slice and reloads when stale.

View changes:
- `file-tree.tsx`: separate `onFileClick` (preview) / `onFileDblClick` (pinned) callbacks.
- `editor-tab-strip.tsx`: italic label on preview tabs, dblclick handler pins, contextual title attribute explains the affordance.

### 2. Centered error panel for invalid files

When a tab fails to load, CodeMirror is suppressed and a full-pane card replaces it: icon + file path + error message + Close-tab button. Operational errors *after* content loaded (e.g. save failures) still surface via the existing small top banner.

### 3. Binary-content / long-line guard

A NUL-byte sniff (first 4 KB) rejects binary files; a single-line-length sniff (>100 K chars) rejects pathological text. The trigger case was a Windows `NTUSER.DAT` regtrans-ms file (512 KB on one line) that hung CodeMirror's layout thread indefinitely. The 10 MB read cap didn't catch it. The refusal message flows through `TabContentLoadFailed` → the centered error panel.

### 4. Reactivity fix for slice projections

The slice's `tabsAtom` / `activeIdAtom` / `activeTabAtom` / `contentAtom` `createMemo`s were created in the model constructor **outside any reactive root**, so subscribers wouldn't reliably re-run on slice mutations. Wrapped in `useBlockAtom` (which uses `createRoot` under the hood) — same fix PR #1084 applied to `zoomAtom`. Without this, clicking a file would fire all the right dispatches but the `createEffect` in `editor-view` would never re-run and CodeMirror would never build for the loaded content.

### 5. Module-level event-sink fan-out

Replaced the per-instance `setEventSink` in `EditorViewModel` with a module-level singleton that fans out to a `Set` of per-instance handlers. The previous design was last-writer-wins: opening a second editor pane silently disabled the first pane's slice subscription.

## Diagnostics

Kept the `[editor-tabs]` `console.log` lines — we're still iterating and the host-log output has been the key to landing the right fixes (slice-sink wake-up, reactivity wrap, stale-cache evict). Strip in a follow-up once the surface stabilizes.

## Test plan

- [x] `npx vitest run frontend/app/store/editor-pane-state-store.test.ts` → 35/35 passing
- [x] `npx tsc --noEmit` clean for editor files
- [ ] Smoke test in `task dev`:
  - Tree single-click → italic preview tab opens; subsequent single-clicks REPLACE the file in the same tab.
  - Tree double-click → pinned tab (non-italic).
  - Double-click preview tab in strip → pins it (italic gone).
  - Open `.regtrans-ms` or any binary → centered error card with "looks binary" message, **renderer doesn't freeze**.
  - Open valid file → loads normally.
  - Open multiple panes simultaneously → both work independently (the fan-out fix).

### DIFF
--- a/.changesets/1779875373-feat-editor-preview-tabs-centered-error-panel-binary-file-gu-e0ny.md
+++ b/.changesets/1779875373-feat-editor-preview-tabs-centered-error-panel-binary-file-gu-e0ny.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(editor): preview tabs + centered error panel + binary-file guard + memo reactivity fix
--- a/frontend/app/store/editor-pane-state-store.ts
+++ b/frontend/app/store/editor-pane-state-store.ts
@@ -49,6 +49,11 @@ export interface EditorTab {
     loadError: string | null;
     /** Transient — true once the lazy fetch resolved. Not persisted. */
     contentLoaded: boolean;
+    /** Preview tab — at most one per pane. A single-click in the tree
+     *  opens into the preview slot (replacing the current preview's file).
+     *  Double-click opens as pinned. Editing or explicit pin promotes a
+     *  preview to pinned. Matches VS Code semantics. */
+    isPreview: boolean;
 }
 
 export interface ClosedTab {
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

AgentY-asaf:
Reagent findings addressed in a248ea46:

- **P1 post-await race**: snapshot the slice after the read RPC resolves; if the tab's filePath no longer matches the canonical path we requested, bail. Prevents path A's content from landing in a tab the slice has since swapped to path B.
- **P1 dblclick pin during loading**: the early-return path for in-flight loads now still dispatches OpenFile so the slice runs its pin-if-existing logic. The RPC continues populating content as usual.
- **P2 error panel hiding CM on save failure**: TabContentLoadFailed in the slice now preserves contentLoaded instead of forcing it to false. Save failures stay on the small top banner; initial-load failures route to the centered panel as before.
- **P2 diagnostic logs**: stripped all [editor-tabs] console.logs — the surface is stable now.

Slice tests still 35/35 green.

PR #1089, opened 2026-05-27
feat(floating-pane): floater = free-floating docked pane, no extra chrome

AgentX-asaf:
## Summary

User feedback: \"the pane header on the floating window is the same as the pane header on a docked pane .. treat it as if you tore off the docked pane and it appeared exactly like it did, except freefloating .. make sure no windows or OS-level controls added\".

Three changes to make the floating window render identically to a docked pane:

1. **Removed the custom 30 CSS-px title bar** from \`frontend/app/workspace/floating-pane-workspace.tsx\` (the placeholder header with \"Floating pane\" text + custom × button). The block renders its standard \`BlockFrame_Header\` instead — same icons, title, close, magnify, mic, endIconButtons as docked.

2. **\`DwmExtendFrameIntoClientArea\`** added to \`agentmux-cef/src/floating_pane.rs::create_owned_popup\` after \`CreateWindowExW\`. Without it, DWM kept painting the standard Win32 caption (with min/max/close buttons) on top of our client area, even though \`WM_NCCALCSIZE → 0\` reserved no space for it. Mirrors the main window's \`setup_native_frameless\` pattern.

3. **Updated \`floating_pane_wndproc\` HTCAPTION zone**:
   - height: 33 CSS px (matches \`--header-height\` in \`theme.scss:97\`)
   - excluded right-anchored 130 CSS px so the docked-pane action cluster (close, magnify, mic, endIconButtons) remains clickable. All DPI-scaled via \`GetDpiForWindow\`.

## Net effect

Tear off a pane → floating window appears with the exact docked-pane visual, no OS chrome, no extra widgets. Drag by the title-text area of the pane header moves the window; click the action buttons (right side) to act on the block.

## Test plan

- [ ] Tear off a pane → floating window has no system title bar / min-max-close buttons
- [ ] The block's standard pane header is the only header
- [ ] Drag the title-text area of the pane header → window moves
- [ ] Click the × in the pane header → window closes (single block → workspace empty → cascade)
- [ ] Magnify / mic / endIconButtons on the right still respond to clicks
- [ ] Resize at edges works (DPI-correct 6 CSS px borders)

## Out of scope

- **Full JS-driven drag** — would let action buttons live anywhere in the header (not just the right cluster). Requires data-drag-region threading + host \`set_window_position\` label routing fix. Tracked as a follow-up.
[...]

PR #1094, opened 2026-05-27
fix(floating-pane): suppress focus-ring + route window-drag by label

AgentX-asaf:
Two iterative fixes from user testing of #1089:

## 1. Suppress the colored focus ring on the floater

The standard block-frame draws a 2 CSS-px colored ring (\`--accent-color\` or \`--block-agent-color\`, see \`block.scss:488,500\`) when the block is \`.block-focused\`. In a single-pane floater the block is *always* the focused one, so the ring is always visible and reads as an unwanted "colored border" around the window.

Added a scoped SCSS override on the new \`.floating-pane-workspace\` root class that sets \`.block-mask\` border-color transparent. \`!important\` is needed because the rules being overridden compile to specificity (0,4,0) and (0,5,0); matching that chain manually would still tie on the agent-color variant.

## 2. Route window-drag IPC by label, not Z-order

**User report:** dragging the MAIN window moved the FLOATING window instead.

**Root cause:** \`get/set_window_position\` resolved the HWND via \`find_own_top_level_window\` — process-wide \`EnumWindows\` returning the topmost-Z visible top-level. Owned floating panes sit *above* their owner in Z-order, so as soon as one exists, every label-less position IPC accidentally targets the floater (including the main window's own \`useWindowDrag\` hook).

This is the latent bug called out in \`docs/analyses/ANALYSIS_FLOATING_PANE_HEADER_DRAG_2026-05-27.md\` as the open follow-up before Phase 4 (re-dock).

### Backend

- New \`resolve_window_hwnd(state, label)\` helper in \`commands/window.rs\`:
  - Prefers reducer registry: \`state.get_browser(label)\` → host → \`window_handle()\` → \`GetAncestor(hwnd, GA_ROOT)\` so we always return the outer top-level.
  - Falls back to \`find_own_top_level_window\` when no label given (back-compat).
- \`get_window_position\` now takes \`args\` and reads \`label\` (IPC dispatcher in \`ipc.rs\` updated).
- \`set_window_position\` already accepted \`label\` but was ignoring it on Windows; now uses the helper.

### Frontend

- \`useWindowDrag.win32.ts\` (main window): reads \`windowLabel\` from URL query params (defaults to \`"main"\`) and passes it on every position IPC call.
- \`floating-pane-workspace.tsx\` (floater): captures the windowLabel memo's value at mount, threads it through the IPC calls.

## Test plan

- [x] \`cargo check -p agentmux-cef\` clean
- [x] \`tsc --noEmit\` clean for touched files
- [ ] With a floating pane open, drag the **main** window's title bar → main window moves (not the floater)
- [ ] Drag the floating pane's header → floater moves (not main)
- [ ] No colored ring around the floating pane
- [ ] Buttons inside the floater's pane header still respond (close/magnify/mic/etc.)

## Cross-references

- \`docs/analyses/ANALYSIS_FLOATING_PANE_HEADER_DRAG_2026-05-27.md\` — original analysis flagging this as the open follow-up.
- \`docs/specs/SPEC_FLOATING_PANE_REDOCK_2026-05-27.md\` — re-dock plan; this fix unblocks the off-target floater repositioning during Phase 4.
- #1089 — preceding chrome cleanup + JS-driven drag PR.

### DIFF
--- a/.changesets/1779899676-fix-floating-pane-suppress-focus-ring-border-route-window-dr-b2bs.md
+++ b/.changesets/1779899676-fix-floating-pane-suppress-focus-ring-border-route-window-dr-b2bs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): suppress focus-ring border + route window-drag/close by label
--- a/agentmux-cef/src/commands/window.rs
+++ b/agentmux-cef/src/commands/window.rs
@@ -158,20 +158,63 @@ pub fn maximize_window(state: &Arc<AppState>, args: &serde_json::Value) -> Resul
     Ok(serde_json::Value::Null)
 }
 
+/// Resolve a top-level HWND for the given label. Prefer the reducer
+/// registry (`state.get_browser(label)` → host → window_handle → walk
+/// to root) over the process-wide `find_own_top_level_window` fallback.
+///
+/// Why the label matters: `find_own_top_level_window` does an
+/// `EnumWindows` and returns the *first* visible top-level of the
+/// current process. Z-order puts **owned** windows ABOVE their owner,
+/// so as soon as a floating-pane window exists, every label-less
+/// `get/set_window_position` call (e.g. the main window's
+/// `useWindowDrag`) accidentally targets the floater — dragging the
+/// main window moves the floater instead.
+///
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #948, opened 2026-05-21
feat(menu): menu positioning framework — paintable-area guard (spec + Phases 1-4)

AgentX-asaf:
## Summary

Unifies AgentMux's menu positioning. Spec + full implementation (Phases 1-4). Every DOM menu now offsets through one paintable-area-aware primitive.

## Commits

| Commit | Phase |
|---|---|
| spec | Analysis + design — `SPEC_MENU_PAINTABLE_AREA_GUARD_2026_05_20.md` |
| Phase 1 | `useMenuPosition` primitive — `frontend/app/util/menu-position.ts` + 15 tests |
| Phase 2 | FlyoutMenu + Popover migrated; ad-hoc submenu offset deleted |
| Phase 3 | More dropdown + TokenBreakdownPopover migrated |
| Phase 4 | `assertMenuInPaintableArea` dev guard + `check-menu-positioning.sh` CI gate |

## What changed

- **The paintable area** = window viewport **minus** native browser-pane HWND rects. Browser panes are `CefBrowserView` native child windows that paint *above* the DOM — a DOM menu over one is invisible regardless of viewport math. The primitive uses the paintable area (not the viewport) as floating-ui's overflow boundary.
- **One positioner** — `useMenuPosition` / `computeMenuPosition`: fixed `offset`+`flip`+`shift`+`size` middleware. FlyoutMenu, MenuButton, Popover, NotificationPopover, More dropdown, TokenBreakdownPopover all route through it.
- **Deleted** the hand-rolled offset code: FlyoutMenu's one-shot `flipped` submenu flag, the More dropdown's manual `innerWidth - rect.right`, TokenBreakdownPopover's bespoke gutter.
- **Guard** — dev-only `[menu-guard]` runtime assertion + a CI gate (`check:menu-positioning` task) that fails on stray `computePosition` outside sanctioned files.
- Native context menus stay native (correctly exempt).
- Tooltip left as-is (spec Q3 — already correct).

## Status

- [x] `npm run build` green
- [x] `npm test -- menu-position` — 15/15
- [x] `check-menu-positioning.sh` gate passes
- [ ] **Runtime regression testing pending** — the FlyoutMenu/Popover migration changes how every DOM menu positions; needs a `task dev` pass over the spec §9 matrix (menus near window edges, submenus, menu over a browser pane, tiny window). Not yet done.

## Note

[...]

PR #1098, opened 2026-05-27
perf(pane-overlay): three fixes from discussion #1097 — Agent2's plan

AgentY-asaf:
Picks up Agent2's analysis in #1097 of the menu-hover lag vs VS Code. Lands fixes 1, 2, and 3 of the proposed seven (all HIGH-impact / LOW-effort items). The remaining four (PR #948 dependency, \`bRedraw=FALSE\` + explicit paint, v8 in-process binding, perf budget alarm) are deferred to follow-ups.

## Background

Every overlay mount/unmount/resize fired an HTTP IPC → Rust → \`SetWindowRgn\` (with \`bRedraw=TRUE\` → \`WM_WINDOWPOSCHANGING/CHANGED\` + redraw) on every live pane HWND. One hover across menu items emitted 2–3 fetch round-trips per gesture. The pane redraw is off-frame from the DOM menu paint — until it lands, the menu pixels are still occluded by the pane's HWND. **That's the felt lag.**

### Fix 1 — short-circuit when no overlay intersects any pane

New \`frontend/app/platform/pane-rect-registry.ts\`: a per-window \`Map<blockId, PaneRect>\` populated by the browser view on \`browser_pane_create\` / \`browser_pane_resize\`, cleared on close. Stored in **CSS pixels** (matches overlay-rect space — no dpr math at the intersection check).

\`pane-overlay.ts:flushClip\` now bails when:
- there are **zero registered panes** in this window (workspaces with no browser pane — the common case for the hamburger / settings menus today), OR
- none of the overlay rects **intersects any registered pane** (AABB overlap test, zero allocation).

The dispatched-clip cache catches the "non-empty → empty" transition so closing the last overlay still clears the clip exactly once.

### Fix 2 — coalesce sendClip on requestAnimationFrame

The dominant gain. Previously every Map mutation fired an immediate IPC; the FlyoutMenu hover transition produces 2–3 in the same microtask checkpoint (Show toggle off, Show toggle on, overflow-correction style re-write). Now \`sendClip\` sets a flag and the actual \`flushClip\` runs **once per rAF tick** after the DOM mutations have settled.

### Fix 3 — drop \`class\` from the style-observer's filter

\`pane-overlay-auto.ts:track\` was watching \`attributeFilter: ["style", "class"]\`. Class mutations don't change box geometry — the \`ResizeObserver\` covers that case. Solid's \`classList\` toggles + portal \`data-attrs\` were firing \`updateRect\` on every hover-driven \`.active\` flip for no benefit. Down to \`["style"]\`.

## Tests

New \`pane-rect-registry.test.ts\` — **8/8 passing**:
- register/unregister counts (idempotent on duplicate unregister)
- AABB overlap on every edge + corner
- multi-pane "any intersects" semantics
- edge-touching rects (non-overlapping — boundary exclusive)
- zero-sized overlays
- overwrite-on-register
- empty-registry case

Existing tests untouched. \`npx tsc --noEmit\` clean for the touched files.

## Verification (per discussion #1097)

- [ ] DevTools Performance recording of 5s hamburger hover: count \`/ipc\` POSTs with \`browser_panes_set_overlay_clip\` in body — should drop dramatically (workspaces with no panes near the menu hit zero post-flush)
- [ ] \`muxlog host '\[pane-airspace\] applied overlay clip'\` during idle hover approaches 0
- [ ] Subjective: rapid hover across menu items, no perceived lag when the menu doesn't overlap a pane

## What's NOT in this PR

The other four fixes from Agent2's plan are deferred:
- Fix 4 — PR #948 (\`useMenuPosition\`) — orthogonal, already in flight
- Fix 5 — \`SetWindowRgn(bRedraw=FALSE)\` + explicit \`InvalidateRect\` — Rust-side, separate PR
- Fix 6 — v8 in-process binding for the hot IPC — high effort, defer
- Fix 7 — Perf budget alarm — observability, separate PR

## References

- [Discussion #1097](https://github.com/agentmuxai/agentmux/discussions/1097) — Agent2's root-cause + proposal
- [PR #948](https://github.com/agentmuxai/agentmux/pull/948) — paintable-area positioning (orthogonal)
- \`frontend/app/platform/pane-overlay.ts\`, \`pane-overlay-auto.ts\`, \`pane-rect-registry.ts\`
- \`frontend/app/view/browser/browser-view.tsx\`
- \`agentmux-cef/src/browser_panes.rs\` (unchanged this PR; downstream consumer)

### DIFF
--- a/.changesets/1779878889-perf-pane-overlay-short-circuit-ipc-coalesce-on-raf-drop-cla-24xe.md
+++ b/.changesets/1779878889-perf-pane-overlay-short-circuit-ipc-coalesce-on-raf-drop-cla-24xe.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+perf(pane-overlay): short-circuit IPC + coalesce on rAF + drop class observation (Agent2 #1097)
--- a/agentmux-cef/src/browser_panes.rs
+++ b/agentmux-cef/src/browser_panes.rs
@@ -423,7 +423,8 @@ impl BrowserPaneManager {
     ) {
         use windows_sys::Win32::Foundation::{POINT, RECT};
         use windows_sys::Win32::Graphics::Gdi::{
-            CombineRgn, CreateRectRgn, DeleteObject, MapWindowPoints, SetWindowRgn, RGN_DIFF,
+            CombineRgn, CreateRectRgn, DeleteObject, InvalidateRect, MapWindowPoints, SetWindowRgn,
+            RGN_DIFF,
         };
         use windows_sys::Win32::UI::WindowsAndMessaging::{
             GetAncestor, GetParent, GetWindowRect, GA_ROOT,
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

AgentY-asaf:
Two changes on top of the existing review:

**P1 fix (reagent feedback)** — `lastDispatchedKey` is no longer mutated on the non-dispatch branch. The gate now explicitly dispatches a CLEARING IPC (empty rects) when transitioning from a previously-sent non-empty clip to a state where nothing intersects, so Rust resets each pane's region instead of leaving a stale hole.

**Fix 5 from Agent2's plan (Rust-side async paint)**: `SetWindowRgn(... bRedraw=TRUE)` runs a synchronous redraw inside the call. Switched to `bRedraw=FALSE` followed by `InvalidateRect(hwnd, NULL, FALSE)` — the region updates immediately and WM_PAINT fires on the next message-pump tick. Stops the IPC handler from blocking on per-pane redraws.

Phase 2 follow-up: track previous overlay rects in AppState and invalidate only the affected sub-rect instead of the whole pane.

PR #1090, opened 2026-05-27
feat(fe-log): runtime source-map resolver for piped error stacks

AgentA-asaf:
## Summary

- Rewrites V8 minified `error.stack` positions (the `bpe → wp → ...` flavor) into original-file frames *before* they reach the host log, via a chunk-keyed `SourceMapConsumer` cache.
- Sync-first emit: primary log line goes out immediately with whatever frames are already cached (`stack_resolved: true | false | "partial"`); partial stacks trigger a one-shot async fetch and a follow-up `(stack-resolved)` line. `stack_raw` always travels alongside the resolved stack.
- Vite now emits `.map` files for every build (~30MB cost per portable; the trade-off is documented in `vite.config.ts`).

## Why

The agent-pane `replaceChild` crash on send-message is open with two falsified hypotheses (see `docs/analysis/AGENT_PANE_REPLACECHILD_CRASH_ON_SEND_2026_05_27.md`). Iterating on minified stacks via the DevTools paste-and-decode dance is the bottleneck — this PR removes it for that bug *and* every future renderer crash.

## Files

- `frontend/log/source-map-resolver.ts` — module-scope cache, lazy `.map` fetch, V8 frame regex, one-shot WARN per failing chunk.
- `frontend/log/source-map-resolver.test.ts` — 6 tests covering seeded sync resolve, async fetch + resolve, 404 caching, malformed frames.
- `frontend/log/error-forwarder.ts` — resolver wiring inside `forward()`: sync emit + async follow-up.
- `vite.config.ts` — `sourcemap: true` always.
- `package.json` / lockfile — adds `source-map-js@^1.2.1`.
- `docs/specs/SPEC_FE_SOURCE_MAP_RESOLVER_2026_05_27.md` — full design + acceptance criteria + future patterns.
- `docs/analysis/AGENT_PANE_REPLACECHILD_CRASH_ON_SEND_2026_05_27.md` — the open crash investigation that motivated this work.
- `.changesets/1779856022-...md` — patch changeset.

## Test plan

- [x] `npx vitest run frontend/log/source-map-resolver.test.ts` — 6/6 pass.
- [x] `npx tsc --noEmit -p .` — clean across resolver + forwarder.
- [x] `task build:frontend` — clean (built in 28s).
- [ ] After merge: build a portable, reproduce the agent-pane crash, confirm the host log shows resolved frames pointing at real `.tsx` files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1101, opened 2026-05-27
fix(agent-pane): unbreak send-message — id-collision render gap + replaceChild crash

AgentA-asaf:
## Yes, you saw them this time

Confirmed manually on v0.38.18 portable: send a message, agent response renders at the bottom of the conversation as a new node. No crash. 🎉

### Render gap — agent responses invisible after a turn

`ClaudeCodeStreamParser` was constructed fresh on every mount with `nodeIdCounter = 0`, so the first markdown chunk always got id `node_0`. When the agent pane mounted against a resumed-session snapshot that already had `node_0` in its document, the agent-document reducer's StreamFlush handler did the "id collision = merge into existing node" thing:

```rust
if (idx != null) {
    arr[idx] = mergeReplacement(arr[idx], n);  // collidedAndUpdated++
}
```

The agent's text was silently merged INTO the historical `node_0`, scrolled far above the visible area (n-124 nodes back in the repro). The user saw their own message, saw the spinner, saw nothing else.

**Fix:** per-parser-instance 6-char random `idPrefix`, rotated on `reset()`. Every generated id becomes `{prefix}_{counter}` — globally unique against snapshot ids.

### replaceChild crash on send

SolidJS reconciler `NotFoundError: Failed to execute 'replaceChild'` on every send-message. The partition split was count-based, so each doc append shifted one node from `streamingNodes` → `virtualizedNodes` in one reactive tick. The same node id existed in both `<For>`/`<Index>` subtrees for one render frame, the reconciler tripped.

**Fix:** sticky frontier id, anchored once when the doc first crosses `STREAMING_BUFFER_SIZE`. Appended nodes flow into `streamingNodes`; the virtualized head stays fixed; no cross-list migration on simple growth.

## Supporting infrastructure (kept)

- `frontend/log/render-trail.ts` — 50-entry ring buffer for "reactive activity just before the throw." Dumped by `BlockErrorBoundary` on catch. **This is what surfaced the id-collision bug** — `[stream-flush-diag] new=1 upd=1 newTypes=markdown newIds=node_0` next to `[partition-diag] total=125` (unchanged) told us the new node was a collision, not an append.
- `BlockErrorBoundary` now wires the source-map resolver from #1090 (sync + async follow-up) and dumps `render_trail` on catch.
- `trail()` taps at `dispatchPane`, `sendMessage`, `TurnStart`, every partition recompute.
- `@/log/*` tsconfig path mapping.

## Tests

- [x] 29/29 stream-parser tests pass (one updated for the id format)
- [x] 19/19 streaming-buffer tests pass (sticky-frontier covered with 4 new tests)
- [x] 8/8 BlockErrorBoundary tests pass at runtime
- [x] 9/9 source-map-resolver tests pass
- [x] Manual: v0.38.18 portable, "u there" round-trip, response renders, no crash

### DIFF
--- a/.changesets/1779885643-fix-agent-pane-sticky-frontier-partition-render-trail-diagno-hpm9.md
+++ b/.changesets/1779885643-fix-agent-pane-sticky-frontier-partition-render-trail-diagno-hpm9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): render-gap (stream-parser id collision) + replaceChild crash (sticky-frontier partition)
--- a/frontend/app/block/BlockErrorBoundary.tsx
+++ b/frontend/app/block/BlockErrorBoundary.tsx
@@ -23,6 +23,8 @@
 //     the existing cascade-detection warnings.
 
 import { invokeCommand } from "@/app/platform/ipc";
+import { resolveStack, resolveStackSync, type ResolveStatus } from "@/log/source-map-resolver";
+import { getTrail } from "@/log/render-trail";
 import { ErrorBoundary as SolidErrorBoundary, createSignal, Show } from "solid-js";
 import type { JSX } from "solid-js";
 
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
Codex's P2 is real but represents a fundamental design trade-off we can't fully close without a separate refactor:

- **Without `skipIds`**: live parser starts at `node_0` → collides with resumed-session snapshot → agent's response silently merged into the historical first node → **invisible to user**. This is the render-gap bug this PR fixes.
- **With `skipIds`**: live parser advances past snapshot's ids → resumed-session new responses render correctly. But on a reconnect that replays overlapping lines, the live parser produces fresh ids for lines `parseHistoryLines` would have given `node_0..N` → reducer can't dedup by id → potential duplicate transcript rows.

The trade-off favors the current fix:
1. The render-gap bug is **data-loss-visible** (response invisible). The reconnect-duplicate bug is **content-visible** (extra rows, user can still see all content).
2. The reconnect-overlap scenario requires the WPS broker to replay file events on resubscribe; whether that happens in production hasn't been verified in this session.
3. The user has confirmed the user-facing bug is fixed on the v0.38.18 portable.

Filing a follow-up (will link from discussion #1095) to properly close the dedup contract via content-based or timestamp-aware matching rather than counter-based ids. That's a larger refactor than this PR's scope.

Merging on reagent green.

PR #1107, opened 2026-05-27
chore: bump version to v0.39.1 (portable build)

AgentY-asaf:
Auto-bump from `task package` — the bump is required because the portable artifact carries its version in the binary + data-dir path, so v0.39.0 portable would collide with the in-tree v0.39.0 codebase.

This is the standard portable-release pattern: every `task package` advances the counter. The portable `agentmux-0.39.1-x64-portable.zip` (172 MB) will be uploaded to the v0.39.1 GitHub release after this merges.

Note: v0.39.0 release tag (created earlier) will be deleted — the v0.39.0 → v0.39.1 gap is a single `chore: bump` with no semantic content (no changesets), so the v0.39.0 tag would be confusing (release-with-no-asset vs release-with-asset). Replacing with v0.39.1.

Single `chore` commit, no semantic changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.39.0"
+version = "0.39.1"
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1108, opened 2026-05-27
fix(workspace): fade-in opacity reveal eliminates tab-switch FOUC flash

AgentA-asaf:
## Summary

- Adds a 120ms opacity fade on top of the existing `tab-reveal` visibility gate to eliminate the "negative-of-a-photo" flash users saw on tab switch with agent panes open.
- Lands the full investigation spec for the broader tab-switch perf work (`docs/specs/SPEC_AGENT_PANE_TAB_SWITCH_PERF_2026_05_27.md`), including negative-result log so future investigations don't re-chase the same dead ends.

## The flash

The existing `tab-reveal` gate uses `visibility: hidden` to suppress paint while the active tab settles. visibility is a binary flip — once the gate lifts the first paint frame can briefly expose un-cascaded theme colors / unstyled content before the agent pane settles, perceived as a negative-of-a-photo flash.

`opacity: 0 → 1` over 120ms ease-out blends those frames into invisibility. Both rules are keyed on the same `tabSwitching()` signal so they release together — visibility flips instantly when the gate lifts; opacity carries the perceived smoothness.

## Investigation spec — what we ruled out (preserved in spec §"Investigation Log")

The 500-600ms long-task on tab switch is browser-side layout/paint on the `display: none → block` transition, opaque to JS instrumentation. Negative results captured:

| Hypothesis | Result |
|---|---|
| Markdown render not memoized | Already memoized at `markdown.tsx:466`; 0 invocations during switches |
| Sticky-frontier partition slow | 0 slices >5ms |
| TanStack `measureElement` slow | 0.3ms total over 50 calls |
| `<For>` reconcile cascade | Only 3-15 row mounts per switch — not the cost |
| `content-visibility: auto` on rows | Measured 700-1090ms (worse than 500-600 baseline); reverted |

The perceived-smoothness fix lives at the reveal-gate layer, not in the document tree. Anyone resuming this work should start from the spec.

## Test plan

- [x] Smoke tested locally on portable v0.39.1 — negative-photo flash is gone, switching between 3 tabs with agent panes open feels clean.
- [ ] reagent + codex pass
- [ ] No regression in tab-switch latency (the visibility gate itself is unchanged; only opacity-fade overlay added)

### DIFF
--- a/.changesets/1779913634-fix-tab-switch-fade-in-opacity-reveal-eliminates-fouc-flash--fayj.md
+++ b/.changesets/1779913634-fix-tab-switch-fade-in-opacity-reveal-eliminates-fouc-flash--fayj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tab-switch): fade-in opacity reveal eliminates FOUC flash + investigation spec
--- a/docs/specs/SPEC_AGENT_PANE_TAB_SWITCH_PERF_2026_05_27.md
+++ b/docs/specs/SPEC_AGENT_PANE_TAB_SWITCH_PERF_2026_05_27.md
@@ -1,0 +1,270 @@
+# SPEC: Agent pane tab-switch perf
+
+**Date:** 2026-05-27
+**Author:** AgentA
+**Status:** Design — perf investigation + multi-phase remediation plan.
+
+---
+
+## The problem
+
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1113, opened 2026-05-28
feat(agents): agent_def_get reads from db_agents (Phase 3b.5a)

AgentA-asaf:
## Summary

Third true read-flip under Phase 3b ([SPEC_AGENT_ARCHITECTURE_2026_05_27.md](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md)). Sub-PR split because only one of the three originally-grouped 3b.5 targets is actually flippable:

| Sub-PR | Scope | Status |
|---|---|---|
| 3b.5a | `agent_def_get` → `db_agents` | ✅ this PR |
| 3b.5b | `user_clone_defs_for_template` → `db_agents` | ⏳ deferred (schema ambiguity + no live callers) |
| 3b.5c | `agent_def_delete_seeded` cascade-capture | ⏳ deferred (write-path helper) |

Tracking discussion: **#1095**.

## Why split

- **3b.5b** hits the same `is_template=0 AND parent_template_id=X` ambiguity that 3b.3a documented — user-clone projections and template-instance projections aren't schema-distinguishable. The function has zero live callers (test-only), so deferring is low-cost; it retires when a distinguishing column is added or the function goes away.
- **3b.5c** isn't a "data read" — it's grabbing cascaded INSTANCE ids before the legacy FK cascade fires so the dual-write can drop projections. Tied to the legacy cascade pattern by design; retires when Phase 3c drops that pattern.
- **3b.5a** is safe: sole caller (`template_promote` idempotency check) uses a deterministic id format (`template-promote-v1-<tpl_id>`) that collides with neither template ids nor instance UUIDs, so the schema-ambiguity is moot.

## Mapping

- Query: `FROM db_agents WHERE id = ?1`. Both seeded templates (`is_template=1`) and folded user-clone projections (`is_template=0`, keyed by def.id) surface.
- Column rename: `parent_id` → `parent_template_id` in the SELECT (the consolidated table uses the new name).

## Test plan

- [x] `cargo test -p agentmux-srv --release agent_def_get_phase_3b5` — 3/3 pass
- [x] `cargo test -p agentmux-srv --release agent_def` — 12/12 pass (no regression)
- [x] `cargo test -p agentmux-srv --release storage` — 157/157 pass
- [ ] reagent + codex pass


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Closing — codex P1 on the latest commit identified a fundamental data-shape divergence that makes a direct flip of `agent_def_get` to `db_agents` unsafe: the folded user-clone row in `db_agents` has `working_directory` / `identity_id` / `memory_id` / `github_context` overwritten by the most-recent instance binding (per the dual-write's `def.is_seeded == 0` branch). `template_promote`'s retry-reuse path would carry the instance-tinted cwd back into a promoted clone.

Per CLAUDE.md no-doc-only-PRs rule, the remaining diff after reverting the code flip is doc-only; better folded into a future PR. Spec entry will move 3b.5a from ✅ shipped to ⏳ deferred in the next data-layer PR (likely R.4 or 3c prep).

Learning captured for future 3b.5a attempts: either (a) keep def-shape columns separate from instance-binding overwrites in the dual-write, or (b) retire `AgentDefinition.working_directory` (and friends) before the flip.

PR #1115, opened 2026-05-28
fix(package-portable): refuse to wipe a running install

a5af:
## Summary

- `scripts/package-portable.sh:31` does an unconditional `rm -rf "\$PORTABLE"`. With no output-dir arg, that's `\$HOME/Desktop/agentmux-<version>-x64-portable/` — the path a portable user runs from.
- NTFS unlinks happily through mapped exe/dll files, so today this silently wiped a live install's on-disk assets while the process kept running on its mapped pages. The damage stayed invisible until the user opened a new window (~4½ hours later), at which point `resolve_frontend_base_url`'s `frontend/index.html` existence check failed, the fallback URL was `localhost:5173`, the new renderer aborted, and `on_render_process_terminated` looped 139k times in 22 minutes.
- This PR adds a pre-`rm` guard that uses PowerShell's `Get-Process` to detect any process whose `.Path` lives under `\$PORTABLE`. On hit, bail with a remediation hint pointing at the alternate-output-dir form.

Two independent follow-ups, called out in the retro and tracked as separate PRs:
1. `resolve_frontend_base_url` returns \`Result\` — never silently emit a dev URL in production when assets are missing.
2. Crash-budget on \`on_render_process_terminated\` — per-browser cap, then a terminal page instead of a loop (per the prime directive of \`SPEC_SERVICE_SUPERVISION_AND_RECOVERY_2026_05_20\`).

Full forensic write-up: \`docs/retro/retro-portable-rm-running-install-2026-05-28.md\`.

## Test plan

- [x] Positive: guard fires against the running 0.39.1 portable (PID 719104 detected).
- [x] Negative: guard does NOT fire against an idle older portable on the same Desktop.
- [ ] CI runs through the script on Linux/macOS where \`powershell.exe\` is not on PATH — guard is gated on \`command -v powershell.exe\` so it should be skipped cleanly. Verify in CI.

### DIFF
--- a/.changesets/1779969832-fix-package-portable-refuse-to-wipe-a-running-install-i5ub.md
+++ b/.changesets/1779969832-fix-package-portable-refuse-to-wipe-a-running-install-i5ub.md
@@ -1,0 +1,24 @@
+---
+type: patch
+---
+
+fix(package-portable): refuse to wipe a running install
+
+`scripts/package-portable.sh:31` did an unconditional `rm -rf "$PORTABLE"`.
+With no output-dir arg, `$PORTABLE` defaults to
+`$HOME/Desktop/agentmux-<version>-x64-portable/` — the canonical location a
+portable user runs from. NTFS unlinks happily through mapped exe/dll files,
+so the script silently deleted the on-disk assets of a live instance; the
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

Issue #1117, opened 2026-05-28
Renderer crash-loop hardening — follow-ups from 2026-05-28 incident


a5af:
Tracking the remaining work surfaced by the 2026-05-28 renderer crash-loop incident. PR #1115 (script guard) closed the original trigger; the items here are defense-in-depth + a pre-existing leak observation.

Retro: [`docs/retro/retro-portable-rm-running-install-2026-05-28.md`](https://github.com/agentmuxai/agentmux/blob/main/docs/retro/retro-portable-rm-running-install-2026-05-28.md)

### 1. `resolve_frontend_base_url` must not silently emit a dev URL in production
- **Where:** `agentmux-cef/src/commands/window.rs:967-999`
- **Problem:** When `frontend/index.html` is missing next to the exe in production, the function silently falls back to `http://localhost:5173` (the dev Vite URL) instead of erroring. This was the proximate cause of the renderer crash loop on 2026-05-28 — and is the *same fallback* that caused the prior `ANALYSIS_DEV_VITE_PORT_HARDCODE_2026-05-26.md` incident; that fix patched the dev branch only.
- **Fix sketch:** Change signature to `Result<String, FrontendUrlError>`. Callers in `window_pool.rs:196`, `window.rs:712`, `drag.rs:431`, `floating_pane.rs:209`, `client/mod.rs:1280` propagate the error and either render a built-in "assets missing" page or refuse to navigate. Recovery-page reload URL specifically should never be a network URL.
- **Risk:** Low. Happy path unchanged. Removes only the silent-fallback footgun.

### 2. Crash budget on `on_render_process_terminated`
- **Where:** `agentmux-cef/src/client/mod.rs:1247-1453`
- **Problem:** Handler unconditionally loads a fresh `data:` recovery page on every renderer death. No `crash_count`, no `last_crash_at`, no per-browser budget. If the recovery load itself triggers a new termination (it can — observed 139k+ times here), the loop is unbounded.
[...]

PR #1119, opened 2026-05-28
fix(cef): resolve_frontend_base_url returns Result (no silent localhost:5173 fallback in prod)

a5af:
## Summary

- `agentmux-cef/src/commands/window.rs`'s `resolve_frontend_base_url` historically fell back to `http://localhost:<vite_port>` when `frontend/index.html` was missing next to the exe. In production that points at nothing, the renderer aborts with `ERR_ABORTED`, and the recovery handler loops (139k crashes / 22 min on 2026-05-28).
- Change the signature to `Result<String, FrontendUrlError>` so the failure is explicit and unforgeable.
- Add `assets_missing_data_url(&err)` — a self-contained \`data:\` URL with a static "install is broken" page, no auto-reload, no link back into the broken install.
- All 5 callers (`open_window`, pool warmup, tab tear-off, floating pane, recovery handler) updated to handle the error. The recovery-handler case is special-cased: if the recovery itself can't resolve a URL, it loads the static page directly so the Reload button never targets a network URL that would crash again.

This is fix #1 from issue #1117 (the three-fix defense-in-depth chain). #2 (crash budget) and #3 (log rate-limit) coming next.

## Test plan

- [x] \`cargo check -p agentmux-cef\` clean (no new warnings).
- [ ] CI: full workspace check + tests.
- [ ] Smoke (manual): in a dev build, delete \`dist/cef-dev/runtime/frontend/index.html\`, open a new window, confirm the static "install broken" page renders instead of crashing.
- [ ] Smoke: same with a renderer-crash trigger (force-kill a renderer process from Task Manager) — recovery should now show either the normal recovery page (assets present) or the static install-broken page (assets gone); never a loop.

Related:
- Closes one item of #1117
- Retro: \`docs/retro/retro-portable-rm-running-install-2026-05-28.md\`
- Earlier analysis of the dev-side variant: \`docs/analyses/ANALYSIS_DEV_VITE_PORT_HARDCODE_2026-05-26.md\`

### DIFF
--- a/.changesets/1779970865-fix-cef-resolve-frontend-base-url-returns-result-instead-of--xq0k.md
+++ b/.changesets/1779970865-fix-cef-resolve-frontend-base-url-returns-result-instead-of--xq0k.md
@@ -1,0 +1,36 @@
+---
+type: patch
+---
+
+fix(cef): resolve_frontend_base_url returns Result instead of silently emitting localhost:5173 in production
+
+The function's no-frontend-bundle fallback historically returned
+`http://localhost:<vite_port>` — the dev Vite URL — when
+`frontend/index.html` was missing next to the exe. In production
+nothing is listening on that port; the new browser would abort with
+ERR_ABORTED, terminate its renderer, and trigger the unbounded crash
+loop that caused the 2026-05-28 incident (139k crashes in 22 min).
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

PR #1120, opened 2026-05-28
fix(cef): bound on_render_process_terminated with a crash budget (no infinite recovery loop)

a5af:
## Summary

- Handler in \`agentmux-cef/src/client/mod.rs\` had no per-browser crash budget. On 2026-05-28 a wedged Browser meant every recovery-page load itself triggered another renderer termination, looping the handler at ~108/sec for 22 min — 139k crashes, 884 MB log, UI thread CPU-pegged on synchronous log writes.
- Add per-browser ring of crash timestamps keyed by \`Browser::identifier()\`. After **3 crashes within 10 s**, log \`crash_loop_aborted\` and load a terminal "give up" page with only a Quit button — no Reload, no \`frame.load_url\` target, so it cannot re-enter the handler.
- History entry dropped in \`on_before_close\` so the map doesn't grow over a session.

Composes with #1119 (already merged): together, the 2026-05-28 incident becomes a single new-window failure with a clear error page — never a loop, never a degraded host process. Satisfies the prime directive of \`SPEC_SERVICE_SUPERVISION_AND_RECOVERY_2026_05_20\`: *"Bounded recovery — never an infinite restart loop."*

Closes one item of #1117.

## Test plan

- [x] \`cargo check -p agentmux-cef\` clean.
- [ ] CI workspace tests.
- [ ] Smoke (manual): from Task Manager, force-kill the same renderer process 4 times in <10 s; window should switch from the recovery page to the terminal "stopped recovering" page after the 4th kill, with no further \`renderer_terminated\` events in the log.
- [ ] Smoke (manual): with #1119 + this PR, simulate the 2026-05-28 trigger (delete \`runtime/frontend/\` while running, open a new window) — confirm the install-broken page renders once on the new window, no loop, log clean.

Related:
- Closes one item of #1117
- Composes with #1119 (merged)
- Retro: \`docs/retro/retro-portable-rm-running-install-2026-05-28.md\`
- Prime directive: \`docs/specs/SPEC_SERVICE_SUPERVISION_AND_RECOVERY_2026_05_20.md\`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1121, opened 2026-05-28
fix(cef): auto-close crash-loop terminal page so dead window releases its instance number

a5af:
## Summary

- When the crash budget (#1120) trips, the "Window stopped recovering" page sits until the user clicks Quit. During that wait the dead browser stays in the host's window registry, so the launcher's instance counter still includes it — UI shows "2 windows" even when only 1 is real. This was one of the user-visible symptoms of the 2026-05-28 incident.
- Add a 30s countdown to the terminal page that calls \`window.close()\` on expiry. \`window.close()\` triggers the existing \`on_before_close\` path → \`ReportWindowClosed\` → \`Event::WindowInstanceReleased\` → UI window count decrements. No new IPC protocol or host state.
- Any keystroke / mouse-down cancels the auto-close (countdown hides), so the message stays readable as long as the user wants. Listeners are single-shot.

This is the "decouple window count from window lifecycle" follow-up from issue #1117, implemented as the minimal viable fix (reuses existing cleanup chain rather than adding a new "instance abandoned" protocol).

## Test plan

- [x] \`cargo check -p agentmux-cef\` clean.
- [ ] Manual smoke: trip the crash budget (force-kill a renderer 4× in <10s), confirm terminal page renders with countdown, wait 30s, confirm window closes and host log shows \`Event::WindowInstanceReleased\` for that label.
- [ ] Manual smoke: trip budget, click anywhere within 30s, confirm countdown disappears and window stays open until manual close.

### DIFF
--- a/.changesets/1779971690-fix-cef-auto-close-crash-loop-terminal-page-so-the-dead-wind-amu1.md
+++ b/.changesets/1779971690-fix-cef-auto-close-crash-loop-terminal-page-so-the-dead-wind-amu1.md
@@ -1,0 +1,30 @@
+---
+type: patch
+---
+
+fix(cef): auto-close crash-loop terminal page so the dead window releases its instance number
+
+When `on_render_process_terminated` trips the crash budget (added in
+the prior PR), the resulting "Window stopped recovering" page sits
+indefinitely waiting for the user to click Quit. During that wait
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1123, opened 2026-05-28
fix(launcher,cef): anchor asset lookup on AGENTMUX_HOME env from launcher

a5af:
## Summary

- Windows's \`GetModuleFileName\` (called by \`std::env::current_exe\`) returns the .exe's *load-time* path forever, even after that dir is renamed or unlinked. That's how the 2026-05-28 incident stayed invisible to the running process and tripped the silent fallback URL.
- \`agentmux-launcher\` already walks from its own \`current_exe\` to find the runtime dir before spawning the host — that path is always correct because the launcher is freshly invoked. Export that resolved path as \`AGENTMUX_HOME\` on host spawn.
- \`agentmux-cef\`'s \`resolve_frontend_base_url\` (and \`RuntimeMode\` detection) now prefer \`AGENTMUX_HOME\` over \`current_exe().parent()\`. New \`resolve_host_runtime_dir()\` helper centralises the preference chain.
- \`current_exe()\` fallback preserved so \`task dev\` (Linux/macOS, host invoked directly) keeps working, standalone smoke tests stay green, and pre-AGENTMUX_HOME launcher builds degrade gracefully.

The previous PRs in this series (#1115 + #1119 + #1120 + #1121) catch the *symptom* of a wedged install. This one removes the *root architectural hazard*.

Closes one item of #1117. Composes with all prior PRs in the series.

## Test plan

- [x] \`cargo check -p agentmux-cef -p agentmux-launcher\` clean.
- [ ] CI workspace tests.
- [ ] Smoke: \`task dev\` on Windows still resolves the Vite URL correctly (AGENTMUX_HOME absent → falls back to current_exe path → RuntimeMode::Dev detected).
- [ ] Smoke: portable install still resolves frontend correctly (AGENTMUX_HOME present, points at runtime/).
- [ ] Smoke: simulate the 2026-05-28 trigger (move portable dir while running, open new window) — confirm AGENTMUX_HOME still points at original moved path? Or at the new path? (The launcher set it at startup, so it'll point at wherever the runtime was at launch. If runtime is *moved*, this protects against the case where the directory still exists at its old path; if runtime is *deleted entirely*, the prior PRs catch it.)

### DIFF
--- a/.changesets/1779971942-fix-launcher-cef-anchor-asset-lookup-on-agentmux-home-env-fr-yb0w.md
+++ b/.changesets/1779971942-fix-launcher-cef-anchor-asset-lookup-on-agentmux-home-env-fr-yb0w.md
@@ -1,0 +1,38 @@
+---
+type: patch
+---
+
+fix(launcher,cef): anchor asset lookup on AGENTMUX_HOME env from launcher instead of fragile current_exe()
+
+Windows's `GetModuleFileName` (called by `std::env::current_exe`)
+keeps returning the path the .exe was *originally loaded from*, even
+after that directory is renamed or unlinked. The 2026-05-28 incident
+exploited this: an external `rm -rf` of the running portable's
+directory left `current_exe()` pointing at an empty path, so
+`current_exe().parent().join("frontend/index.html")` returned ENOENT,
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1124, opened 2026-05-28
fix(cef): rate-limit renderer_terminated log on the crash target

a5af:
## Summary

- The 2026-05-28 incident wrote 884 MB of host log in 22 min (139k identical \`renderer_terminated\` lines on the \`crash\` target). Synchronous file writes on the UI thread starved the live renderer's IPC — the "input is hard" symptom.
- Per-browser crash budget from #1120 already caps worst-case to ~3 events / browser / 10s. **Defense in depth**: add a process-wide rate limit on the same event so a fan-out (many browsers crashing simultaneously) can't reproduce the log-volume failure even before per-browser budgets trip.
- Two \`static AtomicU64\`s track last-emitted timestamp + suppressed counter. Events within 100 ms of the last log are silently counted; the count is emitted as \`suppressed_since_last\` on the next un-throttled event, so no information is lost — only duplicate volume.

Intentionally NOT a full custom tracing Layer: in-place check is ~15 lines, zero allocation per crash, targeted at the only event known to spam. A generic rate-limit Layer would be nicer ergonomically but solves nothing the per-browser budget + this in-place throttle don't already.

Closes the last actionable item of #1117.

## Test plan

- [x] \`cargo check -p agentmux-cef\` clean.
- [ ] Smoke: induce repeated renderer crashes (force-kill from Task Manager), confirm \`grep renderer_terminated\` shows at most ~10 lines/second per process (vs the 108/sec of the original incident), and that the \`suppressed_since_last\` field is populated on lines that follow a throttled gap.

### DIFF
--- a/.changesets/1779972239-fix-cef-rate-limit-renderer-terminated-log-on-the-crash-targ-69mz.md
+++ b/.changesets/1779972239-fix-cef-rate-limit-renderer-terminated-log-on-the-crash-targ-69mz.md
@@ -1,0 +1,32 @@
+---
+type: patch
+---
+
+fix(cef): rate-limit renderer_terminated log on the crash target (100ms gap, suppressed_count rolled forward)
+
+The 2026-05-28 incident wrote 884 MB of host log in 22 minutes —
+139,205 identical `renderer_terminated` lines on the `crash` target.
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1112, opened 2026-05-28
feat(floating-pane): MVP re-dock — drop floater over agentmux window

AgentX-asaf:
## Summary
- Implements floating-pane re-dock (Phase 4a of #1079): drag a torn-off floater back over an agentmux window and drop to redock as a docked pane.
- Backend `RedockFloatingPane` saga handles cross-tab/cross-window MoveBlock with target-side layout sync via `pendingbackendactions`.
- Host adds Win32 Z-order cursor resolution (skipping the source floater) plus hover IPCs that drive a ghost drop-slot preview in the target window.
- Frontend renders per-tile drop placeholders using `determineDropDirection` (same affordance as within-window pane drag), JS-driven floater drag (window follows cursor), and an auto-close watcher when the floater's tab empties.

## What's in
**Backend (agentmux-srv):**
- New `sagas/redock_floating_pane.rs` — validates source/target tabs+block, executes MoveBlock with optional dst_index. Unit tests for happy path, same-tab rejection, missing-block rejection.
- New `workspace.RedockFloatingPane` RPC in `service.rs`.
- `queue_target_layout_insert` helper appends an `"insert"` action to the target layout's `pendingbackendactions` (LayoutModel doesn't sync from external WaveObj writes — the queue is the working pattern).
- Explicit `event_bus.broadcast_event` for `waveobj:update` to reach the target renderer.

**Host (agentmux-cef):**
- `resolve_window_at_cursor` — Win32 Z-order walk via `GetTopWindow`+`GetWindow(GW_HWNDNEXT)`, skips source floater via `exclude_label`.
- `update_floating_redock_hover` / `clear_floating_redock_hover` — emit `floating-redock:hover-state` events with cursor + target label.
- `resolve_window_hwnd` reordered: cache lookup first (no GA_ROOT walk on cache hits), then reducer registry + GA_ROOT, then class-aware EnumWindows fallback.
- `capture_hwnd_for_label`: early-return if label already registered (preserves floater's outer HWND).
- `floating_pane_wndproc`: WS_POPUP + WS_THICKFRAME, no caption; DwmExtendFrameIntoClientArea; WM_NCCALCSIZE→0, WM_NCACTIVATE→1, edge-resize hit testing.
- `find_main_window`: EnumWindows skipping the `AgentMuxFloatingPane` class.

**Frontend:**
[...]

PR #315, opened 2026-04-08
feat: double-click titlebar to maximize/restore

AgentX-asaf:
## Summary
- Add `dblclick` listener to `useWindowDrag.win32.ts` that toggles maximize/restore when double-clicking the window header drag region
- Reuses existing `isInDragRegion()` check so tabs, buttons, and other interactive elements are excluded
- Cancels in-progress drag state on double-click to prevent conflicts
- Linux already works natively via CEF Views compositor — no change needed
- macOS implementation TBD (separate PR)

## Test plan
- [ ] Double-click empty space in window header → maximizes
- [ ] Double-click maximized header → restores to previous size/position
- [ ] Double-click on a tab → does NOT maximize
- [ ] Double-click on window control buttons → does NOT maximize
- [ ] Single click + drag → still drags window normally

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1132, opened 2026-05-28
feat(floating-pane): resize + maximize support

Agent2-asaf:
## Summary

Hand-rolled resize and maximize for the floating-pane WS_POPUP + WS_EX_TOOLWINDOW window. CEF Views isn't available for floaters (the no-taskbar / owned-cascade requirement forces native Win32 — see `SPEC_FLOATING_PANE_TEAROFF_2026_05_11 §10`), so the wndproc carries the load.

Spec: `docs/specs/SPEC_FLOATING_PANE_RESIZE_AND_MAXIMIZE_2026-05-28.md`

## What changes

**Resize (Win32 wndproc)** — `agentmux-cef/src/floating_pane.rs`
- `WM_GETMINMAXINFO`: 320×180 CSS-px min size + maximize work-area clamp (so maximize respects the taskbar instead of going full-monitor like raw WS_POPUP defaults to).
- `WM_SIZE`: track the embedded CEF `WS_CHILD` onto the outer's client size. Children of WS_POPUP do not auto-track parent size, so without this the rendered content stayed at the initial (W, H).
- `WM_NCHITTEST` band 6 → 8 CSS px for discoverability while staying clear of header buttons.

**Maximize (label-routed)** — `agentmux-cef/src/commands/window.rs`
- `maximize_window` now routes via `resolve_window_hwnd` (PR #1094's `window_hwnds` cache pattern) instead of `find_own_top_level_window` — fixes the long-standing bug where a label-less call from main accidentally hit the floater (owned windows draw above their owner, so Z-order EnumWindows returned the floater first).
- For `floating-…` labels, stashes the pre-maximize outer rect into a new `state.floating_restored_rects` map so the post-restore origin can be replayed atomically by `restore_window_and_move` (drag-from-maximized) or implicitly by the next toggle.
- Returns `{state}` so the FE can update its icon glyph without a follow-up `get_window_state` query.

**New IPCs**
- `set_window_size({label, w, h})` — companion to `set_window_position`; CSS px, label-routed.
- `get_window_state({label})` — returns `"normal" | "maximized" | "minimized"` from `GetWindowPlacement`.
- `consume_restored_rect({label})` — drains the floater's pre-maximize rect; used by drag-from-maximized.
- `restore_window_and_move({label, x, y, w, h})` — atomic `SW_RESTORE` + `SetWindowPos` to avoid a flash of the old maximized origin mid-drag.

**Frontend** — `frontend/app/`
- New `store/floating-pane-context.ts`: a process-wide signal holding `{windowLabel, state}` when the renderer is the floating shell. Null elsewhere, so docked panes are unaffected.
- `block/blockframe.tsx` EndIcons: maximize/restore button immediately left of close-X, only when the floating context is non-null. Toggle reuses `maximize_window`'s return value to swap the glyph.
- `workspace/floating-pane-workspace.tsx`:
  - Seed the floating context on mount from `get_window_state`; clear on cleanup.
  - **dblclick listener** on the pane header (mirrors PR #315's main-window pattern); suppresses the bubble-phase magnify-toggle since magnify on a single-pane floater is meaningless.
  - **Drag-from-maximized branch** in mousedown — Windows-explorer-style: dragging a maximized window restores it under the cursor and continues the drag. Uses `consume_restored_rect` + `restore_window_and_move` so there's no IPC flash between the SW_RESTORE and the move.

## Why this approach

History (full table in spec §0):
- Secondary windows switched to CEF Views (PR #263) because raw `browser_host_create_browser` had no resize. But that lesson doesn't transfer — floaters need `WS_EX_TOOLWINDOW + owner` semantics CEF Views doesn't expose, so the parent spec chose the native popup.
- PR #1082 + PR #271 established the proven `WM_NCHITTEST` + `WM_NCCALCSIZE=0` + `DwmExtendFrameIntoClientArea` pattern for resize on the native popup. We extend that pattern; we do not replace it.
- PR #1089 dropped WS_CAPTION, which means we cannot add WS_MAXIMIZEBOX (the standard system maximize button). All maximize affordance is rendered by the frontend inside `BlockFrame_Header`.
- PR #315 wired dblclick maximize for the main window; we mirror that shape for the floater.
- `SPEC_MAXIMIZE_ZOOM_ARCHITECTURE §3.P5` flags missing window-state events. This PR deliberately **does not** add a `window_state_changed` event channel — `maximize_window`'s return value covers user-initiated toggles (button + dblclick). Win+Up / programmatic ShowWindow paths from outside the IPC are a follow-up; the spec calls them out.

## Caveats / things to validate

- **I cannot compile or run this in the container.** Windows-only host code, container is Linux, no cargo. The Rust changes follow the proven `install_frameless_resize_hook` template in `wndproc.rs:64-136` and the `set_window_position` label-routing template in `commands/window.rs:388-422` — but a `task dev` smoke is the first real verification.
- The Linux/macOS branch of `maximize_window` still routes through `crate::ui_tasks::post_maximize_window` (unchanged) but no longer returns `Value::Null` — it now returns `{state: "normal"}` as a placeholder. Floating-pane is Windows-only, so cross-platform maximize behavior shouldn't regress, but worth a quick check on `task dev` on each OS.
- The icon names `window-maximize` / `window-restore` are already in use elsewhere (`command-registry.ts:207`, `tabbar.tsx:658`), so no new icon assets needed.

## Test plan

- [ ] Resize floater from each edge / corner; cursor changes; CEF child follows outer 1:1; cannot shrink below 320×180 CSS px.
- [ ] Click maximize button in header → floater fills monitor work area (taskbar visible). Icon swaps to restore.
- [ ] Click restore → floater returns to pre-maximize rect.
- [ ] Double-click pane header (non-interactive zone) toggles maximize. Magnify does NOT also toggle.
- [ ] Drag a maximized floater's header → floater restores under cursor and follows mouse. Release over main window → redocks per PR #1112.
- [ ] Resize floater on a 100% monitor, drag to a 200% monitor, maximize → fills 200% monitor's work area (not stretched 2×).
- [ ] Open two floaters, maximize one, drag/move/resize the other — independent. `floating_restored_rects` entries don't collide.
- [ ] Close a maximized floater — no leaked `floating_restored_rects` entry (labels are UUID-tagged so collision would never matter, but worth confirming).
- [ ] Main-window dblclick maximize still works (the existing `useWindowDrag.win32.ts` path is untouched).

## Follow-ups deferred from spec

- `window_state_changed` event channel (for keyboard Win+Up / programmatic maximize paths that bypass the IPC).
- Header maximize-button overflow at narrow widths (collapse into existing overflow menu — see `SPEC_AGENT_PANE_BOTTOM_BUTTONS_2026_04_22.md` precedent).
- Optional: docked-pane "maximize tile within window" affordance (separate spec interacting with TileLayout).


a5af:
@codex review

a5af:
@codex review

Addressed:

- **Codex P2 `floating_pane.rs:398`** (ptMaxPosition off-screen): fixed. `ptMaxPosition` is now the work-area origin directly (`work.left`, `work.top`), not a delta from `wr`. The old delta math placed the maximized window off-screen for any floater not already at the work-area origin.
- **Codex P2 `floating-pane-workspace.tsx:205`** (restore on every mousedown): fixed via a 4-CSS-px movement threshold matching Windows `SM_CXDRAG`. `pendingMaximizedRestore` is armed on mousedown over a maximized header, cleared on mouseup, and triggers the restore-under-cursor dance only when mousemove crosses the threshold. Plain clicks and the mousedown half of a dblclick fall through harmlessly.
- **ReAgent P2 `floating-pane-workspace.tsx:228`** (DPI-scaled offset): fixed. The `16` cursor-below-top-edge offset is now `Math.round(16 * dpr)`. At 200% DPI it stays at 16 CSS px instead of collapsing to 8 and hitting the top resize band.
- **ReAgent P2 `commands/window.rs` `maximize_window` null label**: fixed. The Windows branch now returns `Err("...label not resolved to a window")` instead of silently falling through to the cross-platform `Ok({state:"normal"})` default. The frontend will keep its old glyph and surface the failure.
- **ReAgent P2 `commands/window.rs` `set_window_size` speculative API**: left in place. The PR body explicitly cites spec §5.2 as the deferred-use rationale, and the IPC + Rust handler are mirror-images of the `set_window_position` template — removing them now would force the next FE-side PR to add them back. Happy to drop them if the team prefers no dead APIs at all.

a5af:
@codex review

Round 2 of fixes pushed in c7bd5759:

- floating_pane.rs:407 ptMaxPosition is now work.{left,top} minus mon.{left,top} (work-area offset within its monitor). Per Raymond Chen, Windows treats ptMaxPosition as primary-monitor-relative and adjusts by the target offset, so absolute coords get offset twice on non-primary monitors. The new form computes (0, 0) on a primary monitor with bottom taskbar, and (0, 0) on a non-primary with the same layout — the OS then adds the target monitor offset.

- commands/window.rs maximize_window label cache strict: for floating- labels we now consult state.window_hwnds directly and treat cache miss as null, skipping resolve_window_hwnd's fallback to find_own_top_level_window. Stale floater labels now surface as the Err path so the FE keeps the old glyph. Non-floater labels keep the prior resolution chain.

- Floating-pane resize regression from user smoke test: CEF child was filling the entire outer client area, so the OS dispatched WM_NCHITTEST to the child (returns HTCLIENT) and never bubbled to the outer wndproc. Inset the child by resize_border_px (8 CSS px) in WM_SIZE. The inset matches the hit-band width and is visually indistinguishable from a system resize border.

a5af:
@codex review

Round 3 pushed in 87e16a6e + merged main (6a454e8a):

- commands/window.rs:173 IsWindow validation for floater cache lookups. After the cache hit, `IsWindow(hwnd) != 0` check rejects destroyed HWNDs so the Err path fires instead of toggling a zombie. PR #1133 (already on main) adds eviction on close; this is defense in depth against the race window between close and cleanup.

- floating_pane.rs initial-creation inset. The CEF child is now created with `(inset, inset, w - 2*inset, h - 2*inset)` rect, using the same DPI-scaled `RESIZE_BORDER_CSS` math as WM_NCHITTEST. The first WM_SIZE on the outer fires from inside create_owned_popup's ShowWindow before the child exists, so the WM_SIZE inset path no-ops on initial open — applying the inset at creation closes that gap.

Plus merged in the 5 commits that landed on main since the branch opened (including PR #1133 which is the complementary cache-eviction-on-close fix).

a5af:
@codex review

ebf015a5 addresses the one open P2 (ReAgent, on 6d471091): the WM_NCHITTEST forwarder's ORIGINAL_WNDPROCS map is now pruned on the child's WM_NCDESTROY — forward to the original wndproc for CEF teardown, then remove our entry, so a recycled HWND value can't short-circuit the installer and silently break resize on a future floater.

Note: the magenta-debug-paint P0 that was flagged earlier was on the superseded diagnostic commit c15392da; the subclass commit 6d471091 already removed it (no WM_ERASEBKGND paint, no per-paint logging, no WM_NCPAINT suppression remain). The only floating-pane-diag line left is a one-time install log per floater.

a5af:
## Handoff notes — state + open investigation (AgentY)

Parking this PR open. Summary of where it stands so it can be picked up cleanly.

### Resolved
- **White border around floater** — fixed via the WM_NCHITTEST forwarder approach (`6d471091`): the CEF child fills the window and returns `HTTRANSPARENT` on edge bands so the outer wndproc handles resize. No inset strip → no DWM-overpainted white border.
- **Floater resize** — works (edge bands resize via the forwarder).
- **ptMaxPosition off-screen on non-primary monitor** — fixed (`work - mon` offset).
- **Magenta debug paint P0** (ReAgent 22:20) — was on the *superseded* diagnostic commit `c15392da`; `6d471091` already removed all of it (no WM_ERASEBKGND paint, no per-paint logging, no WM_NCPAINT suppression). The only `floating-pane-diag` line left is a one-time install log per floater.
- **ORIGINAL_WNDPROCS map never pruned P2** (ReAgent 22:28) — fixed in `ebf015a5`: the forwarder now handles `WM_NCDESTROY`, forwards to the original wndproc for CEF teardown, then removes its own map entry, so a recycled HWND can't short-circuit the installer.

### Open — two runtime regressions, NOT yet root-caused
1. **Parent (main) window drag stopped working** in the dev build while a floater is open. Dev log showed **zero** `set_window_position` IPC fires on main-drag attempts.
2. **Redock ghosts (hover placeholder) not showing** during a floater drag-over-main.

### Investigation finding — the diff doesn't explain it
Walked `origin/main...HEAD`:
- `useWindowDrag.win32.ts` (main-window drag hook) and `app-init.ts` (redock-hover listener) are **untouched** by this PR — byte-identical to main (both from #1112).
- `resolve_window_at_cursor`, `find_main_window`, and the redock-hover host handlers in `commands/window.rs` are **untouched**. #1132 only changed `maximize_window` to route via `resolve_window_hwnd`, and added 4 new fns (`set_window_size`, `get_window_state`, `restore_window_and_move`, `consume_restored_rect`).
- `floating-pane-workspace.tsx` changes are **purely additive** — they only alter behavior when the floater is *maximized*; a normal floater drag falls straight through to the original redock path and still calls `pushRedockHover`.
- The HTTRANSPARENT forwarder is installed **only on the floater's own CEF child** (`GW_CHILD`).

So statically, #1132 has no obvious mechanism to break normal docking/ghosts or main-window drag.

### Prime suspect for pickup
Runtime mouse-routing during cross-window drag. Specifically: **if `resolve_window_at_cursor` (or any redock target-finding) uses `WindowFromPoint`, that API *skips windows that return HTTRANSPARENT* to `WM_NCHITTEST`** — so the new forwarder could perturb which window is found under the cursor during a drag near a floater edge. Worth checking whether redock resolution uses `WindowFromPoint` vs `EnumWindows` Z-order (the latter is unaffected by HTTRANSPARENT).

### Decisive next step (settles the merge question)
A/B test: **does redock + ghosts work on plain `main` (no #1132)?**
- Broken on main too → #1132 didn't cause it; merge is purely additive.
- Works on main, breaks on branch → real regression; hunt the HTTRANSPARENT-vs-`WindowFromPoint` interaction before merging.

Also re the parent-drag zero-IPC observation: rule out a dev-mode artifact first — in `task dev` the main window can be served from a pool window whose URL carries `windowLabel=window-pool-…`, so `useWindowDrag`'s `ownWindowLabel()` may target a label the host can't resolve for `set_window_position`. May not reproduce in a portable build.

Branch HEAD: `ebf015a5`. ReAgent will re-review on it; the only finding it had on the prior commit (the map-pruning P2) is addressed.

a5af:
Closing — stale base, P0 version downgrade from stale rebase. Reopen rebased on current main.

PR #1133, opened 2026-05-28
fix(cef): evict stale HWND from window_hwnds cache + cleanup on close

AgentA-asaf:
## Summary

Title-bar close button silently broken on v0.39.1 portable: clicking the X resolved an HWND from `state.window_hwnds`, posted `WM_CLOSE` to it, and the window stayed alive. Diagnosis showed the cache was pointing at a dead HWND — CEF Views can swap the main window's outer HWND on recreate, but the cache has no eviction path.

This PR fixes the surface bug and ships the analysis + spec docs the user asked for.

## Evidence

Live host log on v0.39.1, two close clicks ~158ms apart:

```
2026-05-28T17:16:17.366Z [win-resolve] resolved via window_hwnds cache cache_hwnd=0x2b03e4
2026-05-28T17:16:17.524Z [win-resolve] resolved via window_hwnds cache cache_hwnd=0x2b03e4
```

Live process check via PowerShell: actual main window HWND is `0x1404f0` (PID 3404, title "Starter workspace - tab3 - AgentMux"). `PostMessage(0x2b03e4, WM_CLOSE)` posted into a dead window. `on_before_close` never fires, no further log entries.

## Fix — belt + suspenders

1. **`resolve_window_hwnd` validates cache liveness** (`window.rs:235`): wrap the cache hit in `IsWindow(hwnd)`. If stale, evict and fall through to the registry / `EnumWindows` fallback.
2. **`on_before_close` evicts on clean close** (`client/mod.rs:637`): after the existing `UnregisterBrowser` dispatch, remove the closing browser's label from `window_hwnds` so a subsequent open re-captures cleanly.

#1 catches edge cases where `on_before_close` doesn't fire (crash, CEF lifecycle quirks). #2 keeps the cache clean on the happy path so #1 rarely activates.

## Docs

Two docs included per CLAUDE.md "no doc-only PRs" rule — they ride with the code:

- `docs/specs/SPEC_WINDOW_HWND_CACHE_STALE_FIX_2026_05_28.md` — implementation spec for the fix above.
- `docs/analysis/ANALYSIS_LARGE_FILE_MODULARIZATION_CANDIDATES_2026_05_28.md` — survey of fat source files that could benefit from the R-phase modularization treatment applied to store.rs today. `window.rs` (user-flagged) gets a concrete 6-sub-PR carve plan inside. Priority list: R.1 store.rs agents → window.rs → agent_handlers.rs → client/mod.rs.

## Test plan

- [x] `cargo check -p agentmux-cef --release` — clean (36 pre-existing warnings, no new).
- [x] Fresh portable v0.39.3 built and ready on Desktop for manual smoke (close button on title bar should now work end-to-end on second click after a recreate).
- [ ] Manual smoke: open main → drag a tab into a tear-off (forces a recreate path) → click X on main. Window should close on first click.
- [ ] reagent + codex.

## Why no automated test

The bug only surfaces when CEF Views recreates the outer HWND, which we can't reproduce in a unit test without a real Win32 message loop. The fix is defensive and small; `IsWindow` is a sub-microsecond syscall vs the IPC round-trip we're already inside.

### DIFF
--- a/.changesets/1779989590-fix-cef-evict-stale-hwnd-from-window-hwnds-cache-on-before-c-vq9p.md
+++ b/.changesets/1779989590-fix-cef-evict-stale-hwnd-from-window-hwnds-cache-on-before-c-vq9p.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): evict stale HWND from window_hwnds cache + on_before_close cleanup
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.39.1"
+version = "0.39.2"
 dependencies = [
  "anyhow",
  "base64",
@@ -28,7 +28,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.39.1"
+version = "0.39.2"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -58,7 +58,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.39.1"
+version = "0.39.2"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #997, opened 2026-05-23
refactor(agent-pane): drop legacy turnActive/stopping/streaming.active (turn-phase PR G)

AgentA-asaf:
**Final PR in the turn-phase migration series.** Drops the legacy `turnActive` / `stopping` (top-level on `AgentPaneState`) and `active` (on `StreamingState`) booleans that PR A (#987) introduced as a dual-write and PR B (#992) migrated every view consumer off.

## Scope

**Pure cleanup. No user-visible behaviour change.**

Every view and hook now reads:
- `isWorking(state)` / `workingFromPhase(turnPhase)` for the "working" animation gate
- `state.turnPhase.kind === "Interrupting"` for the "Stopping…" label
- `isDisconnected(state)` for the disconnect banner (PR F)

The reducer-internal "is the stream subscribed?" gate (previously `state.streaming.active`) is now `state.lastEventMs !== null` — both fields always moved in lockstep (only `StreamSubscribe` / `StreamUnsubscribe` touched either), so the substitution is mechanical. A new `isStreamSubscribed(state)` selector wraps the check.

## Files touched (12)

- `frontend/app/store/agent-pane-state/types.ts` — drop `turnActive`, `stopping` from `AgentPaneState`; drop `active` from `StreamingState`; add `isStreamSubscribed` selector
- `frontend/app/store/agent-pane-state/reducer.ts` — every arm stops writing the legacy fields; subscription gate switches to `lastEventMs !== null`
- `frontend/app/store/agent-pane-state/reducer.test.ts` — assertions migrated to `isWorking(state)` / `state.turnPhase.kind` / `state.lastEventMs`
- `frontend/app/store/agent-pane-state-store.ts` — drop `turnActive` / `stopping` projection setters
- `frontend/app/store/agent-pane-state-store.test.ts` — drop the now-impossible projection setters from `noopProj()`
- `frontend/app/view/agent/state.ts` — drop `turnActiveAtom`, `stoppingAtom`; reshape `streamingStateAtom` initial value
- `frontend/app/view/agent/state.test.ts` — assertions migrated to `turnPhaseAtom`
- `frontend/app/view/agent/types.ts` — drop `active` from `StreamingState`
- `frontend/app/view/agent/agent-view.tsx` — drop legacy atoms from `registerAgentPaneStatePane` projection bundle and from `useAgentStream` / `useAgentCommands` options; pass `turnPhaseAtom` instead
[...]

PR #992, opened 2026-05-23
feat(agent-pane): view reads isWorking(state) (turn-phase PR B)

AgentA-asaf:
## PR B of the agent-pane turn-phase migration

Follow-up to #987 (PR A — TurnPhase discriminated union + `isWorking(state)` selector + dual-write logic).

This PR migrates the agent VIEW layer off the legacy `turnActive` / `stopping` / `streaming.active` boolean reads onto the new single-source-of-truth phase. The reducer continues to dual-write the legacy fields; **PR G will drop them once every consumer is migrated.**

### View migration
Three view-layer reads in `frontend/app/view/agent/agent-view.tsx` switched off legacy fields:

| Location | Before | After |
| --- | --- | --- |
| `<AgentStatusLine loading={...}>` (the **working animation binding** the user identified as broken) | `status.isLoading() \|\| turnActive \|\| stopping` | `status.isLoading() \|\| workingFromPhase(turnPhase)` |
| `<AgentStatusLine stopping={...}>` (the "Stopping…" label override) | `stoppingAtom[0]()` | `turnPhaseAtom[0]().kind === "Interrupting"` |
| `<PendingMessagesPanel showSendNow={...}>` ("Send now" button visibility) | `turnActiveAtom[0]() && pending.length > 0` | `workingFromPhase(turnPhase) && pending.length > 0` |

### Plumbing
- `AgentAtoms` gains `turnPhaseAtom: SignalPair<TurnPhase>` (defaults to `{ kind: "Idle" }`).
- `AgentPaneProjections` gains an optional `turnPhase` setter; the store calls it on every dispatch alongside the existing legacy setters (`turnActive`, `stopping`, etc.).
- `agent-view.tsx` wires the new projection into the slot registration.
- New `workingFromPhase(phase: TurnPhase): boolean` exported from `agent-pane-state/types.ts`. View code only has the phase signal, not the full reducer state, so this avoids forcing callers to fabricate a `{turnPhase}`-shaped `AgentPaneState`. `isWorking(state)` now delegates to it.

### Tests
- `frontend/app/view/agent/state.test.ts` — three new tests verifying `turnPhaseAtom` defaults to Idle, `workingFromPhase` returns true only for Submitting/Streaming/Interrupting, and the Interrupting-as-stopping-label predicate flips correctly.
[...]

PR #1139, opened 2026-05-28
fix(agent-pane): suppress Send-now flash by excluding Submitting

AgentA-asaf:
## Summary

The pending-queue "Send now" button flickered for ~50–200 ms on every send-message from an agent pane, including from a fully idle pane. Root cause: the predicate at `agent-view.tsx:819` gated on `workingFromPhase(turnPhase)` which returns true for `{Submitting, Streaming, Interrupting}` — but during `Submitting` the message you just typed is itself in the pending queue waiting for `agent-message-accepted`. Both predicates were briefly true on every send → button paints → ~50–200 ms later it hides.

## Fix

- Add `isInterruptibleTurn(phase: TurnPhase): boolean` next to `workingFromPhase` — same shape, strict subset, true only for `{Streaming, Interrupting}`.
- Swap the predicate at the single call site. Inline comment updated.
- Three unit-test cases pinning the new predicate, including an anti-regression test asserting the exact `Submitting` divergence point between the two helpers.

Behavior can only go "button visible → button hidden", never the reverse. The queued-zone itself (amber border, spinner, message list) is unchanged.

## Test plan

- [x] `npx vitest run frontend/app/view/agent/state.test.ts` — 9/9 pass (3 new).
- [ ] Smoke: idle pane, type a message, hit Enter; no "Send now" button paints during the Submitting window.
- [ ] Smoke: start a long Streaming turn, type a second message; "Send now" appears, click it → first turn stops, queued message processes.

## Background

- PRs #495 (Send-now intro), #992 (turn-phase B switched the predicate to `workingFromPhase`), #997 (PR G removed legacy `turnActive`).
- Diagnosis: `docs/analysis/ANALYSIS_SEND_NOW_FLASH_2026_05_28.md` (in the parallel diagnostic branch — not in this PR).

### DIFF
--- a/.changesets/1780001472-fix-agent-pane-suppress-send-now-flash-by-excluding-submitti-3rgt.md
+++ b/.changesets/1780001472-fix-agent-pane-suppress-send-now-flash-by-excluding-submitti-3rgt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): suppress Send-now flash by excluding Submitting from the predicate
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.39.2"
+version = "0.39.3"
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #1140, opened 2026-05-28
fix(agent-pane): collapse 3 hover events into one auto-expand panel

AgentA-asaf:
## Summary

User reported three overlapping visual events on tool-row hover: small browser-native \`title=\` tooltip showing summary (which includes \`(N.Ns)\` — what they perceived as a "time popup"), larger log panel, and a fast expand/collapse animation. All three together = flicker every cursor pass.

This PR collapses them into a single in-flow panel that opens only when there's reason to: tool is \`running\` / \`pending_approval\`, or the user has clicked to pin. **Hover is no longer a trigger.**

## Behavior change

| Trigger | Before | After |
|---|---|---|
| Hover completed tool | title tooltip + 150 ms then panel expands | Nothing |
| Tool starts running | Panel auto-expands | Same |
| Tool completes | Stays open 3 s post-hold | Same |
| Click row | Pins panel | Same |
| Hover running tool | Panel stays open (auto) | Same |

## Header dedup

Old overlay header: status icon · tool name · summary · duration · status label — every field except the status label was already on the collapsed row above it. New header: \`HH:MM:SS\` (from \`ToolNode.timestamp\`) · status label. Two slots, zero duplication.

## What was deleted

- \`ToolBlock.tsx\`: \`hovering\` signal, \`expandDirection\`, \`overlayMaxHeight\`, enter timer, hover-anchor imports, \`onMouseEnter\`/\`onMouseLeave\` on both root + panel divs, \`title={summary}\` on \`.agent-tool-name\`.
- \`ToolBlockOverlay.tsx\`: status icon / tool / summary / duration spans in the header.
- SCSS: \`.agent-tool-panel--overlay-above\`, \`--overlay-below\`, and the 4 deleted header sub-classes.

## What was preserved

- Auto-expand for \`running\` / \`pending_approval\`.
- 3 s \`POST_COMPLETION_HOLD_MS\`.
- Click-to-pin.
- \`UserMessageBlock\` hover-anchor (untouched — still uses \`hover-anchor.ts\`).
- The action bar, log body rendering, scroll-stick.

## Test plan

- [x] \`npx vitest run frontend/app/view/agent/components/ToolBlock.test.tsx\` — 7/7 pass, including:
  - Anti-regression: \`mouseenter\` on a completed tool + 1 s fake-timer advance → panel stays \`--hidden\`, no overlay class, no inline max-height. Guards against the 150 ms hover timer creeping back.
  - \`mouseenter\` on a running tool → still \`--flow\`. Guards against an over-eager removal also killing auto-expand.
- [ ] Smoke: hover a completed Bash tool — nothing visible happens. Click — panel opens with timestamp + status label in header. Click — collapses. Send a new message that runs a tool — auto-expands; status flips to success; 3 s later collapses; hover after that → nothing.

## Spec + diagnosis

\`docs/specs/SPEC_TOOL_HOVER_CONSOLIDATION_2026_05_28.md\` is in this PR — full diagnosis, the three observed events mapped to source lines, the implementation rules each section of the diff was checked against.

## Out of scope

- Discoverability of click-to-pin: only the \`cursor: pointer\` style remains. Noted in spec §5 as a possible follow-up.
- AI Chat Panel (\`aitooluse.tsx\`) has analogous hover behavior — separate PR if prioritized.

### DIFF
--- a/.changesets/1780002518-fix-agent-pane-collapse-3-hover-events-into-one-auto-expand--p5w5.md
+++ b/.changesets/1780002518-fix-agent-pane-collapse-3-hover-events-into-one-auto-expand--p5w5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): collapse 3 hover events into one auto-expand panel
--- a/docs/specs/SPEC_TOOL_HOVER_CONSOLIDATION_2026_05_28.md
+++ b/docs/specs/SPEC_TOOL_HOVER_CONSOLIDATION_2026_05_28.md
@@ -1,0 +1,203 @@
+# SPEC: Tool-hover consolidation (2026-05-28)
+
+**Author:** AgentA
+**Reporter:** user (this session) — "in the agent pane, there appears to be 3 separate events happening on a tool hover, a small time popup, a larger log popup, and a fast expand/collapse. There should be NO expand/collapse on hover. It only is supposed to happen on new messages. We want the small popup (it's just the time) to be at the top of the larger popup, so it's just 1 popup. Also, in the large log popup, we repeat the same info in the line and in the popup, lets get rid of the dupe."
+**Affected files (primary):**
+- `frontend/app/view/agent/components/ToolBlock.tsx` (335 LOC)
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #1142, opened 2026-05-29
diag(statusbar): instrument sysinfo channel to bisect agent-pane cascade freeze

AgentA-asaf:
## What

Diagnostic instrumentation (no behavior change) for the agent-pane cascade crash whose side effect froze **all** status-bar perf indicators (CPU/Mem/Net/Disk/uptime) until a sysinfo widget pane was opened — which re-armed the broker subscription. See \`docs/analysis/ANALYSIS_AGENT_PANE_CRASH_2026_05_28_PM.md\` (included) for the full timeline + sidecar dive.

The instrumentation bisects two candidate mechanisms on the next repro:

- **FE** (\`BackendStatus.tsx\`, \`SystemStats.tsx\`): a \`console.log\` inside each sysinfo handler.
  - silent + DOM frozen → backend stopped delivering (broker dropped the subscriber)
  - fires + DOM frozen → SolidJS scheduler/render-effect stalled
- **Sidecar** (\`wps.rs\` \`Broker::publish\`): a \`tracing::warn!(target: "wps-diag", …)\` when the client is \`None\` OR when a \`sysinfo\` event publishes to **zero** routes. The "0 routes for sysinfo" line is the smoking gun for the broker-drop path.

## Why merge it (we never reproduced)

The freeze is intermittent and tied to the (separately-fixed) cascade crash family. Landing the instrumentation means the **next** occurrence is captured in the host + sidecar logs instead of needing a live repro. All three log sites are explicitly tagged for removal once the root cause lands.

## Also included (analysis docs)

- \`ANALYSIS_AGENT_PANE_CRASH_2026_05_28_PM.md\` — timeline, the 6-fix history of this crash class, the sidecar eventsub dive (eventsub aggregator is NOT the bug; it's the FE re-sub on cascade unwind).
- \`ANALYSIS_SEND_NOW_FLASH_2026_05_28.md\` — diagnosis behind the merged #1139 fix.
- \`ANALYSIS_LARGE_FILE_MODULARIZATION_CANDIDATES_2026_05_28.md\` — refreshed post-R.6 survey with 4 carve plans.

## Verification

FE \`tsc\` clean (pre-existing errors only); \`cargo check -p agentmux-srv\` clean. Rebased on current main, no conflicts.

### DIFF
--- a/.changesets/1780029130-diag-statusbar-sysinfo-channel-instrumentation-to-bisect-age-afum.md
+++ b/.changesets/1780029130-diag-statusbar-sysinfo-channel-instrumentation-to-bisect-age-afum.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+diag(statusbar): sysinfo channel instrumentation to bisect agent-pane cascade freeze
--- a/agentmux-srv/src/backend/wps.rs
+++ b/agentmux-srv/src/backend/wps.rs
@@ -315,10 +315,26 @@ impl Broker {
 
         let client = match &inner.client {
             Some(c) => c,
-            None => return,
+            None => {
+                // Diagnostic for agent-pane cascade freeze (2026-05-28):
+                // if the bridge gets dropped, every publish silently no-ops.
+                tracing::warn!(target: "wps-diag", event = %event.event,
+                    "broker.publish: client is None — no delivery");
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #1141, opened 2026-05-28
build(package): label local builds instead of bumping the committed version

AgentA-asaf:
## Problem

\`task package\` step 1 committed a \`bump patch\` before any build step could fail, on whatever branch you were on. That broke three ways, repeatedly:

1. **Cross-branch collision.** The "monotonic counter" is only monotonic on a linear branch. Two feature branches off \`main@0.39.2\` both bump to \`0.39.3\` → same version → same Desktop folder → same data dir → the first build's running instance locks the second out. Parallel branches are the *normal* multi-agent state here.
2. **Stranded commit on failure.** Bump is step 1, packaging is last; any failure between strands a \`chore: bump\` commit for a build that never shipped.
3. **Fights changesets (RFC #857).** Feature PRs must not bump; the auto-bump put version-file mutations on feature branches that polluted PRs or had to be reverted every iteration.

Root cause: a local smoke build mutated committed *release* state. Research (semver §10, git-describe, changesets docs) is unanimous — versions are for releases, identifiers are for builds.

## Change

\`task package\` no longer bumps and no longer touches git. New orchestrator \`scripts/package.sh\` computes once per build:

- **VERSION** — semver core, unchanged. Still drives binary names + the verification grep.
- **LABEL** = \`<version>+g<sha>[.dirty].<stamp>\` — semver build metadata (ignored for precedence). Names the folder + ZIP. \`.dirty\` flags uncommitted builds; the timestamp makes every folder unique so the running-instance lock can't fire.
- **CHANNEL** = \`dev-portable-<branch>[-<stamp> if --fresh]\` — the data-dir key. Branch-scoped so rebuilds share a session; \`--fresh\` for a throwaway dir.

New \`agentmux-common/build.rs\` declares \`rerun-if-env-changed=AGENTMUX_BUILD_CHANNEL_DEFAULT\` so a branch-switch / \`--fresh\` actually rebakes the channel instead of serving a stale \`option_env!\` cache. The label is deliberately *not* tracked (it changes every build; tracking would kill caching). \`package-portable.sh\` also skips the VERSION_HISTORY.md size-table append for local labeled builds (another per-build git mutation). No \`data_paths.rs\` change — the channel mechanism already existed.

## Verification (end-to-end, this branch)

Ran \`task package\` on this branch. Result:
- Folder: \`agentmux-0.39.2+ge80fac55.20260528T220340-x64-portable\` ✅ (version unchanged, sha = this commit, no \`.dirty\`)
[...]

PR #1143, opened 2026-05-29
feat(statusbar): show local build label in the instance panel

AgentA-asaf:
## Summary

Follow-up to #1141 (local build versioning). A \`task package\` build now carries an ephemeral label — \`<version>+g<sha>[.dirty].<stamp>\` — in its folder/ZIP name, but the running app showed only the bare version, so there was no way to tie a window back to the artifact on disk.

Adds a **"Label" row** to the status-bar instance panel (the popover under the version chip), copyable verbatim so it pastes straight into a folder-name match / grep.

## How the label reaches the app

Read at **runtime** from the \`agentmux-portable.marker\` file the packaging script already writes next to the launcher — **not** baked via \`option_env!\`. Why:

- The label changes every build, and the marker is rewritten on every \`task package\`, so a runtime read is **always accurate**.
- A compile-time bake would go **stale** whenever \`agentmux-cef\` itself wasn't rebuilt (e.g. a frontend-only \`task package\` produces a fresh marker but an unchanged host binary → mismatched About vs folder). It would also force a recompile every build.

Released / installed / dev builds have no marker → \`buildLabel\` is \`null\` and the row is hidden. Version + git hash + build time still show as before.

## Changes

- **platform.rs**: \`read_build_label()\` reads + parses the marker relative to \`current_exe()\` (host runs from \`<portable>/runtime/\`, marker one level up); surfaced as \`buildLabel\` in \`get_about_modal_details\`.
- **InstancePanel.tsx**: render the Label row when present; copy raw (no \`"label: "\` prefix).

## Test plan

- [x] \`tsc\` clean (no InstancePanel errors).
- [x] Release build via \`task package\` (in progress) — debug \`cargo check\` fails only on the unrelated \`cef-dll-sys\` CMake wrapper (known build race); release path builds clean.
- [ ] Smoke: open the new labeled portable → click the version chip → "Label" row shows the exact folder-name label; copy button yields the raw string.

### DIFF
--- a/.changesets/1780029884-feat-statusbar-show-local-build-label-in-the-instance-panel-3oyu.md
+++ b/.changesets/1780029884-feat-statusbar-show-local-build-label-in-the-instance-panel-3oyu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(statusbar): show local build label in the instance panel
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -126,6 +126,42 @@ pub fn get_env(args: &serde_json::Value) -> serde_json::Value {
     }
 }
 
+/// The ephemeral build label of a local portable, read from the
+/// `agentmux-portable.marker` file the packaging script writes next to
+/// the launcher (`scripts/package-portable.sh`). Format on disk:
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1146, opened 2026-05-29
feat(agent-pane): IME composition + agent-keystroke perf marks + slash TODO

Agent1-asaf:
## Summary

First PR landing the input-responsiveness work scoped in [\`SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md\`](../blob/main/docs/specs/SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md) (spec ships in a separate doc PR). Implements **items 1, 2, and 5** from the spec's action items. Items 3, 4, 6, 7 ship as follow-up PRs per the [execution plan](../blob/main/docs/specs/PLAN_INPUT_RESPONSIVENESS_EXECUTION_2026_05_29.md).

## Item 1 — IME composition handling (real user-facing bug fix)

The agent composer's Enter handler previously submitted unconditionally, which incorrectly fired the prompt when a CJK / Vietnamese / IME user pressed Enter to confirm a composition candidate.

- Early return in \`handleKeyDown\` if \`e.isComposing || e.keyCode === 229\` (Safari quirk).
- \`onCompositionStart\` / \`onCompositionEnd\` JSX props on the textarea flip a \`composingRef\` boolean.
- \`updateAutocomplete()\` skips while composing — slash matching on intermediate IME state would flash the dropdown.
- \`compositionend\` re-runs \`updateAutocomplete()\` so the dropdown reflects committed text once the IME finalizes.

## Item 2 — \`agent-keystroke\` perf marks

Four mark pairs mirroring the terminal pattern from \`SPEC_TERMINAL_LATENCY_BENCHMARK_2026_05_19\`:

| Mark | Span |
|---|---|
| \`agent-keystroke\` | \`onInput\` entry → return (handler cost only) |
| \`agent-input-raf\` | RAF enqueue → callback start (queue depth) |
| \`agent-input-raf-cb\` | callback start → end (coalesced scroll cost) |
| \`agent-submit\` | submit handler entry → \`onSendMessage\` return |

Uses existing \`markStart\` / \`markEnd\` from \`@/perf\` (~50 ns each, production-safe). Enables the agent-keystroke bench in PR C and surfaces agent input latency in the dev-mode Perf HUD.

## Item 5 — slash matcher TODO comment

[...]

PR #1150, opened 2026-05-29
test(bench): agent composer keystroke-latency benchmark via CDP

Agent1-asaf:
## Summary

Implements **item 3** from [\`SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md\`](../blob/main/docs/specs/SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md) per the execution plan §Item 3. Counterpart to \`bench-term-echo.mjs\`, but for the agent pane composer.

**Depends on PR #1146** (the perf marks). Mergeable independently — the bench fails fast with a clear error message if the marks aren't present, so it's safe to land before #1146 and will start producing real data after #1146 lands.

## What it measures

| Mark | What it captures | Spec target |
|---|---|---|
| \`agent-keystroke\` | Synchronous \`onInput\` handler cost | P95 ≤ 5 ms |
| \`agent-input-raf-cb\` | RAF callback / coalesced scroll cost | P95 ≤ 5 ms |
| \`agent-submit\` | Submit handler cost (sanity; should be 0 in a no-submit run) | — |

Default end-to-end keystroke P95 threshold: 50 ms (spec §2 internal "snappy" target). Exits non-zero on threshold violation.

## How it works

1. Connects to CEF's CDP debug port (9223 dev / 9222 release) — same port \`agentmux-cef/src/browser_api/cdp.rs\` uses.
2. Discovers the main page target, opens \`Runtime\` + \`Input\` sessions.
3. Verifies \`.agent-input\` exists. **Operator must open an agent pane first** (no \`pane.open\` App API call — could add as a follow-up).
4. Focuses the textarea.
5. Drives N synthetic keystrokes via \`Input.dispatchKeyEvent\` (rawKeyDown → char → keyUp).
6. Reads measures via \`Runtime.evaluate\` + \`performance.getEntriesByType('measure')\`.
7. Reports P50 / P95 / P99 / max.
8. Clears the textarea on exit. **No submit, no agent invocation, zero token cost.**

## Limitations (documented in script header + README)

- Dispatches DOM-level events; does NOT exercise the OS keyboard pipeline. End-to-end real-user latency may be slightly higher.
- Requires an open agent pane (operator-driven setup, like \`bench-term-echo.mjs\` requires a terminal pane).
- Returns zero measures until PR #1146's perf marks merge. The bench distinguishes this from real failure with a targeted error message.

## CLI

\`\`\`bash
node tools/tests/bench-agent-keystroke.mjs                 # 200 keystrokes, 50 ms apart, threshold 50 ms P95
node tools/tests/bench-agent-keystroke.mjs --count 500 --p95-threshold-ms 30 --output-file results.json
node tools/tests/bench-agent-keystroke.mjs --cdp-port 9222 # release build (default is 9223 dev)
\`\`\`

## Test plan

- [x] \`node --check\` passes (valid JS).
- [ ] After PR #1146 merges: run against \`task dev\` instance, confirm P50/P95 are below targets.
- [ ] Before/after profile of a synthetic regression (e.g. inject a 30 ms blocking loop in handleInput) — bench should catch.

## Why CDP, not App API

The terminal bench uses the App API because the PTY input path is a single RPC call. There's no equivalent for "dispatch a keystroke into the agent textarea" — adding one purely for benching would be over-engineering. CDP is the natural fit and already running in dev (port 9223).

### DIFF
--- a/.changesets/1780033475-test-bench-agent-composer-keystroke-latency-benchmark-via-cd-c3eo.md
+++ b/.changesets/1780033475-test-bench-agent-composer-keystroke-latency-benchmark-via-cd-c3eo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+test(bench): agent composer keystroke-latency benchmark via CDP
--- a/tools/tests/README.md
+++ b/tools/tests/README.md
@@ -49,6 +49,41 @@ Reports p50/p95/p99/max in ms. Save `--output-file` before and after a
 fix to compare distributions. Full spec at
 [`docs/specs/SPEC_TERMINAL_LATENCY_BENCHMARK_2026_05_19.md`](../../docs/specs/SPEC_TERMINAL_LATENCY_BENCHMARK_2026_05_19.md).
 
+## `bench-agent-keystroke.mjs`
+
+Node.js benchmark for **agent composer keystroke latency** — the
+synchronous cost of the `<textarea>` `onInput` handler plus the
+RAF-coalesced scroll callback. Counterpart to `bench-term-echo.mjs`,
+but for the agent pane.
+
+CDP-driven (not App-API-driven) — connects to CEF's debug port
+(9223 dev / 9222 release), dispatches synthetic keystrokes via
+`Input.dispatchKeyEvent`, and reads `agent-keystroke:*` and
+`agent-input-raf-cb:*` perf measures via `performance.getEntriesByType`.
+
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1156, opened 2026-05-29
fix(layout): animate pane open/close/split reflow (DOM + native browser panes)

AgentX-asaf:
## Problem
Opening, closing, splitting, or rebalancing a pane made panes **snap** to their new geometry — a visible jerk.

## What this does
Coordinated reflow animation across **both** pane layers, on a single frontend clock:

**DOM panes (terminal / agent / editor)** — pure CSS:
- `DefaultAnimationTimeS` 0 → **0.15s**, ease-out; the `.tile-node` wrapper transition (already present) now actually runs.
- The inner rect is applied **immediately** on open/close (it was debounced by `animationTimeS`, which held the body at its old size and snapped it). `.block-content` transitions its size in lockstep with the wrapper.

**Native browser panes (HWND — CSS can't move them)** — per-frame tracking:
- New `frontend/app/platform/pane-anim.ts` signal, pinged from the win32 `DisplayNode` when a node's geometry changes during an animation.
- While active, each browser pane re-samples its (CSS-animating) DOM placeholder rect every frame and forwards it via the existing `browser_pane_resize` → `SetWindowPos`. The native window tracks the DOM exactly — **one clock, no drift, zero new Rust.** (Rejected a host-side tokio loop precisely because it would drift on a separate clock.)

## Accessibility — essential-motion exemption
Pane reflow + the drag-rearrange placeholder glide animate **even under `prefers-reduced-motion`** — a deliberate product decision: pane reflow is a spatial cue, not decoration. The global `.prefers-reduced-motion * { transition-property: none !important }` reset in `app.scss` is beaten by higher-specificity `!important` rules on `.tile-node` / `.block-content` / `.placeholder-sizer`. The decorative `.placeholder` fade is intentionally left subject to the reset.

## Verified
`task dev` with OS reduce-motion **on** (the strict case): tile reflow, content, browser panes, and the drag placeholder all glide (~150ms). Confirmed via runtime diagnostic that the tile transition resolves to `0.15s` / `width,height,transform` despite reduced-motion.

## Scope / risk
Frontend-only; **no Rust changes** (reuses `browser_pane_resize`). No version bump — changeset included per the repo workflow. Resize-drag path unchanged (`.animate` off during drag). Browser-pane size animation per-frame may relayout heavy pages; fallback (smooth position / snap size) noted in the spec if it stutters.

## Docs
- `docs/specs/SPEC_PANE_REFLOW_ANIMATION_2026_05_29.md` (design)
- `docs/architecture/PANE_LAYOUT_AND_REFLOW_ARCHITECTURE.md` (full cross-stack reference)
- `docs/analysis/ANALYSIS_PANE_OPEN_CLOSE_ANIMATION_2026_05_29.md` (investigation)

### DIFF
--- a/.changesets/1780034495-fix-layout-animate-pane-open-close-split-so-panes-glide-into-qq3f.md
+++ b/.changesets/1780034495-fix-layout-animate-pane-open-close-split-so-panes-glide-into-qq3f.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): animate pane open/close/split so panes glide into place instead of snapping
--- a/docs/analysis/ANALYSIS_PANE_OPEN_CLOSE_ANIMATION_2026_05_29.md
+++ b/docs/analysis/ANALYSIS_PANE_OPEN_CLOSE_ANIMATION_2026_05_29.md
@@ -1,0 +1,171 @@
+# ANALYSIS: Pane Open/Close Animation — why panes "jerk into place" and what a real fix takes
+
+- **Date:** 2026-05-29
+- **Author:** AgentX
+- **Area:** `frontend/layout/` (SolidJS tiling layout) + `frontend/app/block/` + (for full coverage) `agentmux-cef` host
+- **Goal:** When a pane opens/closes/splits, neighbors should glide into their new geometry instead of snapping.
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

PR #1158, opened 2026-05-29
refactor(agents): partial-update API for updateagentinstance (Phase 3b unblock)

AgentA-asaf:
**PR α** of the partial-update refactor ([scope](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/SPEC_UPDATEAGENTINSTANCE_PARTIAL_UPDATE_2026_05_29.md)). Pure decoupling — no read-flip yet.

## Problem
The `updateagentinstance` handler did **fetch-and-merge**: `instance_get` the full row (transient per-launch fields included) → fill the unspecified of 5 mutable fields → write the whole struct back. That read is the **only** production caller needing `instance_get`'s transient columns, pinning it to legacy `db_agent_instances` and blocking 3b.3c.

## Change
- New `InstanceUpdate` (all-`Option`) + `Store::instance_update_partial(id, &upd)` — dynamic `UPDATE … SET <only Some cols> WHERE id=?`, same post-write reload (registry mirror + dual-write) as `instance_update`. Returns the post-update row, so the handler keeps its `agentinstances:changed:<definition_id>` event + response **from the reload it already runs** (no extra query).
- Handler rewritten to call it directly — fetch-and-merge gone.
- No-fields update returns the unchanged row; `None` reserved for not-found (handler can still distinguish).
- `instance_update(&AgentInstance)` kept as test/internal helper.

SQL behavior identical (same 5 columns). The reload's consumers read only **non-transient** fields, so the follow-up can flip `instance_get` → db_agents (3b.3c) without breaking them.

## Resolved open question
`UpdateAgentInstanceCommand` has **zero frontend callers** → response shape is free; returns the fresh row.

## Tests
5 unit cases (status-only · session-only · `Some("")` clear · all-None no-op · not-found). `cargo test -p agentmux-srv test_instance_update_partial` green; `cargo check -p agentmux-srv` clean.

Follow-ups: **β** flips `instance_get` → db_agents (3b.3c); **γ** deletes the `instance_list` status-filter legacy branch (3b.3b, no live caller).

### DIFF
--- a/.changesets/1780036825-refactor-agents-partial-update-api-for-updateagentinstance-d-0x37.md
+++ b/.changesets/1780036825-refactor-agents-partial-update-api-for-updateagentinstance-d-0x37.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(agents): partial-update API for updateagentinstance (drops fetch-and-merge)
--- a/agentmux-srv/src/backend/storage/agents.rs
+++ b/agentmux-srv/src/backend/storage/agents.rs
@@ -182,6 +182,21 @@ pub struct GitHubContext {
     pub workflow_run_id: Option<u64>,
 }
 
+/// Partial-update payload for [`Store::instance_update_partial`]. Each
+/// `Some` field is written; `None` leaves the column untouched. Mirrors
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1163, opened 2026-05-29
fix(agents): un-fold working_directory in db_agents dual-write (Phase 3c, Decision B)

AgentA-asaf:
**Phase 3c, Decision B** (settled in [Discussion #1095](https://github.com/agentmuxai/agentmux/discussions/1095)). Standalone correctness fix that also unblocks 3b.5.

## Problem
`db_agents.working_directory` should hold the agent **definition's configured** cwd, but `agents_dual_write_instance_create` wrote `inst.working_directory` (the instance's **resolved** launch dir) into it in all three branches — folding per-launch state over def config.

- **Live bug:** `agent_def_list` (powers the configured-cwd shown in agent settings) reads db_agents → returned the wrong (resolved) cwd after any launch.
- **3b.5 blocker:** `agent_def_get`→db_agents couldn't return the pristine def cwd template_promote needs (`agent_session.rs:854`).

## Fix
Source `working_directory` from `def.working_directory` (already loaded in the fn) instead of `inst.working_directory`, in all three branches (user-clone UPDATE, continuation UPDATE, head INSERT). `identity_id`/`memory_id` correctly stay per-instance bindings.

The instance's **resolved** cwd is unaffected — it already lives on the block (`block.meta["cmd:cwd"]`), reachable via the `last_block_id` column the next 3c PR adds. **db_agents = durable agent config; block = per-launch runtime state.**

## Tests
The two dual-write tests asserted the old folded value. Updated to give each def a *distinct* configured cwd and assert db_agents reflects **that** (not the instance's) — regression-locking the un-fold. `cargo test -p agentmux-srv dual_write` → 12 passed; `cargo check` clean.


a5af:
@codex review

PR #1164, opened 2026-05-29
feat(storage): db_agents.last_block_id + dual-write maintenance (Phase 3c PR1)

AgentA-asaf:
## What

Adds `db_agents.last_block_id` — the **only transient field** `db_agents` needs to retain — plus its dual-write maintenance. This is the additive, write-only column that unblocks the **My Agents** read flip (Phase 3b.1b).

## Why

`MyAgentsList` ← `listrecentsessions` ← `instance_list_named` (the 3b.1b flip) needs the instance's `block_id` to locate the filestore `output.state.json` snapshot (preview / node_count / last_active). The other transient instance fields don't survive the migration:

- `status` / `session_id` / `ended_at` — cross the wire today but have **no live consumer**; they drop with the legacy `db_agent_instances` table.
- `started_at` → use `db_agents.updated_at`.

So one additive column unblocks 3b.1b; everything else dissolves in Phase 3c.

## Changes

**Schema** (`migrations.rs`):
- `OBJECT_SCHEMA_VERSION` 4 → 5
- `last_block_id TEXT NOT NULL DEFAULT ''` on `db_agents` (in `CREATE TABLE` + the additive `ALTER TABLE … ADD COLUMN` list — idempotent dup-column swallow)

**Dual-write** (`dual_write.rs`): `agents_dual_write_instance_create` now mirrors `inst.block_id` → `last_block_id` in all three branches (user-clone UPDATE, continuation UPDATE, template-instance head INSERT + ON CONFLICT DO UPDATE).

> Note: unlike `working_directory` (def-configured, un-folded in #1163), `last_block_id` **is** instance-derived — it names the latest launch's block. That's correct: it's per-launch runtime state.

## Safety

**Write-only by design.** Nothing reads `last_block_id` until the 3b.1b flip. Existing SELECTs use explicit column lists, so the additive column is safe. Backfill of pre-existing rows is deferred to the 3b.1b PR that first reads it.

## Tests

Extended the two create-path dual-write tests with non-vacuous `last_block_id` assertions (head INSERT → `"blk-head"`; user-clone fold UPDATE → `"blk-fold"`). `cargo test -p agentmux-srv dual_write` + `migration` green (12 + 26).

Phase 3c PR1. Tracking: #1095.

### DIFF
--- a/.changesets/1780053165-feat-storage-add-db-agents-last-block-id-column-dual-write-m-xntr.md
+++ b/.changesets/1780053165-feat-storage-add-db-agents-last-block-id-column-dual-write-m-xntr.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(storage): add db_agents.last_block_id column + dual-write maintenance (Phase 3c PR1)
--- a/agentmux-srv/src/backend/storage/dual_write.rs
+++ b/agentmux-srv/src/backend/storage/dual_write.rs
@@ -267,7 +267,8 @@ impl Store {
                     github_context = ?6,
                     instance_name = ?7,
                     updated_at = ?8,
-                    user_hidden = ?9
+                    user_hidden = ?9,
+                    last_block_id = ?10
                  WHERE id = ?1",
                 params![
                     def.id,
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1165, opened 2026-05-29
fix(cef): never bind a window label to an off-screen warm-pool HWND (fixes window drag)

AgentA-asaf:
## Symptom

Window drag silently no-ops — the title bar responds, `set_window_position` fires ~60×/sec, `SetWindowPos` returns success, but the visible window never moves.

## Root cause (confirmed live, not theory)

`window_hwnds["main"]` was bound to an **unpromoted warm-pool window**.

The warm pool keeps `POOL_TARGET_SIZE` (=2) windows parked off-screen at `(-32000, -32000)` for instant tab tear-off. They stay `IsWindowVisible` until promoted. When `capture_hwnd_for_label("main")`'s fast path returns `NULL` (CEF Views hides the outer HWND on Win32), the fallback picks *"the first `IsWindowVisible` process window not already mapped"* — and in EnumWindows Z-order that can be a pool window. Every subsequent drag faithfully moved an **off-screen** pool window, so nothing visible happened.

**Live evidence** (running 0.39.3 portable):
- Host log: `[win-resolve] resolved via window_hwnds cache  label=main  cache_hwnd=0xff04bc` on every drag tick.
- `0xff04bc` = top-level `Chrome_WidgetWin_1` titled `'A'` at rect `(-32170, -32054)` — a parked pool window.
- The **real** visible window was `0x1a057c` (`'S'`) at `(135, 235)`.

### Relationship to #1133

This is the **live-but-wrong** sibling of the **dead-HWND** case #1133 fixed. #1133 added `IsWindow`-based eviction — but a pool window is *alive*, so eviction never fires and the wrong bind sticks for the whole session (which is why drag is reliably broken, not flaky). 

**Not** caused by the window.rs carve (#1144): `git show` confirms that was a byte-identical pure move of this code. The bug predates it.

## Fix

Both EnumWindows fallbacks now refuse off-screen pool windows:

- **`is_offscreen_pool_window(hwnd)`** — `rect.left < OFFSCREEN_POOL_THRESHOLD_X` (`-20000`). Well below any real monitor origin (a 4K monitor left of primary sits at ~-3840), yet generous enough to still catch a pool window nudged a few hundred px by a failed drag (that's why the observed coord was -32170, not exactly -32000).
- **`find_main_window`** (the resolve fallback for `"main"`) — skips such windows alongside its existing floating-pane-class skip (pool windows share `Chrome_WidgetWin_1`, so a class check can't catch them — needs a position check).
- **`capture_hwnd_for_label`** fallback — skips them for on-screen labels; the skip is **disabled for `window-pool-*` labels** (those legitimately *are* off-screen).

Net: an on-screen label can no longer bind to a hidden pool window, so the wrong HWND can't enter the cache in the first place. This also closes the residual close-button case #1133 only half-fixed.

## Verification

`cargo check -p agentmux-cef --release` clean. Verified empirically on a portable build against the exact failing instance (see PR thread). Pool-label capture path unaffected (skip gated off for `window-pool-*`).

### DIFF
--- a/.changesets/1780054572-fix-cef-bind-window-label-to-promoted-window-not-off-screen--nxw8.md
+++ b/.changesets/1780054572-fix-cef-bind-window-label-to-promoted-window-not-off-screen--nxw8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): bind window label to promoted window, not off-screen warm-pool HWND (fixes window drag)
--- a/agentmux-cef/src/commands/window/lifecycle.rs
+++ b/agentmux-cef/src/commands/window/lifecycle.rs
@@ -115,6 +115,39 @@ pub fn close_window_by_label(
 #[cfg(target_os = "windows")]
 const FLOATING_PANE_CLASS_NAME: &str = "AgentMuxFloatingPane";
 
+/// X-coordinate threshold (in screen px) below which a top-level window is
+/// treated as an off-screen warm-pool member, not a real user window.
+///
+/// Unpromoted pool windows are created at `POOL_OFFSCREEN_X` = -32000
+/// (`window_pool.rs`) and stay there until promoted — but they remain
+/// `IsWindowVisible`, so the EnumWindows-based fallbacks below would
+/// otherwise enumerate one and bind `"main"` to a window the user can't
+/// see. Drags and closes then act on the hidden pool window instead of the
+/// visible one (root cause of the "window won't drag" regression: every
+/// `set_window_position` faithfully moved an off-screen pool window). A
+/// failed drag can nudge the parked window a few hundred px before the bind
+/// is fixed, so we test against a generous threshold rather than the exact
+/// parking coordinate. No real monitor origin is anywhere near -20000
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentA-asaf:
## ✅ Verified empirically on the fixed build

Built a portable from this branch (`0.39.3+gb204eb18`) and reproduced against the same machine that exhibited the bug.

**Before (broken build `0.39.3+g52f86eed`):** `"main"` → `0xff04bc`, a top-level titled `'A'` at rect `(-32170,-32054)` — an off-screen warm-pool window. Drag moved it; nothing visible happened.

**After (this fix):** `"main"` resolves via the class-aware `find_main_window` fallback (cache + reducer-registry empty) →

```
[win-resolve] class-aware EnumWindows fallback  label=main  fallback_hwnd=0xaf0568
```

`0xaf0568` = title `'S'`, rect `(-3, 349)`, size 630×1098 — the **on-screen** window. Title-bar drag now moves the visible window. The off-screen pool windows are correctly skipped.

PR #1166, opened 2026-05-29
fix(floating-pane): deterministic redock-onto-main (no longer depends on the HWND-capture race)

AgentA-asaf:
## Problem

Tear-off → drag a floating pane back over the **main window** → redock worked only **intermittently** ("doesn't always redock").

`resolve_window_at_cursor` (the redock detector) reverse-looks-up the cursor's window HWND in `window_hwnds` to get its label. But `window_hwnds["main"]` is populated by a startup `capture_hwnd_for_label` that **races** the main window becoming enumerable. When the capture loses that race, `"main"` is **absent** from the map, so the reverse lookup can't recognise the main frame under the cursor → `target_label: null` → the floater silently fails to re-dock. Once unresolved, it stays unresolved for the whole session (same session-long stickiness as the drag bug).

**Diagnosed live this session:** a "bad" instance logged `capture_hwnd_for_label: no available HWND for label=main` (0 cache hits, all fallback resolutions) and redock-onto-main was dead; a "good" instance that won the race worked. Same binary.

## Fix

In `resolve_window_at_cursor`, resolve the main frame **independently of the cache** via `find_main_window()` — which already skips floaters and off-screen warm-pool windows (#1165) — and recognise a cursor-over window as `"main"` by HWND **even when it isn't in `window_hwnds`**.

- Resolved **once per call** (not per Z-order iteration).
- **Null-safe**: an unresolved main (`0`) never matches → no behavior change in that case.
- The **cached path still wins** when `"main"` is present; this only adds a deterministic fallback for the race-lost case.

`find_main_window` is made `pub(super)` so the sibling `motion` module can call it.

## Why not "fix the capture race" instead?

The cache is also consumed by `resolve_window_hwnd` (drag/close), which **already** falls back to `find_main_window()` — so drag was never affected by the race. This change brings redock detection to the same cache-independent footing. (Continuously/authoritatively populating `window_hwnds["main"]` at promote-time is a reasonable future hardening, but this is the minimal deterministic fix for the user-visible symptom.)

## Test

`cargo check -p agentmux-cef --release` clean. `resolve_window_at_cursor` is entirely Win32 (EnumWindows/GetWindowRect) — not unit-testable without mocking, consistent with the other window handlers; verified by build + live smoke (redock onto main).

### DIFF
--- a/.changesets/1780061050-fix-floating-pane-redock-onto-main-resolves-via-find-main-wi-n0h4.md
+++ b/.changesets/1780061050-fix-floating-pane-redock-onto-main-resolves-via-find-main-wi-n0h4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): redock onto main resolves via find_main_window when window_hwnds cache misses
--- a/agentmux-cef/src/commands/window/lifecycle.rs
+++ b/agentmux-cef/src/commands/window/lifecycle.rs
@@ -155,7 +155,7 @@ unsafe fn is_offscreen_pool_window(hwnd: *mut std::ffi::c_void) -> bool {
 /// skip, the floater (owned, drawn ABOVE its owner) would be
 /// enumerated first and we'd target it instead.
 #[cfg(target_os = "windows")]
-unsafe fn find_main_window() -> *mut std::ffi::c_void {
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1168, opened 2026-05-29
fix(browser-pane): deterministic re-create-after-close so redocked panes always load

AgentA-asaf:
## Problem

Browser panes **sometimes don't finish loading after tear-off → redock** (land blank / errored).

Redock **moves** the block (same `block_id`) into the target window while the floater's old browser pane for that `block_id` is still tearing down. The target's re-create hits the reducer's `Closing` guard → `BrowserPaneManager::create` returns `Err("still closing; retry after on_before_close")` — on the assumption *"the frontend retries on next tick."* **That retry was never implemented**: `browser-view.tsx::createPane` catches the error and calls `model.onError` (`LoadFailed`), with no retry. So loading depended on a **race** between the floater's teardown drain and the target's re-create → intermittent.

Full diagnosis: `docs/analysis/ANALYSIS_BROWSER_PANE_REDOCK_LOAD_RACE_2026_05_29.md`.

## Reducer-coverage note

The lifecycle state machine is **correct** — `Closing → reject` is necessary, else the floater's label-keyed `DrainBrowserPaneByLabel` would evict the freshly-created entry (create reuses the block_id). The gap was a **dropped handoff**: nothing re-created the pane once the close completed.

## Fix — deterministic host-side replay (no frontend retry / timer)

- `AppState.pending_browser_pane_creates: Mutex<HashMap<block_id, {url, rect, window_label}>>`.
- `BrowserPaneManager::create`, on `RegisterResult::Closing`: **stash** the pending create and return `Ok` (deferred) instead of `Err`.
- `drain_closed_label` (the `on_before_close → DrainBrowserPaneByLabel` path): once the drain removes the old entry, **replay** the stashed create. The entry is now gone → `TryRegisterBrowserPaneLive` returns `Fresh` → posts `CreateBrowserPaneTask`. Single-shot (removed from the map).

This closes the loop on the **deterministic close-completion signal the host already owns** (the drain it dispatches), rather than punting to a frontend retry — consistent with the project's "no timers / find the deterministic signal" rule.

## Test

`cargo check -p agentmux-cef --release` clean. The reducer `Closing` return is already covered (`reducer/tests.rs`); the defer/replay path is host-side CEF glue (posts `CreateBrowserPaneTask`), not unit-testable without CEF — verified by build + live tear-off↔redock cycling (previously intermittent, should now always load).

### DIFF
--- a/.changesets/1780071000-fix-browser-pane-replay-deferred-create-after-close-drain-so-pnxh.md
+++ b/.changesets/1780071000-fix-browser-pane-replay-deferred-create-after-close-drain-so-pnxh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(browser-pane): replay deferred create after close-drain so redocked panes always load
--- a/agentmux-cef/src/browser_panes.rs
+++ b/agentmux-cef/src/browser_panes.rs
@@ -187,14 +187,32 @@ impl BrowserPaneManager {
                 Ok(())
             }
             RegisterResult::Closing => {
-                // Reject rather than overwrite: the old CEF Browser is
-                // mid-teardown and its on_before_close will call
-                // DrainBrowserPaneByLabel — if we let create overwrite, drain
-                // would evict the NEW entry. Frontend retries on next tick.
-                Err(format!(
-                    "browser pane for block_id={} is still closing; retry after on_before_close",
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

Issue #1138, opened 2026-05-28
bug(macOS): pane drag crashes the host with EXC_BREAKPOINT — uncaught NSException "unrecognized selector sent to instance"


a5af:
## Summary

Dragging a pane in the layout reproducibly crashes the `agentmux-cef` host on macOS with `EXC_BREAKPOINT` / `SIGTRAP`. Backend (`agentmux-srv`) survives the host death and is left orphaned; the next launch sees a stale `SingletonLock` symlink and hands off to a non-existent PID. Six crash reports of this exact shape produced by today's testing session.

This is *unrelated* to the macOS compile / runtime / traffic-light work tracked in #1130 / #1134 / #1131. Filing as its own issue because the symptom is a hard crash that escapes Rust's `panic` machinery (it's an Objective-C exception caught by AppKit, which calls `_objc_terminate`).

## Environment

- macOS 26.2 (Darwin 25.2)
- Apple Silicon (`aarch64-apple-darwin`, model `Mac14,2`)
- AgentMux v0.39.2 dev build (`task dev`)
- Public `cef = 146.7.0+146.0.12` from crates.io
- CEF framework: prebuilt from cef-dll-sys cargo cache, copied via the `bundle:darwin` flow added in #1134
- Branch under test: `agenta/macos-cef-framework-bundling` (rebased on `2abe5968`-era main + the #1134 runtime fix)

## Reproduction

1. Launch `task dev` and wait for the main window with the default starter layout (a few panes).
2. Drag any pane header to start a tab/pane drag operation (i.e., grab the pane title bar with the intent of re-docking).
3. Host process exits immediately with `EXC_BREAKPOINT`; renderers + GPU + utility subprocesses die with it; backend survives orphaned.

100% reproducible on this branch.

## Crash signature

From `~/Library/Logs/DiagnosticReports/agentmux-cef-2026-05-28-120731.ips`:

```
Exception:       EXC_BREAKPOINT (SIGTRAP) — (Breakpoint) brk 1
Faulting thread: CrBrowserMain (com.apple.main-thread)
asi[0]:          "unrecognized selector sent to instance %p"
asi[1]:          "unrecognized selector sent to instance 0x118001a3f00"
```

Top stack frames (Cocoa exception path):

```
+[NSApplication _crashOnException:]              + 256
-[NSApplication reportException:]                + 460
NSApplicationUncaughtExceptionHandler            + 152
__handleUncaughtException                        + 820
_objc_terminate()                                + 144
[...]

Issue #1134, opened 2026-05-28
bug(macOS): task dev panics in cef-rs LibraryLoader — bundle:darwin is a no-op stub, no Chromium Embedded Framework on disk


a5af:
> Companion to #1130 / PR #1131. Once #1131 lands, `agentmux-cef` compiles on macOS but `task dev` still cannot launch because no build step produces the CEF Framework bundle the host needs at runtime. This issue captures the spec for the fix; the spec file is in `docs/specs/SPEC_MACOS_CEF_FRAMEWORK_BUNDLING_2026_05_28.md` (local; happy to upstream as a docs PR if useful).

## Repro

```bash
git clone git@github.com:agentmuxai/agentmux.git
cd agentmux
# (after #1131 merges; today, cherry-pick that PR first)
task dev
```

Output:

```
task: [build:host:darwin] cargo build --release -p agentmux-cef
    Finished `release` profile [optimized] target(s)
task: [bundle:darwin] echo "macOS bundling not yet implemented"
macOS bundling not yet implemented
task: [dev:serve] (host launched)
thread 'main' (36953425) panicked at /Users/.../cef-146.7.0+146.0.12/src/library_loader.rs:20:14:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound,
                                                  message: "No such file or directory" }
task: Failed to run task "dev": task: Failed to run task "dev:serve": exit status 101
```

## Root cause

`cef-146.7.0+146.0.12/src/library_loader.rs:8–23` does:

```rust
const FRAMEWORK_PATH: &str = "Chromium Embedded Framework.framework/Chromium Embedded Framework";

pub fn new(path: &std::path::Path, helper: bool) -> Self {
    let resolver = if helper { "../../.." } else { "../Frameworks" };
    let path = path.parent().unwrap()
        .join(resolver)
        .join(Self::FRAMEWORK_PATH)
        .canonicalize()
        .unwrap();        // ← panic here when the path doesn't exist
    Self { path }
}
```

So with the dev layout `dist/cef/agentmux-cef`, cef-rs looks for `dist/Frameworks/Chromium Embedded Framework.framework/Chromium Embedded Framework`. Nothing in the current build pipeline ever puts a Framework there:

```yaml
# Taskfile.yml:507–511
bundle:darwin:
    internal: true
    platforms: [darwin]
    cmds:
        - echo "macOS bundling not yet implemented"
```

[...]

PR #444, opened 2026-04-18
feat(window): macOS traffic-light caption buttons

a5af:
## Summary

- Adds classic red/yellow/green traffic-light window controls to the **top-left** of the title bar on macOS, matching native macOS HIG behaviour
- Windows and Linux are unchanged — Win11-style Minimize/Maximize/Close buttons remain on the right
- Platform split via Vite's `platformResolve` plugin (`.platform.tsx` → `.darwin/.win32/.linux`)

## Changes

| File | Change |
|------|--------|
| `window-controls.darwin.tsx` | Traffic lights (close/minimize/zoom circles); `WindowControlsLeft` renders them, `WindowControlsRight` returns null |
| `window-controls.win32.tsx` | Win11 SVG caption buttons (moved from `system-status.tsx`); `WindowControlsRight` renders them, `WindowControlsLeft` returns null |
| `window-controls.linux.tsx` | Re-exports from win32 |
| `window-controls.platform.tsx` | Barrel stub for TS type-checking |
| `window-controls.darwin.scss` | 12 px circles, 8 px gap, `#FF5F57`/`#FFBD2E`/`#28CA41`; glyphs fade in on hover |
| `window-controls.win32.scss` | Win11 button styles (moved from `system-status.scss`) |
| `window-header.tsx` | Renders `<WindowControlsLeft />` before `<WindowDrag>` |
| `system-status.tsx` | Renders `<WindowControlsRight />` instead of inlined `WindowActionButtons` |
| `window-header.darwin.scss` | `--default-indent: 0px` — traffic lights are the left anchor |

## Windows / Linux impact

None. On win32 and linux, `WindowControlsLeft` returns `null` (no extra DOM) and `WindowControlsRight` renders the same Win11 buttons as before, just moved from `system-status.tsx` to the platform file. The `system-status.scss` Win11 styles moved identically to `window-controls.win32.scss`.

## Test plan

- [ ] macOS: traffic lights visible top-left; hover reveals ×/–/⤢ glyphs; close/minimize/zoom all work
- [ ] Windows: Win11 Minimize/Maximize/Close on right, behaviour unchanged
- [ ] Linux: same as Windows
- [ ] No TS errors (`npx tsc --noEmit`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #403, opened 2026-04-15
fix(macos): patch NSApplication for macOS 26 Tahoe drag crash

a5af:
## Summary

- CEF 146 calls private `NSApplication` selectors (`isHandlingSendEvent`, `setEffectiveAppearance:`, etc.) that were removed in macOS 26 Tahoe, causing `EXC_BREAKPOINT` crashes whenever the user drags a pane or the window
- Injects `+resolveInstanceMethod:` into `NSApplication`'s metaclass so unknown selectors get typed stubs before the `___forwarding___` machinery is invoked
- `isHandlingSendEvent` and similar BOOL guard getters return `0` (NO) — returning truthy blocked CEF's `sendEvent:` routing and broke window drag
- All other missing selectors get a void stub
- 100% macOS-only (`#[cfg(target_os = "macos")]`); no impact on Windows or Linux

Also included (macOS bring-up work from this session):
- `package:macos` task — builds `.app` bundle + DMG
- `cef:bundle:darwin` task — copies CEF framework, sidecar, frontend into `dist/cef/`
- `--use-mock-keychain` Chromium switch to suppress the macOS Keychain dialog on first launch
- `NativeKeyEvent` type alias fix (`cef::sys::MSG` is Windows-only, broke macOS compile)
- `framework_dir_path` / `resources_dir_path` set to CEF `.framework/Resources` on macOS (fixes `icudtl.dat not found in bundle` crash)

## Test plan

- [ ] `task dev` launches on macOS 26 Tahoe without Keychain prompt
- [ ] Dragging a pane to rearrange within a window works without crash
- [ ] Dragging the window by its title bar works without crash
- [ ] Windows build unaffected (patch is `#[cfg(target_os = "macos")]` only)
- [ ] Check logs for `macOS 26 compat:` lines on startup to confirm stubs are registered

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1169, opened 2026-05-29
fix(macos): bundle CEF Framework + traffic-light buttons + drag region — make task dev launchable on macOS

a5af:
## Summary

Makes \`task dev\` launchable on macOS after PR #1131's compile fix. Three layered fixes (one commit each) that together take a fresh checkout from "compiles but crashes" to "launches, renders, drags, traffic-lights work".

**Depends on:** [#1131](https://github.com/agentmuxai/agentmux/pull/1131) (compile fix). The first commit on this branch sits on top of #1131's compile fix.

**Closes:** [#1134](https://github.com/agentmuxai/agentmux/issues/1134) (CEF Framework runtime resolution)

## What's in this PR

| Commit | What it fixes |
|---|---|
| \`fix(macos): make task dev launchable on a fresh checkout\` | \`bundle:darwin\` was an \`echo\` stub. Implements it with a new \`scripts/resolve-cef-runtime-darwin.sh\` (3-tier resolver: \`$AGENTMUX_CEF_RUNTIME_DIR_DARWIN\` → \`~/cef-build/darwin/<arch>\` → cef-dll-sys cargo cache). Copies the framework to \`dist/Frameworks/\` via \`ditto\` (preserves symlinks) + copies the framework's GL libraries (\`libGLESv2.dylib\`, \`libEGL.dylib\`, \`libvk_swiftshader.dylib\`, \`libcef_sandbox.dylib\` + the swiftshader ICD) next to the host exe in \`dist/cef/\`. Then in \`main.rs\`: sets \`resources_dir_path\` + \`locales_dir_path\` to the framework's \`Resources/\` directory and adds the macOS-only \`framework_dir_path\` so Chromium's NSBundle-based lookups resolve. In \`app.rs\`: adds \`--password-store=basic\` cross-platform and \`--use-mock-keychain\` on macOS so a fresh dev launch doesn't pop an "AgentMux wants to use confidential information stored in Login" Keychain dialog. |
| \`fix(macos): port PR #444 traffic-light buttons; hide Win11 caption buttons on macOS\` | Minimal-touch port of [PR #444](https://github.com/agentmuxai/agentmux/pull/444) (which couldn't merge as-is — its base is 7 weeks old and would revert subsequent action-widgets / system-status / etc. work). Adds \`window-controls.{darwin,win32,linux,platform}.tsx\` + scss with the Vite \`platformResolve\` pattern. macOS renders \`<WindowControlsLeft />\` as red/yellow/green traffic lights (glyphs fade in on hover, matches Sequoia/Sonoma) and \`<WindowControlsRight />\` as null. Windows + Linux render \`<WindowControlsLeft />\` as null and \`<WindowControlsRight />\` as Win11 caption buttons (moved verbatim out of \`system-status.tsx\` into \`window-controls.win32.tsx\`). \`window-header.tsx\` gets \`<WindowControlsLeft />\` inserted before the WindowDrag spacer; \`system-status.tsx\` swaps the inline \`<WindowActionButtons />\` for \`<WindowControlsRight />\`. No Rust changes (PR #444's Rust diff was against 7-week-old code and would revert recent work). |
| \`fix(macos): keep 10 px WindowDrag spacer between traffic lights and tabs\` | PR #444 collapsed \`--default-indent\` to 0 px on darwin. Kept it at 10 px — the visible 10 px area is the drag handle right of the traffic lights, and the traffic lights themselves are \`-webkit-app-region: no-drag\` so clicks land on close/minimize/zoom buttons (not on the window). |
| \`fix(macos): make the borderless title bar actually draggable\` | The repo had **no working macOS window-drag mechanism**: \`useWindowDrag.darwin.ts\` only set \`data-drag-region=true\` as an HTML attribute with no native reader (the comment about WebView/OS-level handling was aspirational), no CSS declared \`-webkit-app-region: drag\`, and the patched-libcef \`begin_window_drag\` path requires a libcef fork that has no macOS variant. Adds \`-webkit-app-region: drag\` to \`.window-header\` on darwin only — Chromium for macOS respects that, reports it via \`on_draggable_regions_changed\`, and the existing Rust handler calls \`cef::Window::set_draggable_regions\` which makes the underlying NSWindow movable from those rects. Tabs / hamburger / widgets / traffic lights are already \`-webkit-app-region: no-drag\`. Linux deliberately does NOT use this (Wayland suppresses events on drag regions — see existing comment in \`window-header.linux.scss\`). |

## What's NOT in this PR

- The **\`EXC_BREAKPOINT / unrecognized selector\` crash** that triggers on pane drag, tab tear-off, and (with this PR's drag CSS) some window drag. **Pre-existing macOS 26 Tahoe bug** — CEF 146 calls private \`NSApplication\` selectors that Apple removed. Already filed as [#1138](https://github.com/agentmuxai/agentmux/issues/1138). Solved by a separate, scoped PR (porting the NSApplication \`+resolveInstanceMethod:\` injector from [PR #403](https://github.com/agentmuxai/agentmux/pull/403)) coming right after this one. See [\`docs/analysis/REPORT_MACOS_TEAROFF_DRAG_CRASH_2026_05_29.md\`](https://github.com/agentmuxai/agentmux/blob/agenta/macos-nsapp-tahoe-compat/docs/analysis/REPORT_MACOS_TEAROFF_DRAG_CRASH_2026_05_29.md) for the full diagnosis (lands with the NSApp patch PR).
- The **stale-SingletonLock UX issue** (host crash → next \`task dev\` hands off to dead PID → "AgentMux failed to start"). Also in its own follow-up PR. See [\`docs/specs/SPEC_MACOS_TEAROFF_STABILITY_2026_05_29.md\`](https://github.com/agentmuxai/agentmux/blob/agenta/macos-nsapp-tahoe-compat/docs/specs/SPEC_MACOS_TEAROFF_STABILITY_2026_05_29.md) §item 2.
- **Real macOS tear-off** — C1 in \`SPEC_FLOATING_PANE_TEAROFF_CROSS_PLATFORM_2026-05-26.md\`. ~400 lines of \`NSPanel\` + \`addChildWindow:ordered:\` + native \`NSDraggingDestination\` view. Long-running work, not in scope here.

## Test plan

Verified end-to-end during development on macOS 26.2 / Apple Silicon / cef 146.7.0+146.0.12:

- [x] Fresh \`task dev\` builds + bundles + launches without errors
- [x] No Keychain prompt on first launch
- [x] No \`icudtl.dat not found in bundle\` errors
- [x] No \`Failed to load libGLESv2.dylib\` GPU process errors
- [x] Traffic lights render top-left (red/yellow/green, glyphs fade in on hover)
- [x] Win11 caption buttons no longer render top-right on macOS
- [x] Window drag from the title bar works
- [x] \`muxlog\` works after sourcing the shell integration helper from \`agentmux-srv/src/backend/shellintegration/bash.sh\`
- [ ] Windows: no behavioral change (all platform-gated; no Win32 paths touched)
- [ ] Linux: no behavioral change (the drag CSS is darwin-only; bundle:darwin is darwin-only)

### DIFF
--- a/.changesets/1779991094-fix-macos-make-task-dev-launchable-on-a-fresh-checkout-bundl-k6pu.md
+++ b/.changesets/1779991094-fix-macos-make-task-dev-launchable-on-a-fresh-checkout-bundl-k6pu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): make task dev launchable on a fresh checkout — bundle CEF Framework + suppress keychain prompts
--- a/.changesets/1779994356-fix-macos-port-pr-444-traffic-light-buttons-hide-win11-capti-dj40.md
+++ b/.changesets/1779994356-fix-macos-port-pr-444-traffic-light-buttons-hide-win11-capti-dj40.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): port PR #444 traffic-light buttons — hide Win11 caption buttons on macOS, add platform-resolved window-controls
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -499,7 +499,37 @@ tasks:
         internal: true
         platforms: [darwin]
         cmds:
-            - echo "macOS bundling not yet implemented"
+            - |
+                # Resolve macOS CEF runtime via shell helper. Three-tier order:
+                #   1. $AGENTMUX_CEF_RUNTIME_DIR_DARWIN (explicit override)
+                #   2. ~/cef-build/darwin/<arch> (custom builds)
+                #   3. cef-dll-sys cargo cache (fallback — public upstream)
+                # See docs/specs/SPEC_MACOS_CEF_FRAMEWORK_BUNDLING_2026_05_28.md.
+                CEF_DIR="$(bash scripts/resolve-cef-runtime-darwin.sh)"
+                mkdir -p dist/Frameworks dist/cef
+                rm -rf "dist/Frameworks/Chromium Embedded Framework.framework"
+                # `ditto` preserves the Versions/Current symlink chain that the
+                # CEF loader follows on macOS; `cp -R` flattens it and breaks
+                # framework lookup. The cef-dll-sys cargo-cache framework is
+                # already flat (no Versions/), so `ditto` works for both that
+                # and a real cef-build framework.
+                ditto "$CEF_DIR/Chromium Embedded Framework.framework" "dist/Frameworks/Chromium Embedded Framework.framework"
+                # Chromium's GPU process discovers GL libraries via
+                # base::PathService::Get(DIR_MODULE), which resolves to the
+                # directory of the *host* executable — not the framework's
+                # Libraries/ directory. Without these next to the exe the
+                # GPU process exits at gl_implementation.cc:522 ("Failed to
+                # load libGLESv2.dylib") and the renderer surfaces an
+                # app-wide "renderer crashed" banner. Copy rather than
+                # symlink because dev:serve's `cp -r dist/cef/* dist/cef-dev/`
+                # would otherwise leave dangling relative links once the
+                # cef-dev copy lives one level lower than dist/cef.
+                # 22 MB on disk — negligible.
+                for lib in "$CEF_DIR/Chromium Embedded Framework.framework/Libraries/"*.dylib; do
+                    cp -f "$lib" "dist/cef/$(basename "$lib")"
+                done
+                cp -f "$CEF_DIR/Chromium Embedded Framework.framework/Libraries/vk_swiftshader_icd.json" "dist/cef/vk_swiftshader_icd.json"
+                echo "✓ Bundled CEF Framework for macOS (dist/Frameworks/ + GL libs in dist/cef/)"
 
     bundle:linux:
         internal: true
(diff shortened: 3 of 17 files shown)


a5af:
@codex review

a5af:
Quick response to reagent's findings on the most recent two passes:

**P2 \`.window-unfocused\` unused (Kimi)**: legitimate finding, fixed in \`9f42a47c\` — replaced the dead selector with a comment tracking the omission. The macOS HIG focus/blur polish needs a JS observer the original PR #444 also lacked; deferring it to a follow-up keeps this PR scoped to crash + layout.

**P1 \`scripts/resolve-cef-runtime-darwin.sh:45\` arch should be arm64 not aarch64 (Kimi, earlier pass)**: false positive. Verified the actual cef-dll-sys output directory:

\`\`\`
$ ls target/release/build/cef-dll-sys-*/out/ | grep cef_macos
cef_macos_aarch64
\`\`\`

cef-dll-sys uses Rust's \`target_arch\` convention (\`aarch64\`) for its build-script output dir, not the Apple ecosystem's \`arm64\`. \`sidecar.rs:386\` maps \`aarch64 → arm64\` only for the sidecar **binary file name** (\`agentmux-srv-X.Y.Z-darwin.arm64\`), which is a completely separate convention. The resolver script has to match the cef-dll-sys directory layout to find the framework, so \`ARCH="aarch64"\` is required — using \`arm64\` would silently miss every candidate and fall through to the "no framework found" error. Tested end-to-end repeatedly during dev.

I've left both \`aarch64\` in the script and the spec, and added a code comment is not yet there but happy to add one if reviewers want the convention difference called out explicitly.

PR #1170, opened 2026-05-29
fix(macos): patch NSApplication for macOS 26 Tahoe drag crash

### LINKED ISSUE #1138: bug(macOS): pane drag crashes the host with EXC_BREAKPOINT — uncaught NSException "unrecognized selector sent to instance"

a5af:
## Summary

Closes [#1138](https://github.com/agentmuxai/agentmux/issues/1138).

Inject `+[NSApplication resolveInstanceMethod:]` at host startup, before `cef::initialize`, so that the private NSApplication selectors CEF 146 calls during NSDraggingSession setup (which Apple removed in macOS 26 Tahoe) get typed stubs *before* the ObjC `___forwarding___` machinery runs. The stubs prevent `objc_exception_throw` and the subsequent `_objc_terminate()` that kills the host with `EXC_BREAKPOINT` before Rust panic machinery sees anything.

**Depends on:** [#1131](https://github.com/agentmuxai/agentmux/pull/1131) (the macOS compile fix). Without #1131, `agentmux-cef` doesn't build on macOS so this PR can't run.

**Ported from:** [PR #403](https://github.com/agentmuxai/agentmux/pull/403) (a5af, 2026-04-15) which mixed this patch with other macOS bring-up work that has since been done differently in #1131 and #1169. This PR ports just the load-bearing patch with attribution.

## What's in the PR

1. `agentmux-cef/src/main.rs` — adds `patch_nsapp_unrecognized_selector()` and calls it once, after logging init, before `cef::initialize`. `#[cfg(target_os = "macos")]`, ~100 lines of `unsafe extern \"C\"` against libobjc, no new crate dependencies.
2. `docs/specs/SPEC_MACOS_TEAROFF_STABILITY_2026_05_29.md` — implementation spec covering this PR + the follow-up SingletonLock UX fix.
3. `docs/analysis/REPORT_MACOS_TEAROFF_DRAG_CRASH_2026_05_29.md` — code-walk report tying the crash signature in 6 user-machine diagnostic reports to the macOS Phase 7 / C1 work gap.

## Why this approach

`+resolveInstanceMethod:` is the **earliest** point in the ObjC dispatch chain — called before `___forwarding___`. We add a typed stub on the metaclass and return `YES`; the runtime retries the original message send against the real method, message lands, no exception.

The alternative — swizzling `-[NSObject doesNotRecognizeSelector:]` — does **not** work: that method is called *from inside* `___forwarding___`, and returning normally without throwing corrupts forwarding state, causing a secondary crash in `___forwarding___` itself.

Return type matters: `isHandlingSendEvent` and `isSendingEvent` return `BOOL`. A `void` stub leaves `x0 = self` (truthy) on ARM64, and CEF reads that as "the app is already inside a `sendEvent:` call, skip event routing" — breaking window drag silently *even after the crash stops*. So we maintain a small allowlist of BOOL-returning selectors that get a `bool_no_stub` returning 0; everything else gets the void stub (safe for the unbounded set of removed Apple-private APIs).

## What this does NOT do

- **Real macOS tear-off.** That's the C1 work block in `SPEC_FLOATING_PANE_TEAROFF_CROSS_PLATFORM_2026-05-26.md` §3.2 — `NSPanel` + `addChildWindow:ordered:` + native `NSDraggingDestination` view. ~400 lines, separate effort. With this PR's patch in place, the existing no-op tear-off Rust stub still doesn't actually tear off — it just no longer **crashes** when CEF's macOS code tries to drive the drag session through Apple-private APIs.
- **Auto-prune stale `SingletonLock`.** Crashes leave a stale lock pointing at a dead PID; next `task dev` hands off via process_singleton and exits code 24. Fix in a follow-up PR per the spec §item 2.

## Test plan

After this lands on top of #1131:

- [ ] `task dev` launches on macOS 26 Tahoe; no Keychain prompt
- [ ] Drag a pane header → no `EXC_BREAKPOINT`; `muxlog host '\"macOS 26 compat: adding\"'` shows each stubbed selector once
- [ ] Drag the window title bar → no crash; `isHandlingSendEvent` + `isSendingEvent` get BOOL(NO) stubs
- [x] Builds on macOS (verified locally, \`cargo check -p agentmux-cef --release\` finishes 3.24s after rebase on #1131)
- [ ] Windows: zero behavioral change (patch is `#[cfg(target_os = \"macos\")]`)
- [ ] Linux: zero behavioral change (same `cfg` gate)

### DIFF
--- a/.changesets/1780075756-fix-macos-patch-nsapplication-for-macos-26-tahoe-drag-crash-pqlt.md
+++ b/.changesets/1780075756-fix-macos-patch-nsapplication-for-macos-26-tahoe-drag-crash-pqlt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): patch NSApplication for macOS 26 Tahoe drag crash
--- a/agentmux-cef/src/main.rs
+++ b/agentmux-cef/src/main.rs
@@ -254,6 +254,18 @@ fn main() {
         "Initializing CEF browser process"
     );
 
+    // macOS 26 Tahoe compat: CEF 146 calls private NSApplication selectors
+    // (e.g. `isHandlingSendEvent`, `isSendingEvent`) during NSDraggingSession
+    // setup. Apple removed them in macOS 26, so the calls go through
+    // `___forwarding___`, find nothing, and `objc_exception_throw` fires —
+    // AppKit's default uncaught-exception handler then calls `_objc_terminate`
+    // and the host dies (EXC_BREAKPOINT / SIGTRAP on CrBrowserMain). Inject
+    // `+[NSApplication resolveInstanceMethod:]` to add typed stubs *before*
+    // the forwarding machinery is entered. Must run before `cef::initialize`.
+    // See docs/specs/SPEC_MACOS_TEAROFF_STABILITY_2026_05_29.md and PR #403.
+    #[cfg(target_os = "macos")]
+    unsafe { patch_nsapp_unrecognized_selector() };
+
     // Phase B.6 (post-fix): the named-pipe bind in the launcher is
     // the AUTHORITATIVE single-instance lock — a second launcher
     // hits ERROR_ACCESS_DENIED and never reaches the host. We still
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1175, opened 2026-05-29
fix(dnd): suppress 'operation not allowed' cursor during cross-window tear-off (macOS+Linux)

a5af:
## Symptom

On macOS (WebKit) and Linux (WebKitGTK), tearing out a tab or pane shows the prohibited-circle cursor animation throughout the drag, but the tear-off operation completes successfully. Mismatch between cursor affordance and actual behavior — user wonders if they did something wrong.

## Root cause

HTML5 drag-and-drop shows the prohibited cursor whenever the cursor sits over a region whose \`dragover\` handler doesn't call \`preventDefault()\`. AgentMux's cross-window tear-off drives the *actual* tear-off from the \`dragend\` event (\`CrossWindowDragMonitor.{darwin,linux}.tsx:handleCrossWindowDragEnd → performTearOff\`), but installs **no** \`dragover\` listener — so every pixel outside the source tab bar shows the "no" cursor right up until drop. Then \`dragend\` fires, JS opens the tear-off window, and the user is left with a misleading visual.

## Fix

In both \`CrossWindowDragMonitor.darwin.tsx\` and \`CrossWindowDragMonitor.linux.tsx\`, install a document-level \`dragover\` listener that calls \`preventDefault()\` and sets \`dataTransfer.dropEffect = "move"\` whenever \`_currentDragPayload\` is non-null. The browser then treats the whole page as a valid drop target for the duration of the cross-window drag, and the cursor stays on the "move" affordance — matching the actual outcome. Gated on the payload so we don't hijack unrelated DnD sessions (file drops, future drop targets, etc.).

## What this does NOT do

- **Win32**: unaffected — \`CrossWindowDragMonitor.win32.tsx\` doesn't have this issue because Win32 tab tear-off uses a different path (pointer capture + native modal-drag loop per [PLAN_TAB_TEAROFF_PHASE1_WIN32_2026-05-07.md](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/PLAN_TAB_TEAROFF_PHASE1_WIN32_2026-05-07.md)) and doesn't surface this HTML5 artifact.
- **Tear-off behavior itself**: still entirely driven by the \`dragend\` handler. This PR only changes the cursor affordance during the drag.

## Test plan

- [ ] macOS: drag a tab out of the tab bar → cursor shows the "move" affordance the entire time, not the prohibited circle
- [ ] Linux: same as macOS
- [ ] macOS + Linux: tear-off behavior unchanged (new workspace opens, source tab is removed)
- [ ] Windows: zero behavioral change (win32 monitor untouched)
- [ ] File drops / other DnD sessions: unaffected — listener no-ops when no \`_currentDragPayload\`

### DIFF
--- a/.changesets/1780077925-fix-macos-traffic-light-scss-stylelint-comment-stale-docstri-rcve.md
+++ b/.changesets/1780077925-fix-macos-traffic-light-scss-stylelint-comment-stale-docstri-rcve.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): traffic-light scss stylelint comment + stale docstring on win32 move
--- a/.changesets/1780080618-fix-dnd-suppress-prohibited-cursor-during-cross-window-tear--zbul.md
+++ b/.changesets/1780080618-fix-dnd-suppress-prohibited-cursor-during-cross-window-tear--zbul.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dnd): suppress prohibited-cursor during cross-window tear-off on macOS+Linux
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1176, opened 2026-05-29
test(perf): input-latency bench harness — multi-run baseline-delta + reporting (input-first 0.1)

AgentX-asaf:
## What

**Phase 0.1 of the input-first plan** ([discussion #1161](https://github.com/agentmuxai/agentmux/discussions/1161)) — make the input-latency benches statistically trustworthy so they can eventually gate merges without being flaky.

The existing `bench-agent-keystroke.mjs` / `bench-term-echo.mjs` (PR #1150) are single-run with a hard absolute threshold (`fail if P95 > 50 ms`). On variable hardware that's noisy → gets disabled → theater. This PR adds the layer the input-first reviewers converged on, **without modifying the benches**: median-of-N + **delta-vs-baseline** + **reporting-mode-first**.

## Added

- **`tools/tests/lib/bench-stats.mjs`** — pure stats/aggregation/baseline-delta core: percentiles, multi-run aggregation, run-to-run coefficient-of-variation, and a verdict (`pass`/`regress`/`improve`/`no-baseline`/`noisy`) with a report-vs-gate exit policy.
- **`tools/tests/lib/bench-stats.test.mjs`** — 11 unit tests, `node --test`, **no app required**. ✅ all green locally.
- **`tools/tests/bench-aggregate.mjs`** — runs a bench N times, extracts each run's headline metric (`agent`→`keystroke.stats.p95`, `term`→`busy.p95`), aggregates, compares to a committed baseline, prints a REPORT/GATE verdict. `--update-baseline` captures one. Wraps the existing benches (doesn't touch them).
- **`tools/tests/baselines/`** — schema + README: pinned-device requirement, capture instructions, reporting→gating promotion path.
- **`.github/workflows/input-bench-report.yml`** — `workflow_dispatch`, self-hosted `input-bench` runner, reporting-mode. Also runs the stats unit tests (which work on any runner).
- **`docs/analysis/ANALYSIS_PHASE0_VERIFICATION_2026_05_29.md`** — harness design + the 0.3 verification findings.

## Design notes (per reviewer consensus)

- **Delta-vs-baseline, not absolute** — regress only if median-of-per-run-P95 moves > `--tolerance-pct` (default 20%) vs a baseline captured on the *same* device.
- **Noisy-runner guard** — if run-to-run CoV exceeds `--max-cov` (0.25), the verdict is `noisy` (don't trust the gate) rather than a false regress/pass.
- **Reporting-mode first** — `--mode report` never blocks; promote to `--mode gate` only after variance is characterized on the pinned device.

## 0.3 verification findings (handoff)

The doc also records that the remaining Phase-0.3 "cheap wins" need **runtime/visual confirmation** (past the autonomous boundary):
- **Browser keydown** — `browser-model.ts` has no `keyDownHandler`, but a focused native `CefBrowserView` likely never routes DOM keydown to the host, so adding one may not fire / could break native input. **Verify on the app first.**
- **`block.scss` blur** — line 455 `blur(0.1px)` is a *compositing-order hack* (keeps focus rings above xterm), **not** visual blur; only 313/336/405 are gating candidates, and only with a profile + visual check.
- **Region skip-if-unchanged cache** — an HWND-keyed cache has a **handle-reuse hazard** (stale skip → airspace regression); needs close-path eviction + Windows runtime test. Separate Rust PR.

## Scope / risk

- **Zero runtime change** — test tooling + docs only. Stats core is unit-tested.
- Changeset: `patch`.

## Manual next step

Capture baselines on pinned low-end Windows + macOS runners (`--update-baseline`), characterize variance, then flip the bench to `--mode gate`.

### DIFF
--- a/.changesets/1780081477-test-perf-input-latency-bench-harness-multi-run-baseline-del-rkaj.md
+++ b/.changesets/1780081477-test-perf-input-latency-bench-harness-multi-run-baseline-del-rkaj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+test(perf): input-latency bench harness — multi-run baseline-delta + reporting (input-first 0.1, #1161)
--- a/.github/workflows/input-bench-report.yml
+++ b/.github/workflows/input-bench-report.yml
@@ -8,9 +8,14 @@ name: Input-latency bench (report)
 # tools/tests/baselines/README.md.
 #
 # REQUIRES a self-hosted runner on the pinned reference device with a running
-# AgentMux instance and the relevant pane open (CDP reachable). GitHub-hosted
-# runners are too noisy for this timing and cannot drive a live AgentMux, so
-# this workflow is MANUAL (workflow_dispatch) until those runners exist.
+# AgentMux instance and the relevant pane open. GitHub-hosted runners are too
+# noisy for this timing and cannot drive a live AgentMux, so this workflow is
+# MANUAL (workflow_dispatch) until those runners exist.
+#
+# Pass bench-specific args via `extra_args` (they go after `--`):
+#   - agent: drives the agent pane over CDP, e.g. "--cdp-port 9223 --count 200"
+#   - term:  discovers the WS endpoint/auth via authkey.dev (NO --cdp-port),
+#            e.g. "--count 50"
 
 on:
   workflow_dispatch:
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #1178, opened 2026-05-29
perf(airspace): skip redundant SetWindowRgn when a pane's clip is unchanged (input-first 0.3b)

AgentX-asaf:
## What

**Phase 0.3b of the input-first plan** ([discussion #1161](https://github.com/agentmuxai/agentmux/discussions/1161)) — stop re-applying an unchanged airspace clip.

`set_pane_overlay_clip` (Win32) rebuilt and re-applied a region (`SetWindowRgn` + `InvalidateRect`) to **every** live pane on **every** call — so when one overlay moves, every *other* pane still pays a redundant full-region re-apply + repaint each frame. This caches the last-applied clip per pane and skips the no-op.

## How — and why it's safe (the HWND-reuse hazard)

The obvious design (cache keyed by HWND) is unsafe: Windows recycles HWND integer values, so a stale entry could collide with a recreated pane and wrongly skip its clip → an airspace regression. This PR avoids that **by construction**:

- **Keyed by pane label** (never reused for a different pane), stored in `AppState.pane_clip_cache` (Windows-only; `BrowserPaneManager` is a unit struct).
- **HWND value folded into the signature.** A recreated pane (new HWND, even under a reused label) produces a different signature → always re-applies. We only ever skip when the **same label has the same region on the same HWND** — i.e. the identity case where skipping is correct.
- Therefore **no fragile close-path eviction is needed**; entries are pruned to live labels each call to bound the map.
- Hashing the *inputs* (pane geometry + overlay rects) is faithful: identical inputs ⇒ identical region (no false hit). A changed non-intersecting rect only yields a harmless false **miss** (a redundant re-apply), never a false hit.
- **No host-side per-frame coalescer** (that would race DWM and risk re-introducing airspace tearing) — coalescing stays in the renderer rAF.

## Verification

- `cargo check -p agentmux-cef`: **clean** (only pre-existing warnings).
- Pure optimization — no behavior change when the clip actually changes.

### ⚠️ Requires Windows runtime verification (can't be unit-tested)

Please confirm on a Windows build:
1. Open menus / tooltips / autocomplete over a **browser pane** during rapid hover → no overlay-over-unclipped-pane, no pane stuck hidden.
2. **Close / redock** a browser pane, then open an overlay over its area → clip re-applies correctly (exercises the recreate path the HWND-in-signature defends).
3. Resize/split with an overlay open → clip tracks the new geometry.

## Scope
- Windows-only path; macOS/Linux use a different (Views) airspace mechanism, untouched.
- Changeset: `patch`. Subsumes the region-coalescing portion of #1097.

### DIFF
--- a/.changesets/1780086471-perf-airspace-skip-redundant-setwindowrgn-when-a-pane-s-clip-81vc.md
+++ b/.changesets/1780086471-perf-airspace-skip-redundant-setwindowrgn-when-a-pane-s-clip-81vc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+perf(airspace): skip redundant SetWindowRgn when a pane's clip is unchanged (input-first 0.3b, #1161)
--- a/agentmux-cef/src/browser_panes.rs
+++ b/agentmux-cef/src/browser_panes.rs
@@ -129,6 +129,46 @@ impl<'a> BrowserPaneCloseOps for AppStateCloseOps<'a> {
     }
 }
 
+/// Signature of the clip a pane should have, used by `set_pane_overlay_clip`
+/// to skip a redundant `SetWindowRgn` when nothing changed. The HWND value is
+/// folded in so a recreated pane (new HWND under a reused label) never matches a
+/// stale entry — see `AppState::pane_clip_cache`. Within-process only (the hash
+/// need not be stable across runs).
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1179, opened 2026-05-30
fix(macos): set Dock icon at runtime so task dev shows the AgentMux logo

a5af:
## Summary

`task dev` on macOS shows the generic executable tile in the Dock instead of the AgentMux logo. Root cause: `dev:serve` launches the bare `agentmux-cef` Mach-O directly (not inside an `.app` bundle), so macOS has no `CFBundleIconFile` to read.

Fix: set the Dock icon at runtime via `-[NSApplication setApplicationIconImage:]` right after `cef::initialize` (NSApplication exists by then). Works whether or not we're in a bundle, and overrides a bundle icon in any future packaged build too — one code path for all launch modes.

## Details

- The PNG is embedded at compile time (`include_bytes!`) — no dependency on the `dist/` layout or a runtime resource-path lookup.
- It's the **same normal AgentMux logo the Linux taskbar uses** (`assets/linux/icons/hicolor/512x512/apps/agentmux.png`, wired up in `scripts/install-linux-desktop.sh`) — the panel-layout logo, not the brain-alternate variant — so the Dock/taskbar icon is identical across platforms.
- macOS-only (`#[cfg(target_os = "macos")]`); no new crate dependency (raw libobjc FFI, mirroring the existing `patch_nsapp_unrecognized_selector`).
- The created `NSImage` is intentionally leaked (one per process lifetime): `setApplicationIconImage:` retains it and the icon lives as long as the app does.

## Test plan

- [x] `dock-icon: set NSApplication icon to embedded AgentMux logo` logs on launch (verified locally on macOS 26 / Apple Silicon).
- [x] Dock shows the AgentMux panel logo instead of the generic tile.
- [ ] Windows / Linux: zero behavioral change (`#[cfg(target_os = "macos")]`).

### DIFF
--- a/.changesets/1780127615-fix-macos-set-dock-icon-at-runtime-so-task-dev-shows-the-age-ws5k.md
+++ b/.changesets/1780127615-fix-macos-set-dock-icon-at-runtime-so-task-dev-shows-the-age-ws5k.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): set Dock icon at runtime so task dev shows the AgentMux logo
--- a/agentmux-cef/src/main.rs
+++ b/agentmux-cef/src/main.rs
@@ -622,6 +622,12 @@ fn main() {
 
     tracing::info!("CEF initialized, entering message loop");
 
+    // Set the Dock icon now that NSApplication exists. macOS-only; no-op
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #734, opened 2026-05-07
fix(window): switch title-bar drag to absolute positioning (extracted from #729)

AgentA-asaf:
## Summary

Extract from PR #729 (`fix/window-drag-cursor-drift`), keeping ONLY the drag-drift fix. PR #729 had scope bleed — its branch picked up PR #718's table-formatter content. The drag-drift fix itself is well-motivated; this PR carries it on a fresh branch from main.

## Root cause

`move_window_by` calls `GetWindowRect` then `SetWindowPos` from the IPC thread. Under rapid `mousemove`, multiple in-flight IPC calls all read the same stale rect (SetWindowPos is relayed to the UI thread async via the Windows message queue), so the window accumulates only one delta-worth of movement instead of N — the cursor visibly shifts from its original click point on the title bar.

## Fix

Switch to **absolute positioning**:
- `mousedown` snapshots the initial window origin via `get_window_position`
- Each `mousemove` sends `set_window_position(initWin + totalCursorDelta)`
- Each call is self-contained, so concurrent calls are idempotent — no stale-read race possible

New IPC command `set_window_position(x, y)`:
- Rust handler in `agentmux-cef/src/commands/window.rs`
- Registered in `agentmux-cef/src/ipc.rs`
- Non-Windows CEF Views path via `SetWindowPositionTask` in `agentmux-cef/src/ui_tasks.rs`

## Files

| File | Change |
|---|---|
| `frontend/app/hook/useWindowDrag.win32.ts` | Switch from delta + rolling baseline to absolute target coords |
| `agentmux-cef/src/commands/window.rs` | Add `set_window_position(x, y)` |
| `agentmux-cef/src/ui_tasks.rs` | Add `SetWindowPositionTask` + `post_set_window_position` |
| `agentmux-cef/src/ipc.rs` | Register `set_window_position` command |
| `docs/retro/BUG_WINDOW_DRAG_CURSOR_DRIFT_2026-05-07.md` | Bug report + analysis (moved into `docs/retro/` per repo convention) |

## Test plan

- [ ] Drag title bar slowly — cursor stays fixed at click point throughout
- [ ] Drag title bar fast — cursor stays fixed at click point (the stale-race scenario)
- [ ] Double-click title bar — maximizes/restores correctly
- [ ] Minimize / close buttons still work
- [ ] Second window title bar also drags correctly (find_own_top_level_window is per-process)

## Provenance

[...]

PR #1181, opened 2026-05-30
feat(window-drag): smooth Windows title-bar drag via host-side native move loop

AgentX-asaf:
## What
Windows title-bar drag was JS-driven (per-mousemove **awaited** `set_window_position` IPC), so the window tracked the IPC round-trip rate, not the cursor — jittery, worse at 125% scale.

Replaced with a **host-side native move loop on the CEF UI thread**: frontend fires ONE fire-and-forget `start_window_drag` on the 4px threshold; the host runs `SetCapture` + `GetMessage` loop + `SetWindowPos` per `WM_MOUSEMOVE` in physical px. **Zero per-move IPC**, no `devicePixelRatio` math, no PR #734 race-guarding.

## Why not the OS move loop
The raw `ReleaseCapture` + `WM_NCLBUTTONDOWN(HTCAPTION)` trick is incompatible with CEF — Chromium's frame swallows the NC message (instrumented: move loop returned in 0.4ms, **0 moves**). So we run the loop ourselves. Full trail: `docs/specs/SPEC_WINDOW_DRAG_MANUAL_MOVE_LOOP_2026_05_29.md` (+ `..._NATIVE_MOVE_LOOP`).

## Verification
Runtime-tested via `task dev` on Win11: drags now process **moves=232..1004** per drag at input rate — smooth, and the window stays fully interactive after (the dispatched terminating `WM_LBUTTONUP` keeps Chromium's input state balanced). Native is default; legacy JS path stays behind `localStorage['agentmux.win32NativeDrag']='0'` for one release.

## Follow-ups
- Aero Snap not yet wired for the manual loop.
- Esc cancels + restores start position.
- Floating-pane drag/resize could use the same technique (analysis pending).

### DIFF
--- a/.changesets/1780128496-feat-window-drag-windows-title-bar-drag-via-host-side-native-26s2.md
+++ b/.changesets/1780128496-feat-window-drag-windows-title-bar-drag-via-host-side-native-26s2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(window-drag): Windows title-bar drag via host-side native move loop (VS Code-smooth, zero per-move IPC)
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #1182, opened 2026-05-30
feat(macos): floating-pane tear-off — chromeless floater + draggable header

a5af:
## Summary

Tearing a pane off on **macOS** produced a full workspace window (tab bar + widget bar); Windows produced a chromeless floating window. This makes macOS match — **"just the pane"** — and makes the floater **draggable** by its header.

Spec: [`docs/specs/SPEC_MACOS_FLOATING_PANE_TEAROFF_2026_05_29.md`](https://github.com/agentmuxai/agentmux/blob/agenta/macos-floating-pane-tearoff-phase-a/docs/specs/SPEC_MACOS_FLOATING_PANE_TEAROFF_2026_05_29.md) (Phase A + the Phase B.2 header-drag fix).

## The key insight (why this is small)

The prior cross-platform spec assumed macOS needed a ~400-LOC raw `NSPanel` rewrite. It doesn't, because:
- macOS secondary windows are **already frameless** CEF Views windows (`window_create_top_level(frameless=true)`). There's no native macOS title bar — the tab bar + widget bar are **pure frontend** (`<Workspace>` → `WindowHeader` + `SystemStatus`).
- The chromeless renderer `<FloatingPaneWorkspace>` and the `?floatingPaneId=` switch in `app.tsx` are **100% platform-agnostic**.

So "just the pane" needed only the torn-off window's URL to carry `?floatingPaneId=`. No AppKit.

## Phase A — route pane tear-off to the chromeless floating window

- **`CrossWindowDragMonitor.darwin.tsx`** — pane tear-off now calls `open_floating_pane_window` (chromeless), mirroring the Windows pane branch: measure source pane size → `TearOffBlock` → IPC-first → delete the docked layout node on success. Tab tear-off unchanged (full window). Drop position comes from the DOM `dragend` event's `screenX/Y` (top-left-origin CSS px = DIP, exactly what CEF Views wants) since the host `get_cursor_point` is a Windows-only `GetCursorPos` stub.
- **`commands/floating_pane.rs`** — the non-Windows branch (was "not implemented") now creates a frameless window via the same `post_create_window(frameless=true)` path the tear-off cold path uses, with `&floatingPaneId=` in the URL. No DPI scaling on non-Windows (CEF Views is DIP).
- **`tear-off-pool-helper.ts`** — extracted the shared `measureSourcePaneSize` + floater-size constants (were win32-local); win32 now imports them.

## Phase B.2 — make the floater header-draggable on macOS

The floating-pane drag is absolute (read base via `get_window_position`, then `set_window_position(base + delta)`). Two macOS bugs made it unusable:
1. `get_window_position` returned `{0,0}` on macOS (Windows-only `GetWindowRect`) → wrong baseline → window jumped to the corner.
2. The frontend multiplied the CSS delta by `devicePixelRatio` — correct for Windows (physical-px `SetWindowPos`) but wrong for CEF Views (DIP), so on Retina the window moved **2× too fast**.

Fixes:
- **Host** — `get_window_position` now reads the CEF Views window's DIP bounds on macOS/Linux. Because `bounds()` must run on the CEF UI thread but the IPC command runs on the IPC thread, a new `GetWindowPositionTask` + `get_window_position_blocking` bounces through the UI thread over a bounded channel (250 ms timeout). `set_window_position` already used `set_bounds` (DIP) — no change.
- **Frontend** — `posScale()` returns `devicePixelRatio` only on Windows (physical px) and `1` on macOS/Linux (CEF Views DIP), replacing the raw `devicePixelRatio` in the move-delta math.

## Scope / platform impact

- **Windows**: unchanged — its pane path already used `open_floating_pane_window` with raw-HWND physical-px positioning.
- **Linux**: the host floating-pane branch + `posScale()` already handle Linux, but the **Linux monitor still routes panes through the full-window path** in Phase A (its monitor is unchanged) — a trivial follow-up.
- Out of scope (tracked in the spec as Phase B.3 / B.4): macOS **redock** (`resolve_window_at_cursor` is still a macOS stub) and **edge-resize** parity (same DIP coordinate issue as the drag).
- Pre-existing, unrelated: `create_isolated_request_context` logs `cache_path is not a child of root_cache_path → in-memory storage` for **every** macOS secondary window (tear-off tabs too) — not introduced here; its own follow-up.

## Test plan

Verified on macOS 26 / Apple Silicon / cef 146.7.0:
- [x] Tear a pane → frameless window showing **only the pane** (no tab bar, no widget bar), at the drop point, sized to the source pane.
- [x] Drag the floater by its header → follows the cursor 1:1 (no 2× overshoot, no corner-jump).
- [x] Tear a **tab** → still a full workspace window (unchanged).
- [x] Host + frontend compile/type-check clean.
- [ ] Windows: no behavioral change (paths unchanged).

### DIFF
--- a/.changesets/1780129794-feat-macos-floating-pane-tear-off-chromeless-floater-working-x4i0.md
+++ b/.changesets/1780129794-feat-macos-floating-pane-tear-off-chromeless-floater-working-x4i0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(macos): floating-pane tear-off — chromeless floater + working header drag
--- a/agentmux-cef/src/commands/floating_pane.rs
+++ b/agentmux-cef/src/commands/floating_pane.rs
@@ -154,14 +154,76 @@ pub fn open_floating_pane_window(
 
     #[cfg(not(target_os = "windows"))]
     {
-        // Phase 1 is Windows-only per spec §10. macOS uses
-        // `NSWindow::addChildWindow`; Linux varies. Both are explicit
-        // follow-ups; return a clear error so callers fail loudly.
-        let _ = parsed;
-        let _ = window_label;
-        Err(
-            "open_floating_pane_window: not yet implemented on this platform (Windows only in Phase 1)"
-                .to_string(),
-        )
+        // Phase A (SPEC_MACOS_FLOATING_PANE_TEAROFF_2026_05_29.md): create a
+        // frameless top-level window via the SAME CEF Views path the regular
+        // tear-off uses (`open_window_at_position` →
+        // `ui_tasks::post_create_window(..., frameless=true)`), but with a
+        // `floatingPaneId` in the URL so the frontend renders
+        // `<FloatingPaneWorkspace>` (chromeless: no tab bar, no widget bar)
+        // instead of `<Workspace>`. Secondary windows are already frameless on
+        // macOS/Linux (window/creation.rs), so this alone produces "just the
+        // pane" — the user's request.
+        //
+        // No DPI scaling on non-Windows: the Windows-only block above is
+        // skipped, and CEF Views positions/sizes in DIP (logical px), which is
+        // exactly what `width`/`height` (from `getBoundingClientRect`) and
+        // `x`/`y` (from the DOM drop event's `screenX/Y`) already are.
+        //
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #1183, opened 2026-05-30
docs(macos): floating-pane redock — analysis + implementation plan

a5af:
## Summary

Standalone investigation (requested) of why **redock** — dropping a floating pane back onto another AgentMux window to merge it into that window's tab — works on Windows but not macOS. **Report only; no implementation** (that's a separate follow-up PR).

Report: [`docs/analysis/REPORT_MACOS_FLOATING_PANE_REDOCK_2026_05_30.md`](https://github.com/agentmuxai/agentmux/blob/agenta/macos-floating-pane-redock-report/docs/analysis/REPORT_MACOS_FLOATING_PANE_REDOCK_2026_05_30.md)

Builds on PR #1182 (macOS tear-off — chromeless floater + draggable header); this is the Phase B.3 deep-dive the tear-off spec deferred.

## Does redock integrate with the global reducer system?

**Yes — three layers, cleanly separated. The mutation is reducer-integrated and already cross-platform; the window resolution depends on host projections not populated on macOS.**

1. **Backend (agentmux-srv) reducer + WaveObject graph — the actual effect.** `RedockFloatingPane` is a saga dispatching `Command::MoveBlock` to the backend reducer (block leaves source tab's `block_ids`, joins target tab). Platform-agnostic; **already works on macOS.** The frontend reads the same WaveObject graph to resolve the target tab.
2. **Host (agentmux-cef) reducer — window lifecycle** (`HostCommand::BrowserRegistered`, `EnqueuePendingWindowCreation`). Redock reads host state, doesn't add commands.
3. **Launcher-event → host shadow projection.** `resolve_window_at_cursor` returns the target's `window_id` from `backend_window_id(label)` → `shadow_backend_window_ids`, fed **only by the Windows launcher** (`apply_event_to_shadow` is dead code on macOS). **So `backend_window_id` is always `None` on macOS** → redock's target `window_id` resolves to null and the frontend bails.

## The macOS gaps (none require new reducer/saga work)

1. **`resolve_window_at_cursor` has no macOS impl** — Windows walks HWND Z-order; macOS needs a CEF Views `bounds()` hit-test over `state.windows` (UI-thread read, same blocking-channel pattern PR #1182 added).
2. **`backend_window_id(label)` is always `None` on macOS** (no launcher to feed the shadow map). Fix: `register_backend_window` populates it directly on non-Windows (~10 LOC) — the frontend already hands the host the mapping.
[...]

PR #1185, opened 2026-05-30
feat(macos): floating-pane redock — drop a floater onto a window to merge the pane

a5af:
## Summary

Dragging a floating pane over another AgentMux window and dropping it now **merges the pane into that window's active tab** (with a live drop-zone highlight), and the emptied floater **auto-closes** — matching Windows. Implements the plan in [`REPORT_MACOS_FLOATING_PANE_REDOCK_2026_05_30.md`](https://github.com/agentmuxai/agentmux/blob/main/docs/analysis/REPORT_MACOS_FLOATING_PANE_REDOCK_2026_05_30.md) (PR #1183).

Follows PR #1182 (macOS tear-off — chromeless floater + draggable header). **Verified working on macOS.**

## The redock mutation was already cross-platform

`RedockFloatingPane` is a saga over the backend (agentmux-srv) reducer dispatching `Command::MoveBlock`; the frontend reads the same WaveObject graph to resolve the target tab. **No new backend-reducer or saga code** — the macOS gaps were all in window *resolution*:

### Gap 1 — `resolve_window_at_cursor` had no macOS impl
Add a CEF Views hit-test: `ResolveWindowAtCursorTask` / `resolve_window_at_cursor_blocking` (`ui_tasks.rs`) iterate the registered top-level windows on the UI thread and hit-test the DIP cursor against each `bounds()`, excluding the drag source. Overlap rule (pragmatic first cut): prefer a non-`main` match over `main`; true Z-order among overlapping non-main windows is a follow-up (would need `[NSApp orderedWindows]` + a label↔NSWindow registry). Wired into the non-Windows branch of `resolve_window_at_cursor` (`motion.rs`); dispatch wrapped in `spawn_blocking` (`ipc.rs`) — along with `update_floating_redock_hover` which calls it — since both now block the UI thread up to 250 ms (same pattern as `get_window_position` in #1182).

### Gap 2 — `backend_window_id(label)` was always `None` on macOS
`shadow_backend_window_ids` is fed only by the Windows launcher (`apply_event_to_shadow`). Without a launcher, populate it directly in `register_backend_window` on non-Windows (`meta.rs`), so the resolved target carries a real `window_id` — else the frontend redock bails (`if (!target.label || !target.window_id) return`).

### Gap 3 — coordinate space
The frontend now sends the redock cursor in DIP on macOS/Linux (`posScale()` in `pushRedockHover` + `tryRedockAtCursor`, `floating-pane-workspace.tsx`), and the hover-highlight receiver (`app-init.ts`) inverts it platform-aware (divide by DPR only on Windows) so the drop-zone highlight lands under the cursor on Retina.

## Scope

- **Windows**: unchanged (HWND Z-order walk + physical px throughout).
- Out of scope (follow-ups): true Z-order among overlapping non-main windows (Gap 1 note); edge-resize DIP parity (same coordinate issue as the drag, Phase B.4 in the spec); the pre-existing `cache_path` in-memory-storage warning on all macOS secondary windows.

## Test plan

Verified on macOS 26 / Apple Silicon:
- [x] Tear a pane → drag the floater over the main window → drop-zone highlight appears → release → pane merges into main's active tab, floater auto-closes.
- [x] `registered backend window ID label=main window_id=…` populates the shadow map (Gap 2).
- [x] Drop over the desktop → floater stays put, no errors.
- [x] Host + frontend compile / type-check clean.
- [ ] Windows: no behavioral change (HWND path untouched).

### DIFF
--- a/.changesets/1780144758-feat-macos-floating-pane-redock-drop-a-floater-onto-a-window-6qyb.md
+++ b/.changesets/1780144758-feat-macos-floating-pane-redock-drop-a-floater-onto-a-window-6qyb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(macos): floating-pane redock — drop a floater onto a window to merge the pane
--- a/agentmux-cef/src/commands/window/meta.rs
+++ b/agentmux-cef/src/commands/window/meta.rs
@@ -199,6 +199,21 @@ pub fn register_backend_window(_state: &Arc<AppState>, args: &serde_json::Value)
         // `Event::BackendWindowIdRegistered` (delivered via the CEF
         // JS bridge) carries the label → windowId mapping change to
         // every renderer's reducer. No sync emit here.
+        //
+        // On non-Windows there is NO launcher, so the
+        // `Event::BackendWindowIdRegistered` → `apply_event_to_shadow` round
+        // trip that fills `shadow_backend_window_ids` never runs
+        // (`report_backend_window_id_registered` no-ops without a launcher
+        // channel). The host is the sole authority here, so populate the
+        // projection directly. `resolve_window_at_cursor` reads it via
+        // `backend_window_id(label)` to resolve a redock target's window_id;
+        // without this it is always `None` on macOS/Linux and redock silently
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #1189, opened 2026-05-30
docs(analysis): blur audit — backdrop-filter is not an input-first hotspot (input-first 0.3)

AgentY-asaf:
## What

Phase 0.3 "cheap verified win" from [discussion #1161](https://github.com/agentmuxai/agentmux/discussions/1161): the **`backdrop-filter: blur` audit**.

## Finding: blur is **not** an input-first hotspot — no code change needed

Audited all 10 `backdrop-filter` / `filter: blur` sites in `frontend/`. Every one is:
- **`<Show>`-gated** (absent from the DOM when inactive): connstatus-overlay (disconnect), magnified/ephemeral-node-backdrop, modal-backdrop, or
- **interaction-transient**: `:hover` (copy buttons, codeblock actions), drag, resize, magnify, or
- the **single always-on** `.block-mask { blur(0.1px) }` — a deliberate, documented compositing hack (promotes the focus-ring layer above xterm's scroll layer); 0.1px ⇒ layer promotion, not Gaussian work.

**No backdrop blur is live over a pane during steady-state typing**, so blur cannot be taxing the ≤16 ms keystroke frame. Verdict: *reviewed, nothing to fix.*

Full per-site table + render-guard tracing in the doc. This closes the blur cheap-win and documents it so it isn't re-investigated; if a future regression points at blur, the first check is whether a new always-mounted (`<Show>`-less) blurred element was added over a pane.

### Optional, deliberately not done
`.connstatus-overlay` uses `blur(50px)` (larger radius than the effect needs). Left as-is: the upside applies only while disconnected (negligible for input responsiveness) and it would alter the overlay's look. Recorded in the doc as a known intentional choice.

## Scope
Docs-only (one analysis file). Changeset: `patch`.

### DIFF
--- a/.changesets/1780147527-docs-blur-audit-input-first.md
+++ b/.changesets/1780147527-docs-blur-audit-input-first.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(analysis): blur audit — backdrop-filter is not an input-first hotspot (input-first 0.3, #1161)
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1187, opened 2026-05-30
docs: spec — integrate agentmux-launcher into macOS/Linux task dev (Phase 7)

a5af:
## Summary

The launcher runs in the loop on **Windows** `task dev` but is bypassed on **macOS/Linux** (the host is invoked directly; the launcher's Unix `main()` is a stub that `exec`s the host). [`SPEC_LAUNCHER_DEV_INTEGRATION_2026-05-13.md`](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/SPEC_LAUNCHER_DEV_INTEGRATION_2026-05-13.md) explicitly deferred Unix to "Phase 2 / Phase 7." **This spec is that phase.**

**Report only; no code** (implementation follows).

Spec: [`SPEC_LAUNCHER_MACOS_DEV_INTEGRATION_2026_05_30.md`](https://github.com/agentmuxai/agentmux/blob/agenta/launcher-macos-dev-spec/docs/specs/SPEC_LAUNCHER_MACOS_DEV_INTEGRATION_2026_05_30.md)

## Key finding

The launcher is **Windows-biased in implementation, portable in architecture.**

**Already cross-platform (no work):** saga coordinator, SQLite saga durability + startup recovery, event bus + disk log, path resolution (`agentmux_common::DataPaths`), env-var handoff, srv binary resolution, ESTART handshake, host supervision loop (crash-budget retry + `--disable-gpu` rung). The host already detects launcher-managed mode via `use_launcher_endpoints()` — **zero host changes** needed once the Unix launcher passes `AGENTMUX_BACKEND_*`.

**Windows-locked → needs a Unix port (~5 modules):**
| Piece | Win32 | Unix |
|---|---|---|
| IPC transport | named pipe | `tokio::net::UnixListener` (`launcher.sock`) |
| Single-instance | `first_pipe_instance` bind | `flock` on `launcher.lock` (auto-frees on crash — no stale-PID issue) |
| Process-tree containment | Job Object `KILL_ON_JOB_CLOSE` | process group (`setpgid`+`killpg`) **+** srv's existing kqueue/pidfd parent-death watcher as backstop |
| Suspend/resume spawn | `CREATE_SUSPENDED`+`ResumeThread` | drop — spawn normally |
| Splash | layered popup | skip on Unix |

## Phasing

[...]

PR #1193, opened 2026-05-30
feat(launcher): drive srv + host via agentmux-launcher on macOS/Linux task dev (Phase 1)

a5af:
## What

Integrates **agentmux-launcher** into `task dev` on **macOS/Linux** — previously the launcher was in the loop only on Windows. The launcher now spawns srv, hands the host its backend endpoints via env, and supervises both with the same crash budget Windows uses. So the saga coordinator / srv-lifecycle / host-supervision paths are now exercised in dev on every platform.

Implements **Phase 1** of `SPEC_LAUNCHER_MACOS_DEV_INTEGRATION_2026_05_30.md` (#1187, the spec — review/merge that separately). Continues the deferred Phase 7 of the original launcher dev-integration spec.

`task dev:standalone` remains the direct host-invoke escape hatch.

## How

**`agentmux-launcher/src/main.rs`**
- `run_unix()` — Unix supervisor mirroring `run_windows` minus Windows-only machinery (Job Object, `CREATE_SUSPENDED`/resume, named-pipe handle, splash). Resolves data dirs → spawns srv via the already-Unix-clean `spawn_srv` → spawns host with `AGENTMUX_BACKEND_*` in env → supervises both with the shared `HOST_RESTART_BUDGET` crash budget (+ `--disable-gpu` degraded rung).
- `spawn_host_unix()`, `terminate_child_gracefully()` (SIGTERM so CEF reaps its render subprocesses; SIGKILL grace-window backstop), SIGINT/SIGTERM handlers. srv's stdin write-end is held for the launcher's lifetime (its EOF is srv's parent-death backstop).
- `runtime_dir` falls back to `exe_dir` when there's no `runtime/` subdir → **flat dev layout**, keeping the host's `../Frameworks` resolution + asset anchoring **byte-identical** to the legacy direct-invoke path.

**Two gotchas found + fixed during bring-up:**
1. **Launcher fork bomb** — in the flat layout the launcher is the host's *sibling*, and `find_cef_binary`'s scan matched `agentmux-launcher` (it's `agentmux-*`, not `-cef`/`-srv`), so the launcher resolved **itself** as the host and spawned recursively. Fixed by excluding `agentmux-launcher` from the scan, plus a canonicalized self-spawn guard as a loud backstop.
2. **Double srv** — a dev-build host deliberately ignores `AGENTMUX_BACKEND_WS` (it may be a stale value inherited from a parent agentmux pane), so it double-spawned its own srv next to the launcher's. Fixed with `launcher_is_genuine_parent()` in **`agentmux-cef/src/main.rs`**: the launcher stamps `AGENTMUX_LAUNCHER_PID` (= its pid == the host's `getppid`, since it spawns the host directly), and the host adopts the launcher's srv only when that matches. A stale inherited value can't satisfy the `getppid` match, so `task dev:standalone` still spawns its own srv. Unix-only; **Windows unchanged**.

**`Taskfile.yml`** — `build:host:darwin` builds + stages `agentmux-launcher`; `dev:serve`'s Unix branch co-locates launcher + host + srv in `dist/cef-dev/` (flat) and invokes `./agentmux-launcher`.

**Deps** — launcher gains a unix-only `libc` (graceful SIGTERM) + the tokio `signal` feature.

## Verified on macOS

- Single **launcher → {host, srv}** tree — host **adopts** the launcher's srv, no double-spawn; host's only children are CEF's own helpers (gpu/network/storage/renderers).
- Launcher log: `resolved CEF binary: …/agentmux-cef` → `srv … ready` → `spawned CEF host pid=… (unix)` → `entering supervised host + srv wait (unix)`.
- Frontend boots (startup-bench completes, renderers up).
- `SIGTERM` tears the whole tree down cleanly: `received SIGTERM → terminating children (SIGTERM → grace → SIGKILL) → launcher exiting with code 0`.
- Per-branch dev data dir (`~/.agentmux/dev/<branch>/`) preserved → test-session continuity.

## Out of scope (later phases)
- Phase 2: flock single-instance + Unix-socket IPC transport (srv launched without a pipe path here, matching today's dev mode).
- Phase 3: supervision/cleanup parity (host parent-death backstop for the launcher-SIGKILL case).

### DIFF
--- a/.changesets/1780148517-feat-launcher-drive-srv-and-host-via-agentmux-launcher-on-ma-83i3.md
+++ b/.changesets/1780148517-feat-launcher-drive-srv-and-host-via-agentmux-launcher-on-ma-83i3.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(launcher): drive srv and host via agentmux-launcher on macOS/Linux task dev (Phase 1)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,6 +76,7 @@ dependencies = [
  "agentmux-common",
  "chrono",
  "dirs",
+ "libc",
  "png",
  "proptest",
  "rusqlite",
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -466,10 +466,16 @@ tasks:
         # before the macOS libcef story was sorted; on Linux we always had
         # the patched runtime).
         cmds:
-            - bash -c 'source "$HOME/.cargo/env" 2>/dev/null; PATH="$HOME/.cargo/bin:$PATH" cargo build --release -p agentmux-cef --features patched-libcef'
+            # Phase 1 (SPEC_LAUNCHER_MACOS_DEV_INTEGRATION_2026_05_30): build
+            # the launcher alongside the host so the flat dev layout (launcher
+            # + host + srv co-located in dist/cef-dev/) has it — same as
+            # build:host:darwin. Without this, dev:serve's launcher-not-found
+            # guard hard-fails on Linux.
+            - bash -c 'source "$HOME/.cargo/env" 2>/dev/null; PATH="$HOME/.cargo/bin:$PATH" cargo build --release -p agentmux-cef --features patched-libcef && PATH="$HOME/.cargo/bin:$PATH" cargo build --release -p agentmux-launcher'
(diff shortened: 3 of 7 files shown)


a5af:
@codex review

PR #1186, opened 2026-05-30
fix(dnd): suppress tear-off drag snapback on macOS/Linux (pane + tab)

a5af:
## Summary

Dragging a pane or tab out to tear it off played WebKit/WebKitGTK's "drop rejected" **snapback** animation — the drag ghost flying back into the source window ("it acts like it doesn't want to go") — even though the tear-off completed. The drag is released **outside any pragmatic-dnd drop target** (the new floater/window is created on `dragend`), so the browser treats the drop as unhandled and snaps the preview back to origin.

Fix: pragmatic-dnd's [`preventUnhandled.start()`](https://atlassian.design/components/pragmatic-drag-and-drop/core-package/utilities/prevent-unhandled) on drag start (stopped in `onDrop`). It makes every element a drop target in the browser's eyes, so the drop is "handled" → **no snapback**. The drag preview still follows the cursor during the drag (not suppressed); in-window rearrange still works via pragmatic-dnd's own drop targets.

## Changes

- **Pane**: `TileLayout.darwin.tsx` + `TileLayout.linux.tsx` (platform-split; **win32 unchanged**).
- **Tab**: `droppable-tab.tsx` (shared) gated `!isWindows()`.

## Scope

Scoped to **macOS/Linux**, matching PR #1175 (the move-cursor fix): on Windows the SC_MOVE tear-off path relies on the current behavior and the move cursor there isn't fixed, so suppressing the snapback could leave worse feedback. The stale "don't suppress the OS ghost — it leaves a no-drop cursor with zero feedback" comment in `droppable-tab.tsx` is now moot — #1175 fixed the cursor, and this keeps the ghost (just no snapback).

## Test plan

- [ ] macOS: tear a **pane** out → no ghost flying back into the source window; the floater appears at the drop point.
- [ ] macOS: tear a **tab** out → same, no snapback.
- [ ] macOS: rearranging a pane/tab **within** the window still works (pragmatic-dnd drop targets unaffected).
- [ ] Windows: no behavioral change (`!isWindows()` gate / win32 files untouched).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1194, opened 2026-05-30
fix(macos): suppress AppKit drag slide-back animation on pane/tab tear-off

a5af:
## Symptom

After tearing off a pane (or tab) on macOS, the drag image **animates back into the source window** ("slide-back" / fly-back) before the new floating window appears — an inappropriate "rejection" animation. Still present after PR #1186.

## Root cause

A tear-off releases the pointer **outside any DOM drop target** — the floating window is created on `mouseup`. So blink hands AppKit `NSDragOperationNone`, and AppKit plays its native `NSDraggingSession` slide-back animation (`animatesToStartingPositionsOnCancelOrFail`, default `YES`).

PR #1186's `preventUnhandled` only suppresses the **WebKit-level** snapback for **in-document** drops — it can't reach this **AppKit-level** animation, which is why the animation survived.

## Fix

CEF/Chromium starts every drag via `-[NSView beginDraggingSessionWithItems:event:source:]` and never flips the flag. We swizzle that method (raw-libobjc FFI, mirroring the existing `patch_nsapp_unrecognized_selector` Tahoe shim): call the original to create the session, then set `animatesToStartingPositionsOnCancelOrFail = NO` on the returned session. Swizzled at the `NSView` level so it covers whichever Chromium content view starts the drag.

The flag **only** affects the cancel/fail slide-back — successful in-window drops (e.g. tab reorder) are unaffected — so disabling it app-wide is safe; AgentMux never wants a drop-rejected slide-back.

## Scope / safety
- macOS-only (`#[cfg(target_os = "macos")]`); no effect on Windows/Linux.
- Runs once at startup, before `cef::initialize`, on the main thread.
- Single file (`agentmux-cef/src/main.rs`); complements (does not replace) #1186's frontend `preventUnhandled`.

## Verification
Build + `task dev`, tear off a pane/tab → the slide-back animation is gone; the floating window appears at the release point with no fly-back. (Manual GUI confirmation — please verify on your machine.)

### DIFF
--- a/.changesets/1780149042-fix-macos-suppress-appkit-drag-slide-back-animation-on-pane--m764.md
+++ b/.changesets/1780149042-fix-macos-suppress-appkit-drag-slide-back-animation-on-pane--m764.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): suppress AppKit drag slide-back animation on pane/tab tear-off
--- a/agentmux-cef/src/main.rs
+++ b/agentmux-cef/src/main.rs
@@ -266,6 +266,19 @@ fn main() {
     #[cfg(target_os = "macos")]
     unsafe { patch_nsapp_unrecognized_selector() };
 
+    // macOS tear-off polish: suppress AppKit's native drag slide-back
+    // ("poof" / fly-back-to-origin) animation. When a pane/tab is torn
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1195, opened 2026-05-30
fix(window): canonical label-based window resolution (P1)

AgentA-asaf:
## P1 — canonical label-based window resolution

Implements **P1** from `docs/architecture/ARCHITECTURE_FLOATING_PANE_DOCKING_2026_05_30.md` §9: collapse window-HWND resolution behind the canonical `resolve_window_hwnd(label)` and **eliminate `find_own_top_level_window` from call sites that have a label** (it returns the process's first-visible top-level — the *floater* when one exists — root cause #1 of the recurring docking regressions).

### What changed
- **minimize / maximize / drag** (`chrome.rs`, `motion.rs`) — resolve via `resolve_window_hwnd(label)`, not `find_own_top_level_window`, so they act on the labeled window instead of whatever floater the OS enumerates first.
- **browser-pane creation parent** (`browser_pane/creation.rs`) — parent comes from `window_label` via `resolve_window_hwnd`, so a redocked pane is never parented to the dying floater (child-before-parent destroy → `ERR_ABORTED` → black render).
- **redock-onto-main** (`resolve_window_at_cursor`) — see below; the headline fix.
- `find_own_top_level_window` gets a ⚠️ doc warning; `resolve_window_hwnd` widened to `pub(crate)`.

### The redock-onto-main fix (proven this session)
Long-standing symptom: **dragging a floater onto the main window showed no landing ghost and never docked** (terminal *and* browser panes). Proven via the new `redock-resolve` trace:

- Every fresh launch boots windows in `pool mode — deferring init until promote`, then **promotes a warm-pool window on-screen to serve as main but never relabels it**. `window_hwnds` keeps `window-pool-<uuid> → its HWND` while `main` is left with **no live HWND** (`resolve_window_hwnd("main")` logs *"no available HWND for label=main"*).
- So `find_main_window` returns the pool HWND, and `resolve_window_at_cursor` handed back the **stale pool label** — which never matched the target window's frontend identity (`myLabel = ?windowLabel ?? "main"`), so the ghost gate (`target_label === myLabel`) failed and the drop had no valid `main` target.
- **Fix:** when the HWND under the cursor is the cache-independent main frame (`find_main_window`) **and** its reverse-map label is a lingering `window-pool-*`, resolve it as `"main"`, reconciling the OS-HWND cache with the frontend's identity. Real floaters / additional windows are untouched (zero multi-window regression surface).

The deeper cure — make **promotion relabel** the window so the cache, the backend, and the frontend agree — is captured as a P1/P2 follow-up in the architecture doc.

### Instrumentation (permanent, kept per the regression history)
- `browser_pane::trace` `[pane-trace]` lifecycle log.
- `redock-resolve` window-resolution trace (`muxlog host redock-resolve`) — logs `find_main_window`'s answer vs. the window under the cursor + `is_main_hwnd`.

### Verification
- `cargo check -p agentmux-cef --release` clean; `cargo test -p agentmux-cef` → **112 passed, 0 failed**.
- Built `--fresh` and smoke-tested end-to-end: floater→main now shows the landing ghost **and** docks (`workspace.RedockFloatingPane` completes); floater→floater unchanged.

### DIFF
--- a/.changesets/1780145316-fix-window-resolve-minimize-maximize-drag-pane-parent-by-win-xqs3.md
+++ b/.changesets/1780145316-fix-window-resolve-minimize-maximize-drag-pane-parent-by-win-xqs3.md
@@ -1,0 +1,19 @@
+---
+type: patch
+---
+
+fix(window): canonical label-based window resolution (P1)
+
+Route minimize / maximize / drag / browser-pane-parent through the canonical
+`resolve_window_hwnd(label)` instead of `find_own_top_level_window`, which
+returns the process's first-visible top-level — the floater when one exists,
+so those actions hit the wrong window.
+
+Also fixes redock-onto-main silently failing (no landing ghost, no dock): a
+warm-pool window promoted on-screen to serve as main keeps its `window-pool-*`
+label in the HWND cache while `main` is left with no live HWND, so
+`resolve_window_at_cursor` handed back the stale pool label and it never matched
+the target window's frontend `main` identity. It now resolves the
+cache-independent main frame (`find_main_window`) as `main` even when the
+reverse-map label is a lingering pool label. Adds permanent `redock-resolve`
+and `browser_pane` lifecycle instrumentation (kept, per the regression history).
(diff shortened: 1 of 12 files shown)


a5af:
@codex review

PR #1192, opened 2026-05-30
fix(window): gate macOS/Linux floating-pane redock resolver off Windows (restore Windows build)

AgentX-asaf:
## What

Floating-pane **maximize/restore**, reducer-backed, built fresh on `main` — the floating half of the shared maximize button from `SPEC_PANE_STATE_REDUCER §3.3a`. Docked panes keep layout magnify; a torn-off floating pane now gets an OS-window maximize/restore.

This supersedes the parked resize work in #1132. It deliberately ships **only** maximize (a `ShowWindow` toggle) and **not** edge-resize — so it carries none of #1132's redock risk.

## Commits

1. **`fix(host-reducer)`** — re-key pane window-placement by the floater's window label (`floating-<uuid>`), not `block_id`. Floaters live in `AppState.window_hwnds`, never in `browser_panes`, so the original block_id keying + `browser_panes` co-eviction could never fire for them (latent leak). Adds `EvictFloatingPaneWindowState { label }`, dispatched from `client/mod.rs::on_before_close` (gated to `floating-` labels) — the floater's real teardown hook. Reverts the dead block_id co-eviction from `panes.rs`. Reducer tests rewritten label-keyed (46 pass, 6 in `pane_window`).
2. **`feat(floating-pane)`** — the maximize button + wiring.

## How

- **Host** (`commands/window/chrome.rs`): new `toggle_floating_maximize` IPC handler dispatches `HostCommand::ToggleFloatingMaximize { label }`, then applies `ShowWindow(SW_MAXIMIZE/SW_RESTORE)` **after** dispatch from the emitted `PaneWindowStateChanged` (snapshot-and-drop — no I/O inside the pure reducer, mirroring `set_window_opacity`). Resolves the floater's outer HWND via `resolve_window_hwnd(window_hwnds)`. Returns `{ placement }` so the button syncs its icon (no host→frontend push channel on `main`). Registered in `ipc.rs` `route_command`.
- **Frontend** (`blockframe.tsx` `EndIcons`): `FloatingMaximizeButton`, shown only when the window carries a `?windowLabel=floating-…` param. Reuses `MagnifyIcon` so maximize and magnify read as one operation.

Windows' own `WINDOWPLACEMENT` remembers the pre-maximize rect across `SW_RESTORE`, so basic maximize/restore needs no `last_known_normal_rect` tracking (that's a later reducer phase).

## Test plan

- [x] `cargo build -p agentmux-cef` — clean
- [x] reducer unit tests — 46 pass
[...]

PR #1196, opened 2026-05-30
fix(airspace): clip DOM overlays correctly over browser panes at display scale != 100%

AgentX-asaf:
## What

DOM overlays (right-click context menus, submenus, tooltips, popovers) over a **browser pane** showed **black spots**, appeared **offset**, or were **hidden** at display scales other than 100%. Root cause is a coordinate-unit mismatch in the airspace clip.

## Root cause

- Overlay rects are produced from `getBoundingClientRect()` — **CSS pixels**, no `devicePixelRatio` multiply (`pane-overlay.ts`; the contract is even spelled out in `pane-rect-registry.ts`).
- The host's `browser_panes.rs::set_pane_overlay_clip` computes each pane's geometry from `GetWindowRect` — **physical pixels** — and subtracts the overlay rects **directly, with no DPI scaling**.

At `dpr = 1` (100% scale) these match. At 125%/150%/… the punched `SetWindowRgn` hole lands **offset and undersized** relative to where the overlay actually paints:
- hole lands where the menu does **not** paint → see-through to the parent window = **black spot**;
- menu paints where **no** hole was punched → native pane covers it = **hidden**;
- net = menu and its show-through region **drift apart** = **"offset."**

## Fix

Convert the dispatched rects to physical px with `Math.round(v * dpr)` — mirroring `browser-view.tsx::paneRect`'s **exact** convention so the hole and the pane HWND share rounding and never leave a 1px seam. The CSS-px intersection gate (`pane-rect-registry`) is intentionally left in CSS px. One-file change.

```ts
const dpr = window.devicePixelRatio || 1;
const rectsToSend = (intersects ? rects : []).map((r) => ({
    x: Math.round(r.x * dpr), y: Math.round(r.y * dpr),
    w: Math.round(r.w * dpr), h: Math.round(r.h * dpr),
}));
```

## Verification

Built and run on Windows (`task dev`) at a **non-100% display scale**. Before: black spots / offset / hidden context menus over a browser pane. After: context menus **and submenus** clip cleanly and are correctly positioned. Confirmed by the reporter.

## Scope & docs

- Windows airspace path only; macOS/Linux use the Views hide-whole-pane path (untouched) and are DPI-correct already.
- Includes `docs/analysis/ANALYSIS_BROWSER_PANE_AIRSPACE_ARCHITECTURE_2026_05_30.md` — a full reassessment of the airspace architecture (why rect-punching has inherent fidelity limits, and an OSR/shared-texture path as the principled long-term fix). This PR is the immediate, surgical fix; the architecture discussion is separate.
- Changeset added (`type: patch`); no version-file bump.

## ⚠️ Stacked PR

Based on `agentx/fix-windows-build-redock-cfg` (**#1192**), not `main`, because `main` does not currently compile on Windows (regression from #1185 — see #1192). Stacking keeps this branch buildable/testable on Windows. **Merge #1192 first**, then this retargets cleanly to `main`. The diff shown here is only the airspace change.

### DIFF
--- a/.changesets/1780149540-fix-airspace-send-overlay-clip-rects-in-physical-px-so-dom-o-8wy3.md
+++ b/.changesets/1780149540-fix-airspace-send-overlay-clip-rects-in-physical-px-so-dom-o-8wy3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(airspace): send overlay-clip rects in physical px so DOM overlays clip correctly over browser panes at display scale != 100%
--- a/docs/analysis/ANALYSIS_BROWSER_PANE_AIRSPACE_ARCHITECTURE_2026_05_30.md
+++ b/docs/analysis/ANALYSIS_BROWSER_PANE_AIRSPACE_ARCHITECTURE_2026_05_30.md
@@ -1,0 +1,188 @@
+# Browser-Pane Airspace — Architecture Reassessment
+
+**Date:** 2026-05-30 · **Author:** AgentX · **Trigger:** runtime test of #1178 on Windows surfaced
+"right-click context menus appear wrong around browser panes — black spots in offset menus, menus
+sometimes hidden."
+
+> **Verdict (one line):** Our understanding of the *mechanism* is rigorous and well-documented.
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1197, opened 2026-05-30
docs(spec): terminal flow control (PTY backpressure) design — input-first

AgentY-asaf:
## What

Design spec for the **ACK-based terminal flow control** lever — the genuinely-open xterm.js recommendation from SPEC_INPUT_RESPONSIVENESS §5.1 / PLAN item 7 ([discussion #1161](https://github.com/agentmuxai/agentmux/discussions/1161)). Predictive/mosh echo is shelved; flow control is what actually protects the keystroke frame under heavy output. PLAN item 7 explicitly says this 'needs its own spec doc' — this is it.

## Grounded in the real architecture

Verified the PTY path is **pull-based** (`agentmux-srv/src/backend/blockcontroller/shell.rs` read loop): PTY → `append_output` (FileStore) + `publish_data_event` (WPS) → frontend `term-rpc` pulls. So there's no backpressure today; the FileStore byte cap is the only safety net, and a flood starves keystrokes.

## Design

- **Accounting:** per-block `appended` / `acked` counters; `outstanding = appended − acked`.
- **Pause/resume:** the read loop parks when `outstanding ≥ HIGH_WATERMARK`, resumes below `LOW_WATERMARK` (hysteresis). Parking the PTY read fills the kernel PTY buffer → blocks the child's `write()` → OS backpressure at the producer (xterm flow-control pattern pushed one layer down).
- **Ack channel:** new `termack { blockId, consumedOffset }` WS message; cumulative + monotonic so lost/reordered acks self-heal.
- **Never-block contract (the deadlock-avoidance part):** the pause is bounded + interruptible — `stop()` wakes the parked loop; a `RESUME_TIMEOUT` covers a dead/disconnected client (degrades to today's cap-bounded behavior, never a frozen pane or wedged child); never park while holding the controller inner lock.
- **Validation:** bench #1176 + load generator; gate P95 keystroke echo < 50 ms under sustained output; positive check that a hidden pane actually throttles the producer.

## Scope

Docs-only (design). Implementation (backend accounting + pause/resume + ack handler; frontend consumed-accounting + ack send) follows as separate PRs, profiling-gated per PLAN. Changeset: `patch`.

Note: line-level touch points are listed but exact signatures are to be pinned at implementation — the editor tooling was intermittently returning padded reads during this design pass.

### DIFF
--- a/.changesets/1780149696-spec-terminal-flow-control.md
+++ b/.changesets/1780149696-spec-terminal-flow-control.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(spec): terminal flow control (PTY backpressure) design — input-first (#1161)
--- a/docs/specs/SPEC_TERMINAL_FLOW_CONTROL_2026_05_30.md
+++ b/docs/specs/SPEC_TERMINAL_FLOW_CONTROL_2026_05_30.md
@@ -1,0 +1,136 @@
+# SPEC: Terminal flow control (PTY backpressure)
+
+**Status:** Draft — design, pre-implementation
+**Date:** 2026-05-30
+**Author:** AgentY
+**Tracks:** [`SPEC_INPUT_RESPONSIVENESS_*` §5.1](./SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md) · [`PLAN_INPUT_RESPONSIVENESS_EXECUTION_*` item 7](./PLAN_INPUT_RESPONSIVENESS_EXECUTION_2026_05_29.md) · [discussion #1161](https://github.com/agentmuxai/agentmux/discussions/1161)
+**Implements:** the "ACK-based flow control" terminal lever — the genuinely-open xterm.js recommendation (predictive echo is shelved; see the #1161 update).
+
+---
+
+## TL;DR
+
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentY-asaf:
## Corrected §1 against the real code (addresses all 4 review findings)

Thanks — the review was right on every point. I'd written the "Confirmed in shell.rs" block from an inaccurate read; verified against the tree and rewrote it. Fixes in `649e2751`:

- **[P1] read-loop mechanics** — corrected to the real loop (`shell.rs:776`): `tokio::task::spawn_blocking`, `PTY_READ_BUF_SIZE = 4096` (4 KiB, `shell.rs:87`), single `handle_append_block_file(broker, &block_id, "term", &buf[..n], None)` (`shell.rs:809`). Removed the fabricated `std::thread::spawn` / 64 KiB / `append_output` + `publish_data_event` (those methods don't exist).
- **[P1] push vs pull / FileStore** — corrected: delivery is **inline push via the broker** — `filestore = None` (`shell.rs:814`), bytes ride in the `block:file` append event as `data64` (decoded `termwrap.ts:449`). Dropped the "pull from FileStore" model and the "FileStore byte cap is the safety net" claim; §2.1 accounting renamed `appended`→`published`; §3.4 reconnect rewritten to resubscribe-based.
- **[P2] deadlock/`spawn_blocking`** — §3 reworked: the loop is a `spawn_blocking` task, so park coordination is shutdown-flag + `RESUME_TIMEOUT` wake (no `JoinHandle`-based join), with an explicit "don't pin the blocking pool" note.
- **[P2] WS path** — corrected `agentmux-srv/src/ws/term.rs` → `agentmux-srv/src/server/websocket.rs` (wscommand dispatch at `websocket.rs:419`).

Added a revision note at the bottom of the spec recording the correction. Ready for re-review.

PR #1199, opened 2026-05-30
fix(ui): canonical maximize/restore icons + tighter hamburger fit

Agent1-asaf:
## Summary

- **Pane maximize button**: replace the custom rotating two-arrow SVG with the canonical Windows title-bar pattern — `fa-window-maximize` (single square) when not magnified, `fa-window-restore` (overlapping squares) when magnified. Two distinct glyphs per state, not one animated icon. Reads instantly as "maximize" / "restore".
- **Relabel** action: "Magnify/Minimize" → "Maximize/Restore". Matches the new glyphs *and* matches what users actually call this action.
- **Hamburger SVG** bumped from 18×14 to 26×22 inside the unchanged 28×27 button (per "tight against the borders of its visual container" feedback). Bars are now ~93% wide / ~81% tall — visually filling the button.
- **Quick-tips** entry for "Magnify a Block" gets the same new icon + new label for consistency.
- **Dead code removed**: the `MagnifyIcon` component, its SCSS, and both SVG assets (`magnify.svg`, `magnify-disabled.svg`) are deleted, along with the `.block-frame-magnify svg #arrow1/2` style rules that were targeting them.

Follow-up to #1191.

## Test plan

- [ ] Open any pane → header now shows a single-square icon; tooltip says "Maximize"
- [ ] Click it → icon swaps to overlapping squares; tooltip says "Restore"
- [ ] Click again → returns to single square
- [ ] Hamburger button in the top bar reads as a fuller, denser icon — bars hug the button edges with minimal padding
- [ ] Press Cmd+H (or wherever Quick Tips opens) → "Maximize a Block" row shows the same window-maximize icon
- [ ] Torn-off floating-pane shell still shows its fixed Maximize button (single square, no state animation per the pre-existing spec — see SPEC_PANE_STATE_REDUCER §3.3a comment in blockframe.tsx)
- [ ] Grep confirms no remaining `MagnifyIcon` or `magnify.svg` references anywhere in the bundle

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1191, opened 2026-05-30
feat(ui): fatten pane control + hamburger icons, raise pane body text default to 15px

Agent1-asaf:
## Summary

- Pane header end-icons (close, magnify, mic, etc.) and the top-bar hamburger get fatter inside their existing slots. **No chrome grows in height** — header is locked at 33 px, hamburger button at 28×27 px. The icons just stop floating in oversized slots and start *filling* them (~64–70 % optical fill, hitting the Material 3 / Fluent 2 norm and WCAG 2.2 AA target-size).
- Pane body text default goes from 12/13 → **15 px**. The "most prominent visibility" lever: changes what every new pane looks like out of the box, no toggle required. Existing per-pane and per-setting overrides still win.
- The **widget bar is intentionally untouched** — it is the visual anchor that pane controls converge toward.

Full design rationale (Fitts's law, fill ratio research, the constraint of zero vertical growth, file-by-file change list): `docs/specs/SPEC_PANE_ICON_AND_TEXT_VISIBILITY_2026_05_30.md`.

### What changed
- `theme.scss` — `--header-icon-size` 14→18, `--header-icon-width` 16→20
- `block.scss` — end-icon button 24→26 wide, `font-size: 18px`, centered
- `iconbutton.scss` — rest opacity 0.7→0.85
- `tabbar.tsx` — hamburger SVG 14×12 → 18×14 with 3 px-thick bars
- `term/termSettingsMenu.ts` + `term/termViewModel.ts` — `term:fontsize` default 12→15
- `agent-view.scss` — `--termfontsize` fallback 13→15
- `term.scss` — `.agent-runtime-badge` 10→12

## Test plan

- [ ] Open a tab with a mix of pane types (term, agent, browser, sysinfo, editor); confirm pane headers visually match the widget bar's weight
- [ ] Verify header is still 33 px tall — DevTools "show layout" on `.block-frame-default-header`
- [ ] Verify hamburger button is still 28×27 — the bars are fatter but the button is the same size
- [ ] Narrow a pane to ~180 px width and verify the title still ellipsises sensibly (didn't lose space to the slightly wider end-icons)
- [ ] Open term settings → font size → 10 on one pane, confirm only that pane shrinks (global default stays at 15)
- [ ] Visual check at 100 %, 125 %, 150 % CSS zoom — icons should remain crisp (SVG bars, no rasterization unevenness)
- [ ] Existing `term:fontsize` user settings still take precedence over the new default

[...]

PR #1201, opened 2026-05-30
feat(dnd): file drop on terminal + agent panes (phase 1)

Agent1-asaf:
## Summary

Phase 1 of the file-drop spec. Drop a file from the OS file manager onto a Terminal or Agent pane → it gets copied into the pane's `cmd:cwd` and (for Agent panes) `@filename` is spliced into the composer at the caret.

Spec: [`docs/specs/SPEC_PANE_FILE_DROP_2026_05_30.md`](../blob/agent1/pane-file-drop-phase1/docs/specs/SPEC_PANE_FILE_DROP_2026_05_30.md).

## What changed

**CEF host (Rust)**
- New `drag_stash` module — single-slot `Vec<String>` with 5 s TTL.
- `AgentMuxDragHandler::on_drag_enter` reads `DragData::file_paths` and stashes; returns 0 so the JS drop event still fires (no behavior change for existing draggable-regions logic).
- New `consume_drag_paths` IPC drains the stash.

**Frontend**
- New `frontend/util/dnd.ts` — `consumeDragPaths` + `copyFilesToDir` with concurrency-limited (4) per-file iteration; de-collision (`report (1).csv`) is server-side already.
- `term.tsx` — placeholder branch replaced; one summary toast instead of per-file spam.
- `agent-view` — new `useAgentDropAttach` hook owns the drop UX + the composer `@filename` splice. Uses the native HTMLTextAreaElement value setter so reactive bindings see the change; pads with a leading space when the caret isn't preceded by whitespace.
- Settings: documented `dnd:enabled`, `dnd:maxfilesizemb`, `dnd:concurrency`, `dnd:agentinserttoken` in `settings-template.jsonc`. Defaults applied in code: enabled=true, concurrency=4, agentinserttoken=true (size cap is documented but enforced in Phase 2 — see below).

## Closes the "placeholder" gap

Before: dropping a file on a terminal pane showed *"Dropped N files. Full path copy requires CefDragHandler integration."* That toast is gone — the CEF DragHandler now supplies real OS paths and the existing `copy_file_to_dir` IPC actually runs.

## Phase 2 follow-ups (explicit non-goals here)

Carved out of spec §3.4 and §3.8 for follow-up PRs:

- SSH / Docker / WSL transport backends in the sidecar (this PR's `copy_file_to_dir` is host-local-FS only — same as it was before)
- Server-side per-file size cap (`dnd:maxfilesizemb`)
- Large-file progress toast (> 8 MiB)
- Directory drop recursive copy semantics across transports

Local-FS host agents work end-to-end today; remote agents will fall back to the existing host-local copy (which lands the file in the host's filesystem, not inside the container/remote — Phase 2 fixes).

## Test plan

- [ ] Drop a single file onto a terminal pane → file appears in `pwd`; one toast "Copied X to Y"
- [ ] Drop three files at once → single toast "Copied 3 files to Y", no per-file spam
- [ ] Drop a file whose name already exists → de-collision lands as `name (1).ext`, toast and `@token` reflect the new name
- [ ] Drop a file onto an agent pane with empty composer → `@filename ` appears in textarea
- [ ] Drop with caret mid-word (`summarise|`) → ` @filename ` spliced with leading + trailing space
- [ ] Drop with caret already after a space → no double space
- [ ] Drop on agent pane with `dnd:agentinserttoken=false` → file copied, no splice, info toast says "Mention them in your next message"
- [ ] TTL: leave a drag entered for 6+ seconds, then drop into the source app instead → next drop into AgentMux gets fresh paths, not stale ones
- [ ] Drop on pane with no `cmd:cwd` → "No working directory detected" toast, no copy attempted
- [ ] `dnd:enabled=false` → drag still shows browser default, no overlay, no copy

Follow-up to #1191 / #1199.

### DIFF
--- a/.changesets/1780153219-feat-dnd-file-drop-on-terminal-agent-panes-phase-1-o3a8.md
+++ b/.changesets/1780153219-feat-dnd-file-drop-on-terminal-agent-panes-phase-1-o3a8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(dnd): file drop on terminal + agent panes (phase 1)
--- a/agentmux-cef/src/client/handlers.rs
+++ b/agentmux-cef/src/client/handlers.rs
@@ -117,6 +117,32 @@ wrap_drag_handler! {
     }
 
     impl DragHandler {
+        // File drop path-capture. The HTML5 drop event in CEF/Chromium hides
+        // full filesystem paths from JS (security model: pages can read `File`
+        // objects but not their host paths). CefDragData::get_file_paths
+        // exposes the real paths during OnDragEnter; we stash them so the
+        // JS-side drop handler can consume them via the consume_drag_paths
+        // IPC a few ms later. Returning 0 (don't cancel) lets the JS drop
+        // event fire normally — required for the existing DragOverlay UI to
+        // run unchanged. Spec: docs/specs/SPEC_PANE_FILE_DROP_2026_05_30.md §3.3.
+        fn on_drag_enter(
+            &self,
+            _browser: Option<&mut Browser>,
+            drag_data: Option<&mut DragData>,
+            _mask: DragOperationsMask,
+        ) -> ::std::os::raw::c_int {
+            if let Some(dd) = drag_data {
+                let mut list = CefStringList::new();
+                let _ = dd.file_paths(Some(&mut list));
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

PR #1202, opened 2026-05-30
fix(macos): show the AgentMux instance in the Dock (Regular activation policy)

a5af:
## Symptom

With the launcher driving `task dev` (and also in plain direct-invoke), the AgentMux instance had **no Dock tile** — it never appeared in the macOS taskbar. The window showed, but you couldn't see/switch to the instance from the Dock, and the #1179 Dock icon had nothing to paint on.

## Root cause

A bare `agentmux-cef` Mach-O launched **outside an `.app` bundle** (both `task dev` direct-invoke and the launcher's flat `dist/cef-dev/` layout) has no `Info.plist`, so LaunchServices leaves it as a **background/accessory** process: it can open windows but gets no Dock tile and no menu bar. `lsappinfo` confirmed it: `type="BackgroundOnly"`.

## Fix

Set `NSApplicationActivationPolicyRegular` (raw value `0`) via `-[NSApplication setActivationPolicy:]` at host startup, **before** painting the Dock icon (the icon needs a tile to land on). New `set_macos_activation_policy_regular()` mirrors the existing raw-libobjc FFI of `set_macos_dock_icon` — no new crate deps. Idempotent; harmless in a future packaged `.app` (already Regular there). macOS-only (`#[cfg(target_os = "macos")]`).

## Verified (under the launcher build: `task dev` → launcher → host)

- Host log: `activation-policy: set NSApplication to Regular (Dock-visible)` → `ok:true`.
- `lsappinfo`: host now `type="Foreground"` (was `BackgroundOnly`).
- `osascript … whose background only is false` now lists `agentmux-cef` → **Dock tile present**.
- The embedded AgentMux logo (#1179) paints on the tile.
- Launcher tree healthy: launcher → host + **one** adopted srv.

Single-file change (`agentmux-cef/src/main.rs`).

### DIFF
--- a/.changesets/1780154383-fix-macos-show-the-agentmux-instance-in-the-dock-set-regular-ulo2.md
+++ b/.changesets/1780154383-fix-macos-show-the-agentmux-instance-in-the-dock-set-regular-ulo2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): show the AgentMux instance in the Dock (set Regular activation policy)
--- a/agentmux-cef/src/main.rs
+++ b/agentmux-cef/src/main.rs
@@ -677,11 +677,19 @@ fn main() {
 
     tracing::info!("CEF initialized, entering message loop");
 
-    // Set the Dock icon now that NSApplication exists. macOS-only; no-op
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1203, opened 2026-05-30
test(bench): terminal key-repeat hiccup diagnostic + typing-perf measurement findings

AgentY-asaf:
## What

Adds `tools/tests/term-keyrepeat-hiccups.mjs` — a CDP-driven diagnostic for the **"not silky" stutter felt while HOLDING a key** in a terminal pane (sustained key-repeat). This is a distinct failure mode from the two existing benches:

| Bench | Measures |
|---|---|
| `bench-term-echo.mjs` | single keystroke → PTY → echo round-trip latency |
| `bench-agent-keystroke.mjs` | agent composer textarea handler cost |
| **`term-keyrepeat-hiccups.mjs` (new)** | **frame pacing during sustained key-repeat** |

Tracks the input-first umbrella ([#1161](https://github.com/agentmuxai/agentmux/discussions/1161)).

## How it works

Attaches an in-page monitor to **every** shell window over CDP and, during the hold, records per window:
- **rAF frame intervals** → dropped/janky frames (>20 / >33 / >50 / >100 ms)
- **`longtask`** entries → main-thread JS blocking ≥ 50 ms
- **perf measures** → `term-keypress` / `term-echo-render` / `term-raf-write`
- **focus + active-element sampling** and an **xterm-rows `MutationObserver`** — these *prove* the terminal was focused and actually echoing in the measured window. A run is marked **VALID** only if the keystrokes truly landed (and rendered) there, so a mis-focused capture can't masquerade as data.

It auto-selects the active terminal window and classifies the hiccup as **main-thread JS** (schedulable/fixable), **xterm write cost**, or **compositor/GPU stall** (render-path).

```
# focus a terminal pane, run this, then HOLD a key for the whole countdown
node tools/tests/term-keyrepeat-hiccups.mjs [--cdp-port 9223] [--secs 18] [--json out.json]
```

## Findings so far (local, v0.40.1, validated 18s hold)

```
~56 fps   ·   frame p50=16.7ms  p95=16.8ms  p99=50ms  max=50ms
jank: 42 frames >33ms · 5 >50ms · 0 >100ms
longtasks (JS ≥50ms): 0   ·   0ms blocked
xterm marks: term-keypress p95=0.4ms · term-echo-render p95=0.4ms · term-raf-write p95=0.2ms
```

→ ~2.3 dropped frames/sec with **zero JS blocking** and **sub-ms** input handling. The stutter is in the **render/compositor path, not JS** — so scheduler/flow-control work would not fix it.

### Leading hypothesis (for the reviewing agent to scrutinize)

We run a **double rAF coalescer** on the PTY→screen path: our own `requestAnimationFrame` batch in `termwrap.ts` (`scheduleRafWrite`/`armRaf`) feeds `terminal.write()`, which then schedules **xterm's own `RenderDebouncer` rAF**. Two unsynchronized frame gates in series can add up to ~32ms and beat against each other → the uneven pacing observed. VS Code (same xterm.js, same Chromium) relies solely on xterm's single built-in debouncer.

Our Stage-1 rAF exists for a real reason (the "Tier-3 scroll fix": coalescing Ink's cursor-up + content into one write to stop a viewport double-flash). Open question: does xterm's debouncer alone also prevent that flash, letting Stage 1 be removed?

## Caveats (honest)
- `longtask` only fires for ≥50ms blocks, so a 17–49ms main-thread task could drop a frame uncounted. A CDP `Tracing` capture is the definitive next step for frame-by-frame attribution.
- WebSearch was unavailable during the VS Code research; the comparison is from reading VS Code's `xtermTerminal.ts` + xterm source directly, not a profiler diff.

## Scope
Test tooling only (no app code changed). Changeset: `patch`.

### DIFF
--- a/.changesets/1780154398-test-term-keyrepeat-hiccups.md
+++ b/.changesets/1780154398-test-term-keyrepeat-hiccups.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+test(bench): term key-repeat hiccup diagnostic (CDP frame/longtask/mark capture) — input-first (#1161)
--- a/tools/tests/README.md
+++ b/tools/tests/README.md
@@ -84,6 +84,41 @@ exercise the OS keyboard pipeline.
 
 Spec: [`docs/specs/SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md`](../../docs/specs/SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md) §7.2.
 
+## `term-keyrepeat-hiccups.mjs`
+
+CDP-driven diagnostic for the **"not silky" stutter felt while HOLDING a key**
+in a terminal pane (sustained key-repeat) — a distinct failure mode from
+single-keystroke echo latency (`bench-term-echo.mjs`) and the agent composer
+(`bench-agent-keystroke.mjs`).
+
+Attaches an in-page monitor to **every** shell window over CDP (9223 dev /
+9222 release) and, while you hold a key, records per window:
+
+- **rAF frame intervals** → dropped/janky frames (>20 / >33 / >50 / >100 ms)
+- **`longtask`** entries → main-thread JS blocking ≥ 50 ms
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #276, opened 2026-04-02
fix(term): eliminate character echo delay — bypass RAF for small PTY writes

AgentX-asaf:
## Problem

Typing felt sluggish — **some characters appeared with noticeable delay** instead of instantly.

### Root cause

\`scheduleRafWrite\` (the Tier-3 scroll-flicker fix) added up to **16ms latency** to *all* PTY output, including single-byte character echoes. Worse, when a slow write was in flight (\`writeInFlight = true\`), new data didn't even get a RAF scheduled — it waited for the full current write to finish, then waited another RAF cycle.

**Total potential delay per typed character:** write duration (can be 8–50ms on large scrollback) + RAF cycle (≤16ms) = **24ms+ delay** for characters typed during PTY output.

This is why *some* characters felt delayed — specifically those typed while the terminal was processing output.

A second issue: \`console.log\` fired on every RAF write in the non-slow path, adding IPC overhead hundreds of times/second during streaming output.

## Fix

**Small writes (≤512 bytes) bypass RAF entirely and write directly to xterm.js.**

The RAF scroll-flicker fix only applies to large multi-chunk outputs. A single echoed character (1–4 bytes) is never split, so there is no flicker risk. This gives **zero-latency character echo** regardless of PTY output state.

xterm.js serialises \`terminal.write()\` calls internally, so writing directly while \`writeInFlight\` is true is safe.

Also removes the \`console.log\` from the non-slow path — slow writes (>8ms) still warn.

## Changes

- \`frontend/app/view/term/termwrap.ts\`
  - Add \`RAF_BYPASS_THRESHOLD = 512\` constant
  - \`scheduleRafWrite\`: fast-path small data directly via \`doTerminalWrite\`
  - \`armRaf\`: remove non-slow-path \`console.log\`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1206, opened 2026-05-30
feat(term): term:disablerafcoalesce — opt-in double-rAF tear-out for typing smoothness

AgentY-asaf:
## What

Adds a default-off setting **`term:disablerafcoalesce`** that **tears out our Stage-1 rAF write-coalescer** in the terminal, so PTY output writes straight to xterm.js and xterm's own `RenderDebouncer` is the only frame gate.

Tracks the input-first umbrella ([#1161](https://github.com/agentmuxai/agentmux/discussions/1161)). Companion to the measurement tooling in #1203.

## Why — the "double rAF"

The PTY→screen path runs **two `requestAnimationFrame` coalescers in series**:

```
PTY msg -> [our Stage-1 rAF: termwrap.scheduleRafWrite/armRaf] -> terminal.write()
        -> [xterm's own rAF: RenderDebouncer] -> paint
```

xterm.js already renders **at most once per animation frame** (`RenderDebouncer` guards on `_animationFrame`, merges dirty rows). Our Stage-1 rAF is a **second, unsynchronized frame gate** — can add up to ~32 ms and **beat against** xterm's rAF → uneven frame pacing.

Measured with the #1203 diagnostic (VALID 18s hold-key, v0.40.1): **~56 fps, frame p50 16.7 ms with a 33–50 ms tail, 0 JS long-tasks, sub-ms xterm marks** → the stutter is render-path, not JS. VS Code drives the same xterm.js with **only** the built-in debouncer.

## Provenance — why Stage-1 exists (don't merge-flip blindly)

4-commit arc:

| Commit | PR | What |
|---|---|---|
| `ab9c38d6` | #208 | **Introduced it** — Ink TUIs emit cursor-up + content as separate WS messages -> viewport snapped up/down = **double flash**. **Windows 10** DWM shows each snap as a frame; **Windows 11 coalesces within vsync (invisible).** |
| `7f442735` | #235 | `writeInFlight` guard (slow write let a 2nd concurrent rAF reintroduce the flash) |
| `5bcf503f` | #276 | <=512 B echo fast-path (rAF was taxing echo) |
| `e58707766` | #926 | removed `writeInFlight` from fast path (still stalled echo) |

#235->#276->#926 is a progressive retreat carving the echo path back *out* of the coalescer. This is the end of that arc — but the **original flash is Windows-10-specific**, so removal must be Win10-verified.

## Design — why a default-off flag

A/B on the **same build** (settings hot-reload, no rebuild): flip `term:disablerafcoalesce` and compare. Default off = zero risk to other users. Wired `term.tsx` -> `TermWrap.coalesceWrites` -> early-return in `scheduleRafWrite()`; plumbed through `schema/settings.json`, `gotypes.d.ts`, and `wconfig/types.rs` so the key round-trips.

## Verification protocol (full detail in the spec)

1. **Win10 scroll-flash regression (BLOCKING)** — run an Ink TUI on real Windows 10, flag off vs on; PASS = no viewport flash in tear-out mode.
2. **Hiccups** — `term-keyrepeat-hiccups.mjs` (#1203) hold-key, flag off vs on; expect the 33–50 ms frame tail to shrink.
3. **Throughput/ordering** — `bench-term-echo.mjs --stream --busy`, no new ordering violations or echo regression.

Flip the default only if Test 1 passes on Win10 and 2–3 improve. `docs/specs/SPEC_TERM_DOUBLE_RAF_TEAROUT_2026_05_30.md`.

## Build / verify
- `npx tsc --noEmit`: **no new errors** (the 27 reported all pre-exist on main — test fixtures + `React` namespace + layout test types; none touch term write path or the new setting)
- `cargo check -p agentmux-srv`: **clean (0 errors)**

## Scope
Frontend term write path + setting plumbing (schema/gotypes/wconfig) + spec. Default-off, behavior unchanged unless the setting is set. Changeset: `patch`.

### DIFF
--- a/.changesets/1780156312-feat-term-disablerafcoalesce.md
+++ b/.changesets/1780156312-feat-term-disablerafcoalesce.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(term): term:disablerafcoalesce — opt-in double-rAF tear-out for typing smoothness (input-first #1161)
--- a/agentmux-srv/src/backend/wconfig/types.rs
+++ b/agentmux-srv/src/backend/wconfig/types.rs
@@ -67,6 +67,9 @@ pub struct SettingsType {
     #[serde(rename = "term:disablewebgl", default, skip_serializing_if = "is_false")]
     pub term_disable_web_gl: bool,
 
+    #[serde(rename = "term:disablerafcoalesce", default, skip_serializing_if = "is_false")]
+    pub term_disable_raf_coalesce: bool,
+
     #[serde(rename = "term:localshellpath", default, skip_serializing_if = "String::is_empty")]
     pub term_local_shell_path: String,
 
--- a/docs/specs/SPEC_TERM_DOUBLE_RAF_TEAROUT_2026_05_30.md
+++ b/docs/specs/SPEC_TERM_DOUBLE_RAF_TEAROUT_2026_05_30.md
@@ -1,0 +1,90 @@
+# SPEC: Terminal double-rAF tear-out (experiment)
+
+**Status:** Experimental — behind a default-off setting
+**Date:** 2026-05-30
+**Author:** AgentY
+**Setting:** `term:disablerafcoalesce` (boolean, default `false`)
+**Tracks:** [`SPEC_INPUT_RESPONSIVENESS_*`](./SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md) · [discussion #1161](https://github.com/agentmuxai/agentmux/discussions/1161) · bench PR #1203
+
+---
+
+## TL;DR
+
+The terminal PTY→screen path runs **two `requestAnimationFrame` coalescers in series**:
+
+```
+PTY msg → [our Stage-1 rAF: termwrap.scheduleRafWrite/armRaf] → terminal.write()
+        → [xterm's own rAF: RenderDebouncer] → paint
+```
+
+xterm.js already renders at most **once per animation frame** (its `RenderDebouncer`
+guards on `_animationFrame` and merges dirty rows). Our Stage-1 rAF is therefore a
+**second, unsynchronized frame gate**. Two gates in series can add up to ~32 ms and
+**beat against each other** — the likely cause of the uneven frame pacing ("not
+silky" hiccups) measured while holding a key: ~56 fps, frame p50 16.7 ms with a
+33–50 ms tail, **0 JS long-tasks**, sub-ms xterm marks (PR #1203 diagnostic).
+
+VS Code (same xterm.js, same Chromium) uses **only** xterm's built-in debouncer.
+
+This change adds `term:disablerafcoalesce`. When set, `scheduleRafWrite()` writes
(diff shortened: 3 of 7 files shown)


a5af:
@codex review

AgentY-asaf:
## Findings + Windows-10 handoff (AgentY, 2026-05-30)

### Confirmed: this fixes the "not silky" hiccup (on Windows 11)

Tested live with `task dev` on this branch (Win11). With `term:disablerafcoalesce: true` and a terminal reopened, sustained key-repeat is **subjectively much smoother** — the user confirmed "much better."

**Objective baseline (flag OFF, from the #1203 diagnostic, VALID 18s hold, v0.40.1):**
```
~56 fps · frame p50 16.7ms · p99 50ms · max 50ms
jank: 42 frames >33ms · 5 >50ms · 0 >100ms
longtasks (JS ≥50ms): 0 · xterm marks sub-ms (term-keypress p95 0.4ms)
```
0 JS blocking + sub-ms handler cost ⇒ the dropped frames were the **double-rAF beat**, not JS. Removing our Stage-1 rAF (this PR) is what the user felt improve. (A flag-ON capture to quantify the delta wasn't recorded this session — worth grabbing, but the subjective + mechanism are consistent.)

### Why this is safe to merge as-is
**Default-off / opt-in.** Nobody gets the tear-out unless they explicitly set `term:disablerafcoalesce: true`, so it cannot regress anyone — including Win10. It ships as a measurable experiment.

### ⚠️ Windows-10 verification — REQUIRED before flipping the default (@win10 agent)
The Stage-1 rAF this PR removes exists **specifically** to fix a **Windows-10 DWM scroll-flash** with Ink TUIs (PR #208; invisible on Win11, so the Win11 result above does NOT clear it). Before this can become the default (or be hard-removed), run on **real Windows 10**:

1. **Flash regression (BLOCKING).** Open a terminal, run an Ink TUI that redraws heavily (a Claude Code / Codex session is ideal). Baseline with flag off, then set `"term:disablerafcoalesce": true`, reopen the terminal, repeat the same workload. **PASS = no viewport up/down flash** in tear-out mode. **FAIL = flash returns** → xterm's own debouncer alone is insufficient; do not flip the default.
2. **Hiccup delta.** `node tools/tests/term-keyrepeat-hiccups.mjs --secs 18` (from #1203) while holding a key, flag off vs on. Expect the >33ms jank count / p99 tail to shrink.

Full protocol + provenance: `docs/specs/SPEC_TERM_DOUBLE_RAF_TEAROUT_2026_05_30.md`.

### If the Win10 flash regresses (likely real next step)
The clean fix is **backend-side coalescing**: the root cause per PR #208 is *"PTY data arrives as separate WebSocket messages."* Merge consecutive PTY reads within a ~2-4ms window into one WS message in the sidecar (the read loop is at `agentmux-srv/src/backend/blockcontroller/shell.rs`, `tokio::spawn_blocking`, 4 KiB buf, `handle_append_block_file`). That kills the flash at the source AND removes the frontend frame gate — smooth on both platforms, no per-frame beat. This PR's front-end half (write-straight-to-xterm) then composes with it. I can spec/build that once the Win10 test confirms the flash still reproduces.

PR #1208, opened 2026-05-30
fix(cef): never request OS credential/keychain access in any runtime mode

a5af:
## Symptom

A packaged AgentMux build (installed/portable) pops the macOS dialog **"AgentMux wants to use your confidential information stored in 'Login' in your keychain"** on first launch. On macOS this modal **blocks Chromium startup**, so the app hangs with no window (the GPU subprocess times out with "no connection"). Equivalent prompts exist on Linux (gnome-keyring/kwallet) and Windows (CredVault).

## Root cause

`agentmux-cef/src/app.rs` gated the switches that keep Chromium's `OSCrypt` away from the OS credential store — `--password-store=basic` and (macOS) `--use-mock-keychain` — to `RuntimeMode::Dev`. Installed/portable builds skipped them, so `OSCrypt` derives its cookie encryption key from the **platform-native credential store** and prompts the first time it's touched — even though AgentMux never saves a password.

The gate was a deliberate prior decision ("the prompt is appropriate for released builds") but was **never exercised on macOS** because `task package:macos` was a TODO stub until now — so the conflict with the product requirement (*no OS credential prompts*) was latent.

## Fix

Ungate the switches so they apply in **all** runtime modes (dev / installed / portable):
- `--password-store=basic` — appended on every platform.
- `--use-mock-keychain` — appended in all modes, **macOS only** (`#[cfg]` kept).

The diff is a **net deletion** of the `is_dev_runtime` condition around two platform-agnostic `append_switch` calls — no type/state/`cfg` changes, so **no Windows compile risk** (the #1192 failure mode is avoided).

### Tradeoff (documented in the spec + code comment)

Chromium's cookie jar is then encrypted at rest with an obfuscation key rather than an OS-credential-derived one. Acceptable for AgentMux's **local single-user** threat model (data dir already `0700`; the cookie jar holds the user's own session cookies, not shared secrets). This reverses the earlier "prompt is appropriate for released builds" decision.

## Verification

Re-packaged the installed-mode `.app` and launched it:
- **Keychain prompt: gone.** Before the fix the app hung GPU-only at the keychain modal; after, it progresses *past* OSCrypt init and spawns the full subprocess set + attempts the renderer. The credential-prompt hang is resolved.
- A **separate** signed-bundle subprocess crash-loop surfaced (renderer/GPU under hardened runtime) — that's the independent hard part of macOS CEF packaging (Helper apps / hardened-runtime), **not** a credentials issue; tracked with the `task package:macos` work.

## Docs
- Spec: `docs/specs/SPEC_SUPPRESS_OS_CREDENTIAL_PROMPTS_2026_05_30.md`
- Retro (incident + mis-diagnosis lesson): `docs/retro/retro-macos-keychain-prompt-2026-05-30.md`

### DIFF
--- a/.changesets/1780158422-fix-cef-never-request-os-credential-keychain-access-in-any-r-cyfh.md
+++ b/.changesets/1780158422-fix-cef-never-request-os-credential-keychain-access-in-any-r-cyfh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): never request OS credential/keychain access in any runtime mode
--- a/agentmux-cef/src/app.rs
+++ b/agentmux-cef/src/app.rs
@@ -423,52 +423,36 @@ wrap_app! {
                 // Upstream: https://github.com/chromiumembedded/cef/issues/3603
                 cmd.append_switch(Some(&CefString::from("disable-chrome-login-prompt")));
 
-                // Suppress OS credential prompts on first-run dev / portable
-                // launches. **Dev-only** — gated to `RuntimeMode::Dev` because
-                // these switches weaken Chromium's at-rest encryption of
-                // cookies / passwords by routing OSCrypt to an unencrypted
-                // basic store (and, on macOS, mocking the Keychain entirely).
-                // Enabling them in a packaged AppImage / .app / installer
-                // build would silently degrade credential security on user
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1209, opened 2026-05-30
feat(packaging): Microsoft Store MSIX packaging for the CEF app

AgentX-asaf:
## What

Adds **Microsoft Store MSIX packaging** for the current CEF app via `task package:msix`. Replaces the stale **Tauri-era** tooling (`agenty/msix-packaging` branch + `agentmux-builder` repo, both pre-CEF — they stage `src-tauri/`/`WebView2Loader.dll`/`wsh.exe` that no longer exist). Full design in `docs/specs/SPEC_MSIX_PACKAGING_2026_05_30.md`.

## How

- **`scripts/package-msix.ps1`** — stages the `task package` portable, **packs** with `makeappx`:
  - Excludes `agentmux-portable.marker` + `data/` so the packaged app runs in **installed mode** (per-user `~/.agentmux` data dir), not portable mode — which would fail against the read-only `WindowsApps` install. *(Verified against `agentmux-common::runtime_mode`: no marker + non-dev path ⇒ Installed — there's even a unit test for it. Spec §4.)*
  - **Identity guard:** computes the MSIX publisher-hash of the manifest `Publisher` and asserts it equals the **live published PFN** suffix `vqr1k32tkfk4y` (Store ID `9P9QCXNNCRK3`) before packing — a wrong Publisher fails the build instead of getting rejected by the Store.
  - `-Sign` self-signs for local `Add-AppxPackage` testing (Store re-signs on ingest, so unsigned is correct for submission).
- **`packaging/msix/AppxManifest.xml.template`** — FullTrust desktop app, entry `agentmux.exe`, `runFullTrust` + internet capabilities. Live Store identity baked in (`Name=AgentMux.AgentMux`, `Publisher=CN=C2BCB530-…`), recovered and hash-verified from PR #240. A Publisher CN is an *identity*, not a secret.
- **`packaging/msix/generate-assets.ps1` + `assets/`** — Store logos generated from the 512px app icon (committed for reproducible builds; no image toolchain needed at build time).
- **Taskfile** — `task package:msix` wired (was a TODO `echo`).

## Verification

- `task package:msix` → `makeappx` **packs 412 files** (172 MB), "Package creation succeeded"; identity guard passes.
- `task package:msix -- -Sign` → signs cleanly.
- ⚠️ **Local install/launch (spec §9) not yet run here** — `Add-AppxPackage` of the self-signed package fails with `0x800B0109` (root cert not trusted) and Developer Mode is off; trusting the test cert needs **admin**. To validate locally:
  ```powershell
  Export-Certificate -Cert Cert:\CurrentUser\My\<thumbprint> -FilePath agentmux-test.cer
  Import-Certificate -FilePath agentmux-test.cer -CertStoreLocation Cert:\LocalMachine\TrustedPeople   # admin
  Add-AppxPackage dist\msix\AgentMux_0.40.1_x64.msix
  ```
  Then confirm the spec §9 checklist (launches via launcher; host+srv spawn; data dir = `~/.agentmux`; terminals/agents spawn; single-instance). This exercises the only real risk — CEF + launcher behavior under MSIX — and may need a follow-up.

## Not in scope
- Signing (Store re-signs), CI wiring, arm64. x64 only.


a5af:
@codex review

AgentX-asaf:
Addressed the [P2] in `f107a8dd`.

Added a version-consistency guard in `scripts/package-msix.ps1` immediately after the portable is resolved: it reads the bundled host binary `runtime/agentmux-<ver>.exe` (its filename is the binaries' own version of record), parses the version, and asserts it equals `package.json`. On skew the build throws with a clear message instead of silently emitting an MSIX whose manifest version diverges from its binaries.

```powershell
$hostExe = Get-ChildItem (Join-Path $PortableDir 'runtime') -Filter 'agentmux-*.exe' |
  Where-Object { $_.Name -match '^agentmux-(\d+\.\d+\.\d+)\.exe$' } | Select-Object -First 1
$portableVersion = [regex]::Match($hostExe.Name, '^agentmux-(\d+\.\d+\.\d+)\.exe$').Groups[1].Value
if ($portableVersion -ne $semver) { throw "Version skew: package.json is $semver but the portable bundles agentmux-$portableVersion.exe ..." }
```

Verified against the live `0.40.1` portable: matches and the build proceeds; the regex correctly excludes `agentmux-srv-0.40.1-windows.x64.exe`. Re-review requested.

AgentX-asaf:
Pushed `dc2d9fbf` — fixes a Store ingest rejection found on upload.

**Partner Center error:** `PublisherDisplayName` was `AgentMux Corp`, which doesn't match the account publisher display name `AgentMux`. The manifest had the *legal-entity* name (correct for copyright headers) in the Store publisher field, which must match Partner Center verbatim.

**Third occurrence** — fixed in `bb391461` and #240 (March), regressed when the corrected `src-tauri` manifest was deleted (`12333fa2`) and the CEF manifest was recreated in `cc5f7368`. Retro: `docs/retro/retro-msix-publisherdisplayname-regression-2026-05-30.md`.

**Changes:**
- `AppxManifest.xml.template` → `PublisherDisplayName=AgentMux` (+ explanatory comment)
- `package-msix.ps1` → build-time guard asserts `PublisherDisplayName == 'AgentMux'` before `makeappx` (alongside the existing Publisher-hash guard) so it can't regress a fourth time
- spec §10 corrected (catalog `DeveloperName` ≠ publisher display name)

Verified: the packed `.msix` manifest reads `PublisherDisplayName=AgentMux`; all three guards (binver / identity / pubName) pass.

PR #1212, opened 2026-05-30
feat(agent-pane): new-message enter animation

AgentA-asaf:
## Summary

- New nodes streaming into the agent pane now animate in with a 120ms fade + 4px `translateY`, matching the existing tool-panel collapse easing (`cubic-bezier(0.4, 0, 0.2, 1)`)
- Scoped to `.agent-document-streaming-buffer` only — virtualizer re-inserts (scroll recycling, `loadOlder`) are **not** affected
- `prefers-reduced-motion: reduce` guard included — users with the OS setting see instant appearance as before
- CSS-only, no JS changes

## Test plan

- [ ] Start `task dev`, open an agent pane, send a message — new tool blocks / markdown / user messages should fade+slide in over ~120ms
- [ ] Scroll up and back down — history rows replayed by the virtualizer should NOT animate
- [ ] Click "Load older" — prepended rows should NOT animate
- [ ] Confirm tool-panel expand/collapse animation is unaffected
- [ ] Enable "Reduce motion" in OS accessibility settings — rows should snap in with no animation
- [ ] No layout shift or scroll-position jump during the enter animation

### DIFF
--- a/.changesets/1780175946-feat-agent-pane-new-message-enter-animation-swift-fade-slide-4us9.md
+++ b/.changesets/1780175946-feat-agent-pane-new-message-enter-animation-swift-fade-slide-4us9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): new-message enter animation — swift fade+slide on streaming rows
--- a/docs/specs/SPEC_AGENT_PANE_MESSAGE_ENTER_ANIMATION_2026_05_30.md
+++ b/docs/specs/SPEC_AGENT_PANE_MESSAGE_ENTER_ANIMATION_2026_05_30.md
@@ -1,0 +1,171 @@
+# SPEC: Agent Pane — New Message Enter Animation
+
+**Date:** 2026-05-30
+**Status:** Ready to implement
+**Area:** `frontend/app/view/agent/styles/_document.scss` + `_document-nodes.scss`
+**Goal:** New nodes that stream into the agent pane should enter with the same swift fade-in as the tool-panel collapse, instead of snapping in instantly.
+
+---
+
+## 1. The animation we're matching
+
+The log panel collapse (`agent-tool-panel`) already ships a polished transition:
+
+```scss
+// frontend/app/view/agent/styles/_document-nodes.scss:275-279
+.agent-tool-panel {
+    transition:
+        max-height 120ms cubic-bezier(0.4, 0, 0.2, 1),
+        padding    120ms cubic-bezier(0.4, 0, 0.2, 1),
+        margin     120ms cubic-bezier(0.4, 0, 0.2, 1),
+        opacity    100ms ease-out;
+}
+```
+
+Duration: **120 ms**. Easing: **Material's standard cubic-bezier(0.4, 0, 0.2, 1)**. Properties: structural geometry + opacity. This is the target feel.
+
+---
+
+## 2. Problem
+
+New agent-pane rows — tool blocks, markdown chunks, user messages, bash outputs — appear
+instantly. There is no enter animation on `.agent-document-row` or the streaming buffer rows.
+The snap is especially visible during a fast-streaming response where multiple nodes land in
+quick succession; each pops into existence abruptly.
+
+---
+
+## 3. Architecture of the streaming buffer
+
+New nodes during an active turn land in `.agent-document-streaming-buffer`, a plain flex
+column in normal document flow (no virtualization). They are **not** virtualized while
+streaming, so CSS `@keyframes` / transition on mount is safe — the elements ARE in the DOM at
+paint time.
+
+Once the turn ends and the virtualizer absorbs them, rows are re-inserted as absolutely-
+positioned tiles inside `.agent-document-virtualizer`. Replayed history rows also enter via
+the virtualizer. Those paths are covered by §5 (reduced-motion guard) and should NOT
+animate on virtualized re-insert (that would fire every time the user scrolls a row into
+the virtual window).
+
+---
+
+## 4. Implementation
+
+### 4.1 Keyframe definition
+
+Add to `_document.scss` inside `.agent-view`:
+
+```scss
+@keyframes agent-node-enter {
+    from {
+        opacity: 0;
+        transform: translateY(4px);
+    }
+    to {
+        opacity: 1;
+        transform: translateY(0);
+    }
+}
+```
+
+`4px` translateY keeps it subtle — enough to convey direction (downward stream) without
+feeling like a full slide-in. Matches the "swift" character of the tool-panel collapse.
+
+### 4.2 Apply to streaming-buffer rows only
+
+```scss
+// _document.scss — inside .agent-view
+.agent-document-streaming-buffer .agent-document-row {
+    animation: agent-node-enter 120ms cubic-bezier(0.4, 0, 0.2, 1) both;
+}
+```
+
+Scoped to `.agent-document-streaming-buffer` — not `.agent-document-virtualizer` — so
+the animation fires ONLY when a new node mounts into the live streaming region. Virtualized
+rows that scroll in/out of the DOM window are NOT affected.
+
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

AgentY-asaf:
## Confirmed the P1 — and here's a fix (AgentY)

I traced reagent's P1 against the code and it's correct. `partitionForVirtualization` (`streaming-buffer.ts:73`) returns **all** nodes in `streamingNodes` whenever `nodes.length <= STREAMING_BUFFER_SIZE` (50). Most agent transcripts are under 50 nodes, so the *entire* document renders in `.agent-document-streaming-buffer` — and on pane open/restore (`HistoryLoaded`/`HistoryRestored`) every row mounts there and fires the keyframe. Result: opening a pane replays a fade+slide cascade over the whole visible transcript. Even for >50-node docs, the trailing ~50 history rows animate on restore.

Root cause: **pure CSS can't distinguish a genuinely-new streamed node from an initial/restored mount** — both are real DOM mounts. The fix needs a tiny JS gate so the animation only applies to rows that mount *after* the buffer's initial population.

### Proposed fix (verified: tsc clean on the two .tsx files, no new errors)

**1. `AgentDocumentVirtualList.tsx`** — add a gate that opens one frame after the component mounts (children mount before the parent's `onMount`, so initial rows sample it as `false`):

```ts
// near the top of the component, with the other `let`s:
let streamingReady = false;

// in the existing onMount:
onMount(() => {
    startAgentLayoutShiftObserver();
    requestAnimationFrame(() => { streamingReady = true; });
});

// pass it to the STREAMING-buffer DocumentRow only (NOT the virtualized one):
<DocumentRow ... animateOnEnter={() => streamingReady} />
```

**2. `DocumentRow.tsx`** — sample the gate once on mount, toggle a class:

```ts
import { createSignal, onMount, Show, ... } from "solid-js";

// add to DocumentRowProps:
animateOnEnter?: () => boolean;

// in the component body:
const [entering, setEntering] = createSignal(false);
onMount(() => {
    close();                                   // existing markRowMount closer
    if (props.animateOnEnter?.()) setEntering(true);
});

// on the wrapper div:
classList={{ ...existing, "agent-node-entering": entering() }}
onAnimationEnd={(e) => { if (e.animationName === "agent-node-enter") setEntering(false); }}
```

**3. `_document.scss`** — gate the selector on the class instead of the buffer alone:

```scss
.agent-document-streaming-buffer .agent-document-row.agent-node-entering {
    animation: agent-node-enter 120ms cubic-bezier(0.4, 0, 0.2, 1) both;
}
@media (prefers-reduced-motion: reduce) {
    .agent-document-streaming-buffer .agent-document-row.agent-node-entering {
        animation: none;
    }
}
```

### Why this holds
- **Initial / restored history** (incl. the ≤50-node whole-buffer case): rows mount before the rAF opens the gate → sample `false` → no class → no animation. Fixes the cascade.
- **Genuinely-new streamed node**: mounts after the gate is open → `true` → animates. (`<Index>` keys the buffer by position, so an appended node creates a fresh slot whose `onMount` fires.)
- **Virtualized region / scroll-recycle / loadOlder**: never receives `animateOnEnter` → never animates.
- `onAnimationEnd` drops the class so it doesn't linger; reduced-motion still honored.

Happy to push this to your branch if you'd prefer, or you can apply it directly. Spec §4.2/§5 should be updated to describe the gate (the current text claims history is "always in the virtualizer," which isn't true for ≤50-node docs).

AgentA-asaf:
Fixed in 2b08d106.

**reagent P1 (history rows animating on open/restore):** The root cause was the CSS selector firing on every DOM mount in the streaming buffer — including the initial HistoryLoaded/HistoryRestored batch.

Fix: gate the animation on `[data-animate]` attribute.

- Added `animateEnabled` signal (initially `false`) in `AgentDocumentVirtualList`.
- A `createEffect` watches `nodes().length` — when it first becomes non-zero (history settled), a `queueMicrotask` flips it to `true`. The microtask defers AFTER Solid flushes the DOM mutation batch, so the history rows are already mounted before `[data-animate]` appears and cannot re-trigger their CSS animation.
- Streaming buffer container gets `data-animate={animateEnabled() || undefined}` (absent = no attribute when false).
- CSS updated to `.agent-document-streaming-buffer[data-animate] .agent-document-row` — history rows mount without it, new streaming rows mount with it.

a5af:
@codex review

AgentA-asaf:
Fixed in 206f864a.

codex P2 is correct — adding `[data-animate]` to a parent after rows are already in the DOM still triggers a CSS `@keyframes` animation (the rule starts matching → animation fires). The JS gate controls *when* the attribute appears but can't suppress an animation that fires on selector-match change.

**Fix:** switched from `@keyframes animation` to `transition` + `@starting-style`.

`@starting-style` fires ONLY at actual mount time (element first rendered, or `display:none`→visible) — NOT when a CSS rule becomes newly applicable to an already-mounted element:

- History rows mount while `[data-animate]` absent → rule never matched → `@starting-style` never fires → instant appearance ✓
- `queueMicrotask` adds `[data-animate]` → rule now matches mounted rows, but `@starting-style` does NOT retroactively fire → no animation ✓  
- New streaming rows mount while `[data-animate]` present → `@starting-style` fires → 120ms fade+slide ✓

`@starting-style` is available on Chromium 117+ / cef-146. No JS changes needed beyond the existing `animateEnabled`/`[data-animate]` logic.

a5af:
@codex review

AgentA-asaf:
Fixed in 10c1bcfb — both findings addressed:

**reagent P1 (@starting-style fires on first style resolution, not DOM insert):** The `queueMicrotask` approach was wrong: microtasks drain before the browser's style recalc, so all pending microtasks run BEFORE the first paint. Rows inserted into the streaming buffer and having `[data-animate]` added by the microtask both have their first style resolution WITH the attribute — `@starting-style` fires for them and they cascade. The fix: force a synchronous style/layout flush (`void scrollRef.scrollTop`) BEFORE calling `setAnimateEnabled(true)`. This commits the history rows' first resolution WITHOUT `[data-animate]`; the subsequent attribute addition is a style UPDATE, not a first resolution — `@starting-style` doesn't fire.

**codex P2 (empty conversation first row snaps in):** Added `historyReady`/`markHistoryReady` to `AgentViewState`. `useHistoryPagination` calls `opts.onHistoryReady?.()` at every terminal init point (HistoryRestored, HistoryLoaded, total=0 fast-exit, InitFailed). Bridged from `agent-view.tsx` → `AgentDocumentView` (new `registerHistoryReadyCallback` prop) → `viewState.markHistoryReady`. For empty panes, `historyReady()` fires immediately — no rows in the buffer yet, forced reflow is no-op, `[data-animate]` is present before the first streaming row arrives → first row animates correctly.

**reagent P2 (stale spec):** Spec §3/§4/§5 updated to document the actual shipped implementation (`@starting-style` + `transition` + `[data-animate]` gate + `historyReady` signal + forced reflow).

a5af:
@codex review

AgentA-asaf:
Fixed in 733ec38e — three stale comment/spec items:

1. **AgentDocumentVirtualList.tsx:102** — described the superseded queueMicrotask mechanism. Updated to explain the actual approach: `historyReady()` signal + forced reflow (`void scrollRef.scrollTop`) before `setAnimateEnabled(true)`.

2. **_document.scss:212** — claimed "`queueMicrotask` adds `[data-animate]` → `@starting-style` does NOT retroactively fire". This was inaccurate: `@starting-style` fires at PAINT TIME (after all microtasks), not at DOM insert — without the forced reflow, history rows WOULD animate. Updated to state clearly that the gate only works because of the reflow in `AgentDocumentVirtualList.tsx`, and that this CSS is intentionally silent on the mechanism.

3. **Spec §7 "Files to change"** — said "one file, ~15 lines of CSS". Updated to list all 6 files changed by the shipped implementation.

a5af:
@codex review

PR #1215, opened 2026-05-31
fix(agent-pane): cap tool-output rendering to bound conversation DOM

AgentA-asaf:
## Summary

A long tool call rendered **one `<pre>` per chunk** (or one giant `<pre>` for completed Bash output) with **no windowing**, so the conversation accumulated thousands of in-flow DOM nodes and scroll tanked (the original "extremely slow, can hardly scroll" report).

## Fix

A shared `output-cap` helper + a non-silent marker, applied across **every** tool-output renderer:
- `output-cap.ts` — `MAX_TOOL_OUTPUT_LINES = 1000`, `capText(head/tail)`, `capChunksByLines` (O(kept), cheap per streamed append)
- `OutputHiddenMarker.tsx` — one "… N lines/blocks hidden" affordance (never truncates silently)
- Applied: `ToolOverlayLog` (chunk log + Read content), `BashOutputViewer` (stdout + stderr), `DiffViewer`, `CompactResult`
- `_document-nodes.scss` marker style (+ drops a stray 24px `.agent-tool-panel` left-indent)

Bounds the rendered DOM regardless of output size; the full output stays in `ToolNode` state (Phase 2 of the spec adds a "view full output" pane).

## Test plan

- [x] 12 unit tests (`output-cap.test.ts`) — head/tail, exact-budget boundary, empty, multi-line chunks, single-oversized-chunk, reference stability.
- [ ] Scroll a conversation with a long Bash / Read / diff output → smooth; muted "… N hidden" marker on big outputs; small outputs render unchanged.

## Notes

Full spec: `docs/specs/SPEC_TOOL_OUTPUT_CAP_2026_05_30.md` (Phase 2 = view-full pane, Phase 3 = virtualize it).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1217, opened 2026-05-31
fix(agent-pane): byte-cap single huge lines in tool output

AgentA-asaf:
## Summary

Follow-up to #1215 — addresses the remaining byte-cap item from codex's review there.

`capText` bounds tool output by **lines**, so a payload that fits the 1000-line budget but is **one enormous line** (minified JSON, base64, a long compiler line) was returned **uncapped** — still a multi-MB `<pre>`, reproducing the scroll/DOM bloat the caps prevent. Every render path (Bash stdout/stderr, Read, DiffViewer, CompactResult, the oversized stream-chunk boundary) funnels through `capText`, so all were affected.

## Fix

- `MAX_TOOL_OUTPUT_CHARS` (~1 MB) — comfortably above any real multi-line output, well below the multi-MB blobs that bloat the DOM.
- `capText` trims the result to that budget (head/tail, matching its line behavior) with a visible `…(truncated)` marker, so it is never silent.

## Test plan

- [x] 24 output-cap tests (2 added: char-caps a single ~2 MB line; leaves normal multi-line output untouched).
- [ ] Smoke: a tool emitting a multi-MB single-line payload renders a bounded `<pre>` with the truncation marker; normal output unchanged.

### DIFF
--- a/.changesets/1780245813-fix-agent-pane-byte-cap-single-huge-lines-in-tool-output-vef0.md
+++ b/.changesets/1780245813-fix-agent-pane-byte-cap-single-huge-lines-in-tool-output-vef0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): byte-cap single huge lines in tool output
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Merging on reagent's approval. codex's remaining point — a *total* character budget across the streamed chunk window (vs the per-chunk `capChars` here) — is valid for a large blob streamed across many sub-1 MB chunks. Node count is already bounded (≤ MAX_TOOL_OUTPUT_LINES `<pre>`s via the line cap), so this is bounded-but-large in that rare case, not unbounded. Tracking the total-byte budget as a Phase-2 item alongside chunk-log virtualization rather than expanding this follow-up further.

PR #1218, opened 2026-05-31
fix(agent-pane): gate send-now zone on an interruptible turn

AgentA-asaf:
## Summary

The queued-messages zone (`PendingMessagesPanel`) rendered whenever `pendingMessages()` was non-empty. But during the **`Submitting`** phase of every idle send, the message is optimistically enqueued for a frame before the turn starts — so the **"Queued" box flashed on every normal send**, with no running turn to queue behind.

## Fix

Gate the whole zone on a genuinely **interruptible** turn (`Streaming` / `Interrupting`) via a new `interruptibleTurn` accessor (`isInterruptibleTurn(turnPhase)`) — so the zone only appears when there's actually a running turn the queued messages sit behind. Defaults to shown when the prop is absent (back-compat).

## Test plan

- [ ] Send a message while the agent is **idle** → no "Queued" flash.
- [ ] Send while a turn is **streaming** → queued zone + "Send now" appear as before.

### DIFF
--- a/.changesets/1780246791-fix-agent-pane-gate-send-now-zone-on-an-interruptible-turn-091f.md
+++ b/.changesets/1780246791-fix-agent-pane-gate-send-now-zone-on-an-interruptible-turn-091f.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): gate send-now zone on an interruptible turn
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1221, opened 2026-05-31
feat(macos): standard notarized DMG — CEF 148, patched renderer, launcher splash, lean packaging

a5af:
Produces a **standard, double-clickable, notarized macOS DMG** for AgentMux on macOS 26 (Tahoe). Consolidates all macOS packaging tooling plus two stability fixes.

### CEF 148 + renderer fix
- **CEF 146 → 148 bump** (Chromium 148.0.7778) — cross-platform code update, compiles clean on all platforms.
- **Patched CEF 148 framework fixes the macOS-26 renderer crash.** The signed/packaged renderer crash-looped with `process_requirement.cc … Error -67030`. Fixed at source: `GetPeerValidationPolicy() → kNoValidation` (`docs/cef-patches/`). Framework is **Metal-accelerated** (ANGLE-Metal / AGXMetal verified at runtime).
- **Per-type helper apps** — CEF 148 requires all six variants: Renderer/GPU/Plugin/Alloy/generic + **Alerts** (CEF 148 spawns an alert-service helper; its absence caused endless `posix_spawnp` retries and crash-loops). `package-macos.sh` creates and signs all six.

### Launcher-owned instant splash (A0)
- **`agentmux-launcher` is now `CFBundleExecutable`** — the entry point of the packaged `.app`. The launcher is a tiny binary up in milliseconds; it paints the native splash *before* CEF loads, then spawns + supervises srv and the host (`run_unix`).
- **`splash_mac.rs`** — native AppKit splash (raw ObjC FFI, no new deps): dark rounded card (RGB 26,26,31, matching the Win32 splash) with the brain logo pulsing (fade-in → 1.1 Hz sine), fades out on host first-frame signal. Runs on the main thread; Tokio supervisor runs on a worker thread.
- **Host dismiss signal** (`client/mod.rs`) — touches `AGENTMUX_SPLASH_READY_FILE` (inherited by the host via env) on first CEF paint; launcher polls it and fades the splash out.
- Also gains the **launcher-supervised srv model** on macOS: srv survives a host crash.

### Crash fix — `CreateWindowTask` null-client guard
- **Root cause:** tearing off a pane with a 2nd window open could SIGABRT (`CrBrowserMain`). `CreateWindowTask::execute()` selected a client via `list_browsers()` + label-prefix heuristic; if concurrent `on_before_close → UnregisterBrowser` raced the snapshot, `client = None` → `browser_view_create(None, …)` → CEF C++ `CHECK` → abort.
- **Fix:** switch to `list_top_level_browsers()` (the proper reducer helper, filters by `kind: TopLevel`) and add an explicit early-return guard if no live client is found — log an error and bail before calling CEF, instead of crashing the process.

### Lean DMG — ~169 MB
- `strip -S -x` on libcef + all dylibs/helper binaries (564 → 321 MB framework)
- Dedup helper GL libs (only GPU + generic carry them)
- Trim Chromium locales → en* only (−52 MB)
- **ULMO (LZMA) compression** — same compression class as Linux's SquashFS AppImage; was 248 MB with UDZO (zlib)
- Icon-view drag-to-install DMG layout (AgentMux.app + /Applications alias)
- Cert auto-detected from keychain (no hardcoded identity in public repo)

## Verified
- `spctl -a -t open` → **accepted, source=Notarized Developer ID**
- `xcrun stapler validate` → ticket stapled  
- All six helper apps present and signed
- Splash appears instantly on launch, dismisses on first frame
- One Dock tile (host), no duplicate (launcher runs as accessory)
- Host stable from `/Applications` — no crash-loop, no `--disable-gpu` degradation
- `CreateWindowTask` guard tested: no SIGABRT on multi-window tear-off

## Notes
- **Run from `/Applications`, not the mounted DMG.** Chromium's `code_sign_clone` needs a same-device writable path; launching from the DMG volume causes cross-device errors and a crash-loop.
- Workstream A1/A2 (Unix socket IPC for full launcher reducer/saga coordination on macOS) is a follow-on PR — the null-client guard is the immediate stability fix.

### DIFF
--- a/.changesets/1780159576-feat-macos-package-macos-signed-app-dmg-with-cef-helper-app-i709.md
+++ b/.changesets/1780159576-feat-macos-package-macos-signed-app-dmg-with-cef-helper-app-i709.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(macos): package:macos — signed .app/.dmg with CEF Helper app
--- a/.changesets/1780224165-feat-macos-patched-cef-148-framework-per-type-helpers-fix-th-ko6e.md
+++ b/.changesets/1780224165-feat-macos-patched-cef-148-framework-per-type-helpers-fix-th-ko6e.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(macos): patched CEF 148 framework + per-type helpers fix the renderer on macOS 26
--- a/.changesets/1780248067-fix-macos-lean-notarized-dmg-ulmo-lzma-strip-locale-trim-167-re5q.md
+++ b/.changesets/1780248067-fix-macos-lean-notarized-dmg-ulmo-lzma-strip-locale-trim-167-re5q.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): lean notarized DMG (ULMO/LZMA + strip + locale trim) — 167MB
--- a/.changesets/1780253873-feat-macos-launcher-owned-instant-splash-launcher-as-package-cfts.md
+++ b/.changesets/1780253873-feat-macos-launcher-owned-instant-splash-launcher-as-package-cfts.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(macos): launcher-owned instant splash + launcher as packaged entry point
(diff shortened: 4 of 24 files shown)


a5af:
@codex review

PR #1222, opened 2026-05-31
perf(block): swap `.block-mask` backdrop-filter for `will-change: transform`

AgentU-asaf:
## Summary

`.block-mask` carries the per-pane focus border (2 px ring around every workspace pane). It used `backdrop-filter: blur(0.1px)` as a layer-promotion hack to ensure it composites above xterm's `.xterm-viewport` GPU-scroll layer. This swaps that to `will-change: transform` — same layer-promotion guarantee, an order of magnitude cheaper per frame.

## Why backdrop-filter was the wrong tool

`backdrop-filter` is the **most expensive** layer-promoter in Chromium. Even at 0.1 px blur it creates a backdrop layer that **samples and blurs the layers below every frame**, scaling linearly with pane count.

Linux Chromium-Ozone-Wayland measurements before this commit (10 panes, idle, via `scripts/capture-trace.cjs` + CDP `Profiler.start`):

| | |
|---|---|
| `PaintArtifactCompositor::Update` | ~48 ms per `BeginMainFrame` |
| `BeginMainFrame` total | ~92 ms (16 ms budget) |
| Effective frame rate | ~1 fps |
| V8 idle wall time | 91 % |
| Bottleneck | **100 % compositor** |

Visible symptom on Linux: hold a key in a terminal pane → nothing appears → release → all characters dump at once. PTY echoes arrive in JS land but can't paint until the next frame budget arrives, which is up to a second away.

## Why `will-change: transform` is safe

The original comment cited "Chromium may squash or reorder" `will-change` layers. That's a non-issue here:
- The rule already has `z-index: var(--zindex-block-mask-inner)`, which sets paint order **within the parent stacking context**.
- Chromium squashing groups **preserve relative z-order** across squashed siblings.
- Empirical: focus ring renders correctly above xterm's scroll layer in local testing.

## Empirical impact

`BeginMainFrame` drops from ~92 ms to ~80 ms (~13 % per-frame reduction). **This is not a complete fix** for the Linux terminal-typing latency (separate spec: `aa6f56b9 docs(spec): terminal flow control`). The remaining frame budget is split across:

- Other backdrop-filter sites — modal (`blur(8px)`), magnify overlay (`blur(--magnified-block-blur)`), conn-status overlay (`blur(50px)`), code-block copy-button (`blur(8px)`). All transient / conditional — only cost when the relevant state is active.
- Wayland-side compositor cost (the `WaylandZwpLinuxDmabuf::OnTrancheFlags Not implemented` warning at startup — Chromium-Mutter dmabuf negotiation falling back to a slower path).
- Sysinfo's 1 Hz status-bar re-render triggering the frame in the first place.

But this is the **largest single steady-state per-pane compositor cost** in the layer tree, applied to *every* pane all the time, and the obvious first win to unblock the next steps (CEF launch-flag pass to match VSCode's `EarlyEstablishGpuChannel`/`EstablishGpuChannelAsync`, audit of the conditional backdrop sites, broader compositor-layer reduction).

## Test plan

- [ ] `task build:frontend` passes (✅ verified locally)
- [ ] Focus ring on the active pane renders correctly (above xterm content) — Linux + macOS + Windows
- [ ] Unfocused panes still show the default border color
- [ ] No regression in pane magnify backdrop (separate `backdrop-filter` rule at `block.scss:313`)
- [ ] Capture a fresh trace via `node scripts/capture-trace.cjs` and confirm `PaintArtifactCompositor::Update` drops


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
codex P2 is valid — the release script treats every non-blank line after the frontmatter as a separate VERSION_HISTORY bullet. The changeset body needs to be trimmed to the single summary line; the detailed perf data (measurements, rationale, squash analysis) should stay in the PR body and commit message where it already lives.

Suggested changeset body (only the first line after the frontmatter divider):

```
---
type: patch
---

perf(block): replace `.block-mask` `backdrop-filter` with `will-change: transform`
```

Everything else (Linux trace numbers, squash/reorder analysis, scope note) can be dropped from the changeset — it's all in the PR body. Once that's pushed, reagent is already APPROVED so we can merge immediately.

a5af:
@codex review

PR #1223, opened 2026-05-31
feat(term): predictive local echo for terminal input

AgentA-asaf:
## Summary

VS Code's terminal and AgentMux's are **both xterm.js on WebGL** — identical rendering. The only thing that makes our typing feel less immediate is the **echo round-trip**: VS Code's PTY (`node-pty`) is in-process (a µs callback); ours lives in the sidecar, so every keystroke crosses `frontend → WS → sidecar PTY → echo → WS → frontend → paint`.

This predicts the echo: paint a just-typed printable character **in the keydown frame**, then reconcile byte-exact against the authoritative PTY echo (consume if it matches, roll back if it diverges). The authoritative data path is untouched; predictions are a transient, self-correcting overlay that always converges to the PTY stream.

## Design (full spec: `docs/specs/SPEC_TERMINAL_PREDICTIVE_LOCAL_ECHO_2026_05_31.md`)

- **Observational arming (§7.1)** — never predict without a *confirmed* echo first, so a password prompt (echo off) never flashes plaintext: the first keystroke of every burst is authoritative-only, and confirmations stop the instant echo stops.
- **Byte-exact reconcile + rollback** — divergence (autocap, `^C`, TUI output) rolls back all outstanding predictions; the buffer converges to exactly the PTY stream.
- **Alt-screen / non-printable / CJK** fall through to the authoritative path (outside the Phase-1 safe set).
- **RTT-gate + cooldown** — gate opt-in via `term:predictiveecho:thresholdms` (default 0 = always predict once armed).

## Files

- `predictive-echo.ts` — pure state machine (queue, reconcile, arm, RTT); xterm side effects injected via a `PredictSink` for testability.
- `predictive-echo.test.ts` — 11 tests: arming, **password-no-flash**, predict/confirm/diverge, timeout, flush, RTT gate, **convergence property**.
- `termwrap.ts` — `onInput` paints in the keydown frame; `doTerminalWrite` reconciles (consume painted / roll back on divergence); `ptyOffset` tracks full PTY bytes. **Default-on** (`term:predictiveecho=false` to opt out).

## Test plan

- [x] 11 unit tests incl. convergence.
- [ ] **Manual:** type in a terminal → chars paint instantly; `read -s` / `sudo` → no plaintext flash; `vim` → no corruption.

## Notes

Phase 1 of the spec (cooked-mode, single-line, ASCII). Phases 2–4 (explicit Unix `tcgetattr` echo signal + alt-screen parser; CJK/wrap width-awareness; tuning) tracked in the spec. Part of the typing-perf umbrella (#1161).

### DIFF
--- a/.changesets/1780268491-feat-term-predictive-local-echo-for-terminal-input-z86z.md
+++ b/.changesets/1780268491-feat-term-predictive-local-echo-for-terminal-input-z86z.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(term): predictive local echo for terminal input
--- a/docs/specs/SPEC_TERMINAL_PREDICTIVE_LOCAL_ECHO_2026_05_31.md
+++ b/docs/specs/SPEC_TERMINAL_PREDICTIVE_LOCAL_ECHO_2026_05_31.md
@@ -1,0 +1,195 @@
+# SPEC: Terminal Predictive Local Echo
+
+**Date:** 2026-05-31
+**Author:** AgentA
+**Status:** Draft → implementation
+**Tracks:** Discussion #1161 (typing-perf umbrella) · `SPEC_INPUT_RESPONSIVENESS_TERMINAL_AND_AGENT_2026_05_29.md`
+**Supersedes the "shelved" call on predictive echo** (AgentY 2026-05-30) — see §2.
+
+---
+
+## 1. Motivation
+
+AgentMux's terminal and VS Code's integrated terminal are **the same renderer**: xterm.js on the WebGL addon (`termwrap.ts:150`, `loadRendererAddon`; WebGL is the Windows/macOS default via `useWebGl = !term:disablewebgl`). We are **at parity on rendering**. Measured on Windows the JS path is sub-millisecond (`term-keypress` p95 0.4ms, `term-echo-render` p95 0.3ms, zero long-tasks, ~60fps).
+
+The gap to VS Code is **structural and lives entirely in the echo round-trip**:
+
+| | VS Code | AgentMux |
+|---|---|---|
+| PTY location | **in-process** (`node-pty` in the renderer/extension host) | **out-of-process** (`agentmux-srv` sidecar) |
+| keystroke → echo | one in-process callback (µs) | `controllerinput` RPC **over WS** → sidecar async runtime → `portable-pty` writer (`shell.rs:761`) → PTY → echo → reader loop (`shell.rs:804`) → WPS broker → **WS** → `handleNewFileSubjectData` → base64 decode → `xterm.write` |
+
+Every keystroke pays a cross-process, multi-hop round-trip — two WS traversals, the sidecar's async scheduler, the `#951` seq-reorder buffer, base64 — that VS Code does not. At local-PTY speeds this is small in absolute ms but it is **a meaningful fraction of a frame on top of an already-full frame budget**, and it is the only thing differentiating us from VS Code once rendering is equal. The fix is to **stop waiting for the round-trip to paint the character the user just typed.**
+
+## 2. Why this was shelved, and why we are un-shelving it
+
+AgentY (2026-05-30) shelved predictive echo: *"it optimizes a latency that doesn't exist locally; the ≤512B RAF-bypass already paints echo same-frame."* That reasoning conflates two things:
+
+- **Render latency** (echo *data* → pixels): yes, already same-frame (#278 RAF-bypass).
+- **Transport latency** (keystroke → echo *data arrives*): **not** zero — it is the full cross-process round-trip above, and it is exactly what the user perceives as "less immediate than VS Code."
+
+AgentX's RFC explicitly left the door open: *"Only revisit with a sidecar-signaled tty mode + RTT telemetry."* This spec satisfies both conditions. The two real objections — **(a)** flashing plaintext over a password prompt (a security bug, not a latency one) and **(b)** corrupting CJK / TUI — are handled head-on in §6/§7, the same way VS Code's `TypeAheadAddon` handles them in production.
+
+## 3. Goals / Non-goals
+
+**Goals**
+- Paint a just-typed printable character **in the same frame as the keydown**, before the authoritative echo returns — matching VS Code's perceived immediacy.
+- **Never** render predicted text that the PTY would not have echoed (passwords, raw mode, TUI).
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Addressed in 5e365d62 — all three findings share one root (sticky `armed` across mode boundaries):

**reagent P0 / codex P1 (input path):** `flush()` now clears `armed`, not just rolls back. Every password prompt / TUI is entered *through* a non-printable key (the Enter that runs `sudo`/`ssh`/`vim`), so the first keystroke of the new context starts unarmed → observed, never painted. The armed→echo-off hole is closed. New test: `disarms at the Enter boundary so a following echo-off password char is never painted`.

**codex P1 (output path):** `reconcile` detects alt-screen **enter** (`CSI ?1049h|?47h|?1047h`) and disarms *before* the switch lands (the erase is still on the normal buffer), so a TUI that takes over via output can't be mispredicted and no `CSI K` rollback ever runs inside the alt buffer. New test: `disarms on alt-screen enter output`.

**reagent P1 (sweep needs a chunk):** the unconfirmed-prediction sweep is now also driven by the **keystroke stream** — `onInput` sweeps before handling each key. Deliberately **no wall-clock timer** (project rule: no grace timers); the keystroke stream is the deterministic trigger, so if echo stalls with no chunk the next key ages out the stale prediction. With the disarm fix this path is essentially unreachable anyway (you can only paint while armed, and arming requires a live confirmed echo), but it's there as defense-in-depth. New test: `sweeps a stale painted prediction on the next keystroke, with no PTY chunk and no timer`.

14/14 unit tests pass. Spec §6.3/§7.1/§9 updated to match.

a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
P2 (disable-mid-prediction double render) fixed in 34e57e4d. The output path (`doTerminalWrite`) now detects `!isEnabled() && pending > 0` and calls `reset()` — rolling back the speculative glyph before the authoritative echo writes — so it renders exactly once instead of duplicating until the next keystroke. Regression test added (`reset() rolls back a pending painted prediction so the authoritative echo writes cleanly`); 15/15 pass.

AgentA-asaf:
Fixed in 8691d447:

**reagent P1 (settings schema):** Added `term:predictiveecho` (boolean) and `term:predictiveecho:thresholdms` (number) to `SettingsType` in `gotypes.d.ts` and to `schema/settings.json` — both were missing, causing a type error on `getSettingsKeyAtom` calls and schema rejection.

**reagent P2 (dormant decode cost):** `doTerminalWrite` now guards the decode + `reconcile` (which runs the `ALT_SCREEN_ENTER` regex) behind `pending > 0 || isArmed` — when fully dormant the authoritative path is untouched, zero cost until the first arming confirmation.

**codex P1 (in-burst programmatic echo-off):** This is a known Phase 1 limitation, documented inline. The non-printable disarm handles the common case (every real password prompt is entered via the Enter that runs the command). The remaining edge (a program switches echo off between printable chars with no non-printable) self-heals via sweep within 600ms; the deterministic fix is Phase 2 tcgetattr. Not a Phase 1 blocker.

a5af:
@codex review

AgentA-asaf:
Fixed in 44c382c7:

**reagent P1 (burst scramble):** `onInput` painted whenever `armed=true` regardless of unconfirmed observations ahead in the queue. During a high-RTT burst, if "a" is observed (unarmed) and "b" is typed before "a" echoes, queue=`[{a,obs}]`. Painting "b" places its glyph before "a"'s authoritative echo writes, scrambling "ab" to "ba" with no divergence/rollback triggered (each echo still matches its queue head). Fix: `hasPendingObservation` guard — block painting when any `painted:false` entry is ahead. Test: `burst ordering`.

**codex P2 (alt-screen re-arm in same chunk):** A PTY chunk containing both an echo confirmation AND `CSI ?1049h` disarmed in the alt-screen block then immediately re-armed at `this.armed = true` in the confirmation loop — predictor ends up armed after the alt-screen enter. Fix: track `altScreenEntered` locally in `reconcile` and skip the re-arm assignment when it's set. Test: `same-chunk disarm`.

**codex P2 (TextDecoder buffering):** `stream:true` accumulated partial multibyte UTF-8 across WS chunks. If predict is disabled/flushed before a continuation chunk arrives, leading bytes strand in the decoder while subsequent chunks bypass it → CJK/emoji corruption. Fix: `stream:false` per call (stateless), same as xterm's raw fallback.

17/17 tests pass.

a5af:
@codex review

AgentA-asaf:
Fixed in 84083aef.

Both bots converged: converting `Uint8Array` through `TextDecoder` (any mode) corrupts multibyte UTF-8 sequences split across WS chunks.

**Fix: reconcile now works at the byte level.**

- Signature changed to `reconcile(chunk: Uint8Array): { auth: string; rest: Uint8Array }`.
- Queue entries are single printable ASCII bytes (0x20-0x7e) — byte-by-byte comparison is exact.
- The unconsumed tail (`rest`) is returned as the ORIGINAL `Uint8Array` subarray — **never decoded** — so xterm receives raw bytes and assembles multibyte chars natively.
- `auth` (observation echoes that must be written) is always ASCII → safe as string.
- Alt-screen regex check uses a one-shot `TextDecoder` scan only (the sequence is pure ASCII control chars, no multibyte possible).
- `echoDecoder` field removed from `termwrap.ts`. `doTerminalWrite` chains two `terminal.write()` calls when both auth and rest are non-empty.

Right-margin rollback (codex P2 #2) is a known Phase 1 limitation — `CSI n D` + `CSI K` only works within the current line; Phase 2 cursor-column tracking will fix this. Deferred per the spec's Phase 2 scope boundary.

17/17 tests pass.

a5af:
@codex review

AgentA-asaf:
Fixed in 4a019359.

**codex P1 / reagent P2 (right-margin wrap):** `CSI n D` + `CSI K` only operates within the current terminal row — when predicted glyphs wrap onto the next line, a rollback runs from the post-wrap cursor position and clears the wrong row, leaving speculative chars on the prior row (convergence violated).

Fix: in `handleTermData`, immediately before `predict.onInput`, read `cursorX` and `cols` from the live xterm buffer. If `cursorX >= cols - 1` (cursor at or past the last column), call `reset()` to flush pending predictions and disarm. This ensures predictions are always committed within a single row where `CSI K` rollback is exact. Re-arming resumes normally on the next confirmed echo after the wrap.

a5af:
@codex review

PR #1224, opened 2026-06-01
perf(cef): enable EarlyEstablishGpuChannel + EstablishGpuChannelAsync

AgentU-asaf:
## Summary

Adds `--enable-features=EarlyEstablishGpuChannel,EstablishGpuChannelAsync` to the CEF browser-process command line. Both ship enabled in stable Chrome on Linux and are explicitly set by VSCode's Electron (confirmed via `/proc/<pid>/cmdline` on a running VSCode on the same machine); CEF does not enable them by default.

`EarlyEstablishGpuChannel` requests the GPU process channel before the renderer's first paint instead of synchronously on first paint; `EstablishGpuChannelAsync` treats establishment as non-blocking. Together they let the compositor start producing frames against the GPU process sooner.

## Empirical impact (Linux Chromium-Ozone-Wayland, 10 panes, 12 s held key)

Captured via `scripts/capture-trace-ipv6.cjs`:

| | Before | After |
|---|---|---|
| `BeginMainFrame` avg | 39.1 ms | **35.7 ms** (-9%) |
| `BeginMainFrame` count | 12 | 12 (unchanged) |
| Frame cadence | ~1 Hz | ~1 Hz (unchanged) |
| `LayerTreeHostImpl::DidNotProduceFrame` | 66 | 60 |

Small per-frame win (~3 ms / frame). The Wayland frame-production stall that holds the cadence at ~1 Hz under sysinfo invalidation is a separate problem — the renderer rejects 60+ Mutter `BeginFrame` requests as `DidNotProduceFrame` because xterm.js WebGL canvas writes don't surface as page-level dirtiness. Predictive echo (#1223) writes to xterm's internal buffer instantly, but the render-to-pixels step is gated on those 1 Hz frames.

**This PR is not the typing-latency silver bullet — it's the matching VSCode flag set so we're not unnecessarily off-default while we investigate the real fix.**

## Test plan

- [ ] `task build:host` succeeds (✅ verified locally)
- [ ] App launches; renderer/utility subprocesses show `--enable-features=EarlyEstablishGpuChannel,EstablishGpuChannelAsync` in `/proc/<pid>/cmdline`
- [ ] `scripts/capture-trace-ipv6.cjs` reports no new errors
- [ ] No GPU process / WebGL regressions in normal workflows

### DIFF
--- a/.changesets/1780305472-perf-cef-gpu-channel-flags-v674.md
+++ b/.changesets/1780305472-perf-cef-gpu-channel-flags-v674.md
@@ -1,0 +1,35 @@
+---
+type: patch
+---
+
+perf(cef): enable EarlyEstablishGpuChannel + EstablishGpuChannelAsync
+
+Adds `--enable-features=EarlyEstablishGpuChannel,EstablishGpuChannelAsync`
+to the CEF browser-process command line in
+`AgentMuxApp::on_before_command_line_processing`. Both features ship
+enabled in stable Chrome on Linux and are explicitly set by VSCode's
+Electron (confirmed via `/proc/<pid>/cmdline` on a running VSCode); CEF
+does not enable them by default.
+
+They (a) request the GPU process channel before the renderer's first
+paint instead of synchronously on first paint and (b) treat the channel
+establishment as non-blocking, which lets the compositor start producing
+frames against the GPU process sooner.
+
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1226, opened 2026-06-01
perf(package): strip source maps from release portables (~28 MB saved)

AgentA-asaf:
## Summary

- Adds `task package:release` — builds a portable with source maps stripped (saves ~28 MB, 7% of ZIP).
- `task package` (dev portables) is unchanged — maps stay so the runtime source-map resolver works during local iteration.
- Only `task package:release` should be used when building artifacts for GitHub releases.

## How it works

`scripts/package.sh` checks `STRIP_MAPS=1` (set by `task package:release`) and runs `find dist/frontend -name "*.map" -delete` after the Vite build and before bundling. No Vite config changes — `sourcemap: true` stays unconditional so dev builds always have maps.

## Docs

- `docs/specs/SPEC_PORTABLE_SOURCE_MAPS_2026_06_01.md` — the rule, how it works, future-agent guidance
- `docs/analysis/ANALYSIS_PORTABLE_SIZE_REDUCTION_2026_06_01.md` — full portable size breakdown

## Test plan

- [ ] `task package` on a feature branch → ZIP contains `.map` files ✓
- [ ] `task package:release` → ZIP contains no `.map` files ✓
- [ ] `task dev` unaffected ✓

### DIFF
--- a/.changesets/1780308635-perf-package-strip-source-maps-from-release-portables-28-mb--by8c.md
+++ b/.changesets/1780308635-perf-package-strip-source-maps-from-release-portables-28-mb--by8c.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+perf(package): strip source maps from release portables (~28 MB saved)
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -99,10 +99,10 @@ tasks:
             - bash scripts/package.sh {{.CLI_ARGS}}
 
     package:release:
-        desc: 'Like `task package` but strips source maps from the frontend bundle (~28 MB). Use this when building the artifact that goes into a GitHub release. Dev portables (`task package`) keep maps for local debugging. See docs/specs/SPEC_PORTABLE_SOURCE_MAPS_2026_06_01.md.'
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentA-asaf:
Fixed in 72a86a5a.

codex P1: `scripts/package.sh` unconditionally set `AGENTMUX_BUILD_CHANNEL_DEFAULT` to the branch-scoped `dev-portable-<branch>` channel. A release portable built with `task package:release` would boot users into their dev channel rather than `stable` — appearing as a fresh install with no data.

Fix:
- `scripts/package.sh` now respects a `RELEASE_CHANNEL` env var and uses it to override the computed dev-portable channel.
- `task package:release` now passes `RELEASE_CHANNEL=stable` alongside `STRIP_MAPS=1`.
- `task package` (dev portables) is unchanged — still uses branch-scoped channel.
- Spec `SPEC_PORTABLE_SOURCE_MAPS_2026_06_01.md` updated to document the channel distinction.

a5af:
@codex review

AgentA-asaf:
Fixed in d67a059f:

**P1 (version downgrade):** Rebased on `main` — branch is now at 0.41.1.

**P2 (SKIP_MAP_STRIP):** Spec referenced `SKIP_MAP_STRIP=1` which the script never checks. Removed the stale example; correct workaround is to comment out the `find … -delete` line temporarily.

**P2 (--fresh + RELEASE_CHANNEL conflict):** The fresh-stamp suffix was silently overwritten by `RELEASE_CHANNEL`. Added an explicit `exit 1` with a clear error message — the two flags are mutually exclusive (a distributed artifact uses a fixed channel; a throwaway fresh dir doesn't apply to it).

PR #1227, opened 2026-06-01
fix(launcher): include build version in pipe name hash (version isolation P0)

AgentA-asaf:
## Summary

- Two different release versions (e.g. 0.40.2 and 0.41.0) sharing the same channel (`stable`) resolved to the same named-pipe path → second version silently activated the first's window instead of launching.
- `data_dir_hash16` now takes `version: &str` and hashes `canonical_path\x00version`. Different versions → different pipes → both launch independently.
- 4 call sites updated (main.rs × 1, diag.rs × 2, all already had `version` in scope). New invariant test added.

## Root cause

`BUILD_CHANNEL_DEFAULT = "stable"` is baked into every release binary at compile time. All release versions share `~/.agentmux/channels/stable/data/`. Pipe name = `hash(data_dir)` — identical across versions → single-instance collision.

## What this does NOT change

- Data dir layout is unchanged — all versions of `stable` still share the same `channels/stable/data/`. This is intentional for now (agent/conversation persistence across updates). Phase 2 (version-scoped data dirs) is specced in `docs/specs/SPEC_VERSION_ISOLATION_2026_06_01.md`.
- Channel semantics are unchanged.

## Test plan
- [ ] `cargo test -p agentmux-launcher` — 185 tests pass including `different_versions_same_dir_produce_different_hashes`
- [ ] Build 0.41.0 portable + have 0.40.x running → 0.41.0 launches as independent instance

### DIFF
--- a/.changesets/1780311728-fix-launcher-include-build-version-in-pipe-name-hash-two-rel-7ql0.md
+++ b/.changesets/1780311728-fix-launcher-include-build-version-in-pipe-name-hash-two-rel-7ql0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(launcher): include build version in pipe name hash — two releases no longer share single-instance domain
--- a/.changesets/1780311728-fix-launcher-include-build-version-in-pipe-name-hash-two-rel-7ql0.md
+++ b/.changesets/1780311728-fix-launcher-include-build-version-in-pipe-name-hash-two-rel-7ql0.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(launcher): include build version in pipe name hash — two releases no longer share single-instance domain
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.41.0"
+version = "0.41.1"
 dependencies = [
  "anyhow",
  "base64",
@@ -28,7 +28,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.41.0"
+version = "0.41.1"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -58,7 +58,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.41.0"
+version = "0.41.1"
 dependencies = [
  "dirs",
  "serde",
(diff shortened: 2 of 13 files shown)


a5af:
@codex review

AgentA-asaf:
ipc-port P1 fixed in b431537a.

**Root:** host writes to `ipc-port` (shared across versions); forward_open_new_window reads from it. With different-versioned pipes, a double-click of the older version fails its pipe bind, reads the shared file (now pointing at the newer version's host), and activates the wrong window.

**Fix:** launcher passes `AGENTMUX_IPC_HASH` (already version-scoped = hash(data_dir + version)) to the host at spawn. Host writes to `ipc-port-{hash}`. forward_open_new_window takes `dir_hash` and reads `ipc-port-{hash}`. Both Windows and Unix spawn paths updated. Falls back to plain `ipc-port` for callers that don't set AGENTMUX_IPC_HASH.

a5af:
@codex review

AgentA-asaf:
Phase 2 (version-scoped data dirs) implemented in c3883356 — addresses the codex P1 directly.

**New layout for Installed/Portable:**
```
channels/<ch>/versions/<v>/data/       ← objects.db, sagas.db (version-isolated)
channels/<ch>/versions/<v>/cef-cache/
channels/<ch>/versions/<v>/runtime/    ← ipc-port-{hash}, lock
channels/<ch>/config/                  ← settings (channel-wide, survives upgrades)
channels/<ch>/agents/                  ← agent defs (channel-wide, survives upgrades)
```

Two concurrent releases now have fully independent data dirs — no concurrent SQLite writers, no schema migration hazard.

**Migration:** `migrate_legacy_data_dir()` in launcher's `data_dir.rs` copies `channels/<ch>/data/db/` → `channels/<ch>/versions/<v>/data/db/` on first run when old path exists. Copy, not move — older binaries keep their data.

76 (agentmux-common) + 185 (agentmux-launcher) tests pass including `two_installed_versions_have_distinct_data_dirs`.

a5af:
@codex review

AgentA-asaf:
Fixed in c69c8b65:

**codex P1 (migration always skipped):** `ensure_dirs()` was called first, creating `<new>/data/db/` as an empty dir. Then `migrate_legacy_data_dir` checked `new_db.exists()` — always true — and returned without copying. Users lost their data on first launch of the new version.

Fix: migration runs **before** `ensure_dirs`. Also replaced the dir-existence check with a db-file check (`.db` extension scan), so an empty dir from a previous failed launch doesn't block migration.

**reagent P2 (stale call-site count):** spec updated to 2 call sites in `main.rs` at both the prose and §6 table.

a5af:
@codex review

PR #1229, opened 2026-06-01
fix(host): gate renderer OOM recovery on system memory (Phase 1a)

AgentA-asaf:
## Summary

Makes renderer-crash recovery **memory-aware**. A renderer OOM while the Windows system commit limit is exhausted is transient OS pressure — not a broken renderer — but the existing recovery (`SPEC_GRACEFUL_CRASH_HANDLING`) treated it like any crash:

- the **3-in-10s crash budget** falsely declared "give up" on a window that's fully recoverable once commit frees, and
- the Reload button **re-OOMs** if clicked while commit is still exhausted.

This was the 2026-06-01 incident: `avail_page_gb=0` for ~20 min (many concurrent AgentMux instances × Chromium trees), renderer OOM-terminated. The host main process survived — only a renderer died — but the recovery UX was poor.

## What changed

1. **`memory_heartbeat.rs`** — `commit_free_mb()`: an on-demand `GlobalMemoryStatusEx` probe (µs) plus a published atomic the crash path reads. Returns `u64::MAX` on non-Windows → behavior there unchanged.

2. **`on_render_process_terminated`** — when `status == PROCESS_OOM` **and** commit-free `< RESUME_FLOOR_MB` (512 MB): **do not consume the wedged-slot crash budget**. Show a recoverable "low memory" paused page — **Resume** (navigate to the live app URL → re-project all state from srv, losing nothing) / **Quit** — guided so the user frees memory first. Bounded by a separate `MEMORY_PAUSE_BUDGET` (5/30s) so *total* exhaustion (even the tiny page can't render, re-firing the handler) still converges on the existing give-up page rather than looping. **Non-OOM and OOM-with-headroom paths are unchanged.**

3. **`record_memory_pause()`** — extracted prune+budget step, **unit-tested** (3 tests: floor sanity, budget convergence, window pruning).

The key discrimination: **OOM-under-memory-pressure is the OS's fault and transient — it must not burn the wedged-slot budget; OOM-with-memory-available is the renderer's fault — it still does.**

## Why state is safe

The renderer is state-poor; the sidecar (`srv`, a separate process that survives the renderer death) holds all durable state. Resume re-projects it exactly — nothing is lost. This is the same free-ride the existing Reload uses.

## Scope / verification

- **Phase 1a (this PR):** the memory discrimination + budget bypass + manual recoverable paused page.
- **Phase 1b (follow-up):** host-driven, memory-gated **auto-resume** (`post_delayed_task` probing commit-free) so the window comes back *on its own*, plus a renderer-free **native overlay** (reusing `splash.rs`) for total exhaustion. Split out because it needs a real induced-OOM smoke.
- **Verification:** `cargo check --release -p agentmux-cef` clean (zero new warnings) + 3 unit tests pass. **Runtime OOM smoke is still pending** (inducing real commit exhaustion) — see spec §11. Flagging explicitly rather than claiming runtime-verified.

Spec: `docs/specs/SPEC_GATED_RENDERER_RECOVERY_2026_06_01.md` (full design incl. Phases 1b/2/3).

## Test plan
- [x] `cargo check --release -p agentmux-cef` — clean
- [x] `cargo test --release -p agentmux-cef gated_recovery` — 3/3 pass
- [ ] Runtime: induce commit exhaustion (small page file + many panes), confirm a renderer OOM shows the paused page (not give-up) and Resume re-projects state — Phase 1b smoke

### DIFF
--- a/.changesets/1780341505-fix-host-gate-renderer-oom-recovery-on-system-memory-pause-i-dkgh.md
+++ b/.changesets/1780341505-fix-host-gate-renderer-oom-recovery-on-system-memory-pause-i-dkgh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(host): gate renderer OOM recovery on system memory — pause instead of false give-up
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -158,6 +158,24 @@ impl AgentMuxHandler {
         }))
     }
 
+    /// Reverse-lookup this browser's window label from the reducer's browsers
+    /// map (label → browser), by object identity. Mirrors the find-by-identity
+    /// loop in `on_before_close`. Used by the crash-recovery navigation so a
+    /// resumed *secondary* window preserves its `windowLabel` instead of
+    /// defaulting to `main` (the frontend treats a missing label as `main`,
+    /// which would re-register/route the recovered window as the wrong one).
+    /// (codex P2 on #1229.)
+    fn window_label_for(&self, browser: &mut Browser) -> Option<String> {
+        self.state
+            .list_browsers()
+            .into_iter()
+            .find(|(_, b)| {
+                let mut b = b.clone();
+                b.is_same(Some(&mut *browser)) != 0
+            })
+            .map(|(k, _)| k)
+    }
+
     fn on_title_change(&mut self, browser: Option<&mut Browser>, title: Option<&CefString>) {
         debug_assert_ne!(currently_on(ThreadId::UI), 0);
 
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Both addressed in a2fb2d1e.

**codex P2 (windowLabel dropped):** Real bug — and it affected the **pre-existing** crash recovery page too (same root), so I fixed both rather than just my new page. Added `window_label_for(&mut Browser)` (reverse-lookup from the reducer's browsers map by identity, mirroring `on_before_close`) and a shared `recovery_navigation_url()` builder that appends `&windowLabel=<label>` when known. Both the low-memory paused page and the existing recovery page now preserve the label, so a resumed secondary window comes back as itself instead of re-registering as `main`. Main-window / unknown-label behavior is unchanged.

**reagent P2 (stale spec status):** Updated the Status header to "Phase 1a implemented (PR #1229); Phases 1b/2/3 designed".

`cargo check --release` clean (no new warnings); 3/3 unit tests pass.

a5af:
@codex review

AgentA-asaf:
Rebased onto latest main (force-push abccb664). 

Both reagent P2s were about `ANALYSIS_PORTABLE_SIZE_REDUCTION` / `SPEC_PORTABLE_SOURCE_MAPS` — those belong to the already-merged #1226, not this feature. Root cause: this branch was accidentally cut from `agenta/strip-maps-from-portables` instead of `main`, so the PR diff duplicated #1226's (squash-merged) files. Rebased so the diff now contains **only** the gated-recovery changes (4 files: the changeset, `agentmux-cef/src/client/mod.rs`, `agentmux-cef/src/memory_heartbeat.rs`, `SPEC_GATED_RENDERER_RECOVERY`). The strip-maps files are gone from the diff, so those P2s no longer apply here.

The windowLabel fix (codex P2) and Phase 1a are intact; 3/3 unit tests pass.

(If the `v0.41.1+` version refs in #1226's already-merged docs are genuinely stale on main, that's a separate follow-up on main, not this PR.)

a5af:
@codex review

AgentA-asaf:
Fixed in d1af0e03 — generalized to preserve **all** creation params, not just windowLabel.

`recovery_target_url` now reads the crashed window's **own pre-crash URL** from its main frame (before any recovery page loads over it) and reuses it verbatim. That URL already carries everything — `windowLabel`, and for tear-off / floating-pane windows `workspaceId` / `floatingPaneId` — plus the still-valid in-process `ipc_port`/`ipc_token`. So a resumed floater/tear-off re-projects its exact state instead of opening blank. Falls back to the reconstructed `?ipc_port&ipc_token&windowLabel` URL only when the frame URL isn't a live `http://127.0.0.1` app URL (renderer died pre-commit, or it's a prior data: recovery page). Applied to both recovery sites.

cargo check --release clean; 3/3 tests pass.

a5af:
@codex review

PR #1230, opened 2026-06-01
fix(host): match dev localhost origin in crash-recovery URL reuse (follow-up to #1229)

AgentA-asaf:
## Summary

Follow-up to #1229 (gated renderer recovery), addressing the last codex P2 that landed just after that PR merged.

`recovery_target_url` reuses a crashed window's own pre-crash URL (to preserve `windowLabel` / `workspaceId` / `floatingPaneId`), but the "is this a live app URL" filter was hardcoded to `http://127.0.0.1`. In **dev mode**, `resolve_frontend_base_url` returns `http://localhost:<vite_port>`, so dev tear-off / floating-pane crash recovery fell back to the reconstructed URL and dropped `workspaceId` / `floatingPaneId` — bringing the window back as the wrong workspace/view.

**Fix:** match the pre-crash URL against the resolved `base_url` origin (127.0.0.1 in prod, localhost in dev) rather than a hardcoded host, via a new `url_on_origin` helper. The helper guards the origin boundary so a port that's a numeric prefix of another (`:5173` vs `:51730`) can't false-match.

## Test plan
- [x] `cargo check --release -p agentmux-cef` — clean
- [x] `cargo test --release -p agentmux-cef gated_recovery` — 3/3 pass

### DIFF
--- a/.changesets/1780344124-fix-host-match-dev-localhost-origin-in-crash-recovery-url-re-hip6.md
+++ b/.changesets/1780344124-fix-host-match-dev-localhost-origin-in-crash-recovery-url-re-hip6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(host): match dev localhost origin in crash-recovery URL reuse
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -2017,12 +2017,16 @@ setTimeout(tick, 1000);
 /// True when `url` is on the same origin as `base_url` (i.e. a live app URL we
 /// can safely reuse for recovery). The boundary check after the prefix guards
 /// against a port that is a numeric prefix of another (e.g. `:5173` matching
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Both fixed in 1cecad6a.

- **`#` boundary:** `url_on_origin` now accepts `#` as a valid post-origin boundary, so a hash-routed pre-crash URL (`http://localhost:5173#/route`) is reused instead of discarded.
- **Unit tests:** added `url_on_origin_matches_only_real_same_origin_urls` covering exact origin, origin+path, origin+query, origin+fragment, the port-numeric-prefix false positive (`:5173` vs `:51730`), a mismatched origin, and a `data:` recovery page. 4/4 gated_recovery tests pass.

a5af:
@codex review

PR #1231, opened 2026-06-02
fix(agent-pane): normalize zoom in virtualizer measurement — history overlap (Phase 1)

AgentA-asaf:
## Summary

Scrolling agent-pane **history** with per-pane zoom ≠ 100% made rows overlap (thinking text on top of messages), getting worse the further you scroll. At 100% zoom it was invisible.

## Root cause — empirically verified (CDP on cef-146)

The agent view is wrapped in CSS `zoom` (`agent-view.tsx:725`). The virtualizer's `measureElement` returns `getBoundingClientRect().height`, but `estimateSize` returns **fixed, unzoomed CSS-px** estimates. A CDP probe simulating the virtualizer's exact pattern (measure a 100px row, position the next at `translateY(measured)`):

| zoom | `getBoundingClientRect().height` | `offsetHeight` | gap between rows |
|---|---|---|---|
| 0.5 | **50** (scaled) | 100 (neutral) | **−25px → OVERLAP** |
| 1.0 | 100 | 100 | 0 (flush) |
| 1.5 | 150 | 100 | +75px (gap) |
| 2.0 | 200 | 100 | +200px (gap) |

So `getBoundingClientRect` **is** zoom-scaled in cef-146; the unzoomed estimates and zoomed measurements disagree at zoom ≠ 1, and because the resulting `translateY` offsets are applied *inside* the zoomed subtree, the zoom factor is **double-counted** → overlap when zoomed out, gaps when zoomed in, exactly 0 at 100%.

## Fix

Divide the measurement by the live `zoomFactor` so the virtualizer works entirely in zoom-independent CSS px (matching the estimates); the single ancestor `zoom` then scales offsets correctly. `zoomFactor` (the same `agent-view.tsx` memo that drives the CSS zoom — one source of truth) is plumbed through `AgentDocumentView` → `AgentDocumentVirtualList`. **No zoom-change re-measure needed** — the normalized measure is zoom-invariant and a row's layout height doesn't change with zoom.

(`offsetHeight` is zoom-neutral and would also work as a one-liner, but `GBCR / zoomFactor` keeps fractional precision and doesn't silently depend on `offsetHeight`'s neutrality across CEF upgrades.)

## Scope
- **Phase 1 (this PR):** the zoom normalization — removes the reported overlap.
- **Phase 2 (specced):** re-measure on collapse/expand/pin + a per-row ResizeObserver for in-renderer height changes (a secondary, any-zoom overlap source).

## Verification
- Mechanism empirically proven via CDP (table above) — not a model.
- Type-clean plumbing; dev server hot-reloaded with no errors.
- Manual: set a Terminal/Agent pane zoom ≠ 100% (`term:zoom`), scroll a long history with thinking blocks — no overlap at 0.5 / 1.5 / 2.0.

Spec: `docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_ZOOM_OVERLAP_2026_06_01.md`

### DIFF
--- a/.changesets/1780364832-fix-agent-pane-normalize-zoom-in-virtualizer-measurement-fix-kdun.md
+++ b/.changesets/1780364832-fix-agent-pane-normalize-zoom-in-virtualizer-measurement-fix-kdun.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): normalize zoom in virtualizer measurement — fixes history overlap
--- a/docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_ZOOM_OVERLAP_2026_06_01.md
+++ b/docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_ZOOM_OVERLAP_2026_06_01.md
@@ -196,8 +196,10 @@ priority; do only if jank-on-scroll-in remains visible after 4.1/4.2.
 
 ## 5. Edge cases
 
-- **Zoom mid-scroll:** §4.1's `createEffect` on `zoomFactor()` re-measures, so a
-  zoom change while scrolled deep doesn't strand stale offsets.
+- **Zoom mid-scroll:** no special handling needed (§4.1). A value cached at one
+  zoom (`GBCR/Z₁ = h` CSS px) stays correct at any other zoom, and the browser
+  re-applies the new `zoom` to the whole subtree (offsets *and* heights), so a
+  zoom change while scrolled deep does not strand stale offsets.
 - **`getBoundingClientRect` vs `offsetHeight`:** §3.0 measured `offsetHeight` as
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentA-asaf:
Fixed in 24c7bd50 — you're right, that was a stale doc contradiction. §5's 'zoom mid-scroll' edge case and the §8 file table still referenced a `zoomFactor` `createEffect` re-measure that §4.1 concluded is unnecessary (normalized CSS-px measurements are zoom-invariant; the browser re-applies `zoom` to offsets and heights identically). Both now align with §4.1/§7, and the collapse/pin re-measure is correctly attributed to Phase 2. No code change — the implementation already has no such effect.

a5af:
@codex review

a5af:
@codex review

PR #1233, opened 2026-06-02
fix(agent-pane): observe every virtualized row (data-index measure race) — kills residual overlap

AgentA-asaf:
## Summary

Follow-up to #1231. The agent-pane virtualization overlap (thinking text overlaying neighboring messages) had a **second, dominant cause** that #1231's zoom normalization did not address — the user reported *"still overlaps when zoomed"* after #1231 merged.

**Live CDP inspection of the running virtualizer** (not a synthetic probe) found a subset of on-screen rows **permanently stuck at their `estimateSize` (32px)** while rendering at a different real height — producing gaps and overlaps at **any** zoom, including 100%:

| row | virtualizer size | real `offsetHeight` | result |
|---|---|---|---|
| measured rows | = `offsetHeight` ✓ | — | flush (#1231 working) |
| 6 | **32** (estimate) | 46 | **−9.2px overlap** |
| 3, 8, 9, 10 | **32** (estimate) | 23/24 | gaps/overlap |

## Root cause

TanStack's `measureElement(node)` reads `node.getAttribute("data-index")` **first** and **returns early — before `observer.observe(node)`** — when it is `null`. `data-index` was bound reactively (`data-index={props.dataIndex}`), and SolidJS sets reactive attributes in a **render-effect that races the `ref` callback**. When the ref (`virtualizer.measureElement`) wins, `data-index` is still `null`, so the row is **never registered with the `ResizeObserver`** and never re-measured — frozen at the estimate. A race, so only some rows are hit, persistently.

## Fix

Set `data-index` **synchronously inside the ref**, on the line before `measureElement` reads it, so it is never `null` at measure time. The reactive binding is retained for index shifts.

```ts
ref={(el) => {
    el.setAttribute("data-index", String(virtualItem.index));
    virtualizer.measureElement(el);
}}
dataIndex={virtualItem.index}
```

## Diagnostics ruling out alternatives

- A **±1px scroll nudge** did not rescue the stuck rows → not a mid-scroll transient.
- A forced **container width-change** — which fires the `ResizeObserver` on *every observed row* — also did not re-measure them, **proving they were never observed**.

## Verification (live CDP)

- **Before:** rows 3/6/8/9/10 stuck at 32px at zoom 0.63, `−9.2px` overlap at row 6.
- **After:** a 50-row virtualized region at the same zoom reports **0 stuck rows**.

## Spec

`docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_ZOOM_OVERLAP_2026_06_01.md` §3.4 / §4.4 / Phase 1.5.

### DIFF
--- a/.changesets/1780371938-fix-agent-pane-set-data-index-before-measureelement-so-all-r-9v81.md
+++ b/.changesets/1780371938-fix-agent-pane-set-data-index-before-measureelement-so-all-r-9v81.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): set data-index before measureElement so all rows are observed (fixes virtualization overlap)
--- a/docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_ZOOM_OVERLAP_2026_06_01.md
+++ b/docs/specs/SPEC_AGENT_PANE_VIRTUALIZATION_ZOOM_OVERLAP_2026_06_01.md
@@ -1,9 +1,19 @@
 # Agent-Pane Virtualization Overlap Under Zoom
 
-**Status:** Root cause empirically verified (§3.0); Phase 1 implementing
+**Status:** Phase 1 (zoom) MERGED #1231; **Phase 1.5 (data-index measure race) is the dominant residual cause — §3.4 / §4.4**, fix verified live
 **Date:** 2026-06-01
 **Author:** AgentA
 **Tracking:** open
+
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

a5af (approved):
Clean from both bots (codex 'Swish', reagent no findings). Root cause fix verified live via CDP (0 stuck rows post-fix). Approving per author request.

PR #1234, opened 2026-06-02
perf(agent-pane): lazy-mount collapsed tool overlays — 10× cheaper zoom/scroll relayout

AgentA-asaf:
## Summary

The user reported **slow zooming** in agent panes with many tool calls. Root cause: collapsed `ToolBlock`s keep their full `<ToolBlockOverlay>` body in the DOM, hidden only via `max-height: 0; overflow: hidden`. That clips the content *visually* but the browser **still lays out every descendant** — so every collapsed tool is full-cost layout dead-weight, paid on any full-subtree invalidation. CSS `zoom` (per-pane) invalidates the whole `.agent-view` subtree, making it most visible.

## Measured live (CDP) — 56-row pane, 32 collapsed tools

| | before |
|---|---|
| hidden DOM nodes (laid out) | **1,220** |
| hidden text | **~746 KB** |
| **zoom-reflow median** | **299 ms** (max 434 ms) |
| same with hidden bodies `display:none` | 29.5 ms |

→ the clipped-but-laid-out overlays were **~90% of the zoom cost** (~10×).

## Fix

Keep the lightweight panel **container** mounted (so the 120 ms `max-height` collapse transition still animates), but gate the heavy overlay child behind a `mountOverlay()` signal — mounted only while **expanded or animating closed**:

- **History tools** load already-collapsed (`expanded()===false` at init) → overlay **never mounts** → cost gone, 1,220 fewer nodes.
- **Live auto-collapse** (running → 3 s hold → collapse) keeps the overlay for `COLLAPSE_UNMOUNT_MS` (160 ms, > the 120 ms transition) so the close animation plays, then drops it.

The effect subscribes to `expanded()` only and reads `mountOverlay` via `untrack` (no self-subscription — the self-loop class already documented in this file for `postCompletionHold`).

## Verified live after the fix

- collapsed tools: **0 hidden DOM nodes** (was 1,220)
- **zoom-reflow median: 29.2 ms** (was 299 ms) — at the `display:none` floor
- expand → overlay mounts (content shows); **streaming-buffer collapse holds the overlay through the full ~160 ms transition** then unmounts (animation plays)
- virtualized history tools unmount slightly early on *manual* unpin (the virtualizer remounts the row on height change) — acceptable; they don't auto-animate
- **no virtualization-overlap regression** (0 stuck rows / 0 overlaps across zoom 0.5–2.0)

## Spec

`docs/specs/SPEC_TOOL_BLOCK_COLLAPSED_OVERLAY_LAYOUT_2026_06_02.md`

### DIFF
--- a/.changesets/1780391864-perf-agent-pane-lazy-mount-collapsed-tool-overlays-to-cut-zo-ro4j.md
+++ b/.changesets/1780391864-perf-agent-pane-lazy-mount-collapsed-tool-overlays-to-cut-zo-ro4j.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-perf(agent-pane): lazy-mount collapsed tool overlays to cut zoom/scroll relayout cost
+perf(agent-pane): skip layout of collapsed tool overlays via content-visibility (cuts zoom/scroll relayout ~10x); guard virtualizer against undefined virtual item
--- a/.changesets/1780391864-perf-agent-pane-lazy-mount-collapsed-tool-overlays-to-cut-zo-ro4j.md
+++ b/.changesets/1780391864-perf-agent-pane-lazy-mount-collapsed-tool-overlays-to-cut-zo-ro4j.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+perf(agent-pane): lazy-mount collapsed tool overlays to cut zoom/scroll relayout cost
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #1241, opened 2026-06-02
perf(linux): default to XWayland (X11 ozone) — 5–8× fewer frame stalls

AgentU-asaf:
## Summary

Defaults Linux to `--ozone-platform=x11` (XWayland) in `AgentMuxApp::on_before_command_line_processing`. Native-Wayland Ozone in CEF 146 / Chromium 146 has broken Mutter↔Chromium GPU buffer negotiation, which manifests as up-to-**8-second** typing stalls on this distro.

Opt-out via `AGENTMUX_OZONE_PLATFORM=wayland` env var for regression testing.

## The bug

Chromium emits `WaylandZwpLinuxDmabuf::OnTrancheFlags Not implemented` at startup, then responds `LayerTreeHostImpl::DidNotProduceFrame` to ~**89 %** of Mutter's `BeginFrame` requests. The renderer's `requestAnimationFrame` callbacks (including predictive local echo's render path, #1223) are gated on those frames, so typing visibly hangs and pumps out on key release.

## Empirical impact

Measured on the same host (`scripts/capture-trace-ipv6.cjs` + CDP `Profiler.start`, 10 panes, sustained held key):

| | Wayland (native) | XWayland (this PR) |
|---|---|---|
| rAF firing rate | 2.5 Hz | **6.4 Hz** (2.5×) |
| rAF gap p50 | 138 ms | 136 ms (~same) |
| rAF gap p95 | 1182 ms | **224 ms** (5×) |
| rAF gap max | 8280 ms | **1024 ms** (8×) |

p50 is unchanged — that's the residual per-frame Blink/CC compositor cost, addressed separately. p95 and worst-case drop **5×** and **8×**: the pathological "hold a key, nothing happens, release, dump" pattern is gone.

VSCode on the same machine (Electron 39 / Chromium 142) defaults to the XWayland path and runs smoothly here — this PR brings the AgentMux runtime onto the same well-trodden path until native Wayland is fixed upstream.

## What this PR does NOT fix

- **136 ms p50 rAF gap remains** — per-frame Blink CC compositor work. Layer-tree audit follow-up needed.
- **Mouse responsiveness gap vs VSCode** — same compositor path; bounded by the residual p50.
- **CEF 148 native-Wayland fix** — source bump landed in #1221, but the patched Linux libcef.so isn't bundled yet. Once that's in place, native Wayland may behave better and we can revisit.

## Test plan

[...]

PR #1242, opened 2026-06-02
fix(term): predictive-echo — short stall cooldown so slow rAF doesn't lock predictions out for 1.2 s

AgentU-asaf:
## Summary

Splits the predictive-echo cooldown into two: divergence (the existing 1200 ms `cooldownMs`) and **stall** (new 100 ms `stallCooldownMs`). Sweep timeouts use the short stall cooldown instead of the long divergence one.

Fixes the Linux user-visible symptom: hold a key, see ~10 chars, then ~1 s of nothing, then a huge burst. Holds even with `--ozone-platform=x11` (#1241) because rAF *occasionally* still stalls past 600 ms on this Mutter/Chromium GPU handoff.

## The bug

The state machine in `predictive-echo.ts` had two rollback paths sharing one cooldown:

- **`reconcile()` rollback**: PTY echoed bytes that didn't match the prediction — a real divergence. Penalising with the full `cooldownMs` (1200 ms) makes sense; it rides out a mode change without thrash.
- **`sweep()` rollback**: no echo arrived within `predictTimeoutMs` (600 ms). **Not** divergence — just a slow echo. Penalising this with the same 1200 ms was wrong.

The Linux "10 chars, ~1 s nothing, burst" pattern is exactly the second case looping:

1. User holds key → ~12 chars predicted+painted in first 600 ms (~20 Hz × 50 ms).
2. Next rAF stalls past 600 ms (we measure occasional gaps near 1 s on Linux even after #1241).
3. `sweep()` → `rollback()` erases painted chars + `enterCooldown()` sets a **1200 ms** dead zone.
4. For next 1.2 s, every keystroke → `observe()` not `paint()` → nothing visible, chars accumulate invisibly.
5. Cooldown ends → re-arms → backlog pours out at once.

`predictTimeoutMs` ≈ key-repeat × ~12 chars is what gives the "10 chars" feel; `cooldownMs` is what gives the "~1 s of nothing"; the re-arm + accumulated authoritative echoes give the "burst."

## The fix

Two cooldowns instead of one:

```ts
this.cooldown      = opts.cooldownMs      ?? 1200;  // divergence: unchanged
this.stallCooldown = opts.stallCooldownMs ?? 100;   // sweep timeout: new
```

`sweep()` now calls `enterStallCooldown(now)`, `reconcile()` still calls `enterCooldown(now)`. The next rAF cycle resumes painting almost immediately. Tunable per-platform via the `term:predictiveecho:stallcooldownms` setting if a future runtime needs a different value.

## Test plan

- [x] All 17 existing `predictive-echo.test.ts` tests pass
- [x] New test specifically covers the stall path (sweep rollback re-paints within ~150 ms, not stuck for ~1.2 s)
- [ ] Live Linux confirmation: hold-key burst no longer pauses for ~1 s; predictions resume within ~100 ms of the rAF unblocking
- [ ] Divergence cooldown unaffected: actual byte-mismatch rollback still suppresses for 1200 ms (the test at "rolls back and cools down on divergence" covers this)
- [ ] No regression on macOS/Windows where rAF rarely stalls past 600 ms — the new code path simply doesn't fire there

### DIFF
--- a/.changesets/1780425456-fix-predict-echo-short-stall-cooldown-q2xm.md
+++ b/.changesets/1780425456-fix-predict-echo-short-stall-cooldown-q2xm.md
@@ -1,0 +1,50 @@
+---
+type: patch
+---
+
+fix(term): predictive-echo — short stall cooldown so a slow rAF doesn't lock predictions out for 1.2 s
+
+The user-visible symptom on Linux: type a sustained burst, see ~10
+chars, then ~1 s of nothing, then a huge burst of all the accumulated
+chars. Holds even with `--ozone-platform=x11` (PR #1241) because rAF
+*occasionally* still stalls past 600 ms on broken Mutter/Chromium GPU
+handoff.
+
+The state machine in `predictive-echo.ts` had two rollback paths
+sharing one cooldown:
+
+- **`reconcile()` rollback** (line ~182): PTY echoed bytes that didn't
+  match the prediction — a real divergence. Penalising with the full
+  `cooldownMs` (1200 ms default) makes sense; it rides out a mode
+  change without thrash.
+- **`sweep()` rollback** (line ~197): no echo arrived within
+  `predictTimeoutMs` (600 ms default) — *not* divergence, just a slow
+  echo. Penalising this with the same 1200 ms was wrong.
+
+The Linux symptom is exactly the second case looping:
+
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

AgentU-asaf:
@codex review

AgentU-asaf:
@codex review

PR #1248, opened 2026-06-03
fix(gemini): stop Claude-model leak + map turn-end to session_end (2 of 3)

AgentA-asaf:
Fixes **2 of the 3** gemini provider issues. Issue 1 — an auth-escalation loop — is the **live blocker** and is bigger; it's tracked in the analysis doc for a separate fix.

## Issue 2 — `--model sonnet` leak (`buildRuntimeArgs.ts`)
gemini was in `supportsModel`, so it received `--model sonnet` — a Claude model gemini rejects. Removed gemini from `supportsModel`; it now uses its CLI default. (Per-provider gemini model selection is a follow-up.)

## Issue 3 — translator drops the turn boundary (`gemini-translator.ts`)
`result` (gemini's turn-end: `{"type":"result","status":"success","stats":{…}}`) was returned as `[]` (a lifecycle event), so the reducer never received `session_end` → the working spinner never stopped. Now mapped to `session_end` with stats — same class as the merged codex fix.

## Not in this PR — issue 1 (live blocker)
The gemini pane "does nothing / never logs in" because the stored OAuth token is expired (~Mar 25) and the auth orchestration loops on `CheckCliAuth` without ever escalating to `gemini auth login`. That's a `cli_handlers` auth-flow fix (bigger) — see `docs/analysis/GEMINI_PROVIDER_ISSUES_2026_06_02.md`. Until it lands, an expired-token gemini pane needs a manual `gemini auth login`.

## Tests
- `gemini-translator.test.ts` — `result` → `session_end`; `init` still dropped; content unchanged.
- `buildRuntimeArgs.test.ts` — gemini no longer gets `--model`/`sonnet`.
- 11/11 green.

Note: gemini's exact `result.stats` field names are unverified (the auth blocker prevents live testing) — the fix reads `input_tokens`/`output_tokens`; the spinner-stop (the primary fix) works regardless of the stats shape.

### DIFF
--- a/.changesets/1780487171-fix-gemini-stop-claude-model-leak-map-turn-end-to-session-en-4w34.md
+++ b/.changesets/1780487171-fix-gemini-stop-claude-model-leak-map-turn-end-to-session-en-4w34.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(gemini): stop Claude model leak + map turn-end to session_end
--- a/docs/analysis/GEMINI_PROVIDER_ISSUES_2026_06_02.md
+++ b/docs/analysis/GEMINI_PROVIDER_ISSUES_2026_06_02.md
@@ -1,0 +1,100 @@
+# Gemini provider — auth loop + two latent codex-class bugs
+
+**Date:** 2026-06-02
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #1251, opened 2026-06-03
fix(auth): deliver Claude OAuth code reliably to the CLI

a5af:
## Problem
Logging into Claude from an agent pane stalled: after completing login in the browser and coming back to AgentMux, the pasted code never reached the host, the CLI sat waiting on stdin, and the login timed out.

## Root cause
Two independent issues:
1. **Paste landed in the wrong field.** When the auth-URL box appeared, keyboard focus stayed on the main agent message input (`textarea.agent-input`). The user's paste went there instead of the code box, so `handleSubmitCode` saw an empty field and returned early — `set_provider_auth` was never called (confirmed: its info-level delivery log was absent across attempts).
2. **OSC-8 escapes doubled the captured URL.** `extract_url` stripped CSI (`ESC[…`) sequences but not OSC-8 hyperlinks (`ESC]8;;…BEL`), which the Claude CLI emits — the URL appeared both in the hyperlink params and as visible text, producing a doubled/broken link.

## Fix
- **Auto-focus the auth-code input** when the box appears (`AgentDocumentView.tsx`), deferred one frame past the pane's focus-reclaim so it wins the race. The paste now lands in the code box.
- **Strip OSC-8 in `extract_url`** (`platform.rs`), discarding the hyperlink sequence and falling back to its URI if the visible link text isn't itself the URL. Validated standalone against the doubling case + 4 regression cases (CSI-only, plain, wrapped, link-text-not-url), all pass.

## Testing
- `extract_url` unit-validated standalone (5/5 cases).
- Built & ran via `task dev` on macOS; host compiles clean with the change.

### DIFF
--- a/.changesets/1780487002-fix-auth-auto-focus-oauth-code-box-strip-osc-8-escapes-so-cl-nv9k.md
+++ b/.changesets/1780487002-fix-auth-auto-focus-oauth-code-box-strip-osc-8-escapes-so-cl-nv9k.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): auto-focus OAuth code box + strip OSC-8 escapes so Claude login code lands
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1253, opened 2026-06-03
fix(agent): status-bar token stats — record claude token usage (cache-inclusive)

AgentA-asaf:
## Problem
The status-bar token indicator stays at `↑0 ↓0` while agents run (seen live on 0.42.0 DEV). System stats (CPU/Mem/Net/Disk) are fine — this is the **token usage** indicator only.

## Root cause
The claude translator's `result` case (`claude-translator.ts`) extracted `cost_usd`/`duration_ms`/`num_turns` but **dropped `usage.{input,output}_tokens`**. So claude's `session_end` carried no tokens → the `recordTurn` fallback in `useAgentStream` got `null` → the global token-usage store was never fed → indicator stuck at zero. (The live `message_start`/`message_delta` path only fires under `--include-partial-messages`, and was itself cache-blind.)

The "reactivity" angle is a red herring — the indicator memo tracks the store leaves via `getTotal()`, so it was never touched.

## Fix
- **claude-translator**: carry input (`input_tokens` + `cache_creation_input_tokens` + `cache_read_input_tokens`) and `output_tokens` from the `result` event into `session_end` stats.
- **useAgentStream**: make the live `message_start` token path cache-inclusive too (was reading only `input_tokens` ≈ 2 for cached prompts).

## Test
+2 `claude-translator` tests (cache-inclusive input sum; omit token fields when usage absent/all-zero). **19/19 pass**; `tsc` clean for the changed files (pre-existing unrelated errors only).

### DIFF
--- a/.changesets/1780499435-fix-agent-status-bar-token-stats-record-claude-usage-cache-i-ceid.md
+++ b/.changesets/1780499435-fix-agent-status-bar-token-stats-record-claude-usage-cache-i-ceid.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): status-bar token stats — record claude usage (cache-inclusive)
--- a/frontend/app/store/agent-pane-state/reducer.test.ts
+++ b/frontend/app/store/agent-pane-state/reducer.test.ts
@@ -124,6 +124,21 @@ describe("agent-pane-state reducer", () => {
             });
         });
 
+        it("TurnEnd prefers token-bearing result totals over live last-message tokens", () => {
+            // Live turnTokens hold only the last message_start/message_delta
+            // (TokensIn/TokensOut overwrite); the result carries the
+            // cache-inclusive turn total, which must win.
+            const s0 = ready(100);
+            const s1 = update(s0, { type: "TurnStart", at: 110 }).state;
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Addressed the P2 in `2f3b222f`: flipped the precedence in `finalizeTurn` to prefer the result event's turn-total usage over the live `turnTokens` (which overwrite via TokensIn/TokensOut, so they only hold the last `message_start`/`message_delta`). Live tokens are now the fallback, used only when `session_end` carries no usage.

a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Both P2s addressed:
- **@236 (per-pane mergeStats)** — fixed in `84dc03f4`: `mergeStats` now prefers the token-bearing result totals over the live `turnTokens`, using `||` (not `??`) so a `{0,0}` stats line still falls back to the live totals (covered by the existing test). Added a reducer test for the multi-call case.
- **@514 (global aggregate)** — already addressed in `2f3b222f`: `finalizeTurn` now prefers the result total (`statsTokens ?? liveTokens`); the re-post re-anchored on the `TokensIn` source line.

a5af:
@codex review

PR #1254, opened 2026-06-03
fix(agent-pane): HMR-safe useProcessCount RPC guard (DEV crash)

AgentA-asaf:
## Problem
The block-error-boundary trips in DEV with `TypeError: Cannot read properties of undefined (reading 'rpcCall')` at `useProcessCount.ts` → `AgentProcessListCommand`, on agent-pane re-mount (e.g. after a tab switch). Recovers on reload.

## Root cause
`useProcessCount` fires `RpcApi.AgentProcessListCommand(TabRpcClient, …)` **synchronously in `onMount`**. Under a Vite HMR module re-eval, `rpc-util.ts` re-runs `let TabRpcClient` (→ `undefined`) without `initWshrpc()` re-binding it. The undefined client throws synchronously during argument evaluation — before the Promise exists — so the existing `.catch()` can't catch it, and it escapes to the error boundary. **DEV/HMR-only**; production builds have no HMR, so this can't happen there. Worth fixing as the unguarded `onMount` RPC is a latent fragility.

## Fix
Guard `TabRpcClient` before the call; when unbound, seed from buffered deltas only (the same fallback `.catch()` already provides). Kept as `if/else` (not early `return`) so `onCleanup` still registers — an early return would leak the two WPS subscriptions.

### DIFF
--- a/.changesets/1780499728-fix-agent-pane-hmr-safe-useprocesscount-rpc-guard-buw5.md
+++ b/.changesets/1780499728-fix-agent-pane-hmr-safe-useprocesscount-rpc-guard-buw5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): HMR-safe useProcessCount RPC guard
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1255, opened 2026-06-03
fix(build): guard CEF bundle version against linked cef crate; loud host startup failure

AgentX-asaf:
## Problem

A `task package` portable (v0.42.0) showed a **splash, then no window** on every launch. `runtime/debug.log` had only:

```
ERROR cef\libcef_dll\libcef_dll2.cc:92] Request for unsupported CEF API version 14800
```

**Root cause:** the `cef` crate (`features = ["build-util"]`) **stages a CEF binary SDK into `target/cef-sdk/` and reuses it across crate-version bumps** — the staging is not version-keyed. The 146→148 crate bump (#1221) left the March CEF‑146 staging in place, so every build bundled a **146 `libcef.dll` into a 148‑linked host** → API mismatch → silent splash‑then‑exit. `Cargo.lock` is consistent at 148 (both `cef` and `cef-dll-sys` = `148.3.0+148.0.9`); there is no version skew there and no `[patch.crates-io]`.

## Changes

- **`scripts/verify-cef-version.sh`** — asserts the bundled `libcef.dll` MAJOR matches the linked `cef` crate (read from `Cargo.lock`). `bundle:windows` now calls it and **fails the build on mismatch** instead of shipping a broken portable.
- **`clean:cef`** — now removes `target/cef-sdk/` (the build‑util staging, the real culprit) plus the downstream `cef-dll-sys` build dirs and `target/release` copies, so the next `task build:host` re‑downloads the CEF matching the linked crate.
- **`agentmux-cef`** — on CEF init failure, shows a `MessageBoxW` with an actionable message (points to `cef-debug.log`, suggests `task clean:cef && task build:host`) instead of vanishing after the splash.

## Verification

- Guard fires on the broken state (146 vs 148 → exit 1, clear message) and passes once corrected (148 vs 148 → exit 0).
- `clean:cef` + `task build:host` corrected `target/release/libcef.dll` to **148.0.9 / chromium 148.0.7778.180**.
- `task package` ran the guard inside `bundle:windows` (passed) and produced a working v0.42.0 portable whose `runtime/libcef.dll` is **148** — the "splash then nothing" build is fixed.

## Spec

`docs/specs/SPEC_WINDOWS_CEF_BUNDLE_VERSION_INTEGRITY_2026_06_03.md`

Follow-ups noted in the spec: a version-keyed staging marker so `build:host` re-stages automatically (removing the manual `clean:cef` step); a negative regression test; and applying the guard to `bundle:darwin` / the Linux resolver.

### DIFF
--- a/.changesets/1780499068-fix-build-guard-bundled-cef-version-against-linked-cef-crate-id3l.md
+++ b/.changesets/1780499068-fix-build-guard-bundled-cef-version-against-linked-cef-crate-id3l.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): guard bundled CEF version against linked cef crate; surface a user-facing dialog on CEF init failure instead of a silent splash
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -938,11 +938,13 @@ tasks:
               ignore_error: true
             # Drop cef-dll-sys build dirs so cargo re-runs its build script and
             # re-copies the correct CEF runtime into target/ on the next build.
-            - cmd: '{{.RMRF}} "target/release/build/cef-dll-sys-*"'
-              ignore_error: true
-            - cmd: '{{.RMRF}} "target/fast-release/build/cef-dll-sys-*"'
-              ignore_error: true
-            - cmd: '{{.RMRF}} "target/debug/build/cef-dll-sys-*"'
+            # Use `bash -c` so the globs expand on EVERY OS: {{.RMRF}} double-
+            # quotes its path, and POSIX `rm -rf` does not glob inside quotes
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentX-asaf:
Addressed the P2 in ac1044e1: routed the `cef-dll-sys-*` removals through `bash -c 'rm -rf …'` so the globs expand on every OS (the quoted `{{.RMRF}}` form only globbed under Windows PowerShell). Verified: quoted glob = no-match; `bash -c` unquoted = expands to the real `cef-dll-sys-<hash>` dirs. Spec snippet updated to match.

PR #1256, opened 2026-06-03
feat(launcher): multi-instance isolation telemetry, per-instance splash class, invariants

AgentX-asaf:
## Context

A host instance died shortly after a freshly-built v0.42.0 portable was launched on the same machine, and the launch was (reasonably) suspected of killing it. A launcher + data-dir audit shows **isolation is sound by design**, so a parallel build cannot, by code, crash a running instance:

- single-instance pipe keyed on `hash(data_dir + version)` (`agentmux-launcher/src/hash.rs`)
- **unnamed** Job Objects — no other process can obtain a handle to kill your processes (`main.rs` `create_job_object`)
- version-scoped data dirs (`channels/<ch>/versions/<v>/`)

A v0.42.0 portable (`dev-portable-main-*` channel) and a running v0.41.0 (`stable` channel) therefore share **no** keyed resource. (The actual reason that portable showed splash-then-nothing was an unrelated CEF runtime mismatch — see #1255.)

The gap: nothing **proved** the invariant or would **detect** a real violation. This PR makes the guarantee observable and enforced — it does not change the isolation mechanism.

## Changes

- **`main.rs`** — emit one `instance_claim pid=… version=… data_dir=… dir_hash=… pipe=…` line at startup. A real cross-instance collision becomes greppable (two live PIDs claiming the same `dir_hash`) instead of inferred after a vanished window.
- **`splash.rs`** — key the splash window class per `dir_hash` (`AgentMuxSplash-<hash>`); it was the one remaining *global* `RegisterClassExW` name. Now every named OS object the launcher creates is instance-keyed.
- **`hash.rs`** — `portable_and_installed_never_collide` regression test: the exact 0.42.0-portable vs 0.41.0-installed scenario asserts distinct pipe hashes. (186 launcher tests pass.)
- **`CLAUDE.md`** — document isolation invariants **I1–I6** as a review contract for any diff touching job/pipe/process code.

## Verification

- `cargo test -p agentmux-launcher` → **186 passed**, including the new test.
- `cargo build --release -p agentmux-launcher` → clean.

## Spec

`docs/specs/SPEC_MULTI_INSTANCE_ISOLATION_HARDENING_2026_06_03.md`

Deferred follow-up (in spec §5.1): a full launcher↔launcher *survival* integration test ("launch B ⇒ A stays alive"). Left out of this PR to keep it low-risk; the telemetry + invariants + unit test land the high-confidence pieces first.

### DIFF
--- a/.changesets/1780500267-feat-launcher-log-instance-claim-isolation-telemetry-and-key-3nzp.md
+++ b/.changesets/1780500267-feat-launcher-log-instance-claim-isolation-telemetry-and-key-3nzp.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(launcher): log instance_claim isolation telemetry and key the splash window class per instance; document isolation invariants (I1-I6)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -91,6 +91,33 @@ This means:
 - `task dev` is always safe alongside a running portable instance.
 - **NEVER kill by image name** (`taskkill //im agentmux-cef.exe`) — it kills ALL instances. Always kill by PID.
 
+#### Isolation invariants (I1–I6)
+
+These are the contract that makes parallel instances safe — launching a new build
+must never crash a running one. Any change to the launcher's process/pipe/job code
+must be reviewed against them (see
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

Issue #1235, opened 2026-06-02
Agent-pane virtualizer accumulates measurement drift under expand/collapse churn (overlap/gaps)


AgentA-asaf:
## Summary

The agent-document virtualizer (`AgentDocumentVirtualList`) **accumulates measurement drift** under repeated row-height changes (tool expand/collapse, and likely any dynamic-height settle): cached row sizes diverge from rendered heights, producing overlaps/gaps that **do not self-correct**. It affects **non-tool rows too** (markdown, user-message), so it's a virtualizer-level issue, not specific to tool blocks.

This is **pre-existing on `main`** and distinct from:
- #1233 (data-index/`measureElement` race — fixed the *initial-load* stuck-estimate case)
- #1234 (collapsed-tool layout cost — perf; its `virtualItem` guard stops a *crash*, not this drift)

## Reproduction (live, deterministic)

Fresh reload of an agent pane with a long history, scroll to ~40%, then drive expand/collapse cycles on visible tool rows. Measured via CDP (virtualizer row size = `translateY` delta vs `offsetHeight`; overlap = rendered rows whose tops cross):

| state | tool mismatches | non-tool mismatches | overlaps |
|---|---|---|---|
| **fresh reload (baseline)** | 0 | 0 | 0 |
| **after 9 expand/collapse cycles** | 2 | **3** | 1 |

Reproduces with `content-visibility` (from #1234) **toggled off**, confirming it's independent of that change. Rapid/heavy churn accumulates more (observed 15–20 mismatched rows and multiple −40px+ overlaps in earlier runs), e.g. a markdown row cached at `96` (real `29`), a user-message at `32` (real `87`), a section at `264` (real `313`).

## CDP diagnostic recipe

```js
// per .agent-document-row: ty = DOMMatrixReadOnly(transform).m42
// virtualizer-size = ty[i+1]-ty[i]; mismatch if |that - offsetHeight| > 2
// overlap = sort by rendered top; next.top - cur.bottom < -1.5
```
Dev CDP on `:9223`, page target = the workspace tab; eval against `.agent-document-virtualizer`.

## Hypotheses to investigate

1. **Cache doesn't reconverge after a dynamic height change.** When a row grows/shrinks, the virtualizer updates that row but neighbors measured during the reflow churn cache transient values and aren't re-measured once settled. TanStack's `shouldMeasureDuringScroll` / re-measure gating may skip them.
[...]

PR #1257, opened 2026-06-03
fix(agent-pane): restore scroll-to-bottom when a pane returns from hidden (codex)

AgentA-asaf:
## Problem
Running 2 claude + 1 codex agents, switching tabs and returning left the **codex pane pinned at a stale scroll position** ("didn't scroll with"); the claude panes were fine. Seen live on 0.42.0 DEV.

## Root cause
Auto-follow (sticky-bottom) only re-fires on a new `nodes()` emission. An inactive tab is `display:none` (0×0), so `scrollTo(MAX)` while hidden is a **no-op**, and nothing re-issues it on reactivation. Claude kept streaming text deltas → re-glued on return. **Codex emits complete items and had gone quiet → no emission after return → stayed at the stale position.** That's why only codex did it.

## Fix
A `ResizeObserver` on the scroll container re-applies the bottom scroll on a `0 → non-zero` `clientHeight` transition (hidden→visible), **only when already sticky** — it never *engages* stick, so a user who scrolled up stays put. Deterministic (no timers).

## Verification
Typecheck clean; hot-reload didn't trip the error boundary. The full tab-switch behaviour is best confirmed live (hard to smoke deterministically).

Note: the structural fix for the underlying "unmeasured while hidden" virtualizer class is the layout-reducer work (#1235); this is the targeted fix for the missing re-trigger.

### DIFF
--- a/.changesets/1780500347-fix-agent-pane-restore-scroll-to-bottom-when-a-pane-returns--k2em.md
+++ b/.changesets/1780500347-fix-agent-pane-restore-scroll-to-bottom-when-a-pane-returns--k2em.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): restore scroll-to-bottom when a pane returns from hidden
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1259, opened 2026-06-04
fix(macos): survive external accessibility queries (Magnet/Synergy) without crashing

a5af:
## Problem
On macOS-26, clicking the title-bar menu while a window manager / KVM is running (**Magnet**, **Synergy**) crash-looped the host with `EXC_BREAKPOINT`. The launcher respawned it repeatedly (6 generations in one session).

## Root cause (diagnosed from crash reports + instrumented logging)
Not the menu's CSS/position. Those tools query the app's **accessibility tree**, and the read routes through:
```
-[NSApplication accessibilityParent]
  → NSAccessibilityGetObjectForAttributeUsingLegacyAPI
  → NSAccessibilitySetUnsupportedAttributeError
  → +[NSString stringWithFormat:]  → CF string-format trap
```
i.e. **CEF #3512 family** (a CEF web-content AX object is the `%@` arg whose description faults). Any flyout/portal under an external AX query faults the same way — the old left-side hamburger and every context menu included. The trigger attribute (`AXEnhancedUserInterface`) is overloaded: VoiceOver *and* window managers set it ([Firefox bug 1664992](https://bugzilla.mozilla.org/show_bug.cgi?id=1664992)).

## Fix — Phase 1, host-only (no CEF rebuild), **validated** (zero crashes under Magnet+Synergy)
See `docs/specs/SPEC_MACOS_ACCESSIBILITY_ROBUSTNESS_2026-06-03.md` for the full layered design.

- **Layer 2a (load-bearing):** guard `-[NSApplication accessibilityParent] → nil`. NSApplication is the AX root — its parent is legitimately nil — so returning nil directly short-circuits **before** the crashy legacy bridge runs. Not a dodge: windows/title AX still answer.
- **Layer 1 (defense in depth):** govern AX activation by swizzling `accessibilitySetValue:forAttribute:` — gate `AXEnhancedUserInterface` (the window-manager path) while honoring `AXManualAccessibility`. `AGENTMUX_A11Y_HONOR_ENHANCED=1` overrides.
- Heavy logging on the AX path so the activation route stays observable in the field.

Deliberately **not** `--disable-renderer-accessibility` (the blanket dodge) — the explicit (`AXManualAccessibility`) enable path stays open.

## Scope / follow-ups
Full VoiceOver support — making the web-content AX tree itself non-fatal (CEF source patch / Chromium bump) — is **Phase 2**, tracked in the spec, along with the launcher crash-loop breaker (L3) and a deterministic AX repro harness.

## Testing
- Reproduced the crash deterministically with Magnet+Synergy + menu clicks (crash reports `agentmux-cef-2026-06-03-{125025,143426}.ips`, identical `accessibilityParent` stacks).
- With the guard: **0 crashes** across repeated menu/submenu/context-menu hammering. Host pid stable, no respawns.
- Compiles clean; macOS-gated (`#[cfg(target_os = "macos")]`), no effect on Windows/Linux.

### DIFF
--- a/.changesets/1780539826-fix-macos-survive-external-accessibility-queries-magnet-syne-iltd.md
+++ b/.changesets/1780539826-fix-macos-survive-external-accessibility-queries-magnet-syne-iltd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): survive external accessibility queries (Magnet/Synergy) without crashing
--- a/agentmux-cef/src/main.rs
+++ b/agentmux-cef/src/main.rs
@@ -782,6 +782,13 @@ fn main() {
     unsafe {
         set_macos_activation_policy_regular();
         set_macos_dock_icon();
+        // Layer 1 of SPEC_MACOS_ACCESSIBILITY_ROBUSTNESS_2026-06-03: govern
+        // Chromium's macOS accessibility activation so a window manager / KVM
+        // poking the AX tree (Magnet, Synergy, …) can't force the crash-prone
+        // web-content AX mode (CEF #3512). Must run after cef::initialize so
+        // the CEF NSApplication subclass (which owns the legacy AX setter)
+        // exists.
+        install_macos_accessibility_governor();
     }
 
     // Start memory heartbeat — logs system/process memory stats every 20s.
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1260, opened 2026-06-04
fix(cef): pin Cargo to CEF 146 until 7778 drag/right-click/transparency patches are forward-ported

AgentU-asaf:
## Summary

PR #1221 (CEF 146 → 148 source bump) inadvertently broke `build:host:linux` because the agentmux patches that add `_cef_window_t::begin_window_drag` (the source code addition behind `StartWindowDragTask` for native Wayland/X11 window drag) **only exist on the `agentmuxai/cef@agentmux/7680-drag-rightclick-and-transparency` branch** — they have not been forward-ported to the `agentmux/7778-…` branch yet.

## What happened

- `cargo build --release -p agentmux-cef --features patched-libcef` fails at `ui_tasks.rs:215` with `error[E0609]: no field begin_window_drag on type _cef_window_t` (the cef-dll-sys 148 binding lacks the field).
- The Linux Taskfile path has been silently building **without** `--features patched-libcef` since the bump, which compiles the `#[cfg(not(feature = "patched-libcef"))]` no-op branch. Every `start_window_drag` IPC since then warns "patched-libcef feature disabled — native drag is a no-op."
- **Title-bar drag has silently been broken on every Linux AppImage built since #1221 was merged**, regardless of ozone platform.

## What this PR does

Pins `cef = "146"` in `agentmux-cef/Cargo.toml` + refreshes `Cargo.lock` accordingly. With this pin, the patched cef-dll-sys 146 binding (which has `begin_window_drag`) is used, `--features patched-libcef` compiles, and the existing patched `libcef.so` (built from `agentmuxai/cef@agentmux/7680-drag-rightclick-and-transparency` via `scripts/resolve-cef-runtime.sh`) is the matching runtime.

**Explicitly temporary.** The proper fix is two follow-up PRs:

1. New branch `agentmuxai/cef@agentmux/7778-drag-rightclick-and-transparency` cherry-picking these commits onto upstream `7778`:
   - `af485ed2` `views: Add CefWindow::BeginWindowDrag()`
   - `010f616f` `views: PR review on BeginWindowDrag` (the X11 fix using `display::Screen::Get()->GetCursorScreenPoint()` — without this `_NET_WM_MOVERESIZE` ignores the drag request because `gfx::Point()` gives an invalid anchor)
   - `130af663` `views: Annotate CefWindow::BeginWindowDrag with added=14600`
   - `41802fe6` `Patch A: Let right-clicks on HTCAPTION fall through to renderer`
[...]

PR #1261, opened 2026-06-04
perf(linux): default to XWayland (X11 ozone) — 5–8× fewer frame stalls (supersedes #1241)

AgentU-asaf:
> **Supersedes #1241**, rebased onto current main + addressing the reagent review comments. Pairs with #1260 which restores the broken `--features patched-libcef` build so the title-bar drag still works under the X11 ozone path (X11 fix lives in `agentmuxai/cef@010f616f`).

## Summary

Defaults Linux to `--ozone-platform=x11` (XWayland) in `AgentMuxApp::on_before_command_line_processing`. Opt-out via `AGENTMUX_OZONE_PLATFORM=wayland` env var.

## The bug

Native-Wayland Ozone in CEF 146 / Chromium 146 has broken Mutter↔Chromium GPU buffer negotiation. Chromium emits `WaylandZwpLinuxDmabuf::OnTrancheFlags Not implemented` at startup, then the `cc::Scheduler` replies `LayerTreeHostImpl::DidNotProduceFrame` to ~**89 %** of Mutter's `BeginFrame` requests. With `BeginMainFrame` production stuck near sysinfo's ~1 Hz status-bar invalidation cadence, the renderer's `requestAnimationFrame` callbacks (including predictive local echo's render path, #1223) fire only when sysinfo dirties Blink — typing visibly hangs and pumps out on key release.

## Empirical impact

Measured locally (`scripts/capture-trace.cjs` + CDP `Profiler.start`, 10 panes, sustained held key):

| | Wayland (native) | XWayland (this PR) |
|---|---|---|
| rAF firing rate | 2.5 Hz | **6.4 Hz** (2.5×) |
| rAF gap p50 | 138 ms | 136 ms (~same) |
| rAF gap p95 | 1182 ms | **224 ms** (5×) |
| rAF gap max | 8280 ms | **1024 ms** (8×) |

p50 is unchanged — that's residual per-frame Blink/CC compositor work, addressed separately. p95 and worst-case drop **5×** and **8×**: the pathological "hold a key, nothing happens, release, dump" pattern is gone.

VSCode (Electron 39 / Chromium 142) on the same machine defaults to the same XWayland path and runs smoothly here.

## Drag

The X11 `BeginWindowDrag` path (`agentmuxai/cef@010f616f`, the `display::Screen::Get()->GetCursorScreenPoint()` fix) is already in the patched `libcef.so` we ship — without it `_NET_WM_MOVERESIZE` ignores the request because `gfx::Point()` gives an invalid anchor. Verified locally: title-bar drag works on both Wayland and XWayland ozone modes with this PR applied **on top of #1260**.

## Changes from #1241

Addressing the reagent review:

1. Removed the dangling `docs/specs/SPEC_LINUX_X11_OZONE_DEFAULT_2026_06_02.md` reference (the spec was never committed).
2. Reconciled the inline comment in `app.rs` — clarified that the rAF gating is the result of the `BeginMainFrame`-production stall (which then **fires** at the sysinfo cadence because sysinfo is the only Blink-side invalidation source), not "gated on sysinfo" as a direct cause. Mechanism is the `DidNotProduceFrame` storm; sysinfo is the only thing that flips the answer to "yes."

## Test plan

- [x] `task build:host` succeeds with `#1260` applied (CEF 146 pin + patched-libcef)
- [x] Subprocess `cmdline` shows `--ozone-platform=x11` by default
- [x] `AGENTMUX_OZONE_PLATFORM=wayland` opt-out: `cmdline` shows `--ozone-platform=wayland`
- [x] Title-bar drag works under XWayland (the `010f616f` cursor-screen-point fix)
- [x] Title-bar drag works under Wayland (regression check via opt-out env var)
- [ ] `scripts/capture-trace.cjs` confirms `DidNotProduceFrame` rate drops + rAF gap p95 < 250 ms after CI build
- [ ] No window-creation regressions on a GNOME/Wayland session under XWayland

### DIFF
--- a/.changesets/1780422623-perf-linux-default-to-xwayland-x11-ozone-e42g.md
+++ b/.changesets/1780422623-perf-linux-default-to-xwayland-x11-ozone-e42g.md
@@ -1,0 +1,48 @@
+---
+type: patch
+---
+
+perf(linux): default to XWayland (X11 ozone) — 5–8× fewer frame stalls
+
+Linux CEF 146 (Chromium 146) on Mutter (GNOME) has broken native-Wayland
+GPU buffer negotiation — Chromium logs
+`WaylandZwpLinuxDmabuf::OnTrancheFlags Not implemented` at startup and
+then responds `LayerTreeHostImpl::DidNotProduceFrame` to ~89 % of
+Mutter's `BeginFrame` requests. The renderer's `requestAnimationFrame`
+callbacks (including predictive local echo's render path, #1223) are
+gated on those frames, so typing visibly hangs and pumps out on key
+release.
+
+Setting `--ozone-platform=x11` routes the renderer through XWayland's
+X11 present path — the wire-format Linux Chromium has shipped
+reliably for years. Measured locally on the same host
+(`scripts/capture-trace-ipv6.cjs` + CDP `Profiler.start`, 10 panes,
+sustained held key):
+
+|                 | Wayland (native) | XWayland (this PR) |
+| --------------- | ---------------- | ------------------ |
+| rAF firing rate | 2.5 Hz           | **6.4 Hz**         |
+| rAF gap p50     | 138 ms           | 136 ms             |
+| rAF gap p95     | 1182 ms          | **224 ms**         |
+| rAF gap max     | 8280 ms          | **1024 ms**        |
+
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1262, opened 2026-06-04
fix(window): collapse widget labels before tabs clip, shrink tabs when crowded

AgentA-asaf:
## What

In a narrow window the tabs and the widget bar compete for width, with two responsive gaps. With 2 tabs open, the labeled widget bar (~367px) starved the tab strip to ~117px and the trailing tab was clipped *under* the widget buttons.

## Two gaps

**1. Widget labels collapsed at the wrong time.** `action-widgets.tsx` reserved a *fixed* `MIN_TAB_WIDTH = 120` for the tab strip regardless of tab count, so the collapse trigger (`labeled + buttons + 120 > header`) thought 2 tabs fit and never dropped the labels. Now it reserves `tabCount * per-tab floor`, and re-measures on tab add/remove via a `MutationObserver` (adding a tab doesn't resize the header, so the `ResizeObserver` alone wouldn't fire).

**2. Tabs couldn't get thinner.** Tabs were pinned at the `--ws-tab-min: 100px` floor while `.tab-bar-fill` (the draggable title-bar slack) grabbed the leftover, so a crowded strip *clipped* instead of shrinking. Floor lowered to `56px` (Chrome-style) and `.tab-bar-scroll` is now `flex: 1 1 auto` (was `0 1 auto`) so it reclaims the slack from the fill — tabs stay wide when there's room, shrink when crowded.

Cascade: **labels collapse -> tabs shrink -> strip scrolls**, matching Chrome's tab behavior.

## Verification (live, via CDP on a running dev instance)

| | before | after |
|---|---|---|
| widget bar | 367px labeled | 180px icon-only |
| tab strip | 117px | 305px |
| 2 tabs | trailing tab clipped | both visible, ~99px each |

- `tabbar-dnd.test.ts` — 22/22 pass (CSS-only change to the strip)
- `stylelint` clean on `tabbar.scss`
- `tsc` — 0 errors in `action-widgets.tsx` (repo gates on esbuild/Vite, not full tsc)

### DIFF
--- a/.changesets/1780541630-fix-window-collapse-widget-labels-then-shrink-tabs-when-the--k83i.md
+++ b/.changesets/1780541630-fix-window-collapse-widget-labels-then-shrink-tabs-when-the--k83i.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(window): collapse widget labels then shrink tabs when the title bar is crowded
--- a/frontend/app/tab/tabbar.scss
+++ b/frontend/app/tab/tabbar.scss
@@ -13,14 +13,15 @@
     // unrecoverable from the bar today, unlike editor tabs which clip).
     // Spec: specs/SPEC_WORKSPACE_TAB_SIZING_2026-05-27.md.
     // Basis ~ VS Code "fit" mode default (~240px) — enough room for
-    // ~25 chars of file/title before ellipsis. Floor at 100px keeps
-    // a crowded tab readable rather than collapsing into a pill. Cap
+    // ~25 chars of file/title before ellipsis. Floor at 56px is the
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentA-asaf:
Re codex's collapse-threshold note: intentional — keeping the 100px reserve. It implements the spec'd cascade *labels collapse first, then tabs shrink*: labels drop while tabs are still comfortable, then tabs shrink toward the 56px floor. Lowering the reserve to 56 would invert the order (tabs hit the floor first, then labels collapse). Renamed the constant to `TAB_COLLAPSE_RESERVE_PX` + clarified the comment so the trigger width and the shrink floor aren't conflated.

PR #1243, opened 2026-06-03
feat(macos): CEF 148 notarized DMG — patched renderer, launcher splash, DCHECK-off build

a5af:
Delivers a **stable, notarized macOS DMG** on CEF 148 / macOS 26 (Tahoe). This is the follow-on to the merged #1221 — it contains the launcher-owned splash, the launcher as bundle entry point, and the resolution of the pane-drag/window-close crash saga.

## The headline fix: DCHECK-off CEF build
The pane-drag/window-close SIGABRTs were **macOS-26 DCHECKs** in the from-source CEF framework (`ScopedSendingEvent` CrAppControlProtocol conformance on every drag event; `CefShutdownChecker` on exit) — developer-only assertions that **production CEF compiles out**. They were live only because a from-source `is_official_build=false` build defaults `dcheck_always_on=true`. The fix is to build the framework the way CEF ships: **`dcheck_always_on=false`**. Validated at the binary level (DCHECK assertions physically compiled out; framework 567→446 MB) and via a full drag/tear-off/close repro with 0 crashes.

Root cause writeup: `docs/retro/retro-macos26-cef-dcheck-root-cause-2026-06-02.md`. The breakthrough was symbolizing with `llvm-symbolizer` (DWARF) on a UUID-matched framework, after the shared CHECK-abort machinery offsets misled three earlier wrong guesses.

## What's in this PR

**Launcher-owned instant splash (A0)**
- `agentmux-launcher` is the packaged `.app` `CFBundleExecutable` — a tiny binary that paints a native AppKit splash *before* CEF loads, then spawns + supervises srv and the host.
- `splash_mac.rs`: dark rounded card + pulsing brain (matches the Win32 splash), fades out on host first-frame signal. Raw ObjC FFI, no new deps.

**Packaging (`scripts/package-macos.sh`)**
- Launcher as entry point; host nested (its `current_exe`-relative resolution + Dock tile unchanged).
- All six CEF 148 helper apps (generic + GPU/Plugin/Renderer/Alloy + **Alerts** — CEF 148 spawns an alert-service helper; its absence caused `posix_spawnp` loops).
- Bundle `dist/schema` (srv resolves schema from `AGENTMUX_APP_PATH/schema`).
- Strip `.js.map` source maps (~28 MB), strip libcef + dedup GL libs + trim locales, **ULMO** compression → ~140 MB DMG.

**CEF source fix (documented; lives in the fork)**
[...]

PR #1265, opened 2026-06-04
Revert #1260: per-target CEF version is impossible (links-tagged crate), restore main for macOS+Windows

AgentU-asaf:
## Urgent revert — #1260 broke macOS and Windows builds

I missed that the macOS notarized DMG (#1243) and the Windows bundle pipeline (#1221 follow-up at \`bundle:windows\` in Taskfile.yml) are both **already running CEF 148 libcef binaries**. PR #1260's workspace-wide Cargo pin to \`cef = \"146\"\` put the host source on 146 against 148 runtimes → next macOS or Windows build crashes at startup with the same \`unsupported CEF API version 14800\` error that PR #1221 explicitly called out fixing.

## I tried a per-target Cargo dependency first — it doesn't work

\`\`\`toml
[target.'cfg(target_os = \"linux\")'.dependencies]
cef = { version = \"146\", … }
[target.'cfg(not(target_os = \"linux\"))'.dependencies]
cef = { version = \"148\", … }
\`\`\`

\`cargo check\` fails:

> error: Only one package in the dependency graph may specify the same \`links\` value. \`cef-dll-sys\` … failed to select a version for \`cef\` which could resolve this conflict.

Cargo computes the resolver graph globally across all targets in the workspace; with a \`links\`-tagged native dependency you cannot have two versions resolvable, even if mutually exclusive at compile time. Confirmed locally on this branch state — there is no workspace-Cargo-level fix.

## So: straight revert, restore main to the pre-#1260 state

Reverts commit \`50557091\` (the #1260 squash). Cargo goes back to \`cef = \"148\"\`. Linux \`task build:host\` returns to silently dropping \`--features patched-libcef\` because CEF 148's binding still lacks \`begin_window_drag\` → title-bar drag stays broken on Linux until the proper fix lands. macOS + Windows return to working order.

## What's next (the spec the user already asked for)

The proper fix is the **CEF 148 forward-port** I'll now write a spec for:

1. \`agentmuxai/cef@agentmux/7778-drag-rightclick-and-transparency\` — cherry-pick 11 commits from \`agentmux/7680\` (BeginWindowDrag + X11 cursor-screen-point + HTCAPTION right-click + 5 transparency cascade follow-ups) onto upstream \`7778\`.
2. \`cef-dll-sys\` 148 binding fork that adds \`begin_window_drag\` at the end of \`_cef_window_t\` (or a [patch.crates-io] entry pointing at a fork branch).
3. Rebuild a CEF 148 libcef.so for Linux from #1 + bundle it via \`scripts/resolve-cef-runtime.sh\`.
4. Once #1–3 land, Linux \`--features patched-libcef\` compiles + works against the new libcef.so. No source-version pin needed on any platform.

Spec coming in a follow-up PR.

## Test plan

- [x] \`cargo check --release -p agentmux-cef\` passes (no \`--features patched-libcef\`)
- [x] Cargo.lock restored to CEF 148 resolution
- [x] No changes to Taskfile.yml — the silent \`patched-libcef\` drop is restored, matching pre-#1260 main behavior
- [ ] CI macOS build succeeds (no \`unsupported CEF API version\`)
- [ ] CI Windows build succeeds


a5af:
@codex review

PR #1266, opened 2026-06-04
fix(auth): run Claude login under a PTY so the CLI survives to receive the OAuth code

a5af:
## Problem
Logging into Claude from an agent pane always appeared **stuck on the OAuth URL**. The `claude auth login` CLI exited cleanly **~5 s after printing the URL** — long before the user finished browser auth — so the pasted code (and any callback) had nothing to be delivered to. Logs showed `child exited exit_code=Some(0)` 5 s after every `run_cli_login`.

## Root cause
Diagnosed by reproducing the CLI under many conditions. The CLI exits early **specifically when spawned terminal-less by the GUI host**. Standalone runs with a controlling terminal stay alive at the `Paste code here >` prompt indefinitely; only the host's pipe-spawn dies at 5 s. Ruled out individually: stdout draining, the controlling terminal (`setsid` stays alive), and the environment (`TERM`, `CI`, `CLAUDECODE` all identical).

## Fix
- **Mark Claude `requiresLoginTty: true`** (`providers/index.ts`) so `run_cli_login` takes its existing **PTY branch** (the one OpenClaw already uses). Under a real PTY the CLI is fully interactive and **stays alive**, so the login can complete. **Verified**: the PTY-spawned CLI survives indefinitely where the pipe-spawned one died at 5 s.
- **Harden the pipe branch** (used by other providers): keep draining stdout/stderr for the child's lifetime instead of dropping the readers right after the URL is captured, so a CLI's later prompt writes can't hit a closed pipe (EPIPE).

## Scope / note
`claude auth login` uses the **manual paste-code flow** (`redirect_uri=platform.claude.com`) **even in a plain terminal** — the loopback / auto-complete-on-authorize flow is not what this command exposes. So this PR makes the (paste-based) login *work*; true async login is a separate investigation (tracked).

## Testing
- Reproduced the 5 s exit deterministically; confirmed PTY-spawned CLIs stay alive across browser auth + a 40 s wait.
- Frontend type-checks; host compiles. macOS/Unix PTY branch already in use for OpenClaw.

### DIFF
--- a/.changesets/1780568270-fix-auth-run-claude-login-under-a-pty-so-the-cli-survives-to-g1k6.md
+++ b/.changesets/1780568270-fix-auth-run-claude-login-under-a-pty-so-the-cli-survives-to-g1k6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): run Claude login under a PTY so the CLI survives to receive the OAuth code
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -435,45 +435,74 @@ pub async fn run_cli_login(
         *stored_stdin = child.stdin.take().map(CliLoginStdin::Pipe);
     }
 
-    // Capture the OAuth URL from stdout/stderr. The CLI prints it within the
-    // first few hundred ms after spawn. We read until we find "https://..."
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1268, opened 2026-06-04
fix(macos): window close exits the instance instead of hanging hidden

a5af:
## Problem
Clicking the close (red traffic-light) button on the last window made AgentMux **"disappear but stay open hidden"** — the window vanished but the host process kept running with no visible window, and the launcher hung with it (Dock tile lingering, instance never actually quits).

## Root cause
Pinpointed live with a `[shutdown-diag]` trace around each shutdown step. The close → quit path is entirely correct (`LastWindowClosed` → `quit_message_loop` → CEF message loop returns → graceful shutdown). The wedge is the **very last step, `drop(runtime)`**:

```
CEF message loop exited, shutting down
[diag] cef shutdown() returned          ← CEF shutdown fine (~50ms)
   ← then wedged here forever, host still alive
```

Dropping the multi-thread tokio runtime blocks waiting for every background task / blocking thread to finish, and a `spawn_blocking` (a pipe/PTY reader) is parked on a blocking read that never returns. The window is already destroyed (so it "closes"), but the host hangs in the runtime drop and never exits; the launcher (which exits only when it observes the host exit) hangs too.

*(The earlier reverted `process::exit(0)` was placed **after** the never-reached "shutdown complete" log — past the wedge — which is why it never helped.)*

## Fix
On macOS, replace the blocking `drop(runtime)` with **`runtime.shutdown_background()` + `std::process::exit(0)`**, after the graceful CEF shutdown + sidecar kill + port-file cleanup. Reaching this code only happens on the intended `LastWindowClosed` quit (the CEF message loop returned), so a hard exit is correct, not a crash. Other platforms keep the plain `drop(runtime)`.

## Verified
- Host reaches `AgentMux host shutdown complete (fast exit)` (it never got past `drop(runtime)` before).
- The whole instance — host + launcher + srv — exits cleanly in ~60 ms; **no lingering processes**, Dock tile gone.
- macOS-scoped (`#[cfg(target_os = "macos")]`); Windows/Linux unchanged.

Spec with full diagnosis + ruled-out hypotheses: `docs/specs/SPEC_MACOS_WINDOW_CLOSE_LIFECYCLE_2026-06-04.md`.

### DIFF
--- a/.changesets/1780584179-fix-macos-window-close-now-exits-the-instance-instead-of-han-aywa.md
+++ b/.changesets/1780584179-fix-macos-window-close-now-exits-the-instance-instead-of-han-aywa.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): window close now exits the instance instead of hanging hidden
--- a/agentmux-cef/src/main.rs
+++ b/agentmux-cef/src/main.rs
@@ -861,16 +861,37 @@ fn main() {
     // Clean shutdown.
     shutdown();
 
-    // Drop the tokio runtime after CEF shutdown.
-    drop(runtime);
-
     // Phase B.6 (post-fix): clean up the forwarding hint so a stale
     // file doesn't survive a graceful exit. (Hard crashes will leave
     // it behind; harmless because pipe-bind on next launch is
     // authoritative — see comment at the port_file declaration.)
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1269, opened 2026-06-04
feat(tab): VSCode-style flush tabs + folder content surface

AgentA-asaf:
## What

A perf-safe VSCode-style tab pass, in two layers:

**The look** — flush tabs with a 1px hairline divider (no whitespace gap), regular weight, no text-shadow, left-aligned names, the active close button at full opacity, and the active-tab accent on the **top** edge (VSCode default).

**The folder** — the tab content area + tile gaps paint `--workspace-surface`, so panes read as one dark folder instead of floating chips on the light header. The active tab **opens into** that surface (its boundary hairline drops via `.tab:not(.active)`); inactive tabs carry the 1px strip/content line.

## Perf (the important part)

Tab-switch speed comes from a keep-alive cache + a reveal gate (`store/tab-reveal.ts`) that lifts after 80ms of no >50ms long-tasks (else an 800ms cap). An earlier cut of this used `:has(.tab.active)` + stacked semi-transparent fills, which fired long-tasks on every streaming recalc/paint and **trapped the gate at 800ms** ("frozen" switches). This version stays off that path:
- boundary via **`.tab:not(.active)`** — re-evaluates only on a switch, not on every streaming recalc
- **single-fill** surface + a real `border` for the pane hairline — no 2nd semi-transparent layer, so no added overdraw

Verified live: a tab switch generates **1 long-task (68ms)** in a 2s window (was a continuous stream) → the gate lifts ~150ms, not 800ms.

## Tests
- `tabbar-dnd.test.ts` — pass
- `stylelint` clean on the modified scss

Spec (design + the full future surface model): `docs/specs/SPEC_TAB_CONTENT_FOLDER_SURFACE_2026_06_03.md`. Follows #1262 (responsive collapse, merged).

### DIFF
--- a/.changesets/1780584516-feat-tab-vscode-folder-look.md
+++ b/.changesets/1780584516-feat-tab-vscode-folder-look.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(tab): VSCode-style flush tabs + folder content surface
--- a/docs/specs/SPEC_TAB_CONTENT_FOLDER_SURFACE_2026_06_03.md
+++ b/docs/specs/SPEC_TAB_CONTENT_FOLDER_SURFACE_2026_06_03.md
@@ -144,3 +144,30 @@ Layer B is the smallest structural addition that makes the connection real and r
 - Changing tab sizing/responsive behavior (covered by #1262).
 - Reworking the TileLayout's gap/resize mechanics — only its *background reveal*.
 - Floating/torn-off pane chrome.
+
+---
+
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

AgentA-asaf:
Both P2s addressed: (1) active-tab hover now keeps `--block-bg-color` so the active tab stays continuous with the folder surface while hovered; (2) the full-bleed surface (panes ~0.75 effective vs gaps at 0.5) is the accepted look — recorded as shipped in spec §9, with gap-only noted as a future refinement, so the doc now matches the implementation.

PR #1271, opened 2026-06-05
fix(auth): single-flight login CLI lifecycle (supersede-kill + timeout reaper)

a5af:
## Problem (confirmed live leak)

The `claude auth login` CLI runs under a PTY and is *designed* to stay alive at the `Paste code here >` prompt. But the login lifecycle had three gaps:

1. **No supersede-kill** — `run_cli_login`/`run_cli_login_pty` spawned a fresh child on every attempt without killing the previous one.
2. **Lost handle** — the stored `cli_login_pty_pid` was overwritten on each attempt, so older children became permanently unkillable via `cancel_cli_login`.
3. **No timeout reaper** — the PTY reaper only exited when the child self-exited.

Result: abandoned/superseded logins accumulated as orphaned `node` processes. Caught two live `claude auth login` children at once during the audit. (See `ANALYSIS_AGENT_PANE_LIFECYCLE_LEAKS_2026-06-04.md`.)

## Fix

- **Supersede:** `run_cli_login` now `cancel_cli_login(&state)`s any in-progress login before spawning.
- **Generation guard** (`AppState.cli_login_generation`, `AtomicU64`): each attempt bumps it; a superseded login's reaper clears the shared `cli_login_*` slots **only if it still owns the current generation**. This is the subtle part — without it, the old reaper (when its killed child finally dies) would null out the *new* login's stdin handle, stranding the pasted code (the "stuck login" symptom).
- **Timeout reaper:** both transports kill the child after `LOGIN_REAP_TIMEOUT_SECS` (6 min, just past the frontend's 5-min poll) if it never completes. The PTY path polls `try_wait()` instead of a blocking `wait()`.
- **Frontend:** reaps the CLI when the attempt concludes (`launch-flow.ts`, success/timeout/error) and **unconditionally** on pane unmount (`useAgentControllerStatus.ts`) — closing the window between `runCliLogin()` and `setLoginWaiting(true)` where a close could orphan the child.

Three independent reap triggers now exist: supersede, frontend completion/unmount, backend timeout.

## Verification

- `cargo build -p agentmux-cef --release` ✓ (warning-clean for the touched code)
- `npx tsc --noEmit` ✓ for `launch-flow.ts` + `useAgentControllerStatus.ts`
- Repro to confirm post-merge: trigger login → trigger a second login → assert exactly one `claude auth login` PID survives; close pane mid-login → child gone ≤1s; complete a real login → `.credentials.json` written, "Logged in as …" fires.

### DIFF
--- a/.changesets/1780650174-fix-auth-single-flight-login-cli-lifecycle-supersede-kill-ti-io8v.md
+++ b/.changesets/1780650174-fix-auth-single-flight-login-cli-lifecycle-supersede-kill-ti-io8v.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): single-flight login CLI lifecycle — supersede-kill + timeout reaper
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -363,6 +363,13 @@ impl CliLoginStdin {
     }
 }
 
+/// Hard cap on how long a login CLI may sit at its paste prompt before the
+/// reaper kills it. Slightly longer than the frontend's 5-minute auth poll so
+/// the frontend (which also reaps on completion/cancel) wins normal cases;
+/// this is the backstop for a login whose frontend driver vanished (e.g. the
+/// pane was closed without its cleanup firing).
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1273, opened 2026-06-05
fix(agent): stop and drop subagent file watcher on agent unregister

a5af:
## Problem

`SubagentWatcher.watched_agents` was **push-only**. Every distinct agent that ever ran leaked one OS filesystem-watch handle + mpsc channel + idle debounce task for the rest of the process lifetime — surviving pane/agent deletion. (MED leak in `ANALYSIS_AGENT_PANE_LIFECYCLE_LEAKS_2026-06-04.md`.)

## Fix

Add `unwatch_agent(agent_id)`:
- `watched_agents.retain(|w| w.agent_id != agent_id)` drops the `WatchedAgent`, which drops its `notify` watcher. That closes the only surviving sender on the debounce channel, so the processing task self-terminates on its next `rx.recv()` returning `None`. No manual task abort needed.
- Idempotent — a no-op if the agent isn't watched.

Wire it into **`handle_reactive_unregister`** — the exact symmetric counterpart to the `watch_agent()` call in `handle_reactive_register` (both keyed on `agent_id`, both with `state.subagent_watcher` in hand).

Session records in `sessions` are plain data (not handles), so they're left as-is — the leak was the watcher/channel/task.

## Verification
- `cargo build -p agentmux-srv --release` ✓
- Symmetric with the existing register/unregister pair, so it fires on the same lifecycle event that created the watch.

### DIFF
--- a/.changesets/1780650433-fix-agent-stop-and-drop-subagent-file-watcher-on-agent-unreg-veti.md
+++ b/.changesets/1780650433-fix-agent-stop-and-drop-subagent-file-watcher-on-agent-unreg-veti.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): stop and drop subagent file watcher on agent unregister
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1274, opened 2026-06-05
fix(agent): own tool_chunk sub at body scope; drop dead detached login spawn

a5af:
Two MED lifecycle-leak fixes from `ANALYSIS_AGENT_PANE_LIFECYCLE_LEAKS_2026-06-04.md`.

## 1. `useAgentStream` — `tool_chunk` subscription torn down even on early-return
The per-block `tool_chunk` WPS subscription (`blockChunkUnsub`) is created at **hook-body** scope, but its only teardown lived inside `onMount`'s `onCleanup` — which is **skipped** when `onMount` early-returns (`if (!enabled || !blockId) return;`). So a mount with `enabled:false`/empty `blockId` would leak one global handler per mount (latent today: the sole caller hardcodes `enabled:true`).

Fix: register the teardown at **body scope** (where the subscription is created), and drop the now-redundant call from the `onMount` cleanup.

## 2. `cli_handlers` `runclilogin` — remove the dead detached-spawn
This srv-side login handler has **no live caller** (the active path is the CEF host IPC `run_cli_login`). It spawned a **detached, unkillable** `claude auth login` child via `tokio::spawn(async { child.wait().await })` — zero cancellation, zero timeout. Removed the spawn so it can't orphan a login process if ever wired up. `make_cli_cmd`/`RunCliLoginResult` stay used elsewhere → no unused-symbol warnings.

## Verification
- `npx tsc --noEmit` ✓ (useAgentStream)
- `cargo build -p agentmux-srv --release` ✓ (no new warnings)

### DIFF
--- a/.changesets/1780650869-fix-agent-own-tool-chunk-sub-at-body-scope-drop-dead-detache-68hs.md
+++ b/.changesets/1780650869-fix-agent-own-tool-chunk-sub-at-body-scope-drop-dead-detache-68hs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): own tool_chunk sub at body scope; drop dead detached login spawn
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1272, opened 2026-06-05
feat(linux): restore --features patched-libcef on CEF 148 via cef-dll-sys fork patch

AgentU-asaf:
## Summary

- PR #1221 bumped the workspace from CEF 146 → 148, which switched `cef-dll-sys` from our patched 146.7.0+146.0.12 (with the `begin_window_drag` slot on `_cef_window_t`) to upstream 148.3.0+148.0.9 (no slot).
- Result: `cargo build --features patched-libcef` has been broken on the workspace since the 148 bump, leaving Linux native window-drag silently downgraded to the no-op fallback in every AppImage shipped without the feature enabled.
- This PR adds a workspace `[patch.crates-io]` entry redirecting `cef-dll-sys` at [`AgentU-asaf/cef-rs#agentmux/148-begin-window-drag`](https://github.com/AgentU-asaf/cef-rs/pull/1), which re-applies the same mechanical edit (append `begin_window_drag` to the linux_x86_64 binding, struct size 888 → 896).

## How was this verified?

```
$ cargo build --release -p agentmux-cef --features patched-libcef
   Compiling cef-dll-sys v148.3.0+148.0.9 (https://github.com/AgentU-asaf/cef-rs?branch=agentmux%2F148-begin-window-drag)
   …
    Finished `release` profile [optimized] target(s) in 3m 28s

$ strings target/release/agentmux-cef | grep "BeginWindowDrag returned"
-[start_window_drag] BeginWindowDrag returned
```

That second grep is the key: if the no-op stub were compiled in, the string would read `patched-libcef feature disabled — native drag is a no-op` instead. The patched code path is now wired.

End-to-end (renderer up + drag interaction) is gated on §8.6 — the AppImage bundle scripts need to deploy the matching CEF 148 runtime files (paks/locales/icudtl) so the renderer process initializes. That's a separate PR.

## Companion PRs (spec §8 sequence)

- agentmuxai/cef#3 — CEF source forward-port (`agentmux/7778-drag-rightclick-and-transparency`)
- AgentU-asaf/cef-rs#1 — `cef-dll-sys` binding fork (`agentmux/148-begin-window-drag`)

## Blast radius

- macOS & Windows: unaffected. They don't enable `--features patched-libcef`, and the fork's non-Linux binding files are byte-identical to crates.io 148.3.0+148.0.9.
- CI: unaffected (no jobs build the Linux host with the feature today, since the feature was unbuildable until this PR).

## Spec

[...]

PR #1275, opened 2026-06-05
fix(linux): bundle CEF 148 Vulkan SwiftShader files so renderer process can start

AgentU-asaf:
## Summary

CEF 148 changed its software-rasterizer fallback from GLES SwiftShader to **Vulkan SwiftShader** (`libvk_swiftshader.so` + `vk_swiftshader_icd.json` + `libvulkan.so.1`). The current Linux bundle scripts don't copy those files. Without them, Chromium's GPU process can't find the Vulkan ICD at init, the renderer SIGTRAPs ~600ms after host start, and the AppImage lands on the recovery screen.

CEF 148 also added `headless_command_resources.pak` to the resource set.

This PR teaches `Taskfile.yml::bundle:linux` and `scripts/build-appimage-linux.sh` to copy the four new files alongside the existing GL/snapshot/icudtl set.

## Verification

```
task build:host && task build:backend && task build:frontend && task bundle
bash scripts/build-appimage-linux.sh /tmp
/tmp/AgentMux_0.42.0_amd64.AppImage
```

Result at +18s:
- 8 procs alive (main + GPU + utility + zygote + 2× renderer + 2× sidecar)
- libcef.so mapped: `Chrome/148.0.7778.180`
- Backend serving RPC (`object.GetObject`, `command=listrecentsessions`, etc.)
- Renderer firing focus events, agent cards rendering, no SIGTRAP
- Strip pass works: libcef.so 629 MB → 434 MB

The pre-fix AppImage from yesterday SIGTRAP'd at GPU init within ~600ms with the same libcef.so.

## Why no `--features patched-libcef` in the test build

This PR was smoke-tested with `cargo build --release -p agentmux-cef` (default features) because the workspace `[patch.crates-io]` required to compile `--features patched-libcef` on CEF 148 lives in independent PR #1272 (§8.5). The bundle-script logic exercised here doesn't depend on that feature — it's purely about CEF runtime file presence.

Once both this PR + #1272 land, `task package:linux` produces an AppImage with native window-drag fully restored on Mutter/Wayland.

## Blast radius

- Linux only. The macOS bundle path (`bundle:darwin`) is untouched.
- New `cp -f … 2>/dev/null || true` lines mean missing files don't break older CEF builds — the runtime stays backward-compatible with CEF 146 if someone bisects against a CEF 146 build dir.

## Spec

docs/specs/SPEC_CEF_148_LINUX_FORWARD_PORT_2026_06_04.md §8.6

## Test plan

- [x] CEF 148 AppImage launches + renderer comes up + backend serves RPC (verified on this machine)
- [x] Bundle picks up new files when `~/cef-build/.../Release_GN_x64` is the source
- [ ] CI bundle:linux runs against either the cef-dll-sys cached path or the explicit `AGENTMUX_CEF_RUNTIME_DIR` env override

### DIFF
--- a/.changesets/1780651010-fix-cef-148-bundle-runtime-files.md
+++ b/.changesets/1780651010-fix-cef-148-bundle-runtime-files.md
@@ -1,0 +1,22 @@
+---
+type: fix
+scope: linux
+title: Bundle CEF 148 Vulkan SwiftShader runtime files into AppImage
+---
+CEF 148 ships its software-rasterizer fallback as Vulkan SwiftShader
+(`libvk_swiftshader.so` + `vk_swiftshader_icd.json` + `libvulkan.so.1`),
+replacing the GLES-based SwiftShader path that CEF 146 used. The current
+bundle scripts don't copy these — Chromium's GPU process fails to find
+the Vulkan ICD and the renderer process crashes with SIGTRAP at GPU
+init, leaving the AppImage on a recovery screen.
+
+Also adds `headless_command_resources.pak`, a new resource bundle in
+CEF 148.
+
+Verified end-to-end:
+  task build:host && task build:backend && task build:frontend && task bundle
+  bash scripts/build-appimage-linux.sh /tmp
+  /tmp/AgentMux_0.42.0_amd64.AppImage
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1277, opened 2026-06-05
fix(auth): clearer 2-step Claude login box with paste-to-submit

a5af:
Redesigns the agent-pane auth box into an explicit two-step flow so the manual-paste login is actually discoverable (this is what tripped up real logins — the CLI has no auto-callback, you must copy the code the page shows and paste it back).

- **Step framing:** "Sign in to Claude" → "1 · Authorize in your browser" (URL + Copy) → "2 · Paste the code from that page".
- **Hint text** explains the page shows an **authorization code** to copy + paste.
- **Auto-submit on paste** (pasting the code submits it — no extra click) + a **"Paste & submit"** button.
- **DOM-value fallback** in `submitCode` so a focus-desynced paste isn't silently dropped.

Pairs with the login-CLI lifecycle fix (#1271): together the login is now single-child, self-healing, and clearly explained.

### DIFF
--- a/.changesets/1780649913-fix-auth-clearer-2-step-claude-login-box-with-paste-to-submi-etwb.md
+++ b/.changesets/1780649913-fix-auth-clearer-2-step-claude-login-box-with-paste-to-submi-etwb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1281, opened 2026-06-05
fix(agent-pane): Phase 4 — correct zoom units (single ÷zoom at the measure boundary)

AgentA-asaf:
## Phase 4 — formalize zoom (INV-2)

Completes the agent-pane layout-reducer epic (Phases 0–3 merged in #1270). The
slice operates in **unzoomed CSS px**; this PR makes the component feed it in
that unit consistently, collapsing the scattered `÷zoom` to a single boundary.

### What the measurement found

A CDP probe (`tools/tests/ph4-zoom-probe.mjs`) injects a faithful scroll
container under the ancestor CSS `zoom` and reads each property the agent pane
relies on, at zoom 0.5 / 1 / 2:

| property | ratio vs zoom 1 | unit |
|---|---|---|
| `scrollTop` (set 100 → reads 100) | 1.000 | **unzoomed** |
| `clientHeight` / `scrollHeight` | 1.000 | **unzoomed** |
| `offsetTop` / `offsetHeight` | 1.000 | **unzoomed** |
| `getBoundingClientRect().height` | 0.5 / 2.0 | **zoomed** |

So `getBoundingClientRect` (the measure RO) is the **only** zoomed read.

### The latent bug this fixes

`handleScroll`, the hidden→visible RO, and `scrollToNode` were dividing
`scrollTop`/`clientHeight` by zoom — but those are *already* unzoomed. At any
**zoom ≠ 1** the slice received a wrong `scrollTop`/`viewport` (off by a factor
of zoom) → mis-windowed rows and a mis-targeted scroll-to-node. The cutover only
exercised zoom = 1 (where `÷1` is a no-op), so it never surfaced.

### The change

Feed `scrollTop` / `clientHeight` / `offsetTop` **raw**; keep the lone `÷zoom`
at the measure RO. This is exactly the spec's *"single ÷zoom at the `RowMeasured`
boundary"* (Phase 4, INV-2). Single file: `AgentDocumentVirtualList.tsx`.

### Verification

- **Units probe** — live in cef-146, definitive (table above).
- **zoom = 1 is byte-identical** (`x / (1||1)` ≡ `x`), so the cutover's §7 live
  result (0 overlap / 0 drift) carries over unchanged; only the zoom ≠ 1 path
  differs.
- **Reducer** is zoom-invariant by construction and already unit-tested
  (`INV-2 — zoom is layout-invariant`); no reducer change here.
- **Frontend build** clean.
- ⚠️ **Not run:** the live §7 drift harness *at zoom 0.5/2* on a real pane — the
  throwaway measurement dev has no virtualized conversation to seed. The units
  probe + zoom-1-identity cover correctness; a real-session §7-at-zoom is a
  recommended final smoke.

### DIFF
--- a/.changesets/1780657695-fix-agent-pane-fix-virtualized-layout-under-zoom-1-phase-4-s-odw6.md
+++ b/.changesets/1780657695-fix-agent-pane-fix-virtualized-layout-under-zoom-1-phase-4-s-odw6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): fix virtualized layout under zoom != 1 (Phase 4 - single zoom-normalize at the measure boundary)
--- a/frontend/app/view/agent/virtualization/AgentDocumentVirtualList.tsx
+++ b/frontend/app/view/agent/virtualization/AgentDocumentVirtualList.tsx
@@ -75,9 +75,11 @@ export interface AgentDocumentVirtualListProps {
     onTogglePin: (id: string) => void;
     /**
      * Live per-pane zoom factor (the same value applied as CSS `zoom` on
-     * `.agent-view` in agent-view.tsx). Used to normalize the measure RO + the
-     * Scrolled / scrollToNode math into unzoomed CSS px so slice positions don't
-     * double-count zoom and overlap — see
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1282, opened 2026-06-05
fix(dev): frontend-load error page Retry actually retries

a5af:
## The bug (why Retry did nothing)

When the host can't reach the Vite dev server on launch (`ERR_CONNECTION_REFUSED`), `on_load_error` renders a "Failed to load AgentMux frontend" page **as a `data:text/html;base64,…` URI**. Its Retry button was:

```html
<button class="retry" onclick="location.reload()">Retry</button>
```

`location.reload()` reloads the **current** document — which is now the `data:` error page itself, **not** the original `http://localhost:5289/…` URL. Worse, reloading a `data:` URI returns `ERR_ABORTED`, which `on_load_error` early-returns on — so each click silently re-rendered the same error page. The original URL was never re-tried.

Host log proof (one session): original `ERR_CONNECTION_REFUSED` for `http://localhost:5289/?ipc_port=…`, then repeated `[load-error][ENTRY] url="data:text/html;base64,…" error="ERR_ABORTED"` on every Retry.

## Fix

- Retry now navigates to the original `failed_url` (`location.href='{failed_url}'`).
- Added a **1.2s auto-retry** to the same URL, so the common launch race — the host (on a warm build) starting before Vite binds its port — **self-heals** instead of stranding the user on the error page.

Dev-only path (packaged builds bundle the frontend and never hit this error page).

### DIFF
--- a/.changesets/1780667487-fix-dev-frontend-load-error-retry-navigates-to-the-real-url--nea1.md
+++ b/.changesets/1780667487-fix-dev-frontend-load-error-retry-navigates-to-the-real-url--nea1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dev): frontend-load error Retry navigates to the real URL + auto-retry
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -1358,6 +1358,22 @@ impl AgentMuxHandler {
         let failed_url = failed_url.map(CefString::to_string).unwrap_or_default();
         let error_code_i32 = error_code_raw as i32;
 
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

a5af:
Addressed both:
- **P1 (auto-retry loop on browser panes):** the 1.2s auto-retry now fires only when `!self.is_browser_pane` — i.e. the dev frontend / main window. Browser panes get a manual Retry only, so a failed/offline pane URL can't loop.
- **P2 (unescaped URL):** `failed_url` is now JSON-encoded into a JS string literal (`var __amxTarget = {json};`) and the button/timer call `__amxRetry()`, so a `'` in a real URL can't break it.

PR #1283, opened 2026-06-05
fix(auth): seed isolated claude creds from global to stop the validate spin

AgentA-asaf:
## The bug

Creating an agent leaves it stuck "not authenticated" forever, CPU pinned and
login timing out — even though `claude` is logged in globally.

Logs show the two-phase claude auth check (SPEC_AUTH_CHECK_FALSE_POSITIVE):
- **Phase 1** finds credentials in EITHER the isolated `CLAUDE_CONFIG_DIR` **or**
  the global `~/.claude` fallback (`cli_handlers.rs:304-323`).
- **Phase 2** runs `claude auth status --json` against ONLY the isolated
  `CLAUDE_CONFIG_DIR`.

On a fresh instance the isolated dir is empty but the user is logged in
globally → Phase 1 says "found", Phase 2 says `loggedIn: false`. The frontend
re-polls every ~2s, each poll re-spawns the CLI (the **CPU churn**), and
`run_cli_login` PTY-times-out at **15s**. An unrecoverable "found-but-not-
logged-in" spin — credentials present, never recognized.

## The fix

When tokens were found via the global fallback but the isolated dir has no
`.credentials.json`, copy the global file in **once**. Phase 2 then validates a
populated dir, and the spawned agent (same dir) authenticates too. Only copies
when **absent** → a real per-instance login still wins, so multi-account
isolation is preserved.

## Verification

Mechanism validated **live** against the bundled CLI (v2.1.160):

| `CLAUDE_CONFIG_DIR` | `claude auth status --json` |
|---|---|
| seeded (global creds copied in) | `loggedIn: true` |
| empty (current fresh-instance state) | `loggedIn: false` |

`cargo check -p agentmux-srv` clean. End-to-end relaunch verification next.

### DIFF
--- a/.changesets/1780668404-fix-auth-seed-isolated-claude-creds-from-global-to-stop-the--8asi.md
+++ b/.changesets/1780668404-fix-auth-seed-isolated-claude-creds-from-global-to-stop-the--8asi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): seed isolated claude creds from global to stop the validate spin
--- a/agentmux-srv/src/server/cli_handlers.rs
+++ b/agentmux-srv/src/server/cli_handlers.rs
@@ -334,6 +334,34 @@ pub fn register_cli_handlers(engine: &Arc<WshRpcEngine>, state: &AppState) {
                         return Ok(Some(serde_json::to_value(&result).unwrap()));
                     }
 
+                    // Bootstrap the isolated CLAUDE_CONFIG_DIR from the global login.
+                    // Phase 2 (and the spawned agent) validate against the isolated dir;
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Superseded by the provider-pinned-auth model (per directive: every provider's requirements is a pinned version, completely independent of installed instances). The real fix re-roots provider config+auth to a shared, instance-independent location (`~/.agentmux/providers/<provider>/`) so the empty-per-instance auth dir that caused the validate-spin stops existing — no per-instance credential seeding needed (codex's P1 dissolves under that model). Root-cause diagnosis preserved in the commits; new PR to follow.

PR #1284, opened 2026-06-05
fix(macos): native close button on the DevTools window

a5af:
## Problem

On macOS, opening DevTools gave a window with **no close button** — you couldn't close it with the X. The DevTools popup is created Chrome-style with `frameless=false` (`app.rs:on_popup_browser_view_created`), and the intended native title bar with close/min/zoom **works on Windows** (see `docs/analysis/devtools-secondary-window.md`) but was never wired for macOS — CEF's Chrome-style window comes up there without the standard traffic-light buttons.

## Fix

In `on_window_created`, for **non-frameless** popups (DevTools) on macOS, reach the `NSWindow` via `[NSView window]` (the CEF Views window handle is the NSView ptr on macOS) and:
- force a standard `Titled | Closable | Miniaturizable | Resizable` style mask, and
- un-hide the three standard window buttons.

The frameless main window (custom HTML title bar) is explicitly left untouched (`if !self.frameless`). Raw libobjc FFI, mirroring the existing `set_macos_app_display_name` shim.

Also drops the stale `defwidget@devtools` reference in `CLAUDE.md` — DevTools is a menu/command (`dev:devtools`), not a widget.

## Verification

- `cargo build -p agentmux-cef --release` ✓
- **Confirmed on macOS**: DevTools now opens with a working red close button. ✅

### DIFF
--- a/.changesets/1780655977-fix-macos-native-close-button-on-the-devtools-window-fxko.md
+++ b/.changesets/1780655977-fix-macos-native-close-button-on-the-devtools-window-fxko.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): native close button on the DevTools window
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1287, opened 2026-06-05
fix(auth): macOS Claude login completes — detect Keychain-stored creds

a5af:
## The real bug (not what it looked like)

The Claude login on macOS *always succeeded* — the browser auto-opens, you authorize, and the CLI auto-completes the **loopback happy path** (no code to paste) and stores credentials in the **macOS Keychain** (`Claude Safe Storage`), exactly as designed.

It only *looked* stuck because of a **detection** bug. `CheckCliAuth` has a fast path: "no `.credentials.json` file → **definitely** not authenticated, skip the CLI." On macOS there is **never** a credentials file (the Keychain is the store), so a completed login was never detected and the launch flow's 2s poll spun forever.

Proof from the new instrumentation:
```
[login-pty] Paste code here if prompted > Login successful.   ← CLI auto-completed
claude auth status → "loggedIn": true, asafebgi@gmail.com
creds → macOS Keychain "Claude Safe Storage"  (no .credentials.json anywhere)
```

## Fix

- **`CheckCliAuth` (agentmux-srv):** on macOS, when no credentials *file* is present, fall through to `claude auth status` (which reads the Keychain — and does so without a prompt, since the CLI owns that item) instead of returning not-authenticated. Windows/Linux keep the fast file-only short-circuit (the file *is* the store there).
- **`run_cli_login_pty` (agentmux-cef):** keep draining the login CLI's PTY *after* the URL and **log** its output as `[login-pty]`. Before, the CLI's `… > Login successful.` was discarded — a completed login was a black box. Draining also prevents the CLI blocking on a full PTY output buffer (the pipe path already drains; the PTY path didn't).

## Verification (macOS)

After a completed login + this fix:
```
claude auth check: no credentials file — checking Keychain via CLI (macOS)
[rpc-perf] checkcliauth handler=243ms        ← ran claude auth status
subprocess spawned                            ← agent launched, no login box
```

### DIFF
--- a/.changesets/1780672511-fix-auth-macos-claude-login-completes-detect-keychain-stored-kx3f.md
+++ b/.changesets/1780672511-fix-auth-macos-claude-login-completes-detect-keychain-stored-kx3f.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): macOS Claude login completes — detect Keychain-stored creds
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -658,16 +658,31 @@ async fn run_cli_login_pty(
     tokio::task::spawn_blocking(move || {
         use std::io::BufRead;
         let mut reader = std::io::BufReader::new(reader);
-        let mut found: Option<String> = None;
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #850, opened 2026-05-14
feat(auth): AuthFlowController + RPC bindings (PR B-2)

AgentA-asaf:
**Replaces #846** (auto-closed when its stacked base merged via #849).

Stacks naturally on top of the now-landed #849 (AuthState reducer). Adds the side-effect-bearing partner.

## Summary

- **5 typed RPC bindings** in \`rpc-api.ts\`: AuthStartCommand, AuthPollCommand, AuthSubmitCallbackCommand, AuthCancelCommand, AuthSubmitApiKeyCommand.
- **\`AuthSessionStatus\` global type** in \`gotypes.d.ts\` mirroring the Rust enum (\`tag = "status", rename_all = "kebab-case", rename_all_fields = "camelCase"\`).
- **\`AuthFlowController\` class** — owns one AuthState reducer instance + an injectable poll loop. Async actions: \`selected\` / \`connect\` / \`cancel\` / \`submitCallback\` / \`submitApiKey\` / \`dispose\`. Stale-poll guard short-circuits both at dispatch-time (reducer gate) AND at RPC-time (don't fire poll for sessions the user already left behind).
- 11 controller tests + 32 reducer tests (43 total) all green.

## Test plan

- [x] \`npx vitest run frontend/app/view/agent/auth\` — 43 pass
- [x] \`tsc --noEmit\` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1291, opened 2026-06-05
fix(auth): provider auth in one shared, instance-independent dir (kills the validate-spin)

AgentA-asaf:
## The bug (a regression)

A fresh agent never authenticates — UI stuck "not authenticated", CPU pinned
~100%, login times out at 15s — even though `claude` is logged in. Full
evolution: `docs/retro/retro-provider-auth-isolation-regression-2026-06-05.md`.

**Root cause — two compounding changes.** The original design
(`provider-auth-isolation.md`, 2026-03-21) shared provider auth per *version*
with a stated invariant (§4: *validate the dir the agent runs in*). Then:
- **#850** added a two-phase `CheckCliAuth` whose phase 1 falls back to global
  `~/.claude` while phase 2 validates only the isolated dir → "check global / run
  isolated", the §4 hazard.
- **#1027 (channels)** re-rooted config (hence auth) to **per-channel** dirs.

→ a fresh channel / branch / build starts with an **empty** isolated auth dir;
phase 1 finds global creds, phase 2 validates the empty dir → `loggedIn:false`
forever → a 2 s CLI re-spawn spin.

## The fix — providers are pinned + instance-independent

- Root the **default** provider auth at `~/.agentmux/shared/providers/<provider>/`
  (`DataPaths::provider_auth_dir`) — account-wide, version/channel-independent,
  like the CLI. One login shared everywhere → the empty-per-instance dir cannot
  exist by construction.
- **Restore §4** in `CheckCliAuth`: validate the **same dir the agent runs in**
  (`CLAUDE_CONFIG_DIR` if set, else global) — never "isolated OR global".
- **One-time import** of an existing global login into the shared dir,
  sentinel-gated so a `logout` sticks (a single shared bootstrap, **not**
  per-instance reseeding).
- Per-identity bundle override (multi-account) unchanged.

## Verification

- `cargo check -p agentmux-srv` clean; the **`provider_auth_dir` contract test
  passes** (channel/mode-independent, never under `config_dir`).
- The import → CLI-validates → authed mechanism was empirically confirmed earlier
  (seeded `CLAUDE_CONFIG_DIR` → `loggedIn:true`); this PR only changes *which* dir.
- The cef change is a one-line re-point of the proven `version_config_dir`
  pattern (confirmed field + type). ⚠️ The **local full build** hit the known
  CEF-download / Defender rename race (env, not code) — **CI builds it clean**; a
  full GUI smoke is pending the env.

Spec: `docs/specs/SPEC_PROVIDER_PINNED_AUTH_2026_06_05.md`. Supersedes the closed
band-aid #1283. Follow-ups (in the spec): pin `pinned_version`, migration sweep,
cross-module §4 integration test.

### DIFF
--- a/.changesets/1780675514-fix-auth-shared-provider-auth-dir-restore-the-validate-the-d-3zgu.md
+++ b/.changesets/1780675514-fix-auth-shared-provider-auth-dir-restore-the-validate-the-d-3zgu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): shared provider auth dir; restore the validate-the-dir-you-run-in invariant
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -100,13 +100,21 @@ pub fn ensure_auth_dir(
         ));
     }
 
-    let config_dir = state.version_config_dir.lock();
-    let config_dir = config_dir
+    // The DEFAULT provider auth/config dir lives under the account-wide, version-
+    // and channel-independent shared root (`~/.agentmux/shared/providers/<provider>/`),
+    // NOT the per-channel config dir. One login is shared across every instance /
+    // channel / version — the structural fix for the per-channel validate-spin
+    // regression (docs/retro/retro-provider-auth-isolation-regression-2026-06-05.md).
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #1292, opened 2026-06-05
fix(a11y): don't freeze loading spinners under prefers-reduced-motion

a5af:
## Symptom

On macOS the loading spinner is **stuck/frozen**; on Windows it spins fine.

## Cause (not platform code — an OS setting + an over-eager rule)

macOS "Reduce Motion" (Accessibility → Display) was **on** (`defaults read com.apple.universalaccess reduceMotion` → `1`), which sets `prefers-reduced-motion: reduce`. Two spinners then disable themselves:
- `_composer-strip.scss` — the agent "working" spinner, via `@include mixins.respect-reduced-motion { … animation: none }`
- `_install-modal.scss` — `@media (prefers-reduced-motion: reduce) { .agent-install-modal-spinner { animation: none } }`

It "works on Windows" only because Reduce Motion is off there. Nothing platform-specific in the code.

## Fix

A **frozen loading indicator reads as a hung app** — worse for everyone, and *especially* for motion-sensitive users, who then can't tell it's busy. Loading spinners are **functional status indicators**, not decorative motion (a small continuous rotate is low vestibular risk), so they now **keep spinning under reduced motion, just slower/gentler** (2.4s / 3.2s) instead of freezing.

Genuinely decorative animations (modal entrance/exit in `modal-layer.scss`, message-enter in `_document.scss`) **stay suppressed** under reduced motion — only the loading spinners are exempted.

SCSS-only; hot-reloads. Verify with macOS Reduce Motion **on**: the agent spinner now rotates (gently).

### DIFF
--- a/.changesets/1780689507-fix-a11y-keep-loading-spinners-spinning-under-reduced-motion-cnr6.md
+++ b/.changesets/1780689507-fix-a11y-keep-loading-spinners-spinning-under-reduced-motion-cnr6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(a11y): keep loading spinners spinning under reduced-motion (macOS)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1293, opened 2026-06-05
fix(virt): cap streaming buffer to prevent replaceChild crash across turns

a5af:
## The bug (root-caused from diagnostics added in this PR)

After the first turn, the sticky frontier in `AgentDocumentVirtualList` never advanced. Subsequent turns kept appending to `streamingNodes`, growing it past `STREAMING_BUFFER_SIZE=50` (e.g. 53, 57…) while `virtCount` stayed fixed. On the next `StreamFlush`, Solid's position-keyed `<Index>` saw a longer array than its previous render cycle and `reconcileArrays` threw:

```
NotFoundError: Failed to execute 'replaceChild' on 'Node':
  The node to be replaced is not a child of this node.
  at reconcileArrays (solid-js/web/dist/dev.js:185)
```

`BlockErrorBoundary` caught it, unmounting/remounting the pane — producing the `CASCADE_DETECTED(StreamFlush)` + `dispatchPane dropped(StreamFlushObserved)` sequence. Pane left blank.

**Evidence from the `render_trail` in the error-boundary dump** (from new diagnostics also in this PR):
- Two repros, identical pattern: turn 1 ends (`TurnEnd`), turn 2 starts (`TurnStart`), `streamCount` grows 53→54→56→57 while `virtCount=38` or `45` and `frontier=toolu_01…` stays frozen → crash on `StreamFlush`.

## Fix

In the `partition` memo, after the stale-frontier re-anchor (`splitIndex === -1`), also re-anchor when `streamingNodes.length > STREAMING_BUFFER_SIZE`. This caps the buffer at 50 on every recompute, advancing the frontier across turns. Safe: the memo runs synchronously before the next `<Index>` render, so no node crosses subtrees mid-render.

## Also: dispose-reason diagnostics in `agent-view.tsx`

Adds `onCleanup` instrumentation that logs the disposing stack + `render_trail` + last 40 reducer dispatches when the pane disposes unexpectedly during a turn — the instrumentation that exposed this root cause. Low cost in normal operation (only dumps on unexpected mid-turn dispose).

## Prior art

The sticky-frontier design was specifically built to prevent this class of crash (`docs/analysis/AGENT_PANE_REPLACECHILD_CRASH_ON_SEND_2026_05_27.md`) but only handled the "node migrates on every append" case, not the "buffer grows past 50 across turns" case.

### DIFF
--- a/.changesets/1780692379-fix-virt-cap-streaming-buffer-to-prevent-replacechild-crash--83u9.md
+++ b/.changesets/1780692379-fix-virt-cap-streaming-buffer-to-prevent-replacechild-crash--83u9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(virt): batch StreamFlush dispatches to prevent replaceChild crash
--- a/frontend/app/view/agent/agent-view.tsx
+++ b/frontend/app/view/agent/agent-view.tsx
@@ -32,6 +32,8 @@ import {
 import { isInterruptibleTurn, workingFromPhase } from "@/app/store/agent-pane-state/types";
 import type { SubagentLinkNode } from "./types";
 import { openSubagentPane, isSubagentPaneOpen } from "@/app/store/subagent-pane-manager";
+import { getRecentDispatches } from "@/app/store/command-source";
+import { getTrail } from "@/log/render-trail";
 import { useAgentStream } from "./useAgentStream";
 import { useActivityLog } from "./hooks/useActivityLog";
 import { useSessionDigest } from "./hooks/useSessionDigest";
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
Addressed both:

**Codex P1 (per-append migration):** Changed threshold to `2× STREAMING_BUFFER_SIZE`. After a re-anchor we land at exactly 50 streaming nodes; the next re-anchor only fires after another full 50-node turn's worth of appends — so a single append never triggers a migration. The original invariant (no migration on simple append) holds.

**Reagent P2 (stale comment):** Updated the `stickyFrontierId` declaration comment to document the 2× exception explicitly.

a5af:
Addressed P1 + P2:

**P1 (re-anchor on every append + 2× threshold above crash range):** Both previous approaches were wrong. Replaced with a **prev/cur stream-count crossing detector**: re-anchor only when `prevStreamCount ≤ 50 && curStreamCount > 50`. This fires exactly once per 'below-threshold window' — i.e. once when the second turn's appends first cross 50, then never again for that turn's subsequent appends (prevStreamCount is already >50). No migration on simple appends; no threshold above the observed crash range (53–57).

**P2 (`[object Object]` in dispose log):** Changed ``${phase}`` to ``${JSON.stringify(phase)}`` so the discriminated-union is serialized correctly.

a5af:
Addressed reagent P1 + P2 (round 3):

**P1 (per-append re-anchor):** Fixed the `prevStreamCount` assignment order. It now gets set to the **raw overflow value** (before re-anchoring) inside the firing branch, so after re-anchoring to 50 the next memo sees `prev=51 > 50` and the condition is false. The `else` branch updates normally. Reset to 0 in the `splitIndex=-1` path so the detector re-arms after history clear/reset. Trace: Turn1 ends (prev=50) → Turn2 append1 (cur=51): fires once, prev=51 → Turn2 append2 (cur=51): prev=51>50, no re-anchor → all subsequent appends: no re-anchor.

**P2 (unconditional stack log):** Gated behind `if (midTurn)` — only logs on unexpected mid-turn dispose.

a5af:
Withdrew all the frontier-advancement attempts — reagent was right that they all re-introduce per-append migration.

**New approach: fix the race at the dispatch level.**

The `replaceChild` crash is a concurrent-reactive issue: `flushPendingNodes` makes two sequential Solid store writes (`StreamFlush` to the document store, then `StreamFlushObserved` to pane state). Without `batch()`, the first write synchronously triggers all dependents — including the `<Index>` render-effect reconciler — before the second write lands. If a `DocumentRow` update (triggered by the first write) also runs mid-reconcile, the `<Index>` outer reconciler calls `replaceChild` on a node that was just moved by the inner update.

Wrapped both writes in `batch()` (`useAgentStream.ts`). In Solid, `batch()` defers render effects (`createRenderEffect`, including `<Index>`) until after the batch completes, so the entire `<Index>` reconcile happens once with the final fully-consistent document state — no interleaved DOM mutations. The partition memo (`createMemo`) still runs synchronously inside the batch as expected.

`AgentDocumentVirtualList` is reverted to its original clean state (no frontier-advancement logic added). `agent-view.tsx` dispose log: gated on `midTurn` as before.

a5af:
Addressed both P2s: stale comment in VirtualList replaced with a note pointing to the `batch()` fix in `useAgentStream.ts`; changeset message updated from 'cap streaming buffer' to 'batch StreamFlush dispatches'.

PR #1299, opened 2026-06-06
fix(virt): accumulate tool chunks in RAF buffer to prevent replaceChild crash

a5af:
## Root cause

`ToolChunkAppend` was dispatched **directly** from the `tool_chunk` WPS event handler (`useAgentStream.ts:136–158`) — one immediate `documentAtom` signal write per chunk, outside any `batch()`.

PR #1293 wrapped `StreamFlush`+`StreamFlushObserved` in `batch()`, which serialized those two writes. But `ToolChunkAppend` was a separate writer, still firing independent `runUpdates` frames. During active tool streaming, a `ToolChunkAppend` WPS event and a RAF `flushPendingNodes` call can land in the same browser task. Two independent `runUpdates` frames → `<Index>` reconciler holds a stale `current` array when the interleaved write lands → `reconcileArrays` calls `replaceChild` on a node already moved → `NotFoundError`.

This is why crashes persisted at `streamCount=85–86` after #1293: the `StreamFlush`/`StreamFlushObserved` interleave was fixed, but the `ToolChunkAppend` writer was untouched.

Full analysis: `docs/retro/RETRO_REPLACECHILD_CRASH_2026-06-06.md`

## Fix

Accumulate tool chunks in a `pendingChunks[]` buffer and drain them inside the **same `batch()`** as `StreamFlush`/`StreamFlushObserved` in `flushPendingNodes`. All `documentAtom` writes from streaming now originate from one code path and one Solid reactive frame — no two-writer interleave possible.

## Verification

- `npx tsc --noEmit` ✓
- Hot-reloaded in the running dev app; test with a multi-turn conversation that uses tool calls (bash runner, read, etc.)

### DIFF
--- a/.changesets/1780725744-fix-virt-accumulate-tool-chunks-in-raf-buffer-to-prevent-rep-yo3v.md
+++ b/.changesets/1780725744-fix-virt-accumulate-tool-chunks-in-raf-buffer-to-prevent-rep-yo3v.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(virt): accumulate tool chunks in RAF buffer to prevent replaceChild crash
--- a/frontend/app/view/agent/useAgentStream.ts
+++ b/frontend/app/view/agent/useAgentStream.ts
@@ -15,7 +15,7 @@ import { batch, createEffect, onCleanup, onMount } from "solid-js";
 import { createTranslator } from "./providers/translator-factory";
 import type { PendingMessage, SignalPair } from "./state";
 import { ClaudeCodeStreamParser, STARTUP_HEADING_RE } from "./stream-parser";
-import type { DocumentNode, SessionStats, UserMessageNode } from "./types";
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

a5af:
Addressed all three:

**Codex P1 (chunks dropped if ToolNode not yet created):** Moved tool-chunk appends to AFTER `StreamFlush` inside the batch. `StreamFlush` creates the `ToolNode` first; `ToolChunkAppend` then has a node to append to. The late-subscribe race (WPS chunks arriving before the tool node exists) is not new behavior — `findToolIndex` dropped those before too — but ordering StreamFlush first minimises the window.

**Reagent P2a (truncate cleanup misses `pendingChunks`):** Added `pendingChunks = []` to the truncate cleanup block.

**Reagent P2b (stream-json `tool_chunk` branch still direct-dispatches):** Routed through `pendingChunks.push() + scheduleFlush()` to hold the single-writer invariant.

PR #1302, opened 2026-06-06
fix(virt): cap streaming buffer at STREAMING_BUFFER_SIZE to fix replaceChild crash

AgentA-asaf:
## Root Cause

`<Key>` from `@solid-primitives/keyed` returns a plain JS array. SolidJS's `insertExpression` passes that array to `reconcileArrays` (solid-js/web), which corrupts its sentinel tracking when the array grows past the initial render size — producing:

```
NotFoundError: replaceChild: The node to be replaced is not a child
  reconcileArrays (solid-js/web:593)
  insertExpression (solid-js/web:568)
  createRenderEffect
```

The keying strategy (`<Index>` position-keyed vs `<Key>` id-keyed) does not change this — both return arrays that `insertExpression` reconciles. **Array growth is the trigger, not keying.**

This means the fix in #1300 (`<Index>` → `<Key>`) was necessary but not sufficient: the streaming buffer still grew unbounded across turns (frontier set once, never advanced), so `reconcileArrays` still saw a growing array.

## Fix

In the `partition` createMemo, after the stale-frontier recompute, add a cap check:

```ts
if (result.streamingNodes.length > STREAMING_BUFFER_SIZE) {
    stickyFrontierId = initialStickyFrontierId(nodes, STREAMING_BUFFER_SIZE);
    result = partitionForVirtualization(nodes, STREAMING_BUFFER_SIZE, stickyFrontierId);
}
```

This pins `streamingNodes` at ≤ 50 items always. The oldest buffer node migrates to the virtualized head on each turn boundary.

**Why safe with `<Key by={n => n.id}>`**: the migrated node exits the `<Key>` map by id with no position-based state leakage (no `<Index>` position reuse). `reconcileArrays` in the streaming buffer only ever sees a fixed-size 50-item array.

## History

- #1293 (`3347cc3f`) — described this exact fix in the commit message but only updated a comment; cap code was never added
- #1299 (`887e8482`) — agento added `batch()` in `useAgentStream.ts` (prevents interleaved writes but doesn't cap the array)
- #1300 (`82cff777`) — switched `<Index>` → `<Key>`, still no cap, crash persists in 0.43.1

## Test

All 19 `streaming-buffer.test.ts` tests pass. The `partitionForVirtualization` invariants are unaffected — the function itself never moves the frontier; the cap-advance is in the caller.

Follow-up to #1300.

### DIFF
--- a/frontend/app/view/agent/virtualization/AgentDocumentVirtualList.tsx
+++ b/frontend/app/view/agent/virtualization/AgentDocumentVirtualList.tsx
@@ -706,10 +706,15 @@ export function AgentDocumentVirtualList(props: AgentDocumentVirtualListProps): 
                     (render_trail 2026-06-05; fixed in #1300.)
                   - <Key by={n => n.id}> keys each slot by stable node
                     id. Token updates (new object, same id) fire the
-                    slot's accessor signal, no remount. Array growth
-                    simply adds new keyed slots at the end — no DOM
-                    sentinel corruption, no position shifting, no
-                    cross-slot state leakage. */}
+                    slot's accessor signal — no remount, no position
+                    shifting, no cross-slot state leakage on cap-advance
+                    migrations. NOTE: <Key> still returns a plain array;
+                    insertExpression passes it to reconcileArrays, which
+                    crashes identically to <Index> if the array grows past
+                    the initial render size. The cap-advance in the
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #1303, opened 2026-06-06
fix(virt): defer SessionEnd, batch HistoryLoaded, Show-guard streaming buffer

a5af:
## Summary

Closes the three remaining `documentAtom` write races identified in `SPEC_REPLACECHILD_CRASH_FULL_ANALYSIS_AND_FIX_2026-06-06.md` that can produce the `replaceChild NotFoundError` in the `<Index>` streaming-buffer reconciler. Two crashes were observed today (05:45 UTC and 11:46 UTC). The first was fixed by #1293 + #1299; this PR closes the residual paths.

### Fix 1 — `queueMicrotask(SessionEnd)` in `useAgentStream.ts` `onCleanup`

During error-boundary cleanup the `<Index>` streaming-buffer scope is still partially live while `onCleanup` runs synchronously. The existing `dispatchDoc({ type: "SessionEnd" })` writes `documentAtom` in that window, which re-triggers `reconcileArrays` on a half-torn-down DOM → `replaceChild` fails. Deferring to a `queueMicrotask` ensures all scope disposal completes before the write fires. By then the `<Index>` effect is removed from the computation graph; the write still runs the orphan-scrub reducer but no reactive effect reconciles the dead DOM. `model.dispatchDoc` uses the soft `dispatchIfRegistered` variant so a gone slot is a silent no-op.

### Fix 2 — `batch(HistoryLoaded)` in `useHistoryPagination.ts`

`HistoryLoaded` was dispatched from an `async` RPC callback without any `batch()` wrapper — a second independent `documentAtom` writer. If this microtask resolves in the same browser task as a RAF `StreamFlush` (possible under certain Promise scheduling), two `runUpdates` frames can interleave. Wrapping in `batch()` is a one-liner that closes the race.

### Fix 3 — `<Show when={partition()}>` guard in `AgentDocumentVirtualList.tsx`

Belt-and-suspenders: the narrowed accessor form `{(p) => ...}` isolates `<Index>` in a child reactive scope so it is fully disposed before the outer `partition` memo can produce a stale read. The `<Show>` condition short-circuits on falsy to prevent the secondary `TypeError: Cannot read properties of undefined (reading 'streamingNodes')` that surfaces when `indexArray`'s disposed slot accessor is called after scope teardown. Since `partition()` always returns a truthy object, there is no visible behaviour change.

### Docs

- `docs/retro/RETRO_REPLACECHILD_CRASH_2026-06-06.md` — retro for the original crash class (the pre-existing `RETRO_REPLACECHILD_CRASH_2026-06-06.md` already existed; this adds it to the repo).
- `docs/specs/SPEC_REPLACECHILD_CRASH_FULL_ANALYSIS_AND_FIX_2026-06-06.md` — full root-cause analysis of all three observed crashes, resolved open questions (pre-crash unmount ruled out, WPS delivery confirmed synchronous, HistoryLoaded interleave mechanics, residual crash-2 mechanism with new diagnostic recommendation), complete fix table.

## Test plan

- [ ] Run a long multi-tool conversation (5+ turns, parallel tools) — pane must not crash
- [ ] Scroll up to trigger older-history load mid-turn — pane must not crash  
- [ ] Verify `[useAgentStream] subscribed` appears once after each turn completes (no error-boundary loop)
- [ ] Confirm `agentActivity busyCount` transitions correctly (no stuck "Working…")

### DIFF
--- a/.changesets/1780752741-fix-virt-defer-sessionend-batch-historyloaded-show-guard-str-hamm.md
+++ b/.changesets/1780752741-fix-virt-defer-sessionend-batch-historyloaded-show-guard-str-hamm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(virt): defer SessionEnd, batch HistoryLoaded, Show-guard streaming buffer
--- a/docs/retro/RETRO_REPLACECHILD_CRASH_2026-06-06.md
+++ b/docs/retro/RETRO_REPLACECHILD_CRASH_2026-06-06.md
@@ -1,0 +1,165 @@
+# Retro — `replaceChild` crash in the agent pane virtualizer
+
+**Date:** 2026-06-06  
+**Severity:** Hard blocker — agent pane blanks mid-conversation, repeatable  
+**Status:** Root cause confirmed; fix required (see §6)  
+**Related:** `docs/analysis/AGENT_PANE_REPLACECHILD_CRASH_ON_SEND_2026_05_27.md`, PR #1293
+
+---
+
+## 1. What the user experienced
+
+After 1–3 turns in an agent pane conversation the pane goes blank. The "Reload pane" fallback appeared (sometimes), or the pane just stayed dead. The agent's backend subprocess kept running — the conversation was preserved — but the UI was gone.
+
+The crash is **100% reproducible** once a conversation accumulates enough nodes (observed at `streamCount` 53–57, later at 85–86 after the partial fix).
+
+---
+
+## 2. Observed crash signatures
+
+**Primary crash** — every time, same stack:
+```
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
Fixed both P2s — `batch()` now wraps all three `dispatchDoc` call sites in `useHistoryPagination.ts`: `HistoryRestored` (snapshot restore, line 202), initial-load `HistoryLoaded` (NDJSON replay, line 274), and the `loadOlder` path (line 137). All four `documentAtom` writes from async callbacks in this file are now covered by the spec §5 invariant.

a5af:
@reagent review

a5af:
@reagent review

PR #1308, opened 2026-06-08
fix(floating-pane): restore JS-driven drag on macOS (BeginWindowDrag not in dev libcef)

a5af:
## Summary

- macOS libcef is unpatched — `begin_window_drag` was never compiled into dev builds (`patched-libcef` feature disabled for darwin)
- PR #1276 unified all platforms to use `start_window_drag` → `BeginWindowDrag`, silently breaking floater drag on macOS
- Restores the pre-#1276 JS-driven `get/set_window_position` polling behind an `isMacOS()` branch; Windows and Linux unchanged

## How it works

- `onMouseDown` (macOS): fetches window origin via `get_window_position`, arms `dragging=true` in `.then()`, race-guarded by `macMouseDownId`
- `onMouseMove` (macOS): computes delta from mousedown origin, calls `set_window_position` with a one-in-flight + coalesce guard (`macSendPos`)
- `onMouseUp`: increments `macMouseDownId` to cancel any in-flight IPC; existing `hasMoved` gate drives `tryRedockAtCursor` (redock still works)

## Test plan

- [ ] Tear off a pane into a floating window on macOS
- [ ] Drag the floating pane header — window follows cursor
- [ ] Release over another AgentMux window — redocks correctly
- [ ] Plain header click (no drag) — no false redock
- [ ] Verify Windows/Linux drag unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1249, opened 2026-06-03
fix(floating-pane): dwell + velocity gate prevents accidental redock on transit

Agent2-asaf:
## Problem

When dragging a floating pane across the screen, the cursor frequently *transits* over another AgentMux window on its way somewhere else. The dock indicator appears the instant the cursor enters, and if the user releases the mouse during that transit — even unintentionally — the pane docks.

User report (2026-06-03):
> "Sometimes when dragging the pane over a main window, I don't want it to dock, but it immediately does."

Root cause: every `mousemove` during a drag pushes `update_floating_redock_hover` (throttled only at 50 ms cadence), and `tryRedockAtCursor` on `mouseup` commits unconditionally on cursor-over-window — no intent gate.

## Approach

Researched industry patterns: dwell-time delay (Baymard, Windows hover default), velocity gate (USPTO 11163428), and explicit compass widgets (VS Code, Infragistics, Telerik, Dockview). Chose **dwell + velocity** — composes naturally, no new UI affordances. Compass widget is the natural next step if this proves insufficient.

## Change (single file: `frontend/app/workspace/floating-pane-workspace.tsx`)

Two-stage commitment:

1. **Armed**: dock indicator only appears after the cursor has been moving slowly (≤ 400 CSS-px/s) for ≥ 180 ms.
2. **Committed**: `mouseup` invokes `tryRedockAtCursor` *only if armed at release time*.

Velocity is sampled per-mousemove from the previous sample. Exceeding threshold resets the dwell timer and immediately fires `clear_floating_redock_hover` if the indicator was visible. Returning to slow motion restarts dwell from scratch.

### Tuning (local constants — easy to revisit)

- `REDOCK_DWELL_MS = 180` — below Windows' 400 ms hover default because drag-aiming feels more deliberate than passive hover.
- `REDOCK_VELOCITY_PX_PER_S = 400` — relaxed aim ≈ 100 px / 250 ms (= 400 px/s); transit drags hit 1500+ px/s.

## Behavior matrix

| User action | Old | New |
|---|---|---|
| Drag fast across main window → release elsewhere | Indicator flashes briefly | No indicator |
| Drag fast across main window → accidentally release while over it | **Docks** (the reported bug) | No dock — floater stays put |
| Drag slowly to aim → ~200 ms → release | Indicator + dock | Indicator + dock (unchanged) |
| Drag slowly to aim → see indicator → change mind, drag fast away | Indicator persists | Indicator clears within 1 frame |

## Spec

[...]

PR #1309, opened 2026-06-08
fix(linux): floater header drag — JS-driven so DOM events stay live (landing ghosts + redock)

AgentU-asaf:
## Root cause

`BeginWindowDrag` → `_NET_WM_MOVERESIZE` on Linux gives the X11 compositor exclusive input ownership for the duration of the drag. **DOM `mousemove` and `mouseup` events never fire inside the floater window.** This broke two things Linux-only:

1. **No landing ghosts** — `update_floating_redock_hover` is called from `onMouseMove`, which bails early when `!dragging`. But `dragging` was set back to `false` immediately when `window_drag_ended { moved: false }` fired synchronously right after `BeginWindowDrag()` returned (before the user even moved the mouse).
2. **No redock on drop** — `tryRedockAtCursor` is gated on `hasMoved`, which is set in `onMouseMove`. Since mousemove never fires during `_NET_WM_MOVERESIZE`, `hasMoved` stays `false`, and `tryRedockAtCursor` is never called from `onMouseUp`.

macOS works correctly because `performWindowDragWithEvent:` is a **blocking modal run loop** that still delivers DOM events to the renderer while the drag is active. The code had a `// F3 verification pending` comment for Linux — this PR confirms the events don't flow on Linux.

## Fix

On Linux, skip `start_window_drag` (which triggers `BeginWindowDrag`) and implement **JS-driven positioning** instead — the same mechanism already used for edge-resize, which works correctly on Linux:

1. `mousedown`: fetch the window's DIP origin via `get_window_rect`; save cursor anchor
2. `mousemove`: compute `new_pos = origin + (cursor − anchor)` → call `set_window_position`; also call `update_floating_redock_hover` (already there)
3. `mouseup`: `tryRedockAtCursor` fires normally since `hasMoved = true`

DOM events remain live throughout, so hover ghosts and redock both work. No Rust changes needed.

macOS is unchanged (still uses `BeginWindowDrag`). Windows is unchanged (still uses `Win32BeginMoveTask`).

## Test plan
- [ ] Linux: drag floater header → landing ghost appears on target window tiles ✓
- [ ] Linux: release floater over target window → pane redocks ✓
- [ ] Linux: drag floater to empty desktop → no false redock ✓
- [ ] macOS/Windows: floater drag unaffected (code path unchanged)
- [ ] Linux: floater edge-resize unaffected (separate pointer-capture path)

### DIFF
--- a/.changesets/1780885834-fix-linux-floater-header-drag-js-driven-positioning-so-dom-e-3nt9.md
+++ b/.changesets/1780885834-fix-linux-floater-header-drag-js-driven-positioning-so-dom-e-3nt9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(linux): floater header drag — JS-driven positioning so DOM events stay live (landing ghosts + redock)
--- a/frontend/app/workspace/floating-pane-workspace.tsx
+++ b/frontend/app/workspace/floating-pane-workspace.tsx
@@ -17,16 +17,20 @@
  *  - no extra title bar — the floater renders the block's standard
  *    `BlockFrame_Header` (33 CSS px, `--header-height` in
  *    `theme.scss:97`) as its sole chrome.
- *  - window drag is **host-driven**, installed by the `onMount` below:
+ *  - window drag is platform-specific, installed by the `onMount` below:
  *    a targeted document mousedown listener scoped to
- *    `[data-role="block-header"]` fires `start_window_drag` IPC on all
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentU-asaf:
Superseded by #1314. After main landed the macOS JS-driven floater drag (#1308) + dwell/velocity gate (#1249), the cleaner fix is to extend that proven path to Linux rather than hand-roll a separate one. #1314 does that.

PR #1311, opened 2026-06-09
fix(agent-pane): scroll race + ToolOverlayLog content-visibility layout warnings

a5af:
## Summary

- **`AgentDocumentVirtualList` — deferred measure RO (primary crash fix):** The measure ResizeObserver was dispatching `RowMeasured` synchronously inside its callback, which started a new Solid `runUpdates` frame while the previous `batch()`'s render effects were still executing. Two concurrent `reconcileArrays` passes on the same streaming-buffer `<Key>` parent → `replaceChild NotFoundError`. This is also what caused the 6× `ResizeObserver loop completed with undelivered notifications` bursts in the log (RowMeasured → layout height change → DOM mutation → new RO callbacks → tight sync loop). Fix: capture `getBoundingClientRect()` synchronously at RO time, accumulate pending measurements, dispatch all in the next RAF via a single `batch()`.

- **`AgentDocumentVirtualList` — `queueMicrotask` stickToBottom race:** The auto-scroll effect queued a `scrollTo(MAX_SAFE_INTEGER)` microtask without re-checking `stickToBottom`. If the user scrolled up between the effect firing and the microtask running, the scroll fired anyway, fighting their gesture during streaming. Fix: re-check `stickToBottom()` inside the microtask before scrolling.

- **`ToolOverlayLog` — content-visibility hidden subtree layout warnings:** The `requestAnimationFrame` scroll callback read `scrollRef.scrollHeight`/`scrollTop` even when the tool overlay panel was collapsed (`.agent-tool-panel--hidden` applies `content-visibility: hidden`). Fix: `MutationObserver` on the parent panel tracks the `--hidden` class; RAF callback skips layout access when hidden.

## Test plan

- [ ] Run an agent with several parallel/nested tool calls; scroll up mid-run — verify auto-scroll no longer fights the gesture
- [ ] Run a long agent session (50+ nodes) and watch DevTools console — no `ResizeObserver loop` warnings, no `replaceChild` errors
- [ ] Collapse a tool block while its log is streaming — no `content-visibility` layout warnings in console
- [ ] Scroll up then back down to bottom — stickToBottom re-engages normally
- [ ] Run agent past 50 nodes to exercise cap-advance and stale-frontier paths without crash

### DIFF
--- a/.changesets/1780884564-fix-floating-pane-restore-js-driven-drag-on-macos-beginwindo-sj94.md
+++ b/.changesets/1780884564-fix-floating-pane-restore-js-driven-drag-on-macos-beginwindo-sj94.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): restore JS-driven drag on macOS (BeginWindowDrag not available in dev)
--- a/.changesets/1780984172-fix-agent-pane-guard-queuemicrotask-auto-scroll-against-stal-bat4.md
+++ b/.changesets/1780984172-fix-agent-pane-guard-queuemicrotask-auto-scroll-against-stal-bat4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): guard queueMicrotask auto-scroll against stale stickToBottom; fix ToolOverlayLog content-visibility layout warnings
--- a/frontend/app/view/agent/components/ToolOverlayLog.tsx
+++ b/frontend/app/view/agent/components/ToolOverlayLog.tsx
@@ -15,7 +15,7 @@
  * ANSI parsing lands in Phase γ (perf + worker offload) per the spec.
  */
 
-import { For, Match, Show, Switch, createEffect, type JSX } from "solid-js";
+import { For, Match, Show, Switch, createEffect, onCleanup, onMount, type JSX } from "solid-js";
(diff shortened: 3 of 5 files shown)


a5af:
@codex review

a5af:
Closing as **superseded** — `main` has overtaken this branch and it now conflicts on all three files (`mergeable: dirty`).

The substantive fixes have each landed independently:

- **Floating-pane JS-driven drag** (the +129/-23 in `floating-pane-workspace.tsx`, changeset `1780884564…`) is the same fix already merged as #1308, with #1314 (Linux) and #1249 (dwell/velocity gate) on top.
- **queueMicrotask `stickToBottom` re-check** is already present on main (`AgentDocumentVirtualList.tsx` re-checks `stickToBottom()` inside the microtask).
- **`replaceChild` crash class** (the primary crash this PR targeted) was closed via a different, broader approach: #1319, #1326/#1327 (Key in virt list), and #1322 (Index in ChunkList).

Two minor, non-crash items are not on main yet but arent worth reviving this conflicted branch for: the `ToolOverlayLog` guard for `content-visibility: hidden` reads, and deferring the measure-RO `RowMeasured` dispatch to RAF/batch (the benign *ResizeObserver loop* console warning). If we want those, theyre a small fresh PR against current main.

PR #1314, opened 2026-06-09
fix(linux): floater drag uses JS-driven positioning so redock works (mirrors macOS #1308)

AgentU-asaf:
## Symptom
On Linux, floating-pane **redock doesn't work** — no landing ghosts when dragging a floater over a window, and dropping doesn't re-dock the pane.

## Root cause
The Linux floater header drag used `CefWindow::BeginWindowDrag` → `_NET_WM_MOVERESIZE`, which hands **all input to the X11/Wayland compositor** for the duration of the drag. The renderer receives **no DOM `mousemove`/`mouseup`**, so:
- `update_floating_redock_hover` never fires → no landing ghosts
- the `hasMoved` gate never trips → `tryRedockAtCursor` never runs → no redock

## Fix
macOS already solved this exact problem in **#1308** (it falls back to JS-driven `get/set_window_position` polling because `BeginWindowDrag` isn't in dev libcef). The mechanism is identical to what Linux needs, and it already carries the dwell + velocity gate from **#1249**.

This PR extends that proven path to Linux via a shared `JS_DRIVEN_DRAG = isMacOS() || isLinux()` gate. The window follows the cursor through `set_window_position`, which keeps the cursor over the header so DOM `mousemove` keeps firing (self-sustaining — no pointer capture needed), and the hover + redock logic runs exactly as on macOS.

Windows is untouched (`Win32BeginMoveTask` emits hover host-side while the renderer is dark).

## Why not #1276's BeginWindowDrag (smooth native drag)?
Keeping BeginWindowDrag would require a host-side move loop on Linux (like Windows' `Win32BeginMoveTask`) to poll the cursor + emit hover while the renderer is dark — a much larger change (X11/Wayland move loop in Rust). JS-driven is the pragmatic, proven choice; macOS lives with the same trade-off.

## Supersedes
Closes #1309 — that PR hand-rolled a separate Linux JS-drive before main landed the macOS path (#1308). This reuses the proven path and inherits the dwell/velocity gate.

## Test
- [ ] Linux: tear off a pane, drag the floater over a window → landing ghost appears on tiles
- [ ] Linux: drop over a tile → pane redocks
- [ ] Linux: fast transit over a window → no accidental redock (velocity gate)
- [ ] macOS/Windows: unchanged

### DIFF
--- a/.changesets/1781005247-fix-linux-floater-drag-uses-js-driven-positioning-mirrors-ma-qirn.md
+++ b/.changesets/1781005247-fix-linux-floater-drag-uses-js-driven-positioning-mirrors-ma-qirn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(linux): floater drag uses JS-driven positioning (mirrors macOS) so redock hover + drop work
--- a/frontend/app/workspace/floating-pane-workspace.tsx
+++ b/frontend/app/workspace/floating-pane-workspace.tsx
@@ -172,18 +172,18 @@ function FloatingPaneWorkspaceElem(): JSX.Element {
         // tryRedockAtCursor in onMouseUp so a plain header click (no pixel
         // motion) doesn't false-redock when the floater overlaps another window.
         let hasMoved = false;
-        // macOS JS-driven drag state: macOS libcef lacks the patched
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1315, opened 2026-06-09
fix(launcher): rename dev-portable channel to local; fix per-build pipe isolation

AgentA-asaf:
## Summary

- **Rename `dev-portable-*` → `local-*`**: the old name conflated `task dev` (hot-reload runtime mode) and portable (ZIP format). `local-<branch>-<hash>` accurately describes the scope — a locally-built binary — without implying build format or runtime mode. Reclaims 7 chars of the 64-char channel-name cap; branch slug grows from ≤20 to ≤27 chars.
- **Fix per-build pipe isolation**: two successive `task package` runs at the same semver shared the same single-instance pipe key, so the second build joined the first's running window instead of starting fresh. The launcher now uses `AGENTMUX_BUILD_LABEL` (which includes a per-build timestamp stamp) as the pipe key when present. Session data dir is unchanged (keyed on channel+semver), so agents/auth carry over between rebuilds.
- New test: `successive_local_builds_produce_different_hashes`.
- Retro: `docs/retro/retro-local-build-isolation-regression-2026-06-09.md`.

## Test plan

- [ ] `cargo test -p agentmux-launcher -p agentmux-common` — all pass (7 hash tests, 76 data_paths tests)
- [ ] `task package` produces a folder/channel named `local-<branch>-*` not `dev-portable-*`
- [ ] Two successive `task package` runs while first is running: second starts its own window, not a second window in the first instance


a5af:
@codex review

PR #1316, opened 2026-06-09
fix(cef): repair inert Resume button on low-memory pause page

a5af:
## Problem

A long-lived AgentMux window was OOM-paused (host hit 99% commit-limit exhaustion) and showed the gated-recovery **"Paused — system memory low"** page. **Clicking Resume did nothing**, even after memory was freed — the window was permanently wedged. Recovery required driving the renderer back to the app URL by hand over the Chrome DevTools Protocol.

## Root cause

`memory_paused_page()` (`agentmux-cef/src/client/mod.rs`) built the button as:

```rust
<button class="primary" onclick="location.href = {app_url_js}">Resume</button>
```

`js_string_literal()` returns a **double-quoted** JS string and is a `<script>`-context literal (it escapes `<`/`>`/`&` for `</script>` defense). Dropped into a **double-quoted `onclick="…"` attribute**, the quotes collide — the attribute terminates at the first inner quote, leaving `onclick="location.href = "` (a no-op) and turning the URL into stray markup:

```html
<button class="primary" onclick="location.href = "http://127.0.0.1:63627/?ipc_port=63627&ipc_token=…">Resume</button>
```

So the gated-recovery pause page has **never** been resumable via its button — any low-memory renderer OOM wedges the window until the process is killed.

## Fix

Attach the handlers in a `<script>` block via `addEventListener` (the context `js_string_literal` is designed for), keyed by button id. Same UX, quoting-safe. Adds a pure-string regression test (`memory_paused_page_resume_button_has_working_handler`) asserting the inert `onclick="location.href` antipattern is gone and the URL is emitted as a valid JS literal.

## Test

`cargo test -p agentmux-cef gated_recovery_tests` → 5 passed (incl. the new test). `cargo check -p agentmux-cef --tests` clean.

## Scope / follow-up

This restores the **manual** one-click recovery. Memory-gated **auto-resume** (so no human/agent is ever needed) is SPEC_GATED_RENDERER_RECOVERY **Phase 1b** — fully specified in `docs/recovery/FIX_MEMORY_PAUSE_RESUME_BUTTON_2026_06_09.md`, to land as a separate PR (it needs `Arc<AppState>` wiring into the memory heartbeat + a host build to validate threading).

> Review note: worth a glance that the crash-recovery / give-up pages don't construct buttons the same way.

### DIFF
--- a/.changesets/1781012789-fix-cef-low-memory-resume-button-inert-onclick-agc1.md
+++ b/.changesets/1781012789-fix-cef-low-memory-resume-button-inert-onclick-agc1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): low-memory pause page Resume button was inert (double-quoted JS string inside a double-quoted onclick attribute) — wire it via a <script> + addEventListener so OOM-paused windows can actually resume
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -2161,11 +2161,19 @@ window resumes.</p>
 other apps — then click Resume. Resuming before memory recovers will just
 pause again.</p>
 <div class="actions">
-<button class="primary" onclick="location.href = {app_url_js}">Resume</button>
-<button onclick="window.close()">Quit this window</button>
+<button id="amx-resume" class="primary">Resume</button>
+<button id="amx-quit">Quit this window</button>
 </div>
 <div class="footer">error_code={error_code} · commit_free={commit_free_mb}MB</div>
 </div>
+<script>
+(function(){{
+  var r = document.getElementById('amx-resume');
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
Thanks @reagentx-workflow — good catch. Fixed the identical bug at `commands/window/creation.rs` in 91c14ba0:

- Extracted a testable `assets_missing_html()` and moved both buttons ("Copy path" / "Quit") into a `<script>` block via `addEventListener`, matching the pause-page fix.
- Added regression test `copy_path_button_has_working_handler`.
- Confirmed via `grep -rn 'onclick="' agentmux-cef/src/ | grep '{'` → **no remaining** `onclick="…{js_string_literal}…"` call sites in the CEF host.

`cargo test -p agentmux-cef` passes (new test + existing `gated_recovery_tests`). PR now covers both inert-button instances of this quoting bug.

@codex review

PR #1320, opened 2026-06-09
feat(agent-pane): hold expanded tool on hover; expand Glob results by default

AgentA-asaf:
## Summary

- **Interaction hold**: When a tool block is auto-expanded (running or post-completion hold) and the user's mouse enters it, the block stays open until they leave — the post-completion timer can no longer collapse it mid-read. Hover on already-collapsed tools is unchanged (no peek, per `SPEC_TOOL_HOVER_CONSOLIDATION_2026_05_28`).
- **Glob auto-expand**: Glob tool results now start expanded so the full file list is visible immediately without requiring a chevron click.

## Files changed

- `frontend/app/view/agent/components/ToolBlock.tsx` — `userHolding` signal latched on `mouseEnter` while `autoExpanded() || pinned`, cleared on `mouseLeave`; included in `expanded()` guard
- `frontend/app/view/agent/components/CompactResult.tsx` — `createSignal(tool === "Glob")` instead of `createSignal(false)`
- `docs/specs/SPEC_TOOL_BLOCK_INTERACTION_HOLD_AND_GLOB_EXPAND_2026_06_09.md` — spec documenting both changes

## Test plan

- [ ] Run a tool call, let it complete — hover inside the block while the 3s hold is counting down, confirm it stays open after the timer fires, confirm it collapses when mouse leaves
- [ ] Hover over a collapsed (already-completed) tool — confirm it does NOT open (hover-to-peek still disabled)
- [ ] Trigger a Glob tool call — confirm file list is expanded on arrival without clicking
- [ ] Click the Glob chevron — confirm it can still be manually collapsed
- [ ] Pin a tool block — confirm pin still overrides everything

### DIFF
--- a/.changesets/1781046953-fix-agent-pane-hold-expanded-tool-on-hover-expand-glob-resul-n6gb.md
+++ b/.changesets/1781046953-fix-agent-pane-hold-expanded-tool-on-hover-expand-glob-resul-n6gb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): hold expanded tool on hover; expand Glob results by default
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #1319, opened 2026-06-09
fix(agent-pane): Send Now panel always visible + replaceChild crash on cap-advance

AgentA-asaf:
## Summary

Fixes two related crash scenarios in the agent pane:

- **Bug A — Send Now panel absent during stream stall**: `<PendingMessagesPanel>` outer `<Show>` was gated on `interruptibleTurn`, which is `false` during `Done.errored`/stream-stall. Removed the gate; zone shows whenever `pendingMessages().length > 0`.

- **Bug B — `replaceChild: node not a child` on streaming buffer** (commit 1): `<Key by={n=>n.id}>` in the streaming buffer calls `keyArray`, which disposes departing slots synchronously while `reconcileArrays` still holds stale element refs. Rapid consecutive cap-advances (>50 nodes) exploit this window. Switched to `<Index>` — position-keyed, cap-advances are pure signal updates, crash is structurally impossible.

- **Bug C — same `replaceChild` crash in virtual head** (commit 2): The virtual head container (`windowedRows()`) had the identical `<Key by={r=>r.nodeId}>` pattern. Rapid window shifts (stick-to-bottom + streaming) trigger the same disposal-before-reconcile race. Switched to `<Index>` with a reactive `elNodeId` update so the ResizeObserver dispatches `RowMeasured` for the correct nodeId after a window shift. After this commit `@solid-primitives/keyed` is no longer imported in `AgentDocumentVirtualList.tsx`.

Also adds a `prevNodeId` guard in `ToolBlock` to reset `postCompletionHold` state when `<Index>` slot-position reassignment places a different node at the same slot.

## Root cause (shared)

`<Key>` is designed for identity-stable lists. Both the streaming buffer and virtual head are positional sliding windows. `<Key>` vs `<Index>` mismatch is the crash root cause; `<Index>` fixes both.

## Test plan

- [ ] Start an agent with a long task; let the streaming buffer hit 50+ nodes; observe no crash
- [ ] Trigger a Send Now scenario (stream stalled with a queued message); verify panel appears
- [ ] Scroll a long agent conversation to exercise virtual head window shifts; no crash
- [ ] TypeScript check: no new errors in changed files

Fixes #1317.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #1322, opened 2026-06-10
fix(agent-pane): Index in ChunkList — close last replaceChild crash site (§7.4)

AgentA-asaf:
## Problem

`v0.43.1` still crashes with `replaceChild: node not a child` on block `aa6a64f`, even after the `<Key>→<Index>` fix from #1319. Render trail: `StreamFlushObserved (virtCount=102→103)` → `StreamUnsubscribe` (post-crash error-boundary cleanup).

The `<Index>` fix in #1319 eliminated `reconcileArrays` from the outer streaming-buffer and virtual-head containers. But a **nested `<For>`** inside `ChunkList` (ToolOverlayLog) was still reachable:

```tsx
// ToolOverlayLog.tsx — ChunkList
<For each={capped().chunks}>
    {(chunk) => <pre ...>{capChars(chunk.content)}</pre>}
</For>
```

When the `<Index>`-keyed streaming buffer shifts slot N to a **different ToolNode** (e.g. the frontier advances because the user sent a new message), `props.chunks` changes to the new node's chunk array. `mapArray` (used by `<For>`) keys by item reference — the entire incoming array is "new refs", so SolidJS calls `reconcileArrays(parent, old_pre_elements, new_pre_elements)`. If the `<Switch>` branch in `ToolOverlayLog` also changed in the same reactive batch (e.g. Match1→Match4 because the new node has no log yet), the old `<pre>` elements can already be detached before `reconcileArrays` runs → `replaceChild` throws.

This is the §7.4 crash class from `SPEC_REPLACECHILD_CRASH_FULL_ANALYSIS_AND_FIX_2026-06-06.md`, previously marked "mechanism unresolved — new diagnostic needed".

## Fix

Replace `<For>` with `<Index>` in `ChunkList`:

```tsx
<Index each={capped().chunks}>
    {(chunk) => (
        <pre class={`agent-tool-log-line ${KIND_CLASS[chunk().kind] ?? ""}`}>
            {capChars(chunk().content)}
        </pre>
    )}
</Index>
```

`<Index>` uses `indexArray` internally:
- **Same-length updates** (streaming appends within same node): slot signals update in-place → **no `reconcileArrays`**
- **Length-change updates** (new node with different chunk count): `indexArray` appends/removes from the tail only — all existing `<pre>` elements remain in their parent → **no stale DOM**

This is the same structural fix applied to `AgentDocumentVirtualList.tsx` in #1319, extended to the one remaining `<For>` in the document-rendering hot path.

## Why this is the last crash site

After this PR, there are **no remaining reactive `<For>` components** in the document rendering chain (streaming buffer → virtual head → DocumentRow → ToolBlock → ToolOverlayLog). All other `<For>` uses in the agent view are in modal/panel components outside the rendering hot path, or use static arrays.

## Test plan

- [ ] Build and launch `task package`
- [ ] Open an agent with a long conversation (150+ nodes) that has active tool streaming
- [ ] Send a new message while the tool is streaming — triggers the `virtCount` window shift that caused the crash
- [ ] Verify: no `replaceChild` error boundary, chunks continue rendering correctly in the overlay
- [ ] Verify: streaming chunks still display (no regression on ChunkList rendering)

Follow-up to #1319 — closes §7.4 of the replaceChild crash spec.

### DIFF
--- a/.changesets/1781088143-fix-agent-pane-replace-for-with-index-in-chunklist-to-close--tir5.md
+++ b/.changesets/1781088143-fix-agent-pane-replace-for-with-index-in-chunklist-to-close--tir5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): replace For with Index in ChunkList to close the last replaceChild crash site
--- a/frontend/app/view/agent/components/ToolOverlayLog.tsx
+++ b/frontend/app/view/agent/components/ToolOverlayLog.tsx
@@ -15,7 +15,7 @@
  * ANSI parsing lands in Phase γ (perf + worker offload) per the spec.
  */
 
-import { For, Match, Show, Switch, createEffect, type JSX } from "solid-js";
+import { Index, Match, Show, Switch, createEffect, type JSX } from "solid-js";
 // `Show` retained for fallback ToolOverlayResult sub-tree.
 import type { ToolNode } from "../types";
 import { BashOutputViewer } from "./BashOutputViewer";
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1323, opened 2026-06-10
feat(scripts): import-agents wires real Claude sessions so conversations resume

AgentA-asaf:
## Problem

`scripts/import-agents.sh` copied agent **definitions** + empty-session stubs, so imported agents showed only the session-context line — no conversation. The transcript is **not** in `objects.db`: it lives in the provider CLI's account-wide session file (`~/.claude/projects/<slugified-cwd>/<session_id>.jsonl`), replayed when the agent re-spawns with `--resume <session_id>` (`agentmux-srv/.../subprocess.rs`).

## Change

Wire each imported instance to its **largest existing on-disk session** so a single message resumes the real (long) conversation — no transcript bytes are copied:

- resolve `~/.agentmux/agents/<slug>-*` working dirs, slugify each to its Claude project dir (`: \ . /` → `-`), pick the largest `.jsonl` (longest transcript = best repro)
- set `session_id` + `working_directory` on `db_agent_instances` (and mirror the cwd into `db_agents` for a fresh launch)
- **backfill already-imported stubs**: the "definition already present" case no longer early-skips — re-running wires sessions onto prior imports. Every copy is `INSERT OR IGNORE`, so re-running is a harmless no-op.

Agents with no on-disk session import exactly as before (empty stub) — the import never fails on a session-less agent.

## Test plan

- [x] `bash -n` syntax check
- [x] Ran against a fresh test channel: Mopeo wired to its 1.1 MB session, Maksi to its 718 KB session; verified `session_id` + `working_directory` populated in `db_agent_instances`
- [ ] Send an imported agent a message → confirms `--resume` replays the conversation in the pane

### DIFF
--- a/.changesets/1781088104-feat-scripts-import-agents-wires-real-claude-sessions-so-imp-d9fa.md
+++ b/.changesets/1781088104-feat-scripts-import-agents-wires-real-claude-sessions-so-imp-d9fa.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(scripts): import-agents wires real Claude sessions so imported agents resume their conversation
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

Issue #1317, opened 2026-06-09
Agent pane: Send Now panel absent + replaceChild crash when sending mid-stream


AgentA-asaf:
## Summary

Two related failures when the user sends a message while an agent is actively streaming a response:

1. **Send Now panel never appears** — the pending message enters a limbo state with no UI feedback
2. **`replaceChild` DOM crash** — after the pending message expires (~30s) and the stream completes, SolidJS `reconcileArrays` throws, collapsing the agent pane to the error boundary

## Reproduction steps

1. Open an agent pane with a long-running conversation (virtCount ≥ ~50 items)
2. Trigger a response that streams for several seconds
3. While the agent is mid-stream, send a new message
4. **Expected:** "Send Now" interrupt panel appears below the input
5. **Actual:** Nothing appears; agent continues streaming but stops appending new items
6. Wait ~30s → `PendingMessageExpired` fires
7. When the stream eventually completes and `StreamUnsubscribe` fires → crash

## Log evidence

```
[17:47:38] ERROR [block-error-boundary] NotFoundError: Failed to execute 'replaceChild' on 'Node':
           The node to be replaced is not a child of this node.
           (block=aa6a64f7-c8b2-4bb9-bb98-f3bf591de4ab, view=agent)

Stack (resolved):
  at reconcileArrays    (solid-js/web/dist/web.js:593:13)
  at insertExpression   (solid-js/web/dist/web.js:568:16)
  at runComputation     (solid-js/dist/solid.js:677:2)
  at createRenderEffect (solid-js/web/dist/web.js:565:4)
  at insertExpression   (solid-js/web/dist/web.js:330:32)
  at runComputation     (solid-js/dist/solid.js:677:2)
```

Render trail key events:
```
ToolStart              → frontier=node_26,  streamCount=50, virtCount=54
StreamFlushObserved    → frontier=user_178, streamCount=50, virtCount=55
PendingMessageQueued   → user sent message mid-stream
PendingMessageExpired  → t+30s, message expired with no Send Now shown
TokensIn               → stream response arrived post-expiry
StreamFlushObserved    → frontier=node_27,  streamCount=50, virtCount=56
StreamFlushObserved    → frontier=node_28,  streamCount=50, virtCount=57
StreamUnsubscribe      → CRASH
```

## Root cause hypothesis

[...]

PR #1327, opened 2026-06-10
fix(agent-pane): close replaceChild crash class — <Key> in virt list, <Index> in DiffViewer (#1326)

### LINKED ISSUE #1326: agent-pane: replaceChild crash persists in v0.44.0 (block e913dc1) — #1322 For→Index was incomplete, §7.4 mechanism still unresolved
## Summary

The `replaceChild` SolidJS reconciliation crash **still occurs in v0.44.0**, the release that already contains the `<For>`→`<Index>` ChunkList fix (#1322). So #1322 removed one `reconcileArrays` site but did **not** close the crash class — the mechanism documented as unresolved in §7.4 of `docs/specs/SPEC_REPLACECHILD_CRASH_FULL_ANALYSIS_AND_FIX_2026-06-06.md` is still live.

## Error (observed on a v0.44.0 dogfood instance, Windows)

```
NotFoundError: Failed to execute 'replaceChild' on 'Node':
The node to be replaced is not a child of this node.
block e913dc1 · view agent

    at xbe (index-C97xjv_D.js:6:17326)
    at jp  (index-C97xjv_D.js:6:23199)
    at Object.fn (index-C97xjv_D.js:6:22830)
    at q0e (index-C97xjv_D.js:5:23957)
    at Fm  (index-C97xjv_D.js:5:23882)
    at G   (index-C97xjv_D.js:5:20086)
    at jp  (index-C97xjv_D.js:6:22778)
    at Object.fn (index-C97xjv_D.js:6:19663)
    at q0e (index-C97xjv_D.js:5:23957)
    at Fm  (index-C97xjv_D.js:5:23882)
```

- Build: **v0.44.0** release (`index-C97xjv_D.js`), `127.0.0.1:64282`.
- Block: **`e913dc1`** (a different block than the earlier `aa6a64f`).
- The frame shape (`q0e`→`Fm`→`G` nested twice) is the same two-level `insertExpression` → `reconcileArrays` → `replaceChild` pattern as before.

## Why the stack can't be resolved as-is

The release build is produced by `task package:release`, which **strips source maps** (`STRIP_MAPS=1`). So the host can't resolve the stack and it stays minified. The diagnostic value is the **`render_trail`** (reducer-action sequence) in the host log, which is still emitted regardless of maps:
`~/.agentmux/channels/stable/versions/0.44.0/logs/agentmux-host-*.log*`

## History

- #1317 (closed) — original manifestation: Send Now panel + replaceChild mid-stream
- #1319 — Send Now always-visible + replaceChild on cap-advance (virtual-head/streaming-buffer `<Key>`→`<Index>`)
- #1322 — `<For>`→`<Index>` in `ChunkList` (ToolOverlayLog) — **this release; insufficient**

## Remaining suspects (now that ChunkList is `<Index>`)

- `DiffViewer`'s `<For each={lines}>` (Edit/diff tool results) — `lines` is recomputed in `PlainDiff` on each render; a Show-branch flip could detach its `<pre>`s mid-reconcile.
- A virtual-head / streaming-buffer `<Index>` boundary under a transition #1319/#1322 didn't cover (e.g. a length-change flush coinciding with a `<Switch>` branch change in `ToolOverlayLog`).
- Any remaining reactive `<For>` reachable from the document render tree.

## Next steps

1. Implement the §7.4 diagnostic: instrument the `reconcileArrays` caller site to log `a.length`/`b.length` and whether each `a[i]` is still a child of `parent` before the call — the next crash then names the offending node + the flush that detached it.
2. Reproduce on a **maps-included** local build (`task package`, not `package:release`) so the host resolves the stack and the `block-error-boundary` frames are legible.
3. Capture the `render_trail` from the crashing instance to confirm whether the trigger is the same streaming window-shift (`virtCount` transition) as #1319/#1322.

cc: spec `docs/specs/SPEC_REPLACECHILD_CRASH_FULL_ANALYSIS_AND_FIX_2026-06-06.md` §7.4

AgentX-asaf:
## Summary

Closes the remaining `replaceChild NotFoundError` crash sites identified in issue #1326 (still occurring in v0.44.0 after #1322).

Three sites fixed:

### 1. Virtual head — `<Index>` → `<Key by={r => r.nodeId}>`

`<Index>` (position-keyed) required a `createEffect` to keep `elNodeId` current whenever the scroll window shifted a different node into the same slot. Without it, `measureRO` would dispatch `RowMeasured` for the wrong nodeId. `<Key>` ties each DOM slot to a nodeId for life — no extra effect needed, no stale-ref window.

### 2. Streaming buffer — `<Index>` → `<Key by={n => n.id}>`

The root crash cause was **never keying strategy** — it was **array growth**. `reconcileArrays` corrupts its sentinel tracking when the array grows past the initial render size, regardless of whether `<For>`, `<Index>`, or `<Key>` is used. The cap-advance in the partition memo already keeps `streamingNodes.length ≤ STREAMING_BUFFER_SIZE`, so `reconcileArrays` always sees a fixed-size array. With that guarantee in place, `<Key by={n => n.id}>` is the right primitive:
- No remount on token updates (same id, new object ref)
- No position-slot state leaks across cap-advance migrations (`prevStatus`, `expanded` flags)

The comments in #1317/#1319 explaining why `<Index>` was chosen over `<Key>` were based on an incomplete model of the crash mechanism; this PR updates them.

### 3. DiffViewer — `<For>` → `<Index>` + memoize `lines`

`PlainDiff` used `<For each={lines}>` where `lines = parseDiffLines(diff()!)` was recomputed on every reactive sweep (new array ref = `<For>` treats all items as changed). A `<Show>` branch flip when Shiki resolves can race with a second `reconcileArrays` pass on the same `<pre>` DOM nodes → crash. Fix:
- `createMemo` wrapping `parseDiffLines(diff())` so the array is only recomputed when `diff()` actually changes
- `<Index>` (position-keyed) instead of `<For>` so each position slot updates in place; no DOM moves when the diff is stable

## Crash history
- #1317 — original: Send Now panel + replaceChild mid-stream
- #1319 — cap-advance on virtual head (streaming-buffer `<Key>`→`<Index>`)
- #1322 — ChunkList `<For>`→`<Index>` in ToolOverlayLog
- **#1326 (this PR)** — remaining sites: virt head, streaming buffer, DiffViewer

## Test plan
- [ ] Run a long agent session (50+ nodes, multiple Edit tool calls) — pane must not crash
- [ ] Trigger a cap-advance: start a session, let it exceed 50 streaming nodes, verify no crash
- [ ] Open a file with Edit tool result containing a large diff — DiffViewer must render without crash
- [ ] Scroll up to trigger virtualized head, then back down — no crash, rows correct

Closes #1326

### DIFF
--- a/.changesets/1781093156-fix-agent-pane-replace-remaining-index-with-key-in-virt-list-vxq9.md
+++ b/.changesets/1781093156-fix-agent-pane-replace-remaining-index-with-key-in-virt-list-vxq9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): replace remaining <Index> with <Key> in virt list and <For> with <Index> in DiffViewer to close replaceChild crash class (#1326)
--- a/frontend/app/view/agent/components/DiffViewer.tsx
+++ b/frontend/app/view/agent/components/DiffViewer.tsx
@@ -20,8 +20,7 @@
  * output.
  */
 
-import { createEffect, createMemo, createSignal, onCleanup, Show, type JSX } from "solid-js";
-import { For } from "solid-js";
+import { createEffect, createMemo, createSignal, Index, onCleanup, Show, type JSX } from "solid-js";
 import type { EditParams, EditResult } from "../types";
 import { detectLanguage } from "./detectLanguage";
 import { OutputHiddenMarker } from "./OutputHiddenMarker";
@@ -188,20 +187,30 @@ export const DiffViewer = (props: DiffViewerProps): JSX.Element => {
         );
     }
 
+    // Memoized so parseDiffLines doesn't re-run on every reactive sweep
+    // and so <Index> sees a stable array reference between renders.
+    const lines = createMemo(() => {
+        const d = diff();
+        return d ? parseDiffLines(d) : [];
+    });
+
     // --- Plain fallback (shown until Shiki resolves, or permanently if Shiki fails) ---
-    const PlainDiff = () => {
-        const lines = parseDiffLines(diff()!);
-        return (
-            <pre class="agent-diff">
-                <div class="agent-diff-header">{filePath()}</div>
-                <For each={lines}>
-                    {(line) => (
-                        <div class={`agent-diff-${line.type}`}>{line.raw}</div>
-                    )}
-                </For>
-            </pre>
-        );
-    };
+    // Uses <Index> (position-keyed) rather than <For> (reference-keyed).
+    // Diffs are immutable once rendered, but a <Show> branch flip when
+    // Shiki resolves can race with a reactive update that drives a
+    // second reconcileArrays pass on the same DOM nodes → replaceChild
+    // NotFoundError. <Index> avoids that: each position slot updates
+    // in place; no DOM moves occur when the array is stable. (#1326)
+    const PlainDiff = () => (
+        <pre class="agent-diff">
+            <div class="agent-diff-header">{filePath()}</div>
+            <Index each={lines()}>
+                {(line) => (
+                    <div class={`agent-diff-${line().type}`}>{line().raw}</div>
+                )}
+            </Index>
+        </pre>
+    );
 
     return (
         <>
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1330, opened 2026-06-10
fix(term): reclaim dead space right of terminal text at all zoom levels

AgentX-asaf:
## Problem

A dead band (up to ~22px at DPR 1.25) sat between the last column of terminal text and the pane edge, and its **width changed with zoom level** — thin at some zooms, wide at others, never zero.

## Root cause (verified live over CDP)

Full write-up: `docs/analysis/ANALYSIS_TERM_SCROLLBAR_GAP_ROOTCAUSE_2026_06_10.md` (+ spec `docs/specs/SPEC_TERM_SCROLLBAR_ZERO_GAP_2026_06_10.md`).

- **FitAddon reserves a fixed 14px** scrollbar lane whenever `scrollback > 0` — *even when no scrollbar is showing*. That always-reserved-but-empty lane is the dead band.
- The previous correction added `floor(14 / cellWidth)` columns back. That's **0 whenever `cellWidth > 14`** (zoomed in / higher DPR) and **1** otherwise — the quantized on/off flip is exactly the zoom-dependent width.
- **`overflow: overlay` is a no-op** — Chromium deprecated it and aliases it to `auto` (confirmed: it still steals 14px when a scrollbar is present).
- The gap is already painted the terminal background color (`rgb(15,15,16)` = text bg) by `.block-frame-default-inner`.

## Fix

- **`termwrap.ts` `customFit()`**: recompute `cols` from the real container width in pixel space, reserving the 14px lane **only when a scrollbar is actually present**:
  - no scrollbar → reserve 0 → grid fills full width (gap → 0)
  - scrollbar present → reserve 14 → the scrollbar fills its lane (gap → 0)
  - the residual sub-cell remainder (`< cellWidth`, irreducible — no partial columns) is background-colored, so it is invisible at every zoom.
- **`setupScrollbarRefit()`**: `ResizeObserver` on the viewport re-fits when the scrollbar toggles. (xterm v6's DOM renderer scrolls virtually and never shows a native scrollbar; the WebGL path scrolls natively — the conditional reservation is correct for both.)
- **`term.scss`**: removed the no-op `overflow: overlay`; set `overflow-y: auto` so the lane is conditional; kept the transparent track.
- Removed the obsolete `FIT_WIDTH_CORRECTION` constants.

## Verified (live dev build, DPR 1.25)

| | before | after |
|---|---|---|
| `.xterm` width | 1040px (72 cols) | **1055px (73 cols)** |
| gap (`connect − xterm`) | **22px** dead | **7px** sub-cell, bg-colored |
| text reaches edge | no | **yes** (200-char lines run edge-to-edge) |
| wheel scrollback | works | **works** |
| last-column clipping | n/a | **none** |

A literal 0px gap is impossible (terminals render whole cells only — the sub-cell remainder is irreducible), but correct pixel-space reclaim + the existing background match makes the gap **visually zero at every zoom level**.

## Note (out of scope, worth tracking)

CDP probing found **WebGL is unavailable in CEF** (`getContext('webgl')` returns null) → every terminal silently uses the slower DOM renderer, which has **no native scrollbar at all** (scroll is wheel-only). Flagged in §7 of the analysis for a separate ticket.

## Test plan

- [ ] Open a terminal — text reaches the right edge at default zoom
- [ ] Ctrl+= / Ctrl+- through several zoom levels — no dead band appears at any
- [ ] Run output that overflows vertically — scroll works, last column not clipped
- [ ] Resize the pane slowly — column count steps cleanly, no contrasting strip
- [ ] On a machine where WebGL loads — scrollbar appears, reserves 14px, no clipping

### DIFF
--- a/.changesets/1781140153-fix-term-reclaim-dead-space-to-the-right-of-terminal-text-at-61h4.md
+++ b/.changesets/1781140153-fix-term-reclaim-dead-space-to-the-right-of-terminal-text-at-61h4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): reclaim dead space to the right of terminal text at all zoom levels
--- a/docs/analysis/ANALYSIS_TERM_SCROLLBAR_GAP_ROOTCAUSE_2026_06_10.md
+++ b/docs/analysis/ANALYSIS_TERM_SCROLLBAR_GAP_ROOTCAUSE_2026_06_10.md
@@ -1,0 +1,318 @@
+# Root-Cause Analysis: Terminal Right-Edge Gap (zoom-dependent width)
+
+**Status:** Complete — empirically verified against the live dev build via CDP
+**Author:** AgentX
+**Date:** 2026-06-10
+**Companion spec:** `docs/specs/SPEC_TERM_SCROLLBAR_ZERO_GAP_2026_06_10.md`
+**PR:** #1330 (`agentx/fix-xterm-dead-space`)
+**Method:** Chrome DevTools Protocol (port 9223, dev build PID 116308), live DOM
+measurement + diagnostic layer colorization + composited-pixel sampling + screenshots.
+
+---
+
+## 1. Executive summary
+
+The "dead space" to the right of terminal text is **22px** of empty `.term-connectelem`
+between the right edge of the `.xterm` element and its parent. It is composed of:
+
+- **~14px** — FitAddon's hard-coded scrollbar/overview-ruler reservation that PR #1330's
+  column correction **fails to reclaim** at most zoom levels (the dominant, fixable part).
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

Agent3-asaf:
@reagentx-workflow review

(Re-triggering: the original review on this PR completed but was silently dropped on a GitHub App token 401 during write-back — see a5af/reagent#171, fix in a5af/reagent#172.)

Agent3-asaf:
@reagentx-workflow review

AgentX-asaf:
@reagentx-workflow review

AgentX-asaf:
@reagentx-workflow review --force

AgentX-asaf:
@reagentx-workflow review --force

Agent3-asaf:
@reagentx-workflow review

PR #1331, opened 2026-06-10
fix(scripts): rehydrate Claude sessions into isolated home so resume replays (history P0)

AgentA-asaf:
## Problem

Imported agents showed an **empty pane**. `import-agents.sh` wired a `session_id` it found in the user's **global** `~/.claude/projects/`, but AgentMux spawns claude with `CLAUDE_CONFIG_DIR=~/.agentmux/shared/providers/claude`, so `claude --resume <sid>` reads from the **isolated** home and never finds the global file → no replay. (Root-caused in `SPEC_UNIFIED_AGENT_HISTORY_STORE_2026-06-10.md` §2.5.)

## Fix (P0 of the history-store spec)

- `best_session_for_slug` now searches **both** the isolated home (`shared/providers/claude/projects`) and global `~/.claude/projects`, and returns the source transcript path.
- New `rehydrate_claude_session` copies the chosen transcript into `$CLAUDE_CONFIG_DIR/projects/<slug(cwd)>/<sid>.jsonl` (no-op if already there) — so a single message replays the real conversation.

## Verified

Against the running 0.44.1 channel: Mopeo's 1.1 MB session (`4f03e3ed`, previously only in global `~/.claude`) is now copied into `shared/providers/claude/projects/.../4f03e3ed.jsonl`. Agents with no session import as before (empty stub).

## Also included

The full architecture spec `docs/specs/SPEC_UNIFIED_AGENT_HISTORY_STORE_2026-06-10.md` — verified current state (provider isolation, cwd-slug fragmentation, orphaned pre-isolation histories), and the P1–P3 plan (per-provider `HistoryAdapter`s, turn-end capture, `index.db` + History pane, organize/clear/GC). P0 marked done.

### DIFF
--- a/.changesets/1781112727-fix-scripts-import-agents-rehydrates-claude-sessions-into-th-jb9x.md
+++ b/.changesets/1781112727-fix-scripts-import-agents-rehydrates-claude-sessions-into-th-jb9x.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(scripts): import-agents rehydrates Claude sessions into the isolated CLAUDE_CONFIG_DIR home so resume actually replays the conversation (history store P0)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1332, opened 2026-06-10
fix(history): scan AgentMux-isolated Claude homes in history browse (P1a)

AgentA-asaf:
## Problem

`ClaudeHistoryAdapter` scanned only `~/.claude/projects` and `~/.config/claude-*/projects`. But AgentMux spawns claude with `CLAUDE_CONFIG_DIR=~/.agentmux/shared/providers/claude`, so **agent conversations write under the isolated home** — which the adapter never scanned. The result: the existing `HistoryService` browse (wired to `useHistoryPagination` / `parseHistoryLines` in the frontend) **missed every AgentMux-spawned agent's history**, showing only global sessions.

## Fix (P1a)

`ClaudeHistoryAdapter::new()` now also scans, via the launcher-exported `AGENTMUX_SHARED_DIR` (fallback `~/.agentmux/shared`):
- `<shared>/providers/claude/projects/` — default, account-wide
- `<shared>/identities/<bundle_id>/claude/projects/` — per-identity bundles

Existing scan roots and behavior are unchanged.

## Why this is the right first increment

A discovery/browse layer **already exists** — `history/{adapter,claude_adapter,index,mod}.rs` (the `HistoryAdapter` trait + `SessionIndex` + `HistoryService`), wired to RPC and the frontend. It just wasn't looking where AgentMux writes. This change makes it user-visible immediately, and the rest of the spec (P1b durable mirror/capture + clear, P1c Codex/Gemini, P2 pane UX) builds on the same infra rather than rebuilding it.

## Verification

`cargo check -p agentmux-srv` → clean (Finished, no errors). Spec updated (`SPEC_UNIFIED_AGENT_HISTORY_STORE_2026-06-10.md`) with the corrected phasing + the "implementation reality" note.

### DIFF
--- a/.changesets/1781114538-fix-history-scan-agentmux-isolated-claude-homes-so-the-histo-q373.md
+++ b/.changesets/1781114538-fix-history-scan-agentmux-isolated-claude-homes-so-the-histo-q373.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(history): scan AgentMux-isolated Claude homes so the history browse surfaces agent conversations, not just global ~/.claude (history P1a)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1333, opened 2026-06-10
feat(history): clear/delete past conversations (P1b)

AgentA-asaf:
## What

Adds the **"clear"** half of the history-store work (the explicitly-requested ability to clear/organize history):

- `SessionIndex::delete(session_id)` — removes the native transcript file **and** Claude's sibling `<session_id>/` subagents dir, then drops it from the index.
- `SessionIndex::clear(provider, project)` — bulk delete by filter (no filter = clear all). Snapshots ids under the lock, then does fs ops lock-free.
- `HistoryService::delete` / `clear` + RPC methods `history.Delete` / `history.Clear` (alongside the existing `history.List`/`Get`/`Refresh`).
- Unit tests for delete + filtered clear (mock adapter + temp files) — both green.

## Mirror DESCOPED (heads-up — deviates from the original P1b sketch)

The spec originally proposed re-copying every transcript into `shared/history/`. On inspection that's **low value for high cost**: provider histories already live **inside** AgentMux's home (`shared/providers/<p>/`, not GC'd by the CLIs), so a duplicate would double ~540 MB (Claude alone) for marginal durability. Instead — P1a made them browsable in place, this PR adds clear in place, P1c extends to more providers. The mirror's real sub-goals (orphan migration — already done on resume in P0; agent-anchored defragmentation) are better as targeted features. Rationale documented in the spec. **Shout if you'd rather I build the full mirror anyway.**

## Verify

`cargo check -p agentmux-srv` clean; `cargo test -p agentmux-srv history::index::` → 2 passed.


a5af:
@codex review

PR #1334, opened 2026-06-10
diag(agent-pane): name the component behind the replaceChild crash

AgentA-asaf:
## Why

The agent pane crashes with `replaceChild: node not a child` across v0.43.x–v0.44.1 (block `aa6a64f`, #1326). **Four keying fixes (#1319/#1322/#1327 + the Switch fix) did not close it**, and even with source maps the resolved JS stack is **100% SolidJS-internal** — the offending render effect runs detached from the component call stack, so the stack can never name the component. The crash is **keying-independent** (reproduced with both `<Index>` and `<Key>`), so the real mechanism is an inner render-effect racing the outer virt-window-shift, not a `For`/`Index`/`Key` choice.

## What

A transparent guard on `Node.replaceChild` that, the instant the node-not-a-child condition is true (one tick before SolidJS throws), logs the **parent container** + the **detached child**'s tag/class/`data-*` + **ancestry** + the parent's current children. `[fe]` logs are forwarded into the host log, landing right above the `block-error-boundary` `render_trail` — together they pinpoint the component and what shifted underneath it.

- Cost: one `parentNode` comparison per `replaceChild`; the heavy logging only runs on the anomaly (which always precedes a crash).
- Behaviour-preserving (calls through to native `replaceChild` unchanged) — safe to keep permanently; it'll also catch any future occurrence.
- Installed first in `app.tsx` so it's active before the agent pane mounts.

## Verify

`tsc --noEmit` clean for the new files. To exercise: reproduce the agent-pane crash and read `[replace-child-diag]` in the host log — it names the container + detached node.

### DIFF
--- a/.changesets/1781119327-diag-agent-pane-log-parent-container-detached-child-ancestry-9l76.md
+++ b/.changesets/1781119327-diag-agent-pane-log-parent-container-detached-child-ancestry-9l76.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+diag(agent-pane): log parent container + detached child + ancestry the instant replaceChild would throw, to finally name the reconcileArrays crash component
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #1336, opened 2026-06-11
fix(gpu): SwiftShader software-GL fallback on Windows + status-bar GPU indicator

AgentX-asaf:
## Problem

When Chromium's hardware GPU process can't initialize a GL device — **virtual display adapters (Parsec, RDP/headless), or broken drivers** — it crashes at init, and after 3 crashes Chromium disables GL entirely (`gl=disabled`). WebGL becomes unavailable and xterm **silently** falls back to its slower DOM renderer, which in xterm v6 has **no scrollbar** (wheel-only).

Full evidence-backed root cause (CDP `SystemInfo`, CEF debug log, `Win32_VideoController`): `docs/analysis/ANALYSIS_WEBGL_GPU_DISABLED_ROOTCAUSE_2026_06_10.md`. Confirmed on this machine across **both** the dev build and the downloaded portable v0.44.0 — it's environment + bundle, not a version mismatch (`libcef.dll` is correctly 148.0.9).

The AgentMux-side contributing factor: the **Windows bundle stripped SwiftShader**, so there was no software-GL fallback (macOS/Linux bundles keep it). Chromium went straight from "hardware GPU crashed" → "GPU fully off."

## GPU fix

- **`Taskfile.yml` `bundle:windows`** now ships `vk_swiftshader.dll`, `vulkan-1.dll`, `vk_swiftshader_icd.json` (parity with macOS/Linux). ~18 MB.
- **`agentmux-cef/src/app.rs`** appends `--enable-unsafe-swiftshader`. Chromium 110+ gates the SwiftShader-for-WebGL fallback behind this switch; without it, even with the DLLs present, `getContext('webgl')` returns null once hardware GPU is off. Hardware GL is still preferred when available → **fallback only, no cost on healthy machines**.

Net: a hardware-GPU failure degrades to **software WebGL** instead of disabling it — the xterm WebGL renderer (and its scrollbar) keep working on virtual-display / headless / broken-driver machines (the remote-agent environments AgentMux runs in).

## Make the degraded state visible

- **`frontend/util/gpuutil.ts`** — cached WebGL probe → `hardware` / `software` / `unavailable` + unmasked renderer/vendor.
- **`statusbar/GpuStatus.tsx`** — bottom-left `GPU HW/SW/off` line (color-coded) with a detail tooltip.
- **`BackendStatus` popover** (click the uptime / "time-up") gains a **GPU section**: WebGL class, renderer, vendor, and the terminal's actual renderer (WebGL vs DOM).
- **`termwrap.ts`** records the chosen renderer via a new global `termRendererAtom`.

## Verification

- ✅ Frontend typechecks (`tsc --noEmit`, no errors in changed files)
- ✅ `cargo check -p agentmux-cef` compiles
- ✅ SwiftShader DLLs present in the CEF runtime source, so the bundle copy is real (not a silent no-op)
- ⏳ End-to-end "WebGL returns under Parsec" needs a clean rebuild + relaunch (manual — the running dev instance is on another branch). Expected: `SystemInfo` `webgl` flips from `disabled_off` to software, status bar shows `GPU SW`.

## Test plan

- [ ] Build a Windows portable (`task package`) — confirm `vk_swiftshader.dll` + `vulkan-1.dll` are in `runtime/`
- [ ] Launch on a machine with a virtual display adapter (or after a GPU-process crash) — terminals render via WebGL (software), scrollbar present; status bar shows `GPU SW`
- [ ] On a healthy GPU machine — status bar shows `GPU HW`, hardware path unchanged (no perf regression)
- [ ] Click the uptime in the status bar — GPU section shows WebGL class / renderer / vendor / terminal renderer

### DIFF
--- a/.changesets/1781170152-fix-gpu-bundle-swiftshader-on-windows-status-bar-gpu-indicat-0doe.md
+++ b/.changesets/1781170152-fix-gpu-bundle-swiftshader-on-windows-status-bar-gpu-indicat-0doe.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(gpu): bundle SwiftShader on Windows + status-bar GPU indicator
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -536,18 +536,29 @@ tasks:
                 cp -f "$cefDir/libEGL.dll" dist/cef/ 2>/dev/null || true
                 cp -f "$cefDir/libGLESv2.dll" dist/cef/ 2>/dev/null || true
                 cp -f "$cefDir/d3dcompiler_47.dll" dist/cef/ 2>/dev/null || true
+                # Software-GL fallback (SwiftShader). When the HARDWARE GPU process can't
+                # initialize a GL device — virtual display adapters (Parsec, RDP/headless
+                # sessions), or broken drivers — Chromium 110+ will only fall back to
+                # SwiftShader for WebGL if these are present AND --enable-unsafe-swiftshader
+                # is set (see agentmux-cef/src/app.rs). Without them, GL is disabled entirely
+                # (gl=disabled), WebGL becomes unavailable, and xterm drops to its slower DOM
+                # renderer (no scrollbar). macOS/Linux bundles already ship SwiftShader; this
+                # restores parity on Windows. ~18 MB.
+                # See docs/analysis/ANALYSIS_WEBGL_GPU_DISABLED_ROOTCAUSE_2026_06_10.md.
+                cp -f "$cefDir/vk_swiftshader.dll" dist/cef/ 2>/dev/null || true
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

AgentX-asaf:
@reagentx-workflow review

AgentX-asaf:
Superseded. The status-bar GPU indicator moved to #1337 (focused, frontend-only). The SwiftShader software-fallback approach here is dropped: live CDP evidence shows VS Code (Electron) drives the **hardware NVIDIA RTX 3060** via ANGLE D3D11 on this exact machine while AgentMux/CEF's GPU process crashes — so the GPU is reachable and the right fix is to steer CEF onto the hardware GPU (ANGLE D3D11 + high-performance preference, like VS Code), not fall back to CPU-bound SwiftShader. That hardware-steering fix will be a separate PR.

PR #1337, opened 2026-06-11
feat(statusbar): GPU status indicator — enabled/disabled + driver info

AgentX-asaf:
## What

A bottom-left **status-bar GPU indicator** so the rendering path is visible instead of silent. When Chromium's GPU process is unavailable, WebGL is disabled and xterm silently falls back to its slower DOM renderer (no scrollbar); this makes that state observable.

This is the **first step** toward VS-Code-style robust GPU operation. Hardware-GPU steering (ANGLE D3D11 / high-performance preference) and driver-info-while-disabled are the follow-up.

## Changes (frontend-only — no backend/CEF changes)

- **`frontend/util/gpuutil.ts`** — cached WebGL probe → `hardware` / `software` / `unavailable`, plus the unmasked GL **renderer + vendor**. The renderer string is the GPU model + D3D driver version (same as `chrome://gpu`'s `glRenderer`, e.g. `ANGLE (NVIDIA RTX 3060 … D3D11-32.0.15.9186)`).
- **`statusbar/GpuStatus.tsx`** — compact bottom-left line `GPU HW/SW/off` (green / amber / red) with a detail tooltip.
- **`BackendStatus` popover** (click the uptime) gains a **GPU section**: Status (Enabled hardware / Enabled software / Disabled), WebGL version, **Driver** (renderer), Vendor, and the terminal's actual renderer (WebGL vs DOM).
- **`termwrap.ts`** records the chosen renderer via a new global `termRendererAtom`.

## Why it matters

On a machine where the GPU process can't start, every terminal silently uses the DOM renderer. Today there's no way to see that from the UI. Evidence that this is real (same machine, same Chromium 148):

| | AgentMux (CEF) | VS Code (Electron) |
|---|---|---|
| webgl | `disabled_off` | `enabled` |
| glRenderer | `Disabled` | `ANGLE (NVIDIA RTX 3060 … D3D11)` |

The indicator makes this immediately visible.

## Verification

- ✅ `tsc --noEmit` clean (no errors in changed files)
- ⏳ Live visual check pending a clean isolated dev run (frontend-only; mirrors the existing `BackendStatus`/`SystemStats` patterns)

## Test plan

- [ ] On a healthy-GPU machine: status bar shows `GPU HW`; popover shows Status = Enabled (hardware) + Driver/Vendor + Terminal = WebGL (GPU)
- [ ] On a machine with the GPU disabled: shows `GPU off`; popover shows Status = Disabled + Terminal = DOM (software)
- [ ] Click the uptime → GPU section renders; tooltip on the line shows the renderer

### DIFF
--- a/.changesets/1781171432-feat-statusbar-gpu-status-indicator-enabled-disabled-driver--bp14.md
+++ b/.changesets/1781171432-feat-statusbar-gpu-status-indicator-enabled-disabled-driver--bp14.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(statusbar): GPU status indicator — enabled/disabled + driver info
--- a/frontend/app/statusbar/BackendStatus.tsx
+++ b/frontend/app/statusbar/BackendStatus.tsx
@@ -1,9 +1,10 @@
 // Copyright 2026, AgentMux Corp.
 // SPDX-License-Identifier: Apache-2.0
 
-import { atoms, backendDeathInfoAtom, getApi, setBackendStatusAtom } from "@/store/global";
+import { atoms, backendDeathInfoAtom, getApi, setBackendStatusAtom, termRendererAtom } from "@/store/global";
 import { setRestartInProgress } from "@/store/backendStatus";
 import { waveEventSubscribe } from "@/app/store/wps";
+import { getGpuInfo } from "@/util/gpuutil";
 import { createEffect, createSignal, onCleanup, onMount, Show, type JSX } from "solid-js";
 
 function pad2(n: number): string {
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

AgentX-asaf:
@reagentx-workflow review

AgentX-asaf:
Addressed both P2s in ca07344c:

- **WebGL context leak** — `probe()` now calls `gl.getExtension("WEBGL_lose_context")?.loseContext()` right after reading the capability params, so the throwaway probe no longer holds a per-page WebGL context slot (which xterm's WebGL terminals draw from). Good catch — that would've been self-defeating for a GPU-rendering feature.
- **Two "GPU" labels** — relabeled the compact line to **`GFX HW/SW/off`** (graphics-acceleration state) to disambiguate from SystemStats' `GPU {n}%` (utilization). The detailed popover section stays titled "GPU".

AgentX-asaf:
@reagentx-workflow review --force

PR #1341, opened 2026-06-11
feat(agent): fork prompt for parallel sessions

AgentA-asaf:
## Summary

- **Root cause**: session zones keyed on `definition_id` — opening the same agent in two panes causes both to share one `output.state.json`, clobbering each other's conversation on app restart
- **Fix**: detect collision at the picker level; surface an inline fork prompt that creates an independent definition (new zone) instead of silently reattaching

### Backend
- `forkagentdefinitionsuggest` RPC — read-only, returns `"Name #N"` suggested label based on existing fork count under `parent_id`; called when the name input opens for instant pre-fill
- `forkagentdefinition` handler: `branch_label` is now the fork's display name directly (was `"Name [label]"`), auto-numbers as `"Name #2"` when empty

### Frontend
- `getOpenDefinitionMap()` exported from `agent-pane-state-store` — O(slots) scan returning `definition_id → blockId` for all live panes
- `ForkAgentDefinitionSuggestCommand` added to `rpc-api.ts`
- `MyAgentsList`: active badge (green dot) on rows with open definitions; inline fork prompt with **Open new session** / **Switch to existing** buttons; name input pre-filled with backend suggestion, Start disabled when empty, Enter confirms, Escape cancels
- `AgentPicker`: `openDefinitions` signal computed from pane store (refreshed on `agents:changed`); `handleFork` → `ForkAgentDefinitionCommand` + `launchAgentDefinition`; `handleSwitchToExisting` → `refocusNode`

## UX flow
```
User clicks active row
  → fork prompt appears (row stays in place)
  → "Open new session" → name input expands, pre-filled "Mopeo #2"
  → user edits / confirms → fork created → new pane opens
  → "Switch to existing" → focus jumps to the already-open pane
```

## Test plan
- [ ] Open Mopeo in two panes — second click shows fork prompt (active badge visible on first click)
- [ ] "Open new session" → name input appears pre-filled with "Mopeo #2"
- [ ] Edit name → "Start" enabled only when non-empty
- [ ] Confirm → new pane opens with a fresh conversation zone
- [ ] "Switch to existing" → focus jumps to the first pane
- [ ] Restart app → both sessions remain independent (no clobber)
- [ ] Third fork → suggested name is "Mopeo #3"

### DIFF
--- a/.changesets/1781173320-fix-agent-pane-idle-send-messages-no-longer-flash-in-the-que-09vf.md
+++ b/.changesets/1781173320-fix-agent-pane-idle-send-messages-no-longer-flash-in-the-que-09vf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): idle-send messages no longer flash in the queued zone
--- a/.changesets/1781189133-feat-agent-fork-prompt-for-parallel-sessions-detect-active-p-6sqa.md
+++ b/.changesets/1781189133-feat-agent-fork-prompt-for-parallel-sessions-detect-active-p-6sqa.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): fork prompt for parallel sessions — detect active panes and offer named fork instead of silent collision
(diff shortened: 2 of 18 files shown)


a5af:
@codex review

PR #1342, opened 2026-06-11
fix(layout): resize border no longer triggers pane tear-off

AgentA-asaf:
## Summary

- **`ResizeHandle.onPointerDown`** now calls `event.preventDefault()` before `setPointerCapture`. This suppresses the subsequent `mousedown` event; since HTML5 DnD requires `mousedown` to start, no drag can initiate from a press on the resize handle even if an underlying element would otherwise react.
- **`DisplayNode.canDrag`** now rejects drags whose initial pointer position (`input.clientX/Y`) falls within the ±halfSize zone of any resize handle's `centerPx`. Defense-in-depth for the race: if the pointer barely misses the handle element and lands on a header pixel near the border, the drag is cancelled before `onDragStart` ever sets `_currentDragPayload`.

## Root cause

The resize handle (6 px, centered at the pane boundary) overlaps 1.5 px with the top of the adjacent pane's header. In that overlap zone the resize handle wins via `z-index: 3` vs the header's `z-index: auto`, but under WebView2 timing edge cases (layout recomputing during a reactive update, subpixel rounding at the boundary) the header occasionally received `mousedown` instead — starting a pragmatic-dnd drag that set `_currentDragPayload`, which `CrossWindowDragMonitor` then interpreted as a tear-off request.

Pane tear-off from the header remains fully functional. Only drags whose origin is within the resize-handle zone are blocked.

## Test plan

- [ ] Resize panes by dragging the border — no tear-off, resize works correctly
- [ ] Hover over the resize border so the resize cursor appears, click and drag — should resize, not tear off
- [ ] Drag a pane header away from the border — tear-off still works
- [ ] Drag a pane header that's adjacent to a border (e.g. left edge of right pane in a side-by-side layout) — tear-off still works when dragging from > 3 px inside the header

### DIFF
--- a/.changesets/1781240500000-fix-agent-composer-remove-hint-soften-input-focus-border-kq81.md
+++ b/.changesets/1781240500000-fix-agent-composer-remove-hint-soften-input-focus-border-kq81.md
@@ -1,0 +1,14 @@
+---
+type: patch
+---
+
+fix(agent): remove keyboard hint line; soften input focus border
+
+Removes the "Enter to send • Shift+Enter for newline • Esc to clear / stop"
+hint line from the agent pane composer footer — it consumes vertical space
+without adding value for returning users.
+
+Changes the textarea focus border from full `--accent-color` to
+`color-mix(in srgb, var(--accent-color) 40%, transparent)` — a lighter
+variation of the pane-selected border color that stays visually connected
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1344, opened 2026-06-11
fix(gpu): bundle SwiftShader software-GL fallback on Windows

AgentX-asaf:
## Problem

On Windows the GPU is **fully disabled** (`gl=disabled`, `webgl: disabled_off`). The CEF GPU process crashes `STATUS_BREAKPOINT` ×3 at "create shared context for virtualization", falls to software GL, which **also** fails (no SwiftShader bundled), so Chromium disables all GPU → every terminal uses the slow xterm **DOM renderer** (no scrollbar).

Full diagnosis: `docs/analysis/ANALYSIS_WINDOWS_GPU_DISABLED_ROOTCAUSE_2026_06_11.md`.

**Root cause** (the real hardware fix, *separate*): the `agentmuxai/cef` fork's **Windows `libcef.dll` is a non-official / DCHECK-enabled build** — a `DCHECK` in GPU context init fires as a fatal `int3`, which upstream release Chrome (VS Code) and AgentMux's **macOS** build both survive on the same machine. Evidence: the CEF log prints full source paths/line numbers (`gpu_channel_manager.cc:927`) which official Chrome strips. Filed as an issue on `agentmuxai/cef` (build Windows libcef with `is_official_build=true`). **Ruled out:** transparency (opaque A/B crashes identically), adapter selection (NVIDIA correctly chosen, fully capable — `chrome://gpu` Hardware profile init 192ms), AMD-switchable (false).

## This PR — the software-GL safety net

So a hardware-GPU failure degrades to **software WebGL** instead of the fully-disabled cliff:
- **`Taskfile.yml` `bundle:windows`** ships `vk_swiftshader.dll`, `vulkan-1.dll`, `vk_swiftshader_icd.json` (**~6 MB**; parity with macOS/Linux which already bundle it).
- **`app.rs`** adds `--enable-unsafe-swiftshader` (Chromium 110+ gates the SwiftShader WebGL fallback behind it).

With both, when the GPU process can't boot, Chromium settles on SwiftShader (software WebGL) instead of `crashed too many times with software GL → disable all`. Keeps the xterm WebGL renderer + scrollbar working.

**Not the end goal** — software is CPU-bound; hardware (the official libcef build) is. But this removes the disabled-DOM-renderer cliff, and **hardware GL is still preferred when it boots, so there is no cost on healthy machines.**

## Verification
- ✅ `cargo check -p agentmux-cef` compiles
- ✅ SwiftShader DLLs present in the CEF runtime to bundle (~6.1 MB → ~3 MB in the ZIP)
- ⏳ Runtime "software WebGL kicks in" confirmation pending (needs the build; or validates naturally once the official-libcef fix lands and hardware works)

## Test plan
- [ ] `task package` → confirm `vk_swiftshader.dll` + `vulkan-1.dll` in `runtime/`
- [ ] On a machine where the hardware GPU process can't boot → terminals render via software WebGL (status bar `GFX SW`, scrollbar present) instead of `GFX off`
- [ ] On a healthy-GPU machine → hardware path unchanged (`GFX HW`), no regression

## Follow-ups (tracked, not here)
- **Real fix:** official (DCHECK-off) Windows `libcef` build in `agentmuxai/cef`.
- Stability roadmap from `CRASH_GPU_PROCESS_FATAL_2026_05_20.md`: `--disable-gpu-process-crash-limit`, launcher relaunch w/ `--disable-gpu`, `SetErrorMode`, launcher supervision + restore.

### DIFF
--- a/.changesets/1781194353-fix-gpu-bundle-swiftshader-software-gl-fallback-on-windows-zfgh.md
+++ b/.changesets/1781194353-fix-gpu-bundle-swiftshader-software-gl-fallback-on-windows-zfgh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(gpu): bundle SwiftShader software-GL fallback on Windows
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -536,18 +536,29 @@ tasks:
                 cp -f "$cefDir/libEGL.dll" dist/cef/ 2>/dev/null || true
                 cp -f "$cefDir/libGLESv2.dll" dist/cef/ 2>/dev/null || true
                 cp -f "$cefDir/d3dcompiler_47.dll" dist/cef/ 2>/dev/null || true
+                # Software-GL fallback (SwiftShader, ~6 MB). When the HARDWARE GPU process
+                # can't boot, Chromium falls back to SwiftShader for WebGL ONLY if these are
+                # present AND --enable-unsafe-swiftshader is set (app.rs). Without them, the
+                # software-GL fallback also fails ("crashed too many times with software GL")
+                # and Chromium disables GL entirely → WebGL unavailable → xterm drops to its
+                # slower DOM renderer (no scrollbar). This is the current Windows failure mode
+                # (GPU process STATUS_BREAKPOINTs at init — see issue / the DCHECK-build root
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentX-asaf:
@reagentx-workflow review

PR #1346, opened 2026-06-11
fix(package): copy SwiftShader DLLs into portable runtime (follow-up to #1344)

AgentX-asaf:
Follow-up to #1344. That PR's Taskfile bundle step copies `vk_swiftshader.dll` / `vulkan-1.dll` / `vk_swiftshader_icd.json` into `dist/cef`, but `scripts/package-portable.sh` has its **own** explicit GPU-DLL copy list (`libEGL`/`libGLESv2`/`d3dcompiler`) that didn't include SwiftShader — so the portable's `runtime/` shipped **without** it and the software-GL fallback never reached users.

**Caught by building from main and inspecting the artifact:** SwiftShader DLLs were present in `dist/cef/` but missing from `runtime/`. This adds the runtime copy.

## Test plan
- [ ] `task package` → confirm `vk_swiftshader.dll` + `vulkan-1.dll` + `vk_swiftshader_icd.json` are in the portable's `runtime/` (not just `dist/cef/`)

### DIFF
--- a/.changesets/1781196402-fix-package-copy-swiftshader-dlls-into-portable-runtime-foll-5c45.md
+++ b/.changesets/1781196402-fix-package-copy-swiftshader-dlls-into-portable-runtime-foll-5c45.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(package): copy SwiftShader DLLs into portable runtime (follow-up to #1344)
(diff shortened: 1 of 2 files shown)


AgentX-asaf:
@reagentx-workflow review

a5af:
@codex review

PR #1347, opened 2026-06-11
feat(container): Phase 4 — container agent image (Dockerfile + GHCR CI)

AgentY-asaf:
## Summary

- Adds `docker/Dockerfile.agent-agentmux` — AgentMux's container agent image, fully independent of claw
- Adds `.github/workflows/container-image.yml` — builds and pushes `ghcr.io/agentmuxai/agent-claude` on version tags
- Adds `docs/specs/SPEC_CONTAINER_PANE_SUPPORT_2026_06_11.md` — full implementation spec

## Image design

| Choice | Rationale |
|---|---|
| `node:22-slim` base | glibc required — Alpine/musl crashes Claude Code |
| `tini` as PID 1 | `docker exec -i` doesn't propagate signals; tini handles reaping and forwarding |
| `DISABLE_AUTOUPDATER=1` | Version pinned at build time, not runtime |
| `~/.claude` as named volume | Credential isolation per agent; not a host bind mount (Anthropic recommendation) |
| `/workspace` bind mount | AgentMux runtime mounts the host workspace here |

## Independence boundary

This image is built and distributed by AgentMux. It shares design patterns with claw's `Dockerfile.agent` (same base image, same user setup) but has no runtime or build dependency on claw.

## Test plan

- [ ] `docker build -f docker/Dockerfile.agent-agentmux docker/` builds without error
- [ ] `docker run --rm ghcr.io/agentmuxai/agent-claude:local claude --version` returns version
- [ ] CI workflow triggers on version tag push

### DIFF
--- a/.changesets/1781224588-feat-container-add-dockerfile-and-ci-workflow-for-agent-clau-lvc5.md
+++ b/.changesets/1781224588-feat-container-add-dockerfile-and-ci-workflow-for-agent-clau-lvc5.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(container): add Dockerfile and CI workflow for agent-claude image
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -45,6 +45,9 @@ jobs:
             # refs/tags/vX.Y.Z here; the previous refs/heads/main check was
             # never true in this workflow and suppressed :latest entirely)
             type=raw,value=latest,enable=${{ github.ref_type == 'tag' }}
+            # workflow_dispatch does not push a semver tag so type=semver emits nothing;
+            # produce a dispatch-<sha> tag so manual runs actually push an image.
+            type=raw,value=dispatch-{{sha}},enable=${{ github.event_name == 'workflow_dispatch' }}
 
       - name: Resolve Claude Code version
         id: claude_ver
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -41,14 +41,22 @@ jobs:
           tags: |
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
-            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
+            # publish :latest on every version-tag push (github.ref is always
+            # refs/tags/vX.Y.Z here; the previous refs/heads/main check was
+            # never true in this workflow and suppressed :latest entirely)
+            type=raw,value=latest,enable=${{ github.ref_type == 'tag' }}
 
       - name: Resolve Claude Code version
         id: claude_ver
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed all three findings from round 1:

**P1 — `:latest` tag condition** — changed `enable=${{ github.ref == 'refs/heads/main' }}` to `enable=${{ github.ref_type == 'tag' }}`. The workflow only fires on `push: tags: ['v*']` so `github.ref` is always `refs/tags/vX.Y.Z`, never `refs/heads/main` — the previous condition suppressed `:latest` on every version-tag build. Now publishes `:latest` on all tag pushes.

**P1 — shell injection** — bound `github.event.inputs.claude_version` to an env var (`CLAUDE_VERSION_INPUT`) and reference `$CLAUDE_VERSION_INPUT` in the script body. Direct `${{ }}` expansion before the shell executes allows arbitrary command injection from a crafted input.

**P2 — changeset** — added changeset file (minor, as this is a new feature artifact).

AgentY-asaf:
Addressed all three findings from round 2:

**P0 — UID 1000 conflict** — `node:22-slim` ships a `node` user at UID 1000; `useradd -u 1000 agent` fails without `--non-unique`. Fixed by renaming the existing user/group via `usermod -l agent -d /home/agent -m node` + `groupmod -n agent node`. This preserves UID 1000 (for Linux host-FS UID parity) without creating a duplicate.

**P1 — ENTRYPOINT exec model** — replaced `ENTRYPOINT ["tini", "--", "claude"]` with `ENTRYPOINT ["tini", "--"]` + `CMD ["sleep", "infinity"]`. The container now stays alive with `sleep infinity` as the idle keeper process; agent turns run via `docker exec`, not by replacing PID 1. tini remains for signal forwarding and zombie reaping.

**P2 — workflow_dispatch produces no tags** — added `type=raw,value=dispatch-{{sha}},enable=${{ github.event_name == 'workflow_dispatch' }}`. Without this, `type=semver` emits nothing on non-tag refs and `latest` is disabled, leaving the manual trigger with an empty tag list and no image push.

AgentY-asaf:
Addressed P2 from round 3:

**P2 — `~/.claude` not owned by agent** — added `RUN mkdir -p /home/agent/.claude && chown agent:agent /home/agent/.claude` before `USER agent`. A fresh Docker named volume is initialised `root:root`; without this the agent user cannot write its config/session state on first run.

AgentY-asaf:
Addressed P2 from round 4:

**P2 — amd64-only build** — added `setup-qemu-action` + `setup-buildx-action` and set `platforms: linux/amd64,linux/arm64` on the build-push step. Without arm64, the image runs under QEMU emulation on Apple Silicon (OrbStack/Colima/Docker Desktop). The manifest now includes both architectures.

PR #1348, opened 2026-06-11
fix(bashwrap): collapse lone CR so spinner animations don't bleed as multiple lines

AgentA-asaf:
## Summary

- Spinner/progress animations that use `\r` to overwrite the current line were rendering as one line per frame in the agent pane's `<pre>` log view. HTML normalizes bare `\r` to a line break, so each animation frame appeared on its own line.
- Fix: in `strip_ansi` (PTY path), a lone `\r` now truncates the current visual line in the output buffer back to after the last `\n`, simulating what a real terminal does (cursor to column 0, next write overwrites).
- CRLF (`\r\n`) handling is unchanged.

## Test plan

- [ ] Run a command that uses a spinner (e.g. `npm install`, a Python `tqdm` bar, or `printf 'frame1\rframe2\rfinal\n'`) in an agent pane — only the final frame should appear in the log, not one line per frame.
- [ ] Confirm normal multi-line output (with `\n`) is unaffected.
- [ ] Confirm CRLF output is still normalized to LF.

### DIFF
--- a/.changesets/1781172595-feat-agent-pane-persistent-shell-node-shellnode-type-reducer-wiiq.md
+++ b/.changesets/1781172595-feat-agent-pane-persistent-shell-node-shellnode-type-reducer-wiiq.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): persistent shell node — ShellNode type, reducer, PersistentShellBlock component and styles (Phase 1 frontend skeleton)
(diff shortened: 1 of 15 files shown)


a5af:
@codex review

AgentA-asaf:
Superseded by #1349 — rebased on main, upgraded to pending-buffer-level CR collapse that handles cross-read spinner frames.

PR #1351, opened 2026-06-11
fix(bashwrap): collapse lone CR so spinner animations don't bleed as multiple lines

AgentA-asaf:
## Summary

Spinner/progress animations (npm install, tqdm, cargo, etc.) use `\r` to overwrite the current line. In the agent pane's `<pre>` log view, HTML normalizes bare `\r` to a line break, so each frame appeared on its own line instead of overwriting.

**Root cause:** the previous approach tried to collapse `\r` inside `strip_ansi` (per-PTY-read). But real spinners emit each frame as a separate `process.stdout.write`, which arrives as a separate PTY read — by the time the `\r` arrives, the prior frame has already been flushed as its own chunk.

**Fix:** add `collapse_cr()` that operates on the `pending` accumulator buffer *after* each read is appended. Since `pending` accumulates content across reads until a `\n` drains it, cross-read `\r` overwrites are collapsed correctly:

- Read N: `"⣾ Loading..."` → `pending = "⣾ Loading..."` → partial flush → chunk emitted... **wait**
- Actually: `pending` is only flushed as a partial line if it's non-empty at end of read. `collapse_cr` runs before that flush, so if the current pending content ends without a `\r` yet, it still goes out. The key case: when the *next* read brings `"\r⣽ Loading..."`, `pending` is empty (previous flush drained it), so `collapse_cr` on `"\r⣽ Loading..."` sees `\r` at start, `line_start=0`, truncates to 0, then pushes `"⣽ Loading..."`.

Result: only the final non-`\r`-terminated frame (the `\n`-ended completion line) renders in the log. Applied to both the PTY path (`pty_reader_loop`) and the pipe fallback path (`stream_reader`).

## Changes

- `collapse_cr(&mut Vec<u8>)` — new function, O(n) scan, fast-paths if no `\r`
- Called in `pty_reader_loop` and `stream_reader` after each `pending.extend_from_slice`
- `strip_ansi` reverted to pass lone `\r` through (CRLF→LF still handled there)

## Test plan

- [ ] `printf 'frame1\rframe2\rfinal\n'` — only `final` appears in agent pane log
- [ ] `npm install` / `cargo build` spinner — only last frame (done/error line) visible per animation cycle
- [ ] Normal multi-line output (`\n`-separated) renders unchanged
- [ ] CRLF output still normalized to LF


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Pushed f2c36bed: merged latest main (resolved the bash_wrap.rs conflict — kept the thread-bridge/quiet-window CR-collapse structure, preserved main's CONIN-writer-lifetime + brace-group-redirect changes and the raw_n debug log) and addressed the last P2 — removed the dead `pending_cr_override.take()` in the PTY drain loop (the override is already consumed/prepended before the loop, matching the pipe path in stream_reader). All 31 bashwrap tests pass. Ready for re-review.

a5af:
@codex review

PR #1350, opened 2026-06-11
feat(container): add container schema fields to AgentDefinition (Phase 0)

AgentY-asaf:
## Summary

Phase 0 of SPEC_CONTAINER_PANE_SUPPORT_2026_06_11.md — schema foundation for container-type agent panes.

- Adds container_image, container_volumes, container_name to AgentDefinition struct
- Updates all storage SELECT/INSERT/UPDATE queries (db_agent_definitions + db_agents)
- Updates dual_write.rs upsert and instance_create helpers
- Adds 6 idempotent ALTER TABLE migrations (3 per table), bumps OBJECT_SCHEMA_VERSION 5→6
- Adds fresh-database CREATE TABLE columns to both tables
- Extends CommandAgentDefineData and CommandUpdateAgentDefinitionData RPC types
- container_name is server-managed only (never accepted from clients)
- All host-agent rows default to '', '[]', '' — no data loss on existing databases

Independence boundary: AgentMux and claw are fully independent. No Docker runtime code in this phase.

## Phase sequence

| Phase | Branch | Status |
|-------|--------|--------|
| 0 — Schema | agenty/container-schema | this PR |
| 1 — ContainerManager | agenty/container-manager | next |
| 2 — Spawn routing | agenty/container-spawn | after Phase 1 |
| 4 — Dockerfile + CI | agenty/container-image | open (#1347) |

## Test plan
- [ ] cargo build -p agentmux-srv passes
- [ ] cargo test --no-run compiles clean
- [ ] Existing agent definitions survive migration (columns default to ''/'/[]'/'')
- [ ] agent.define with no container fields creates host agent with empty container fields
- [ ] updateagent preserves container_name and accepts container_image/container_volumes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1352, opened 2026-06-11
feat(container): ContainerManager with bollard Docker SDK (Phase 1)

AgentY-asaf:
## Summary

Phase 1 of SPEC_CONTAINER_PANE_SUPPORT_2026_06_11.md — Docker runtime management.

Adds `agentmux-srv/src/backend/container.rs` with `ContainerManager`:

**Design (cross-platform, independence boundary — no claw dependency):**
- `connect()` via `Docker::connect_with_local_defaults()` — honors `DOCKER_HOST`, Windows named pipe, macOS socket, Linux `/var/run/docker.sock`
- Persistent container model (one per agent, exec per turn) — same pattern as VS Code Dev Containers
- `ensure_running()`: create+start on first use, restart stopped containers, no-op if running
- `exec()`: `docker exec -i` (no `-t`) — preserves NDJSON newlines for Claude Code output
- `stop()` / `remove()` for lifecycle management
- Named Docker volume for `/home/agent/.claude` — avoids host credential leakage, follows Anthropic recommendation
- `container_name_for_slug()` derives stable `agentmux-<slug>` names independent of claw
- Startup check in `main.rs` warns (non-fatal) if Docker unavailable
- `bollard = "0.18"` added to `Cargo.toml`

**Depends on:** PR #1350 (Phase 0 — schema fields)

## Phase sequence

| Phase | Branch | Status |
|-------|--------|--------|
| 0 — Schema | agenty/container-schema | PR #1350 |
| 1 — ContainerManager | agenty/container-manager | **this PR** |
| 2 — Spawn routing | agenty/container-spawn | next |
| 4 — Dockerfile + CI | agenty/container-image | PR #1347 |

## Test plan
- [ ] `cargo check -p agentmux-srv` compiles clean
- [ ] Server starts without Docker installed (warning logged, no crash)
- [ ] Server starts with Docker installed (ping succeeds, info logged)
- [ ] `ContainerManager::connect()` + `check_available()` returns Ok when Docker running

### DIFF
--- a/.changesets/1781209707-feat-container-add-containermanager-with-bollard-docker-sdk--3e3o.md
+++ b/.changesets/1781209707-feat-container-add-containermanager-with-bollard-docker-sdk--3e3o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(container): add ContainerManager with bollard Docker SDK (Phase 1)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,6 +99,7 @@ dependencies = [
  "async-stream",
  "axum 0.7.9",
  "base64",
+ "bollard",
  "chrono",
  "clap",
  "crash-handler",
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed all four findings from round 1:

**P1 — Cargo.lock** — committed updated lockfile with bollard 0.18 entries via `cargo fetch`.

**P1 — `tracing::warn!(spec = spec, …)`** — changed to `spec = %spec` (Display formatter); `&String` doesn't implement `tracing::Value`.

**P2 — TOCTOU in `ensure_running()`** — added `ensure_locks: Mutex<HashMap<String, Arc<Mutex<()>>>>` to `ContainerManagerInner`. Each `ensure_running()` call acquires the per-container lock before touching Docker, serializing concurrent callers for the same name. Added a second cache-check after acquiring the lock (to skip the Docker round-trip when the first caller already finished). Always queries Docker rather than trusting a potentially stale cache entry.

**P2 — misleading `ExecSession` doc comment** — updated to accurately describe bollard's multiplexed `LogOutput::StdOut` / `LogOutput::StdErr` frame model with `tty: false`.

AgentY-asaf:
Addressed both P2 findings from round 2:

**P2 — stale-cache fast-path** — removed the in-memory `ContainerState` cache entirely. The pre-lock fast-path that returned `Ok(())` when the cache said `Running` (without querying Docker) contradicted the doc comment and would silently miss an externally-killed container. `ensure_running` now always queries Docker under the per-container lock. The lock still serializes concurrent callers for the same container to prevent 409 create races; once the first caller finishes, the second sees the container as `running` via the Docker query and returns immediately.

**P2 — Windows drive-letter parse** — replaced `spec.splitn(3, ':')` with a `parse_volume_spec()` helper that detects the `X:\` / `X:/` drive-letter prefix and skips the drive colon before splitting on the next `:`. The old code produced `source="C"` and `target="\Users\me\repo"` for `C:\Users\me\repo:/workspace`. Added unit tests for Unix named volumes, Unix bind mounts, and Windows bind mounts (backslash + forward-slash).

Issue #1345, opened 2026-06-11
Windows GPU disabled: libcef is a non-official/DCHECK build — GPU process STATUS_BREAKPOINTs at init


AgentX-asaf:
**The real hardware-GPU fix** (the SwiftShader bundle in #1344 is only a software safety net).

## Problem
On Windows the GPU is fully disabled — the CEF GPU process crashes `STATUS_BREAKPOINT` (`0x80000003`) ×3 at "create shared context for virtualization", so Chromium disables all GPU and terminals use the slow xterm DOM renderer (no scrollbar). The NVIDIA RTX 3060 hardware path is fully capable (`chrome://gpu` Hardware profile, init 192ms) — it just never survives the checked init.

## Root cause
The `agentmuxai/cef` fork's **Windows `libcef.dll` is a non-official / DCHECK-enabled build** (consumed via `AgentU-asaf/cef-rs`). A `DCHECK` in GPU context-virtualization init fires as a fatal `int3`. **Upstream release Chrome (VS Code) and AgentMux's macOS build both survive it on the same machine.**

Evidence it's non-official: `cef-debug.log` prints full source paths/line numbers (`gpu_channel_manager.cc:927`, `viz_main_impl.cc:190`) — official Chrome strips these.

Ruled out: transparency (opaque A/B crashes identically), adapter selection (NVIDIA correctly chosen), AMD-switchable (false). Full analysis: `docs/analysis/ANALYSIS_WINDOWS_GPU_DISABLED_ROOTCAUSE_2026_06_11.md` (#1344).

## Fix
Build the **Windows** libcef from `agentmuxai/cef` with **`is_official_build=true`** (or `dcheck_always_on=false` + `is_debug=false`), matching the release/macOS build. Re-point the `cef-rs` download and rebuild. (The CEF fork has issues disabled, so tracking here.)

## Also worth doing (stability roadmap — `CRASH_GPU_PROCESS_FATAL_2026_05_20.md`)
`--disable-gpu-process-crash-limit`, launcher relaunch with `--disable-gpu` after repeated crashes, `SetErrorMode(SEM_NOGPFAULTERRORBOX)`, launcher supervision + state restore — so any GPU fault becomes invisible recovery.

Related: #1344 (SwiftShader software-GL safety net).

AgentX-asaf:
**Correction — this root cause was wrong, and the real fix is found (#1354).**

[...]

PR #1354, opened 2026-06-11
fix(gpu): embed supportedOS manifest so the GPU process can initialize on Windows

AgentX-asaf:
## The real fix for Windows GPU being disabled

**Root cause:** `agentmux-cef.exe` and `agentmux-launcher.exe` shipped **without a Windows application manifest** declaring `supportedOS`. Windows version-lies to unmanifested exes — `GetVersionEx` reports **6.2 (Windows 8)** on a Windows 11 box. `chrome://gpu` confirmed it: `Operating system: Windows NT 6.2.9200`.

Chromium's GPU init is **OS-version-gated** (DirectComposition, D3D feature detection, driver-bug workarounds). Fed a fake "Windows 8", the GPU process took a path that `CHECK`-crashed (`STATUS_BREAKPOINT`) at "create shared context for virtualization" on this Win11 / multi-GPU / Parsec-virtual-display hardware → 3 crashes → Chromium disabled GL → xterm DOM renderer (no scrollbar, `GFX off`).

**Why VS Code works on the identical machine:** Electron embeds exactly this manifest and reports the true OS, so its GPU process initializes hardware NVIDIA D3D11. (Thanks to @a5af for pointing at VS Code — that's what cracked it.)

## Fix
`winres` now embeds an app manifest with `supportedOS` GUIDs (Win7→Win10/11) + `asInvoker` (preserves prior UAC behavior) on **both** the host and the launcher. The GPU process runs as `agentmux-cef.exe`, so its manifest is what determines the reported OS.

## Verified — packaged build, virgin data dir

| | before | after |
|---|---|---|
| OS reported | `Windows NT 6.2.9200` | **`Windows NT 10.0`** |
| GPU-process crashes | 3 (every launch) | **0** |
| `shared context` failure | yes | **none** |
| GPU process `--use-gl` | `disabled` | **(none → default hardware path)** |

## Corrects earlier wrong analysis
My prior conclusions (custom/DCHECK libcef build; transparency) were **wrong** — the Windows `libcef.dll` is the stock official CEF 148.0.9 release (pulled from `cef-builds.spotifycdn.com`). The problem was purely the missing exe manifest. Issue #1345 and the related analysis doc should be corrected/closed.

## Supersedes / reframes
- The SwiftShader bundle (#1344/#1346) stays as general robustness, but **isn't needed for this** — hardware GPU now works.
- The conditional-SwiftShader-fallback idea is moot for this machine.

## Test plan
- [ ] `task package`; launch on a Win11 machine → `chrome://gpu` shows `Windows NT 10.0`, hardware ANGLE D3D11, status bar `GFX HW`
- [ ] No regression to UAC (asInvoker), DPI, or window behavior

### DIFF
--- a/.changesets/1781212885-fix-gpu-embed-supportedos-manifest-so-windows-reports-the-tr-x1e7.md
+++ b/.changesets/1781212885-fix-gpu-embed-supportedos-manifest-so-windows-reports-the-tr-x1e7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(gpu): embed supportedOS manifest so Windows reports the true OS version — fixes GPU process crash
--- a/agentmux-cef/build.rs
+++ b/agentmux-cef/build.rs
@@ -1,6 +1,33 @@
 // Copyright 2025-2026, AgentMux Corp.
 // SPDX-License-Identifier: Apache-2.0
 
+// Windows application manifest. The `supportedOS` GUIDs (Vista, 7, 8, 8.1,
+// 10/11) make Windows report the true OS version to the process instead of
+// capping at 6.2 — required for correct Chromium GPU initialization. Mirrors
+// the manifest Electron embeds. `asInvoker` preserves winres's default UAC
+// behavior (no elevation prompt).
+#[cfg(target_os = "windows")]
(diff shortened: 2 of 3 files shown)


AgentX-asaf:
@reagentx-workflow review

a5af:
@codex review

PR #1358, opened 2026-06-12
fix(term): make terminal scrollbar always visible and clickable

AgentX-asaf:
## Summary

- Removes the show/hide observer div mechanism — the observer divs in the scrollbar zone were intercepting the clicks they were meant to let through
- Makes the scrollbar thumb permanently visible (removes the `.terminal:hover` gate that hid the thumb when the cursor moved from the text area onto the scrollbar track, mid-reach)
- Sub-cell gap fix: `customFit` injects a per-terminal `<style>` that widens the scrollbar to `14px + remainder` so text columns tile flush to the scrollbar left edge with no zoom-dependent gap
- **Fixes URL hover regression** (reagent P1 on first two commits): the initial fix incorrectly added `.xterm-screen { pointer-events: none }`. Since `pointer-events` is inherited, that made all canvas/row children inherit `none` too, removing the entire screen subtree from hit-testing — events fell through to `.xterm-viewport`, bypassing xterm's Linkifier2 (bound to `.xterm-screen`), so `WebLinksAddon` and `FilePathLinkProvider` lost hover-underline and click-activation. The third commit removes the rule. The scrollbar remains clickable: `customFit` sizes `.xterm-screen` to exactly `cols × cellWidth`, so it ends flush with the scrollbar left edge and never overlaps it.

## Test plan

- [ ] Open a terminal pane, run a command with lots of output to generate scrollback
- [ ] Hover over the scrollbar — thumb stays visible (no disappearing on approach)
- [ ] Click the scrollbar thumb — scrolls correctly
- [ ] Click-drag the scrollbar thumb — drag works
- [ ] Click in the scrollbar track (above/below thumb) — jumps scroll position
- [ ] Hover over a URL/file path in the terminal — cursor changes to pointer, link underlines
- [ ] Click the URL — opens link
- [ ] Text selection by click-drag in the terminal still works
- [ ] Keyboard input still works

### DIFF
--- a/.changesets/1781253502-fix-term-make-scrollbar-thumb-always-visible-and-clickable-d0rc.md
+++ b/.changesets/1781253502-fix-term-make-scrollbar-thumb-always-visible-and-clickable-d0rc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): make scrollbar thumb always visible and clickable
--- a/.changesets/1781304156-fix-term-restore-url-hover-in-terminal-pane-4n5e.md
+++ b/.changesets/1781304156-fix-term-restore-url-hover-in-terminal-pane-4n5e.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): restore URL hover in terminal pane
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1359, opened 2026-06-12
fix(term): fix xterm paste truncation — chunked send, BPM on, larger buffer

AgentY-asaf:
## Summary

Fixes silent paste truncation in xterm terminal panes when pasting large amounts of text.

- **`shell.rs`** — `SHELL_INPUT_CH_SIZE` 32 → 256: prevents burst-drop when rapid input arrives faster than the PTY write loop drains
- **`termViewModel.ts`** — `sendDataToController` now chunks input >4 KB into 4 KB frames with 5 ms inter-chunk delay; shows a "Pasting…" notification for pastes ≥8 KB that auto-dismisses when complete
- **`term.tsx`** — Bracketed paste mode default changed `false` → `true`; modern shells (bash 4+, zsh, fish) all support BPM and it prevents the shell from executing partial lines mid-paste. Override per-pane via `term:allowbracketedpaste=false` for legacy shells

Full root-cause analysis: `docs/specs/SPEC_XTERM_PASTE_TRUNCATION_2026_06_12.md`

## Root cause

Large pastes were sent as a single PTY write. Windows ConPTY and the backend 32-slot input channel (`SHELL_INPUT_CH_SIZE`) could not absorb bursts, silently dropping data with only a backend log warning (`"input channel full, dropping"`).

## Test plan

- [ ] Paste 1 KB text → works as before (fast path, no chunking)
- [ ] Paste 10 KB text → "Pasting…" notification appears, full text arrives
- [ ] Paste 100 KB text → all text arrives in order, no truncation
- [ ] Multi-line paste with BPM on → shell does not execute partial lines mid-paste
- [ ] `term:allowbracketedpaste=false` disables BPM as expected
- [ ] Regular keystrokes unaffected (< 4 KB threshold, fast path taken)

### DIFF
--- a/.changesets/1781253563-fix-term-fix-xterm-paste-truncation-chunked-send-larger-inpu-u698.md
+++ b/.changesets/1781253563-fix-term-fix-xterm-paste-truncation-chunked-send-larger-inpu-u698.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(term): fix xterm paste truncation — chunked send, larger input buffer, BPM on by default
--- a/agentmux-srv/src/backend/blockcontroller/shell.rs
+++ b/agentmux-srv/src/backend/blockcontroller/shell.rs
@@ -40,10 +40,17 @@ use crate::backend::storage::store::Store;
 use crate::backend::obj::{self, MetaMapType};
 use crate::backend::wps;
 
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1360, opened 2026-06-12
fix(agent): inject Bash(agentmux-bashwrap*) into permissions.allow

AgentX-asaf:
## Summary

- The PreToolUse hook rewrites every Bash tool call to `agentmux-bashwrap exec --tool-id=... --b64-cmd=...`
- Claude Code treats that rewritten command as a new Bash invocation and checks it against `permissions.allow`
- Without a matching entry, Claude Code blocks it — agents see "bash-wrap unusable / permissions error"
- `build_settings_with_hooks` now appends `"Bash(agentmux-bashwrap*)"` to `permissions.allow`, merging with any user-supplied allow entries

## Test plan

- [ ] Start an agent pane
- [ ] Run a bash command from the agent (e.g. `ls`)
- [ ] Live streaming output appears — no permissions prompt or block
- [ ] Check `.claude/settings.json` in the agent workspace contains `"permissions": { "allow": ["Bash(agentmux-bashwrap*)"] }`
- [ ] User-supplied `permissions.allow` entries in agent settings are preserved alongside the injected entry

### DIFF
--- a/.changesets/1781257955-fix-agent-inject-bash-agentmux-bashwrap-into-permissions-all-suy5.md
+++ b/.changesets/1781257955-fix-agent-inject-bash-agentmux-bashwrap-into-permissions-all-suy5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): inject Bash(agentmux-bashwrap*) into permissions.allow so bashwrap exec is not blocked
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

AgentX-asaf:
Closing as **superseded**. `main` already injects the bashwrap allow entry in `agent_config.rs` — in a safer form: `Bash(agentmux-bashwrap *)` (with a space), which restricts the match to `agentmux-bashwrap <args>` and avoids matching unrelated executables like `agentmux-bashwrapXYZ`. This PR's `Bash(agentmux-bashwrap*)` (no space) would match those too.

This branch is also now CONFLICTING with main. The need it addressed is already met, so closing rather than resolving the conflict (which would regress the glob). 🤖

PR #1361, opened 2026-06-12
fix(agent-pane): schema v2 snapshot — lightweight overlay, no nodes[] serialization

AgentA-asaf:
## Summary

- **Root cause fixed**: `writeSnapshotNow` was serializing full `DocumentNode[]` (7–10 MB for long sessions) into a JSON blob every 30s on the main render thread, causing 126ms long-tasks and renderer OOM crashes (`error_code=-36861`)
- **Schema v2**: The snapshot payload is now always < 1 KB — `highWaterMark` + `documentState` (collapsedNodeIds, pinnedNodeIds, scrollPosition, filter) + `paneState.detailsOpen`; `nodes[]` is removed entirely
- **v2 restore path**: On open, fetches the trailing `RESTORE_WINDOW_LINES = 5_000` lines from `blockfile output` via `BlockfileReadRangeCommand` and replays through `parseHistoryLines`; older history remains on disk and is accessible via load-older
- **Legacy support**: v1 snapshots (with `nodes[]`) continue to restore via the existing path; schema version mismatch falls back to NDJSON ring-buffer replay

## Details

The session history is now unlimited (disk is the only constraint). The `RESTORE_WINDOW_LINES` constant is a render viewport, not a storage cap — it controls how many lines are loaded on pane open for a fast initial render. All prior history remains readable via the existing load-older pagination.

Closes the renderer crash reported in `docs/analysis/ANALYSIS_RENDERER_CRASH_WRITE_STATE_OOM_2026_06_12.md`.

Spec: `docs/specs/SPEC_WRITE_STATE_NDJSON_RESTORE_2026_06_12.md`

## Test plan

- [ ] Open an existing agent pane with a long session (>200 lines) — verify history restores and scroll position is preserved
- [ ] Let `writeSnapshotNow` fire (30s idle) — verify no memory spike in DevTools Memory tab
- [ ] Close and reopen pane — verify v2 restore path (check console for `v2 restore: N nodes from lines [X, Y)`)
- [ ] Open a pane with an old v1 snapshot — verify legacy path still works (console: `v1 restore: N nodes from snapshot`)
- [ ] Scroll to top, trigger load-older — verify pages load correctly past the initial 5000-line window

### DIFF
--- a/.changesets/1781283396-fix-agent-write-state-schema-v2-replace-nodes-snapshot-with--tq14.md
+++ b/.changesets/1781283396-fix-agent-write-state-schema-v2-replace-nodes-snapshot-with--tq14.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): write-state schema v2 — replace nodes[] snapshot with lightweight overlay + NDJSON restore to eliminate renderer OOM
--- a/.changesets/1781283396-fix-agent-write-state-schema-v2-replace-nodes-snapshot-with--tq14.md
+++ b/.changesets/1781283396-fix-agent-write-state-schema-v2-replace-nodes-snapshot-with--tq14.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-fix(agent): write-state schema v2 — replace nodes[] snapshot with lightweight overlay + NDJSON restore to eliminate renderer OOM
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

AgentA-asaf:
Per the review discussion: scoped this PR to the same-block reopen case (the renderer OOM fix). Cross-block structural continuation now gracefully falls back to NDJSON replay (no orphaning, no blank pane) and is deferred to a unified per-agent log — tracked in #1397. See spec §15 for the design.

PR #1362, opened 2026-06-12
feat(blockfile): output.idx byte-offset index for O(1) line seek on large sessions

AgentA-asaf:
## Summary

- **Problem**: `blockfile:read_range` loads the entire NDJSON file into memory before slicing by line number. For a 100GB session this is catastrophic — a 50+ second scan that OOMs the sidecar.
- **Solution**: `handle_append_block_file` now maintains a companion `output.idx` file. Each entry is the start byte of one line (u64-LE, 8 bytes). `blockfile:read_range` checks for the index first and calls `read_at` to load only the requested bytes — O(1) seek regardless of file size.
- **Fallback**: sessions without `output.idx` (all existing sessions) continue to use the existing full-scan path — fully backward compatible.

## How it works

**Write path** (`shell.rs::handle_append_block_file`):
1. Stat the file before append to get `prev_size` (start byte for new data)
2. After successful `append_data("output", ...)`, scan `data` for `\n` bytes
3. For each complete line, append its start byte as u64-LE to `output.idx`

**Read path** (`app_api.rs::register_blockfile_read_range`):
1. Check if `output.idx` exists; total lines = `idx_size / 8`
2. Read `idx[offset * 8..+8]` → `byte_start`; read `idx[(offset+limit) * 8..+8]` → `byte_end` (or output file size if at end)
3. `read_at("output", byte_start, byte_end - byte_start)` — loads only the needed bytes

**Index size**: 8 bytes per line. A 100M-line session = 800MB index — acceptable alongside the multi-GB output file.

## Test plan

- [ ] Write N lines to a block, verify `output.idx` exists and has `N * 8` bytes
- [ ] `blockfile:read_range` with offset/limit on a session with index returns correct lines
- [ ] `blockfile:read_range` on a session WITHOUT `output.idx` still returns correct lines (fallback path)
- [ ] Load a pane with a long session — verify the DevTools Network tab shows the sidecar returning only the requested line range's bytes (not the full file size)

### DIFF
--- a/.changesets/1781284240386-feat-blockfile-output-idx-o1-seek-dbruxcdm980g7ihb.md
+++ b/.changesets/1781284240386-feat-blockfile-output-idx-o1-seek-dbruxcdm980g7ihb.md
@@ -1,0 +1,12 @@
+---
+'agentmux-srv': patch
+---
+
+feat(blockfile): output.idx byte-offset index for O(1) line seek
+
+Appends a companion `output.idx` file alongside every NDJSON output file.
+Each entry is a u64-LE value recording the byte offset where line N starts.
+`blockfile:read_range` checks for this index first; if present it seeks
+directly to the requested line range without loading the full file.
+
+Existing sessions without an index fall back to the previous full-scan path.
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1363, opened 2026-06-12
fix(scroll): stickToBottom can now disengage on short conversations

AgentX-asaf:
## Root cause

`isNearBottom` (in `anchor.ts`) uses a fixed 200 px threshold. For short conversations where `scrollHeight - clientHeight < 200 px`, **every scroll position** satisfies `distFromBottom < 200`, so `stickToBottom` is impossible to disengage.

Consequence with three fresh agents open:
- Scroll wheel moves the position by a few px, a streaming token arrives, `queueMicrotask(scrollTo(MAX_SAFE_INTEGER))` fires and snaps back → wheel appears broken
- Fast drag same: snaps back faster than the user perceives movement
- Slow drag "works" only because tokens don't arrive between each slow step
- Long conversations work fine because `maxScroll >> 200`

## Fix

**`anchor.ts` — `isNearBottom`**

When `maxScroll < threshold`, cap the sticky zone at `max(1, floor(maxScroll / 2))` instead of the full threshold. This guarantees a reachable "scrolled away" position exists regardless of content height.

```
maxScroll=50  →  effectiveThreshold=25  →  sticky only for scrollTop > 25
maxScroll=1   →  effectiveThreshold=1   →  sticky only at absolute bottom
maxScroll≥200 →  effectiveThreshold=200 →  unchanged (long conversations)
```

**`AgentDocumentVirtualList.tsx` — microtask guard**

The queued `queueMicrotask` was not checking `stickToBottom` at fire time. A token arriving just before the user scrolls away would queue a snap that fired after disengagement. Added a re-check inside the callback.

## Tests

New cases in `anchor.test.ts`:
- short conversation: caps sticky zone at half range
- no overflow: always near bottom
- range=1: only at absolute bottom

All 16 tests pass.

## Checklist
- [x] `anchor.ts` fix + tests
- [x] `AgentDocumentVirtualList.tsx` microtask stale-guard
- [x] Changeset added

### DIFF
--- a/.changesets/1781257955-fix-agent-inject-bash-agentmux-bashwrap-into-permissions-all-suy5.md
+++ b/.changesets/1781257955-fix-agent-inject-bash-agentmux-bashwrap-into-permissions-all-suy5.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-fix(agent): inject Bash(agentmux-bashwrap*) into permissions.allow so bashwrap exec is not blocked
+fix(agent): inject Bash(agentmux-bashwrap *) into permissions.allow so bashwrap exec is not blocked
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #1364, opened 2026-06-12
fix(macos): window drag + right-click context menus coexist on title bar

a5af:
## What

macOS title-bar **window drag** and **right-click context menus** now work on the same surface. Previously they were mutually exclusive: the header used `-webkit-app-region: drag`, which lets the OS move the window but makes Chromium swallow every event (including `contextmenu`) on those regions — so right-click never fired on empty title-bar space, and the only workaround broke dragging.

Fixes the reported regression — right-click context menus stopped firing on empty title-bar areas (still worked on tabs/widgets).

## How — adopt the Linux JS-driven model, with a host-side native drag

- **`frontend/app/hook/useWindowDrag.darwin.ts`** — header stays HTCLIENT; a **left-button-only** JS listener sends one `start_window_drag` IPC on a 4px threshold crossing, and dblclick maximizes. Right/middle-click pass straight through to `contextmenu`.
- **`frontend/app/window/window-header.darwin.scss`** — drop `-webkit-app-region: drag` and the `.tab-bar-fill { no-drag }` opt-out it required. The whole header is now HTCLIENT, so the shared `onContextMenu` on `.window-header` finally fires on empty areas.
- **`agentmux-cef/src/ui_tasks.rs`** — macOS `start_window_drag` runs a native AppKit drag: reach the NSWindow via `[NSView window]` and call `[NSWindow performWindowDragWithEvent:[NSApp currentEvent]]` through raw libobjc FFI (same idiom as `ensure_macos_native_window_buttons`). The window server moves the window — no per-frame IPC, **no patched libcef**. `StartWindowDragTask` (`CefWindow::BeginWindowDrag`) is now Linux-only.

### Why host-side native drag (not patched libcef)

Stock libcef ships no programmatic drag API, and the fork's `BeginWindowDrag` is Ozone-only (a no-op on macOS). Patching libcef would chain every macOS release to a patched-Chromium framework and force a re-port on each Chromium bump. Reaching the NSWindow directly keeps macOS on **stock libcef**.

## Verification

- `cargo check -p agentmux-cef` — clean (the objc FFI compiles).
- `stylelint` on the scss — clean; `esbuild` transform of the hook — valid TS; imports + return shape match the working Linux hook exactly.

## ⚠️ Needs on-device testing

`performWindowDragWithEvent:` needs the live mouse event. Because `start_window_drag` is async, the host uses `[NSApp currentEvent]` (button still held, browser mid-drag → it should be the current left-mouse-dragged event) and **guards against nil** (skips rather than passing nil — never UB). Please verify on-device that dragging actually moves the window. If `currentEvent` proves stale/nil in practice, the fix is localized: swap `begin_macos_native_window_drag` to a manual `nextEventMatchingMask` move loop (documented in the function's doc comment).

### DIFF
--- a/.changesets/1781253993-fix-macos-right-click-contextmenu-on-empty-title-bar-area-hjnb.md
+++ b/.changesets/1781253993-fix-macos-right-click-contextmenu-on-empty-title-bar-area-hjnb.md
@@ -9,7 +9,8 @@ window, but Chromium swallows every event (including `contextmenu`) on those
 regions — so right-click context menus never fired on empty title-bar space,
 and the only workaround broke dragging. Switch macOS to the JS-driven drag
 model already used on Linux: the header stays HTCLIENT (right-click works
-everywhere) and a left-button-only drag is handed off to a native AppKit drag
-(`[NSWindow performWindowDragWithEvent:]`) in the host — no patched libcef.
+everywhere) and a left-button-only drag is handed to the host, which runs a
+manual move loop — pumping the drag events and repositioning the window until
+the button is released — with no patched libcef.
 Dragging the window and right-clicking the title bar now both work on the same
 surface.
--- a/.changesets/1781253993-fix-macos-right-click-contextmenu-on-empty-title-bar-area-hjnb.md
+++ b/.changesets/1781253993-fix-macos-right-click-contextmenu-on-empty-title-bar-area-hjnb.md
@@ -1,0 +1,15 @@
+---
+type: patch
+---
+
+fix(macos): window drag and right-click context menus now coexist on the title bar
+
+The macOS title bar used `-webkit-app-region: drag` so the OS could move the
+window, but Chromium swallows every event (including `contextmenu`) on those
+regions — so right-click context menus never fired on empty title-bar space,
+and the only workaround broke dragging. Switch macOS to the JS-driven drag
+model already used on Linux: the header stays HTCLIENT (right-click works
+everywhere) and a left-button-only drag is handed off to a native AppKit drag
+(`[NSWindow performWindowDragWithEvent:]`) in the host — no patched libcef.
+Dragging the window and right-clicking the title bar now both work on the same
+surface.
--- a/agentmux-cef/src/client/handlers.rs
+++ b/agentmux-cef/src/client/handlers.rs
@@ -100,16 +100,17 @@ wrap_focus_handler! {
 
 // ---------------------------------------------------------------------------
 // DragHandler — handles `-webkit-app-region: drag` regions reported by the
-// renderer (used on macOS/Windows where native draggable regions work).
+// renderer (used on Windows, where native draggable regions work).
 //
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

a5af:
Addressed both codex findings in edbc6bed.

**P1 — `[NSApp currentEvent]` unreliable under async IPC.** Replaced `performWindowDragWithEvent:[NSApp currentEvent]` with a host-side manual move loop, mirroring the existing Windows `Win32BeginMoveTask`. On the UI thread it pumps `NSLeftMouseDragged`/`NSLeftMouseUp` and repositions via CEF `set_bounds`, so the window tracks the cursor with **no** dependency on `currentEvent` being the original mouse-down — the drag can't silently fail to start. Deltas are mouse-location based (points == DIP on macOS); only the vertical axis is flipped (NSEvent y-up → CEF y-down). Like the Windows loop it blocks the UI thread until mouse-up, the accepted trade-off for a window drag.

**P2 — existing `no-drag` controls not honored.** Added `data-drag-region="false"` to the `.traffic-lights` container. It was the only macOS header control opting out solely via `-webkit-app-region: no-drag` (which the JS hit-test doesn't read now that the header is HTCLIENT); every other widget (tabs, action widgets, hamburger) already carries the data attribute. A press-and-move on a traffic-light button no longer starts a window drag.

Still pending on-device verification: that the drag tracks the cursor in the correct direction (the y-flip) — noted in the PR description.

a5af:
Addressed the new round in b1ecdfa0.

**P1 — move loop could hang the UI thread (codex + reagent).** The loop blocked on `nextEventMatchingMask:` with `distantFuture` and no guard, so a fast flick-and-release (NSLeftMouseUp consumed before the posted task starts pumping) would wait forever and freeze the app. Mirrored the Windows `Win32BeginMoveTask` guards using AppKit's **live** button state:
- **Pre-loop bail:** `[NSEvent pressedMouseButtons] & 1 == 0` → return without dragging (the Windows `!lbutton_down()` bail).
- **100ms timeout instead of `distantFuture`:** fresh deadline each iteration; on timeout, re-check `pressedMouseButtons` and exit if the button is up — catches a missed `NSLeftMouseUp` (the Windows 100ms `SetTimer` wake). Active drags are unaffected: matching events return immediately, so the timeout only fires while the pointer is idle.

**P2 ×3 — stale comments.** Updated `post_start_drag`'s doc, the `useWindowDrag.darwin.ts` header, and `window-header.darwin.scss` — all three now describe the manual `set_bounds` move loop instead of `performWindowDragWithEvent:`.

`cargo check -p agentmux-cef` clean; scss lints clean.

Note: I scoped this to the flagged hang. The Windows mirror also synthesizes a balancing mouse-up so Chromium's mousedown isn't left unbalanced (its `reagent P1 / §5.1`); that wasn't raised here and the renderer's JS drag state is reset via `window_drag_ended` regardless, so I left it for a follow-up if on-device testing shows a stuck-button symptom.

PR #1365, opened 2026-06-12
fix(bashwrap): keep CONIN writer alive until after child.wait()

AgentX-asaf:
## Problem

`run_via_pty` called `pair.master.take_writer()`, wrote the DSR response (`\x1b[1;1R`), then immediately dropped the writer in a scoped block. On Windows, closing the ConPTY CONIN pipe's write-end while a process is attached sends `CTRL_C_EVENT` to the child. Result: every bash command exited 130 in ~30 ms with no output.

The module comment already documented the ConPTY lifetime invariant for `pair.master` — the writer requires the same treatment.

## Fix

Return the writer out of the DSR block and move it into the `spawn_blocking` wait closure alongside `pair`, so both drop only after `child.wait()` returns:

```rust
// before: writer dropped here ← sends CTRL_C_EVENT
{
    let mut writer = pair.master.take_writer()?;
    writer.write_all(b"\x1b[1;1R")?;
}

// after: writer held until child exits
let writer = {
    let mut w = pair.master.take_writer()?;
    w.write_all(b"\x1b[1;1R")?;
    w
};
// ... later in spawn_blocking:
drop(writer);  // safe — child already exited
drop(pair);
```

The `exec </dev/null;` command prefix still redirects bash's stdin to `/dev/null` after startup, so no future writes to the writer are needed — the handle just needs to stay open.

## Test

Build passes (`cargo build -p agentmux-bashwrap`). Live test: bash commands that previously exited 130 immediately now run to completion.

### DIFF
--- a/.changesets/1781302822-fix-bashwrap-keep-conin-writer-alive-until-after-child-wait--dg0t.md
+++ b/.changesets/1781302822-fix-bashwrap-keep-conin-writer-alive-until-after-child-wait--dg0t.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(bashwrap): keep CONIN writer alive until after child.wait() — prevents CTRL_C_EVENT exit 130 on Windows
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1366, opened 2026-06-12
fix(term): file path links now open in Explorer/Finder on click

AgentX-asaf:
## Summary

- Implements the missing `reveal_in_file_explorer` IPC backend handler — the frontend called `invokeCommand("reveal_in_file_explorer")` but `ipc.rs` had no matching arm, so every file-path link click silently dropped
- Adds `vscode://` to `open_external`'s scheme allowlist so that paths with a `:line` suffix (which use `vscode://file/<path>:<line>`) actually open in VS Code instead of being rejected

## Platform behaviour

| Platform | File path (no line) | Path with `:line` |
|----------|--------------------|--------------------|
| Windows  | `explorer /select,<path>` — opens Explorer with file selected | `vscode://file/<path>:<line>` via `open_external` |
| macOS    | `open -R <path>` — reveals in Finder with file selected | same |
| Linux    | `xdg-open <parent dir>` — opens parent folder | same |

## Test plan

- [ ] In a terminal, `echo /absolute/path/to/file.ts` — hover should underline, click should open Explorer/Finder with the file selected
- [ ] `echo src/components/App.tsx` from a directory with that file — relative path resolves and opens
- [ ] `echo /some/file.ts:42` — click should open VS Code at line 42 (requires VS Code installed)
- [ ] Invalid scheme (`ftp://...`) still rejected by `open_external`

### DIFF
--- a/.changesets/1781305151-fix-term-file-path-links-now-open-in-explorer-finder-on-clic-sd4e.md
+++ b/.changesets/1781305151-fix-term-file-path-links-now-open-in-explorer-finder-on-clic-sd4e.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): file path links now open in Explorer/Finder on click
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1368, opened 2026-06-12
fix(bashwrap): remove exec </dev/null prefix — root cause of exit-130 on ConPTY

AgentX-asaf:
## Root cause

Every PTY-backed bashwrap command was exiting with code 130 (SIGINT) in ~20ms, producing `chunks_published=0`. Traced to the `exec </dev/null;` prefix injected into every `bash -c` command string (line 476, commit `199818cd`).

**What happened:** When bash ran `exec </dev/null`, it closed its PTY stdin handle from within the child process. On Windows ConPTY, closing the child's console input handle triggers a `CTRL_C_EVENT` delivery to the process group — the same mechanism as dropping the host-side CONIN write handle. Bash received SIGINT and exited before the actual command ran, producing zero output.

This is why PR #1365's writer-lifetime fix (which correctly addressed the host-side CONIN drop) did not resolve the issue — there were two independent CTRL_C sources.

## Fix

Remove the `exec </dev/null;` prefix. Bash's stdin remains the PTY slave for the duration of the command. Commands that explicitly read stdin (plain `cat`, `read`) will block on it — but typical agent commands (`git`, `ls`, `npm`, `cargo`, `python`) do not read stdin and are unaffected.

## Verification

```
$ AGENTMUX_AUTH_KEY=test AGENTMUX_LOCAL_URL=http://localhost:1 \
    ./target/release/agentmux-bashwrap.exe exec --tool-id test-fix \
    --b64-cmd ZWNobyBoZWxsb19mcm9tX3B0eV9maXg

<exited 0 in 2.36s>
hello_from_pty_fix
```

Before fix: `<exited 130 in 0.02s>` with empty body, `chunks_published=0` in every log entry ever recorded.

## Also in this PR

- `exit_code` info log in the PTY wait task (was missing — made diagnosing this hard)
- `raw_bytes` / `post_strip` debug trace in `pty_reader_loop` (confirms whether data arrives before strip)
- Retrospective written to `docs/retros/RETRO_BASHWRAP_EXIT130_2026_06_12.md` (not included in commit — local analysis doc)

### DIFF
--- a/.changesets/1781315893-fix-bashwrap-wrap-commands-in-a-dev-null-brace-group-redirec-dkln.md
+++ b/.changesets/1781315893-fix-bashwrap-wrap-commands-in-a-dev-null-brace-group-redirec-dkln.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(bashwrap): wrap commands in a /dev/null brace-group redirect instead of exec </dev/null — the exec form closed the child's ConPTY console input and ConPTY killed every streamed bash command with exit 130 before it ran; the group redirect gives stdin-readers EOF without firing ctrl-c (#1368)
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

PR #1369, opened 2026-06-13
fix(term): make xterm-6 overlay scrollbar clickable (z-index above link-layer)

AgentX-asaf:
## Symptom

Clicking the terminal scrollbar does nothing — the thumb won't drag, the track won't page-scroll.

## Root cause (xterm 5 → 6 migration debt)

xterm **6.0.0** replaced the native-scrollbar viewport with a custom **scrollable-element + overlay scrollbar DIVs**. The live DOM is now:

```
.xterm
├── .xterm-viewport            ← now EMPTY (scrollHeight === clientHeight always)
├── .xterm-scrollable-element
│   ├── .xterm-screen          ← text + canvases, incl. .xterm-link-layer (z-index: 2)
│   ├── .scrollbar.horizontal
│   └── .scrollbar.vertical > .slider   ← the REAL scrollbar (z-index: auto)
└── canvas
```

`.xterm-link-layer` (the canvas the WebLinks / FilePath link providers hit-test) carries **`z-index: 2`**. The overlay scrollbar is a *later sibling* of `.xterm-screen`, so by DOM order it should paint on top — but **an explicit positive z-index beats `z-index: auto` regardless of DOM order**, so the link layer painted over the scrollbar.

While `.xterm-screen` still had `pointer-events: none`, this was harmless — the link-layer *inherited* it and was click-through. Removing `pointer-events: none` to fix link hover (the earlier regression) let the `z:2` link-layer **swallow every click in the scrollbar lane**.

As a side effect of the xterm-6 rework, the `customFit` scrollbar-reservation logic and the `term.scss ::-webkit-scrollbar` rules (both written for xterm 5's native scrollbar) are now dead — `viewport.scrollHeight > clientHeight` is permanently false. They're left in place as no-ops/DOM-renderer-fallback and annotated; the real scrollbar is now styled in the new `.scrollbar` block.

## Fix

```scss
.xterm-scrollable-element > .scrollbar {
    z-index: 10;            // above .xterm-link-layer (z:2) and .xterm-helpers (z:5)
    .slider { cursor: pointer; background-color: var(--scrollbar-thumb-color); ... }
}
```

## Verification (live, via CDP `elementFromPoint` over the scrollbar lane)

| | hit-test result |
|---|---|
| before | `CANVAS.xterm-link-layer` — clicks blocked |
| after  | `DIV.slider` — clicks reach the scrollbar |

`elementFromPoint` resolves exactly what a real mouse click hits, so this is conclusive. The scrollbar occupies only the right 14px → link hover over the text area is unaffected; with no overflow the scrollbar is `visibility:hidden` and not hit-tested, leaving the full width available for hover.

## Note on terminal bug cluster

This is the same xterm 5→6 migration debt behind the recent run of terminal bugs (pointer-events/link hover, scrollbar). The custom terminal layer (`customFit`, `term.scss` scrollbar styling) was written against xterm 5's DOM and is subtly mismatched to xterm 6. Worth a follow-up pass to retire the dead `customFit` reservation path and `::-webkit-scrollbar` rules now that xterm 6 owns scrollbar layout.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.44.1"
+version = "0.44.2"
 dependencies = [
  "anyhow",
  "base64",
@@ -28,7 +28,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.44.1"
+version = "0.44.2"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
@@ -58,7 +58,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-common"
-version = "0.44.1"
+version = "0.44.2"
 dependencies = [
  "dirs",
  "serde",
@@ -71,7 +71,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-launcher"
-version = "0.44.1"
+version = "0.44.2"
 dependencies = [
  "agentmux-common",
  "chrono",
@@ -93,7 +93,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-srv"
-version = "0.44.1"
+version = "0.44.2"
 dependencies = [
  "agentmux-common",
  "async-stream",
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

AgentX-asaf:
Superseded by #1370 — the scrollbar z-index fix from this branch is included in #1370, which also carries the cursor fix, the xterm-5 dead-code reconciliation, the smoke guard, and is converted to the changesets workflow. Merging there instead to avoid the stacked-version churn.

PR #1370, opened 2026-06-13
fix(term): xterm-6 terminal scrollbar clickable + default cursor, retire dead xterm-5 code

AgentX-asaf:
**Stacked on #1369** (base: `agentx/fix-terminal-scrollbar-click`). The follow-up reconciliation you asked for after the scrollbar root-cause turned out to be xterm 5→6 migration debt.

## Why

Nearly every recent terminal bug — the URL-hover regression and the scrollbar click — traced to one root cause: **xterm 6.0.0 reworked the viewport/scrollbar DOM**, but the custom terminal layer was written for xterm 5. The leftover code wasn't just dead; its confident-but-wrong comments actively sent debugging down the wrong path.

## xterm 6 facts (verified live via CDP)

- `.xterm-viewport` is now an **empty positioning anchor**. The real scroll container is `.xterm-scrollable-element` with a transform-driven **overlay** scrollbar (`.scrollbar.vertical > .slider`) — not a native `::-webkit-scrollbar`.
- Consequently `viewport.scrollHeight > clientHeight` is **permanently false**.
- FitAddon v0.11 still unconditionally subtracts a 14px gutter (`scrollback>0 ? 14 : 0`), which `customFit` legitimately reclaims — so that part stays.

## Removed (all provably dead under xterm 6)

| Removed | Why dead |
|---|---|
| `customFit` `scrollbarVisible`/`reservation` branch | reservation was always 0 → only ever took the full-width path (kept) |
| native-scrollbar-width injection (`scrollbarStyleEl`, injected `::-webkit-scrollbar` `<style>`) | targets a pseudo-element xterm 6 doesn't create |
| `setupScrollbarRefit()` + viewport ResizeObserver | watched a toggle on the now-empty viewport that never changes |
| `SCROLLBAR_WIDTH`, `lastScrollbarVisible`, `scrollbarStyleEl` (+ dispose) | only referenced by the above |
| `term.scss .xterm-viewport::-webkit-scrollbar*` rules | no native scrollbar in xterm 6 |

**Kept:** the full-width cols recompute (load-bearing — reclaims FitAddon's dead gutter), `overflow-y: auto` (suppresses a permanent native track the vendored `xterm.css`'s `overflow-y: scroll` would paint on the empty viewport), and the `.scrollbar` z-index fix from #1369.

**Behavior is unchanged** — every removed branch was already inert. `termwrap.ts` drops ~91 lines, `term.scss` ~37.

## Regression guard

`scripts/cdp-term-smoke.mjs` — a CDP hit-test guard. Via `document.elementFromPoint` (exactly what a real click resolves) it asserts:
1. scrollbar lane → `.slider` (not `.xterm-link-layer`) — else scrollbar unclickable
2. text area → `.xterm-link-layer` + `.xterm-screen` pointer-events:auto — else URL/file hover dead

**Either assertion would have caught its regression instantly.** Run `node scripts/cdp-term-smoke.mjs` (dev=9223, release=9222) against an instance with a terminal open. Validated: clean syntax, connects, evaluates, and correctly skips (exit 2) when no terminal is mounted; the two probes are the same ones used to prove #1369 live.

### DIFF
--- a/.changesets/1781315995-fix-term-xterm-6-terminal-scrollbar-is-clickable-and-shows-t-on2m.md
+++ b/.changesets/1781315995-fix-term-xterm-6-terminal-scrollbar-is-clickable-and-shows-t-on2m.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): xterm-6 terminal scrollbar is clickable and shows the default cursor — lift the overlay scrollbar above the link-layer canvas (z-index) and force cursor:default; retire the dead xterm-5 fit/reservation + native-scrollbar code and add a CDP hit-test smoke guard (#1369, #1370)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agentmux-bashwrap"
-version = "0.44.3"
+version = "0.44.4"
 dependencies = [
  "anyhow",
  "base64",
@@ -28,7 +28,7 @@ dependencies = [
 
 [[package]]
 name = "agentmux-cef"
-version = "0.44.3"
+version = "0.44.4"
 dependencies = [
  "agentmux-common",
  "axum 0.8.9",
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #300, opened 2026-04-05
fix: Win11 focus border black color + packaging script resilience

AgentX-asaf:
## Summary

- **Fix Win11 black focus border**: The focused pane border rendered black on Windows 11 because the default accent color used an RGB variable from `:root`, which CEF's compositor fails to resolve on `backdrop-filter` elements. Agent colors (HEX inline styles) rendered correctly. Unified both paths into a single `--block-focus-color` inline style that always uses HEX.
- **Fix packaging script MSYS2 regression**: `scripts/package-cef-portable.sh` failed silently due to MSYS2 bash stdout FD corruption combined with `set -euo pipefail`. Dropped `set -e`, added resilient `log()`/`die()` helpers with explicit error checks.
- **Fix `powershell` → `pwsh`**: Changed PowerShell 5.1 call to PowerShell 7 per project rules.

## Changes

| File | Change |
|------|--------|
| `frontend/app/block/blockframe.tsx` | Replaced `--block-agent-color` + `has-agent-color` class with unified `--block-focus-color` inline style |
| `frontend/app/block/block.scss` | Collapsed two CSS rules (agent color override + default) into single rule using `var(--block-focus-color)` |
| `scripts/package-cef-portable.sh` | Dropped `set -e`, added `log()`/`die()` helpers, explicit error checks, `powershell` → `pwsh` |
| Version files | Bumped to v0.33.44 |

## Root cause

CEF on Windows 11 has a compositor bug where `:root` CSS custom properties containing RGB values render as black when applied to elements with `backdrop-filter`. The `.block-mask` element uses `backdrop-filter: blur(0.1px)` to force correct GPU compositing order above xterm's scroll layer. Agent-loaded terminals worked because their border color was set as an inline HEX style, bypassing the `:root` variable path entirely.

## Test plan

- [ ] Launch portable build (v0.33.44 on Desktop)
- [ ] Open a standard terminal pane — focus border should be blue (#419FE0), not black
- [ ] Open an agent terminal (e.g., AgentX with red) — focus border should be red
- [ ] Verify unfocused panes still show dark border
- [ ] Run `task cef:package:portable` — should complete without silent failures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1371, opened 2026-06-13
fix(block): focused ring falls back to accent when agent color is unusable; "Terminal" is not an agent id

AgentY-asaf:
## The bug (the "Win11 terminal pane has no focus highlight" saga)

Plain terminal panes showed **no border highlight when selected** — the focused border painted **literal black**. Root-caused live (CDP + objects.db): a machine-local PowerShell profile exported `AGENTMUX_AGENT_ID="Terminal"` + `AGENTMUX_AGENT_COLOR="#000000"` into every plain pwsh pane. Shell integration persisted that into block `cmd:env`, `detectAgentFromEnv` accepted `"Terminal"` as a real agent, `has-agent-color` applied, and `block.scss`'s `.block-focused.has-agent-color .block-mask { border-color: var(--block-agent-color) }` beat the accent rule → focused ring = `rgb(0,0,0)`.

Verified on a healthy compositor (post-#1354: `gpu_compositing: enabled`): `block-focused` was correctly applied; the ring was simply painted black. Claw agent panes (AgentX `#ef4444`, AgentY `#f87171`) highlighted fine through the identical mechanism — which is what falsified every compositor theory.

**History note:** this same symptom was chased as a Win11 DWM/compositor bug in #267/#296/#300 (April). #300's "border renders black on Win11" is this signature verbatim. The per-machine confounders were the shell profile and the now-fixed GPU-disabled regime (#1354) — not Windows 11 rendering.

## The fix — env data must never erase the selection affordance

1. **`detectAgentFromEnv` rejects `"Terminal"`** (case-insensitive) — it's the default pane title, not an agent identity. Whitespace-only ids now return null too.
2. **`blockAgentColor` gates the ring color with `isUsableFocusRingColor`** — unparseable, low-alpha (< 0.5), or near-black (WCAG relative luminance < 0.03) identity colors fall back to the theme accent ring. The pane **header** keeps the raw identity color (separate `agentColor` memo, untouched); only the focus ring is gated.

Threshold calibration: claw palette passes — including the darkest legit color, AgentA's `#1e3a5f` (L≈0.041). `#000` (L=0), `#111` (L≈0.005), `#222`-class near-blacks (L≈0.016) are rejected.

## Tests

12 new unit tests in `autotitle.test.ts` (Terminal rejection, palette acceptance, near-black/transparent/unparseable rejection, hex/rgb/rgba syntaxes). Full file: 44/44 pass. `tsc --noEmit` clean on touched files.

## Out of scope (tracked in the analysis doc)

- Single `paneFrameColors` resolver consolidating the 3 scattered color sources (meta override / agent env / accent).
- Verifying `will-change: transform` (#1222) composites the ring above xterm's scroll layer on Windows now that GPU is healthy.

### DIFF
--- a/.changesets/1781315130-fix-block-focused-ring-falls-back-to-accent-when-agent-color-uh2a.md
+++ b/.changesets/1781315130-fix-block-focused-ring-falls-back-to-accent-when-agent-color-uh2a.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(block): focused ring falls back to accent when agent color is unusable; 'Terminal' is not an agent id
--- a/frontend/app/block/autotitle.test.ts
+++ b/frontend/app/block/autotitle.test.ts
@@ -3,12 +3,75 @@
 
 import { assert, describe, test } from "vitest";
 import {
+    detectAgentFromEnv,
     detectAgentFromPath,
     detectAgentFromWorkspacesPath,
     generateAutoTitle,
     getEffectiveTitle,
+    isUsableFocusRingColor,
     shouldAutoGenerateTitle,
 } from "./autotitle";
+
+describe("detectAgentFromEnv", () => {
+    test("detects agent from AGENTMUX_AGENT_ID", () => {
+        assert.equal(detectAgentFromEnv({ AGENTMUX_AGENT_ID: "AgentY" }), "AgentY");
+        assert.equal(detectAgentFromEnv({ AGENTMUX_AGENT_ID: "  Agent2  " }), "Agent2");
+    });
+
+    test("returns null for missing/blank env", () => {
+        assert.equal(detectAgentFromEnv(undefined), null);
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentY-asaf:
**F4 follow-up verified (post-merge):** ran a live compositor test on Windows 11 with healthy GPU (post-#1354) — overrode a focused terminal pane's ring color to sentinel green via CDP, captured the composited border strip (`Page.captureScreenshot`, `fromSurface`), and pixel-sampled it: 20.5% sentinel pixels, matching the expected 2px-ring geometry. The `will-change: transform` promotion from #1222 correctly composites `.block-mask` above xterm's `.xterm-viewport` scroll layer on Windows. No remaining open legs on the focus-ring saga. Test tool: `cdp-verify-ring-compositing.mjs` (agenty workspace).

PR #1372, opened 2026-06-13
fix(term): terminal fills block body edge-to-edge (remove pane gaps)

AgentX-asaf:
## Symptom

Terminal panes had a visible gap/frame on the right, bottom, and left — same width on the right as height on the bottom.

## Root cause

`.term-connectelem` (the terminal body element) had:
```scss
margin: 5px;
margin-left: var(--space-1);  /* 4px */
```
So every terminal pane was inset by 5px top/right/bottom and 4px left, and that inset exposed the `.block-frame-default-inner` background (`rgba(13,14,15,0.5)`) as a frame around the terminal — exactly the "right == bottom, also left" gaps reported.

## Fix

`margin: 0` — the terminal fills the block body edge-to-edge.

## Verified (CDP layout measurement, `scripts/cdp-gap-probe.js`)

| gap (connect → block frame) | before | after |
|---|---|---|
| left | 4px | **0** |
| right | 5px | **0** |
| top | 5px | **0** (33px = the term header, not a gap) |
| bottom | 5px | **0** |

The only residue is the inherent sub-cell grid remainder (text grid can't fill a fractional last cell: ~5px right / ~2px bottom). A zoomed corner screenshot confirms it blends into the near-black terminal background — no visible seam. If you still want that absolute-zeroed, the follow-up is to paint the terminal container background to match the theme bg; flagging rather than doing it blind since it can interact with transparent terminal themes.

### DIFF
--- a/.changesets/1781316529-fix-term-remove-the-5px-term-connectelem-margin-that-framed--gfl7.md
+++ b/.changesets/1781316529-fix-term-remove-the-5px-term-connectelem-margin-that-framed--gfl7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): remove the 5px term-connectelem margin that framed every terminal pane with block-background gaps on all four sides — the terminal now fills the block body edge-to-edge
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1375, opened 2026-06-13
fix(term): terminal scrollbar flush to the pane's right edge

AgentX-asaf:
## Symptom

A gap to the **right of the terminal scrollbar** — annoying, and its width **changed with zoom level**.

## Root cause

`.term-connectelem` is a `display: flex` row, and `.xterm` is a flex item that sizes to its **content** width (`cols × cellWidth`), not the full pane width. That left a strip of free flex space — `connectWidth − cols·cellWidth` (0 to <1 cell) — on the right. The xterm-6 overlay scrollbar is anchored `right: 0` inside `.xterm-scrollable-element` (which inherits `.xterm`'s narrow width), so the scrollbar floated that strip-width away from the pane edge. Because the remainder is `width mod cellWidth`, it changed every time the cell size changed → **zoom-dependent gap**.

## Fix

```scss
.terminal { flex-grow: 1; }   /* .xterm fills the connect width */
```

`.xterm` now fills the connect width, so `.xterm-scrollable-element` and its `right: 0` scrollbar sit **flush** against the pane's right edge — at every zoom, since this no longer depends on the grid remainder. The text grid (`.xterm-screen`, still `cols·cellWidth`) stays left-aligned; the now-interior remainder is terminal background, and the fading overlay scrollbar covers the rightmost sliver.

## Verified (CDP)

`scrollbar-right-edge → pane-right-edge` gap: **7px → 0**. `.xterm` width 1064 → 1071 (= connect width). A flush-edge screenshot confirms the terminal background runs to the pane border with no strip.

### DIFF
--- a/.changesets/1781334657-fix-term-terminal-scrollbar-sits-flush-against-the-pane-s-ri-gvrl.md
+++ b/.changesets/1781334657-fix-term-terminal-scrollbar-sits-flush-against-the-pane-s-ri-gvrl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): terminal scrollbar sits flush against the pane's right edge — grow .xterm to fill the flex connect width so the overlay scrollbar (anchored right:0) no longer floats a zoom-varying sub-cell remainder away from the edge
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1378, opened 2026-06-13
fix(tear-off): keep the brain splash covering bootstrap until content settles

AgentX-asaf:
## Symptom

Tab tear-off is fast (the warm window pool makes the window appear instantly), **but it flashes** during the bootstrap. The user asked for the same pulsing-brain splash to cover it.

## Root cause

The brain splash `#startup-loading` (a full-cover overlay: `position:fixed; inset:0; bg; z-index:99999`) was removed **mid-mount** inside `initWave` (`app-init.ts`). But the content-reveal gate (`tab-reveal.ts::scheduleRevealLift`) holds the tab content hidden until "settled" (80ms clean frames / 800ms cap), and `initWaveWrap` un-hides the body immediately. So in the gap between brain-removal and gate-lift, the window showed **bare chrome → empty pane → piecemeal mount, uncovered** — the flashing. Most visible on tear-off: a pool window is shown instantly *with* the brain, then the brain died before the torn-off content had rendered.

## Fix — tie the splash to the reveal gate

- **Don't remove the brain in `initWave`** — let it survive past mount.
- **`tab-reveal.ts`:** new `liftGate()` calls `setTabSwitching(false)` **and** `fadeOutStartupSplash()` at every gate-lift site (settled, hard-cap, and the no-`PerformanceObserver` fallbacks).
- **`startup-splash.ts`** (new): idempotent cross-fade + remove (adds `.fading`, removes on `transitionend` with a timeout safety net). No-op once gone, so ordinary tab switches don't re-trigger.
- **`index.html`:** `#startup-loading.fading { opacity: 0 !important; transition: opacity 200ms }`.
- 30s body safety net also drops the splash as a last resort.

Result: **brain → (cross-fade) → settled content**, nothing uncovered in between. Applies uniformly to all window-init paths (main / new / pool), since they all funnel through `initWaveWrap` → `scheduleRevealLift`.

## Verification

- `tsc --noEmit` clean on touched files.
- `tab-reveal.test.ts` + `error-display.test.ts` — 10/10 pass.
- Live: HMR-applied; verify by tearing off a tab — brain should hold over the whole bootstrap, then fade to ready content with no bare-chrome flash.

## Note

Also lands `docs/analysis/ANALYSIS_TEAR_OFF_PERF_2026_06_13.md` — the pane-vs-tab tear-off perf/flash analysis. This PR is the shared **P0** (brain coverage). The pane-specific blank-window flash (P0a: show-after-paint) and the floating-pane pool (P0b) are separate follow-ups.

### DIFF
--- a/.changesets/1781336214-fix-tear-off-keep-the-pulsing-brain-splash-covering-the-whol-9dpf.md
+++ b/.changesets/1781336214-fix-tear-off-keep-the-pulsing-brain-splash-covering-the-whol-9dpf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tear-off): keep the pulsing-brain splash covering the whole window bootstrap until content has settled, then cross-fade — instead of removing it mid-mount, which exposed the bare-chrome/empty/piecemeal-mount flashes (most visible on tear-off)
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #1380, opened 2026-06-13
fix(scripts): make import-agents.sh resilient to bad source DBs

AgentA-asaf:
## Problem

`import-agents.sh` scans every `objects.db` under `~/.agentmux` and copies user agents into a target version. The per-source-DB loop runs under `set -euo pipefail` with several **unguarded** sqlite calls, so the **first error on any one source DB aborts the entire import** — silently dropping every agent in DBs not yet scanned, and exiting non-zero with no explanation.

**Observed live:** importing into a freshly-built 0.44.2 dropped a real agent (`Spixo`) whose data lived in a later-sorted DB. A stale `*.bak` channel (and old pre-agent-schema `versions/0.33.x` DBs) mid-scan tripped the abort. Running with `--name Spixo` "worked" only because the filter skipped the offending DBs before they could abort.

## Root causes fixed

| # | Abort point | Fix |
|---|---|---|
| 1 | `resolved=$(best_session_for_slug ...)` — the function returns non-zero when an agent has **no on-disk session** (a normal case), which `set -e` turned into a full abort | `\|\| resolved=""` at the call site |
| 2 | The definition-copy `INSERT` was unguarded; an old/`*.bak` source schema **missing a referenced column** errored → abort | capture stderr, `WARN` + skip that agent, `continue` |
| 3 | Unreadable / locked / **table-less** source DBs aborted the agents query | probe each DB first; pre-agent-schema DBs (no `db_agent_definitions`) skip **silently**, genuine corruption/locks `WARN` |
| 4 | `sqlite3.exe` emits **CRLF on Windows** → trailing `\r` on the last `char(1)` field (the slug) → silently broke on-disk session lookup | `tr -d '\r'` on the query result |
| 5 | `[ "$imported" -gt 0 ] && echo …` poisoned the function's exit code → a zero-import run exited 1 | converted to an `if` |

Net behavior: a single bad/old/locked source DB can no longer abort the import; problems are surfaced as `WARN` lines instead of a silent early exit.

## Testing

Re-ran the exact unscoped import that previously aborted:

- **Before:** aborted after the first 2 agents, exit 1, `Spixo` dropped, no summary.
- **After:** traverses all DBs (60+ ancient `0.33.x` DBs skipped silently), `exit 0`, prints `Imported: N  Skipped: M`, and `Maksi`/`Mopeo`/`Spixo` import correctly with their sessions wired.

`bash -n` clean.

### DIFF
--- a/.changesets/1781338880-fix-scripts-import-agents-sh-skips-bad-old-source-dbs-instea-2hmt.md
+++ b/.changesets/1781338880-fix-scripts-import-agents-sh-skips-bad-old-source-dbs-instea-2hmt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(scripts): import-agents.sh skips bad/old source DBs instead of aborting; strip CRLF from slug
--- a/scripts/import-agents.sh
+++ b/scripts/import-agents.sh
@@ -214,6 +214,24 @@ import_into() {
 
         local src_win; src_win=$(win_path "$src_db")
 
+        # Resilience: one unreadable / locked / corrupt / table-less source DB
+        # must never abort the whole import. Under `set -euo pipefail` the first
+        # sqlite error would propagate and silently drop every agent in DBs not
+        # yet scanned (observed: a stale `*.bak` channel mid-scan dropped a real
+        # agent whose data lived in a later-sorted DB). Probe the source first.
+        # Old pre-agent-schema versions (no db_agent_definitions table) are
+        # expected and numerous — skip those silently; WARN only on genuine
+        # problems (corruption, locks).
+        local _probe_err
+        if ! _probe_err=$(sqlite3 "$src_win" \
+            "SELECT 1 FROM db_agent_definitions LIMIT 1;" 2>&1 >/dev/null); then
+            case "$_probe_err" in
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Re codex's recurring `dst.` finding (*abort when target definitions table missing*): this is already handled, and I verified it empirically. The `*"dst."*` case precedes the `no such table` skip branch in the `case`, and SQLite **does** emit the attached-schema qualifier for a missing target table:

```
Error: in prepare, no such table: dst.db_agent_definitions   # target missing  → matches *"dst."*  → ABORT
Error: in prepare, no such column: nonexistent_col           # source col gone → matches *no column* → SKIP
```

So a target-side missing table aborts (never reaches the per-agent skip); only source-side `no such column` / `no such table` (no `dst.` prefix) downgrade to a skip. No code change needed for this one. The other P2 (agent SELECT failure vs empty) was fixed in `ff8ab30e`.

a5af:
@codex review

PR #1381, opened 2026-06-13
fix(dev-badge): show DEV only under task dev, not on portable/release builds

a5af:
## What

The status-bar **DEV** badge — and the macOS **"AgentMux DEV"** menu name — showed on **portable and release builds**, not just under `task dev`. (Reported on macOS.)

## Root cause

Build identity was resolved from **`AGENTMUX_RUNTIME_MODE`** (env-first). A running `task dev` AgentMux exports `AGENTMUX_RUNTIME_MODE=dev:<branch>` into its environment, which **every descendant process inherits**. Launch a packaged build from a terminal / agent pane *inside* a dev instance — exactly how builds get launched while developing — and it inherits `dev:main` → self-reports `Dev` → DEV badge on a non-dev build. (Same class as the `AGENTMUX_VITE_PORT=5289` collision: AgentMux leaking its own `AGENTMUX_*` identity to descendants.)

## Fix — self-identify by path (spec Fix A + Fix D)

Build identity is a property of the **binary on disk**, not the launching env. Added one shared helper and routed the three previously-divergent copies through it:

- **`agentmux_common::is_dev_self()`** — path + portable-marker via `current_path_only()`, **ignores** `AGENTMUX_RUNTIME_MODE`.
- `platform.rs::get_is_dev` (the badge) → `is_dev_self()`
- `macos_menu.rs` (the "AgentMux DEV" menu name) → `is_dev_self()` (local copy deleted)
- `app.rs` (dev-only Vite frontend fallback) → `is_dev_self()`

`task dev` (`dist/cef-dev/…`) → `Dev` (badge shown); a packaged `.app` / portable → `Installed` / `Portable` (badge hidden), regardless of any inherited env.

## Tests

- `cargo test -p agentmux-common` — new regression `current_path_only_ignores_leaked_dev_env`: with `AGENTMUX_RUNTIME_MODE=dev:main` exported, `current_path_only` returns `Installed` (not `Dev`); `current()` still honors the explicit override. ✅

## Deferred (documented in the spec, with rationale)

- **Fix B** — write the macOS portable marker so a `.app` classifies as `Portable` (not `Installed`). The `is_portable_marker_present` probe expects the marker at the bundle root **inside the signed `.app`** (`exe_dir` two-up), so it needs a notarized-build test to confirm it doesn't break the codesign seal. Not required for the badge (both `Installed` and `Portable` hide DEV).
- **Fix C** — make the launcher's `data_dir` resolution path-only (the actual `AGENTMUX_*`-leak root cause; fixes the whole instance + the vite-port collision). Touches the launcher's I1–I6 isolation invariants → warrants its own carefully-reviewed PR.

### DIFF
--- a/.changesets/1781339330-fix-dev-badge-show-dev-only-under-task-dev-not-on-portable-r-0prg.md
+++ b/.changesets/1781339330-fix-dev-badge-show-dev-only-under-task-dev-not-on-portable-r-0prg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dev-badge): show DEV only under task dev, not on portable/release builds — self-identify build type by exe path, not the leaked AGENTMUX_RUNTIME_MODE env
--- a/agentmux-cef/src/app.rs
+++ b/agentmux-cef/src/app.rs
@@ -706,16 +706,13 @@ wrap_browser_process_handler! {
             // in dev mode — in release builds, localhost:5173 doesn't exist and
             // would show a raw browser error page.
             let base_url = if base_url.is_empty() {
-                // Use the launcher's mode if it set the env, else
-                // fall back to detecting from the host exe path
-                // (covers standalone `task dev` runs).
-                let mode = agentmux_common::RuntimeMode::from_env().or_else(|| {
-                    std::env::current_exe()
-                        .ok()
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #1382, opened 2026-06-13
fix(build): agents run the app's own bashwrap, not a stale system-PATH copy

AgentX-asaf:
## Root cause

The exit-130 fix (#1368) has been on `main` for a while, yet agents in a fresh `task dev` build kept hitting exit 130. **The code was right; the build never shipped it to the agent.** Full write-up: `docs/retros/RETRO_BASHWRAP_STALE_BUNDLE_2026_06_13.md`.

The agent's `agentmux-bashwrap` hook is a **bare PATH lookup**. It resolved on the system PATH to a **stale leftover portable**:
```
C:\Users\asafe\Downloads\agentmux-…g6fc6d864…-portable\runtime\tools\bin\agentmux-bashwrap.exe   (pre-fix)
```
…because:
- **`task dev` never populated its own `runtime/tools/bin`** → the bundled store was empty in dev.
- **The sidecar *appended* tool dirs** ("system PATH wins", `shell.rs`) → even a present bundled binary couldn't win.

So every agent ran whatever `agentmux-bashwrap` happened to be first on the system PATH.

## Fix

**A — `Taskfile.yml` `dev:serve`** copies the freshly-built `agentmux-bashwrap` into the dev runtime's `tools/bin` (Windows: `$DEV_DIR/runtime/tools/bin`; Unix flat: `$DEV_DIR/tools/bin`), mirroring `scripts/package-portable.sh:121`. The bundled store now exists in dev with the **current** binary.

**B — `shell.rs`** now **prepends** the bundled (app-owned, version-locked) tools dir to the agent PATH and **appends** only the user store. `agentmux-bashwrap` is the app's own streaming hook and must match the running build — it now wins over any stale copy anywhere on PATH. (Bundled `jq`/`rg` winning too is intended: agents get deterministic tool versions. User-managed `~/.agentmux/tools/bin` stays appended so the user's system PATH still wins for those.)

## Why this is the real fix vs. the stopgap

The immediate unblock (overwriting the PATH-winning stale binary) patched a random Downloads build — non-deterministic. This makes the dev build **bundle and prefer its own** version-locked binary, so it can't happen again regardless of what's on the host PATH.

## Verification

- After this, a fresh `task dev` produces `dist/cef-dev/runtime/tools/bin/agentmux-bashwrap.exe` (current build), and the sidecar prepends it → the agent's `agentmux-bashwrap` resolves to it. Confirm via `bashwrap-debug.log`: the `PTY child exited exit_code=N` line (present only in the fixed binary) + `chunks_published > 0`.

## Follow-ups (in the retro, not this PR)
- `agentmux-bashwrap --version` + log the resolved hook path/version at agent spawn (cheap guardrail that would've caught this instantly).
- Get stale portable `tools/bin` dirs off the persistent system PATH.

### DIFF
--- a/.changesets/1781339585-fix-build-agents-always-run-the-app-s-own-bashwrap-not-a-sta-nmpt.md
+++ b/.changesets/1781339585-fix-build-agents-always-run-the-app-s-own-bashwrap-not-a-sta-nmpt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): agents always run the app's own bashwrap, not a stale system-PATH copy — dev build now bundles agentmux-bashwrap into the runtime tools/bin, and the sidecar PREPENDS the bundled (version-locked) tools dir to the agent PATH instead of appending it (was the exit-130 root cause: a stale Downloads-portable bashwrap on the system PATH shadowed the fixed bundled one)
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -463,11 +463,11 @@ tasks:
             # launcher now drives srv + host on macOS dev too, so build it
             # alongside the host. They co-locate in dist/cef-dev/ (flat
             # layout) — see dev:serve's Unix branch.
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1383, opened 2026-06-13
feat(registry): session_id on named-agent record (schema v2) — cross-channel P0.1

AgentA-asaf:
**P0.1 of cross-channel agent persistence** — see `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` (§11 has the code-map-driven plan).

## Why
The named-agent registry record (`NamedAgentRecordV1`) is an *instance projection*: it references a definition by FK and carries no `session_id`. Today the read path joins session/definition data from the **current channel's** SQLite — which won't exist for an agent that lives in another channel. So before the registry can go global, a record must carry enough to `--resume` on its own.

## What
- Schema **v2**: add `session_id: Option<String>` (`#[serde(default)]`); `MAX_SUPPORTED_SCHEMA` 1 → 2.
- **Lazy schema bump** (`NamedAgentRecordV1::min_schema_version()`): a record is stamped v2 **only when `session_id` is populated**, so v1 binaries keep reading session-less records and only session-wired records are hidden from them — strictly better back-compat than always-MAX.
- `registry_mirror` populates `session_id` from the live instance. The per-channel SQLite consolidation (`migrate.rs`) stays session-less (v1) until the **global** migration (P0.3) reads `session_id` from `db_agent_instances`.
- All 7 record constructors updated; added `session_id_drives_min_schema_version` test.

## Risk
Additive and back-compatible. **No read-path or behavior change yet** — this only widens the record. `cargo check -p agentmux-srv` clean; 45 registry tests pass.

## Series
P0.1 (this) → P0.2 global definitions registry → P0.3 re-root instances registry + global migration → P0.4 read path → P1 registry-primary writes → P2 cutover → P3 retire legacy.

### DIFF
--- a/.changesets/1781344105-feat-registry-add-session-id-to-named-agent-record-schema-v2-3uaz.md
+++ b/.changesets/1781344105-feat-registry-add-session-id-to-named-agent-record-schema-v2-3uaz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(registry): add session_id to named-agent record (schema v2, lazy-bump)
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

PR #1384, opened 2026-06-13
feat(registry): global agent-definition store — cross-channel P0.2a

AgentA-asaf:
**P0.2 step 1** of cross-channel agent persistence — `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` (§11.4).

## Why
The instance registry references definitions by FK and joins their names/providers from the **current channel's** SQLite — which won't exist for an agent from another channel. So definitions need a **global** home of their own. This PR adds that store (the foundation); later steps mirror writes to it and read from it.

## What
- **`def_schema.rs`** — `DefinitionRecord` carrying all 25 `AgentDefinition` fields. The `registry` module can't depend on `backend::storage` (that would cycle — `backend::storage` already depends on `registry`), so the fields are mirrored here and the `AgentDefinition ↔ record` conversion lives on the dependent side. Includes a forward-compat (unknown-field) round-trip test.
- **`def_store.rs`** — `DefinitionStore`: file-per-definition CRUD at `~/.agentmux/shared/agents/definitions/<id>.json` with a `retired/` tombstone tree. `upsert` preserves unknown fields, refuses schema downgrades, never clobbers a corrupt/newer file; `retire`/`unretire` give resurrection-proof deletes; `hard_delete`, `list_active`.

## Scope / risk
**Store-only and unwired** (behind the module's existing `#![allow(dead_code)]`) — mirrors exactly how the instance registry's "PR A" staged its store before wiring. No behavior change. `cargo test`: **12 new tests pass**; no change to the in-production instance `Registry`.

The forward-compat merge helpers intentionally parallel `store.rs` to keep this isolated from the live instance registry; a follow-up can extract a shared generic file-store (noted in the module docs).

## Series
P0.1 (merged, #1383) → **P0.2a (this)** → P0.2b write-path mirror → P0.2c read-first + startup wiring + global migration → P0.3 re-root instances registry → P0.4 → P1 → P2 → P3.

### DIFF
--- a/.changesets/1781346106-feat-registry-global-agent-definition-store-cross-channel-p0-3uyy.md
+++ b/.changesets/1781346106-feat-registry-global-agent-definition-store-cross-channel-p0-3uyy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(registry): global agent-definition store (cross-channel P0.2a)
--- a/agentmux-srv/src/registry/def_schema.rs
+++ b/agentmux-srv/src/registry/def_schema.rs
@@ -34,9 +34,37 @@ pub struct DefinitionRecord {
     pub data: DefinitionRecordV1,
 }
 
+/// A content blob (system prompt / mcp / env / soul / startup) attached to
+/// a definition — mirrors a `db_agent_content` row. Carried in the global
+/// record so a cross-channel agent launches with its instructions even
+/// though `db_agent_content` is per-version. (codex P1 on #1384.)
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct DefContentBlob {
+    pub content_type: String,
+    pub content: String,
+}
+
+/// A skill attached to a definition — mirrors a `db_agent_skills` row.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct DefSkillBlob {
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Re codex's P1 (persist content/skills with the definition): the record **now carries them** — `DefinitionRecordV1.content` (`DefContentBlob`) + `.skills` (`DefSkillBlob`) were added in `81ea7077`, so the global payload can hold an agent's full launch config. That satisfies the "persist in the payload" ask at the schema level.

This PR is deliberately **store-only and unwired** (behind the module's `#![allow(dead_code)]`), matching how the instance registry's PR A staged its store: `agent_def_list` still reads SQLite and **nothing reads the global store yet**, so no cross-channel agent is broken (there is no cross-channel read path until P0.2c). The remaining work codex describes is the explicitly-staged next phases:
- **P0.2b** — def write-path mirror reads `db_agent_content`/`db_agent_skills` into the record (population).
- **P0.2c** — `agent_def_list` reads global-first + global migration (the read rewire codex points at).

Because the schema is already complete, those phases need no further schema bump. I'll land P0.2b+c as one end-to-end PR so the full cross-channel read path is reviewable together and codex can green it. Merging this foundation on reagent's approval.

PR #1385, opened 2026-06-13
feat(registry): cross-channel agent definitions — write-mirror + read-first (P0.2b+c)

AgentA-asaf:
**P0.2b+c** of cross-channel agent persistence — `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` (§11.4). Builds on the global DefinitionStore from #1384 (P0.2a).

## What this makes work
A **user agent created in one channel now appears — and launches with its instructions/skills — in every channel.**

## Write mirror (best-effort; SQLite stays authoritative)
- `agent_def_insert`/`update` → `registry_def_upsert`; `agent_def_delete` → `registry_def_retire` (resurrection-proof tombstone).
- `agent_content_set` + `agent_skill_insert`/`update`/`delete` → re-mirror, so the global record carries **current content + skills**.
- **Skips seeded templates** — they're manifest-managed/re-seeded per channel and their hide flag is per-channel UI state; only `is_seeded=0` user agents go global.
- All mirror calls are **conn-drop-safe** (the mirror re-acquires the SQLite lock, so the writers were restructured to drop `conn` first).

## Read-first (global wins for user agents; SQLite fallback)
- `agent_def_list` merges the global user-agent roster over local templates (sorted `updated_at DESC, created_at ASC`).
- `agent_content_get_all` + `agent_skill_list` **fall back to the global record** when local SQLite is empty → a cross-channel agent launches with its configured instructions/skills. **This closes the content/skills P1 codex raised on #1384.**

## Wiring
Channel-independent `resolve_shared_definitions_dir` (via `AGENTMUX_SHARED_DIR`, *not* the channel-scoped `AGENTMUX_DATA_DIR` walk-up) · `Store.def_registry` + `set_def_registry`/`shared_def_registry` · main.rs startup construction · `DefinitionStore::get` · `record_to_agent_definition`.

## Scope / risk
Mirror is best-effort (failures logged, never propagated; SQLite remains the source of truth for the local channel). When the global store can't be resolved, everything falls back to today's SQLite-only behavior. **26 def tests pass**, including a read-first integration test (`Store` + global store → roster + content/skills fallback).

## Deliberately NOT here → immediate follow-up **P0.2d**
The **one-time global migration** backfilling *existing* agents from every channel's per-version `objects.db`. Until that lands, an existing agent goes global on its **next edit**; newly-created agents go global immediately.

## Series
[...]

PR #1386, opened 2026-06-13
feat(registry): backfill existing agents into the global store (cross-channel P0.2d)

AgentA-asaf:
**P0.2d** — the final piece of cross-channel agent persistence. `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` (§11.4).

## What this finishes
P0.2b+c (#1385) made **new/edited** agents go global via the live write-mirror. This one-shot startup migration backfills agents created **before** it shipped — so the existing roster (Maksi/Mopeo/Spixo, …) appears in **every** channel on the next launch, no edit required.

## `migrate_definitions_global_once(home, store)`
- Scans `<home>/channels/*/versions/*/data/db/objects.db`, reading `is_seeded=0` `db_agent_definitions` + their `db_agent_content` / `db_agent_skills`, and upserts each as a `DefinitionRecord` into the global store.
- **Dedup by id**, keeping the highest `updated_at` (most recent edit wins across channels/versions).
- **Respects tombstones** (`upsert` won't resurrect a deleted agent); **skips seeded templates** (manifest-managed, per-channel).
- **Per-DB resilient**: an unreadable / locked / old-schema (missing-column) `objects.db` is skipped with a warning, never aborting the pass.
- **Idempotent** via a `.migrated_definitions` marker, written once after the pass. Read-only on every SQLite.

## Why the marker is written even when DBs are skipped
The live write-mirror backfills any transiently-skipped agent on its next edit, so the marker can't strand a row — and writing it unconditionally avoids looping forever on a permanent old-schema DB.

## Wiring / risk
Runs in the main.rs def-store construction before `set_def_registry`; failures are logged and non-fatal (the live mirror is the backstop). **31 def tests pass**, incl. backfill+content/skills+idempotency and seeded-skip+dedup-by-latest.

## Series
P0.1 (#1383) → P0.2a (#1384) → P0.2b+c (#1385) → **P0.2d (this)** → P0.3 re-root instances registry → P0.4 → P1 → P2 → P3.

### DIFF
--- a/.changesets/1781352983-feat-registry-backfill-existing-agents-into-global-store-cro-ygqf.md
+++ b/.changesets/1781352983-feat-registry-backfill-existing-agents-into-global-store-cro-ygqf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(registry): backfill existing agents into global store (cross-channel P0.2d)
--- a/agentmux-srv/src/main.rs
+++ b/agentmux-srv/src/main.rs
@@ -439,6 +439,24 @@ async fn main() {
     if let Some(def_dir) = registry::resolve_shared_definitions_dir() {
         match registry::DefinitionStore::open(def_dir.clone()) {
             Ok(def_store) => {
+                // P0.2d: one-shot backfill of EXISTING user agents from every
+                // channel's per-version objects.db into the global store, so
+                // agents created before this shipped become cross-channel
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1387, opened 2026-06-13
refactor(registry): decouple instance working-dir base from registry root (cross-channel P0.3a)

AgentA-asaf:
## What

Prep refactor for **P0.3b** (re-rooting the instance registry to the global `~/.agentmux/shared/agents/registry/`). Decouples the base directory that instance `working_directory` values are stored **relative to** from the registry's own file location.

## Why

Today the instance registry sits at `channels/<ch>/agents/registry/`, so its parent — `reg.agents_root()` — *happens to equal* the channel agents dir (`AGENTMUX_AGENTS_DIR`). Two production sites rely on that coincidence:

- **Write** (`registry_mirror.rs`): strips `inst.working_directory` against `reg.agents_root()` to compute the relative `working_dir`.
- **Read** (`listnamedagents` handler): re-joins the relative `working_dir` onto `reg.agents_root()` to reconstruct the absolute path.

Once P0.3b moves the registry out of `channels/<ch>/agents/`, `reg.agents_root()` becomes `shared/agents` — which does **not** contain any live instance. Every mirror write would fail `relative_workdir()` and every read would re-join to a wrong path, **dropping the entire My Agents roster**.

## How

- New `Store::registry_agents_base` field + `set_registry_agents_base()` / `registry_agents_base()`.
- Set once from `AGENTMUX_AGENTS_DIR` in `main.rs` (the current channel's agents dir).
- Both the write mirror and the read handler now use `registry_agents_base()`.
- When unset (in-memory tests, odd envs) the accessor **falls back to the registry parent** — identical to today — so existing mirror tests are unchanged.
- The base is **never read from ambient env inside the Store**, so tests running inside an AgentMux pane don't pick up the host's `AGENTMUX_AGENTS_DIR`.

**Behavior-preserving today** (the explicit base and the fallback are equal in the current layout).

## Tests

- `registry_agents_base_falls_back_to_registry_parent`
- `registry_agents_base_override_decouples_from_registry_parent` — registry rooted at a global path whose parent is **not** the channel agents dir; the explicit base still mirrors the instance with the correct relative `working_dir`.
- All 54 `instance_` tests pass (mirror suite unchanged).

Per `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` §11.4 **P0.3**. Follow-up: **P0.3b** re-roots the registry + generalizes the migration with per-channel anchoring.

### DIFF
--- a/.changesets/1781354523-refactor-registry-decouple-instance-working-dir-base-from-re-2ced.md
+++ b/.changesets/1781354523-refactor-registry-decouple-instance-working-dir-base-from-re-2ced.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(registry): decouple instance working-dir base from registry root (cross-channel P0.3a)
--- a/agentmux-srv/src/backend/storage/registry_mirror.rs
+++ b/agentmux-srv/src/backend/storage/registry_mirror.rs
@@ -129,11 +129,17 @@ impl Store {
         let Some(reg) = self.shared_agent_registry() else {
             return;
         };
-        let Some(agents_root) = reg.agents_root() else {
-            tracing::warn!("registry: agents_root has no parent — skipping mirror");
+        // Anchor the relative working_directory on the CURRENT channel's
+        // agents dir, not the registry's own parent — once P0.3 re-roots the
+        // registry under ~/.agentmux/shared/, its parent no longer coincides
+        // with channels/<ch>/agents/. `registry_agents_base()` returns the
+        // explicit channel dir (AGENTMUX_AGENTS_DIR) and falls back to the
+        // registry parent for tests / pre-re-root layouts.
+        let Some(agents_root) = self.registry_agents_base() else {
+            tracing::warn!("registry: no agents base — skipping mirror");
             return;
         };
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentA-asaf:
Good catch on the dev-mode divergence — confirmed. `AGENTMUX_AGENTS_DIR` (`~/.agentmux/dev/<branch>/agents`) ≠ the pre-re-root channel-local `reg.agents_root()` (`~/.agentmux/agents`), and pre-change the dev mirror silently skipped those rows, so wiring the base here would have leaked branch-local workspaces into the shared dev registry. Fixed in `38be9b64`: P0.3a is now a pure no-op — the base stays unset (fallback → `reg.agents_root()` in every mode). The base is wired from `AGENTMUX_AGENTS_DIR` in P0.3b, atomically with the re-root to the global shared registry, where all modes share it and the anchor is correct per channel/branch; cross-channel workspace reconstruction is P0.4's source-encoding job. 👍

a5af:
@codex review

PR #1388, opened 2026-06-13
chore(diag): log the agent's resolved bashwrap at spawn (stale-binary guardrail)

AgentX-asaf:
Follow-up to the exit-130 / stale-bundle saga (#1368, #1382; `docs/retros/RETRO_BASHWRAP_STALE_BUNDLE_2026_06_13.md`).

The whole saga cost time because nothing told us *which* `agentmux-bashwrap` the agent actually ran. This adds that one line.

At agent spawn (`shell.rs`, where the agent PATH is wired) the sidecar now logs (`target = "agent-tools"`):
- **INFO** with the resolved path when the **bundled, version-locked** binary is used (prepended to PATH).
- **WARN** when the bundled store is missing or lacks the binary, so the agent will fall through to a **system-PATH copy** (the stale-binary trap) — pointing at the retro.

`agentmux-bashwrap --version` already works (clap derives it from the workspace version), so **path + version answers "which binary ran?" in seconds.**

### Follow-up 2 (PATH hygiene) — no action needed
Investigated: the stale Downloads-portable `tools/bin` is **not** in the persistent User or Machine PATH, nor in any shell profile — it's injected into the session at runtime (by the parent app/shell). There's nothing persistent to remove, and #1382's prepend already makes the bundled binary win over it regardless. So the hygiene item is effectively closed by #1382 + this guardrail.

`cargo check -p agentmux-srv` clean.

### DIFF
--- a/.changesets/1781355111-chore-diag-log-which-agentmux-bashwrap-an-agent-will-run-at--kw68.md
+++ b/.changesets/1781355111-chore-diag-log-which-agentmux-bashwrap-an-agent-will-run-at--kw68.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1389, opened 2026-06-13
feat(registry): re-root instance registry to global shared dir + scan all channels (cross-channel P0.3b)

AgentA-asaf:
## What

Completes **P0.3**: the named-**instance** registry moves from the channel-scoped `channels/<ch>/agents/registry/` to the global `~/.agentmux/shared/agents/registry/` — sibling of the P0.2 definition store. An agent named in one channel is now **visible in every channel**. The startup migration is generalized to scan **every channel + dev tree** on the machine, anchoring each row on its own source channel's agents dir.

Builds directly on **#1387** (P0.3a), which pre-decoupled the working-dir relative base into `Store::registry_agents_base`.

### `paths.rs` — re-root
- `resolve_shared_registry_dir()` now resolves via `resolve_global_shared_root()` (the channel-independent root, same as the definition store) instead of walking `AGENTMUX_DATA_DIR` up three levels. Removed the now-dead `resolve_shared_home()`.
- Tests: override → `root/shared/agents/registry`; new test asserts the registry resolves from the shared root and **ignores** a channel-scoped `AGENTMUX_DATA_DIR`; registry + definitions are siblings under `shared/agents/`.

### `main.rs` — home derivation + base wiring
- `home = registry-root.ancestors().nth(3)` (registry → agents → shared → `~/.agentmux`), fed to the generalized migration.
- Wire the working-dir base from `AGENTMUX_AGENTS_DIR` right after `set_registry` — now safe in **all** modes (the registry is global, so `AGENTMUX_AGENTS_DIR` is the correct per-channel/per-dev anchor). This is the wiring P0.3a deferred to dodge the dev-leak (codex P2 on #1387).

### `migrate.rs` — generalized scan, per-source anchoring (**landmine #1**)
- `enumerate_sources()` walks `channels/<ch>/versions/<v>/data/db/objects.db` (anchored on `channels/<ch>/agents`) **plus** the dev tree `dev/<branch>[/<sub>]/data/db/objects.db` (anchored on `<instance_dir>/agents`, located via a bounded recursive search since dev depth varies).
- Each `RowSnapshot` carries its **own** source `agents_root`; `row_to_record` strips `working_directory` against **that** — so rows from other channels aren't dropped as "unmappable".
- Stat renamed `versions_scanned` → `dbs_scanned`. Dedup / forget-intent / pre-v8-skip / marker-defer logic unchanged.

## Landmines (§11.5)
- **#1 per-source anchoring** — handled (each row carries its source agents root). New test `migrate_anchors_each_row_on_its_own_channel`.
- **#2 concurrent multi-process writes** — every running channel's srv now writes the same global registry. Safety rests on the existing atomic-rename upsert + `retired/` tombstone tree; concurrent startup migrations converge (`exists_anywhere` skips written ids; same-id writes are content-identical, last atomic rename wins).
- **#3 marker relocation** — the `.migrated_from_sqlite` marker moves with the registry root, so the migration re-runs once into the global location, idempotent via `exists_anywhere`. Expected/acceptable.

## Tests
- **14 migrate** (incl. `migrate_anchors_each_row_on_its_own_channel`, `migrate_handles_dev_layout`), **5 paths**, **54 instance_** — all pass.

Per `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` §11.4/§11.5.

**Follow-up P0.4**: encode the source base per record so cross-channel rows reconstruct their absolute `working_directory` instead of re-joining under the current channel's agents dir.

### DIFF
--- a/.changesets/1781355693-feat-registry-re-root-instance-registry-to-global-shared-dir-66ja.md
+++ b/.changesets/1781355693-feat-registry-re-root-instance-registry-to-global-shared-dir-66ja.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(registry): re-root instance registry to global shared dir + scan all channels (cross-channel P0.3b)
--- a/agentmux-srv/src/backend/storage/store.rs
+++ b/agentmux-srv/src/backend/storage/store.rs
@@ -51,16 +51,17 @@ pub struct Store {
     /// stops coinciding with the channel agents dir, and using it to strip /
     /// re-join `working_directory` would drop every live instance.
     ///
-    /// `None` for in-memory test stores and — for now — production too: the
-    /// accessor falls back to the registry's parent, which equals the channel
-    /// agents dir in the pre-re-root layout for installed/portable builds, so
-    /// behavior is unchanged. It is wired from `AGENTMUX_AGENTS_DIR` in P0.3b,
-    /// atomically with the re-root to the global shared registry (setting it
-    /// before the re-root would diverge in dev mode, where
-    /// `AGENTMUX_AGENTS_DIR` ≠ the channel-local registry parent). When set,
-    /// it must be passed in explicitly — never read from ambient env inside
-    /// the Store — so tests running inside an AgentMux pane don't pick up the
-    /// host's `AGENTMUX_AGENTS_DIR`. See
+    /// In production it is wired from `AGENTMUX_AGENTS_DIR` in `main.rs`
+    /// (P0.3b), atomically with the re-root to the global shared registry —
+    /// which is why the wiring waited for the re-root: setting it earlier would
+    /// diverge in dev mode, where `AGENTMUX_AGENTS_DIR` ≠ the (then
+    /// channel-local) registry parent. `None` only for in-memory test stores
+    /// and odd envs where the var is unset; the accessor then falls back to the
+    /// registry's parent (which equals the channel agents dir in the pre-re-root
+    /// layout), so existing mirror tests are unchanged. When set, it is passed
+    /// in explicitly — never read from ambient env inside the Store — so tests
+    /// running inside an AgentMux pane don't pick up the host's
+    /// `AGENTMUX_AGENTS_DIR`. See
     /// `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` (P0.3).
     registry_agents_base: Mutex<Option<PathBuf>>,
 }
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentA-asaf:
Both addressed in `d65499d5`:
- **codex P1** — agreed, widening the scan made the `complete` gate dangerous. Decoupled attach from `complete`: `main.rs` now attaches the registry whenever the migration runs, so a corrupt/locked DB in an unrelated channel no longer disables cross-channel My Agents. Readable records are served immediately; the live mirror backfills the current channel regardless; the marker stays deferred on any skipped DB so a future launch retries it (idempotent via `exists_anywhere`). Added a `dbs_skipped` counter for observability.
- **reagent P2** — refreshed the stale `resolve_shared_definitions_dir` doc (the registry is no longer channel-scoped; both are global siblings under `shared/agents/`).

a5af:
@codex review

AgentA-asaf:
Both stale comments fixed in `f08320ae` (doc-only): refreshed the main.rs attach-policy comment to match the codex-P1 fix (attaches whenever the migration runs, serves readable records), and the `registry_agents_base` field doc to reflect that P0.3b now wires it from `AGENTMUX_AGENTS_DIR` in production (None now only for in-memory tests / unset envs). The earlier codex P1 inline on this HEAD is a stale re-anchor of the 13:06:33Z comment — already addressed.

AgentA-asaf:
Fixed in `7fe70f24` (doc-only). I grepped the crate for the whole class (`channel-scoped` / `channel-local` / "instance registry above") — the def-block comment was the only remaining stale one; the other hits correctly describe the old layout being migrated from, the def-store's own fallback, or use historical "then" framing.

AgentA-asaf:
Both fixed in `da409d62`:
- **dev-scan bug (real)** — `collect_dev_sources` now skips instance-internal dirs (`data`, `agents`, `logs`, `cef-cache`, `runtime`, `config`) during descent, so it only traverses branch/sub-hash containers and never walks into an agent workspace to pick up a nested `objects.db`. Added 2 tests (nested-ignored with/without an instance-level DB).
- **stale comment (migrate.rs:179)** — rewrote the `complete=false` note: after the codex-P1 fix it only defers the marker, it no longer detaches the registry.

a5af:
@codex review

AgentA-asaf:
Both fixed in `86601925` — good catches, both genuinely new:
- **unreadable dir → defer marker** — `enumerate_sources` now returns an `incomplete` flag via `read_dir_tracking` (NotFound = fine, any other error = retry); the caller seeds `any_db_failed` from it so an unreadable `channels/` or `versions/` defers the marker instead of silently finalizing it. Test added.
- **dev branch named like an internal dir** — replaced the skip-list with an explicit 2-level dev scan (`dev/<branch>/data/...` and `dev/<branch>/<sub>/data/...`), so any branch slug is scanned while the bounded depth still never reaches a nested agent-workspace `objects.db`. Test added. The three earlier dev tests (incl. nested-workspace-ignored) still pass.

a5af:
@codex review

PR #1390, opened 2026-06-13
feat(registry): per-record source agents base for cross-channel workdir reconstruction (P0.4)

AgentA-asaf:
## What

Completes the **P0** read path. After P0.3 the instance registry is global, so a record surfaced in a **different** channel than it was created in re-joined its relative `working_dir` under the **reader's** current channel agents dir — pointing at the wrong workspace (and the wrong cwd for `--resume`).

This adds the missing per-record provenance so cross-channel rows reconstruct their **real** absolute `working_directory`.

## How

New schema **v3** field `NamedAgentRecordV1.source_agents_base: Option<String>` — the absolute agents dir `working_dir` is relative to (the channel/dev instance the agent actually lives in).

- Additive: `#[serde(default)]`; `MAX_SUPPORTED_SCHEMA` 2→3, `MIN` stays 1.
- `min_schema_version()` returns 3 when set, so a pre-v3 reader **skips** such a record rather than mis-resolving it (same discipline as `session_id`→v2). Legacy v1/v2 records have `None` and keep the current-channel fallback — which is correct, since they could only ever represent same-channel rows.

Producers/consumer:
- **migrate.rs** `row_to_record`: `source_agents_base` = the row's **own** source channel agents dir (`row.agents_root`, the per-source anchor from P0.3b).
- **registry_mirror.rs** (live mirror): `source_agents_base` = the **current** channel agents base it strips `working_dir` against — live-mirrored rows live in this channel.
- **agent_handlers.rs** `listnamedagents`: reconstruct the absolute `working_directory` against `source_agents_base` when present, else fall back to the current channel base (pre-P0.4 behavior, correct for legacy records).

## Tests
- schema: v3 drives `min_schema_version`; unknown-future-field round-trip (additive contract).
- migrate: anchoring test asserts each record's `source_agents_base` is its source channel + `schema_version == 3`.
- live mirror: asserts the current channel base is stamped.
- **18 migrate + 67 registry + 55 instance_** tests pass.

Per `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` §11.4 **P0.4**. This closes **P0** — named agents are now visible **and** correctly resolvable across every channel. Follow-up: **P1** (registry-primary writes + tombstones) → **P2** (cutover) → **P3** (retire per-version tables).

### DIFF
--- a/.changesets/1781359077-feat-registry-per-record-source-agents-base-for-cross-channe-a20n.md
+++ b/.changesets/1781359077-feat-registry-per-record-source-agents-base-for-cross-channe-a20n.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(registry): per-record source agents base for cross-channel workdir reconstruction (P0.4)
--- a/agentmux-srv/src/backend/storage/registry_mirror.rs
+++ b/agentmux-srv/src/backend/storage/registry_mirror.rs
@@ -59,8 +59,12 @@ fn agent_instance_to_record(
         created_by_version: version.clone(),
         last_launched_by_version: version,
     };
-    // Lazy schema bump: v2 only when session_id is set, so session-less
-    // records stay readable by v1 binaries (see schema::min_schema_version).
+    // Stamp the lowest envelope schema that faithfully represents the payload
+    // (schema::min_schema_version). Since P0.4 always sets source_agents_base,
+    // that is v3 for every live-mirrored record — a pre-v3 reader rejects it
+    // (intended: better to hide than mis-resolve a cross-channel workdir; the
+    // only pre-v3 readers of the GLOBAL registry are pre-P0.4 builds, which
+    // ship together with this).
     Ok(NamedAgentRecord {
         schema_version: data.min_schema_version(),
         data,
--- a/agentmux-srv/src/backend/storage/registry_mirror.rs
+++ b/agentmux-srv/src/backend/storage/registry_mirror.rs
@@ -49,6 +49,11 @@ fn agent_instance_to_record(
         // record can `--resume` without joining the current channel's SQLite.
         session_id: empty_to_none(&inst.session_id),
         working_dir: rel,
+        // v3: the agents dir `working_dir` is relative to — this row lives in
+        // the CURRENT channel, so its source base is the channel agents dir we
+        // just stripped against. Lets a different channel reconstruct the
+        // absolute path instead of re-joining under its own agents dir (P0.4).
+        source_agents_base: Some(agents_root.to_string_lossy().to_string()),
         created_at_ms: inst.created_at,
         last_launched_at_ms: inst.started_at,
         created_by_version: version.clone(),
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

AgentA-asaf:
Both fixed in `7579120b` (doc-only). Correct: since `source_agents_base` is always set, every P0.4 record is v3, and a pre-v3 reader rejects it via the `schema_version > MAX` upper-bound (not MIN). That's intended — hiding is safer than mis-resolving a cross-channel workdir, and the only pre-v3 readers of the global registry are pre-P0.4 builds (which ship with this). I swept the rest of the class (agent_handlers, schema.rs) — those correctly describe *legacy* v1/v2 records and are accurate.

AgentA-asaf:
Good catch, and agreed it's a real gap — `.migrated_from_sqlite` short-circuits, so a registry already migrated by a P0.3 build that predates this field keeps `source_agents_base: None` and the handler falls back to the wrong cross-channel workspace.

Plan: I'll land this as a dedicated **P0.4b backfill** PR immediately after this merges — same pattern as P0.2d (separate marker, idempotent one-shot). It must update **only** `source_agents_base` on existing records (re-deriving each row's source channel from the SQLite sources), preserving live-mirror-enriched fields like `session_id` — a blind re-upsert of the SQLite-derived record would clobber those, so it needs its own careful pass + review rather than riding in this PR.

Practical note: P0.3b merged minutes ago with no release cut, and this field ships in the same release train as P0.3, so there is currently **no** registry on any machine with the marker-but-no-source_agents_base state — P0.4b lands before any release consumes these changesets, so the real-world window is zero. 👍

AgentA-asaf:
Addressed in `4963118f` — the backfill now ships **in this PR** (both of you flagged it; reagent as blocking P1). `backfill_source_bases_once` runs once under its own `.backfilled_source_bases` marker (P0.2d pattern), re-derives each pre-v3 record's source channel from the SQLite sources, and sets **only** `source_agents_base` on the existing record — re-stamping v3 while preserving `session_id`/`identity`/timestamps (no blind re-upsert that would clobber live-mirror fields). Records whose source channel is gone are counted `unresolved` and left None (handler falls back; relaunch re-mirrors). Wired in main.rs after the migration. 3 tests incl. session-id-preserved + idempotent.

a5af:
@codex review

AgentA-asaf:
Both fixed in `fe248190`:
- **backfill dedup (real)** — `backfill_source_bases_once` now collects the winning **latest-`started_at`** row per id across all sources before anchoring, exactly like `migrate_from_sqlite_once` derives `working_dir`, so source_agents_base anchors on the same channel `working_dir` was stripped against. Added a test (same id in two channels → anchors on the latest-started_at one).
- **stale comment (main.rs:467)** — corrected; the current-channel base is the fallback for legacy v1/v2 records only, v3 carries its own `source_agents_base`.

a5af:
@codex review

PR #1391, opened 2026-06-13
fix(registry): cross-channel agent backfill captures all existing agents

AgentX-asaf:
## Problem

The cross-channel feature (#1383–#1390) is supposed to retain agents across channels+versions, but on this machine the global store (`~/.agentmux/shared/agents/definitions/`) held only **1 of ~6** custom agents. A fresh channel/version (e.g. a new `task package` portable) showed none of the user's agents.

Full root-cause: `docs/analysis/ANALYSIS_CROSS_CHANNEL_AGENT_RETENTION_2026_06_13.md`. The read/merge path is correct; the **one-shot backfill migration** (`registry/def_migrate.rs`) was broken three ways. Log evidence: `dbs_scanned=12 dbs_skipped=10 rows_seen=1 records_written=1` with `error: "no such column: container_image … FROM db_agent_definitions"`.

## Fixes

- **F1 — schema-resilient query.** It `SELECT`ed `container_image`/`_volumes`/`_name`; older DBs lack those columns → `no such column` → the **whole DB skipped**. Now `PRAGMA table_info` introspects the present columns and substitutes a default literal for any absent (order/count fixed, so the index-based row mapper stays valid); the `is_seeded` filter is dropped when that column is absent.
- **F2 — scan `dev/`.** It walked only `channels/*/versions/*`. Agents in `dev/<branch>[/<sub>]` (e.g. **Qooma** in `dev/main`) were invisible. New `collect_scan_dbs()` scans channels **and** dev, matching the instance migration.
- **F3/F4 — recoverable marker.** It wrote `.migrated_definitions` unconditionally after the incomplete pass and never re-ran. Replaced with a **versioned marker** (`MIGRATION_VERSION = 2`): a marker older than the current version — including the legacy `migrated` text (→ 0) — re-runs once, so **existing users recover automatically** on next launch.

## Tests

`cargo test -p agentmux-srv def_migrate` — **7 pass**: schema-resilient read of a container-less DB, dev-branch scan, legacy-marker re-run, plus the existing idempotency/dedup/freshness tests.

## Verification plan

After this, on next launch the (re-running) migration backfills every readable channel+dev DB. I'll confirm on this machine that Qooma / Clamk / Naki / CodexPo / GeminiOpp / Mazs all land in the global store and appear in a fresh channel.


a5af:
@codex review

PR #1393, opened 2026-06-13
fix(registry): repopulate "My Agents" cross-channel — anchor instance migration on the global agents root

AgentX-asaf:
## Problem
"My Agents" (the agent **sessions/instances** list) is empty in every channel even though the agents exist on disk. The cross-channel registry migration silently dropped all of them.

`migrate.rs::row_to_record` stripped each instance's `working_directory` against the **per-channel** `channels/<ch>/agents` dir. But agent workspaces live at the **global** `~/.agentmux/agents/<name>` (verified on disk: `working_directory = ~/.agentmux/agents/mazs-0527n`). So `strip_prefix` failed for every row → all "unmappable" → 0 written.

On the affected machine: `dbs_scanned=52, rows_seen=10, records_skipped_unmappable=9, records_written=0`.

## Fix
1. **`row_to_record` anchors on the real workspace root `<home>/agents`** (`home.join("agents")`; `home` = `~/.agentmux`, which main.rs derives as the registry root's 3rd ancestor: `registry → agents → shared → home`). It falls back to the row's own source-channel agents dir for any legacy in-channel workspace, and skips a workspace under neither root (matching the live mirror's `relative_workdir`). The reader (`agent_handlers.rs`) reconstructs `source_agents_base.join(working_dir)`, so storing `source_agents_base = <home>/agents` + `working_dir = <name>` resolves the real workspace in any channel.
2. **Recoverable marker** (same pattern as #1391's definitions migration): `MIGRATION_VERSION = 2` + a `migration_version:` line. A registry an older build already finalized has a legacy stats-only marker (no version line → read as `0`), so the migration **re-runs exactly once**; `exists_anywhere()` prevents duplicates.

### Note on the first push
The initial commit anchored on `registry.root().parent()` and claimed parity with the live mirror — **wrong** on both counts (reagent caught it). The re-rooted registry sits at `<home>/shared/agents/registry`, so its parent is `<home>/shared/agents`, **not** the workspace root `<home>/agents`; and the live mirror anchors on the per-channel `AGENTMUX_AGENTS_DIR`. Corrected in the follow-up commit; the test now asserts the anchor ≠ `registry.root().parent()` so it can't regress.

## Tests
`migrate_anchors_global_workspace_not_per_channel` (production shape + anti-regression guard), `migrate_legacy_marker_reruns_then_settles`; existing per-channel-anchoring and skip-unmappable tests stay green. **24/24 pass.**

Complements #1391 (definitions). Context: `docs/architecture/ARCHITECTURE_AGENT_DATA_AND_CROSS_CHANNEL_2026_06_13.md` §4–§5 (I1).

### DIFF
--- a/.changesets/1781378511-fix-instances-my-agents-cross-channel.md
+++ b/.changesets/1781378511-fix-instances-my-agents-cross-channel.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(registry): anchor instance migration on the global agents root so "My Agents" repopulates cross-channel (instances; complements #1391's definitions fix)
--- a/agentmux-srv/src/registry/migrate.rs
+++ b/agentmux-srv/src/registry/migrate.rs
@@ -15,13 +15,18 @@
 //!   <home>/dev/<branch>[/<sub>]/data/db/objects.db         (dev)
 //! ```
 //!
-//! **Landmine #1 — per-source anchoring.** A row's `working_directory` is
-//! absolute under *its own* channel's agents dir (`channels/<ch>/agents`, or
-//! `<instance_dir>/agents` for dev). Stripping every row against a single
-//! global agents root would mark rows from other channels "unmappable", so
-//! each source DB carries the agents dir its rows are anchored on, and
-//! [`row_to_record`] strips against that. See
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1395, opened 2026-06-14
fix(launcher): one AgentMux instance per local build (per-build channel + nested env-leak guard)

AgentX-asaf:
## Problem
A freshly-built `task package` portable doesn't run its own binary:
- **Standalone, same branch:** a second build joins the first build's running window — `CEF early exit … exit_code=24` / *"Opening in existing browser session"*.
- **Nested inside another AgentMux** (e.g. an agent terminal): the build adopts the **parent** pane's channel and forwards into the parent.

## Root cause — three isolation keys disagreed
| Key | Was keyed on |
|---|---|
| Single-instance pipe | `(data_dir, build label)` → per-**build** (#1315) |
| Data dir | `(channel, semver)` → per-**branch** |
| CEF user-data-dir singleton | cef-cache = `(channel, semver)` → per-**branch** |

**#1315** made the *pipe* per-build but left the data dir + cef-cache per-branch. cef-cache is derived from the channel (`data_paths.rs:209`), so Chromium's singleton re-coupled what the pipe split — which is why #1315's retro workaround was literally *"close all instances first."* Separately, portables honor an inherited `AGENTMUX_CHANNEL` (dev builds don't), so a nested launch adopts the parent's channel via the pane's leaked env.

## Why this is fixable *now*
#1315 kept the data dir per-branch to preserve session persistence (agents/auth/panes). The **cross-channel work (#1387–#1393)** moved agents **and auth** to global stores, so a per-build data dir keeps every agent and stays logged in. The blocker is gone.

## Fix
1. **`package.sh`** — bake a per-build `BUILD_ID` (8-char hash of the full label) into `AGENTMUX_BUILD_CHANNEL_DEFAULT` → channel `local-<slug>-<hash>-<build-id>`. Data dir + cef-cache + pipe are now **all per-build**. Releases unaffected (`RELEASE_CHANNEL` wins; ≤55/64 chars). `--fresh` becomes a no-op.
2. **`data_dir.rs`** — when launched **nested** (`AGENTMUX` set), resolve path-only and ignore the leaked ambient `AGENTMUX_CHANNEL` — symmetric with dev builds. Launcher env is already authoritative for host+srv (`to_env_vars()` overwrites inherited), so no downstream change. Explicit *standalone* override still honored (PR #1027).
3. **GC** in `package.sh` — keep newest `AGENTMUX_LOCAL_CHANNELS_KEEP` (default 5) per-build channels per branch, prune older (skip <30 min old) so disk stays bounded.

## Safety (verified, code-anchored)
- **Invariants:** strengthens **I1/I4/I6**, neutral to I2/I3/I5. Two builds of one branch run simultaneously safely (separate SQLite, pipe, cef-cache, unnamed jobs).
- **What survives per-build:** agents (definitions, My-Agents, workspaces, transcripts, `--resume`) ✅ and **auth/oauth** ✅ (global). **What starts fresh:** pane/tab layout and **memories** (`db_memory_bundles` — not yet globalized; flagged as a follow-up).

## Tests
- Extended `resolve_path_only_ignores_env_channel` to assert **Portable** mode ignores the leaked channel (the nested guarantee).
- Added `per_build_channels_isolate_data_dir_even_at_same_version` (data-dir axis).
- **188 launcher + 22 data_paths tests pass.**

Retro: `docs/retro/retro-per-build-launch-isolation-2026-06-13.md`. Completes the chain from `retro-local-build-isolation-regression-2026-06-09.md` (#1315).


a5af:
@codex review

PR #1396, opened 2026-06-14
fix(agents): surface cross-channel agents in "My Agents"

AgentX-asaf:
## Problem
"My Agents" showed only the current instance's local sessions, never your agents from other builds/channels — even though the cross-channel registry exists. Two root causes:

1. **The live mirror was dropping every new agent.** `registry_mirror.rs` anchored `working_directory` on the per-channel `AGENTMUX_AGENTS_DIR`, but workspaces are **global** (`~/.agentmux/agents/<name>`), so `strip_prefix` failed and each new agent was skipped (`instance not representable as record`). This is the **live-write twin of the bug #1393 fixed in the one-shot migration** — so the registry only ever got agents from that one migration, and nothing new (e.g. a just-created "Mzap") propagated.
2. **"My Agents" never read the registry.** The grid's `listrecentsessions` handler read only local SQLite.

## Fix
1. **`registry_mirror.rs`** — anchor **global-first** (per-channel fallback), deriving `<home>/agents` from the registry root exactly as `migrate.rs`/`main.rs` do. New agents now register.
2. **`listrecentsessions` ("My Agents")** — source from the global registry when available: `list_active` → dedup by `(definition_id, instance_name)` keeping newest → reconstruct workdir → **overlay** live local rows (running `block_id`/`session_id`) and **append** local-only agents. Cross-channel rows become synthetic instances; the existing snapshot/preview loop + `RecentSessionRow` shape are untouched, so **zero frontend change**. SQLite path kept as the no-registry fallback.

## Why this is safe / how surfaces interact
- Agent definitions + auth are global (#1387–#1393), so cross-channel rows resolve their definition + stay logged in; reattach uses the reconstructed `working_directory` + `continueOfInstanceId` (existing flow).
- Dedup by `(definition_id, instance_name)` closes the gap `registry_mirror.rs:133` called out (the registry read path lacks SQLite's chain-root collapse).

## Tests
`mirror_anchors_global_workspace`, `mirror_per_channel_legacy_workspace_still_works`, `mirror_skips_workspace_under_neither_root` (store.rs). 91 store + 24 migrate tests pass; full srv suite green.

Completes the cross-channel chain: #1391 (definitions migration) → #1393 (instances migration) → this (live mirror + My-Agents read).

### DIFF
--- a/.changesets/1781405424-my-agents-global-registry.md
+++ b/.changesets/1781405424-my-agents-global-registry.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agents): surface cross-channel agents in "My Agents". Fix the live registry mirror to anchor on the GLOBAL workspace root (the live-write twin of #1393 — newly-created agents were silently dropped as "not representable"), and source the My-Agents list from the global registry (deduped by definition+name, enriched with local running state, local-only agents appended) so agents created in any build/channel/version appear.
--- a/agentmux-srv/src/backend/storage/registry_mirror.rs
+++ b/agentmux-srv/src/backend/storage/registry_mirror.rs
@@ -28,16 +28,33 @@ use super::store::{AgentInstance, Store};
 /// cross-version dropdown.
 fn agent_instance_to_record(
     inst: &AgentInstance,
-    agents_root: &Path,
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1399, opened 2026-06-14
feat(agent): globalize agent transcript for cross-channel history

AgentX-asaf:
## Problem

After #1396, agents from other builds/channels appear in **My Agents**, but opening one shows an **empty pane** — the conversation doesn't load. Root cause: a conversation is stored **per-channel** (`filestore.db`), so when you open the agent from a different build there's nothing local to load. Conversation history is the last per-channel agent surface (definitions, instances, workspaces, and auth went global in #1387–#1396).

## Fix

Back the `agent:<defId>:current` zone with a **global** transcript store, and read from it on open. Storage + addressing only — no new render path or history format (the O(1) paginated loader #336–#342 is reused verbatim).

| Phase | Change |
|---|---|
| **1** | `resolve_shared_transcripts_dir()` → `<shared>/agents/transcripts`; second `FileStore` in `main.rs` as `Option<Arc<FileStore>>`; `AppState.global_transcript_store` + a process-global handle so the controller hot path reaches it without threading. |
| **2** | Hot-path mirror: the agent `output` stream is mirrored into the global zone (subprocess/persistent/acp readers). Zone resolved once from block meta `agentId`. Fire-and-forget; PTY `term` never mirrors. |
| **3** | Read fallback in **`blockfile:read_range` and `blockfile:line_count`** — when this channel has no local `output` for the block, read the global zone. Zero frontend change. |
| **4** | Snapshot overlay mirrored to + read-fallback from the global store. Additive — per-channel stays primary, same-channel behaviour byte-identical. |

## The #1361 interaction (important)

The driving analysis predates **#1361**, which made the snapshot a *lightweight overlay* (no `nodes[]`; just `highWaterMark`). The pane now reconstructs history from the `output` NDJSON via `read_range`. So:

- The snapshot-only "smallest slice" no longer restores anything — the load-bearing core is **1 → 2 → 3**.
- A cross-channel open is a **fresh local block** whose id never equals the v2 snapshot's `sourceBlockId`, so the frontend falls through to the **line-count path** (gated on `count === 0` → empty pane). That's why **`line_count` needed the global fallback too**, not just `read_range`.

## Deferred

Backfill of the **9 pre-existing migrated agents** + §6 `session_id` recovery → verified follow-up. Go-forward cross-channel history works without it (the Phase-2 live mirror writes the exact correct format). The backfill *source* (provider `.jsonl` transcript) is **not** byte-identical to the stream-json the renderer expects, and `ClaudeHistoryAdapter` yields a lossy summary — so it needs app-level verification, not a write-blind migration.

## Tests

9 new unit tests (zone resolution, hot-path mirror round-trip, read/line_count fallback, snapshot read fallback). Full `agentmux-srv` suite green: **1128 passed, 0 failed**.

Design doc: `docs/analysis/ANALYSIS_CROSS_CHANNEL_CONVERSATION_HISTORY_2026_06_14.md` (§9 = implementation status).


a5af:
@codex review

PR #1357, opened 2026-06-12
feat(container): Phase 2 — ContainerManager + docker exec spawn

AgentY-asaf:
## Summary

- **ContainerManager** (`agentmux-srv/src/backend/container.rs`, bollard v0.18): persistent container model — one container per agent, `docker exec` per turn
  - `ensure_running()`: always queries Docker (no stale cache), per-container serialization mutex to prevent TOCTOU 409 race on create
  - `parse_volume_spec()`: Windows drive-letter–aware volume string parsing
  - `container_name_for_slug(slug)` → `agentmux-<slug>`
- **AppState**: `container_manager: Option<Arc<ContainerManager>>` — `None` when Docker unavailable; async ping on startup logs availability
- **websocket.rs `agentinput`**: branches on `agentMode == "container"`, calls `ensure_running`, then spawns `docker exec -i [-w <cwd>] <container> <cli> [args]` instead of the native CLI
- **app_api.rs `agent.send`**: same container branch for the `SubprocessController` path (picker reattach + REST-path turns)
- **app_api.rs `agent.open`**: writes `agent:container_image` and `agent:container_volumes` to block meta so send/input handlers can read them at turn time

## Design notes

- `docker exec -i` (NOT `-t`) — avoids CR/LF corruption of NDJSON output from the agent CLI
- Container image: `ghcr.io/agentmuxai/agent-claude:latest` (default); override via `agent:container_image` block meta
- Named volume for `~/.claude` keeps credentials inside the container, not on the host
- `ContainerManager::connect()` is infallible for local Docker; ping is async so startup is never blocked by a slow daemon

## Test plan

- [ ] `cargo check -p agentmux-srv` passes (verified: 0 errors)
- [ ] Open a container-type agent pane → `agentMode` = `"container"` in block meta
- [ ] Turn triggers `docker exec -i agentmux-<name> claude ...` (verify with `docker ps` + logs)
- [ ] Non-container agents unaffected (host path taken when `agentMode != "container"`)
- [ ] Docker unavailable → error surfaced to client, not a panic

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1401, opened 2026-06-14
fix(container): pass global_output_zone through container-exec path (unbreak main)

AgentA-asaf:
## Problem

`main` (`3e866dc0`) does not compile:

```
error[E0061]: this function takes 6 arguments but 5 arguments were supplied
  --> agentmux-srv/src/backend/blockcontroller/subprocess.rs:1309:13
  argument #6 of type `Option<&str>` is missing
note: function defined here
  --> agentmux-srv/src/backend/blockcontroller/shell.rs:1220:8
```

## Root cause — semantic merge conflict (both PRs green individually)

- **#1399** (globalize agent transcript) added a 6th param `global_output_zone: Option<&str>` to `handle_append_block_file` and updated the **subprocess / persistent / acp** stdout readers to resolve the zone once and pass it.
- **#1357** (container Phase 2) merged in parallel and added a *new* call site — `SubprocessController::publish_line`, the container-exec output reader — built against the **old 5-arg** signature.

Neither PR's CI compiled the *merged* tree, so the break only surfaces on `main`.

## Fix

Resolve the agent's global transcript zone **once** before the container-exec reader loop (via the existing `resolve_global_output_zone`), thread it through `publish_line`, and pass it to `handle_append_block_file`. This is byte-for-byte the same pattern the other three controllers already use — in particular it does **not** add a per-line store lookup on the hot output path.

## Verification

- `task build:backend` → `Finished release profile [optimized] in 4m 06s` (only pre-existing warnings; the E0061 error is gone).

This is the same fix already carried on the `agenta/release-v0.45.0` branch; landing it standalone unblocks `main` for everyone independent of the release.

### DIFF
--- a/.changesets/1781451735-fix-container-global-output-zone.md
+++ b/.changesets/1781451735-fix-container-global-output-zone.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(container): thread global_output_zone through the container-exec output path (publish_line) so main compiles — a semantic merge conflict between #1399 (added the 6th arg to handle_append_block_file) and #1357 (added the publish_line call site with the old signature).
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1403, opened 2026-06-14
feat(agent): backfill pre-existing agent conversations into the global transcript store

AgentX-asaf:
## Problem

#1399 made agent conversations cross-channel by mirroring **new** agent output into the global `agent:<defId>:current` zone. But conversations that existed **before** #1399 shipped were never mirrored — so the 9 cross-channel agents (Smark, Smike, …) opened **empty**: with no global history and `session_id: null`, opening them just started a fresh CLI turn ("No prior memory for this session").

## Fix

A one-shot, marker-gated backfill at srv startup. For each user agent definition it scans every channel/dev `objects.db` + `filestore.db`, finds the **richest block `output`** for that agent, and seeds the global zone with it.

- **Source = the channel block's `output`** — the exact stdout stream-json the renderer already parses (`parseHistoryLines` + `claude-stream-json`). **Not** the provider `~/.claude/projects/<slug>/<sid>.jsonl` transcript, which is a different envelope (`queue-operation` / `user`+`parentUuid`) the renderer can't consume. Zero translation.
- Seeded snapshot overlay uses **`sourceBlockId: ""`**, which the frontend treats as "matches any block" — so opening the agent (a fresh local block with no local `output`) takes the v2 fast-path and #1399's `read_range`/`line_count` global fallback serves the backfilled zone.
- Marker-gated + version-stamped (`.transcripts_backfilled`), read-only on every scanned SQLite, mirrors `registry::def_migrate`. Won't clobber a global zone that already has equal/more content.

## Gotcha found

`db_block.data` is declared `TEXT` but the wstore stores the JSON as a **BLOB** (`typeof = 'blob'`). Reading it as a Rust `String` errors and silently skips every row → `seeded=0`. Fixed by reading raw bytes (`from_slice` parses either).

## Verified live (`task dev`)

`transcript backfill finished agents=12 seeded=10 skipped_no_source=2`. Smark's global zone went **20 KB → 124 KB** and the conversation **renders on open** (confirmed by the owner). The 2 skipped agents have no recoverable block output in any channel.

## Tests

3 unit tests: richest-channel-wins seeding, don't-clobber-richer-global, and BLOB-storage round-trip. Full `agentmux-srv` suite green.

Companion to #1399. Does **not** address `--resume` (continuing the conversation needs `session_id` recovery — a separate follow-up); this is display/load only.

### DIFF
--- a/.changesets/1781453182-feat-agent-backfill-pre-existing-agent-conversations-into-th-jjpi.md
+++ b/.changesets/1781453182-feat-agent-backfill-pre-existing-agent-conversations-into-th-jjpi.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): backfill pre-existing agent conversations into the global transcript store
--- a/agentmux-srv/src/backend/mod.rs
+++ b/agentmux-srv/src/backend/mod.rs
@@ -26,6 +26,7 @@ pub mod schema;
 pub mod service;
 pub mod session_archive;
 pub mod shellexec;
+pub mod transcript_backfill;
 pub mod shellintegration;
 pub mod sysinfo;
 pub mod storage;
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1406, opened 2026-06-14
fix(build): slim the Linux release — AppImage strips + official-build libcef

AgentU-asaf:
Two-part Linux release size reduction.

## 1. AppImage packaging (`scripts/build-appimage-linux.sh`)
- **Strip frontend source maps** (~28 MB) — debug-only, matches the Windows/macOS `STRIP_MAPS` policy.
- **Max-level zstd** (`-Xcompression-level 22`). (xz+BCJ would be smaller, but the AppImage type2-runtime's squashfuse only supports zlib/zstd — an xz image won't mount.)

## 2. The real lever: libcef was a non-official build (`docs/cef-build/`, `specs/`)
The patched fork `libcef.so` was built **without `is_official_build`** — a plain `is_debug=false` Release — so it ran **~2× stock-official**. Corrects the build doc's gn args to set `is_official_build=true` + `use_thin_lto=true` (CFI/PGO off to skip their deps), and fixes `specs/cef-size-reduction.md`, whose "libcef is already official" premise was wrong for our build.

## Verified (CEF 148.0.7778.180)
- `libcef.so` **414 → 263 MB** stripped (below stock-official ~280 MB); `dist/cef` 709 → 313 MB.
- **AppImage 197.7 → 134.6 MB** — now smaller than Windows (168) and macOS (139).
- Launches clean (no V8 snapshot mismatch), hardware WebGL via the cap-probe (GaneshGL), terminal on the WebGL renderer.

## Notes / gotchas documented
- Regenerate the gitignored `ctocpp` wrappers (`translator.py`) before building, or it dies on a missing `window_ctocpp.cc`.
- Copy the new V8 snapshots alongside the rebuilt libcef (stale ones crash on checksum mismatch).
- **Distribution follow-up (separate):** the slim `libcef.so` is local; CI/release must fetch the official-built runtime for releases to ship the 135 MB AppImage.

### DIFF
--- a/.changesets/1781456067-fix-build-slim-the-linux-appimage-strip-frontend-source-maps-2mlx.md
+++ b/.changesets/1781456067-fix-build-slim-the-linux-appimage-strip-frontend-source-maps-2mlx.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): slim the Linux AppImage — strip frontend source maps (~28MB) and use max-level zstd compression
--- a/docs/cef-build/build-patched-libcef.md
+++ b/docs/cef-build/build-patched-libcef.md
@@ -91,6 +91,10 @@ If you see "class member cannot be redeclared" compile errors later, the patcher
 cd ~/cef-build/chromium_git/chromium/src
 gn gen out/Release_GN_x64 --args='
   is_debug=false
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #743, opened 2026-05-08
build(linux): patched-libcef.so resolver for bundle:linux

AgentU-asaf:
## Summary

\`task bundle:linux\` and \`task package:linux\` previously sourced \`libcef.so\` directly from cef-dll-sys's cargo cache, which holds the *upstream* prebuilt. AgentMux's Linux runtime needs a patched libcef built from the \`agentmux/7680-drag-rightclick-and-transparency\` branch — it adds \`CefWindow::BeginWindowDrag()\`, called via raw FFI by \`agentmux-cef/src/ui_tasks.rs::StartWindowDragTask\` for left-click window drag on Wayland (PR #663). Every clean bundle silently regressed to upstream → drag IPC silently no-opped → required a manual overlay step after every build.

## Approach

3-tier resolver invoked from \`bundle:linux\`:

1. \`\$AGENTMUX_CEF_RUNTIME_DIR\` — explicit override
2. \`~/cef-build/chromium_git/chromium/src/out/Release_GN_x64\` — standard cef-build layout
3. cef-dll-sys cargo cache — current behavior, fallback for non-maintainers

If the chosen libcef.so is >1 GB the resolver warns it's likely the unstripped upstream debug build (patched stripped build is ~613 MB), so the failure mode is debuggable at build time instead of "left-click drag broke and I don't know why".

The runtime ABI guard (\`_cef_window_t.size\` check in \`creation_views.rs\`-style code) remains the contract — this resolver is the build-time soft signal.

Three rejected alternatives are documented in the spec:
- In-tree vendor (~600 MB binary in git, infeasible)
- Fork cef-dll-sys (third place to track patches)
- Cargo xtask (overkill for one binary on one platform)

## Files

**New:**
- \`scripts/resolve-cef-runtime.sh\` — POSIX shell helper (~80 lines including comments)
- \`docs/cef-build/build-patched-libcef.md\` — how-to-build doc consolidating the cef-build machine notes (OOM-resistant ninja wrapper via \`systemd-run --user --scope\`, patcher.py idempotency gotcha, etc.)
- \`docs/specs/patched-libcef-bundling-2026-05-08.md\` — full design spec

**Edited:**
- \`Taskfile.yml::bundle:linux\` — replace inline \`find target -name libcef.so\` with \`bash scripts/resolve-cef-runtime.sh\`. Also pick up \`chrome_crashpad_handler\` + \`snapshot_blob.bin\` (needed by the AppImage build but missing from the prior bundle step).

## Verified

[...]

PR #1408, opened 2026-06-14
fix(build): commit canonical CEF GN args + de-noise libcef resolver

AgentU-asaf:
## Why

Follow-up to #1406 (the slim-libcef release work). I'd flagged "wire the slim libcef into CI" as the open item — but **there is no CI release path**; releases are built locally via `task package:linux`. `resolve-cef-runtime.sh` already auto-resolves to the official-rebuilt `~/cef-build/.../Release_GN_x64` tree and the packager strips it (1.5 GB → 263 MB → 134 MB AppImage), so the next local release is *already* slim. This PR hardens that real path instead of building CI nobody uses.

## What

- **`scripts/cef-build/args.gn`** — version-control the exact GN config that built the shipped v0.45.0 libcef (`is_official_build=true` + thin-LTO — the ~2x size lever). Previously these flags lived only in an untracked `out/Release_GN_x64/args.gn` on one machine; wipe `~/cef-build` and a rebuild silently regresses to the non-official ~414 MB binary.
- **`scripts/cef-build/configure-cef-build.sh`** — one-command configure step: regenerate the gitignored translator C-API wrappers (the `window_ctocpp.cc` "missing and no known rule" gotcha), install `args.gn`, run `gn gen`. Idempotent; honors `AGENTMUX_CEF_SRC`.
- **`scripts/resolve-cef-runtime.sh`** — the `>1 GB` size check used to print `WARNING: ... likely unpatched upstream debug build` on the **correct** unstripped official build (false alarm that would send a future maintainer chasing a ghost). The warning is now scoped to the cef-dll-sys **cargo-cache** candidate — the only one that's genuinely unpatched regardless of size. The dev/override tree gets a soft `INFO` ("unstripped, packager strips at bundle time") instead.
- **docs** — step 4 of `build-patched-libcef.md` now points at the committed args + configure script.

## Verification

- Both scripts pass `bash -n`.
- `resolve-cef-runtime.sh` on the live 1.5 GB official tree now emits the soft INFO and still prints the correct path on stdout.
- Override path still hard-fails on a bad `AGENTMUX_CEF_RUNTIME_DIR` (no silent fallthrough — preserves the #743 invariant).
- No `dist/` / binary artifacts staged; 5 files, all source/docs.

### DIFF
--- a/.changesets/1781474409-fix-build-commit-canonical-cef-gn-args-configure-script-so-l-i2tx.md
+++ b/.changesets/1781474409-fix-build-commit-canonical-cef-gn-args-configure-script-so-l-i2tx.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): commit canonical CEF GN args + configure script so libcef rebuilds are slim by default; stop resolve-cef-runtime crying 'unpatched upstream' on the correct unstripped official build
--- a/docs/cef-build/build-patched-libcef.md
+++ b/docs/cef-build/build-patched-libcef.md
@@ -87,22 +87,24 @@ If you see "class member cannot be redeclared" compile errors later, the patcher
 
 ### 4. Configure the build
 
+The canonical GN args are version-controlled at **`scripts/cef-build/args.gn`** in
+the agentmux repo — the exact configuration that built the shipped v0.45.0 libcef.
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1407, opened 2026-06-14
feat(editor): scratch buffer, file-tree context menu, widget defaults

a5af:
## Summary

- **Editor widget default UX** — clicking the Editor widget now opens with the file tree collapsed and an `Untitled` scratch buffer already open and focused. Users can start typing immediately without touching the file tree.
- **Scratch file infrastructure** — new `createscratchfile` / `movescratchfile` RPCs back scratch buffers with real files in `~/.agentmux/cache/scratch/<uuid>.md`. Content survives crashes; Save As promotes to a user-chosen path.
- **File tree right-click context menu** — right-clicking a file, folder, or empty tree background opens a context menu with: Open, Copy Path, Copy Relative Path, Rename (inline), Delete (with confirm), New File (inline placeholder), New Folder (inline placeholder), Open in Terminal, Reveal in Explorer.
- **Reusable `<ContextMenu>` component** — Portal-based, keyboard-navigable (Esc), closes on outside click, viewport-clamped.
- **7 new backend handlers** — `openinshell` (platform-aware reveal in Explorer/Finder/Files), `renameeditorfile`, `createeditorfile`, `createeditordir`, `deleteeditorfile`, `createscratchfile`, `movescratchfile`. All path-validated against home dir.
- **Optimistic tree updates** — add/remove entries immediately after mutations without waiting for a full re-fetch.
- **Specs** — two new specs (`SPEC_EDITOR_WIDGET_DEFAULT_UX`, `SPEC_FILE_TREE_CONTEXT_MENU`) + freshened `app-api-pane-open.md`.

## Files changed

| Area | Files |
|------|-------|
| Widget config | `agentmux-srv/src/config/widgets.json` |
| Backend handlers | `agentmux-srv/src/server/websocket.rs` |
| ContextMenu component | `frontend/app/components/context-menu.{tsx,scss}` |
| State store | `frontend/app/store/editor-pane-state-store.ts` |
| RPC API | `frontend/app/store/rpc-api.ts` |
| Editor model | `frontend/app/view/editor/editor-model.ts` |
| Tab strip | `frontend/app/view/editor/editor-tab-strip.tsx` |
| Editor view | `frontend/app/view/editor/editor-view.{tsx,scss}` |
| File tree | `frontend/app/view/editor/file-tree.{tsx,model.ts}` |
[...]

PR #1409, opened 2026-06-14
feat(editor): Save As inline path entry for scratch buffers

a5af:
## Summary

- Ctrl+S on an unsaved scratch tab now shows an **inline path-entry field** inside the tab label instead of saving to the cache path
- Ctrl+Shift+S is always available as a Save As shortcut regardless of tab state
- Typing a path and pressing Enter calls `movescratchfile` (already in backend from PR #1407), promotes the tab via the `PromoteScratch` reducer command, and re-opens the file from its new location
- Esc or clicking outside cancels and leaves the tab in scratch state

## How it works

- `editor-view.tsx` — adds `saveAsTabId` signal + `triggerSaveAs()` helper; wires both keymaps inside CodeMirror's keymap extension; passes `saveAsTabId` + handlers down to `EditorTabStrip`
- `editor-tab-strip.tsx` — when `saveAsTabId === tab.id`, renders a `<SaveAsInput>` component in place of the tab label; auto-focuses, commits on Enter, cancels on Esc/blur with a `committed` guard to prevent double-fire
- `editor-view.scss` — `.editor-tab-saveas-input` styled to match the inline rename input in the file tree; `.editor-tab--saveas` widens the tab to fit the input

## Test plan

- [ ] Open editor widget → scratch "Untitled" tab appears
- [ ] Press Ctrl+S → inline path input appears inside the tab with `~/` pre-filled
- [ ] Type a path (e.g. `~/notes.md`), press Enter → file appears at that path, tab label updates to `notes.md`, `isScratch` is cleared
- [ ] Press Esc while input is open → input dismisses, tab stays as Untitled scratch
- [ ] Click elsewhere while input is open → same cancel behavior as Esc
- [ ] On a non-scratch tab, Ctrl+S → normal save (no path input)
- [ ] Ctrl+Shift+S on any scratch tab → opens path input

### DIFF
--- a/.changesets/1781474525-feat-editor-save-as-inline-path-entry-for-scratch-buffers-ct-f2va.md
+++ b/.changesets/1781474525-feat-editor-save-as-inline-path-entry-for-scratch-buffers-ct-f2va.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(editor): Save As inline path entry for scratch buffers (Ctrl+S / Ctrl+Shift+S)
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1418, opened 2026-06-15
fix(srv): unsubscribe WPS broker on WebSocket disconnect — stops ~1000/min handle leak

### LINKED ISSUE #1125: agentmux-srv handle leak — ~1000/min growth, eventsub-correlated
Spun out of #1117 as an independent investigation — it surfaced during the 2026-05-28 renderer-crash-loop forensics but is **not caused by that incident** and pre-dates it.

## Observation

\`agentmux-srv\` accumulates handles at a steady ~1,000/min on a running session, even at low UI activity. Measured 2026-05-28 (UTC):

| Time | Handles |
|---|---:|
| 11:35 (first snapshot) | 38,950 |
| 12:00 | 40,979 |
| 12:02 | 41,721 |
| 12:09 | 43,718 |

Linear growth. WebSocket churn over the same window was minimal (29 connects / 27 disconnects in the entire session, 2 steady-state connections). So the leak isn't per-connection.

\`eventsub\` count was 1,000 vs \`eventunsub\` 38 over the session — that 26× ratio is suspicious and may correlate with the leak.

## Impact today

Low immediate impact: at ~1k handles/min, hitting Windows's per-process handle limit (16M theoretical, lower in practice) takes ~11 days of continuous uptime. \`agentmux-srv\` typically restarts daily on the user's box, which is what's been masking this.

But it's the kind of leak that gets very expensive if a user ever leaves the app running over a long weekend or vacation — and it complicates future supervision logic that wants to keep the sidecar alive across host restarts.

## Suspects (in priority order)

1. **\`eventsub\` handler leak** — high \`eventsub:eventunsub\` ratio suggests subscribers aren't being cleaned up on disconnect. Each subscription may be allocating a mojo / IPC channel handle that isn't dropped when the subscriber goes away.
2. **Unawaited \`tokio::spawn\` per event** — if event-fanout spawns a task per emission without joining or aborting, each spawned task may hold handles to file / socket resources.
3. **File handles for shared session state** — block-file / session-archive paths may be opened with handles that outlive the request scope.
4. **WebSocket per-message resource** — unlikely given the steady-state connection count, but worth confirming.

## Investigation steps

1. **Capture a Sysinternals \`handle.exe -p <srv-pid>\` dump** while handles are high. Group by Type column (File, Event, Mutant, Semaphore, etc.) — the most-leaked type narrows the suspect list dramatically.
2. **Snapshot \`tokio-console\` or equivalent** to see if there are unbounded growing task counts.
3. **Grep \`agentmux-srv/src\`** for \`eventsub\` / subscription registration paths; trace whether unsubscribe is wired up on every disconnect (websocket close, client drop, etc.).
4. **Check for \`Arc\` cycle leaks** — if a subscriber holds a strong ref back to the channel and the channel holds the subscriber, neither side ever drops.

## Out of scope

- Fixing the actual leak (this issue scopes to **diagnosis**; the fix lands in a separate PR once the source is confirmed).
- Generic handle-accounting infrastructure for the sidecar — would be nice but is orthogonal.

## References

- Original surfacing: #1117 (renderer crash-loop hardening) — handle count was noted as an independent observation during forensics
- Retro: \`docs/retro/retro-portable-rm-running-install-2026-05-28.md\`

a5af:
## Root cause

Two bugs in `agentmux-srv/src/server/websocket.rs`:

**Bug 1 — no broker cleanup on disconnect.**
`handle_ws_connection` calls `event_bus.unregister_ws(&conn_id)` when a WebSocket closes, but never calls `broker.unsubscribe_all`. Every `eventsub` the frontend sent during the session stayed registered in the broker permanently. Observed in the wild: ~1,000 `eventsub` calls vs ~38 `eventunsub` calls over a session (issue #1125), with handle count growing linearly at ~1,000/min.

**Bug 2 — all connections shared the hardcoded route ID `"ws-main"`.**
The three broker handlers (`eventsub`, `eventunsub`, `eventunsuball`) all used the string literal `"ws-main"` as the subscriber key. Connections stomped each other's subscription tracking, and the broker's internal `replayed` HashSet accumulated `("ws-main", event, scope)` tuples that were never cleared across reconnects.

## Fix

- Pass `conn_id` into `register_handlers` and use it as the broker `route_id` for all three handlers.
- Add `state.broker.unsubscribe_all(&conn_id)` to the disconnect path, after `event_bus.unregister_ws`.

`EventBusBridge::send_event` ignores `route_id` (it broadcasts to all connections via the event bus), so delivery behavior is unchanged — only per-connection subscription tracking is corrected.

## Impact

Subscriptions are now bounded to the lifetime of the WebSocket connection that created them. On disconnect, the broker clears all entries and `replayed` tuples for that connection. Handle growth stops.

## Test plan

- [ ] Open 5 panes, verify events flow normally (agent stream, sysinfo updates, blockfile events)
- [ ] Close all panes, reopen — confirm no stale event delivery from prior subscriptions
- [ ] On a long session (>30 min), verify `agentmux-srv` handle count stays flat in `muxlog srv` (no ~1000/min growth)

Closes #1125

### DIFF
--- a/.changesets/1781492869-fix-srv-unsubscribe-wps-broker-on-websocket-disconnect-yyi7.md
+++ b/.changesets/1781492869-fix-srv-unsubscribe-wps-broker-on-websocket-disconnect-yyi7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): unsubscribe WPS broker on WebSocket disconnect — stops ~1000/min handle leak (closes #1125)
--- a/agentmux-srv/src/backend/eventbus.rs
+++ b/agentmux-srv/src/backend/eventbus.rs
@@ -96,6 +96,23 @@ impl EventBus {
         }
     }
 
+    /// Send an event to a single connection by conn_id. No-op if not found.
+    pub fn send_to_conn(&self, conn_id: &str, event: &WSEventType) {
+        let data = match serde_json::to_value(event) {
+            Ok(v) => v,
+            Err(e) => {
+                tracing::error!("cannot marshal event: {}", e);
+                return;
+            }
+        };
+        let watches = self.watches.lock().unwrap();
+        if let Some(watch) = watches.get(conn_id) {
+            if watch.sender.send(data).is_err() {
+                tracing::warn!("failed to send event to conn {}", conn_id);
+            }
+        }
+    }
+
     /// Send an event to all connected WebSocket clients.
     pub fn broadcast_event(&self, event: &WSEventType) {
         let data = match serde_json::to_value(event) {
@@ -156,7 +173,7 @@ impl EventBusBridge {
 }
 
 impl WpsClient for EventBusBridge {
-    fn send_event(&self, _route_id: &str, event: WaveEvent) {
+    fn send_event(&self, route_id: &str, event: WaveEvent) {
         // Wrap as RPC eventrecv message (format expected by frontend)
         let ws_event = WSEventType {
             eventtype: WS_EVENT_RPC.to_string(),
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
Good catch — addressed in 633089e8.

Added `EventBus::send_to_conn(conn_id, event)` and updated `EventBusBridge::send_event` to use it when `route_id` is a specific conn_id. Replay delivery now goes only to the subscribing connection. The `"ws-main"` broadcast path is retained as a fallback (no live callers remain after this fix) to avoid a silent no-op if any external path passes it.

a5af:
Both stale comments updated in c3806be2 —  references replaced with per- framing in the  type alias doc and the  doc block.

PR #1420, opened 2026-06-15
feat(build): release-time BeginWindowDrag-patch gate for Linux libcef.so

AgentU-asaf:
## Why

Closes the last open item from the slim-libcef work (#1406, #1408): nothing forced a check that the shipped `libcef.so` still carries AgentMux's **BeginWindowDrag** patch. Bundle the upstream prebuilt CEF (the cef-dll-sys cargo-cache fallback) and left-click window drag silently no-ops — the runtime ABI guard only surfaces it *after* the user clicks the title bar and nothing happens.

## How

`scripts/verify-cef-patch.sh <runtime-dir | libcef.so>` greps `.symtab` for the patch symbol (`window_begin_window_drag_<apiver>` / `CefWindowImpl::BeginWindowDrag`):

| Exit | Meaning |
|---|---|
| `0` | patched (symbol found) |
| `1` | **unpatched upstream** — symbol table present, no slot (the regression) |
| `2` | stripped / can't verify (run on the unstripped build) |

**Key constraint:** the symbol survives only in the **unstripped** `.so` — `strip` removes `.symtab` entirely, so the shipped 263 MB binary has zero trace. The check therefore runs on the unstripped runtime, *before* packaging strips it. (Empirically confirmed: `strings`/`nm` find the symbol in the 1.5 GB build output, nothing in the stripped copy.)

## Wiring

- **`bundle:linux`** — advisory (`|| true`): prints a warning but proceeds, so `task dev` still works on the upstream cef-dll-sys fallback (no 3-hour CEF build needed).
- **`build-appimage-linux.sh`** — **hard gate** (release-only script; never run by `task dev`): refuses to package an AppImage whose `libcef.so` lacks the patch. If `dist/cef` was pre-stripped out-of-band, it falls back to verifying the resolved unstripped source. `AGENTMUX_SKIP_CEF_PATCH_CHECK=1` overrides (emergency only).

Mirrors the existing `verify-cef-version.sh` gate on the Windows bundle.

## Verification

All three exit codes tested against real binaries on this machine:
- patched build tree (1.5 GB unstripped) → **exit 0** (~5s nm scan)
- upstream cargo-cache `.so` (1.3 GB unstripped) → **exit 1** ✅ catches the regression
- stripped `dist/cef/libcef.so` (263 MB) → **exit 2** → gate falls back to resolved source → **pass**

`task --list` + PyYAML confirm the Taskfile edit is valid; both scripts pass `bash -n`.

### DIFF
--- a/.changesets/1781494069-feat-build-add-release-time-beginwindowdrag-patch-gate-for-l-z63w.md
+++ b/.changesets/1781494069-feat-build-add-release-time-beginwindowdrag-patch-gate-for-l-z63w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(build): add release-time BeginWindowDrag-patch gate for Linux libcef.so (verify-cef-patch.sh) — advisory in bundle:linux, hard gate in build-appimage-linux.sh
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -631,6 +631,10 @@ tasks:
                 # See docs/specs/patched-libcef-bundling-2026-05-08.md.
                 CEF_DIR="$(bash scripts/resolve-cef-runtime.sh)"
                 echo "Using CEF runtime from: $CEF_DIR"
+                # Advisory BeginWindowDrag-patch check — non-fatal here so `task dev`
+                # still works on the upstream cef-dll-sys fallback (no 3-hour CEF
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1424, opened 2026-06-15
fix(startup): self-heal + recovery UI for host-bridge init failure

AgentX-asaf:
## Problem

When the frontend can't establish `window.api` (the CEF host bridge), the app died on a bare dead screen — *"AgentMux Failed to Start … Check the browser console (F12)"* (rendered by `bootstrap.ts`'s catch) — with **no recovery action**.

The failure is almost always **transient**: a Vite full-reload mid-change in dev (e.g. a `git checkout` in the served clone), a slow backend spawn, or a stale pooled window whose host IPC went away. The backend is usually reachable a moment later.

### Observed incident (2026-06-15)
A dev instance logged `window.api still undefined after 5s` **385 times** — a reload-and-fail *storm* ~31 min after a clean startup. Root cause: `git checkout`/`rebase` in the clone Vite was serving → mass file churn → Vite full reloads over a momentarily-inconsistent module tree → `bootstrap()` couldn't finish the bridge handshake before `initApp`'s 5s guard. And `Ctrl+R` did nothing because the **CEF app window has no reload keybinding** — so close+reopen was the only way out.

## Fix (mirrors the existing WebGL-context-loss recovery in `bootstrap.ts`)

- **`error-display.ts`**: bounded, `sessionStorage`-guarded **auto-reload** (max 3, backoff, reset on success) — a persistent failure can **never** storm (the 385× guard). On exhaustion, a friendly **recovery card** with a one-click **Reload** + **Reopen** and collapsible technical details replaces the dead div. Adds `showReconnecting` overlay + `tryAutoRecover`. `location.reload()` works here (real `localhost` URL, not the host's `data:` crash page) and preserves the IPC port/token query params.
- **`bootstrap.ts`**: catch → `tryAutoRecover` instead of the bare div; clears the reload budget on successful startup; and wires **`Ctrl+R` / `F5` → reload** (registered capture-phase *before* the handshake, so it works even when the UI is wedged — the exact thing the user reached for).

## Test
- `task build:frontend:dev` → clean.
- Manual (per spec §6): with a dev instance running, `git checkout` another branch in the served clone → expect "Reconnecting… (n/3)" → auto-recovers once the tree settles, instead of the 385× dead-screen storm. Killing the host IPC → recovery card with a working Reload.

## Operational note
Don't run git branch ops in a clone a dev instance is actively serving — use `git worktree` or stop `task dev`. This PR makes that mistake **recoverable**, not impossible.

Spec: `docs/specs/SPEC_BRIDGE_INIT_RECOVERY_2026_06_15.md`.

### DIFF
--- a/.changesets/1781522760-fix-startup-self-heal-recovery-ui-for-host-bridge-init-failu-lrmz.md
+++ b/.changesets/1781522760-fix-startup-self-heal-recovery-ui-for-host-bridge-init-failu-lrmz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(startup): self-heal + recovery UI for host-bridge init failure (auto-reload guard, Reload button, Ctrl+R/F5 keybinding)
--- a/docs/specs/SPEC_BRIDGE_INIT_RECOVERY_2026_06_15.md
+++ b/docs/specs/SPEC_BRIDGE_INIT_RECOVERY_2026_06_15.md
@@ -1,0 +1,226 @@
+# SPEC: Host-Bridge Init Failure — Self-Heal + Recovery UI
+
+**Date:** 2026-06-15
+**Status:** Draft
+**Scope:** `frontend/app/init/error-display.ts` + `frontend/app-init.ts` startup-error path
+
+---
+
+## 1. Problem
+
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1425, opened 2026-06-15
fix(agent-pane): scope subagent ⚡ panels to the owning pane

a5af:
## Problem

Agent panes accumulate panels with a thunder symbol (⚡) and a short hash code. These are **subagent link badges** (`SubagentLinkBlock`) representing Claude Task-tool subagents. They were appearing in agent panes **unrelated** to where the subagent was spawned — most visibly for subagents spawned by Claude instances running in **terminal** panes, which should never surface in an agent pane at all. Clicking one opens the subagent's raw session JSONL, which renders unformatted.

## Root cause

A missing parent-pane filter on **both** ends:

- **Backend** (`subagent_watcher.rs`): `subagent:spawned` / `:activity` / `:completed` are broadcast **globally** to every websocket client, and the payload carried only the parent Claude `agentId` — no identifier for the owning pane/block.
- **Frontend** (`useSubagentEvents.ts`): the hook is mounted in **every** agent pane and appended a `subagent_link` node for **any** event with an `agentId` — zero filtering. So every pane rendered every subagent spawned anywhere.

## Fix

Thread the registering **block id** through the watcher and stamp it on every emitted event as `parentBlockId`. The block id is already available at the `reactive/register` call (`reactive.rs`). The frontend hook now receives the pane's own `blockId` and renders a subagent only when `parentBlockId` matches — so terminal-spawned (and other panes') subagents are dropped.

### Changes
- `agentmux-srv/src/backend/subagent_watcher.rs` — `watch_agent` takes `parent_block_id`, threaded through `scan_existing_subagents` / `process_jsonl_change`; added `parentBlockId` to all three event payloads.
- `agentmux-srv/src/server/reactive.rs` — pass `req.block_id` into `watch_agent`.
- `agentmux-srv/src/server/service.rs` — `subagent.WatchAgent` RPC accepts an optional block_id arg (defaults to `""`).
- `frontend/app/view/agent/hooks/useSubagentEvents.ts` — new required `blockId` option; filter `parentBlockId === blockId` on spawned + completed.
- `frontend/app/view/agent/agent-view.tsx` — pass `model.blockId`.

## Scope / notes

- Subagent watching is triggered **only** by the `reactive/register` HTTP path, which fires only for **terminal** blocks (via the `AGENTMUX_AGENT_ID` OSC). Agent panes don't register their own Claude for watching today, so this filter cleanly eliminates the cross-pane leak with no loss of legitimate behavior.
- **Follow-ups (out of scope):** (1) wiring agent panes to surface their *own* subagents, and (2) the poorly-formatted raw-JSONL rendering shown when a subagent panel is expanded.

## Testing
- `cargo check -p agentmux-srv` passes.
- Frontend `node_modules` not installed in this checkout; changes are a single new required string option + a string equality guard (no new types).

### DIFF
--- a/.changesets/1781523209-fix-agent-pane-scope-subagent-panels-to-the-owning-pane-so-t-llx1.md
+++ b/.changesets/1781523209-fix-agent-pane-scope-subagent-panels-to-the-owning-pane-so-t-llx1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): scope subagent ⚡ panels to the owning pane so terminal-spawned subagents stop leaking into unrelated agent panes
--- a/agentmux-srv/src/backend/subagent_watcher.rs
+++ b/agentmux-srv/src/backend/subagent_watcher.rs
@@ -109,7 +109,18 @@ impl SubagentWatcher {
 
     /// Start watching a Claude Code agent's session directory for subagent files.
     /// Spawns a background tokio task for debounced file event processing.
-    pub fn watch_agent(self: &Arc<Self>, agent_id: &str, config_dir: PathBuf) {
+    ///
+    /// `parent_block_id` is the pane/block that owns this Claude instance (from
+    /// the reactive register request). It is stamped onto every emitted subagent
+    /// event so the frontend can route the ⚡ panel to the originating pane only,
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1427, opened 2026-06-15
feat(editor): theme-bound colors + tight line-number gutter

a5af:
## What

Two editor polish changes requested:

1. **Colors bound to the global theme.** The editor previously used a hardcoded `oneDark` CodeMirror theme. Now the chrome — background, text, gutters, cursor, selection, search matches, tooltips — is bound to the app's theme tokens (`--block-bg-color`, `--main-text-color`, `--accent-color`, `--secondary-text-color`, `--highlight-bg-color`, `--border-color`, …), so the editor follows whatever theme the user has active (tokyo-night, nord, monokai, dracula, gruvbox, high-contrast, …). Syntax-token colors still come from oneDark's highlight style — the theme system has no per-token color tokens, and that palette reads well on every dark theme.

2. **Tight line-number gutter.** Removed the large gap between the line numbers and the code. There's no debugger/breakpoint margin to reserve, so:
   - the unused **fold-gutter** column is hidden,
   - the **lint-gutter** marker column is hidden (collapsed to 0 width),
   - line-number right padding is collapsed to **1px**.

   `lintGutter()` stays in the extension set because it installs the lint state field that backs the **inline diagnostic underlines** (LSP pushes via `setDiagnostics`). Only its empty marker column is hidden in CSS, so diagnostics still render — just without the reserved gap.

## Files

- `frontend/app/view/editor/editor-theme.ts` (new) — `editorTheme` extension: CSS-var-bound chrome + `syntaxHighlighting(oneDarkHighlightStyle)`.
- `frontend/app/view/editor/editor-view.tsx` — swap `oneDark` → `editorTheme`; keep `lintGutter()` with an explanatory comment.
- `frontend/app/view/editor/editor-view.scss` — hide `.cm-foldGutter` / `.cm-gutter-lint`, tighten `.cm-lineNumbers` padding.

### DIFF
--- a/.changesets/1781524995-feat-editor-theme-bound-colors-tight-gutter.md
+++ b/.changesets/1781524995-feat-editor-theme-bound-colors-tight-gutter.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(editor): bind CodeMirror colors to the global theme and tighten the line-number gutter
--- a/frontend/app/view/editor/editor-theme.ts
+++ b/frontend/app/view/editor/editor-theme.ts
@@ -1,0 +1,73 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+//
+// CodeMirror chrome theme bound to the app's global theme tokens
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1431, opened 2026-06-15
fix(widgets): repair phantom drop indicator after right-click unpin

a5af:
## Summary

- **Bug**: After unpinning a widget via right-click → context menu, hovering over the remaining widget bar showed the accent-colored drop indicator. Clicking any widget made it disappear. The indicator should only appear during an active left-button drag.
- **Root cause**: `onPointerDown` fired for all mouse buttons including right-click (button 2). The right-button `pointerup` was swallowed by the context-menu overlay (which receives focus before the button releases), leaving `dragStartRef` set. Subsequent mouse movement > 5px over any slot activated drag mode.
- **Structural issue**: Drag state was split across 3 mutable refs + 2 reactive signals — two sources of truth that could diverge on any escape path.

## Changes

**`frontend/app/window/action-widgets.tsx`**
- Replaced `dragStartRef` / `draggingKeyRef` / `dropIndexRef` / `draggingKey` signal / `dropIndex` signal with a single `DragState` discriminated union signal (`idle | pending | dragging`)
- Added `e.button !== 0` guard in `handlePointerDown` — right/middle clicks never enter `pending` state
- Added `createEffect` that registers global `pointercancel` + `visibilitychange` cleanup while not idle — handles Alt+Tab, OS interrupts, and other escape scenarios
- `handlePinnedContextMenu` resets drag state as belt-and-suspenders before opening menu
- `handlePointerMove` now computes `dropIndex` before updating state — single `setDragState` per frame instead of two

**`specs/SPEC_WIDGET_PINBAR_DND_STATE_2026_06_15.md`** — full analysis and design spec

## Test plan

- [ ] Right-click a pinned widget → "Unpin from bar" → hover over remaining widgets → **no drop indicator appears**
- [ ] Left-drag a pinned widget to reorder → drop indicator appears during drag, clears on release — reorder commits correctly
- [ ] Left-drag → release outside the window → no stale indicator on return
- [ ] Alt+Tab while left-dragging → return → indicator is cleared
- [ ] Right-click opens context menu normally, closes normally — no visual regression

### DIFF
--- a/.changesets/1781500000-fix-widget-pinbar-phantom-drop-indicator-after-rightclick-unpin-x3k2.md
+++ b/.changesets/1781500000-fix-widget-pinbar-phantom-drop-indicator-after-rightclick-unpin-x3k2.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+fix(widgets): repair phantom drop indicator after right-click unpin — replaces brittle ref/signal dual-state with a single DragState machine, adds primary-button guard and global OS-interrupt cleanup (closes #1432)
--- a/frontend/app/window/action-widgets.tsx
+++ b/frontend/app/window/action-widgets.tsx
@@ -322,13 +322,37 @@ const ActionWidgets = (): JSX.Element => {
     });
 
     // ── Drag-to-reorder (pinned only, saves to widget:pinned) ─────────────────
+    //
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1434, opened 2026-06-15
fix(tabs): more padding + live resize while renaming

AgentA-asaf:
## Summary

- `.tab-inner` padding increased from `6px` → `10px` each side — matches VS Code's tab breathing room
- `TAB_PADDING_BUDGET` updated from `36` → `52px` so `measureTabWidth()` accounts for the real CSS padding
- Added `onInput` handler to the contentEditable rename field — tab width tracks keystrokes live as you type
- `handleBlur` now also calls `onNaturalWidth` with the committed name so the tab snaps to the correct size immediately on Enter/click-away/Escape

## Test plan

- [ ] Tabs look visually roomier (not clipped at text edges)
- [ ] Double-click rename → tab grows/shrinks in real time as you type
- [ ] Press Enter → tab locks to the new name width
- [ ] Press Escape → tab snaps back to the original name width
- [ ] Short label still produces a narrow tab (not over-widened)

### DIFF
--- a/.changesets/1781540118-fix-tab-padding-and-live-resize.md
+++ b/.changesets/1781540118-fix-tab-padding-and-live-resize.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tabs): increase tab padding to match VS Code feel; resize live while typing during rename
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #1435, opened 2026-06-15
fix(agent): tool live-tail shows last stdout/stderr, not system chunks

AgentA-asaf:
## Problem

For a Bash tool call like `sleep 10 && cat file.txt`, the collapsed ToolBlock row showed:

```
↳ [bashwrap] starting: 184 chars
```

…for the entire duration of the sleep. This is a `kind:"system"` internal bashwrap housekeeping chunk, not agent-visible output. The bug was `ToolBlock.tsx:249` blindly taking `chunks[chunks.length - 1]` regardless of chunk kind.

## Changes

- **`ToolBlock.tsx`** — walk chunks backwards to find the last `kind:"stdout"` or `kind:"stderr"` chunk; skip `kind:"system"` chunks in the live-tail
- **`ToolBlock.tsx`** — when no stdout/stderr has arrived yet (e.g. during `sleep`), show a `ToolElapsedTicker` component (`3s…`) so the user knows the tool is alive
- **`stream-parser.ts`** — stamp `timestamp: Date.now()` on tool_call creation so the elapsed ticker has a reliable start time
- **`_document-nodes.scss`** — add `.agent-tool-live-tail--waiting` modifier (dim, no background) so the timer reads as pending, not real output

## Before / After

| State | Before | After |
|---|---|---|
| `sleep 10 && cat file` during sleep | `↳ [bashwrap] starting: 184 chars` | `3s…` → `7s…` |
| First stdout arrives | `↳ [bashwrap] starting: 184 chars` | `↳ <first line of output>` |
| Multiple lines | May show a system chunk | Shows last stdout/stderr line |

## Test plan

- [ ] `sleep 5 && echo done` — row shows `0s…` → increments → `↳ done` when echo fires
- [ ] Fast command `echo hello` — no ticker flicker, shows `↳ hello` immediately
- [ ] Command with only stderr (`cat nonexistent`) — live-tail shows the stderr line
- [ ] Overlay panel still shows full chunk list including system chunks (unchanged)

### DIFF
--- a/.changesets/1781541229-fix-tool-live-tail-stdout-filter.md
+++ b/.changesets/1781541229-fix-tool-live-tail-stdout-filter.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): tool live-tail skips system chunks — shows last stdout/stderr or elapsed timer while waiting
--- a/frontend/app/view/agent/components/ToolBlock.tsx
+++ b/frontend/app/view/agent/components/ToolBlock.tsx
@@ -38,6 +38,19 @@ import { createBlock } from "@/store/global";
 import type { ToolNode } from "../types";
 import { ToolBlockOverlay } from "./ToolBlockOverlay";
 
+// Ticks every second while a tool is running with no output yet.
+function ToolElapsedTicker(props: { startMs: number }): JSX.Element {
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1436, opened 2026-06-15
feat(trust-center): secure API-key entry UI (Phase 2b)

a5af:
## Summary

Phase 2b of `specs/SPEC_TRUST_CENTER_2026_06_15.md` — the **frontend key lifecycle**, wired to the Phase 2a backend (`account.key.verify`). This makes the secure key flow usable end-to-end (§5/§5.1/§6).

- **New "OS Keychain (validated)" secret backend** in the account form — now the default for new accounts. Selecting it swaps the generic env/secrets-manager/plaintext fields for the secure key flow.
- **Entry → Validate → Locked**: a password input (autocomplete off) with two actions — **Validate & Save** (the single user-initiated egress probe) and **Save without validating** (air-gapped escape hatch).
- **§5.1 egress transparency**: an inline ⓘ note next to Validate names the **exact destination endpoint** per provider (`api.github.com/user`, `api.anthropic.com/v1/models`) so the user sees where the key goes *before* clicking. Providers without a validator show save-without-validating guidance instead.
- **Non-recoverable**: plaintext is dropped from the form immediately after submit and never read back. Existing keychain accounts render **locked** (masked tail, e.g. `••••••••3f9a`) and require **Replace key** to re-enter.
- **AccountDetail** shows "Stored in: OS keychain" + the masked tail.

## Verification
- `npm run build` — clean.
- `stylelint` on the new SCSS — clean.

## End-to-end (with Phase 1 + 2a)
Trust Center → Accounts → Add → pick GitHub/Anthropic → paste key → **Validate & Save** → backend probes the live service, stores the key in the OS keychain, returns only metadata + masked tail → account shows locked + validated. At agent spawn, the resolver pulls the key from the keychain.

## Deferred (Phase 3+)
Broader service catalog + OAuth (Google/Slack/etc.), AWS key validation (SigV4), account↔identity quick-bind.

### DIFF
--- a/.changesets/1781530200-feat-trust-center-key-ui.md
+++ b/.changesets/1781530200-feat-trust-center-key-ui.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(trust-center): secure API-key entry UI — paste a key, Validate (one user-initiated probe with an inline egress notice) or save without validating; keys are stored in the OS keychain and shown masked + non-recoverable with a Replace action
--- a/agentmux-srv/src/server/agent_handlers.rs
+++ b/agentmux-srv/src/server/agent_handlers.rs
@@ -112,6 +112,20 @@ struct VerifyKeyReq {
     /// Set to replace the key on an existing account; empty mints a new one.
     #[serde(default)]
     account_id: String,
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1437, opened 2026-06-15
fix(startup): retry IPC on stale token; stop the recovery-reload loop

AgentX-asaf:
## Problem

A window could get stuck in the recovery UI looping "Reconnecting…" — clicking **Reload** never recovered. Two compounding bugs:

1. **The frontend gave up on a transient stale token.** The host re-injects fresh, self-consistent `ipc_port` + `ipc_token` into the window on **every** page load (`on_load_end`, `agentmux-cef/src/client/mod.rs:1214`). But on a reload the frontend's first `invokeCommand` can race *ahead* of that injection, hit a **401** (or missing creds), and fail bridge init outright — wedging the window, even though valid creds arrive a moment later.
2. **The recovery card looped.** Its Reload button reset the auto-reload budget and re-entered the bounded `Reconnecting…` cycle, so a *persistent* failure (stopped/restarted host) looped forever instead of telling the user to restart.

(Root cause understood after mapping the windows / IPC-token / reducer architecture — the host already provides a fresh token on every load; the frontend just wasn't waiting for it.)

## Fix

- **`ipc.ts` `invokeCommand` (cef):** re-read the `window.__AGENTMUX_IPC_PORT__/TOKEN__` globals **each attempt** and retry (≤6 × 250ms) on missing creds / network error / **401**, so the host's freshly-injected token self-heals the race. Non-401 HTTP errors and IPC-level failures still throw immediately.
- **`error-display.ts`:** stop resetting the reload budget on exhaustion *and* on the card's manual Reload (that was the loop). Once the budget is spent, a reload is a **single** attempt that returns to the card instead of looping the spinner; the budget clears only on **successful** startup. When reloading isn't working, the card escalates to **"Can't reconnect — restart AgentMux"**.

## Result
- Transient stale-token race (the common case after a reload) **self-heals silently** via the IPC retry — the recovery UI usually isn't even reached.
- A genuinely dead/restarted host shows a **terminal** card with honest guidance, **no infinite loop**.
- Manual **Reload** is meaningful: reload → host re-injects creds → retry succeeds if the host is back; otherwise straight back to the card.

## Test
- `task build:frontend:dev` clean.
- Follows up `SPEC_BRIDGE_INIT_RECOVERY_2026_06_15.md` (the recovery UI from #1424).

### DIFF
--- a/.changesets/1781542495-fix-startup-retry-ipc-on-stale-token-stop-the-recovery-reloa-8xu9.md
+++ b/.changesets/1781542495-fix-startup-retry-ipc-on-stale-token-stop-the-recovery-reloa-8xu9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(startup): retry IPC on stale token + stop the recovery-reload loop
--- a/.changesets/1781542495-fix-startup-retry-ipc-on-stale-token-stop-the-recovery-reloa-8xu9.md
+++ b/.changesets/1781542495-fix-startup-retry-ipc-on-stale-token-stop-the-recovery-reloa-8xu9.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-fix(startup): retry IPC on stale token + stop the recovery-reload loop
+fix(startup): wait for re-injected IPC creds before bridge init; retry on stale token; stop the recovery-reload loop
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #1440, opened 2026-06-15
feat(trust-center): add OpenAI key-based account provider (Phase 3 — catalog)

a5af:
## Summary

The key-based slice of Phase 3 (`specs/SPEC_TRUST_CENTER_2026_06_15.md`) — expand the Accounts catalog to **OpenAI**, which the Phase 2a key validator already probes (`GET api.openai.com/v1/models`). **No backend change** — the validator supports it; only the frontend catalog needed it.

- `AccountProvider += "openai"`; `PROVIDER_LABELS` / `PROVIDER_COLORS` entries.
- Account form: OpenAI option + API Key kind; `KEY_VALIDATION_ENDPOINT` entry so the §5.1 egress note names the exact endpoint.
- `accountsByProvider` order, `AssignmentsTab` providers, and the per-agent `AgentIdentityPanel` provider list all include OpenAI.
- `ProviderLogo` already ships an OpenAI icon.

Users can now add + validate an OpenAI key through the same secure keychain lifecycle shipped in 2a/2b (Validate → masked/non-recoverable → OS keychain).

## ⚠️ OAuth is NOT in this PR (blocked)

Full OAuth for Google / Microsoft / Slack (spec §4.2, §12.1) is **blocked on a product decision**: it needs registered OAuth-app **client IDs** (which don't exist yet) plus the **PKCE-vs-BYO-secret** choice (§11 Q2). Building the OAuth client now would be untestable scaffolding, so it's deferred until those are settled. Slack is also omitted from the key catalog for now — no bundled brand icon (the backend validator already supports it whenever an icon is added).

## Verification
- `npm run build` — clean.

### DIFF
--- a/.changesets/1781533200-feat-trust-center-openai-account.md
+++ b/.changesets/1781533200-feat-trust-center-openai-account.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(trust-center): add OpenAI as a key-based account provider (validated against api.openai.com/v1/models), with icon and per-agent assignment support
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #1441, opened 2026-06-15
feat(agent): interactive AskUserQuestion panel + answer delivery

a5af:
## What

Makes the **`AskUserQuestion`** tool work inside AgentMux. When an agent calls it, AgentMux now renders an interactive question panel and routes the user's answer back to the agent — previously the call rendered as a generic "running" tool that **never resolved** (the agent hung; this is the failure that surfaced when an `AskUserQuestion` call errored out in an agent pane).

## Validated first (SPEC §10)

Before building, I ran a two-phase smoke test against the bundled `claude` CLI with the exact persistent stream-json args. Result: **`emits_AskUserQuestion = True`** and **`consumes_tool_result = True`** — claude emits the tool_use, blocks the turn, and continues once fed a flat-text `tool_result`. Confirmed the exact wire schema (`input.questions[]`). This de-risked the one load-bearing unknown; all spec open questions are resolved.

## How

- **Detect** (`stream-parser.ts`): `AskUserQuestion` already arrives as a `tool_call`; special-case it → `ToolNode{ status: "awaiting_answer", question }`.
- **Render** (`AgentQuestionPanel.tsx/.scss`): N questions, single/multi-select, required free-text **Other**, submit / answer-later, theme-bound styling. Collected via `pendingQuestions()` (mirror of `pendingDecisions()`).
- **Deliver**: `handleAnswer()` → `RpcApi.AgentAnswerCommand` → `agent.answer` handler → `PersistentSubprocessController.send_tool_result()`, which writes a `{type:"user", content:[{type:"tool_result", tool_use_id, content}]}` block to the **live stdin** of the host agent process. Optimistic `awaiting_answer → success` transition dismisses the panel; the agent's continuation drives the rest.

## Scope (Phase 1)

- **Host (persistent) agents** — they keep stdin open (`--input-format stream-json`). Container/one-shot agents return `UNSUPPORTED_CONTROLLER` (Phase 2, per the analysis).
- Flat labelled-text answer encoding (validated). Idempotency/structured JSON deferred.

## Verification

`cargo check -p agentmux-srv` passes (only pre-existing warnings). Frontend couldn't be `tsc`'d in this checkout (`node_modules` absent); changes mirror the existing `AgentDecisionPanel`/`pendingDecisions`/`ToolDecisionCommand` patterns.

Spec: `docs/specs/SPEC_ASK_USER_QUESTION_2026_06_15.md` (includes the §10 validation results).

### DIFF
--- a/.changesets/1781543065-feat-agent-ask-user-question.md
+++ b/.changesets/1781543065-feat-agent-ask-user-question.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): render AskUserQuestion as an interactive panel and deliver the answer to the agent
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -167,6 +167,30 @@ impl PersistentSubprocessController {
             .map_err(|e| format!("stdin send failed: {e}"))
     }
 
+    /// Deliver an AskUserQuestion answer to the running CLI as a `tool_result`
+    /// on the live stdin, unblocking the agent's turn. Mirrors `send_message`
+    /// but emits a user turn whose content is a single `tool_result` block
+    /// keyed on the original `AskUserQuestion` tool_use id. The process must
+    /// already be running (the agent is mid-turn, blocked on this answer) — we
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

PR #1442, opened 2026-06-15
fix(markdown): make table-of-contents reactive to streaming text

### LINKED ISSUE #789: Markdown component: TOC items not reactive when text streams
## Summary

Codex flagged this as a P2 during PR #786 review (virt redesign cascade). Filed as follow-up since it's tangential to the agent pane virtualization fix that PR addressed.

## The bug

In \`frontend/app/element/markdown.tsx\` around line 382:

When \`showTocAtom\` is enabled and \`props.text\` changes while \`Markdown\` stays mounted, the \`renderedMarkdown\` memo re-runs (correctly) but headings are accumulated into a non-reactive \`tocItems\` array. The TOC \`<Show>\`/\`map\` below isn't driven by that memo, so the visible TOC stays stuck on the initial heading set or shows stale/duplicated items after a remount.

## Affected scenarios

- Settings panes with TOC enabled
- Help pages with TOC enabled
- Any other long-form markdown view that opts into TOC

**Not affected**: agent pane (never sets \`showTocAtom\` — short-lived per-message markdown only).

## Fix sketch

Move \`tocItems\` and the heading-accumulator side effect into a memo or store keyed off \`renderedMarkdown\`. Re-derive on each text change so the TOC tracks the current state.

## Origin

- PR #786 (merged) — fixed Markdown's `text` reactivity for streaming agent pane (codex P1)
- This issue — the same review's P2 flagging that TOC isn't reactive

a5af:
## Problem (#789)

In `frontend/app/element/markdown.tsx`, `tocItems` was a plain mutable array that the `renderedMarkdown` memo **pushed** into on every render and **never cleared**. When `props.text` changes while `Markdown` stays mounted (settings/help panes with TOC enabled), two bugs followed:

1. The `<Show>`/`.map()` over the plain array wasn't reactive — the visible TOC stayed stuck on the initial heading set.
2. `tocItems.push(...)` accumulated headings across re-renders — producing stale/duplicated TOC items.

(Originally flagged by Codex as a P2 during the PR #786 virtualization review.)

## Fix

- `renderedMarkdown` now returns `{ element, toc }`. The toc is collected into the memo-local `tocRef` (fresh each run), so it reflects only the current text.
- A derived `const tocItems = createMemo(() => renderedMarkdown().toc)` drives the TOC reactively.
- JSX consumers updated: `renderedMarkdown().element` for content, `tocItems()` for the TOC list/empty-state.

Net effect: the TOC re-derives on each text change and no longer accumulates duplicates.

## Scope

Single file. Agent pane is unaffected (it never sets `showTocAtom`). No behavior change for non-TOC markdown — `renderedMarkdown().element` is the same JSX as before.

## Testing

`node_modules` isn't installed in this checkout, so `tsc` wasn't run; the refactor is type-safe by construction (memo return type changed to `{ element, toc }`, all consumers updated accordingly).

Closes #789

### DIFF
--- a/.changesets/1781543103-fix-markdown-make-the-table-of-contents-reactive-to-streamin-xv89.md
+++ b/.changesets/1781543103-fix-markdown-make-the-table-of-contents-reactive-to-streamin-xv89.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(markdown): make the table-of-contents reactive to streaming text
--- a/frontend/app/element/markdown.tsx
+++ b/frontend/app/element/markdown.tsx
@@ -373,7 +373,6 @@ const Markdown = (props: MarkdownProps) => {
         onClickExecute,
     } = props;
     const textAtomValue = useAtomValueSafe<string>(textAtom as any);
-    const tocItems: TocItem[] = [];
     const showToc = useAtomValueSafe(showTocAtom) ?? false;
     const [focusedHeading, setFocusedHeading] = createSignal<string | null>(null);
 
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

a5af:
@reagentx-workflow review

PR #1443, opened 2026-06-15
feat(trust-center): service OAuth 2.0 client scaffold (Phase 3 — OAuth)

a5af:
## Summary

Backend OAuth 2.0 client for Trust Center **service accounts** (Google/GitHub/Microsoft/Slack), built to the native-app security RFCs. Researched against **RFC 8252** (Native Apps), **7636** (PKCE), **8628** (Device Grant), and **9700** (Security BCP). Distinct from the CLI-provider OAuth in `auth_session.rs` — this drives the flow itself.

**`oauth_client.rs`:**
- **Authorization Code + PKCE (S256)** over a transient **loopback** `127.0.0.1:<ephemeral>` redirect (Google, Microsoft) — `127.0.0.1` literal, not `localhost`, per §7.3.
- **Device Authorization Grant** (RFC 8628) for **GitHub** (no client secret), with `interval`/`slow_down`/`expired_token` handling.
- High-entropy **`state`** generated and verified on the callback (CSRF, §8.9).
- Per-provider **catalog** with endpoints + flow choice; `client_id` intentionally `None`.
- **BYO** client_id/secret for secret-mandatory providers (Slack) and as an override; `resolve_client()` **gates** with a clear "not configured" error until a client id exists.
- Tokens stored in the **OS keychain** (JSON blob via `secret_store`); DB carries only the pointer; rollback on persist failure.
- Process-global `OAuthSessionManager` with a 5-min timeout sweep.

**RPCs:** `account.oauth.start` / `poll` / `cancel`, modeled on the `auth.*` set.

**Tests (7, all pass):** PKCE S256 derivation, verifier/state uniqueness, callback parse + percent-decode, catalog flow choices, config gating, BYO-secret requirement.

## ⚠️ Scaffold status (intentional)

No client ids are provisioned yet, so `start()` returns the **"not configured"** error and **no flow runs**. This is the reviewed, security-correct engine — dropping in client ids (or supplying BYO) **activates it**. The frontend Connect button + the spawn-time access-token refresh for OAuth accounts land when the ids arrive (both are untestable without them).

## Verification
- `cargo test … oauth_client::tests` — 7 passed.

### DIFF
--- a/.changesets/1781536800-feat-trust-center-oauth-scaffold.md
+++ b/.changesets/1781536800-feat-trust-center-oauth-scaffold.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(trust-center): service OAuth 2.0 client scaffold — Authorization Code + PKCE (loopback) and Device Flow per RFC 8252/7636/8628/9700, with account.oauth.* RPCs; gated on per-provider client ids (inert until provisioned or supplied as BYO)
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1444, opened 2026-06-15
feat(agent): AskUserQuestion answers for one-shot/container agents (Phase 2)

a5af:
## What

Phase 2 of AskUserQuestion (`SPEC_ASK_USER_QUESTION_2026_06_15.md`): make the question panel functional for **one-shot / container** agents, not just persistent host agents.

## The finding that reshaped the approach

I empirically tested the originally-planned mechanism (resume the session with `--resume <sid>` and feed back a `tool_result`). It **does not work**:

```
one_shot_emits_askq   = True     # claude -p DOES emit the AskUserQuestion tool_use
one_shot_turn_ends    = True     # the turn then ends with a `result`
resume + tool_result  = no-op    # the model logs "the user sent an empty message" and just greets
```

When a one-shot turn ends, the CLI **abandons** the pending `AskUserQuestion` tool_use, so a later `tool_result` for it is dropped. There's no dangling tool_use to complete. (Recorded in SPEC §10.1.)

## What ships instead

Deliver the answer as a **normal follow-up user turn** over the existing `--resume` `agentinput` path. The agent resumes the session and continues from the question with the answer as context — the ordinary, working resume path; only the *tool_result block* is what the abandoned tool_use can't consume.

**Frontend-only** change: the panel first tries `agent.answer` (Phase 1 tool_result path, for persistent agents); on `UNSUPPORTED_CONTROLLER` it routes `answer_text` through the standard `handleSendMessage`. Optimistic success is kept; the panel no longer dead-ends on one-shot/container agents. Other failures still roll the node back.

## Notes

- Backend unchanged — the Phase 1 `UNSUPPORTED_CONTROLLER` response is the trigger.
- `handleAnswer` moved below `handleSendMessage` to avoid use-before-define.
- Semantic caveat (documented): for one-shot/container the answer is a *message*, not a protocol tool_result, so the tool_use stays unanswered in the transcript (harmless — the turn already ended). Persistent agents remain on the true tool_result path.

### DIFF
--- a/.changesets/1781547328-feat-agent-askq-phase2-oneshot.md
+++ b/.changesets/1781547328-feat-agent-askq-phase2-oneshot.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): deliver AskUserQuestion answers for one-shot/container agents as a follow-up turn
--- a/docs/specs/SPEC_ASK_USER_QUESTION_2026_06_15.md
+++ b/docs/specs/SPEC_ASK_USER_QUESTION_2026_06_15.md
@@ -1,7 +1,9 @@
 # SPEC: AskUserQuestion — interactive agent questions in the agent pane
 
 **Date:** 2026-06-15
-**Status:** Validated — implementing Phase 1 (smoke test passed 2026-06-15; see §10)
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1445, opened 2026-06-15
fix(agent): answered AskUserQuestion no longer re-surfaces on reload (Phase 3)

a5af:
## What

Phase 3 of AskUserQuestion: fix a real reload bug. On history reload, an **answered** `AskUserQuestion` re-parsed as `awaiting_answer` and the question panel **re-opened** for a question that was already answered (the stream transcript has the tool_use but no following tool_result, since the answer was delivered out-of-band via stdin or a follow-up turn).

## Fix

Extend the agent-document orphan-scrub (`scrubOrphanedInProgress`, which already runs on `SessionEnd` / `HistoryLoaded` / `HistoryRestored`):

- An `awaiting_answer` tool node **with content after it** → the conversation continued past the question, so it was answered → resolve to `success` and clear `question`. The panel no longer re-surfaces.
- A **tail** `awaiting_answer` is **left untouched** — deliberately. A one-shot agent's turn ends (`SessionEnd` fires) with the question as the tail node, and it's still answerable via a follow-up resume turn (Phase 2). Scrubbing it would kill the panel before the user could answer. This reuses the exact last-node heuristic the existing thinking-node scrub uses.

## Why this matters / the subtle part

The naive version (resolve all `awaiting_answer` on scrub) would have broken Phase 2: a one-shot agent legitimately has the question as the tail when `SessionEnd` fires, so it must survive the scrub. Only questions with *content after them* are provably answered.

## Tests

Two new reducer tests: answered (content after) → `success` + `question` cleared; tail-pending → untouched (no allocation, no events). `node_modules` isn't present in this checkout so I couldn't run vitest locally — the logic mirrors the adjacent running-tool / thinking-node scrub tests.

Spec §9 Phase 3 updated (also records why structured-JSON and cross-provider parity are not pursued — flat text is validated and `AskUserQuestion` is Claude-specific).

### DIFF
--- a/.changesets/1781549878-fix-agent-askq-reload-dedup.md
+++ b/.changesets/1781549878-fix-agent-askq-reload-dedup.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): answered AskUserQuestion no longer re-surfaces its panel on history reload
--- a/docs/specs/SPEC_ASK_USER_QUESTION_2026_06_15.md
+++ b/docs/specs/SPEC_ASK_USER_QUESTION_2026_06_15.md
@@ -385,10 +385,21 @@ update in 7.2 (and/or a matching `tool_result` echo if the provider emits one). 
   the true tool_result path (Phase 1).
 
 **Phase 3 — refinements.**
-- Structured (JSON) tool_result instead of the flat text block, if the model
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

Issue #1190, opened 2026-05-30
Browser pane: wire keyboard shortcuts (Ctrl+L / Ctrl+T / Ctrl+W / Ctrl+F) — input-first Phase 0.3


AgentY-asaf:
## Context

Phase 0.3 "cheap verified win" from [discussion #1161](https://github.com/agentmuxai/agentmux/discussions/1161): *"the browser pane has no keydown handler (Ctrl+L/T/F dead)."*

## Why they're dead (architecture)

Browser panes are **native CEF child windows** (not OSR). When a browser pane is focused, keystrokes go to the **CEF browser**, not to the host WebView running the SolidJS UI, so the app-level dispatcher (`appHandleKeyDown` at `frontend/app/store/keymodel.ts:369`) never sees them — standard chrome accelerators are swallowed.

Note the frontend address-bar plumbing **already exists**: `frontend/app/view/browser/browser-view.tsx` has `addressInputRef`, `addressBar` signal, and `handleAddressKeyDown` (`browser-view.tsx:242`). What's missing is getting the accelerator from the focused CEF pane to that code.

## Proposed fix

1. **Rust** — in the CEF keyboard handler (`agentmux-cef/src/client/handlers.rs`; browser-pane plumbing in the `agentmux-cef/src/browser_pane/` module) match the reserved accelerators in `OnPreKeyEvent`; on hit, send an IPC event to the frontend (`browser.shortcut` with `{ paneId, action }`) and return handled so CEF doesn't forward them to the page.
2. **Frontend** (`frontend/app/view/browser/browser-view.tsx`) — handle `browser.shortcut`:
   - `Ctrl+L` → focus + select `addressInputRef` (wiring already present)
   - `Ctrl+T` → new tab · `Ctrl+W` → close tab *(depends on browser-pane tabs — `specs/SPEC_BROWSER_PANE_TABS_2026-05-27.md`)*
   - `Ctrl+F` → in-page find (CEF `Find` API) or the app find overlay
   - *(optional)* `Ctrl+R` reload, `Alt+←/→` back/forward

## Acceptance
- With a browser pane focused, each shortcut performs its action; none leak to the page.
- No regression to typing into web-page form fields (only reserved chrome accelerators intercepted).
- Cross-platform: same set with `Cmd` on macOS.

## Sequencing
[...]

Issue #768, opened 2026-05-09
Phantom browser pane (orphan + tearoff): host/frontend lifecycle divergence


AgentA-asaf:
## Summary

The frontend browser-pane slot store and the host CEF browser registry can diverge under two paths, both observed in dev session 2026-05-09:

### 1. Orphan (renderer/host-cleanup)
- Host emits `BrowserUnregistered` (renderer crash, GPU loss, or other host-side cleanup).
- Frontend `BrowserViewModel` is unaware → keeps dispatching `Navigate` against a dead slot.
- User sees a **black pane** that quietly drops typed URLs.
- Closing/reopening the pane recovers.

Concrete log evidence from this session:
```
15:36:18  host-reducer: event=BrowserUnregistered label=browser-pane-<bid>-1
15:36:18  Unregistered browser: label=browser-pane-<bid>-1 (remaining: 3)
15:40:26  [fe] [browser-pane:diag][<bid>] navigate(url=…) closed=false
15:40:26  [fe] [browser-pane:diag][<bid>] dispatch type=Navigate src=navigate
   …       (no nav-state event ever returns; the CEF browser is gone)
```

### 2. Tearoff regression
- Tear off a tab containing 2+ browser panes.
- New window shows **2 black browsers**.
- DOMs from the source window **flicker at stale screen coords** atop the first window.
- Same root: host reparented the HWNDs but frontend never re-handshook against the new chrome.

## Root cause

Three sources of truth on browser-pane lifecycle:

1. Host `BrowserPaneManager` registry (actual CEF state)
2. Host-reducer state (`BrowserRegistered` / `BrowserUnregistered` events)
3. Frontend slice #9 slot store (`browser-pane-state-store.ts`)

(1)/(2) and (3) drift on every code path that doesn't go through the layout's block-removal saga.

## Fix design (high level)

Two coordinated host→frontend lifecycle events plus a recovery overlay:

- **`browser-pane-unregistered`** — host emits with `{ block_id, label, reason }`. Frontend listens, calls `BrowserViewModel.dispose()` (idempotent), so the slot store flips `closed = true` and post-close gating engages.
- **`browser-pane-reparented`** — host emits per pane on tear-off completion with `{ block_id, new_window_id, new_hwnd_parent }`. Frontend re-runs the host-attach handshake.
- **Browser disconnected overlay** — view shows `⚠ Browser disconnected (<reason>) — close pane to reset` with Close / New buttons, so users see the failure and the recovery action.

[...]

PR #1446, opened 2026-06-15
fix(browser-pane): focus handoff on pane HWND destroy (global typing-lock)

a5af:
## Summary

Fixes the **global typing lock** after redocking (or closing) a browser pane — the bug where every pane (agent, terminal, even the address bar) stops accepting keystrokes and the only recovery is to "open another window and come back." This is **Defect A** from the triage doc included here.

### Root cause
Browser panes are native CEF child windows (HWNDs). On redock/close, `destroy_hwnd()` hides + destroys the pane HWND and repaints the parent, **but never hands Win32 keyboard focus back**. If the pane held focus — the common case on redock, since the dragged pane is focused — focus is orphaned on a dead HWND and the OS has nowhere to route keystrokes, so typing dies app-wide. Switching windows fires `main_window_focus → MainFocusReclaimTask`, the only path that re-seats focus on the main render widget; that's the manual workaround. This PR makes the reclaim automatic.

### Changes
- **`destroy_hwnd`** — `hwnd::forget_focus_for_child()` drops the `LAST_FOCUSED_BY_ROOT` entry for the dead child, so the `WM_ACTIVATE` restore hook never targets it.
- **`close()` + `drain_closed_label()`** — after the pane HWND is destroyed, post `MainFocusReclaimTask` to return Win32 focus to the surviving window's main render widget (which hosts the entire DOM UI — agent/terminal panes included) and `defocus_all()` native panes.
- **`MainFocusReclaimTask`** — an empty label now resolves the **foreground** agentmux window (`GetForegroundWindow` → `window_hwnds` reverse-lookup, fallback `"main"`), correct for both redock (floater gone → target window foreground) and in-window close.

Windows-only — the orphaning is a Win32 native-child-window issue.

### Triage doc
`docs/analysis/ANALYSIS_BROWSER_PANE_REDOCK_BLACK_TYPING_LOCK_2026_06_15.md` — full analysis. **Defect B** (black page / empty URL on redock — slot re-init to `url:""` + deferred re-create not re-navigating) and **Defect C** (reducer divergence: `MoveBlock` doesn't clear focus or sync slice #9) are separate follow-ups.

## Verification
- Manual (Windows): float a loaded browser pane, redock it, confirm typing works immediately in the redocked pane *and* sibling agent/terminal panes without opening a new window. (Win32 focus path can't be unit-tested headless.)

Refs #768, #1190.

### DIFF
--- a/.changesets/1781540400-fix-browser-pane-redock-focus.md
+++ b/.changesets/1781540400-fix-browser-pane-redock-focus.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(browser-pane): hand keyboard focus back when a pane HWND is destroyed — redocking/closing a browser pane no longer locks typing app-wide (no more "open another window to get typing back")
--- a/agentmux-cef/src/browser_pane/hwnd.rs
+++ b/agentmux-cef/src/browser_pane/hwnd.rs
@@ -121,6 +121,23 @@ pub unsafe fn record_intentional_focus(child: *mut std::ffi::c_void) {
     }
 }
 
+/// Forget any `LAST_FOCUSED_BY_ROOT` entry whose recorded child is `child`.
+/// Called when a pane HWND is destroyed so the `WM_ACTIVATE` focus-restore
+/// hook never tries to `SetFocus` a dead child, and the destroyed pane isn't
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1448, opened 2026-06-15
diag(browser-pane): instrument the redock create path (black-page triage)

a5af:
## Summary

Diagnostics-only instrumentation to pin down the **black-page-on-redock** symptom (Defect B in `docs/analysis/ANALYSIS_BROWSER_PANE_REDOCK_BLACK_TYPING_LOCK_2026_06_15.md`). The exact cause can't be reproduced headless, so this makes the next live repro conclusive instead of guessing — and it surfaces a previously-invisible race.

- **`AlreadyLive` arm was completely silent.** It's the prime suspect: if a redock's `browser_pane_create` races *ahead* of the `browser_pane_close` (old pane still `Live`), the host only re-navigates the existing browser — which is still parented to the dying floater window — and **never creates a pane in the requested window**, so the target shows black. Now logs `label` + `requested_window` + `rect` (both `tracing` and `pane_trace`).
- **`Closing` + `Fresh` arms** now also log `requested_window` + `rect`, so a zero/garbage-bounds pane (the other black-page suspect — an invisible HWND) is visible too.

No behavior change.

## How to use
Reproduce the redock → black pane, then check the host log for the `block_id`:
- A `create-already-live` / "hit AlreadyLive" line on redock → **confirmed cause** (create raced the close); fix = recreate-in-target on window-label mismatch.
- A zero/garbage `rect=(…)` → invisible-bounds cause.
- Only `create-deferred-closing` then a successful replay → load race already handled; black is elsewhere (stale GPU surface).

## Verification

Refs #768.

### DIFF
--- a/.changesets/1781544000-diag-browser-pane-redock-create.md
+++ b/.changesets/1781544000-diag-browser-pane-redock-create.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+diag(browser-pane): log the create register-result (Fresh/Closing/AlreadyLive), requested window, and rect on the redock path — surfaces the previously-silent AlreadyLive race so the black-page-on-redock cause is conclusive in a repro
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1449, opened 2026-06-15
fix(host): eliminate ghost taskbar icon from pre-warmed pool window

a5af:
## Root Cause

With 2 instances open, 3 taskbar icons appeared. The third was a **pre-warmed pool window** (tear-off pool, `POOL_TARGET_SIZE=2`) positioned off-screen at `-32000,-32000` that never got `WS_EX_TOOLWINDOW` applied.

**Failure chain:**

1. `spawn_pool_window` creates a pool window via `post_create_window` → `window_create_top_level`
2. `on_after_created` fires → `register_pool_window` calls `BrowserHost::window_handle()` to get the HWND
3. `BrowserHost::window_handle()` returns **null** after the page loads — this is a documented CEF behaviour, diagnosed in the 2026-05-06 run and covered by the `POOL_HWND_CACHE` comment in `window_pool.rs`
4. `set_taskbar_hidden` is skipped (line 304-310) with the warning: `"HWND null at register time — taskbar hide skipped"`
5. Pool window has no `WS_EX_TOOLWINDOW` → Windows shows a taskbar button for the off-screen blank window

## Fix

Add `init_pool_window_hwnd()` to `window_pool.rs`, called from `AgentMuxWindowDelegate::on_window_created` in `app.rs`.

`CefWindow::GetWindowHandle()` (available in `on_window_created`) is reliably non-null — it's the same underlying Win32 HWND but accessed through the CEF Views Window object rather than `BrowserHost`. This fires **before** the page loads, so before the CEF bug that nullifies `BrowserHost::window_handle()`.

As a bonus, `init_pool_window_hwnd` also **pre-populates the HWND cache** — meaning `promote_pool_window` now has a guaranteed cache hit even on instances where `register_pool_window` encountered a null HWND.

`register_pool_window` (the `on_after_created` path) remains untouched and continues as belt-and-suspenders; its `set_taskbar_hidden` call is now a no-op because the style was already applied.

## Changes

- `agentmux-cef/src/commands/window_pool.rs` — new `pub fn init_pool_window_hwnd`  
- `agentmux-cef/src/app.rs` — call it in `on_window_created` for `window-pool-*` labels on Windows

## Test plan

- [ ] Open 2 instances — confirm exactly 2 taskbar icons, no ghost third icon
- [ ] Check `muxlog host '[pool]'` — should see `"HWND cached + taskbar hidden at on_window_created"` at startup, NOT the `"HWND null at register time"` warning
- [ ] Tear off a tab → confirm the promoted window appears in the taskbar normally (`set_taskbar_hidden(false)` still runs in `promote_pool_window`)
- [ ] Pool refills after promotion — confirm the refill pool window is also hidden from taskbar

### DIFF
--- a/.changesets/1781501000-fix-host-pool-window-ghost-taskbar-icon-hwnd-early-hide-k7m9.md
+++ b/.changesets/1781501000-fix-host-pool-window-ghost-taskbar-icon-hwnd-early-hide-k7m9.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+fix(host): eliminate ghost taskbar icon from pre-warmed pool window — apply WS_EX_TOOLWINDOW at on_window_created (reliable HWND) instead of on_after_created where BrowserHost::window_handle() can return null after page load
--- a/agentmux-cef/src/app.rs
+++ b/agentmux-cef/src/app.rs
@@ -63,6 +63,24 @@ wrap_window_delegate! {
                 window.set_bounds(Some(&Rect { x, y, width: w, height: h }));
             }
 
+            // Windows: hide pool windows from the taskbar at creation time.
+            // CefWindow::GetWindowHandle() is reliable here; BrowserHost::
+            // window_handle() (used in register_pool_window / on_after_created)
+            // can return null after the page loads — see window_pool.rs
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1450, opened 2026-06-15
fix(agent-pane): persistent shell hung on inherited stdin (dev servers now start)

AgentX-asaf:
## Problem

The persistent Shell tool / activity dock **could not run dev servers** — the headline use case. `Shell("npm run dev")` (or `task dev`, vite, any watcher) produced **no output, never bound its port, and never exited**. The shell row just sat there dead.

## Root cause

`ShellNodeRunner` set `stdout`/`stderr` to piped but **never set `stdin`** — so the spawned `cmd /C …` inherited the **srv's stdin**, which is not a usable TTY. Tools that probe/read stdin at startup **hang before doing any work**.

Confirmed with a process-tree trace of a stuck shell:
```
cmd.exe → node.exe (npm-cli.js)   ← stuck here; no vite grandchild, no output, never exits
```
npm spawned but never launched the `dev` script. The identical command with **`< NUL`** starts vite **instantly** — isolating stdin as the cause.

This silently broke the whole long-running-process feature for npm/node dev servers, and very likely is *why* agents fall back to backgrounded `Bash` for dev servers (the Shell tool just hangs for them).

## Fix

One line: give the spawned shell a **null stdin**.
```rust
child_cmd.stdin(std::process::Stdio::null());
```
(When `ShellInput` lands — Phase 3b — this becomes a piped stdin.)

Also adds concise `shell.spawn` / `shell.exit` tracing (pid, exit_code, line_count) for future observability.

## Verified end-to-end

Rebuilt the backend, fired a **plain** `Shell("npm run dev")` (no `< NUL` workaround) against the landing-page repo → **HTTP 200 on :4600**, serving `<title>AgentMux: Agentic Workflow Environment</title>`. `cargo check` clean.

Spec: `SPEC_PERSISTENT_SHELL_PHASE3_STOP` / the long-running-shell dock track.

### DIFF
--- a/.changesets/1781562782-fix-agent-pane-persistent-shell-hung-on-inherited-stdin-npm--ieim.md
+++ b/.changesets/1781562782-fix-agent-pane-persistent-shell-hung-on-inherited-stdin-npm--ieim.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): persistent shell hung on inherited stdin — npm/vite dev servers now start (null stdin)
--- a/agentmux-srv/src/backend/shell_node.rs
+++ b/agentmux-srv/src/backend/shell_node.rs
@@ -158,6 +158,14 @@ impl ShellNodeRunner {
 
         child_cmd.stdout(std::process::Stdio::piped());
         child_cmd.stderr(std::process::Stdio::piped());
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1451, opened 2026-06-15
feat(agent): answer AskUserQuestion via the Agent SDK control protocol

AgentU-asaf:
## Problem

Calling **AskUserQuestion** from a Claude agent pane renders no question panel — the agent instantly gets `Error: Answer questions?` back and continues.

**Root cause (confirmed in a live v0.46.0 srv log, block `27b6f287`, model `claude-opus-4-8`):** the Claude provider defaults to the **subprocess (`-p`) controller**. In `-p`/print mode the Claude CLI has no interactive stdin, so it **auto-rejects its own AskUserQuestion tool_use** ~7ms after emitting it:

```
22:17:52.361  assistant tool_use  name=AskUserQuestion  input={questions:[…]}   ✅ valid
22:17:52.368  user tool_result    content="Answer questions?"  is_error=true     ❌ +7ms, from the CLI itself
```

The frontend node flips `awaiting_answer → failed` before first paint, so `pendingQuestions()` never catches it → no panel.

AskUserQuestion (`SPEC_ASK_USER_QUESTION_2026_06_15.md`, #1441/#1444/#1445) was designed + validated **only for the persistent controller** (`--input-format stream-json`, bidirectional), where the turn genuinely **blocks** on the tool_use and consumes a `tool_result` over live stdin. Phase 2's subprocess fallback can't engage either, because the CLI's auto-error resolves the tool mid-turn.

## Fix

Flip Claude to `ControllerType::Persistent` in **both** provider configs:
- `agentmux-srv/src/backend/providers.rs` (`static CLAUDE`)
- `frontend/app/view/agent/providers/index.ts` (seeds the block's `controller` meta at agent creation)

`persistent_launch_args` were already defined in both; `controllerType` selects them automatically.

## Why this is safe now

#406 originally switched Claude `Persistent → Subprocess` to dodge a **Windows anonymous-pipe "no response" hang in the then-Node.js CLI**. The Claude CLI is now a **native binary (v2.1.x)** — both the user's standalone install and the npm-bundled `@anthropic-ai/claude-code` (which now just pulls `claude-code-linux-x64`). That Node-process root cause is gone, and the spec's §10 smoke test already confirmed persistent mode streams correctly against the native CLI on Windows.

## Tests

- `agentmux-srv` provider tests: **11/11** pass (incl. updated `claude_persistent_with_persistent_args_present`)
- frontend provider tests: **25/25** pass

## Caveats for reviewers

- Persistent hasn't been Claude's default since #406 — runtime behavior (multi-turn, resume, stop) should get a manual smoke test, not just unit tests.
- **Existing** Claude agents keep `controller: subprocess` in their persisted block meta; only **newly created** Claude agents get persistent. A meta migration for existing agents is out of scope here.
- Re-validating persistent on Windows against the *npm-bundled* native binary (vs the standalone the spec tested) is recommended before relying on it there.

### DIFF
--- a/.changesets/1781563278-fix-agent-run-claude-on-persistent-controller-so-askuserques-cd4x.md
+++ b/.changesets/1781563278-fix-agent-run-claude-on-persistent-controller-so-askuserques-cd4x.md
@@ -1,5 +1,12 @@
 ---
-type: patch
+type: minor
 ---
 
-fix(agent): run Claude on persistent controller so AskUserQuestion works
+feat(agent): answer AskUserQuestion via the Agent SDK control protocol
+
+Claude now runs on the persistent controller with `--permission-prompt-tool stdio`
+and the sidecar answers the CLI's `can_use_tool` control_request with a
+control_response (`updatedInput.answers`). This is the only mechanism the CLI
+accepts for AskUserQuestion — delivering a tool_result on stdin does not work
+(the CLI auto-rejects it within the turn). Ordinary tools are auto-allowed to
+preserve current behavior. Spec: docs/specs/SPEC_AGENT_CONTROL_PROTOCOL_2026_06_15.md.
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

PR #1453, opened 2026-06-15
fix(ui): scrollbars use the arrow cursor, not the link hand

a5af:
## What

Scrollbars (notably on **agent panes**) showed the link **hand** (`cursor: pointer`) instead of the default **arrow**. A scrollbar is a scroll affordance, not a hyperlink — it should be the arrow. This is **Phase 0** of the cursor audit.

## Root cause — two global lines in `app.scss`

- `*::-webkit-scrollbar-thumb { cursor: pointer }` — every Chromium scrollbar.
- `.os-scrollbar-handle { cursor: pointer }` — **OverlayScrollbars**, which is what the agent conversation/markdown surfaces use (`markdown.tsx` → `<Markdown scrollable>`). That's why agent panes were the obvious victim.

## Fix

- Removed the thumb's `cursor` override → it inherits the default arrow.
- Set the OverlayScrollbars handle to `cursor: default`.

No spec/changeset ever justified the hand; it was a hard-coded default that rode along while the scrollbar *colors* got tokenized.

## Follow-up (proposed, not in this PR)

The full audit + a clean cursor-token/mixin refactor (Phase 1: semantic `--cursor-*` tokens + `_cursor.scss`; Phase 2: a stylelint/CI guard forbidding `cursor:` on scrollbar selectors so this bug class can't recur) is written up in `docs/analysis/ANALYSIS_CURSOR_STYLING_2026_06_15.md`. I'll ship that as a small follow-up PR.

### DIFF
--- a/.changesets/1781564876-fix-scrollbar-cursor-arrow.md
+++ b/.changesets/1781564876-fix-scrollbar-cursor-arrow.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ui): scrollbars use the default arrow cursor, not the link hand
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1455, opened 2026-06-15
chore(ui): cursor design tokens + utilities + scrollbar lint guard (Phase 1+2)

a5af:
## What

Phase 1+2 of the cursor audit (`docs/analysis/ANALYSIS_CURSOR_STYLING_2026_06_15.md`). Phase 0 (#1453) already fixed the scrollbar hand; this generalizes cursor styling and **prevents the bug class from recurring**.

Cursors were the one interaction primitive never tokenized — **229** raw `cursor:` declarations, no tokens/mixins/utilities — while spacing, type, radius, motion, and z-index all are. That asymmetry is exactly why a wrong global default (link-hand on scrollbars) could be set once and go unnoticed.

## Phase 1 — generalization

- **`theme.scss`**: semantic `--cursor-*` tokens (`interactive`, `default`, `text`, `disabled`, `grab`, `grabbing`, `resize-x/-y`). Name the intent, not the keyword.
- **`app.scss`**: `.u-cursor-*` utility classes resolving to the tokens (for markup that can't easily reach a component stylesheet; SCSS authors can also write `cursor: var(--cursor-interactive)`). `.os-scrollbar-handle` now reads `cursor: var(--cursor-default)`.

## Phase 2 — guardrail

- **`.stylelintrc.json`**: `rule-selector-property-disallowed-list` forbids the `cursor` property on any `::-webkit-scrollbar*` selector, with a message pointing at the analysis doc. This models the existing `z-index` allowed-list and `color-no-hex` design-system enforcement. After Phase 0 nothing sets cursor on those selectors, so it passes today and fails the moment someone re-adds the hand.

## Scope / not done

The 229-site migration to the utilities is **deliberately not** in this PR — it's opportunistic per-area hygiene (agent pane first), not a big-bang rewrite. Rationale in the analysis doc's phasing table.

## Verification

`node_modules` isn't present in this checkout, so I couldn't run `stylelint` locally — but the JSON validates, the rule name/shape match stylelint 17 core, and I scanned the tree to confirm no remaining `cursor` declaration sits inside a `::-webkit-scrollbar*` rule (the lone hit was the explanatory *comment* in the thumb rule, which stylelint doesn't evaluate).

### DIFF
--- a/.changesets/1781565397-chore-cursor-tokens-and-guard.md
+++ b/.changesets/1781565397-chore-cursor-tokens-and-guard.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+chore(ui): add cursor design tokens + utilities and a stylelint guard against cursor on scrollbars
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -12,6 +12,14 @@
                 "auto", "unset", "inherit", "initial"
             ]
         },
+        "rule-selector-property-disallowed-list": [
+            {
+                "/-webkit-scrollbar/": ["cursor"]
+            },
+            {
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1456, opened 2026-06-15
fix(activity-dock): higher-contrast pinned rows + larger stop/dismiss buttons

AgentX-asaf:
## What

Polish on the pinned activity dock (`.agent-activity-*`):

- **Higher-contrast rows** — title is now full-strength `--main-text-color` (weight 600), the live tail is monospace at 0.9 opacity (was dim `--secondary-text-color` at 0.75), elapsed brightened. The dock sits on an elevated surface with an accent-tinted 2px divider + drop shadow so it reads as distinct from the transcript instead of blending in.
- **Larger controls** — stop (■) is a 32×32 button with a 22px glyph; dismiss (×) glyph scaled to 34px bold for an obvious target. Both `align-self: center` so the taller glyph centers in the row instead of hanging off the text baseline (the row uses `align-items: baseline` for the text).

## Scope

CSS-only, scoped entirely to `.agent-activity-*`. No `overflow`/scroll-container/global-scrollbar rules touched.

## Test

Hot-reloads in `task dev`; verified the contrast + button sizing + alignment live in the 0.46.0 dev instance.

### DIFF
--- a/.changesets/1781565587-fix-activity-dock-higher-contrast-pinned-rows-larger-stop-di-rnqu.md
+++ b/.changesets/1781565587-fix-activity-dock-higher-contrast-pinned-rows-larger-stop-di-rnqu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(activity-dock): higher-contrast pinned rows + larger stop/dismiss buttons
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1454, opened 2026-06-15
fix(toolchain): enrich GUI-launch PATH so node/npm/git resolve (P0)

a5af:
## What

Root-cause fix for the **"NPM failed"** install error. When AgentMux launches from Finder / Dock / a mounted DMG as a `.app`, it inherits launchd's stripped PATH (`/usr/bin:/bin:/usr/sbin:/sbin`). nvm/Homebrew `node`/`npm`/`git` live in the user's **login-shell** PATH, which a GUI launch never sees — so `install.start`'s `Command::new("npm")` died with `npm: command not found`, and spawned agent CLIs couldn't resolve their interpreter. This hits the majority of macOS users (nvm/Homebrew Node) — the central "do users have the right prereqs" risk for the alpha.

**P0** of `docs/specs/SPEC_TOOLCHAIN_MANAGER_2026-06-15.md` (the Toolchain Manager modal — P1–P3 — builds on this).

## How

- **`agentmux-common::toolchain_path`** — `resolve_login_path()` reconstructs a usable PATH from `$SHELL -lic 'printf … "$PATH"'` (2s hard timeout; sentinel-isolated parse so rc-file stdout banners can't corrupt it) unioned with existence-checked well-known dirs (nvm current/default, Homebrew, `/usr/local/bin`, `~/.local/bin`, `~/.cargo/bin`; +Snap/Flatpak on Linux). **Purely additive** — never drops an inherited system dir. **No-op on Windows** (`#[cfg(not(unix))]` early return — Windows GUI apps already inherit a usable PATH). Returns a `PathSource` (`login-shell` / `fallback-dirs` / `inherited`) for diagnostics + the future modal.
- **host `sidecar.rs`** — sets the enriched PATH on the srv it spawns (logs the source).
- **srv `main.rs`** — cheap direct-launch fallback: enriches only when the current PATH looks like the stripped launchd default (won't pay for a shell spawn when already healthy).

## Cross-platform

| OS | Behavior |
|----|----------|
| macOS | Enriches from login shell + Homebrew/nvm dirs |
| Linux | Same + Snap/Flatpak export dirs |
| Windows | No-op (GUI inherits usable PATH) |

## Tests

[...]

PR #1457, opened 2026-06-15
feat(toolchain): Toolchain manager modal (hamburger ▸ Toolchain) — P1

a5af:
## What

A first-class **Toolchain** surface (hamburger ▸ Toolchain) so users can see — and diagnose — the dev toolchain AgentMux runs CLIs in. Builds on the P0 PATH fix (#1454); this is **P1** of `SPEC_TOOLCHAIN_MANAGER_2026-06-15` (read-only).

A window-scoped modal with three sections:
- **Environment** — the *effective PATH* the srv resolves tools in + a **source badge** (`login-shell` / `fallback-dirs` / `inherited`), OS/arch, and a Show-PATH toggle. This makes the GUI-launch PATH problem (the "NPM failed" class) visible instead of mysterious.
- **Core tools** — node / npm / git / docker: detected version + resolved path + status pill.
- **Agent CLIs** — every provider: version + source (`local_install` vs `system_path`).

Missing tools get an **Install** link (+ copyable command). Docker is marked optional, so its absence reads as info, not a warning.

## Design / decomposition

- **Reuses `resolvecli`** per row (it already does versioned-dir → PATH → `--version` for any tool, incl. non-npm like docker) — no new batch resolver, lower risk.
- Adds only a tiny **`toolchain.env`** RPC → `{ path, pathSource, os, arch }`. `pathSource` comes from `AGENTMUX_PATH_SOURCE`, set by the host (P0) when it enriches the srv PATH (srv also sets it on the direct-launch fallback).
- New `toolchain-catalog.ts` is the single place to add more tools later (ripgrep, uv/python for kimi…).

## Cross-platform

Probes use `which`/`where`; the modal is platform-agnostic; the PATH-source badge reflects P0's per-OS enrichment (no-op on Windows, where the badge reads "inherited" as muted, not a warning).

## Scope

Read-only by design. **Install/repair in place** (reuse `install.start`) is **P2**; **one-click `brew`** is **P3** — both tracked in the spec.

## Verification

- `cargo check` clean on `agentmux-srv` + `agentmux-cef`; `cargo test -p agentmux-common` (P0) green.
- `tsc --noEmit`: zero new errors from the added files (toolchain-modal, toolchain-catalog, rpc-api, hamburger-menu).

### DIFF
--- a/.changesets/1781565739-feat-toolchain-manager-modal.md
+++ b/.changesets/1781565739-feat-toolchain-manager-modal.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(toolchain): Toolchain manager in the hamburger menu — a new "Toolchain" item opens a modal showing the effective PATH (and how it was derived), OS/arch, and the detected version + path + status of node, npm, git, docker and every provider CLI, with install links for anything missing. P1 of SPEC_TOOLCHAIN_MANAGER (read-only; install-in-place is P2).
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

PR #1462, opened 2026-06-16
feat(agent): shared PaneRow auxiliary-pin primitive (forks Phase 1)

a5af:
## What

**Phase 1** of `SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15.md` — the shared **auxiliary-pin row primitive**, `<PaneRow>`.

One status-accented row — *sigil + title + meta + tail + actions + optional inline-expanded body* — that the pinned **ActivityDock** rows use and the forthcoming **fork bar** (bottom-of-pane conversation switcher) will reuse. The goal is that every pin-shaped surface shares **one** look, **one** status-accent palette, and **one** interaction model, instead of the current bespoke-per-surface markup the audit found.

- **Presentational only**: owns no data, derives nothing — the "derive pins from a source of truth" rule stays in the caller (the spec's §5.2/§5.3 best-practice).
- **Token-driven SCSS** (no raw hex; passes the design-system stylelint rules), interactive cursor via the cursor tokens shipped earlier.

## Why land it ahead of consumers

Per the spec's phasing, Phase 1 is the foundation. The **immediate next consumers** are (a) migrating `ActivityDock`/`ActivityRow` onto `<PaneRow>` and (b) the Phase-2 fork bar. It's landed now, fully unit-tested, so those build on a verified base. It is intentionally not yet wired into the render tree — that migration is pixel-sensitive and gets its own PR so the just-shipped dock isn't refactored blind.

## Also in this PR

The umbrella spec, including the **§6.4 `--fork-session` validation gate** — which I ran empirically against the bundled `claude` and **passed**: a fork inherits the parent's history, gets a **new** session id, leaves the **parent intact**, and two forks are **independent**. That de-risks the Phase-3 fork mechanism before any of it is built.

## Verification (toolchain now available locally)

- **vitest 8/8** — render (sigil/title/meta/tail), accent modifier, default neutral, `onActivate` on summary click, action handler fires with `stopPropagation` (no double-activate), expanded body slot, expanded modifier.
- **stylelint** — clean on `PaneRow.scss`.
- **tsc** — clean (no errors in the new files).

## Files

- `frontend/app/view/agent/components/PaneRow.tsx` (new)
- `frontend/app/view/agent/components/PaneRow.scss` (new)
- `frontend/app/view/agent/components/PaneRow.test.tsx` (new)
- `docs/specs/SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15.md` (new — the umbrella design + validation results)

### DIFF
--- a/.changesets/1781568779-feat-agent-pane-row-primitive.md
+++ b/.changesets/1781568779-feat-agent-pane-row-primitive.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): shared PaneRow auxiliary-pin primitive (forks Phase 1)
--- a/docs/specs/SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15.md
+++ b/docs/specs/SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15.md
@@ -1,0 +1,401 @@
+# SPEC: Agent pane forks + a cohesive auxiliary-pins architecture
+
+**Date:** 2026-06-15
+**Status:** Draft — architecture + best-practices (no code landed)
+**Scope:** Agent pane (`frontend/app/view/agent/**`), block/layout model, agent runtime
+**Related:** `SPEC_LONG_RUNNING_SHELL_PINNED_DOCK_2026_06_15.md` (the top dock that just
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1463, opened 2026-06-16
feat(muxlog): cross-instance log discovery, NDJSON rendering, filters & recipes

AgentX-asaf:
## Why

`muxlog` was a version-pinned pointer tail — host/srv only, raw NDJSON, single instance. In a real debugging session (a startup reconnect loop) it actively got in the way: it resolved a **6-day-stale** instance's log, was blind to the `~/.agentmux/dev/<branch>/` and `channels/local-*` log trees, and grepping the sidecar hit agent-transcript noise. Every hunt began with a manual `find` + giant-JSON `grep`.

## What

A shared **Node core** (`muxlog.mjs`, deployed next to the shell rcfiles) that the bash/zsh/pwsh/fish `muxlog` functions delegate to — one tested implementation instead of four divergent shell ports.

- **Discovery across all three log roots** (shared · `dev/<branch>` · `channels/local-*`), ranked by mtime → defaults to the **most-recently-active** instance.
- **`muxlog ls`** — every instance's logs with version, source (`shared`/`dev:<branch>`/`channel:…`), age, size, path.
- **NDJSON rendering** — `time  level  target  message` (shortened tracing targets); `--raw` / `--verbose`.
- **Targets** host·srv·launcher·fe; **filters** `-i` (instance), `--grep` (message field only), `--level`, `--since`, `--target`; **agent-transcript noise excluded by default** (`-a` to include).
- **Recipes** — `muxlog errors` (host+srv WARN/ERROR), `muxlog bridge` (startup-handshake correlation, purpose-built for the reconnect-loop class of bug).
- **Fallback** — if `node` isn't on PATH, degrades to the legacy pointer tail so logs are never inaccessible.

## Docs (emphasis of this PR)

- README **Logs** section rewritten with the new command set.
- New **`docs/MUXLOG.md`** — full reference (targets, actions, options, recipes, how discovery works).
- `CLAUDE.md` Log Access table updated (what agents read).

## Test

`node --check` clean; smoke-tested against live logs on Windows — `ls` (discovers shared+dev+channel logs newest-first), `bridge` (renders the exact handshake trace), `grep`/`cat`/filters/rendering all verified. Default `tail` follows like `tail -f`.

> Follow-ups (not in this PR): a combined time-interleaved `all` view; deploy redeploys only on version bump (`VERSION_MARKER` = cargo version) — a content-hash marker would redeploy on script edits too; and the docs-site (`docs.agentmux.ai`) page to mirror `docs/MUXLOG.md`.

### DIFF
--- a/.changesets/1781568947-feat-muxlog-cross-instance-log-discovery-ndjson-rendering-fi-ovsl.md
+++ b/.changesets/1781568947-feat-muxlog-cross-instance-log-discovery-ndjson-rendering-fi-ovsl.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxlog): cross-instance log discovery, NDJSON rendering, filters and recipes
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,18 +152,29 @@ These views exist in the codebase but are **not** widget-bar entries — do not 
 
 ## Log Access
 
-`$AGENTMUX_LOG_DIR` in AgentMux terminals is the shared `~/.agentmux/logs/` directory. The sidecar log lives there directly; the host log lives in the per-instance data dir and registers a pointer (`current-host-v<v>.path`) in the shared dir so `muxlog host` can resolve it.
+`muxlog` discovers and renders AgentMux logs across **every** running instance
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #1464, opened 2026-06-16
feat(agent-pane): surface classified failure cause on non-zero agent exit

AgentA-asaf:
Completes the half of [`SPEC_AGENT_FAILURE_DIAGNOSTICS`](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/SPEC_AGENT_FAILURE_DIAGNOSTICS_2026_06_11.md) that #1353 left undone: surfacing the **real cause** of a failed agent run **in the pane**.

## Problem
#1353 shipped the pure `failure::classify()` taxonomy but only wired it into the **headless** runner (`agents/runner.rs`). The interactive **pane** path (`blockcontroller/subprocess.rs` — what an agent like *Spixo* runs in) never classified anything: its stderr reader explicitly *"log warnings, don't publish"*, and it surfaced only the raw exit code → the pane shows an opaque **"exited with code 1."** (Spec §line 45, G2.)

## Change
**Backend (`subprocess.rs`):**
- The stderr reader now retains a bounded tail (last ~40 lines) in a shared buffer.
- On a **genuine non-zero exit** (the clean-exit arm; a user-initiated stop via the kill arm is intentionally *not* a failure), it runs `(exit_code, signal, stderr_tail)` through `agents::failure::classify` and publishes a dedicated **`agentfailure`** wave event scoped to the block.
- A dedicated event (rather than a field on `BlockControllerRuntimeStatus`) avoids touching ~20 status constructors and models a failure as an event, not recurring status.

**Frontend (`useControllerStatusEvents.ts` + `gotypes.d.ts`):**
- Subscribes to `agentfailure` and renders the real reason — `title — detail [exit N] (retryable)` + the stderr tail — as error log lines, instead of just the exit code.
- Adds the `AgentFailure` wire type.

## Scope
Covers the **main subprocess agent path** (Spixo's). The container-exec path and the `persistent`/`acp` controllers, plus a structured pane *banner* (spec P3) and the exit-0-error-frame case (G3), remain follow-ups.

## Verification
`task build:backend` ✓ (release, warnings only) · `task build:frontend` ✓. So an agent that dies on e.g. an auth/rate-limit error will now show *that*, not "exit 1" — which is also the fastest way to see why Spixo is dying.

### DIFF
--- a/.changesets/1781569327-feat-agent-pane-failure-wiring.md
+++ b/.changesets/1781569327-feat-agent-pane-failure-wiring.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): surface the classified failure cause when an agent exits non-zero (SPEC_AGENT_FAILURE_DIAGNOSTICS Phase 2 — pane path). The `SubprocessController` now captures a stderr tail, runs it through `failure::classify`, and emits an `agentfailure` event; the pane shows the real reason (auth, rate-limit, OOM, context, …) + stderr tail instead of a bare "exited with code N".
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1465, opened 2026-06-16
feat(agent): fork-set derivation — data model for the fork bar (forks Phase 2)

a5af:
## What

The **pure data model** behind the bottom **fork bar** (`SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15.md` §6): `computeForkSet(definitions, openBlockByDef, activeDefinitionId)`.

Given the agent-definition list (lineage via `parent_id`) and which definitions are currently open as panes, it derives the active conversation's lineage **root** plus every **fork** descended from it, in the order the fork bar renders them (root first, then breadth-first, oldest-sibling-first — reads like a timeline of branches).

It's a **pure function of a source of truth** — no parallel state — per the spec's §5.3 "derive pins from a source of truth" rule. The fork bar (next phase) renders these via `<PaneRow>`; the in-pane block-stack swap wires each entry's `blockId` to the active view.

## Edge cases covered (and tested)

- Deep lineages → walks `parent_id` up to the true root.
- **Orphaned fork** (parent definition deleted) → becomes its own root.
- Blank/whitespace `branch_label` → falls back to the definition `name`.
- `openBlockByDef` attaches a `blockId` only to forks that are actually open.
- **Corrupt `parent_id` cycle** → terminates (visited-set + walk cap), no duplicates.
- Unknown active id → `[]` (caller renders no bar).
- Exactly one entry flagged `isActive`.

## Verification

- **vitest 9/9** (the cases above).
- **tsc** clean.

## Scope

Pure logic only — lands ahead of its consumer (the fork bar component) per the spec's phasing, so the UI builds on a tested data model. No production wiring yet (that needs the Phase-2 block-stack swap).

## Files

- `frontend/app/view/agent/fork/fork-set.ts` (new)
- `frontend/app/view/agent/fork/fork-set.test.ts` (new)

### DIFF
--- a/.changesets/1781569700-feat-agent-fork-set.md
+++ b/.changesets/1781569700-feat-agent-fork-set.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): fork-set derivation — the data model for the fork bar (forks Phase 2)
--- a/frontend/app/view/agent/fork/fork-set.test.ts
+++ b/frontend/app/view/agent/fork/fork-set.test.ts
@@ -1,0 +1,95 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from "vitest";
+import { computeForkSet, type ForkDefinition } from "./fork-set";
+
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1466, opened 2026-06-16
feat(statusbar): click CPU to open an adaptive per-core usage panel

a5af:
## What

Clicking the **CPU** readout in the bottom status bar now opens a panel showing **live per-core CPU usage**. Previously the status bar showed only the system-wide average and the readout was not interactive.

## Why it's frontend-only

The backend already collects and publishes per-core CPU — `get_cpu_data()` (`agentmux-srv/src/backend/sysinfo.rs:29-41`) inserts both `cpu` (avg) and `cpu:0..N` (per-core) into the existing `sysinfo`/`local` event every ~1s. The panel just consumes keys matching `^cpu:\d+$`. **No backend changes, no new events/RPCs.**

## Adaptive density (scales 4 → 256+ cores)

No single layout reads well from 8 to 128 cores, so the panel picks a representation from the core count:

| Cores | Tier | Unit |
|-------|------|------|
| ≤16 | Rows | `Core 0 ▕███░▏ 52%` (label + bar + %) |
| 17–64 | Cells | compact tile: index + % + mini bar (auto-fill grid) |
| 65+ | **Heatmap** | small squares colored by a continuous idle→busy ramp; detail on hover/focus + legend |

The heatmap square size is **computed** from panel width + core count (near-square grid), shrinking toward a 10px floor as the count grows, then scrolling (height-capped `min(60vh, 420px)`) for very high counts. So 64 → ~16px squares, 128 → ~12px, 256 → ~10px + light scroll — a glanceable "which cores are hot" block instead of a long list.

## Interaction & a11y

- Mirrors the canonical `TokenUsageIndicator → TokenBreakdownPopover` pattern: `usePaneOverlay` (paints over browser-pane HWNDs), `computeMenuPosition` placement `top-end`, `@floating-ui` `autoUpdate`, Esc + outside-click close.
- Never color-only: each square has `aria-label="Core N, X%"`, squares are focusable/arrow-navigable, and the header keeps the exact aggregate number. A single live readout line shows the hovered/focused core (cheaper + keyboard-friendly vs. 128 tooltips).

## Files

- **New** `frontend/app/statusbar/CpuCoresPopover.tsx` — the panel (3 tiers + computed square sizing + readout).
- **New** `frontend/app/statusbar/cpu-color.ts` — `cpuColor` (threshold, header) + `loadColor` (continuous ramp). Extracted so `SystemStats` and the popover share it without a circular import.
- **New** `frontend/app/statusbar/_cpu-cores-popover.scss` — styles for all three tiers + legend; registered in `StatusBar.scss`.
- **Modified** `frontend/app/statusbar/SystemStats.tsx` — CPU `<span>` → `<button>`; open/anchor state + Esc/outside-click; renders the popover in a `<Portal>`.
- **Spec** `docs/specs/SPEC_STATUSBAR_CPU_CORES_PANEL_2026_06_15.md`.

## Testing

`node_modules` isn't installed in this checkout, so `tsc` wasn't run locally; the change is self-contained (one new component + helpers, consuming an existing event). Manual verification target: 8, 32, 64, 128 cores (synthesize `cpu:N` payloads if no high-core machine is handy).

### DIFF
--- a/.changesets/1781569814-feat-statusbar-click-cpu-to-open-an-adaptive-per-core-usage--sm3y.md
+++ b/.changesets/1781569814-feat-statusbar-click-cpu-to-open-an-adaptive-per-core-usage--sm3y.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(statusbar): click CPU to open an adaptive per-core usage panel
--- a/docs/specs/SPEC_STATUSBAR_CPU_CORES_PANEL_2026_06_15.md
+++ b/docs/specs/SPEC_STATUSBAR_CPU_CORES_PANEL_2026_06_15.md
@@ -1,0 +1,194 @@
+# SPEC: Status-bar CPU → per-core panel
+
+**Date:** 2026-06-15
+**Repo:** agentmuxai/agentmux @ v0.46.0
+**Status:** Draft / ready to implement
+**Effort:** ~Small–Medium. Frontend-only. ~1 new component with three adaptive render tiers (~200–240 LOC incl. the computed square-sizing + `loadColor` ramp) + ~15 LOC change to `SystemStats.tsx` + ~1 SCSS partial. **No backend changes.**
+
+---
+
+## 1. Goal
+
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

a5af:
@reagentx-workflow review

a5af:
@reagentx-workflow review

a5af:
@reagentx-workflow review

a5af:
@reagentx-workflow review

PR #1471, opened 2026-06-16
docs(retro): cross-channel agent history empty — global snapshot stores a channel-local sourceBlockId

AgentX-asaf:
## Incident

Closing **Naki** in a running 0.44.1 instance and reopening it in a freshly-built 0.46.0 portable showed an **empty conversation**. No data lost — the 32 MB transcript is intact in `agent:f81f9785…:current`.

## Root cause (pinned from disk + code, no runtime drive)

The agent-anchored **global** snapshot (`output.state.json` in `agent:<defId>:current`) carries `sourceBlockId`, which `writeSnapshotNow` always sets to **`model.blockId`** — the *writing channel's* local block id — and the snapshot write mirrors that to the global zone. The only reader, `global_output_source` (`app_api.rs:1066`), resolves the agent zone by looking that block up **locally** (`wstore.get::<Block>(block_id)`). In another channel the block doesn't exist → fallback skipped → empty render, even though the global `output` is present.

- Worked in 0.44.1 (`sourceBlockId` *is* the local block there); empty in 0.46.0 (foreign block).
- `session_id: null` is a **separate** `--resume` gap, not this.

## Blast radius

Any agent that has run live since the #1403 backfill is poisoned → opening it in any fresh build/channel renders empty. Per-build isolation makes this hit on nearly every `task package`.

## Fix direction (separate PR)

Normalize `sourceBlockId → ""` when mirroring the snapshot to the **global** zone (local keeps the real id); one-shot heal for already-poisoned global snapshots; add a 'run in channel A, open in fresh channel B → renders' e2e.

Full analysis: `docs/retro/retro-legacy-agent-history-cross-channel-2026-06-16.md`.

### DIFF
--- a/docs/retro/retro-legacy-agent-history-cross-channel-2026-06-16.md
+++ b/docs/retro/retro-legacy-agent-history-cross-channel-2026-06-16.md
@@ -1,0 +1,125 @@
+# Retro: agent opens with no conversation history in a fresh build/channel
+
+**Date:** 2026-06-16
+**Author:** AgentX
+**Severity:** P1 (history *appears* lost — it is intact but unreachable on open)
+**Status:** Root-caused (from on-disk data + code; no data lost). Fix pending.
+
+> **Correction kept on purpose:** my first pass blamed the registry's
+> `session_id: null`. That's a **separate** gap (provider `--resume` / *continue*),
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #1472, opened 2026-06-16
fix(agent-pane): global snapshot stored a channel-local sourceBlockId → empty history cross-channel

AgentX-asaf:
Fixes the incident in #1471 (retro).

## Root cause

The agent-anchored **global** snapshot (`output.state.json` in `agent:<defId>:current`) was mirrored **verbatim** from the per-channel copy, including `sourceBlockId = model.blockId` — the *writing channel's* local block. The only reader, `global_output_source` (`app_api.rs:1066`), resolves the agent zone by looking that block up **locally** (`wstore.get::<Block>`). Opening the agent in another channel/build can't anchor that block → the read fallback is skipped → empty pane, though the agent's `output` is intact in the global zone (verified: Naki, 32 MB).

## Fix

1. **Normalize `sourceBlockId → ""` when mirroring to the global store** (`write_session_state`). The per-channel copy keeps the real id for same-channel restore; the global copy is agent-anchored, so a cross-channel open anchors on its own fresh local block (which maps to the agent → the fallback resolves).
2. **One-shot idempotent startup heal** (`heal_global_snapshot_source_block_ids`, wired in `main.rs` after the backfill) rewrites `sourceBlockId → ""` for snapshots already poisoned before this fix, so existing agents (the 9 in the registry) recover on next boot.
3. **Unit test** for the normalization (strip + preserve other fields + idempotent + non-JSON passthrough).

Backend-only; per-channel restore unchanged. Builds clean (`cargo build --release -p agentmux-srv`).

## Test

Open a previously-poisoned agent (e.g. Naki) in a fresh `task dev`/portable → history renders (heal runs on startup; new mirrors stay agent-anchored).

Follow-up (noted in the retro, not here): a 'run in channel A, open in fresh channel B → renders' e2e; and the separate `session_id` backfill for `--resume`.

### DIFF
--- a/.changesets/1781572123-fix-agent-pane-cross-channel-history-empty-global-snapshot-s-ccbi.md
+++ b/.changesets/1781572123-fix-agent-pane-cross-channel-history-empty-global-snapshot-s-ccbi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): cross-channel history empty — global snapshot stored a channel-local sourceBlockId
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1474, opened 2026-06-16
feat(agent): PaneRegions declarative region container (forks Phase 1)

a5af:
## What

The **§5.1 region model** of `SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15.md`: `<PaneRegions>`, a declarative `region → content` map that replaces the hand-ordered ~16-surface JSX stack in `agent-view.tsx`.

Surfaces **register into a named region** — `top-fixed / dock / stream / alert / queue / status / input / forks / overlay` — instead of being positioned by JSX accident. Each region owns its layout contract in `PaneRegions.scss`: `stream` is the lone flex-grow band (the scrolling conversation), `overlay` is z-stacked over the column (the pane's *real* layers — focused panel, pane-scoped modals), and the rest are shrink-to-content. An absent or empty region renders **no wrapper** (zero layout cost).

Presentational only — it owns no state, matching the "derive from a source of truth" discipline.

## Why now (it's unblocking another agent)

This is the shared dependency the **session-digest accessory** is explicitly blocked on — `SPEC_SESSION_DIGEST_AS_PANE_ACCESSORY_2026_06_15.md` states "Building `<PaneRow>` / `<PaneRegions>` — owned by the forks/aux-pins spec Phase 1," and its Phase B registers the digest into the `top-fixed` region map. The fork bar will register into the `forks` region the same way.

## Scope

Lands the **container + region model + tests**. The pixel-sensitive migration of `agent-view.tsx`'s render tree onto it gets its **own** follow-up PR, so the whole render tree isn't refactored in one blind shot.

## Verification

- **vitest 6/6** — region modifier classes + `data-region`, canonical top→bottom ordering (provided out of order → reordered), absent-region skip, empty-array/all-nullish treated as no content, array-of-nodes in one region, reactive show/hide.
- **stylelint** clean; **tsc** clean.

## Files

- `frontend/app/view/agent/components/PaneRegions.tsx` (new)
- `frontend/app/view/agent/components/PaneRegions.scss` (new)
- `frontend/app/view/agent/components/PaneRegions.test.tsx` (new)

### DIFF
--- a/.changesets/1781572475-feat-agent-pane-regions.md
+++ b/.changesets/1781572475-feat-agent-pane-regions.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): PaneRegions declarative region container (forks Phase 1, aux-pins)
--- a/frontend/app/view/agent/components/PaneRegions.scss
+++ b/frontend/app/view/agent/components/PaneRegions.scss
@@ -1,0 +1,43 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+//
+// PaneRegions — per-region layout contracts. The agent pane is a flex column;
+// `stream` is the only flex-grow region (the scrolling conversation), every
+// other region is fixed-height / shrink-to-content, and `overlay` is z-stacked
+// over the column. Token-driven; no raw values beyond structural zeros.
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1475, opened 2026-06-16
docs(retro): cross-channel conversation continuity 'fixed then broke again' (all agents)

AgentX-asaf:
Companion to #1471 — the **full regression** (display + session continuity).

## Why it looked fixed and broke again

After #1403 (Jun 14) agents rendered their history on open — but that was verified only in the freshly-**seeded** state (`sourceBlockId:""`, agents not run since). **Normal use undoes it**, via two independent mechanisms that mask each other:

- **Display (since #1399):** `write_session_state` mirrors the snapshot to the global zone verbatim, including a **channel-local `sourceBlockId`**; the reader needs that block locally → blank cross-channel. *(fixed: #1472)*
- **Continuity (never implemented):** the registry `session_id` that would carry the provider session across channels is **never written by production code** (only tests). So a fresh-build open has no sid to `--resume` → the CLI **starts a new session**, which latches in after one turn and shadows the original.

The blank display is what makes you interact with a 'fresh' agent, which spawns the shadowing session — that's the compound.

## Data is safe

Every original is intact in the global `output` (Naki: 32 MB) and the provider jsonl (Naki: 6.28 MB). The damage is pointer/anchor, not bytes.

## Action items (in the retro)

Populate `session_id` on each turn + read it on cross-channel open (`--resume`); backfill from the **original** session (not the post-bug short one); guard 'resume-before-new-session'; recover already-affected agents; add the missing run-live-then-open-fresh e2e.

Full analysis: `docs/retro/retro-cross-channel-conversation-continuity-regression-2026-06-16.md`.

### DIFF
--- a/docs/retro/retro-cross-channel-conversation-continuity-regression-2026-06-16.md
+++ b/docs/retro/retro-cross-channel-conversation-continuity-regression-2026-06-16.md
@@ -1,0 +1,124 @@
+# Retro: cross-channel agent conversation continuity "fixed, then broke again" — all agents
+
+**Date:** 2026-06-16
+**Author:** AgentX
+**Severity:** P1 (every agent opens to a fresh/short conversation cross-channel; original intact but orphaned)
+**Status:** Root-caused. Display half fixed (#1472); continuity half + recovery pending.
+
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #1479, opened 2026-06-16
fix(agent-pane): backfill registry session_id so cross-channel open resumes the original conversation

AgentX-asaf:
Second half of the continuity regression (#1475 retro). Companion to the display fix #1472.

## Root cause

The registry record `session_id` is **read** on launch (surfaced to the picker, which passes `--resume <sid>` on the first turn of a reattached block) but is **never written by production code** — every `.session_id = Some(..)` in the tree is in tests. So a cross-channel / fresh-build open has no session to resume → the CLI starts a **new** provider session, which after one turn latches in and **shadows the original conversation**.

## Fix

An idempotent startup backfill (`backend::session_backfill`) that populates `session_id` from each agent's **largest** provider session — *largest, not newest*, so an accidental fresh session (started when a blank cross-channel open failed to resume) can never win over the real conversation. Once set, `--resume` keeps the same id across turns, so a single startup pass keeps continuity solid without per-turn wiring. This **also recovers** already-affected agents (e.g. Naki → its original 6 MB session).

## Tests (high-value)

- `encode_slug_matches_claude_convention` — workspace→project-slug encoding (verified against a real on-disk slug).
- `largest_session_beats_a_fresh_short_one` — the recovery guarantee: pick the 6 MB original over a 15 KB fresh one.
- `backfill_fills_empties_picks_largest_and_is_idempotent` — only fills empties, picks largest, skips records with no transcript, idempotent on re-run.

`cargo test -p agentmux-srv session_backfill` → **3 passed**. Backend builds clean.

> Pairs with #1472 (display) — both are needed for a fully-recovered cross-channel open: #1472 heals the snapshot anchor, this resumes the original session. Follow-up (retro): a run-live-in-A → open-in-fresh-B e2e.

### DIFF
--- a/.changesets/1781600326-fix-agent-pane-backfill-registry-session-id-so-a-cross-chann-61b5.md
+++ b/.changesets/1781600326-fix-agent-pane-backfill-registry-session-id-so-a-cross-chann-61b5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): backfill registry session_id so a cross-channel open resumes the original conversation
--- a/agentmux-srv/src/backend/mod.rs
+++ b/agentmux-srv/src/backend/mod.rs
@@ -25,6 +25,7 @@ pub mod rpc_types;
 pub mod schema;
 pub mod service;
 pub mod session_archive;
+pub mod session_backfill;
 pub mod shellexec;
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1483, opened 2026-06-16
fix(dev): reap a stale Vite squatting the dev port on relaunch

AgentX-asaf:
## Problem

`task dev` appeared to 'keep crashing on launch.' Root cause (retro: `docs/retro/retro-dev-instance-vite-orphan-port-2026-06-16.md`): when the host window closes, the backgrounded **Vite orphans** and squats the **deterministic per-clone port**; the next `task dev` then dies immediately on the `--strictPort` pre-check (`❌ Port … already in use` → `dev:serve exit status 1`). Confirmed live: a `node` Vite holding the port with **0** host processes alive.

A shell `trap`/`$!` cleanup is **impossible** here — `dev:serve` runs in go-task's embedded shell (mvdan/sh) where `$!` is a goroutine id, not a killable OS PID (the code even notes this).

## Fix

Make the busy-port pre-check **self-heal** instead of erroring: reap the process holding the port (`taskkill /PID /T /F` on Windows, `kill -9` on Unix via `lsof -ti`), re-check, and only fail if something **foreign** still holds it. Additive + best-effort: if the reap can't run, it falls through to the same error as before — **no regression**. Keeps `--strictPort` (loud failure beats silent rebind).

## Test

`task --list` parses. The reap path degrades safely to the prior behavior; full exercise needs a live `task dev` (the foreground host-loop can't run in CI). Follow-up noted in the retro: a `dev:doctor` reaper + applying the same to `dev:standalone`.

### DIFF
--- a/.changesets/1781602735-fix-dev-reap-a-stale-vite-squatting-the-dev-port-instead-of--9o62.md
+++ b/.changesets/1781602735-fix-dev-reap-a-stale-vite-squatting-the-dev-port-instead-of--9o62.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dev): reap a stale Vite squatting the dev port instead of erroring on relaunch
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1484, opened 2026-06-16
feat(agent): send-now queue — instant panel, boundary delivery, ArrowUp recall, no drop

a5af:
Reworks the **"send now" panel** for messages typed while an agent is busy, per the three reported issues.

## Fixes

**1. Panel appears instantly.** `handleSendMessage` dispatched `TurnStart` unconditionally; doing so while a turn is already running regresses the phase `Streaming → Submitting`, and the "Send now" button is gated on `isInterruptibleTurn` (false during `Submitting`) — so it only reappeared on the next stream event ("a couple seconds late"). Now `TurnStart` fires only when the agent is idle.

**2. Queued messages no longer drop.** Root cause: the pending→conversation promotion is driven entirely by the backend `agent-message-accepted` event, which **only the subprocess controller emitted**. Claude now uses the **persistent** controller (#1451), which never emitted it — and the persistent CLI does **not** echo the user prompt in its stream (verified). So a queued message had *no* render path; the only thing that ever removed it was the 30s `PENDING_TIMEOUT_MS`, which silently dropped it. 

Now:
- The persistent controller emits `agent-message-accepted` on receive (threaded `message_id` from the `agentinput` RPC), so sent messages render and clear from the queue — fixes idle-send rendering too.
- A queued-while-busy message is **held in the frontend** (no RPC, no expiry) until delivered — it cannot drop while waiting.

**3. Delivered at the next tool-call boundary + ArrowUp recall.** The held message is flushed at the next tool-call boundary (`currentTool` change) or turn end, so it lands just before the agent's next tool call — the agent finishes its current train of thought, then picks it up (the CLI consumes a stdin message at its next inference). Because the message is *held* (never sent until then), **ArrowUp on an empty composer** pulls it back into the composer and removes it from the queue — a true un-send, matching the Claude Code CLI gesture.

## Changes
- `persistent.rs`: `send_message(message, message_id, config)` + `publish_message_accepted` (mirrors subprocess); threaded in `websocket.rs`; `app_api.rs` programmatic send passes `None`.
- `useAgentCommands.ts`: held queue + `flushHeldMessages` / `recallLatestHeld` / `hasHeldMessages`; idle sends keep immediate delivery + lost-delivery expiry.
- `agent-view.tsx`: skip `TurnStart` when working; tool-boundary/turn-end effect drives the flush; wire ArrowUp recall.
- `AgentFooter.tsx`: ArrowUp recall handler + prop. `PendingMessagesPanel.tsx`: updated copy.

## Testing
- `tsc --noEmit` — clean for all touched files (unrelated pre-existing errors only surface under a wrong local node; node 24 required).
- No existing tests broken (reducer/state unchanged; no Footer/Panel/commands test files; panel copy not asserted).
- **Not run locally:** frontend vitest (needs node ≥24, not installed here) and the full app. Recommend a real-app pass: send to a busy Claude pane → panel shows instantly → ArrowUp recalls → otherwise it lands at the next tool call → it never drops.

### DIFF
--- a/.changesets/1781605307-feat-agent-send-now-queue-instant-panel-hold-until-next-tool-sdbj.md
+++ b/.changesets/1781605307-feat-agent-send-now-queue-instant-panel-hold-until-next-tool-sdbj.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): send-now queue — instant panel, hold-until-next-tool-call delivery, ArrowUp recall, no 30s drop
--- a/frontend/app/view/agent/agent-view.tsx
+++ b/frontend/app/view/agent/agent-view.tsx
@@ -736,7 +736,7 @@ const AgentPresentationView = ({ model, agentId }: { model: AgentViewModel; agen
         prevTool = tool;
         const turnIdle = phaseKind === "Idle" || phaseKind === "Done";
         if ((newToolCall || turnIdle) && commands.hasHeldMessages()) {
-            commands.flushHeldMessages();
+            void commands.flushHeldMessages();
         }
     });
 
--- a/frontend/app/view/agent/agent-view.tsx
+++ b/frontend/app/view/agent/agent-view.tsx
@@ -672,7 +672,15 @@ const AgentPresentationView = ({ model, agentId }: { model: AgentViewModel; agen
         // where the message flashed in the amber zone between Streaming
         // promotion and agent-message-accepted. See ANALYSIS_IDLE_SEND_RACE_2026_06_11.md.
         const wasAlreadyWorking = workingFromPhase(paneSnapshot(model.blockId)?.turnPhase ?? { kind: "Idle" });
-        dispatchPane(model.blockId, { type: "TurnStart", at: Date.now() }, "user");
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

a5af:
Addressed the P2 (send ordering) in 68d74e5f: `deliverToBackend` now awaits `AgentInputCommand` and `flushHeldMessages` awaits each delivery in turn, so queued messages reach the persistent CLI's stdin in submission order (no more racing cmd:args round-trips).

a5af:
Rebased onto latest main. Main's #1492 independently landed the same backend fix (persistent controller emits `agent-message-accepted` via `config.message_id`), so I **dropped my redundant backend changes** — this PR is now **frontend-only**: the send-now queue UX (instant panel, hold-until-next-tool-call delivery, ArrowUp recall, no 30s drop) riding on main's backend emit. The ordering fix is preserved.

a5af:
Addressed the residual-race P2 in b91a8444: added a single-flight guard to `flushHeldMessages` — only one flush drains at a time, looping over the queue (shift) so messages queued mid-flush are delivered in order. Overlapping boundary triggers can no longer start concurrent flushes.

PR #1487, opened 2026-06-16
fix(agent-session): read global snapshot first — permanent fix for cross-channel blank pane

AgentX-asaf:
## Summary

- **Root fix:** `read_session_state` now reads the **global** filestore first. The per-channel copy carries a channel-local `sourceBlockId` that is stale the moment any other block opens the agent. By preferring global (always `sourceBlockId=\"\"`), cross-channel opens never get a foreign block ID, so `v2SameBlock=true` and full history renders.

- **Invariant G1 (debug_assert):** `normalize_snapshot_for_global` now asserts in debug builds that the resulting global snapshot has `sourceBlockId=\"\"`. Catches any future write-path regression before it ships.

- **schemaVersion 3 + startup migration:** `migrate_snapshots_v2_to_v3` runs at startup (after `backfill_session_ids`), reads each agent's v2 global snapshot, injects `sessionId` from the registry, and writes a v3 snapshot. This makes the snapshot self-contained for cross-channel `--resume` — no registry lookup needed at open time.

- **Architecture spec:** `docs/specs/SPEC_AGENT_GLOBAL_PORTABILITY_2026-06-16.md` defines the complete final design: data zone model, invariants G1/G2, schema versioning contract, migration sequence, and portability guarantee for all future version bumps.

## Root cause (eliminated)

`read_session_state` read per-channel first → per-channel had `sourceBlockId=\"658d3caf\"` (written on last close in this channel) → `v2SameBlock=false` → NDJSON fallback → last 200 lines shown, not full history.

Global had `sourceBlockId=\"\"` all along (correct since #1472), but was never reached because per-channel shadowed it. This was the root cause that survived all three prior patches (#1472, #1479, #1486).

## Test plan

- [ ] Build backend (`task build:backend`) — already done, compiles clean
- [ ] Restart `task dev` to pick up the new binary
- [ ] Open Naki → full history renders (not last 200 lines, not blank)
- [ ] Close + reopen Naki → still full history (writeSnapshotNow does not re-poison)
- [ ] Open any other agent → full history
- [ ] Check srv log: `agent:session:read` log line shows `sourceBlockId=\"\"` in restored snapshot
- [ ] Check srv log: startup shows `upgraded N` if any agents have v2 snapshots

### DIFF
--- a/.changesets/1781012789-fix-cef-low-memory-resume-button-inert-onclick-agc1.md
+++ b/.changesets/1781012789-fix-cef-low-memory-resume-button-inert-onclick-agc1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): low-memory pause page Resume button was inert (double-quoted JS string inside a double-quoted onclick attribute) — wire it via a <script> + addEventListener so OOM-paused windows can actually resume
--- a/.changesets/1781093156-fix-agent-pane-replace-remaining-index-with-key-in-virt-list-vxq9.md
+++ b/.changesets/1781093156-fix-agent-pane-replace-remaining-index-with-key-in-virt-list-vxq9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #1488, opened 2026-06-16
fix(term): terminal I/O has priority over perf telemetry on WS egress

a5af:
## Problem

In a terminal pane, after a long session, typed keystrokes echo back with a periodic stutter — the echo is "delayed by a tick" and the tick **lines up with the CPU/perf readout refreshing** (sysinfo, ~1 Hz). Invisible early, perceptible late.

## Root cause (confirmed in code)

Terminal echo output and droppable perf telemetry share **one** FIFO per-connection event channel (`EventBus::register_ws` → single `event_rx`), drained by an **unbiased** `tokio::select!` (`websocket.rs`):

1. **Same channel.** Terminal `blockfile` echo and `sysinfo`/`blockstats` telemetry both funnel through `broker.publish` → `EventBusBridge` → the one `event_rx`. A telemetry burst enqueued microseconds before an echo blocks it head-of-line.
2. **Unbiased select.** Even across branches, `select!` without `biased;` polls in random order, so a ready telemetry frame can win the coin-flip and make a keystroke wait for a `socket.send().await`.
3. **Grows with use.** The collector emits one `blockstats` event **per open block** every tick and refreshes **all** processes (`ProcessesToUpdate::All`) — both scale with session length, widening the window in which an echo lands behind telemetry.

## Fix — Phase 1

- **Two egress lanes per connection.** `register_ws` now returns `WsReceivers { priority, background }`. `EventBusBridge::send_event` classifies by `WaveEvent` type (the only spot the raw type is visible before RPC-wrapping): `sysinfo` + `blockstats` → background; everything else → priority.
- **Biased select.** The egress loop is now `biased;`, ordered: incoming keystrokes → RPC replies → agent bus → priority events (echo) → background events (telemetry) → ping. Telemetry only flushes when the priority lanes are momentarily idle, so it **can never delay a keystroke echo**.

All existing direct `broadcast_event` callers keep priority-lane behavior (none send telemetry) — tiny blast radius. The shared event-forwarding logic is factored into one `forward_event` helper used by both lanes.

This is intentionally "complete priority": under a flood of terminal output, telemetry yields and the gauges resume the instant the priority lanes drain — exactly the requested behavior.

## Not a regression

Distinct from the prior frontend typing fixes (#926 writeInFlight, double-rAF tearout, predictive echo) — those were all frontend frame-pacing; this is server→client egress scheduling. Predictive echo *masks* this for printable chars; fixing the channel benefits every byte (control sequences, program output, cursor moves).

## Tests / verification

- New `eventbus` tests: `test_lane_separation` (background send never lands on priority lane and vice versa) and `test_bridge_routes_telemetry_to_background` (bridge demotes sysinfo/blockstats, keeps blockfile on priority). All 6 eventbus tests pass.
- Manual repro/verify via the existing `term-echo-render` perf mark: record while typing continuously ~30s with many panes open; the ~1 Hz outlier correlated with the sysinfo tick should be gone.

## Follow-ups (in spec)

Phase 2 (coalesce telemetry to latest-only + trim per-tick process refresh), Phase 3 (frontend popover idle-scheduling), Phase 4 ("step aside while typing") — all documented in `docs/specs/SPEC_TERMINAL_INPUT_PRIORITY_OVER_SYSINFO_2026_06_16.md`. Phases 1+2 remove the latency; 3–4 are conditional on residual measurement.

### DIFF
--- a/.changesets/1781625002-fix-term-give-terminal-i-o-priority-over-perf-telemetry-on-t-cqk0.md
+++ b/.changesets/1781625002-fix-term-give-terminal-i-o-priority-over-perf-telemetry-on-t-cqk0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): give terminal I/O priority over perf telemetry on the WebSocket egress
--- a/agentmux-srv/src/backend/eventbus.rs
+++ b/agentmux-srv/src/backend/eventbus.rs
@@ -10,11 +10,32 @@ use std::sync::{Arc, Mutex};
 
 use serde::{Deserialize, Serialize};
 
-use super::wps::{WaveEvent, WpsClient};
+use super::wps::{WaveEvent, WpsClient, EVENT_SYS_INFO, EVENT_BLOCK_STATS, EVENT_BLOCK_FILE};
 
 // ---- Event type constants ----
 
 pub const WS_EVENT_RPC: &str = "rpc";
+
+/// Egress priority lane for a server→client event.
+///
+/// `Background` is reserved for droppable perf telemetry (sysinfo + per-block
+/// stats) that must never delay interactive terminal I/O; everything else is
+/// `Priority`. The WebSocket egress loop drains the priority lane before the
+/// background lane (see `server/websocket.rs` and
+/// `docs/specs/SPEC_TERMINAL_INPUT_PRIORITY_OVER_SYSINFO_2026_06_16.md`).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Lane {
+    Priority,
+    Background,
+}
+
+/// The pair of receivers handed to a WebSocket connection on registration.
+/// Terminal echo + interactive events arrive on `priority`; perf telemetry on
+/// `background`.
+pub struct WsReceivers {
+    pub priority: tokio::sync::mpsc::UnboundedReceiver<serde_json::Value>,
+    pub background: tokio::sync::mpsc::UnboundedReceiver<serde_json::Value>,
+}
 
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
Good catch on the `biased;` starvation — acknowledged. It's the known tradeoff of "complete priority": the background lane is only drained when the priority lanes are momentarily empty, so a *sustained* priority flood (e.g. `! yes`) can let sysinfo/blockstats accumulate on the unbounded `background_rx` until the flood pauses, then flush as a stale burst. You're right that before this PR telemetry shared the drained channel and didn't accumulate separately.

The intended bound is **Phase 2 coalescing** (latest-value-per-`(event, scope)`), already scoped in `docs/specs/SPEC_TERMINAL_INPUT_PRIORITY_OVER_SYSINFO_2026_06_16.md` §4.2 — a coalesced background lane keeps only the newest sysinfo/blockstats frame, so it can't grow during a flood and the post-flood flush is a single current reading rather than a stale backlog. That's the correct home for the bound; this PR is just the lane split + bias.

Two mitigating realities in the meantime: (1) telemetry ingress is low-rate (sysinfo ~1/s, blockstats a few/s), so accumulation is slow; (2) a sustained terminal flood also saturates the socket writes, so the loop spends most of its time awaiting `socket.send`, during which the priority lanes do drain and the background lane gets serviced — true never-empty starvation is the rare worst case. Net: a slow, bounded-in-practice growth that Phase 2 closes properly.

Happy to pull Phase 2 coalescing into this PR if you'd prefer to land them together rather than as a follow-up.

PR #1489, opened 2026-06-16
chore(release): add --as <type> override to force a specific bump

a5af:
## Why
`scripts/release.sh` computes the release bump as the **highest** type among pending changesets (major > minor > patch). With ~37 changesets queued (14 `minor` + 23 `patch`) on `main`, a release is forced to **minor (0.47.0)** — there's currently **no way to cut a deliberate patch** even when you want one.

## What
Adds an opt-in override to `release.sh`:

```bash
task release -- --as patch     # force a patch even while minors are queued
task release -- --as minor      # or any of patch|minor|major  (--bump is an alias)
```

- **Default behavior unchanged** — no flag = highest-type computation as before.
- The override **still consumes and changelogs every pending changeset**; only the version *number* is forced. The higher-typed changes ride along under the forced version (a deliberate under-claim).
- Prints a loud **`WARNING`** when you force a bump *lower* than the changesets request, so it can never happen silently.
- Validates input (`--as bogus` and unknown args error out).

## Validation (`--dry-run`, no mutation)
| Invocation | Result |
|---|---|
| `release.sh --dry-run` | `Release type: minor` (unchanged default) |
| `release.sh --dry-run --as patch` | `NOTE` + `WARNING: forcing a LOWER bump…` → `Release type: patch` |
| `release.sh --dry-run --as=minor` | silent (matches computed) → `minor` |
| `release.sh --as bogus` | `ERROR: --as must be one of patch\|minor\|major` |
| `release.sh --frobnicate` | `ERROR: unknown argument` |

## Notes
- Script + `CLAUDE.md` doc update + changeset (patch).
- Semver caveat is intentional and documented: forcing patch while minors are queued means those features ship under a patch version. For a continuously-shipped app that's a valid choice; the warning keeps it deliberate.

### DIFF
--- a/.changesets/1781625491-chore-release-add-as-type-override-to-force-a-specific-versi-nmfk.md
+++ b/.changesets/1781625491-chore-release-add-as-type-override-to-force-a-specific-versi-nmfk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+chore(release): add --as <type> override to force a specific version bump (e.g. patch even when minor changesets are queued)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -207,6 +207,8 @@ git push -u origin agenta/release-vX.Y.Z
 
 The release script picks the highest bump type across pending changesets (major > minor > patch), runs `scripts/bump-wrapper.sh`, appends to `VERSION_HISTORY.md`, and deletes the consumed changesets.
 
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
Thanks @reagentx-workflow — both P2s fixed in 49b35faf:
- **#1 (`--as` as final token):** args are now consumed within each `case` branch (no loop-bottom `shift`), so an empty value no longer trips `set -e`; it prints `ERROR: --as must be one of patch|minor|major (got '')` and exits 1.
- **#2 (`--as=` empty):** validation now gates on a `FORCE_SET` flag instead of `[[ -n $FORCE_TYPE ]]`, so an empty value errors instead of silently falling back.

Re-validated all combinations via `--dry-run` (incl. `--as` last, `--as=`, `--as bogus`, value+flag).

PR #1491, opened 2026-06-16
feat(agent): tool blocks stay expanded until scrolled off the top

a5af:
## What

Replaces the 3 s post-completion auto-collapse on agent-pane tool blocks with **scroll-position-driven collapse**: a completed tool stays **expanded while it's on screen** and collapses once it **scrolls off the top** of the transcript. Everything else is unchanged — pin override, hover-hold, auto-expand while running/pending, immediate collapse for denied/canceled, and the ✗/red collapsed-row treatment for failures.

## Why

The 3 s timer collapsed tools while you were still reading them. Position-driven collapse keeps a tool open exactly as long as it's visible, and folds it only once it's out of view — and it removes a wall-clock timer (the codebase's "no grace timers" convention).

## How

A pane-level `documentState.expandedTools` set is the post-completion hold, durable across virtualization unmount and shared by **both** the visual render and the layout-slice height decision — closing the component-vs-layout divergence that `expansion-source.ts` already documented as a known gap.

- **Add** the id on `ToolBlock`'s `active → inactive` transition (the same trigger that armed the old timer), so loaded-history tools — which never transition this session — are never added and stay collapsed (no load-flash; preserves the existing guard).
- **Release** the id in `AgentDocumentVirtualList` once the row's bottom edge has scrolled above the viewport top — DOM-based (`data-node-id` + `getBoundingClientRect`), zoom-safe, works for both virtualized and streaming-buffer rows. **Latched**: once released it won't re-add, so scrolling back up leaves it collapsed (click to pin to re-open).
- `currentExpansion()` returns open while held, so virtualized row heights track the visual.

### Phase 1 (this PR) vs Phase 2

The release is **gated on stick-to-bottom**, where collapsing a row above the fold causes no scroll jump — the primary live-streaming case ("new tool completes → more output → it scrolls off the top → collapses"). The scrolled-up case (collapse while reading history, needing `captureTopmostAnchor`/`restoreScrollFromAnchor` compensation) is a documented **Phase 2** follow-up; those tools are off-screen so only scroll position, not the collapse, would be visible.

## Tests / verification

- New `expansion-source` cases (held tool resolves open; pin/running still win); new `ToolBlock` tests (`heldOpen` renders in-flow; `onHoldOpen` fires once on completion and **never** on already-completed mount).
- `tsc -p tsconfig.json --noEmit`: **no new errors** (confirmed the only matches in touched-area, `ActivityDock.tsx`, are a pre-existing baseline — they error identically on `main`).
- **511** agent-view vitest tests pass.
- ⚠️ Not verifiable headlessly: the actual on-screen feel (expand → stays while visible → collapses off the top while streaming). Needs a running app.

## Docs

- Plan: `docs/specs/PLAN_TOOL_BLOCK_SCROLL_DRIVEN_COLLAPSE_2026_06_16.md`
- Analysis: `docs/analysis/ANALYSIS_TOOL_BLOCK_SCROLL_DRIVEN_COLLAPSE_2026_06_16.md`

### DIFF
--- a/.changesets/1781643855-feat-agent-tool-blocks-stay-expanded-until-scrolled-off-the--mwow.md
+++ b/.changesets/1781643855-feat-agent-tool-blocks-stay-expanded-until-scrolled-off-the--mwow.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): tool blocks stay expanded until scrolled off the top (replaces 3s collapse timer)
--- a/docs/analysis/ANALYSIS_TOOL_BLOCK_SCROLL_DRIVEN_COLLAPSE_2026_06_16.md
+++ b/docs/analysis/ANALYSIS_TOOL_BLOCK_SCROLL_DRIVEN_COLLAPSE_2026_06_16.md
@@ -1,0 +1,319 @@
+# Analysis: Scroll-Driven Tool-Block Collapse (replace the post-completion timer)
+
+**Date:** 2026-06-16
+**Author:** smike (agent)
+**Status:** Analysis / design proposal (no code yet)
+**Area:** agent-pane transcript — tool-call block expand/collapse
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

PR #1492, opened 2026-06-16
fix(persistent-agent): sent messages invisible + history truncated after #1487

AgentX-asaf:
## Summary

- **Bug 2 — sent message invisible (regression)**: `PersistentSubprocessController.send_message` never emitted `agent-message-accepted`. The frontend has no other signal to add the user_message node for persistent agents (unlike subprocess agents, the persistent CLI does not echo user turns on stdout). Fix: add `message_id` to `PersistentSpawnConfig`, mirror the `emit_message_accepted` helper from subprocess.rs, pass `cmd.message_id` from the AgentInput websocket handler.
- **Bug 2b — user messages lost on reopen (persistent agents)**: Because the persistent CLI outputs only the assistant side, `parseHistoryLines` found no `user_message` events in the NDJSON and so user turns were invisible in history. Fix: `persist_to_blockfile_silent` (new helper in shell.rs) writes the user-message line to the per-block output file + global zone WITHOUT a WPS event — no live duplicate, but future history loads reconstruct the node.
- **Bug 1 — cross-channel history truncated**: `global_output_source` bailed as soon as the local output was non-empty. For per-build/portable channels the local file accumulates only the current session; the global zone has the full record. Removing the bail makes the global zone always preferred, matching its role as the canonical history store.

## Test plan

- [ ] Open a persistent agent (e.g. Naki) — full conversation history loads (not just the most recent session)
- [ ] Send a message to a persistent agent — the message appears immediately in the pane (amber → blue promotion)
- [ ] Reload the pane after sending one or more messages — sent messages are present in history, not just the agent's responses
- [ ] `cargo test -p agentmux-srv -- global_output_source` — 4 tests pass (one renamed to reflect new invariant)
- [ ] Archived-block sessions still open in archive state (suppression guard in `global_output_source` unchanged)


a5af:
@codex review

PR #1494, opened 2026-06-16
feat(host): debounced memory-pressure detection + observability (mem_pressure)

AgentA-asaf:
## Why

Foundation slice of `docs/specs/SPEC_MEMORY_PRESSURE_SUPERVISION_2026_06_16.md` §5.A, following the merged P0 (#1493, memory-aware host *relaunch*). The 2026-06-16 crash had `avail_page_gb ≈ 0` for ~20 min with **no pressure signal** — the host saw the numbers (the 20s heartbeat) but never turned them into an actionable level. This adds that.

## What

- **New `memory_pressure` module (pure + unit-tested):** a hysteresis classifier over commit-free → `PressureLevel { Normal, Warn, Critical }`.
  - Enter **Warn** below 1 GB, **Critical** below 512 MB (aligned with the spec's `WARN_FLOOR` / `RESUME_FLOOR`).
  - **Hysteresis:** leaving a band requires recovering 256 MB past the enter threshold, so a reading parked at a boundary can't flap the log/banner.
  - Publishes a `PRESSURE_LEVEL` atom (`current_level()`) for the proactive-shedding responses (warm-pool drain, CDP purge, banner — later slices) to read lock-free.
- **`memory_heartbeat`** feeds the tracker each 20s sample and emits a structured **`mem_pressure`** warn line (level + `commit_free_mb`) **only on a transition**.

## Scope discipline

**Detection + observability only** — no action on renderers, windows, or the warm pool. Those responses mutate delicate CEF subsystems (window lifecycle / renderer process model) and need runtime verification under real memory pressure, so they land in their own verifiable slices. The classifier is pure, so this PR is **zero behavior change** to anything but the log.

## Tests

5 unit tests (band entry from Normal, change-only emission, Warn/Critical boundary hysteresis, Critical→Warn→Normal step-down). `cargo build` + `cargo test -p agentmux-cef` green.

## Sequence

Part of driving the memory-pressure spec to completion. Next verifiable slices: proactive shedding (CDP `Memory.simulatePressureNotification`/`forciblyPurgeJavaScriptMemory` + warm-pool drain) and the low-memory banner — each gated on a runtime memory-pressure smoke before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1493, opened 2026-06-16
feat(launcher): memory-aware host relaunch — wait out system OOM instead of crash-looping (P0)

AgentA-asaf:
## Why

`docs/retro/retro-oom-crash-2026-06-16.md`: the v0.44.1 host hard-crashed with Chromium's intentional OOM exception **`0xe0000008`** under **system commit-limit exhaustion** (`errno 1455`). The launcher's relaunch ladder (`HOST_RESTART_BUDGET` 3/60s) is **memory-blind** — it relaunched straight back into the same starved condition, burned the budget in seconds, and the user's window vanished with no signal.

This is **P0** of `docs/specs/SPEC_MEMORY_PRESSURE_SUPERVISION_2026_06_16.md` — the **host/instance-level** complement to the existing renderer-level `SPEC_GATED_RENDERER_RECOVERY_2026_06_01.md` (whose §9 explicitly leaves the host-relaunch layer out of scope).

## What

**New `mem_supervisor` module (pure + unit-tested):**
- `classify_host_exit()` splits a **system-OOM** exit from a genuine host fault. OOM is detected two ways (defensive — Chromium sometimes reports OOM as a generic crash, per electron#40426): the exact `0xe0000008` exit code, **or** *any* abnormal exit taken while commit-free was already below `RESUME_FLOOR`.
- `commit_free_mb()` probe (`GlobalMemoryStatusEx` on Windows, `/proc/meminfo` on Linux, fail-open elsewhere), `next_backoff()` (exp, capped 30s), `budget_exhausted()` windowed budget, `await_commit_recovery()` (wait until commit clears the floor, 5-min deadline).

**Both supervisor loops (`run_windows` / `run_unix`) gain a `SystemOom` arm:**
- **Commit-gated, backed-off relaunch** drawn from a **separate** OOM budget (`OOM_RESTART_BUDGET` 5 / 10 min) — *not* the wedged-host budget. Relaunching into a starved system just re-OOMs; waiting is the only lever.
- Relaunch **degraded** (`--disable-gpu`) since the GPU process is a large commit consumer.
- The existing wedged-host **`Abnormal` path is byte-for-byte unchanged** — a deterministic host bug still trips the fast 3/60s budget + degraded ladder exactly as before.

[...]

PR #1496, opened 2026-06-16
feat(host): proactive --disable-gpu at startup under critically-low commit

AgentA-asaf:
## Why

`SPEC_MEMORY_PRESSURE_SUPERVISION_2026_06_16` §5.H, building on the merged P0 (#1493, memory-aware *relaunch*) and the merged detection foundation (#1494). On a **fresh** launch the GPU process spawns unconditionally; when commit is already critically low (the 2026-06-16 condition — several instances + dev tooling saturating the commit pool), that GPU process can be the allocation that tips commit over the edge and OOM-aborts the host before first paint.

## What

In `on_before_command_line_processing` (**browser process only** — only it decides whether to spawn the GPU process), probe `memory_heartbeat::commit_free_mb()`; if below **512 MB**, append `disable-gpu` (software rendering) and log a `mem_pressure` warn line.

- Reuses the exact switch the launcher's degraded-relaunch rung already uses, so the software-rendering path is proven.
- On non-Windows `commit_free_mb()` is `u64::MAX` → never fires (commit-limit exhaustion is a Windows concern).
- Complements #1493: that adds `--disable-gpu` only on a degraded *relaunch*; this is the proactive *first-launch* guard.

Low-risk: a single conditional that reuses an already-exercised flag; no change to the multi-process model or window/renderer lifecycle. `cargo build -p agentmux-cef` green.

## Sequence note

This is the last of the **blind-safe** memory-pressure slices. The remaining responses (warm-pool drain, CDP renderer purge, low-memory banner, Job-Object soft limit, full Reopen dialog) mutate delicate CEF window/renderer/launcher internals and need a smoke under **real** memory pressure before merge — they won't be blind-merged.

### DIFF
--- a/.changesets/1781649131-feat-host-proactive-disable-gpu-at-startup-when-commit-is-cr-we0r.md
+++ b/.changesets/1781649131-feat-host-proactive-disable-gpu-at-startup-when-commit-is-cr-we0r.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(host): proactive --disable-gpu at startup when commit is critically low
--- a/agentmux-cef/src/app.rs
+++ b/agentmux-cef/src/app.rs
@@ -535,10 +535,35 @@ wrap_app! {
     impl App {
         fn on_before_command_line_processing(
             &self,
-            _process_type: Option<&CefString>,
+            process_type: Option<&CefString>,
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1497, opened 2026-06-16
feat(pane.open): OpenEditor collapse_tree + floating-pane support

a5af:
## What

Extends the `OpenEditor` MCP tool (and the underlying `pane.open` RPC / `POST /api/v1/pane/open`) with two options:

- **`collapse_tree`** — open the editor with its file-tree sidebar collapsed (just the file, no explorer). Writes `editor:tree_expanded=false` into block meta; the frontend `EditorViewModel` already restores it on init, so **no frontend change**.
- **`floating`** — open the file in a floating window instead of a docked split, reusing the existing tear-off floating-pane pipeline.

## How (floating)

srv can't open OS windows — only the host can, driven by the frontend. So `open_pane_floating`:

1. Creates the block via the **reducer `CreateBlock`** (not `wcore::create_block`) so it lands in `state.blocks` — the `tear_off_block` saga's precondition checks the reducer-canonical block map; `BlockCreated` also mirrors meta into the wstore Block so the editor renders. **Layout placement is skipped** so it never renders docked (no flash).
2. Runs the existing **`tear_off_block` saga** (`CreateWorkspace`→`CreateTab`→`MoveBlock`) into a fresh floating workspace + `setup_torn_off_block_layout`.
3. Broadcasts `waveobj:update`s, then publishes a **window-scoped `openfloatingpane` directive via the WPS broker**.
4. The source window's frontend (`global.ts`, scoped to `windowId` like `userinput`) handles it and calls the existing host command **`open_floating_pane_window`** → `FloatingPaneWorkspace`.

No new host command, reducer `Event` variant, or saga.

## Verification

- `cargo check -p agentmux-srv -p agentmux-mcp` — green.
- `tsc --noEmit` — clean for `global.ts` (project-wide pre-existing TS errors untouched).
- ⚠️ The **floating** srv→frontend→host cross-process path still needs a `task package`/`task dev` (CEF) **smoke test** — it can't be unit-tested.

Spec: `docs/specs/SPEC_OPENEDITOR_FLOATING_AND_COLLAPSED_TREE_2026_06_16.md`

### DIFF
--- a/.changesets/1781649151-feat-pane-open-openeditor-collapse-tree-floating-pane-suppor-oazj.md
+++ b/.changesets/1781649151-feat-pane-open-openeditor-collapse-tree-floating-pane-suppor-oazj.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(pane.open): OpenEditor collapse_tree + floating-pane support
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -76,7 +76,9 @@ const OPEN_EDITOR_TOOL: &str = r#"{
     "properties": {
       "file":  { "type": "string", "description": "Absolute path to the file to open" },
       "title": { "type": "string", "description": "Optional tab/pane title (defaults to the file name)" },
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #1499, opened 2026-06-16
fix(muxbus): register persistent agents for Tier-1 delivery (closes #1470)

a5af:
Closes #1470.

## Problem
Two live **persistent** Claude panes in the same instance can't muxbus-message each other — `SendMessage → Naki` returns **`agent not found`**.

muxbus Tier-1 delivery (`ReactiveHandler::inject_message`) already dispatches on controller type via the `MessageSender`: persistent (stream-json) agents get a `{type:"user"}` NDJSON line (`persistent.rs::send_user_message`), shell agents get PTY keystrokes (`handler.rs:262-318` + `blockcontroller/mod.rs:305`). **But** only the PTY shell controller ever *registered* an `agent_id → block_id` mapping (`shell.rs:716-747`). Persistent/no-PTY controllers never did, so the lookup at `handler.rs:227` failed before reaching the dispatch. The delivery half of #1470 was built; the **registration** half was missing.

## Fix
Register persistent agents from `persistent.rs::spawn_process` once stdin is live, and unregister in **both** `process_waiter` exit arms — mirroring the shell controller's auto-register/teardown and reusing the same reactive-handler + Tier-2 file-registry calls (`get_global_handler().register_agent/unregister_block`, `reactive::registry::write/remove`).

- **Address** = the agent's display name (`AGENTMUX_AGENT_ID` = `agent.name`; `WAVEMUX_AGENT_ID` legacy fallback), lowercased on register (`handler.rs:112`) — so a sender addressing the agent by name resolves regardless of controller type.
- New pure helper `muxbus_agent_id_from_env()` for id resolution, with unit tests.
- No-op for non-agent persistent blocks (no `AGENTMUX_AGENT_ID` → not registered → unchanged).

## Verification
- ✅ `cargo check -p agentmux-srv --tests` clean (no new errors/warnings).
- ✅ New unit tests pass: `muxbus_registration_tests` (4 tests — canonical/legacy/precedence/blank).
- ⚠️ **Live end-to-end not run from the authoring env.** The acceptance test — `SendMessage → Naki` lands as a `{type:"user"}` turn in Naki's pane, and `GET /agentmux/reactive/agents` lists the persistent agent — requires a build carrying this change in the instance where both persistent panes run. The registration path is a line-for-line mirror of the already-proven PTY shell path, and the delivery path it feeds is already tested (`reactive/tests.rs:330-442`).

## Scope / follow-ups
[...]

PR #1500, opened 2026-06-16
feat(muxbus): unified agent-discovery endpoint + DiscoverAgents tool

a5af:
Follow-up to #1499 (persistent-agent delivery). Together they close the "directory ≠ delivery registry" gap: an agent can now *see* what exists and what's actually reachable, and (with #1499) persistent panes are reachable.

## What
A unified, agent-facing discovery surface across the muxbus tiers.

### `GET /agentmux/discovery` (authed)
```jsonc
{
  "host": {
    "version": "0.46.0",
    "local_url": "http://127.0.0.1:PORT",
    "addressable": [ /* reactive_handler.list_agents() — Tier-1/2 reachable NOW (carries block_id) */ ],
    "agents":      [ { "name": "...", "id": "...", "definition_id": "...",
                       "working_directory": "...",
                       "addressable": true,
                       "block_id": "blk_..." /* live block when addressable, else null */ } ]
  },
  "lan": [ /* LanInstance[] — Tier-3 mDNS peers + their agents */ ],
  "wan": { "subscribed_agents": [ /* Tier-4 cloud subscriptions */ ] }
}
```
- **`host.addressable`** is the authoritative reachable set (every entry has a live `block_id`).
- **`host.agents`** is this host's directory (SQLite `instance_list`), each flagged `addressable` iff its name is in the reachable set (case-insensitive). `block_id` is sourced from the reachable set — the real live delivery block when addressable, `null` otherwise. (The consolidated `instance_list` path doesn't carry block_id/status, so those aren't faked here.)
- Registered in the authed route block next to `/api/lan-instances`; agents reach it via `AGENTMUX_LOCAL_URL` + `X-AuthKey`.

### `DiscoverAgents` MCP tool
First-class tool in `agentmux-mcp` (mirrors `SendMessage`): GETs the endpoint and returns the JSON, so an agent can enumerate valid targets *before* `SendMessage` instead of guessing.

### Plumbing
One new public accessor — `CloudSubscriber::subscribed_agents()` (read-only snapshot). No AppState changes.

## Why
This is the gap that made testing muxbus painful: `Naki` was visible in "My Agents" (directory) yet `SendMessage → Naki` returned `agent not found` (delivery registry). `DiscoverAgents` surfaces that distinction directly via the `addressable` flag.

## Verification
- ✅ `cargo check -p agentmux-srv -p agentmux-mcp` clean.
- ✅ reagent P1/P2 addressed (block_id now live-or-null, dead hidden-filter removed).
- ⚠️ Live response not captured from the authoring env. Once running: `curl -s -H "X-AuthKey: $AGENTMUX_AUTH_KEY" "$AGENTMUX_LOCAL_URL/agentmux/discovery" | jq`. With #1499 also merged, persistent panes show `addressable: true`.

## Notes
- Design doc (`SPEC_MUXBUS_AGENT_DISCOVERY_AND_PERSISTENT_DELIVERY_2026_06_16.md`) lands with #1499.
- Known upstream gap (out of scope): `LanInstance.agents` is currently populated empty on the advertise/resolve path, so LAN peers list but their agent arrays may be empty until that's filled.

### DIFF
--- a/.changesets/1781651198-feat-muxbus-unified-agent-discovery-endpoint-discoveragents--wzms.md
+++ b/.changesets/1781651198-feat-muxbus-unified-agent-discovery-endpoint-discoveragents--wzms.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxbus): unified agent-discovery endpoint + DiscoverAgents tool (host/LAN/WAN, addressable flag)
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -68,6 +68,15 @@ const SEND_MESSAGE_TOOL: &str = r#"{
   }
 }"#;
 
+const DISCOVER_AGENTS_TOOL: &str = r#"{
+  "name": "DiscoverAgents",
+  "description": "List the agents and instances reachable from here across the muxbus delivery tiers (host, LAN, cloud), so you can pick a valid target before SendMessage. Returns JSON: host.addressable (agents reachable right now via local delivery), host.agents (this host's agent directory, each with an `addressable` flag), lan (LAN peers and the agents on them), and wan.subscribed_agents (cloud-subscribed agents). Addressing is by agent name, case-insensitive. Takes no arguments.",
+  "inputSchema": {
+    "type": "object",
+    "properties": {}
+  }
+}"#;
+
 const OPEN_EDITOR_TOOL: &str = r#"{
   "name": "OpenEditor",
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
Addressed both findings in f4476396:
- **P1** (block_id/status always empty): dropped `status`; `block_id` is now sourced from the reactive reachable set — the real live delivery block for an addressable agent, `null` otherwise. No more fields that `instance_list(None,None)` can't populate.
- **P2** (dead `display_hidden` filter): removed; hidden/template rows are already excluded by `user_hidden = 0 AND is_template = 0` in the SQL, and the comment was misleading.

PR description schema updated to match.

PR #1501, opened 2026-06-16
feat(host+ui): low-memory warning banner driven by mem_pressure

AgentA-asaf:
## Why

`SPEC_MEMORY_PRESSURE_SUPERVISION_2026_06_16` §5.F — the **user-facing half** of the merged detection (#1494). It warns the user *before* the memory cliff instead of after, so they can free memory (the single most reliable lever — close a window or another app) while it still helps. On 2026-06-16 commit-free sat near zero for ~20 min with **no signal** before the host OOM-crashed; this surfaces exactly that.

Tracked in [Discussion #943](https://github.com/agentmuxai/agentmux/discussions/943); the lowest-risk slice of the memory-pressure response set.

## What

**Host** (`agentmux-cef`): on a `mem_pressure` level transition the heartbeat posts a `memory-pressure` event (level + `commit_free_mb`) to the frontend via the existing `emit_event_to_top_level_windows` (the same path `pool:promote` uses), wrapped in a UI-thread task (`ui_tasks::post_memory_pressure`) because the heartbeat runs on a background `std::thread`. `memory_heartbeat::start` now takes the `AppState`.

**UI**: a non-modal, dismissible, **app-wide** banner (`memory-pressure-banner.tsx`, mounted in `app.tsx` next to the existing `BundleManagerElsewhereBanner`):
- amber on **Warn** (< 1 GB commit-free), error-tinted on **Critical** (< 512 MB), cleared on the return to Normal.
- **Dismiss is sticky per severity** — dismissing at Warn keeps it hidden until pressure escalates to Critical or a fresh episode begins, so it never nags while the user is already acting.

## Risk

**Purely informational** — unlike the warm-pool drain / CDP purge, this response *never touches renderers or windows*. It only renders a banner. That's why it's the safe first response to ship.

## Verification

- `cargo build -p agentmux-cef` ✓ · `task build:frontend` ✓ · `tsc` clean on touched files.
- **6 unit tests** over the pure show/dismiss/escalation logic (`shouldShow` / `severity`).
- The banner renders independently of real pressure — exercisable from DevTools:
  ```js
  window.dispatchEvent(new CustomEvent('agentmux-event', { detail: { event: 'memory-pressure', payload: { level: 'warn' } } }))
  ```
- The host-emit-at-threshold is established by #1494's tested tracker; the full under-real-pressure path is observable in the wild (per the deferred-verification decision in Discussion #943).

### DIFF
--- a/.changesets/1781651220-feat-host-ui-low-memory-warning-banner-driven-by-mem-pressur-6pd8.md
+++ b/.changesets/1781651220-feat-host-ui-low-memory-warning-banner-driven-by-mem-pressur-6pd8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(host+ui): low-memory warning banner driven by mem_pressure
--- a/agentmux-cef/src/main.rs
+++ b/agentmux-cef/src/main.rs
@@ -815,8 +815,10 @@ fn main() {
     macos_menu::install_menu_bar(app_state.clone());
 
     // Start memory heartbeat — logs system/process memory stats every 20s.
-    // Provides forensic data if the process later crashes from OOM / VA exhaustion.
-    memory_heartbeat::start();
+    // Provides forensic data if the process later crashes from OOM / VA
+    // exhaustion, and drives the debounced mem_pressure level + low-memory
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #1503, opened 2026-06-16
fix(agent): apply model/effort changes to running persistent (Claude) agents

a5af:
## Symptom

Changing the model (e.g. Opus → Sonnet) — or effort / permission-mode — on a **running Claude agent** silently does nothing; it keeps responding on the old model. "It was working before."

## Root cause

Model/effort/permission are **spawn-time CLI flags** (`--model`, `--effort`, …) built from `block.meta["agent:runtime"]`.

- **Subprocess** controllers (codex/gemini/…) re-spawn a fresh process **per turn** with `--resume`, so a change applies next turn — the handlers' *"applies to next turn"* wording is true there.
- **Persistent** controllers (Claude stream-json) keep **one process alive** across turns (`persistent.rs`); turns just write to stdin. The frontend still rewrites `cmd:args` and `agent:runtime`, but **nothing restarts the live process**, so the new `--model` never lands.

## How it regressed

**PR #1451 (`a05060b0`, 2026-06-15)** — *"answer AskUserQuestion via the Agent SDK control protocol"* — flipped Claude from `subprocess` → `persistent` (`providers/index.ts:170`, confirmed via `git blame`). That silently invalidated the *"applies on next turn"* assumption the runtime-change path relies on, and the runtime path was never updated to restart persistent controllers. So `/model` (and the inline picker) have no-op'd on Claude since 2026-06-15.

**Live evidence** (this Naki pane, srv v0.46.0): block spawned **once** at 21:35:38 with `--model opus`; later `SetMeta agent:runtime` + `cmd:args` at 23:04–23:05, but **no respawn**, no `sonnet` anywhere in the srv log, live model still `claude-opus-4-8`.

## Fix

- **Frontend** (`commands/global/runtime.ts`): on a runtime change for a persistent controller, rebuild `cmd:args` (the same `buildRuntimeArgs` the per-turn path uses) and call `ControllerResync{forcerestart:true}`. `resync_controller` swaps in a fresh controller instance — no process-waiter race.
- **Server resume parity** (`persistent.rs` + `websocket.rs` + `app_api.rs`): `PersistentSpawnConfig` gains `resume_flag` + `session_id` (read from `agent:resume_flag` / `agent:sessionid` meta — the keys the subprocess path already reads). `spawn_process` hydrates the session id and appends `--resume <sid>` on respawn — a verbatim mirror of `SubprocessController::spawn_turn`. So the restart respawns Claude with the new `--model` **and** resumes the same conversation (no context loss).

Subprocess behavior is unchanged.

## Verification

- `cargo check -p agentmux-srv` green; `tsc --noEmit` clean for `runtime.ts`.
- ⚠️ Needs a `task dev`/`task package` (CEF) smoke test: switch `/model` on a running Claude pane → next turn runs on the new model and the conversation continues.

Analysis: `docs/analysis/ANALYSIS_PERSISTENT_AGENT_MODEL_SWITCH_REGRESSION_2026_06_16.md`

### DIFF
--- a/.changesets/1781652454-fix-agent-apply-model-effort-permission-changes-to-running-p-wbrq.md
+++ b/.changesets/1781652454-fix-agent-apply-model-effort-permission-changes-to-running-p-wbrq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): apply model/effort/permission changes to running persistent (Claude) agents via resume-preserving restart
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -50,6 +50,17 @@ pub struct PersistentSpawnConfig {
     pub working_dir: String,
     pub env_vars: HashMap<String, String>,
     pub session_id_field: String,
+    /// Resume flag for this provider (e.g. "--resume"), read from
+    /// `agent:resume_flag` meta. Empty = provider has no simple-flag resume.
+    /// Mirrors `SubprocessSpawnConfig::resume_flag` so a respawn (after a
+    /// runtime/model change or the picker reattach path) continues the same
+    /// conversation instead of starting fresh.
+    pub resume_flag: String,
+    /// Session id to hydrate `inner.session_id` with BEFORE spawning, when the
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1505, opened 2026-06-17
fix(agent): seed PTY size at spawn to fix initial agent-pane resize race

a5af:
## Problem

Loading a new agent pane intermittently logged **`resize to N cols failed after 3 attempts`** (`[pty]`, warn). The agent pane is a custom UI (not xterm.js), so `usePtyWidth` pushes the pane width to the PTY over `controllerinput` — but the initial send raced controller startup and rejected with "controller is not running". Two root causes:

- **`ShellController::start()` opened the PTY at a hardcoded `200x25`** and relied on a post-spawn resize RPC to correct it. That RPC races the controller coming up and can fail outright → the agent CLI / tools wrap at 200 cols for the whole session.
- **`publish_status()` broadcast `"running"` before `input_tx` existed**, so even a frontend that waited for the `"running"` event could still hit the reject.

## Fix

- **Seed the PTY at spawn** using the existing `rtopts.termsize` plumbing (the resync command already carries `rtopts` end-to-end). `launch-flow` Phase 3 computes the pane size and the backend opens the PTY at it (`pty_size_from_rt_opts`, clamped, fallback `200x25`). The post-spawn resize is now a correction, not a load-bearing race.
- **Make `"running"` truthful**: move `publish_status()` to after `input_tx` is set.
- **Gate the frontend initial send** on readiness (`controllerstatus: "running"` + a one-shot `GetControllerStatus` probe for re-mounts), coalescing the latest width and clearing readiness on `"done"` (the agent process is per-turn). Retries are now classified (only controller-not-ready is retried) and jittered.

## Tests
- `pty_size_from_rt_opts` unit tests (absent / serde-default / honored / cols-only / clamp / garbage).
- New `usePtyWidth.test.ts`: cols math, readiness gate (defer→flush, already-running re-mount), retry classification.
- `useAgentFailure.test.ts` re-run green — the `publish_status` move preserves `running`/`done` event ordering.

## Verification
- ✅ `task dev` builds (backend + host + bundle) and launches.
- ⏳ Live before/after (PTY opens at pane width, no warning) not yet captured — needs an agent pane opened in the running app; can attach once confirmed.

## Notes / follow-ups
- The runtime-resize path (`shell.rs`) still does an unchecked `i64 as u16` cast — independent hardening.
- Only the launch-flow resync passes `termsize`; other resync callers (restart paths) could too.

Analysis: `docs/analysis/AGENT_PANE_PTY_RESIZE_RACE_2026_06_16.md`.

### DIFF
--- a/.changesets/1781651181-fix-agent-seed-pty-size-at-spawn-so-the-agent-pane-stops-hit-k1k7.md
+++ b/.changesets/1781651181-fix-agent-seed-pty-size-at-spawn-so-the-agent-pane-stops-hit-k1k7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): seed PTY size at spawn so the agent pane stops hitting 'resize to N cols failed after 3 attempts'
--- a/agentmux-srv/src/backend/blockcontroller/shell.rs
+++ b/agentmux-srv/src/backend/blockcontroller/shell.rs
@@ -37,7 +37,7 @@ use crate::backend::eventbus::EventBus;
 use crate::backend::shellexec::{ConnInterface, ShellProc};
 use crate::backend::storage::filestore::FileStore;
 use crate::backend::storage::store::Store;
-use crate::backend::obj::{self, MetaMapType};
+use crate::backend::obj::{self, MetaMapType, RuntimeOpts};
 use crate::backend::wps;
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #1507, opened 2026-06-17
fix(agent-pane): persistent controllers accept the stop signal (SIGINT/SIGTERM)

a5af:
## Bug

Stopping a **Claude agent** (any persistent-controller agent) always fails with:

> stop failed: persistent controller does not accept raw input; use send_message()

## Root cause

The Stop button / Esc sends an interrupt as a **SIGINT signal** via `ControllerInputCommand({signame:"SIGINT"})` (`useAgentCommands.ts:494`), which the backend turns into `BlockInputUnion::signal("SIGINT")` and routes to `controller.send_input()` (`websocket.rs:2316`).

- `SubprocessController::send_input` handles `SIGINT`/`SIGTERM` → `stop_subprocess(true)` ✓
- `PersistentSubprocessController::send_input` rejected **all** input → `Err("… does not accept raw input; use send_message()")` ✗

The signal case was never wired, even though the persistent controller owns a working kill path (`stop_process(true)`, used by its `stop()`).

## Fix

Handle `SIGINT`/`SIGTERM` in `PersistentSubprocessController::send_input` by routing to `stop_process(true)` — the same path `stop()` uses, mirroring `SubprocessController`. Raw (non-signal) input is still rejected. The `session_id` is retained, so the next message resumes the conversation (kill-the-turn semantics, matching subprocess).

## Verification

`cargo check -p agentmux-srv` — clean (pre-existing warnings only).

### DIFF
--- a/.changesets/1781662115-fix-agent-pane-persistent-controllers-accept-sigint-sigterm--hp3p.md
+++ b/.changesets/1781662115-fix-agent-pane-persistent-controllers-accept-sigint-sigterm--hp3p.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): persistent controllers accept SIGINT/SIGTERM so Stop/Esc interrupts Claude (stream-json) agents instead of failing with 'does not accept raw input'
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1508, opened 2026-06-17
feat(agent-pane): ArrowUp/ArrowDown recall previously sent messages (shell-style history)

a5af:
## What

Pressing **ArrowUp** in the agent composer now brings back **old messages** — shell-style history. ArrowDown walks forward again toward your live draft.

## Behavior

- **ArrowUp** on an empty composer:
  1. first un-queues the most-recently queued "send now" held message (existing gesture), then
  2. once nothing is queued, recalls the previous **sent** message; repeated ArrowUp goes further back.
- **ArrowDown** while navigating walks toward the present; past the newest entry it restores the in-progress draft that was stashed when navigation began.
- A manual edit (or sending) resets the cursor, so the next ArrowUp starts from the newest message.
- Caret-on-first/last-line guards keep multiline editing of a recalled message moving line-by-line, only crossing into history at the top/bottom edge.

## Implementation

All in `frontend/app/view/agent/components/AgentFooter.tsx` (the composer):
- Per-pane, in-memory history captured in `handleSend` (capped at 200, consecutive duplicates skipped). The component body runs once in SolidJS, so the closure `let`s persist for the footer's lifetime — same pattern as the existing voice/typing flags.
- A `setComposerValue` helper writes the uncontrolled textarea directly (no `input` event dispatch), reused by both queued-recall and history.
- The existing `onRecallLatestQueued` (held-message un-queue) is preserved and takes precedence on an empty composer.

No backend or RPC changes. Scoped entirely to the composer keydown path.

## Notes

- History is per-pane and in-memory (resets if the pane is closed/reopened) — matches shell-session semantics. Persisting across remounts could be a follow-up.

## Verification

`tsc --noEmit`: clean (31 pre-existing baseline errors unchanged, none in `AgentFooter.tsx`).

### DIFF
--- a/.changesets/1781662243-feat-agent-pane-arrowup-arrowdown-recall-previously-sent-mes-46g4.md
+++ b/.changesets/1781662243-feat-agent-pane-arrowup-arrowdown-recall-previously-sent-mes-46g4.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): ArrowUp/ArrowDown recall previously sent messages in the composer (shell-style history)
--- a/frontend/app/view/agent/components/AgentFooter.tsx
+++ b/frontend/app/view/agent/components/AgentFooter.tsx
@@ -698,13 +698,11 @@ export const AgentFooter = (props: AgentFooterProps): JSX.Element => {
                 }
             }
 
-            // Older: enter from an empty composer, or continue while navigating
-            // with the caret on the first line.
-            if (
-                e.key === "ArrowUp" &&
-                histPos > 0 &&
-                (empty || (navigating && caretOnFirstLine))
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

a5af:
Addressed both Codex P2s in 80936581:
1. **History from non-empty drafts** — ArrowUp now enters history whenever the caret is on the first line (not just when empty); the live draft is stashed into `histDraft` and restored on ArrowDown.
2. **Reset cursor on Esc-clear** — the Esc clear path now resets `histPos`/`histDraft` (it sets `value=""` without dispatching `input`, so handleInput's reset wouldn't fire). Next ArrowUp no longer skips the newest message.

PR #1509, opened 2026-06-17
fix(term): enable reverse-wraparound so Backspace crosses soft-wrapped lines

AgentU-asaf:
## Problem
On **Linux/macOS**, pressing **Backspace at column 0 of a soft-wrapped row** does nothing — the cursor is stuck and won't delete back into the previous row. **Windows is fine.**

## Root cause (evidence-based)
- Captured from a PTY: bash/readline emits a **bare `BS` (`0x08`) + `CSI K`** for the cross-wrap delete — identical bytes to a same-line backspace. It relies on the terminal doing **reverse-wraparound** (`infocmp xterm-256color` confirms **no `bw`** capability).
- xterm.js v6 gates reverse-wraparound behind **DEC private mode 45**, default **OFF**. Its `backspace()` clamps at column 0 when the mode is off:
  ```js
  backspace(){ if(!decPrivateModes.reverseWraparound) return _restrictCursor(), x>0 && x--, true; ... }
  ```
- **Platform split:** on a raw PTY (Linux/macOS) the bare `0x08` reaches xterm.js and clamps. On Windows, **ConPTY re-renders with absolute cursor positioning**, so xterm.js never receives the bare BS — it just works.

## Fix
`frontend/app/view/term/termwrap.ts`: assert `\x1b[?45h` (DECSET reverse-wraparound) on terminal init, and **re-assert after RIS** (`reset`/`tput reset` clears DEC private modes). xterm.js's enabled branch only wraps when the row is genuinely `isWrapped`, so **hard line breaks are never crossed**. No effect on Windows (ConPTY path doesn't need it).

## Verification
Confirmed in a live dev terminal on Linux: Backspace across a soft wrap now pulls the cursor up to the end of the previous row. Frontend-only; type-clean.

### DIFF
--- a/.changesets/1781691242-fix-term-enable-reverse-wraparound-so-backspace-crosses-soft-2fn6.md
+++ b/.changesets/1781691242-fix-term-enable-reverse-wraparound-so-backspace-crosses-soft-2fn6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): enable reverse-wraparound so Backspace crosses soft-wrapped lines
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1510, opened 2026-06-17
fix(agent-pane): persistent controller no-ops termsize resize

a5af:
## Problem

The agent pane still logs `resize to <N> cols failed ...` in its activity log — even after #1505 ("seed PTY size at spawn"). #1505 fixed the **ShellController** path (the only code that emits `"controller is not running"`), but **Claude/Codex agents don't use it**.

## Root cause

Agent panes run on the **PersistentController** (`controllerType: "persistent"`, e.g. Claude `--input-format stream-json`) — verified live: the sidecar logs `persistent controller registered (spawns on first message)`. Persistent controllers **have no PTY**.

`usePtyWidth` can't tell a PTY-backed controller from a PTY-less one, so it sends a `termsize` resize via `ControllerInputCommand` on every running turn:

- `SubprocessController::send_input` already treats `termsize` as an accepted no-op (`subprocess.rs:1482`).
- `PersistentController::send_input` had **no such branch** — it rejected *all* non-signal input with `Err("persistent controller does not accept raw input; use send_message()")`.

`usePtyWidth` doesn't classify that string as retryable, so it logged a one-shot `resize to N cols failed (not retryable): ...` warning in the pane's "aux bar" on every turn.

## Fix

Mirror `SubprocessController`: reject only raw `input_data`; accept `termsize`/other benign inputs as a no-op. A persistent controller has no PTY, so there is nothing to resize — the resize is meaningless, not an error.

## Tests

Added `send_input_tests` in `persistent.rs`:
- `termsize_resize_is_accepted_noop` — a `termsize` resize returns `Ok`
- `raw_input_is_still_rejected` — raw keystrokes still error

Both pass (`cargo test -p agentmux-srv send_input_tests`).

## Test plan

- [ ] Open a Claude (persistent) agent pane, send a message, confirm no `resize to N cols failed` warning appears in the activity log
- [ ] Stop button / Esc (SIGINT) still interrupts a running persistent turn
- [ ] Sending a normal message still works (routes via `send_message`, not `send_input`)

### DIFF
--- a/.changesets/1781692991-fix-persistent-controller-termsize-noop.md
+++ b/.changesets/1781692991-fix-persistent-controller-termsize-noop.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): persistent controller accepts termsize resize as a no-op so the agent pane stops logging spurious "resize to N cols failed" warnings
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -980,7 +980,21 @@ impl Controller for PersistentSubprocessController {
                 "persistent controller: unsupported signal {sig} (only SIGINT/SIGTERM)"
             ));
         }
-        Err("persistent controller does not accept raw input; use send_message()".to_string())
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1511, opened 2026-06-17
feat(tool-output): tee redirected output to feed + render tool output as terminal

a5af:
Implements `docs/specs/SPEC_TOOL_OUTPUT_TEE_AND_TERMINAL_RENDER_2026_06_17.md` — two independent fixes, "two complaints, one theme."

## F1 — tee a trailing file redirect to the feed (backend)

**Problem:** `task package > build.log 2>&1` sends stdout to the file, so the agent's tool feed shows *nothing* while it runs.

**Fix:** the PreToolUse hook (`agentmux-bashwrap/src/hook.rs`) now recognizes an unambiguous **top-level trailing** redirect and rewrites it to a `tee` pipeline before base64-encoding, so the same bytes reach **both** the file **and** the PTY (hence the live chunk feed) — **zero frontend changes**, reuses the whole existing pipeline.

```
CMD > F        →  set -o pipefail; { CMD ; } | tee -- F
CMD >> F 2>&1  →  set -o pipefail; { CMD ; } 2>&1 | tee -a -- F
CMD &> F       →  set -o pipefail; { CMD ; } 2>&1 | tee -- F
```

A conservative, **quote/group/comment-aware** scanner handles `>`/`>>`/`&>`/`&>>`, trailing `2>&1`, **pipelines** (`a | b > f`), and **subshells** (`(a && b) > f`). Everything ambiguous **bails to today's verbatim behavior**: quotes, comments, heredocs, process substitution, `/dev/null`, fd-specific/dup redirects (`2>f`, `>&2`), the `2>&1 > F` order (different semantics), top-level lists (`&&`/`||`/`;`/`&`), non-trailing, and multiple redirects. `set -o pipefail` preserves the command's exit code through the pipe (bashwrap mirrors it to the model + terminal event).

## F2 — render task/tool output as a terminal, not a JSON blob (frontend)

**Problem:** `Task` / `TaskOutput` / unknown-tool results route to `CompactResult`, whose expanded view is `JSON.stringify(result, null, 2)` — escaped-newline JSON instead of the actual output.

**Fix:**
- new **`TerminalOutput`** renders a result's terminal string body through the existing **`AnsiLine`** (SGR color) in a monospace, scroll-capped panel — reusing `capText`/`MAX_TOOL_OUTPUT_LINES`/`OutputHiddenMarker`. No xterm.js in the feed (§4.3).
- new **`terminalText(result)`** extracts `stdout`(+`stderr`) / `output` / `content`; `null` for purely structured or empty results.
- `CompactResult`'s expanded branch renders the terminal when there's a string body, else falls back to JSON — fixes **Task, Agent, TaskOutput, and all unknown tools at once**.

## Tests / verification

- **`cargo test -p agentmux-bashwrap`**: 40 pass, incl. 9 new F1 cases — every rewrite form + round-trip through `build_response` b64 + every bail path.
- **549 agent-view vitest** pass (3 new suites: `terminal-text`, `TerminalOutput`, `CompactResult`).
- `tsc --noEmit` and `stylelint` clean for touched files (pre-existing baseline noise excluded).
- ⚠️ Not verified headlessly: the live feed actually streaming a redirected command's output, and a Task result rendering as a terminal in the running app — needs `task dev` (per spec §8).

## Notes
- `TerminalOutput` is placed in `view/agent/components/` (not `element/` as the spec sketched) to keep the **view→element** dependency direction correct — it imports the view-layer cap utilities; `AnsiLine` stays in `element/`. Noted in the committed spec.
- Live-chunk **color** (F2 §4.2.4) stays a follow-up — bashwrap strips SGR before publishing chunks today.

### DIFF
--- a/.changesets/1781693591-feat-tool-output-tee-redirected-output-to-the-feed-render-to-stdl.md
+++ b/.changesets/1781693591-feat-tool-output-tee-redirected-output-to-the-feed-render-to-stdl.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(tool-output): tee redirected output to the feed + render tool/task output as a terminal
--- a/agentmux-bashwrap/src/hook.rs
+++ b/agentmux-bashwrap/src/hook.rs
@@ -76,7 +76,15 @@ pub fn build_response(stdin_payload: &str) -> Value {
         return passthrough();
     }
 
-    let b64 = URL_SAFE_NO_PAD.encode(command.as_bytes());
+    // Tee a trailing file redirect back through the PTY so the agent's tool
+    // feed shows the output live while the file is still written. Conservative:
+    // only unambiguous trailing redirects are rewritten; anything else is
+    // encoded verbatim (today's behavior). See
+    // SPEC_TOOL_OUTPUT_TEE_AND_TERMINAL_RENDER_2026_06_17.md §3.
+    let effective_command =
+        tee_redirect_rewrite(command).unwrap_or_else(|| command.to_string());
+
+    let b64 = URL_SAFE_NO_PAD.encode(effective_command.as_bytes());
     let wrapped = format!(
         "{} exec --tool-id={} --b64-cmd={}",
         WRAPPER_BINARY,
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #1512, opened 2026-06-17
fix(trust-center): bind already-authorized CLI OAuth to its brand

a5af:
## Problem

Talking to a running Claude instance means you're **clearly authorized with Anthropic** — yet the Trust Center showed **no** Anthropic connection.

Two namespaces never met. The Default-bundle migration discovers the ambient Claude login (`~/.claude/.credentials.json`) and stores it as an `IdentityAccount` with **`provider = "claude"`** (`kind: oauth`, `SecretRef::OAuthConfigDir`, live-probed `status`). But the Trust Center only knows **brand** ids — `"claude"` isn't in `AccountProvider`, isn't a gallery tile, and isn't in the `accountsByProvider` grouping `order` — so it was **filtered out and invisible**.

## Fix — unify the namespace at the grouping key

- **`provider-brand.ts`** — `brandForProvider()` maps CLI-OAuth ids → brands: `claude→anthropic`, `codex→openai`, `gemini→google`, `copilot→github`. **Display-only** — the account's stored `provider` is unchanged, so spawn-time env injection (keyed by the real CLI id via the resolver) is untouched.
- **`accountsByProvider()`** groups by `brandForProvider(provider)`. Since it feeds **both** the gallery tile count (`AccountsGallery.countFor`) **and** the Connected-accounts list (`AccountsTab`), the detected account now appears under its **Anthropic** tile *and* in the list with its **live status dot** — one change, fixed everywhere.

## Two-way binding
This is the **read** side: the surfaced row is the *same* one the resolver re-probes at every agent spawn (`probe_oauth_status` → status upsert → `identitybundlebindings:changed` broadcast), so the status stays live, not a snapshot. The account is already bound to the Default bundle, so it composes into identities today.

**Follow-ups (noted in the spec):** reconnect-from-tile via the existing `auth.start` flow, and an ambient live-probe RPC for machines where the migration didn't create the row.

## Verification
- `npm run build` — clean.
- `vitest` — identity + accounts pass, incl. new `provider-brand.test.ts` (4 cases).
- Stored `provider` stays `"claude"` → agent env injection unaffected; brand-native (key) accounts group as before.

Spec: `docs/specs/SPEC_TRUST_CENTER_CLI_AUTH_BINDING_2026_06_17.md`.

### DIFF
--- a/.changesets/1781570000-trust-center-cli-auth-binding.md
+++ b/.changesets/1781570000-trust-center-cli-auth-binding.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(trust-center): surface already-authorized CLI OAuth under its brand — a Claude CLI login now shows as a connected Anthropic account (and codex→OpenAI, gemini→Google, copilot→GitHub), with its live probe status, instead of being invisible
--- a/docs/specs/SPEC_TRUST_CENTER_CLI_AUTH_BINDING_2026_06_17.md
+++ b/docs/specs/SPEC_TRUST_CENTER_CLI_AUTH_BINDING_2026_06_17.md
@@ -1,0 +1,61 @@
+# SPEC: Trust Center ↔ CLI-OAuth two-way binding
+
+**Date:** 2026-06-17
+**Status:** Draft — implementation (folded into the PR)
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1514, opened 2026-06-17
feat(agent): tool-result renderer registry + WebSearch cards + record tables (Phases 0–3)

a5af:
Implements `docs/specs/SPEC_TOOL_RESULT_RENDERER_REGISTRY_2026_06_17.md` Phases 0–2: the architecture for rich per-tool result UIs, **plus the first one** (WebSearch).

## The problem
`ToolNode.tool` is a **closed 9-value union** (`…|"Other"`), and `stream-parser.ts:508` collapsed every unknown tool to `"Other"` — so `WebSearch`, `TodoWrite`, and all `mcp__*` tools were indistinguishable, and `renderToolResultBody`'s `switch` could only reach `default` → JSON. You couldn't `case "WebSearch"` over an enum the data can't populate.

## Phase 0 — preserve the raw tool name (no behavior change)
`ToolNode` gains optional **`toolName`** (`"WebSearch"`, `"mcp__github__search_issues"`, …); `stream-parser` sets it at the three node-construction sites. The coarse `tool` enum stays as the *kind*; matchers fall back to it when `toolName` is absent. The data already flowed from every translator — it was just dropped at `normalizeToolName`.

## Phase 1 — renderer registry (pure refactor)
`tool-renderers/registry.ts`: register `{ priority, match, render, label }`, resolve the highest-priority match (ties by registration order). Matchers `byKind`/`byName`/`byNamePrefix("mcp__")`/`byShape`/`anyTool`. `ToolOverlayLog`'s switch is migrated to **registered built-ins** — the former arms **verbatim** at priority 0 + a catch-all (`CompactResult`) — **byte-identical** output.

## Phase 2 — WebSearch result cards (the payoff)
`SearchResults` renders WebSearch results as **clickable cards** (title → `openExternal` to the system browser, host, snippet) instead of a JSON blob. Registered for `WebSearch`/`web_search` by name at priority 10. `extractSearchResults` is **tolerant** (top-level array or `results`/`content`/`items`/… arrays; `url|link|uri`, `title|name|heading`, `snippet|description|text|content|page_age`) and **gracefully falls back to today's JSON** for anything it doesn't recognize — so registering by name can never break an unexpected payload.

#1511's `terminalText`→`TerminalOutput` is the seed of this pattern and is the terminal entry of the fallback chain.

## Tests / verification
- `registry.test.ts` (10): resolution (priority/tie/catch-all/null), matchers, parity (all 8 built-ins register; every kind resolves).
- `search-results.test.ts` + `SearchResults.test.tsx` (11): classifier shapes/variants/rejections; card render + click-to-open (mocked `getApi`) + JSON fallback + registry routing.
- `tsc` + `stylelint` clean for touched files; **570 agent-view vitest pass** (no regression).
- ⚠️ Not verified headlessly: real WebSearch output in the running app. The exact result shape varies by provider; the extractor is deliberately tolerant and no-ops to JSON on unrecognized shapes, so it may need a **real-sample tune** to confirm field coverage. Needs `task dev` + an agent running WebSearch.

## Follow-ups (Phase 3)
`mcp__*` prefix renderers, more kinds (image/table/file-list), and folding in the two *other* tool-keyed dispatches (`TOOL_ICONS` + the ad-hoc AskUserQuestion panel).

### DIFF
--- a/.changesets/1781694790-refactor-agent-tool-result-renderer-registry-foundation-raw--2vn7.md
+++ b/.changesets/1781694790-refactor-agent-tool-result-renderer-registry-foundation-raw--2vn7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(agent): tool-result renderer registry foundation (raw tool name + registry)
--- a/.changesets/1781694790-refactor-agent-tool-result-renderer-registry-foundation-raw--2vn7.md
+++ b/.changesets/1781694790-refactor-agent-tool-result-renderer-registry-foundation-raw--2vn7.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
-refactor(agent): tool-result renderer registry foundation (raw tool name + registry)
+feat(agent): tool-result renderer registry — rich per-tool result UIs (WebSearch cards, record tables) routed by tool name/shape instead of a closed switch
--- a/docs/specs/SPEC_TOOL_RESULT_RENDERER_REGISTRY_2026_06_17.md
+++ b/docs/specs/SPEC_TOOL_RESULT_RENDERER_REGISTRY_2026_06_17.md
@@ -1,0 +1,163 @@
+# SPEC: Tool-result renderer registry (rich, per-tool result UIs that scale)
+
+**Date:** 2026-06-17
+**Status:** Proposed (analysis + design; not implemented)
+**Author:** smike
+**Builds on:** `SPEC_TOOL_OUTPUT_TEE_AND_TERMINAL_RENDER_2026_06_17.md` (PR #1511 — `TerminalOutput` + `terminalText`, the first result-shape classifier)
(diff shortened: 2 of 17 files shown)


a5af:
@codex review

a5af:
Updated:
- **Rebased onto latest `main`** (#1512).
- **Fixed the P2** — hidden WebSearch cards now use `noun="result"` (widened `OutputHiddenMarker`'s union), so the cap reads "… N more **results** hidden" instead of "lines".
- **Phase 3 added** — a generic `RecordTable`: a top-level array of flat records renders as a table instead of a JSON blob, registered **by shape at priority -1** so it only improves the otherwise-JSON unknown-tool path (mcp__*, provider-specific) and leaves the coarse-kind built-ins + name-matched renderers untouched. Conservative classifier (top-level array, flat scalar records, bounded 12 cols / 200 rows; anything else stays JSON).

(The earlier `[P1] hook.rs` finding was against the pre-rebase version that had wrongly bundled #1511's commit — it's no longer part of this PR. That redirect-scanner concern is real and tracked on **#1511**.)

567 agent-view vitest pass; tsc + stylelint clean.

PR #1517, opened 2026-06-17
fix(agent-pane): control-bar model/effort changes apply to running Claude

a5af:
## Verification finding

Verifying "**does setting the model (Sonnet/Opus) work in AgentMux**" surfaced a real gap:

| Path | Applies to a *running* Claude agent? |
|------|--------------------------------------|
| Launch-time `--model` (`buildRuntimeArgs`) | ✅ (10/10 unit tests) |
| Slash `/model` | ✅ — #1503 rebuilds `cmd:args` + restarts the persistent controller |
| **GUI control-bar Opus/Sonnet dropdown** | ❌ **silently no-op'd** |

#1503 (which fixed the #1451 subprocess→persistent regression) patched the **slash** path in `commands/global/runtime.ts`, but the **control-bar dropdown** has its *own* `updateRuntime` that only wrote `agent:runtime` meta. For a **persistent** (Claude stream-json) controller the model/effort are baked in at spawn, so a meta-only write never changes the live process — picking Opus/Sonnet in the dropdown did nothing until the next respawn.

## Fix — one shared apply path
- **`runtime-apply.ts::applyRuntimeChange`** (new): write `agent:runtime`, and for persistent controllers rebuild `cmd:args` (`buildRuntimeArgs`) + `ControllerResyncCommand({ forcerestart })` (resumes via `--resume`, conversation preserved).
- `commands/global/runtime.ts` (`/model`·`/effort`·`/mode`) now calls the shared helper — de-duplicates the inline #1503 logic.
- `AgentControlBar.updateRuntime` calls `applyRuntimeChange(getProvider(providerId), …)` instead of the meta-only write — **fixes the dropdown**.

## Verification
- `npm run build` — clean.
- `vitest buildRuntimeArgs.test.ts` — 10/10 (`--model opus/sonnet`, `--effort` skipped on haiku, codex namespace, stale-flag strip).
- Subprocess providers unaffected (meta write only; they respawn per turn).

### DIFF
--- a/.changesets/1781572000-fix-control-bar-model-apply.md
+++ b/.changesets/1781572000-fix-control-bar-model-apply.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): model/effort/permission picked in the control-bar dropdown now applies to a running Claude agent — previously only the /model slash command did (the GUI write was meta-only and silently no-op'd for persistent controllers)
--- a/frontend/app/view/agent/commands/global/runtime.ts
+++ b/frontend/app/view/agent/commands/global/runtime.ts
@@ -16,11 +16,8 @@
  * (current)`, and Enter is a no-op confirmation.
  */
 
-import { RpcApi } from "@/app/store/rpc-api";
-import { TabRpcClient } from "@/app/store/rpc-util";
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1523, opened 2026-06-17
fix(agent): working-state shows in-progress on stream resume + phase-guard Esc

a5af:
Ships the two high-value fixes from `docs/analysis/ANALYSIS_AGENT_WORKING_STATE_AND_INPUT_REDUCER_2026_06_17.md` (the analysis from the live "Smark agent stuck/unresponsive" incident).

## 1. The reported bug — "in progress" off while the agent is actively streaming

**Root cause (code-grounded):** during a long upstream model stall (overloaded/rate-limited — observed across all agents simultaneously), the agent CLI is **killed + respawned**, which drops and re-opens the stream. On reconnect the turn phase lands in `Idle`/`Disconnected` (`StreamUnsubscribe → Disconnected`, then `StreamSubscribe → Idle`, PR F). But the live-activity commands only promoted to `Streaming` **from `Submitting`** — so the resumed response streamed into the document while the phase sat at `Idle` → `isWorking()` false → **indicator off while content printed**.

**Fix:** `StreamFlushObserved` and `bumpEvent` (`ToolStart`/`TokensIn`/`TokensOut`) now re-enter `Streaming` from `Idle`/`Disconnected` too. These commands are dispatched **only for live stream content** (never history replay — verified in `useAgentStream`), so observable content is proof the agent is working. `Done` (completed turn) and `Interrupting` (user stopping) are intentionally left alone.

> **Correction:** I first suspected the `StreamStalled` streaming-idle watchdog (`Streaming → Done.errored`). It turns out to be **unwired dead code** — `schedule-stream-watchdog`/`StreamStalled` are emitted but **never consumed** anywhere in production, so the watchdog never fires. Flagged for a cleanup follow-up (delete or wire it).

## 2. Esc edge case (you emphasized meticulous Esc state)

Empty-composer **Esc** called `stopAgent` unconditionally → on an idle/finished/disconnected agent it fired a SIGINT that landed nowhere and logged a spurious **"stop failed"**. **Fix:** `stopAgent` no-ops when the phase isn't working (`workingFromPhase`). Stop still works in `Submitting`/`Streaming`/`Interrupting`.

## Tests / verification
- 2 new reducer tests for the reconnect→resume promotion (`StreamFlushObserved`/`ToolStart`/`TokensIn` from `Idle` → `Streaming`). No existing test asserted the old behavior.
- **747 agent + store vitest pass**; `tsc` clean.
- ⚠️ Not verifiable headlessly: the live UI showing the spinner through a real stream-drop/resume — needs `task dev`.

## Deferred (in the analysis)
Indicator deriving from `turnPhase` only (startup-flicker polish), the dead-watchdog cleanup, held-message drain-point, question-panel "Other" IME/Esc, double-submit gate, and a shared alert-surface helper.

### DIFF
--- a/.changesets/1781717402-fix-agent-show-in-progress-when-streaming-resumes-after-a-st-t08o.md
+++ b/.changesets/1781717402-fix-agent-show-in-progress-when-streaming-resumes-after-a-st-t08o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): show 'in progress' when streaming resumes after a stream drop; quiet Esc on idle
--- a/docs/analysis/ANALYSIS_AGENT_WORKING_STATE_AND_INPUT_REDUCER_2026_06_17.md
+++ b/docs/analysis/ANALYSIS_AGENT_WORKING_STATE_AND_INPUT_REDUCER_2026_06_17.md
@@ -1,0 +1,217 @@
+# Analysis: agent-pane working-state desync, input-reducer Esc edge cases, and state cohesion
+
+**Date:** 2026-06-17
+**Author:** smike
+**Status:** Analysis (no code) — findings + prioritized recommendations
+**Trigger:** Live incident — an agent ("Smark") in a 0.46.1 pane showed **not "in progress"
+while it was actively streaming output**. Also covers the conversation input reducer, Esc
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1522, opened 2026-06-17
feat(editor): styled markdown preview with rendered/source toggle

a5af:
## Summary

`.md` files now open **styled by default** in the editor, with a toggle to edit source. Implements your ask: "md files should open in styled version."

**No new dependency, no hand-rolled parser** — reuses the editor stack AgentMux already ships:
- The editor's existing **`frontend/app/element/markdown.tsx`** `<Markdown>` component (the same one the agent conversation pane uses): remark-gfm + remark-flexible-toc, rehype-sanitize (XSS-safe), rehype-raw/slug/highlight + **shiki**, themed via `markdown.scss`.

## Behavior

- Open a `.md` file → renders styled (default).
- **Toggle:** a `👁 Preview` / `✎ Source` button (top-right of the editor body, markdown tabs only) and **Ctrl/Cmd+Shift+V** flip the active tab between rendered and editable CodeMirror source.
- Mode is remembered **per tab**; non-markdown files are unaffected; mode state is cleared on tab close.

## Implementation note

The preview is an **overlay** over the CodeMirror body — CM stays mounted underneath. This:
- preserves cursor / scroll / undo when toggling back to source,
- avoids CodeMirror build-timing races (no unmount/remount on toggle),
- reads `model.contentAtom()` reactively, so edits made in source appear when you switch back to preview.

(Markdown isn't LSP-backed, so keeping CM mounted for a previewed file costs nothing meaningful.)

## Checks

- `tsc --noEmit`: no errors in the changed files.
- `stylelint`: no new violations (added SCSS uses only theme tokens; pre-existing violations elsewhere in the file are untouched).

## Test plan

- [ ] Open a `.md` → shows styled (headings, lists, code blocks highlighted, tables).
- [ ] Click `✎ Source` / press Ctrl+Shift+V → CodeMirror source; edit; toggle back → preview reflects edits.
- [ ] Per-tab memory: one .md in source, another in preview, switch between them.
- [ ] Non-markdown files open in source as before (no toggle button).
- [ ] Ctrl+S still saves; relative-image rendering is best-effort (no resolveOpts wired yet — follow-up if needed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1525, opened 2026-06-17
fix(editor): scratch buffer opens as plain editable text

a5af:
## Problem

Opening a new editor pane creates an untitled **scratch buffer**. It was tagged `language: "markdown"` (the backing file is `{uuid}.md` on disk), so the just-merged styled-preview feature (#1522) opened a fresh, empty scratch as a **blank rendered markdown pane** instead of an editable buffer.

Reported: "when I open a new editor pane I should get a scratch file with no extension, but it's a new md."

## Fix

- **`editor-model.ts`**: the scratch tab's language is now `"text"` — an untitled buffer is for typing, not a markdown document. (The on-disk file stays `.md` for persistence/reuse; that's invisible to the user. Save As `.md` gives full markdown behavior.)
- **`editor-view.tsx`**: `showRendered()` never auto-renders a scratch tab — defensive, so a scratch is always editable even if it were markdown.

Real `.md` files still open styled; non-markdown files are unaffected.

## Notes

- `tsc`: the one error in `editor-model.ts` (rpcCall arg count) is **pre-existing** on origin/main (verified by stashing) — unrelated to this change.

## Test plan

- [ ] New editor pane → editable empty scratch (no blank rendered pane, no preview toggle).
- [ ] Type into it; Ctrl+S / Save As `.md` → becomes markdown.
- [ ] Opening a real `.md` still renders styled with the toggle.

### DIFF
--- a/.changesets/1781723223-fix-editor-scratch-plaintext.md
+++ b/.changesets/1781723223-fix-editor-scratch-plaintext.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(editor): new scratch buffer opens as plain editable text, not a blank rendered markdown pane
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1531, opened 2026-06-17
perf(layout): reduce pane reflow settle window from 220ms to 32ms

a5af:
## Summary

- The 220ms `REFLOW_WINDOW_MS` was originally tracking a CSS pane reflow animation that was removed
- The placeholder rect is now set synchronously, so there's no animation to follow frame-by-frame
- 32ms (≈2 rAF frames) is sufficient for Solid's reactive flush + ResizeObserver to settle before the per-frame HWND sampling loop stops
- This eliminates a ~188ms artificial delay on every pane open/close/split

## Test plan

- [ ] Open a new pane — browser panes should reposition correctly with no lag
- [ ] Close a pane — verify adjacent browser panes snap into position without the old ~200ms settle delay
- [ ] Resize window — steady-state `ResizeObserver` path is unaffected (no regression)
- [ ] Split a pane — HWND settles correctly within the new 32ms window

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.changesets/1781726783-perf-layout-reduce-pane-reflow-settle-window-from-220ms-to-3-kmzz.md
+++ b/.changesets/1781726783-perf-layout-reduce-pane-reflow-settle-window-from-220ms-to-3-kmzz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1513, opened 2026-06-17
fix(ui): pin arrow cursor on scrollbars — they inherit the host cursor

AgentA-asaf:
## Problem

The 2026-06-15 cursor work (#1453 + #1455) did **not** fix the scrollbar cursor — it changed the symptom. The current `main` shows:

- **Main agent-pane conversation scrollbar → text I-beam**
- **Live-tool log scrollbar → link hand**

## Root cause

`cursor` is an **inherited** CSS property, and a WebKit `::-webkit-scrollbar*` pseudo-element takes the **computed cursor of its scroll-host element** — not the OS default. #1453 *deleted* `cursor: pointer` from the thumb on the assumption "no cursor → inherits the arrow." That's false: the scrollbar then inherits the host's cursor.

- Main conversation: `.agent-document { overflow-y: auto; cursor: text }` (`_document.scss:16`) → scrollbar inherits **text**.
- Live-tool log: `.agent-tool-overlay-log` (no cursor) nested under `.agent-tool-block { cursor: pointer }` (`_document-nodes.scss:141`) → scrollbar inherits **pointer**.

Worse, #1455's stylelint guard **forbids `cursor` on any `::-webkit-scrollbar*` selector** — i.e. it banned the only working fix, and its message ("Remove it so the thumb keeps the default arrow (it inherits it)") codified the wrong model.

Full post-mortem: `docs/retro/retro-scrollbar-cursor-regression-2026-06-17.md`.

## Fix

- **`app.scss`** — pin `cursor: var(--cursor-default)` on `*::-webkit-scrollbar`, `-track`, `-thumb`, `-corner`. (OverlayScrollbars handle already pinned it correctly.)
- **`.stylelintrc.json`** — remove the inverted `rule-selector-property-disallowed-list` entry (stylelint can't express the real, value-scoped rule).
- **`scripts/check-scrollbar-cursor.sh`** — new grep gate forbidding `pointer`/`text` (and `--cursor-interactive`/`--cursor-text`) on scrollbar/os-scrollbar selectors while allowing the arrow and cursor-less blocks. Wired into `task build:frontend` (mirrors `check:menu-positioning`).
- **`theme.scss` + analysis doc** — correct the stale "deleting it inherits the arrow" guidance.
- **changeset** — `patch`, no version bump.

## Verification

[...]

PR #1532, opened 2026-06-18
feat(ui): thin scrollbars everywhere except agent-pane main and terminal

AgentA-asaf:
## What

Make **half-width (7px)** the default for scrollbars, and keep **full-width (14px)** only on the two primary reading surfaces:

- **Agent pane main conversation** (`.agent-document`) — stays full.
- **Terminal** — xterm's own overlay scrollbar, already sized independently (`view/term/term.scss`), so no change needed.

Everything else — tool-call logs, the agents list, command palette, identity panels, markdown code/TOC blocks — now gets the thin scrollbar.

## How

- `app.scss`: global `*::-webkit-scrollbar` width/height **14px → 7px**; new `.agent-document::-webkit-scrollbar { 14px }` exception (more specific than the universal rule, so it wins on size only); `.os-scrollbar { --os-size: 7px }` for OverlayScrollbars (library default was 10px).
- Removed the redundant hard-coded `width: 14px` from `command-palette.scss` and the three identity panels (`_accounts`/`_detail`/`_form-overlay`) so they inherit the new thin default instead of restating the old magic number.
- `patch` changeset; no version bump.

## Verification

- Confirmed in `task dev` (HMR): agent main + terminal stay full; tool logs, agents list, command palette, identity panels, markdown blocks are now thin.
- Mechanical: only two webkit-scrollbar widths remain (global 7 / agent-document 14); `--os-size: 7px` set; `scripts/check-scrollbar-cursor.sh` green; touched files lint clean (aside from the unrelated pre-existing `z-index:9999` at `app.scss:48`).

Follow-up to #1513 (scrollbar cursor fix).

### DIFF
--- a/.changesets/1781744050-feat-ui-thin-scrollbars-everywhere-except-agent-main-and-terminal-6956.md
+++ b/.changesets/1781744050-feat-ui-thin-scrollbars-everywhere-except-agent-main-and-terminal-6956.md
@@ -1,0 +1,12 @@
+---
+type: patch
+---
+
+feat(ui): thin scrollbars everywhere except the agent-pane main scroll and terminal
+
+Make half-width (7px) the default for all native scrollbars and OverlayScrollbars
+(`--os-size`), and keep full-width (14px) only on the two primary reading
+surfaces: the agent conversation (`.agent-document`) and the terminal (xterm's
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #1533, opened 2026-06-18
feat(editor): support non-UTF-8 file encodings

a5af:
## Summary

The editor only handled UTF-8 — opening a **Windows-1252 `.ini`**, a **UTF-16** file, **Shift_JIS**, Latin-1, etc. errored (`std::fs::read_to_string` → "stream did not contain valid UTF-8"), and saves always wrote UTF-8 (silently changing a file's encoding). This adds encoding detection + transcoding so a wide variety of encodings open and **round-trip**.

Spec: `docs/specs/SPEC_EDITOR_FILE_ENCODINGS_2026_06_17.md`. (And to the original question: an `.ini` has no fixed encoding — on Windows it's typically Windows-1252/ANSI, sometimes UTF-8/UTF-16; it must be detected from the bytes.)

## What changed

**Backend — `agentmux-srv/src/backend/text_encoding.rs` (new):**
- `decode_file`: BOM sniff → valid-UTF-8 → `chardetng` heuristic → decode via `encoding_rs`. Returns content + `encoding`/`bom`/`line_ending`/`had_decode_errors`.
- `encode_file`: re-encode to the target encoding + BOM + line endings. UTF-16 is encoded by hand (`encoding_rs` has no UTF-16 *encoder* per the Encoding Standard — its `encode()` emits UTF-8).
- `readeditorfile` now reads **bytes** and returns the metadata; `writeeditorfile` accepts `encoding`/`bom`/`line_ending` (absent ⇒ UTF-8, so old callers are unaffected).

**Frontend:** the editor model captures per-tab encoding on read and passes it back on save (`saveFile`/`saveFileAs`), so a non-UTF-8 file saves back in its original format instead of UTF-8. Types extended (`gotypes.d.ts`).

**Deps:** `encoding_rs` (Firefox's codec, also used by ripgrep) + `chardetng` (Firefox's detector). Pure Rust, no external binary — the Rust equivalent of VS Code's `iconv-lite` + `jschardet`.

## Tests

8 unit tests in `text_encoding.rs` (all pass): Windows-1252 `.ini` decode + round-trip, UTF-8/UTF-16 BOM handling, CRLF on write, unknown-label → UTF-8 fallback. Backend builds clean; frontend `tsc` clean (the one pre-existing `rpcCall` baseline error is unrelated).

## Test plan

- [ ] Open a Windows-1252 `.ini` with non-ASCII (`é`, smart quotes) → opens correctly (was erroring).
- [ ] Edit + save → file stays Windows-1252 (not converted to UTF-8); BOM + CRLF preserved.
- [ ] Open a UTF-16 LE file (with BOM) → opens; saves back as UTF-16 LE.
- [ ] Normal UTF-8 files unchanged.

Phase 3 (status-bar encoding indicator + Reopen/Save-with-Encoding UI) is a follow-up per the spec.

### DIFF
--- a/.changesets/1781744589-feat-editor-file-encodings.md
+++ b/.changesets/1781744589-feat-editor-file-encodings.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(editor): open and round-trip non-UTF-8 files (Windows-1252 .ini, UTF-16, Shift_JIS, …) via encoding detection
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,10 +123,12 @@ dependencies = [
  "axum 0.7.9",
  "base64",
  "bollard",
+ "chardetng",
  "chrono",
  "clap",
  "crash-handler",
  "dirs",
+ "encoding_rs",
  "flate2",
  "futures-util",
  "hex",
@@ -833,6 +835,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "chardetng"
+version = "0.1.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14b8f0b65b7b08ae3c8187e8d77174de20cb6777864c6b832d8ad365999cf1ea"
+dependencies = [
+ "cfg-if",
+ "encoding_rs",
+ "memchr",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

a5af:
Addressed both P2s: UTF-32 BOM is now sniffed before UTF-16 and decoded/encoded by hand (encoding_rs has no UTF-32 codec); and `writeeditorfile` now **refuses** a lossy save (unmappable char → error suggesting UTF-8) instead of writing `&#NNNN;` corruption. Added tests for both (11 passing).

a5af:
Addressed P1+P2: `had_decode_errors` now threads to the editor model and **blocks saving** a lossily-decoded file (symmetric with the encode-side refusal — prevents overwriting the original with the U+FFFD buffer); line-ending detection now uses the **first** line break (VS Code behavior) instead of any-CRLF. 12 tests passing.

PR #1536, opened 2026-06-18
fix(agent): auto-resume after AskUserQuestion on persistent panes (dead-air fix)

a5af:
## Summary

Fixes a reliability bug where a Claude (persistent) agent **stalls with no output after the user answers an `AskUserQuestion`** — "dead air" — until manually nudged. Diagnosed live; full write-up in `agentmux-askuserquestion-dead-air-report.md`.

## Root cause

`AskUserQuestion` blocks the model's turn; the answer is delivered as a `control_response` that should resume the same turn. But — already documented for one-shot agents in `SPEC_ASK_USER_QUESTION_2026_06_15.md` §9/§10.1 — **when a turn ends while the question is still parked, the CLI abandons the pending tool_use and the `control_response` is silently dropped** (the model "sees an empty message"). Phase 1 (persistent) had **no recovery**: `answer_question` fired the `control_response` and returned. `proc_status` can't detect this — it tracks the *process* (alive across turns), not the *turn* boundary (which lives in the stdout stream).

## Fix

Give the persistent path the same resilience the one-shot path already has (§9 Phase 2). After sending the `control_response`, `answer_question` snapshots `HealthMonitor::last_output_at()` and arms a short check (`ANSWER_RESUME_FALLBACK_MS` = 4s). If **no stdout line** arrives by then, the turn didn't resume → re-deliver the answer as a **directive follow-up user message** (`build_answer_resume_message`) on the live stdin, so the agent resumes with the decision in context.

**Safe by construction:** gated on output activity → **mutually exclusive** with a real resume, so it can never double-deliver. The happy path (turn still live → `control_response` consumed) is byte-for-byte unchanged. The directive wording ("Resume the task… do not wait for further input") also defends against the secondary failure where a vague "continue" nudge is read as a no-op.

## Scope

- `agentmux-srv/src/backend/blockcontroller/persistent.rs` — `answer_question` fallback + `build_answer_resume_message` + tests
- `agentmux-srv/src/backend/blockcontroller/health.rs` — `last_output_at()` getter
- `docs/specs/SPEC_ASK_USER_QUESTION_2026_06_15.md` — §11 amendment

No protocol or frontend change.

## Test plan

- [x] `cargo check -p agentmux-srv` — clean (only pre-existing warnings)
- [x] `cargo test -p agentmux-srv` — green (new: `answer_resume_message_renders_qa_pairs`, `answer_resume_message_handles_non_object`; existing blockcontroller/health suites unaffected — 84 passed)
- [ ] **Live smoke (needs a real pane):** ask a question, answer it, confirm the turn resumes; force the dead-air case (answer after the turn ends) and confirm the follow-up message re-delivers exactly once. The 4s window is a heuristic that may want tuning under real timing — like the §10 cross-process smokes, it can't be unit-tested.

### DIFF
--- a/.changesets/1781747957-fix-agent-askuserquestion-dead-air-resume.md
+++ b/.changesets/1781747957-fix-agent-askuserquestion-dead-air-resume.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): auto-resume after AskUserQuestion on persistent (Claude) panes. When a turn ends while a question is still parked, the CLI abandons the pending tool_use and the answer's control_response is silently dropped — the agent stalls ("dead air"). `answer_question` now arms a short post-answer check: if no stdout activity follows the control_response, the answer is re-delivered as a directive follow-up user message so the turn resumes. Gated on output activity, so it is mutually exclusive with a real resume and never double-delivers. Mirrors the one-shot controllers' existing message fallback (SPEC_ASK_USER_QUESTION §9/§10.1).
--- a/agentmux-srv/src/backend/blockcontroller/health.rs
+++ b/agentmux-srv/src/backend/blockcontroller/health.rs
@@ -223,6 +223,14 @@ impl HealthMonitor {
         self.inner.lock().unwrap().active_turn
     }
 
+    /// `Instant` of the most recent stdout line (meaningful or not). Used by the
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1539, opened 2026-06-18
fix(agent): harden AskUserQuestion dead-air fallback (codex review on #1536)

a5af:
## Summary

Fast-follow addressing the two **P2** Codex review notes on #1536 (the dead-air fix). Both are about the *resume-detection* signal; the user-visible behavior and happy path are unchanged.

## Codex points addressed

1. **Snapshot before the send** (`persistent.rs`) — the activity snapshot was taken *after* `tx.try_send(control_response)`. Now taken **before**, so a fast resume that emits between the send and the snapshot can't be misread as "no activity."
2. **Count control frames** (`persistent.rs`) — the old signal (`HealthMonitor::last_output_at()`) is only updated by `record_output`, which the stdout reader **skips for `control_request`/`control_response` frames** (it `continue`s first). So a resume whose first activity was a tool-permission round-trip would look like "no output" and spuriously re-deliver the answer.

## How

Introduce a dedicated `stdout_seq: Arc<AtomicU64>` bumped at the **top of the reader loop for every non-empty stdout line** (incl. control frames). `answer_question` snapshots it **before** sending, then after the 4s window re-delivers only if the counter is **unchanged**. Removed the now-unused `HealthMonitor::last_output_at()` getter (so `health.rs` is back to its pre-#1536 state).

Still gated on activity → **mutually exclusive with a real resume, never double-delivers.** Happy path (turn still live → `control_response` consumed) is byte-for-byte unchanged.

## Test plan

- [x] `cargo check -p agentmux-srv` — clean (only pre-existing warnings); no dangling `last_output_at` refs
- [x] `cargo test -p agentmux-srv blockcontroller` — 84 passed (incl. the `answer_resume_message` tests)
- [ ] Live smoke (carried over from #1536): the 4s window still wants real-timing validation — cross-process timing isn't unit-testable.

Spec `SPEC_ASK_USER_QUESTION_2026_06_15.md` §11 updated to match.


a5af:
@codex review

PR #1541, opened 2026-06-18
fix(tabbar): uniform hamburger ↔ first-tab spacing

a5af:
## Analysis

On Windows/Linux the hamburger sits at the left of the tab strip:

```
.tab-bar [flex, no gap]
  HamburgerMenu  (28px button, icon centered, padding:0, no margin)
  .tab-bar-scroll [flex, no gap]
    tab0 · .tab-separator(1px) · tab1 · .tab-separator · tab2 …
```

Each tab label has 10px inner padding; adjacent tabs are divided by a **1px** `.tab-separator`. But the **hamburger→first-tab** boundary was built differently:

1. **No separator** — the loop only inserts a `.tab-separator` before index > 0, so the first tab was **flush** against the hamburger with no hairline (unlike every tab-to-tab boundary).
2. **Tighter** — icon→first-label ≈ button centering inset (~6px) + first-tab padding (10px) ≈ **16px**, vs the inter-tab label gap of 10 + 1 + 10 = **21px**.

So the first tab hugged the hamburger ~5px tighter *and* lacked the divider every other tab has → non-uniform.

## Fix

- Render a leading `.tab-separator` inside `.tab-bar-scroll` when the hamburger is present (non-macOS) → same 1px hairline as every tab boundary.
- Add `margin-right` to the base `.hamburger-btn` so icon→separator matches a tab's inner padding → `icon · 10px · 1px sep · 10px · label` = 21px, uniform. The `--right` (macOS, far-right placement) variant overrides it.

Markup + SCSS only; `tsc`/`stylelint` clean.

## Test plan
- [ ] Windows/Linux: first tab sits the same distance from the hamburger as tabs sit from each other, with the 1px hairline.
- [ ] macOS unaffected (hamburger is far-right; no leading separator).
- [ ] Active/hover first tab no longer glued to the hamburger.

### DIFF
--- a/.changesets/1781752186-fix-tabbar-hamburger-spacing.md
+++ b/.changesets/1781752186-fix-tabbar-hamburger-spacing.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tabbar): uniform spacing/separator between the hamburger and the first tab
--- a/frontend/app/tab/tabbar.tsx
+++ b/frontend/app/tab/tabbar.tsx
@@ -605,6 +605,15 @@ function TabBar(props: TabBarProps): JSX.Element {
                 <HamburgerMenu />
             </Show>
             <div ref={tabBarScrollRef!} class="tab-bar-scroll" data-drag-region="false">
+                {/* When the hamburger sits to the left of the tabs (Windows/
+                    Linux), give the hamburger→first-tab boundary the SAME 1px
+                    separator every tab-to-tab boundary has — otherwise the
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
Fixed the P1: the tear-off anchor now selects the first `.tab-drop-wrapper` (the actual first tab) instead of `firstElementChild`, which the new leading separator had taken over. No more anchor skew on Windows/Linux.

PR #1542, opened 2026-06-18
chore(agent-pane): remove dead StreamStalled streaming-idle watchdog

a5af:
## What

Removes the **dead `StreamStalled` streaming-idle watchdog** from the agent-pane state machine — a 4-link chain whose middle link (the dispatcher) was never built, so the whole feature never executed in production.

## Why it's dead

The intended pipeline was:

```
reducer emits          a dispatcher schedules     timer fires      reducer transitions
schedule-stream-   ->   setTimeout(deadline)   ->  StreamStalled -> Streaming -> Done.errored
watchdog                (NEVER BUILT)              command          ("stream-stalled") + sound
```

The reducer *emitted* `schedule-stream-watchdog` and *held* the terminal `StreamStalled` arm, but **no dispatcher ever consumed the schedule** (`EventSink` has no handler; no `setTimeout`/`deadlineMs` consumer; `StreamStalled` is never dispatched). So the command never fired, the turn was never force-ended, and the `agent.stream.stalled` sound it gated was unreachable — a dead consumer hanging off dead code.

Full analysis: `docs/analysis/ANALYSIS_DEAD_STREAM_STALLED_WATCHDOG_2026_06_18.md` (included).

## What's removed

- `StreamStalled` command + its reducer arm; `schedule-stream-watchdog` / `stream-stalled` events; the `streamWatchdogEvent()` helper + its **6 call sites**; the `STREAMING_IDLE_TIMEOUT_MS` constant.
- The unreachable `agent.stream.stalled` sound (`sounds.ts`), its `sound-service` case + test, and the `notify:sound:agent.stream.stalled` setting across `schema/settings.json`, `settings-template.jsonc`, the `wconfig` Rust field, and the `gotypes.d.ts` binding.

## What's intentionally kept

- The **alive** diagnostic path — `StreamWatchdogTick` (5 s) → `stream-stuck` — is untouched (it's a different, live mechanism; emits diagnostics, no transition).
- The **#1523 reconnect promotion** (resumed live content re-enters `Streaming` from `Idle`/`Disconnected`) and its **regression tests** are preserved (relocated into a new `Stream resume after reconnect` describe block).
- `SubmitTimeoutElapsed` / `InterruptTimeoutElapsed` arms are left as-is (out of scope here — see note below).

## Risk

Pure deletion — **no behavior change**, because the removed code never ran. Removing the `wconfig` settings field is backward-compatible (`SettingsType` has no `deny_unknown_fields`, so existing configs carrying the old key are ignored).

## Verification

- `npx tsc -p tsconfig.json --noEmit` — **zero errors in any modified file** (pre-existing unrelated baseline errors elsewhere are untouched).
- `npx vitest run` on the agent-pane reducer + notification/sound suites — **164 passed**, including the kept reconnect regression tests.
- `cargo check -p agentmux-srv` — passes (warnings only).

## Note (not addressed here)

The agent-pane audit found the dead-watchdog family is **larger** than just `StreamStalled`: the `schedule-submit-timeout` / `schedule-interrupt-timeout` events (and their `SubmitTimeoutElapsed` / `InterruptTimeoutElapsed` commands) follow the same emitted-but-never-dispatched pattern. Left for a separate, explicitly-scoped change.


a5af:
@codex review

PR #1543, opened 2026-06-18
fix(agent): context meter — per-model window, learned from usage

AgentU-asaf:
## Problem
The agent-pane context meter (composer aux bar, `ContextWindowBar`) showed **`/ 200,000` on Opus** and **overflowed without ever reflecting compaction**.

**Root cause:** `frontend/app/view/agent/providers/index.ts` hardcodes the Claude provider `contextWindow: 200_000` — that's **Haiku's** window. Opus/Sonnet are **1,000,000**, and the value is **per-provider when it must be per-model**. So Opus context (e.g. ~300K) ÷ 200K read as 150% (pinned, "300k / 200k"), and since the CLI auto-compacts near Opus's *real* ~987K ceiling, the meter showed it sailing past 200K with no relief.

The meter's **numerator was already correct** (per-turn `message_start.usage` = `input + cache_creation + cache_read`, overwritten not accumulated). Only the denominator + presentation were wrong.

## Why it's a *learned* window, not a static map
The CLI never reports the effective window (**verified**: `system/init` and `result` carry `model` but no window field), and **Sonnet is 200K-or-1M** depending on the `context-1m` beta — so a static `model→window` table can't stay aligned. Resolution layers:
1. **Seed** from the resolved model id the stream reports (`message.model` → `claude-opus-4-8`): Opus/Fable → 1M, Haiku → 200K, Sonnet → **200K conservative**.
2. **High-water upgrade:** a prompt can't exceed the real window, so if observed context crosses the assumed window, promote to the next tier — **this catches Sonnet-1M** (jumps to 1M once context passes 200K).
3. **Band against the auto-compaction threshold** (~`window − 33K`), so a "critical" bar means *about to compact*; fill is clamped (never implies >100%).

The provider's static `contextWindow` stays as a fallback for unrecognised / non-Claude models.

## Changes
- New pure helper `store/agent-pane-state/context-window.ts` (`contextWindowForModel`, `learnContextWindow`, `compactionThreshold`) + **11 unit tests**.
- `TokensIn` carries the resolved `model`; reducer learns `lastContextWindow`; projected through the store to a `contextWindowAtom`; `agent-view` feeds it to the bar (`?? provider().contextWindow` fallback).
- `ContextWindowBar` bands against the compaction threshold and clamps the fill.

## Verification
- **11 new tests + 159 reducer/store tests pass; frontend type-clean.**
- Live-verifiable in the dev build: a fresh **Opus** agent now meters against **1,000,000** and colours toward the compaction threshold instead of pinning at "/200,000".

## Scope / follow-ups (in the spec)
This is steps 1–2 of `SPEC_CONTEXT_VISIBILITY_2026_06_17.md` (included). Deliberately **not** in this PR: seeding the catalog from the Anthropic Models API on CLI install (verified reachable via the agents' OAuth creds — no API key needed), exact-window calibration at the compaction boundary, and a compaction transcript marker. The spec + `docs/reports/REPORT_CONTEXT_VISIBILITY_2026_06_17.md` capture the full roadmap.

### DIFF
--- a/.changesets/1781753670-fix-agent-context-meter-resolves-the-window-per-model-and-le-e7i6.md
+++ b/.changesets/1781753670-fix-agent-context-meter-resolves-the-window-per-model-and-le-e7i6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): context meter resolves the window per-model and learns it from usage
--- a/docs/reports/REPORT_CONTEXT_VISIBILITY_2026_06_17.md
+++ b/docs/reports/REPORT_CONTEXT_VISIBILITY_2026_06_17.md
@@ -1,0 +1,91 @@
+# Report: Context & Token Visibility — findings and next steps
+
+**Date:** 2026-06-17
+**Scope:** Why the agent-pane context meter is wrong, what token/context data AgentMux can actually get from the CLIs it drives, how to source model context windows reliably, and the recommended path forward.
+**Design detail:** `docs/specs/SPEC_CONTEXT_VISIBILITY_2026_06_17.md` (this report is the executive summary + decisions).
+
+---
+
+## 1. The reported bug, diagnosed
+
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

PR #1544, opened 2026-06-18
test(rpc): RPC frontend↔backend contract guard (architecture audit A1)

a5af:
## What

Implements **A1** — the highest value/effort item from the architecture audit (included here as `docs/analysis/ANALYSIS_CODEBASE_ARCHITECTURE_AUDIT_2026_06_18.md`): a CI guardrail for the hand-maintained RPC wire contract.

## Why

The FE↔BE RPC contract lost its code generator (`cmd/generate/main-generatets.go`) when the Go backend was rewritten in Rust, and it was never replaced. Today the contract spans three hand-synced files — `rpc-api.ts` (FE bindings), `rpc_types.rs`, and the `engine.register_handler(...)` registry — with **no codegen, no contract test, no CI check** (only two prose "Keep in sync" comments). It is already drifting, and nothing prevents the next field/command rename from shipping a silent runtime break.

## How

`test/contract/rpc-contract.test.ts` re-derives both sides from source at run time and freezes the three drift sets as **shrink-only baselines**:

| set | count | meaning |
|-----|-------|---------|
| `liveUnregistered` | 13 | commands the FE actually **calls** (`RpcApi.X(…)` or a direct `rpcCall("name")`) with **no backend handler** — latent not-found calls (Wave-inherited telemetry/conn/wsl surface) |
| `declaredUnregistered` | 71 | `rpc-api.ts` methods with no handler (dead binding surface — audit item A12) |
| `registeredUndeclared` | 13 | backend handlers with no FE method (`agent.*` server verbs, `pane.open`, RPC engine test stubs) |

A **new** drift — a fresh `rpc-api.ts` method without a handler, a new live call to an unhandled command, or a removed handler still bound by the FE — changes one of these sets and fails the test. Cleanups that shrink a set update the baseline. This is the guardrail the deleted generator used to provide, at a fraction of the cost.

### Robustness
- Resolves **both** `register_handler("literal", …)` and `register_handler(COMMAND_CONST, …)` forms (the string-literal form is what made an earlier draft over-report 28 false gaps — fixed; `registered` now = 144, matching the audit).
- Distinguishes **live** calls from the ~70 dead-but-declared methods, so the actionable `liveUnregistered` set stays small and meaningful.
- Extractor-sanity assertions (`registered>120`, `declared>180`, a known `setmeta` round-trip) prevent a broken regex from passing the diff assertions vacuously.
- An `unresolved register_handler args` assertion fails loudly if a registration form is ever missed.

## Verification

- `npx vitest run test/contract/rpc-contract.test.ts` → **4 passed**.
- **Negative test:** injecting a `rpcCall("zzz-nonexistent-cmd")` into a FE file makes the live-call assertion fail with `+ "zzz-nonexistent-cmd"`; removing it restores green — confirming the guard catches drift rather than passing vacuously.
- No source changes — purely additive (one test + the audit doc). No tsc impact.

## Scope

This is A1 only. The audit's other items (A2–A15 — e.g. extracting the mcp/bashwrap HTTP DTOs, splitting the backend god-files, collapsing the agent-pane state systems) are left for follow-ups. The two prose "Keep in sync" comments in `rpc-api.ts` / `gotypes.d.ts` could later point at this test.

### DIFF
--- a/.changesets/1781754556-test-rpc-add-frontend-backend-rpc-command-contract-guard-arc-ooae.md
+++ b/.changesets/1781754556-test-rpc-add-frontend-backend-rpc-command-contract-guard-arc-ooae.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+test(rpc): add frontend<->backend RPC command contract guard (architecture audit A1)
--- a/docs/analysis/ANALYSIS_CODEBASE_ARCHITECTURE_AUDIT_2026_06_18.md
+++ b/docs/analysis/ANALYSIS_CODEBASE_ARCHITECTURE_AUDIT_2026_06_18.md
@@ -1,0 +1,319 @@
+# AgentMux — Codebase Architecture & Refactor Audit
+
+**Date:** 2026-06-18
+**Author:** smike
+**Scope:** Whole repo — frontend app structure, the agent-pane subsystem, the `agentmux-srv`
+Rust backend, and cross-cutting contracts / provider pipeline / workspace boundaries / docs.
+**Method:** Four parallel deep-dive audits, each citing `file:line` evidence, synthesized here.
+**Status:** Analysis (no code change). Companion to the just-merged
+[`ANALYSIS_DEAD_STREAM_STALLED_WATCHDOG_2026_06_18.md`](ANALYSIS_DEAD_STREAM_STALLED_WATCHDOG_2026_06_18.md)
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
Good catch — fixed in `7f3e1f3b`. The method→command extractor now scans within each binding's own body (`[signature, next signature)`) and treats both `rpcCall` and `rpcStream` as bindings, so the 9 stream commands are no longer dropped/mismapped (`declared` 202 → 211). Added a hard assertion that no binding is left unresolved plus a `fileliststream` sanity check so this can't regress. The `declaredUnregistered` baseline was updated to include the stream commands (also unregistered Wave surface); the other two sets are unchanged.

PR #1545, opened 2026-06-18
refactor(agent/providers): shared ToolCorrelator for tool-call/result mapping (audit A7)

a5af:
## What

Implements **A7** from the architecture audit: removes the four-way duplicated tool-call ↔ tool-result correlation in the stream translators.

## Why

`codex`, `gemini`, `kimi` and `acp` each independently maintained a `toolNameById` map, registered `id → name` on a tool call, looked it up on the result, and hand-built the provider-agnostic `tool_call`/`tool_result` event literals — the same ~40-line block copy-pasted four times, differing only in **which wire fields** the name/id/params come from. That made "the tool name didn't resolve" a four-place bug.

## How

New `tool-correlation.ts`:
- **`ToolCorrelator`** — owns the `id → name` map, the `call()` / `result()` event builders, and `reset()`. `result()` resolves the name (with a fallback), so correlation lives in exactly one place.
- **`wrapOutput`** — the repeated `typeof output === "string" ? { output } : output` normalisation (codex/gemini/acp).

Each translator keeps its **own field extraction** — the genuinely provider-specific part (`call_id` vs `tool_id` vs `toolCallId`; `arguments` vs `parameters` vs `input`; etc.). Only the identical mechanics moved.

## Behaviour preserved exactly

Verified against the existing suites + a new direct unit test:
- codex's `_raw` JSON-parse fallback and `call-${Date.now()}` id prefix
- kimi's object-arg passthrough and `{ content }` result shape (not `{ output }`)
- acp's `map ?? params.toolName ?? "unknown"` fallback chain — preserved via `fallbackName`, since `a ?? b ?? c === a ?? (b ?? c)`
- gemini/codex/acp success/failed status logic

**Claude is intentionally left untouched** — its streaming `content_block` accumulation and piecewise `tool_call` emission make it genuinely distinct, not duplication (audit §4).

## Result

- Translators shrink ~40 LOC (-75/+35); the mechanics are centralised + directly tested.
- **All 6 provider suites pass** (codex/gemini/kimi/claude/index + new `tool-correlation.test.ts`). No tsc impact.

### DIFF
--- a/.changesets/1781758234-refactor-agent-providers-extract-shared-toolcorrelator-for-t-vbf9.md
+++ b/.changesets/1781758234-refactor-agent-providers-extract-shared-toolcorrelator-for-t-vbf9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(agent/providers): extract shared ToolCorrelator for tool-call/result mapping (architecture audit A7)
--- a/frontend/app/view/agent/providers/acp-translator.ts
+++ b/frontend/app/view/agent/providers/acp-translator.ts
@@ -1,8 +1,9 @@
 // Copyright 2026, AgentMux Corp.
 // SPDX-License-Identifier: Apache-2.0
 
-import type { StreamEvent, ToolCallEvent, ToolResultEvent } from "../types";
+import type { StreamEvent } from "../types";
 import type { OutputTranslator } from "./translator";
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

a5af:
@codex review

PR #1551, opened 2026-06-18
fix(srv): harden error paths — poison-tolerant mutex, log silent drops, guard take().unwrap() (A15)

AgentA-asaf:
## Summary

Addresses A15 from the architecture refactor tracking board (`docs/analysis/TRACKING_ARCHITECTURE_REFACTOR_A1_A15_2026_06_18.md`).

- **`rpc/engine.rs`** — add `lock_inner()` helper that recovers from mutex poison via `unwrap_or_else(|e| e.into_inner())`, replacing all 15 `lock().unwrap()` call sites. Also logs when the output channel is closed (`send_output`) and when a response channel is full or closed (`handle_response`).
- **`subprocess.rs`, `persistent.rs`, `acp.rs`** — replace `child.stdin/stdout.take().unwrap()` with `.ok_or_else(...)` so a missing pipe returns `Err` instead of panicking the Tokio task.
- **`acp.rs`** — log `warn`/`error` when `try_send` silently drops ACP handshake messages (`initialize`, `initialized`, `session/create`) or shutdown messages (`shutdown`, `exit`). ACP agents that fail to start will now appear in logs.

## Test plan

- [ ] `cargo check -p agentmux-srv` passes (no new errors) ✅ (36s, 66 pre-existing warnings)
- [ ] Start an ACP agent, observe normal startup in logs
- [ ] Kill an ACP agent mid-run, confirm graceful stop path logs debug (not error)
- [ ] Confirm no panic under `lock_inner()` usage in a normal run

### DIFF
--- a/.changesets/1781776672392-fix-srv-error-hardening-94e3c9878898.md
+++ b/.changesets/1781776672392-fix-srv-error-hardening-94e3c9878898.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+Harden srv error paths: poison-tolerant mutex in rpc/engine, log silent channel drops, guard unguarded take().unwrap() in subprocess/persistent/acp controllers with proper Err returns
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1559, opened 2026-06-18
fix(floater): dwell gate for redock ghost + retry on back-to-back tear-off

a5af:
## Summary

- **Bug 1 — redock ghost delay**: `installFloatingRedockHoverListener` in `app-init.ts` now gates the ghost placeholder behind 180ms of continuous hover over the same target window, matching `REDOCK_DWELL_MS` already enforced by the floater for the actual drop. On Windows, `Win32BeginMoveTask` fires `floating-redock:hover-state` every 50ms with no dwell — the ghost was appearing immediately on cursor entry and surprising users who just wanted to reposition the floater.

- **Bug 3 — second tear-off silently drops**: All three `CrossWindowDragMonitor` variants (win32, darwin, linux) now retry `open_floating_pane_window` once after 350ms when the error message contains `"currently closing"`. Root cause: the first tear-off auto-closes its source tab (`service.rs:2182`), briefly setting `BrowserPaneLifecycle::Closing`, which makes `any_browser_pane_closing()` return true and rejects any concurrent `open_floating_pane_window` call. The 350ms window covers the typical 100–200ms tab-close lifecycle.

## Out of scope (tracked separately)

Bug 2 — floater Z-order always above main window — is a Win32 owned-window invariant and requires removing the owner relationship + manual minimize/restore/destroy cascade. Filed as a separate issue.

## Test plan

- [ ] Drag floating pane quickly across main window border — ghost should NOT appear (< 180ms pass-through)
- [ ] Hold cursor over main window for > 180ms with floater being dragged — ghost should appear and snap correctly on drop
- [ ] Tear off a pane, then immediately tear off a second pane — both should produce floating windows (no silent drop)
- [ ] Non-"currently closing" IPC failures still log and bail (no retry loop)

### DIFF
--- a/.changesets/1781781025-fix-floater-dwell-gate-for-redock-ghost-retry-on-back-to-bac-n07u.md
+++ b/.changesets/1781781025-fix-floater-dwell-gate-for-redock-ghost-retry-on-back-to-bac-n07u.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floater): dwell gate for redock ghost + retry on back-to-back tear-off
--- a/frontend/app-init.ts
+++ b/frontend/app-init.ts
@@ -161,6 +161,16 @@ function installFloatingRedockHoverListener(): void {
         }
     };
 
+    // Dwell gate — mirrors the 180ms gate in floating-pane-workspace.tsx.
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #1561, opened 2026-06-18
fix(agent): Edit tool expand view shows diff instead of 'No diff available'

a5af:
## Summary

- **Root cause:** `DiffViewer` read `props.result?.diff` to get the diff text, but Claude's translator returns a plain-text acknowledgement string as the tool result — not a structured `EditResult` object — so `result.diff` is always `undefined`. The component hit its early-return empty-state on every Edit tool call.
- **Fix:** When `result.diff` is absent, fall back to computing a unified diff client-side from `params.old_string` / `params.new_string`, which are always present on every Edit call. LCS line-diff (O(mn)) for spans ≤ 300 lines; plain del+add block for larger spans. No new dependencies.
- `result.diff` is still preferred when present (future-proof for providers that do populate it).

## Test plan

- [ ] Trigger any Edit tool call in an agent pane, click to expand — should show a syntax-highlighted diff (green adds / red deletes) instead of "No diff available"
- [ ] Test with a small edit (few lines) — LCS path, shows proper context lines
- [ ] Test with a large edit (> 300 lines) — fallback del+add block path
- [ ] Verify file path header renders correctly in all cases
- [ ] Verify Shiki highlight loads asynchronously (PlainDiff shown first, then highlights in)

### DIFF
--- a/.changesets/1781780526-fix-agent-edit-tool-expand-panel-now-shows-a-diff-instead-of-ga49.md
+++ b/.changesets/1781780526-fix-agent-edit-tool-expand-panel-now-shows-a-diff-instead-of-ga49.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): Edit tool expand panel now shows a diff instead of 'No diff available'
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1571, opened 2026-06-18
fix(toolchain): docker/git/node show as not installed on Windows

a5af:
## Summary

- `resolve_cli_on_path()` in `cli_handlers.rs` was appending `.cmd` to every command name on Windows before calling `where`. This works for `npm` (which genuinely ships as `npm.cmd`) but silently breaks detection for `docker`, `git`, and `node` — all of which install as `.exe` files.
- `where docker.cmd` returns nothing even with Docker Desktop fully installed → toolchain manager reports Docker as not installed.
- Fix: pass the bare command name to `where`. Windows resolves the correct extension via PATHEXT and returns the full path including extension (`.exe` or `.cmd`). `make_cli_cmd()` already branches on the extension to handle native `.exe` spawn vs `.cmd` wrapper unwrapping, so invocation is unaffected.

**Affected tools on Windows (all now fixed):** `docker`, `git`, `node`  
**Unaffected:** `npm` (still found as `npm.cmd`, `make_cli_cmd` continues to unwrap it correctly)

## Test plan

- [ ] Open Toolchain Manager on Windows with Docker Desktop installed — Docker should show version + path (was: "not installed")
- [ ] Same for Git for Windows and Node.js
- [ ] npm detection still works (finds `npm.cmd`, version reported correctly)
- [ ] On macOS/Linux — no change to `which` path, unaffected

### DIFF
--- a/.changesets/1781807290-fix-toolchain-detect-docker-git-node-on-windows-where-lookup-pb2t.md
+++ b/.changesets/1781807290-fix-toolchain-detect-docker-git-node-on-windows-where-lookup-pb2t.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(toolchain): detect docker/git/node on Windows — where lookup now uses bare command name so .exe tools are found via PATHEXT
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1568, opened 2026-06-18
feat(macos): bundle id follows channel + unix open_new_window forward

a5af:
## What & why

On macOS every build shipped a constant `CFBundleIdentifier` (`ai.agentmux.cef`), so LaunchServices couldn't tell versions/channels apart. Double-clicking one build while another ran reactivated the *other* instance and hung → **"AgentMux is not responding"**, and a relaunch never opened a new window (the `open_new_window` forward was Windows-only). Full analysis: `docs/specs/SPEC_MACOS_LAUNCH_COHERENCE_2026_06_18.md`.

Two "attach layers" must agree — LaunchServices (`CFBundleIdentifier`) and the launcher (`channels/<channel>/versions/<version>/` data dir). They didn't. This aligns them.

## Changes

1. **`scripts/package-macos.sh` — bundle id follows the channel.** `BUNDLE_ID="ai.agentmux.cef.${CHANNEL}"` from `${AGENTMUX_BUILD_CHANNEL_DEFAULT:-stable}` (the same channel compiled into the binaries), sanitized; helpers/app inherit it. ✅ **verified** — a `.stable` build launches next to a running 0.45.0 with no "not responding".
2. **`agentmux-launcher` — wire the unix `open_new_window` forward.** The second-instance path in `run_unix` / `bind_socket_with_recovery` now calls `forward_open_new_window` (mirroring the Windows pipe path) before exiting. ✅ **verified** — `open -n` of a second instance opens a new window; the second launcher forwards + exits.
3. **`agentmux-cef` — `kAEReopenApplication` handler.** ⚠️ **Included but currently inert.** Chromium owns the reopen AppleEvent and re-registers after our raw `NSAppleEventManager` handler, so a plain double-click never reaches it. The working fix is to swizzle the NSApp delegate's `applicationShouldHandleReopen:` (the pattern the a11y governor already uses) — tracked as a follow-up (spec §7). **Reviewers: keep this as documented scaffolding, or strip it and land the swizzle separately?**

## Verification (on-device, Apple Silicon)
- Notarized `.stable` DMG launches with no "not responding" alongside a running `ai.agentmux.cef` instance; LaunchServices registers it as a distinct app.
- `open -n` second instance → new window in the running instance; 2nd launcher exits.
- Reopen handler logs `installed` but never `fired` → confirms #3 is inert.

## Scope / safety
[...]

PR #1572, opened 2026-06-18
feat(macos): plain double-click of a running app opens a new window

a5af:
## What & why

Follow-up to #1568 (macOS launch coherence). That PR's `kAEReopenApplication` handler was **inert** — CEF re-registers its own AppleEvent handler after ours, so a plain Finder/Dock double-click of the running app never reached it (it just focused the existing window). This wires the **NSApplication delegate** path instead.

## Change

`agentmux-cef/src/macos_menu.rs` — replace the raw AE handler with a dedicated **`applicationShouldHandleReopen:hasVisibleWindows:` delegate**. CEF sets no delegate of its own, so we install one (and, defensively, hook the method onto an existing delegate's class if one ever appears). On reopen it calls `open_new_window` and returns YES.

## Verification (on-device, Apple Silicon)

Built a notarized `.stable` DMG, launched it, triggered a reopen:
- Install log: `reopen-hook: installed dedicated reopen delegate (NSApp had none)`
- On reopen: `reopen-hook:fired — applicationShouldHandleReopen opened a new window`
- Renderer helpers **2 → 3** — a real second window appeared.

This is the difference from the raw-AE attempt, which logged `installed` but never `fired`.

## Scope / safety

- macOS-only (`#![cfg(target_os = "macos")]` module). Windows/Linux unaffected.
- Mirrors the existing libobjc patterns in this crate (menu `make_target`, a11y governor swizzle).
- Updates `docs/specs/SPEC_MACOS_LAUNCH_COHERENCE_2026_06_18.md` §7 to mark #2 fixed.


a5af:
@codex review

a5af:
@codex review — addressed both P2s: `hook_class` now uses `class_addMethod`-first (never swizzles an inherited superclass IMP), and the stale call-site comment is corrected to describe the delegate approach.

a5af:
@codex review — fixed: `applicationShouldHandleReopen:` now returns **NO** (we've taken responsibility for the reopen by opening a window; AppKit must not also run its default reopen), and the inverted comment is corrected. Earlier P2s (inherited-IMP swizzle + stale call-site comment) were addressed in the prior commit.

PR #1574, opened 2026-06-18
fix(win32): remove owned-window z-order invariant; add floater cascade + diagnostics

### LINKED ISSUE #1560: fix(floater): owned window Z-order — floater always above main window
## Problem

The floating pane always renders above the main window, even when the user clicks the main window and expects it to come to the foreground. This makes it impossible to hide the floater "behind" the main window.

## Root cause

`create_owned_popup` in `agentmux-cef/src/floating_pane.rs` creates the floater as a **Win32 owned window** (owner = main HWND). This is a Win32 invariant: owned windows are always Z-ordered above their owner and cannot be pushed below with `SetWindowPos`. No amount of `SetWindowPos`/`BringWindowToTop`/`SetForegroundWindow` can override this — the invariant is enforced by the kernel.

The floater also suppresses its NC activation border via `WM_NCACTIVATE => return 1` (purely cosmetic) and uses `SW_SHOWNOACTIVATE` on show. Neither of these is the cause — they're intentional.

## What a fix requires

Remove the owner relationship by passing `HWND_DESKTOP` (NULL) as the owner in `CreateWindowExW`. Then manually implement the cascade behaviours that owned windows get for free:

1. **Minimize together** — subscribe to `WM_SIZE` on the main window; when main minimizes, minimize/hide floaters
2. **Restore together** — on main restore, show floaters again
3. **Destroy on close** — on `WM_CLOSE`/`WM_DESTROY` of main, destroy all orphaned floaters
4. **Z-order follow** — optionally call `SetWindowPos(floater, mainHWND, ...)` when main is brought forward so floater stays relatively near but NOT above

Without removing the owner, there is no workaround.

## Scope

- `agentmux-cef/src/floating_pane.rs` — `create_owned_popup`: remove owner HWND; add Z-order tracking callback
- `agentmux-cef/src/floating_pane.rs` — `floating_pane_wndproc`: no change needed for Z-order; the NC suppress and SHOWNOACTIVATE are still correct
- `agentmux-cef/src/ui_tasks.rs` / main wndproc — subscribe to WM_SIZE, WM_DESTROY on main to cascade to owned floaters (now unowned, so needs explicit tracking)

## Current behaviour workaround

None available. This is a Win32 architectural constraint.

## References

- `agentmux-cef/src/floating_pane.rs:create_owned_popup` — where owner HWND is set
- Win32 docs: [Owned Windows](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#owned-windows) — "An owned window is always above its owner in the Z order"

a5af:
## Summary

- **Removes Win32 owned-window relationship** from floating panes. The owned-window invariant forces owned windows permanently above their owner — it cannot be overridden with `SetWindowPos`. Floaters are now created as unowned `WS_POPUP + WS_EX_TOOLWINDOW` windows.
- **Adds explicit cascade hook** (`install_main_window_floater_cascade_hook` in `client/wndproc.rs`) that subclasses the main window's WndProc to replicate owned-window behaviour manually: floaters are placed below the main window on `WM_ACTIVATE`, hidden/shown on minimize/restore (`WM_SIZE`), and closed on `WM_DESTROY`.
- **Global floater HWND registry** (`ACTIVE_FLOATER_HWNDS` in `floating_pane.rs`) tracks every live floater; populated at creation, pruned on `WM_DESTROY` in `floating_pane_wndproc`.
- **`get_pane_debug_state` IPC command** — returns `any_browser_pane_closing`, `pool_queue_size`, `pending_window_creations`, and active floater HWNDs for structured diagnostics.
- **Pre-flight diagnostic logging** in `CrossWindowDragMonitor.win32.tsx` calls `get_pane_debug_state` before every `open_floating_pane_window` so intermittent tear-off failures can be correlated with pane-closing state in the frontend log.

Closes #1560.

## Test plan

- [ ] Tear off a pane — main window should come to foreground when clicked; floater should sit below it
- [ ] Minimize main window — floaters should hide; restore should bring them back
- [ ] Close main window — floaters should close
- [ ] Back-to-back tear-offs still work (retry path from PR #1559 unaffected)
- [ ] Check frontend logs on a failed tear-off: pre-flight snapshot should show `any_browser_pane_closing: true`

### DIFF
--- a/.changesets/1781823967-fix-win32-bind-floater-cascade-to-source-window-fix-z-order--38iw.md
+++ b/.changesets/1781823967-fix-win32-bind-floater-cascade-to-source-window-fix-z-order--38iw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(win32): bind floater cascade to source window; fix z-order and multi-window regression
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -134,13 +134,15 @@ mod handlers;
 pub(crate) mod helpers;
 #[cfg(target_os = "windows")]
 mod wndproc;
+#[cfg(target_os = "windows")]
+pub(crate) use wndproc::install_main_window_floater_cascade_hook;
 
 pub use handlers::AgentMuxClient;
 
 use helpers::{js_string_literal, html_escape, backend_close_window};
 #[cfg(target_os = "windows")]
 use wndproc::{
-    install_main_window_floater_cascade_hook, install_top_level_focus_restore_hook,
+    install_top_level_focus_restore_hook,
     set_window_icon, skip_taskbar,
 };
 
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -139,7 +139,10 @@ pub use handlers::AgentMuxClient;
 
 use helpers::{js_string_literal, html_escape, backend_close_window};
 #[cfg(target_os = "windows")]
-use wndproc::{install_top_level_focus_restore_hook, set_window_icon, skip_taskbar};
+use wndproc::{
+    install_main_window_floater_cascade_hook, install_top_level_focus_restore_hook,
+    set_window_icon, skip_taskbar,
+};
 
 impl AgentMuxHandler {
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #1573, opened 2026-06-18
feat(agent-pane): host vs container runtime differentiation

AgentA-asaf:
## Summary

- **RuntimeBadge** — new green/amber pill component (`fa-box` container, `fa-server` host)
- **Running pane** — always-visible `PaneRow` pin shows sandbox vs full-access mode (green accent = container, amber = host)
- **My Agents list** — runtime badge on each agent row (requires `agent_type` added to `RecentSessionRow`)
- **Template picker** — runtime badge on each AgentCard tile
- **Default flipped** — new agents now default to container (sandbox) instead of host
- **Host warning** — amber callout in the launch modal when "On this computer" is selected, explaining full system access
- **Seed templates** — 6 of 7 templates updated to `agent_type: "container"` (openclaw stays host — needs host access to orchestrate agents); seed version bumped 6→7

## Spec

`docs/specs/SPEC_HOST_VS_CONTAINER_AGENTS_2026_06_18.md`

## Commits

1. `feat(agent-pane): host vs container runtime badges across all agent surfaces` — purely additive, zero behavior change
2. `feat(agent-pane): default new agents to container; warn when host is selected` — behavior change: container is now default

## Test plan

- [ ] Create a new agent from a template — launch modal should pre-select "In a safe sandbox"
- [ ] Select "On this computer" — amber warning callout should appear below the runtime fieldset
- [ ] Create an agent as container — My Agents list shows green Container badge
- [ ] Create an agent as host — My Agents list shows amber Host badge; running pane shows amber PaneRow
- [ ] Open a container agent — running pane shows green PaneRow "Container — isolated Docker sandbox"
- [ ] Template picker (+ New from template section) — each card shows Container or Host badge
- [ ] OpenClaw template card should show amber Host badge (containerSupported: false)
- [ ] Existing agents created before this PR (agent_type = "host") continue to launch as host

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1576, opened 2026-06-18
fix(agent-pane): choose host/container at template instantiation; drop bogus template runtime badge

AgentA-asaf:
**Follow-up to #1573** — a correction to the host/container differentiation feature, stacked on its branch.

## The problem (found while dogfooding 0.46.4 on a Windows box with no Docker)

Two symptoms reported by the user, one root cause:

1. **"agents not responding"** — their seeded agents were all `agent_type = "container"`, and container agents spawn via the Docker socket API. With no Docker daemon (`Docker daemon not reachable; container agent panes will fail to start` in the srv log), the container never starts, so the agent never comes alive to answer.
2. **"can't create a host agent — it's hard-coded to container"** — of the agent-create paths, only the full launch modal exposes a host/container radio. The **+ New from template** path inherits the template's `agent_type` (now `container`) with **no host escape hatch**.

The conceptual error #1573 introduced: it treats **runtime (host vs container) as a property of the template**. It isn't. A "Claude" / "Codex" template is a CLI you can run *either* way. Stamping `agent_type="container"` on templates and badging the template cards is what produced both the mislabeling and the Docker-less container trap.

## The fix — move the choice to instantiation time

- **`AgentCard`**: remove the `RuntimeBadge` from template cards (templates are runtime-agnostic). The badge stays on **My Agents** (`MyAgentsList`), where a launched session has a concrete runtime.
- **`AgentCreateFromTemplateModal`**: add a **Docker-aware host/container dropdown**.
  - Defaults to **host** so the new agent always actually starts.
  - The **container** option is only enabled when the CLI supports containers *and* a Docker runtime is resolvable (same `ResolveCli` probe the launch flow gates on).
  - When Docker **is** present and the template suggests container, it defaults to container — so #1573's secure-by-default intent is preserved where it can actually work. Never auto-selects a runtime that can't run.
- **Threading**: the chosen runtime flows `onSubmit → onCreatedAndLaunch → launch override`, and is **persisted on the cloned definition** (`agentdefcreatefromtemplate` gains an `agent_type` field) so later reattach / auto-continue uses the runtime the user actually picked, not the template's.

## Tests / smoke
- `AgentCreateFromTemplateModal.test.tsx` +2: container option disabled when no Docker; defaults to container when Docker available + container template. Existing modal + `AgentPicker` suites green (6/6, 9/9).
- `tsc` shows no new errors in touched files (repo builds via Vite).

## Note / open question for @parko
The seeded templates still carry `agent_type` in the DB — now used only as the modal's **Docker-gated default suggestion**, which is the right role for it. If you'd rather templates store no runtime at all, that's a separate seed change (gen-seed.js + schemaVersion bump) we can do on your branch.

### DIFF
--- a/.changesets/1781822953-fix-agent-pane-pick-host-container-when-creating-from-a-temp-da2a.md
+++ b/.changesets/1781822953-fix-agent-pane-pick-host-container-when-creating-from-a-temp-da2a.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): pick host/container when creating from a template; drop bogus runtime badge on templates
--- a/agentmux-srv/src/backend/rpc_types.rs
+++ b/agentmux-srv/src/backend/rpc_types.rs
@@ -1425,6 +1425,13 @@ pub struct CommandAgentDefCreateFromTemplateData {
     /// Same semantics as `identity_id` above.
     #[serde(default)]
     pub memory_id: String,
+    /// Runtime to persist on the cloned definition: "host" or
+    /// "container". Empty/absent → keep the template's `agent_type`.
+    /// Runtime is chosen at instantiation time, not a property of the
+    /// template, so the clone records the user's pick rather than
+    /// inheriting the (now container-defaulted) template value.
+    #[serde(default)]
+    pub agent_type: String,
 }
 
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

PR #1578, opened 2026-06-18
fix(layout): pane resize on Linux — drive resize from window pointermove

AgentU-asaf:
## Problem
Dragging a pane divider to resize does nothing on **Linux**. Works on Windows/macOS.

## Root cause
On Linux/CEF (Wayland), `PointerEvent.pointerId` is **not stable across a press → drag**: the `pointerdown` arrives as one id (e.g. `1`) but the subsequent `pointermove` events arrive as a *different* id (e.g. `2`), and `setPointerCapture(1)` does not route those `id=2` moves once the cursor leaves the thin handle.

The shared `ResizeHandle` guarded each move with `trackingPointer() === event.pointerId` (with `trackingPointer` set from `onGotPointerCapture`). On Linux that comparison is **always false**, so `onResizeMove` never fires. win32/darwin work because their `pointerId` stays consistent.

Verified by instrumenting the handler — every move logged:
```
[resize-dbg] pointerdown id=1 → setPointerCapture OK id=1
[resize-dbg] pointermove tracking=1 id=2 match=false buttons=1   (×many)
```

## Fix
Stop depending on `pointerId` and pointer-capture routing on Linux. The handle now:
- arms on `pointerdown` (primary button) and `preventDefault()`s to stop the press also starting a tear-off/native window drag,
- attaches `pointermove`/`pointerup`/`pointercancel` listeners to **`window`** — which receive every move regardless of `pointerId` or whether the cursor is still over the thin handle — driving `onResizeMove` while the primary button is held,
- tears down on release (or when `buttons` goes to 0).

`setPointerCapture` is kept best-effort (helps over native panes on platforms where it works). **`TileLayout.win32.tsx` / `TileLayout.darwin.tsx` are unchanged.**

## Verification
Built locally (`task dev`) on Linux/CEF. After the fix, dragging a divider resizes smoothly; the log confirms `onResizeMove` firing throughout the drag (`id=2 buttons=1`, 99× in one drag).

## Scope
- One focused hunk in `frontend/layout/lib/TileLayout.linux.tsx` (the `ResizeHandle` component) + a `patch` changeset.
- No backend or shared-type changes.

## Test plan
- [ ] Linux: drag horizontal and vertical pane dividers — panes resize.
- [ ] Linux: pane tear-off / rearrange still works (drag behavior unchanged).
- [ ] Windows / macOS: resize unaffected (files untouched).

### DIFF
--- a/.changesets/1781825760-fix-layout-pane-resize-on-linux-drive-resize-from-window-poi-j1lb.md
+++ b/.changesets/1781825760-fix-layout-pane-resize-on-linux-drive-resize-from-window-poi-j1lb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): pane resize on Linux — drive resize from window pointermove instead of the unstable CEF pointerId
--- a/frontend/layout/lib/TileLayout.linux.tsx
+++ b/frontend/layout/lib/TileLayout.linux.tsx
@@ -685,37 +685,67 @@ interface ResizeHandleComponentProps {
 
 const ResizeHandle = (props: ResizeHandleComponentProps) => {
     let resizeHandleRef: HTMLDivElement | undefined;
-    const [trackingPointer, setTrackingPointer] = createSignal<number | undefined>(undefined);
 
-    const handlePointerMove = throttle(10, (event: PointerEvent) => {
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1579, opened 2026-06-18
fix(agent-failure): surface auth 401 errors delivered as in-band assistant messages

AgentA-asaf:
## Summary

- Claude wraps API 401 errors as synthetic `assistant` messages (`isApiErrorMessage:true`, `error:"authentication_failed"`) and exits 0 with `is_error:false` on the `result` frame
- The agentfailure gate at `subprocess.rs` only checks `exit_code != 0 || frame_is_error` — both false for these → gate never trips → pane shows "Worked" with no error UI, no recovery row
- This was the reason Parko silently swallowed every "u there" message during the auth outage

## Changes

**`subprocess.rs`**
- Track `last_inband_error` Arc for `assistant` frames with `isApiErrorMessage:true` or `error` field set
- Trip the gate when `inband_is_api_error` (alongside the existing `exit_code != 0 || frame_is_error`)
- Merge the in-band error text (the `error` field + `message.content[0].text`) into the tail passed to `classify()` so keyword matchers see it even though it arrived on stdout, not stderr

**`failure.rs`**
- Add `"authentication_failed"` keyword to Auth class (the exact string in the `error` field)
- Add `"invalid authentication"` keyword (covers "Invalid authentication credentials" from content text)
- Add `"error: {code}"` variant to `mentions_http_status` so "API Error: 401" (lowercased → "api error: 401") matches

## Tests

3 new regression tests in `failure.rs` covering:
1. `authentication_failed` string (the `error` field value)
2. `invalid authentication credentials` (content text)
3. `api error: 401` (the colon variant that appears in real transcripts)

All verified against a live Parko agent session JSONL where the 401 pattern was confirmed.

## Test plan

- [ ] Build `agentmux-srv` and relaunch with an expired credential store (`~/.agentmux/shared/providers/claude/.credentials.json`)
- [ ] Send a message to an agent in that expired state — should now show the "Not authenticated" recovery row instead of silently completing
- [ ] `cargo test -p agentmux-srv failure` — all 26 pass

### DIFF
--- a/.changesets/1781826736-fix-agent-failure-surface-auth-401-errors-delivered-as-in-ba-y864.md
+++ b/.changesets/1781826736-fix-agent-failure-surface-auth-401-errors-delivered-as-in-ba-y864.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-failure): surface auth 401 errors delivered as in-band assistant messages
--- a/agentmux-srv/src/agents/failure.rs
+++ b/agentmux-srv/src/agents/failure.rs
@@ -177,6 +177,8 @@ pub fn classify(
         );
     }
     if hay.contains("authentication_error")
+        || hay.contains("authentication_failed")
+        || hay.contains("invalid authentication")
         || hay.contains("invalid api key")
         || hay.contains("invalid x-api-key")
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1582, opened 2026-06-19
fix(dnd): three tearoff smoke-test fixes — pool warn noise, ghost race, wrong-tab DeleteBlock

a5af:
## Summary

Three bugs found during pane/tab tear-off smoke testing on macOS:

- **Pool warn noise** (`drag.rs`): The warm-path pool is Windows-only by design (`#[cfg(not(target_os = "windows"))]`). On macOS/Linux, hitting the cold path logged a spurious `WARN dnd:tearoff:pool pool exhausted`. Demoted to `DEBUG` on non-Windows; `WARN` is preserved on Windows where pool exhaustion is a real problem.

- **Ghost race on redock** (`floating-pane-workspace.tsx`): The hover IPC (`update_floating_redock_hover`) was fired at the same threshold (`REDOCK_DWELL_MS` = 180ms) as the redock arm condition in `onMouseUp`. With ~20–50ms IPC round-trip latency, the ghost hadn't appeared when the user released the mouse — block docked with no visual confirmation. Fixed by firing the IPC at `REDOCK_DWELL_MS - HOVER_THROTTLE_MS` (130ms), giving the round-trip a 50ms runway to complete before `onMouseUp` can arm.

- **Wrong-tab DeleteBlock on floating pane close** (`service.rs`): `DeleteBlock` was extracting `tab_id` from `uicontext.active_tab_id` — which for a floating pane window is the window's own tab, not the block's source tab. The `delete_block` saga's pre-condition (`block.tab_id != tab_id`) always tripped, leaving the block alive and logging `[BUG-TRACE] onNodeDelete ERROR`. Fixed by looking up `s.blocks.get(&block_id).tab_id` from server state instead.

## Test plan

- [ ] Tear off a pane on macOS — no `WARN dnd:tearoff:pool` in log (only `DEBUG`)
- [ ] Slow-cursor approach to a tab bar for ~200ms then release — ghost should be visible before the block docks
- [ ] Close a pane from a floating pane window — block should be deleted cleanly, no `[BUG-TRACE]` in log

Full analysis: `docs/analysis/ANALYSIS_PANE_TAB_TEAR_SMOKE_2026_06_19.md`

### DIFF
--- a/.changesets/1781844521-fix-dnd-suppress-misleading-pool-exhaustion-warn-on-non-wind-189f.md
+++ b/.changesets/1781844521-fix-dnd-suppress-misleading-pool-exhaustion-warn-on-non-wind-189f.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dnd): suppress misleading pool-exhaustion WARN on non-Windows tearoff
--- a/.changesets/1781844523-fix-dnd-fire-redock-hover-ipc-before-arm-threshold-to-preven-jp22.md
+++ b/.changesets/1781844523-fix-dnd-fire-redock-hover-ipc-before-arm-threshold-to-preven-jp22.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dnd): fire redock hover IPC before arm threshold to prevent ghost race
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

a5af:
Fixed in d8aa50a3. Good catch — the response handler's target-changed branch always reset `dwellSlowSince = null`, including on the initial null → target confirmation. That nulled out `onMouseUp`'s condition-2 arm so a user releasing right after the ghost appeared found nothing armed.

Fix: capture `prevTarget` before the assignment and only reset `dwellSlowSince` when `prevTarget !== null` (the slow-transit case the reset actually guards against).

PR #1584, opened 2026-06-19
feat(memory): global memory bundles seeded from workspace patterns

a5af:
## Summary

- Adds two-tier memory architecture: **Trust Center (global)** bundles are injected into every agent's `CLAUDE.md` at launch; **per-agent** bundles remain available for individual selection
- Seeds two bundles from a5af/claw workspace patterns: **Workspace Rules** (global — git, task, GitHub access tiers, safety rules) and **AgentMux Development** (opt-in — stack, build commands, version workflow)
- Global badge shown in the Memory manager list rail and detail view so users can identify which bundles auto-inject

## Changes

**Backend**
- `migrations.rs` — adds `is_global` column to `db_memory_bundles` (fresh install + ALTER TABLE for existing DBs)
- `memory_bundles.rs` — adds `is_global: bool` to `Memory` struct; adds `bundle_memory_list_global()` method
- `app_api.rs` — `write_agent_config_files()` now injects global bundle `instructions` into the `memory` content slot of every agent's `CLAUDE.md`
- `agent_seed.rs` — seeds `Vec<SeedMemory>` from the manifest; runs unconditionally on startup (upsert is idempotent)
- `agent-seed.json` — regenerated at version 10 with two seeded memory bundles

**Frontend**
- `gotypes.d.ts` — adds `is_global?: boolean` to `Memory` type
- `memory-manager.tsx` — shows `Global` chip in list items and a "Scope" row in the read-only detail view
- `memory-view.scss` — styles `.memory-view-global-badge` accent chip

## Test plan

- [ ] Fresh install: verify `db_memory_bundles` has `is_global` column and two seeded rows appear in Memory manager
- [ ] Existing install: verify `ALTER TABLE` migration runs without error and seeded rows appear
- [ ] Open any agent: verify `CLAUDE.md` contains the Workspace Rules instructions block
- [ ] Memory manager: verify "Workspace Rules" shows the Global badge; "AgentMux Development" does not
- [ ] Click "Workspace Rules" in detail view: verify "Scope: Global — injected into every agent at launch" row appears
- [ ] Create a non-global bundle: verify it does NOT appear in any agent's `CLAUDE.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1585, opened 2026-06-19
fix(memory): re-enable autoMemoryEnabled so agents can write native memory

AgentA-asaf:
## Summary

Reverts the `autoMemoryEnabled: false` default set in #1584.

- **Why it was set:** prevent autonomous MEMORY.md writes from accumulating alongside AgentMux's bundle system
- **Why it was wrong:** GitHub issue #63903 (open, unfixed upstream) confirms the 11,300-token memory preamble loads regardless of this flag — we were paying the full token cost with no write benefit
- **New architecture:** clean layer separation by content type (see spec below), so blocking writes is unnecessary

## Two-layer memory model

| Layer | Content | Written by |
|---|---|---|
| Global bundles → `CLAUDE.md` | Rules, policies, standards | Human via AgentMux Trust Center |
| Native memory (`memory/*.md`) | Facts, discoveries, session insights | Claude autonomously + user via memory modal |

No overlap by design. No concurrent-write risk: each AgentMux agent has an isolated working directory, giving it its own `~/.claude/projects/<path>/memory/` with no shared writers.

## Spec

`docs/specs/SPEC_AGENT_PANE_MEMORY_IDENTITY_MODALS_2026_06_19.md` — the upcoming memory modal PR that surfaces this folder in the agent pane UI.

## Test plan

- [ ] Launch an agent and verify `autoMemoryEnabled` is no longer forced to `false` in `.claude/settings.json`
- [ ] Confirm Claude can write to native memory during a session
- [ ] Verify global bundle content still injects into `CLAUDE.md` (unaffected by this change)

### DIFF
--- a/.changesets/1781650205-feat-mcp-loop-loopstop-tools-recurring-prompt-injection-clau-6s1b.md
+++ b/.changesets/1781650205-feat-mcp-loop-loopstop-tools-recurring-prompt-injection-clau-6s1b.md
@@ -1,5 +1,0 @@
----
-type: minor
----
-
-feat(mcp): Loop + LoopStop tools — recurring prompt injection (Claude /loop analogue)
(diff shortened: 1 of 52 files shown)


a5af:
@codex review

PR #1589, opened 2026-06-20
feat(agent): persist agent failure state in block meta (P1.1/P1.2)

AgentA-asaf:
## Problem

`agentfailure` WPS events were fire-and-forget (`persist:0`). For a 401 auth failure the subprocess exits in ~2 seconds — if the user glances away or the tab isn't actively watching, the recovery banner evaporates. The user sees **"Worked"** (the send-accepted toast) and then **silence**. No auth error, no "Login Again" button, nothing.

Root cause confirmed via blockfile inspection: the `classify()` path is correct and DOES emit the event, but `persist:0` means it's only delivered to real-time subscribers. The failure event and the meta write were both missing.

Retro: two prior specs ([Jun 11 diagnostics](#1353), [Jun 16 recovery UI](#1464)) both assumed delivery — neither questioned the `persist:0` guarantee. Full retro in `docs/specs/SPEC_AGENT_ERROR_FRAMEWORK_2026_06_20.md §1`.

### P1.1 — backend (`core.rs` + `subprocess.rs`)
- Add `META_LAST_FAILURE = "agent:last_failure"` constant
- Add `persist_last_failure(block_id, failure, wstore, event_bus)` helper in `core.rs`:
  - `Some(failure)` → writes JSON to block meta
  - `None` → sets null (triggers `merge_meta` delete path) to clear on clean exit
  - Broadcasts `waveobj:update` so live subscribers see the change via the block atom
- Capture `wstore` + `event_bus` in the `process_waiter` closure
- Call `persist_last_failure()` before the WPS publish
- Bump `persist:0 → persist:1` on the `agentfailure` event (belt-and-suspenders)

### P1.2 — frontend (`useAgentFailure.ts` + `gotypes.d.ts`)
- Add `"agent:last_failure"?: AgentFailure` to `MetaType`
- On pane mount, read `getBlockMetaKeyAtom(blockId, "agent:last_failure")` and seed `setFailure()` — the recovery banner now appears on any cold load without the WPS event

## Test plan
- [ ] Send a message to an agent with expired/missing credentials → auth failure banner appears within 2s
- [ ] Switch to another tab, switch back → banner still visible (recovered from meta)
- [ ] Hard-reload the app → banner still visible on pane open
- [ ] Successful agent turn → banner clears
- [ ] Transient failure (rate-limit) → auto-retry still fires; banner clears on success
- [ ] `cargo check -p agentmux-srv` — no errors
- [ ] `tsc --noEmit` — no new errors

## Spec
`docs/specs/SPEC_AGENT_ERROR_FRAMEWORK_2026_06_20.md` — full retro + P1–P2 roadmap (P1.3 inline error nodes, P1.4 health transition nodes, P1.5 system notification are follow-on PRs)


a5af:
@codex review

AgentA-asaf:
Addressing reagent P2: changeset type is intentionally `patch` per project convention (CLAUDE.md / `feedback_use_patch_bumps_for_iteration` memory): all iteration changesets use `patch`; `minor` is reserved for milestone releases. The bump type does not affect the API shape delivered to users — it only controls the release version increment when these changesets are consumed. No code change needed.

PR #1590, opened 2026-06-20
feat(agent): inline API error node in transcript (P1.3)

AgentA-asaf:
## Summary

- Adds `AgentErrorNode` + `ErrorResultEvent` types to the agent stream type system
- `ClaudeTranslator` now emits an `error_result` event when the CLI `result` frame has `is_error:true` + `api_error_status` (e.g. 401 auth failures, 429 rate-limits)
- `StreamParser` converts `error_result` → `AgentErrorNode` and inserts it inline in the transcript — same position as where the session ended
- `DocumentRow` renders the error node as a compact left-bordered block (HTTP code + full error message)
- `renderers.ts` adds `agent_error` to `STREAMING_CAPABLE`, `estimateNode`, `estimateNodeForState`, and `NodeRendererRegistry`

## Why

Before this PR, auth/rate-limit errors produced by the Claude CLI appeared in the backend blockfile but were silently discarded by the translator (`type:"result"` was handled as a `session_end` only). The user saw "Worked" (message accepted) and then nothing — no indication the turn failed. This is the frontend-side delivery of the P1 error framework (spec §P1.3).

**Companion PRs:** P1.1/P1.2 (#1589) persist the failure to block meta so the FailureRow banner survives tab switches; P1.3 makes the error visible directly in the transcript stream.

## Test plan

- [ ] Send a message to an agent with `apiKeySource:"none"` (no valid credentials); verify a red `HTTP 401` error block appears in the transcript below the user message
- [ ] Send a message to a correctly-authenticated agent; verify no error block appears
- [ ] Verify the `session_end` event still fires (cost/turn stats still update) even when an error node is emitted ahead of it
- [ ] `npx tsc --noEmit` passes with no new errors on changed files
- [ ] Existing tests pass (`npm test`)


a5af:
@codex review

PR #1592, opened 2026-06-20
feat(agent): inline "Login Again" CTA on 401/403 error nodes (reauth P2.3)

AgentA-asaf:
## What

When an agent hits an auth error, the inline `agent_error` transcript node (added in #1590) now shows a **"Login Again →"** button for recoverable auth statuses (401 Unauthorized / 403 Forbidden). Clicking it re-runs the provider login flow — the same `status.startLaunchFlow()` action the failure banner already drives.

Errors with no in-place fix (any non-auth code, and code `0` = non-HTTP network/CLI errors) render **no** button.

## Why

Closes the gap from the auth-failure-visibility work: the user can now recover from an expired Claude OAuth **directly from the error in the transcript**, not just the banner. Two entry points, identical behavior.

## How

- New optional `onAgentErrorLogin` callback threaded `agent-view.tsx → AgentDocumentView → AgentDocumentVirtualList → DocumentRow`.
- `DocumentRow` renders the CTA only when the callback is present **and** `code ∈ {401, 403}`.
- SCSS: right-aligned outlined button matching the error block's error-color theme.

## Scope

First slice of **SPEC_REAUTH_FROM_AUTH_ERROR_2026_06_20** (§7), included in this PR (rides with code per the no-doc-only-PR rule). The `ReauthBrowserModal` pieces (§6 — open OAuth in a CEF browser window with URL fallback, P2.1/P2.2) follow in a separate PR.

## Tests

`DocumentRow.test.tsx` — 5 cases: 401 renders CTA + fires callback on click; 403 renders CTA; 500 renders none; code 0 renders none + shows "Error" not "HTTP 0"; missing callback renders none.

### DIFF
--- a/.changesets/1781921078-feat-agent-inline-login-again-cta-on-401-403-error-nodes-rea-wshy.md
+++ b/.changesets/1781921078-feat-agent-inline-login-again-cta-on-401-403-error-nodes-rea-wshy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): inline Login Again CTA on 401/403 error nodes (reauth P2.3)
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #1594, opened 2026-06-20
feat(agent): open OAuth re-auth in an in-app browser pane with URL backup (reauth P2.1)

AgentA-asaf:
## What

When a CLI login surfaces an OAuth URL, AgentMux now opens it in an **in-app browser pane** (`createBlock({ meta: { view: "browser", url } })`) instead of immediately punting to the system browser. The pane splits the current tab so the agent's **AuthUrlBox** (URL text + Copy + the paste-the-code input) stays visible beside it — needed for providers like Claude that hand back an authorization code to paste rather than redirecting.

System browser (`openExternal`) remains the **fallback** if the pane can't be created. The helper never throws.

## Why

This is the "browser window + url as backup" the user asked for after the auth-failure-visibility work. The in-app browser keeps the OAuth flow inside AgentMux and right next to the paste-code box.

## How

- **New helper** `frontend/app/view/agent/flows/open-oauth-pane.ts` — `openOAuthBrowserPane(url)` returns `"pane" | "external" | "failed"` (createBlock → openExternal → nothing).
- `launch-flow.ts` swaps its `openExternal(loginUrl)` for the helper and logs which path opened. Applies to **all** CLI logins (first launch + re-auth) since they share the flow.
- `AuthUrlBox` gains an **"Open"** button (beside Copy) to re-open the URL in a pane on demand.

## Design note (modal → pane)

The spec originally proposed a `ReauthBrowserModal` with an embedded CEF browser. A codebase study found that's **architecturally impossible**: a browser pane is a real native HWND child window positioned in physical pixels synced to layout, and can't be CSS-positioned inside a modal overlay. The spec is revised to the browser-pane approach (**§6′**), which is also the idiomatic way AgentMux opens content. Spec rides with this code (no doc-only PRs).

## Tests

`open-oauth-pane.test.ts` — pane success (asserts the browser BlockDef), system-browser fallback when the pane throws, and the `"failed"` no-throw path.

## Series

- #1592 — P2.3 inline 401/403 "Login Again" CTA (**merged**)
- **this PR** — P2.1 in-app browser pane + URL backup
- Spec: `SPEC_REAUTH_FROM_AUTH_ERROR_2026_06_20.md`

### DIFF
--- a/.changesets/1781921659-feat-agent-open-oauth-re-auth-in-an-in-app-browser-pane-with-ovni.md
+++ b/.changesets/1781921659-feat-agent-open-oauth-re-auth-in-an-in-app-browser-pane-with-ovni.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): open OAuth re-auth in an in-app browser pane with URL backup (reauth P2.1)
--- a/docs/specs/SPEC_REAUTH_FROM_AUTH_ERROR_2026_06_20.md
+++ b/docs/specs/SPEC_REAUTH_FROM_AUTH_ERROR_2026_06_20.md
@@ -1,8 +1,17 @@
 # SPEC: Re-authentication from Agent Auth Failure
 **Date:** 2026-06-20  
 **Author:** AgentA  
-**Status:** Draft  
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1599, opened 2026-06-20
fix(sysinfo): throttle chart SVG rebuild to fix sustained GPU/renderer CPU

a5af:
## Root cause

The sysinfo chart was rebuilding its entire Observable Plot SVG every second (on every sysinfo event at 1Hz). Profiling the stable build after 24h of uptime revealed:

- **GPU process (13.4% CPU sustained)**: `CA::Transaction::commit()` / `CA::Layer::commit_if_needed()` fired on every sysinfo update — the compositor committing the freshly rebuilt SVG layer on every tick
- **Renderer Compositor thread (6.9% CPU)**: `fontations_ffi BridgeHintingInstance` — font hinting for the x-axis tick labels every second; `temporal_rs_PlainTime_millisecond` — Temporal API date formatting for those labels

With `"h:mm:ss A"` format every label was unique per second, making glyph-cache reuse impossible. The full SVG tear-down + rebuild + font rasterization ran at 1Hz in perpetuity.

## Fix

**`sysinfo-view.tsx`**: Replace the reactive `createMemo(() => model.dataAtom())` (fires on every sysinfo event) with a `createSignal` + `setInterval(2000)` throttle. The chart updates every 2 seconds instead of every second. The 5-minute history chart makes a 2s lag imperceptible.

The status-bar `SystemStats` component is unaffected — it subscribes to sysinfo events directly and does cheap text-only DOM updates (no SVG, no font hinting), so it still refreshes at full 1Hz.

**`sysinfo-plot.tsx`**: Change x-axis `tickFormat` from `"h:mm:ss A"` to `"h:mm A"`. Shorter strings, and labels repeat once per minute allowing the browser to reuse cached glyph bitmaps. The interactive hover tooltip keeps second precision (renders only on user mouse movement, not continuously).

## Expected impact

- GPU process CPU: ~13% → <5%
- Renderer CPU: ~7% → <3%
- Total sustained AgentMux idle CPU: ~20% → <8% on two open windows

## Test plan

- [ ] Open AgentMux with sysinfo pane visible
- [ ] Verify chart still updates (refreshes every ~2s)
- [ ] Verify x-axis shows `h:mm` format (no seconds on ticks)
- [ ] Hover over data points — confirm tooltip still shows `h:mm:ss A` with full precision
- [ ] Let run for 5+ minutes and confirm Activity Monitor shows reduced GPU Helper CPU

### DIFF
--- a/.changesets/1781923404-fix-sysinfo-throttle-chart-svg-rebuild-from-1hz-to-0-5hz-to--y3mg.md
+++ b/.changesets/1781923404-fix-sysinfo-throttle-chart-svg-rebuild-from-1hz-to-0-5hz-to--y3mg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(sysinfo): throttle chart SVG rebuild from 1Hz to 0.5Hz to fix sustained GPU CPU
--- a/frontend/app/view/sysinfo/sysinfo-plot.tsx
+++ b/frontend/app/view/sysinfo/sysinfo-plot.tsx
@@ -138,7 +138,7 @@ function SingleLinePlot(props: SingleLinePlotProps): JSX.Element {
             x: {
                 grid: true,
                 label: "time",
-                tickFormat: (d: number) => `${dayjs.unix(d / 1000).format("h:mm:ss A")}`,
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1604, opened 2026-06-20
fix(agent): force re-login on "Login Again" instead of trusting the auth-status check (reauth P2.5)

AgentA-asaf:
## The bug

Dogfooding v0.46.6: an agent 401s, the user clicks **Login Again**, and **nothing happens** — no browser, no error, no change.

## Root cause (confirmed from live logs)

The click ran `startLaunchFlow()`, whose Phase 2 trusts `CheckCliAuth`:

```
ResolveCli claude            ✓
CheckCliAuth claude.cmd       ✓  → authenticated:true   ← FALSE POSITIVE
needsLogin = false           → login skipped (no browser opens)
ControllerResync force=false → no-op on the running persistent controller
```

`claude auth status` only confirms a credential is **present**, not **valid**. An expired/revoked token still reports `authenticated:true`, so the gated flow trusts it, skips login, and the non-forced resync no-ops. The **401 itself** is the ground truth that auth is broken (and P1.3 already surfaces it) — not the status check.

## Fix

The explicit "Login Again" actions now **force** a login, bypassing the check (the 401 already told us the token is bad):

- **New `forceProviderLogin()`** (`flows/force-login.ts`): runs `runCliLogin` unconditionally → `openOAuthBrowserPane` (in-app browser pane, #1594) → `setAuthUrl`. **Never** calls `CheckCliAuth`.
  - Deliberately **omits the status-check success-poll** the gated flow uses: with a present-but-expired token that poll would "succeed" on the first tick and reap the in-flight login CLI before the user finishes.
- **`useAgentControllerStatus.relogin()`** reads the CLI path + auth env from block meta (`cmd` / `cmd:env`, written at launch) and calls the helper.
- Both **"Login Again" entry points** (failure banner `onLoginAgain`, inline error node `onAgentErrorLogin`) call `relogin()` instead of `startLaunchFlow()`.
- **`/login` refactored** onto the same helper (DRY; also gains the in-app pane).

`startLaunchFlow()` keeps trusting `CheckCliAuth` for **first launch**, where it's reliable (genuinely no credential → correctly returns false). The running persistent agent re-reads its credential per request, so the next message uses the fresh token and clears the failure row.

## Verification

- Diagnosed live on a dev build (logs in spec §11); `/login` was the manual workaround (it already bypassed the gate).
- `tsc` clean, `task build:frontend` green.
- `force-login.test.ts` — 4 cases: runs `runCliLogin` with the right args + opens the URL; **never** calls the auth-status check; warns + no `setAuthUrl` on a null URL; resolves on system-browser fallback.

## Series

P1.1/P1.2 (#1589) · P1.3 (#1590) · P2.3 inline CTA (#1592) · P2.1 browser pane (#1594) · **P2.5 force-login (this PR)** · Spec: `SPEC_REAUTH_FROM_AUTH_ERROR_2026_06_20.md` §11

## Follow-up (noted in spec)

Force-restart / health-revalidate after login so providers that cache the token in-process pick it up without a manual message resend.

### DIFF
--- a/.changesets/1781929749-fix-agent-force-re-login-on-login-again-instead-of-trusting--a6vg.md
+++ b/.changesets/1781929749-fix-agent-force-re-login-on-login-again-instead-of-trusting--a6vg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): force re-login on Login Again instead of trusting the auth-status check (reauth P2.5)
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

PR #1611, opened 2026-06-20
fix(linux): prefer native Wayland ozone on CEF 148 instead of forcing XWayland

AgentU-asaf:
## Background
The Linux host forced `--ozone-platform=x11` (XWayland) to work around a **CEF 146 / Chromium 146** native-Wayland bug:
- `WaylandZwpLinuxDmabuf::OnTrancheFlags Not implemented` at startup (the dmabuf/GPU-buffer handshake with Mutter)
- → `cc::Scheduler` replied `DidNotProduceFrame` to **~89%** of Mutter's `BeginFrame` requests
- → `requestAnimationFrame` starved → terminal typing visibly **hung and burst** (predictive-echo path, #1223/#1241)

The inline comment even anticipated this: *"Future CEF 148 + a forward-ported Wayland-aware libcef.so may make native Wayland viable."*

## What changed
We're on **CEF 148** now, and the bug no longer reproduces. So instead of pinning XWayland:

- When the session exposes `WAYLAND_DISPLAY` → select **native Wayland**.
- Otherwise → append nothing, let Chromium use its **X11 default**.
- `AGENTMUX_OZONE_PLATFORM=<wayland|x11>` still **pins** a specific backend (regression testing / per-machine override).

One file: `agentmux-cef/src/app.rs` (the Linux ozone block) + a `patch` changeset.

## Evidence (CEF 148, GNOME/Mutter)
Launched with native Wayland engaged and captured `cef-debug.log`:

| Check | CEF 146 (old) | CEF 148 (now) |
|---|---|---|
| `Not implemented` callbacks (the bug cluster) | OnTrancheFlags, WaylandSeat::OnName, XdgToplevel… | **0** |
| `DidNotProduceFrame` / GPU fallback | ~89% of BeginFrame | **0** |
| dmabuf line | `OnTrancheFlags Not implemented` (fatal) | `Binding to zwp_linux_dmabuf_v1 version 4` (benign warning) |

Then verified the **new default** (no env override) brings up native Wayland on its own: 7 `ui/ozone/platform/wayland/` log lines, `0` `Not implemented`, `0` frame-production distress. Typing in a terminal pane confirmed smooth (the predictive-echo render path is the one that used to stall).

## Caveat
Validated so far only on a **GNOME/Mutter VM (vmwgfx)**. The ANGLE GPU-tier selection, OOM `--disable-gpu` recovery, and other GPU paths are untouched. Broader testing (other compositors, real GPUs, X11-only sessions) is welcome before this is fully relied on — and the `AGENTMUX_OZONE_PLATFORM` override remains as an escape hatch.

## Test plan
- [ ] Wayland session (GNOME/Mutter): launches native Wayland, terminal typing smooth, no `Not implemented` / `DidNotProduceFrame` in `cef-debug.log`.
- [ ] X11-only session: still launches (Chromium X11 default, no forced switch).
- [ ] `AGENTMUX_OZONE_PLATFORM=x11` still pins XWayland; `=wayland` still pins Wayland.
- [ ] Other compositors (KDE/Plasma, wlroots) + real-GPU hardware.

### DIFF
--- a/.changesets/1781955205-fix-linux-prefer-native-wayland-ozone-on-cef-148-instead-of--o0wn.md
+++ b/.changesets/1781955205-fix-linux-prefer-native-wayland-ozone-on-cef-148-instead-of--o0wn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(linux): prefer native Wayland ozone on CEF 148 instead of forcing XWayland
--- a/agentmux-cef/src/app.rs
+++ b/agentmux-cef/src/app.rs
@@ -596,44 +596,37 @@ wrap_app! {
                 );
                 cmd.append_switch_with_value(Some(&key), Some(&val));
 
-                // ── Linux: default to XWayland (X11 ozone) ─────────────────
-                // Native-Wayland Ozone in CEF 146/Chromium 146 has broken
-                // Mutter↔Chromium GPU buffer negotiation
-                // (`WaylandZwpLinuxDmabuf::OnTrancheFlags Not implemented`
-                // at startup), which causes the cc::Scheduler to reply
-                // `LayerTreeHostImpl::DidNotProduceFrame` to ~89 % of
-                // Mutter's `BeginFrame` requests. With BeginMainFrame
-                // production stuck near sysinfo's ~1 Hz invalidation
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1613, opened 2026-06-20
fix(auth): host CLI login capture for Claude Code v2.1.x — instrument + seed-from-global

AgentA-asaf:
Follow-up to #1604. Spec: `docs/specs/SPEC_HOST_CLI_LOGIN_CAPTURE_2026_06_20.md`.

## Problem
Claude Code CLI **v2.1.183** turned `claude auth login` into a self-driving TUI: it opens its own browser + localhost OAuth callback and **copies the login URL to the clipboard on `c`** — it no longer prints a scrapeable `https://…` line. The host's PTY `extract_url` scrape (`run_cli_login_pty`) therefore never fires, so after the #1604 force-login fix, **"Login Again" calls `run_cli_login` but no browser/URL box appears**. (`extract_url` already strips ANSI/OSC-8, so this is *not* a parsing gap — there is simply no URL to scrape.)

Research against the current Anthropic docs surfaced three officially-supported, non-scrape paths (`setup-token` → `CLAUDE_CODE_OAUTH_TOKEN`, the v2.1 paste-code fallback, and the env/credential precedence rules). This PR lands the two pieces that don't require an interactive capture run.

## Slice 1 — §4 instrumentation
- `run_cli_login_pty` now logs **every PTY line from the first byte** (was: only *after* a URL was captured, so a login that never prints a URL was a total black box).
- Both spawn paths log an **auth-env precedence snapshot** — `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` / `CLAUDE_CODE_OAUTH_TOKEN` = `override|inherited|absent` (names/states only, never values) — to check the "loggedIn but 401" precedence hypothesis (a stray `ANTHROPIC_API_KEY` overrides subscription OAuth). Visible via `muxlog host grep login-pty`.

## Slice 2 — §5.5 "Use my existing Claude login" (ship-now recovery)
- New host IPC command **`seed_provider_auth_from_global`**: reads the user's GLOBAL Claude credential (`$CLAUDE_CONFIG_DIR` if the host inherits one, else `~/.claude/.credentials.json`), validates `claudeAiOauth.expiresAt` is in the future (mirrors `agentmux-srv` `identity::resolver::probe_oauth_status`), and copies it verbatim — incl. `refreshToken` — into the agent's isolated dir (`~/.agentmux/shared/providers/claude/.credentials.json`) via temp+rename. Returns `{ seeded, status: seeded|missing|expired, expiresAt }` with **no token material**.
- Frontend: `seedProviderAuthFromGlobal` API, a `seedGlobalLogin` flow, `useGlobalLogin` on `useAgentControllerStatus`, and a 🌐 **"Use existing login"** action beside "Login Again" on the 401/403 failure CTA. No controller restart — the running agent re-reads its credential per request, so the next message clears the failure row.

## Verification
- `vitest` — **18/18** on the two affected suites (incl. a new CTA assertion)
- `tsc --noEmit` — **zero errors in any touched file** (pre-existing errors are all in unrelated files)

## Scoped out (tracked in the spec)
- Pre-launch auth panel integration (that path goes through the srv `auth.*` state machine, not the host `runCliLogin`/failure-CTA path).
- Live smoke test (needs a real expired isolated cred + valid global on a host build).
- **5.1 (`setup-token`) / 5.2 (paste-code)** OAuth-capture fixes — sequenced *after* a §4 instrumented capture run.

### DIFF
--- a/.changesets/1781950297-fix-auth-log-every-pty-login-line-and-auth-env-precedence.md
+++ b/.changesets/1781950297-fix-auth-log-every-pty-login-line-and-auth-env-precedence.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): instrument host CLI login — log every PTY line from byte 0 + auth-env precedence snapshot (host-login-capture §4)
--- a/.changesets/1781952589-fix-auth-use-existing-claude-login-seed-from-global.md
+++ b/.changesets/1781952589-fix-auth-use-existing-claude-login-seed-from-global.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): "Use existing login" — seed an agent from a valid global Claude credential instead of a fresh OAuth (host-login-capture §5.5)
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -446,6 +446,34 @@ pub async fn run_cli_login(
         .and_then(|v| v.as_bool())
         .unwrap_or(false);
 
+    // §4 instrumentation (SPEC_HOST_CLI_LOGIN_CAPTURE): snapshot the
+    // auth-precedence env keys the child will see. A stray ANTHROPIC_API_KEY
+    // (precedence #3) overrides subscription OAuth (#6) and, if it belongs to a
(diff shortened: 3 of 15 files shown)


a5af:
@codex review

PR #1614, opened 2026-06-20
fix(agent): move activity dock below conversation, above composer

a5af:
## Summary

- Relocates the `ActivityDock` from the `"dock"` region (just below the control bar, **top** of pane) to render after `"queue"` — **just above the composer** — by reordering `PANE_REGION_ORDER` in `PaneRegions.tsx`
- Lowers `max-height` from `45%` to `30%` and adds a `border-top` separator in `PaneRegions.scss` so the dock reads as part of the composer area rather than the transcript area
- Adds spec `docs/specs/SPEC_ACTIVITY_DOCK_BOTTOM_MOVE_2026_06_20.md`

**Why:** The original top placement meant activity status appeared at the furthest point from where users look after sending a message. In a tiling layout with short panes, a 2–3 row dock also consumed 15–25% of visible height and pushed the conversation top edge down whenever tasks started. Bottom placement keeps the conversation stable and surfaces task status where attention already is.

No component, data model, backend, or WPS changes — the `ActivityDock` component is position-agnostic.

## Test plan

- [ ] Start a shell task (`task dev` or any long-running shell via agent) — verify dock appears **above the composer**, not below the control bar
- [ ] Verify conversation top edge does not shift when dock appears/disappears
- [ ] Verify dock auto-hides when no tasks are running (no dead space)
- [ ] Verify expand/collapse, stop, and dismiss still work
- [ ] Verify dock scrolls internally when more than 3 rows are expanded (30% cap)

### DIFF
--- a/.changesets/1781957910-fix-agent-move-activity-dock-below-conversation-above-compos-gdng.md
+++ b/.changesets/1781957910-fix-agent-move-activity-dock-below-conversation-above-compos-gdng.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): move activity dock below conversation, above composer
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #1617, opened 2026-06-20
feat(muxbus): auto-register agents with cloud subscriber at spawn/stop

a5af:
## Summary

- Calls `cloud_subscriber::add_agent(&blockid)` immediately after `spawn_turn()` in the `agentspawn` handler so every started agent is registered for cloud injection
- Calls `cloud_subscriber::remove_agent(&blockid)` after `ctrl.stop()` in the `agentstop` handler so stopped agents are deregistered
- Also includes `docs/specs/SPEC_MUXBUS_GITHUB_REVIEW_NOTIFICATIONS_2026_06_20.md` — the full gap analysis spec

## Why

Previously, agents only registered with the cloud subscriber if they explicitly called the reactive subscribe endpoint (`POST /api/v1/reactive/subscribe`). Any agent that never called that endpoint was invisible to cloud-side injection — GitHub PR review notifications would be queued under the agent's ID but never polled and delivered.

With this change, registration is automatic at spawn, so all agents receive cloud injections from the moment they start regardless of which MCP tools they use.

## Test plan

- [ ] Spawn an agent pane — verify `cloud_subscriber` logs `add_agent smike-06122` (or equivalent block ID)
- [ ] Stop the agent — verify `cloud_subscriber` logs `remove_agent smike-06122`
- [ ] Inject a test message via `/reactive/inject` targeting the block ID — verify delivery

Part of the MuxBus GitHub review notifications MVP (see companion cloud PR).

### DIFF
--- a/.changesets/1781972947-feat-muxbus-auto-register-agents-with-cloud-subscriber-at-sp-fnyo.md
+++ b/.changesets/1781972947-feat-muxbus-auto-register-agents-with-cloud-subscriber-at-sp-fnyo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(muxbus): auto-register agents with cloud subscriber at spawn/stop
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

PR #1620, opened 2026-06-20
fix(mcp): prefer AGENTMUX_BLOCKID over AGENTMUX_AGENT_BUS_ID for shell scope

a5af:
## Root Cause

The `block_id` used by `agentmux-mcp` for `shell/create` requests was resolved in the wrong order:

```rust
// Before (broken):
let bus = std::env::var("AGENTMUX_AGENT_BUS_ID").unwrap_or_default();
if bus.is_empty() { AGENTMUX_BLOCKID } else { bus }
```

`AGENTMUX_AGENT_BUS_ID` is set in every agent's `.mcp.json` as the MuxBus routing identifier — typically `"claude"` (the agent type, not a block UUID). Because it's non-empty, agentmux-mcp used it as `agent_block_id` in shell create requests. The backend then published `shell_node_create` under `block:claude` instead of `block:<real-pane-uuid>`. The frontend subscribes to `block:<real-uuid>` and never received the event — **the ActivityDock stayed empty for all persistent (stream-json) Claude Code agent panes**.

## Fix

Flip the priority: read `AGENTMUX_BLOCKID` first (the canonical block UUID injected by `agent_handlers.rs` into every persistent subprocess env), falling back to `AGENTMUX_AGENT_BUS_ID` only for older deployments that pre-date the block ID injection.

```rust
// After (fixed):
let blockid = std::env::var("AGENTMUX_BLOCKID").unwrap_or_default();
if blockid.is_empty() { AGENTMUX_AGENT_BUS_ID } else { blockid }
```

## Test plan

- [ ] Ask a Claude Code agent to run a long-running shell (`ping -n 15 localhost`)
- [ ] Verify the ActivityDock appears **above the composer** while the shell is running
- [ ] Verify the dock row shows the shell command title and running status
- [ ] Verify the dock auto-hides 3s after the shell exits
- [ ] Verify `muxlog srv grep shell.create` shows the real block UUID (not `"claude"`) in the `block_id` field

### DIFF
--- a/.changesets/1781976316-fix-mcp-prefer-agentmux-blockid-over-agentmux-agent-bus-id-f-5j8a.md
+++ b/.changesets/1781976316-fix-mcp-prefer-agentmux-blockid-over-agentmux-agent-bus-id-f-5j8a.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(mcp): prefer AGENTMUX_BLOCKID over AGENTMUX_AGENT_BUS_ID for shell scope
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -12,17 +12,15 @@
 //! conversation as a `ShellNode` row without blocking the agent.
 //!
 //! Env vars (inherited from agentmux-srv's agent env injection):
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1621, opened 2026-06-20
fix(win32): suppress console windows for bashwrap and shell spawns

a5af:
## Problem

On Windows, every Bash tool call and every Shell tool execution spawned a blank transparent console window. After a session with many tool calls, dozens of orphaned windows accumulated on screen.

**Root cause:** Windows auto-creates a visible console window for any `CUI`-subsystem process unless `CREATE_NO_WINDOW` is passed by the spawner. Both spawn sites were missing this:

1. `agentmux-bashwrap` — spawned by Claude Code's PreToolUse hook for every Bash tool call. Each call creates a new process; if it exits non-zero or the parent doesn't wait, the window lingers.
2. `ShellNodeRunner` (shell_node.rs) — spawns `cmd /C <command>` on Windows without `CREATE_NO_WINDOW`, so every Shell tool (`task dev`, `ping`, etc.) creates a visible cmd.exe window.

## Fix

**`agentmux-bashwrap/src/main.rs`** — add `#![cfg_attr(windows, windows_subsystem = "windows")]`. The binary communicates via pipes and HTTP only; GUI subsystem is correct and suppresses window creation entirely.

**`agentmux-srv/src/backend/shell_node.rs`** — add `creation_flags(0x0800_0000)` (`CREATE_NO_WINDOW`) to the tokio `Command` on Windows. stdout/stderr are already piped so no output is lost.

## Test plan

- [ ] Bash tool call in an agent pane → no new console window appears
- [ ] Shell tool (`ping -n 10 localhost`) → no new console window appears
- [ ] Shell output still streams correctly in the conversation
- [ ] Run `tasklist | grep agentmux-bashwrap` before and after — count doesn't grow with each tool call

**One-time cleanup of existing orphans** (run outside AgentMux): `taskkill /F /IM agentmux-bashwrap.exe`

### DIFF
--- a/.changesets/1781977194-fix-win32-suppress-console-windows-for-bashwrap-and-persiste-78b1.md
+++ b/.changesets/1781977194-fix-win32-suppress-console-windows-for-bashwrap-and-persiste-78b1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(win32): suppress console windows for bashwrap and persistent shell spawns
--- a/agentmux-bashwrap/src/main.rs
+++ b/agentmux-bashwrap/src/main.rs
@@ -1,5 +1,11 @@
 // Copyright 2026, AgentMux Corp.
 // SPDX-License-Identifier: Apache-2.0
+// Suppress console windows on Windows — bashwrap communicates via pipes and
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #983, opened 2026-05-23
feat(oauth): seed Default identity bundle from ambient ~/.claude on startup (OAuth Bundles PR E)

AgentA-asaf:
## OAuth Bundles PR E

Fifth of the 6-PR sequence (`SPEC_OAUTH_IDENTITY_BUNDLES_2026_05_22.md` §5). **The migration that makes Maks (and every other ambient-creds user) valid for real** — seeds a "Default" identity bundle whose claude/codex/openclaw bindings point at the existing `~/.claude` / `~/.codex` / `~/.openclaw` config dirs. No token movement. No fabrication. The pointer genuinely resolves.

### What changed

- **`agentmux-srv/src/identity/migration.rs`** (new) — `run_default_bundle_migration(wstore, broker, home_override)`. Enumerates oauth-class providers from the registry (`get_provider_list().filter_map(provider_class)`), so a future oauth provider auto-picks up. For each:
  - Skip if any bundle already has a binding for it (idempotent across restarts).
  - Skip if no ambient config dir exists at `<HOME>/.<auth_dir_name>/.credentials.json`.
  - Else: lazily create / reuse a fixed "Default" bundle (`id="default"`), `IdentityAccount { kind: "oauth", secret_ref: OAuthConfigDir { dir: <HOME>/<auth_dir_name> }, status: <probe result> }`, bind, publish `identitybundlebindings:changed:default`.
- **`WaveStore::instance_backfill_identity_id(new_id) -> usize`** — single UPDATE that rewrites empty + legacy `"blank"` `identity_id` rows to the Default bundle. Gated on `default_bundle_created` so the FK is always valid.
- **Startup wiring** (`main.rs:480-498`) — migration runs after broker setup + wstore bootstrap, before listeners come up. Production: `Some(&broker)`, `home_dir_override = None` (resolves via `dirs::home_dir()`). Tests inject both.
- **Probe at seed time** — uses `probe_oauth_status` from PR D, so stale ambient creds land as `expired` / `needs_reauth` (not always `valid`).

### Lazy Default-bundle creation

Important behavior: if claude is already covered by a user-created bundle and codex/openclaw have no ambient creds, the migration creates NO Default bundle. Avoids a phantom empty bundle in the UI. Test `existing_binding_in_other_bundle_skips_provider` pins this.

### Tests (6 new, all passing)

| Scenario | Behavior |
|---|---|
| Empty home (no ambient creds) | 0 seeded, no Default bundle, no back-fill |
[...]

PR #1626, opened 2026-06-20
fix(auth): provider environment isolation — agents never touch the user's ~/.claude

AgentA-asaf:
Spec: `docs/specs/SPEC_PROVIDER_ISOLATION_2026_06_20.md` · Diagnosis: `docs/reports/REPORT_AGENT_AUTH_DIVERGENCE_2026_06_20.md`

## Problem
AgentMux was designed so each provider keeps its **own** auth, isolated from the user's environment. That drifted: the **#983** Default-bundle migration pointed each Default account's `secret_ref` **directly at `~/.claude`** ("the CLI keeps reading + refreshing in place"), so every agent bound to `default` **refreshes tokens inside the user's personal login**. Two agents on the same identity diverge purely by *when* they launched (one frozen on `~/.claude`, one on a stale isolated dir) — see the diagnosis report.

Directive: *"I do not want to pollute a user's environment. Same applies to the executable."*

## Auth half (this PR)
- **§4.1** — New agents: import ambient `~/.claude` into the AgentMux dir **once** (read-only copy, sentinel-gated) and point the Default account at the **AgentMux dir** — reverses the #983 pointer-to-ambient. Tokens refresh in `~/.agentmux/shared/providers/<provider>`, never `~/.claude`.
- **§4.2** — Sweep: repoint any pre-existing Default account still pointing at `~/.claude` to the AgentMux dir (copy once) on next srv startup, so existing agents un-pollute with no manual re-login.
- **§4.3** — *Verified satisfied server-side*: `inject_identity_env` **overwrites** `CLAUDE_CONFIG_DIR` from the binding at every spawn (`resolver.rs:563`), so the migration fix propagates with no frontend change.
- **§4.4** — *Verified already-correct (#1291)*: `CheckCliAuth` validates the dir the agent runs in.
- **§4.5** — The "Use existing login" 🌐 seed now targets the agent's **resolved** dir (from `cmd:env`), **guarded** to `~/.agentmux` — a stale `~/.claude` is rejected → shared default, so the seed can never write the user's env (INV-R).

## Out of scope (follow-up, specced)
Executable isolation (§5): install-pinned-only, drop the global `claude` run fallback, pin the version.

## Verification
- `cargo check` (srv + cef) clean
- **8/8 migration tests**, incl. new `T-A1` (account points at AgentMux dir; `~/.claude` byte-unchanged) and `T-A2` (sweep repoints + idempotent)
- `tsc --noEmit` clean for all touched files

### DIFF
--- a/.changesets/1781984071-fix-auth-provider-env-isolation-no-claude-pollution.md
+++ b/.changesets/1781984071-fix-auth-provider-env-isolation-no-claude-pollution.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): provider environment isolation — agents read/refresh credentials in the AgentMux dir, never the user's ~/.claude (provider-isolation auth half: migration reversal of #983 pointer-to-ambient + sweep + hardened seed target)
--- a/agentmux-cef/src/commands/providers.rs
+++ b/agentmux-cef/src/commands/providers.rs
@@ -452,16 +452,34 @@ pub fn seed_provider_auth_from_global(
         }
     }
 
-    // ISOLATED destination — the same path `ensure_auth_dir` builds.
+    // ISOLATED destination (SPEC_PROVIDER_ISOLATION §4.5). Prefer the agent's
+    // RESOLVED config dir (passed as `config_dir` from its `cmd:env`), so a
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #1628, opened 2026-06-20
fix(editor): md preview width, first-open blank render, tab path tooltip

a5af:
Three editor fixes. Analysis: `docs/analysis/ANALYSIS_EDITOR_MD_BUGS_2026_06_20.md`.

## 1. Markdown blank on first open (the real bug)
Opening an `.md` rendered blank until you clicked **✎ Source** then **👁 Preview**.

**Root cause:** the preview binds to the `liveDoc` signal, which was only seeded as a side effect of building/restoring CodeMirror. On first open the CM container (`<div ref=…>`) mounts only *after* content finishes loading (the body `<Show>` gates on `!loadingAtom()`), so the tab-change `createEffect` ran while `containerRef` was still undefined, hit its `!containerRef` guard, early-returned, and never seeded the preview. `containerRef` wasn't reactive, so nothing re-ran when it mounted.

**Fix:** make `containerRef` a signal (`createSignal`). The effect now tracks it and re-runs when the container mounts, seeding the preview reliably. The existing tab-switch restore path (cached CodeMirror state → `setLiveDoc`) is untouched, so unsaved-edit previews aren't clobbered.

## 2. Markdown preview fixed width → fills the pane
`.editor-md-preview .markdown` had `max-width: 860px; margin: 0 auto`, a narrow centered column. Removed so it flows to the pane width (the container already supplies padding + scroll). Editor-scoped — does not affect the shared `Markdown` component.

## 3. In-editor tab: full path on hover
The tab already set a native `title` with the full absolute path, but native `title` is slow/inconsistent in CEF, and a pure-CSS tooltip would be clipped (`.editor-tab-strip` and `.editor-tab` are `overflow:hidden`). Switched to the shared Portal-based `Tooltip`. A wrapper (`.editor-tab-tip`) carries the flex sizing the tab used to own, so layout is unchanged; the tab keeps its mousedown (middle-click close) / click / dblclick (pin preview) handlers.

## Verification
- `tsc` clean for all touched files (31 pre-existing unrelated errors remain, incl. `editor-model.ts:788`).
- `stylelint`: 13 violations before == 13 after — zero introduced (all pre-existing/grandfathered in `editor-view.scss`).
- Frontend-only, no backend/schema.

Bug 1 (first-open blank) is reasoned statically from the render/effect ordering; worth a quick manual confirm in a running instance.

### DIFF
--- a/.changesets/1781984501-fix-editor-markdown-preview-width-first-open-blank-render-an-uiuf.md
+++ b/.changesets/1781984501-fix-editor-markdown-preview-width-first-open-blank-render-an-uiuf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(editor): markdown preview width, first-open blank render, and tab path tooltip
--- a/docs/analysis/ANALYSIS_EDITOR_MD_BUGS_2026_06_20.md
+++ b/docs/analysis/ANALYSIS_EDITOR_MD_BUGS_2026_06_20.md
@@ -1,0 +1,184 @@
+# Editor — Markdown rendering bugs (analysis)
+
+**Date:** 2026-06-20  
+**Scope:** `view: "editor"` pane (`frontend/app/view/editor/`)  
+**Status:** Analysis only — no code changed.
+
+Three reported issues, in priority order (Bug 3 is the real defect; 1 and 2 are
+small polish).
+
+---
+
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1629, opened 2026-06-20
feat(agent-pane): aggressive URL hyperlinking in all text node types

a5af:
## Summary

- Adds `linkify-it` (4KB, 11.9M weekly downloads — the same engine markdown-it uses) for URL detection
- Bare domains (`github.com/foo`), `localhost:3000`, IPs with ports, and `https://` links all become clickable in every text-bearing pane region
- Opens in system browser via existing `openExternal` IPC

## Changes

**New files:**
- `frontend/app/element/rehype-linkify.ts` — rehype plugin that walks HAST text nodes (skipping `<a>`, `<code>`, `<pre>` ancestors) and injects `<a href>` for every detected URL; inserted before `rehype-sanitize` in the unified pipeline so all markdown + thinking blocks get linkification automatically
- `frontend/app/element/linkified-text.tsx` — `<LinkifiedText>` SolidJS component for plain-text nodes: splits on `linkify-it` matches, renders interleaved text + `<a>` elements wired to `openLink()`

**Edited files:**
- `markdown.tsx` — adds `rehypeLinkify` to the rehype plugin array (covers MarkdownNode + reasoning/thinking blocks)
- `AgentMessageBlock.tsx` — wraps `<pre>` content with `<LinkifiedText>`
- `UserMessageBlock.tsx` — wraps `<pre>` content with `<LinkifiedText>`
- `PersistentShellBlock.tsx` — wraps shell output lines with `<LinkifiedText>`
- `_document-nodes.scss` — link styles for `.linkified-url` inside plain-text containers

## Edge cases handled
- Code blocks / inline code: ancestor check skips them
- Existing `[text](url)` markdown links: already `<a>` in HAST before plugin runs — not double-processed
- Trailing `.`, `,`, `)`: `linkify-it` strips these correctly by default
- `javascript:`/`data:` URIs: blocked by `isSafeHref()` guard in `LinkifiedText` + `rehype-sanitize` strips them from markdown
- Bare `localhost` (no port): not matched — avoids false positives

## Test plan
- [ ] Paste `check github.com/foo and localhost:3000` in user message → both clickable
- [ ] Paste `visit https://example.com today.` → link detected, trailing `.` not included
- [ ] Inline code `` `localhost:3000` `` in markdown → NOT linkified
- [ ] Fenced code block with URL → NOT linkified
- [ ] Existing `[text](url)` markdown link → unchanged, still works
- [ ] Thinking block with a URL → linkified
- [ ] Shell output with a URL (e.g. `vite dev server running at http://localhost:5173`) → clickable

### DIFF
--- a/.changesets/1781995722-feat-agent-pane-aggressive-url-hyperlinking-linkify-it-detec-xddu.md
+++ b/.changesets/1781995722-feat-agent-pane-aggressive-url-hyperlinking-linkify-it-detec-xddu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): aggressive URL hyperlinking — linkify-it detects bare domains, localhost:PORT, and https:// links in all agent pane text nodes
--- a/docs/specs/SPEC_AGENT_PANE_HYPERLINKS_2026_06_20.md
+++ b/docs/specs/SPEC_AGENT_PANE_HYPERLINKS_2026_06_20.md
@@ -301,7 +301,8 @@ match.url includes "://"    →  use as-is
 otherwise (bare domain)     →  "https://" + match.url
 ```
 
-`localhost:3000` → `https://localhost:3000`  
+`localhost:3000` → `http://localhost:3000` (loopback uses http)  
+`127.0.0.1:8080` → `http://127.0.0.1:8080` (loopback uses http)  
 `github.com/foo` → `https://github.com/foo`  
 `http://example.com` → `http://example.com` (unchanged)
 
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

a5af:
Addressed P1 and P2:

- **P1 (changeset)**: Added.
- **P2 (localhost http://)**: `normalizeHref` now checks for `localhost`, `127.x.x.x`, `0.0.0.0`, and `[::1]` and uses `http://` instead of `https://`. Fixed in both `rehype-linkify.ts` and `linkified-text.tsx`.
- **P2 (lockfile churn)**: The 138 `"dev": true` → removal changes and new peer entries (`react`, `terser`, etc.) are npm 11's lockfile normalization — they would appear on any dependency change with this npm version. The only genuinely new packages are `linkify-it`, `@types/linkify-it`, `unist-util-visit-parents`, and their transitive deps (`uc.micro`, `@jridgewell/source-map`). Re-running with `--package-lock-only` confirmed the state is already minimal for npm 11.

PR #1634, opened 2026-06-20
fix(macos): LSUIElement on launcher plist — eliminate dual Dock icon on macOS 26 Tahoe

a5af:
## Problem

On macOS 26.5.1 (Tahoe, Build 25F80), two AgentMux icons appear in the Dock: the CEF host (correct) and the launcher (incorrect). Nothing is pinned — both are live processes.

**Root cause confirmed via `lsappinfo`:**

```
pid = 90285  type="Foreground"  executable: agentmux-launcher   ← should not be here
pid = 90609  type="Foreground"  executable: agentmux-cef        ← correct, owns the tile
```

The launcher calls `setActivationPolicy(.accessory)` in `splash_mac.rs:249` to release its Dock slot before CEF claims its own. On macOS ≤15 Sequoia this worked; on macOS 26 Tahoe, `lsappinfo` still reports `type="Foreground"` for the launcher after the call — the runtime downgrade is silently ignored.

## Fix

Add `<key>LSUIElement</key><true/>` to the launcher's `Info.plist` in `scripts/package-macos.sh`. The OS never registers the launcher as Foreground, so no runtime downgrade is needed. CEF's `setActivationPolicy(.regular)` then creates the sole Dock entry.

The splash window continues to work correctly — `LSUIElement` apps can show `NSWindow`s, call `makeKeyAndOrderFront`, and animate alpha exactly as before.

Full evidence and analysis: `docs/specs/BUG_MACOS26_DUAL_DOCK_ICON_2026_06_20.md`

## Test plan

- [ ] Build with `NOTARIZE=0 task package:macos`
- [ ] Launch the new `.dmg` — verify only **one** icon appears in the Dock
- [ ] Confirm splash screen still displays during startup
- [ ] Confirm `lsappinfo list | grep agentmux` shows launcher as `type="UIElement"` and cef as `type="Foreground"`

### DIFF
--- a/.changesets/1782000000-fix-macos26-dual-dock-icon-lsuielement-launcher-vtbt.md
+++ b/.changesets/1782000000-fix-macos26-dual-dock-icon-lsuielement-launcher-vtbt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): add LSUIElement to launcher Info.plist to eliminate duplicate Dock icon on macOS 26 Tahoe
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1635, opened 2026-06-20
fix(linux): restrict pane tear-off to header only

AgentU-asaf:
## Summary

- On Linux, any click-drag anywhere on a pane body (terminal, browser, agent transcript) triggered a window tear-off. Should only fire when dragging the pane header.
- Root cause: `TileLayout.linux.tsx` registered `draggable()` on `tileNodeRef` (the full pane), so every drag anywhere fired `onDragStart` → tear-off. This dates to PR #182 which worked around a WebKitGTK constraint by widening the drag surface to the whole tile.
- Fix: register `draggable()` directly on the header element (`querySelector('[data-role="block-header"]')`). Only the header gets `draggable="true"`; the tile root gets no `draggable` attribute. No explicit `draggable="false"` on any parent → WebKitGTK is satisfied, drag/tear-off restricted to header. Same pattern as `TileLayout.win32.tsx`.
- Adds 100ms polling loop (matching win32) to handle async Show-gate header mount and ErrorBoundary element swaps.

## Test plan

- [ ] Drag on pane body (terminal text, browser, transcript) — no tear-off, text selection works
- [ ] Drag on pane header — tear-off fires, pane detaches into floating window
- [ ] In-window rearrange via header drag — pane repositions, no floater
- [ ] Cross-window header drag — pane transfers to target window
- [ ] Magnified/ephemeral pane header drag — no drag (canDrag returns false)
- [ ] ErrorBoundary recovery — drag still works after header element is swapped by Show gate

Retro: `docs/retro/2026-06-20-linux-pane-tearoff-anywhere.md`
Spec: `docs/specs/SPEC_LINUX_TEAROFF_HEADER_ONLY_2026-06-20.md`

### DIFF
--- a/.changesets/1781997497-fix-linux-restrict-pane-tear-off-to-header-whole-tile-drag-t-2uur.md
+++ b/.changesets/1781997497-fix-linux-restrict-pane-tear-off-to-header-whole-tile-drag-t-2uur.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(linux): restrict pane tear-off to header — whole-tile drag triggered tear-off anywhere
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #1610, opened 2026-06-20
feat(pool): pane tear-off pool on macOS/Linux

a5af:
## Summary

Introduces a dedicated pre-warmed pane window pool (`floating-pool-{uuid}`, `frameless=true`) for macOS and Linux. Pane tear-off latency drops from ~3 s cold path to ~30 ms.

**Why a separate pool?** `frameless` is a CEF creation-time flag and cannot be changed afterward. Tab pool windows (with title bar) cannot be repurposed as frameless pane windows without a visible chrome flash — a separate pool is required.

## What changed

**Backend:**
- `PanePoolState` in `state.rs` + `pane_pool` field in `HostState` — mirrors `PoolState`/`pool` for the new pool
- `reducer/pane_pool.rs` — 4 new reducer arms (`PanePoolWindowSpawnStart`, `PanePoolWindowReady`, `PanePoolWindowDestroyedBeforePromote`, `PopAndPromoteFrontPanePoolWindow`)
- `window_pool.rs` — `PANE_POOL_TARGET_SIZE=1`, `spawn_pane_pool_window` (URL: `?pane-pool=1`, `frameless=true`, 900×600), `init_pane_pool`, `register_pane_pool_window`, `mark_pane_pool_window_renderer_ready`, `on_pane_pool_window_destroyed`, `promote_pane_pool_window`
- `ui_tasks.rs` — `PromotePanePoolWindowTask` (non-Windows): `set_bounds` + `show` + emit `pool:pane-promote { paneId, workspaceId }`
- `floating_pane.rs` — pool-first path in `open_floating_pane_window` (non-Windows)
- `client/mod.rs` — `floating-pool-*` in `on_after_created` (init), `on_before_close` (cleanup), `on_load_end` (focus-steal guard)
- `ipc.rs` + `drag.rs` — `pane_pool_window_ready` IPC command

**Frontend:**
- `pool.ts` — `isPanePoolMode()` checks `?pane-pool=1`; `awaitPanePoolPromote()` listens for `pool:pane-promote`, injects `floatingPaneId`+`workspaceId` into URL, removes `pane-pool=1` — then `initHostNewWindow()` reattaches the workspace and wave renders `FloatingPaneWorkspace` (same URL convention as cold-path `?floatingPaneId=X&workspaceId=Y`)
- `app-init.ts` — `isPanePoolMode()` branch parallel to existing `isPoolMode()` branch

[...]

PR #1639, opened 2026-06-21
fix(ui): pane tear-off renders as full window instead of floating pane (pool path)

a5af:
## Summary

- **Bug**: After PR #1610 (pane tear-off pool), tearing off a pane produced a full AgentMux window (tab bar, widget bar, status bar) instead of a chromeless floating pane.
- **Root cause**: `IS_FLOATING_PANE` in `app.tsx` was a module-level IIFE evaluated at JS parse time — before `awaitPanePoolPromote()` could call `replaceState()` to inject `floatingPaneId` into the URL. The IIFE always saw `?pane-pool=1` (no `floatingPaneId`) and froze `IS_FLOATING_PANE=false`.
- **Fix**: Move the URL check inside `AppInner` so it evaluates at component-mount time, after the pool promote handshake has updated the URL. Cold path (non-pool) is unaffected — the URL already has `floatingPaneId` from the start in that case.

## What changed

- `frontend/app/app.tsx`: remove module-level `IS_FLOATING_PANE` IIFE; read `window.location.search` at the top of `AppInner` instead.
- `docs/retro/retro-pane-tearoff-full-window-regression-2026-06-20.md`: full retrospective.

## Test plan

- [ ] Tear off a pane (drag a pane out of the main window) — should produce a chromeless floating window with just the pane content, no tab bar or widget bar
- [ ] Tear off a tab — should still produce a full window with tabs (unaffected path)
- [ ] Cold-path floating pane (first tear-off before pool is ready) — should also render `FloatingPaneWorkspace`
- [ ] Normal windows — should render `<Workspace>` (no `floatingPaneId` in URL)

### DIFF
--- a/.changesets/1782002647-fix-cef-disable-macos-renderer-sandbox-in-dev-skip-seatbelt--kj87.md
+++ b/.changesets/1782002647-fix-cef-disable-macos-renderer-sandbox-in-dev-skip-seatbelt--kj87.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): disable macOS renderer sandbox in dev — skip Seatbelt init and pass no_sandbox=1 when libcef_sandbox.dylib is absent (task dev flat layout)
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1633, opened 2026-06-20
security(cef): Windows renderer sandbox — DLL wrapper pattern (Phase 3, #1374)

a5af:
## Summary

- **Adds `[lib]` cdylib target** to `agentmux-cef` so `bootstrap.exe` (renamed `agentmux-cef.exe`) can load us as a DLL and call `RunWinMain` with a pre-initialized `cef_sandbox_info_t *`
- **Extracts `pub fn run(sandbox_info)`** from `fn main()` into `src/lib.rs`; `main.rs` shrinks to a 3-line wrapper calling `agentmux_cef::run(null_mut())`
- **Exports `RunWinMain`** (`#[no_mangle] pub unsafe extern "system"`) on Windows + sandbox feature, forwarding bootstrap.exe's sandbox pointer into `run()`
- **Removes `target_os = "windows"` from the `no_sandbox` guard** so Windows + sandbox feature now passes `no_sandbox = 0` to `CefInitialize`
- **Updates `build:host:windows` Taskfile step** to detect the sandbox DLL and stage `bootstrap.exe → agentmux-cef.exe` + `agentmux_cef.dll → agentmux-cef.dll`; falls back to the plain binary for `--no-default-features` builds
- **Sandbox pointer plumbed** through `CefExecuteProcess` and `CefInitialize` (was `null_mut()` on both)

Depends on: PR #1622 (Phases 1+2, macOS + Linux sandbox, merged).
Spec: `docs/specs/SPEC_CEF_SANDBOX_WIN_PHASE3_2026_06_20.md`

## Test plan

- [ ] `cargo check -p agentmux-cef` — both bin and lib targets compile (verified locally ✓)
- [ ] `cargo check -p agentmux-cef --no-default-features` — sandbox-off bin-only path compiles (verified locally ✓)
- [ ] Windows `task dev`: app starts, browser pane renders, no regression
- [ ] Windows `chrome://sandbox`: renderer Sandbox = active
- [ ] `AGENTMUX_UNSAFE_NOSANDBOX=1` on Windows: sandbox skipped, warning logged
- [ ] macOS/Linux: unaffected (Phase 1+2 behavior unchanged)

PR #1642, opened 2026-06-21
fix(cef): harden two CEF-init log errors — cache_path under root + dynamic debug port

AgentA-asaf:
Spec: `docs/specs/SPEC_CEF_LOG_ROBUSTNESS_2026_06_20.md`. Two non-fatal CEF-init log errors that degrade silently — surfaced while verifying the Phase 3 sandbox build (#1633); both predate it.

## 1. `cache_path` not a child of `root_cache_path` → in-memory storage (every user)
Per-window `RequestContext` cache paths were built from `version_data_dir` (`…/data/browser-contexts/<label>`) — a **sibling** of `root_cache_path` (`…/cef-cache`). CEF requires `cache_path` to be a descendant of `root_cache_path`, so it rejected it and fell back to **in-memory** storage (pool/tear-off/browser-pane windows lost disk cache/cookies/storage on close).
**Fix:** store `root_cache_path` in `AppState.cef_cache_dir`; root `browser-contexts/<label>` under it.

## 2. `bind()` collision on the hardcoded remote-debugging port (multi-instance)
`debug_port` was hardcoded `9222`/`9223`. A 2nd instance can't bind it (`WSAEADDRINUSE` 0x2740) → no DevTools server → the browser DOM API (`/agentmux/browser/*`) can't connect for that instance. AgentMux is designed to run many instances in parallel (isolation I1–I6).
**Fix:** bind-probe the preferred port, fall back to an OS-assigned free port, store the **actual** port in `AppState.debug_port` (which `browser_api` already reads).

## Verification (real build, other instances holding 9222)
- `CEF remote-debugging port: 52489 (preferred 9222)` — fell back to a free port; **0** `tcp_socket_win.cc:530 bind()` errors.
- **0** `context.cc:161/185` errors; contexts now under `cef-cache/browser-contexts/window-pool-*` (disk-backed), **none** under `data/`.

`cargo check -p agentmux-cef` clean.

### DIFF
--- a/.changesets/1782003483-fix-cef-cache-path-and-dynamic-debug-port.md
+++ b/.changesets/1782003483-fix-cef-cache-path-and-dynamic-debug-port.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): root per-window browser contexts under root_cache_path (stop in-memory fallback) + pick a free remote-debugging port per instance (stop multi-instance bind() collision)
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1644, opened 2026-06-21
fix(cef): restore the AgentMux icon on Windows sandbox builds (#1633 regression)

AgentA-asaf:
The Phase 3 Windows sandbox (#1633) makes the host process CEF's **bootstrap.exe**, which ships Chrome's icon and carries none of the winres-embedded AgentMux resources (those went into the cdylib + the old raw bin). Two surfaces showed Chrome.

## 1. Running window / taskbar / Alt-Tab
`set_window_icon` used `GetModuleHandleW(null)` → the *process exe* (bootstrap.exe) → no resource ID 1 → "no icon found" → Chrome.
**Fix:** resolve the module that *contains the code* via `GetModuleHandleExW(FROM_ADDRESS)` — the cdylib (sandbox build) or the exe (non-sandbox), whichever carries the embedded icon.
**Verified:** host log now logs `Set window icon from embedded resource` (was "no icon found"). User confirmed visually.

## 2. Host exe-file icon (Explorer / Task Manager / Alt-Tab thumbnail)
winres can't edit an already-built PE, so we stamp the AgentMux icon onto the staged host exe at package time with electron's **rcedit** (pinned by sha256, cached under `~/.agentmux/build-tools`, idempotent on the raw bin). Wired into `package-portable.sh` **and** `build:host:windows`.
rcedit can transiently fail (`Unable to commit changes`) when Defender / the search indexer still holds a handle on the just-copied exe — the same race that bites CEF extraction — so the helper **retries, then degrades to a warning** (a cosmetic stamp must never break packaging; the runtime window icon is set independently).
**Verified:** the stamped exe icon hash (`530460D2…`) differs from the un-stamped Chrome bootstrap (`84303ABE…`).

## Installer
Already correct — `SetupIconFile=agentmux.ico`, `AppExe=agentmux.exe` (the *launcher*, which has the icon), shortcuts use `IconFilename: agentmux.ico`. The installed host exe inherits the stamp via the packaged portable. No change needed.

## Files
- `agentmux-cef/src/client/wndproc.rs` — Fix A (`GetModuleHandleExW(FROM_ADDRESS)`)
- `scripts/inject-exe-icon.sh` (new) — rcedit helper (pinned, cached, retry+degrade)
- `scripts/package-portable.sh`, `Taskfile.yml` — wire the stamp into both build paths

`cargo check -p agentmux-cef` clean; built + verified on a real portable.

### DIFF
--- a/.changesets/1782005606-fix-sandbox-taskbar-and-exe-icon.md
+++ b/.changesets/1782005606-fix-sandbox-taskbar-and-exe-icon.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): restore the AgentMux icon on Windows sandbox builds (Chrome-icon regression from #1633). Load the window/taskbar icon from the host module (cdylib) via GetModuleHandleExW(FROM_ADDRESS) instead of the bootstrap.exe process exe, and stamp the AgentMux icon onto the bootstrap.exe host at package time with rcedit (fixes the Explorer / Task Manager / Alt-Tab exe-file icon).
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -493,6 +493,10 @@ tasks:
                 } else {
                   Copy-Item target/release/agentmux-cef.exe dist/cef/agentmux-cef.exe -Force
                 }"
+            # Stamp the AgentMux icon onto the staged host exe — bootstrap.exe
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1648, opened 2026-06-21
fix(linux): center the startup splash on Wayland via the XWayland backend

AgentU-asaf:
## Problem
On native-Wayland sessions the startup splash (the pulsing brain) appears in the **top-left** of the screen instead of centered.

## Root cause
Since #1611 the host defaults to native Wayland, and the splash *mirrored the host's ozone choice* — so it took the native-Wayland `xdg_toplevel` backend. Wayland forbids a client from positioning its own toplevel, and the spec's assumption that "Mutter centers small toplevels" is **false**: Mutter drops them at the top-left.

The X11 backend, by contrast, draws an **override-redirect** window it positions itself (centered on the RANDR primary), and XWayland honors that — it's the pre-#1611 behavior that always centered.

## Fix
Choose the splash backend by **centering ability**, not by mirroring the host (the two need not share a display protocol — dismiss is via the ready-file, not the wire). `detect()` now prefers the self-centering X11/XWayland backend whenever an X server is actually reachable (probed via a real connection, so a stale `DISPLAY` can't strand the splash), and falls back to the native-Wayland backend only when there is no X display. `AGENTMUX_OZONE_PLATFORM=x11|wayland` still pins a backend explicitly.

## Verification
Empirically confirmed on this GNOME/Mutter Wayland session: a standalone override-redirect window at the computed center lands at **exactly (585, 1162)** on the 1354×2509 primary — i.e. Mutter/XWayland honors override-redirect positioning. `cargo check -p agentmux-launcher` is clean.

## Files
- `splash_linux/mod.rs` — `detect()` now picks by centering ability + probes a live X server
- `splash_linux/x11.rs` — adds `server_reachable()` probe
- `splash_linux/wayland.rs` — doc: now the no-X-server fallback (can't self-center)
- spec `SPEC_LINUX_SPLASH_SESSION_AWARE_2026_06_20.md` — corrected the false "Mutter centers" claims

### DIFF
--- a/.changesets/1782035709-fix-linux-center-the-startup-splash-on-wayland-by-routing-it-zleg.md
+++ b/.changesets/1782035709-fix-linux-center-the-startup-splash-on-wayland-by-routing-it-zleg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(linux): center the startup splash on Wayland by routing it through the self-centering XWayland backend (Mutter drops uncentered toplevels top-left)
--- a/agentmux-launcher/src/splash_linux/mod.rs
+++ b/agentmux-launcher/src/splash_linux/mod.rs
@@ -8,9 +8,14 @@
 //! over a dark backdrop in that gap — the Linux analogue of `splash.rs` (Win32)
 //! and `splash_mac.rs` (Cocoa).
 //!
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1650, opened 2026-06-21
fix(srv): seed default 3-pane layout for new windows (no more blank "Open another window")

a5af:
## Problem

Status bar → **"Open another window"** opens a **blank** window. It should have the default 3-pane launch layout (agent + sysinfo + swarm), same as the first window.

## Root cause

The default layout was only ever built in `wcore::ensure_initial_data` — the **first-launch** path (empty store). "Open another window" takes a different path:

> `openNewWindow()` → host pool `pool:new-window` → `initHostNewWindow` → `WindowService.CreateWindow(null, "")` → reducer `CreateWorkspace{name}` + `CreateTab`

…which creates a workspace whose tab has **`rootnode = None`** (`service.rs:749` "default rootnode = None matches wcore behaviour"). The frontend's `CreateWorkspace(name, icon, color, applyDefaults)` has carried an `applyDefaults` arg since the initial commit, but the backend `Command::CreateWorkspace { name }` never read it — so nothing seeds defaults on demand. Full archaeology in the retro.

## Fix

- **Extract `wcore::seed_default_layout(store, tab_id)`** — the shared primitive that builds the agent/sysinfo/swarm blocks + `rootnode`/`leaforder` (previously inlined in `ensure_initial_data`, which now calls it). One definition of "the default layout."
- **Seed it on the new-window path:** `service.rs` `CreateWindow` **empty-workspace branch** calls it on the fresh tab. The tear-off branch (existing workspace) is untouched. Non-fatal: a seed failure logs and leaves a blank tab rather than failing window creation.
- Chose the targeted seed over threading `apply_defaults` through the IPC command + reducer + persist subscriber (larger/riskier surface; equivalent effect). Wiring `applyDefaults` end-to-end remains a follow-up cleanup.
- **Regression test** `seed_default_layout_creates_three_pane_layout`.

## Verification

- `cargo check -p agentmux-srv`: clean.
- `cargo test -p agentmux-srv`: **1222 passed, 0 failed, 2 ignored** (incl. the new test; first-launch refactor intact).

Retro (root cause, regression vector, lessons): `docs/retro/retro-blank-new-window-2026-06-21.md`.

### DIFF
--- a/.changesets/1782037055-fix-srv-seed-the-default-3-pane-layout-for-new-windows-so-op-51bk.md
+++ b/.changesets/1782037055-fix-srv-seed-the-default-3-pane-layout-for-new-windows-so-op-51bk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): seed the default 3-pane layout for new windows so 'Open another window' is not blank
--- a/agentmux-srv/src/backend/wcore/mod.rs
+++ b/agentmux-srv/src/backend/wcore/mod.rs
@@ -111,22 +111,39 @@ pub fn ensure_initial_data(store: &Store) -> Result<bool, StoreError> {
     // Create initial tab in workspace (pinned, matching Go's isInitialLaunch=true)
     let tab = create_tab_with_opts(store, &ws.oid, "", true)?;
 
-    // Seed the default 3-block layout:
-    //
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #1609, opened 2026-06-20
feat(pool): Windows new-window pool

a5af:
## Summary

Removes the Windows cold-path stub in `promote_pool_window_for_new_window` and wires it to the existing `promote_pool_window` function. New window (Cmd+N / File → New Window) is now pool-accelerated on all three platforms.

**The DPI problem and how it's solved:**
- `get_offset_position()` and `get_secondary_window_size()` on Windows return **physical pixels** (`GetWindowRect` / `GetMonitorInfoW`)
- `promote_pool_window`'s `to_physical()` conversion only runs when `width.is_some()` — passing `None` skips it and falls back to the `POOL_WIDTH=1200` / `POOL_HEIGHT=800` constants without double-scaling
- `pos_x/pos_y` pass as `tab_anchor_x/y` so they feed directly into `SetWindowPos` bypassing the cursor-centering `screen_x - win_w / 2` formula

**Fresh workspace via empty workspaceId:**
- `workspace_id=""` → host emits `pool:promote { workspaceId: "" }` (Windows reuses the existing promote event vs macOS/Linux `pool:new-window`)
- `awaitPoolPromote` in `pool.ts` receives it, calls `url.searchParams.set("workspaceId", "")` — empty string is falsy, so `initHostNewWindow` takes the "create fresh workspace" branch ✓

## Coverage after this PR

| Operation | Windows | macOS | Linux |
|-----------|---------|-------|-------|
| Tab tear-off | ✅ pool | ✅ pool | ✅ pool |
| New window (Cmd+N) | ✅ pool (this PR) | ✅ pool | ✅ pool |
| Pane tear-off | cold path | cold path | cold path |

## Test plan

- [ ] Windows: Cmd+N / File → New Window — second window appears instantly (~30ms), no `[create-window] task entered UI thread` cold-path log
- [ ] Windows: `muxlog host 'pool:new-window\|pool:promote'` shows promote + refill; fresh workspace (no existing tabs) opens
- [ ] Windows: back-to-back Cmd+N (pool exhausted) gracefully cold-paths
- [ ] Windows: tab tear-off still works (promote_pool_window path unchanged)
- [ ] macOS/Linux: no regression (cfg-gated, code path untouched)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1652, opened 2026-06-21
fix(pool): keep pool-promoted new windows on-screen on HiDPI (DPI positioning)

a5af:
## Problem

On HiDPI (125%) Windows, status-bar **"Open another window"** can open a **blank/invisible** window. CDP inspection of the running build showed the promoted pool window **rendered its content** (3-pane layout, 75 block-frames) but was parked **off-screen**:

| Window | screen pos | size | blocks |
|--------|-----------|------|--------|
| Window 2 / 4 | (421,246) | 960×640 | 75 ✅ |
| **Window 3** | **(-25600,-25600)** | **159×27** | 75 ❌ off-screen |

`-25600 = -32000 × 0.8` — the window stayed at the **DPI-scaled `POOL_OFFSCREEN`**; the reposition that should move it on-screen intermittently didn't take. The layout-seed fix (#1650) is unaffected — this is pure **window placement**, from the Windows new-window pool (#1609) reusing the tear-off promote path, whose DPI/physical-pixel handling is unstable and has no on-screen guarantee.

## Fix (`agentmux-cef/src/commands/window_pool.rs`, Windows-only)

1. **`clamp_rect_within`** (pure, unit-tested): shrink-to-fit + shift so a rect is always fully inside a monitor work area.
2. **Clamp before `SetWindowPos`** to the *destination* monitor's work area — a HiDPI miscalc can no longer strand the window off-screen.
3. **Re-assert the rect after `ShowWindow`** (`SWP_NOACTIVATE`): the pre-show move can be lost to the create/show + per-monitor DPI-context settle; a second idempotent move makes placement deterministic. Logs `error` if still off-screen (telemetry).

Deferred follow-ups (noted in the plan): size normalization to 70%-work-area (currently 960×640, which is fine and now guaranteed on-screen) and an explicit repaint.

## Verification

- `cargo check -p agentmux-cef`: clean.
- `cargo test -p agentmux-cef clamp_tests`: **4/4 pass** (off-screen rect, oversized, valid-unchanged, negative-monitor-origin).
- Runtime HiDPI verification via a fresh portable + CDP position sweep to follow (will confirm 0 windows land off-screen across repeated opens).

Plan with full coordinate-flow analysis: `docs/specs/PLAN_POOL_NEW_WINDOW_DPI_POSITIONING_2026_06_21.md`.

### DIFF
--- a/.changesets/1782052109-fix-pool-clamp-re-assert-pool-promoted-new-window-placement--d5cm.md
+++ b/.changesets/1782052109-fix-pool-clamp-re-assert-pool-promoted-new-window-placement--d5cm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(pool): clamp + re-assert pool-promoted new window placement so it can't land off-screen on HiDPI
--- a/agentmux-cef/src/app.rs
+++ b/agentmux-cef/src/app.rs
@@ -409,6 +409,32 @@ pub fn get_monitor_work_area(px: i32, py: i32) -> Option<(i32, i32, i32, i32)> {
     }
 }
 
+/// Like [`get_monitor_work_area`] but returns the work area in **physical**
+/// pixels (no DIP division). Win32 `SetWindowPos`/`GetWindowRect` operate in
+/// physical pixels, so clamping a physical-pixel window rect must use physical
+/// work-area bounds — using the DIP variant over-constrains placement on HiDPI
+/// (reagent P1 on PR #1652). Returns `(left, top, width, height)`.
+#[cfg(target_os = "windows")]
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
Addressed in c2588a08. **P1/P2 (DIP vs physical):** correct — `get_monitor_work_area` returns DIP (divides rcWork by the DPI scale), but the promote rect and `SetWindowPos`/`GetWindowRect` are all physical px, so clamping/telemetry against DIP over-constrained placement and mis-evaluated the off-screen check on HiDPI. Added `app::get_monitor_work_area_physical` (rcWork, no division) and pointed both the clamp and the telemetry at it. Units are now physical end to end. cargo check clean; clamp_tests still 4/4.

PR #1653, opened 2026-06-21
fix(floating-pane): redock ghost target, maximize on macOS, dock-click focus

a5af:
Fixes three issues documented in `docs/analysis/ANALYSIS_FLOATING_PANE_ISSUES_2026_06_21.md`.

## 1 — Redock ghost appears on idle floating pane instead of main window (P1, both platforms)

The cursor resolver (`ui_tasks.rs` macOS/Linux, `motion.rs` Windows) had no filter for `floating-*` window labels. When an idle floater was stacked over the main window, it would win the Z-order / bounding-rect check and receive the `floating-redock:hover-state` event — painting a ghost on the wrong window. Fix: skip any label starting with `"floating-"` in both resolvers; continue the walk so the main window behind it is found instead.

## 2 — Maximize button no-op on macOS/Linux (P2)

`toggle_floating_maximize` had the `SetWindowPos` resize block inside `#[cfg(target_os = "windows")]` only. On macOS/Linux the reducer dispatch ran (flipping `Normal ↔ Maximized` in state) but the window geometry never changed. Fix: add a `#[cfg(not(target_os = "windows"))]` branch calling `post_maximize_window(state, &label)` — the existing CEF Views `window.maximize()` / `window.restore()` task handles the toggle and native restore geometry.

## 3 — Dock icon click opens new window instead of focusing (P2, macOS)

`should_handle_reopen` in `macos_menu.rs` ignored the `hasVisibleWindows` flag and always called `open_new_window`. Standard macOS convention: Dock click with a visible window = bring it to front; only open a new window when no windows are visible. Fix: branch on `_has_visible_windows != 0` — focus via `post_focus_window("main")` when YES, `open_new_window` when NO. The launcher unix-socket forward path stays as `open_new_window` (that path is always an explicit second-process launch).

## Test plan

- [ ] Drag floating pane over main window with another idle floater partially overlapping — ghost appears on the main window, not the idle floater
- [ ] Floating pane maximize button expands window to fill monitor work area on macOS; clicking again restores
- [ ] Dock single-click with AgentMux running brings the main window to front (no new window opened)
- [ ] Dock click with all windows minimised opens a new window
- [ ] `cargo check -p agentmux-cef` clean

### DIFF
--- a/.changesets/1782056025-fix-macos-floating-pane-maximize-button-now-works-on-macos-l-ywtc.md
+++ b/.changesets/1782056025-fix-macos-floating-pane-maximize-button-now-works-on-macos-l-ywtc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): floating pane maximize button now works on macOS/Linux via CEF Views window.maximize()
--- a/.changesets/1782056025-fix-redock-floating-panes-are-never-valid-redock-targets-gho-4ckc.md
+++ b/.changesets/1782056025-fix-redock-floating-panes-are-never-valid-redock-targets-gho-4ckc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(redock): floating panes are never valid redock targets — ghost no longer appears on idle floaters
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #1654, opened 2026-06-21
fix(pool): paint pool-promoted new/tear-off windows on Windows (CEF Views show + chrome binding + taskbar name)

a5af:
## Problem

On Windows, windows served from the pre-warm window pool (used for instant **"Open another window"** and **tab/pane tear-off**) rendered **blank** despite a valid DOM. macOS was unaffected. Three distinct defects, all on the pool-promote path:

1. **Blank surface** — the Windows promote positioned the raw HWND via Win32 but never drove the CEF Views `Window.show()` that flips the browser's compositor visibility, so the renderer's frame was evicted and the window painted blank.
2. **Chrome acted on the wrong window** — promoted pool windows never landed in `state.window_hwnds`, so `resolve_window_hwnd` fell back to the main window; dragging/closing/min/max on the new window's chrome hit the original.
3. **Taskbar name** — the host exe is CEF's `bootstrap.exe`, so the taskbar jump-list / Properties showed **"CEF Bootstrap application"**.

## Fix

1. **Render (load-bearing):** cache the CEF Views `Window` at `on_window_created` in a global `Mutex` (because `browser_view.window()` is `None` for pool windows post-load **and** the Windows promote runs on the IPC thread), then **post a UI task** that runs `set_bounds()` + `show()` — the exact macOS sequence, on the UI thread. This is the genuine hidden→visible transition Chromium needs.
2. **Chrome binding:** register the promoted outer HWND under its label in `window_hwnds` at promote, so drag/close/min/max resolve to the new window.
3. **Taskbar name:** `inject-exe-icon.sh` now also stamps `FileDescription`/`ProductName` = `AgentMux` via rcedit.

The pool is fully retained (no cold-path fallback). Research + implementation notes: `docs/research/RESEARCH_CEF_PREWARM_WINDOW_BLANK_ON_WINDOWS_2026_06_21.md`, `docs/plans/PLAN_POOL_HIDDEN_PREWARM_2026_06_21.md` (§0 RESOLUTION documents what actually fixed it vs. the abandoned "hidden pre-warm" framing).

## Verified on HiDPI Windows

- ✅ "Open another window" paints correctly (was blank)
- ✅ New-window chrome: drag / close / minimize / maximize act on the new window
- ✅ Tab tear-off and pane tear-off into a new window
- ✅ Taskbar right-click shows "AgentMux"

## Known follow-up (not in this PR)

Redocking a torn-off floating pane back onto a pool-served main window leaves an **empty spot** — the block re-docks in the workspace model but the target window's frontend never instantiates the pane's child browser (no `createPane` reaches the host). Signature is independent of the pool-render work here; tracked separately in #1662.

### DIFF
--- a/.changesets/1782057571-fix-pool-show-promoted-pool-window-on-screen-first-fix-windo-vpts.md
+++ b/.changesets/1782057571-fix-pool-show-promoted-pool-window-on-screen-first-fix-windo-vpts.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-fix(pool): show promoted pool window on-screen first (fix Windows blank-on-promote)
+fix(pool): paint pool-promoted windows on Windows — drive CEF Views Window.show() on the UI thread at promote (macOS parity), register the promoted HWND for chrome ops (drag/close/min/max), and stamp FileDescription/ProductName=AgentMux on the host exe so the taskbar no longer shows "CEF Bootstrap application"
--- a/.changesets/1782057571-fix-pool-show-promoted-pool-window-on-screen-first-fix-windo-vpts.md
+++ b/.changesets/1782057571-fix-pool-show-promoted-pool-window-on-screen-first-fix-windo-vpts.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

PR #1656, opened 2026-06-21
fix(hamburger): Exit closes the current window, not main

a5af:
## Summary

- `Exit` in the hamburger menu called `invokeCommand("close_window", {})` with no window label
- The Rust handler fell back to `"main"`, so clicking Exit in any non-main window closed the **main** window instead
- Fixed to use `getApi().closeWindow()`, which reads `windowLabel` from the page's URL search params — the same path used by the close button and Cmd+W

## Test plan

- [ ] Open a second window (hamburger → New Window)
- [ ] In the new window, hamburger → Exit — the **new** window should close, original stays open
- [ ] Confirm the original window is still alive and responsive


a5af:
@codex review

PR #1661, opened 2026-06-21
fix(editor): use CSS zoom property for correct per-pane zoom scaling

a5af:
## Problem

Editor zoom in/out barely changed the visible size. Two bugs compounding:

1. **Wrong zoom mechanism** — zoom was applied via `--editor-zoom` CSS variable with `calc(Xpx * var(--editor-zoom))` on individual `font-size` rules in `editor-view.scss`. Only text scaled; padding, borders, icons, scrollbars, and layout stayed at base size. Compare: agent/swarm panes use `style={{ zoom }}` on their root element, which scales the entire subtree uniformly.

2. **`stepZoom` used terminal base font size (15px) for editor** — `getBaseFontSize()` in `zoom.{win32,darwin,linux}.ts` always returned 15 (terminal default). The editor's CodeMirror renders at 13px, so the integer-snapping guard that prevents no-visible-change steps was computing against the wrong base, requiring extra keypresses before zoom moved perceptibly.

## Fix

- `editor-view.tsx`: replace `--editor-zoom` style with `zoom: model.zoomAtom()` on `.editor-view` root
- `editor-view.scss`: remove all `calc(* var(--editor-zoom, 1))` expressions; restore flat `font-size` / `padding` values
- Both resize handlers (tree column width + preview panel height) now divide their mouse delta by `model.zoomAtom()` — `document` mousemove events are in viewport CSS pixels, but stored dimensions are in local CSS pixels; without the correction the resize handle moves at the wrong speed when zoomed
- `zoom.{win32,darwin,linux}.ts`: `getBaseFontSize()` returns 13 for editor blocks instead of 15

## Result

`Ctrl++` / `Ctrl+-` / `Ctrl+Scroll` on an editor pane now scales text, padding, borders, icons, and the file tree together, matching the visual behaviour of the agent and swarm panes.

### DIFF
--- a/.changesets/1782063152-fix-editor-apply-css-zoom-property-instead-of-calc-font-size-6x35.md
+++ b/.changesets/1782063152-fix-editor-apply-css-zoom-property-instead-of-calc-font-size-6x35.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(editor): apply CSS zoom property instead of calc font-size scaling for correct per-pane zoom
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #1664, opened 2026-06-21
fix(packaging): use APFS for intermediate DMG to avoid macOS 26 HFS+ decompressor regression

a5af:
## Summary

- Switches the intermediate UDRW disk image from `-fs HFS+` to `-fs APFS` in `scripts/package-macos.sh`
- macOS 26 Tahoe has a regression in the HFS+ transparent file decompressor (the `decmpfs` layer, indicated by the `z` attribute on files). hdiutil applies this compression automatically to all compressible files (JS, HTML, CSS) when building an HFS+ image
- When such a DMG is mounted, reading those compressed files fails with I/O error after the kernel page cache is evicted
- The CEF IPC server (`ServeDir`) reads file sizes correctly from stat() metadata, sends the right `Content-Length` header, but then silently drops the connection when the actual `read()` fails — Chromium gets `ERR_CONTENT_LENGTH_MISMATCH (-354)` → "Failed to load agentmux frontend" on every window

## Root cause trace

```
agentmux-cef ipc.rs: ServeDir::new(&frontend_dir)
 → stat("/Volumes/AgentMux N/.../frontend/index.html") → 11447 bytes ✓
 → Content-Length: 11447 sent
 → tokio::fs::File::open() + read() → I/O error (HFS+ decmpfs broken on macOS 26)
 → connection closed with 0 bytes transferred
 → CEF: ERR_CONTENT_LENGTH_MISMATCH (-354)
```

## Fix

APFS images get the same `z` attribute from hdiutil but their decompressor works correctly on macOS 26. Switching the intermediate UDRW from HFS+ to APFS avoids the regression entirely. The final ULMO-compressed DMG is unaffected (that's a container-level format, not a filesystem format).

APFS DMGs have been mountable since macOS 10.14 (Mojave); we target 11.0+, so this is safe.

## Test plan

- [ ] Build fresh DMG with `task package:macos`
- [ ] Mount and verify `cat AgentMux.app/Contents/Resources/frontend/index.html` succeeds
- [ ] Launch from DMG — frontend should load without ERR_CONTENT_LENGTH_MISMATCH
- [ ] Verify Finder window shows drag-to-install layout (APFS-backed volume still works with the osascript layout script)

### DIFF
--- a/.changesets/1782077760-fix-packaging-use-apfs-for-intermediate-dmg-to-fix-err-conte-s5bb.md
+++ b/.changesets/1782077760-fix-packaging-use-apfs-for-intermediate-dmg-to-fix-err-conte-s5bb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(packaging): use APFS for intermediate DMG to fix ERR_CONTENT_LENGTH_MISMATCH on macOS 26 Tahoe
--- a/scripts/package-macos.sh
+++ b/scripts/package-macos.sh
@@ -342,7 +342,16 @@ ln -s /Applications "$STAGE/Applications"
 # bundle as folders. Falls back to a plain DMG if Finder automation is
 # unavailable (headless / no TCC grant).
 RW="$(mktemp -u).dmg"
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1665, opened 2026-06-21
fix(agent): show session digest summary in pane header

a5af:
## Summary

`session:digest_summary` (AI-generated 10-word-or-fewer summary, stored in block meta by `useSessionDigest` via the backend `SessionDigest` RPC) was never being read by `AgentViewModel.viewText`, so it never appeared in the agent pane header.

- Show the digest summary as the header label when available
- Fall back to `term:activity` (Claude Code OSC window title) when no digest exists yet
- One-line change in `agent-model.ts` — no new signals, no new hooks; just reads an already-populated meta key

## Test plan

- [ ] Open an agent pane that has had at least one completed conversation — header should show the AI-generated summary (e.g. "Implementing auth middleware")
- [ ] Fresh agent pane with no digest yet but active Claude Code session — header falls back to `term:activity` OSC title
- [ ] Pane with neither → header shows nothing (same as before)
- [ ] `continued Xm ago` chip still appears alongside the label

### DIFF
--- a/.changesets/1782077925-fix-agent-show-session-digest-summary-in-pane-header-fall-ba-06l0.md
+++ b/.changesets/1782077925-fix-agent-show-session-digest-summary-in-pane-header-fall-ba-06l0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): show session digest summary in pane header; fall back to term:activity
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1667, opened 2026-06-21
fix(sysinfo): robust CPU chart — ZOH gap fill, debounced resize, no blank-on-reload

a5af:
## Summary

Three bugs in the sysinfo CPU chart, all rooted in fragile state management under stress:

- **Empty chart during heavy load** — streaming handler called `loadInitialData()` on any late sample, setting `loadingAtom=true` which hid the chart. Created a drop/reload feedback loop during sustained load.
- **Dock/undock distortion** — `ResizeObserver` fired every animation frame; overlapping SVGs sharing the same gradient `id` produced a line sliding across the chart.
- **Small dots / no fill after heavy CPU run** — gap-triggered reloads dropped streaming samples; NaN-filled history rendered as disconnected micro-segments by Observable Plot.

## Changes

**`sysinfo-model.ts`** — replace `addInitialData`/`addContinuousData` with a pure `sampleReducer`:
- `RESET` action: used by `loadInitialData()`, preserves NaN boundary markers for true history gaps
- `APPEND` action: zero-order hold for 1–2 missed ticks (carry-forward last value), NaN sentinel only for true breaks (>3 missed ticks / connection loss)

**`sysinfo-view.tsx`** — streaming handler now calls `model.appendData()` unconditionally; removed gap-triggered `loadInitialData()`; removed `!loading()` from `<Show>` so chart stays visible during connection-change reloads.

**`sysinfo-plot.tsx`** — 150 ms debounce on `ResizeObserver`, so dock/undock renders once at the settled size.

Spec: `docs/specs/SPEC_SYSINFO_CHART_ROBUSTNESS_2026_06_21.md`

### DIFF
--- a/.changesets/1782078641-fix-sysinfo-robust-cpu-chart-zoh-gap-fill-debounced-resize-n-tua9.md
+++ b/.changesets/1782078641-fix-sysinfo-robust-cpu-chart-zoh-gap-fill-debounced-resize-n-tua9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(sysinfo): robust CPU chart — ZOH gap fill, debounced resize, no blank-on-reload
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1658, opened 2026-06-21
feat(ui): pane minimize button + failed tool call immediate collapse

a5af:
## Summary

- **Pane minimize button** — adds a `window-minimize` icon button to every pane header between the aux buttons and the magnify/close triad. Click collapses the block to header-only height; click again restores. State persisted per block via `layout:minimized` block meta (driven reactively via WOS — no extra signal needed).
- **Aux button separator** — a subtle 1px vertical pipe (`block-frame-btn-separator`) appears between block-type-specific aux buttons and the standard triad when aux buttons are present.
- **Failed tool call immediate collapse** — `failed`, `denied`, and `canceled` tool calls now collapse immediately on completion rather than holding open until the row scrolls off screen. The user can still click to expand manually.

## Files changed

- `frontend/app/view/agent/components/ToolBlock.tsx` — add `isFailTerminal()`, short-circuit `heldOpen` in `autoExpanded()` for fail-terminal statuses
- `frontend/app/block/blockframe.tsx` — add `isMinimized`/`toggleMinimize` to `EndIcons` props; add separator + minimize button JSX; wire minimize signal from blockData meta; add `minimized` class to root div
- `frontend/app/block/block.scss` — add `.block-frame-minimize`, `.block-frame-btn-separator`, and `&.minimized` content-hide rule

## Test plan

- [ ] Click minimize on any pane → content hides, only header visible; click again → content restores
- [ ] Close and reopen the block → minimized state persists (block meta `layout:minimized`)
- [ ] Agent pane with aux buttons (brain/identity icons) → separator pipe visible between aux and triad
- [ ] Pane with no aux buttons (e.g. plain terminal) → no separator shown
- [ ] Failed Bash command → tool call collapses immediately without waiting to scroll off
- [ ] Successful Bash command → tool call still holds open as before until scroll-off
- [ ] Minimized pane: double-click header still toggles magnify; close button still closes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1668, opened 2026-06-21
fix(ui): remove stub pane minimize button that did nothing on macOS/Linux

a5af:
## Summary

- Removes the pane minimize button added in #1658
- The button called `SetMetaCommand` to set `"layout:minimized"` on the block, which applied a CSS class hiding the content area — but the tiling layout engine never reads this meta key, so the pane container keeps its full allocated size
- On macOS/Linux users saw the pane become an empty header-only rectangle that still occupied full space — indistinguishable from a no-op
- Removes: `isMinimized` signal, `toggleMinimize` callback, `EndIcons` props, `BlockFrame_Header` props, the `IconButton` render, `.minimized` SCSS rule, `.block-frame-minimize` SCSS rule

## Why not implement it properly?

Real pane collapse requires the tiling layout geometry engine (`layoutGeometry.ts`) to read block meta and constrain a pane's computed height to header-height, redistributing freed space to siblings. That's a larger feature tracked separately.

## Test plan

- [ ] Verify minimize button no longer appears in pane headers
- [ ] Verify close (×) and magnify buttons still render correctly
- [ ] Verify no TypeScript errors in blockframe.tsx

### DIFF
--- a/.changesets/1782078840-fix-ui-remove-stub-pane-minimize-button-that-did-nothing-no--thex.md
+++ b/.changesets/1782078840-fix-ui-remove-stub-pane-minimize-button-that-did-nothing-no--thex.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ui): remove stub pane minimize button that did nothing (no layout collapse on macOS/Linux)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

Issue #1662, opened 2026-06-21
Redock onto a pool-served main window leaves an empty spot (pane never instantiated)


a5af:
## Symptom

On Windows, tearing off a pane into a floating window and then **redocking** it back onto a window leaves an **empty spot** where the pane should be — the floater closes, but nothing paints in the target slot.

## Root cause (from host log analysis)

Traced on build `0.47.2+gf52adb61` (branch `agentc/pool-hidden-prewarm`):

1. `resolve_window_at_cursor` correctly targets the main window (`0x4810c10`, a pool-served main; its reverse-map label is `window-pool-…`, rewritten to `"main"` / window_id `5fe9187d`). **Target resolution is correct.**
2. `workspace.RedockFloatingPane` succeeds (~15ms) — the block moves in the workspace data model.
3. **No pane-create ever reaches the host** for the redocked block — no `createPane`, no `CreateBrowserPaneTask`, no `AlreadyLive`/`Fresh`/`Closing` arm in `browser_panes.rs`, no "browser pane created on UI thread". The target window's frontend re-docks the block into its layout but never instantiates the pane's child browser → empty spot.

So this is **not** the compositor-visibility blank-surface bug (the child browser is never created) and **not** a target-resolution bug (it resolved to the correct window). The block lands in the target workspace without the frontend mounting its pane.

## Where to look

- Frontend layout/`createPane` path after `RedockFloatingPane` resolves — why the redocked block doesn't mount a pane in the target window.
- Possible interaction with redocking specifically onto a **pool-served main** window (label `window-pool-*` rewritten to `"main"`), vs. a cold main.
- Confirm whether this reproduces pre-`agentc/pool-hidden-prewarm` (i.e. pre-existing) or is specific to the pool-served-main path.

## Repro

1. Open AgentMux (Windows) where the main window was served from the pool.
2. Tear off a pane into a floating window.
3. Drag the floater back over the main window and drop to redock.
4. Observe: floater closes, target slot is an empty spot.

Split out from #1654 (which fixes pool-window render / chrome binding / taskbar name and is verified working).

a5af:
## Deeper root cause (traced on 0.47.3 with added redock tracing)

Earlier theory ("target window never inserts a layout leaf") was **wrong** — the target window **keeps its leaf** across the whole tear-off/redock cycle (host log: `[floating-redock] leaf already present — no-op` on every redock). So no leaf needs inserting.

[...]

PR #1670, opened 2026-06-21
fix(floating-pane): full floater independence + unjam tear-off after a failed drag

a5af:
## Symptoms (reported on 0.47.3)

Tearing off panes/windows worked, then after closing a window/floater, **a second unrelated floating pane closed along with it**, and **tear-off stopped working entirely**.

## Root cause — two compounding state-management bugs (traced in the 0.47.3 host log)

**Bug 1 — window close cascaded into closing an independent floater.**
The issue #1560 cascade hook (`client/wndproc.rs::floater_cascade_wndproc`) posts `WM_CLOSE` to every floater in `floater_hwnds_for_parent(hwnd)` on a FullInstance `WM_DESTROY`. That set matched `parent == hwnd` **OR `parent == 0`**, so closing one window could destroy floaters it didn't own.

**Bug 2 — a failed tear-off permanently jammed all future tear-offs.**
When Bug 1 destroyed the block under an in-flight drag, `TearOffBlock` threw `block not found: …`. `CrossWindowDragMonitor`'s `catch` logged but **never released the host drag session**, so `active_drag` stayed `Some` forever and every later tear-off was rejected with `start_drag: drag session already active` until restart.

## Fix

**Floaters are now fully independent top-level windows.** Removed the `WM_ACTIVATE` z-order / `WM_SIZE` hide-show / `WM_DESTROY` close cascades — a window's lifecycle never touches floaters. The subclass is retained only as an observer-passthrough that self-prunes `FLOATER_CASCADE_ORIGINALS` on `WM_DESTROY` (HWND-reuse safety). `floater_hwnds_for_parent` deleted.

**Tear-off can no longer be jammed by a failed drag.** `CrossWindowDragMonitor` (`win32`/`linux`/`darwin`) now calls `cancelCrossDrag(dragId)` in its `catch`. Belt-and-suspenders: `start_cross_drag` self-heals a stale session (`active_drag` older than 30s) via a new `AppState::active_drag_snapshot()`.

## Verified on Windows

- ✅ Closing a window leaves its floaters open and independent
- ✅ Tear-off keeps working after closing windows/floaters (no permanent jam)

## Not in this PR

The floating-pane **redock empty-slot** (#1662) is a separate, deeper lifecycle race (the floater's `onNodeDelete` → `DeleteBlock` destroys the redocked block) and is tracked there.

### DIFF
--- a/.changesets/1782080390-fix-floating-pane-make-floaters-fully-independent-of-their-p-9l28.md
+++ b/.changesets/1782080390-fix-floating-pane-make-floaters-fully-independent-of-their-p-9l28.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): make floaters fully independent of their parent window and stop a failed tear-off from jamming all future tear-offs
--- a/agentmux-cef/src/client/wndproc.rs
+++ b/agentmux-cef/src/client/wndproc.rs
@@ -260,43 +260,24 @@ unsafe extern "system" fn floater_cascade_wndproc(
     lparam: isize,
 ) -> isize {
     use windows_sys::Win32::UI::WindowsAndMessaging::{
-        CallWindowProcW, DefWindowProcW, PostMessageW, SetWindowPos, ShowWindow,
-        SWP_NOACTIVATE, SWP_NOMOVE, SWP_NOSIZE, SW_HIDE, SW_SHOWNOACTIVATE,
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

a5af:
Superseded by the consolidated floating-pane DnD PR (independence + tear-off un-jam + redock/tear-off block survival + no white flash). Redock now works, so keeping it all in one PR.

PR #1671, opened 2026-06-21
feat(agent): live Haiku mini-summary in pane header

a5af:
## Summary

- Adds a `session:activity_summary` backend RPC that uses `claude-haiku-4-5-20251001` (15s timeout) to generate a short phrase from the last 30 lines of session output, writing the result directly to `term:activity` block meta.
- New `useAgentActivitySummary` SolidJS hook watches `turnPhaseAtom` — clears the label on `Submitting` (new turn starts fresh), fires the RPC on `Done` (turn complete).
- Word target is derived from pane width: `Math.max(5, Math.min(12, floor((rootWidth - 280) / 48)))` — narrow panes get ~5 words, wide panes up to 12.
- `.term-activity` CSS `max-width` widened from 300px to 500px to accommodate wider panes.

## Motivation

The agent pane header (`term:activity` key, shown in `viewText`) was always blank. The original plan was to pull this from Claude Code's OSC window title sequences, but `PersistentSubprocessController` uses piped NDJSON (not PTY), so no OSC sequences flow. The Claude Code hooks system doesn't expose equivalent data either. This replaces that dead path with a simple per-turn Haiku call.

## Test plan

- [ ] Start an agent and send a message — header label should clear while agent is working, then show a short phrase once the turn completes (~1-2s after `Done`)
- [ ] Verify narrow panes (~400px) show 5-word phrases; wider panes show more
- [ ] Verify second turn replaces the previous label (not accumulates)
- [ ] Verify header stays blank if CLI path is missing (no `cmd` meta) or if Haiku call fails


a5af:
@codex review

PR #1673, opened 2026-06-21
fix(macos): disable MacAppCodeSignClone so the packaged app launches from any volume

a5af:
## Problem

The packaged macOS host **crash-loops at startup and never opens a window** when the `.app` is launched from a read-only **DMG** volume (e.g. double-clicking the icon inside the disk image instead of dragging it to `/Applications`).

Crash report (`agentmux-cef-*.ips`):

```
EXC_BREAKPOINT (SIGTRAP, signal 5)
  → Chromium Embedded Framework  cef_initialize   ← crashes here, before any window
procPath: /Volumes/VOLUME/AgentMux.app/Contents/MacOS/agentmux-cef
os: macOS 26.5.1
```

The launcher's crash budget relaunches 3× (incl. `--disable-gpu`, which rules out graphics) then gives up and exits code 1. No host log pointer is ever written because the host dies in stage-0 init.

## Root cause

Chromium's **`MacAppCodeSignClone`** feature ("code-sign safe updates"): at launch it `clonefile()`s the whole `.app` into a boot-volume temp dir and runs from the clone, so an in-place auto-update can replace the on-disk bundle without invalidating the running process's code signature. `clonefile()` **cannot cross volumes** — from a DMG (a separate volume from the temp dir) it fails `EXDEV`, Chromium `CHECK`-aborts, and the host traps in `cef_initialize`.

"It used to work" because older builds bundled a pre-clone CEF; the bump made it fatal.

## Fix

`MacAppCodeSignClone` protects against Chromium's own bundle being swapped underneath a running process by a background updater. AgentMux sets `--disable-component-update` and ships its own launcher/updater (updates on relaunch, never live-swaps the running bundle), so the feature guards against a scenario that **structurally cannot happen here** — it's pure liability (the DMG crash + a >1 GB temp clone per run).

Append `MacAppCodeSignClone` to the existing `--disable-features` list in `agentmux-cef/src/app.rs` (`on_before_command_line_processing`). One switch, one feature name on the comma-joined value.

## ⚠️ Verify from a DMG

There's a known trap in this same function: the MachPort peer policy is read **before** `FeatureList` init, so its runtime flag can't apply in time and a source patch was needed (`docs/cef-patches/agentmux_disable_mach_rendezvous_validation.patch`). `MacAppCodeSignClone` is a genuine `base::Feature` (evaluated **after** FeatureList init, like the other working entries), so the switch *should* land in time — but the clone runs early, so **this must be verified by building and launching from a DMG**. If it proves too-late, fall back to a `code_sign_clone_manager` source patch (same pattern as MachPort).

## Test plan

- [ ] Build a signed DMG (`NOTARIZE=0 task package:macos`).
- [ ] Mount the DMG and **launch the app in-place from `/Volumes/...`** (do not drag to `/Applications`).
- [ ] App opens its main window; no `cef_initialize` SIGTRAP in `~/Library/Logs/DiagnosticReports/`.
- [ ] Sanity: also launches normally from `/Applications`.

## Notes

- Adds `docs/retro/retro-macos-codesignclone-dmg-crash-2026-06-21.md`.
- Operator workaround for already-built apps: drag `AgentMux.app` to `/Applications` and launch from there.

### DIFF
--- a/.changesets/1782083516-fix-macos-disable-macappcodesignclone-so-packaged-app-launch-zy8d.md
+++ b/.changesets/1782083516-fix-macos-disable-macappcodesignclone-so-packaged-app-launch-zy8d.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): disable MacAppCodeSignClone so packaged app launches from any volume
--- a/agentmux-cef/src/app.rs
+++ b/agentmux-cef/src/app.rs
@@ -630,6 +630,30 @@ wrap_app! {
                 //                                  → content-autofill.googleapis.com
                 // The last two were the actual Google QUIC traffic observed via
                 // net-log; AgentMux makes no such calls itself.
+                //
+                //   MacAppCodeSignClone — Chromium's "code-sign safe updates"
+                //     feature: at startup it clonefile()s the whole .app into a
+                //     temp dir on the boot volume so an in-place auto-update can
+                //     swap the on-disk bundle without invalidating the running
+                //     process's code signature. clonefile() is single-volume, so
+                //     when the bundle runs from a *read-only DMG* (a different
+                //     volume from the temp dir) the clone fails EXDEV and Chromium
+                //     CHECK-aborts inside cef_initialize → SIGTRAP, before any
+                //     window is created. AgentMux ships its own launcher/updater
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1676, opened 2026-06-22
fix(lifecycle): quit the instance on last-window close + clean Windows exit (CEF Views recycle-on-close)

AgentA-asaf:
## Summary

Closing the **last window** of an instance orphaned the entire process tree (launcher + host + srv + agent + CEF subprocesses). This PR fixes that end-to-end. Full diagnosis, churn history, and resolution: **Discussion #1680**.

> Branch is named `…stage0` and this PR opened as "the decision + tests," but it now contains the **complete fix** (see commits).

## Root cause (confirmed via instrumented smokes)

A CEF **Views** window close HIDES/recycles the window (warm pool) instead of destroying the browser — the close fires `EVENT_OBJECT_HIDE` for the **child** `Chrome_RenderWidgetHostHWND`, not the top-level frame — so **`on_before_close` never fires**. The host's quit cascade never starts, the launcher (gated on `host_child.wait()`) never returns, and the Job Object never reaps. J0 never regressed; its **trigger chain** was severed.

## The fix (triggers off the reliable OS signal, not `on_before_close`)

1. **`wrr/win_event.rs`** — on every `EVENT_OBJECT_HIDE`/`DESTROY` (NOT gated on `is_app_class` — the close's HIDE is for the child render widget), count OS-visible user windows (`Chrome_WidgetWin_` + visible + on-screen; excludes off-screen warm-pool + floaters), gated on the CEF registry having ≥1 user window. When zero visible → `quit_message_loop()` on the UI thread.
2. **`lib.rs`** — after the message loop returns, Windows uses `TerminateProcess(self, 0)` to skip the CEF teardown that fail-fasts (`0xC0000005`/`0xC0000602`) on the still-alive recycled browsers → clean code 0 → launcher reaps via Job Object (no relaunch). Extends macOS #1268.
3. **`commands/window/lifecycle.rs` + `ui_tasks.rs`** — close via the registered browser's `CefWindow` (`window.close()`), reliable for Views (NULL `window_handle`), not `WM_CLOSE` to a guessed HWND.
4. **`reducer/quit.rs`** — `count_live_user_windows` excludes pane-pool (`floating-pool-*`) windows + the level-triggered `reconcile_quit` decision/tests.

## Validation (user-driven smokes)
- Single + multi-window (3) close-all → clean code-0 exit, no orphan, no reopen.
- **Heavy floating-pane smoke on latest `main`**: multiple windows + floaters, drag/dock/redock/tear-off, then close everything → clean exit, no spurious quit, no crash, no reopen.

## Follow-ups (non-blocking, on #1680)
- Should a lone visible floater keep the instance alive (currently excluded)?
- No CI test runner exists — worth adding.


a5af:
@codex review

PR #1677, opened 2026-06-22
fix(floating-pane): DnD lifecycle — independence, tear-off un-jam, redock/tear-off block survival, darker floater

a5af:
Floating-pane DnD lifecycle fixes. **Verified working on a single window**; the new 2nd-window-tear-off-with-pool path and the last bit of floater flash are explicitly **out of scope** (frontier — see the rethink spec, below).

## Shipping (confirmed)
1. **Floaters fully independent** — closing/minimizing a window never touches its floaters (the #1560 cascade matched `parent==hwnd` OR `parent==0`, destroying unrelated floaters). Cascade actions removed; subclass kept only as a self-pruning observer.
2. **Tear-off can't be permanently jammed** — a failed drop left `active_drag` stuck `Some` forever. `CrossWindowDragMonitor` now releases it in `catch`; `start_cross_drag` self-heals a stale (>30s) session.
3. **Tear-off / redock no longer delete the moved block** — node removal during a move fired `tabcontent.tsx`'s `onNodeDelete → DeleteBlock`, destroying the just-moved block (empty-slot redock, logo-only floater). A renderer-local, time-boxed **block-move guard** suppresses the delete while a move is in flight. *(Known to be a band-aid over the real issue — see rethink spec.)*
4. **Darker floater** — dark window-class brush + opaque dark `BrowserSettings.background_color` remove one of the two tear-off white frames.

## Explicitly NOT in this PR (frontier → rethink spec)
- **2nd-window tear-off with the tear-off pool** — never validated (the pool feature blanked 2nd windows; #1654 un-blanked them; this combination is new). Tracked in `docs/specs/SPEC_FLOATING_PANE_DND_RETHINK_*`.
- **The remaining floater white flash** (show-before-paint + transparent CefSettings bg + DWM-extended glass). Needs CEF-Views floaters or a paint-gated reveal, not a hide-until-load hack (which destabilized tear-off and was reverted here).

Supersedes #1670.


a5af:
@codex review

PR #1674, opened 2026-06-21
feat(splash): user@host + version footer + splash:disabled setting

AgentU-asaf:
Implements `SPEC_SPLASH_USERINFO_AND_DISABLE_2026_06_21.md` (in this PR).

## What's in
- **Identity footer near the bottom of the splash** — `user@host` and `v<version>` (+ a `(dev)` label on non-stable builds). Muted, centered, middle-ellipsized so a long hostname never widens the card.
- **Disable toggle** — `"splash:disabled": true` in `settings.json` (read early by the launcher, before srv/host), plus `AGENTMUX_SPLASH=0` env override. **Fail-safe:** any settings read error → splash stays enabled.

## Cross-platform status
| Piece | Linux | macOS | Windows |
|---|---|---|---|
| Disable toggle | ✅ guard | ✅ guard | ✅ guard (`spawn_splash`) |
| Footer rendering | ✅ **implemented + demoed** | ⏳ next slice | ⏳ next slice |

The footer's hard part is rendering **dynamic text** on the software-buffer backends (the splash previously only blitted a static brain image). This PR lands the **shared scaffolding** — `SplashInfo` gatherer, a baked DejaVu Sans Mono glyph atlas (`splash_font.rs`, regenerated by `scripts/dev/gen-splash-font.py`) and a software text blitter (`splash_text.rs`) — plus the **Linux** footer. **macOS** (`NSTextField`) and **Windows** (atlas into the DIB + square→W×H card) footer rendering are the next slices (spec §5/§10); they consume the same `SplashInfo`. Those need their own machines to build/verify, so they're intentionally not in this Linux-built PR.

## Verification (Linux)
- `--splash-selftest` (new launcher flag) renders the splash in isolation. Confirmed on GNOME/Mutter Wayland: card **312×381 centered at +521+1064** (exact), footer shows `snowbark@snow...ual-Platform` / `v0.47.2 (dev)`, brain still centered + pulsing above.
- `cargo build -p agentmux-launcher` green; 4 unit tests for the `splash:disabled` settings read pass (true / false / absent / fail-safe).

## Notes
- The disable setting `splash:disabled` should be registered in the settings schema/defaults (`agentmux-srv/.../wconfig`) so a future settings UI can surface a toggle — flagged as follow-up in the spec.

PR #1678, opened 2026-06-22
feat(splash): identity footer on macOS + Windows (completes cross-platform footer)

AgentU-asaf:
Completes the cross-platform splash footer started in #1674 (Linux). macOS and Windows now show the same `user@host` / `v<version>` (+ dev label) footer near the bottom, consuming the shared `SplashInfo`.

## Changes
- **Windows** (`splash.rs`) — software DIB path: split the square `SPLASH_SIZE` into `SPLASH_W`×`SPLASH_H`, re-center the card, and draw the footer into the premultiplied DIB via the shared `splash_text` blitter. `window_alpha = 1.0` because the layered window fades uniformly (`fade_out`/`push_layered`), so the footer needs no per-pixel fade.
- **macOS** (`splash_mac.rs`) — native AppKit: grow the card by a footer band, keep the brain in the upper region (Y-up coords), and add two `NSTextField` labels via `labelWithString:` (borderless/transparent) in muted gray. They fade with the window on dismiss and don't pulse.

## Verification status ⚠️
- **Linux build green** (the new code is `cfg`-gated to its platform, so Linux is unaffected — confirmed).
- **macOS / Windows: NOT built here** — this dev box has no macOS/Windows/cross toolchain (no `apple-darwin`/`windows-gnu` target, no mingw; rusqlite's bundled C build blocks a cross `cargo check`). The code is type-matched to the **existing** FFI/GDI patterns in each file (same `objc_msgSend` transmute helpers; same DIB/`UpdateLayeredWindow` model), but it needs an **on-machine / CI build** to confirm it compiles and renders. Flagging explicitly rather than claiming verification I can't do.

The shared pieces (`SplashInfo`, the baked atlas, `splash_text`) are unchanged and already verified on Linux.

### DIFF
--- a/.changesets/1782114529-feat-splash-render-the-user-host-version-footer-on-macos-and-lct0.md
+++ b/.changesets/1782114529-feat-splash-render-the-user-host-version-footer-on-macos-and-lct0.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(splash): render the user@host + version footer on macOS and Windows too (completes the cross-platform footer)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

Issue #1681, opened 2026-06-22
Floating-pane DnD lifecycle rethink (move-vs-close intent, CEF-Views floaters, pool identity)


a5af:
Tracking issue for the floating-pane DnD work intentionally deferred out of PR #1677. Full design: `docs/specs/SPEC_FLOATING_PANE_DND_RETHINK_2026_06_22.md`.

PR #1677 ships the **single-window-stable** subset (independence, tear-off un-jam, move-doesn't-delete-block via a **band-aid guard**, darker floater). The items below are the durable fixes + the two open symptoms.

## Open symptoms
- [ ] **2nd-window tear-off with the tear-off pool** — fails (`TearOffBlock: block not found`; block also doesn't persist torn off). New path: the pool blanked 2nd windows, #1654 un-blanked them, this combination was never validated.
- [ ] **Remaining floater white flash** on tear-off (one of two frames; the deferred-show fix for it was reverted in #1677 because it destabilized tear-off).

## Root causes → durable fixes (sequenced)
- [ ] **R1 — `onNodeDelete` conflates move vs close.** Node removal must carry `reason: "close" | "move"`; `DeleteBlock` only on `"close"`. **Then delete the block-move guard.** Highest leverage, smallest change — dissolves redock/tear-off block-survival at the root.
- [ ] **R3 — pool-promoted secondary windows keep `window-pool-*` identity.** Re-key to a real window identity (label + `window_hwnds` + `backend_window_id`) at promote. Fixes 2nd-window tear-off + the redock/focus resolution bugs.
- [ ] **R2 — floater is a raw Win32 popup w/ the browser-pane handler.** Make floaters CEF-Views windows like the main window → kills the flash at the source, retires `create_popup` / floater wndproc / HWND registry.

Suggested order: R1 → R3 → R2. R1+R3 should make 2nd-window-tear-off-with-pool work; R2 is the polish/cleanup tier.

Related: #1662 (redock empty-slot, fixed via guard in #1677), #1677.

a5af:
**R1 implemented → PR #1685** (branch `agentc/r1-node-delete-intent`).

[...]

PR #1685, opened 2026-06-22
fix(layout): move-vs-close intent — a block move no longer deletes the moved block (R1)

a5af:
**R1 of the floating-pane DnD rethink (#1681).** Replaces the band-aid block-move guard (shipped in #1677) with a root-cause fix.

## Root cause
Every backend `"delete"` layout action is a **MOVE** — `tear_off_block` (`dnd.rs`) and the `queue_source_layout_delete` sites (`TearOffBlock` / `RedockFloatingPane` / `PromoteBlockToTab`). Their documented intent is *"remove the node from the source layout tree"* — the block lives on in its new tab. But the frontend applied them via `handleBackendAction → model.closeNode → onNodeDelete → ObjectService.DeleteBlock`, **destroying the just-moved block**. That single move/close conflation caused: empty-slot redock, logo-only floater on tear-off, block snapping back, and `TearOffBlock: block not found`.

## Fix
`handleBackendAction`'s `DeleteNode` case now removes the node directly via `treeReducer(DeleteNode)` (no block delete) — matching the action's intent and the orphaned-block branch right beside it. A genuine pane **close** still deletes the block through the frontend `closeNode` path (no backend `"delete"` action is ever emitted for a close), so close is unaffected.

With move/close separated at the source, the renderer-local, time-boxed **block-move guard is obsolete and deleted entirely**: `block-move-guard.ts` + arming sites in all three `CrossWindowDragMonitor` platforms + `floating-pane-workspace` + the check in `tabcontent`.

## Verified (single window)
- ✅ Tear-off persists; redock renders in place, block survives (repeatable)
- ✅ Close-a-pane still deletes the block
- ⬜ 2nd-window tear-off still fails — **expected**, that's R3 (pool-window identity), tracked in #1681

## Not in this PR
R3 (pool-window identity → fixes 2nd-window tear-off) and R2 (CEF-Views floaters → kills the floater flash) — see #1681.


a5af:
@codex review

PR #1688, opened 2026-06-22
fix(splash): center the macOS splash footer (unified NSTextAlignment value)

a5af:
## Problem

The new splash footer (`user@host` / `v<version>`, #1674/#1678) renders **right-aligned and clipped against the right edge** of the card on macOS instead of centered.

## Root cause

The footer `NSTextField` labels were set with `setAlignment: 2` and a comment claiming `NSTextAlignmentCenter`. But on **macOS 10.12+** the `NSTextAlignment` enum adopted UIKit's *unified* values:

| value | meaning |
|-------|---------|
| 0 | Left |
| **1** | **Center** |
| **2** | **Right** |
| 3 | Justified |
| 4 | Natural |

So `2` is `NSTextAlignmentRight`. (Value `2` for center is the *pre-Sierra* macOS encoding — the author used the old constant.) Fix: pass `1`.

## Verification

Rendered the splash offscreen before/after (no Screen Recording permission needed):

- **Before:** footer flush against the right edge.
- **After:** footer centered with equal margins, both lines centered.

This PR also adds `Splash::dump_png()`, wired into the existing `--splash-selftest` via `AGENTMUX_SPLASH_DUMP_PNG=<path>`, which renders the splash to a PNG with `cacheDisplayInRect:`. That makes the selftest actually fulfill its documented purpose ("eyeballing the footer + centering") on machines/CI without Screen Recording access.

Launcher-only change; no host/srv impact.

### DIFF
--- a/.changesets/1782133923-fix-splash-center-the-macos-splash-footer-unified-nstextalig-w112.md
+++ b/.changesets/1782133923-fix-splash-center-the-macos-splash-footer-unified-nstextalig-w112.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(splash): center the macOS splash footer (unified NSTextAlignment value)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1690, opened 2026-06-22
fix(#1681): secondary-window DnD — 2nd-window tear-off + redock into any window

a5af:
Two fixes for the long-running #1681 "secondary-window drag-and-drop is broken" theme. Both stem from the same root: **pool-promoted secondary windows had no stable identity downstream** — one in the reducer (blocks), one in window-label resolution (redock).

---

## Fix 1 — 2nd-window tear-off (`block not found`)

**Problem:** tearing a pane/tab off a secondary window silently failed. The TearOffBlock saga had been rejecting at **0.05ms with no log line**:
```
block_not_found  source_ws_exists:true  source_tab_exists:true
block_exists:false  known_workspaces:4  known_tabs:4  known_blocks:3
```
Workspace and tab exist in the reducer; the **block does not**, and `known_blocks` is pinned at the main window's bootstrap count.

**Root cause:** "Open another window" seeded its 3 default panes **straight into SQLite** (`seed_default_layout` → `create_block`), bypassing the reducer. `CreateWindow` runs *after* bootstrap, and `srv_state` (which TearOffBlock checks) is only hydrated from SQLite once at startup. So new-window blocks lived only in SQLite — rendered fine, invisible to the reducer.

**Fix:** `CreateWindow`'s fresh path now creates the 3 seed blocks **through the reducer** (`CreateBlock`) → both `srv_state` and SQLite. Split `seed_default_layout` into block-creation + shared `write_default_three_pane_layout`. Kept a loud ERROR diagnostic in the tear-off saga (a silent reject is how this hid). **Regression test** drives a real `CreateWindow` → asserts blocks in `srv_state` → tear-off succeeds end-to-end.

**Verified** by user on portable build: 2nd-window tear-off works.

## Fix 2 — redock into secondary windows (no landing ghost)

**Problem:** floating-pane redock ghosts only appeared on the home window; dragging a floater over a 2nd/3rd window showed no drop preview → couldn't dock there.

**Root cause:** `resolve_window_at_cursor` rewrote a tracked `window-pool-*` label to `"main"` whenever `find_main_window()` pointed at that HWND. But `find_main_window()` returns the topmost non-floater window — with several promoted windows open it picks a SECONDARY one, which then resolved to `"main"`. The ghost is gated on `target_label === windowLabel` (app-init.ts), and the secondary's frontend `windowLabel` is `window-pool-*`, so it never matched; `backend_window_id("main")` also mis-targeted the drop. Confirmed via the `redock-resolve` host diagnostic:
```
cursor inside window  hwnd=0x60a26  label=window-pool-99ff…  is_main_hwnd=true
```

**Fix:** the `window_hwnds` label is always the exact string the target window's renderer carries as `?windowLabel=`, so resolve now returns the tracked label **verbatim** (removed the pool→"main" rewrite for tracked HWNDs). Kept the `None if is_main_frame => "main"` fallback for genuine cold-main-before-cache. Robust even when `find_main_window` picks the wrong frame.

---

## Known same-root-cause follow-up

`agent.open` / `pane.open` (`app_api.rs:362, 926`) also create blocks store-only post-bootstrap, so tearing off a *freshly-opened* pane likely fails identically. Scoped out; tracked for a focused follow-up.

### DIFF
--- a/.changesets/1782134938-fix-layout-seed-new-window-default-blocks-through-the-reduce-bbiy.md
+++ b/.changesets/1782134938-fix-layout-seed-new-window-default-blocks-through-the-reduce-bbiy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): seed new-window default blocks through the reducer so secondary-window tear-off works (R3, #1681)
--- a/.changesets/1782135834-fix-dnd-redock-floating-panes-into-secondary-windows-r3-wind-7ft0.md
+++ b/.changesets/1782135834-fix-dnd-redock-floating-panes-into-secondary-windows-r3-wind-7ft0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dnd): redock floating panes into secondary windows (R3 window-label resolve, #1681)
--- a/agentmux-cef/src/commands/window/motion.rs
+++ b/agentmux-cef/src/commands/window/motion.rs
@@ -377,28 +377,38 @@ pub fn resolve_window_at_cursor(
                                 main_hwnd = %format!("{:#x}", main_hwnd),
                                 "cursor inside window"
                             );
-                            // Resolution precedence (P1). The HWND→label
-                            // reverse map can carry a STALE `window-pool-*`
-                            // label for a promoted main frame: a warm-pool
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

PR #1691, opened 2026-06-22
fix(dnd): tear-off/redock a freshly-opened pane — route docked block creation through the reducer (#1681)

a5af:
Follow-up to #1690. Same root cause, different call sites.

## Problem

Tearing off (or redocking) a **freshly-opened pane** — one created via the widget bar (`agent.open`, or the docked branch of `pane.open`: terminal/editor/browser/sysinfo/help) — was rejected `block not found`, even in the main window. Only the *original* seeded panes could be torn off.

## Root cause

`agent.open` (`app_api.rs:362`) and docked `pane.open` (`app_api.rs:926`) created their block with the store-only `wcore::create_block`, bypassing the reducer. These RPCs run **after bootstrap**, and the in-memory reducer `srv_state` (which the TearOffBlock / RedockFloatingPane saga pre-conditions check) is only hydrated from SQLite at startup. So the block existed in SQLite (the frontend rendered it) but never in `state.blocks` → tear-off/redock rejected it.

This is the exact mechanism #1690 fixed for the new-window *seed* blocks; these are the sibling call sites flagged there.

## Fix

Create the block via `dispatch_to_reducer(CreateBlock)` + apply the `BlockCreated` event to the wstore — mirroring the already-correct `open_pane_floating` path (which had a comment explicitly calling out this requirement). `register_agent_open` now captures the `Clone` `AppState` so its handler closure can reach the reducer. `BlockCreated` carries meta → `apply_block_created` writes the wstore Block, so `agent.open`'s controller resync still reloads it by id.

## Test

`docked_pane_open_block_is_in_reducer_and_tears_off`: opens a docked pane, asserts the block is in `srv_state`, then tears it off and asserts success. Would fail on old code; full app_api + tear-off suite green (17 passed).

> Note: stacks conceptually on #1690 but is independent at the file level (different functions); both target #1681. Recommend merging #1690 first.

### DIFF
--- a/.changesets/1782137046-fix-dnd-create-docked-agent-pane-blocks-through-the-reducer--mchi.md
+++ b/.changesets/1782137046-fix-dnd-create-docked-agent-pane-blocks-through-the-reducer--mchi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dnd): create docked agent/pane blocks through the reducer so freshly-opened panes can tear off/redock (#1681)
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1692, opened 2026-06-22
feat(macos): open a new window on app reopen (fix "not responding")

a5af:
## Problem

Double-clicking `AgentMux.app` in Finder (or `open AgentMux.app` without `-n`) while AgentMux is already running shows:

> You cannot open the application "AgentMux" because it is not responding.

…instead of opening another window.

## Root cause

`CFBundleExecutable = agentmux-launcher`, so LaunchServices delivers the Finder/`open` reopen Apple Event to the **launcher** process. The existing reopen handler (#1572) lives in the **host** (`agentmux-cef`), which only ever receives the **Dock-tile** reopen (it owns the tile). The launcher had **no** reopen handler → the event went unhandled → LaunchServices timed out → "not responding".

**Proof:** the running 0.47.3 host binary already contains the #1572 handler (`reopen-hook` symbols), yet a Finder double-click still times out — so the event never reaches the host. `lsappinfo` confirms two ASNs for one bundle: launcher = `UIElement` rooted at the Dock; host = `Foreground` "in front".

## Fix

Install `applicationShouldHandleReopen:hasVisibleWindows:` on the launcher's splash `NSApplication` (same Chromium-proof delegate technique the host uses). On reopen it **always** forwards `open_new_window` over the single-instance socket this launcher bound — reusing the exact `(data_dir, dir_hash)` published by `run_unix` via a `OnceLock`, **never recomputed**, so a leaked `AGENTMUX_CHANNEL` / `AGENTMUX_IPC_VERSION_OVERRIDE` can't misroute it to the wrong `(channel, version)` instance.

The host's Dock-click handler is unchanged (native focus-on-visible); its log lines now carry `proc=host` to disambiguate from `proc=launcher`.

## Spec

`docs/specs/SPEC_MACOS_REOPEN_NEW_WINDOW_2026_06_22.md` — both open questions resolved with evidence (lsappinfo ASN topology + binary-symbol presence).

## Verification

- ✅ Compiles (`cargo build -p agentmux-launcher`).
- ✅ `install_reopen_handler()` runs at runtime without crashing (logs `reopen-hook: installed dedicated launcher reopen delegate`).
- ⏳ **Runtime delivery gate (§6.5):** confirm AppKit calls the launcher delegate while it parks in `CFRunLoopRunInMode` — verified on a fresh build by double-clicking the running app and watching `muxlog launcher` for `reopen-hook:fired proc=launcher` + a second window. `finishLaunching` (which the launcher calls) installs the standard reopen AE handler and default-mode CFRunLoop dispatches it, so delivery is expected; fallback is `[NSApp run]` if not.

Launcher + host (log-string only) change; no srv impact. Windows/Linux unaffected (§10).

### DIFF
--- a/.changesets/1782137474-feat-macos-open-a-new-window-on-app-reopen-fix-not-respondin-s6i1.md
+++ b/.changesets/1782137474-feat-macos-open-a-new-window-on-app-reopen-fix-not-respondin-s6i1.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(macos): open a new window on app reopen (fix 'not responding' on Finder double-click)
--- a/agentmux-cef/src/macos_menu.rs
+++ b/agentmux-cef/src/macos_menu.rs
@@ -158,12 +158,12 @@ unsafe extern "C" fn should_handle_reopen(
                 // post_focus_window is non-blocking and safe to call from the
                 // AppKit main thread.
                 crate::ui_tasks::post_focus_window(state, "main");
-                tracing::info!("reopen-hook:fired — focused main window (hasVisibleWindows=YES)");
+                tracing::info!("reopen-hook:fired proc=host — focused main window (hasVisibleWindows=YES)");
             } else {
                 // No visible windows — open a new one.
                 match crate::commands::window::open_new_window(state) {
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1693, opened 2026-06-22
chore(portable): tidy package layout — fix README data location, drop vestigial data/, marker into runtime/

a5af:
Tidies the portable package's extract layout. Three related changes; no app behavior change beyond packaging.

## 1. README data location was wrong
Both portable READMEs claimed user data lives in the local `data/` folder next to `agentmux.exe`. It doesn't: portable builds run in Portable mode, whose state lives in `~/.agentmux/versions/<version>/` (`RuntimeMode` maps `Installed | Portable -> "versions"`; *"portable binaries are stateless on disk"*). Top-level README now points there and lists the real subfolders; the `data/README.txt` is removed with the folder (below).

## 2. Dropped the vestigial `data/` folder
The bundled `data/` folder was empty-but-for-its-README and **nothing ever wrote to it** — the last artifact of the pre-`SPEC_DATA_DIR_UNIFICATION` "portable = local data" design. It actively misled users into backing up an empty dir. Removed from packaging.

## 3. Moved the marker into `runtime/`
`agentmux-portable.marker` now lives in `runtime/` so the extract root is just `agentmux.exe` + `README.txt` + `runtime/`. The marker still earns its keep (it carries the ephemeral build label the UI shows via `read_build_label`, and flags portable mode). `is_portable_marker_present` now also checks one level down in `runtime/` (the launcher sits at the root; host/srv run from `runtime/` and find it next to themselves). **Backward compatible** — a root-level marker is still accepted. Detection test + stale comments (`runtime_mode` module doc, `sidecar.rs`, `platform.rs`) updated.

## Note surfaced during this work
Portable vs Installed mode is otherwise a **no-op**: both resolve data to `~/.agentmux/versions/<v>/`, and the only other `Portable`-specific code sets a diagnostic-only log flag. The mode distinction is effectively vestigial; the marker's sole live purpose is the UI build label. Flagging in case we want to retire the split later (out of scope here).

### DIFF
--- a/.changesets/1782139565-docs-portable-correct-readme-data-location-agentmux-versions-ffw5.md
+++ b/.changesets/1782139565-docs-portable-correct-readme-data-location-agentmux-versions-ffw5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(portable): correct README data location (~/.agentmux/versions, not local data/ folder)
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -171,14 +171,16 @@ fn is_sensitive_env_key(key: &str) -> bool {
 /// rebuild). Released / installed / dev builds have no marker → returns
 /// `None` and the UI falls back to the plain version + git hash.
 ///
-/// The host runs from `<portable>/runtime/`, so the marker sits one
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1694, opened 2026-06-22
feat(agent-pane): marching-ants busy indicator

a5af:
## Summary

- Replaces the single-glint gradient sweep on the agent pane progress bar with a marching-ants diagonal stripe animation
- Bar height bumped 2→3px so the stripe pattern reads clearly at the top of the pane
- Animation still uses `transform: translateX()` (GPU compositor) — no regression on the `background-position` promotion issue noted in the existing comments
- `repeating-linear-gradient(-45deg, accent 0→4px, accent@25% 4→8px)` + `translateX(11.314px)` per 0.5s cycle = seamless loop (one stripe period = 8px × √2)
- Fallback `background: var(--accent-color)` retained for Chromium < 111 (color-mix unsupported)
- `prefers-reduced-motion` still disables animation

## Test plan
- [ ] Build with `task dev` and trigger an agent run — progress bar should show animated diagonal stripes marching left-to-right
- [ ] Stripes should be visible and not flicker; motion should be smooth and GPU-composited
- [ ] Bar fades out when agent finishes (`opacity: 0` transition)
- [ ] `--stopping` state applies 0.55 opacity dimming as before

### DIFF
--- a/.changesets/1782140264-feat-agent-pane-replace-gradient-sweep-with-marching-ants-pr-wj9z.md
+++ b/.changesets/1782140264-feat-agent-pane-replace-gradient-sweep-with-marching-ants-pr-wj9z.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): replace gradient sweep with marching-ants progress bar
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1695, opened 2026-06-22
fix(dnd): redock into the first window when it's a promoted pool window (#1681)

a5af:
Follow-up to the redock fix in #1690. That PR made `resolve_window_at_cursor` return the tracked `window_hwnds` label verbatim — which fixed redock into promoted **secondary** windows but **regressed the first window**: no landing ghost, so a floater wouldn't dock into the primary window.

## Root cause (from the `redock-resolve` + `capture_hwnd_for_label` host logs)

The primary window can be served by a **promoted pool window**. The host knows it everywhere as `window-pool-<id>` (browser registry + `window_hwnds`), but that window's renderer **re-identifies as `"main"`** — it calls `register_backend_window("main")`, not the pool label. So:

- `window_hwnds` maps the HWND only to the pool label; `"main"` never gets a `window_hwnds` entry (`capture_hwnd_for_label` can't claim the already-pool-mapped HWND → log: `no available HWND for label=main`).
- The redock ghost gates on `target_label === the renderer's windowLabel` (`"main"`), but resolve returned the pool label → never matched → no ghost.

## The distinguisher

A genuine promoted **secondary** keeps its pool label as its `windowLabel` and **registers** it (`backend_window_id` is `Some`). The **primary-served-by-pool** registers `"main"` instead, leaving its pool label **unregistered**. resolve now uses that authoritative signal:

```
pool label WITH a backend_window_id        -> verbatim   (genuine secondary)
pool label WITHOUT one, on the main frame  -> "main"      (primary via pool)
```

This fixes the first window **without re-breaking secondaries**, and is robust to `find_main_window` picking the wrong frame (a registered pool label is never rewritten).

## Verification
Building for manual confirmation: redock into the first window AND into secondary windows. (The earlier secondary-window fix from #1690 stays covered by the `Some(l) if registered` branch.)

### DIFF
--- a/.changesets/1782141143-fix-dnd-redock-into-the-first-window-when-it-is-served-by-a--2653.md
+++ b/.changesets/1782141143-fix-dnd-redock-into-the-first-window-when-it-is-served-by-a--2653.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dnd): redock into the first window when it is served by a promoted pool window (#1681)
--- a/agentmux-cef/src/commands/window/motion.rs
+++ b/agentmux-cef/src/commands/window/motion.rs
@@ -377,38 +377,54 @@ pub fn resolve_window_at_cursor(
                                 main_hwnd = %format!("{:#x}", main_hwnd),
                                 "cursor inside window"
                             );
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1697, opened 2026-06-22
fix(packaging): purge inactive pages before ULMO DMG compression

a5af:
LZMA (`hdiutil convert -format ULMO`) peaks at 4-6GB RAM. After a full Rust compile the system holds significant inactive pages and macOS OOM-kills the compression step (SIGKILL, exit 137), silently falling through to a corrupt/missing DMG.

`sudo purge` before the conversion releases inactive pages so LZMA has headroom. `|| true` means CI/sandboxed builds that can't run purge continue without error.

Confirmed: previous build OOM-killed at ULMO step → workaround of converting an already-built UDZO DMG to ULMO post-compile succeeded → this adds purge so the inline step works too.


a5af:
@codex review

PR #1698, opened 2026-06-22
fix(statusbar): wrap long tooltips instead of overflowing the balloon

a5af:
## Problem
The GFX status-bar item's tooltip overflows its balloon: the text spills outside the bordered box.

## Root cause
The shared `[data-tip]` balloon rule (`StatusBar.scss`) set both:
```scss
white-space: nowrap;
max-width:   320px;
```
That's contradictory — `nowrap` forces a single line, so `max-width` can't constrain it. A long tip (the GFX item appends the full GPU renderer string, e.g. `… — ANGLE (NVIDIA GeForce RTX …) · terminal: WEBGL`) lays out wider than 320px on one line and overflows the balloon.

## Fix
Let the balloon wrap within its max-width:
```scss
white-space:   normal;
width:         max-content;   // short tips stay one line; long ones wrap at max-width
max-width:     320px;
overflow-wrap: anywhere;      // break unbreakable renderer strings instead of overflowing
```
One shared-rule change → fixes GFX, Backend, Host, Config, and every other status-bar tooltip. No markup changes, no regression for short tips (they stay single-line via `max-content`).

### DIFF
--- a/.changesets/1782144490-fix-statusbar-wrap-long-data-tip-tooltips-instead-of-overflo-hwjh.md
+++ b/.changesets/1782144490-fix-statusbar-wrap-long-data-tip-tooltips-instead-of-overflo-hwjh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(statusbar): wrap long data-tip tooltips instead of overflowing the balloon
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1699, opened 2026-06-22
perf(layout): reduce native-pane reflow settle window 32ms -> 4ms

a5af:
Reduces `REFLOW_WINDOW_MS` (`frontend/app/platform/pane-anim.ts`) from 32ms to 4ms, as requested.

## What it is
After a pane geometry change, `notifyPaneReflow()` opens a settle window during which `browser-view.tsx`'s `requestAnimationFrame` loop re-samples the placeholder rect and re-positions the native pane HWND every frame. `REFLOW_WINDOW_MS` is how long that loop keeps running after the change. The CSS reflow animation it once tracked was removed (the rect is final synchronously), so the window now only covers scheduler slack: 220ms → 32ms → **4ms**.

## Honest caveats (called out in the commit + code comment)
- The resample loop is **rAF-gated**, so 4ms floors at ~1 frame in practice; the real effect is a shorter post-change settle *tail*, not a higher repaint rate.
- This same constant governs the settle after **pane open/close/split**, not just window resize. The prior ~2-frame (32ms) margin was there to catch Solid's async reactive flush / ResizeObserver tick — so those paths deserve a smoke for panes landing on a stale rect. `~16ms` (one full frame) is the safer fallback if 4ms regresses anything.

### DIFF
--- a/.changesets/1782144871-perf-layout-reduce-native-pane-reflow-settle-window-from-32m-yn2c.md
+++ b/.changesets/1782144871-perf-layout-reduce-native-pane-reflow-settle-window-from-32m-yn2c.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+perf(layout): reduce native-pane reflow settle window from 32ms to 4ms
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1700, opened 2026-06-22
feat(agent): persist per-agent zoom across pane close/reopen

a5af:
## Problem
Zooming an agent pane (Ctrl+Wheel) is lost on close. Reopening the same agent comes back at default 1.0. Zoom is a per-agent reading preference but was bound to the throwaway block.

## Root cause
Agent zoom lives in the block's `term:zoom` meta (`agent-view.tsx:917` reads it, `term.tsx:222` writes it). Closing the pane deletes the block; `agent.open` builds a fresh block seeding `agentId`/provider/etc. but **not** `term:zoom`, so it defaults to 1.0. Nothing tied zoom to the agent's stable id.

## Design (full spec: `docs/specs/SPEC_AGENT_ZOOM_PERSISTENCE_2026_06_22.md`)
Persist zoom per `agent.id` in the existing per-agent content store (`agent_content`, new `content_type = "ui:zoom"`) — the same vehicle as `env`/`instructions`, already **global cross-channel**, so zoom follows the agent everywhere. No schema change, additive.

- **Restore** — `agent.open` seeds the new block's `term:zoom` from `ui:zoom` (`parse_seed_zoom`: parseable, non-default, clamped to the frontend's `[0.5, 2.0]`). Live render path unchanged.
- **Persist** — the `SetMeta` handler mirrors an agent block's `term:zoom` change into `ui:zoom`, **debounced ~300ms** via a per-`agent_id` generation counter, so a Ctrl+Wheel burst collapses to one durable write (+ one global re-mirror). `term:zoom: null` (the frontend's reset-to-1.0 convention) **deletes** the row, so default agents store nothing. Only blocks carrying `agentId` participate.
- **No frontend change** for the core.

## Tests
- `parse_seed_zoom`: valid/non-default → `Some`; default/out-of-range/garbage → `None`.
- mirror debounce: a 3-update burst persists only the final value, and nothing before the window elapses.
- reset-to-default (`None`) deletes the saved row.

(4 new tests, all green; full srv suite unaffected.)

## Out of scope (noted in spec §8)
Live cross-pane sync of an in-flight zoom across two simultaneously-open panes of the same agent; terminal-pane zoom (no stable cross-open identity); a per-pane "transient override" mode.

### DIFF
--- a/.changesets/1782145929-feat-agent-persist-per-agent-zoom-across-pane-close-reopen-gsqo.md
+++ b/.changesets/1782145929-feat-agent-persist-per-agent-zoom-across-pane-close-reopen-gsqo.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): persist per-agent zoom across pane close/reopen
--- a/agentmux-srv/src/backend/storage/content.rs
+++ b/agentmux-srv/src/backend/storage/content.rs
@@ -170,17 +170,25 @@ impl Store {
     }
 
     /// Delete a specific content blob. Returns true if a row was deleted.
-    #[allow(dead_code)]
     pub fn agent_content_delete(
         &self,
         agent_id: &str,
         content_type: &str,
     ) -> Result<bool, StoreError> {
-        let conn = self.conn.lock().unwrap();
-        let rows = conn.execute(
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #733, opened 2026-05-07
fix(InstancePanel): use Reconcile (replace) instead of Seed (additive) on refresh

AgentA-asaf:
## Summary

Codex P3 follow-up to PR #732. The InstancePanel-on-open refresh added in #732 used `ApplySeed`, which is **intentionally additive** (codex P1 #603: protects against typed events racing the boot snapshot fetch). That meant in dev mode where the launcher never pushes WindowClosed events, closed windows persisted as ghost rows in the panel even after re-opening.

## Fix

Add a new reducer command `ReconcileFromSnapshot` that **replaces** `knownEntries` wholesale:
- Adds labels in the snapshot but not in `knownEntries`
- **Removes** labels in `knownEntries` but not in the snapshot
- Filters non-instance labels (same `isInstanceLabel` rule as `ApplySeed`)
- No tombstone handling (post-seed; pre-seed tombstones are no longer relevant)

Distinct from `ApplySeed`:
- `ApplySeed` — boot-time, additive, race-protected
- `ReconcileFromSnapshot` — steady-state, replace, used for periodic refresh

`InstancePanel.tsx::onMount` now calls `reconcileKnownEntriesFromSnapshot` instead of `seedKnownEntriesFromSnapshot`.

## Test plan

- [x] 3 new regression tests in `reducer.test.ts`:
  - `removes labels absent from snapshot (closed windows actually disappear)`
  - `adds new labels and removes missing ones in one pass`
  - `filters non-instance labels (parity with ApplySeed)`
- [x] All 29 reducer tests pass
- [ ] Smoke in dev: open InstancePanel → close a window → re-open InstancePanel; the closed window's row should be gone

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1701, opened 2026-06-22
fix(window): keep the status-bar window count accurate (close report + gap resync)

AgentA-asaf:
## Summary
Fixes the status-bar window count `(N)` (next to the version) going stale/inconsistent — a sibling of the #1676 lifecycle bug (Discussion #1680), same root cause (CEF Views recycle-on-close), two distinct symptoms.

## Two causes
1. **Missing emit** — closing a Views window HIDES/recycles it, so `on_before_close` (the only caller of `report_window_closed`) never fires → the launcher never emits `WindowClosed` → the count stayed stale (`(2)` with 1 window). **Fix:** `close_window` reports `report_window_closed(label)` itself (idempotent; skips `browser-pane-*`).
2. **Lossy delivery** — each renderer reconstructs the window list from a versioned but best-effort launcher event stream; a dropped event left renderers permanently diverged (`3 vs 4` across windows). The gap detector (`event-buffer.ts`) only logged. **Fix:** a detected version gap bumps `launcherEventGapSeq`; the launcher-event reducer re-pulls the authoritative `list_window_instances` and reconciles (`ReconcileFromSnapshot`), with a version race-guard (the codex P1 #733 concern) so an in-flight event can't be clobbered by a stale snapshot.

## Verification (live, packaged build)
- 3 windows → close 1 → **all** windows converge to `(2)`.
- Minimize last window → **stays alive** (confirmed `IsIconic=True` while the tree is up) — no spurious quit.
- Close-all → clean teardown, no orphans.
- 66 vitest green (launcher-events + event-buffer + reducer); touched files type-clean.

## Specs
- `SPEC_WINDOW_COUNT_STALE_ON_VIEWS_CLOSE_2026_06_22.md` — both causes + fixes.
- `SPEC_REDUCER_SSOT_CONSOLIDATION_2026_06_22.md` — the broader single-source-of-truth audit this fix is **Track 3** of (six window-count computations across three data sources, the post-#1676 win_event quit authority that bypasses the reducer, the unwired `reconcile_quit`/H.6 runner, pane-pool typing, the HWND/role registry). Roadmap for follow-ups; this PR ships only Track 3.

Sibling of #1676 · Discussion #1680

### DIFF
--- a/.changesets/1782167695-fix-window-accurate-status-bar-window-count-on-close-event-s-uny9.md
+++ b/.changesets/1782167695-fix-window-accurate-status-bar-window-count-on-close-event-s-uny9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(window): accurate status-bar window count on close + event-stream gaps
--- a/agentmux-cef/src/commands/window/lifecycle.rs
+++ b/agentmux-cef/src/commands/window/lifecycle.rs
@@ -38,6 +38,19 @@ use cef::{ImplBrowser, ImplBrowserHost};
 pub fn close_window(state: &Arc<AppState>, args: &serde_json::Value) -> Result<serde_json::Value, String> {
     let label = args.get("label").and_then(|v| v.as_str()).unwrap_or("main");
 
+    // Report the logical window-close to the launcher so the frontend's window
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #1703, opened 2026-06-22
refactor(lifecycle): typed Floater BrowserKind — floaters excluded from quit gate by type

AgentA-asaf:
## Summary
SSOT keystone (Track 1 / finding **L4** of `SPEC_REDUCER_SSOT_CONSOLIDATION_2026_06_22.md`): give floaters a typed `BrowserKind::Floater` variant so the last-window quit gate excludes them **by type**, not by a fragile label-prefix string.

## The bug it fixes
Floaters had no `BrowserKind` variant — `floating-*` registered as `TopLevel { is_pool: false }`, and the count gate only excluded `floating-pool-`. So:
- `floating-<uuid>` (direct floater) → **counted** as an instance-keeping window
- `floating-pool-<uuid>` (pane-pool floater) → **excluded**

Same user-facing thing (a torn-off floating pane), opposite keep-alive behavior, decided purely by which code path created it.

## Change
- `BrowserKind::Floater { is_pool }` — both floater labelings classify into it; pane-pool promote flips `is_pool`.
- `is_live_user_window` = `matches!(kind, TopLevel { is_pool: false })` → floaters excluded **by type**; the `!starts_with("floating-pool-")` string and `counts_as_live_user_window` are removed.
- `list_top_level_browsers` now includes `Floater` (floaters are trusted renderers needing host JS events — the `matches!` site that would have silently broken floater event delivery).
- New `is_unpromoted_pane_pool_label` (mirror of `is_unpromoted_pool_label`).

## Policy — canonized
**Invariant FP-LIFE: floaters die with the last top-level window** — `ARCHITECTURE_FLOATING_PANE_DOCKING_2026_05_30.md` §1.1 (user-confirmed). Aligns the reducer count with the #1676 win_event quit trigger, which already excludes floaters by window-class. The doc records that changing this must touch BOTH the win_event trigger and the reducer count together.

## Scope / out of scope
- `host_counts_snapshot` (launcher-mirror feed) and the `pending_window_creations` prefix classifier are deliberately left label-based — separate concerns (noted in spec L4).
- L4 of the roadmap; L2 (wire `reconcile_quit`), L3 (collapse the six counts), L1 (retire the win_event parallel quit authority) are follow-ups.

## Verification
- 140 `agentmux-cef` lib tests green (incl. new `is_live_user_window_counts_only_top_level_by_type`); cargo check clean; all 8 `BrowserKind` match sites audited.
- Live smoke on a packaged build — tear-off/redock floaters, FP-LIFE behavior intact. (A pre-existing, unrelated browser-black-on-redock race surfaced during smoke — diagnosed via `[pane-trace]`, tracked separately on Discussion #1205; not this change.)

Track 1 / L4 of SPEC_REDUCER_SSOT_CONSOLIDATION · sibling of #1676 / #1701

### DIFF
--- a/.changesets/1782171094-fix-lifecycle-typed-floater-browserkind-floaters-excluded-fr-g6fe.md
+++ b/.changesets/1782171094-fix-lifecycle-typed-floater-browserkind-floaters-excluded-fr-g6fe.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(lifecycle): typed Floater BrowserKind; floaters excluded from quit gate by type
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -425,18 +425,29 @@ impl AgentMuxHandler {
         // `browser-pane-`). LABEL is the source of truth. See
         // docs/retro/smoke-test-0.33.586-and-pr5-plan-2026-05-02.md.
         //
-        // Classification:
-        //   - label `browser-pane-<uuid>-<seq>` → Pane { block_id: uuid }
-        //   - label `window-pool-*` + still in unpromoted_pool_labels →
-        //     TopLevel { is_pool: true }
+        // Classification (LABEL is the source of truth at registration; after
+        // this the typed `BrowserKind` is authoritative — don't re-parse labels):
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #1705, opened 2026-06-22
fix(agent-pane): busy bar freezes under reduced-motion; make bar opaque

a5af:
Refines the #1694 marching-ants busy indicator. Two defects seen running v0.48.1 on Windows 11.

## 1. "Stuck" bar under reduced-motion (the Win11 report)
`@media (prefers-reduced-motion: reduce)` set `animation: none` with no replacement, so the diagonal stripe froze mid-phase at full opacity and read as a hung bar.

CEF maps `prefers-reduced-motion: reduce` to the Windows **Animation effects** toggle (`SPI_GETCLIENTAREAANIMATION`). Verified **OFF** on the affected Win11 box, so it lands in that branch; a Win10 box with animations on plays normally. This is **per-machine OS setting**, not a Win10-vs-Win11 code path — and it **regressed #1694**, which deleted the prior aurora bar's static reduced-motion fallback.

**Fix:** replace the bare `animation: none` with a non-positional opacity `agent-ant-breathe` on a solid bar, so "working" still reads without motion. Scope `will-change` to the active state only (no permanent idle layer).

## 2. Text bleeds through the stripe gaps
Gap stops used `color-mix(..., transparent)` and the bar had no opaque background, so content reaching the top edge showed through as the ants marched.

**Fix:** opaque base `background: var(--main-bg-color)` (the pane's own surface), and mix the faded stripe against `--main-bg-color` instead of `transparent` so every stop is opaque and the pattern stays crisp in any theme.

## Scope
- One file: `frontend/app/view/agent/styles/_control-bar.scss`. No JS/markup changes.
- Spec: `docs/specs/SPEC_AGENT_BUSY_ANTS_REFINEMENT_2026_06_22.md`
- Frontend build verified (SCSS compiles).

## Verify
- DevTools ▸ Rendering ▸ **Emulate `prefers-reduced-motion`** → bar breathes instead of freezing; disable → ants march.
- Start a turn with text reaching the top edge → no bleed-through, light + dark themes.

### DIFF
--- a/.changesets/1782171756-fix-agent-pane-busy-bar-no-longer-freezes-under-reduced-moti-nm9r.md
+++ b/.changesets/1782171756-fix-agent-pane-busy-bar-no-longer-freezes-under-reduced-moti-nm9r.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): busy bar no longer freezes under reduced-motion + opaque bar (no text bleed-through)
--- a/docs/specs/SPEC_AGENT_BUSY_ANTS_REFINEMENT_2026_06_22.md
+++ b/docs/specs/SPEC_AGENT_BUSY_ANTS_REFINEMENT_2026_06_22.md
@@ -1,0 +1,240 @@
+# Agent Busy Bar (Marching Ants) Refinement
+
+**Date:** 2026-06-22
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #1669, opened 2026-06-21
fix(swarm): show agent panes in Swarm view

a5af:
## Summary

- **Root cause:** `AgentTrackedBlocksCommand` sourced block IDs only from `process_tracker`, which only registers blocks when an OS subprocess spawns via `ensure_tracker()`. Claude Code agent panes never spawn OS-level subprocesses, so they were never visible in Swarm.
- **Rust fix:** Added `list_active_blocks()` to `ReactiveHandler` (reads `block_to_agent` keys) and unioned those IDs with process tracker IDs in `AgentTrackedBlocksCommand`.
- **Frontend fix:** `buildTree()` in `swarm-model.ts` also includes unique `parent_block_id` values from active subagents as a safety net for race conditions where subagents spawn before their parent's registration propagates.

## Test plan

- [ ] Launch AgentMux and open Swarm pane
- [ ] Open one or more Agent panes — they should appear immediately in Swarm tree
- [ ] Verify agent status badge (running/idle) reflects actual agent activity
- [ ] Launch a deep-research / workflow that spawns subagents — parent + children should appear in tree
- [ ] Close an agent pane — confirm it disappears from Swarm on next refresh

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1708, opened 2026-06-23
fix(swarm): emit agent:process-added on reactive register so panes appear in Swarm view

a5af:
## Root cause

The Swarm model's `loadTrackedBlocks()` is only triggered by `agent:process-added` and `agent:process-exited` events, which the OS process tracker emits for subprocess-based agents. Claude Code agent panes (Korp, Lark, Loap, Naki, etc.) register via the reactive handler (`/agentmux/reactive/register`) and have no OS subprocess — so no event was ever emitted, and the Swarm view never refreshed to show them.

PR #1669 already fixed `AgentTrackedBlocksCommand` to union process tracker + reactive handler blocks, but that only helps when the Swarm view actually re-polls. Without an event to trigger a re-poll, the panes remained invisible.

## Fix

In `handle_reactive_register` (server/reactive.rs): publish `agent:process-added` after a successful registration.
In `handle_reactive_unregister`: look up the block_id, then publish `agent:process-exited` after removal.

No frontend changes needed — the Swarm model's existing subscriptions already call `loadTrackedBlocks()` on these events.

## Test

1. Open the Swarm pane
2. Open a new Agent pane — it should appear in Swarm within the next event cycle
3. Close the Agent pane — it should disappear from Swarm


a5af:
@codex review

PR #1711, opened 2026-06-23
fix(agent-pane): make the busy bar zoom-invariant

a5af:
Refines the marching-ants busy bar so it stays a **fixed size regardless of pane zoom**.

## Problem
The bar lives inside `.agent-view`, which applies per-pane zoom via the CSS `zoom` property (`agent-view.tsx`, range 0.5-2.0). As a child, the bar scaled with it: 6px tall with a doubled stripe period/march at 2x, shrunk at 0.5x. The busy indicator is chrome, not content, so it should read at a constant size.

## Fix
Expose the zoom factor as `--agent-pane-zoom` alongside the `zoom` on the root div, and divide every px dimension by it:
- bar `height`
- `::before` overshoot (`left`/`right`)
- stripe period (`repeating-linear-gradient` stops)
- march distance (`@keyframes agent-ant-march` translateX)

`zoom` re-multiplies all lengths in the subtree, so dividing first renders the bar at a fixed screen size at any pane zoom. Overshoot is scaled by the same factor so it stays larger than the (also-scaled) march distance and edge coverage holds. The custom property inherits to `::before` (so the keyframe resolves it per element); fallback `1` preserves current behavior if absent.

Width still spans the pane, by design.

## Scope
- `frontend/app/view/agent/agent-view.tsx` (expose the var)
- `frontend/app/view/agent/styles/_control-bar.scss` (counter-scale)
- Spec updated: `SPEC_AGENT_BUSY_ANTS_REFINEMENT_2026_06_22.md` §3b
- Frontend build verified.

## Verify
Zoom an agent pane in/out (Ctrl+wheel) during a turn — the busy bar's thickness, stripe size, and march speed stay constant; only the pane content scales.

### DIFF
--- a/.changesets/1782176589-fix-agent-pane-keep-the-busy-bar-a-fixed-size-regardless-of--56wl.md
+++ b/.changesets/1782176589-fix-agent-pane-keep-the-busy-bar-a-fixed-size-regardless-of--56wl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): keep the busy bar a fixed size regardless of pane zoom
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #1712, opened 2026-06-23
fix(memory): brain pane agent-not-found, autoMemory guard, inline new-file input, seed discipline

a5af:
## Summary

- **`agent_def_get` global registry fallback** — brain pane RPCs (`agent:memory:list`, `agent:memory:read_file`, `agent:memory:write_file`) were failing with `"agent not found"` for cross-channel agents. `agent_def_list` already overlaid the global `DefinitionStore`; `agent_def_get` didn't. Fixed by checking the global registry when SQLite returns `None`.

- **`autoMemoryEnabled` guard tightened** — the guard was disabling Claude Code's autonomous memory writes for _any_ path under `~/.agentmux/agents/`, including agents with unique per-instance workdirs like `~/.agentmux/agents/lark-06209`. The shared-slug concern only applies to the bare fallback path (`~/.agentmux/agents/<slug>`). Fixed to exact-match the slug-fallback path only; all other workdirs keep memory writes enabled.

- **Brain pane inline new-file input** — replaced `window.prompt()` with an inline input row in the file list sidebar. Validates filename before the RPC round-trip, supports Enter/Escape keyboard shortcuts, shows inline error on invalid names, dismisses cleanly.

- **`seed-agent-memory` global brain bundle** — new seeded memory bundle (global, injected into every agent's CLAUDE.md) teaching agents when and how to use the native memory system: MEMORY.md index discipline, topic file format, what to save vs not save. Also fixed `agent-seed.json` encoding from cp1252 to UTF-8.

## Test plan

- [ ] Open brain pane for a cross-channel agent — no "agent not found" error
- [ ] Open brain pane for any agent — "+ New file" shows inline input, Enter creates, Escape cancels, invalid name shows inline error
- [ ] Agent with explicit (non-bare-slug) workdir starts a session — `autoMemoryEnabled` not forced to false in settings.json
- [ ] New install — Trust Center > Brain shows "Agent Memory" as a global bundle

### DIFF
--- a/.changesets/1782176393-fix-memory-agent-def-get-global-registry-fallback-automemory-4d5x.md
+++ b/.changesets/1782176393-fix-memory-agent-def-get-global-registry-fallback-automemory-4d5x.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(memory): agent_def_get global-registry fallback, autoMemory guard, brain UI inline input, seed memory discipline
--- a/agentmux-srv/agent-seed.json
+++ b/agentmux-srv/agent-seed.json
@@ -4,7 +4,7 @@
     {
       "id": "claude",
       "name": "Claude",
-      "icon": "✖",
+      "icon": "\u00e2\u0153\u2013",
       "provider": "claude",
       "agent_type": "container",
       "container_image": "ghcr.io/agentmuxai/agent-claude:latest",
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

a5af:
Fixed both P1s in 056bbf1c:

- **agent-seed.json encoding**: reverted to the original valid-UTF-8 source before adding the new bundle. The cp1252-round-trip approach was wrong direction — the original was already UTF-8. New entry uses `ensure_ascii=True` so the file stays ASCII-safe on any platform. Icons and em-dashes are intact.
- **autoMemoryEnabled guard**: replaced slug-comparison with `agent.working_directory.is_empty()`. The slug passed to `write_agent_config_files` is `routing_id` (= `agent.slug`), but `work_dir` was built from a name-derived slug — they diverge on renamed agents. `working_directory.is_empty()` is authoritative: the caller only uses the fallback path when this field is unset.

PR #1659, opened 2026-06-21
fix(cef): disable Web Notifications to stop duplicate macOS permission toasts

a5af:
## Summary

- On every new-version install, macOS shows **two** "Notifications may include alerts, sounds, and icon badges" permission toasts — one from AgentMux (banner-style) and one from AgentMux Helper (Alerts) (alert-style)
- This happens because CEF registers both processes as OS notification sources and requests permission for each at startup; PR #1646's per-version bundle IDs make every upgrade trigger a fresh permission request
- We don't use `new Notification()` anywhere in the frontend — the Web Notifications API is unused
- Fix: add `--disable-notifications` Chromium switch in `on_before_command_line_processing`, which prevents CEF from registering either notification channel and eliminates both permission toasts

## Test plan

- [ ] Build DMG and install fresh (or reset notification permissions: `tccutil reset Notifications ai.agentmux.stable.X.Y.Z`)
- [ ] Launch — no notification permission toasts appear
- [ ] Agent turn-complete sounds still play (sound service is separate from Web Notifications)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1713, opened 2026-06-23
fix(macos): remove AgentMux Helper (Alerts) from bundle to eliminate duplicate notification toasts

a5af:
## Problem

On macOS 26 Tahoe, every fresh install of a new version shows **2 notification permission toasts** ("Notifications may include alerts, sounds, and icon badges"). PR #1659 added `--disable-notifications` which silenced the CEF→OS request path and was verified working in 0.47.3. The toasts are back in 0.48.1.

## Root cause

macOS 26 **proactively scans new app bundles on first launch** and prompts for both the main bundle (`ai.agentmux.stable.<ver>`) and the `AlertNotificationService` helper (`ai.agentmux.stable.<ver>.helper.alerts`) independently of whether the app calls `requestAuthorization`. `--disable-notifications` only silences the JS/CEF request path — it doesn't prevent the OS-level bundle discovery prompt.

## Fix

Remove `AgentMux Helper (Alerts).app` from the package script. AgentMux never uses OS notifications. With `--disable-notifications` in place, CEF never spawns the helper, so the binary in the bundle is dead weight that only serves to trigger the prompt.

**Verified:** built and notarized without the helper — `codesign --verify --deep --strict` passes, Apple accepted the notarization. `--disable-notifications` remains in place as a belt-and-suspenders guard against the JS API.

## Churn history

This is the 4th PR touching this issue. Adds a retro (`docs/retro/retro-macos-notification-double-prompt-regression-2026-06-22.md`) documenting why each prior approach was insufficient and how to prevent re-regression.

### DIFF
--- a/.changesets/1782176744-fix-macos-remove-alertnotificationservice-helper-from-bundle-rojg.md
+++ b/.changesets/1782176744-fix-macos-remove-alertnotificationservice-helper-from-bundle-rojg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(macos): remove AlertNotificationService helper from bundle to eliminate duplicate notification prompts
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1709, opened 2026-06-23
chore(ci): fast (PR) + nightly (CEF) test runners on standard GitHub-hosted runners

AgentA-asaf:
## Summary
Stands up the missing CI test runner. Until now there was **no workflow running `cargo test` or `vitest`** — the lifecycle fixes shipped this week (#1676/#1701/#1703) had zero automated coverage. Two workflows, both on **standard** GitHub-hosted runners (free on public repos; spec invariant CI-1 forbids larger/self-hosted — those are always billed):

- **`ci-fast.yml`** (`pull_request` + `push`, ubuntu-latest): the CEF-free crates (`launcher`/`srv`/`common`) + `vitest`. ~5 min, no CEF build. *This PR triggers it — the runner self-tests on its own introduction.*
- **`ci-nightly.yml`** (cron 09:00 UTC + `workflow_dispatch`, windows-latest): full `cargo test --workspace` incl. `agentmux-cef` (the reducer/lifecycle tests need the CEF link) + `vitest`. Windows so the `#[cfg(windows)]` host/lifecycle code compiles + runs. Installs Ninja; CMake ships with VS.

## Pre-existing test debt surfaced (the runner already earning its keep)
Standing this up flushed out **four** accumulated failures/flakes — none from this change. Handled minimally + tracked (spec §6.4) so the gate is green + meaningful now, with a "test-health" follow-up to remove the workarounds:
1. **Rust serial** (`--test-threads=1`) — `agentmux-cef` `allow_pane_focus_once_*` race a process-global `AtomicBool`.
2. **`agentmux-srv::write_then_read_roundtrip` `#[ignore]`d** — a read-cache keyed by id (not by `FileStore`) pollutes across in-memory stores; fails even serially.
3. **`tools/**` excluded from the frontend vitest** — standalone `.mjs` tooling tests, not frontend jsdom.
4. **`AgentLaunchModal` "auth state on memory change" `it.skip`'d — a REAL pre-existing frontend regression** (auth-ready state resets on a Memory selection change). Highest-priority follow-up.

## Verification
Both lanes green locally: vitest **1550 pass / 1 skip**; rust serial all `ok`.

## Open (spec §6)
- Whether `windows-latest` has enough disk for the CEF build (verify on first nightly; if not it's a maintainer cost decision, not a silent larger-runner opt-in).
- Cadence/matrix + caching iteration.

Implements `SPEC_CI_TEST_RUNNER_2026_06_22.md`.

PR #1714, opened 2026-06-23
fix(test): un-skip AgentLaunchModal preset test — stale Memory→Preset label

AgentA-asaf:
## Summary
The CI runner (#1709) flagged `AgentLaunchModal` "memory change must not reset auth state (§6.10)" as a possible auth regression, and it was skipped to unblock CI. **Triage result: stale test, NOT a product bug.**

The `Memory bundle` `<select>` was renamed to **Preset** (the Memory→Preset rename), so the test's `findByLabelText("Memory bundle")` failed *before* reaching the auth assertion — which read as an "auth regression." The §6.10 requirement holds: changing the preset does **not** reset auth (the assertion passes once the selector is correct).

- Fixed the selector → `findByLabelText("Preset")` + un-skipped. Test passes (verified locally + via this PR's `ci-fast` vitest).
- Spec §6.4 item 4 marked resolved.

First of the six CI-surfaced test-debt items closed — and a reminder to *triage* "real regression vs stale test", not assume.

### DIFF
--- a/.changesets/1782178546-fix-test-un-skip-agentlaunchmodal-preset-test-stale-memory-p-x16s.md
+++ b/.changesets/1782178546-fix-test-un-skip-agentlaunchmodal-preset-test-stale-memory-p-x16s.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1715, opened 2026-06-23
fix(agent-pane): keep marching ants under reduced-motion (gentler), drop the breathe

a5af:
Follow-up to #1705. A user on Windows 11 with **Animation effects off** reported the busy bar "appears now as aurora."

## Cause
#1705's reduced-motion fallback swapped the marching ants for an opacity **breathe** on a solid bar. On a reduced-motion machine (Windows "Animation effects" off → CEF reports `prefers-reduced-motion: reduce`), that renders as a soft pulsing glow that reads as "aurora" — a **different effect** from the marching ants the indicator is meant to show, and not what motion-enabled machines see. (There is no actual aurora animation in the code; it was the breathe.)

## Fix
Keep a single effect everywhere. Under reduced motion, run the **same `agent-ant-march` keyframe but slower** (override only `animation-duration` to 1.5s). The busy bar is a small (3px) looping *essential* "working" indicator, so a gentle march is an acceptable reduced-motion treatment — and it no longer looks like a different effect or a stuck/frozen bar.

- Motion-enabled machines: 0.5s march (unchanged).
- Reduced-motion machines: 1.5s march (gentler), same ants.
- Removes `agent-ant-breathe`.

Frontend build verified. Spec updated (`SPEC_AGENT_BUSY_ANTS_REFINEMENT_2026_06_22.md`).

Note: enabling Windows "Animation effects" also restores the full-speed 0.5s march.

### DIFF
--- a/.changesets/1782178592-fix-agent-pane-keep-marching-ants-under-reduced-motion-slowe-tr48.md
+++ b/.changesets/1782178592-fix-agent-pane-keep-marching-ants-under-reduced-motion-slowe-tr48.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): keep marching ants under reduced-motion (slower) instead of an opacity breathe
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1716, opened 2026-06-23
feat(tool-log): in-place spinner animation in tool overlay

a5af:
## Summary

- Consecutive single-character spinner frames (⠋⠙⠹⠸… braille, ◐◓◑◒ quarter-circle, `/-\|` ASCII) were stacking as separate `<pre>` lines in the tool overlay and shell log panels
- They now animate **in place**: a trailing run of spinner chars is held in a single reactive slot that Solid updates in-place as each frame arrives, then freezes on the last character when the tool finishes or non-spinner output follows — matching real terminal behaviour
- Frontend-only change (~40 lines). No backend or protocol changes

## How it works

`ChunkList` (in `ToolOverlayLog.tsx`) and the log panel in `PersistentShellBlock.tsx` now run a `createMemo` that walks the capped chunk list:
- **Trailing spinner run** → goes into `spinnerSlot`, rendered as a single `<pre>` whose text Solid updates in place
- **Completed spinner run** (followed by non-spinner output) → last frame frozen as a static line
- **No spinner chars** → identical to today, no overhead

`SPINNER_CHARS` covers braille (ora/listr/tqdm), quarter-circle, and ASCII classic sets.

## Test plan
- [ ] Run an agent task that uses npm install or cargo build — spinner chars should animate in one line, not stack
- [ ] After the tool finishes, the last spinner char should remain frozen (no CSS spin)
- [ ] Non-spinner output following a spinner run should appear on a new line with the last spinner char frozen above it
- [ ] Tools with no spinner output are unaffected

### DIFF
--- a/.changesets/1782209542-feat-tool-log-collapse-consecutive-spinner-char-frames-to-in-z5vb.md
+++ b/.changesets/1782209542-feat-tool-log-collapse-consecutive-spinner-char-frames-to-in-z5vb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(tool-log): collapse consecutive spinner-char frames to in-place animation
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1724, opened 2026-06-23
fix(cef): disable GPU process in macOS dev mode to avoid sandbox crash

a5af:
## Problem

`task dev` on macOS crashes immediately. The GPU process dies 6 times in rapid succession:

```
GPU process exited unexpectedly: exit_code=5
GPU process has crashed 1 time(s)
...
GPU process has crashed 6 time(s)
FATAL: GPU process isn't usable. Goodbye.
```

## Root cause

The dev binary runs unsigned from a flat directory (`dist/cef-dev/`), not inside a signed `.app` bundle. macOS gates Metal GPU context initialization on code-signing + proper bundle structure. Without it the GPU subprocess's sandbox blocks Metal access — exit_code=5 every time.

The **installed** app (0.47.x, 0.48.x) is properly signed and its GPU helper runs fine. Only `task dev` hits this.

## Fix

When `AGENTMUX_DEV=1` is set (macOS, browser process only), add `--disable-gpu` so CEF falls back to software rendering. UI visuals are identical; GPU compositing is unnecessary for dev iteration.

```rust
#[cfg(target_os = "macos")]
if process_type.is_none() && std::env::var("AGENTMUX_DEV").is_ok() {
    cmd.append_switch(Some(&CefString::from("disable-gpu")));
}
```

## Test plan

- [ ] `task dev` on macOS: app launches, no GPU crash in logs
- [ ] `task dev` on macOS: UI renders correctly (software rendering)
- [ ] Installed 0.48.x on macOS: GPU helper still runs (not affected — `AGENTMUX_DEV` not set)

### DIFF
--- a/.changesets/1782213380-fix-cef-disable-gpu-process-in-macos-dev-mode-to-avoid-sandb-8ury.md
+++ b/.changesets/1782213380-fix-cef-disable-gpu-process-in-macos-dev-mode-to-avoid-sandb-8ury.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-fix(cef): disable GPU process in macOS dev mode to avoid sandbox crash
+fix(cef): use --no-sandbox in macOS dev mode so unsigned dev binary can init GPU and network service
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1726, opened 2026-06-23
feat(layout): pane minimize — collapse to header bar

a5af:
## Summary

- Adds a working **minimize button** (chevron-down/chevron-up) to every tiled pane header bar
- Clicking minimize collapses the pane to its 33 px header bar and gives the freed space to an adjacent sibling; clicking again restores the original size
- Minimized state persists across page reloads via the existing `rootNode` WaveObject (stored in `node.minimizedSize`)
- Button is only visible when `numLeafs > 1` (can't minimize the only pane) and hidden when the pane is magnified or ephemeral

## Architecture

New file `layoutMinimize.ts` handles all geometry:
- `minimizeNodeToggle(model, nodeId)`: reads parent's `pixelToSizeRatio` from `additionalProps`, sets `node.size = (33 + gap) * ratio`, saves original in `node.minimizedSize`, redistributes freed units to sibling. Restore reclaims the same amount, clamped so sibling can't go below its own header height.
- `rebuildMinimizedSet(model)`: scans the loaded tree for `minimizedSize !== undefined` to rebuild the reactive `minimizedNodeIds` Set on startup.

The layout geometry engine (`updateTreeHelper`) naturally clips content via `overflow: hidden` on `.tile-leaf` — no extra CSS needed to hide pane content.

`layoutResize.ts` now skips the `MinNodeSizePx` guard for minimized nodes so the resize handle beside a minimized pane doesn't block dragging.

## Test plan

- [ ] Open multiple panes; click chevron-down on one — it collapses to header bar, neighbor expands to fill
- [ ] Click chevron-up (same button) on the collapsed pane — it restores to original size
- [ ] Reload page — minimized pane stays minimized
- [ ] Drag the resize handle between a minimized pane and its neighbor — should move freely
- [ ] With only 1 pane open — minimize button should not appear
- [ ] Magnify a pane — minimize button should not appear
- [ ] Test on macOS, Linux, Windows (all three TileLayout variants)

### DIFF
--- a/frontend/app/block/block.scss
+++ b/frontend/app/block/block.scss
@@ -280,6 +280,12 @@
                         padding: 0;
                     }
 
+                    .block-frame-minimize {
+                        justify-content: center;
+                        align-items: center;
+                        padding: 0;
+                    }
+
                     .block-frame-btn-separator {
                         width: 1px;
                         height: 14px;
--- a/frontend/app/block/blockframe.tsx
+++ b/frontend/app/block/blockframe.tsx
@@ -170,6 +170,16 @@ function getViewIconElem(viewIconUnion: string | IconButtonDecl, blockData: Bloc
     }
 }
 
+function OptMinimizeButton(props: { minimized: boolean; toggleMinimize: () => void }): JSX.Element {
+    const decl = createMemo<IconButtonDecl>(() => ({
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #1727, opened 2026-06-23
fix(container-agent): empty image fallback, skip empty meta, guard null blockId in swarm

a5af:
## Summary

Three bugs causing container agents (e.g. Pconat) to silently fail to load:

- **`app_api.rs`**: `meta_get_string` returns `""` when `"agent:container_image"` is present-but-empty (the frontend always writes the key). Added empty-string guard to fall back to `ghcr.io/agentmuxai/agent-claude:latest` so agents with no explicit image still start. Root cause: `meta_get_string(meta, key, default)` returns the value, not the default, when the value is `""`.

- **`agent-model.ts`**: `launchAgentDefinition` always wrote `"agent:container_image": ""` to block meta when neither `overrides.containerImage` nor `agent.container_image` was set, defeating the backend default. Now skips the key if both are blank, so the backend default applies.

- **`swarm-view.tsx`**: `AgentRow` called `focusBlock(node.blockId)` even when `blockId` was `null` (agent never launched — not yet registered with muxbus). Click was a silent no-op that confused users into thinking the pane opened. Added `node.blockId &&` guard.

## Reproduction
Create a container agent without setting `container_image` → agent definition gets `container_image: ""` → open pane → first user message → `ensure_running("...", "", ...)` → Docker fails → pane shows nothing.

## Test plan
- [ ] Container agent with `container_image: ""` in definition: send first message → container starts with default image (not a Docker error)
- [ ] Container agent with explicit `container_image` set: still uses the explicit image
- [ ] New container agent opened via agent picker: `agent:container_image` key absent from block meta when image is blank
- [ ] Click unstarted agent in swarm view: click does nothing (no console error) rather than a broken focusBlock(null)

### DIFF
--- a/.changesets/1782232002-fix-container-agent-empty-image-fallback-skip-empty-meta-gua-l8c6.md
+++ b/.changesets/1782232002-fix-container-agent-empty-image-fallback-skip-empty-meta-gua-l8c6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(container-agent): empty image fallback, skip empty meta, guard null blockId in swarm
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #1732, opened 2026-06-23
fix(auth): Use existing login primary for Claude; Login via terminal fallback

AgentA-asaf:
## Summary

- **"🌐 Use existing login"** is now the **primary** action for Claude auth failures in the in-agent failure banner — this is the reliable seed-from-global path (no browser spawn needed)
- **"🖥 Login via terminal"** replaces "Login Again" for Claude — opens a real `cmd.exe` window with `CREATE_NEW_CONSOLE` so the OS can launch the browser for OAuth, then polls for credentials every 5s for up to 5 minutes and auto-seeds when they appear
- **Non-Claude providers** keep "Login Again" as primary (PTY path may work for them)
- 14 tests green (2 new cases: `canSeed=false` legacy path, `canSeed=true` Claude path)

**Root cause confirmed in logs:** clicking "Login Again" called `run_cli_login_pty` → "no auth URL captured within 15s" → silent hang → user sees nothing happen. The PTY is headless; Claude v2.1.x needs a real console to open the browser.

Closes the "Login Again does nothing" UX regression for Claude agents.

### DIFF
--- a/.changesets/1782229694-fix-auth-seed-from-global-primary-redact-tokens-in-login-pty-gqip.md
+++ b/.changesets/1782229694-fix-auth-seed-from-global-primary-redact-tokens-in-login-pty-gqip.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): seed-from-global primary; redact tokens in login PTY logs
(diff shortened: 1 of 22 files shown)


a5af:
@codex review

PR #1733, opened 2026-06-23
fix(container-agent): force subprocess controller at block creation, guard resync

a5af:
## Summary

- Container agents (agentMode == "container") require `SubprocessController` for per-turn docker exec. Claude's provider returns `"persistent"` as its default controller type, so new container agent blocks were created with `controller:"persistent"` in their meta.
- `AgentInput` matched the persistent branch, saw `agentMode=="container"`, and returned an RPC error that never reached the blockfile — the pane appeared blank with no feedback.
- **Root cause traced to**: `app_api.rs:246` — `controller_type = provider.controller_type_str()` with no override for container agents.

**Fix 1 — `app_api.rs`**: Override `controller_type` to `"subprocess"` when `agent.agent_type == "container"` before writing block meta. New container agent blocks will be created with the correct controller type.

**Fix 2 — `blockcontroller/mod.rs`**: In `resync_controller`, override stale `"persistent"` → `"subprocess"` when `agentMode == "container"`. Existing blocks created before this fix are self-correcting at the next `ControllerResync` without a DB patch.

## Test plan

- [ ] Open a container agent (e.g. Pconat with `container_image` set) — pane should load and start Claude inside Docker
- [ ] Send a message — docker exec fires, no "requires subprocess controller" RPC error
- [ ] Non-container (host) Claude agents continue to use persistent controller unchanged
- [ ] Existing container agent blocks with stale `controller:"persistent"` self-correct on `ControllerResync`

### DIFF
--- a/.changesets/1782238012-fix-container-agent-force-subprocess-controller-type-at-bloc-go6c.md
+++ b/.changesets/1782238012-fix-container-agent-force-subprocess-controller-type-at-bloc-go6c.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(container-agent): force subprocess controller type at block creation; guard resync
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1735, opened 2026-06-23
perf(layout): remove 50ms debounce on window resize — live HWND tracking

a5af:
## Summary

- The 50ms trailing debounce on `useOnResize` for `onContainerResize` caused native browser pane HWNDs to freeze for the entire drag and snap to the final position ~50ms after the user stopped resizing the window. This made the REFLOW_WINDOW_MS reduction in #1699 (32ms → 4ms) invisible on the window-resize path.
- Root cause traced: 50ms debounce → delayed `updateTree()` → delayed placeholder rect update → delayed `browser_pane_resize` IPC. All downstream steps were already instant.
- Fix: drop the debounce (`debounceMs = null`, the default), so `onContainerResize` fires per animation frame via ResizeObserver's own batching — same rate as the splitter-drag path (`onResizeMove` calls `updateTree()` every `mousemove`).

## Test plan

- [ ] Resize the window by dragging the edge — panes should track the window edge at native frame rate with no trailing snap
- [ ] Pane open/close/split still works (different path: `notifyPaneReflow` / `paneReflowActive`, unaffected)
- [ ] Splitter drag still works (`onResizeMove`, unaffected)

### DIFF
--- a/.changesets/1782240141-perf-layout-remove-50ms-debounce-on-container-resize-hwnd-st-w68r.md
+++ b/.changesets/1782240141-perf-layout-remove-50ms-debounce-on-container-resize-hwnd-st-w68r.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+perf(layout): remove 50ms debounce on container resize — HWND stays in sync during window resize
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1739, opened 2026-06-23
fix(auth): inline agent_error Login Again uses seedGlobalLogin for Claude

AgentA-asaf:
The inline `agent_error` node (HTTP 401/403 from Claude API mid-turn) surfaced a "Login Again →" button that called `relogin()` → `forceProviderLogin` → PTY spawn → hangs silently for 15s on Claude v2.1.x (the CLI login TUI is headless, browser never opens).

For Claude specifically, route `onAgentErrorLogin` to `useGlobalLogin()` (seed-from-global) instead — same fix as the failure banner in #1732 but for the inline error path that was missed.

Non-Claude providers keep the existing `relogin()` path.

Follow-up to #1732.


a5af:
@codex review

PR #1740, opened 2026-06-24
fix(composer): ArrowUp/Down respects soft-wrapped visual lines

a5af:
## Summary

- ArrowUp inside a multi-line soft-wrapped composer (no `\n`) immediately jumped to the previous sent message instead of moving the cursor up one visual row
- Root cause: `caretOnFirstLine` used `includes("\n")` — only detects physical newlines, not CSS word-wrap boundaries
- Fix: `caretVisualEdge()` helper uses a mirror div to measure actual visual row Y via `span.offsetTop`, so soft-wrapped lines are counted correctly

## How it works

`caretVisualEdge(ta)` replicates the textarea's CSS (font, padding, width, `white-space:pre-wrap`) in a temporary off-screen `div`, inserts text up to the caret position, and measures the zero-width-space marker's `offsetTop`. Comparing that Y to the baseline Y (position 0) detects "first visual line"; comparing to the end-of-text Y detects "last visual line".

Fast-path: when a physical `\n` exists before/after the cursor the answer is settled by character scan alone — no DOM element is created.

## Test plan

- [ ] Type a long message that soft-wraps to 3+ lines (no Shift+Enter). ArrowUp on line 3 should move cursor up, not load previous message
- [ ] ArrowUp when cursor is on the first visual row (even of a soft-wrapped message) should navigate to previous sent message
- [ ] ArrowDown on last visual row navigates to next history entry; on other rows moves cursor down
- [ ] Single-line message: ArrowUp navigates history as before
- [ ] Multi-line message with `\n` (Shift+Enter): ArrowUp on line 2 moves cursor up; ArrowUp on line 1 navigates history

Retro: `docs/retros/RETRO_COMPOSER_ARROWKEY_HISTORY_SOFT_WRAP_2026_06_23.md`

### DIFF
--- a/.changesets/1782259316-fix-composer-use-mirror-div-for-visual-line-detection-in-arr-nnsq.md
+++ b/.changesets/1782259316-fix-composer-use-mirror-div-for-visual-line-detection-in-arr-nnsq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(composer): use mirror-div for visual line detection in ArrowUp/Down history nav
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #1741, opened 2026-06-24
fix(agent-pane): re-enter Streaming from Done on tool-continuation rounds

a5af:
## Summary

- `session_end` fires after every model API round, not just at the end of a user turn — after the first round, `TurnPhase` transitions to `Done`, and the next model round (driven by a tool result) had no way to re-enter `Streaming`
- `StreamFlushObserved` and `bumpEvent` now include `Done` in the set of phases that promote to `Streaming`, so the busy indicator and working row stay active through multi-round tool-continuation turns
- `Interrupting` remains excluded (an in-flight interrupt should not be un-done by a stray flush)

See `docs/retro/retro-busy-indicator-turnphase-gap-tool-continuation-2026-06-23.md` for full root-cause analysis.

## Test plan

- [ ] Send a message that triggers multiple tool calls in sequence — verify busy indicator (marching ants) stays active throughout all tool rounds, not just the first
- [ ] Interrupt mid-turn — verify indicator stops and does not re-activate spuriously
- [ ] Single-round turn (no tools) — verify indicator still clears normally on `TurnEnd`

### DIFF
--- a/.changesets/1782261020-fix-agent-pane-guard-done-streaming-promotion-to-outcome-com-d34v.md
+++ b/.changesets/1782261020-fix-agent-pane-guard-done-streaming-promotion-to-outcome-com-d34v.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): guard Done→Streaming promotion to outcome=completed only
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1742, opened 2026-06-24
fix(dev): avoid EXE lock on Windows when re-running task dev

a5af:
## Summary

- On Windows, `rm -rf dist/cef-dev` fails when `agentmux-launcher.exe` is running from that dir — the OS locks the EXE image for the process lifetime, so a re-run of `task dev` exits 201 and requires a manual `Stop-Process`
- `dev:serve` now uses `dist/cef-dev-<epoch>` on Windows (fresh dir per run, no lock contention); a best-effort prune loop removes old stamped dirs whose launchers have since exited
- Unix keeps the fixed `dist/cef-dev` name so `install-linux-desktop.sh` retains a stable `Exec=` path between dev sessions
- `clean:host` gains a glob step to wipe `dist/cef-dev-*`

See `docs/retro/retro-task-dev-isolation-multi-agent-2026-06-23.md` for full root-cause analysis.

## Test plan

- [ ] Run `task dev` as a background task, let it launch — then run `task dev` again without stopping the first one; verify the second run succeeds instead of exiting 201
- [ ] Verify `task clean:host` removes all `dist/cef-dev-*` dirs
- [ ] On Linux: verify `task dev` still produces a stable `dist/cef-dev/` path (desktop entry unaffected)

### DIFF
--- a/.changesets/1782259563-fix-dev-timestamp-stamp-cef-dev-dir-on-windows-to-avoid-exe--qurq.md
+++ b/.changesets/1782259563-fix-dev-timestamp-stamp-cef-dev-dir-on-windows-to-avoid-exe--qurq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dev): timestamp-stamp cef-dev dir on Windows to avoid EXE lock on re-launch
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #1751, opened 2026-06-24
refactor(frontend): useTick shared interval — 11 polling sites → 1 interval per period

a5af:
## Summary

- Adds `frontend/app/hook/useTick.ts` — a 50-line ref-counted shared `setInterval` primitive. All callers with the same period share one underlying interval; the interval clears automatically when the last subscriber unmounts.
- Migrates 11 sites that each owned an independent 1 s or 60 s `setInterval` to `useTick`: `ActivityRow`, `PersistentShellBlock`, `ActivityDock`, `AgentFooter` (elapsed), `AgentComposerStrip` (elapsed), `ToolBlock` (`ToolElapsedTicker`), `DiagPanel`, `HostSection` + `LanSection` (warden), `useNotification`, `MyAgentsList`.
- Peak: ≥13 independent 1 s intervals in a loaded session → 1 shared interval; 2 independent 60 s intervals → 1.
- Removes `globalThis.__nowMinuteInterval` singleton hack from `termViewModel.ts` (the 60 s sites now share the same `tickers` map).
- No behaviour changes — elapsed timers, retention windows, and age columns all work identically.

## Key design decisions

- **`??` short-circuit subscription gating** (e.g. `ActivityRow`, `PersistentShellBlock`): `a.endedAt ?? (tick(), Date.now())` naturally subscribes to `tick` only while the activity is running, because SolidJS prunes reactive subscriptions that weren't called during the last evaluation.
- **`createMemo(() => (tick(), Date.now()))`** for components that always need current time while mounted (warden, diag-panel, ActivityDock): reads `tick()` on every second, returns fresh `Date.now()`. Signal-compatible — callers keep `now()` unchanged.
- **`loadStartMs` signal + `tick` memo** for elapsed-since-loading-start timers (AgentFooter, AgentComposerStrip): a `createSignal<number | null>` tracks when loading started; the memo computes `Date.now() - loadStartMs` on every tick.

## Test plan

- [ ] Open multiple agent panes with running tools → elapsed timers tick correctly in all
- [ ] Open Warden → Host section age column updates every second
- [ ] Open an activity dock with a running shell → elapsed shows; after shell exits, dock row fades within retention window
- [ ] Trigger a notification with an expiration → it disappears after its timeout
- [ ] Open My Agents list → relative timestamps update on the minute

### DIFF
--- a/.changesets/1782282367-refactor-frontend-usetick-shared-interval-11-1s-60s-sites-1--sbvd.md
+++ b/.changesets/1782282367-refactor-frontend-usetick-shared-interval-11-1s-60s-sites-1--sbvd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(frontend): useTick shared interval — 11 1s/60s sites → 1 interval per period
--- a/docs/analysis/ANALYSIS_POLLING_LANDSCAPE_2026_06_23.md
+++ b/docs/analysis/ANALYSIS_POLLING_LANDSCAPE_2026_06_23.md
@@ -1,0 +1,237 @@
+# Polling Landscape Analysis — AgentMux
+
+**Date:** 2026-06-23  
+**Scope:** All timer-based repeating work in frontend (TS/SolidJS) and backend (Rust/tokio)  
+**Question:** Is a unified polling framework worth building?
+
+---
+
+## TL;DR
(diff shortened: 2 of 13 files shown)


a5af:
@codex review

PR #1752, opened 2026-06-24
fix(swarm): provider icon + fix persistent agents showing working when idle

a5af:
## Summary

- Replace the ⬡ hexagon icon with the provider's brand logo (`ProviderLogo` component), driven by `block.meta["agentProvider"]`. Falls back to the AgentMux brain mark when no provider is set.
- Fix false "working" status for persistent/reactive agents (e.g. Claude Code): `shellprocstatus="running"` only means the process is alive, not that it's actively doing LLM work. When no `TurnPhase` is registered for a block, the status now defaults to "idle" instead of "working".

## Root cause of status bug

`phaseToDisplayStatus` fell back to `shellprocstatus === "running" → "working"` when `getBlockTurnPhase` returned null. For persistent controller agents (Claude Code, Codex), the shell process is always alive between turns, so the fallback always returned "working". Fixed by returning "idle" when TurnPhase is absent — TurnPhase is the only reliable signal for active LLM work.

### DIFF
--- a/.changesets/1782280161-fix-editor-markdown-preview-source-hidden-guards-aria.md
+++ b/.changesets/1782280161-fix-editor-markdown-preview-source-hidden-guards-aria.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

PR #1754, opened 2026-06-24
fix(zoom): persist zoom for cross-channel agents via global registry

a5af:
## Summary

- **Root cause**: Agents created in a different channel (e.g. Naki, created in `local-agentx-fix-term-scrollbar-c-99fefa`) have no `db_agent_definitions` row in the current channel. The `db_agent_content` FK constraint rejects `INSERT`, so `agent_content_set` fails silently (warn-swallowed in `schedule_agent_zoom_mirror`). `ui:zoom` is never saved → zoom resets on every reopen.
- **Fix**: `agent_content_set` and `agent_content_delete` now check `db_agent_definitions` first. Cross-channel agents are routed to the new `registry_def_update_content_field` helper, which updates just the targeted content blob in the global registry record in-place (preserving all other fields + content). `agent_content_get`'s existing cross-channel fallback then surfaces the value on reopen.
- **Also fixed**: `agent_content_delete` for cross-channel agents no longer falls through to `registry_def_upsert` (which would have overwritten the global record with empty local content, silently wiping instructions/env).
- **Tests**: Two regression tests added in `def_registry_mirror.rs`.

## Test plan
- [ ] Rebuild and open Naki (cross-channel agent), Ctrl+Wheel zoom to e.g. 75%
- [ ] Close the pane, reopen Naki — zoom should restore to 75%
- [ ] Ctrl+Wheel back to 100%, close/reopen — zoom should reset to default (no zoom)
- [ ] Verify local agents (created in current channel) are unaffected: zoom still persists via local SQLite path
- [ ] `cargo check --tests` passes with no errors

### DIFF
--- a/.changesets/1782260462-feat-agent-pane-right-click-header-to-set-2-tone-hue-color-t-w5m4.md
+++ b/.changesets/1782260462-feat-agent-pane-right-click-header-to-set-2-tone-hue-color-t-w5m4.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): right-click header to set 2-tone hue color theme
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

PR #1757, opened 2026-06-24
fix(claude-translator): emit session_end per turn in persistent mode

a5af:
## Summary

- **Root cause:** In interactive/persistent mode, Claude Code only emits `{"type":"result"}` on process exit, not at the end of each turn. This left `TurnPhase` stuck in `Submitting` permanently, causing the swarm pane to show "working" indefinitely after a response finished.
- **Fix:** `handleAssistantMessage` now detects turn completion: a non-partial `assistant` message with no `tool_use` blocks is always the final text response of a turn. It emits a synthetic `session_end` so `TurnPhase` transitions `Done → idle`.
- **Subprocess mode safety:** In `--print` mode, the `result` event fires a second `TurnEnd` dispatch — this is a no-op due to first-done-wins in the reducer.

## How it works

```
assistant event (tool_use blocks) → tool call in progress → no session_end
assistant event (text/thinking only) → turn complete → emit session_end → TurnPhase = Done
```

## Test plan

- [ ] Send a message to a persistent agent; verify swarm shows "working" during response and transitions to "idle" when done
- [ ] Send a message that triggers tool calls; verify swarm stays "working" through all tool call rounds, then goes "idle" on the final text response
- [ ] Send a message in subprocess (--print) mode; verify swarm still transitions to idle correctly (no regression from duplicate TurnEnd)

### DIFF
--- a/.changesets/1782280161-fix-editor-markdown-preview-source-hidden-guards-aria.md
+++ b/.changesets/1782280161-fix-editor-markdown-preview-source-hidden-guards-aria.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(editor): guard sourceHidden display behind isMarkdown(), fix Mod+F in preview-only, add aria-label to </> toggle, restore previewOpen on exit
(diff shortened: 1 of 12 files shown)


a5af:
@codex review

PR #1758, opened 2026-06-24
fix(layout): solo-column pane slips header into adjacent column on minimize

a5af:
## Summary

- **Caret direction** — already correct on main (`chevron-up` when minimized, `chevron-down` when expanded via `OptMinimizeButton` in `blockframe.tsx:182`). No code change needed; confirmed via the reactive chain `minimizedNodeIds → isMinimized memo → EndIcons JSX prop → OptMinimizeButton createMemo`.

- **Solo-column minimize** — when a pane's layout-tree parent is a `Row` node, the previous code reduced `node.size` along the Row's flex direction, making the pane horizontally narrow (thin strip). Now it "slips" the header into the adjacent column instead:
  1. Pane is removed from the Row; its width goes to the sibling column.
  2. The pane's header is inserted at the top of that column at header-bar height.
  3. `node.slipMinimize` stores restore context (original Row index, width, whether the target was a leaf that needed column-conversion).
  4. Restore reverses: removes from column, re-inserts in Row at original index.

- Leaf-sibling edge case: when the adjacent Row slot is a plain leaf (not already a column branch), it is converted to a Column in-place during slip and unwrapped back to a leaf on restore.

- `rebuildMinimizedSet` extended to include `slipMinimize` nodes (needed for tree-reload after persistence).

## Test plan
- [ ] Open 2+ panes arranged side-by-side (horizontally). Click minimize on one — it should disappear from its column slot, and its header bar should appear at the top of the adjacent column.
- [ ] Click restore on the slipped header — it should re-appear at its original column position.
- [ ] Minimize a pane that shares a column with others (normal case) — should still collapse vertically to header height within the column (unchanged behavior).
- [ ] Verify the caret flips: chevron-down when expanded, chevron-up when minimized.
- [ ] Reload the app with a minimized pane in the slipped state — verify it restores correctly from persisted tree.

### DIFF
--- a/.changesets/1782311382-fix-layout-solo-column-pane-slips-header-into-adjacent-colum-a17h.md
+++ b/.changesets/1782311382-fix-layout-solo-column-pane-slips-header-into-adjacent-colum-a17h.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): solo-column pane slips header into adjacent column on minimize instead of going thin
--- a/docs/specs/SPEC_PANE_MINIMIZE_REFINEMENTS_2026_06_24.md
+++ b/docs/specs/SPEC_PANE_MINIMIZE_REFINEMENTS_2026_06_24.md
@@ -1,0 +1,278 @@
+# SPEC — Pane Minimize Refinements
+
+**Date:** 2026-06-24
+**Status:** Proposed
+**Builds on:** SPEC_PANE_MINIMIZE_AND_TOOLCALL_FAILCOLLAPSE_2026_06_21.md
+
+---
+
+## Overview
+
+Two refinements to the pane minimize button introduced in the prior spec:
+
+1. **Caret direction** — the minimize button's chevron should flip to point upward when the pane is collapsed, signaling "click to expand."
+2. **Solo-column slip behavior** — a pane that occupies its own column slot (no vertical sibling) should collapse vertically and slip its header into the adjacent column rather than shrinking horizontally into a thin strip.
+
+---
+
+## Background: current implementation
+
+The minimize button lives in `frontend/app/block/blockframe.tsx` (`OptMinimizeButton`, line 179). It is visible when `numLeafs() > 1` (more than one pane in the workspace) and the pane is not magnified or ephemeral.
+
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

a5af:
Fixed the P1:

**Root cause confirmed:** After slip removes A from the two-pane Row, the Row has one child (B, a Column with children). `balanceNode`'s `beforeWalkCallback` flatMap rule fires on the Row's parent: `child.children?.length == 1 && child.children[0].children` → hoists B's children directly into the grandparent, scattering them and making `targetColumnId` unreachable on restore.

**Fix:** In `_slipMinimize`, after removing A and before `_finishToggle` calls `updateTree`, detect `parentRow.children.length === 1` and hoist B ourselves:
- Walk up to grandparent via `findParent(rootNode, parentRow.id)`
- Set `sibling.size = parentRow.size` (B inherits the Row's Column-slot size)
- Replace the Row with B in grandparent's children array
- Record the Row's restore context in `slipMinimize.promotedRow` (`rowParentId`, `rowIdx`, `rowSize`, `originalTargetRowSize`)

On restore when `promotedRow` is set: find B in grandparent by `targetColumnId`, restore B's original Row-width, set A's size, create a new Row with both children in original order, splice it into the grandparent at the recorded index.

a5af:
Fixed. The `promotedRow` approach is gone.

**Root cause recap:** After slip removes the pane from `parentRow`, the Row has exactly one child (Column B). `balanceNode`'s flatMap rule (`child.children?.length == 1 && child.children[0].children`) would hoist B's children directly into the grandparent — a Column. `balanceNode` then saw two Column children in a Column parent and flipped B's direction to Row, inverting the entire subtree's axis.

**Fix:** Added `_slipAnchor?: true` to `LayoutNode`. `_slipMinimize` sets `parentRow._slipAnchor = true` when the Row is left with a single child, and `balanceNode` now checks `!child._slipAnchor` before applying the hoist rule. The `Row(→Column)` direction alternation is preserved — no hoist, no direction flip. Flag is cleared on restore via `rowNode._slipAnchor = undefined`.

Three files changed (`types.ts`, `layoutNode.ts`, `layoutMinimize.ts`). All 60 layout tests pass.

PR #1759, opened 2026-06-24
feat(tab-chrome): Phase 2+3 folder boundary line and recessed strip

a5af:
## Summary

- **Boundary line (Phase 2):** `border-bottom: 1px solid var(--border-color)` added to `.tab-bar-fill`, `.system-status`, and `.window-drag` on all platforms. Combined with the existing `.tab:not(.active)` rule, the 1px hairline now runs unbroken across the full header width — absent only under the active tab, completing the folder look.
- **Win32+Linux button height:** bumped from `calc(100% + --space-1-5)` → `calc(100% + --space-1-5 + 1px)` so caption-button hover backgrounds still reach the window edge after `.system-status` gains its 1px border.
- **Recessed strip (Phase 3 / Layer A):** `background: var(--tab-strip-bg)` on `.window-header` — the near-transparent white tint (`rgba(255,255,255,0.03)`) makes the strip fractionally lighter than `--workspace-surface`, so the folder content visually advances.
- **Spec updated (§10):** documents online triangulation findings — why `margin-bottom: -1px` cannot work with `overflow-x: auto` scroll containers, VSCode colour-match reference, Win32 DWM border note, and macOS gap caveat (traffic-lights + hamburger use `align-self: center` and don't reach the bottom edge — tracked as follow-up in §10.5).

## Test plan

- [ ] Win32: boundary line visible at the bottom of the header, absent under active tab, present under inactive tabs / fill / system-status area
- [ ] Win32: caption button hover backgrounds (including close-button red) reach the window's bottom edge — no 1px gap
- [ ] Win32: Layer A strip is barely perceptibly lighter than the pane content (near-invisible at window:transparent=false, more noticeable at window:transparent=true)
- [ ] macOS: line present under tab fill and system-status; gap acceptable under traffic lights and hamburger
- [ ] No regressions in tab drag-and-drop, content-aware sizing, or tab switch performance

### DIFF
--- a/.changesets/1782315895-feat-tab-chrome-phase-2-3-folder-boundary-line-and-recessed--lkfc.md
+++ b/.changesets/1782315895-feat-tab-chrome-phase-2-3-folder-boundary-line-and-recessed--lkfc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(tab-chrome): Phase 2+3 folder boundary line and recessed strip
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #1765, opened 2026-06-24
fix(ActivityDock): replace hasCandidates memo with createEffect+setTimeout gate

AgentA-asaf:
## Summary

Follow-up to #1751 — the `hasCandidates` gate that landed in main still subscribes to `tick()` inside the memo itself, so it recomputes every second for the dock's entire lifetime (exited ShellNodes linger in `allActivities()` forever via `shell-adapter.ts`).

This replaces it with a `hasExpiring` signal driven by `createEffect` + `setTimeout`:
- The effect subscribes only to `allActivities()` (no tick dependency)
- It computes the latest retention-window expiry and schedules a timer to flip the gate off
- `onCleanup` cancels the stale timer on re-run
- `visible` calls `tick()` only while `hasExpiring()` is true — once all terminal rows age out, zero tick subscriptions remain

Reagent approved this approach on the PR #1751 branch (`efd0c6cc`) immediately before that PR merged; cherry-picked here against main.

## Test plan
- [ ] Open agent pane, run a shell command to completion → dock row appears, ticks for retention window, then disappears
- [ ] After dock is empty, verify no per-second recompute in browser profiler (no `ActivityDock` in the flame chart)
- [ ] Run multiple shells, let them all expire → gate flips false, tick subscription drops

### DIFF
--- a/.changesets/1782334542-fix-activity-dock-replace-hascandidates-memo-with-createeffect-settimeout-gate.md
+++ b/.changesets/1782334542-fix-activity-dock-replace-hascandidates-memo-with-createeffect-settimeout-gate.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ActivityDock): replace hasCandidates memo with createEffect+setTimeout gate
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #1768, opened 2026-06-24
fix(agent): remove session digest, fix per-turn Haiku summary in pane header

a5af:
## Summary

- **Removes session digest entirely** — `SessionDigestBanner`, `useSessionDigest`, `register_session_digest` (backend), `digest-accessory.ts`, `_session-digest.scss`, `SessionDigestCommand` RPC, and all associated types. Not wanted.
- **Fixes the per-turn Haiku activity summary** so it actually populates the pane header (`term:activity` → `viewText()`).

## Root cause of the bug

`register_session_activity_summary` was reading recent agent output from the WPS ring buffer (`broker.read_event_history(EVENT_BLOCK_FILE, ...)`), but `EVENT_BLOCK_FILE` events have `persist: 0` — they are transient (delivered to live subscribers only, never stored). So `read_event_history` always returned `[]`, the activity window was always empty, Haiku was never invoked, and `term:activity` was never written.

The fix reads from the FileStore directly (same pattern `register_session_digest` already used). The full analysis is in `docs/retro/retro-haiku-activity-pane-header-2026-06-24.md`.

## What now shows where

| Surface | Key | Source |
|---|---|---|
| Pane header (right of "Agent" label) | `term:activity` | `useAgentActivitySummary` → Haiku RPC → `viewText()` |
| Session digest banner | — | **Removed** |


a5af:
@codex review

PR #1769, opened 2026-06-24
fix(browser-pane): black flash + UI freeze on macOS/Linux when opening browser pane

a5af:
## Summary

- **Black flash on open**: `set_visible(1)` was called before `layout()` in `create_browser_pane_view`, making the OverlayController briefly visible at an uncommitted position with its opaque black baseline (`background_color: 0xFF000000`). Fixed by swapping order — `layout()` first, `set_visible(1)` after.
- **UI freeze after open**: `on_after_created_browser_pane` was a complete no-op on macOS/Linux (`let _ = (state, browser)`). On Windows a `WM_SETFOCUS` redirect subclass catches CEF's automatic focus grant to the new pane browser; nothing equivalent existed on macOS/Linux, so the `OverlayController` (`can_activate=1`) held focus and its `BrowserView` intercepted all input for the parent window. Fixed by explicitly calling `set_focus(1)` on the main window browser immediately after `add_overlay_view` returns, and adding a non-Windows `on_after_created_browser_pane` implementation as a safety net for navigation-driven focus re-steals.
- **Borrow-after-move in `creation.rs`**: pre-existing compile error introduced in the latest pull (`initial_view` moved into `promote_pool_window_for_new_window` then used again as `.as_deref()` on the cold path); fixed with `.clone()`.

## Root cause (how we found it)

Diagnosed via live CDP debugging of a frozen 0.47.4 instance:

1. Connected to the renderer on `--remote-debugging-port=9222`
2. DOM inspection confirmed `activeDrag = false` (only 1 element with `pointer-events: none`: the layout drag placeholder — correct when not dragging). The CSS layer was fine; the freeze was native.
3. Identified browser pane block ID (`view: browser`) from the SQLite objects DB
4. Called `browser_pane_close` via the IPC (`POST /ipc`) → freeze lifted immediately, confirming the `OverlayController` was the sole input interceptor

Full write-up: `docs/analysis/ANALYSIS_BROWSER_PANE_BLACK_FREEZE_MACOS_2026_06_24.md`

## Test plan

- [ ] Open a browser pane on macOS — no black flash, no UI freeze
- [ ] Click outside the browser pane immediately after opening — clicks reach the SolidJS UI
- [ ] Navigate to a cross-origin URL in the browser pane — focus returns to main window after redirect (safety net in `on_after_created_browser_pane`)
- [ ] Open a browser pane on Linux — same behaviour
- [ ] Windows behaviour unchanged (Windows path untouched)
- [ ] `cargo check -p agentmux-cef` passes with no errors

### DIFF
--- a/agentmux-cef/src/browser_pane/callbacks.rs
+++ b/agentmux-cef/src/browser_pane/callbacks.rs
@@ -70,7 +70,40 @@ pub fn on_after_created_browser_pane(state: &Arc<AppState>, browser: &Browser) {
     }
     #[cfg(not(target_os = "windows"))]
     {
-        let _ = (state, browser);
+        // On macOS/Linux, the focus redirect subclass used on Windows doesn't
+        // exist. CEF's on_after_created fires from inside add_overlay_view
+        // (when the BrowserView is added to the widget hierarchy) and gives
+        // focus to the new pane browser. creation_views::create_browser_pane_view
+        // immediately returns focus to the main window after add_overlay_view
+        // returns — this callback is an additional safety net for any navigation-
+        // driven focus steals that bypass that initial return (e.g. cross-origin
+        // redirects that recreate the renderer, hitting on_after_created again).
+        //
+        // Only refocus if we can identify the parent window; don't do a blanket
+        // "focus main" that would disrupt multi-window setups.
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #1770, opened 2026-06-24
fix(cef): clone initial_view before move into promote_pool_window_for_new_window

AgentA-asaf:
## Summary

- `initial_view` was moved into `promote_pool_window_for_new_window()` (pool-fast path, line 267) then borrowed via `.as_deref()` at line 282 (cold path), causing `E0382: borrow of moved value`
- Fix: `.clone()` at the call site — the clone goes to the pool path, the original survives for the cold path

Blocked the local build (`task package:local`) — discovered when clearing disk space and rebuilding from scratch.

## Test plan
- [ ] `cargo check -p agentmux-cef` passes (verified clean)
- [ ] "Open in New Window" still works via pool path and cold path


a5af:
@codex review

PR #1772, opened 2026-06-24
feat(tab-chrome): full-width boundary line below tab strip

a5af:
## Summary

- Moves the tab-strip/content separator from per-element `border-bottom` properties (which had gaps at the hamburger button, tab separators, and system-status right margin) to a single `border-bottom` on `.window-header` itself — one rule, zero gaps, all platforms
- Uses `var(--snap-chrome)` for the border width so it snaps to exactly 1 physical pixel inside the header's zoom context at any DPR/zoom factor, eliminating the thinner-border artifact at 125% DPI
- Updates Win32 caption-button height compensation from hardcoded `1px` to `var(--snap-chrome)` to match the border width precisely
- Removes leftover per-element `border-bottom` from `.tab:not(.active)`, `.tab-bar-fill`, `.system-status`, and `tabcontent.tsx` padding workaround

## Test plan

- [ ] Win32: verify a single continuous `--border-color` line appears full-width below the tab strip with no gaps
- [ ] Win32 at 125% DPI: line should be exactly 1 physical pixel (no thinner/thicker inconsistency)
- [ ] Caption button hover backgrounds should still reach the full header height (compensation formula)
- [ ] macOS / Linux: same continuous line via their respective platform SCSS files

### DIFF
--- a/.changesets/1750808000-feat-tab-chrome-boundary-line.md
+++ b/.changesets/1750808000-feat-tab-chrome-boundary-line.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(tab-chrome): full-width boundary line below tab strip via container border-bottom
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #1774, opened 2026-06-25
fix(identity): m0012 — deduplicate identity accounts sharing same credential

a5af:
## Summary

- 101 `claude-oauth` entries were accumulating in `store.db` — one per channel/version launch — all pointing to the same `shared/providers/claude` dir
- Root cause: the old per-channel `backfill_shared_store_once` inserted each channel's unique UUID without deduplication against existing rows with the same credential
- Adds `m0012_dedup_identity_accounts` (Global-scoped, runs once at next launch via the migration framework): groups by `(provider, secret_ref)`, keeps the bound account (or oldest), reroutes bundle bindings + agent links, deletes dupes via `identity_delete` which cascades to joins

## Test plan

- [ ] Verify 0.49.2 launcher log shows `m0012` applied with `deleted=100`
- [ ] Verify `listidentityaccounts` returns 1 claude entry after migration
- [ ] Verify Default bundle binding still points to the correct claude account
- [ ] Verify no migration error log written


a5af:
Closing in favor of #1775 — rebased cleanly onto main

PR #1776, opened 2026-06-25
fix(macos): inject CFBundleIdentifier into dev builds for LaunchServices isolation

a5af:
## Summary

- **Root cause:** Dev builds run as raw binaries without an `Info.plist`, so macOS LaunchServices has no `CFBundleIdentifier` to distinguish them from the packaged stable `.app`. The OS coalesces both under the same Dock tile — clicking the stable `.app` can focus the dev instance instead of launching a new stable window.
- **Fix:** Inject `CFBundleIdentifier = ai.agentmux.<channel>.<version>` at runtime in `set_macos_app_display_name()`, which already mutates the main bundle's mutable info dictionary for `CFBundleName`/`CFBundleDisplayName`. `AGENTMUX_CHANNEL` is always set by the launcher before the host starts.
- **Launcher UX:** When a second `task dev` invocation finds the dev socket already bound, the "already running" message now names the channel and suggests `task dev:local`.
- **Spec:** `docs/specs/SPEC_DEV_ENV_ISOLATION.md` documents the full four-layer isolation stack (data dir, socket hash, bundle ID, pane sentinel).

## Files changed

| File | Change |
|------|--------|
| `agentmux-cef/src/lib.rs` | Add `CFBundleIdentifier` write in `set_macos_app_display_name()` |
| `agentmux-launcher/src/main.rs` | Pass `channel` into `bind_socket_with_recovery`; dev-aware message |
| `docs/specs/SPEC_DEV_ENV_ISOLATION.md` | New spec — isolation stack and known gaps |
| `docs/specs/SPEC_DEV_ENV_ISOLATION.fix-plan.md` | Fix rationale and implementation notes |

## Test plan

- [ ] Run `task dev:local` while stable `.app` is running
- [ ] Double-click the stable `.app` — a new stable window opens, dev instance unaffected
- [ ] Run `task dev` a second time while first dev session is active — message reads "AgentMux dev instance already running … Use `task dev:local`"
- [ ] Verify `AGENTMUX_CHANNEL` in the dev instance matches the injected `CFBundleIdentifier` suffix

### DIFF
--- a/agentmux-cef/src/lib.rs
+++ b/agentmux-cef/src/lib.rs
@@ -1370,7 +1370,17 @@ unsafe fn set_macos_app_display_name() {
                 // Fall back to "dev" only for unmanaged direct invocations.
                 let channel = std::env::var("AGENTMUX_CHANNEL")
                     .unwrap_or_else(|_| "dev".to_string());
-                let bundle_id = format!("ai.agentmux.{}.{}", channel, env!("CARGO_PKG_VERSION"));
+                // Sanitize channel to match the normalization applied by
+                // scripts/package-macos.sh (lines 67-68):
+                //   tr -C 'A-Za-z0-9.-' '-' | tr '[:upper:]' '[:lower:]'
+                // Replace any char not in [A-Za-z0-9.-] with '-', then lowercase,
+                // so runtime and build-time CFBundleIdentifiers always match.
+                let channel_sanitized: String = channel
+                    .chars()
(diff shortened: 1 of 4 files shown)


a5af:
Fixed: apply same CFBundleIdentifier sanitization as `package-macos.sh` (replace non-[A-Za-z0-9.-] with -, lowercase) so runtime and build-time IDs always match.

PR #1778, opened 2026-06-25
fix(browser-pane/macos): fix black screen by deferring overlay bounds commit

a5af:
## Summary

- **Black pane rendered, page loads** — the primary bug is fixed. Opening a browser pane on macOS no longer shows a blank black rectangle.
- **Scroll works in pane** — scroll-wheel events reach the Chromium renderer correctly.
- **Sidebar stays usable** — main window regains key focus after pane setup, so the SolidJS sidebar remains clickable.
- **Hidden-pane creation handled** — panes created while hidden (inactive tab / modal overlay-clip) now have their NSWindow frame committed at creation time so they display correctly when later shown.
- **Splitter-drag resize works** — `resize_browser_pane_view` now posts an ObjC setFrame task so post-creation resizes update the NSWindow frame (previously only `set_size`/`set_position` were called, which are no-ops on macOS).

## Root Cause

`overlay_view_host.cc::SetOverlayBounds()` does:
```cpp
bounds_.Intersect(window_view_->bounds());
```
At pane creation time, the parent `CefWindowView` has empty (0×0) bounds on macOS because the native layout is deferred. The intersection with 0×0 always produces 0×0 — so every `setFrame:display:YES` on the `NativeWidgetMacNSWindow` fired `windowDidResize:` → CEF layout → reset to 0×0 → black screen.

Additional macOS issues:
- `CefOverlayController::set_size`/`set_position`/`set_bounds` are no-ops for the overlay's `NativeWidgetMacNSWindow` frame; the only working path is ObjC `[NSWindow setFrame:display:YES]`.
- Hidden panes (created while `compute_pane_visible=false`) had their `SetPaneBoundsViewsTask` return early without committing the frame, leaving the NSWindow unpositioned for when the pane was later shown.

## Fix

Call `controller.set_bounds()` from the **deferred UI-thread task** (`SetPaneBoundsViewsTask`), after `set_visible(1)`. By then the parent `CefWindowView` has non-empty bounds (1200×800 DIP), so Intersect preserves the desired pane rect.

Additional changes:
- **macOS 26 Tahoe compat**: `[NSWindow windowWithWindowNumber:]` removed — replaced with a pre-`set_visible(0)` scan of `[NSApp windows]` to capture the overlay `windowNumber` before it is hidden.
- **ObjC `setFrame:display:YES`** on the `NativeWidgetMacNSWindow` to sync the NSWindow frame in addition to CEF Views bounds.
- **`makeKeyAndOrderFront:`** on main window after bounds commit to restore sidebar click routing (sidebar-first policy).
- **Hidden-pane fix**: `SetPaneBoundsViewsTask` now always calls `set_visible(1)` before scanning (forces NSWindow creation and ensures it appears in `[NSApp windows]`), commits the frame, then restores visibility via `set_visible(0)` if `compute_pane_visible=false`. Discovered wnum stored in `AppState::browser_pane_overlay_wnums`.
- **Resize fix**: `resize_browser_pane_view` posts a `SetPaneBoundsViewsTask` (retry=2, physical-pixel path, wnum match) so splitter-drag resizes actually move the NSWindow frame.

## Known Limitation — Pane Mouse Clicks

Mouse button clicks (left/right) inside the browser pane are still unresponsive. Scroll works because `NSEventTypeScrollWheel` is routed by cursor position, independent of key-window status. Mouse button events require the window to be KEY or the content view to return `acceptsFirstMouse:YES`. Full investigation documented in `docs/analysis/RETRO_BROWSER_PANE_MACOS_FIX_2026_06_26.md`.

## Test Plan

- [ ] Open agentmux, open a browser pane — pane should render the web page (not black)
- [ ] Scroll within the pane — scroll events should reach the page renderer
- [ ] Click in the sidebar (SolidJS UI) — should still respond normally
- [ ] Open, close, re-open pane — no crash, pane renders on each open
- [ ] Open pane while another tab is active (pane hidden) — switch to tab, pane should render at correct position
- [ ] Drag a splitter to resize the pane — pane should resize (NSWindow frame moves)

### DIFF
--- a/.changesets/1782432000-fix-browser-pane-black-screen-freeze-macos-objc-nswindow-resize.md
+++ b/.changesets/1782432000-fix-browser-pane-black-screen-freeze-macos-objc-nswindow-resize.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(browser-pane): black screen + mouse freeze on macOS when opening a browser pane
--- a/agentmux-cef/src/browser_pane/creation_views.rs
+++ b/agentmux-cef/src/browser_pane/creation_views.rs
@@ -366,6 +366,29 @@ pub fn resize_browser_pane_view(state: &Arc<AppState>, label: &str, rect: Rect) 
     if let Some(window) = state.windows.lock().get(&window_label).cloned() {
         window.layout();
     }
+
+    // macOS: set_size/set_position/layout() are no-ops for the overlay's
+    // NativeWidgetMacNSWindow frame — the only working path is ObjC
+    // [NSWindow setFrame:display:YES].  Post a bounds task (retry=2 →
+    // physical-pixel path, wnum match, no delta-detection pre-scan) so the
+    // NSWindow frame is updated on this UI-thread tick.  retry=2 is used
+    // rather than retry=0 to skip the delta-detection pre-scan (the window
+    // already exists) and rather than retry=1 to avoid the
+    // controller.bounds()-as-DIP code path (we have fresh physical pixels).
+    #[cfg(target_os = "macos")]
+    {
+        let wnum = state.browser_pane_overlay_wnums.lock()
+            .get(label).cloned().unwrap_or(0);
+        if wnum > 0 {
+            crate::ui_tasks::post_set_pane_bounds_views(
+                state, label, &window_label,
+                rect.x, rect.y, rect.width, rect.height,
+                2,    // retry=2: physical pixels, wnum match, no pre-scan
+                wnum,
+            );
+        }
+    }
+
     tracing::debug!(
         label = %label, window_label = %window_label,
         x = rect.x, y = rect.y, w = rect.width, h = rect.height,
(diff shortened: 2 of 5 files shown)


a5af:
Addressed P1/P2 review feedback:

- **P1**: `controller.set_visible(1)` on macOS is now inside the ObjC block, called only after frame is committed (guarded by overlay NSWindow found check). On Linux it remains in the `#[cfg(not(target_os = "macos"))]` early-return path.
- **P1**: Added changeset file `.changesets/1782432000-fix-browser-pane-black-screen-freeze-macos-objc-nswindow-resize.md`.
- **P2**: UI-thread FIFO ordering invariant now documented in code comment.
- **P2**: `post_set_pane_bounds_views` call in `creation_views.rs` is already cfg-gated (module is non-Windows only per `mod.rs`); added inline comment noting this.

Root cause was fully identified and fixed: `CefOverlayController::SetSize/SetPosition` are permanent no-ops on macOS. The overlay is a `NativeWidgetMacNSWindow` (CEF popup NSWindow). We now resize it directly via `[overlayWindow setFrame:display:YES]` with physical-px → points conversion via `backingScaleFactor`.

PR #1782, opened 2026-06-25
fix(term): eliminate 1Hz typing jerk from sysinfo blocking Tokio workers

AgentU-asaf:
## Root cause

`sysinfo.rs` called `refresh_processes_specifics(ProcessesToUpdate::All)` synchronously inside an async Tokio task. On Linux with many CEF renderer/helper processes, the `/proc` scan takes 5–20ms and occupies the Tokio worker thread for the full duration. Since the WebSocket egress loop (which sends terminal echo) runs on the same thread pool, it gets starved at exactly 1Hz — producing the noticeable jerk when holding a key in a terminal.

This is the regression of the fix documented in `SPEC_TERMINAL_INPUT_PRIORITY_OVER_SYSINFO_2026_06_16.md`.

### `agentmux-srv/src/backend/sysinfo.rs` — block_in_place (primary fix)
All three blocking `/proc` refresh calls wrapped in `tokio::task::block_in_place()`. The runtime spawns an extra worker thread to handle async work (echo egress, RPC) while the current thread is occupied by the scan. Echo never waits for sysinfo anymore.

### `agentmux-srv/src/server/websocket.rs` — two-lane egress + coalescing
- Priority lane (terminal echo, RPC replies) and background lane (sysinfo, blockstats) — `biased;` select! always drains priority before touching background
- Background lane coalesces to latest-only per `event:scope` key, dropping stale intermediate telemetry that would pile up behind the priority lane

### `frontend/app/store/wps.ts` — defer sysinfo dispatch
Sysinfo/blockstats events deferred via `setTimeout(0)` to a new macrotask, so the current frame's xterm.js canvas render completes before sysinfo DOM updates land.

### `agentmux-launcher/src/splash_info.rs` — DEV label style
Shows `v0.49.2 DEV` (uppercase, no parens) to match the status bar's existing DEV badge style.

## Verification

Tested: hold backspace on a long terminal line with sysinfo polling active. No visible 1Hz jerk. Sysinfo gauges continue updating normally.


AgentU-asaf (commented):
Fixed — updated the doc comment on line 33 to describe the actual format (`v<version> DEV`, uppercase, no parens).

PR #1801, opened 2026-06-27
fix(statusbar): use full viewport boundary for CPU/token popovers on Windows

AgentA-asaf:
## Summary

- `CpuCoresPopover` and `TokenBreakdownPopover` were calling `computeMenuPosition` with the default `avoidNativePanes: true`
- When a browser pane fills the workspace, `getPaintableArea()` resolves `largestFreeRect` to the DOM title-bar strip (33px) — larger than the 22px status-bar strip below the pane
- Floating-ui's `shift` middleware pins the `"top-end"` popover to that boundary, placing it near the top of the window instead of above the status bar button
- Both popovers already call `usePaneOverlay` which cuts the airspace transparency hole — `avoidNativePanes` is redundant and harmful here
- Fix: pass `avoidNativePanes: false` so floating-ui uses the full viewport as boundary

Root-cause analysis and full test matrix: `docs/specs/SPEC_STATUSBAR_POPOVER_POSITION_WIN10_2026_06_26.md`

## Files changed

| File | Change |
|------|--------|
| `frontend/app/statusbar/CpuCoresPopover.tsx` | `avoidNativePanes: false` |
| `frontend/app/statusbar/TokenBreakdownPopover.tsx` | `avoidNativePanes: false` |
| `docs/specs/SPEC_STATUSBAR_POPOVER_POSITION_WIN10_2026_06_26.md` | Root-cause spec |

## Test plan

- [ ] Windows 10 with a browser pane open — click CPU stat — popover appears directly above button
- [ ] Windows 11 with a browser pane open — same verification
- [ ] macOS — no regression (already worked)
- [ ] No browser pane open (agent/terminal panes only) — popover still correct
- [ ] Token breakdown popover — same check on Windows with browser pane open
- [ ] Chrome zoom at non-1.0 — popover still correctly anchored

### DIFF
--- a/.changesets/1782537051-fix-statusbar-popover-position-windows-avoidnativepanes-false-8kzm.md
+++ b/.changesets/1782537051-fix-statusbar-popover-position-windows-avoidnativepanes-false-8kzm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(statusbar): status-bar popovers (CPU, token breakdown) now position correctly when a browser pane is open on Windows
--- a/docs/specs/SPEC_STATUSBAR_POPOVER_POSITION_WIN10_2026_06_26.md
+++ b/docs/specs/SPEC_STATUSBAR_POPOVER_POSITION_WIN10_2026_06_26.md
@@ -1,0 +1,175 @@
+# SPEC: Status-Bar Popover Position Bug (Windows — browser pane open)
+
+**Date:** 2026-06-26  
+**Status:** Root-cause confirmed, fix identified  
+**Reported:** Windows 10 — CPU stat popover appears in the middle of the screen instead of above the status bar. Works correctly on macOS and Windows 11 (without a browser pane open).  
+**Affects:** `CpuCoresPopover`, `TokenBreakdownPopover` — any status-bar popover using `computeMenuPosition` with default `avoidNativePanes: true`.
+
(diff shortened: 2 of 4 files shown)


AgentA-asaf:
**P1 items — pre-existing in `main`, not introduced by this PR:**

`git diff main...fix/statusbar-popover-position-win10 --name-only` shows only 3 files (the two popovers + spec). `ZOOM_MIRROR_GEN` and the 300ms debounce at `service.rs:2823-2828` exist in `main` before this branch was cut — confirmed with `git show main:agentmux-srv/src/server/service.rs | sed -n '2818,2835p'`. The missing changeset `1782529685-fix-agent-zoom-persists-immediately-drop-300ms-debounce-that-fse0.md` is also a pre-existing gap in `main`. Neither was touched by this PR. Will track in a separate issue.

**P2 items addressed in latest push (fb402dc2):**

- Removed `assertMenuInPaintableArea` calls from both `CpuCoresPopover` and `TokenBreakdownPopover` — these popovers use `usePaneOverlay` for airspace transparency, so the guard's native-pane intersection check would always fire a false-positive `[menu-guard]` warning when a browser pane is open.
- Updated Evidence Trail table in the spec to remove the "missing `avoidNativePanes: false`" annotation (fix is applied in this PR).

PR #1803, opened 2026-06-27
fix(window-count): detect OS-level window hides, emit report_window_closed

AgentA-asaf:
## Summary

**Gap A fix** from `docs/retro/retro-window-count-stale-post-1701-2026-06-27.md`:

Alt+F4, taskbar "Close window", and any non-RPC close path bypass `close_window()`, where PR #1701 Part 1 added `report_window_closed`. Because CEF Views recycles windows on close (`on_before_close` never fires), those paths leave the launcher's instance list permanently stale — the status bar count never decrements.

**Changes:**
- `agentmux-cef/src/state.rs`: new `label_for_hwnd(HWND) -> Option<String>` reverse lookup on `window_hwnds` (cfg windows only). O(n) over open window count (typically 1–5).
- `agentmux-cef/src/wrr/win_event.rs`: `EVENT_OBJECT_HIDE` arm — after the existing `report_hwnd_visibility_changed` call, checks if the hiding window is on-screen (left ≥ `OFFSCREEN_POOL_THRESHOLD_X`), non-minimised, and has a registered label that isn't `browser-pane-*`. If all conditions hold, calls `report_window_closed(label)`. The launcher silently no-ops duplicate reports.

## Test plan
- [ ] Open 2 windows, close one via Alt+F4 — status bar should drop to `(1)` immediately
- [ ] Close via taskbar right-click "Close window" — same
- [ ] Normal ✕ close still works (double-report from RPC + HIDE is a launcher no-op)
- [ ] Pool windows (HWND left < −20000) do not trigger false reports

### DIFF
--- a/.changesets/1782535243-fix-window-count-report-os-level-window-closes-via-win-event-6i64.md
+++ b/.changesets/1782535243-fix-window-count-report-os-level-window-closes-via-win-event-6i64.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(window-count): report OS-level window closes via win_event HIDE hook
--- a/.changesets/1782535243-fix-window-count-report-os-level-window-closes-via-win-event-6i64.md
+++ b/.changesets/1782535243-fix-window-count-report-os-level-window-closes-via-win-event-6i64.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-fix(window-count): report OS-level window closes via win_event HIDE hook
+fix(window-count): detect OS-level window closes via LOCATIONCHANGE pool-move
--- a/agentmux-cef/src/state.rs
+++ b/agentmux-cef/src/state.rs
@@ -1055,6 +1055,22 @@ impl AppState {
         crate::reducer::count_live_user_windows(&self.host_state.lock())
     }
 
+    /// Reverse lookup: find the label whose cached HWND matches `hwnd`.
+    /// Used by the win-event HIDE handler to map an OS-level window hide to a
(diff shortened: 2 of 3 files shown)


AgentA-asaf:
Thanks for catching this. The fix (commit a96e080a) debounces the `report_window_closed` call by 200ms:

- On `EVENT_OBJECT_HIDE`: instead of calling `report_window_closed` immediately, a cancel flag is stored in a per-label `pending_os_closes` map and a short-lived thread sleeps 200ms.
- On `EVENT_OBJECT_SHOW`: if the same label has a pending entry, the flag is set to cancelled — the deferred thread no-ops. Net effect for virtual-desktop switches (HIDE → SHOW within ~50ms): zero reports.
- Real closes (Alt+F4, taskbar): no paired SHOW arrives, so the thread fires after 200ms.

AgentA-asaf:
Replaced the debounce approach entirely (commit b061f02e). The HIDE-based detection is gone.

**Root insight**: CEF recycle-on-close moves the HWND to `x < -20000` via `WM_WINDOWPOSCHANGED` → `EVENT_OBJECT_LOCATIONCHANGE`. Virtual-desktop switches do NOT move the rect — the window stays at its on-screen coords; only visibility changes. So pool-position detection on `LOCATIONCHANGE` distinguishes a real close from a desktop switch with no debounce, no paired-SHOW cancellation, and no thread spawning.

The pool-move check runs before the position debounce so a close following a drag within 50ms is still caught. `WM_WINDOWPOSCHANGED` only fires when the position changes, so a stationary pool window never re-fires — no duplicate-report risk.

Net diff: −104 / +39 lines.

PR #1818, opened 2026-06-27
feat(app-api): implement identity, preset, and memory namespaces

a5af:
## Summary

Implements the `identity.*`, `preset.*`, and `memory.*` App API namespaces per `SPEC_AGENT_APP_API_IDENTITY_PRESETS_BRAIN_2026_06_27.md` (spec also included on this branch).

### What's in this PR

- **Spec** (`specs/SPEC_AGENT_APP_API_IDENTITY_PRESETS_BRAIN_2026_06_27.md`) — full design doc for all three namespaces
- **12 RPC handlers** in `app_api.rs` — identity.self.accounts, identity.account.upsert, identity.account.validate, identity.self.unlink, preset.list/get/upsert/delete/self.get, memory.list/read/write
- **S1 auth gate** on all identity and memory handlers — `ctx.agent_id` non-empty + matches `req.agent_id`
- **Keychain + optional probe** — `identity.account.upsert` writes to OS keychain first, optionally calls `key_validator::validate`, then DB-upserts with compensation on failure for new accounts
- **S4a field strip** — `preset.upsert` strips caller-supplied `is_global` / `sort_order`
- **Seed guard** — `preset.delete` catches `StoreError::Other("seed...")` → FORBIDDEN
- **Blank-singleton two-step** — `preset.self.get` lists all bundles to find `is_blank=true` id, then fetches full object
- **CLAUDE_CONFIG_DIR fix** — `native_memory_handlers.rs` now resolves the memory root from the agent's stored env blob instead of hardcoding `~/.claude/projects`
- **Atomic write** — `memory.write` writes to a `.tmp` then renames, fires `agent:memory:changed:<agent_id>`
- **WPS events** — upsert/unlink/write each publish the right wave events

### Files changed

| File | Change |
|------|--------|
| `rpc_types.rs` | Added `agent_id: String` to `RpcContext`; 12 `COMMAND_*` constants |
| `websocket.rs` | Stamp `agent_id` into `RpcContext` on `bus:register` |
| `native_memory_handlers.rs` | Resolve `CLAUDE_CONFIG_DIR` from agent env blob; pub helpers |
| `app_api.rs` | 12 handler registration functions + `check_s1` helper (~650 LOC) |

## Test plan

- [ ] `cargo build -p agentmux-srv` — 0 errors ✅
- [ ] `identity.self.accounts` returns empty list for agent with no linked provider
- [ ] `identity.account.upsert` stores key in keychain, links provider, fires WPS events
- [ ] `identity.account.validate` probes a stored or ad-hoc key
- [ ] `identity.self.unlink` removes link, fires event only if a link existed
- [ ] `preset.list` returns summary-only fields (no instructions blob)
- [ ] `preset.get` by id and by name (most-recent-updated tie-break)
- [ ] `preset.upsert` rejects blank preset, strips `is_global`/`sort_order`
- [ ] `preset.delete` rejects blank/seed bundles with FORBIDDEN
- [ ] `preset.self.get` falls back to blank singleton when instance has no `memory_id`
- [ ] `memory.list/read/write` use `CLAUDE_CONFIG_DIR` from agent env (not `~/.claude`)
- [ ] S1 check rejects empty `agent_id` or mismatched caller

### DIFF
--- a/.changesets/1782635266-feat-app-api-implement-identity-preset-and-memory-namespaces-4ibc.md
+++ b/.changesets/1782635266-feat-app-api-implement-identity-preset-and-memory-namespaces-4ibc.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(app-api): implement identity, preset, and memory namespaces for agents
--- a/agentmux-srv/src/backend/rpc_types.rs
+++ b/agentmux-srv/src/backend/rpc_types.rs
@@ -384,6 +384,20 @@ pub const COMMAND_BLOCKFILE_READ_RANGE: &str = "blockfile:read_range";
 pub const COMMAND_BLOCKFILE_READ_STATE: &str = "blockfile:read_state";
 pub const COMMAND_BLOCKFILE_WRITE_STATE: &str = "blockfile:write_state";
 
+// App API — identity/preset/memory namespaces
+pub const COMMAND_IDENTITY_SELF_ACCOUNTS: &str = "identity.self.accounts";
+pub const COMMAND_IDENTITY_ACCOUNT_UPSERT: &str = "identity.account.upsert";
+pub const COMMAND_IDENTITY_ACCOUNT_VALIDATE: &str = "identity.account.validate";
+pub const COMMAND_IDENTITY_SELF_UNLINK: &str = "identity.self.unlink";
+pub const COMMAND_PRESET_LIST: &str = "preset.list";
+pub const COMMAND_PRESET_GET: &str = "preset.get";
+pub const COMMAND_PRESET_UPSERT: &str = "preset.upsert";
+pub const COMMAND_PRESET_DELETE: &str = "preset.delete";
+pub const COMMAND_PRESET_SELF_GET: &str = "preset.self.get";
+pub const COMMAND_MEMORY_LIST: &str = "memory.list";
+pub const COMMAND_MEMORY_READ: &str = "memory.read";
+pub const COMMAND_MEMORY_WRITE: &str = "memory.write";
+
 // App API Tier 1 — session archival commands
 pub const COMMAND_SESSION_ARCHIVE: &str = "session:archive";
 pub const COMMAND_SESSION_RESTORE: &str = "session:restore";
@@ -1351,6 +1365,11 @@ pub struct RpcContext {
     pub tabid: String,
     #[serde(default, skip_serializing_if = "String::is_empty")]
     pub conn: String,
+    /// Slug of the authenticated agent from `bus:register`. Empty for non-agent
+    /// connections (plain WebSocket, CEF UI). App API handlers use this for S1
+    /// enforcement — reject if empty (unauthenticated), reject if ≠ request agent_id.
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub agent_id: String,
 }
 
 /// Matches Go's `CommandVarData`
--- a/agentmux-srv/src/server/app_api.rs
+++ b/agentmux-srv/src/server/app_api.rs
@@ -3189,11 +3189,15 @@ fn register_identity_account_upsert(engine: &Arc<WshRpcEngine>, state: &AppState
                 // Steps 2+4: replace provider link (unlink old, link new).
                 let _ = id_store.agent_identity_unlink(&req.agent_id, &req.provider);
                 if let Err(e) = id_store.agent_identity_link(&req.agent_id, &account_id, &req.provider) {
-                    // Unlink already ran — keychain secret is now orphaned. Best-effort delete.
-                    let aid = account_id.clone();
-                    let _ = tokio::task::spawn_blocking(move || {
-                        crate::identity::secret_store::delete(&aid)
-                    }).await;
+                    // Only clean up keychain for new accounts — on the update path the
+                    // account still exists in the DB and may be linked to other providers,
+                    // so deleting the keychain secret would destroy a valid credential.
+                    if is_new {
+                        let aid = account_id.clone();
+                        let _ = tokio::task::spawn_blocking(move || {
+                            crate::identity::secret_store::delete(&aid)
+                        }).await;
+                    }
                     return Err(format!("identity.account.upsert: link: {e}"));
                 }
 
--- a/agentmux-srv/src/server/app_api.rs
+++ b/agentmux-srv/src/server/app_api.rs
@@ -3232,12 +3232,13 @@ fn register_identity_account_validate(engine: &Arc<WshRpcEngine>, state: &AppSta
     let id_store = state.id_store.clone();
     engine.register_handler(
         COMMAND_IDENTITY_ACCOUNT_VALIDATE,
-        Box::new(move |data, _ctx| {
+        Box::new(move |data, ctx| {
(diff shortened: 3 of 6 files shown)


AgentA-asaf:
Addressed all review findings (55bf70a9):

**P1 — S5 contradiction (§8):** Reworded — "never writes to keychain or DB" dropped. S5 now reads: may update `status` column on existing row only (when `account_id` supplied + probe resolves `unknown` status); never creates/deletes rows, never writes keychain.

**P1 — `identity.account.upsert` link-replace atomicity (§5.1, §7):** `linkagentidentity` inserts, it does not replace. Documented explicit sequence: `account.key.verify` → `unlinkagentidentity` (if existing link for provider) → `linkagentidentity`. §7 "New logic" column updated to match.

**P2 — `preset.get { name }` tie-breaking (§5.2):** Preset names are not unique. Added: return most-recent `updated_at` on collision; return not-found error on zero matches.

**P2 — `preset.self.get` null case (§5.2):** Returns the blank singleton object (not null) when `memory_id` is unset — via `listmemories` filter `is_blank: true`.

**P2 — `memory.write` event timing (§5.3):** Event fires only after successful write. No event on error.

**P2 — `memory.list` `metadata_type` (§5.3):** Pinned to `metadata.type` YAML frontmatter key.

**Minor — `rpc_types.rs` path (§3 OQ1, §9):** Corrected to `agentmux-srv/src/backend/rpc_types.rs`.

**Minor — Error format (§8):** Added: FORBIDDEN errors use the existing string convention (`"FORBIDDEN: <reason>"`), no new error type.

**Minor — §10 phasing:** P0 (`RpcContext` extension) can ship in the same PR as P1 — no need for a bare prereq commit.

AgentA-asaf:
Addressed all round-2 review findings (1ddc0ca1):

**P1 — S1 empty-string bypass (§8):** Added two-step check: (1) reject if `ctx.agent_id` is empty (`FORBIDDEN: unauthenticated agent connection`), *then* (2) reject if `ctx.agent_id != req.agent_id`. Equality-only check was insufficient — `ctx=""` + `req=""` passed it, letting unauthenticated connections call agent-scoped commands.

**P1 — `identity.account.upsert` keychain compensation (§5.1):** Added compensation path: if `linkagentidentity` (step 3) fails after the keychain write (step 1), handler must attempt to delete the orphaned keychain entry. If that also fails, log + return error to caller. Caller retry re-enters step 1 and overwrites the orphaned entry — same best-effort-cleanup pattern used elsewhere in `identity_handlers.rs`.

**P1 — `preset.upsert` `is_global`/`sort_order` strip (§5.2, §7, S4a):** New S4a invariant: handler must strip `is_global` and `sort_order` before delegating to `upsertmemory`. Confirmed: `bundle_memory_upsert` at `memory_bundles.rs:156` does `is_global = excluded.is_global` — caller-supplied `true` would inject instructions into every agent's context at launch.

**P1 — WPS `persist` values (§6):** Corrected to `persist: 0` for all identity and preset events. Confirmed: every event in `agent_handlers.rs` uses `persist: 0`. Added re-poll guidance for reconnecting subscribers.

**P1 — `preset.self.get` slug→UUID lookup (§5.2, §7):** `CommandGetAgentInstanceData` takes `id: String` (UUID). Documented correct path: `instance_get_by_name(agent_id)` maps the agent slug → `AgentInstance` → `memory_id` → `getmemory`. `getagentinstance` cannot be called directly with a slug.

**P1 — `preset.delete` `seed-` guard (§5.2, §7, S4):** Confirmed `bundle_memory_delete` at `memory_bundles.rs:194` rejects `id.starts_with("seed-")` with `StoreError::Other`. Handler must catch and re-surface as `FORBIDDEN: cannot delete a seeded bundle`, not a raw storage error string. S4 updated to cover both blank singleton and seeded bundles.

**P2 — `mcp_servers` type (§5.2):** Fixed example from `{}` to `[]` — `Memory.mcp_servers` is a JSON-encoded array string.

a5af:
Addressed latest review findings (5bbe0f48):

**P1 — `preset.upsert` S4 bypass via id (app_api.rs:3383):** The `memory.is_blank` flag check was bypassable — sending `{ "id": "blank", "is_blank": false }` passes it since `is_blank` defaults false. Fixed: guard now checks `memory.id == "blank" || memory.id.starts_with("seed-")` first, then `memory.is_blank` as a belt-and-suspenders fallback. The id check fires before any mutation, regardless of caller-supplied flags.

**P1 — `identity.account.upsert` no ownership check (app_api.rs:3099):** When a caller supplies an existing `account_id` (update path), the handler now verifies that id is linked to the calling agent via `agent_identity_list_for_agent(req.agent_id)` before writing to keychain or DB. Unknown or foreign `account_id` → `FORBIDDEN: account not linked to this agent`.

**P1 — Version downgrade:** Rebased onto latest main (0.49.7).

a5af:
Addressed latest P2 findings (35f61add):

**P2 — `preset.upsert` demotes/corrupts existing global presets (app_api.rs:3402):** Before upserting a non-empty id, handler now fetches the existing record. If `is_global=true`, returns `FORBIDDEN: cannot mutate a global preset`. Agents can only upsert presets they created (no existing row, or existing row with `is_global=false`).

**P2 — `identity.account.upsert` link failure orphans keychain (app_api.rs:3192):** If `agent_identity_link` fails after `agent_identity_unlink` already ran, handler now attempts best-effort `secret_store::delete` before returning the error — matching the §5.1 compensation contract.

a5af:
Addressed P1 (ac7e3296):

**P1 — `memory_dir_for_agent` slug vs UUID mismatch (native_memory_handlers.rs:88):** `agent_def_get(agent_id)` queries `db_agents WHERE id = ?` (UUID primary key), but `req.agent_id` arriving from the App API is the agent slug (`AGENTMUX_AGENT_ID` / `bus:register` identity). Slug ≠ UUID, so every `memory.*` call returned "agent not found".

Fixed: replaced `agent_def_get(slug)` with `instance_get_by_name(slug)` (queries `WHERE instance_name = ?`). The returned `AgentInstance` has `working_directory` directly and `id` (UUID) for the subsequent `agent_content_get("env")` lookup.

a5af:
Addressed P1 (a7af6eaa):

**P1 — link-failure keychain delete not gated on `is_new` (app_api.rs:3194):** On the update path (`is_new=false`) the account row was already successfully upserted and the secret may still be linked to other providers. Unconditional `secret_store::delete` on `agent_identity_link` failure would destroy a live credential while leaving the DB row intact.

Fixed: keychain delete on link failure is now gated on `is_new`, matching the DB-failure compensation block a few lines above.

a5af:
Addressed P2 (d34545e1):

**P2 — `identity.account.validate` no ownership check on stored-account path (app_api.rs:3249):** Handler was using `_ctx` and had no `agent_id` field in the request, letting any connection probe another agent's credential by supplying its UUID.

Fixed for the stored-account path (`account_id` supplied): added `agent_id` to the request struct, enforce S1 (`ctx.agent_id` non-empty + matches `req.agent_id`), then verify the account is linked to the calling agent via `agent_identity_list_for_agent` before reading from keychain. The ad-hoc path (`provider + secret`, no stored account) is unaffected — the caller supplies their own secret.

PR #1852, opened 2026-06-30
feat(app-api): IdentityLink MCP tool — bind owned account into default bundle

AgentY-asaf:
## Summary

Second-wave identity write tool, building on #1840 (`feat(app-api): MCP bindings`). **Merge after #1840.**

**The gap**: `identity.account.upsert` links an account via `db_agent_identity_links` (legacy), which `IdentityAccounts` reads. But the resolver at spawn reads `db_identity_bundles`/`db_identity_bindings`. An agent that owns an account via the App API never gets it injected at launch.

**The fix**: `IdentityLink` bridges the two systems — given an `account_id` (from `IdentityAccounts`) and a `provider`, it:
1. Verifies ownership (S1 — account must be in the agent's legacy links)
2. Finds or creates `bundle-<agent_id>-default`
3. Calls `bundle_identity_bind` so the resolver picks it up at next launch

**No secret in the tool call or transcript** — only `account_id` + `provider`. Closes the §7.2 concern from the spec.

### Files changed
- `agentmux-mcp/src/main.rs` — `IDENTITY_LINK_TOOL` schema + match arm → `POST /api/v1/agent/identity/link`
- `agentmux-srv/src/server/mod.rs` — route + `handle_agent_identity_link`
- `agentmux-srv/src/server/app_api.rs` — `identity_bundle_link_impl`

`cargo build -p agentmux-srv -p agentmux-mcp` — clean.

### DIFF
--- a/.changesets/1782858157-feat-app-api-identity-link-mcp-tool-b4a2.md
+++ b/.changesets/1782858157-feat-app-api-identity-link-mcp-tool-b4a2.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(app-api): IdentityLink MCP tool — bind owned account into default bundle
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -368,6 +368,19 @@ const IDENTITY_VALIDATE_TOOL: &str = r#"{
   }
 }"#;
 
+const IDENTITY_LINK_TOOL: &str = r#"{
+  "name": "IdentityLink",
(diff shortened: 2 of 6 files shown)


AgentY-asaf:
Closing — superseded by the composable agent model (#1861). The IdentityLink bridge targets  → , both of which are removed in the new model (Account becomes a direct primitive, Bundle is the optional collection). The resolver gap this addressed will be closed by the new Account binding path instead.

PR #1875, opened 2026-07-01
fix(floating-pane): relabel promoted pane-pool windows to floating-* (status-bar count)

a5af:
## ⚠️ Do NOT merge until runtime-verified on Windows
The highest-risk piece (the promoted floater's renderer label consistency) is a **runtime** property. See "Verification" below.

## Bug
On Windows, a user floating pane opened via the pane-pool fast path keeps its warm-pool label `floating-pool-<uuid>` after promotion, so the status-bar InstancePanel counter (`isFloatingPaneLabel`, which excludes `floating-pool-*`) reports **0 floating panes** while one is visible. Cosmetic; pre-existing (not a regression).

## Fix — Option A (rename on promotion)
Full scoping in `SPEC_FLOATING_PANE_POOL_RELABEL_2026_06_30.md`. On promotion, rename `floating-pool-<uuid>` → `floating-<uuid>` so the pane is counted like any cold-path floater. (The count is fed by the `list_window_instances` snapshot, which already returns promoted panes — just under the excluded label — so the rename alone is sufficient.)

- **New pure reducer command** `HostCommand::RelabelBrowser` (`browsers::handle_relabel_browser`): re-keys `browsers` (+ the duplicated `BrowserHandle.label`) and, when present, `window_opacities` / `pane_window_states`; errors on missing-old / existing-new. Added to the Debug impl + `is_quit_relevant` exclusion.
- **`promote_pane_pool_window` (Windows)**: re-keys the AppState maps `window_hwnds` / `window_meta` under lock, then registers the floater hook and emits `pool:pane-promote` **under the new label** (the emit resolves via `get_browser`, so it must run after the re-key), and returns the new label.
- **`pool:pane-promote` payload** now carries `windowLabel`; the renderer's `awaitPanePoolPromote` rewrites `?windowLabel=` so every label-addressed IPC targets the host's current key — the subtlest loose end.

## Scope
- **Windows path only** (the repro + where it's runtime-verifiable). Non-Windows promote is annotated as a follow-up.
- **Launcher / srv untouched** — neither stores floater labels (verified).
- **Counter unchanged** — already accepts `floating-*`, rejects `floating-pool-*` (warm pool panes still read 0).

## Verification (required before merge)
- `cargo check -p agentmux-cef` ✅ clean; `tsc --noEmit` ✅ clean (no errors in `init/pool.ts`).
- No reducer unit test for `RelabelBrowser` — cef::Browser can't be constructed in unit tests, so the browser-registry arms (register/unregister/relabel) are runtime-verified by design.
- **Manual (Windows), pending:** open a floating pane → InstancePanel counts 1 (after reconcile); open a 2nd → 2; close → decrements; a warm un-opened pool pane still reads 0. **Then** exercise the promoted floater — move / resize / maximize / focus / redock / close — to confirm no label-addressed IPC broke.

### DIFF
--- a/.changesets/1782874440-fix-floating-pane-relabel-promoted-pane-pool-windows-floatin-6x2q.md
+++ b/.changesets/1782874440-fix-floating-pane-relabel-promoted-pane-pool-windows-floatin-6x2q.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): relabel promoted pane-pool windows floating-pool-* -> floating-* so the status bar counts them (Windows)
--- a/agentmux-cef/src/commands/window_pool.rs
+++ b/agentmux-cef/src/commands/window_pool.rs
@@ -1588,32 +1588,70 @@ pub fn promote_pane_pool_window(
             let _ = ShowWindow(outer_hwnd as HWND, SW_SHOWNORMAL);
         }
 
+        // Rename `floating-pool-<uuid>` → `floating-<uuid>` so the promoted
+        // pane is counted as a real user floater rather than filtered out as a
+        // warm pool window (SPEC_FLOATING_PANE_POOL_RELABEL_2026_06_30). Re-key
+        // every per-label store BEFORE the emit below — `emit_event_to_window`
+        // resolves the target via `get_browser(label)`, which after the reducer
+        // re-key lives under `new_label`.
+        let new_label = label.replacen("floating-pool-", "floating-", 1);
+        // Reducer state: `browsers` (+ the duplicated label field) and, if
+        // present, `window_opacities` / `pane_window_states`.
+        let _ = state.host_dispatch(crate::reducer::HostCommand::RelabelBrowser {
+            old_label: label.clone(),
+            new_label: new_label.clone(),
+        });
+        // AppState label maps (not reducer state).
+        {
+            let mut hwnds = state.window_hwnds.lock();
+            if let Some(h) = hwnds.remove(&label) {
+                hwnds.insert(new_label.clone(), h);
+            }
+        }
+        {
+            let mut meta = state.window_meta.lock();
+            if let Some(mut m) = meta.remove(&label) {
+                m.label = new_label.clone();
+                meta.insert(new_label.clone(), m);
+            }
+        }
+
         // Bind this floater to the source window's cascade hook. Deferred from
         // spawn time because the parent HWND is only known at tear-off.
         crate::floating_pane::register_floater_hwnd(
-            label.clone(),
+            new_label.clone(),
             outer_hwnd as isize,
             parent_hwnd,
         );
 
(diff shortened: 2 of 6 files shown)


a5af:
## ✅ Runtime-verified on Windows

Ran a live dev build of this branch and exercised the pane-pool floating path:

- **4 promotions → 4 clean relabels**, e.g. `[relabel] browser re-keyed old_label=floating-pool-39c4e71b… new_label=floating-39c4e71b…`, each followed by the frontend `pool:pane-promote received — bootstrapping floating pane`.
- **Label-addressed IPC from the promoted floater all succeeded** under the new label — `set_floating_redock_target` / `clear_floating_redock_hover` (drag/redock), `close_window` (close), `get_double_click_time`. These are the exact calls that would have failed if the re-key had stranded a label. They show only as `[perf]` **latency** WARNs, not failures.
- **0** `not found` / `orphan cleanup` / `browser not in state` / `get_browser` misses / panics.
- Pool correctly **refilled** with fresh unpromoted `floating-pool-*` panes after each promotion (correctly *not* counted).

The only ERROR in the log was a benign `ResizeObserver loop completed with undelivered notifications` (standard browser noise, unrelated).

The do-not-merge gate is cleared for the exercised paths (promote / drag-redock / close / count). `ToggleFloatingMaximize` wasn't explicitly exercised but keys off the same `window_hwnds`/`pane_window_states` re-key that's covered.

a5af:
## Review responses

**reagent P1 (version downgrade):** ✅ Rebased onto `main` (now 0.49.10); version files restored.

**reagent P1 (dispatch result discarded):** ✅ Fixed. `DispatchOutput` now carries `relabel_ok`; `handle_relabel_browser` sets it on success, and `promote_pane_pool_window` gates the `window_hwnds`/`window_meta` re-key **and** the emit on it — on failure (the concurrent close-before-promote race) it aborts via `cleanup_failed_pane_promote_orphan(old_label)` and returns `None`, so `browsers` and the AppState maps can never disagree.

**Codex P2 (launcher close signal):** does **not** apply, verified empirically in the live run. Codex's premise is that `registerBackendWindow` emits `backend_window_id_registered` for the new label, adding a floating-pane-map entry that close can't remove. But floaters never register a backend window id: across 4 promotions, `registerBackendWindow` fired **only for `label=main`**, with **zero** `backend_window_id_registered` for any `floating-<uuid>` (the `FloatingPaneEntry.windowId` type is even nullable for exactly this reason). The floating count is fed by the `list_window_instances` snapshot — which reads the live `browsers` map, so it's symmetric across open/close (close removes the browser → next reconcile drops it). No event-path add ⇒ no asymmetric stale-on-close. No code change needed.

Rebuilt clean (`cargo check -p agentmux-cef`); the relabel + floater IPC paths were runtime-verified on Windows (see prior comment).

PR #1897, opened 2026-07-02
refactor(frontend): extract self-contained hooks from agent-view.tsx

Agent1-asaf:
## Summary
- Extracts four self-contained logic groups from `agent-view.tsx` (1,345 → 1,074 lines) into dedicated `hooks/` files, following the existing `useAgentCommands`/`useAgentStream` pattern:
  - `useSnapshotPersistence` — 30s dirty-flag snapshot interval + on-close write
  - `useAgentDecisions` — permission decision queue + decide handler
  - `useAgentQuestions` — AskUserQuestion queue, waiting-tone effect, answer handler
  - `useAgentCloseConfirm` — process-count close-confirm onClose wrapper
- Reactive semantics preserved — same signals, same effect ordering; only relocation
- Left deeply-entangled logic (registration, launch flow, send/flush effects, startup sequence) inline to avoid fragile extraction

## Test plan
- [x] vite build passes
- [x] full vitest suite green (1574 passed)

### DIFF
--- a/.changesets/1782957522-refactor-frontend-extract-hooks-from-agent-view-if7v.md
+++ b/.changesets/1782957522-refactor-frontend-extract-hooks-from-agent-view-if7v.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(frontend): extract hooks from agent-view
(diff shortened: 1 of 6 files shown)


Agent1-asaf (commented):
Addressed both P2s: removed the unused `dispatch as dispatchDoc` import (only dispatchDocIfRegistered is used here now), and corrected the useSnapshotPersistence JSDoc to describe the actual synchronous createEffect/setInterval placement (no onMount). tsc clean.

PR #1911, opened 2026-07-02
fix(agent): composer strip near input now shows real session totals

a5af:
## Summary
- The composer strip above the textarea was documented as "final session totals" but bound to the same per-turn `sessionStats` atom as the per-query "Worked" row below the transcript, so both always showed the same (latest-query-only) numbers instead of a running sum across the session.
- Adds a genuine `sessionTotals` accumulator to `AgentPaneState` that sums cost/tokens/duration on every `TurnEnd` (persists across `TurnStart`, clears only on `TurnReset`), and points the composer strip at it. The per-turn `AgentWorkingRow` display is unchanged — it was already correct.
- See `docs/specs/SPEC_AGENT_SESSION_COST_TOTALS_2026_07_02.md` for the full root-cause writeup and design.

## Test plan
- [x] `npx vitest run` on the touched suites (`agent-pane-state/reducer.test.ts`, `agent-pane-state-store.test.ts`, `agent-pane-registration.test.ts`, `agent-pane-model.test.ts`, `view/agent/state.test.ts`) — 192/192 passing, including 2 new tests covering cross-turn accumulation and reset-on-`TurnReset`
- [ ] Manual: send several queries in an agent pane; confirm the "Worked" row per query stays scoped to that query while the composer-strip total near the input grows across queries and resets on session wipe

### DIFF
--- a/.changesets/1782997838-fix-agent-compute-real-running-totals-for-composer-strip-ins-mx7z.md
+++ b/.changesets/1782997838-fix-agent-compute-real-running-totals-for-composer-strip-ins-mx7z.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): compute real running totals for composer strip instead of reusing per-query stats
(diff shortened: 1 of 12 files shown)


a5af:
Closing and reopening with the correct agentmux agent-id routing marker.

PR #1912, opened 2026-07-02
fix(agent): promote Mode to composer strip, un-nest Controls from Log, style Session buttons

AgentA-asaf:
## Summary

The agent-pane composer strip's **Log** toggle also revealed a nested, independently-collapsible **Controls** panel (`AgentControlBar`) that duplicated the strip's Model/Effort selects and buried the only editable **Mode** control behind a second click. This resolves the open (a)/(b) decision from `SPEC_AGENT_MODEL_DROPDOWN_CLI_PIN_LOG_2026_07_02` Part C as **(b)**: Log now shows only the activity log, and all three runtime controls live at the strip's top level.

Also fixes three related bugs the user reported and a CSS defect.

## Changes

- **Promote Mode to the strip** — a top-level `<select>` next to Model and Effort, with the color-coded left border. Removes the read-only permission pill.
- **Un-nest Controls from Log** — `AgentControlBar` loses its "▸ Controls" chevron and its duplicate Mode/Model/Effort selects. It now renders **session management only** (Archive / Export / Restore / Dismiss), directly, with no inner toggle. Log reveals only `ActivityLogPanel`.
- **Fix "Session: ArchiveExport"** — the session buttons had *zero* CSS and rendered as run-together text. They now have real button chrome (border/bg/padding/hover/focus ring) and row spacing, so they read as distinct buttons.
- **Bug — Bypass indicator missing on first load** (only appeared after changing a dropdown): the Mode select now reads `runtime()`, which falls back to `DEFAULT_RUNTIME_CONFIG` (bypass), so the displayed mode matches the flags the agent actually runs with from first paint.
- **Bug — changing Effort appeared to change Model**: bind all three selects with `prop:value`, and move the registry `default: true` from Opus to **Sonnet** to match `DEFAULT_RUNTIME_CONFIG.model`. Defensive — root cause not yet reproduced live (see below).
- **Tooltips on every control** (Mode / Model / Effort / Log + session buttons) per user request.
- Housekeeping: remove dead control-bar CSS; restore the `agent-spin` keyframe (the picker uses it); add the missing archived-banner layout.

## Docs
- `docs/retro/retro-composer-strip-log-controls-nesting-2026-07-02.md` — root-cause of the Log/Controls nesting (shipped option (a) instead of (b) without confirmation).
- `docs/specs/SPEC_COMPOSER_STRIP_MODE_TOPLEVEL_2026_07_02.md` — full design, incl. Fix 4 (Bypass) and Fix 5 (Effort→Model).

## Testing / caveats
- ⚠️ **Not compiler-verified or run**: `node_modules` isn't installed in this workspace and installing would spike memory on the dev machine. Changes were manually reviewed + swept for dangling references (clean), but have **not** been typechecked, built, or exercised in the app.
- **Deferred**: Fix 4's deeper launch-time seeding of `agent:runtime` meta (the user-visible symptom is already resolved by the Mode-select fallback); a live repro of the Effort→Model bug to confirm the `prop:value` fix is the right one.


AgentA-asaf:
Addressed both review items in 415e8276:

- **[P1]** `AgentControlBar.tsx` — the session-actions row now excludes `isLargeSession()` (`lineCount() > 0 && !isArchived() && !isLargeSession()`), so its Archive button no longer renders alongside the large-session banner's Archive for sessions ≥ 500k lines.
- **[P2]** `_composer-strip.scss` — merged the two `.agent-composer-strip-select--mode` rules (min-width + padding-left) into one.

The 14 "deletion of recently-added code" flags are intentional: this PR removes the nested Model/Effort dropdowns from `AgentControlBar` because they were promoted to the top-level composer strip (per `SPEC_COMPOSER_STRIP_MODE_TOPLEVEL_2026_07_02`).

PR #1918, opened 2026-07-02
feat(armory): rename Preset to Bundle (App API + UI, phase 2)

Agent1-asaf:
## Summary
Phase 2 of the composable-agent-model refactor (SPEC_PRESET_TO_BUNDLE_REFACTOR_2026_07_02.md): retire "Preset" in favor of **Bundle** at the App API + UI layer. Storage + resolver untouched (Phases 3-4).

- App API `preset.*` → `bundle.*`; old `preset.*` commands kept as **aliases for one release** (both registered under the same handler logic)
- Frontend command bindings + labels + types "Preset" → "Bundle"
- Armory "Presets" tab label → "Bundles" (internal id `memories` unchanged; icon `sliders` → `layer-group`)
- Storage (`db_memory_bundles`, `bundle_memory_*`) and resolver unchanged — Phase 4
- Persisted `view: "memory"` key unchanged
- Contract baseline: added the 5 `bundle.*` entries to `KNOWN_REGISTERED_UNDECLARED`; the `preset.*` aliases stay (backend-only, both surfaces are registered-but-undeclared)

## Test plan
- [x] `cargo check` clean (no new warnings)
- [x] `tsc --noEmit` clean
- [x] `vitest run rpc-contract AgentLaunchModal` green (5/5)
- [x] `task build:frontend` passes


Agent1-asaf (commented):
Addressed the P2: `bundle_list_impl` now returns `{ "bundles": …, "presets": … }` — new `bundle.list` callers read `bundles`; the retained `preset.list` alias's existing agent/REST consumers (`/api/v1/agent/preset/list`) still read `presets` for back-compat. The `presets` key drops in Phase 4 when the alias is removed. Rather than a straight key rename (which would break existing `.presets` consumers), the dual key preserves the alias contract for the compat window.

Agent1-asaf (commented):
Addressed the P2: memory-model.ts viewIcon 'sliders' → 'layer-group' so the standalone bundle pane icon matches the Armory rail's Bundles tab (armory-view.tsx). tsc clean.

PR #1923, opened 2026-07-02
feat(srv): model catalog fetch layer — GET /v1/models via OAuth (phase 1)

AgentA-asaf:
## Summary
First slice of `SPEC_MODEL_CATALOG_REFRESH_2026_07_02` — make the agent-pane model dropdown **API-sourced** instead of a hand-curated list whose version labels drift (today it shows "Sonnet 4.6" though **Sonnet 5** is current).

This PR lands the **foundation**: `agentmux-srv/src/backend/model_catalog.rs`
- `read_oauth_access_token()` — pulls the agent's OAuth token from the isolated `.credentials.json` (`claudeAiOauth.accessToken`).
- `fetch_model_catalog()` — `GET https://api.anthropic.com/v1/models` with `Authorization: Bearer <token>`; returns `None` on any failure (macOS Keychain, 401/expiry, network) so callers fall back to the bundled catalog.
- `parse_models()` — unit-testable body→`CatalogModel` mapping (3 tests).

**Verified live (2026-07-02):** a Claude **Max subscription** OAuth token returns HTTP 200 (no API key, no beta header) with the current models incl. `claude-sonnet-5` | "Claude Sonnet 5". Metadata-only — this token is never used for `/v1/messages`.

## Remaining (tracked in the spec — follow-up PRs)
- Cache per `(provider, cli_version)`; `providers.models` RPC (+ `ProviderConfig.models`).
- Refresh trigger on CLI install/upgrade (+ optional lazy >30d refetch).
- Frontend: overlay `providers/index.ts` with the cached catalog; converge the strip / `/model` dropdowns. **This is the phase that makes "Sonnet 5" visibly appear** — not this PR.

## Testing / caveats
- ⚠️ **Not build-verified locally** (Rust toolchain not built here) — relies on CI `cargo` build + the included unit tests.
- Pure additive: new `pub mod` + doc; no existing behavior changes yet.

PR #1926, opened 2026-07-03
feat(agent): providers.models RPC + live model-catalog overlay

AgentA-asaf:
## What

End-to-end wiring so the composer-strip **Model** drop-up shows the authoritative Claude model list from `GET /v1/models` instead of hand-curated labels that drift each release ("Sonnet 4.6" → "Sonnet 5" …). Builds on the fetch layer merged in #1923.

## Backend
- **New `providers.models` RPC** (`server/providers_handlers.rs`): resolves the account-global OAuth token via `DataPaths::provider_auth_dir("claude")` and calls the existing `fetch_model_catalog()`. Best-effort — returns `{models:[]}` when there is no token (logged out / macOS Keychain), so the frontend keeps its static list. Non-Claude providers return empty. Registered next to the tool-store handlers.

## Frontend
- `rpc-api/misc.ts`: `ProvidersModelsCommand` binding. Identical command string on both sides keeps the `rpc-contract` baseline green.
- `providers/index.ts`: signal-backed model **overlay**. `setProviderModels` folds the catalog in, **behavior-preserving**:
  - curated family entries (`opus`/`sonnet`/`haiku`) keep their short `value` (which `--model` resolves to the current version), the `default` marker, `aliases`, and `description`; only the **label** is refreshed to the newest matching model. Version-tuple comparison means **Sonnet 5 → 5.x → 6 auto-wins** regardless of API ordering.
  - families the curated list does not cover (**Fable** today, any future family) are appended automatically, newest per family, using the concrete id as `value`. **No family is hardcoded.**
- `app-init.ts`: one-time fire-and-forget fetch after render; the drop-up shows the static list until it resolves, then re-renders with fresh labels.

## Why this shape
The static Claude entries intentionally use family aliases (`sonnet`) as `--model` values so the CLI resolves to the current version, and `sonnet` carries `default: true` that `DEFAULT_RUNTIME_CONFIG` readers depend on. A naive full replace with concrete API ids would break both. The overlay refreshes only what drifts (labels) and adds only what is genuinely new (families).

## Verification
- `cargo check -p agentmux-srv` — clean (pre-existing warnings only)
- `tsc --noEmit -p tsconfig.json` — exit 0
- `vitest run test/contract/rpc-contract.test.ts` — 4/4 pass (command present in both `registered` and `declared`; baseline diffs unchanged)

### DIFF
--- a/.changesets/1783039588-feat-agent-providers-models-rpc-catalog-overlay.md
+++ b/.changesets/1783039588-feat-agent-providers-models-rpc-catalog-overlay.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): `providers.models` RPC + live model-catalog overlay for the strip drop-up
--- a/.changesets/1783039588-feat-agent-providers-models-rpc-catalog-overlay.md
+++ b/.changesets/1783039588-feat-agent-providers-models-rpc-catalog-overlay.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
 feat(agent): `providers.models` RPC + live model-catalog overlay for the strip drop-up
--- a/agentmux-srv/src/server/mod.rs
+++ b/agentmux-srv/src/server/mod.rs
@@ -14,6 +14,7 @@ mod reactive;
 pub(crate) mod service;
 mod shell_handlers;
 mod tool_handlers;
+mod providers_handlers;
 mod voice;
 pub(crate) mod wave_obj_bridge;
 mod websocket;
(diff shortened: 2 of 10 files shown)


AgentA-asaf:
Addressed the [P2]: getProviderList() now routes through getProvider() (commit 4868a561), so list consumers reading .models get the same overlay (refreshed labels / new families) as getProvider() callers instead of the raw static list. 	sc --noEmit still exit 0.

AgentA-asaf:
Addressed both P2s (commit 953b6218): changeset type \patch\->\minor\ (this PR adds a feat RPC), and \providers.models\ now returns \mpty_result()\ on \DataPaths::from_env()\ failure so the RPC stays best-effort per the module contract instead of erroring. \cargo check\ clean.

Issue #1624, opened 2026-06-20
Reconcile per-agent keychain with the live identity-bundle system


a5af:
## Background

Shelved from active work (was "Priority 3: migrate `AgentDefinition.accounts` blob → `db_agent_identity_links` junction" in `SPEC_MEMORY_IDENTITY_ARCH_2026_06_19.md`).

**That plan was a functional no-op.** Verified that agent launch resolves credentials from **neither** the blob **nor** the v6 junction — both are dead. Launch uses the v7 **identity-bundle** path:

```
block → AgentInstance.identity_id → db_identity_bindings → db_identity_accounts → env
```
(`agentmux-srv/src/identity/resolver.rs:415`, `inject_identity_env_with_broker`).

Three per-agent identity mechanisms exist; two are dead:

| Mechanism | Written by | Read at launch? |
|---|---|---|
| `AgentDefinition.accounts` blob | ID-card/keychain modal | ❌ |
| `db_agent_identity_links` (v6 junction) | link/unlink RPCs (no UI caller) | ❌ |
| identity bundle (v7) | launch modal Identity dropdown | ✅ |

Code confirms: `storage/agents.rs:69` ("migration never completed"), `storage/identities.rs:13` ("legacy"), `agent_handlers.rs:2691` ("bundles replacing per-agent `db_agent_identity_links` semantics").

**Consequence:** the per-agent keychain modal (ID-card icon) is decorative today — it writes a blob launch ignores. Real per-agent creds come from the identity bundle picked at launch.

## The real decision

How should the per-agent keychain relate to the bundle system that drives launch?

- **Option A (recommended) — bundle-backed keychain.** Each agent gets a dedicated *private* identity bundle (lazily created); the keychain modal edits its bindings; launch defaults `instance.identity_id` to it. Keychain becomes functional; shared bundles remain a launch override; removes both dead layers. New `AgentDefinition.default_identity_id` + `db_identity_bundles.is_private` (schema v10) + one-time blob→bundle backfill.
- **Option B — remove the dead layers.** Delete keychain/blob/junction; standardize on launch-time bundle selection. Smaller, but a UX regression for per-agent setup.

## Open questions
1. Private bundle via `is_private` column (recommended) vs `agent:<id>` id-convention.
2. Add `default_identity_id` to `db_agents` consolidation table too?
3. Launch override = one-off (recommended) vs mutates the default.
4. Deprecation timeline to drop the blob column + v6 junction RPCs.
[...]

PR #1927, opened 2026-07-03
feat(identity): direct agent-account bindings — additive dual-read + backfill (Phase 3 slice 2 PR-A)

Agent1-asaf:
## Summary
PR-A of Phase 3 Slice 2 (auth-critical, but **additive & behavior-preserving**). Revives the already-existing `db_agent_identity_links` direct agent↔account table as a resolution path, and backfills it from existing bundle bindings — without changing spawn-auth behavior on existing DBs. Sets up PR-B (flip to direct-only) and resolves the direction in #1624.

- **Resolver dual-read:** prefers direct links (`db_agent_identity_links` keyed on `instance.definition_id`), falls back to bundle bindings. Existing DBs (no direct links) → identical behavior.
- **Backfill migration `m0013`:** copies each non-sentinel bundle's bindings into direct links on the instance's definition (shared store; idempotent; `default` OAuth bundle included; `blank`/`''` skipped; deterministic last-write-wins on collisions).
- **No change** to injection logic, provider→env mapping, secret resolution, or the blank/ambient short-circuit.
- Bundle read path stays fully intact (this is dual-read, not a flip).

## Test plan
- [x] `cargo check` + `cargo check --tests` clean, no new warnings
- [x] `cargo test resolver identity` green (incl. new direct-link + direct-wins tests)
- [x] new m0013 migration tests: backfill correct, idempotent, sentinel-skipped, default-OAuth included

Refs #1624.

### DIFF
--- a/.changesets/1783040229-feat-identity-revive-direct-agent-account-bindings-additive--zwe5.md
+++ b/.changesets/1783040229-feat-identity-revive-direct-agent-account-bindings-additive--zwe5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(identity): revive direct agent-account bindings (additive, dual-read + backfill)
--- a/agentmux-srv/src/identity/resolver.rs
+++ b/agentmux-srv/src/identity/resolver.rs
@@ -414,6 +414,64 @@ pub async fn inject_identity_env_async(
     }
 }
 
+/// Resolve the identity bindings for an instance, preferring the
+/// **direct** agent↔account links over the bundle bindings.
+///
+/// Phase 3 slice 2 PR-A dual-read: the resolver historically walked
+/// `instance.identity_id → db_identity_bindings (bundle)`. This helper
+/// first consults `db_agent_identity_links` keyed on the instance's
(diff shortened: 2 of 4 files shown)


Agent1-asaf (commented):
Addressed the P2: m0013 now matches on `instance_list` and `tracing::warn!`s on error instead of `unwrap_or_default()` silently dropping instances — a partial read is now visible in logs. Left non-fatal (the dual-read resolver still falls back to the bundle path for any un-backfilled instance, so a source-read failure degrades gracefully rather than aborting startup). cargo check clean, m0013 tests 5/5.

Agent1-asaf (commented):
Fixed the second occurrence (line 154): bundle_identity_bindings now logs on error + continues, consistent with the instance_list guard. No remaining unwrap_or_default() on fallible reads in m0013. cargo check clean, m0013 tests pass.

Agent1-asaf (commented):
Addressed: 'written' now counts distinct (definition, provider) rows, not upsert calls — a collision overwrite no longer inflates the logged count, so it matches the actual db_agent_identity_links row count. m0013 tests pass.

PR #1937, opened 2026-07-03
fix(scripts): resolve task via PATHEXT instead of hardcoded .exe extension

AgentA-asaf:
## Summary

- `dev-agent.cmd` / `package-agent.cmd` hardcoded `task.exe`, assuming `task` was installed via `go install` / WinGet. On machines where `task` was installed via npm (go-task's npm wrapper), the binary is `task.cmd`, not `task.exe` — calling `task.exe` fails immediately with "not recognized as an internal or external command" (exit 1, <100ms).
- Since these `.cmd` wrapper scripts already run under `cmd.exe` (that's what fixes the bash-can't-resolve-.cmd-files gap — no bash lookup needed to invoke the wrapper itself), calling bare `task` lets `cmd.exe`'s `PATHEXT` resolve to whichever extension actually exists, instead of assuming one specific install method.
- Also adds `package-agent.cmd`, the `task package` counterpart to the existing `task dev` bridge script, for portable-build invocation from an agent's MCP Shell.

## Test plan

- [x] `cmd /C scripts\package-agent.cmd --help` now resolves and runs `task` (previously: `'task.exe' is not recognized as an internal or external command`)
- [x] Confirmed via `mcp__agentmux__Shell` — build now actually compiles (npm/frontend steps observed running) instead of failing in <100ms

### DIFF
--- a/.changesets/1783063556-fix-scripts-resolve-task-via-pathext-instead-of-hardcoded-ex-2dyj.md
+++ b/.changesets/1783063556-fix-scripts-resolve-task-via-pathext-instead-of-hardcoded-ex-2dyj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(scripts): resolve task via PATHEXT instead of hardcoded .exe extension in agent bridge scripts
--- a/scripts/dev-agent.cmd
+++ b/scripts/dev-agent.cmd
@@ -34,8 +34,11 @@ set "GIT_BIN=C:\Program Files\Git\bin"
 set "GIT_USR=C:\Program Files\Git\usr\bin"
 set "PATH=%GIT_BIN%;%GIT_USR%;%PATH%"
 
-:: Call task.exe by explicit extension (Gap A fix — no bash lookup needed here).
-:: task is installed as task.exe (via go install / WinGet), not task.cmd.
+:: Call bare `task` (no extension) — this .cmd wrapper itself runs under cmd.exe
+:: (Gap A only applies to bash resolving .cmd files, not to us), so PATHEXT
+:: resolution finds task.exe (go install / WinGet) OR task.cmd (npm global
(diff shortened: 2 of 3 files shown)


AgentA-asaf:
Verified this is safe across all 3 platforms:

- **macOS/Linux: zero effect.** These \`.cmd\` scripts are Windows-only — \`task dev\`/\`task package\` are called directly on those platforms (see the scripts' own header comments + CLAUDE.md). This PR never touches non-Windows code paths.
- **Windows: this closes a whack-a-mole, not opens a new one.** Git history on this file:
  1. PR #1816 originally hardcoded \`task.cmd\` — broke for anyone whose \`task\` came from \`go install\`/WinGet (produces \`task.exe\`, no \`.cmd\` shim).
  2. Commit \`aa0f51a3\` (2026-06-28) "fixed" that by hardcoding \`task.exe\` instead — which fixed *that* dev's machine but broke mine (npm-installed \`task\`, giving \`task.cmd\`, no \`.exe\`). This is exactly the failure I hit and diagnosed this session.
  3. This PR uses bare \`task\` instead of guessing an extension.

Bare \`task\` is safe here specifically because these wrapper scripts already run under \`cmd.exe\` (that's the whole point of the wrapper — see the file's Gap A/B comments). \`cmd.exe\` resolves bare commands via \`PATHEXT\` (tries \`.com\`/\`.exe\`/\`.bat\`/\`.cmd\` in order) natively — no bash involved, no ambiguity. The only reason an explicit extension was ever added was Gap A, which is specifically about *MSYS2 bash* failing to resolve extensionless files — a problem that doesn't exist once you're already inside the cmd.exe-run wrapper. So this isn't a third guess at "the right extension," it's removing an unnecessary assumption that two prior commits each got wrong for the other's machine.

PR #1940, opened 2026-07-03
feat(cef): report CEF dlopen + cef_init sub-stages to splash telemetry

AgentO-asaf:
## Summary

Sixth PR in the `SPEC_MACOS_LAUNCH_SPEED_AND_SPLASH_TELEMETRY_2026_07_02.md` stack, prompted directly by a user question: after seeing the splash's stage list, the total time didn't add up to the sum of the listed stages. Root cause: the `"host"` stage (#1931) only measures process-spawn latency, leaving everything CEF does after spawn — framework `dlopen`, `cef::initialize()` — invisible. This PR closes that gap by having the host process report those phases back to the launcher, live, over the existing IPC connection.

**Wire protocol** — two new `Command` variants in the shared `agentmux-common::ipc::Command` enum: `ReportStartupStageBegin`/`ReportStartupStageEnd`. Status is a plain string on the wire (`"ok"`/`"warn"`/`"error"`) rather than importing the launcher's `StartupStatus` type across the crate boundary.

**Launcher side** — forwarded directly into `StartupEventSink`, bypassing the reducer entirely (same short-circuit-before-reducer pattern already used for `GetEvents` — not mirrored State, not broadcast, purely a side-channel for the splash panel). `ServerCtx` gained a `startup_sink` field threaded from the same `launcher_setup` async block (#1932) that already owns the sink. `StartupEvent`'s `stage`/`label` fields are `&'static str` (every existing caller passes a compile-time constant) — the host's runtime JSON strings are `Box::leak`'d to a handful of small, bounded allocations rather than widening the shared type (which Windows' `splash.rs` and Linux's `splash_linux/` also consume, neither of which I can build/verify here).

**Host side** — `dlopen` can only be reported **retroactively**: it finishes before `connect_to_launcher` even runs, so there's no live connection to report through yet. Timed locally with an `Instant`, sent as a `StageBegin`+`StageEnd` pair with the already-elapsed duration once the connection exists. `cef::initialize` is genuinely **live** — `connect_to_launcher` always runs before it, on every platform, not just macOS — so this closes the gap for Windows/Linux's existing splash consumers too, not only the macOS one added in #1933.

## Test plan
- [x] `cargo build`/`cargo test` — `agentmux-common` (86), `agentmux-cef` (134), `agentmux-launcher` (199), all passing, zero failures
- [x] **Verified beyond unit tests — a real launch, not just selftest.** Rebuilt the full signed `.app`, launched a real (non-selftest) instance on an isolated `AGENTMUX_CHANNEL` (specifically to avoid touching the actual already-running AgentMux session on this machine), and confirmed via temporary logging (added, verified, then removed before this commit) that the full IPC round-trip fires with real measured numbers:
  ```
  dlopen:   2960ms
  cef_init: 2613ms
  ```
  ~5.6s combined — confirms the audit's "dlopen dominates" hypothesis directly rather than leaving it as speculation. Spec updated with these numbers.
- [x] Windows: same limitation as prior PRs in this stack — no MSVC C toolchain available here to cross-compile. The `windows.rs`/shared-crate changes are structurally identical to the already-proven `report_window_opened` pattern in the same files.


AgentO-asaf:
Added a second, independent verification pass per review request — this time with **permanent** `tracing::debug!` calls in `report_startup_stage_begin`/`end` (agentmux-cef side) instead of temporary logging, so future verification doesn't need another add/remove cycle. Enabled via the existing `RUST_LOG` env var (`agentmux_cef=debug`), no new dependency.

Confirmed on a fresh real launch (isolated `AGENTMUX_CHANNEL`, same safety precaution as the first verification — didn't touch the actual running AgentMux instance on this machine):

```
dlopen:   2166ms
cef_init:  627ms
```

Different numbers from the first commit's measurement (2960ms/2613ms) — expected run-to-run system-load variance (this machine had other AgentMux instances + background builds running), not a bug. The mechanism reports correctly and reliably across two independent runs.

PR #1961, opened 2026-07-04
feat(agent): ghost-text next-prompt suggestion via the Ambient Model Call gateway

Agent3-asaf:
## Summary

Implements `docs/specs/SPEC_AMBIENT_GHOST_TEXT_NEXT_PROMPT_2026_07_03.md`: a
second consumer of the Ambient Model Call gateway (`agentmux-srv/src/ambient`,
#1947). After a turn completes, predict a short, plausible next user message
and show it as dimmed ghost text in the empty composer — Tab accepts it.

This reimplements Claude Code CLI's own interactive-mode feature. Confirmed
by inspecting the installed CLI binary directly: a real `prompt_suggestion`
stream-json event exists, gated by four independent checks (env var,
GrowthBook remote flag, non-interactive-mode, "swarm teammate"). It's
unreachable from AgentMux because the non-interactive check unconditionally
suppresses it whenever Claude runs via `-p`/`--print` — which is how
AgentMux always drives the CLI. Verified empirically: ran the real CLI with
`-p --output-format stream-json` and its own enabling env var set — zero
`prompt_suggestion` events. (One web search result claimed a
`--prompt-suggestions` CLI flag exists; it doesn't — `claude --help` and a
live invocation both confirm that's fabricated blog content. Full writeup
in the spec.)

## Backend

- New RPC `session:next_prompt_suggestion`, structurally identical to
  `session:activity_summary` — same gateway admission/cancellation, same
  Haiku CLI invocation, same result/usage parsing.
- Renamed `invoke_cli_for_activity` → `invoke_ambient_haiku_call` and
  extracted `read_recent_activity_digest` as a shared helper — both ambient
  call sites need the identical tail-read + digest-extract + CLI-invoke
  logic now. Pure dedup, no behavior change to the existing activity-summary
  path.
- New meta key `term:next_prompt_suggestion`; tokens tagged
  `"ambient:next_prompt_suggestion"` in the total-tokens view.

## Frontend

- New hook `useNextPromptSuggestion.ts`, mirroring
  `useAgentActivitySummary.ts` but with sharper lifecycle rules — a stale
  suggestion here can put words in the user's mouth, not just show a stale
  label:
  - Cleared on every new turn start (`Submitting`), not just checked against
    the current generation — `term:ambient_summary` deliberately persists
    across turns, this key deliberately does not.
  - Checked again at write time against composer-emptiness (not just at
    request time), closing the race where the RPC resolves after the user
    already started typing.
- `AgentFooter.tsx`: composer placeholder now prefers the ghost-text
  suggestion (falls back to the existing voice/default placeholder); Tab
  accepts it into the real input (matches Claude Code's own terminal UX so
  it's a consistent mental model for users who know the CLI); first
  keystroke into a previously-empty box clears any pending suggestion so it
  can't silently reappear if the box is emptied again.

## Test plan

- [x] Full backend suite: 1296 passed, 0 failed (includes the existing
      `ambient::tests::*` — generic over purpose, so this validates the
      gateway's second real consumer for free)
- [x] Full frontend suite: 1581 passed (105 files), 0 failed
- [x] `tsc --noEmit` clean
- [x] `cargo check` clean, no new warnings
- [ ] **Not live-tested end-to-end** in a running dev instance (needs a real
      completed agent turn to trigger). Backend logic is unit-tested and
      reuses the already-verified `activity_summary` pipeline; frontend
      logic typechecks cleanly. Flagging this explicitly rather than
      claiming full manual verification — happy to do a `task dev` pass if
      wanted before merge.

### DIFF
--- a/.changesets/1783156315-feat-agent-ghost-text-next-prompt-suggestion-via-the-ambient-3vcb.md
+++ b/.changesets/1783156315-feat-agent-ghost-text-next-prompt-suggestion-via-the-ambient-3vcb.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): ghost-text next-prompt suggestion via the Ambient Model Call gateway
--- a/.changesets/1783198154-fix-agent-clear-stale-next-prompt-ghost-text-on-session-end-nu35.md
+++ b/.changesets/1783198154-fix-agent-clear-stale-next-prompt-ghost-text-on-session-end-nu35.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): clear stale next-prompt ghost text on session end
--- a/agentmux-srv/src/backend/rpc_types/commands.rs
+++ b/agentmux-srv/src/backend/rpc_types/commands.rs
@@ -352,6 +352,11 @@ pub const COMMAND_SESSION_EXPORT: &str = "session:export";
 // routed through the Ambient Model Call gateway, crate::ambient)
 pub const COMMAND_SESSION_ACTIVITY_SUMMARY: &str = "session:activity_summary";
 
+// Per-turn ghost-text next-prompt suggestion (Haiku-powered, writes
+// term:next_prompt_suggestion; routed through the Ambient Model Call gateway,
+// crate::ambient). See docs/specs/SPEC_AMBIENT_GHOST_TEXT_NEXT_PROMPT_2026_07_03.md.
+pub const COMMAND_SESSION_NEXT_PROMPT_SUGGESTION: &str = "session:next_prompt_suggestion";
+
 // Option E (PR 1 of 2) — agent-anchored session zones.
 // A session zone is bound to the *agent definition* (`definition_id`),
 // not the identity bundle. Every block of the same agent reads/writes
--- a/agentmux-srv/src/backend/rpc_types/session.rs
+++ b/agentmux-srv/src/backend/rpc_types/session.rs
@@ -38,6 +38,32 @@ pub struct ActivitySummaryResult {
     pub tokens: Option<TokenCounts>,
 }
 
+// ---- Ghost-text next-prompt suggestion types ----
+
+/// Request for session:next_prompt_suggestion — predict a plausible next
+/// user message via Haiku, routed through the Ambient Model Call gateway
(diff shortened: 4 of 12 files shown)


AgentY-asaf:
Fixed in de045a15 — good catch. `useBlockActivity.ts`'s `clearActivity` already owns the session-end clear for `term:osc_title`/`term:ambient_summary`; extended it to clear `term:next_prompt_suggestion` there too, rather than adding a second `ControllerStatus` subscription. Updated both hooks' doc comments to keep the invariant list accurate.

AgentY-asaf:
Fixed both in bcb7a760:
- **P1** (AgentFooter.tsx): replaced the `value.length === 1` check with tracking the box's emptiness immediately before each edit (`boxWasEmpty`), so paste/voice-transcript inserts that add multiple characters to an empty box in one `input` event also clear the stale suggestion, not just a single typed keystroke.
- **P2** (session.rs): corrected `NextPromptSuggestionResult`'s doc comment — the frontend writes `term:next_prompt_suggestion`, not the backend handler (matching the accurate inline comment already on `activity_summary`'s equivalent call site).

`cargo test -p agentmux-srv` (1296 passed) and `npx vitest run` (1581 passed) both clean after the change; no test infra exists yet for AgentFooter.tsx to add targeted coverage against.

Agent3-asaf:
Fixed in bdd3ba61 — good catch, this was a real reintroduction of the bug via a different trigger. `handleSend` and the Escape-clear handler both wrote `textareaRef.value = ""` directly, bypassing the `input` event `boxWasEmpty` relies on.

Rather than patch those two sites individually (which just invites a fourth review round finding yet another bypass), introduced `writeComposerValue` as the single choke point for every programmatic write to `textareaRef.value` — `acceptCompletion` and `setComposerValue` route through it too now (they had the same latent bug in the opposite direction: writing non-empty text without clearing a stale `boxWasEmpty=true`). No call site assigns `.value` directly anymore outside `writeComposerValue` itself.

`tsc --noEmit` clean. `vitest run` — 1580/1581 passed; the one failure (`tool-renderers/registry.test.ts`, unrelated file) was a timeout from full-suite parallel contention, confirmed passing in isolation (1.8s, well under its 5s timeout) and untouched by this change.

PR #1969, opened 2026-07-05
fix(window): mid-session window-close leak — pool demote (round 6) + backend_close_window auth fix

AgentA-asaf:
## Summary

Fixes the mid-session window-close leak: closing any secondary window permanently leaked a live renderer process (~100MB commit each), the reducer `browsers` map entry, and the entire srv-side window/workspace/tab/block state — because CEF 148 Views **parks** the hosted browser on every close/destroy sequence and `on_before_close` (where all cleanup lived) never fires.

**Six investigation rounds** (full evidence trail in `docs/retro/retro-window-lifecycle-leak-2026-07-04.md`):
- Rounds 2–3 (AgentA): `close_browser(1)` after/before `window.close()` — destruction initiates (`do_close` fires) but parks.
- Round 4 (Agent2): native `DestroyWindow` (new strict HWND resolver + main-guard) — window genuinely dies, **zero** CEF callbacks; a Views browser does not tear down with its HWND.
- Round 5: arm destruction (`close_browser(1)`) then `DestroyWindow` — `can_close→1` fires; `on_before_close` still never. Break pinned inside CEF's BrowserView→Browser destruction.
- **Round 6 (the fix): pool demote — embrace the recycle.** Closing a promoted pool window now:
  1. Runs the srv cleanup **imperatively** (`demote_srv_cleanup`): `backend_close_window` → srv `CloseWindow` → `delete_workspace` cascade, with the #1965 registration-race retry.
  2. Returns the window to the warm pool: `HostCommand::DemotePoolWindow` (is_pool flip + `unpromoted` re-insert), park offscreen + hide, HWND/Views-window re-cached, reload to the `pool=1` boot URL — queue re-entry rides the normal `PoolWindowReady` handshake. **The renderer is reused, not leaked.**
  3. Demote cap `POOL_TARGET_SIZE + 2`; beyond it (or when strict HWND resolution fails — mutation-free fallback), the round-5 destroy path runs with srv state still cleaned.

**Also fixes a latent May-11 regression found live:** `backend_close_window` authenticated via the `?authkey=` query param, disabled for HTTP routes in the 2026-05-11 security audit — every call 401'd since, unnoticed because its only caller never fires. Now sends `X-AuthKey`.

## Verification (live, `AGENTMUX_DEBUG_CLOSE=1`, per spec §4b)
- Demotes confirmed per close up to the cap; `backend_close_window` **HTTP 200**; workspace cascade runs
- **Reopen promotes a recycled window and its close re-demotes it — the loop is in equilibrium; steady-state closes leak zero renderers**
- 3 new reducer tests for `DemotePoolWindow` (round-trip, idempotency, semaphore untouched); cef suite green

## Follow-ups (tracked in the retro)
- srv `CloseWindow` leaves the bare window row (workspace/tabs/blocks are cascaded) — same dual-write class as #864
- Non-Windows keeps the Views close path (no parked-browser evidence collected there)
- Round-5 instrumentation kept (load-bearing forensics)


AgentY-asaf:
## Rounds 4 + 5 executed (Agent2) — negative results, break point pinned one level deeper

Picked up the handoff and ran the recommended round 4, plus a round 5 the results demanded. Full evidence in the updated retro; short version:

**Round 4 — native `DestroyWindow`, strict HWND resolution (new `resolve_window_hwnd_strict`: validated cache/registry only, EnumWindows fallback banned on the destroy path, main-HWND guard):**
- `ret=1`, same-thread ownership verified, `IsWindow_after=0`, visible window genuinely gone
- **ZERO CEF callbacks** (no can_close/do_close/on_before_close), renderer leaked anyway
- Key negative result: a **Views-hosted browser does not tear down when its top-level HWND is destroyed natively** — the #1957 WM_DESTROY-cascade mechanism is specific to `set_as_child` panes, not BrowserView plumbing.

**Round 5 — `close_browser(1)` to arm destruction, THEN `DestroyWindow`:**
- `do_close` and `can_close → 1` now fire per close; window death delivered
- `on_before_close` **still never fires**; renderers 5→8 over 3 cycles
- Break point now pinned: **between `can_close → true` and `OnBeforeClose`, inside CEF 148's BrowserView→Browser destruction** — it parks regardless of whether Views or native code delivers the window death.

**Conclusion after 5 rounds:** no host-side sequence of sanctioned close APIs + native destruction completes browser destruction on this build. Recommended round 6 is your fallback candidate, now primary: **pool-demote (embrace the recycle)** — close = hide + imperative srv cleanup (label is known; no on_before_close needed) + re-enqueue as a warm pool entry, so the renderer is *reused* rather than leaked. Fallback: imperative cleanup only (fixes state/chip drift, accepts the parked renderer).

Round-5 code + forensic instrumentation left on the branch per the handoff protocol.

AgentY-asaf:
## Round 6 implemented and VERIFIED — this PR is no longer WIP

**Pool demote works.** Closing a promoted pool window now recycles it instead of leaking it:

- **srv cleanup runs imperatively** at close time (`demote_srv_cleanup`): `backend_close_window` → srv `CloseWindow` → `delete_workspace` cascade with the #1965 retry — no `on_before_close` needed.
- **The window returns to the warm pool**: new `HostCommand::DemotePoolWindow` (is_pool flip + `unpromoted` re-insert, 3 reducer tests), park offscreen + hide, HWND/Views-window re-cached, reload to the `pool=1` boot URL — queue re-entry rides the normal `PoolWindowReady` handshake.
- **Verified live** (3 open/close cycles + reopen): closes demote (`demoted back into pool` per close up to the cap), `backend_close_window` returns **HTTP 200**, workspace cascade confirmed, and — the money shot — **the next open promoted a RECYCLED window and its close re-demoted it: the loop is in equilibrium.** Steady-state closes leak zero renderers.

**Latent bug found en route (also fixed here):** `backend_close_window` authenticated via the `?authkey=` query param, which the 2026-05-11 security audit deliberately disabled for HTTP routes — **it has 401'd on every call since May**, unnoticed because its only caller never fires and pre-#1965 the response was never read. Now sends the `X-AuthKey` header. Even with a future CEF fix for the destruction parking, the cleanup chain would have been silently broken without this.

**Follow-ups noted in the retro:** srv `CloseWindow` still leaves the bare window ROW behind (workspace/tabs/blocks are cascaded; the row is the same dual-write class as the #864 authority work); non-Windows keeps the Views close path; demote cap overflow takes the round-5 destroy fallback (srv state still cleaned).

Rounds 4/5 negative results + full round-6 evidence: `docs/retro/retro-window-lifecycle-leak-2026-07-04.md`.

PR #632, opened 2026-05-01
feat(E.4-A): layout reducer focused/magnified arms

AgentA-asaf:
## Summary

Step 3 (Option A) of the architecture-completeness plan. Migrates the focused/magnified layout fields onto the srv reducer; rootnode / leaforder / pendingbackendactions stay on the existing wcore-direct paths per the deferred Option B decision.

Closes (partial): [reducer-architecture-gaps §1 E.4](../blob/main/docs/retro/reducer-architecture-gaps-2026-05-01.md). The `handle_move_tab` strict-mode flip (gap §4) ships as PR 2 of this step after Option A soaks.

Spec: `docs/specs/SPEC_PHASE_E4_LAYOUT_REDUCER_2026-05-01.md`
Plan: `docs/retro/next-steps-architecture-completeness-2026-05-01.md` step 3

## What changed

- Wire format: `Command::{SetFocusedNode,SetMagnifiedNode}` + `Event::{FocusedNodeChanged,MagnifiedNodeChanged}`.
- Reducer arms with no-op short-circuit + unknown-tab error path; `TabRecord` gains `focused_node_id` + `magnified_node_id` (bootstrap-loaded from each tab's `LayoutState` row).
- Persist subscriber writes the field within the existing layout transaction via the `Tab.layoutstate` join (`apply_focused_node_changed` / `apply_magnified_node_changed`).
- RPC routing: `("object", "UpdateObject")` for `OTYPE_LAYOUT` extracts focusednodeid/magnifiednodeid and dispatches the new reducer commands before the existing wcore write. Other LayoutState fields (rootnode/leaforder/pendingbackendactions) keep flowing through wcore unchanged. Frontend send-sites untouched.
- Launcher reducer / event_log / ipc-server: match-arm wiring for the new srv-pipe-only commands and events.
- Unit tests for round-trip, no-op, unknown-tab, magnify-clear (7 new tests).

## What's NOT in this PR

- Option B (rootnode / leaforder migration) — deferred per spec §3.
- `handle_move_tab` strict-mode flip — PR 2 of step 3.

## Test plan

- [x] `cargo check -p agentmux-srv` clean
- [x] `cargo check -p agentmux-launcher` clean
- [x] `cargo check -p agentmux-cef` clean
- [x] `cargo test -p agentmux-srv reducer::` — 68 passing (7 new)
- [ ] Smoke: switch tabs, click panes, verify focused-pane state preserved per-tab
[...]

PR #1970, opened 2026-07-05
fix(layout): UpdateObject routes layout pushes through LayoutSetTree — SPEC_864 Phase 2 (weak single-writer cutover)

AgentY-asaf:
## Summary

**SPEC_864 Phase 2 — the load-bearing single-writer change** (per `SPEC_864_LAYOUT_SINGLE_WRITER_2026_06_30.md` §4 Phase 2 and the 2026-07-02 weak-cutover scope note / DISCUSSION §7b): an `OTYPE_LAYOUT` `UpdateObject` push now dispatches **one** `Command::LayoutSetTree` through the srv reducer instead of the legacy `update_raw` whole-row write + separate `SetFocusedNode`/`SetMagnifiedNode` dispatches.

- **One `db_layout` write, one version bump per push** (was two — the "double-write" at `object.rs`)
- **`TabRecord.rootnode` stays authoritative mid-session** — previously a passive shadow that diverged on the first push and stayed stale for the whole session (the Pillar-1 reproject coherence gap)
- New `LayoutClientSlices` (common) carries leaforder / focus / magnify / pendingbackendactions with REPLACE semantics — pushing without processed actions still **clears the backend-action queue** (the frontend's ack path), exactly as the wholesale row write did
- Granular/tree-only `LayoutSetTree` callers (`slices: None`) leave the client columns untouched
- Unowned layout rows and rootnode parse failures fall back to the legacy path **loudly** (`tracing::warn`); a SQLite apply failure compensates by re-dispatching the pre-push row so reducer state rolls back
- **Frontend unchanged** — wire format identical; also updates the stale "4 of 11 arms" scaffolding comment in `reducer.rs`

## Verification

**Unit/integration (all green — `cargo test --workspace`, 1,796 tests):**
- New router-level invariant test `update_object_layout_push_single_write_and_coherent_reducer`: real HTTP push → exactly one version bump, row content matches, `TabRecord.rootnode == db_layout.rootnode`, omitted pendingbackendactions clears the queue
- New reducer tests: slices apply focus/magnify (empty = clear); empty-tree contract wins over skewed slices
- New persist tests: full-row slice write, ack-clears-pending, tree-only write leaves client columns untouched

**App-running E2E (per spec §5 — this is the "sharp edge" phase):**
- `task dev` on this branch; drove `POST /api/v1/pane/open` (backend queues action → frontend processes → pushes tree via UpdateObject → **new reducer route**)
- Persisted row verified via GetObject: new pane present in rootnode, leaforder written (4 entries), `pendingbackendactions: None` (acked), version low/coherent
- **0 "falling back to legacy" warnings**; no layout errors in srv log

⚠️ Per the spec: **do not auto-merge on bot approval** — this is the highest-frequency layout writer. Suggest a human smoke of split/resize/drag + an app restart (persistence) before merge.

Unblocks: SPEC_864 Phases 3–5 (seeder/tear-off reroute, pendingbackendactions reducer-routing, `heal_layout` deletion) → Pillar 1.

### DIFF
--- a/.changesets/1783247349-fix-layout-updateobject-routes-layout-pushes-through-the-red-hjid.md
+++ b/.changesets/1783247349-fix-layout-updateobject-routes-layout-pushes-through-the-red-hjid.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): UpdateObject routes layout pushes through the reducer as a single LayoutSetTree (SPEC_864 Phase 2)
--- a/agentmux-common/src/ipc.rs
+++ b/agentmux-common/src/ipc.rs
@@ -585,10 +585,19 @@ pub enum Command {
     },
     /// Bulk-replace the tree. Used during Phase 7a writer migration and
     /// for tear-off where the whole subtree changes atomically.
+    ///
+    /// SPEC_864 Phase 2 — when the command originates from the frontend's
+    /// full-row `UpdateObject` push, `slices` carries the remaining
+    /// frontend-owned `LayoutState` columns (leaforder / focus / magnify /
+    /// pendingbackendactions) with REPLACE semantics, so the single
+    /// dispatch fully supersedes the legacy `update_raw` whole-row write.
+    /// `slices: None` (tree-only callers) leaves those columns untouched.
     LayoutSetTree {
         tab_id: String,
         new_tree: Option<crate::LayoutNode>,
         correlation_id: String,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        slices: Option<crate::LayoutClientSlices>,
     },
 
     /// Phase D.3 — request an `Event::EventList` reply containing the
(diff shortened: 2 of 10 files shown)


AgentY-asaf:
Addressed the P1: `update_layout_via_reducer` now runs the reducer dispatch AND the `apply_event_to_wstore` SQLite write (plus the failure-path rollback re-dispatch) inside a single `srv_state` critical section, so persist order always equals dispatch order — the B→A inversion window is gone. The `db_layout` row UPDATE is a sub-millisecond local write; the doc comment records why this route deliberately deviates from `dispatch_to_reducer`'s release-before-I/O contract. Full workspace suite green (1,796 tests).

AgentY-asaf:
Review-3 P1 addressed: the owned-row parse-failure fallback now dispatches the focus/magnify slice to the reducer after the legacy write succeeds — exactly the pre-Phase-2 Option-A behavior for that branch (with the #632 write-then-dispatch ordering). New router-level test `update_object_layout_parse_failure_falls_back_with_focus_dispatch` covers it: malformed rootnode → wholesale raw write + reducer focus updated. Full srv suite green.

AgentY-asaf:
Review-4 P1 addressed: the rollback snapshot (`old_row`) is now read INSIDE the `srv_state` critical section, immediately before the forward dispatch — a concurrent same-tab push can no longer commit between the snapshot and the lock, so an apply-failure rollback can never restore stale state over a newer committed push. Full srv suite green (1,305 tests).

PR #1975, opened 2026-07-06
fix(agent): smooth the running -> completed tool-preview transition

AgentY-asaf:
## Summary

When a running tool call finishes, its preview jerks — an abrupt layout shift instead of a smooth transition — because the DOM shape rendered for `running` differs from the terminal-state shape, and nothing bridges the two. Full root-cause analysis: `docs/analysis/ANALYSIS_TOOL_PREVIEW_RUNNING_TO_COMPLETED_JERK_2026_07_05.md`.

**Root cause:** `mergeReplacement()` (`reducer.ts`) flips `status` away from `"running"` *and* `log.open` to `false` in the same tick, so two independent pieces of UI swap DOM shape simultaneously with zero transition anywhere.

## Fix

**Summary row** (`ToolBlock.tsx` / `_document-nodes.scss`) — CSS only:
- Shared `agent-tool-summary-fade-in` keyframe applied to `.agent-tool-result-pill` (mounts the instant status leaves running) and `.agent-tool-live-tail` (unmounts the same tick) — softens the pop instead of an instant appear/disappear.
- `transition: color 150ms` on the status icon so its warning→success/error color change eases.
- Same fade-in applied to `ToolBlockOverlay`'s overlay header (shown for non-success terminal states).
- All gated behind the existing `respect-reduced-motion` mixin.

**Panel body** (`ToolOverlayLog.tsx`) — the bigger one:
- The exclusive `<Switch>` swaps `ChunkList` (streaming) ↔ `ToolOverlayResult` (terminal) — two structurally different component trees with different natural heights, currently zero transition.
- Added a FLIP-style height transition: a `branch()` accessor mirrors the `<Switch>` conditions exactly (kept as a plain function, not `createMemo`, per this file's own documented reactivity caveat around `props.node.log?.chunks`). An effect keyed off `branch()`/`chunks()` reuses its **own previous invocation's** measured height as the FLIP "before" state — the only way to get it, since Solid effects only ever see the DOM *after* it's already patched for the triggering change. Freezes the box at that height, forces a reflow, then eases to the freshly-measured new height.
- Skips entirely while the panel is `content-visibility: hidden` (same layout-forcing hazard the existing auto-scroll effect already guards against) or when `prefersReducedMotionAtom()` is set.

## Tests

5 new tests in `ToolOverlayLog.test.tsx` covering the FLIP mechanics directly (mocking `scrollHeight` per branch since jsdom has no real layout engine): freeze→ease→transitionend-cleanup, no-op on mount, no-op when the branch doesn't actually change, no-op when heights already match, no-op under reduced motion.

- `npx tsc --noEmit` — clean
- `npx stylelint` on the touched scss file — same 27 pre-existing errors as baseline (confirmed via `git stash`), zero new ones introduced
- Full frontend suite: **107 files / 1599 tests, all green**

### DIFF
--- a/.changesets/1783303965-fix-agent-smooth-the-running-completed-tool-preview-transiti-1f8w.md
+++ b/.changesets/1783303965-fix-agent-smooth-the-running-completed-tool-preview-transiti-1f8w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): smooth the running -> completed tool-preview transition (fade-in summary elements, FLIP height transition on panel body)
--- a/docs/analysis/ANALYSIS_TOOL_PREVIEW_RUNNING_TO_COMPLETED_JERK_2026_07_05.md
+++ b/docs/analysis/ANALYSIS_TOOL_PREVIEW_RUNNING_TO_COMPLETED_JERK_2026_07_05.md
@@ -1,0 +1,195 @@
+# ANALYSIS: Tool preview "jerk" on running → completed transition
+
+**Date:** 2026-07-05
+**Status:** Root causes confirmed by direct file inspection (see citations). Not previously
+documented — `SPEC_TOOL_PREVIEW_SCROLL_CHAINING_2026_07_03.md` describes a different, already-fixed
+bug (native scroll-chaining dead zone / double-scrollbar); its own addendum explicitly says "Not a
+jerk, a dead zone" — that "jerk" refers to browser scroll-bounce feel, not this layout reflow.
+
+---
+
+## Problem (as reported)
+
+In the Agent pane, when a running tool call finishes, the tool-call preview visibly jerks — an
+abrupt layout shift rather than a smooth transition — because the DOM shape rendered for the
+`running` state differs from the DOM shape rendered for the terminal (`success`/`failed`/etc.)
+state, and nothing bridges the two.
+
+---
+
+## Root cause: one state update flips two independent DOM shapes at once, and neither is transitioned
+
+The `running`→terminal transition is driven by a single object replacement,
+`mergeReplacement()` (`frontend/app/store/agent-document/reducer.ts:659-679`). When a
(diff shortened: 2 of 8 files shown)


AgentY-asaf:
Both findings addressed:

**P1 (hidden-panel FLIP resync)** — the height-FLIP effect now records the logical branch even while \`content-visibility:hidden\` (cheap, no DOM read — just tracks which branch, not the height) and marks the measured height as stale. A failed/denied/canceled tool auto-collapses the instant it leaves \`running\`, so the branch change commonly happens entirely while hidden; without this, the first re-measurement after the user expands the panel to inspect the error would FLIP from the stale pre-collapse height instead of just showing the final state. New test: \`resyncs without animating when a branch change happened entirely while the panel was hidden\`.

**P2 (resize retriggering the pill fade-in)** — moved the fade-in off a persistent \`animation:\` on \`.agent-tool-result-pill\` (whose \`display\` also toggles via an unrelated \`@container\` width breakpoint in \`_responsive.scss\`, independent of tool status) onto a one-shot \`.agent-tool-fade-in-once\` class added by a \`ref\` callback that only fires at genuine Solid mount — so resizing across that breakpoint never replays it for an already-completed, unrelated tool call.

Writing the regression test for P2 surfaced a **second bug my own fix introduced**, not flagged by reagent: Solid's \`ref\` callbacks fire at element creation, *before* the effect for a dynamic \`class={...}\` expression runs (that effect sets \`className\` wholesale) — a synchronous \`classList.add\` inside the ref was being clobbered immediately, making the fade-in a complete no-op. Fixed by deferring the classList mutation a microtask so it runs after Solid's own class effect commits.

Full suite green: 107 files / 1602 tests. \`tsc --noEmit\` clean. Stylelint on the touched scss file: same 27 pre-existing errors as baseline, zero new.

(Apologies for the mangled commit message on the fix commit — a shell backtick-quoting mistake ate some inline code spans from the text; the actual diff is untouched, confirmed via \`git show\`.)

PR #1979, opened 2026-07-06
feat(tabs): active tab's custom color traces the tab strip's boundary line

Agent3-asaf:
## Summary
- Extends `SPEC_TAB_CONTENT_FOLDER_SURFACE_2026_06_03.md`'s existing boundary-line mechanic (a hairline under the tab strip, absent under the active tab) with a colored variant: when the active tab has a custom `tab:color`, that color now traces a 2px line under `.tab-bar-scroll` — a persistent, colored indication of which tab you're on.
- Two earlier approaches tried live and rejected during iteration with the user:
  - Coloring `TabContent`'s root (the content area) via a top-only inset shadow, then a full 4-side inset frame — too visually intense, didn't render on the right edge (that root's width didn't cleanly match the intended span), and competed with the existing per-pane focus border.
  - Landed on `.tab-bar-scroll` specifically (not `.tab-bar`, which also contains the hamburger) so the line spans exactly the tab strip's own width — right of the hamburger, left of the header widgets — matching the folder-boundary line's existing geometry instead of the content area's.
- Reactive to both which tab is active *and* that tab's own color changing while it stays active, via the same two-level `createMemo(atom)` pattern `tabcontent.tsx` already uses for the same WaveObject.

## Test plan
- [x] `npx vitest run` — 1594/1594 tests pass (one unrelated file, `tool-renderers/registry.test.ts`, showed a full-suite-contention timeout; confirmed passing in isolation, untouched by this change)
- [x] Manual, live in `task dev`, iterated directly with the user across three approaches before landing on this one: confirmed the line follows the active tab, updates when switching tabs, spans the correct region (hamburger to widgets), and doesn't visually compete with the per-pane focus border

### DIFF
--- a/.changesets/1783333377-feat-tabs-active-tab-s-custom-color-traces-the-tab-strip-s-b-05dk.md
+++ b/.changesets/1783333377-feat-tabs-active-tab-s-custom-color-traces-the-tab-strip-s-b-05dk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(tabs): active tab's custom color traces the tab strip's boundary line
--- a/frontend/app/tab/tabbar.scss
+++ b/frontend/app/tab/tabbar.scss
@@ -24,11 +24,6 @@
     min-width: 0;
     height: 100%;
     overflow: hidden;
-    // Needed for .active-tab-color-line, an absolutely-positioned sibling of
-    // .tab-bar-scroll — deliberately NOT a child of .tab-bar-scroll itself,
-    // since that's the horizontal scroll container and an absolute
-    // descendant of it moves with its scrollLeft.
-    position: relative;
     // Drag cascades from window-header; individual tabs/buttons opt out
 
     .tab-bar-scroll {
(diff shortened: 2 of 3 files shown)


Agent3-asaf:
Good catch, confirmed live: the line was a child of `.tab-bar-scroll` (the scroll container), so scrolling it by 500px shifted the line's `getBoundingClientRect().x` by exactly -500px — the persistent-boundary effect broke completely once the strip overflowed.

Fixed in d488cf12 — moved the line to be a sibling of `.tab-bar-scroll` (both children of the never-scrolling `.tab-bar`), positioned via measured `offsetLeft`/`clientWidth` rather than CSS. Re-verified live: identical rect before/after a forced 500px scroll (both `x: 39`), and still exactly aligned to the first tab's left edge at rest.

Agent3-asaf:
- **P1**: already moot — d1e990d9 (pushed before this review landed) switched the right boundary to the viewport edge unconditionally, so there's no more platform-conditional stop point to undershoot.
- **P2**: fixed in 7aed9125 — removed the dead `position: relative` and its now-inaccurate comment; the line hasn't been a descendant of `.tab-bar` since the Portal move. Verified live the line's rect is unchanged without it.

PR #2027, opened 2026-07-08
feat(jekt): Phase 3 — JektBubble renders [JEKT:...] markers in the agent pane

AgentY-asaf:
## Summary

SPEC_JEKT_SECURITY_AND_VISIBILITY_2026_07_01.md's Phase 1 (marker injection, commit 37d64dc8) and Phase 2 (CLAUDE.md security rules) shipped 2026-07-01. Phase 3 — the frontend rendering — was never picked up, so jekt messages have been rendering as raw `[JEKT:FROM=... TIER=... ...]` marker text mixed into the conversation ever since, instead of the distinct labeled block the spec calls for (design goals G1/G2).

- `stream-parser.ts`: detects a `[JEKT:...]...[/JEKT]` block occupying a `user_message` payload, parses `FROM`/`TO`/`TIER`/`DELIVERY`/`TRUST`/`MSGID`/`PRIORITY`, strips the divider/header/reply scaffolding down to the actual message text, and emits a `JektMessageNode` instead of a plain `UserMessageNode`. Falls back to plain text for anything malformed or partial, so an unrecognized wrapper variant degrades gracefully instead of losing content.
- `components/JektBubble.tsx` (new): collapsed-by-default row — direction icon, sender/recipient, tier badge, delivery-tier badge. Expands on click to show the message body, metadata (MSGID, trust, priority, timestamp), and a raw-payload `<details>` — spec §3.3.
- `DocumentRow.tsx` / `expansion-source.ts` / `renderers.ts`: wire the new node kind through the (now-exhaustive) per-kind switches — expansion default, height estimator, streaming-capability flag, renderer-registry slot.
- `_document-nodes.scss`: badge/border styling matching the existing `.agent-message-block` conventions — accent border by direction, red border for `tier-sensitive`.

## Scope note

Outgoing/right-aligned rendering (spec §3.2, the sender-side echo) is implemented in the parser/component but untested against real data — no backend path emits an outgoing echo today, so only the incoming path is actually exercised in production.

## Test plan
- `npx tsc --noEmit` clean.
- `npx vitest run frontend/app/view/agent` — 638/638 passing, including 8 new jekt-detection cases in `stream-parser.test.ts` (well-formed parse, envelope stripping, incoming/outgoing direction, unknown-tier defaulting, malformed/unterminated fallback).
- `npx stylelint` on the touched SCSS file — the new rules introduce zero lint errors (all pre-existing errors in that file are outside the added block).

### DIFF
--- a/frontend/app/view/agent/components/JektBubble.tsx
+++ b/frontend/app/view/agent/components/JektBubble.tsx
@@ -1,0 +1,92 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * JektBubble — agent-pane rendering of a muxbus jekt message.
+ *
+ * Replaces the raw `[JEKT:FROM=... TIER=... ...]` marker text with a
+ * distinct, labeled block so the human operator can always tell a jekt
+ * apart from a typed user message or agent output (design goals G1/G2).
+ *
+ * Collapsed by default, like AgentMessageBlock: a one-line summary showing
+ * direction, sender/recipient, and the tier + delivery badges. Clicking
+ * expands to show the message body plus metadata (full MSGID, timestamp,
+ * raw marker payload) — spec §3.3's "click the bubble shows metadata".
+ *
+ * Spec: docs/specs/SPEC_JEKT_SECURITY_AND_VISIBILITY_2026_07_01.md §3.3.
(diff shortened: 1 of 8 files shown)


AgentY-asaf:
Closing to reopen under the correct agent identity per operator request.

PR #2042, opened 2026-07-09
fix(bashwrap): suppress orphaned console window on PTY-fallback pipe path

AgentY-asaf:
## Summary

Diagnosed a live issue: dozens of orphaned, blank ("ghosted") console windows accumulating in the taskbar across this shared multi-agent machine, traced to `agentmux-bashwrap`.

`run_via_pipes` (`agentmux-bashwrap/src/bash_wrap.rs`) — the safety-net path used only when PTY allocation fails — spawned `bash.exe` via a plain `tokio::process::Command` with no `CREATE_NO_WINDOW` flag. `bash.exe` is a console-subsystem binary, so every fallback exec popped a new, visible console window that was never suppressed.

`SPEC_ELIMINATE_BASHWRAP_CONSOLE_WINDOWS_2026_06_20` (#1621) already fixed two spawn sites — bashwrap's own process (`windows_subsystem = "windows"`) and the persistent Shell tool (`shell_node.rs`, `CREATE_NO_WINDOW`) — but missed this one. It's an inner spawn bashwrap does of its own accord, only reachable when the primary PTY path fails, so it's easy to miss unless that fallback is actually exercised during verification. Confirmed via `~/.agentmux/logs/agentmux-launcher.log` ("PTY allocation failed — falling back to pipes") and via `Get-CimInstance Win32_Process`, which showed multiple `bash.exe` processes (some nearly 24h old, near-zero CPU) with `agentmux-bashwrap.exe` as an ancestor.

## Fix

Same pattern as `shell_node.rs`: `creation_flags(CREATE_NO_WINDOW)` on the child `Command`, gated `#[cfg(windows)]`, no behavior change on other platforms.

## Test plan

- [x] `cargo check -p agentmux-bashwrap` clean

### DIFF
--- a/.changesets/1783561904-fix-bashwrap-suppress-orphaned-console-window-on-pty-fallbac-sesf.md
+++ b/.changesets/1783561904-fix-bashwrap-suppress-orphaned-console-window-on-pty-fallbac-sesf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(bashwrap): suppress orphaned console window on PTY-fallback pipe path
--- a/agentmux-bashwrap/src/bash_wrap.rs
+++ b/agentmux-bashwrap/src/bash_wrap.rs
@@ -663,6 +663,12 @@ async fn run_via_pipes(
     // rarely-exercised inner spawn). Same fix as shell_node.rs.
     #[cfg(windows)]
     {
+        // `cmd` is `tokio::process::Command`, which provides `creation_flags`
+        // as a native inherent method on Windows — unlike `std::process::
+        // Command`, no `use std::os::windows::process::CommandExt` is needed
+        // to call it here. Verified via a clean `cargo check --target
+        // x86_64-pc-windows-msvc` (0 errors) and this crate's windows-latest
+        // CI job, both passing without the import. See PR #2042 discussion.
         const CREATE_NO_WINDOW: u32 = 0x0800_0000;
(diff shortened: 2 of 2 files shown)


AgentY-asaf:
Re: the P0 finding above (missing `use std::os::windows::process::CommandExt`) — investigated and this is a false positive, not fixing.

`tokio::process::Command` (which is what `cmd` is here — see the `use tokio::process::Command;` import at the top of this file) provides `creation_flags` as a native inherent method under `#[cfg(windows)]`; unlike `std::process::Command`, it does not require importing `std::os::windows::process::CommandExt` to call it. Confirmed three ways:

1. `cargo clean -p agentmux-bashwrap && cargo check -p agentmux-bashwrap --target x86_64-pc-windows-msvc` on a real Windows host — 0 errors.
2. Experimentally added the import anyway: it compiles, but rustc flags it `warning: unused import: 'std::os::windows::process::CommandExt'` — i.e. the compiler itself confirms the method resolves without it.
3. This PR's own CI already proved it independently: `check --tests + test (windows-latest)` on this exact branch passed (https://github.com/agentmuxai/agentmux/actions/runs/28988491353/job/86022861976).

(The reference sites the review cites, `shell_node.rs`/`sidecar.rs`, do add the import — possibly redundant there too, out of scope to dig into for this PR. Not adding a dead import here just to mirror that.)

No code change needed. Merging once this is acknowledged.

PR #2048, opened 2026-07-09
fix(window): end the startup window storm — single-run initApp + slow-path reproject GC

AgentA-asaf:
## The regression (user-visible)

Both stable-channel instances open **dozens of windows on every launch** since v0.51.x. The earlier fixes (#2017/#2032 deferred closes, the 20-window safety cap) did not stop it — they bounded a single boot to 20 windows while the store kept **growing ~20 rows per boot**, so every restart storms again, forever.

Live evidence (this machine, 2026-07-09, stable channel):
- v0.51.2: `Client.windowids` grew **24 → 43 → 62 → 82** across four boots; every boot logged the cap warning and recreated 20 windows.
- v0.52.0: same pattern, 25 → 45 after one boot.
- srv log shows **40 `window.CreateWindow` calls per boot for 20 reprojected windows** (2 per window), and 42 of the 82 persisted window rows have **no `kind`** (never registered).

## Root cause — two bugs compounding

**1. `initApp` runs twice per page in CEF (the fuel).**
`app-init.ts` has a module-level "dev-only" self-start guarded by `!isHostApp()`. `isHostApp()` reads `window.__AGENTMUX_IPC_PORT__`, which `setupCefApi()` only sets *later*, at bootstrap time — so at module evaluation the guard is **always** false in CEF and the self-start always schedules a second `initApp` alongside `bootstrap.ts`'s own call. Each non-main window therefore calls `WindowService.CreateWindow` twice. Only one row ever completes `registerBackendWindow` (and `initWaveWrap`'s `savedInitOpts` latch hid the double-run from the UI), stranding one **never-registered orphan row + `Client.windowids` entry per window creation**. Verified live: the startup-bench timeline shows `fonts-ready`/`isMainWindow-start` firing twice ~2ms apart within a single page load, and 40 `[initTauriNewWindow] Creating new backend objects` lines for 20 windows.

**2. The Step-4 slow-path reproject resurrects those orphans as real windows (the match).**
`reproject_from_srv` runs on every cold boot and defaulted rows with no persisted `kind` to `FullInstance` ("no persisted kind (pre-Step-3 window row?)" warning — that warning fired once per storm window). A kind-less row cannot have completed a `register_backend_window` round trip (the Step-3 write-through stamps `kind` on **every** registration, main included) — it is an orphan the user never saw, not a restorable window. Each resurrected orphan then registered a fresh row (+1) while the deferred close removed only the old one (−1), and the resurrected window's own double-init leaked another orphan — hence the +20/boot growth and a loop that never converges.

## The fix

1. **`frontend/app-init.ts`** — latch `initApp` on a single promise so both callers (bootstrap and the self-start) await the same run. One `CreateWindow` per window, no more orphan rows.
2. **`agentmux-cef/src/commands/window/creation.rs`** — the slow path now **garbage-collects** kind-less rows via `CloseWindow` instead of recreating them: the windowids entry is pruned, the row deleted, and the orphan's empty auto-created workspace cascade-deleted (the existing "any other window references this workspace" guard protects shared workspaces, e.g. a tear-off twin). Polluted stores **self-heal on the first launch of a fixed build** — no manual DB surgery needed in the field. The 20-window safety cap is kept and now applies to the genuinely-restorable set only.

`main` is excluded from the slow path by value (as before) and additionally always has `kind` stamped, so it can never be GC'd.

Trade-off: genuine pre-Step-3 window rows (created ≤ v0.50, before topology persistence existed) also lack `kind` and are now GC'd rather than restored. Given they cannot be restored with correct kind/parent anyway and the alternative is this storm, deletion-with-loud-log is the safer behavior.

## Verification

- `cargo check -p agentmux-cef` clean; all 184 crate unit tests pass; `tsc --noEmit` clean; full vitest suite 1653 passed.
- **Live smoke (fixed build, `task dev`, seeded with a copy of the polluted stable DB — main + 42 orphan rows, `Client.windowids` = 43):**
  - `[reproject] slow path: garbage-collecting never-registered window rows count=42`
  - `[reproject] slow path: recreating windows from srv window_count=0` — **zero windows opened**
  - srv received exactly 42 `window.CloseWindow` calls; post-boot DB: `db_window` 43 → **1** (main only), `windowids` 43 → **1**, `db_workspace` 82 → 40 (exactly the 42 orphan workspaces cascade-deleted; the guard kept every still-referenced one).
  - Fix 1 verified in the same build: the startup-bench timeline shows `isMainWindow-start` firing **once** (previously twice ~2ms apart within one page load).

Fixes the storm reported today on both stable instances. Follow-up candidates (separate PRs): remove the CEF-dead self-start block entirely; srv-side rejection of `CloseWindow` no-op divergence (reducer silently returns no events for unknown windows, which hid this class of leak).

### DIFF
--- a/.changesets/1783632173-fix-window-end-the-startup-window-storm-single-run-initapp-r-6wsn.md
+++ b/.changesets/1783632173-fix-window-end-the-startup-window-storm-single-run-initapp-r-6wsn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(window): end the startup window storm — single-run initApp, reproject GCs never-registered rows
--- a/agentmux-cef/src/commands/window/creation.rs
+++ b/agentmux-cef/src/commands/window/creation.rs
@@ -693,7 +693,7 @@ pub(crate) fn reproject_from_srv(state: &Arc<AppState>, main_window_id: String) 
         // translated to the `"main"` sentinel here — otherwise the remap
         // lookup finds nothing and the subwindow is silently skipped as if
         // its parent were missing.
-        let mut extra_ids: Vec<&String> = window_ids.iter().filter(|id| *id != &main_window_id).collect();
+        let extra_ids: Vec<&String> = window_ids.iter().filter(|id| *id != &main_window_id).collect();
         if extra_ids.is_empty() {
             tracing::debug!(
                 target: "reproject",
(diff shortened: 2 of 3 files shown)


AgentA-asaf:
Addressed the reagent P1 in 51480e462: added `MAX_SLOW_PATH_SCAN` (200), applied to `extra_ids` **before** the topology-fetch/GC loop — per-boot blocking round trips are hard-bounded again (≤200 GetWindow + ≤200 CloseWindow worst case), independent of how polluted `Client.windowids` is. Entries beyond the budget are left untouched for the next launch, so the store still converges across boots. The scan budget is deliberately larger than `MAX_SLOW_PATH_RECREATE` (20, unchanged — still bounds windows actually opened) so a storm-polluted store (60+ orphans per instance observed live) self-heals in one or two launches rather than 4+.

PR #2056, opened 2026-07-10
fix(identity): App API link-table ops resolve S1 slug to definition id (#1624 PR-C precursor)

AgentY-asaf:
## Summary

Fixes the `identity.account.upsert` bug agent3 reported while self-linking a GitHub identity account, plus the wider blast radius found while verifying it. Part of the #1624 PR-C track (folds agent3's findings in ahead of the main PR-C write-path deprecation).

`db_agent_identity_links.agent_id` stores the agent **definition id** (`AgentDefinition.id` — see m0013 and `identity/resolver.rs`, #2041). But every App API identity handler passed `req.agent_id` — the S1-authenticated **slug** (`AGENTMUX_AGENT_ID` / `bus:register` id) — straight into the link-table calls. Failure mode depends on which store `id_store` routes to:

- **Per-channel fallback** (0011 shared-store backfill not applied): the links table carries `FOREIGN KEY (agent_id) REFERENCES db_agent_definitions(id)` → every upsert fails loudly with `FOREIGN KEY constraint failed` (agent3's repro).
- **Shared store**: its links-table variant has **no** agent_id FK → the write "succeeds" but keys the row by slug, which the resolver (reading by definition id since #2041) can never match. Silently broken instead of loudly.

Reads had the mirror-image bug — listing links by slug always returned empty — so `identity.self.accounts` returned nothing, the upsert **ownership check always rejected updates** (any `account_id`-carrying call got FORBIDDEN), `identity.account.validate`'s stored path always rejected, and `identity.self.unlink` always no-oped. All four surfaces were effectively unusable for normal (slug-authenticated) agent callers.

## Fix

- New `resolve_agent_definition_id()` (`app_api/mod.rs`): slug → definition id via `instance_get_by_name` (the codebase's established slug-lookup convention, cf. `native_memory_handlers.rs`), with passthrough for callers that already hold a definition id (`agent_def_get` verified) so internal non-S1 callers of the shared `*_impl` helpers stay valid. Unknown ids error instead of writing resolver-invisible rows.
- Applied in: `identity.account.upsert` (ownership check + unlink + link), `identity_self_accounts_impl`, `identity_account_validate_stored_impl`, `identity.self.unlink`.
- **Orphaned-row cleanup** (agent3's second finding): a failed link during upsert now deletes the just-upserted `db_identity_accounts` row for new accounts — previously only the keychain secret was compensated, stranding an orphaned unlinked account row per failed call.
- Change events (`agentidentities:changed:*`) keep the slug key — no subscribers exist yet, and callers know themselves by slug.

## Test plan

- [x] New test `resolve_agent_definition_id_maps_slug_and_passes_through_def_id` (slug→def-id, def-id passthrough, unknown-id error)
- [x] `cargo test -p agentmux-srv identity` — 101/101 passing

cc agent3 — both findings from your jekt are in here; the OAuth-flow + bind/unbind bundle-write gaps documented in #2041 remain for the main PR-C.

### DIFF
--- a/.changesets/1783663406-fix-identity-app-api-link-table-ops-resolve-the-s1-slug-to-t-n2oq.md
+++ b/.changesets/1783663406-fix-identity-app-api-link-table-ops-resolve-the-s1-slug-to-t-n2oq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): App API link-table ops resolve the S1 slug to the definition id; failed upsert no longer orphans the account row
--- a/agentmux-srv/src/server/app_api/identity.rs
+++ b/agentmux-srv/src/server/app_api/identity.rs
@@ -152,8 +152,16 @@ fn register_identity_account_upsert(engine: &Arc<WshRpcEngine>, state: &AppState
                     return Err(format!("identity.account.upsert: db: {e}"));
                 }
 
-                // Steps 2+4: replace provider link (unlink old, link new).
-                let _ = id_store.agent_identity_unlink(&def_id, &req.provider);
+                // Step 2/4: (re)point the agent's provider link at this account.
+                // agent_identity_link's own `ON CONFLICT(agent_id, provider) DO
+                // UPDATE` already overwrites whatever account_id was linked
+                // before, so no separate unlink is needed for the success path.
+                // A preceding unlink-then-link was here previously (reagent P1 on
+                // PR #2056): now that def_id resolves correctly, an unlink that
+                // succeeds followed by a link that fails would permanently drop
+                // the agent's existing provider link with no compensation (the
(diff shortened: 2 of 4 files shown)


AgentY-asaf:
Fixed in 301c868e — dropped the unlink call entirely rather than adding another compensating delete for the link-fails case. \`agent_identity_link\`'s own \`ON CONFLICT(agent_id, provider) DO UPDATE SET account_id = excluded.account_id\` already repoints an existing link atomically, so the preceding unlink was both redundant on success and unsafe on partial failure. Existing coverage: \`test_agent_identity_link_and_unlink\` (store/tests.rs) already exercises the overwrite-in-place semantics this relies on. Full backend suite green (1452/1452).

PR #2059, opened 2026-07-10
fix(agent): resume real session on pane reopen + hydrate stats bar from history

AgentY-asaf:
## Summary

Reported bug: after fully closing and reopening AgentMux, opening an existing agent pane showed its historical conversation, but the actually-spawned CLI process was a brand-new session with no memory of it, and the token/context stats bar showed wrong/blank numbers.

This closes out the still-open "continuity" half of `docs/retro/retro-cross-channel-conversation-continuity-regression-2026-06-16.md` plus a related, previously undocumented stats-bar bug. Plan: `docs/plans/PLAN_PANE_REOPEN_SESSION_RESUME_AND_STATS_BAR_2026_07_10.md`.

- **`agent_open.rs`**: when opening an agent creates a *new* block (no existing block found in the target tab — the case an app restart or new-tab reopen hits), seed the new block's `agent:sessionid` meta from the shared registry's known `session_id` for that agent. Previously only the picker's explicit "Continue" flow did this; the default open path silently started fresh and orphaned the original conversation.
- **`main.rs`**: run the registry's `session_id` backfill as a best-effort pass on every startup, not just once via the `m0010` migration, so agents created after the migration already ran still get a resumable `session_id`.
- **`parseHistoryLines.ts`**: history replay now surfaces the last `session_end` stats it finds instead of silently discarding them.
- **`useHistoryPagination.ts` / pane-state reducer**: new `ReconcileContextFromHistory` action hydrates the composer strip's context-fill bar from that historical stats payload at mount — seed-only-if-still-default, mirroring the existing `ReconcileTurnActive` pattern, so it never clobbers real live data.

Scoping note: did not wire live per-turn writes of the captured session id back through SQLite (`db_agent_instances` → registry mirror) — that hits real complexity in the consolidated-vs-legacy instance model that the codebase already sidesteps via the file-based backfill. Flagged as a possible follow-up in the plan doc rather than solved here.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 1454 passed
- [x] `npx vitest run frontend/app/view/agent` — 641 passed
- [x] `npx vitest run frontend/app/store/agent-pane-state/reducer.test.ts` — 144 passed (incl. 3 new `ReconcileContextFromHistory` cases)
- [ ] Manual repro: close app, reopen, open an existing agent pane — confirm resume + correct stats (not yet verified against a packaged build)

### DIFF
--- a/.changesets/1783681059-fix-agent-resume-the-real-session-on-pane-reopen-and-hydrate-hl96.md
+++ b/.changesets/1783681059-fix-agent-resume-the-real-session-on-pane-reopen-and-hydrate-hl96.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): resume the real session on pane reopen and hydrate stats bar from history
--- a/agentmux-srv/src/main.rs
+++ b/agentmux-srv/src/main.rs
@@ -423,6 +423,21 @@ async fn main() {
         match registry::Registry::open(root.clone()) {
             Ok(reg) => {
                 tracing::info!(root = %root.display(), "registry: shared agent registry attached");
+                // Best-effort catch-up pass, run every startup (not just the
+                // one-shot m0010 migration): fills session_id for any agent
+                // record that still lacks one, e.g. one named after m0010
+                // already ran. Idempotent — skips records that already carry
+                // a non-empty session_id. Without this, an agent created
+                // after the migration ran would never get a resumable
+                // session_id in the registry, so a cross-tab/cross-restart
+                // open would silently orphan its conversation (the still-open
+                // half of docs/retro/retro-cross-channel-conversation-continuity-regression-2026-06-16.md).
+                if let Some(shared_dir) = root.parent().and_then(|p| p.parent()) {
+                    let filled = backend::session_backfill::backfill_session_ids(&reg, shared_dir);
+                    if filled > 0 {
+                        tracing::info!(filled, "registry: backfilled session_id for cross-channel resume (startup pass)");
+                    }
+                }
                 wstore_raw.set_registry(Arc::new(reg));
                 if let Some(base) = std::env::var_os("AGENTMUX_AGENTS_DIR") {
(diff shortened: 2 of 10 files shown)


AgentY-asaf:
Addressed in 5556848b: `agent_open.rs` now checks whether any other block with the same `agentId` meta currently has a controller registered anywhere in this process (not just the target tab) before seeding `agent:sessionid` from the registry. If the agent is already live elsewhere, the new block falls back to the pre-PR behavior (no resume seed) rather than risking two controllers concurrently `--resume`ing the same provider session.

No dedicated unit test added for this guard — `agent_open.rs` has no existing test harness (it requires a full `AppState`/wstore/event_bus/broker setup), and mocking the global controller registry would need scaffolding beyond this fix's scope. Verified via `cargo check` (clean) and the full backend suite (`cargo test -p agentmux-srv --bin agentmux-srv` — 1454 passed, no regressions).

AgentY-asaf:
Good catch — fixed in 7ece185a. Serialized \`agent.open\` per agent-definition-id with a global \`tokio::sync::Mutex\` held across the whole check/seed/create/register sequence (acquired right after the agent definition resolves, dropped on every return path). Two concurrent opens for the same agent now fully serialize instead of racing past the live-elsewhere check independently; different agents are unaffected. \`cargo check\` clean, full backend suite still 1454/1454.

AgentY-asaf:
Fixed in ac892905 — good catch, that's a real gap for exactly the default persistent-controller case this PR targets. \`parseHistoryLines\` now only updates \`lastSessionStats\` when the \`session_end\` event actually carries \`input_tokens\`/\`output_tokens\`, so a later empty per-turn boundary marker (\`stats: {}\`) can no longer clobber real stats from an earlier turn in the replayed window. Added a regression test (\`does not let a later empty-stats session_end clobber real historical stats\`). \`tsc --noEmit\` clean, full agent-view suite 642/642.

PR #2068, opened 2026-07-10
fix(titlebar): move Icon Only under New Tab, rename to Show Icons Only

AgentA-asaf:
## Summary
- Moved the tab bar's empty-space right-click menu's "Icon Only" toggle from the very bottom (below the conditional Pin Widgets section) to directly under New Tab, and renamed it to "Show Icons Only" (trailing "Only" matches the existing widget:icononly setting key and common UI convention like "Read Only").
- Removed a second, duplicate context menu: right-clicking the gaps between pinned widgets in the tab bar opened its own separate one-item ("Icon Only") popup instead of the unified menu everywhere else — action-widgets.tsx's handleBarContextMenu was stopping propagation and showing its own ContextMenuModel popup rather than letting the click bubble up to window-header's handler. Removed entirely; that area now falls through to the same shared TitleBarContextMenu.

Net result: right-clicking any empty space in the tab bar — including between widgets — opens one single, consistent menu with "Show Icons Only" right under "New Tab".

## Test plan
- [x] npx tsc --noEmit — clean
- [ ] Not click-tested live — small, low-risk menu-consolidation change

### DIFF
--- a/.changesets/1783702443-fix-titlebar-move-icon-only-under-new-tab-rename-to-show-ico-4kz5.md
+++ b/.changesets/1783702443-fix-titlebar-move-icon-only-under-new-tab-rename-to-show-ico-4kz5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(titlebar): move Icon Only under New Tab, rename to Show Icons Only
--- a/frontend/app/window/action-widgets.tsx
+++ b/frontend/app/window/action-widgets.tsx
@@ -528,28 +528,13 @@ const ActionWidgets = (): JSX.Element => {
     }
 
     // ── Context menus ─────────────────────────────────────────────────────────
-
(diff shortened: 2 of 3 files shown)


AgentA-asaf:
Re: the P1 version-bump finding — this looks like a gate false positive, not something to act on. Per CLAUDE.md's changesets workflow, feature PRs add a changeset and do **not** bump `package.json`/`Cargo.toml`/`VERSION_HISTORY.md` — that happens later in a dedicated release PR. This PR touches none of those three files, which is exactly the scope CLAUDE.md documents for when the version-consistency gate should apply ("when reviewing any PR that touches VERSION_HISTORY.md, package.json, or Cargo.toml"). The changeset file (`.changesets/1783702443-...md`) is present and correctly formatted, matching the identical pattern PR #2060 used without triggering this flag.

Not bumping the version here — that would be the wrong action per the documented process.

PR #2074, opened 2026-07-10
fix(tabbar): pane drag over tab bar no longer mis-fires as a tear-off

AgentY-asaf:
## Summary
- Fixes a real, reproducible bug: dragging a pane and releasing it over the tab bar spawns an unwanted floating pane window (the release is misread by `CrossWindowDragMonitor` as a tear-off, since nothing clears `currentDragPayload` for an in-window pane drop over the bar today).
- Adds a `tileItemType` monitor on the tab bar that consumes the drop (clearing the payload) and, piggybacking on the same hit-test, tracks which tab (if any) is currently hovered via a new `hoveredDropTabId` signal + `computeHoveredTab` helper in `tabbar-dnd.ts`.
- Wires `hoveredDropTabId` into `DroppableTab` via a `tile-drop-hover` class — this class and its accent-color pulse keyframe already existed in `tabbar.scss`, fully styled but never referenced anywhere in the codebase; this PR is the first thing to actually use it.
- No visible cross-tab move yet (that's the schematic preview + ghost + backend RPC, tracked as follow-up work) — this PR only fixes the tear-off regression and lights up the tab-hover highlight.

Implements steps 1–2 of `specs/SPEC_PANE_DRAG_TO_TAB_2026_07_10.md`'s implementation order (see #2073).

## Test plan
- [x] `npx vitest run frontend/app/tab/` — 28/28 passing, including 6 new `computeHoveredTab` cases
- [ ] Manual: drag a pane over a sibling tab and release — confirm no floating window spawns, confirm the sibling tab pulses while hovered and stops pulsing over the pane's own current tab

### DIFF
--- a/.changesets/1783723001-fix-tabbar-pane-drag-over-tab-bar-no-longer-mis-fires-as-a-t-br8j.md
+++ b/.changesets/1783723001-fix-tabbar-pane-drag-over-tab-bar-no-longer-mis-fires-as-a-t-br8j.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tabbar): pane drag over tab bar no longer mis-fires as a tear-off, adds tab-hover highlight
--- a/frontend/app/tab/droppable-tab.tsx
+++ b/frontend/app/tab/droppable-tab.tsx
@@ -17,6 +17,7 @@ import {
     insertionPoint,
     setInsertionPoint,
     bouncingTabId,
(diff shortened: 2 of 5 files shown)


AgentY-asaf:
Consolidated into #2073 (spec + implementation in one PR) — cherry-picked this branch's commit there. Closing to avoid duplicate review.

PR #2076, opened 2026-07-11
fix(jekt): echo outgoing SendMessage jekts into the sender's own pane

AgentY-asaf:
## Summary

`SPEC_JEKT_SECURITY_AND_VISIBILITY_2026_07_01.md` §3.2 calls for the host to echo an outgoing jekt back into the sending agent's own pane, symmetric to the incoming `JektBubble` that already renders on the recipient's side (shipped in v0.52.0). That half never actually shipped — `SendMessage`'s tool result was just a bare `"Message sent to X"` string, and the frontend only ever ran jekt-detection inside user-message events, never tool results. Net effect: a human watching the sender's pane saw a generic collapsed `ToolBlock`, nothing distinctive, while the recipient's pane correctly showed a `JektBubble`.

- **`agentmux-srv`**: `wrap_jekt_message` gains a `JektTrailer` (`ReplyHint` vs `DeliveredStatus`). Reusing the identical wrapped string for both directions would tell the sender "Reply: bus:inject to {themselves}" — caught this while designing, so the echo gets its own trailer instead of a byte-for-byte copy of the recipient's block. `InjectionResponse.echo: Option<String>` carries it, `Some` only on the two success paths, `None` on every failure path.
- **`agentmux-mcp`**: `SendMessage` returns `response.echo` as the tool result when present, falling back to the old plain string against an older server (forward/back compat during a mixed-version rollout).
- **frontend**: `tryParseJekt`'s body is factored into `tryParseJektText(text, timestamp?)`, shared by `userMessageToNode` (incoming, unchanged behavior) and a new check in `toolResultToNode`, gated on tool name `"SendMessage"` + `status: "success"` — not "any tool output that happens to look jekt-shaped." The returned node's `id` is forced to `event.id` so it replaces the running tool placeholder in place instead of leaving it stuck alongside an orphaned jekt node.

Full design doc: `docs/specs/SPEC_JEKT_OUTGOING_ECHO_2026_07_10.md`.

## Test plan
- [x] `cargo check -p agentmux-srv -p agentmux-mcp` — clean
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 1467 passed (incl. 4 new: `wrap_jekt_message` trailer variants, echo present-on-success/absent-on-failure)
- [x] `cargo test -p agentmux-mcp` — 4 passed
- [x] `npx vitest run frontend/app/view/agent/stream-parser.test.ts` — 52 passed (incl. 6 new: outgoing echo parses to `jekt_message`, id-replacement, failure/wrong-tool/malformed all correctly fall back to a plain tool node)
- [x] `npx vitest run frontend/app/view/agent` (full suite, regression check) — 656 passed
- [ ] Manual: two agents, `SendMessage` A→B — confirm A's own pane now shows an outgoing `JektBubble` (right-aligned, paper-plane icon per the governing spec §3.3) alongside B's existing incoming bubble

### DIFF
--- a/.changesets/1783731370-fix-jekt-echo-outgoing-sendmessage-jekts-into-the-sender-s-o-z4it.md
+++ b/.changesets/1783731370-fix-jekt-echo-outgoing-sendmessage-jekts-into-the-sender-s-o-z4it.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(jekt): echo outgoing SendMessage jekts into the sender's own pane
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -919,7 +919,14 @@ async fn call_tool(
                 .map_err(|e| anyhow::anyhow!("response parse failed: {e}"))?;
 
             if result.get("success").and_then(|v| v.as_bool()) == Some(true) {
-                Ok(format!("Message sent to {to}"))
+                // Prefer the server's [JEKT:...] echo (SPEC_JEKT_OUTGOING_ECHO_2026_07_10.md
+                // §2.4) so the sender's own pane renders a JektBubble like the
+                // recipient's does. Falls back to the plain confirmation string
+                // against an older server that doesn't send `echo` yet.
+                match result.get("echo").and_then(|v| v.as_str()) {
+                    Some(echo) => Ok(echo.to_string()),
+                    None => Ok(format!("Message sent to {to}")),
+                }
             } else {
                 let err = result
                     .get("error")
--- a/agentmux-srv/src/backend/reactive/handler.rs
+++ b/agentmux-srv/src/backend/reactive/handler.rs
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 use std::sync::{Mutex, OnceLock};
 use std::time::{Duration, Instant};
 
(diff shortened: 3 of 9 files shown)


AgentY-asaf:
Superseded by #2077, which already merged to main and covers the same ground more completely.

#2077's mechanism (`echo_jekt_to_sender` appending the `[JEKT:...]` block to the sender's own blockfile as a `user` line, rendered via the existing `userMessageToNode`/`tryParseJekt` path) fully replaces what this PR's `toolResultToNode` detection + `agentmux-mcp` echo-string change were trying to accomplish — and additionally fixes a bug this PR never touched: persistent (stream-json) agents were writing injected jekts only to stdin, so operators never saw *incoming* jekts either, live or in history replay. It also wires `setAgentId` into production code paths for the first time, which this PR's direction-detection logic depended on but never actually got in real usage (only in my own tests, where I called `setAgentId` manually).

Verified no direct file conflict in `agentmux-mcp/src/main.rs` (untouched by #2077), but merging this as-is would produce a real functional bug: both mechanisms would fire on every `SendMessage` call, rendering two duplicate outgoing jekt bubbles per message.

Closing rather than rebasing/reconciling — #2077's approach is strictly more complete, not just different.

Issue #1569, opened 2026-06-18
macOS/Linux: orphan reconciler can't detect crash-orphans (liveness check hard-coded Live)


a5af:
## Gap

WRR (Window Reality Reconciliation) is Windows-only by design, and that's the right call (see `docs/specs/ANALYSIS_WRR_CROSS_PLATFORM_2026_06_18.md`). But there is **one** cross-platform consequence worth fixing.

On Windows, a renderer/window that dies **without firing `on_before_close`** (a crash-orphan) is caught via WRR's `WM_DESTROY` → `OrphanDestroy` → `HostShouldQuit`. On macOS/Linux there is no such native signal, and the host-side orphan reconciler can't compensate because its per-browser liveness check is **hard-coded `Live`** off Windows:

```rust
// agentmux-cef/src/commands/orphan_reconcile.rs:432
#[cfg(not(target_os = "windows"))]
{
    let _ = host;
    HwndStatus::Live
}
```

Only the Windows arm does a real `IsWindow` probe (#702). So on macOS/Linux a zombie browser is indistinguishable from a live one, and orphaned pool windows / the last-window-closed quit cascade can be missed.

## Proposed fix (targeted, ~tens of LoC — NOT a WRR port)

1. Make the `#[cfg(not(windows))]` liveness arm a **real check** — query `cef::Window` / `BrowserHost` validity (e.g. `has_view`/`is_valid` on the browser host) instead of assuming `Live`.
2. Trigger `reconcile_and_drain` off the already-cross-platform **`on_render_process_terminated`** (`agentmux-cef/src/client/handlers.rs:396`, `client/mod.rs:1491`) so a renderer crash kicks reconciliation without needing a native window-event hook.

This delivers the crash-orphan safety net WRR gives on Windows, on all platforms, without the native-observation machinery (which would be three separate backends and is infeasible on Wayland).

## Out of scope

A full macOS/Linux WRR observation layer (`OffMonitor`/`HiddenSinceOpen`/`LingeringHwnd`) — low value, high cost, Wayland-infeasible. The native window bugs (tear-off ghost-pane, redock hit-test, Wayland pool visibility) are separate native work.

Refs: `docs/specs/ANALYSIS_WRR_CROSS_PLATFORM_2026_06_18.md`, `docs/retro/wrr-design-2026-04-28.md`.

PR #2081, opened 2026-07-11
refactor(lifecycle): Pillar 2 Phase 1 — orphan_reconcile becomes sanitize-then-decide

AgentA-asaf:
## Summary

Phase 1 of `docs/specs/SPEC_PILLAR2_SANITIZE_THEN_DECIDE_2026_07_11.md`. **Stacked on #2080** (Phase 0 — arming bit, `ReconcileQuit` poke, executor extraction); the base is set to that branch so this diff shows only the Phase-1 change. Merge #2080 first.

`orphan_reconcile` stops being an independent quit authority. Its parallel drain predicate — `ReconcilePlan.begin_drain`, computed from launcher-shadow membership plus the Race-B `freshly_promoted` guard — is deleted. The verdict now comes from `reconcile_quit` via a `HostCommand::ReconcileQuit` dispatch, and executes through the shared `ui_tasks::begin_drain_and_cascade`.

## Why the Race-B guard can go (spec §1.A)

Promotion flips `is_pool: false` **synchronously inside the promote dispatch** (`reducer/pool.rs`), so `count_live_user_windows` already counts a freshly-promoted window and `reconcile_quit` refuses to drain while one exists. The guard compensated for `plan_reconcile` reading the *slower* launcher-shadow projection instead of the reducer's own `browsers` map. `freshly_promoted` stays as race telemetry (log line preserved), with no authority.

## Flow (spec §2.1)

probe → plan → **sanitize** (hostless live-user entries unregistered unconditionally — the stale projections that block `reconcile_quit` forever; the old drain-gating was circular: cleanup waited on a decision that read the very registration being cleaned) → **decide** (`ReconcileQuit` poke) → **execute** (Stage-1 cascade *before* zombie closes, preserving the old BeginDrain-before-closes pool-refill suppression; drain-gated hostless-pool cleanup; the two documented direct `quit_message_loop` Stage-2 executions, unchanged in context and rationale).

Dead-zombie closes remain self-sanitizing: `close_browser(force=1)` → `on_before_close` → `UnregisterBrowser` + its own `request_drain` consumption (PR #1993). For user-kind zombies the poke-time verdict is deliberately `None` (they are still counted live); the drain fires from the close chain — same authority, one event later.

Also deleted: the `drainable` live-pool close list — Stage 1 already closes `window-pool-*`/`floating-pool-*` inventory when a drain fires; the planner was duplicating that mechanism.

## Intentional behavioral deltas

- **Lone live floater no longer blocks the orphan-path drain** (old shadow-count included mirrored floaters). Aligns with invariant FP-LIFE, `reconcile_quit`, and WRR's visible count (spec §1.D).
- **Hostless live-user repair is unconditional** rather than waiting for a drainable moment — projection repair is correct at any time and is precisely what unblocks the decision.
- Ordering/refill nuance: with mixed pool-kind + user-kind zombies, a pool zombie's close can now trigger one pool refill spawn before the (close-chain-driven) drain lands; it converges via Stage 1. Pre-#1993 code suppressed this by deciding early from shadow state — accepting one wasteful spawn in a rare crash-orphan path is the cost of removing the second authority.

## Coverage & platform containment

- `classify_hwnd` hard-codes `Live` on macOS/Linux (#1569) → every zombie/hostless branch in this file is Windows-only in practice; macOS/Linux orphan flows run purely through `on_before_close` consumption + Stage-2, untouched here.
- The residual Windows mixed case (user-kind zombies alongside hostless entries, where `client::browser_list` can never empty for Stage 2) is bounded by the #2043 quit watchdog (quit-late-with-loud-log).
- Planner tests rewritten as classification-only (14 tests covering the Dead/Hostless/Live × user-kind × shadow × pool-set × pending matrix); the decision half of the old truth table lives in `reducer/tests.rs`'s `should_begin_drain`/`reconcile_quit` tables (extended in #2080). **182/182 crate tests**, parallel and single-threaded.

## Honest verification note

The live trigger for this path (`Event::HostShouldQuit` from launcher orphan detection) is hard to stage deterministically, and one semantic — "a freshly-promoted window blocks the composed verdict through the reducer count" — cannot be unit-tested CEF-free (registering a browser requires a real `cef::Browser`; the promote-flip and the count predicate are each pinned separately, but not composed). Per the spec's §3 Phase-1 notes this PR leans on: every *mechanism* (probe, close actions, direct-quit contexts, threading) being untouched; the moved truth table; and preserved diagnostics (`freshly_promoted` log, close plans) for field attribution. The close-matrix live verification lands with Phase 2, which exercises the same executor on the dominant (non-orphan) path.

Follow-up to #2080. Refs SPEC_PILLAR2_WIRE_RECONCILE_QUIT §3.3 (orphan_reconcile row).

### DIFF
--- a/.changesets/1783748841-refactor-lifecycle-orphan-reconcile-becomes-sanitize-then-de-gqw4.md
+++ b/.changesets/1783748841-refactor-lifecycle-orphan-reconcile-becomes-sanitize-then-de-gqw4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(lifecycle): orphan_reconcile becomes sanitize-then-decide — drain verdict moves to reconcile_quit
--- a/agentmux-cef/src/commands/orphan_reconcile.rs
+++ b/agentmux-cef/src/commands/orphan_reconcile.rs
@@ -1,27 +1,33 @@
 // Copyright 2026, AgentMux Corp.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Host-side orphan-instance reconciliation. Spec:
-//! `docs/specs/SPEC_HOST_ORPHAN_RECONCILIATION_2026_05_05.md`.
+//! Host-side orphan-instance reconciliation. Specs:
+//! `docs/specs/SPEC_HOST_ORPHAN_RECONCILIATION_2026_05_05.md` (origin),
+//! `docs/specs/SPEC_PILLAR2_SANITIZE_THEN_DECIDE_2026_07_11.md` §2.1 (the
+//! Phase-1 restructure this module now implements).
 //!
-//! When the launcher detects that the last user-visible window
-//! has closed but the host is still alive, it emits
-//! `Event::HostShouldQuit`. The host's handler invokes
-//! `reconcile_and_drain` here, which closes any orphan
-//! `window-pool-*` browsers (promoted out of the warm pool, but
-//! the launcher mirror has dropped them — typically because their
-//! HWND was destroyed without the host's `on_before_close` running).
+//! When the launcher detects that the last user-visible window has closed but
+//! the host is still alive, it emits `Event::HostShouldQuit`. The host's
+//! handler invokes `reconcile_and_drain` here.
 //!
-//! Each close funnels back through `client::on_before_close`,
-//! whose Stage 2 hook fires `quit_message_loop()` once
-//! `browser_list` empties — so the reconciler doesn't drive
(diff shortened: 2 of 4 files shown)


AgentA-asaf:
Addressed the P0: `is_live_top_level_browser`'s `cfg(windows)` gate removed (the body is a pure reducer read; the gate existed only because its sole caller used to be WRR — doc comment updated with the rationale). Cross-platform callers now compile. Commit: 8563a5a8

PR #2082, opened 2026-07-11
fix(lifecycle): Pillar 2 Phase 2 — consume the drain verdict at every count-lowering dispatch site

AgentA-asaf:
## Summary

Phase 2 of `docs/specs/SPEC_PILLAR2_SANITIZE_THEN_DECIDE_2026_07_11.md`. **Stacked on #2081** (base set accordingly; merge order #2080 → #2081 → this).

Since PR #1993 the reducer computes `DispatchOutput.request_drain` after every quit-relevant command — and outside `on_before_close`, every site discarded it. This PR adds `ui_tasks::consume_request_drain` (posts the shared Stage-1 executor as a UI task) and wires it at exactly the sites whose command can flip the verdict to `Some` (i.e., lower the live-user count or clear the last user-pending creation):

| Site | Why it can flip the verdict |
|---|---|
| `unregister_after_parking_close` | The dominant Windows close path (#2043). **`QuitState` now transitions on main-window close for the first time**; Stage-1 pool drain runs. |
| WRR LOCATIONCHANGE recycle-close | #2043 Step A's dispatch — last-window recycle close. |
| `DemotePoolWindow` | The one close flow that lowers the count **without** an `UnregisterBrowser` (kind flip only, browser stays registered for pool reuse). |
| Failed-promote orphan cleanups (both platforms) | Label was already popped+promoted (user-kind); no close chain will ever consume for it. |

Deliberately **not** consuming at `on_after_created`'s Dequeue→Register pair — the gap between them would surface a spurious verdict for a second window opening into an otherwise-empty session. Documented in `quit.rs`'s wiring-status header (refreshed by this PR). This consumption-site analysis is only safe because of #2080's arming bit — without it, a pre-main-registration dispatch could consume a spurious startup drain.

Consumption **posts** rather than calling inline: sites include WINEVENT callbacks and non-UI command threads; `BeginDrain` is monotonic/idempotent so racing consumers (e.g. `on_before_close`'s existing inline consumption) are benign — first wins, rest no-op.

This also closes the original racing-pool-refill obligation from `SPEC_PILLAR2_WIRE_RECONCILE_QUIT` §3.2: a refill that kept the close-time count non-zero gets re-evaluated when its own settling dispatch (demote/unregister) lands.

WRR's own quit gate is intentionally untouched this phase — it remains the live cross-check (the cascade should now win every close race; a scenario where WRR's gate fires with `QuitState` still `Running` = a consumption site we missed). Phase 3 (next PR) demotes it to a Draining-gated executor.

## Test plan

- `cargo check -p agentmux-cef` clean; no new warnings in touched files (`window_pool.rs`'s 3 warnings are pre-existing on base — verified by stash-compare).
- 182/182 crate unit tests.
- **Live close matrix (spec §3 Phase 2 gate) — in progress on a packaged build; results will be posted as a PR comment before merge.** Scenarios: solo main close (expect `[drain-consume]` + Stage-1 before WRR's gate line, clean tree exit); 4 windows close-one-non-last (no drain, no watchdog); sequential close-all (clean exit, no orphan — #1676 regression check); demote-close of last promoted window.

Follow-up to #2080/#2081. Refs SPEC_PILLAR2_WIRE_RECONCILE_QUIT §3.2/§3.3.

### DIFF
--- a/.changesets/1783749343-fix-lifecycle-consume-the-drain-verdict-at-every-count-lower-o807.md
+++ b/.changesets/1783749343-fix-lifecycle-consume-the-drain-verdict-at-every-count-lower-o807.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(lifecycle): consume the drain verdict at every count-lowering dispatch site (Pillar 2 Phase 2)
--- a/agentmux-cef/src/commands/window_pool.rs
+++ b/agentmux-cef/src/commands/window_pool.rs
@@ -761,6 +761,12 @@ pub fn demote_promoted_pool_window(
         ));
         return false;
     }
+    // Pillar 2 Phase 2 (sanitize-then-decide §2.4) — a demote-close is the one
+    // close flow that lowers the live-user count WITHOUT an UnregisterBrowser
+    // (the kind flip excludes the window from the count while the browser
+    // stays registered for pool reuse). When this was the last user window,
+    // the drain verdict fires here.
+    crate::ui_tasks::consume_request_drain(state, &dispatch, "demote_pool_window");
 
     // 3. Park the HWND offscreen + hide, mirroring fresh-spawn state.
     unsafe {
@@ -1431,11 +1437,15 @@ fn cleanup_failed_promote_orphan(state: &Arc<AppState>, label: &str) {
     // inline since CEF won't fire it for this label.
     // Phase H.2.d — legacy `state.browsers.lock().remove` removed;
     // reducer's UnregisterBrowser is sole canonical mutation site.
-    state.host_dispatch(
+    let out = state.host_dispatch(
         crate::reducer::HostCommand::UnregisterBrowser {
             label: label.to_string(),
         },
     );
+    // Pillar 2 Phase 2 — the label was already popped+promoted (user-kind), so
+    // this inline unregister can zero the live count; on_before_close's own
+    // consumption never runs on this path (no CEF close will fire).
+    crate::ui_tasks::consume_request_drain(state, &out, "failed_promote_orphan_cleanup");
     state.window_meta.lock().remove(label);
     crate::launcher_ipc::report_window_closed(label.to_string());
     // Refill (graceful path gets this via on_pool_window_destroyed).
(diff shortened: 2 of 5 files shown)


AgentA-asaf:
Addressed the P1: `consume_request_drain` now runs `begin_drain_and_cascade` **inline** when `cef::currently_on(ThreadId::UI)` (parking-close, LOCATIONCHANGE, and demote sites — WINEVENT callbacks and CloseWindowTask both run on the UI thread), matching the on_before_close pattern, so the cascade always completes before the LOCATIONCHANGE handler's synchronous gate re-run. Posting remains only for genuinely non-UI contexts (promote-failure cleanup), where nothing quits synchronously. The Phase 3 PR (#2083) extends the same fix so its deterministic-quit re-check runs on the inline path too.

AgentA-asaf:
## Live close-matrix results (packaged build `0.52.4+gc9a105e7` = this stack's tip incl. #2083 + both review fixes; isolated instance, full env scrub, CDP-driven)

**Scenario 1 — solo main close: 6/6 PASS.**
`[drain-consume] site=unregister_after_parking_close reason=LastWindowClosed — running begin_drain_and_cascade inline (UI thread)` → `quit_state=Draining` → Stage 1 → `all user windows hidden/closed (registered=0, 0 visible, draining) — quitting message loop (Stage-2 executor)`. Full tree exit code 0 in **255ms**, zero survivors, zero watchdog fires. QuitState transitions on the dominant Windows close path for the first time, exactly as designed.

**Scenario 2 — 4 windows, close one: 4/4 PASS.** The closed window happened to be **main itself** (with 3 others open) — the strongest variant: host stays alive, no drain, no watchdog fire, no quit. (#2043 regression check holds.)

**Scenario 3 — sequential close-all (hermetic fresh channel): 3/3 PASS.** Non-last closes armed the watchdog which stood down cleanly each time (`N window(s) visible again — stand down`); the last close ran the **demote consumption site** (`site=demote_pool_window`, inline) → Draining → Stage 1 closed 5 pool browsers → gated Stage-2 quit 35ms later. Exit code 0, no orphan (#1676 check), no watchdog fire.

**Bonus coverage from a contaminated first run** (shared per-build data dir picked up state from a prior scenario's `taskkill`): Step-4 crash-reproject restored the killed windows (`restoring=1`) and the close-all still ended with **zero survivors and zero orphans**; `orphan_reconcile` fired mid-drain and correctly answered `no drain verdict — already draining` (Phase 1's monotonic decide step, live-verified). One observation from that run for the record: an in-flight pool refill can complete after Draining (`PoolWindowReady` doesn't gate on quit_state), leaving one parked pool browser for the OS-exit path to reclaim — pre-existing behavior, bounded, no user impact.

Not live-verified (as flagged in the PR body): the deliberately-broken-consumption watchdog injection (needs a scratch build; the pure decision is unit-tested) and macOS/Linux runtime (the P0 fix is compile-level; no mac hardware here).

AgentA-asaf:
Superseded by #2084 (same branch rebased onto main after the #2081 stack merge auto-closed this PR — the base branch was deleted before retargeting). All review context above carries over.

PR #2083, opened 2026-07-11
refactor(lifecycle): Pillar 2 Phase 3 — WRR quit demoted to Draining-gated Stage-2 executor

AgentA-asaf:
## Summary

Phase 3 (final wiring phase) of `docs/specs/SPEC_PILLAR2_SANITIZE_THEN_DECIDE_2026_07_11.md`. **Stacked on #2082** (merge order #2080 → #2081 → #2082 → this).

`wrr::win_event`'s happy-path `quit_message_loop()` now requires the reducer to have **decided** to quit: `should_quit_on_last_window(armed, draining, visible, registered)` with `draining = QuitState left Running` (i.e. a `reconcile_quit` verdict was consumed at one of Phase 2's dispatch sites). WRR stops being an independent quit authority and becomes the **Windows Stage-2 executor** — on Windows, parked browsers never fire `on_before_close`, so `client::browser_list` can never empty and the on_before_close Stage-2 gate is structurally unreachable (spec §1.F); something OS-signal-driven must run the final quit, and now it executes only a decision the reducer already made.

- `is_reducer_lagging_os` covers both desync flavors: count disagreement (`registered > 0`, #2043 Step D's original target) and **missing decision** (counts zero but `!draining` — a missed consumption site). Both arm the same watchdog; its quit-on-OS-signal-with-loud-log behavior is unchanged, so the #2043 guarantee holds: any future missed path degrades to quit-late-with-telemetry, never hang-forever.
- Deterministic-quit gap closed: `BeginDrainCascadeTask` re-runs the gate right after `QuitState` flips (new `reevaluate_last_window_quit` wrapper) — a drain with zero pool inventory produces no further OS window events and would otherwise wait out the 3s watchdog.
- The gate reads count + `quit_state` in one `host_state` lock scope (two separate reads could tear).

With this, `reconcile_quit` is the only quit decision-maker in the codebase. Remaining `quit_message_loop` call sites are all executors of its decision: on_before_close Stage 2 (macOS/Linux), the watchdog backstop, `orphan_reconcile`'s two documented direct executions, and this gated WRR executor (spec §6 grep-gate).

## Test plan

- `cargo check -p agentmux-cef` clean; **183/183** crate tests (new: all-zero-without-decision must NOT quit; both lagging flavors arm the watchdog), parallel + single-threaded.
- **Live close matrix (spec §3 Phase 3 gate) — running on a packaged build of this branch (validates Phases 2+3 together); results will be posted as a PR comment before merge.** Scenarios: solo main close (expect `[drain-consume]` → Stage-1 → gated WRR quit, clean tree exit, no watchdog); 4-windows close-one (all survive); sequential close-all (#1676 orphan check); watchdog path (counts/decision desync → quit-late-with-log, no hang).

Follow-up to #2080/#2081/#2082. Closes the WRR row of SPEC_PILLAR2_WIRE_RECONCILE_QUIT §3.3 ("full retirement of the parallel authority").

### DIFF
--- a/.changesets/1783749719-refactor-lifecycle-wrr-quit-demoted-to-draining-gated-stage--b6vb.md
+++ b/.changesets/1783749719-refactor-lifecycle-wrr-quit-demoted-to-draining-gated-stage--b6vb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(lifecycle): WRR quit demoted to Draining-gated Stage-2 executor (Pillar 2 Phase 3)
--- a/agentmux-cef/src/reducer/quit.rs
+++ b/agentmux-cef/src/reducer/quit.rs
@@ -81,9 +81,16 @@ pub(super) fn handle_confirm_drained(state: &mut HostState) -> DispatchOutput {
 // opening into an otherwise-empty session; the Register lands in the same
 // lock scope of activity and re-blocks it — do not consume there).
 //
-// NOT YET WIRED (Phase 3): `wrr::win_event::maybe_quit_on_last_user_window`
-// still calls `quit_message_loop()` from its own predicate without requiring
-// `QuitState::Draining` — the last independent quit authority.
+// WIRED (Phase 3): `wrr::win_event::maybe_quit_on_last_user_window` now
+// requires `QuitState::Draining` (a decision this module made and a Phase-2
+// site consumed) before its happy-path `quit_message_loop()` — it is the
+// Windows Stage-2 EXECUTOR of this module's decision, not an authority
+// (parked browsers never fire `on_before_close` on Windows, so the
+// on_before_close Stage-2 gate is structurally unreachable there). The quit
+// watchdog remains the bounded backstop for both desync flavors
+// (registered > 0, or counts-zero-but-never-decided) and is the only
+// remaining direct `quit_message_loop` caller outside the documented
+// Stage-2 executors. `reconcile_quit` is the sole quit decision-maker.
 
 /// Background PENDING-CREATION labels — warm-pool tab refills (`window-pool-`),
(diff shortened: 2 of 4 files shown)


AgentA-asaf:
Live close matrix for this stack's tip (including this PR's draining gate) posted on #2082 — headline: solo main close exits in 255ms via `quitting message loop (Stage-2 executor)` with all three signals (`registered=0, 0 visible, draining`); close-all shows the watchdog arming and standing down cleanly on every non-last close and the demote-site consumption driving the final drain. Zero watchdog fires, zero orphans across all scenarios.

AgentA-asaf:
## Final live verification (build `0.53.0+gcd523713` — this branch's code incl. the watchdog re-arm fix)

- **Solo main close**: tree exit code 0 in **257ms** via `[drain-consume] site=unregister_after_parking_close` (inline) → Stage 1 → `quitting message loop (Stage-2 executor)` with `registered=0, 0 visible, draining`. Zero survivors (verified by process enumeration), zero watchdog fires.
- **4 windows, close one**: host alive, no drain, no watchdog, no quit.
- **Zombie reproduction** — the exact preconditions that produced the permanently hung draining corpse on the pre-fix build (`taskkill`ed multi-window session on the same channel → 3 crash-reprojected `restoring=1` windows → sequential close-all): now exits **cleanly, code 0, zero survivors**. Watchdog armed and stood down correctly on every non-last close (real windows visible); final close ran the demote consumption site → Stage-2 executor quit 56ms later. The re-arm path wasn't needed on this run's timing — it remains the bounded backstop for the late-pool-spawn window documented in the re-arm commit.

Rebased onto main post-#2084; `git diff` against the verified tip over `agentmux-cef/src` is empty (code byte-identical). 183/183 unit tests.

PR #793, opened 2026-05-11
feat(overlay): declarative data-pane-overlay auto-clip — close the menu/popover/tooltip airspace bug class

AgentA-asaf:
## Summary

Closes the recurring class of "DOM menu paints under native browser pane" bugs on Windows. Any DOM element tagged with `data-pane-overlay` is auto-discovered by a singleton MutationObserver + per-element ResizeObserver service. The service measures each tagged element's screen rect and feeds it into the same overlay-rect map that `usePaneOverlay()` uses, dispatching the existing `browser_panes_set_overlay_clip` IPC. The host already handles the clipping via `SetWindowRgn(hwnd, RGN_DIFF, ...)`. **No new host API; no migration forced on existing modal callers.**

## What changes

- `frontend/app/platform/pane-overlay.ts` keeps a second `Map<Element, OverlayRect>` for auto-discovered overlays, unioned into the existing rect set in `sendClip()`. Existing `usePaneOverlay()` hook callers are unchanged.
- `frontend/app/platform/pane-overlay-auto.ts` (new) is the auto-discovery service. Watches `document.body` for `[data-pane-overlay]` add/remove, attaches a per-element `ResizeObserver`, sweeps on window resize/scroll. RAF-coalesced.
- `frontend/app-init.ts` starts the service after the SolidJS render.
- `data-pane-overlay` added to:
  - FlyoutMenu main menu Portal div
  - FlyoutMenu sub-menu Portal div
  - `cef-api.ts` JS context menu (the `#cef-context-menu-overlay`'s inner menu element)
  - Tooltip Portal div
  - Popover Portal div

Modal callers (modal-v2, TabModalLayer, action-widgets, etc.) already call `usePaneOverlay()` and continue to work via the existing hook path. They could migrate to the attribute later, but no behavior change here.

## Why an attribute, not a hook?

- Discoverable via `grep data-pane-overlay`
- Hard to forget when adding new overlays — a code reader sees the attribute and knows immediately
- Centralized lifecycle (one observer, one map) vs. N hook calls + N cleanups
- Composes with existing `usePaneOverlay()` — both feed the same sendClip path

## Spec

Detailed design at `docs/specs/SPEC_PANE_OVERLAY_AUTO_CLIP_2026_05_11.md` (currently on PR #792's branch; once #792 merges, the spec is on main).

## Test plan

- [ ] Hover the hamburger ≡ over a browser pane — menu fully covers, no leak-through
- [ ] Open Theme submenu over a browser pane — both clipped
[...]

PR #2098, opened 2026-07-11
fix(pane): flyout menus over browser panes on macOS — occlusion, click routing, themed hide background

AgentO-asaf:
## Problem

On macOS, opening the hamburger (or any FlyoutMenu) over a browser/messenger pane (e.g. Discord/Slack) failed three ways:

1. **Pane painted over the menu** — the menu was invisible under the native pane surface.
2. **Menu items overlapping the pane were unclickable** — the menu drew on top but the native pane swallowed the clicks.
3. **The pane "turned black" while the menu was open** — and occlusion only worked *sometimes* (nondeterministic).

## Root causes & fixes

**1. Menu never registered its occlusion rect on macOS/Linux** (`frontend/app/element/flyoutmenu.tsx`)

`FlyoutMenu`/`SubMenu` relied solely on the `data-pane-overlay` attribute, but the auto-discovery service that watches for it is deliberately gated to Windows (`pane-overlay-auto.ts`, PR #793). Fixed by adding explicit `usePaneOverlay()` calls — the same belt-and-suspenders pattern `MoreDropdown` already uses. The top-level menu body is split into a `MenuBody` component so the hook mounts fresh per open, mirroring `SubMenu`.

**2. Airspace hide compared against a stale, wrong-scale rect** (`agentmux-cef/src/browser_panes.rs`, `state.rs`, `browser_pane/creation_views.rs`)

`SetPaneOverlayClipViewsTask` intersected overlay rects with `OverlayController::bounds()`. On macOS that value is frozen at a DIP-scale creation-time value because CEF Views `set_size`/`set_position` are permanent no-ops on `NativeWidgetMacNSWindow` — the real frame is committed via the raw ObjC `setFrame:` path. The intersection test silently missed real overlaps, so `set_visible(0)` never fired and the pane kept receiving events. Fix: new `AppState.browser_pane_physical_rects` cache mirroring the last committed physical rect (seeded at create, refreshed on resize, dropped on detach); the airspace task reads it instead of `bounds()`.

Verified live: before — `visible:true, pane_w:591` (stale DIP); after — `visible:false, pane_w:1182` (real physical frame).

**3. Occlusion was nondeterministic — a mount-time race** (`frontend/app/platform/pane-overlay.ts`)

`usePaneOverlay` measured its element once at mount, while floating-ui still had the menu parked at its `left:0;top:0` placeholder position. Nothing re-measured on position-only changes (ResizeObserver is silent on moves), so whether the clip landed at the menu's real position was a race between the ResizeObserver's initial callback and floating-ui's async position commit — the long-observed "works sometimes, breaks again" flakiness. Fixed with a style-attribute MutationObserver (mirroring `pane-overlay-auto.ts`).

**4. Freeze-frame instead of black-out** (`browser-view.tsx/.scss`, `pane-overlay.ts`, `ipc.ts`)

macOS has no `SetWindowRgn` equivalent, so the airspace strategy hides the **entire pane window** — which read as the pane "going black". Now: `flushClip` mirrors each dispatched clip to a `pane-overlay-clip-changed` DOM event; the pane view listens, and when an overlay intersects it, captures the pane's last rendered frame (CDP `Page.captureScreenshot` via the existing browser DOM API — the compositor keeps producing frames while the overlay NSWindow is ordered out, verified empirically) and displays it in the placeholder until the overlay clears. The placeholder also gets the theme background as the pre-capture fallback. New `invokeBrowserApi()` helper in `ipc.ts` gives the frontend access to `/agentmux/browser/*` routes.

**5. Pre-existing browser-API bug surfaced along the way** (`browser_api/routes.rs`)

`open_cdp_for_block` failed the *current* request on a stale cached CDP target (pane navigated → target rotated; nothing calls `forget()` proactively) and only self-healed the *next* call. Now retries once in-place after dropping the stale cache entry.

## Not in scope

Routing overlay-clip state through the host reducer was considered and deliberately skipped — the mutation path is already UI-thread-serialized and single-decision-point (`compute_pane_visible`). A fork-side CALayer hole-punch (`agentmuxai/cef`) would give macOS Windows-parity precise clipping and is the real long-term fix for the whole-pane hide.

## Testing

- `npx vitest run frontend/app/platform/pane-overlay-auto.test.ts` — 6/6 pass
- `tsc --noEmit` clean; cargo release build clean
- Live CDP verification on `task dev`: two consecutive open→freeze→close cycles — clip rect correct both rounds, freeze snapshot present at full pane resolution (1182×1404), removed on close, zero capture failures

### DIFF
--- a/.changesets/1783794700-fix-pane-flyout-menus-over-browser-panes-on-macos-occlusion--7xrb.md
+++ b/.changesets/1783794700-fix-pane-flyout-menus-over-browser-panes-on-macos-occlusion--7xrb.md
@@ -2,4 +2,4 @@
 type: patch
 ---
 
-fix(pane): flyout menus over browser panes on macOS — occlusion, click routing, and themed hide background
+fix(pane): flyout menus over browser panes on macOS — deterministic occlusion, click routing, and freeze-frame instead of black-out
--- a/.changesets/1783794700-fix-pane-flyout-menus-over-browser-panes-on-macos-occlusion--7xrb.md
+++ b/.changesets/1783794700-fix-pane-flyout-menus-over-browser-panes-on-macos-occlusion--7xrb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(pane): flyout menus over browser panes on macOS — occlusion, click routing, and themed hide background
--- a/agentmux-cef/src/browser_api/routes.rs
+++ b/agentmux-cef/src/browser_api/routes.rs
@@ -729,19 +729,34 @@ async fn open_cdp_for_block(
     state: &Arc<AppState>,
     block_id: &str,
 ) -> Result<CdpSession, String> {
-    let target = state
-        .browser_api
-        .target_cache
-        .resolve(state, block_id)
-        .await?;
     let debug_port = *state.debug_port.lock();
-    let ws_url = format!("ws://127.0.0.1:{debug_port}/devtools/page/{target}");
-    CdpSession::connect(&ws_url).await.map_err(|e| {
-        // On connect failure the cached target may be stale; drop it
-        // so the next call re-probes.
-        state.browser_api.target_cache.forget(block_id);
-        format!("CDP connect: {e}")
-    })
+    // Two attempts: the cached target id goes stale whenever the pane's
+    // CDP target rotates (navigation-driven process swap, etc. — nothing
+    // proactively calls `forget`), and failing the caller's request just
+    // to self-heal the NEXT one made every first call after a navigation
+    // error out. Attempt 1 uses the cache; on connect failure, forget the
+    // stale entry and attempt 2 re-resolves from a fresh `/json` probe.
+    let mut last_err = String::new();
+    for attempt in 0..2 {
+        let target = state
+            .browser_api
+            .target_cache
+            .resolve(state, block_id)
+            .await?;
+        let ws_url = format!("ws://127.0.0.1:{debug_port}/devtools/page/{target}");
+        match CdpSession::connect(&ws_url).await {
+            Ok(session) => return Ok(session),
+            Err(e) => {
+                state.browser_api.target_cache.forget(block_id);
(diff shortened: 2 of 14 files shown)


AgentO-asaf:
Addressed both review items:

- **Floating-pane hit-test mismatch** (P1 ×3): `onOverlayClipChanged`'s hit-test now derives its rect from `paneRect()` itself (the same inset, DPR-rounded rect that seeds `browser_pane_physical_rects` on the host side) instead of the raw placeholder box — the DOM-side freeze decision and the host-side hide decision now evaluate the identical rect, including the `FLOATER_EDGE_RESIZE_BORDER` inset on `floating-*` windows. `79ff095c`.
- **Version drift / conflict** (P1): merged `origin/main` (v0.53.2) — the one conflict was both branches appending a `<Show>` block at the same place in `browser-view.tsx` (freeze snapshot vs the new loading-brain overlay); resolved by keeping both. `2f03a2ba`. Post-merge: `tsc --noEmit` clean, `cargo check -p agentmux-cef` clean, pane-overlay tests 6/6.

Note the freeze path these fixes touch is now Linux-only — macOS switched to the CALayer hole-punch mask (`ui_tasks/pane_hole_mask.rs`) in `f56fc0ab`/`f49d15a3`, where the pane stays live and no freeze-frame is involved.

AgentO-asaf:
Addressed the post-creation race (P1): `browser_pane_overlay_wnums` is now seeded synchronously at pane creation from the pre-hide ObjC scan's already-resolved windowNumber (`0726a428`), so `apply_pane_overlay_hole_mask`'s lookup succeeds immediately — the whole-pane-hide fallback no longer triggers for menus opened during the window before the deferred `SetPaneBoundsViewsTask` runs. A `0` wnum (scan missed the window) stays unseeded and is filled in by the bounds task's highest-wnum fallback as before; in that residual case the pane is still in its creation-hidden phase, so the hide fallback is a visual no-op. `cargo check` clean.

PR #2100, opened 2026-07-11
fix(tabbar): invert strobe, remove reduced-motion respect app-wide, fix lingering source pane

AgentY-asaf:
## Summary
Field-testing round 2 on the pane drag-to-tab feature (spec addendum A1/A2 updated in the same PR):

- **Strobe revised**: the 20ms×10 top-edge bar was invisible — partly killed by reduced-motion handling, partly too fast. Now the whole destination tab flashes to its **negative** (`filter: invert(1)`) and back, 10 flashes at 50ms intervals (**500ms total**, one-shot), then the steady accent pulse takes over. Pure CSS on `.tile-drop-hover .tab`, restarts on every hover entry.
- **Reduced-motion respect removed app-wide** (product decision): the OS "reduce motion" setting was silently disabling functional motion cues (strobe, drop pulse, insertion indicators). `prefersReducedMotionAtom` is hard `false` (setting plumbing kept for easy revisit), the `respect-reduced-motion` SCSS mixin is a no-op, and all raw `@media (prefers-reduced-motion: reduce)` blocks are removed (modal-layer, control-bar, document, install-modal).
- **Lingering source pane fixed**: after a cross-tab move the pane appeared at the destination but also stayed rendered in the source tab — the backend's queued source `DeleteNode` races the source frontend's debounced persist (the documented `INVESTIGATION_LAYOUT_DEAD_SPACE_STALE_TREE_RESURRECTION` mechanism). `redockDraggedPane` now deletes the source leaf locally immediately after the RPC succeeds (the same frontend-mutates-then-persists contract as in-tab moves); the queued backend delete no-ops idempotently.

## Test plan
- [x] `npx tsc --noEmit` clean; stylelint clean on touched files
- [x] `npx vitest run frontend/app/tab/ frontend/layout/` — 108/108 passing
- [ ] Manual: hover a pane over a tab → the tab visibly inverts/flashes 10× over half a second, regardless of OS animation settings
- [ ] Manual: move a pane across tabs → it disappears from the source tab immediately

### DIFF
--- a/.changesets/1783796014-fix-tabbar-invert-strobe-at-500ms-remove-reduced-motion-resp-j0vv.md
+++ b/.changesets/1783796014-fix-tabbar-invert-strobe-at-500ms-remove-reduced-motion-resp-j0vv.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tabbar): invert strobe at 500ms, remove reduced-motion respect app-wide, source pane no longer lingers after cross-tab move
--- a/frontend/app/element/modal-layer.scss
+++ b/frontend/app/element/modal-layer.scss
@@ -20,9 +20,3 @@
     from { opacity: 0; transform: translateY(2px); }
     to   { opacity: 1; transform: translateY(0); }
 }
-
-@media (prefers-reduced-motion: reduce) {
-    .modal-layer-content {
-        animation: none;
-    }
-}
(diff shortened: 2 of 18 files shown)


AgentY-asaf:
Superseded by #2105 (merged) — that branch forked from this one's tip and carried every fix here forward (plus the root-cause placeholder-container fix, full age-gated prune, WCAG-safe strobe, and diagnostic removal). Merging this now would revert main to the pre-#2105 state. Closing without merging.

PR #2102, opened 2026-07-11
feat(pool): adopt foreign window-{uuid} labels into the warm pool on close (Residual B)

AgentA-asaf:
PR B of `SPEC_POOL_ADOPTION_AND_WINDOW_ROW_CRUMB_2026_07_11.md` (task #15). Follow-up to #2094 (Residual A).

## Problem

Cold-path and tear-off windows (`window-{uuid}` labels, minted whenever the pool is exhausted) could never re-enter the warm pool: `CloseWindowTask`'s demote gate keyed on the `window-pool-` label prefix, so their close fell to park-and-blank — srv state cleaned (post-#2087), but **~100MB of renderer stranded per close**, accumulating over long sessions (the "Known residual" in `CloseWindowTask`'s own comment).

## Design — no relabeling

The spec's fallback design (relabel-at-adopt) proved unnecessary: Pillar 2 already made everything that matters classify pool-ness **by type** — the reducer's `is_pool` flag + `unpromoted`/`queue` sets, "decided PURELY BY TYPE … never by label-prefix" (`reducer/quit.rs`). This PR completes that doctrine instead of fighting it:

- `CloseWindowTask` attempts demote for **every** `window-*` label. Demote's own gates (capacity cap, strict HWND resolution, reducer membership) still apply; a refusal falls through to park-and-blank exactly as before.
- The two quit-path pool enumerations still filtering by prefix go type-based via a new `AppState::pool_side_top_level_labels()`: **Stage 1's drain sweep** (an adopted browser invisible to it would block Stage 2's `browser_list.is_empty()` gate) and **orphan_reconcile's nothing-will-pump inventory** (a miscount of 0 would drive `quit_message_loop` with a live pool browser registered). The pane pool (`floating-pool-`) stays prefix-matched — pane adoption is out of scope.
- `srv_event_bridge`'s pool skip adds the type-based check (adopted parked renderers show the pool boot page; no srv-event consumers).
- `cache_pool_window_view` / `mark_pool_window_renderer_ready` caller-bug guards relax from `window-pool-` to `window-`; the real membership gate is the reducer's `handle_pool_ready`, which idempotently no-ops any label not in `unpromoted`.

## Verification

- New reducer test: foreign-label adopt → ready → **re-promote** cycle + double-adopt idempotency. 189 host tests pass. (One unrelated flake observed once — `allow_pane_focus_once_swap_returns_prev_and_clears`, a shared-static parallel-test race; passes on rerun and on unmodified main.)
- **Live on a packaged build, full cycle**: exhausted the pool (4 rapid opens → 2 pool-served + 2 cold `window-{uuid}`); closed a cold one → close trace shows `demoted back into pool` + srv row closed (HTTP 200) + its CDP target reloaded to the `pool=1` boot URL under the foreign label; a subsequent `openNewWindow` **was served by the adopted label** (instant open, renderer reused, fresh workspace, correct title); close-all then exits cleanly with **zero** leftover processes — the type-based sweep reaps adopted entries.

Remaining from the spec after this: Residual C (non-Windows close-path verification) — needs macOS/Linux hardware.

### DIFF
--- a/.changesets/1783796274-feat-pool-adopt-foreign-window-uuid-labels-into-the-warm-poo-5fhv.md
+++ b/.changesets/1783796274-feat-pool-adopt-foreign-window-uuid-labels-into-the-warm-poo-5fhv.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(pool): adopt foreign window-{uuid} labels into the warm pool on close (renderer reuse instead of park-and-blank)
--- a/agentmux-cef/src/commands/orphan_reconcile.rs
+++ b/agentmux-cef/src/commands/orphan_reconcile.rs
@@ -408,10 +408,16 @@ fn ui_thread_reconcile(state: &Arc<AppState>) {
     // flight and Stage 1 found no pool inventory to close — no CEF lifecycle
     // event will ever advance Stage 2 (stale `client::browser_list` entries
     // we can't see from here may even be blocking it). Drive quit directly.
+    // Type-based tab-pool membership (+ prefix for the pane pool) — same
+    // reasoning as Stage 1's sweep: an ADOPTED pool window's foreign label
+    // (SPEC_POOL_ADOPTION_AND_WINDOW_ROW_CRUMB Residual 1) would make a
+    // prefix-only count read 0 here and drive quit_message_loop with a live
+    // pool browser still registered.
+    let tab_pool_labels = state.pool_side_top_level_labels();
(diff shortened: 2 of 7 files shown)


AgentA-asaf:
P2 addressed in the follow-up commit: the park_and_blank_window doc comment now describes the post-adoption reality (fallback fires only when a demote gate refuses — cap, strict HWND, or reducer rejection — not categorically for foreign labels).

PR #2104, opened 2026-07-11
feat(theme): add a light theme + fix phantom tokens, context menu, and pane title bar theming

AgentY-asaf:
## Summary

Turns `docs/analysis/ANALYSIS_THEME_SYSTEM_LIGHT_THEME_AND_DEPTH_GAPS_2026_07_07.md` into an implementation spec (`docs/specs/SPEC_LIGHT_THEME_AND_DEPTH_FIXES_2026_07_11.md`) and ships it. Terminal penetration (the analysis's dimension 2a) was already shipped separately in PR #2010 and is **not** part of this change.

- **Phantom tokens**: defined 7 CSS custom properties that were referenced via `var(--token, #fallback)` but never actually declared anywhere — every consumer was permanently pinned to its own local fallback. Uses each site's established fallback value, so today's rendering is unchanged across all 9 existing themes.
- **Context menu**: replaced literal Catppuccin Mocha hex values with theme tokens — was the most visible "doesn't theme at all" surface in the app.
- **Pane title bar**: replaced hardcoded `rgba(0,0,0,0.6-0.8)` (would've rendered as an opaque black bar on a light pane) with `color-mix(in srgb, var(--block-bg-solid-color) N%, transparent)` — theme-relative, numerically identical to the old value for the default theme.
- **First light theme** (`themes/light.scss`): follows the existing 8-file pattern. Hover/highlight/button/scrollbar tokens invert polarity (darkening, not lightening, since a lightening tint on a light surface is invisible). Terminal ANSI colors individually darkened for legibility on a light background.
- Registered in `themes/index.scss`, `schema/settings.json`'s `window:theme` enum, and `THEME_OPTIONS`.

**Deferred** (documented in the spec, not this PR): `tab.scss`'s colored-tab white text (separate contrast issue, not light-theme-blocking), Mermaid/Shiki fixed-theme wiring, native/CEF pre-paint flash guards, scattered bare-hex long tail.

## Test plan

- [x] `npx vitest run` — 1677 passed
- [x] `npx vite build` — compiles cleanly; spot-checked the compiled CSS output directly (correct token values; `[data-theme="light"]` sits in the same cascade position, after `:root`, as the already-working `[data-theme="dracula"]`)
- [x] `npx stylelint` on all touched files — clean (theme.scss, context-menu.scss, titlebar.scss); light.scss matches the same pre-existing (non-CI-gated) pattern as all 8 sibling theme files
- [ ] Manual visual verification in the running app — **not done**. No project skill or browser-automation tooling is set up to drive this custom Rust+CEF desktop app, and a full native build was judged out of scope for this change's risk/cost. Recommend a human (or `/run-skill-generator` to build a driver) spot-check the Light theme in Settings before/after merge.

### DIFF
--- a/.changesets/1783800919-feat-theme-add-a-light-theme-fix-phantom-tokens-context-menu-8y91.md
+++ b/.changesets/1783800919-feat-theme-add-a-light-theme-fix-phantom-tokens-context-menu-8y91.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(theme): add a light theme + fix phantom tokens, context menu, and pane title bar theming
--- a/docs/specs/SPEC_LIGHT_THEME_AND_DEPTH_FIXES_2026_07_11.md
+++ b/docs/specs/SPEC_LIGHT_THEME_AND_DEPTH_FIXES_2026_07_11.md
@@ -1,0 +1,59 @@
+# Spec: Light theme + theme-system depth fixes
+
+**Date:** 2026-07-11
+**Author:** Agent1
+**Status:** Phases 1–3 implemented (this PR); Phase 4 (light theme) implemented (this PR); Phases 5–7 not started
+**Related:** `frontend/app/theme.scss`, `frontend/app/themes/`, `frontend/app/components/context-menu.scss`, `frontend/app/block/titlebar.scss`, `schema/settings.json`, `frontend/app/menu/base-menus.ts`
+**Governing context:** `docs/analysis/ANALYSIS_THEME_SYSTEM_LIGHT_THEME_AND_DEPTH_GAPS_2026_07_07.md` (the research pass this spec turns into concrete work), `SPEC_TERMINAL_THEME_PENETRATION_2026_07_07.md` (terminal penetration — separately proposed and already shipped in PR #2010, 2026-07-08)
+
+---
+
+## 1. Summary
+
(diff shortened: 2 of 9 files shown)


AgentY-asaf:
Both fixed in 2c8f7251:

1. \`light.scss\` now overrides \`--error-text-color\`/\`--success-text-color\`/\`--warning-text-color\`, aliased to the theme's own already-contrast-tuned \`--error-color\`/\`--success-color\`/\`--warning-color\` — same pattern \`--accent-text-color\` already uses.
2. \`--input-bg-color\` is now its own literal \`rgba(0, 0, 0, 0.2)\` in \`theme.scss\` (no longer aliased to \`--form-element-bg-color\` → \`--main-bg-color\` → window-opacity-coupled), plus a light-theme-appropriate override. Fixes the editor inline-input see-through-at-low-opacity issue.

Re your "byte-for-byte unchanged" note on \`--error-text-color\` vs PreLaunchAuthPanel.scss's prior \`#ef4444\`: that was a deliberate majority-fallback pick (5 of 6 call sites used \`#ff8080\`), documented as such — not a functional bug, just an imprecise claim in my commit message. No code change needed there; flagging so it's not mistaken for unaddressed.

\`tsc --noEmit\` clean, \`vitest run\` 1677/1677, rebuilt and spot-checked the compiled CSS output directly for both fixes.

Issue #2092, opened 2026-07-11
Launcher-side teardown backstop: force-close J0 when no user window remains (Discussion #1680 §9.4)


AgentA-asaf:
The single undelivered item from Discussion #1680's §9 scorecard, carved out as its own tracker so the discussion can close.

Everything else in the lifecycle-churn program landed: the close path is instrumented and single-routed (#1676, round-6 demote, #2087/#2088/#2089), quit authority is reducer-owned and level-triggered (Pillar 2: #2080/#2081/#2084/#2083, watchdog backstop #2043), and CI + E2E coverage exists (ci-pr required checks, #2088 windowids-baseline suite).

**The gap:** every backstop shipped so far is *host-side* (the #2043 watchdog runs inside agentmux-cef). If the host itself wedges — alive but not pumping, so no `on_before_close`, no watchdog tick — nothing above it notices that zero user-visible windows remain, and the process tree lingers until the user kills it manually. The launcher already owns Job Object J0 and a window-mirror (WRR reducer state) that knows the user-visible window count; a launcher-side rule of the shape

> "mirror says zero user windows for N seconds AND host pipe unresponsive to ping → terminate J0"

would bound the worst case host-independently. Needs care against the isolation invariants (I2/I3: only its own job) and against false positives during legitimate zero-window states (startup, crash-restart gap, pool-only states during reproject — the mirror models these, but the rule must consume that state, not raw counts).

Related: #942 Phase 2 (srv supervision) and its §8 responsiveness ping — a host liveness ping is a shared prerequisite for both.

GenericAgentX-asaf:
Closing — fully implemented per `docs/specs/SPEC_LAUNCHER_TEARDOWN_BACKSTOP_2026_07_11.md`: "Phase 2 implemented 2026-07-16 — armed state machine in `agentmux-launcher/src/teardown_backstop.rs` ... teardown execution in the supervisor's 5s check tick (TerminateJobObject(J0)...)". Confirmed shipped per `docs/status/STATUS_LIFECYCLE_AND_CRASH_ARCHITECTURE_2026_07_16.md`.

PR #2106, opened 2026-07-11
feat(launcher): observe-only UI-thread liveness probe (teardown backstop Phase 1)

AgentA-asaf:
Phase 1 of `SPEC_LAUNCHER_TEARDOWN_BACKSTOP_2026_07_11.md` (in this PR), toward **#2092** — the one undelivered item from Discussion #1680's §9 scorecard — and #942 §8's responsiveness gap.

## Why

Every quit backstop shipped so far runs **inside the host** (the #2043 watchdog, Stage-2 executors, orphan_reconcile's quit drive) and dies with a wedged host. The launcher can't tell a hung UI thread from a healthy one: **pipe traffic is not liveness** — a wedged host's tokio IPC reader keeps pumping happily.

## What

- **Protocol:** `Command::ProbeUiThread { nonce }` (launcher → host, over the host pipe like the saga `IssueCmd::Host` commands) + `Command::ReportUiThreadAlive { nonce }` (host → launcher via the `Report*` path).
- **Host:** the pipe reader posts a `wrap_task!` UI task; the reply is sent **only** from the task's `execute()` on the UI thread. A wedged — or pre-ready (the known `post_task` silent drop) — UI thread never replies. Silence is the signal; deliberately no fallback reply path, no host-side timeout.
- **Launcher:** new `ui_liveness` telemetry module (probe bookkeeping, nonce-matched RTT, `last_alive()` as Phase 2's consumer surface; 3 unit tests). Recording happens at the **transport layer** (`ipc/server.rs`); the reducer arms are deliberate no-ops — this is thread telemetry about the host process, not domain state. 60s prober in `run_windows`'s select loop; first tick delayed one full interval so a booting host isn't a false miss; pipe send failures are logged as transport, never counted as liveness evidence.
- **Observe-only:** nothing consumes the signal yet. Phase 2 (the armed `TerminateJobObject(J0)` state machine — arm on the existing `was_last`/OrphanInstance signals, grace, disarm-on-window-open) lands separately after the probe bakes, mirroring how Step 4 Phase 1 (GetSnapshot, observe-only) de-risked crash-reproject.

## Verification

- Live on a packaged build: probe `nonce=1` answered `rtt=0ms` on a healthy host; after killing the inner host, the **supervised respawn's** UI thread answered `nonce=2` seamlessly across the restart — the prober needs no restart-awareness (the pipe reconnect machinery already owns that).
- Suites: launcher 202 (3 new), host 189, all green. Exhaustive-match sites for the new variants covered in the launcher reducer + pre-Register table.

### DIFF
--- a/.changesets/1783808987-feat-launcher-observe-only-ui-thread-liveness-probe-teardown-mip8.md
+++ b/.changesets/1783808987-feat-launcher-observe-only-ui-thread-liveness-probe-teardown-mip8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(launcher): observe-only UI-thread liveness probe (teardown backstop Phase 1)
--- a/agentmux-cef/src/launcher_ipc.rs
+++ b/agentmux-cef/src/launcher_ipc.rs
@@ -253,6 +253,20 @@ pub async fn connect_to_launcher(
                             tracing::info!("[launcher-ipc] received event: {:?}", event);
                             apply_event_to_shadow(&state_for_reader, &event);
                         }
+                        Ok(HostFrame::Command(Command::ProbeUiThread { nonce })) => {
+                            // SPEC_LAUNCHER_TEARDOWN_BACKSTOP Phase 1 — NOT a
+                            // saga command (no saga_id, no idempotency LRU:
+                            // duplicate probes are harmless, replies are
+                            // at-most-once per posted task). The reply MUST
+                            // come from the UI thread — replying from here
+                            // (the tokio reader) would prove nothing about
+                            // the thread the backstop actually cares about.
+                            // If the UI thread isn't pumping (wedged, or the
+                            // pre-ready post_task silent drop), the posted
+                            // task never executes and no reply is sent —
+                            // silence IS the signal.
+                            crate::ui_tasks::post_probe_ui_thread_reply(nonce);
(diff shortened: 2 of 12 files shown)


AgentA-asaf:
Both P1s addressed: send-failure now retracts the outstanding probe (nonce-scoped, with a regression test), and UiLiveness logic moved onto the struct so tests run on their own instances — no shared global slot.

AgentA-asaf:
Round-2 P1 addressed: the prober now uses a real try_send_command_no_buffer (the fail-fast variant the HostPipeError docs had reserved) — disconnected sends error immediately and retract, never buffer. Two pipe tests pin the semantics (fail-fast leaves the pending buffer empty; connected delivery reaches the wire). The residual check-to-write race is documented on the method with why it's bounded and attributable.

AgentA-asaf:
Round-3 P1 addressed: the no-buffer send now purges any ProbeUiThread frames from the pending buffer on its error path — covering the write-failure-rebuffers route (send_frame pushes before returning Err). Test mirrors write_failure_clears_writer_and_rebuffers with the opposite assertion (Err + empty buffer). With this, no path can leave a probe frame buffered: pre-check covers already-disconnected, purge covers write-failure, and any hypothetical residual is scrubbed on the next failed tick.

PR #2107, opened 2026-07-11
feat(launcher): srv supervision via host recycle (#942 Phase 2)

AgentA-asaf:
Implements `SPEC_SRV_SUPERVISION_RECYCLE_2026_07_11.md` (in this PR) — the "srv is NOT yet supervised" gap that #942's status map identified as the framework's first genuinely-remaining phase.

## Design — recycle, don't rewire

The original #942 sketch ("SrvManager; host tolerates srv reconnect") predates crash-reproject. The host is disposable now, so srv supervision composes two baked mechanisms: **respawn srv** (same path as cold boot; durable SQLite state intact) → **rebind `srv_result`** (host endpoints flow exclusively through spawn-time env — the rebind IS the rewire) → **deliberately kill the host** and let the existing supervised-restart machinery do the rest: restoring-session splash, new host against the new endpoints, crash-reproject rebuilds the windows.

- Bounded by `SRV_RESTART_BUDGET` (3 per 120s); exhaustion = loud give-up (today's behavior).
- A recycle kill is flagged (`srv_recycle_kill`, consumed once) so the host-restart arm **skips the deterministic-crash ladder** — an srv-driven recycle must never mark the host deterministic-crashing or degrade it to `--disable-gpu` — while still counting against the host budget as runaway protection.

## The bug live verification caught

The first respawned srv died **7ms** after starting (`stdin closed, shutting down`): tokio's `Child::wait()` proactively drops the child's stdin, and srv reads stdin-EOF as parent death. Cold boot parks srv's stdin in a keepalive binding for exactly this reason (the v0.33.447 lesson, documented at the binding); the respawn path now re-parks the new child's stdin identically.

## Verification (all live, packaged build)

- Kill srv mid-session with a second window open → `srv exited UNEXPECTEDLY … respawning srv + recycling host (restart 1/3)` → new endpoints → restoring splash → **main and "Window 2" fully reprojected and functional against the new srv** (RPC round-trip confirmed; `Client.windowids` intact at 2; the recycled frontend's `__WAVE_SERVER_WEB_ENDPOINT__` shows the new port).
- Two further kills exercised repeat recycles (each hit a freshly-respawned srv).
- A third kill inside the window: `srv restart budget exhausted (3 in 120s) — terminating launcher` → clean teardown, **zero** leftover processes.
- Launcher suite 206 passed.

Windows-first per the program convention; `unix.rs` mirrors after bake. srv hang-while-alive detection stays a separate follow-up in the #942 family.

### DIFF
--- a/.changesets/1783813107-feat-launcher-srv-supervision-via-host-recycle-an-srv-crash--kgxb.md
+++ b/.changesets/1783813107-feat-launcher-srv-supervision-via-host-recycle-an-srv-crash--kgxb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(launcher): srv supervision via host recycle — an srv crash no longer kills the session
--- a/agentmux-launcher/src/supervisor/mod.rs
+++ b/agentmux-launcher/src/supervisor/mod.rs
@@ -14,6 +14,17 @@
 pub(crate) const HOST_RESTART_BUDGET: usize = 3;
 pub(crate) const HOST_RESTART_WINDOW: std::time::Duration = std::time::Duration::from_secs(60);
 
+/// SPEC_SRV_SUPERVISION_RECYCLE_2026_07_11 (#942 Phase 2) — srv gets its
+/// own crash budget: an unexpected srv exit respawns srv and recycles the
+/// host through the existing supervised-restart path (crash-reproject
+/// restores the session), at most this many times per window. A wider
+/// window than the host's: each srv recycle also costs a host restart, so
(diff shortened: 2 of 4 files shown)


AgentA-asaf:
P1 addressed: the recycle flag is consumed (mem::replace) before classification and forces the Abnormal path — a deliberate kill can no longer route to the OOM branch (degraded relaunch) regardless of commit pressure, and the flag structurally cannot leak into the next iteration since the Abnormal arm's ladder-skip reads the consumed local.

PR #2109, opened 2026-07-12
feat(sysinfo): track pagefile-volume free disk + warn on a stuck page file

AgentA-asaf:
Implements P0 item 1 of `SPEC_WIN10_PAGEFILE_OOM_CRASH_2026_06_29.md` §5.2 (task #10's cross-reference).

## The gap

The existing commit gauge (`mem:commit:used/total`, StatusBar `PF x/y`) only sees the **symptom** the spec's regression produces: commit near its limit. It's blind to the **cause**: a system-managed page file wants to auto-grow toward `min(3×RAM, ⅛ volume)` but silently can't if the volume it lives on doesn't have the free space, pinning the real commit ceiling well below what the RAM figure alone suggests — invisible to every existing gauge.

## What

- **`agentmux-srv/src/backend/sysinfo.rs`**: new `get_pagefile_volume_data`, emitting `disk:pagefile_volume:{free_gb,total_gb,free_pct}` and `disk:pagefile_system_managed`. Resolves which volume to watch by parsing the `PagingFiles` registry value — `initial=0 && maximum=0` means "system managed" per Microsoft's documented format; an **empty** value (the common default) means Windows fully auto-manages a single pagefile on the system drive, so that case is correctly treated as system-managed rather than "no data" (a naive reading would get this backwards). Registry read is memoized for the process lifetime (machine config, needs a reboot to change); free disk space is re-read every tick since it genuinely moves during a session. 7 unit tests on the pure parsing/resolution logic.
- **`frontend/app/statusbar/SystemStats.tsx`**: new `"Disk NN%"` stat beside the existing `PF` gauge. Colored using the spec's own numbers (**<15% free = warning, <8% = error**) **only** when the page file is system-managed — a fixed-size page file isn't gated by free disk this way, so it's never flagged as at-risk. Tooltip explains the mechanism and shows the free-space figure in GB.

## Verification

Live on a packaged build, against **this machine's actual current disk state**: `df` showed 20G/447G free (~4.5%); the UI rendered **"Disk 4%" in error-red**, tooltip `"...19.4G free"` — matching exactly, and correctly surfacing the live risk condition the spec set out to make visible. Neighboring `PF`/`Mem` gauges unaffected (no regression). `cargo test -p agentmux-srv`: 1487 passed (7 new). `npx vitest run`: 1684 passed.

## Scope note

This is P0 item 1 of the spec's 3-item P0 list. Item 2 (confirm `0xE0000008` is caught by gated recovery) turned out to already be fully covered by existing code on inspection — both the renderer-level path (`crash_recovery.rs`'s dedicated `PROCESS_OOM` → memory-paused page, #1229) and the host-level path (`mem_supervisor::classify_host_exit` explicitly matching `CHROMIUM_OOM_EXIT_CODE = 0xE000_0008`) — no code change needed there. Item 3 (commit-aware turn scheduler re-derived from `PrivateUsage`) and the P1 items (log rotation, shut-down-old-version-on-upgrade) remain open follow-ups.

### DIFF
--- a/.changesets/1783865214-feat-sysinfo-track-pagefile-volume-free-disk-warn-when-a-sys-4cch.md
+++ b/.changesets/1783865214-feat-sysinfo-track-pagefile-volume-free-disk-warn-when-a-sys-4cch.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(sysinfo): track pagefile-volume free disk + warn when a system-managed page file can't grow
--- a/.changesets/1783865214-feat-sysinfo-track-pagefile-volume-free-disk-warn-when-a-sys-4cch.md
+++ b/.changesets/1783865214-feat-sysinfo-track-pagefile-volume-free-disk-warn-when-a-sys-4cch.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
 feat(sysinfo): track pagefile-volume free disk + warn when a system-managed page file can't grow
--- a/agentmux-srv/Cargo.toml
+++ b/agentmux-srv/Cargo.toml
@@ -80,6 +80,8 @@ windows-sys = { version = "0.59", features = [
     "Win32_System_Threading",
     "Win32_System_ProcessStatus",
     "Win32_System_SystemInformation",
+    "Win32_System_Registry",
+    "Win32_Storage_FileSystem",
 ] }
 
 [target.'cfg(windows)'.build-dependencies]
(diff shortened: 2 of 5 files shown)


AgentA-asaf:
Both addressed: GetDiskFreeSpaceExW now runs inside the same block_in_place block as the other synchronous OS reads (it's real filesystem I/O, unlike GlobalMemoryStatusEx's pure in-memory read, so it belongs there); changeset bumped patch → minor to match the new user-visible feature.

PR #2118, opened 2026-07-13
fix(editor): file-tree rename/create/delete silently failed outside $HOME

AgentO-asaf:
## Problem

Renaming a file in the editor's file tree "appears to be a no-op" — you type a new name, press Enter, and the tree just reverts to the old name with no explanation.

## Root cause

Two bugs, compounding:

1. **`renameeditorfile`/`createeditorfile`/`createeditordir`/`deleteeditorfile`** (`agentmux-srv/src/server/editor_handlers.rs`) hard-restrict every mutation to strictly inside `$HOME`. But `readeditorfile`/`writeeditorfile` have **no such restriction** — the editor already lets you open and save any file the OS permits. `SPEC_FILE_TREE_CONTEXT_MENU_2026_06_14.md`'s own "Path Safety" section calls for rename/create/delete to follow "the same policy as `readeditorfile` / `writeeditorfile`" — this was never actually implemented that way.
2. **The frontend swallows every failure silently** — `renameFile`/`deleteFile`/`createFile`/`createDir` in `editor-model.ts` only `console.error`'d on RPC rejection. No toast, no inline error, nothing user-visible.

Together: any rename outside `$HOME` (a mounted drive, a project checked out elsewhere, or simply a **symlink inside `$HOME` that canonicalizes to a target outside it** — common on real dev machines) silently fails and looks exactly like a no-op.

## Verification (live, via CDP on a running dev instance)

- Renamed a file under `/tmp` (outside `$HOME`) → backend returned `"path outside or is the home directory"`. Reproduced the exact symptom.
- Applied the fix → the same rename call succeeded.
- Confirmed the new filesystem-root guard still blocks `rename("/")` → `"refusing to rename the filesystem root"`.
- Forced a destination-collision rename → a visible toast now appears: **"Rename failed — renameeditorfile: destination already exists"**.
- Re-ran the original in-`$HOME` rename flow end-to-end (right-click → Rename… → type → Enter) → still works, tree updates, file renamed on disk.

## Fix

1. `editor_handlers.rs`: all four handlers now match `readeditorfile`/`writeeditorfile`'s actual policy — `expand_home_dir_safe`, no home-only gate. Added one new guard neither read/write nor the old code needed: refuse to rename/delete the filesystem root itself (`is_filesystem_root()`).
2. `editor-model.ts`: new shared `reportFileOpError()` helper pushes a visible error notification (`pushNotification`, same pattern `term.tsx` already uses) on any file-tree mutation failure, not just `console.error`. This makes any future silent-failure mode (permissions, disk full, etc.) visible too, not just this specific bug.

## Testing

- `tsc --noEmit` clean.
- `cargo check -p agentmux-srv` clean, no new warnings in the touched file.
- Live CDP verification (4 scenarios above) on `task dev`.


AgentO-asaf:
All three findings are correct, and thank you for catching them — I made a real research error in the original diagnosis, not just a missed edge case.

**What was actually wrong**: I checked `readeditorfile` (genuinely no home restriction) and incorrectly assumed `writeeditorfile` matched it without re-verifying `writeeditorfile`'s own code closely enough. It doesn't — `writeeditorfile` (lines 268-288) enforces the exact same `canonical.starts_with(&canonical_home)` check I removed from rename/create/delete, with an explicit comment: *"Path safety: restrict writes to under the user's home directory. Allowlist approach — safer than an incomplete denylist."* That's a deliberate, existing security policy — read is unrestricted (lower risk), every mutating operation (write, and originally rename/create/delete too) is home-scoped. My PR's premise ("mutations should match read/write's permissive policy") was based on a policy that doesn't actually exist; the real asymmetry is read-permissive vs. write-restrictive, and rename/create/delete were already correctly aligned with write before my change.

**Fix**: reverted `agentmux-srv/src/server/editor_handlers.rs` to byte-identical with `main` (confirmed via `git diff main -- <file>`, empty) — full restoration of the original `$HOME` boundary checks on all four handlers, including the `canonical_old == canonical_home` / `canonical == canonical_home` guards you flagged as P0. Dropped the `is_filesystem_root()` helper — no longer needed, since the restored home-boundary check already transitively blocks the actual filesystem root in every realistic case (home is never `/` unless running as root, and `/` fails `starts_with(&canonical_home)` regardless).

**What's still fixed and still correct**: the frontend `reportFileOpError()` → `pushNotification()` change (`editor-model.ts`) is untouched and remains the actual root-cause fix for "rename appears to be a no-op." That symptom is now fully explained without touching the path policy at all: a rename outside `$HOME` is *supposed* to be rejected (by design, matching write), but the rejection was completely silent — no toast, no inline error, `console.error` only. The user saw the inline input close and nothing else happen. Now every rejection (home-boundary, destination-collision, whatever) surfaces a visible "X failed: <reason>" toast — which resolves the reported confusion without loosening any security boundary.

Verified live again post-correction: rename within `$HOME` still works end-to-end; rename outside `$HOME` now correctly fails *with a visible toast* instead of silently; renaming `$HOME` itself is blocked again.

PR #2127, opened 2026-07-13
fix(agent-pane): stop O(n) rebuilds on scroll and streamed chunks

AgentA-asaf:
## Summary

Task #39. A prior investigation found that the agent pane's DOM windowing (visible rows + a 50-node streaming buffer) works correctly, but the **data layer underneath** did O(total-conversation-length) work on every streamed chunk AND on every scroll event — making the pane get measurably slower as a session's conversation grows, independent of any LLM/API context-size cost.

This PR is the fix, in priority order from the task brief:

1. **Scroll no longer triggers a full O(n) layout recompute.** `agent-pane-layout-store.ts`'s `dispatch` used to gate the full `computeLayoutView()` (which rebuilds the O(n) prefix-sum `positions()` array) on ANY of `orderedIds/heights/estimates/scrollTop/viewportPx/scrollMarginPx/overscan` changing. Split into `positionInputsChanged` (order/heights/estimates/scrollMargin — real position-affecting changes, needs full O(n) rebuild) vs `windowInputsChanged` (scrollTop/viewportPx/overscan — reuses the cached positions array, just re-derives the visible window via the already-existing O(log n) `windowRangeOf` binary search). A pure scroll now costs O(log n), not O(n).

2. **Scroll handler is now RAF-coalesced.** `AgentDocumentVirtualList.tsx`'s `handleScroll` used to dispatch on every native `scroll` event (which can fire dozens of times per drag/wheel gesture). Now coalesced to at most one dispatch per animation frame, mirroring the `scheduleFlush`/`flushRafId` idiom `useAgentStream.ts` already uses for stream-chunk batching.

3. **Streamed chunks no longer rebuild an id→index Map or linear-scan `nodes` on every chunk.** Added an incrementally-maintained `nodeIndexById: Map<string, number>` to `AgentDocumentState`, kept in lockstep with `nodes` the same way `nodeIdSet` already is (issue #728 gap 4 pattern). `StreamFlush`, `ToolChunkAppend`, `ShellChunkAppend`, `ShellStatusUpdate` all resolve their target node in O(1)/O(k) now, instead of an O(n) Map rebuild (`StreamFlush`) or O(n) linear scan (the others) on every single chunk.

## Residual / known limitation

`state.nodes.slice()` itself is still O(n) per chunk. `nodes` is exposed to the reactive layer as a plain-array **signal** (`agent-document-store.ts`'s `DocumentSetter`), not a SolidJS `createStore` — so a brand-new array reference is structurally required for the signal to fire. Eliminating that too would mean switching the document state to a store-based `produce`/`reconcile` model for fine-grained in-place updates, which is a bigger architectural change than this task's scope (the task explicitly said not to touch the virtualization architecture). Flagging this as a legitimate follow-up rather than forcing it into this PR.

## Test plan

- [x] Verified all cited file:line locations against current code (mostly matched; `StreamFlush` had already grown a lazy-clone guard since the investigation, but still rebuilt the O(n) `indexById` Map every flush)
- [x] Added tests pinning the new behaviors:
  - `agent-pane-layout-store.test.ts`: scroll-only dispatch reuses the cached `rows` array reference (proves no O(n) recompute); a subsequent data-changing dispatch still rebuilds `rows`; the reused-positions window is still correct (not stale) across repeated scrolls
  - `reducer.test.ts`: `nodeIndexById` stays correct across every command that touches `nodes` (StreamFlush append/collision, HistoryLoaded/HistoryRestored prepend-shift, UserClear/StreamTruncate reset, ShellNodeCreate/ShellChunkAppend/ShellStatusUpdate, and the non-tool-node drop path)
- [x] Full frontend suite: `npx vitest run` → 1701 passed, 3 skipped (unrelated e2e tests needing a build dir), 0 failures

### DIFF
--- a/.changesets/1783908843-fix-agent-pane-stop-o-n-rebuilds-on-scroll-and-streamed-chun-suba.md
+++ b/.changesets/1783908843-fix-agent-pane-stop-o-n-rebuilds-on-scroll-and-streamed-chun-suba.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): stop O(n) rebuilds on scroll and streamed chunks in agent-pane conversation rendering
--- a/.changesets/1783910275-chore-agent-pane-add-scroll-stream-dispatch-bench-suite-and--0d8b.md
+++ b/.changesets/1783910275-chore-agent-pane-add-scroll-stream-dispatch-bench-suite-and--0d8b.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+chore(agent-pane): add scroll/stream dispatch bench suite and perf-probe dispatch timing (task #40)
--- a/frontend/app/devtools/agent-pane-perf-section.tsx
+++ b/frontend/app/devtools/agent-pane-perf-section.tsx
@@ -4,8 +4,10 @@
 /**
  * Agent-pane perf section for the diag panel — Phase 3 of the
  * virtualization redesign. Polls `agentPerfStore.snapshot()` at 1 Hz
- * and renders per-kind row mount times, estimator-miss rates, and
- * recent layout shifts.
+ * and renders per-kind row mount times, estimator-miss rates, recent
+ * layout shifts, and (task #40) layout/document store DISPATCH times —
+ * the reducer/store cost underneath the DOM that the original probe
+ * never covered.
  *
  * No-op in production: snapshot returns empty when probing is
  * disabled, so the section renders nothing.
@@ -28,6 +30,7 @@ const EMPTY_SNAPSHOT: AgentPerfSnapshot = {
     estimatorMissRateByKind: new Map(),
     recentEstimatorMisses: [],
     recentLayoutShifts: [],
+    dispatchByKind: new Map(),
 };
 
 function formatMs(n: number | undefined): string {
@@ -67,7 +70,8 @@ export function AgentPanePerfSection(): JSX.Element {
         const s = snapshot();
         return s.rowMountByKind.size > 0
             || s.recentEstimatorMisses.length > 0
-            || s.recentLayoutShifts.length > 0;
+            || s.recentLayoutShifts.length > 0
+            || s.dispatchByKind.size > 0;
     };
 
     return (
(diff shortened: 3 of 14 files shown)


AgentA-asaf:
## Follow-up: benchmark diagnostics (task #40)

Added a Vitest bench suite (new convention for this repo — first `.bench.ts` files, run manually via `npx vitest bench <path>`, not wired into CI) plus dispatch-timing in the dev-only perf probe. Actual numbers observed on this machine:

**`agent-pane-layout-store.bench.ts`** — scroll-only `dispatch()` at 100/1,000/10,000 already-positioned rows:

| rows | mean | vs 100 rows |
|---|---|---|
| 100 | 0.0062ms | 1x |
| 1,000 | 0.0105ms | 1.69x |
| 10,000 | 0.0279ms | 4.51x |

100x more rows → only ~4.5x slower (sub-linear, consistent with the O(log n) `windowRangeOf` re-derivation over cached positions). The full O(n) prefix-sum rebuild this replaced would show ~50-100x growth over the same range.

**`agent-document/reducer.bench.ts`** — `StreamFlush`, two scenarios:

| scenario | 100→1,000 nodes | 100→10,000 nodes |
|---|---|---|
| append (new node) | 7.74x slower | 197x slower |
| update-only (dominant per-token streaming case) | 2.25x slower | 14x slower |

Honest finding, documented in the bench file's comments: **neither scenario is flat/O(1)**. `agent-document-store.ts` exposes `nodes` as a plain-array Solid signal (not a `createStore`), so `state.nodes.slice()` stays an O(n) full-array copy on every nodes-changing flush — that's the residual already called out in this PR's description. What the `nodeIndexById` fix actually removed was a *second* O(n) pass stacked on top of that copy (the unconditional id→index Map rebuild-by-scan that used to run on every flush regardless of whether anything changed). That shows up exactly as expected: "update-only" (map cost now zero) scales far better than "append" (map clone is still O(n)) over the same size range.

Also extended `perf-probe.ts` with `markDispatch("layout" | "document")` / `recordDispatchTiming`, mirroring the existing `markRowMount` pattern, wired into both stores' `dispatch()` call sites (not the reducers — both are documented pure/no-I/O) and surfaced in the existing dev diagnostics panel (`agent-pane-perf-section.tsx`) as a new "Store dispatch duration" table. This closes the blind spot the original investigation flagged: the probe measured DOM row-mount cost but never the reducer/store dispatch cost underneath.

Full suite: 1705 passed, 3 skipped (unrelated e2e), 0 failures. `tsc --noEmit` clean.

PR #2151, opened 2026-07-14
fix(linux): gate startup window-show/splash-dismiss on real first paint

### LINKED ISSUE #2146: Linux: startup blank/white-flash before window content settles
## Summary

On Linux only, after the native launcher splash finishes and the main window appears, users see a blank/transparent window, then a white flash, then a loading indicator, then a brief blank moment again, before the UI settles. Windows is smooth by comparison — same host source runs on both platforms.

## Diagnosis (done)

Full write-up, real profiling data (a `[startup-bench]` timeline from an actual run), and a phased fix proposal are in:

**`docs/specs/SPEC_LINUX_STARTUP_PAINT_GATING_2026_07_13.md`** (branch [`docs/linux-startup-white-flash-spec`](https://github.com/agentmuxai/agentmux/tree/docs/linux-startup-white-flash-spec/docs/specs/SPEC_LINUX_STARTUP_PAINT_GATING_2026_07_13.md), not yet merged)

Short version: `agentmux-cef/src/client/navigation.rs`'s `on_load_end` shows the native window and dismisses the launcher splash on **main-frame load-complete**, not on **first actual paint** — the same code path runs on Windows and Linux, but Windows got a narrower fix for this exact class of bug back in March/April (`docs/analysis/cef-white-flash-on-startup.md`, `cef-white-flash-retro.md`) that was never re-verified for Linux's slower GPU/EGL init path. Ruled out the Track-2 per-pixel-transparency Blink bug (`SPEC_TRANSPARENCY_MACOS_LINUX_2026_07_01.md`) as the cause — that's a different, already-tracked issue.

## Proposed plan (not yet implemented)

1. **Telemetry**: add a "frontend painted" IPC signal (first `requestAnimationFrame` after the splash SVG is in the DOM) so we can measure the exact blank-to-painted gap instead of inferring it.
2. **Fix**: gate `window.show()` / focus / splash-ready-file on that signal instead of `on_load_end`, with a safety-net timeout. Linux-only, no CEF rebuild needed.
3. **A/B test** `--disable-gpu-compositing` on Linux — this was "essential" in the Windows testbed but isn't applied by default anywhere today.
4. **Separate follow-up**: real frontend perf debt independent of the native flash — ~15 local IPC calls in the startup invoke-batch running serially at 130–380ms each, plus a 1.16s `document.fonts.ready` wait. Contributes to the "some loading" part of the symptom.

See the spec for full profiling numbers, root-cause evidence (file:line citations), and open questions.

AgentU-asaf:
## Summary

Fixes #2146 — on Linux, the native window and launcher splash were shown/dismissed from `on_load_end` (CEF's main-frame *load-complete* event), not from an actual compositor paint. Windows/macOS got away with this; Linux's slower GPU/EGL init path exposed it as a blank/transparent window, a white flash, then the loading brain, then a brief blank moment again before settling.

- `frontend/bootstrap.ts`: fires a new `report_first_paint` IPC call on the second `requestAnimationFrame` after script start (the standard "a frame was actually presented" proxy), sent directly via `invokeCommand` rather than waiting on the slower `setupCefApi()` bridge.
- `agentmux-cef/src/client/navigation.rs`: on Linux only, `on_load_end` no longer shows/focuses the window or writes the splash-ready-file immediately. It arms a pending-gate entry + a safety-net timeout; the real reveal happens in `reveal_gated_window`, triggered by whichever of the paint signal or timeout fires first. Windows/macOS behavior is unchanged.
- `agentmux-cef/src/commands/backend.rs` / `ipc.rs` / `state.rs`: new `report_first_paint` command + `linux_paint_gate_pending` state (keyed by arm-time, so reveals log which path won and how long the window was gated).

Full diagnosis + profiling data: `docs/specs/SPEC_LINUX_STARTUP_PAINT_GATING_2026_07_13.md` (on the `docs/linux-startup-white-flash-spec` branch, not yet merged).

## Verification

Built + ran a fresh `task package:linux` AppImage twice:

1. First pass used a 1500ms safety timeout. Cross-referencing `on_load_end` and `report_first_paint` timestamps showed the real signal for the main window landing ~2.08s out (pool-window prewarms, unaffected by this gate, landed 1.36–1.84s out — confirming this is real compositor-first-frame latency in this environment, not an artifact of the change). The 1500ms cap fired *before* the real signal, silently degrading the fix to "always show after 1.5s" instead of waiting for real paint.
2. Widened the timeout to 4000ms and added reveal-path logging. Rebuilt, killed the stale same-version instance (single-instance pipe would've otherwise forwarded to it), relaunched fresh:

   ```
   [startup-paint] revealing gated window label=main reason=signal elapsed_ms=1487
   ```

   Confirms the window is now revealed by the real paint signal, not the timeout fallback. No errors in the host log; `startup-bench` shows `frontend-painted` firing before `isMainWindow-done`, i.e. the window reveals as soon as there's real content to show rather than waiting for full app boot.

`cargo check -p agentmux-cef` and `tsc --noEmit` both clean.

## Test plan

- [x] `cargo check -p agentmux-cef`
- [x] `tsc --noEmit`
- [x] Built `task package:linux`, verified via host log that `main` reveals via `reason=signal` not `reason=timeout`
- [ ] Manual visual confirmation on a real Linux desktop (this sandbox's screenshot tooling wasn't reliable for a before/after visual — log-based verification only)
- [ ] Windows/macOS smoke test to confirm no behavior change (code is `#[cfg(target_os = "linux")]`-gated, but worth a sanity check)

### DIFF
--- a/.changesets/1783994551-fix-linux-gate-startup-window-show-splash-dismiss-on-real-fi-oy1y.md
+++ b/.changesets/1783994551-fix-linux-gate-startup-window-show-splash-dismiss-on-real-fi-oy1y.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(linux): gate startup window-show/splash-dismiss on real first paint, not load-complete
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -53,7 +53,7 @@ mod handlers;
 pub(crate) mod helpers;
 mod lifecycle;
 mod display;
-mod navigation;
+pub(crate) mod navigation;
 mod crash_recovery;
 mod recovery_pages;
 #[cfg(target_os = "windows")]
--- a/agentmux-cef/src/client/navigation.rs
+++ b/agentmux-cef/src/client/navigation.rs
@@ -34,16 +34,50 @@ use crate::state::AppState;
 #[cfg(target_os = "linux")]
 const PAINT_GATE_SAFETY_TIMEOUT_MS: i64 = 4000;
 
+/// Monotonic arm counter for the Linux paint gate. Each `on_load_end` call
+/// that (re-)arms a label's gate gets a fresh epoch; its safety-net timeout
+/// task captures that epoch and only acts if it's still current when the
+/// timeout fires (see `reveal_gated_window`). Mirrors the identical
+/// stale-timeout guard in `browser_pane::auth` (`NEXT_EPOCH`/`Entry::epoch`).
+#[cfg(target_os = "linux")]
+static PAINT_GATE_NEXT_EPOCH: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+
 /// Reveal (show + focus the real window, dismiss the native splash) once —
 /// whichever of the real `report_first_paint` IPC signal or the safety-net
 /// timeout fires first wins; the other becomes a no-op via the
-/// `linux_paint_gate_pending` membership check. Must run on the CEF UI thread.
+/// `linux_paint_gate_pending` membership check.
+///
+/// `expected_epoch`: `None` for the real-signal path (always reveal whatever
+/// is currently armed for `label` — a stale signal from a torn-down document
+/// can't fire; CEF cancels pending rAF/timers on navigation). `Some(epoch)`
+/// for the safety-timeout path — reveals only if `label`'s current arm is
+/// still the one this specific timeout was scheduled for; if `on_load_end`
+/// re-armed the gate since (reload/retry mid-startup), this stale timeout is
+/// a no-op instead of revealing the window ahead of the *current*
+/// navigation's real paint (docs/specs/SPEC_LINUX_STARTUP_PAINT_GATING_2026_07_13.md,
+/// reagent PR #2151 review).
+///
+/// Must run on the CEF UI thread.
 #[cfg(target_os = "linux")]
-pub(crate) fn reveal_gated_window(state: &Arc<AppState>, label: &str, reason: &'static str) {
+pub(crate) fn reveal_gated_window(
+    state: &Arc<AppState>,
+    label: &str,
+    reason: &'static str,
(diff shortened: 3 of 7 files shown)


AgentU-asaf:
Fixed the stale-timeout race flagged above (09a2ed10).

Root cause was real: `on_load_end`'s gate-arm block had no guard against re-arming the same label before the first reveal — a stale delayed task from an earlier arm would fire at its original deadline and could reveal the window ahead of a later navigation's real paint.

Fix mirrors the identical pattern already in `browser_pane::auth.rs` (`NEXT_EPOCH`/`Entry::epoch`): each arm gets a monotonic epoch, and the safety-timeout task only reveals if its epoch is still current for that label when it fires. The real `report_first_paint` signal path doesn't need the check — a stale signal from a torn-down document can't fire (CEF cancels pending rAF/timers on navigation), same as `auth.rs`'s `take()` skipping the epoch check on the resolve side.

Re-verified on a fresh `task package:linux` build — normal (non-race) path still reveals via `reason=signal` with no regression:
```
[startup-paint] revealing gated window label=main reason=signal elapsed_ms=1449
```

AgentU-asaf:
Fixed the ordering race flagged above (0d2255e2).

Confirmed: \`on_load_end\`'s main-frame load-complete isn't guaranteed to fire after the frontend's first compositor frame — render-blocking stylesheets can resolve, and a frame can paint, before other load-blocking resources finish. So \`report_first_paint\` could reach the UI thread before \`on_load_end\` armed the gate, and the old \`None => return\` silently dropped that real signal, falling through to the 4s safety timeout.

Added \`linux_first_paint_seen\`: when the signal arrives before the gate is armed, the label is recorded there instead of dropped. \`on_load_end\` checks it before arming and reveals immediately if present. Both orderings (arm-first, signal-first) now resolve correctly.

Re-verified on a fresh build — normal path unaffected:
\`\`\`
[startup-paint] revealing gated window label=main reason=signal elapsed_ms=1428
\`\`\`

PR #2162, opened 2026-07-15
Settings pane: fill out Window & Panes, Sounds, Network, Advanced sections

AgentY-asaf:
## Summary
- Wires the ~46 remaining `schema/settings.json` keys that were `StubSection` placeholders, completing Phases 3-4 of the Settings pane work.
- Restructures the rail from 7 to 6 sections: drops **Agent** and **Files** (the schema no longer backs a dedicated Agent section, and Files was down to a single key), adds **Window & Panes** to hold window-behavior keys that don't belong in the purely-visual Appearance section.
- Adds two new reusable controls: a key-value editor (`cmd:env`) and a string-array editor (`term:localshellopts`).
- Extends Appearance (bg color, pane hover-magnify group) and Terminal (predictive echo) with keys added to the schema since the original spec.
- Corrects a stale `CLAUDE.md` line describing Settings as opening `settings.json` directly in an editor.

Full key-by-key catalog reconciliation and rationale: `specs/SPEC_SETTINGS_PANE_COMPLETION_2026_07_14.md`.

Deliberately excluded from any UI (see spec §6): `tab:preset` (dead, zero consumers), `window:dimensions` (app-managed geometry string, not user-facing), `app:dismissarchitecturewarning` (one-shot dismiss flag).

Not in this PR (spec Phase 5, still open): `settings:section` persistence across pane remounts, theme preview swatch, search/filter across all keys.

## Test plan
- [x] `tsc --noEmit` across the frontend — passes with zero errors
- [x] `stylelint` on `settings.scss` — no new violations (7 pre-existing hex-color grandfather violations unchanged)
- [ ] Manual click-through of all 6 sections in a running build (not done in this session — recommend before merge)


AgentY-asaf:
Addressed the review findings in cfdd1f84:

- Verified against Rust source: 11 keys had zero reads anywhere beyond their serde struct declaration (`window:showmenubar`/`nativetitlebar`/`disablehardwareacceleration`/`maxtabcachesize`/`confirmclose`/`savelastwindow`/`zoom`, `app:globalhotkey`, `widget:showhelp`, `telemetry:enabled`, `preview:showhiddenfiles`) — removed from the UI.
- `conn:wshenabled`/`conn:askbeforewshinstall` have no matching `SettingsType` struct field at all — the live versions are per-connection in `schema/connections.json`. Removed.
- `network:lan_discovery` is real but already has a working toggle in `HostPopover.tsx` — dropped the duplicate rather than shipping two disconnected controls for the same key.
- This collapsed the Network section to zero keys, so it's removed entirely (5 sections now, not 6). Window & Panes trimmed from 11 keys to 4.
- Fixed `KeyValueEditor` to key `<For>` by stable `Object.keys()` strings instead of fresh `Object.entries()` tuples, so editing one row no longer remounts all of them.

Also credit where due: `specs/settings-cleanup.md` had already done this audit properly back in May and I missed cross-referencing it — updated `specs/SPEC_SETTINGS_PANE_COMPLETION_2026_07_14.md` with a correction note rather than silently rewriting history.

AgentY-asaf:
Addressed round 3 (b53a0170):

- \`term:localshellpath\`/\`term:localshellopts\`: verified against \`blockcontroller/shell/lifecycle.rs:252\` — the Windows shell-spawn path calls \`detect_local_shell_path_windows()\` unconditionally and never reads either setting; both constants are \`#[allow(dead_code)]\`. Removed. This was the same mistake as round 2, one level deeper: a declared Rust const is evidence of deserialization, not evidence of a read site — I should have grepped for actual call sites the first time, not just matching const names across two files.
- Since that was \`StringArrayEditor\`'s only caller, deleted the component entirely rather than leaving it unused — which also resolves the P2 finding on it (duplicate array entries colliding under \`<For>\`'s primitive-value keying) by removing the code instead of patching a bug in something nothing calls anymore.
- Spec doc updated again with a second correction note, plus a standing lesson for future settings work: verify usage sites, not just declaration sites, before wiring UI to a key.

AgentY-asaf:
- Fixed the stale changeset text/filename (8f5016df) — it still said "Network" after that section was cut in an earlier fix commit.
- The windows-latest CI failure (\`bash_wrap::tests::run_via_pty_does_not_misclassify_fast_success_as_idle_timeout\`) is unrelated to this PR — this PR touches only \`frontend/app/view/settings/\`, \`CLAUDE.md\`, \`specs/\`, and \`.changesets/\`; the failing test lives in \`agentmux-bashwrap\`, a crate nothing here touches, and the assertion message itself describes an "idle_rx/wait_task race" — a timing-sensitive test. Requested a rerun of the failed job to confirm.

PR #2167, opened 2026-07-15
docs: retro + universal identity gateway spec (PR #2164's gate doesn't gate what's actually running)

AgentA-asaf:
## What happened

Live user report, minutes after the previous auth-isolation retro (#2165) merged: Armory shows **zero** accounts, yet the "Lzop" agent kept responding.

Verified directly against both running instances' live SQLite databases — not a synthetic repro:
- `db_accounts` is genuinely empty in the shared store.
- **2 of 2** agent-shaped blocks actually observed running (Lzop in the dev:main channel; this very conversation's own pane in the v0.53.3 channel) have zero matching `db_agent_instances` row — and for Lzop, zero matching `db_agents` row either.
- Lzop's block carries a **frozen** `CLAUDE_CONFIG_DIR` pointing at the old shared isolated dir, whose credentials are still being actively refreshed by something else — it keeps working not because it passed a check, but because nothing ever asked it to prove anything.

## Root cause

`identity/resolver.rs`'s `inject_identity_env_with_broker` has **two unconditional early-returns that were never touched by tonight's PR #2164**:
- Step 1: no `db_agents`/`db_agent_instances` row for the block → `Ok(())`, no gate.
- Step 2: row exists but `identity_id` is blank → `Ok(())`, no gate.

Every PR #2164 test seeds its fixture via `instance_create`/`agent_def_insert` directly — exactly the shape real agents don't reliably have. The App API `agent.open` path never creates an instance row; the frontend launch modal's own attempt is best-effort, wrapped in try/catch, called *after* the spawn already happened.

## What this PR does (docs only — implementation is task #50)

- `docs/retro/retro-identity-gate-bypassed-by-missing-instance-binding-2026-07-15.md` — the incident, verified live, with exact citations.
- `docs/specs/SPEC_UNIVERSAL_IDENTITY_GATEWAY_2026_07_15.md` — the fix design, **plus the full inventory of every place a provider CLI could be spawned** (the user's "identify all the places login may happen" ask). Conclusion: this is NOT a scattered-spawn-paths problem — exactly two call sites ever invoke injection (`agentinput`, `agent.send`), both correctly wired. The bug is entirely inside the one shared function's own decision tree, which currently trusts two disconnected identity signals instead of one.
- Cross-link added to `identity/resolver.rs`'s module doc, matching the pattern from the prior same-day retro (#2165).

## Verification

- `cargo check -p agentmux-srv` clean (comment-only code change)
- All root-cause claims verified against live production data via direct SQLite queries, not assumed from code reading alone

### DIFF
--- a/agentmux-srv/src/identity/resolver.rs
+++ b/agentmux-srv/src/identity/resolver.rs
@@ -20,6 +20,18 @@
 //! and it was never restored — today's gate only chooses between "block"
 //! and "true ambient" (`use_ambient_login=true`, zero isolation), not the
 //! isolated-auto-provision option that used to exist implicitly.
+//!
+//! **Second, read this too:** `inject_identity_env_with_broker`'s Steps 1/2
+//! (`instance_get_active_for_block` returning `None`; `identity_id` empty
+//! or `"blank"`) are UNCONDITIONAL early-returns that skip the Step 3/4
+//! gate entirely — and live verification on 2026-07-15 found this is not a
+//! rare edge case: every real running agent inspected hit one of these two
+//! paths. See `docs/retro/retro-identity-gate-bypassed-by-missing-instance-binding-2026-07-15.md`
+//! and `docs/specs/SPEC_UNIVERSAL_IDENTITY_GATEWAY_2026_07_15.md` before
+//! assuming the Step 3/4 gate actually runs for the agent you're testing
+//! with — check whether its block has a `db_agents`/`db_agent_instances`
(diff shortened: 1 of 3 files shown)


AgentA-asaf:
Per updated workflow guidance: docs-only PRs should stay local until bundled with a feature PR that touches related code. Content preserved on the branch (docs/universal-identity-gateway-retro-and-spec) — will fold into task #50's implementation PR.

PR #2168, opened 2026-07-15
docs: live-measure the New Window color-flash gap PR #2163 couldn't verify

AgentA-asaf:
## What

PR #2163's own investigation report flagged its §4.3 fix (the pool-promote path's native-show-vs-CEF-compositor-show ordering) as the fix it had the **least** confidence was fully resolved — explicitly because verifying it needed a real Windows GUI session the authoring environment didn't have.

This session has one. Triggered a real `openNewWindow()` against the running dev instance (built on a commit including #2163) and read the exact `pool:new-window` timing telemetry that PR added for this measurement:

```
18:54:53.511Z  — openNewWindow() called
18:54:53.779Z  — CEF Views compositor first told to show   (268ms gap)
18:54:53.7796Z — CEF Views show completed (elapsed_ms: 2)
```

**268ms** between the trigger and CEF's compositor being told to show anything — well inside human-perceptible flash range, and confirms the report's own ranking (§4.3 as "the dominant cause") holds up under live data.

## What this does NOT do

No reorder attempted. The naive fix (show CEF Views before the native Win32 show) already caused a *worse* regression once — a fully blank window, per the cited prior research doc — because `set_taskbar_hidden`'s own hide→show cycle unconditionally re-hides whatever was shown before it. That risk calculus doesn't change just because the gap is now measured instead of suspected.

## What's here

- `docs/specs/REPORT_NEW_WINDOW_COLOR_FLASH_LIVE_MEASUREMENT_2026_07_15.md` — the measurement, method, and an acceptance test for whoever implements the real fix (same CDP-trigger + telemetry method, gap should shrink toward ~2ms).
- Cross-link added to the original report.
- Implementation tracked as task #51, not attempted here.

### DIFF
--- a/agentmux-srv/src/identity/resolver.rs
+++ b/agentmux-srv/src/identity/resolver.rs
@@ -20,6 +20,18 @@
 //! and it was never restored — today's gate only chooses between "block"
 //! and "true ambient" (`use_ambient_login=true`, zero isolation), not the
 //! isolated-auto-provision option that used to exist implicitly.
+//!
+//! **Second, read this too:** `inject_identity_env_with_broker`'s Steps 1/2
+//! (`instance_get_active_for_block` returning `None`; `identity_id` empty
+//! or `"blank"`) are UNCONDITIONAL early-returns that skip the Step 3/4
+//! gate entirely — and live verification on 2026-07-15 found this is not a
+//! rare edge case: every real running agent inspected hit one of these two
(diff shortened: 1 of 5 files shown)


AgentA-asaf:
Per updated workflow guidance: docs-only PRs should stay local until bundled with a feature PR that touches related code. Content preserved on the branch (docs/new-window-flash-live-measurement) — will fold into task #51's implementation PR.

Issue #2179, opened 2026-07-16
Pane minimize/collapse — locked-state redesign (consolidated tracking)


AgentO-asaf:
Consolidated tracking thread for the pane minimize/collapse workstream. Gathers all prior scattered work (previously tracked only through PRs) and the newly-reported hole into one place.

## Current problem (open)

**A minimized pane can still be resized.** Minimize is supposed to be a locked state; today it is advisory data inside the flex tree that every mutation path must individually respect — and almost none do. Full audit + redesign proposal: `docs/specs/SPEC_LAYOUT_MINIMIZE_LOCKED_STATE_REDESIGN_2026_07_16.md`.

Audit highlights (see spec §4 for the full matrix):
- Resize handles are generated on minimized panes' edges (`layoutGeometry.ts`), and the drag path (`layoutResize.ts:101-108`) explicitly **exempts** minimized nodes from the min-size guard — backwards awareness that makes the lock resizable.
- The backend reducer's `resize_nodes` blindly writes any size to any node — agent/API-driven resizes bypass the frontend entirely.
- Move/insert/swap/split/DnD/magnify have zero minimize awareness on either side.

## Plan (spec §5)

- **Option B (now):** write-point enforcement, same pattern as `prune_dangling_block_refs` — reject locked-target ops in both reducers (`LayoutError::NodeLocked`), snap-back normalization pass (`enforce_minimized_locks`) at the balance/prune choke points, suppress resize handles on locked edges (which deletes the backwards exemption).
- **Option C (follow-up decision):** structural redesign — minimized panes leave the flex tree into a per-column "minimized dock", deleting the entire slip/dissolve/`_slipAnchor` special-case machinery. Correct by construction; needs its own spec + migration.

## Merged history consolidated here

- #2036 — dead-space point fix: delete-block saga now notifies frontends via `pendingbackendactions`.
- #2039 — `prune_dangling_block_refs`: reducer-level self-healing invariant, "no layout leaf may reference a nonexistent block" enforced at every write ("WRR-style", event-driven at the point of write).
- #2176 — dissolved-column direction flip: `balanceNode`/`balance_node` no longer flips a `columnDissolve` column's own flexDirection ("minimized pane becomes narrow instead of short").
- `docs/investigations/INVESTIGATION_LAYOUT_DEAD_SPACE_STALE_TREE_RESURRECTION_2026_07_08.md` — root-cause report for the dead-space class.

[...]

PR #2180, opened 2026-07-16
fix(layout): minimize is a locked state (Option B of the locked-state redesign)

AgentO-asaf:
## Summary

A minimized pane was freely resizable — dragging the adjacent handle overwrote its header-height size while the minimize state still said "minimized," corrupting the restore arithmetic and persisting the broken state through `db_layout`. The one minimize-aware line in the resize path was aware **backwards**: `layoutResize.ts` exempted minimized nodes from the `MinNodeSizePx` guard so adjacent handles would keep working, which is exactly what made the lock draggable. An audit (spec §4) found every other mutation surface — move/swap/split/DnD on the frontend, `resize_nodes`/`move_node`/`swap_nodes`/`split_impl` on the backend — entirely minimize-blind.

Implements **Option B** of `docs/specs/SPEC_LAYOUT_MINIMIZE_LOCKED_STATE_REDESIGN_2026_07_16.md` (included in this PR), tracked in #2179. Same enforcement shape as `prune_dangling_block_refs` (#2039): the invariant lives at the write choke points, not in each caller.

## Changes

- **Lock definition:** `isNodeLocked` (TS) / `is_node_locked` (Rust) — locked = `minimizedSize` (minimized leaf) ∪ `slipMinimize` (slipped header) ∪ `columnDissolve` (dissolved column).
- **Lock recording:** new `minimizedLockedSize` field written at all three squeeze sites (normal minimize, slip, dissolve), cleared at all three restore sites. Additive — old persisted trees simply lack it until the next minimize.
- **Reducer guards (both sides):** frontend `resizeNode`/`moveNode`/`swapNode`/`splitHorizontal`/`splitVertical` reject locked targets; resize rejects the whole before/after pair (applying half would leak flex units). Rust `resize_nodes`/`move_node`/`swap_nodes`/`split_impl` return the new `LayoutError::NodeLocked`; `move_node` also rejects a locked *destination* (insertion into a dissolved column). This closes the agent/API path that never touches the frontend.
- **Snap-back enforcement:** `enforceMinimizedLocks` runs in `updateTree` before `balanceNode`; `enforce_minimized_locks` runs at the reducer's set-tree / insert-at-index / `apply_atomic` sites right after the prune. A tampered locked size is snapped back and the delta repaid to the nearest unlocked sibling, so the parent's flex budget is conserved.
- **Affordance suppression:** resize handles are no longer generated on locked edges. This surfaced a latent bug: handle `parentIndex` was `resizeHandles.length`, correct only while every gap got a handle — now it's the child-array index (its only consumer, `onResizeMove`, looks up flanking children by it). The backwards `MinNodeSizePx` exemption is deleted (nothing left for it to serve — the guard is unconditional again) and a stale-handle guard added at resize-context creation.

## Known limitations (spec §6, deliberate)

- Lock is in flex units: minimized pixel height still scales with container resizes (pre-existing). Pixel-true headers need Option C's dock (or a ratio-aware pass).
- `findNextInsertLocation`'s heuristic is still not minimize-aware for blind inserts.

## Test plan

- [x] 7 new Rust tests: guard rejections for resize/move (source + destination)/swap/split, snap-back + redistribution, no-op on clean trees, and a reducer-level end-to-end (`layout_set_tree_snaps_tampered_minimized_size_back_to_its_lock`) proving the choke-point wiring and that the emitted event carries the healed tree.
- [x] 6 new vitest tests: lock recording/clearing (leaf + dissolved column), atomic resize rejection (the reported bug), split rejection, snap-back + sibling repayment, no-op.
- [x] Full `cargo test --bin agentmux-srv`: 1498 passed.
- [x] Full `npx vitest run frontend/layout`: 89 passed.

Refs #2179.

### DIFF
--- a/.changesets/1784196051-fix-layout-minimize-is-a-locked-state-reject-snap-back-resiz-uk7s.md
+++ b/.changesets/1784196051-fix-layout-minimize-is-a-locked-state-reject-snap-back-resiz-uk7s.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): minimize is a locked state — reject/snap-back resize and structural ops on minimized panes
--- a/agentmux-srv/src/backend/layout/mod.rs
+++ b/agentmux-srv/src/backend/layout/mod.rs
@@ -255,7 +255,12 @@ pub fn enforce_minimized_locks(root: &mut Option<LayoutNode>) -> usize {
                         .find(|&j| !locked_flags[j])
                         .or_else(|| (0..i).rev().find(|&j| !locked_flags[j]));
                     if let Some(j) = beneficiary {
-                        node.children[j].size += delta;
+                        // Floor at 1 flex unit (same floor as the slip/
+                        // undissolve restore paths): a tampered size BELOW
+                        // the lock makes the delta negative, and the
+                        // repayment must not drive the beneficiary's size
+                        // to zero or negative.
+                        node.children[j].size = (node.children[j].size + delta).max(1.0);
                     }
                     snapped += 1;
                 }
@@ -402,6 +407,12 @@ pub fn insert_node_at_index(
     }
 
     let parent = unsafe { &mut *current };
+    // Minimized is a locked state: an `index_arr` that resolves into a
+    // dissolved column (locked container) or onto a minimized leaf (which
+    // `ensure_group_node` would otherwise promote into a group) is rejected.
+    if is_node_locked(parent) {
+        return Err(LayoutError::NodeLocked { id: parent.id.clone() });
+    }
     ensure_group_node(parent);
     let insert_at = (clamped_idx + 1).min(parent.children.len());
     parent.children.insert(insert_at, node);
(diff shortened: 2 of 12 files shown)


AgentO-asaf:
All four review findings addressed in 391161a:

- **P1 (`handle_layout_insert_node`):** explicit-`parent_id` path now has a dedicated match arm rejecting a minimize-locked parent before the `data.is_none()` group check (which a dissolved column satisfied). New reducer-level test `layout_insert_node_rejects_minimize_locked_parent`.
- **P1 (`insert_node_at_index`):** the resolved insertion parent is now checked with `is_node_locked` before `ensure_group_node` — rejecting both an `index_arr` landing in a dissolved column and one landing on a minimized leaf (which would have been promoted into a group). New test `insert_node_at_index_rejects_locked_resolution` covers both.
- **P2 (negative-delta repayment):** both `enforceMinimizedLocks` and `enforce_minimized_locks` clamp the beneficiary at the 1-unit floor (same floor as the slip/undissolve restore paths). New tests on both sides.
- **P2 (type):** `enforceMinimizedLocks` signature widened to `LayoutNode | undefined` to match the `updateTree` call site.

Full suites re-run: 1501 Rust / 90 vitest, all green.

PR #2193, opened 2026-07-17
fix(cef): login CLI crashes on libcef's interposed close() — the real 'FD ownership violation' cause

AgentU-asaf:
## What

Every in-app provider login via `runCliLogin` on Linux crashed the CLI ~1ms after spawn with `Crashing due to FD ownership violation:` **before** it could print an OAuth URL. So "Login Again", `/login`, and the fresh-agent auto-login phase all silently no-oped — the keystone auth blocker behind `retro-agent-auth-relogin-noop-2026-07-01` and the PR #2192 symptoms.

## Root cause (this corrects the record)

The crash is emitted by **libcef.so, not Bun** — superseding the Bun-fd-guard theory in `SPEC_HOST_CLI_LOGIN_CAPTURE_2026_06_20.md` §2.

Chromium interposes a strong `close` symbol (`base/files/scoped_file_linux.cc`) that `IMMEDIATE_CRASH()`es when a `ScopedFD`-owned descriptor is closed. `portable_pty` 0.9's Unix spawn runs `close_random_fds()` in the forked child's `pre_exec`, calling the libc `close` **symbol** on every fd ≥ 3. In this process that resolves to libcef's interposed `close`, executing in the child while it still runs the host image (pre-exec) — so it crashes before `execve`, for **any** program.

**Proven live** (this session, DevTools `Runtime.evaluate` → `window.api.runCliLogin` against the running host):
- Target `/bin/sh` (no Bun runtime) → **identical crash stack**. Bun cannot be the cause.
- Crash addresses constant across runs and targets → the code at crash time is the parent host image, not the child's `main`.
- A standalone `portable_pty` 0.9 probe **not linking libcef** runs the pinned Claude 2.1.198 cleanly even with a 154-fd polluted table + the full host environ.
- `agentmux-srv` uses the same portable_pty and never hit this — it doesn't link libcef.

Full write-up: `docs/retros/RETRO_LOGIN_PTY_FD_OWNERSHIP_LIBCEF_CLOSE_2026_07_16.md`.

## Fix

`spawn_login_pty_unix` replaces portable_pty's Unix spawn for the login child: hand-rolled `openpty` + `std::process::Command` + `pre_exec` that does fd hygiene with the **raw `close_range(2)` syscall** (`CLOSE_RANGE_CLOEXEC`) instead of the `close` symbol. The raw syscall is invisible to symbol interposition, so Chromium's guard never runs; `CLOSE_RANGE_CLOEXEC` also just marks fds close-on-exec rather than closing them in-process. Pre-exec otherwise mirrors portable_pty (signal reset, `setsid`, `TIOCSCTTY`). Windows keeps the ConPTY/portable_pty path untouched via `#[cfg]`, with a shared `LoginChildWait` trait so both platforms use one reaper. macOS uses `fcntl(FD_CLOEXEC)`.

## Tests / verification

- `cargo check -p agentmux-cef` clean; new `spawn_login_pty_tests` (spawn+capture on a pty; child stdin is a TTY) pass.
- **End-to-end:** rebuilding the AppImage and driving `runCliLogin` against the pinned Claude CLI to confirm the `[login-pty]` capture shows the CLI's real output instead of the FD crash. _(Verification build in progress; result will be posted as a comment.)_

## Related

- Frontend companion #2192 makes a no-URL outcome user-visible, so a future spawn regression can't present as a silently dead button.
- Follow-up: any agentmux-cef child whose `pre_exec` calls `close`/`closefrom` is a latent copy of this bug — prefer `close_range(2)`.

### DIFF
--- a/.changesets/1784253719-fix-cef-login-cli-no-longer-crashes-on-libcef-interposed-clo-t84z.md
+++ b/.changesets/1784253719-fix-cef-login-cli-no-longer-crashes-on-libcef-interposed-clo-t84z.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): login CLI no longer crashes on libcef interposed close() (FD ownership violation)
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -639,6 +639,198 @@ pub async fn run_cli_login(
     Ok(serde_json::json!({ "auth_url": auth_url }))
 }
 
+/// Minimal wait/kill surface over the two login-child representations:
+/// portable_pty's boxed Child on Windows (ConPTY path) and std's Child on
+/// Unix (hand-rolled openpty spawn — see `spawn_login_pty_unix` for why).
+/// Lets `run_cli_login_pty` keep one shared reaper for both.
+trait LoginChildWait: Send {
+    fn try_exit_code(&mut self) -> std::io::Result<Option<i64>>;
+    fn kill_child(&mut self) -> std::io::Result<()>;
+    fn wait_child(&mut self);
+}
+
+#[cfg(windows)]
+impl LoginChildWait for Box<dyn portable_pty::Child + Send + Sync> {
+    fn try_exit_code(&mut self) -> std::io::Result<Option<i64>> {
+        Ok(self.try_wait()?.map(|status| status.exit_code() as i64))
+    }
+    fn kill_child(&mut self) -> std::io::Result<()> {
+        self.kill()
+    }
+    fn wait_child(&mut self) {
+        let _ = self.wait();
+    }
+}
+
+#[cfg(unix)]
+impl LoginChildWait for std::process::Child {
+    fn try_exit_code(&mut self) -> std::io::Result<Option<i64>> {
+        Ok(self
+            .try_wait()?
+            .map(|status| status.code().map(i64::from).unwrap_or(-1)))
+    }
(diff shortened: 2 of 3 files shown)


AgentU-asaf:
## ✅ End-to-end verification — fix confirmed on a real build

Rebuilt the AppImage with this change and drove `window.api.runCliLogin` against the pinned Claude CLI (2.1.198) via DevTools, against the **live host** (not a probe).

**Before the fix** (unpatched 0.53.6 host, same call):
```
[login-pty] Crashing due to FD ownership violation:
[login-pty] #0 … #2 close … (20-frame stack)
run_cli_login_pty: no auth URL captured within 15s
run_cli_login: 15650.7ms   → returned auth_url: null
```

**After the fix** (this branch):
```
run_cli_login: spawned (PTY), waiting for OAuth URL   cli=…/claude
[login-pty] Opening browser to sign in…
[login-pty] If the browser didn't open, visit: https://claude.com/cai/oauth/authorize?code=true&client_id=…&code_challenge=…&code_challenge_method=S256&state=…
run_cli_login_pty: captured auth URL
[fe] ipc run_cli_login 1099.7ms
```

- `grep -c "FD ownership violation"` in the host log: **0**
- The Claude CLI ran to completion, printed its login flow, and the host scraped the real OAuth URL — returned to the frontend in **1.1 s** instead of the previous 15 s timeout.

The interposed-`close` crash is gone; the raw `close_range(2)` pre-exec spawns the child cleanly.

PR #2197, opened 2026-07-17
feat(layout): minimize is a display mode (i3 pattern) — geometry derived, slip/dissolve deleted

AgentO-asaf:
## Summary

Research-backed redesign of pane minimize, tracked in #2179. The prior in-tree size-arithmetic model produced four bug classes in two weeks (direction flip #2176, resize-through-lock #2180, marker fields stranded on branches, cascading dissolves computing **negative sizes** — caught live by the layout doctor from #2184). Deep research across i3, tmux, Eclipse, IntelliJ, FlexLayout, AvalonDock, and Dockview — 21 sources, top 25 claims adversarially verified 3-0, report included in this PR (`docs/research/RESEARCH_PANE_MINIMIZE_BEST_PRACTICES_2026_07_16.md`) — found **no mature system stores minimize as in-tree size state**. tmux's maintainer rejected exactly our prior design in writing; i3 derives collapse geometry per render pass so drift is structurally impossible. This PR implements the research's recommendation (§7.3, the i3 pattern).

## The model

- **State:** one leaf-only flag, `minimized: true`. Stored flex sizes are **never touched** by minimize; restore = clear the flag — original geometry intact by construction.
- **Geometry derived per render pass** (`computeMainAxisAllocation`, pure + unit-tested): header-height strip in a Column parent; fixed-width chip (`MinimizedRowSlotWidthPx`, cross-axis clamped to header height) in a Row parent; a fully-minimized subtree renders as a stacked chip strip (one header height per leaf); chips scale down proportionally in tiny containers. Expanded siblings absorb the remainder proportionally to their untouched flex sizes; `pixelToSizeRatio` now describes only the flex-distributed space so resize-drag math stays exact.
- **Deleted wholesale:** slip/dissolve/undissolve structural surgery, `minimizedSize` restore arithmetic, `slipMinimize`/`columnDissolve`/`_slipAnchor` bookkeeping, and the frontend `enforceMinimizedLocks` lock layer (~450 lines of the machinery that generated every bug).
- **Kept:** reducer guards on both sides (`isNodeLocked`/`is_node_locked` keyed on the flag + legacy markers — minimized panes remain untargetable by resize/move/swap/split/insert), the last-expanded-pane guard, the collapsed-header reduction, and the layout doctor (I2 extended to the flag; I4–I7 now legacy detection). `balanceNode`'s carve-outs and Rust `enforce_minimized_locks` remain as inert protection for unmigrated trees.
- **Migration:** `rebuildMinimizedSet` converts legacy persisted state in place at load (`minimizedSize` → size restored + flag; `slipMinimize` → flag in place; `columnDissolve`/`minimizedLockedSize`/`_slipAnchor` → dropped), with a one-line log of the migrated count.

## Why each historical bug class is now structurally impossible

| Bug class | Why it can't recur |
|---|---|
| 1 — direction flip renders headers sideways | Chip geometry is derived; there is no stored orientation-coupled state to flip |
| 2 — resize through the header-height invariant | There is no squeezed size to resize and no handle generated on chip edges |
| 3 — marker fields stranded on branches | One leaf-only flag instead of four marker fields (doctor I2 still watches promotions) |
| 4 — cascade computes negative sizes | No steal-from-neighbor arithmetic exists; allocation is a pure function clamped ≥ 0 |

## Test plan

- [x] Full rewrite of `layoutMinimize.test.ts`: display-mode toggle (sizes untouched), no-structural-surgery assertion, last-expanded guard, 6 derived-geometry cases (Column/Row chips, stacked strips, scale-down clamp, no-minimize parity with plain proportional split), reducer-guard rejections, 3 legacy-migration cases, legacy balance carve-out
- [x] `layoutInvariants.test.ts`: flag-on-branch I2, flag-based ALL_LEAVES_LOCKED, distinct legacy/flag describe output; 0.53.6 corruption fixture unchanged
- [x] Rust: flag-based lock rejection + flag-on-branch + flag ALL_LEAVES_LOCKED; all legacy-marker tests still pass unchanged
- [x] `cargo test --bin agentmux-srv`: **1515 passed** · `npx vitest run frontend/layout`: **96 passed**

Also includes the research report and the spec update (§8, status superseded) per "no separate PRs for docs."

Refs #2179.

### DIFF
--- a/.changesets/1784269908-feat-layout-minimize-is-a-display-mode-derived-geometry-slip-5x81.md
+++ b/.changesets/1784269908-feat-layout-minimize-is-a-display-mode-derived-geometry-slip-5x81.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(layout): minimize is a display mode — derived geometry, slip/dissolve machinery deleted
--- a/agentmux-srv/src/backend/layout/mod.rs
+++ b/agentmux-srv/src/backend/layout/mod.rs
@@ -219,7 +219,15 @@ pub fn prune_dangling_block_refs(
 /// through the untyped `extra` catch-all on this side. Mirrors
 /// `isNodeLocked` in `frontend/layout/lib/layoutMinimize.ts`.
 pub fn is_node_locked(node: &LayoutNode) -> bool {
-    node.extra.contains_key("minimizedSize")
+    // Current model: the `minimized` display-mode flag (geometry derived at
+    // render on the frontend; stored sizes never touched). The remaining keys
+    // are legacy markers from the pre-display-mode model, recognized until
+    // persisted trees are migrated by the frontend's `rebuildMinimizedSet`.
+    node.extra
+        .get("minimized")
+        .and_then(|v| v.as_bool())
+        .unwrap_or(false)
+        || node.extra.contains_key("minimizedSize")
         || node.extra.contains_key("slipMinimize")
         || node.extra.contains_key("columnDissolve")
 }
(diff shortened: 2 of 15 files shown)


AgentO-asaf:
Both findings addressed: chip allocations are now gap-compensated (`header + gapSizePx` per chip, `MinimizedRowSlotWidthPx + gap` in Row parents, cross-axis clamp included) so the inset inner box comes out exactly header-sized — with a dedicated test covering both parent directions; and the `minimizedLockedSize` doc now correctly describes it as a legacy field read only by the Rust-side enforcement. Suites re-run: 97 vitest green.

AgentO-asaf:
Round-2 P1 addressed: the slip/dissolve migration paths now restore `node.size` from the marker's recorded `originalRowSize` (when positive), matching the `minimizedSize` path — a migrated slipped leaf or dissolved branch no longer keeps its collapsed-era sliver size, and the guard also heals the negative-size corruption from the cascade bug. Migration test extended to assert both restored sizes (including a `-0.0039` dissolved-branch fixture). 98 vitest green.

AgentO-asaf:
Round-3 P1s addressed: (1) every reducer guard on both sides now checks `isEffectivelyMinimized`/`is_effectively_minimized` — a fully-collapsed branch is as untargetable as a minimized leaf, since its stored size becomes load-bearing again on child restore; covered by new TS + Rust tests (resize atomic-reject and both move endpoints). (2) Migration no longer restores `originalRowSize` across unit spaces — a slipped/dissolved node's size heals to the mean of its current siblings (fallback `DefaultNodeSize`), asserted against the actual nested parent including the `-0.0039` corrupted fixture. Suites: 1517 cargo / 98 vitest green.

PR #2198, opened 2026-07-17
fix(menu): route pane context-menu submenus through the paintable-area framework

AgentA-asaf:
## Problem

Right-click a pane header near the window's right edge and hover **"Replace With..."** — the second-level submenu paints off-window and is cut off. Same for "Pane Color" and any `type: "submenu"` item.

## Root cause

`showJsContextMenu` (`frontend/util/cef-api.ts`) — the renderer behind every pane right-click menu — routes its **top-level** menu through `computeMenuPosition`, but its **submenu branch** kept pre-framework CSS: `position:absolute; left:100%; top:0` inside the hovered row. No flip, no shift, no size cap, and no `assertMenuInPaintableArea` call, so the framework's dev guard was structurally blind to it (the guard is opt-in per call site). This is the SPEC_MENU_PAINTABLE_AREA_GUARD_2026_05_20 Phase 3 ("manual-math surfaces") site that was missed.

Full analysis: `docs/analysis/submenu-offscreen-second-level-2026-07-16.md` (bundled).

## Fix

- On row hover, the submenu is placed via `computeMenuPosition({anchor: row.getBoundingClientRect(), placement: "right-start", avoidNativePanes: false})` — flip sends it left of the parent near the right edge, shift pulls it up near the bottom. Coordinates are `position:fixed`, so the DOM nesting the hover logic needs doesn't affect placement. Held `visibility:hidden` until placed (same pane-overlay-clip rationale as the top level) and guarded with `assertMenuInPaintableArea(sub, "context-submenu")`.
- Secondary gap fixed in the same pass: `computeMenuPosition`'s `maxHeight` cap was returned but dropped by every consumer. Both `showJsContextMenu` levels and `FlyoutMenu`'s `styleToString` now apply `max-height` + `overflow-y:auto` (+ `justify-content:flex-start`, since `.menu`'s `flex-end` makes overflow unreachable in a scroll container — a no-op when the menu fits). `maxWidth` is deliberately **not** applied: an inline max-width would override, and can loosen, the `.menu` 400px CSS cap; horizontal fit is already guaranteed by flip+shift for menus at or under that cap.

## Verification

- `tsc --noEmit` clean; `vitest run` on `cef-api.test.ts` + `menu-position.test.ts`: 21/21 pass.
- Manual repro to check on a live build: window near the screen's right edge → right-click a pane header in the window's right half → hover "Replace With..." and "Pane Color" — both must open to the LEFT of the parent and stay fully visible; dev console shows no `[menu-guard]` errors.

### DIFF
--- a/.changesets/1784285080-fix-menu-context-submenu-offscreen.md
+++ b/.changesets/1784285080-fix-menu-context-submenu-offscreen.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(menu): route pane context-menu submenus through the paintable-area framework — "Replace With..."/"Pane Color" no longer cut off at the window edge; menus taller than the free space now scroll internally
--- a/docs/analysis/submenu-offscreen-second-level-2026-07-16.md
+++ b/docs/analysis/submenu-offscreen-second-level-2026-07-16.md
@@ -1,0 +1,132 @@
+# Second-level submenus escape the paintable-area framework
+
+**Date:** 2026-07-16
+**Symptom (user report):** right-click a pane header → hover **"Replace With..."** — the
+second-level submenu gets cut off at the window edge. The top-level menu is always
+placed correctly.
+**Verdict:** the paintable-area framework is working as designed; the pane
+context-menu renderer's **second level never calls it**. The submenu is positioned by
+two lines of hand-rolled CSS from before the framework existed.
+
+---
+
+## 1. What the framework guarantees (and to whom)
+
+`frontend/app/util/menu-position.ts` (SPEC_MENU_PAINTABLE_AREA_GUARD_2026_05_20) is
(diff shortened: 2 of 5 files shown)


AgentA-asaf:
Addressed both review items in 54f655de:

- **P2 fallback**: restored the static `left:100%;top:0` CSS (+ `row.position:relative`) as the degraded fallback — a `computeMenuPosition` rejection now reveals an edge-anchored submenu rather than an unpositioned one; successful placement overrides it with fixed viewport coords.
- **P2 coverage**: `showJsContextMenu` is now exported for tests, with 4 new cases in `cef-api.test.ts`: fallback state at render, hidden-until-placed reveal (fixed px coords + max-height cap + overflow), stale-promise guard on early mouseleave, and top-level framework placement. `tsc --noEmit` clean, 8/8 tests pass.

PR #2200, opened 2026-07-17
feat(launcher): unix teardown-backstop parity (issue #2188 Part 2)

AgentO-asaf:
## Summary

Part 2 of #2188: ports the wedged-UI-thread teardown backstop to macOS/Linux. The platform-neutral pieces (`teardown_backstop.rs`'s armed state machine, `ui_liveness.rs`'s probe telemetry, and the arm/disarm hooks on `PoolDrained`/`OrphanInstance`/`WindowOpened` in `ipc/server.rs`) already ran unconsumed on unix — this wires up the missing supervisor-side tick and a real kill mechanism.

## Changes

- **`process_group(0)` added to both unix spawn paths** (`spawn_host_unix`, `srv_spawner::spawn_srv`) — a code survey ahead of this change confirmed **no unix child of the launcher was ever placed in its own process group**, so there was no safe target for a group-kill. Host and srv each become their own group leader (they spawn concurrently via `tokio::join!`, so they can't share one group without coupling the two spawn call sites).
- **`kill_process_group_forcefully`** (`host_spawn.rs`): `kill(-pgid, SIGKILL)` — the unix analog of `TerminateJobObject`, scoped to exactly what the launcher spawned (I2/I3 blast-radius bound).
- **Two new `select!` arms in `run_unix`**, ported directly from `run_windows`'s equivalent Phase 1/2 logic: a 60s UI-liveness prober (`ProbeUiThread` via `host_pipe.try_send_command_no_buffer`, fail-fast so a probe sent during a crash-restart gap can't age into a false miss) and a 5s teardown-check tick polling `teardown_backstop::should_teardown`, which on true kills both process groups and exits with the same distinct code (86) Windows uses.
- **`teardown_backstop::disarm()`** added to the host-exit arm — every host exit (clean, crash, or future recycle) stands the armed machine down, the crash-restart-gap false-positive guard.
- `launcher_setup`'s returned tuple widened to include `host_pipe` (unix previously discarded it; the new prober arm needs it).

## Test plan

- [x] 3 new tests in `host_spawn.rs`, run against **real spawned processes** rather than asserting on the builder call alone (a wrong `process_group` argument would compile fine and silently fail to create a new group): group-leader pgid check via `getpgid`, actual SIGKILL delivery check, and a no-op-on-already-exited-child case.
- [x] Full `cargo test -p agentmux-launcher`: 208 passed, no regressions. Existing `teardown_backstop.rs`/`ui_liveness.rs` state-machine unit tests untouched and pass.
- [x] `cargo check --workspace`: clean.
- [ ] **Not yet live-verified**: the manual recipe from #2188 (`AGENTMUX_DEBUG_HANG=1` + `debug:hang_ui` + last-window close via direct IPC → watch for `ARMED` → `terminating` within grace + probe intervals → zero surviving processes) hasn't been run on real hardware yet.

Part 1 of #2188 (verifying #2186's `on_before_close` changes are correct live on macOS) is a separate, manual-verification track — not touched by this PR.

Refs #2188.


AgentO-asaf:
P1 addressed: `.process_group(0)` genuinely breaks the documented "shares our foreground process group" assumption — a terminal Ctrl+C no longer reaches host/srv directly (only the launcher does, since it's the only one left in the terminal's foreground group). No functional regression — shutdown still completes correctly via the launcher's own explicit `terminate_child_gracefully` SIGTERM in the cleanup epilogue — but the doc comments describing WHY the group-shutdown guards exist were now inaccurate. Rewrote them (module header + both wait() arm guards) to describe the actual post-change signal-propagation path, and strengthened the process-group test to explicitly assert the child's pgid differs from our own (not just that it equals its own pid) — the behavioral half of the invariant this whole PR depends on. 208 tests green.

P0 acknowledged — the do-not-merge note is intentional, this PR is held for @asafebgi's explicit approval per standing instruction, not merging regardless of review outcome.

PR #2201, opened 2026-07-17
fix(activity-dock): group subagent rows by workflow, one dock row per Agent-tool call

AgentY-asaf:
## Summary
- Live-reported: the pinned activity dock showed dozens of subagent rows for what was actually 1-2 Agent-tool calls.
- Root cause: `subagent-adapter.ts` mapped every `ActiveSubagent` to its own `PinnedActivity`, regardless of whether it was a standalone Task-tool subagent or one member of a Workflow-tool batch spawning dozens at once (the Swarm docs cite 45 observed live in one run).
- The Swarm pane already solved exactly this with `groupSubagentsByWorkflow` (shared `workflow_id` → one `WorkflowGroup`; same-name loose subagents → one `NameGroup`) — the dock adapter never adopted it. This PR reuses that grouping and its `groupCacheKey` identity scheme directly instead of re-deriving the rules, so the dock and the Swarm tree always agree on what counts as "one call".
- Known simplification (documented in a code comment): a grouped row's expanded view and tail text show only the most-recently-active member, not every member's transcript — `ActivityRow.tsx` renders one `subagent` per row today; a real multi-member expanded view is a follow-up, not required to fix the row-count explosion.

## Test plan
- [x] `tsc --noEmit` — passes
- [x] New tests: workflow-batch collapse, workflow-group-done-once-all-terminal, same-name loose-subagent collapse, lone-subagent-stays-ungrouped — all pass
- [x] Existing `subagent-adapter.test.ts` (unchanged behavior for all-loose input) — passes
- [x] `swarm-model.test.ts` (shared grouping logic, unmodified) — 34/34 pass

### DIFF
--- a/.changesets/1784287906-fix-activity-dock-group-subagent-rows-by-workflow-one-dock-r-ih5r.md
+++ b/.changesets/1784287906-fix-activity-dock-group-subagent-rows-by-workflow-one-dock-r-ih5r.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(activity-dock): group subagent rows by workflow, one dock row per Agent-tool call
--- a/frontend/app/view/agent/activity/subagent-adapter.test.ts
+++ b/frontend/app/view/agent/activity/subagent-adapter.test.ts
@@ -67,4 +67,52 @@ describe("subagentActivities", () => {
     it("returns an empty array when no subagents match the block", () => {
         expect(subagentActivities([mk({ agent_id: "a1", parent_block_id: "block-2" })], "block-1")).toEqual([]);
     });
+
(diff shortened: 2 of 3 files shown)


AgentY-asaf:
Closing in favor of #2203 (AgentY-asaf), which landed the same fix independently and more completely: it adds a proper \`subagentGroup\` field with a full multi-member roster in the expanded view (my version showed only one representative member — noted as a known simplification), and more accurate status derivation (distinguishes \`stopped\` for an all-abandoned group instead of collapsing it to \`done\`). Same core approach otherwise (reusing the Swarm pane's \`groupSubagentsByWorkflow\` + \`groupCacheKey\`), so no conflict in direction — just redundant now.

PR #2204, opened 2026-07-17
feat(launcher): Linux teardown backstop — process-group isolation + wedged-host kill

### LINKED ISSUE #2189: Linux: verify window-close/quit lifecycle chain + teardown-backstop parity
Platform follow-up to the Windows lifecycle work that landed 2026-07-16 (#2186 close-path fixes, #2187 teardown backstop Phase 2). Sibling of the macOS issue (#2188) — same two parts, Linux specifics below.

## Part 1 — verify the close/quit chain (the #2186 fixes are UNVERIFIED here)

On Linux, `window.close()` → `can_close` → `on_before_close` runs the full cleanup chain (unlike Windows/CEF-148, which parks the browser and never fires it). That makes `on_before_close` the **live** path here — and #2186 changed it:

- `agentmux-cef/src/client/lifecycle.rs::on_before_close` — the srv-notify block (`backend_close_window` + retry) now runs **unconditionally** (it used to be skipped for the last window), and `quit_message_loop()` for the last window is **deferred** via `ui_tasks::QuitMessageLoopTask`, posted back to the UI thread from the background notify thread once the notify completes.

Verify live on Linux (X11 and, if available, Wayland):
1. Multi-window session: close a non-last window → srv log shows the full `srv_window_closed → delete_workspace saga` sequence; `db_window` row removed.
2. Close the **last** window → same saga fires (this was the #2186 leak on Windows), the app quits cleanly afterward (the deferred `QuitMessageLoopTask` path — watch for a hang; it is new code that Windows never exercises), and `db_window` count matches what was open on next launch (no ghost windows resurrected by crash-reproject).
3. Regression: floating-pane/pool closes during the quit cascade still work (they share the same handler). Note Linux floaters have their own close lifecycle (`SPEC_LINUX_FLOATING_PANE_TEAROFF_2026_05_30.md`) — include one floater in the pass.

This also covers the standing "#1969 residual C: non-Windows close-path verification" item (session task #15): confirm `Client.windowids` / renderer-process count return to baseline across open+close cycles, per `test/e2e/window-close-baseline.e2e.test.ts`'s assertions done manually.

## Part 2 — teardown backstop parity (deferred by design)

`SPEC_LAUNCHER_TEARDOWN_BACKSTOP_2026_07_11.md` (Phase 2 merged for Windows in #2187) explicitly defers Linux parity until after the Windows bake. The portable pieces already exist and are platform-neutral:
- `agentmux-launcher/src/teardown_backstop.rs` — the armed state machine (unit-tested).
- `agentmux-launcher/src/ipc/server.rs` — arm/disarm hooks on the post-reducer events (PoolDrained / OrphanInstance / WindowOpened).
- `agentmux-launcher/src/ui_liveness.rs` — probe telemetry + `consecutive_misses`.

What's missing on Linux: the supervisor-side check tick + teardown execution (`agentmux-launcher/src/supervisor/windows.rs` has the 5s tick calling `teardown_backstop::should_teardown(misses)` then `TerminateJobObject(J0)`, exit code 86). The unix supervisor needs the equivalent tick and a platform kill of the spawned tree — likely `killpg`/process-group or explicit host+srv PID kills, honoring the same I2/I3-style blast-radius bound: only processes this launcher spawned, never a PID outside its own tree.

Also verify Phase 1 telemetry works here at all: `[ui-liveness] UI thread alive — probe nonce=N rtt=…ms` lines in the launcher log during a normal session (the probe reply rides `wrap_task!` on the CEF UI thread — platform-neutral in theory, unverified on Linux).

Verification recipe (from the Windows run): launch with `AGENTMUX_DEBUG_HANG=1`, invoke the `debug:hang_ui` IPC command, close the last window via direct IPC (NOT via CDP — CDP response routing needs the UI thread you just parked and will deadlock your driver), then watch the launcher log for `ARMED (OrphanInstance drift…)` → `host wedged with zero user windows — terminating job` within ~30s grace + 2 probe intervals, and confirm zero surviving processes.

## References
- `docs/specs/SPEC_LAUNCHER_TEARDOWN_BACKSTOP_2026_07_11.md`
- `docs/retro/retro-last-window-close-quit-race-2026-07-16.md` (#2186 root cause + fix)
- `docs/specs/SPEC_WINDOW_LIFECYCLE_CLOSE_RELIABILITY_2026_07_04.md` §4c
- `docs/status/STATUS_LIFECYCLE_AND_CRASH_ARCHITECTURE_2026_07_16.md` §6 item 2

AgentU-asaf:
Closes #2189.

## Part 1 — verify the Linux close/quit chain (#2186 fallout)

No code changes needed. Live-verified on `task dev`:
- Non-last window close → full srv saga (`srv_window_closed → delete_workspace → tab_deleted → active_tab_changed → workspace_deleted → saga_completed`), `db_window` row correctly removed.
- Last window close → clean quit through the pool-window cascade, `db_window` ends empty, zero orphaned processes.
- Floating pane close mid-cascade → clean (exercised for free during the last-window test).
- `CloseWindowTask`'s "main" double-notify branch is `#[cfg(target_os = "windows")]`-only — structurally can't fire on Linux.

**A false start worth reading:** initial testing looked like a serious regression (`on_before_close` never firing) — turned out to be two `task dev` instances colliding on the same default DevTools port, with my driver silently talking to the wrong (stale) one the whole time. Full writeup, including how it was caught and corrected, in the retro linked below — it's the same "confusing log sinks" trap the original #2186 retro warned about, one level up (confusing *live instances*, not just logs).

## Part 2 — Linux teardown-backstop execution (new code)

`SPEC_LAUNCHER_TEARDOWN_BACKSTOP_2026_07_11`'s Phase 1 (arm/disarm state machine, UI-liveness probing) was already platform-neutral and unused on Linux; only the Windows-only Phase 2 execution (`TerminateJobObject`) had no Linux equivalent.

Linux has no Job Object, and — critically — launcher/host/srv currently all share the launcher's own ambient process group (intentional, for terminal Ctrl+C). A naive `killpg` on "our own group" would kill the launcher itself mid-teardown, violating I2/I3.

**Fix:** give host+srv their own separate process group at spawn time, leaving the launcher's untouched:
- srv gets `.process_group(0)` (Linux-only) — becomes leader of a brand-new group.
- host joins that same group (`.process_group(backstop_pgid)`) on every spawn/relaunch.
- The launcher's `run_unix` loop gains the same two ticks `run_windows` already has (60s UI-liveness probe, 5s teardown check). On trigger: `killpg(pgid, SIGKILL)` then `break` (not an immediate exit) so the existing saga-cleanup path still runs — mirrors Windows' control flow exactly.
- `backstop_pgid` is `Option<i32>`, never a bare `0` — `killpg(0, ...)` means "my own group," so a stray literal 0 could kill the launcher.
- Scoped to `#[cfg(target_os = "linux")]` specifically (not `unix`) — no behavior change on macOS, which is #2188's separate follow-up.

### End-to-end verification (live)

`AGENTMUX_DEBUG_HANG=1` + `debug:hang_ui` to park the UI thread, then closed the last window via **raw HTTP IPC** (not CDP — `window.api` calls actually ride a plain Bearer-token HTTP server the host runs, not DevTools Protocol; a CDP round-trip would deadlock against a parked UI thread, this doesn't).

```
ARMED instantly (report_window_closed doesn't depend on the wedge)
+84s  probe nonce=3 unanswered
+144s probe nonce=4 unanswered
+149s teardown fires — killpg, launcher exits cleanly with code 86
```

149s total, inside the spec's ~150s worst-case envelope (30s grace + 2×60s probes). Post-teardown: **zero** surviving processes in the killed group; the launcher exited via its own graceful post-loop path (not killed) in a completely different process group the whole time; the driving shell (outside the tree entirely) was unaffected throughout — the concrete proof I2/I3 held.

Full trace + methodology in `docs/retro/retro-linux-close-quit-teardown-verification-2026-07-17.md`.

## Test plan

- `cargo test -p agentmux-launcher` — 205 passed.
- `cargo check` clean on both changed crates, no new warnings.
- Live end-to-end verification as described above (not just unit tests — this is process-kill logic, so it was exercised against a real running instance, not simulated).

### DIFF
--- a/.changesets/1784290408-feat-launcher-linux-teardown-backstop-process-group-isolatio-a44k.md
+++ b/.changesets/1784290408-feat-launcher-linux-teardown-backstop-process-group-isolatio-a44k.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(launcher): Linux teardown backstop — process-group isolation + wedged-host kill (issue #2189)
--- a/agentmux-launcher/src/host_spawn.rs
+++ b/agentmux-launcher/src/host_spawn.rs
@@ -107,6 +107,16 @@ pub(crate) fn spawn_host_supervised(
 /// byte-identical to the legacy direct-invoke (`task dev:standalone`)
 /// path. Phase 2 will set it once the production runtime/ layout lands
 /// on macOS.
+/// `backstop_pgid`: on Linux, the process group host must join at spawn (see
+/// SPEC_LAUNCHER_TEARDOWN_BACKSTOP_2026_07_11, issue #2189) — srv's own pid,
+/// captured once after srv spawns (`srv_spawner.rs`'s `.process_group(0)`
+/// makes srv the leader of that group). Passed on every spawn, including
+/// crash-budget relaunches, since srv itself never restarts mid-session so
+/// the value never goes stale. `None` = teardown backstop disabled for this
+/// session (srv had no pid to key off of) — host spawns normally, just stays
+/// in the launcher's own ambient group as it did before this feature existed.
+/// Unused on macOS — Linux is the only platform this backstop is verified on
+/// so far (issue #2188 is the macOS follow-up).
 #[cfg(not(target_os = "windows"))]
 pub(crate) fn spawn_host_unix(
     real_exe: &std::path::Path,
@@ -114,6 +124,7 @@ pub(crate) fn spawn_host_unix(
     srv: &crate::srv_spawner::SrvSpawnResult,
     host_env: &[(&'static str, std::ffi::OsString)],
     disable_gpu: bool,
+    #[cfg_attr(not(target_os = "linux"), allow(unused_variables))] backstop_pgid: Option<i32>,
 ) -> Option<tokio::process::Child> {
     let mut host_cmd = tokio::process::Command::new(real_exe);
     host_cmd
@@ -163,6 +174,21 @@ pub(crate) fn spawn_host_unix(
                 libc::prctl(libc::PR_SET_PDEATHSIG, libc::SIGKILL, 0, 0, 0);
                 Ok(())
             });
+        }
+        // SPEC_LAUNCHER_TEARDOWN_BACKSTOP_2026_07_11 (issue #2189): join srv's
+        // process group instead of the launcher's ambient one — see the
+        // matching comment in srv_spawner.rs for the full rationale. Race-free
+        // because `run_unix` only calls this after srv's ESTART handshake, by
+        // which point srv's own `.process_group(0)` has long since taken
+        // effect (setpgid's target must be an existing pgid in the same
+        // session — this is NOT just "after fork", it's "after srv's pre_exec
+        // ran", which the ESTART wait guarantees). Applied on every relaunch
+        // (crash-budget retries reuse the same `backstop_pgid`). `None` (srv
+        // had no pid) skips this — host stays in the launcher's own group,
+        // same as before this feature existed; never falls back to a bare `0`
+        // (see the caller's comment on why that would be unsafe).
+        if let Some(pgid) = backstop_pgid {
+            host_cmd.process_group(pgid);
         }
     }
     match host_cmd.spawn() {
--- a/agentmux-launcher/src/srv_spawner.rs
+++ b/agentmux-launcher/src/srv_spawner.rs
@@ -354,6 +354,29 @@ pub async fn spawn_srv(
         }
     }
 
+    // SPEC_LAUNCHER_TEARDOWN_BACKSTOP_2026_07_11 (Linux Phase 2, issue #2189):
+    // srv becomes the leader of a BRAND NEW process group (pgid == srv's own
+    // pid) instead of inheriting the launcher's ambient group. host joins this
+    // same group at its own spawn (host_spawn.rs). This creates a bounded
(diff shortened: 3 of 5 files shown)


AgentU-asaf:
Closing in favor of #2200, which merged while this PR was in flight and already covers this ground more broadly (Linux **and** macOS, vs. this PR's Linux-only scope) — implementing the same mechanism (process-group isolation + wedged-host `killpg` teardown) via two independently-owned process groups (host's own, srv's own) rather than this PR's single shared group.

Reagent's outstanding findings on this PR (P1: process-group isolation loses the terminal-Ctrl+C fast path — #2200 hit the identical finding and resolved it by updating the doc comment, which I agree is the right call; P2: `killpg`'s return value discarded) are being folded into a small, focused follow-up PR against `main` as-is now, which also:
- Ports my P1 fix from this PR's last commit (SIGTERM-first, giving srv's own graceful shutdown handler — which reaches agent-spawned shells via their own tracked pid/pgid, independent of ambient process-group membership — a real chance to clean those up before the hard SIGKILL escalation) onto #2200's two-group design.
- Applies the same return-value check to `kill_process_group_forcefully`.
- Re-runs my live end-to-end wedge-teardown verification against the actual merged implementation (this PR's own extensive live verification was against my single-group design, which is no longer what's on `main`).

No further action needed on this PR.

PR #2211, opened 2026-07-18
feat(layout): restore minimize slip-docking + fix insert-into-minimized-leaf corruption

AgentO-asaf:
## Summary

Two fixes from live testing of the display-mode minimize redesign (#2197).

### 1. Slip-docking restoration

The display-mode redesign incorrectly deleted a real, deliberately spec'd product requirement (`SPEC_PANE_MINIMIZE_REFINEMENTS_2026_06_24.md`, `SPEC_PANE_MINIMIZE_COLUMN_DISSOLVE_2026_06_27.md`) along with the buggy tree-surgery mechanism that used to implement it: a minimized pane's header should dock onto an adjacent pane (which absorbs its freed space), not render as its own separate narrow strip. Full retro on why this got lost and why restoring it correctly wasn't a hard problem: `docs/retro/retro-minimize-display-mode-lost-slip-requirement-2026-07-17.md` (included here).

Restored as pure derived geometry — no tree mutation, no restore-context bookkeeping, matching the rest of the redesign's philosophy:

- **`resolveRowSlipTargets`**: for each Row child, finds the nearest non-minimized sibling to dock onto (right-preferred, else left, scanning past other minimized siblings) — unifies the original spec's per-pane "slip" and cascading full-column "dissolve" into one rule via `isEffectivelyMinimized`.
- **`computeMainAxisAllocation`** extended with an optional `slipChildIds` set: a slip child contributes zero pixels to its row's width split, so the freed space flows entirely to its docking target.
- **A new docking pass** in `updateTreeHelper` shrinks the target's rect from the top and stacks the slip children's chip rects in that reserved space.
- Resize handles correctly disappear on a docked edge.

### 2. Insert-into-minimized-leaf corruption fix

Live testing surfaced a real corruption: a blind pane insert (heuristic-targeted, no explicit destination) had no concept of "minimized" and could select an already-minimized leaf as its insertion point. Promoting that leaf into a branch moved its data into a fresh intermediate child (inheriting the leaf's id) but left the `minimized` flag stranded on the outer node — now a branch, violating the layout doctor's leaf-only I2. The separately-maintained `minimizedNodeIds` cache then kept referencing the inherited id from before promotion, so an **expanded** pane showed a "Restore" button.

Root-caused directly from the layout doctor's own `MIN_MARKER_ON_BRANCH` log line, timestamped immediately after the triggering `CreateBlock` action.

Three-part fix, mirrored TS + Rust:
- `findNextInsertLocation`/`collect_insert_candidates` skip minimize-locked candidates entirely; `insertNode`/`insertNodeAtIndex` (both sides) gained explicit reject guards as defense-in-depth.
- `addIntermediateNode`/`ensure_group_node` hardened to correctly move the flag onto the intermediate child if ever reached anyway.
- `minimizedNodeIds` converted from an incrementally-maintained cache to a value **derived fresh from the tree on every render pass** — closes the entire desync bug class structurally, not just this one path.

### Platform scope

Both fixes are pure frontend TS + cross-platform srv Rust — no `#[cfg]`/platform-specific code paths. Applies identically on macOS/Windows/Linux.

## Test plan

- [x] 12 new tests for slip-docking (9 `resolveRowSlipTargets` edge cases, 2 width-reclaim math, 1 end-to-end integration test via the real `LayoutModel` reproducing the exact user repro)
- [x] 5 new tests for the corruption fix (2 TS pure-function, 1 `addIntermediateNode` flag-preservation, 1 TS end-to-end via the real `LayoutModel`, 1 Rust mirror) — each verified red-before/green-after against the actual reported bugs
- [x] Full `cargo test --bin agentmux-srv`: 1518 passed
- [x] Full `npx vitest run frontend/layout`: 116 passed

### DIFF
--- a/.changesets/1784349691-feat-layout-restore-minimize-slip-docking-fix-insert-into-mi-ihh2.md
+++ b/.changesets/1784349691-feat-layout-restore-minimize-slip-docking-fix-insert-into-mi-ihh2.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(layout): restore minimize slip-docking + fix insert-into-minimized-leaf corruption
--- a/agentmux-srv/src/backend/layout/mod.rs
+++ b/agentmux-srv/src/backend/layout/mod.rs
@@ -476,6 +476,18 @@ fn collect_insert_candidates<'a>(
     depth: usize,
     out: &mut Vec<InsertCandidate<'a>>,
 ) {
+    // Minimized is a locked state: never a blind-insert target. Without
+    // this, `ensure_group_node` promoting a minimized leaf into a branch
+    // could leave the tree in a state where an agent/API-driven insert
+    // silently un-collapses a fully-minimized subtree (every child
+    // minimized → one child isn't anymore, with no user action on that
+    // branch at all). Mirrors `findNextInsertLocationHelper`'s guard in
+    // `frontend/layout/lib/layoutNode.ts` (2026-07-18). For a branch this
+    // also skips recursing into its children — they're all effectively
+    // minimized too by definition, so recursion would find nothing anyway.
+    if is_effectively_minimized(node) {
+        return;
+    }
     // Leaf node (has data but no children) — TS returns index 1.
     if node.data.is_some() && node.children.is_empty() {
         out.push(InsertCandidate { node, index: 1, depth });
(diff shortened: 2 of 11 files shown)


AgentO-asaf:
P1 addressed: chip heights in the slip-docking pass now scale down proportionally when a slip group's combined stacked height exceeds its target's available cross-axis space — mirrors `computeMainAxisAllocation`'s existing `scale` factor for its analogous fixed-chip path. New test precisely isolates the exact overflow (verified red with just the scale factor neutralized, green restored). 117 vitest / cargo test unaffected (Rust side untouched by this fix — pure frontend geometry).

PR #2242, opened 2026-07-20
feat(lan-discovery): UDP broadcast probe fallback for filtered mDNS

AgentY-asaf:
## Summary

Layer 2 of the LAN-discovery reliability plan: a UDP broadcast probe/responder
fallback for networks where mDNS multicast is filtered (corporate/guest
WiFi). This is the desktop/backend (Rust) half; a Flutter mobile agent is
building the client half against the identical wire protocol in parallel.

Folded into the existing `LanDiscovery` struct (`agentmux-srv/src/backend/lan_discovery.rs`)
rather than a sibling module, since it already owns `instance_id` /
`hostname` / `version` / `port` / `auth_key` and already spawns background
tasks from `start()`.

### Wire protocol (JSON over UDP, fixed port **47891**)

Port 47891 is above both the well-known and IANA-registered ranges — fixed
(so a broadcast probe can target it) and easy to grep in logs/firewall
rules. Documented in a code comment at `UDP_DISCOVERY_PORT`.

Probe (mobile → desktop, broadcast to `255.255.255.255:47891` or subnet
broadcast):
```json
{"type":"agentmux_discover","v":1}
```

Response (desktop → mobile, unicast back to the probe's `recv_from` source
address):
```json
{"type":"agentmux_discover_response","v":1,"instance_id":"...","hostname":"...","version":"...","port":12345,"auth_key":"..."}
```

Malformed/unrelated/wrong-version packets are silently ignored
(`is_valid_probe`) — logged at `tracing::debug!` only, since this port will
see LAN noise and the task explicitly calls for non-chatty logging.

### Gating & shutdown

Gated on the exact same `network_lan_discovery` setting as the mDNS daemon,
including on a **live toggle**, not just at boot. `LanDiscovery::shutdown()`
now also fires a `tokio::sync::oneshot` cancellation channel that the UDP
responder's `tokio::select!` races against `recv_from`, so `apply(false)`
actually stops answering probes — previously `shutdown()` only touched the
mDNS daemon.

### SO_REUSEADDR trade-off

Deliberately **not** set. This codebase explicitly supports running multiple
AgentMux instances on one host simultaneously; on Windows, `SO_REUSEADDR`
lets a second process silently steal a UDP port already owned by another
process — inappropriate for a socket that hands out `auth_key` in its
responses. If the bind to :47891 fails (most likely a second local instance
that also has LAN discovery enabled), the UDP task logs at `debug!` and
exits quietly — `LanDiscovery::start()` still succeeds and mDNS discovery is
unaffected. This mirrors the security posture already documented at
`find_agent()`: the auth_key is broadcast because LAN traffic here is
already a trusted-network assumption, and this doesn't add new exposure.

`set_broadcast(true)` was considered but is not needed: it's only required
to *send* to a broadcast address; this socket only receives probes and
replies via plain unicast.

### Testing

Extracted two small, directly-testable functions used by both the
production loop and the tests: `is_valid_probe(bytes) -> bool` and the free
function `probe_response_json(...) -> serde_json::Value` (wrapped by
`LanDiscovery::build_probe_response(&self)` for the production path). Added
a `#[tokio::test]` round-trip test binding two `UdpSocket`s on
`127.0.0.1:0` (ephemeral, not the fixed port, to avoid CI collisions) plus
unit tests for probe validation edge cases, in the existing
`#[cfg(test)] mod tests` block.

## Verification

- `cargo build -p agentmux-srv` — succeeds, no new warnings from
  `lan_discovery.rs`.
- `cargo test -p agentmux-srv lan_discovery` — 12/12 passed.
- `cargo test -p agentmux-srv` (full suite) — 1558 passed, 0 failed, 4
  ignored (pre-existing, unrelated).

## Test plan

- [x] `cargo build -p agentmux-srv`
- [x] `cargo test -p agentmux-srv lan_discovery` (new tests)
- [x] `cargo test -p agentmux-srv` (full suite, no regressions)

### DIFF
--- a/.changesets/1784573464-feat-lan-discovery-udp-broadcast-probe-fallback-for-filtered-h4w8.md
+++ b/.changesets/1784573464-feat-lan-discovery-udp-broadcast-probe-fallback-for-filtered-h4w8.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(lan-discovery): UDP broadcast probe fallback for filtered mDNS
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -13,15 +13,71 @@ use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use mdns_sd::{ServiceDaemon, ServiceEvent, ServiceInfo};
-use parking_lot::RwLock;
+use parking_lot::{Mutex, RwLock};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use tokio::net::UdpSocket;
+use tokio::sync::oneshot;
 
 use super::eventbus::{EventBus, WSEventType};
 
 const SERVICE_TYPE: &str = "_agentmux._tcp.local.";
 const LAN_AGENT_CACHE_TTL_SECS: u64 = 60;
 const LAN_PEER_QUERY_TIMEOUT_SECS: u64 = 2;
+
+/// UDP broadcast discovery fallback (Layer 2), for LANs where mDNS multicast
+/// is filtered (common on corporate/guest WiFi). Mobile clients broadcast a
+/// small JSON probe to this port; any listening desktop instance unicasts a
+/// response straight back to the sender's source address.
+///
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2253, opened 2026-07-21
fix(agent): vault icon on Agent-setup + eliminate horizontal scroll in the agent armory

AgentO-asaf:
## Summary
Two rounds so far:

1. **Vault icon** — the existing "Agent setup" icon changes from `id-card` to `vault`, plus icons + a container-query breakpoint on the modal's tab bar (matching the global Armory pane's icon choices and responsive technique). See §1-6 of the spec.

2. **Eliminate horizontal scroll** — live-tested by Asaf via `task dev`: scrolling appeared on every tab, not just the ones in round 1's review. Root cause + fix in full: `docs/specs/SPEC_AGENT_PANE_ARMORY_HEADER_ICON_2026_07_20.md` §7.
   - **Root cause (every tab):** the modal's own sizing used a raw `width: 780px; max-width: 92vw`. Since it opens pane-scoped, `92vw` resolves against the browser window, not the pane — on a narrow pane in a wide window it constrained nothing, so the modal always rendered at 780px and overflowed. Switched to `width: min(780px, 100%)` (the same fluid-width technique `modal.scss`'s own `data-size` modals already use) + a scoped `min-width: 0` rescue.
   - **Memories tab:** migrated off a fixed 220px list column onto `PrimitiveListDetail` — the same single-pane list/detail component MCP Servers and Skills already use. Also dropped the old auto-select-first-file behavior (existed only to avoid an empty second column that no longer exists; keeping it would make this the one tab that doesn't open to its list like its siblings).
   - **Accounts tab:** the provider-row grid had a non-shrinking `auto` actions column with a ~370-420px floor. Added a breakpoint that reflows to a 2-row stack instead of trying to compress content with nowhere left to go.
   - MCP Servers, Skills, Startup — no changes needed, already fully fluid.

## Test plan
- [x] `npm run lint:scss` — no new errors
- [x] `npx vitest run frontend/app/view/agent` — 673/674 passing (1 pre-existing unrelated timeout flake, confirmed passing in isolation)
- [ ] Live check — `task dev` is up (labeled "agentx: agent-armory no-scroll fix"); please confirm no horizontal scroll appears on any tab at narrow pane widths, and that Memories' list/detail toggle + Accounts' row-reflow both look right

### DIFF
--- a/.changesets/1784593673-feat-agent-open-armory-icon-in-the-pane-header-header-now-na-khou.md
+++ b/.changesets/1784593673-feat-agent-open-armory-icon-in-the-pane-header-header-now-na-khou.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): vault icon on the Agent-setup button, responsive tabs in the per-agent armory modal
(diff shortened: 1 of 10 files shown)


a5af:
@codex review

a5af:
@codex review

PR #2255, opened 2026-07-21
fix(auth): single-point login enforcement for oauth providers

GenericAgentX-asaf:
## Summary

Claude Code v2.1.x's OAuth flow can't open a browser from AgentMux's headless
CLI spawn. Before this PR, `/login`, the "Login Again" failure-banner action,
and the gated launch flow (behind "Retry Login") each dead-ended on this
independently — three separate, drifted implementations of "spawn login CLI,
wait for a URL." Live testing on a real stuck agent (repro details in the
first retro below) surfaced a second, deeper problem: the recovery path that
did work (copying a valid global `~/.claude` login) wrote into the shared
default credential directory instead of a real, Armory-tracked account — so
a fully working agent could exist with no visible account anywhere in
Armory, no expiry tracking, and (via the `use_ambient_login` escape hatch) no
enforcement that a real account ever existed.

- Consolidates `/login`, "Login Again", and the launch flow's auto-login onto
  one orchestrator (`runProviderLogin`) with a three-tier fallback:
  URL-capture → copy-from-a-valid-global-login → real terminal window,
  polled for the result.
- Ports `open_login_terminal` (the real-console fallback) to macOS/Linux —
  previously Windows-only, so it was the *only* working escape hatch and
  didn't exist on 2 of 3 platforms.
- Retires the `use_ambient_login` bypass in `identity/resolver.rs`'s spawn
  gate — an oauth-class provider now unconditionally requires a real bound
  `IdentityAccount`. No silent fallback to an untracked credential, for any
  provider.
- Reworks the seed-from-global recovery path to mint a real per-account
  isolated dir (`identity.ensureaccountdir`, wrapping the same
  `compute_and_ensure_account_dir` a real OAuth handshake already uses),
  seed it, and persist + link the account — instead of writing into the
  shared dir and faking "connected" UI state locally.
- Adds a proactive token refresh to the MuxBus cloud subscriber (found via
  the same investigation — a long-lived connection never re-checked its
  stored credential's expiry, so it could sit expired while still working).

**Behavior change reviewers should know about:** any agent currently relying
on `use_ambient_login=true` (pre-existing agents grandfathered by an earlier
migration) will fail its next turn until a real account is bound via "Use
existing login" — which now actually creates one instead of just seeding a
file.

Full investigation and design notes:
- `docs/retro/retro-headless-login-browser-open-2026-07-20.md`
- `docs/retro/retro-login-three-code-paths-2026-07-20.md`
- `docs/specs/PLAN_LOGIN_SINGLE_PATH_CONSOLIDATION_2026_07_20.md`

## Test plan

- [x] `cargo check --workspace` — clean, no errors
- [x] `cargo test -p agentmux-srv identity::resolver` — 28/28 pass (2 rewritten to pin the new invariant)
- [x] `cargo test -p agentmux-srv identity_handlers` — 16/16 pass
- [x] `npx vitest run frontend/app/view/agent frontend/app/store/launch-flow-state` — 733/733 pass
- [x] Live-verified end-to-end on a real stuck agent (empty shared-dir credential, valid global login): tier-1 timeout → tier-2 auto-recovery → real credential confirmed on disk, all via `muxlog` trace, before the account-registration work landed
- [ ] Live-verify the account-registration + enforcement changes together (rebuild + restart pending)

### DIFF
--- a/.changesets/1784593520-fix-auth-single-point-login-enforcement-every-oauth-class-ag-zdqv.md
+++ b/.changesets/1784593520-fix-auth-single-point-login-enforcement-every-oauth-class-ag-zdqv.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(auth): single-point login enforcement — every oauth-class agent requires a real bound Armory account
--- a/.changesets/1784697483-fix-auth-runproviderlogin-cancels-abandoned-tier-1-login-chi-l2p1.md
+++ b/.changesets/1784697483-fix-auth-runproviderlogin-cancels-abandoned-tier-1-login-chi-l2p1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): runProviderLogin cancels abandoned tier-1 login child, dedupes tier2/3 account minting; open_login_terminal fixes on macOS/Linux
--- a/.changesets/1784699000-fix-auth-non-claude-oauth-providers-were-permanently-unable--snou.md
+++ b/.changesets/1784699000-fix-auth-non-claude-oauth-providers-were-permanently-unable--snou.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): non-Claude oauth providers were permanently unable to log in via terminal fallback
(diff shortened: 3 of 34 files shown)


a5af:
@codex review

PR #2258, opened 2026-07-22
feat(browser-pane): individual Ctrl+Wheel zoom per pane via CSS injection

Agent1-asaf:
## Summary

Browser panes share their parent window's `RequestContext` (a deliberate tradeoff to avoid a `ThemeService` crash — `docs/specs/pane-shares-window-request-context-linux-2026-05-13.md`), so Ctrl+Wheel zoom (previously unhandled — `hwnd.rs`'s WndProc hook only logged `WM_MOUSEWHEEL`, letting Chromium's own native page zoom handle it by default) was scoped by a shared `HostZoomMap`: zooming one browser pane visually zoomed every other pane on the same host.

Full root-cause + alternatives investigation: `docs/specs/REPORT_ARMORY_ZOOM_AND_PER_PANE_BROWSER_ZOOM_2026_07_20.md`. Short version — a native CEF patch (`CefBrowserHost::SetZoomIsolated()`) was written and opened for review (`agentmuxai/cef#6`, `AgentU-asaf/cef-rs#2`, both closed unmerged) before finding this CSS-injection approach, which needs no native patch and no per-platform libcef build at all.

**Known limitation — Windows-only for now.** The Ctrl+Wheel *interception* point lives in `hwnd.rs`'s WndProc subclass hook, which is `#![cfg(target_os = "windows")]`-gated. No macOS/Linux equivalent hook exists yet in this codebase, so on those platforms Ctrl+Wheel over a browser pane still falls through to CEF's native, `HostZoomMap`-shared zoom (the original bug this PR fixes). The CSS-injection mechanism itself (`BrowserPaneManager::zoom_in`/`zoom_out`, `apply_zoom`) is plain, platform-agnostic Rust — the only missing piece on macOS/Linux is the input-interception hook. Tracked as a follow-up rather than implemented blind for platforms unavailable in this dev environment.

## Changes

- `hwnd.rs`: Ctrl+`WM_MOUSEWHEEL` is now intercepted and **consumed** (never reaches the default WndProc, so Chromium's native zoom never fires for it). Windows-only, disclosed above and in a code comment at the interception point.
- `browser_panes.rs`: `BrowserPaneManager::zoom_in`/`zoom_out` apply the new factor via `document.documentElement.style.zoom` through `execute_java_script` on the pane's own main frame — the same pattern `events.rs`'s `emit_event` already uses. Per-`CefFrame` by construction, so no `RequestContext`/`HostZoomMap` involvement and no cookie/session-sharing tradeoff either way.
- `state.rs`: new `AppState.browser_pane_zoom: Mutex<HashMap<String, f64>>`, keyed by `block_id`, clamped 0.5–2.0 matching `frontend/app/store/zoom.win32.ts`'s existing convention for every other pane type.
- `callbacks.rs`: `reapply_zoom` called from `on_load_end_browser_pane` so zoom survives navigation (a fresh page load replaces the previous DOM/inline-style state entirely).
- `zoom.darwin.ts` / `zoom.linux.ts`: brought in line with `zoom.win32.ts` — added `"armory"` to `getBlockZoom`'s view-type allow-list so Armory's own Ctrl+Wheel/keyboard zoom (a separate, already-merged feature, PR #2251) doesn't silently no-op on those platforms.

## Test plan

- [x] `cargo test -p agentmux-cef --lib browser_panes` — 7/7 passing (5 new: the pure clamp+step arithmetic, split out testable-without-a-real-browser the same way `BrowserPaneCloseOps` already is for the close path)
- [x] Live verification via `task dev` — confirmed Ctrl+Wheel reaches the handler, injected CSS renders correctly, and two panes on the same host zoom independently.


a5af:
@codex review

PR #2260, opened 2026-07-22
feat(auth): credential broker skeleton — single-flight refresh + keychain-backed MuxBus tokens

GenericAgentX-asaf:
## Summary

Phase A of a broader auth-architecture rethink documented in
`docs/specs/REPORT_AUTH_ARCHITECTURE_STATE_AND_RETHINK_2026_07_21.md` (included in this
PR) — AgentMux currently runs three independent OAuth/credential systems (provider-CLI
identity, MuxBus cloud login, an unused Armory service-account scaffold) with no shared
refresh or storage model. This PR introduces the first piece of the consolidated
**Credential Broker** and proves it out on MuxBus, the system closest to broker-shaped
already (it's the only one that does active token refresh today).

- **`agentmux-srv/src/broker/`** (new module) — `RefreshScheduler`: a generic,
  single-flight-guarded, proactively-scheduled credential refresher. Any number of
  credentials can register with one scheduler instance; it holds no credential state
  itself, only refresh coordination.
  - `ensure_fresh(id)` collapses concurrent callers for the same credential onto exactly
    one in-flight refresh via a per-id async mutex + double-checked freshness read. This
    closes the exact failure class Claude Code itself has two open production issues for
    (anthropics/claude-code#25609, #29896) — unlocked concurrent refresh attempts racing
    on a shared credentials file, which for providers that rotate refresh tokens can
    revoke the entire token family.
  - `run_sweep_loop` proactively walks every registered credential on a timer and
    refreshes anything stale, instead of only refreshing reactively when something
    happens to ask for a token.
- **MuxBus migrated onto it** (`muxbus/cloud_subscriber.rs`): previously had two
  independent, uncoordinated refresh call sites (the outer reconnect loop's
  `load_valid_token`, and an inline check on every 240s WS ping tick). Both now go
  through one `scheduler.register("muxbus:global", ...)` call; the broker's background
  sweep keeps the stored token fresh even while the WS session is disconnected/backing
  off, which the old ping-tick-only check could not.
- **MuxBus tokens moved off plaintext SQLite onto the OS keychain**
  (`backend/storage/muxbus.rs`): `access_token`/`refresh_token`/`id_token` now go through
  the existing `identity::secret_store` keychain wrapper (already used for
  `SecretRef::Keychain` elsewhere in this codebase) instead of living as plaintext TEXT
  columns in `db_muxbus_credentials`. Lazy-migrated on first read (no hard schema
  migration, no downtime) — a pre-migration row's plaintext columns are read once, written
  into the keychain, then blanked. Non-secret metadata (`cognito_domain`, `client_id`,
  `expires_at`, `user_email`, `user_sub`) stays in SQLite as before.

## Test plan

- [x] `cargo test -p agentmux-srv broker::` — single-flight dedup (10 concurrent callers
      → exactly 1 refresh), already-fresh no-ops, failed-refresh doesn't mark fresh,
      unregistered-id errors cleanly.
- [x] `cargo test -p agentmux-srv muxbus::` — token JSON round-trip, existing
      `cloud_subscriber` WS-handshake regression test still passes.
- [x] `cargo test -p agentmux-srv` (full suite) — 1586 passed, 0 failed, 0 regressions.
- [ ] Live end-to-end MuxBus reconnect/refresh cycle in a running instance — not
      performed (would require live cloud credentials in this session); relying on unit
      coverage of the single-flight/preserve-on-failure/proactive-sweep contracts instead.
- Not tested: real OS-keychain read/write I/O, matching this codebase's existing
  convention for `identity/secret_store.rs` (its own tests avoid real keychain I/O for
  CI-portability reasons — headless Linux runners may have no Secret Service daemon).

### DIFF
--- a/.changesets/1784593520-fix-auth-single-point-login-enforcement-every-oauth-class-ag-zdqv.md
+++ b/.changesets/1784593520-fix-auth-single-point-login-enforcement-every-oauth-class-ag-zdqv.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(auth): single-point login enforcement — every oauth-class agent requires a real bound Armory account
--- a/.changesets/1784593529-feat-armory-ctrl-wheel-zoom-reusing-the-per-block-term-zoom--dg05.md
+++ b/.changesets/1784593529-feat-armory-ctrl-wheel-zoom-reusing-the-per-block-term-zoom--dg05.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
 feat(armory): Ctrl+Wheel zoom, reusing the per-block term:zoom pipeline
--- a/.changesets/1784593673-feat-agent-open-armory-icon-in-the-pane-header-header-now-na-khou.md
+++ b/.changesets/1784593673-feat-agent-open-armory-icon-in-the-pane-header-header-now-na-khou.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): vault icon on the Agent-setup button, responsive tabs in the per-agent armory modal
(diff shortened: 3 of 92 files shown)


a5af:
@codex review

PR #2261, opened 2026-07-22
feat(term): in-pane tabs for terminal panes (Phase 5)

Agent3-asaf:
## Summary

Phase 5 of `docs/specs/SPEC_PANE_TAB_STRIP_AGENT_TERMINAL_2026_07_20.md`. Terminal panes get the same `<PaneTabStrip>` the editor and agent panes already have, reusing everything Phases 1–4 already built.

Confirms the spec's own prediction: **no new backend surface was needed at all.** `pane.open`'s `skip_placement` (added for Phase 4's agent-fork `+`) is already fully view-agnostic — it just creates a `Block` with whatever `meta`/`cwd` is passed and returns `{ block_id }`. This PR is pure frontend wiring.

Unlike the agent pane's fork strip (which derives its tab list from a cross-pane definition/lineage lookup, since a fork can exist as its own separate top-level pane before ever joining a stack), a terminal "tab" has no equivalent prior existence — there's no multi-session concept for shell panes anywhere else in the app (one block = one PTY = one `ShellController`, always). So the tab list is simply "whatever's in this pane's own `blockStack` right now", derived reactively off `layoutModel.localTreeStateAtom` (the same signal `getNodeModel`'s own `isFocused`/`isMagnified` memos already depend on).

When no stack exists yet (the common default case), the tab list synthesizes the pane's own one-tab entry rather than rendering empty, so the strip always shows the active terminal as a labeled tab next to `+`.

`+` spawns a new shell block (`skip_placement`, inheriting the current tab's `cwd`) and `pushBlockOntoStack`s it onto this pane. Closing the sole/last tab correctly falls through `closeBlockInStack` to the ordinary `closeNode` path (closes the whole pane) — no terminal-specific handling needed.

Tab labels are simple position-based ("Terminal 1", "Terminal 2", …) for now — a dormant tab's block isn't mounted (switching is a remount, not a reactive update), so a richer label is a follow-up.

## Test plan

- [x] `npx tsc -p tsconfig.json --noEmit` — clean (two pre-existing, unrelated errors already on `main`, confirmed via stash-and-check)
- [x] `npx vitest run` — 1863 passed, 3 pre-existing skips
- [x] No Rust changes in this PR
- [ ] Manual: click `+` on a terminal pane, confirm a second shell tab opens in the same pane inheriting the cwd, switch between tabs, close one

### DIFF
--- a/.changesets/1784695324-feat-term-in-pane-tabs-for-terminal-panes-phase-5-mbb5.md
+++ b/.changesets/1784695324-feat-term-in-pane-tabs-for-terminal-panes-phase-5-mbb5.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(term): in-pane tabs for terminal panes (Phase 5)
--- a/frontend/app/view/term/term.tsx
+++ b/frontend/app/view/term/term.tsx
@@ -17,10 +17,12 @@ import { makeTerminalModel, setTerminalViewComponent, TermViewModel } from "./te
 import { TermWrap } from "./termwrap";
 import "./xterm.css";
 import { DragOverlay } from "@/app/element/dragoverlay";
+import { PaneTabStrip } from "@/app/element/PaneTabStrip";
 import { detectHost, invokeCommand } from "@/app/platform/ipc";
 import { RpcApi } from "@/app/store/rpc-api";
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2262, opened 2026-07-22
fix(auth): New Agent modal Connect button — real terminal fallback for requiresLoginTty providers

GenericAgentX-asaf:
## Summary

Phase B of the auth-architecture rethink (report + Phase A in #2260). Follows up on
PR #2255's plan doc, which flagged (but deferred, pending a design question) that the
New Agent modal's "Connect" button is a structurally separate implementation with **no
terminal fallback at all** for providers whose CLI can't complete OAuth in a headless
spawn.

**One correction to the original framing, found by actually reading the render logic
rather than trusting the plan doc's prose:** `ConnectCta`'s `canSeed` branch
(`provider.id === "claude"`) already routes Claude to "Use my existing login" instead
of ever rendering the generic Connect button — so the dead-end `auth.start`/`auth.poll`
path is not currently *reachable* for Claude through this UI. It **is** live and
broken today for **OpenClaw** specifically (also `requiresLoginTty: true`, but no
`canSeed` bypass). This PR fixes the underlying `connect()` call for both, since it's
unconditionally broken for any `requiresLoginTty` provider regardless of which one the
UI currently happens to route there.

## Changes

- `PreLaunchAuthPanel.tsx`'s `startConnect()` routes `requiresLoginTty` providers
  through `runProviderLogin` (PR #2255's consolidated orchestrator) instead of
  `AuthFlowController.connect()`'s `auth.start`/`auth.poll` RPC path — giving them the
  same real terminal-window fallback (tier 3) and seed-from-global (tier 2, Claude)
  that pane-level logins already have.
- Extended `runProviderLogin` with three small, additive, backward-compatible params
  (all optional; existing 3 callers unaffected):
  - `existingAccountId` — reconnect into the same account instead of always minting a
    new one, needed for this modal's Reconnect case.
  - `onAccountRegistered` — lets a caller learn the resulting account id (to select it
    in a dropdown, etc.) without threading a new return shape through the shared
    function's existing 3 call sites.
  - `skipTier1` — skip the headless URL-capture attempt for providers where it's a
    documented, unconditional dead end, so "opened but no completion signal" (tier 1
    alone doesn't detect completion, unlike tiers 2/3) stays structurally unreachable
    for this caller rather than something it has to defend against.
- Widened `auth-state.ts`'s `Seeded` reducer case to also honor dispatch from
  `waiting` — the new call path dispatches `ConnectClicked` up front so the panel shows
  progress during tier 3's up-to-5-minute wait, then `Seeded` on success. The original
  single-phase "Use my existing login" caller never enters `waiting`, so this widening
  doesn't change its behavior (covered by an updated test).

## Scope note

Deliberately did **not** touch `controller.connect()`/`auth.start` for
non-`requiresLoginTty` providers (their piped/headless capture already works) or
remove `spawn_auth_cli`/`spawn_auth_cli_pty` from the backend — both still have a live
caller for those providers. The plan's Phase 4 cleanup items (dead-code removal, a
topology test pinning `auth.start` to zero frontend callers) don't apply until/unless
a later phase replaces that remaining path too.

## Test plan

- [x] `npx vitest run` (full suite) — 1877 passed, 3 skipped (pre-existing e2e,
      unrelated), 0 failed.
- [x] `npx tsc --noEmit` — no new type errors (one pre-existing, unrelated error in
      `swarm-model.test.ts`, confirmed present before this branch's changes too).
- [x] New/updated test coverage: `run-provider-login.test.ts` (+5 tests: `skipTier1`,
      `existingAccountId` threaded to both tier 2 and tier 3, `onAccountRegistered`
      firing on success and NOT firing when persist fails), `auth-state.test.ts`
      (`Seeded` honored from `waiting`, and still correctly dropped from
      `ready`/`saving`/`authenticated`/`idle`).
- [ ] Manual New Agent modal → OpenClaw → Connect in a running dev build — not
      performed this session; relying on the underlying `runProviderLogin` orchestrator
      already being exercised end-to-end by the pane-level flows it's shared with.
- No component-level test exists for `PreLaunchAuthPanel.tsx` itself (none existed
  before this change either) — `startConnect` is a module-private function; the new
  branching logic it adds is thin orchestration over `runProviderLogin` (tested) and
  the reducer (tested), so coverage is at those two layers rather than a new
  SolidJS-rendering test harness for the component.

Follow-up to #2260 (Phase A) and #2255 — addresses PR #2255's own plan doc's deferred
Phase 3 item for the New Agent modal's Connect flow specifically.

### DIFF
--- a/.changesets/1784593520-fix-auth-single-point-login-enforcement-every-oauth-class-ag-zdqv.md
+++ b/.changesets/1784593520-fix-auth-single-point-login-enforcement-every-oauth-class-ag-zdqv.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(auth): single-point login enforcement — every oauth-class agent requires a real bound Armory account
--- a/.changesets/1784593529-feat-armory-ctrl-wheel-zoom-reusing-the-per-block-term-zoom--dg05.md
+++ b/.changesets/1784593529-feat-armory-ctrl-wheel-zoom-reusing-the-per-block-term-zoom--dg05.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
 feat(armory): Ctrl+Wheel zoom, reusing the per-block term:zoom pipeline
--- a/.changesets/1784593673-feat-agent-open-armory-icon-in-the-pane-header-header-now-na-khou.md
+++ b/.changesets/1784593673-feat-agent-open-armory-icon-in-the-pane-header-header-now-na-khou.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): vault icon on the Agent-setup button, responsive tabs in the per-agent armory modal
--- a/.changesets/1784644126-fix-agents-stop-dropping-working-directory-in-agent-consolid-eaza.md
+++ b/.changesets/1784644126-fix-agents-stop-dropping-working-directory-in-agent-consolid-eaza.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agents): stop dropping working_directory in agent consolidation migration and surface Claude API errors/tool-results correctly in the translator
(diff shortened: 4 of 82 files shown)


a5af:
@codex review

PR #2263, opened 2026-07-22
fix(identity): gemini/copilot now get real OAuth isolation-dir + spawn-gate treatment

GenericAgentX-asaf:
## Summary

Phase C of the auth-architecture rethink (partial — see scope note below).
**Stacked on #2260 (Phase A)** — this PR's base branch includes Phase A's broker
module since gemini/copilot classifying as OAuth means they'd otherwise flow
through the same account-dir minting path Phase A touched; the diff here shows
only this PR's own changes.

`resolver::provider_class` is the single source of truth both the layer-3 spawn
gate **and** `identity_handlers.rs`'s per-account isolation-dir minting key off
(confirmed by reading both call sites directly, not assuming). It only classified
`claude`/`codex`/`openclaw` as OAuth-class, while the frontend's `ProviderDefinition`
table already marks `gemini` and `copilot` as `authType: "oauth"` too — so neither
mechanism actually applied to them despite the UI already presenting them as
oauth-class accounts. This was flagged in the auth-architecture report as unexamined
drift, not a deliberate choice.

- Extended `provider_class`'s match arm to include `gemini`/`copilot`.
- Added `oauth_class_matches_frontend_authtype_oauth_set`, a regression test pinning
  the backend OAuth-class set equal to the frontend's `authType: "oauth"` set going
  forward. No automated cross-language check exists in this codebase, so it's a
  manually-mirrored list — the test's doc comment tells future changes to update both
  together in the same change, not as a follow-up.
- Extended the existing `provider_class_oauth_providers` test with the two new
  providers' expected config-dir env vars (`GEMINI_CLI_HOME`, `COPILOT_HOME` —
  verified against `backend/providers.rs` directly).

## Scope note — one item from the plan dropped, with reasoning

The plan's other Phase C item was routing the OAuth expiry probe
(`inject_identity_env_with_broker`'s `probe_oauth_status` call) through the Phase A
broker. On closer look at the actual code, this doesn't apply yet: CLI-provider
credentials have **no active AgentMux-side refresh mechanism** today (unlike MuxBus,
which calls `pkce::refresh_token` itself) — the provider CLI manages its own token
refresh internally, and AgentMux only ever reads its token file read-only. Routing
that existing read-only probe through the broker without a real `refresh` behavior
behind it would be a hollow indirection — register a closure that does nothing but
re-run the same probe — not an actual improvement, and the broker's `HashMap` would
need per-account entry lifecycle handling that Phase A didn't design for. Revisiting
this makes more sense once there's a real CLI-provider refresh mechanism to broker
(plausibly alongside a future device-flow shim), not before.

## Test plan

- [x] `cargo test -p agentmux-srv identity::resolver::` — 29 passed, including the
      new regression test.
- [x] `cargo test -p agentmux-srv` (full suite) — 1587 passed, 0 failed, 4 ignored
      (pre-existing, unrelated).
- [ ] Manual: connect a `gemini` account via the New Agent modal and confirm it gets a
      real isolated account dir the same way `claude` does — not performed this
      session (would benefit from Phase B being available in the same build first,
      since that's what makes the New Agent modal's Connect flow work reliably for
      TTY-requiring providers; gemini itself doesn't require TTY, so this is
      lower-risk, but still worth a live check before merge).

Follow-up to #2260 (Phase A, base of this PR) — part of the same auth-architecture
rethink as #2262 (Phase B).

### DIFF
--- a/.changesets/1784696421-fix-identity-gemini-copilot-now-get-real-oauth-isolation-dir-2gnx.md
+++ b/.changesets/1784696421-fix-identity-gemini-copilot-now-get-real-oauth-isolation-dir-2gnx.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(identity): gemini/copilot now get real OAuth isolation-dir + spawn-gate treatment, closing a frontend/backend drift gap
--- a/.changesets/1784704633-docs-identity-fix-stale-claude-codex-openclaw-comment-in-per-atv1.md
+++ b/.changesets/1784704633-docs-identity-fix-stale-claude-codex-openclaw-comment-in-per-atv1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(identity): fix stale claude/codex/openclaw comment in per-account isolation-dir gate
--- a/.changesets/1784706352-docs-identity-fix-4th-stale-claude-codex-openclaw-comment-in-8wcg.md
+++ b/.changesets/1784706352-docs-identity-fix-4th-stale-claude-codex-openclaw-comment-in-8wcg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(identity): fix 4th stale claude/codex/openclaw comment in resolver test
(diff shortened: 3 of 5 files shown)


a5af:
@codex review

PR #2265, opened 2026-07-22
fix(agent): fork tab strip no longer leaks unrelated agents cloned from the same template

AgentY-asaf:
## Summary
- Fixes a bug where opening two completely unrelated agents (no fork relationship) in separate panes caused both panes' fork tab strips to show both agents as tabs.
- Root cause: `agentdefcreatefromtemplate` writes the template's id into `parent_id` for template-provenance (`db_agents.parent_template_id`, wire-mapped to `parent_id`). `computeForkSet` walks `parent_id` to find each pane's fork-lineage root — so two agents cloned from the same template shared that template as a "lineage root," and every other clone of the same template rendered as a fork.
- Fix: `hasParent()` in `fork-set.ts` now excludes parents that are seeded templates (`is_seeded === 1`), so only real `ForkAgentDefinitionCommand` lineages are walked. Genuine forks of a user-owned agent are unaffected.

## Test plan
- [x] New unit tests in `fork-set.test.ts`: two agents cloned from the same template no longer appear as each other's forks; a real fork of a user-owned agent still walks correctly.
- [x] `npx vitest run app/view/agent` — 55 files / 676 tests pass.
- [x] `npx tsc -p tsconfig.json --noEmit` — no new errors (2 pre-existing, unrelated failures: missing `qrcode` module, `swarm-model.test.ts` cron type mismatch).

### DIFF
--- a/.changesets/1784703901-fix-agent-fork-tab-strip-no-longer-leaks-unrelated-agents-cl-6d1v.md
+++ b/.changesets/1784703901-fix-agent-fork-tab-strip-no-longer-leaks-unrelated-agents-cl-6d1v.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): fork tab strip no longer leaks unrelated agents cloned from the same template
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

Issue #2155, opened 2026-07-14
Editor pane: rename-input caret/focus bug + CreateTab activate arg silently dropped


AgentO-asaf:
Two open items from the editor-pane work in PRs #2118, #2125, #2130 (all merged).

## 1. Rename input: no blinking caret, arrow keys / click don't reposition it (unresolved)

**Symptom:** In the file tree, click "Rename…" → filename shows highlighted/selected →
pressing Left/Right does not move or collapse the caret → clicking inside the text does not
reposition it either. Text stays highlighted regardless of click/keypress.

**Fix attempts so far (both failed in real-world testing):**

1. Added a `main_window_focus` IPC call (same pattern already used by the browser
   address bar for this class of bug) into `InlineInput`'s mount effect and its
   `onMouseDown`, in `frontend/app/view/editor/file-tree.tsx`.
2. Traced into `MainFocusReclaimTask::execute()` in `agentmux-cef/src/ui_tasks/window.rs`
   and found it was Windows-complete but macOS-incomplete (only `host.set_focus(1)`,
   which is Chromium-internal focus, not real AppKit focus — the latter is required for
   Blink to paint a caret). Added `makeKeyAndOrderFront:` + `makeFirstResponder:` ObjC
   calls ahead of the existing call. Compiled clean; confirmed via host log that the IPC
   path fires correctly on real user clicks. Still did not fix the symptom.

**Why this needs fresh eyes / a different diagnostic angle:** every layer observable
remotely (DOM `document.activeElement`, computed styles, CDP synthetic input including
`Input.dispatchKeyEvent`/`Input.dispatchMouseEvent`) looks correct in reproduction
attempts — the exact symptom has never been reproduced outside the user's real
interaction, which is why two evidence-backed fixes both missed.

**Unexplored angles, roughly in order of suspicion:**
- The NSView from `browser.host().window_handle()` may not be CEF's actual
  render-widget-host view that Blink needs as first responder (could be an outer
  container view — `makeFirstResponder:` would report success but focus the wrong node).
- Possible focus-stealing race immediately after `MainFocusReclaimTask` runs.
- May not be a native-focus issue at all — could be a SolidJS remount/stale-ref issue
  specific to real click+keyboard sequences that CDP structurally can't reproduce.

## 2. `WorkspaceService.CreateTab`'s `activate` argument is silently dropped

Found while implementing "Open in New Tab" (#2125). `CreateTab(oid, name, activate, ...)`'s
[...]

PR #2268, opened 2026-07-22
fix: batch of small triaged issue fixes (#1397, #780, #859, #2155)

### LINKED ISSUE #780: Default agent name on launch — let users click Launch immediately
## Summary

The agent-name field is empty on launch form open. User must type a name to click Launch. Friction with no payoff — most users want to just spin up an agent.

## Proposed

Pre-populate with `<Provider> Agent` (e.g., `Claude Agent`). If that name's taken, suffix `2`, `3`, ... — `Claude Agent`, `Claude Agent 2`, `Claude Agent 3`. User can still edit; default just removes the type-something-to-proceed friction.

## Spec

`docs/specs/SPEC_DEFAULT_AGENT_NAME.md` on branch `agenta/spec-default-agent-name`.

Covers: algorithm, where to compute, dirty-flag handling on provider change, validation, effort (~55 LOC, ~0.5 day).

## Related

- #779 — zombie HWND eats keystrokes; this spec is a partial mitigation (no typing needed for the happy path).

### LINKED ISSUE #859: Settings types drift: `gotypes.d.ts` declares keys absent from `schema/settings.json` (waveterm vestiges)
## Summary

`frontend/types/gotypes.d.ts` declares a number of settings keys (mostly under the `ai:`, `editor:`, and `autoupdate:` namespaces) that no longer exist in `schema/settings.json` or `agentmux-srv/src/backend/wconfig/types.rs`, and that I couldn't find consumed by current code outside of test fixtures.

Per maintainer confirmation, these are **vestiges of the old waveterm backend** that the Rust rewrite no longer uses. The typings file (`gotypes.d.ts` — name suggests Go provenance) was never pruned, so the type system still advertises them as live settings.

Discovered while triangulating the source against `agentmuxai/agentmux-docs` — the docs site enumerates ~25 of these keys in `settings.md`, which gives users the impression they're functional levers. (Docs cleanup will happen separately in agentmux-docs.)

## Concrete examples

Declared in `frontend/types/gotypes.d.ts` but **not** in `schema/settings.json`:

| Key | gotypes.d.ts | schema | Used anywhere outside types? |
|---|---|---|---|
| `ai:preset` | ✅ | ❌ | no |
| `ai:apitype` | ✅ | ❌ | no |
| `ai:baseurl` | ✅ | ❌ | no |
| `ai:apitoken` | ✅ | ❌ | no |
| `ai:name` | ✅ | ❌ | no |
| `ai:model` | ✅ | ❌ | only `agentmux-srv/src/backend/wconfig/mod.rs` test fixture |
| `ai:orgid` | ✅ | ❌ | no |
| `ai:apiversion` | ✅ | ❌ | no |
| `ai:maxtokens` | ✅ | ❌ | no |
| `ai:timeoutms` | ✅ | ❌ | no |
| `ai:backendtype` | ✅ | ❌ | no |
| `ai:local` | ✅ | ❌ | no |
| `editor:fontsize` | ✅ | ❌ | no |
| `autoupdate:enabled` | ✅ | ❌ | only test fixture (declared **twice** in gotypes.d.ts too) |

The double declaration of `autoupdate:enabled` in `gotypes.d.ts` is another small smell — the file is likely hand-maintained or generated from a stale source.

## Why this matters

1. **Typed-but-unused keys are footguns.** Frontend code can write `cfg["ai:model"] = "..."` and the type system approves, but nothing in the backend will read it.
2. **Docs followed the types.** The public docs site documents these keys as real settings; users who copy them get no effect.
3. **Gotypes.d.ts appears orphaned by the Rust rewrite.** It still describes the waveterm-era Go backend shape.

## Proposed resolution

Since maintainers have confirmed these are waveterm vestiges:

1. **Delete the vestigial entries from `frontend/types/gotypes.d.ts`** — `ai:*`, `editor:*` (except any that are actually in the Rust schema), `autoupdate:*`, plus the duplicate `autoupdate:enabled` line.
2. **Sweep any TS code that still imports these keys** (likely none — they're typed-only) so the deletion doesn't break compile.
3. **Treat `gotypes.d.ts` as suspect long-term.** If it was once auto-generated, the generator is gone; either rename to reflect that it's now hand-maintained, or regenerate from `schema/settings.json` / `wconfig/types.rs`.

The docs site will get a parallel cleanup PR removing the corresponding rows from `settings.md`.

## Verification commands

```bash
# Keys in canonical schema
grep -oE '"(window|term|app|ai|agent|editor|autoupdate|markdown|mcp):[a-z]+"' schema/settings.json | sort -u

# Keys declared in gotypes.d.ts
grep -oE '"(window|term|app|ai|agent|editor|autoupdate|markdown|mcp):[a-z]+"' frontend/types/gotypes.d.ts | sort -u

# Diff:
diff <(grep -oE '"(ai|editor|autoupdate):[a-z]+"' schema/settings.json | sort -u) \
     <(grep -oE '"(ai|editor|autoupdate):[a-z]+"' frontend/types/gotypes.d.ts | sort -u)
```

## Context

Surfaced from a docs triangulation audit run on `main@29105883` (2026-05-15). Full report lives in agent1's workspace (`AGENTMUX_DOCS_TRIANGULATION_2026-05-15.md`); not committed to either repo.

### LINKED ISSUE #1397: Unified per-agent conversation log for cross-block continuation (follow-up to #1361)
## Background

PR #1361 replaced the schema-v1 agent snapshot (which embedded the full composite `nodes[]`) with a lightweight schema-v2 overlay + NDJSON restore, eliminating a renderer OOM crash (`writeSnapshotNow` serializing ~18 MB of nodes every 30 s — see `docs/analysis/ANALYSIS_RENDERER_CRASH_WRITE_STATE_OOM_2026_06_12.md`).

To stay safe, #1361 **scoped v2 restore to same-block reopen** (the case the OOM actually occurs in). Cross-block "structural continuation" (Option E / PR #1007 — open a *new* pane for an existing agent and pick up its conversation) now falls back to NDJSON replay on the new block instead of restoring the prior history.

## Why cross-block is hard

- The durable NDJSON conversation log is stored **per-block** (`<blockId>/output`, written by `handle_append_block_file` from the subprocess stdout reader).
- The v2 snapshot is **agent-anchored** (`agent:<defId>:current`, shared across all of an agent's blocks).
- A continuation pane gets a fresh `blockId` whose per-block output is empty, so a single `(sourceBlockId, highWaterMark)` cannot describe a conversation physically split across blocks. v1 only worked because it embedded the whole composite document.

## Proposed fix: unified per-agent log

1. **Backend** — mirror each agent output line into the agent-anchored zone `agent:<defId>:current/output` via the existing `append_session_output` (currently defined but **never called** by the live path). Write site: `agentmux-srv/src/backend/blockcontroller/subprocess.rs:658-671` (and the parallel `persistent.rs:428`, `acp.rs:405`). The `definition_id` is resolvable once per turn from block meta `agentId` via the captured `wstore`.
2. **Frontend** — have v2 `writeSnapshotNow` count `highWaterMark` and v2 restore read history by `definition_id` (the agent zone), not per-block. `sourceBlockId` then becomes unnecessary.

## Must handle

- **Concurrency**: multiple live panes of the *same* agent definition would interleave appends into one zone → corruption. Need single-writer arbitration or per-pane sub-streams.
- **Archive**: `archive_session` clears `agent:<defId>:current/output`; confirm the mirror's "current session" scoping matches intent.
- **Hot-path cost**: a second `append_data` per output line (acceptable, but measure).
- **Migration**: existing per-block logs won't be in the agent zone; decide whether to backfill or only apply going forward.

## References
- Spec: `docs/specs/SPEC_WRITE_STATE_NDJSON_RESTORE_2026_06_12.md` §15
- Scoped fix: PR #1361

GenericAgentX-asaf:
## Summary

Four independently-scoped fixes from a GitHub issue triage pass, bundled since each is small and had an exact, evidence-backed location before implementation started. Closes #1397, closes #780, closes #859. Addresses only the "CreateTab activate arg" half of #2155 (leaves it open — the rename-input caret/focus bug is a separate, deeper issue).

**#1397 — cross-block conversation continuation restore**
`useHistoryPagination`'s v2 snapshot fast-path only restored when the snapshot's `sourceBlockId` matched (or was agent-anchored/empty); a genuine cross-block continuation (a NEW block opened for the same agent definition) fell through to legacy NDJSON replay on the new block's own empty output file, rendering blank. Fix: extend the fast path to a real cross-block match whenever `opts.definitionId` is available — `BlockfileLineCountCommand`/`BlockfileReadRangeCommand` already resolve to the agent's GLOBAL output zone by the block's own `agentId` meta, independent of the snapshot's recorded `sourceBlockId`. Only the no-definitionId legacy/picker path still needs the NDJSON fallback.

**#780 — default agent name on launch**
Pre-fills `<Provider> Agent` (e.g. "Claude Agent"), suffixed `2`, `3`, ... on collision — only in "New" mode, only when nothing already seeded the field. New `default-agent-name.ts` module.

**#859 — settings type drift (gotypes.d.ts vs schema/settings.json)**
Diffed `SettingsType` against the JSON schema exactly and found drift in both directions:
- 11 real, actively-used settings (`dnd:*`, `voice:*`, `notify:sound:agent.waiting.for.input`, `notify:sounds:waiting:volume`) existed in the TS type and are read by real code, but were missing from `schema/settings.json` — since that schema sets `additionalProperties: false`, any user with one of these in their real settings.json hits a validation error in the raw-editor escape hatch. Added to the schema.
- `term:zoom`/`tab:color` were misplaced in `SettingsType` (they're per-block/per-tab `.meta` keys, already correctly modeled in `MetaType`/`ConnKeywords`); `dnd:*`/`dnd:maxfilesizemb` were genuinely dead. Removed. `tab:color`/`tab:color-initialized` added to `MetaType`, where the tab-color code actually reads them (via a workaround `as MetaType` cast today).
- `conn:askbeforewshinstall`/`conn:wshenabled`/`tab:skipcloseconfirm` existed in the schema but not the TS type — added.

Net: 0 remaining drift (verified by re-running the diff after the fix).

**#2155 (CreateTab-activate half)** — the "activate arg silently dropped" comment in `editor-model.ts` was accurate when written but is stale: the backend's CreateTab handler already dispatches a compensating `SetActiveTab`. Corrected the comment.

## Test plan
- [x] `npx tsc --noEmit -p .` — clean (also fixed a pre-existing unrelated type error in `swarm-model.test.ts`'s `mkCron` helper)
- [x] `npm run build` — clean
- [x] `npx vitest run` — 122/124 files, 1875/1878 tests passing (2 pre-existing skips)
- [ ] Manual: open a new "New agent" launch modal, confirm the name pre-fills and Launch is clickable immediately; open the raw settings.json editor and confirm no schema-validation errors on `voice:*`/`dnd:*` keys


a5af:
@codex review

PR #2267, opened 2026-07-22
chore: delete confirmed-dead Rust + frontend code (audit §2.1/§5.1)

AgentX-asaf:
## Summary

Implements the mechanical dead-code deletion pass from `docs/reports/REPORT_REPO_HEALTH_AUDIT_2026_07_20.md` (§2.1 Rust, §5.1 frontend) — the remaining scope from the audit after PR #2257 fixed the 5 real bugs it surfaced. Every item was individually re-verified as zero-caller/zero-importer immediately before deletion (2 days had passed since the audit; other PRs had landed in the meantime), not just deleted on the audit's say-so.

**Rust (~20 items):**
- Superseded-by-`_unique`/newer-helper: `mcp_server_upsert`, `skill_upsert`, `get_stdin_tx`, `resolve_provider_alias`
- Zero-caller: `send_message` (ACP), `send_to_conn`, `is_armed`, `pending_count`, `peek_back_pending_window_creation`/`has_browser`/`list_browser_labels`
- Never-constructed/never-read: `GitHubContext` struct, `OscEvent.ps`, `ToolSpec.bundled`, `WebhookMessage.msg_type`, `AgentMuxHandler.ipc_port` (threaded out through its constructors + all 5 call sites), `oauth_status::UNKNOWN`
- Windows-dead: the `wndproc.rs` frameless-resize cluster (`setup_native_frameless`/`ORIGINAL_WNDPROCS`/`install_frameless_resize_hook`) and the Windows `post_start_drag` no-op stub — both already flagged in the 07-05 audit, still present 15+ days later

Deliberately **not** deleted: `into_bundle_id` field on `auth_session.rs`'s `Session` (entangled with a 10+ call-site public API for marginal value) and the two migration no-op bodies (kept inert for migration-id stability, per the audit's own "safe to leave inert" phrasing).

**Frontend:**
- Entire `frontend/app/view/agent-def/` directory (12 files) + the stray `"agent-def"` `VIEW_LABELS` entry
- `FilterControls.tsx`, `NewAgentCard.tsx`
- `identity-view.tsx`'s dead `IdentityView`/`IdentityPanel`/`AssignmentsTab` exports (the file's still-live `AccountsTab`/`AccountForm`/`AccountDetail` are untouched)
- `.agent-document-node-wrapper` SCSS class, `icon` field on `BlockKindMeta`, `detectAgentFromPath()`

## Test plan

- [x] `cargo build --workspace` — 0 errors
[...]

PR #2271, opened 2026-07-22
fix(identity): restore resolve_provider_alias, broken by #2267's dead-code sweep

Agent2-asaf:
## Summary
`main` is currently broken for `cargo check --workspace --tests` (confirmed via CI on an unrelated PR — both `windows-latest` and `ubuntu-latest` fail with `error[E0425]: cannot find function resolve_provider_alias`).

Root cause: #2267 ("delete confirmed-dead Rust + frontend code") deleted `resolve_provider_alias` from `backend/providers.rs`, believing it had zero callers per its audit re-verification pass. But #2263 (merged immediately before it) had just added a real caller in `identity/resolver.rs:324`'s `provider_class()` — a P1 fix ensuring `provider_class` and `get_provider` can't disagree on an alias ID. The "zero-caller" re-check missed this call site.

Fix: restored the function verbatim from #2267's own diff. Nothing else changed — the same commit's unrelated deletion of `resolver.rs`'s genuinely-unused `oauth_status::UNKNOWN` const is correct and left alone.

## Test plan
- [x] `cargo check -p agentmux-srv --tests` — now succeeds (was failing before this fix, reproduced locally).
- [x] `cargo test -p agentmux-srv --bin agentmux-srv identity::resolver` — 30 passed, 0 failed, including `provider_class_resolves_aliases_to_the_same_result_as_canonical`.

## Priority
This blocks CI for every open PR in the repo, not just this fix. Requesting a fast merge.

### DIFF
--- a/.changesets/1784743591-fix-identity-restore-resolve-provider-alias-deleted-by-2267--fo6w.md
+++ b/.changesets/1784743591-fix-identity-restore-resolve-provider-alias-deleted-by-2267--fo6w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): restore resolve_provider_alias deleted by #2267's dead-code sweep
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #2272, opened 2026-07-22
fix(auth): open_login_terminal no longer nulls stdin on Windows

GenericAgentX-asaf:
## Summary

`open_login_terminal`'s Windows path (`CREATE_NEW_CONSOLE` + `cmd.exe /k`) explicitly set `.stdin(Stdio::null())`. Live-tested this session: the exact same `claude.cmd auth login` command, with the same isolated `CLAUDE_CONFIG_DIR` AgentMux sets, opened a real browser and completed OAuth correctly when run interactively (PowerShell) — but produced no browser at all when spawned via AgentMux's actual terminal-fallback path, whose only difference was the NUL'd stdin.

Rust's `Command` only sets `STARTF_USESTDHANDLES` (overriding whatever handles `CREATE_NEW_CONSOLE` would otherwise wire up for the new console) when a stdio method is explicitly called. Removing the explicit `.stdin()` call lets `CreateProcess` give the child fresh handles tied to its own new console — required both for the CLI's own TTY detection (gates whether it even attempts the browser-open call) and for the "Paste code here if prompted >" manual fallback the same command prints, which a NUL'd stdin made impossible to use regardless.

macOS/Linux are unaffected — neither path touches stdio explicitly; both already spawn via a real terminal app/emulator.

## Test plan

- [x] Live-tested: ran the exact command manually via PowerShell with the same isolated `CLAUDE_CONFIG_DIR` — browser opened, OAuth completed successfully (confirmed by the user completing login twice during diagnosis).
- [x] `cargo check -p agentmux-cef` clean.
- [x] Built a fresh portable with the fix; awaiting live in-app confirmation.

### DIFF
--- a/.changesets/1784760432-fix-auth-open-login-terminal-no-longer-nulls-stdin-on-window-yjrg.md
+++ b/.changesets/1784760432-fix-auth-open-login-terminal-no-longer-nulls-stdin-on-window-yjrg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): open_login_terminal no longer nulls stdin on Windows, restoring browser-open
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #2273, opened 2026-07-22
feat(process-broker): unify agent process/status tracking behind one broker (Phase A)

AgentY-asaf:
## Summary
Implements Phase A of `docs/specs/REPORT_PROCESS_ARCHITECTURE_STATE_AND_RETHINK_2026_07_22.md` (included in this PR). Reported symptom: agent process/liveness data showed up inconsistently between the Agent pane and the Swarm overview. Traced to **six independent, only-partially-overlapping tracking mechanisms** with no single source of truth (`blockcontroller::CONTROLLER_REGISTRY`, `process_tracker`, the `reactive` handler's own registration list, `pidregistry`, `HealthMonitor`'s output-silence heuristic, `watchdog.rs`'s PTY-idle timers).

## What this does (wrap-first, matching the Credential Broker's own sequencing)
- New `agentmux-srv/src/broker/process.rs` — `ProcessBroker`, a sibling module to the existing Credential Broker (`broker::scheduler`) — same "one broker mediates a fragmented concern" pattern applied a second time, per the report's §6.6.
- `lifecycle_from()` is a pure reducer over `BlockControllerRuntimeStatus`, mirroring `reducer/lifecycle.rs`'s shape (explicit per-case transition rules, not a bare field copy) — report §3.1.
- `ProcessBroker::list()` replaces `agent.tracked-blocks`'s `.chain()` of `process_tracker`'s registry + the `reactive` handler's own registration list (two structurally different registries unioned with **no reconciliation** — a provider-chain anti-pattern) with `blockcontroller::get_all_controllers()`, which is already authoritative for every controller type and closes the coverage gap `process_tracker` alone has for `shell`/`acp` blocks.
- `ProcessBroker::status()`/`forget()` single-flight-guard reads through a `parking_lot::Mutex`-backed cache (sync, not async — `compute_status()`'s two reads are cheap in-memory lookups, not I/O, matching `process_tracker::registry`'s own choice of a sync lock for the same reason), and emit a `processbroker:status-changed` WPS event on real transitions.
- Wired into `AppState`, `main.rs` construction, and `blockcontroller::delete_controller` (cache cleanup on pane close).

**`agent.tracked-blocks`'s RPC response shape is unchanged** — this is a backend-only fix, no frontend/wire-type changes needed.

## Deferred to later phases (report §6)
- Migrating each `Controller` impl to register with the broker directly at spawn (closing the coverage gap at the write side, not just the read side).
- Consolidating the two independent activity-summary pipelines (push-sweep `activity_watcher.rs` vs. pull-RPC `session.rs`).
- Moving Swarm's client-side-only fine-grained activity chip server-side.
- A real batch/aggregate status RPC + Swarm frontend migration off its N sequential `GetControllerStatus` calls (this PR fixes the block-*discovery* list; the *status* fan-out is a separate, larger change touching the frontend and wire types).

## Verification
- `cargo build -p agentmux-srv` — clean.
- `cargo test -p agentmux-srv` — full suite green, **1610 passed**, 10 new (`broker::process::tests`).
- New tests cover: every `lifecycle_from()` transition rule, unknown-block handling, concurrent-caller consistency (10 real OS threads hitting the cache), cache cleanup, and that `list()`'s block_ids exactly track `get_all_controllers()`'s current keys (written to be safe under the shared-global-registry test parallelism this crate's test binary already has).

### DIFF
--- a/.changesets/1784760543-feat-process-broker-unify-agent-process-status-tracking-behi-a33g.md
+++ b/.changesets/1784760543-feat-process-broker-unify-agent-process-status-tracking-behi-a33g.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(process-broker): unify agent process/status tracking behind one broker (Phase A)
--- a/agentmux-srv/src/backend/blockcontroller/mod.rs
+++ b/agentmux-srv/src/backend/blockcontroller/mod.rs
@@ -247,6 +247,12 @@ pub fn delete_controller(block_id: &str) {
     if let Some(registry) = crate::backend::process_tracker::registry::global() {
         registry.remove(block_id);
     }
+    // Drop the Process Broker's cached status for this block, so a closed
+    // pane's stale entry doesn't linger unreachable (get_all_controllers()
+    // already won't list it — see ProcessBroker::forget's doc comment).
+    if let Some(broker) = crate::broker::process::global() {
+        broker.forget(block_id);
+    }
 }
 
 /// Get all controllers (snapshot).
--- a/agentmux-srv/src/broker/mod.rs
+++ b/agentmux-srv/src/broker/mod.rs
@@ -1,7 +1,15 @@
 // Copyright 2026, AgentMux Corp.
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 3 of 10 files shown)


a5af:
@codex review

PR #2275, opened 2026-07-22
refactor(broker): reducer-based state machine for credential refresh coordination

GenericAgentX-asaf:
## Summary

Refactors the Credential Broker (`agentmux-srv/src/broker/`, shipped as PR #2260) from implicit state (a bare `is_fresh()` boolean + an unqueryable per-credential `AsyncMutex<()>`) to an explicit reducer-based state machine, matching AgentMux's existing 4-tier reducer convention (`agentmux-launcher/src/reducer/`, `agentmux-srv/src/reducer.rs`, `agentmux-cef/src/reducer/mod.rs`, and the frontend reducer spec).

- **New:** `agentmux-srv/src/broker/state.rs` — pure `update(&mut HashMap<String, CredentialState>, Command) -> Vec<Event>` reducer. No I/O, no async, sub-millisecond, easily unit-testable in isolation.
- **`CredentialState`:** `Fresh` / `Refreshing { prior_failures }` / `Failed { consecutive_failures, last_error }` / `NeedsReauth { since_unix, reason }`. Escalates to `NeedsReauth` after 5 consecutive transient failures, or immediately on a classified-permanent error — closing a real gap where the old scheduler retried a dead refresh_token forever with no distinct "give up, a human needs to log in again" state.
- **`agentmux-srv/src/broker/scheduler.rs`** rewritten as a thin async orchestrator: dispatches commands to the reducer, executes whatever effects the returned events call for (`is_fresh()`/`refresh()`), dispatches the result back in, wakes waiters via a per-id `Notify`. Public API (`register`/`unregister`/`ensure_fresh`/`run_sweep_loop`) unchanged — no caller changes needed beyond the error-type update below.
- **`agentmux-srv/src/muxbus/pkce.rs`:** `refresh_token` now returns a typed `RefreshTokenError` (`NoRefreshToken` / `Network` / `Rejected{status,body}` / `ParseFailed`) instead of a bare `String`, so callers can classify failures instead of string-sniffing.
- **`agentmux-srv/src/muxbus/cloud_subscriber.rs`:** classifies `RefreshTokenError` into the broker's `RefreshErrorKind` (4xx rejection or no-refresh-token → `PermanentAuthFailure`; everything else → `Transient`).
- Every transition logs under `auth.broker.*`, which lands directly in `muxlog auth`'s existing filter regex with zero changes needed there — this was a real diagnostic gap surfaced while debugging the login-terminal stdin bug (#2272) this session.

## Test plan

- [x] `cargo test -p agentmux-srv broker::` — 20/20 passing (reducer + orchestrator, including a register-while-refreshing race test the old design had no coverage for)
- [x] `cargo test -p agentmux-srv cloud_subscriber::` — new `classify_refresh_token_error` unit tests (permanent vs. transient classification)
- [x] `cargo test -p agentmux-srv` (full workspace) — 1620/1620 passing, 0 regressions
- [ ] Manual: force a MuxBus refresh failure and confirm `muxlog auth` shows `Failed` accumulating and `NeedsReauth` once escalated, and that the sweep loop stops retrying at that point

### DIFF
--- a/.changesets/1784761584-refactor-broker-reducer-based-state-machine-for-credential-r-rp7z.md
+++ b/.changesets/1784761584-refactor-broker-reducer-based-state-machine-for-credential-r-rp7z.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(broker): reducer-based state machine for credential refresh coordination
--- a/agentmux-srv/src/broker/mod.rs
+++ b/agentmux-srv/src/broker/mod.rs
@@ -19,8 +19,10 @@
 //! (Phase C+), not touched here.
 
 pub mod scheduler;
+pub mod state;
 
 pub use scheduler::RefreshScheduler;
+pub use state::{CredentialState, RefreshErrorKind};
 
 use std::sync::{Arc, OnceLock};
 
--- a/agentmux-srv/src/broker/scheduler.rs
+++ b/agentmux-srv/src/broker/scheduler.rs
@@ -140,6 +140,19 @@ impl RefreshScheduler {
     pub async fn ensure_fresh(&self, credential_id: &str) -> Result<(), String> {
         let id = credential_id.to_string();
         loop {
+            // Create the Notified future *before* dispatching CheckRequested
+            // (and thus before releasing the states lock that gates whether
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

PR #2276, opened 2026-07-23
fix(launcher): macOS splash rows now animate instead of snapping

AgentO-asaf:
## Summary
- Fixes the reported "splash rows don't count up" bug: rows whose Begin+End both land in the same ~8-24ms render tick were applied back-to-back and created already-done, so no "running" frame ever painted — fast steps just snapped to their final value.
- Fix defers an End to the next tick whenever its matching Begin was also seen in the same drain pass, guaranteeing every row spends at least one tick visibly "running" before flipping done.
- Adds `docs/specs/PLAN_SPLASH_TELEMETRY_OPEN_ITEMS_2026_07_22.md`, consolidating the splash-telemetry spec's scattered open items (7 found vs. the spec's own tally of 3) into one tracker, and resolves two of them for free via verification (no code risk): confirmed `cef`/`frontend` stage events were never wired end-to-end, and confirmed Linux's actual shipped hold-duration default is 450ms (not the 1500ms the original spec assumed) — and that it's a semantically different knob (min on-screen time vs. macOS's post-ready hold) despite sharing an env var name.
- Adds `docs/retro/retro-splash-screen-rows-dont-count-up-2026-07-22.md` explaining why PR #2244 never fixed this (it closed a different, unrelated complaint about the "total" row not matching the sum of item rows).

## Test plan
- [x] `cargo test -p agentmux-launcher splash_mac` — 11/11 passing (6 pre-existing + 5 new covering the deferral logic)
- [x] `cargo test -p agentmux-launcher` (full crate) — 222/222 passing
- [x] `cargo check` / `cargo clippy` clean on touched code
- [ ] Live visual verification on a real macOS launch (can't screen-record/verify AppKit rendering from this environment — Asaf to confirm rows visibly count up)

### DIFF
--- a/.changesets/1784781999-fix-launcher-macos-splash-rows-now-animate-instead-of-snappi-ygwe.md
+++ b/.changesets/1784781999-fix-launcher-macos-splash-rows-now-animate-instead-of-snappi-ygwe.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(launcher): macOS splash rows now animate instead of snapping to their final value
--- a/agentmux-launcher/src/splash_mac.rs
+++ b/agentmux-launcher/src/splash_mac.rs
@@ -138,6 +138,15 @@ struct StageRow {
     subs: Vec<SubRow>,
 }
 
+/// Identifies which row a `Begin` event created, so `run_until_dismissed`'s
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2277, opened 2026-07-23
Fix stale Claude Code version doc and exe-return-codes.md

Agent2-asaf:
## Summary
- Doc-analyst findings DOC-003, DOC-007 from the Weekly Documentation Analyst's first report.
- `spec-claude-code-versioning.md` said current pin was 2.1.197 across "two locations"; actual current pin is 2.1.198 across five locations, and the Dockerfile ARG (not covered by `pin-consistency.test.ts`) had silently drifted behind the rest. Bumped it to match and documented all five pin sites.
- `exe-return-codes.md` referenced renamed/retired binaries (`agentmuxsrv-rs.exe` -> `agentmux-srv`, `wsh.exe` retired) and an NSIS installer table; Windows packaging moved to Inno Setup (`packaging/windows/agentmux.iss`). Updated names and pointed at Inno Setup's own exit-code reference instead of fabricating a replacement table.

## Test plan
- [x] Verified pin values via `pin-consistency.test.ts`, `providers.rs` (srv + cef), `frontend/.../index.ts`, and `container-image.yml`
- [x] Verified binary names via `agentmux-cef/src/sidecar.rs`
- [x] Verified installer via `packaging/windows/agentmux.iss` / `scripts/package-installer.ps1` and repo-wide grep for NSIS vs. Inno Setup references
- [ ] No app code paths changed besides a Docker build-arg default; no build/test run needed beyond the above

### DIFF
--- a/docker/Dockerfile.agent-agentmux
+++ b/docker/Dockerfile.agent-agentmux
@@ -33,7 +33,7 @@ RUN usermod -l agent -d /home/agent -m node \
 
 # Claude Code — version pinned at build time for reproducible images.
 # Managed by rebuilding with a new CLAUDE_VERSION arg, not npm update.
-ARG CLAUDE_VERSION=2.1.197
+ARG CLAUDE_VERSION=2.1.198
 RUN npm install -g @anthropic-ai/claude-code@${CLAUDE_VERSION}
 
 # Prevent in-container auto-updates — version is managed at the image level.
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

Agent2-asaf:
Fixed in d99e9740 — added back an agentmux.exe/launcher section, verified against agentmux-launcher/src/main.rs's actual exit sites. Kept the codes I could verify directly (0/1 for supervisor outcome and fatal startup errors, 2 for \`--diag\` usage errors) rather than asserting single-instance/IPC-specific codes I couldn't confirm exist. Resolved the thread.

Agent2-asaf:
Fixed in f5c99555 — confirmed second_instance.rs's bind_socket_with_recovery() also process::exit(2)s on macOS/Linux for IPC socket setup failures (bind, lockfile open, flock, post-lock/post-recovery rebind), separate from Windows' --diag usage-error case. Documented both under exit code 2.

Agent2-asaf:
Fixed in 42d1f209 — confirmed \`supervisor/windows.rs:213,226\` are IPC named-pipe bind/forward failures (same class as macOS/Linux's Unix-socket case, added to the exit-2 row), and that the table wasn't exhaustive: added \`TEARDOWN_BACKSTOP_EXIT_CODE = 86\` (wedged-host teardown) and an explicit "other" row documenting that \`break code\` at 7 sites passes through the host/srv's own raw exit code once a restart or OOM-recovery budget is exhausted.

PR #2278, opened 2026-07-23
feat(agent): drop the shell's activity-log panel, write output into the terminal instead

AgentO-asaf:
## Summary
- Clicking "Shell" in the agent pane no longer shows a separate activity-log panel above the terminal — just the terminal, filling the drawer.
- Bang-command (`!cmd`) output and other status/error messages (auth prompts, launch-flow status, ~15 call sites) now write directly into the terminal's local render buffer instead of the removed panel, via a new `onTermReady`/`onTermDispose` prop pair on `AgentShellSubblock`. `Terminal.write()` is local-only (never touches the PTY/stdin), matching the existing pattern in `AgentInstallModal.tsx`.
- `useActivityLog`'s backlog is kept and replayed into the terminal each time it (re)mounts, so reopening the drawer after closing it doesn't lose history.
- `ActivityLogPanel.tsx` deleted (fully unused now). Trimmed dead log-specific CSS out of `_activity-log.scss`/`_composer-strip.scss`, keeping unrelated auth-UI styling.

## Test plan
- [x] `npm run lint:scss` clean on touched files
- [x] `npx vitest run frontend/app/view/agent` — 706/707 (1 pre-existing unrelated flaky timeout in `tool-renderers/registry.test.ts`, not touched by this change)
- [x] Live-verified via CDP against a `task dev` instance: opened an agent pane, confirmed `.agent-activity-log` no longer exists in the DOM and `.agent-shell-subblock` renders in its place; screenshot confirms normal rendering, no layout breakage
- [ ] Live bang-command → terminal-write redirect not independently confirmed end-to-end — the only pane available to test against turned out to be this very agent's own live session (recursive/unsafe to poke at further), so I backed off rather than risk it. Code path is short and typechecked (`Terminal.write` call mirrors `AgentInstallModal.tsx`'s proven usage) but worth a manual `!echo hi` sanity check before merge.


a5af:
@codex review

PR #2279, opened 2026-07-23
fix(term): terminal scrollback now survives reconnect (all view:term panes)

AgentO-asaf:
## Summary
- Fixes a pre-existing gap discovered while testing PR #2278: every terminal in the app — standalone Terminal panes and the agent-shell drawer alike — loses all scrollback on remount (drawer close/reopen, tab teardown/rebuild). Not specific to #2278's changes; both surfaces share the same `ShellController` and `TermWrap`.
- `TermWrap.loadInitialTerminalData()` was already correct and already calls the right two restore endpoints — both were dead ends server-side:
  - **Part A**: raw PTY output write-through was hardcoded off (`handle_append_block_file(..., None, None)`). `ShellController` never received the `filestore` that `PersistentController`/`SubprocessController` already get from the same factory call site (`blockcontroller/mod.rs`) — threaded it through the constructor and the PTY read loop.
  - **Part B**: the periodic-snapshot RPC (`SaveTerminalState`) was an unimplemented stub silently discarding what the frontend already sends every ~5s. Implemented it — persists to the `cache:term:full` blockfile with `ptyoffset`/`termsize` meta, matching exactly what the frontend reads back.
- No frontend changes needed (its restore logic was already correct) and no new storage primitives needed (reuses `FileStore::write_file`/`write_meta`/`append_data` already used by other controllers).
- Full design writeup: `docs/specs/SPEC_TERMINAL_SCROLLBACK_PERSISTENCE_2026_07_23.md`.

## Test plan
- [x] `cargo build -p agentmux-srv` clean
- [x] `cargo test -p agentmux-srv` — 1637/1637 passing (4 new: constructor wiring + 3 `SaveTerminalState` behavior tests)
- [x] `cargo clippy -p agentmux-srv --tests` clean on all touched files
- [ ] Live end-to-end verification (close/reopen a terminal or the agent shell drawer, confirm scrollback reappears) — not yet done in a real running instance; recommend a manual check before merge

### DIFF
--- a/.changesets/1784822275-fix-term-terminal-scrollback-now-survives-reconnect-all-view-d1q7.md
+++ b/.changesets/1784822275-fix-term-terminal-scrollback-now-survives-reconnect-all-view-d1q7.md
@@ -1,5 +1,5 @@
 ---
-type: minor
+type: patch
 ---
 
 fix(term): terminal scrollback now survives reconnect (all view:term panes)
--- a/.changesets/1784822275-fix-term-terminal-scrollback-now-survives-reconnect-all-view-d1q7.md
+++ b/.changesets/1784822275-fix-term-terminal-scrollback-now-survives-reconnect-all-view-d1q7.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(term): terminal scrollback now survives reconnect (all view:term panes)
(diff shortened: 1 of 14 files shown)


a5af:
@codex review

PR #2282, opened 2026-07-23
feat(tabs): shrink-to-fit pane tab strip + double-click rename

AgentY-asaf:
## Summary
- **Shrink-to-fit strip**: with one tab open, the strip is exactly `[tab][+]` wide instead of stretching to full pane width with dead trailing space (`display: inline-flex` + dropping the tabs' flex-grow).
- **Content-sized tabs**: tabs size to their label + padding instead of an 80–200px flex-grow band, capped at 240px with existing ellipsis truncation for long names.
- **Double-click to rename**: both agent fork tabs and terminal tabs. No changes needed to the shared `PaneTabStrip` component — reuses its existing `onTabDoubleClick`/`renderLabel` hooks via a new shared `PaneTabRenameInput`.
  - Terminal tabs persist via the existing `pane-title` block-meta mechanism (same one the block title bar already uses).
  - Agent fork tabs persist via a new dedicated `renameagentdefinitiontitle` RPC that writes `branch_label` when the definition already has one (a fork) else `name` (a lineage root) — whichever field `titleOf()` actually displays. Kept deliberately separate from `updateagent`, which stays immutable on `branch_label` for every other caller.

Full design/rationale: `docs/specs/SPEC_PANE_TAB_STRIP_COMPACT_SIZING_AND_RENAME_2026_07_22.md`.

## Test plan
- [x] New tests: `PaneTabRenameInput.test.tsx` (7 tests — autofocus/select, Enter/blur/Escape commit-cancel semantics, empty/unchanged no-op, no double-fire).
- [x] `PaneTabStrip.test.tsx` (11 tests, unchanged) still pass — sizing change is SCSS-only.
- [x] `npx vitest run` (frontend) — 1868/1869 pass (1 pre-existing, unrelated flaky timing test, confirmed passes in isolation).
- [x] `cargo test -p agentmux-srv` — 1589 + 4 + 7 pass, including `dual_write_agent_def_update_refreshes_name_in_db_agents` and the full `agent_handlers`/template suite.
- [x] `npx tsc --noEmit` / `cargo check` / `cargo build` — clean (one pre-existing unrelated `qrcode` module error).
- [ ] Manual: open a single agent/terminal pane, confirm no trailing dead space; add a second tab, confirm width matches its label; double-click a tab, confirm rename persists across pane close/reopen.

### DIFF
--- a/.changesets/1784830298-feat-tabs-shrink-to-fit-pane-tab-strip-sizing-double-click-r-q2ac.md
+++ b/.changesets/1784830298-feat-tabs-shrink-to-fit-pane-tab-strip-sizing-double-click-r-q2ac.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(tabs): shrink-to-fit pane tab strip sizing + double-click rename for agent fork and terminal tabs
--- a/agentmux-srv/src/backend/rpc_types/agent.rs
+++ b/agentmux-srv/src/backend/rpc_types/agent.rs
@@ -635,3 +635,11 @@ pub struct CommandForkAgentDefinitionSuggestData {
 pub struct ForkAgentDefinitionSuggestResult {
     pub suggested_label: String,
 }
+
+/// SPEC_PANE_TAB_STRIP_COMPACT_SIZING_AND_RENAME_2026_07_22.md §4 — renames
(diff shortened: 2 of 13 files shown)


a5af:
@codex review

PR #2284, opened 2026-07-23
docs(reports): Armory architecture and naming review

Agent2-asaf:
## Summary
Reassesses the Armory's architecture (app-wide pane + per-agent Agent-setup modal) and researches whether "Accounts" should be renamed to "Connectors", per request.

**Findings:**
- The incremental, spec-driven evolution (Trust Center -> Armory -> Preset->Bundle -> Phase 4/5 consolidation -> architecture doc) was handled well; the doc's own "reusable resource in Armory, scoped binding lives with the agent" principle is sound and followed by 4/5 primitives.
- Real rough edges: an `ArmorySection` id/label mismatch (`id:"brain"` -> labeled "Memories", `id:"memories"` -> labeled "Bundles"), hand-duplicated (not shared) model code between `AgentMcpModel`/`McpCatalogModel` and `AgentSkillModel`/`SkillCatalogModel`, leftover "Preset" copy/RPC alias past its stated removal phase, and Startup being a structural outlier by design (no catalog of its own).
- **One real bug, not just an architecture note:** the per-agent "Accounts" tab (`AgentIdentityModalPanel`) writes to `AgentDefinition.accounts`, a column the spawn-time credential resolver never reads (it reads `db_agent_identity_links` exclusively, written only by the launch flow) -- the tab looks functional but silently no-ops. Recommend filing as a follow-up bug.
- **Naming recommendation: keep "Accounts", don't rename to "Connectors".** A draft spec already reserves "Connectors" for a different concept (preloaded MCP server integrations) one tab over. Industry precedent for this exact UI slot (Notion, Zapier) converges on "Connections"/"Connected accounts", not "Connectors". If renamed at all, "Connected Accounts" is the only zero-collision, precedent-backed option.

This is a research/analysis report, not a code change -- no functional changes in this PR.

PR #2285, opened 2026-07-23
fix(agent): Agent setup's Accounts tab wrote to a column nothing reads

Agent2-asaf:
## Summary
Found during the Armory architecture review (docs/reports/REPORT_ARMORY_ARCHITECTURE_AND_NAMING_REVIEW_2026_07_23.md, #2284, §2.2). `AgentSetupModal`'s "Accounts" tab looked like fully functional per-agent credential-assignment UI, but wrote to `AgentDefinition.accounts` -- a legacy JSON blob column that the spawn-time credential resolver (`identity/resolver.rs::resolve_bindings_for_instance`) never reads. It reads only `db_agent_identity_links`, written exclusively by the agent-launch flow. So a user picking "assign this account to this agent" in that modal saw it save successfully, with zero effect on what credentials the agent actually launches with.

**Fix:** replaced the write-path modal with `AgentIdentityLinksPanel` -- the existing, already-tested, read-only view built against the table the resolver actually reads (it already backs the agent pane's own Identity tab). It only needs `agentId`, so as a bonus it now also works for quick-launch panes that previously hit an "unavailable" fallback. Removed the `agent`/`blockId` plumbing that only existed to support the removed write path (`AgentSetupRequest`, `modal-dispatch.tsx`, `agent-view.tsx`'s `modalLayer.open` call), and the CSS rule that only ever styled the removed component.

`AgentIdentityModalPanel`/`AgentIdentityPanel.tsx` are left in place -- they were already unreachable dead code before this change (the standalone `"agent-identity"` modal-layer kind they back has no live caller anywhere, before or after this PR) -- not touched here to keep this fix scoped to the actual bug.

## Test plan
- [x] `npx tsc --noEmit` -- clean
- [x] `npx vitest run` across `frontend/app/view/agent` + `frontend/app/element` -- 738/739 passing (1 unrelated pre-existing flaky timeout in `tool-renderers/registry.test.ts`, confirmed passing standalone in 1.8s)
- [x] `frontend/app/view/identity/agent-identity-links-panel.test.tsx` already covers the reused component's per-agentId scoping behavior
- [x] Verified no other caller of the removed `agent`/`blockId` fields on `AgentSetupRequest`

### DIFF
--- a/.changesets/1784846138-fix-agent-accounts-tab-in-agent-setup-wrote-to-a-dead-column-qwyy.md
+++ b/.changesets/1784846138-fix-agent-accounts-tab-in-agent-setup-wrote-to-a-dead-column-qwyy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): Accounts tab in Agent setup wrote to a dead column
--- a/frontend/app/element/modal-dispatch.tsx
+++ b/frontend/app/element/modal-dispatch.tsx
@@ -284,11 +284,9 @@ export function renderRequest(
                 label: requestLabel(req),
                 panel: (
                     <AgentSetupModal
-                        agent={req.agent}
                         agentId={req.agentId}
                         agentName={req.agentName}
                         workingDirectory={req.workingDirectory}
-                        blockId={req.blockId}
                         initialTab={req.initialTab}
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2286, opened 2026-07-23
fix(subagent-watcher): scope live fs events to the owning block and tear down watchers on block delete

Agent1-asaf:
## Summary

Fixes the root cause behind a user report of an unfamiliar agent slug appearing in the Swarm/conversation UI ("nobile-percolating-ritchie ab9384" — log-verified actual spelling: `noble-percolating-ritchie`). Full investigation with evidence: `docs/retro/retro-subagent-watcher-shared-dir-fanout-and-leak-2026-07-23.md`.

Two independent bugs in `agentmux-srv/src/backend/subagent_watcher.rs`, both real and both required to fully close the symptom:

**Bug A — unscoped shared-directory fan-out.** Agents without a per-identity bundle override all resolve to the same default Claude config dir (`resolve_claude_config_dir`), so their `notify` filesystem watches observe the *same* directory tree. The live-watch dispatch loop attributed any changed file to its own `parent_agent`/`parent_block_id` with no check that the file's session actually belongs to it — one agent's real subagent write fanned out to every other watcher sharing that path, each misattributing it to itself.

Fix: a new `session_belongs_to_block(block_id, session_id)` check — compares the changed file's derived session id against the block's own currently-persisted `agent:sessionid` meta (read live from `wstore`) — gates every live fs event before it's processed.

**Bug B — leaked watchers on ungraceful pane close.** The only teardown for a block's filesystem watcher (`unwatch_agent`) runs through the frontend's graceful `/agentmux/reactive/unregister` round trip — something a crash, API-driven delete, or cascade delete can all skip (already documented in the existing `prune_block` doc comment). The "robust backstop" built for exactly that gap (`prune_block`, driven by `BlockDeleted`/`TabDeleted`/`WorkspaceDeleted`) only pruned derived session/dispatch/pending-activity state — it never tore down the underlying watcher. So a closed block's watcher leaked indefinitely, continuing to receive shared-directory events and (per Bug A) re-creating fresh misattributions for days after the pane was gone — confirmed live: five closed panes' block_ids no longer exist in any `objects.db` on the machine, yet their misattributions were still firing today.

Fix: added `parent_block_id` to `WatchedAgent`, a new block-scoped `unwatch_block` (mirrors `unwatch_agent`'s teardown mechanism, keyed by block instead of agent name so an agent identity reused across blocks over time isn't affected), called from `prune_block`.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv subagent_watcher::` — 56/56 passing, including 4 new tests:
  - `session_belongs_to_block_matches_only_the_blocks_own_persisted_session`
  - `session_belongs_to_block_is_false_for_a_block_that_no_longer_exists`
  - `prune_block_also_tears_down_that_blocks_filesystem_watcher`
  - `live_fs_event_is_not_misattributed_to_a_block_that_does_not_own_the_session` — end-to-end: two blocks share one `config_dir` (the real-world scenario), a file is written to a session owned by only one of them, and only that block records the subagent.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` (full suite) — 1643/1643 passing, 0 failed, 4 ignored (pre-existing).
- [x] `cargo check -p agentmux-srv` — clean.

### DIFF
--- a/.changesets/1784846624-fix-subagent-watcher-scope-live-fs-events-to-the-owning-bloc-p215.md
+++ b/.changesets/1784846624-fix-subagent-watcher-scope-live-fs-events-to-the-owning-bloc-p215.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(subagent-watcher): scope live fs events to the owning block and tear down watchers on block delete
--- a/agentmux-srv/src/backend/subagent_watcher/mod.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/mod.rs
@@ -37,7 +37,13 @@
 //!     `Event::BlockDeleted`/`TabDeleted`/`WorkspaceDeleted`). Without this,
 //!     `ListActive`/`ListDispatches` kept returning a closed block's
 //!     subagents forever, so the Swarm pane kept rendering a ghost row for
-//!     it until the whole app/srv restarted.
+//!     it until the whole app/srv restarted. `prune_block` also tears down
+//!     the block's own filesystem watcher (`unwatch_block`) — without that,
+//!     a closed block's watcher leaked indefinitely and kept re-creating
+//!     fresh (mis-)attributions the next time another agent sharing its
+//!     watched directory wrote to its own subagent transcript; see
+//!     `session_belongs_to_block` and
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2287, opened 2026-07-23
refactor: implement Tier 3 large-file modularization items

AgentX-asaf:
## Summary

Implements the "Tier 3 — genuine value but requires reading and respecting an in-code incident warning before touching" items from `docs/reports/REPORT_LARGE_FILE_MODULARIZATION_SCAN_2026_07_22.md`. These were the highest-risk items in the whole scan, each carrying a documented production-incident or security-invariant warning, so each was handled deliberately with explicit constraints and independently re-verified by me (beyond the implementing agent's own checks) before committing.

| # | Item | Constraint honored |
|---|---|---|
| 1 | `main.rs` → `bootstrap.rs` | 1165-line `async fn main()` extracted into named phase functions, byte-for-byte identical logic. Verified via the real subprocess-spawn integration test (`sigterm_exits_process`), not just unit tests. 1509 → 178 lines. |
| 2 | `lib.rs` → `macos_compat.rs` + `logging.rs` | `run()` itself (the CEF bootstrap, ordering-hazard comments throughout) was **not touched** — only the two genuinely self-contained pieces it calls into were relocated. 1949 → 1121 lines. |
| 3 | `identity/resolver.rs` → `resolver/*` | `inject_identity_env_with_broker` (enforces `SPEC_PROVIDER_ISOLATION`'s INV-A — see `docs/retro/retro-auth-isolation-invariant-silently-orphaned-2026-07-14.md` for the incident where it was silently orphaned once already) moved as one atomic unit with its full 21-test suite and its warning comment kept verbatim and directly attached. I independently re-read the final file and re-ran the tests myself before trusting this one. |
| 4 | `useAgentStream.ts` → `stream-flush-queue.ts` + 4 hooks | Per `RETRO_REPLACECHILD_CRASH_2026-06-06.md`, extracted the shared single-RAF/single-`batch()` flush mechanism **first**, then routed every event-source producer through it rather than giving any of them independent scheduling. I independently grepped all 6 files myself and confirmed exactly one `requestAnimationFrame` and one `batch()` call site before committing. |

**Explicitly not touched:** `client/lifecycle.rs` and `supervisor/windows.rs` — both flagged KEEP-AS-IS in the audit, needing a precursor context-struct refactor before a file split is safe. Forcing a split without that groundwork was out of scope.

## Test plan

- [x] `cargo build --workspace` — clean, only pre-existing warnings
- [x] `cargo test --workspace` — all pass, 0 failed across every crate
- [x] `cargo test -p agentmux-srv --test integration_test` — 4/4 pass (real subprocess-spawn tests, exercises the new bootstrap path end-to-end)
- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx vitest run` — 1912/1915 passed, 3 skipped
- [x] Manually re-verified the two highest-risk items myself (INV-A warning placement + resolver test count; single-RAF/single-batch invariant) rather than relying solely on the implementing agents' self-reports


a5af:
@codex review

PR #2289, opened 2026-07-24
fix(tabs): hide single tab pill until a 2nd tab exists, transparent strip bg

AgentY-asaf:
## Summary
Live-QA follow-up on PR #2282. Two visual issues reported while running the change:
- A lone conversation/terminal still showed its own tab pill next to the `+`. Now the tab list passed to `PaneTabStrip` is empty until there's a 2nd tab to switch between — only `+` renders for a single tab; the first tab appears alongside it the moment a 2nd one opens.
- `.pane-tab-strip`'s own background fill painted a colored block over any stray container width beyond the actual tabs/`+`. Removed it — only `.pane-tab` (active/hover) and `.pane-tab-strip-add` (hover) paint a background now, so empty area reads as transparent.

## Test plan
- [x] `npx vitest run app/element app/view/agent app/view/term` — 63 files / 761 tests pass.
- [ ] Manual: confirm a single agent/terminal pane shows only `+`, and opening a 2nd tab reveals both.

### DIFF
--- a/.changesets/1784870112-fix-tabs-hide-the-single-tab-pill-until-a-2nd-tab-exists-tra-t4p4.md
+++ b/.changesets/1784870112-fix-tabs-hide-the-single-tab-pill-until-a-2nd-tab-exists-tra-t4p4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

Issue #121, opened 2026-03-13
perf: PTY spawn serializes on concurrent pane creation (~10ms × N stall)


AgentU-asaf:
## Summary

Opening multiple terminal panes quickly produces a noticeable UI stall. Each new pane triggers a `controllerresync` RPC that takes **~10ms** on the backend (PTY + shell spawn). When N panes are opened concurrently, these calls serialize, producing an **N × ~10ms** stall before any terminal becomes interactive.

Timing logs from v0.31.120 (4 panes opened in a burst):

```
Pane 1: 12.73 ms  (14:23:33.588)
Pane 2:  9.66 ms  (14:23:33.708)
Pane 3:  9.43 ms  (14:23:33.845)
Pane 4:  9.83 ms  (14:23:35.963)
──────────────────────────────────
Total:  ~41.65 ms serialized — perceived as sluggishness
```

---

### 1. `spawn_command()` runs inline on the async executor

In `shell.rs::start()`, `pair.slave.spawn_command(cmd)` (i.e., `fork()` + `exec()`) is a blocking OS call taking 3–10ms. It runs inline in a Tokio async task, violating the <100µs rule. This starves other tasks sharing the executor.

### 2. Frontend `await`s the full spawn before displaying the pane

In `termwrap.ts::init()`:

```typescript
// Subscribe to PTY data ← happens BEFORE resync
this.mainFileSubject.subscribe(this.handleNewFileSubjectData.bind(this));
// ...
await this.resyncController("init"); // ← awaits full fork+exec (~10ms)
```

The PTY data subscription is already registered before `resyncController` is called, so the pane will receive output regardless. **For a new (empty) pane there is no correctness reason to await.** The `await` was the only thing serializing the four concurrent open operations from the frontend side.

### 3. `fork()` page table cost scales with parent VSS

As the app grows (more modules loaded, more memory mapped), `fork()` latency increases because Linux must copy the parent's page table even with CoW. Research shows this can reach 10–50ms in large processes ([Rust issue #87764](https://github.com/rust-lang/rust/issues/87764), [fork latency analysis](https://www.whexy.com/en/posts/fork)).

---

### Fix A — Drop `await` on `resyncController` for new panes (1 line, highest ROI)

**`frontend/app/view/term/termwrap.ts`**

```typescript
// Before:
await this.resyncController("init");

// After:
this.resyncController("init"); // fire-and-forget; subscription already active
```

[...]

PR #2291, opened 2026-07-24
fix(term): drop await on init resync to stop N-pane spawn serialization

GenericAgentX-asaf:
## Summary
Implements Fix A from #121: opening N terminal panes concurrently produced an N×~10ms UI stall because `termwrap.ts::init()` awaited `resyncController("init")` before returning, even though the PTY data subscription is already active by that point (there's no correctness reason to block on it for a new pane). Moves `hasResized = true` ahead of the call (per the issue's own review comment, so a reconnect event arriving mid-spawn during the fire-and-forget window isn't missed by the `TermResyncHandler` gate) and fires `resyncController` without awaiting, matching the codebase's existing `void this.foo()` fire-and-forget convention.

Verified via `git grep` that this was still unapplied (the SolidJS migration mentioned in the issue's comments landed, but this specific line was untouched).

**Scope note:** this is Fix A only. Fix B (wrapping the backend's `spawn_command` fork+exec in `tokio::task::spawn_blocking`) is real, agreed-upon follow-up work, but turned out to touch 3 separate async RPC call sites (`websocket.rs`, `agent_open.rs` ×2) with Send/'static implications across the PTY-spawn hot path — bigger than a same-PR change I wanted to make without dedicated review. Left #121 referenced but not auto-closed by this PR since B is still open; will note in the issue.

Fixes (partially) #121.

## Test plan
- [x] `tsc --noEmit` passes with no new errors
- [ ] Manually verify opening 4+ terminal panes in a burst no longer visibly serializes (dev build)

### DIFF
--- a/.changesets/1784893740-fix-term-drop-await-on-init-resync-to-stop-n-pane-spawn-seri-anu6.md
+++ b/.changesets/1784893740-fix-term-drop-await-on-init-resync-to-stop-n-pane-spawn-seri-anu6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): drop await on init resync to stop N-pane spawn serialization (#121)
--- a/frontend/app/view/term/termwrap.ts
+++ b/frontend/app/view/term/termwrap.ts
@@ -329,8 +329,12 @@ export class TermWrap {
         // At this point we are fully subscribed and ready to receive data.
         this.customFit();
         this.sendTermSize();
-        await this.resyncController("init");
+        // hasResized must flip before the fire-and-forget resync below: it gates the
+        // TermResyncHandler effect, and the terminal is already sized at this point
+        // (customFit/sendTermSize above), so a reconnect event arriving mid-spawn must
+        // not be dropped. See issue #121.
         this.hasResized = true;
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed in 229d203d: Codex's P2 was right — the rAF post-paint re-fit's `sendTermSize()` could fire before the fire-and-forget `resyncController("init")` had actually created the block's controller backend-side. Confirmed the mechanism in `agentmux-srv/src/backend/blockcontroller/mod.rs`: `send_input` returns `Err("no controller for block {block_id}")` when the controller doesn't exist yet, and the websocket handler (`websocket.rs:543`) just logs that at `debug!` and drops it — no queue, no retry. So a dimension change caught by the re-fit (the exact case that code exists for) could leave the PTY at stale geometry until a manual resize.

Fix: store the "init" resync's promise and await it before the re-fit's corrective `sendTermSize()` — but only on the branch where a resize is actually needed, so the common no-change case (most panes, most of the time) stays instant and N panes still spawn concurrently. `npx tsc --noEmit` clean.

PR #2292, opened 2026-07-24
feat(agent): fix silent auto-scroll drops, extend scrollbar, replace Host row with a tag

AgentY-asaf:
## Summary
Three related fixes from live QA in the agent pane:

- **Silent auto-scroll-follow drops.** The stick-to-bottom effect in `AgentDocumentVirtualList` only re-ran when the message count changed, not when an existing (possibly off-screen) row's height changed — a tool panel collapsing/expanding, code re-highlighting, an image loading. Those events flow through a separate `layoutView` signal the effect never depended on, so `scrollTop` silently desynced from true bottom with no visible state change. Fixed by tracking `layoutView().totalSize` as a second dependency (an existing, already-correct signal — no new state introduced).
- **Removed the "Host — full system access" / "Container — isolated Docker sandbox" row** (confirmed inert — no click handler was ever wired to it) and replaced it with a compact **HOST** (red) / **SANDBOX** (white) tag next to the model selector in the composer strip.
- **Scrollbar now reaches the bottom of the pane.** With the Host row gone, only `AgentWorkingRow` ("Working…"/"✓ Worked") sat between the message list and the composer. Moved it out of normal flex flow into an absolutely-positioned overlay anchored to the bottom of a new `.agent-document-scroll-region` wrapper, so the scrollbar now spans that whole region instead of stopping short by the row's height. A `ResizeObserver` keeps the message list's bottom padding in sync with the row's actual height (0 when hidden) so scrolling to true bottom never hides the last message underneath it.

Full design/rationale: `docs/specs/SPEC_AGENT_PANE_SCROLL_FOLLOW_AND_STATUS_OVERLAY_2026_07_24.md`.

## Test plan
- [x] `npx vitest run app/view/agent` — 59 files / 721 tests pass.
- [x] Manual, live: verified via CDP against the running dev build — confirmed the HOST tag renders correctly next to the model selector with no separate row above the composer.
- [ ] Manual: reproduce the original auto-scroll-drop bug (expand/collapse an off-screen tool panel while streaming) and confirm the view stays pinned. Confirm scrolling to true bottom with `AgentWorkingRow` visible doesn't hide the last message.


a5af:
@codex review

PR #2293, opened 2026-07-24
fix(agent): auto-clear error activity-dock rows after 15s, add landing/departure flash

GenericAgentX-asaf:
## Summary
- `ActivityDock` (the strip of pinned shell/subagent rows above the composer) gave `error`-status rows `Infinity` retention -- "persist until manually dismissed." Reported live: a session with several failed background shells accumulated permanent rows above the composer with no self-clearing ("docks racking up above the model selector bar").
- `RETENTION_MS.error`: `Infinity` -> `15_000` (15s) -- still well above `done`'s 8s / `stopped`'s 3s, long enough to actually read the error. The manual × dismiss button still works for clearing one early.
- Added a landing/departure flash so a row's appearance and (now genuinely imminent) disappearance are both conspicuous: reuses `tab-bounce` (`tabbar.scss`) -- the same spring-bounce already played when a dragged tab lands -- rather than a new bespoke animation. `ActivityDock`'s visibility window is extended by `EXIT_FLASH_MS` (400ms) past nominal retention so the exit flash has time to play before the row actually leaves the DOM.
- Retro: `docs/retro/retro-activity-dock-error-retention-2026-07-24.md` -- also documents 4 other independent ad-hoc "flash" implementations already in the codebase as a scoping note for a possible follow-up reusable primitive (not built in this PR, deliberately out of scope).

## Test plan
- [x] `tsc --noEmit` clean
- [ ] Manually verify: spawn a background shell that fails, confirm the row flashes in, sits ~15s, flashes out and clears without manual dismissal

### DIFF
--- a/.changesets/1784922128-fix-agent-auto-clear-error-activity-dock-rows-after-15s-add--odmy.md
+++ b/.changesets/1784922128-fix-agent-auto-clear-error-activity-dock-rows-after-15s-add--odmy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): auto-clear error activity-dock rows after 15s, add landing/departure flash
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

Issue #2218, opened 2026-07-18
tracking: renderer pool commit-charge growth — driver-committed memory not reclaimed until restart (§B.4/B.5)


AgentA-asaf:
## Summary

Consolidated tracker for the recurring "pagefile/commit charge grows until the machine needs a restart" problem. Supersedes #1936 (closed 2026-07-04, before the follow-up experiment below ran). Confirmed live again today (2026-07-18) via the newly-merged commit-attribution diagnostic (#2207): with 2 AgentMux instances running, `agentmux_mb` (1.8GB) + `panes_mb` (0) + `other_mb` (~5GB) summed to ~6.4GB while system-wide commit was **24.4GB** — the same ~18GB unattributed gap both retros below already diagnosed.

## Root cause (already diagnosed, not yet fully fixed)

- `docs/retro/retro-commit-charge-pagefile-growth-2026-07-02.md` — first measurement: ~43 of 52GB commit was **unattributed to any process**, traced to AgentMux's per-pane isolated `RequestContext` design (`agentmux-cef/src/commands/mod.rs:25-30`). Every window/pane gets its own GPU-accelerated Chromium renderer; each one makes the **GPU driver** commit kernel-mode, pagefile-backed system memory (WDDM paging buffers, GPU-VA structures) charged to the **System process (PID 4)**, invisible to any per-process tool including our own attribution diagnostic.
- `docs/retro/retro-commit-restart-reclaim-2026-07-16.md` — follow-up experiment (direct response to "why does it grow? can we stop it somehow?"): proved the commit is **live, not leaked** — closing AgentMux fully released 43.6GB instantly. Measured rule of thumb: **~1GB of invisible driver commit per AgentMux renderer process**, on top of its own visible private bytes.
- `docs/specs/SPEC_MEMORY_COMMIT_ATTRIBUTION_CORRECTION_2026_07_02.md` §B — the concrete fix list.

## Status of the §B fix list

| Item | Description | Status |
|---|---|---|
| B.1 | Prune WPS broker `persist_map` on block/shell close | ✅ shipped (`wps.rs::purge_scope`) |
| B.2 | Bound per-connection WS egress channels | ✅ shipped (`eventbus.rs` uses bounded `mpsc::channel`) |
| B.3 | Cap `SubagentWatcher` per-session event Vec | ✅ shipped (`MAX_SUBAGENT_EVENTS`) |
[...]

PR #2294, opened 2026-07-24
feat(cef): add dev-only GPU memory trace commands for #2218 diagnostics

GenericAgentX-asaf:
## Summary
Implements Tier 2 of `docs/specs/SPEC_GPU_MEMORY_TRACING_SCAFFOLDING_2026_07_24.md` -- diagnostics scaffolding for #2218 (system commit charge grows ~0.8 GB/hour, independent of both renderer count and logged content activity per that issue's investigation). Process-level counters can't see inside the GPU process's own allocator, which is exactly where the "unattributed" growth lives -- this adds the ability to actually look.

- `begin_gpu_trace` / `end_gpu_trace` RPC commands (`agentmux-cef/src/commands/window/gpu_trace.rs`), wrapping CEF's `begin_tracing`/`end_tracing` to capture Chromium's memory-infra GPU categories (`gpu`, `cc`, `skia`, `GPUMemoryBuffer`, `disabled-by-default-memory-infra`) for an arbitrary-length, operator-controlled capture window -- unlike `--trace-startup`, which is bounded and requires restarting before the capture window starts.
- Modeled directly on the existing `toggle_devtools` handler (`commands/window/meta.rs`), same signature shape, wired into `ipc.rs` the same way.
- Dev-only diagnostics -- no UI affordance, not a user feature.
- Resolves the one open unknown from the spec's first pass: verified against the exact resolved crate version (`cef 148.3.0+148.0.9`, read from the local Cargo registry cache, not guessed) that `begin_tracing`/`end_tracing` are exposed at the crate root, along with the `wrap_end_tracing_callback!` macro (first use of that pattern in this codebase).

## Test plan
- [x] `cargo check -p agentmux-cef` clean, zero warnings/errors in the new file
- [ ] Live capture during a confirmed-idle multi-hour session (blocked on this machine right now by an unrelated `task dev` PATH issue -- code is verified by typecheck only, not yet run)

### DIFF
--- a/.changesets/1784926378-feat-cef-add-dev-only-gpu-memory-trace-commands-for-2218-dia-w3mj.md
+++ b/.changesets/1784926378-feat-cef-add-dev-only-gpu-memory-trace-commands-for-2218-dia-w3mj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(cef): add dev-only GPU memory trace commands for #2218 diagnostics
--- a/.changesets/1784926378-feat-cef-add-dev-only-gpu-memory-trace-commands-for-2218-dia-w3mj.md
+++ b/.changesets/1784926378-feat-cef-add-dev-only-gpu-memory-trace-commands-for-2218-dia-w3mj.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
 feat(cef): add dev-only GPU memory trace commands for #2218 diagnostics
--- a/agentmux-cef/src/commands/window/gpu_trace.rs
+++ b/agentmux-cef/src/commands/window/gpu_trace.rs
@@ -13,13 +13,22 @@
 // Spec: docs/specs/SPEC_GPU_MEMORY_TRACING_SCAFFOLDING_2026_07_24.md §2.2.
 // Not a user feature — no UI affordance. Diagnostics only, same tier as
 // `toggle_devtools`/`inspect_element_at` (meta.rs), which this is modeled on.
+//
+// `cef::begin_tracing`/`cef::end_tracing` must run on the CEF UI thread, same
+// as every other CEF-touching call in this codebase (see `post_show_dev_tools`
+// / `ShowDevToolsTask`, `ui_tasks/window.rs`) — this handler is invoked from
+// the IPC/Tokio task, not that thread, so both calls are marshaled via
+// `post_task(ThreadId::UI, ...)` rather than called inline. That makes both
+// RPC commands fire-and-forget: the real begin/end outcome is logged from
+// inside each `Task::execute()`, not returned synchronously to the caller
+// (matches `post_show_dev_tools`'s own fire-and-forget shape).
 
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use cef::{
-    rc::Rc, wrap_end_tracing_callback, CefString, EndTracingCallback, ImplEndTracingCallback,
-    WrapEndTracingCallback,
+    post_task, rc::Rc, wrap_end_tracing_callback, wrap_task, CefString, EndTracingCallback,
+    ImplEndTracingCallback, ImplTask, Task, ThreadId, WrapEndTracingCallback, WrapTask,
 };
 
 use crate::state::AppState;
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed the P1 (UI-thread marshaling) and the second P2 (spec's semicolon vs. space-separated \`AGENTMUX_CEF_EXTRA_FLAGS\` bug — verified against the actual parsing code, real bug, fixed) in the latest commit.

On the first P2 (\`DEFAULT_TRACE_CATEGORIES\` mixing \`-*\` with included categories) — pushing back rather than changing it. This is Chromium's own documented category-filter shorthand, not something this PR invented: their official startup-tracing docs (https://chromium.googlesource.com/chromium/src/+/112.0.5615.165/docs/memory-infra/memory_infra_startup_tracing.md) use the identical pattern — \`--trace-startup=-*,disabled-by-default-memory-infra\`. \`-*\` means "start from a blank category set instead of the default one"; everything after it in the comma-separated list is what gets included. My string is the same shape with more categories listed. Since CEF's \`begin_tracing\` wraps the same underlying Chromium \`TraceConfig\` string parser (not a CEF-specific format), it should accept it identically. Happy to be shown a concrete case where this breaks, but wanted to leave the reasoning on record rather than silently drop or silently "fix" something I believe is actually correct.

GenericAgentX-asaf:
Pushing back on the P0 (\`path = written\` at gpu_trace.rs:198) — this doesn't reproduce.

- \`cargo check -p agentmux-cef\` against \`f23642d9\` (the exact commit this review is against — confirmed via the review's own \`commit_id\`) compiles clean, zero errors, including that line. Ran it twice just now to be sure.
- Stronger than a typecheck: this exact code is live right now in a running \`task dev\` instance — I called \`begin_gpu_trace\` over the \`/ipc\` RPC a few minutes ago (confirmed via the srv log: \`gpu_trace: started\`), and the callback containing this exact line (\`on_end_tracing_complete\`) will execute in ~15 more minutes when the capture auto-stops. None of that would be possible if this were a real compile error — there's no such thing as a Rust compile error that only manifests when a code path actually runs; the whole crate fails to build or it doesn't.

\`tracing::field::Value\` does have an impl for owned \`String\` in the version this workspace pins — that's presumably where the review's claim is wrong. Not changing this line; the evidence above should be reproducible by anyone (\`git checkout f23642d9 && cargo check -p agentmux-cef\`).

GenericAgentX-asaf:
Went through all 5 review rounds against current HEAD rather than assuming each was still live:

**Fixed (46833d2d):** round 5's P1 — `resolve_trace_path` blocked `/`, `\`, `\0`, `.`, `..` but not `:`. On Windows, `PathBuf::join`/`push` replaces the base path entirely for a component with a prefix-but-no-root (e.g. `C:evil.json` is drive-relative), so `dir.join("C:evil.json")` would silently discard the `TRACE_SUBDIR` confinement. Extracted the check into `is_valid_trace_filename` with unit tests covering the escape case, plain filenames, and the existing separator/special-component rejections.

**Already fixed by earlier commits on this branch** (verified by reading current code, not just trusting round numbering): round 1's dev-gate P1 (`require_dev_mode()` is present on both commands) and round 1's changeset-type P2 (already `minor`), round 2's CEF-UI-thread P1 (`post_task(ThreadId::UI, ...)` used in both `wrap_task` blocks) and the `AGENTMUX_CEF_EXTRA_FLAGS` semicolon-splitting P2.

**Turned out to be false positives, verified rather than assumed:**
- Round 3's P0 (`tracing::info!(path = written, ...)` allegedly failing to compile because `String` doesn't implement `tracing::field::Value`) — `cargo test -p agentmux-cef` actually compiles and passes clean. This version of `tracing` does implement `Value` for owned `String`.
- Round 2's P2 (`DEFAULT_TRACE_CATEGORIES` mixing `-*` with included categories, claimed unsupported) — this is the officially documented Chromium memory-infra startup-tracing pattern. Confirmed against `docs/memory-infra/memory_infra_startup_tracing.md` in chromium/chromium: `--trace-startup=-*,disabled-by-default-memory-infra` is the exact precedent already cited in this file's own doc comment, and the doc's custom-config equivalent (`included_categories: [...], excluded_categories: ["*"]`) confirms include+exclude mixing is intentional, supported syntax.

`cargo test -p agentmux-cef` — 3/3 new tests pass, full crate compiles clean.

PR #2295, opened 2026-07-25
fix(agent): fix invalid Sass compound-selector syntax in RuntimeBadge

GenericAgentX-asaf:
## Summary
\`&--tag&--host\` / \`&--tag&--container\` in \`RuntimeBadge.scss\` used a second bare \`&\` mid-compound-selector, which Dart Sass rejects: \`"&" may only be used at the beginning of a compound selector\`. This broke \`task dev\`'s Vite server with a full-screen compile-error overlay -- found live while using a dev instance for unrelated work.

Both rules were meant to compile to a compound selector combining two BEM modifier classes at once (per the comment directly above them, overriding the single-modifier \`&--host\`/\`&--container\` rules via specificity). \`#{&}\` interpolation produces the identical compiled CSS (\`.runtime-badge--tag.runtime-badge--host\`) without hitting the syntax restriction -- no behavior change, pure syntax fix.

## Test plan
- [x] Applied live against a running \`task dev\` instance -- the compile error cleared, no new Sass errors in the log
- [ ] Visual check: RuntimeBadge still renders the intended override styling when both \`--tag\` and \`--host\`/\`--container\` modifiers are applied together

### DIFF
--- a/.changesets/1784943486-fix-agent-fix-invalid-sass-compound-selector-syntax-in-runti-i4hp.md
+++ b/.changesets/1784943486-fix-agent-fix-invalid-sass-compound-selector-syntax-in-runti-i4hp.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): fix invalid Sass compound-selector syntax in RuntimeBadge, blocking dev server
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #2296, opened 2026-07-25
fix(identity): oauth_config_dir secret_ref serde rename mismatch

GenericAgentX-asaf:
## Summary
Reported live: "what happened to My Agents? its an empty list" -- on a dev instance with real agent definitions (7 rows in `db_agents`) but an empty recent-sessions view.

Root cause, traced end to end:
- `SecretRef::OAuthConfigDir`'s enum-wide `#[serde(rename_all = "snake_case")]` derives `o_auth_config_dir` -- serde's snake_case conversion splits at *every* capital letter, including inside "OAuth" itself.
- Every real account in `~/.agentmux/shared/store.db` (the global, cross-channel account store) was written with `oauth_config_dir` (no extra underscore -- how a human actually spells "OAuth").
- `identity_list()` (`agentmux-srv/src/backend/storage/identities.rs`) aborts entirely on the first row that fails to deserialize -- so one legacy-spelled row breaks the whole query, for every channel/dev-branch that shares the global store (which per `docs/specs/SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md` is deliberately *all* of them -- "My Agents" is designed to show agents from any build/channel/version, not just the current instance).
- `listrecentsessions` (which "My Agents" calls) depends on this query, so it failed too -- silently, because `MyAgentsList.tsx` had `catch { return []; }` with zero logging, indistinguishable from "genuinely no sessions."

## Fix
- Explicit `#[serde(rename = "oauth_config_dir")]` on the `OAuthConfigDir` variant, overriding the derive for just this one acronym-containing name -- matches what's actually on disk.
- New test (`secret_ref_deserializes_real_stored_oauth_config_dir_json`) that pins the literal JSON copied from a real account row, rather than round-tripping through this crate's own serializer (which is how the existing test suite missed this -- write and read both used the same self-consistent-but-wrong tag).
- `MyAgentsList.tsx`'s silent catch now logs the real error via `Logger.error` (including `Error.message`/`.stack`, which don't survive structured-clone/JSON serialization without being pulled out explicitly) -- this is what actually surfaced the root cause during live debugging.

## Test plan
- [x] New unit test passes; full `cargo test -p agentmux-srv` clean (1655 passed, 0 failed)
- [x] Verified live: found the exact error via the newly-added frontend logging, traced it to the exact stored JSON, confirmed the fix compiles
- [ ] Full end-to-end re-verification (needs a backend restart to pick up the fix in a running instance -- `restart_backend` is disabled under launcher-owned lifecycle per this build; next full app restart will pick it up)

### DIFF
--- a/.changesets/1784946191-fix-identity-oauth-config-dir-secret-ref-variant-mismatched--g3hs.md
+++ b/.changesets/1784946191-fix-identity-oauth-config-dir-secret-ref-variant-mismatched--g3hs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): oauth_config_dir secret_ref variant mismatched its own derived serde rename, breaking My Agents for every real account
--- a/agentmux-srv/src/backend/storage/identities.rs
+++ b/agentmux-srv/src/backend/storage/identities.rs
@@ -68,7 +68,16 @@ pub enum SecretRef {
     /// conventionally spelled), so the derived name silently stopped
     /// matching stored data — this makes the wire format explicit instead
     /// of relying on the derive macro to guess right for an acronym.
-    #[serde(rename = "oauth_config_dir")]
+    ///
+    /// The `alias` keeps accepting the derive-produced `o_auth_config_dir`
+    /// tag too: any account persisted by a backend running between this
+    /// variant's introduction and this fix would have round-tripped
+    /// self-consistently under that buggy tag, and `identity_list()` aborts
+    /// entirely on the first unparseable row — so dropping read support for
+    /// it would just move the "My Agents empty" incident to a different set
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed in 5df3ba09: added `#[serde(alias = "o_auth_config_dir")]` alongside the `rename = "oauth_config_dir"`, so deserialization accepts both the legacy derive-produced tag and the canonical one, while serialization always writes the canonical tag. Added two regression tests: one pinning the legacy-tag round-trip, one asserting serialization never emits the legacy tag. `cargo test -p agentmux-srv identities::` passes (7/7).

PR #2298, opened 2026-07-25
feat(armory): add TouchDesigner to the MCP preload catalog

AgentY-asaf:
## Summary
- Adds TouchDesigner as the second entry in the Armory's MCP Servers preload catalog (`frontend/app/view/mcp/mcp-preload-catalog.ts`), following the Ableton Live pattern from #2030 — picking it in "+ Browse catalog" pre-fills the add-server form with `npx -y touchdesigner-mcp-server@latest --stdio`.
- This is the first expansion past Ableton-alone, gated in spec §7/§8 Q1 on "expand only after this pattern validates." That candidate research already existed in `docs/specs/SPEC_ARMORY_PRELOADED_CREATIVE_MCP_CONNECTORS_2026_07_10.md` — this PR is validation, not just research: I set up a real TouchDesigner + Ableton Live instance end-to-end this session (AbletonOSC bridge, the touchdesigner-mcp connector's WebServer DAT component, and confirmed the MCP router comes up with all 12 routes and live tempo/beat data flowing from Ableton), then wired the catalog entry against what actually worked.
- `prereqNote` calls out that this connector's tool list includes `exec_python_script` (arbitrary Python execution inside the running TouchDesigner instance) as a first-class tool — more exposed than Ableton's scoped bridge — matching the existing pattern of surfacing real setup/risk info inline rather than a bare "Error" (spec §6).

## Test plan
- [x] `npx tsc --noEmit` — no new errors (one pre-existing, unrelated `qrcode` module error in `HostPopover.tsx`)
- [x] Manually verified the underlying connector end-to-end (not just this catalog entry): TouchDesigner MCP router initialized with 12/12 routes, OpenAPI schema loaded, HTTP server live on 127.0.0.1:9981
- [ ] Armory catalog picker UI not smoke-tested in a running `task dev` session — the picker component (`McpCatalogPicker.tsx`) iterates `MCP_PRELOAD_CATALOG` generically with no per-entry logic, so this should render without further changes, but flagging since I didn't launch the app to confirm visually

### DIFF
--- a/.changesets/1784977373-feat-armory-add-touchdesigner-to-the-mcp-preload-catalog-zevc.md
+++ b/.changesets/1784977373-feat-armory-add-touchdesigner-to-the-mcp-preload-catalog-zevc.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): add TouchDesigner to the MCP preload catalog
--- a/frontend/app/view/mcp/McpCatalogPicker.tsx
+++ b/frontend/app/view/mcp/McpCatalogPicker.tsx
@@ -32,6 +32,7 @@ export function McpCatalogPicker(props: { model: McpCatalogModel }): JSX.Element
                             <button type="button" class="mcp-picker-entry" onClick={() => model.startFromCatalog(entry)}>
                                 <span class="mcp-picker-entry-name">{entry.name}</span>
                                 <span class="mcp-picker-entry-note">{entry.prereqNote}</span>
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

AgentY-asaf:
Fixed both in 196ab60d: pinned to `touchdesigner-mcp-server@1.5.0` (the version validated this session), and split the exec_python_script warning into a new `riskNote` field rendered persistently (both in the catalog picker and McpStatusPill), independent of connection status — unlike `prereqNote`, which is intentionally suppressed once connected.

PR #2299, opened 2026-07-26
feat(media): add Media pane widget with live directory-watch updates

Agent2-asaf:
## Summary

New widget-bar pane (`defwidget@media`) for viewing local image/video files — built for agent generation pipelines (e.g. ComfyUI/Comfy Cloud video rendering) that currently have no way to surface their output inside AgentMux. Point the pane at a directory and it shows the newest matching file, live-updating whenever a new render lands — no manual reload.

Design doc: `docs/specs/SPEC_MEDIA_PANE_2026_07_26.md` (includes a corrected path-scoping section — the initial draft assumed an enforced Editor-pane root allowlist that turned out not to actually exist in this codebase, per `editor_handlers.rs`'s own "root scoping, not a sandbox" comment).

## What's included

**Backend**
- Implements the previously-stubbed `GET /agentmux/stream-local-file` route (`agentmux-srv/src/server/files.rs`) — as a side effect, this also fixes existing dead code in `frontend/app/view/term/termsticker.tsx` that already assumed this endpoint worked (terminal image stickers were silently 501ing).
- New `MediaFileWatcher` (`agentmux-srv/src/backend/media_file_watcher.rs`), a directory-mode sibling of the existing `EditorFileWatcher` — watches a directory, filters by extension, debounces, and publishes a new `media:file_changed` WPS event scoped per-block.
- New `watchmediadir`/`unwatchmediadir` RPC handlers.

**Frontend**
- `MediaViewModel`/`MediaView` (`frontend/app/view/media/media.tsx`), registered as `defwidget@media`.
- Directory mode resolves the newest matching file via the existing `ListEditorDirCommand` (reused, not a new listing endpoint) and subscribes to `media:file_changed` for live updates, with an opacity crossfade on src-swap to avoid the "running→completed jerk" UI pitfall documented elsewhere in this repo.
- Single-file mode just displays the file directly (no watch — nothing to watch for on a fixed path in v1).

## Test plan

- [x] `task build:backend` (release) — clean
- [x] `cargo test -p agentmux-srv media_file_watcher` — 3/3 pass
- [x] `task build:frontend` — clean
- [x] `npx vitest run --no-file-parallelism` — 1912/1912 pass (0 regressions), plus 6 new tests for the pane's pure path helpers
- [ ] **Live visual smoke test not done** — this is a shared dev machine with other agents actively running; I didn't launch `task dev` to click through the actual pane in a window. Static verification (build/tests/JSON validity) is solid, but someone should open the Media pane, point it at a real directory, and confirm it renders/updates before merging.


a5af:
@codex review

Agent2-asaf:
Addressed all four review findings in ae70451d:

**ReAgent (P1) / Codex (P2) — `handle_stream_local_file` blocking sync read.** Switched from `std::fs::read` to `tokio::fs::File` + `tokio_util::io::ReaderStream` + `Body::from_stream` — no more blocking a shared Tokio worker or fully buffering up to 500MB in memory per request.

**ReAgent (P1) — no `onError` handler, pane goes silently blank on failed load.** Added `onError` on both `` and `<video>`. More importantly this overlaps with the Codex finding below — most real failures (auth, 404) are now caught at the `fetch()` layer with an actual error message, `onError` is the remaining defense-in-depth for a valid-but-corrupt/undecodable blob.

**Codex (P1) — ``/`<video src>` can't carry the required `X-AuthKey` header.** Real bug, would have 401'd every request in practice (`stream-local-file` is in `authed_routes`, and the query-string `?authkey=` fallback is deliberately `/ws`-only per the 2026-05-11 security audit). Fixed by fetching the bytes with the header (same pattern as `fetchWaveFile` in `wave-file.ts`) and rendering a `URL.createObjectURL(blob)` instead of the raw endpoint URL. Also explains why the pre-existing `termsticker.tsx` dead code calling this same endpoint was dead — identical bug, never exercised.

**Codex (P2) x2 — stale empty-directory error not cleared on new file, and same-path overwrite not re-fetched.** Both real: the WPS handler wasn't clearing `errorMsg`, and setting a Solid signal to its current value is a no-op so an in-place file overwrite never re-triggered anything. Fixed with an explicit `revision` counter bumped on every change event, independent of whether the path string itself changed.

Verified: `cargo test -p agentmux-srv` (1660/1660), `npx vitest run` (1918/1918), `npx tsc --noEmit`, and full backend release build — all clean, no regressions.

`docs/specs/SPEC_MEDIA_PANE_2026_07_26.md` updated with a "Post-implementation corrections" section recording what the original design got wrong, for future reference.

Agent2-asaf:
@codex review

Agent2-asaf:
Follow-up work since the last review pass, driven by live testing in a running `task dev` instance:

**UX changes (per live feedback):**
- Replaced the path text-input with a native "open file" dialog (new `rfd` dependency in `agentmux-cef`, `show_open_file_dialog` IPC command), filtered to the pane's supported types.
- Clicking anywhere in the pane's background (not just a text label) opens the picker; added explanatory copy about supported types and live-update behavior.
- Unified file/directory handling: picking any file now always watches its containing directory, replacing the earlier separate `ListEditorDirCommand`-based directory-detection path (simpler, same live-update result).

**New format support:**
- Added WAV/audio (`<audio controls>`, `audio/wav` MIME, dialog filter). No codec concerns — PCM WAV is natively supported by Chromium unconditionally.
- Deliberately did **not** add MKV — Chromium's `<video>` element doesn't reliably accept the Matroska container for direct playback regardless of codec support (documented in the spec).

**Real finding while testing MP4 playback:** improved the `onError` handlers to surface the actual `MediaError` code/message instead of a generic string, and used that to confirm a real, separate, whole-app issue: `PipelineStatus::DEMUXER_ERROR_NO_SUPPORTED_STREAMS` on a standard H.264/AAC MP4. Checked the entire repo for `proprietary_codecs`/`ffmpeg_branding` build flags — zero matches anywhere (dev, CI, or release), meaning this app's CEF binary lacks H.264/AAC support **everywhere**, not just in this pane. Written up in `docs/reports/REPORT_CEF_PROPRIETARY_CODEC_GAP_2026_07_26.md` with root cause and options (rebuilding the `agentmuxai/cef` fork with proprietary codecs is the last-resort option per direction from the repo maintainer, not the default path). Not fixed here — this pane's own code is correct; MP4/MOV would just start working if the underlying CEF build ever changes. WebM (VP9) and WAV are unaffected and confirmed working.

Re-verified after all of the above: `cargo check` on both `agentmux-cef` and `agentmux-srv` clean, `npx tsc --noEmit` clean, `npx vitest run` green (media suite + no regressions elsewhere).

Agent2-asaf:
Addressed the open P2: \`defwidget@media\` now has \`"display:pinned": true\`, matching the CLAUDE.md widget table already added in this PR (was incorrectly \`false\`, so the widget only surfaced in "More").

Also live-verified end-to-end: the Media pane now plays a previously-failing MP4 with proprietary codec support, once tested against the new codec-enabled Windows CEF build from #2308.

PR #2300, opened 2026-07-26
fix(auth): deterministic Claude login UX — skip doomed URL-capture wait

GenericAgentX-asaf:
## Summary
- Claude's CLI never prints a scrapeable OAuth URL (v2.1.x moved to a clipboard-copy-on-keypress flow), so every mount-time/relogin/`/login` attempt burned `cli_login.rs`'s ~15s PTY URL-capture wait on a guaranteed dead end — not a rare timeout, every single time.
- Added `headlessLoginUrlUnsupported` to the provider catalog and wired `skipTier1` at the three auto-login call sites (`launch-flow.ts`, `useAgentControllerStatus.ts`'s relogin, `commands/global/login.ts`) so Claude skips straight past tier 1. No Rust changes needed for the fix itself — `runProviderLogin`'s existing `skipTier1` param already did exactly this, just wasn't wired to the automatic paths.
- Shortened the Rust-side `URL_CAPTURE_TIMEOUT_SECS` from 15s → 5s as a safety margin for any future/direct caller (Codex/Gemini/OpenClaw still get the real bounded wait).
- Added a `LaunchPhase` status (new `flows/launch-phase.ts`, mirrors `auth-state.ts`'s `AuthState.kind` pattern) so the footer shows what's actually happening ("Resolving CLI", "Waiting for login link… up to 12s", "Waiting for you to finish logging in… up to 4m 50s") instead of a generic "Working…" for every phase — every remaining real wait is now labeled with its deadline, never silent.
- Added a cancel-login button (`AuthUrlBox` + `AgentWorkingRow`) wired to the existing `cancelLogin()`, which previously only had a button on the separate pre-launch modal, not this mount-time flow.

## Test plan
- [x] `tsc --noEmit` and `cargo check` clean
- [x] New `launch-flow.test.ts` (3 tests) pins that Claude gets `skipTier1: true`, Codex doesn't, and every launch phase is reported including a terminal `ready`
- [x] Existing `run-provider-login.test.ts` (20 tests) and `run-cli-login-single-caller.test.ts` pass unmodified
- [x] New `cli_login.rs` unit test pins `URL_CAPTURE_TIMEOUT_SECS` as a short safety margin, not the old 15s
- [x] Full frontend suite (`npx vitest run`) — 724 passed
- [x] `cargo test -p agentmux-cef --lib cli_login::` — 5 passed
- [x] Full release-mode dev build compiled and booted cleanly (verified twice)
- [ ] Manual click-through in a running dev instance — blocked by an unrelated, pre-existing dev-server crash bug (documented separately, not caused by this change)

### DIFF
--- a/.changesets/1785050454-fix-auth-skip-doomed-headless-url-capture-wait-for-claude-lo-fwcn.md
+++ b/.changesets/1785050454-fix-auth-skip-doomed-headless-url-capture-wait-for-claude-lo-fwcn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(auth): skip doomed headless URL-capture wait for Claude login, surface phase status + cancel
--- a/agentmux-cef/src/commands/cli_login.rs
+++ b/agentmux-cef/src/commands/cli_login.rs
@@ -63,6 +63,16 @@ impl CliLoginStdin {
 /// pane was closed without its cleanup firing).
 const LOGIN_REAP_TIMEOUT_SECS: u64 = 6 * 60;
 
+/// Cap on how long run_cli_login_pty blocks waiting for a scrapeable OAuth
+/// URL before giving up and returning auth_url=None. Frontend callers that
+/// know in advance a provider never prints one (Claude — see catalog.ts's
+/// headlessLoginUrlUnsupported flag) skip this call entirely via
+/// runProviderLogin's skipTier1, so in practice this only bounds providers
+/// that plausibly DO print a URL (Codex/Gemini/OpenClaw). Kept short as a
+/// safety margin for any caller that reaches this function without going
+/// through that gate.
+const URL_CAPTURE_TIMEOUT_SECS: u64 = 5;
+
 /// Spawn a CLI auth login flow.
 pub async fn run_cli_login(
     state: Arc<AppState>,
@@ -620,7 +630,8 @@ async fn run_cli_login_pty(
 
     // Synchronously read from the master in a blocking task, scanning
     // each line for an OAuth URL. portable_pty's reader is sync.
-    // The 15 s cap is enforced async-side via tokio::time::timeout —
+    // The URL_CAPTURE_TIMEOUT_SECS cap is enforced async-side via
+    // tokio::time::timeout —
     // BufRead::read_line itself blocks indefinitely without per-read
     // timeout support, so a child that pauses before its first line
     // (or sits at a prompt with no newline) would wedge `url_rx.await`
@@ -694,7 +705,7 @@ async fn run_cli_login_pty(
     });
 
     let auth_url: Option<String> = match tokio::time::timeout(
-        std::time::Duration::from_secs(15),
+        std::time::Duration::from_secs(URL_CAPTURE_TIMEOUT_SECS),
         url_rx,
     )
     .await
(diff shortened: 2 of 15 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed all three P1 findings:

1. **Visibility gate**: `AgentWorkingRow`'s `launchPhase` label + Cancel button (and the top progress bar, and the composer status strip) were gated only on `status.isLoading()`, which never becomes true during `relogin()`/`loginViaTerminal()`/`useGlobalLogin()`. Added a shared `showingLaunchActivity()` that also checks `status.launchPhase() != null`, used at all three call sites.
2. **Stale phase during runProviderLogin's black-box tiers**: added `onTierChange` to `RunProviderLoginParams` — fires `{tier: "fallback"}` when tier 1 gives up, and `{tier: "polling", deadlineMs}` (a live deadline) once a terminal actually opens and polling begins. Wired at all three call sites (`launch-flow.ts`, `relogin()`, `loginViaTerminal()`). Added regression tests in both `launch-flow.test.ts` and `run-provider-login.test.ts` pinning the fix.
3. **URL_CAPTURE_TIMEOUT_SECS**: reverted to 15s — Claude no longer reaches that function at all, so shortening it only risked breaking OpenClaw's legitimate up-to-15s URL print time for no benefit. Updated the regression test accordingly. Also deduplicated the frontend's matching `LOGIN_LINK_CAPTURE_LABEL_MS` constant into `launch-phase.ts` (was drifting as a second copy).

Full suite green (724 tests) plus 3 new regression tests for these fixes specifically.

GenericAgentX-asaf:
Addressed the P2 from the 18:19 re-review, pushed in 873c617d:

**`useAgentControllerStatus.ts:616`** — the comment claimed `onTierChange`'s "fallback" event never fires when `skipTier1` is true. Confirmed that's wrong: `run-provider-login.ts:288-307` only gates the tier-1 *attempt* itself behind `if (!p.skipTier1)` — the code right after it (cancel tier-1's child, then `p.onTierChange?.({ tier: "fallback" })`) runs unconditionally regardless of `skipTier1`. Corrected the comment to state that accurately and explain why the handler still only needs to react to "polling" here (this flow never displays anything for tier 1 in the first place, so there's nothing to update on "fallback").

The 3 P1s from the earlier review round (visibility gate, stale phase label, 5s timeout breaking OpenClaw) were already fixed in 18:16:00's commit — confirmed via the 14:17→18:19 review sequence, where the 18:19 re-review found no P1s, only this one P2.

Typecheck and the `launch-flow.test.ts` + `run-provider-login.test.ts` suites (26 tests) pass.

GenericAgentX-asaf:
Fixed both P2s from the 18:53 re-review, pushed in c8d49f4d:

**Double Cancel button** — `AgentFooter.tsx:76`'s `CANCELLABLE_LAUNCH_PHASES` includes `waiting-for-login-completion`, which `launch-flow.ts` also sets for tier 1's "opened" outcome (a URL was captured) — so `AuthUrlBox`'s own Cancel button and the working-row's Cancel button rendered simultaneously, contradicting the row button's own doc comment. Added a `hasAuthUrl` prop (from `status.authUrl()`) that suppresses the row's button whenever `AuthUrlBox` already has one.

**CLI-resolve mislabeled** — `useAgentControllerStatus.ts`'s `relogin()` (line 323) and `loginViaTerminal()` (line 598) both set their "real" phase label (`checking-auth` / `opening-login-terminal`) *before* `resolveCliForRecovery`, which can itself take up to 5 minutes on a fresh install — so that whole wait showed the wrong label. Both now show `resolving-cli` during the actual resolve, then switch to the correct label once it completes, matching `launch-flow.ts`'s own phase for the same step.

Typecheck and the `launch-flow.test.ts` + `run-provider-login.test.ts` suites (26 tests) pass.

PR #2301, opened 2026-07-26
fix(diag): restore WER crash-dump capture for agentmux-srv, document 9-min crash investigation

GenericAgentX-asaf:
## Summary
- Found while manually verifying #2300: dev instances reliably die ~9m36-39s after startup, fully idle, no user activity. Reproduced 6 times total (2 in this investigation, 4 in a prior session).
- Traced the crash-diagnostics gap: WER `LocalDumps` was registered for the pre-rename binary name `agentmuxsrv-rs.exe` (set up in `retro-recurring-sidecar-crash-0xC0000409.md`, March 2026) — the actual binary today is `agentmux-srv-{version}-windows.x64.exe`, so **no crash of the real binary has produced a diagnosable minidump since that rename**, and nobody noticed.
- Added `scripts/register-crash-dumps.ps1` to fix the registration for the current version. It's idempotent, self-elevates via UAC, and reads the version from `package.json`. **Limitation documented in the script and retro:** WER matches by exact filename, so this needs re-running on every version bump — no code-level fix attempted (a `SetUnhandledExceptionFilter`-based in-process handler would be more durable but is out of scope here).
- Root cause of the actual 9-minute crash is **not found**. The retro (`docs/retro/retro-agentmux-srv-9min-crash-2026-07-26.md`) documents everything ruled out with hard evidence (no timer in the workspace explains the timing, the crash isn't a genuine Windows fault/exception — zero event log entries and zero dumps even with WER correctly configured, not srv's own graceful shutdown path, not any known internal kill site) and the next diagnostic step for whoever picks it up (Process Monitor tracing the exact PID through one more reproduction).

## Test plan
- [x] Ran `scripts/register-crash-dumps.ps1` — confirmed it registers the correct current-version binary name in the registry
- [x] Verified idempotency (re-running produces the same key)
- N/A — the retro doc is a written report of an unresolved investigation, nothing to test beyond the script itself

### DIFF
--- a/.changesets/1785098338-fix-diag-restore-wer-crash-dump-registration-for-agentmux-sr-dm9o.md
+++ b/.changesets/1785098338-fix-diag-restore-wer-crash-dump-registration-for-agentmux-sr-dm9o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(diag): restore WER crash-dump registration for agentmux-srv, correcting a stale binary filename
--- a/docs/retro/retro-agentmux-srv-9min-crash-2026-07-26.md
+++ b/docs/retro/retro-agentmux-srv-9min-crash-2026-07-26.md
@@ -1,0 +1,224 @@
+# Retro: `agentmux-srv` Dies ~9m36s-9m40s Into Every Idle Dev Instance
+
+**Date:** 2026-07-26
+**Affected versions:** 0.54.4 (dev build, branch `agenta/deterministic-login-ux`); likely others — nothing found is version-specific
+**Severity:** High — blocks unattended dev-instance use past ~9.5 minutes; unresolved
+**Status:** Root cause NOT found. Multiple plausible mechanisms ruled out with hard evidence. Next diagnostic step identified (Process Monitor), not yet run.
+
+---
+
+## How this was found
+
+Discovered as a side effect while manually verifying an unrelated fix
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Added the missing changeset, pushed in d2ba1cbf — `task changeset -- patch "fix(diag): restore WER crash-dump registration for agentmux-srv, correcting a stale binary filename"`, per RFC #857 Phase 2.

AgentA-asaf:
Fixed both P2s, pushed in 8931bac6:

**Unquoted elevation argv** — `$argList` elements went to `Start-Process -ArgumentList` unquoted; a space in `$PSCommandPath` or `$DumpFolder` (both realistic — repo paths and a user-supplied dump folder) would split into extra, garbled arguments during the UAC relaunch. Wrapped both in explicit quotes.

**Masked elevated-child exit code** — `exit $LASTEXITCODE` after `Start-Process ... -Wait` (no `-PassThru`) doesn't reflect the elevated child's actual outcome — `Start-Process` returns nothing without `-PassThru`, so a real failure in the elevated instance (e.g. the `throw` on a missing `package.json` version) silently read as success. Added `-PassThru` and now exit on the returned process object's `ExitCode`.

Verified the script still parses cleanly (`Parser::ParseFile`, no syntax errors).

PR #2302, opened 2026-07-26
fix(window): restore exact secondary-window geometry after a full app restart

AgentX-asaf:
## Summary

Investigated during a user question about close/reopen persistence: tabs/panes/layout and agent conversation history already survive a full app restart well, but secondary windows land at an approximate default position instead of their exact last size/location on the slow-path reproject (a full process-tree restart, launcher included — not just the main window closing). The fast path (launcher still alive) already had this from its live in-memory `WindowMirror.last_rect`; the slow path (`reproject_from_srv`) never had a rect source at all.

`Window.pos`/`winsize` already existed as srv fields (`agentmux-srv/src/backend/obj.rs`) and the `SetWindowPosAndSize` RPC already existed to write them (`agentmux-srv/src/server/service/window_mutate.rs`) — both unused until now.

- New `commands/window/position_persist.rs`: debounced (1.5s trailing-edge, generation-counter keyed by label) srv write-through, mirroring `transparency.rs`'s proven opacity-write-through pattern exactly. Hooked into `wrr::win_event`'s existing ~20Hz position stream (`EVENT_OBJECT_LOCATIONCHANGE`), alongside (not replacing) the existing launcher-IPC forward.
- `client/helpers.rs`: `backend_set_window_pos_and_size` (write) and `backend_get_window_pos_and_size` (read), same raw-TCP/blocking shape as the existing opacity helpers.
- `commands/window/creation.rs::reproject_from_srv`: was hardcoding `last_rect: None` for every recreated window; now fetches the persisted rect per window and passes it through to `open_window_with_kind`'s existing `explicit_rect` parameter — the exact path the fast path already uses via `WindowMirror.last_rect`.

**Scope note:** floating-pane placement restore (a pane that was torn off into its own window when the app last closed) turned out to be a separate, larger gap on investigation — floating panes aren't recreated on app restart at all today (they have no srv `Window` row and aren't part of the `Client.windowids` reproject sweep), so there's no existing call site to wire a geometry fix into. Out of scope here; flagging separately rather than silently dropping it or forcing an unrelated call site to serve double duty.

## Test plan

- [x] `cargo check -p agentmux-cef -p agentmux-srv` — clean, no new warnings
- [x] `cargo test -p agentmux-cef position_debounce_tests` — 3/3 new tests pass (burst-collapses-to-last-write, single-report-is-current, per-label-independence — mirroring `transparency.rs`'s existing opacity-debounce test suite)
- [ ] Manual: open a secondary window, drag it to a distinctive position, fully quit the app (kill the whole process tree via Task Manager, not just the main window), relaunch, confirm the secondary window reappears at its last position/size instead of a default one

### DIFF
--- a/.changesets/1785052913-fix-window-restore-exact-secondary-window-geometry-after-a-f-numt.md
+++ b/.changesets/1785052913-fix-window-restore-exact-secondary-window-geometry-after-a-f-numt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(window): restore exact secondary-window geometry after a full app restart
--- a/agentmux-cef/src/client/helpers.rs
+++ b/agentmux-cef/src/client/helpers.rs
@@ -248,6 +248,88 @@ pub(crate) fn backend_set_window_opacity(web_endpoint: &str, auth_key: &str, win
     }
 }
 
+/// Write-through a window's position/size to srv's `Window.pos`/`winsize`
+/// mirror (via `backend_get_window_pos_and_size` below) so a full
+/// process-tree restart — where the launcher's live in-memory
+/// `WindowMirror.last_rect` is also gone, not just the main window — can
+/// still recreate secondary windows at their exact last geometry instead of
+/// a default placement. Fire-and-forget, same shape as
+/// `backend_set_window_opacity`: raw TCP, always called from its own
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2303, opened 2026-07-26
docs(spec): agent pane mount/auth notifications + LaunchAuthState reducer

GenericAgentX-asaf:
## Summary
- Manually testing PR #2300 surfaced that reopening/resuming an agent pane is fully silent on the happy path, and can pop open a login (browser/terminal) with zero warning if the cached credential has gone stale — the mount flow treats a fresh agent and a resumed one identically.
- This spec proposes `LaunchAuthState`, a single reducer subsuming PR #2300's `LaunchPhase` plus the scattered `needsLogin`/`loginWaiting` booleans, `onLoginSuccess` callback, and Phase 3's init/done branch — with an explicit notification behavior for every state, so "does every terminal state tell the user something" is answerable from one table (§6) instead of re-deriving it from five files.
- Key decisions already resolved in §8 (not left open): notify-then-proceed (not block) on an expired-token relogin; fold the "already authenticated" resume into a single end-of-flow line rather than two; named `LaunchAuthState` specifically to avoid colliding with the unrelated `CredentialState` broker in `agentmux-srv/src/broker/state.rs` (different system — see §7).
- Explicitly does NOT touch the pre-launch modal's `AuthState` reducer or the MuxBus `CredentialState` broker — see §3/§7 for why they stay separate.

Docs-only — no code changes in this PR. Implementation will follow as a separate PR against this spec's §9 acceptance criteria.


a5af:
@codex review

GenericAgentX-asaf:
Addressed the P1 and the fixable P2s:

- **P1 (shellprocstatus "running" missing)**: confirmed real via `persistent.rs`'s `STATUS_RUNNING` and `useControllerStatusEvents.test.ts`. §5/§6 updated — a persistent controller resumed while still alive/mid-turn is now treated as a resume (same as "done"), not silently dropped. **Also fixed the actual implementation gap this pointed at in PR #2304** (the spec's own reference implementation), not just the doc.
- **P2 (stale line citations)**: replaced exact line numbers — which had already drifted once and will keep drifting across this stacked PR set — with landmark references to the actual condition/function text, plus an explicit note that line numbers are approximate as of PR #2300's current revision.
- **P2 (retro doc "doesn't exist")**: it exists on PR #2301's branch (the sibling PR this doc cites), not yet merged to `main` at review time — not a broken reference, just not-yet-true. Clarified this in §10.
- **P2 (PR #2300 infra not yet landed)**: acknowledged as accurate for a standalone read of `main` — this spec is explicitly written to build on PR #2300 (stated in the header's "Depends on" list), so `LaunchPhase`/the cancel button not existing on `main` yet is expected, not an error.

GenericAgentX-asaf:
Addressed the recurring line-citation issue, pushed in 6468205:

The 3 stale line-number citations (`agent-view.tsx`'s onMount call, the `if (needsLogin)` block start, Phase 3's init/done check) kept drifting because they cite code on PR #2300's branch, which is still under active review and gains a new commit (shifting those lines) every review round — the earlier "treat as approximate" caveat next to a specific number wasn't enough to stop it from being flagged as wrong each time. Removed the literal numbers entirely; the landmark text (function names, condition text like `if (needsLogin) {...}`) was already the actual stable reference, so nothing about the spec's substance changes. Also updated the Phase 3 landmark to `status === "done" || status === "running"`, matching the fix already reflected in §5/§6's tables.

The other two items from your last round — the "running" status gap in the ready-state table and the retro-doc-not-yet-merged clarification — were already addressed in an earlier commit (§5/§6 cover "running"; §Depends-on / the retro citation at the bottom note it lands via PR #2301, not yet merged to main). Re-checked both are still present on this branch.

Note PR #2300 continues to iterate (I just pushed another commit there addressing a stale-comment P2) — if it lands more substantive changes to the mount/Phase-2/Phase-3 shape before merge, I'll re-verify this spec's landmarks still match.

GenericAgentX-asaf:
Fixed the last P2, pushed in 428a1800: §10's References note still said line numbers were "approximate as of PR #2300's current revision," directly contradicting §1's already-corrected statement that this doc cites no line numbers at all. Made both say the same thing.

PR #2305, opened 2026-07-26
docs(report): bashwrap is a seventh, uncoordinated process-liveness mechanism

GenericAgentX-asaf:
## Summary
- Assesses whether `agentmux-bashwrap` is coupled to the long-running-process/reducer/dock/Swarm architecture, per a live user request after `task dev` was silently killed by bashwrap's idle-kill mid manual-testing today.
- Finding: the opposite of the assumption — bashwrap has **zero** coupling to any of it, despite being the highest-frequency shell-execution path in the system (every native Bash tool call). Traces this to `hook.rs`'s `PreToolUseInput` struct, which never captures `run_in_background` from Claude's own Bash-tool schema.
- Confirms (via grep + direct source reads) that `agentmux-bashwrap` and `shell_node.rs` (behind AgentMux's own `Shell` MCP tool) are two fully independent execution surfaces with diverging bug histories — the same console-window-flash bug was found and fixed twice, once per surface.
- Cross-references `docs/retro/retro-persistent-agent-working-status-stuck-2026-07-16.md` (the Agent1 12h-stuck incident) as the same root cause, independently reproduced ten days earlier via a different downstream symptom (stuck status vs. today's silent kill).
- Separately diagnoses the "docked long-running processes don't act deterministically" half of the brief as a **different, already-well-documented** issue (two disagreeing liveness clocks, too few modeled states) — not caused by bashwrap, addressed by the existing `REPORT_PROCESS_ARCHITECTURE_STATE_AND_RETHINK_2026_07_22.md` / `REPORT_LONGRUNNING_*` reports.
- Report-only — no code changes. Recommendations and open questions are listed for follow-up.

## Test plan
- [x] Every structural claim (missing `run_in_background` field, zero coupling to the six tracking mechanisms, no timeout logic in `shell_node.rs`) verified via direct source reads, not inferred from prior docs alone.

### DIFF
--- a/docs/specs/REPORT_BASHWRAP_LONGRUNNING_PROCESS_DETERMINISM_2026_07_26.md
+++ b/docs/specs/REPORT_BASHWRAP_LONGRUNNING_PROCESS_DETERMINISM_2026_07_26.md
@@ -1,0 +1,378 @@
+# Bashwrap, the Dock, and the Process Broker — a Seventh Mechanism Nobody Wired Up
+
+**Date:** 2026-07-26
+**Author:** AgentA
+**Status:** Report — analysis, grounded in two fresh, personally-reproduced incidents from today plus this
+codebase's own prior-art trail. No implementation in this PR.
+**Ground truth basis:** `agentmuxai/agentmux` `main` at commit `6962f81f`, pulled fresh for this report.
+**Related (read in full while preparing this report, not just cited secondhand):**
+- [`REPORT_PROCESS_ARCHITECTURE_STATE_AND_RETHINK_2026_07_22.md`](REPORT_PROCESS_ARCHITECTURE_STATE_AND_RETHINK_2026_07_22.md)
+  — the six existing process-liveness mechanisms + the proposed Process Broker. **This report's central claim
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentA-asaf:
Both P2s fixed, pushed in 1419177:

**Incident B's diagnosis was wrong, as flagged** — verified against source (`shell_node.rs:313-321`): `ShellNodeRunner::run` spawns `cmd /C <raw string>` directly on Windows, no Git Bash or `sh` anywhere in that path. There was never a ".cmd auto-dispatch forwarding the raw string" layer to blame — the actual mechanism is simpler: `cmd.exe` has no concept of `;`/`$?`, so it passed my POSIX-shell syntax straight through as literal argv tokens. Corrected the section and left an inline note flagging the correction (rather than silently rewriting history).

**Agent1 retro citation softened** — that retro genuinely hedges its own root-cause attribution ("almost certainly", "wasn't confirmed without a live repro"), and my report presented it as confirmed. Reframed throughout (§0 executive summary, §3.3's heading and body) as corroborating-but-unconfirmed prior evidence rather than an established match — this report's own Incident A remains the only *directly observed* instance of the two.

PR #2306, opened 2026-07-26
docs(report): reducer stack audit — same shape, zero shared code

GenericAgentX-asaf:
## Summary
- Post-audit of the entire reducer stack per user request: "clear out tangled spaghetti code, DRY it out, modularize where appropriate."
- 6 parallel research passes covering: `agentmux-srv`'s Phase E reducer, `agentmux-launcher`'s reducer + saga coordinator, `agentmux-cef`'s own reducer, the frontend pane-state reducer family, the full auth/credential state-machine cluster (frontend + backend), and a check on the subagent-lifecycle "no reducer" claim from an earlier report.
- **Central finding**: 3 independent Rust reducers (srv/launcher/cef) + a saga coordinator + a `UiThreadGate` that are reducers in all but name, plus 5+ frontend reducers, all correctly repeat the same `update(state, command) -> events` convention — with zero shared code behind it. Concretely: ~70 copy-pasted rejection-arm blocks across the Rust trio, a triplicated poll loop in the auth flow, and two structurally near-identical backend session managers (`AuthSessionManager`/`OAuthSessionManager`).
- **Decisively answers a standing question**: "why isn't `AuthState` + `CredentialState` one reducer" — correctly separate (different lifecycles / process boundaries), but two specific accidental-duplicate pairs inside that separation should merge.
- Deliberately complements, not duplicates, `docs/reports/REPORT_LARGE_FILE_MODULARIZATION_SCAN_2026_07_22.md` (already answered "should this file be split" for 84 files) — this report only asks "is this logic duplicated across separate implementations."
- Prioritized recommendations (P0/P1/P2) and an explicit "what's already right, don't touch" section so nothing here reads as "everything is spaghetti."
- Report only — no code changes.

## Test plan
- [x] Every structural claim (file:line citations, duplication counts, cross-references to existing reports) verified by direct source reads in the 6 research passes, not inferred from documentation alone.
- [x] Cross-reference links to sibling reports verified to resolve to real files.

### DIFF
--- a/docs/specs/REPORT_REDUCER_STACK_AUDIT_2026_07_26.md
+++ b/docs/specs/REPORT_REDUCER_STACK_AUDIT_2026_07_26.md
@@ -1,0 +1,431 @@
+# Reducer Stack Audit — Post-Mortem on Duplication, Coupling, and Modularization
+
+**Date:** 2026-07-26
+**Author:** AgentA
+**Status:** Report — analysis only, no code changes. Synthesizes 6 parallel research passes.
+**Ground truth basis:** `agentmuxai/agentmux` `main` at commit `35af4958`, pulled fresh for this report.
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2309, opened 2026-07-27
feat(agent): promote long-running Bash tool calls to the activity dock

AgentY-asaf:
## Summary
- Any Bash tool call still running past 30s promotes to a live `PinnedActivity` row in the existing agent-pane `ActivityDock`, via a new `activity/tool-adapter.ts` mirroring `shell-adapter.ts`. Classification is duration-first (not text-pattern matching), so it generalizes to any long-running command — `sleep`, dev-server launches, long builds — with no regex/false-positive surface.
- Promotion timing follows the dock's existing "no continuous tick" discipline (see `ActivityDock.tsx`'s `hasExpiring`): a one-shot `setTimeout` scheduled for the exact instant a running Bash call crosses the threshold, not a per-second recompute.
- `AgentWorkingRow` suppresses its own "tool · arg" text once the pane's current tool call has a live dock entry, falling back to the calm, cycling "Working…" phrase instead of repeating the same info the dock now shows — the turn-phase state machine (`workingFromPhase`) is untouched, this is rendering-only.
- Row expand/collapse, retention-window auto-dismiss, and log-tail rendering are inherited for free from the existing dock/`ActivityRow` machinery — `ActivityRow.tsx` was extended to source its expanded log view from either `shell.log` or `tool.log`.

Implements step 1 of `docs/specs/REPORT_LONGRUNNING_TOOLCALL_AUTODETECT_STATUS_2026_07_26.md` §3 (duration-threshold promotion), including its §4.2 (duration-over-regex classification) and §4.3 (dock-takes-over suppression) resolutions.

## Test plan
- [x] `npx vitest run app/view/agent` — 741 passed, including new `tool-adapter.test.ts` coverage (promotion threshold boundary, non-Bash/finished-call exclusion, one-shot-timer scheduling helper)
- [ ] Manual: run a Bash `sleep 60` in a live agent pane, confirm it docks at ~30s and the Working banner goes calm
- [ ] Manual: confirm the dock row's expand/collapse and auto-dismiss-on-completion behave the same as an existing shell row


a5af:
@codex review

PR #2310, opened 2026-07-27
feat(macos): tab redock via CGEventTap — cross-window tab drag now merges directly

AgentO-asaf:
## Summary

Ports Windows' `HookMode::TabDrag` cross-window tab redock to macOS via a `CGEventTap`-based global hook — dragging a tab from one AgentMux window's strip and dropping it onto another window's strip now merges it directly (with a live cursor-tracked insertion indicator), matching Windows' existing behavior. Previously this only worked on Windows; macOS fell back to an append-only, indicator-less drop.

Full design writeup: `docs/specs/SPEC_MACOS_TAB_REDOCK_PARITY_2026_07_24.md`

**Scope correction made along the way:** the original plan included porting Windows' `SC_MOVE` live-follow tear-off window-move. That turned out to be dead code on *every* platform today (Windows itself moved to a commit-on-release tear-off model in its most recent tab-drag work) — so it was dropped from scope. Only the in-strip cross-window merge gesture was built, which is the literal ask and the one thing still genuinely missing on macOS.

## What's in this PR

- `agentmux-cef/src/commands/tear_off_hook.rs` — macOS `CGEventTap` hook: Accessibility permission gate (silent check → one-time OS prompt if ungranted), point-in-rect cross-window hit-test via `CGWindowListCopyWindowInfo`, `tearoff:hover-changed`/`tearoff:hover-cleared`/`tabdrag:merge-direct` emission — identical event names/payloads to Windows, so the frontend needed almost no changes.
- `agentmux-cef/src/app/mod.rs` — windowNumber→label cache, populated on the CEF UI thread (the hook thread never touches AppKit/CEF Views directly — see the code comments referencing `docs/investigations/tab-drag-tearoff-crash-macos.md` for the exact crash class this avoids).
- `agentmux-cef/Cargo.toml` — `core-foundation`/`core-graphics`, scoped to macOS, for the CGEventTap/CFRunLoop/CGWindowList plumbing.
- `frontend/app/tab/tab-tearoff-events.ts` — DPR-conversion fix (CGEventTap reports coordinates in points, not physical pixels like Windows) + new `tabdrag:escape-pressed` listener.
- `frontend/app/tab/tab-reorder.ts`, `tabbar-dnd.ts`, `droppable-tab.tsx` — Escape-to-abort: the native HTML5 drag session doesn't cancel itself on Escape, and (confirmed via live testing) a page-level `keydown` listener doesn't fire during an active native drag on macOS either. Fixed by having the CGEventTap hook — which sees the raw OS-level keystroke outside the renderer's event pipeline — signal the frontend via IPC event instead.

## Fixes found via live testing (not just static review)

1. **Escape didn't actually abort a pending tear-off** — fixed as above.
2. **Hot-path performance regression** — the hit-test called `CGWindowListCopyWindowInfo` (a full system-wide window enumeration) on every single mouse-move event, up to ~120 Hz. Unlike Windows' `WindowFromPoint` (an O(1) OS-maintained lookup), this is not free. Throttled to ~30 Hz for the hover-indicator hot path; the mouseup finalize decision stays uncached for correctness.
3. **Accessibility permission was silently checked but never requested** — with no way to grant it, the hook just never installed. Added a minimal one-prompt-per-process-lifetime request path.

## Known follow-ups (not blocking, tracked in the spec)

- The landing-bounce/strobe animation (`tab-drop-invert-strobe`) doesn't play consistently on merge — confirmed a pre-existing issue on Windows too, not something this PR introduced or needs to fix.
- Phase 7c (in-app Accessibility-permission explanation UI, "already asked" persistence across launches, settings deep-link) and Phase 7d (telemetry, stress testing) are follow-up work, not in this PR.

## Test plan

- [x] `cargo build -p agentmux-cef` — clean, frameworks link correctly.
- [x] `cargo test -p agentmux-cef` — 184/184 passing.
- [x] Live-tested end-to-end via CDP + manual testing: in-strip cross-window drag merges directly with a live insertion indicator; Escape aborts cleanly (no reorder, no tear-off); Accessibility permission prompt appears and the hook installs once granted.

### DIFF
--- a/.changesets/1785003182-feat-macos-tab-redock-via-cgeventtap-cross-window-tab-drag-n-ny0j.md
+++ b/.changesets/1785003182-feat-macos-tab-redock-via-cgeventtap-cross-window-tab-drag-n-ny0j.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(macos): tab redock via CGEventTap — cross-window tab drag now merges directly with a live insertion indicator, matching Windows
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,6 +47,8 @@ dependencies = [
  "axum 0.8.9",
  "cef",
  "chrono",
+ "core-foundation",
+ "core-graphics",
  "dirs",
  "futures-util",
  "json_comments",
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #2313, opened 2026-07-27
docs(report): cross-branch agent-definition DB gap breaks auth for pre-existing agents

AgentA-asaf:
## Summary
- Investigates 2 auth symptoms from manual end-to-end testing on a fresh `main`-branch dev instance (browser opens immediately with no click; successful login shows no notification), reproduced with two different pre-existing agents (Oozp, then Parko).
- **Root cause**: `task dev` keeps a separate SQLite database per git branch. An agent used successfully on one branch's database has no `db_agent_definitions`/`db_agents` row in another branch's database — even though its on-disk config folder is shared/global. Every downstream DB write that needs that row (`createagentinstance`, `linkagentidentity`, `persistSeededAccount`) fails on the same foreign key.
- This explains the missing notification as a **correct response to genuinely-failed account persistence**, not a regression in this session's `LaunchAuthState`/notification work — the credential seed itself succeeds; only the DB-backed confirmation of it doesn't.
- Confirmed systemic (both agents hit the identical two-error sequence), confirmed via exhaustive search across 100+ database files on this machine that neither agent's row exists anywhere, and confirmed via a forced `PRAGMA wal_checkpoint(FULL)` that this isn't a read-visibility artifact.
- One open question flagged explicitly rather than guessed at: how the UI still lists/launches an agent whose definition row doesn't exist in the connected database — stale frontend cache is the leading hypothesis, not yet confirmed.
- Report only — no code changes.

## Test plan
- [x] Every "does not exist" claim verified by direct `sqlite3` query (including a forced WAL checkpoint), not inferred.
- [x] Failure-chain claims cross-checked against both the live host log evidence and the current `run-provider-login.ts`/`launch-flow.ts` source.

### DIFF
--- a/docs/specs/REPORT_AGENT_DEFINITION_DB_GAP_2026_07_27.md
+++ b/docs/specs/REPORT_AGENT_DEFINITION_DB_GAP_2026_07_27.md
@@ -1,0 +1,273 @@
+# Cross-Branch Agent-Definition Gap — Why "Existing" Agents Fail Auth on a Fresh Dev Database
+
+**Date:** 2026-07-27
+**Author:** AgentA
+**Status:** Report — root-cause investigation complete for the confirmed chain; one mechanism (how a
+non-existent agent definition still appears launchable in the UI) remains an open question, flagged
+explicitly in §5. No code changes in this PR.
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2315, opened 2026-07-27
fix(armory): Skill content renders as Markdown; fix broken Bind-to-agent

AgentY-asaf:
## Summary
- **Skill content → Markdown**: the Armory's Skills tab and the per-agent Skills tab (`AgentSkillsModal`) both rendered `content` via a raw `<pre>`. Swapped to the existing `<Markdown>` component (`frontend/app/element/markdown.tsx` — already used for agent chat/editor/tool output, no new dependency), scoped to `content` only (`description`/`trigger` are single-line inputs, left as plain text).
- **"Bind to agent" fixed** — live-verified via CDP against a running dev build, this previously failed 100% of the time with `Bind failed: FORBIDDEN: unauthenticated agent connection`. Root cause: `skill.bind` is `check_s1`-gated (requires the calling WebSocket to be an authenticated agent connection via `bus:register`), but the Armory dashboard's connection never authenticates as an agent. Added `skill.catalog.bind`, a `check_s1`-free sibling mirroring the existing `skill.catalog.list/upsert/delete` trust tier, and repointed the Armory's Bind action to it. `skill.bind` itself is untouched.
- **Bonus, batched in per request**: pane modals dispatched via `ModalLayer` now close on backdrop click when they carry no in-flight form/submit state to lose (currently just the per-agent Stash/setup modal) — previously every `ModalLayer`-hosted modal hardcoded `closeOnBackdropClick={false}` regardless of whether it was an actual form or a pure browse/view surface.

Full investigation + root-cause writeup: `docs/reports/REPORT_ARMORY_SKILLS_MARKDOWN_AND_BIND_BUG_2026_07_27.md`.

**Known follow-up, not fixed here**: MCP Servers' catalog bind has the identical `check_s1` root cause (same originating PR introduced both) — flagged in the report, out of scope for this PR since only Skills was asked for.

## Test plan
- [x] `npx stylelint` on the touched scss — clean
- [x] `cargo test -p agentmux-srv` — 1657/1657 (2 pre-existing, unrelated `agent_session` tests flake under parallel execution; pass cleanly in isolation, unrelated to this diff)
- [x] `npx vitest run app/view/skill app/view/agent app/element` — 769 passed
- [x] Live-verified via CDP against a running `task dev` build: clicked Bind on a real skill → "Used by 1 agent" (was 0, no error), and visually confirmed heading/bold/link/ordered-list Markdown rendering in the Content field
- [ ] Manual: same live check for the per-agent Skills tab (AgentSkillsModal) Markdown rendering

### DIFF
--- a/.changesets/1785132254-fix-armory-render-skill-content-as-markdown-and-fix-the-brok-99xq.md
+++ b/.changesets/1785132254-fix-armory-render-skill-content-as-markdown-and-fix-the-brok-99xq.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(armory): render Skill content as Markdown and fix the broken catalog Bind-to-agent action
--- a/.changesets/1785132255-fix-agent-pane-modals-close-on-backdrop-click-when-they-carr-ec7i.md
+++ b/.changesets/1785132255-fix-agent-pane-modals-close-on-backdrop-click-when-they-carr-ec7i.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): pane modals close on backdrop click when they carry no in-flight form state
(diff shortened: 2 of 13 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed both reagentx P1s (this round + last) and codex's matching P1 — see the two follow-up commits. The detector is now generic (any non-readonly/disabled `<textarea>` within `.modal-root`, not tied to a specific component's class), live-verified by injecting a raw textarea unrelated to any existing tab and confirming it's protected too.

Codex's P2 (cross-channel bind routing via the shared def registry) is real but pre-existing — `Store::skill_bind` itself was never channel-aware; my new `skill.catalog.bind` just makes it reachable from a human-driven dropdown that can list shared-registry-only agents, where the original agent-self-bind path never could. Deferring rather than rushing a fix into a multi-channel/shared-registry area I don't have full context on yet — will follow up separately.

PR #2316, opened 2026-07-27
fix(warden): decouple container-query context from scrolling pane; send auth on LAN fetch

Agent1-asaf:
## Summary

Investigation + fixes for a user report of the Warden pane rendering as a solid black area in a `task dev` instance. Full evidence chain: `docs/retro/retro-warden-pane-black-render-2026-07-27.md`.

**The black pane itself was NOT a Warden bug.** Live diagnosis (in-app DOM instrumentation, native HWND enumeration, `PrintWindow` pixel captures) proved the pane's DOM was fully laid out, visible, and topmost while showing black, with no native surface covering it — and a falsification test (reverting the CSS fix below and reloading) rendered fine, disproving the initial CSS hypothesis. The condition was a sticky renderer/rasterization failure under severe commit pressure (68–75GB of a 78.6GB limit; the in-app memory-pressure banner was actively showing), cleared only by a full page reload, and it survived pane re-creation. The retro documents this as a data point for the #2218 GPU-memory investigation, including a capture-before-reload checklist so the evidence isn't destroyed next time.

## Changes

- **`warden.scss`/`warden.tsx` — container-query context moved to a dedicated non-scrolling `.warden-container` wrapper.** `.warden-pane` was the only element in the codebase combining `container-type: inline-size` with `overflow: auto` — every other consumer (armory, settings, agent-setup modal) uses the wrapper convention `armory-view.scss` documents. The same-element combination produced a real, measured resize feedback loop: a burst of 18 `ResizeObserver loop completed with undelivered notifications` per pane creation, zero after the split. Code comments are explicit that this is convention/hygiene hardening, not the black-pane root cause.
- **`warden.tsx` — LAN section now sends `authedHeaders()` on `GET /api/lan-instances`.** The route has required `X-AuthKey` since the 2026-05-11 security audit, which predates the LAN section's ship date — its "public route (no auth required)" comment was wrong from birth, so the fetch has silently 401'd since day one (visible as an inline `→ 401` error in the section).
- **`docs/retro/retro-warden-pane-black-render-2026-07-27.md`** — the incident retro, including the disproven-hypothesis trail and follow-ups.

## Test plan

- [x] Live-verified in the running `dev-main` instance via `PrintWindow` captures: Warden renders all three sections with the wrapper fix in place.
- [x] Falsification test: old CSS restored + reload → still renders (documented in retro §3).
- [x] LAN 401 fix live-verified: the inline `GET /api/lan-instances → 401` error is gone after the auth header change.
- [x] ResizeObserver loop burst: 18 errors per pane creation before the wrapper split, zero after.

### DIFF
--- a/.changesets/1785133835-fix-warden-decouple-container-query-context-from-the-scrolli-h1lb.md
+++ b/.changesets/1785133835-fix-warden-decouple-container-query-context-from-the-scrolli-h1lb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(warden): decouple container-query context from the scrolling pane; send auth on LAN fetch
--- a/docs/retro/retro-warden-pane-black-render-2026-07-27.md
+++ b/docs/retro/retro-warden-pane-black-render-2026-07-27.md
@@ -1,0 +1,56 @@
+# Retro: Warden pane renders as a solid black area
+
+**Date:** 2026-07-27
+**Severity:** Medium (one widget unusable while the state persists — no data loss, no crash)
+**Area:** Renderer/compositor state, NOT Warden's own code (see §4 — Warden's DOM was proven healthy while the pane showed black)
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2317, opened 2026-07-27
fix(armory): MCP Servers Bind-to-agent — same check_s1 + cross-channel bugs as Skills

AgentY-asaf:
## Summary
Follow-up to #2315. MCP Servers' catalog "Bind to agent" had the exact same two bugs Skills had, since both primitives were built and catalog-wired together in the same original PR:

- `mcp.bind` is `check_s1`-gated (agent binds a server to itself over its own authenticated connection), but the Armory dashboard's connection never authenticates as an agent — every catalog bind attempt failed with `FORBIDDEN: unauthenticated agent connection`. Added `mcp.catalog.bind`, a `check_s1`-free sibling mirroring the existing `mcp.catalog.list/upsert/delete/probe` trust tier. `mcp.bind` itself is untouched.
- `mcp_server_bind` now checks `db_agent_definitions` for the agent id first (same as `skill_bind`), so binding a cross-channel-only agent id fails loudly instead of the FK on `db_agent_mcp_ref.agent_id` silently swallowing the insert and reporting success while creating nothing.

Audited for any other bind-shaped commands beyond Skill/MCP (grepped `commands.rs` for BIND/LINK/ASSIGN/ATTACH) — `linkagentidentity` is a different, older handler with no `check_s1` gate at all, never affected. Skill and MCP are the only two.

## Test plan
- [x] `cargo test -p agentmux-srv` — 1657/1657 passed
- [x] `npx vitest run app/view/mcp app/view/skill app/view/agent app/element` — 769 passed
- [x] Live-verified via CDP against a running dev build: clicked Bind on a real MCP server (`git`) → "Used by 1 agent", no error (previously failed 100% of the time)

### DIFF
--- a/.changesets/1785137808-fix-armory-mcp-servers-bind-to-agent-had-the-same-check-s1-a-h7b9.md
+++ b/.changesets/1785137808-fix-armory-mcp-servers-bind-to-agent-had-the-same-check-s1-a-h7b9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(armory): MCP Servers Bind-to-agent had the same check_s1 auth gap as Skills — fixed identically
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2318, opened 2026-07-27
fix(auth): remove mount-time auto-login trigger, fix login persistence and turn-state races

AgentA-asaf:
## Summary

- **Removed the mount-time auto-login trigger.** Opening an agent pane that needs authentication no longer opens a browser/terminal automatically — it posts a notification and waits for an explicit "Log in" click (wired to `relogin()`). This was the original, long-standing complaint: a login window appearing with no click and no way to decline.
- Added a durable red/green "Logged in" / "Not logged in" status pill to the composer strip.
- Gated `relogin()`'s auto-retry-last-message behavior (`retryAfterLogin`) so a fresh mount-time login doesn't silently resend an unrelated old message — only the failure-row "Login Again" path (a genuinely failed turn) retries.

## Three deeper bugs found and fixed during manual verification

Logins had never previously persisted far enough in this app's history to hit these — they're pre-existing, not introduced by the above:

1. **`IdentityAccount.created_at`/`updated_at` lacked `#[serde(default)]`** — every login's account-persist RPC call failed deserialization 100% of the time. Credentials seeded to disk correctly, but no Armory account or agent↔account link was ever created, so the resolver's spawn gate blocked the very next turn with "not logged in" even after a reported "Login successful."
2. **Lost-update race in `update_object_meta`** — read-merge-write ran across two separate connection-mutex acquisitions instead of one transaction. Two independent writers on the same block (the stale-`--resume`-session-id self-heal vs. the process-exit cleanup) raced on every persistent-controller exit; whichever wrote second silently reverted the other's change. Fixed by wrapping the whole operation in `Store::with_tx`.
3. **Optimistic `TurnStart` never reverted on a synchronous send failure** — `handleSendMessage` sets "Working…" before `AgentInputCommand` is even awaited; if that RPC failed (no controller, spawn gate, network hiccup — anything), nothing ever reverted `turnPhase`, leaving the pane stuck on "Working…" forever regardless of the actual cause.

Full writeup with logs, root-cause traces, and a not-yet-implemented HTTP 429 exponential-backoff design: `docs/specs/REPORT_LOGIN_PERSIST_FAILURE_AND_STUCK_WORKING_2026_07_27.md`.

## Test plan

- [x] Full frontend suite: 1943 passed, 3 skipped (pre-existing e2e skips)
- [x] Full backend suite: 1659 passed
- [x] New regression tests: `launch-flow.test.ts` rewritten for no-auto-login behavior, `run-provider-login.test.ts` extended for tier-3 persist retry, `useAgentCommands.test.ts` (new file) for the turnPhase-revert fix, `object_helpers.rs` (new multi-threaded test) for the lost-update race
- [x] Manual end-to-end: fresh login → persist → force-refresh of an already-running stale process → message delivery, verified live against a running dev instance with real Claude Code
- [ ] More manual testing in progress — flagging this PR is up for early review, not necessarily ready to merge yet

### DIFF
--- a/.changesets/1785165307-fix-auth-remove-mount-time-auto-login-trigger-fix-login-pers-vozj.md
+++ b/.changesets/1785165307-fix-auth-remove-mount-time-auto-login-trigger-fix-login-pers-vozj.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(auth): remove mount-time auto-login trigger, fix login persistence and turn-state races
--- a/.changesets/1785177296-feat-dev-opt-in-isolated-auth-store-for-destructive-armory-t-5v2s.md
+++ b/.changesets/1785177296-feat-dev-opt-in-isolated-auth-store-for-destructive-armory-t-5v2s.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(dev): opt-in isolated auth store for destructive Armory testing (AGENTMUX_ISOLATED_AUTH=1)
--- a/.changesets/1785179340-fix-auth-gate-relogin-opened-tier-success-on-persist-result--k6zw.md
+++ b/.changesets/1785179340-fix-auth-gate-relogin-opened-tier-success-on-persist-result--k6zw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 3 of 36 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #2314, opened 2026-07-27
rename per-agent "Agent setup" modal to "Stash" (backpack icon)

AgentY-asaf:
## Summary
- Renamed the per-agent header-button modal (informally "the agent armory" in code/specs, user-facing label "Agent setup", `vault` icon) to **Stash**, with a distinct `backpack` icon — disambiguating it from the global, fleet-wide **Armory** pane, which is unchanged and keeps `vault`.
- `AgentSetupModal.tsx`/`.scss` → `AgentStashModal.tsx`/`.scss` (via `git mv`), `agent-setup-modal` CSS classes/container-query name → `agent-stash`, modal-layer request `kind: "agent-setup"` → `"agent-stash"`, `_openAgentSetupModal` → `_openAgentStashModal`.
- Internal-only identifiers left untouched (RPC names, DB tables, `AgentMcpModel`/`AgentSkillModel`), matching the original Trust Center → Armory rename's own minimal-risk scope.

See `docs/reports/REPORT_ARMORY_STASH_NAMING_2026_07_27.md` for the full audit (icon-collision analysis, naming alternatives considered, why "Stash" fits) that this PR implements.

## Test plan
- [x] `npx tsc -p tsconfig.json --noEmit` — clean
- [x] `npx stylelint` on the touched/renamed scss files — clean (repo-wide `lint:scss` has 526 pre-existing, unrelated hex-color findings in other files)
- [x] `npx vitest run app/view/agent app/element` — 769 passed
- [ ] Manual: open an agent pane, confirm the header button now shows a backpack icon with tooltip "Stash" and opens the same 5-tab modal at the same width/responsiveness as before

PR #2319, opened 2026-07-27
fix(window): first window's title no longer leaks the unrenamed bootstrap workspace name

AgentY-asaf:
## Summary
Root-caused "sometimes the first window titles as 'Window 1', sometimes as 'Starter workspace'." The window-title 3-tier resolution (`window:displayname` → workspace name → `"Window N"`) let the literal seeded default workspace name (`"Starter workspace"`, `agentmux-srv/wcore/mod.rs`) count as a real tier-2 name. That name is never actually chosen by the user — every fresh data directory gets it — so whether it showed up depended on whether workspace data happened to resolve before the first title paint, producing the inconsistency.

- Excluded the bootstrap default literal from tier 2 in both `resolveWindowName` and `resolveFloatingPaneName` (same bug shape affects floating panes too).
- Window 1 now titles the same deterministic way every other unnamed window does — matching `SPEC_WINDOW_TITLE_FORMAT_2026-05-13.md`'s own "Default unnamed window" example (`Window 1 - Shell - AgentMux`), and giving the user a reliably-distinguishable taskbar entry as requested.
- A workspace the user has actually renamed to something meaningful still shows via tier 2, unchanged — only the never-renamed bootstrap default is excluded.
- Also includes a small unrelated housekeeping fix: `ModalLayer.tsx` still referenced the pre-rename `"agent-setup"` modal-layer kind key; #2314 (merged) renamed it to `"agent-stash"`, flagged as a follow-up at the time.

## Test plan
- [x] `npx vitest run util/window-title.test.ts app/statusbar app/view/agent app/element` — 842 passed, including new coverage for the bootstrap-default exclusion (both `resolveWindowName` and `resolveFloatingPaneName`)
- [x] Live-verified against a fresh `task dev` launch via CDP: OS/taskbar title reads `Window 1 - tab1 - AgentMux`

### DIFF
--- a/.changesets/1785179751-fix-window-first-window-titles-consistently-as-window-1-inst-a177.md
+++ b/.changesets/1785179751-fix-window-first-window-titles-consistently-as-window-1-inst-a177.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(window): first window titles consistently as 'Window 1' instead of sometimes leaking the unrenamed bootstrap workspace name
--- a/frontend/app/element/ModalLayer.tsx
+++ b/frontend/app/element/ModalLayer.tsx
@@ -60,13 +60,12 @@ export const ModalLayer: Component<ModalLayerProps> = (props) => {
     // add-account/new-memory) — so clicking outside should just close
     // them, like any other dismissible overlay.
     //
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2322, opened 2026-07-27
feat(armory): add Agent Skills (SKILL.md) format for skills — ABF Phase 0

AgentX-asaf:
## Summary
Implements **Phase 0** of the rollout plan in [`docs/specs/REPORT_ARMORY_BUNDLE_STANDARD_RESEARCH_2026_07_16.md`](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/REPORT_ARMORY_BUNDLE_STANDARD_RESEARCH_2026_07_16.md) / [docs.agentmux.ai/abf](https://docs.agentmux.ai/abf/) — the report's own "highest-leverage single step": aligning Armory skills with the [Agent Skills](https://agentskills.io/specification) standard so any community `SKILL.md` becomes importable as-is.

**Zero schema migration.** Repurposes the existing `skill_type` column — confirmed via a full-repo search to always be `"prompt"` and never branched on anywhere in the codebase today — as the format discriminator, rather than adding a new column. `skill_type == "agent-skill"` now materializes a skill as `.claude/skills/<slug>/SKILL.md` (native Claude Code / Agent Skills consumption) instead of `.claude/commands/<trigger>.md`.

## Changes
- **`agentmux-srv/src/backend/agent_config.rs`** (authoritative launch path): branches `build_config_files`'s skill-materialization loop on `skill_type`. New `render_skill_md()` writes the two required Agent Skills frontmatter fields (`name`, `description`) via `serde_json` string escaping — which is valid YAML double-quoted-scalar syntax per YAML 1.2 §7.3.1 — avoiding hand-rolled YAML escaping or a new `yaml` crate dependency (this workspace has neither today) for two scalar fields. Slug derived via the existing `derive_slug()` (already used for agent workdir names).
- **`frontend/app/view/agent/agent-config-builder.ts`** (TS mirror — must stay in sync with the Rust path per the research report's own note, or the two diverge): identical branch, plus a `deriveSlug()` that exactly mirrors Rust's algorithm. Verified this needed to be a new function, not a reuse of the existing `slugifyInstanceName` — that helper strips invalid characters rather than collapsing runs of them into a single dash, which diverges from Rust on inputs like `"Deploy!!!Checklist"` (see the regression test).
- **UI**: `skill_type` was previously stored end-to-end but never surfaced in any form — the Armory catalog's `skill-manager.tsx` and the per-agent `AgentSkillsModal.tsx` both hardcoded `"prompt"`. Both now expose a **Format** select (Slash command / Agent Skill — beta ABF format); the Trigger field hides itself when Agent Skill format is selected since it's unused for that materialization path.

## Test plan
- [x] 2 new Rust unit tests (`agent_config.rs`): SKILL.md materialization path, and YAML-special-character escaping regression test.
- [x] 9 new Vitest tests (`agent-config-builder.test.ts`): `deriveSlug` parity with Rust (including the divergence-from-`slugifyInstanceName` regression case), `renderSkillMd`, and the `buildConfigFiles` branch (both formats + empty-content skip).
- [x] Full `agent_config::` Rust test module: 13/13 pass.
- [x] Full frontend `tsc --noEmit`: clean, no errors.
- [x] Full Vitest run across `frontend/app/view/agent/` + `frontend/app/view/skill/`: 251/251 pass.
- [ ] Manual: create an Agent Skill-format skill in the Armory, launch an agent bound to it, confirm `.claude/skills/<slug>/SKILL.md` is written with correct frontmatter and Claude Code picks it up natively.

## Scope note
This is Phase 0 only (of 6 phases in the rollout plan). Phases 1-4 (exporter, importer, Armory export/import UI, OCI distribution) are separate, larger follow-ups — not attempted here.

### DIFF
--- a/.changesets/1785185774-feat-armory-add-agent-skills-skill-md-format-for-skills-abf--s97c.md
+++ b/.changesets/1785185774-feat-armory-add-agent-skills-skill-md-format-for-skills-abf--s97c.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): add Agent Skills (SKILL.md) format for skills, ABF Phase 0
--- a/agentmux-srv/src/backend/agent_config.rs
+++ b/agentmux-srv/src/backend/agent_config.rs
@@ -562,6 +562,21 @@ pub fn expand_template(content: &str, vars: &HashMap<String, String>) -> String 
 /// Agent Skills spec caps `description` at 1024 characters.
 const SKILL_DESCRIPTION_MAX_LEN: usize = 1024;
 
+/// Truncate `s` to at most `max_units` UTF-16 code units, on a UTF-8 char
+/// boundary. Mirrors JS `.slice(0, n)` semantics (JS strings are indexed in
+/// UTF-16 code units, so a Rust byte-length cap diverges from the TS mirror
+/// for any non-ASCII text — reagent P2, PR #2322).
+fn truncate_utf16_units(s: &str, max_units: usize) -> String {
+    let mut units = 0usize;
+    for (byte_idx, ch) in s.char_indices() {
+        units += ch.len_utf16();
+        if units > max_units {
+            return s[..byte_idx].to_string();
+        }
+    }
+    s.to_string()
+}
+
 /// Render an Agent Skills-format `SKILL.md`: YAML frontmatter with the two
 /// required fields (`name`, `description` — per the spec's six-field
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #2323, opened 2026-07-27
feat(agent): guarantee pane close/reopen resumes the real conversation, or discloses it couldn't

Agent1-asaf:
## Summary

Implements `docs/specs/SPEC_PANE_CLOSE_REOPEN_CONTINUITY_GUARANTEE_2026_07_27.md`: **closing an agent's pane and reopening it must resume the same conversation, or the user must be told it couldn't.** Written after a user question about whether Armory workspace rules are "always available" to a running agent surfaced that pane-reopen's `--resume` success was unverified end-to-end — investigating that turned up two real, evidenced gaps, both closed here.

### Gap 1 — the picker could be handed a permanently-stale session id

`ListRecentSessionsCommand` prefers the local `db_agent_instances` row over the shared registry, but that row's `session_id` was set to `""` at instance creation and **never updated again in production** — the RPC that would keep it live (`updateagentinstance`) had zero frontend callers. Any agent launched at least once in the running app had a stale local session id, so closing its pane and reopening via "My Agents" could silently start a genuinely fresh conversation instead of resuming (per `agent-model.ts`'s own documented invariant: an empty `continueSessionId` clears `agent:sessionid` on the new block, by design, to avoid a *different* prior bug).

**Fix:** `persist_session_id` (the live-turn call site every controller already uses) now also writes through to the matching `db_agent_instances` row, via a new `Store::instance_get_by_block_id` + `instance_update_partial`. Added a registry fallback in `ListRecentSessionsCommand` too, for defense-in-depth regardless of whether a given row predates this fix.

### Gap 2 — a failed resume recovered silently

`persistent.rs` already recovers gracefully when a `--resume <sid>` is rejected (stale/unreachable session, e.g. after a relogin moves an agent to a different config dir): it clears the poisoned id and lets the next respawn start fresh. But this was completely silent — only a `tracing::warn!`, no signal to the frontend at all. The user would just get an agent that's forgotten everything, with no indication anything happened.

**Fix:** a new `session:resume_failed` meta flag (`session_recovery.rs`, same frontend-only-signal contract as the existing `session:was_interrupted` flag/banner), set at the exact point `persistent.rs`'s `poison_resume` path fires, plus a matching `AgentControlBar` banner + dismiss action mirroring the existing `was_interrupted` banner's shape and copy style.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 1661 passed, 0 failed. 5 new/updated tests: `instance_get_by_block_id`, `persist_session_id_writes_through_to_the_instance_row`, `persist_session_id_is_a_noop_for_a_block_with_no_instance_row`, `mark_resume_failed_sets_the_flag`.
- [x] `npx vitest run` frontend suite — 751 passed, 0 failed, no regressions.
- [x] `cargo check -p agentmux-srv` — clean.
- [~] Live UI verification of the banner itself was attempted in a `task dev` instance but inconclusive — a manual out-of-band SQLite edit used to set the flag (to avoid needing to trigger a real `--resume` failure) desynced the running srv's in-memory object-version tracking and left that one pane stuck loading. Not a regression from this PR's code (confirmed via host log — no errors, purely a side effect of writing to the live DB file outside the srv's own update path); the banner mechanism itself is code-identical to the already-shipped, working `session:was_interrupted` banner (same signal-read/dismiss/JSX pattern). Flagging this transparently rather than claiming a clean live verification that didn't actually happen.

### DIFF
--- a/.changesets/1785186317-feat-agent-guarantee-pane-close-reopen-resumes-the-real-conv-j2it.md
+++ b/.changesets/1785186317-feat-agent-guarantee-pane-close-reopen-resumes-the-real-conv-j2it.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): guarantee pane close/reopen resumes the real conversation, or discloses it couldn't
--- a/agentmux-srv/src/backend/blockcontroller/core.rs
+++ b/agentmux-srv/src/backend/blockcontroller/core.rs
@@ -121,6 +121,19 @@ pub(crate) fn spawn_health_watchdog(health_monitor: &Arc<HealthMonitor>) {
 /// missing this step (it only set `inner.session_id` in memory) — A5 fixes it
 /// by routing ACP through this same function.
 ///
+/// Also writes through to the matching `db_agent_instances` row's
+/// `session_id` column (SPEC_PANE_CLOSE_REOPEN_CONTINUITY_GUARANTEE_2026_07_27.md
+/// §4.1). Without this, that column was set to `""` at instance creation and
+/// never updated again in production — `ListRecentSessionsCommand` prefers
+/// the local row over the shared registry whenever both exist, so the "My
+/// Agents" picker's reattach flow (`AgentPicker.tsx`'s `handleReattach`) was
+/// handed an empty session id even mid-session, and — per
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #2324, opened 2026-07-27
fix(armory): align ArmorySection ids with their labels

AgentY-asaf:
## Summary

The Armory rail's Phase 5 rename (Memory -> Memories, Bundles) changed the display labels but left the underlying `ArmorySection` ids pointing at the wrong-sounding pane:

- `id: "brain"` sat behind the **"Memories"** label and the `GlobalBrainManager` pane.
- `id: "memories"` sat behind the **"Bundles"** label and the `MemoryManager` pane.

So `section() === "memories"` actually finds the Bundles pane, not the one labeled "Memories" — misleading for anyone reading or extending `armory-view.tsx`.

## Fix

Renamed the two ids to match their labels/purpose: `"brain"` -> `"memory"`, `"memories"` -> `"bundles"`. This also aligns with the sibling per-agent `AgentStashModal`'s `StashTabId`, which already uses `"memory"` for its own Memories tab.

No behavior change — same five sections, same order, same panes, same labels. Confirmed (via grep) no other code references the old id strings; the existing rail test only asserts on labels/DOM, not ids.

## Verification

- `npx tsc -p tsconfig.json --noEmit` — clean
- `npx vitest run app/view/armory` — 9/9 passed
- `npx vitest run app/view` (full view suite, regression check) — 882/882 passed

### DIFF
--- a/.changesets/1785195417-fix-armory-align-armorysection-ids-with-their-labels-memory--0itj.md
+++ b/.changesets/1785195417-fix-armory-align-armorysection-ids-with-their-labels-memory--0itj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(armory): align ArmorySection ids with their labels (memory/bundles, not brain/memories)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2321, opened 2026-07-27
feat(telemetry): [wave-turn] and [health] diagnostic logging for the Working-state watchdog

AgentY-asaf:
## Summary

Audits every way the pane "Working..." indicator can appear with no real cause (triggered by a live incident this session: a leftover `task dev` background process left a pane looking stuck). Full findings in `docs/reports/REPORT_WORKING_STATE_TELEMETRY_AUDIT_2026_07_27.md`:

- §1 ranks nine false-positive paths through the `TurnPhase` state machine and its watchdog.
- §2 identifies telemetry gaps — the reducer's `dispatch()` logged almost nothing, the backend `HealthMonitor` only logs on composite-health *category* change (not on raw `turn_active` flips), and tool-call activity never reaches the backend at all.
- §3 recommends four fixes, ranked. This PR implements the top three (§3.1-§3.3); §3.4 (backend process-registry long-lived-process logging) is deferred, blocked on a prerequisite fix to `started_at_ms` being hardcoded to `0` on Windows.

## Changes (logging-only, no behavior change)

- `frontend/app/store/agent-pane-state-store.ts`: `dispatch()` now logs a `[wave-turn]` line on every `TurnPhase` transition, plus the watchdog's own reasoning for why it did or didn't force-recover a pane — mirroring the existing `[wave-title]` debug-log precedent, so it's greppable via `muxlog host '[fe] [wave-turn]'`.
- `agentmux-srv/src/backend/blockcontroller/health.rs`: `HealthMonitor::set_active_turn`, `mark_turn_active_returning_was_active`, and `set_exited` now log every `turn_active` flip as `"[health] turn_active flip"`, greppable via `muxlog srv health`.

## Verification

- `npx tsc -p tsconfig.json --noEmit` — clean
- `cargo check -p agentmux-srv` — clean (no new warnings)
- `cargo test -p agentmux-srv health::` — 9/9 passed
- `cargo test -p agentmux-srv` (full suite) — 1660 passed, 0 failed, 4 ignored
- `npx vitest run app/store` — 638/638 passed
- Live-verified in a `RUST_LOG=debug task dev` instance via CDP: launched an agent pane, sent a real message, confirmed both `[wave-turn]` and `[health] turn_active flip` lines appeared with correct structured content across a genuine `Idle → Submitting → Streaming → Done` cycle — including a live "Done → Streaming → Idle" re-promotion event that the new logging fully explained.

PR #2326, opened 2026-07-27
feat(agent): harden Working-state and scroll-follow against 4 recurring desync bugs

Agent1-asaf:
## Summary

Folds four related live incidents from this session into one hardening pass:

1. This agent's own pane got stuck showing "Working…" after a PR had genuinely merged.
2. The message list's auto-scroll-follow — previously "fixed" in SPEC_AGENT_PANE_SCROLL_FOLLOW_AND_STATUS_OVERLAY_2026_07_24.md — was still observed silently drifting off true bottom.
3. A pane was observed going "Worked" → "Working…" again with no user input in between.
4. Agent2's pane got stuck on an unanswerable AskUserQuestion prompt (documented only — root cause is a hypothesis, not live-confirmed; not fixed this pass).

Full writeup: `docs/specs/SPEC_WORKING_STATE_AND_SCROLL_FOLLOW_HARDENING_2026_07_27.md`, with supporting root-cause docs `docs/reports/REPORT_WORKING_STATE_TELEMETRY_AUDIT_2026_07_27.md` (pre-existing), `docs/specs/REPORT_LOGIN_PERSIST_FAILURE_AND_STUCK_WORKING_2026_07_27.md` (pre-existing), and `docs/reports/REPORT_WORKING_STATE_REGRESSION_AND_STUCK_QUESTION_PANEL_2026_07_27.md` (new, this session).

## Changes

**Stuck-"Working" (backend fire-once event gap):**
- `blockcontroller/mod.rs`: `publish_controller_status`'s `persist: 0 → 1`, mirroring the existing `EVENT_AGENT_FAILURE` precedent — a WS reconnect now replays the last known turn state instead of nothing.
- `blockcontroller/persistent.rs`: new `spawn_status_heartbeat`, wired alongside every `spawn_health_watchdog` call site — republishes status every 20s while a turn is active, a self-healing backstop independent of reconnect/persist.
- `agent-view.tsx`: new focus/visibility-triggered reconcile via the pre-existing `makeWindowFocusSignal()` — covers the same-connection pane-remount case persist alone doesn't.
- `[wave-turn]` telemetry (report 1 §3.1/§3.2) turned out to already be implemented via a concurrent PR (#2321) from another agent mid-session — verified present, no duplicate work.

**Scroll-follow drift (still recurring after the 07-24 fix):**
- `AgentDocumentVirtualList.tsx`: generalized the scroll container's `ResizeObserver` to re-pin on ANY `clientHeight` change while `stickToBottom()` is true, not just hidden→visible. Root cause: a sibling row (retry bar, decision/question panel, pending-messages panel) appearing/growing mid-turn shrinks the flex-1 scroll container via pure CSS reflow — no scroll event fires, none of the three previously-tracked dependencies change, so `stickToBottom` stays silently true while the view falls short of true bottom.
- `usePendingMessageAcceptance.ts` → `useAgentStream.ts` → `agent-view.tsx`: a new turn starting via the queue-drain path (auto-accepted queued message) now re-engages `stickToBottom` too — previously only composer-driven turns did.

**"Worked" reverting to "Working…" with no input:**
- `agent-view.tsx`: additive-only settled-grace window (500ms). The reducer's `Done.completed → Streaming` re-promotion is intentional (multi-round tool continuations need it — removing it would be worse), but once the UI has shown a *settled* "Worked" for the grace window, a later re-promotion now posts a visible "Picked up more work — starting another round…" notification instead of silently reverting the indicator.

**Explicitly deferred** (see spec §5): the frontend tool-liveness gap (unconditional `toolsActive > 0` watchdog exemption), scroll input-gating (the programmatic-vs-user-scroll race), and Agent2's stuck-question panel (needs live confirmation first).

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 1675 passed / 0 failed / 4 ignored (net +1 from the new persist-replay regression test)
- [x] `npx vitest run app/view/agent app/store/agent-pane-state app/window` — 66 files / 931 tests passed, 0 regressions
- [ ] No live UI verification this pass — deliberately avoided direct DB manipulation of a live srv instance (documented hazard from earlier this session); relying on automated test coverage only

### DIFF
--- a/.changesets/1785196188-feat-agent-harden-working-state-and-scroll-follow-against-4--f1fl.md
+++ b/.changesets/1785196188-feat-agent-harden-working-state-and-scroll-follow-against-4--f1fl.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): harden Working-state and scroll-follow against 4 recurring desync bugs
--- a/agentmux-srv/src/backend/blockcontroller/mod.rs
+++ b/agentmux-srv/src/backend/blockcontroller/mod.rs
@@ -489,7 +489,19 @@ pub fn resync_controller(
     }
 }
 
-/// Publish a controller status event via WPS broker.
+/// Publish a controller status event via WPS broker. The sole publish point
+/// for `controllerstatus`, used by every controller type (persistent CLI,
+/// subprocess CLI, ACP agents, plain shell/PTY panes — 13 call sites).
+///
+/// `persist: 1` so a reconnecting subscriber (WS reconnect, srv restart)
+/// replays the last known status instead of seeing nothing until the next
+/// live event — mirrors `EVENT_AGENT_FAILURE`'s identical `persist: 1` in
+/// `subprocess/host_spawn.rs`. Closes the "missed the one turn-end push,
+/// stuck showing Working forever" gap for the reconnect case; see
+/// REPORT_LOGIN_PERSIST_FAILURE_AND_STUCK_WORKING_2026_07_27.md §3/§4 item 5.
+/// (A same-connection pane remount, which doesn't clear the broker's
+/// per-route replay tracking, isn't covered by persist alone — see the
+/// focus-triggered reconcile in `agent-view.tsx` for that case.)
 pub fn publish_controller_status(
     broker: &super::wps::Broker,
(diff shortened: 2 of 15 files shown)


a5af:
@codex review

PR #2327, opened 2026-07-27
fix(agent): harden My Agents against single-source failures

AgentA-asaf:
## Summary
Root-caused in `docs/retro/retro-my-agents-fresh-channel-regression-2026-07-27.md` (included in this PR), written after a user report that a fresh portable build's "My Agents" list appeared empty.

Investigation found no live data corruption on the reporting machine, but surfaced a proven, still-exploitable structural fragility: PR #2296 (merged 2026-07-25) already hit this exact user-facing symptom once — one unparseable `db_accounts` row aborted the entire `listrecentsessions` response. That fix patched the one instance; the all-or-nothing handler shape was untouched, so any other malformed row in any of six sequential data sources could reproduce it.

- **Backend** (`session.rs`): each of the six data sources now degrades to empty + logs a warning on its own failure, instead of `?`-aborting the whole response. Added a completion-level trace (row count, which sources degraded, latency) — the investigation's biggest evidentiary gap was that this RPC wasn't traced at all.
- **Frontend** (`MyAgentsList.tsx`): a failed fetch and a genuinely empty list used to render identical UI — exactly what let PR #2296's regression go unnoticed as "expected empty state" the first time. Added a distinct error state with a retry button.
- **Tests**: a Rust integration test seeding a real cross-channel `Registry` + `DefinitionStore` and asserting all agents populate on a fresh channel (this exact path has broken 3 times with no regression test covering it), a second test reproducing PR #2296's exact malformed-row shape via raw SQL and asserting the list still returns the other agents, plus frontend tests for the new error state and its retry action.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 1676 passed, including the 2 new regression tests
- [x] `npx vitest run` — 1962 passed, including the 2 new MyAgentsList tests
- [ ] CI green on this PR

### DIFF
--- a/.changesets/1785196299-fix-agent-harden-my-agents-against-single-source-failures-di-56hd.md
+++ b/.changesets/1785196299-fix-agent-harden-my-agents-against-single-source-failures-di-56hd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): harden My Agents against single-source failures, distinguish error from empty
--- a/agentmux-srv/src/backend/rpc_types/instance.rs
+++ b/agentmux-srv/src/backend/rpc_types/instance.rs
@@ -178,6 +178,26 @@ pub struct RecentSessionRow {
     pub agent_type: String,
 }
 
+/// Response envelope for `listrecentsessions`. Introduced alongside the
+/// per-source degradation hardening in `session.rs` (reagent P1 on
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

a5af:
@codex review

PR #2328, opened 2026-07-28
fix(agent): keep My Agents loading state visible during a retry

AgentA-asaf:
## Summary
Follow-up to #2327 (merged) — chatgpt-codex-connector's post-merge re-review of commit e5de8e15 found a real gap in that PR's fix: `frontend/app/view/agent/components/MyAgentsList.tsx:149`.

Solid's `createResource` keeps the previous resolved value (`[]`) visible while a refetch is in flight (stale-while-revalidate). Since `fetchError` is cleared synchronously the instant Retry starts, and `isLoading` checked `rows() === undefined` (only true before the very first fetch ever resolves), clicking Retry after a failure fell straight through to the empty-state branch for the retry's entire in-flight duration — flashing "No agents yet, pick a template below" and recreating the exact error/empty ambiguity #2327 exists to remove.

`isLoading` now reads the resource's own `.loading` flag instead, which Solid sets for any in-flight fetch (initial or refetch) regardless of whether a stale value is still visible.

## Test plan
- [x] Added a regression test that clicks Retry, holds the retry promise pending, and asserts neither the empty nor error state renders during that window
- [x] Verified the test actually catches the bug: reverted the fix, confirmed the test fails, restored the fix
- [x] `npx vitest run` — 1966 passed (frontend), same pre-existing unrelated flake as #2327
- [ ] CI green on this PR

Codex finding: https://github.com/agentmuxai/agentmux/pull/2327#discussion_r3661921580

### DIFF
--- a/.changesets/1785199520-fix-agent-keep-my-agents-loading-state-visible-during-a-retr-iar3.md
+++ b/.changesets/1785199520-fix-agent-keep-my-agents-loading-state-visible-during-a-retr-iar3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): keep My Agents loading state visible during a retry, don't flash empty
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2329, opened 2026-07-28
fix(stash): MCP Servers and Skills tabs no longer fail unauthorized

AgentY-asaf:
## Summary

The Stash modal's MCP Servers and Skills tabs called `mcp.list`/`upsert`/`delete`/`bind`/`unbind` and `skill.list`/`upsert`/`delete`/`bind`/`unbind` — every one of which is gated by `check_s1`, which requires the calling connection to *be* that agent's own authenticated bus connection. The Stash modal runs over the dashboard window's connection, so **every action failed "unauthorized," including the very first list-load** — the whole tab looked broken from the moment it opened.

## Fix

Following the precedent already set for `mcp.catalog.bind`/`skill.catalog.bind` (added earlier for the identical reason) and for the Accounts tab (already converted to a read-only reactive view over Armory-managed bindings), both tabs are now a **reactive read + bind/unbind view of the Armory catalog** instead of a separate, always-broken authoring surface:

- New window-scoped, no-`check_s1` commands: `mcp.catalog.list_for_agent`, `mcp.catalog.unbind`, `skill.catalog.list_for_agent`, `skill.catalog.unbind` — mirror `mcp.list`/`skill.list`'s own computation (every global item plus this agent's own, each annotated with `bound_to_agent`), just without the gate that can never be satisfied from a UI connection.
- `mcp.catalog.bind`/`skill.catalog.bind` now also publish `mcp:changed`/`skills:changed` (previously silent), so `AgentMcpModel`/`AgentSkillModel` can subscribe and stay in sync across windows without a manual refresh.
- `AgentMcpModal`/`AgentSkillsModal`: removed the New/Edit/Delete UI for private per-agent items — it never worked (`check_s1` blocked it from day one), so nothing functional is lost. Kept the read-only detail view and Bind/Unbind toggle for global (Armory-managed) items. A private item an agent created via its own tool calls still shows up (read-only), since those are real per-agent data.

## Verification

- `npx tsc -p tsconfig.json --noEmit` — clean
- `cargo check -p agentmux-srv` — clean (no new warnings)
- `cargo test -p agentmux-srv` (full suite) — 1677 passed, 0 failed, 4 ignored
- `npx vitest run app/view` — 895/895 passed
- Live-verified in `task dev` via CDP against a real agent (`openclaw`):
  - `mcp.catalog.list_for_agent` / `skill.catalog.list_for_agent` — both return real data, zero "unauthorized" errors (previously both threw immediately).
  - Full bind → unbind → rebind roundtrip on a real global MCP server (`git`), confirming `bound_to_agent` flips correctly at each step and the final state matches the starting state.

### DIFF
--- a/.changesets/1785199510-fix-stash-mcp-servers-and-skills-tabs-no-longer-fail-unautho-q3oi.md
+++ b/.changesets/1785199510-fix-stash-mcp-servers-and-skills-tabs-no-longer-fail-unautho-q3oi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(stash): MCP Servers and Skills tabs no longer fail unauthorized; reactive read+bind view over the Armory catalog
--- a/agentmux-srv/src/backend/rpc_types/commands.rs
+++ b/agentmux-srv/src/backend/rpc_types/commands.rs
@@ -337,6 +337,14 @@ pub const COMMAND_SKILL_CATALOG_DELETE: &str = "skill.catalog.delete";
 // catalog-tier sibling the Armory's "Bind to agent" action actually calls.
 // See docs/reports/REPORT_ARMORY_SKILLS_MARKDOWN_AND_BIND_BUG_2026_07_27.md.
 pub const COMMAND_SKILL_CATALOG_BIND: &str = "skill.catalog.bind";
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

PR #2330, opened 2026-07-28
feat(agent): collapse mid-line spinner/progress redraws in tool preview

AgentX-asaf:
## Summary
Implements `docs/specs/SPEC_TOOL_LOG_UNIVERSAL_ANIMATION_COLLAPSE_2026_07_27.md` — generalizes the existing spinner-collapse fix (`SPEC_TOOL_LOG_INPLACE_ANIMATION_2026_06_22.md`) beyond "the animated character is the entire line" to also cover a spinner glyph or progress percentage trailing/leading static text on the same line (`Installing deps... ⠋`, `Downloading (45%)`) — the actual shape real install/build tools use, and the literal repeated-line bug reported during installs in the agent pane's tool preview.

## Backend (`agentmux-bashwrap/src/bash_wrap.rs`)
Applied identically to both `pty_reader_loop` and `stream_reader`:
- **A1**: a pending line with no leading/trailing `\r` is now speculatively deferred for one extra quiet-window tick (~50-100ms) instead of flushed immediately, so a static label printed once and later overwritten via `\r` doesn't get permanently committed as its own line before the overwrite arrives.
- **A2**: CHA-to-column-1 (`\x1b[1G`/`\x1b[G`) and EL (`\x1b[K`/`\x1b[2K`/`\x1b[0K`) are normalized to a literal `\r` before `strip_ansi` deletes them, so CSI-based redraws collapse the same way `\r`-based ones already do.
- **A2 extension (discovered via the e2e test, not anticipated by the spec)**: Windows ConPTY re-serializes a literal `\r` as CUP-to-home (`\x1b[H`) rather than passing it through unchanged — verified by capturing raw PTY bytes. Handled with the same `\r`-equivalence, gated on a `still_on_first_line` flag so a genuine multi-line CUU-style redraw (spec's explicitly out-of-scope §A3) is never misread as "restart the current line."
- CUB (`\x1b[<n>D`) intentionally left unhandled — needs true per-column tracking this line-offset model doesn't have; documented follow-up alongside multi-line redraw.

## Frontend (`frontend/app/view/agent/components/output-cap.ts`)
Consumed unchanged by `ToolOverlayLog.tsx` and `PersistentShellBlock.tsx`. `collapseSpinnerChunks` now also collapses a run via `looksLikeRedraw` — normalize away spinner glyphs/digits/percent/fill-runs, then either exact-match or a Levenshtein-similarity fallback. **Threshold raised from the spec's initial 0.82 estimate to 0.92** after testing showed 0.82 false-positived on same-shape-different-content lines common in real build output (`"Compiling src/main.rs"` vs `"Compiling src/lib.rs"` scores ~0.86). The dominant real cases (glyph/percentage trailing static text) hit an exact-match fast path regardless of this threshold, so raising it only trims false positives, not real coverage.

## Test plan
- [x] `cargo test -p agentmux-bashwrap`: 60 tests incl. a real-PTY end-to-end test that reproduces the exact duplicate-line bug against actual bash + ConPTY (retried up to 5x internally for real OS scheduling jitter around the defer window — `FLUSH_QUIET_WINDOW` has no test override). 8 consecutive full-suite runs clean.
- [x] `cargo build --workspace` clean.
- [x] `vitest output-cap.test.ts`: 36 passed (10 new, incl. explicit false-positive regression tests).
- [x] `vitest ToolOverlayLog.test.tsx`: 7 passed, unaffected.
- [x] `tsc --noEmit` clean.
- [ ] Manual: run a real `npm install`/`cargo build` as an agent Bash tool call, confirm a single settling line in the expanded tool preview / persistent-shell panel.

### DIFF
--- a/.changesets/1785201516-feat-agent-collapse-mid-line-spinner-progress-redraws-in-too-sink.md
+++ b/.changesets/1785201516-feat-agent-collapse-mid-line-spinner-progress-redraws-in-too-sink.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): collapse mid-line spinner/progress redraws in tool preview + bashwrap
--- a/agentmux-bashwrap/src/bash_wrap.rs
+++ b/agentmux-bashwrap/src/bash_wrap.rs
@@ -440,12 +440,16 @@ where
     // new frame, collapsing throttled spinner frames (>50 ms apart) before
     // they become separate LineEvents.
     let mut pending_cr_override: Option<Vec<u8>> = None;
+    // A1: speculative one-tick hold for a pending line with no leading OR
+    // trailing `\r` yet. Mirrors the same slot in `pty_reader_loop`. See
+    // docs/specs/SPEC_TOOL_LOG_UNIVERSAL_ANIMATION_COLLAPSE_2026_07_27.md §A1.
+    let mut deferred_pending: Option<Vec<u8>> = None;
     loop {
         match tokio::time::timeout(FLUSH_QUIET_WINDOW, reader.read(&mut buf)).await {
             Ok(Ok(0)) => {
-                // EOF: flush any held CR override, then drain remainder,
-                // stripping a dangling lone \r (stream ended mid-spinner
-                // or mid-CRLF pair).
+                // EOF: flush any held CR override or A1 deferred line, then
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2331, opened 2026-07-28
fix(agent): let an AskUserQuestion answer survive a pane reopen or process respawn

Agent1-asaf:
## Summary

Live-diagnosed this session from a real stuck pane (Agent2), and directly from the user's own report of the exact symptom: "even if I answer, it flickers and comes back to the same state."

**Root cause (confirmed in code, not hypothesized):** `PersistentSubprocessController::answer_question` tracks pending AskUserQuestion state in `pending_questions: HashMap<tool_use_id, ...>` — **in-memory only, scoped to one controller instance.** A fresh instance (created on every pane reopen, or any process respawn for any reason) starts with an empty map, even though the persisted transcript still correctly shows the question as the tail `awaiting_answer` node (`scrubOrphanedInProgress` deliberately preserves it as "may still be answerable"). Answering it after ANY respawn therefore always failed at the backend with `"no pending AskUserQuestion for tool_use_id …"` — a different error shape than `UNSUPPORTED_CONTROLLER`, the only string `useAgentQuestions.ts`'s `handleAnswer()` checked for before falling back to the existing (already-built, for non-persistent agents) follow-up-message delivery path. Every other failure just rolled the optimistic "answered" UI back to `awaiting_answer` — this is the precise mechanism behind "it flickers and comes back to the same state," and it meant a question surviving ANY pane reopen was permanently unanswerable through the UI.

## Changes

- `frontend/app/view/agent/hooks/useAgentQuestions.ts` — `handleAnswer`'s fallback is no longer gated on the one `UNSUPPORTED_CONTROLLER` string; it now retries via the follow-up-message path on **any** `AgentAnswerCommand` failure, only rolling back if the fallback itself also fails. This is strictly safer, and since `send_message` (backend) already auto-spawns/resumes a dead process, it also self-heals a genuinely dead process without requiring a manual pane reopen at all.
- `agentmux-srv/src/backend/blockcontroller/persistent.rs` — `answer_question`'s "not found" error message now explicitly names the tool_use_id and the likely cause (process respawn), for `muxlog` diagnosability — the frontend fix no longer depends on the exact string, but a clear message still matters for debugging.
- `docs/reports/REPORT_WORKING_STATE_REGRESSION_AND_STUCK_QUESTION_PANEL_2026_07_27.md` §2.7/§2.8 — documents both the live-confirmed process-death diagnosis (§2.7, from srv log analysis) and this fix (§2.8).

## Test plan

- [x] New backend test `answer_question_on_untracked_tool_use_id_is_descriptive` — asserts the error names the tool_use_id + likely cause
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 1678 passed / 0 failed / 4 ignored
- [x] New frontend test `useAgentQuestions.test.ts` (3 tests) — success path (no fallback), non-`UNSUPPORTED_CONTROLLER` failure (falls back, keeps optimistic state), fallback-also-fails (rolls back)
- [x] `npx vitest run app/view/agent app/store/agent-pane-state app/window` — 69 files / 949 tests passed, 0 regressions


a5af:
@codex review

PR #1177, opened 2026-05-29
feat(floating-pane): edge-resize floaters by dragging edges/corners

AgentA-asaf:
## What
Floating panes can now be **resized by dragging any edge or corner** — terminal, agent, and browser floaters. Previously a no-op.

## Approach — JS-driven (the HTTRANSPARENT forwarder was abandoned)
The forwarder in the original PR A **does not work on cef-rs 146**: Chromium consumes `WM_NCHITTEST` before any child-wndproc subclass sees it, so the forwarder got **zero** hits (confirmed with live diagnostics). A `WM_SYSCOMMAND(SC_SIZE)` fallback also failed — the post succeeds, but Chromium holds the OS mouse capture from the DOM pointerdown and the host's `ReleaseCapture` is cross-thread (no-op), so the native resize loop can't track the mouse.

Shipped instead — **JS-driven resize**, mirroring the already-working JS header move:
- `floating-pane-workspace.tsx` detects an edge pointerdown within a 12px band, takes **pointer capture** (keeps receiving moves after the cursor leaves the window), reads the start rect via `get_window_rect`, and on each move computes the new rect (delta + which of 8 edges) and calls `set_window_rect` → `SetWindowPos`. Moves coalesce one-IPC-in-flight.
- New host IPC `get_window_rect` / `set_window_rect` (`commands/window/motion.rs`).
- **Browser floaters** have a separate web-content child window over the content region; `browser-view.tsx` insets it by the band depth on left/right/bottom (the top edge is over the 33px header, already frontend) so the frontend owns the grab band. The placeholder paints the strip → **no native white border**.
- Shared band constant (`floater-resize.ts`) keeps detector + inset in sync.

## Removed
All forwarder machinery (`floating_pane.rs`, `client/mod.rs`, `browser_pane/callbacks.rs` reverted to base) and the `SC_SIZE` `start_window_resize`.

## Review notes
- **codex P2** (forwarder reinstall-on-navigation): resolved by deleting the forwarder.
- prior **reagent** CHANGES_REQUESTED was the degraded `kimi+haiku-timeout` `incomplete-review` meta-note, not a code finding.

## Test (Windows, local portable)
- Terminal/agent floater: all 8 edge/corner zones resize ✓
- Browser floater: content-area edges/corners resize (12px inset frame visible) ✓
- Redock: tear-off → drop back onto another window still works ✓

## Follow-ups (not in this PR)
[...]

PR #2332, opened 2026-07-28
fix(floating-pane): restore resize hit-target from 4px to 8px

Agent2-asaf:
## Summary

Floating-pane resize handles regressed from a comfortable 12px grab target down to an unusable 4px, as an unintended side effect of an unrelated cosmetic fix.

## Root cause

`FLOATER_EDGE_RESIZE_BORDER` (`frontend/app/workspace/floater-resize.ts`) drives two things with one constant: the invisible mouse hit-test band width for resizing a floating pane by its edge, *and* how much a floating browser pane's separate OS child window is inset so that band stays clickable underneath it.

- Shipped at **12px** in #1177 (2026-05-29) — comfortable, no complaints for non-browser floaters.
- #1829 (2026-06-29) fixed a real visual bug — browser panes' web-content inset reading as a background-colored border around the content — by shrinking the *shared* constant, first to 6px then to **4px** in the same PR.

That fixed the browser-pane matte, but also shrank the grab target for every other floating pane type (agent, terminal, editor, sysinfo, drone, ...) — none of which have a matte problem at all, since they have no separate child window sitting over the frontend DOM. Two docs still cited the original 12px and were never updated, so there was no trail pointing at the regression.

## Fix

Restored to **8px** — roughly double the unusable 4px, comfortably under the 12px that read as a border on browser panes. Kept a single shared constant rather than splitting hit-test width from the browser-pane inset: the two must mechanically match for the exposed band to actually receive pointer events (a wider hit-test zone than the inset would just have the CEF child window swallow clicks in the extra region), and per-pane-type branching in the hot pointer-move path was judged more risk than the marginal gain over a single well-chosen shared value.

Also fixed two stale doc references (both still cited 12px; one also pointed at `browser-view.tsx` for inset logic that's since moved to `use-pane-rect-sync.ts`) and wrote a retro documenting the regression pattern for future reference: `docs/retro/retro-floating-pane-resize-hit-target-2026-07-27.md`.

Cross-platform: this code path runs identically on Windows, macOS, and Linux (verified — no platform gating around the resize detection or the browser-pane inset; `posScale()` handles the Windows-physical-px vs. macOS/Linux-DIP coordinate difference, unrelated to this fix).

## Test plan

- [x] Changeset added
- [ ] Manual test — will run `task dev` for the user to verify the grab feel directly before merge

### DIFF
--- a/.changesets/1785201955-fix-floating-pane-restore-resize-hit-target-from-4px-to-8px-sfnn.md
+++ b/.changesets/1785201955-fix-floating-pane-restore-resize-hit-target-from-4px-to-8px-sfnn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): restore resize hit-target from 4px to 8px
--- a/docs/architecture/ARCHITECTURE_FLOATING_PANE_DOCKING_2026_05_30.md
+++ b/docs/architecture/ARCHITECTURE_FLOATING_PANE_DOCKING_2026_05_30.md
@@ -82,7 +82,7 @@ doc exists to prevent).
 - During drag, throttled `update_floating_redock_hover` (~50ms) drives the drop-target highlight.
 
 ### 2.3 Edge-resize (floater only)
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2334, opened 2026-07-28
docs(retro): pane tear-off cursor + redock reliability investigations

Agent2-asaf:
## Summary

Two investigations from live testing after the floating-pane resize-hit-target fix (#2332). Docs only, no code changes -- these lay out root cause + fix direction so the actual fixes can be scoped/reviewed separately.

### 1. Pane tear-off cursor (`retro-pane-tearoff-cursor-never-fixed-2026-07-27.md`)

Windows pane tear-off shows the OS "not-allowed" cursor instead of a "creates a new window" affordance. Not strictly a regression -- tab tear-off got the fix for this exact problem (#2175) but it was scoped to tabs only; panes were never picked up. Fix direction is clear and low-risk (port `tab-reorder.ts`'s `dragover`/`dropEffect` pattern into `TileLayout.win32.tsx`).

### 2. Redock ghost-landing-spot reliability (`retro-redock-ghost-landing-reliability-2026-07-27.md`)

Floating-pane redock no longer reliably lands where the ghost preview showed. Deeper issue -- a 7-week-old subsystem with its own architecture doc that already flagged 5 recurring structural themes after a prior regression cycle. Last deliberately-fixed state: 2026-07-05 (#1968). Several large unrelated refactors since then touch dependencies (shared RPC caller, HWND-identity timing, layout-geometry rewrite) without anyone re-running the redock checklist. Two existing, never-executed structural proposals already address the "framework level" fix the user asked for (atomic backend move; replace floaters' raw Win32-popup model) -- this retro recommends scheduling those plus a regression harness rather than another point patch.

## Next steps (not in this PR)

- Implement the pane tear-off cursor fix (small, scoped).
- Scope a spec for the redock structural work before touching code there.

### DIFF
--- a/docs/retro/retro-pane-tearoff-cursor-never-fixed-2026-07-27.md
+++ b/docs/retro/retro-pane-tearoff-cursor-never-fixed-2026-07-27.md
@@ -4,6 +4,36 @@
 **Severity:** Low — cosmetic, no functional breakage (the tear-off itself works), but reads as broken/unpolished on every single Windows pane drag.
 **Observed by:** user, while testing the floating-pane resize-hit-target fix (PR #2332).
 
+> **Update (2026-07-29):** the "Fix direction" below was implemented
+> (`fix/pane-tearoff-cursor-windows`, PR #2335) and **found insufficient in
+> live testing** — automated review at the time also flagged (P2, unaddressed
+> before this update) that a window-level `dragover` listener setting
+> `dropEffect="copy"` whenever the cursor is outside the tile layout would
+> overwrite the tab strip's own drop-target `dropEffect` for in-app tab
+> redock, a correctness bug on top of the cursor not actually landing. Root
+> cause of the insufficiency, confirmed via direct research in a follow-up
+> session: Chromium's own `IDropSource::GiveFeedback` calls `SetCursor`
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

Agent2-asaf:
Added two follow-on specs, scheduling the fixes these retros pointed at rather than leaving them as open recommendations:

- \`docs/specs/SPEC_REDOCK_FRAMEWORK_HARDENING_2026_07_27.md\` — schedules the P3 (atomic backend redock), P4 (unowned floaters), and P6 (regression harness) work that two existing docs already designed and explicitly deferred, updated with what's changed since the last functional fix (a second RPC caller, an HWND-identity timing change, a layout-geometry rewrite).
- \`docs/specs/SPEC_FLOATING_PANE_MULTI_MONITOR_TASKBAR_2026_07_27.md\` — new feature spec (separately requested): give a floating pane a real taskbar/Dock entry when it's dragged to a different monitor than its main window. Researched Windows/macOS/Linux conventions; scoped the implementation phase to Windows only, since macOS/Linux multi-monitor work-area detection doesn't exist in this codebase yet. Reuses the existing \`set_taskbar_hidden()\` primitive rather than inventing new Win32 taskbar plumbing.

PR #2335, opened 2026-07-28
fix(dnd): tab + pane tear-off show correct cursor, not circle-slash (Windows)

Agent2-asaf:
## Summary

Both tab tear-off and pane tear-off showed the OS "not-allowed" circle-slash cursor on Windows when dragged **outside the app window** — instead of an affordance indicating a new window will be created.

## Root cause (revised from this PR's first commit)

The first commit here (and tabs' own earlier fix, #2175) used `dropEffect = "copy"` on `dragover`. That only reaches the OS cursor while the pointer is still over this app's own HTML content. The moment a tear-off drag crosses the window's own boundary onto the desktop or another window, no `dragover` event fires here at all — Windows' OLE drag-feedback takes over and shows its default no-drop cursor regardless of what `dropEffect` was last set to. Confirmed live: both tabs and panes still showed the circle-slash once dragged past the window edge, despite the dropEffect fix being in place for both.

`set_drag_cursor`/`restore_drag_cursor` already existed in `agentmux-cef/src/commands/drag.rs`, fully wired end-to-end (IPC dispatch, frontend wrapper) — but dead code, called from nowhere, using a known-broken implementation (`SetSystemCursor(OCR_NO)`). An unmerged March 2026 branch (`agenta/fix-tab-drag-cursor`) had already diagnosed why: OLE's `IDropSource::GiveFeedback` caches its own no-drop cursor handle on the *first* `GiveFeedback` call of a drag session, so replacing the system cursor resource afterwards never reaches that cached handle. That branch's final commit fixed it with a 2ms polling thread that repaints `SetCursor(IDC_CROSS)` out from under whatever last set it — but the branch was never merged.

## Fix

- Ported the working polling-thread implementation into `set_drag_cursor`/`restore_drag_cursor`.
- Wired both draggable sources — `droppable-tab.tsx` (tab tear-off) and `TileLayout.win32.tsx` (pane tear-off) — to call `setDragCursor()` on drag start and `restoreDragCursor()` on drop, each with a `dragend`-based safety net for the swallowed-dragend failure mode this codebase already guards against elsewhere (Win11 snap-layouts/Alt+Tab interrupting a drag).
- Kept the `dropEffect` fixes from the first commit — still correct for the in-window-but-outside-the-drop-target case, just not sufficient on their own for tear-off past the window edge.

## Test plan

- [x] `cargo check -p agentmux-cef` clean
- [x] Changeset added
- [ ] Manual test — will run `task dev` for the user to verify both tab and pane tear-off cursors directly before merge

### DIFF
--- a/.changesets/1785203678-fix-layout-pane-tear-off-shows-correct-cursor-on-windows-4wsr.md
+++ b/.changesets/1785203678-fix-layout-pane-tear-off-shows-correct-cursor-on-windows-4wsr.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): pane tear-off shows correct cursor on Windows
--- a/agentmux-cef/src/commands/drag.rs
+++ b/agentmux-cef/src/commands/drag.rs
@@ -249,34 +249,80 @@ pub fn get_mouse_button_state() -> Result<serde_json::Value, String> {
     Ok(serde_json::json!(false))
 }
 
-/// Replace the system no-drop cursor with a crosshair during drag.
+/// Whether the cursor-override thread should keep running.
+#[cfg(target_os = "windows")]
+static DRAG_CURSOR_ACTIVE: std::sync::atomic::AtomicBool = std::sync::atomic::AtomicBool::new(false);
+
+/// Start a cursor-override thread that calls `SetCursor(IDC_CROSS)` every 2ms,
+/// so the cursor reads as "creates something new" instead of the OS's default
+/// no-drop circle-slash while a tab/pane tear-off drag is outside the window.
+///
+/// # Why a thread instead of SetSystemCursor
+///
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

Agent2-asaf:
Closing without merging — confirmed insufficient in live testing. Chromium's own `IDropSource::GiveFeedback` calls `SetCursor` continuously during the OLE drag session this fix runs on top of, based on its own drop-target hit-testing; no cursor-polling-thread interval reliably out-races that.

A follow-up session explored eliminating the OLE session entirely (native Pointer Events + `setPointerCapture`, no `draggable()`), confirmed that mechanism does structurally fix the cursor, but found the full rewrite (needed to replace pragmatic-dnd's drop-zone hit-testing once the native drag session is gone) introduces real regressions of its own. That attempt was also shelved in favor of a narrower follow-up. Full history: #2354 (retro) and #2337 (design spec, updated with outcome).

PR #2336, opened 2026-07-28
feat(agent): surface a Restart action when an agent process goes unresponsive

Agent1-asaf:
## Summary

Implements the minimum-viable fix scoped in `docs/reports/REPORT_WORKING_STATE_REGRESSION_AND_STUCK_QUESTION_PANEL_2026_07_27.md` §4, closing the gap directly exposed by this session's Agent2 incident: reaching `AgentHealth::Dead` was previously diagnostic-only (a WPS `agenthealth` event with zero frontend subscribers) — a pane whose process went unresponsive just sat there until a human happened to notice and manually reopened it.

## Changes

- **New `FailureClass::Unresponsive`** (`agentmux-srv/src/agents/failure.rs` + `frontend/types/gotypes.d.ts`), published as a full `AgentFailure` the moment `HealthMonitor` transitions into `Dead` — durable meta persist then live WPS push, mirroring the existing exit-classification pattern exactly. `retryable: false`: the process is alive but wedged, not exited, so a plain "re-send the last message" retry wouldn't reach anything.
- **New "Restart" action** in the failure row (`failure-accessory.ts`), wired to `ControllerResync{forcerestart: true}` — the exact mechanism `forceControllerRefresh` already used internally for the post-login stale-process case, now exposed on `useAgentControllerStatus`'s public interface so the failure row can reuse it.
- **Self-heal handling** (identified during research, closed before shipping): `compute_health()`'s silence-based Dead check has no hysteresis — late output arriving just after the 120s threshold tripped flips health straight back to `Healthy`. Without handling this, a stale "Restart" button would linger over a process that's actually fine again. A `Dead -> anything else` transition now publishes a `data: None` clearing event on the same channel; the frontend only clears the row if it's currently classed `unresponsive`, so it can never clobber an unrelated concurrent failure (e.g. auth).
- **`HealthMonitor` now takes `wstore`/`event_bus`**, threaded through all three call sites (`persistent.rs`, `subprocess/mod.rs`, `acp.rs`) — previously store/event-bus-free by design; needed only for this publish/clear.

**Deliberately NOT built:** auto-restart without waiting for a click. Per the report's own scoping, that needs a product decision (when is it actually safe to assume "nothing to lose"?), not just engineering — left as an explicit follow-up.

## Test plan

- [x] `cargo check -p agentmux-srv --tests` clean
- [x] New backend tests `dead_transition_publishes_unresponsive_failure` / `dead_recovery_clears_the_unresponsive_failure` (`health.rs`) — drive `Dead` via `record_error(Fatal, ...)` since the 120s silence branch isn't practical to wait out in a test; verify publish/clear via `Broker::read_event_history`
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 1701 passed / 0 failed / 4 ignored (one unrelated pre-existing test-parallelism flake in `agent_session::tests` reproduced once, confirmed passes in isolation and on a clean re-run — documented pre-existing "process-global read cache leaks" issue, unrelated to this change)
- [x] New frontend tests: `failureToRow`'s `unresponsive` case (Restart primary, no plain Retry offered), two `useAgentFailure.ts` tests (null-data event clears an `unresponsive` row; does NOT clear an unrelated concurrent one)
- [x] `npx vitest run app/view/agent app/store/agent-pane-state app/window` — 70 files / 980 tests passed, 0 regressions


a5af:
@codex review

PR #2337, opened 2026-07-28
docs(spec): native pointer-capture drag for tab+pane tear-off (Windows)

Agent2-asaf:
## Summary

Spec for a performance-first replacement of HTML5/OLE drag-and-drop on the Windows tab+pane tear-off gesture, per explicit direction that performance takes priority over engineering cost for this work.

## Why this instead of the cursor-polling-thread fix on #2335

Investigated the current architecture in depth: Windows' OLE `IDropSource::GiveFeedback` genuinely owns cursor painting for the entire drag session, continuously, including outside the app window. The polling-thread fix on #2335 fights that by brute-force repainting — live testing found it insufficient. Rather than keep fighting OLE, this spec eliminates the OLE drag session entirely: drive the whole gesture from raw Pointer Events with `setPointerCapture`, which keeps delivering events (including outside the window) with **no** OS-level drag negotiation — nothing to override, because nothing paints a competing cursor in the first place.

This isn't a new idea — an unmerged May 2026 plan (`PLAN_TAB_TEAROFF_PHASE1_WIN32_2026-05-07.md`, status "READY TO IMPLEMENT") already spiked and validated the core mechanism for tabs. This spec updates that plan for ~3 months of drift, extends scope to panes (May plan deferred them), and removes its "degrade cross-window tab drop for v1" compromise by reusing infrastructure that didn't exist in May (the `WH_MOUSE_LL` hook's existing hit-test) instead of accepting a regression.

## Performance case

Eliminates OLE/COM drag-session overhead and Blink's HTML5 drag-event pipeline per mouse-move, replacing both with direct native message handling — the same pattern already proven in this codebase for floating-pane window-move (`Win32BeginMoveTask`, documented "zero per-move IPC"). Also structurally eliminates the whole class of Win11 swallowed-dragend bugs this codebase has repeatedly built safety nets around, since `pointerup`/`pointercancel` don't have that failure mode.

## Scope

Windows only — macOS/Linux don't have the cursor bug (`preventUnhandled` already avoids it there) and have no equivalent native-loop precedent to build the performance case on.

This is a spec/planning doc only — no code changes in this PR.

### DIFF
--- a/docs/specs/SPEC_NATIVE_POINTER_DRAG_TEAROFF_2026_07_28.md
+++ b/docs/specs/SPEC_NATIVE_POINTER_DRAG_TEAROFF_2026_07_28.md
@@ -1,6 +1,20 @@
 # Spec: replace HTML5/OLE drag with a native pointer-capture drag loop for tab + pane tear-off (Windows)
 
-**Status:** proposed
+**Status:** attempted 2026-07-29, shelved — see
+`docs/retro/retro-native-pointer-drag-tearoff-shelved-2026-07-29.md` for the
+full outcome. Summary: the core mechanism (§ Architecture below) does work —
+tab and pane tear-off both function, and the circle-slash cursor is
+structurally eliminated (no OLE session at all). But implementing it
+required rebuilding pragmatic-dnd's in-window drop-zone hit-testing and drag-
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2338, opened 2026-07-28
fix(auth): fast-fail sends from a pane already known-unauthenticated

AgentA-asaf:
## Summary

- User report: while a pane was showing the mount-time "Log in" bar (`canRetry() === true`), typing a message sent it anyway — the button disappeared, the pane went into a long "Working…" phase, and only after a real network round-trip did it surface `Not logged in · Please run /login`. Nothing should have needed that round-trip since the UI already knew the pane was logged out.
- Root cause (full writeup in `docs/retro/retro-send-while-unauthenticated-2026-07-28.md`): the composer had zero auth gating — `canRetry()`/`authStatus()`/`agentReady()` exist but were never referenced anywhere in the send path — and a controller already gets registered unconditionally at agent-launch time, so the backend's identity-injection gate (which only validates credential *shape*, not *validity*) lets the spawn through on a stale/`needs_reauth` credential.
- Fix: `deliverToBackend` (`useAgentCommands.ts`, the single choke point both the immediate-send and held-message-flush paths converge on, after slash/bang-command interception) now checks `opts.canRetry()` first and fails immediately with the same `PendingMessageRejected` + `TurnStartFailed` cleanup already used for synchronous `AgentInputCommand` rejections — no CLI spawn, no "Working…" delay.
- Scope note: the retro's recommendation #2 (backend-side fast-fail in `agentmux-srv`'s `input.rs`, so every caller of `AgentInputCommand` gets this protection, not just this composer) is intentionally left as a follow-up — this PR ships the complete, tested fix for the reported user-facing symptom.

## Test plan

- [x] New unit tests in `useAgentCommands.test.ts`: never calls `AgentInputCommand` when `canRetry()` is true (reverts `turnPhase` to Idle, surfaces an auth notice); does NOT fast-fail an already-in-flight turn's held-message flush if `canRetry()` only flips true mid-turn.
- [x] Verified both new tests actually catch the regression (temporarily reverted the guard to a no-op, confirmed the first test fails with the real `AgentInputCommand` payload, restored the fix).
- [x] `npx vitest run frontend/app/view/agent/hooks/useAgentCommands.test.ts` — all 4 tests pass.
- [ ] Live verification in a running dev instance: open a pane showing "Log in", type a message, confirm immediate auth-notice feedback instead of a "Working…" hang.


a5af:
@codex review

AgentA-asaf:
Fixed in 9be87b6a7 — gated the fast-fail guard on \`initiatesTurn\` so a held message already accepted into an active turn's queue is no longer silently ejected if \`canRetry()\` flips mid-turn; it now falls through to the normal \`AgentInputCommand\` attempt instead. Added an assertion to the existing held-flush test to actually check delivery (not just \`turnPhase\`), and verified it fails against the prior unconditional guard before restoring the fix.

a5af:
@codex review

AgentA-asaf:
Fixed in 61485b848 — real gap in both directions, thanks for catching it:

- **relogin() window**: added `loginWaiting()` to the guard alongside `canRetry()`, since `canRetry` is cleared synchronously on click, well before the OAuth attempt (up to 5 min) actually resolves.
- **/login never clears canRetry**: threaded `notifyControllerHealthy` (already built for "the agent became healthy through a different path") into `SlashCommandContext` and call it from both of /login's success branches.

Checked `useGlobalLogin()`/`loginViaTerminal()` too — both are wired only to the failed-turn failure banner, which is mutually exclusive with the mount-time `canRetry` bar (one requires a turn to have already started and failed; the other means Phase 2 bailed before any turn could run), so they don't share this gap.

All three test changes were verified to fail against the prior behavior before restoring each fix.

a5af:
@codex review

AgentA-asaf:
Fixed in c4022726d — real gap, confirmed by re-reading useGlobalLogin(): unlike relogin()/loginViaTerminal(), it only ever touched a local seedInFlight boolean, never the reactive loginWaiting signal the fast-fail guard checks. Added setLoginWaiting(true)/false around the same try/finally seedInFlight already guards, mirroring the other two recovery paths exactly.

This hook had no prior unit tests, so I added useAgentControllerStatus.test.ts with a deferred-promise test that inspects loginWaiting() mid-flight (before registerSeededAccount resolves) — verified it fails against the prior behavior before restoring the fix.

a5af:
@codex review

AgentA-asaf:
Fixed both in 92ad461a2:

**reagent P0** — real regression: relogin()/useGlobalLogin()/loginViaTerminal() only cleared loginWaiting in their trailing finally, which runs AFTER onRecovered() synchronously resends the failed turn. Moved setLoginWaiting(false) to run in every success branch immediately before onRecovered()/onReady() fire.

**codex P1** — also real: neither canRetry nor loginWaiting reflects a live mid-turn 401/403 (that's the separate failure-banner state.failure/data.code==="auth" mechanism). Tried to check it live inside deliverToBackend first, but TurnStart (dispatched by the caller before every send) unconditionally clears state.failure — so a live re-check always sees it already gone, for both a genuine unresolved failure AND a legitimate post-recovery auto-retry. Fixed by having handleSendMessage capture the failure state BEFORE dispatching TurnStart (mirroring how wasAlreadyWorking is already captured the same way) and threading it through as hadLiveAuthFailure; onRecovered now explicitly dispatches FailureCleared before its auto-retry so that capture correctly reads false for the legitimate-recovery case.

All new/changed test assertions verified to fail against the prior behavior before restoring each fix.

a5af:
@codex review

AgentA-asaf:
Fixed in 74d0f3b9a — good catch, real UX regression: the rejection was silently wiping the failure banner (and its "Login Again"/"Use existing login" actions) with nothing to replace it, since TurnStart unconditionally clears state.failure before the guard even runs.

Changed sendMessage/deliverToBackend's third param from a boolean to the captured AgentFailure itself, and on rejection re-dispatch FailureObserved with that same data (after TurnStartFailed, so it doesn't hop through a transient Done state) — restoring the exact banner and its recovery actions instead of a generic notice that referenced a "Log in" button not even shown in this context.

a5af:
@codex review

AgentA-asaf:
Fixed the P1 in 7fc1d5386 — real gap: /login's success path called notifyControllerHealthy() but never cleared a live state.failure "auth" row, unlike relogin()/useGlobalLogin()/loginViaTerminal() (which all clear it via onRecovered→FailureCleared). A stale pre-existing failure would survive a successful /login and get re-captured by the caller's next send, reproducing the exact bug this PR fixes. Added clearAuthFailure to SlashCommandContext, called alongside notifyControllerHealthy in both success branches. Verified the new assertion fails without the fix.

On the P2 (discarded prompt text on rejection): confirmed real — AgentFooter.handleSend clears the textarea optimistically before the async send resolves, and PendingMessageRejected then removes the only other copy. Deliberately deferring this one: fixing it properly means restoring rejected text into the composer (or a retry queue), which is a separate composer-level UI feature (AgentFooter owns that state, not this guard), not a guard-logic fix. Flagging as a follow-up rather than scope-creeping it into this PR.

a5af:
@codex review

AgentA-asaf:
Looked into this P2 carefully (both codex's original flag and reagent's re-raise) before deciding whether to change anything — the claim that the message is "permanently lost... must be retyped" doesn't hold up against the actual composer code:

- `AgentFooter.tsx:673` (`handleSend`) pushes the typed message into `sentHistory` **unconditionally**, immediately after the fire-and-forget `onSendMessage` call and *before* the async send can even reach the guard that might reject it.
- `handleKeyDown`'s ArrowUp handling (`AgentFooter.tsx:790-795`) recalls the most recent `sentHistory` entry into the (now-empty) composer on the very next ArrowUp press. `onRecallLatestQueued()` is checked first but returns `null` here (nothing was queued — this was an idle send, not a held/queued one), so it falls straight through to the sentHistory recall.

So the rejected text is one ArrowUp away, exactly like recalling any other previously-sent message — the same shell-history convention Claude Code's own CLI uses. It's not literally in the composer, but it's not destroyed either, and no new retry-queue mechanism is needed since this one already covers it. Not making a code change here since I couldn't find an actual gap once I traced it through — happy to revisit if there's a scenario I'm missing.

AgentA-asaf:
Fixed in 57ad68c96 — verified real: confirmed against failure-accessory.ts that only the generic "Retry now" action has a disabled binding, so useGlobalLogin() and relogin()/loginViaTerminal() genuinely can run concurrently. Replaced the three functions' independent setLoginWaiting(true/false) calls with a shared activeRecoveryFlows counter — loginWaiting only clears when it hits zero, so whichever flow finishes first no longer clears it out from under a still-running sibling flow. New regression test starts loginViaTerminal, lets a concurrent useGlobalLogin() fully resolve, and asserts loginWaiting() stays true until loginViaTerminal itself finishes — verified it fails against the prior independent-boolean behavior before restoring the fix.

(The other two inline comments on this round are the same pre-existing ones already replied to earlier in the thread — the auth-failure-preservation one was fixed in 74d0f3b9a, and the blocked-prompt one was investigated and found to already be covered by AgentFooter's sentHistory/ArrowUp recall, no code change needed.)

a5af:
@codex review

AgentA-asaf:
Fixed in 576da3644 — real, direct consequence of the shared-counter fix in 57ad68c96: a flow that just succeeded shouldn't have its own onRecovered auto-retry blocked by a *different*, unrelated flow's unconfirmed state. Threaded a trustedAfterRecovery flag from onRecovered's retryLastTurn call through to deliverToBackend, bypassing only the loginWaiting() check (canRetry()/authFailureToPreserve are already false/null by the time onRecovered fires, so nothing else needs bypassing). The generic failure-banner "Retry now" button is unaffected — it calls retryLastTurn with no argument, keeping the normal guard.

Verified the new test fails against the prior unconditional loginWaiting() check before restoring the fix.

a5af:
@codex review

AgentA-asaf:
Fixed in a52b2b715 — real, verified: turnPhase (wasAlreadyWorking) and canRetry()/loginWaiting() are tracked by independent signals with no invariant enforcing "active turn implies good auth," so a controller reporting an active turn while the mount-time check has also shown "Log in" is genuinely reachable (e.g. a reopened pane whose backend turn is still streaming while its own mount-time auth check independently reports auth_failed).

Captured canRetry()||loginWaiting() at the moment a message is pushed onto heldQueue, and flushHeldMessages() now rejects exactly the items that were already known-bad at queue time — without touching turnPhase, since a real turn is still genuinely active for whatever else put the pane in the held-message branch. The existing "don't drop a held message just because auth flips bad AFTER it was queued" test still passes (it queues at canRetry=false, only flips true later) — this fix is specifically for the opposite ordering.

Verified the new test fails against the prior unconditional-flush behavior before restoring the fix.

a5af:
@codex review

AgentA-asaf:
Fixed in 6a38b9d33 — real and important: confirmed /login's handler had zero references to forceControllerRefresh while all three OTHER recovery paths (relogin/useGlobalLogin/loginViaTerminal) call it before declaring success. send_message only spawns a fresh process when one isn't already running, so a pane whose controller was already alive would stay on the stale credential after a successful /login — the next message bypasses every guard this PR added (all correctly cleared by then) and still reaches that stale process.

Exposed the existing forceControllerRefresh (already used internally by the other three paths) on useAgentControllerStatus's public interface, threaded it through SlashCommandContext, and call it in both of /login's success branches before declaring success, matching the other paths' ordering.

Verified the new assertion fails against the prior behavior before restoring the fix.

a5af:
@codex review

AgentA-asaf:
Addressed codex's P1 from the eighth re-review (against `6a38b9d33`): `useControllerStatusEvents`' `onTurnActive` was calling `notifyControllerHealthy()` for **every** controllerstatus event on the pane, including idle heartbeats — not just events proving an active turn. Verified directly against `agent-view.tsx`/`useControllerStatusEvents.ts`: a persistent controller left alive from before a just-failed recovery attempt (`canRetry` just set `true` to show "Login Again" again) still emits idle `{is_agent_pane: true}` events, and one arriving would silently clear that recovery's own `canRetry=true` — letting the very next message bypass the fast-fail guard and reach the still-known-bad process. That's the exact bug class this PR exists to close.

Fix (`9fe8cbddd`): split the hook's callback — `onTurnActive` still fires on every event (TurnPhase reconciliation legitimately wants idle transitions too), but a new `onActiveTurnConfirmed` fires ONLY when `turn_active` is genuinely `true`, and `agent-view.tsx` now calls `notifyControllerHealthy()` from that one. Mirrors the file's existing `deriveTurnActive`/`didTurnJustEnd` pattern — pure wire-contract predicates with direct hook-level tests. Verified the new assertions fail against the prior unconditional-fire behavior before restoring the fix.

Also merged `origin/main` (`cd50c005e`) to clear the `CONFLICTING` mergeable state — main had independently landed PR #2336's own `forceControllerRefresh` exposure (for an "unresponsive agent → Restart" action, with an added `context` param) after this branch forked. Combined both call sites' needs into one signature; `d2d761ffd` cleans up a duplicate-object-key merge artifact from that resolution (caught via `tsc`, not by any reviewer — no functional impact, just an accidental double-insert).

`cargo check --workspace --tests`, `tsc --noEmit`, and the full frontend/agent vitest suite (843 tests) are all green locally. CI didn't report any checks against `6a38b9d33` at all (confirmed via check-suites/check-runs API, not just `gh pr checks`) — watching to see if it picks up on this push.

a5af:
@codex review

AgentA-asaf:
Addressed two new P1s from codex against `d2d761ffd` (ninth re-review) — both verified directly against the current code before fixing, both genuinely new (checked timestamps against the `git log` push time to rule out re-anchoring):

**1. `agent-view.tsx:1154` / `useAgentCommands.ts` — local commands silently discard a live auth failure.** `handleSendMessage` captures a live "auth" `state.failure` as `authFailureToPreserve` before `TurnStart` clears it, so a rejected turn-initiating send can restore it. But `sendMessage`'s LOCAL command paths (bang commands, and any slash command the dispatcher marks "handled" — `/help`, `/clear`, etc.) returned early without ever using it — the failure just vanished, and the next *normal* send saw a null `authFailureToPreserve` with `canRetry`/`loginWaiting` both still false (mid-turn auth failures never touch either), sailing straight through the guard to the still-known-bad credential.

Fixed by restoring it (`FailureObserved`) at every local-command return point, unless the command itself resolved it — tracked by wrapping `ctx.clearAuthFailure` for the slash-command dispatch (only `/login`'s success path calls it; restoring a stale failure over a credential `/login` just fixed would be wrong).

**2. `useAgentCommands.ts:477` — `/login`'s own poll isn't tracked as an in-flight recovery.** Unlike `relogin()`/`useGlobalLogin()`/`loginViaTerminal()` (all wired to the shared `loginWaiting` counter via `beginRecoveryFlow`/`endRecoveryFlow`), `/login`'s slash-command handler never touched it. A second message sent while `/login`'s up-to-5-minute poll is in flight got held with `authWasKnownBadAtQueueTime: false` (both signals false), so a `/login` that ultimately times out flushed that held message straight to the still-known-bad controller.

Fixed by exposing `beginRecoveryFlow`/`endRecoveryFlow` on `SlashCommandContext` and having `/login` register itself for the duration of its attempt (`finally`-guarded, mirroring the other three flows).

Verified both: reverted each fix, confirmed the new tests fail with the expected assertion errors, restored. `tsc --noEmit` and the full frontend/agent suite (844 tests) are green.

a5af:
@codex review

AgentA-asaf:
Addressed three findings from codex against `12f2a923e` (tenth re-review) — all verified against the actual current code, all genuine:

**1. (P1) `login.ts:145` — restarting the controller from `/login` could kill an active turn.** `forceControllerRefresh()` sends `ControllerResyncCommand` with `forcerestart:true`; traced into `agentmux-srv`'s `resync_controller` and confirmed `needs_replace` unconditionally stops the existing controller process when forced (`blockcontroller/mod.rs`). Since `sendMessage` intercepts slash commands *before* the busy/held-queue branch, `/login` submitted while a turn is actively streaming would run immediately and — on success — kill that turn mid-stream, discarding its work.

Fixed: `/login` now checks `ctx.isTurnActive()` (same `turnPhase` the fast-fail guard reads) and skips the restart when a turn is active. The refreshed credential takes effect on the pane's next natural respawn instead — narrower than the original stale-controller bug this call exists to fix, and strictly safer than killing live output.

**2. (P1) `login.ts:153` — a failed refresh still declared the pane healthy.** `forceControllerRefresh()` swallows its own RPC failures internally (best-effort, warns, resolves normally) — every caller ignored the return value, so `/login` cleared every fast-fail guard (`notifyControllerHealthy`/`clearAuthFailure`) even when the refresh itself failed and the controller was still on the stale credential.

Fixed: `forceControllerRefresh` now returns whether the resync actually succeeded (`Promise<boolean>`); `/login`'s `finalizeLoginSuccess` only clears the guards when it did (or was intentionally skipped per #1), returning an error otherwise.

**3. (P2) `useAgentCommands.ts:627` — the auth guard could go stale across an async gap.** It ran once, before the runtime-args `SetMetaCommand` round-trip — a real window another recovery flow or a mid-turn auth failure could land in before the actual send.

Fixed: extracted the guard into `checkAuthGuard()` and re-invoke it immediately before `AgentInputCommand`, not just once at the top.

Verified all three: reverted each fix individually, confirmed the new tests fail with the expected assertion errors, restored. `tsc --noEmit` and the full frontend/agent suite (849 tests, 1 pre-existing unrelated flaky timeout confirmed passing in isolation) are green.

a5af:
@codex review

AgentA-asaf:
Addressed a genuine P1 from codex against \`461a822d3\` (eleventh re-review) — verified directly against the actual code, and it's a real one that would have quietly defeated last round's fix.

**\`useAgentCommands.ts:336\` — \`isTurnActive()\` always read \`true\`, even on a genuinely idle pane.** It read \`paneSnapshot(blockId)?.turnPhase\` live. But \`handleSendMessage\` (agent-view.tsx) dispatches \`TurnStart\` *optimistically* before \`sendMessage\` ever runs — so by the time \`buildCommandContext()\` executed, \`turnPhase\` was already \`"Submitting"\` for the completely ordinary case of typing \`/login\` on an idle pane. \`isTurnActive()\` therefore reported \`true\` unconditionally, which meant \`finalizeLoginSuccess\` (from last round's fix) *always* skipped \`forceControllerRefresh()\` — including for the most common path — so a refreshed credential would never reach an already-running but genuinely idle stale controller. That's exactly the bug \`forceControllerRefresh\` was added to \`/login\` to fix, silently reintroduced.

Fixed by threading \`wasAlreadyWorking\` — the snapshot \`handleSendMessage\` already captures *before* dispatching \`TurnStart\` — through \`buildCommandContext(wasAlreadyWorking)\` into \`isTurnActive\`'s closure, instead of re-deriving it from a live (and by-then-stale) \`turnPhase\` read.

Verified: reverted to the live-read version, confirmed the new "genuinely idle pane" test fails while the "already streaming" test still correctly passes either way (proving the test actually discriminates the bug), restored. \`tsc --noEmit\` and the full frontend/agent suite (851 tests, 1 pre-existing unrelated flaky timeout confirmed passing in isolation) are green.

a5af:
@codex review

AgentA-asaf:
Addressed a genuine P1 from codex against \`97a91a688\` (twelfth re-review) — verified directly against the code.

**\`useAgentCommands.ts:352\` (via \`agent-view.tsx:1778\`) — invoking \`/login\` from the help panel during an active turn could kill it.** \`SlashHelpPanel\`'s \`onInvoke\` called \`commands.sendMessage(\`/${cmd.name}\`)\` directly, omitting \`wasAlreadyWorking\` — whose default is \`false\` regardless of whether a turn is actually active. Since round eleven's fix made \`isTurnActive()\` consume that argument verbatim (correctly fixing the composer path, which always passes it explicitly), this call site fed it a hardcoded \`false\`: invoking \`/login\` from the help panel while a turn was genuinely streaming made \`isTurnActive()\` lie, and \`finalizeLoginSuccess()\` force-restarted the controller on login success — killing that active turn and discarding its work. Exactly the failure mode round ten's fix exists to prevent, reached through a second call site I hadn't audited.

Fixed by routing this invocation through \`handleSendMessage\` — the same entry point the composer uses — which computes \`wasAlreadyWorking\` from a live pre-\`TurnStart\` \`turnPhase\` read at the moment of the click.

Note on verification: this one is a call-site wiring fix in \`agent-view.tsx\`, which has no test file in this codebase (confirmed — no \`.test.tsx\` exists for it). I verified it by direct inspection (confirmed \`handleSendMessage\` is in scope and computes the value correctly) rather than a new unit test; the downstream mechanism it now feeds correctly (\`isTurnActive()\` honoring \`wasAlreadyWorking\`) is already covered by round eleven's regression tests. \`tsc --noEmit\` and the full frontend/agent suite (851 tests, 1 pre-existing unrelated flaky timeout already diagnosed as environmental) are green.

Also did a broader audit for any other direct \`commands.sendMessage(...)\` call sites bypassing \`handleSendMessage\` — this was the only other one.

a5af:
@codex review

AgentA-asaf:
Addressed a genuine, deeper P1 from codex against \`d97469924\` (thirteenth re-review) — this one correctly identified that round ten's "skip the restart while a turn is active" fix was only a partial fix.

**\`login.ts:46\` — skipping the restart while declaring the pane healthy left the credential permanently stale.** Round ten's \`finalizeLoginSuccess\` skipped \`forceControllerRefresh()\` when a turn was active, but still called \`notifyControllerHealthy()\`/\`clearAuthFailure()\` immediately — correct only if the controller stops being stale the moment the turn ends. But persistent providers (\`agentmux-srv\`'s \`persistent.rs\`) keep the *same* controller process alive across many turns, not just one. So the skip-and-declare-healthy approach left the controller on the stale credential indefinitely, with every fast-fail guard already cleared — reproducing the exact bug \`forceControllerRefresh\` was added to \`/login\` to fix in the first place, just delayed by one turn instead of immediate.

Fixed by actually deferring the work instead of abandoning it:
- \`login.ts\`'s \`finalizeLoginSuccess\` now calls \`ctx.deferControllerRefreshUntilIdle()\` when a turn is active, and returns *without* touching \`notifyControllerHealthy\`/\`clearAuthFailure\` — the guards correctly stay up until the credential is actually confirmed refreshed.
- \`useAgentCommands.ts\` tracks the pending flag and exposes \`flushPendingControllerRefresh()\`, which runs the real \`forceControllerRefresh()\` and only then clears the guards on success (leaves them up on failure, matching the existing non-deferred error path).
- \`agent-view.tsx\` calls \`flushPendingControllerRefresh()\` from its *existing* turn-just-ended edge detector (\`trackTurnJustEnded\`) — the same signal that already drives the ambient-summary trigger — so the deferred restart runs the instant the blocking turn actually finishes.

Verified: reverted \`login.ts\`'s deferral back to round ten's skip-and-declare-healthy behavior and confirmed the new test fails; separately reverted \`flushPendingControllerRefresh\` to a no-op and confirmed both of its tests fail; restored both. \`tsc --noEmit\` and the full frontend/agent suite (854 tests, 1 pre-existing unrelated flaky timeout already diagnosed as environmental) are green.

a5af:
@codex review

AgentA-asaf:
Addressed two more findings from codex against \`0222d6714\` (fourteenth re-review) — verified both against the actual code, both genuine.

**1. (P1) \`useAgentCommands.ts:393\` — the frozen \`isTurnActive()\` snapshot went stale during \`/login\`'s own 5-minute poll.** Round eleven's fix froze \`wasAlreadyWorking\` at submission time to avoid an optimistic-\`TurnStart\` corruption — but froze the \`true\` case too. If the turn that was genuinely active at \`/login\` submission ends *during* the OAuth poll, \`finalizeLoginSuccess\` still saw the stale frozen \`true\`, deferred the refresh again after the only triggering edge (turn-just-ended) had already fired, and stranded it — the controller stays on the stale credential until some unrelated future turn happens to end.

Fixed: only \`wasAlreadyWorking === false\` stays frozen (that's the case an optimistic \`TurnStart\` actually corrupts); \`true\`/unset now fall through to a live read on every call — nothing corrupts that signal when no fresh \`TurnStart\` was dispatched for the message in the first place.

**2. (P2) \`useAgentCommands.ts:694\` — bypassing \`loginWaiting()\` for a just-succeeded recovery's auto-retry ignored a still-running sibling's own eventual restart.** \`trustedAfterRecovery\` bypassed the guard reasoning that a *different* overlapping flow's uncertainty didn't matter to this flow's own confirmed success. But that bypass's own precondition (this flow's counter decrement already happened before \`onRecovered\` fires) means \`loginWaiting()\` could only ever read true here because of a genuinely-still-active *sibling* — and \`relogin()\`/\`useGlobalLogin()\`/\`loginViaTerminal()\` all unconditionally force-restart the controller once *they* finish (none check \`isTurnActive()\` — only \`/login\`'s slash-command path does). So the bypass was only ever exercised in exactly the case where it's unsafe.

Removed \`trustedAfterRecovery\` entirely rather than leave a parameter that's provably inert-or-harmful (touched \`sendMessage\`/\`deliverToBackend\`/\`handleSendMessage\`/\`retryLastTurn\` signatures). The auto-retry now blocks like any other send while a sibling is active, and fires again once the *last* remaining flow's own \`onRecovered\` runs (\`retryLastTurn\` always resends the most recent message, so whichever flow finishes last ends up triggering it).

Verified finding 1 with a new test (reverted, confirmed failure, restored). Finding 2's fix reverts to the pre-\`trustedAfterRecovery\` unconditional \`loginWaiting()\` check — reverting that one line and confirming the existing loginWaiting-block test fails demonstrates the exact regression the removal prevents. \`tsc --noEmit\` and the full frontend/agent suite (855 tests, all green including the previously-flaky one) pass.

a5af:
@codex review

AgentA-asaf:
Addressed two more findings from codex against \`145254d73\` (fifteenth re-review) — both verified against the actual code.

**1. (P1) Held-message flush could race a pending controller restart.** \`/login\` succeeding mid-turn (deferring its restart) and \`flushHeldMessages\` draining the queue are triggered by the SAME turn-just-ended moment via two independent signals — a live controllerstatus event (fire-and-forget \`flushPendingControllerRefresh\`) vs. a reactive \`turnPhaseAtom\` effect (\`flushHeldMessages\`). Neither awaited the other, so a held message could issue \`AgentInputCommand\` concurrently with \`ControllerResyncCommand\` still stopping/replacing the controller.

Fixed: \`flushPendingControllerRefresh\` now tracks its own in-flight promise, not just a boolean flag — a caller arriving after the OTHER trigger already claimed the flag still awaits the real completion instead of seeing an already-cleared flag and treating it as "nothing to wait for." \`flushHeldMessages\` awaits it unconditionally before draining (no-op when nothing's pending/in-flight).

**2. (P2) A held message known-bad at queue time was rejected forever, even if auth recovered before the flush ran.** Trusting the frozen \`authWasKnownBadAtQueueTime\` snapshot unconditionally discarded a message that would now succeed, with no path back (composer already cleared it) — a real gap since recovery can complete (or \`/login\`'s poll can finish) before the next tool-boundary/turn-end flush.

Fixed: only reject when the item was known-bad at queue time AND is still bad now (live \`canRetry()\`/\`loginWaiting()\`). The frozen flag still gates whether to even consider rejecting — an item good at queue time is still never retroactively dropped, a separate already-settled rule — it just no longer overrides a since-recovered credential.

Verified both: reverted each fix individually, confirmed the new tests fail with the expected assertion errors, restored. \`tsc --noEmit\` and the full frontend/agent suite (857 tests) are green.

a5af:
@codex review

AgentA-asaf:
Addressed the CHANGES_REQUESTED review plus codex's findings against \`b7f6259be\` (sixteenth re-review) — read every review before proceeding per standing discipline. All three verified against the actual code.

**1. (reagent P1) Held-message flush never checked a live mid-turn auth failure.** A message queued while a turn was genuinely healthy (\`authWasKnownBadAtQueueTime: false\`) that then fails with a 401/403 — \`FailureObserved\` ends the turn without touching \`canRetry\`/\`loginWaiting\` — sailed straight through to \`AgentInputCommand\` on the exact credential that just failed, since \`deliverToBackend\`'s guard is gated on \`initiatesTurn\` (always false for a flush).

**2. (codex P1) Last round's live re-check let a still-bad message through after a FAILED recovery.** \`relogin()\`/\`useGlobalLogin()\`/\`loginViaTerminal()\`'s default \`retryAfterLogin:true\` failure path clears \`loginWaiting()\` and never sets \`canRetry()\` back to true — so both signals read false after a failure too, not just a success. My round-fifteen "re-check live signals before rejecting" fix let a known-bad message through in exactly that case.

Fixed both by rejecting whenever EITHER the item was flagged bad at queue time (now unconditional again — no live re-check) OR there's a live "auth" \`state.failure\` right now. Trades back the rare lost-message UX papercut for never sending on a credential with no live proof it's fixed — the same trade this guard makes everywhere else.

**3. (codex P2) \`onRecovered\` could clear the retry banner and lose the auto-retry for good.** It unconditionally cleared the failure banner and called \`retryLastTurn()\`, even when a sibling recovery flow was still active (correctly blocked by last round's \`loginWaiting()\` fix) — but if that sibling ultimately FAILS instead of succeeding, it never calls \`onRecovered\` (success-only), so nothing retries again, and the banner (the user's only manual path back) was already cleared.

Fixed by checking \`loginWaiting()\` first: if a sibling is still active, leave the banner and everything else untouched and bail — the user can retry manually once it settles, and if the sibling instead succeeds, its own \`onRecovered\` re-runs this same check and completes the clear+retry then.

Verified 1+2 with new/updated tests (reverted, confirmed failure, restored). Finding 3 is an \`agent-view.tsx\`-only call-site fix (no test file exists for this file in the codebase, consistent with prior rounds) — verified by direct inspection. \`tsc --noEmit\` and the full frontend/agent suite (858 tests) are green.

a5af:
@codex review

AgentA-asaf:
Addressed the CHANGES_REQUESTED review plus codex's findings against \`79acc1a54\` (seventeenth re-review) — three P1s this round, all verified against the actual code.

**1. (reagent P1) A turn can end via a path the deferred-refresh mechanism never watched.** \`useTurnLifecycle.ts\`'s \`finalizeTurn\` (\`session_end\` → \`TurnEnd\`) flips \`turnPhase\` to idle immediately, independent of the \`controllerstatus\` event stream that fed the only two drains (\`trackTurnJustEnded\`, \`flushHeldMessages\`). A fresh idle send landing right after that \`TurnEnd\` — before the lagging \`controllerstatus\` event — would pass every already-cleared guard and reach \`AgentInputCommand\` while the deferred \`ControllerResyncCommand\` still hadn't run.

Fixed: \`sendMessage\`'s idle-send path now awaits \`flushPendingControllerRefresh()\` before \`deliverToBackend\` — this branch is unconditionally "pane idle, initiating a new turn," so no turn-active check is needed there.

**2. (codex P1) Last round's fix could itself force-restart a still-active controller.** \`flushHeldMessages\` also runs at a mid-turn tool-call boundary (not just turn-end) and from the Esc-to-steer handler — both while a turn can be genuinely active. My round-sixteen unconditional \`flushPendingControllerRefresh()\` call inside it could therefore kill an in-progress turn — exactly what deferring the refresh was meant to prevent.

Fixed: gated that call on the LIVE turn state — only runs when the pane is actually idle right now.

**3. (codex P1) A reopened pane could strand the deferred refresh forever.** A pane mounting onto an already-active turn never initializes \`trackTurnJustEnded\`'s edge-detector state (deliberately, to avoid a false edge on the first live event). If \`/login\` succeeds during that pre-existing turn and it ends before any other live event arrives, the edge detector never fires — and with no held messages, nothing else would ever run the deferred refresh.

Fixed: \`agent-view.tsx\`'s existing reactive effect (already watching \`turnPhaseAtom\`) now also runs the deferred refresh whenever \`turnIdle\` is true, independent of held messages and independent of the edge-detector — it reacts to the atom's current value directly, which updates from any path.

Verified 1 and 2 with new tests (reverted, confirmed failure, restored) — also had to update an existing round-16 test that was implicitly relying on the now-gated unconditional behavior. Finding 3 is an \`agent-view.tsx\`-only fix (no test file exists for this file, consistent with prior rounds) — verified by direct inspection. \`tsc --noEmit\` and the full frontend/agent suite (860 tests) are green.

a5af:
@codex review

AgentA-asaf:
Addressed the CHANGES_REQUESTED review against \`77055856c\` (eighteenth re-review) — one P1, self-inflicted by last round's own fix, verified against the actual code.

**\`useAgentCommands.ts:636\` — the idle-send path's new \`flushPendingControllerRefresh()\` await made \`authFailureToPreserve\` stale.** \`authFailureToPreserve\` is captured by the caller (\`handleSendMessage\` in \`agent-view.tsx\`) *before* \`TurnStart\` — well before last round's new \`await flushPendingControllerRefresh()\` call ever runs. \`flushPendingControllerRefresh()\`, on success, dispatches \`FailureCleared\`. So the stale \`authFailureToPreserve\` parameter still held the OLD (now-resolved) failure, and \`checkAuthGuard\` rejected the send and re-dispatched \`FailureObserved\` with a failure the credential fix had just resolved — resurrecting the "Not logged in" banner immediately after \`/login\` fixed it.

A live re-read of \`state.failure\` at that point can't distinguish this from the ordinary case, since \`TurnStart\` already unconditionally clears \`state.failure\` regardless of whether a refresh ran — both look identical from a live read.

Fixed by having \`flushPendingControllerRefresh()\` report whether it actually ran a refresh that succeeded (\`Promise<boolean>\` instead of \`Promise<void>\`); the idle-send path now discards \`authFailureToPreserve\` specifically when that's true, since a successful refresh is newer, authoritative proof the controller is fine regardless of what was captured earlier.

Verified: reverted to passing the stale parameter through unconditionally, confirmed the new test fails with the expected assertion error, restored. \`tsc --noEmit\` and the full frontend/agent suite (861 tests) are green.

a5af:
@codex review

AgentA-asaf:
Addressed codex's two findings against \`ae01cc302\` (nineteenth re-review) — reagentx itself reported no new issues after independently retracing the full current state (LGTM), but codex caught two more, both verified.

**1. (P1) \`isTurnActive()\` trusted the frontend's \`turnPhase\`, which can transiently lie.** A premature per-round \`session_end\` can demote \`turnPhase\` to "Done" even while the backend controller genuinely still reports \`turn_active: true\` — a divergence \`useControllerStatusEvents.ts\`'s \`didTurnJustEnd\` is *deliberately* independent of \`turnPhase\` to avoid, for exactly this reason (confirmed via its own doc comment). If \`/login\`'s OAuth poll completes during that window, trusting the falsely-idle \`turnPhase\` would force-restart a controller that's still genuinely working.

Fixed: \`isTurnActive()\` now ORs in a new \`isBackendTurnActive()\` accessor, backed by \`agent-view.tsx\`'s existing \`wasTurnActive\` — the same last-confirmed-backend-\`turn_active\` state \`trackTurnJustEnded\`'s edge detector already tracks from live \`controllerstatus\` events only.

**2. (P2) The pre-send re-check missed a live mid-turn auth failure.** \`checkAuthGuard\`'s re-check (added specifically to catch changes during the \`SetMetaCommand\` round-trip) only re-read the live \`canRetry()\`/\`loginWaiting()\` accessors — neither of which a mid-turn 401/403 updates. \`authFailureToPreserve\` is a captured parameter that stays \`null\` forever if nothing was showing at the initial pre-\`TurnStart\` capture, so a NEW failure arriving during that exact window slipped through both checks.

Fixed: \`checkAuthGuard\` now also reads \`paneSnapshot(...).failure\` live at each call. Safe here (unlike the very first call, which runs right after \`TurnStart\` already cleared \`state.failure\`) because anything a live read finds at this point necessarily arrived *after* \`TurnStart\` — it can't double-count the already-captured snapshot.

Verified both: reverted each fix individually, confirmed the new tests fail with the expected assertion errors, restored. \`tsc --noEmit\` and the full frontend/agent suite (863 tests, 1 pre-existing unrelated flaky timeout already diagnosed as environmental) are green.

a5af:
@codex review

AgentA-asaf:
Addressed codex's two findings against \`2231950b7\` (twentieth re-review) — reagentx APPROVED after independently retracing the full current state; codex found two deeper extensions of the same signal-divergence class last round's fix only partially closed. Both verified.

**1. \`isTurnActive()\`'s frozen \`false\` branch never consulted \`isBackendTurnActive()\`.** Last round's OR was only added to the live-read branch. But the same premature per-round \`session_end\` that corrupts a *live* \`turnPhase\` read can ALSO make \`handleSendMessage\` capture \`wasAlreadyWorking === false\` in the first place (\`turnPhase\` already read Done/Idle at capture time), even though the backend was never actually idle. Freezing \`false\` for that branch is still correct for its own reason (an optimistic \`TurnStart\` corrupting a *live* read) — but \`isBackendTurnActive()\` can't suffer that specific corruption, since it's fed only by real backend events, never the frontend's own optimistic dispatch.

Fixed: now ORed in unconditionally, rescuing this branch without reintroducing the bug freezing it was meant to fix.

**2. \`flushPendingControllerRefresh()\` itself never checked the authoritative signal.** It trusted whichever caller invoked it (\`flushHeldMessages\`, the idle-send path, the turn-just-ended edge detector, the reactive \`turnIdle\` effect) to have already verified idle-ness — all of which can fire based on the same falsely-idle \`turnPhase\`.

Fixed by moving the check *inside* this shared function instead: it now leaves the flag pending (doesn't claim it, doesn't run the resync) while the backend still confirms activity, regardless of why it was called — closing the gap for every caller at once rather than requiring each to re-derive it.

Verified both: reverted each fix individually, confirmed the new tests fail with the expected assertion errors, restored. \`tsc --noEmit\` and the full frontend/agent suite (865 tests, 1 pre-existing unrelated flaky timeout already diagnosed as environmental) are green.

a5af:
@codex review

AgentA-asaf:
Pushed aead0b083 addressing both findings from the fourth CHANGES_REQUESTED round (against af8320b01):

**reagent P1** — `flushPendingControllerRefresh`'s old `isBackendTurnActive()` guard (`wasTurnActive === true`) treated "never confirmed either way" (a pane that mounts mid-turn, before its first live controllerstatus event) the same as confirmed-idle. Fixed by adding `isBackendTurnConfirmedIdle()` (`wasTurnActive === false` specifically) and gating the destructive force-restart on that instead — positive proof of idle required, not just absence of proof of activity. `isBackendTurnActive()` is unchanged and still used by `isTurnActive()`'s OR.

**codex P1** — `trackTurnJustEnded` updated `wasTurnActive = active` *after* calling `flushPendingControllerRefresh()`, so that function's synchronous internal gate check saw the stale pre-update value on exactly the genuine turn-end edge it exists to react to — stranding the deferred refresh on that trigger. Fixed by reordering: capture `turnJustEnded` first (needs the old value), then update `wasTurnActive`, then call the flush.

Both verified via the revert-and-confirm-fails-then-restore ritual used throughout this PR: temporarily reverted the gate to the old `isBackendTurnActive()`-based check, confirmed the new "UNCONFIRMED backend state" regression test fails with the exact expected assertion, restored, confirmed all 31 useAgentCommands tests + full 866-test frontend/app/view/agent suite pass.

New regression test added: `flushPendingControllerRefresh leaves the flag pending while the backend is still active > does NOT run the refresh while the backend state is UNCONFIRMED` — directly encodes reagent's `isBackendTurnActive() false ⇏ isBackendTurnConfirmedIdle()` distinction.

a5af:
@codex review

AgentA-asaf:
Pushed 7ded9da4b addressing both findings from the fifth CHANGES_REQUESTED round (against aead0b083):

**reagentx P1** — ACP-mode controllers (Gemini/Codex/Kimi in catalog.ts) called `health.set_active_turn(false)` on a normal `stopReason` turn-end (`acp.rs:354`) with no adjacent `publish_controller_status` — the only other publishes were on kill or process exit. Confirmed by direct comparison against `persistent.rs`'s matching normal-turn-end path (`parsed.get("type") == "result"`), which pairs `set_active_turn(false)` with a publish every time. Without the publish, the frontend's `wasTurnActive` signal (fed only by live controllerstatus events) never observed the flip for ACP agents, permanently stranding `flushPendingControllerRefresh`'s `isBackendTurnConfirmedIdle` gate after a mid-turn `/login` until an unrelated event (kill/exit) happened to republish status. Fixed by publishing the status flip inline in the stopReason branch, mirroring `persistent.rs`'s exact pattern using the already-captured `inner_clone`/`broker_clone`/`block_id_stdout`.

**codex P1** — `flushPendingControllerRefresh` returning `false` conflated two different states: "nothing was ever deferred" (safe to proceed) vs. "a refresh IS deferred but the backend hasn't confirmed idle yet, so it's still blocked" (NOT safe — the controller may still hold the stale pre-refresh credential). Both `sendMessage`'s idle-send path and `flushHeldMessages` treated both cases identically and proceeded to deliver. Fixed by having both check `controllerRefreshPendingUntilIdle` after the flush attempt and hold (rather than deliver) when it's still true. This surfaced a follow-on gap while implementing: the pane reducer's `ReconcileTurnActive` is a no-op (same state reference, no re-render) when local `turnPhase` already reads idle/Done — meaning nothing was guaranteed to re-trigger a drain once the block eventually cleared. Closed by having a successful refresh proactively drain any held messages itself (`flushHeldMessages`'s own single-flight guard makes any reentrant/concurrent call safe).

All three fixes verified via the revert-and-confirm-fails-then-restore ritual: neutralized each fix independently, confirmed its dedicated regression test fails with the exact expected assertion, restored, confirmed all 33 useAgentCommands tests + full 868-test frontend/app/view/agent suite pass (one pre-existing, previously-diagnosed environmental flake in `tool-renderers/registry.test.ts`, confirmed unrelated — passes in isolation). The ACP fix additionally verified via `cargo check -p agentmux-srv` (clean) and the full `blockcontroller` Rust test module (109 passed) — no unit test exists for the stdout-reader closure itself (a pre-existing testability gap predating this PR, not introduced by it), so this one relies on direct-inspection verification against `persistent.rs`'s established pattern rather than a new regression test.

New regression tests added:
- `holds (does not deliver) a fresh idle send while a deferred refresh is still blocked...`
- `does NOT drain the held queue while local turnPhase reads idle but the backend hasn't confirmed idle yet...`

a5af:
@codex review

AgentA-asaf:
Pushed e4282a1fd addressing all three findings from the sixth CHANGES_REQUESTED round (against 7ded9da4b):

**reagentx P1** — `sendMessage`'s idle path called `deliverToBackend` directly right after `flushPendingControllerRefresh()`, but when THAT call is the one that resolves a deferred refresh, its success path independently fires `flushHeldMessages()` in the background to drain any pre-existing held items — the two `AgentInputCommand` deliveries raced with no ordering guarantee, violating the file's own documented FIFO submission-order invariant. Fixed by converting `flushHeldMessages`'s `flushing` boolean guard into a proper in-flight-promise (`inFlightHeldFlush`, mirroring the existing `inFlightControllerRefresh` pattern) so a caller can actually *await* an already-running flush instead of getting a silent no-op, and having `sendMessage` await it whenever its own call resolved the refresh.

**codex P1** — ACP's `send_input` marked the turn active via `health_monitor` but never published the flip — only spawn, kill, and exit (fixed last round) publish `controllerstatus` for ACP-mode controllers. A stale `wasTurnActive === false` left over from an earlier turn's end was indistinguishable from genuine current idleness, letting `flushPendingControllerRefresh` force-restart a controller that had just started a *new* turn, killing it. Fixed by calling `publish_status()` right after `mark_turn_active_returning_was_active()`, mirroring `persistent.rs`'s `send_message`/`send_user_message`.

**codex P2** — a message held because its deferred refresh was still blocked discarded the caller's captured `authFailureToPreserve`, recording only `canRetry()`/`loginWaiting()` (which a mid-turn auth failure never sets). Since `TurnStart` already cleared the live failure by queue time, a subsequently *failed* refresh left the item looking clean on every check `flushHeldMessages` had, delivering it to the still-bad controller. Fixed by carrying `authFailureToPreserve` on the held item and restoring its recovery banner (`FailureObserved`) when rejected because of it.

All three fixes verified via the revert-and-confirm-fails-then-restore ritual. The ACP fix additionally has a real regression test this time (`send_input_publishes_the_turn_active_flip`, using a real `wps::Broker` + `read_event_history` to assert the published status) — full 866-870-test frontend/app/view/agent suite passes (35/35 in `useAgentCommands.test.ts`, two new regression tests: FIFO-order and auth-failure-preservation), plus 110 Rust `blockcontroller` tests pass. One pre-existing, previously-diagnosed environmental flake in `tool-renderers/registry.test.ts` recurred (confirmed unrelated, passes in isolation, same as prior rounds).

a5af:
@codex review

AgentA-asaf:
Pushed e5c778f1a addressing the codex P2 finding from the seventh review round (against e4282a1fd, which reagentx approved).

**codex P2** — when a message is sent while the agent is already busy and `canRetry()`/`loginWaiting()` is ALSO already true at that exact moment, the message used to be queued into `heldQueue` anyway, deferring its inevitable rejection until some later trigger (a tool-call boundary or turn-end) happened to call `flushHeldMessages`. A tool-less or stuck turn could leave the message sitting in the "send now" panel with no feedback indefinitely — recreating the non-immediate-feedback problem this PR (retro-send-while-unauthenticated) exists to fix, since `flushHeldMessages` would reject it unconditionally anyway once it finally ran. Fixed by rejecting immediately at push time instead.

Scoped deliberately to the busy-path push only, not the blocked-deferred-refresh push: `canRetry()`/`loginWaiting()` being true there can mean the SAME in-flight recovery that hold is already waiting on, so immediate rejection there would be premature and would defeat the whole point of the hold-and-retry mechanism.

Two existing tests needed updating: their held-then-rejected-at-flush scenario is no longer reachable via the busy path post-fix (the message is now rejected before ever entering `heldQueue`). Repointed them at the still-reachable blocked-refresh push site, which preserves the underlying frozen-flag invariant they exist to pin.

Verified via the revert-and-confirm-fails-then-restore ritual. All 35 `useAgentCommands.test.ts` tests pass, full 870-test `frontend/app/view/agent` suite passes cleanly (no flakes this run).

a5af:
@codex review

AgentA-asaf:
Pushed 83296e089 addressing the codex P2 finding from the eighth review round (against e5c778f1a, which reagentx approved).

**codex P2** — `send_input`'s `self.publish_status()` call (added two rounds ago) ran BEFORE attempting `tx.try_send(req)`, along with the `mark_turn_active_returning_was_active()` call and conditional watchdog spawn. If the enqueue itself fails (channel full, or the stdin writer task already exited because the process died), no turn actually starts, but the health/status state claimed one did anyway — a rejected prompt could re-promote the frontend to a working state and clear auth-recovery UI via `notifyControllerHealthy`, and the health watchdog would be armed for work that never happened. Fixed by moving the enqueue attempt first, only marking active + spawning the watchdog + publishing after `try_send` confirms success.

Note: this exact ordering pattern (mark/publish before the send attempt) already exists in `persistent.rs`'s `send_message`/`send_user_message` — I mirrored it there in an earlier round per explicit review guidance. This fix only touches the ACP code this PR owns; `persistent.rs`'s pre-existing instances are out of scope here.

New regression test: `send_input_does_not_mark_or_publish_turn_active_when_enqueue_fails` — drops the mpsc receiver so `try_send` fails with `Closed`, asserts neither the health monitor nor a published controllerstatus event claim the turn is active. Verified via the revert-and-confirm-fails-then-restore ritual (temporarily restored the old ordering, confirmed the new test fails with the exact expected assertion, restored). Full 111-test Rust `blockcontroller` module passes; TypeScript surface is untouched this round so no frontend re-run was needed beyond the prior round's already-green 870-test suite.

a5af:
@codex review

AgentA-asaf:
Pushed 2a7d42137 addressing both findings from the ninth review round (against 83296e089).

**reagentx P1** — `flushPendingControllerRefresh` cleared `controllerRefreshPendingUntilIdle` before awaiting `opts.forceControllerRefresh()`, and never restored it (or any other guard) when that refresh FAILED. `forceControllerRefresh`'s failure path only logs a warning — it never sets `canRetry()`/`loginWaiting()`/`state.failure` — so once a deferred refresh failed, every guard in `checkAuthGuard` read clean and the very next fresh idle-sent message was delivered straight to the still-stale controller, reproducing the exact doomed "Working…" send this whole PR chain exists to prevent. Fixed by re-arming `controllerRefreshPendingUntilIdle` on failure so the next trigger retries instead of treating it as resolved — every existing caller already handles "still pending" correctly (hold, don't deliver), so this closes the gap without inventing a new signal.

**codex P2** — the previous round's ACP fix (mark/publish only after the stdin enqueue succeeds) introduced a genuine race: a fast-responding ACP agent could have the stdout reader task observe and publish the resulting `stopReason` (`turn_active: false`) on a different tokio worker *before* `send_input`'s own thread reached the mark, which would then overwrite it back to `true` — leaving a completed turn incorrectly reported as active. Fixed by restoring the original ordering (mark+publish BEFORE the enqueue attempt, which guarantees happens-before against any possible response), while keeping the prior round's enqueue-failure fix via an explicit rollback — conditioned on `!was_active` so a mid-turn steering send's own enqueue failure never clobbers a turn genuinely active from an earlier successful send.

Both fixes verified via the revert-and-confirm-fails-then-restore ritual. Two new Rust regression tests replace the now-superseded enqueue-failure test: `send_input_rolls_back_turn_active_when_enqueue_fails` (asserts the *latest* published status reflects the rollback) and `send_input_does_not_roll_back_an_already_active_turn_on_a_failed_steering_send` (the `!was_active` guard specifically). Full 112-test Rust `blockcontroller` module passes, all 35 `useAgentCommands.test.ts` tests pass (one new test strengthening the previously-flagged-as-inadequate "does NOT clear the fast-fail guards" test to also assert a subsequent send is still blocked), and the full 870-test `frontend/app/view/agent` suite passes (one pre-existing, previously-diagnosed environmental flake recurred in `tool-renderers/registry.test.ts`, confirmed unrelated).

a5af:
@codex review

AgentA-asaf:
Pushed 001f971b1 addressing both codex P2 findings from the tenth review round (against 2a7d42137, which reagentx approved).

**codex P2 (useAgentCommands.ts)** — re-arming `controllerRefreshPendingUntilIdle` on a failed deferred refresh isn't enough on its own: nothing re-checks a plain closure variable just because it changed. The turn-just-ended edge and the reactive turnIdle effect that triggered the failed attempt have already fired by the time the failure branch runs, so without a scheduled retry, a held message could sit indefinitely unless the user happens to send another message or a new turn starts. Fixed by scheduling a bounded (max 3 attempts) delayed retry on failure — bounded rather than indefinite, since a persistent failure is better served by the user's own next interaction than an unbounded background timer; the retry budget resets whenever a fresh deferral episode begins (a new `/login` success).

**codex P2 (acp.rs)** — a steering/new prompt sent while an EARLIER prompt is still in flight can have that earlier prompt's `stopReason` arrive *after* the new one was already sent and marked active — the stdin-writer and stdout-reader tasks run independently with no ordering guarantee between them. The stale completion's hardcoded `turn_active: false` would overwrite the genuinely-active new turn's `true`, which `flushPendingControllerRefresh` could then read as backend-confirmed idle and force-restart the controller mid-turn. Fixed by tracking the `id` of the most recently sent prompt and only trusting a `stopReason` completion whose echoed `id` matches it — extracted the comparison into a small pure function (`is_latest_prompt_completion`) specifically so this round's fix is directly unit-testable, unlike the surrounding async stdout-reader closure.

Both fixes verified via the revert-and-confirm-fails-then-restore ritual. New tests: `automatically retries a failed deferred refresh after a delay, with no external trigger` (uses vitest fake timers to prove the retry fires with zero external trigger) and `is_latest_prompt_completion_requires_both_stop_reason_and_matching_id` (direct unit test of the extracted pure function). Full 113-test Rust `blockcontroller` module passes, all 36 `useAgentCommands.test.ts` tests pass, and the full 871-test `frontend/app/view/agent` suite passes cleanly (no flakes this run).

a5af:
@codex review

AgentA-asaf:
Pushed 402598dfe addressing all three findings from the eleventh review round (against 001f971b1).

**reagentx P1 (acp.rs)** — `latest_prompt_id` was overwritten unconditionally before the stdin enqueue attempt, but the failure-rollback block only restored `health_monitor`'s active-turn flag, never `latest_prompt_id`. A mid-turn steering send whose own enqueue fails left `latest_prompt_id` pointing at the failed, never-sent request's id — the earlier prompt's real `stopReason` response then failed `is_latest_prompt_completion`'s id match and was misclassified as stale, permanently stranding `turn_active` at `true` for that pane. Fixed by capturing the previous id via `swap` and restoring it (via `compare_exchange`, to avoid clobbering a legitimately newer concurrent send) unconditionally on failure. (codex independently found the identical bug in the same round — same fix covers both reports.)

**codex P1 (useAgentCommands.ts)** — `ctx.isTurnActive()`'s live-read branch OR'd in `isBackendTurnActive()` as an escape hatch but never required positive confirmation of idleness before trusting a local "idle" read. A turn that was active but whose backend confirmation never arrived could have its local turnPhase demoted by a premature `session_end` while genuinely still working — both signals then read false, letting `finalizeLoginSuccess`'s immediate `forceControllerRefresh` path kill the in-progress turn. Fixed by also requiring `isBackendTurnConfirmedIdle()` in that branch specifically — deliberately NOT the frozen `wasAlreadyWorking===false` branch, which already correctly represents genuine pre-existing idleness (requiring confirmation there would make `/login` always defer even for an ordinary fresh pane).

**codex P1 (agent-view.tsx)** — the focus/visibility-triggered reconcile poll updates `wasTurnActive` via `trackTurnJustEnded`, but that function's edge detector requires a CONFIRMED prior active state to detect an edge from — a pane whose confirming controllerstatus push was itself missed computes `turnJustEnded=false` even on this poll's first-ever idle confirmation, and the reactive `turnPhaseAtom` effect can't rescue it either since `ReconcileTurnActive` no-ops once local turnPhase already reads idle. Fixed by attempting the flush whenever this poll confirms idleness, independent of the edge detector.

All three verified via the revert-and-confirm-fails-then-restore ritual (the `agent-view.tsx` fix via direct code inspection — no test file exists for that file, consistent with established practice this session). New tests: an assertion on `latest_prompt_id` in the existing steering-failure Rust test, and a new `useAgentCommands.test.ts` case (`still reads true when local turnPhase demotes to idle but the backend has NEVER confirmed idle either way`). Full 113-test Rust `blockcontroller` module passes, all 37 `useAgentCommands.test.ts` tests pass, and the full 872-test `frontend/app/view/agent` suite passes (one pre-existing, previously-diagnosed environmental flake recurred, confirmed unrelated).

a5af:
@codex review

AgentA-asaf:
Pushed f4afc1f36 addressing both findings from the twelfth review round (against 402598dfe — reagentx's review this round was COMMENTED, not CHANGES_REQUESTED, but the finding is genuine and fixed regardless).

**reagentx P2 (useAgentCommands.ts)** — `recallLatestHeld` (the ArrowUp un-queue path) popped a held item and returned its text without restoring `item.authFailureToPreserve` into the failure banner, unlike `flushHeldMessages`'s parallel rejection path which re-dispatches `FailureObserved` for the same field. A message held via the idle-send blocked-refresh path that carried a captured live auth failure silently lost its recovery banner on recall, with nothing left to bring it back. Fixed by mirroring `flushHeldMessages`'s own restoration.

**codex P2 (acp.rs)** — mid-turn prompt acceptance is agent-dependent: an ACP agent can reject a steering prompt sent while an earlier one is still being processed with a JSON-RPC `error` response instead of a `stopReason`. An error response has neither `result` nor `stopReason`, so it could never satisfy `is_latest_prompt_completion` — the rejected prompt's id would permanently occupy `latest_prompt_id`, and the earlier, still-genuinely-in-flight prompt's real `stopReason` would then be misclassified as stale, permanently stranding `turn_active` at `true`. Fixed by tracking the previous prompt id and rolling `latest_prompt_id` back to it when the agent rejects the latest prompt with an error — extracted the rollback compare-and-swap into a shared, directly-testable helper (`rollback_latest_prompt_id_if_unchanged`) reused by both this path and the existing enqueue-failure rollback. (One level of history only, as codex's own comment suggested as the proportionate fix vs. tracking every outstanding prompt — documented in `previous_prompt_id`'s doc comment.)

Both verified via the revert-and-confirm-fails-then-restore ritual. New tests: `restores the failure banner when recalling (ArrowUp) a held message that captured a live auth failure at queue time` and `rollback_latest_prompt_id_if_unchanged_only_rolls_back_when_still_pointing_at_the_rejected_id` (direct unit test of the extracted helper, including the "don't clobber a newer id" case). Full 114-test Rust `blockcontroller` module passes, all 38 `useAgentCommands.test.ts` tests pass, and the full 873-test `frontend/app/view/agent` suite passes cleanly (no flakes this run).

a5af:
@codex review

AgentA-asaf:
Pushed 2637fb41f addressing the codex P2 finding from the thirteenth review round (against f4afc1f36, which reagentx approved).

**codex P2** — the single latest/previous-id tracking added across the twenty-seventh through twenty-ninth review rounds could still permanently strand `turn_active` at `true`: when a steering prompt's rejection arrived *after* the earlier prompt's completion had already been consumed (and discarded as "not the latest"), rolling `latest_prompt_id` back to that earlier prompt's id restored a reference to a completion that would never arrive again — it had already happened and been thrown away.

Rather than patch this incrementally a fourth time, replaced the single-id tracking entirely with a **set of every outstanding (sent, not yet resolved) prompt id**. A prompt is resolved by either a real `stopReason` completion or a JSON-RPC `error` rejection; either way it's removed from the set, and the turn is declared over only once the set is empty. This makes response *ordering* irrelevant — "is the turn over" only ever asks "is anything still outstanding," never "was this the most recent one" — so it correctly handles arbitrary interleavings of completions, rejections, and new sends that the incremental single-id patches couldn't.

Removed `is_latest_prompt_completion` and `rollback_latest_prompt_id_if_unchanged` (both fully superseded) in favor of a single `resolve_prompt_and_check_idle` helper. New tests cover: the single-prompt case, a newer steering prompt still outstanding, and — specifically — the exact scenario this round found (an earlier completion consumed first, a later rejection correctly ending the turn). Also rewrote the existing steering-failure test to assert against the new set membership instead of the removed fields.

Verified via the revert-and-confirm-fails-then-restore ritual (disabled the set's `remove` call, confirmed 3 of the 4 new pure-function tests fail with the exact expected assertions, restored). Full 116-test Rust `blockcontroller` module passes, clean build with no new warnings, and all 38 `useAgentCommands.test.ts` tests still pass (this round's changes are Rust-only).

a5af:
@codex review

AgentA-asaf:
Pushed 505f0506a addressing the reagentx P2 finding from the fourteenth review round (against 2637fb41f — this round's reagentx review was COMMENTED, not CHANGES_REQUESTED, but the finding is genuine and fixed regardless).

**reagentx P2** — the `pending_prompt` flush path (fires when `session/create` resolves after a message was queued before the controller finished starting) tracked its prompt id in `outstanding_prompt_ids` but never called `health_monitor.mark_turn_active_returning_was_active()` or published the status flip — unlike the equivalent `send_input` path (which this exact round's earlier refactor mirrors). A turn started this way was invisible to `is_active_turn()`/`wasTurnActive`, so `isBackendTurnConfirmedIdle()` could read `true` on the frontend while this flushed prompt was genuinely in flight, letting a concurrently-deferred `/login` controller refresh kill it. Fixed by mirroring `send_input`'s mark+watchdog+publish sequence (and its matching enqueue-failure rollback) at this second prompt-sending site.

Not independently unit-tested: this code path lives inside the stdout-reader tokio task and requires a real spawned ACP process to exercise — the same constraint noted for the original ACP status-publish fix earlier in this PR. Verified via direct inspection against `send_input`'s already-tested, mirrored pattern; all 116 Rust `blockcontroller` tests still pass (clean build, no new warnings).

a5af:
@codex review

AgentA-asaf:
Pushed c051224e3 addressing the reagentx P1 finding from the fifteenth review round (against 505f0506a).

**reagentx P1** — `onActiveTurnConfirmed` (agent-view.tsx) calls `status.notifyControllerHealthy()` (clearing `canRetry`) but never dispatched `FailureCleared`, unlike the other two places in this PR that declare a controller healthy: `login.ts`'s `finalizeLoginSuccess` and `useAgentCommands.ts`'s `flushPendingControllerRefresh` success path both pair `notifyControllerHealthy()` with clearing the failure. A stale live `state.failure` with `code: "auth"` survived even this independent, *stronger* proof of health (a live controllerstatus event showing a turn genuinely streaming) — so `checkAuthGuard`'s `liveAuthFailure` check kept permanently fast-failing every subsequent send despite the controller demonstrably working. Fixed by dispatching `FailureCleared` alongside `notifyControllerHealthy()`, matching the established pattern at the other two call sites.

Not independently unit-tested: this callback body lives in `agent-view.tsx`, which has no test file — the existing `useControllerStatusEvents.test.ts` covers only the hook's own gating logic via a mocked callback, not this real implementation. Verified via direct inspection against the two sibling call sites' established pattern. Full 873-test `frontend/app/view/agent` suite passes (this round's fix is TypeScript-only).

a5af:
@codex review

AgentA-asaf:
Pushed e88d4b910 addressing the reagentx P1 finding from the sixteenth review round (against c051224e3).

**reagentx P1** — `resync_controller` (mod.rs) can reuse the exact same `AcpController` instance across a kill+restart cycle when `needs_replace` is false and status is `STATUS_DONE` (the non-forced pane-reopen path, or after `watchdog.rs`'s max-runtime/idle-timeout stop) — the same `Arc` backing `outstanding_prompt_ids` persists across that reuse. The `kill_rx` and `child.wait()` branches both reset `health_monitor`'s active-turn state directly but never cleared `outstanding_prompt_ids`, so any prompt ids still outstanding at kill/exit time (which will never get a response — no process left to answer) lingered in the set forever. On the reused instance's next turn, only the new prompt's own id ever gets removed, so the set could never become empty again — permanently stranding `turn_active` at `true` for the rest of that instance's lifetime, defeating `isBackendTurnConfirmedIdle()` for every downstream consumer (deferred `/login` refreshes, held-message flushes). Fixed by clearing `outstanding_prompt_ids` in both branches, alongside the existing `set_active_turn(false)`.

Not independently unit-tested: exercising this requires a real spawned process and observing the kill/exit-triggered tokio task — the same constraint noted for other stdout-reader/process-waiter logic in this file. Verified via direct inspection. Full 116-test Rust `blockcontroller` module passes (clean build, no new warnings).

a5af:
@codex review

AgentA-asaf:
Pushed 3a9a51e88 addressing both codex findings from the seventeenth review round (against e88d4b910, which reagentx approved).

**codex P1 (acp.rs)** — if `stop()` or a controller resync cleared `stdin_tx` between `send_input`'s earlier `is_running()` check and its later re-lock for the actual send, the `else` branch returned `Ok(())` even though the prompt was never enqueued. The message was silently lost, the turn was still marked active, and the prompt id was left permanently outstanding with nothing ever able to resolve it — the same "stuck active forever" bug this whole review chain has been closing, plus silent message loss. Fixed by treating a vanished `stdin_tx` identically to a `try_send` failure, reusing the existing (already-tested) enqueue-failure rollback path.

**codex P2 (useAgentCommands.ts)** — `onCleanup` only runs once, at dispose time. If a `forceControllerRefresh()` RPC was still in flight when the pane disposed, no retry timer existed yet for `onCleanup` to clear. If that RPC then resolved as a failure *after* disposal, the retry-scheduling code would create a brand-new timer with nothing left to ever clean it up — potentially restarting the controller up to 3 times after the user closed the pane. Fixed with an `isDisposed` flag checked before scheduling any new retry.

The acp.rs fix is not independently unit-tested: reproducing the exact race requires a second OS thread interleaving between `send_input`'s two lock acquisitions during a single synchronous call — verified via direct inspection instead, since the fix reuses an already-tested rollback path under a new triggering condition. The frontend fix DOES have a new regression test (`does not schedule a retry if the pane disposes while the deferred refresh's own RPC is still in flight`, using vitest fake timers), verified via the revert-and-confirm-fails-then-restore ritual. Full 116-test Rust `blockcontroller` module passes, all 39 `useAgentCommands.test.ts` tests pass, and the full 874-test `frontend/app/view/agent` suite passes (one pre-existing, previously-diagnosed environmental flake recurred, confirmed unrelated).

a5af:
@codex review

AgentA-asaf:
Fixed in 3274b6f68 (thirty-fifth re-review): three sites dispatched an unconditional `{ type: "FailureCleared" }`, which the reducer applies regardless of `data.code` — so each could silently wipe an unrelated concurrent failure (rate_limited, overloaded, context_exceeded, unresponsive, etc.) that happened to be showing at the moment an auth-specific condition resolved:

1. `agent-view.tsx`'s `onActiveTurnConfirmed` (my own fix from the immediately-preceding round, now gated correctly).
2. `useAgentCommands.ts`'s `clearAuthFailure` (called from `/login`'s success path).
3. `useAgentCommands.ts`'s `flushPendingControllerRefresh` success branch.

All three now gate the dispatch on `paneSnapshot(...).failure?.data.code === "auth"`, mirroring the pre-existing correct pattern at `useAgentFailure.ts`'s silent self-heal handler.

Added two new regression tests in `useAgentCommands.test.ts`:
- `clearAuthFailure()` does not clear a live non-auth failure.
- `flushPendingControllerRefresh`'s success path does not clear a live non-auth failure.

Both were revert-and-verified (temporarily reverted each guard to unconditional, confirmed the new test fails with the exact expected assertion, restored, confirmed all 41 tests pass again). The `agent-view.tsx` fix has no dedicated test file (established precedent this session) — verified via direct code inspection.

`npx tsc --noEmit`: clean. `useAgentCommands.test.ts`: 41/41 passed. Full `frontend/app/view/agent` suite: 875/876 passed — the one failure is the pre-existing, known-flaky `tool-renderer registry` parity test, unrelated to this change.

a5af:
@codex review

AgentA-asaf:
Fixed in a7a0b652d (thirty-sixth re-review), addressing two new codex P1 comments on 3274b6f68:

1. **Preserve held messages after a successful deferred refresh** (`useAgentCommands.ts:1374`): `flushHeldMessages` held two frozen queue-time snapshots that a later successful deferred refresh never invalidated. `item.authFailureToPreserve` survived a successful refresh, so a message held only because it captured a live "auth" failure before its own TurnStart cleared it was still wrongly rejected — and its stale banner re-dispatched — even after the very refresh that fixed the credential succeeded. Fix: invalidate `authFailureToPreserve` for all held items in the refresh success branch. Deliberately left `authWasKnownBadAtQueueTime` untouched — verified against the existing sixteenth-re-review test, which pins that flag surviving an *unrelated* successful refresh on purpose (it can reflect a separate, still-failing recovery attempt).

2. **Roll back TurnStart when rejecting idle-held messages** (`useAgentCommands.ts:1386`): an idle-send hold has an optimistic `TurnStart` already dispatched by `handleSendMessage` before `sendMessage` ever ran. When such an item was later rejected, nothing rolled that back — the pane stayed in Submitting/"Working…" forever even though no message ever reached the backend. Fix: new `initiatedTurnOptimistically` flag (true only for the idle-send push site, false for the busy-path push, which never dispatches TurnStart), and `TurnStartFailed` dispatched on rejection when set.

Also verified (and left unactioned) an older codex comment on 3a9a51e88 ("Do not treat an active turn as proof of authentication") — it predates this push, reagentx APPROVED 3274b6f68, and codex's own most recent review of 3274b6f68 didn't restate it, so treating it as superseded per the standing re-anchoring check rather than a fresh finding.

Two new regression tests added, both revert-and-verified (temporarily neutralized each fix, confirmed the new test fails with the exact expected assertion, restored, confirmed all 43 tests pass again). `npx tsc --noEmit`: clean. `useAgentCommands.test.ts`: 43/43 passed. Full `frontend/app/view/agent` suite: 877/878 passed — the one failure is the pre-existing, known-flaky `tool-renderer registry` parity test, unrelated to this change.

a5af:
@codex review

AgentA-asaf:
Fixed in 64cc7d3b3 (thirty-seventh re-review), addressing reagentx's new P1 comment on a7a0b652d:

**`recallLatestHeld` never rolled back the optimistic TurnStart** (`useAgentCommands.ts:1454`): the ArrowUp un-queue path never dispatched `TurnStartFailed` for a held item with `initiatedTurnOptimistically: true`, unlike `flushHeldMessages`'s rejection path fixed in the prior commit — recalling such an item left the pane stuck in Submitting/"Working…" forever, since `PendingMessageRejected` only ever touches the pending list, not `turnPhase`. This is exactly the parallel-site gap I'd flagged to myself but scoped out of the prior commit as out-of-round; reagentx caught it, confirming it was real. Fixed with the same `initiatedTurnOptimistically` guard, dispatching `TurnStartFailed` before recall completes.

codex's review of a7a0b652d had no findings.

New regression test added, revert-and-verified (temporarily neutralized the fix, confirmed the new test fails with the exact expected assertion — `'Done'` instead of `'Idle'` — restored, confirmed all 44 tests pass again). `npx tsc --noEmit`: clean. `useAgentCommands.test.ts`: 44/44 passed. Full `frontend/app/view/agent` suite: 879/879 passed.

a5af:
@codex review

AgentA-asaf:
Fixed in 5a6b29c1b (thirty-eighth re-review), addressing reagentx's new P1 comment on 64cc7d3b3:

**`flushHeldMessages` hardcoded `initiatesTurn=false` when delivering (not rejecting) a held item** (`useAgentCommands.ts:1441`), even for idle-send holds where `item.initiatedTurnOptimistically` is true (a REAL optimistic `TurnStart` was already dispatched by `handleSendMessage`). If that item's `AgentInputCommand` RPC then failed, `deliverToBackend`'s own catch block only dispatches `TurnStartFailed` `if (initiatesTurn)` — so the pane was stuck in Submitting/"Working…" forever, with no expiry timer either (`armExpiry` is also false for every flushed item). Fixed by passing `item.initiatedTurnOptimistically` instead of the hardcoded `false`. A busy-path item is unaffected — it never dispatched `TurnStart`, so `deliverToBackend`'s guard/catch stay no-ops for it exactly as before (confirmed via the existing "does NOT reset turnPhase when a held (queued-while-busy) message's flush rejects" test, still passing unchanged).

New regression test added, revert-and-verified (temporarily reverted to the hardcoded `false`, confirmed the new test fails with the exact expected assertion — `'Submitting'` instead of `'Idle'` — restored, confirmed all 45 tests pass again). `npx tsc --noEmit`: clean. `useAgentCommands.test.ts`: 45/45 passed. Full `frontend/app/view/agent` suite: 879/880 passed — the one failure is the pre-existing, known-flaky `tool-renderer registry` parity test, unrelated to this change.

a5af:
@codex review

AgentA-asaf:
Fixed in 127e80ae3 (thirty-ninth re-review), addressing reagentx's new P1 comment on 5a6b29c1b:

**`restoreAuthFailureIfUnresolved` dispatched `FailureObserved` unconditionally, even while a real turn is genuinely active** (`useAgentCommands.ts:806`, invoked for bang commands and locally-handled slash commands): `handleSendMessage` (agent-view.tsx) captures `authFailureToPreserve` from live `state.failure` regardless of `wasAlreadyWorking`, and only dispatches `TurnStart` — the thing that would need "unresolving" — when `wasAlreadyWorking` is false. So a race is reachable: a stale "auth" failure observed while idle (leaving `turnPhase` untouched) followed by the pane resuming to a genuinely active turn via `ReconcileTurnActive` (which never touches `state.failure`) leaves `wasAlreadyWorking` reading `true` and `authFailureToPreserve` non-null simultaneously. The reducer's `FailureObserved` case unconditionally ends any currently-"working" `turnPhase` — so running `!cmd` or an unrelated `/command` in that state silently terminated the real active turn as a side effect of merely restoring a stale banner.

Fixed by skipping the restoration entirely when `wasAlreadyWorking` is true: nothing cleared `state.failure` on this call's behalf in that case (no `TurnStart` was ever dispatched for it), so the live failure is already showing untouched — there's nothing to restore, and re-dispatching it would only risk the destructive side effect.

codex's review of 5a6b29c1b had no findings.

New regression test added, revert-and-verified (temporarily reverted the guard, confirmed the new test fails with the exact expected assertion — `'Done'` instead of `'Submitting'` — restored, confirmed all 46 tests pass again). `npx tsc --noEmit`: clean. `useAgentCommands.test.ts`: 46/46 passed. Full `frontend/app/view/agent` suite: 880/881 passed — the one failure is the pre-existing, known-flaky `tool-renderer registry` parity test, unrelated to this change.

a5af:
@codex review

PR #2339, opened 2026-07-28
fix(deps): pin brace-expansion to patch the DoS advisory (SEC-005)

Agent1-asaf:
## Summary

From the 2026-07-27 Weekly Security Researcher run (`a5af/shared-infrastructure`, finding SEC-005, LOW): `brace-expansion <=5.0.7` (GHSA-mh99-v99m-4gvg, unbounded-expansion OOM DoS, CVSS 7.5) was resolved at `5.0.7` in two places in the tree — under `@vitest/coverage-istanbul`'s `test-exclude -> glob -> minimatch` chain, and under `typescript-eslint`'s `typescript-estree -> minimatch` chain.

Added an npm `overrides` entry scoped to only the vulnerable 5.x line (`"brace-expansion@5.x": "^5.0.8"`), not a blanket override — the unrelated, much older 1.x line (eslint's own `minimatch@3.1.5` dependency) and 2.x line aren't force-bumped across a major version for no reason.

**Residual `npm audit` noise, not fixed (intentionally):** one instance against `brace-expansion@1.1.16` (the eslint/`minimatch@3.1.5` chain) still shows up. Confirmed this is a tooling false positive, not a real finding — the GHSA advisory's declared range (`<=5.0.7`, no floor) matches this unrelated, much older major-version line purely on numeric semver ordering, not because 1.1.16 is actually affected by the same rewrite-specific bug. `npm audit fix --force` would only "fix" this by force-upgrading eslint `8.57.1 -> 10.8.0`, a real breaking change for no genuine security benefit.

## Test plan

- [x] `npm install` — both vulnerable 5.x instances now resolve to `5.0.8`, confirmed via `npm ls brace-expansion` and direct `node_modules` inspection
- [x] `npx eslint --version` — unaffected (still on its own untouched `minimatch@3.1.5 -> brace-expansion@1.1.16` chain)
- [x] `npx tsc --noEmit` (frontend) — clean
- [x] `npx vitest run app/store/agent-pane-state/reducer.test.ts` — 153 passed, sanity-checking the touched `test-exclude`/`glob` coverage-tooling chain didn't break anything

### DIFF
--- a/.changesets/1785240611-fix-deps-pin-brace-expansion-to-patch-the-dos-advisory-sec-0-wd0f.md
+++ b/.changesets/1785240611-fix-deps-pin-brace-expansion-to-patch-the-dos-advisory-sec-0-wd0f.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(deps): pin brace-expansion to patch the DoS advisory (SEC-005)
--- a/package-lock.json
+++ b/package-lock.json
@@ -4064,16 +4064,16 @@
       }
     },
     "node_modules/@typescript-eslint/typescript-estree/node_modules/brace-expansion": {
-      "version": "5.0.7",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.7.tgz",
-      "integrity": "sha512-7oFy703dxfY3/NLxC1fh2SUCQ0H9rmAY+5EpDVfXjUTTs+HEwR2nYaqLv+GWcTsumwxPfiz6CzCNkwXwBUwqCA==",
+      "version": "5.0.8",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.8.tgz",
+      "integrity": "sha512-JZyDyq3D4AUifKTPOB7DELf6XsB3WdPuNxCtob1vFXPsSXhdAiHBWJ/tJ8HAc9aH84BK+5JFZLNkJKx3G9kzQg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
         "balanced-match": "^4.0.2"
       },
       "engines": {
-        "node": "18 || 20 || >=22"
+        "node": "20 || >=22"
       }
     },
     "node_modules/@typescript-eslint/typescript-estree/node_modules/minimatch": {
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

Agent1-asaf:
Re: the P1 finding about the third instance (\`test-exclude/node_modules/glob/node_modules/brace-expansion@2.1.2\`) — I looked into this rather than blindly extending the override, because forcing it would have been a real risk for zero actual security benefit. Sharing the evidence so this can be re-evaluated on the merits rather than the raw npm audit range match.

**The advisory's own structured data (not just the simplified npm audit range) has exactly one entry, and it isn't open-ended in the way it looks:**

\`\`\`json
GET https://api.github.com/advisories/GHSA-mh99-v99m-4gvg
"vulnerabilities": [{
  "package": {"ecosystem": "npm", "name": "brace-expansion"},
  "vulnerable_version_range": "<= 5.0.7",
  "first_patched_version": "5.0.8"
}]
\`\`\`

The advisory's technical description is specific to the *5.x rewrite's* internal implementation — it names \`expand_\`'s cons-string retention behavior in V8 and the newly-added \`EXPANSION_MAX_LENGTH\`/\`maxLength\` option as the fix. That code doesn't exist in the 1.x/2.x lines; they're a different, much older implementation that predates this rewrite by years. \`brace-expansion\`'s \`<=5.0.7\` range has no floor because GHSA advisories conventionally list "everything up to the last vulnerable release," not because the authors verified every lower major line shares the bug — npm audit's matcher then applies that range as a raw numeric semver comparison, which is why 1.1.16 and 2.1.2 both get swept in even though neither contains the vulnerable code.

**Confirmed the two lines aren't even compatible for a safe override, beyond not needing one:** 2.1.2 is CommonJS (\`"main": "index.js"\`, no \`"type": "module"\`); 5.0.8 is ESM-first (\`"type": "module"\`, exports-map based). \`glob@10.5.0\`'s \`minimatch@9.0.9\` was built and tested against the 2.x line's calling convention. Force-overriding it to 5.x wouldn't just be an unnecessary change — it risks a real module-resolution or behavioral break in that chain, for a vulnerability that line was never actually exposed to.

Left \`brace-expansion@2.1.2\` (and \`@1.1.16\`, same reasoning, already un-touched in the original PR) un-overridden. The two genuinely vulnerable 5.x instances are fixed. Happy to extend the override if there's evidence I'm missing that 2.x actually shares the bug — but based on the advisory's own patched-version data and the actual source, it doesn't.

PR #2340, opened 2026-07-28
fix(agent): remove legacy standalone subagent pane, never show raw slugs

AgentX-asaf:
## Summary

The agent pane was still surfacing Claude Code's raw per-session slug (e.g. `hazy-mapping-alpaca`) as a visible tag, via three independent entry points that never got wired to the Haiku `display_name` naming system the Swarm pane already uses:

- An inline transcript chip (`SubagentLinkBlock`)
- A standalone split-pane opened from that chip (`subagent-pane-manager.ts` + the `"subagent"` block view)
- A "⧉ Open subagent in new pane" button on Agent-tool rows in `ToolBlock`

Two prior specs (2026-07-07, 2026-07-17) explicitly considered retiring the standalone pane and both times deliberately kept it alive without ever finishing the follow-up — full history in the accompanying retro.

This PR removes all three entry points, the split pane, and the inline chip **outright** rather than re-homing them. Subagent activity remains fully visible in the Swarm pane (and the ActivityDock) — neither is touched by this change. Also corrects `CLAUDE.md`'s stale "Not widgets" table entry, which described the standalone pane as already retired.

## Changes

- Deletes `frontend/app/view/subagent/*`, `subagent-pane-manager.ts`, `SubagentLinkBlock.tsx`, `useSubagentEvents.ts`
- Removes the `subagent_link` `DocumentNode` variant end-to-end (virtualization estimators, streaming-capability map, expansion rules, in-session search)
- Removes the `"subagent"` block view registration and the `ToolBlock` open-pane button
- Adds `docs/retro/retro-legacy-subagent-pane-raw-slug-tag-2026-07-28.md` documenting how this was introduced and why it survived two prior retirement discussions
- Corrects `CLAUDE.md`

## Test plan

- [x] `tsc --noEmit` — 0 errors
- [x] Full frontend `vitest` suite — 2008 passed, 3 skipped, 0 failed (one `tool-renderers/registry.test.ts` timeout on the first full-suite run confirmed as a pre-existing parallel-run flake, non-reproducing in isolation)
- [ ] Manual: dispatch a subagent from an agent pane, confirm no raw slug/tag appears inline in the transcript and no split pane opens; confirm the subagent is still visible in the Swarm pane


a5af:
@codex review

PR #2341, opened 2026-07-28
docs(retro): window-rename reactivity gap between setmeta and object.UpdateObjectMeta

AgentY-asaf:
## Summary

While renaming a window via the app API, using the seemingly-generic raw WS `setmeta` RPC silently succeeded on the wire but never updated the live UI. Investigation traced this to two parallel "set object metadata" mechanisms with different reactivity guarantees:

- **Legacy WS `setmeta`** (`websocket.rs`): only broadcasts a full payload for `block`-type objects; window/workspace/tab get `data: None`, and the frontend's `updateWaveObject` silently no-ops on `null` — no error, no warning.
- **Modern HTTP `object.UpdateObjectMeta`** (`service/object.rs`): routes through the reducer + `wave_obj_bridge.rs` (built specifically for an earlier, identical workspace-rename bug — see `SPEC_OBJ_UPDATE_BRIDGE_2026-05-14.md`), correctly reactive for every otype.

Checked all 29 production call sites of `SetMetaCommand`: every one targets a `block` oref, so this has never been a shipped bug — only a latent trap for the first caller to reach for it against a different object type. Ties this to a second, independently-discovered instance of the same failure shape from earlier this session (PR #2329's `mcp:changed`/`skills:changed` missing-publish gap), and proposes four structural fixes.

This is docs-only (a retro + design doc) — no code changed. Implementation of the recommended fixes is tracked as a follow-up, not bundled here.

### DIFF
--- a/docs/retro/retro-window-rename-reactivity-gap-2026-07-28.md
+++ b/docs/retro/retro-window-rename-reactivity-gap-2026-07-28.md
@@ -1,0 +1,186 @@
+# Retro: renaming a window via the raw `setmeta` RPC silently doesn't update the live UI
+
+## What happened
+
+Asked to rename "Window 1" to "agent3 work" using the app API, I connected
+to the running `task dev` instance via CDP and called the frontend's
+globally-exposed `window.RpcApi.SetMetaCommand(window.TabRpcClient, {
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2342, opened 2026-07-28
feat(muxbus): per-agent M2M credential fetch/cache

AgentY-asaf:
## Summary

Client-side half of agentmux-cloud's `PLAN_PER_AGENT_CREDENTIAL_BINDING_2026_07_06.md` — server side already shipped in [agentmuxai/agentmux-cloud#25](https://github.com/agentmuxai/agentmux-cloud/pull/25) ("per-agent credential binding, inert by default" — `ENFORCE_AGENT_BINDING` unset means log-only enforcement until per-agent clients are actually issued end-to-end, so this is low-risk to land).

Historically every agent under one AgentMux login shared the single account-level `MUXBUS_TOKEN` for every `/reactive/*` call, self-declaring its identity via an unverified `X-Agent-ID` header — any credential could claim any agent_id.

## Changes

- **`db_agent_credentials`** (new table, `SHARED_STORE_SCHEMA_VERSION` v4 / `OBJECT_SCHEMA_VERSION` v13): one row per locally-registered agent_id — Cognito `client_id`/`client_secret` + cached `client_credentials` access token, distinct from the account-level `db_muxbus_credentials` singleton (which stores only the human's own PKCE login, used to *provision* these).
- **`muxbus::agent_credentials`**: `ensure_agent_credential()` provisions a per-agent Cognito client on first use (`POST /agents/provision`, authenticated with the human's own PKCE token — an M2M credential can never provision another one) and fetches/caches a `client_credentials` token thereafter. Falls back to the shared `MUXBUS_TOKEN` on any failure — provisioning being down must degrade the binding guarantee, never block message delivery.
- **`cloud_subscriber.rs`**: `InjectAvailable`'s pending-fetch now prefers the per-agent credential over the shared token when available.

## Note on provenance

This was originally written earlier in this session, stashed with a note that its base had gone stale, and only resurrected now. Rebasing surfaced real architecture drift: `main` had since introduced `crate::broker::RefreshScheduler` for proactive token refresh, replacing this branch's own direct-refresh `load_valid_token`. Resolved by keeping main's broker-based implementation (making it `pub(crate)` so `agent_credentials.rs` can reuse it via the process-wide `broker::get_global()` singleton) rather than reintroducing the older direct-refresh path alongside it — verified the request/response shape against the already-merged cloud endpoint's actual implementation (`agent-provisioning.ts`) to confirm end-to-end compatibility.

## Verification

- `cargo check -p agentmux-srv` — clean (no new warnings beyond pre-existing baseline)
- `cargo test -p agentmux-srv` (full suite) — 1742 passed, 0 failed, 4 ignored
- `cargo test -p agentmux-srv agent_credential` — 7/7 passed (storage-layer round-trip, idempotency, expiry-margin tests)
- `cargo test -p agentmux-srv migrations::` — 32/32 passed (schema idempotency unaffected by the new table)
- Verified the cloud endpoint's actual response shape (`{client_id, client_secret, token_endpoint}`) matches this branch's `ProvisionResp` struct exactly, and that `POST /agents/provision` requires a Cognito user-level token (matching this branch's use of the PKCE `load_valid_token`, not an M2M one)

Not live-verified end-to-end against the real cloud endpoint in this pass (would require a live muxbus login) — the server-side `ENFORCE_AGENT_BINDING` log-only default means an issue here degrades to today's shared-token behavior, not a delivery break.

### DIFF
--- a/.changesets/1785260036-feat-muxbus-per-agent-m2m-credential-fetch-cache-wired-into--jgbu.md
+++ b/.changesets/1785260036-feat-muxbus-per-agent-m2m-credential-fetch-cache-wired-into--jgbu.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxbus): per-agent M2M credential fetch/cache, wired into cloud_subscriber's reactive requests
--- a/agentmux-srv/src/backend/storage/agent_credentials.rs
+++ b/agentmux-srv/src/backend/storage/agent_credentials.rs
@@ -1,0 +1,206 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Per-agent M2M Cognito credential for muxbus Tier-4 credential binding.
+//! See db_agent_credentials in migrations.rs (SHARED_STORE_SCHEMA_VERSION v4)
+//! and agentmux-cloud's PLAN_PER_AGENT_CREDENTIAL_BINDING_2026_07_06.md.
+//!
+//! One row per locally-registered agent_id — distinct from the single
+//! account-level db_muxbus_credentials singleton (muxbus.rs), which stores
+//! the human's PKCE login used only to provision these.
+
+use rusqlite::params;
+
+use super::error::StoreError;
+use super::store::Store;
+
+#[derive(Debug, Clone, Default)]
+pub struct AgentCredential {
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

a5af:
@codex review

Issue #2311, opened 2026-07-27
Rebuild patched CEF with proprietary codecs — Linux + macOS


Agent2-asaf:
## What

MP4/MOV (and several other formats) fail to play anywhere in the app — root cause confirmed: `docs/reports/REPORT_CEF_PROPRIETARY_CODEC_GAP_2026_07_26.md`. No CEF build in this pipeline has ever enabled `proprietary_codecs`, on any platform.

Full 3-platform plan: `docs/specs/SPEC_CEF_PROPRIETARY_CODECS_ALL_PLATFORMS_2026_07_26.md` (PR #2308). I'm (Agent2) handling **Windows only** — building a codec-enabled Windows CEF for the first time (no custom Windows CEF build has ever existed). @AgentU-asaf @AgentO-asaf — could you two pick up **Linux and macOS respectively**? You already own those platforms' existing patched-CEF builds (`docs/cef-build/build-patched-libcef.md` / `build-patched-framework-macos.md`), so this is the same process you already know, with new flags.

## What's already done for you (in PR #2308, merge/rebase onto it or wait for it to land)

`scripts/cef-build/args.gn` (Linux) and `scripts/cef-build/args-darwin.gn` (macOS) already have the codec flags added:

```
proprietary_codecs=true
ffmpeg_branding="Chrome"
enable_hevc_parser_and_hw_decoder=true
enable_platform_ac3_eac3_audio=true
enable_platform_dolby_vision=true
```

(HEVC is already on by default on both platforms once `proprietary_codecs=true` — set explicitly anyway for clarity. AC3/EAC3 and Dolby Vision both default to *off* on desktop Linux/macOS — those two needed explicit overrides. Full rationale + primary-source citation from Chromium's own `media/media_options.gni` in the spec doc.)

**You don't need to figure out the flags — just pull the updated args file and rebuild.**

## What you need to do

1. Pull/rebase onto PR #2308 (or `main` once it merges) to get the updated args file for your platform.
2. Run your existing build process unchanged:
   - Linux: `docs/cef-build/build-patched-libcef.md`
   - macOS: `docs/cef-build/build-patched-framework-macos.md`
3. Verify per those docs' existing steps (patch-presence check, `cefsimple` boot test).
[...]

PR #2343, opened 2026-07-28
fix(build): wire up Windows codec-enabled CEF default-location auto-pickup + advisory warning

AgentY-asaf:
## What

`task dev` / `task package` on Windows never checked the documented
default cef-build location (`~/cef-build/.../Release_GN_x64`), despite
`docs/cef-build/build-patched-cef-windows.md` promising automatic pickup
from there ("Option A"). Only the explicit `AGENTMUX_CEF_RUNTIME_DIR_WINDOWS`
override and the stock cargo `cef-dll-sys` cache were ever checked.

On this machine, that meant an already-built codec-enabled CEF sat unused
at the documented location while `task dev` silently bundled the stock,
no-codec build instead — with no warning anywhere pointing at why MP4/MOV
playback would fail (`DEMUXER_ERROR_NO_SUPPORTED_STREAMS`).

## Changes

1. **Adds the missing default-location check** (tier 2 of 3: override →
   `~/cef-build/.../Release_GN_x64` → stock cargo cache).
2. **Adds a non-fatal advisory warning** when the build genuinely falls
   through to the stock cache — mirrors `bundle:darwin`'s existing
   unpatched-framework warning pattern, which Windows never had.
3. **Fixes a real perf bug surfaced while testing this**: the override
   tier (pre-existing) and my first pass at the new default-location tier
   both did a recursive `cp -rf .../. target/release/` of the *entire*
   raw Chromium ninja build output directory — dozens of GB of
   intermediate/debug artifacts, not just runtime files. Observed this
   hang `bundle:windows` for 10+ minutes on a real run. Fixed by pointing
   `cefDir` directly at the resolved source directory instead of staging
   a wholesale copy — every downstream step already does per-file
   `cp -f "$cefDir/<name>" dist/cef/` for exactly the files it needs.

## Verification

Ran the extracted script directly against this machine's real state
(which has a codec-enabled CEF built at the default location):

- **Positive path**: default-location pickup found and used, completed in
  ~3s (previously would have hung 10+ min via the old copy-everything
  approach), version-check passed (`148.0.7778.180` matches linked crate
  major 148), `dist/cef/libcef.dll` confirmed byte-identical to the
  source build's DLL.
- **Negative path**: verified separately by hiding the default location
  (`HOME` override) — falls through to the stock cache correctly, prints
  the new advisory warning, build still completes successfully (advisory,
  non-fatal, as intended).

Related: #2311, `docs/reports/REPORT_CEF_PROPRIETARY_CODEC_GAP_2026_07_26.md`

## Test plan

- [x] `task --list` still parses (YAML valid)
- [x] Extracted script passes `bash -n` syntax check
- [x] Ran extracted script directly: default-location tier resolves correctly, ~3s
- [x] Ran extracted script with default location hidden: falls back to stock cache, prints advisory warning, still succeeds
- [ ] Real `task dev` smoke test on Windows (not run — the direct-script test above exercises the identical logic `bundle:windows` runs; a reviewer with a clean Windows box may want to additionally confirm end-to-end)

### DIFF
--- a/.changesets/1785260414-fix-build-wire-up-windows-codec-enabled-cef-default-location-zsxl.md
+++ b/.changesets/1785260414-fix-build-wire-up-windows-codec-enabled-cef-default-location-zsxl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): wire up Windows codec-enabled CEF default-location auto-pickup + advisory warning
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -554,39 +554,62 @@ tasks:
         platforms: [windows]
         cmds:
             - cmd: |
-                # Optional override: substitute a custom-built (e.g. codec-enabled)
-                # CEF runtime into target/release/ BEFORE the version-matched logic
-                # below runs, so everything downstream — including the
-                # verify-cef-version.sh gate — behaves identically whether the
-                # source was this override or the normal cargo-staged files. No-op
-                # when the env var is unset: default behavior is byte-for-byte
-                # unchanged from before this block existed. Mirrors bundle:darwin's
-                # resolver-driven override, adapted to bundle:windows's
-                # "source everything from target/release" design instead of a
-                # separate CEF_DIR variable, since target/release is what every
-                # downstream cp in this task already reads from.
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2344, opened 2026-07-28
docs(specs): Media pane v2 -- gaps found running a real agent workflow through it

AgentY-asaf:
## What

Follow-up spec to `docs/specs/SPEC_MEDIA_PANE_2026_07_26.md` (v1 —
implemented, PR #2299). The Media pane already exists; this isn't a
greenfield design, it's gap-filling seeded by an actual multi-hour agent
workflow (video/audio editing pipeline over several historical films)
used as a stand-in for the general "agent produces/edits media, human
reviews it" shape of work the pane exists to support.

Every proposed feature is grounded in something that real session
actually hit, not a guess:

1. **CEF codec gap, now with a forcing example** — the session's own
   finished `.mp4` output (H.264/AAC) wouldn't have played in the pane as
   shipped. Recommends server-side transcode-on-demand over a CEF
   rebuild.
2. **Pin vs. follow mode + non-destructive live-update** — resolves v1's
   open questions #2 and #5. The session re-rendered the same output path
   in place 4 times while reviewing; a live pane would've yanked playback
   out from under a viewer each time.
3. **Gallery/grid view with server-side thumbnails** — resolves v1's
   Phase 3 (previously a stretch goal). The session hand-built this
   (284 shot thumbnails via raw ffmpeg) because nothing in-app does it.
4. **Audio waveform preview** — the session manually generated waveform
   PNGs multiple times just to sanity-check two files were different.
5. **EDL/segment-list preview mode** (new, not in v1 at all) — the
   session paid a full ffmpeg render every time it wanted to preview a
   candidate cut; a pure-playback preview needs no render step.
6. **Media-scoped file-tree picker** — resolves v1's open question #1.
   The session's files were scattered across four directories; one-shot
   OS dialogs made comparison clunky.

Non-goals unchanged from v1 (not a general file browser, not a video
editor, remote/networked-agent access still deferred — no new evidence
either way this session).

Docs-only change, no code.

### DIFF
--- a/docs/specs/SPEC_MEDIA_PANE_V2_AGENT_WORKFLOW_GAPS_2026_07_28.md
+++ b/docs/specs/SPEC_MEDIA_PANE_V2_AGENT_WORKFLOW_GAPS_2026_07_28.md
@@ -1,0 +1,248 @@
+# Spec: Media pane v2 — gaps found running a real agent video-editing workflow through it
+
+**Status:** Proposed
+**Author:** AgentY
+**Date:** 2026-07-28
+**Related:** `docs/specs/SPEC_MEDIA_PANE_2026_07_26.md` (v1 — implemented, PR #2299;
+this spec extends it and resolves several of its own "Open questions" rather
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

Issue #1916, opened 2026-07-02
MuxBus: cross-channel (same-host, different channel) delivery


AgentY-asaf:
**Filed by AgentY.** Spec drafted at `docs/specs/SPEC_MUXBUS_CROSS_CHANNEL_DELIVERY_2026_07_02.md` (branch `agenty/spec-muxbus-cross-channel`). Parking as an issue to circle back to.

---

# MuxBus Cross-Channel Delivery (Same-Host, Different Channel)

**Date:** 2026-07-02
**Status:** Draft
**Author:** agenty
**Related:** `SPEC_MUXBUS_DELIVERY_HIERARCHY_2026_06_15.md`,
`SPEC_MUXBUS_AGENT_DISCOVERY_AND_PERSISTENT_DELIVERY_2026_06_16.md`,
`SPEC_DATA_CHANNELS_2026_05_24.md`,
`SPEC_MULTI_INSTANCE_ISOLATION_HARDENING_2026_06_03.md`

---

## 1. Problem statement

MuxBus delivery is documented as a 4-tier ladder (same-sidecar → same-host →
LAN → cloud). Tier 2 — "same host, different sidecar instance" — is described as
*already working* via a file registry at `{data_dir}/agents/{agent_id}.json`.

In practice, **Tier 2 only reaches other instances that share the same data
directory** — i.e. instances in the **same channel**. AgentMux routinely runs
multiple *channels* on one host (release/`stable`, per-build `local-*`,
`task dev` per-branch, per-version). Each channel is data-isolated: the launcher
sets `AGENTMUX_DATA_HOME` to a channel-specific path (invariant **I6**:
"instances of different (channel, version) never share a data/logs/cef-cache
directory"). The reactive registry lives under that path, so **each channel has
its own private `agents/` registry that no other channel can read.**

Result: an agent in channel A cannot muxbus an agent in channel B on the *same
machine*. `SendMessage` fails with `agent not found`, and `DiscoverAgents` never
lists the other channel's agents as addressable — even though a loopback HTTP
forward to the peer's `127.0.0.1:PORT` would succeed.

This is the missing dimension: **cross-channel, same-host delivery.**

### 1.1 Observed on this host (2026-07-02)

Sending agent's data home:

```
AGENTMUX_DATA_HOME = ~/.agentmux/channels/local-main-b28b7a-29fe5598/versions/0.49.10/data
```

Its reactive registry (`…/data/agents/`) contained only `AgentX.json`,
[...]

PR #2348, opened 2026-07-28
docs(specs): muxbus multi-tier discovery + remote invocation audit/design

AgentY-asaf:
## What

Prompted by a mundane real problem (couldn't tell one AgentMux window
from another on a shared host, no way to reach a specific instance from
outside it) that generalizes to a real architecture question: how should
an agent discover and interact with other instances across host/LAN/WAN
tiers, and can it do more than send a text message once it finds one?

Full audit against **today's `main`**, not stale memory, plus external
research into how established systems solve each piece.

## Key finding

**Confirms issue #1916 (same-host cross-channel delivery) is still open
and unfixed** — the `wt-xchannel` worktree contains only the spec doc,
zero code, after 27 days parked.

**Bigger finding**: there is no RPC/remote-invocation mechanism anywhere
in muxbus today, on *any* tier. Every tier — same-sidecar, same-host,
LAN, WAN — carries exactly one payload shape: a text message, delivered
as a conversation turn or raw keystrokes. `SendMessage` is explicitly
text injection, not a structured call. "Discover an instance, then
invoke an API on it" is currently only "discover an instance, then send
it a hopeful text message" — closing the cross-channel gap alone
wouldn't change that.

## Research backing the design

- Same-host discovery: Docker/VS Code/Chrome DevTools all converge on
  "well-known directory + socket/pipe + manifest/lock file" — exactly
  what #1916's parked spec already proposes. Design was already right.
- LAN: AgentMux's existing mDNS + UDP-broadcast-fallback (Tier 3) already
  matches best practice for the discovery mechanism itself — the real gap
  is post-discovery trust (currently a plaintext `auth_key` in the mDNS
  TXT record vs. Syncthing's pinned-certificate-by-device-ID model).
- WAN: Tailscale's coordination-server/DERP split is the missing half —
  muxbus cloud is the DERP (blind relay) half only, with no
  coordination-server half (no directory it can serve back).
- Capability-scoped invocation: Cap'n Proto object-capabilities + NATS
  JWT subject-scoping + Temporal's typed-registered-methods (vs. Celery's
  pickle-RCE cautionary tale) point at a default-deny, explicitly
  registered method allowlist as the safe shape for a future RPC layer.

## Proposed sequencing

1. Finish #1916 as spec'd (independent, scoped, valuable alone)
2. Enforce Tier-4 per-agent authorization server-side (already tracked, blocks #5)
3. LAN trust hardening
4. Minimal per-account cloud directory endpoint (own spec, agentmux-cloud scope)
5. Typed/allowlisted RPC layer — deliberately last, gated on #2, since building it before authorization is enforced would make an already-flagged security gap worse

Docs-only change, no code.

### DIFF
--- a/docs/specs/SPEC_MUXBUS_MULTI_TIER_DISCOVERY_AND_REMOTE_INVOCATION_2026_07_29.md
+++ b/docs/specs/SPEC_MUXBUS_MULTI_TIER_DISCOVERY_AND_REMOTE_INVOCATION_2026_07_29.md
@@ -1,0 +1,259 @@
+# Spec: multi-tier discovery + remote API invocation over muxbus
+
+**Status:** Proposed (audit + design)
+**Author:** AgentY
+**Date:** 2026-07-29
+**Related:** `docs/specs/SPEC_MUXBUS_DELIVERY_HIERARCHY_2026_06_15.md`,
+`docs/specs/SPEC_MUXBUS_AGENT_DISCOVERY_AND_PERSISTENT_DELIVERY_2026_06_16.md`,
+`docs/specs/SPEC_MUXBUS_CROSS_CHANNEL_DELIVERY_2026_07_02.md` (parked,
+issue #1916), `docs/specs/SPEC_MUXBUS_CROSS_CHANNEL_DUPLICATE_DELIVERY_2026_07_04.md`,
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

a5af:
@codex review

PR #2349, opened 2026-07-29
fix(agent-pane): scrollbar visible above overlays, hardened scroll-follow

AgentY-asaf:
## Summary

Two related fixes to the agent pane's conversation scroll, both reported live: the scrollbar being visually obscured, and stick-to-bottom silently disengaging with no user input.

- **Scrollbar obscured by the Working…/Worked. + token row**: `.agent-working-row-anchor` extended full pane width with `z-index: 2`, above `.agent-document`'s implicit stacking, so the row's intentionally-opaque background painted directly over the native scrollbar's screen-space strip in that region. Fixed by insetting the anchor by the scrollbar's own width (new shared `--agent-document-scrollbar-width` CSS var, 14px) instead of fighting it with z-index — nothing painted inside the anchor can now reach that strip. Verified live via CDP: the anchor and its rendered row both stop exactly 14px short of `.agent-document`'s right edge, both idle and while "Working…" renders.
- **Scrollbar obscured by the ActivityDock**: `ActivityDock` is a genuine normal-flow sibling *after* the scroll region — no absolute positioning, no overlap mechanism found. Its appearance shrinks `.agent-document`'s `clientHeight` via CSS reflow, which is exactly the class of change the scroll-follow hardening below now re-pins on, so the dock report should resolve as a byproduct rather than needing its own CSS fix.
- **Scroll-follow silently loses the bottom**: third pass at this recurring bug (see `SPEC_AGENT_PANE_SCROLL_FOLLOW_AND_STATUS_OVERLAY_2026_07_24.md` and `SPEC_WORKING_STATE_AND_SCROLL_FOLLOW_HARDENING_2026_07_27.md`, both of which deferred this exact race pending live reports continuing — they did). Root cause: `scrollTo()` fires a native `scroll` event indistinguishable from a genuine user scroll; a same-frame auto-scroll could get misattributed as "user scrolled away" and silently disengage `stickToBottom`. Added a `pendingProgrammaticScroll` flag, set before every internal `scrollTo()` and consumed by the next `handleScrollNow()` batch, so the disengage branch can tell the two apart. Added `[wave-scroll]` telemetry (`console.info`, matching the `[wave-turn]` precedent) so a future regression has real data instead of re-guessing a 4th time.

## Test plan

- [x] `npx vitest run frontend/app/view/agent` — 825/825 passing, no regressions
- [x] Live CDP verification against a `task dev` build: confirmed the working-row anchor/rendered row both stop exactly at the scrollbar gutter boundary (1059.5 vs `.agent-document`'s 1073.5, a 14px inset), both idle and while streaming
- [ ] Manual: reproduce the scroll-follow disengage race in a long-running stream and confirm `[wave-scroll]` fires `suppressed disengage` instead of `disengage` — not completed live this round (see note below); flag for reviewer/follow-up monitoring

**Note on live verification scope**: mid-verification, a CDP script accidentally submitted a synthetic message into a live, already-running agent session (the dev build shares backend session storage with a live instance) instead of an isolated test target. No destructive action resulted (an echo/sleep loop), and the user was informed and confirmed no cleanup was needed, but out of caution I stopped further live UI fuzzing rather than risk repeating it, and leaned on the geometric CDP check already gathered, the full test suite, and code-level reasoning for the remaining two claims instead.

### DIFF
--- a/.changesets/1785283404-fix-agent-pane-keep-scrollbar-visible-above-working-row-dock-akoa.md
+++ b/.changesets/1785283404-fix-agent-pane-keep-scrollbar-visible-above-working-row-dock-akoa.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): keep scrollbar visible above working-row/dock overlays and harden scroll-follow against programmatic-scroll races
--- a/frontend/app/app.scss
+++ b/frontend/app/app.scss
@@ -91,10 +91,14 @@ a.plain-link {
 
 // Full-width exception: the agent pane's main conversation scroll. More specific
 // than the universal rule above, so it wins on size only (cursor/colour still
-// come from the shared rules).
+// come from the shared rules). Width driven by --agent-document-scrollbar-width
+// (declared on .agent-view, view/agent/styles/_document.scss — inherited here
+// since .agent-document is its descendant) so .agent-working-row-anchor can
+// inset itself by exactly this much and never paint over the scrollbar
+// gutter, instead of fighting it with z-index.
 .agent-document::-webkit-scrollbar {
-    width: 14px;
-    height: 14px;
+    width: var(--agent-document-scrollbar-width, 14px);
+    height: var(--agent-document-scrollbar-width, 14px);
 }
 
 *::-webkit-scrollbar-track {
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2350, opened 2026-07-29
feat(muxbus): same-host cross-channel reactive delivery (Tier 2b)

### LINKED ISSUE #1916: MuxBus: cross-channel (same-host, different channel) delivery

AgentY-asaf:
## Summary

Closes #1916. MuxBus Tier 2 same-host forwarding previously only ever reached an
agent registered in the *caller's own channel* — a file registry rooted at
`AGENTMUX_DATA_HOME`, which is per-channel. An agent in `dev-a` had no way to
reach an agent in `dev-b` running on the same machine; `SendMessage` failed with
"agent not found" even though both instances were local.

This adds a sibling **host-global shared registry** at
`~/.agentmux/shared/agents/reactive/` (one JSON file per agent name, holding an
array of one entry per channel currently registering that name) and wires it in
as **Tier 2b**, between the existing Tier 2a (same-channel) and Tier 3 (LAN):

- `handle_reactive_register`/`unregister` now mirror into the shared registry
  alongside the existing per-channel one.
- `handle_reactive_inject` tries Tier 2b candidates freshest-first when Tier 2a
  has no entry or its forward failed; a failed forward evicts just that
  channel's entry and falls through to the next candidate (same evict-on-fail
  shape Tier 3 already uses).
- `GET /agentmux/discovery` gains `host.cross_channel[]`, listing other
  channels' live agents on this host.
- Startup sweep (`cleanup_stale_shared`) additionally evicts entries whose
  owning PID is no longer alive on this host — catches a channel that crashed
  without a clean unregister, which the existing TTL-only sweep could miss for
  up to 4 hours.

Implements the design in `SPEC_MUXBUS_CROSS_CHANNEL_DELIVERY_2026_07_02.md`
(parked in the `wt-xchannel` worktree since 2026-07-02), corrected against
current code — notably using the real, already-established `AGENTMUX_SHARED_DIR`
resolver rather than the spec's proposed (and since-superseded) env var name.

Same-user trust boundary as the existing per-channel registry (0600 file
perms, loopback-only forward target, auth_key required for the forward).

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 1762 passed, 0 failed (13 new shared-registry tests: round-trip, multi-channel coexistence, replace-not-duplicate, freshest-first ordering, evict-one-channel, evict-last-channel-deletes-file, list-all, TTL eviction, dead-PID eviction, keep-fresh-live, path-traversal sanitization, pid_alive true/false)
- [x] `registry::paths` tests (existing + new `resolve_shared_reactive_dir` cases) — 11 passed

### DIFF
--- a/.changesets/1785284522-feat-muxbus-same-host-cross-channel-reactive-delivery-tier-2-8d9c.md
+++ b/.changesets/1785284522-feat-muxbus-same-host-cross-channel-reactive-delivery-tier-2-8d9c.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(muxbus): same-host cross-channel reactive delivery (Tier 2b, issue #1916)
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -920,11 +920,20 @@ impl PersistentSubprocessController {
                         agent_id = %agent_id,
                         "muxbus: auto-registered persistent agent"
                     );
-                    // Also write the cross-instance (Tier-2) file registry.
+                    // Also write the cross-instance (Tier-2) file registry,
+                    // and its host-global sibling (Tier 2b, issue #1916) —
+                    // this auto-register path bypasses the HTTP register
+                    // handler entirely, so it needs its own mirror call
+                    // exactly like that handler does.
                     if let Ok(local_url) = std::env::var("AGENTMUX_LOCAL_URL") {
                         let data_dir = crate::backend::base::get_wave_data_dir();
                         crate::backend::reactive::registry::write(
                             &data_dir,
+                            agent_id,
+                            &local_url,
+                            &self.block_id,
+                        );
+                        crate::backend::reactive::registry::write_shared_from_env(
                             agent_id,
                             &local_url,
                             &self.block_id,
@@ -1169,6 +1178,7 @@ impl PersistentSubprocessController {
                     if let Some(ref agent_id) = agent_id_wait {
                         let data_dir = crate::backend::base::get_wave_data_dir();
                         crate::backend::reactive::registry::remove(&data_dir, agent_id);
+                        crate::backend::reactive::registry::remove_shared_from_env(agent_id);
                         if let Some(sub) = crate::muxbus::cloud_subscriber::get_global_subscriber() {
                             sub.remove_agent(agent_id);
                         }
@@ -1232,6 +1242,7 @@ impl PersistentSubprocessController {
                     if let Some(ref agent_id) = agent_id_wait {
                         let data_dir = crate::backend::base::get_wave_data_dir();
                         crate::backend::reactive::registry::remove(&data_dir, agent_id);
+                        crate::backend::reactive::registry::remove_shared_from_env(agent_id);
                         if let Some(sub) = crate::muxbus::cloud_subscriber::get_global_subscriber() {
                             sub.remove_agent(agent_id);
                         }
--- a/agentmux-srv/src/backend/blockcontroller/shell/lifecycle.rs
+++ b/agentmux-srv/src/backend/blockcontroller/shell/lifecycle.rs
@@ -482,11 +482,19 @@ impl Controller for ShellController {
                         agent_id = %agent_id,
                         "jekt: auto-registered"
                     );
-                    // Also write to cross-instance file registry.
+                    // Also write to cross-instance file registry, and its
+                    // host-global sibling (Tier 2b, issue #1916) — this
+                    // auto-register path bypasses the HTTP register handler
+                    // entirely, so it needs its own mirror call too.
                     if let Ok(local_url) = std::env::var("AGENTMUX_LOCAL_URL") {
                         let data_dir = crate::backend::base::get_wave_data_dir();
                         crate::backend::reactive::registry::write(
                             &data_dir,
+                            agent_id,
+                            &local_url,
+                            &self.block_id,
+                        );
+                        crate::backend::reactive::registry::write_shared_from_env(
                             agent_id,
                             &local_url,
                             &self.block_id,
@@ -768,6 +776,7 @@ impl Controller for ShellController {
             if let Some(ref agent_id) = agent_id_wait {
                 let data_dir = crate::backend::base::get_wave_data_dir();
                 crate::backend::reactive::registry::remove(&data_dir, agent_id);
+                crate::backend::reactive::registry::remove_shared_from_env(agent_id);
                 if let Some(sub) = crate::muxbus::cloud_subscriber::get_global_subscriber() {
                     sub.remove_agent(agent_id);
                 }
--- a/agentmux-srv/src/backend/reactive/registry.rs
+++ b/agentmux-srv/src/backend/reactive/registry.rs
@@ -234,6 +234,28 @@ pub fn remove_shared(shared_dir: &Path, agent_id: &str, channel: &str) {
     }
 }
 
+/// Convenience wrapper over [`write_shared`]: resolves the shared dir and
+/// this process's channel itself, no-op if the shared root can't be
+/// resolved. Exists so call sites outside `server/reactive.rs`'s explicit
+/// register/unregister handlers (PTY shell auto-register, persistent
+/// stream-json controller auto-register — both call the per-channel
+/// `write`/`remove` directly, not through the HTTP handler) don't each
+/// repeat the same resolve-dir-then-read-env dance.
+pub fn write_shared_from_env(agent_id: &str, local_url: &str, block_id: &str) {
+    if let Some(shared_dir) = crate::registry::resolve_shared_reactive_dir() {
+        let channel = std::env::var("AGENTMUX_CHANNEL").unwrap_or_else(|_| "stable".to_string());
+        write_shared(&shared_dir, agent_id, local_url, block_id, &channel);
+    }
+}
+
+/// Convenience wrapper over [`remove_shared`] — see [`write_shared_from_env`].
+pub fn remove_shared_from_env(agent_id: &str) {
+    if let Some(shared_dir) = crate::registry::resolve_shared_reactive_dir() {
+        let channel = std::env::var("AGENTMUX_CHANNEL").unwrap_or_else(|_| "stable".to_string());
+        remove_shared(&shared_dir, agent_id, &channel);
+    }
+}
+
 /// Return every live candidate for `agent_id` across all channels,
 /// freshest-first (§4.3: Tier 2b prefers the freshest candidate). Does
 /// NOT filter by staleness/pid-liveness — callers needing that should use
--- a/agentmux-srv/src/backend/reactive/registry.rs
+++ b/agentmux-srv/src/backend/reactive/registry.rs
@@ -174,8 +174,12 @@ pub fn cleanup_stale(data_dir: &Path, max_age_ms: u64) {
 // in by the caller rather than resolved here to keep this module free of
 // a dependency on the `registry` crate module).
 //
-// Layout: `<shared_dir>/agents/<agent_name>/<channel>.json`, one file per
-// (agent, channel) pair — NOT a single JSON array shared across channels.
+// Layout: `<shared_dir>/<agent_name>/<channel>.json` (`shared_dir` here is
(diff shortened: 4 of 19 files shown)


a5af:
@codex review

a5af:
@codex review

PR #2351, opened 2026-07-29
docs(retro): dev-build test session collided with a live production agent

AgentY-asaf:
## Summary

- While live-verifying PR #2349, a throwaway `task dev` build resumed what looked like a disposable test session via the agent picker's "Continue Agent3" — it was actually the live, currently-running production conversation. Submitting a synthetic test message spawned a genuinely independent second agent process that inherited full context and autonomously continued the standing scrollbar task.
- Traced the mechanism end-to-end: agent registry/definitions/transcript storage resolves via a channel-independent `~/.agentmux/shared` root by design (so a named agent's history follows you across builds), but turn execution/ownership is purely per-process with no cross-process lock — so two processes can both believe they're free to drive the same session's next turn.
- `git reflog` on the shared working directory showed this went beyond a file collision: the second process ran an actual `git checkout -b` in the same physical clone, silently switching the branch this session's own shell commands were operating against.
- No code fix shipped here — this is a process/tooling gap documented for follow-up (cross-process session locking, dev-picker safety, and the broader shared-working-directory hazard), not a defect in the merged scrollbar PR.

## Test plan

N/A — documentation only.

### DIFF
--- a/docs/retros/RETRO_DEV_BUILD_SHARED_AGENT_SESSION_COLLISION_2026_07_29.md
+++ b/docs/retros/RETRO_DEV_BUILD_SHARED_AGENT_SESSION_COLLISION_2026_07_29.md
@@ -1,0 +1,51 @@
+# Retro: a throwaway `task dev` test build resumed a live production agent session and spawned an unaware second turn
+
+**Date:** 2026-07-29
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2355, opened 2026-07-29
feat(srv): cross-process session-ownership lease foundation

AgentY-asaf:
## Summary

PR 1 of 2 for the session-ownership lease work (see plan context below). Foundation only — inert, no behavior change.

- New `registry::LeaseStore` (`agentmux-srv/src/registry/leases.rs`): claim/renew/release a lease on an agent `instance_id`, backed by a dedicated per-session file (`<registry_root>/leases/<instance_id>.lease.json`), not new fields on the registry's own JSON record — `Registry::upsert` is a confirmed blind read-modify-write with no cross-process CAS, so a lock needs a real atomic primitive instead. Claim uses `OpenOptions::create_new`, the same exclusive-create pattern already established in this codebase for scratch-file claims (`server/editor_handlers.rs`). TTL-based expiry (15s, 3x the planned 5s renew interval) means a crashed owner's lease self-heals without any separate liveness-detection process.
- `AppState` gains a `boot_id: Arc<str>`, generated once per process boot via `uuid::Uuid::new_v4()`. No per-boot identifier existed anywhere in the codebase before this — `config.instance_id` (the `--instance` channel name) is stable across restarts and shared by every process on the same channel, which is exactly backwards for a lease owner id.

No call sites are wired up yet (turn spawn doesn't claim/renew/release anything in this PR) — that enforcement lands in a follow-up PR once this foundation is reviewed.

**Context**: this follows a real incident — a throwaway `task dev` build resumed a live production agent session and a second, unaware process ran real `git checkout` commands in the same shared working directory. Full writeups: `docs/retros/RETRO_DEV_BUILD_SHARED_AGENT_SESSION_COLLISION_2026_07_29.md` and `docs/analysis/ANALYSIS_MULTI_AGENT_SESSION_AND_WORKDIR_ISOLATION_2026-07-29.md`.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv registry::leases` — 11/11 new unit tests passing (claim/renew/release happy paths, TTL expiry reclaim, corrupt-file eviction, stale-owner protection against clobbering, an 8-thread concurrent-claim race with exactly one winner)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` (full suite) — 1757/1757 passing, run twice to confirm no flakiness (an earlier draft had two real bugs surfaced only under full-suite parallel load: a Windows-specific double-file-touch race in claim's write path, and overly tight sleep/TTL margins in three timing-sensitive tests — both fixed)

### DIFF
--- a/.changesets/1785332715-feat-srv-cross-process-session-ownership-lease-foundation-re-efno.md
+++ b/.changesets/1785332715-feat-srv-cross-process-session-ownership-lease-foundation-re-efno.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(srv): cross-process session-ownership lease foundation (registry::LeaseStore)
--- a/agentmux-srv/Cargo.toml
+++ b/agentmux-srv/Cargo.toml
@@ -82,6 +82,7 @@ windows-sys = { version = "0.59", features = [
     "Win32_System_SystemInformation",
     "Win32_System_Registry",
     "Win32_Storage_FileSystem",
+    "Win32_System_IO",
 ] }
 
 [target.'cfg(windows)'.build-dependencies]
--- a/agentmux-srv/src/bootstrap.rs
+++ b/agentmux-srv/src/bootstrap.rs
@@ -1382,6 +1382,7 @@ pub fn build_app_state(
 
     AppState {
         auth_key: config.auth_key.clone(),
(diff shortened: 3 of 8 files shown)


a5af:
@codex review

PR #2359, opened 2026-07-29
feat(srv): enforce cross-process session lease on host-mode agent turns

AgentY-asaf:
## Summary

PR 2 of 2 for the session-ownership lease work (PR 1: #2355, merged). Wires `registry::LeaseStore` into the real turn-spawn path.

- `SubprocessController::spawn_turn` (host-mode) now claims the lease for `config.instance_id` before flipping status to running. A refusal (`HeldByOther`) unwinds cleanly — releases the run lock, returns a clear error naming the other process and how long ago it last renewed — before any child process is spawned.
- Renewed every 5s for as long as the turn is active, via a small dedicated task (not by widening `spawn_health_watchdog`'s signature — that helper has 7 call sites across ACP/persistent/container/host controllers; only host-mode claims a lease in this PR).
- Released on every exit path: spawn failure, normal completion, and kill (all converge through the same process_waiter release point).
- `instance_id` (`block.meta["agentId"]`) threaded through `SubprocessSpawnConfig` and the controller constructor from every real construction site: `input.rs`'s live `agentinput` handler, `agent_io.rs`'s `agent.send` handler, and `resync_controller`'s three callers.

**Explicitly deferred** (per the approved plan, not an oversight): container-mode (`spawn_container_turn`) and persistent/background agents (`PersistentSubprocessController`) share the identical exposure but aren't wired in this PR — fast-follow once this lands.

Context: `docs/retros/RETRO_DEV_BUILD_SHARED_AGENT_SESSION_COLLISION_2026_07_29.md` (the incident this closes) and `docs/analysis/ANALYSIS_MULTI_AGENT_SESSION_AND_WORKDIR_ISOLATION_2026-07-29.md` (the design).

## Test plan

- [x] `cargo check -p agentmux-srv` / `--tests` — clean
- [x] New integration test `spawn_turn_refuses_when_lease_held_by_another_process`: pre-claims a lease under a different `boot_id` in a real temp registry, then asserts `spawn_turn` refuses with "already owned by another AgentMux process" *before* any child process spawns, and that the run lock is released on refusal (so a legitimate retry against the same process isn't permanently blocked by its own busy-lock).
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` (full suite) — 1765/1765 passing, run twice for stability

### DIFF
--- a/.changesets/1785335800-feat-srv-enforce-cross-process-session-lease-on-host-mode-ag-7h8c.md
+++ b/.changesets/1785335800-feat-srv-enforce-cross-process-session-lease-on-host-mode-ag-7h8c.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(srv): enforce cross-process session lease on host-mode agent turns
--- a/agentmux-srv/src/backend/blockcontroller/mod.rs
+++ b/agentmux-srv/src/backend/blockcontroller/mod.rs
@@ -359,6 +359,8 @@ pub fn resync_controller(
     event_bus: Option<Arc<EventBus>>,
     wstore: Option<Arc<Store>>,
     filestore: Option<Arc<FileStore>>,
+    registry: Option<Arc<crate::registry::Registry>>,
+    boot_id: Arc<str>,
 ) -> Result<(), String> {
     let block_id = &block.oid;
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #2360, opened 2026-07-29
fix(agent): transparently retry a stale --resume session id on a persistent controller's first message

AgentA-asaf:
## Summary
- Fixes a live-testing find on the fresh `0.54.7` portable build: sending the first message to an already-logged-in, globally-known agent ("Lzop") failed with a generic **"Agent encountered an error"** and the message was lost.
- Root cause: agent session ids are deliberately global (follow the user across every build/channel), but each channel's CLI binary + its own native session storage are provisioned fresh per version. The very first resume of an existing session under a brand-new channel's CLI install fails immediately (`No conversation found with session ID: ...`, exit code 1) before producing any response. The backend already detects this exact case and clears the stale id — but only the *next* send benefited; the triggering message itself was lost.
- Adds a transparent retry: `send_message` stashes the (config, formatted stdin line) for a fresh spawn's resume attempt. If the stderr reader confirms that session id unreachable, the process-waiter task (via a new weak self-reference, mirroring `SubprocessController`'s existing queued-message-drain pattern) respawns fresh with no `--resume` and redelivers the exact same message — without re-persisting it or re-emitting `agent-message-accepted` (both already happened correctly on the original attempt).
- Full investigation write-up: `docs/retros/RETRO_STALE_RESUME_SESSION_ID_ACROSS_CHANNELS_2026_07_29.md` (a sibling finding to `RETRO_DEV_BUILD_SHARED_AGENT_SESSION_COLLISION_2026_07_29.md` from the same day — same root tension: agent identity/session data is deliberately global, but a narrower-scoped execution layer isn't always aware of it).
- Unrelated to PR #2338 (fast-fail-send-while-unauthenticated, merged earlier today) — that PR's entire review surface was the ACP controller's auth-guard/deferred-refresh machinery; this is the **persistent** controller's session-resume path, a completely different codepath that PR never touched.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv blockcontroller::persistent` — 19/19 passed, including 3 new tests covering `pending_resume_retry`'s set/clear lifecycle
- [x] Revert-and-verify: temporarily reverted `try_capture_session_id`'s new clearing line, confirmed both new "cleared on success" tests fail with the exact expected assertion, restored, confirmed all 19 pass again
- [x] `cargo test -p agentmux-srv --bin agentmux-srv blockcontroller` — 119/119 passed (no regressions across the whole module)
- [x] `cargo build -p agentmux-srv` — clean, no new warnings
- [ ] Fresh portable build + live resend of the exact failing scenario (first message to a globally-known agent from a brand-new channel)

### DIFF
--- a/.changesets/1785340373-fix-agent-transparently-retry-a-stale-resume-session-id-on-a-gx9b.md
+++ b/.changesets/1785340373-fix-agent-transparently-retry-a-stale-resume-session-id-on-a-gx9b.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): transparently retry a stale --resume session id on a persistent controller's first message
--- a/agentmux-srv/src/backend/blockcontroller/mod.rs
+++ b/agentmux-srv/src/backend/blockcontroller/mod.rs
@@ -465,6 +465,7 @@ pub fn resync_controller(
                 filestore,
             );
             let ctrl = Arc::new(ctrl);
+            ctrl.set_self_ref();
             register_controller(block_id, ctrl.clone());
             ctrl.start(block_meta.clone(), rt_opts, force)
         }
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -165,6 +165,19 @@ struct PersistentInner {
     /// Cleared on kill and on a normal exit for the same
     /// reused-instance reason as `pending_resume_retry`.
     confirmed_stale_resume_retry: Option<(String, PersistentSpawnConfig, String)>,
+    /// Bumped by `spawn_process` every time it successfully spawns a child
+    /// (the original send_message-triggered spawn, or a retry's own
+    /// respawn). `send_message` captures this right after its own spawn
+    /// attempt (if any) and re-checks it just before writing to
+    /// `stdin_tx` — codex P1 on PR #2360 (third review pass): a
+    /// fast-failing stale-resume spawn can let the process-waiter task
+    /// confirm and install a retry's brand-new process (with its own
+    /// fresh `stdin_tx`) BEFORE `send_message`'s own later write runs.
+    /// If the epoch has moved on, that write is skipped — the retry
+    /// already delivers this exact message on the new process, so
+    /// writing here too would duplicate it, and erroring on a
+    /// mid-swap `stdin_tx` would falsely report a failed send.
+    spawn_epoch: u64,
     current_pid: Option<u32>,
     /// Channel to send messages to the stdin writer task.
     stdin_tx: Option<mpsc::Sender<String>>,
@@ -321,6 +334,7 @@ impl PersistentSubprocessController {
                 resume_poisoned: None,
                 pending_resume_retry: None,
                 confirmed_stale_resume_retry: None,
+                spawn_epoch: 0,
                 current_pid: None,
                 stdin_tx: None,
                 kill_tx: None,
@@ -510,6 +524,21 @@ impl PersistentSubprocessController {
         if is_fresh_spawn {
             self.spawn_process(config.clone(), Some(json_str.clone()))?;
         }
+        // Captured AFTER the spawn above (if any) so it reflects whatever
+        // is CURRENT at this point — see the tail-end check below and
+        // `spawn_epoch`'s own doc comment. codex P1 on PR #2360 (third
+        // review pass): on a fast-failing stale-resume spawn, the
+        // process-waiter task (on another thread) can detect the exit,
+        // confirm the retry, and install a BRAND NEW process — with its
+        // own fresh `stdin_tx` — before this function's own later stdin
+        // write below ever runs. Without tracking which spawn attempt this
+        // call's write belongs to, that write would either duplicate the
+        // message on the new process (if it read the new stdin_tx) or
+        // spuriously report a failed send (if it raced the narrow window
+        // where the old stdin_tx was already cleared but the new one
+        // wasn't installed yet) — even though the retry already delivers
+        // this exact message on its own.
+        let my_spawn_epoch = self.inner.lock().unwrap().spawn_epoch;
         // spawn_process already marks a fresh process's first turn active
         // (and starts its watchdog); for an already-running process (the
         // common case — every turn after the first) this is the only place
@@ -552,6 +581,18 @@ impl PersistentSubprocessController {
         );
 
         let inner = self.inner.lock().unwrap();
+        if inner.spawn_epoch != my_spawn_epoch {
+            // A NEWER spawn (a stale-resume retry, or any other respawn)
+            // has already taken over delivery for this exact attempt —
+            // `retry_after_resume_failure` already sends this SAME
+            // already-formatted `json_str` on the new process. Skip our
+            // own write entirely rather than duplicating it or risking a
+            // false "not running" error against a `stdin_tx` mid-swap.
+            // codex P1 on PR #2360 (third review pass).
+            drop(inner);
+            self.emit_message_accepted(config.message_id.as_deref());
+            return Ok(());
+        }
         let tx = inner.stdin_tx.as_ref()
             .ok_or("persistent process not running after spawn")?;
         tx.try_send(json_str)
@@ -933,23 +974,35 @@ impl PersistentSubprocessController {
             format!("failed to spawn persistent process: {e}")
         })?;
 
-        // Stash the TENTATIVE resume-retry payload synchronously, right
-        // here — before any background task (stdin writer, stdout/stderr
-        // readers, process-waiter) is created below. reagentx P1 on PR
-        // #2360: stashing this later, back in send_message after this
-        // function returned, left a window where a process that dies fast
-        // enough (the exact case this exists to catch) lets the
-        // process-waiter task — already racing on another thread once it's
-        // spawned — observe the exit and take() this payload while it's
-        // still `None`, silently losing the retry for the very case it's
-        // meant to catch. Keyed on the EXACT sid this spawn attempted (not
-        // `config.session_id`, which can differ from what's actually held
-        // in `inner.session_id` once an earlier call has already
-        // hydrated it) so `poison_resume`'s later confirmation check is
-        // unambiguous.
-        if let (Some(sid), Some(retry_json)) = (attempted_resume_sid.clone(), resume_retry_payload) {
+        // Bump spawn_epoch and stash the TENTATIVE resume-retry payload
+        // synchronously, right here — before any background task (stdin
+        // writer, stdout/stderr readers, process-waiter) is created below.
+        //
+        // The epoch bump (see its own doc comment) marks THIS as the
+        // newest spawn attempt for this controller — codex P1 on PR #2360
+        // (third review pass): without it, send_message's own later stdin
+        // write (after this function returns) has no way to notice that a
+        // stale-resume retry already installed a NEWER process (and
+        // `stdin_tx`) in the meantime, and would either duplicate the
+        // message on it or spuriously report a failed send.
+        //
+        // The retry-payload stash (reagentx P1 on PR #2360): doing this
+        // later, back in send_message after this function returned, left a
+        // window where a process that dies fast enough (the exact case
+        // this exists to catch) lets the process-waiter task — already
+        // racing on another thread once it's spawned — observe the exit
+        // and take() this payload while it's still `None`, silently losing
+        // the retry for the very case it's meant to catch. Keyed on the
+        // EXACT sid this spawn attempted (not `config.session_id`, which
+        // can differ from what's actually held in `inner.session_id` once
+        // an earlier call has already hydrated it) so `poison_resume`'s
+        // later confirmation check is unambiguous.
+        {
             let mut inner = self.inner.lock().unwrap();
-            inner.pending_resume_retry = Some((sid, config.clone(), retry_json));
+            inner.spawn_epoch += 1;
+            if let (Some(sid), Some(retry_json)) = (attempted_resume_sid.clone(), resume_retry_payload) {
+                inner.pending_resume_retry = Some((sid, config.clone(), retry_json));
+            }
         }
 
         let pid = child.id().unwrap_or(0);
@@ -1856,6 +1909,46 @@ mod send_input_tests {
              not race ahead of it"
         );
     }
+
+    /// codex P1 on PR #2360 (third review pass) added a `spawn_epoch` check
+    /// to `send_message`'s own stdin write, to detect a stale-resume retry
+    /// having already installed a newer process in the meantime. This
+    /// confirms that check is a pure no-op in the overwhelmingly common
+    /// case — no concurrent retry raced ahead — so an already-running
+    /// process still receives its message exactly as before. The actual
+    /// race this check exists to close (a genuine multi-threaded
+    /// interleaving between this function's own two lock acquisitions) is
+    /// not practical to inject deterministically without a test-only hook
+    /// in production code; verified via direct code inspection instead —
+    /// the check reads the SAME `inner.spawn_epoch` both times, so any
+    /// intervening bump (which only `spawn_process` performs, and only
+    /// while holding the same lock) is unconditionally visible.
+    #[tokio::test]
+    async fn send_message_still_delivers_normally_when_the_epoch_has_not_moved() {
+        let c = controller();
+        let (tx, mut rx) = mpsc::channel::<String>(4);
+        {
+            let mut inner = c.inner.lock().unwrap();
+            inner.stdin_tx = Some(tx);
+        }
+
+        let config = PersistentSpawnConfig {
+            cli_command: "unused".to_string(),
+            cli_args: vec![],
+            working_dir: String::new(),
+            env_vars: HashMap::new(),
+            session_id_field: "session_id".to_string(),
+            resume_flag: String::new(),
+            session_id: String::new(),
+            message_id: None,
+        };
+
+        c.send_message("hello".to_string(), config)
+            .expect("delivery to an already-running process must succeed");
+
+        let received = rx.try_recv().expect("the message must have been written to stdin_tx");
+        assert!(received.contains("hello"));
+    }
 }
 
 /// Covers the stderr-poison / stdout-capture race directly against
@@ -1874,6 +1967,7 @@ mod resume_poison_tests {
             resume_poisoned: None,
             pending_resume_retry: None,
             confirmed_stale_resume_retry: None,
+            spawn_epoch: 0,
             current_pid: None,
             stdin_tx: None,
             kill_tx: None,
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -702,15 +702,34 @@ impl PersistentSubprocessController {
             // behind it (from other callers already told "accepted"),
             // hand off to a bounded fallback respawn rather than
             // stranding it with nobody responsible for delivering it.
+            //
+            // codex P2 on PR #2360 (round 13, commit e9678091f): clearing
+            // `spawning_in_progress` in a SEPARATE, later lock acquisition
+            // left a window between the emptiness check above and that
+            // clear where a concurrent `send_message` could observe
+            // `spawning_in_progress` still `true`, enqueue its message via
+            // `decide_send_action`'s `Queued` branch, and be told
+            // "accepted" — then this function's second lock would clear
+            // the claim without ever rechecking the queue, stranding that
+            // accepted message with no spawner and no drain ever
+            // responsible for it. The emptiness check and the flag clear
+            // must be one atomic decision under a single lock acquisition
+            // (same shape as the round-9 regression this whole PR already
+            // fixed once): only clear the claim here if the queue is STILL
+            // empty at the exact moment we're about to clear it; otherwise
+            // leave the claim held and hand off to the fallback respawn.
             let leftovers = {
                 let mut inner = self.inner.lock().unwrap();
                 inner.pending_send_messages.pop_front();
-                !inner.pending_send_messages.is_empty()
+                if inner.pending_send_messages.is_empty() {
+                    inner.spawning_in_progress = false;
+                    false
+                } else {
+                    true
+                }
             };
             if leftovers {
                 self.respawn_once_for_leftover_queue(retry_config);
-            } else {
-                self.inner.lock().unwrap().spawning_in_progress = false;
             }
             return;
         }
@@ -3164,6 +3183,76 @@ mod send_input_tests {
             "queued-by-someone-else",
             "a message queued by a DIFFERENT caller (already told \"accepted\") must survive for the next spawn"
         );
+    }
+
+    /// Exercises the actual race with real OS threads — codex P2 on PR
+    /// #2360 (round 13, commit e9678091f): a FAILED spawn's claim used to
+    /// be released in a lock acquisition SEPARATE from the emptiness check
+    /// that decided whether to release it at all. That left a window,
+    /// between the two, where a concurrent `send_message` could observe
+    /// `spawning_in_progress` still `true`, enqueue via `decide_send_action`'s
+    /// `Queued` branch, and be told "accepted" — then this function's
+    /// second lock would clear the claim without ever rechecking the
+    /// queue, stranding that accepted message with nobody left responsible
+    /// (no drain, no respawn, no disclosure). The fix merges the emptiness
+    /// check and the flag clear into one lock acquisition, so a racer's
+    /// push can now only land fully before or fully after that atomic
+    /// block — never inside it. Across many iterations and threads, the
+    /// invariant that must always hold: whenever a message is left queued
+    /// with the claim released, it's because a fallback respawn was
+    /// actually attempted (and disclosed its failure via a published
+    /// status) — never silently, with no attempt at all.
+    #[test]
+    fn release_spawn_claim_and_drain_queue_never_silently_strands_a_racing_send() {
+        use std::sync::Arc as StdArc;
+
+        for iteration in 0..30 {
+            let broker = StdArc::new(crate::backend::wps::Broker::new());
+            let block_id = format!("block-race-{iteration}");
+            let c = StdArc::new(PersistentSubprocessController::new(
+                "tab".to_string(),
+                block_id.clone(),
+                Some(broker.clone()),
+                None,
+                None,
+                None,
+            ));
+            c.set_self_ref();
+            {
+                let mut inner = c.inner.lock().unwrap();
+                inner.spawning_in_progress = true;
+                inner.pending_send_messages.push_back(QueuedMessage::fresh("the-one-that-failed".to_string()));
+            }
+
+            let handles: Vec<_> = (0..8)
+                .map(|i| {
+                    let c = StdArc::clone(&c);
+                    std::thread::spawn(move || {
+                        let _ = c.decide_send_action(&format!("racer-{iteration}-{i}"), false);
+                    })
+                })
+                .collect();
+
+            c.release_spawn_claim_and_drain_queue(false, unreachable_fallback_config());
+
+            for h in handles {
+                h.join().unwrap();
+            }
+
+            let inner = c.inner.lock().unwrap();
+            let stranded = !inner.spawning_in_progress && !inner.pending_send_messages.is_empty();
+            if stranded {
+                let published = !broker
+                    .read_event_history(crate::backend::wps::EVENT_CONTROLLER_STATUS, &format!("block:{block_id}"), 10)
+                    .is_empty();
+                assert!(
+                    published,
+                    "iteration {iteration}: a racing send was left queued with the claim already \
+                     released and no respawn attempt ever disclosed via a status publish — stranded \
+                     with nobody responsible for it"
+                );
+            }
+        }
     }
 
     /// codex P2 on PR #2360 (sixth review pass, round 3): `retry_after_
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -548,12 +548,28 @@ impl PersistentSubprocessController {
     ///   `json_str` alongside that claim, and returns `BecomeSpawner` —
     ///   the caller must then call `spawn_process` and, regardless of
     ///   outcome, call `release_spawn_claim_and_drain_queue`.
-    fn decide_send_action(&self, json_str: &str) -> SendAction {
+    ///
+    /// `skip_if_already_queued` — always `false` for a genuine new message
+    /// (`send_message`): a user legitimately re-sending the exact same
+    /// text while an unrelated spawn is in flight must still queue both,
+    /// so this must never dedup by content there. `true` only for
+    /// `retry_after_resume_failure`, whose `json_str` is a KNOWN re-
+    /// delivery of content that may ALREADY be sitting in
+    /// `pending_send_messages` — pushed by the very spawn attempt whose
+    /// failure triggered this retry, if that spawn's own drain hasn't
+    /// reached it yet. codex P1 on PR #2360 (sixth review pass, round 4):
+    /// blindly queueing another copy there let a fallback spawn eventually
+    /// deliver the same prompt twice.
+    fn decide_send_action(&self, json_str: &str, skip_if_already_queued: bool) -> SendAction {
         let mut inner = self.inner.lock().unwrap();
         if inner.stdin_tx.is_some() && !inner.spawning_in_progress {
             SendAction::DeliverDirect
         } else if inner.spawning_in_progress {
-            inner.pending_send_messages.push_back(json_str.to_string());
+            let already_queued =
+                skip_if_already_queued && inner.pending_send_messages.iter().any(|m| m == json_str);
+            if !already_queued {
+                inner.pending_send_messages.push_back(json_str.to_string());
+            }
             SendAction::Queued
         } else {
             inner.spawning_in_progress = true;
@@ -762,6 +778,23 @@ impl PersistentSubprocessController {
         );
     }
 
+    /// Persists a formatted stdin JSON line to the blockfile + global zone
+    /// so `parseHistoryLines` can reconstruct the `user_message` node on
+    /// the next pane open. No WPS event is published here — the
+    /// live-display is handled by the `agent-message-accepted` path (UUID
+    /// node), avoiding a duplicate.
+    fn persist_message_to_blockfile(&self, json_str: &str) {
+        let global_zone = super::shell::resolve_global_output_zone(&self.wstore, &self.block_id);
+        let line_with_newline = format!("{json_str}\n");
+        super::shell::persist_to_blockfile_silent(
+            &self.block_id,
+            crate::backend::agent_session::OUTPUT_FILE,
+            line_with_newline.as_bytes(),
+            self.filestore.as_ref(),
+            global_zone.as_deref(),
+        );
+    }
+
     pub fn send_message(&self, message: String, config: PersistentSpawnConfig) -> Result<(), String> {
         // Format as stream-json user message.
         let json_msg = serde_json::json!({
@@ -773,25 +806,14 @@ impl PersistentSubprocessController {
         });
         let json_str = json_msg.to_string();
 
-        // Silently persist the user message to the blockfile + global zone so
-        // `parseHistoryLines` can reconstruct `user_message` nodes on the next
-        // open. No WPS event is published here — the live-display is handled by
-        // the `agent-message-accepted` path (UUID node), avoiding a duplicate.
-        // Done unconditionally, before deciding delivery below, so a message
-        // that ends up queued (see `decide_send_action`) is still durably
-        // persisted immediately rather than only once eventually delivered.
-        let global_zone = super::shell::resolve_global_output_zone(&self.wstore, &self.block_id);
-        let line_with_newline = format!("{json_str}\n");
-        super::shell::persist_to_blockfile_silent(
-            &self.block_id,
-            crate::backend::agent_session::OUTPUT_FILE,
-            line_with_newline.as_bytes(),
-            self.filestore.as_ref(),
-            global_zone.as_deref(),
-        );
-
-        match self.decide_send_action(&json_str) {
+        match self.decide_send_action(&json_str, false) {
             SendAction::Queued => {
+                // Persisted unconditionally: this caller was already told
+                // "accepted" (below) and a spawn/drain genuinely IS
+                // working through the queue this just joined, so — unlike
+                // the BecomeSpawner case — there is no "rejected before
+                // ever really being attempted" outcome to guard against.
+                self.persist_message_to_blockfile(&json_str);
                 self.emit_message_accepted(config.message_id.as_deref());
                 Ok(())
             }
@@ -803,6 +825,7 @@ impl PersistentSubprocessController {
                 // since the persistent process never exits between turns.
                 // Without this, `turn_active` would go stale after turn 1.
                 self.mark_turn_active_and_publish();
+                self.persist_message_to_blockfile(&json_str);
                 let inner = self.inner.lock().unwrap();
                 let tx = inner.stdin_tx.as_ref()
                     .ok_or("persistent process not running after spawn")?;
@@ -828,15 +851,20 @@ impl PersistentSubprocessController {
                 // `--resume`, which is what the retry payload is for.
                 let message_id = config.message_id.clone();
                 let retry_config = config.clone();
-                let spawn_result = self.spawn_process(config, Some(json_str));
-                // Only emit "accepted" on success — matches this
-                // function's prior (pre-queue) behavior, which propagated
-                // a spawn failure via `?` before ever reaching the
-                // "accepted" emission. A failed spawn's message is
-                // discarded by `release_spawn_claim_and_drain_queue` below
-                // (see its own doc comment), so telling the frontend
-                // "accepted" for it here would be a lie.
+                let spawn_result = self.spawn_process(config, Some(json_str.clone()));
+                // Only persist/emit "accepted" on success — codex P2 on PR
+                // #2360 (sixth review pass, round 4): persisting
+                // unconditionally (an earlier cut of this fix did) let a
+                // rejected spawn (missing executable, bad launch config)
+                // leave a "user_message" line in the blockfile for a
+                // prompt that was NEVER actually delivered — reopening the
+                // pane would reconstruct it as if it had been sent, and a
+                // manual retry then produced a misleading duplicate.
+                // Matches this function's pre-queue behavior, which
+                // propagated a spawn failure via `?` before ever reaching
+                // either the persist or the "accepted" emission.
                 if spawn_result.is_ok() {
+                    self.persist_message_to_blockfile(&json_str);
                     self.mark_turn_active_and_publish();
                     self.emit_message_accepted(message_id.as_deref());
                 }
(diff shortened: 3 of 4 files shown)


a5af:
@codex review

AgentA-asaf:
Fixed in e361d5855, addressing all three findings from reagentx's first review pass:

**P0 — `retry_after_resume_failure` only cleared `config.session_id`, not `inner.session_id`, with no ordering guarantee vs. `poison_resume`**: correct — `poison_resume` (stderr-reader task) and this retry (process-waiter task) are independently scheduled with no ordering guarantee, so the retry could re-attach `--resume` to the same dead id if it won the race. Fixed by clearing `inner.session_id` directly inside `retry_after_resume_failure`, not relying on `poison_resume`'s timing.

**P1 — the retry fired on ANY exit before the first session id capture, not just the confirmed stale-resume case**: correct — this would have silently discarded the user's existing conversation on an unrelated auth/network/rate-limit failure, with no disclosure. Fixed by splitting the state into a TENTATIVE `pending_resume_retry` (a resume was attempted, outcome unknown) and a new CONFIRMED `confirmed_stale_resume_retry`, which `poison_resume` only sets when the sid it just confirmed dead matches the one the tentative record was captured for. The process-waiter now only ever acts on the confirmed field.

**P1 — `pending_resume_retry` was stashed in `send_message` after `spawn_process` returned and its process-waiter task was already `tokio::spawn`'d, racing a fast-failing child process**: correct — moved the stash inside `spawn_process` itself, synchronously, before any background task (including the waiter) is created. `spawn_process` is a plain (non-async) function, so this ordering is now guaranteed by construction rather than a runtime race.

5 new/updated tests cover the state-machine fixes (revert-and-verified: temporarily reverted each, confirmed the relevant test fails with the exact expected assertion, restored, confirmed all 22 pass again). The stash-timing fix is verified via direct code inspection rather than a runtime test — the synchronous, non-async body of `spawn_process` makes the ordering a compile-time guarantee, not something a test could meaningfully toggle.

`cargo check`/`cargo build -p agentmux-srv`: clean, no new warnings. `cargo test -p agentmux-srv --bin agentmux-srv blockcontroller`: 122/122 passed.

a5af:
@codex review

AgentA-asaf:
Fixed in d9d0a65ac, addressing all three findings from codex's second review pass:

**P1 — a fast-failing process could let the process-waiter wipe the tentative retry before the stderr reader ever promotes it**: correct — `child.wait()` resolving and the stderr reader processing the "No conversation found" line are independently scheduled with no ordering guarantee. If the waiter ran first, my own defensive `pending_resume_retry = None` (added to avoid leaking tentative state into a reused controller instance) would delete the payload before `poison_resume` ever got to promote it, permanently losing the retry for the exact case it exists to catch.

**P2 — publishing the terminal status before deciding on the retry let the UI treat a to-be-retried failure as a real turn end**: correct — the mounted UI would process `done`/`turn_active:false` synchronously (activity summaries, a deferred controller refresh) before the retry's own fresh "running" status ever landed.

**P2 — the stderr reader's own `persist_session_id("")` call could run after a confirmed retry's fresh session id was already persisted, silently overwriting it**: correct — this could break continuity on a later pane reopen despite the retry having actually succeeded.

All three share the same root cause, so one fix closes them all: the stderr reader's `JoinHandle` is now captured and awaited (bounded by a 500ms timeout — generous, since the pipe drains almost immediately once the process has genuinely exited) before the process-waiter makes ANY decision — reading `confirmed_stale_resume_retry`, clearing `pending_resume_retry`, or publishing a terminal status. By the time that wait completes, the stderr task's entire body (`poison_resume`, `persist_session_id("")`, the disclosure banner) is guaranteed to have already run if it was ever going to. The terminal-status publish now only happens on the non-retry path.

Added a lightweight test of the core synchronization primitive (a `JoinHandle` awaited under a bounded timeout is fully awaited when it completes well within the bound). A full subprocess-based integration test proving the exact race ordering isn't practical to write reliably cross-platform (this file's own established precedent avoids spawning real CLI processes for this subsystem); verified via direct code inspection — the await now unconditionally precedes every decision point it needs to guard.

`cargo build -p agentmux-srv`: clean. `cargo test -p agentmux-srv --bin agentmux-srv blockcontroller`: 123/123 passed.

a5af:
@codex review

AgentA-asaf:
Fixed in 772132692, addressing codex's third-review-pass finding:

**P1 — `send_message`'s own tail-end stdin write can race against a retry it just triggered**: correct. On a fast-failing stale-resume spawn, the process-waiter task (on another thread) can detect the exit, confirm the retry, and install a brand-new process — with its own fresh `stdin_tx` — before `send_message`'s own later write ever runs. Depending on timing, this either duplicates the message on the new process, or spuriously reports a failed send during the narrow window where the old `stdin_tx` was already cleared but the new one wasn't installed yet — even though the retry successfully delivers the message moments later.

Fixed with a monotonic `spawn_epoch` on `PersistentInner`, bumped by `spawn_process` every time it successfully spawns (the original attempt or a retry's own respawn). `send_message` captures the epoch right after its own spawn attempt, then re-checks it just before writing to `stdin_tx`: if the epoch moved on, a newer spawn already owns delivery for this exact attempt, so `send_message` skips its own write (no duplicate) and returns `Ok` (no false failure) instead of erroring.

One new deterministic test confirms the check is a pure no-op in the common (non-racing) case. The actual race — genuine multi-threaded interleaving between `send_message`'s own two lock acquisitions — isn't practical to inject without a test-only hook in production code; verified via direct code inspection, matching this PR's two earlier hard-to-test concurrency fixes.

`cargo build -p agentmux-srv`: clean. `cargo test -p agentmux-srv --bin agentmux-srv blockcontroller`: 124/124 passed.

a5af:
@codex review

AgentA-asaf:
Fixed in 7c426575d, addressing codex's fourth-review-pass finding — a gap in the epoch fix from the previous commit:

**P1 — `my_spawn_epoch` was captured via a separate lock acquisition after `spawn_process` returned, not atomically**: correct. If a stale-resume child exits and its retry installs the replacement process in the window between `spawn_process` returning and that later read, `my_spawn_epoch` captures the REPLACEMENT's epoch, not the original attempt's — the tail-end comparison then always "matches" (both reads see the same already-bumped value), so `send_message` proceeds to duplicate the message on the replacement process even though the retry already delivered it there. This exactly reproduces the bug the epoch check exists to prevent, just via a different timing window than the one already fixed.

Fixed exactly as suggested: `spawn_process` now returns the epoch value it assigned (`Result<u64, String>`) as part of the same atomic lock acquisition that bumps it, and `send_message` binds to that return value directly for its own fresh-spawn case — there's no longer any intervening read for a race to land in. The already-running (non-fresh-spawn) case is unaffected, since nothing tied to that message's own delivery could race ahead of it.

No new test for this specific commit — it closes a gap in an already direct-inspection-verified concurrency fix (this PR's third), and the atomicity guarantee (no read between a value being produced and being used) is now enforced by the type system itself (the value is a function return, not a separate read), not something a test could usefully assert on beyond the existing coverage.

`cargo build -p agentmux-srv`: clean. `cargo test -p agentmux-srv --bin agentmux-srv blockcontroller`: 124/124 passed.

a5af:
@codex review

AgentA-asaf:
Fixed in d833772f4, addressing all three findings from codex's fifth review pass:

**P1 — the epoch-skip check fired for ANY message observing a mismatch, not just the one whose own fresh spawn was superseded**: correct, and the most severe finding on this PR so far — a second message arriving on an already-running (soon-to-be-replaced) process never has its content captured into `pending_resume_retry` (only a fresh-spawn attempt does). If it happened to observe the SAME retry's epoch bump, it would silently skip its own delivery — assuming the retry handles it, which it doesn't (the retry only redelivers the FIRST message's payload) — while still reporting success and emitting `agent-message-accepted`. Worse than a duplicate: the content vanishes with no signal at all. Fixed by gating the skip on `is_fresh_spawn`, extracted into a pure `should_skip_own_delivery` function so the exact condition is directly, deterministically unit-testable (new test, revert-and-verified).

**P1 — the stderr-reader `JoinHandle` timeout only stopped waiting, it didn't cancel the task**: correct — if `persist_session_id` blocked past 500ms, the still-running task could later call `poison_resume`/`persist_session_id("")` AFTER the process-waiter had already moved on, either losing an unpromoted retry or overwriting a confirmed retry's fresh session id after the fact. Fixed by capturing an `AbortHandle` before the timeout and actually aborting the task if it elapses.

**P2 — a failed retry respawn left the pane hanging forever with no signal**: correct — the outer waiter had already suppressed its own terminal-status publish (a retry was in flight), and `send_message` had already returned success, so nothing ever told the frontend the turn was over if the retry itself then failed to spawn. Fixed by publishing the terminal status (already `STATUS_DONE`/`turn_active:false` from the original exit's own cleanup) in that failure branch.

One new deterministic test for the first fix, revert-and-verified (temporarily reverted the `is_fresh_spawn` gate, confirmed the test fails with the exact expected assertion, restored, confirmed all 25 tests pass again). The other two are concurrency-timing fixes verified via direct code inspection, matching this PR's established precedent.

`cargo build -p agentmux-srv`: clean. `cargo test -p agentmux-srv --bin agentmux-srv blockcontroller`: 125/125 passed.

a5af:
@codex review

AgentA-asaf:
Re: the sixth-review-pass finding (\`send_message\`'s \`is_running()\`/\`spawn_process()\` not being atomic, allowing a concurrent second spawn) — this is genuine, and reagentx's framing is accurate: this exact TOCTOU race was already theoretically reachable before this PR (any two truly-simultaneous \`send_message\` calls could hit it), but the automatic stale-resume retry makes the window reliably sub-second instead of requiring human-timescale coincidence.

A proper fix mirrors \`SubprocessController\`'s existing \`run_lock\` (\`AtomicBool\` + \`compare_exchange\`) and pending-message-queue pattern (\`subprocess/mod.rs\`, \`host_spawn.rs\`) — a real subsystem addition, not a narrow timing-window fix like the five findings this review already closed on this PR. Given how deep this review cycle has already gone, I'm deferring this specific gap to a tracked follow-up rather than adding a new subsystem this late: [#2362](https://github.com/agentmuxai/agentmux/issues/2362), which documents the exact mechanism and points at \`SubprocessController\`'s pattern as the model fix.

This is a pre-existing architectural gap, not a new regression introduced by this PR's own retry feature — it just became easier to trigger. The five concurrency findings from rounds 1–5 (session-id race, stderr/waiter sync, send_message/retry delivery, epoch atomicity, and the message-loss/hang/corruption trio) are all fixed and verified. Proceeding to merge on that basis; #2362 tracks the remaining gap for dedicated follow-up work.

AgentA-asaf:
Implemented the full fix for the concurrent-spawn TOCTOU race (round 6) rather than deferring it — closes issue #2362.

Added `PersistentInner::spawning_in_progress` + `pending_send_messages`, and a `decide_send_action`/`release_spawn_claim_and_drain_queue` pair that atomically decides "deliver directly / become the exclusive spawner / queue behind another in-flight spawn" under one lock acquisition — mirroring `SubprocessController`'s existing `run_lock`/`pending_messages` pattern. `retry_after_resume_failure` now goes through the same decision function, since its own `spawn_process` call is exactly the other side of this race.

This also let me retire the `spawn_epoch`/`should_skip_own_delivery` mechanism from rounds 3-5: that check only existed to catch a delivery landing in the window between a caller's own `spawn_process` returning and its own tail-end stdin write. With the new spawn-claim lock, that window no longer exists — delivery now happens as part of the same atomic claim, before the lock releases — so the epoch check became a permanent no-op. Removed it along with its dedicated tests, replacing them with tests for the new mechanism, including one that exercises the actual race with 16 real OS threads (not just sequential state assertions) and asserts exactly one becomes the spawner.

Full `blockcontroller` suite (135 tests) and the full crate suite (1778 tests) pass; `cargo build`/`cargo clippy` clean on this file.

Will close #2362 once this lands.

a5af:
@codex review

AgentA-asaf:
Addressed both new findings from the last codex pass on commit 0266632222:

1. **P1 (line 574, "Remove failed-spawn messages before returning")**: confirmed real — a failed spawn left its own message queued, so an unrelated later successful spawn could silently execute (or duplicate) a prompt the caller was already told had failed. Fixed: `release_spawn_claim_and_drain_queue` now takes `spawn_succeeded: bool` and on failure discards exactly the front item (the failed spawner's own message, provably at the front — nothing else can be queued before `spawning_in_progress` is set), leaving anything queued after it (other callers already told "accepted") intact. Also stopped emitting `agent-message-accepted` for a failed spawn's own message, matching this function's pre-refactor behavior.

2. **P2 (line 589, "Requeue messages when the stdin channel is full")**: confirmed real — the synchronous `try_send` drain loop discarded a message outright on `Full` instead of backpressuring. Fixed: on a successful spawn, the drain now runs on a background task using `Sender::send` (awaits capacity) instead of `try_send`, so a burst of concurrent callers can no longer silently lose input.

New/updated tests cover both cases. Full suite (1778 tests) + clippy clean. Pushed as 30b2d8cb5.

a5af:
@codex review

AgentA-asaf:
Addressed the three new findings from reagentx (commit 0266632222) and codex (commit 30b2d8cb5) — both converged on the same core gap, plus one more:

1. **reagentx/codex P1 ("Respawn when releasing a nonempty queue")**: confirmed real. A fast-dying child can let the process-waiter run `retry_after_resume_failure` (which gets `SendAction::Queued`, since the original spawn's claim hadn't been released yet) *before* that original caller's own thread reaches `release_spawn_claim_and_drain_queue`. My previous fix (publishing status when stalled) only closed the "indefinite hang" half of this — it didn't actually get the message delivered. Now `release_spawn_claim_and_drain_queue` and its background drain hand off to a new `respawn_once_for_leftover_queue` in this exact situation: one bounded fallback spawn attempt (session_id forced empty, so it can't repeat a stale `--resume`), success drains normally, failure publishes status. Bounded to exactly one extra hop so a second stall can't cascade indefinitely.

2. **codex P2 ("Hand off queued messages after a spawn failure")**: same root gap, on the synchronous-failure side (a message queued behind an in-flight spawn that then fails). Same fix applies — the failure path now hands off to the same bounded fallback respawn when leftovers remain.

3. **codex P2 ("Preserve a session captured by a concurrent respawn")**: confirmed real and separate — `retry_after_resume_failure` cleared `inner.session_id` unconditionally at the top of the function, before knowing whether this call would even be the one spawning. If it instead resolved via `DeliverDirect` (a concurrent, unrelated spawn already running with its own freshly-captured session id), that clear would erase it. Scoped the clear to only the `BecomeSpawner` branch.

New/updated tests cover all three. Full suite (1780 tests) + clippy clean. Pushed as 6ec17a805.

a5af:
@codex review

AgentA-asaf:
Addressed both findings from codex on commit 6ec17a805, and both from reagentx on the same commit (posted separately, see prior comment):

1. **codex P1 ("Do not enqueue the stale retry behind its original")**: confirmed real and severe — `retry_after_resume_failure` calls `decide_send_action` with the same content the original (now-dying) spawn already pushed onto the queue. If that spawn's own drain hasn't reached it yet, this pushed a second copy, and the fallback respawn would deliver both — executing the user's prompt twice. `decide_send_action` now takes a `skip_if_already_queued` flag: `false` for `send_message` (a user legitimately re-sending identical text must still see both delivered — never dedup a genuine new message), `true` for `retry_after_resume_failure` (whose payload is a *known* re-delivery, not an independent message).

2. **codex P2 ("Persist the prompt only after accepting its spawn")**: confirmed real — `send_message` was persisting the blockfile line unconditionally, before knowing whether `spawn_process` would even succeed. A rejected first spawn left a "user_message" line for a prompt that was never delivered — reopening the pane reconstructed a phantom message, and a manual retry produced a misleading duplicate. Persistence now happens only once the message is durably in play (unconditional for Queued/DeliverDirect, only on success for BecomeSpawner) — matching this function's pre-queue behavior.

New tests cover the dedup logic directly (genuine-duplicate-text still queues both; known-retry-of-already-queued content doesn't duplicate; a retry whose payload isn't present still queues normally). The persist-ordering fix is a straightforward code-path reorder verified by inspection (not independently unit-testable without a FileStore test harness this file doesn't otherwise use).

Full suite (1784 tests) + clippy clean. Pushed as 954a97f66. (The reagentx findings on this same commit were addressed in the prior comment/commit f70a11951.)

AgentA-asaf:
Correction/addendum to my last comment: I referenced "the prior comment" for reagentx's two findings on commit 6ec17a805, but never actually posted it — doing so now for the record.

**reagentx P1 x2 on commit 6ec17a805** (fixed in commit f70a11951, both in code introduced by the round-3 fallback-respawn fix):

1. **`decide_send_action` gated `DeliverDirect` purely on `stdin_tx.is_some()`**: `spawn_process` sets `stdin_tx` synchronously, well before the queued message that triggered the spawn is actually delivered by the background drain task. A second, genuinely concurrent `send_message` landing in that window took `DeliverDirect` and wrote straight to stdin via `try_send`, racing ahead of the drain's own `Sender::send().await` for the message that actually triggered the spawn — silently reordering user input. Now also requires `!spawning_in_progress` (which stays true for a drain's entire lifetime), routing a late arrival into the queue behind whatever the active drain is still working through instead.

2. **`drain_queue_after_successful_spawn`'s stalled branches cleared `spawning_in_progress` to `false` before `respawn_once_for_leftover_queue` had a chance to re-claim it**: reopened the exact orphaned/duplicate-process window this whole mechanism exists to close. Now only releases the claim when NOT about to hand off to a fallback respawn, mirroring `release_spawn_claim_and_drain_queue`'s own failed-spawn branch (which reagentx confirmed was already correct).

New test added for the `DeliverDirect`-vs-`Queued` gating; the claim-retention fix is a narrow synchronous-code-only race verified by inspection (no `.await` point exists to inject a competing thread deterministically). Both included in the 1781-test suite passing as of f70a11951 (now 1784 as of 954a97f66).

a5af:
@codex review

AgentA-asaf:
reagentx's review of f70a11951 (issue IDs P1 line 882, P2 line 785) described the exact same two findings codex raised on commit 6ec17a805 — already fixed in commit 954a97f66 (dedup + persist-ordering for BecomeSpawner). No further action needed for that review; noting it here for the record since it was filed against an intermediate commit.

**New finding from reagentx on commit 954a97f66 (fixed in 2dec5e2d7)**:

- **P1**: `send_message`'s `DeliverDirect` branch persisted to the blockfile *before* the `tx.try_send` call that can still fail — `stdin_tx` can go `None` (process died) between `decide_send_action`'s check and this later lock re-acquisition, or `try_send` can fail with `Full` under load. A failure returned `Err` to the caller while the blockfile already showed the message as sent — reproducing, for this path, the exact bug the immediately prior commit just fixed for `BecomeSpawner`. Moved the persist call to after a successful send, matching `send_user_message`'s existing correct ordering (which reagentx pointed to as the reference pattern).

Verified by direct code inspection (no FileStore test harness exists in this file to assert persistence timing directly — same as the BecomeSpawner persist-ordering fix in the prior commit). Full suite (1784 tests) + clippy clean. Pushed as 2dec5e2d7.

AgentA-asaf:
Addressed all three findings from codex on commit 954a97f66:

1. **P1 ("Retain every message sent to a stale-resume process")**: confirmed real and severe — `confirmed_stale_resume_retry` only ever carried the ONE message that triggered the spawn. Once other callers could queue additional messages behind that same spawn attempt, the drain could hand SEVERAL of them to the doomed process's stdin channel before it turned out unreachable — channel acceptance is not proof the CLI ever read anything, and only the first was ever redelivered. `pending_resume_retry`/`confirmed_stale_resume_retry` now carry a `Vec<String>`: `spawn_process` still seeds it synchronously with the triggering message (must happen before any background task exists), and the drain appends every later delivery. `retry_after_resume_failure` now loops the whole batch through the same per-message decision logic.

2. **P2 ("Persist queued messages in their delivery order")**: confirmed real — persistence happened wherever each caller's own synchronous code ran, not in delivery order. Moved persistence into the drain itself, at the exact moment each message is actually sent, so blockfile order now matches delivery order.

3. **P1 ("Cancel the pending drain when stopping the process")**: confirmed real — found and fixed the same class of bug: `stop_process`'s kill path cleared the stale-resume retry state but left `pending_send_messages`/`spawning_in_progress` untouched. Since the background drain is a separate task a kill doesn't cancel, it would see `stdin_tx` gone with messages still queued, treat that as a stall, and silently respawn a fresh CLI moments after the user explicitly stopped it. Now clears the queue and releases the claim as part of kill cleanup.

New tests cover the batch redelivery and the drain's append-without-duplicating-the-first logic directly. The kill-path fix is verified by direct code inspection (it's deep inside `spawn_process`'s own async process-waiter `tokio::select!`, which this file's own established precedent avoids unit-testing with a real subprocess).

Full suite (1786 tests) + clippy clean. Pushed as 7087c0690.

a5af:
@codex review

AgentA-asaf:
Addressed 3 of 4 findings from codex on commit 7087c0690; deferred the 4th to a follow-up (#2363), with reasoning below.

1. **P1 ("Keep appending to the promoted retry batch")**: confirmed real — the drain's append-to-retry-batch logic only checked `pending_resume_retry`, but `poison_resume` (running concurrently on the stderr-reader task) can promote it to `confirmed_stale_resume_retry` at any point. A message delivered right after that promotion found `None` where it expected the tentative slot and was silently dropped from the batch. Now checks `confirmed_stale_resume_retry` too.

2. **P2 x2 ("Queue the entire retry batch before spawning" + "Preserve repeated identical messages during retry")**: both confirmed real, same root cause — looping `decide_send_action` once per message in the retry batch left a window between iterations where a genuinely new, unrelated message could interleave into the middle of the batch (reordering it relative to what the doomed process actually received), and the same-content dedup check (meant to guard against a different caller's already-queued copy) would wrongly collapse two genuinely-identical messages within the SAME batch (e.g. the user sent "hello" twice). Replaced the per-message loop with a new `decide_retry_batch_action` that enqueues the whole remaining batch atomically in one lock acquisition; dedup now only ever applies to the first entry against pre-existing queue content, never within the batch.

3. **P2 ("Requeue failed direct retry deliveries")**: confirmed real — the `DeliverDirect` path is the retry's own last-chance delivery attempt (nothing else will ever resend it), but discarded a message on `try_send` failure instead of requeueing it. Fixed.

4. **P2 ("Defer fallback respawn until old-process cleanup finishes")**: confirmed as a genuine theoretical race — the fallback respawn (triggered by the drain's stall-detection) and the original exit-handler's own muxbus/`clear_active_pid` cleanup are independently-scheduled tasks with no ordering guarantee, and `clear_active_pid` isn't PID/generation-aware, so it could wipe out a fresh replacement's registration. **Deferring this one** — closing it properly means adding generation-awareness to `session_recovery`'s shared API (used by other controller types too), a broader change than the rest of this PR's fixes. Assessed likelihood is low in practice: winning this race requires a real OS process spawn (`cmd.spawn()`, several ms) to beat a handful of in-memory/local-store cleanup calls. Filed as #2363 with the fix direction; happy to implement it here instead if you'd rather not merge with it outstanding.

New tests cover all three fixed findings directly (atomic batch enqueue, duplicate-content preservation within a batch, requeue-on-failed-delivery, and the confirmed-retry append race). Full suite (1791 tests) + clippy clean. Pushed as f4442859d.

a5af:
@codex review

AgentA-asaf:
Addressed all findings from commit f4442859d — 1 from reagentx, 2 genuinely new from codex (the rest were stale re-anchors of already-fixed comments from earlier commits):

1. **reagentx P1**: `send_user_message` (the muxbus/jekt steering path) still gated purely on `stdin_tx.is_some()` — the same reordering bug fixed for `send_message`'s `DeliverDirect` path back in round 4. Unlike `send_message` it has no spawn config and its persistence is a live, visible append rather than the drain's silent one, so there's no safe way to queue it behind the drain — it now errors in that narrow window instead of reordering, and the caller can retry.

2. **codex P2 ("Avoid persisting stale-retry messages twice")**: confirmed real — the shared drain persisted every message unconditionally, including a stale-resume retry's own redelivery, which was already correctly persisted on its original attempt. `pending_send_messages` now holds a small `QueuedMessage{json_str, already_persisted}` instead of a bare `String`; the drain skips persisting entries already marked as such.

3. **codex P2 ("Schedule a drain for failed direct retries")**: confirmed real — a failed `DeliverDirect` retry delivery pushed onto the queue without claiming `spawning_in_progress`, so nothing would ever drain it for as long as the same process stayed alive (every future send also takes `DeliverDirect`). Now claims the flag and hands off to the same drain used after a fresh spawn.

New tests cover all three directly. Full suite (1795 tests, one unrelated pre-existing flaky test in `subagent_watcher` confirmed to pass in isolation) + clippy clean. Pushed as e1ae30eea.

a5af:
@codex review

AgentA-asaf:
Addressed reagentx's finding on commit e1ae30eea (codex had no genuinely new findings — its inline comments on this commit are all stale re-anchors of already-fixed items from earlier commits, confirmed by timestamp):

**P1**: confirmed real, and a regression in my own immediately-prior commit — `retry_after_resume_failure`'s `DeliverDirect` branch pushed failed-delivery messages onto the queue under one lock acquisition, then only later re-acquired the lock to set `spawning_in_progress = true`. In that gap, a genuinely concurrent `send_message` could observe `stdin_tx.is_some() && !spawning_in_progress` and take `DeliverDirect` itself — jumping ahead of the just-queued retry messages — or, if the process also died in that window, take `BecomeSpawner` and spawn a duplicate process. Both operations (push + claim) now happen inside the same lock acquisition.

This is a straightforward atomicity fix (two operations merged into one existing lock scope) — verified by inspection alongside the existing `retry_after_resume_failure_claims_the_spawn_flag_when_direct_delivery_fails` test, which already covers the end-state behavior; the fix doesn't introduce new observable behavior to test, just closes the window between two already-tested operations.

Full suite (1795 tests) + clippy clean. Pushed as 7782b2e44.

a5af:
@codex review

AgentA-asaf:
Addressed reagentx's finding on commit 7782b2e44 (codex had no genuinely new findings on that commit):

**P1**: confirmed real — a deeper variant of the round-6 "keep appending to the promoted retry batch" fix. The drain's own "send, then append to the retry batch" sequence has an unavoidable gap at the send's `.await` (a mutex can't be held across it). The process-waiter's exit-handling only waited for the stderr-reader task before deciding the retry batch was final — it never accounted for the drain itself (a separate, independently-scheduled task with no `JoinHandle` kept anywhere). If the drain successfully sends msg2, but the stderr reader (already finished, since the CLI's failure fires almost immediately) promotes the retry batch and the process-waiter takes it *before* the drain gets back around to appending msg2 — msg2 silently vanishes from the retry batch, while still being marked accepted and persisted in the transcript. It's never actually delivered again.

Added `PersistentInner::drain_send_in_flight`, set around the drain's send-then-append sequence; the process-waiter's exit-handling now waits (bounded, 500ms, same bound as the existing stderr-reader wait) for it to clear before deciding the retry batch is final.

New tests cover both the flag lifecycle (set/cleared correctly across a real drain delivery) and the underlying bounded-polling primitive directly (the full cross-task race isn't practical to reproduce deterministically, same reasoning as this file's other cross-task timing fixes).

Full suite (1796 tests) + clippy clean. Pushed as bb782f6cb.

a5af:
@codex review

AgentA-asaf:
Found it — thanks for the nudge; my non-paginated review queries were silently truncating results and missing the reviews that landed on commit bb782f6cb roughly 20 minutes after I pushed it (not stalled after all).

Addressed reagentx's finding (codex had no new findings on that commit):

**P0**: confirmed real — the same resurrection bug the round-6 kill fix closed for force-kill, in a code path that fix didn't touch. Graceful (non-force) `stop_process` clears `stdin_tx` immediately to trigger EOF, then waits up to 5s for the child to exit — but only cleared `pending_send_messages`/`spawning_in_progress` afterward, in the shared cleanup block once that wait resolved. During that window, the drain's independently-scheduled background task could observe `stdin_tx.is_none()` with messages still queued, classify it as a stall, and spawn a brand-new CLI process while the user's graceful stop was still in progress. The force-kill path doesn't have this gap (all three clears already happen together there). Now clears the queue and the claim in the same early lock acquisition as `stdin_tx`, for the graceful path too.

This is deep inside `spawn_process`'s own process-waiter `tokio::select!` (same as the original round-6 kill fix) — verified by inspection, matching this file's established precedent for not unit-testing this exact class of real-subprocess-lifecycle logic.

Full suite (1796 tests) + clippy clean. Pushed as 9a4441862.

a5af:
@codex review

AgentA-asaf:
Addressed codex's finding on commit 9a4441862 (its top-level review body was boilerplate; the finding came as an inline comment — no reagentx review yet at time of this check, will address separately if one lands with anything new):

**P2 ("Prepend retried messages ahead of later leftovers")**: confirmed real — `decide_retry_batch_action`'s `Queued` branch always appended the retry batch to the *back* of the queue. But that batch represents messages the doomed process already accepted *before* whatever's currently queued (which arrived after the original spawn claimed `spawning_in_progress`, so it's chronologically later). Appending let the fresh process receive that later-arriving input before the earlier-accepted retry batch.

Now prepends: if the batch's own first message is still sitting un-popped in the queue (always at index 0 — nothing but `push_back` ever touches this queue, and it's only non-empty because that exact message hasn't been drained yet), the rest of the batch is inserted immediately after it, preserving order. Otherwise the whole batch goes ahead of whatever's already queued. New test covers the exact scenario codex described.

Full suite (1797 tests, one unrelated pre-existing flaky test in `subagent_watcher` confirmed passing in isolation) + clippy clean. Pushed as b06febf80.

a5af:
@codex review

AgentA-asaf:
Addressed reagentx's finding on commit 9a4441862 (I'd missed this review in my prior check — found it via a paginated re-check):

**P0**: confirmed real — the same class of bug already fixed once for `retry_after_resume_failure`, never applied to `respawn_once_for_leftover_queue` (a sibling fallback path added later in this same round). It cleared `config.session_id` to force a resume-less respawn, but `spawn_process`'s own `--resume` decision reads `inner.session_id` directly — `config.session_id` is only consulted to hydrate `inner.session_id` when it's still `None`, which no-ops here since it's now empty. If the doomed process's stderr reader hasn't cleared `inner.session_id` yet (poison_resume racing the drain's own `tx.send()` failure — exactly the fast-fail case this whole mechanism targets), this fallback would reattach `--resume` to the same dead sid and reproduce the identical failure. Since this call passes `resume_retry_payload: None`, nothing re-arms to catch the repeat — the message is silently dropped for good.

Now explicitly clears `inner.session_id`, mirroring `retry_after_resume_failure`'s own existing correct pattern. New test mirrors the existing analogous test for that function.

Full suite (1798 tests) + clippy clean. Pushed as d4cf8ac71.

a5af:
@codex review

AgentA-asaf:
Addressed both findings from reagentx on commit d4cf8ac71:

**P1** (a subtler variant of the round-11 session_id fix): confirmed real, on careful re-verification. `respawn_once_for_leftover_queue` is reached from the drain's own stall detection with no guarantee `poison_resume` has already run for the doomed process's sid — its stderr-reader task can get aborted past its 500ms bound before ever calling `poison_resume` (this path, unlike `retry_after_resume_failure`, doesn't require `confirmed_stale_resume_retry` to have been set, so it fires independently of that promotion). In that case the same process's stdout-reader task — which has no bound at all — could still be racing to echo-capture that exact sid, re-setting `inner.session_id` right after a plain clear. Now calls `poison_resume` on whatever sid was held instead of just clearing it, so a later capture attempt for that exact sid is refused regardless of ordering.

**P2** (content-based identity ambiguity in the round-11 prepend fix): confirmed real, but deferred to #2365. `decide_retry_batch_action` checks "is the batch's own trigger still queued" via content equality — if a genuinely unrelated, later message happens to have identical text (plausible for short replies like "yes"/"ok"), it's misidentified, causing the retry batch to be misordered. I verified neither position-based nor the existing `already_persisted`-flag-based disambiguation works here (both a genuine still-queued trigger and a coincidentally-matching new message start out `already_persisted: false`) — a proper fix needs real per-message identity (sequence numbers) threaded through the queue AND the retry-batch `Vec<String>` representation, a larger change than proportionate to add here. Severity is lower than this PR's other fixes (reordering, not data loss), and requires an actual content collision. Filed with the fix direction in #2365.

New test covers the P1 fix directly. Full suite (1799 tests) + clippy clean. Pushed as fc0ece635.

a5af:
@codex review

AgentA-asaf:
Reverted the round-12 fix in e9678091f — the unconditional \`poison_resume\` call in \`respawn_once_for_leftover_queue\` was itself a regression, and this round's review correctly caught it.

**Why it was wrong:** this fallback fires from two triggers that have nothing to do with a *confirmed* stale `--resume` — `release_spawn_claim_and_drain_queue`'s `!spawn_succeeded` branch (any `spawn_process` OS-level failure: missing binary, OS error) and `drain_queue_after_successful_spawn`'s stall branch (any process crash with a leftover queue, not specifically a stale-resume death). In both, `inner.session_id` could just as easily be a legitimately hydrated-but-unattempted id, or a genuinely valid, already-captured session from a process that ran fine and crashed for an unrelated reason. `poison_resume` sets `resume_poisoned`, documented as permanent (never reset back to `None`) — poisoning a sid never actually confirmed dead by the CLI would silently and permanently break that session's resume capability, with no disclosure, for the rest of the controller instance's lifetime. That's worse than the narrower race it was meant to close.

**Fix:** reverted to the original plain `session_id = None` clear (round 11's behavior). Swapped the test from asserting poisoning happened to asserting a valid sid survives this fallback uncorrupted and remains capturable afterward (`respawn_once_for_leftover_queue_does_not_poison_the_sid_it_held`).

**The narrower round-12 race is not forgotten** — filed as #2366 for proper tracking: a still-racing stdout-reader task (unbounded, unlike the stderr-reader's 500ms wait) could theoretically re-capture the exact dead sid right after this fallback's plain clear, causing one respawn to still reattach a dead `--resume`. Closing it properly needs the dying spawn's specific `attempted_resume_sid` threaded through to this call site so it can compare against the *specific* sid rather than poisoning based on ambient state — a real fix, but a larger structural change than proportionate to fold into this PR's remaining scope, and not something to rush after just causing one over-broad regression by trying to close it too quickly.

Full suite: 1799 passed, clippy clean. Requesting a fresh look.

a5af:
@codex review

AgentA-asaf:
Fixed in 8c2bc99ab — codex's P2 on \`e9678091f\`.

**Confirmed the finding is real**: \`release_spawn_claim_and_drain_queue\`'s failed-spawn path checked the queue for leftovers under one lock acquisition, then cleared \`spawning_in_progress\` in a *separate*, later lock acquisition. A concurrent \`send_message\` landing in that gap would see \`spawning_in_progress\` still \`true\`, take \`decide_send_action\`'s \`Queued\` branch, and get told "accepted" via \`emit_message_accepted\` — then this function's second lock would clear the claim without ever rechecking the queue, leaving that accepted message stranded with no drain and no respawn ever responsible for it (exactly as described: "remains undelivered until an unrelated later send happens to spawn another process").

**Fix**: merged the emptiness check and the flag clear into one lock acquisition — same shape as the round-9 atomicity fix already applied elsewhere in this file. Only clears the claim if the queue is *still* empty at that exact instant; otherwise leaves it held and hands off to the existing fallback respawn (\`respawn_once_for_leftover_queue\`), matching how the sibling "successful spawn stalls with leftovers" case already behaves.

**Verification**: added a real-thread concurrency test, \`release_spawn_claim_and_drain_queue_never_silently_strands_a_racing_send\` — 30 iterations × 8 racing threads calling \`decide_send_action\` concurrently with the release path, asserting the invariant that a stranded message (claim released, queue non-empty) can only happen alongside a disclosed respawn-attempt status publish, never silently. Traced through all reachable interleavings by hand to confirm the fix closes this structurally (the racer's push can now only land fully before or fully after the atomic block, never inside it) rather than just probabilistically.

Full suite: 1800 passed (up from 1799 — one net new test), clippy clean. Requesting a fresh look.

a5af:
@codex review

AgentA-asaf:
Round 15, commit fdb8db6fd — addressed both new inline findings from the last round (\`8c2bc99ab\`).

**Fixed — codex P2 at \`persistent.rs:651\` ("Keep the spawner message identifiable behind leftovers")**

Confirmed real: \`release_spawn_claim_and_drain_queue\`'s failed-spawn path assumed the caller's own message is always at the front of \`pending_send_messages\`. That assumption breaks specifically after a "second stall" (\`drain_queue_after_successful_spawn\` with \`allow_fallback_respawn: false\`), which deliberately releases \`spawning_in_progress\` while leaving genuine leftover messages queued. A later \`send_message\` claiming \`BecomeSpawner\` appends its own message *behind* those leftovers — so when its spawn fails, blindly popping the front discarded an older, unrelated, already-accepted prompt instead of the message that actually failed (real data loss), and handed the wrong, already-failed message to the fallback respawn.

**Fix**: threaded the exact message each \`BecomeSpawner\` claim pushed through to the failed-spawn path (\`own_message: &str\`), which now removes that specific entry by content match instead of assuming position. This fully resolves the common case (differing content). The narrow residual — two genuinely different messages sharing identical content — is the same content-identity-ambiguity gap already tracked in #2365; either duplicate is content-equivalent to discard here, so this fix doesn't need to close that separately.

Added \`release_spawn_claim_and_drain_queue_discards_the_right_entry_when_it_is_not_at_the_front\`, reproducing the leftover-then-append scenario directly.

**Deferred, filed as #2367 — codex P2 at \`persistent.rs:1268\` ("Claim stale retries before accepting later sends")**

Also confirmed real: a confirmed stale-resume retry's \`decide_retry_batch_action\` can resolve via \`DeliverDirect\` into an entirely unrelated, later-spawned process if that process's own spawn+drain cycle completes during the exit-handler's synchronous cleanup window (between clearing \`stdin_tx\`/taking \`confirmed_stale_resume_retry\` and actually calling \`retry_after_resume_failure\`) — reversing accepted input order.

Not fixed inline: the obvious first idea (re-assert \`spawning_in_progress = true\` as soon as the retry is known) doesn't work — \`decide_retry_batch_action\` needs to see it as \`false\` to know IT should become the spawner; pre-setting it would starve the retry forever instead (queued behind a spawn that never happens). A correct fix needs the two decisions ("retry still pending for this exited process" vs "a genuinely new spawn is allowed") to be distinguishable, which the current single bool can't express — likely a spawn-generation counter or a more granular claim state. Reasoning and full race trace in #2367. Same general race shape as the already-deferred #2363 (an OS-level spawn winning against a few synchronous cleanup calls), different specific defect (reordering vs PID-generation), so filed separately rather than folded in.

Full suite: 1801 passed (up from 1800), clippy clean. Requesting a fresh look.

a5af:
@codex review

AgentA-asaf:
Round 16, commit ce1642d90 — reagentx APPROVED \`fdb8db6fd\` and codex left two more genuinely new inline P2 findings on that same commit. Confirmed both real; both fixed.

**Fixed — codex P2 at \`persistent.rs:818\` ("Track the seeded retry message instead of the queue front")**

The drain's \`is_first_delivery\` flag identified \`spawn_process\`'s synchronously-seeded retry message by POSITION (whatever this drain pops first) — the same assumption I'd already fixed for \`release_spawn_claim_and_drain_queue\` two commits ago, just resurfacing in a sibling location. A prior "second stall" can leave an older leftover queued ahead of a later spawner's own triggering message, so the first thing this drain pops isn't necessarily the seeded one. Treating it as if it were: silently dropped the older leftover from retry-batch tracking entirely (permanent loss if this process also later dies from a stale resume), while recording the ACTUAL trigger message a second time (duplicate execution on retry).

**Fix**: match the seeded entry by content (\`is_the_seed\`, matched at most once via \`seed_already_matched\`) instead of position — same approach as the earlier fix, same known content-identity-ambiguity limits as #2365 for the narrow case of genuinely duplicate content. Added \`drain_identifies_the_seed_by_content_even_when_a_leftover_is_delivered_first\`.

**Fixed — codex P2 at \`persistent.rs:1169\` ("Queue the rest of the retry batch after the first full send")**

\`retry_after_resume_failure\`'s \`DeliverDirect\` branch kept attempting \`try_send\` for every message in a multi-message batch even after an earlier one failed. Since the stdin receiver runs concurrently, it can free capacity between loop iterations, letting a LATER message succeed via \`try_send\` while an EARLIER, failed one sits queued behind it — reordering the batch relative to itself once the drain eventually redelivers the queued one.

**Fix**: a one-line short-circuit (\`!any_failed && ...\`) — once any message fails, every remaining one is queued in order instead of re-attempting direct delivery. Added \`retry_after_resume_failure_queues_the_rest_of_the_batch_in_order_once_one_fails\` covering the deterministic all-fail case; the actual race (an earlier failure, a later success) needs genuine concurrent capacity freed mid-loop with no yield point to hook a test into, so it's verified by direct code inspection instead — same established practice this file already uses for cross-task async races that can't be reproduced deterministically without flakiness.

Full suite: 1803 passed (up from 1801), clippy clean. Requesting a fresh look.

a5af:
@codex review

AgentA-asaf:
Round 17, commits e4fc20451 + 517d42eac — addressed both findings from the last round (\`ce1642d90\`): reagentx's COMMENTED review (a P2) and codex's inline P1.

**Fixed — codex P1 at \`persistent.rs:2244\` ("Cancel a stale-session retry when Stop arrives")**

Confirmed real, and more severe than most findings this PR has seen — a broken Stop button. If Stop/Esc arrives after a process has already exited from a confirmed stale \`--resume\` but before the exit-handler gets around to invoking the retry, \`stop_process\` finds nothing to signal (kill_tx already cleared) and returns success recording no cancellation anywhere — worse, even when \`kill_tx\` is still \`Some\` at that moment, sending through it can be equally futile, since \`tokio::select!\` may have already committed to its exit branch before \`stop_process\`'s send ever lands, so the \`kill_rx\` arm is never polled again regardless. Either way, the exit-handler proceeds to spawn a fresh CLI and deliver the same prompt — directly overriding an explicit stop.

**Fix**: added a monotonic \`spawn_generation\` counter (bumped once per \`spawn_process\` call, atomically with stashing \`pending_resume_retry\`) and a \`stop_requested_generation\` field that \`stop_process\` now *always* records — not just when \`kill_tx\` is already gone. The exit-handler's retry decision consumes that field against its own captured generation before deciding whether to retry, honoring the stop instead. Monotonic, never-reused generation numbers mean a stale unconsumed value can never spuriously match a later, unrelated spawn, so no separate leak-prevention was needed. Extracted the check into a \`PersistentInner\` method (\`consume_stop_requested_for_generation\`, mirroring \`poison_resume\`/\`try_capture_session_id\`) so it's directly unit-testable without a real child process exit. Added three tests covering the match/mismatch/no-live-kill_tx cases.

Note: this is a **different** mechanism from the \`spawn_epoch\`/\`should_skip_own_delivery\` counter this PR removed earlier (see \`spawning_in_progress\`'s own doc comment) — that one deduped a spawn-claim race, a job \`spawning_in_progress\` now fully owns. This is a narrower, independent signal solely for stop-cancellation.

**Fixed — reagentx P2 at \`persistent.rs:638\` (stale doc comment)**

Confirmed: \`decide_send_action\`'s doc comment claimed \`skip_if_already_queued=true\` is used by \`retry_after_resume_failure\`, but that function was refactored in round 6 to call \`decide_retry_batch_action\` instead — a separate function with its own dedup logic. \`decide_send_action(..., true)\` is only exercised by this file's own unit tests today, never a production call site. Corrected the comment to say so and point at the function that actually implements the production behavior now, without touching the (still test-covered, still-documented-for-reasoning) dead branch itself.

Full suite: 1806 passed (up from 1803), clippy clean. Requesting a fresh look.

a5af:
@codex review

PR #2361, opened 2026-07-29
fix(agent): backfill cross-channel agent definitions before instance create

AgentY-asaf:
## Summary

Launching an old agent whose definition exists only in the global cross-channel registry (created/last touched under a different channel or version than the one currently connected) fails with two SQLite `FOREIGN KEY` errors — `createagentinstance` and `linkagentidentity` — and the agent's pane/container opens but never processes a message, since the underlying `db_agent_instances` row is never actually created.

Root-caused live against a real production agent ("Moras", a container agent, `created_by_version: 0.46.5`). The identical symptom was independently diagnosed for two host agents ("Oozp"/"Parko") two days earlier in `docs/specs/REPORT_AGENT_DEFINITION_DB_GAP_2026_07_27.md`, which left open (§5) exactly how a definition-less agent could still appear launchable at all — this investigation answers it: agent definitions live in two places (local per-channel SQLite, and a separate cross-channel/global JSON registry the picker reads from), and nothing backfills the former from the latter before `instance_create`'s raw INSERT.

**Backend** (`agentmux-srv/src/backend/storage/agents.rs`): `instance_create` now backfills a local `db_agent_definitions`/`db_agents` shadow row (plus a best-effort local content/skills copy) from the registry when the local row is absent, before its own INSERT. Deliberately does **not** reuse `agent_def_insert`'s registry-mirror step for this backfill — confirmed by reading `registry_def_upsert` directly that doing so would read back only the still-empty just-inserted local content/skills and overwrite the registry's real content/skills with empty arrays. If the registry also has nothing, the original FK error still surfaces unchanged — a genuinely orphaned `definition_id` isn't silently swallowed.

**Frontend**: two call sites (`AgentPicker.tsx`, `useContinueOrNewMode.ts`) duplicated an incomplete blacklist for carrying over identity/memory ids on a continuation launch (`!== "blank"` only) — `"default"`, a legacy pre-real-UUID-accounts sentinel, passed straight through and triggered `linkagentidentity`'s `account_id -> db_accounts` FK failure. Replaced with a shared UUID-shape check (`identity-carry-over.ts`) — real ids are always UUIDs — so any legacy sentinel is filtered, not just the ones already observed, matching what the existing code comment already said was intended.

## Test plan

- [x] `cargo check -p agentmux-srv` / `--tests` — clean
- [x] 3 new backend tests: `instance_create` succeeds when the local definition is registry-only; still errors when missing everywhere; preserves the registry's real `updated_at`. Critically, one test asserts the registry's content/skills survive the backfill (the regression guard for the corruption bug found while designing this fix).
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` (full suite) — 1768/1768, run twice
- [x] 8 new frontend tests (`identity-carry-over.test.ts`) + `npx vitest run frontend/app/view/agent` (full agent-view suite) — 889/889, no regressions
- [ ] Manual: relaunch Moras against the live instance and confirm the FK errors are gone and it actually responds — not yet done as part of this PR; will verify post-merge against the live production instance this bug was found on.

### DIFF
--- a/.changesets/1785341697-fix-agent-backfill-cross-channel-agent-definitions-before-in-mn26.md
+++ b/.changesets/1785341697-fix-agent-backfill-cross-channel-agent-definitions-before-in-mn26.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): backfill cross-channel agent definitions before instance create, filter legacy identity sentinels
--- a/agentmux-srv/src/backend/storage/agents.rs
+++ b/agentmux-srv/src/backend/storage/agents.rs
@@ -324,26 +324,7 @@ impl Store {
     /// "does the promote-target clone for this template already
     /// exist?" and either reuses it or inserts it.
     pub fn agent_def_get(&self, id: &str) -> Result<Option<AgentDefinition>, StoreError> {
-        // Try local SQLite first (conn released before global registry check).
-        let local = {
-            let conn = self.conn.lock().unwrap();
-            let mut stmt = conn.prepare(
-                "SELECT id, slug, name, icon, provider, description,
-                        working_directory, shell, provider_flags, auto_start,
-                        restart_on_crash, idle_timeout_minutes, created_at,
-                        agent_type, environment, agent_bus_id, is_seeded,
-                        accounts, parent_id, branch_label, updated_at,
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2364, opened 2026-07-30
fix(ci): Windows nightly build failed to extract codec CEF (backslash zip paths)

Agent2-asaf:
## Summary
- The Windows job of `ci-nightly-artifacts.yml` (and `build-windows.yml`) has failed outright on every run since at least 2026-07-28 whenever the CEF-runtime cache misses -- no nightly Windows artifact, codec or otherwise, has actually been produced in that window.
- Root cause: the codec-enabled CEF release zip is created with PowerShell's `Compress-Archive` (`docs/cef-build/build-patched-cef-windows.md` §7), which stores nested directory entries with backslash path separators. Info-ZIP's `unzip` (bash) can't parse those as nested directories, so the `libcef.dll` lookup always came up empty.
- Fixed by switching both workflows' extraction to PowerShell's `Expand-Archive` -- same .NET zip implementation that created the archive, so it round-trips correctly regardless of internal path separator style. No change needed to the zip-creation side, no need to rebuild/re-release the already-published CEF binary.
- Live-verifying that fix surfaced a second, distinct bug: `Taskfile.yml`'s `bundle:windows` task failed one step later on a missing optional file (`vulkan-1.dll`, absent from this particular codec CEF build) with a raw `GetFileAttributesEx` task-engine error -- Task's shell (`mvdan.cc/sh`) runs `cp` as a native Go builtin, and on Windows a missing *source* file bypasses the `2>/dev/null || true` suppression the script relied on. Fixed by switching every optional-file copy in that block to an explicit `[ -f src ] && cp` existence check (the pattern the locale-file copy already used successfully).
- Full root-cause writeup, live-verification evidence, and a note on scope (only `bundle:windows` touched -- `bundle:linux`'s lookalike lines are unaffected, confirmed by that job completing successfully in the same runs) in the included retro.

## Verification
- Triggered two live `workflow_dispatch` runs of `ci-nightly-artifacts.yml` against this branch:
  - First run (extraction fix only): CEF download/extract step succeeded, version-match confirmed (`✓ CEF runtime OK: libcef.dll ... matches linked cef crate major 148`); failed one step later on the `vulkan-1.dll` bug.
  - Second run (both fixes): **all three platform jobs succeeded**, including Windows.
- `python3 -c "import yaml; yaml.safe_load(...)"` on both workflow files and `Taskfile.yml` after edits.

## Test plan
- [x] Live CI verification (see above) -- this is a CI-infrastructure fix, so the actual test *is* a green CI run, already done.
- [ ] Next scheduled nightly run (06:00 UTC) should now produce a Windows artifact for the first time since 2026-07-27.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -139,24 +139,37 @@ jobs:
           path: ${{ env.CEF_RUNTIME_DIR_WINDOWS }}
           key: cef-runtime-windows-${{ steps.cef.outputs.tag }}
 
+      # `Expand-Archive`, not bash `unzip`: the release zip is produced by
+      # PowerShell's `Compress-Archive` (docs/cef-build/build-patched-cef-windows.md
+      # §7), which stores internal entries with backslash path separators.
+      # Info-ZIP's `unzip` (the bash job's prior tool here) can't parse those
+      # as nested directories — it silently flattens/mangles them, so the
+      # libcef.dll lookup below always failed with "not found in zip" on any
+      # cache miss. Extracting with the same .NET zip implementation that
+      # created the archive round-trips correctly. See
+      # docs/retro/retro-nightly-windows-cef-zip-backslash-paths-2026-07-29.md.
       - name: Download codec-enabled CEF runtime
         if: steps.cef-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2369, opened 2026-07-30
fix(agent-stream): require real explanation text before ending a persistent-mode turn

AgentX-asaf:
## Summary

In persistent/interactive mode, `handleAssistantMessage()` synthesized `session_end` (turn "Done") whenever an assistant message merely lacked a `tool_use` block — including thinking-only or entirely empty messages. That let the UI settle to "Worked" a beat before the real explanation streamed in, then immediately reopen, firing the "Picked up more work — starting another round…" notification.

Fix: only emit `session_end` when the message has **no** `tool_use` block **and** contains real, non-empty explanation text. Every genuinely finished turn (`stop_reason: end_turn`) carries at least some user-facing text by definition, so this only excludes the anomalous/transitional messages that were causing the false-positive settle. The existing 180s liveness-recovery watchdog remains as a backstop against the original "stuck forever" bug this heuristic was added to prevent (PR #1757), so this can't regress that.

Two existing tests asserted the old (buggy) behavior literally — reversed them and added coverage for whitespace-only text and tool_use+text in the same message.

Full design writeup: `docs/specs/SPEC_PERSISTENT_TURN_END_TEXT_GATE_2026_07_30.md`.

## Test plan

- [x] `claude-translator.test.ts` — 27/27 passing (2 reversed, 3 new)
- [x] Full agent-pane suite (`frontend/app/view/agent`, `frontend/app/store/agent-pane-state`) — 1003/1004 passing (1 unrelated pre-existing timeout flake in `tool-renderers/registry.test.ts`, confirmed passes in isolation)
- [x] `tsc --noEmit` clean
- [ ] Manual `task dev` verification: multi-tool-call turn no longer shows a premature "Worked" checkmark between rounds

### DIFF
--- a/.changesets/1785419249-fix-agent-stream-require-real-explanation-text-before-ending-od56.md
+++ b/.changesets/1785419249-fix-agent-stream-require-real-explanation-text-before-ending-od56.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-stream): require real explanation text before ending a persistent-mode turn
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2370, opened 2026-07-30
fix(agent-pane): pin scroll from content ResizeObserver, not an itemized signal whitelist

Agent3-asaf:
## Summary

The message-list scrollbar was visibly rising as content streamed in, then snapping back down once some unrelated signal happened to re-fire the pin effect — not the smooth, unconditional pin best-practice chat UIs achieve.

**Root cause:** the pin-to-bottom effect only re-scrolled when one of three hand-picked signals changed (`nodes().length`, `layoutView().totalSize`, `workingRowHeight`). Any other source of content growth left the scrollbar visibly above true bottom until an unrelated signal happened to fire. A confirmed culprit: `MarkdownBlock`'s trailing syntax-highlight re-render lands via a bare `setTimeout` ~90ms after the last token — completely outside that signal graph — reflowing code blocks to a different height with no re-pin.

**Fix:** add a `ResizeObserver` on the virtualized region + streaming buffer — the elements whose box size *is* the content height — and re-pin from any resize while `stickToBottom()` is true. `ResizeObserver`'s notification step runs after layout but before paint, so the `scrollTop` write never produces a visible flash regardless of what caused the growth. This mirrors the pattern used by production chat UIs (e.g. `use-stick-to-bottom`, Stream Chat's React SDK) — see the audit report for full research.

## Audit + research

`docs/specs/REPORT_AGENT_PANE_SCROLL_PIN_FLICKER_AUDIT_2026_07_30.md` has the full code audit and external research this is based on.

## Test plan

- [x] `npx tsc --noEmit` — clean (no new errors from this change)
- [x] `npx vitest run app/view/agent` — 69 files / 889 tests passed, 0 regressions
- [x] Live-verified via `task dev` + CDP: streamed real long responses (code blocks in Python/Rust/Go) through fresh, isolated agent identities and sampled scroll gap (`scrollHeight - clientHeight - scrollTop`) at 100ms intervals for the full ~45s duration of each turn. Result: 0/443 and 0/443 samples exceeded a 5px gap (max transient 0.6px and 42.8px respectively), vs. `main`'s baseline of 263/443 samples over a 5px gap and a ~21px resting gap on the same test.

### DIFF
--- a/.changesets/1785398290-fix-agent-pane-pin-scroll-from-content-resizeobserver-not-an-98kb.md
+++ b/.changesets/1785398290-fix-agent-pane-pin-scroll-from-content-resizeobserver-not-an-98kb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): pin scroll from content ResizeObserver, not an itemized signal whitelist
--- a/docs/specs/REPORT_AGENT_PANE_SCROLL_PIN_FLICKER_AUDIT_2026_07_30.md
+++ b/docs/specs/REPORT_AGENT_PANE_SCROLL_PIN_FLICKER_AUDIT_2026_07_30.md
@@ -1,0 +1,113 @@
+# Report: why the message-list scrollbar still visibly drifts up before snapping to bottom, and how to make it pin unconditionally
+
+**Date:** 2026-07-30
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2372, opened 2026-07-30
fix(agent-composer): two-line responsive layout for narrow panes

AgentX-asaf:
## Summary

The composer strip's narrow-pane container queries targeted `@container modal-mount` — a container name only declared by `<Modal>` portal mounts, never by the docked agent pane (whose real containment context is `.agent-view`'s `container-name: agent-pane`). They never matched, so narrowing the pane produced **no responsive behavior at all** — content just crowded/overlapped.

Fix, keyed off the correct `agent-pane` container:
- **≤480px** — switch from a single-row 3-column grid to a 2-row grid via `grid-template-areas`: controls + right-zone (badges/ctx/auth/Shell) share row 1, token/elapsed stats move to their own full-width, centered row 2. Nothing is hidden at this tier — giving stats a dedicated row removes them from the row-1 space contest entirely.
- **≤300px / ≤260px** — shed the auth tag, then the process badge (informational, lowest priority). Context text and Shell stay visible at every tier (ctx = the one safety signal in the bar, Shell = the primary action).
- **≤220px** — runtime-dropup trigger and Shell button compact, matching the pre-existing (previously dead) intent.

Full design: `docs/specs/SPEC_COMPOSER_STRIP_TWO_LINE_RESPONSIVE_2026_07_30.md`.

## Test plan

- [x] `npx sass --style=compressed` on the changed file — compiles clean, no syntax errors
- [ ] `task dev` — manually resize an agent pane through the tiers and confirm: stats move to row 2 with no clipping around 480px; auth tag disappears first, then process badge; context text + Shell remain usable at the narrowest widths tested

### DIFF
--- a/.changesets/1785433477-fix-agent-composer-two-line-responsive-layout-for-narrow-pan-i73s.md
+++ b/.changesets/1785433477-fix-agent-composer-two-line-responsive-layout-for-narrow-pan-i73s.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-composer): two-line responsive layout for narrow panes
--- a/docs/specs/SPEC_COMPOSER_STRIP_TWO_LINE_RESPONSIVE_2026_07_30.md
+++ b/docs/specs/SPEC_COMPOSER_STRIP_TWO_LINE_RESPONSIVE_2026_07_30.md
@@ -1,0 +1,271 @@
+# SPEC — Composer strip: two-line wrap when the pane narrows
+
+**Date:** 2026-07-30
+**Type:** UI responsiveness spec
+**Status:** Proposed — audited and designed, not yet implemented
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentX-asaf:
Addressed both review findings in fd87687e:

- **reagentx P2** (stale ≤320px comment reference) — updated to reference the actual ≤220px tier.
- **codex P2** (row overflow at ≤220px) — confirmed: `1fr` in CSS Grid keeps an implicit content-based minimum, so the controls column couldn't actually shrink past the trigger+badge's intrinsic width and the row could overflow instead of clipping. Fixed with `minmax(0, 1fr)` on the track plus `min-width: 0` on `.agent-composer-strip-controls` (the classic nested-flex-in-grid min-width trap).

PR #2376, opened 2026-07-31
feat(process-tracker): register ShellController/AcpController spawns (Process Broker Phase B)

AgentA-asaf:
## Summary

Call-site parity + refactor for `process_tracker::registry`, the source of a `ProcessStatus`'s OS-process detail in `agentmux-srv/src/broker/process.rs`. Previously only `SubprocessController` and `PersistentSubprocessController` registered spawned children; `ShellController` (shell/cmd panes) and `AcpController` never did, so the broker's `processes` field was always empty for those two controller types.

- Factored the duplicated `ensure_tracker`/`assign_process` block into one `process_tracker::registry::track_spawned(block_id, pid)` helper (was copy-pasted verbatim at the two existing call sites).
- Wired `ShellController` (`shell/lifecycle.rs`) and `AcpController` (`acp.rs`) to call it with the PID each already extracts today.
- Added test coverage for the underlying mechanism, which had none before.

## Design doc

`docs/specs/SPEC_PROCESS_BROKER_PHASE_B_SHELL_ACP_REGISTRATION_2026_07_31.md` — full scoping, including the Windows Job Object risk this PR surfaced (see below).

## Re-scoped after the requested live check (see #2405)

The live check requested below happened, and found the risk was real, not a sandboxed-harness artifact. #2405 confirms `AssignProcessToJobObject` fails Access Denied on every controller type, including `SubprocessController`/`PersistentSubprocessController`, both live on this branch under `task dev` and in the shipped v0.54.9 production build (3/3 real spawns failed in production logs). This directly contradicts this PR's original working assumption that the two existing controllers "already exercise this path successfully in production."

What that means for this PR specifically: the `track_spawned` refactor and the two new call sites are correctly wired (#2405 verified this live), but they register into a currently-dead path (`ProcessStatus.processes` stays empty for every controller type, not just Shell/ACP, until #2405's underlying fix lands). This PR delivers call-site parity plus the dedup refactor only, not a functional fix. The real capability fix is tracked in #2405, which also evaluates two fix directions (nested-Job-Object configuration vs. pivoting the broker's enrichment to PID-tree enumeration).

Merging this now is still worth it despite the dead path: once #2405 lands a fix in `track_spawned`, all four controllers benefit immediately with no further call-site work. Holding this PR until #2405 resolves would just mean redoing this exact parity wiring later.

The ignored test (`ensure_tracker_and_assign_process_track_a_real_short_lived_child`) stays ignored - the underlying capability genuinely doesn't work yet, so a real-spawn assertion would be a legitimate, currently-expected failure, not a false one.

## Test plan

- [x] `cargo test -p agentmux-srv` - 1854 passed, 0 failed, 5 ignored (incl. the newly-added, deliberately-ignored test above)
- [x] Live `task dev` check (via #2405): call-site wiring confirmed correct; `AssignProcessToJobObject` itself confirmed broken for all controller types, tracked separately

### DIFF
--- a/.changesets/1785512603-feat-process-tracker-register-shellcontroller-acpcontroller--4sqs.md
+++ b/.changesets/1785512603-feat-process-tracker-register-shellcontroller-acpcontroller--4sqs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(process-tracker): register ShellController/AcpController spawns with process_tracker::registry (Process Broker Phase B, shell+acp coverage)
--- a/agentmux-srv/src/backend/blockcontroller/acp.rs
+++ b/agentmux-srv/src/backend/blockcontroller/acp.rs
@@ -237,6 +237,14 @@ impl AcpController {
             "ACP agent process spawned"
         );
 
+        // Assign to this block's process tracker — same path SubprocessController
+        // and PersistentSubprocessController already use. Closes the Process
+        // Broker Phase B coverage gap for ACP-type agent panes (see
+        // docs/specs/SPEC_PROCESS_BROKER_PHASE_B_SHELL_ACP_REGISTRATION_2026_07_31.md).
+        if pid != 0 {
+            crate::backend::process_tracker::registry::track_spawned(&self.block_id, pid);
+        }
+
         let (kill_tx, kill_rx) = tokio::sync::oneshot::channel::<bool>();
         let stdin = child.stdin.take()
             .ok_or_else(|| format!("[acp] stdin not captured for block {}", self.block_id))?;
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -1958,17 +1958,7 @@ impl PersistentSubprocessController {
         // Matches `SubprocessController`'s identical path — both controller
         // types share the same swarm-pane visibility story.
         if pid != 0 {
-            if let Some(registry) = crate::backend::process_tracker::registry::global() {
-                let tracker = registry.ensure_tracker(&self.block_id);
-                if let Err(e) = tracker.assign_process(pid) {
-                    tracing::warn!(
-                        block_id = %self.block_id,
-                        pid = pid,
-                        err = %e,
-                        "[process-tracker] assign_process failed"
-                    );
-                }
-            }
+            crate::backend::process_tracker::registry::track_spawned(&self.block_id, pid);
         }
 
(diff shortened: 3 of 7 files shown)


a5af:
@codex review

a5af:
@codex review

GenericAgentX-asaf:
## Live check performed — result: ❌ fails, including in production

I ran the live verification this PR requested (shell pane + ACP pane under `task dev` on this branch, checked via `agent.process-list` / `broker` process data). Full evidence in **#2405**; summary:

**The new wiring itself is correct** — logs confirm `track_spawned` fires for both `ShellController` and `AcpController` spawns (`created Windows Job Object` → `registered tracker`), so the coverage-gap change does what it says.

**But `assign_process` failed with `AssignProcessToJobObject … Access is denied (os error 5)` on all four reproductions**, including a `task dev` launched from a **plain `conhost` cmd window with no agent/harness tooling anywhere in the parent chain** — ruling out the agent-shell job-nesting confound this PR's ⚠️ section leaned on. `processes` stayed `[]` for both pane types every time.

**More importantly, the PR's key assumption doesn't hold:**

> `SubprocessController`/`PersistentSubprocessController` already exercise the identical `AssignProcessToJobObject` path successfully in production

The shipped v0.54.9 portable on this machine logged **3/3 `assign_process failed` (same Access Denied) for `PersistentSubprocessController` spawns today**, and shared srv logs from v0.54.4 → v0.54.9 show the same warning on every active day with no evidence of a single success. Phase A's Job-Object registration appears to have been non-functional on Windows since it shipped — the `#[ignore]`d test was reproducing a real product defect, not a harness artifact. Likely root cause (launcher J0 job vs. tracker job nesting) and fix directions are scoped in #2405; notably ACP fails too despite spawning via plain `Command` (no ConPTY), so it isn't PTY-specific.

**Suggested disposition:**
- Don't check the two live-check boxes — they fail as written.
- Either hold the merge, or re-scope the PR description to "call-site parity + `track_spawned` refactor" (the refactor and tests are good, and the new call sites are harmless warn-and-continue) with the actual capability tracked in #2405. Merging as-is without that re-scope would mark the Phase B coverage gap "closed" while the underlying capability has never worked.

*Side finding, unrelated to this PR (will file separately if wanted): `agent_open.rs:129` resolves provider CLIs at `~/.agentmux/<version>/cli/<provider>` while `install_handlers.rs::provider_install_dir` installs to `~/.agentmux/instances/v<version>/cli/<provider>` — so `agent.open` reports `CLI_NOT_AVAILABLE` even immediately after a successful `install.start`.*

PR #2378, opened 2026-07-31
feat(agent): detect and surface context compaction in real time

AgentA-asaf:
## Summary

AgentMux drives Claude Code as `claude --print --output-format=stream-json` and translates its frames into an internal `AgentEvent`. Claude Code already emits an authoritative `system`/`compact_boundary` frame with exact trigger/token/duration data whenever compaction completes — but the translator silently discards every `system`-type frame, so AgentMux fell back to *inferring* compaction from a ≥50% token-count drop on the next turn, with **no signal at all while it's happening**.

This surfaced concretely: figuring out when a long-running agent session had gone through compaction required cross-referencing external git commit timestamps, because AgentMux itself couldn't answer "when did this start." Full writeup: `docs/specs/SPEC_COMPACTION_DETECTION_AND_HANDLING_2026_07_31.md`.

This PR ships **Tier 1** (live "compaction started" signal + pane status) and **Tier 2** (elapsed-time counter, finalized with the real backend duration on completion), plus the backend fix both depend on. Tier 3 (predictive countdown) and Tier 4 (estimated progress bar) are explicit follow-ups, not built here — see spec §7.

## What changed

- **`agentmux-srv`**: stop discarding `compact_boundary` frames — forward a real `AgentEvent::CompactionBoundary` (trigger, pre/post tokens, cumulative dropped, duration) instead of throwing the data away.
- **`agentmux-bashwrap`**: new `precompact` subcommand, registered as Claude Code's `PreCompact` hook, pings the sidecar over the existing WPS HTTP transport (same one `exec` already uses for `tool_chunk`) the instant compaction *starts*. `PreCompact` requires an explicit `matcher` and its stdin payload carries no `trigger` field, so two hook entries (`matcher: "manual"` / `"auto"`) are registered with the trigger baked into argv instead.
- **`agent_config.rs`**: auto-injects those two `PreCompact` entries with the same user-hook-preserving merge discipline already used for `PreToolUse` — naively reusing the existing generic merge path here would have silently and permanently dropped any user-supplied `PreCompact` hook the moment this key became auto-injected, the same bug class PR #813 fixed for `PreToolUse`. Added two regression tests covering both merge paths (legacy `content_map["hooks"]` and `settings.json`'s own `hooks` key).
- **Frontend**: real `CompactionBoundary` data now drives the transcript's `context-compacted` node (real trigger + duration, not just inferred before/after counts) and a live "Compacting… Ns" readout in the composer strip. The existing ≥50%-drop heuristic is kept as a fallback for providers with no structured signal (codex/gemini/copilot) and suppressed for a window after a real boundary lands, so the same compaction event can't double-fire both paths.

## Test plan

- [x] `cargo test -p agentmux-srv translator::claude` — 26 passed (incl. 6 new: manual/auto trigger, unrelated subtype, missing/malformed metadata all degrade to no-op rather than panicking)
- [x] `cargo test -p agentmux-srv agent_config::` — 24 passed (incl. 2 new PreCompact merge-safety regression tests)
- [x] `cargo test -p agentmux-bashwrap` — 79 passed
- [x] `cargo build -p agentmux-srv` / `cargo build -p agentmux-bashwrap` — clean
- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx vitest run` on touched reducer/DocumentRow/renderer suites — 171+ passed, including the "exactly one context-compacted event, not two" invariant
- [ ] Live smoke test (spawn a Claude pane, run it past its context limit or type `/compact`, confirm the "Compacting…" status + live timer appear and the transcript shows the real trigger/duration) — not done in this environment; worth a manual check before merge


a5af:
@codex review

AgentA-asaf:
Addressed all three review findings in daa8da32:

- **reagent [P1]** (`compacting` can get permanently stuck): now also cleared by `StreamUnsubscribe`, `TurnEnd`, `TurnReset`, and `RequestStop`, not just a matching `CompactionBoundary`. New tests cover all four.
- **codex [P1]** (TS hook builder never installs `PreCompact`): `frontend/app/view/agent/agent-config-builder.ts`'s `buildSettingsWithHooks` now mirrors the Rust builder exactly — same two-entry (manual/auto) injection, same user-hook-preserving merge discipline. New tests mirror the Rust suite's PreCompact coverage.
- **codex [P1]** (stale replayed `compaction_started` after reconnect): `useCompactionStream` now validates the payload's own `startedAt` (previously ignored) and rejects anything older than a real compaction could plausibly still be running, extracted into a pure `resolveCompactionStart` with its own unit tests.

All touched suites re-verified: `cargo test` (unchanged, no Rust touched this round), `npx tsc --noEmit` clean, `npx vitest run` — 201 passed across the touched files.

@codex review

a5af:
@codex review

AgentA-asaf:
Addressed round 2 in 092d2207:

- **codex [P1]** (stale replay within the 10-min window): root-caused to the wrong mechanism — an age heuristic can't distinguish "recently finished" from "still running." Fixed at the source: `compaction_started` now publishes with `persist: 0` (never retained/replayed), since a disconnected pane already clears `compacting` on its own (previous fix) — there was nothing worth replaying on reconnect to begin with. The `startedAt` check remains as defense-in-depth only.
- **codex [P1]** (watchdog trips during a real compaction): `StreamWatchdogTick` now suspends entirely while `state.compacting` is set — a ~230s compaction was previously tripping both the 45s stream-stuck warning and the 180s force-demote-to-Idle.
- **codex [P2]** (compaction markers lost on history replay): extracted the shared frame-parsing into `compact-boundary.ts` and wired it into `parseHistoryLines.ts`. While fixing this, found and fixed a related bug in **both** the live and replay paths: bypassing the provider translator for this frame also bypasses the parser's text-accumulator reset, so text after a mid-turn compaction was silently merging into the text from before it (caught by a test asserting node order — `flushPending()` now runs explicitly at the boundary in both paths).

Full suite re-verified: `cargo test` (bashwrap 79, srv unchanged) + `cargo build` both crates clean, `npx tsc --noEmit` clean, `npx vitest run` — full suite 2148 passed, 0 failed.

@codex review

a5af:
@codex review

AgentA-asaf:
Round 3 addressed in 1e917e91:

- **reagent/codex [P1]** (both independently found the same bug): `FailureObserved` didn't clear `compacting` — fixed in 609d322a, both bots' latest reviews confirm this against the same root cause.
- **codex [P2]**: `RequestStop` was clearing `compacting` before the stop was confirmed. If `StopFailed` rolls back to Streaming, the compaction had actually kept running the whole time — removed the eager clear entirely; the four transitions that already, authoritatively end a turn are sufficient.
- **codex [P2]**: fixed a stale `persist: 1` reference in the spec doc left over from round 2's fix.

Full suite re-verified clean (one `tool-renderers/registry.test.ts` timeout under full-suite parallel load, confirmed flaky — passes in 2.8s isolated, well under its 5s timeout, and untouched by this round's changes).

@codex review

a5af:
@codex review

AgentA-asaf:
Round 4 addressed in 993c944b:

- **reagent [P1] + codex [P1]** (converged on the same finding): `InterruptTimeoutElapsed` was the fifth authoritative terminal transition that didn't clear `compacting` — fixed.
- Proactively swept the rest of the reducer for the same bug class rather than waiting for another round to find it piecemeal: `SubmitTimeoutElapsed` and `ReconcileTurnActive(active: false)` are two more authoritative transitions that could plausibly race with an in-flight compaction, both now clear it too. `TurnStartFailed` was deliberately left alone — its own doc comment establishes it only fires before any CLI process could exist for that turn attempt, so `compacting` can never legitimately be set when it does.

Full suite re-verified: reducer tests 174 passed, `npx tsc --noEmit` clean.

@codex review

a5af:
@codex review

AgentA-asaf:
Round 5 addressed in 315cb252 — this one's a root-cause fix, not another patch:

**reagent [P1]**: every prior round's fix gated *clearing* \`compacting\` on a transition out of a working phase, but nothing gated *setting* it on actually being in one. \`compaction_started\` arrives over a separate transport (WPS) from the primary NDJSON stream carrying \`TurnEnd\`/\`compact_boundary\`, so it can race and land after that same turn's \`TurnEnd\` already fired — orphaning \`compacting\` on an Idle/Done pane that none of the five existing clearing fixes ever touch again.

Fixed at the source: \`CompactionStarted\` now also no-ops unless the turn is actually in the working set, using the same \`workingFromPhase\` predicate \`FailureObserved\` already uses. This should close the whole bug class rather than requiring a round 6 to find transition #6 — the five existing clearing fixes stay as-is, since they cover the complementary case (compaction legitimately starts while working, then the turn ends before \`compact_boundary\` arrives).

Tests updated: the original "sets compacting while subscribed" test used an Idle base state, which was actually exercising the bug this fix closes — updated to use a genuinely-working base state, plus two new tests for the Idle and post-TurnEnd no-op cases. 176 reducer tests pass, \`tsc --noEmit\` clean.

@codex review

a5af:
@codex review

AgentA-asaf:
Round 6 addressed in d16ad023:

- **reagent [P1]**: found a real bug in round 5's own fix — the reducer now correctly rejects a stray \`compaction_started\` ping that races past \`TurnEnd\`, but \`useCompactionStream.ts\` ignored \`dispatchPane\`'s return value and pushed the transcript node regardless. Fixed by gating on a new, tested \`wasCompactionStartedAccepted()\`.
- **codex [P2]** (from round 4, missed at the time — caught while addressing round 6): the synchronous \`PreCompact\` hook inherited a 5s HTTP timeout meant for async tool-output streaming, which could delay real compaction by up to 5s if the backend was slow to respond — added a 500ms hook-specific timeout. Also added the missing \`compaction_boundary\` variant to \`gotypes.d.ts\`, the TS mirror of the Rust wire union.

Full suite re-verified: bashwrap 80 passed, \`tsc --noEmit\` clean, full vitest suite 2158 passed (one confirmed-flaky, unrelated \`tool-renderers/registry.test.ts\` timeout under parallel load — passes in isolation, same as two rounds ago).

@codex review

a5af:
@codex review

AgentA-asaf:
Round 7 addressed in 97e352e2:

- **reagent [PLAUSIBLE]**: a narrower race than round 5's fix — a stale \`compaction_started\` ping can arrive *after* its own matching \`CompactionBoundary\` while the turn is still working, since \`workingFromPhase\` alone doesn't distinguish "before" from "after" the boundary. \`CompactionStarted\` now also rejects any start at-or-before \`lastCompactionBoundaryAt\`.
- **codex [P2]**: \`wps_client.rs\` and \`precompact.rs\` each had their own private \`ENV_LOCK\`, which didn't actually synchronize anything since both mutate the same env vars — extracted one shared lock.

178 reducer tests pass, bashwrap 80 pass, \`tsc --noEmit\` clean.

@codex review

a5af:
@codex review

AgentA-asaf:
Round 8 addressed in 8763b46b:

- **codex [P2]**: the live path built its `context_compacted` node id from `Date.now()`, while the history-replay path (round 2's fix) built it from the frame's own stable `timestamp`. When the same compaction is processed live *and* separately re-seen in a mount-time history range, the mismatched ids meant the document store's dedup couldn't merge them — the compaction showed up twice. Threaded the frame's raw timestamp through as a dedicated dedup key (kept separate from `at`, which stays `Date.now()`-based since it drives real state like the watchdog and must reflect actual delivery time, not the frame's possibly-delayed embedded timestamp).

217 tests across the touched files pass, full suite 2163 passed (0 failed — same confirmed-flaky unrelated timeout as previous rounds), `tsc --noEmit` clean.

reagent has already approved (97e352e2) — this round is a codex-only fix on top of that.

@codex review

a5af:
@codex review

AgentA-asaf:
Round 9 addressed in 9d74f6b3:

- **codex [P2]**: `TurnStartFailed` didn't clear `compacting` — missed in round 4 on the wrong assumption that a synchronously-failed turn-start could never race a live `compaction_started` ping. Round 5's `workingFromPhase` gate explicitly permits it during `Submitting`, so it can.
- **codex [P2]**: `CompactionBoundary` cleared `compacting` unconditionally, but the two transports (NDJSON boundary vs. WPS start) have no ordering guarantee — a delayed boundary for an OLDER compaction could wipe a genuinely-active NEWER one. Now compares the boundary's own `frameTimestamp` against `compacting.startedAt` and only clears when it's for the currently-tracked (or older) compaction.

182 reducer tests pass, `tsc --noEmit` clean.

@codex review

a5af:
@codex review

GenericAgentX-asaf:
Round 10: addressed codex's P2 finding on commit 9d74f6b3.

- **codex P2**: `isStaleVsLastBoundary` compared `CompactionStarted.at` (the WPS payload's embedded true start time) against `lastCompactionBoundaryAt`, which was set from the boundary's frontend *receipt* time (`command.at`) rather than its own parsed completion time (`frameTimestamp`). A delayed boundary delivery could inflate that receipt time past a legitimately-later next compaction's true start, falsely rejecting it as stale. Fixed by using the parsed `frameTimestamp` when available, falling back to `command.at` when it's missing/unparseable (same fallback the sibling `preservesNewerCompaction` check already used).

Added a regression test reproducing the exact race (delayed boundary receipt vs. an earlier-but-genuinely-later true start) and updated one existing test whose assertion had baked in the old (buggy) receipt-time behavior. 183/183 reducer tests pass, `tsc --noEmit` clean.

Commit: 3259f9632

a5af:
@codex review

GenericAgentX-asaf:
Round 11: addressed reagentx's P2 finding on commit 3259f963.

- **reagent P2**: the `CompactionBoundary` handler set `lastContextTokens: command.postTokens` unconditionally, even when `preservesNewerCompaction` is true (a delayed boundary for an OLDER compaction arriving after a newer one already started). This overwrote the live context-fill reading with the older/stale value while a newer compaction was confirmed still in flight. Now gated the same way as `compacting`: preserved when this boundary belongs to a superseded compaction, only updated when it's the current one. The `context-compacted` transcript event still reports this boundary's own true tokens either way — only the live state gate changed.

Updated the existing test that had baked in the old (buggy) behavior, and added a dedicated test proving a real pre-existing context value (not just null) survives a stale delayed boundary. 184/184 reducer tests pass, `tsc --noEmit` clean.

Commit: b0170e5e8

a5af:
@codex review

GenericAgentX-asaf:
Round 12: addressed both codex findings on commit b0170e5e.

Both findings trace to the same root cause: the live path (`useAgentStream.ts`) and the history-replay path (`parseHistoryLines.ts`) each independently reimplemented the `context_compacted` node's id/timestamp fallback logic instead of sharing one definition, and drifted:

- **codex P2** (timestamp): the live path recorded the frontend's receipt time as the persisted node's timestamp, discarding the frame's own completion time (`frameTimestamp`) even when available. Since live/replay nodes now share an id, a later history replay could no longer correct a live node that already won the dedup.
- **codex P2** (id): the two paths used different fallbacks for a timestamp-less frame — live used `Date.now()`, replay used a batch-relative node count — so the same boundary seen by both could land on two different ids and show up twice.

Extracted `contextCompactedNodeId` / `contextCompactedLiveTimestamp` into `compact-boundary.ts` (the module that already exists specifically to stop these two consumers from drifting) and made both call the shared functions instead of reimplementing locally. The heuristic-sourced path is unaffected (still keys on `Date.now()` — nothing in history replay can ever collide with it).

Added tests for both shared functions and a dedicated history-replay test proving the timestamp-less id now matches what the live path would produce for the same frame. 215/215 tests pass (reducer + compact-boundary + parseHistoryLines), `tsc --noEmit` clean.

Commit: b0a1d97da

a5af:
@codex review

PR #2325, opened 2026-07-27
feat(armory): add bundle.export RPC — ABF Phase 1 exporter

AgentX-asaf:
## Summary
Implements **Phase 1** of the rollout plan in [`docs/specs/REPORT_ARMORY_BUNDLE_STANDARD_RESEARCH_2026_07_16.md`](https://github.com/agentmuxai/agentmux/blob/main/docs/specs/REPORT_ARMORY_BUNDLE_STANDARD_RESEARCH_2026_07_16.md) / [docs.agentmux.ai/abf](https://docs.agentmux.ai/abf/): "`armory export bundle <id>`" — serialize a bundle + its referenced skills/MCP servers into the ABF v0.1 on-disk layout. Pure read-side, no schema changes.

**Stacked on #2322 (Phase 0)** — this genuinely depends on `SKILL_TYPE_AGENT_SKILL`/`render_skill_md`/`unique_skill_slug`, all introduced there (now bumped to `pub(crate)` for reuse). Base branch is the Phase 0 branch, not `main` — this diff will shrink once #2322 merges and this rebases onto `main`.

## Changes
- **New `agentmux-srv/src/backend/bundle_export.rs`**: pure `export_bundle()` (`Memory` + resolved `Skill` rows → ABF file list + `armory.json` manifest) and `zip_bundle_export()` (in-memory `ZipWriter` — first writer usage in this workspace; `tool_store.rs` only ever reads zips).
- **New `bundle.export` RPC** (`app_api/bundle.rs`, `COMMAND_BUNDLE_EXPORT`): reads the bundle from `id_store`, resolves its `skills` id-array against `wstore` (bundles and skills live in different `Store` instances — traced from the existing `bundle.*` vs `skill.*` handlers), calls the pure exporter. `format:"zip"` returns a base64-encoded archive; otherwise returns the raw file list for the caller to write out.

## Design decisions worth flagging
- **No CLI binary exists** in this workspace (checked all 5 `[[bin]]` entries) — RPC is the idiomatic Phase-1 surface, matching the existing `bundle.*`/`skill.*`/`mcp.*` App API convention. "`armory export bundle <id> --dir/-o zip`" from the plan maps onto `format:"dir"` (default) vs `format:"zip"` on this one RPC.
[...]

PR #2379, opened 2026-08-01
feat(armory): ABF single-file format + importer (Phase 2)

AgentA-asaf:
## Summary

Follow-up to `docs/specs/REPORT_ARMORY_BUNDLE_STANDARD_RESEARCH_2026_07_16.md` — that report's research stands unchanged; this PR refines its packaging decision and implements the importer its Phase 2 left as a one-paragraph sketch.

**Ground truth check first** (see the new spec's §1): Phase 0 (skills alignment) and Phase 1 (exporter) were already shipped, including a zip option (`zip_bundle_export`) the original report had filed under "later, optional phase." Phase 2 (importer) did not exist at all — zero `bundle.import`, zero `bundle_import.rs`.

**This PR:**
1. **Spec** (`docs/specs/SPEC_ABF_V0_1_SINGLE_FILE_AND_IMPORTER_2026_08_01.md`) — promotes the `.abf` zip archive to the primary v0.1 interchange format (same internal layout, just packaged as one file instead of a bare directory tree — "an ABF file that gets loaded" needs a file). Also specifies the importer precisely enough to implement against.
2. **A spec self-correction, made during implementation, not after**: an earlier draft said a single unambiguous account match should be "substituted" into an imported MCP server's `${VAR}` placeholder. That's wrong — MCP server `env` values have no spawn-time resolution indirection the way CLI-provider identity does (checked `resolver/inject.rs` before assuming), so "substituting" would mean writing a real secret into the same DB column `bundle_export.rs`'s own redaction logic (env/header/args/URL scanning, added across PRs #2325/#2333) exists specifically to keep secrets out of. Account resolution is **read-only and purely informational** — a bundle's placeholders are never rewritten by import, full stop.
3. **`bundle_import.rs`** — pure parsing/validation, mirroring `bundle_export.rs`'s shape and its "warn, don't silently drop; reject structurally, not per-entry" philosophy. Reuses `sanitize_context_relative_path` (now `pub(crate)`) for the same path-traversal defense on the way *in* that the exporter already applies on the way *out* — an untrusted `.abf` needs it at least as much. Includes a SKILL.md frontmatter parser (the inverse of `render_skill_md` — didn't exist anywhere in the codebase) and a zip reader that strips the exporter's root-slug wrapping directory.
4. **`bundle.import` RPC handler** — accepts either `zip_base64` or a raw `files` list, creates one `Memory` bundle row (`is_global: false` — an import must not silently start injecting into every agent's CLAUDE.md without explicit user action) and one global `Skill` row per parsed skill, resolves account requirements read-only.

## Test plan

- [x] `bundle_import.rs`: 16 unit tests — missing/malformed `armory.json` rejected structurally; missing referenced files/skills warn-and-skip; path-traversal and duplicate-entry defense; `accounts/` allowlist (only `requirements.json` ever read); SKILL.md frontmatter round-trip (including special characters); a full `export_bundle` → `zip_bundle_export` → `unzip_bundle_import` → `parse_bundle_import` round trip.
- [x] `cargo test -p agentmux-srv bundle` — 54 passed (30 export + 16 import + others), 0 failed.
- [x] `cargo test -p agentmux-srv` (full suite) — 1871 passed, 0 failed, 4 ignored (pre-existing).
- [x] `cargo build -p agentmux-srv` — clean.
- [ ] No UI yet (Phase 3, separately tracked per the spec) — this ships backend/RPC only, so there's no click-through to verify beyond the RPC contract itself.

### DIFF
--- a/.changesets/1785618688-feat-armory-abf-single-file-abf-zip-format-importer-phase-2-bvcf.md
+++ b/.changesets/1785618688-feat-armory-abf-single-file-abf-zip-format-importer-phase-2-bvcf.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): ABF single-file (.abf zip) format + importer (Phase 2)
--- a/agentmux-srv/src/backend/bundle_export.rs
+++ b/agentmux-srv/src/backend/bundle_export.rs
@@ -323,12 +323,19 @@ fn redact_mcp_entry(entry: &Value) -> (Value, Vec<String>) {
     (redacted, redacted_names)
 }
 
-/// Validate a context-file's relative path is safe to place under
-/// `instructions/context/` in an exported bundle: non-empty, not absolute,
-/// no drive letter, no `..` traversal component. Pure string validation
-/// (no filesystem access) so [`export_bundle`] can stay a pure function.
-/// Returns `None` for anything that fails; callers skip that entry.
-fn sanitize_context_relative_path(path: &str) -> Option<String> {
+/// Validate a relative path is safe to place under a bundle export/import
+/// root: non-empty, not absolute, no drive letter, no `..` traversal
+/// component. Pure string validation (no filesystem access) so
+/// [`export_bundle`] can stay a pure function. Returns `None` for anything
+/// that fails; callers skip that entry.
+///
+/// `pub(crate)` (not private) so `bundle_import.rs` can reuse the exact
+/// same check on the way IN — an untrusted `.abf` file needs this defense
+/// at least as much as export needs it on the way out, and a single shared
+/// implementation means the two can never drift apart on what counts as
+/// safe (see `docs/specs/SPEC_ABF_V0_1_SINGLE_FILE_AND_IMPORTER_2026_08_01.md`
+/// §4.3.4).
+pub(crate) fn sanitize_context_relative_path(path: &str) -> Option<String> {
     if path.is_empty() {
         return None;
     }
--- a/agentmux-srv/src/backend/bundle_import.rs
+++ b/agentmux-srv/src/backend/bundle_import.rs
@@ -113,20 +113,31 @@ pub fn parse_bundle_import(files: &[BundleImportFile]) -> Result<ParsedBundleImp
     // re-export of that same bundle, written unredacted straight into
     // `instructions/AGENTS.md`) — defeating the accounts/ allowlist this
     // exact loop exists to enforce.
-    let by_path: HashMap<&str, &str> = files
-        .iter()
-        .filter(|f| {
-            let rejected = f.path.starts_with("accounts/") && f.path != "accounts/requirements.json";
-            if rejected {
-                warnings.push(format!(
-                    "{}: rejected — only accounts/requirements.json is ever read from the accounts/ directory",
-                    f.path
-                ));
-            }
-            !rejected
-        })
-        .map(|f| (f.path.as_str(), f.content.as_str()))
-        .collect();
+    // Codex P1, PR #2379 (round 2): the accounts/ allowlist check above
+    // only ever saw a ZIP entry's already-normalized path (unzip_bundle_import
+    // runs every entry through `sanitize_context_relative_path` first). The
+    // raw `files` RPC input skips that step entirely, so a spelling like
+    // `./accounts/secrets.json` or `accounts\secrets.json` doesn't match the
+    // literal `starts_with("accounts/")` check and sails through unrejected
+    // — while a manifest `components.*` entry using the exact same raw
+    // spelling still resolves it via `by_path.get(path)` below. Normalize
+    // every file's path the same way regardless of source (zip or raw
+    // list) before the allowlist check and before it becomes a lookup key,
+    // so both intake paths enforce the same rule on the same canonical form.
+    let mut by_path: HashMap<String, &str> = HashMap::new();
+    for f in files {
+        let Some(safe_path) = sanitize_context_relative_path(&f.path) else {
+            warnings.push(format!("{}: not a safe path; skipped", f.path));
+            continue;
+        };
+        if safe_path.starts_with("accounts/") && safe_path != "accounts/requirements.json" {
+            warnings.push(format!(
+                "{safe_path}: rejected — only accounts/requirements.json is ever read from the accounts/ directory"
+            ));
+            continue;
+        }
+        by_path.insert(safe_path, f.content.as_str());
+    }
 
     let manifest_raw = by_path
         .get("armory.json")
@@ -329,6 +340,15 @@ const MAX_ENTRY_UNCOMPRESSED_BYTES: u64 = 10 * 1024 * 1024;
 /// each individually pass [`MAX_ENTRY_UNCOMPRESSED_BYTES`].
 const MAX_TOTAL_UNCOMPRESSED_BYTES: u64 = 50 * 1024 * 1024;
 
+/// Maximum number of entries an archive may contain. A legitimate bundle
+/// (instructions, a handful of context files, a handful of skills, one
+/// manifest) never comes close to this; it exists purely to bound the CPU
+/// cost of iterating + path-sanitizing + hashing every entry, independent
+/// of the per-entry/aggregate byte caps below (reagent P2, PR #2379 round
+/// 2) — a crafted archive of many tiny/valid entries passes both size caps
+/// while still forcing unbounded iteration work.
+const MAX_ENTRY_COUNT: usize = 10_000;
+
 /// Unpack a `.abf` zip archive into a flat file list, applying the same
 /// path-safety check as everywhere else in this module to every entry
 /// name before it's trusted (§4.3.4) — a zip's own internal paths are
@@ -350,6 +370,13 @@ pub fn unzip_bundle_import(zip_bytes: &[u8]) -> Result<(Vec<BundleImportFile>, V
     use std::io::Read;
     let cursor = std::io::Cursor::new(zip_bytes);
     let mut archive = zip::ZipArchive::new(cursor).map_err(|e| format!("not a valid zip archive: {e}"))?;
+
+    if archive.len() > MAX_ENTRY_COUNT {
+        return Err(format!(
+            "zip archive: {} entries exceeds the limit ({MAX_ENTRY_COUNT}) — rejecting the whole import",
+            archive.len()
+        ));
+    }
 
     // First pass: collect every non-dir entry's raw name, to DETECT
     // whether every single one shares one common wrapping directory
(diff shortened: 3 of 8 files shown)


a5af:
@codex review

a5af:
@codex review

AgentA-asaf:
Round 1 addressed across two commits:

**3fb271fb** — findings from both bots, one confirmed independently by each:
- **reagent [P1] + codex [P2]** (same bug, found separately): \`unzip_bundle_import\` unconditionally stripped the first path segment of every entry, assuming a wrapper directory always exists — silently breaking a hand-built, unwrapped \`.abf\` (which the spec's §2 explicitly requires to also round-trip). Now detects whether *every* entry agrees on the same wrapper before stripping anything.
- **codex [P1]**: a rejected \`accounts/*\` file was excluded from warnings but not from the lookup map — a malicious bundle referencing \`accounts/secrets.json\` from \`components.instructions\` would still have gotten that content imported, defeating the whole allowlist. Fixed at the source: rejected files are excluded from lookups entirely.
- **codex [P1]**: \`read_to_string\` had no size bound — added a per-entry declared-size check, a hard backstop on the actual read, and an aggregate cap that fails the whole import.
- **codex [P2]** x2: the RPC handler treated every skill-creation error as a name conflict (masking real infra failures as a "successful" lossy import), and didn't roll back already-created skills on a later failure. Fixed both.

**ba72c28a** — the CI vitest failure: a static contract test verifies every backend RPC command has a frontend binding (or an explicit allowlist entry for "backend-only for now"). \`bundle.import\` has no frontend caller yet by design (Phase 3 UI is separately tracked) — added it alongside \`bundle.export\`, which already sits in the same allowlist for the identical reason.

28 backend tests pass (24 import + 4 new for these fixes... actually 24 total in bundle_import, +58 across all bundle tests), full `cargo test -p agentmux-srv` clean, `tsc --noEmit` clean, contract test passes.

@codex review

a5af:
@codex review

GenericAgentX-asaf:
Round 2: addressed all 4 findings from reagent + codex on commit ba72c28a.

- **reagent P1**: aggregate decompression cap accumulated the zip's own declared/attacker-controlled `entry.size()` instead of actual bytes read — an archive with many entries lying about their declared size while really containing content up to the per-entry cap would bypass the aggregate zip-bomb defense. Now accumulates actual bytes read after the per-entry backstop.
- **reagent P2**: no cap on entry count — added `MAX_ENTRY_COUNT` (10,000), checked upfront.
- **codex P1**: raw `files`-list paths (as opposed to zip entries) skipped path normalization entirely, so `./accounts/secrets.json` or `accounts\secrets.json` bypassed the literal `accounts/` allowlist check while a manifest reference using the identical raw spelling would still resolve it. Every file's path (zip or raw list) is now normalized via `sanitize_context_relative_path` before the allowlist check and before it becomes a lookup key.
- **codex P2**: account-requirement resolution queried `wstore` instead of `id_store` (where account CRUD actually lives — see `identity.rs`'s own handler registrations), reporting false negatives for accounts that actually exist.

Added regression tests for all four, including a byte-patched zip test that directly proves the aggregate cap survives a declared-size lie (patches the local file header's uncompressed-size field post-write while leaving real content untouched). 64/64 bundle-scoped Rust tests pass, full crate builds clean.

Commit: 7fc83994b

a5af:
@codex review

GenericAgentX-asaf:
Round 3: addressed codex's P1 finding on commit 7fc83994, and want to explain why I'm not applying the P2 suggestion as-is.

**codex P1 (fixed)**: `components.instructions`/`skills`/`mcpServers` are manifest-controlled arrays whose length isn't bounded by the decompression size caps — those cap file *content*, not how many times a manifest can reference the same path. A manifest repeating one path thousands of times cloned that content once per repetition, letting a sub-50MB archive expand to many gigabytes before the join. All three component loops now dedupe by path (warn + skip on repeat). Added tests for instructions, skills, and mcpServers.

**codex P2 (not applying the suggested fix)**: the observation — a shared/global bundle referencing a skill ID that only exists in the importing channel's local store — is real, but it's a pre-existing property of the whole app's architecture, not something this PR introduces or worsens. `bundle.upsert` (the pre-existing, non-import RPC handler) already writes bundles through `id_store` (shared — see its own doc comment: "identity/memory/drone/muxbus store... survive across version upgrades"). Every `skill.*` handler (`skill.list`, `skill.get`, `skill.upsert`, `skill.delete`, `skill.bind`, `skill.list_global`) consistently reads/writes `wstore` (channel-local) — skills were never made global. Moving the importer's skill writes to `id_store` to "match" the bundle would make imported skills invisible to every other skill handler in the app (list/get/bind would all miss them), which is a worse regression than the cross-channel gap it's meant to close. Any bundle with a `skills` reference — manually assembled via the UI or imported — has this same characteristic today; fixing it requires a broader skills-globalization decision outside this PR's scope, not a local change in the importer. Happy to file a follow-up if that's wanted.

Commit: b2f3d8ea7

a5af:
@codex review

GenericAgentX-asaf:
Round 4: addressed reagent's P1 and all 3 codex findings on commit b2f3d8ea.

- **reagent P1**: the accounts/ allowlist check was case-sensitive — `sanitize_context_relative_path` never case-folds, so `ACCOUNTS/secrets.json` bypassed the reject-and-warn branch while a manifest reference using the identical casing still resolved it. Now checks the directory prefix case-insensitively (the canonical-name exception stays an exact match against the lowercase spelling the exporter always emits).
- **codex P2** (path normalization): `by_path`'s keys are normalized (an earlier round's fix), but `components.instructions`/`skills`/`mcpServers` references were still looked up using the raw, un-normalized manifest string — a valid non-canonical spelling like `./instructions/AGENTS.md` no longer matched its normalized key and got reported "not found" even though the file was genuinely present. All three component loops now normalize their own reference the same way before dedup and lookup.
- **codex P1** (unbounded requirements DoS): `accounts/requirements.json`'s `requirements` array was unbounded, letting a small archive (well within the per-entry size cap) drive an unbounded number of downstream per-provider account lookups. Added a 1,000-entry cap in the parser, and the RPC handler now additionally queries each distinct provider once instead of once per requirement row.
- **codex P2** (silent rollback failure): skill rollback on a later failure discarded every `skill_delete` error via `let _ = ...` — if the same failing store also rejected the cleanup deletes, previously-inserted global skills could silently survive an RPC that reports failure. No transaction primitive exists on `Store` today, so full atomicity is out of scope for this PR; rollback failures are now collected and appended to the returned error instead of discarded, so the caller at least learns cleanup may not have fully succeeded.

Added regression tests for the 3 findings living in `bundle_import.rs` (case-insensitivity bypass, non-canonical reference resolution, requirements cap). The two `bundle.rs` handler-level changes (provider dedup, rollback surfacing) don't have dedicated new tests — there's no existing RPC-handler test harness for this handler (the only existing `bundle.rs` tests cover the pure `normalize_bundle_upsert_input` helper, not a full handler), so these are verified by code review rather than a new test harness built for this PR. 69/69 bundle-scoped Rust tests pass, full crate builds clean.

Commit: d58fda59f

a5af:
@codex review

GenericAgentX-asaf:
Round 5: addressed reagent's P1+P2 and codex's P1+P2 on commit d58fda59.

This round unified the size/DoS caps instead of patching around them again — the pattern across 4 rounds of size-cap findings pointed at a structural gap rather than another one-off bug:

- **reagent P1**: the raw `files` RPC branch had **zero** size/count enforcement, even though the spec treats it as an equally untrusted alternate ingestion path to `zip_base64` — a hostile bundle submitted via `files` bypassed every zip-bomb/DoS defense the zip path enforced.
- **codex P1**: even on the zip path, an entry whose *actual* decompressed size exceeded the per-entry cap was fully decompressed (paying the real ~10MB read) and then skipped *without* counting that work toward the aggregate budget — up to 10,000 such entries could force ~100GB of real decompression work while the 50MB aggregate cap never tripped.

Fix: extracted `check_entry_size` as a single choke point both intake paths now call — it accounts an entry's bytes toward the aggregate budget **unconditionally**, before deciding whether to keep the entry, closing the accounting gap structurally. `unzip_bundle_import` now calls it after its backstop read instead of inline checks; a new `enforce_raw_files_caps` applies the identical caps (plus the entry-count cap) to the raw `files` path.

- **reagent P2**: `by_path` construction had no duplicate-path detection for the raw files list (unlike zip's explicit "duplicate entry; first occurrence kept") — silently last-write-wins. Now warns and keeps the first occurrence.
- **codex P2**: the version-recognition check used a bare prefix match (`v.starts_with("0.1")`), so `"0.10.0"` and `"0.1garbage"` were both treated as recognized. Now matches `"0.1"` or `"0.1."` exactly.

Added regression tests for all four, including a direct unit test of `check_entry_size` proving the accounting-before-discard invariant (a zip-level reproduction would need to forge a central-directory-level declared size, which the `zip` crate reads instead of the local file header this module can cheaply patch — verified at the actual implementation site instead). 77/77 bundle-scoped Rust tests pass, full crate builds clean.

Commit: df6bd89aa

a5af:
@codex review

GenericAgentX-asaf:
Round 6: addressed reagent's P2 and all 3 codex P1 findings on commit df6bd89a.

- **reagent P2**: `components.accounts` was the one remaining manifest-reference lookup still comparing the raw, un-normalized string — the same bug class round 4 fixed for `instructions`/`skills`/`mcpServers`, just missed here. Now normalized the same way.
- **codex P1** (requirements.json leak): `accounts/requirements.json` is intentionally readable via `by_path` (the dedicated parser needs it), but was ALSO reachable through the generic `components.instructions`/`mcpServers` lookups — a manifest listing it there would copy its raw JSON into instructions (later written unredacted to `AGENTS.md` on export) or into an mcpServers entry. Added `is_requirements_json()` and rejected it from both generic lookups.
- **codex P1** (uncounted read-error bytes): an entry whose `read_to_end` ultimately errors (e.g. a forged CRC) can still leave real decompressed bytes in `buf` — those were previously discarded without accounting, so up to 10,000 corrupt entries could each force ~10MB of real decompression work while the aggregate cap never moved. `check_entry_size` now runs before the read-error branch, unconditionally.
- **codex P1** (warning amplification): the round-3 dedup fix avoided cloning content per duplicate reference, but still pushed one warning string per duplicate — its own amplification vector. All three component loops now push one bounded summary warning after the loop instead of one per duplicate.

Added regression tests for all four. 81/81 bundle-scoped Rust tests pass, full crate builds clean.

Commit: 319d11c7d

a5af:
@codex review

GenericAgentX-asaf:
Round 7: addressed reagent's P2 and both codex P1 findings on commit 319d11c7.

- **reagent P2**: `components.accounts` silently dropped a non-string value with no warning — inconsistent with every other component category in this module. Now warns explicitly.
- **codex P1** (warning amplification, take 2): round 6 bounded only the "duplicate reference" warning specifically; a manifest filled with non-string junk values hit a different (still unbounded) warning path for the same effect. Rather than special-casing this one too, added `capped_component_array` — truncates each `components.instructions`/`skills`/`mcpServers` array to `MAX_ENTRY_COUNT` entries **before** any per-entry processing, so one cap bounds every warning type a per-entry loop can produce, not just the ones already special-cased.
- **codex P1** (unbounded skill writes): the archive-entry-count cap (10,000) is far too high a ceiling for the RPC handler's *write* side — each imported skill becomes a separate synchronous Store transaction creating a permanent, globally-visible row. Added `MAX_IMPORTED_SKILLS` (200), applied in the parser so the handler never sees more than the cap.

Added regression tests for all three. 84/84 bundle-scoped Rust tests pass, full crate builds clean.

Commit: 789e38d39

a5af:
@codex review

GenericAgentX-asaf:
Merged. One outstanding non-blocking note from codex on 789e38d3 for the record: imported \`\$schema\`/\`version\` metadata isn't preserved through \`ParsedBundleImport\` today, so a later export always writes the hard-coded current \`0.1.0\` schema/version rather than round-tripping whatever was originally imported. Not an active bug (only v0.1.x exists right now, so there's nothing to actually mislabel yet), but worth tracking as a forward-compatibility follow-up once/if a v0.2+ format ever exists.

PR #2381, opened 2026-08-02
docs(specs): ABF import UI (Phase 3) -- selective import + collision handling

GenericAgentX-asaf:
## Summary

Spec for the ABF import modal (Phase 3), following up on PR #2379's importer backend (Phase 2), whose own spec explicitly deferred the UI as separately-tracked. Covers the three pieces the user asked for: file selection, item-selection UI, and collision handling.

Key design points:
- Splits `bundle.import` into `bundle.import.preview` (pure parse, no writes) and `bundle.import.commit` (writes only the user's selection) — the current RPC is all-or-nothing and can't support a real selection checklist without this.
- Audited the codebase for what can actually collide: **skills are the only item type with a real collision surface** today (global name uniqueness, create-only-if-unique). MCP servers have none — they're never promoted to a global catalog, a flagged-but-not-resolved asymmetry with skills. Bundle names and context files have none either.
- Three-step modal flow (select → preview/select → confirm) following the existing sequential-modal-chain pattern already used elsewhere in the app, since there's no existing wizard-panel component to imitate.
- Skill-collision resolution: instant client-side validation against a one-time-fetched skill list, with the server remaining the sole authority at commit time (race-safe).

Also generated a sample `.abf` fixture for manual testing once the UI exists (not checked in — regeneration steps are in the spec).

## Test plan

- [ ] N/A — spec only, no code changes in this PR.

### DIFF
--- a/docs/specs/SPEC_ABF_IMPORT_UI_PHASE3_2026_08_02.md
+++ b/docs/specs/SPEC_ABF_IMPORT_UI_PHASE3_2026_08_02.md
@@ -1,0 +1,300 @@
+# Spec: ABF Import UI (Phase 3) — Selective Import + Collision Handling
+
+**Date:** 2026-08-02
+**Status:** Spec — not yet implemented.
+**Relationship to prior work:** builds on
+`docs/specs/SPEC_ABF_V0_1_SINGLE_FILE_AND_IMPORTER_2026_08_01.md`, whose §4.1
+explicitly scoped Phase 2 to "backend + RPC only... no UI (Phase 3, still
+separately tracked)." This is that Phase 3. The existing `bundle.import` RPC
+(merged, `agentmux-srv/src/backend/bundle_import.rs` +
+`agentmux-srv/src/server/app_api/bundle.rs`) is all-or-nothing: it parses an
+`.abf`/directory tree and writes everything found in one shot. This spec adds
+a preview step and a selective commit, and specifies collision handling for
+each importable item type — none of which the current RPC does.
+
+---
+
+## 1. Why a new phase, not an extension of `bundle.import`
+
+The user-facing requirement is: pick a file, see what's inside, choose what
+to bring in, get told about (and resolve) anything that collides with
+existing data. `bundle.import` today can't support that even as a UI-only
+addition — it's a single atomic write, so "select items" would either be
+decorative (everything gets imported regardless of the checkboxes) or
+require a genuine backend split. This spec makes that split explicit:
+
+- **`bundle.import.preview`** — pure parse, zero Store writes. Returns
+  everything the modal needs to render a selection screen, including
+  collision status computed against current Store state.
+- **`bundle.import.commit`** — takes the same file bytes plus the user's
+  selections (and any collision resolutions) and writes only what was
+  selected.
+
+Both RPCs are stateless request/response, like everything else in this
+engine — no server-side session or preview-caching token. The frontend holds
+the picked file's bytes in modal state and sends them to both calls; a
+multi-MB base64 payload twice is a non-issue at the sizes this format
+already caps itself to (`MAX_TOTAL_UNCOMPRESSED_BYTES` = 50MB uncompressed,
+typically far less compressed).
+
+## 2. What can collide, and what can't (ground truth, verified against code)
+
+This determines which parts of the modal need collision UI at all.
+
+| Item | Collision surface today | Source |
+|---|---|---|
+| **Skills** | Yes — `skill_upsert_unique_global` rejects a new skill whose `name` matches an existing **global** skill (`is_global = 1`), regardless of id. Create-only-if-unique; no overwrite-by-name path exists. | `agentmux-srv/src/backend/storage/skills.rs` |
+| **MCP servers** | **None.** A global MCP catalog (`db_mcp_servers`, `mcp_server_upsert_unique_global`) exists, but `bundle.import` never calls it — imported MCP servers are written straight into the new bundle row's own `mcp_servers` JSON blob, never promoted to the global catalog. There is nothing for them to collide with. | `agentmux-srv/src/backend/storage/mcp_servers.rs`; confirmed absent from `bundle_import.rs`/`bundle.rs` |
+| **Bundle name** | **None enforced.** `bundle_memory_upsert` only conflicts on `id` (always a fresh UUID for an import); `name` has no uniqueness constraint — two bundles can freely share a display name. | `agentmux-srv/src/backend/storage/memory_bundles.rs` |
+| **Context files / instructions** | **None.** Each import creates a brand-new bundle row; there's nothing existing for a fresh bundle's `instructions`/`context_files` to collide with. | — |
+| **Account requirements** | **None** — read-only, informational (`resolved_requirement_ids` / `unresolved_requirements`), never written anywhere. | `bundle.rs` §4.5 handling |
+
+**Implication:** the only *hard* collision this UI needs to resolve is
+**skills**. Everything else either has no collision concept (MCP servers,
+context files, requirements) or is a soft, non-blocking naming nicety
+(bundle name). The spec below reflects that — most of the "collision
+handling" work is one well-scoped mechanism (skill name conflicts), not a
+generic system for every item type. MCP servers never being promoted to the
+global catalog is a **known, deliberate asymmetry with skills** carried over
+from the current backend, not something this UI spec resolves — flagged as
+an open question in §7.
+
+## 3. RPC design
+
+### 3.1 `bundle.import.preview`
+
+**Request** — identical shape to today's `bundle.import`:
+```jsonc
+{ "zip_base64": "..." }   // or: { "files": [{ "path": "...", "content": "..." }] }
+```
+
+**Response:**
+```jsonc
+{
+  "name": "Backend Dev Bundle",
+  "description": "...",
+  "instructions_preview": "Be concise. Prefer existing patterns...",
+  "context_files": [
+    { "path": "conventions.md", "size_bytes": 39 }
+  ],
+  "skills": [
+    { "slug": "deploy-checklist", "description": "...", "collision": "none" },
+    { "slug": "code-review", "description": "...", "collision": "name_conflict" }
+  ],
+  "mcp_servers": [
+    { "name": "github", "command": "npx", "args": ["-y", "@modelcontextprotocol/server-github"] }
+  ],
+  "requirements": [
+    { "id": "req-1", "provider": "github", "env": "GITHUB_TOKEN", "resolved": false, "match_count": 0 }
+  ],
+  "warnings": [ "components.instructions: ... skipped" ],
+  "name_collision": false   // true if an existing bundle already has this exact name (soft, informational)
+}
+```
+
+Implementation notes:
+- `parse_bundle_import` already produces everything here except
+  `collision`/`name_collision` — those are two new, cheap read-only lookups
+  the RPC handler adds: `wstore.skill_list_global()` filtered by name for
+  each parsed skill's slug, and a `id_store.bundle_memory_list()` (or
+  equivalent) name scan. Both are read-only; no new Store methods needed.
+- `instructions_preview` is the full instructions string, not truncated
+  server-side — truncate for display client-side if needed (keeps the RPC
+  contract simple; the size cap already bounds this to something reasonable
+  in the worst case).
+- Same `MAX_ENTRY_COUNT`/`MAX_ENTRY_UNCOMPRESSED_BYTES`/
+  `MAX_TOTAL_UNCOMPRESSED_BYTES`/`MAX_ACCOUNT_REQUIREMENTS`/
+  `MAX_IMPORTED_SKILLS` caps from Phase 2 apply unchanged — preview reuses
+  `parse_bundle_import`/`unzip_bundle_import`/`enforce_raw_files_caps`
+  as-is, just skips the Store-write half of today's handler.
+
+### 3.2 `bundle.import.commit`
+
+**Request** — the same file payload, plus selections:
+```jsonc
+{
+  "zip_base64": "...",             // or files[], same as preview — re-sent, not a cache token
+  "bundle_name": "Backend Dev Bundle (2)",   // user-editable, defaults to the parsed name
+  "include_instructions": true,
+  "include_context_files": ["conventions.md"],       // paths to include; omitted path = excluded
+  "include_skills": [
+    { "slug": "deploy-checklist" },
+    { "slug": "code-review", "import_as": "code-review-team-x" }   // rename to dodge a collision
+  ],
+  "include_mcp_servers": ["github"]
+}
+```
+
+**Response** — same shape as today's `bundle.import` response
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Round 1: addressed all 4 codex findings on commit 798b16f2.

- **P1 (file dialog)**: verified — `showOpenFileDialog()` takes no args and its host handler (`agentmux-cef/src/commands/platform.rs`) has a hard-coded image/video/audio filter. Spec now calls for a new `show_open_bundle_dialog` host command instead of reusing the existing one.
- **P1 (skill RPC)**: verified — `skill.list_global` doesn't exist. Corrected to `skill.catalog.list` (the window-scoped, no-`agent_id` route Armory already uses).
- **P2 (requirement resolution)**: verified — `parse_bundle_import` doesn't resolve requirements against connected accounts; that logic only exists in `bundle.rs`'s current commit handler. Spec now calls for extracting it into a shared helper both preview and commit call.
- **P1 (stable selection IDs)**: verified — neither a skill's slug nor an MCP server's JSON `name` field is guaranteed unique across parsed entries. Added §3.0, a required small Phase 2 amendment: `ParsedSkill` gains a `source_dir` field, `mcp_servers` becomes `Vec<ParsedMcpServer{source_path, config}>` — both already-computed values the parser currently discards. All selection wire formats now key on these instead of slug/name.

Commit: a29db5e15

a5af:
@codex review

GenericAgentX-asaf:
Round 2: addressed both codex findings on commit a29db5e1. Both were gaps in round 1's own §3.0 amendment, not new issues.

- **P1 (MCP config wrapper leak)**: verified — `bundle.rs`'s existing write path serializes `parsed.mcp_servers` directly into `Memory.mcp_servers`. Once that type changes to `Vec<ParsedMcpServer>`, that line would persist `{source_path, config}` wrappers instead of raw configs. Made explicit in §3.0: every write site (the retained `bundle.import` route AND the new commit handler) must project to `.config` before serializing.
- **P1 (intra-bundle duplicate slugs)**: verified — preview only checked the global catalog, so two skills within the SAME bundle sharing a slug not already global were both marked `"none"`, letting the user check both with commit silently skipping the second and no rename ever offered. Added a second detection pass (new `collision: "duplicate_in_bundle"` value) and made client-side rename validation check the union of the global list and other selected rows' current slugs, not just the global list.

Commit: 380925e55

a5af:
@codex review

GenericAgentX-asaf:
Round 3: addressed codex's P1 on commit 380925e5.

**P1 (WebSocket payload ceiling)**: verified against the pinned dependency versions — `axum 0.7.9` → `tokio-tungstenite 0.24.0`'s `WebSocketConfig::default()` sets `max_message_size` to 64 MiB, unmodified anywhere in `websocket.rs`. A bundle near the 50MB uncompressed cap base64-encodes to ~66.7MB — over that ceiling — and this spec's own preview-then-commit design would've hit it twice, on a payload the original draft wrongly assumed was a non-issue. This is a pre-existing Phase 2 limitation, not something Phase 3 introduces, but Phase 3's UI is what turns it from theoretical into routine.

Fix (new §3.0.5): both new RPCs gain a `file_path` input, read server-side — the natural shape for this flow since the file came from a local dialog in the first place (a path, not bytes). Mirrors an existing precedent in this exact codebase (`/agentmux/stream-local-file` already reads an arbitrary server-local path outside the WS transport for the same reason). `zip_base64`/`files` remain available for callers without a local path; their exposure to this limit is left as a separate, pre-existing, explicitly-noted limitation.

Commit: 35932542d

a5af:
@codex review

GenericAgentX-asaf:
Round 4: addressed both codex P1s on commit 35932542, both gaps in the file_path mechanism added in round 3.

- **P1 (unbounded pre-cap read)**: verified — `unzip_bundle_import`'s content caps all apply after the zip is already open and being read; nothing bounds the `file_path` intake step itself. Added an on-disk size check (`std::fs::metadata`, cheap) against a new `MAX_ABF_FILE_SIZE_BYTES` (100 MiB) **before** the file is opened for reading at all — same "cheap check first" pattern the per-entry declared-size pre-filter already uses one layer down.
- **P1 (TOCTOU on file_path)**: verified — a `file_path` is a live pointer to a mutable resource; the file can change between preview and commit with no detection, and stable `source_dir`/`source_path` values don't protect against a replacement archive keeping the same paths while changing actual content. Added a `content_digest` (SHA-256) returned by preview and required back at commit as `expected_content_digest` — a mismatch is a hard rejection, not a partial import. Scoped specifically to `file_path`; `zip_base64`/`files` callers already hold identical bytes on both calls by construction, so nothing can drift for them.

Commit: c0a53c9ef

a5af:
@codex review

GenericAgentX-asaf:
Round 5: addressed all 3 codex findings on commit c0a53c9e.

- **P1 (unbounded preview response)**: verified — a parser-accepted bundle can put nearly all of its 50MiB budget into `instructions`, and JSON string-escaping can expand pathological content up to 6x — a ~300MiB response, which would blow the same 64MiB WS ceiling round 3 fixed on the request side, just on the way back out. Added a `MAX_INSTRUCTIONS_PREVIEW_CHARS` (50,000) server-side cap + `instructions_truncated` field. The committed bundle's actual `instructions` is unaffected — only the preview response is capped.
- **P2 (digest scoped too narrowly)**: verified — round 4 scoped digest-binding to `file_path` only, reasoning `zip_base64`/`files` callers "already hold the bytes by construction." Correctly called out as an unenforced assumption, not an API guarantee. Generalized: `content_digest`/`expected_content_digest` now apply uniformly to all three input modes (canonical hash per mode — raw bytes, decoded zip bytes, or a sorted/length-prefixed hash of `files[]` so ordering can't change the digest).
- **P2 (metadata-then-read TOCTOU)**: verified — checking size via a separate `std::fs::metadata(path)` call, then re-resolving the same path to read it, leaves a window for another process to grow/replace the file in between. Fixed to open the file exactly once and read through a bounded reader on that same handle — the identical hard-backstop-read pattern `unzip_bundle_import`'s own per-entry decompression already uses.

Commit: 9062c73e8

a5af:
@codex review

GenericAgentX-asaf:
Round 6: addressed all 3 codex findings on commit 9062c73e (confirmed with the user to continue the review loop given this was the 6th round on a docs-only PR — all 3 verified as genuine design bugs, not nitpicks).

- **P1 (files digest doesn't reflect parser's actual ordering)**: verified — `parse_bundle_import`'s `by_path` construction normalizes paths first and keeps whichever entry appears first for two that normalize to the same key. A naive path-sort could make two differently-ordered request bodies hash identically while the parser would import different content from each. Fixed to hash the parser's own effective, order-resolved (normalize-then-first-wins) representation instead of raw input.
- **P2 (cross-mode digest claim was unimplementable)**: verified — `file_path`/`zip_base64` hash raw zip bytes while `files` hashes canonicalized extracted entries; these never produce equal digests for the same content, so the earlier claim that commit could freely switch input modes from preview was wrong. Fixed: commit must use the same input mode preview used — the actual modal flow always uses `file_path` for both calls anyway, so this costs nothing real.
- **P2 (symlink bypass)**: verified — `File::open` transparently follows symlinks, so round 5's "open once, check metadata on that handle" never actually rejected them. Fixed to require an explicit no-follow open.

Commit: ad6927f1e

a5af:
@codex review

GenericAgentX-asaf:
Round 7: addressed both codex P2 findings on commit ad6927f1 (continuing without pausing, per your standing confirmation to keep the loop running).

- **P2 (digest mode-binding gap)**: verified — `file_path` and `zip_base64` both canonicalize to the raw zip bytes, so a `file_path` preview and a `zip_base64` commit of the identical archive produce the same `content_digest`, silently passing exactly the mode-mismatch case round 6 was meant to reject. Fixed by mixing a mode tag into the hash domain (`mode_byte || canonical_representation`) so different modes always produce different digests for identical content.
- **P2 (unbounded MCP preview data)**: verified — preview returned the full raw `config` JSON per MCP server, but a parser-accepted bundle can put a large share of its content budget into one of these blobs — the same "lightweight response actually approaches the aggregate cap" problem `instructions_preview`'s round-5 cap exists to prevent, just uncapped here. Fixed: preview returns a bounded `{name, command}` display projection instead; the full config is read fresh at commit time.

Commit: 4ec48265b

a5af:
@codex review

GenericAgentX-asaf:
Round 8: addressed both codex findings on commit 4ec48265.

- **P1 (unbounded warnings)**: verified — the same unbounded-response class as `instructions_preview` (round 5) and `mcp_servers[].config` (round 7), a third field nobody had capped yet. An archive with thousands of unsafe/malformed entries pushes one warning per bad entry, each embedding the entry's raw (uncapped-length) name — accumulated text could reach hundreds of MB. Fixed: each warning truncated to a char limit, the array capped to a fixed count, with `warnings_truncated` + a summary entry when either cap is hit — mirrors the bounded-array-plus-summary shape `bundle_import.rs` already uses for its own duplicate-reference warnings.
- **P2 (missing instructions_total_chars)**: verified — the spec promised the modal shows "N characters total" for truncated instructions but never specified a field carrying that count. Added `instructions_total_chars`, always present regardless of truncation.

Commit: 70fc7a558

a5af:
@codex review

GenericAgentX-asaf:
Round 9: addressed codex's P1 on commit 70fc7a55.

**P1 (commit response still unbounded)**: verified, and this one is worse than the preview-side version round 8 fixed — round 8's warnings cap was applied to `preview` only; `commit`'s response still described "the same shape as today's `bundle.import`" (unbounded). Since `commit`'s Store writes can **succeed** before response serialization fails against the same 64MiB WS ceiling, the caller would see a failure for an import that actually went through — and since `bundle.import.commit` isn't idempotent (always mints a fresh bundle UUID), a plausible retry would create a **second, duplicate bundle**.

Fixed by applying the identical bounded-warning projection to commit's response, computed after the write loop completes (the writes stay based on full untruncated data; only the reported-back warnings are capped). Also specified this as one shared function both RPCs call, not two copies — the exact kind of drift that just happened is what that prevents going forward.

Commit: f6cf46000

a5af:
@codex review

GenericAgentX-asaf:
Round 10: addressed codex's P2 on commit f6cf4600.

**P2 (unbounded skill descriptions)**: verified — the fourth unbounded-response field found across rounds 5/7/8/9: preview's `skills[].description` returns SKILL.md's frontmatter description verbatim, with no length limit. With up to 200 imported skills and up to 10MiB allowed per entry, a bundle could return tens of megabytes of description text, unlike instructions/MCP-display/warnings which are already bounded for the same reason. Fixed with a fixed-character truncation (skill descriptions are meant to be short one-liners by convention, so no separate total-length field is needed here unlike `instructions_preview`). Full description unaffected at commit time.

Commit: b52cfea38

a5af:
@codex review

GenericAgentX-asaf:
Round 11: addressed all 3 codex findings on commit b52cfea3.

- **P2 (requirement fields unbounded)**: verified — `requirements[]`'s `id`/`provider`/`env` strings, a fifth field in the same class as skill descriptions. Fixed with the same fixed-character truncation.
- **P2 (bundle_name inert — functional bug, not just a size issue)**: verified, and this one's important — the write loop as described never actually substituted the request's `bundle_name` for `parsed.name`, so the entire rename-before-import UX (editable name field, collision-suggested alternate) would have been silently inert regardless of what the user typed. Fixed: commit's write loop now uses `bundle_name` for `Memory.name`, mirroring how `import_as` already correctly overrides a skill's slug.
- **P2 (warnings capped too late)**: verified — rounds 8/9's cap ran at the RPC response boundary, after the full unbounded warning list was already built in memory. Moved enforcement to the parser's own warning-push sites via an explicit budget parameter — the existing `bundle.import` route passes an effectively-unbounded budget (preserving its current behavior exactly), the new preview/commit handlers pass the tight budget. The response-boundary projection is now a cheap final consistency pass, not the primary defense.

Commit: a18883630

a5af:
@codex review

GenericAgentX-asaf:
Round 12: addressed both codex findings on commit a1888363, plus a self-audit catch.

- **P2 (unbounded skill slugs)**: verified — `parse_skill_md` places no length limit on SKILL.md's `name:` frontmatter. Fixed with the same truncation `description` already had; `source_dir` remains the real selection key, and commit's `import_as` validates against the full-length slug regardless of what preview displayed.
- **P2 (commit response not reusing preview's bounds)**: verified — `skipped_skills`/`resolved_requirement_ids`/`unresolved_requirements` never got the bounding preview's equivalent fields have, the same drift class round 9 already fixed once for `warnings`. Fixed those specific fields, and generalized to a governing rule: every bounded-display projection is one shared function both RPCs call, stated once instead of re-discovered field by field.
- **Self-audit (not a codex finding)**: per the instruction to proactively check before implementing, verified `parse_bundle_import`'s actual code — confirmed bundle-level `name`/`description` and `context_files[].path` also have no length bound anywhere. Fixed proactively with the same pattern before it became round 13.

Commit: 86f799f24

a5af:
@codex review

GenericAgentX-asaf:
Round 13: addressed all 3 codex findings on commit 86f799f2. All three exposed the same underlying design flaw in round 12's approach: truncate-for-display only works for pure-display fields, not fields also used as selection keys or edit-prefills that get submitted unchanged.

- **P2 (context_files selection silently drops truncated entries)**: verified — `include_context_files` matched against paths, but a truncated `display_path` wouldn't match the freshly-parsed full path, silently dropping that file with no error. Fixed by giving context files the same `source_dir`/`slug`-style split skills already have: a stable `id` (deterministic index) as the real selection key, `display_path` as the truncatable display value.
- **P2 (bundle name silently shortened on no-edit)**: verified — a user who never touched the pre-filled name field would import under a truncated name, since the frontend was never given the full value. Fixed by bounding `name` at the parse source instead of the response boundary — preview and commit converge on the identical canonical name deterministically, no separate "real" value to lose.
- **P2 (rename input could auto-submit a truncated slug)**: verified — pre-filling the rename input with the truncated slug meant importing-without-editing could silently rename a colliding skill to a truncated prefix. Fixed by having the rename input start blank (slug shown as read-only label instead), so "empty = skip" has no ambiguous case.

Commit: 3f81af29d

a5af:
@codex review

GenericAgentX-asaf:
**Round 14 fix pushed** (`84dfa58ae`)

codex P2 was real: the round-12 implementation note under skill `slug` truncation still described the *old* rename-input behavior ("pre-filled from this possibly truncated display `slug` — that's fine...") even though round 13 explicitly changed the contract so the rename input **starts blank** (§4.1 point 5), with `slug` shown only as read-only label text. Two contradictory instructions for implementers in the same doc.

Fixed by removing the stale "pre-filled ... that's fine" claim and pointing implementers at §4/§4.1 as the single source of truth for how the input initializes. The passage's only remaining (still-true) point is narrower: the *write decision* at commit never depends on the truncated display value, regardless of how the UI initializes the field.

Re-requesting codex review.

a5af:
@codex review

PR #2382, opened 2026-08-02
ABF import UI (Phase 3): selective import + collision handling

GenericAgentX-asaf:
Implements docs/specs/SPEC_ABF_IMPORT_UI_PHASE3_2026_08_02.md end to end -- the spec merged via #2381 after 14 review rounds.

## Summary

- **Backend struct amendments** (`bundle_import.rs`): `ParsedSkill.source_dir`, `ParsedMcpServer{source_path, config}`, `ImportedContextFile.id`, parse-time bundle-name length bounding -- stable selection keys distinct from truncatable display values, per §3.0/§3.1.
- **Shared helpers**: `WarningBudget`/`WarningSink` (parser-level warning bounding, unbounded default preserving today's `bundle.import` route exactly), bounded-display-projection helpers (`truncate_display`, `bounded_instructions_preview`, `mcp_server_display`), two-pass skill collision classification (`duplicate_in_bundle_slugs`/`classify_skill_collision`), and `resolve_account_requirements` extracted out of the existing `bundle.import` handler for reuse.
- **`bundle.import.preview` / `bundle.import.commit` RPC handlers**: `file_path` input support (no-follow single-handle bounded read -- `O_NOFOLLOW` on Unix, `FILE_FLAG_OPEN_REPARSE_POINT` + `is_symlink()` on Windows), mode-tagged SHA-256 content digest for all three input modes (`file_path`/`zip_base64`/`files`), digest-gated commit, `bundle_name` actually applied to `Memory.name`, context files selected by stable `id` not truncatable path, skills server-side-authoritatively skipped when left with an empty rename on a collision.
- **`show_open_bundle_dialog`** host command (agentmux-cef) + `cef-api.ts` binding, mirroring `show_open_file_dialog` with an `.abf` filter.
- **Frontend 3-step modal** (select → preview/select → confirm), following the existing `ModalLayerApi.replace()` chain pattern (install→launch precedent), wired into the Armory Bundles tab's new "Import Bundle" button.

## Testing

- 68 new/updated Rust unit + integration-style tests (`bundle_import.rs` + a new `bundle::import_preview_commit_tests` module using the existing `test_state()` harness) -- collision classification against both a seeded global skill and an intra-bundle duplicate, digest-mismatch rejection writing nothing, `bundle_name` override, context-file id-based selection survives truncation, skip-vs-rename skill collision handling, raw MCP config persistence (not the `{source_path, config}` wrapper), `read_abf_file_path`'s missing-path/directory/oversized-via-sparse-file/symlink(Unix) rejection paths, and the digest mode-tag differentiation. Full backend suite: 1940 passed, 0 failed.
- `tsc --noEmit` and the full `vitest` suite pass (one unrelated pre-existing flaky timeout in `tool-renderers/registry.test.ts`, confirmed clean in isolation).
- `agentmux-cef` builds clean.

### DIFF
--- a/.changesets/1785660132-feat-armory-abf-import-ui-phase-3-selective-import-collision-2z7d.md
+++ b/.changesets/1785660132-feat-armory-abf-import-ui-phase-3-selective-import-collision-2z7d.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): ABF import UI (Phase 3) -- selective import + collision handling
--- a/agentmux-cef/src/commands/platform.rs
+++ b/agentmux-cef/src/commands/platform.rs
@@ -704,6 +704,30 @@ pub async fn show_open_file_dialog(
     })
 }
 
+/// File picker for Armory Bundle Format (`.abf`) files — Phase 3 of
+/// docs/specs/SPEC_ABF_IMPORT_UI_PHASE3_2026_08_02.md §4 Step 1.
+/// `show_open_file_dialog` can't be reused: it takes no filter argument and
+/// its own filter list is hard-coded to image/video/audio extensions, with
+/// no generic filter mechanism elsewhere in this module — this mirrors its
+/// shape with an `.abf` filter instead. The filter is advisory only (a
+/// non-`.abf` file picked here still just fails `unzip_bundle_import`'s
+/// "not a valid zip archive" check server-side, same as today).
+pub async fn show_open_bundle_dialog(
+    _args: &serde_json::Value,
+) -> Result<serde_json::Value, String> {
+    let path = tokio::task::spawn_blocking(|| {
+        rfd::FileDialog::new()
+            .add_filter("Armory Bundle", &["abf"])
+            .pick_file()
+    })
+    .await
+    .map_err(|e| format!("show_open_bundle_dialog: task join error: {e}"))?;
+    Ok(match path {
+        Some(p) => serde_json::json!(p.to_string_lossy()),
+        None => serde_json::Value::Null,
+    })
+}
+
 fn extract_commented_setting_key(line: &str) -> Option<&str> {
     let trimmed = line.trim_start();
     let rest = trimmed.strip_prefix("//")?;
--- a/agentmux-cef/src/ipc.rs
+++ b/agentmux-cef/src/ipc.rs
@@ -273,6 +273,7 @@ async fn route_command(
         "open_external" => commands::platform::open_external(args),
         "reveal_in_file_explorer" => commands::platform::reveal_in_file_explorer(args),
         "show_open_file_dialog" => commands::platform::show_open_file_dialog(args).await,
+        "show_open_bundle_dialog" => commands::platform::show_open_bundle_dialog(args).await,
         "set_window_transparency" => commands::window::set_window_transparency(state, args),
         "set_window_opacity" => commands::window::set_window_opacity(state, args),
         "get_window_opacity" => commands::window::get_window_opacity(state, args).await,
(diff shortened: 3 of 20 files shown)


a5af:
@codex review

GenericAgentX-asaf:
**Round 1 fixes pushed** (\`b3f2dff72\`)

Addressed reagentx's 3 findings on commit \`d418a4822\` (codex had no findings this round):

- **P1 (missing changeset)**: verified — the queue's last changeset was Phase 2's. Added one for this Phase 3 implementation.
- **P2 (stale doc comment)**: verified, and investigating it surfaced a second, related bug the same code extraction introduced — `parse_bundle_import`'s own doc comment had gotten merged into `dedup_files_by_path`'s (missing blank-line separator), leaving `parse_bundle_import` completely undocumented. Fixed both: each function now has its own clean, accurate doc comment.
- **P2 (dead condition in BundleImportConfirmModal.tsx)**: verified — `s.renameValue.trim().length >= 0` is always true. Removed; the adjacent comment already states the intended behavior (count checked rows regardless of rename state), which the fix now matches.

\`cargo test -p agentmux-srv\`: 1940 passed, 0 failed. \`tsc --noEmit\`: clean.

Re-requesting codex review.

a5af:
@codex review

GenericAgentX-asaf:
**Round 2 fixes pushed** (\`3ffae39b0\`)

Addressed reagentx's 2 findings on commit \`b3f2dff72\` (codex had no findings this round):

- **P2 (rename-conflict check only used preview's own flags)**: verified — my own comment claimed "the global catalog" but the code never actually fetched it, only reused `preview.skills`' already-known `name_conflict` rows. Spec §4.1 point 2 explicitly requires fetching `skill.catalog.list` up front. Fixed by fetching on mount and unioning into the taken-slugs check; added 2 component tests pinning both the fetch-and-warn case and the no-false-positive case.
- **P2 (unbounded `import_as` in commit's "already exists" warning)**: verified, and it's worse than described — `StoreError::Other`'s own message text embeds the same unbounded value a *second* time internally. Fixed by using a fixed message string (the branch guard already establishes "already exists") plus the shared `bounded_display` projection. Added a backend test that seeds a global skill whose name IS an oversized `import_as` value, forcing the exact path, and asserts every warning stays bounded.

\`cargo test -p agentmux-srv\`: 1941 passed, 0 failed. \`tsc --noEmit\`: clean. \`vitest\`: 2 new tests passing.

Re-requesting codex review.

a5af:
@codex review

GenericAgentX-asaf:
**Round 3 fixes pushed** (\`ae76c6cce\`)

Note: reagentx attributed finding #1 to "an unaddressed codex P1 left in round 2" — I checked, and that's incorrect. Every codex review on this PR so far (rounds 1-3) has been empty boilerplate with no findings text; codex never left that finding. Flagging the misattribution for the record, but both underlying technical claims were independently verified against the actual code and are real, so both are fixed:

- **P1 (unbounded/undeduped `include_skills` write loop)**: verified — a client repeating the same `source_dir` with a different `import_as` each time (or sending a large distinct array) could drive far more `skill_upsert_unique_global` writes/iterations than the archive's own `MAX_IMPORTED_SKILLS` cap allows. Fixed with dedup (first occurrence wins) + a cap at the same `MAX_IMPORTED_SKILLS` the parser enforces (now `pub`, reused rather than duplicated). Two new tests cover the dedup and prove the cap truncates the selection itself.
- **P2 (unbounded `bundle_name` override)**: verified — unlike `parsed.name` (bounded at parse time specifically so preview/commit converge), a client override bypassed that bound entirely. Extracted the plain-truncation logic into a shared `bound_bundle_name`, used by both paths. New test confirms.

\`cargo test -p agentmux-srv\`: 1944 passed, 0 failed. \`tsc --noEmit\`: clean.

Re-requesting codex review.

a5af:
@codex review

PR #2386, opened 2026-08-03
fix(frontend): consolidate 5 duplicated elapsed-time formatters, fix negative-duration gap

AgentA-asaf:
## Summary
- Consolidates 5 independently-duplicated elapsed-time formatters (`AgentComposerStrip.tsx`, `AgentFooter.tsx`, `ActivityRow.tsx`, `PersistentShellBlock.tsx`, `swarm-view.tsx`) into `frontend/util/format-time.ts`.
- Two genuinely different display conventions were already live in the product — prose ("3m 5s") vs clock ("3:05") — so this exports both (`formatElapsedCompact`, `formatElapsedClock`) rather than forcing one everywhere. See `docs/specs/SPEC_TOKEN_STATS_NUMBER_FORMATTING_2026_08_02.md` §7.1.
- Fixes a real bug along the way: `PersistentShellBlock.tsx`'s copy was missing the `Math.max(0, ...)` floor guard the other clock-style copies already had — it could render a negative duration (e.g. `-1:05`) on a clock-skew edge case.
- Second PR in the display-formatting consolidation sequence (follows #2385, format-count.ts — merged).

## Test plan
- [x] New `frontend/util/format-time.test.ts` — 5 tests covering both formatters, boundary values, and negative-duration flooring.
- [x] Full frontend suite: 2216/2216 passing (verified with a real `npm install`, not a dev-worktree node_modules junction).
- [x] `tsc --noEmit` clean, project-wide.

PR #2385, opened 2026-08-03
fix(frontend): consolidate 7 duplicated token-count formatters, add k/m/b rollover

AgentA-asaf:
## Summary
- Consolidates 7 independently-duplicated "format a count as Xk" implementations across 6 files into a single shared `frontend/util/format-count.ts` (`formatCompactNumber`, `formatExactNumber`).
- Fixes the actual bug that motivated this: none of the 7 duplicates ever rolled over past "k" — a heavy session's cumulative token totals would render as an ever-growing `12345.7k` instead of `12.3m`.
- Migrated call sites: `AgentComposerStrip.tsx` (`fmtTokens`, context-fill text, `contextTitle`'s tooltip), `AgentFooter.tsx` (`fmtTokens`), `TokenUsageIndicator.tsx`, `TokenBreakdownPopover.tsx`, `swarm-view.tsx` (`fmtCtx`), `DocumentRow.tsx` (compaction transcript node).
- Full audit + design in `docs/specs/SPEC_TOKEN_STATS_NUMBER_FORMATTING_2026_08_02.md`, including a broader sweep (§7) for other duplicated display-formatting utilities (elapsed-time, string truncation, sleep) — those are scoped as separate follow-up PRs, not included here.

## Test plan
- [x] New `frontend/util/format-count.test.ts` — 9 tests covering every tier boundary, the pre-existing floating-point rounding-at-boundary quirk (documented, not a regression), and negative-number defensiveness.
- [x] Updated 2 `DocumentRow.test.tsx` fixtures whose hardcoded expectations assumed the old integer-only-under-10k precision (now `4.0k`/`5.0k`, matching the documented decimal-precision rule).
- [x] Full frontend suite: 2205/2205 real tests passing (2 unrelated failures confirmed as sandbox/filesystem artifacts of my dev worktree setup, not regressions — reproduced clean on the main tree).
- [x] `tsc --noEmit` clean, project-wide.

PR #2387, opened 2026-08-03
fix(frontend): consolidate string-truncation formatters, remove dead code

AgentA-asaf:
## Summary
- Consolidates 3 independently-duplicated truncate/abbreviate implementations (`block/autotitle.ts`, `view/drone/drone-view.tsx`, `AgentFooter.tsx`) into `frontend/util/format-text.ts`.
- **Bonus find**: `autotitle.ts`'s copy was genuinely dead code — not exported, zero call sites anywhere in the file. Deleted outright rather than migrated.
- Real design correction made during implementation (documented in the spec): an early draft auto-detected `/`/`\` in any input string and left-truncated it, framed as a straightforward generalization of `AgentFooter.tsx`'s path-aware behavior. That would have silently changed `drone-view.tsx`'s truncation direction for URLs/expressions/templates (where a `/` is incidental, not a path) — caught by tracing its actual call sites before shipping. Fixed: `pathAware` is an explicit opt-in (default `false`), and only `AgentFooter.tsx` (whose inputs are genuinely file paths / shell args) opts in.
- Third PR in the display-formatting consolidation sequence (follows #2385 format-count.ts and #2386 format-time.ts, both merged).

## Test plan
- [x] New `frontend/util/format-text.test.ts` — 5 tests covering both truncation directions, the `pathAware` opt-in default, and the real-ellipsis character.
- [x] Full frontend suite: 2220/2220 real tests passing (1 unrelated pre-existing timing flake in `registry.test.ts`, already confirmed independent of this change on the main tree).
- [x] `tsc --noEmit` clean, project-wide.


a5af:
@codex review

AgentA-asaf:
@codex review

PR #2392, opened 2026-08-03
feat(agent-pane): hover-to-peek on tool calls and thinking clumps

AgentA-asaf:
## Summary

Implements `docs/specs/SPEC_TRANSCRIPT_NODE_HOVER_PEEK_2026_08_03.md` — hover-to-peek on tool calls (extending existing behavior) and thinking clumps (new).

**Important context, spelled out in the spec's §0**: this is the third pass at broadly this idea. Two prior specs explicitly *removed* earlier versions — `SPEC_TOOL_HOVER_CONSOLIDATION_2026_05_28.md` killed ToolBlock's three-overlapping-popups-on-hover after a direct user complaint, and `SPEC_REMOVE_NODE_HOVER_STRIP_2026_06_15.md` deleted a generic per-row hover strip entirely. This PR extends the ONE hover pattern that survived (ToolBlock's narrow, read-only command-text tooltip) rather than reintroducing what was removed — no hover-triggered expand/collapse anywhere, no new floating strip, one tooltip per node anchored to a specific existing element.

## What changed

- **`ToolBlock.tsx`**: existing tool-name tooltip gains exact time + time-ago + an estimated token count, prepended above the already-shown word-wrapped command text.
- **`MarkdownBlock.tsx`**: thinking clumps get the same peek tooltip — previously had no hover surface at all.
- **Styling**: new `.agent-node-peek-tooltip*` CSS (13px, more generous spacing) replaces the old cramped `.agent-tool-cmd-tooltip` (11px) — reads as comfortably as the native browser tooltip, per direct user feedback comparing the two.
- **`frontend/util/format-time.ts`**: adds `formatTimeAgo`/`formatExactTime`, migrating `MyAgentsList.tsx`'s tested `formatRelative` rather than adding a 7th duplicate relative-time implementation.
- **`frontend/util/format-count.ts`**: adds `estimateTokenCount` (chars÷4, labeled "(est.)") — no real per-node token data exists anywhere today; see `docs/specs/SPEC_PER_NODE_TOKEN_ACCOUNTING_2026_08_03.md` for the real, Claude-only derivation designed as a follow-up phase.
- **`stream-parser.ts`**: real gap found during the audit — `thinkingToNode()` never stamped a timestamp at all, unlike every other node kind. Fixed (stamped once, on first creation of the clump).

**Deliberately deferred** (flagged in the spec, not silently dropped): a derived duration line for thinking clumps needs a new prop threaded through `AgentDocumentVirtualList.tsx`'s windowed-rows/streaming-buffer machinery — performance-critical, carefully tuned code with several past-bug references in its own comments. Not worth the risk for a nice-to-have; follow-up.

## Test plan
- [x] New/updated tests: `format-time.test.ts` (+7), `format-count.test.ts` (+2), `stream-parser.test.ts` (+1), `ToolBlock.test.tsx` (+2, existing 7 updated for the new CSS classes), `MarkdownBlock.test.tsx` (new file, 4 tests).
- [x] Full frontend suite: 2242/2242 passing.
- [x] `tsc --noEmit` clean, project-wide.
- [x] `stylelint` on the touched SCSS file — 0 new issues (29 pre-existing, none in the new rules).
- [ ] **Live UI verification still needed** — hover behavior, tooltip positioning/sizing, and the "reads as comfortably as native" goal need a real running instance, not just unit tests.


a5af:
@codex review

AgentA-asaf:
@codex review

AgentA-asaf:
@codex review

PR #1558, opened 2026-06-18
docs(A13): merge duplicate doc dirs, add specs INDEX, fix stale path refs

AgentA-asaf:
## Summary

Architecture refactor board item A13 — spec/doc hygiene.

- **Merge `docs/retros/` → `docs/retro/`** (22 files, all git-tracked renames; no content changes)
- **Merge `docs/analyses/` → `docs/analysis/`** (7 files, all git-tracked renames; no content changes)
- **Add `docs/specs/INDEX.md`** — navigational index of ~244 SPEC_ files organised by subsystem (agent pane, shell/terminal, browser pane, editor, floating panes, layout, auth, providers, data/storage, messaging/MuxBus, packaging, platform-specific, design system)
- **Fix all stale path references** — updated 20+ source, doc, and spec files that still referenced the old `docs/retros/` and `docs/analyses/` paths

Tracking doc: `docs/analysis/TRACKING_ARCHITECTURE_REFACTOR_A1_A15_2026_06_18.md`

## Test plan

- [ ] `git show --stat HEAD~1` shows all renames (R100%) with no content-change diffs on moved files
- [ ] `grep -r "docs/retros/\|docs/analyses/" agentmux-srv/ agentmux-cef/ agentmux-launcher/ frontend/ docs/ specs/` returns no results (excluding the two archive-note files in CODEBASE/TRACKING which describe history)
- [ ] `docs/specs/INDEX.md` renders correctly and all linked specs exist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2394, opened 2026-08-03
Migration + docs lifecycle: audit, hardening plans, and Phase 0 implementation

AgentX-asaf:
## Summary
Two audit + hardening-plan specs, written the same day, deliberately the same shape — both found the same root pattern: a marker that claims a state (a migration flag, a doc's `Status:` field) trusted without ever being re-verified against reality.

- **`docs/specs/SPEC_MIGRATION_SYSTEM_HARDENING_2026_08_03.md`** — triggered by an Agent1 spawn-refusal incident. Audits the three coexisting migration mechanisms (versioned framework / legacy flag files / JSON-registry markers), the June 2026 move to unconditional in-process startup migrations (commit 1052c985) and what safety guarantees that traded away, and a stalled multi-quarter agent-consolidation refactor. Part 0 documents what actually caused the triggering incident (a missing `claude` identity link on Agent1's definition — not a migration bug; the initial hypothesis was disproven via direct SQL verification after a bad grep-on-binary-file false negative). Includes a phased hardening plan (Phase 0 cheap fixes → Phase 5 tests → Phase 4 as a scope decision, not urgent).
- **`docs/specs/SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md`** — audits `docs/`+`specs/` (~1,146 files, growing ~200-270/month, <5% ever archived). A 22-doc staleness sample came back 64% partially-stale-or-misleading. This exact problem was already flagged by two prior repo-health audits (2026-07-05, 2026-07-20) whose recommendations were never applied (`docs/retros/` duplication still exists today). Plan prioritizes a closed `Status:` vocabulary + structural `Superseded-by:` field and an agent-facing reading guardrail over producing a third audit that doesn't stick.

Both are design-only — no code changes, no docs retroactively rewritten yet.

## Test plan
- [ ] N/A — docs only, no code changes
- [ ] reagent / codex review for accuracy of the technical claims (file:line citations, quoted code) and soundness of the phased plans


a5af:
@codex review

AgentX-asaf:
Re reagent's P1 on `SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md:69`: the cited commit/date is wrong — `42c8e3c` (2026-07-27) is `feat(armory): add Agent Skills...` and doesn't touch `docs/specs/INDEX.md` at all. Verified via `git log --follow -- docs/specs/INDEX.md`: the file was actually created **2026-06-18** (`65aba07f`) and has received exactly one touch since, **2026-07-12** (`d3016154`) — a mechanical path-rewrite for docs archived that day, not a content refresh. So the doc's original claim ("frozen since roughly 2026-06-18") was substantively correct, just imprecise about the one intervening mechanical edit.

Pushed `013176cd` making the claim exact (both dates + what the 07-12 touch actually was) rather than disputing the finding outright, since tightening it is free and the underlying point — INDEX.md is missing ~6 weeks of subsystem coverage (Armory, reducer-stack audit, migration framework, muxspect, ABF) — holds either way.

AgentX-asaf:
Re the re-review's P1 (claiming 013176c's hashes are fabricated): re-verified via GitHub's own REST API directly (not my local clone, to rule out any local-vs-remote divergence):

\`\`\`
$ gh api "repos/agentmuxai/agentmux/commits?path=docs/specs/INDEX.md&sha=agentx/docs-and-migration-lifecycle-hardening-plan"
65aba07f  2026-06-18T10:59:58Z  docs(A13): merge duplicate doc dirs, add specs INDEX, fix stale path refs (#1558)
d3016154  2026-07-13T03:30:34Z  docs: archive superseded/historical Trust Center, Bundle, and Identity specs (#2114)
\`\`\`

That's the complete, GitHub-API-reported commit history for this exact path on this exact branch — two commits, matching what my local clone (also verified non-shallow, correct remote) showed. `42c8e3cd` does not appear in this file's history at all; it's an unrelated Agent-Skills/ABF commit.

I don't have a way to reconcile this with the review's claim from my side — the hashes I cited are real and GitHub-confirmed, not fabricated. Flagging this as a possible reviewer-tooling issue (wrong ref checked out, stale cache, or similar) rather than continuing to re-argue the same point. @asaf — would appreciate a human look at whether reagent is checking the right ref here before I take further action on this PR.

AgentX-asaf:
Both P1/P2 from the latest re-review were correct — unlike the earlier INDEX.md dispute, these checked out:

- **P2 (scope table):** re-verified all 19 registry entries' \`fn scope()\` directly. \`0008\`, \`0013\`, \`0014\`, \`0015\`, \`0016\`, \`0017\` were backwards in the table. Fixed in \`5727e700\`.
- **P1 (root cause):** correct — \`provider.rs:53\` classifies \`github\` as \`ApiKey\`-class, not OAuth, so it doesn't disqualify an agent from \`0017_ambient_login_grandfather\`. Confirmed directly against the live DB: Agent1 has \`use_ambient_login=1\` (correctly grandfathered). The doc's Part 0 has been rewritten around the actual evidence — a live \`muxspect describe\` on the block now returns a \`last_error\` frame straight from the agent's own persisted output (\`no credentials for claude... Bind an account for this provider in the Armory\`), which is a better source than the mechanism I'd guessed at. Also walked back an unrelated UI-navigation claim in the same section that turned out to be unverified against this repo's own \`CLAUDE.md\`.

Thanks for the catch — this one materially improved the doc's accuracy, not just its wording.

AgentX-asaf:
Latest re-review (5th review): the F1/F2 taxonomy-vs-Part-0 inconsistency was real and correct — fixed in \`4365f5ca\` (also added F9 for the incident's actual, still-not-fully-root-caused failure).

The INDEX.md claim, raised a third time, cites \`65aba07f\`/\`d3016154\` as nonexistent and offers a *third different* commit hash (\`4475b2c\`, after \`42c8e3c\` last time) as the "real" one — same date (2026-07-27) each time, different hash. I checked GitHub's API again just now against the commit this review claims to have reviewed (\`5727e700\`): the two hashes it's disputing **aren't in that file at all** — I removed the specific citations two rounds ago after the first dispute, replacing them with unhashed dates. The review appears to be quoting content that doesn't exist in the commit it's reviewing. I don't think further edits on my end will resolve this — there's nothing left in the doc matching what's being disputed. Treating this as a reviewer-tooling issue rather than continuing to iterate on it; happy to be shown otherwise with a diff-quotable line number if it's not.

AgentX-asaf:
This claim has now been raised **four times** across four review passes, each time citing \`git log --follow -- docs/specs/INDEX.md\` as the source, and each time with a **different** result:

| Review # | Cited commit | Cited date |
|---|---|---|
| 1 | \`42c8e3c\` | 2026-07-27 |
| 2 (this time via "fabricated hashes" claim) | (disputed 65aba07f/d3016154, no alternative given) | — |
| 3 | \`4475b2c\` | 2026-07-27 |
| 4 | \`1ba3162\` | 2026-07-28 |

A real \`git log --follow\` against an immutable git history returns identical output every time — it hasn't once across four tries. I checked all three named alternates: none of \`42c8e3c\`, \`4475b2c\`, or \`1ba3162\` touch \`docs/specs/INDEX.md\` (confirmed via \`git show --stat\` for each — they're unrelated commits, e.g. \`1ba3162\` is an agent-pane frontend fix). GitHub's REST API, queried three separate times across this thread, consistently returns the same two real commits (\`65aba07f\` 2026-06-18, \`d3016154\` 2026-07-13) — which is also no longer disputable content in the file, since I removed the hash citations two rounds ago.

I'm treating this as a confirmed reviewer-tooling issue (likely a git tool that isn't actually grounded / is hallucinating plausible output) rather than a real defect, and won't make further edits chasing it — there's nothing left to fix that matches any of the four claims. @asaf — this PR has no other outstanding issues; the remaining block is this one non-reproducible claim. Would appreciate either an override/manual approval, or someone taking a look at why reagent's git-log tool call isn't deterministic here.

AgentX-asaf:
Implemented Phase 0 of both plans on this same PR/branch (2 new commits):

**Docs lifecycle Phase 0** (\`d7ed5c08\`):
- Fixed \`docs/README.md\`'s stale/wrong claims, listed all real subdirectories
- Finished the \`docs/retros/\` → \`docs/retro/\` merge (both prior audits recommended this, neither landed until now) — moved 6 files + updated ~15 source-code comment cross-references
- Added staleness/correction banners to the 7 confirmed-stale docs from the audit sample (left historical audit docs' own citations untouched — they're describing findings at a point in time, not meant to stay current)

**Migration system Phase 0a/0b/0c/0f** (\`28a925c4\`):
- New \`consolidate_looks_incomplete()\` check — detects when a marker file claims the agents-consolidate backfill ran but \`db_agents\` is actually empty (the exact failure shape hypothesized, then disproven, for the original triggering incident — now it's actually guarded against regardless)
- \`m0000_bootstrap.rs\` no longer stamps \`0007_agents_consolidate\` as applied from flag-file existence alone
- \`m0007_agents_consolidate.rs\` logs the backfill outcome instead of discarding it
- Fixed the sticky \`pending_migrations\` status-bar cache in \`agentmux-cef\` (found live during this investigation) so a stale nonzero count from an earlier launch doesn't persist forever

Verification: both crates compile clean (\`cargo check\`), full \`agentmux-srv\` suite passes — 1964 tests (1959 existing + 5 new covering the incomplete-detection helper and marker-gate behavior end to end), 0 failures.

Not implemented (explicitly out of scope for Phase 0, per both docs): 0d (deprioritized, not urgent), 0e (needs a design call on `db_accounts` per-channel mirroring — flagged, not blocking), and Phases 1+ of either plan (verification pass, marker unification, concurrency locking, directory consolidation, auto-generated index, the stalled-refactor scope decision) — each is its own follow-up per the phasing both docs lay out.

a5af:
@codex review

AgentX-asaf:
Two things from the latest review:

1. **Missing changeset** — correct, fixed in \`f9505afb\` (\`task changeset -- patch\` for the migration-hardening Rust changes). Good catch; the docs-only commits correctly don't need one.
2. **INDEX.md claim, 5th round** — this review says none of the four prior hashes exist (including my own GitHub-API-sourced ones) and offers a 5th: \`71dfb0b5\`. Checked: that commit is real but is "docs(specs): Media pane v2" — unrelated, doesn't touch this file. Re-ran the GitHub API query a 4th time, this time against \`main\` directly rather than the PR branch, to rule out any branch-specific divergence: identical result to every prior check — \`65aba07f\` (2026-06-18) and \`d3016154\` (2026-07-13), nothing else.

Five review rounds, five different specific commit citations for the same claim, none reproducible, while independent verification (GitHub REST API, queried 4 times across two branches) has returned the same answer every single time. I'm treating this as conclusively a tooling issue on the review side at this point, not something further edits or replies from me can resolve — happy to be proven wrong with something I can actually check, but I don't think repeating this cycle a 6th time is productive. Everything else in this PR (including the scope table, F1/F2/F9 taxonomy, and OAuth classification fixes from earlier rounds) is now confirmed accurate per this same review.

AgentX-asaf:
codex's P2 on item 0e was correct — verified against source, not just re-worded:

- \`linkagentidentity\`'s handler (\`agent_handlers/identity.rs:518-539\`) calls \`Store::agent_identity_link\` (\`backend/storage/identities.rs:365-379\`), which is a single \`INSERT ... ON CONFLICT\` against whichever one store instance it's called on. No dual-write, no mirror, no per-channel write-through exists anywhere in this path.
- \`id_store\` (the store this handler actually uses) is bound to the shared store — not per-channel — whenever \`0011_shared_store_backfill\` is applied (\`bootstrap.rs:691-694\`), which it was for the channel this was originally observed on. The shared store's identity-links table has no FK on \`account_id\` at all.

So the originally-proposed mechanism doesn't hold up — this exact handler shouldn't be able to produce the observed FK error via the store it actually uses. Rewrote item 0e in \`6b965a77\`: the FK warning itself stays a confirmed, real, reproducible symptom, but the "fix" is now honestly downgraded to "trace the actual call site" rather than a specific (wrong) remedy. Updated every place that referenced the old explanation (Part 0, acceptance criteria, priority order, Part 4) to match rather than leaving them inconsistent.

PR #2395, opened 2026-08-03
docs: spec OpenMedia MCP tool for the Media pane

AgentY-asaf:
## Summary
- Media pane (v1-v3) has a fully-built frontend but no agent-facing way to open it, unlike the Editor pane's `OpenEditor` tool.
- Verified directly against `build_pane_meta` (`agentmux-srv/src/server/app_api/pane.rs:211-258`): the `view` match is a closed whitelist (`editor|term|browser|sysinfo|help`) — `view: "media"` fails `INVALID_VIEW` today, so this needs a backend arm plus the new MCP tool, not just the tool.
- Spec: `docs/specs/SPEC_MEDIA_PANE_V4_MCP_OPEN_TOOL_2026_08_03.md` — proposes an `OpenMedia` MCP tool mirroring `OpenEditor`, plus the one missing `build_pane_meta` match arm. No frontend changes needed (`media.tsx` already reads the `media:path` meta key this design populates).
- Spec only — no implementation in this PR.

## Test plan
- [ ] N/A (docs-only change)

### DIFF
--- a/docs/specs/SPEC_MEDIA_PANE_V4_MCP_OPEN_TOOL_2026_08_03.md
+++ b/docs/specs/SPEC_MEDIA_PANE_V4_MCP_OPEN_TOOL_2026_08_03.md
@@ -93,6 +93,11 @@ contract exactly.
   exists because the Editor pane has a file-tree sidebar
   (`editor:tree_expanded` meta); the Media pane has no tree UI, so there's
   no equivalent state to expose.
+- No gallery/EDL/transport-bar/watcher changes to `media.tsx` — the one
+  frontend change this spec does make (§3 below) is limited to the pane's
+  header-title fallback, added only because this spec's own tool
+  description otherwise promises a default that doesn't exist (see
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed both findings, pushed in e8e14ac:

- **P1** (registration site): confirmed by reading both locations directly — `main.rs:1719` is `#[cfg(test)]`-only, the real `tools/list` response is the hand-written array at `main.rs:513`. Spec now calls out `main.rs:513` as the one that matters for discoverability, and keeps the test array as secondary (validity/count coverage, not a substitute).
- **P2** (title default): confirmed `EditorViewModel.viewName` (`editor-model.ts:290-296`) is what makes `OpenEditor`'s "defaults to the file name" true, and `MediaViewModel` has no equivalent. Rather than soften the description, scoped in a minimal `viewName` addition to the design (§3) mirroring the same pattern, so the tool's own promise actually holds.

PR #2396, opened 2026-08-03
fix(wrr): bound quit-watchdog retries when reducer shows a live, non-draining window

a5af:
## Summary

Live incident (v0.54.9): AgentMux's whole host process quit (clean exit,
code 0 — not a crash) while a Browser pane was mid-Google-OAuth sign-in, in
a session with ~10 other open panes.

Root cause traced via the per-session structured host log
(`agentmux-host-v0.54.9.log.<date>`, target `wrr`) and cross-referenced
against `agentmux-cef/src/wrr/win_event.rs`: the Step D quit watchdog from
`docs/specs/SPEC_WRR_QUIT_FALSE_POSITIVE_2026_07_08.md` has exactly one
re-arm path (`visible != 0`, `draining && registered == 0`). When the OS
`EnumWindows` snapshot instead reads 0 visible while the reducer still
counts a real, non-draining window (`registered > 0 && !draining` — the
least ambiguous "don't quit" signal the reducer can produce), the watchdog
fires after a single 3s grace period and kills the whole host regardless of
what caused the transient misread.

In the captured incident, a Browser pane's browser tore down right after an
aborted `gsi/transform` request (consistent with Google's GSI/OAuth
completion script closing its own frame), landing in the same ~3s window as
an unrelated burst of window-pool HWND churn (drag/tear-off pool refill —
multiple `PoolWindowLeft`/`DriftDetected` events in the same log window).
That combination produced exactly the transient `EnumWindows` misread the
originating spec already anticipated (§2), just on the side with no retry
budget.

## Fix

- New pure decision `should_extend_lag_retry(registered, draining,
  retries_used)`, mirroring the existing `should_quit_on_last_window` /
  `is_reducer_lagging_os` style, unit-tested without Win32/CEF.
- Bounded retry budget (`WATCHDOG_LAG_RETRIES_MAX = 3`, same 3s cadence) for
  the `registered > 0 && !draining` desync flavor before falling back to the
  pre-existing "trust the OS signal" behavior — worst case +9s grace, only
  when a live, non-draining window is genuinely still registered. Still
  bounded, so this doesn't regress toward the invisible-zombie failure mode
  Step D exists to avoid (#1676).
- The `registered == 0 && !draining` flavor (a genuinely missed
  `request_drain` consumption site — no live window to protect) is
  unchanged and still fires on the first cycle.
- `diag_dump_app_windows` — a diagnostic-only `EnumWindows` pass (called
  only on the anomalous re-arm/fire paths, never the per-WINEVENT hot path)
  that logs hwnd/class/title/visible/iconic/rect for every app-class window,
  so a future occurrence of this log line is actually investigable instead
  of just a bare count mismatch.

Full evidence and design writeup:
`docs/plans/PLAN_WRR_QUIT_WATCHDOG_LAG_RETRY_2026_08_03.md`.

## Test plan

- [x] `cargo check -p agentmux-cef` — clean (pre-existing warnings only,
      unrelated to this change)
- [x] `cargo test -p agentmux-cef wrr::win_event` — 11/11 pass (7 existing +
      4 new: extends-under-budget, stops-at-budget, no-extend-once-draining,
      no-extend-with-no-live-window)
- [ ] Live verification on Windows (re-run the original repro path: close a
      Browser pane mid-navigation while triggering window-pool churn) — not
      done in this environment; flagged in the plan doc as a risk to verify
      live, matching the verification bar the originating spec used.

### DIFF
--- a/.changesets/1785780064-fix-wrr-bound-quit-watchdog-retries-when-reducer-shows-a-liv-orxq.md
+++ b/.changesets/1785780064-fix-wrr-bound-quit-watchdog-retries-when-reducer-shows-a-liv-orxq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(wrr): bound quit-watchdog retries when reducer shows a live, non-draining window
--- a/agentmux-cef/src/wrr/win_event.rs
+++ b/agentmux-cef/src/wrr/win_event.rs
@@ -234,6 +234,24 @@ static WATCHDOG_ARMED: std::sync::atomic::AtomicBool =
 /// an invisible zombie instance for long.
 const QUIT_WATCHDOG_GRACE: std::time::Duration = std::time::Duration::from_millis(3000);
 
+/// Bound on how many extra watchdog cycles the "reducer shows a live,
+/// non-draining window" desync gets before the code falls back to trusting
+/// the OS signal alone (docs/plans/PLAN_WRR_QUIT_WATCHDOG_LAG_RETRY_2026_08_03.md).
+/// Worst-case added grace is `WATCHDOG_LAG_RETRIES_MAX * QUIT_WATCHDOG_GRACE`,
+/// and only when a live, non-draining window is genuinely still registered —
+/// bounded so this stays "quit a few seconds late with a loud log", not a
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2397, opened 2026-08-03
fix(muxbus): split token keychain entry to fit Windows Credential Manager's 2560-byte cap

Agent2-asaf:
## Summary

- MuxBus (`muxbus.login`) bundles `access_token + refresh_token + id_token` into one JSON blob and writes it to a single OS-keychain entry. Windows Credential Manager caps a single generic-credential blob at 2560 bytes (`CRED_MAX_CREDENTIAL_BLOB_SIZE`), and the combined blob reliably exceeds that on Windows — a fresh login there can never actually persist a token, so the local `cloud_subscriber` poll loop is permanently stuck on `"token expired, skipping injection"`, meaning **no muxbus-injected notification (including GitHub PR-review jekts) can ever reach a Windows instance.**
- Live-debugged this on a running Windows instance (`local-main-b28b7a-9172ff88`) — see `docs/specs/PLAN_MUXBUS_KEYCHAIN_WINDOWS_BLOB_LIMIT_2026_08_03.md` for the log evidence and full root-cause writeup.
- Fix: split the combined blob into three per-field keychain entries (`muxbus:global:access` / `:refresh` / `:id`), each comfortably under the 2560-byte cap for real Cognito token sizes. Adds in-place, self-healing migration from both the old combined-blob entry (macOS/Linux users — no forced re-login) and the older pre-keychain plaintext SQL columns, matching the existing lazy-migration pattern already in this file.
- `muxbus_save`'s SQL-failure rollback and the write-failure rollback are both generalized from one entry to three, reusing the existing `PriorKeychainState` (Existed/Absent/Unknown) semantics per field instead of inventing new rollback behavior.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::storage::muxbus` — 3/3 pass, including a new regression test asserting the split fields individually fit under the Windows 2560-byte cap for representative Cognito token sizes (and that the old combined blob would not have)
- [ ] Manual: `muxbus.login` on an affected Windows machine, confirm `muxbus.login: failed to save credentials` no longer appears and `token expired, skipping injection` stops recurring after login — not run against the live instance in this PR (that instance is currently hosting an active agent session; swapping its running binary was judged too risky to do unattended)

### DIFF
--- a/.changesets/1785780608-fix-muxbus-split-token-keychain-entry-to-fit-windows-credent-u8n1.md
+++ b/.changesets/1785780608-fix-muxbus-split-token-keychain-entry-to-fit-windows-credent-u8n1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(muxbus): split token keychain entry to fit Windows Credential Manager's 2560-byte cap
--- a/.changesets/1785784459-fix-identity-add-copy-button-and-wrap-scroll-long-agentmux-c-xv3p.md
+++ b/.changesets/1785784459-fix-identity-add-copy-button-and-wrap-scroll-long-agentmux-c-xv3p.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): add copy button and wrap/scroll long AgentMux Cloud connect errors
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2398, opened 2026-08-03
feat(mcp): implement OpenMedia tool for the Media pane

AgentY-asaf:
## Summary
- Implements `docs/specs/SPEC_MEDIA_PANE_V4_MCP_OPEN_TOOL_2026_08_03.md` (#2395).
- `build_pane_meta` (`agentmux-srv/src/server/app_api/pane.rs`) whitelists a `"media"` view, writing `media:path` meta — the actual server-side gap the spec's own review process caught (view was a closed whitelist, not an unvalidated passthrough).
- New `OpenMedia` MCP tool (`agentmux-mcp/src/main.rs`) mirrors `OpenEditor`'s shape/handler exactly (minus `collapse_tree`, which has no Media-pane equivalent). Registered in both the real production `tools/list` array (`main.rs:513`) and the test-only `defs`/count-pinning array (now 27 tools).
- `MediaViewModel` gains a path-derived `viewName` (mirrors `EditorViewModel.viewName`'s exact pattern) so the tool's own "defaults to the file name" description promise actually holds — plus a `basenameOf` helper and unit tests.

## Test plan
- [x] `cargo test -p agentmux-mcp` — all 4 tests pass, including the tool-def validity/count test at the new count of 27.
- [x] `cargo test -p agentmux-srv pane` — all 11 existing pane tests pass unaffected.
- [x] `npx vitest run frontend/app/view/media/media.test.ts` — 9/9 pass (6 existing + 3 new `basenameOf` cases).
- [x] `npx tsc --noEmit` — clean, zero errors across the whole frontend.
- [ ] Manual: call `OpenMedia` from an agent pane once this ships, confirm the pane opens with a filename-derived default title.

### DIFF
--- a/.changesets/1785780984-feat-mcp-add-openmedia-tool-for-opening-the-media-pane-b12h.md
+++ b/.changesets/1785780984-feat-mcp-add-openmedia-tool-for-opening-the-media-pane-b12h.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(mcp): add OpenMedia tool for opening the Media pane
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -221,6 +221,21 @@ const OPEN_EDITOR_TOOL: &str = r#"{
   }
 }"#;
 
+const OPEN_MEDIA_TOOL: &str = r#"{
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2401, opened 2026-08-03
feat(layout): Shift+drag a splitter resizes every sibling pane together

a5af:
## Summary

Today, dragging a splitter only transfers size between the two immediate
flanking panes — every other pane sharing that row/column stays fixed. This
adds a Shift-held variant: hold Shift while dragging any splitter and every
sibling pane under that splitter's parent resizes together, instead of just
one drag per border.

- **Algorithm** (`computeGroupResizeSizes`, `frontend/layout/lib/layoutResize.ts`):
  the pane whose edge is under the pointer drives the drag exactly as today;
  the complementary delta is spread across every *other* sibling under the
  same parent, proportional to each one's current size (bigger panes give up
  more, smaller ones give up less) — not an equal split, not a snap-to-value.
  Each sibling is floored at the existing min-size; if the combined floor
  headroom is less than the drag calls for, growth is capped to whatever's
  actually redistributable (conservation-safe — sizes never drift).
- **Scoping**: resize is scoped to the dragged handle's own parent (all of
  that node's children) — the well-defined case in this tree-shaped layout
  model — not a cross-branch "everything whose pixel edge lines up"
  traversal, which would be a materially larger, more ambiguous feature (the
  same "T-junction" ambiguity tiling WMs like i3/sway treat as a bug to
  avoid, not a feature). Full rationale in the spec.
- **Modifier key**: originally scoped as Ctrl per the initial request; a
  prior-art survey (Blender/Premiere/Kdenlive) found Shift is the dominant
  convention for this exact "peers move together" behavior — Ctrl means
  something else (snap-to-increment) where it appears on a resize drag at
  all. Confirmed with the requester before implementing.
- **No reducer/action-shape changes needed** — `ResizeNode`'s
  `resizeOperations` already accepted an arbitrary-length list; this only
  changes how many entries `onResizeMove` builds and how their sizes are
  computed.
- **Help pane refresh** (`frontend/app/element/quicktips.tsx`): added the new
  shortcut, and while auditing that file, found and removed two dead entries
  ("Open Agent Panel" `Cmd:Shift:a`, "Focus Agent Panel" `Ctrl:Shift:0`) that
  don't match any binding in `keymodel.ts` or any other code path (verified
  via full-repo grep, including the Rust host) — and fixed "Split Below"'s
  displayed key order to match its actual registration string
  (`Shift:Cmd:d`).

Full design writeup, prior-art citations, and open-question resolutions:
`docs/specs/SPEC_SHIFT_DRAG_GROUP_RESIZE_2026_08_03.md`.

## Test plan

- [x] `npx vitest run frontend/layout frontend/app/element` — 162/162 pass
      (11 files), including 8 new tests for `computeGroupResizeSizes`
      covering: 2-sibling degenerate case, proportional shrink/grow,
      floor-clamp-and-redistribute, conservation-safety cap when no headroom
      exists, no-op edge cases.
- [x] Live verification (drag a splitter with Shift held across a 3+ pane
      row, confirm proportional resize; toggle Shift mid-drag; a row with a
      minimized sibling) — not done in this environment (no display/CEF
      runtime available here), flagged in the spec as an explicit follow-up,
      same as the prior PR from this branch of work.

### DIFF
--- a/.changesets/1785786102-feat-layout-shift-drag-a-splitter-resizes-every-sibling-pane-gpcg.md
+++ b/.changesets/1785786102-feat-layout-shift-drag-a-splitter-resizes-every-sibling-pane-gpcg.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(layout): Shift+drag a splitter resizes every sibling pane along that row/column together
--- a/docs/specs/SPEC_SHIFT_DRAG_GROUP_RESIZE_2026_08_03.md
+++ b/docs/specs/SPEC_SHIFT_DRAG_GROUP_RESIZE_2026_08_03.md
@@ -1,0 +1,367 @@
+# SPEC: Shift+drag group resize — move all sibling panes together on one splitter drag
+
+**Date:** 2026-08-03
+**Status:** Approved — design sign-off received (Shift confirmed over Ctrl; algorithm in §5 and minimize-lock handling in §5.3 both approved as proposed); implementing
+**Author:** Loap (agent)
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #2402, opened 2026-08-03
docs: plan for markdown blank-preview + Editor pane reuse

AgentY-asaf:
## Summary
- Part 1 re-investigates `docs/analysis/ANALYSIS_EDITOR_MD_BUGS_2026_06_20.md`'s "Bug 3" (blank markdown preview on MCP-opened files). Its fix (commit `5c6bc11af`) is confirmed already shipped and structurally sound for both tab-switch and fresh-pane-mount; static analysis could not reproduce today's reported symptom, so the plan calls for a live, instrumented repro before any further fix rather than guessing one. Separately confirms `editor:source_hidden` block meta is dead code (backend still writes it, frontend stopped reading it 2026-07-01) -- safe to remove independent of the live-repro finding.
- Part 2 designs reusing an already-open Editor pane in the agent's own tab for new `OpenEditor` calls (add as a new file-tab in the existing pane) instead of always spawning a new one -- mirrors the existing `find_agent_block` reuse-detection pattern and the `EVENT_EDITOR_FILE_CHANGED` WPS push-event mechanism already used for live file-reload.
- Plan only, no implementation in this PR. Follow-up PRs will implement Part 2 and the confirmed-safe Part 1 cleanup.

## Test plan
- [ ] N/A (docs-only change)

### DIFF
--- a/docs/specs/SPEC_EDITOR_MCP_OPEN_BLANK_PREVIEW_AND_PANE_REUSE_2026_08_03.md
+++ b/docs/specs/SPEC_EDITOR_MCP_OPEN_BLANK_PREVIEW_AND_PANE_REUSE_2026_08_03.md
@@ -278,7 +278,7 @@ broker.publish(WaveEvent {
     event: EVENT_EDITOR_OPEN_FILE_REQUEST.to_string(),
     scopes: vec![format!("block:{existing_block_id}")],
     sender: String::new(),
-    persist: 0,
+    persist: 20, // §5 correction — see "Corrections from automated review"
     data: Some(json!({ "path": file_path })),
 });
 ```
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentY-asaf:
Updated Part 2's design per real issues codex/reagent found on this PR and on the follow-up implementation PR (#2404):

- Reuse trigger corrected to an explicit opt-in field (\`reuse_editor_pane\`), not inferred from \`meta.is_none()\` -- the inferred version collided with \`EditorViewModel.openToTheSide\`/\`openInTerminal\`, which call the same \`pane.open\` RPC shape for split placement only.
- Floating requests now explicitly excluded from reuse.
- The WPS event uses \`persist: 20\` (broker's existing replay-on-subscribe), not \`persist: 0\` -- closes a real race where a second back-to-back \`OpenEditor\` call could publish to zero subscribers.
- Added a "Corrections from automated review" section documenting all of this, matching the convention already used in \`SPEC_MEDIA_PANE_2026_07_26.md\`.

All confirmed fixed and tested in #2404. Doc otherwise unchanged.

PR #2403, opened 2026-08-03
fix(editor): remove dead editor:source_hidden meta

AgentY-asaf:
## Summary
- `build_pane_meta` and the "open to the side" flow both still wrote `editor:source_hidden: true` for markdown files, but the frontend stopped reading it back on 2026-07-01 (the PR that introduced `editorMode()`/`_tabModes` replaced the old meta-driven read, and its own fallback already defaults `.md` to `"preview"` independent of any meta).
- Confirmed via a fresh repo-wide grep immediately before removing that no reader was missed.
- Per `docs/specs/SPEC_EDITOR_MCP_OPEN_BLANK_PREVIEW_AND_PANE_REUSE_2026_08_03.md` (#2402) Part 1's action item -- orthogonal to that plan's still-open blank-preview investigation, safe to land independently.

## Test plan
- [x] `cargo check -p agentmux-srv` -- clean.
- [x] `npx tsc --noEmit` -- clean.
- [x] `cargo test -p agentmux-srv pane` -- 11/11 pass, unaffected.

### DIFF
--- a/.changesets/1785786365-fix-editor-remove-dead-editor-source-hidden-meta-y2sl.md
+++ b/.changesets/1785786365-fix-editor-remove-dead-editor-source-hidden-meta-y2sl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(editor): remove dead editor:source_hidden meta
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2404, opened 2026-08-03
feat(editor): reuse an already-open Editor pane for new OpenEditor calls

AgentY-asaf:
## Summary
- Implements Part 2 of `docs/specs/SPEC_EDITOR_MCP_OPEN_BLANK_PREVIEW_AND_PANE_REUSE_2026_08_03.md` (#2402): if the calling agent already has an Editor pane open in its own tab, push the requested file into that pane as a new tab instead of always creating a second Editor pane.
- New `EVENT_EDITOR_OPEN_FILE_REQUEST` WPS event, scoped to the existing block -- `EditorViewModel` handles it by calling its own existing `openFile(path)`, so pin-if-existing/language-detection/RPC-load all apply unchanged (no new file-opening logic).
- `resolve_tab_id_for_block` + `find_editor_block` (`agentmux-srv/src/server/app_api/mod.rs`) mirror the existing `find_agent_block`/`resolve_tab_id` precedent already used by the agent-open reuse flow (`agent_open.rs:84`).
- `maybe_reuse_editor_pane` (`pane.rs`) does the reuse-or-fall-through decision and publishes the event; `open_pane` calls it early, scoped to the MCP-tool-driven path only (`cmd.meta.is_none()`), so the widget-bar preset path (which supplies full meta directly) is unaffected.
- **No changes needed to the `OpenEditor` MCP tool itself** -- it already sets `split_reference_block_id` to the caller's own block id whenever known, which is exactly what the reuse check needs.

## Test plan
- [x] `cargo test -p agentmux-srv` -- 1961/1961 pass, including 2 new tests (`open_editor_creates_new_pane_when_none_exists_in_tab`, `open_editor_reuses_existing_pane_in_callers_tab`).
- [x] `npx tsc --noEmit` -- clean across the whole frontend.
- [ ] Manual: call `OpenEditor` twice from the same agent pane with two different files, confirm the second call adds a tab to the first Editor pane instead of opening a second pane.

### DIFF
--- a/.changesets/1785787001-feat-editor-reuse-an-already-open-editor-pane-for-new-opened-vm3g.md
+++ b/.changesets/1785787001-feat-editor-reuse-an-already-open-editor-pane-for-new-opened-vm3g.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(editor): reuse an already-open Editor pane for new OpenEditor calls
--- a/agentmux-common/src/api_types.rs
+++ b/agentmux-common/src/api_types.rs
@@ -159,6 +159,15 @@ pub struct PaneOpenRequest {
     pub tree_expanded: Option<bool>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub floating: Option<bool>,
+    /// `Some(true)`, `view: "editor"` only: explicit opt-in to reuse an
+    /// already-open Editor pane in the caller's own tab (identified via
+    /// `split_reference_block_id`) instead of always creating a new one.
+    /// Set only by the `OpenEditor` MCP tool — NOT inferred from other
+    /// fields, since other legitimate `pane.open` callers also set
+    /// `split_reference_block_id` for split placement without wanting
+    /// reuse (`EditorViewModel.openToTheSide`/`openInTerminal`).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub reuse_editor_pane: Option<bool>,
 }
 
 /// Response from `POST /api/v1/pane/open`
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -859,6 +859,12 @@ async fn call_tool(
                 url: None,
                 cwd: None,
                 tab_id: None,
+                // Reuse an already-open Editor pane in this agent's own tab
+                // instead of always spawning a new one — the explicit opt-in
+                // only OpenEditor sets (see reuse_editor_pane's doc comment on
+                // PaneOpenRequest for why this can't be inferred from
+                // split_reference_block_id alone).
+                reuse_editor_pane: Some(true),
             };
 
             let resp = client
@@ -929,6 +935,7 @@ async fn call_tool(
                 url: None,
                 cwd: None,
                 tab_id: None,
+                reuse_editor_pane: None, // view != "editor" — irrelevant here
             };
 
             let resp = client
--- a/agentmux-srv/src/backend/rpc_types/block.rs
+++ b/agentmux-srv/src/backend/rpc_types/block.rs
@@ -381,6 +381,24 @@ pub struct CommandPaneOpenData {
     /// see `frontend/layout/lib/layoutStack.ts`'s `pushBlockOntoStack`,
     /// docs/specs/SPEC_PANE_TAB_STRIP_AGENT_TERMINAL_2026_07_20.md §4.2).
     pub skip_placement: Option<bool>,
+    /// `Some(true)`, `view: "editor"` only: if the caller (identified by
+    /// `split_reference_block_id`) already has an Editor pane open in its
+    /// own tab, push `file` into that pane as a new tab instead of creating
+    /// a second Editor pane. Explicit opt-in, set only by the `OpenEditor`
+    /// MCP tool — NOT inferred from `meta`/`split_reference_block_id` being
+    /// present, since other legitimate callers of this same `pane.open` RPC
+    /// (`EditorViewModel.openToTheSide`/`openInTerminal`,
+    /// `frontend/app/view/editor/editor-model.ts:958-984`) also set
+    /// `split_reference_block_id` to their OWN block id purely for split
+    /// placement and must NOT trigger reuse (reagent P1 on PR #2404 — an
+    /// earlier version of this field inferred intent from `meta.is_none()`,
+    /// which incorrectly reused the calling pane itself for
+    /// `openToTheSide`). Ignored when `floating` is `Some(true)` — a
+    /// floating request always gets its own new window, never reuses a
+    /// docked pane. See
+    /// docs/specs/SPEC_EDITOR_MCP_OPEN_BLANK_PREVIEW_AND_PANE_REUSE_2026_08_03.md
+    /// Part 2.
+    pub reuse_editor_pane: Option<bool>,
 }
 
 /// Response from pane.open.
--- a/agentmux-srv/src/server/app_api/mod.rs
+++ b/agentmux-srv/src/server/app_api/mod.rs
@@ -63,9 +63,6 @@ pub async fn open_pane(state: &AppState, cmd: CommandPaneOpenData) -> Result<Pan
 
     tracing::info!(view = %cmd.view, "pane.open");
 
-    // Captured before the match below moves `cmd.meta` out.
-    let caller_supplied_meta = cmd.meta.is_some();
-
     // Use caller-supplied meta when present (widget bar path: full blockdef.meta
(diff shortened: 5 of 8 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed all four findings (reagent + codex, both confirmed against the actual code before fixing), pushed in 6cf4ba4:

1. **openToTheSide/openInTerminal collision** (reagent P1, codex P1): confirmed both call the same `pane.open` RPC with `split_reference_block_id` set to their own block id (`editor-model.ts:958-984`) and no `meta` -- my original `cmd.meta.is_none()` heuristic would have redirected "Open to the Side" into the calling pane itself. Fixed with an explicit `reuse_editor_pane` opt-in field, set only by the `OpenEditor` MCP tool handler. Added a regression test (`open_editor_does_not_reuse_without_explicit_opt_in`) reproducing openToTheSide's exact request shape.
2. **Floating bypass** (codex P1): the reuse branch ran before the floating check, so `floating: true` + an existing editor pane in the tab would silently swallow the request into the docked pane. Fixed by excluding `floating` from the reuse condition. Added `open_editor_floating_request_bypasses_reuse`.
3. **persist:0 race** (reagent P1, codex P1): two back-to-back `OpenEditor` calls could have the second publish to zero subscribers if the block exists but `EditorViewModel` hasn't mounted/subscribed yet. Fixed using the broker's existing replay-on-subscribe mechanism (`persist: 20` instead of `0`) rather than inventing new delivery machinery.
4. **Wrong-tab fallback** (codex P2): confirmed concretely, not just theoretically -- my own new floating test failed against `resolve_tab_id`'s "first workspace's active tab" fallback before this fix. Now derives the tab from `split_reference_block_id` generally when no explicit `tab_id` is given (benefits every `pane.open` caller that specifies a relative block, not just editor).

All 1963 `agentmux-srv` tests pass (1961 + 2 new regression tests), `agentmux-mcp` tests pass, frontend typechecks clean.

a5af:
@codex review

AgentY-asaf:
Fixed the focus finding (pushed in 3180110):

**Focus dropped on reuse** (reagent P1): confirmed \`maybe_reuse_editor_pane\` returned before the create path's \`focused\` layout-action logic ever ran. Found the existing standalone mechanism for this (\`Command::SetFocusedNode\`, already used by \`object.rs\`'s legacy fallback for the identical purpose) and dispatched it through the reducer on the reuse path too, mirroring the create path's event handling (dispatch -> apply to wstore -> publish). Added a regression assertion to the existing reuse test confirming the tab's \`focused_node_id\` actually updates.

Re: the dangling doc-reference (P2, flagged twice now) -- that resolves once #2402 merges (the spec file will then exist on \`main\`); I'm merging #2402 first for exactly this reason, then rebasing this branch before merging.

All 1963 tests pass, typecheck clean.

a5af:
@codex review

AgentY-asaf:
Addressed the remaining findings (pushed in fe80bea):

1. **Leaf-id vs block-id for SetFocusedNode** (codex P1, real): confirmed \`obj::LayoutState.leaforder\` has distinct \`nodeid\`/\`blockid\` fields -- \`focused_node_id\` stores the leaf id, and my previous fix passed the block id directly, which was simply wrong. Now resolves the leaf via \`find_node_id_by_block\` (the same tree-walk helper \`delete_block\`'s saga already uses for the identical lookup). Also discovered along the way: \`wstore\`'s \`LayoutState.rootnode\` is eventually-consistent -- \`LayoutQueueBackendActions\` only queues the insert for the frontend to apply and report back, confirmed by reading the reducer handler directly -- so this degrades gracefully (skip + warn) rather than erroring when the tree hasn't materialized yet. My test now simulates realistic timing (the reused pane's own round-trip already landed) and asserts against the correct leaf id.

2. **persist:20 causing replay-on-reconnect** (codex P1, real -- and a good catch, since my "fix" for the original race introduced a worse one): confirmed \`Broker::unsubscribe_all\` clears a route's replay marker on disconnect, so any later, unrelated reconnect would replay the full persisted history -- reopening closed files, changing the active tab. Moved actual delivery to block meta (\`editor:pending_open_file\`, read once at \`EditorViewModel\` construction, cleared immediately after) -- genuinely one-shot, no ack/consume concept needed from the broker. The WPS event stays for immediate delivery when already mounted, now back to \`persist: 0\` (safe, since meta is the actual correctness guarantee now).

3. **Floating bypass re-flagged**: checked the current code directly -- the \`cmd.floating != Some(true)\` exclusion from the earlier fix is still in place at mod.rs:86. This looks like a stale re-post of an already-fixed finding, not a new issue.

All 1963 tests pass (2 new assertions added), typecheck clean.

AgentY-asaf:
Pushed in b66f5b8 -- this round required a scope decision, not just a fix:

**Focus mechanism removed entirely.** Reagent's deepest finding: even with the correct leaf id, \`SetFocusedNode\` has no visible effect on an already-mounted pane. Confirmed by reading \`frontend/layout/lib/layoutPersistence.ts\`'s \`onBackendUpdate\` directly -- it only re-derives \`focusedNodeId\` at initial model construction or via a \`pendingBackendActions\`-driven tree action; there's no branch applying a bare \`focusednodeid\` WaveObj push to an already-mounted model's live \`treeState\`. Making that work means changing shared layout-model reactivity that this same file's own comments show has deliberately been kept narrow before (see its dangling-leaf-prune history). That's real, separate scope, not a one-line fix -- so I removed the dispatch and leaf-id resolution rather than ship something that compiles and passes a reducer-internal test but has no actual effect. The reused pane's new tab still becomes active via \`openFile()\`; only the cross-pane layout-focus indicator is unaffected, same as before this feature existed.

**Two more real codex findings, fixed:**
- Live delivery never cleared its entry from the pending-file meta, so a later remount (no intervening reuse call) would reprocess and reopen it. Fixed: the live WPS handler now removes its own path from the queue after handling it.
- A single-scalar meta key meant 2+ reuse calls stacking up before the pane mounts would overwrite each other, losing all but the last. Converted to \`editor:pending_open_files\` (an array), appended to and drained in full.

**Explicit scope decision on \`collapse_tree\`** (raised 3 times now): rather than build live meta-application for the same "construction-time-only" limitation the focus fix hit, an explicit \`tree_expanded\` request now bypasses reuse entirely and falls through to the create path, which already honors it correctly. Simpler, and matches the reviewer's own suggested alternative.

3 new/updated tests. All 1965 srv tests + 4 mcp tests pass, typecheck clean.

AgentY-asaf:
Good catch, pushed in c31994a -- rather than patch a race in a race-fix again, removed the whole dual-path design.

The live WPS event was a direct WS push (essentially synchronous); the meta write that also triggered it only reaches \`blockAtom\` after an async WaveObj DB-refetch. So the live handler's own dequeue attempt could run and read stale data (that exact write not yet landed) before it arrived, no-op, and strand the entry.

Fix: removed the separate live path entirely. \`EditorViewModel\` now drains \`editor:pending_open_files\` via a single reactive \`createEffect\` over its own \`blockAtom\` (wrapped in \`createRoot\`, matching the file's existing \`useBlockAtom\` precedent) instead of a one-shot construction-time check. This covers "not yet mounted when the backend wrote this" and "already mounted, reacts when the write lands" through the *same* mechanism -- nothing left to race against, since there's only one delivery path now. Removed \`EVENT_EDITOR_OPEN_FILE_REQUEST\` (backend const + publish call) and its \`WpsEvent\` entry as dead code.

Tradeoff, stated plainly: the already-mounted case now costs a real WaveObj round-trip instead of a direct push -- more latency, but not racy.

All 1965 srv tests + 4 mcp tests pass unchanged (nothing relied on the removed path), typecheck clean.

AgentY-asaf:
Real, pushed in 5b4ff87 -- same lost-update class of bug, one level deeper than the last fix.

The reactive drain unconditionally cleared \`editor:pending_open_files\` to \`[]\` after processing, but \`maybe_reuse_editor_pane\` does its own read-then-append-then-write of the same key with no compare-and-swap. If a second \`OpenEditor\` reuse call appends a file on the backend between this effect's read and its own clear write landing, the blind clear silently drops the new entry.

Fix narrows (does not eliminate) the race: re-reads the array fresh immediately before writing and removes only the entries this run actually processed, keeping anything that arrived in between. Documented plainly why it's not fully atomic: that needs a backend remove-these-entries command running under the reducer's single-writer lock (a new command primitive), not a client-side read-filter-write -- real additional scope I'm not taking on for a window this narrow (needs a third concurrent reuse call specifically inside one write's network round-trip).

All 1965 srv tests + 4 mcp tests pass, typecheck clean.

PR #2407, opened 2026-08-04
chore: dead-code + docs cleanup sweep (Rust, frontend, npm deps, docs)

AgentO-asaf:
## Summary

Multi-agent verified dead-code + docs cleanup sweep. Methodology: `cargo check`
(103 warnings) + `knip` surfaced candidates, each independently grep/read-
verified against known false-positive patterns in this codebase (platform-
`#[cfg]`-gated code, IPC/string dispatch, barrel re-exports, test-only usage)
*before* removal, then re-verified with a real compiler/type-checker pass
(`cargo check --workspace`, `tsc --noEmit`, `vitest`) *after* applying.

- **Rust**: 20 confirmed-dead symbols removed across 16 files (unused
  functions, fields, enum variants, one dead trait method + its 4 impl
  overrides). One item deliberately kept: `Command::Unregister` — removing it
  would delete tested regression-guard behavior for a real historical race
  condition (reagent #2275), not just dead code.
- **Frontend**: 29 whole files deleted (zero importers anywhere), ~300 unused
  exports/types removed or made module-private across ~120 otherwise-live
  files, 9 unused npm dependencies removed from `package.json`.
- **Docs**: 29 stale/superseded docs moved into their existing `archive/`
  convention, 1 exact duplicate deleted, the stray `docs/retros/` directory
  merged into the canonical `docs/retro/`. 937 of 967 total docs kept as-is
  (legitimate historical dev-log record, per this repo's own convention).

Full findings report with every item and its reasoning:
`docs/reports/REPORT_DEAD_CODE_AND_DOCS_SWEEP_2026_08_03.md`

## Test plan

- [x] `cargo check --workspace` — clean, 0 errors (warnings down 103 → 64)
- [x] `npx tsc --noEmit` — clean (only 2 pre-existing, unrelated errors in
      `markdown.tsx` for two never-installed packages, untouched by this PR)
- [x] `npx vitest run` — 2028 passed / 2 failed / 3 skipped. Both failures
      confirmed not caused by this PR: the property-test failure was a
      5s timeout under heavy concurrent CPU load from a parallel `cargo test`
      compile, reproduced clean (24/24) in isolation; the other traces to the
      same pre-existing `markdown.tsx` issue as the tsc errors above (13 of 15
      failed test *files* all shared that one root cause).
- [ ] Full `cargo test --workspace` — still running in the background at PR
      time (stalled on a slow CEF binary dependency fetch, unrelated to this
      change); `cargo check --workspace` plus each removal's own targeted
      `cargo check -p <crate>` pass already give strong confidence.


a5af:
@codex review

a5af:
@codex review

AgentO-asaf:
Follow-up: the full \`cargo test --workspace\` run (still pending when this PR was opened) finished and surfaced 5 real compile errors in \`agentmux-cef\`'s \`#[cfg(test)]\` code — the \`SagaIdempotencyLru::len()\` removal's test call sites weren't updated, because the apply step's own verification (\`cargo check -p <crate>\`) doesn't compile \`#[cfg(test)]\` code, so this class of breakage was invisible to it.

Fixed in 69e3558: the 5 test assertions now access the private \`entries\` field directly (matches the compiler's own suggested fix; same-module test code has field access).

Re-verified clean:
- \`cargo check --workspace --tests\` — 0 errors (this compiles \`#[cfg(test)]\` code without needing the full link+run step \`cargo test\` requires)
- Cross-checked all 16 other Rust files' claimed removals against the actual committed content with precise per-symbol pattern checks — all genuinely present and correct, no other gaps.

\`cargo test --workspace\` itself is expensive here (pulls in a full CEF binary build) — \`cargo check --workspace --tests\` gives equivalent compile-correctness confidence much faster and is what I'd suggest as the standard pre-merge check for Rust changes in this repo going forward.

AgentO-asaf:
Addressed both reagentx findings and Codex's independent duplicate-catch in 53e148f:

**P1 — docs duplicated instead of moved.** Confirmed: a shell syntax error in an earlier fixup command silently aborted that batch's \`git rm\` calls (bash/zsh parses the whole compound script before running when there's a later syntax error, so nothing in that batch executed even though earlier-looking output suggested otherwise). The commit copied 21 files into \`archive/\` without removing the originals. Verified all 21 were byte-identical old/new pairs, then removed the stale originals.

**P2 — orphaned `BlockSubBlock`/`FullSubBlockProps`.** Confirmed dead after `SubBlock`'s removal (zero other callers). Removed both.

**Also found independently** (not in either review, caught by the full `cargo test --workspace` run finishing after this PR opened): `SagaIdempotencyLru::len()`'s removal broke 5 `#[cfg(test)]` call sites, invisible to the apply step's plain `cargo check`. Fixed in 69e3558.

**And the Windows CI failure**: a second, `#[cfg(target_os = "windows")]`-gated \`Ctx { conn_id, ... }\` construction site in \`srv_ipc/server.rs\` (\`dispatch_synthetic_goodbye\`) that macOS-only local verification structurally couldn't see. Fixed, and I audited every other removed symbol in this PR for the same platform-cfg blind spot (struct-field construction sites, enum-variant match arms, trait impls) — no other instances found.

Re-verified after all fixes:
- \`cargo check --workspace --tests\` — clean
- \`tsc --noEmit\` — clean (pre-existing unrelated errors aside)

The remaining \`vitest\` CI failure is pre-existing and unrelated to this PR — all 14 failed test files trace to the same root cause (\`frontend/app/element/markdown.tsx\` importing \`hast-util-to-jsx-runtime\`, a package never installed in this repo; that file is completely untouched by this PR).

AgentO-asaf:
Correction to my earlier claim that the \`vitest\` failure was pre-existing — it wasn't, and I was wrong to say so without tracing it further. Fixed in 07f8197.

**Root cause**: removing the \`streamdown\` package (genuinely unused after \`streamdown.tsx\`, its only consumer, was correctly deleted by this PR) also removed its *transitive* dependencies \`hast-util-to-jsx-runtime\` and \`remark-rehype\`. \`markdown.tsx\` imports both of these directly but never declared them as its own dependencies — a pre-existing phantom-dependency gap (this PR's own earlier findings report already flagged all 24 "Unlisted dependencies" in this repo, these two among them, but I hadn't acted on it since it wasn't in scope for dead-code removal).

**Why other PRs' CI stays green**: they don't touch \`package.json\`, so \`streamdown\` stays installed and keeps silently providing these two packages transitively. Only a PR that removes \`streamdown\` (like this one) surfaces the gap.

**Fix**: declared both as explicit direct dependencies at the versions already resolved on \`main\`.

Re-verified clean:
- \`tsc --noEmit\` — 0 errors (previously 2)
- \`vitest run\` — 2250/2254 passing, 0 failed test files (previously 14 failed, all traced to this). The 1 remaining "failed" test is the same environmental property-test timeout noted earlier in this thread — reproduced passing 24/24 in isolation immediately after.

a5af:
@codex review

AgentO-asaf:
Good catch — this landed just before merge and slipped through. Did a full sweep, not just this one file: found and fixed 8 broken internal links (in the 3 moved files that had relative links) and 5 broken inbound links (from 3 other docs pointing at pre-move paths), all traced to the same root cause you flagged. Fixed in agentmuxai/agentmux@b38fbb6 on #2409.

PR #2408, opened 2026-08-04
fix(agent-pane): composer strip centers at wide widths, pairs HOST/SANDBOX with Shell

a5af:
## Summary

PR #2393 (`SPEC_COMPOSER_STRIP_LEFT_JUSTIFIED_TIERED_WRAP_2026_08_03.md`) correctly fixed a real clipping bug by making the composer strip always left-justify, wrapping onto additional lines instead of overflowing under the textarea. But "always" went further than intended — wide panes with a full line of room to spare lost the true-centered stats zone as a side effect, not a deliberate goal. User feedback, same day: it should stay centered as long as content fits on one line, and only switch to left-justified once a 2nd line actually appears.

- **Restores centering at a new widest tier** (`@container agent-pane (min-width: 482px)`): a symmetric `flex: 1 1 0` split on `.agent-composer-strip-controls` and `.agent-composer-strip-right`, with `.agent-composer-strip-stats-zone` centered between them — the standard "true-center a middle flex item regardless of unequal side content width" pattern. Same visual result as the pre-08-03 `1fr auto 1fr` grid, **without** reintroducing `display: grid` (that's what caused the original clipping bug) and **without** touching the narrow-tier `flex-wrap` logic that actually fixed it — below 482px, behavior is byte-for-byte unchanged from #2393.
- **482px is live-measured, not estimated.** First shipped as a `640px` guess; live testing in a real `task dev` agent pane found the actual 1-line/2-line wrap boundary sits at ~482px, so 640px left a ~160px dead zone where content already fit on one line but the breakpoint hadn't kicked in — rendering left-justified when it should've stayed centered. Corrected to the measured value.
- **Pairs the HOST/SANDBOX runtime tag with the Shell toggle** as a single unit, moved out of the controls zone to sit immediately left of Shell (direct user request). Wrapped in a shared `inline-flex` span (`.agent-composer-strip-host-shell`) rather than just adjacent DOM order, so the pair is one atomic flex item inside the right zone's own internal `flex-wrap` — they move together across a wrap instead of independently landing on separate sub-lines.

Full design writeup and rationale: `docs/specs/SPEC_COMPOSER_STRIP_LEFT_JUSTIFIED_TIERED_WRAP_2026_08_03.md` (addendum section).

## Test plan

- [x] `npx stylelint frontend/app/view/agent/styles/_composer-strip.scss` — one pre-existing hex-color error unrelated to this change (already documented in the original #2393 spec)
- [x] `npx vitest run frontend/app/view/agent` — 954/954 tests pass across 74 files (no regressions)
- [x] **Live verification in `task dev`** — the gap the original #2393 PR explicitly left open. Resized a real agent pane through the full range: confirmed content stays centered at every width down to the exact 2nd-line wrap point (no dead zone), switches to left-justified only once wrapping begins, and the HOST/SANDBOX + Shell pair renders and moves together as intended.

### DIFF
--- a/.changesets/1785805646-fix-agent-pane-composer-strip-centers-at-wide-widths-pairs-h-mmmd.md
+++ b/.changesets/1785805646-fix-agent-pane-composer-strip-centers-at-wide-widths-pairs-h-mmmd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): composer strip centers at wide widths, pairs HOST/SANDBOX with Shell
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

a5af:
@codex review

PR #2410, opened 2026-08-04
feat(auth): in-app Claude OAuth login — core session + launch surface

AgentU-asaf:
## Summary

Live probes today show the factual basis for skipping in-app login for Claude is obsolete: **claude CLI 2.1.198 (our pinned version) prints the full PKCE authorize URL and accepts a pasted code on stdin** via the new `claude auth login` subcommand — and auto-completes by polling once the user authorizes in the browser, no paste needed in the happy path.

The paste-code UI (`AuthUrlBox`) and its `set_provider_auth` → `CliLoginStdin` plumbing were never deleted, only unrouted for Claude (`headlessLoginUrlUnsupported` + `skipTier1`, both decided against v2.1.183 behavior).

The spec (`docs/specs/SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03.md`) revives that machinery as one shared in-app login session used at all three auth surfaces:
1. New-agent launch (`PreLaunchAuthPanel`)
2. Credential-loss relogin (agent pane)
3. Armory Accounts + Agent Stash (closes the gap where a fresh profile has no way to create a Claude account at all — see the v0.54.9 stuck-instance incident: spawn gate correctly blocks, but the only UI outcome is a dead-end "Agent encountered an error")

Terminal login (tier 3) demotes to an explicit manual fallback; seed-from-global (tier 2) stays as the fast path. Feature-gated by URL-capture success, so older CLIs degrade to today's behavior — no version pin.

Phased as 4 PRs (core+launch → relogin → Armory/Stash → cleanup); **this PR is PR 1: spec + core in-app session (§3.1) + catalog/flag changes (§3.2) + the launch surface (§3.3 surface 1)**. Relogin (surface 2) and Armory/Stash (surface 3) land in the follow-up stacked PRs.

## Implementation

**Catalog / behavior-gate (§3.2)** — `providers/catalog.ts` drops `headlessLoginUrlUnsupported` for Claude and replaces the stale "DEAD END for Claude v2.1.x" rationale with the 2026-08-03 probe evidence. The flag mechanism itself stays (`providers/types.ts`) as a behavior-gate for CLIs that genuinely print no URL; older Claude CLIs just time out tier-1 capture (15s) and fall through to tiers 2→3 exactly as before — no version checks anywhere.

**Host capture (`agentmux-cef`)** — `extract_url` already matched the `https://claude.com/cai/oauth/authorize?…` URL incl. OSC-8 hyperlink wrapping; now pinned by `extract_url_claude_authorize_tests` (plain fallback line, OSC-8 BEL/ST-terminated, CSI-wrapped, non-auth-URL negative). One minimal new IPC, `get_cli_login_status` (`{ active }` off the generation-guarded `cli_login_stdin` slot): the child-exit half of the §3.1 completion check — a credential probe alone false-positives on present-but-expired tokens during a reconnect and would reap the login child mid-flight.

**Core session (`flows/run-provider-login.ts`)** — new `awaitTier1Completion` option: after tier 1 captures a URL, the call stays alive as the in-app session, polling for child-exit AND credential material in the minted isolated per-account dir (`CLAUDE_CONFIG_DIR` is pointed at it before any tier runs — existing minting, no new credential storage), then persists + links the account itself and resolves with new outcomes `"inapp-success"` / `"inapp-timeout"` (plus a new `"inapp-waiting"` `onTierChange` event carrying the live deadline). Child exits without a credential → fail fast after grace re-checks. The child is reaped on every exit path. Callers not passing the flag keep the existing `"opened"` + own-poll + `persistAndLinkAccount` contract unchanged.

**Launch surface (`components/PreLaunchAuthPanel.tsx`)** — the `requiresLoginTty` branch no longer hardcodes `skipTier1: true`; it runs the awaited in-app session and renders a new `InAppLoginPanel` (AuthUrlBox's 2-step pattern: authorize URL + Open/Copy, paste-code → `setProviderAuth` → login-child stdin, live phase line, Cancel) via a `"provider-login"` sentinel session id. Terminal login is an explicit secondary "Use terminal instead" action, never auto-launched on the happy path. Connect is the primary CTA for Claude again; "Use my existing login" (seed-from-global) demotes to secondary. `/login` (`commands/global/login.ts`) picks up tier 1 for Claude automatically via the flag drop — its composer `AuthUrlBox` + completion poll already handle it.

**Not touched (later PRs):** `useAgentControllerStatus.ts` (surface 2 — though its flag-derived `skipTier1` means relogin now benefits from tier-1 capture with its existing generic "opened" handling), `frontend/app/view/accounts/`, identity model/view (surface 3).

### Test results
- `tsc --noEmit`: clean (exit 0)
- `vitest run frontend/app/view/agent`: **74 files, 963 tests, all passing** — incl. 8 new awaited-session tests (inapp-success w/ persist+link, completion-requires-child-exit on stale-token reconnect, fail-fast on child-exit-without-credential, cancel, feature-gate fallthrough to tier 2/3 when no URL, `inapp-waiting` deadline event, persist-failure gating, mint-failure downgrade to `"opened"`) and a `/login skipTier1: false` pin
- `cargo test -p agentmux-cef --lib cli_login`: **13 passed** (5 new URL-capture tests)

## Verification evidence
- `claude auth login` under PTY, isolated `CLAUDE_CONFIG_DIR`, **no DISPLAY**: prints `If the browser didn't open, visit: https://claude.com/cai/oauth/authorize?code=true&client_id=…&code_challenge=…` then `Paste code here if prompted >` — on both 2.1.198 (pinned) and 2.1.214.
- End-to-end completion verified live: browser authorize → CLI self-detects via state polling → credential minted, zero terminal interaction.

### DIFF
--- a/.changesets/1785808171-feat-auth-in-app-claude-oauth-login-core-session-catalog-rev-8sgd.md
+++ b/.changesets/1785808171-feat-auth-in-app-claude-oauth-login-core-session-catalog-rev-8sgd.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(auth): in-app Claude OAuth login — core session, catalog revival, and launch surface
--- a/agentmux-cef/src/commands/cli_login.rs
+++ b/agentmux-cef/src/commands/cli_login.rs
@@ -64,18 +64,22 @@ impl CliLoginStdin {
 const LOGIN_REAP_TIMEOUT_SECS: u64 = 6 * 60;
 
 /// Cap on how long run_cli_login_pty blocks waiting for a scrapeable OAuth
-/// URL before giving up and returning auth_url=None. Frontend callers that
-/// know in advance a provider never prints one (Claude — see catalog.ts's
-/// headlessLoginUrlUnsupported flag) skip this call entirely via
-/// runProviderLogin's skipTier1, so in practice this only bounds providers
-/// that plausibly DO print a URL (Codex/Gemini/OpenClaw).
+/// URL before giving up and returning auth_url=None. Bounds every provider
+/// that plausibly prints a URL — Codex/Gemini/OpenClaw, and since
+/// SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03.md §3.2 Claude as well: the
+/// pinned CLI (2.1.198+) prints the full authorize URL
+/// (`https://claude.com/cai/oauth/authorize?…`) under this PTY spawn,
+/// verified by live probes on 2026-08-03, so catalog.ts's
+/// headlessLoginUrlUnsupported flag was dropped for Claude and tier 1
+/// reaches this function again. An OLDER Claude CLI (≤2.1.183 behavior,
+/// which prints nothing) simply times out here — that's the frontend's
+/// behavior-gate: auth_url=None makes runProviderLogin fall through to
+/// tiers 2/3 exactly as before, with no CLI version check anywhere.
 ///
-/// Left at 15s, NOT shortened as a "safety margin" for Claude: Claude never
-/// reaches this function at all now (skipTier1 above), so shortening it
-/// would only affect Codex/Gemini/OpenClaw — providers that legitimately
-/// need up to this long to print their URL. reagent P1 on PR #2300 caught
-/// an earlier attempt to cut this to 5s, which would have killed valid
-/// in-progress OpenClaw logins and forced an unnecessary terminal fallback.
+/// Left at 15s, NOT shortened as a "safety margin": reagent P1 on PR #2300
+/// caught an earlier attempt to cut this to 5s, which would have killed
+/// valid in-progress OpenClaw logins (they can take close to the full 15s
+/// to print their URL) and forced an unnecessary terminal fallback.
 const URL_CAPTURE_TIMEOUT_SECS: u64 = 15;
 
 /// Spawn a CLI auth login flow.
@@ -687,8 +691,11 @@ async fn run_cli_login_pty(
                             t.len()
                         );
                     }
-                    // Still capture an OAuth URL for providers that DO print one
-                    // (Codex/Gemini/OpenClaw); a no-op for Claude v2.1.x.
+                    // Capture the OAuth URL for providers that print one —
+                    // Codex/Gemini/OpenClaw, and Claude 2.1.198+ (its "If the
+                    // browser didn't open, visit: https://claude.com/cai/oauth/
+                    // authorize?…" fallback line, sometimes OSC-8-hyperlinked;
+                    // SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03.md §2).
                     if let Some(tx) = url_tx.take() {
                         if let Some(u) = extract_url(&line) {
                             let _ = tx.send(Some(u));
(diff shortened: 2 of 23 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
Fixed the child-liveness race (`cli_login.rs`) in `e8627f169` — `get_cli_login_status` now reads a dedicated `cli_login_active` flag set at spawn and cleared only by each transport's own reaper task on actual child exit, instead of inferring liveness from `cli_login_stdin`'s presence (which `set_provider_auth` clears the instant a code is pasted, before the child finishes processing it).

Re the `useAgentControllerStatus.ts` relogin child-aware-completion finding: fixed in stacked PR #2413, which moves `relogin()` onto `awaitTier1Completion: true`.

Re the two spec-doc wording findings: the actual implementation already does the right thing in both cases (tier 3 auto-launches only via the unchanged feature-gate path, never after an `awaitTier1Completion` timeout; the Armory tile calls `runProviderLogin` with `provider: PROVIDERS["claude"]`, not the "anthropic" account-provider id — see PR #2414's `ClaudeLoginPanel`). Will tighten the spec's wording in the cleanup PR (#4 of this stack) so it doesn't read as ambiguous against the code that actually shipped.

a5af:
@codex review

a5af:
Both P1 findings fixed in `797582970`:
- `useAgentControllerStatus.ts`: hardcoded `skipTier1: true` for relogin in this PR alone (the real fix — child-aware completion via `awaitTier1Completion` — lives in the stacked #2413, which now supersedes this hardcode via its own rebase).
- `PreLaunchAuthPanel.tsx`: the URL/paste box is now gated on `phase` too, not just `authUrl`'s presence, so it disappears once "Use terminal instead" moves the session past waiting-authorize.

963 frontend tests green, typecheck clean.

a5af:
@codex review

a5af:
Round-2 findings all fixed:
- [P1] `useAgentControllerStatus.ts` reconnect stale-credential risk — superseded by `awaitTier1Completion` in #2413, which now supersedes the temporary hardcode I'd put in this PR.
- [P1] Credential freshness: added `cli_login_cred_baseline`/`credential_changed` (mtime-tracked) — completion now requires child-exit AND authed AND credential_changed, closing the reconnect-crash false-positive.
- [P2] Cross-surface generation confusion: `get_cli_login_status` now reports `generation`; the poll captures it on first read and returns `superseded: true` (skipping `cancelCliLogin()`) the moment it changes.
- [P2] URL query-param logging leak — added `redact_url_query`, used at both log sites.
- [P2] `force-login.ts` cancellation gap before opening a browser.
- [P2] `PreLaunchAuthPanel.tsx`: `isStaleAction` added to the cancellation predicate; missing `catch` clause added (rejections no longer leave the panel stuck).

18 Rust + 966 frontend tests green, typecheck clean.

a5af:
@codex review

a5af (approved):
Approving to clear the required-review gate: reagentx-workflow's most recent re-review (2026-08-04T10:50:37Z, commit 1a09c7bfe) found zero outstanding issues after the fix commits landed, and I've independently re-verified at that same commit — `npx tsc --noEmit` clean, `npx vitest run frontend/app/view/agent` (74 files / 969 tests) all passing, and the relevant `agentmux-cef` cli_login unit tests (22/22) passing. The three stale CHANGES_REQUESTED reviews above (whose findings were fixed by 2e09d3158/1a09c7bfe) were dismissed with rationale linking each to its fix commit.

PR #2411, opened 2026-08-04
feat(launcher): detect srv hang-while-alive and force a recycle (#942)

AgentA-asaf:
## Summary

Closes the named non-goal in `SPEC_SRV_SUPERVISION_RECYCLE_2026_07_11.md`
(#2107): *"srv hang-while-alive detection ... is a separate follow-up in
the #942 family."*

srv exit-triggered recycle (#2107) already recovers cleanly from a crashed
srv, but nothing detected srv staying **alive but wedged** (deadlock,
exhausted blocking-thread pool, stuck `await`) — `srv_child.wait()` never
resolves in that case, so the existing crash-recycle machinery never fires
and the session silently stops responding.

- New `agentmux-launcher/src/srv_liveness.rs`: a 10s HTTP health-check probe
  against srv's existing unauthenticated `GET /` endpoint, 3-consecutive-miss
  threshold (~30s to detect, same order of magnitude as the host teardown
  backstop's 30s grace).
- On detecting a hang: kill srv directly (`start_kill`) and reset the
  counters. The existing `srv_status = srv_child.wait()` arm (next loop
  iteration) drives the already-shipped #2107 respawn/rebind/host-recycle
  path **unmodified** — this module's only job is deciding "treat this as a
  crash," not how to recover from one.
- Simpler by design than the host's `ui_liveness`/`teardown_backstop` pair:
  srv's health check is synchronous (pass/fail within one tick, no
  cross-tick nonce matching) and there's no "legitimate absence" state to
  guard against the way the host's zero-window backstop must.
- Uses `tokio::net` (not the `second_instance.rs` blocking-`std::net`
  precedent) since this probe runs every tick inside the same `select!`
  loop that also detects host/srv exit and the teardown backstop — a
  blocking call would stall all of those on every hiccup.

**Windows-only.** `SRV_RESTART_BUDGET`/`SRV_RESTART_WINDOW` (the machinery
this plugs into) are `#[cfg(target_os = "windows")]` today — srv
crash-recycle itself isn't ported to Unix yet, so building hang detection
on top of it there would be dead code. Unix parity is a natural follow-up
once srv recycle lands on that platform.

Full design writeup: `docs/specs/SPEC_SRV_HANG_WHILE_ALIVE_DETECTION_2026_08_03.md`.

## Test plan

- [x] `cargo test -p agentmux-launcher` — new `srv_liveness` unit tests
      (miss-count/recycle-decision, pure) + 3 `probe()` integration tests
      against a local `tokio::net::TcpListener` (200 response, stalled
      response past timeout, connection refused) — all pass.
- [x] Full launcher suite (213 tests) green, no regressions.
- [x] `cargo check -p agentmux-launcher` clean.
- [ ] Live verification on a Windows machine (needs Sysinternals `pssuspend`
      to freeze srv without killing it — not available in this sandbox):
      boot `task dev`, `pssuspend <srv_pid>`, confirm the launcher log shows
      3 missed probes (~30s) then "srv wedged — forcing recycle", srv is
      killed, the existing "Restoring session…" respawn/reproject sequence
      fires, windows come back, and `SRV_RESTART_BUDGET` (3/120s) still
      exhausts correctly on repeat.

### DIFF
--- a/.changesets/1785806209-feat-launcher-detect-srv-hang-while-alive-and-force-a-recycl-s606.md
+++ b/.changesets/1785806209-feat-launcher-detect-srv-hang-while-alive-and-force-a-recycl-s606.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(launcher): detect srv hang-while-alive and force a recycle (#942)
--- a/agentmux-launcher/src/main.rs
+++ b/agentmux-launcher/src/main.rs
@@ -56,6 +56,10 @@ mod splash_text;
 mod splash_config;
 mod splash_info;
 mod startup_events;
+// srv hang-while-alive detection plugs into srv recycle-on-exit
+// (SRV_RESTART_BUDGET, supervisor/mod.rs), which is Windows-only today.
+#[cfg(target_os = "windows")]
+mod srv_liveness;
 mod srv_spawner;
 mod state;
 mod wrr;
--- a/agentmux-launcher/src/srv_liveness.rs
+++ b/agentmux-launcher/src/srv_liveness.rs
@@ -122,23 +122,23 @@ pub fn should_recycle() -> bool {
 /// `select!`) for up to `timeout` on every hiccup.
 ///
 /// Success = a response starting with `HTTP/1.1 200`. `web_endpoint` is
-/// `srv_result.web_endpoint`, already in `http://127.0.0.1:{port}` form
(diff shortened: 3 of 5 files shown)


a5af:
@codex review

AgentA-asaf:
@codex review

a5af:
@codex review

PR #2412, opened 2026-08-04
fix(agent-pane): remove the model selector's misleading trailing caret

AgentY-asaf:
## Summary
- User-reported: the collapsed model selector ("Mode · Model · Effort" trigger pill) shows an extra dot separator at the end that isn't needed.
- Root-caused via screenshot: `compactSummary()` (`AgentRuntimeDropup.tsx`) is a plain `.join(" · ")` over 3 always-non-empty values -- provably correct, no stray separator possible. The "extra dot" is actually the trigger's own expand/collapse caret (`▴`, `font-size: 8px`, `opacity: 0.7`) sitting right after the label, small/faint enough to read as a 4th separator dot.
- Per explicit user direction: removed the caret entirely rather than just restyling it -- the whole pill is already clickable, so no dedicated indicator glyph is needed.
- Cleaned up the now-dead `.agent-runtime-dropup-trigger-caret` CSS rule and `justify-content: space-between` (irrelevant with a single remaining child).

## Test plan
- [x] `npx tsc --noEmit` -- clean.
- [x] `npx vitest run frontend/app/view/agent` -- 954/954 pass.
- [ ] Manual: open an agent pane, confirm the model selector pill shows "Mode · Model · Effort" with no trailing mark, and still opens the panel on click.

### DIFF
--- a/.changesets/1785807323-fix-agent-pane-remove-the-model-selector-s-misleading-traili-ucxu.md
+++ b/.changesets/1785807323-fix-agent-pane-remove-the-model-selector-s-misleading-traili-ucxu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): remove the model selector's misleading trailing caret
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2416, opened 2026-08-04
fix(ambient): ghost-text suggestions read as direct instructions

AgentY-asaf:
## Summary
- User report: the composer's ghost-text next-prompt suggestion reads like "Yeah let's debug that blank preview bug next" instead of "Debug the blank preview bug next" -- conversational filler, not a direct instruction.
- Root cause: `next_prompt_suggestion`'s prompt (`session.rs`) asked Haiku to predict "a message the user might send to continue the conversation" -- inviting casual chat register. `activity_summary`/`subagent_name`'s prompts are already imperative/factual and don't have this problem, so the fix is scoped to that one prompt.
- Same "prompt nudge + reliable sanitizer" split this codebase already established for the fence/quote-wrapping bug (`SPEC_AMBIENT_SUMMARY_SANITIZATION_AND_TERSENESS_2026_07_08.md`):
  1. Tightened the prompt to explicitly ask for a direct, imperative command, with a concrete before/after example baked in.
  2. Extended `sanitize_ambient_text` (the shared choke point for all three ambient Haiku call purposes) with `strip_conversational_preamble` -- strips a small, non-exhaustive list of known filler openers ("Yeah, let's ", "Go ahead and ", "Sure, ", etc.) from the start of the string only, re-capitalizing what remains. This is the reliable half of the fix, per the same philosophy already documented for the fence/quote case.

## Test plan
- [x] 7 new unit tests covering: filler stripping, mid-sentence "lets" left alone, preamble inside quotes/fences, filler-only input left alone (no trailing content to strip).
- [x] `cargo test -p agentmux-srv` -- 1978/1978 pass.
- [ ] Manual: trigger the ghost-text suggestion after a turn and confirm it reads as a direct instruction, not "Yeah, let's..."

### DIFF
--- a/.changesets/1785875969-fix-ambient-ghost-text-suggestions-read-as-direct-instructio-z6gs.md
+++ b/.changesets/1785875969-fix-ambient-ghost-text-suggestions-read-as-direct-instructio-z6gs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ambient): ghost-text suggestions read as direct instructions, not chat filler
--- a/agentmux-srv/src/server/app_api/session.rs
+++ b/agentmux-srv/src/server/app_api/session.rs
@@ -527,12 +527,17 @@ fn register_session_next_prompt_suggestion(engine: &Arc<WshRpcEngine>, state: &A
                 }
 
                 let prompt = format!(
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2417, opened 2026-08-04
fix(dev): default dev-agent.cmd window title to agent identity

AgentY-asaf:
## Summary
- `scripts/dev-agent.cmd` (the CLAUDE.md-documented Windows entry point agents use to invoke `task dev`) now defaults `TITLE` to the caller's own `$AGENTMUX_AGENT_ID` when no explicit `TITLE=` argument is passed, so the OS taskbar shows whose dev window is whose instead of a generic "AgentMux" for every agent-launched session.
- An explicit `TITLE=` from the caller still always wins — this only fills the gap when it's absent.
- No changes to `Taskfile.yml` or `frontend/app-init.ts` — reuses the existing `TITLE=` → `VITE_DEV_TITLE` → `UpdateObjectMeta` flow from `specs/SPEC_DEV_WINDOW_TITLE_ARG_2026_06_25.md` unmodified; only the caller changed.
- Corrected that spec's stale `Status: Draft` header (the design was already shipped) and added an addendum documenting this follow-up.

## Test plan
- [x] Empirically exercised the batch logic via throwaway `.cmd` files under `cmd.exe`: no explicit TITLE + agent-id set → defaults correctly; explicit TITLE passed → caller's value wins; genuinely-unset agent-id (`env -u AGENTMUX_AGENT_ID`) → no TITLE arg added, `task dev` runs with no override.
- [x] Pure batch script change — no compile step required.

### DIFF
--- a/.changesets/1785877152-fix-dev-default-dev-agent-cmd-window-title-to-agent-identity-tu4r.md
+++ b/.changesets/1785877152-fix-dev-default-dev-agent-cmd-window-title-to-agent-identity-tu4r.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dev): default dev-agent.cmd window title to agent identity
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2418, opened 2026-08-04
feat(cron): disambiguate AgentMux vs native scheduling tools, add expiry bound

AgentX-asaf:
## Summary
Design doc + Phase 0 implementation: `docs/specs/SPEC_AGENT_POLLING_AND_WAKEUP_HARDENING_2026_08_04.md`.

Triggered by live use of `mcp__agentmux__Loop` this session to babysit a GitHub PR's review status — the wrong tool for a same-session self-check. The harness already ships native `ScheduleWakeup`/`CronCreate`/`CronList`/`CronDelete` tools with better-designed backoff, jitter, durability, and expiry semantics than AgentMux's own `Loop`/`Cron`, but AgentMux's tools shadow them under near-identical names (`CronCreate` vs `mcp__agentmux__CronCreate`) with no indication in either description that the other exists — a collision already flagged in `SPEC_CRON_LOOP_ROBUSTNESS_2026_06_25.md` over a month ago and never resolved.

- Rewrote `Loop`'s and all five `Cron*` tools' MCP descriptions to state explicitly they're for targeting a *different* agent, and to recommend native `ScheduleWakeup`/`CronCreate` for same-session self-checks (zero delivery overhead vs. this family's cross-agent muxbus envelope). No schema/functional change to `Loop`/`LoopStop`/`LoopList`.
- Added an optional `max_age_secs` hard expiry bound to AgentMux's own `CronCreate` (matching the spirit of native `CronCreate`'s 7-day auto-expiry) alongside the existing `max_fires` bound. Opt-in, defaults to no expiry — AgentMux's cross-agent jobs are often meant to run indefinitely (e.g. a recurring standup check), so an unconditional default would be a surprising behavior change for existing automations.
- New `db_cron_jobs.max_age_secs` column (`SHARED_STORE_SCHEMA_VERSION` v4→v5, additive/idempotent `ALTER TABLE`, `NULL` for all pre-existing rows = no behavior change for jobs created before this shipped).
- Extracted the age-expiry check into a standalone `is_expired_by_age()` function (matching the existing `should_catchup()` pattern) with unit tests, rather than leaving it inline and untested like the sibling `max_fires` check already was.

Explicitly not done here (see the design doc's Phase 1/2): consolidating `Loop`/`Cron` onto one primitive, or renaming AgentMux's tools outright — both need their own compat/rollout scoping.

## Test plan
- [x] `cargo check -p agentmux-srv -p agentmux-mcp` — clean
- [x] `cargo test -p agentmux-srv` — 1976 passed, 0 failed (5 new)
- [x] `cargo test -p agentmux-mcp` — 4 passed, 0 failed (27-tool-count assertion still holds — no tools added/removed, only description/schema changes)
- [ ] reagent review

### DIFF
--- a/.changesets/1785877245-feat-cron-agentmux-native-scheduling-tool-disambiguation-exp-fsjt.md
+++ b/.changesets/1785877245-feat-cron-agentmux-native-scheduling-tool-disambiguation-exp-fsjt.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(cron): AgentMux/native scheduling tool disambiguation + expiry bound
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -1519,7 +1519,7 @@ async fn call_tool(
                 let next = j["next_fire"].as_str().unwrap_or("—");
                 let fires = j["fire_count"].as_i64().unwrap_or(0);
                 let max = j["max_fires"].as_i64().map(|n| format!("/{n}")).unwrap_or_default();
-                let age_bound = j["max_age_secs"].as_i64().map(|n| format!("  expires_in={}s_of_age", n)).unwrap_or_default();
+                let age_bound = j["expires_in_secs"].as_i64().map(|n| format!("  expires_in={n}s")).unwrap_or_default();
                 lines.push(format!(
                     "  {}  {}  [{}]  fires={fires}{max}  next={}  expr='{}'{age_bound}",
                     j["id"].as_str().unwrap_or("?"),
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

AgentX-asaf:
Both real, both fixed in \`b1eec55e\`:

- **P2**: \`CronList\` was printing the raw \`max_age_secs\` bound, not remaining time. Server now computes \`expires_in_secs\` (clamped to 0) in \`to_view()\`; MCP CLI displays that instead of doing its own date math (avoids adding \`chrono\` to \`agentmux-mcp\` just for this).
- **P1**: confirmed — the age check only ran in "now" terms at loop-top, so a long cron interval could let one fire slip through after the hard bound. Now also checks whether the *next* scheduled fire time itself would land past expiry before sleeping for it, and disables without firing if so. Extracted the disable sequence into a shared `disable_job()` helper while I was touching all four call sites.

Added a regression test naming the exact P1 scenario. Full suites green: `agentmux-srv` 1977/0, `agentmux-mcp` 4/0.

PR #2419, opened 2026-08-05
fix(identity): tolerate a malformed secret_ref row instead of hiding every account

AgentA-asaf:
## Summary

Live-testing the in-app Claude OAuth login (`SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03.md`, #2410/#2414/#2415) surfaced a real data-integrity bug: one `db_accounts` row's `secret_ref` JSON has unescaped Windows-path backslashes — invalid JSON, confirmed as an out-of-band DB edit (not reproducible from any write path in this repo; `identity_upsert` always serializes correctly via `serde_json::to_string`).

`identity_list()` had **zero per-row error isolation**, so this one corrupted row made the *entire* 9-account list fail for every caller: Armory's account list UI, the boot-time identity sweep, and the post-login cache refresh all hit the identical `Conversion error from type Text at index: 5, invalid escape` error.

This is the same failure class a prior incident already flagged in a code comment (the `oauth_config_dir`/`o_auth_config_dir` tag-rename fix) without ever fixing the root no-isolation issue: *"`identity_list()` aborts entirely on the first unparseable row."*

## Changes

- `identity_list` now skips a malformed row (logged with its account id) instead of erroring the whole query. `identity_get` is unchanged (single-row fetch, no "skip to" fallback — its one problematic caller, the identity sweep, already tolerates a fetch failure per-candidate).
- New migration `m0019_repair_malformed_secret_ref`: scans `db_accounts.secret_ref`, repairs the one known corruption shape (escapes bare backslashes that aren't already a valid JSON escape, re-validates, only rewrites if the repair round-trips), and leaves unrepairable rows untouched — never deletes, since a broken row may still be linked via `db_agent_identity_links`.
- Updated one existing test (`session.rs`) whose fixture assumed the old whole-list-degrades-on-one-bad-row behavior — the accounts source is no longer reported as degraded for a single malformed row, since it's now genuinely tolerated at the source rather than caught downstream.

Full incident writeup: `docs/analysis/ANALYSIS_ARMORY_STASH_CREDENTIAL_VISIBILITY_GAP_2026_08_04.md` (not included in this PR — separate doc-only artifact from the investigation, happy to land it here too if useful).

## Test plan

- [x] `cargo test -p agentmux-srv` — 1990 tests pass (1979 + 4 + 7 across the three binaries), including 6 new tests (repair-function correctness, `identity_list` row-skip behavior, migration repair + idempotency + safe-no-op-on-unrecognized-corruption) and the one updated existing test.
- [x] `cargo check -p agentmux-srv` clean.
- [ ] Live verification on `task dev` against the real corrupted row on this machine's shared store — will confirm after review, before merge.

### DIFF
--- a/.changesets/1785892369-fix-identity-tolerate-a-malformed-secret-ref-row-instead-of--5nfj.md
+++ b/.changesets/1785892369-fix-identity-tolerate-a-malformed-secret-ref-row-instead-of--5nfj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): tolerate a malformed secret_ref row instead of hiding every account
--- a/agentmux-srv/src/backend/storage/identities.rs
+++ b/agentmux-srv/src/backend/storage/identities.rs
@@ -141,44 +141,31 @@ fn default_identity_status() -> String {
     "unknown".to_string()
 }
 
-/// Escape every backslash in `raw` that isn't already the start of a valid
-/// JSON escape sequence (`\"`, `\\`, `\/`, `\b`, `\f`, `\n`, `\r`, `\t`,
-/// `\uXXXX`). Targets exactly the one corruption shape seen in production —
-/// a raw Windows path (`C:\Users\...`) pasted into a JSON string without
-/// escaping — without touching JSON that's already correctly escaped, so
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2420, opened 2026-08-05
fix(agent-pane): remove synthetic system notification + decouple Working state from window focus

a5af:
## Summary

Two related fixes from the same investigation ("what causes the 'Picked up more work' message, and why does 'Working' disappear when I click the pane?").

### 1. Removed the "Picked up more work" synthetic system notification

This system (added in `SPEC_WORKING_STATE_AND_SCROLL_FOLLOW_HARDENING_2026_07_27.md` §4) injected a **permanent markdown node into the pane's own conversation transcript** — not a toast — via `postSystemNotification` → `StreamFlush`, whenever `StreamFlushObserved`'s intentional `Done.completed → Streaming` re-promotion happened after the user had already seen a settled "Worked" for ≥500ms. The CLI's `session_end` fires after every model round, not just true turn end, so this re-promotion is a real, necessary mechanism for genuine multi-round tool continuations — the notification was meant to explain an otherwise-confusing checkmark reversal instead of letting it happen silently.

Writing synthetic explanatory text directly into real conversation history is not wanted, even well-intentioned. Removed:
- The `createEffect` in `agent-view.tsx` that called `postSystemNotification("Picked up more work — starting another round…", "info")`.
- `settled-grace.ts` (the pure `SETTLE_GRACE_MS`/`shouldNotifyOnReopen`/`nextDoneCompletedAt` decision logic) and its test — confirmed via repo-wide grep this effect was their only consumer.
- Does **not** touch the underlying `StreamFlushObserved` re-promotion logic in `reducer.ts` — that's a separate, necessary correctness mechanism (removing it would leave the UI falsely showing "Worked" during genuine real work, a worse failure). Only the notification-about-it is removed.
- `postSystemNotification` itself is untouched — it's shared infra also used for login-success notifications.

### 2. Decoupled "Working" state from window focus

Investigating "why does Working leave when I click the agent pane" surfaced the real question underneath: why would that coupling exist at all? Traced to Mechanism 3 of the same spec — a focus-triggered re-poll (`GetControllerStatus` on every window background→foreground transition, added as a fallback for a live WebSocket event the pipeline can silently miss) that called `reconcileTurnActive()` off a single snapshot with no protection against `turn_active` reading transiently false during the same between-CLI-rounds gap described above. Because this poll fires on window focus, and clicking a pane you want to check on very often *also* focuses the window (the app was backgrounded), "Working" would flicker off on the single most common user action for looking at a pane.

Window focus was never conceptually relevant to whether the agent is working — it was only a convenient trigger reused from an unrelated feature (sound-notification suppression). Fix: stop this poll from calling `reconcileTurnActive` at all. Its other two purposes (turnJustEnded edge-tracking for ambient-summary/next-prompt-suggestion, and deferred controller-refresh recovery for a `/login` completed mid-turn) are untouched — both are lower-stakes and/or already idempotent against a stale snapshot.

The original stuck-Working-forever gap this mechanism existed to close is still covered, just via a different existing path: the live `useControllerStatusEvents` subscription reconciles `TurnPhase` from the backend's periodic status heartbeat (`persistent.rs`'s `spawn_status_heartbeat`, republishing every 20s while a turn is active) independent of window focus entirely. Worst case for the narrow "missed push + no WS reconnect" edge case moves from "instant on refocus" to "within ~20s", not "uncovered."

## Test plan

- [x] `npx vitest run frontend/app/view/agent` — 976/977 pass; the one failure (`tool-renderer registry — built-ins (parity)`, a 5000ms timeout) is unrelated to either change (doesn't touch tool-renderers or turn-phase reconciliation) and passes cleanly in isolation (~2000ms) — confirmed flaky under concurrent-suite load, not a regression.


a5af:
@codex review

PR #2421, opened 2026-08-05
fix(agent): surface 429/overloaded failures from the persistent controller for retry

Agent3-asaf:
## Summary

When Claude's API returns 429 (rate-limited) or overloaded and the `claude` CLI surfaces that error, the agent pane silently treats the turn as a normal completion instead of offering a retry — "sometimes it returns too busy, currently it just quits."

**This is a regression, not a missing feature.** The failure-classification + auto-retry framework already exists and already works correctly for every other provider:

- `agents::failure::classify()` already recognizes `429`/"rate limited" → `RateLimited` and `529`/"overloaded" → `Overloaded`, both `retryable: true`.
- `frontend/app/view/agent/hooks/useAgentFailure.ts` already subscribes to the `agentfailure` event and already arms a 5s→10s auto-retry countdown (capped at 2 attempts) for transient classes.
- `subprocess/host_spawn.rs` (used by Codex and other non-Claude providers) already wires classify → persist → publish correctly on exit.

**The gap:** Claude is hard-wired to a different controller, `persistent.rs` (so `AskUserQuestion` can block mid-turn without the process exiting between turns). That controller never calls `classify()` or publishes the failure event anywhere — so Claude specifically never triggers the recovery UI that already exists and already works for everyone else.

Full writeup: `docs/specs/SPEC_PERSISTENT_CONTROLLER_FAILURE_CLASSIFICATION_2026_08_04.md`.

## Changes

Two insertion points in `persistent.rs`, mirroring `host_spawn.rs`'s existing pattern — no new policy, no new event, no frontend changes:

1. **Mid-stream** — an `is_error_result` frame that the stale-`--resume` retry machinery has confirmed is *not* a live retry candidate (`!hold_back_for_resume_retry`) gets classified and published directly.
2. **Process exit** — only the `PersistImmediately`/`FlushErrorLine` `ResumeEffect` arm (the resume machinery has decided this exit is final) classifies and publishes, via a new `classify_exit_line()` helper. Deliberately does **not** touch the `FireRetry` arm, which must stay invisible to the user per its own existing "must never reach the user" invariant (a separate, internal retry for a stale `--resume` session id — untouched by this change).

`classify_exit_line()` is extracted as a small pure function specifically so the classify()-integration logic (JSON-vs-raw-text fallback, suppressing a low-confidence `UnknownNonZero` from noisy flushed text) has direct unit coverage without needing the full async process-spawn harness.

## Test plan

- [x] `cargo check` / `cargo clippy` — clean, no new warnings
- [x] All 85 pre-existing tests in `persistent.rs`/`persistent_resume.rs` still pass — no regression to the stale-`--resume` retry machinery this composes with
- [x] 6 new unit tests for `classify_exit_line()`: overloaded (JSON), rate-limited (JSON), rate-limited (raw-text fallback), auth error surfaced despite non-retryable, unrecognized JSON error suppressed, unrecognized raw text suppressed
- [ ] Manual `task dev` end-to-end verification not performed — forcing a real 429/overloaded from Anthropic on demand isn't practical; flagged in the spec's own testing plan as the weakest link in verification

### DIFF
--- a/.changesets/1785914390-fix-agent-surface-429-overloaded-failures-from-the-persisten-l11c.md
+++ b/.changesets/1785914390-fix-agent-surface-429-overloaded-failures-from-the-persisten-l11c.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): surface 429/overloaded failures from the persistent controller for retry
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -1273,6 +1273,23 @@ impl PersistentSubprocessController {
             self.filestore.as_ref(),
             global_output_zone.as_deref(),
         );
+        // Same gap reagentx flagged (PR #2421 P2) at the other FlushErrorLine
+        // call sites: this is also a previously held-back turn's error only
+        // now confirmed final (a fresh spawn superseding a still-tracking
+        // generation, or a retry batch exhausted with nothing left to
+        // deliver) — give it the same classify/persist/publish treatment so
+        // it isn't silently dropped from the pane's failure-recovery UI. No
+        // exit code exists for this now-superseded turn.
+        if let Some(failure) = classify_exit_line(None, &line) {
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2422, opened 2026-08-05
fix(login): resolve npm .cmd shims before spawning under a PTY on Windows

AgentA-asaf:
## Summary

Root cause of the "tier-1 in-app OAuth URL capture" bug from `SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03.md`: `run_cli_login_pty` (`agentmux-cef/src/commands/cli_login.rs`) spawned `cli_path` — a raw npm `node_modules\.bin\*.cmd` shim path — directly via `portable_pty::CommandBuilder`. On Windows this forces the OS through `cmd.exe /c "<path>.cmd" ...`, and **that layer hangs indefinitely under a real ConPTY**.

Live-confirmed via the actual production IPC path (`POST /ipc` `run_cli_login` with `requires_tty: true`, fresh `CLAUDE_CONFIG_DIR`): the spawned `cmd.exe` child sat alive at 0% CPU, produced zero PTY output, and never even created the target config directory — for the full 15s timeout and beyond (had to `taskkill` it manually). This matches the original bug exactly: `run_cli_login_pty: no auth URL captured within 15s`, silently falling through to the tier-2 fallback, masking the failure.

Ruled out along the way: the URL-matching regex (unit-tested, correct), CLI version drift (confirmed pinned 2.1.198), stray inherited env vars (confirmed absent in the actual failing run's own log), and the original "fresh config dir hits an onboarding wizard" hypothesis (directly refuted — a fresh dir works instantly without a PTY).

## Fix

`agentmux_common::make_cli_cmd`/`build_cli_cmd` already parses `.cmd` shims and resolves them to the real spawn target (a native `.exe` directly — Claude v2+'s actual shape — or `node <script>` for JS-based CLIs like OpenClaw) — and is already used by every other spawn site in the codebase (plain-pipe login, CLI version checks, persistent/subprocess block-controller spawns). `run_cli_login_pty` was the *only* one that bypassed it.

Factored the existing resolution logic into `pub fn resolve_cli_spawn_target(cli_path: &str) -> (String, Vec<String>)`, shared by `build_cli_cmd` (unchanged behavior, now delegates) and the new call site in `run_cli_login_pty`'s Windows branch. Provider-agnostic — fixes both Claude and OpenClaw (the other `requires_tty` PTY-spawn provider) with one change.

## Test plan

- [x] `cargo test -p agentmux-common` — 96 tests pass, including 4 new tests for `resolve_cli_spawn_target` (native `.exe` shim — using the real Claude `.cmd` content as a fixture; `node <script>` shim; non-`.cmd` passthrough; unparseable-shim fallback).
- [x] `cargo check -p agentmux-common` clean.
- [ ] Live re-verification on a rebuilt `task dev`: confirm the host log shows a captured URL within 15s instead of the timeout, and no orphaned `cmd.exe`/`claude.exe` process survives. Batching this with a related PR's live verification (Armory account-list fix, PR #2419) in one rebuild rather than rebuilding twice — will confirm before merge.

### DIFF
--- a/.changesets/1785915999-fix-login-resolve-npm-cmd-shims-before-spawning-under-a-pty--mh56.md
+++ b/.changesets/1785915999-fix-login-resolve-npm-cmd-shims-before-spawning-under-a-pty--mh56.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(login): resolve npm .cmd shims before spawning under a PTY on Windows
--- a/agentmux-cef/src/commands/cli_login.rs
+++ b/agentmux-cef/src/commands/cli_login.rs
@@ -602,7 +602,17 @@ async fn run_cli_login_pty(
             })
             .map_err(|e| format!("openpty for {cli_path}: {e}"))?;
 
-        let mut cmd = CommandBuilder::new(&cli_path);
+        // Resolve past the `.cmd`/`.bat` npm shim before spawning under the
+        // PTY — CommandBuilder on the raw shim path forces Windows through
+        // `cmd.exe /c`, which hangs indefinitely under a real ConPTY (no
+        // output, target config dir never created, confirmed live). Every
+        // other spawn site in this codebase already resolves through this;
+        // this was the one PTY-specific gap.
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2423, opened 2026-08-05
fix(accounts): unify ClaudeLoginPanel onto the canonical Modal, fix Stash nesting

AgentA-asaf:
## Summary

`ClaudeLoginPanel` hand-rolled its own overlay (`accounts-chooser-overlay`/`accounts-chooser` divs, manual backdrop-click-to-close, manual `role="dialog"`) instead of using the app's canonical `Modal` ("the one modal system for the whole app"). Not just an inconsistency — when opened from the Agent Stash's Accounts tab, it stacked a hand-rolled dialog *on top of* `AgentStashModal` (itself a real canonical `Modal`) with none of the shared system's coordinated focus-trap/ESC/backdrop stacking logic.

## Changes

- Replaced the overlay with `<Modal>`, added a `scope` prop so each caller supplies the right region:
  - Armory's Accounts tile (no `ModalLayer` ancestor) → `scope="tab"`
  - Agent Stash's Accounts tab (already inside `agent-view.tsx`'s `ModalLayer scope="pane"`) → `scope="pane"`
- Converted `inFlight` from a plain variable to a signal so `Modal`'s `closeOnBackdropClick` can reactively preserve the original "don't dismiss mid-login" guard.
- **Confirmed empirically, not just by inspection**: added a test that opens `ClaudeLoginPanel`'s `Modal` from inside another open `Modal` and asserts the outer one is *not* closed — `Modal` instances stack independently via `modal-stack.ts`'s containment-based reachability, as long as the inner one is used directly (not routed through `useModalLayer().open()`, which would replace rather than stack — a mistake that's easy to make since that's the established pattern for most of this app's other dialogs).

Part of the Fix 3 DRY-the-login-UI work from `docs/analysis/ANALYSIS_ARMORY_STASH_CREDENTIAL_VISIBILITY_GAP_2026_08_04.md`. A follow-up PR wires the same primitive into the previously-unbuilt credential-loss relogin surface (agent pane).

## Test plan

- [x] All 11 pre-existing `ClaudeLoginPanel.test.tsx` assertions pass unchanged — switched from `render()`'s container-scoped `findByText` to `screen.findByText` (document-wide), since `Modal` portals outside the test's render container. Query-scoping fix, not a behavior change.
- [x] New nested-modal-stacking regression test passes.
- [x] `agent-identity-links-panel.test.tsx` (13 tests, mocks `ClaudeLoginPanel` at the module level) unaffected — confirms the new `scope` prop doesn't break existing prop-passing assertions.
- [ ] Live re-verification on `task dev`: Armory → Accounts → Connect (shared `Modal` chrome); Agent Stash → Accounts tab → Connect/Re-login (dialog stacks correctly, Stash stays open behind it). Batching with the other open PRs' (#2419, #2422) live verification in one rebuild.

### DIFF
--- a/.changesets/1785916655-fix-accounts-unify-claudeloginpanel-onto-the-canonical-modal-42wj.md
+++ b/.changesets/1785916655-fix-accounts-unify-claudeloginpanel-onto-the-canonical-modal-42wj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(accounts): unify ClaudeLoginPanel onto the canonical Modal, fix Stash nesting
--- a/frontend/app/view/accounts/ClaudeLoginPanel.test.tsx
+++ b/frontend/app/view/accounts/ClaudeLoginPanel.test.tsx
@@ -10,9 +10,15 @@
  * old alias — otherwise the orphaned alias row lingers and the resolver's
  * inject.rs aborts every future spawn on it, even though a healthy
  * canonical "claude" link now exists right alongside it.
+ *
+ * Uses `screen.findByText` (document-wide), not `render()`'s own bound
+ * `findByText` — ClaudeLoginPanel now renders through the canonical `Modal`
+ * (ANALYSIS_ARMORY_STASH_CREDENTIAL_VISIBILITY_GAP_2026_08_04.md's Fix 3),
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2424, opened 2026-08-05
fix(build): exclude target/ from the Vite dev-server watcher

a5af:
## Summary

Hit the CEF extraction corruption class documented in `docs/retro/RETRO_CEF_BUILD_RACE_2026_04_24.md` twice in one session (`task dev` failing with `CMake Error: Cannot find source file: ../include/cef_id_mappers.h`), with a corruption pattern the Taskfile's own automatic `repair-cef-extract.sh` fallback doesn't catch: it only restores a destination subdirectory that's fully **missing or empty**, not one that's present, non-empty, but short exactly one file (which is what both occurrences looked like).

Full investigation, including a new suspect the original retro didn't have: `vite.config.ts`'s dev-server watcher had no `target/` exclusion. `task dev` runs the Vite watcher and the Rust/CEF build concurrently against the same working tree, so the dev server's own file watcher is one more process potentially touching/opening/statting files mid-extraction — alongside whatever the original retro already attributed to Windows Defender's real-time scan.

This PR applies the one clear, low-risk fix: excludes `target/**` from the dev-server watcher, removing AgentMux's own long-running process as a possible contributing actor (no functional loss — nothing under `target/` should ever trigger a frontend HMR reload anyway). Two further recommendations are documented but **not** applied here:
- A Windows Defender exclusion for the repo's `target/` directory (per the original retro's Option A) — needs admin rights this session didn't have.
- Hardening `repair-cef-extract.sh`'s detection to catch a *partially* populated destination (compare file counts/contents, not just presence/emptiness) — a more delicate change to an already carefully-reasoned script, left as a follow-up.

Full writeup: `docs/retro/RETRO_CEF_EXTRACT_PARTIAL_REPAIR_GAP_AND_VITE_WATCHER_2026_08_05.md`.

## Test plan

- Docs-only + a 2-line watcher config change; no runtime behavior change to verify beyond the config being valid TS.

### DIFF
--- a/.changesets/1785917467-fix-build-exclude-target-from-the-vite-dev-server-watcher-uqtv.md
+++ b/.changesets/1785917467-fix-build-exclude-target-from-the-vite-dev-server-watcher-uqtv.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): exclude target/ from the Vite dev-server watcher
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2426, opened 2026-08-05
Surface --resume outcome as an explicit pane transcript event

AgentY-asaf:
## Summary

Follow-up to an audit of how agent-pane history is managed
(`audits/AUDIT_AGENT_PANE_HISTORY_2026_08_05.md`, written outside this repo)
which found two independently-loaded stores per agent pane:

1. **What the pane displays** — AgentMux's own NDJSON transcript, persisted
   unbounded, loaded 200 lines at a time.
2. **What the model actually remembers** — the provider CLI's own session,
   reached only if AgentMux's `--resume <session_id>` handshake succeeds.

These can silently disagree: if a `--resume` attempt is stale/unreachable,
`persistent_resume.rs`'s existing state machine already knows precisely
(per-generation) that the model started a fresh session with no memory of
the visible scrollback above it — but that fact was only ever logged
(`tracing::info!`), never shown to the user. The pane keeps rendering
continuous-looking scrollback either way.

This PR adds `docs/specs/SPEC_AGENT_PANE_HISTORY_ALIGNMENT_2026_08_05.md`
(Part A design + implementation) and implements it:

- **Backend** (`persistent_resume.rs`, `persistent.rs`): a new
  `ResumeEffect::EmitSessionOutcome` fires at the two points where a
  `--resume` attempt's fate becomes unambiguously known (a `SessionCaptured`
  resolution, or a confirmed-dead retry about to fire fresh), persisted as an
  `agentmux_session_outcome` transcript frame — same shape/interception
  pattern as the existing `compact_boundary` frame. Purely additive: no
  existing state/effect/condition changes, only new effects alongside ones
  already produced. All 1999 backend tests pass (24 in
  `persistent_resume` are new/updated for this).
- **Frontend** (`session-outcome.ts` + wiring through
  `useAgentStream.ts`/`parseHistoryLines.ts`/`DocumentRow.tsx`/
  `renderers.ts`/`expansion-source.ts`): renders the event as a distinct
  divider — "Session continued" vs "New session started — prior
  conversation is not available to this agent" — mirroring the existing
  `context_compacted` divider pattern exactly. `tsc --noEmit` and the full
  agent-view vitest suite pass (990/991; the one failure is a pre-existing
  flaky timeout in an unrelated tool-renderer test, confirmed passing in
  isolation).

Parts B/C (rehydrating a resumable session's transcript into a new
instance/version before attempting `--resume`, and backfilling
`session_id`-less migrated agents) are scoped in the design doc §6 but
deferred to a follow-up PR.

## Test plan

- [x] `cargo test -p agentmux-srv` — 1999 passed, 0 failed
- [x] `npx vitest run frontend/app/view/agent/` — 990/991 passed (1
      pre-existing unrelated flake, passes in isolation)
- [ ] Manual verification in a running pane (not done — no browser/CEF host
      available in this environment; the div render is a straightforward
      mirror of the already-shipped `context_compacted` divider)

### DIFF
--- a/.changesets/1785955383-feat-agent-pane-surface-resume-outcome-as-an-explicit-transc-zb0y.md
+++ b/.changesets/1785955383-feat-agent-pane-surface-resume-outcome-as-an-explicit-transc-zb0y.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): surface --resume outcome as an explicit transcript event
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -45,6 +45,34 @@ pub const PERSISTENT_OUTPUT_SUBJECT: &str = "output";
 
 pub const BLOCK_CONTROLLER_PERSISTENT: &str = "persistent";
 
+/// Builds the NDJSON line for a `persistent_resume::ResumeEffect::
+/// EmitSessionOutcome` — a free function (not a method) so it's callable
+/// both from `PersistentSubprocessController::emit_session_outcome_line`
+/// and directly from the stdout-reader/process-waiter tasks below, which
+/// only hold `_read`-suffixed clones, not `&self`. See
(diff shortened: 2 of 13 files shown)


a5af:
@codex review

PR #2427, opened 2026-08-05
test(agent-pane): regression coverage for resize scroll-pin bug

AgentX-asaf:
## Summary

Investigated the reported bug: dragging a pane splitter while the agent pane is scrolled to the latest message causes it to lose its bottom-pinned scroll position.

- Added `AgentDocumentVirtualList.resize.test.tsx` — mounts the real, unmodified `AgentDocumentVirtualList` component with deterministic `ResizeObserver`/`requestAnimationFrame` fakes (jsdom implements neither) and simulates a splitter-drag resize: grow, shrink, both possible native-clamp/ResizeObserver orderings, and rapid multi-tick drags. All 5 cases pass against unmodified `main` — this **disproves** the leading "grow-clamp ordering race" theory from the earlier investigation and closes this file's previously-zero test coverage (it's the third documented incident of this bug class in two weeks, and this exact file had no regression suite).
- `docs/specs/PLAN_AGENT_PANE_RESIZE_SCROLL_PIN_2026_08_05.md` has the full writeup: original hypotheses, the math that disproves two of them, and the one remaining open question — whether `ResizeObserver` actually fires on every splitter-drag tick in the live app. That can't be verified in jsdom (no real ResizeObserver to falsify against) and needs a ~1 minute manual repro with DevTools open in a running build; this agent has no GUI automation tool to drive that itself.

No production code changed — this PR is test + docs only, so no changeset per the changesets README ("fixing a small thing and don't need a release? then don't add a changeset").

## Test plan

- [x] `npx vitest run frontend/app/view/agent/virtualization/AgentDocumentVirtualList.resize.test.tsx` — 5/5 passing
- [x] `npx vitest run frontend/app/view/agent/` — 982/982 passing (no regressions)
- [x] `npx tsc --noEmit` — clean on the new file
- [ ] Live repro per plan §7 Q2 — needs a human with the running app (see plan doc)

### DIFF
--- a/docs/specs/PLAN_AGENT_PANE_RESIZE_SCROLL_PIN_2026_08_05.md
+++ b/docs/specs/PLAN_AGENT_PANE_RESIZE_SCROLL_PIN_2026_08_05.md
@@ -1,0 +1,314 @@
+# Fix Plan: Agent Pane Loses Bottom-Pin on Pane Resize
+
+**Date:** 2026-08-05
+**Author:** AgentX (agent)
+**Status:** Updated 2026-08-05 (same day) — H1 disproven and the scroll-pin
+JS logic empirically validated via a new real-component test suite
+(`AgentDocumentVirtualList.resize.test.tsx`, 5/5 passing against unmodified
+`main`). See §3.5. No code fix has been made to `AgentDocumentVirtualList.tsx`
+— the tests show that file's logic is already correct for every resize
+ordering this plan could construct in jsdom. The remaining open question is
+live-only (§7) and needs one manual repro in a running build, which this
+agent cannot drive (no GUI automation tool available).
+
+**Reported symptom:** Dragging a pane splitter (gutter resize) while the agent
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

AgentX-asaf:
Fixed in fa8be24dd — good catch. The `clampFirst` boolean reordered two side-effecting calls but never flushed `requestAnimationFrame` between them, so both scroll events landed in the same coalesced `handleScrollNow()` batch regardless of order — the two "orderings" were byte-identical from the SUT's perspective.

Replaced it with explicit `clamp()`/`notify()` steps the test interleaves itself, plus a genuinely adversarial case that flushes rAF right after the native clamp — before the ResizeObserver notification runs at all — which is the literal worst-case ordering the (now-disproven) H1 theory relied on. Verified with a throwaway mutation test (forced `isNearBottom` to always return `false`) that the new test actually fails when the SUT is broken, and passes cleanly again after reverting the mutation. Full suite (982 tests) still green.

PR #2428, opened 2026-08-05
Fix App-API self-lookup for agents with mixed-case display names

AgentY-asaf:
## Summary

Investigation follow-up to the pane-history-alignment work — while
verifying the new native-memory bridge live against my own running pane
(agent "AgentY"), `MemoryList` failed with `"agent agenty not found"`.
Traced to a real, reproducible bug affecting several MCP tools, not just
memory.

**Root cause:** `AGENTMUX_AGENT_ID` in an agent's MCP server env is the
deliberately-lowercased `derive_slug(display_name)` routing slug (e.g.
`"agenty"` for `"AgentY"`) — confirmed intentional by
`build_mcp_config`'s own doc comment. Every App-API "tell me about
myself" endpoint that resolves this slug back to a persisted agent
record compared it against `instance_name` (which keeps the *original*
display casing) with **case-sensitive equality** — both in `db_agents`
SQL (`WHERE instance_name = ?1`) and the global cross-channel registry
(`r.data.instance_name == agent_id`). `"agenty" != "AgentY"`, so every
lookup failed for any agent whose name wasn't already all-lowercase.

**Confirmed live** against my own real running agent pane (not just
read from code):

| Tool | Result |
|---|---|
| `MemoryList` | HTTP 400 `"memory: agent agenty not found"` |
| `IdentityAccounts` | HTTP 500 `"unknown agent 'agenty': no instance with that name..."` |
| `PresetGet` (self mode) | No error — silently returned the generic blank preset instead of the real bound one |

## Fix

- `instance_get_by_name` (`agents.rs`) keeps its exact-match fast path
  and falls back to a slug-normalized scan (`derive_slug` on both
  sides) when that misses. This is the shared resolver behind
  `memory.*`, `identity.self.accounts`, `identity.account.validate`,
  and `bundle.self.get` (via `resolve_agent_definition_id`) — one fix
  covers all of them.
- `find_active_registry_record_by_slug` (`native_memory_handlers.rs`)
  — extracted from `memory_dir_from_registry`, fixed to slug-normalize,
  and exposed so `bundle_self_get_impl` can reuse the same lookup
  instead of reimplementing it.
- `bundle_self_get_impl` (`app_api/mod.rs`) now also falls back to the
  registry when no local `db_agents` row exists — it previously had no
  registry tier at all (unlike the memory path's two-tier lookup), so a
  registry-only agent silently got the blank preset with no error.
- `identity_auth_dirs.rs`'s test module was using its own module-local
  env lock for `AGENTMUX_HOME_OVERRIDE` instead of the crate-wide
  `ISOLATED_AUTH_ENV_LOCK`. This PR's own new tests exposed a genuine
  cross-module test race as a result (passed every solo run,
  intermittently failed full-suite parallel runs) — switched to the
  shared lock, closing exactly the gap it exists to prevent.

Every change is additive/corrective to existing lookup logic; no schema
or API shape changes.

## Test plan

- [x] `cargo test -p agentmux-srv` — 2001 passed, 0 failed
- [x] Full suite re-run 4x to confirm the identity_auth_dirs race is
      actually fixed, not coincidentally passing (it reproduced
      reliably before the fix, in ~4/4 full-suite runs)
- [x] New unit tests reproduce the exact live failure (mixed-case
      display name "AgentY" / slug "agenty") for both
      `instance_get_by_name` and the registry fallback, plus a
      registry-only `bundle_self_get_impl` case
- [ ] Not verified against the actual live srv instance serving this
      session's MCP tool calls — restarting that shared, running
      process (used by other concurrent agents on this machine) is out
      of scope for this PR; the unit tests reproduce the exact live
      error strings observed

### DIFF
--- a/.changesets/1785959990-fix-agent-api-resolve-app-api-self-lookup-by-the-agent-s-rou-v2vn.md
+++ b/.changesets/1785959990-fix-agent-api-resolve-app-api-self-lookup-by-the-agent-s-rou-v2vn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-api): resolve App-API self-lookup by the agent's routing slug, not its display-cased name
--- a/agentmux-srv/src/backend/storage/agents.rs
+++ b/agentmux-srv/src/backend/storage/agents.rs
@@ -1596,32 +1596,23 @@ impl Store {
     /// detection only cares about identity / cwd; ContinueNamed only
     /// cares about `id` + bindings).
     /// Spec: docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md §3b.
-    /// Resolve an instance by `instance_name` — but App-API callers (App
-    /// API self-lookup endpoints: `memory.*`, `identity.self.*`,
-    /// `bundle.self.get`) actually pass the `AGENTMUX_AGENT_ID` routing
-    /// slug, not the literal display name — it's stamped onto the MCP
-    /// server's env straight from `agent.slug` (`app_api::agent_open`),
-    /// the same stable, collision-resolved value stored in
-    /// `db_agents.slug` (`agent_def_insert_local_only`, never changes
-    /// after creation — see `AgentDefinition::slug`'s own doc comment).
-    /// `instance_name` is the renameable display name, overwritten on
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2432, opened 2026-08-06
feat(muxspect): diagnose and clear stuck Activity Dock entries

AgentA-asaf:
## Summary

Tonight's live-debugging session hit the exact bug `muxspect`'s own original spec was born from (its stated trigger was literally "user asked to locate 2 stuck sleep dock items") — an outer-harness-rejected Bash call left a `ToolNode` stuck at `status: "running"` in the Activity Dock, and no `muxspect` command could see or fix it. This adds the missing diagnostic + remediation layer.

Full design/root-cause: `docs/specs/SPEC_MUXSPECT_DOCK_DIAGNOSIS_AND_REMEDIATION_2026_08_06.md`. **This is not the fix for the underlying bug** — the spec explicitly reconciles against `SPEC_PROCESS_AND_TURN_STATE_TRACKING_CONSOLIDATION_2026_07_31.md`'s already-recommended first-party fix (a generation/max-age self-expiry on the dock itself) and positions this as the complementary manual/scriptable layer, useful before and after that fix ships.

## Root cause (traced, not fixed here)

A `ToolNode` is created optimistically at `status: "running"` the instant a tool call is *dispatched* (`stream-parser.ts`'s `toolCallToNode()`), before any confirmation it executed. The only way out is a matching `tool_result` for that exact call id, or a session-boundary scrub (`scrubOrphanedInProgress`, which only fires on pane close/reload/reconnect). An outer-harness pre-execution rejection surfaces as a top-level turn-ending error frame instead of a scoped `tool_result`, so the node never clears until the pane's session actually tears down.

## Changes

- **`agentmux-srv/src/backend/dock_snapshot.rs`** (new) — in-memory, latest-per-block cache of `ToolNode` status deltas. Never persisted (matches the dock's own nature).
- **New RPC command `docknodestatus`** — fire-and-forget push from the renderer on every `ToolNode` status change (`websocket.rs`, `rpc_types/{block,commands}.rs`).
- **`GET /api/v1/muxspect/dock?block_id=X`** — the cached snapshot, cross-referenced against `ProcessBroker`'s lifecycle to flag entries that look stuck (`running`, past the 30s promotion threshold, nothing srv-side backing the block).
- **`POST /api/v1/muxspect/dock/clear`** — muxspect's first mutating route. Validates the node is in the cache, then publishes a new scope-routed WPS event (`EVENT_DOCK_CLEAR`, `block:<id>` — same real server-side routing mechanism `EVENT_SHELL_NODE_CREATE` already uses, not client-side self-filtering).
- **`ForceCancelToolNode`** — new `agent-document` reducer command; the live renderer holding that block dispatches it on receipt, flipping the node to `canceled` with a distinguishing summary (`"⏹ Canceled — cleared via muxspect"`).
- **`muxspect dock <block_id>` / `muxspect dock clear <block_id> <node_id>`** — new CLI subcommands (`muxspect.mjs`), `--json` supported on `dock`.
- **`docs/MUXSPECT.md`** — new section documenting the feature.

## Test plan

- [x] `cargo test -p agentmux-srv` — 2032 + 4 + 7 passed, 0 failed (includes new `dock_snapshot` tests, `muxspect_handlers` dock/clear handler tests, and the stuck-flagging pure-function tests).
- [x] `cargo test -p agentmux-common` — 96 passed, unaffected baseline.
- [x] `npx vitest run` (full suite) — 152 test files / 2326 tests passed, 2 pre-existing skips. Includes new `ForceCancelToolNode` reducer tests (7) and `muxspect dock`/`dock clear` arg-parsing tests (4).
- [x] `npx tsc --noEmit -p .` — clean, project-wide.
- [ ] **Live verification not yet done**: boot `task dev`, trigger a stuck dock entry (e.g. a rejected Bash call), confirm `muxspect dock <block_id>` flags it `STUCK?`, and confirm `muxspect dock clear <block_id> <node_id>` actually clears it in the live UI without a reload. Flagging for reviewer/live-verification pass — this PR's own diagnostic value is best confirmed against a real stuck entry, not just unit tests.


a5af:
@codex review

PR #2434, opened 2026-08-06
docs: document task package MCP Shell invocation + retro

AgentA-asaf:
## Summary

While producing a fresh v0.54.11 desktop build, `task package` failed 6 consecutive times across three different agent-tool invocation mechanisms before a working path was found and verified. Full investigation: `docs/retro/retro-task-package-mcp-timeout-and-shell-output-gap-2026-08-06.md`.

Key findings:
- The Bash tool's timeout (not part of this repo — confirmed by reading `agentmux-bashwrap`'s source, which has no wall-clock cap beyond the documented 600s *idle* timeout) killed 3 attempts at the same point each time, even with an active heartbeat proving the idle-timeout specifically wasn't the cause.
- `nohup ... & disown` doesn't survive past the wrapping tool call — Windows Job Object membership isn't something POSIX detachment constructs can opt out of.
- `mcp__agentmux__Shell` has no way to retrieve a shell's output after the fact (`ShellStatus` only returns `running`/`exit_code`/`line_count`) — confirmed by reading `shell_node.rs`.
- **The actual fix already existed**: `scripts/package-agent.cmd`, built for exactly this, was referenced nowhere except itself.

## Changes

- `CLAUDE.md`: new "Launching `task package` from an agent / MCP Shell (Windows)" section, mirroring the existing `task dev` section, documenting the correct invocation (`package-agent.cmd` directly, output redirected to a file, since `ShellStatus` can't return content).
- `scripts/package-agent.cmd`: cross-links back to CLAUDE.md and the retro.
- New retro doc with full investigation, evidence, and fix plan.

## Test plan

- [x] The fix itself was verified live during this same session — `package-agent.cmd` invoked via `mcp__agentmux__Shell` with output redirected to a file produced a genuinely running (not instant-fail) build, readable via `Read` on the log file. Cited in the retro's "What Worked" section.

### DIFF
--- a/.changesets/1786043924-docs-document-task-package-mcp-shell-invocation-retro-bb1g.md
+++ b/.changesets/1786043924-docs-document-task-package-mcp-shell-invocation-retro-bb1g.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: document task package MCP Shell invocation + retro
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -73,6 +73,22 @@ grep "shell\." ~/.agentmux/logs/agentmuxsrv-*.log.$(date +%Y-%m-%d)
 
 See `docs/retro/retro-task-dev-agent-shell-path-2026-06-27.md` for full analysis.
 
+#### Launching `task package` from an agent / MCP Shell (Windows)
+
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2435, opened 2026-08-06
fix(context-meter): seed Claude Sonnet 5 context window at 1M

Agent2-asaf:
## Summary
- `contextWindowForModel()` treated every model id matching `sonnet` as the beta-gated Sonnet 4.x default: seed at 200K, only promote to 1M once observed usage actually exceeds it (`context-1m-2025-08-07` beta behavior).
- Sonnet 5 has no such beta gate — it ships with a 1M window by default — so a freshly opened agent pane that hadn't yet pushed past 200K of context stayed stuck showing "200K" in the dock, while another pane running the identical model that had already exceeded 200K showed "1M". Same model, two different displayed context sizes.
- Fix: seed directly at 1M for Sonnet 5+ (detected via the version number following `sonnet-` in the resolved model id reported by the CLI stream). Sonnet 4.x and earlier keep the existing conservative seed-then-learn-up behavior, since those really do need the beta header for 1M.

## Test plan
- [x] Added unit tests for `contextWindowForModel("claude-sonnet-5")` → 1M, case-insensitivity, and the bare `"sonnet"` family-alias fallback staying conservative
- [x] Added unit tests for `learnContextWindow` seeding Sonnet 5 at 1M on first observation (no learning needed) and a mid-session model-switch case (Sonnet 4.6 learned-1M → Sonnet 5 stays 1M)
- [x] `npx vitest run frontend/app/store/agent-pane-state/context-window.test.ts` — 18/18 passed
- [x] `npx tsc --noEmit` — no new errors

### DIFF
--- a/.changesets/1786044997-fix-context-meter-seed-claude-sonnet-5-context-window-at-1m--l4np.md
+++ b/.changesets/1786044997-fix-context-meter-seed-claude-sonnet-5-context-window-at-1m--l4np.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(context-meter): seed Claude Sonnet 5 context window at 1M (no beta gate)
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2425, opened 2026-08-05
feat(agent): wire in-app OAuth login into credential-loss relogin, retiring AuthUrlBox

AgentA-asaf:
## Summary

Closes the last of the three surfaces `SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03.md` originally planned (§3.3 surface 2) — its own phasing explicitly listed this as a separate PR that was never built. Until now, a mid-session credential-loss relogin (an agent pane whose auth died) rendered the older, separate `AuthUrlBox` component inline in the transcript.

**Design update (2026-08-07):** originally built as a `Modal`-wrapped `InAppLoginPanel`, matching the Armory/Stash surfaces. Based on user feedback while live-testing #2429 ("the login UI needs to be pinned to the bottom, like question/answer"), switched to rendering `InAppLoginPanel` directly inside `AgentAuthPanel`'s bottom-docked slot (`agent-view.tsx`, introduced by #2436) instead — same flex-sibling band as `AgentDecisionPanel`/`AgentQuestionPanel`, no Modal/backdrop-dimming. Still reuses the shared `InAppLoginPanel` component (one fewer hand-rolled auth UI), just without the floating-dialog presentation. Rebased onto main after #2436 merged to pick up `AgentAuthPanel`.

`AuthUrlBox` and its dead CSS (`frontend/app/view/agent/styles/_activity-log.scss`) are fully retired — addresses the reagentx P2 finding from the original review.

Added `toInAppLoginPhase` (`frontend/app/view/agent/components/to-in-app-login-phase.ts`) to derive `InAppLoginPanel`'s phase enum from `useAgentControllerStatus.ts`'s existing `authUrl`/`launchPhase` signals, without touching that hook's own internal state tracking — purely additive.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run` on `login.test.ts`, `useAgentControllerStatus.test.ts`, `to-in-app-login-phase.test.ts` — 30/30 passing
- [x] Confirmed zero remaining code references to `AuthUrlBox`/its CSS classes (only stale prose comments remain, same as originally scoped)

PR #2436, opened 2026-08-07
fix(login): make tier-1 PTY capture actually work end-to-end

### LINKED ISSUE #2429: Tier-1 Claude in-app login: no PTY output captured under ConPTY despite correct binary resolution
## Summary

PR #2422 fixed a confirmed P0 (raw `.cmd` npm shim spawned via `portable_pty::CommandBuilder` hangs indefinitely under Windows ConPTY — 0% CPU, zero output, target config dir never even created). That fix is merged and correct as far as it goes: `run_cli_login_pty` now resolves through `agentmux_common::resolve_cli_spawn_target` and spawns the real `claude.exe` directly instead of the `.cmd` shim.

However, live verification (2026-08-05, on a fresh `task dev` build with the fix applied) shows tier-1 URL capture **still does not work end-to-end** — it's a different, narrower bug than the one #2422 fixed.

## Reproduction

1. Trigger a fresh Claude Connect flow (Armory → Accounts → Connect → OAuth), which spawns `run_cli_login_pty` with a fresh isolated `CLAUDE_CONFIG_DIR`.
2. Host log shows: `spawned (PTY), waiting for OAuth URL` → 15s later → `no auth URL captured within 15s` → child killed cleanly.
3. **Zero `[login-pty]` output lines are logged during the entire 15s window** (the code logs every non-empty PTY line from the first byte — see `cli_login.rs`'s PTY reader task). This means the spawned process produced no visible output at all, not that a URL line failed to parse.
4. Reproduced twice, consistently, same result both times.

## What's confirmed NOT the problem

- **Shim resolution is correct.** `ResolveCliCommand` returns `cli_path = ...\node_modules\.bin\claude.cmd`; that file's content matches the native-exe shim shape; `resolve_cli_spawn_target` resolves it to the real `claude.exe` at `...\node_modules\@anthropic-ai\claude-code\bin\claude.exe` — confirmed that file exists.
- **The binary and args are correct and functional.** Running the exact resolved binary directly — `claude.exe auth login` with `CLAUDE_CONFIG_DIR` set to a fresh dir, no ConPTY at all (plain process spawn) — prints the OAuth URL (`Opening browser to sign in… If the browser didn't open, visit: https://claude.com/cai/oauth/authorize?...`) in well under a second, then prompts `Paste code here if prompted >`. Exactly the behavior `PROVIDERS['claude']`'s catalog doc comment describes.
- **The child process does start and run** (not hung/stuck) — `auth.start (direct-account)` logs show the isolated config dir gets wired, and the process shows up in `tasklist` with growing memory during the window (though this specific signal is noisy on a busy dev machine with many other Claude sessions running).

## Leading hypothesis

**stdout buffering.** Many CLIs (this one is Node/V8-based under the hood per its shim) switch from line-buffered to fully-buffered stdout when `isatty()`/console-detection reports the output isn't a real interactive terminal. A `portable_pty`-allocated Windows ConPTY handle may not satisfy whatever check `claude.exe` uses to decide "this is an interactive terminal," causing it to buffer output that would otherwise flush immediately — output would then only appear on a full buffer flush or process exit, both of which lose the race against the 15s capture timeout.

## Suggested next steps

- Check whether `portable_pty`'s Windows ConPTY allocation reports as attached-to-a-console in whatever way `claude.exe`/Node checks (e.g. `GetConsoleMode`/`isatty`) — compare against a plain `CreateProcess` + real conhost.
- Try env vars that might force unbuffered/TTY-like behavior in the child (`FORCE_COLOR=1`, `CI=` unset if set, etc.) as a diagnostic, not necessarily the fix.
- Check `portable_pty` crate version/changelog for known Windows ConPTY buffering quirks.
- Add a diagnostic: log PTY reader activity (bytes-read count, not just parsed lines) so a future repro can distinguish "zero bytes received" from "bytes received but not treated as a line yet."

## Impact

Not a regression — tier-2 (seed from existing global login) and tier-3 (terminal fallback) still work and mask this gracefully for any user who already has a global Claude login. The specific capability still missing is: a **brand new** in-app Claude login with **no pre-existing credentials anywhere** on the machine, fully self-contained (SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03.md's stated goal).

AgentA-asaf:
## Summary

Fixes #2429. Tier-1 in-app PTY capture for Claude CLI login was completely non-functional; this closes it out end-to-end, plus two follow-on bugs discovered while live-verifying the fix.

1. **Root cause — nothing was ever captured.** Claude Code, on detecting a real TTY, immediately sends a cursor-position query (`ESC[6n`) and blocks waiting for a reply before printing anything. Nothing was answering it, so tier 1 always timed out at 15s with zero bytes seen and silently fell back to the terminal tiers. `run_cli_login_pty` now transparently answers this query the moment it appears in the byte stream via a `DsrRespondingReader` wrapper. Confirmed live: URL capture now takes ~450ms instead of timing out.

2. **Follow-up — captured URL was missing `client_id`.** Once capture started working, live testing surfaced "Invalid OAuth Request — Missing client_id parameter" in the browser. Root cause: `extract_url()` preferred the CLI's plain-text fallback line ("If the browser didn't open, visit: ..."), which the CLI's terminal-width-aware wrapping splits mid-query-string at the PTY's 80-column width — truncating the URL before `client_id` ever appears. The OSC-8 hyperlink payload carries the complete, unwrapped URL the whole time (it's inside an escape sequence, not printed/wrapped text); `extract_url()` now prefers it, falling back to the plain text only for CLIs that don't emit OSC-8.

3. **Follow-up — login UI pinned to the top of the pane.** Also surfaced by the same live test. The auth-URL box rendered inside `AgentDocumentView`'s scrollable header slot, which pinned it to the top of the scroll area instead of near the composer. Extracted into a new `AgentAuthPanel` component, now rendered by `agent-view.tsx` as a flex sibling directly after the scroll region — the same slot band as `AgentDecisionPanel`/`AgentQuestionPanel`, so it's bottom-docked the same way.

## Test plan

- [x] `cargo test -p agentmux-cef cli_login` — 25/25 passing, including a new regression test for the DSR reader and one reproducing the exact wrapped-line URL truncation
- [x] `npx vitest run` on the affected login/auth-status test files — 24/24 passing, no regressions
- [x] Live-verified DSR fix in a dev instance: URL capture completes in ~450ms instead of timing out
- [ ] Live re-verification of the client_id + UI-positioning follow-ups pending a fresh manual click-through (structural fix confirmed via tests + exact pattern parity with AgentQuestionPanel's proven-working bottom-docked layout)

Note for reviewers: PR #2425 (`agenta/relogin-inapp-modal`) also touches this exact render site (retiring `AuthUrlBox` for a Modal-wrapped `InAppLoginPanel`). Once this merges, #2425 will be rebased to render `InAppLoginPanel` inside this PR's `AgentAuthPanel` bottom-docked slot instead of a Modal, per user direction — chosen over the Modal so the login UI stays consistent with `AgentQuestionPanel`'s inline, non-blocking bottom-docked pattern.

### DIFF
--- a/.changesets/1786054204-fix-login-answer-the-cli-s-cursor-position-query-so-tier-1-p-g7wn.md
+++ b/.changesets/1786054204-fix-login-answer-the-cli-s-cursor-position-query-so-tier-1-p-g7wn.md
@@ -1,0 +1,13 @@
+---
+type: patch
+---
+
+fix(login): make tier-1 PTY capture actually work end-to-end
+
+Fixes #2429, in three parts:
+
+1. **Root cause — the CLI never printed anything.** Claude Code, on detecting a real TTY, immediately sends a cursor-position query (`ESC[6n`) and blocks waiting for a reply before printing anything — nothing was ever answering it. `run_cli_login_pty` now transparently answers this query the moment it appears in the stream, unblocking the CLI. Confirmed live: URL capture now takes ~450ms instead of timing out at 15s with zero bytes seen.
+
+2. **Follow-up — the captured URL was missing `client_id`.** Once capture started working, `extract_url()` preferred the CLI's plain-text fallback line, which gets word-wrapped by the PTY's 80-column width mid-query-string. The OSC-8 hyperlink payload (immune to that wrapping, since it's inside an escape sequence, not printed text) carries the complete URL the whole time — `extract_url()` now prefers it.
+
+3. **Follow-up — the login UI rendered pinned to the top of the pane.** It was rendered inside `AgentDocumentView`'s scrollable header slot. Extracted into `AgentAuthPanel`, now rendered by `agent-view.tsx` as a flex sibling after the scroll region — same bottom-docked slot band as `AgentDecisionPanel`/`AgentQuestionPanel`.
--- a/agentmux-cef/src/commands/cli_login.rs
+++ b/agentmux-cef/src/commands/cli_login.rs
@@ -575,6 +575,96 @@ fn spawn_login_pty_unix(
 ///   - Unix: `spawn_login_pty_unix` — portable_pty's spawn is UNUSABLE in
 ///     this host because its pre-exec fd cleanup trips libcef's interposed
 ///     close() and crashes the child before exec (see that fn's doc).
+
+/// Device Status Report cursor-position query — some CLIs (confirmed:
+/// Claude Code, on detecting a real TTY) send this immediately and BLOCK
+/// waiting for the terminal to reply with `ESC[<row>;<col>R` before
+/// printing anything else at all. A bare `portable_pty` handle has no
+/// attached terminal emulator to answer this automatically — nothing
+/// does, on any platform, without code specifically watching for it — so
+/// the child hangs forever and the whole capture loop below times out
+/// having seen zero bytes. This is the confirmed root cause of issue
+/// #2429 ("no PTY output captured... despite correct binary resolution"):
+/// isolated repro (a standalone portable_pty harness spawning the exact
+/// same binary/args) showed the child's first and ONLY output was these
+/// 4 bytes, then silence: answering this one query unblocked it
+/// immediately, and it printed its OAuth URL within half a second.
+const DSR_CURSOR_POSITION_QUERY: &[u8] = b"\x1b[6n";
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2437, opened 2026-08-07
fix(agent-pane): remove meaningless Resumed/Ready transcript notifications

AgentX-asaf:
## Summary

Per user request: remove the AgentMux-generated messages that show up in the agent pane looking like they came from the agent but actually say nothing.

- `"Resumed — continuing where you left off"` and `"Ready — type a message to start"` are posted via `postSystemNotification` → `StreamFlush` on every pane mount — the exact same path real streamed CLI output goes through, landing as permanent nodes in the actual transcript, not a toast/banner. Both use `"info"` style, which (per the code's own comment) renders with **no visual prefix**, unlike `"success"`/`"warning"`'s `✓`/`⚠` glyph — so on screen they're indistinguishable from something the agent actually said, and they narrate nothing the user couldn't already see (existing history, or an empty composer right below).
- Removed both `notify()` calls. Kept the `resumed-ready`/`fresh-ready` `setPhase()` transitions untouched — `formatPhaseLabel` already renders no footer label for either phase today, so the notification was the only user-visible effect of the branch, not the branch itself.
- Same "no artificial messages mixed into transcript content" rule as PR #2420's earlier removal (the "Picked up more work" notification / settled-grace).
- Left the resync-failed warning and the auth-expired/first-login `notify()` calls alone — they're either visibly `"warning"`-styled or convey real actionable info (you need to log in) the user doesn't otherwise have.

Full audit of every AgentMux-synthesized string in the agent pane — what's meaningless narration vs. what's real information, and why — is in `docs/specs/REPORT_AGENT_PANE_SYNTHESIZED_TEXT_AUDIT_2026_08_06.md`. This PR implements the narrow slice confirmed with the user; the report also flags a broader "ephemeral chrome" category (auth banners, failure-recovery row, rate-limit labels) that's explicitly out of scope here since those never touch the transcript and back real bug fixes.

## Test plan

- [x] Updated the two `launch-flow.test.ts` cases that asserted on the removed notification text — they now assert `notices` stays empty while the `resumed-ready`/`fresh-ready` phase transition still fires correctly.
- [x] `npx vitest run frontend/app/view/agent/flows/launch-flow.test.ts` — 16/16 passing
- [x] `npx vitest run frontend/app/view/agent/` — 995/996 passing (the 1 failure is `tool-renderers/registry.test.ts`'s unrelated 5s-timeout flake under full-suite parallelism; passes cleanly in isolation, unrelated to this change)

### DIFF
--- a/.changesets/1786069930-fix-agent-pane-remove-meaningless-resumed-ready-transcript-n-kdj1.md
+++ b/.changesets/1786069930-fix-agent-pane-remove-meaningless-resumed-ready-transcript-n-kdj1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): remove meaningless Resumed/Ready transcript notifications
--- a/docs/specs/REPORT_AGENT_PANE_SYNTHESIZED_TEXT_AUDIT_2026_08_06.md
+++ b/docs/specs/REPORT_AGENT_PANE_SYNTHESIZED_TEXT_AUDIT_2026_08_06.md
@@ -1,0 +1,340 @@
+# Report: Audit of AgentMux-Synthesized Text in the Agent Pane
+
+**Date:** 2026-08-06
+**Author:** AgentX (agent)
+**Status:** Research only — no code changed. This catalogs everything found so a
+scope decision can be made before any removal work starts.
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2438, opened 2026-08-07
fix(swarm): stop rendering a phantom "Agent" row for a block that doesn't exist

GenericAgentX-asaf:
## Summary

While live-testing the Swarm view this session, the user hit two related symptoms:
- A row displaying the literal placeholder name **"Agent"** — not a real agent — confirmed independently on a **second AgentMux instance, different OS** (rules out local leftover state as the sole explanation).
- Top-level rows whose only visible content is **"No activity yet"**.

Root cause (full writeup in `docs/retro/RETRO_SWARM_PHANTOM_ROWS_AND_STALE_TRACKING_2026_08_06.md`): `buildTree()` adds every subagent's `parent_block_id` to the tree's row-id set unconditionally (a registration-ordering fallback), even when no WOS block object exists for that id at all — e.g. a parent block whose registration never completed, or one that was pruned while a subagent record referencing it lingered. `prune_block` only fires on `BlockDeleted`/`TabDeleted`/`WorkspaceDeleted`; a block that never finished registering was never "deleted," so nothing cleans up the orphaned reference.

That phantom id then fell back to the `"Agent"` placeholder string (a fallback that's actually meant for "block exists but `agentName` hasn't propagated to meta yet" — a different, legitimate transient state, unchanged by this PR) and rendered as its own top-level row. Any dispatch grouped under it — genuinely zero events, since the parent never really existed — shows as an empty "No activity yet" entry beside it. Same bug, two symptoms.

## Changes

- New `hasRenderableBlock()` predicate in `swarm-model.ts` (mirrors the existing `buildShellRows`/`buildCronRows` pattern of extracting pure, directly-unit-testable helpers rather than inlining logic in `buildTree()`).
- `buildTree()` now skips a row entirely (`flatMap` + `[]`) when its block doesn't resolve to a real object, instead of falling through to the `"Agent"` placeholder.
- 3 new unit tests for `hasRenderableBlock`.
- Also included: the retro (root-cause investigation) and a separately-scoped spec (`SPEC_SWARM_ROW_AUTO_LINGER_COUNTDOWN_2026_08_06.md`) for an auto-expiry countdown on completed rows the user also flagged this session — not implemented yet, sequenced to land after this fix per the spec's own note (arming a countdown on a still-phantom row would just delay the bug by 60s).

## What this does NOT fix

This is the frontend-side defensive filter only. The retro's open items — live-reproducing the exact orphaned `parent_block_id`, confirming why it reproduces across two independent instances, and a backend-side cleanup path for blocks that never finish registering — are still open. This PR stops the phantom row from being *visible*; it doesn't address why the orphaned record exists in the backend in the first place.

## Test plan

- [x] All existing `swarm-model.test.ts` (54) + `swarm-view.test.ts` tests pass unchanged.
- [x] 3 new tests for `hasRenderableBlock` (undefined, null, real-block-with-empty-meta).
- [ ] Live verification: reproduce the original phantom-row report and confirm it no longer renders (I couldn't do this myself — I don't have live access to either instance where it was observed).

### DIFF
--- a/.changesets/1786071566-fix-swarm-phantom-agent-rows-e186afd3.md
+++ b/.changesets/1786071566-fix-swarm-phantom-agent-rows-e186afd3.md
@@ -1,0 +1,15 @@
+---
+type: patch
+---
+
+fix(swarm): stop rendering a phantom "Agent" row for a block that doesn't exist
+
+`buildTree()` added every subagent's `parent_block_id` to the swarm tree's row
+list unconditionally, even when no WOS block object exists for that id (a
+parent block whose registration never completed, or was pruned while a
+subagent record referencing it lingered). That row rendered as a placeholder
+named literally "Agent," and any dispatch grouped under it showed as an empty
+"No activity yet" entry beside it. `buildTree()` now skips a row entirely when
+its block doesn't resolve to a real object, via the new `hasRenderableBlock()`
+predicate — distinct from the pre-existing "Agent" fallback for a real block
+whose `agentName` meta just hasn't propagated yet, which is unchanged.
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2439, opened 2026-08-07
fix(agent-pane): continuous Working row background through the scrollbar gutter

Agent3-asaf:
## Summary

- The agent pane's "Working…"/"✓ Worked" status row (`AgentWorkingRow`) has an opaque background that visibly stopped short of the pane's right edge, right where the message list's scrollbar lives — most noticeable scrolled away from true bottom.
- Root cause: `.agent-working-row-anchor` is deliberately inset from the right edge by `--agent-document-scrollbar-width` (14px) — that inset is the fix for a *real earlier bug* where the row's opaque background painted over (hid) `.agent-document`'s native scrollbar, since a custom-styled `::-webkit-scrollbar` renders as part of its owning element's own box, and the anchor's `z-index: 2` sat above `.agent-document`'s `z-index: auto`. But nothing ever filled that gutter with a matching color, and the scrollbar track itself is `background-color: transparent` (`theme.scss`), so the gutter showed the plain pane background instead of continuing the row's color.
- Fix: a purely decorative full-width backdrop, stacked *below* `.agent-document` (via a real stacking context on `.agent-document-scroll-region` + DOM order) so the scrollbar still paints on top of it, while the row's color now reaches the pane's true right edge. The existing text-bearing anchor is unchanged — same inset, same z-index, same click-through behavior.
- Full stacking-order analysis in `docs/specs/SPEC_AGENT_WORKING_ROW_SCROLLBAR_GAP_2026_08_06.md`.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] Live-verified in a `task dev` session: started a turn, scrolled the conversation away from true bottom while streaming — the row's background now reads as one continuous bar to the pane's right edge, scrollbar still visible/clickable on top
- [ ] Reviewer: also worth a quick look at the `--worked` (completed) state and at true-bottom scroll position

### DIFF
--- a/.changesets/1786072136-fix-agent-pane-continuous-working-row-background-through-the-wtzo.md
+++ b/.changesets/1786072136-fix-agent-pane-continuous-working-row-background-through-the-wtzo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): continuous Working row background through the scrollbar gutter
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2440, opened 2026-08-07
feat(swarm): auto-retire a finished row after a 60s countdown

GenericAgentX-asaf:
## Summary

Implements `SPEC_SWARM_ROW_AUTO_LINGER_COUNTDOWN_2026_08_06.md` (written earlier this session after a live-spawned subagent's row was expected to auto-expire and didn't — landed on top of #2438's phantom-row fix, since this spec explicitly assumed that lands first).

Before this PR: a completed Agent Tool row or Workflow row lingered in the Swarm tree indefinitely until a human manually clicked Retire. A finished 12-second subagent and a finished 6-hour workflow behaved identically.

## Changes

- **`swarm-model.ts`**: `reconcileCountdowns()` arms a 60s countdown (via the existing `retireRow()`/`_retiredRowKeys` un-retire-on-new-activity mechanism) for any subagent with `status === "completed"` or Workflow dispatch with `status === "completed"` — called after every subagent/dispatch reload, same timing as the existing `pruneRetiredRowKeys()`. `pauseCountdown`/`resumeCountdown` support hover-to-pause. `AUTO_RETIRE_DELAY_MS` is exported so the view computes against the same constant the timer actually runs on.
- **`swarm-view.tsx`**: `SubagentRow`/`WorkflowDispatchRow` render a live "disappearing in Ns" label next to the status chip/badge, and wire `onMouseEnter`/`onMouseLeave` to pause/resume. The visible countdown's 1s re-render reuses the existing `useTick(1000)` hook (ref-counted, auto-cleanup) rather than a second ViewModel-owned interval — simpler than what the spec's first draft proposed, noted in the spec's status line.
- **Excluded**: `"interrupted"` (abandoned/failed) rows never auto-count-down — matches `canRetire`'s existing scope, so a failure can't silently vanish before being read.
- **Excluded**: `shellRows`/`cronRows` — out of scope per the spec, `agentToolRows`/`workflowRows` only.

## Design decision worth a second look

Per the spec: on mouse-leave, the countdown resumes with a **fresh 60s**, not the remaining time from before the hover. This was the spec's stated default ("simpler to reason about... matches how most hover-to-pause toast patterns behave") but flagged as worth confirming rather than assumed. Flagging here too in case reviewers feel a resume-from-remaining model is preferable — it's a small, isolated change if so (`resumeCountdown` is the only place this would need to move).

## Test plan

- [x] All existing `swarm-model.test.ts` (60) + `swarm-view.test.ts` (10 prior) pass unchanged.
- [x] 5 new tests for the exported, pure `countdownSecondsRemaining` (arm instant, counts down, floors at 0, ignores unrelated rows, null when not counting down) — 65 total across both files.
- [x] `stylelint` on the touched scss file — confirmed the 12 pre-existing errors are unchanged/unrelated (same count on main before this PR).
- [ ] Live verification: spawn a subagent, watch it complete, confirm the countdown appears and the row disappears at 0; hover-pause; trigger new activity mid-countdown to confirm it clears. I don't have live access to run this myself.

### DIFF
--- a/.changesets/1786075165-feat-swarm-row-auto-linger-countdown-490473ee.md
+++ b/.changesets/1786075165-feat-swarm-row-auto-linger-countdown-490473ee.md
@@ -1,0 +1,17 @@
+---
+type: minor
+---
+
+feat(swarm): auto-retire a finished row after a 60s countdown
+
+A completed Agent Tool or Workflow row in the Swarm tree used to linger
+indefinitely until manually dismissed — a finished 12-second subagent and
+a finished 6-hour workflow behaved identically, cluttering the "what's
+happening now" view with things that are no longer happening. A clean
+terminal row (idle subagent, or a fully-completed Workflow dispatch) now
+shows a live "disappearing in Ns" countdown and auto-retires at 0, reusing
+the existing retire/un-retire-on-new-activity mechanism. Hovering a row
+pauses its countdown so reading a just-finished row's output doesn't race
+it disappearing; manual dismiss still works immediately at any point.
+Failed/interrupted rows are unaffected — they still require an explicit
+dismiss, so a failure never silently vanishes before being read.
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2441, opened 2026-08-07
feat(agent-pane): AskUserQuestion auto-selects the recommended option after 30s

Agent2-asaf:
## Summary

An unanswered `AskUserQuestion` tool call blocks the agent's turn indefinitely today — there's no timeout anywhere in that path. This adds a 30s auto-timeout with a visible countdown so a pending question can never stall work forever.

- **Recommended-option detection.** No such flag exists in the wire schema, so it's heuristic: a `"(Recommended)"` suffix on the option label (Claude Code's own documented convention for this tool), falling back to the first listed option when nothing is flagged.
- **Merge at timeout, not disarm-on-interaction.** An earlier design disarmed the whole timer the moment the user touched anything — rejected during spec review because it directly undercuts the goal: a user who answers question 1 of a 2-question set and steps away would otherwise cancel the safety net entirely, leaving question 2 blocked forever. Instead, the timer always fires at 30s; at zero it only fills in whichever questions are still genuinely untouched, leaving anything the user already answered (even partially) exactly as-is.
- **Countdown UI** in both the full panel header and the minimized chip, with color escalation (`--warning-color` at ≤10s, `--error-color` + a subtle pulse at ≤5s) using the app's existing severity tokens. Recommended option(s) get a dashed highlight so a watching user can predict the outcome.
- **Audit trail, no wire-format change.** The answer delivered to the agent is identical in shape to a manual one — same `answers_map`/`answer_text`, no backend change. A new `autoFilledCount` on `AnswerOutcome` only feeds the local transcript summary, distinguishing fully-manual / partly-auto / fully-auto so anyone reviewing later can tell how much of an answer was actually a human choice.

Spec (design decisions + rationale, including why the disarm-on-interaction design was rejected): `docs/specs/SPEC_ASK_USER_QUESTION_AUTO_TIMEOUT_2026_08_06.md`

## Test plan
- [x] New unit tests: `recommendedOptions()` (flagged/case-insensitive/multi-flag/fallback/empty), full-idle auto-submit, merge case (one question answered, one auto-filled), manual-submit-before-30s prevents any later auto-submit, countdown decrements exactly once per second to 0
- [x] `npx vitest run` — full suite: 2343 passed / 3 skipped, no regressions
- [x] Manually caught and fixed a real bug during testing: the Submit button's `onClick={submit}` was passing the click `PointerEvent` as `submit`'s new `autoFilledCount` parameter — fixed to `onClick={() => submit()}`

### DIFF
--- a/.changesets/1786077015-feat-agent-pane-askuserquestion-auto-selects-the-recommended-ier2.md
+++ b/.changesets/1786077015-feat-agent-pane-askuserquestion-auto-selects-the-recommended-ier2.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): AskUserQuestion auto-selects the recommended option after 30s
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -45,6 +45,7 @@ See also:
 | [SPEC_AGENT_GLOBAL_PORTABILITY_2026-06-16](SPEC_AGENT_GLOBAL_PORTABILITY_2026-06-16.md) | Agent definitions shared across workspaces |
 | [SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13](SPEC_CROSS_CHANNEL_AGENT_PERSISTENCE_2026-06-13.md) | Cross-channel conversation continuity |
 | [SPEC_ASK_USER_QUESTION_2026_06_15](SPEC_ASK_USER_QUESTION_2026_06_15.md) | `ask_user_question` tool |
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

Agent2-asaf:
Fixed in cbe41d4 — thanks for catching this.

Root cause confirmed: `applyRecommendedDefaults()` had nothing to select for a question with `options: []`, so it stayed unanswered, `submit()`'s `allAnswered()` gate silently no-op'd, and the interval was already cleared unconditionally before that check — no further retry possible.

Fix: `applyRecommendedDefaults()` now falls back to a free-text placeholder (`"No option was available to auto-select"`) when a question has no options to recommend, so every question is guaranteed answerable by the time the merged outcome submits. Added a regression test (`still auto-submits when a question has zero options...`) and a new row in the spec's edge-case table documenting this.

PR #2442, opened 2026-08-07
fix(login): widen login PTY so the OAuth URL never wraps, restyle the login card

AgentA-asaf:
## Summary

Follow-up to #2436. User live-tested the merged fix and reported the captured URL was still truncated (`https://claude.com/cai/oauth/authorize?code=t`) — identical to the pre-#2436 symptom.

**Root cause of the persisting bug:** my earlier fix in #2436 made `extract_url()` prefer an OSC-8 hyperlink URI over the wrapped plain-text line, on the assumption the Claude CLI emits one. A live capture from this session shows it does **not** — the actual output only contains an OSC-0 window-title sequence, never OSC-8. So that fix had nothing to prefer and kept falling through to the plain-text `"If the browser didn't open, visit: ..."` line, which the CLI's own renderer hard-wraps at the PTY's 80-column width, truncating the URL before `client_id` (and everything after it).

**The real fix:** widen the login PTY from 80 to 4096 columns, on both the Windows (`PtySize`) and Unix (`winsize`) spawn paths in `agentmux-cef/src/commands/cli_login.rs`. With no width to wrap against, the CLI prints the URL on one line and `extract_url()` captures it whole. The OSC-8 preference from #2436 is left in place as defense-in-depth (harmless, and would help if a CLI version ever does emit one), with its doc comment corrected to not assert something not actually observed.

**Also restyled the in-app login card** in the agent pane (user feedback: it looked unorganized compared to `AgentQuestionPanel`). Root cause: every other caller of `InAppLoginPanel` gets box chrome for free — wrapped in `.pre-launch-auth-panel` in the launch modal, or inside a `Modal` dialog for Armory/Stash — but the new bottom-docked `AgentAuthPanel` context (from #2436) rendered it completely bare. Added a card wrapper matching `AgentQuestionPanel`'s bordered/shadowed treatment, boxed each step ("Authorize in your browser" / "Paste the authorization code") as a distinct section, and changed the backup URL from single-line ellipsis-truncated text to a fully visible wrapped block — that URL is the whole point of the backup-link affordance, so it needs to actually be readable/copyable.

## Test plan

- [x] `cargo test -p agentmux-cef cli_login` — 25/25 passing
- [x] `npx vitest run` on the login/auth-status test files — 30/30 passing
- [x] **Live-verified in a rebuilt dev instance**: captured URL log shows zero wrapped continuation fragments (previously always 3-4 follow-up `[login-pty]` lines), and login completes end-to-end (`Login successful` → credentials written)
- [x] User confirmed the login flow and restyled card look correct

### DIFF
--- a/.changesets/1786089201-fix-login-widen-the-login-pty-so-the-oauth-url-never-h-x7q2.md
+++ b/.changesets/1786089201-fix-login-widen-the-login-pty-so-the-oauth-url-never-h-x7q2.md
@@ -1,0 +1,11 @@
+---
+type: patch
+---
+
+fix(login): widen the login PTY so the OAuth URL never hard-wraps, and restyle the in-app login card
+
+Two follow-ups to #2429/#2436:
+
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2443, opened 2026-08-07
fix(agent-pane): flush AskQuestion/disconnected-banner margins and remove scrollbar padding gap

Agent3-asaf:
## Summary

Three related visual-continuity fixes in the agent pane, found while verifying #2439 (the Working row scrollbar-gap fix):

1. **AgentQuestionPanel/AgentDisconnectedBanner top margin vs. the Working row.** Both panels carry `margin: <space> 0` for standalone breathing room. When `AgentWorkingRow` is visible directly above them, that top margin left a gap between the row's now-continuous color (#2439) and the panel's top border. Zeroed via `.agent-document-scroll-region + <panel>`, scoped to a new `.agent-view--working-row-visible` class (driven by the same `workingRowVisible`/`workingRowLoading` memos the row itself uses — hoisted once, reused for both) so a panel appearing on its own, without an active working row above it, keeps its normal margin.
2. **Same panels' bottom margin vs. the composer strip.** Left a gap before `AgentComposerStrip` when it's the very next thing below — unconditional this time, since the composer strip is always there. Uses `:has()` to target the panel's own margin from a "what's my next sibling" condition (an existing pattern in this codebase — `_pending-footer.scss`, `tabbar.scss`, etc.).
3. **`.agent-document`'s right padding vs. its own scrollbar.** 4px (`var(--space-1)`) of right padding left conversation content wrapping short of the scrollbar track, reading as a small gap between content and scrollbar. Removed (left padding kept for the pane's own left-edge breathing room).

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] All three verified live via `task dev` + Vite HMR, confirmed by the user before commit: AskQuestion panel flush against the Working row's blue above and the composer strip's stats bar below; conversation text now reaches the scrollbar with no gap

### DIFF
--- a/.changesets/1786093177-fix-agent-pane-flush-askquestion-disconnected-banner-margins-293h.md
+++ b/.changesets/1786093177-fix-agent-pane-flush-askquestion-disconnected-banner-margins-293h.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): flush AskQuestion/disconnected-banner margins and remove scrollbar padding gap
--- a/frontend/app/view/agent/agent-view.tsx
+++ b/frontend/app/view/agent/agent-view.tsx
@@ -1480,6 +1480,7 @@ const AgentPresentationView = ({ model, agentId }: { model: AgentViewModel; agen
         <div
             ref={rootRef}
             class="agent-view agent-view--presentation"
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2447, opened 2026-08-07
fix: resolve @emnapi/* package-lock.json drift breaking npm ci on CI

AgentA-asaf:
## Summary

Diagnostic PR — probing a fix for the `npm ci` EUSAGE failure hitting #2445/#2446 (and every future PR, since it's on `main`'s own lockfile).

## Root cause

`@tailwindcss/oxide-wasm32-wasi` is a `cpu: wasm32`-gated optional dependency of `@tailwindcss/oxide` that declares `@emnapi/core`, `@emnapi/runtime`, `@emnapi/wasi-threads` etc. as `bundleDependencies`. When `package-lock.json` is regenerated on Windows/macOS, npm correctly omits separate lockfile entries for these (they're vendored in the tarball, never needed off-wasm32). But `npm ci` on Linux (`ubuntu-latest`, Node 24) flags them as "Missing from lock file" anyway — a known cross-platform npm/napi-rs issue, see tailwindlabs/tailwindcss#20324.

This is not caused by the js-yaml or mermaid bumps in #2446/#2445 — reproduces identically on `main`'s current lockfile (`v0.54.12`).

This PR currently has a **temporary diagnostic step** in `.github/workflows/ci-pr.yml` that regenerates `package-lock.json` via `npm install` on the actual `ubuntu-latest` runner and uploads it as an artifact, since this can't be reproduced on Windows (verified: plain `npm install`, `npm install` with npm 11, and a from-scratch lockfile regen all fail to reproduce it locally). Once the artifact confirms the fix, the workflow diagnostic will be reverted and the corrected lockfile committed as the real fix, restoring plain `npm ci`.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -73,13 +73,16 @@ jobs:
         with:
           node-version: '24'
           cache: npm
+      # TEMPORARY, TRACKED in #2449 — do not let this become permanent.
       # `npm ci` fails on this platform with EUSAGE ("Missing: @emnapi/core... from
       # lock file") even though the lockfile is correct: @tailwindcss/oxide-wasm32-wasi
       # is a cpu:wasm32-gated optional dep that declares @emnapi/* as bundleDependencies
       # (vendored in its own tarball, never installed on real x64/arm64 hosts), and
       # npm ci's lockfile-sync validator incorrectly expects separate top-level entries
       # for them anyway. Known upstream bug: tailwindlabs/tailwindcss#20324.
-      # `npm install` doesn't run this overly-strict check and installs correctly.
+      # `npm install` doesn't run this check — it also won't fail on real
+      # package.json/package-lock.json drift, unlike npm ci. Revert to `npm ci`
+      # per #2449 once the upstream bug is fixed.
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

Agent2-asaf:
Landed a narrower fix in #2448 (merged) that keeps `npm ci` fully intact — pinned `@emnapi/core`/`@emnapi/runtime`/`@emnapi/wasi-threads` as explicit `devDependencies` at the exact versions `npm ci` was reporting missing, so they get real top-level lockfile entries and satisfy the validator directly, rather than switching the CI job to `npm install`.

Your root-cause diagnosis here (the `bundleDependencies`-vs-`dependencies` field cross-check in npm's `npm ci` validator, and the confirmation that regenerating the lockfile doesn't change its content) is exactly right and is what made #2448 possible without needing to reproduce this on Linux myself — thank you for digging into it. Closing this in favor of #2448 since reagent flagged a real concern with the `npm install` swap (permanently removes the lockfile-drift check with nothing enforcing the planned revert), and the more surgical fix is now merged and confirmed green on `ubuntu-latest`.

PR #2448, opened 2026-08-07
fix(ci): pin @emnapi/* devDependencies to fix npm ci EUSAGE on Linux

Agent2-asaf:
## Summary

`npm ci` fails on `ubuntu-latest` with:

```
npm error Missing: @emnapi/core@1.11.3 from lock file
npm error Missing: @emnapi/runtime@1.11.3 from lock file
npm error Missing: @emnapi/wasi-threads@1.2.3 from lock file
```

Confirmed reproducing on #2446 (dependabot js-yaml bump) — **not caused by that bump**; it's a pre-existing issue on `main` itself.

**Root cause** (diagnosis credit to #2447, which found this first): `@tailwindcss/oxide-wasm32-wasi` is a `cpu: wasm32`-gated optional dependency that declares `@emnapi/core`, `@emnapi/runtime`, `@emnapi/wasi-threads` as `bundleDependencies` (vendored inside its own tarball, never installed separately on real x64/arm64 hosts) — but it *also* lists them in its own `dependencies` field. `npm ci`'s lockfile-sync validator cross-checks that field and expects top-level lockfile entries regardless of the `bundleDependencies` exemption. This is a known upstream npm/napi-rs cross-platform quirk (`tailwindlabs/tailwindcss#20324`). #2447 also confirmed regenerating the lockfile doesn't help — a fresh install produces identical (correct) content either way, since the bug is in `npm ci`'s validator itself, not the lockfile content.

## Fix

Add the three exact packages as explicit `devDependencies`, at the same semver ranges `@tailwindcss/oxide-wasm32-wasi` already declares (`^1.11.1`, `^1.11.1`, `^1.2.2`). This makes them genuinely-resolved top-level dependencies with real lockfile entries, satisfying `npm ci`'s cross-check.

- **No CI workflow changes.** `npm ci` stays exactly as it is — its lockfile-drift/integrity protection is fully intact.
- **No functional impact.** These are tiny pure-JS shim packages; installing them as ordinary top-level devDependencies on non-wasm32 platforms is a no-op at runtime.
- Purely additive diff: `+62 -0` across `package.json` / `package-lock.json`.

This supersedes #2447's approach (`npm ci` → `npm install` in the workflow), which reagent correctly flagged (twice) as permanently removing the lockfile-drift check with no enforced follow-up to ever revert it. Closing #2447 in favor of this once CI confirms green here — full credit to its diagnosis, which is exactly right and made this fix possible without needing to reproduce the bug myself (I'm on Windows, where this doesn't reproduce, same as it didn't for #2447's author).

## Test plan
- [x] `npm install` locally regenerates `package-lock.json` with real top-level `node_modules/@emnapi/{core,runtime,wasi-threads}` entries at the exact versions `npm ci` was missing (verified byte-for-byte against the error message)
- [x] `npx vitest run` — full suite passes, no regressions
- [ ] **Needs CI on `ubuntu-latest` to confirm `npm ci` now passes** — can't reproduce the original failure on Windows, so this PR's own CI run is the real test

### DIFF
--- a/.changesets/1786124150-fix-ci-pin-emnapi-devdependencies-to-fix-npm-ci-eusage-on-li-q07e.md
+++ b/.changesets/1786124150-fix-ci-pin-emnapi-devdependencies-to-fix-npm-ci-eusage-on-li-q07e.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): pin @emnapi/* devDependencies to fix npm ci EUSAGE on Linux
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,6 +69,9 @@
         "ws": "^8.21.1"
       },
       "devDependencies": {
+        "@emnapi/core": "^1.11.1",
+        "@emnapi/runtime": "^1.11.1",
+        "@emnapi/wasi-threads": "^1.2.2",
         "@eslint/js": "^8.57.0",
         "@solidjs/testing-library": "^0.8.10",
         "@tailwindcss/vite": "^4.1.18",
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2454, opened 2026-08-07
fix(swarm): drop redundant agent-id tag on solo feeds, colorize ANSI in activity feed

AgentX-asaf:
## Summary

Investigated a report of "hex codes on nearly every line" in the Swarm view's expanded per-dispatch activity feed. Two issues found and fixed:

1. **Redundant agent-id tag.** `DispatchActivityFeedEntry` unconditionally prepends `entry.agentId.substring(0, 7)` to every entry. For the common case — a solo Agent Tool dispatch (one subagent per row) — every single line in the feed carries the exact same 7-char hex fragment, styled as unmarked 10px secondary-color text with no badge/tooltip. It reads as pure noise. The tag genuinely earns its keep in a multi-member Workflow dispatch's concatenated feed (it's the only thing saying which member a line belongs to), so the fix gates it on `showAgentTag = backfillAgentId === undefined` — the same discriminator `DispatchActivityFeed` already uses to distinguish solo vs. multi-member feeds — rather than removing it outright.
2. **No ANSI handling.** The feed's `text`/`tool_use`/`tool_result` bodies rendered raw `<pre>{text}</pre>` with zero ANSI-escape processing, unlike the Agent pane's own tool-result view (`CompactResult` → `TerminalOutput` → `AnsiLine`). A subagent's captured shell output (colorized `git diff`, a test reporter, etc.) has no ANSI stripping on the backend either (`subagent_watcher/parse.rs`), so it would render as literal escape-sequence text. Added a small `AnsiText` helper — `AnsiLine` per line, styled inline with the feed's existing `.swarm-subagent-detail-text` class rather than pulling in `TerminalOutput`'s bordered-panel chrome (which doesn't match this feed's compact list). `progress.output` deliberately left untouched — it's a short inline status label next to a spinner, not a captured command body, and `AnsiLine`'s per-line `<div>` would break that row's inline flex layout for no real benefit.

## Test plan

- [x] New `swarm-view.render.test.tsx` — 7 tests covering tag visibility (shown for multi-member, hidden for solo, consistent across all four entry kinds) and ANSI colorization (text entries, expanded tool_result preview, multi-line splitting, and confirming progress intentionally stays unrouted)
- [x] Verified with a throwaway mutation (swapped `AnsiLine` for a plain `<span>`) that the ANSI tests actually fail when the fix is broken, then reverted
- [x] `npx vitest run frontend/app/view/swarm/` — 74/74 passing

### DIFF
--- a/.changesets/1786127987-fix-swarm-drop-redundant-agent-id-tag-on-solo-feeds-colorize-7dlw.md
+++ b/.changesets/1786127987-fix-swarm-drop-redundant-agent-id-tag-on-solo-feeds-colorize-7dlw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(swarm): drop redundant agent-id tag on solo feeds, colorize ANSI in activity feed
--- a/frontend/app/view/swarm/swarm-view.render.test.tsx
+++ b/frontend/app/view/swarm/swarm-view.render.test.tsx
@@ -1,0 +1,124 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Component-level coverage for DispatchActivityFeedEntry — the Swarm view's
+ * per-dispatch activity feed row. Two fixes covered here:
+ *
+ *   1. The `showAgentTag` gate: a solo Agent Tool dispatch's feed only ever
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2455, opened 2026-08-08
feat(backend): add shared FsWatchPool filesystem-watcher framework

Agent2-asaf:
## Summary

Three of the four `notify`-based watchers in this codebase (`config_watcher_fs.rs`, `editor_file_watcher.rs`, `media_file_watcher.rs`) independently duplicate the same ~80 lines of plumbing — their own doc comments say so explicitly ("generalizes `config_watcher_fs.rs`'s pattern", "directory-mode sibling of `editor_file_watcher.rs`"). None of the four have any recovery if a watch fails to start or silently dies later — every one just logs a warning and gives up, with zero signal anywhere that live-update quietly stopped working.

Full audit + design rationale: `docs/specs/SPEC_SHARED_FS_WATCHER_FRAMEWORK_2026_08_07.md`.

`FsWatchPool` (new, `agentmux-srv/src/backend/fs_watch/`) owns the shared plumbing plus a genuinely new recovery layer:
- **Retry-with-backoff** on initial watch failure (3 attempts, ~4.2s bounded).
- **Automatic `PollWatcher` fallback** if the native backend won't construct at all.
- **Periodic self-healing sweep** (every 30s) that re-verifies every watched path is still live — a documented-safe no-op when nothing's wrong, so this recovers from a silent death without needing a true "is this watch dead" signal from the OS.
- **Queryable health** (`FsWatchHealth`: active backend, active watch count, degraded paths + last error) for future diagnostics.
- Bakes in the correctness-critical "watch the parent directory, not the file, so an atomic rename-based save is still detected" rule every existing implementer had to independently discover.

**Pure addition — no existing watcher migrated onto it yet.** Zero behavior change to anything live. Migration (config → editor+media → native-memory-as-a-new-consumer) is scoped as separate follow-up PRs per the spec's §5.

## Test plan
- [x] 7 new unit tests (refcounted subscribe/unsubscribe, file-subscription watches the parent dir, a real fs event is observable on the broadcast stream, health reports the active backend, recovery bookkeeping)
- [x] `cargo test --workspace -- --test-threads=1` (matches CI exactly) — 2045+ passed, 0 failed, no regressions

### DIFF
--- a/.changesets/1786157012-feat-backend-add-shared-fswatchpool-filesystem-watcher-frame-lgn0.md
+++ b/.changesets/1786157012-feat-backend-add-shared-fswatchpool-filesystem-watcher-frame-lgn0.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(backend): add shared FsWatchPool filesystem-watcher framework
--- a/agentmux-srv/src/backend/fs_watch/mod.rs
+++ b/agentmux-srv/src/backend/fs_watch/mod.rs
@@ -1,0 +1,33 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared filesystem-watcher framework.
+//!
+//! Before this module, three independent `notify`-based watchers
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2456, opened 2026-08-08
fix(backend): migrate config_watcher_fs onto the shared FsWatchPool

Agent2-asaf:
## Summary

First migration target per `SPEC_SHARED_FS_WATCHER_FRAMEWORK_2026_08_07.md` §5 — smallest/simplest of the three duplicated watchers, so it's the cheapest correctness check that `FsWatchPool`'s subscribe/events API is actually usable by a real caller, before migrating the more complex editor/media watchers in a follow-up PR.

- `spawn_settings_watcher` no longer constructs its own `notify::Watcher` — it subscribes to `settings.json` via `pool.subscribe_file(...)` and filters the pool's shared broadcast stream for that specific path, keeping its own domain-specific debounce/reload/broadcast logic exactly as it was.
- Return type dropped from `Option<RecommendedWatcher>` to `()` — the pool now owns watcher lifecycle end-to-end, so there's no handle left for `bootstrap.rs` to keep alive.
- `FsWatchPool` is now constructed once in `spawn_background_subsystems` and threaded through `BackgroundSubsystems`/`AppState` (as `fs_watch_pool`) so it's available to future migrations (editor/media watchers, and native memory as a new consumer) without re-threading.
- Added the first end-to-end test for this path — there wasn't one before: a real `settings.json` write on disk, detected via the pool, reloads into `ConfigWatcher`.

No behavior change from the user's perspective — settings.json live-reload works exactly as before, just via the shared framework instead of a bespoke watcher.

## Test plan
- [x] New end-to-end test: writing `settings.json` on disk is detected via `FsWatchPool` and reloads `ConfigWatcher` within the existing 300ms debounce window
- [x] New unit test for the filename-match helper
- [x] `cargo test --workspace -- --test-threads=1` (matches CI exactly) — 2047 passed, 0 failed, no regressions

### DIFF
--- a/.changesets/1786157802-fix-backend-migrate-config-watcher-fs-onto-the-shared-fswatc-zlu4.md
+++ b/.changesets/1786157802-fix-backend-migrate-config-watcher-fs-onto-the-shared-fswatc-zlu4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(backend): migrate config_watcher_fs onto the shared FsWatchPool
--- a/agentmux-srv/src/backend/config_watcher_fs.rs
+++ b/agentmux-srv/src/backend/config_watcher_fs.rs
@@ -3,16 +3,22 @@
 
 //! Filesystem watcher for settings.json — detects saves and pushes updated
 //! config to all connected WebSocket clients in real time.
+//!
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2457, opened 2026-08-08
feat(context-menu): Swarm copy menu, agent-pane paste menu, and app-wide fix for the same gap

AgentA-asaf:
## Summary

Two requested fixes, plus a scour of the app for the same class of gap:

- **Swarm right-click Copy menu** (previously nothing at all — rows are `user-select: none`, so there was no way to copy an agent's name or block ID out of Swarm).
- **Agent-pane composer right-click Paste menu** (previously hijacked by the pane body's generic Copy-on-selection-only handler, which no-ops with nothing selected — the common case when right-clicking to paste).

## Root cause (one bug, ~20 fixed spots)

`block/pane-actions.ts`'s pane-body context menu only offers Copy-on-selection and Paste for `view: "term"` panes. Any `<input>`/`<textarea>` living directly in a non-terminal pane's body (not inside a portalled `Modal`, which already worked correctly) had no way to paste via right-click.

## Fix

- Extracted `app.tsx`'s root Cut/Copy/Paste fallback logic into a shared, exported `showTextInputContextMenu()` in `store/contextmenu.ts` — single source of truth, `app.tsx` now uses it too instead of a local duplicate.
- Attached it directly to every affected input/textarea found in a full sweep: agent decision-panel deny-feedback, agent question-panel "Other" answer, browser address bar, skill/MCP/memory/brain manager forms, drone node-editor fields (task/URL/body/condition/template/variables/instance).
- Swarm's primary agent row gets a native `ContextMenuModel` Copy menu ("Copy agent name" / "Copy block ID"), following the existing `AgentPicker.tsx` per-row pattern.

Full gap inventory — including what's intentionally left for a follow-up pass (list/table copy-id polish on drone/warden/list-managers, and lower-value single-line inputs in settings/toolchain/editor/identity) — is in `docs/specs/REPORT_CONTEXT_MENU_GAP_AUDIT_2026_08_07.md`.

## Test plan

- [x] `npx vitest run` — 2370 passed, 0 failed, 3 skipped (no regressions)
- [ ] Manual click-through recommended: right-click the Swarm agent row, the agent-pane composer, and a couple of the manager forms to confirm the menus render correctly

### DIFF
--- a/.changesets/1786158722-feat-context-menu-swarm-copy-menu-agent-pane-composer-paste--7b7y.md
+++ b/.changesets/1786158722-feat-context-menu-swarm-copy-menu-agent-pane-composer-paste--7b7y.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(context-menu): Swarm copy menu, agent-pane composer paste menu, and shared Cut/Copy/Paste fix for every in-pane text input missing it
--- a/docs/specs/REPORT_CONTEXT_MENU_GAP_AUDIT_2026_08_07.md
+++ b/docs/specs/REPORT_CONTEXT_MENU_GAP_AUDIT_2026_08_07.md
@@ -1,0 +1,105 @@
+# REPORT — context-menu gap audit: Swarm copy, agent-pane paste, and beyond
+
+**Date:** 2026-08-07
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

PR #2458, opened 2026-08-08
fix(backend): migrate editor and media file watchers onto the shared FsWatchPool

Agent2-asaf:
## Summary
- PR3 in the FsWatchPool migration series (SPEC_SHARED_FS_WATCHER_FRAMEWORK_2026_08_07.md): moves `EditorFileWatcher` and `MediaFileWatcher` onto the shared pool.
- Both constructors now take `Arc<FsWatchPool>` and always return `Arc<Self>` (construction failure is absorbed into the pool's own health tracking rather than propagated as `None`) — `AppState.editor_file_watcher`/`media_file_watcher` are no longer `Option<Arc<...>>`.
- Per-path/per-dir subscribe/unsubscribe now delegates to `pool.subscribe_file`/`subscribe_dir`/`unsubscribe`; each watcher consumes its own `broadcast::Receiver` from `pool.events()`.
- Stacked on #2456 (PR2, `config_watcher_fs` migration) since this branch was cut from it — please merge #2456 first, or this will be rebased onto `main` automatically once that lands.

## Test plan
- [x] `cargo test --workspace -- --test-threads=1` — 2050 passed, 0 failed
- [x] Kept all pre-migration unit tests for both watchers; added a real end-to-end fs-write test per watcher asserting a `WaveEvent` is published via the shared pool

### DIFF
--- a/.changesets/1786159108-fix-backend-migrate-editor-and-media-file-watchers-onto-the--zhu6.md
+++ b/.changesets/1786159108-fix-backend-migrate-editor-and-media-file-watchers-onto-the--zhu6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(backend): migrate editor and media file watchers onto the shared FsWatchPool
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #2459, opened 2026-08-08
feat(backend): durable, cross-channel native memory mirror

Agent2-asaf:
## Summary
- Implements SPEC_NATIVE_MEMORY_DURABLE_SYNC_2026_08_07.md: makes native memory (Claude Code's own autonomous memory files, surfaced in Stash's Memory tab) durable and location-consistent across AgentMux channels/instances.
- Root cause: `agent:memory:list/read_file/write_file` only ever touched the live filesystem path, which is channel-relative by design (`agent.working_directory` resolves against the channel's own agents base) — so the same agent opened from two channels/builds resolves two different memory dirs, and appears to have no memory on whichever channel it wasn't last used from.
- Adds `db_agent_native_memory` (keyed by the stable `AgentDefinition.id`, not any live path) as a durable mirror — `SHARED_STORE_SCHEMA_VERSION` v6 / `OBJECT_SCHEMA_VERSION` v14, duplicated into both schemas the same way `db_agent_credentials` already is, for `id_store`'s pre-`0011_shared_store_backfill` fallback.
- All three RPC handlers write through into the mirror on every file they touch; `list`/`read_file` merge the live-FS view with the mirror (live wins when both have the file) so content written from one channel stays visible from another with no distinguishing UI treatment — the user never sees "not found on this channel."
- Also includes the status report + spec docs this was scoped from (written earlier in the same initiative, previously uncommitted).

## Test plan
- [x] `cargo test --workspace -- --test-threads=1` — all green (2047+ passed)
- [x] Store-level unit tests for upsert/read/list-meta (`agent_native_memory.rs`)
- [x] End-to-end RPC integration tests simulating two channels sharing one `id_store` but different live-FS paths for the same `agent_id`: a file written from channel A is visible (list + read) from channel B; live FS wins over a stale mirror row; read_file errors cleanly when absent from both

### DIFF
--- a/.changesets/1786160984-feat-backend-durable-cross-channel-native-memory-mirror-09og.md
+++ b/.changesets/1786160984-feat-backend-durable-cross-channel-native-memory-mirror-09og.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(backend): durable, cross-channel native memory mirror
--- a/agentmux-srv/src/backend/storage/agent_native_memory.rs
+++ b/agentmux-srv/src/backend/storage/agent_native_memory.rs
@@ -25,6 +25,16 @@ pub struct NativeMemoryMirrorRow {
     pub size_bytes: i64,
     pub updated_at: i64,
     pub last_seen_path: String,
+    /// The live file's own on-disk mtime (ms epoch) as of the last upsert —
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2460, opened 2026-08-08
fix(agent-pane): runtime dropup stays open on select for real, add close button

AgentX-asaf:
## Summary

Follow-up to `docs/specs/SPEC_AGENT_RUNTIME_DROPUP_CLOSE_BUTTON_2026_08_07.md`, which itself amends `docs/specs/SPEC_AGENT_RUNTIME_DROPUP_2026_07_09.md` §9.2. Two fixes for the Mode/Model/Effort runtime panel (`AgentRuntimeDropup`):

1. **Root cause of "closes right after I pick a value."** The panel's own §9.2 decision already says selection shouldn't close the panel, and `applySelection()` never calls `setOpen(false)` — but the option rows are plain non-focusable `<div>`s, so clicking one lets the browser's default mousedown behavior blur the trigger and shift `document.activeElement` to `<body>`. `handleFocusChange` treats that as "focus left the panel" and closes it — a second, independent close path the original §9.2 decision never accounted for. Fixed with `onMouseDown={(e) => e.preventDefault()}` on each row, the standard fix for this exact browser quirk in menu/listbox implementations.
2. **Added an explicit close (✕) button**, scaled down from `Modal`'s `showCloseButton` pattern to match this panel's compact chrome. Required splitting the single `role="listbox"` element into an outer positioned wrapper (holding the button) and an inner `role="listbox"` div (holding just the option rows) — dropping an interactive button in as a sibling of the `role="option"` rows would be an invalid listbox structure for assistive tech.

## Test plan

- [x] New `AgentRuntimeDropup.test.tsx` — this component had zero prior coverage. Covers: selection keeps the panel open, the close button works and sits outside the listbox subtree, click-outside still works.
- [x] The focus-blur fix needed a mechanism-level test (asserting `mousedown`'s `preventDefault` directly) since jsdom doesn't reproduce the real-browser blur-to-`<body>` quirk — confirmed via mutation testing (temporarily removed the fix) that the higher-level "stays open" tests alone do NOT fail without it, so they wouldn't have caught this bug on their own; the direct mechanism test does.
- [x] `npx vitest run frontend/app/view/agent/` — 1020/1020 passing

### DIFF
--- a/.changesets/1786161250-fix-agent-pane-runtime-dropup-stays-open-on-select-for-real--bg6y.md
+++ b/.changesets/1786161250-fix-agent-pane-runtime-dropup-stays-open-on-select-for-real--bg6y.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): runtime dropup stays open on select for real, add close button
--- a/docs/specs/SPEC_AGENT_RUNTIME_DROPUP_CLOSE_BUTTON_2026_08_07.md
+++ b/docs/specs/SPEC_AGENT_RUNTIME_DROPUP_CLOSE_BUTTON_2026_08_07.md
@@ -1,0 +1,320 @@
+# SPEC: Explicit close button on the Runtime (Mode/Model/Effort) dropup
+
+**Date:** 2026-08-07
+**Status:** Draft — ready for review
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2462, opened 2026-08-08
fix(backend): migrate editor and media file watchers onto the shared FsWatchPool

Agent2-asaf:
## Summary
- PR3 in the FsWatchPool migration series (SPEC_SHARED_FS_WATCHER_FRAMEWORK_2026_08_07.md): moves `EditorFileWatcher` and `MediaFileWatcher` onto the shared pool. Replaces the original #2458 (auto-closed when its stacked base branch was deleted on PR2's merge) — same content, rebased onto current `main`.
- Both constructors now take `Arc<FsWatchPool>` and always return `Arc<Self>` (construction failure is absorbed into the pool's own health tracking rather than propagated as `None`) — `AppState.editor_file_watcher`/`media_file_watcher` are no longer `Option<Arc<...>>`.
- Per-path/per-dir subscribe/unsubscribe now delegates to `pool.subscribe_file`/`subscribe_dir`/`unsubscribe`; each watcher consumes its own `broadcast::Receiver` from `pool.events()`.
- Restores the pre-migration Create/Modify-only event filter (a Removed event must not republish "this file changed" — reagent P1 x2 on the original #2458).

## Test plan
- [x] `cargo test --workspace -- --test-threads=1` — all green (2045+ passed, run with isolated AGENTMUX_DATA_DIR/CONFIG_DIR/SHARED_DIR/HOME_OVERRIDE/INSTANCE_DIR to avoid touching real data)
- [x] Kept all pre-migration unit tests for both watchers; added a real end-to-end fs-write test per watcher, plus a regression test confirming a Removed event does NOT trigger a publish

### DIFF
--- a/.changesets/1786159108-fix-backend-migrate-editor-and-media-file-watchers-onto-the--zhu6.md
+++ b/.changesets/1786159108-fix-backend-migrate-editor-and-media-file-watchers-onto-the--zhu6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(backend): migrate editor and media file watchers onto the shared FsWatchPool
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #2464, opened 2026-08-08
fix(identity): close two single-point-enforcement gaps (#2463)

### LINKED ISSUE #2463: Auth: single-point-enforcement has two live gaps — legacy identity_id sentinels FK-crash, blank identity_id bypasses the gate entirely
## Context

Pulled latest `main` (through `4ca646d3e`, which includes all of AgentA's recent auth work — #2431 isolated-auth-by-channel default, #2436 tier-1 PTY capture fix, #2442 login-PTY width fix, #2425 in-app OAuth wired into credential-loss relogin, #2423 ClaudeLoginPanel/Modal unification — plus the in-app OAuth login surfaces from #2410/#2413/#2414/#2415), built a fresh `task package:linux` portable (v0.54.12+g046e0cbae, channel `local-main-b28b7a-5e305ccf`), and live-tested it. Filing this because two real gaps in the "single point, not global" auth invariant (`PLAN_LOGIN_SINGLE_PATH_CONSOLIDATION_2026_07_20.md` §7) turned up immediately, and the second one produced contradictory results between a Windows tester and this Linux build for a reason that's worth documenting precisely.

Both are pre-existing gaps, not regressions from the pull — but they collided head-on with `SPEC_ISOLATED_AUTH_DEFAULT_BY_CHANNEL_2026_08_06.md` making isolated (empty) per-build account lists the default everywhere, so they're now hit on literally every fresh local build instead of only in edge cases.

## Finding 1 — legacy `identity_id` sentinels (`"default"`, stale bundle ids) crash the write-through and silently block spawn

Repro: relaunch a pre-existing named agent ("AgentU", agent definition `c438a823-cfbc-4be2-89e1-b8e3577c6bc2`, created 2026-06-30) in a fresh isolated-auth channel.

**What happens:**
1. `db_agent_instances.identity_id` for this agent's continuation rows is the literal string `"default"` — not a real account UUID. This is the "accepted migration gap" `agent-model.ts`'s write-through comment already documents (a pre-#1624-PR-C row that predates the account-linking system, carrying an old bundle id instead of a real account id).
2. On launch, the write-through (`frontend/app/view/agent/agent-model.ts:657-670`) calls `LinkAgentIdentityCommand({ agent_id, account_id: "default", provider })`. `db_agent_identity_links.account_id` has `FOREIGN KEY (account_id) REFERENCES db_accounts(id) ON DELETE CASCADE` (`agentmux-srv/src/backend/storage/migrations.rs:280-288`) — in a fresh isolated channel `db_accounts` is genuinely empty (0 rows, verified directly against `identity-store.db`), so this INSERT always throws `FOREIGN KEY constraint failed`.
3. The write-through only `console.warn`s and swallows the error ("best-effort, doesn't abort the launch"). No real link ever gets created.
4. The spawn gate (`agentmux-srv/src/identity/resolver/inject.rs:340`) then correctly refuses the spawn — `identity.spawn.blocked: no credentials for provider claude … spawn refused`.
5. User-visible result: the pane shows the generic **"Agent encountered an error"** — not the "No account linked — Sign in to link one" Auth-classified banner (`agentmux-srv/src/agents/failure.rs:200-218`, PR #2413) I'd expect. Haven't fully traced why that classification isn't reaching this specific failure (first-turn-of-a-fresh-channel vs. the mid-life credential-loss path #2413/#2425 actually targeted) — flagging as a loose end, not confirmed root-caused.
6. **Workaround that DOES work today:** opening this agent's Identity tab (cog → Identity) and clicking "Connect Claude account" (#2414) runs a real in-app login and creates a proper link, replacing the stale `"default"` reference. So the recovery path exists, it's just not what a user hits first — they hit the crash-on-relaunch before ever finding the recovery button.

**Ask:** should the write-through validate `account_id` actually exists in `db_accounts` before attempting the link (or check for known legacy sentinels like `"default"`/`"blank"`) and, on a stale reference, surface something actionable — ideally routing straight into the same in-app reconnect flow the Identity tab button already offers — instead of a swallowed console warning + an opaque backend-side spawn refusal?

## Finding 2 — a blank `identity_id` bypasses the gate entirely, before it's ever evaluated (environment-dependent, not actually enforced)

Repro: create a **brand-new** agent ("Mzop", definition `b13188b9-94f3-44d0-b195-be6b256e5953`, created this session, `use_ambient_login=0`) in the same fresh isolated channel, with no account selected (none exist to select — `db_accounts` is empty).

**What happens:**
1. `db_agent_instances.identity_id` ends up `''` (genuinely empty) for this instance.
2. `inject_identity_env_with_broker` (`agentmux-srv/src/identity/resolver/inject.rs:249-263`) short-circuits on `instance.identity_id.is_empty()` and returns `Ok(())` **before the layer-3 "must have a real bound account" gate is ever reached** — logged as `instance {id} has empty/blank identity_id — falling back to ambient creds. Legacy row or UI regression?`.
3. The controller spawns with whatever's ambiently on disk. On this Linux dev box, `~/.claude/.credentials.json` (a leftover, still-valid ambient credential from 2026-07-27, `subscriptionType: max`) was sitting there — so the agent worked immediately, with **nothing bound in Armory** for this channel.
4. The same code, tested on a clean Windows machine with no ambient credential file present, correctly appeared to *require* Armory — but only because there was nothing to fall back to, not because the gate blocked it. Same bypass, different observable outcome purely as a function of whether a stray ambient credential happens to exist on the test machine.

This directly contradicts the "single point, not global" invariant `inject.rs`'s own `gate_oauth_failure` comment describes (`use_ambient_login`'s escape hatch was supposed to be fully retired). The empty-`identity_id` short-circuit predates that retirement and was never folded into it — its own comment says the UI "no longer produces these for new launches… so seeing one here means either a legacy continuation row or a UI regression," but a genuinely brand-new agent launch with an empty local account list produces exactly this today, so it's not just legacy rows hitting it.

**Ask:** should `identity_id.is_empty()` (for a definition whose provider is genuinely oauth-class) route into the same layer-3 gate as everything else, rather than short-circuiting before it? That would make the Windows behavior the *guaranteed* behavior everywhere instead of an accident of a clean profile.

## Why this matters now specifically

`SPEC_ISOLATED_AUTH_DEFAULT_BY_CHANNEL_2026_08_06.md` made empty per-build Armory account lists the default for every non-`stable` channel specifically to force the login/relogin code paths to actually execute during routine testing. Both gaps above mean that goal is currently undermined for two different reasons: Finding 1 crashes instead of prompting login, and Finding 2 silently skips login entirely when an ambient credential happens to be present — so a tester can go a long way without ever actually exercising the in-app login surfaces this whole epic (#2410/#2413/#2414/#2415/#2425/#2436/#2442) was built to harden, depending entirely on which agents/machine they happen to test with.

AgentA-asaf:
Closes #2463's two confirmed findings.

## Finding 1 — legacy `identity_id` sentinels FK-crash the write-through

`AgentPicker.tsx`'s `handleReattach`/`handleFork` forwarded `row.identity_id` unfiltered as `accountId`, so a legacy sentinel (`"default"`, a stale bundle id) reached `LinkAgentIdentityCommand` and threw `FOREIGN KEY constraint failed` against the now-empty `db_accounts` table in an isolated-auth channel — silently swallowed by the write-through's `try/catch`, leaving the user staring at a generic "Agent encountered an error."

`looksLikeRealAccountId` already exists and is already used for exactly this at the two call sites in `useContinueOrNewMode.ts` and the auto-continue path in this same file (`AgentPicker.tsx:438`) — its own header comment literally describes this bug class ("Forwarding an unresolvable id like 'default' as an account_id causes a real backend failure"). It just never made it to `handleReattach`/`handleFork`. Guarded both, plus added the same check as defense-in-depth inside the write-through itself (`agent-model.ts`), so a future/missed call site fails safe instead of crashing.

## Finding 2 — blank `identity_id` bypassed the layer-3 auth gate entirely

`inject_identity_env_with_broker` short-circuited to `Ok(())` (silent ambient-credential fallback) for empty/`"blank"` `identity_id`, before the "must have a bound account" gate ever ran. A brand-new agent launched with no account selected got `identity_id: ""` and spawned on whatever ambient credential happened to exist on disk — with nothing bound in Armory. This is exactly why a Windows tester and a Linux tester saw contradictory results from the identical code: same bypass, different outcome purely as a function of whether a stray credential file happened to exist on that machine.

Removed the short-circuit. Empty/blank now falls through to the same gate a non-empty-but-unresolvable `identity_id` (e.g. `"default"`) already correctly went through — safe because `resolve_bindings_for_instance` keys off `definition_id`, not `identity_id`, so the fallthrough doesn't depend on what `identity_id` actually contains.

Together: a fresh isolated-auth channel now correctly and consistently requires signing in, instead of either crashing (Finding 1) or silently working depending on legacy row shape / ambient credential presence (Finding 2).

## Not addressed

The issue's third point — the resulting spawn refusal sometimes surfaces as a generic "Agent encountered an error" instead of the Auth-classified "No account linked" banner — was explicitly flagged by the issue's own author as unconfirmed/not root-caused, and I couldn't nail it down from static reading either. Traced as far as I could: `failure::classify()` exists, has the correct substring-matching logic for `SpawnGateError`'s `Display` text, and is wired into several call sites (`runner.rs`, `persistent.rs`, `health.rs`, `host_spawn.rs`) — but confirming which path this specific first-turn-of-a-fresh-channel case takes needs a live repro. Worth flagging: this fix makes that refusal path fire *more* often (fewer spawns silently bypass the gate now), so it's more likely to be hit going forward. Left as a follow-up rather than a speculative fix to auth-error-UX code.

## Test plan

- [x] `cargo test -p agentmux-srv` — 2062 passed, 0 failed, 5 ignored. Rewrote the test that encoded the old (buggy) "blank bypasses the gate" behavior into the corrected assertion, added a companion test for the empty-string case.
- [x] `npx vitest run` (targeted: identity-carry-over, AgentPicker, agent-model) — 17 passed, 0 failed

### DIFF
--- a/.changesets/1786190173-fix-identity-close-two-single-point-enforcement-gaps-legacy--ne7m.md
+++ b/.changesets/1786190173-fix-identity-close-two-single-point-enforcement-gaps-legacy--ne7m.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): close two single-point-enforcement gaps — legacy identity_id sentinel FK-crash and blank identity_id bypassing the layer-3 auth gate
--- a/agentmux-srv/src/identity/resolver/inject.rs
+++ b/agentmux-srv/src/identity/resolver/inject.rs
@@ -38,9 +38,14 @@ use super::secret::resolve_secret;
 /// 1. Look up the active `AgentInstance` for this block. If none
 ///    exists, the caller didn't go through the launch modal — return
 ///    immediately, no injection.
-/// 2. Read its `identity_id`. Empty / "blank" → no injection (the
-///    user picked the blank singleton at launch, meaning "use ambient
-///    creds").
+/// 2. Read its `identity_id`. Empty / "blank" no longer short-circuits to
+///    ambient creds (see #2463 — this used to bypass the layer-3 gate
+///    entirely, so whether a spawn required a bound account depended on
+///    whether a stray ambient credential happened to exist on the test
+///    machine). It now falls through to the same steps below, which gate
+///    purely on the agent definition's own provider — identical to how a
+///    non-empty-but-unresolvable `identity_id` (e.g. a legacy sentinel)
+///    already behaved.
 /// 3. Read the `db_agent_identity_links` rows for the instance's definition.
 /// 4. For each binding: fetch the account, resolve its `SecretRef`,
 ///    look up the provider's env-var matrix, write each var into
@@ -246,23 +251,29 @@ pub fn inject_identity_env_with_broker(
     };
 
     // Step 2: identity_id check.
+    //
+    // #2463 Finding 2: this used to `return Ok(())` here (silent ambient-
+    // creds fallback), which bypassed the layer-3 "must have a bound
+    // account" gate below entirely — a genuinely brand-new agent launched
+    // with no account selected got identity_id="" and spawned on whatever
+    // ambient credential happened to be sitting on disk, with nothing
+    // bound in Armory. Observed behavior therefore depended on whether the
+    // test machine happened to have a stray credential file, not on any
+    // actual policy decision. The UI no longer produces empty/"blank"
+    // identity_id for new launches (identity is required at submit-time —
+    // SPEC_LAUNCH_MODAL_STATE_MACHINE_2026_05_19.md), so seeing one here
+    // means a legacy continuation row or a UI regression; either way it
+    // must be gated the same as any other unresolvable identity_id rather
+    // than silently bypassed. `resolve_bindings_for_instance` below keys
+    // strictly off `instance.definition_id`, not `identity_id`, so falling
+    // through here is safe regardless of what identity_id contains.
     if instance.identity_id.is_empty() || instance.identity_id == "blank" {
-        // Empty or legacy "blank" sentinel → ambient creds (no
-        // injection). The UI no longer produces these for new
-        // launches (identity is now required at submit-time —
-        // SPEC_LAUNCH_MODAL_STATE_MACHINE_2026_05_19.md), so seeing
-        // one here means either a legacy continuation row or a UI
-        // regression. Warn so the regression is visible in logs.
-        // Deliberately NOT gated by layer 3: this sentinel predates the
-        // managed-account model and was an explicit "ambient creds"
-        // choice at launch time, not a silent fallback.
         tracing::warn!(
             target: "identity",
-            "instance {} has empty/blank identity_id — falling back to ambient creds. \
-             Legacy row or UI regression?",
+            "instance {} has empty/blank identity_id — falling through to the \
+             layer-3 gate instead of ambient creds. Legacy row or UI regression?",
             block_id
         );
-        return Ok(());
     }
 
     // Layer-3 gate inputs: the agent definition's ambient opt-in flag and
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2465, opened 2026-08-08
docs: status doc for a live agentmux-srv Section-handle leak

AgentX-asaf:
## Summary

Live investigation this session (triggered by a "pagefile growing unbounded" question) found `agentmux-srv` itself — not Audiosrv, not the already-resolved GPU-driver-commit mechanism — holding 214,636 anonymous Windows Section handles at check time. Same leak fingerprint as the earlier resolved Audiosrv bug (`STATUS_PF_COMMIT_GROWTH_INVESTIGATION_2026_07_24.md`), but this time genuinely inside AgentMux's own process. System commit was at 95% of the commit limit at check time.

Root cause narrowed to `sysinfo`'s 30s whole-machine process-refresh loop (`agentmux-srv/src/backend/sysinfo.rs:148`) as the leading hypothesis, not yet proven at the exact API-call level — `memmap2` and `portable-pty` were ruled out via direct source inspection.

Live mitigation (restart srv by PID) identified but deliberately not executed without explicit confirmation — traced the actual supervision path first and found it's a much bigger blast radius than the Audiosrv fix: killing srv recycles the entire CEF host and hard-kills every live CLI subprocess across every pane the shared-channel srv instance backs, not just one.

Docs-only change, no code/behavior modified.

### DIFF
--- a/docs/status/STATUS_SRV_SECTION_HANDLE_LEAK_2026_08_08.md
+++ b/docs/status/STATUS_SRV_SECTION_HANDLE_LEAK_2026_08_08.md
@@ -1,0 +1,168 @@
+# Status: `agentmux-srv` Leaking Windows Section Handles — Live Finding (2026-08-08)
+
+**Related, but a different mechanism:** `docs/status/STATUS_PF_COMMIT_GROWTH_INVESTIGATION_2026_07_24.md`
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2467, opened 2026-08-08
fix(agent-pane): tool preview scrollbar sits flush at the panel edge

Agent2-asaf:
## Summary
- First of a tool-preview refinement series: removes the dead strip between every tool preview's scrollbar and the preview surface's edge.
- Root cause (measured in `docs/specs/SPEC_TOOL_PREVIEW_SCROLLBAR_EDGE_PADDING_2026_08_08.md`, included): all tool previews (Read, Bash, Write, Edit, Grep/Glob, Agent, Task, default) share one scroll container, `.agent-tool-overlay-log`, inside `.agent-tool-panel` whose `padding: 6px 8px` inset the log's entire box — scrollbar included — 8px inboard of the visible surface edge. A `::-webkit-scrollbar` renders inside its owning element's box, so the panel padding is what floated the track away from the edge.
- Fix: panel padding → 0; the log carries the inset itself (left/top/bottom unchanged breathing room, right tightened to 4px since it now only separates content from the scrollbar). Two SCSS declarations changed total.
- Intended side effects: the overlay header's divider and per-tool section bars (bash cmd/exit rows, diff header) now run full-bleed across the panel, and preview content gains ~8px usable width per side (content inset was a double-stacked 16px, now 8px, matching the header).
- Not affected (verified via cascade order, documented in the spec): ActivityRow's live activity feed reuses the log class but overrides padding with a later-imported rule.

## Test plan
- [x] `npx stylelint` on both edited files — identical error output before/after (all pre-existing grandfathered issues; zero new)
- [x] `npx vitest run` — 2385 passed, 0 failed
- [ ] Visual pass per the spec's §6 matrix (Read code/markdown, Bash long output + horizontal scroll, Write, Edit plain/highlighted diff, Grep/Glob, expand/collapse animation, preview zoom, persistent shell block, one failed/denied state) — reviewer or post-merge dev-build check

### DIFF
--- a/.changesets/1786193214-fix-agent-pane-tool-preview-scrollbar-sits-flush-at-the-pane-b8yi.md
+++ b/.changesets/1786193214-fix-agent-pane-tool-preview-scrollbar-sits-flush-at-the-pane-b8yi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): tool preview scrollbar sits flush at the panel edge
--- a/docs/specs/SPEC_TOOL_PREVIEW_DEDENT_2026_08_08.md
+++ b/docs/specs/SPEC_TOOL_PREVIEW_DEDENT_2026_08_08.md
@@ -1,0 +1,192 @@
+# SPEC: Tool preview common-indentation stripping (dedent)
+
+**Date:** 2026-08-08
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2468, opened 2026-08-08
fix(build): fail production builds when VITE_MUXBUS_CLIENT_ID resolves empty

AgentY-asaf:
## Summary

The 2026-08-05 Desktop portable shipped with the **entire MuxBus Cloud UI invisibly absent** — the HostPopover section, the per-agent Identity panel section, and the Armory Accounts tile all gone. Root cause (forensically confirmed against the shipped bundle): that one build run didn't load `frontend/.env.production`, so `VITE_MUXBUS_CLIENT_ID` compiled to `""`, `isConfigured()` folded to `false`, and the `<Show>` gates hid everything. The minifier even dead-code-eliminated the whole login path (the Cognito domain string is absent from the shipped bundle too), which is what made a bad build byte-for-byte indistinguishable from a good one until a user noticed missing UI.

The env file was present and correct at the build commit, the tree was clean (no `.dirty` label), and rebuilds from the same checkout bake the ID correctly today — the env loss was transient and unreproducible. Which is exactly why this needs a guard rather than a one-off fix.

## Change

New `requireMuxBusClientId()` Vite plugin: **production builds now fail loudly at config time** if the resolved env lacks the client ID, with a message pointing at the env pipeline. Since `frontend/.env.production` is committed, a correctly-functioning build can never trip this — if it fires, the build environment is broken and the artifact must not ship. `--mode dev` builds and the dev server are never blocked.

## Test plan

- [x] Production build with `VITE_MUXBUS_CLIENT_ID=` shadowed empty (the exact suspected Aug 5 failure mode) → build **fails** with the guard's message, exit 1
- [x] Normal production build → passes; verified `26odoigu…` client ID present in the output bundle
- [x] `--mode dev` build with the same empty shadow → guard does not fire, build proceeds

### DIFF
--- a/.changesets/1786194007-fix-build-fail-production-builds-when-vite-muxbus-client-id--v10c.md
+++ b/.changesets/1786194007-fix-build-fail-production-builds-when-vite-muxbus-client-id--v10c.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): fail production builds when VITE_MUXBUS_CLIENT_ID resolves empty
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -65,6 +65,42 @@ function platformResolve(): Plugin {
 }
 
 /**
+ * Fails a production build outright when VITE_MUXBUS_CLIENT_ID resolves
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2469, opened 2026-08-08
fix(agent-stream): don't end the turn on mid-turn narration text (stop_reason gate)

AgentX-asaf:
## Summary

Recurrence of the premature-"Worked" symptom that `SPEC_PERSISTENT_TURN_END_TEXT_GATE_2026_07_30.md` / #2369 fixed: status flips from "Working…" to "Worked" while the agent is clearly mid-work (tool calls still coming). User-reported with the exact invariant this fix encodes: *"the agent will always output a normal text message at the end — 'Worked' can never happen except after a text output."*

**Root cause — the #2369 text gate is intact but the CLI's frame shape defeats it.** Verified against a real captured session (1,790 assistant frames): the CLI emits **one assistant frame per content block** — zero combined `[text, tool_use]` frames exist in the whole session. So narration text preceding a tool call ("Now let me check the config…") arrives as its own text-only frame, passes the `hasText && !hasToolUse` gate, and fires `session_end` mid-turn. **377 of 409 text-only frames in that session were this mid-turn kind** — every one a premature "Worked" flip.

**The fix, from the same data:** every frame is stamped with its API message's `stop_reason`. All 377 mid-turn text frames carry `"tool_use"`; genuine finals carry `"end_turn"` (32) or `"stop_sequence"` (1). New gate:

```ts
if (!hasToolUse && hasText && message.stop_reason !== "tool_use")
```

Only an **affirmative** `"tool_use"` suppresses — `null`/`undefined` (an older CLI that doesn't stamp it) and every terminal reason still end the turn, so this cannot reintroduce #1757's stuck-forever failure; the 180s liveness watchdog remains the backstop either way.

## Test plan

- [x] 4 new tests: mid-turn narration (`stop_reason: "tool_use"`) suppressed; `end_turn`, `stop_sequence`, and missing-`stop_reason` all still end the turn
- [x] Mutation-tested: reverting the gate to the old condition makes the narration test fail (confirmed, then restored)
- [x] `npx vitest run frontend/app/view/agent/providers/claude-translator.test.ts` — 31/31
- [x] `npx vitest run frontend/app/view/agent/` — 1024/1024

### DIFF
--- a/.changesets/1786194396-fix-agent-stream-don-t-end-the-turn-on-mid-turn-narration-te-coie.md
+++ b/.changesets/1786194396-fix-agent-stream-don-t-end-the-turn-on-mid-turn-narration-te-coie.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-stream): don't end the turn on mid-turn narration text (stop_reason gate)
--- a/frontend/app/view/agent/providers/claude-translator.test.ts
+++ b/frontend/app/view/agent/providers/claude-translator.test.ts
@@ -51,6 +51,61 @@ describe("ClaudeTranslator", () => {
             expect(events).toHaveLength(0);
         });
 
+        it("does NOT end the turn on a text-only frame stamped stop_reason 'tool_use' (mid-turn narration)", () => {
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2471, opened 2026-08-08
feat(statusbar): Disk readout explains its % in the tooltip; click opens a per-drive popover

AgentY-asaf:
## Summary

Per direct user feedback on the status bar's Disk pill:

1. **Hover no longer talks about the page file.** The tooltip now explains what the number actually is — free ÷ capacity on the named drive, with live figures: *"Free space on C: — 120.5G free of 460.2G (% = free ÷ capacity). Click for all drives."* The page-file significance stays on the PF gauge's own tooltip where it belongs. Pill metric and color thresholds unchanged.
2. **Clicking the pill opens a per-drive breakdown** — every mounted volume with free/total and a used-share bar — mirroring the CPU readout's per-core panel (same positioning/airspace/outside-click/Esc pattern, `DiskVolumesPopover` as the sibling of `CpuCoresPopover`).

Backend: `get_disk_data` additionally emits per-volume capacity keyed by mount point (`disk:vol:<mount>:free_gb` / `:total_gb`), deduped by mount point and skipping zero-capacity media (empty card readers). `get_pagefile_volume_data` marks its watch volume (`disk:vol:<mount>:watch`) so the tooltip can name the drive — the values map is f64-only, so the letter rides in the key.

Parsing/formatting/color helpers are pure (`disk-volumes.ts`), covering the tricky Windows case where the mount's own colons/backslashes sit inside the key (`disk:vol:C:\:free_gb`) and dropping torn half-pairs rather than rendering a fabricated 0.

## Test plan

- [x] `vitest` — 9/9 new tests pass (`disk-volumes.test.ts`)
- [x] `tsc --noEmit` — clean
- [x] Full production vite build — clean (new SCSS compiles)

### DIFF
--- a/.changesets/1786195261-feat-statusbar-disk-readout-explains-its-in-the-tooltip-clic-toq5.md
+++ b/.changesets/1786195261-feat-statusbar-disk-readout-explains-its-in-the-tooltip-clic-toq5.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(statusbar): Disk readout explains its % in the tooltip; click opens a per-drive free-space popover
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #2472, opened 2026-08-08
fix(agent-stream): flush the turn's tail before TurnEnd so panes don't stick on Working

AgentX-asaf:
## Summary

User-reported (and live-log-confirmed on this machine): a pane shows "Working…" after its turn has genuinely ended, with the agent fully idle. The exact trace from the reporting pane:

```
13:19:17.210  Streaming → Done   cmd=TurnEnd
13:19:17.216  Done → Streaming   cmd=StreamFlushObserved  toolsActive=0
(then nothing — phantom Working until the 180s watchdog, which ReconcileTurnActive can immediately re-promote, so it oscillates)
```

**Mechanism:** `session_end` dispatches `TurnEnd` synchronously (`useTurnLifecycle.ts:113`), but the turn's final text nodes are still sitting in the RAF flush queue (`stream-flush-queue.ts`). They flush a frame later, landing in `Done.completed` — where the reducer's `StreamFlushObserved` re-promotion (deliberately kept by #2420 for genuine multi-round continuations, `reducer.ts:206-260`) reads the turn's own tail as "the agent picked up another round" and re-enters `Streaming`, with no further `session_end` ever coming.

**Fix:** ordering, not heuristics — `queue.flushNow()` before `finalizeTurn()`, so the tail lands while still `Streaming` and only then does `TurnEnd` settle the phase. The re-promotion mechanism itself is untouched; a genuine next round's flushes arrive after this point and still re-promote correctly. `flushNow` cancels the armed RAF, and the RAF path already no-ops on empty queues, so no duplicate or empty `StreamFlushObserved` can land after `TurnEnd`.

Distinct from today's #2469 (premature "Worked" via the stop_reason gate) — that fixed the turn ending too *early*; this fixes it never visually ending at all. Both bugs were live in the reported build.

## Test plan

- [x] New `stream-flush-queue.test.ts` (module had zero coverage): synchronous flush, no RAF double-dispatch after `flushNow`, empty-queue no-op (no empty `StreamFlushObserved`), and a genuine next-round flush still flowing normally
- [x] `npx vitest run frontend/app/view/agent/` — 79/79 files passing

### DIFF
--- a/.changesets/1786195522-fix-agent-stream-flush-the-turn-s-tail-before-turnend-so-pan-0j4n.md
+++ b/.changesets/1786195522-fix-agent-stream-flush-the-turn-s-tail-before-turnend-so-pan-0j4n.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-stream): flush the turn's tail before TurnEnd so panes don't stick on Working
--- a/frontend/app/view/agent/stream-flush-queue.test.ts
+++ b/frontend/app/view/agent/stream-flush-queue.test.ts
@@ -1,0 +1,117 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Coverage for `flushNow()` — the synchronous flush added for the
+ * session_end path (turn-tail flush race: the turn's own trailing document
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2474, opened 2026-08-08
fix(armory): accounts appear live everywhere without a reload

AgentA-asaf:
## Summary

New/changed accounts didn't appear in the Armory (or the per-agent identity panel, or account pickers) until a reload. Found while live-testing the #2463 auth fixes on a fresh isolated-auth dev build.

## Root cause

The shared account cache (`identity-model.ts`) is the single source every account view reads from — but it was only refreshed by its own CRUD methods' explicit `refreshAccountCache()` calls. It never subscribed to the backend's `identityaccounts:changed` broadcast, so any backend-originated change (completing an in-app OAuth login via `identity_auth_persist.rs`, API-key verify, upsert/delete from another tab, the spawn-time expiry probe) left every consumer stale. `AgentLaunchModal` had already hand-rolled a private subscription to work around exactly this — its own comment documents the gap: *"the shared account cache (identity-model.ts) isn't itself subscribed to the backend's identityaccounts:changed broadcast — callers refresh it explicitly."*

## Fix

`primeAccountCache()` (already called once at app startup, after event subscriptions initialize) now also installs a single app-lifetime `identityaccounts:changed` subscription that refreshes the cache. Every consumer wired via `subscribeAccountChanges()`/`accountsAtom` — Armory Accounts tab, per-agent identity links panel, pickers — goes live automatically, with no per-view subscriptions needed. Guarded against duplicate installation on repeated calls.

Backend event coverage verified before relying on it: upsert, key-verify, OAuth login persist, delete, and the expiry probe all already publish the event.

## Test plan

- [x] New `identity-model.livesync.test.ts`: exactly one subscription installed even across repeated `primeAccountCache()` calls; the broadcast handler refreshes the cache and notifies cache listeners
- [x] Full vitest suite: 2390+ passing; two unrelated tests flaked in different full runs (each passes in isolation, machine was concurrently running a dev instance) — CI on a quiet runner is the arbiter
- [ ] Live check on the running dev build: complete an in-app OAuth login → account appears in Armory immediately

### DIFF
--- a/.changesets/1786197128-fix-armory-accounts-appear-live-everywhere-subscribe-the-sha-c7w1.md
+++ b/.changesets/1786197128-fix-armory-accounts-appear-live-everywhere-subscribe-the-sha-c7w1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(armory): accounts appear live everywhere - subscribe the shared account cache to identityaccounts:changed
--- a/frontend/app/view/agent/components/AgentLaunchModal.tsx
+++ b/frontend/app/view/agent/components/AgentLaunchModal.tsx
@@ -20,7 +20,6 @@ import { RpcApi } from "@/app/store/rpc-api";
 import { TabRpcClient } from "@/app/store/rpc-util";
 import { isAvailable, watchCapability } from "@/app/store/toolchain-capabilities";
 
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2475, opened 2026-08-08
fix(app-api): window rename validates existence; naming routes return real status codes

Agent2-asaf:
## Summary
- Live-probing the window-title App API (via a dev instance's `authkey.dev` bridge) found that `POST /api/v1/window/name` with a **well-formed UUID matching no window** returns `200 {"success": true}` while changing nothing — a false positive for any agent holding a stale `window_id`.
- Root cause: the reducer's `UpdateWindowMeta` arm was the **only** meta arm with no existence guard (workspace/tab/block all validate and error), and the persist subscriber silently no-ops on a wstore miss — so nothing anywhere caught the phantom. The code comment at the dispatch site even documents the intended contract ("validates entity exists") that this arm violated.
- Fix 1: add the same guard its three sibling arms have. Verified safe: `bootstrap_state_from_wstore` hydrates every pre-existing window (including the wcore-seeded first-launch window, which is created before hydration runs) and all runtime creation goes through `CreateWindow` — the old "wcore-direct paths won't appear in reducer state" comment predates that hydration and is updated accordingly.
- Fix 2: the naming-route family (`window/name`, `tab/name`, `pane/title`, `workspace/name`) mapped every service error to 500. `finish_name_call` now maps "not found" → 404 and "invalid" (e.g. malformed object id) → 400, following the existing `app_api_error_status` text-matching precedent; `handle_window_name`'s inlined duplicate of `finish_name_call` is converged onto it.
- Spec with the probe transcript baseline + deferred follow-ups (clear-to-default parity, dev-instance discovery ergonomics, grapheme-safe clamping): `docs/specs/SPEC_WINDOW_NAME_API_HARDENING_2026_08_08.md`.

## Test plan
- [x] New reducer tests: phantom window id → `Event::Error`; known window → `WindowMetaUpdated` (happy path unregressed)
- [x] New HTTP tests: phantom UUID → 404; malformed id → 400; seeded window (srv_state hydrated via the same `bootstrap_state_from_wstore` production runs) → 200 + `window:displayname` persisted in wstore
- [x] `cargo test --workspace -- --test-threads=1` — all green (2067 passed, 0 failed; run with isolated AGENTMUX_* env)

### DIFF
--- a/.changesets/1786197197-fix-app-api-window-rename-validates-existence-naming-routes--olxd.md
+++ b/.changesets/1786197197-fix-app-api-window-rename-validates-existence-naming-routes--olxd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(app-api): window rename validates existence; naming routes return real status codes
--- a/agentmux-srv/src/reducer/window.rs
+++ b/agentmux-srv/src/reducer/window.rs
@@ -109,19 +109,34 @@ pub(super) fn handle_switch_workspace(
 /// through the reducer's broadcast bus" so the WaveObjUpdate bridge
 /// can fan out to the frontend.
 ///
-/// The validation is best-effort: reducer state's `windows` map only
-/// holds windows for which a workspace mapping has been registered
-/// (via `handle_create_window`). Windows created via wcore-direct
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2479, opened 2026-08-09
fix(statusbar): shorten the Disk pill's hover tooltip to match the other stat pills

AgentY-asaf:
## Summary

Follow-up to #2471 per direct feedback: the Disk tooltip was a full sentence with live figures and a call to action ("Free space on C: — 120.5G free of 460.2G (% = free ÷ capacity). Click for all drives."), unlike every sibling stat tooltip's short descriptive form ("Per-core CPU usage", "Memory used and total", "Commit charge used and total (RAM + page file budget)."). Shortened to match: **"Free share of system drive (C:)"** — names what the % is a share of, nothing else. Live figures and the drive list already live in the click-open panel, not the hover.

Also live-verified the click-open panel's positioning in a running dev instance via CDP (attached both a CPU-popover and Disk-popover screenshot for comparison) — it's correctly anchored above its button with the same `top-end` placement logic as the CPU per-core panel, and the airspace cut over native panes is working. Could not reproduce a placement issue in that session.

## Test plan

- [x] `vitest` — 9/9 tests pass (updated to assert the exact short string)
- [x] Live CDP verification against a running `task dev` instance: clicked the Disk pill, confirmed real data renders (`C: 472.2G free of 3.6T`, `G: 205.5G free of 402.0G`) and the popover paints on top of native panes correctly

### DIFF
--- a/.changesets/1786279599-fix-statusbar-shorten-the-disk-pill-s-hover-tooltip-to-match-wvw3.md
+++ b/.changesets/1786279599-fix-statusbar-shorten-the-disk-pill-s-hover-tooltip-to-match-wvw3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(statusbar): shorten the Disk pill's hover tooltip to match the other stat pills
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2482, opened 2026-08-09
fix(agent-pane): surface the spawn gate's refusal detail instead of bare 'Agent encountered an error'

AgentA-asaf:
## Summary

Live repro on a fresh v0.54.14 install (claudius, pre-existing Agent1): the identity spawn gate correctly refused the spawn — logs show `identity.spawn.blocked: no credentials for provider claude … spawn refused` — and wrote its `error_during_execution` frame with the full actionable text in `error.message` ("…Bind an account for this provider in the Armory."). But the pane rendered only the bare **"Agent encountered an error"**.

## Root cause

`ClaudeTranslator`'s result-frame error path only read `rawEvent.result` for the message. AgentMux's own synthesized frames (the spawn-gate refusals from `agent_io.rs`/`input.rs`) carry their detail in `error.message` — a field the translator never read — so every gate refusal fell through to the hardcoded generic fallback. The detail was produced, persisted, and then dropped at the last hop.

The backend classifier (`agents/failure.rs::frame_error_text`) already reads `/error/message` first; the translator now mirrors its exact field priority: `result` → `error.message` → `error` (string) → generic fallback.

Codex and Gemini translators already read `error?.message` — Claude was the only gap (Kimi is api-key-class; the OAuth gate never refuses it).

**This also closes the "loose end" from #2463/#2464**: the reason Auth-classified detail never reached the pane on first-turn gate refusals was never the classifier — it was this dropped field.

## Test plan

- [x] 3 new translator tests: the exact gate frame shape surfaces its `error.message`; string-typed `error` also surfaced; `result` still wins when both present
- [x] Full claude-translator suite: 34/34 pass

### DIFF
--- a/.changesets/1786282080-fix-agent-pane-surface-the-spawn-gate-s-actual-refusal-text--mqeh.md
+++ b/.changesets/1786282080-fix-agent-pane-surface-the-spawn-gate-s-actual-refusal-text--mqeh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): surface the spawn gate's actual refusal text instead of the bare 'Agent encountered an error'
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2483, opened 2026-08-09
feat(diagnostics): mem_attribution logs unattributed commit gap + handle-count anomalies

AgentX-asaf:
## Summary

Automates the exact manual methodology that isolated both prior pagefile-exhaustion incidents — the Windows 10 **Audiosrv** Section-handle leak (`STATUS_PF_COMMIT_GROWTH_INVESTIGATION_2026_07_24.md` §16) and **agentmux-srv's own** Section-handle leak (`STATUS_SRV_SECTION_HANDLE_LEAK_2026_08_08.md`). Both took days of manual Sysinternals work to find because `mem_attribution` only bucketed process-*private* commit, which is structurally blind to the leak class both incidents shared: pagefile-backed Section objects held via leaked handles are charged to system commit but to **no process's private bytes**.

Two additions to the existing 30s `mem_attribution` cycle:

1. **`unattributed_gb`** = `commit_used − (agentmux + panes + other)`. Flat process buckets + climbing commit = this gap climbing = the leak-class signature — now visible in one `muxlog srv grep mem_attribution` instead of requiring an external PowerShell sampler running for hours.
2. **Per-process handle-count sweep** (one `OpenProcess`/`GetProcessHandleCount`/`CloseHandle` triplet per live PID — microseconds each, no handles retained) with two-tier anomaly WARNs:
   - any `agentmux*` process ≥ **20K** handles (healthy baseline is under 1K; the 08-08 leak hit 215K) → "possible handle leak in an AgentMux process," pointing at the `handle64 -s -p <pid>` type-breakdown step and the runbook doc;
   - any external process ≥ **50K** (Audiosrv hit 85K; the highest legitimately-handle-heavy process observed on the 08-08 machine snapshot was ~25K) → "Audiosrv-class OS leak?" with the restart-that-service remediation pointer.
   - Top-3 handle counts also ride the info line for trend history even below thresholds.

Either incident would have been flagged automatically within 30s of crossing threshold, instead of being found weeks later at 85K/215K handles with commit at 95% of the limit.

`classify_handle_anomalies` is factored out pure and unit-tested (same pattern as the existing `classify_commit_attribution`), plus a live smoke test of the winapi wrapper against the test process itself. Windows-only, matching the rest of this file's attribution machinery.

## Test plan

- [x] 5 new tests (healthy baselines trip nothing; agentmux-tier threshold; external-tier threshold; worst-first ordering; live `GetProcessHandleCount` smoke test), plus the existing `log_memory_attribution` real-process-table test now exercises the full new path
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2072/2072
- [x] No new windows-sys features needed (`Win32_System_Threading`/`Win32_Foundation` already enabled)

### DIFF
--- a/.changesets/1786297813-feat-diagnostics-mem-attribution-logs-unattributed-commit-ga-45dr.md
+++ b/.changesets/1786297813-feat-diagnostics-mem-attribution-logs-unattributed-commit-ga-45dr.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(diagnostics): mem_attribution logs unattributed commit gap + handle-count anomalies
--- a/agentmux-srv/src/backend/sysinfo.rs
+++ b/agentmux-srv/src/backend/sysinfo.rs
@@ -125,6 +125,89 @@ fn classify_commit_attribution(
     }
 }
 
+/// Handle-count anomaly thresholds. Both real incidents this detects were
+/// found manually with Sysinternals Handle after weeks of commit growth
+/// (docs/status/STATUS_PF_COMMIT_GROWTH_INVESTIGATION_2026_07_24.md §16:
+/// Audiosrv at ~85K handles ≈ 18GB of leaked pagefile-backed Sections;
+/// docs/status/STATUS_SRV_SECTION_HANDLE_LEAK_2026_08_08.md: agentmux-srv
+/// itself at ~215K, 99.5% Sections). A leaking process crosses these within
+/// hours of a healthy baseline (both incidents' healthy baselines were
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentX-asaf:
P1 resolved: the referenced runbook doc `docs/status/STATUS_SRV_SECTION_HANDLE_LEAK_2026_08_08.md` now exists on `main` — it landed via PR #2465 (which was open-but-unmerged when this PR was first reviewed; merged just now, including corrections for two Codex P1s on the doc's own content). This branch has merged `main`, so the path resolves in-tree.

PR #2484, opened 2026-08-09
docs: specs for account⇄agent binding — Armory 'Bind to Agent' menu (primary) + account adoption (companion)

AgentA-asaf:
Design doc for letting an unlinked agent adopt an existing Armory account instead of forcing a fresh OAuth — motivated by the claudius v0.54.14 live test, where a legacy agent was correctly gate-refused while a valid claude account sat unused on the same machine.

Key points:
- Two of three piggyback surfaces already exist (launch-modal picker, seed-from-global-login); the gap is adopting an existing **Armory** account from the failure-recovery row and the Identity tab.
- Pure frontend feature: `LinkAgentIdentityCommand` upsert + the same `cmd:env` live-refresh step `useGlobalLogin` already does. No backend changes.
- Silent auto-adoption explicitly rejected as default (attributability / "single point, not global" §7); cross-machine sync explicitly out of scope.

Docs only — no code changes.


a5af:
@codex review

AgentA-asaf:
Folded into the implementation PR — specs + code land together there.

PR #2485, opened 2026-08-09
feat(armory): 'Bind to Agent' context menu on account rows

AgentA-asaf:
## Summary

Implements `SPEC_ARMORY_BIND_TO_AGENT_CONTEXT_MENU_2026_08_09.md` (included in this PR, along with its companion account-adoption spec — supersedes #2484, which carried the specs alone).

Right-click an Armory account row → **Bind to Agent ▸** submenu listing the channel's user-owned agents, plus **Copy account ID**. The submenu doubles as a live binding overview — the motivating case is a user with three claude accounts who has no way to see or change which agent uses which:

- ✓ checkmark on agents already bound to this account
- sublabel naming the *different* account an agent is currently bound to ("bound: work-claude") or "no account bound" — rebinds are disclosed before the click, so no confirm dialog on a non-destructive upsert
- running agents (open panes) sort first with a marker; non-running agents still bindable (pre-provisioning)
- CLI-OAuth accounts only offer same-provider agents (alias-canonicalized both sides — the #2377 lesson); service accounts offer all agents; empty set renders disabled-with-reason instead of hiding

**Bind action:** the existing `LinkAgentIdentityCommand` upsert; for a running agent, also mirrors `useGlobalLogin`'s `cmd:env` config-dir refresh so the new binding takes effect next turn without restart. Account list re-renders live via #2474's cache sync. Zero backend changes.

## Test plan

- [x] Menu builder is a pure function — 11 unit tests (provider filtering + aliases, boundHere/boundElsewhere annotations incl. alias-stored links, running-first sort, template exclusion, sublabel composition)
- [x] Full vitest suite: 2423 passed, 0 failed
- [ ] Live check (per spec §5): 3 claude accounts → right-click each → verify binding overview, rebind an agent between accounts, confirm next turn's spawn log shows the new account id

### DIFF
--- a/.changesets/1786300545-feat-armory-right-click-an-account-row-to-bind-it-to-an-agen-1z96.md
+++ b/.changesets/1786300545-feat-armory-right-click-an-account-row-to-bind-it-to-an-agen-1z96.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): right-click an account row to bind it to an agent - Bind to Agent submenu with live binding overview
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

PR #2487, opened 2026-08-09
fix(dev): Vite watcher ignores never matched (34x handle reduction); cool down handle-anomaly warns

AgentX-asaf:
## Summary

Two follow-ups from the handle-leak diagnostics' first live run (#2483), which immediately flagged two `task dev` Vite processes at **75K and 129K Windows handles**:

### 1. `vite.config.ts`'s `server.watch.ignored` patterns were ALL no-ops

Chokidar matches ignore globs against **absolute paths**, so the relative `"dist/**"`/`"target/**"` forms matched nothing, silently. Worse, even the canonical `"**/dist/**"` form fails on any checkout under a dot-directory (every agent checkout lives under `~/.agentmux/...`): picomatch's `**` refuses to traverse dot-segments like `.agentmux` without `dot:true`, which chokidar doesn't set. Both failure modes verified empirically against picomatch directly.

Net effect: the watcher was watching `dist/cef-dev`'s entire extracted CEF runtime (and `target/`) — which also means **#2424's watcher-vs-CEF-extraction race fix was silently ineffective on these checkouts**. Fix: anchor all four patterns to the absolute, forward-slashed repo root — the dot-directories become literal pattern text, which matches fine.

**Live A/B on this machine** (`task dev`, same checkout, watcher settled):
| | handles |
|---|---|
| before (relative globs) | 75,005 (~1h uptime, climbing) |
| after (absolute globs) | **2,196** (stable across repeated samples) |

This is also a real lead on the user-observed "commit drops when I exit AgentMux" pagefile pattern — Vite dev-server processes are part of `task dev`'s tree, and each was holding tens of thousands of kernel handles it shouldn't have.

### 2. Handle-anomaly WARNs cool down per-PID (10 min)

The first live run re-warned every 30s attribution cycle (plus the urgent-on-pressure path) for a persisting anomaly. Now one WARN per PID per 10 minutes; the info line's `handles_top` still carries counts every cycle so trend data is not lost. A PID that recovers below threshold is pruned from the cooldown map, so a later re-crossing warns immediately. Pure `filter_anomalies_for_warn` factored out + 3 unit tests.

**Live-verified in the same dev run**: 7 attribution cycles, the still-anomalous external Vite (129K handles) warned exactly once.

## Test plan

- [x] 3 new cooldown unit tests (immediate first warn / per-PID independence / prune-on-recovery)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2075/2075
- [x] Live A/B of the Vite fix and the cooldown, numbers above
- [x] No frontend JSON imports exist, so the (now actually working) `*.json` ignore can't break HMR for imported modules

### DIFF
--- a/.changesets/1786303976-fix-dev-vite-watcher-ignores-never-matched-34x-handle-reduct-nwt0.md
+++ b/.changesets/1786303976-fix-dev-vite-watcher-ignores-never-matched-34x-handle-reduct-nwt0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dev): Vite watcher ignores never matched (34x handle reduction); cool down handle-anomaly warns
--- a/agentmux-srv/src/backend/sysinfo.rs
+++ b/agentmux-srv/src/backend/sysinfo.rs
@@ -142,6 +142,17 @@ const HANDLE_WARN_AGENTMUX: u32 = 20_000;
 const HANDLE_WARN_EXTERNAL: u32 = 50_000;
 #[cfg(target_os = "windows")]
 const HANDLE_TOP_N: usize = 3;
+/// Minimum gap between repeat WARNs for the SAME still-anomalous PID. The
+/// first live run of this detector (2026-08-09) re-warned every 30s
+/// attribution cycle — and more under the urgent-on-pressure path — for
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2488, opened 2026-08-09
fix(agent-pane): new-tab '+' keeps the existing conversation as a visible tab

GenericAgentX-asaf:
## Summary

Clicking the tab-strip `+` on an agent pane pushed a blank (agentId-less) block onto the pane's stack — correct — but the entire pane then appeared to vanish: the tab strip only rendered inside `AgentPresentationView`, and a blank active block falls through `AgentViewWrapper`'s gate to `AgentPicker`, which had no strip at all. The previous conversation (e.g. "Camper") was still in `blockStack` but had no visible pill to switch back to.

## Fix

- Move `PaneTabStrip` up to `AgentViewWrapper` so it renders for BOTH branches (picker and presentation).
- Drive the tab list primarily from the pane's own `blockStack` (mirroring `term.tsx`'s `termTabs`), merging cross-pane fork tabs in as deduped extras so fork-switching keeps working.
- A blank tab shows as "New Agent" until an agent is launched into it, then picks up the agent's name; double-click-rename only offered once a definition exists.

Behavior after: 1 tab → just the `+`; click `+` → two pills ("Camper" + "New Agent"), switchable, existing conversation intact.

## Verification

- `tsc --noEmit` clean (verified the checker catches injected errors first).
- `vitest` agent+layout suites: 1052/1053 pass; the 1 failure (`tool-renderers/registry`) is a pre-existing timeout flake, passes in isolation.
- Not yet clicked through live — the dev instance died at session teardown (see `docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_09.md`, lands in a follow-up PR). Reviewer or a follow-up session should verify the `+` flow in a running build.

### DIFF
--- a/.changesets/1786314154-fix-agent-pane-new-tab-keeps-the-existing-conversation-as-a--r7z4.md
+++ b/.changesets/1786314154-fix-agent-pane-new-tab-keeps-the-existing-conversation-as-a--r7z4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): new-tab '+' keeps the existing conversation as a visible tab instead of replacing the pane
--- a/frontend/app/view/agent/agent-view.scss
+++ b/frontend/app/view/agent/agent-view.scss
@@ -33,6 +33,27 @@
 @use "styles/shell-node";
 @use "styles/loading-overlay";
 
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2493, opened 2026-08-10
fix(agent-pane): expanded dock row shows the full command word-wrapped at the top

GenericAgentX-asaf:
## Summary

Expanding an ActivityDock row (long-running shell / promoted Bash call) jumped straight into the log — with the collapsed title ellipsis-truncated, a long command was unidentifiable. This adds a full-command header at the top of the expanded view: monospace, `pre-wrap` + `overflow-wrap: anywhere` so it wraps fully, user-selectable, separated from the log by a divider.

- Shell rows use `shell.cmd` (the real command, even when the row has a custom display title).
- Promoted tool rows use `BashParams.command`, falling back to the title.

Found live while watching a long-running process in a dev instance (user report 2026-08-10).

## Verification

- `tsc --noEmit` clean.
- Verified live via Vite hot reload in the running dev instance.

### DIFF
--- a/.changesets/1786324878-fix-agent-pane-expanded-dock-row-shows-the-full-command-word-pzr5.md
+++ b/.changesets/1786324878-fix-agent-pane-expanded-dock-row-shows-the-full-command-word-pzr5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2489, opened 2026-08-09
feat(agent-pane): wire the attached-task status axis — 'Running in background' instead of stuck 'Working…'

GenericAgentX-asaf:
## Summary

**Stacked on #2488** (both touch `agent-view.tsx`); retarget to `main` after it merges.

The `attachedTask` reducer slice from `SPEC_ATTACHED_TASK_STATUS_AXIS_2026_08_02.md` landed a week ago with **zero producers and zero consumers** — the spec's §6 deferred the wiring pending live verification and no session ever picked it up. This PR lands all three deferred pieces:

1. **Dispatch call site** — `hasLiveAttachedActivity()` (new pure helper mirroring the ActivityDock's shell + subagent + tool aggregate, the spec §4's chosen data source) drives an edge-triggered effect in `agent-view.tsx` dispatching `AttachedTaskObserved` / `AttachedTaskCleared` on the 0→1 / 1→0 transitions, with the same one-shot wall-clock re-check timer discipline as the existing tool-promotion effect.
2. **Projection plumbing** — `attachedTask` setter in `AgentPaneProjections`, projected in `dispatch()`, backed by a new `attachedTaskAtom`, registered in the pane's projections.
3. **Rendering** — `AgentWorkingRow` gains a calm third state: when the turn is idle but ≥1 attached task is live, it shows "● Running in background · 14m" (slow-pulse dot, secondary text — deliberately reads as *healthy*, not *busy*) instead of "✓ Worked"/nothing. Reverts to the worked summary when the task ends.

This is rung 1 of the ladder in `docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_09.md` (included in this PR) — the doc consolidates why the "stuck Working…" problem survived five prior passes and what remains (`run_in_background` threading, bashwrap idle-kill exemption, session-teardown survival — to be filed as issues).

## Verification

- `tsc --noEmit` clean.
- 253/253 tests pass across `activity/` + `agent-pane-state` suites, including a new `attached-task.test.ts` (shell running/exited, subagent block-scoping, tool promotion threshold, finished-call retention).
- **Live verification still needed** (spec §6's original condition): a promoted foreground `sleep 60` and a real backgrounded shell in a running build. Note: a harness-`run_in_background` Bash call will NOT light this up yet — its ToolNode finishes instantly and `BashParams` has no `run_in_background` field (rung 2, out of scope here). MCP Shell tasks and ≥30s foreground calls will.

PR #2494, opened 2026-08-10
fix(agent-pane): scrub ghost 'running' tool nodes at turn end

GenericAgentX-asaf:
## Summary

Live report (2026-08-10): a `cd … && git remote -v && git status && git branch --show-current` call — a ~1-second command — showed as **running · 45m** in the ActivityDock. The OS process tree under the agent had zero children: the process was long gone, but its ToolNode was stuck at `running` because the `tool_result` never matched, and `scrubOrphanedInProgress` only runs at session boundaries (SessionEnd / HistoryLoaded). Known gap from `RETRO_TASK_DEV_IDLE_KILL_FALSE_POSITIVE_2026_07_31.md` §"stuck dock entry", follow-up finally implemented.

**The invariant that makes this safe:** a foreground tool call blocks its turn (cannot outlive it), and a backgrounded harness call resolves its ToolNode immediately — so a tool node still `running` shortly *after* TurnEnd is provably an orphan.

- `ScrubOrphanedInProgress` gains `scope: "tools-only"`: running tools scrub; thinking markdown, live MCP shells (turn-independent by design), and `awaiting_answer` questions keep their session-bounded lifecycles.
- `agent-view.tsx` dispatches it 2s after every turn-end edge (`turnJustEndedAtom`), skipped if a new turn already started.

Also protects the attached-task axis (#2489) from being pinned on by the same ghost — it consumes the same running-tool signal.

## Verification

- `tsc --noEmit` clean.
- 87/87 document-reducer tests, incl. 2 new: tools-only scrubs the ghost while leaving shell/thinking/question nodes untouched; no-op when nothing is running.
- Live diagnosis of the triggering incident: dock row's command header (from #2493) identified the call; `Get-CimInstance` process-tree walk confirmed no matching OS process.

### DIFF
--- a/.changesets/1786325635-fix-agent-pane-scrub-ghost-running-tool-nodes-at-turn-end-st-a582.md
+++ b/.changesets/1786325635-fix-agent-pane-scrub-ghost-running-tool-nodes-at-turn-end-st-a582.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): scrub ghost 'running' tool nodes at turn end (stuck dock rows)
--- a/frontend/app/store/agent-document/reducer.test.ts
+++ b/frontend/app/store/agent-document/reducer.test.ts
@@ -865,6 +865,42 @@ describe("agent document reducer", () => {
             expect(r.events).toEqual([]);
         });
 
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

Issue #2492, opened 2026-08-09
Backgrounded agent tasks die on session teardown — a running dev server shouldn't die because the agent session restarted


GenericAgentX-asaf:
Rung 4 of docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_09.md.

Observed 2026-08-09: a heartbeat-wrapped `task dev` launched via a background Bash call survived the bashwrap idle-timeout, the user was actively using the dev instance — then the agent's own session restarted and the background task (and the entire dev process tree under it) was killed with no completion record.

A *declared* long-running background task should be adoptable/reparented (launcher Job Object or srv-owned process group) rather than dying with the agent session that spawned it, while remaining visible and stoppable through the dock. Design needed — interacts with isolation invariants I1–I6 (blast-radius rules in CLAUDE.md / SPEC_MULTI_INSTANCE_ISOLATION_HARDENING_2026_06_03.md).

Also worth folding in: Windows `AgentProcessRegistry.started_at_ms` is hardcoded 0 (`process_tracker/windows.rs:198`, needs NtQueryInformationProcess) — blocks any process-age display/heuristic for these tasks.

Issue #2491, opened 2026-08-09
bashwrap idle-timeout should exempt declared long-running tasks (kills successful 'task dev' runs, forces the heartbeat hack)


GenericAgentX-asaf:
Rung 3 of docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_09.md.

The idle-kill (RETRO_BASHWRAP_PAGER_HANG_LEAK_2026_07_14.md, correct for pager hangs) fires on the structurally different "GUI/daemon succeeded and is now legitimately silent forever" case — reproduced again 2026-08-09: a fully successful `task dev` (build done, Vite up, launcher up) was killed ~10 min after going quiet (RETRO_TASK_DEV_IDLE_KILL_FALSE_POSITIVE_2026_07_31.md is the first documented occurrence).

The documented workaround (120s heartbeat loop) creates the *next* bug: 120s < the pane watchdog's 180s LIVENESS_RECOVERY_MS, so the heartbeat pins the pane on "Working…" for the process's entire lifetime (retro-persistent-agent-working-status-stuck-2026-07-16.md — the 12h Agent1 incident). The two mechanisms are in direct conflict; no amount of frontend polish resolves it.

Fix direction: once a task is *declared* long-running (backgrounded — see the run_in_background threading issue), bashwrap's idle-timeout should not apply to it (or apply a far longer policy). That removes the need for the heartbeat entirely, which un-pins the indicator without touching the watchdog.

Issue #2490, opened 2026-08-09
Thread run_in_background through BashParams + dock adapter — backgrounded Bash tasks are invisible to the ActivityDock


GenericAgentX-asaf:
Rung 2 of docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_09.md (see it for full context).

A harness Bash call with `run_in_background: true` returns its tool result in under a second — the `ToolNode` goes terminal immediately, so the 30s duration-promotion (`tool-adapter.ts`) never fires and the actual long-lived process tree (e.g. `task dev` → cargo → Vite → launcher) is completely invisible to the dock and to the attached-task axis (#2489). Observed live 2026-08-09.

Work:
- Add `run_in_background?: boolean` to `BashParams` (`frontend/app/view/agent/types.ts`) and preserve the flag at stream-parse time (it's already in the transcript's tool_use input JSON; it's just dropped).
- New dock adapter (or tool-adapter branch): a backgrounded call produces a `PinnedActivity` that stays `running` until the harness's task-completion notification lands, not until the tool_result.
- The attached-task axis (`hasLiveAttachedActivity`) picks this up with no changes by construction (spec §4).

This is step 2 of REPORT_LONGRUNNING_TOOLCALL_AUTODETECT_STATUS_2026_07_26.md §3, unbuilt since July.

PR #2495, opened 2026-08-10
fix(agent-pane): Working shows instantly on send; drop redundant Running-in-background footer state

GenericAgentX-asaf:
## Summary

Two working-indicator refinements from live use of the dev build:

**1. "Working…" appears the instant you press Enter.** The `TurnStart` state flip was already synchronous with the send (verified via `[wave-turn]` telemetry — `Idle → Submitting cmd=TurnStart` with zero suppressions), but the row's type-out reveal played on *entry* into the loading state: ~28ms × 8 chars ≈ 250ms of a nearly-empty row, perceived as the indicator lagging the keypress. The reveal now fires only on text *transitions* while already visibly working (tool → tool → phrase); the first text renders in full immediately. Reduced-motion behavior unchanged.

**2. Remove the "Running in background" footer state (#2489's render half).** User feedback: the ActivityDock's running row — same data source, pinned right above the composer — already IS the indicator for a live attached task, with strictly more affordance (full command since #2493, live log, stop button). The footer copy was redundant by construction. The reducer axis (`attachedTask`), its dispatch wiring, and its tests remain — the state's other consumers (liveness-watchdog exemption per the spec's item 3, Swarm surfacing) are tracked follow-ups that need it.

## Verification

- `tsc --noEmit` clean.
- Live via Vite hot reload in the running dev instance.

### DIFF
--- a/.changesets/1786336124-fix-agent-pane-working-shows-instantly-on-send-drop-redundan-51jh.md
+++ b/.changesets/1786336124-fix-agent-pane-working-shows-instantly-on-send-drop-redundan-51jh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): Working shows instantly on send; drop redundant Running-in-background footer state
--- a/frontend/app/view/agent/agent-view.tsx
+++ b/frontend/app/view/agent/agent-view.tsx
@@ -1226,12 +1226,7 @@ const AgentPresentationView = ({
         () => showingLaunchActivity() || workingFromPhase(agentAtoms().turnPhaseAtom[0]()),
     );
     const workingRowVisible = createMemo(
-        () =>
-            workingRowLoading() ||
-            agentAtoms().sessionStatsAtom[0]() != null ||
-            // Attached-task state ("Running in background · Ns") renders in
-            // the same row — see AgentWorkingRow's attachedSince prop.
-            agentAtoms().attachedTaskAtom[0]() != null,
+        () => workingRowLoading() || agentAtoms().sessionStatsAtom[0]() != null,
     );
 
     onMount(() => {
@@ -1806,7 +1801,6 @@ const AgentPresentationView = ({
                             currentTool={agentAtoms().currentToolAtom[0]()}
                             currentToolArg={agentAtoms().currentToolArgAtom[0]()}
                             toolPromoted={hasPromotedTool()}
-                            attachedSince={agentAtoms().attachedTaskAtom[0]()?.since ?? null}
                             sessionStats={agentAtoms().sessionStatsAtom[0]()}
                             turnTokens={agentAtoms().turnTokensAtom[0]()}
                             launchPhase={status.launchPhase()}
--- a/frontend/app/view/agent/components/AgentFooter.tsx
+++ b/frontend/app/view/agent/components/AgentFooter.tsx
@@ -79,15 +79,6 @@ interface AgentWorkingRowProps {
      *  launchPhase to "waiting-for-login-completion" too, so without this the
      *  row rendered a second, redundant Cancel button alongside AuthUrlBox's. */
     hasAuthUrl?: boolean;
-    /** Unix ms the current attached-task episode began (reducer
-     *  `attachedTask.since`), or null when nothing long-running is attached.
-     *  When set and the turn is otherwise idle, the row shows a calm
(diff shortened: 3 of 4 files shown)


a5af:
@codex review

GenericAgentX-asaf:
## Handoff — AgentA

This PR closes out Camper's 2026-08-09/10 working-indicator + long-running-task session. Full history and rationale: `docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_09.md` (on main). What landed across the session: #2488 (agent-pane tab strip: stack tabs, close ×, rename), #2489 (attachedTask reducer axis wired end-to-end), #2493 (full command word-wrapped in expanded dock rows), #2494 (turn-end ghost-tool scrub), and this PR (instant "Working…", footer running-state removed).

### Design decision to preserve

**The ActivityDock is THE indicator for live attached tasks.** The footer "Running in background" state was removed here because it derived from the same aggregate as the dock rows and could never appear without the dock already showing a strictly richer version (command header, live log, stop button). Don't re-add a footer/status copy — extend the dock instead.

### The attachedTask axis now has state but no UI consumer — that's intentional, and these are its intended consumers

1. **Liveness-watchdog exemption** (`retro-persistent-agent-working-status-stuck-2026-07-16.md` fix-direction item 3): `StreamWatchdogTick`'s "quiet = hung" heuristic should not apply to a pane with `state.attachedTask != null`. The reducer field is right there; this is a small reducer-arm change + tests.
2. **Swarm / muxspect surfacing** (`SPEC_ATTACHED_TASK_STATUS_AXIS_2026_08_02.md` §6.4): expose the axis via block meta or the process-broker snapshot so cross-pane surfaces can show "this agent has a live task".

### Tracked follow-ups (the remaining ladder, in dependency order)

- **#2490** — thread `run_in_background` through `BashParams` + a dock adapter. Highest leverage: a harness-backgrounded call's ToolNode resolves in <1s, so `task dev`-style tasks are invisible to the dock AND to the attachedTask axis today. Until this lands, the axis only fires for MCP shells, ≥30s foreground calls, and subagents.
- **#2491** — bashwrap idle-timeout exemption for declared long-running tasks. Kills the 120s-heartbeat workaround, which is what pins panes on "Working…" (heartbeat < 180s watchdog threshold). Depends conceptually on #2490's "declared" bit.
- **#2492** — background tasks must survive agent-session teardown (observed twice this session: `task dev` died with the spawning session). Interacts with isolation invariants I1–I6. Also folds in Windows `started_at_ms` (hardcoded 0, `process_tracker/windows.rs:198`).

### Known remaining gaps in what shipped

- **Turn-end scrub (#2494) only fires on turn-end edges.** A ghost `running` ToolNode in a pane whose session died without a TurnEnd still waits for a session boundary (HistoryLoaded/SessionEnd). Possible extension: also dispatch the tools-only scrub from the mount-time `ReconcileTurnActive(false)` path.
- **Tab strip rename** (`#2488`) renames definition + block meta, but a *dormant* stack member's pill relies on the in-session `titleOverrides` echo; after an app restart the pill re-reads meta — verified in code, not live-tested across restart.
- **AgentPicker still calls `useAgentDefinitions()` independently** of AgentViewWrapper (reagent's P2 was fixed for AgentPresentationView only). Duplicate RPC exists transiently while a picker tab is showing. Low priority.
- **Live verification coverage:** new-tab +, close ×, composer fix, dock command header, and instant-Working were all verified live in the dev instance. NOT yet live-verified: rename round-trip across restart, the attachedTask axis firing for an MCP Shell (only code+unit-tested), and the turn-end scrub clearing a real ghost (the triggering ghost was still pending Smike's next turn-end at session close).

### Operational notes

- Dev-loop from an agent: `scripts/dev-agent.cmd` wrapped in the 120s-heartbeat pattern (`RETRO_TASK_DEV_IDLE_KILL_FALSE_POSITIVE_2026_07_31.md`), `run_in_background: true`. Stopping the dock's row for that shell kills the entire dev instance — known sharp edge.
- Pushing as an agent: `scripts/gh-agent.sh` for `gh`; for `git push`, resolve the PAT via `secrets get services/infra --path gh-token-<agent>` and pass inline. reagent findings on these PRs were consistently real — fix, don't dismiss; `pr merge --squash --auto` after it approves.

Issue #2368, opened 2026-07-30
Persistent controller: transparent stale-resume retry still flashes a visible 'Agent encountered an error' before the real response


AgentA-asaf:
## Context

Live-verified on the fresh portable built from PR #2360's merge (`0.54.7+g22dbe7ff9`), agent "Lazo", first message ("u there") sent after logging in on a session whose `--resume <sid>` was carried over from a different build/channel — the exact scenario PR #2360 (merged as `22dbe7ff9`) fixed.

## What the backend actually did (confirmed correct)

From `~/.agentmux/logs/agentmuxsrv-v0.54.7.log.2026-07-30`, block `bef65a18-b00f-4193-9358-3ec537905a35`:

1. `13:39:33.552` — "u there" sent; persistent process spawned with `--resume d46aa723-...`.
2. `13:39:35.475` — CLI stderr: `"No conversation found with session ID: d46aa723-..."`; process exits code 1.
3. `13:39:35.945` — `"stale --resume session id caused this exit — retrying fresh, without --resume"` — the PR #2360 retry fired correctly.
4. `13:39:35.959` — fresh process spawned (14ms later), no `--resume`.
5. `13:39:39.295` — fresh session ID captured (`11cb9d45-...`); turn completes normally ~2s later.

The message was **not** lost — the retry succeeded and got a real response ("Hey! What are we working on?"). This is exactly what PR #2360 was meant to guarantee, and it worked.

## The remaining gap

Despite the backend recovering transparently and quickly (14ms), the user still saw a visible **"Error" / "Agent encountered an error"** bubble in the transcript, appearing *before* the real response. So from the user's point of view this looked like "the same bug as before happened again," even though no data was actually lost this time.

This is the second, not-yet-addressed half of the original retro's explicit follow-up list (`docs/retros/RETRO_STALE_RESUME_SESSION_ID_ACROSS_CHANNELS_2026_07_29.md`, "Explicit follow-ups (not fixed here)"): *"it could transparently retry the same send once, without `--resume`, before ever reporting failure to the user — turning this into a silent (or disclosed-but-non-error) fresh-conversation start rather than a lost message."* PR #2360 implemented "don't lose the message, retry transparently" at the `persistent.rs` controller layer. It did **not** touch whatever renders the visible error bubble for the first (failed) attempt.

## Likely mechanism (needs confirmation, not yet traced end-to-end)

[...]

Issue #2367, opened 2026-07-30
PersistentSubprocessController: a confirmed stale-resume retry can DeliverDirect into an unrelated later spawn, reversing order


AgentA-asaf:
## Context

Found by reagentx/codex during PR #2360's sixth review pass, round 14 (commit 8c2bc99ab), inline comment at `persistent.rs:1268` ("Claim stale retries before accepting later sends"). Verified against actual code; deliberately not fixed inline in that PR — see reasoning below.

## The race

1. A process is spawned with `--resume <sid>` for a message (`decide_send_action` returns `BecomeSpawner`, `spawning_in_progress = true`).
2. The drain (`drain_queue_after_successful_spawn`) delivers that message via `tx.send(...).await`. Since the queue is then empty and `stdin_tx.is_some()`, the drain's own loop concludes "not stalled" and clears `spawning_in_progress = false` almost immediately — well before the process has any chance to write its stale-resume diagnostic to stderr and exit. This is expected/correct: once a process is alive, the common case is `DeliverDirect` for all further sends, same as any healthy long-running persistent process.
3. The process then actually dies (stale `--resume` confirmed via stderr). The process-waiter's exit-handling clears `stdin_tx = None`, takes `confirmed_stale_resume_retry`, and only THEN does non-trivial synchronous cleanup (`unregister_block`, `registry::remove`, `remove_agent`, `clear_active_pid`) BEFORE finally calling `retry_after_resume_failure` → `decide_retry_batch_action`.
4. During that cleanup window, `spawning_in_progress` is already `false` (cleared in step 2) and `stdin_tx` is `None`. A concurrent, unrelated `send_message` can see "not running, not spawning" → `BecomeSpawner`, spawn its OWN fresh process, and have its OWN drain fully deliver-and-clear before the original exit-handler ever reaches `decide_retry_batch_action`.
5. When the original exit-handler's `retry_after_resume_failure` finally calls `decide_retry_batch_action`, it now sees `stdin_tx.is_some() && !spawning_in_progress` (the CONCURRENT caller's fresh process) → `SendAction::DeliverDirect` — sending the chronologically OLDER, already-accepted stale-resume retry batch directly into that newer process, AFTER the newer, unrelated message was already delivered to it. This reverses accepted input order and could execute a dependent follow-up before its prerequisite.

## Why not fixed inline

[...]

Issue #2366, opened 2026-07-30
PersistentSubprocessController: respawn_once_for_leftover_queue's plain session_id clear can still race a lingering stdout-reader capture


AgentA-asaf:
## Context

Found by reagentx during PR #2360's sixth review pass, round 12 (commit d4cf8ac71). A defensive fix was attempted (poisoning instead of clearing) but had to be reverted in round 13 (commit fc0ece635 -> the revert) because it introduced a worse regression — see below.

`respawn_once_for_leftover_queue` clears `inner.session_id` before calling `spawn_process`, to force the fallback respawn to skip `--resume`. This fallback can be reached with no guarantee that `poison_resume` has already run for whatever process just died — the exit-handler's own bounded wait (500ms) for that process's stderr-reader task can time out and abort it before it ever calls `poison_resume`. In that case, the same process's stdout-reader task — which has no bound or abort mechanism at all — could still be concurrently racing to echo-capture that exact sid via `try_capture_session_id`, re-setting `inner.session_id` right after this fallback's plain clear and before `spawn_process`'s own `--resume` read, moments later. If that race is lost, this fallback respawn reattaches `--resume` to the dead sid and reproduces the original failure, with `resume_retry_payload: None` meaning nothing re-arms to catch the repeat — the message is silently dropped.

## Why not fixed directly

A first attempt called `inner.poison_resume(&old_sid)` unconditionally instead of a plain clear, reasoning that `poison_resume`'s `resume_poisoned` marker would block any later re-capture regardless of ordering. reagentx caught the regression this introduced: `respawn_once_for_leftover_queue` is reached from triggers that have nothing to do with a *confirmed* stale `--resume` — a plain `spawn_process` failure (missing binary, OS error) or any process crash with messages still queued. In both cases `inner.session_id` could just as easily be a legitimately hydrated-but-unattempted id, or a genuinely valid, already-captured session from a process that ran fine and crashed for an unrelated reason. `poison_resume` is documented as permanent (`resume_poisoned` is never reset) — poisoning a sid never actually confirmed dead by the CLI permanently breaks that session's resume capability, with no disclosure, for the rest of the controller instance's lifetime. That regression was worse than the narrow race it was meant to close, so it was reverted back to a plain clear.

## Suggested fix

[...]

Issue #2365, opened 2026-07-30
PersistentSubprocessController: content-based matching in decide_retry_batch_action can misidentify which queue entry is the retry's own trigger


AgentA-asaf:
## Context

Found by reagentx during PR #2360's sixth review pass (round 12), on commit d4cf8ac71.

`decide_retry_batch_action`'s prepend logic (added in round 11 to fix a reordering bug) checks whether the retry batch's own trigger message (`first`) is still sitting un-popped in `pending_send_messages` via content equality:

```rust
let first_already_queued = inner.pending_send_messages.iter().any(|m| m == first);
```

`QueuedMessage`'s `PartialEq` impls only compare the JSON payload string, with no notion of per-message identity. If a genuinely unrelated, later-arrived message (queued via `decide_send_action`, e.g. a second RPC or muxbus delivery) happens to have identical text to the retry batch's first entry — plausible for short, common replies like "yes", "ok", "continue" — this incorrectly treats it as "the same message, still queued," and splices the rest of the retry batch in right after that unrelated entry instead of prepending the whole batch ahead of the queue. The retry batch ends up misordered relative to newer input.

## Why not fixed directly in #2360

Neither position (checking only the front of the queue) nor the existing `already_persisted` flag on `QueuedMessage` can disambiguate this: `first`'s own entry, if genuinely still queued, was pushed via `decide_send_action`'s `BecomeSpawner` branch with `already_persisted: false` — identical to how a genuinely new, unrelated message is marked. A proper fix requires real per-message identity (e.g. a monotonically-increasing sequence number attached at push time), threaded through `QueuedMessage`, `pending_resume_retry`/`confirmed_stale_resume_retry`'s `Vec<String>` batches, and `retry_after_resume_failure`'s signature — a larger, more invasive change than the rest of this PR's fixes, which was already at its 12th review sub-round.

Assessed severity: lower than most of this PR's other fixes — the consequence is reordering (the retry batch and the coincidentally-matching later message both still get delivered, just possibly in the wrong relative order), not data loss. Requires a real content collision (identical JSON payload) between the retry's own trigger and a genuinely different, later message.

## Suggested fix

- Add a `sequence: u64` (or similar) field to `QueuedMessage`, assigned from a per-controller monotonic counter at push time.
[...]

PR #2496, opened 2026-08-10
docs: spec — spawn generations + message identity for the persistent-controller race cluster

AgentA-asaf:
Consolidated design for the five deferred follow-ups from PR #2360's stale-resume retry work (#2363, #2365, #2366, #2367, #2368).

Core insight: all five are shrapnel from two missing primitives — **spawn identity** (cleanup/retry decisions key on block_id alone, so they can't distinguish "the process I belong to" from "whatever is live now") and **message identity** (queue membership checked by content equality). The spec introduces a per-controller spawn-generation counter + per-message sequence numbers, then fixes each race against those primitives, honoring every "why not fixed inline" constraint documented in the issues (the round-13 poison revert, the round-14 spawning_in_progress starvation analysis).

Phased delivery: (1) generation-aware cleanup, (2) seq-based retry matching, (3) generation-aware retry decision, (4) attempted-sid-aware fallback, (5) the #2368 error-flash suppression — gated on live-repro confirmation of which channel renders the bubble.

Docs only — implementation PRs follow per phase.


a5af:
@codex review

PR #2497, opened 2026-08-10
feat(agent-pane): Undo for Esc-cleared composer drafts — right-click menu + Ctrl/Cmd+Z

GenericAgentX-asaf:
## Summary

Esc in the agent composer clears everything typed — and because the clear is a programmatic `.value` write, the browser's native undo stack can't restore it. An accidental Esc destroyed the draft with no recovery (user report 2026-08-10).

- **Snapshot on Esc-clear** (one level deep — Esc-clear is the only destructive programmatic edit; ordinary typing keeps native undo).
- **"Undo" in the composer's right-click menu** — `showTextInputContextMenu` gains an optional `leadingItems` param (rendered above the standard Cut/Copy/Paste block, menu shows even when the standard items are all disabled). Explicit `click` handler, no reliance on native menu roles.
- **Platform-standard shortcut**: Ctrl+Z (Windows/Linux) / Cmd+Z (macOS). Intercepted ONLY when the composer is empty and a cleared draft exists — every other state falls through to the browser's own undo, so normal text-edit undo is untouched. `keyutil.ts` already whitelists Cmd:z as an input-owned key; no keymodel conflict.

## Verification

- `tsc --noEmit` clean.
- Live verification pending: the dev instance was down when this was authored; a fresh `task dev` is coming up now — will confirm type → Esc → Ctrl+Z restore and the right-click Undo entry there before/shortly after merge.

### DIFF
--- a/.changesets/1786337657-feat-agent-pane-undo-for-esc-cleared-composer-drafts-right-c-9jcl.md
+++ b/.changesets/1786337657-feat-agent-pane-undo-for-esc-cleared-composer-drafts-right-c-9jcl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): Undo for Esc-cleared composer drafts — right-click menu + Ctrl/Cmd+Z
--- a/frontend/app/store/contextmenu.ts
+++ b/frontend/app/store/contextmenu.ts
@@ -118,17 +118,29 @@ async function getClipboardURL(): Promise<URL | null> {
     }
 }
 
-async function showTextInputContextMenu(e: MouseEvent): Promise<void> {
+/**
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed both reagent's P1 and codex's P2 — same root cause: `escClearedDraft` was only cleared by the restore itself, so a stale snapshot could resurface via Ctrl/Cmd+Z after a send or after typing-then-deleting back to empty.

Now invalidated in `handleInput` (any real keystroke), `setComposerValue` (ghost-text accept, history recall), and `handleSend`. Added `AgentFooter.test.tsx` with 3 regression tests — confirmed 2 of them fail against the pre-fix code.

PR #2499, opened 2026-08-10
docs(specs): session-scoped pane scrollback + Agent History view

AgentX-asaf:
## Summary

Spec for making the agent pane's working scrollback reflect the agent's **actual knowledge** — after a `fresh` session boundary the pane shows only post-boundary content, with a link row into a new read-only, per-agent **Agent History** view (full retained history, day separators, session dividers). Display-scope only; nothing stored changes.

Extends `SPEC_AGENT_PANE_HISTORY_ALIGNMENT_2026_08_05.md` (Part A's persisted `agentmux_session_outcome` line is the scope anchor). Orthogonal to the account-wide unified-history-store spec — this is the in-pane surface.

Second commit adds two refinements diagnosed live on 2026-08-09:

- **§3.5 — demote the "Session continued" divider.** `resumed` outcomes fire on user-invisible process recycles and (because emission waits for `is_confirmed_success`) land *below* the first post-resume exchange. Working view renders only `fresh`; `resumed` stays persisted and shows subtly in Agent History.
- **§4.4 — node timestamps become first-class.** Additive `output.tsidx` receive-time sidecar stamped at append time, wired into replay — fixes the shipped-but-starved hover-to-peek time display (gated on `node.timestamp != null`, which restored nodes never have), gives day separators exact edges, and kills the `timestamp: 0` → 1970 sentinel leak.

## Verification

Docs-only. All cited symbols/paths verified against `main` @ `9720ce4c2` (emission arms in `persistent_resume.rs:283-363`, restore path in `useHistoryPagination.ts`, hover gate in `UserMessageBlock.tsx`/`ToolBlock.tsx`, unused archive RPCs).

### DIFF
--- a/docs/specs/SPEC_AGENT_PANE_SESSION_SCOPED_SCROLLBACK_AND_AGENT_HISTORY_VIEW_2026_08_09.md
+++ b/docs/specs/SPEC_AGENT_PANE_SESSION_SCOPED_SCROLLBACK_AND_AGENT_HISTORY_VIEW_2026_08_09.md
@@ -1,0 +1,375 @@
+# SPEC: Session-scoped pane scrollback + a full "Agent History" view
+
+**Date:** 2026-08-09
+**Status:** Proposed
+**Severity:** Medium — UX/correctness follow-up, no data loss involved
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2501, opened 2026-08-10
fix(persistent): retry batch flushes under a drain claim — the queue is the ordering authority (PR B, #2367)

### LINKED ISSUE #2367: PersistentSubprocessController: a confirmed stale-resume retry can DeliverDirect into an unrelated later spawn, reversing order

AgentA-asaf:
## Summary

**PR B** of `docs/specs/SPEC_PERSISTENT_SPAWN_GENERATION_AND_MESSAGE_IDENTITY_2026_08_09.md` (§4, **option 2** — the spec's preferred resolution). Follow-up to #2500 (PR A). Closes #2367.

### The bug

`decide_retry_batch_action` had no generation input: when a stale-resume retry fired and an unrelated NEWER spawn had already installed a live `stdin_tx` during the exit-handler's cleanup window, the batch was `try_send`-delivered straight into that newer process from the caller — and a concurrent `send_message` (also seeing `DeliverDirect`) could jump ahead of, or interleave into the middle of, messages the doomed process had already accepted.

### The fix (spec §4 option 2)

- `decide_retry_batch_action(retry_generation, ...)` returns its own `RetryBatchAction`. The live-process case is by construction a newer spawn (the retry's own process exited; its exit-handler clears `stdin_tx` under the generation gate — debug-asserted) and resolves to **`FlushClaimed`**: the batch is prepended to the queue AND a new `drain_claim` flag is taken **in the same lock acquisition**.
- While `drain_claim` is held, `decide_send_action` routes `DeliverDirect` → `Queued` — exactly how it already treats `spawning_in_progress` — so the queue is the single ordering authority. `spawning_in_progress` semantics are untouched (spec non-goal; round-14 starvation analysis stands — the round-7/8 borrow of the spawn flag for a non-spawn is retired).
- The flush reuses the existing queue-drain loop via `drain_queue_with_claim(.., QueueDrainClaim)` — one parameterized release point, so `Sender::send` backpressure, seed-aware retry-batch appends, delivery-order persistence and stall handling are shared, not duplicated. `allow_fallback_respawn: false` (a stall means the target process died; leftovers stay queued and the stalled branch publishes status).
- The accepted residue matches the spec: only a `DeliverDirect` *decided before* the claim was taken can land mid-batch — that send was already racing the process exit itself.

### Bonus: a strand gap in the drain loop itself

The new ordering test exposed a pre-existing hole affecting the spawn-claim drain too: a message enqueued between the drain observing an empty queue and releasing its claim (two separate lock acquisitions) was stranded — no claim holder left to drain it, and every future send `DeliverDirect`s past it while the process lives. The release now re-checks the queue under its own lock and loops back to drain instead.

## Verification

- Full `cargo test -p agentmux-srv`: **2097 passed, 0 failed** (plus 4 + 7 integration). One unrelated parallel-run flake (`m0020_agent_color_backfill`), passes in isolation and on re-run — same flake the #2488/#2489 session noted.
- New tests: the retry takes `drain_claim` and never the spawn flag, releases it against a dead sender with the message preserved; a concurrent send during a flush is `Queued` and arrives strictly after the entire batch; existing batch redelivery/ordering/dedup tests pass unchanged.

Remaining from the spec after this: #2368 (live stale-resume repro to close with evidence).

### DIFF
--- a/.changesets/1786347745-fix-persistent-retry-batch-flushes-under-a-drain-claim-queue-xn2c.md
+++ b/.changesets/1786347745-fix-persistent-retry-batch-flushes-under-a-drain-claim-queue-xn2c.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(persistent): retry batch flushes under a drain claim — queue is the ordering authority (#2367)
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -225,6 +225,21 @@ struct PersistentInner {
     /// spawn or arrived while someone else's spawn was already in flight.
     /// Drained by `release_spawn_claim_and_drain_queue`.
     pending_send_messages: VecDeque<QueuedMessage>,
+    /// Exclusive claim held by a stale-resume retry batch flush (issue
+    /// #2367; spec §4 option 2 of
+    /// SPEC_PERSISTENT_SPAWN_GENERATION_AND_MESSAGE_IDENTITY_2026_08_09).
+    /// Taken by `decide_retry_batch_action` in the SAME lock acquisition
+    /// in which it decides a live, newer-generation process should
+    /// receive the batch; while held, `decide_send_action`'s
+    /// `DeliverDirect` branch routes to `Queued` instead (exactly how it
+    /// already treats `spawning_in_progress`), making the queue the
+    /// single ordering authority — a concurrent send can no longer
+    /// `try_send` ahead of earlier-accepted retry messages. Released by
+    /// the flush task (`drain_queue_with_claim`) once the queue runs dry
+    /// or the process dies. Deliberately a NEW flag, not a reuse of
+    /// `spawning_in_progress` — round-14's starvation analysis stands:
+    /// that flag is never pre-asserted while no spawn is in progress.
+    drain_claim: bool,
     /// Monotonic counter behind `QueuedMessage::seq` (issue #2365) —
     /// pre-incremented at every fresh enqueue, so real seqs start at 1
     /// and are never reused for this controller's lifetime. Redelivery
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2505, opened 2026-08-10
feat(agent): introduce model vendor as a concept distinct from harness

GenericAgentX-asaf:
## Summary

AgentMux's "provider" concept conflates two independent things: the **harness** (the CLI driving a session — `claude`, `codex`, `gemini`...) and the **model vendor** (the LLM backend actually serving responses — Anthropic, a proxy, Bedrock...). Claude Code already supports `ANTHROPIC_BASE_URL` to point at a non-Anthropic-hosted backend, so harness ≠ vendor in reality — but the data model had no explicit "model vendor" concept. Confirmed via full-repo grep: `ANTHROPIC_BASE_URL` appeared nowhere in the codebase before this PR.

A generic, unstructured escape hatch already existed (`CommandAgentDefineData.env`, a free-form `KEY=VALUE` env blob merged at spawn time), so the gap wasn't mechanical — it was that there was no **explicit, discoverable, validated, first-class concept**: no way to query which agents use a non-default vendor, no validation that the chosen provider even supports redirection, nothing typed or surfaced in `agent.define`'s result.

## What this adds

- **`ProviderConfig.base_url_env_var`** (Rust) / **`ProviderDefinition.baseUrlEnvVar`** (TS): declares which env var (if any) a harness reads to redirect its backend. Only set for `claude` (`ANTHROPIC_BASE_URL`) — the one case independently verified as real, documented Claude Code behavior; explicit `None`/omitted everywhere else rather than guessed.
- **`AgentDefinition.model_vendor_base_url`**: per-agent override, schema v15 — added to **both** `db_agent_definitions` and `db_agents` (the dual-write mirror table `agent_def_list` actually reads from; a single-table ALTER would have silently no-op'd the feature after the first read).
- **Validation** in `agent_define_core`: rejects a non-empty override unless the resolved provider declares `base_url_env_var` support — a pure, unit-tested `validate_vendor_base_url` helper.
- **Actual env-var injection at spawn time, on BOTH paths**: Rust (`agent_open.rs`, the `agent.open` App-API/MCP path) and TypeScript (`agent-model.ts`'s `launchAgentDefinition`, the human "Launch Agent" UI path) — the feature works for real agent launches, not just MCP-spawned ones.

## Deliberately out of scope

This traces back to a spec found on an unmerged branch (`feat/harness-model-decoupling-antigravity`) whose core architectural observation is sound and independently verified here, but whose scope (onboarding a new "Antigravity" CLI harness with a hardcoded `--yolo` auto-approve flag) was explicitly rejected. This PR:
- **No new harness/CLI onboarding** — no Antigravity integration, no `--yolo`, no bypassing the tool-permission gate. Formalizes the concept using only providers AgentMux already trusts.
- **No UI wizard rework** — settable via `agent.define` only; the human "New Agent"/edit UI doesn't surface it yet (`createagent`/`updateagent` don't carry the field — `updateagent` explicitly preserves the existing value so a UI-driven save never silently wipes an override set via `agent.define`).
- **No cross-channel registry mirror** — channel-local only for now; a cross-channel reopen of the same agent won't carry the override. Documented limitation in `def_registry_mirror.rs`, not an oversight.

## Test plan

- [x] `cargo check --workspace --tests` — clean (all 6 workspace crates)
- [x] `cargo test -p agentmux-srv` — 2107 passed, 0 failed (includes 2 new `providers.rs` tests, 4 `agent_define.rs` validation tests, 3 `agent_open.rs` env-resolution tests, 1 schema test asserting the new column on both tables + idempotency)
- [x] `npx vitest run` — 2468 passed (includes 3 new `vendor-env.ts` tests; `pin-consistency.test.ts`/`provider-id-aliases.test.ts` — the drift guards between the Rust and TS provider catalogs — still pass)
- [ ] Not manually smoke-tested via `task dev` (full CEF rebuild) — verification here is thorough unit coverage of every new code path (validation, both injection sites, schema) rather than a live UI walkthrough; flagging this explicitly rather than claiming untested things work

### DIFF
--- a/.changesets/1786370573-feat-agent-introduce-model-vendor-as-a-concept-distinct-from-6686.md
+++ b/.changesets/1786370573-feat-agent-introduce-model-vendor-as-a-concept-distinct-from-6686.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): introduce model vendor as a concept distinct from harness
--- a/agentmux-srv/src/backend/agent_seed.rs
+++ b/agentmux-srv/src/backend/agent_seed.rs
@@ -196,6 +196,7 @@ pub fn seed_agents(wstore: &Arc<Store>) -> Result<SeedReport, StoreError> {
             container_volumes: "[]".to_string(),
             container_name: String::new(),
             use_ambient_login: 0,
+            model_vendor_base_url: String::new(), // manifest doesn't declare this yet
         };
         wstore.agent_def_insert(&mut agent)?;
 
@@ -442,6 +443,7 @@ fn reseed_if_needed(
             container_volumes: "[]".to_string(),
             container_name: String::new(),
             use_ambient_login: 0,
+            model_vendor_base_url: String::new(), // manifest doesn't declare this yet
         };
 
         if let Some(existing_agent) = existing_map.get(agent_def.id.as_str()) {
@@ -451,6 +453,7 @@ fn reseed_if_needed(
             agent.provider = existing_agent.provider.clone();
             agent.agent_type = existing_agent.agent_type.clone();
             agent.environment = existing_agent.environment.clone();
+            agent.model_vendor_base_url = existing_agent.model_vendor_base_url.clone();
             agent.shell = if existing_agent.shell.is_empty() {
                 agent_def.shell.clone()
             } else {
@@ -549,6 +552,7 @@ mod tests {
             container_volumes: "[]".to_string(),
             container_name: String::new(),
             use_ambient_login: 0,
+            model_vendor_base_url: String::new(),
         };
         wstore.agent_def_insert(&mut def).unwrap();
     }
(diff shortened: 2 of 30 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed the P1 — added the missing changeset (`task changeset -- minor "..."`), pushed as a follow-up commit. Missed it on the first push; caught by this exact review.

GenericAgentX-asaf:
Fixed both — both were real, not nitpicks:

- **P0:** `agent_def_list`'s global overlay was zeroing `model_vendor_base_url` on every read for ANY agent also mirrored into the global registry — which is the normal case (every `agent.define` write auto-mirrors). Confirmed this made the feature a no-op even in default single-instance operation, not just genuinely cross-channel agents. Fixed by preserving the local row's value before the overlay insert.
- **P1:** the update path couldn't distinguish "don't touch" from "clear," and a provider change while a stale override sat there could permanently lock out all future edits to that agent (no way to ever un-set it). Switched `model_vendor_base_url` to `Option<String>` (`None` = don't touch, `Some("")` = explicit clear), mirroring the existing `use_ambient_login` idiom in this codebase.

4 new regression tests (1 for the P0 overlay case, 3 for the P1 clear/don't-touch/recovery cases), all confirmed red-before/green-after against the pre-fix code. Full suite still green: 2111 Rust tests, full workspace `cargo check --tests` clean.

PR #2507, opened 2026-08-10
feat(agent-pane): session-scoped scrollback — clamp at fresh session boundary, demote 'Session continued' divider

AgentX-asaf:
## Summary

P1 of `SPEC_AGENT_PANE_SESSION_SCOPED_SCROLLBACK_AND_AGENT_HISTORY_VIEW_2026_08_09.md` (merged as #2499): the working pane now reflects the agent's **actual knowledge**.

- **Session-scope clamp (§3.1-3.3).** The visible document is clamped at the newest `fresh` session-outcome boundary. One reducer helper (`clampToSessionScope`) funnels all three merge paths — `HistoryRestored`, `HistoryLoaded`, `StreamFlush` — so restore, backward pagination, and live arrival all obey the same rule. The live path clamps *inside the same reduction* that lands the boundary node, making the trim race-free against same-batch pre-boundary nodes. `useHistoryPagination` stops paging past the boundary in all three of its paths. Display-scope only: the persisted stream (per-channel blockfile + global mirror) is untouched, and legacy streams with no boundary render exactly as before — no migration.
- **"Session continued" demoted (§3.5).** `resumed` outcomes fire on user-invisible process recycles and (because emission waits for `is_confirmed_success`) land *below* the first post-resume exchange — a mis-placed divider announcing a non-event. Both parse paths (`parseHistoryLines`, `useAgentStream`) now skip materializing them; the persisted line stays for diagnostics and for the P2 Agent History view, which will render them subtly.
- The fresh divider's detail line now reads "…preserved in the agent's history" (the §3.4 link row ships with P2, per the spec's P1 phasing note).

## Verification

- 9 new reducer tests (clamp-at-boundary, newest-wins, `resumed`-not-an-anchor, strictly-older-page drop, mid-page boundary, live trim incl. same-batch nodes, idempotent re-flush, id-set/index rebuild) + 2 new `parseHistoryLines` tests — all passing.
- Full affected suites: reducer + parseHistoryLines + session-outcome (125 tests), virtualization + useAgentStream + hooks (241 tests, 17 files) — all passing. `tsc --noEmit` clean.
- StreamFlush hot path unchanged unless a batch actually carries a fresh boundary (one `some()` scan of the small batch otherwise).

### DIFF
--- a/.changesets/1786370383-feat-agent-pane-session-scoped-scrollback-clamp-at-fresh-ses-4ubt.md
+++ b/.changesets/1786370383-feat-agent-pane-session-scoped-scrollback-clamp-at-fresh-ses-4ubt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): session-scoped scrollback — clamp at fresh session boundary, demote 'Session continued' divider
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

PR #2508, opened 2026-08-10
feat(agent-pane): output.tsidx receive-time sidecar — replayed transcript nodes get real timestamps

AgentX-asaf:
## Summary

First slice of P2 of `SPEC_AGENT_PANE_SESSION_SCOPED_SCROLLBACK_AND_AGENT_HISTORY_VIEW_2026_08_09.md` (§4.4): **node timestamps become first-class and survive restore.**

**Backend.** Every agent-transcript append batch now also appends one NDJSON `{off, ms}` record to an `output.tsidx` sidecar in the same zone — covering `handle_append_block_file`, `persist_to_blockfile_silent` (user-message lines), and the global-zone mirror (offset-keyed against the global zone's own `output`). Deliberately a sidecar, not an envelope: `output`'s format and every existing parser are untouched, writes are fire-and-forget, and PTY `term` data never gets one (gated on `global_output_zone`). `blockfile:read_range`'s `output.idx` fast path joins the sidecar against each returned line's byte offset (one slice read + binary search) and returns an optional `stamps` array parallel to `lines` — old frontends ignore it, new frontends tolerate its absence.

**Frontend.** `parseHistoryLines` stamps freshly-created nodes from their line's batch time when the wire frame carries no timestamp; a wire timestamp always wins; same-id replacements (`tool_call` → `tool_result`, accumulated deltas) carry the stamp forward instead of wiping it; `0`/absent means unknown and never stamps a node — closing the 1970-hover sentinel hole. `useHistoryPagination` threads `stamps` through all three read paths.

**Net effect:** the hover-to-peek time display (#2392 lineage) — dead after any pane reopen because replayed nodes had no timestamps — now works on restored history for all post-upgrade content. Legacy pre-sidecar lines degrade to today's behavior (no backfill is possible; their receive times were never recorded). This is also the data source for the upcoming Agent History view's day separators.

## Verification

- Backend: 4 new tests (per-batch stamping + offset keying, term-data exclusion, global-mirror stamping, persist-silent stamping); full `cargo test -p agentmux-srv` suite green (2101 passed).
- Frontend: 5 new tests (batch stamping, wire-timestamp-wins, replacement stamp carry, 0-sentinel hygiene, absent-array no-op); affected suites green (142 tests); `tsc --noEmit` clean.
- Sidecar growth is one ~30-byte line per append batch; the read-side join reads it whole — worth a size cap/rotation only if profiling ever shows it hot (noted for P3 alongside archive rotation).

### DIFF
--- a/.changesets/1786371477-feat-agent-pane-output-tsidx-receive-time-sidecar-replayed-t-wkfs.md
+++ b/.changesets/1786371477-feat-agent-pane-output-tsidx-receive-time-sidecar-replayed-t-wkfs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): output.tsidx receive-time sidecar — replayed transcript nodes get real timestamps (hover peek works after restore)
--- a/agentmux-srv/src/backend/agent_session/mod.rs
+++ b/agentmux-srv/src/backend/agent_session/mod.rs
@@ -37,7 +37,7 @@ pub use global_store::{
 pub use migrations::{migrate_block_zones_v1, migrate_promote_template_sessions_v1};
 pub use session_io::{
     append_session_output, heal_global_snapshot_source_block_ids, read_session_state,
-    write_session_state, OUTPUT_FILE, SNAPSHOT_FILE,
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

PR #2509, opened 2026-08-10
feat(agent-pane): Agent History view — read-only full-history reader with day separators

AgentX-asaf:
## Summary

Completes P2 of `SPEC_AGENT_PANE_SESSION_SCOPED_SCROLLBACK_AND_AGENT_HISTORY_VIEW_2026_08_09.md` (on top of #2507's session-scoped clamp and #2508's tsidx timestamps): the full retained history is now **perusable in-pane**.

- **`bodyMode` swap (§4.2).** The agent pane gains `"live" | "history"`; history mode swaps the transcript + composer subtree for `AgentHistoryView` and unmounts the live subtree (stream subscription never doubled). Not a new pane type/widget — a mode of the agent pane; always reopens live.
- **`AgentHistoryView` (§4.1/4.3).** Read-only, boundary-blind reader over the agent's full transcript stream. Deliberately bypasses the reducer store — the working view's session-scope clamp must not apply here, and a read-only reader has none of the races the store exists to serialize. Plain local signals fed by `parseHistoryLines` (new `includeResumedOutcomes` option — `resumed` markers render here as subtle restart landmarks while staying demoted in the working view), rendered through the reused `AgentDocumentView` + virtual list under a synthetic layout key (`<blockId>:history`) so the live pane's layout slot is untouched. Pages 600 lines at a time back to line 0.
- **Day separators (§4.4).** New `day_divider` render-time synthetic node injected wherever the best-known timestamp (own, else carried forward) crosses a local calendar day — exact for post-#2508 content, session-start-day granularity for pre-sidecar history. Stable `day-<YYYY-MM-DD>` ids merge across pagination recomputes.
- **Entry points (§3.4 + §4.2).** The scope clamp is no longer silent: `useHistoryPagination` surfaces `scopeClamped`, which gates an "Earlier conversations — Open Agent History" `PaneRow` above the working scrollback. Plus a provider-agnostic "View full history" button in `AgentControlBar` (rendered before the Claude-only gate, so non-Claude panes get it too).

Remaining after this: P3 only (archives section wiring, jump-to-date, link-row session count).

## Verification

- New tests: 5 `injectDayDividers` cases (day-change insertion, carry-forward, leading-unknown, id stability across prepends, empty) + `includeResumedOutcomes` parse case.
- Full agent-view suite: **1084 tests / 84 files pass**; `tsc --noEmit` clean. No backend changes.
- Live-drive note: the view composes the existing virtual list/renderers against the same RPCs the live pane already exercises; a `task dev` walkthrough (open history from the link row and the control bar, scroll to line 0, day chips + resumed markers visible, back to live resumes streaming) is the remaining manual check — happy to run it on the next dev-instance session.

### DIFF
--- a/.changesets/1786372533-feat-agent-pane-agent-history-view-read-only-full-history-re-74vs.md
+++ b/.changesets/1786372533-feat-agent-pane-agent-history-view-read-only-full-history-re-74vs.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): Agent History view — read-only full-history reader with day separators; 'Earlier conversations' link row
--- a/docs/specs/SPEC_AGENT_PANE_SESSION_SCOPED_SCROLLBACK_AND_AGENT_HISTORY_VIEW_2026_08_09.md
+++ b/docs/specs/SPEC_AGENT_PANE_SESSION_SCOPED_SCROLLBACK_AND_AGENT_HISTORY_VIEW_2026_08_09.md
@@ -1,7 +1,7 @@
 # SPEC: Session-scoped pane scrollback + a full "Agent History" view
 
 **Date:** 2026-08-09
-**Status:** active — P1 (session-scope clamp §3.1-3.3 + `resumed`-divider demotion §3.5 + static preserved-note on the fresh divider) shipped in PR #2507; §4.4's `output.tsidx` capture + replay stamping implemented 2026-08-10 (this restores hover-peek times after reopen). Remaining: the Agent History view itself (§4.1-4.3, day separators, §3.4 link row) and P3.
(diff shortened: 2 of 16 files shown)


a5af:
@codex review

AgentX-asaf:
Addressed all five findings in 46bd7c629: codex P1 (wholesale reparse — per-page parser ids collided and the dedup dropped real history), codex P2 (link row now derives from the document's first node, covering the live clamp), reagent P1×2 ('View full history' is an unconditional control-bar row; search bar moved inside the live branch + Ctrl+F no-ops in history mode), reagent P2 (explicit loading state). Full agent-view suite 1084 pass, tsc clean.

AgentX-asaf:
Rebased onto main (which had picked up #2510 in the meantime — real conflict in `parseHistoryLines.ts` between #2510's fresh-boundary stats reset and this branch's `includeResumedOutcomes` option, both preserved). Re-verified after merge: backend suite 2117 pass, full agent-view suite 1089 pass (85 files), tsc clean. Ready for re-review.

PR #2510, opened 2026-08-10
fix(agent-pane): address codex P2s from #2507/#2508 — stats reset, tsidx archive lifecycle, exact stamp offsets

AgentX-asaf:
## Summary

Follow-up addressing the three codex P2 review comments that landed on #2507/#2508 after they merged (they were unanswered — the review-notification jekts only relay reagent reviews; codex comments need an explicit PR check, which is now part of my merge routine).

1. **Stats reset at a fresh boundary** ([#2507 comment](https://github.com/agentmuxai/agentmux/pull/2507#discussion_r-)): `parseHistoryLines` now nulls `lastSessionStats` when it materializes a `fresh` session outcome — a restore window shaped [old usage-bearing result → fresh boundary → no new result] was hydrating the context-fill bar with the dead session's token count. Post-boundary usage still hydrates.
2. **tsidx through the archive lifecycle** (#2508 comment 1): the sidecar now travels with `output` everywhere — `agent_session::archive_session` copies it into the archive zone (per-channel + global-source paths) and clears it from `:current` in all three clear paths; `session_archive` archives it to `<block>.tsidx.gz`, deletes it alongside `output`, and restores it on session restore. Without this, a stale sidecar under a fresh offset-0 `output` mis-times the next session's lines, and archived history loses its stamps.
3. **Exact stamp offsets** (#2508 comment 2): new `FileStore::append_data_at` performs the size read and part writes under one connection lock and returns the offset the batch actually landed at; all three tsidx stamp sites use it instead of the racy pre-append stat. `append_data` becomes a delegating wrapper — zero behavior change for existing callers. (The broadcast event's `offset` field keeps the pre-append stat — pre-existing contract, reconcile-only consumer, unchanged.)

## Verification

- New tests: fresh-boundary stats reset + post-boundary rehydration (frontend, 2), `archive_carries_and_clears_tsidx_sidecar`, `tsidx_offsets_match_actual_append_positions` (backend, 2).
- Full backend suite: 2103 passed. parseHistoryLines suite: 22 passed. `tsc --noEmit` clean.

### DIFF
--- a/.changesets/1786373216-fix-agent-pane-address-codex-review-on-2507-2508-stats-reset-vgfo.md
+++ b/.changesets/1786373216-fix-agent-pane-address-codex-review-on-2507-2508-stats-reset-vgfo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): address codex review on #2507/#2508 — stats reset at fresh boundary, tsidx archive lifecycle, exact stamp offsets
--- a/agentmux-srv/src/backend/agent_session/archive.rs
+++ b/agentmux-srv/src/backend/agent_session/archive.rs
@@ -7,7 +7,7 @@ use crate::backend::storage::filestore::FileStore;
 
 use super::global_store::global_transcript_store;
 use super::helpers::{now_ms, write_zone_file};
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #2512, opened 2026-08-10
fix(lan-discovery): registered instances were never actually announced on the wire

GenericAgentX-asaf:
## Summary

Investigating "why can't I reach AgentA on Area54 via LAN jekt" surfaced a real bug in LAN peer discovery itself, not a network/environment problem.

Diagnosed empirically before touching any code:
- Confirmed `Area54` is genuinely network-reachable (mDNS hostname resolution + ping succeed, same L2 segment as the real Ethernet adapter — not a VM/WSL virtual one).
- Confirmed local Windows Firewall isn't blocking anything (agentmux-srv already has broad allow rules: inbound UDP+TCP, any port, Private+Public profiles).
- Confirmed AgentMux's own LAN discovery daemon starts cleanly with no error logged (`"LAN discovery started (mDNS)"`).
- Confirmed basic mDNS multicast fundamentally works on this machine — an independent Python `zeroconf` client successfully advertised and discovered its own test service.
- **But browsing for AgentMux's own `_agentmux._tcp.local.` service type found nothing — not even this instance's own self-advertisement**, despite the "started successfully" log line.

## Root cause

`LanDiscovery::start()` calls:
```rust
ServiceInfo::new(SERVICE_TYPE, &service_name, &host_name_mdns, "", port, &properties[..])
```
with the comment `// empty = auto-detect IP`. That's not what the `mdns-sd` crate does. `AsIpAddrs for &str`'s own doc comment: *"If the string is empty, will return an empty set."* Auto-detection is a **separate opt-in** — `ServiceInfo`'s `.enable_addr_auto()` builder method — which this code never called. `ServiceInfo::new` also hardcodes `addr_auto: false`.

Net effect: every registered service had **zero addresses** and no instruction to the daemon to ever fill them in. `ServiceDaemon::register()` has no synchronous validation for this — it just enqueues a `Command::Register` and returns `Ok(())` — so startup logged success on every boot while the service was never actually discoverable by anyone, including another mDNS client on the same host.

## Fix

Chain `.enable_addr_auto()` onto the `ServiceInfo` builder.

**Both machines need this fix to see each other** — registration was broken on whichever side runs this code, so a peer still running the old build won't be found until it's updated too.

## Test plan

- [x] New test: `a_registered_instance_is_discoverable_by_an_independent_mdns_client` — registers a real `LanDiscovery` instance (the exact production code path) and browses for it with a **separate, independent** `mdns_sd::ServiceDaemon` (a stand-in for a real peer's browse side) over real loopback multicast. Confirmed **failing** (5s timeout, never found) against the pre-fix code; **passing** (found in <1s) after the fix.
- [x] `cargo test -p agentmux-srv` — 2112 passed, 0 failed (includes all 22 `lan_discovery` tests)

### DIFF
--- a/.changesets/1786396855-fix-lan-discovery-registered-instances-were-never-actually-a-j0u7.md
+++ b/.changesets/1786396855-fix-lan-discovery-registered-instances-were-never-actually-a-j0u7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(lan-discovery): registered instances were never actually announced on the wire
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -201,15 +201,29 @@ impl LanDiscovery {
             ("instance_id", instance_id.as_str()),
             ("auth_key", auth_key.as_str()),
         ];
+        // `""` alone does NOT mean "auto-detect" despite how that reads —
+        // `AsIpAddrs for &str`'s own doc comment: "If the string is empty,
+        // will return an empty set." Auto-detection is a SEPARATE opt-in:
+        // `.enable_addr_auto()` tells the daemon to fill in (and keep
+        // updated) this service's addresses from the host's own interfaces.
+        // Without it, a service registered via `ServiceInfo::new(..., "",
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2514, opened 2026-08-10
feat(agent-pane): AskUserQuestion countdown hides on hover, resumes after 15s

Agent1-asaf:
## Summary

Refines the AskUserQuestion 30s auto-timeout (#2441) so hovering the panel
suppresses the countdown instead of leaving it ticking under the mouse:

- Hovering the panel hides the countdown chip entirely ("show nothing"),
  for a flat 15s window timed from that hover.
- If nothing was submitted by the time the window elapses, the countdown
  resumes at a fresh 30s — **unconditionally**, regardless of whether the
  mouse is still over the panel.
- A fresh `mouseenter` (the mouse genuinely leaving and coming back)
  restarts the window from that point — repeatable indefinitely.

## Why bounded, not open-ended

The design originally drafted kept the countdown hidden for as long as the
mouse stayed over the panel (grace only started on `mouseleave`). A
pre-existing test (`merges: keeps a manually-answered question...`) caught
the regression before this ever reached a PR: `userEvent.click()` on any
option fires a real `mouseenter` on the way to the click, because the
pointer has to already be over the target — true in a real browser exactly
as much as in the test harness. Under the open-ended design, a user who
answers by mouse click and then steps away (cursor left parked over the
panel, `mouseleave` never fires) would have the safety net paused
indefinitely — reintroducing the exact "work does not stop" failure
`SPEC_ASK_USER_QUESTION_AUTO_TIMEOUT_2026_08_06.md` §5.1 already rejected,
just through hover instead of a permanent disarm.

Fixed by making the hide window flat and unconditional: worst case is
bounded at exactly 15s per fresh hover, never indefinite. Full account in
`docs/specs/SPEC_ASK_USER_QUESTION_TIMEOUT_HOVER_PAUSE_2026_08_10.md` §9.

## Test plan

- [x] `AgentQuestionPanel.test.tsx` — 21/21 passing, including the new
      hover-pause suite and the updated pre-existing "merges" test
- [x] Full `frontend/app/view/agent/` suite — 85 files / 1095 tests passing
- [ ] Manual `task dev` verification of live hover behavior (not run in
      this sandbox — no display)

### DIFF
--- a/.changesets/1786399858-feat-agent-pane-askuserquestion-countdown-hides-on-hover-res-3yf1.md
+++ b/.changesets/1786399858-feat-agent-pane-askuserquestion-countdown-hides-on-hover-res-3yf1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): AskUserQuestion countdown hides on hover, resumes after 15s
--- a/docs/specs/SPEC_ASK_USER_QUESTION_TIMEOUT_HOVER_PAUSE_2026_08_10.md
+++ b/docs/specs/SPEC_ASK_USER_QUESTION_TIMEOUT_HOVER_PAUSE_2026_08_10.md
@@ -1,0 +1,467 @@
+# SPEC: Hover-pause for the AskUserQuestion auto-timeout countdown
+
+**Date:** 2026-08-10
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2515, opened 2026-08-10
fix(agent-pane): ghost-text next-prompt suggestion restores when composer is cleared

Agent1-asaf:
## Summary

Fixes the composer's ghost-text next-prompt suggestion (`term:next_prompt_suggestion`)
so it survives being typed over and cleared, instead of being permanently
dropped on the first keystroke.

**Bug:** type one character into the composer while a suggestion is showing,
then delete it back to empty — the suggestion is gone for good, falling back
to "Send message to `<agent>`..." instead of reappearing.

**Root cause:** `handleInput` unconditionally nulled `term:next_prompt_suggestion`
from block meta the moment the box went from empty to non-empty (and
Tab-accept did the same). That was a deliberate implementation of the
original ghost-text spec's §4.3 — but that section conflated two separate
concerns into one bullet. The actual race it names (a late suggestion-RPC
response overwriting text the user already started typing) is fully closed
by an independent guard already in `useNextPromptSuggestion.ts`
(`isComposerEmpty()`, checked at RPC-response time, not tied to anything in
`AgentFooter.tsx`). Clearing meta on every first keystroke was redundant for
that purpose — its only real effect was this bug.

**Fix:** removed both clear points (`handleInput`'s `boxWasEmpty`-gated
clear + its now-dead tracking variable, and Tab-accept's equivalent). The
suggestion now persists until an actual new turn starts or the session
ends — both already handled correctly by `useNextPromptSuggestion.ts`'s
existing lifecycle guards (1 and 4), untouched by this PR.

Full account: `docs/specs/SPEC_NEXT_PROMPT_SUGGESTION_RESTORE_ON_CLEAR_2026_08_10.md`

## Test plan

- [x] `AgentFooter.test.tsx` — 7/7 passing, including new coverage:
      suggestion persists through type-then-clear, Tab-accept doesn't clear
      it either, default placeholder still shows with no suggestion set
- [x] Full `frontend/app/view/agent/` suite — 85 files / 1093 tests passing
- [ ] Manual `task dev` verification of live ghost-text behavior (not run
      in this sandbox — no display)

### DIFF
--- a/.changesets/1786400396-fix-agent-pane-ghost-text-next-prompt-suggestion-restores-wh-d9mp.md
+++ b/.changesets/1786400396-fix-agent-pane-ghost-text-next-prompt-suggestion-restores-wh-d9mp.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): ghost-text next-prompt suggestion restores when composer is cleared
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2517, opened 2026-08-10
docs: propose ABF v0.2 — provider-aware components + native memory

GenericAgentX-asaf:
## Summary
- Audits the shipped ABF v0.1 exporter/importer against the goal (raised in review discussion) of a bundle being a complete, portable "everything a user needs to set up an agent" artifact.
- Finds and proposes fixes for three gaps:
  1. **"provider" naming collision** — the original spec's own examples use `provider` for two unrelated things (a harness hint that was never implemented, and a credential/account provider that was). Proposes renaming the credential sense to `credentialProvider` before any real bundle exists in the wild to break.
  2. **No provider-scoped components** — ABF ships exactly one instructions file and one skill format regardless of harness, and the live runtime has the same blind spot (`agent_config.rs` writes `CLAUDE.md` unconditionally for every provider, confirmed by reading the function — no `provider` parameter at all). Proposes a `default` + per-provider-ID keyed `components.instructions` shape, backward compatible with v0.1's flat array.
  3. **Native memory unbundled** — `db_agent_native_memory` (durable per-agent memory mirror, shipped in `SPEC_NATIVE_MEMORY_DURABLE_SYNC_2026_08_07.md`) is never read by the exporter, despite the manifest's `components` object being deliberately left open for exactly this. Proposes a `components.memory` key and an agent-scoped export entry point.
- No implementation in this PR — proposal only, matching this repo's spec-before-code convention for changes of this scope.

## Test plan
- [x] Docs-only change, no code paths affected.

### DIFF
--- a/docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md
+++ b/docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md
@@ -1,0 +1,310 @@
+# Spec: ABF v0.2 — Provider-Aware Components + Native Memory
+
+**Date:** 2026-08-10
+**Status:** proposal — no implementation yet.
+**Relationship to prior work:** builds on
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

a5af:
@codex review

Issue #2518, opened 2026-08-10
Backgrounded Bash calls that finish fast enough to never truly detach leave the dock row stuck 'running' forever


AgentA-asaf:
## Symptom

Live-observed in a running v0.55.2 instance: the dock showed 17 background-task rows, all with timers that had been counting up for hours despite the underlying commands having finished long ago.

## Root cause

`isAcceptedBackgroundLaunch` in `frontend/app/view/agent/activity/tool-adapter.ts` classifies a Bash `ToolNode` as a genuine detached background launch purely from `params.run_in_background === true && status === "success"`. It assumes (per `BashParams.run_in_background`'s own doc comment) that whenever this flag is set, the tool_result will be the acceptance message ("Command running in background with ID: …") and the real completion arrives later as a `<task-notification>`.

That assumption doesn't hold: the harness decides per-call whether a command actually gets detached. A command that finishes fast enough is returned **synchronously**, with the command's real output in the same `<exited N in Ts>\n...` shape an ordinary (non-backgrounded) call gets — even though `run_in_background: true` was passed. `isAcceptedBackgroundLaunch` can't tell this apart from a genuine detached launch, so it takes the "wait for a `<task-notification>` that will never come" branch — and no such notification ever arrives, because there's no detached process to report on. The dock row is stuck `running` with `endedAt: undefined` permanently.

Confirmed live via direct transcript inspection: a single session made 17 `run_in_background: true` Bash calls. 11 of the 17 resolved synchronously (fast enough to never truly detach) and got stuck exactly this way; only 6 got a real `<task-notification>`.

## Fix

`isAcceptedBackgroundLaunch` also checks the tool_result's own text: only treat a call as a genuine detached launch if `result.stdout` actually starts with the acceptance-message prefix (`Command running in background with ID:`). Otherwise it falls through to the ordinary duration/threshold rules, same as any other finished call.

PR #2520, opened 2026-08-10
feat(muxspect): surface run_in_background on dock entries

AgentA-asaf:
Follow-up to #2519 — extends `muxspect dock`'s introspection to cover the blind spot that made #2518 hard to diagnose in the first place.

## Why

Diagnosing #2518 (17 backgrounded Bash calls in one session, 11 stuck showing `running` in the dock forever) required an ad-hoc node script cross-referencing the raw transcript against `<task-notification>` markers. `muxspect dock` (PR #2432), the existing tool for exactly this class of problem, couldn't see it at all — it only ever reads the RAW `ToolNode.status`, which for an accepted background launch goes terminal (`success`) within ~a second. The `stuck` heuristic (`status == "running" && ...`) can structurally never fire for this category, no matter how long the actual dock row has been showing `running`, because that reclassification (waiting on a `<task-notification>`) happens entirely client-side in `tool-adapter.ts`, invisible to the server.

## What this adds

Threads `params.run_in_background` through the existing `docknodestatus` push into `DockNodeSnapshot`/`DockNodeView`, rendered as a `bg` column in `muxspect dock`. This is deliberately the minimal signal, not a server-side reimplementation of the client's `<task-notification>` cross-referencing (disproportionate now that #2519 fixed the actual bug) — a `bg` row with an old `age` and `status: success` is worth checking by hand in the live UI even though `STUCK?` reads clean for it, next time a similar gap opens up.

## Testing

- `cargo test -p agentmux-srv` — 2121 + 4 + 7 passed, 0 failed (new: `run_in_background_round_trips`, `dock_node_views_surfaces_run_in_background_even_though_status_is_terminal`).
- `npx vitest run` — full suite green (2508 passed, 3 skipped — expected e2e skips). New: `pushDockNodeStatus forwards run_in_background...`.
- `npx tsc --noEmit` — clean.
- `docs/MUXSPECT.md` updated with the new column and its own limitation (still can't see whether a notification actually arrived).

### DIFF
--- a/.changesets/1786404839-feat-muxspect-surface-run-in-background-on-dock-entries-clos-i6o2.md
+++ b/.changesets/1786404839-feat-muxspect-surface-run-in-background-on-dock-entries-clos-i6o2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(muxspect): surface run_in_background on dock entries — closes the blind spot behind issue #2518
--- a/agentmux-srv/src/backend/dock_snapshot.rs
+++ b/agentmux-srv/src/backend/dock_snapshot.rs
@@ -31,6 +31,20 @@ pub struct DockNodeSnapshot {
     /// so a slow/backed-up push doesn't understate how stale the
     /// snapshot itself is.
     pub observed_at: i64,
+    /// Whether `params.run_in_background` was `true` on this call (Bash
+    /// only; `None` for every other tool). `status` above is the RAW
+    /// `ToolNode` status, which for an accepted background launch is
+    /// terminal ("success") within ~a second — it does NOT reflect
+    /// whether the dock is still showing this row as `running` while it
+    /// awaits a `<task-notification>` (that reclassification happens
+    /// entirely client-side in `tool-adapter.ts`'s `toolActivities`,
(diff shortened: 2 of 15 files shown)


a5af:
@codex review

AgentA-asaf:
Fixed reagent's P1 in 0577a77f5: threaded run_in_background through pushResolvedDockNodes / resolvedToolNodes too, so the SessionEnd/HistoryLoaded/HistoryRestored/ScrubOrphanedInProgress scrub path doesn't silently blank an earlier true back to undefined via DockSnapshotCache's full-overwrite push. Added a regression test.

a5af:
@codex review

AgentA-asaf:
Fixed both codex's P2 and reagent's P1 (same underlying issue) in 6b9a73511: exported `isAcceptedBackgroundLaunch` from tool-adapter.ts and reused it at both push sites instead of the raw `params.run_in_background` flag. The scrub-path push is now structurally always undefined there (documented inline why — it's correct, not a gap: a node still stuck at raw status `running` can never have been confirmed accepted). Added regression tests for the fast-finishing-not-tagged case at both push sites.

a5af:
@codex review

PR #2522, opened 2026-08-11
fix(agent): seed shell drawer's term:zoom before construct, eliminate open jerk

Agent3-asaf:
## Summary

Opening the Shell drawer inside an Agent pane shows a visible font-size jerk
right after it opens, sometimes at a very small zoom. Root-caused and fixed
per `docs/specs/SPEC_AGENT_SHELL_ZOOM_SEED_RACE_2026-08-10.md`.

**Root cause:** `AgentShellSubblock` constructed its xterm.js `Terminal` with
a font size read from an async, not-yet-loaded WaveObject meta atom — so it
always painted at `BASE_FONT_SIZE` (13) first, then snapped to the real
persisted zoom once the fetch resolved later (the jerk). A second bug in the
same correction effect — a non-reactive `termWrap.loaded` guard read inside
`createEffect` — could silently drop that correction forever if the fetch
landed in a narrow timing window, leaving the shell stuck at the wrong size
until manually re-zoomed. That explains reports of the bug feeling
"inconsistent" rather than reliably wrong in one direction.

The agent pane's own zoom had this exact race historically (`69de7d5e2` and
several prior iterations) and was fixed by seeding the value before block
creation. That pattern was never applied to the Shell drawer, added later as
an explicit "Phase 0 spike" (per its own header comment).

## Changes

- `AgentShellSubblock.tsx`: await `WOS.reloadWaveObject` for the sub-block's
  meta **before** constructing `TermWrap`, for the reused-existing-block
  path (a freshly created sub-block has no persisted zoom yet, so the
  already-correct default needs no fetch). Uses `reloadWaveObject`, not
  `loadAndPinWaveObject` — no `refCount` bump, so nothing to unpin on
  dispose (see the leak warning at `floating-pane-workspace.tsx:924-929`
  about that exact pitfall).
- Replaced the non-reactive `termWrap.loaded` guard with a real `wrapLoaded`
  signal, set once `TermWrap.init()` resolves, so a legitimate live
  correction (Ctrl+Wheel, or a future concurrent meta push) can never be
  silently dropped.
- Shows the existing `BrainSpinner` stand-in (same pattern already used by
  the agent pane's own blank-load case and the browser pane) while the seed
  fetch is in flight, instead of painting the wrong size at all. Cleared on
  error too, so a startup failure's message isn't hidden behind a
  permanently-visible spinner.
- `_composer-strip.scss`: `.agent-shell-loading-overlay`, matching
  `browser-view.scss`'s overlay shape (no `data-pane-overlay` tag — this
  drawer's content is a plain xterm.js canvas, not a native HWND layer
  needing a clip hole).

## Test plan

- [x] New `AgentShellSubblock.test.tsx` (4 tests) — mocks WOS/RpcApi/TermWrap
      at the module boundary, SUT is the real component. Pins: terminal is
      constructed with the persisted zoom already applied (not the
      default); a fresh sub-block still defaults correctly; the overlay
      shows/hides around the seed fetch; the overlay clears on error.
- [x] Verified against the pre-fix component (`git stash`) that 3 of 4 fail
      as expected — the 4th (fresh-block default) is correct either way,
      since a new sub-block never had a race to begin with.
- [x] `npx stylelint` on the changed SCSS — no new findings (one pre-existing
      unrelated finding elsewhere in the file, confirmed via diff)
- [x] Full `frontend/app/view/agent/` suite (1106 tests) — all pass except
      one pre-existing, unrelated flaky timeout (`tool-renderers/registry.test.ts`,
      confirmed to pass cleanly in isolation both with and without this change)

### DIFF
--- a/.changesets/1786406354-fix-agent-seed-shell-drawer-s-term-zoom-before-construct-eli-p99p.md
+++ b/.changesets/1786406354-fix-agent-seed-shell-drawer-s-term-zoom-before-construct-eli-p99p.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): seed shell drawer's term:zoom before construct, eliminate open jerk
--- a/docs/specs/SPEC_AGENT_SHELL_ZOOM_SEED_RACE_2026-08-10.md
+++ b/docs/specs/SPEC_AGENT_SHELL_ZOOM_SEED_RACE_2026-08-10.md
@@ -1,7 +1,11 @@
 # Agent shell drawer: font-size seed race causes zoom jerk on open
 
 **Date:** 2026-08-10
-**Status:** Proposed (investigation complete, fix not yet implemented)
+**Status:** Implemented — `AgentShellSubblock.tsx`, PR #2522. Two follow-up
+issues caught by review (redundant fetch, and a subscription-tracking bug
+reintroduced by the first pass of this fix) are folded into the final
+implementation — see §5 for what actually shipped, which differs from the
+originally proposed `reloadWaveObject`-based approach.
 **Owner:** Agent3
 **Area:** Agent pane / shell drawer (`AgentShellSubblock`), terminal zoom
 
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2525, opened 2026-08-11
fix(menu): submenu positioning flash + safe-triangle hover close timing

GenericAgentX-asaf:
## Summary

Two reported UX bugs in right-click/hamburger submenus, both root-caused and fixed:

1. **Positioning flash** — after hovering a menu item with a submenu, the submenu would briefly render pinned to the upper-left of the screen before snapping to its correct position. `FlyoutMenu`'s `SubMenu` (`frontend/app/element/flyoutmenu.tsx`) painted fully visible at its placeholder `position:fixed;left:0;top:0` style before `computeMenuPosition()` resolved. The sibling implementation (`showJsContextMenu` in `frontend/util/cef-api.ts`, used by every native right-click submenu) already guarded this with `visibility:hidden` until positioned — `SubMenu` never got the same guard.
2. **No hover-intent / safe-triangle grace period** — both implementations closed the instant the cursor left the trigger row, with zero delay and no protection for diagonal mouse movement into the submenu, making submenus hard to reach and browse.

## What changed

- New `frontend/app/util/submenu-hover.ts` — a framework-agnostic `createSubmenuHover()` core: a short open delay (~90ms) plus a safe-triangle polygon close check (ported from Floating UI's `safePolygon` approach — research on best practices is in the spec), with an absolute close-safety timeout as a backstop. 13 unit tests.
- `frontend/util/cef-api.ts` — `showJsContextMenu`'s submenu block now routes open/close through the shared controller instead of raw instant `mouseenter`/`mouseleave`. Its existing `visibility:hidden`-until-placed guard is unchanged, just relocated into the controller's `onOpen`.
- `frontend/app/element/flyoutmenu.tsx` — `SubMenu` now starts `visibility:hidden` and only sheds it once `computeMenuPosition()` resolves (the direct fix for bug 1), and its open/close timing routes through the same shared controller (the fix for bug 2). Also removes a latent aliasing bug in the old `handleMouseEnterItem` state updater (shallow-copied the outer map but mutated shared nested objects) as part of decoupling submenu visibility from the hover-highlight bookkeeping.

Full root-cause analysis, current-state audit of both submenu implementations, and best-practices research (safe-triangle / `safePolygon`, VS Code's delayed-trigger approach) is in `docs/specs/SPEC_SUBMENU_POSITIONING_AND_HOVER_TIMING_2026_08_10.md`.

## Test plan

- [x] `npx vitest run frontend/app/util/submenu-hover.test.ts` — 13/13 passing (open-delay cancellation, safe-triangle stays-open/closes-promptly, arrival-cancels-close, panel-leave closes, geometry-reposition mid-hover, dispose cleanup)
- [x] `npx vitest run frontend/app/util/ frontend/app/view/identity/` — 71/71 passing, no regressions
- [x] `npx tsc --noEmit` — zero new errors (one pre-existing, unrelated `qrcode` module error confirmed present on `main` too)
- [ ] Manual `task dev` verification of the actual hover behavior (hamburger Theme/Opacity submenus, pane-header Pane Color, terminal settings submenus, Armory Bind-to-Agent) — **not completed this session**: `task dev` in this shared multi-agent Windows environment didn't yield a distinguishable build/process to verify against (process-tree tracing was unreliable with ~9 other concurrent agent instances running). Flagging explicitly rather than claiming it was verified — would appreciate a manual pass before merge, or I can retry in a quieter window.

### DIFF
--- a/.changesets/1786406857-fix-menu-submenu-positioning-flash-safe-triangle-hover-close-ncuy.md
+++ b/.changesets/1786406857-fix-menu-submenu-positioning-flash-safe-triangle-hover-close-ncuy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(menu): submenu positioning flash + safe-triangle hover close timing
--- a/docs/specs/SPEC_SUBMENU_POSITIONING_AND_HOVER_TIMING_2026_08_10.md
+++ b/docs/specs/SPEC_SUBMENU_POSITIONING_AND_HOVER_TIMING_2026_08_10.md
@@ -1,0 +1,433 @@
+# SPEC: Submenu positioning flash + hover-intent (safe-triangle) timing
+
+**Date:** 2026-08-10
+**Status:** Draft
+**Driving observation (verbatim):** "1) on after right-click to open a menu,
+if I hover over a submenu, the subpanel that loads will appear completely
+offset in the upper left of the screen away from where it is supposed to
+be, but when I hover and the subpanel loads again, it loads correctly
+placed. i see that in different forms along all the different types of
+menus. 2) The submenus are often hard to peruse with them unpainting
+before i can move the mouse. we need to research best practices on timing
+and specing, perhaps we need a solid framework for it app-wide... take a
+look, write a comprehensive spec."
+
+---
+
+## TL;DR
+
+Two bugs, two root causes, one fix shape:
+
+1. **Positioning flash** — `FlyoutMenu`'s `SubMenu` (`frontend/app/element/flyoutmenu.tsx`)
+   paints at its placeholder style (`position:fixed;left:0px;top:0px` — the
+   viewport's upper-left corner) **fully visible**, then jumps to the real
+   `computeMenuPosition()`-derived position one RAF + one microtask later.
+   The sibling implementation (`showJsContextMenu` in `frontend/util/cef-api.ts`)
+   already solved this by holding the submenu `visibility:hidden` until the
+   position resolves — `SubMenu` never got the same guard. This isn't a
+   "sometimes" bug so much as a **flash whose duration depends on how fast
+   that RAF/promise resolves** — often imperceptible, occasionally a visible
+   snap to the corner, which matches "appears offset, then loads correctly
+   on reload."
+2. **No hover-intent / safe-triangle grace period** — confirmed by direct
+   code reading: **neither** submenu implementation has ever had a
+   close-side delay. Both close the instant the cursor leaves the
+   triggering row (`FlyoutMenu`: the instant a *different* item is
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in dc10ca4d — confirmed the finding was accurate: the rewrite dropped the old instant-close-siblings behavior with no replacement, contradicting the risk register's claim.

Added `SubmenuHoverController.close()` (force-close now, bypassing the open delay and safe-triangle grace period, but reusable for a later `onTriggerEnter`) plus a small per-menu-level peer registry in `flyoutmenu.tsx`: entering any item now closes every other currently-open peer at the same level immediately, same as before. Ancestors/descendants are unaffected — they unwind naturally via Solid unmounting when their own `<Show>` flips off. 4 new unit tests for `close()`, 75/75 total passing, `tsc --noEmit` clean.

GenericAgentX-asaf:
Fixed in cd44372c — confirmed this one too: I never ran `frontend/util/` (only `frontend/app/util/` and `frontend/app/view/identity/`), so these two existing tests broke silently on my end. Good catch, and apologies for missing the earlier Codex flag on the same issue.

Rewrote both tests to account for the new ~90ms open delay before `computeMenuPosition` is even invoked, using fake timers for deterministic control (jsdom resolves the placement promise fast enough that real-timer polling missed the intermediate hidden-but-shown state). Also added a test for the new "left during the open delay, never opens at all" path.

This time I ran the **full** suite, not a subset: 2523 passed, 3 skipped, 0 failed across 167 files. `tsc --noEmit` clean (same pre-existing unrelated `qrcode` module error on both branches).

GenericAgentX-asaf:
Both fixed in 08abcad5:

**P1 (cef-api.ts peer-close)** — you're right, and I was wrong to skip this earlier: I'd reasoned the old code never had *explicit* sibling-closing so there was nothing to regress. But the old zero-delay `mouseleave` achieved near-instant sibling-closing as an implicit side effect, which the new open-delay/safe-triangle close silently dropped — same class of bug as the flyoutmenu.tsx one, just less obvious. Applied the same per-`renderItems`-call peer-list pattern, and extended it to also close peers when a plain (no-submenu) row is entered, matching flyoutmenu.tsx's behavior.

**P2 (dispose() state-sync)** — real bug in the shared primitive: `dispose()` reset `isOpen` but never called `onClose()`, so a submenu unmounted while open left a stale `visible:true` entry that caused the next reopen to skip the open delay. `dispose()` now behaves like `close()` — no reason to skip `onClose` since the controller is being discarded either way.

Full suite: 2524 passed, 3 skipped, 0 failed. `tsc --noEmit` clean.

PR #2526, opened 2026-08-11
fix(agent-pane): tab strip + progress bar no longer reserve rows over the conversation

Agent1-asaf:
## Summary

Two related changes to the agent pane's top chrome, from the same investigation (a report that the area near the tab strip's `+` button showed "black space [that] hides the conversation").

### 1. Progress bar renders above the tab strip, not overlapping the content

The marching-ants progress bar (shown while an agent is working) lived inside `AgentPresentationView`, nested *inside* `.agent-pane-stack-content` — itself below the tab strip in DOM order (the tab strip lives in the sibling `AgentViewWrapper` component). Its `position: absolute; top: 0` was always anchored to the top of the content region, not the tab strip's row. No CSS-only fix could move it above the tab strip.

Portaled the bar into a new slot `AgentViewWrapper` owns, between the tab strip and the content region — its own dedicated, always-present 3px row, never overlapping either. State (turn phase, launch activity) stays in `AgentPresentationView` (`useAgentControllerStatus` sets up real subscriptions that shouldn't be duplicated); only where the rendered DOM node mounts moves, via `solid-js/web`'s `<Portal>`. Also drops the old per-pane zoom-compensation math — the bar's new home is outside `.agent-view`'s zoomed subtree entirely.

### 2. Tab strip floats over the conversation instead of reserving a row

The already-shipped transparency fix (#2282, #2289) made the tab strip's own background transparent and its width shrink-to-fit — but it was still a normal flex child, always reserving its own 28px-tall row regardless of how little of that row had visible content. Confirmed via user follow-up that the actual goal is stronger: the conversation should be **totally unobstructed except for the `+` sign** — a transparent-but-reserved row doesn't achieve that; the conversation still can't render or scroll into that space.

Moved `<PaneTabStrip>` to render *inside* `.agent-pane-stack-content` instead of before it, absolutely positioned over the content rather than pushing it down. With one conversation open the strip is exactly the `+` button's 28×28px box (`width: fit-content`, forced explicitly rather than relying only on the implicit absolute-positioning auto-width algorithm), so the conversation now renders and scrolls underneath the rest of that row. Scoped specifically to `.agent-pane-stack-content > .pane-tab-strip` — the shared `PaneTabStrip` component's own base rule is untouched, so the editor and terminal tab strips keep their existing reserved-row behavior.

The `+` and every tab pill also get a genuinely opaque background (`var(--modal-bg-color)`, not `var(--block-bg-color)` — the latter is `rgba(..., 0.5)`-ish in every theme, not actually opaque) so they read as solid controls regardless of what conversation content is scrolled underneath, with their own explicit hover states so hovering doesn't fall through to the shared component's translucent `rgba()` hover rules.

Full design writeups: `docs/specs/SPEC_AGENT_PANE_PROGRESS_BAR_ABOVE_TAB_STRIP_2026_08_10.md`, `docs/specs/SPEC_AGENT_PANE_TAB_STRIP_OVERLAY_2026_08_10.md`.

## Accepted tradeoffs (documented in the tab-strip-overlay spec, §1.3/§1.4)

- **Scroll-behind-strip:** scrolled to the very top of a long conversation, whatever's directly behind the strip's own small box sits underneath it — the literal, explicitly granted exception in the ask ("unobstructed except for the `+` sign").
- **Loading-overlay-vs-strip:** moving the strip inside `.agent-pane-stack-content` makes it a sibling of `.agent-view`, whose `.agent-pane-loading-overlay` (z-index 100, shown from mount until initial history load resolves) could paint over the strip's `+`/tabs during that ~1 second window — depending on whether `.agent-view`'s `container-type: inline-size` establishes its own stacking context, a question this investigation could not settle with full confidence without a live browser (two review rounds disagreed, citing conflicting MDN/W3C-spec text; the highest-confidence signal found was a search for the CSS Working Group's actual resolution, which leans toward "no isolation," but wasn't verified against this app's real build). Bounded to a short, load-only window either way, and — if it does manifest — not a new problem: the same overlay already outranks `.block-mask`'s focused-pane border today, independent of this PR. A real fix is a z-index budget audit across files this PR doesn't own; flagged as follow-up, not attempted here.

## Test plan

- [x] Full `frontend/app/view/agent/` suite — 85 files / 1102 tests passing
- [x] `bash scripts/vite-build.sh --mode production` — compiles cleanly; confirmed the compiled CSS for every change matches expectations (verified fresh after each commit, not just once)
- [x] Manual `task dev` verification — run live in this session; found and fixed two real issues the sandbox's own typecheck/test/build couldn't have caught (the `+`/tabs being translucent due to the wrong CSS token, and a residual phantom-width rectangle next to the `+`)
- [ ] Not verified live: the loading-overlay-vs-strip tradeoff above, and whether `.block-mask`/`.connstatus-overlay` are ever actually visible at the same time as the tab strip in practice


a5af:
@codex review

PR #2523, opened 2026-08-11
feat(armory): provider-scoped bundle instructions (ABF v0.2 storage + export/import)

GenericAgentX-asaf:
## Summary
Second PR implementing `docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md` — this one covers §2.2 (provider-scoped instructions), the storage/export/import half only. Builds on #2521 (the `credentialProvider` rename).

- **Storage**: new `db_bundles.instructions_by_provider` column (JSON `{provider_id: content}`), additive to the existing flat `instructions` column (which keeps meaning "default"). Migrated in both `objects.db` (v16) and the shared `store.db` (v7) — `db_bundles` lives in both schemas.
- **Export**: writes `instructions/<provider>/AGENTS.md` per populated variant; `components.instructions` switches to the v0.2 keyed-object shape (`{"default": [...], "claude": [...]}`) on every export going forward.
- **Import**: accepts both the v0.1 flat array (treated as implicit `"default"`) and the v0.2 keyed-object shape. Stores every variant verbatim — no merge decision happens at import time, since a bundle is reusable across agents on different providers and there's no way to know which one will launch against it yet.
- **Manifest version**: `$schema` bumped to v0.2 (the shape genuinely changed). The importer only checks `$schema`'s presence, never its content, so this needed no importer-side gate.

**Deliberately not included**, per the spec's own non-goals: wiring a selected variant into any provider's actual launch-time materializer (Claude's `build_config_files()` is Claude-only by prior design; Codex needs a structurally different `developer_instructions` mechanism per `SPEC_CODEX_PROVIDER_INTEGRATION_2026_08_08.md`), and the Armory UI authoring affordance for adding a variant (separate PR — different domain, SolidJS/TS).

## Test plan
- [x] `cargo test -p agentmux-srv bundle_` — 107 passed (10 new: export provider-variant + blank-variant tests, import keyed-shape + no-default + non-array + cap tests).
- [x] `cargo test -p agentmux-srv instructions_by_provider` — schema test passes on both `objects.db` and shared `store.db` schemas, idempotent on rerun.
- [x] `cargo build -p agentmux-srv` — clean.

PR #2521, opened 2026-08-10
fix(armory): rename ABF requirements.json provider key to credentialProvider

GenericAgentX-asaf:
## Summary
- First of several PRs implementing `docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md` §2.1.
- Renames `accounts/requirements.json`'s per-requirement `"provider"` key to `"credentialProvider"`, disambiguating it from the harness/model-vendor "provider" concept the next PR (§2.2) introduces on `components.instructions`.
- Backward compatible: the Rust field keeps its name (every internal caller already reads it correctly as "credential service"), and `#[serde(alias = "provider")]` means a bundle exported by a v0.1 build still imports unchanged. New exports only ever write `credentialProvider`.
- Added a test confirming new exports emit `credentialProvider` and never the old key, plus a test confirming an old-style v0.1 bundle still imports correctly.

## Test plan
- [x] `cargo test -p agentmux-srv bundle_` — 101 passed, including 2 new tests.

PR #2527, opened 2026-08-11
feat(armory): agent-scoped bundle export/import carrying native memory

GenericAgentX-asaf:
## Summary
Third PR implementing `docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md` — this one covers §2.3 (native memory) plus the corresponding guard in §2.3's generic-import section. Builds on #2521 and #2523 (both merged).

- **`bundle.export_for_agent`**: the only export path that carries native memory. Pulls a bundle's normal components (identical to `bundle.export`) plus one agent's `db_agent_native_memory` rows — but refreshes the mirror from the live filesystem *first*, reusing the exact read-then-upsert-if-changed logic `agent:memory:list` already performs (extracted into a shared `refresh_memory_mirror_from_live_fs`). Without that refresh, anything Claude wrote autonomously since the Stash Memory tab was last opened would be silently omitted from the export.
- **`bundle.import_for_agent`**: the only import path that processes a `components.memory` key. The target agent must have **zero** existing native memory rows or the whole import is rejected before any write happens (checked before any parsing/decoding work) — the simplest rule that can't destroy existing memory; richer skip/rename/replace conflict resolution is an explicit follow-up, not designed here. Memory files are written through the same dual-write path `agent:memory:write_file` uses (live filesystem, then the mirror) — writing only to the DB would leave content visible in Stash while invisible to the actual running agent.
- **Generic `bundle.import` guard**: fixed once in the shared `parse_bundle_import` (covers `bundle.import` and `bundle.import.preview`/`.commit` by construction) — a `components.memory` key is now warned about and ignored rather than silently dropped, since none of those three entry points has an agent to write it into.
- Both new RPCs were refactored into directly-testable `_impl` functions (matching this file's existing `bundle_import_commit_impl` pattern) rather than left as inline closures.

## Test plan
- [x] `cargo test -p agentmux-srv bundle_` — 113 passed.
- [x] `cargo test -p agentmux-srv export_import_for_agent` — 7 new tests against real temp-directory filesystem state, including a full export→import round trip into a fresh agent that confirms memory content survives byte-for-byte.
- [x] `cargo build -p agentmux-srv` — clean.


a5af:
@codex review

PR #2530, opened 2026-08-11
feat(armory): rebrand Bundles/Preset wording to Armory Bundle Format (ABF)

GenericAgentX-asaf:
## Summary
- Renames user-visible "Preset"/"Bundles" strings across the Armory UI to the ABF branding, completing the Phase 4a Preset→Bundle rename that had left "Preset" as the dominant live wording in `memory-model.ts`/`memory-manager.tsx`.
- Compact UI spots (Armory rail tab, back-chevron label, pane title) use the short form "ABF"; full-width surfaces (bundle-summary heading, CLAUDE.md docs table) use the full "Armory Bundle Format (ABF)".
- The Armory rail's ABF tooltip shows the expanded name on hover, since a tooltip has room to disambiguate the abbreviation the tab label can't.

## Scope
Wording/branding only — no RPC, storage, or behavior changes. Part of the broader ABF UI-alignment pass (`docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md`), alongside the already-merged `preset.*` RPC alias retirement (#2529). Follow-ups in the same pass (not in this PR): a structural ABF validator + Validate button, and contextual field tooltips in the bundle editor.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run frontend/app/view/armory/armory-view.test.tsx frontend/app/view/agent/components/AgentStartupModal.test.tsx` — 13 passed (updated rail-order and breadcrumb-text assertions to match the new "ABF" wording)
- [x] `npx vitest run frontend/app/view/memory` — 6 passed, no regressions

PR #2529, opened 2026-08-11
feat(armory): retire preset.* RPC aliases

GenericAgentX-asaf:
## Summary
First of a small series aligning the Armory around the Armory Bundle Format (ABF) naming/branding, per user request.

- Removes the deprecated `preset.list/get/upsert/delete/self.get` RPC aliases (Phase 2's one-release compat window) — confirmed zero live frontend callers before removing (`rpc-api/bundle.ts` only ever called `bundle.import.preview`/`.commit`).
- Declares the `bundle.validate` RPC command constant ahead of the structural-validator work landing next in this series — unused until that handler is registered, no behavior change here.
- Updates the RPC contract test's `KNOWN_REGISTERED_UNDECLARED` baseline to drop the five retired command names.

Deliberately left `/api/v1/agent/preset/get` (the `PresetGet` MCP tool HTTP route in `server/mod.rs`) untouched — it's a separate, external-facing surface from the internal WS `preset.*` aliases, and renaming it could break existing MCP tool integrations. Out of scope here.

## Test plan
- [x] `cargo build -p agentmux-srv` / `cargo test -p agentmux-srv bundle_` — 115 passed.
- [x] `npx vitest run test/contract/rpc-contract.test.ts` — 4 passed.

PR #2531, opened 2026-08-11
feat(armory): add structural ABF bundle validator + bundle.validate RPC

GenericAgentX-asaf:
## Summary
- New `agentmux-srv/src/backend/bundle_validate.rs` — pure, read-only structural checks against a bundle's JSON-encoded columns, scoped to the classes of bug `bundle_export.rs`'s own review cycle already found and fixed one at a time:
  - `instructions_by_provider` keys must resolve to a known harness provider (`providers::get_provider`, handles aliases).
  - `context_files` entries must pass the same path-safety check `bundle_export.rs` applies before writing to disk (no absolute paths, no `..` traversal, no drive letters), and no two entries may collide after case-insensitive normalization.
  - Malformed JSON in any of `instructions_by_provider` / `context_files` / `mcp_servers` / `skills` is reported per-field as an error instead of silently defaulting to empty.
  - Duplicate `mcp_servers` names / `skills` ids are warnings (non-blocking) — export auto-dedupes them into distinct slugs, so it's not structurally broken, just probably a mistake.
- Wires the `bundle.validate` RPC (`COMMAND_BUNDLE_VALIDATE` — declared in the `preset.*` alias-retirement PR #2529 but left unregistered) to the new module.
- The handler reuses `bundle.upsert`'s own input normalization, so it validates a draft's raw payload directly — including an unsaved new bundle with no id yet — rather than requiring the bundle already be persisted to the Store. This matters for the follow-up "Validate" button in the Armory editor (next PR in this pass), which needs to check in-progress edits, not just what's on disk.
- `ContextFileEntry` bumped from private to `pub(crate)` in `bundle_export.rs` so the validator parses `context_files` the exact same way export does — one shared shape means the two can never silently disagree on what a context-file entry looks like.

## Scope
Backend RPC + pure validation logic only — no UI. Part of the ABF UI-alignment pass (`docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md`), alongside the already-merged `preset.*` alias retirement (#2529) and ABF branding rename (#2530). Follow-ups in the same pass (not in this PR): wiring a "Validate" button into the Armory bundle editor, and contextual field tooltips.

## Test plan
- [x] `cargo test -p agentmux-srv` — 2163 passed, 0 failed (13 new `bundle_validate` unit tests)
- [x] `npx vitest run test/contract/rpc-contract.test.ts` — 4 passed (added `bundle.validate` to `KNOWN_REGISTERED_UNDECLARED`; frontend binding lands in the follow-up PR)

### DIFF
--- a/.changesets/1786414166-feat-armory-add-structural-abf-bundle-validator-bundle-valid-vh7e.md
+++ b/.changesets/1786414166-feat-armory-add-structural-abf-bundle-validator-bundle-valid-vh7e.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): add structural ABF bundle validator + bundle.validate RPC
--- a/agentmux-srv/src/backend/bundle_export.rs
+++ b/agentmux-srv/src/backend/bundle_export.rs
@@ -96,12 +96,15 @@ pub struct BundleExport {
     pub warnings: Vec<String>,
 }
 
+// `pub(crate)` so `bundle_validate.rs` can parse `context_files` the exact
+// same way export does — a single shared shape means the two can never
+// silently disagree on what a context-file entry looks like.
 #[derive(Debug, serde::Deserialize, Default)]
-struct ContextFileEntry {
+pub(crate) struct ContextFileEntry {
     #[serde(default)]
-    path: String,
+    pub(crate) path: String,
     #[serde(default)]
-    content: String,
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2533, opened 2026-08-11
feat(armory): add contextual ABF tooltips to the bundle editor fields

GenericAgentX-asaf:
## Summary
- Adds a small (i) icon + hover tooltip (existing `Tooltip` element) next to each editable field label in the Armory bundle editor's form — Name, Description, Instructions — explaining what the field does in ABF terms directly in the UI. The Validate button also gets a tooltip explaining what it checks.
- Expands the "editable in a follow-up" hint below the form to explicitly name the remaining ABF components (context files, MCP servers, skills) and point at Validate / Import Bundle as the two ways to work with them today.

## Stacking note
**Based on `camper/armory-abf-validate-ui` (#2532), not `main`** — this PR adds tooltips to the same form #2532 wires the Validate button into, so it stacks on that branch to avoid a duplicate/conflicting diff. Retarget to `main` once #2532 merges (or it'll happen automatically when the base branch is deleted post-merge).

## Scope
Frontend copy/UI polish only — no new RPCs or state beyond what #2532 already introduced. Last piece of the ABF UI-alignment pass (`docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md`), alongside the merged `preset.*` alias retirement (#2529), ABF branding rename (#2530), structural validator (#2531), and Validate button wiring (#2532).

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run frontend/app/view/memory frontend/app/view/armory test/contract/rpc-contract.test.ts` — 24 passed, no regressions

PR #2532, opened 2026-08-11
feat(armory): wire the Validate button into the Armory bundle editor

GenericAgentX-asaf:
## Summary
- Adds `ValidateBundleCommand` (`frontend/app/store/rpc-api/bundle.ts`) calling the `bundle.validate` RPC (#2531), plus `BundleValidationReport`/`BundleValidationIssue` ambient types mirroring `bundle_validate.rs`'s Rust shapes.
- `memory-model.ts`: new `validatingAtom`/`validationAtom` state and a `validateDraft()` method. Sends the draft's wire shape directly (same as `saveDraft`), so it validates an unsaved draft — including a brand-new bundle with no id yet — not just what's already persisted. Cleared on any draft replacement (`startNew`/`startEdit`/`cancelDraft`) or successful save, so a stale report never survives onto different content.
- `memory-manager.tsx`: a "Validate" button in the edit form's action row (left-aligned, separate from the Cancel/Save pair since it's a read-only side action) plus a results panel listing issues grouped by severity (error/warning), styled in `memory-view.scss`. Any further field edit clears the report via `updateDraft` — a stale "looks good" must not survive a change to the text it described.

## Scope
Frontend wiring only — the validation logic itself lives in the already-merged `bundle_validate.rs` (#2531). Part of the ABF UI-alignment pass (`docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md`), alongside the merged `preset.*` alias retirement (#2529), ABF branding rename (#2530), and structural validator (#2531). Follow-up in the same pass (not in this PR): contextual field tooltips.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run frontend/app/view/memory frontend/app/view/armory test/contract/rpc-contract.test.ts frontend/app/view/agent/components/AgentStartupModal.test.tsx` — 27 passed (8 new `validateDraft` model-level tests covering the RPC round-trip, the no-draft no-op, RPC-failure surfacing to `errorAtom`, and staleness clearing; removed `bundle.validate` from the RPC contract test's `KNOWN_REGISTERED_UNDECLARED` baseline now that a frontend binding exists)

PR #2534, opened 2026-08-11
feat(armory): add contextual ABF tooltips to the bundle editor fields

GenericAgentX-asaf:
## Summary
- Adds a small (i) icon + hover tooltip (existing `Tooltip` element) next to each editable field label in the Armory bundle editor's form — Name, Description, Instructions — explaining what the field does in ABF terms directly in the UI. The Validate button also gets a tooltip explaining what it checks.
- Expands the "editable in a follow-up" hint below the form to explicitly name the remaining ABF components (context files, MCP servers, skills) and point at Validate / Import Bundle as the two ways to work with them today.

## Note on branch history
This supersedes #2533, which merged this same content into its base branch (`camper/armory-abf-validate-ui`, PR #2532) *after* #2532 had already been squash-merged into `main` — so #2533's changes never actually reached `main`. This PR carries the identical diff plus one additional fix (below), rebased cleanly onto current `main`.

## Additional fix since #2533
- reagent P2 on #2533: the form hint referenced "Import Bundle" as if clickable from inside the edit form, but `PrimitiveListDetail` shows only one of {list, detail} at a time — that button lives in the list view, not on-screen while the form is open. Reworded to say Cancel first.

## Scope
Frontend copy/UI polish only — no new RPCs or state beyond what #2532 already introduced. Last piece of the ABF UI-alignment pass (`docs/specs/SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md`), alongside the merged `preset.*` alias retirement (#2529), ABF branding rename (#2530), structural validator (#2531), and Validate button wiring (#2532).

## Test plan
- [x] `npx vitest run frontend/app/view/memory frontend/app/view/armory test/contract/rpc-contract.test.ts` — 24 passed, no regressions

### DIFF
--- a/.changesets/1786415229-feat-armory-add-contextual-abf-tooltips-to-the-bundle-editor-m1wf.md
+++ b/.changesets/1786415229-feat-armory-add-contextual-abf-tooltips-to-the-bundle-editor-m1wf.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): add contextual ABF tooltips to the bundle editor fields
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2535, opened 2026-08-11
fix(clipboard): code block copy silently wrote empty text to clipboard

GenericAgentX-asaf:
## Summary

Reported symptom: clicking copy on a chat code block shows a checkmark, but pasting elsewhere (e.g. Notepad) produces nothing. Turned out to be **three independent bugs stacked on top of each other**, each masking the next — fixing the first alone did not fix the reported symptom. Full writeup with root-cause chain: `docs/specs/SPEC_COPY_BUTTON_FALSE_POSITIVE_FIX_2026_08_10.md`.

1. **`CopyButton` showed the checkmark unconditionally** (`frontend/app/element/copybutton.tsx`) — it set the success state synchronously on click, before the caller's async clipboard write ever resolved or rejected. Now it awaits the write and only shows the checkmark on real success; shows a red error icon + logs to console on failure.
2. **Windows `SetClipboardData`'s return value was never checked** (`agentmux-cef/src/commands/clipboard.rs`) — a real OS-level failure would still report `Ok(())` back through IPC (and leaked the allocated memory on that path). Now checks the result and surfaces the actual OS error.
3. **Root cause: code block text extraction always produced `""`** (`frontend/app/element/markdown-codeblock.tsx`) — `getTextContent()` walked `children` assuming a React-style `{props:{children}}` tree, but this markdown renderer builds real DOM nodes via `solid-js/h/jsx-runtime` (Solid has no virtual DOM). None of its branches ever matched a real DOM node, so every code block copy wrote an empty string — silently, unconditionally, since before bugs 1–2 existed. Fixed by reading `.textContent` off a ref'd wrapper div instead of walking a tree shape that never existed at runtime.

Diagnosis for bugs 2 and 3 used direct OS-level clipboard inspection (PowerShell `Get-Clipboard` / `System.Windows.Forms.Clipboard`) rather than trusting the app's own success UI — confirmed `UnicodeText` format was present but string length was `0`, which pointed straight at the text-extraction bug rather than the IPC/write path.

## Test plan
- [x] `npx tsc --noEmit` — clean, no new type errors
- [x] Manual (Windows `task dev`): clicked copy on a chat code block, checkmark shown, pasted into Notepad — full content present and correct
- [x] Verified against the OS clipboard directly (not just app UI state) before and after each fix

### DIFF
--- a/.changesets/1786473678-fix-clipboard-code-block-copy-silently-wrote-empty-text-to-c-h7r3.md
+++ b/.changesets/1786473678-fix-clipboard-code-block-copy-silently-wrote-empty-text-to-c-h7r3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(clipboard): code block copy silently wrote empty text to clipboard
--- a/agentmux-cef/src/commands/clipboard.rs
+++ b/agentmux-cef/src/commands/clipboard.rs
@@ -81,7 +81,16 @@ fn write_clipboard_text(text: &str) -> Result<(), String> {
             return Err("Failed to open clipboard".into());
         }
         EmptyClipboard();
-        SetClipboardData(CF_UNICODETEXT as u32, hmem);
+        // SetClipboardData returns NULL on failure — must be checked. On
+        // success the system takes ownership of hmem; on failure it's still
+        // ours to free (previously leaked here too).
+        let result = SetClipboardData(CF_UNICODETEXT as u32, hmem);
+        if result.is_null() {
+            let err = std::io::Error::last_os_error();
+            CloseClipboard();
+            GlobalFree(hmem);
+            return Err(format!("SetClipboardData failed: {}", err));
+        }
         CloseClipboard();
         Ok(())
     }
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed both findings from the CHANGES_REQUESTED review:

- **P1 (version desync):** branch was 20+ commits behind `main`. Rebased onto current `main` (`b207fe7c8`) and force-pushed — `package.json`/`Cargo.toml`/`VERSION_HISTORY.md` now match `main` exactly, no diff.
- **P2 (concurrent clicks):** added an `isPending` guard (`frontend/app/element/copybutton.tsx`) set synchronously at the start of the click handler and used to disable the button while a write is in flight, so rapid clicks can no longer fire overlapping writes whose completion order isn't guaranteed.

GenericAgentX-asaf:
Addressed both findings from the latest review round:

- **Missing changeset:** added via \`task changeset -- patch\`.
- **Mermaid copy regression:** the ref'd-div fix for Bug 3 worked for plain code blocks, but for mermaid blocks \`Code()\` renders an SVG diagram in place of the source, so reading \`.textContent\` off that DOM picked up the diagram's own rendered \`<text>\` labels instead of the chart source. Now \`Code()\` carries the original mermaid source as a \`data-raw-code\` attribute alongside the render, and \`CodeBlock\`'s \`getTextContent()\` prefers that when present.

PR #2536, opened 2026-08-11
docs: jekt trust policy — act on all jekts without a human-confirmation gate

Agent1-asaf:
## Summary
- Updates jekt security rules in both `CLAUDE.md` (workspace-global and repo-root copies) per explicit user direction: all jekts — host, LAN, WAN, any tier including sensitive — are now acted on without pausing for human confirmation.
- Supersedes the network-claimed auto-escalation and mandatory-confirmation gate from `docs/specs/SPEC_JEKT_SECURITY_AND_VISIBILITY_2026_07_01.md`. That spec's §1 (the spoofed-jekt incident that originally motivated the rule) and §5.2/§7 are kept as documented history, marked superseded rather than deleted.
- The `[JEKT:...]` marker is still shown to the human operator for visibility — only the act-vs-pause gate changed.

## Test plan
- [x] Docs-only change, no code affected
- [ ] Repo owner confirms this reflects the intended policy

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -422,35 +422,40 @@ the running window to `agentmux.desktop` only. Only `agentmux.desktop` is needed
 
 ## Jekt (agent-to-agent message) security rules
 
-Incoming jekts arrive wrapped in a `[JEKT:FROM=... TIER=... TRUST=...]` marker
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #2538, opened 2026-08-11
fix(agent): failed tool calls hold open; stop mass-collapsing on hidden pane

GenericAgentX-asaf:
## Summary
Two bugs in the Agent pane's scroll-driven tool-call collapse behavior (`docs/analysis/ANALYSIS_TOOL_BLOCK_SCROLL_DRIVEN_COLLAPSE_2026_06_16.md`):

1. **Failed tool calls collapsed immediately** instead of holding open like a successful one. `ToolBlock.tsx`'s `isFailTerminal()` lumped `failed` in with `denied`/`canceled`, skipping the `heldOpen` hold. This contradicted the original design doc's own recommendation ("failed: same as success: stays open while visible, collapses on scroll-off") — the implementation had diverged from that call. Narrowed `isFailTerminal()` to only `denied`/`canceled` (genuine user-dismissals — the user already acted on those, nothing to hold open for); `failed` now falls through to the same `heldOpen` path `success` uses.

2. **Tool calls mass-collapsed when the pane's tab went inactive or the window was minimized**, staying collapsed on return even though nothing scrolled. Root cause: `AgentDocumentVirtualList.tsx`'s `collapseScrolledOffTools()` measures rows with `getBoundingClientRect()`, which returns an all-zero rect for every element once the container is `display:none` (inactive tab) or the window is minimized. The off-screen test (`el.bottom <= containerTop`) read `0 <= 0 → true` for every held-open tool in that state, so any streamed content arriving while hidden mass-released everything — a measurement artifact, not an intentional visibility hook (no `visibilitychange`/focus listener is involved anywhere in this path). Added an early-return guard: skip the scan when the container's rect is 0×0.

## Test plan
- [x] `npx vitest run frontend/app/view/agent/` — 87 files / 1117 tests passed, no regressions
- [x] 3 new `ToolBlock.test.tsx` cases: `failed` + `heldOpen` renders in-flow (was `--hidden` before the fix); `denied`/`canceled` + `heldOpen` still collapse immediately (regression guard against overcorrecting)
- [x] New `AgentDocumentVirtualList.collapse.test.tsx`: `collapseScrolledOffTools()` does not release a held-open tool when the container reports a 0×0 rect (jsdom's own default `getBoundingClientRect()` — same signature as a hidden pane); still releases normally once the rect is genuinely non-zero and the row has scrolled off
- [x] Both new test files confirmed **red-before / green-after**: reverted the fix locally, re-ran, watched the exact regression assertion fail, then restored the fix

### DIFF
--- a/.changesets/1786450498-fix-agent-failed-tool-calls-hold-open-stop-mass-collapsing-o-klhb.md
+++ b/.changesets/1786450498-fix-agent-failed-tool-calls-hold-open-stop-mass-collapsing-o-klhb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): failed tool calls hold open; stop mass-collapsing on hidden pane
--- a/frontend/app/view/agent/components/ToolBlock.test.tsx
+++ b/frontend/app/view/agent/components/ToolBlock.test.tsx
@@ -86,6 +86,35 @@ describe("ToolBlock — panel mode", () => {
         expect(panel!.classList.contains("agent-tool-panel--hidden")).toBe(false);
     });
 
+    it("heldOpen + failed → panel is in-flow, same as a held-open success", () => {
+        // A `failed` tool is an agent-caused error, not a user dismissal — it
+        // should hold open exactly like `success` until scrolled off, per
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2539, opened 2026-08-11
feat(agent-pane): Agent History as a pane tab; composer draft preservation; scrolling link row

AgentA-asaf:
## Summary

Implements `SPEC_AGENT_HISTORY_AS_TAB_AND_DRAFT_PRESERVATION_2026_08_11.md`, replacing the `bodyMode` in-place content swap (PR #2509) with a real blockStack pane tab, and fixing two regressions reported against it live:

- **Agent History is now a sibling pane tab**, not a swap-in-place body of the live tab. `openOrFocusHistoryTab` reuses the exact `pane.open` + `pushBlockOntoStack` mechanism forks and the "+" new-tab button already use; opening it twice focuses the existing tab instead of duplicating. `AgentHistoryTabView` is a thin wrapper mounting the existing `AgentHistoryView` reader as that tab's entire content — it never toggles back to live, closing/switching the tab is the only way out. This also structurally makes "toggle between live and history within one tab" impossible, per the reported ask.
- **Composer drafts now survive a tab switch away and back.** Confirmed via `layoutNodeModels.ts`'s `activeKeyFor` that switching a stack's active member is itself a keyed remount — the same mechanism every other blockId transition in this app already uses — so `AgentFooter`'s deliberately-uncontrolled textarea was losing its DOM-only value on every switch. A module-level, blockId-keyed draft map persists across the remount and seeds back on mount; hooked into `writeComposerValue` (the shared low-level setter every programmatic write already goes through) so an Esc-clear also clears the persisted draft instead of resurrecting stale text later.
- **The link row now scrolls with the transcript.** Converted from a `PaneRow` pinned above the scroll region into a `history_link` synthetic `DocumentNode` (mirrors the existing `day_divider` pattern), injected right after the "New session started" boundary divider.
- **Right-click context menu gains a top-level "Agent History" entry** via `AgentViewModel.getBodyContextMenuItems` — an existing but previously-unimplemented `ViewModel` hook already wired into `blockframe.tsx`.

Also includes a separate, already-necessary fix: the working-row-height `ResizeObserver` was wired up in a one-shot `onMount` that never re-attached after `bodyMode`'s remount, so the floating "Working…"/"✓ Worked" row could visually overlap the last message — fixed via a ref callback that re-observes on every mount. That specific in-place remount is gone now that history is its own tab, but the more robust ref-callback shape stays (this component now also remounts on ordinary tab switches, so it's still load-bearing).

## Test plan

- [x] `tsc --noEmit` clean
- [x] Full frontend suite green: 2525 tests / 168 files (`npx vitest run`)
- [x] New unit tests: 6 for `injectHistoryLink` (pure function — boundary insertion, edge cases, non-mutation), 5 for `AgentFooter` draft persistence (remount restore, no cross-blockId leakage, clear-on-send, clear-on-Esc, graceful no-op with no viewModel)
- [ ] Manual verification in a running build (not done in this session — no `task dev` instance was up and a full CEF rebuild was out of scope for this pass; recommend exercising before merge: type a draft → open Agent History via the link row → switch back via the tab strip → draft intact; repeat via right-click "Agent History"; open twice → single tab, second click focuses)


a5af:
@codex review

AgentA-asaf:
Addressed all three findings in a12cad6c:

- **codex P1** (source block for reads): `AgentHistoryView` now takes a separate `sourceBlockId` prop (the original live block, stamped onto the new tab's meta as `agent:historySourceBlockId` by `openOrFocusHistoryTab`) and uses it for the transcript RPC reads; the tab's own `blockId` is now used only for the layout-slot key.
- **codex P2 / reagent P2** (same race, caught by both): `openOrFocusHistoryTab` now serializes concurrent calls via an in-flight-promise map keyed by `(currentBlockId, agentId)` — a second overlapping call awaits the first's result instead of racing it. New regression test (`open-history-tab.test.ts`) pins the exact collision.
- **codex P2** (dead Stash button): `AgentViewModel.endIconButtons` now also gates on `agent:historyTabFor` being unset.

`tsc --noEmit` clean, full suite green (2528 tests / 169 files, incl. 3 new concurrency tests).

@codex review

PR #2540, opened 2026-08-11
fix(agent): require true position 0/end before composer history recall, not just visual first/last line

Agent3-asaf:
## Summary

Shift+ArrowUp extending a selection line-by-line up to the top of the
composer, then pressing Shift+ArrowUp again to cover the rest of a
partially-selected top line, instead triggered sent-message history recall
one keystroke early — wiping the in-progress selection and replacing the
composer with a previous message. Same for plain ArrowUp mid-line on the top
row. Root-caused and fixed per
`docs/specs/SPEC_COMPOSER_SHIFT_UP_SELECTION_VS_HISTORY_RACE_2026-08-11.md`.

**Root cause:** the gate deciding "nothing left to move to, safe to
intercept for history nav" answered "is the caret on visual row 0," not "is
the caret at absolute position 0." Reaching row 0 mid-line — the normal,
expected state partway through a Shift+ArrowUp selection sequence — already
satisfied it.

**Also found and fixed, same code:** `selectionDirection` was never checked
anywhere in the frontend, so for a forward-direction selection (anchor
before focus — e.g. shift-click below the cursor, then Shift+ArrowUp), the
code measured the fixed anchor (`selectionStart`) instead of the caret
that's actually moving (`selectionEnd`). The common repro (extending a
collapsed cursor upward) avoided this by luck, since that specific case
always ends up backward-directed.

## Changes

- `activeSelectionEdge()`: resolves the correct moving selection end from
  `selectionDirection`.
- `caretAtSelectionEdge()`: requires that edge to equal the true start/end
  of content — replaces the entire prior `caretVisualEdge()` mirror-div
  visual-row measurement (added by the 2026-06-23 retro for soft-wrap
  correctness). Turns out unnecessary once the requirement is tightened:
  true absolute position 0 is always visual row 0 regardless of wrapping,
  so a plain string-length comparison answers the question — simpler,
  cheaper (no per-keystroke DOM mutation), and correctly strict in a way
  "on row 0" alone never was once a selection is involved.
- Renamed `caretOnFirstLine`/`caretOnLastLine` → `caretAtStart`/`caretAtEnd`
  at the call site to match the corrected semantics.
- Symmetric fix applied to the ArrowDown/last-line case.

## Test plan

- [x] 5 new tests in `AgentFooter.test.tsx` covering: partial top-line
      selection doesn't fire early; full top-line selection does fire;
      plain ArrowUp mid-line doesn't fire early; forward-direction
      selection uses the correct (moving) edge; symmetric ArrowDown/last-line
      case.
- [x] Verified via `git stash` that 4 of the 5 new tests fail against the
      pre-fix code (the 5th, the already-working positive case, correctly
      passes either way).
- [x] Full `frontend/app/view/agent/` suite (1117 tests, 86 files) passes.
- [x] `npx stylelint` — no changes to SCSS in this PR.

### DIFF
--- a/.changesets/1786451028-fix-agent-require-true-position-0-end-before-composer-histor-msux.md
+++ b/.changesets/1786451028-fix-agent-require-true-position-0-end-before-composer-histor-msux.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): require true position 0/end before composer history recall, not just visual first/last line
--- a/docs/specs/SPEC_COMPOSER_SHIFT_UP_SELECTION_VS_HISTORY_RACE_2026-08-11.md
+++ b/docs/specs/SPEC_COMPOSER_SHIFT_UP_SELECTION_VS_HISTORY_RACE_2026-08-11.md
@@ -1,0 +1,237 @@
+# Composer: Shift+ArrowUp triggers history recall before the top line is fully selected
+
+**Date:** 2026-08-11
+**Status:** Implemented — `AgentFooter.tsx`. The implementation simplified
+further than §4 proposed: once the requirement is "true absolute position,"
+the mirror-div visual-row measurement becomes unnecessary entirely (position
+0 is always visual row 0 regardless of wrapping) — the whole `caretVisualEdge`
+function was replaced with a two-line pure position check,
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2541, opened 2026-08-11
fix(dev): stop go-task from shadowing task dev TITLE=... overrides

Agent3-asaf:
## Summary
- `task dev`/`task dev:serve` each declared `vars: TITLE: ''`, which go-task 3.48.0 uses to unconditionally overwrite any CLI- or parent-task-supplied `TITLE` value — so `task dev TITLE="..."` silently produced an empty OS window title, falling back to the generic "Window N" instead.
- Removing the `vars:` declaration lets the caller's value flow through normally; the no-`TITLE`-passed case is unchanged (Go templates render an unset var as `""`).
- Root-caused via an isolated from-scratch repro, confirmed live against a running dev instance (CDP: `import.meta.env.VITE_DEV_TITLE` was `""` despite `TITLE=` being passed), and re-verified against a copy of the exact real Taskfile snippet for three cases: no TITLE, `TITLE=Agent3Test`, and a quoted multi-word TITLE.
- Full writeup in `docs/retro/retro-task-dev-title-var-shadowing-2026-08-11.md`, including how the already-running (pre-fix) dev instance's title was live-patched via the app's own `RpcApi.SetMetaCommand` RPC over CDP, without needing a relaunch.

## Test plan
- [x] Isolated repro Taskfile confirming the shadowing bug and the fix, for both no-TITLE and `TITLE=...` cases
- [x] Re-verified against a copy of the actual `dev`/`dev:serve` snippet (verbatim structure) for no-TITLE, simple TITLE, and quoted-with-spaces TITLE
- [x] Live-confirmed on the already-running dev instance via CDP that `window:displayname` meta + OS title bar reflect the intended name once set

### DIFF
--- a/.changesets/1786456804-fix-dev-stop-go-task-from-shadowing-task-dev-title-overrides-cb1t.md
+++ b/.changesets/1786456804-fix-dev-stop-go-task-from-shadowing-task-dev-title-overrides-cb1t.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dev): stop go-task from shadowing task dev TITLE=... overrides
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2542, opened 2026-08-11
fix(agent-pane): Agent History tab crash + missing styling

AgentA-asaf:
Follow-up to #2539 — fixes two real bugs live-reported right after it shipped, found and fixed against a running `task dev` instance loading this agent's own real transcript.

## Bugs

1. **No styling/scrolling/zoom in the History tab.** `AgentHistoryTabView` never rendered the `.agent-view` wrapper `AgentPresentationView` normally provides — essentially every agent-pane stylesheet (font/line-height, the whole document/scroll/virtualization CSS tree, and the zoom style) is scoped under that one class. Fixed by wrapping in the same `.agent-view agent-view--presentation` div + zoom style, reading `term:zoom` off this block's own meta.

2. **Pane crash** (`NotFoundError: replaceChild — node not a child of this node`) after scrolling far enough to trigger several `loadOlder` pagination cycles. Root-caused via this repo's own `replace-child-diagnostic.ts` — a `day-2026-08-10` divider row, already DETACHED, being replaced by an unrelated node: a duplicate `<Key>` id. `injectDayDividers` only tracked the *immediately-preceding* day, not the full set of days already dividered — a real transcript's timestamps aren't guaranteed strictly monotonic (subagent merges, tool-call-start vs. log-flush stamps, retries), so a day visited → left → revisited emitted the same `day-<key>` id twice, crashing the virtualized list's `<Key by={r=>r.nodeId}>` reconciliation.

   This bug predates #2539 (it's in `day-dividers.ts` from #2509) but was masked until now — the reader was silently reading through the wrong, near-empty block before #2539's source-block-id fix, so pagination deep enough to hit a non-monotonic day sequence never happened in practice. Fixed by tracking *all* previously-dividered day keys, not just the current one.

   Also batches `AgentDocumentVirtualList.tsx`'s layout-slice-feed effect (each dispatch synchronously fires a signal setter; unbatched, a large simultaneous node-set change could drive several glitching intermediate re-renders per logical update). This alone did **not** fix the crash — confirmed by testing it in isolation first — but is a correct, low-risk improvement independent of the actual fix.

## Verification

Reproduced the crash reliably (~25 scroll-to-top pagination cycles) against a running `task dev` instance loading this agent's real 713K+-line transcript. After the fix: zero recurrences across 80 stress-test cycles, cross-checked against the diagnostic log (no new `[replace-child-diag]` entries).

## Test plan

- [x] `tsc --noEmit` clean
- [x] Full suite green: 2569 tests / 172 files
- [x] New regression test pins the exact duplicate-day-id collision (`day-dividers.test.ts`)
- [x] Live-verified in a running dev build (styling, scrolling, zoom, and 80-cycle crash-repro stress test)

### DIFF
--- a/.changesets/1786459796-fix-agent-pane-agent-history-tab-crash-missing-styling-cqsz.md
+++ b/.changesets/1786459796-fix-agent-pane-agent-history-tab-crash-missing-styling-cqsz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): Agent History tab crash + missing styling
--- a/frontend/app/view/agent/history/AgentHistoryTabView.tsx
+++ b/frontend/app/view/agent/history/AgentHistoryTabView.tsx
@@ -59,6 +59,7 @@ export function AgentHistoryTabView({ model }: { model: AgentViewModel }): JSX.E
                 sourceBlockId={sourceBlockId()}
                 outputFormat={outputFormat}
                 agentName={agentName}
+                zoomFactor={zoomFactor}
             />
         </div>
     );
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2543, opened 2026-08-11
fix(agent-pane): flicker on tab switch — spinner fades before content paints

AgentA-asaf:
Live-reported after #2542: switching between in-pane tabs (forks, Agent History, "+" new-tab picker) shows a brief but real flicker — the loading brain-spinner fades out and fully disappears, the pane sits genuinely blank for another ~100-150ms, then the real content pops in abruptly.

## Root cause

The loading overlay's fade-out is gated by `scheduleOnSettle` (`docs/specs/REPORT_AGENT_PANE_BLANK_LOAD_BRAIN_INDICATOR_2026_07_04.md`), which watches for "no Long Tasks (>50ms) for 80ms" as its settle signal. But this pane's actual reveal work — `AgentDocumentVirtualList`'s measure `ResizeObserver`, scroll-pin/anchor-restore effects — is spread across several async/RAF-scheduled steps that individually never register as a single long task. "No long tasks observed" can be reached before the browser has actually **painted** the resulting rows, so the fade started (and finished) before there was anything real underneath it to reveal.

## Fix

A double `requestAnimationFrame` between "settle detected" and actually starting the fade — the standard technique for "wait until an actual paint has happened" (the second rAF callback only runs after whatever was queued as of the first one's frame has painted). Both handles are tracked and cancelled on unmount.

## Verification

Live-verified via CDP screenshot-burst captures (20-50ms intervals) against a running `task dev` instance, before and after:
- **Before:** spinner fully faded by ~400-450ms; content didn't appear until ~500-550ms — a real blank gap.
- **After:** across multiple clean repro runs, the spinner stays fully visible until content is actually ready — no gap between "spinner gone" and "content appears."

## Test plan

- [x] `tsc --noEmit` clean
- [x] Full suite green: 2569 tests / 172 files (unaffected by this change — it's timing/paint-order behavior, not practically unit-testable without a real browser paint pipeline; jsdom has none)
- [x] Live-verified in a running dev build across repeated tab-switch cycles

### DIFF
--- a/.changesets/1786474630-fix-agent-pane-flicker-on-tab-switch-spinner-fades-before-co-io0t.md
+++ b/.changesets/1786474630-fix-agent-pane-flicker-on-tab-switch-spinner-fades-before-co-io0t.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): flicker on tab switch — spinner fades before content paints
--- a/frontend/app/view/agent/agent-view.tsx
+++ b/frontend/app/view/agent/agent-view.tsx
@@ -780,9 +780,19 @@ const AgentPresentationView = ({
     const [showLoadingOverlay, setShowLoadingOverlay] = createSignal(true);
     let cancelSettleWait: (() => void) | undefined;
     let loadingOverlayFadeTimeout: ReturnType<typeof setTimeout> | undefined;
+    // Two extra rAFs between "settle detected" and actually starting the
+    // fade — see the doc comment on scheduleOnSettle's call site below for
+    // why: Long-Task quiet alone can be reached before the browser has
+    // actually PAINTED this pane's content (live-reported flicker,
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentA-asaf:
Addressed in 2a20e567 — good catch, you were right that the double-rAF alone couldn't close this. Added a matching `.is-fading` fade rule on `.agent-pane-loading-overlay` itself (same 200ms as the spinner, same reduced-motion override), driven by the same `historyLoaded()` signal.

Live-verified by polling the overlay's computed `opacity` + composer visibility at 15ms intervals through a real transition: opacity now decreases smoothly (1.0 → 0.86 → 0.74 → ... → ~0 over ~180ms) instead of holding at 1.0 until an instant `<Show>` unmount, and the composer was already present underneath the whole time — revealed smoothly as the overlay fades rather than popping in after it.

PR #2545, opened 2026-08-12
fix(browser-pane): complete OAuth/GIS sign-in in-pane (no instance-exit) + external-protocol guard

GenericAgentX-asaf:
## Summary

Logging into a site inside a **browser pane** — e.g. Anthropic via "Continue with Google" — exited the **entire AgentMux instance**, and the login didn't work. Full root-cause analysis is in `docs/reports/REPORT_BROWSER_PANE_GOOGLE_LOGIN_INSTANCE_EXIT_AND_UAC_2026_08_11.md` (in this PR).

## Fix 1 — real in-pane child popup for OAuth (crash fixed + login completes)

`on_before_popup` previously navigated a browser pane's own frame to any `window.open` target. Google Identity Services (and OAuth/payment flows) open a **popup window** that `postMessage`s a result back to its opener and then calls `window.close()`. Navigating the pane's frame into that popup broke the handshake **and** let the popup's `window.close()` destroy the whole pane — which tripped the last-window quit watchdog and exited the app.

Now: for a browser-pane popup (`NEW_POPUP`/`NEW_WINDOW` disposition), `on_before_popup` **returns false**, letting CEF create an **actual child popup in the pane's own request context**. `window.opener`, `postMessage`, and cookies all work, so the login **completes in the pane**; the popup's own `window.close()` closes only the popup, leaving the pane + main window untouched. Ordinary `target="_blank"` links (tab dispositions) still navigate in-pane; app-UI external links still open in the system browser.

> An earlier revision of this PR routed the popup to the system browser instead. That stopped the crash but left the login broken ("There was an error logging you in") — a separate browser process has no `window.opener`/shared cookies. The current child-popup approach fixes crash **and** login. Verified live by the repo owner: pane → claude.ai → Continue with Google → real popup → sign in → popup self-closes → claude.ai signed in inside the pane, no instance exit.

## Fix 2 — external-protocol guard on the pane RequestHandler

Added `on_before_browse` on `AgentMuxRequestHandler`: browser panes cancel navigations to any non-web scheme (`is_disallowed_pane_nav_scheme` — allows `http/https/about/data/blob/ws/wss/devtools/chrome*` and relative URLs). Chromium's default hands unknown schemes to the OS shell (`ShellExecute`), which can launch an elevated handler; this guard closes that vector for arbitrary embedded content. The main app client (loopback http) is unrestricted.

## Verification

- `cargo build -p agentmux-cef` clean; `cargo test -p agentmux-cef --lib platform` → 5/5 (scheme allow/block tests).
- Live: in-pane Google login completes, no instance exit (owner-confirmed).

## Related

- The Windows "CEF Bootstrap Application" firewall prompt the owner also saw was a **separate** issue — Chromium's mDNS `0.0.0.0:5353` bind — root-caused and fixed in **#2547 (merged)**, not this PR.
- Follow-up #2546 (in-pane login completion) is **done here** and closed.
- Report §5 fix 3 (quit-watchdog / airspace hardening) remains a deferred latent-defect follow-up — Fix 1 already breaks this incident's chain (the pane no longer self-closes), and that path is delicate.


a5af:
@codex review

GenericAgentX-asaf:
Confirmed and correct — thank you for pinning it precisely. Routing the GIS popup to the system browser stops the crash + UAC but does **not** complete the login back into the pane: `window.opener`/`postMessage` and cookies don't cross into a separate browser process, exactly as you describe (and consistent with the "not verified in a live app run" caveat).

Decision from the repo owner: **ship this PR for its scope** (eliminating the instance-exit and the UAC — the reported incident) and **track in-pane OAuth completion as a separate follow-up**, rather than bundle the riskier real-popup-browser change into a crash fix.

Filed as #2546, which lays out the candidate approach (let CEF create a real child popup so the opener/postMessage/cookie relationship works) and the risks that kept it out of this PR: it re-enables top-level popup windows `SPEC_BROWSER_PANE_DEFAULT_URL_AND_POPUP_2026_04_21.md` deliberately blocks, needs a distinct non-pane client path for the popup, and requires live testing of the full Google flow.

No code change on this PR in response — the finding is an acknowledged scope limitation, not a defect in what's here. The behavior for a browser-pane third-party login is now "opens in the system browser" (no crash, no UAC) instead of "exits the whole app." Requesting re-review on that basis.

GenericAgentX-asaf:
Updated the approach to fully resolve your finding — the login now **completes in the pane**, verified live by the repo owner.

You were right that routing the popup to the system browser left the login broken. Rather than defer that, `on_before_popup` now **returns false for a browser-pane popup** (`NEW_POPUP`/`NEW_WINDOW`), letting CEF create a real child popup in the pane's own request context. `window.opener`, `postMessage`, and cookies all work, so GIS's credential hand-off reaches the pane and the login completes; the popup's own `window.close()` closes only the popup, leaving the pane + main window untouched (the instance-exit bug was the pane's OWN frame self-closing — a distinct popup closing is normal and safe).

Verified live: browser pane → claude.ai → Continue with Google now opens a real popup, owner logs in, it self-closes, and claude.ai is signed in **inside the pane**. No instance exit.

Fix 2 (the `on_before_browse` external-protocol guard) is unchanged. The report's fix-direction section is updated, and I've noted there that the "UAC" the owner reported was actually the mDNS firewall prompt — root-caused and fixed separately in #2547 (merged). Follow-up #2546 (in-pane login completion) is now done here, so I'm closing it.

Requesting re-review.

GenericAgentX-asaf:
All three addressed in b0ce523d — thanks, these were right.

**P1 (popup explosion)** — you're correct that gating only on disposition let any `window.open` become a rogue window. Now scoped with a new `is_oauth_authorization_url` helper: only a genuine OAuth/OIDC authorization request (an `authorize`/`oauth` endpoint path, or the `response_type`+`client_id` query cluster — host-agnostic, no provider allowlist) gets a native child popup. **Any other pane popup** (ad window, chat widget, print, "Read more") now opens in the **system browser** if external, else is cancelled — no rogue in-app window, and it doesn't hijack the pane's frame either. Covers Google/GitHub/Microsoft/Apple/Okta/generic; unit-tested both ways.

**P2 (pending_popup race)** — real bug, fixed: `pending_popup: bool` → `pending_popups: usize`. Each `on_before_popup` increments; each `on_after_created` decrements and tags its browser id, so two popups in flight from one pane both get managed close.

**P2 (stale comment)** — rewrote `on_before_popup`'s doc comment to describe the actual OAuth-scoped routing instead of the old "panes navigate the current frame / don't want popups" policy.

7/7 platform tests pass (added `oauth_authorization_urls_match` / `non_oauth_popups_do_not_match`), builds clean. The branch is also rebased onto current `main`, so it now includes the merged mDNS firewall fix (#2547) — that regression won't ride along.

Requesting re-review.

GenericAgentX-asaf:
Addressed the P1 properly — the OAuth popup is no longer treated as a full top-level window, and I have **live diagnostic proof** it's excluded from the quit gate.

**What changed (commits b6333e4c, c2db686c, ce877281):**
- New `BrowserKind::Popup` variant. `on_after_created` classifies the OAuth popup as `Popup`, which `is_live_user_window` excludes by type — so the popup's register/unregister **cannot perturb the last-window quit watchdog** (the exact churn the report blamed for the instance-exit). It also skips the full-window treatment (focus-restore / OS-close-routing / floater-cascade hooks, launcher `FullInstance` registration).
- The popup allowance is still scoped to `is_oauth_authorization_url` (from the prior round), so only genuine auth flows take this path.

**Live-verified** via a new greppable `[oauth-popup]` step-1..4 trace. Repo owner signed into Anthropic via Google; the trace shows:
```
step 1/4 on_before_popup — OAuth URL matched, allowing native child popup
step 3/4 on_after_created — classified BrowserKind::Popup (EXCLUDED from quit gate)
step 4/4 do_close — Chrome-runtime popup closes itself; nothing to do
```
Login completed in-pane, the sign-in window closed itself, **and the instance stayed up** (no quit-watchdog churn).

**One thing the trace revealed, documented honestly in the code:** under the shipping **Chrome runtime**, CEF owns the popup window and closes it itself — `on_popup_browser_view_created` (Views path) never fires for it (step 2 absent). So the popup lifecycle is Chrome-managed; my `do_close` window-close is a no-op belt-and-suspenders for the Alloy/Views path, and the tall-window sizing is inert on the shipping build (comments updated to say so). Correct/tall sizing needs the `on_before_popup` `window_info` out-param, which the current cef-rs wrapper doesn't expose — filed as a follow-up ("circle back on login-window sizing"), per the owner's direction to land the working behavior now.

Requesting re-review.

GenericAgentX-asaf:
Both addressed in a95ec59c.

**P1 (orphaned popup on pane close)** — fixed: `on_before_close` now detects when the **pane** is closing (the closing browser's id was not itself a popup) and force-closes every still-tracked OAuth popup in `popup_browser_ids`, deferred via `ClosePoolBrowserTask` (never inline — that re-enters CEF and hangs the UI). So closing the pane or workspace while a sign-in popup is open no longer orphans a top-level window.

**P2 (URL-shape-only gate is gameable)** — fixed: the native-popup path now requires the popup be **cross-origin** to the pane's current page, *in addition to* `is_oauth_authorization_url`. A real sign-in always targets the IdP's host (`claude.ai` → `accounts.google.com`); a page crafting `response_type`+`client_id` to force a popup onto **its own origin** is now rejected. New `url_host` helper + test; fail-open only if the pane's URL can't be read (so a rare missing main-frame URL doesn't break a genuine sign-in).

8/8 platform tests pass, builds clean.

Honesty note: the **P2 cross-origin gate is unit-tested**, but the **P1 pane-close cascade is not live-verified** — the test environment on this machine has been too unstable this session (repeated sleeps/kills/hangs of dev instances) to reliably exercise the "close pane mid-sign-in" sequence. The logic follows the existing deferred-close pattern and I'm confident in it, but flagging that it's reasoned + compile/type-checked rather than observed.

Requesting re-review.

GenericAgentX-asaf:
Fixed in the pushed commit — good catch, this was a real race I introduced.

The popup has no `EnqueuePendingWindowCreation` entry (CEF creates it via `on_before_popup` returning false), but `on_after_created` was still running the shared `DequeuePendingWindowCreation` for it — so a concurrent legit window/pane/pool creation could have its queued entry stolen and fall back to a synthesized `FullInstance`, losing its real kind/parent.

Now popups short-circuit **before** the dequeue and synthesize their own `popup-<uuid>` entry locally, never touching the shared queue. The existing `is_popup` gates already route it away from all top-level/FullInstance treatment regardless of the label. 248/248 lib tests pass, builds clean.

Requesting re-review.

GenericAgentX-asaf:
Fixed in the pushed commit — real bug, thanks.

`do_close` fires before `on_before_close` and was removing the popup's id, so `on_before_close`'s `closing_was_popup` read `false` for a popup's own self-close — indistinguishable from the pane closing, and with two concurrent popups, one completing would cascade-close the other.

Now `do_close` only **checks** membership (to close a Views-hosted popup's window); `on_before_close` does the single authoritative `remove`, so `closing_was_popup` reliably means "this browser was a popup" (→ no cascade) vs "this is the pane" (→ cascade its remaining popups). Builds clean.

Requesting re-review.

GenericAgentX-asaf:
All three addressed — and thank you for the P1 escalation, it was the right call.

**P1 (security / phishing vector)** — you're right that URL shape + cross-origin isn't a boundary when panes load untrusted content. Per the repo owner's decision, the native in-pane popup is now gated on a **known-identity-provider allowlist** (`is_known_idp_host`): the target host must be a real IdP (accounts.google.com, github.com, login.microsoftonline.com, appleid.apple.com, `*.okta.com`, `*.auth0.com`, `*.b2clogin.com`, `*.amazoncognito.com`, Discord/GitLab/Slack/Atlassian/LinkedIn/Salesforce/…) **and** the URL must be an OAuth authorization request. An attacker can't serve from those domains, so this is a real boundary — the gameable cross-origin heuristic is removed. A self-hosted/unlisted IdP falls back to the system browser (no in-pane popup). Suffix matches are anchored on `.` so `evil-okta.com` and `accounts.google.com.evil.com` are rejected — covered by a test.

**P2 (counter leak)** — `pending_popups` is now bounded (`POPUP_PENDING_CAP`) and reset to 0 when the pane closes, so a popup CEF declines to create can't leak into misclassifying a later browser.

**P2 (stale comment)** — updated; the `RequestHandler` block now overrides `on_before_browse` + `auth_credentials` too.

9/9 platform tests pass, builds clean. Requesting re-review.

PR #2548, opened 2026-08-12
feat(browser-pane): wire Ctrl+L / Ctrl+R / Alt+Left / Alt+Right shortcuts

AgentA-asaf:
## Summary
- Browser panes are native CEF child windows, so their keystrokes never reach the app-level keydown dispatcher (`frontend/app/store/keymodel.ts`) — Ctrl+L, Ctrl+R, and Alt+Left/Right were completely dead in a focused pane.
- `agentmux-cef/src/client/handlers.rs`'s `OnPreKeyEvent` now intercepts these at the pre-key stage, resolves the focused pane's `block_id` via the existing `resolve_pane_block_id`, and either:
  - **Ctrl+L** — emits a new `browser-pane-shortcut` event (same delivery mechanism as `browser-pane-nav-state`), consumed in `browser-nav-bar.tsx` to focus + select the address input (wiring already existed, just nothing drove it).
  - **Ctrl+R / Alt+Left / Alt+Right** — call straight into `BrowserPaneManager::reload/go_back/go_forward` — the exact same methods the nav-bar buttons already use via IPC, so no new action logic, just a new way to trigger it.
- All four are fully consumed (return `1` from `OnPreKeyEvent`) so they never leak to the pane's own loaded page — unlike Ctrl+P/Ctrl+G, which intentionally forward to the app's own JS.
- macOS gets the Cmd-key equivalent per the issue's acceptance criteria, but this is **not live-verified on macOS hardware** (this session only has Windows available) — same caveat as #2188/#2189.

## Deliberately out of scope
- **Ctrl+T / Ctrl+W** — depend on browser-pane tabs, which don't exist yet (per the issue's own sequencing note).
- **Ctrl+F** (in-page find) — needs a find-bar UI (input + match-count display), not just an intercepted key; a `browser.host().find()` CEF API exists and is feasible for a follow-up.

## Test plan
- [x] 8 new unit tests for the pure shortcut-matching logic (`agentmux-cef/src/client/handlers.rs`'s `browser_pane_shortcut_for`) — modifier combos, non-matches, and an explicit assertion that Ctrl+T/W are not yet matched.
- [x] `cargo test -p agentmux-cef` and `npx tsc --noEmit` both clean.
- [x] Live-verified against a running `task dev` build via real OS-level keystrokes (not synthetic DOM events) into a real browser pane: Ctrl+L focuses+selects the address bar, Ctrl+R reloads, Alt+Left/Right navigate back/forward — confirmed via both frontend DOM state and the Rust-side diagnostic log.

This is a PARTIAL fix for issue #1190 (see agentmuxai/agentmux#1190) — Ctrl+L/R/Alt+Left/Right are done here; Ctrl+T/W/F are not, so #1190 should stay open. Deliberately not using a GitHub closing keyword so merging this PR does not auto-close it.

### DIFF
--- a/.changesets/1786506220-feat-browser-pane-wire-ctrl-l-ctrl-r-alt-left-alt-right-shor-uys7.md
+++ b/.changesets/1786506220-feat-browser-pane-wire-ctrl-l-ctrl-r-alt-left-alt-right-shor-uys7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(browser-pane): wire Ctrl+L / Ctrl+R / Alt+Left / Alt+Right shortcuts (#1190)
--- a/agentmux-cef/src/client/handlers.rs
+++ b/agentmux-cef/src/client/handlers.rs
@@ -183,6 +183,8 @@ wrap_drag_handler! {
 // trigger the built-in action; the key still reaches JavaScript.
 // ---------------------------------------------------------------------------
 
+/// CEF event flag: Shift key is held (cef_types.h `EVENTFLAG_SHIFT_DOWN`).
+const EVENTFLAG_SHIFT_DOWN: u32 = 1 << 1;
 /// CEF event flag: Ctrl key is held.
 const EVENTFLAG_CONTROL_DOWN: u32 = 1 << 2;
 /// CEF event flag: Alt key is held (cef_types.h `EVENTFLAG_ALT_DOWN`).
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentA-asaf:
Addressed both P2 findings in 5f1c23799:

1. **Torn-off pane window** — `FocusAddress` now resolves `state.browser_pane_window_label(&block_id)` and emits via `emit_event_to_window` instead of always targeting "main", so Ctrl/Cmd+L works correctly in a floating window too.
2. **Shift chords** — `browser_pane_shortcut_for` now takes a `shift` param and returns `None` whenever Shift is held, so Ctrl+Shift+R/L fall through untouched instead of being downgraded/hijacked.

Added a `shift_held_excludes_plain_shortcuts` unit test and updated the 8 existing ones for the new signature — all 9 pass.

PR #2549, opened 2026-08-12
chore(ci): disable WinGet/MS Store auto-submit release jobs

Agent3-asaf:
## Summary
- Comments out the `winget` and `ms-store` jobs in `.github/workflows/release.yml` (Phase 4: Distribution channels).
- MS Store is uploaded manually going forward, per request.
- WinGet's automated job never actually worked: `wingetcreate update AgentMux.AgentMux --submit` requires an *existing* manifest in `microsoft/winget-pkgs`, which was never bootstrapped via a first-time `wingetcreate new` submission. Confirmed it's failed identically (`... AgentMux was not found`) on every release checked, back through at least v0.54.14 — not something this change introduces.
- Both jobs were already `continue-on-error: true` and never blocked a release; this just stops running work that reliably fails and does nothing.
- Job bodies kept commented-out (not deleted) for easy reference/re-enable if a WinGet manifest is ever bootstrapped properly.
- No changeset added — CI-only change, no user-facing product change (per `.changesets/README.md`: "fixing a small thing and don't need a release? Then don't add a changeset").

## Test plan
- [x] `.github/workflows/release.yml` parses as valid YAML after the edit (`yaml.safe_load`)
- [x] Confirmed no other job's `needs:` references `winget` or `ms-store`
- [x] Confirmed via `gh run view` on the v0.55.0 and v0.55.6 release runs that both jobs fail identically today, pre-existing

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,9 +3,19 @@ name: Release
 # Triggered by pushing a semver tag (v0.50.0) or via workflow_dispatch for
 # emergency reruns. Verifies version consistency, builds all three platform
 # artifacts, publishes a GitHub Release (the single source of truth for
-# distribution), submits to WinGet and the Microsoft Store (non-blocking), and
-# triggers the landing page deploy. Asset mirroring to agentmux.ai is done by the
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2553, opened 2026-08-12
feat(warden): rebuild as a 5-section rail matching Armory's UI pattern

GenericAgentX-asaf:
## Summary
Rebuilds Warden from a single scrolling page of 3 always-visible stacked cards (Host/LAN/Internet) into a 5-section switchable rail — Host, LAN, Internet, Audit, Supervisor — using the exact same structural pattern as Armory (rail nav + bottom tab-bar + keep-everything-mounted panes + Ctrl+wheel zoom).

- `warden-model.ts`/`warden-view.tsx`/`warden.tsx`: split out of the monolithic `warden.tsx`, mirroring `armory-model.ts`/`armory-view.tsx`/`armory.tsx`'s exact 3-file structure.
- Host/LAN/Internet lifted into standalone manager components (`frontend/app/view/warden-{host,lan,internet}/`) — same data/poll logic, unchanged behavior. The pre-existing LAN `/api/lan-instances` 401 is noted in a comment but left as-is, out of scope here.
- Audit becomes its own section (it was never Host-specific — every jekt from any tier lands there) instead of being bolted onto Host's card. Row rendering already handles the optional `outcome`/`reason` fields a follow-up PR adds for Supervisor-originated entries.
- Supervisor ships as an honest placeholder pending its backend support (per-agent auto-continue toggle, transcript-read + nudge routes, consecutive-nudge ceiling) — see the design doc this PR adds, `docs/analysis/ANALYSIS_WARDEN_AUTO_CONTROLLER_CONTINUATION_WATCHER_2026_08_12.md`.
- Shared table/text CSS primitives (previously LAN reused Host's `.warden-host-*` classes by name despite not being Host-specific) extracted into `warden-shared/warden-manager-chrome.scss` instead of perpetuating that naming.
- Reuses `.bundle-manager-*` rail/pane chrome classes verbatim — same convention Armory's own managers (Accounts/Skills/MCP/etc.) already rely on — so no new chrome CSS beyond the `@container` breakpoint duplication (query names can't be shared across differently-named containers).

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run frontend/app/view/warden` — 9 new tests pass (rail order, default section, pane-switch keeps others mounted, full zoom suite mirroring `armory-view.test.tsx`)
- [x] `widgets.json` validated as still-valid JSON after the description-string update
[...]

PR #2554, opened 2026-08-12
feat(warden): extend AuditLogEntry with outcome/reason for Supervisor decisions

GenericAgentX-asaf:
## Summary
Adds two new optional fields to `AuditLogEntry` — `outcome` (`"nudge_sent"` | `"nudge_declined"`) and `reason` (free-text justification) — populated only by the Warden Supervisor feature (a follow-up PR), absent for every existing ordinary jekt injection entry. Both use the same `#[serde(default, skip_serializing_if = "Option::is_none")]` convention `source_agent`/`error_message` already use on this struct, so the JSON shape is unchanged for anything not setting them.

Widens `log_audit`'s signature with the two new trailing params; all 6 existing call sites in `inject_message` (plus the one test call site) pass `None, None` — mechanical, zero behavior change for the ordinary jekt path.

## Scope
Backend schema only — no route/RPC changes in this PR. Part of the Warden Auto-Controller design (`docs/analysis/ANALYSIS_WARDEN_AUTO_CONTROLLER_CONTINUATION_WATCHER_2026_08_12.md`), alongside the already-merged rail restructure (#2553), which already renders these fields when present. Follow-ups in the same design (not in this PR): per-agent `auto_continue_enabled` config field, transcript-read + nudge REST routes/MCP tools, and the consecutive-nudge ceiling guardrail.

## Test plan
- [x] `cargo test -p agentmux-srv reactive::` — 72 passed (3 new: ordinary-jekt-has-no-outcome, log_audit-stores-outcome-when-provided, serde-roundtrip-omits-when-none)

### DIFF
--- a/.changesets/1786574599-feat-warden-extend-auditlogentry-with-outcome-reason-for-sup-krmt.md
+++ b/.changesets/1786574599-feat-warden-extend-auditlogentry-with-outcome-reason-for-sup-krmt.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(warden): extend AuditLogEntry with outcome/reason for Supervisor decisions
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2557, opened 2026-08-12
feat(warden): add SupervisorNudge route/MCP tool, nudge ceiling, and Supervisor manager UI

GenericAgentX-asaf:
## Summary

- New `POST /agentmux/reactive/supervisor-decision` route
  (`handle_reactive_supervisor_decision`): a Warden Supervisor watcher
  agent's decision about a target agent it just polled via
  `GetAgentTranscript`. `action: "nudge"` delivers `message` to the target
  through the same delivery path `SendMessage`/`Loop` use, audited as
  `outcome: "nudge_sent"`. `action: "decline"` sends nothing, audited as
  `outcome: "nudge_declined"`.
- **Entitlement gate**: a Nudge is refused with HTTP 403 unless the target
  agent's `AgentDefinition.auto_continue_enabled` is set — checked at the
  HTTP boundary (`Handler` has no `Store` access by design).
- `Handler::inject_message`'s body is now `inject_message_inner(req,
  outcome, reason)` — a shared delivery path threading `outcome`/`reason`
  into every `log_audit` call, so `record_supervisor_decision`'s `Nudge` arm
  reuses it instead of duplicating sanitize/deliver logic.
- **Consecutive-nudge ceiling** (`MAX_CONSECUTIVE_AUTO_CONTINUES = 5`):
  tracked per target agent in `Handler::nudge_counters`, keyed on the
  target's `registration_nonce` AND `block_id` (a respawn resets the
  counter — nonce alone misses PTY/shell-registered agents, which always
  register with nonce 0), and also reset after `NUDGE_COOLDOWN_RESET_MS`
  (30min) of inactivity. Swept on every Nudge to bound memory growth. A
  nudge that would exceed the ceiling is refused (forced decline, audited)
  and the caller gets `Err`/HTTP 429 — a signal to stop and escalate to a
  human instead of retrying.
- New `SupervisorNudge` MCP tool wraps the route, same `reqwest` POST +
  `X-AuthKey` pattern `SendMessage` already uses.
- **Supervisor manager UI**: `CommandUpdateAgentDefinitionData` gained
  `auto_continue_enabled` (`Option<i64>`, same shape as `use_ambient_login`).
  `WardenSupervisorManager` replaces the placeholder stub — lists user-owned
  agents with a toggle bound to the field (via
  `RpcApi.UpdateAgentDefinitionCommand`), plus a recent-decisions feed
  (`fetchWardenAudit`, filtered to entries with `outcome` set).

This completes the Warden Supervisor feature — see
`docs/analysis/ANALYSIS_WARDEN_AUTO_CONTROLLER_CONTINUATION_WATCHER_2026_08_12.md`.
Continues #2553/#2554/#2555/#2556.

## Test plan

- [x] `cargo build -p agentmux-srv` and `cargo build -p agentmux-mcp` — clean
- [x] `cargo test -p agentmux-srv` — 2180 tests passing, including:
  - Nudge-ceiling tests (both `registration_nonce`- and `block_id`-based respawn reset)
  - Entitlement-gate tests (rejected when unregistered/opted-out, passes when opted-in)
  - `record_supervisor_decision(Decline, ...)` audits without delivering
  - HTTP-level validation tests for `/agentmux/reactive/supervisor-decision`

### DIFF
--- a/.changesets/1786577572-feat-warden-add-supervisornudge-route-mcp-tool-consecutive-n-im48.md
+++ b/.changesets/1786577572-feat-warden-add-supervisornudge-route-mcp-tool-consecutive-n-im48.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(warden): add SupervisorNudge route + MCP tool + consecutive-nudge ceiling
--- a/.changesets/1786578516-feat-warden-add-supervisor-manager-ui-fix-nudge-entitlement--y0pg.md
+++ b/.changesets/1786578516-feat-warden-add-supervisor-manager-ui-fix-nudge-entitlement--y0pg.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(warden): add Supervisor manager UI + fix nudge entitlement/ceiling gaps
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -142,6 +142,21 @@ const GET_AGENT_TRANSCRIPT_TOOL: &str = r#"{
   }
 }"#;
 
+const SUPERVISOR_NUDGE_TOOL: &str = r#"{
+  "name": "SupervisorNudge",
+  "description": "For a Warden Supervisor watcher agent: record a decision about a target agent that just ended its turn, after inspecting it with GetAgentTranscript. action=\"nudge\" delivers `message` to the target as an ordinary jekt telling it to continue, and logs the decision. action=\"decline\" sends nothing and just logs that you chose not to nudge (e.g. the target isn't opted in, or looks genuinely done/blocked, not merely pausing to ask). A nudge that would exceed the server-side consecutive-nudge ceiling is refused (HTTP 429/tool error) — treat that as a signal to stop nudging this target and escalate to a human via SendMessage instead of retrying.",
+  "inputSchema": {
+    "type": "object",
+    "properties": {
+      "target_agent": { "type": "string", "description": "Name of the agent this decision is about (its AGENTMUX_AGENT_ID value)" },
+      "action":       { "type": "string", "enum": ["nudge", "decline"], "description": "Whether to nudge the target to continue or decline to" },
+      "message":      { "type": "string", "description": "Required when action is \"nudge\" — the message delivered to the target as a jekt" },
+      "reason":       { "type": "string", "description": "Your stated reasoning for this decision, recorded in the audit log" }
+    },
+    "required": ["target_agent", "action"]
+  }
+}"#;
+
 const SET_ACTIVE_TAB_TOOL: &str = r#"{
   "name": "SetActiveTab",
   "description": "Switch the active (foreground) tab to the given tab_id within its workspace. Get tab ids from Layout(query:\"tabs\") or Layout(query:\"layout\").",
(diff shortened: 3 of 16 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed in c5717f7f9:

- **[P1] auto_continue_enabled never checked before nudging** — fixed at the HTTP boundary (`handle_reactive_supervisor_decision`), since `Handler` has no `Store` access by design. A Nudge for a target with no matching `AgentDefinition`, or `auto_continue_enabled: 0`, is now refused with 403 before it reaches the handler. New tests: `reactive_supervisor_decision_nudge_rejected_when_not_opted_in`, `reactive_supervisor_decision_nudge_rejected_when_opted_out`, `reactive_supervisor_decision_nudge_passes_gate_when_opted_in`.
- **[P1] registration_nonce-only respawn detection missed PTY agents** — `register_agent` (PTY/shell/HTTP-register paths) always passes nonce 0, so nonce equality alone could never fire a reset for them. `NudgeCounterState` now also tracks `block_id`; a relaunch into a new pane resets the ceiling even when nonce stays 0. New test: `test_record_supervisor_decision_nudge_ceiling_resets_on_new_block_id_when_nonce_is_zero`.
- **[P2] nudge_counters unbounded growth** — now swept on every Nudge, dropping entries already past the cooldown window (behavior-neutral, since those would be treated as stale on next use anyway).

Also folded in the Supervisor manager UI (was going to be its own follow-up PR per the original plan, but it's tightly coupled to the `auto_continue_enabled` RPC wiring this fix touches anyway) — `CommandUpdateAgentDefinitionData` now carries the field, and `WardenSupervisorManager` replaces the placeholder stub with a real per-agent toggle list + recent-decisions feed.

Full verification: `cargo build`/`cargo test -p agentmux-srv` (2180 passing) and `npx tsc --noEmit`, all clean.

GenericAgentX-asaf:
Addressed in 2d341bee0 (second-round review):

- **[P1] free-form `message` violated §4.3's fixed-template guardrail** — `SupervisorAction::Nudge` is now a unit variant; delivered text is always `handler::NUDGE_MESSAGE`, a fixed constant. Neither the REST route nor the MCP tool accept a `message` parameter anymore. `reason` (the Supervisor's own stated reasoning) still travels separately, into the audit log only. New test: `reactive_supervisor_decision_nudge_ignores_a_stray_message_field` (a stray `message` field in the request body is silently ignored, not honored).
- **[P2] ceiling consumed by failed deliveries** — the counter now only increments after `inject_message_inner` reports `success`. New test: `test_record_supervisor_decision_nudge_failure_is_audited_and_not_counted`.
- **[P2] audit outcome hardcoded to "nudge_sent"** — `inject_message_inner` now takes separate `outcome_on_success`/`outcome_on_failure` params; a failed nudge is audited as the new `"nudge_failed"` outcome. Updated `AuditLogEntry`'s doc, the Audit manager's status rendering, and the Supervisor manager's decision feed to handle it (styled as an error, distinct from the neutral "declined").
- **[P2] MCP tool only checked HTTP status** — the route always returns 200 once past the entitlement/ceiling gates, so a failed delivery was reported to the calling agent as success. `SupervisorNudge` now also checks the response body's `success` field.

Full verification: `cargo test -p agentmux-srv` (2181 passing), `cargo build -p agentmux-mcp`, `npx tsc --noEmit` — all clean.

GenericAgentX-asaf:
Addressed in 04063e636:

- **[P0] Entitlement gate matched by \`AgentDefinition.name\` instead of \`.slug\`** — every delivery path keys registration on \`AGENTMUX_AGENT_ID\`, which \`agent_open.rs\` sets to the agent's stable \`slug\`, not its renameable display \`name\`. Fixed the lookup to match \`d.slug\`. Updated the two existing entitlement tests to use distinct slug/name values (so a regression can't pass by accident), and added \`reactive_supervisor_decision_nudge_slug_match_does_not_cross_into_a_colliding_display_name\`, mirroring \`agents.rs\`'s own \`instance_get_by_name_and_by_slug_never_cross_the_others_namespace\` precedent — agent A's slug is unrelated but its name collides with agent B's slug; a nudge for that shared string must resolve to (and use the entitlement of) agent B, never agent A.

Full verification: \`cargo test -p agentmux-srv\` — 2182 passing.

PR #2558, opened 2026-08-13
feat(agent): expose model vendor / custom endpoint UI + add Antigravity harness

GenericAgentX-asaf:
## Summary

Finishes the harness-vs-model-vendor decoupling work from #2505, which shipped the backend plumbing (`AgentDefinition.model_vendor_base_url`, `ProviderConfig.base_url_env_var`) but explicitly scoped out UI: *"settable via `agent.define` only; the human 'New Agent'/edit UI doesn't surface it yet."* Also completes the Antigravity (AGY) harness integration originally prototyped on the unmerged `feat/harness-model-decoupling-antigravity` branch (single commit, never opened as a PR, 73 commits stale) — rebuilt fresh against current `main` rather than cherry-picked.

**No new `AgentDefinition`/database fields.** The vendor concept is computed, not stored: `resolveEffectiveVendor(provider, model_vendor_base_url)` returns `"custom"` when an override is set, else the provider's own default vendor (a new static `supportedVendors` catalog field, mirrored Rust `ProviderConfig` ↔ TS `ProviderDefinition`).

### Vendor/endpoint UI
- `CommandAgentDefCreateFromTemplateData` and `CommandUpdateAgentDefinitionData` both gained an `Option<String> model_vendor_base_url` field (`None` = inherit/preserve, `Some("")` = explicit clear), wired through `template.rs`'s and `core.rs`'s handlers with the same `validate_vendor_base_url` check `agent.define` already uses (widened to `pub(crate)` and re-exported instead of duplicated).
- `AgentCreateFromTemplateModal` gained a "Model Vendor / Custom Endpoint" field, shown only when the template's provider declares `baseUrlEnvVar` (currently just claude).
- `AgentIdentityModal` gained the same field for editing an existing agent, reusing its `basePayload()`/`handleUpdate` round-trip pattern.

### Dual-icon display
- New `DualProviderLogo` component (primary harness icon + overlaid vendor badge, omitted when vendor === harness) wired into `AgentCard` and `MyAgentsList`. `RecentSessionRow` gained `model_vendor_base_url` so `MyAgentsList` can compute the badge without a second round trip.

### Antigravity (AGY) harness
- Added as a full provider entry (Rust `ProviderConfig` + registry/aliases, TS catalog + provider-id-aliases, `ProviderLogo` icon mapping reusing Gemini's, seed template) so it's actually pickable as a "New Agent" template — the harness has no route into the UI otherwise, since creation is template-driven.

## Deliberately out of scope (explicit decision during planning)

- A from-scratch 3-step creation wizard — investigation found agent creation has no from-scratch flow today (every agent clones a template, which already fixes the provider/harness), so the wizard proposal doesn't map onto the current UX.
- New SVG icons for `openrouter`/`ollama`/`alibaba` vendors — letter-circle fallback is fine for v1.
- `AgentPaneIcon.tsx`'s pane-tab icon — resolves provider from block meta, not an `AgentDefinition`; no vendor-base-url meta key exists, and plumbing one through the launch flow for a 16px icon wasn't judged worth it this pass.

## Test plan

- [x] `cargo test -p agentmux-srv` — 2185 tests passing (new: antigravity provider tests, alias tests)
- [x] `npx vitest run` — 2588 tests passing (175 files), including new `DualProviderLogo.test.tsx`, `resolve-effective-vendor.test.ts`, and 2 new `AgentCreateFromTemplateModal` tests (field shown/hidden by provider, submit payload)
- [ ] Manual smoke test (`task dev`, launching the real desktop app) — **not performed**, consistent with how the rest of this session's UI work was verified (typecheck + component tests only, no live GUI launch attempted in this environment)

### DIFF
--- a/.changesets/1786604409-feat-agent-expose-model-vendor-custom-endpoint-ui-add-antigr-81md.md
+++ b/.changesets/1786604409-feat-agent-expose-model-vendor-custom-endpoint-ui-add-antigr-81md.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): expose model vendor / custom endpoint UI + add Antigravity harness
--- a/agentmux-srv/agent-seed.json
+++ b/agentmux-srv/agent-seed.json
@@ -75,6 +75,30 @@
       ]
     },
     {
+      "id": "antigravity",
+      "name": "Antigravity",
+      "icon": "\u269b",
+      "provider": "antigravity",
+      "agent_type": "host",
+      "description": "Google's agentic coding CLI harness (AGY)",
+      "working_directory": "",
+      "shell": "pwsh",
+      "agent_bus_id": "antigravity",
+      "content": {
+        "env": "AGENT_NAME=antigravity\nAGENTMUX_AGENT_ID=Antigravity",
+        "startup": "__SKIP__"
+      },
+      "skills": [
+        {
+          "name": "Startup Verification",
+          "trigger": "startup",
+          "skill_type": "prompt",
+          "description": "Re-run tool verification checks and report status table",
+          "content": "Run the verification round from your startup instructions. Check identity (gh, git), dev-tools (@a5af/* packages \u2014 install if missing), secrets access, and MCP servers. Report a status table with OK/FAIL for each check. Fix any failures automatically if possible."
+        }
+      ]
+    },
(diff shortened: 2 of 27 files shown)


a5af:
@codex review

PR #2559, opened 2026-08-13
docs: retro + spec for pane-restore-on-relaunch and saved Layouts

AgentX-asaf:
## Summary
- Retro: the "restore panes on reopen" behavior users remember was never a deliberate feature — it was a leaked-window-row bug (traces commits `e3a6f85c2` and `4cbf856b7`/PR #2186) that made a graceful quit accidentally look like session restore. Fixing the leak correctly (as it should have been) means every quit now wipes the workspace, which is the regression being reported.
- Spec: scopes the real fix — (1) restore-on-relaunch, decoupled from the destroy-on-close cascade, and (2) a separate, explicitly requested "Layouts" feature (named, on-demand saved pane arrangements), including researched prior art (Wave Terminal Workspaces, tmux-resurrect, iTerm2 Arrangements, Zellij) and UI placement/naming decided with the user (hamburger menu, "Layouts" between Opacity and Settings).

Companion implementation PR for the spec's Feature 1 (restore-on-relaunch) follows separately.

## Test plan
- [x] Docs only — no code change in this PR.


a5af:
@codex review

PR #2560, opened 2026-08-13
feat(window): restore last session's tabs/panes on relaunch

AgentX-asaf:
## Summary
- Adds a real restore-on-relaunch mechanism, independent of the existing (correct) destroy-on-close cascade: `session_restore.rs` snapshots a workspace's tabs/blocks/layout tree into `Client.meta["session:last_topology"]` just before close, and replays it on the next cold start instead of always reseeding the hardcoded default 3-pane layout.
- Wired into `window_close.rs` (write) and `window_create.rs` (read, gated behind a new opt-in `restoreIfAvailable` arg set only by the frontend's genuine cold-start call site — tear-off / "Open New Window" are unaffected).
- See companion docs PR #2559 for the retro (why this regressed) and spec (full design, including a separately-specced "Layouts" save/load feature not included here).

Not included here (tracked as explicit follow-ups): a settings toggle to disable this, and the "Layouts" menu UI.

## Test plan
- [x] `cargo test -p agentmux-srv` — full suite (2193 tests) passes, no regressions.
- [x] New unit tests in `session_restore.rs` covering the snapshot/restore round trip, placeholder remapping, and idempotency.
- [x] New live integration test (`restore_last_session_survives_a_real_process_restart`) — spawns two separate real `agentmux-srv` processes against the same data dir, closes the first gracefully, kills it, cold-starts the second, and confirms a marker block created before the restart comes back. Proves durability across an actual process restart, not just an in-process call.
- [x] `npx tsc --noEmit` clean on the frontend change.
- [ ] Not live-tested through the actual Electron/CEF UI (`task dev`, closing/reopening the real app) — only through the real HTTP RPC API, which exercises the same server-side path but not the frontend/host wiring end-to-end.

### DIFF
--- a/.changesets/1786628662-feat-window-restore-last-session-s-tabs-panes-on-relaunch-o9os.md
+++ b/.changesets/1786628662-feat-window-restore-last-session-s-tabs-panes-on-relaunch-o9os.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(window): restore last session's tabs/panes on relaunch
--- a/.changesets/1786713585-fix-window-close-toctou-races-in-session-restore-snapshot-sa-rki0.md
+++ b/.changesets/1786713585-fix-window-close-toctou-races-in-session-restore-snapshot-sa-rki0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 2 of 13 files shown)


a5af:
@codex review

PR #2561, opened 2026-08-13
feat(armory,warden,settings): pane title reflects the selected rail section

AgentO-asaf:
## Summary
- Armory and Warden panes showed a static title ("Armory"/"Warden") no matter which rail section was active. Their selected-section state also lived as unpersisted local view state.
- Moved `section` onto each model as a meta-backed `sectionAtom` (mirrors the existing `zoomAtom`/`term:zoom` pattern already in both files), and made `viewName` a reactive memo over a label table. Side effect: the selected tab now survives a block remount instead of always resetting.
- Settings has the same rail UI but its section signal was already model-owned — only needed `viewName` wired to read it directly, no new persistence infrastructure.
- Added test coverage for all three (Settings previously had none at all). Along the way, fixed a real regression the change exposed in Warden's existing rail-click test: the RPC mock needed to actually round-trip into a reactive block-meta store for section switching to visibly take effect in tests, matching how it works against the real backend.

Full design/rationale: `docs/specs/SPEC_SECTIONED_PANE_DYNAMIC_TITLE_2026_08_12.md`

## Test plan
- [x] `tsc --noEmit` — clean, zero errors
- [x] Full `vitest run` — 176 files / 2604 tests passed, 0 failures (2 e2e skipped, no build dir — expected)
- [x] Armory/Warden: unit tests cover default title, click-to-switch (rail + tab bar) writing meta and updating `viewName()`/visible pane, and fallback for invalid meta
- [x] Settings: new baseline rail tests + `viewName()` reactivity tests
- [ ] Manual `task dev` check (not run in this environment — no display)

### DIFF
--- a/.changesets/1786629295-feat-armory-dynamic-pane-title-reflects-selected-rail-sectio-7se4.md
+++ b/.changesets/1786629295-feat-armory-dynamic-pane-title-reflects-selected-rail-sectio-7se4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(armory): dynamic pane title reflects selected rail section (also warden, settings)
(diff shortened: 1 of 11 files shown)


a5af:
@codex review

PR #2563, opened 2026-08-13
fix(agent-view): stop transcript divider rules from crossing pill text

AgentO-asaf:
## Summary
- The "context compacted" and "new session started" / "session continued" transcript dividers positioned their flanking horizontal rules at `top: 50%` of the whole pill+detail-line container instead of just the pill row — the rule landed near the pill's bottom edge (or crossed into the detail text on narrow panes) instead of framing the label cleanly.
- The pills also had no background, so the rule painted straight through the transparent label text instead of being masked behind it (unlike the already-correct `.agent-day-divider`, which has both).
- Fix: wrap each pill in its own single-row rule element (mirroring `.agent-day-divider`'s shape) and give the pill an explicit background. The rule now always bisects exactly the pill, and the detail line lives entirely outside the ruled row.
- `compaction_started` ("Compacting conversation…") is untouched — it never had rule lines to begin with, so it wasn't affected by this bug (documented as a correction in the spec).

See `docs/specs/SPEC_DIVIDER_PILL_RULE_MISALIGNMENT_2026_08_12.md` for full root-cause analysis and before/after.

## Test plan
- [x] `tsc --noEmit` — no errors
- [x] `stylelint` on the changed file — same 31 pre-existing errors before/after, nothing new introduced
- [x] `vitest run DocumentRow.test.tsx` — all 10 tests pass, including compaction-node tests
- [ ] Manual visual check in `task dev` (not run in this environment — no display)

### DIFF
--- a/.changesets/1786665803-fix-agent-view-stop-transcript-divider-rules-from-crossing-p-p7e3.md
+++ b/.changesets/1786665803-fix-agent-view-stop-transcript-divider-rules-from-crossing-p-p7e3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-view): stop transcript divider rules from crossing pill text
--- a/.changesets/1786665803-fix-agent-view-stop-transcript-divider-rules-from-crossing-p-p7e3.md
+++ b/.changesets/1786665803-fix-agent-view-stop-transcript-divider-rules-from-crossing-p-p7e3.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(agent-view): stop transcript divider rules from crossing pill text
--- a/.changesets/1786665908-fix-agent-view-stop-transcript-divider-rules-from-crossing-p-wuve.md
+++ b/.changesets/1786665908-fix-agent-view-stop-transcript-divider-rules-from-crossing-p-wuve.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-view): stop transcript divider rules from crossing pill text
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentO-asaf:
Re: [reagentx re-review](https://github.com/agentmuxai/agentmux/pull/2563#pullrequestreview-4932620017) — the P2 it raises (`--panel-bg-color` too translucent to mask the rule behind the pill) was Codex's original finding on the first commit, and it's already fixed as of `295cd59c1`/`364ab3855`.

That commit replaced the `position: absolute; width: 25%` + background-masking approach with real flex items (`flex: 1 1 auto` on `::before`/`::after`, inside a `display: flex` `.agent-*-rule` row). The lines and the pill now share the row's width via normal flex distribution — there's no `position: absolute` left anywhere in this diff, so there's nothing for the pill's background to need to mask. Overlap between a rule line and the pill is structurally impossible regardless of the pill's width or the background's opacity.

The `background: var(--panel-bg-color, ...)` on the pill that the review flagged (line reference is from the pre-fix version) is now purely cosmetic — it gives the pill a subtle chip background, matching `.agent-day-divider-label`'s existing styling — not doing any masking work.

Happy to adjust further if I'm missing something the flex layout doesn't actually cover — but as far as I can verify, this specific finding no longer applies to the current diff. Full root-cause writeup + the flex-layout rationale is in \`docs/specs/SPEC_DIVIDER_PILL_RULE_MISALIGNMENT_2026_08_12.md\`.

PR #2565, opened 2026-08-13
feat(jekt): host-tier per-agent HMAC signing for sender verification

AgentX-asaf:
## Summary
- Implements the never-built "Phase 5" from `SPEC_JEKT_SECURITY_AND_VISIBILITY_2026_07_01.md` §5.3 — see the new `docs/specs/SPEC_JEKT_TRUST_LAYER_COMPLETION_2026_08_13.md` for the full design/audit this PR implements.
- A raw call to the local jekt injection endpoint could previously self-declare any `source_agent` with zero verification, defaulting to the most-trusted delivery tier when omitted. This adds real, cryptographic host-tier sender verification: a per-agent HMAC-SHA256 key, minted server-side and injected only into that one agent's own MCP process env (`AGENTMUX_JEKT_KEY`, alongside `AGENTMUX_AGENT_ID`) — `SendMessage`/`Loop` sign outgoing jekts with it, and `handle_reactive_inject` verifies against the claimed sender's key on file.
- Verification is a genuine three-state result (`host-verified` / `unverified` / `self-declared`), not a bool — collapsing "never checked" into "verified" would mislabel every non-agent caller (Slack/Discord/Telegram/WhatsApp bridges) as cryptographically proven. Only escalates to `TIER=sensitive` when a key exists but the signature is missing or wrong — a caller with no key at all (rollout-safety case) is left unescalated, same as today.
- Network-tier (LAN/WAN) jekts remain unconditionally `TIER=sensitive` regardless of this signal — unchanged, this only strengthens host-tier identity, it doesn't touch the existing safety net.
- CLAUDE.md (this repo's copy and the outer agent-level mirror, kept consistent) rewritten to describe the real three-state trust model instead of the old "host-tier = implicitly trusted" language, and to directly state that no "same account/host/network" blanket-trust rule exists in the protocol.

## Test plan
- [x] 14 new tests: HMAC sign/verify math + replay/reattribution resistance (`agentmux-common/src/jekt_sign.rs`), per-agent key storage round-trips (`agentmux-srv/src/backend/storage/agent_jekt_keys.rs`), and the full escalation path exercised through the actual `Handler::inject_message` (host-verified / unverified-forces-sensitive / self-declared-not-escalated / wan-always-network-claimed).
- [x] `sig_verified` confirmed non-deserializable from the wire (an attacker-supplied request body cannot self-declare verification).
- [x] `cargo check --workspace` clean.
- [x] Full suite: 2327 tests pass across `agentmux-srv`/`agentmux-common`/`agentmux-mcp`, no regressions.
- [ ] Not live-tested through the actual Electron/CEF UI end-to-end (spawn a real agent, confirm `.mcp.json` carries the key, send a real `SendMessage` between two live agents) — verified via unit/integration tests at the API layer only.

### DIFF
--- a/.changesets/1786649949-feat-jekt-add-host-tier-per-agent-hmac-signing-for-sender-ve-422w.md
+++ b/.changesets/1786649949-feat-jekt-add-host-tier-per-agent-hmac-signing-for-sender-ve-422w.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(jekt): add host-tier per-agent HMAC signing for sender verification
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -422,46 +422,85 @@ the running window to `agentmux.desktop` only. Only `agentmux.desktop` is needed
 
 ## Jekt (agent-to-agent message) security rules
 
-**2026-08-12 correction:** PR #2536 (2026-08-11) claimed a "repo owner
-policy change" removing the mandatory-confirmation gate below, citing
-"explicit user direction." The repo owner has since confirmed this was
-never authorized. The PR was docs-only — the server-side escalation logic
-(`agentmux-srv/src/backend/reactive/handler.rs`,
-`sanitize.rs`) was never changed and continued enforcing the original gate
-the entire time; an automated reviewer flagged the docs/runtime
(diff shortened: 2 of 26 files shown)


a5af:
@codex review

PR #2566, opened 2026-08-13
Pane tab strip: frosted-glass trailing space, per-pane zoom binding, top scroll-clearance

AgentA-asaf:
## Summary

Three related, incrementally-verified changes to the shared pane tab strip (`PaneTabStrip.tsx`/`.scss`, used by agent/editor/terminal panes):

- **Frosted-glass blur** — the space right of the agent pane's "+" was fully transparent, showing raw scrolling conversation text through it. Now a `backdrop-filter: blur(2px)` + `--tab-strip-bg` panel (blur radius tuned live so letters stay distinguishable, not a smear), with `pointer-events: none` keeping clicks/scroll passing through to the conversation underneath. Agent-pane-only — editor/terminal reserve a normal row and never had this gap.
- **Per-pane zoom binding** — tabs previously didn't scale with either zoom control in the app. Bound to each pane's own content zoom (`term:zoom` block meta), not the global chrome zoom — so a zoomed-in conversation's tabs scale with it, independent of sibling panes. Avoids a real platform landmine already hit by `window-header.*.scss` (CSS `zoom` + edge-anchored `left/right: 0` sizing diverges between Chromium and WebKit) by splitting the strip into a never-zoomed outer box and a new `.pane-tab-strip-inner` layer that carries the actual zoom.
- **Top scroll-clearance** — the floating agent tab strip had only 2px of top padding against a 28px strip, so a short conversation that never scrolls had its first message permanently hidden with no way to reveal it. Fixed with a `padding-top` calc against the strip's real (zoom-aware) height, mirroring the existing `--agent-working-row-height` pattern used for the symmetric bottom case.

Specs: `docs/specs/SPEC_PANE_TAB_STRIP_TRAILING_BLUR_2026_08_12.md`, `docs/specs/SPEC_PANE_TAB_STRIP_CHROME_ZOOM_AND_SCROLL_CLEARANCE_2026_08_12.md` (the latter documents an earlier wrong draft — global chrome zoom instead of per-pane — that was corrected before landing here).

## Test plan

- [x] `stylelint` on all changed files — only pre-existing warnings, none introduced
- [x] Full test suite (93 files / 1177 tests) — passing, including one intentionally-updated test (`PaneTabStrip.test.tsx`, DOM structure now has an inner wrapper layer)
- [x] Production `vite build` — compiles cleanly
- [x] Live-verified in `task dev` via CDP: blur radius, pointer-events passthrough on the trailing area, `+` button still hit-testable, real `Ctrl+Wheel` zoom gesture scaling tabs/content together, and the strip's right edge staying pixel-identical to the pane's real right edge at zoom (zero drift — confirms the platform-landmine avoidance)
- [ ] Real macOS/Linux check — not possible from this Windows dev machine; the design was chosen specifically to avoid needing platform-specific compensation, but that's inferred from a comment on a different component, not independently confirmed on WebKit

### DIFF
--- a/.changesets/1786666690-feat-pane-tab-strip-frosted-glass-trailing-space-per-pane-zo-6jlz.md
+++ b/.changesets/1786666690-feat-pane-tab-strip-frosted-glass-trailing-space-per-pane-zo-6jlz.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(pane-tab-strip): frosted-glass trailing space, per-pane zoom binding, top scroll-clearance for short agent conversations
--- a/docs/specs/SPEC_PANE_TAB_STRIP_CHROME_ZOOM_AND_SCROLL_CLEARANCE_2026_08_12.md
+++ b/docs/specs/SPEC_PANE_TAB_STRIP_CHROME_ZOOM_AND_SCROLL_CLEARANCE_2026_08_12.md
@@ -2,22 +2,32 @@
 
 **Date:** 2026-08-12 (Part A corrected before implementation — see §A.0's
 note).
-**Status:** implemented and live-verified against this corrected design.
-(History: an earlier draft bound to the wrong zoom control — global
-chrome zoom, `--zoomfactor` — was implemented and live-verified, then
-corrected to the per-pane design below before landing anywhere.) Live
-check used a **real Ctrl+Wheel zoom gesture** (CDP-dispatched, not a
-simulated CSS variable override) on an agent pane: tabs, `+`, and
-conversation content scaled together (0.97 → 1.3, matching the pane's
-actual persisted `term:zoom`), `--pane-tab-strip-zoom` and the
-pre-existing `--agent-pane-zoom` agreed exactly at every step (both
-derive from the same block meta via separate paths — §A.3/§B.3), and the
-strip's right edge stayed pixel-identical to the pane's real right edge
-at zoom (312.5 = 312.5, zero drift) — the one measurement that directly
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

AgentA-asaf:
Addressed reagentx's review:

- **P1 (editor double-zoom)**: fixed by removing the `zoomFactor` prop from the `<PaneTabStrip>` call in `editor-tab-strip.tsx` entirely, with a comment explaining why — `<EditorTabStrip>` renders inside `.editor-view`'s already-`zoom`-scaled subtree, so ambient zoom already covers the whole tab strip for free; passing `zoomFactor` was compounding it. Re-checked terminal too (not flagged, but same class of bug was possible) — confirmed it has no ambient CSS `zoom` anywhere in its tree, so its wiring was already correct.
- **P1 (agent scroll-clearance double-scale)**: fixed by dropping the `* var(--agent-pane-zoom, 1)` multiplication from `_document.scss`'s `padding-top` calc — `.agent-document` is already inside `.agent-view`'s zoomed subtree, so the ambient zoom was already scaling the written value; multiplying again double-scaled it. Live-verified via CDP that the clearance gap now scales linearly with zoom (0.75x → 4.5px, 1.24x → 7.42px, ratio 1.65 ≈ the 1.65x zoom ratio, not the ~2.7x a squared bug would produce).
- **P2 (missing changeset)**: added via `task changeset -- minor`.

Spec doc updated with both corrections and an honest accounting of why the original live verification (agent-pane-only) missed these.

PR #2567, opened 2026-08-13
fix(menu): context-submenu wrong-strategy positioning + main window silent show() no-op

AgentY-asaf:
## Summary

Two bugs found while trying to live-verify PR #2525
("fix(menu): submenu positioning flash + safe-triangle hover close timing")
at the operator's request — that PR's own test plan flagged that manual
`task dev` verification was never completed, and it turns out there was
a real gap.

**1. `cef-api.ts`'s context-submenu still renders off the paintable area**
("Replace With...", "Pane Color", etc.) — e.g. `top:-393px` on a
1176px-tall window, despite #2525's `visibility:hidden`-until-positioned
guard.

Root cause: `sub` starts `position:absolute` (a deliberate row-relative
fallback for when `computeMenuPosition` rejects — see the comment it's
already carrying), but `computeMenuPosition` computes coordinates for
`strategy:"fixed"`. floating-ui resolves the floating element's offset
parent from its *current* position at measurement time, so measuring
while `sub` is still `position:absolute` (nested inside `row`, itself
`position:relative`) resolves against the wrong containing block
entirely. This reproduced **deterministically regardless of anchor
position** — the exact same wrong constant every time — which is what
ruled out a mere layout-timing race (my first, incorrect theory).

`flyoutmenu.tsx`'s Solid `SubMenu` sibling never hits this because its
placeholder style is already `position:fixed;left:0px;top:0px` *before*
it ever calls `computeMenuPosition`. Fix: switch `sub` to
`position:fixed` before measuring, and restore the absolute row-relative
fallback explicitly if placement itself fails (so a rejection doesn't
regress to `position:fixed; left:100%`, which would pin it off the right
edge of the screen). Also deferred measurement one `requestAnimationFrame`
(matching `flyoutmenu.tsx`'s own `registerSubMenu`) so the
`display:none`→`""` reflow settles first.

**2. `agentmux-cef`'s main window intermittently never appears** — stuck
showing a stray pool window instead (pool windows are meant to stay
invisible off-screen until explicitly promoted for tab tear-off).

`on_load_end`'s `window.show()` call sits inside two nested
`if let Some(...)` guards (`browser_view_get_for_browser`, `bv.window()`)
that silently no-op if either resolves to `None` — no log, no retry, no
diagnostic trace at all. This is the same CEF Views timing quirk the
pool-window path already diagnosed and worked around
(`POOL_HWND_CACHE`, `SPEC_POOL_WINDOW_HWND_NULL_2026_05_06.md`), but the
workaround was never ported to the main-window path. Converted the
silent no-op into a logged, bounded retry (`ShowWindowRetryTask`, mirrors
the existing Linux paint-gate retry pattern already in the same file).
Full writeup: `docs/retro/RETRO_MAIN_WINDOW_SHOW_SILENT_NOOP_2026_08_13.md`.

Both confirmed live via `task dev` on the affected dev machine, including
the operator directly reproducing and then confirming the fix.

## Test plan

- [x] `cargo check -p agentmux-cef` — clean, no new warnings
- [x] Live `task dev` verification — reproduced both bugs, applied fixes,
      reproduced again on a fresh instance to confirm each fix, operator
      confirmed the submenu positioning is fixed
- [ ] Full `npm test` suite — not run this session (frontend change is in
      a `.ts` utility file with no existing unit tests around this exact
      code path; flagging rather than claiming coverage)

### DIFF
--- a/.changesets/1786650533-fix-menu-context-submenu-wrong-strategy-positioning-main-win-f75t.md
+++ b/.changesets/1786650533-fix-menu-context-submenu-wrong-strategy-positioning-main-win-f75t.md
@@ -1,0 +1,4 @@
+---
+type: patch
+---
+fix(menu): context-submenu wrong-strategy positioning + main window silent show() no-op
--- a/agentmux-cef/src/client/navigation.rs
+++ b/agentmux-cef/src/client/navigation.rs
@@ -163,6 +163,72 @@ pub(crate) fn on_frontend_first_paint(state: Arc<AppState>, label: String) {
     post_task(ThreadId::UI, Some(&mut task));
 }
 
+/// Bound on how many times `on_load_end`'s top-level `window.show()` retries
+/// when `browser_view_get_for_browser()`/`BrowserView::window()` return
+/// `None` at load-end, instead of the previous silent no-op that left the
+/// window permanently hidden with zero diagnostic trace. This is the same
+/// class of CEF Views timing quirk already diagnosed and worked around on
+/// the pool-window path (`commands/window_pool.rs`'s `POOL_HWND_CACHE`,
+/// `SPEC_POOL_WINDOW_HWND_NULL_2026_05_06.md`) but was never ported to the
+/// main/top-level window path — see
+/// `docs/retro/RETRO_MAIN_WINDOW_SHOW_SILENT_NOOP_2026_08_13.md`.
+const SHOW_WINDOW_MAX_RETRIES: u32 = 10;
+/// Delay between retries — short enough that a handful still resolves within
+/// a user-imperceptible instant if the window becomes resolvable a frame or
+/// two later, long enough not to busy-loop the UI thread.
+const SHOW_WINDOW_RETRY_DELAY_MS: i64 = 50;
+
+/// Show + focus a top-level (non-pool) window by label if its `BrowserView`/
+/// `Window` are resolvable and it isn't already visible. Returns `true` once
+/// resolved (shown or already visible), `false` if the CEF Views objects
+/// aren't available yet — caller should retry rather than give up silently.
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentY-asaf:
Fixed in ff91528 — extracted the platform-branching show/gate logic into a shared \`reveal_top_level_window()\` used by both \`on_load_end\`'s primary path and \`ShowWindowRetryTask\`'s retry path, so the Linux paint-gate (\`linux_paint_gate_pending\`/\`reveal_gated_window\`) is honored regardless of which pass actually resolves the window. Thanks reagentx and codex — same issue independently caught by both.

Also already fixed in e24ec91 (pushed before these reviews landed): the submenu placement tests now flush the added \`requestAnimationFrame\` deferral via \`vi.advanceTimersByTimeAsync(20)\` (this vitest version, 3.2.7, has no \`advanceTimersToNextFrameAsync\`).

\`cargo check -p agentmux-cef\` clean (53 warnings, matching pre-existing baseline exactly — zero new). Full \`npx vitest run\` passes (2588/2588, up from the 1 pre-existing failure this PR introduced and then fixed).

AgentY-asaf:
The latest reagentx review (2026-08-14T00:12:00Z) analyzed commit \`e24ec91\` — that's my *test* fix, pushed one commit before the actual Linux paint-gate fix in \`ff91528\`. Current HEAD is \`ff91528\`; \`navigation.rs\` no longer has an unconditional \`window.show()\` in the retry path — see \`reveal_top_level_window()\` (now shared by both the primary \`on_load_end\` path and \`ShowWindowRetryTask\`, both routed through the same Linux gate). This looks like review-trigger lag against an intermediate commit rather than a live finding — flagging so it doesn't block on a stale analysis.

PR #2569, opened 2026-08-14
fix(bashwrap): harden two Windows-CI-flaky PTY tests against runner load

AgentX-asaf:
## Summary
- `bash_wrap::tests::run_via_pty_does_not_misclassify_fast_success_as_idle_timeout` and `a1_e2e_static_label_then_delayed_cr_overwrite_collapses_to_one_line` have been intermittently failing `check --tests + test (windows-latest)` — confirmed reproduced on `main` itself (run 31756237799, an unrelated tab-color-picker commit), not caused by any recent bashwrap change or by PRs #2559/#2560 that this was originally spotted blocking.
- Fast-success test: `AGENTMUX_BASHWRAP_IDLE_TIMEOUT_SECS=1` was tight enough that a loaded CI runner's PTY-reader-thread scheduling delay alone could exceed it before `echo hello` produced output — a genuine idle-kill by the test's own short timeout, not the idle_rx/wait_task misclassification race it exists to catch. Raised to 3s.
- A1 e2e test — **superseded by a full rewrite of the retry mechanism** (reagentx caught a real process leak in the first version of this fix): a hung/slow attempt used to be retried by wrapping `run_via_pty` in an external `tokio::time::timeout` and `continue`ing past the deadline on expiry. That leaked a process — dropping the outer future doesn't abort `wait_task`/`pty_reader_loop` (independent `spawn_blocking` tasks already running by the time the timeout could preempt anything), so the underlying bash/PTY child from the timed-out attempt kept running detached, up to `MAX_ATTEMPTS - 1` of them per flaky run, none ever cleaned up.

  The fix instead reuses `run_via_pty`'s own correct internal idle-kill mechanism (already used for the "hung command" case in production — tree-kills the child via `kill_process_tree` + `clone_killer`, returns exit code 124) rather than fighting it with a second, incomplete external timeout: `AGENTMUX_BASHWRAP_IDLE_TIMEOUT_SECS` is lowered to `15` for just this test (generous for the ~70ms command under test, still bounded), and the loop awaits `run_via_pty` directly — no external timeout, no orphaned process. Exit code `124` is treated as "this attempt got stuck, retry" (already cleaned up by the time it returns), same retry treatment as a wrong-occurrence-count attempt.
- Neither change loosens what's actually being verified (idle-kill classification correctness, single-occurrence collapse) — both are pure CI-runner-load tolerance for tests whose own doc comments already acknowledged this exact flakiness class.
- Follow-up fixes from review: the idle-kill retry path wasn't updating the diagnostic message shown on a final failure (only the occurrence-mismatch path was) — now both paths update it. The env var override was being restored manually at the end of the function, which skipped on a panicking exit (unwinding past the restore code) — switched to `EnvVarGuard`, an existing RAII guard in this same file, whose `Drop` impl restores it on every exit path including a panic.

## Test plan
- [x] `cargo test -p agentmux-bashwrap` — 91 + 1 passed, 0 failed
- [x] Confirmed the two target tests individually pass (`--test-threads=1 --nocapture`)
- [x] Confirmed sibling positive-case test `run_via_pty_kills_idle_child_and_returns_promptly` (idle-kill SHOULD fire) still passes unchanged
- [x] CI (windows-latest / ubuntu-latest / vitest) — green

### DIFF
--- a/.changesets/1786712782-fix-bashwrap-stop-leaking-orphaned-bash-pty-processes-on-a1--618o.md
+++ b/.changesets/1786712782-fix-bashwrap-stop-leaking-orphaned-bash-pty-processes-on-a1--618o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(bashwrap): stop leaking orphaned bash/PTY processes on a1 test retry timeout
--- a/agentmux-bashwrap/src/bash_wrap.rs
+++ b/agentmux-bashwrap/src/bash_wrap.rs
@@ -3270,7 +3270,12 @@ mod tests {
         }
 
         const MAX_ATTEMPTS: u32 = 5;
-        let mut last_blob = String::new();
+        // reagentx P2 on PR #2559 (re-review): must be updated on EVERY
+        // retry path, not just the occurrence-mismatch one — the idle-kill
+        // branch below used to leave this at its initial empty value, so if
+        // every attempt got idle-killed the final panic misleadingly
+        // reported an empty blob instead of indicating a timeout occurred.
+        let mut last_outcome = String::new();
         let mut collapsed = false;
         let mut skipped = false;
         for attempt in 1..=MAX_ATTEMPTS {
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2572, opened 2026-08-14
fix(lan): stop broadcasting the full-access auth_key to LAN peers

AgentX-asaf:
## Summary
- LAN discovery (mDNS TXT record + UDP broadcast responder) used to hand out the SAME `auth_key` that gates the entire `/agentmux/service` API surface to anything that could receive an mDNS multicast packet or send a UDP probe from a private-looking source address. A passive listener who captured it once had standing full-instance access for as long as it kept running, not just visibility into forwarded jekt traffic.
- Mints a separate `lan_key` (`Config::lan_key`), freshly generated per process launch, broadcast in place of `auth_key`.
- New `lan_or_full_auth_middleware` accepts it ONLY on the two routes LAN peer forwarding actually needs (`/agentmux/reactive/agent`, `/agentmux/reactive/inject`) — applied via a standalone `lan_forward_routes` router merged at the top level, not nested inside `authed_routes` (nesting would put `authed_routes`'s own strict `auth_middleware` on the outside, rejecting the `lan_key` before the permissive middleware ever ran). Every other route keeps requiring the full `auth_key`, completely unaffected.
- The wire field name (mDNS TXT property / UDP JSON key) stays `"auth_key"` for backward compatibility with peer instances running an older version — only the *value* it carries changed, not the wire shape.

## What this doesn't change
- LAN discovery stays opt-in, default off.
- Broadcasting a credential in cleartext to the LAN is still this feature's accepted trade-off (mDNS/UDP have no confidentiality of their own) — this shrinks *what a captured value is worth* (two narrow routes, not the whole instance), it doesn't hide it in transit. LAN P1-1 (encrypt/authenticate the transport itself) from `docs/specs/SPEC_JEKT_LAN_WAN_TRUST_HARDENING_2026_08_13.md` remains open, not attempted here.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2210 passed (8 new: `lan_key` accepted on both LAN-forward routes, full `auth_key` still works, garbage key rejected, `lan_key` rejected on the general service route and on other reactive routes, `lan_key` generated non-empty/distinct/fresh-per-launch)
- [x] `cargo check --workspace` clean
- [ ] CI — pending

### DIFF
--- a/.changesets/1786693117-fix-lan-stop-broadcasting-the-full-access-auth-key-to-lan-pe-0yzz.md
+++ b/.changesets/1786693117-fix-lan-stop-broadcasting-the-full-access-auth-key-to-lan-pe-0yzz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(lan): stop broadcasting the full-access auth_key to LAN peers
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -581,16 +581,26 @@ impl LanDiscoveryController {
     }
 
     /// Query LAN peers for which one hosts `agent_id`. Returns `(peer_url,
-    /// auth_key)` for the first peer that responds 2xx to the agent-lookup
+    /// lan_key)` for the first peer that responds 2xx to the agent-lookup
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2573, opened 2026-08-14
fix(muxbus): treat a 403 (agent-binding rejection) the same as 401 in cloud_subscriber

AgentX-asaf:
## Summary
- `sync_agent_reactive` only ever retried-with-shared-token on HTTP 401 — a `checkAgentBinding` rejection (403) fell into the generic non-2xx branch, silently stalling that agent's WAN delivery for the rest of the poll cycle with no retry and no user-visible error.
- Both mean the same thing operationally ("this credential can't be used for this agent"), so both now trigger identical recovery: invalidate the per-agent credential and retry immediately with the shared token, or reconnect the whole session if the shared token itself was what got rejected.
- Extracted the status check into a pure `is_credential_rejected()` so the fix is unit-testable without a live muxbus endpoint (`MUXBUS_REST_URL` is a hardcoded constant, not mockable).

## Context
This closes the operational gap `docs/specs/SPEC_JEKT_LAN_WAN_TRUST_HARDENING_2026_08_13.md` §5.2 identified as a prerequisite before `ENFORCE_AGENT_BINDING` is safe to flip on in `agentmux-cloud` — without it, a genuine binding mismatch would degrade to a silent stall instead of falling back the same way an expired token already does. Does **not** itself flip the flag or change server-side binding enforcement — both remain separate, larger pieces of work (tracked separately).

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2206 passed (4 new: 401 and 403 both treated as credential-rejected, other 4xx/5xx/2xx are not)
- [x] `cargo check --workspace` clean
- [ ] CI — pending

### DIFF
--- a/.changesets/1786693223-fix-muxbus-treat-a-403-agent-binding-rejection-the-same-as-4-9qky.md
+++ b/.changesets/1786693223-fix-muxbus-treat-a-403-agent-binding-rejection-the-same-as-4-9qky.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(muxbus): treat a 403 (agent-binding rejection) the same as 401 in cloud_subscriber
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2574, opened 2026-08-14
fix(agent-pane): stop truncating AskUserQuestion answers in tool summary row

Agent1-asaf:
## Summary

- The collapsed tool row's default styling (`white-space: nowrap` +
  `text-overflow: ellipsis` on `.agent-tool-name`) is right for command
  names and file paths, but it was also silently clipping the printed
  answer after an `AskUserQuestion` prompt to one line — a longer selected
  answer, or a free-text "Other" response, became unreadable without
  hovering for the peek tooltip.
- Scoped via the `data-tool="askuserquestion"` attribute already set on
  `.agent-tool-block` (same pattern already used for per-tool colors right
  below it in `_document-nodes.scss`): switches to `white-space: pre-wrap`
  + `overflow-wrap: anywhere`, matching how normal typed user messages
  already wrap in `UserMessageBlock.tsx`. Font-size/line-height untouched
  — both inherit the same `var(--termfontsize)`/`1.35` from `.agent-view`.
- Applies to both the live pane and the read-only history tab, since both
  render tool rows through the same `ToolBlock.tsx`.

## Test plan

- [x] `npx stylelint frontend/app/view/agent/styles/_document-nodes.scss`
      — 31 pre-existing errors, confirmed identical count on `main` before
      this change (none introduced)
- [ ] Visual check: ask a question with a long answer, confirm it now wraps
      and stays visible in both the live pane and the history tab

### DIFF
--- a/.changesets/1786714074-fix-agent-pane-stop-truncating-askuserquestion-answers-in-to-67jz.md
+++ b/.changesets/1786714074-fix-agent-pane-stop-truncating-askuserquestion-answers-in-to-67jz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): stop truncating AskUserQuestion answers in tool summary row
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #994, opened 2026-05-23
feat(agent-pane): bounded Submitting → Done.errored on timeout (turn-phase PR D)

AgentA-asaf:
**PR D of the agent-pane turn-phase migration.** Reducer-only — bounded `Submitting → Done.errored` watchdog. Closes the second gap in issue #728 from the reducer side.

## Why

Today, a turn enters `Submitting` via `TurnStart` and depends entirely on the backend acking with a chunk / `TokensIn` / `TokensOut` / `ToolStart` event to promote to `Streaming` (via `StreamFlushObserved` or `bumpEvent`). If the backend never responds — network drop, agent crash before any output, RPC hang post-send — the pane stays in `Submitting` forever. There is no `isWorking → false` signal, so the working animation runs forever.

## What

Mirror PR C's bounded-state pattern. New contract:

- **`SUBMIT_TIMEOUT_MS = 30_000`** constant exported from `types.ts` alongside `INTERRUPT_TIMEOUT_MS`. 30 s is the outer bound for a healthy backend ack — handshake, agent spawn, and the first model token round-trip.
- **`SubmitTimeoutElapsed { at: number }`** added to `AgentPaneCommand` union.
- **`schedule-submit-timeout { deadlineMs: number }`** added to `AgentPaneEvent` union, emitted by `TurnStart` whenever the phase crosses INTO `Submitting`.
- **`submit-timed-out { at: number }`** audit event emitted when the bounded transition fires.
- **`TurnStart` arm**: emits `schedule-submit-timeout` only on entry — does not re-arm if the phase is already `Submitting` (mirrors PR C's entry-only rule for the interrupt watchdog).
- **`SubmitTimeoutElapsed` arm**:
  - If `state.turnPhase.kind === "Submitting"`: force-transition to `Done.errored` with `finishedAt = command.at`, plus the standard legacy cleanup (`turnActive=false`, `stopping=false`, `currentTool=null`, `turnTokens=null`). Emits `submit-timed-out`.
  - Else (`Streaming`, `Interrupting`, `Done`, `Disconnected`, `Idle`): no-op (same-ref).

The promotion arms (`bumpEvent`, `StreamFlushObserved`) implicitly cancel the timeout via the dispatch-side cancel-flag pattern — the reducer doesn't need to know about that; the timeout's own arm checks the current phase.

Three independent paths out of `Submitting` are now pinned:
1. promotion via `bumpEvent` (`ToolStart` / `TokensIn` / `TokensOut`)
[...]

PR #2575, opened 2026-08-14
fix(agent-pane): wire up SubmitTimeoutElapsed dispatch + visibility catch-up watchdog tick

GenericAgentX-asaf:
## Summary

Root-caused a live "AgentA stuck showing Working…" incident to two confirmed dispatch-side gaps in the turn-phase state machine. Full investigation, log/code/git-history evidence, and ruled-out hypotheses: `docs/reports/REPORT_AGENTA_STUCK_WORKING_INVESTIGATION_2026_08_14.md` (included in this PR).

1. **`SubmitTimeoutElapsed` had zero consumers anywhere in the app.** PR D (#994, 2026-05-23) shipped the reducer-only half of the bounded `Submitting → Done.errored` timeout and explicitly deferred "dispatch-side setTimeout in useAgentStream / model layer" to a follow-up — confirmed via a repo-wide grep that it never landed. A turn whose backend ack was lost (dropped RPC, network blip, crash before first output) left the pane in `Submitting` — "Working…" — forever, with no recovery path at all.

2. **The stream watchdog's periodic tick showed zero evidence of running during a real, confirmed 29+ minute incident.** A stray/late `StreamFlushObserved` re-promoted a legitimately-completed turn back into `Streaming` (an accepted, documented edge case needed for real multi-round tool continuations), and the watchdog's own 180s unconditional recovery for that exact shape never fired. Component-unmount (`workspace.tsx` keeps inactive tabs mounted via `display:none`) and a reducer logic bug were both ruled out; the leading remaining explanation is renderer/window-level timer throttling for a backgrounded pane in this app's multi-window CEF architecture.

## Fix

Both changes are in `useTurnLifecycle.ts`, mirroring the file's own existing `stopFallbackTimer` pattern (a reactive effect on `turnPhase.kind`, not a new event-consumption mechanism):

- A `createEffect` arms/clears a real `SUBMIT_TIMEOUT_MS` `setTimeout` while `turnPhase.kind === "Submitting"`, dispatching the reducer's own `SubmitTimeoutElapsed` command so the transition goes through the documented, already-tested contract.
- A `document.visibilitychange` listener fires one extra `StreamWatchdogTick` the instant the document becomes visible again — a backstop for the interval. `visibilitychange` fires reliably even for throttled content, and the watchdog's recovery math is already wall-clock-based (`nowMs - lastEventMs`), so a single catch-up tick correctly recovers a stuck pane regardless of how long the interval stalled.

Both are additive to an existing, already-tested reducer contract — no `reducer.ts` changes.

## Test plan

- [x] `npx tsc --noEmit` — zero new errors (2 pre-existing, unrelated errors confirmed present on clean `main` too)
- [x] `frontend/app/store/agent-pane-state/reducer.test.ts` — 191/191 pass, no regression
- [x] New `useTurnLifecycle.test.ts` (6 tests, all passing): submit-timeout fires/doesn't-fire-if-phase-moved/re-arms-on-second-episode; watchdog fires-on-visible/not-on-hidden/cleans-up-on-dispose
- [x] Full `frontend/app/view/agent` suite — 1149/1149 pass, no regression
- [ ] Not verified live in the running app (no access to a live repro of the original incident within this session) — the leading hypothesis for gap 2 (renderer timer throttling) is evidence-consistent but not confirmed via live DevTools inspection; see report §6


a5af:
@codex review

GenericAgentX-asaf:
Addressed both review rounds:

1. **codex P1** (accept-but-slow-to-stream misfire): fixed in `3a01dc9a` — the submit-timeout timer is now cancelled on `AgentMessageAccepted`, narrowing it to only "message never reached the backend at all" instead of also bounding legitimate slow-but-accepted turns (which are already covered server-side by `HealthMonitor`'s Stalled/Dead detection).

2. **reagentx P1, round 2** (stale accepted-event cross-episode disarm): fixed in `e237084e` — moved both the timer handle and the `AgentMessageAccepted` subscription inside the `createEffect` as fresh per-episode local closures, so a stale/late accepted event for an earlier episode has nothing left to disarm once a newer episode has started. New test captures a stale handler reference directly and proves it can't affect a later episode's timer.

3. **reagentx P2** (changeset filename missing collision-avoidance suffix): fixed in `3a01dc9a` — regenerated through `scripts/changeset.sh` instead of hand-authoring.

All fixes verified: `tsc --noEmit` clean (2 pre-existing, unrelated errors only), 9/9 `useTurnLifecycle.test.ts`, 1343/1343 full agent-view + reducer suite.

PR #2576, opened 2026-08-14
feat(jekt): relax TIER=sensitive for WAN jekts verified against reagent's Ed25519 signature

AgentX-asaf:
## Summary
- A WAN jekt carrying `SIG=verified` (Ed25519 signature checked against reagent's pinned public key) is no longer forced to `TIER=sensitive` by delivery tier alone -- mirrors how host-tier's `TRUST=host-verified` already doesn't force it. Declared-sensitive and keyword-match escalation still apply unconditionally on top of this.
- `TRUST` is unchanged -- still renders `network-claimed` regardless of verification. Crossing the network boundary is still true; what changes is whether a human must confirm before acting.
- Deliberately scoped to reagent specifically, not WAN traffic in general: arbitrary agent-to-agent WAN jekts have no per-message signing mechanism yet (account-scoped Cognito binding authenticates the API *caller*, not a signature over the message itself), so self-declared `source_agent` claims remain exactly as forgeable as before and stay forced to sensitive.
- This is a genuine change to the jekt security policy in `CLAUDE.md` -- explicitly requested and confirmed by the repo owner in-conversation (an `AskUserQuestion` round distinguishing "just document what SIG=verified means" from "actually relax the sensitive-tier gate," the latter picked explicitly). Recorded in `docs/specs/SPEC_JEKT_REAGENT_TRUST_RELAXATION_2026_08_14.md` with the full rationale, specifically so it's traceable and distinguishable from an unverified inline claim (see that spec's §3, and CLAUDE.md's updated history note referencing the prior PR #2536 fake-policy-change incident for contrast).
- Updated both copies of the jekt security rules (`amx/CLAUDE.md`, the version-controlled source of truth, and the parent `agents/CLAUDE.md` which must match it exactly).

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` -- 2251 passed, 0 failed
- [x] New/updated coverage: verified-WAN-relaxes-to-coord + TRUST label unchanged + no sensitive-warning banner; still-escalates-on-declared-sensitive; still-escalates-on-keyword-match; invalid-signature-stays-sensitive (unchanged); no-signature-stays-sensitive (unchanged, now with an explicit assertion)


a5af:
@codex review

PR #2577, opened 2026-08-14
fix(agent-pane): composer strip stable width + deliberate edge-split tiers

GenericAgentX-asaf:
## Summary
- Fixes the composer strip (Shell/stats/model-selector row above the agent pane textarea) toggling annoyingly between 1 and 2 lines with no pane resize: the live-ticking stats text (tokens/elapsed, `useTick(1000)`) had no width floor, so a wrap decision sitting near the fit boundary could flip every second. Added `min-width: 12ch` to absorb the common tick-to-tick jitter.
- Reverts the "always left-justify below the widest tier" design (SPEC_COMPOSER_STRIP_LEFT_JUSTIFIED_TIERED_WRAP_2026_08_03) per direct feedback, and corrects an intermediate "center everything as one group" revision (also per feedback, same day) to the final design: each zone (controls/stats/right) keeps a fixed left/center/right identity at every tier — matching the widest (≥482px) tier's own edge-pinned layout — redistributed across up to 3 lines as the pane narrows, rather than collapsing into a centered blob or wrapping wherever content organically ran out of room.
- Full design writeup, including two same-day corrections from live `task dev` review, in `docs/specs/SPEC_COMPOSER_STRIP_CENTERED_SMART_SPLIT_2026_08_14.md`.

## Test plan
- [x] `npx stylelint frontend/app/view/agent/styles/_composer-strip.scss` — one pre-existing unrelated hex-color warning, same as before this change
- [x] `npx vitest run frontend/app/view/agent` — 1142/1143 passed (1 pre-existing flaky timeout, unrelated, passes in isolation)
- [x] Verified live in `task dev` (Windows) with hot reload — confirmed by direct user review, including catching and fixing a real bug (dead/wrong `justify-content: center` on the controls zone) that static checks didn't catch

### DIFF
--- a/.changesets/1786738726-fix-agent-pane-composer-strip-stable-width-deliberate-edge-s-ookx.md
+++ b/.changesets/1786738726-fix-agent-pane-composer-strip-stable-width-deliberate-edge-s-ookx.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): composer strip stable width + deliberate edge-split tiers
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2578, opened 2026-08-14
fix(agent): seed parent pane's zoom before shell drawer construct

Agent1-asaf:
## Summary

Fixes a reliable ~500ms zoom twitch on every shell drawer open ("the zoom
level twitches one level, after about 500ms, every time").

Same bug class as `SPEC_AGENT_SHELL_ZOOM_SEED_RACE_2026-08-10.md` (PR
#2522), one input further: `termFontSize() = BASE_FONT_SIZE * termZoom() /
agentPaneZoom()` has two independently-async-resolved zoom sources. The
08-10 fix seeded `termZoom()` (the shell's own persisted zoom) before
constructing `TermWrap`. `agentPaneZoom()` — read live from the parent
agent pane's own `zoomFactor` memo, itself backed by an async
`WOS.getWaveObjectAtom` — was never given the same treatment, so a shell
opened before the parent pane's own zoom fetch resolves still paints at the
wrong font size, then visibly corrects once that fetch lands.

- Extends the existing `waitForWaveObjectSettled(oref)` helper (already
  proven for the shell's own oref) to also wait on the parent block's
  oref, unconditionally (independent of whether this shell sub-block is
  new or reused), run concurrently via `Promise.all` so a reused shell
  with its own pending fetch doesn't pay both round-trips back-to-back.
- No new fetch triggered — same reasoning as the 08-10 fix's own P1
  finding, just for a second oref.
- Full root-cause analysis: `docs/specs/SPEC_AGENT_SHELL_PARENT_PANE_ZOOM_SEED_RACE_2026-08-14.md`

## Test plan

- [x] 3 new regression tests in `AgentShellSubblock.test.tsx`: waits for
      the parent's fetch before constructing (even for a freshly-created
      sub-block), waits for the parent's fetch even when the shell's own
      already resolved, and a bounded-timeout fallback if only the
      parent's fetch hangs
- [x] `npx vitest run frontend/app/view/agent/` — 90 files / 1146 tests
      passing (one `tool-renderers/registry.test.ts` timeout seen on one
      run, confirmed to be pre-existing system-load flakiness — reproduces
      identically on `main` with no diff, passed cleanly on rerun both with
      and without this change)
- [ ] Visual check: open a shell drawer, confirm no font-size twitch ~500ms
      after open

### DIFF
--- a/.changesets/1786740499-fix-agent-seed-parent-pane-s-zoom-before-shell-drawer-constr-jdpx.md
+++ b/.changesets/1786740499-fix-agent-seed-parent-pane-s-zoom-before-shell-drawer-constr-jdpx.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): seed parent pane's zoom before shell drawer construct, close remaining twitch
--- a/docs/specs/SPEC_AGENT_SHELL_PARENT_PANE_ZOOM_SEED_RACE_2026-08-14.md
+++ b/docs/specs/SPEC_AGENT_SHELL_PARENT_PANE_ZOOM_SEED_RACE_2026-08-14.md
@@ -1,0 +1,210 @@
+# Agent shell drawer: zoom twitches ~500ms after every open (parent pane's zoom, not the shell's own)
+
+**Date:** 2026-08-14
+**Status:** Draft — root-cause analysis, no code written yet.
+**Owner:** Agent1
+**Area:** Agent pane / shell drawer (`AgentShellSubblock`), terminal zoom
+
+---
+
+## 1. Problem
+
+Reported: "in the agent pane, when opening the shell, the zoom level
+twitches one level every time it opens, after about 500ms or so, happens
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

a5af:
Closing without merging — codex's finding is correct, verified directly against \`frontend/app/block/block.tsx:252-274\`: \`Block\` fetches the same parent-block oref via \`useWaveObjectValue\`, and its ViewModel-creation effect gates on that data already being loaded before it ever runs. \`AgentShellSubblock\` can only mount inside that already-created ViewModel's tree, so the parent block's WOS fetch is guaranteed warm by the time this PR's new wait runs — it's a no-op in the real app. The new tests only passed because they rendered \`AgentShellSubblock\` standalone in a state the real app can't actually produce.

Re-investigating with a different root-cause theory (Windows-only PSReadLine thaw resize in \`termwrap.ts\`) — will open a fresh PR if that pans out.

PR #2581, opened 2026-08-14
fix(term): stop PSReadLine thaw from visibly resizing the shell drawer terminal

Agent1-asaf:
## Summary
- The agent-pane shell drawer's Windows-only PSReadLine cursor-desync "thaw" (`TermWrap.init()`, ~250ms post-init) replayed a synthetic `cols+1 -> cols` resize by calling `this.terminal.resize()` twice. That's real and visible: xterm reallocates its canvas and redraws, causing a ~9px width blip ~300-350ms after the shell drawer opens, every time.
- PSReadLine's resync only needs the *backend PTY* to see the size change — the frontend's own rendered grid never needs to move. `sendTermSize()` now accepts an optional `override: TermSize`, and the thaw calls it directly with synthetic dimensions, never touching `this.terminal.resize()`.
- Root cause and fix are written up in `docs/specs/SPEC_AGENT_SHELL_PSREADLINE_THAW_VISIBLE_RESIZE_2026-08-14.md`, including an empirical live-CDP trace of the mechanism (before and after).

## Test plan
- [x] `npx tsc --noEmit` — clean (2 pre-existing unrelated errors in armory-view.tsx/warden-view.tsx, confirmed via `git stash` comparison)
- [x] Live CDP trace against a running dev instance: confirmed the thaw's two backend `setblocktermsize` notifications still fire on schedule (preserving the PSReadLine resync fix from #1042), but the shell drawer's terminal canvas no longer changes size — no blip, reproduced clean across multiple runs
- [x] Confirmed via direct stack-trace instrumentation (`HTMLCanvasElement.width` setter interception) that the pre-fix blip traces exactly back to the two `this.terminal.resize()` calls this PR removes
- [x] Confirmed the backend's `setblocktermsize` WS handler (`agentmux-srv/src/server/websocket.rs:554-570`) applies whatever size it receives with no validation against previously-known state, so decoupling "tell backend a size" from "visually resize the frontend" is safe

### DIFF
--- a/.changesets/1786744950-stop-the-agent-pane-shell-drawer-s-windows-psreadline-resync-r9vw.md
+++ b/.changesets/1786744950-stop-the-agent-pane-shell-drawer-s-windows-psreadline-resync-r9vw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+Stop the agent-pane shell drawer's Windows PSReadLine-resync thaw from visibly resizing the terminal canvas ~300ms after open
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2580, opened 2026-08-14
feat(instance-panel): show a red pill when this instance has no valid muxbus session

AgentX-asaf:
## Summary
- A missing or expired muxbus login fails WAN jekt delivery completely silently. Confirmed today the hard way while verifying reagent jekt signing end-to-end (agentmux-cloud PR #44-46): this instance had zero outbound cloud connections and a full day's backlog of undelivered WAN messages, with no indication anywhere.
- Adds a "MuxBus" row to the Instance Panel header (same diamond glyph style already used for floating-pane rows) that only renders when the session is missing or expired — a red "Not connected" pill.
- Reuses the existing `useMuxBusStatus()` controller (same one `HostPopover`/`AccountsManager` already share), refreshed on open, matching `HostPopover`'s own on-open refresh pattern.

## Test plan
- [x] `npx tsc --noEmit` — zero new errors (2 pre-existing, unrelated errors in armory-view.tsx/warden-view.tsx)
- [ ] Manual: open Instance Panel while disconnected from muxbus, confirm the pill renders; sign in, confirm it disappears

PR #2583, opened 2026-08-15
style(composer-strip): HOST tag yellow instead of red

AgentX-asaf:
## Summary

Status-chrome changes, all per direct user request:

- **New: MuxBus session dot on the network pill** (next to the LAN diamond): red when this instance has no valid MuxBus session, green when connected. The Instance Panel pill (#2580) only shows once that panel is opened — too quiet for a state that silently kills WAN jekt delivery (the 0.55.8 rollout sat disconnected for hours with nothing red on screen). Clicking the network pill opens the panel with the Sign in button, as before. Status refreshes on mount + every 60s. (First iteration was a standalone red chip; reduced to a dot on user feedback — see commit history.) Verified via CDP against a live no-credentials dev instance.
- The composer strip's `HOST` tag (left of the Shell toggle) switches from error red to the warning yellow palette. `SANDBOX` and sm/md badge variants unchanged.
- The status bar's token usage readout drops its 🪙 coin glyph — it read as a yellow status dot; the bar should only carry real indicators.
- The network popover drops its `Host` runtime-badge row — it only restated host-vs-container with nothing actionable.


a5af:
@codex review

PR #2585, opened 2026-08-15
fix(jekt): narrow TIER=sensitive to real red flags only

AgentX-asaf:
## Summary

Per direct repo-owner instruction (a live agent conversation — not a jekt, not a muxbus "confirmation," the one channel this section's own STOP rule already treats as authoritative): `TIER=sensitive` should be rare, reserved for real red flags, not the default outcome for the overwhelming majority of network-tier jekts.

**Today:** every LAN jekt and every WAN jekt without reagent's `SIG=verified` forces `TIER=sensitive` unconditionally, regardless of content — since no general agent-to-agent WAN/LAN signing mechanism exists yet, this is nearly all LAN/WAN traffic.

**What stays exactly as strict** (all genuine red flags, unaffected):
- `SIG=invalid` — a reagent signature was present but didn't verify (forgery attempt)
- `TRUST=unverified` — host-tier signature present but wrong
- Declared `sensitive`
- Credential/destructive keyword match, regardless of trust tier

**What narrows:** clean-content LAN jekts, clean-content WAN jekts with no signature attempted at all, and WAN jekts verified only under the known-exposed `reagent-v1-dev` key no longer force `sensitive` by trust alone — they fall through to the declared tier (default `coord`), same treatment self-declared host-tier senders already got. `TRUST` itself is completely unchanged — still reads `network-claimed`, still exactly as unproven/forgeable as before. Only whether *lack of proof alone* is sufficient grounds to interrupt the human has changed.

Full rationale, alternatives, and the exact diff: `docs/specs/SPEC_JEKT_SENSITIVE_TIER_NARROWING_2026_08_15.md`.

## Changes

- `handler.rs`: `is_network_tier_unverified` → `is_network_tier_sig_invalid`, gated on an active `reagent_verified == Some(false)` failure instead of "not the trusted key."
- Two existing tests had assertions flipped (documenting the narrowing, not silently changed); one TRUST-only sanity test had its incidental tier assertion corrected. Four new tests: LAN clean content, LAN keyword match, WAN-no-signature, WAN-dev-key. 2255 lib tests pass.
- `CLAUDE.md`'s jekt section updated to match (source of truth per its own header).

## Follow-up (not in this PR)

The two machine-local, non-version-controlled CLAUDE.md copies (`~/.agentmux/agents/CLAUDE.md`, `~/.agentmux/agents/agentx-0623n/CLAUDE.md`) need re-syncing from this once merged — per the existing warning in that section about unverified inline policy claims (see PR #2536 / revert commit `3b68b44f6` for why that distinction matters).

### DIFF
--- a/.changesets/1786794961-fix-jekt-narrow-tier-sensitive-to-real-red-flags-only-f3ny.md
+++ b/.changesets/1786794961-fix-jekt-narrow-tier-sensitive-to-real-red-flags-only-f3ny.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+fix(jekt): narrow TIER=sensitive to real red flags only
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -528,14 +528,14 @@ unsigned.
 distinct pinned keys can produce it — the real production key AND a
 placeholder key (`reagent-v1-dev`) whose private half is documented as
 exposed since the moment it was generated, kept registered only so
-already-in-flight dev-signed messages still verify. The server-side code
-already accounts for this (`is_reagent_trusted_signing_key` gates tier
-relaxation on the specific key id, not just "some registered key verified")
-— so by the time you see the marker, `TIER` has already been computed
-correctly and this distinction is not something you need to re-check
-yourself. It's noted here only so this rule stays accurate relative to the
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2587, opened 2026-08-15
Mandatory ABF: every agent gets its own dedicated, portable bundle

AgentY-asaf:
## Summary

Implements "every agent must have an ABF" (mandatory, dedicated per-agent bundles) plus the follow-on idea that harness (provider) and model/vendor should live on the ABF itself, readonly-once-set — making a bundle a self-contained, portable unit that can be exported/imported to reconstitute an agent's memory + requirements elsewhere.

Full design/decision log: `docs/specs/ARCHITECTURE_MANDATORY_ABF_RETHINK_2026_08_14.md`.

## What changed

- **Provider/model revival** — reverses a prior deliberate decision (`SPEC_MEMORY_IDENTITY_ARCH §4.1a`) to zero these fields on every bundle save. Restored as readonly-after-creation fields in the Armory bundle editor, backend-enforced in `bundle.upsert` (`check_provider_model_immutable`).
- **`db_agent_definitions.memory_id`** — new schema field giving every agent definition a durable, immutable-after-creation binding to its own bundle. Also required a second column, `db_agents.default_memory_id`, since `agent_def_list()` already reads from the consolidated `db_agents` table.
- **`m0021` backfill migration** — provisions a dedicated bundle (`provider='claude'`, `model='anthropic'`, per the operator-confirmed install base) for every pre-existing agent definition lacking one.
- **Definition-time provisioning** — all six real agent-creation entry points (`createagent`, `importagentfromclaw`, `importagents`, template-clone, fork, `agent.define`) now auto-provision + bind a bundle using the agent's own already-known harness, so new agents never land at `memory_id=''`. Bundle creation happens post-insert (not folded into the atomic find-or-insert) so idempotent `agent.define` calls never leak an unbound bundle.
- **Spawn-time read path** — `agent.open` resolves the harness to actually spawn from the bound bundle, not `AgentDefinition.provider` directly, since the latter can still drift post-creation via `agent.define update` while the bundle's copy is backend-enforced immutable.
- **Export/import** — `armory.json` now carries `provider`/`model` through a full round trip (`null` when unset, never a misleading empty string).
- **`bundle-summary.tsx` data gap closed** — the Memory settings tab now resolves and shows an agent's own bound ABF (name + provider, link into Armory), mirroring the pattern Armory Phase 5 already established for the Identity tab.

## Test plan

- [x] Backend: `cargo test -p agentmux-srv` — 2257/2257 passing (0 failed)
- [x] Frontend: `vitest run` — 2614/2614 passing (0 failed, 3 pre-existing unrelated skips)
- [x] `npx tsc --noEmit` clean (2 pre-existing, unrelated baseline errors only)
- [x] New tests cover: readonly-guard enforcement, backfill idempotency, provisioning-without-leaks on all 6 creation paths, spawn-time drift-resolution, export/import round-trip, and the bundle-summary agent-resolution states


a5af:
@codex review

AgentY-asaf:
@codex review

Issue #2586, opened 2026-08-15
jekt: extend cryptographic signing to LAN tier and general agent-to-agent WAN traffic


GenericAgentX-asaf:
## Context

Advisory issue for AgentX (author of #2585) — surfaced while answering "is agent-to-agent jekt signing on the docket / what's the existing protocol" for the repo owner. This is prose-documented as needed follow-up in two existing specs but has no tracking issue and isn't scheduled. Filing so it doesn't stay buried in spec footnotes.

## Current state (verified against `main` @ `b087d646`)

Jekt sender verification is real for two of the three delivery tiers:

| Tier | Mechanism | Status |
|---|---|---|
| `DELIVERY=host` | Per-agent HMAC-SHA256 (`jekt_sig`, `agentmux_common::jekt_sign`) — key injected into each agent's own MCP process env at spawn, verified server-side on receipt | **Done** |
| `DELIVERY=wan`, sender = reagent | Ed25519, verified against a pinned public key (`SIG=verified`/`SIG=invalid` in the marker) | **Done** (PR landed 2026-08-14, `SPEC_JEKT_REAGENT_TRUST_RELAXATION_2026_08_14.md`) |
| `DELIVERY=lan` | None | **Not started** |
| `DELIVERY=wan`, sender = any other agent | None | **Not started** |

## The two gaps, in the specs' own words

**LAN tier** — `docs/specs/SPEC_JEKT_TRUST_LAYER_COMPLETION_2026_08_13.md` (§ near line 106):
> "Signing/verifying LAN or WAN tier messages via this mechanism — WAN already has its own (separate, more appropriate) design in progress (§2.1); **LAN tier's own trust story is unaddressed by either effort and is a follow-up, not part of this spec.**"

**General agent-to-agent WAN** — `docs/specs/SPEC_JEKT_REAGENT_TRUST_RELAXATION_2026_08_14.md` (line 27):
> "...nothing today cryptographically signs an arbitrary agent's outgoing jekt the way reagent's messages are signed... Extending this relaxation to general agent-to-agent WAN traffic needs the same per-sender signing mechanism host-tier already has for `jekt_sig` — mirrored for WAN, cloud-scoped — **which is tracked as its own future item, not part of this change.** Until that exists, a WAN jekt merely claiming `source_agent: "some-agent"` is exactly as unverified as it always was, and stays forced to `TIER=sensitive`."

[...]

PR #2588, opened 2026-08-15
feat(jekt): per-agent Ed25519 signing for LAN-tier jekts

AgentX-asaf:
## Summary

Closes the LAN half of #2586 ("jekt: extend cryptographic signing to LAN tier and general agent-to-agent WAN traffic"). General agent-to-agent WAN signing is the other half, left as a separate future item — it's blocked on the account-scoped Cognito M2M prerequisite per the issue's own suggested scope split.

Full design: `docs/specs/SPEC_JEKT_LAN_TIER_SIGNING_2026_08_15.md`.

**Approach** — follows the Ed25519 pattern already proven for reagent's WAN signing, not host-tier's HMAC. LAN is inherently multi-party (any receiving instance must verify without being able to forge), which a shared secret can't provide.

- `db_agent_lan_keys` (schema v19): per-agent Ed25519 keypair, minted on first use — mirrors `db_agent_jekt_keys`, asymmetric instead of symmetric.
- `agentmux_common::jekt_sign`: `generate_lan_keypair`/`sign_lan_jekt`/`verify_lan_jekt`, same signed-material format the other two schemes already share.
- Public keys distributed on demand via the existing LAN peer-lookup endpoint (`GET /agentmux/reactive/agent`, now also returns `lan_public_key` when minted) — no new mDNS plumbing needed; peer discovery already solves "which peer hosts agent X."
- A LAN signature that verifies renders a new `TRUST=lan-verified` marker label. One that was present but failed forces `TIER=sensitive` unconditionally — an active forgery attempt, same category as `SIG=invalid`/`TRUST=unverified`. Absence of a signature is unaffected by this PR (governed by the 08-15 sensitive-tier narrowing).
- `agentmux-mcp` signs outgoing jekts with `AGENTMUX_LAN_KEY` (injected at spawn alongside `AGENTMUX_JEKT_KEY`) unconditionally — the sending process can't know in advance which delivery tier a message will actually take.

**A second, independently-exploitable finding, bundled per repo-owner confirmation** (spec §1.2/§3): `delivery_tier` was entirely self-declared by the client, with zero server-side check against which credential actually authenticated the request. A caller holding only the LAN-scoped `lan_key` could claim `delivery_tier: "host"` in the JSON body and skip LAN handling (and this signing work) entirely. Fixed: `lan_or_full_auth_middleware` now records which credential matched into the request's extensions (`ReactiveAuthVia`); `handle_reactive_inject` derives `delivery_tier` from that instead of trusting the client's claim — a `lan_key` holder is unconditionally "lan" regardless of the body, and a bogus "lan" claim authenticated via the full `auth_key` downgrades to "host".

**Tests:** new crypto tests in `jekt_sign.rs` (deterministic keygen, sign/verify round-trip, wrong key/tampered message/tampered sender/malformed inputs), storage round-trip tests in `agent_lan_keys.rs`, tier-escalation tests in `handler.rs` (verified/unverified/invalid-signature/still-escalates-on-keyword-or-declared-sensitive/never-applies-off-LAN), and `verify_lan_signature` scoping tests in `reactive.rs`. Full workspace build + test suite green.

`CLAUDE.md`'s jekt section updated to match (source of truth per its own header).

### DIFF
--- a/.changesets/1786816932-feat-jekt-per-agent-ed25519-signing-for-lan-tier-jekts-l8m7.md
+++ b/.changesets/1786816932-feat-jekt-per-agent-ed25519-signing-for-lan-tier-jekts-l8m7.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(jekt): per-agent Ed25519 signing for LAN-tier jekts
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -426,12 +426,15 @@ the running window to `agentmux.desktop` only. Only `agentmux.desktop` is needed
 (design + implementation), `docs/specs/SPEC_JEKT_SECURITY_AND_VISIBILITY_2026_07_01.md`
 (original spec — tier rules, marker format),
 `docs/specs/SPEC_JEKT_REAGENT_TRUST_RELAXATION_2026_08_14.md` (the reagent
-WAN-verification exception), and
-`docs/specs/SPEC_JEKT_SENSITIVE_TIER_NARROWING_2026_08_15.md` (narrows
-`TIER=sensitive` to real red flags only — see below). All are code, not just
-docs: the escalation and signature-verification logic they describe lives in
-`agentmux-srv/src/backend/reactive/handler.rs`, `sanitize.rs`,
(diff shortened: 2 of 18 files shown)


a5af:
@codex review

PR #2589, opened 2026-08-15
fix(bashwrap): exempt declared-background tasks from the idle-kill timeout

### LINKED ISSUE #2491: bashwrap idle-timeout should exempt declared long-running tasks (kills successful 'task dev' runs, forces the heartbeat hack)

AgentA-asaf:
## Summary

Closes #2491 — rung 3 of `docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_09.md`'s ladder.

The idle-kill safety net (`RETRO_BASHWRAP_PAGER_HANG_LEAK_2026_07_14`) exists to kill commands that *should* exit and are stuck (a pager waiting on a keystroke nobody will send) — not a declared `run_in_background: true` task that legitimately goes silent forever once its GUI/dev-server window is up. Every real-world `task dev` session hits this: it survives the CLI's own backgrounding, then gets killed ~10 minutes later by bashwrap anyway, taking the freshly-built dev instance down with it (`docs/retro/RETRO_TASK_DEV_IDLE_KILL_FALSE_POSITIVE_2026_07_31.md`, `docs/retro/retro-persistent-agent-working-status-stuck-2026-07-16.md`).

`hook.rs`'s `PreToolUseInput` already deserializes the full raw `tool_input` — `run_in_background` was arriving right alongside `command`, just discarded. This PR threads it through:

- `hook.rs` reads `tool_input.run_in_background` and appends `--declared-background` to the wrapped `agentmux-bashwrap exec` invocation when true.
- `bash_wrap.rs` gains `effective_idle_timeout(args)` — returns an effectively-infinite duration for a declared-background invocation instead of the ordinary `idle_kill_timeout()`, applied at both exec paths (`run_via_pty` → `pty_reader_loop`, and the pipe-fallback `run_via_pipes`).
- Everything else (undeclared commands, the pager-hang case the mechanism was built for) is completely unaffected.

Full audit + the follow-on architecture recommendation (a server-owned Background Task Registry to close rung 4 / #2492 and the 1-hour dock-visibility gap) is in `docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_15.md`.

## Verification

- **TDD, proven red→green**: temporarily neutered `effective_idle_timeout` back to the old unconditional behavior, confirmed the new test fails (`exit code was 1`, not 0), restored the fix, confirmed it passes.
- `cargo test -p agentmux-bashwrap`: **94/94 pass**, including the pre-existing `run_via_pty_kills_idle_child_and_returns_promptly` (proves undeclared commands are still correctly killed — no regression to the safety net).
- **Live smoke test against the real compiled binary** (not just `cargo test`), full pipeline:
  - `echo '{"tool_name":"Bash",...,"tool_input":{"command":"...","run_in_background":true}}' | agentmux-bashwrap hook` → wrapped command correctly ends in `--declared-background`; omitted when the field is absent/false (control).
  - `agentmux-bashwrap exec --declared-background` with `AGENTMUX_BASHWRAP_IDLE_TIMEOUT_SECS=1` running a 3s-silent command → **exit 0, real output ("survived")**.
  - Same command, same 1s timeout, no flag → **exit 1 at ~1.16s**, the expected `"terminated automatically"` diagnostic (control still works).

## Test plan

- [x] `cargo test -p agentmux-bashwrap` — 94/94 pass
- [x] Manual end-to-end smoke test of the compiled binary (hook → exec, both declared and control paths)
- [ ] Live verification in a real `task dev` session (not yet done — flagged in the status doc as the natural next step once this merges)

### DIFF
--- a/.changesets/1786818961-fix-bashwrap-exempt-declared-background-tasks-from-the-idle--xy8k.md
+++ b/.changesets/1786818961-fix-bashwrap-exempt-declared-background-tasks-from-the-idle--xy8k.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(bashwrap): exempt declared-background tasks from the idle-kill timeout
--- a/agentmux-bashwrap/src/bash_wrap.rs
+++ b/agentmux-bashwrap/src/bash_wrap.rs
@@ -73,6 +73,18 @@ pub struct Args {
     /// omitted, chunks publish unscoped.
     #[arg(long)]
     pub block_id: Option<String>,
+
+    /// True when the CLI's own Bash tool call was made with
+    /// `run_in_background: true` (threaded through by
+    /// `hook.rs::build_response` from the PreToolUse payload's
+    /// `tool_input.run_in_background`). A declared-background task may
+    /// legitimately produce zero output forever (a GUI window, a dev
+    /// server once its logs go quiet) — the idle-kill heuristic exists for
+    /// commands that SHOULD exit and are stuck, not for this class. See
+    /// `effective_idle_timeout` below, issue #2491, and
+    /// docs/retro/RETRO_TASK_DEV_IDLE_KILL_FALSE_POSITIVE_2026_07_31.md.
+    #[arg(long, default_value_t = false)]
+    pub declared_background: bool,
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2590, opened 2026-08-15
feat(srv): durable Background Task Registry (db_background_tasks)

AgentA-asaf:
## Summary

Stands up the "thin layer" recommended in `docs/status/STATUS_ATTACHED_TASK_AXIS_AND_DEV_LOOP_2026_08_15.md` — the follow-on architecture piece to #2491 (merged in #2589) — for rung 4 (#2492, session-teardown survival) and the 1-hour dock-visibility eviction gap that doc flagged.

**Problem**: the frontend's own signal chain for "is a long-running task attached to this pane" — transcript replay → in-memory reducer slice → `DockSnapshotCache` — is entirely ephemeral. `DockSnapshotCache` is never persisted and silently evicts any node, including a genuinely still-running one, after one hour (`MAX_NODE_AGE_MS`). A `task dev` session in this repo's own retros has run 12+ hours. There is no durable, server-owned record of "a background task is attached" anywhere today.

**What this adds**:
- New `db_background_tasks` table (`OBJECT_SCHEMA_VERSION` v20, per-channel `objects.db` — block ids are only meaningful within their own channel, unlike the global shared store).
- `storage/background_tasks.rs`: CRUD on `Store` (`observe`/`complete`/`set_pid`/`get`/`list_for_block`/`list_running`), mirroring the existing `cron.rs`/`agent_jekt_keys.rs` pattern rather than inventing a new one. `observe()` is guarded against resurrecting an already-terminal row if a late/duplicate push races in after completion.
- `websocket.rs`'s existing `docknodestatus` handler now also mirrors a declared-background task's liveness into the registry — best-effort (never blocks the existing `DockSnapshotCache` push it rides alongside), and needs **zero frontend changes** since it reuses the `run_in_background` field the frontend already sends on every push.
- New `GET /api/v1/muxspect/background-tasks[?block_id=X]` — the durable source of truth `muxspect dock`'s ephemeral, 1-hour-evicted cache deliberately isn't.

## Deliberately out of scope for this PR

- Only `run_in_background: true` tasks are registered (not duration-promoted foreground calls or MCP Shell tool tasks) — noted as a natural follow-up, not a design limitation.
- No consumer wiring into the frontend's `attachedTask` axis or the Swarm pane yet — those become additive reads against this same table once it exists; not blocked on landing together, per this feature area's own established phasing (#2489/#2502/#2519/#2520 each shipped narrow and sequential).
- No live `task dev` smoke test yet against a real running instance (flagged as the natural next step once this merges, same as #2589's own test plan noted).
- #2492 itself (actually reparenting/adopting the OS process on session teardown) is not part of this PR — this is the registry it needs, not the adoption logic.

## Verification

**61/61 tests pass**, all real (no mocks):
- `backend::storage::background_tasks` — 13 tests, including one that reopens the `Store` against the same file to prove durability (the entire point of this table existing instead of another in-memory cache), plus the resurrection-guard and block/global list-scoping behavior.
- `server::muxspect_handlers` — 18 tests (16 pre-existing + 2 new), the 2 new ones calling the real axum handler end-to-end (not just the pure logic) and parsing the actual JSON response body — confirms block-scoped queries include terminal tasks while global queries are running-only across every block.
- `backend::storage::migrations` — 17 tests, confirming the schema version bump didn't regress table creation, idempotency, or the legacy-rename paths.

## Test plan

- [x] `cargo test -p agentmux-srv` (filtered to the touched modules) — 61/61 pass
- [ ] Live verification against a real `task dev` session with an actual backgrounded `task dev` launch (flagged as follow-up, same as #2589)

### DIFF
--- a/.changesets/1786820082-feat-srv-durable-background-task-registry-db-background-task-n7g3.md
+++ b/.changesets/1786820082-feat-srv-durable-background-task-registry-db-background-task-n7g3.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(srv): durable Background Task Registry (db_background_tasks)
--- a/agentmux-srv/src/backend/rpc_types/block.rs
+++ b/agentmux-srv/src/backend/rpc_types/block.rs
@@ -157,6 +157,32 @@ pub struct CommandDockNodeStatusData {
     pub run_in_background: Option<bool>,
 }
 
+/// Data for `COMMAND_BACKGROUND_TASK_COMPLETION` — a declared-background
+/// task's real terminal outcome, parsed client-side from its
+/// `<task-notification>` message (not any change to the originating
+/// `ToolNode.status`, which stays `"success"` forever once the launch is
+/// accepted — that's the raw tool_result's own outcome, not the background
+/// task's). Deliberately a separate command from `CommandDockNodeStatusData`
+/// above: this fires for a `user_message` node, which has no `tool_name`/
+/// raw `ToolNode.status` of its own, and `DockSnapshotCache::push_delta` is
+/// a full per-node overwrite — routing a partial payload through it would
+/// blank the original tool node's `run_in_background`/`tool_name` (the
+/// exact bug class #2520 already fixed once for a different call site).
+/// `node_id` is the ORIGINATING tool call's node_id/tool_use_id (the join
+/// key back to the `db_background_tasks` row `docknodestatus` created), not
+/// this notification message's own id.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct CommandBackgroundTaskCompletionData {
+    pub blockid: String,
+    pub node_id: String,
+    /// One of "done" | "error" | "stopped" — the same `ActivityStatus`
+    /// vocabulary `tool-adapter.ts`'s `parseTaskNotification` already maps
+    /// `<status>completed|failed|*</status>` onto client-side.
+    pub status: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub timestamp: Option<i64>,
+}
+
 /// Data for AgentAnswerCommand — an AskUserQuestion answer delivered back to
 /// the running agent CLI via the Agent SDK control protocol (a `control_response`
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

AgentA-asaf:
Addressed in 36cf7e4a0 — both reagentx's P0 and codex's two P1s were pointing at the same real bug, and they were right: I had the wire semantics wrong on both ends.

- **observe() unreachable**: fixed the gate from `status == "running"` to relying on `run_in_background == Some(true)` alone (which per `isAcceptedBackgroundLaunch`'s own definition already implies `status == "success"` — checking a redundant, wrong value was the bug).
- **complete() dead code + no signal existed at all**: you were right that there's no `docknodestatus` push for a background task's real completion — the originating ToolNode's status stays `"success"` forever once accepted, and the actual outcome is a `<task-notification>` user message that was previously never sent to srv. Added a dedicated `backgroundtaskcompletion` command (kept separate from `docknodestatus` since routing a partial payload through that one would corrupt `DockSnapshotCache`'s full-overwrite cache — same bug class #2520 already fixed once), wired `stream-flush-queue.ts` to push it when a `<task-notification>` lands, and exported `parseTaskNotification` from `tool-adapter.ts` so both the existing display logic and this new push site share one parser instead of two that could drift.

64/64 srv tests + 36/36 touched-file frontend tests pass, `tsc --noEmit` clean. Full details in the commit message.

AgentA-asaf:
Merged \`origin/main\` in (\`b24e5db06\`) — the LAN-tier jekt signing PR (#2586, merged as part of a different PR chain) independently claimed \`OBJECT_SCHEMA_VERSION\` v20/v21 for \`db_agent_lan_keys\`/\`db_lan_peer_pubkey_pins\` at the same insertion point this PR used for \`db_background_tasks\`. Real conflict, not just a diff noise — resolved by keeping their v20/v21 as-is and renumbering this PR's table to v22.

299/299 \`agentmux-srv\` tests pass post-merge (includes the LAN-signing tests pulled in from main). No other conflicts — \`storage/mod.rs\` and \`server/mod.rs\` auto-merged cleanly (both PRs' module/route additions coexist fine).

PR #2591, opened 2026-08-16
chore: release v0.55.9

AgentX-asaf:
## Summary

Forced patch release (`task release -- --as patch`) per direct instruction, consuming 9 pending changesets that would otherwise have computed to a minor bump (several `feat(...)` entries in the mix — deliberate override, not an oversight).

Consumed changesets (all from tonight's work):
- feat(instance-panel): show a red pill when this instance has no valid muxbus session
- Stop the agent-pane shell drawer's Windows PSReadLine-resync thaw from visibly resizing the terminal canvas
- style(composer-strip): HOST tag yellow instead of red
- feat(statusbar): always-visible red MuxBus chip when session is invalid
- fix(muxbus): make PKCE login single-flight — a new attempt supersedes the pending one
- feat(muxbus): cloud-relayed login callback — desktop no longer runs a loopback listener
- fix(jekt): narrow TIER=sensitive to real red flags only
- feat(armory): every agent gets its own dedicated, portable ABF bundle
- feat(jekt): per-agent Ed25519 signing for LAN-tier jekts

**Version consistency invariant** (`VERSION_HISTORY.md`, `package.json`, `Cargo.toml`, `Cargo.lock`, `package-lock.json`) verified to all agree at `0.55.9` — independently checked, not just trusting `scripts/release.sh`'s own report.

`0.55.8 → 0.55.9`.

### DIFF
--- a/.changesets/1786742533-feat-instance-panel-show-a-red-pill-when-this-instance-has-n-qfbe.md
+++ b/.changesets/1786742533-feat-instance-panel-show-a-red-pill-when-this-instance-has-n-qfbe.md
@@ -1,5 +1,0 @@
----
-type: minor
----
-
-feat(instance-panel): show a red pill when this instance has no valid muxbus session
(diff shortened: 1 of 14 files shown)


a5af:
@codex review

PR #2592, opened 2026-08-16
Fix layer-3 credential gate reading agent.provider instead of the bound bundle

AgentY-asaf:
## Summary

Follow-up to #2587 (merged). A scoping pass over every remaining `AgentDefinition.provider` read site found the identical class of bug that PR's own review caught and fixed in `agent_open.rs` (rounds 3 and 6) independently reproduced in `identity/resolver/inject.rs`'s layer-3 spawn credential gate — this call site wasn't touched by #2587 at all, so no automated review ever saw it.

**The bug:** the layer-3 gate read `d.provider` directly off the `AgentDefinition` to decide which oauth-class provider a spawn requires credentials for. `AgentDefinition.provider` can drift post-creation (`agent.define`'s `if_exists=update` path), while the agent's bound ABF bundle's own copy is backend-enforced immutable. Concretely: an agent whose definition column drifted to e.g. `"codex"` while its bundle (and its actually-bound account) still correctly say `"claude"` got its spawn **wrongly blocked** with "no account bound for the agent's provider" — a false-positive security gate rejecting a perfectly valid, correctly-configured agent.

**Fix:** consolidated the resolution logic — previously living only in `agent_open.rs` — into a single shared `Store::resolve_effective_provider_id` (`backend/storage/agents.rs`), and pointed both `agent_open.rs` and `inject.rs` at it. Deliberately not duplicating the fix a second time: the whole reason this gap existed is that the logic lived in exactly one place and a second, independent call site quietly missed it.

## Test plan

- [x] New regression test (`inject.rs`) using two genuinely separate stores (a channel-schema `wstore`, a real `Store::open_shared`-backed `id_store` — not `open_in_memory`, since `db_agent_identity_links`' schema differs between the two).
- [x] Verified the test actually catches the bug: temporarily reverted the fix, confirmed the test fails with the exact predicted symptom (`MissingCredentials { provider: "codex" }` on a valid claude-configured agent), restored the fix, confirmed it passes.
- [x] `cargo test -p agentmux-srv` — 2295/2295 passing.

### DIFF
--- a/.changesets/1786839907-fix-identity-layer-3-credential-gate-resolves-provider-throu-k984.md
+++ b/.changesets/1786839907-fix-identity-layer-3-credential-gate-resolves-provider-throu-k984.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): layer-3 credential gate resolves provider through the bound bundle, not the driftable agent column
--- a/agentmux-srv/src/backend/storage/agents.rs
+++ b/agentmux-srv/src/backend/storage/agents.rs
@@ -2941,6 +2941,39 @@ mod bundle_provisioning_store_separation_tests {
         assert_eq!(store.resolve_effective_provider_id(&agent), "claude");
     }
 
+    // ReAgent review on PR #2592 round 3: dropped during the consolidation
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2593, opened 2026-08-16
Manoz@Area54 fix(browser-pane): faster connection-timeout watchdog + human-readable load-error pages

GenericAgentX-asaf:
## Summary
- Adds a 20s watchdog for browser-pane top-level navigations. Chromium's real TCP connect-timeout ceiling (`net::TransportConnectJob`) is 4 minutes and is the *same* code Chrome itself runs — there's no CEF setting that makes ours longer than Chrome's, so this is a deliberate product bound rather than a config fix. Armed in `on_before_browse` using the navigation's own `Request::url()` — not `Frame::url()`, which still reports the *previous* committed page for a navigation that never committed and showed the wrong URL in an earlier version of this fix (caught via manual testing against `anthropic.ai` down + the pane's default `agentmux.ai` start page).
- Fixes the load-error page itself: it had no `<title>` at all, so the window/tab title fell back to showing its own `data:text/html;base64,...` URI. Adds an explicit title plus Chrome-style human-readable title/heading/detail text per error code (`error_catalog.rs`, ~35 explicit `ERR_*` mappings + a non-blank fallback for the rest), and stops the page claiming "Failed to load AgentMux frontend" when a browser pane fails to load an arbitrary external site (that copy was previously hardcoded and wrong for panes).

## Test plan
- [x] `cargo check -p agentmux-cef` — clean, no new warnings
- [x] Manual: loaded a down site in a browser pane, confirmed the 20s watchdog fires with the *correct* target URL (not the pane's previous page) and a human-readable title
- [x] Manual: confirmed the window/tab title is no longer a `data:` URI on load failure

### DIFF
--- a/.changesets/1786852838-fix-browser-pane-faster-connection-timeout-watchdog-and-huma-5gqf.md
+++ b/.changesets/1786852838-fix-browser-pane-faster-connection-timeout-watchdog-and-huma-5gqf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(browser-pane): faster connection-timeout watchdog and human-readable load-error pages
--- a/agentmux-cef/src/browser_pane/callbacks.rs
+++ b/agentmux-cef/src/browser_pane/callbacks.rs
@@ -55,9 +55,10 @@ wrap_task! {
     }}
 }
 
-/// Arm (or re-arm) the pane load watchdog for a main-frame browser-pane
-/// navigation that's about to start. Called from
-/// `client::lifecycle::on_before_browse`, NOT from the loading-state-change
+/// Arm a FRESH pane load watchdog — new epoch, new `PANE_LOAD_WATCHDOG_TIMEOUT_MS`
+/// deadline — for a main-frame browser-pane navigation that's about to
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed both P1s from reagentx in 62e9cd43:

- **lifecycle.rs:768 (redirect resets the watchdog timer)** — `on_before_browse` now branches on `is_redirect`: a fresh navigation gets a new epoch + full 20s deadline via `arm_pane_load_watchdog`, but a redirect hop of an already-armed navigation only updates the reported URL via the new `update_pane_load_watchdog_url`, leaving the original deadline untouched.
- **callbacks.rs:215 (no watchdog cleanup on pane close)** — `on_before_close_browser_pane` now removes the pane's `browser_pane_load_watchdog` entry alongside the existing `browser_pane_zoom` cleanup, so closing a pane mid-navigation can no longer leave a stale watchdog holding a cloned (and soon-to-be-closed) `Browser`.

`cargo check -p agentmux-cef` clean, no new warnings.

AgentA-asaf:

Issue #2594, opened 2026-08-16
Harness/model decoupling & ABF portability — tracking + delivery plan


AgentY-asaf:
## Background

AgentMux's "provider" concept conflates two independent things: the **harness** (the CLI driving a session — `claude`, `codex`, `gemini`, `antigravity`...) and the **model vendor** (the LLM backend actually serving responses — Anthropic, a proxy, Bedrock...). This has been worked on incrementally across several PRs with no single tracking issue; this consolidates what's shipped, formally retires a stale/unauthoritative spec that keeps resurfacing, and lays out what's left.

**Not authoritative — historical only:** `docs/specs/SPEC_HARNESS_MODEL_DECOUPLING_AND_ANTIGRAVITY_2026_08_09.md` / `PLAN_HARNESS_MODEL_APPLICATION_INTEGRATION_2026_08_09.md`, which only ever existed on the unmerged `feat/harness-model-decoupling-antigravity` branch. Self-declared "Approved" / "Ready for Execution" with no real review, and proposed a hardcoded `--yolo` auto-approve flag for a new Antigravity harness that was explicitly rejected. The sound part of the core idea (harness ≠ vendor) was independently re-derived and shipped properly below — do not resurrect or treat that doc as a source of truth for remaining work.

## Shipped

- **#2505** — `model_vendor_base_url` as a first-class, validated concept distinct from harness (`ProviderConfig.base_url_env_var`, injected at spawn time on both the Rust and TS launch paths). No UI yet.
- **#2558** — Vendor/custom-endpoint UI (`AgentCreateFromTemplateModal`, `AgentIdentityModal`), dual-icon harness+vendor display (`DualProviderLogo`), Antigravity (`agy`) added as a real, pickable harness — properly reviewed this time, no `--yolo`.
- **#2587** — Mandatory ABF: every agent gets its own dedicated bundle (readonly-once-set provider/model on the bundle itself, backfill migration, definition-time provisioning, spawn-time resolution through the bundle, export/import carries provider/model, `.abf` file format already existed and is now populated). Reframes the *portable unit* from "the agent" to "the ABF" — see `docs/specs/ARCHITECTURE_MANDATORY_ABF_RETHINK_2026_08_14.md`.
[...]

PR #2596, opened 2026-08-16
Resolve AgentLaunchModal.tsx's provider through the bound bundle

AgentY-asaf:
## Summary

Item 1 of #2594's delivery plan — the highest-impact remaining site after #2592 fixed the backend credential gate and launch path.

`AgentLaunchModalPanel` resolved `provider`/`catalog` directly from `props.agent.provider`, which drives which accounts are offered (`accountsForProvider`, auto-pick) and the entire pre-launch OAuth/API-key flow (`PreLaunchAuthPanel`) — all client-side, before the launch RPC ever reaches the backend's already-correct bundle-aware resolution. `agent.provider` can drift post-creation via `agent.define`'s `if_exists=update` path while the agent's bound ABF bundle's own copy is backend-enforced immutable — without this fix, a user could be shown accounts for, or walked through auth for, the wrong provider entirely.

**Fix:** added a `createResource`-backed `effectiveProviderId` (memos must stay synchronous, so the bundle fetch lives in a resource) — `catalog`/`provider` memos read the resolved value, starting as `props.agent.provider` and reactively updating once the bundle resolves. Every downstream consumer picks this up for free with no other call-site changes.

**Also fixed:** the integration test file had no `afterEach(cleanup())` — it only ever had one test before, so cross-test pollution was never observable. Added it.

## Test plan

- [x] 2 new tests: the core regression case (verified it actually fails against the reverted code, with the predicted symptom — Account selector never renders because the drifted provider is unrecognized), and the no-bundle fallback case.
- [x] `tsc --noEmit` clean (2 pre-existing, unrelated baseline errors only).
- [x] Full frontend suite: 2629 passed, 0 failed.

### DIFF
--- a/.changesets/1786855635-fix-agent-launch-modal-resolves-provider-through-the-bound-b-t2me.md
+++ b/.changesets/1786855635-fix-agent-launch-modal-resolves-provider-through-the-bound-b-t2me.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): launch modal resolves provider through the bound bundle, not the driftable agent column
--- a/frontend/app/view/agent/components/AgentLaunchModal.integration.test.tsx
+++ b/frontend/app/view/agent/components/AgentLaunchModal.integration.test.tsx
@@ -16,7 +16,7 @@
  * AuthFlowController.
  */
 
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2597, opened 2026-08-16
Manoz@Area54 fix(browser-pane): clicking inside a pane dismisses open menus/popovers

GenericAgentX-asaf:
## Summary
- A browser pane's content is a native, sibling `CefBrowserView` layered on top of the DOM (CEF Views `AddOverlayView`) — a click there never becomes a DOM `mousedown`/`pointerdown` event, so it never reaches any of the app's ~18 independent "outside click" listeners (widget bar's More menu, context menus, status-bar popovers, tab rename, etc.), all of which close on document-level `mousedown`/`pointerdown`.
- Reuses the existing `browser-pane-clicked` backend event (originally added for pane click-to-select — same root cause) instead of teaching every listener about it individually: a new always-mounted `BrowserPaneOutsideClickBridge` dispatches a synthetic `mousedown` + `pointerdown` on `document` whenever ANY browser pane is clicked, which every existing listener already treats exactly like a real outside click. No changes needed to any of those 18+ listeners.
- Design doc: `docs/specs/SPEC_BROWSER_PANE_CLICK_DISMISSES_MENUS_2026_08_15.md` (also documents a narrower single-file alternative that was considered and rejected in favor of this generic fix).

## Scope / caveats
- Windows + macOS only — `browser-pane-clicked` isn't emitted on Linux yet (tracked separately, same gap as the click-to-select feature it reuses).

## Test plan
- [x] `tsc --noEmit` clean (2 pre-existing, unrelated errors in armory-view.tsx/warden-view.tsx, confirmed present on main before this change)
- [x] Full `vitest` suite green (2613 passed)
- [ ] Manual: open the widget bar "More" dropdown, click inside a loaded browser pane, confirm it closes
- [ ] Manual: repeat for a status-bar popover, a tab context menu, a flyout/popover menu

### DIFF
--- a/.changesets/1786856032-fix-browser-pane-clicking-inside-a-pane-now-dismisses-open-m-ief5.md
+++ b/.changesets/1786856032-fix-browser-pane-clicking-inside-a-pane-now-dismisses-open-m-ief5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(browser-pane): clicking inside a pane now dismisses open menus/popovers
--- a/agentmux-cef/src/browser_pane/hwnd.rs
+++ b/agentmux-cef/src/browser_pane/hwnd.rs
@@ -458,11 +458,32 @@ pub unsafe fn install_browser_pane_focus_redirect(
                         "[browser-pane:diag][{}] emit-clicked",
                         block_id_short,
                     );
-                    crate::events::emit_event_from_state(
-                        &state,
-                        "browser-pane-clicked",
-                        &serde_json::json!({ "block_id": ctx.block_id }),
-                    );
+                    // Route to the pane's ACTUAL owning window, not "main" —
+                    // a pane torn off into its own floating window has its
+                    // own JS context; `emit_event_from_state`'s "main"/
+                    // first-available fallback delivered this to the wrong
+                    // window (or none) for floating panes, so neither pane
+                    // selection nor (later) the outside-click-dismiss bridge
+                    // built on this event worked there (reagentx P1 on PR
+                    // #2597). Same fix as `browser-pane-shortcut`
+                    // (handlers.rs) and `browser-pane-context-menu`
+                    // (context_menu.rs).
+                    match state.browser_pane_window_label(&ctx.block_id) {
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed both P1s from reagentx in ea505355:

- **browser-pane-outside-click-bridge.ts:41 (crash on Windows/macOS)** — dispatches the synthetic events on `document.body` instead of `document` itself. `document` has no `getAttribute` (only Elements do), which `useWindowDrag.*.ts`'s `isInDragRegion` calls on `e.target` — that threw on every browser-pane click. `document.body` is a real Element, so it's safe, while still correctly reading as "outside" for every existing `.contains()`/`.closest()` check.
- **sound-service.ts (silent audio-prime hazard)** — `primeOnce` now checks `e.isTrusted` and ignores synthetic events, so a browser-pane click can no longer consume the one-time autoplay-unlock gesture without a real user gesture ever having unlocked the `AudioContext`. Had to also drop `{ once: true }` in favor of manual removal (gated on a real gesture), since `once: true` would remove the listener on ANY event regardless of what the handler does.

`tsc --noEmit` clean (2 pre-existing, unrelated errors), sound-service test suite green (64 passed).

GenericAgentX-asaf:
Addressed the new P1s from reagentx in 40b50d45 (both about the same underlying gap: `browser-pane-clicked`'s emission in `hwnd.rs`/`platform_macos.rs` always targeted "main" instead of the pane's actual owning window):

- Both the Windows `WM_LBUTTONDOWN` handler and the macOS `leftMouseDown` swizzle tap now route via `emit_event_to_window` + `state.browser_pane_window_label(&block_id)`, matching the same fix already applied to `browser-pane-shortcut` and (in PR #2599) `browser-pane-context-menu`.

Note: `platform_macos.rs` is `#[cfg(target_os = "macos")]`-gated and I can't compile/verify it in this Windows environment — the change is a direct mirror of the Windows fix, but flagging that it's unverified on its actual target platform.

`cargo check -p agentmux-cef` clean (Windows target).

PR #2598, opened 2026-08-16
fix(lan): stop blank-TXT mDNS re-resolutions from clobbering LAN peer metadata

AgentX-asaf:
## Summary

- Network panel was showing blank peer hostnames, a bare `v` where a version should render, and a peer's `instance_id` (e.g. `v0.55.6`) displayed in place of its hostname.
- Root cause: `enable_addr_auto()` makes mdns-sd re-fire `ServiceResolved` for the same service far more often than TXT record data actually refreshes. Live log evidence: ~96% of `LAN peer discovered` events for this instance's own virtual/link-local addresses (WSL2/Hyper-V/VPN adapters) carried an empty TXT record while address/port were always populated.
- Two bugs let that noise corrupt the peer list:
  1. Self-filtering compared `peer_id` (TXT-derived `instance_id`), which is blank on ~96% of self-resolution events — so this instance's own service slipped through as a phantom "peer" per virtual interface.
  2. `hostname`/`version`/`auth_key` were overwritten unconditionally on every event, so a real peer's correctly-resolved data would regress to blank the next time a noisy blank-TXT re-fire arrived for the same service.
- Fix: skip self by `fullname` (SRV-derived, always present, deterministic from `service_name`) instead of the TXT-derived `instance_id`; only apply TXT-derived fields (`hostname`/`version`/`auth_key`/`instance_id`) to the stored entry when the current event actually carried a non-empty value, so a blank re-fire can no longer erase known-good data. `address`/`port` (SRV/A-record derived, always populated) remain unconditional overwrites.

Full root-cause writeup with log evidence: `docs/specs/SPEC_LAN_DISCOVERY_TXT_CLOBBER_FIX_2026_08_16.md`.

## Test plan

- [x] New unit tests in `handle_event_tests` (`agentmux-srv/src/backend/lan_discovery.rs`) exercise `handle_event` directly against hand-built `ServiceInfo`/`ServiceEvent` values (no real daemon register/browse, so none of the existing `#[ignore]`d test's documented multicast flakiness applies):
  - self-resolution with blank TXT is never inserted as a peer (the exact live-log failure mode)
  - self-resolution with full TXT is also never inserted (no regression)
  - a blank-TXT re-fire does not clobber a previously-known-good peer's hostname/version/auth_key/instance_id
  - address/port still update unconditionally on a blank-TXT re-fire
  - a genuinely new peer first seen via blank TXT has no data to preserve (documents expected behavior, not a regression)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::lan_discovery::` — 28 passed, 0 failed, 1 ignored (pre-existing, documented-flaky real-multicast test, untouched)
- [x] `cargo clippy -p agentmux-srv --bin agentmux-srv --tests` — zero warnings in the new/changed code
- [ ] Manual: rebuild + run multi-instance, confirm network panel shows stable correct hostnames/versions and zero phantom self-entries over time (not yet done — noted as follow-up in the spec)

### DIFF
--- a/.changesets/1786856101-fix-lan-stop-mdns-re-resolutions-with-blank-txt-records-from-xyyw.md
+++ b/.changesets/1786856101-fix-lan-stop-mdns-re-resolutions-with-blank-txt-records-from-xyyw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(lan): stop mDNS re-resolutions with blank TXT records from clobbering known-good LAN peer metadata
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -396,15 +396,29 @@ impl LanDiscovery {
     fn handle_event(&self, event: ServiceEvent) {
         match event {
             ServiceEvent::ServiceResolved(info) => {
+                let fullname = info.get_fullname().to_string();
+
+                // Skip self. Compared by `fullname` (deterministic from
+                // `service_name`, set once at registration — see `start()`)
+                // rather than the TXT-record `instance_id` property: on this
+                // machine's own virtual/link-local interfaces (WSL/Hyper-V/
+                // VPN adapters), `enable_addr_auto()` re-fires
+                // `ServiceResolved` far more often than fresh TXT data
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2599, opened 2026-08-16
Manoz@Area54 feat(browser-pane): unified context menu replaces Chromium's native right-click menu

GenericAgentX-asaf:
## Summary
- Right-clicking a browser pane's content showed Chromium's own native context menu (Back/Forward/Reload/Print/Save As/View Page Source/Inspect) instead of AgentMux's own pane menu (split/replace/color/close). Implements CEF's `ContextMenuHandler` for browser panes only (`run_context_menu`, gated on `is_browser_pane` — the main app's own right-click menu is unaffected, it's handled entirely in DOM via `app.tsx`'s `onContextMenu={showTextInputContextMenu}`). Always suppresses the native menu (`callback.cancel()`, never `.cont()` — no CEF-native command is ever run through this path) and emits `browser-pane-context-menu` with the click point (pane-local coordinates) plus link URL / selection text / editable state / back-forward history.
- The frontend already has an extension point for exactly this — `getBodyContextMenuItems()`, already used by `termViewModel.ts`/`sysinfo-model.ts` — but a browser pane's native overlay content has no real DOM `contextmenu` event to trigger it. `blockframe.tsx` now also listens for `browser-pane-context-menu`, translates the pane-local click point into window-client coordinates via the pane's own DOM wrapper rect (`frameEl`, which the native overlay is always positioned to exactly match), and drives the *identical* `onBodyContextMenu` code path via a synthetic event — so item composition/ordering matches every other pane type exactly, no duplicated menu-building logic.
- `BrowserModel.getBodyContextMenuItems()` provides Back / Forward / Reload (reusing the existing `goBack()`/`goForward()`/`reload()` — no new backend surface needed), Copy Link Address (only when right-clicking an actual link), Print, View Page Source, and Inspect Element.
- Print / View Page Source / Inspect Element needed new backend surface: three `browser_pane_*` IPC commands backed by CEF's own `BrowserHost::print()`/`show_dev_tools()` and a `view-source:` navigation (added `"view-source"` to `PANE_ALLOWED_NAV_SCHEMES` — Chromium renders it internally, same as `chrome-devtools:`, so it carries none of the OS-handoff/UAC risk that allowlist exists to block).

Design doc: `docs/specs/SPEC_BROWSER_PANE_UNIFIED_CONTEXT_MENU_2026_08_15.md`.

## Scope / caveats
- Unlike the two sibling native-overlay fixes (click-to-select, click-dismisses-menus), CEF's context-menu callback is NOT part of the Windows-HWND-subclass/macOS-swizzle machinery — it's a standard CEF client handler, so this should work identically on all three platforms including Linux. Flagged as needing verification during manual testing, not assumed.
- Out of scope for v1 (noted in the spec): Save As / Save Image As / spell-check suggestions / Cast.

## Test plan
- [x] `cargo check -p agentmux-cef` clean, no new warnings
- [x] `tsc --noEmit` clean (2 pre-existing, unrelated errors in armory-view.tsx/warden-view.tsx, present on main before this change)
- [x] Full `vitest` suite: 2626 passed. One unrelated flaky test (`tool-renderers/registry.test.ts`, a timing-sensitive test with no connection to browser panes) failed under full-suite parallel load but passes cleanly in isolation (2755ms) — not a regression from this change.
- [ ] Manual: right-click a loaded browser pane, confirm the app-style menu appears (not Chromium's), Back/Forward reflect real history state, Reload/Print/View Source/Inspect Element all work, Copy Link Address appears only over a link, split/replace/color/close all still work
- [ ] Manual: confirm the pane HEADER's right-click menu (already real DOM, untouched by this change) still works identically

### DIFF
--- a/.changesets/1786856221-feat-browser-pane-unified-context-menu-replaces-chromium-s-n-ysgw.md
+++ b/.changesets/1786856221-feat-browser-pane-unified-context-menu-replaces-chromium-s-n-ysgw.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(browser-pane): unified context menu replaces Chromium's native right-click menu
--- a/agentmux-cef/src/browser_pane/callbacks.rs
+++ b/agentmux-cef/src/browser_pane/callbacks.rs
@@ -146,9 +146,12 @@ pub fn on_before_close_browser_pane(state: &Arc<AppState>, label: &str) {
 
     // Cross-platform: drop this pane's zoom-factor entry so
     // `browser_pane_zoom` doesn't grow unboundedly as panes are opened and
-    // closed over a session.
+    // closed over a session. Same for the stashed context-menu frame
+    // (SPEC_BROWSER_PANE_UNIFIED_CONTEXT_MENU_2026_08_15.md) — also holds a
+    // cloned `Frame` that must not outlive the closing browser.
     if let Some(block_id) = block_id {
         state.browser_pane_zoom.lock().remove(block_id);
+        state.browser_pane_context_menu_frame.lock().remove(block_id);
     }
 
     // Windows only:
--- a/agentmux-cef/src/browser_panes/navigation.rs
+++ b/agentmux-cef/src/browser_panes/navigation.rs
@@ -128,19 +128,37 @@ impl BrowserPaneManager {
     /// PR #2599). `Frame::copy/cut/paste` operate on whatever is currently
     /// selected/focused in that frame, exactly like CEF's own native menu
     /// commands would have.
+    ///
+    /// Uses `state.browser_pane_context_menu_frame`'s stashed frame — the
+    /// ACTUAL frame `client::context_menu::run_context_menu` saw the
+    /// right-click land on — instead of unconditionally
+    /// `b.main_frame()`. For a page with sub-frames/iframes (ads, embeds,
+    /// widgets), the selection/focus can live in a sub-frame; acting on
+    /// `main_frame()` there would silently no-op or paste into the wrong
+    /// place (reagentx P1, second finding on PR #2599). Falls back to
+    /// `main_frame()` only if nothing was stashed (e.g. this pane never had
+    /// its context menu invoked in this session).
+    fn edit_target_frame(&self, block_id: &str, state: &Arc<AppState>, browser: &mut Browser) -> Option<Frame> {
+        state
+            .browser_pane_context_menu_frame
+            .lock()
+            .get(block_id)
+            .cloned()
+            .or_else(|| browser.main_frame())
+    }
     pub fn copy(&self, block_id: &str, state: &Arc<AppState>) {
         if let Some(mut b) = self.live_browser(state, block_id) {
-            if let Some(frame) = b.main_frame() { frame.copy(); }
+            if let Some(frame) = self.edit_target_frame(block_id, state, &mut b) { frame.copy(); }
         }
     }
     pub fn cut(&self, block_id: &str, state: &Arc<AppState>) {
         if let Some(mut b) = self.live_browser(state, block_id) {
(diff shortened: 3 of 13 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed both findings from reagentx in e36f9b0a:

- **context_menu.rs:75 (P1, floating-window delivery)** — routes `browser-pane-context-menu` via `emit_event_to_window` + `state.browser_pane_window_label(&block_id)` instead of `emit_event_from_state` (which always targeted "main"), mirroring the existing fix for `browser-pane-shortcut`. Also changed the suppression logic: the native menu is now only suppressed when delivery actually succeeds — every failure path (unresolvable block_id/params, or no owning window) falls through to CEF's native menu instead of showing nothing.
- **navigation.rs:97 (P2, double view-source prefix)** — `view_source()` now checks for an existing `view-source:` prefix before adding one.

`cargo check -p agentmux-cef` clean, no new warnings.

GenericAgentX-asaf:
Addressed all three findings from reagentx's second-pass review in 14561142:

- **browser-model.ts (P1, Reload no-op)** — the menu item now invokes `browser_pane_reload` directly instead of `this.reload()` (which only did local reducer state tricks, never touching the real native CEF browser view — turned out to have no other real caller in the app; `browser-nav-bar.tsx`'s own Reload button already bypassed it the same way).
- **blockframe.tsx (P1, menu offset)** — now translates the click point using `.browser-placeholder`'s rect (queried within `frameEl`) instead of `frameEl`'s own rect, since that's what CEF's `xcoord`/`ycoord` are actually relative to — `frameEl` also includes the header/title bar/nav bar above it.
- **browser-model.ts / navigation.rs / ipc.rs (P2, lost Cut/Copy/Paste)** — added three new `browser_pane_copy/cut/paste` IPC commands backed by CEF's `Frame::copy/cut/paste`, plus the corresponding menu items (Copy when a selection exists, Cut/Paste when the target is editable).

`cargo check -p agentmux-cef` clean, `tsc --noEmit` clean (2 pre-existing unrelated errors), browser+block test suites green.

GenericAgentX-asaf:
Addressed both findings from reagentx's third-pass review in a78d0bac:

- **handlers.rs/context_menu.rs (P1, wrong frame for Copy/Cut/Paste)** — the CEF `frame` param (the actual frame that was right-clicked) was being discarded; for a page with sub-frames/iframes, Copy/Cut/Paste would silently act on the top-level frame instead of wherever the selection/focus actually was. Now stashes the right-clicked frame per-block (`state.browser_pane_context_menu_frame`), consumed by copy/cut/paste, cleared on pane close.
- **context_menu.rs (P2, stale doc reference)** — fixed the comment referencing a `browser-pane-context-menu-bridge.ts` file that was never created; the real listener is inline in `blockframe.tsx`.

`cargo check -p agentmux-cef` clean, no new warnings.

GenericAgentX-asaf:
Addressed both findings from reagentx in a438a4fb:

- **platform.rs:695 (P0, view-source scheme bypass)** — this was a real gap. Adding "view-source" to the flat allowlist wasn't enough since \`url_scheme()\` only inspects up to the first colon; \`view-source:file:///etc/passwd\` had outer scheme "view-source" (allowed) with the nested target never validated. Fixed: \`is_disallowed_pane_nav_scheme\` now recursively validates the nested target when the outer scheme is "view-source", restricted to http(s) specifically. Added unit tests for both directions (11/11 pass, including the 2 new ones).
- **context_menu.rs:84 (P2, stale comment)** — fixed the comment to reference \`.browser-placeholder\` (the actual element used) instead of the outer \`.block-<id>\` wrapper.

\`cargo check\` + \`cargo test -p agentmux-cef external_url_tests\` both clean.

PR #2600, opened 2026-08-16
fix(drop): stop an unhandled file drop from navigating the whole window away

GenericAgentX-asaf:
## Summary

A user dragged a `.md` file onto a running AgentMux window "just to see what it would do." It landed outside any pane's own drop zone, and Chromium's default action for an unhandled file drop — navigate the top-level frame to the dropped file — ran unopposed. Since AgentMux's window controls are rendered inside that same page (not native OS chrome), the navigation destroyed the entire UI, including the close/minimize/maximize controls, with no in-app way to recover. Full root-cause writeup: `docs/retro/retro-md-drop-window-hijack-and-55-6-relaunch-failure-2026-08-16.md`.

Root cause: file drop is only handled by Terminal (`term.tsx`) and Agent (`useAgentDropAttach.ts`) pane elements — there was no window/document-level guard for drops landing on the tab strip, title bar, splitters, or any other pane type.

## Changes

- `frontend/util/dnd.ts`: add `isFileDrag(e)` — true only when `dataTransfer.types` includes `"Files"`.
- `frontend/app-init.ts`: add `installGlobalDropGuard()`, wired in as early as possible in `initAppInner()`. Registers `window`-level `dragover`/`drop` listeners (no `capture`, so pane-level handlers still get first chance) that call `preventDefault()` only for file drags — text/URL drags are untouched, so the composer's default text-insertion-on-drop behavior keeps working.
- `frontend/util/dnd.test.ts`: unit tests for the new helper (plus `baseName`, which had none).

## Test plan

- [x] `npx tsc --noEmit` — clean (3 pre-existing, unrelated errors in untouched files)
- [x] `npx vitest run frontend/util/dnd.test.ts` — 4/4 passing
- [ ] Manual click-test in a packaged/dev build: drag a file onto the tab strip / title bar / pane gaps and confirm the window no longer navigates away (not exercised — requires a `task package`/`task dev` build cycle)

### DIFF
--- a/.changesets/1786881678-fix-drop-stop-an-unhandled-file-drop-from-navigating-the-who-tlo4.md
+++ b/.changesets/1786881678-fix-drop-stop-an-unhandled-file-drop-from-navigating-the-who-tlo4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(drop): stop an unhandled file drop from navigating the whole window away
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2604, opened 2026-08-16
Clamk@claudius: fix jekt signing-key injection gap on the normal launch path

GenericAgentX-asaf:
## Summary

Root-caused and fixed why normally-launched agents' jekts render `TRUST=self-declared` instead of `TRUST=host-verified` even on builds well past when host-tier (v0.55.7) and LAN-tier (v0.55.9) signing shipped.

**Root cause, confirmed with live evidence (see the included report for full detail):** the signing-key injection code only ever lived in `agent.open`'s Rust handler. The actual "click Launch" path — `WriteAgentConfig`, driven entirely by the frontend's `agent-config-builder.ts` — writes `.mcp.json` independently and never called it. Verified directly on this machine: a running v0.55.9 instance's `.mcp.json`, rewritten today, has no `AGENTMUX_JEKT_KEY`/`AGENTMUX_LAN_KEY`; the server log shows the write came from `WriteAgentConfig` with zero `agent_open`/`jekt` log lines anywhere in the entire file.

**Fix:** extracted the injection logic into one shared, pure function (`agent_config::inject_jekt_signing_keys_into_mcp_json`), called from both `agent.open` (replacing its former inline duplicate) and `WriteAgentConfig` (new — this is the actual fix). Matches this module's own existing pattern for why shared logic lives here (its doc comment already warns about exactly this kind of two-call-site drift for the skill-file cleanup logic it centralizes).

Full evidence trail and research method: `docs/specs/REPORT_JEKT_SIGNING_KEY_INJECTION_GAP_2026_08_16.md`.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2350 passed, 0 failed, 6 pre-existing ignores, no regressions
- [x] New tests: `inject_jekt_signing_keys_into_mcp_json_{patches_both_keys_into_the_env_block,reuses_the_same_key_on_a_second_call,returns_none_when_theres_no_env_object_to_patch,returns_none_on_malformed_json}` (agent_config.rs)
- [x] New test: `writeagentconfig_extracts_slug_and_injects_signing_keys_into_the_real_mcp_json_shape` (editor_handlers.rs) — exercises the real `rpc_types::AgentConfigFile{path, content}` shape, not just the pure function directly

### DIFF
--- a/.changesets/1786886693-fix-jekt-inject-host-lan-signing-keys-on-the-normal-launch-p-5ce3.md
+++ b/.changesets/1786886693-fix-jekt-inject-host-lan-signing-keys-on-the-normal-launch-p-5ce3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(jekt): inject host/LAN signing keys on the normal launch path, not just agent.open
--- a/agentmux-srv/src/backend/agent_config.rs
+++ b/agentmux-srv/src/backend/agent_config.rs
@@ -877,6 +877,50 @@ pub fn write_managed_skill_file_manifest(
     }
 }
 
+/// Best-effort: mint-or-reuse `agent_slug`'s jekt/LAN signing keys
+/// (`SPEC_JEKT_TRUST_LAYER_COMPLETION_2026_08_13.md` §2.2,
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

Issue #2603, opened 2026-08-16
Clamk@claudius: Agent identity/history persistence protocol (tracking)


GenericAgentX-asaf:
## Tracking issue — COMPLETE

Sequenced fix for named agents' conversation history fragmenting across channel/identity UUIDs on every rebuild (root cause: PR #2431's `identities_dir()` default-isolation flip, 2026-08-06, never reconciled with `ClaudeHistoryAdapter`'s scan list — full writeup in `docs/specs/REPORT_AGENT_IDENTITY_HISTORY_FRAGMENTATION_2026_08_16.md`).

Design: `docs/specs/SPEC_AGENT_IDENTITY_HISTORY_PERSISTENCE_PROTOCOL_2026_08_16.md` (5-invariant protocol P1-P5, synthesized with the Mandatory ABF work from PR #2587). Reviewable checklist for future PRs in this area: `docs/specs/CHECKLIST_AGENT_DATA_SCOPE_ROUTING_2026_08_17.md`.

All six steps merged.

### Steps

- [x] **Step 1** — Fix stale Phase 3a/3b migration-phase doc comments, add a test pinning the real per-function read split (protocol P4). #2602 — **merged**
- [x] **Step 2** — Confirmed DEFINITION/PORTABLE CONFIG store-routing regression tests already exist. No new work needed.
- [x] **Step 3** — Split CREDENTIAL isolation from CONVERSATION HISTORY storage in `identities_dir()` (protocol §4.1). #2605 — **merged**
- [x] **Step 4a** — Fix `ClaudeHistoryAdapter`'s read path to also scan per-channel identity dirs (protocol §4.4, first half). #2606 — **merged**
- [x] **Step 4b** — Build a fast, `agent_id`-keyed history index (protocol §4.4, second half). #2611 — **merged**
- [x] **Step 5** — Add an opt-in `bundle.export_for_agent_with_history` action, separate from the default ABF export (protocol §3.3). #2613 — **merged**, after 4 review rounds (junction-metadata dispute — disproven, not fixed; index staleness for a "complete" export; blocking sync I/O on the async worker, in two separate spots; unbounded per-session AND aggregate memory).
- [x] **Step 6** — Distill P1-P5 into a short, reviewable PR checklist (protocol §4 step 6). #2614 — **merged**

Side quest, not part of this sequence but discovered while working on it: #2604 (jekt host/LAN signing-key injection gap) — **merged**.

PR #2431, opened 2026-08-06
feat(identity): default isolated auth for every non-stable channel

AgentA-asaf:
## Summary

`AGENTMUX_ISOLATED_AUTH` (shipped in `SPEC_ISOLATED_AUTH_DEV_TESTING_2026_07_27.md`) gives a `task dev`/`task package` boot its own channel-scoped identity store and OAuth credential dirs instead of sharing the real global account list — but it's opt-in, so almost nobody sets it, and every routine dev/test boot silently inherits a fully-authenticated global session. That defeats the point for a repo actively hardening its OAuth login/relogin surfaces (#2422, #2423, #2425, #2429): the login code paths mostly never execute during normal testing.

This flips the default: `isolated_auth_enabled()` now isolates by default for any channel other than `stable`. `stable` (real installed/portable releases — the actual daily-driver) is explicitly unaffected. `AGENTMUX_ISOLATED_AUTH=0`/`=1` still override explicitly in either direction.

Full rationale, resolution-order design, and phased rollout: `docs/specs/SPEC_ISOLATED_AUTH_DEFAULT_BY_CHANNEL_2026_08_06.md` (amends the July 27 spec — mechanism unchanged, only the default).

## Changes

- `agentmux-common/src/data_paths.rs`: `isolated_auth_enabled()` now resolves via a new `isolated_auth_reason()` (single source of truth, exposed for diagnostics) — explicit `AGENTMUX_ISOLATED_AUTH=1`/`=0` always wins; otherwise isolates iff `AGENTMUX_CHANNEL != "stable"`; stays global if `AGENTMUX_CHANNEL` is unset entirely (conservative fallback for bare `cargo test`/`cargo run`).
- `agentmux-common/src/lib.rs`: re-export `isolated_auth_reason`/`IsolatedAuthReason`.
- `agentmux-srv/src/bootstrap.rs`: the "shared store: attached" boot log line now includes which of the four states applied (`explicit opt-in` / `explicit opt-out` / `channel default — isolated` / `channel default — global`).
[...]

PR #2605, opened 2026-08-16
Clamk@claudius: keep conversation history global under credential isolation

GenericAgentX-asaf:
## Summary

Step 3 of #2603 (agent identity/history persistence protocol). Fixes the actual, currently-live regression root-caused in `docs/specs/REPORT_AGENT_IDENTITY_HISTORY_FRAGMENTATION_2026_08_16.md`: `identities_dir()` isolating a provider's credential directory per-channel (the default for every non-`stable` channel since PR #2431) silently isolates that provider's conversation transcripts too, since e.g. Claude Code's `projects/` lives inside the same directory tree as its credentials.

**Fix:** `DataPaths::identity_history_dir()` always resolves globally, independent of `isolated_auth_enabled()`. `ensure_history_link()` makes an isolated credential dir's `projects/` subpath a directory junction (Windows) / symlink (Unix) pointing at that global location — so the provider CLI is unaware anything changed, but its session data lands in one stable place regardless of which channel/build spawned it. Migrates any pre-existing real directory's contents into the global target first (never overwriting, never deleting), so history that predates this fix isn't orphaned.

Used the `junction` crate on Windows rather than `std::os::windows::fs::symlink_dir` (requires `SeCreateSymbolicLinkPrivilege` — admin or Developer Mode, which this app can't assume) or hand-rolled `DeviceIoControl`/`FSCTL_SET_REPARSE_POINT` FFI (this is credential-adjacent storage — not a place to improvise low-level filesystem code).

Wired into both `identity_auth_dirs.rs` call sites, gated on `isolated_auth_enabled()` (a no-op when off, since `dir/projects` already IS the global location in that case).

## Verification

Beyond `cargo test`, verified empirically on this real Windows machine before wiring into any credential-handling code:
- Created a real junction via the new code, confirmed its type (`Junction`) and target via PowerShell `Get-Item`
- Confirmed a file written at the global target reads back through the isolated link path, both via Rust's own `fs` APIs and native `Get-Content`
- The migration test caught a real bug during development (`junction::create` needs the link's parent dir to pre-exist, not just the target) — fixed before this PR, not left for review to catch

## Test plan

- [x] `cargo test -p agentmux-common --lib` — 148 passed, including 4 new tests on `ensure_history_link` (creation + read-through, idempotency, pre-existing-directory migration, never-overwrite-on-collision)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2346 passed, 0 failed, 6 pre-existing ignores, including a new end-to-end test (`compute_account_dir_links_projects_to_the_global_history_location_when_isolated`) exercising the real `compute_and_ensure_account_dir` call site, not just the linking primitive in isolation

### DIFF
--- a/.changesets/1786887880-fix-identity-keep-conversation-history-global-when-credentia-9wuk.md
+++ b/.changesets/1786887880-fix-identity-keep-conversation-history-global-when-credentia-9wuk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): keep conversation history global when credential isolation is on
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,6 +82,7 @@ dependencies = [
  "dirs",
  "ed25519-dalek",
  "hmac",
+ "junction",
  "serde",
  "serde_json",
  "sha2",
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

PR #2606, opened 2026-08-16
Clamk@claudius: ClaudeHistoryAdapter now scans per-channel isolated identity dirs

GenericAgentX-asaf:
## Summary

Step 4a of #2603 (protocol §4.4's read-path fix, first half). `ClaudeHistoryAdapter` never scanned `<home>/channels/*/identities/*/claude/projects` or `<home>/dev/*/(*/)?identities/*/claude/projects` — the per-channel-isolated locations PR #2431's default flip sends history to. Step 3 (#2605, merged) links those forward to the global location going forward, but history written before that fix landed, or for a bundle whose link creation failed (best-effort, can fail), still has real data only under the per-channel path. This makes it discoverable through the adapter again instead of requiring a manual filesystem grep — which is literally how this whole investigation started (see `docs/specs/REPORT_AGENT_IDENTITY_HISTORY_FRAGMENTATION_2026_08_16.md`).

Bounded scan (depth 2, not unbounded recursion), covering both known layouts. Deduplicates by canonical path — after Step 3, a per-channel `projects/` can be a junction pointing at the exact same physical location as an already-scanned global entry, and without dedup every session under it would surface twice.

Zero test coverage existed for this file's directory-discovery logic before this change. Extracted `push_deduped_dir` out of an inline closure into an independently-testable function; added 7 tests, including one that creates a real Windows junction and confirms it dedupes correctly against its target.

**Splitting scope:** protocol step 4's other half (a fast, `agent_id`-keyed history index so lookup doesn't need a filesystem walk at all) is a distinct piece of work and will follow as its own PR — this one is scoped to making existing lookups correct first.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2361 passed, 0 failed, 6 pre-existing ignores
- [x] New: 7 tests on `push_deduped_dir`/`scan_isolated_identities_under`, including a real-Windows-junction dedup case

### DIFF
--- a/.changesets/1786904766-fix-history-claudehistoryadapter-now-scans-per-channel-isola-543k.md
+++ b/.changesets/1786904766-fix-history-claudehistoryadapter-now-scans-per-channel-isola-543k.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(history): ClaudeHistoryAdapter now scans per-channel isolated identity dirs too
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #2607, opened 2026-08-16
fix(identity): resolve provider through the bound bundle at the three backend clone-sites

AgentY-asaf:
## Summary

Delivery-plan item 2 of #2594 (harness/model decoupling & ABF portability) — the three backend clone-sites (`template.rs:119,132`, `template.rs:357`, `v1_templates.rs:265`).

`agentdefcreatefromtemplate`, `forkagentdefinition`, and the template-promote migration all cloned a source/template definition's `.provider` column directly instead of resolving through its own bound bundle — the same "gate vs. actual launch can disagree" drift class #2592 fixed for the credential gate and launch path. If a template/source's `provider` column had drifted from its (immutable) bundle copy, a clone/fork/promote would silently inherit the stale value instead of the real one.

- `agentdefcreatefromtemplate` (`template.rs:119,132`) and `forkagentdefinition` (`template.rs:357`) now call `id_store.resolve_effective_provider_id()`, already available in both RPC handler scopes. The create-from-template vendor-base-url validation now uses the same resolved value it validates the new definition against, instead of a separately-read (and possibly different) column.
- The template-promote migration (`v1_templates.rs:265`) calls `wstore.resolve_effective_provider_id()` — only `wstore` is available at that point in the boot sequence (no id_store/shared-store handoff has happened yet), which safely falls back to the drifted column when the bundle isn't found via that store — same behavior as before this fix, strictly better when the bundle is reachable.

## Test plan

- [x] Regression tests for all three sites seed a template/source with a drifted `.provider` (`"codex"`) against a bundle with the real provider (`"claude"`) and assert the clone/fork/promote carries `"claude"`.
- [x] Verified all three new tests fail against the pre-fix code (temporarily reverted the three fixes, kept the tests — all three failed with `left: "codex", right: "claude"`), then confirmed they pass again with the fix restored.
- [x] `cargo check -p agentmux-srv` — clean, no new warnings/errors.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::storage::` — 282/282 pass.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::agent_session::` — 35/35 pass (1 pre-existing ignore).
- [x] `cargo test -p agentmux-srv --bin agentmux-srv server::agent_handlers::` — 30/30 pass, including the 2 new RPC-level tests.
- [x] Merged latest `main` into the branch before pushing (picked up #2606).

### DIFF
--- a/.changesets/1786905565-fix-identity-resolve-provider-through-the-bound-bundle-at-th-y6jm.md
+++ b/.changesets/1786905565-fix-identity-resolve-provider-through-the-bound-bundle-at-th-y6jm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): resolve provider through the bound bundle at the three backend clone-sites
--- a/agentmux-srv/src/backend/agent_session/migrations/v1_templates.rs
+++ b/agentmux-srv/src/backend/agent_session/migrations/v1_templates.rs
@@ -257,12 +257,21 @@ pub fn migrate_promote_template_sessions_v1(
             // at the deterministic id. Field copies mirror
             // `agent_def_create_from_template`.
             let now = now_ms() as i64;
+            // Resolve through the template's own bound bundle when it has
+            // one, not the possibly-drifted `db_agent_definitions.provider`
+            // column directly (#2594, same pattern as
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2608, opened 2026-08-16
fix(identity): remove the post-creation model-vendor-base-url editor from AgentIdentityModal

AgentY-asaf:
## Summary

Delivery-plan item 3 of #2594 (harness/model decoupling & ABF portability) — the "needs a product decision" item on `AgentIdentityModal.tsx`. **Decision made directly with the human operator in this conversation** (not via jekt/relay): remove the edit surface.

`AgentIdentityModal.tsx` let a user actively edit `model_vendor_base_url` on the agent definition after its bundle already exists — the one remaining post-creation write path for a field the Mandatory ABF architecture (`ARCHITECTURE_MANDATORY_ABF_RETHINK_2026_08_14.md` §7.4.1) treats as set-once at definition/bundle-creation time, same as `provider` (which this modal has never let a user edit either — there's no provider field in this UI at all).

Unlike `provider`, the bundle (`Memory`) has **no field to mirror the raw base-url string into at all** — `m0021`'s backfill migration only derives a coarse `"custom"` vendor *label* into the bundle's `model` column, never the actual URL string. So there's no bundle-side value to resolve reads through the way #2592/#2607 fixed provider drift (`resolve_effective_provider_id`) — every post-creation edit here just wrote a second, permanently-diverging copy onto `AgentDefinition` with nowhere for it to reconcile. The only consistent fix available at this scope is to stop allowing that second write at all, matching how `provider` is already handled.

The definition-time editor in `AgentCreateFromTemplateModal.tsx` is unaffected — it still sets the value once, before/at bundle provisioning, which is the correct place for it.

## Changes

- Removed the "Model Vendor / Custom Endpoint" section (`supportsCustomEndpoint` memo, `modelVendorBaseUrl`/`vendorSaving`/`vendorError` signals, `handleVendorSave`) from `AgentIdentityModal.tsx`.
- Removed the now-dead `.agent-identity-vendor-*` SCSS rules.
- No backend changes — `updateagent`'s handling of `model_vendor_base_url` is untouched (still supports `AgentCreateFromTemplateModal`'s definition-time write via `agentdefcreatefromtemplate`, a separate RPC).

## Test plan

- [x] New regression test renders `AgentIdentityModalPanel` for an agent whose bundle already exists (`memory_id` set) and provider supports a custom endpoint (`claude`), asserts no vendor-editing input/save-button/label is present.
- [x] Verified the test fails against the pre-fix code (found the real `<input data-testid="agent-identity-vendor-base-url-input">` in the DOM).
- [x] `npx tsc --noEmit -p tsconfig.json` — same 2 pre-existing, unrelated errors as `main` (`armory-view.tsx`/`warden-view.tsx` `MetaType` — confirmed via `git stash`), nothing new.
- [x] `npx vitest run app/view/agent/components/` — 341/342 pass; the 1 failure (`tool-renderers/registry.test.ts`, a timeout) is a pre-existing flake unrelated to this change — passes cleanly in isolation.
- [x] Confirmed no other frontend references to the removed CSS classes/handlers remain.

### DIFF
--- a/.changesets/1786924587-fix-identity-remove-the-post-creation-model-vendor-base-url--34n9.md
+++ b/.changesets/1786924587-fix-identity-remove-the-post-creation-model-vendor-base-url--34n9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): remove the post-creation model-vendor-base-url editor from AgentIdentityModal
--- a/frontend/app/view/agent/components/AgentIdentityModal.scss
+++ b/frontend/app/view/agent/components/AgentIdentityModal.scss
@@ -16,64 +16,10 @@
 // The ambient-login opt-in row that used to live here was removed (reagent
 // P1 on #2262 — see AgentIdentityModal.tsx's doc comment): the layer-3
 // spawn gate no longer honors use_ambient_login at all, so the toggle was
-// dead UI promising a fallback that no longer existed.
-
-.agent-identity-vendor-section {
-    display: flex;
-    flex-direction: column;
-    gap: var(--space-1-5);
-    padding: var(--space-3) var(--space-5);
-    border-top: 1px solid var(--border-color);
-}
-
-.agent-identity-vendor-label {
-    font-size: var(--text-sm);
-    font-weight: var(--font-weight-medium);
-    opacity: 0.75;
-}
-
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2609, opened 2026-08-17
fix(agent): AgentPicker resolves provider through the bound bundle for install-check, prereq-probe, and add-account

AgentY-asaf:
## Summary

Delivery-plan item 4 of #2594 (harness/model decoupling & ABF portability) — the `AgentPicker.tsx` portion of "AgentPicker.tsx, AgentInstallModal.tsx, and the two smaller identity-linking sites." The issue text explicitly allows these to "land together or split further" — splitting here: `AgentInstallModal.tsx` and the identity-linking sites (`bind-to-agent-menu.ts`, `agent-identity-links-panel.tsx:271`) each need their own treatment (the install modal needs an async-resolution-with-loading-guard rewrite similar in scope to #2596) and will follow as separate PRs.

Five live call sites in `AgentPicker.tsx` read `agent.provider` directly instead of resolving through the agent's bound bundle — the same "gate vs. actual launch can disagree" risk class #2592/#2596/#2607 fixed:

- `checkInstalled` / `handleTemplateSelect`'s install-check gating — could check/install the WRONG provider's CLI, or skip the check entirely, while the actual launch (`agent-model.ts`, already resolves correctly via `resolveEffectiveLaunchProvider`) spawns the real, uninstalled provider.
- `onRequestAddAccount` — could open an "add account" flow for the drifted provider, creating an account the agent can never actually use at launch.
- `probeMissingPrereqs` — could probe system prereqs for the wrong provider, letting a launch through with the real provider's prereqs never checked.
- `buildTemplateInstallRequest`'s `onInstalled` cache-invalidation — compared the just-installed agent's own drifted `.provider` against the resolved canonical id, so its own cache entry could never match and get marked installed (every subsequent click re-triggered an install check).

## Other changes

- `handleTemplateSelect` had a second, redundant `agent.provider` read duplicating `checkInstalled`'s own npm-installable check — removed outright in favor of always calling `checkInstalled` (which already handles the "no install needed" case internally), so there's one fewer place for the two to independently drift again.
- `buildInstallRequest` — the generic, pre-two-tier install-request builder containing the same unfixed pattern — turned out to be **dead code** (its only caller, the legacy single-tier `handleSelect`, was already removed in the Phase 1 rewrite; `buildTemplateInstallRequest` is the only live install-request builder). Deleted rather than patched.
- The bulk cache-invalidation loop over ALL agents in `buildTemplateInstallRequest`'s `onInstalled` still compares *other* agents via their own `.provider` column as a heuristic — a deliberate scope boundary: any of them that has also drifted just misses the bulk shortcut and gets a correct, resolved answer the next time it hits `checkInstalled` itself (now fixed), never a wrong CLI/credential outcome.

## Test plan

- [x] Regression tests seed a template with a drifted `.provider` (`"claude"`, not npm-installable) against a bundle resolving to a different, npm-installable provider (`"codex"`), and assert: `InstallCheckCommand`/`ResolvePrereqsCommand` are called with the bundle's provider, and the just-installed agent's own cache entry is correctly marked (a second click doesn't re-check).
- [x] Verified all three new tests fail against the pre-fix code (temporarily reverted via `git stash`, kept the tests) — presence/absence of RPC calls and wrong `kind` assertions all failed as expected.
- [x] `npx vitest run app/view/agent/ app/view/identity/` — 1206/1207 pass; the 1 failure (`tool-renderers/registry.test.ts`, a timeout) is the same pre-existing flake already confirmed unrelated in #2608's PR.

### DIFF
--- a/.changesets/1786925818-fix-agent-agentpicker-resolves-provider-through-the-bound-bu-04jt.md
+++ b/.changesets/1786925818-fix-agent-agentpicker-resolves-provider-through-the-bound-bu-04jt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): AgentPicker resolves provider through the bound bundle for install-check, prereq-probe, and add-account
--- a/frontend/app/view/agent/components/AgentPicker.test.tsx
+++ b/frontend/app/view/agent/components/AgentPicker.test.tsx
@@ -29,6 +29,12 @@ vi.mock("@/app/store/rpc-api", () => {
         ListNamedAgentsCommand: vi.fn().mockResolvedValue([]),
         InstallCheckCommand: vi.fn().mockResolvedValue({ installed: true }),
         ResolvePrereqsCommand: vi.fn().mockResolvedValue({ results: [] }),
+        // Backs `resolveEffectiveLaunchProvider`'s bound-bundle resolution
+        // (#2594) — resolves to `undefined` by default so existing tests
+        // (none of whose agent fixtures set `memory_id`) never even
+        // trigger the fetch; the drift-regression tests below set their
+        // own `.mockResolvedValue`.
+        GetMemoryCommand: vi.fn().mockResolvedValue(undefined),
         AgentSessionReadCommand: vi
             .fn()
             .mockResolvedValue({ content: null, modts: null }),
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2610, opened 2026-08-17
fix(agent): AgentInstallModal installs the bound bundle's provider, not a drifted agent.provider

AgentY-asaf:
## Summary

Delivery-plan item 5 of #2594 (harness/model decoupling & ABF portability) — `AgentInstallModal.tsx`, "determines which CLI package literally gets installed." Same "gate vs. actual launch can disagree" risk class #2592/#2596/#2607/#2609 fixed.

`catalog`/`provider` (and everything derived from them — displayName, version, and critically `startInstall`'s `InstallStartCommand` payload: `providerId`/`cliCommand`/`npmPackage`/`pinnedVersion`) read `agent.provider` directly. If that column had drifted from the agent's bound bundle, this modal — the **one place that actually runs `npm install` for a provider's CLI** — would install the wrong provider's package while the agent's real (bundle) provider stayed uninstalled, even though AgentPicker's `checkInstalled` (already fixed in #2609) correctly identified the real provider needed installing in the first place.

## Fix

- Header display (icon/displayName/version) now resolves through a `createResource` over `resolveEffectiveLaunchProvider`, falling back to `agent.provider` while loading — same fallback contract the function itself documents.
- `startInstall` re-resolves directly (not via the resource) at click time, so a click racing the resource's own in-flight fetch still installs the correct provider rather than trusting a possibly-stale snapshot for the one call that actually matters.

## Test plan

- [x] Regression tests: a drifted-provider agent (`"claude"` column, `"codex"` bundle) must install `"codex"` (`InstallStartCommand` called with `codex`'s `providerId`/`cliCommand`/`npmPackage`/`pinnedVersion`) and show "Codex" in the header, not `"claude"`/"Claude Code".
- [x] An unbound agent (no `memory_id`) installs its own `agent.provider` directly with no bundle fetch at all — the no-drift baseline stays unchanged.
- [x] Verified the two drift-sensitive tests fail against the pre-fix code (temporarily reverted via `git stash`, kept the tests); the baseline test correctly passes both ways (proving it's a genuine no-regression check, not a tautology).
- [x] `npx vitest run app/view/agent/` — 1169/1169 pass, 94/94 files, no flakes this run.

### DIFF
--- a/.changesets/1786926699-fix-agent-agentinstallmodal-installs-the-bound-bundle-s-prov-k2ij.md
+++ b/.changesets/1786926699-fix-agent-agentinstallmodal-installs-the-bound-bundle-s-prov-k2ij.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): AgentInstallModal installs the bound bundle's provider, not a drifted agent.provider
--- a/frontend/app/view/agent/components/AgentInstallModal.test.tsx
+++ b/frontend/app/view/agent/components/AgentInstallModal.test.tsx
@@ -1,0 +1,177 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * AgentInstallModal must install the agent's EFFECTIVE (bundle-resolved)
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2611, opened 2026-08-17
Clamk@claudius: fast agent_id-keyed Conversation History lookup

GenericAgentX-asaf:
## Summary

Step 4b of #2603 (protocol §4.4, second half — the fast index, split from Step 4a's read-path fix in #2606, merged).

Before this, "this agent's conversation history" wasn't a queryable concept at all. The only lookup (`history.List`) filters by provider + a `working_directory` substring match across every session found on disk (personal + AgentMux both), and `SessionMeta` had no `agent_id`/identity field anywhere. This is the actual gap behind the operator's original ask that started this whole investigation: finding a specific agent's last conversation fast, without a manual filesystem grep.

- `SessionMeta` gains `identity_id`: the bundle/account id a session was discovered under (parsed from its file path), empty for personal/default non-bundle sessions.
- `SessionIndex` maintains a `by_identity: HashMap<identity_id, Vec<session_id>>` alongside the existing sessions map, rebuilt on every `refresh()`. New `list_for_identity()` is O(sessions for this identity), not O(total sessions on disk) like `list()`.
- `HistoryService::list_for_agent(store, agent_id, ...)` resolves `agent_id` to its linked identity bundle(s) via `Store::agent_identity_list_for_agent` (the same `db_agent_identity_links` table `identity_auth_dirs.rs` already keys off).
- New RPC: `history.ListForAgent(agent_id, offset, limit, sort_by, sort_dir)`.

**Not yet wired to a frontend UI action** — this PR is the backend lookup path; surfacing it as an actual "Conversation History" button is a follow-up.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2367 passed, 0 failed, 6 pre-existing ignores
- [x] 9 new tests, including one exercising the real `db_agent_identity_links` FK path end to end (agent definition row + account row + link row, not a mocked lookup)

### DIFF
--- a/.changesets/1786904766-fix-history-claudehistoryadapter-now-scans-per-channel-isola-543k.md
+++ b/.changesets/1786904766-fix-history-claudehistoryadapter-now-scans-per-channel-isola-543k.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(history): ClaudeHistoryAdapter now scans per-channel isolated identity dirs too
--- a/.changesets/1786927133-feat-history-fast-agent-id-keyed-conversation-history-lookup-g22r.md
+++ b/.changesets/1786927133-feat-history-fast-agent-id-keyed-conversation-history-lookup-g22r.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(history): fast agent_id-keyed Conversation History lookup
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2612, opened 2026-08-17
fix(identity): bind-to-agent menu and Connect-Claude CTA resolve provider through the bound bundle

AgentY-asaf:
## Summary

Delivery-plan item 6 of #2594 (harness/model decoupling & ABF portability) — the last two frontend sites: `bind-to-agent-menu.ts` ("filters which agents an OAuth account can bind to") and `agent-identity-links-panel.tsx:271` ("gates the 'Connect Claude account' CTA"). Same "gate vs. actual launch can disagree" risk class #2592/#2596/#2607/#2609/#2610 fixed. This is the last item on the frontend read-site list — only `exportagents`/`importagents` (lowest priority, no live spawn/credential consequence) remains after this.

## `bind-to-agent-menu.ts`

`computeBindCandidates` filtered CLI-OAuth accounts to same-provider agents by reading `agent.provider` directly. A drifted agent could be wrongly offered (or wrongly hidden) in the "Bind to Agent" submenu relative to what it actually launches as.

`computeBindCandidates` stays a pure, sync function — its own doc comment emphasizes it's "unit-testable without a DOM or RPC mocks." Rather than making it async (which would break that property and require every existing test to mock bundle fetches), I added an optional `resolveAgentProviderId` resolver parameter, defaulting to `(a) => a.provider` so every existing test is unaffected. `buildAccountRowMenu` (the async orchestrator, already doing one RPC round-trip for links) now batch-resolves every agent's effective provider first and passes a map-backed resolver.

## `agent-identity-links-panel.tsx`

The empty-state "Connect Claude account" CTA was gated on `agent()?.provider` directly — could offer the CTA for an agent that doesn't actually resolve to claude (linking an unrelated account to the wrong agent), or hide it for one that does. Now resolves via `createResource` over `resolveEffectiveLaunchProvider`, falling back to the raw column while loading (cosmetic-only risk here — button visibility, not a spawn/credential decision).

## Test plan

- [x] `computeBindCandidates` gains two resolver-aware cases: offers on drift-INTO-match, hides on drift-AWAY.
- [x] `buildAccountRowMenu` gains two batch-resolution cases exercising the same via `GetMemoryCommand` mocks.
- [x] `agent-identity-links-panel` gains two drift fixtures (agent-4 drifts INTO claude, agent-5 drifts AWAY from it) asserting the CTA follows the resolved provider, not the raw column.
- [x] All six new tests verified to fail against the pre-fix code (temporarily reverted each file via `git stash`, kept the tests).
- [x] All pre-existing tests in both files pass unchanged (the resolver defaults preserve prior behavior exactly).
- [x] `npx vitest run app/view/identity/ app/view/agent/` — 1216/1216 pass, 99/99 files, no flakes this run.

### DIFF
--- a/.changesets/1786927621-fix-identity-bind-to-agent-menu-and-connect-claude-cta-resol-6e84.md
+++ b/.changesets/1786927621-fix-identity-bind-to-agent-menu-and-connect-claude-cta-resol-6e84.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): bind-to-agent menu and Connect-Claude CTA resolve provider through the bound bundle
--- a/frontend/app/view/identity/agent-identity-links-panel.test.tsx
+++ b/frontend/app/view/identity/agent-identity-links-panel.test.tsx
@@ -14,10 +14,16 @@ import { cleanup, render, screen, waitFor, within } from "@solidjs/testing-libra
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const listAllAgentIdentities = vi.fn();
+const getMemoryCommand = vi.fn().mockResolvedValue(undefined);
 
 vi.mock("@/app/store/rpc-api", () => ({
     RpcApi: {
         ListAllAgentIdentitiesCommand: (...args: unknown[]) => listAllAgentIdentities(...args),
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2613, opened 2026-08-17
Clamk@claudius: opt-in bundle.export_for_agent_with_history action

GenericAgentX-asaf:
## Summary

Step 5 of #2603 (protocol §3.3). Deliberately a **separate, explicit** action from `bundle.export_for_agent` — a bundle is meant to be shared/reused across agents and machines, while conversation history is private, potentially large, and belongs to exactly one agent. Embedding it in the default export would conflate those two lifetimes/audiences.

Refactored `bundle_export_for_agent_impl`'s body into a shared `build_export_for_agent` helper (bundle+agent resolution, base export, native-memory splice) so the two export paths can't drift on that logic. The with-history variant additionally resolves the agent's sessions via `HistoryService::sessions_for_agent` (a typed core extracted from `list_for_agent`, from Step 4b — #2611, merged) and appends each readable transcript under `history/<provider>/<session_id>.jsonl` in the zip. Always zips (a multi-file, potentially large archive has no sensible inline-JSON shape). An unreadable transcript is warned about and skipped, never fails the whole export.

New RPC: `bundle.export_for_agent_with_history(bundle_id, agent_id)`.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2370 passed, 0 failed, 6 pre-existing ignores
- [x] All 42 pre-existing `bundle.rs` export/import tests still pass (confirms the `build_export_for_agent` refactor is behavior-preserving)
- [x] 3 new tests for the export-with-history path: includes sessions correctly, succeeds with zero sessions for an unlinked agent, warns-and-continues on an unreadable transcript

### DIFF
--- a/.changesets/1786927950-feat-armory-opt-in-bundle-export-for-agent-with-history-acti-1ayu.md
+++ b/.changesets/1786927950-feat-armory-opt-in-bundle-export-for-agent-with-history-acti-1ayu.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): opt-in bundle.export_for_agent_with_history action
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2602, opened 2026-08-16
Clamk@claudius: fix stale db_agents migration-phase docs + add identity/history persistence protocol

GenericAgentX-asaf:
## Summary

Step 1 of a sequenced fix for named agents' conversation history fragmenting across channel/identity UUIDs on every rebuild (root-caused in the included report). This PR only does the cheapest, lowest-risk piece: correcting stale documentation and pinning the real behavior down with a test. No runtime behavior changes.

- Three separate doc comments (`OBJECT_SCHEMA_VERSION`'s v4 entry in `migrations.rs`, `dual_write.rs`'s module doc, `agents.rs`'s module doc) claimed the `db_agent_definitions`/`db_agent_instances` → `db_agents` consolidation was still "reads on the legacy tables" / "old tables authoritative" — false since `agent_def_list()` was quietly flipped to read `db_agents` (discovered during PR #2587's review). Corrected all three to state the real, per-function split precisely.
- Added `agent_def_list_and_agent_def_get_read_different_tables_by_design`, a test that builds a row in each table independently and asserts which function sees which, so the doc comments can't silently drift out of sync with the code again (protocol invariant P4, see below).
- Added `docs/specs/REPORT_AGENT_IDENTITY_HISTORY_FRAGMENTATION_2026_08_16.md` — root-causes the actual bug (PR #2431's `identities_dir()` default flip, 2026-08-06, never reconciled with `ClaudeHistoryAdapter`'s scan list).
- Added `docs/specs/SPEC_AGENT_IDENTITY_HISTORY_PERSISTENCE_PROTOCOL_2026_08_16.md` — synthesizes that root cause with the Mandatory ABF work (PR #2587) into a proposed 5-invariant protocol (P1-P5) and a build sequence. This PR is step 1 of that sequence; steps 2-6 will follow as separate PRs.

## Test plan

- [x] `cargo check -p agentmux-srv` — clean, no new warnings/errors
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::storage::agents::tests::` — 8/8 pass, including the new test
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::storage::` — 282/282 pass, no regressions

PR #2614, opened 2026-08-17
Clamk@claudius: add agent data scope-routing PR checklist

GenericAgentX-asaf:
## Summary

Step 6 of #2603 (protocol §4 step 6) — the final step in this series. Distills `SPEC_AGENT_IDENTITY_HISTORY_PERSISTENCE_PROTOCOL_2026_08_16.md`'s P1-P5 into a short, reviewable checklist for anyone touching CREDENTIAL/DEFINITION/PORTABLE CONFIG/CONVERSATION HISTORY routing.

Written after actually implementing steps 1-5 (#2602, #2605, #2606, #2611, #2613), not purely from the design doc — every item cites the real incident from that work that motivates it (the `wstore`/`id_store` bundle-provisioning bug, the unvalidated spawn-path `entity_id`, the credential-isolation-took-history-with-it regression this whole protocol exists to fix, the history-adapter read path never getting updated, the hardcoded-`"projects"`-subdir bug, the stale migration-phase comment, the two review bots' false claim about Windows junction metadata, and the `ensure_history_link` bug an empirical test caught that type-checking didn't).

Docs-only change, no code.

### DIFF
--- a/.changesets/1786928135-docs-add-agent-data-scope-routing-pr-checklist-m3un.md
+++ b/.changesets/1786928135-docs-add-agent-data-scope-routing-pr-checklist-m3un.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: add agent data scope-routing PR checklist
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #2615, opened 2026-08-17
fix(identity): exportagents resolves provider through the bound bundle, not a drifted agent.provider

AgentY-asaf:
## Summary

Delivery-plan item 7 of #2594 (harness/model decoupling & ABF portability) — the **last remaining item**: `agent_handlers/core.rs:709` (`exportagents`), "weaker case; round-trips a possibly-drifted value through `importagents` into a new definition+bundle." Same "gate vs. actual launch can disagree" risk class #2592/#2596/#2607/#2609/#2610/#2612 fixed.

`exportagents` read `agent.provider` directly for every exported definition. `importagents` (verified while investigating this fix) is internally consistent on its own — it writes the imported value onto the new definition and immediately provisions that same new definition's bundle from it, so there's no separate drift bug on the import side. But if the SOURCE export itself already carried a drifted value, importing it would seed the newly-created agent's bundle with the wrong provider from the very start — the export is the only place this needed fixing.

## Fix

`id_store.resolve_effective_provider_id(&agent)` (already used by #2592/#2607/#2612) replaces the direct `agent.provider` read.

## Test plan

- [x] Regression test seeds an agent with a drifted `.provider` (`"codex"`) against a bundle with the real provider (`"claude"`) and asserts the exported entry carries `"claude"`.
- [x] Verified it fails against the pre-fix code (temporarily reverted the fix inline, ran the test — `left: "codex", right: "claude"` — then restored it).
- [x] `cargo check -p agentmux-srv` — clean, no new warnings/errors.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv server::agent_handlers::` — 31/31 pass, including the new test.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::storage::` — 282/282 pass.

This is the **last unchecked item** on issue #2594's delivery plan — once this merges, the whole harness/model decoupling & ABF portability plan is complete.

### DIFF
--- a/.changesets/1786928533-fix-identity-exportagents-resolves-provider-through-the-boun-7mql.md
+++ b/.changesets/1786928533-fix-identity-exportagents-resolves-provider-through-the-boun-7mql.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): exportagents resolves provider through the bound bundle, not a drifted agent.provider
--- a/agentmux-srv/src/server/agent_handlers/core.rs
+++ b/agentmux-srv/src/server/agent_handlers/core.rs
@@ -672,10 +672,12 @@ pub fn register(engine: &Arc<WshRpcEngine>, state: &AppState) {
 
     // exportagents — export all agent definitions with content and skills
     let wstore_efa = state.wstore.clone();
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2618, opened 2026-08-17
feat(agent): harness-then-model creation flow + pane tab-strip padding fix

AgentY-asaf:
## Summary

Two related pieces of usability work in the "new agent" (AgentPicker) flow. Not part of issue #2594's delivery plan (that's fully merged) — separate follow-up feature work requested directly in conversation, building on #2594's harness/bundle-resolution lineage.

### Harness-then-model creation flow

The template card grid already picks the **harness** (the CLI — clicking "Claude Code" = the claude harness); this adds an explicit **model** picker once a card is clicked, so creation now reads as harness-first-then-model, matching how the in-session runtime switcher (`AgentRuntimeDropup`/`AgentComposerStrip`) already works.

Frontend-only — no backend RPC change. I investigated adding a `model` field to `agentdefcreatefromtemplate` first, but `Memory.model` (the bundle field that looked like "which model") turned out to already be used to store the **vendor** label (`"anthropic"`/`"openai"`/`"custom"`, via `resolve_effective_vendor`), not an actual model choice. The real model picker writes into per-launch block runtime config (`agent:runtime` meta / `applyRuntimeChange`) instead — a purely client-side concept with no bundle involvement at all. So the model chosen at creation time threads through as a new `LaunchOverrides.model` field into the first launch's runtime config, via a new pure, testable `resolveInitialRuntimeConfig` helper (`agent-launch-env.ts`, same pattern as `resolveEffectiveLaunchProvider`).

**Incidental fix along the way**: `launchAgentDefinition` never set `agent:runtime` meta on *any* fresh launch before this — every new agent silently got `DEFAULT_RUNTIME_CONFIG`'s hardcoded `"sonnet"` regardless of harness. Now seeded from the harness's own `default: true` model when no explicit choice is given, for every launch, not just template-creates.

Explanatory copy: inline hint text in `AgentCreateFromTemplateModal.tsx` near the model field, plus a short subtitle under the picker's "+ New from template" header explaining harness vs. model.

### Pane tab-strip padding fix

`.agent-picker` (the "new agent" pane content) had a flat `padding: var(--space-4)` with no extra top clearance for the floating `PaneTabStrip`, which renders even over the blank/picker tab (`agent-view.tsx`: "the strip stays visible whether the active member is a launched conversation OR a blank/picker tab"). The picker's own top row rendered behind it. Fixed by mirroring the exact `padding-top: calc(... + var(--pane-tab-strip-height, 28px))` pattern `.agent-document` (`_document.scss`) already uses for the same problem.

## Test plan

- [x] `resolveInitialRuntimeConfig`: 5 new unit tests (override wins, provider default wins, `DEFAULT_RUNTIME_CONFIG` fallback ×2, permissionMode/effort pass through unchanged).
- [x] `AgentCreateFromTemplateModal`: 4 new tests (model select defaults to the harness's own default, explanatory hint renders, picked model reaches `onSubmit`, both hidden + empty model sent for a harness with no models list).
- [x] `AgentPicker`: extended the existing `onCreatedAndLaunch` integration test to assert the model threads through to `launchAgentDefinition`'s overrides; added a companion test for the omitted-model case; added a test for the new templates-section hint.
- [x] `npx vitest run app/view/agent/ app/element/` — 1202/1202 pass, 97/97 files, no flakes.

Also includes `docs/reports/REPORT_SESSION_STATUS_ABF_HARNESS_MODEL_2026_08_17.md`, a handoff status doc covering this session's #2594 delivery-plan completion and this follow-up work.

### DIFF
--- a/.changesets/1786951148-fix-agent-pane-tab-strip-no-longer-overlaps-the-new-agent-pi-kgz8.md
+++ b/.changesets/1786951148-fix-agent-pane-tab-strip-no-longer-overlaps-the-new-agent-pi-kgz8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): pane tab-strip no longer overlaps the new-agent picker's top content
--- a/.changesets/1786951149-feat-agent-harness-then-model-creation-flow-pick-a-model-whe-7676.md
+++ b/.changesets/1786951149-feat-agent-harness-then-model-creation-flow-pick-a-model-whe-7676.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): harness-then-model creation flow — pick a model when creating from a template, explanatory copy for harness vs model
--- a/.changesets/1786951774-fix-agent-create-from-template-model-picker-reads-through-ge-bg0z.md
+++ b/.changesets/1786951774-fix-agent-create-from-template-model-picker-reads-through-ge-bg0z.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): create-from-template model picker reads through getProvider, not the raw static catalog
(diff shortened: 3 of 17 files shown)


a5af:
@codex review

PR #2621, opened 2026-08-17
feat(agent): filter/search bar atop MyAgentsList

AgentY-asaf:
## Summary

Adds a filter/search box atop the AgentPicker's My Agents list so a user with
several agents can quickly narrow to the one they're interested in, per
user request. Not part of issue #2594's delivery plan (that's fully shipped)
— this is a follow-up UX request from the same session.

- New `AgentPickerFilterBar` component: plain text input + clear button,
  no autofocus (see Q2 below).
- `MyAgentsList` gains an optional `nameFilter` prop: case-insensitive
  substring match against `instance_name || definition_name`, client-side
  over the already-fetched page.
- Fetch limit bumps from the default 20 to the backend's hard cap of 100
  only while a filter query is active (one refetch on the empty→non-empty
  transition, not per keystroke), so the filter searches the full
  available set, not just the first page. Reverts to 20 when cleared.
- Distinct "No agents match ..." empty state, kept separate from the
  existing generic/identity-filtered empty copy.
- Template grid (the "+ New from template" tier) is intentionally
  untouched — spec's Q1, confirmed by the human operator: the ask was
  specifically to find an *existing agent*.

Full design rationale + two open questions (both now resolved) in
`docs/specs/SPEC_AGENT_PICKER_FILTER_SEARCH_2026_08_17.md`.

## Test plan

- [x] New regression tests written first, verified to fail against the
      unfixed component before implementing.
- [x] `MyAgentsList.test.tsx` — 23/23 passing (9 new cases).
- [x] `AgentPicker.test.tsx` — 19/19 passing (5 new cases).
- [x] Full `frontend/app/view/agent` suite: 94 files / 1197 tests passing,
      no collateral regressions.
- [x] `stylelint` on the touched scss — 0 new violations.


a5af:
@codex review

PR #2623, opened 2026-08-17
feat(jekt): TIER=sensitive no longer stops work for a cryptographically verified sender

AgentX-asaf:
## Summary

Repo-owner-confirmed live: Manoz@Area54 received a ReAgent WAN PR review whose Ed25519 signature genuinely verified (`SIG=verified`), but got force-escalated to `TIER=sensitive` anyway by a keyword match in the review text (this repo's own recent credential/agent-binding hardening work is exactly the kind of PR ReAgent legitimately reviews and discusses in those terms) — and stopped work waiting on a human, for a message that was, in fact, exactly who it claimed to be.

- Adds a new, server-computed `ESCALATE=required|none` marker field (only rendered when `TIER=sensitive`) so the receiving agent doesn't have to infer stop-or-not by cross-referencing `TRUST=`/`SIG=` itself.
- `ESCALATE=none` fires only when the sender is cryptographically verified on the delivery tier it arrived on (`TRUST=host-verified`, `TRUST=lan-verified`, or WAN `SIG=verified`) — the `TIER=sensitive` tag is retained for visibility (repo owner: "we want to retain the tag for visual indication"), but no human confirmation is required.
- The three active-forgery signals (`TRUST=unverified`, `SIG=invalid`, a failed LAN signature) stay unconditionally `ESCALATE=required` — this is guaranteed by construction, not just convention: each is keyed on the exact same trust field reading `Some(false)` instead of `Some(true)`, so a message can never be both "actively failed" and "verified" on the same field at once.
- `InjectionResponse` gains `requires_stop`, threaded through the sender-echo path (including cross-instance-forwarded hops) so the sender's own pane renders the same STOP-or-tag-only marker the receiver got.

Full spec: `docs/specs/SPEC_JEKT_SENSITIVE_TIER_VERIFIED_SENDER_NO_STOP_2026_08_17.md`. `CLAUDE.md`'s jekt security section is updated to match (new "Does `TIER=sensitive` always STOP?" subsection).

## Test plan

- [x] New/updated unit tests in `agentmux-srv/src/backend/reactive/tests.rs` — 120 passed, 0 failed (was 119; 6 existing tests gained `requires_stop`/`ESCALATE=` assertions, 1 new host-tier test added):
  - WAN reagent-verified + declared-sensitive / keyword-match → `requires_stop == Some(false)`, `ESCALATE=none`, no STOP text
  - LAN lan-verified + declared-sensitive / keyword-match → same
  - Host `sig_verified=true` + keyword match → same (new test, no host-tier equivalent existed before)
  - WAN `SIG=invalid`, LAN failed signature, host `TRUST=unverified` → all still `requires_stop == Some(true)`, `ESCALATE=required` (regression coverage — active-forgery cases are unaffected)
  - Unverified LAN sender + keyword match → still `requires_stop == Some(true)` (relaxation does NOT apply to unproven senders)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::reactive:: server::reactive::` — 120 passed, 0 failed, 0 ignored
- [x] `cargo build -p agentmux-srv --bin agentmux-srv` — clean
- [x] `cargo clippy -p agentmux-srv --bin agentmux-srv --tests` — no new warning categories; `too_many_arguments` on `wrap_jekt_message`/`echo_jekt_to_sender` pre-existed before this change (10/7 and 9/7 respectively) and is incremented by one parameter each, consistent with how prior PRs added `sig_verified`/`reagent_verified`/`lan_verified` one at a time
- [ ] Manual: live jekt round-trip through a verified sender showing `ESCALATE=none` and no stop — not yet done, noted as follow-up


a5af:
@codex review

PR #2624, opened 2026-08-17
fix(layout): shift-drag group resize — correct border direction, raise min pane size to 128px

GenericAgentX-asaf:
## Summary

Three related fixes to Shift+drag group pane resize (PR #2401), found and
fixed live in this session per a user report ("if I shift+drag a pane
border in one direction, some panes along that direction move in the
opposite direction — is that a design flaw or intentional?"):

- **Border-direction bug (design flaw, not intentional).** `computeGroupResizeSizes`
  previously redistributed the resize delta across every sibling in the
  parent regardless of position. A sibling positioned *past* the driven
  pane could get a growth share that pulled its own shared border *toward*
  the driven pane — i.e. opposite the drag — whenever the driven pane had
  siblings on both sides of it. Rewrote the algorithm as two blocks meeting
  exactly at the dragged handle (`beforeBlock`/`afterBlock`), each scaling
  proportionally as a unit. This guarantees the handle border always
  tracks the cursor exactly, and — the actual fix — that every other
  border in the group moves the same direction as the drag, never
  backward. Trade-off: the driven pane's own size is no longer pixel-exact
  once `afterBlock` has more than one member (documented; the invariant
  that matters is the handle border, not the driven node's individual
  width).
- **Minimum pane size raised from 40px to 128px** (`MinNodeSizePx` in
  `layoutResize.ts`) — scoped to the interactive drag path only (plain and
  Shift-group resize, both row and column share the one constant). Several
  other paths that can also produce a pane's on-screen size — split,
  the resize reducer's own flex-weight range check, moveNode,
  minimize-restore, and window-shrink reflow — do not enforce any pixel
  floor today and were deliberately left alone (confirmed via a full-repo
  scan); scope decision made explicitly with the requester rather than
  silently expanding this PR.
- **Follow-on stall bug**, found via live use immediately after raising the
  floor to 128px ("it stops resizing before the 128px, why is that?"):
  `computeGroupResizeSizes` pre-clamped the raw cursor-implied desired size
  to `minNodeSize` before computing the aggregate transfer delta — correct
  under the old single-node model (driven's final size *was* that clamped
  value), but wrong under the new two-block model, where driven only gets
  a *proportional share* of `afterBlock`'s total change. Pre-clamping
  froze the aggregate delta the instant the raw, unshared cursor position
  crossed the floor, well before the block's true combined headroom was
  exhausted — so the pane being dragged stalled above 128px and further
  dragging did nothing. Fixed by deriving the delta from the unclamped
  desired size always; floor enforcement already happens correctly
  per-member inside the block-shrink helper.

Full write-up, worked numeric examples, and the geometric argument for why
the two-block model is direction-safe:
`docs/specs/SPEC_SHIFT_DRAG_GROUP_RESIZE_DIRECTION_FIX_2026_08_17.md`.

## Test plan

- [x] `npx vitest run frontend/layout frontend/app/element` — 169/169 pass
      (12 files), including 3 new/rewritten regression tests: one asserting
      **border positions** (not just sizes) never move opposite the drag
      direction, one exercising the documented driven-size-vs-desired-value
      trade-off, and one specifically for the pre-clamp stall bug (driven
      reaches the true floor even when the raw cursor position implies it
      should go far below it, and stays there on further dragging).
- [x] `npx tsc --noEmit` — clean (2 pre-existing, unrelated errors in
      `armory-view.tsx`/`warden-view.tsx`, untouched by this change).
- [x] Live verification in a real `task dev` window this session: confirmed
      the border-direction fix, then live-caught and fixed the stall bug
      via direct interactive dragging — not just a code read.

### DIFF
--- a/.changesets/1786983766-fix-layout-shift-drag-group-resize-no-longer-moves-borders-o-rx2n.md
+++ b/.changesets/1786983766-fix-layout-shift-drag-group-resize-no-longer-moves-borders-o-rx2n.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): shift-drag group resize no longer moves borders opposite the drag direction
--- a/.changesets/1786985909-fix-layout-raise-minimum-pane-size-from-40px-to-128px-m1i3.md
+++ b/.changesets/1786985909-fix-layout-raise-minimum-pane-size-from-40px-to-128px-m1i3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): raise minimum pane size from 40px to 128px
--- a/docs/specs/SPEC_SHIFT_DRAG_GROUP_RESIZE_DIRECTION_FIX_2026_08_17.md
+++ b/docs/specs/SPEC_SHIFT_DRAG_GROUP_RESIZE_DIRECTION_FIX_2026_08_17.md
@@ -208,6 +208,40 @@ Degenerate cases:
   runtime in this sandbox); `task dev` started so the requester can verify
   directly.
 
+## 6.1 Follow-on bug found via live use: premature stop before the floor
+
+User report, same session, after raising `MinNodeSizePx` to 128: "if I
+shift+drag a pane, it stops resizing before the 128px, why is that?"
+
+Root cause: `computeGroupResizeSizes` computed
+`clampedDesired = Math.max(drivenDesiredSize, minNodeSize)` and derived
+`totalDelta` from that **clamped** value, before the block split. That was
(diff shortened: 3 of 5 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in fc6b63855 — good catch. \`shrinkBlockBy\` now excludes members already at/under \`minNodeSize\` from its pool up front instead of computing \`cur - minNodeSize\` on them (which went negative and enlarged them). Added a regression test reproducing the exact scenario, hand-traced against the unfixed code first to confirm it actually broke conservation the way described.

@codex review

PR #2625, opened 2026-08-17
fix(agent): HOST/SANDBOX runtime tag reads as a small tag, not full-size text

AgentY-asaf:
## Summary

Follow-up to the earlier MyAgentsList/composer-strip HOST/SANDBOX consistency
fix — the `size="tag"` variant of `RuntimeBadge` never had its own
`font-size`/`padding`, unlike `--sm`/`--md`. It silently inherited whatever
text size surrounded it (composer strip / My Agents row), so it rendered as
normal-size text instead of the small, compact tag it's meant to be.

Gives `.runtime-badge--tag` an explicit `font-size: 9px; padding: 1px 4px;`
— one step below `--sm`'s existing 10px, following this file's own
sm(10px)/md(11px) scale downward for the more minimal tag variant.

## Test plan

- [x] `stylelint` on the touched file — 0 new violations (11 pre-existing,
      unrelated hex-color violations elsewhere in the same file).
- [x] Full `frontend/app/view/agent` suite: 94 files / 1199 tests passing.
- [ ] Not yet visually confirmed in a running instance.

### DIFF
--- a/.changesets/1786993300-fix-agent-host-sandbox-runtime-tag-reads-as-a-small-tag-not--0zn4.md
+++ b/.changesets/1786993300-fix-agent-host-sandbox-runtime-tag-reads-as-a-small-tag-not--0zn4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): HOST/SANDBOX runtime tag reads as a small tag, not full-size text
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2626, opened 2026-08-17
Manoz@Area54 fix(cef): auto-fetch patched Windows CEF runtime when missing

GenericAgentX-asaf:
`task package`/`task dev` silently fell back to the stock (no-codec) CEF whenever `~/cef-build` wasn't set up locally, with only an advisory warning. `docs/cef-build/build-patched-cef-windows.md` documents a ~3-6 hour local Chromium compile to fix it, but that build is already published as a GitHub release on `agentmuxai/cef` and only needs to happen once per CEF version, ever.

`scripts/cef-build/fetch-patched-cef-windows.sh` downloads and installs that release into the standard `cef-build` location when missing, wired into `Taskfile.yml`'s `bundle:windows` tier-2 check right before it would fall through to the stock tier. Non-fatal on any failure (gh missing/unauthenticated, network) — falls back to the existing advisory-warning behavior exactly as before.

**Verified end-to-end:**
- No-op when already present
- Full download+extract+install when missing (caught and fixed a real bug in the process: `mktemp -d`'s Unix-style `/tmp/...` path isn't resolvable by `pwsh`'s `Expand-Archive` — fixed with `cygpath -w`)
- A full portable rebuild now bundles the patched 288MB `libcef.dll` with no `STOCK (NO-CODEC)` warning (confirmed via byte-for-byte size match against the release asset)

### DIFF
--- a/.changesets/1786994163-fix-cef-auto-fetch-patched-windows-cef-runtime-when-missing-txxw.md
+++ b/.changesets/1786994163-fix-cef-auto-fetch-patched-windows-cef-runtime-when-missing-txxw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): auto-fetch patched Windows CEF runtime when missing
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2629, opened 2026-08-17
Manoz@Area54 fix(agent): eliminate Agent Picker loading flicker with a combined reveal gate

GenericAgentX-asaf:
## Problem

Opening a new agent pane (the picker view) had a visible flicker/layout-shift: bottom content painted first, then top content arrived and pushed it down.

## Root cause

Two independent, uncoordinated async data sources fed `AgentPicker`:
1. `useAgentDefinitions()` — no loading/empty distinction. `agents()` starts at `[]`, indistinguishable from "genuinely zero definitions" — so `<Show when={agents().length > 0}>` flashed the "No definitions configured" empty state on *every* picker mount for users who actually have agents, before the first RPC response arrived.
2. `MyAgentsList`'s own `ListRecentSessionsCommand` fetch — rendered an empty `<ul>` with no loading indicator while in flight, and was nested *inside* gate #1 so it couldn't even start until that one resolved.

Each reflowed independently — the reported flicker.

## Fix

- `useAgentDefinitions()` now returns `[agents, loading]` instead of just the accessor. Updated all 5 other call sites (`agent-view.tsx`, `bundle-summary.tsx`, `agent-identity-links-panel.tsx`, `identity-accounts-tab.tsx` ×2) + 2 test mocks that broke from the signature change — verified via `tsc --noEmit` (only 2 pre-existing, unrelated errors remain, confirmed present on `main` too).
- `AgentPicker` combines `definitionsLoading` with a new `onFirstLoad` callback from `MyAgentsList` into one `pickerReady` gate, held behind a single `BrainSpinner` overlay — same hold-then-fade pattern already used by `agent-view.tsx`'s pane-level loading overlay. One reveal, no partial states.

Manually verified live via `task dev` — flicker is gone.

### DIFF
--- a/.changesets/1786998258-fix-agent-eliminate-agent-picker-loading-flicker-with-a-comb-1kiy.md
+++ b/.changesets/1786998258-fix-agent-eliminate-agent-picker-loading-flicker-with-a-comb-1kiy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): eliminate Agent Picker loading flicker with a combined reveal gate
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #2630, opened 2026-08-17
feat(agent-pane): answered questions render as user input; invert user-input surface

GenericAgentX-asaf:
## Summary
- An answered `AskUserQuestion` now renders using the same `.agent-user-message` treatment as a typed message once it scrolls into history, instead of the generic collapsed tool row — it substantively *is* user input. Auto-timeout answers keep a small muted note so a 30s fallback isn't misrepresented as something the user typed. Legacy transcripts (no `answerText`) safely fall back to the old rendering.
- `.agent-user-message`'s colors are inverted to a true negative of normal output — background = the theme's normal text color, text = the theme's normal background color — using tokens every theme already defines, so it inverts correctly across all 13 themes with no new per-theme values.

See `docs/specs/SPEC_ASK_USER_QUESTION_HISTORY_STYLING_2026_08_17.md` for the full design.

## Test plan
- [x] `npx stylelint` on the changed SCSS — no new errors (31 pre-existing, unrelated)
- [x] `ToolBlock.test.tsx` (5 new cases) + `useAgentQuestions.test.ts` (1 new case) + `UserMessageBlock.test.tsx` — 33/33 + 27/27 passing
- [ ] Visual check in `task dev` across a few themes


a5af:
@codex review

PR #2631, opened 2026-08-17
fix(launcher): stop inherited MSYS stdio from breaking CEF host CREATE_SUSPENDED spawn

AgentX-asaf:
## Summary

`agentmux-launcher/src/host_spawn.rs` spawns the CEF host with `CREATE_SUSPENDED` but inherited stdio from its own parent by default. When the launcher is itself spawned from deep inside an MSYS/Git-Bash process chain (e.g. an agent building and launching a fresh portable from its own Bash tool call — exactly what surfaced this), those inherited handles make `CreateProcess` fail outright with `ERROR_NOT_SUPPORTED` (os error 50): no window, no log, no data directory, instant exit with nothing to debug from.

Full root-cause writeup, including everything ruled out along the way (Job Object UI restrictions, desktop/window-station access, process integrity level, and `CREATE_SUSPENDED` itself — none of which were the cause): `docs/retro/retro-cef-host-spawn-fails-under-inherited-nonstandard-stdio-2026-08-17.md`.

**Fix:** give `host_cmd` explicit `Stdio::null()` for all three streams instead of inheriting. `Stdio::piped()` (matching `srv_spawner`'s already-working pattern for its own `CREATE_SUSPENDED` spawn) was considered and rejected — the host has no stdout/stderr protocol the launcher parses or consumes (unlike srv's ESTART readiness signal + log forwarding), so piping without ever reading would trade one bug for a latent pipe-buffer-fill hang risk.

## Test plan

- [x] `cargo build -p agentmux-launcher` — clean
- [x] Built a portable from the **pre-fix** commit, launched via plain `./agentmux.exe` from a Bash tool call (the exact failing scenario) — reproduced instantly: exit code 1, empty stdout/stderr, no data dir created, launcher log confirms `failed to spawn CEF host: The request is not supported. (os error 50)`
- [x] Confirmed the same binary launches cleanly when given real Win32 file handles instead of inherited MSYS ones (`Start-Process -RedirectStandardOutput/-RedirectStandardError`) — isolated the exact trigger before touching any code
- [x] Built a portable from the **post-fix** commit, launched via the identical plain `./agentmux.exe` method — succeeded end-to-end: `spawned CEF host pid=... (suspended)` → `Job Object assigned` → `[ipc] host registered (session=1)` → ran → `CEF host exited cleanly (code 0)` → `launcher exiting with code 0`
- [x] Verified no unintended process/data-dir interference: killed all test-instance processes by exact PID (never by image name), confirmed the 10 shared/live production instance processes on the same machine were untouched throughout

No unit tests added — `spawn_host_supervised` needs a real Windows process/Job-Object/thread-resume environment, which is exactly why this was verified by building and launching real portables rather than a unit test (see existing precedent: `host_spawn.rs`'s only existing tests are the Unix-only process-group ones a few hundred lines down, for the same reason).

### DIFF
--- a/.changesets/1786998817-fix-launcher-stop-inherited-msys-stdio-from-breaking-cef-hos-195k.md
+++ b/.changesets/1786998817-fix-launcher-stop-inherited-msys-stdio-from-breaking-cef-hos-195k.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(launcher): stop inherited MSYS stdio from breaking CEF host CREATE_SUSPENDED spawn
--- a/agentmux-launcher/src/host_spawn.rs
+++ b/agentmux-launcher/src/host_spawn.rs
@@ -47,6 +47,23 @@ pub(crate) fn spawn_host_supervised(
         .env("AGENTMUX_INSTANCE_ID", &srv.instance_id)
         .envs(host_env.iter().cloned())
         .env("AGENTMUX_LAUNCHER_PIPE", pipe_path)
+        // Explicit stdio instead of inheriting the launcher's own —
+        // `CreateProcess(..., CREATE_SUSPENDED, ...)` for this GUI-subsystem
+        // target fails outright with ERROR_NOT_SUPPORTED (os error 50) when
+        // it inherits stdio handles that trace back through an MSYS/Git-Bash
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2632, opened 2026-08-17
fix(identity): agent-account links survive a version/channel switch

### LINKED ISSUE #2627: fix(identity): account/identity-link bindings reset on every non-stable channel (version bump = lost credentials + orphaned history)
## Summary

A named, identity-bound agent's OAuth account binding (`db_agent_identity_links`/`db_accounts`) is
per-channel-isolated by default on every non-`"stable"` channel — i.e. every local/dev/portable build,
which means every version bump. Closing an agent's session and reopening it from "My Agents" after a
version upgrade gets refused with "no credentials — bind an account in the Armory," not a graceful resume.
Re-authenticating mints a fresh, disconnected binding, orphaning the prior conversation history.

Full root-cause writeup, code citations, and recommended fix:
`docs/specs/REPORT_HISTORY_CONTINUITY_ACROSS_VERSION_UPGRADE_2026_08_17.md`.

## Root cause (short version)

`isolated_auth_reason()` (`agentmux-common/src/data_paths.rs`) defaults to isolated for every non-`"stable"`
channel. This flag drives both `identities_dir()` (credential dirs — intentionally per-channel for Armory
testing) AND `registry::resolve_shared_store_path()` (`agentmux-srv/src/registry/paths.rs`), which decides
whether `id_store` — the store `db_agent_identity_links`/`db_accounts` live in — resolves to the global
`shared/store.db` or a per-channel `identity-store.db`. The second consumer was never split out the way
conversation history storage already was (#2605) — `db_agent_identity_links`/`db_accounts` are DEFINITION-
adjacent binding pointers, not raw credential material, and per `migrations.rs`'s own doc comment are
supposed to "survive across channels/versions." They don't today.

The relaunch path (`inject_identity_env_with_broker`,
`agentmux-srv/src/identity/resolver/inject.rs:231`) looks up the binding via `id_store`, finds nothing on a
fresh channel, and refuses the spawn (`SpawnGateError::MissingCredentials`,`inject.rs:605`) before ever
reaching the already-fixed, always-global `identity_history_dir()` machinery from #2605/#2606/#2611.

## Recommended fix

Same pattern as #2605's CREDENTIAL-vs-CONVERSATION-HISTORY split, one level up: route
`db_agent_identity_links`/`db_accounts` through a resolver that's always global, regardless of
`isolated_auth_enabled()` — mirroring `identity_history_dir()`'s unconditional resolution. Add a regression
test asserting these tables' reads/writes never vary by channel, mirroring the existing
`identities_dir_is_shared_on_stable_channel`-style tests. See the report's §4 for the full proposal and §4/§6
for sequencing/scope notes.

## Related

- Issue #2603 (closed) — fixed conversation-history *reachability* once a conversation exists; this issue is
  one step earlier in the same pipeline (the binding lookup that gates whether a spawn even proceeds).
- `docs/specs/SPEC_AGENT_IDENTITY_HISTORY_PERSISTENCE_PROTOCOL_2026_08_16.md`, 
  `docs/specs/CHECKLIST_AGENT_DATA_SCOPE_ROUTING_2026_08_17.md` — same taxonomy/checklist this fix should
  follow.

AgentY-asaf:
## Summary

Fixes #2627 — closes issue.

Root cause (`docs/specs/REPORT_HISTORY_CONTINUITY_ACROSS_VERSION_UPGRADE_2026_08_17.md`,
independently re-verified against live code): `db_agent_identity_links`/`db_accounts` live
in a store whose location flips between global and per-channel based on
`isolated_auth_enabled()`, which defaults to isolated on every non-`"stable"` channel —
i.e. every local/dev/portable build, the exact case a version bump produces. Reopening a
named agent after a version switch finds no binding in the fresh, empty per-channel store
and refuses the spawn ("no credentials — bind an account"), even though the binding still
exists in whichever channel originally created it.

Design (`docs/specs/SPEC_IDENTITY_STORE_SPLIT_2026_08_17.md`, human-operator-approved
"most robust" approach — split by permanent-vs-disposable intent, not by ambient channel):

- New `Store::open_identity_store` / `run_identity_store_schema` /
  `registry::resolve_identity_store_path` — a store deliberately never gated by
  `isolated_auth_enabled()`, structurally (not just conventionally) impossible to
  accidentally isolate — mirrors the pattern `identity_history_dir()` already uses for
  conversation transcripts. `db_agent_identity_links` has no `account_id` FK here (same
  cross-DB-FK reasoning `run_shared_store_schema` already uses for its
  `db_agent_definitions` FK).
- Every agent-identity-link call site rewired from `id_store` to this new store —
  including the actual relaunch/spawn path (`inject_identity_env_with_broker` via
  `resolve_bindings_for_instance`), not just the explicit link-management RPCs. This is
  the piece that actually fixes the reported bug.
- `deleteidentityaccount`'s cascade-delete fixed: it relied on `identity_delete`'s
  same-connection cascade, which no longer reaches links now that they live in a
  different physical file. Added `agent_identity_unlink_by_account` for the
  `identity_store` side, merged both outcomes into the existing notification logic.
- New migration (`0022_identity_store_links_backfill`) backfills existing links from
  every reachable store — current channel, the true global store, and every sibling
  channel's both `objects.db` and isolated `identity-store.db` — into the new store.
  **Without this, shipping the store split alone would regress every existing
  non-stable-channel install** (a working agent's binding would become invisible)
  instead of fixing it.

## Scope

Links only. Bundles/drones/muxbus/native-memory/cron jobs share the same new store's
schema (already in its DDL) but call-site rewiring is deferred to fast-follow PRs (spec
§6 step 1b) to keep this diff reviewable. Accounts (`db_accounts`) stay on the existing
per-channel-isolatable path for now — they have one genuine isolation need (Armory's
disposable test-account-deletion flow) needing its own explicit creation-time scope
split, proposed in the spec but not yet implemented.

## Test plan

- [x] New regression tests, written to prove the actual mechanism, not just exercise
      the code: `registry::paths` (identity-store path never varies by
      channel/isolation; explicitly diverges from `resolve_shared_store_path` under
      isolation — caught and fixed a real test bug of my own along the way, see the
      `identity_store_path_is_sibling_of_shared_store_path` history).
- [x] `backend::storage::identities` — a link written by one process is visible to a
      later, fully independent `Store` open at the same path (the literal close/reopen-
      across-a-version scenario); a link write doesn't require a matching `db_accounts`
      row in this store (confirms the FK removal is actually exercised, not just
      declared).
- [x] `migrations::m0022_identity_store_links_backfill` — backfills a link that only
      ever existed in a sibling channel's *isolated* `identity-store.db` (the literal
      broken-today state, not a synthetic one); idempotent on a second run; never
      mutates a source store.
- [x] Full `agentmux-srv` test suite: 2393 tests passing, 6 pre-existing ignored, 0
      failed — no regressions from the ~15-call-site rewiring.
- [x] `cargo clippy --tests` — zero new warnings in every touched file (verified via
      diff, not just visual scan).
- [x] Merged latest `main` (release v0.55.12) into the branch before pushing; still
      compiles and the full suite still passes post-merge.


a5af:
@codex review

AgentY-asaf:
Two rounds of review fixes pushed:

**Round 1** (reagentx P1/P2): the migration's "true global store" lookup was using the wrong (isolation-aware) field, silently skipping the real global store on every non-`"stable"` channel; and conflicting links across sources resolved non-deterministically. Both fixed, both verified against a test confirmed failing against the pre-fix code first.

**Round 2** (reagentx P0/P1, re-review): the more important one — the link resolved via `identity_store` but the *account* it pointed to could still only exist in the per-channel-isolatable `id_store`, so the spawn was still refused end-to-end. Fixed by adding `db_accounts` to `identity_store` as a read-through fallback mirror (schema v2) — `id_store` stays the authoritative primary (Armory's test-account isolation is unaffected), the mirror is only consulted when the primary lookup misses. `resolve_account()` in `inject.rs` does the fallback read; the status-update write-back now targets whichever store the account actually came from. The migration now backfills accounts alongside links. Also fixed two call sites (`history.ListForAgent`, `bundle.export_for_agent_with_history`) that were still passing `id_store` into `HistoryService`, missed in the original sweep since they're history/export code, not identity code.

Updated summary of what's now true: **accounts remain primarily on `id_store`** (unchanged, preserves Armory's isolation), but a version/channel switch now resolves both the link *and* the account end-to-end via the fallback mirror, for any account that existed before the switch (the migration backfills it). A brand-new account created and linked entirely within one isolated channel still has nothing to fall back to if reopened on a different channel — that's the same disclosed residual gap as before, requiring the deferred per-account `disposable_test` scope split (spec §3.2).

AgentY-asaf:
Addressed round-3 review findings (review 4954613845):

- **P0**: `identity.account.upsert` (and every other live account-write path — `account.key.verify`, both OAuth persist flows, and `deleteidentityaccount`'s cleanup) now dual-writes the `identity_store` fallback mirror via a new `Store::identity_upsert_with_mirror` helper, not just the one-time migration backfill. Without this, an account created or updated after the migration ran would still break on its own next channel switch — the exact bug this PR exists to fix, just shifted to "after" instead of "before" the fix shipped.
- **P1** x2: `identity_self_accounts_impl` and `identity_account_validate_stored_impl` now use `resolve_account`'s `id_store` → `identity_store` fallback, matching the spawn path.

Added a regression test (`inject_resolves_a_freshly_created_account_after_a_channel_switch_when_written_via_the_mirror_helper`) that writes an account through the live `identity_upsert_with_mirror` helper (not migration backfill) and confirms it resolves after a simulated channel switch — verified it reproduces the exact `MissingCredentials` error when the mirror write is skipped, and passes with the fix.

Full suite: 2398 passed, 0 failed.

PR #2628, opened 2026-08-17
chore: release v0.55.12

GenericAgentX-asaf:
## Summary

Release PR consuming all 9 pending changesets, forced to a **patch** bump
(`--as patch`) per explicit request — overrides one queued minor-level
changeset (`feat(agent): filter/search bar atop MyAgentsList`), which ships
under this patch version instead of triggering a minor bump.

`0.55.11 -> 0.55.12`

### Changes shipping in this release

- docs: add CI completion notifications spec (agentmux-cloud PR #48)
- fix(layout): shift-drag group resize no longer moves borders opposite the drag direction
- feat(agent): filter/search bar atop MyAgentsList — quickly narrow to an existing agent by name
- fix(agent): My Agents runtime badge matches composer-strip HOST/SANDBOX tag styling
- fix(agent-pane): hold the brain-spinner overlay until launch-flow's auth check resolves, not just history paint
- fix(layout): raise minimum pane size from 40px to 128px
- feat(jekt): TIER=sensitive no longer stops work for a cryptographically verified sender
- fix(agent): HOST/SANDBOX runtime tag reads as a small tag, not full-size text
- fix(cef): auto-fetch patched Windows CEF runtime when missing

Generated via `task release:patch` (`scripts/release.sh --as patch`) — all
five version-consistency locations (`VERSION_HISTORY.md`, `package.json`,
`Cargo.toml`, `Cargo.lock`, `package-lock.json`) agree at 0.55.12, verified
by the script itself.

## Test plan

- [x] `task release:patch` completed cleanly, no errors, script's own
      post-bump consistency check passed
- [x] `git diff --staged` reviewed before commit — only version files,
      `VERSION_HISTORY.md`, and the 9 consumed changeset deletions
- [x] No source changes in this PR — pure version/changelog bump, so no
      additional test run beyond what already gated each underlying PR

PR #2633, opened 2026-08-17
feat(widgets): parent widgets with grouped submenus — Messengers

Agent1-asaf:
## Summary

Implements `docs/specs/SPEC_WIDGET_BAR_PARENT_SUBMENUS_2026_08_12.md` — a "parent widget" that expands a submenu of child widgets instead of opening a pane, matching the app's existing right-click submenu feel. First concrete use: a **Messengers** parent grouping the 5 existing standalone messenger widgets (Discord, Slack, Telegram, WhatsApp, Teams).

- `WidgetConfigType` gains an optional `children: Vec<String>` (Rust) / `children?: string[]` (TS). A non-empty `children` list makes a widget a "parent"; its own `blockdef` goes unused. Load-time validation warns (not hard-fails) if a widget has neither `blockdef.meta` nor `children`, or both.
- `widgets.json`: new `defwidget@messengers` parent (`children: [discord, slack, telegram, whatsapp, teams]`); the 5 messenger entries gain `display:hidden: true`.
- `getGroupedChildKeys()` (new, in `action-widgets-config.ts`) excludes grouped children from both `getPinnedWidgets()` and `getMoreWidgets()`, and strips a stale bare child short-name out of a pre-existing `widget:pinned` array.
- **Case A (pinned in the bar):** clicking a parent slot toggles a new `PinnedWidgetFlyout` (bottom-start placement, modeled on `MoreDropdown`); hovering opens it via the shared `createSubmenuHover` 90ms-delay/safe-triangle-close primitive. Right-click is scoped to group-level actions ("Pin/Unpin group from bar") instead of the leaf per-pane menu.
- **Case B (a row inside More):** the same `PinnedWidgetFlyout` renders sideways (`right-start`, flips near the edge) off a parent row in `more-dropdown.tsx`, with a trailing chevron — structurally the same hover-intent shape `flyoutmenu.tsx`'s own `SubMenu` already uses.
- `createPeerRegistry()` (force-close a sibling submenu when a new one opens) is hoisted out of `flyoutmenu.tsx` into `submenu-hover.ts` so both `FlyoutMenu` and the new pinned-bar/More-dropdown parent rows share one implementation — kept framework-agnostic (returns an unregister function; callers own their own `onCleanup`).
- Non-goals kept out per spec: nested groups, per-child "promote to standalone pinned," and a UI for authoring new groups.

## Test plan

- [x] `cargo test -p agentmux-srv wconfig` — 40/40 pass, including 2 new tests (`test_build_default_config_loads_messengers_group`, `test_validate_widget_configs_does_not_panic`)
- [x] `cargo check -p agentmux-srv` and `task build:backend` (release profile) — both clean, no new warnings
- [x] `npx vitest run frontend/app/window frontend/app/element frontend/app/util` — 67/67 pass, including 10 new tests in `action-widgets-config.test.ts` (grouped-children exclusion, parent-in-exactly-one-list, stale-pinned-child stripped)
- [x] `npx tsc --noEmit` — clean on every touched file (2 pre-existing unrelated errors in `armory-view.tsx`/`warden-view.tsx` confirmed present on `main` via `git stash`, not introduced here)
- [x] `task build:frontend` — production Vite build succeeds
- [ ] Manual verification in a running `task dev` instance (hover-intent timing, safe-triangle close, Case A/B visual placement) — not done in this session; recommend a reviewer click through both placements before merge

### DIFF
--- a/.changesets/1786999034-feat-widgets-parent-widgets-with-grouped-submenus-messengers-jpel.md
+++ b/.changesets/1786999034-feat-widgets-parent-widgets-with-grouped-submenus-messengers-jpel.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(widgets): parent widgets with grouped submenus — Messengers
--- a/agentmux-srv/src/backend/wconfig/loader.rs
+++ b/agentmux-srv/src/backend/wconfig/loader.rs
@@ -23,6 +23,7 @@ pub fn build_default_config() -> FullConfigType {
 
     match serde_json::from_str::<HashMap<String, WidgetConfigType>>(WIDGETS_JSON) {
         Ok(widgets) => {
+            validate_widget_configs(&widgets);
             config.widgets = widgets;
         }
         Err(e) => {
@@ -31,6 +32,28 @@ pub fn build_default_config() -> FullConfigType {
     }
 
     config
+}
+
+/// Soft validation for widget entries loaded from `widgets.json`: a widget
+/// must be either a leaf (non-empty `blockdef.meta`) or a parent (non-empty
+/// `children`), never neither, never both. Logged as a startup warning, not
+/// a hard failure, matching how other config issues are handled.
+fn validate_widget_configs(widgets: &HashMap<String, WidgetConfigType>) {
+    for (key, widget) in widgets {
+        let has_blockdef = !widget.block_def.meta.is_empty();
+        let has_children = !widget.children.is_empty();
+        if has_blockdef && has_children {
+            eprintln!(
+                "wconfig: widget \"{}\" has both a blockdef and children — children (parent behavior) takes precedence",
+                key
+            );
+        } else if !has_blockdef && !has_children {
+            eprintln!(
+                "wconfig: widget \"{}\" has neither a blockdef nor children — it can't open a pane or expand a submenu",
+                key
+            );
+        }
+    }
 }
 
 // ---- Config loading helpers ----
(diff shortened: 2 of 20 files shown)


a5af:
@codex review

Agent1-asaf:
Pushed a follow-up commit addressing review feedback from manual testing in a running `task dev` instance:

1. **More / pinned-flyout mutual close** — opening "More" while Messengers' pinned flyout was open (or vice versa) left both floating panels open simultaneously. Fixed both directions.
2. **Grouped children leaking into other pickers** — analysis found 3 more surfaces that enumerated `wmap` directly and still showed Discord/Slack/Telegram/WhatsApp/Teams individually instead of only through Messengers:
   - `titlebar-context-menu.tsx`'s "Pin Widgets" submenu (also had a stale duplicate `getPinnedKeys` instead of the shared one — now fixed)
   - `pane-actions.ts`'s "Replace With..." submenu
   - `tabcontent.tsx`'s empty-tab right-click menu

   The latter two now share a new `buildPaneWidgetMenuItems()` helper — grouped children are excluded as flat entries and reachable via a native nested submenu under the parent's label. The Launcher ("+" picker) already hid children via `display:hidden`, but clicking the parent tile tried to open an empty blockdef — it now drills into the group's children with a back breadcrumb + Escape.
3. **Real crash found via `muxlog` while testing #1** — the pinned flyout's `<Show when={openParentKey()}>` callback re-invoked its keyed accessor from several prop closures; one could fire post-unmount, throwing Solid's "stale value from `<Show>`" error. Fixed by capturing the key as a plain string once.

Verified: `cargo test`, `npx tsc --noEmit`, full `vitest` suite (157 passing), `task build:frontend`, and live in a running `task dev` instance via `muxlog`.

Agent1-asaf:
Addressed both findings from the ReAgent review:

- **P1** (`action-widgets.tsx:172`) — `handleParentSlotClick` now goes through a new `openNow()` on `SubmenuHoverController` instead of setting `openParentKey` directly, so the controller's `isOpen` stays truthful and a later mouse-leave starts the normal safe-triangle close-intent instead of no-op'ing forever.
- **P2** (`action-widgets.tsx:274`) — `buildItemMenuItems` no longer offers "Pin to bar" for a grouped child (e.g. right-clicking Discord inside the Messengers flyout); it was a silent no-op since `getPinnedKeys()` always filters grouped children back out.

Added 4 new unit tests for `openNow()` (including one that reproduces the exact P1 failure mode and confirms the fix). Full suite (161 tests), `tsc --noEmit`, and live HMR in the running `task dev` instance all clean.

Agent1-asaf:
Pushed a feature follow-up per additional feedback: users can now pin an individual grouped widget (e.g. Slack) out of its group.

Right-clicking a child inside Messengers' flyout/submenu now supports a real "Pin to bar" — it promotes that widget to a normal standalone entry everywhere (bar, More, the titlebar's "Pin Widgets" checklist, Replace With, the "+" Launcher grid), and it drops out of Messengers' own children list so it's never shown twice. Unpinning it puts it back inside Messengers only, regardless of whether Messengers itself is pinned.

Core mechanism: `getPinnedKeys()` no longer force-strips a grouped child from `widget:pinned` (being listed there is literally how promotion happens). New `getEffectiveGroupedChildKeys(wmap, settings)` — "structurally a child AND not individually pinned" — replaces the old purely-structural check everywhere a picker decides what to hide. `getChildWidgets()` gained an optional `settings` param so a parent's own rendered list excludes anything currently promoted.

Verified: 27 tests in `action-widgets-config.test.ts` (12 new/rewritten for this), full suite (172 passing), `tsc` clean, full frontend build, and live HMR in the running `task dev` instance with no new errors.

Agent1-asaf:
Addressed the P0: rebased onto current \`origin/main\` (was stale — main had moved through the 0.55.12 release (#2628) plus 3 more commits since this branch was cut). Clean rebase, no conflicts. \`package.json\`/\`Cargo.toml\`/lockfiles now correctly read 0.55.12, matching main.

Re-verified post-rebase: \`tsc --noEmit\` clean, full vitest suite (173 passing), \`cargo check\` clean. Force-pushed with \`--force-with-lease\`.

Agent1-asaf:
Pushed a fix for asymmetric hover behavior: hovering onto a pinned parent (e.g. Messengers) already force-closed an open More dropdown, but hovering back onto "More..." never reopened it — only an explicit click did.

The More button now gets its own \`createSubmenuHover\` controller registered as a peer in the same registry pinned parents use, so both directions now open-on-hover and close-the-other symmetrically (same 90ms delay + safe-triangle close as everywhere else). \`MoreDropdown\` also gained the hover-panel plumbing (\`onSubmenuEnter\`/\`onSubmenuLeave\`/\`setSubmenuEl\`) so moving the cursor diagonally from the button into the dropdown doesn't prematurely close it.

Verified: full suite (173 passing), \`tsc\` clean, frontend build clean, live HMR in the running \`task dev\` instance with no new errors.

Agent1-asaf:
Fixed a real regression: right-clicking an entry inside More (or a pinned parent's flyout, e.g. Messengers) was collapsing the menu instead of staying open while you work with pin/unpin.

Root cause: opening the item popover renders it at the cursor position, which makes Chromium re-run hit-testing against the new topmost element and fire a spurious \`mouseleave\` on the dropdown/flyout underneath — even though the cursor never moved. That leave reached the hover controller and started a close-intent. This was introduced for "More" by the symmetric-hover fix, and has been present for pinned parents (Messengers) since Case A shipped originally — the existing outside-click guard (\`.closest(".popover-menu")\`) only covered the click path, never this hover path.

Fix: a new \`isItemMenuOpen()\` check gates every \`onSubmenuLeave\`/\`onTriggerLeave\` call site that could receive this spurious leave — More's own panel, a pinned parent's flyout, and (threaded down as a prop) MoreDropdown's own panel plus its nested Case B parent rows.

Verified: full suite (173 passing), \`tsc\` clean, build clean, live HMR with no new errors.

Agent1-asaf:
Fixed both P1s — great catch, this was a real one.

Root cause: \`getPinnedWidgets()\`/\`getMoreWidgets()\`/\`widgets()\` allocate fresh \`{key, widget}\` object literals on every call, so Solid's reference-diffing \`<For>\` tears down and rebuilds every pinned slot / More row on **any** \`fullConfigAtom\` change — not just when the pinned/more set actually changes. Each row's \`createSubmenuHover\` controller was created (and disposed on teardown) inline inside the \`<For>\` item body, so that churn disposed an *open* controller, firing \`onClose\` and force-closing the flyout — including as a direct consequence of this PR's own "Pin to bar" RPC round-trip.

Fixed by caching each key's controller in a \`Map\` declared **outside** the \`<For>\` (\`parentHoverControllers\` in \`action-widgets.tsx\`, \`parentRowHoverControllers\` in \`more-dropdown.tsx\`) and reusing the existing instance across reference churn instead of recreating it. Controllers now only get disposed at component-level unmount.

Didn't add an automated test for this one — it's specifically about Solid's \`<For>\` reference-diffing interacting with global-store updates, which would need a fairly heavy component-harness (mocking \`fullConfigAtom\`, RPC, etc.) for the coverage it'd add on top of the existing 173-test suite + \`tsc\` + live verification. Verified via \`tsc\` clean, full suite passing, clean build, and live HMR in the running \`task dev\` instance with no new errors.

PR #2635, opened 2026-08-17
feat(reactive): expose ensure-signing-key endpoint for host-tier jekt tests

AgentX-asaf:
## Summary

The only existing way to mint an agent's host-tier jekt signing key (`agent_jekt_key_ensure`) is `agent.open`'s WebSocket (`WshRpcEngine`) path, which also spawns a real provider session — unusable for anything that just wants to exercise host-tier jekt signing/verification without the cost and side effects of a live LLM session.

Surfaced as a real gap while live-verifying PR #2623 (jekt `TIER=sensitive` verified-sender relaxation): the negative/self-declared case was directly testable over the existing `/agentmux/reactive/*` HTTP API, but the positive (genuinely verified sender) case had no reachable path — the key it needs is only ever minted deep inside a full agent launch.

Adds `POST /agentmux/reactive/ensure-signing-key`, gated by the same `X-AuthKey` full-auth boundary as every other route in this group (`register`/`unregister` already let an authenticated caller mint arbitrary `agent_id`→`block_id` bindings — this is the same trust level, not a new category of risk). Idempotent — returns the existing key unchanged if one was already minted for that `agent_id`.

## Test plan

- [x] `cargo build -p agentmux-srv --bin agentmux-srv` — clean
- [x] `cargo clippy -p agentmux-srv --bin agentmux-srv` — no warnings on the new code
- [x] Built a real portable, launched it, called the new endpoint live: `POST /agentmux/reactive/ensure-signing-key {"agent_id":"verify-sender"}` → returned a real base64 key
- [x] Used that key to compute a genuine HMAC-SHA256 jekt signature (matching `agentmux_common::jekt_sign`'s exact algorithm) and sent real signed jekts through the running server's `/agentmux/reactive/inject`, closing the verification gap this endpoint was built to close:

  | sender | signature | `requires_stop` |
  |---|---|---|
  | self-declared | none | `true` (`ESCALATE=required`) |
  | host-verified | valid | `false` (`ESCALATE=none`) |
  | claimed-known, forged | invalid | `true` (`ESCALATE=required`, never relaxed) |

  All three match PR #2623's spec exactly.
- [x] Idempotency: called the endpoint twice for the same `agent_id`, got back the identical key both times.

No new unit test added — `handle_reactive_register`/`handle_reactive_unregister` (the existing routes in this same group) don't have dedicated unit tests either, since they need a full `AppState`; this handler is a thin wrapper around the already-unit-tested `agent_jekt_key_ensure` (`agentmux-srv/src/backend/storage/agent_jekt_keys.rs`), and was verified live above instead.

### DIFF
--- a/.changesets/1787009424-feat-reactive-expose-post-agentmux-reactive-ensure-signing-k-f2se.md
+++ b/.changesets/1787009424-feat-reactive-expose-post-agentmux-reactive-ensure-signing-k-f2se.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(reactive): expose POST /agentmux/reactive/ensure-signing-key for host-tier jekt signing tests
--- a/agentmux-srv/src/server/mod.rs
+++ b/agentmux-srv/src/server/mod.rs
@@ -305,6 +305,10 @@ pub fn build_router(state: AppState) -> Router {
             post(reactive::handle_reactive_unregister),
         )
         .route(
+            "/agentmux/reactive/ensure-signing-key",
+            post(reactive::handle_reactive_ensure_signing_key),
+        )
+        .route(
             "/agentmux/reactive/poller/stats",
             get(reactive::handle_reactive_poller_stats),
         )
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2636, opened 2026-08-17
fix(statusbar): popovers and tooltips paint over native browser-pane HWNDs

Agent2-asaf:
## Summary
- `HostPopover` and `BackendStatus` never adopted the `usePaneOverlay`/`data-pane-overlay`/`Portal` pattern every other status-bar popover already uses, so their dropdowns rendered behind a native browser pane's HWND (the airspace problem) instead of over it. Migrated both to the same Portal + floating-ui positioning pattern as `TokenUsageIndicator`/`TokenBreakdownPopover`, including a dual-ref fix for outside-click-to-close now that the popover content is portaled out of the trigger's own DOM subtree.
- The status bar's `data-tip="..."` hover tooltips had the identical root cause via a different mechanism: a pure-CSS `::after` pseudo-element can never be tagged `data-pane-overlay` or measured for the native-pane clip, since it isn't a real DOM node. Replaced it with `StatusBarTip`, a single delegated listener that renders a real, airspace-clip-aware balloon for any `data-tip` element in the status bar or its popovers — zero changes needed at the ~15 existing call sites, and scoped so it doesn't collide with the editor file-tree's own separate (untouched) copy of the same CSS pattern.

Spec: `docs/specs/SPEC_STATUS_BAR_POPOVER_AIRSPACE_CLIP_2026_08_17.md`

## Test plan
- [x] `npx tsc --noEmit` clean (only 2 pre-existing, unrelated errors in `armory-view.tsx`/`warden-view.tsx`)
- [x] `npx vitest run frontend/app/statusbar` passes (11/11)
- [x] Manually verified in `task dev`: backend-status popover and host-info popover now paint over a browser pane; hover tooltips (CPU/mem/disk stats, backend dot, host chip, LAN toggle) paint over a browser pane; outside-click-to-close still works for both popovers post-portal

### DIFF
--- a/.changesets/1787009771-fix-statusbar-popovers-and-tooltips-paint-over-native-browse-jegz.md
+++ b/.changesets/1787009771-fix-statusbar-popovers-and-tooltips-paint-over-native-browse-jegz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(statusbar): popovers and tooltips paint over native browser-pane HWNDs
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #2637, opened 2026-08-17
Manoz@Area54 fix(agent): remove agent-definitions Import/Export toolbar from the Agent Picker

GenericAgentX-asaf:
Removes `AgentActionBar` — the "↓ Import" / "↑ Export" pair at the bottom of the Agent Picker — per request during live regression testing.

This was a bulk import/export of agent **definitions** as a JSON file, distinct from and unrelated to the per-session history "Export" button in `AgentControlBar`, which is untouched.

## Changes

- Deleted `AgentActionBar.tsx` and its only dependency, `ImportPreviewModal.tsx`
- Deleted `_action-bar.scss` and its `@use` import in `agent-view.scss`
- Removed both `<AgentActionBar />` call sites in `AgentPicker.tsx`
- Removed the now-unnecessary mock in `AgentPicker.test.tsx`
- Removed a stale comment in `agent-view.tsx`
- Backend RPC commands (`ExportAgentDefinitionsCommand` / `ImportAgentDefinitionsCommand`) are **untouched** — scoped to the agent-pane UI entry point only


a5af:
@codex review

PR #2638, opened 2026-08-18
Manoz@Area54 fix(sysinfo): dynamic y-axis scaling, fix 3-metric layout, fix slow first paint

GenericAgentX-asaf:
Three related fixes to the sysinfo widget from live regression testing, plus a research report on why NOT to merge CPU/Mem/Net into one literal combined chart.

## 1. Dynamic y-axis scaling

Network/Disk had a fixed 100 MB/s ceiling that both clipped real spikes above it and wasted most of the chart when idle traffic sat at 0-5 MB/s. Now auto-scales from the visible window's own data (+15% padding, 1 MB/s floor). Memory auto-scales too but is hard-capped at `mem:total` (can't show headroom that doesn't exist). CPU stays fixed 0-100 — auto-scaling an already-bounded percentage would make ordinary noise look like dramatic spikes. Domain rounds via Plot's `nice: true` for human-friendly ticks.

## 2. 3-metric layout fix

"CPU + Mem + Net" (exactly 3 metrics) now stacks as a single column of 3 full-width rows instead of falling into the 2-column grid, which wrapped 3 panels into an awkward 2+1 with a half-empty second row. "All CPU" (can be >3 panels) keeps its existing 2-column layout — scoped to exactly the case that was losing space.

## 3. Slow first-paint fix

`plotData` was a one-time snapshot of `model.dataAtom()` taken at mount, only resynced every `CHART_UPDATE_INTERVAL_MS` (2s) via `setInterval`. If the pane mounted before `loadInitialData()`'s RPC resolved (the common case), the chart sat blank until the next interval tick — up to 2 extra seconds after the real data was already ready. Added an effect scoped via `on()` to only `loadingAtom()` (not `dataAtom()`, or it'd fire on every ~1Hz sample and defeat the throttle) that syncs immediately when the load completes.

## Research report

`docs/reports/REPORT_SYSINFO_COMBINED_CHART_RESEARCH_2026_08_17.md` — covers why a literal single combined chart (dual/triple-axis) was considered and rejected in favor of small multiples + dynamic per-panel scaling, with sources.

### DIFF
--- a/.changesets/1787012397-fix-sysinfo-dynamic-y-axis-scaling-fix-3-metric-layout-fix-s-03d3.md
+++ b/.changesets/1787012397-fix-sysinfo-dynamic-y-axis-scaling-fix-3-metric-layout-fix-s-03d3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(sysinfo): dynamic y-axis scaling, fix 3-metric layout, fix slow first paint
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

Issue #2024, opened 2026-07-08
Armory pane: Brain/Bundle tab decision, full Bundle composition (v2) — Identity->Accounts fold done


AgentY-asaf:
## Background

`specs/PROPOSAL_COMPOSABLE_AGENT_MODEL_2026_06_30.md` lays out a target architecture for the Armory pane where Accounts, Memory, MCP servers, and Skills are all bindable primitives that can be packaged into a Bundle. Follow-up work (#1943/#1946/#1948/#1962, tracked in #1960, closed) shipped the MCP/Skills half of this. This issue originally tracked three pieces of that target architecture not yet shipped in the Armory pane. **Item 1 is now fully resolved (2026-07-13)** — see the update comment below. Items 2 and 3 remain open.

## 1. ~~Fold "Identities" tab into "Accounts"~~ — RESOLVED 2026-07-13

Shipped across Phase 3 (#1624, #2061), Phase 4a-4c (`docs/specs/SPEC_ARMORY_PHASE4_STORAGE_RENAME_COMPLETION_2026_07_12.md`), and Phase 5 (`docs/specs/SPEC_ARMORY_PHASE5_CONSOLIDATION_AND_SKILL_SEEDING_2026_07_13.md`, PR #2140):

- `db_identity_bundles`/`db_identity_bindings` (the tables item 1 originally flagged as "not safe to just hide") are **dropped**. Credential resolution reads only `db_agent_identity_links`/`db_accounts` — confirmed the sole path in `identity/resolver.rs::resolve_bindings_for_instance`.
- The Armory "Identities" tab is **removed** (not just made read-only) — `AgentIdentitiesPanel` deleted, `ArmorySection` no longer has an `"identities"` member.
- The agent-pane's own "Identity" tab (`view: "identity"`) now correctly shows *that specific agent's* linked accounts via a new `AgentIdentityLinksPanel`, closing the data gap that previously made it a dead stub.
- See the update comment below for full detail + verification.

## 2. Fold "Brain" tab into "Bundles" — NOT ACTUALLY DECIDED, needs a product call

[...]

PR #2639, opened 2026-08-18
feat(bundle): Bundle-level MCP Server and Skill references (composable model v2)

AgentY-asaf:
## Summary

Delivery-plan steps 1-3 of `docs/specs/SPEC_BUNDLE_AS_CONTAINER_V2_2026_08_17.md`, scoping GH issue #2024 item 3 ("Bundles packaging Accounts + Memory + MCP + Skills together" — proposed, explicitly deferred to "v2" by the original v1 composable-model spec, never started).

Shipped as one PR rather than split further into schema-only / RPC-only slices: reading `write_agent_config_files` revealed it never consulted `agent.memory_id` (the bound bundle) at all for MCP/skills resolution — a schema-only or RPC-only slice would have been exactly as inert at launch as the bundle's existing (never-consulted) inline `mcp_servers`/`skills` JSON columns already are. This PR is end-to-end functional: a bundle can reference standalone MCP servers/skills via RPC, and those references actually affect what a spawned agent gets.

- **New tables**: `db_bundle_mcp_ref` / `db_bundle_skills_ref` (objects.db v23) — mirror the existing agent-level `db_agent_mcp_ref`/`db_agent_skills_ref` exactly, keyed by `bundle_id`.
- **Store methods**: `bundle_mcp_list/bind/unbind/is_accessible_to`, same for `bundle_skill_*`. `mcp_server_delete`/`skill_delete` now also purge the new bundle ref rows.
- **RPCs**: `mcp.catalog.bind_to_bundle/unbind_from_bundle/list_for_bundle`, same for `skill.catalog.*` — same naming convention and "only global rows may be bound" trust boundary as the existing `*_to_agent` trio (closes the same IDOR class `mcp_server_list_global_for_agent`'s doc comment already documents, applied to `bundle_id` instead of `agent_id`).
- **Launch-time wiring**: new `effective_mcp_servers` + extended `effective_skills` union the agent's bound bundle's referenced servers/skills into the existing agent-ref + global resolution, deduped by id.

**Explicitly deferred** (see the spec doc's "Open questions" section, flagged rather than decided unilaterally):
- Bundle ↔ Account — real FK to a live credential vs. an abstract per-provider requirement; a real security/trust-model decision, not an implementation detail.
- Bundle ↔ Memory (native/learned) — no standalone/referenceable memory primitive exists yet to reference; building one is materially larger scope than this PR.

Frontend Bundle-editor UI (MCP/Skills tabs, reusing the existing `AgentMcpModal`/`AgentSkillsModal` pattern) is the natural next PR — `memory-manager.tsx`'s own hint text already says these are "editable here in a follow-up."

## Test plan

- [x] TDD throughout: every new Store method and the launch-time wiring change has a regression test verified to fail against the unfixed/pre-wiring code before trusting it passes (leak-isolation check, bundle-existence guard, purge-on-delete, bundle-referenced skill/server visible in `effective_skills`/`effective_mcp_servers`, no duplication of globals visible via both paths).
- [x] Full backend suite: 2415 passed, 0 failed (up from 2410 pre-merge baseline + 4 unrelated commits from `main`).
- [x] `cargo check --tests`: clean.
- [x] `cargo clippy --tests`: no new warnings in any touched file.
- [x] Merged latest `main` (4 commits) into the branch before pushing; still compiles and the full suite still passes post-merge.
- [x] Changeset added (`minor`).


a5af:
@codex review

AgentY-asaf:
Addressed round-2 review findings (reviews 4955889703 / 4955983366 / 4956017873):

**P0 x2 — cross-store bundle existence check.** `bundle_mcp_bind`/`bundle_skill_bind` checked bundle existence against `self` (wstore/objects.db), but bundles are authoritatively written through `id_store` (the shared store in a normal production install, per `bundle.rs::register_bundle_upsert`) — `wstore`'s own local `db_bundles` copy is essentially always empty in that case, so every real production bind would fail with "bundle not found." Fixed by threading `id_store` through as an explicit parameter and checking existence there. Dropped the now-cross-store FK to `db_bundles` (SQLite can't FK across two separate database files) — mirrors `db_agent_identity_links.account_id`'s identical no-FK precedent for the same reason.

**P1 — bind_to_bundle had no functional effect.** Correctly flagged: a global server/skill already reaches every agent unconditionally regardless of bundle binding, so any bind `mcp.catalog.bind_to_bundle`/`skill.catalog.bind_to_bundle` could successfully perform had zero effect on a spawned agent. Added `bundle_mcp_upsert_unique`/`bundle_skill_upsert_unique` (+ `mcp.catalog.upsert_for_bundle`/`skill.catalog.upsert_for_bundle` RPCs) — creates a brand-new PRIVATE primitive scoped to one bundle, atomically bound. This closes the gap without reopening the cross-entity IDOR concern `bind_to_bundle`'s "only global" restriction exists to prevent: a brand-new row was never privately owned by anyone else, so there's no theft risk in binding it. Also relaxed `unbind_from_bundle` to allow removing a bundle's own private ref (previously restricted to global-only, which no longer made sense once bundle-owned private refs could exist).

10 new regression tests, each verified to fail against the pre-fix code first: two-store existence-check tests (proving the check now reads `id_store`, and that `self`'s own copy alone is insufficient — directly reproducing the reported "bundle not found" bug), upsert_unique create + duplicate-name-rejection tests, and end-to-end `effective_mcp_servers`/`effective_skills` tests proving a bundle-owned private primitive actually reaches a spawned agent's config.

Full suite: 2436 passed, 0 failed. RPC contract test (frontend↔backend) passing — added TS bindings + gotypes for all new commands.

PR #2640, opened 2026-08-18
fix(reactive): don't evict live cross-channel jekt registry entries

GenericAgentX-asaf:
## Summary

- **Root cause fix**: cross-channel jekt forwarding (`server/reactive.rs`) permanently deleted a channel's shared registry entry on any `success:false` response — including transient ones (target channel up, but the specific agent hadn't finished registering yet). Nothing re-writes the entry except that agent's own one-time registration event, so a single hiccup made the agent unreachable via cross-channel jekt until a restart. Reuses the PID-liveness check `cleanup_stale_shared` already applies at startup (`registry.rs`'s new `is_pid_alive`) at both evict-on-forward-failure call sites (Tier 2a same-channel, Tier 2b cross-channel): only evict when the owning process is confirmed dead.
- Also includes the tool-call-streaming scroll-pin diagnostic instrumentation (`[wave-scroll-shrink]` log) and analysis this branch started with — see the linked docs.

Full writeups:
- `docs/retro/retro-cross-channel-jekt-eviction-2026-08-17.md`
- `docs/analysis/ANALYSIS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_17.md`

## Test plan

- [x] `cargo test -p agentmux-srv reactive::registry` — 17/17 pass, no regressions
- [x] `npx vitest run frontend/app/view/agent/virtualization` — 121/121 pass, no regressions
- [ ] Live cross-channel jekt repro against a currently-unreachable agent, post-restart (needs a coordinated `srv` rebuild/restart — the shared `srv` on this host currently serves multiple live agents, so this wasn't done unilaterally; see retro §"Known gaps")
- [ ] Integration-level regression test for the new PID-alive-guard branch in `server/reactive.rs` (only the underlying `is_pid_alive` primitive has direct unit coverage today) — flagged as a follow-up in the retro

### DIFF
--- a/.changesets/1787013019-fix-reactive-don-t-evict-cross-channel-jekt-registry-entries-mi9o.md
+++ b/.changesets/1787013019-fix-reactive-don-t-evict-cross-channel-jekt-registry-entries-mi9o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(reactive): don't evict cross-channel jekt registry entries whose owning process is still alive
--- a/agentmux-srv/src/backend/reactive/registry.rs
+++ b/agentmux-srv/src/backend/reactive/registry.rs
@@ -302,20 +302,47 @@ fn pid_alive(pid: u32) -> bool {
     sys.process(target).is_some()
 }
 
-/// Public entry point for callers outside this module (`server/reactive.rs`'s
-/// evict-on-forward-failure sites) that need to distinguish "the owning
-/// process is genuinely gone" (safe to evict permanently) from "the process
-/// is alive but this one forward attempt failed" (a transient hiccup — e.g.
-/// the target channel's srv is up but the specific agent hadn't finished its
-/// own registration yet, a momentary lock/timing race, ...). Before this,
-/// `remove_shared`/`remove` were called unconditionally on any `success:false`
-/// response, which is correct for a truly-dead channel but destroys a live
-/// channel's only path to being found again — nothing re-writes the shared
-/// entry except the agent's own one-time registration event, so an evicted
-/// live entry stays unreachable until that agent's controller restarts. See
-/// docs/retro/retro-cross-channel-jekt-eviction-2026-08-17.md.
-pub fn is_pid_alive(pid: u32) -> bool {
-    pid_alive(pid)
+/// Grace period during which a forward failure against a live-process entry
+/// is presumed transient rather than proof the specific agent is gone — see
+/// [`should_evict_on_forward_failure`].
+const FORWARD_FAILURE_GRACE_MS: u64 = 60_000;
+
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both P1s fixed in d50809761 — the `Err(e)` connection-failure branches at Tier 2a and Tier 2b now use the same `is_pid_alive` guard as the `success:false` branches, for exactly the reason you flagged: a connection refusal during the startup-race window is just as transient-plausible as a parsed `success:false` body. `cargo check -p agentmux-srv` clean.

GenericAgentX-asaf:
Good catch — that was a real regression, not just a leftover gap. Fixed in b690b72ac.

`AgentEntry.pid` is the owning **srv process's** PID (shared by every agent registered under that channel/srv), so a PID-only guard could only prove "the whole process died" — a genuinely-dead individual agent on a live shared srv would never be evicted. Replaced `is_pid_alive` with `should_evict_on_forward_failure`, which evicts on either a confirmed-dead process *or* an entry older than a 60s grace period (so a fresh registration-race entry gets protected, but an old dead-agent entry still gets reclaimed regardless of its process's liveness). Added 3 unit tests, including one that directly encodes this regression so it can't come back silently.

`cargo check -p agentmux-srv` clean, 20/20 registry tests pass.

GenericAgentX-asaf:
Right again, and this one was the real hole — fixed in e146e567c.

`entry.updated_at` was only ever stamped once at registration, with nothing to refresh it. The 60s grace window I added only ever covered an agent's first minute of life — every steady-state agent (including this PR's own incident agent, already up "a few minutes" when it failed) was outside that window permanently, so a single transient failure still evicted it immediately. The "add a heartbeat" note I'd left as a deferred nice-to-have in the retro was actually load-bearing for the fix to work past minute one.

Added a 20s heartbeat in `bootstrap.rs` that re-writes every locally-registered agent's Tier 2 + Tier 2b registry entries, driven from `reactive::get_global_handler().list_agents()` (the in-memory Tier-1 registry — always current, no staleness of its own). A live agent's entries now stay fresh indefinitely; a dead one stops appearing in `list_agents()` and ages out normally.

`cargo check -p agentmux-srv` clean, 20/20 registry tests pass.

GenericAgentX-asaf:
Fixed in b8f9a3ef6 — the heartbeat now uses \`write_with_nonce\` / \`write_shared_from_env_with_nonce\`, threading \`reg.registration_nonce\` through instead of dropping it via the plain (nonce=0) variants. That was a real one: it would have silently broken \`remove_if_nonce\`/\`remove_shared_from_env_if_nonce\`'s clean-exit compare-and-remove on every live agent, every 20s.

\`cargo check -p agentmux-srv\` clean, 20/20 registry tests pass.

PR #2641, opened 2026-08-18
feat(agent-pane): pane title tracks the session's overall goal, not per-turn micro-steps

GenericAgentX-asaf:
## Summary
- The Haiku-generated pane title previously answered "what is currently being worked on" — a fresh, stateless, context-blind paraphrase of the last ~30 lines of raw CLI output, regenerated from scratch on every completed turn. In practice it thrashed between micro-steps ("reviewing the diff", "Feature merged to main branch") instead of settling on something like a PR title.
- Fix: anchor each call on the **current title** plus the **user's newest message**, and instruct the model to repeat the title back unchanged unless the message represents a genuinely new/expanded goal.
- Trigger moved from turn-end to submit-time — only the user's own messages can change the goal, the agent's tool calls never do. This also removes the need for a FileStore tail read in the common case: `TurnPhase.Submitting.pendingContent` (a field that existed but was never wired up — the reducer's own comment flagged it as a TODO) now actually carries the literal submitted text, threaded through a new optional `TurnStart.content` field.
- Backend prompt-building extracted into a pure, unit-tested `build_session_title_prompt()`.

See `docs/specs/SPEC_AMBIENT_PANE_TITLE_OVERALL_GOAL_TRACKING_2026_08_17.md` for the full analysis and design.

## Test plan
- [x] `cargo check` / `cargo test -p agentmux-srv` — clean, 4 new prompt-builder unit tests pass
- [x] `npx tsc --noEmit` — no new errors (2 pre-existing, unrelated)
- [x] 291 frontend tests pass across reducer, pane-state, command hooks, and the new 8-case `useAgentActivitySummary.test.ts` (submit-trigger, non-fire on other phases, mount-defer, staleness guard)
- [ ] Visual check in `task dev`: submit a message, confirm the pane title updates once and stays stable across a multi-tool-call turn


a5af:
@codex review

PR #2642, opened 2026-08-18
fix(browser-pane): stop loading-brain flicker on repeated load-state flips

Agent2-asaf:
## Summary
- Root cause: CEF's `on_loading_state_change` callback has no frame parameter — it aggregates loading state across the whole frame tree, not just the main document. A sub-frame/subresource load (ad iframe, chat widget, analytics beacon) can flip it long after the main page has rendered. The loading-brain overlay trusted that raw flag directly and is airspace-clipped to the full pane, so every spurious flip hid and re-revealed the entire native pane HWND — "flash into multiple loads, brain, page, brain, page."
- **Rust** (`agentmux-cef`): added a dedicated, genuinely main-frame-scoped loading tracker (armed in `on_before_browse`, cleared in `on_load_end_browser_pane`/a real main-frame `on_load_error`) — `on_loading_state_change_browser_pane` now emits this corrected signal instead of forwarding CEF's frame-blind parameter. (Diverges slightly from the spec's original proposal to reuse the load-timeout watchdog's armed state — that state disarms at navigation *commit*, too early for this purpose; a separate tracker was needed.)
- **Frontend**: once a pane has painted content once, a later loading flip renders a small corner badge instead of the full-pane overlay, so even a flip the Rust fix doesn't catch can't blank the whole visible page. A `loading:false` dispatch is also held ~200ms and canceled if a fresh `true` arrives within that window (defense in depth against anything neither of the above catches, e.g. a same-tick redirect hop).

Spec: `docs/specs/SPEC_BROWSER_PANE_LOADING_INDICATOR_FLICKER_2026_08_17.md`

## Test plan
- [x] `cargo check -p agentmux-cef` clean
- [x] `cargo test -p agentmux-cef --lib` — 251/251 passing
- [x] `npx tsc --noEmit` clean (2 pre-existing, unrelated errors elsewhere)
- [x] `npx vitest run` on the browser-pane suites — 115/115 passing (2 tests updated for the new debounce timing, 1 new test asserting a flip is actually suppressed)
- [x] Manually verified in `task dev` on a page with embedded widgets — confirmed improved, no more repeated brain/page flashing

### DIFF
--- a/.changesets/1787014443-fix-browser-pane-stop-loading-brain-flicker-on-repeated-load-v0wy.md
+++ b/.changesets/1787014443-fix-browser-pane-stop-loading-brain-flicker-on-repeated-load-v0wy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(browser-pane): stop loading-brain flicker on repeated load-state flips
--- a/agentmux-cef/src/browser_pane/callbacks.rs
+++ b/agentmux-cef/src/browser_pane/callbacks.rs
@@ -142,6 +142,67 @@ fn fire_pane_load_watchdog(state: &Arc<AppState>, block_id: &str, epoch: u64) {
     );
 }
 
+/// Set (or clear) the main-frame-loading tracker for `block_id` and, if the
+/// tracked state actually changed, emit a `browser-pane-nav-state` event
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #2643, opened 2026-08-18
Manoz@Area54 fix(tabs): remove automatic tab color assignment

GenericAgentX-asaf:
Removes two coupled auto-assignment mechanisms so all tabs (default and new) start with no color — users still set one manually via the existing right-click swatch picker, unchanged.

1. `createTab()` (`tab-actions.ts`) assigned a random hex from `TAB_COLORS` to every newly-created tab. Removed `randomNewTabColor()` and its call site.
2. `tabbar.tsx` backfilled a fixed "Blue" onto the startup tab — this existed ONLY to paper over the inconsistency #1 created (first tab neutral, every tab after it randomly colored). With #1 gone there's nothing left to stay consistent with, so removed too, along with the now-dead `tab:color-initialized` guard meta key (`gotypes.d.ts`).

"No color" was already a fully-supported, intentional state before this change — `tab.tsx` already renders gracefully with no `tab:color` set, and the picker already has an explicit "Clear color" action. This just makes that the starting state instead of something a user opts into.

Existing tabs that already have a color (random, backfilled, or manually chosen) are untouched — this only stops future auto-assignment, not a migration.

See `docs/reports/REPORT_REMOVE_AUTO_TAB_COLOR_2026_08_18.md` for the full analysis.

### DIFF
--- a/.changesets/1787014873-fix-tabs-remove-automatic-tab-color-assignment-0zbo.md
+++ b/.changesets/1787014873-fix-tabs-remove-automatic-tab-color-assignment-0zbo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tabs): remove automatic tab color assignment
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2645, opened 2026-08-18
Fix macOS false-positive needs_reauth for Keychain-backed Claude accounts

AgentO-asaf:
## Summary
- Root cause (documented in the new retro): Claude Code CLI stores OAuth credentials in the macOS Keychain, never in `<dir>/.credentials.json`, regardless of `CLAUDE_CONFIG_DIR` — unlike Linux/Windows, where `CLAUDE_CONFIG_DIR` genuinely relocates the credential file. This makes AgentMux's per-agent credential isolation strategy a silent no-op on macOS.
- Concrete bug found and fixed while investigating: `probe_oauth_status()` treated a missing token file as a definitive `needs_reauth` signal, which — given the root cause above — meant every working, Keychain-backed macOS Claude account was almost certainly being mislabeled `needs_reauth`. Verified empirically against a real per-identity bundle dir on the dev machine (working session, zero `.credentials.json`).
- Fix: on macOS, a missing token file for `claude` now returns `None` (status left alone) instead of asserting a false `needs_reauth`. `codex`/`openclaw` are untouched (their macOS credential-storage behavior hasn't been verified).
- Two pre-existing `inject.rs` integration tests that exercised this exact scenario under `"claude"` were repointed to `"codex"` so they keep testing the general probe→upsert plumbing rather than colliding with the new macOS-specific carve-out.
- Adds a plan for the broader isolation gap (`PLAN_MACOS_CLAUDE_KEYCHAIN_CREDENTIAL_ISOLATION_2026_08_17.md`). The riskier remaining item — independently managing OAuth token refresh outside the CLI's own Keychain-backed flow — is intentionally deferred, not started in this PR.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv identity::` — 89 passed, 0 failed
- [x] New tests added: platform-gated coverage in `oauth_probe.rs` (`#[cfg(target_os = "macos")]` / `#[cfg(not(target_os = "macos"))]`) proving the carve-out only applies to `claude` on macOS
- [x] Verified the underlying claim empirically against a real per-identity bundle dir on this machine (no `.credentials.json`, working session)

### DIFF
--- a/.changesets/1787037196-fix-identity-stop-mislabeling-macos-keychain-backed-claude-a-xi4o.md
+++ b/.changesets/1787037196-fix-identity-stop-mislabeling-macos-keychain-backed-claude-a-xi4o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): stop mislabeling macOS Keychain-backed Claude accounts as needs_reauth
--- a/agentmux-srv/src/identity/resolver/inject.rs
+++ b/agentmux-srv/src/identity/resolver/inject.rs
@@ -1938,6 +1938,14 @@ mod tests {
         // bundle dir with NO token file → the probe surfaces
         // `needs_reauth` and the resolver upserts the account row
         // with the new status. Spec §4.4.
+        //
+        // Uses "codex", not "claude": the macOS Keychain carve-out added
+        // in oauth_probe.rs (retro-macos-keychain-credential-isolation-gap-
+        // 2026-08-17.md) makes a missing token file for "claude" return
+        // `None` (no status change) on macOS specifically, since Claude
+        // Code never writes `.credentials.json` there regardless of the
+        // config dir. "codex" isn't covered by that carve-out, so it keeps
+        // this test's original, platform-independent needs_reauth signal.
         let store = make_store();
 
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentO-asaf:
Addressed both review findings in 694de8468:

- **reagent P1** (missing changeset): added `.changesets/1787037196-fix-identity-stop-mislabeling-macos-keychain-backed-claude-a-xi4o.md` via `task changeset -- patch`.
- **Codex P1** (stale needs_reauth not healed): good catch — added a self-heal path in `inject_identity_env_with_broker`. When the macOS Keychain carve-out fires (missing token file, `claude`, macOS) and the account's currently-persisted status is exactly `needs_reauth`, it now resets to `"unknown"` on the next spawn instead of staying stuck. Deliberately not resetting to `"valid"` — the probe genuinely can't tell on macOS, and a fabricated `"valid"` would be exactly as dishonest as the bug being fixed. Added `inject_oauth_class_self_heals_stale_needs_reauth_for_claude_on_macos` covering it.

Full identity suite (90 tests) + workspace `cargo check --tests` green.

PR #2647, opened 2026-08-18
feat(bundle): Bundle editor MCP Servers/Skills sections (composable model v2)

AgentY-asaf:
## Summary

Delivery-plan step 4 (final) of `docs/specs/SPEC_BUNDLE_AS_CONTAINER_V2_2026_08_17.md`, GH issue #2024 item 3 — frontend follow-up to the backend PR #2639 (schema/RPCs/launch-time wiring). Before this, `memory-manager.tsx`'s bundle-editor form still said MCP servers and skills "will be editable here in a follow-up" — this is that follow-up.

- **`BundleMcpModel`/`BundleSkillModel`** (`bundle-mcp-model.ts`, `bundle-skill-model.ts`): view models mirroring the existing `AgentMcpModel`/`AgentSkillModel`, keyed by `bundle_id` instead of `agent_id`, plus `addPrivate()` — the actually-functional path (creates a brand-new bundle-owned private server/skill via `mcp.catalog.upsert_for_bundle`/`skill.catalog.upsert_for_bundle`). Binding an existing global item is offered for parity with the agent-scoped modal, but the UI carries a caveat that it has no effect on a spawned agent — globals already reach every agent unconditionally regardless of bundle binding.
- **`BundleMcpSection`/`BundleSkillsSection`**: compact inline list + bind/unbind + "add private" form, mounted directly in the bundle's readonly detail view. Deliberately a flat list, not a nested `PrimitiveListDetail` — this content already sits inside one side of `MemoryManagerBody`'s own list/detail split, and a second full-height single-pane swap nested inside that would fight for the same space.
- Updated the stale "will be editable here in a follow-up" hint text in the bundle edit form to point at the new sections.

## Test plan

- [x] 22 new model tests (bind/unbind/addPrivate success + failure, `addingAtom` in-flight state, cross-window `mcp:changed`/`skills:changed` reactivity, dispose) — spot-verified meaningful by breaking the `addPrivate` bundle_id wiring and confirming the test catches it before trusting it.
- [x] Full frontend suite: 2759 passed, 0 failed.
- [x] `tsc --noEmit`: no new errors (2 pre-existing, unrelated, in `armory-view.tsx`/`warden-view.tsx`).
- [x] Changeset added (`minor`).
- [x] Branch created fresh off latest `main` (just fast-forwarded from PR #2639) — no staleness to merge.
- [ ] **Not visually verified in a running app** — this requires the full Rust backend + CEF desktop host, which isn't practical to launch interactively in this environment. Noting this explicitly rather than claiming visual verification; recommend a manual pass before merge if UI correctness matters beyond what the model tests + typecheck cover.

### DIFF
--- a/.changesets/1787034731-feat-bundle-bundle-editor-mcp-servers-skills-sections-compos-djrk.md
+++ b/.changesets/1787034731-feat-bundle-bundle-editor-mcp-servers-skills-sections-compos-djrk.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(bundle): Bundle editor MCP Servers/Skills sections (composable model v2)
--- a/frontend/app/view/memory/BundleMcpSection.test.tsx
+++ b/frontend/app/view/memory/BundleMcpSection.test.tsx
@@ -1,0 +1,79 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * reagentx P1 on PR #2647: a failed "add private server" must not silently
+ * discard the name/config the user just typed — `addPrivate` never rejects
+ * (errors go to errorAtom, same convention as bind/unbind), so a naive
+ * `.then(() => clearForm())` ran unconditionally regardless of success.
+ */
+
+import { cleanup, render, screen } from "@solidjs/testing-library";
+import userEvent from "@testing-library/user-event";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+
+vi.mock("@/app/store/wps", () => ({ waveEventSubscribe: vi.fn(() => () => {}) }));
+vi.mock("@/app/store/rpc-api", () => ({
+    RpcApi: {
+        McpCatalogListForBundleCommand: vi.fn(),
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed review findings (review 4958016102):

**P1 x2 — failed add silently discarded typed input.** `addPrivate` never rejects (errors go to `errorAtom`, same convention as `bind`/`unbind`), so `BundleMcpSection`/`BundleSkillsSection`'s `.then(() => clearForm())` ran unconditionally regardless of success — a failed add (duplicate name, invalid JSON, FORBIDDEN) silently wiped the user's typed name/config even though the error banner showed, forcing a full retype.

Fix: `addPrivate` now returns a success boolean; both components only clear their form fields when it resolves `true`.

2 new component tests (type into the form, submit against a mocked RPC rejection, assert the fields still hold what was typed) — verified failing against the pre-fix unconditional `.then()` first. Also updated the existing model-level `addPrivate` tests to assert the new return value.

Full suite: 2763 passed, 0 failed. `tsc --noEmit`: clean (the 2 previously-noted pre-existing errors were fixed by a `main` merge, unrelated to this PR).

PR #2649, opened 2026-08-18
feat(editor): word-wrap toggle + gutter/tab-strip visual polish

GenericAgentX-asaf:
## Summary
- Adds a Word Wrap toggle to the editor's right-click menu (`getBodyContextMenuItems`) — was previously hardcoded always-on with no way to turn it off. Persisted per-pane (`editor:word_wrap` block meta), reconfigured live via a CodeMirror `Compartment` (no rebuild needed), and resynced on tab switch since it's a pane-wide setting, not per-tab.
- Fixes a real gutter bug: `@codemirror/lint`'s `lintGutter()` ships a runtime-injected base theme (`.cm-gutter-lint { width: 1.4em }`) that was winning a CSS specificity tie against our existing `display: none` override, leaving a stray ~18px empty column between the line number and the code even though the lint gutter was supposed to be fully hidden.
- Tightens the line-number-to-text gap and gives the numbers a bit more horizontal breathing room on their own.
- Line numbers now use the theme's accent color instead of a flat secondary-text gray — dimmer for regular lines, full accent strength for the current line, so the active line visibly stands out. The active line's row background is now a light accent tint instead of a flat white tint.
- The tab strip's bottom border now runs the full width of the pane instead of stopping right after the last tab/`+` button. The strip itself stays intentionally shrink-to-fit (per `SPEC_PANE_TAB_STRIP_COMPACT_SIZING_AND_RENAME_2026_07_22.md`) — only a wrapping element around it picked up the full-width border, so agent/terminal tab strips (which share the same underlying `PaneTabStrip` component) are unaffected.

Iterated live against a `task dev` build with a human in the loop for every visual change in this PR.

## Test plan
- [x] Word Wrap: toggled on/off via right-click on the editor body, confirmed it applies immediately and survives a tab switch
- [x] Confirmed the gutter gap fix and remaining spacing/color changes visually against a running dev build
- [x] Confirmed the tab-strip border now reaches the pane's right edge
- [ ] Automated test coverage — none added yet for the new toggle; flagging for reviewer input on whether it's expected before merge

### DIFF
--- a/.changesets/1787036796-feat-editor-word-wrap-toggle-tighter-gutter-spacing-accent-c-c5ub.md
+++ b/.changesets/1787036796-feat-editor-word-wrap-toggle-tighter-gutter-spacing-accent-c-c5ub.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(editor): word-wrap toggle, tighter gutter spacing, accent-colored line numbers, full-width tab strip border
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2650, opened 2026-08-18
fix(editor): + button now always opens a new tab

GenericAgentX-asaf:
## Summary
Follow-up to #2649 (merged before this commit made it in). `openScratch()` reused an existing scratch tab whenever one was already open in the pane — correct for the constructor's idempotency-guarded default-open call, but wrong for the "+" button: since a fresh editor pane opens with a default scratch tab, clicking "+" just silently re-activated that same already-active tab. Indistinguishable from broken.

- Added a `reuseExisting` param to `openScratch()` (default `true`, preserving the constructor's dedup behavior).
- The "+" button now calls `openScratch(false)`, so it always creates a genuinely new tab, matching ordinary tabbed-UI expectations (VS Code's Untitled-1/2/3, etc.).

## Test plan
- [x] Verified live against a `task dev` build: pane opens with its default scratch tab, clicking "+" now opens a second, genuinely new blank tab instead of no-op'ing
- [ ] No automated test coverage added — flagging for reviewer input on whether it's expected before merge

### DIFF
--- a/.changesets/1787041527-fix-editor-button-now-always-opens-a-new-tab-instead-of-re-a-2lod.md
+++ b/.changesets/1787041527-fix-editor-button-now-always-opens-a-new-tab-instead-of-re-a-2lod.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(editor): + button now always opens a new tab instead of re-activating an existing scratch tab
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2651, opened 2026-08-18
fix(agent-pane): stop tool_result echo from clobbering an answered AskUserQuestion

AgentA-asaf:
## Summary
Follow-up to #2630 — addresses a bug found while manually verifying that PR live: an answered `AskUserQuestion` briefly rendered with the new user-message styling, then reverted to the old collapsed tool row.

- **Root cause:** `AskUserQuestion` is a real tool call, so once `handleAnswer`'s optimistic update sets `answerText`/`timeoutNote` and delivers the answer, the CLI's own `tool_result` for the same `tool_use_id` still arrives once the answer resumes the turn (`claude-translator.ts` has no special-casing for it). That event gets parsed into a fresh, generic `ToolNode` via `stream-parser.ts`'s `toolResultToNode()` (no concept of these fields), and `mergeReplacement`'s wholesale-replace path (AskUserQuestion has no log buffer to otherwise gate on) silently wiped `answerText` back to `undefined` — reverting the resolved question to the old collapsed-row rendering shortly after the user answered.
- **Fix:** `mergeReplacement` now preserves `answerText`/`timeoutNote`/`questionText`/`summary` across that echo when the existing node is an already-answered `AskUserQuestion`.

Also, per direct follow-up feedback after #2630 shipped:
- Reverted the enlarged-text treatment for the answer (`--answered-question` SCSS modifier) — it now renders at the same size as ordinary typed input.
- The original question now prints too, as plain agent text (the same treatment `MarkdownBlock` gives a normal assistant message) directly above the answer — flattened into a new `questionText` field in `handleAnswer` before `question` is cleared. The resolved node now reads as an ordinary question-then-answer exchange instead of only ever showing the answer half.

## Test plan
- [x] `npx stylelint` on the changed SCSS — no new errors (31 pre-existing, unrelated — same baseline #2630 reported)
- [x] Full suite: `npx vitest run` — 2741 passed, 3 skipped (2 e2e, no build dir)
- [x] Added a reducer regression test that reproduces the exact clobbering scenario; confirmed it fails without the fix (`answerText` → `undefined`) and passes with it
- [x] Verified live in `task dev`: answered a fresh question, confirmed it now renders normal-size with the question printed as agent text above the answer, and stays that way (no revert)

### DIFF
--- a/.changesets/1787041732-fix-agent-pane-stop-the-cli-s-own-tool-result-echo-from-clob-7j6h.md
+++ b/.changesets/1787041732-fix-agent-pane-stop-the-cli-s-own-tool-result-echo-from-clob-7j6h.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): stop the CLI's own tool_result echo from clobbering an answered AskUserQuestion's answerText
--- a/docs/retro/RETRO_ASK_USER_QUESTION_ANSWER_TEXT_CLOBBER_2026_08_18.md
+++ b/docs/retro/RETRO_ASK_USER_QUESTION_ANSWER_TEXT_CLOBBER_2026_08_18.md
@@ -1,0 +1,146 @@
+# Retro: Answered `AskUserQuestion` silently reverted to the old collapsed row (PR #2630 follow-up)
+
+**Date:** 2026-08-18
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #2653, opened 2026-08-18
feat(muxlog): add muxlog phases — merged turn-phase timeline for one agent pane

AgentA-asaf:
## Summary
Implements `docs/specs/SPEC_AGENT_TURN_PHASE_TIMELINE_LOGGING_2026_08_18.md`, motivated directly by `docs/reports/REPORT_AGENTA_STUCK_WORKING_INVESTIGATION_2026_08_14.md` — answering "why did this pane show Working" required manually cross-referencing two separate log files by hand and inferring watchdog liveness from the *absence* of expected lines rather than confirming it directly. This gives agents (and humans) a direct, one-command answer instead.

- **`muxlog phases [<block-id>]`** merges the frontend's `[wave-turn]` transition log (host) and the backend's `[health] turn_active flip` log (srv) chronologically into one timeline for a single pane. Defaults to the caller's own pane via `$AGENTMUX_BLOCKID` (already set in every agent's shell env), so self-audit needs zero setup.
- Host and srv logs are correlated to the **same running instance** by `(source, version)` together, not independently "most recently active" per target — `source` alone isn't enough (the shared log root routinely holds several version-tagged pairs at once). This is a real bug caught live while manually verifying the recipe against synthetic fixtures — see the spec's "Implementation notes" section for the full story.
- A stray/late `StreamFlushObserved` re-promotion (`Idle`/`Disconnected`/`Done` → `Streaming`, as opposed to the normal `Submitting` hand-off) is now tagged `(stray)` in the existing `[wave-turn]` line.
- The watchdog interval now emits a `watchdog: tick #N` heartbeat every ~60s (every 12th tick) — direct proof of liveness instead of an inference from total silence.
- `document.visibilitychange` firing is now logged explicitly in both directions, since PR #2575's catch-up-tick fix depends on it firing but nothing previously confirmed that it actually does.

## Test plan
- [x] `node --check` on `muxlog.mjs` — syntax valid
- [x] Full suite: `npx vitest run` — 2771 passed, 3 skipped (2 e2e no build dir, 1 registry test flaked under full-suite load, confirmed passing in isolation — unrelated to this change)
- [x] New unit tests: stray-flush tagging (positive + negative case), watchdog heartbeat cadence, both `visibilitychange` directions — all in the existing test files for the touched modules
- [x] Manual end-to-end verification of `muxlog phases` against synthetic host+srv NDJSON fixtures: correct block-id filtering (excludes other panes/blocks), correct chronological interleaving, `(stray)` tagging, `--raw`, `-n`, the `$AGENTMUX_BLOCKID` default, both error paths, and the host/srv correlation fix itself (reproduced the bug, then confirmed the fix)


a5af:
@codex review

PR #2654, opened 2026-08-18
Fix macOS Keychain password prompt on every app launch

AgentO-asaf:
## Summary
- `frontend/app-init.ts` fires the `providers.models` RPC unconditionally on every app launch (fire-and-forget, to refresh the Claude model dropdown's labels).
- On macOS, where Claude Code never writes `.credentials.json`, that RPC fell through to an unconditional read of AgentMux's own Keychain-backed fallback-token store (`identity/secret_store.rs`, `keyring` crate under service `"agentmux"`).
- If that Keychain entry already had something in it (e.g. from an earlier session where `CLAUDE_CODE_OAUTH_TOKEN` was exported once), macOS's Keychain ACL requires approval from any code signature it doesn't already trust — for a locally rebuilt dev binary, that's every fresh build. Result: an interactive password prompt at app open, for a purely cosmetic feature.
- Fix: `resolve_access_token` now takes `allow_keychain_fallback: bool`. The automatic startup call passes `false`, skipping the Keychain-touching fallback steps entirely — no Keychain interaction from that path, ever. The capability itself isn't deleted (still callable with `true` from a future explicit user action); it's just no longer reachable from a silent background call.
- Adds a retro (`docs/retro/retro-startup-keychain-prompt-model-catalog-2026-08-18.md`) documenting the investigation.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv model_catalog::` — 8 passed, including a new test locking "flag off + no file → None, never enters the keychain-fallback branch"
- [x] Root cause confirmed by code tracing (`app-init.ts` → `providers.models` RPC → `resolve_access_token` → `secret_store::get`, unconditional on every launch when `.credentials.json` is absent)

### DIFF
--- a/.changesets/1787062243-fix-model-catalog-stop-the-automatic-startup-model-catalog-r-sfyx.md
+++ b/.changesets/1787062243-fix-model-catalog-stop-the-automatic-startup-model-catalog-r-sfyx.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(model-catalog): stop the automatic startup model-catalog refresh from prompting for the macOS Keychain password
--- a/agentmux-srv/src/backend/model_catalog.rs
+++ b/agentmux-srv/src/backend/model_catalog.rs
@@ -96,37 +96,44 @@ pub fn read_oauth_access_token(config_dir: &Path) -> Option<String> {
 /// Callers should use this instead of calling [`read_oauth_access_token`]
 /// directly. Never logs the token.
 ///
-/// `allow_keychain_fallback` gates steps 2/3 — both hit the OS keychain (via
-/// `identity::secret_store`), which wraps blocking `keyring` syscalls that
-/// can trigger an interactive macOS "App wants to use your confidential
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentO-asaf:
Addressed in 39375a7fc — good catch: reading `CLAUDE_CODE_OAUTH_TOKEN` is a plain process-env read with no keychain interaction, so it now always runs regardless of `allow_keychain_fallback`. Only the two operations that actually touch the OS keychain — persisting a freshly-read env-var token, and reading back a previously-persisted one — are gated behind the flag. Added a test (`resolve_access_token_still_honors_env_var_when_fallback_disallowed`) locking this in, alongside the existing keychain-untouched test. 9/9 model_catalog tests pass, workspace `cargo check --tests` clean.

AgentO-asaf:
Fixed in 47ecd1c4e — the None-branch comment was stale from before the previous fix; updated to correctly describe the two allowed sources (file, env var) and that Keychain is deliberately excluded.

PR #2658, opened 2026-08-19
fix(agent-pane): preserve cache-read/cache-creation token breakdown instead of collapsing it

AgentA-asaf:
## Summary

Both frontend token-usage parse sites (`claude-translator.ts`'s `result` handler and `useAgentStream.ts`'s `message_start` handler) summed Anthropic's four usage fields — `input_tokens` + `cache_creation_input_tokens` + `cache_read_input_tokens` — into one number before anything was stored, discarding the split despite a code comment noting it mattered. That made it impossible to tell after the fact whether a session's input-token total was mostly cheap cache reads (~0.1x cost) or expensive fresh tokens.

- Extends `TurnTokens` / `SessionStats` / the `TokensIn` command / `ServiceUsage` with optional `freshInput` / `cacheCreation` / `cacheRead` (snake_case equivalents on `SessionStats`), threaded additively through both parse sites → the reducer's `TokensIn`/`TokensOut`/`mergeStats`/`accumulateStats` → `useTurnLifecycle`'s `finalizeTurn` → the `token-usage.ts` session store.
- The existing collapsed `input` / `input_tokens` total is unchanged, so every existing consumer keeps working as-is — this is additive only.
- Surfaces a "N% of input served from cache" line in the token breakdown popover once at least one turn has reported the split.
- Full write-up, including compaction's current passive-only architecture and a proposed path to a manual-trigger option (not addressed by this PR): `docs/reports/REPORT_TOKEN_ACCOUNTING_AND_COMPACTION_CONTROL_2026_08_18.md`.

## Test plan

- [x] `npx vitest run` on the three touched suites (`claude-translator.test.ts`, `reducer.test.ts`, `parseHistoryLines.test.ts`) — 249/249 pass
- [x] Updated `claude-translator.test.ts`'s exact-equality assertion to include the new breakdown fields
- [ ] Manual verification of the new "N% served from cache" popover line in a live pane (not done — no live Claude session run as part of this PR)

### DIFF
--- a/.changesets/1787102240-fix-agent-pane-preserve-cache-read-cache-creation-token-brea-xzy9.md
+++ b/.changesets/1787102240-fix-agent-pane-preserve-cache-read-cache-creation-token-brea-xzy9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): preserve cache-read/cache-creation token breakdown instead of collapsing it into input_tokens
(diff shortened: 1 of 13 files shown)


a5af:
@codex review

PR #2659, opened 2026-08-19
feat(agent-pane): add manual "Compact now" trigger for Claude sessions

AgentA-asaf:
## Summary

Follow-up to #2658 — same investigation (`docs/reports/REPORT_TOKEN_ACCOUNTING_AND_COMPACTION_CONTROL_2026_08_18.md`), second recommendation from that report. Independent of #2658's changes — this branch is based on `main`, not on #2658's branch, so it can be reviewed/merged separately.

Adds a small "Compact" button next to the context-fill meter in the composer strip (visible only for Claude panes, only once there's context to compact, disabled while a turn is in flight or a compaction is already running). Clicking it sends the literal text `/compact` through the exact same `handleSendMessage` path the composer's own Send button and the SlashHelpPanel already use.

**Why this is safe:** empirically verified against a live `claude` CLI process (v2.1.218) in an isolated scratch dir, using the same persistent-mode launch args `providers.rs` configures — a stream-json user turn with content `/compact` is recognized as the real command, not literal text. The CLI emits a progress frame (`system/status: compacting`), then a terminal status (`compact_result: "success"|"failed"`), and on success emits the **exact same** `system/compact_boundary` frame (`compact_metadata.trigger: "manual"`) that AgentMux's backend already parses for auto-compaction — the `CompactionTrigger::Manual` variant it maps to already existed in `agents/types.rs` with no prior producer. Net result: **no new backend RPC, no new control-protocol frame, no parser changes needed.**

Deliberately **not** registered as a `SlashCommand` (`commands/providers/claude.ts` stays empty) — `dispatch.ts`'s own comment confirms a *registered* command's handler returning `{kind:"passthrough"}` is logged and ignored, not forwarded. Leaving `/compact` unregistered means it takes the same unknown-command passthrough path a user manually typing it already takes today, which `useAgentCommands.sendMessage` correctly forwards as a real turn.

## Test plan

- [x] `npx vitest run frontend/app/view/agent` — 1222/1222 (one unrelated `tool-renderers/registry.test.ts` flake reproduced on `main` too; passes cleanly in isolation)
- [x] Empirical CLI probe (see report §6) confirming `/compact` over stream-json stdin behaves as claimed, including the full success-path event sequence
- [ ] Manual click-through of the new button in a live pane (not done as part of this PR — no GUI build run)

### DIFF
--- a/.changesets/1787104458-feat-agent-pane-add-manual-compact-now-trigger-for-claude-se-eudr.md
+++ b/.changesets/1787104458-feat-agent-pane-add-manual-compact-now-trigger-for-claude-se-eudr.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): add manual 'Compact now' trigger for Claude sessions
--- a/docs/reports/REPORT_TOKEN_ACCOUNTING_AND_COMPACTION_CONTROL_2026_08_18.md
+++ b/docs/reports/REPORT_TOKEN_ACCOUNTING_AND_COMPACTION_CONTROL_2026_08_18.md
@@ -1,0 +1,193 @@
+# Token Accounting & Compaction Control — Current State, Gaps, and a Path Forward
+
+**Date:** 2026-08-18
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2661, opened 2026-08-19
fix(muxbus): let the browser sign-in flow be cancelled instead of hanging for 5 minutes

AgentY-asaf:
## Summary

Closing the muxbus sign-in browser tab left the status-bar "Sign in" chip stuck disabled indefinitely. Root cause: `run_pkce_login`'s poll loop had no cancellation path and only gave up after the full 5-minute `LOGIN_TIMEOUT_SECS`.

- Adds `cancel_active_login()` in `pkce.rs`, which aborts the in-flight flow with an accurate `"sign-in cancelled"` error — distinguished from the existing `"superseded by a newer login"` case (same abort mechanism, different reason, tracked via a `CANCELLED_BY_USER` flag).
- New `muxbus.login.cancel` RPC + `MuxBusLoginCancelCommand` on the frontend.
- Every button that shows "Connecting…"/"Sign in" (status-bar popover, agent identity panel, Armory connect panel) now shows a Cancel affordance while a login is in flight instead of a stuck disabled state.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv muxbus::pkce::` — 3/3 passing (including a new regression test proving cancel resolves immediately instead of waiting out the timeout), stable across repeated runs and under default parallel test execution
- [x] `cargo build` (agentmux-srv) and `tsc --noEmit` both clean
- [x] Live click-through: dev build, reproduced the original hang, confirmed Cancel now appears and resolves immediately

### DIFF
--- a/.changesets/1787124092-fix-muxbus-let-the-browser-sign-in-flow-be-cancelled-instead-fhhk.md
+++ b/.changesets/1787124092-fix-muxbus-let-the-browser-sign-in-flow-be-cancelled-instead-fhhk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(muxbus): let the browser sign-in flow be cancelled instead of hanging for 5 minutes
--- a/agentmux-srv/src/muxbus/pkce.rs
+++ b/agentmux-srv/src/muxbus/pkce.rs
@@ -43,24 +43,28 @@ fn relay_base_url() -> String {
 // a port another live socket already holds) the browser's redirect lands on
 // whichever listener the OS picks — a 400 state-mismatch coin flip. The newest
 // attempt is the one the user actually wants: abort the predecessor.
-static ACTIVE_LOGIN: std::sync::Mutex<Option<tokio::task::AbortHandle>> =
+//
+// Each flow gets a monotonic generation id (NEXT_LOGIN_ID), stored alongside
+// its abort handle. A prior design tracked "was this abort a user Cancel?"
+// in a single shared CANCELLED_BY_USER flag instead of per-attempt — that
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed both reviews in 332aeecc9:

- **reagent P1 / Codex's second finding** (`CANCELLED_BY_USER` shared flag): fixed by keying cancellation on a per-attempt generation id instead of a single process-global bool, so one flow's cancel can no longer be misattributed to a different concurrent flow regardless of task-resolution order. Added a regression test (`cancelling_one_flow_never_leaks_into_a_concurrent_unrelated_flow`) that starts a cancelled flow concurrently with an unrelated superseded flow and asserts neither can affect the other's reported reason.
- **reagent P2** (missing changeset): added via `task changeset -- patch`.

On **Codex's first finding** (a stale Cancel click from an already-superseded flow's UI could abort the wrong, newer flow instead of no-op'ing): this is a real gap, but closing it fully requires threading an attempt id through the `muxbus.login`/`muxbus.login.cancel` RPC contract end-to-end (frontend would need to learn its own attempt's id before the blocking `muxbus.login` call resolves, which the current synchronous RPC shape doesn't support without a wire-format change). Given the race window is narrow (bounded by "only one login flow may exist per process" combined with the existing supersede-on-start behavior — already a documented, accepted simplification), I'm scoping that as a follow-up rather than expanding this PR's surface. Flagging explicitly rather than leaving it silently unaddressed.

3/3 pkce tests + the new regression test pass, stable across repeated runs; `cargo build` clean.

AgentY-asaf:
Fixed both stale-comment nits in df72ec4ac.

PR #2662, opened 2026-08-19
feat(agent-api): add UIScreenshot/UIClick/UIQuery — agent-facing UI automation

AgentY-asaf:
## Summary

Adds a first-class Agent App API capability: agents can now screenshot, click, and query AgentMux's own UI directly, instead of relying entirely on unit tests + a human click-through to verify a UI change actually works.

This builds on `agentmux-cef/src/browser_api/` — an existing, shipped CDP automation layer built to replace flaky pixel-coordinate clicks in the `pane-focus-stress.ps1` test harness, previously wired only to Browser-widget panes and never exposed to any agent. Design rationale, prior-art survey, and industry-practice research: `docs/specs/SPEC_AGENT_UI_AUTOMATION_CLICK_SCREENSHOT_2026_08_18.md`.

### What changed
- **`browser_api` resolver**: widened target resolution from "browser-pane widgets only" to any pane, via a new DOM-probe fallback (`resolve_by_dom_probe`) that asks each live CDP target whether its DOM contains the block's `[data-blockid]` wrapper.
- **Scoping** (the actual hard part): every `UIClick`/`UIQuery` call is scoped to the caller's own pane **or** shared app chrome (status bar, hamburger menu, window controls — not owned by any one agent) — but never a *different* agent's pane. `UIScreenshot` stays strictly pane-only: its CDP `clip` is one rectangle that can't exclude another pane's pixels sitting between the caller's pane and a chrome element the way an element-level filter can.
- **`host_ipc.Register` handshake**: the CEF host pushes its own `ipc_port`/`ipc_token` to `agentmux-srv` once at startup, so srv can proxy `/api/v1/ui/{screenshot,click,query}` through to the host's `/agentmux/browser/*` CDP routes — closing the one missing piece of cross-process plumbing (srv never had a way to reach the host's CDP layer before).
- **Three new MCP tools**: `UIScreenshot`, `UIClick`, `UIQuery`. `block_id` is stamped server-side from `AGENTMUX_BLOCKID` (the same trusted-env pattern every other App API tool uses) — never agent-suppliable, so "own pane only" holds by construction.

### A bug the live verification actually caught
The resolver initially reused Path 1's (dedicated browser-pane) target-exclusivity rule for Path 2 (shared window) too — so the *first* pane resolved in a window "claimed" the only CDP target, and every *other* pane in that same window then failed to resolve at all. Caught live (querying the agent pane worked; the sysinfo pane, same window, didn't), reproduced with a regression test driving a fake CDP WebSocket responder, then fixed: Path 2 target sharing across multiple blocks is the expected, correct behavior, not a collision to prevent.

## Test plan
- [x] `cargo build` clean across `agentmux-cef`, `agentmux-srv`, `agentmux-mcp`, `agentmux-common`
- [x] `cargo test` — 3 resolver tests + 5 srv proxy/registration tests, all passing; full existing 60-test srv suite still green
- [x] `tsc --noEmit` clean
- [x] Live end-to-end against a real running dev instance: `UIQuery` found the exact status-bar "Sign in" chip; `UIClick` performed a real synthetic click on it, opening the popover; `UIScreenshot` returned a valid, correctly-clipped PNG (visually confirmed) of just the caller's own pane; cross-pane isolation confirmed (a different pane's `block_id` cannot see or reach the agent pane's content); re-verified after the resolver fix across all three panes in the window

## Explicitly out of scope (Phase 1 — see spec §5)
Cross-agent/cross-window targeting (no capability-flag mechanism exists yet to gate it); OS-level whole-desktop capture/input injection; coordinate-based (non-selector) clicking; iframe/subframe targeting.


a5af:
@codex review

AgentY-asaf:
Closed the P0 for real in 4cc0ec8e5 (not just the doc-only pass in c0d3dd2).

`block_id` is no longer a request field on `/api/v1/ui/*` at all — there's nothing left to spoof. Every request now carries `UiAutomationAuth {agent_id, ts_secs, sig}`: an HMAC-SHA256 signature over the caller's own `agent_id`, using that agent's own `AGENTMUX_JEKT_KEY` — the same per-agent signing key already shipped for jekt sender authentication, reused rather than inventing a parallel credential system. `verified_block_id` (`ui_handlers.rs`) verifies the signature against the claimed agent's key on file, and only once that succeeds, derives that agent's **actual current block_id server-side** from the global `ReactiveHandler` registry. An agent without agent X's key cannot produce a valid signature claiming to be X, regardless of what it puts in the request body.

Added a regression test for the exact attack: an agent signs correctly with its own key but substitutes a different agent's `agent_id` into the request — rejected (`ui_click_rejects_a_forged_agent_identity`), since a signature valid for one identity doesn't verify against a different claimed one. Also `ui_click_rejects_an_unsigned_request`.

Full test count: 67 passing (63 srv including 3 new crypto-specific ones, 4 agentmux-cef resolver). Clean build across all 4 touched crates.

What I did NOT do this round: a full live relaunch spawning a real agent process to drive the actual `agentmux-mcp` binary end-to-end (this session is already very long). Instead I directly verified the MCP-side signing call (`sign_ui_automation_auth`) and srv-side verification call (`verified_block_id`) use byte-for-byte identical parameters to `jekt_sign::sign_jekt`/`verify_jekt`, and the existing test suite already exercises the real (non-mocked) sign→verify round-trip through the actual production functions, just not through the compiled `agentmux-mcp` binary specifically.

AgentY-asaf:
Fixed in 57be6f036. My P0 fix's unconditional reject was too strict — you were right that it broke real crash-restart recovery. Rather than a permanent one-shot lock, \`handle_register\` now probes the currently-registered host's own \`/health\` route before rejecting a conflicting re-registration: still alive → reject (hijack protection unchanged), unreachable → accept (the old host is provably gone, so this is recovery, not an attack).

Added a test for exactly the scenario you described (old registration points at an unreachable port → new registration succeeds) alongside the existing "still alive → rejected" test, now using a real live fake server so the liveness probe is genuinely exercised rather than coincidentally passing. 64 srv tests passing, stable across repeated runs.

AgentY-asaf:
Closed the None-state registration race from the latest re-review (commit 8574e40).

**Root cause confirmed:** the liveness probe added in the previous round only helps once `state.host_ipc` already holds a prior registration to check aliveness against. At every srv startup, and the window after any `restart_backend` before the host re-registers, `state.host_ipc` is `None` — nothing to probe — so any caller won that race outright, and could then stand up its own always-200 `/health` responder to permanently defeat the liveness check on every future legitimate registration too.

**Fix:** a new credential, `AGENTMUX_HOST_REG_SECRET`, known only to srv and the paired host — never given to any agent process (unlike `auth_key`, which agents do share).

- **Launcher-managed mode:** minted once per `spawn_srv` call (`agentmux-launcher/src/srv_spawner.rs`, `SrvSpawnResult::host_reg_secret`) and passed to both the host's and srv's spawn env (`host_spawn.rs` / `srv_spawner.rs`). Because `SrvSpawnResult` lives in the launcher's own stack frame across a host-only crash-restart (same lifetime as `auth_key`), a relaunched host automatically gets back the same secret srv already has.
- **Host-owned-spawn / dev mode:** the host generates it itself alongside its own `auth_key` and passes it to srv at `spawn_backend` time (no launcher in the loop, so no chicken-and-egg problem).
- `host_ipc.rs::handle_register` checks the caller-supplied secret against `state.host_reg_secret` via constant-time comparison (`secret_matches`, same "HMAC then `Mac::verify_slice`" idiom already used for the WhatsApp webhook signature check) **before** touching the `state.host_ipc` None/Some branching at all — so this closes the race unconditionally, not just in the already-registered case. If srv has no secret configured, it fails closed (rejects every registration) rather than accepting the first unauthenticated caller.

New tests: `host_ipc_register_rejects_without_the_matching_secret_even_when_state_is_none` (proves the exact race is closed), `host_ipc_register_rejects_everything_when_srv_has_no_secret_configured`, plus `config.rs` coverage for reading/scrubbing the env var. Updated the three existing `host_ipc.Register` tests to carry the secret so they still exercise the liveness/conflict logic they were written for, not just the new gate.

Full workspace build + `cargo test` across `agentmux-srv`, `agentmux-cef`, and `agentmux-launcher` all green (2453 + 254 + 213 tests passing, 0 failed).

PR #2664, opened 2026-08-19
feat(config): non-stable channels default to isolated per-channel settings.json

AgentA-asaf:
## Summary

Implements `docs/specs/SPEC_SETTINGS_ISOLATED_BY_CHANNEL_2026_08_19.md`.

`settings.json` currently resolves to `~/.agentmux/channels/settings.json` — a single file sitting *outside* every individual channel directory, so it's shared by every channel on the machine: every `task dev` branch, every `task package` build past and future, and the real `stable` install. This is documented as **deliberate** in both consumers' own comments (`resolve_settings_dir()` and `window_settings.rs` both explicitly call the shared location "the modern location"), not an oversight — but it means a fresh channel silently inherits whatever the shared file currently holds, indefinitely.

**Motivating case:** a fresh portable build was found shipping with `network:lan_discovery` already `true`, even though the code's own default is `false` in three independent places. Traced to exactly this sharing, not a wrong default — the setting was toggled on once, in some other channel, and every subsequent build has silently inherited it since.

**What this does**, mirroring `SPEC_ISOLATED_AUTH_DEFAULT_BY_CHANNEL_2026_08_06.md`'s already-shipped mechanism almost exactly (substituting `settings.json` for the identity store):

- New `agentmux_common::isolated_settings_enabled()` (+ `IsolatedSettingsReason` for boot-time diagnostics) — isolated by default for every channel except `stable`, explicit override via `AGENTMUX_ISOLATED_SETTINGS=0`/`=1`, conservative global fallback when `AGENTMUX_CHANNEL` is unset. Deliberately a **separate flag** from `AGENTMUX_ISOLATED_AUTH` (verified independent via a dedicated test) — see the spec's Open Questions §1 for why.
- `agentmux-srv`'s `resolve_settings_dir()`: on an isolated channel, uses `AGENTMUX_CONFIG_HOME` directly — that var already carries the correctly channel-scoped `channels/<ch>/config/` dir (per `DataPaths`' own layout comment) — instead of walking up two parent directories to the shared `channels/` root.
- `agentmux-cef`'s `read_window_transparent_setting()`'s `candidates()`: on an isolated channel, checks **only** the per-channel config dir — deliberately does not fall through to the shared-file candidates, so a channel whose own `settings.json` exists but lacks `window:transparent` reads the default instead of silently inheriting the shared file's value.
- **No seeding** — a freshly-isolated channel starts with no `settings.json` at all, same "blank slate" non-goal the auth spec chose, for the same reason: a seed-copy would carry `network:lan_discovery: true` right along with it, defeating the entire point.
- `stable` is completely unaffected — same file, same behavior, before and after.
- `CLAUDE.md` updated (mirrors the existing Armory-identity note's style).

## Test plan

- [x] `cargo test -p agentmux-common` — 153/153 (3 new: reason classification, fail-safe on malformed value, independence from `AGENTMUX_ISOLATED_AUTH`)
- [x] `cargo test -p agentmux-srv` — 2445 + 5 + 7 / all passed (6 new `resolve_settings_dir` cases: stable-stays-global, isolated-by-default, explicit opt-out, explicit opt-in overriding `stable`, unset-channel fallback, `AGENTMUX_SETTINGS_DIR` override still wins)
- [ ] Manual Linux/macOS check of `window:transparent` pre-`CefInitialize` on both an isolated dev channel and `stable` — flagged in the spec §9 item 6 as the highest-risk regression surface for this PR; not done here (Windows dev machine)


a5af:
@codex review

PR #2665, opened 2026-08-19
Fix macOS Keychain prompt storm from MuxBus's 12-entry chunked token storage

AgentO-asaf:
## Summary
- Follow-up to #2654 (which fixed one automatic keychain touch — the model-catalog RPC — but wasn't the only one).
- Root cause: `CloudSubscriber::init_global` (called unconditionally at startup, `bootstrap.rs:997`) reads MuxBus tokens from the keychain on every launch to auto-reconnect a cloud session. MuxBus tokens were chunked into **12 separate keychain entries** (3 fields × 4 sub-entries each) uniformly across every platform, to work around a real 1280-char Windows Credential Manager cap. macOS/Linux have no such cap, but paid for it anyway — each entry is its own independent OS consent decision, so "Always Allow" on one just surfaced the next entry's own prompt, making the whole flow look broken.
- Fix: macOS/Linux now write a single combined JSON blob (reusing the already-existing "legacy blob" format and its migration machinery, pointed in the other direction from what it was originally built for). Windows keeps its untouched, still-necessary chunked layout. A pre-fix macOS/Linux install self-heals the next time tokens are successfully read: reads the 12 old entries once, writes them as one blob, deletes the old ones.
- Adds a retro (`docs/retro/retro-macos-muxbus-keychain-prompt-storm-2026-08-19.md`) documenting the investigation, including a second, more severe finding that's deliberately scoped **out** of this PR per explicit direction: a keychain read requiring interactive OS consent can block indefinitely rather than failing fast (verified live — a diagnostic test blocked past 60s before failing with "User canceled the operation"). Fixing that needs a timeout wrapper around shared `identity/secret_store.rs` plumbing used well beyond MuxBus — tracked as a separate follow-up, not started here.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv muxbus::` — 27 passed, 0 failed (existing suite, unmodified — chunking math tests still cover the Windows path)
- [x] Verified live on `task dev` from an isolated worktree/branch (didn't disturb the concurrent main-branch dev session): fresh build compiles and runs
- [x] Verified via a temporary diagnostic test (removed before commit) against the real keychain on this machine — confirmed the 12 pre-existing chunked entries are genuinely present and that the migration/consolidation logic is reachable

## Note
This machine's actual 12 pre-fix chunked entries didn't get to self-heal during manual verification, because the read attempt itself hit the interactive-consent-hang issue described above (item 2, deliberately out of scope here) before it could complete. The consolidation logic itself is verified correct via the compiling/passing test suite and code review; the self-heal will complete on this machine once a read succeeds (e.g. after the follow-up timeout fix, or once "Always Allow" is granted for the one remaining entry).

### DIFF
--- a/.changesets/1787149266-fix-muxbus-collapse-macos-linux-keychain-token-storage-from--h4mq.md
+++ b/.changesets/1787149266-fix-muxbus-collapse-macos-linux-keychain-token-storage-from--h4mq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(muxbus): collapse macOS/Linux keychain token storage from 12 entries to 1, stopping the multi-prompt Keychain access storm
--- a/agentmux-srv/src/backend/storage/muxbus.rs
+++ b/agentmux-srv/src/backend/storage/muxbus.rs
@@ -684,6 +684,50 @@ impl Store {
                     let tokens: MuxBusTokens = serde_json::from_str(&blob).map_err(|e| {
                         StoreError::Other(format!("muxbus: stored keychain blob is corrupted: {e}"))
                     })?;
+
+                    // Codex P1, PR #2665: a blob and split entries can
+                    // coexist on this same machine if an EARLIER blob→split
+                    // migration (the pre-2026-08-03 upgrade path, which ran
+                    // unconditionally on every platform before this fix
+                    // existed) wrote the split entries successfully but its
+                    // own best-effort blob delete then failed or was
+                    // interrupted. Split is always the fresher data in that
+                    // case — it's only ever written FROM a blob, never the
+                    // reverse — so a hit here doesn't necessarily mean the
+                    // blob is authoritative; check for that coexistence and
+                    // prefer split, finishing the earlier interrupted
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentO-asaf:
Addressed in 8f08d8eaf — good catch. A blob and split entries coexisting (from an earlier interrupted blob→split migration) is now detected on a blob hit: split is preferred (it's always the fresher data in that scenario) and the earlier interrupted cleanup gets finished by migrating it into this fix's single-blob format. 27/27 muxbus tests pass, workspace `cargo check --tests` clean.

AgentO-asaf:
Addressed in afb94bdeb — you're right, the hardcoded direction was wrong. The single-blob format now carries its own `new_generation()` timestamp stamp (same scheme the chunked layout already uses), and coexistence is resolved by an actual freshness comparison instead of an assumed direction. An old pre-generation legacy blob still correctly loses to any real split generation via `#[serde(default)]`. 31/31 muxbus tests pass (27 existing + 4 new covering the comparison logic), workspace `cargo check --tests` clean.

PR #2666, opened 2026-08-19
fix(srv): bump sysinfo 0.34→0.35 to fix CreateToolhelp32Snapshot handle leak

AgentX-asaf:
## Summary

`sysinfo` v0.34.2's Windows `refresh_processes_specifics()` never closes the
handle it gets from `CreateToolhelp32Snapshot` — on any code path (success
or either early-return error branch). That handle is backed by a kernel
`Section` object, so every call leaks one — invisible to the process's own
working-set/private-bytes, only visible as a climbing
`GetProcessHandleCount`.

This is the confirmed root cause of
`docs/status/STATUS_SRV_SECTION_HANDLE_LEAK_2026_08_08.md` (still open) and
today's live recurrence,
`docs/status/STATUS_SRV_SECTION_HANDLE_LEAK_LIVE_RECURRENCE_2026_08_19.md`
(§7 has the full evidence chain — direct read of the vendored dependency
source, cross-checked against Microsoft's own docs and an independent
RustSec advisory for the identical bug pattern in a different crate). Live
on this machine at time of writing: one `agentmux-srv` instance at 517K
total handles, 500K of them `Section` — a 3-day-old process, still growing
(~27 handles/min measured live).

`run_sysinfo_loop` calls `refresh_processes_specifics` ~2×/tick (default 1s
interval) plus once more every 30s in the commit-attribution pass — so this
leaks continuously for the lifetime of every `agentmux-srv` process,
eventually contributing to system-wide page-file/commit exhaustion.

## Fix

Bump `sysinfo` `"0.34"` → `"0.35"` in `agentmux-srv/Cargo.toml`. Confirmed
via direct source comparison that the handle-close fix (an `Owned` RAII
wrapper) landed between 0.34.2 and 0.35.0. Deliberately targeting the
minimum fixed version rather than current stable (0.39.6) — that version
requires Rust 1.95, ahead of this repo's toolchain, and pulls in an
unrelated `windows` crate 0.57→0.62 major bump. A follow-up 0.39 upgrade can
happen separately once the toolchain question is resolved.

## Verification

- `cargo build -p agentmux-srv` (release) — clean.
- `cargo test -p agentmux-srv --bin agentmux-srv backend::sysinfo::` — all
  15 existing tests pass unchanged.
- **New regression test**,
  `refresh_processes_specifics_does_not_leak_a_handle_per_call` — calls
  `refresh_processes_specifics` 500× (the same pattern `run_sysinfo_loop`
  uses) and asserts the process's own handle count doesn't grow linearly.
- **Proved the test is a real discriminator**: temporarily re-pinned to the
  broken `sysinfo = "0.34.2"` and reran — it fails, reporting *"handle count
  grew by 506 over 500 calls"* (≈ exactly 1/call, matching the theorized
  mechanism exactly). Restored `0.35` — passes clean.

Not done in this PR (tracked as follow-up in the status doc): a multi-hour
live-production soak test confirming `handle64 -s -p <pid>`'s `Section`
count stays flat on a real running instance over realistic uptime — the
500-call synthetic test is a fast, proven-accurate proxy but not a
substitute for confirming in the actual production loop over hours/days.

## Test plan

- [x] `cargo build -p agentmux-srv`
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::sysinfo::`
- [x] New regression test passes against the fix, fails against the bug
- [ ] Live soak test on a real instance (follow-up, not blocking)

### DIFF
--- a/.changesets/1787152102-fix-srv-bump-sysinfo-to-0-35-to-fix-createtoolhelp32snapshot-nbgb.md
+++ b/.changesets/1787152102-fix-srv-bump-sysinfo-to-0-35-to-fix-createtoolhelp32snapshot-nbgb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): bump sysinfo to 0.35 to fix CreateToolhelp32Snapshot handle leak
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1017,7 +1017,7 @@ dependencies = [
  "num-traits",
  "serde",
  "wasm-bindgen",
- "windows-link",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -1871,7 +1871,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1bd49230192a3797a9a4d6abe9b3eed6f7fa4c8a8a4947977c6f80025f92cbd8"
 dependencies = [
  "rustix 1.1.4",
- "windows-link",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2668, opened 2026-08-19
fix(sound): tool-call tones default to max volume; centralize gain defaults

Agent2-asaf:
## Summary
- Tool-call tones (`notify:tooltones:volume`) default raised `0.15` → `1.0`. Master notification volume (`notify:sounds:volume`) stays `0.6`, unchanged — per explicit direction after an initial pass that touched the wrong control.
- Fixed the underlying cohesiveness gap that made this exact kind of change error-prone: each gain default was a duplicated literal across 5+ files with no single source of truth. Added `frontend/app/notification/sound/sound-defaults.ts` (`DEFAULT_MASTER_VOLUME`/`DEFAULT_TOOLTONES_VOLUME`/`DEFAULT_WAITING_VOLUME`); every consumer (`sound-player.ts`, `tool-tones-player.ts`, `waiting-tone-player.ts`, `sound-service.ts`, `sounds-section.tsx`) now imports from it.
- Verified (not fixed — it was already correct) that the underlying architecture is a single shared `AudioContext` + one master `GainNode`, with tool-tones and waiting-tones correctly layered underneath it as independent child gains, so master enable/volume genuinely governs everything downstream.
- Includes two specs from the same audit pass:
  - `docs/specs/SPEC_SETTINGS_RECORDING_INPUT_SECTION_2026_08_19.md` — new Settings section design for the currently-UI-less voice/mic system.
  - `docs/specs/SPEC_SETTINGS_AUDIT_GOOD_PICKINGS_2026_08_19.md` — ranked audit of other real, shipped-but-undiscoverable settings.

## Test plan
- [x] `npx vitest run frontend/app/notification/sound` — 64/64 passing
- [x] `npx vitest run frontend/app/view/settings` — 70/70 passing

### DIFF
--- a/.changesets/1787170221-fix-sound-tool-call-tones-default-to-max-volume-centralize-s-xyum.md
+++ b/.changesets/1787170221-fix-sound-tool-call-tones-default-to-max-volume-centralize-s-xyum.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(sound): tool-call tones default to max volume; centralize scattered gain-default literals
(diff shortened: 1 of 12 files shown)


a5af:
@codex review

PR #2670, opened 2026-08-19
feat(settings): make the AskUserQuestion auto-answer timeout user-configurable

Agent2-asaf:
## Summary
Was a hardcoded 30s constant, explicitly flagged as a spec-anticipated follow-up per `SPEC_ASK_USER_QUESTION_AUTO_TIMEOUT_2026_08_06.md` §5.2: *"a per-user or per-agent override is a reasonable follow-up if requested later... nothing here blocks adding one afterward."*

- New setting `agent:askquestiontimeoutms` (integer ms, default 30000, minimum 1000).
- Frontend-only — no Rust struct field, following the same precedent as the existing `dnd:*` settings (schema + `gotypes.d.ts` only, no `types.rs` change needed since nothing server-side reads it).
- Exposed in Settings → Advanced as a seconds-based numeric input, matching the existing `telemetry:interval`/`telemetry:numpoints` pattern in that section.
- The hover-pause grace window (`HOVER_HIDE_GRACE_MS`, 15s) stays hardcoded — only the auto-timeout itself was named as the anticipated injection point in the original spec.

Settings audit candidate #5 — `docs/specs/SPEC_SETTINGS_AUDIT_GOOD_PICKINGS_2026_08_19.md`.

## Test plan
- [x] `npx vitest run frontend/app/view/agent/components/AgentQuestionPanel.test.tsx` — 21/21 passing (unchanged timing behavior, confirms the default-fallback path is exercised correctly by all existing 30s-timing tests)
- [x] `npx vitest run frontend/app/view/settings` — 6/6 passing
- [x] `schema/settings.json` validated as parseable JSON after the edit

### DIFF
--- a/.changesets/1787170999-feat-settings-make-the-askuserquestion-auto-answer-timeout-u-14b4.md
+++ b/.changesets/1787170999-feat-settings-make-the-askuserquestion-auto-answer-timeout-u-14b4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(settings): make the AskUserQuestion auto-answer timeout user-configurable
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2673, opened 2026-08-20
docs(status): record 4-hour soak test confirming the sysinfo handle-leak fix

AgentX-asaf:
## Summary

Follow-up to #2666 — ran the merged fix (`agentmux-0.55.16+g376006730`) as a
live, isolated 4-hour soak test against the real `run_sysinfo_loop` process,
sampling `handle64 -s -p <pid>` every 15 minutes.

**Result: Section handles held flat at 11 for the entire 4-hour run** (one
transient `0` reading, immediately followed by `11` again for the remaining
3h15m — a read race, not a real drop). Total handles/memory grew normally
from ordinary activity (816→3,091 handles, 26→122MB) — expected churn in
other handle types, not the leaked mechanism this fix addresses.

This is the closing verification flagged as outstanding in #2666 — confirmed
in the actual production code path over real uptime, not just the 500-call
synthetic regression test that PR added. Closes out both
`STATUS_SRV_SECTION_HANDLE_LEAK_LIVE_RECURRENCE_2026_08_19.md` and the
original `STATUS_SRV_SECTION_HANDLE_LEAK_2026_08_08.md`.

Docs-only change, no code.


a5af:
@codex review

PR #2674, opened 2026-08-20
feat(memory): version-controlled native memory history (core + MCP tools)

Agent1-asaf:
## Summary

A fabricated memory file sat undetected in an agent's native memory store for 8 days this session (traced in detail in the linked spec's §1) because nothing captures history for native memory or Armory bundles — every write silently overwrites the last, with no diff, no audit trail, no way to know a write traced back to unverified/jekt-originated content.

This PR (1 of 3 stacked PRs) adds real version history for native memory:

- New `db_agent_native_memory_versions` table, added everywhere `db_agent_native_memory` already lives (mirroring its own dual/triple-schema duplication), append-only, chained via `parent_version_id`.
- `agent:memory:write_file` (WS RPC) and `memory.write`/`memory_write_impl` (the App-API path the actual `MemoryWrite` MCP tool hits) both record a version now, source-tagged via optional caller-supplied `provenance` (`human` / `agent_inferred` / `jekt` / `revert`).
- Three new RPCs — `agent:memory:history` / `diff` / `revert` — plus matching App-API + HTTP + MCP surfaces (`MemoryHistory`, `MemoryDiff`, `MemoryRevert` tools). Revert is implemented as a **new** version (git-revert style), never a rewrite of history.
- Full design + research write-up: `docs/specs/SPEC_MEMORY_VERSION_CONTROL_AND_ARMORY_AUDIT_2026_08_19.md` — includes why this does **not** add a Bundle↔Memory FK (§4.2/§2.5: three pre-existing, unreconciled Agent↔Bundle pointers already exist; adding a fourth was rejected after research, operator-confirmed).
- A separate, unimplemented north-star doc surfaced during that research is included for reference: `docs/specs/ARCHITECTURE_ARMORY_FOUNDATION_CONSOLIDATION_2026_08_19.md` (Armory naming/binding cleanup, out of scope for this PR).

Stacked on top of this (separate PRs, both based on this branch):
- fs-watch out-of-band write detection (spec §4.5)
- Frontend UI (Stash history view + new Armory "Native Memory" tab)

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2463 passed, 0 failed, 6 ignored (pre-existing)
- [x] `cargo test -p agentmux-mcp` — 4 passed, 0 failed (includes the tool-count/JSON-validity assertion, updated 27→30 across all 3 stacked PRs' worth of tools)
- [x] Verified this branch compiles and tests pass standalone (stashed the two follow-on PRs' changes before testing)

PR #2675, opened 2026-08-20
feat(memory): detect native memory writes that bypass AgentMux's RPCs

Agent1-asaf:
## Summary

Stacked on #2674 (base branch: `agent1/memory-version-core`) — this PR (2 of 3) closes the gap that PR still leaves open: a memory write made outside AgentMux's own RPCs entirely.

Claude Code's CLI writes memory `.md` files directly to `~/.claude/projects/<hash>/memory/`, with no obligation to call `agent:memory:write_file`/`MemoryWrite`. An agent's own general-purpose filesystem tools can write there directly too — this is not hypothetical, it's what actually happened in the session that motivated the base spec (`docs/specs/SPEC_MEMORY_VERSION_CONTROL_AND_ARMORY_AUDIT_2026_08_19.md` §1).

Two independent layers (spec §4.5), so failure of one degrades rather than blinds the whole system:

- **Fast path** — subscribes to every known agent's memory directory via the existing (previously unused in production) `FsWatchPool`, records a new version tagged `external_fs_write` on any change whose hash doesn't match the latest recorded version.
- **Slow path** — a periodic reconciliation sweep, same cadence class as `FsWatchPool`'s own `HEALTH_SWEEP_INTERVAL` (30s), hashing every known agent's live memory files against their latest recorded version. This is the backstop for a missed fs-watch event, a watch still in retry-backoff, or srv having been down entirely when the write happened.

Neither promises zero data loss — see the spec's §4.5 "precision, honestly bounded" section for the exact, accepted limits (e.g. two out-of-band writes between one sweep and the next collapse into one detected version).

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2470 passed (7 new over the base branch's 2463), 0 failed, 6 ignored (pre-existing)
- [x] Includes a unit test that a plain filesystem write (bypassing every RPC) is detected and recorded correctly

### DIFF
--- a/agentmux-srv/src/backend/mod.rs
+++ b/agentmux-srv/src/backend/mod.rs
@@ -21,6 +21,7 @@ pub mod config_watcher_fs;
 pub mod editor_file_watcher;
 pub mod fs_watch;
 pub mod media_file_watcher;
+pub mod native_memory_drift;
 pub mod ijson;
 pub mod docsite;
 pub mod eventbus;
--- a/agentmux-srv/src/backend/native_memory_drift.rs
+++ b/agentmux-srv/src/backend/native_memory_drift.rs
@@ -39,7 +39,7 @@ use std::time::Duration;
 
 use crate::backend::fs_watch::{FsWatchEventKind, FsWatchPool};
 use crate::backend::storage::store::Store;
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2676, opened 2026-08-20
feat(agent-pane): render Agent/Task/Workflow tool calls as live dispatch cards [Naki@claudius]

GenericAgentX-asaf:
## Summary

Replaces the raw JSON dump the main transcript pane used for `Agent`/`Task`/`Workflow` tool calls with a compact card showing the call's live dispatch status (name, running/done, member count for a Workflow) — reusing the same dispatch data the Swarm panel already tracks for the identical call, instead of duplicating a long unstructured list inline.

- No exact `tool_use_id -> dispatch_id` link exists in the backend, so this uses **ordinal matching**: within one pane, zip its `Agent`/`Task`/`Workflow` tool nodes (transcript order) against its dispatches (spawn order, reconstructed client-side since `ListDispatches`/`ListActive` sort by `last_event_at`, not spawn order) by position. All-or-nothing per pane — falls back to today's generic rendering on any ambiguity (parallel same-turn spawns, stale/pruned dispatch data), since a wrong card is worse than no card.
- Adds `"Workflow"` as a recognized tool type across all 7 switch/case sites that needed it (previously collapsed to `"Other"`).
- New renderer registered via the existing tool-renderer registry's priority system; falls back to the existing `CompactResult` renderer whenever no confident match exists.

See `docs/reports/REPORT_AGENT_TRANSCRIPT_DISPATCH_CARDS_2026_08_19.md` for the full writeup.

## Test plan

- [x] `npx vitest run` — full frontend suite passes (2801 tests / 190 files)
- [x] New tests: `dispatch-correlation.test.ts` (6 cases covering the ordinal-match confidence gate)
- [x] `task build:frontend` production build succeeds
- [ ] Interactive/visual verification not done — native desktop app, this machine runs shared live instances; see report for details


a5af:
@codex review

PR #2677, opened 2026-08-20
fix(swarm): robust dispatch attribution and formalized session lifecycle

GenericAgentX-asaf:
## Summary

Root-causes a real reported symptom: a session with only "a couple" of real `Agent`-tool calls showed **48 entries** in the Swarm panel, including rows that visually read as duplicates. Not one bug — accumulation (the backend never prunes historical subagents, and the only dismiss mechanism was ephemeral in-memory state) plus a workflow-member spillage bug, compounded by dispatches having no "dead" status of their own and a reconciliation pass that could silently no-op.

- `buildDispatchBuckets` no longer spills a lagging Workflow dispatch's members individually into the flat Agent-Tool bucket — grouped into one synthesized placeholder row.
- Added `DispatchStatus::Abandoned` (Rust) with proper aggregation from member reconciliation. Caught and fixed a real bug along the way: `refresh_dispatch_status` (called on *every* `ListDispatches` read) was silently clobbering a freshly-set `Abandoned` back to `Running`/`Completed`.
- Fixed a flagged-but-never-confirmed race where `reconcile_stale_subagents` could run before a block's controller registers, permanently treating "unknown" the same as "confirmed active" — now retries once (bounded, not a loop) before giving up.
- Retired/dismissed rows now persist to `localStorage` (local-machine scope) instead of resetting on every reload/restart, plus a new bulk **"Clear completed (N)"** action so a human can resolve a large historical backlog in one click.
- Also fixed a follow-on gap caught while wiring this up: the frontend `AgentDispatch.status` TS type and the `WorkflowDispatch.status` active/retired mapping didn't account for the new `"abandoned"` value at all — an abandoned Workflow dispatch would have silently read as still "active".

Deliberately descoped: a full Live/Historical collapsed-section UI split — real information-architecture work that deserves its own design pass, tracked as a follow-up in the spec rather than rushed in here.

See `docs/specs/SPEC_SWARM_DISPATCH_ATTRIBUTION_AND_LIFECYCLE_2026_08_19.md` and `docs/reports/REPORT_SWARM_DISPATCH_ATTRIBUTION_AND_LIFECYCLE_2026_08_19.md` for the full investigation and design rationale.

## Test plan

- [x] `cargo test -p agentmux-srv` — full suite passes (2472 tests, 0 failed)
- [x] `npx vitest run` — full frontend suite passes (2801+ tests)
- [x] New tests: 6 Rust (dispatch abandonment aggregation, reconcile retry paths), 13 frontend (orphaned-member grouping, retire persistence, bulk-clear eligibility, abandoned-status mapping regression)
- [ ] Interactive/visual verification not done — see report for why

### DIFF
--- a/.changesets/1787196269-fix-swarm-robust-dispatch-attribution-and-formalized-session-n5fp.md
+++ b/.changesets/1787196269-fix-swarm-robust-dispatch-attribution-and-formalized-session-n5fp.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(swarm): robust dispatch attribution and formalized session lifecycle
--- a/agentmux-srv/src/backend/subagent_watcher/jsonl.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/jsonl.rs
@@ -627,26 +627,24 @@ impl SubagentWatcher {
         state.info.member_count = state.journal_started.max(state.member_files);
         state.info.members_done = state.journal_results.max(state.members_completed);
         state.info.last_event_at = now_millis();
-        Self::refresh_dispatch_status(state);
+        // Called only in response to genuinely NEW member evidence (a spawn
+        // or a completion just landed) — always allowed to recompute, even
+        // overriding a prior Abandoned, mirroring the existing SubAgent-level
+        // precedent that a late-arriving Result always wins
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

PR #2678, opened 2026-08-20
feat(memory): version history UI in Stash and a new Armory Native Memory tab

Agent1-asaf:
## Summary

Stacked on #2674 (base branch: `agent1/memory-version-core`) — this PR (3 of 3) is the UI for the version history/diff/revert RPCs that PR adds.

- Shared `NativeMemoryHistoryPanel` component (`agentId`+`filename` props) — version list with source tags (human/agent_inferred/jekt-with-trust-marker/external_fs_write/revert), diff between any two versions, revert with confirmation.
- Mounted from two places, reading the identical backend RPCs — one source of truth, no copy/migration step between them:
  - Stash's own **Memory** tab (`AgentNativeMemoryModal`) — a "History" toggle next to the existing content view.
  - A new Armory **Native Memory** rail tab (`NativeMemoryManager`) — an agent picker in front of the same panel, for browsing any agent's memory history without opening that agent's own pane.
- Deliberately named/placed away from `frontend/app/view/memory/` (Armory's *Bundles* tab — `MemoryManager`, a legacy name predating the Preset→Bundle rename) to avoid deepening an existing naming collision — see `docs/specs/SPEC_MEMORY_VERSION_CONTROL_AND_ARMORY_AUDIT_2026_08_19.md` §4.3.
- Includes a follow-up fix: threads `agent_id` through the `agent:memory:diff` call to match #2674's ownership-check fix (the Rust/TS wire shapes aren't type-linked, so `tsc` alone didn't catch the mismatch after that PR's fix landed).

## Test plan

- [x] `tsc --noEmit` clean
- [x] `vitest run` — 1236 passed across the touched areas (agent/armory views), 0 failed
- [x] Manually traced the "one source, two views" property: both mount points call the identical `agent:memory:history/diff/revert` RPCs with no intermediate cache/copy

### DIFF
--- a/frontend/app/store/rpc-api/memory.ts
+++ b/frontend/app/store/rpc-api/memory.ts
@@ -74,7 +74,12 @@ export const MemoryApi = {
 
     NativeMemoryDiffCommand(
         client: RpcClient,
-        data: { from_version_id: string; to_version_id: string },
+        // agent_id required (reagent P1 on agent1/memory-version-core):
+        // the backend verifies BOTH versions belong to this agent before
+        // returning their content — every caller shares one instance-wide
+        // X-AuthKey, so without it any caller could read any other
+        // agent's memory content by version id.
+        data: { agent_id: string; from_version_id: string; to_version_id: string },
         opts?: RpcOpts,
     ): Promise<NativeMemoryDiffResult> {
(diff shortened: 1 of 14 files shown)


a5af:
@codex review

PR #2679, opened 2026-08-20
Bound Keychain reads/writes with a 15s timeout to stop indefinite hangs

AgentO-asaf:
## Summary
- Follow-up to #2665, explicitly scoped out of that PR pending its own pass (this is that pass).
- Root cause: `identity::secret_store`'s `get`/`get_optional`/`put`/`delete` are synchronous `keyring`-crate calls that on macOS can require an interactive Keychain consent dialog on first access — and that call has no cancellation mechanism. An unanswered prompt (headless process, no attached display session, dialog on another Space) blocks the underlying platform call forever. Several callers (`muxbus.rs`'s `muxbus_save_lock`-guarded sequence in particular) hold a mutex for the entire duration, so a stuck read doesn't just block its own caller — it blocks every other operation waiting on that same lock.
- Fix: each `secret_store` function now runs the real platform call on a detached thread and waits up to 15s via a channel (`recv_timeout`). **No signature change, no caller updates needed anywhere in the codebase** — a timeout produces the same `Result<_, String>` error shape these functions already returned for other keychain failures, so every existing caller's error handling covers it for free.
- This is a wait-bound, not true cancellation — the detached thread keeps running until the OS resolves it. A leaked thread per stuck prompt is a far better failure mode than an indefinitely blocked, lock-holding caller.
- **Verified live** against this exact machine's real stuck Keychain entries (the ones blocking #2665's self-heal migration): the same read that previously blocked for minutes-plus now returns after exactly 15.006s with a clear timeout error.
- Adds a retro (`docs/retro/retro-secret-store-keychain-read-timeout-2026-08-20.md`) and updates the prior retro's follow-up section to point at it.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv secret_store::` — 3 passed (incl. new `run_with_timeout` tests: returns the value when work finishes in time, gives up when work outlives the deadline)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv muxbus::` — 35 passed, unaffected
- [x] Live-verified against real (not mocked) Keychain state on this machine via a temporary diagnostic test (removed before commit): confirmed the exact stuck read now resolves after 15.006s instead of hanging

### DIFF
--- a/.changesets/1787210241-fix-identity-bound-keychain-reads-writes-with-a-15s-timeout--kwu8.md
+++ b/.changesets/1787210241-fix-identity-bound-keychain-reads-writes-with-a-15s-timeout--kwu8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): bound keychain reads/writes with a 15s timeout so an unanswered OS consent prompt fails fast instead of hanging indefinitely
--- a/agentmux-srv/src/identity/secret_store.rs
+++ b/agentmux-srv/src/identity/secret_store.rs
@@ -48,12 +48,30 @@ pub const SERVICE: &str = "agentmux";
 /// (e.g. `muxbus_save_lock`) indefinitely. See this module's doc comment.
 const TIMEOUT: Duration = Duration::from_secs(15);
 
-/// Run `f` on a detached thread and wait up to `timeout` for it. `Timeout`
-/// is a distinct outcome from any error `f` itself can return — callers
-/// that want it folded into their own error type do that at the call site
+/// Why `run_with_timeout` gave up waiting — reagent P2: collapsing these
+/// into one outcome made a closure PANIC (sender dropped without sending,
+/// `mpsc::RecvTimeoutError::Disconnected`) misreport as "timed out after
+/// 15s" — actively misleading, since the operation actually failed
+/// immediately, not after waiting the full deadline.
+#[derive(Debug, PartialEq)]
+enum RunOutcome {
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentO-asaf:
Addressed both in 6998aa154:
- **P1** (unscrubbed plaintext copy): `put()` now wraps the moved secret in `Zeroizing` before the move, matching `get`/`get_optional`'s existing guarantee.
- **P2** (panic misreported as timeout): `run_with_timeout` now returns a `RunOutcome` distinguishing `TimedOut` from `WorkerPanicked` (a dropped sender via `RecvTimeoutError::Disconnected`), each with an accurate error message.

5/5 secret_store tests pass (incl. 2 new: panic-vs-timeout distinction, message content), 35/35 muxbus tests pass, workspace `cargo check --tests` clean.

AgentO-asaf:
Addressed in be9648519 — you're both right that a timed-out mutation completing later is a real corruption risk I hadn't accounted for, and there's no way to actually cancel the underlying platform call to prevent it. Rather than try to build ordering-preserving compensation for writes in this pass, I scoped the timeout down to `get`/`get_optional` only — reads are safe to bound since nothing acts on a stale/discarded late result. `put`/`delete` revert to their original unbounded (safe-if-slow) behavior. This still fixes the actual, live-verified, motivating bug (the stuck self-heal *read* from #2665), just without introducing a new write-ordering risk to fix it. Retro updated to document the narrower scope and why. All tests still pass, workspace `cargo check --tests` clean.

PR #2683, opened 2026-08-20
fix(agent): declared-background tasks survive a session restart, not just idle-timeout

### LINKED ISSUE #2492: Backgrounded agent tasks die on session teardown — a running dev server shouldn't die because the agent session restarted

AgentA-asaf:
## Summary

Closes #2492. A declared-background task (`task dev` via `run_in_background: true`) was dying whenever its owning agent session restarted, even after #2491 fixed the idle-timeout kill.

**Root cause:** `resync_controller`'s replace path (session restart/resync) called `delete_controller` — the same function real pane/tab/workspace deletion uses — which unconditionally tears down the block's shared process tracker (a Windows Job Object). Any declared-background descendant died with it, even though a restart only intends to swap the CLI controller, not kill everything the block ever spawned.

**Fix:** `AgentProcessRegistry::ensure_tracker` was already idempotent, and its own doc comment says the tracker is meant to survive controller re-creation ("e.g. on `/clear`") — this was a case of calling the wrong cleanup function, not a missing mechanism. Added `Controller::stop_for_replace` (default delegates to `stop()`), overridden by `ShellController` to kill only its own CLI process (single PID, not the whole process group `stop()` targets) via the existing `AgentProcessRegistry::kill_pid`, with a direct single-PID signal fallback on Unix. `resync_controller`'s replace branch now calls `stop_for_replace` + a new `remove_controller_entry_only` (registry-map only, no tracker/broker teardown) instead of `stop` + `delete_controller`. Real deletion paths (`delete_tab`, `delete_block`, `wcore/tab.rs`) are unchanged — a background task still dies on genuine pane close, just not on restart.

**Note on design history (transparency, not asking for re-review of an abandoned path):** an earlier draft of `docs/specs/SPEC_BACKGROUND_TASK_TEARDOWN_SURVIVAL_2026_08_20.md` proposed a Windows Job Object breakaway mechanism. While starting implementation I found the per-block Job Object already carries a deliberate, documented decision *against* allowing breakaway (`process_tracker/windows.rs`'s own comment: "we want those attempts to fail so the child stays tracked"). That design would have weakened a real existing containment guarantee for every process in a block's tree, not just the one task this PR cares about — so it was abandoned in favor of this approach, which **does not touch Job Object flags or any containment guarantee at all**. §0 of the spec (revised in this same PR) documents the abandoned approach and why.

## Test plan
- [x] New unit tests: default `stop_for_replace` delegates to `stop()`; `remove_controller_entry_only` removes only the registry entry (doesn't call `stop`); `resync_controller`'s replace path calls `stop_for_replace` (not `stop`) on the old controller — 3 new tests, all passing
- [x] `cargo test -p agentmux-srv blockcontroller` — 209/209 pass, zero regressions
- [x] `cargo test -p agentmux-srv` (full suite) — 2489 + 5 + 7 pass, zero failures
- [ ] Live verify (background `task dev`, force a session restart, confirm survival) — planned as part of this effort's final end-to-end pass once Phase C also lands, per the spec's §5

### DIFF
--- a/.changesets/1787236804-fix-agent-declared-background-tasks-survive-a-session-restar-gqze.md
+++ b/.changesets/1787236804-fix-agent-declared-background-tasks-survive-a-session-restar-gqze.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): declared-background tasks survive a session restart, not just idle-timeout
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,6 +29,7 @@ dependencies = [
  "chrono",
  "clap",
  "dirs",
+ "libc",
  "portable-pty",
  "reqwest",
  "serde",
--- a/agentmux-bashwrap/Cargo.toml
+++ b/agentmux-bashwrap/Cargo.toml
@@ -33,3 +33,9 @@ portable-pty = "0.9"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 dirs = "5"
+
+[target.'cfg(unix)'.dependencies]
+# setsid() for declared-background invocations — see bash_wrap.rs's
+# `detach_declared_background_session` and
+# docs/specs/SPEC_BACKGROUND_TASK_TEARDOWN_SURVIVAL_2026_08_20.md.
+libc = "0.2"
--- a/agentmux-bashwrap/src/bash_wrap.rs
+++ b/agentmux-bashwrap/src/bash_wrap.rs
@@ -327,6 +327,69 @@ fn effective_idle_timeout(args: &Args) -> Duration {
     }
 }
 
+/// For a `--declared-background` invocation, detach this process into a
+/// brand-new session (`setsid()`) before doing anything else, so it has no
+/// controlling terminal at all.
+///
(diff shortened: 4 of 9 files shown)


a5af:
@codex review

PR #2684, opened 2026-08-20
fix(armory): use block-tile app icon for AgentMux tile, shorten Cloud subtitle

a5af:
## Summary
- The Armory Accounts AgentMux tile was rendering blank (a separate root cause already fixed upstream in #2196: gradient-id collision when the icon mounts twice). Independently of that, swap the tile's icon from the brain-alternate logo to the app's own block-tile icon (same mark as the taskbar/favicon), per user preference.
- Shorten the AgentMux tile subtitle from "AgentMux Cloud sign-in" to "Amux Cloud".

## Test plan
- [x] `tsc --noEmit` passes for the touched files
- [x] Verified the vendored PNG renders correctly (visual check)
- [ ] Manual check in a running dev build: Armory → Accounts tab shows the block-tile icon + "Amux Cloud" subtitle on the AgentMux tile


a5af:
@codex review

PR #2685, opened 2026-08-20
feat(dashboard): make db_background_tasks a real read source for the agent-pane and muxspect dock

AgentA-asaf:
## Summary

Phase C (final phase) of the background-task registry work — `db_background_tasks` (Phase A/B) has been write-only from the frontend's perspective until now: nothing ever reads it back into the live UI, so a task that survives a session restart (#2683) under a fresh controller generation has no transcript history to re-derive its state from.

- New request/response RPC `COMMAND_LIST_BACKGROUND_TASKS` (`listbackgroundtasks`) returns a block's current `db_background_tasks` rows, reusing the existing `muxspect_handlers::BackgroundTaskView` wire type.
- The three existing observe/complete/pid handlers now publish a live-only `background-task-updated` WPS event (block-scoped, `persist: 0`, mirroring `process_tracker::registry`'s `emit()`) as a pure invalidation signal — no task data in the payload, subscribers just re-query.
- New `useBackgroundTaskRegistry` hook seeds/refreshes the `attachedTask` reducer axis from the registry on mount and on that event. **Deliberately additive-only** — never dispatches `AttachedTaskCleared` — since the registry can be a moment stale relative to a live transcript signal; completion stays the more specific `<task-notification>`-parsing path's job alone, per the spec's "registry is a floor, not an override" framing.
- `muxspect dock`'s 1-hour `DockSnapshotCache` TTL blind spot (a `task dev` session in this repo's own retros has run 12+ hours) is fixed by merging in still-`Running` `db_background_tasks` rows the cache evicted — without touching the cache's own eviction semantics, which are correct for what the cache is actually for.

**Explicitly out of scope** (per the spec's §1, verified against current code before writing it): no change to `turnPhase`'s state machine (the "stuck Working…" bug is already fixed, confirmed via a retro's own correction notice), no re-added footer "Running in background" text (tried once, reverted on direct user feedback 2026-08-10 as redundant with the ActivityDock's own row). No Swarm-pane wiring in this PR — flagged in the spec as a natural, low-risk follow-on once this lands, not bundled here to keep review surface focused.

Spec: `docs/specs/SPEC_BACKGROUND_TASK_DASHBOARD_INTELLIGENCE_2026_08_20.md`.

## Test plan
- [x] New unit tests: `merge_background_tasks` (muxspect dock TTL merge logic, 4 tests) and `resolveAttachedTaskObservation` (frontend earliest-running-task logic, 5 tests)
- [x] `cargo test -p agentmux-srv` (full suite) — 2547 + 5 + 7 pass, zero failures
- [x] Full frontend vitest suite — 2799 tests, 2798 pass + 1 pre-existing timing flake unrelated to this change (confirmed passes in isolation: `tool-renderer registry` parity test, no connection to background tasks)
- [ ] Live end-to-end verify (background `task dev`, force a session restart, confirm the dashboard reflects survival accurately) — final step once this merges, closing out the whole three-phase effort

### DIFF
--- a/.changesets/1787240445-feat-dashboard-make-db-background-tasks-a-real-read-source-f-2ghb.md
+++ b/.changesets/1787240445-feat-dashboard-make-db-background-tasks-a-real-read-source-f-2ghb.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(dashboard): make db_background_tasks a real read source for the agent-pane and muxspect dock
--- a/agentmux-srv/src/backend/rpc_types/block.rs
+++ b/agentmux-srv/src/backend/rpc_types/block.rs
@@ -195,6 +195,15 @@ pub struct CommandBackgroundTaskPidData {
     pub pid: u32,
 }
 
+/// Data for `COMMAND_LIST_BACKGROUND_TASKS` — request/response, returns
+/// this block's current `db_background_tasks` rows (as
+/// `muxspect_handlers::BackgroundTaskView`s). See
+/// docs/specs/SPEC_BACKGROUND_TASK_DASHBOARD_INTELLIGENCE_2026_08_20.md §3.1.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct CommandListBackgroundTasksData {
+    pub blockid: String,
+}
+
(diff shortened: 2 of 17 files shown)


a5af:
@codex review

PR #2687, opened 2026-08-20
feat(fleet): multi-agent fleet control — select, broadcast, bulk-stop

AgentY-asaf:
## Summary

Implements `docs/specs/SPEC_MULTI_AGENT_FLEET_CONTROL_2026_08_20.md` — the repo owner asked for tools to control many agents at once, prompted by recent Swarm-pane work. The spec (research-backed: internal codebase audit + external best-practices survey of AutoGen/Magentic-One, LangGraph, K9s/Lens, Ansible, tmux, PM2) recommended building this into **Swarm** rather than Warden or a new pane — Swarm's own widget description is already "Multi-agent orchestration," even though until now it was a pure read-only telemetry viewer.

## What's in this PR

**Backend** (`agentmux-srv`)
- New `db_agent_groups` table (schema v24) for saved, named target sets — the Ansible-inventory-group lesson from the spec's research: a durable group, not just ephemeral selection, is what scales.
- `fleet.broadcast` / `fleet.bulk-stop` / `fleet.group.{create,list,update,delete}` RPCs — thin server-side loops over the *existing* single-target primitives (`ReactiveHandler::inject_message`, `agent_io::stop_one_agent_block`, factored out for reuse), never a new transport. Always return per-target `{succeeded, failed}` results, never a single bool.
- `fleet.bulk-stop` supports an optional staged rollout (fixed batch size + `max_fail_percentage` auto-abort) — a simplified take on Ansible's `serial` + `max_fail_percentage`.
- `POST /api/v1/fleet/bulk-stop` REST route for agent-initiated bulk-stop.

**Agent-facing (MCP)**
- `FleetList` (thin alias of `DiscoverAgents`, framed for target discovery), `FleetBroadcast`, `FleetBulkStop`.
- `FleetBroadcast` deliberately loops the *existing signed* single-target `SendMessage` delivery path client-side in `agentmux-mcp`, rather than routing through a new server-side batch RPC — only the calling agent's own process holds its `AGENTMUX_JEKT_KEY`, so per-message signing can only happen there.

**Frontend (Swarm pane)**
- Per-agent-card checkbox selection, a toolbar that appears once ≥1 agent is selected (inline broadcast, destructive bulk-stop behind a `ConfirmModal` showing the exact resolved target list/count, optional staged-rollout toggle), saved-group management (create from selection / apply / delete), and a per-target result panel after every bulk action — never a single aggregate toast, per the spec's finding that silent partial failure is the most commonly-cited fleet-ops UX pitfall.

## Test plan

- [x] `cargo test` — `agentmux-srv` (2482 tests incl. 8 new fleet-specific + full RPC-engine round-trip tests for group CRUD), `agentmux-mcp` (4 tests), `agentmux-launcher`/`agentmux-cef` untouched/unaffected — all green.
- [x] `tsc --noEmit` clean; full `vitest` suite (2783 tests) green.
- [x] Live-verified against a real running `task dev` instance (not just in-memory test stores): `POST /api/v1/fleet/bulk-stop` via curl, and `fleet.group.create` / `fleet.group.list` / `fleet.broadcast` / `fleet.group.delete` via a raw WS RPC client — all returned correct results with real SQLite persistence confirmed across calls.
- [ ] Visual click-through of the Swarm UI (checkboxes/toolbar/modal actually rendering) was **not** performed — I can't screenshot a separate OS window from within this session. Everything below the UI layer is live-verified; the UI layer itself only has `tsc`/`vitest` coverage plus manual code review. Flagging this gap explicitly rather than claiming full verification.


a5af:
@codex review

AgentY-asaf:
Addressed all findings from Codex and reagent (commit 63227328e):

**P0/P1 (Codex + reagent, independently converged on the same root cause)** — `FleetBroadcast` forwarded `targets` (block_id values, per its own advertised contract) directly as `InjectRequest.target_agent`, but `/agentmux/reactive/inject` resolves that field by registered agent *name* only. Every advertised call would have failed "agent not found." The WS-RPC path (`fleet_broadcast_impl`) already resolved this correctly via `get_agent_by_block`; the MCP tool now does the equivalent by reading `/agentmux/discovery`'s `host.addressable` (the same source `FleetList`/`DiscoverAgents` already use) and mapping block_id → agent_id before signing/injecting. Good catch — neither my added tests nor my live verification exercised this specific MCP-side path, only the WS-RPC side.

**P1 (Codex)** — a tight broadcast loop past ~10 targets would deterministically hit `ReactiveHandler`'s global rate limiter (10/sec, hard reset per second). Both the server-side loop and the MCP client-side loop now send in chunks of 10 with a >1s pause between chunks.

**P2 (Codex)** — the staged-rollout failure check computed a truncated integer percentage before comparing (`1/3 = 33.3%` read as exactly `33`, never exceeding a `33%` threshold). Fixed via cross-multiplication (`batch_failures * 100 > max_pct * batch_len`, no rounding), extracted into a standalone `exceeds_fail_threshold()` with direct unit tests for the exact edge cases you'd expect to catch a truncation bug.

**P2 (Codex)** — `fleet.bulk-stop` never wrote to the audit log despite the spec explicitly promising every bulk action shows up in Warden's Audit tab. Added `ReactiveHandler::log_fleet_action_audit` (thin wrapper around the existing private `log_audit`) and call it once per target, success or failure.

**P2 (reagent)** — `aborted_early` was set whenever a batch's failure rate exceeded the threshold, even when that batch was the *last* one and nothing was actually left unattempted. Now only set when targets were genuinely skipped.

**P2 (Codex)** — Swarm's `FleetToolbar` hid the entire toolbar, including the saved-group picker, whenever nothing was selected — so a saved group could never be the first thing a user reached for to *restore* a selection. The toolbar now also shows when saved groups exist even with an empty selection; only the selection-dependent actions stay gated on count > 0.

All existing + new tests pass: `agentmux-srv` (2481, incl. 6 new/updated fleet tests + 5 new pure-function threshold tests), `agentmux-mcp` (4), `tsc --noEmit` clean, `vitest` (2783). The MCP-side block_id resolution fix is type-checked and logically verified but not re-exercised live (would need multiple live agent panes to test broadcast realistically) — flagging that honestly rather than claiming more than I checked.

PR #2689, opened 2026-08-21
fix(sound,ui): tool-call tone default to 0.25; widget-bar hover-then-click no longer closes menu

Agent3-asaf:
## Summary
- **Tool-call tone volume**: `DEFAULT_TOOLTONES_VOLUME` reverted from `1.0` (PR #2668) back to `0.25`, updated consistently across `sound-defaults.ts`, `schema/settings.json`, `settings-template.jsonc`, and the Rust doc comment in `agentmux-srv/src/backend/wconfig/types.rs`.
- **Widget bar hover/click**: the `···More` button and pinned parent widgets (e.g. "Messengers") could open via hover-intent (~90ms after mouseenter) and then close again on the user's very next click, since the click handler couldn't distinguish "already open from hover" from "already open from a deliberate prior click." Fixed by tracking a per-trigger "confirmed by click" flag: the first click after a hover-open now just confirms the open state; a genuine second click closes it. Root cause and design in `docs/specs/SPEC_WIDGET_BAR_HOVER_CLICK_PREMATURE_CLOSE_2026_08_20.md`.

## Test plan
- [x] `tsc --noEmit` passes
- [x] Verified default-value diff is a clean 1.0 → 0.25 revert at every sync point (no stray literals left)
- [ ] Manual: hover More/Messengers until it opens, click once → stays open; click again → closes
- [ ] CI

### DIFF
--- a/.changesets/1787282050-fix-sound-ui-tool-call-tone-default-to-0-25-widget-bar-hover-arha.md
+++ b/.changesets/1787282050-fix-sound-ui-tool-call-tone-default-to-0-25-widget-bar-hover-arha.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(sound,ui): tool-call tone default to 0.25; widget-bar hover-then-click no longer closes menu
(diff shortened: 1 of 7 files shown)


a5af:
@codex review

PR #2690, opened 2026-08-21
docs(status): cross-channel --resume silently starts a blank conversation on a stale registry session_id

AgentX-asaf:
## Summary

Live-reproduced 2026-08-20: opening a named persistent agent from a fresh
AgentMux channel/instance, after it had run continuously elsewhere for
days, silently dropped all conversation history and started a brand-new
empty session — no error surfaced beyond a backend log line.

Root cause traced end-to-end via srv logs and source: the shared,
cross-channel `Registry` record's `session_id` (the one thing a
genuinely different channel reads to decide what to `--resume`) is
written exactly once, by `backfill_session_ids()`
(`agentmux-srv/src/backend/session_backfill.rs`), which is explicitly
idempotent — skips any record that already carries a non-empty
`session_id`, forever, regardless of whether the live session has since
moved on. The local per-channel `db_agent_instances.session_id` does stay
live (`persist_session_id`, `core.rs`), but that write never reaches any
other channel's copy.

Full writeup with exact log timestamps, code citations, and fix
directions: `docs/status/STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md`.

Filed as GitHub issue (linked separately) for tracking; this PR just adds
the investigation doc. Docs-only, no code changes.

### DIFF
--- a/docs/status/STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md
+++ b/docs/status/STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md
@@ -1,0 +1,182 @@
+# Status: Cross-Channel `--resume` Silently Starts a Blank Conversation on a Stale Registry `session_id` (2026-08-20)
+
+**Status: root cause confirmed, precise code evidence, not yet fixed.**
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2692, opened 2026-08-21
docs(retro): dev instance launched from stale branch after workspace sync

Agent3-asaf:
## Summary
- Adds a retro on a workflow mistake from this session: local `main` was fast-forwarded to latest, but `task dev` was then launched from a different (stale feature-branch) checkout, so the build the user smoke-tested didn't reflect the sync.
- Root cause, fix, and follow-up guidance for next time are in the doc itself.

Docs-only change, no code/behavior affected.


a5af:
@codex review

PR #2693, opened 2026-08-21
fix(agent): recover the largest on-disk session before falling back to a blank conversation on a stale --resume

### LINKED ISSUE #2691: Cross-channel agent resume silently starts a blank conversation when the registry session_id is stale
## Summary

Opening a named, persistent agent (e.g. a named agent like "AgentX") from
a fresh AgentMux instance/channel — while that same agent has been running
continuously in another instance/channel for an extended period — can
silently drop all conversation history and start a brand-new, empty
session. No error is surfaced to the user; the only signal is a backend
log line.

Live-reproduced 2026-08-20. Full investigation with exact log timestamps
and code citations:
`docs/status/STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md`
(PR #2690).

## Root cause

Two independent, non-synchronized places track "this agent's session id":

1. **`db_agent_instances.session_id`** (local, per-channel SQLite column) —
   kept live and correct by `persist_session_id`
   (`agentmux-srv/src/backend/blockcontroller/core.rs:139-188`) on every
   genuine session-id capture. Correct for the originating channel, but
   never reaches any *other* channel's own local DB.
2. **The shared, cross-channel `Registry` record's `session_id`** — the
   ONE thing a genuinely different channel reads to decide what to
   `--resume`. The only production write path to this field is
   `backfill_session_ids()`
   (`agentmux-srv/src/backend/session_backfill.rs:69-80`), which is
   explicitly idempotent:

   ```rust
   if rec.data.session_id.as_deref().map_or(false, |s| !s.is_empty()) {
       continue; // already wired
   }
   ```

   That function's own doc comment: *"Idempotent (skips records that
   already carry a non-empty id)."* Once this field gets ANY value, it is
   **never refreshed again**, no matter how long the agent keeps running or
   how many new top-level sessions it goes through.

## Reproduction

- The registry's `session_id` for the affected agent was
  `d019e2e4-7223-4eeb-a2a6-b16b688b9893` — confirmed present since at
  least 2026-08-14 (6 days before the incident), and it wasn't even the
  agent's own top-level conversation: it belongs to a batch of *subagent*
  dispatches (Claude Code subagents share their parent turn's session id).
- A fresh channel (`0.55.18`) opened the agent and attempted
  `claude.cmd --resume d019e2e4-...`. 1.5s later:
  `WARN: stale --resume session id unreachable under the current config
  dir — clearing so the next message starts a fresh conversation`, followed
  by `WARN: stale --resume session id caused this exit — retrying fresh,
  without --resume`.
- srv respawned with no `--resume` flag, producing a brand-new, empty
  session (8 lines, ~14KB) — while the agent's actual live conversation
  (2.8MB, still being actively written in the original instance) was never
  consulted.

## Impact

Any named, persistent agent kept running continuously in one channel for
an extended period is at risk the next time it's opened from a genuinely
different channel — with no user-facing warning when it happens, just a
pane that quietly "forgets everything" after the first message.

## Suggested fix directions

1. Make the shared registry `session_id` a live pointer instead of a
   write-once value — either propagate `persist_session_id`'s writes
   cross-channel, or stop trusting the stored value as authoritative and
   re-resolve it (e.g. re-run the "largest session file" scan) at the
   moment a `--resume` is actually attempted.
2. At minimum, when a `--resume <sid>` is confirmed unreachable, re-scan
   for the largest on-disk session before falling back to a blank one —
   the correct session was sitting on disk the whole time.
3. Surface the failure to the user (e.g. a system message: "Could not
   resume the previous session — starting fresh") instead of silently
   starting empty.
4. Separately worth its own repro: how a *subagent's* session id ended up
   written into the registry as if it were the parent agent's own
   top-level session in the first place.

AgentX-asaf:
## Summary

Fixes #2691.

When a `--resume <sid>` attempt is confirmed unreachable, the retry
unconditionally cleared `session_id` and started completely blank — even
when the agent's actual, much larger conversation was sitting right there
on disk the whole time. Reproduced case (see
`docs/status/STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md`):
a stale shared-registry `session_id` pointed at a long-dead subagent
session from 6 days earlier, while the real 2.8MB conversation went
unconsulted, and the pane silently started a brand-new empty session with
no error surfaced to the user.

## Fix

- `session_backfill::find_largest_session_for_working_dir` — reuses the
  same "largest session wins" recovery logic `backfill_session_ids`
  already trusts for the analogous cross-channel-open case, scoped to a
  single `CLAUDE_CONFIG_DIR`/`working_dir` pair.
- Wired into `retry_after_resume_failure` via a new
  `find_recovery_session_id` helper: instead of unconditionally clearing
  `session_id` to empty, it now looks for a real recoverable session first.
- Guards against an infinite retry loop: refuses to re-recover a candidate
  that's already the confirmed-`resume_poisoned` id for this attempt — disk
  state can't change between attempts, so an unguarded retry would
  rediscover the identical dead id forever and loop.
- **Additive only**: falls back to the exact pre-fix behavior (blank
  conversation) when `CLAUDE_CONFIG_DIR` isn't set on the spawn config or
  no session exists on disk — no behavior change for the no-signal case.

## Testing

- 7 new tests: 3 for `find_largest_session_for_working_dir`, 3 for
  `find_recovery_session_id` (recovers correctly / no config dir / refuses
  an already-poisoned candidate), 1 end-to-end through
  `retry_after_resume_failure` confirming `inner.session_id` hydrates to
  the recovered session rather than staying blank.
- All 10 pre-existing `retry_after_resume_failure_*` tests pass unchanged
  (confirms no regression to the existing stale-resume state machine).
- Full suite: `cargo test -p agentmux-srv` — 2577 passed, 0 failed.

## Not in scope for this PR

- Making the shared registry `session_id` a live pointer instead of
  write-once (the other fix direction from the status doc) — this PR is
  the safety-net recovery path; the write-once staleness itself is a
  separate, larger change.
- Surfacing the recovery/failure to the user in the UI.
- How a subagent's session id ended up written into the registry as the
  parent's own in the first place (noted as a separate follow-up in the
  status doc).

### DIFF
--- a/.changesets/1787295688-fix-agent-recover-the-largest-on-disk-session-before-falling-q26s.md
+++ b/.changesets/1787295688-fix-agent-recover-the-largest-on-disk-session-before-falling-q26s.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): recover the largest on-disk session before falling back to a blank conversation on a stale --resume
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -1527,6 +1527,33 @@ impl PersistentSubprocessController {
         }
     }
 
+    /// Publish a session-outcome line immediately, via the same append
+    /// path `session_outcome_line`'s other call sites use
+    /// (`persistent_resume::ResumeEffect::EmitSessionOutcome`'s normal
+    /// handling). Used by `retry_after_resume_failure` for the ONE case
+    /// it can decide unambiguously and immediately: no recovery candidate
+    /// exists, so this is genuinely, unconditionally a fresh conversation
+    /// — see that function's own doc comment for why the Resumed case is
+    /// deliberately NOT emitted here.
+    fn emit_session_outcome_now(
+        &self,
+        outcome: persistent_resume::SessionOutcome,
+        attempted_sid: String,
+        actual_sid: Option<String>,
+    ) {
+        let Some(ref broker) = self.broker else { return };
+        let line = session_outcome_line(outcome, attempted_sid, actual_sid);
+        let global_output_zone = super::shell::resolve_global_output_zone(&self.wstore, &self.block_id);
+        super::shell::handle_append_block_file(
+            broker,
+            &self.block_id,
+            PERSISTENT_OUTPUT_SUBJECT,
+            line.as_bytes(),
+            self.filestore.as_ref(),
+            global_output_zone.as_deref(),
+        );
+    }
+
     /// After a confirmed-stale `--resume` failure, try to recover a REAL
     /// session instead of giving up and starting blank
     /// (`docs/status/STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md`).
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentX-asaf:
Addressed all three findings in dc943aa6a:

**Codex P2 — `~` not expanded in `CLAUDE_CONFIG_DIR`**: fixed — `find_largest_session_for_working_dir` now expands via `expand_home_dir_safe`, same as the actual spawn path (`core::apply_working_dir`). New test: `find_largest_session_for_working_dir_expands_a_tilde_config_dir`.

**reagent P1 + Codex P1 (same root cause)** — premature `outcome: Fresh` emission, and no retry-tracking if the recovered id is also rejected: fixed together. `persistent_resume.rs` no longer pre-decides the outcome at the `ConfirmedRetry + ProcessExited` transition — it just threads `attempted_sid` through `FireRetry`. `retry_after_resume_failure` now:
- Emits `Fresh` immediately only when no recovery candidate exists (genuinely known right then).
- When a candidate IS found, emits nothing yet and spawns the recovery attempt via the SAME resume-tracking path a first-time `--resume` uses (`Some(first.clone())`, not `None`) — reusing the already-correct, already-tested `SessionCaptured` handling to emit `Resumed` once genuinely confirmed, or cascade into another `ConfirmedRetry` if the recovered id is ALSO stale. The existing `find_recovery_session_id` poison guard then correctly refuses to re-recover the same dead id on that second pass, falling through to a true blank conversation with `Fresh` correctly emitted at that point.

Two new regression tests lock in the emission-timing behavior directly (`retry_after_resume_failure_emits_fresh_outcome_immediately_when_no_recovery_found`, `retry_after_resume_failure_does_not_emit_an_outcome_yet_when_recovery_is_found`), plus 3 existing `persistent_resume.rs` tests updated (their asserted effects no longer include the removed premature emission — that removal is itself the regression coverage for reagent's finding). Full suite: 2580 passed, 0 failed.

a5af:
@codex review

PR #2694, opened 2026-08-21
[Manoz@Area54] fix(jekt): remove unsafe global-settings agent-id fallback, audit registration events

GenericAgentX-asaf:
## Summary

A PTY-shell pane spawned without its own block-scoped `AGENTMUX_AGENT_ID` fell back to a **global settings value** (`cmd_env.AGENTMUX_AGENT_ID`), shared by every pane in the instance/channel, for jekt auto-registration. `register_agent_with_nonce` unconditionally evicts whoever previously held an `agent_key`, so two panes across instances/channels sharing a `settings.json` would collide — the second to register silently steals the first's jekt identity, and future messages addressed to that agent go to whichever pane registered last. This matches a live reported symptom: jekts addressed to one agent being delivered to a different agent, both running in separate instances on the same host.

The parallel `persistent.rs` path (used for Claude-based agents, `muxbus_agent_id_from_env`) never had this fallback — it already reads identity strictly from the block's own spawn env. This PR brings the PTY-shell path in line with that already-safe pattern.

- Removed the global-settings fallback from `agent_id_for_jekt` resolution in `shell/lifecycle.rs`, extracted into a testable `resolve_agent_id_for_jekt` free function. A pane with no block-scoped identity is now simply not jekt-registered (matches `persistent.rs`).
- Extended `AuditLogEntry` with `event_kind` ("delivery" | "register" | "unregister") and `evicted_block`/`evicted_agent`, and added `Handler::log_audit_registration`, called from `register_agent_with_nonce`/`unregister_agent`/`unregister_block`. Previously registration/eviction events left zero audit trail — only delivery attempts were logged — so an incident like this couldn't be reconstructed after the fact. `GET /agentmux/reactive/audit` now surfaces this with no other changes needed.
- Added regression tests locking in both behaviors (see Test plan).

This is Phase 1 of a 3-phase plan; Phase 2 (recipient-side identity verification at delivery time) and Phase 3 (Stash UI registration indicator) are filed as follow-up issues referencing this PR.

## Test plan

- [x] New test: `resolve_agent_id_for_jekt` never falls back to global settings, block metadata takes priority over legacy `WAVEMUX_AGENT_ID`, legacy env still works when no block-scoped ID is set (`shell/lifecycle.rs`'s `agent_id_for_jekt_tests`)
- [x] New tests: `register_agent`/`unregister_agent`/`unregister_block` now audit `register`/`unregister` events with correct `evicted_block`/`evicted_agent` (`backend/reactive/tests.rs`)
- [x] Fixed 4 pre-existing tests whose exact audit-log-length assertions broke now that `register_agent`'s setup call also audits an event — filtered to `event_kind == "delivery"`, matching each test's actual intent
- [x] `cargo test -p agentmux-srv --bin agentmux-srv reactive::` — 127 passed, 0 failed
- [x] `cargo test -p agentmux-srv --bin agentmux-srv agent_id_for_jekt` — 1 passed, 0 failed

### DIFF
--- a/.changesets/1787297005-fix-jekt-remove-unsafe-global-settings-agent-id-fallback-aud-aogv.md
+++ b/.changesets/1787297005-fix-jekt-remove-unsafe-global-settings-agent-id-fallback-aud-aogv.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(jekt): remove unsafe global-settings agent-id fallback, audit registration events
--- a/agentmux-srv/src/backend/blockcontroller/shell/lifecycle.rs
+++ b/agentmux-srv/src/backend/blockcontroller/shell/lifecycle.rs
@@ -32,25 +32,58 @@ use crate::backend::shellexec::ShellProc;
 use crate::backend::wps;
 
 /// Resolve the effective AGENTMUX_AGENT_ID for jekt auto-registration from a
-/// block's own spawn metadata (or the legacy WAVEMUX_AGENT_ID env var).
-/// Priority: block metadata > WAVEMUX_AGENT_ID env compat. Deliberately does
-/// NOT fall back to the global settings' cmd_env.AGENTMUX_AGENT_ID: that
-/// value is shared across every pane in the instance/channel, so a pane
-/// spawned without its own explicit per-block ID would silently inherit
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

Issue #2696, opened 2026-08-21
Stash: per-agent jekt registration status indicator


GenericAgentX-asaf:
## Context

Follow-up to #2694 (Phase 1) and #2695 (Phase 2). User-requested: after a live incident where jekts addressed to one agent were delivered to a different agent on the same host, there's currently no way to *see* an agent's live jekt registration state from the UI at all -- no indication of which block/channel/instance currently owns an agent's identity, whether it's also (unexpectedly) registered elsewhere, or whether a recent delivery hit an identity mismatch.

## Where this belongs

Per-agent live state belongs in **Stash** (`AgentStashModal.tsx`, opened via the backpack icon in the agent-pane header), not the global Armory pane -- Armory is explicitly scoped to shared/reusable resources, not per-agent-instance live state (see CLAUDE.md's "Not widgets" table). `AgentIdentityLinksPanel` (`frontend/app/view/identity/agent-identity-links-panel.tsx`) is the closest existing precedent: fetch via an `RpcApi.*Command`, live-update via a `waveEventSubscribe` WPS event (not polling), and a small `<span class="...-status-dot is-{badge().dot}">` status badge per row.

## Proposal (not yet designed in detail)

**Backend:** new endpoint `GET /agentmux/reactive/agent/registrations?id=`, merging:
- this instance's local `AgentRegistration` (already exposed via `GET /agentmux/reactive/agent`)
- cross-instance/channel entries from the host-global file registry (`registry::lookup_all_shared`, `AgentEntry` -- currently NOT exposed to the frontend at all)
- the most recent `identity-mismatch` audit entry for this agent_id, if any (from #2695's Phase 2 work)

**Frontend:**
- new `frontend/app/store/rpc-api/reactive.ts` RPC wrapper (currently zero coverage of any `/agentmux/reactive/*` endpoint from the frontend)
- new "Registration" tab in `AgentStashModal.tsx`, following `AgentIdentityLinksPanel`'s fetch/subscribe/badge pattern
- badge states: registered-here (green) / registered-elsewhere-too (amber -- the actual risk signal for this bug class) / recent-mismatch-detected (red, once #2695 lands) / not-registered (gray)

## Related

- #2694 -- Phase 1: global-settings fallback fix + audit trail (prerequisite, merged/in-review)
- #2695 -- Phase 2: recipient identity verification at delivery time (this issue's `recent-mismatch-detected` badge state depends on it)

Issue #2695, opened 2026-08-21
jekt: recipient identity verification at delivery time


GenericAgentX-asaf:
## Context

Follow-up to #2694 (Phase 1 of 3), which removed a global-settings agent-id fallback that could cause a same-host cross-instance jekt to be delivered to the wrong agent, and added an audit trail for registration/eviction events.

Phase 1 closes the specific collision that caused it, but the underlying architectural gap remains: `ReactiveHandler::inject_message_inner`'s delivery path (`agentmux-srv/src/backend/reactive/handler.rs`, ~line 356 lookup into `agent_to_block`) blindly trusts whatever the registry maps `target_agent` to. There is no point where the receiving block itself confirms "yes, I am still agent X" before a message is delivered to it — registration is register-once-then-trusted-until-explicit-unregister, with no live re-check. `AgentRegistration.last_seen`/`Handler::update_last_seen` exist but are dead code (never called anywhere in the crate).

## Proposal (not yet designed in detail)

Add a verification hop in `inject_message_inner`, after resolving `block_id` and before writing to the PTY/controller: compare `req.target_agent` against the receiving block's own immutable, spawn-time-captured identity (a value the `BlockController` itself holds, independent of `ReactiveHandler`'s maps — e.g. via a new `BlockController::confirm_agent_identity(&self, expected: &str) -> bool`). On mismatch: reject delivery (distinct error from "agent not found"), and log via `log_audit`/`log_audit_registration` with a new `outcome: "identity-mismatch"` so it's visible in `GET /agentmux/reactive/audit` and (once #<phase-3-issue> lands) in the Stash UI.

This check only needs to run on the final local-delivery hop (whichever instance actually holds the PTY/controller) — Tier 2a/2b cross-instance forwarding already funnels through this same function on the target instance after the HTTP hop, so no separate check is needed mid-forward.

**Open question, not yet decided:** whether to also revive `update_last_seen` as a periodic heartbeat. Current lean is no — it would catch staleness/liveness, a different failure mode than misdirected identity, and the delivery-time check above is a cheaper, more direct fix for the actual failure surface. Worth a second opinion before ruling it out for good.

## Related

- #2694 — Phase 1 (this issue's prerequisite): global-settings fallback fix + audit trail
- #2595 — same-host cross-channel jekt trust/signing gap (a different concern: sender authentication, not receiver correctness)

PR #2697, opened 2026-08-21
[Manoz@Area54] feat(jekt): verify recipient identity before delivery

GenericAgentX-asaf:
## Summary

Phase 2 of the jekt misdelivery hardening (#2695), stacked on #2694 (Phase 1 — merge that first).

Phase 1 closed the specific collision that caused a live-reported misdelivery incident (jekts addressed to one agent landing on a different agent, same host, separate instances), but the underlying architectural gap remained: `inject_message_inner` resolved a delivery target purely from `Handler`'s own `agent_to_block` map and trusted it blindly. There was no point where the receiving block confirmed its own identity before a message was delivered to it — so a stale or drifted registry entry (e.g. surviving a respawn, or a resync overwriting registration without updating what the block itself believes it is) could still silently misdeliver, with nothing to catch it.

This adds a genuine second, independent source of truth:

- Each block controller now captures its own jekt/muxbus identity once at spawn time — `ShellController` via `resolve_agent_id_for_jekt` (added in #2694), `PersistentSubprocessController` via `muxbus_agent_id_from_env` (pre-existing) — and retains it on the controller struct itself, not just in `Handler`'s maps.
- New `Controller::agent_id() -> Option<String>` trait method (default `None`) exposes it.
- New `Handler::agent_identity_confirmer` hook, mirroring the existing `input_sender`/`message_sender` pattern, wired in `bootstrap.rs` to query the real controller via `blockcontroller::get_controller(block_id)`.
- `inject_message_inner` now calls the confirmer right after resolving `block_id`: an **active mismatch** (confirmer returns `Some(actual)` that differs from `target_agent`) rejects delivery with a distinct error and audits an `"identity-mismatch"` outcome. A confirmer that's unset, or returns `None` (unverifiable — e.g. a controller type that doesn't implement `agent_id()`), leaves delivery unaffected — matching this codebase's existing jekt trust philosophy elsewhere (`TRUST=`/`ESCALATE=`): only an active red flag is acted on, mere absence of proof is not.

Checking `agent_to_block` against itself would have been a tautology and caught nothing — the whole point is that the controller's own captured value is written independently, at a different time, by different code, from the registry's own bookkeeping.

## Test plan

- [x] New test: an active mismatch (confirmer returns a different agent_id) rejects delivery, never reaches the input sender, and is audited with `outcome: "identity-mismatch"`
- [x] New test: confirmer returning `None` does not block delivery (fail-open, no regression)
- [x] New test: confirmer match is case-insensitive (consistent with `agent_to_block`'s own lowercased-key lookup)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv reactive::` — 130 passed, 0 failed
- [x] `cargo build -p agentmux-srv --bin agentmux-srv` — full workspace build clean (pre-existing warnings only)

Follow-up: #2696 (Phase 3 — Stash UI indicator, including a badge for `identity-mismatch` audit entries this PR now produces).

### DIFF
--- a/.changesets/1787299190-feat-jekt-verify-recipient-identity-before-delivery-rcwt.md
+++ b/.changesets/1787299190-feat-jekt-verify-recipient-identity-before-delivery-rcwt.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(jekt): verify recipient identity before delivery
--- a/agentmux-srv/src/backend/blockcontroller/mod.rs
+++ b/agentmux-srv/src/backend/blockcontroller/mod.rs
@@ -235,6 +235,20 @@ pub trait Controller: Send + Sync {
         None
     }
 
+    /// Refresh this block's own captured jekt/muxbus identity (see
+    /// [`agent_id`](Controller::agent_id)'s doc comment). Called whenever
+    /// `ReactiveHandler::register_agent`/`register_agent_with_nonce`
+    /// (re-)registers THIS block's block_id under a (possibly different)
+    /// agent_id, so the two independently-written copies never drift apart
+    /// (reagentx P1 on #2697: `agent_id()` was captured once at spawn and
+    /// never refreshed, while `agent_to_block` gets re-keyed on every
+    /// `register_agent` call — e.g. `handle_reactive_register`'s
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

PR #2699, opened 2026-08-21
fix(agent): expand ~ in working_dir too when recovering a stale-resume session

AgentX-asaf:
## Summary

Follow-up to #2693 (already merged) — Codex found this in a re-review
pass that landed a few minutes after the merge went through, against the
fix commit itself (`dc943aa6a`).

`find_largest_session_for_working_dir` expanded `~` in `CLAUDE_CONFIG_DIR`
(the config-dir half, fixed pre-merge per Codex's earlier P2) but never
did the same for `working_dir`. This half matters more:
`agent_open.rs`'s default cwd for any agent with no explicit
`working_directory` set is the literal string `~/.agentmux/agents/<slug>`
— not an edge case, the single most common configuration. Without this
fix, recovery was hashing the wrong (literal-tilde) project slug and
silently failing for nearly every default-configured agent — including,
plausibly, the exact "AgentX" incident #2693 was written to fix in the
first place.

## Fix

Expand `working_dir` via `core.rs`'s own `expand_home_dir` before encoding
the project slug — a different function from `base::expand_home_dir_safe`
(used for `config_dir`), since `apply_working_dir` (the actual spawn path)
uses each for a different input; this now mirrors both exactly.

## Testing

New test `find_largest_session_for_working_dir_expands_a_tilde_working_dir`,
modeled on `agent_open.rs`'s actual default-cwd shape. Verified it's a
real discriminator: reverted the one-line fix, confirmed the test panics,
restored it. Full suite: 2581 passed, 0 failed.

### DIFF
--- a/.changesets/1787323193-fix-agent-expand-in-working-dir-too-when-recovering-a-stale--e7bz.md
+++ b/.changesets/1787323193-fix-agent-expand-in-working-dir-too-when-recovering-a-stale--e7bz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): expand ~ in working_dir too when recovering a stale-resume session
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

a5af:
@codex review

PR #2701, opened 2026-08-21
fix(agent-pane): reuse fresh output.idx cache instead of full-history rescan on every pane open

Agent1-asaf:
## Summary

Opening a pane for an agent with a lot of accumulated history was slow
on every single reopen — traced to `global_zone_line_count`
(`agentmux-srv/src/server/app_api/mod.rs`), which is hit first on
mount (`useHistoryPagination.ts`'s `needsHwmWidening` check, true on
almost every reopen since #2323 made pane reopen always spawn a new
block).

That function unconditionally called `rebuild_output_idx` — a full
streaming byte-scan of the agent's *entire* transcript — instead of
first checking whether the `output.idx` cache already sitting on disk
is fresh for the current `output` size. Its sibling handler,
`register_blockfile_read_range` (`blockfile.rs`), already does this
freshness check (`covered_size` header comparison) before rebuilding;
`global_zone_line_count` was added later by #1399 (globalize agent
transcript for cross-channel history) and simply never reused it.

Net effect: every pane open for an agent with any global-zone history
paid an O(total transcript size) rescan, even when nothing had
changed since the index was last built.

## Fix

Port the same freshness check `read_range` already uses into
`global_zone_line_count`: if `output.idx`'s 8-byte `covered_size`
header matches the current `output` file size, trust the cached line
count directly instead of calling `rebuild_output_idx`.

## Testing

- New test `global_zone_line_count_reuses_fresh_index_instead_of_rescanning`:
  seeds a deliberately-wrong-but-size-matching index (2 fabricated
  entries) alongside real `output` content that would scan to 3
  non-blank lines — proves the fix trusts the cached 2 rather than
  rescanning to 3.
- Existing `global_zone_line_count_counts_non_blank_lines` still passes
  unchanged (no-cache / empty / absent-zone cases).
- Full suite: `cargo test -p agentmux-srv -- --test-threads=1` — 2581
  passed, 0 failed. (Default parallel run shows one pre-existing flaky
  failure, `sysinfo::handle_anomaly_tests::refresh_processes_specifics_does_not_leak_a_handle_per_call`
  — confirmed unrelated: it also fails under parallel scheduling on a
  clean `main` checkout with no changes, and passes in isolation and
  single-threaded regardless of this diff.)

### DIFF
--- a/.changesets/1787324251-fix-agent-pane-reuse-fresh-output-idx-cache-instead-of-resca-65hl.md
+++ b/.changesets/1787324251-fix-agent-pane-reuse-fresh-output-idx-cache-instead-of-resca-65hl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): reuse fresh output.idx cache instead of rescanning full history on every pane open
--- a/agentmux-srv/src/backend/agent_session/archive.rs
+++ b/agentmux-srv/src/backend/agent_session/archive.rs
@@ -245,6 +245,13 @@ pub fn clear_local_current_zone(filestore: &FileStore, zone: &str) {
 /// a missing file is the expected "agent never mirrored" case (silent), other
 /// errors are logged but never propagated. Keeps the global zone in lockstep
 /// with the per-channel `:current` clear in [`archive_session`].
+///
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2702, opened 2026-08-21
muxspect verify-sender: fast JEKT sender verification

GenericAgentX-asaf:
## Summary

- Verifying a JEKT's claimed `FROM` currently means manual filesystem/process archaeology — no single command answers "is this a real, currently-live agent, and via what tier." See `docs/retro/RETRO_JEKT_CROSS_CHANNEL_TRUST_SELF_DECLARED_2026_08_21.md` for the incident that motivated this (a live-verification exchange with AgentA on `agenta-07017/agentmux-wt-muxspect-dock`'s `SPEC_BACKGROUND_TASK_DASHBOARD_INTELLIGENCE_2026_08_20.md` took ~9 turns to resolve, entirely because of this gap).
- Adds `GET /api/v1/muxspect/verify-sender?name=X`, composing the same four sources `handle_discovery` already aggregates (host, cross-channel, lan, wan) into a `found`/`not_found`/`stale` verdict with a **computed** `trust` value per tier — independent of whatever a JEKT's own `TRUST` field says, which today falls back to an undocumented `self-declared` placeholder for the cross-channel case.
- Adds `muxspect verify-sender <agent_name>` to the CLI. Checks a zero-round-trip "spawner" tier first: a task-dev instance's own `AGENTMUX_CHANNEL`/`AGENTMUX_RUNTIME_MODE` already encode which agent's dev build spawned it — a stronger, cheaper signal than any network call, and the actual signal that resolved the retro's incident once checked.
- Full design in `docs/specs/SPEC_MUXSPECT_VERIFY_SENDER_2026_08_21.md`.

## Testing

- 10 new Rust unit tests (`classify_sender` tier/trust/staleness/case-insensitivity/priority-ordering cases, plus two handler tests) — `cargo test -p agentmux-srv --bin agentmux-srv -- muxspect_handlers`: 32 passed, 0 failed.
- 9 new JS tests (`checkSpawnerTier` cases + a `verify-sender` `parseArgs` case) — `npx vitest run agentmux-srv/src/backend/shellintegration/muxspect.test.mjs`: 19 passed, 0 failed.
- `cargo check -p agentmux-srv --tests` clean (pre-existing warnings only, none in touched files).

## Test plan

- [ ] `muxspect verify-sender <name>` from a task-dev instance whose `AGENTMUX_CHANNEL` names the spawning agent — expect `tier: spawner`, no network call made.
- [ ] `muxspect verify-sender <name>` for a live cross-channel peer — expect `tier: cross-channel`, `trust: cross-channel-verified`.
- [ ] `muxspect verify-sender <name>` for an unknown name — expect `status: not_found`, HTTP 200, non-zero CLI exit code.

Related to the background-task-dashboard-intelligence work on `agenta/spec-foreground-background-abstraction` (`agentmux-wt-muxspect-dock`) — this doesn't touch that spec's own scope, just a tooling gap surfaced while live-verifying it.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.changesets/1787328437-feat-muxspect-verify-sender-8f3q.md
+++ b/.changesets/1787328437-feat-muxspect-verify-sender-8f3q.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxspect): add `verify-sender` — registry-liveness lookup for a JEKT's claimed FROM
--- a/agentmux-srv/src/backend/reactive/registry.rs
+++ b/agentmux-srv/src/backend/reactive/registry.rs
@@ -304,8 +304,11 @@ fn pid_alive(pid: u32) -> bool {
 
 /// Grace period during which a forward failure against a live-process entry
 /// is presumed transient rather than proof the specific agent is gone — see
-/// [`should_evict_on_forward_failure`].
-const FORWARD_FAILURE_GRACE_MS: u64 = 60_000;
+/// [`should_evict_on_forward_failure`]. `pub(crate)` so `muxspect_handlers`'s
+/// `verify-sender` route can reuse the same real, already-tested staleness
+/// judgment for shared-registry entries instead of inventing its own
+/// threshold (see that module's own doc comment for why).
+pub(crate) const FORWARD_FAILURE_GRACE_MS: u64 = 60_000;
 
 /// Should `server/reactive.rs`'s evict-on-forward-failure sites remove this
 /// entry after a failed forward (`success:false` or a connection error)?
--- a/agentmux-srv/src/backend/shellintegration/muxspect.mjs
+++ b/agentmux-srv/src/backend/shellintegration/muxspect.mjs
@@ -43,13 +43,15 @@ Usage:
                                   open — no pane reload needed. muxspect's
                                   only mutating command; see the spec above.
   muxspect verify-sender <agent_name> [--json]
-                                  is <agent_name> a real, currently-live
-                                  agent, and via what tier (spawner / host /
-                                  cross-channel / lan / wan)? For sanity-
-                                  checking a JEKT's claimed FROM before
-                                  acting on it (SPEC_MUXSPECT_VERIFY_SENDER_
-                                  2026_08_21.md). Exits non-zero unless the
-                                  verdict is 'found'.
+                                  is an agent named <agent_name> currently
+                                  REGISTERED, and via what tier (spawner /
+                                  host / cross-channel / lan / wan)?
+                                  Registry-liveness only — NOT cryptographic
+                                  sender verification; does not replace
+                                  checking a JEKT's own TRUST=/SIG= fields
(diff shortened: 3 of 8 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Thanks both — all four issues were real and are fixed in e85b20b3f.

**reagentx-workflow (P1):** `tier_trust()` reused `host-verified`/`network-claimed`, which this repo's actual `CLAUDE.md` already defines to mean cryptographic HMAC-SHA256/Ed25519 sender verification — a real, already-shipped mechanism I was unaware of because I'd sourced the original retro/spec from `agenta-07017/scratch/agentmux-docs` (a scratch copy of a *different* repo) instead of this repo's real docs. Fixed by dropping the `trust` field entirely and reframing the whole command as a registry-liveness lookup, with an explicit "this is not cryptographic verification, check the JEKT's own TRUST=/SIG= fields" note in the CLI output and doc comments.

**reagentx-workflow (P2):** `docs/security/trust-model.md` / `docs/internals/interagent-comms.md` don't exist in this repo — confirmed via `find` + `git log --all`. Corrected both docs; rewrote the retro's root-cause section (turns out `TRUST=self-declared` almost certainly just meant no `AGENTMUX_JEKT_KEY` on file yet — a mundane, already-documented case per the real CLAUDE.md, not an undiscovered protocol gap).

**reagentx-workflow (P2) + changeset:** added `.changesets/1787328437-feat-muxspect-verify-sender-8f3q.md`.

**codex (P1, muxspect.mjs:256, spawner tier):** correct that `AGENTMUX_CHANNEL`/`AGENTMUX_RUNTIME_MODE` are a naming convention, not an attested identity. Added an explicit caveat (both in code comments and CLI output) that this tier is a coordination heuristic, not proof — no `trust`/`verified` field implies otherwise anywhere in the output now. A real fix (launcher-attested `AGENTMUX_SPAWNED_BY`) is out of scope, noted as future work in the spec's Non-goals.

**codex (P1, muxspect_handlers.rs:612, host staleness):** confirmed — no caller of `update_last_seen` anywhere in the codebase, so host-tier `last_seen` is registration-time-only. Fixed: host tier is now never staleness-checked (`ReactiveHandler::list_agents()` is synchronously accurate by design — removed on unregister, not aged out).

**codex (P1, muxspect_handlers.rs:644, LAN units):** confirmed — `LanInstance.last_seen` is `.as_secs()`, my code treated it as ms. Fixed with the missing `× 1000`.

**codex (P2, muxspect_handlers.rs:555, stale-duplicate priority):** fixed — `classify_sender` now prefers a live match over an earlier-tier stale one, falling back to the most-trusted stale match only if every match is stale.

Also reused the real, already-tested `FORWARD_FAILURE_GRACE_MS` (60s, `registry.rs`) for cross-channel/LAN staleness instead of the invented 30s constant. 6 new Rust tests added (35 total) covering the corrected logic; full Rust + JS suites passing, `cargo check` clean.

PR #2704, opened 2026-08-21
docs(analysis): correlate the already-collected tool-call scroll oscillation repro data

AgentX-asaf:
## Summary

Follow-up to `docs/analysis/ANALYSIS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_17.md`.
Its recommended next step ("run a live repro, correlate `wave-scroll-shrink`
against `wave-turn` state") had already happened — a dedicated
`ScrollPinTest` agent ran a real repro on 2026-08-17 on the
`loap-fix-tool-call-scroll-shrink-oscillation` dev branch — but the
collected log data was never pulled or analyzed. This PR does that.

## Findings

Two distinct shrink classes in the 8 collected events, not one:
- **Small (~16-24px)**: confirmed tied to a tool's running→terminal DOM
  swap ("Source A" from the original analysis) — an identical 19px delta
  recurs across two different tool-lifecycle endings (natural completion
  vs. user-interrupt), which is clean confirmation this mechanism is real,
  not just theorized.
- **One large (13,502px)**: a single-frame collapse tied to a
  `Streaming → Idle → Streaming` `ReconcileTurnActive` double-flip 21ms
  earlier — not explained by any of the three originally theorized
  sources, and likely the more user-visible/jarring half of the
  "oscillation" complaint.

Full doc: `docs/analysis/FINDINGS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_21.md`.

Docs-only, no code changes. Recommends the next concrete step: trace what
`ReconcileTurnActive` re-evaluates in `AgentDocumentVirtualList.tsx`/the
agent-document reducer, since that's the highest-value unanswered question
this data raises.

PR #2705, opened 2026-08-21
docs(analysis): correct two overclaims in the scroll-oscillation findings doc

AgentX-asaf:
## Summary

Follow-up to #2704 (merged) — Codex reviewed the merged commit and found
two real overclaims in my analysis:

1. **"Two independent trigger-path confirmations"** — only one of the two
   19px shrinks is actually adjacent to a terminal transition; the other
   fires 2.8s before its own tool's `TurnEnd` while still `Streaming`.
   Also misattributed the component: the 08-17 analysis's own addendum
   already ruled out `ToolBlock.tsx`'s summary row (fixed-height CSS) —
   the one clean example is better explained by that same addendum's
   already-documented `denied`/`canceled`-only FLIP-bypass gap, which the
   one example (an interrupted tool) is consistent with but doesn't
   confirm on its own.
2. **"93% disappears in one frame" / "synchronous DOM mutation"** — the
   diagnostic only samples at `scrollToTrueBottom()` call boundaries
   (~162ms apart here), not per-frame, so this can't be established from
   the data. Same issue for calling `ReconcileTurnActive` causal rather
   than just temporally correlated, from one occurrence.

## Fix

Corrected both claims, downgraded confidence language throughout, and
replaced the recommended-next-step framing with what's actually needed
(frame/mutation-level instrumentation for Class 2; a cheap direct
`heightStale` check for Class 1) rather than implying more of the same
log data would resolve either.

Docs-only, no code changes.

### DIFF
--- a/docs/analysis/FINDINGS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_21.md
+++ b/docs/analysis/FINDINGS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_21.md
@@ -32,14 +32,30 @@ agent "ScrollPinTest-host-7c31"), correlated against the same log's
 23:33:51.921  [wave-scroll-shrink] 1023px -> 1004px (delta=19px)      ← IDENTICAL delta
 ```
 
-The same **19px delta recurs exactly** across two different tool
-lifecycles ending two different ways (one runs to natural completion,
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

a5af:
@codex review

PR #2706, opened 2026-08-21
docs(spec,retro): tool-call bursts restart the Working row's type-out reveal

Agent1-asaf:
## Summary

Formalizes an operator-reported UI bug into a retro + fix spec, no code
changes in this PR.

- **Retro**: root-cause investigation for "tool calls interrupt agent
  thinking dialog" (reproduced live in an agent pane). Traced to
  `AgentWorkingRow`'s type-out reveal effect restarting on every
  intermediate `currentTool`/`toolPromoted` flip during a tool-call
  burst, not to any loss/truncation of the actual thinking-text
  content (that path was checked and ruled out — documented explicitly
  so it isn't re-litigated later).
- **Spec**: proposes batching the per-chunk dispatch loop in
  `useAgentStream.ts` (the layer that actually sees "these N tool
  transitions arrived together") as the fix, with a debounce-the-reveal-effect
  follow-up noted as defense in depth, not required for the initial fix.

## Not in this PR

Implementation of the fix itself — tracked as a follow-up issue.

### DIFF
--- a/docs/retro/retro-agent-working-row-reveal-interrupted-by-tool-burst-2026-08-21.md
+++ b/docs/retro/retro-agent-working-row-reveal-interrupted-by-tool-burst-2026-08-21.md
@@ -1,0 +1,161 @@
+# Retro: Tool-call bursts interrupt the agent-pane "thinking" shimmer/type-out
+
+**Date:** 2026-08-21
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #2719, opened 2026-08-22
test(jekt): regression tests for tier-classification keyword matching and marker rendering

GenericAgentX-asaf:
## Summary

`is_sensitive_message` and `wrap_jekt_message` (`agentmux-srv/src/backend/reactive/sanitize.rs`) are the core of jekt security-tier classification (keyword-based SENSITIVE escalation, and the `TRUST=`/`SIG=`/`ESCALATE=` marker fields), but were previously only exercised transitively through full `Handler::inject_message` integration tests elsewhere in `tests.rs`. This adds direct, isolated fixture coverage for both.

- Keyword-match boundary tests: whole-word positives, substring-false-positive avoidance (`dispatch`/`patch`/`tokenize`/etc. must NOT match), case-insensitivity, string-boundary matches, and the full substring-keyword list.
- Marker-rendering matrix: `TRUST=`/`SIG=`/`ESCALATE=` field combinations across host/LAN/WAN delivery tiers and all verification states, plus a full combinatorial smoke test asserting no combination produces a malformed marker.

100/100 `reactive::tests` pass (87 existing + 13 new) — no behavior changed, this is test-only.

## Findings surfaced (not fixed here — flagging for a separate decision)

Two behaviors were discovered while writing these fixtures and are locked in as explicit, clearly-commented tests documenting *current* behavior, not asserting it's correct:

1. **Plural forms bypass every whole-word keyword.** `contains_whole_word` requires a non-alphanumeric character immediately after the match. Every `SENSITIVE_WHOLE_WORD_KEYWORDS` entry has a common plural (`tokens`, `credentials`, `passwords`, `secrets`, `keychains`) that is NOT currently caught — only the singular form is. A jekt saying "your credentials were leaked" reaches `TIER=coord` today, not `sensitive`.
2. **`"armory"` is a broad false-positive source.** It's a `SENSITIVE_SUBSTRING_KEYWORDS` entry, but also the name of a core AgentMux pane (MCP servers / Skills / ABF bundles hub). Ordinary jekts that merely mention the Armory pane by name get keyword-escalated to `TIER=sensitive`.

Happy to open follow-up issues/PRs for either if the team wants them addressed — wanted eyes on both before changing escalation behavior.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv reactive::tests::` — 100/100 passed

### DIFF
--- a/.changesets/1787386695-test-jekt-add-regression-tests-for-tier-classification-keywo-v21l.md
+++ b/.changesets/1787386695-test-jekt-add-regression-tests-for-tier-classification-keywo-v21l.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+test(jekt): add regression tests for tier-classification keyword matching and marker rendering
--- a/agentmux-srv/src/backend/reactive/tests.rs
+++ b/agentmux-srv/src/backend/reactive/tests.rs
@@ -2212,9 +2212,9 @@ async fn test_handler_inject_wan_trust_is_always_network_claimed_regardless_of_s
 // incidentally through full `Handler::inject_message` calls. Added because a code
 // review of the keyword lists (SENSITIVE_WHOLE_WORD_KEYWORDS /
 // SENSITIVE_SUBSTRING_KEYWORDS, sanitize.rs) found no existing direct unit
-// coverage of the boundary logic itself — see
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

Issue #2718, opened 2026-08-22
Two distinct scroll-height anomalies found in live wave-scroll-shrink data: recurring ~251px shrink, and whole-window collapse to 0px


Agent1-asaf:
## Summary

Pulled full-session `[wave-scroll-shrink]` diagnostic data from two
currently-running local-build instances (0.55.18: 1,434 events / 3
panes over ~8h; 0.55.19: 575 events / 3 panes over ~8h) and found two
patterns not present in the smaller single-pane dataset that originally
shipped this diagnostic (`docs/analysis/FINDINGS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_21.md`).
Full writeup: `docs/analysis/FINDINGS_TOOL_CALL_SCROLL_OSCILLATION_LIVE_INSTANCE_DATA_2026_08_22.md`
(#2717).

## 1. Recurring ~251-252px shrink — likely one specific fixed-height element

54 occurrences in the 0.55.18 sample, magnitude consistently 251 or
252px (vs. a smooth, wide spread for every other magnitude bucket in
the histogram). Hits all 3 open panes independently, 12s-9min apart,
over nearly an hour of sampled time. This consistency strongly suggests
one specific UI element with a fixed/estimated height around
251-252px that periodically collapses — not general
markdown/tool-output content reflow (the existing Class 1/Class 2
taxonomy from the prior analysis).

**Not yet identified which component.** Needs a source search for a
matching fixed/estimated height constant — candidates: `ActivityDock`
row height, a background-task dashboard card, an
`AgentDocumentVirtualList` row-height estimate default.

## 2. Whole-window collapse to 0px — distinct, more severe, unexplained

2 incidents in the 0.55.18 sample where every currently-open pane's
`scrollHeight` drops to exactly 0px simultaneously (within 1-3ms of
each other): 21K-44K px deltas, an order of magnitude past anything in
either dataset otherwise. Confirmed **not** a render-thread freeze —
the host log's own periodic watchdog ticks continued uninterrupted
through both incidents (no logging gap). Zero occurrences in the
0.55.19 sample (575 events, none hit 0px).

**Cause not established.** Multiple independent panes hitting exactly
0px together points at something application-wide (window
resize/minimize/restore, a display-mode change, a CEF repaint/layout
reset) rather than a per-pane document mutation, but this wasn't
chased further — the diagnostic doesn't currently log
window-visibility/resize events alongside scroll height, so there's
nothing to correlate against yet.

[...]

PR #2720, opened 2026-08-22
docs(status): cross-channel opening an old agent can freeze the whole app

Agent1-asaf:
## Summary

Live-investigated an operator-reported freeze: opening AgentY (a
persistent agent, ~1.5M-line transcript, never before opened in this
particular local build instance) locked the entire AgentMux window for
roughly a minute by report; log evidence shows the underlying incident
spans 3m36s end-to-end.

**Root cause (traced, not proven end-to-end):** AgentY's own
`output.idx` was never rebuilt during the incident — PR #2701's
freshness-check fix is working correctly for it. But two *unrelated*
agents' `output.idx` rebuilds (AgentX: ~850MB/1.94M lines, Smike:
~165MB × 3) landed in the same `srv` process during the exact same
window. `rebuild_output_idx` runs synchronously with no
`tokio::task::spawn_blocking` offload, so a large *necessary* rebuild
for any agent can starve the runtime's worker threads and delay
unrelated requests — including AgentY's own, already-fast ones. This
is a materially worse failure mode than PR #2701 addressed: the whole
app can lock up over an unrelated agent's housekeeping.

Also flags a secondary signal worth correlating: a sustained flood of
native `EVENT_OBJECT_LOCATIONCHANGE` window events in the same window
— possibly connected to a separate scroll-height-oscillation
investigation on the same host today (#2718).

Investigation only — no fix implemented, recommended fix directions in
§7 of the doc.

### DIFF
--- a/docs/status/STATUS_CROSS_CHANNEL_AGENT_OPEN_FULL_APP_FREEZE_2026_08_22.md
+++ b/docs/status/STATUS_CROSS_CHANNEL_AGENT_OPEN_FULL_APP_FREEZE_2026_08_22.md
@@ -1,0 +1,196 @@
+# Status: Cross-Channel Opening an Old Agent Locks the Whole App for Minutes, Not Just That Pane
+
+**Status: root cause traced with strong circumstantial evidence, not
+proven end-to-end, not yet fixed.**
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2721, opened 2026-08-22
docs(specs): quick-fork, pane mirroring, and host/LAN/WAN naming specs

AgentX-asaf:
## Summary

Three linked, no-code design specs, written together because they share one underlying naming/addressing substrate:

- **`SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md`** — quick-fork the active agent into a new tab: full independent identity (routing ID + jekt signing key never shared; Armory/credential identity unbound by default, opt-in to inherit), reusing the already-designed `ForkAgentDefinitionCommand` and the already-validated `--resume --fork-session` mechanism from the in-pane fork-bar spec, just targeting a new tab instead of a block-stack row. Trigger is right-click-the-tab ("Duplicate Tab"-style), not a new pane icon, to avoid adding to the pane's already-documented accessory sprawl.

- **`SPEC_AGENT_PANE_CROSS_CHANNEL_LAN_WAN_SYNC_2026_08_21.md`** — mirror a running agent's pane across channels/LAN/WAN. Phased strictly by network scope (same-host → LAN → WAN); floor-control rather than CRDT-merge for the shared composer (backed by research — this is a plain-text field, not a rich document); relay architecture (not P2P WebRTC) for the eventual WAN phase; every phase's input-authorization sits inside the *existing* jekt trust model (DELIVERY/TRUST/ESCALATE) rather than a parallel one.

- **`SPEC_AGENT_NAMING_AND_ADDRESSING_HOST_LAN_WAN_2026_08_22.md`** — the shared naming layer both specs above consume: names are unqualified within a host ("AgentX #4"), qualified with a `HostLabel` only once they cross a boundary (git `origin/branch` style). Verified directly against `agentmuxai/agentmux-cloud` (commit `e0a756e`): the WAN qualifier anchors to the already-shipped `agent-ownership.ts` table rather than a hand-wavy "muxbus account," and the spec explicitly flags that muxbus has **no server-side host-resolution at all** today (WAN delivery is mailbox/poll-based) — the host half of a WAN name must be self-asserted, same trust weight as `TRUST=network-claimed`. Also documents a real synergy for whoever eventually builds jekt's still-missing general WAN signing (issue #2586): the `reagent_sig`/`reagent_key_id` wire format already exists end-to-end for one hardcoded sender; a public key on the existing ownership row could serve as the trust anchor both problems need, instead of two independent identity tables.

## Test plan

- [ ] Docs-only PR — no code, no build/test impact.
- [ ] Cross-links between the three specs (and to prior art: `SPEC_MULTI_SESSION_AGENT_FORK`, `SPEC_AGENT_PANE_FORKS_AND_AUX_PINS`, the jekt trust-layer specs) resolve correctly.
- [ ] `docs/specs/INDEX.md` entries added under the right sections.

### DIFF
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -42,6 +42,8 @@ See also:
 | [SPEC_AGENT_API_FIRST_CLASS_SURFACE_2026_06_17](SPEC_AGENT_API_FIRST_CLASS_SURFACE_2026_06_17.md) | App-API `agent.*` surface (open/fork/define) |
 | [SPEC_APP_API_AGENT_DEFINE_2026_06_06](SPEC_APP_API_AGENT_DEFINE_2026_06_06.md) | `agent.define` RPC |
 | [SPEC_MULTI_SESSION_AGENT_FORK_2026_06_06](SPEC_MULTI_SESSION_AGENT_FORK_2026_06_06.md) | Agent fork / continuation across sessions |
+| [SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21](SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md) | Quick-fork an agent into a new tab (hot clone, full identity) |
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2722, opened 2026-08-22
fix(srv): stop FsWatchPool's health sweep leaking a File+Semaphore handle pair per tick

Agent1-asaf:
## Summary

Found live-debugging an "opening an old agent is slow" complaint on two
long-running local instances: `mem_attribution` flagged 51K-124K open
handles (healthy baseline ~1K). Initially looked like a recurrence of the
already-fixed sysinfo `CreateToolhelp32Snapshot` leak (#2666), but `Section`
handles were flat at 11 — the fixed baseline. `File` and `Semaphore`
handles dominated instead (~98.6% of the total), concentrated identically
(2,142 each) across every agent's memory directory *and* an unrelated
channel config file — a uniform count across unrelated paths pointed at
one shared, path-agnostic, periodic mechanism.

## Root cause

`FsWatchPool::sweep()` re-issues `watch()` on every subscribed target every
`HEALTH_SWEEP_INTERVAL` tick (30s), unconditionally, to self-heal a
silently-dead watch — the code assumed this was a safe no-op for an
already-live watch. Reading the vendored `notify` 7.0.0 Windows backend
directly: `add_watch()` has no check for an existing entry — every call
opens a new `CreateFileW` handle + `CreateSemaphoreW`, then
`HashMap::insert()` silently overwrites the previous `WatchState` with no
cleanup. One redundant watch per target per 30s tick, for the life of the
process.

Full root-cause writeup: `docs/status/STATUS_FS_WATCH_SWEEP_HANDLE_LEAK_2026_08_22.md`.

## Fix

`sweep()` now calls `unwatch()` before every re-`watch()`, so at most one
native watch per target exists at any instant — preserves the silent-death
self-healing this sweep exists for.

## Testing

- New regression test, `sweep_does_not_leak_a_handle_pair_per_call`
  (Windows-only), reusing the existing `process_handle_count` helper from
  the sysinfo leak fix.
- **Proven to discriminate**: temporarily reverted just the fix, reran —
  "handle count grew by 400 over 200 sweep() calls", exactly 2.0/call,
  matching the theorized File+Semaphore pair precisely. Restored the fix,
  reran clean.
- Full suite: `cargo test -p agentmux-srv -- --test-threads=1` — 2639
  passed, 0 failed.

## Not in this PR

This is a code fix only — already-running instances need a restart/rebuild
to pick it up; their already-leaked handles don't self-heal (same caveat
as the sysinfo fix).

### DIFF
--- a/.changesets/1787412644-fix-srv-stop-fswatchpool-s-health-sweep-leaking-a-file-semap-1lf1.md
+++ b/.changesets/1787412644-fix-srv-stop-fswatchpool-s-health-sweep-leaking-a-file-semap-1lf1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): stop FsWatchPool's health sweep leaking a File+Semaphore handle pair per tick
--- a/agentmux-srv/src/backend/fs_watch/pool.rs
+++ b/agentmux-srv/src/backend/fs_watch/pool.rs
@@ -129,9 +129,10 @@ impl FsWatchPool {
         });
 
         // Periodic self-healing sweep — see recovery.rs's HEALTH_SWEEP_INTERVAL
-        // doc comment for why re-issuing watch() on an already-watched path
-        // is a safe, cheap way to detect and recover from a silent death
-        // without needing a true OS-level "is this watch still alive" signal.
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2724, opened 2026-08-22
feat(srv): log rebuild_output_idx's start time and duration

Agent1-asaf:
## Summary

Small, standalone diagnostic improvement split out of the live-debugging
session for `docs/status/STATUS_CROSS_CHANNEL_AGENT_OPEN_FULL_APP_FREEZE_2026_08_22.md`.
The prior completion-only log recorded that a rebuild happened but not how
long it ran or when it started — so correlating a slow RPC against a
concurrent rebuild elsewhere in the process required guessing from
completion timestamps alone (and got that guess wrong once already, in
that doc's own §4 correction).

Kept as a lasting diagnostic (not removed after the debug session) — the
same reasoning `mem_attribution` is kept for.

No behavior change, log-only.


a5af:
@codex review

PR #2725, opened 2026-08-22
fix(agent): fork actually carries conversation history + flat lineage naming

AgentX-asaf:
## Summary

Implements `SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md` Phase 1 — two confirmed bugs in the already-shipped fork flow (`AgentPicker.tsx`'s fork prompt → `ForkAgentDefinitionCommand`), found while writing that spec and verified against real code during PR #2721's review round with reagent and Codex.

**Bug 1 — forking doesn't carry conversation history.** `handleFork` mints a forked `AgentDefinition` (config/instructions/skills clone via `ForkAgentDefinitionCommand`) but never set `continueSessionId`/`forkSession` on the `launchAgentDefinition` call — so every "fork" silently started a brand-new conversation, despite `--fork-session` being empirically validated as working back in June (`SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15` §6.4). Fixed by wiring both through, mirroring how `handleReattach` already does it for plain resume a few lines above.

**Bug 2 — fork-of-a-fork naming.** `template.rs`'s auto-naming ("Name #N") counted only a source's immediate children (`parent_id == source_id`), so forking a fork produced "AgentX #2 #2" instead of the flat "AgentX #3" the naming spec (`SPEC_AGENT_NAMING_AND_ADDRESSING_HOST_LAN_WAN_2026_08_22.md` §4.5) requires. This needed care: `parent_id` is overloaded in this schema — `agentdefcreatefromtemplate` also sets it (to the template's id) for every freshly-instantiated agent, so a naive "walk `parent_id` to the root" fix would wrongly merge unrelated agents created from the same template into one lineage. Fixed by walking only while the current node is itself a fork (non-empty `branch_label` — the one field that actually distinguishes a real fork from a template-instantiation) to find the true lineage root, then counting *and naming* off that root. Naming off the root mattered too — a correctly-*counted* fork-of-a-fork still produced "AgentX #2 #3" when the name was built from the immediate parent's own already-suffixed name instead of the root's.

## Test plan

- [x] New frontend test: `AgentPicker.test.tsx` — `handleFork` sets `continueSessionId`/`forkSession: true` (20/20 tests pass, `npx vitest run`).
- [x] New backend test: `template.rs` — forking a fork produces a flat, lineage-wide name via both the real fork RPC and the suggest-preview RPC (`cargo test -p agentmux-srv --bin agentmux-srv agent_handlers::template::`, 3/3 pass).
- [x] Compiles clean, no new warnings.

### DIFF
--- a/.changesets/1787414854-fix-agent-fork-carries-conversation-history-flat-lineage-nam-cinn.md
+++ b/.changesets/1787414854-fix-agent-fork-carries-conversation-history-flat-lineage-nam-cinn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): fork carries conversation history + flat lineage naming
--- a/agentmux-srv/src/server/agent_handlers/template.rs
+++ b/agentmux-srv/src/server/agent_handlers/template.rs
@@ -26,6 +26,64 @@ use crate::backend::storage::{AgentDefinition, AgentContent, AgentSkill};
 
 use super::super::AppState;
 
+/// Resolve a definition's **fork**-lineage root by walking `parent_id`.
+///
+/// `parent_id` is overloaded in this schema: `agentdefcreatefromtemplate`
+/// sets it to the *template's* id for every freshly-instantiated user
+/// agent (see that handler below), not just `forkagentdefinition`'s actual
+/// forks — so two unrelated agents both created fresh from the same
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2726, opened 2026-08-22
fix(agent-pane): refresh Activity Dock subagent rows on subagent:abandoned

AgentA-asaf:
## Summary

Found via live testing (fresh `task dev` on main): after restarting the app, an agent pane's Activity Dock showed subagent rows stuck on "running" for subagents that were actually long dead — fixed timestamp, static event count, not a live process.

Root cause: `frontend/app/view/agent/activity/subagent-source.ts` (the module-level singleton feeding the dock's subagent rows) never subscribed to `subagent:abandoned`. The backend's own reconciliation (`reconcile_stale_subagents`, runs on every pane reopen with a persisted session id) already correctly downgrades a stale `Active` subagent to `Abandoned` and broadcasts the event — this module just never listened for it, so the dock kept showing the pre-restart snapshot until an unrelated subagent happened to spawn/complete on that pane.

The sibling `dispatch-source.ts` singleton already got this exact fix per reagent/codex's review on PR #2676 — this module predates that fix and was never brought in line with it. One-line addition, mirroring `dispatch-source.ts` exactly.

## Test plan

- [x] `subagent-source.test.ts` (new) — fires a synthetic `subagent:abandoned` event, asserts `ListActive` is re-fetched and the subagent's status updates to `abandoned`; also locks in the pre-existing spawned/completed refresh behavior.
- [x] Full frontend suite green (2852 passed).
- [x] `tsc --noEmit` clean.
- [x] Live-verified the underlying bug via CDP against a fresh `task dev` instance before writing the fix (3 real stale subagent rows observed with frozen timestamps/event counts).

### DIFF
--- a/.changesets/1787416051-fix-agent-pane-refresh-activity-dock-subagent-rows-on-abando-ves3.md
+++ b/.changesets/1787416051-fix-agent-pane-refresh-activity-dock-subagent-rows-on-abando-ves3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): refresh Activity Dock subagent rows on abandon so restarts don't show stale 'running' entries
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2727, opened 2026-08-22
feat(tabs): quick-fork a tab into a new tab with full independent identity

AgentX-asaf:
## Summary

Implements `SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md` Phase 2, on top of Phase 1's fork history-carryover fixes (#2725).

Right-click a tab → **"⑂ Quick-fork to new tab"** → forks the active agent's definition, creates a new tab, creates and places a fresh agent block directly into it, then launches the fork into that block carrying conversation history forward, with an unbound identity by default (spec §5 — a "quick" one-click action shouldn't silently fan out credential access to a second agent).

**Two real things surfaced while building this, verified against actual code rather than guessed:**

1. `launchAgentDefinition` reuses the **source tab's own already-mounted `AgentViewModel`** (via `getBlockComponentModel` — the same block-component registry keybindings/zoom/focus routing already use to reach a live view model from outside its component tree) rather than constructing a new one for the not-yet-rendered destination block. Verified this is safe: the function's only two references to instance state are `this.blockId` (already overridden by the existing `targetBlockId` param) and `this.nodejsError` (a minor, non-blocking rough edge — a fork whose provider isn't installed surfaces that error on the source pane, not the new one).

2. `launchAgentDefinition` unconditionally passed `atoms.staticTabId()` (fixed once at window bootstrap) as `ControllerResyncCommand`'s `tabid`. Harmless for the *existing* same-tab `targetBlockId` use case, but silently wrong for a genuinely different destination tab — the backend stores this value as the spawned controller's own registry `tab_id` and `AGENTMUX_TABID` env var, not just for logging (confirmed in `agentmux-srv/src/backend/blockcontroller/{persistent.rs,shell/lifecycle.rs}`). Added a `targetTabId` override param, defaulting to the old behavior for every existing caller.

The new block is created+placed via `pane.open` with an explicit `tab_id` (`open_pane`'s "explicit tab_id wins" resolution path, `agentmux-srv/src/server/app_api/mod.rs`) — this does create+place in one RPC, so none of `waitForLayoutModel`/`createBlockOnModel`'s extra machinery was needed; that's for the in-pane fork-bar's different same-tab block-stack case.

## Test plan

- [x] 13 new unit tests for `resolveActiveAgentForTab` + `quickForkTabToNewTab` (`frontend/app/tab/quick-fork.test.ts`), covering the resolution edge cases (no layout model, nothing focused, non-agent block, blank picker tab, in-pane block-stack `activeBlockId` precedence) and the full orchestration (fork → create tab → place block → launch, with the right overrides and explicit cross-tab targeting), plus failure paths.
- [x] `npx vitest run` — 2868 tests pass; one unrelated pre-existing flaky timeout in `tool-renderers/registry.test.ts` confirmed passing in isolation.
- [x] `npx tsc --noEmit` clean, zero errors project-wide.
- [ ] **Not manually verified end-to-end in a live running instance** — no access to one in this environment. The RPC sequence is verified against the real backend/frontend code paths and covered by orchestration-level tests, but an actual click-through (right-click → new tab appears → agent is running with history) was not possible here. Flagging this explicitly — please pay particular attention to this in review.


a5af:
@codex review

PR #2728, opened 2026-08-22
feat(srv): retention/GC for native memory version history

Agent1-asaf:
## Summary
- Closes `SPEC_MEMORY_VERSION_CONTROL_AND_ARMORY_AUDIT_2026_08_19.md` §7.1, an open operator question in the original spec ("no pruning is proposed in v1... flag if unbounded growth is a concern").
- Operator-confirmed policy (2026-08-22): hybrid age + min-count floor — a version is pruned only once it is **both** older than 90 days **and** ranked beyond the 50 most-recent versions for its `(agent_id, filename)`. This protects a rarely-touched file's full history from age-only pruning, and caps a hyperactive file's growth that a count-only floor wouldn't bound in time.
- New `Store::agent_native_memory_version_prune` (single `ROW_NUMBER() OVER`-based SQL statement) plus a new daily background sweep (`agentmux-srv/src/backend/native_memory_retention.rs`), spawned once at startup alongside `native_memory_drift`'s own detection sweeps — a much slower cadence (24h vs. 30s) since pruning is housekeeping, not latency-sensitive.

## Test plan
- [x] New unit tests for `agent_native_memory_version_prune`/`agent_native_memory_version_list_distinct_files` (22 tests total in that module) — floor-only, age-only, both-combined, and per-agent/file scoping cases.
- [x] New unit tests for `native_memory_retention::prune_once` (multi-file sweep coverage, no-op cases).
- [x] Proved test discrimination: reverted the count-floor guard in the SQL, confirmed 3 tests fail with the exact expected signature, restored, confirmed clean.
- [x] Full suite: `cargo test -p agentmux-srv --bin agentmux-srv -- --test-threads=1` — 2653 passed, 0 failed.
- [x] `cargo build -p agentmux-srv --bin agentmux-srv` — clean, confirms `main.rs` wiring compiles.

### DIFF
--- a/.changesets/1787418543-feat-srv-retention-gc-for-native-memory-version-history-8e0o.md
+++ b/.changesets/1787418543-feat-srv-retention-gc-for-native-memory-version-history-8e0o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): retention/GC for native memory version history
--- a/.changesets/1787418543-feat-srv-retention-gc-for-native-memory-version-history-8e0o.md
+++ b/.changesets/1787418543-feat-srv-retention-gc-for-native-memory-version-history-8e0o.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
 feat(srv): retention/GC for native memory version history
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2729, opened 2026-08-22
feat(agent-pane): Tier 3 predictive countdown before auto-compact

Agent1-asaf:
## Summary
- Ships Tier 3 of `SPEC_COMPACTION_DETECTION_AND_HANDLING_2026_07_31.md` §7 — the last documented gap here was presentation, not data: `compactionThreshold(contextWindow) - currentTokens` was already computable every turn, but only visible as a number behind a hover tooltip.
- `AgentComposerStrip.tsx` now shows explicit inline countdown text (`~4.2k to auto-compact`) next to the existing context readout, once the fill level reaches the `mid` band or above.
- Escalates at the `critical` band with a bold weight + leading `⚠` glyph, layered on the band's existing pulsing red color — considered and deliberately skipped a new standalone banner component (like `AgentDisconnectedBanner.tsx`) to avoid destabilizing this strip's already-fragile, hand-tuned tiered-wrap responsive layout (its own file header calls this out).
- Labeled everywhere (inline tooltip + main context tooltip) as predicting **auto-compaction only** — a manual `/compact` can happen at any fill level and isn't predicted by this.

## Test plan
- [x] New `AgentComposerStrip.test.tsx` (7 tests): band gating (silent below mid), correct remaining-token math, critical-band escalation class, zero-clamp past the threshold, tooltip wording.
- [x] Proved test discrimination: stubbed `compactionCountdownText` to always return `null`, confirmed 4 of 7 tests fail with the expected signature, restored, confirmed clean.
- [x] `npx vitest run` — 2933 passed, 1 pre-existing unrelated timeout flake (passes in isolation), 21 unrelated pre-existing failures from an untracked `amramebgi/` scratch directory's missing `joi` dependency (not touched by this change).

### DIFF
--- a/.changesets/1787419367-feat-agent-pane-tier-3-predictive-countdown-before-auto-comp-2cwp.md
+++ b/.changesets/1787419367-feat-agent-pane-tier-3-predictive-countdown-before-auto-comp-2cwp.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): Tier 3 predictive countdown before auto-compact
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2730, opened 2026-08-22
feat(browser-pane): add global bookmarks + icon Go button

Agent3-asaf:
## Summary
- Adds a bookmarks button to the browser pane nav bar (between Reload and the address bar), opening a `FlyoutMenu` (reusing the same menu primitive as the hamburger menu / widget bar / right-click menus) with a "Bookmark this page" toggle + the saved bookmark list.
- Bookmarks persist to a new, global (cross-channel) `~/.agentmux/shared/browser-bookmarks.json` file rather than `settings.json` — `settings.json` isolation is whole-file-only per `SPEC_SETTINGS_ISOLATED_BY_CHANNEL_2026_08_19`, so a key inside it can't stay global while the rest of the file is channel-isolated. `shared_dir` is this codebase's existing escape hatch for data that must always be global (agent registry, transcripts, provider auth) — reused here so bookmarks survive across `task dev` branches and local package builds instead of resetting per channel.
- Replaces the nav bar's text "Go" button with an icon-only `fa-arrow-right` button, matching the icon-only Back/Forward/Reload buttons already in the same row.

## Design doc
`docs/specs/SPEC_BROWSER_PANE_BOOKMARKS_AND_GO_ICON_2026_08_22.md` — full analysis, prior-art (a previous message-bookmark feature was removed for scope-creep reasons, `SPEC_REMOVE_BOOKMARKS_2026_06_11`), persistence-mechanism tradeoffs, and a documented unhappy-paths table (empty list, corrupt file, concurrent writes, favicon-load failure, keyboard-only dismissal gap in `FlyoutMenu`, etc.).

## What's new
- `agentmux-srv/src/backend/bookmarks_store.rs` — atomic (tmp+rename) read/write of the shared bookmarks file; missing file → empty list, corrupt file → loud error (never silently discards).
- `agentmux-srv/src/server/app_api/bookmarks.rs` — `bookmarks.list` / `bookmarks.set` RPC handlers.
- `frontend/app/store/rpc-api/bookmarks.ts` — RPC wrapper, spliced into `RpcApi`.
- `frontend/app/view/browser/browser-bookmarks-logic.ts` — pure toggle/dedupe-by-URL logic, unit-tested independently of the Solid component.
- `frontend/app/view/browser/browser-nav-bar.tsx` — the bookmarks button + menu wiring, plus the Go-icon swap.

## Test plan
- [x] `cargo test -p agentmux-srv` — 2647 passed (includes new `bookmarks_store`/`app_api::bookmarks` unit tests)
- [x] `npx vitest run` — full frontend suite passes (one unrelated pre-existing flaky timeout in `tool-renderers/registry.test.ts`, confirmed unrelated and passes in isolation)
- [x] `tsc --noEmit` clean
- [ ] Manual click-through in a running app (`task dev`) — not yet done, worth doing before merge

### DIFF
--- a/.changesets/1787423206-feat-browser-pane-add-global-bookmarks-icon-go-button-wtuj.md
+++ b/.changesets/1787423206-feat-browser-pane-add-global-bookmarks-icon-go-button-wtuj.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(browser-pane): add global bookmarks + icon Go button
--- a/agentmux-srv/src/backend/bookmarks_store.rs
+++ b/agentmux-srv/src/backend/bookmarks_store.rs
@@ -1,0 +1,172 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Browser-pane bookmarks — a flat list of saved URLs, deliberately stored
+//! under `shared_dir` (`~/.agentmux/shared/browser-bookmarks.json`) rather
+//! than `settings.json`. `settings.json` isolation is whole-file only (see
+//! `docs/specs/SPEC_SETTINGS_ISOLATED_BY_CHANNEL_2026_08_19.md` §6 — per-key
+//! tiering was considered and rejected there), so a key inside it can't stay
+//! global while the rest of the file is channel-isolated. `shared_dir` is
+//! this codebase's existing mechanism for data that must always be global
+//! regardless of channel/isolation flags — the same directory already backs
+//! the agent registry, transcripts, and `provider_auth_dir()`. See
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

Agent3-asaf:
Fixed in c5f1e543d — bookmark clicks now route through the same `navigateTo()` path (model update + `browser_pane_navigate`/`createPane`) that the address-bar submit uses, instead of calling `model.navigate()` alone. Good catch.

Agent3-asaf:
Fixed in e6d302ed8 — the loading row is now label-only rather than a fake, non-animating `fa-spinner` icon. `FlyoutMenu`'s default item renderer has no way to pair `fa-spin` through the `icon` field, so trying to force it in would've been the wrong fix.

PR #2731, opened 2026-08-22
fix(agent-pane): "New Agent" heading + harness-only template card icons

GenericAgentX-asaf:
## Summary
- Agent pane template section header: "+ New from template" → "New Agent".
- Template cards (`AgentCard.tsx`) now render a plain harness `ProviderLogo` instead of `DualProviderLogo` — drops the small overlaid vendor-logo badge, which is redundant here since the section is explicitly about picking a harness (the hint text right below the header already says "you'll pick which model it uses next").
- "My Agents" (`MyAgentsList.tsx`) is untouched — still uses `DualProviderLogo` with its badge, where the model vendor is a real fact about an already-launched instance.

Full design: `docs/specs/SPEC_AGENT_PICKER_TEMPLATE_SECTION_CLEANUP_2026_08_22.md`

## Test plan
- [x] New `AgentCard.test.tsx` — asserts `ProviderLogo` renders (not `DualProviderLogo`) for template cards, including a provider with a `model_vendor_base_url` override; install-ribbon rendering still covered.
- [x] `AgentPicker.test.tsx` header-text assertion updated to "New Agent"; all 21 pre-existing tests still pass.
- [x] `tsc --noEmit` clean, zero errors.

### DIFF
--- a/.changesets/1787424964-fix-agent-pane-new-agent-heading-harness-only-template-card--44ym.md
+++ b/.changesets/1787424964-fix-agent-pane-new-agent-heading-harness-only-template-card--44ym.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): New Agent heading + harness-only template card icons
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2732, opened 2026-08-22
fix(agent-pane): "modelit" typo in the New Agent hint text

GenericAgentX-asaf:
## Summary
Follow-up to #2731 — that PR's auto-merge fired on the pre-fix commit due to a race between reagent's approval and a fix I was still pushing, so this typo shipped to `main` for a short window. This PR is exactly the fix commit that PR was supposed to include.

- The hint text's `</strong>` closing tag was directly followed by a line break before "it uses next" — JSX drops whitespace entirely at that exact tag-to-text boundary instead of preserving it, rendering "model" and "it" concatenated as **"modelit"** (user-reported, confirmed live on `main`).
- Fixed by keeping `</strong>` and "it" on the same source line with a real space between.
- The existing hint test's substring checks (`toHaveTextContent("model")`) didn't catch this — "model" is still a substring of "modelit" — strengthened to assert the actual word boundary.
- Also includes the fix for codex P1 on #2731: `AgentCard.test.tsx`'s mock spies now use `vi.hoisted()` since `vi.mock(...)` is hoisted above plain top-level `const` declarations.

## Test plan
- [x] All 24 tests in `AgentPicker.test.tsx` + `AgentCard.test.tsx` pass, including the new word-boundary assertion.
- [x] Confirmed the bug is present in current `main` (`git show origin/main:...AgentPicker.tsx` shows the unfixed `</strong>\n it uses next.` line break) before this fix.

### DIFF
--- a/.changesets/1787425657-fix-agent-pane-modelit-typo-in-the-new-agent-hint-text-xm4g.md
+++ b/.changesets/1787425657-fix-agent-pane-modelit-typo-in-the-new-agent-hint-text-xm4g.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): modelit typo in the New Agent hint text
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2733, opened 2026-08-22
feat(browser-pane): render real per-site favicons in the bookmarks menu

Agent3-asaf:
## Summary
Follow-up to #2730 (merged) — the user asked for real favicons in the bookmarks menu instead of a generic icon, which #2730's spec had deliberately deferred pending confirmation this was wanted.

- Bookmark rows now render the actual saved favicon (an ``), falling back to a plain globe icon on a missing `favicon_url` or an `onError` load failure — matches this app's existing "no favicon → globe" convention (`BrowserViewModel.viewIcon`).
- `FlyoutMenu`'s default item renderer only ever templates a FontAwesome name (`fa-solid fa-fw fa-${icon}`) and never handles the `JSX.Element` branch of `MenuItem.icon` (typed but previously unused by any consumer) — this adds a `renderMenuItem` override scoped to just this menu: render `item.icon` as-is when it's a `JSX.Element`, otherwise fall back to the original string-icon rendering (used by the pinned "Bookmark this page" toggle row and the loading/empty-state rows).
- Spec updated (`docs/specs/SPEC_BROWSER_PANE_BOOKMARKS_AND_GO_ICON_2026_08_22.md`) to reflect this as a same-day follow-up rather than a deferred v1 limitation.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `npx vitest run frontend/app/view/browser frontend/app/store/browser-pane-state` — 123 passed
- [x] **Live-verified** against a running `task dev` instance via CDP (`Runtime.evaluate` over the CEF remote-debugging port): opened the bookmarks menu, confirmed a real bookmark row renders ``, and the star-toggle row is unaffected (`<i class="fa-solid fa-fw fa-star ...">`).

### DIFF
--- a/.changesets/1787425691-feat-browser-pane-render-real-per-site-favicons-in-the-bookm-k7ok.md
+++ b/.changesets/1787425691-feat-browser-pane-render-real-per-site-favicons-in-the-bookm-k7ok.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(browser-pane): render real per-site favicons in the bookmarks menu
--- a/.changesets/1787425691-feat-browser-pane-render-real-per-site-favicons-in-the-bookm-k7ok.md
+++ b/.changesets/1787425691-feat-browser-pane-render-real-per-site-favicons-in-the-bookm-k7ok.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
 feat(browser-pane): render real per-site favicons in the bookmarks menu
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

Agent3-asaf:
Both fixed in e204538ab:

- **P1**: `loadBookmarks()` now only flips `bookmarksLoading` before the pane's first-ever successful load. Later opens keep rendering the last-known list while the fresh fetch resolves, and the result is only written back when it's actually different by value — so `bookmarkMenuItems`' memo doesn't rerun and `<For>` (which diffs by reference) never tears down/remounts unchanged rows. Traced through the reactivity graph to confirm; also re-ran `tsc`/the full browser test suite (123 passed) after the change.
- **P2**: changeset bumped `patch` → `minor`.

PR #2735, opened 2026-08-22
feat(tabs): quick-fork keybinding, non-Claude fallback banner, inherit-identity variant

AgentX-asaf:
## Summary
Phases 3-4 of `SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md` (Phases 1-2 shipped in #2725/#2727):

- **Keybinding**: `Cmd:Shift:t` quick-forks the currently *active* tab into a new tab (deliberately reads `atoms.activeTabId()`, not the stale `staticTabId`).
- **Non-Claude fallback banner**: when a forked agent's effective provider has no `--fork-session` equivalent and there was a parent session to lose, `quick-fork.ts` sets a `quickfork:noHistoryFallback` meta flag on the new block once launch succeeds; `ForkProviderFallbackBanner` (cloned from `AgentDisconnectedBanner`'s no-dismiss-button pattern) reads it and shows a persistent, non-dismissable-by-accident note — per spec §4.4's honesty requirement.
- **Inherit-identity variant** (§5, Phase 4): a second "⑂ Quick-fork (inherit identity)" entry in the tab's context panel opts into the source agent's own bound Armory account (via `ListAgentIdentitiesCommand`) instead of the unbound-by-default path the plain quick-fork action takes.
- Documented the tab-strip "+" split-button (§4.3's deferred item) as a deliberate skip, not a gap — ship right-click-only, revisit only if discoverability proves a real problem.

Spec doc's phasing table updated to mark Phases 3-4 complete.

## Test plan
- [x] `npx vitest run` — 2882 passed, 1 pre-existing/unrelated timeout in `tool-renderers/registry.test.ts` that passes standalone (not touched by this change)
- [x] Added 9 new tests in `quick-fork.test.ts` covering `opts.inheritIdentity` (success, empty-list fallback, RPC-rejection fallback, opt-out) and the non-Claude fallback meta flag (set/not-set across provider, no-session, launch-failure, and `SetMetaCommand`-rejection cases)
- [ ] Manual verification by repo owner once merged


a5af:
@codex review

AgentX-asaf:
Fixed in e39fb7231 — both reagent's P1 and Codex's P1 caught the same bug: `opts.inheritIdentity` filters `ListAgentIdentitiesCommand`'s results to the link matching the fork's own canonical effective provider now, instead of blindly taking `links[0]`. Also added `ForkProviderFallbackBanner.test.tsx` per reagent's P2.

AgentX-asaf:
Fixed in 61f7eed26 — good catch, switched to `lastLinkedAccountId` (`provider-id-aliases.ts`) instead of a raw `.find()`, matching `launch-flow.ts`/`useAgentControllerStatus.ts`'s established pattern for this exact bug class. Added a test for the canonical-vs-alias-row "last wins" case.

PR #2736, opened 2026-08-22
fix(statusbar): stop double-applying chrome zoom to status bar popovers

AgentA-asaf:
## Summary

Found via inspecting my own live pane (`mcp__agentmux__UIQuery`/`UIClick`): every status bar popover (Token Usage, CPU Cores, Disk Volumes, the status-bar-tip balloon, and the shared `.status-bar-popover` class) renders offset and shrunk relative to its own declared CSS whenever chrome zoom (`--zoomfactor`) is not `1`.

**Root cause**: these popovers are Portal'd to `document.body` and positioned via `computeMenuPosition` from an anchor rect that's already real, post-zoom pixels (the trigger button lives under `.status-bar`, which legitimately self-applies `zoom: var(--zoomfactor)` as an in-flow container). Each popover's own root ALSO self-applies `zoom: var(--zoomfactor)` — since `--zoomfactor` is a root-level CSS custom property, Portal-moving the DOM node doesn't escape it, so the already-real coordinates get rescaled a second time.

Full diagnosis: `docs/specs/SPEC_STATUS_BAR_POPOVER_DOUBLE_ZOOM_OFFSET_2026_08_22.md`

**Fix**: remove the self-`zoom:` from the 5 affected selectors. Their position is already correct without it — the tradeoff is losing automatic font/padding scaling for these specific popovers (a cosmetic nicety) in exchange for never being mispositioned (a correctness bug). `.status-bar`/`.window-header` themselves are untouched — they're in-flow containers, not Portal'd/positioned, so their own `zoom:` is correct as-is.

## Test plan

- [x] Live-verified via CDP against a running `task dev` instance: forced `--zoomfactor: 0.65`, confirmed the popover's computed `zoom` is now `1` and its rendered rect exactly matches its declared inline `left`/`top`/`width` (previously diverged by exactly 0.65x on every dimension).
- [x] `tsc --noEmit` clean (CSS-only change).
- [x] `stylelint` on all 5 modified files — 3 pre-existing hex-color violations, none introduced by this change, none near the edited lines.
- No automated test added: this is a pure CSS-composition bug (two `zoom:` declarations compounding on a `position: fixed` box) that jsdom does not implement — CSS `zoom` has no layout effect in jsdom, so nothing in that environment could distinguish buggy from fixed. Guarded with an explanatory comment at each fix site instead.

### DIFF
--- a/.changesets/1787427171-fix-statusbar-stop-double-applying-chrome-zoom-to-status-bar-r5f2.md
+++ b/.changesets/1787427171-fix-statusbar-stop-double-applying-chrome-zoom-to-status-bar-r5f2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(statusbar): stop double-applying chrome zoom to status bar popovers
--- a/docs/specs/SPEC_STATUS_BAR_POPOVER_DOUBLE_ZOOM_OFFSET_2026_08_22.md
+++ b/docs/specs/SPEC_STATUS_BAR_POPOVER_DOUBLE_ZOOM_OFFSET_2026_08_22.md
@@ -1,17 +1,23 @@
 # SPEC — Status bar popovers render offset (and undersized) under Chrome zoom
 
-**Date:** 2026-08-22
-**Type:** Bug diagnosis (root-caused precisely; fix NOT yet implemented — this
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2738, opened 2026-08-22
fix(muxlog): srv logs honor AGENTMUX_LOG_DIR, fix channels/ glob depth mismatch

AgentY-asaf:
## Summary

Ext 1 of `docs/reports/REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22.md` (included in this PR for context — findings from a live debugging session where a dispatched subagent never appeared in Swarm, and the diagnostic tools couldn't reliably say why).

Two compounding bugs found together:

1. **agentmux-srv wrote logs to the shared root regardless of channel.** `agentmux-cef` (host) already writes to a per-instance log dir; the shell-integration scripts (`bash.sh`/`zsh.sh`/`fish.fish`/`pwsh.ps1`) already assume `$AGENTMUX_LOG_DIR` holds both host AND srv pointer files. Only srv didn't honor it — every instance at the same version interleaved into one shared file, un-attributable after the fact. Fixed: srv now prefers `$AGENTMUX_LOG_DIR` (same var `agentmux-cef` already uses), falling back to the old shared-dir behavior when unset.

2. **`muxlog.mjs`'s own `channels/` discovery glob had the wrong depth.** `channels/*/versions/*/*/logs` vs. the real on-disk `channels/<channel>/versions/<v>/logs` (confirmed against a real running instance's own `AGENTMUX_LOG_DIR`) — one wildcard segment too many, so it matched **zero** channel-build logs, on any platform, for as long as that source existed. `logRoots()` now tries both depths (deduped).

Together: srv logs now land where `muxlog` can actually find them per-channel, so `muxlog swarm -i <channel>` / `errors -i <channel>` / etc. resolve to the right instance instead of a same-version sibling's interleaved shared log.

## Test plan
- [x] New `muxlog.test.mjs` (7 cases) pins the glob depth-matching fix against a real temp directory tree — `npx vitest run agentmux-srv/src/backend/shellintegration/muxlog.test.mjs` passes
- [ ] Manual verification against a rebuilt portable (can't retroactively fix already-running instances — needs a relaunch on the new binary)

### DIFF
--- a/.changesets/1787428853-fix-muxlog-srv-logs-honor-agentmux-log-dir-fix-channels-glob-zj2j.md
+++ b/.changesets/1787428853-fix-muxlog-srv-logs-honor-agentmux-log-dir-fix-channels-glob-zj2j.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(muxlog): srv logs honor AGENTMUX_LOG_DIR, fix channels/ glob depth mismatch
--- a/agentmux-srv/src/backend/shellintegration/muxlog.mjs
+++ b/agentmux-srv/src/backend/shellintegration/muxlog.mjs
@@ -17,6 +17,7 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { execFileSync } from "node:child_process";
+import { pathToFileURL } from "node:url";
 
 const HOME = os.homedir();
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2739, opened 2026-08-22
fix(editor): strip Windows' \?\ prefix so live-reload matching works

Agent1-asaf:
## Summary
- Live-confirmed bug: an editor pane opened via the `OpenEditor` MCP tool never picks up an on-disk change made after the pane is open — reproduced end-to-end against a real running instance, confirmed by the repo owner directly checking the pane.
- Root cause: Rust's `Path::canonicalize()` unconditionally prepends the `\?\` extended-length prefix on Windows (`\?\UNC\` for a network share) — well-documented std behavior, and this backend's own comments already flag it in two sibling files (`editor_file_watcher.rs`, `media_file_watcher.rs`) for the backend's own internal path matching. But `EditorFileWatcher`'s published `editor:file_changed` WPS event ships that `\?\`-prefixed path straight to the frontend, and the frontend's `canonicalizePath()` never stripped it. A tab's own `filePath` is never `\?\`-prefixed, so the live-reload comparison in `_handleExternalFileChanged` could never match — silently, unconditionally, on every Windows install.
- Fix: `canonicalizePath()` (`frontend/app/store/editor-pane-state-store.ts`) now strips `\?\` and `\?\UNC\` before its existing normalization.

## Test plan
- [x] Two new regression tests in `editor-pane-state-store.test.ts`, proven to fail without the fix (exact string mismatch) and pass with it restored.
- [x] Full `vitest run` on the file (37 tests) — clean.
- [x] `tsc --noEmit` — clean.
- [x] **Live end-to-end verification**: launched an isolated `task dev` instance from this branch (window titled "agent1", per the repo owner's request), opened a probe file via `OpenEditor`, edited it on disk, confirmed the pane now updates live — repo owner confirmed directly ("yes great it works").

### DIFF
--- a/.changesets/1787434387-fix-editor-strip-windows-verbatim-prefix-so-live-reload-matc-l5fb.md
+++ b/.changesets/1787434387-fix-editor-strip-windows-verbatim-prefix-so-live-reload-matc-l5fb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(editor): strip Windows' verbatim prefix so live-reload matching works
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2740, opened 2026-08-22
fix(jekt): match plural forms of whole-word sensitive keywords

GenericAgentX-asaf:
## Summary

Follow-up to #2719. One of the two findings surfaced while writing the jekt tier-classification regression suite: `contains_whole_word` (`agentmux-srv/src/backend/reactive/sanitize.rs`) required a non-alphanumeric character immediately after a keyword match. A plural "s" is alphanumeric, so it never satisfied that check — every `SENSITIVE_WHOLE_WORD_KEYWORDS` entry (`pat`, `token`, `secret`, `password`, `credential`, `keychain`) matched its singular form but not its plural.

Concretely, before this fix: "your credentials were leaked" reached `TIER=coord`, not `sensitive`.

## Fix

Accept a single trailing "s" before requiring the word-boundary character, so plurals match the same as their singular. Compound words that merely start with a keyword still correctly do NOT match — the boundary is still required right after that trailing "s" — verified with a new false-positive guard test (`tokensmith`, `secretsauce`, `credentialstore.example` all still correctly pass as non-sensitive).

## What changed

- `sanitize.rs`: `contains_whole_word` now also checks for `<keyword>s` + boundary, not just `<keyword>` + boundary.
- `tests.rs`: the old `test_keyword_match_known_gap_plural_forms_not_caught` (which asserted the bug) now asserts the fix (`test_keyword_match_whole_word_plural_forms_are_caught`); added `test_keyword_match_plural_rule_does_not_introduce_new_false_positives` for the compound-word edge cases.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv reactive::tests::` — 101/101 passed

### DIFF
--- a/.changesets/1787434387-fix-jekt-match-plural-forms-of-whole-word-sensitive-keywords-srl5.md
+++ b/.changesets/1787434387-fix-jekt-match-plural-forms-of-whole-word-sensitive-keywords-srl5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(jekt): match plural forms of whole-word sensitive keywords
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2741, opened 2026-08-22
Korp@claudius: fix(muxlog): prefer caller's own $AGENTMUX_CHANNEL over global freshest when resolving a log

AgentY-asaf:
## Summary

Ext 2 of `docs/reports/REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22.md` §2.1.

The report proposed porting `phases`' content-verification resolver to `swarm`/`errors`/`auth`/`bridge`. Those recipes don't take a block-id argument, so there's no "needle" to check log content against. Simpler, more precise fix: `$AGENTMUX_CHANNEL` is already set in every agent pane's environment — use the caller's own identity as ground truth instead.

- `pickCandidate()` (new, pure, extracted from `resolveFile`) now prefers a candidate matching the caller's own channel when no explicit `-i` is given, falling back to "freshest overall" only when nothing matches.
- `errors` had its own separate ad-hoc `.file`-only filter with no own-channel default at all — now routed through the shared resolver like every other recipe.
- `filterByInstance()` (extracted, exported) fixes a case-sensitivity bug: the original inline filter only lowercased `.file`, not `.source`/`.version`.
- `phases`' own filtering reuses `filterByInstance()` too (same case-sensitivity fix), keeping its existing content-verification as the real correctness backstop, unchanged.

## Test plan
- [x] 11 new test cases (`filterByInstance` + `pickCandidate`) — `npx vitest run agentmux-srv/src/backend/shellintegration/` passes 40/40

### DIFF
--- a/.changesets/1787429114-fix-muxlog-prefer-caller-s-own-agentmux-channel-over-global--tos2.md
+++ b/.changesets/1787429114-fix-muxlog-prefer-caller-s-own-agentmux-channel-over-global--tos2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(muxlog): prefer caller's own $AGENTMUX_CHANNEL over global freshest when resolving a log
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2742, opened 2026-08-22
Korp@claudius: feat(muxlog): ls gains a LIVE column via real TCP liveness probe of each instance's ipc-port

AgentY-asaf:
## Summary

Ext 3 of `docs/reports/REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22.md`.

**Stacked on #2741 (Ext 2)** — this branches from `korp/muxlog-own-channel-preference`, not `main`, since it extends the same file. Once #2741 merges and its branch is deleted, GitHub should auto-retarget this PR to `main`.

A log's mtime says "something was written recently," not "this instance is running right now" — the two look identical for a process that died right after its last log line. `agentmux-cef` already writes a real, checkable liveness signal per instance: `<data-dir>/ipc-port-<hash>` (`agentmux-launcher/src/second_instance.rs`), containing `port:token` for the host's cross-window IPC listener.

`checkLiveness(logDir)` finds the sibling `data` dir, globs for any `ipc-port-*` file (doesn't replicate the launcher's FNV-1a hash — just finds whatever's there), and does a raw TCP connect to the recorded port with a short timeout. Returns `"live" | "dead" | "?"` (genuinely unknown, never a false "confirmed dead").

`muxlog ls` now shows this as a **LIVE** column, probing every listed instance concurrently.

## Test plan
- [x] 7 new test cases (`siblingDataDir` + `checkLiveness`, the latter against a real ephemeral TCP listener) — 47/47 passing overall in this file's suite
- [x] **Verified end-to-end against this machine's real running instances**, not just unit tests: both live channel-build instances correctly show `live`; dev-mode/shared entries show `?` (honest — dev builds don't appear to write the same ipc-port file)

### DIFF
--- a/.changesets/1787434874-feat-muxlog-ls-gains-a-live-column-via-real-tcp-liveness-pro-huv1.md
+++ b/.changesets/1787434874-feat-muxlog-ls-gains-a-live-column-via-real-tcp-liveness-pro-huv1.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxlog): ls gains a LIVE column via real TCP liveness probe of each instance's ipc-port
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2745, opened 2026-08-22
Korp@claudius: feat(muxspect): cross-instance find — which channel has this block_id/agent

GenericAgentX-asaf:
## Summary

Ext 4 of `docs/reports/REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22.md`. `SPEC_MUXSPECT_CROSS_INSTANCE_FIND_2026_08_22.md`.

`muxspect` (Phase 1) only ever queries the instance the caller is already inside — it cannot answer "does a *different* running instance have a controller/dispatch matching this block id or agent name." The live debugging session that produced the report needed exactly this and had no tool for it.

New `GET /api/v1/muxspect/find?block_id=X` or `?agent=X`, and matching CLI `muxspect find <block_id_or_agent>`:

- **Host tier** (no network): `state.process_broker.list()`, cross-referenced against `reactive_handler.list_agents()` for agent-name queries.
- **Cross-channel tier**: reads the shared reactive registry (`resolve_shared_reactive_dir` + `list_all_shared` — the exact mechanism `handle_muxspect_conversations` already uses). Each entry already carries `block_id`/`agent_id`/`channel`/`local_url`, so matching costs **zero** network calls — only an actual match gets a forwarded `describe` call for detail.

CLI dispatches by shape: UUID-looking query → `block_id`; anything else → `agent`.

## Known limitations (called out in the spec, not hidden)
- Not verified against a live rebuilt instance — would need a full srv rebuild+relaunch. Verified instead via `cargo check`, CLI unit tests, and direct mirroring of `handle_muxspect_conversations`'s already-proven pattern.
- No dedicated Rust unit/integration test for the handler itself — no existing precedent in this file for testing its async, `AppState`-composing handlers in isolation.

## Test plan
- [x] 3 new `parseArgs` cases (`find <uuid>`, `find <agent>`, flag-anywhere) — 25/25 passing in `muxspect.test.mjs`
- [ ] Live end-to-end verification (blocked on a rebuild — see limitations above)

### DIFF
--- a/.changesets/1787436129-feat-muxspect-cross-instance-find-which-channel-has-this-blo-2qw6.md
+++ b/.changesets/1787436129-feat-muxspect-cross-instance-find-which-channel-has-this-blo-2qw6.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxspect): cross-instance find — which channel has this block_id/agent
--- a/agentmux-srv/src/backend/shellintegration/muxspect.mjs
+++ b/agentmux-srv/src/backend/shellintegration/muxspect.mjs
@@ -54,6 +54,17 @@ Usage:
                                   name — resolves host first, then
                                   cross-channel (other AgentMux channels on
                                   this same host); does not reach LAN/WAN
+  muxspect find <block_id_or_agent> [--json]
+                                  which running instance(s), if any, have a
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2746, opened 2026-08-22
fix(tabs): quick-fork lands as a pane-stack tab, not a new window tab

AgentX-asaf:
## Summary
Corrects the destination + trigger surface of the quick-fork feature (PR #2725/#2727/#2735), caught live by the repo owner during manual testing:

- **What shipped:** right-click a WINDOW tab → "⑂ Quick-fork to new tab" → opens the fork in a brand-new top-level window tab.
- **Why it was wrong:** forking is a per-agent action. The codebase already has a per-agent right-click surface for exactly that (`AgentViewModel.getBodyContextMenuItems`, home to "Agent History") — the window tab-strip's own context menu is ambiguous once a tab can ever hold more than one pane, and puts a per-agent action on a per-tab surface. Separately, `launchAgentDefinition`'s own `targetBlockId` param already existed (shipped 2026-07-20, a month before this feature) specifically for "the fork-tab-strip `+` action" per `SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15.md`'s own (never-wired) design — pushing the fork onto the SAME pane's own block-stack, not a new window tab.
- **Fix:** moved `quickForkTabToNewTab` (`frontend/app/tab/quick-fork.ts`) to `quickForkAgent` (`frontend/app/view/agent/quick-fork.ts`). Forks now push a new block onto the source pane's own stack via `pushBlockOntoStack` — the same primitive `open-history-tab.ts`'s "Agent History" entry and the pane tab strip's own "+" already use — instead of `WorkspaceService.CreateTab`. Triggered from `getBodyContextMenuItems`, right next to "Agent History".
- Removed the now-wrong tab-strip context-menu entries, the `Cmd:Shift:t` keybinding (its target — the active window tab — no longer matches this action), and `launchAgentDefinition`'s now-dead `targetTabId` param.
- Both spec docs (`SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md`, `SPEC_AGENT_PANE_FORKS_AND_AUX_PINS_2026_06_15.md`) updated with correction notices and phasing cross-references — this work is now understood to complete the older spec's Phase 3 "+"-affordance, not a separate sibling feature.

All Phase 1 backend fixes (history carryover, lineage-wide naming), the non-Claude fallback banner, and the inherit-identity provider-matching logic (`lastLinkedAccountId`) carry over unchanged — only *where the fork lands* and *where it's triggered from* changed.

## Test plan
- [x] `npx vitest run` — full suite green (2897 passed), 1 pre-existing/unrelated flaky timeout in `tool-renderers/registry.test.ts` (passes standalone, not touched by this change)
- [x] New `frontend/app/view/agent/quick-fork.test.ts` (19 tests) covering the pane-stack push path, orphaned-block cleanup on a mid-flight pane close, inherit-identity provider filtering, and the non-Claude fallback flag — replacing the deleted window-tab-scoped test file
- [ ] Manual verification by repo owner (right-click an agent pane's body → "Quick-fork" / "Quick-fork (inherit identity)")


a5af:
@codex review

a5af:
@codex review

AgentX-asaf:
Both P1s fixed in c8ece2987 — good catch, this was a real race. Captured `atoms.activeTabId()` once, synchronously, before any of the function's awaits, and threaded it through both `pane.open`'s `tab_id` and `launchAgentDefinition`'s re-added `targetTabId` param.

The P2 (clean up on launch failure) is also fixed — pops the new block back out via `closeBlockInStack` on a failed launch, same as the existing "pane closed mid-flight" cleanup.

a5af:
@codex review

AgentX-asaf:
Checking each of the three latest findings against the actual committed code:

- **"Pin pane.open to the source pane's tab" (P1, quick-fork.ts:163)** — this was already fixed in c8ece2987 (the previous push): the `pane.open` call at that line passes `tab_id: ownerTabId` explicitly (`ownerTabId` captured synchronously via `atoms.activeTabId()` before any `await`). The comment's premise ("this request has neither `tab_id` nor `split_reference_block_id`") no longer matches the code — looks like this review pass re-surfaced a pre-fix comment rather than re-reading the diff.
- **"Remove the pane-stack tab when launch fails" (P2, quick-fork.ts:199)** — also already fixed in c8ece2987 — the `!launched` branch calls `closeBlockInStack(layoutModel, freshNode.id, paneOpenResult.block_id)`. Same likely stale re-surface.
- **"Notify the user when the fork launch returns false" (P2, quick-fork.ts:192)** — this one's real and new: the cleanup added in c8ece2987 was itself silent. Fixed in 8e14564d2 — pushes the same error notification the outer `catch` already uses for other failure modes.

AgentX-asaf:
Re-checked the latest review round (pullrequestreview-5001189551) against the current code:

- The "Pin pane.open to the source pane's tab" (P1) and "Remove the pane-stack tab when launch fails" (P2) findings are unchanged re-posts of the same two already fixed in 8e14564d2 (verified again — `tab_id: ownerTabId` is present at quick-fork.ts:186, `closeBlockInStack` runs in the `!launched` branch).
- The "Notify the user when the fork launch returns false" (P2) finding was also already fixed in 8e14564d2 (the notification landed in that same commit, before this review ran).
- **"Remove the new stack block when launch rejects" (P2) is real and new** — good catch. `launchAgentDefinition`'s own doc comment claims it never throws, but `resolveEffectiveLaunchProvider`, `checkNodejsForProvider`, and `ensureAuthDir` all run unguarded before its internal try/catches, so it genuinely can reject. Fixed in 1cadde8c2: extracted the cleanup into `cleanupPushedBlock()`, now called from both the `!launched` branch and the outer `catch`.

PR #2747, opened 2026-08-22
feat(agent-config): protect a pre-existing project CLAUDE.md from overwrite

GenericAgentX-asaf:
## Summary
Both config-materializing call sites (`agent.open`'s `write_agent_config_files`, and the `WriteAgentConfig` "click Launch" path) wrote `CLAUDE.md` via a raw `std::fs::write` — no read-before-write, no backup, no merge. That path is the exact same slot Claude Code's own project-`CLAUDE.md` discovery uses, so any agent launched with its working directory pointed at a real project with its own hand-authored, git-tracked `CLAUDE.md` had it silently, fully replaced on every launch.

`write_claude_md_respecting_ownership()` fixes this, extending the same "distinguish AgentMux-authored from user-authored" philosophy the existing managed-skill-files manifest already uses:

- AgentMux marks any `CLAUDE.md` it writes with a first-line comment marker. No file yet, or marker present → owned, freely regenerated exactly as before (**zero behavior change for the common case** — a fresh working directory).
- Marker absent on an existing file → foreign, never overwritten. AgentMux's composed content instead goes to an AgentMux-owned side file (`.claude/AGENTMUX_MEMORY.md`), pulled in via one idempotent `@import` line appended to the real file — offered at most once per working directory, and never silently reappears if a user deletes it to opt out.

The relative `@path` import mechanism was smoke-tested directly against the real Claude Code CLI (`claude -p`) before writing any code, not assumed — confirmed it resolves `@.claude/AGENTMUX_MEMORY.md` correctly.

Full design: `docs/specs/SPEC_CLAUDE_MD_OWNERSHIP_PROTECTION_2026_08_22.md`

## Test plan
- [x] 6 new unit tests in `agent_config.rs` covering: fresh dir, AgentMux-owned regeneration, foreign-file content never touched, side-file + import-line creation, no duplicate import across repeated launches, and a user's deliberate opt-out never gets silently reversed.
- [x] All 4 existing `write_agent_config_files_tests` integration tests still pass (skill cleanup, manifest path-traversal defense, MCP blob preservation).
- [x] `cargo check` clean.
- [x] Real Claude Code CLI smoke test confirming the `@path` import mechanism this design depends on actually works as assumed.

### DIFF
--- a/.changesets/1787436504-feat-agent-config-protect-a-pre-existing-project-claude-md-f-xdb8.md
+++ b/.changesets/1787436504-feat-agent-config-protect-a-pre-existing-project-claude-md-f-xdb8.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-config): protect a pre-existing project CLAUDE.md from being overwritten
--- a/agentmux-srv/src/backend/agent_config.rs
+++ b/agentmux-srv/src/backend/agent_config.rs
@@ -877,6 +877,120 @@ pub fn write_managed_skill_file_manifest(
     }
 }
 
+// ============================================================
+// CLAUDE.md ownership protection (I/O — see module doc comment)
+// ============================================================
+
+/// First line AgentMux writes on any `CLAUDE.md` it fully owns (freshly
+/// created, or regenerated on every launch since). Its ABSENCE on an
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

a5af:
@codex review

PR #2749, opened 2026-08-22
Korp@claudius: feat(muxspect): stamp every response with x-agentmux-srv-version, self-diagnosing stale-build 404s

GenericAgentX-asaf:
## Summary

Ext 5 of `docs/reports/REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22.md`. `SPEC_MUXSPECT_SRV_VERSION_HEADER_2026_08_22.md`.

Live during this same debugging session: `muxspect conversations` 404'd against the actual running instance because the running srv binary predated that route (merged to `main`, but this instance hadn't been rebuilt yet). Nothing said so — the 404 looked identical to "route genuinely doesn't exist."

Every response now carries `x-agentmux-srv-version` (a small `axum::middleware::from_fn` layer in `build_router`, applied to the whole router including health/webhook routes). `muxspect.mjs` prints it to **stderr** on every call (`[srv v0.55.19]` — never stdout, so `--json` output stays clean), and folds a hint into a 404's failure message specifically, since a version mismatch is the most plausible explanation for that one status.

## Test plan
- [x] 3 new cases for `logSrvVersion` (header present/absent, logs vs. doesn't) — 25/25 passing overall
- [ ] Live end-to-end verification (blocked on a rebuild, same as Ext 4 — see spec's Non-goals)

### DIFF
--- a/.changesets/1787437319-feat-muxspect-stamp-every-response-with-x-agentmux-srv-versi-njed.md
+++ b/.changesets/1787437319-feat-muxspect-stamp-every-response-with-x-agentmux-srv-versi-njed.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(muxspect): stamp every response with x-agentmux-srv-version, self-diagnosing stale-build 404s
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2750, opened 2026-08-22
Korp@claudius: feat(muxlog): swarm -d/--dispatch <id> filters + prints a correlated match-count verdict

GenericAgentX-asaf:
## Summary

Ext 6 (final) of `docs/reports/REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22.md`. `SPEC_MUXLOG_SWARM_DISPATCH_VERDICT_2026_08_22.md`.

The report's own motivating investigation needed to answer one question repeatedly: "did `subagent_watcher` ever log anything about this specific dispatch, on the instance I'm actually resolved to?" No single command answered that.

`-d`/`--dispatch <id>`:
- Filters on the **raw line text**, before/regardless of JSON parsing — a dispatch id can appear in the message OR as a bare structured-field value; `--grep` only ever matches the message field.
- Combined with `swarm`, prints an explicit **verdict** after the trace: `N lines mention '<id>'`, or a "0 matches" explanation with next steps. Silent zero-line output looks identical to "nothing to show" for every other filter; `-d` gets a verdict because that's exactly the symptom this flag exists to diagnose.
- `printLastLines` now returns the total match count (not just what fit under `-n`), so the verdict is exact.

## Scope (called out in the spec's Non-goals, not silently implied)

This closes the **log-correlation** third of the original 4-signal manual workflow (file written → watcher config → **log line** → registry state) — the filesystem-side check (subagent transcript files under `~/.claude`, outside `muxlog`'s own `~/.agentmux` domain) and a live-registry query (`muxspect`'s domain, needs new backend work) are real follow-ups, not attempted here.

## Test plan
- [x] 5 new cases for `renderLine`'s `--dispatch` filter (message match, field-only match, no match, composes with `--level`, unaffected when unset)
- [x] 2 new cases for `printLastLines`' return value (count exceeds `-n` cap; zero matches)
- [x] 34/34 passing in this file's full suite


a5af:
@codex review

Issue #2671, opened 2026-08-19
Tracking: Settings pane audit — sound defaults, mic/voice UI, and other undiscoverable settings


Agent2-asaf:
## Background

Started from a request to add a volume/gain control to Settings → Sounds. While there, did a broader audit of the Settings pane for real, already-shipped backend/frontend capability with no Settings UI at all — either `settings.json`-only (user has to hand-edit raw JSON) or a hardcoded constant with a real user-facing motivation for being configurable.

Full audit + ranked candidates: `docs/specs/SPEC_SETTINGS_AUDIT_GOOD_PICKINGS_2026_08_19.md`

## Done

- [x] Tool-call tones volume default raised `0.15` → `1.0` (master notification volume stays `0.6`, unchanged). Centralized the previously-scattered gain-default literals (5+ duplicated locations) into one source of truth, `frontend/app/notification/sound/sound-defaults.ts`. — [#2668](https://github.com/agentmuxai/agentmux/pull/2668) (merged)
  - Includes a Codex-caught follow-up: the embedded `settings-template.jsonc` example still showed the old `0.15`.
- [x] Added the missing "Waiting for input" tone enable toggle + volume slider to Settings → Sounds (`notify:sound:agent.waiting.for.input`, `notify:sounds:waiting:volume` — both were fully wired already, just had zero UI). — [#2669](https://github.com/agentmuxai/agentmux/pull/2669) (merged)
- [x] Made the `AskUserQuestion` auto-answer timeout user-configurable (`agent:askquestiontimeoutms`, Settings → Advanced) — was a hardcoded 30s constant explicitly flagged as a spec-anticipated follow-up. — [#2670](https://github.com/agentmuxai/agentmux/pull/2670) (merged)
  - Includes a real bug both reagent (P1) and Codex (P2) independently caught before merge: reading the setting inside a `tool_use_id`-keyed effect made that effect also re-run (wiping in-progress answers) on unrelated setting changes. Fixed via `untrack()`, baked into the shared helper so no future call site can reintroduce it.

## Remaining (ranked by the audit spec)

[...]

PR #2751, opened 2026-08-22
feat(settings): add Recording/Input section — voice engine config, device picker, test-your-mic

GenericAgentX-asaf:
## Summary

Candidate #1 from #2671 / `SPEC_SETTINGS_AUDIT_GOOD_PICKINGS_2026_08_19.md` — the single highest-value gap found in that audit. Voice input is a fully-shipped feature (3 STT engines, per-pane mic button) with **zero Settings UI**, including a Groq API key a user had to hand-edit into `settings.json` in plaintext. Implements `docs/specs/SPEC_SETTINGS_RECORDING_INPUT_SECTION_2026_08_19.md`.

## What's new

New **"Recording"** rail section (between Sounds and Advanced):

- Enable toggle, engine picker (Groq / whisper.cpp local / Web Speech dev-only)
- **Groq API key** via a new `MaskedKeyField` shared primitive — masked-by-default, Replace/Save/Cancel, egress-transparency note. Built generic (not voice-specific) per the audit spec's Open Question 1, so messaging-bridge bot tokens can reuse it later.
- **whisper-local config** (CLI path + model) via a new `PathField`, live-validated against a new `voice.checkPath` RPC (`agentmux-srv/src/server/app_api/voice.rs`) — exposes the exact same `Path::exists()` check `voice.rs` already does at transcribe time, just proactively instead of only failing at first use.
- **Microphone device picker** (`enumerateDevices`), new `voice:inputDeviceId` setting threaded into the real capture path.
- **"Test your microphone"** — the first "test X" pattern in the app: a live level meter (new `useMicLevelMeter` hook, built on a `computeRms` helper factored out of the existing VAD so the two don't duplicate Web Audio boilerplate) + a real transcription round-trip against `/api/v1/voice/transcribe`, surfacing the server's actual error text.
- Fixed the stale "Speech recognition isn't available in this build yet" toast copy.
- Threaded a `lastErrorDetail` signal through `VoiceSession` (additive/optional — `MicButton.tsx`'s existing tooltip logic is unchanged) so the test-mic panel can show the real server error.

## Non-goals (matching the spec)

- Quick mic test only validates capture + Groq — whisper-local's WAV-encoding path isn't duplicated in the test flow (documented in-UI); full validation there means actually recording from an agent pane.
- No auto-detection of a local whisper.cpp install.
- No change to the groq-vs-whisper-local default.

## Test plan

- [x] `npx vitest run` — full suite green (one pre-existing flaky timeout in `tool-renderers/registry.test.ts`, confirmed passing in isolation, unrelated to this change)
- [x] New `masked-key-field.test.tsx` (6 tests, the state machine: locked/entry/save/cancel)
- [x] Updated `settings-view.test.tsx` for the new rail entry

### DIFF
--- a/.changesets/1787438407-feat-settings-add-recording-input-section-voice-engine-confi-9i5s.md
+++ b/.changesets/1787438407-feat-settings-add-recording-input-section-voice-engine-confi-9i5s.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(settings): add Recording/Input section — voice engine config, device picker, test-your-mic
--- a/.changesets/1787439491-fix-voice-redact-groqapikey-from-renderer-fix-engine-switch--0dxc.md
+++ b/.changesets/1787439491-fix-voice-redact-groqapikey-from-renderer-fix-engine-switch--0dxc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(voice): redact groqApiKey from renderer, fix engine-switch/test-cancel bugs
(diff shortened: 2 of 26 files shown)


a5af:
@codex review

PR #2752, opened 2026-08-22
Korp@claudius: feat(muxlog): ls gains a LIVE column via real health-checked liveness probe

GenericAgentX-asaf:
## Summary

Ext 3 of `docs/reports/REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22.md`. Reopens the prior PR (#2742), which was closed unmerged after its stacked base branch (Ext 2, already merged as #2741) was deleted — this is the same, already-reviewed content rebased cleanly onto current `main`, with no further changes.

`muxlog ls` previously inferred an instance's liveness from log mtime alone — silent and stale under any crash, and gave no signal at all when deciding whether an instance is worth targeting with `-i`. This adds a **LIVE** column (`live`/`dead`/`?`) computed from a real check: probing each instance's `ipc-port-*` file(s) with an actual `GET /health` request against `agentmux-cef`'s IPC server.

### Review history carried over from #2742 (already fixed, included in this branch)
- **[P1]** `checkLiveness` now probes **every** `ipc-port-*` file concurrently, not just the first via `entries.find()` — a dev-mode data dir is keyed by branch (not version), and a crashed process's port file isn't cleaned up until a graceful exit, so a stale dead-port file and a live one can coexist; "live" if any probe succeeds.
- **[P2]** `probePort` validates the real `GET /health` response body (`{status:"ok"}`), not just a bare TCP connect — a bare connect can't distinguish AgentMux's IPC listener from an unrelated service that got OS-reassigned the same now-free ephemeral port.
- **[P2]** `docs/MUXLOG.md`'s `muxlog ls` column documentation now lists the LIVE column.

## Test plan
- [x] 36/36 passing in `muxlog.test.mjs` (real `http.createServer`-backed liveness probe tests, `// @vitest-environment node`)
- [x] Diff vs current `main` is exactly Ext 3's two commits — verified via `git diff origin/main --stat` before pushing

### DIFF
--- a/.changesets/1787434874-feat-muxlog-ls-gains-a-live-column-via-real-tcp-liveness-pro-huv1.md
+++ b/.changesets/1787434874-feat-muxlog-ls-gains-a-live-column-via-real-tcp-liveness-pro-huv1.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxlog): ls gains a LIVE column via real TCP liveness probe of each instance's ipc-port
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2748, opened 2026-08-22
feat(settings): add agent watchdog threshold controls to Terminal section

GenericAgentX-asaf:
## Summary

Part of tracked work from #2671 (candidate #2, `SPEC_SETTINGS_AUDIT_GOOD_PICKINGS_2026_08_19.md`). `term:agentmaxruntimehours` / `term:agentidletimeoutmins` (`agentmux-srv/src/backend/wconfig/types.rs:91-99`) are a real, live safety feature — `watchdog.rs`'s 60s poll loop kills any agent pane exceeding a configured max wall-clock runtime or PTY-idle duration — but had zero discoverability.

**Re-verified before implementing**: `SPEC_SETTINGS_PANE_COMPLETION_2026_07_14.md` previously claimed these fields were removed entirely, and the audit spec flagged this as needing a fresh check. Confirmed both fields are live in `types.rs` and consumed in `watchdog.rs` today.

## Changes

- `schema/settings.json` + `frontend/types/gotypes.d.ts`: added the two previously-undocumented keys (matching the format of neighboring `term:*` entries)
- Two new numeric `SettingRow`s in Settings → Terminal, "0 = no limit" per the existing doc comments — no backend changes, both fields already round-trip through the generic settings merge path

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] Manual: confirmed field names/defaults exactly match `types.rs`'s serde rename attributes

PR #2744, opened 2026-08-22
feat(settings): add drag-and-drop file-attach controls to Advanced section

GenericAgentX-asaf:
## Summary

Part of tracked work from #2671 (candidate #4, `SPEC_SETTINGS_AUDIT_GOOD_PICKINGS_2026_08_19.md`). `dnd:enabled`, `dnd:agentinserttoken`, `dnd:concurrency` (`schema/settings.json:281-295`) are fully shipped and consumed by `useAgentDropAttach.ts`, but had zero Settings UI — only configurable by hand-editing `settings.json`.

## Changes

- New "Drag & drop" subsection in Settings → Advanced:
  - Enable file drop (toggle)
  - Insert reference token (toggle)
  - Max concurrent uploads (numeric, blank = unlimited)
- `dnd:concurrency` has no default (absent = unlimited) — clearing the field calls `set(key, null)`, the existing sanctioned deletion mechanism already built into `setconfig`'s merge logic (`config_watcher_fs.rs`'s "Remove keys explicitly set to null" path). No backend changes needed.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] Manual: verified against the schema defaults and `useAgentDropAttach.ts`'s read sites

PR #2753, opened 2026-08-22
docs(spec): design proposal for messaging bridges Settings section

GenericAgentX-asaf:
## Summary

Follow-up spec for candidate #3 from #2671 / `SPEC_SETTINGS_AUDIT_GOOD_PICKINGS_2026_08_19.md`, which explicitly deferred it: *"the most product-shaped gap in the whole audit... left as a follow-up spec of its own given the scope."*

Discord/Telegram/Slack/WhatsApp are all fully wired server-side (~24 settings fields) with zero Settings UI and zero `schema/settings.json` entries — only configurable by hand-editing raw JSON containing bot tokens and app secrets in plaintext.

**Design-only, no code.** New "Integrations" rail section, one collapsible sub-block per bridge gated by its own enabled toggle, reusing the `MaskedKeyField` primitive #2751 introduces (designed generic for exactly this reuse). Flags three open questions before implementation: the section label ("Integrations" vs "Messaging"), which existing agent-picker component to reuse for "target agent" fields, and whether these bridges hot-reload on settings change or need a "restart required" note.

Candidates #1/#2/#4 from the same audit have already shipped (#2751, #2748, #2744) — this is the last item, delivered as a spec rather than code per the audit's own sequencing recommendation.

### DIFF
--- a/docs/specs/SPEC_SETTINGS_MESSAGING_BRIDGES_SECTION_2026_08_22.md
+++ b/docs/specs/SPEC_SETTINGS_MESSAGING_BRIDGES_SECTION_2026_08_22.md
@@ -5,9 +5,15 @@
 **Status:** Draft
 **Scope:** New `frontend/app/view/settings/sections/integrations-section.tsx` (+
 registration in `settings-view.tsx`/`settings-model.ts`), reusing the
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2754, opened 2026-08-22
fix(agent): stop "Agent unresponsive" from firing during legitimate context compaction

Agent1-asaf:
## Summary

`HealthMonitor`'s silence-based Unresponsive/Restart detector fires after 120s with no output on an active turn. Claude Code context compaction is a single opaque LLM call producing zero intermediate output — the repo's own compaction spec captured a real example taking 231.6s, nearly double that threshold — so a normal auto-compaction routinely trips the detector while the pane is simultaneously showing "Compacting…", then self-heals ~90s later once the real `compact_boundary` frame lands. The two features (unresponsive detection, 07-27; compaction visibility, 07-31) were built three weeks apart and never made aware of each other.

Full root-cause analysis and design: `docs/specs/SPEC_UNRESPONSIVE_FALSE_POSITIVE_DURING_COMPACTION_2026_08_22.md`.

## Changes

- `HealthMonitor` gains a `compacting` flag that suspends the normal 30s/120s Stalled/Dead thresholds while true, bounded by a separate, much more generous 600s ceiling — so a genuine hang mid-compaction is still eventually caught, not permanently hidden.
- `set_compacting(false)` deliberately does NOT reset `last_meaningful_ts` — resuming re-arms the normal thresholds against whatever silence has already elapsed, not a fresh grace period.
- New `Controller::health_monitor()` trait accessor (default `None`, overridden by the three controller types that own a `HealthMonitor`) + a small special case in `handle_wps_publish` forwards the existing `compaction_started` WPS event (already published by the `PreCompact` hook) into the target block's health monitor.
- New `is_compact_boundary_frame()` check at each of the three controller stdout-reader loops that already classify raw NDJSON lines — clears `compacting` the instant the real `compact_boundary` frame arrives.
- `ACP` controllers are deliberately untouched — not the Claude Code stream-json path the compaction feature targets, matching the original compaction spec's own provider scoping.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2689 passed, 0 failed (one pre-existing, unrelated, already-tracked flaky handle-count test on this machine skipped — `docs/status/STATUS_SRV_SECTION_HANDLE_LEAK_LIVE_RECURRENCE_2026_08_19.md`)
- [x] New unit tests in `health.rs`: compacting suppresses Stalled/Dead despite stale output; the 600s ceiling still reaches Dead; exactly-under-ceiling stays Healthy; non-compacting behavior is byte-for-byte unchanged; `set_compacting(false)` grants no fresh grace period; a Dead reached via the ceiling still publishes a real `AgentFailure`
- [x] New integration tests in `server/tests.rs`: a `compaction_started` publish reaches the target block's `HealthMonitor`; an unregistered block id is a harmless no-op, never an HTTP error

### DIFF
--- a/.changesets/1787439562-fix-agent-stop-agent-unresponsive-from-firing-during-legitim-wokj.md
+++ b/.changesets/1787439562-fix-agent-stop-agent-unresponsive-from-firing-during-legitim-wokj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): stop 'Agent unresponsive' from firing during legitimate context compaction
--- a/agentmux-srv/src/backend/blockcontroller/health.rs
+++ b/agentmux-srv/src/backend/blockcontroller/health.rs
@@ -210,6 +210,18 @@ impl HealthMonitor {
         let now = Instant::now();
         inner.last_output_ts = now;
         inner.last_meaningful_ts = now;
+        // codex P2 on PR #2754: reset unconditionally (both directions), same
+        // reasoning as last_output_ts/last_meaningful_ts above — a turn
+        // boundary must never carry a stale `compacting=true` forward. Without
+        // this, a process that crashes/is cancelled mid-compaction (no
+        // compact_boundary frame ever arrives to clear it via
+        // is_compact_boundary_frame's own call site) leaves `compacting`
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #2755, opened 2026-08-22
Camper@claudius: fix(agent): make cross-channel resume session_id a live pointer, not write-once

GenericAgentX-asaf:
## Summary

Root-caused after the agent's own pane resumed a 9-day-stale session on
reopen in a fresh channel (see the retro this PR adds:
`docs/retro/retro-agent-resumed-9-day-stale-session-2026-08-22.md`).

This ties directly to an already-documented, only-partially-fixed bug:
`docs/status/STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md`
found that the shared cross-channel registry's `session_id` is written
**once** (`session_backfill::backfill_session_ids`, explicitly idempotent —
skips any record that already carries a non-empty value, forever) and never
refreshed after that, even though the *local* per-channel
`db_agent_instances.session_id` stays genuinely live.

PR #2693/#2699 partially addressed this — but only for the case where a
stale `--resume <sid>` is **confirmed unreachable** (the CLI rejects it
outright), falling back to a largest-on-disk-session scan. This incident is
the harder, still-open case: the stale registry id pointed at a session
file that **still exists and is still valid** — so the CLI resume
**succeeds**, silently, with no WARN, no retry, and no fallback ever
triggered. The pane simply, quietly resumes the wrong (stale-but-valid)
conversation.

## Fix (in progress on this branch)

Implementing status doc recommendation #1: make the shared registry
`session_id` a live pointer instead of a write-once value, by propagating
`persist_session_id`'s already-live local writes to the shared
cross-channel registry record on every genuine capture — not just the
one-time backfill.

Opening this as a draft now per the plan (retro first, fix in the same PR);
will push the implementation as follow-up commits.

### DIFF
--- a/.changesets/1787441148-fix-agent-propagate-resumed-session-id-to-shared-cross-chann-x7xr.md
+++ b/.changesets/1787441148-fix-agent-propagate-resumed-session-id-to-shared-cross-chann-x7xr.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): propagate resumed session_id to shared cross-channel registry
--- a/agentmux-srv/src/backend/storage/agents.rs
+++ b/agentmux-srv/src/backend/storage/agents.rs
@@ -1643,6 +1643,7 @@ impl Store {
             )?;
         }
         self.registry_upsert_if_named(inst);
+        self.registry_propagate_continuation_session_id(inst);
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2756, opened 2026-08-22
fix(tabs): quick-fork always inherits identity, dropping the redundant second menu item

AgentX-asaf:
## Summary
Repo-owner feedback on the merged quick-fork feature: the "unbound by default, explicit inherit-identity opt-in" design (spec §5 / Phase 4, two menu items: "Quick-fork" and "Quick-fork (inherit identity)") was based on a wrong premise.

Verified directly against the backend: `LinkAgentIdentityCommand` (`agentmux-srv/src/server/agent_handlers/identity.rs:559-580`) never copies a credential — it only inserts a `db_agent_identity_links` row (`agent_id`, `account_id`, `provider`). The actual credential (OAuth config dir or keychain entry) lives at a path keyed by `account_id` alone (`identities_dir().join(account_id)`), so *any* agent linked to that account — the fork's own new link or the parent's pre-existing one — reads and writes the exact same shared credential. There's no independent, safer copy that "unbound by default" was actually withholding; "inherit" and "unbound" were never a real risk/no-risk pair.

Given that, offering both as separate menu actions added a decision with no safety difference on the other side of it.

- `quickForkAgent` now always resolves and inherits the source's own bound account, unconditionally — removed the `opts` parameter entirely.
- `getBodyContextMenuItems` now has one "Quick-fork" entry, not two.
- Spec doc (`SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md` §5, phasing table) updated with a correction notice; original reasoning kept for the historical record.

## Test plan
- [x] `npx vitest run` — full suite green (2949 passed), 1 pre-existing/unrelated flaky timeout in `tool-renderers/registry.test.ts`
- [x] `quick-fork.test.ts` updated: removed the opt-out test, renamed the describe block to "identity inheritance", added a "no linked identity at all" case

### DIFF
--- a/.changesets/1787441538-fix-tabs-quick-fork-always-inherits-identity-dropping-the-re-q6iq.md
+++ b/.changesets/1787441538-fix-tabs-quick-fork-always-inherits-identity-dropping-the-re-q6iq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tabs): quick-fork always inherits identity, dropping the redundant second menu item
--- a/docs/specs/SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md
+++ b/docs/specs/SPEC_AGENT_QUICK_FORK_NEW_TAB_2026_08_21.md
@@ -324,12 +324,32 @@ so no LAN/WAN qualifier is ever shown — that only becomes relevant if a
 future capability forks directly onto a remote peer, at which point the
 naming spec's tiering (§4.1 there) already covers it without change here.
 
-## 5. Identity: resolving the open question prior specs left unsettled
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentX-asaf:
Good catch — verified against the backend's OAuth layer-3 spawn gate (`identity/resolver/inject.rs:358-367`): a definition with no resolved binding at all gets blocked there unless it opted into ambient login, so silently proceeding unbound after a failed lookup would let `launchAgentDefinition` report success while the agent can never actually spawn. Fixed in b6cd0b1ae — the lookup failure now propagates to the outer catch, which already handles cleanup + notification identically to every other mid-flight failure.

Also renamed the menu label/notification titles to "Quick Fork" per repo-owner request in the same commit.

PR #2757, opened 2026-08-22
Korp@claudius: fix(subagent-watcher): watch the identity-bound Claude config dir, not a stale spawn-time snapshot

GenericAgentX-asaf:
## Summary

Root-causes the original "dispatched subagent never appears in Swarm" bug left open in `docs/reports/REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22.md` §4, root-caused this session using that report's own newly-shipped tooling (`muxlog swarm -d`, `muxlog ls`'s LIVE column — Ext 6/3, PRs #2750/#2752).

**Root cause:** `subagent_watcher::resolve_claude_config_dir` trusted a block's persisted `cmd:env.CLAUDE_CONFIG_DIR` — a write-once snapshot taken at launch, using the *generic* shared-provider dir (`~/.agentmux/shared/providers/claude`). For any agent bound to an explicit Armory identity, the real CLI process runs (and writes subagent transcripts) under a completely different, identity-bound dir — re-resolved fresh on **every turn** by `inject_identity_env_with_broker`, per `SPEC_PROVIDER_ISOLATION_2026_06_20.md` §4.3, which explicitly documents that `cmd:env` goes stale and was never meant to be trusted after launch. `subagent_watcher` was added later and never joined that re-resolution — it watches the wrong directory forever, deterministically, for every identity-bound agent, not a race or rare edge case.

**Confirmed live**, on my own currently-running agent: a real `Agent`-tool dispatch's transcript landed under `.../shared/identities/<uuid>/claude/projects/.../subagents/agent-<id>.jsonl`; the watcher's own "watching for subagent JSONL files" log line fired once, at registration, against `.../shared/providers/claude/projects` — and never fired again for that dispatch. This supersedes the original report's leading suspect (`session_belongs_to_block`) — that gate is downstream of the file watch and never runs, because the watch itself never sees the write.

## Fix

Added `resolve_bound_oauth_config_dir` (`identity/resolver/inject.rs`) — a new, deliberately **pure, read-only** helper. It does **not** reuse `inject_identity_env`/`inject_identity_env_with_broker` directly: those also resolve API-key secrets into the caller's env map (a leak risk for a caller that only wants a directory path), run a token-expiry filesystem probe, upsert account status to the DB, and publish `identityaccounts:changed` — none of which belong on every agent *registration*, only on an actual spawn/turn. The new helper does only the minimal OAuth-branch lookup and returns `None` (never an error) on any failure.

`resolve_claude_config_dir` gained an optional first-choice `bound_dir` parameter; falls back to the existing `cmd:env`/`derive_claude_config_dir` chain unchanged when `None` — **purely additive**, ambient/unbound agents keep their exact old behavior.

## Test plan
- [x] 5 new unit tests for `resolve_bound_oauth_config_dir`: resolves for an identity-bound agent; `None` when no instance/no binding/API-key-class provider; and an explicit assertion that account status is never mutated (proves the "no side effects" claim, not just states it)
- [x] `resolve_claude_config_dir` tests updated for the new parameter + 1 new test proving the bound dir wins over stale `cmd:env`
- [x] Full `agentmux-srv` suite: 2706 passed, 0 failed, 6 pre-existing ignores

### DIFF
--- a/.changesets/1787442317-fix-subagent-watcher-watch-the-identity-bound-claude-config--py2b.md
+++ b/.changesets/1787442317-fix-subagent-watcher-watch-the-identity-bound-claude-config--py2b.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(subagent-watcher): watch the identity-bound Claude config dir, not a stale spawn-time snapshot
--- a/agentmux-srv/src/backend/subagent_watcher/parse.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/parse.rs
@@ -473,13 +473,27 @@ pub fn derive_claude_config_dir(agent_id: &str) -> Option<PathBuf> {
     Some(config_dir)
 }
 
-/// The authoritative Claude Code config directory for a block: the
-/// `CLAUDE_CONFIG_DIR` the CLI process was actually launched with, read from
-/// the block's own `cmd:env` meta (written by the launch flow before spawn,
-/// from the exact same resolution the CLI process itself used — see
-/// `agentmux-cef`'s `ensure_auth_dir`). Falls back to
+/// The authoritative Claude Code config directory for a block.
+///
+/// `bound_dir` — pass `identity::resolver::resolve_bound_oauth_config_dir`'s
+/// result here. It wins when present: for an agent bound to an explicit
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2759, opened 2026-08-23
Korp@claudius: feat(fleet): FleetBroadcast reaches cross-channel/LAN/WAN targets, not just this instance

GenericAgentX-asaf:
## Summary

Item §3.1 of `docs/reports/REPORT_CROSS_INSTANCE_CONTROL_ROBUSTNESS_AUDIT_2026_08_22.md`.

`FleetList`/`DiscoverAgents` correctly return `host.cross_channel`, `lan`, and `wan.subscribed_agents` entries. `FleetBroadcast`'s own doc comment says "get targets from FleetList" — but its `block_id -> agent_name` resolution only ever read `host.addressable`, so a `host.cross_channel` target (a different channel on this same host — genuinely has a real `block_id`, just under `name` instead of `agent_id`) silently failed with "no registered agent for this block." LAN/WAN targets have no `block_id` at all (they're not local blocks — only agent names), so they never had a way to resolve either.

**Fix:**
- `host.cross_channel` folded into the same resolution map `host.addressable` builds (extracted into a new pure `build_block_to_agent_map`, unit-tested without a live discovery endpoint).
- A target that doesn't match any known block_id now falls through to being used **as a literal agent name** — letting `/agentmux/reactive/inject`'s own cross-tier cascade (cross-channel → LAN → WAN muxbus relay, already proven by `SendMessage`) reach it. A genuinely-wrong target still fails cleanly, just via the inject endpoint's own "agent not found," one layer later than before.
- Tool description + validation error message updated to document the new addressing (block_id for host/cross-channel, agent name for LAN/WAN).

**Deliberately not touched:** `fleet_broadcast_impl` (`agentmux-srv`, the WS-RPC/Swarm-UI path) — per its own doc comment it's intentionally host-only, since a human clicking a UI button has no agent's `AGENTMUX_JEKT_KEY` to sign a LAN/WAN-bound jekt with. Only the MCP tool's client-side path (which does hold the calling agent's key) can extend cross-tier.

## Test plan
- [x] 5 new unit tests for `build_block_to_agent_map`: resolves `addressable`; resolves `cross_channel`; merges both; ignores `lan`/`wan`'s differently-shaped entries without panicking; handles a missing/empty response
- [x] Full `agentmux-mcp` suite: 13 passed, 0 failed

### DIFF
--- a/.changesets/1787446160-feat-fleet-fleetbroadcast-reaches-cross-channel-lan-wan-targ-r06t.md
+++ b/.changesets/1787446160-feat-fleet-fleetbroadcast-reaches-cross-channel-lan-wan-targ-r06t.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(fleet): FleetBroadcast reaches cross-channel/LAN/WAN targets, not just this instance
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -297,11 +297,11 @@ const FLEET_LIST_TOOL: &str = r#"{
 
 const FLEET_BROADCAST_TOOL: &str = r#"{
   "name": "FleetBroadcast",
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2761, opened 2026-08-23
feat(layout): generalize the tab-content reveal gate to leaf/pane scope

AgentX-asaf:
## Summary
Repo owner reported visible UI flashing — "even when clicking the + to get a new agent." Root cause, confirmed against real code (full writeup in `docs/specs/SPEC_PANE_BLOCK_STACK_MOUNT_FLICKER_2026_08_22.md`): every block-stack push/switch (the pane tab strip's "+", Quick Fork, Agent History) forces a full component-subtree remount by design (`layoutStack.ts`'s own doc comment — `NodeModel.blockId` is captured once at construction). Nothing hid that transition. AgentMux already solved this exact class of flicker once, for whole-tab switches (`SPEC_TAB_CONTENT_REVEAL_GATE.md`'s "reveal gate") — but that spec's own "Out of scope" section explicitly flagged this gap: "backend-driven switches... bypass `setActiveTab` and don't currently fire the gate... address if it becomes a pain point." It has.

**Fix**: generalize the reveal gate to leaf/pane scope, alongside (not replacing) the existing whole-tab gate:
- New `gatingNodeIds()` / `holdLeafRevealGate(nodeId)` / `scheduleLeafRevealLift(nodeId)` in `tab-reveal.ts`, built on `@/app/util/settle-detector`'s `scheduleOnSettle` — that module's own doc comment had already anticipated exactly this need ("wrong for N agent panes each waiting on their own settle, since concurrent callers would clobber each other's detector"), pre-built as a per-instance-cancellable primitive for it.
- Wired into all three callers that force this remount: `handleNewAgentTab` (the "+" button), `openOrFocusHistoryTab` (both its create-new and switch-to-existing paths), `quickForkAgent` — each holds the gate before its first RPC and schedules the lift in a `finally` covering every exit path.
- All three platform `TileLayout.*.tsx` files read `gatingNodeIds()` per leaf, merged into the existing tile-transform style object (not a wrapper element, so each leaf's absolute positioning is untouched) — mirrors `workspace.tsx`'s identical visibility+opacity treatment for the whole-tab gate.

The existing whole-tab gate's own behavior and all 6 of its existing tests are untouched — the leaf gate is fully separate, deliberately not unified into one data structure.

Phases 1-2 of the spec (this PR) are the actual fix. Phases 3-4 (cross-fading the intermediate loading/picker stages so the *revealed* frame is always the final one, and a CSS transition on the tab strip's own width) are smaller, independent follow-ups not included here — noted as still-open in the spec.

## Test plan
- [x] `npx vitest run` — full suite green (2984 passed), 1 pre-existing/unrelated flaky timeout in `tool-renderers/registry.test.ts` (passes standalone)
- [x] `tab-reveal.test.ts`: 7 new tests for the leaf gate (per-node independence, independence from the whole-tab gate) — all 6 existing whole-tab tests still pass unchanged
- [x] `quick-fork.test.ts` / `open-history-tab.test.ts`: new tests asserting hold/schedule fire on the correct node id across every exit path (success, RPC rejection, launch failure, launch rejection, pane-closed-mid-flight)
- [ ] Manual verification by repo owner (click "+", Quick Fork, Agent History — no visible flash/flicker)


a5af:
@codex review

AgentX-asaf:
All three findings are real — thanks for the thorough pass. Fixed in 7a56cebce:

1. **Ordinary block-stack switches were ungated** — `handleTabSwitch` (agent-view.tsx) and the identical `handleTermTabSwitch` (term.tsx) call `setActiveBlockInStack` directly, which forces the same remount as creating a new stack member. Both now gated. Also found and fixed the same create-path gap in term.tsx's own `handleTermTabAdd` (the terminal analog of `handleNewAgentTab`, missed in the original PR).
2. **Race: older operation's completion could reveal mid-newer-operation** — fixed via a generation-token pairing: `holdLeafRevealGate` returns a token, `scheduleLeafRevealLift` takes it back, and a call is a no-op if a newer hold has since taken over.
3. **Race: slow operation's hold-timeout then re-hides on schedule** — same generation-token mechanism also tracks "already resolved once," so a stale schedule call never re-hides an already-revealed pane.

Two new regression tests in `tab-reveal.test.ts` cover both races directly.

AgentX-asaf:
Fixed in face3c249 — added \`clearLeafRevealGate(nodeId)\`, wired into \`closeNode\` (\`layoutMagnify.ts\`), the one place every close path (block-stack pop-to-last-member, ordinary pane close, drag tear-off) funnels through when a node is actually removed from the tree. Cancels any pending timer/detector first so a stale callback can't resurrect gating for an id that's already gone. 3 new tests cover it.

AgentX-asaf:
Fixed in f29eef609 — thanks for catching this.

Both `handleTabClose` (agent-view.tsx:372) and `handleTermTabClose` (term.tsx:147) now hold the leaf reveal gate before calling `closeBlockInStack` and schedule the lift in a `.finally()`, matching the pattern already used for the switch handlers. Confirmed `closeBlockInStack` does unconditionally evict the NodeModel on every pop, but the `<Key>` wrapper only actually remounts when `activeKeyFor` (`${node.id}:${node.data.activeBlockId}`) changes — i.e. only when the closed tab was the active member — so gating unconditionally is a harmless no-op for closing a background tab and the real fix for closing the active one.

The last-member-closes-the-pane path already goes through `closeNode`, which already calls `clearLeafRevealGate` from Phase 2's cleanup fix, so no extra handling was needed there.

`tsc --noEmit` and the full reveal-gate test suite (54 tests) pass. Spec doc's Phase 2 row updated with this third round.

AgentX-asaf:
Good catch — fixed in d44391083.

You're right that `gatingNodeIds()` hides the leaf's entire rendered content regardless of whether a remount is actually pending, so gating unconditionally on every close was a genuine regression for closing a background tab. Both `handleTabClose` and `handleTermTabClose` now only gate when the closed tab is the resolved node's own active member (`node.data?.activeBlockId === targetBlockId`), matching the guard `handleTabSwitch`/`handleTermTabSwitch` already had. Closing a non-active tab now skips the gate entirely, going straight to `closeBlockInStack` with no hide/reveal cycle.

`tsc --noEmit` and the full reveal-gate suite (54 tests) still pass. Spec doc updated again.

PR #2762, opened 2026-08-23
Korp@claudius: feat(fleet): FleetBulkStop reaches cross-channel targets, not just this instance

GenericAgentX-asaf:
## Summary

Item §3.2 of `docs/reports/REPORT_CROSS_INSTANCE_CONTROL_ROBUSTNESS_AUDIT_2026_08_22.md`, sibling fix to #2759 (FleetBroadcast).

`fleet_bulk_stop_impl` resolved every target purely via this instance's own in-process controller registry. A `host.cross_channel` target (a different channel on the same host, real `block_id`, visible via `FleetList`) always failed with "no registered agent for this block" — even though the block is live, just owned by a sibling process on the same machine.

**Fix**, at the single shared choke point (`fleet_bulk_stop_impl` — unlike broadcast, bulk-stop involves no jekt signing at all, so there's no signing-locality reason to keep this client-side; this covers both the WS-RPC/Swarm-UI and HTTP/MCP-tool callers identically):
- When a target isn't in the local registry, check the shared cross-channel registry (same mechanism `server/reactive.rs`'s inject cascade tier-2b already uses) and forward an authenticated stop request to that channel's own srv (new `POST /agentmux/agent/stop` endpoint), using **its own** `auth_key` from the registry — same one-hop, same-host, same-user trust model as the existing inject forward, no new trust primitive.
- Falls through to the original `NOT_RUNNING` error when nothing matches anywhere — no regression for the common case.
- `fleet_bulk_stop_impl` is now `async`; both call sites updated to `.await` it.

**LAN/WAN deliberately deferred, not silently dropped** (see the spec's §3): per the audit, no remote-command bus exists on any tier today — jekt is message-only. Reaching LAN/WAN for bulk-stop means designing a genuinely new authenticated destructive-action protocol, not reusing something that already exists the way cross-channel does. WAN specifically is blocked on the same Cognito redesign as general WAN jekt signing (real `agentmux-cloud`/`shared-infrastructure` work, out of this repo's scope). LAN is architecturally feasible via the existing LAN Ed25519 signing, but a new message type's tier/escalation semantics (should a verified LAN sender's stop request skip human escalation, or does a destructive remote action always warrant it regardless of verification?) deserves its own dedicated design pass, not a rushed addition here.

## Test plan
- [x] 4 new unit tests for `forward_stop_to_shared_channel`: no matching entry → `None`; non-loopback entry ignored (security); stale self-entry ignored; succeeds against a real loopback peer (raw-TCP fake responder, mirrors the existing `spawn_fake_browser_api` pattern) and asserts the **peer's own** `auth_key` is sent, not this instance's; propagates a peer-side failure's error text
- [x] 1 new end-to-end test: a target present only in the shared cross-channel registry succeeds via the forward
- [x] Full `agentmux-srv` suite: 2721/2722 passed — the one failure (`refresh_processes_specifics_does_not_leak_a_handle_per_call`) is an unrelated, environment-sensitive Windows handle-count check, confirmed to pass in isolation and untouched by this PR

### DIFF
--- a/.changesets/1787447234-feat-fleet-fleetbulkstop-reaches-cross-channel-targets-not-j-2dg8.md
+++ b/.changesets/1787447234-feat-fleet-fleetbulkstop-reaches-cross-channel-targets-not-j-2dg8.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(fleet): FleetBulkStop reaches cross-channel targets, not just this instance
--- a/agentmux-srv/src/server/app_api/fleet.rs
+++ b/agentmux-srv/src/server/app_api/fleet.rs
@@ -135,7 +135,7 @@ fn register_fleet_bulk_stop(engine: &Arc<WshRpcEngine>, state: &AppState) {
             Box::pin(async move {
                 let cmd: CommandFleetBulkStopData = serde_json::from_value(data)
                     .map_err(|e| format!("fleet.bulk-stop: {e}"))?;
-                let result = fleet_bulk_stop_impl(&state, cmd.targets, cmd.signal.as_deref(), cmd.staged);
+                let result = fleet_bulk_stop_impl(&state, cmd.targets, cmd.signal.as_deref(), cmd.staged).await;
                 Ok(Some(serde_json::to_value(&result).unwrap()))
             })
         }),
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2764, opened 2026-08-23
Korp@claudius: feat(muxspect): Phase B policy infrastructure + jekt tier enforcement for transcript_request

GenericAgentX-asaf:
## Summary

Implements the confirmed jekt rules (`SPEC_JEKT_TRANSCRIPT_REQUEST_TIER_RULES_2026_08_22.md`) for `SPEC_MUXSPECT_CROSS_TIER_CONVERSATION_VISIBILITY_2026_08_21.md`'s Phase B (LAN conversation visibility). Full writeup in `docs/specs/SPEC_MUXSPECT_PHASE_B_POLICY_AND_TIER_ENFORCEMENT_2026_08_22.md`.

**Ships (schema v26):**
- `conversation_visibility` per-agent setting (`private`/`trusted_peers`/`ask`), dual-written to `db_agents`, deliberately channel-local (not added to the cross-channel registry wire format — same precedent as `model_vendor_base_url`/`memory_id`).
- `db_conversation_trust_grants` table + CRUD module, mirroring `db_lan_peer_pubkey_pins`'s exact shape (the explicit precedent the design spec named).
- `transcript_request`/`transcript_response` wire payload (`agentmux-common`) — JSON-encoded inside a jekt's existing `message` field. No new transport: there's no structured `msg_type` anywhere in the jekt wire format (`message` is part of the signed material every signer/verifier depends on), so this rides the exact same `/agentmux/reactive/inject` signing + cross-tier delivery cascade every ordinary jekt already uses.
- **The actual `TIER=sensitive`/`ESCALATE=required` computation** — the security-critical reason this needed a repo-owner conversation in the first place: two new server-computed `InjectionRequest` fields (`#[serde(skip_deserializing)]`, same guarantee as `sig_verified`/`lan_verified` — no attacker-supplied body can set or suppress them), resolved in `handle_reactive_inject` (has `Store` access, unlike `Handler`) and wired into `Handler::inject_message_inner`'s existing tier logic.

[...]

PR #2765, opened 2026-08-23
Korp@claudius: fix(muxspect): wire transcript_request jekt tier enforcement into the WAN delivery path (Phase C)

GenericAgentX-asaf:
## Summary

Phase C of `SPEC_MUXSPECT_CROSS_TIER_CONVERSATION_VISIBILITY_2026_08_21.md`, extending Phase B (#2764).

Phase B's `resolve_transcript_request_tier_fields` was written tier-generically from the start — it reads `req.delivery_tier` and matches trust grants against whichever tier a request actually arrived on. But it was only ever **called** from `handle_reactive_inject` (the HTTP handler). WAN delivery doesn't go through that path at all: `muxbus::cloud_subscriber::sync_agent_reactive` polls the cloud relay for pending injections and calls `Handler::inject_message` **directly**. Without also calling the resolution function there, a WAN-delivered `transcript_request` would silently skip both confirmed jekt rules entirely — not a weaker version, nothing at all.

**Fix:** `resolve_transcript_request_tier_fields` made `pub(crate)`, takes `&Arc<Store>` directly instead of `&AppState` (all it used was `.wstore`, which `sync_agent_reactive` already has as its own parameter — no new plumbing). `server::reactive` module widened to `pub(crate)`. One new call site, right before `sync_agent_reactive`'s existing `inject_message` call.

**Why this is the whole change:** see `docs/specs/SPEC_MUXSPECT_PHASE_C_WAN_TIER_ENFORCEMENT_2026_08_22.md` §3 — the design spec's own WAN-specific differences from Phase B (settings-UI guidance around `trusted_peers`, no-cache `ask` approval, bench-tuned `max_lines`) all concern pieces Phase B already deferred (the auto-responder, the settings UI, the approve/deny CLI) and this PR doesn't build either — so they're not yet applicable. Once this repo builds those, they'll need their own WAN-specific handling; today there's genuinely nothing to differentiate.

## Test plan
- [x] 2 new tests proving the shared resolution function behaves identically for `delivery_tier: "wan"` as it already does for `"lan"` — rule 1 forces sensitive; a matching-tier `trusted_peers` grant relaxes escalation (the existing Phase B suite already covered the negative "wrong tier" case)
- [x] Full `agentmux-srv` suite: 2746 passed (the one intermittent local failure, `refresh_processes_specifics_does_not_leak_a_handle_per_call`, is the same pre-existing environment-sensitive Windows handle-count flake seen throughout this session — confirmed passing 2/3 re-runs in isolation, untouched by this change)
- [x] `sync_agent_reactive` itself is not independently integration-tested, consistent with `cloud_subscriber.rs`'s own existing convention (every test there covers an extracted pure helper, never the top-level async polling orchestration)

### DIFF
--- a/.changesets/1787452355-fix-muxspect-wire-transcript-request-jekt-tier-enforcement-i-a01z.md
+++ b/.changesets/1787452355-fix-muxspect-wire-transcript-request-jekt-tier-enforcement-i-a01z.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(muxspect): wire transcript_request jekt tier enforcement into the WAN delivery path (Phase C)
--- a/agentmux-srv/src/backend/reactive/handler.rs
+++ b/agentmux-srv/src/backend/reactive/handler.rs
@@ -581,9 +581,12 @@ impl Handler {
         // verified sender when the RESPONDING agent's own
         // conversation_visibility is `ask` (or `trusted_peers` with a
         // non-allow-listed requester) — `req.transcript_request_escalate_forced`,
-        // resolved server-side in `handle_reactive_inject` against that
-        // agent's own setting (meaningless/always-false unless
-        // `is_transcript_request` is also true). A valid signature answers
+        // resolved server-side by `resolve_transcript_request_tier_fields`
+        // against that agent's own setting, called from both delivery paths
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2768, opened 2026-08-23
feat(layout): cross-fade the remaining pane-mount hard cuts (Phase 3-4)

AgentX-asaf:
## Summary

Implements Phases 3-4 of `docs/specs/SPEC_PANE_BLOCK_STACK_MOUNT_FLICKER_2026_08_22.md`, the two follow-up phases to #2761's leaf reveal gate (Phases 1-2). Phases 1-2 hide the *transition* between mount stages; these phases fix what's uncoordinated *inside* that transition and one independent layout-shift issue:

- **`Block`'s `ready()` gate** (`block.tsx`): the fallback `BrainSpinner` now cross-fades out on top of the real content (200ms) instead of an instant `<Show>` hard cut. Generic — applies to every block type, not just agent panes. Required making `.tile-leaf` (`tilelayout.scss`) a positioned containing block so the overlay's `inset:0` resolves correctly against its own box (which gets padding in split layouts) instead of skipping past it to `.tile-node`.
- **`AgentPicker` → `AgentPresentationView`** (`agent-view.tsx`): same overlay-fade technique, agent-pane-specific. Launching an agent from a blank "+" tab's picker sets `agentId` on the *same* block in place (no block-stack mutation, so PR #2761's reveal gate never covers this transition at all) — previously an instant hard cut. Now the picker fades out on top of the now-mounting `AgentPresentationView`, which shows its own pre-existing loading-overlay/history-load fade underneath — net effect: picker → spinner → content, no hard cuts.
- **`.pane-tab-strip` width transition** (`PaneTabStrip.scss`): CSS-only via `interpolate-size: allow-keywords` (stable since Chromium 129; this app bundles CEF 148) + `transition: width 160ms`, so the strip's shrink-to-fit width animates across a tab-count change instead of snapping. Independent of the other two.

Both overlay fades use the same pattern: apply the `is-overlay` positioning class the *same* render the transition starts (so there's never a frame where old/new content sit in normal flow together), then defer the actual opacity-fade trigger by one `requestAnimationFrame` (so there's a real painted frame to fade *from*), then unmount after the CSS transition duration elapses.

## Verification

- `tsc --noEmit`: clean.
- `vitest run`: full suite green (2990 passed, 3 skipped) — no regressions.
- `stylelint` on all touched SCSS: clean (pre-existing unrelated errors elsewhere in `agent-view.scss`/`tilelayout.scss` confirmed present on `main` too, not introduced here).
- **Not manually verified in a running browser.** `agent-view.tsx` and `block.tsx` have no existing component-render test harness to extend, and spinning up an isolated `task dev` instance plus UI automation against it was out of scope for this pass. The CSS containing-block/z-index reasoning (documented inline) is the highest-risk part — flagging for reviewer attention.

## Test plan

- [ ] Open a pane, click "+", confirm no hard-cut flash through spinner/picker before the new tab settles
- [ ] Launch an agent from a blank picker tab, confirm the picker fades rather than hard-cutting
- [ ] Open enough tabs to grow the tab strip, confirm the strip width animates rather than snapping
- [ ] Toggle OS-level reduced-motion, confirm all three transitions are suppressed


a5af:
@codex review

AgentX-asaf:
Addressed both rounds of feedback in a76e9f850:

**reagent (branch staleness, P1):** merged latest `main` (picks up #2765 and the v0.55.21 release commit) — version fields now match across `package.json`/`Cargo.toml`/`VERSION_HISTORY.md`.

**Codex (non-functional width transition, P2):** confirmed both points — `interpolate-size: allow-keywords` + `transition: width` doesn't actually animate here since `width`'s specified value never changes (only the DOM-content-driven used size does, and CSS transitions only fire on a discrete specified-value change), and the agent pane already overrides the strip to a fixed `left:0;right:0` full-width box anyway. Replaced with a measured FLIP transition in `PaneTabStrip.tsx` (hold the old measured width → force reflow → transition to the new width → clear back to `auto`), gated behind a new opt-in `animateWidth` prop passed only by `term.tsx`/`editor-tab-strip.tsx` — the two consumers that stay genuinely shrink-to-fit. Added 5 regression tests covering the hold/transition/cleanup sequence, the no-op cases, and confirming no animation on initial mount.

`tsc --noEmit` and the full `vitest run` suite pass (2994 tests; one pre-existing, unrelated flaky test — `tool-renderer registry` — times out only under full-suite parallel load, confirmed passing in isolation both before and after this branch's changes).

AgentX-asaf:
Confirmed and fixed in cd7d0349e — good catch.

Traced through it: under the previous code, when a second `tabs.length` change arrived before the first transition's cleanup timeout fired, `newWidth` was measured while `el.style.width` was still pinned to the first transition's target — so the comparison against `lastMeasuredWidth` (the same stale value) found them equal and silently did nothing, including never clearing that transition's own pending timeout. The strip stayed stuck at the wrong width until the original timer eventually fired.

Fix: clear any still-pinned `width`/`transition` *before* measuring `newWidth`, so it always reflects the true natural width for the current tab count. Also switched the "hold" reference from the stale `lastMeasuredWidth` target to the box's actual current visual position when interrupting an in-flight transition (measured before clearing), so restarting mid-flight doesn't itself introduce a small discontinuity beyond what's unavoidable.

Added a regression test (`re-targets to the current tab count's true natural width when a second change arrives before the first transition's cleanup timeout fires`) that fails against the old code and passes now — traced it manually to confirm: under the old logic it would have asserted `280px` stuck at `180px` instead of correctly reaching `260px`.

Full `vitest run` (2996 tests) and `tsc --noEmit` both clean.

PR #2770, opened 2026-08-23
Camper@claudius: fix(agent): stop Activity Dock flashing stale shell rows on pane load

GenericAgentX-asaf:
## Summary
- Reopening an agent pane briefly showed old, already-finished shell commands in the Activity Dock as running, then made them vanish — confirmed by direct user testing, not a hypothesis (asked timing + content clarifying questions first to rule out the dock's own intended retention-window fade behavior).
- Root cause: `shell_node_create` replays (persist:64 ring) on every pane mount/reconnect for every shell in the block's recent history, indistinguishable from a live spawn in the event payload — the frontend always built the node as `status: "running"` regardless, correcting only via a much slower, separate exit-event replay moments later.
- Fix: new `shellstatus` RPC command thinly wrapping the already-authoritative `ShellSessionRegistry::get_status`, fired immediately after every `shell_node_create`. If the shell already exited, pushes a fast correction via the same `pushShellExit`/`ShellStatusUpdate` path the real exit-event replay would eventually use anyway — a much cheaper single-registry-lookup round trip that in practice wins the race before the "running" row ever paints.
- Node creation itself is unchanged (still synchronous, always "running" at that point) — deliberately avoids introducing any risk of a chunk/exit event racing ahead of its own node's existence in the document (an earlier draft that deferred node creation itself was rejected for this reason; see the retro).

Full investigation + design rationale: `docs/retro/retro-activity-dock-stale-shell-flash-on-load-2026-08-22.md`

## Test plan
- [x] `cargo test -p agentmux-srv` — 2750 passed, 0 failed (4 new tests for the `shellstatus` RPC handler: running / clean exit / nonzero exit / unknown id, exercised through the real `WshRpcEngine` dispatch path)
- [x] `npx vitest run` — 5 new tests for the pure `shellStatusCorrection` decision function; one unrelated pre-existing flaky timeout in `tool-renderers/registry.test.ts` confirmed to pass in isolation
- [x] Changeset added

### DIFF
--- a/.changesets/1787461233-fix-agent-stop-activity-dock-flashing-stale-shell-rows-on-pa-a8zk.md
+++ b/.changesets/1787461233-fix-agent-stop-activity-dock-flashing-stale-shell-rows-on-pa-a8zk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): stop Activity Dock flashing stale shell rows on pane load
--- a/agentmux-srv/src/backend/rpc_types/block.rs
+++ b/agentmux-srv/src/backend/rpc_types/block.rs
@@ -285,6 +285,12 @@ pub struct CommandShellStopData {
     pub shell_id: String,
 }
 
+/// Data for ShellStatusCommand — query a shell's current running state by id.
+#[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #2771, opened 2026-08-23
feat(ci): nightly workflow auto-tags and publishes the latest pending version bump

Agent1-asaf:
## Summary

Adds `.github/workflows/nightly-release.yml`. Every night (chained off `ci-nightly-build.yml` via `workflow_run`, plus `workflow_dispatch` for manual runs), it checks whether the most recent `chore: release vX.Y.Z` commit on `main` has been tagged yet. If not, it tags it and pushes — automating the one manual step (`git tag && git push`) currently standing between a merged release PR and an actual published release. This is the exact command sequence performed manually for `v0.55.21` earlier today.

Once tagged, the existing `release.yml` (build 3 platforms, publish GitHub Release, dispatch landing-page deploy) and `container-image.yml` pipelines fire completely unmodified, exactly as they do for a manual tag push.

**What this workflow deliberately does NOT do:**
- Does not run `task release` or decide a version bump — that stays a human/agent decision.
- Does not touch `.changesets/*.md`.
- Does not build or package anything itself — zero new packaging code anywhere.
- If two `chore: release` commits land before either is tagged, only the newest is ever tagged (matches manual behavior today).

## ⚠️ Action required before this is live

**Requires a new repo secret: `NIGHTLY_TAG_PAT`** — a PAT with `contents:write`, ideally from a dedicated service/bot identity rather than any individual agent's personal PAT. This is **not optional**: GitHub's anti-recursion guard means a tag pushed using the default `GITHUB_TOKEN` will *not* trigger other workflows' `on: push: tags:` listeners — so without this secret, the workflow will find a pending release and then fail at the push step (with a clear `::error::` explaining why), never silently no-op. Documented in the workflow's own header comment.

## Design doc

Full design, including the `gh-reporter` (a5af/shared-infrastructure, cross-repo) half of this change that adds the nightly-release results to the existing "GitHub Nightly Report" email — see `docs/specs/SPEC_NIGHTLY_RELEASE_CHANNEL_2026_08_23.md`, included in this PR.

## Test plan

- [x] YAML syntax validated (`python -c "import yaml; yaml.safe_load(...)"`)
- [x] Detect-and-tag bash logic dry-run locally against real repo history — correctly identifies `v0.55.21` as already tagged (nothing pending, as expected today)
- [ ] End-to-end verification once `NIGHTLY_TAG_PAT` is provisioned and a real pending bump exists (tracked in the spec's rollout plan, §8)

### DIFF
--- a/.changesets/1787462822-feat-ci-nightly-workflow-auto-tags-and-publishes-the-latest--5fhn.md
+++ b/.changesets/1787462822-feat-ci-nightly-workflow-auto-tags-and-publishes-the-latest--5fhn.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(ci): nightly workflow auto-tags and publishes the latest pending version bump
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -83,7 +83,20 @@ jobs:
             exit 1
           fi
 
-          if git ls-remote --exit-code --tags origin "refs/tags/${version}" >/dev/null 2>&1; then
+          # persist-credentials: false (above) means `origin` has no auth
+          # configured, so this read must authenticate itself explicitly
+          # rather than relying on an anonymous request. Falls back to a
+          # plain `origin` read only when NIGHTLY_TAG_PAT isn't set yet
+          # (bootstrap period) — works today since this repo is public, but
+          # an authenticated read is correct regardless of that and doesn't
+          # depend on the repo staying public.
+          if [ -n "${NIGHTLY_TAG_PAT:-}" ]; then
+            remote="https://x-access-token:${NIGHTLY_TAG_PAT}@github.com/${REPO}.git"
+          else
+            remote="origin"
+          fi
+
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

Agent1-asaf:
Both findings confirmed and fixed in 4b0146d49:
- **P0** (persist-credentials): added `persist-credentials: false` to the checkout step.
- **P2** (grep matching full commit body): switched to filtering the formatted subject only. Verified against this repo's real history that 919b9327 (whose body contains "chore: release v0.50.3" but whose subject doesn't) is now correctly excluded, while the actual latest release commit (`v0.55.21`) is still correctly found.

Agent1-asaf:
Confirmed and fixed in 669581f9c: \`persist-credentials: false\` stripped auth from **all** \`origin\` reads, not just the push-overriding extraheader — so the already-tagged check's \`git ls-remote\` became anonymous too. Works today only because this repo is public (auth failure and "ref not found" are otherwise indistinguishable to the \`if\`). Now explicitly authenticates the ls-remote with \`NIGHTLY_TAG_PAT\` (same credential the push already uses), falling back to plain \`origin\` only when the secret isn't configured yet. Verified locally in both modes.

PR #2766, opened 2026-08-23
Smike@claudius: docs: Phase 1 of docs-cleanup audit — correct 6 stale spec statuses

GenericAgentX-asaf:
## Summary

Executes Phase 1 from `SPEC_DOCS_CLEANUP_AUDIT_2026_08_22.md`'s §5.1 — with real per-file verification, given that same spec's §2.2 already documented a case of confidently wrong pattern-matching. For each of the 6 files: read the actual spec content, confirmed the described feature exists in current code with file/line evidence, and found a real implementing PR via `git log` where determinable.

- `SPEC_BROWSER_AND_EDITOR_PANES_2026_04_16.md` → `implemented` (#415 editor, #422/#423 browser, LSP Phase 1 in #1074). One real gap found and noted: no `OpenBrowser` MCP tool exists, only `OpenEditor`.
- `SPEC_AGENT_PANE_ZONE_ORDER_WORKED_FOOTER_2026_04_24.md` → `implemented` (#549) — the code itself cites this spec by filename in a comment.
- `SPEC_AGENT_TOOL_CALL_TONES_2026_06_05.md` → `implemented` (commit `4bda9ce88`, follow-ups #2668/#2669/#2689).
- `SPEC_AGENT_SESSION_COST_TOTALS_2026_07_02.md` → `implemented` (#1920) — `reducer.ts`'s `accumulateStats()` cites this spec by filename.
- `SPEC_AGENT_STARTUP_SEQUENCE_2026_04_16.md` → `implemented` (#408), noting the "Forge" terminology it uses predates a later rename, and flagging an unverified related "Brief" doc rather than asserting a supersession relationship I haven't confirmed.
- `SPEC_AGENT_VIEW_SCSS_SPLIT_2026_04_24.md` → `implemented`, **not** the "half-done" status the parent audit spec guessed from line-count alone. Actually reading the file found it's essentially complete (28 split files matching the spec's target names; the remaining lines are legitimate root-layout, not unmigrated styles). Corrected the audit spec's own §2.1 entry for this file too — that was also a shallow-verification miss.

5 of 6 landed exactly as predicted; the 6th needed correcting in the *other* direction (more complete than guessed, not less) — included as a reminder that both directions of error are worth checking for, not just "did I overclaim."

PR #2772, opened 2026-08-23
fix(agent-spawn): inject per-agent git commit identity

AgentY-asaf:
## Summary

`git commit` falls through to whatever `user.name`/`user.email` sits in this machine's shared `~/.gitconfig` on a multi-agent host — one agent's real, GitHub-verified identity, silently baked into every OTHER agent's commits too. `agentmux-cloud`'s review-notification consumer resolves "who committed this" via GitHub's own commit→account auto-linking (`commit.author.login`), not raw git metadata directly — so whichever agent's real email happened to be in the shared config received every other agent's committer notifications.

**Confirmed live**, not from a bug report alone: every commit across 9+ PRs from 4 different agents (korp, smike, agentx, camper) on this host resolved to one agent's (mine) account this way. Full root-cause trace in the retro, including:
- Ruling out the recently-fixed merge-commit misattribution (agentmux-cloud#57) as a red herring — these are single-parent, genuine commits, a different bug.
- Confirming dual-delivery, not a single misroute: PR #2766's own body tag (`agent_id=smike`) proves the primary PR-author notification correctly resolves to smike; the *separate* committer-notification path is the one with bad input.

## Fix

Inject `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` + `GIT_COMMITTER_NAME`/`GIT_COMMITTER_EMAIL` at agent spawn time, scoped to that agent's own process tree — same pattern `AGENTMUX_AGENT_ID` already uses just above this code. Extracted into a small pure `git_identity_env_vars()` function for testability.

Email intentionally uses a non-existent `.local` domain: GitHub can never link it to a real account, so an agent with no dedicated PAT (the common case — see this repo's own `CLAUDE.md`, "Which GitHub account am I acting as?") simply drops out of `commit.author.login` lookups entirely instead of resolving to someone else's real, verified identity.

## Test plan
- [x] 4 new unit tests (`git_identity_env_vars`) — name/email mapping, casing (natural name, lowercase email), distinct agents never collide, and a locked-down assertion that the email domain stays `.local` (load-bearing for the fix's safety property)
- [x] `cargo check --bin agentmux-srv` clean
- [x] `cargo test --bin agentmux-srv agent_handlers::input::tests` — 4/4 pass

## Deploy note
This only takes effect for newly-spawned agent processes — every currently-running agent (including me) needs to be respawned after this merges and deploys to pick up the new env vars.

### DIFF
--- a/.changesets/1787463150-fix-agent-spawn-inject-per-agent-git-commit-identity-stop-mi-rys7.md
+++ b/.changesets/1787463150-fix-agent-spawn-inject-per-agent-git-commit-identity-stop-mi-rys7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-spawn): inject per-agent git commit identity, stop misattributing every agent's commits to whoever's real identity sits in the shared machine gitconfig
--- a/agentmux-srv/src/server/agent_handlers/input.rs
+++ b/agentmux-srv/src/server/agent_handlers/input.rs
@@ -15,6 +15,40 @@ use crate::backend::blockcontroller;
 
 use super::super::AppState;
 
+/// Per-agent git commit identity env vars for a spawned agent process
+/// (2026-08-22, docs/retro/retro-shared-git-identity-committer-misattribution-2026-08-22.md).
+///
+/// Without this, `git commit` falls through to whatever user.name/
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2773, opened 2026-08-23
fix(activity-dock): coalesce event-triggered refreshes to stop a pane-reopen request storm

AgentX-asaf:
## Summary

Reopening a pane with a nontrivial subagent/dispatch history (e.g. a heavily-used "My Agents" entry) currently triggers a multi-second request storm. Root-caused via a live trace of this session's own agent pane reopening — see `docs/reports/REPORT_AGENT_PANE_REOPEN_SUBAGENT_STORM_2026_08_23.md` for full measurement.

**Root cause:** pane reopen replays the session's (capped, ≤200-event) subagent/dispatch history as a burst of `subagent:spawned`/`completed`/`abandoned`/`dispatch:updated` WebSocket events — an existing, already-fixed OOM-crash guard from `docs/retro/retro-subagent-backfill-storm-oom-2026-07-17.md`. But `dispatch-source.ts` and `subagent-source.ts` (the Activity Dock's two data singletons) each fire their own **uncoalesced** `ListDispatches`/`ListActive` RPC call on *every single event* — no debounce anywhere in the stack (`callBackendService` is a bare `fetch()`). Measured: 155 backfill events → ~300 overlapping RPC calls → round-trip latency for the same two endpoints grew from 84ms to **7.2 seconds** as the queue backed up, ~14 seconds total from pane reopen to settled, with a 291ms main-thread Long Task and a normally-instant `main_window_focus` IPC call taking 2.08s as knock-on effects.

**Fix** (design doc: `docs/specs/SPEC_ACTIVITY_DOCK_REFRESH_COALESCING_2026_08_23.md`): a new shared `createDebouncedRefresh` (trailing-edge debounce + a max-wait ceiling, so a burst collapses into one call once it goes quiet, while a sustained/continuous stream still refreshes at least once per second). Both singletons' event handlers now go through it instead of calling `refresh()` directly. The module-load-time initial refresh stays immediate/undebounced — only event-triggered refreshes are coalesced.

This is a frontend-only, no-backend-change fix (option 1 of 4 sketched in the REPORT) — deliberately the smallest, most targeted change; larger follow-ups (threading the backend's existing `live` flag through to the wire, generic in-flight RPC de-duplication) are documented as explicit non-goals for this pass.

## Test plan

- [x] `debounced-refresh.test.ts` (new, 6 tests): burst collapsing, trailing reset, max-wait ceiling under continuous triggering, no double-fire, instance independence, fresh cycle after settling.
- [x] `subagent-source.test.ts`: existing event-wiring tests updated for the new debounce timing; 2 new tests confirm a 50-event burst and a sustained continuous burst both coalesce correctly.
- [x] `dispatch-source.test.ts`: 3 new tests (previously zero event-wiring coverage in this file) — handler registration, single-event-type burst coalescing, and a mixed-event-type burst coalescing.
- [x] `tsc --noEmit`: clean.
- [x] `vitest run`: full suite green (3013 passed, 3 skipped; one pre-existing, unrelated flaky test — `tool-renderer registry` — confirmed to only time out under full-suite parallel load, not in isolation).

### DIFF
--- a/.changesets/1787508418-fix-activity-dock-coalesce-event-triggered-refreshes-to-stop-4njb.md
+++ b/.changesets/1787508418-fix-activity-dock-coalesce-event-triggered-refreshes-to-stop-4njb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(activity-dock): coalesce event-triggered refreshes to stop a pane-reopen request storm
--- a/docs/reports/REPORT_AGENT_PANE_REOPEN_SUBAGENT_STORM_2026_08_23.md
+++ b/docs/reports/REPORT_AGENT_PANE_REOPEN_SUBAGENT_STORM_2026_08_23.md
@@ -1,0 +1,272 @@
+# Analysis: reopening an existing "My Agents" pane triggers a multi-second frontend request storm
+
+**Date:** 2026-08-23
+**Trigger:** repo owner reopened a heavily-used agent ("AgentX," this session's
+own agent) from My Agents and asked why loading an existing agent takes a
+while, noting "docks and long-running tasks that show during load, then
+disappear."
+**Method:** this session's own agent pane reopening (block
+`537ce695-737f-4a88-b354-7fc67fc1119c`, channel `local-main-b28b7a-e5bfaf58`,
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #2776, opened 2026-08-24
fix(agent): surface a Reconnecting… status during stale-resume retry

AgentA-asaf:
## Summary

Live-incident follow-up to `docs/status/STATUS_STALE_RESUME_LIVE_REPRO_AND_FIX_PLAN_2026_08_23.md` (§6.2/§6.3).

While investigating a real ~39s "did it crash?" gap the operator observed in their own agent pane, root-caused it to the existing stale-`--resume` detect→recover→respawn cycle (#2693/#2699) — which worked correctly, but gave **zero UI feedback** the whole time.

- Publishes a new `agent-resume-retry` WPS event (`{status:"retrying"|"resolved"}`, `persist: 2`) from `retry_after_resume_failure` and both `EmitSessionOutcome` handling sites in `agentmux-srv/src/backend/blockcontroller/persistent.rs`, plus a safety-net publish on the non-retryable-failure path so the indicator can never get stuck showing forever.
- New `reconnecting` pane-state axis (mirrors `attachedTask`'s shape — no staleness/generation gating needed, unlike `compacting`, since both ends of this signal travel over the same reliable channel), wired through a new `useResumeRetryStream` hook into `AgentComposerStrip`'s existing status-chip area: shows `Reconnecting… Ns` alongside the existing `Compacting… Ns` treatment.
- Rewords the misleading `"retrying fresh, without --resume"` log line — it fires *before* the recovery scan runs, so it doesn't reflect that a recovered session id is very often attached.

**Correction on the linked status doc**: its §6.1 recommendation ("make the shared registry `session_id` a live pointer") turned out to already be shipped a day earlier by #2755 — not duplicated here. The status doc has been corrected in place rather than silently reflecting stale info.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — new backend tests (`retry_after_resume_failure_publishes_retrying_then_resolved_when_no_recovery_found`, `retry_after_resume_failure_only_publishes_retrying_when_recovery_is_found`) + full `blockcontroller`/`wps` module suites (259 tests) all pass
- [x] `npx vitest run` — new reducer tests + `useResumeRetryStream` hook tests + full `frontend/app/store` + `frontend/app/view/agent` suites (2067 tests, 133 files) all pass

### DIFF
--- a/.changesets/1787545507-fix-agent-surface-a-reconnecting-status-during-stale-resume--f09n.md
+++ b/.changesets/1787545507-fix-agent-surface-a-reconnecting-status-during-stale-resume--f09n.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): surface a Reconnecting… status during stale-resume retry
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -83,6 +83,42 @@ fn session_outcome_line(
     )
 }
 
+/// Publish a `wps::EVENT_AGENT_RESUME_RETRY` status ping — a free function
+/// (not a method) for the same reason `session_outcome_line` above is one:
+/// callable from the stdout-reader/process-waiter match arms, which only
+/// hold `_read`/`_wait`-suffixed clones, not `&self`. `status` is `"retrying"`
+/// (set when a stale `--resume` is detected and a retry/recovery attempt is
+/// about to fire) or `"resolved"` (set the moment the retry's outcome —
+/// Fresh or Resumed — is actually known). See
+/// `docs/status/STATUS_STALE_RESUME_LIVE_REPRO_AND_FIX_PLAN_2026_08_23.md` §6.2.
+/// No-ops if `broker` is `None` (tests / non-wired controllers), same
+/// posture as every other best-effort WPS publish in this file.
+fn publish_resume_retry_status(broker: &Option<Arc<wps::Broker>>, block_id: &str, status: &str) {
+    let Some(broker) = broker else { return };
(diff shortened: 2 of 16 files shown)


a5af:
@codex review

AgentA-asaf:
Re: codex's P2 on \`useResumeRetryStream.ts\` (pane unmount/remount loses in-progress reconnect visibility) — verified directly: \`registerPane\`/\`unregisterPane\` (\`frontend/app/store/agent-pane-state-store.ts:211-224\`) fully destroy and recreate \`AgentPaneState\` on every pane remount (fresh \`initialState\`, slot deleted on unregister), and \`Broker::replay_to_route\`'s once-per-\`(route, event, scope)\` dedup (\`agentmux-srv/src/backend/wps.rs:198-280\`) only replays fresh on a true route reconnect, not a same-connection remount.

Confirmed this is not a new gap this PR introduces — it's the existing, accepted characteristic of the whole pane-state slot architecture, and \`compacting\` (the closest sibling this PR mirrors) has the identical limitation today: it uses \`persist: 0\` specifically because replay is unreliable across compaction's cross-channel completion gap, and has no remount-survival mechanism either. \`attachedTask\` is the one axis with a fix (an explicit recompute-from-transcript effect in \`agent-view.tsx\`), but that's viable there because attached-task state is derivable from the transcript; a stale-resume retry isn't.

Given this matches an already-accepted limitation rather than regressing anything, and fixing it properly would mean a broader pane-state-persistence change out of scope for this fix (which is scoped to §6.2/§6.3 of the linked status doc), leaving as a known, narrow edge case (pane must unmount+remount within the multi-second retry window) rather than fixing here. Happy to file a follow-up if it's judged worth its own pass.

PR #2777, opened 2026-08-24
fix(agent-pane): stop the block-mount BrainSpinner overlay getting stuck visible forever

AgentY-asaf:
## Summary

Live-reported: a "pulsating brain logo" (`BrainSpinner`) stuck permanently
visible in agent panes, on top of already-loaded, actively-rendering
content — not a cosmetic flicker, an indefinite stuck state. Confirmed
live in the reporting session's own hosting window via `UIScreenshot` +
`UIQuery`: `.block-ready-gate-host.is-overlay` elements mounted at full
opacity, no `is-fading`, in panes that had been actively rendering new
content for the entire (long, ~180k-token) session.

**Root cause** (full writeup: `docs/retro/retro-block-ready-gate-spinner-stuck-visible-race-2026-08-23.md`):
introduced by #2768's `ready()`-gate cross-fade in `block.tsx`. It seeded
`spinnerVisible` from `!ready()` read once at component *construction*
(synchronous), then relied on `on(ready, ..., { defer: true })`'s first
(swallowed) run to treat "already ready" as "nothing to do" — assuming
that first run would see the same `ready()` value the seed did. It won't,
reliably: `createEffect`'s first flush happens a render pass *after*
construction. If `ready()` flips `false → true` in that gap — the common
case for a block whose `blockData`/`viewModel` resolve from an
already-warm cache within a microtask of mount, not a rare edge case —
the seed (`true`) is never corrected: `defer` skips the one callback
invocation that would set it back to `false`, and since `ready()` never
changes again, nothing else ever does either. Spinner stays mounted,
full opacity, pulsing its normal 1.8s CSS animation, forever.

The identical shape existed in `agent-view.tsx`'s `AgentPicker` ->
`AgentPresentationView` cross-fade, added in the same PR, seeded from
`agentId()` instead of `ready()`. Not separately live-reproduced, but
structurally identical — fixed alongside the confirmed one.

Swept the rest of the frontend for the same pattern (construction-time
seed from a signal + `on(that signal, ..., { defer: true })`) — the only
other `{ defer: true }` sites (`agent-view.tsx`'s turn-end re-poll,
`PaneTabStrip.tsx`'s width transition, two unrelated hooks) don't pair a
construction-time seed with the deferred signal, so they don't have this
bug shape.

## Fix

Collapse the two separate reads into one: read the signal once per effect
run and special-case the first run inline via a local (non-reactive)
`initialized` flag, instead of splitting "seed at construction" and
"react to changes" across two different reactive primitives that have to
agree by coincidence.

```ts
const [spinnerVisible, setSpinnerVisible] = createSignal(true);
let spinnerGateInitialized = false;
createEffect(() => {
    const isReady = ready();
    if (!spinnerGateInitialized) {
        spinnerGateInitialized = true;
        setSpinnerVisible(!isReady);
        setSpinnerFading(false);
        return;
    }
    // ...unchanged fade-out / re-show logic
});
```

Now there is exactly one read of `ready()` deciding the initial state,
inside the same effect that reacts to later changes — no gap for the two
to disagree in. Same treatment applied to `agent-view.tsx`'s
`pickerVisible`/`agentId()` pair.

## Test plan

- [x] `npx vitest run app/block app/view/agent` — 1375/1376 passing. The
      one failure (`tool-renderers/registry.test.ts` timeout) is a
      pre-existing flake under full-suite parallel load, already
      documented in `docs/retro/retro-activity-dock-stale-shell-flash-on-load-2026-08-22.md`
      — unrelated to this change, passes in isolation.
- [x] Confirmed the BUG live before this fix: `UIScreenshot` + `UIQuery`
      against a running 0.55.21 portable build showed the exact stuck
      `.block-ready-gate-host.is-overlay` elements described above.
- [ ] Not yet re-verified live with this fix applied (would require a
      fresh build/relaunch) — flagging explicitly rather than claiming
      visual confirmation I don't have.
- No dedicated component-render test exists for `block.tsx` or
  `agent-view.tsx` (`Block`/`AgentViewWrapper` are heavy RPC/store
  orchestration components with no render harness — the same gap #2622
  and #2768 both explicitly flagged and deferred). Followed the same
  precedent rather than introducing a new test harness under a bug fix.

### DIFF
--- a/.changesets/1787550573-fix-agent-pane-stop-the-block-mount-brainspinner-overlay-get-tod8.md
+++ b/.changesets/1787550573-fix-agent-pane-stop-the-block-mount-brainspinner-overlay-get-tod8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): stop the block-mount BrainSpinner overlay getting stuck visible forever
--- a/docs/retro/retro-block-ready-gate-spinner-stuck-visible-race-2026-08-23.md
+++ b/docs/retro/retro-block-ready-gate-spinner-stuck-visible-race-2026-08-23.md
@@ -1,0 +1,123 @@
+# Retro: block-mount BrainSpinner overlay can get stuck visible forever
+
+**Date:** 2026-08-23
+**Owner:** AgentY
+**Area:** `frontend/app/block/block.tsx` (`Block`'s `ready()` gate), `frontend/app/view/agent/agent-view.tsx` (`AgentPicker` -> `AgentPresentationView` cross-fade)
+**Introduced by:** PR #2768 (`f59bb43b7`, 2026-08-22), "feat(layout): cross-fade the remaining pane-mount hard cuts (Phase 3-4)"
+
+---
+
+## Symptom
+
+Reported by the user directly: a "pulsating brain logo" (`BrainSpinner`)
+visible in agent panes, seemingly permanently, on top of already-loaded,
+actively-rendering content. Confirmed live in the reporting session's own
+hosting window via `UIScreenshot` + `UIQuery`: multiple
+`.block-ready-gate-host.is-overlay` elements were mounted at full opacity,
+carrying no `is-fading` class, in panes that had been actively rendering
+new content for the entire (long-running) session. Not a cosmetic flicker —
+a real, indefinite stuck state.
+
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2780, opened 2026-08-24
fix(agent-pane): strip common indentation from mid-file tool previews

Agent2-asaf:
## Summary
- Implements `docs/specs/SPEC_TOOL_PREVIEW_DEDENT_2026_08_08.md` — refinement #2 of the tool-preview series (refinement #1, scrollbar-edge padding, shipped in #2467).
- Root problem: a mid-file `Read` (with `offset`) or an `Edit` hunk deep inside a nested scope renders every line with the source file's full original indentation, even though the preview can't show the enclosing scopes that indentation refers to — wasted columns in an already-tight preview.
- New `frontend/app/view/agent/components/dedent.ts`: literal longest-common-whitespace-prefix stripping (no tab-width guessing — a file mixing tabs/spaces at the same depth correctly no-ops rather than guessing), always preserving *relative* indentation between lines.
  - `stripCommonIndent` — the base algorithm.
  - `stripCommonIndentNumbered` — Read-specific: Claude Code's Read results are `<N>\t<code>` lines (verified against a real session transcript in the spec's research); this splits off that prefix before dedenting so the line-number gutter is untouched, and falls through safely to the plain algorithm if the numbered shape doesn't match.
  - `stripCommonIndentSharedPrefix` — a third helper (not spelled out in the spec's original two-function sketch, added to keep the "one shared prefix across two related strings" concern in this module rather than duplicated at the call site) for Edit's `old_string`/`new_string` pair: computes ONE prefix across both sides together so an independent per-side dedent can't skew add/del alignment and manufacture a phantom indentation diff.
- Wired into the three call sites the spec identified: `renderRead`/`renderWrite` in `ToolOverlayLog.tsx` dedent their visible (post-`capText`) text; `DiffViewer` dedents `old_string`/`new_string` before building its LCS diff. Bash, Grep/Glob, and streaming chunks are explicitly out of scope (spec §2's own reasoning — independent match lines / no stable "file indentation" concept / can't compute a common prefix mid-stream).
- Language detection (`detectLanguage`) still reads the RAW (non-dedented) first line for Read/Write — shebang/content sniffing should see the file as-is; only the *displayed* text is dedented.

## Test plan
- [x] 25 new unit tests (`dedent.test.ts`) covering: uniform space/tab indent, mixed-indent no-op, blank-line handling, CRLF, shallowest-line-wins, the numbered-Read variant (incl. non-uniform/unnumbered fallback, multi-digit line numbers), and the shared-prefix diff variant (incl. asymmetric old/new depths, pure insertion/deletion)
- [x] `npx vitest run` — 3051 passed, 0 failed
- [ ] Live-pane visual check (spec §5's matrix: mid-file Read, deeply-nested Edit diff, normal Write no-op, `.md` Read) — no dev instance was live to check against at PR time; please do this before/at merge if practical, or I'll follow up with a dev-build check

### DIFF
--- a/.changesets/1787554977-fix-agent-pane-strip-common-indentation-from-mid-file-tool-p-9uya.md
+++ b/.changesets/1787554977-fix-agent-pane-strip-common-indentation-from-mid-file-tool-p-9uya.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): strip common indentation from mid-file tool previews
--- a/docs/specs/SPEC_TOOL_PREVIEW_DEDENT_2026_08_08.md
+++ b/docs/specs/SPEC_TOOL_PREVIEW_DEDENT_2026_08_08.md
@@ -1,7 +1,12 @@
 # SPEC: Tool preview common-indentation stripping (dedent)
 
 **Date:** 2026-08-08
-**Status:** proposed — verified unimplemented as of 2026-08-10 (sibling refinement #2467 shipped; this one did not).
+**Status:** implemented 2026-08-24. `stripCommonIndentSharedPrefix` was added
+as a third exported helper (not spelled out in §3.1's two-function sketch) to
+keep the Edit call site's "one shared prefix across two related strings"
+concern inside this module rather than duplicated at the DiffViewer call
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2781, opened 2026-08-24
fix(agent): gate the BrainSpinner on subagent backfill so the Activity Dock never flickers

AgentA-asaf:
## Summary

Follow-up to `docs/retro/retro-activity-dock-flicker-survives-debounce-fix-2026-08-24.md` — repo owner reported (on `0.55.23`) that Activity Dock rows still show-then-disappear on reopening a heavily-used agent pane, even after PR #2773's debounce fix.

Root cause (full writeup in the retro): PR #2773 reduced request *volume* (~300 calls down to a handful) but never claimed to stop the dock from rendering real, transient intermediate states — a large subagent backfill still produces 2-3 genuinely different snapshots as it streams in, and the BrainSpinner's `ready()` gate (`block.tsx`) has zero awareness of the Activity Dock's data (a separate, app-wide singleton source) — it only depends on this block's own `blockData`/`viewModel`, which resolves long before the backfill converges.

### What this PR does
- **Backend**: `SubagentWatcher::scan_session_subagents` now brackets its whole call with a new, scoped, persisted WPS event (`subagent:backfill_status`, `{status:"started"|"done"}`), published through a newly-wired `wps::Broker` (`SubagentWatcher::set_broker`, called once from `bootstrap.rs`) rather than the unscoped/unpersisted `EventBus` every other subagent event uses — this one specifically needs scope+persist so a pane can query current status directly via `EventReadHistoryCommand`, not just rely on live-event timing.
- **Frontend**: new `useSubagentBackfillGate` hook mirrors `useResumeRetryStream`'s exact shape, including the same race fixes that hook needed across its own review rounds (live subscribe + mount-time history read + discard-on-live-race). Wired into `block.tsx`'s `ready()` memo as one more AND condition. No-op for every non-agent block; defaults to "settled" so a block with no persisted session id (never triggers a backfill) can never get stuck gated.
- Also updates the retro's own status line so it doesn't go stale the same way two earlier docs did this session.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — full suite, 2759 tests pass (including 3 new tests for the started/done broadcast + no-broker no-op)
- [x] `npx vitest run` — full suite, 3031 tests pass (including 5 new tests for the pure status resolver)
- [ ] Manual verification in a running build (not done this pass — flagging plainly per this repo's own precedent for `block.tsx`/`agent-view.tsx` changes with no render harness, e.g. the adjacent `retro-block-ready-gate-spinner-stuck-visible-race-2026-08-23.md`)

### DIFF
--- a/.changesets/1787555508-fix-agent-gate-the-brainspinner-on-subagent-backfill-so-the--q1w7.md
+++ b/.changesets/1787555508-fix-agent-gate-the-brainspinner-on-subagent-backfill-so-the--q1w7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): gate the BrainSpinner on subagent backfill so the Activity Dock never flickers
--- a/agentmux-srv/src/backend/subagent_watcher/mod.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/mod.rs
@@ -155,6 +155,20 @@ pub struct SubagentWatcher {
     /// exist) -- every test call site built via bare `new()` skips this
     /// event entirely rather than panicking, same posture as `self_ref`.
     broker: Mutex<Option<Arc<crate::backend::wps::Broker>>>,
+    /// reagentx P2 (PR #2781): `scan_session_subagents` can be called twice
+    /// for the SAME `parent_block_id` in overlapping fashion (the same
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

PR #2782, opened 2026-08-24
feat(armory): AgentMux-controlled, highest-priority Global Memory system tier

Agent2-asaf:
## Summary

Adds a new **system tier** within Global Memory (Armory → Memory tab): an AgentMux-controlled entry that is always injected first into every agent's `CLAUDE.md`/`.claude/AGENTMUX_MEMORY.md`, wrapped in explicit override wording, and structurally isolated from every generic bundle-editing surface.

Full design + threat-model discussion: `docs/specs/SPEC_GLOBAL_MEMORY_SYSTEM_TIER_2026_08_24.md`.

**Backend**
- New `db_bundles.is_system` column (`OBJECT_SCHEMA_VERSION` v27 / `SHARED_STORE_SCHEMA_VERSION` v9, plus the identity-store mirror).
- `Store::bundle_memory_upsert`/`_delete` now refuse outright to touch an existing `is_system=1` row (content included, not just the flag).
- New `Store::bundle_memory_upsert_system`/`_delete_system` are the ONLY paths that can write/remove `is_system=1` — hardcode `is_blank`/`is_global`/`is_system` regardless of caller input.
- `bundle_memory_reorder` silently skips system rows; `bundle_memory_list_global` sorts them first regardless of `sort_order`/name.
- New RPC commands `upsertsystemmemory`/`deletesystemmemory` — never wired to any MCP tool, kept separate from `upsertmemory`/`deletememory`.
- `format_global_brain_block` splits system entries out and renders them first with explicit override wording ("IMPORTANT: ... take the HIGHEST PRIORITY ... OVERRIDE any default behavior ... you MUST follow them exactly as written"), mirroring the language already used by the outer `~/.agentmux/agents/CLAUDE.md`.

**Frontend**
- `GlobalBrainManager` (Armory Memory tab) renders a pinned, badged "AgentMux System" section above ordinary Global Memory entries, wired to the dedicated system-tier commands only.
- `GlobalBrainViewModel` gains `systemSectionsAtom`/`ordinarySectionsAtom` plus `saveSystemEdit`/`removeSystem`, with entirely separate draft-editing state from the ordinary editor.

**Important limitation, documented explicitly in the spec's §2:** there is no cryptographic UI-vs-agent boundary anywhere in this codebase today — the same `X-AuthKey` reaches both the frontend and every agent's own shell environment. This feature delivers a **structural + convention + prompt-priority** boundary (no generic bundle-writing path can reach an `is_system` row even by accident, and the injected content carries real override weight for the model), not a hard security one. Flagged so it's never mistaken for more than it is.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` (isolated `AGENTMUX_DATA_DIR`/`CONFIG_DIR`/`SHARED_DIR`/`INSTANCE_DIR`) — 2763 passed, 0 failed (includes 7 new tests covering every guard + the formatter split).
- [x] `npx tsc --noEmit -p .` — clean.
- [x] `npx vitest run` — 3060 passed, 0 failed (includes 7 new tests: formatter mirror + view-model split).
- [ ] Manual live-pane verification (spec §5) — not done as part of this PR; no live dev instance was available. Same caveat as the sibling dedent PR (#2780) — do before/at merge if practical.

### DIFF
--- a/.changesets/1787561798-feat-armory-add-agentmux-controlled-highest-priority-system--u77c.md
+++ b/.changesets/1787561798-feat-armory-add-agentmux-controlled-highest-priority-system--u77c.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): add AgentMux-controlled, highest-priority system tier to Global Memory
--- a/agentmux-srv/src/backend/agent_seed.rs
+++ b/agentmux-srv/src/backend/agent_seed.rs
@@ -280,6 +280,7 @@ fn seed_memories(wstore: &Arc<Store>, manifest: &SeedManifest) -> Result<usize, 
             sort_order: idx as i64,
             created_at: now,
             updated_at: now,
+            is_system: false,
         };
         // Use warn-and-skip rather than ? so a user bundle whose name
         // collides with the seeded name (UNIQUE constraint on name) does
--- a/agentmux-srv/src/backend/agent_session/tests.rs
+++ b/agentmux-srv/src/backend/agent_session/tests.rs
@@ -671,6 +671,7 @@ fn template_promote_resolves_provider_through_the_templates_bundle_not_the_drift
         sort_order: 0,
         created_at: 0,
         updated_at: 0,
+        is_system: false,
     };
     wstore.bundle_memory_upsert(&bundle).unwrap();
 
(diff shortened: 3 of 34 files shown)


a5af:
@codex review

PR #2783, opened 2026-08-24
feat(linux): recover from AppArmor's unprivileged-userns sandbox restriction

Agent1-asaf:
## Summary

Fixes a real Ubuntu bug report from this session: the `v0.55.21` AppImage failed to launch (no window, no visible error) unless run with `--no-sandbox`.

**Root cause (confirmed, not a regression):** AgentMux uses Chromium's kernel user-namespace sandbox on Linux rather than a root-owned SUID `chrome-sandbox` binary — a reasonable original choice (`SPEC_CEF_SANDBOX_2026_06_20.md`). Ubuntu backported an AppArmor restriction on unprivileged user-namespace creation (23.10, later security-patched into 22.04/20.04 too, ~early 2024) that blocks this for any Chromium/CEF-based app system-wide — not AgentMux-specific, it hit Chrome, VS Code, Discord, and Slack the same way.

Since AppImage has no install step (confirmed during investigation: no `.deb`/`.rpm` actually exists for this project despite docs claiming otherwise — that repo was deleted and never had a working Linux build), a real working sandbox requires one unavoidable one-time privileged action. Full design: `docs/specs/SPEC_LINUX_SANDBOX_APPARMOR_USERNS_2026_08_23.md`.

## What this does

- **`agentmux-cef/src/linux_sandbox.rs` (new)** — proactively probes whether unprivileged userns creation works (spawns `unshare`, not `libc::unshare()` in-process — avoids fork()-in-multithreaded-process hazards) **before** ever calling `CefInitialize`. If blocked, shows a native `zenity`/`kdialog` dialog:
  - **Fix it now** — installs a one-time AppArmor exception via `pkexec`, glob-scoped on the extract-once-cache's version segment so it survives every future AgentMux update without reinstalling (the reason this uses an AppArmor profile rather than reviving SUID `chrome-sandbox`, which wouldn't survive an update).
  - **Continue without sandbox this time** — proceeds unsandboxed for this session only, visibly (not silent).
  - **Cancel** — exits.
  - Never fails silently, even if neither dialog tool is available.
- **`lib.rs`** — wires the probe into the browser-process-only startup path (explicitly skipped for CEF's renderer/GPU/utility subprocess re-execs), plus a Linux sibling of the existing Windows `MessageBoxW` post-init-failure dialog as a safety net for other CEF init failures — same reasoning as that existing code: the web frontend can't render an error when the host itself failed to start.
- **`scripts/install-userns-apparmor-fix.sh` (new)** — the `pkexec`-invoked helper. Mechanical copy+reload only; the profile *text* lives solely in `build_apparmor_profile()` (tested), not duplicated in a separate bundled template that could drift.
- **`docs/linux.md`** — new troubleshooting section. **`SPEC_CEF_SANDBOX_2026_06_20.md`** gets a small forward-reference note so its "no setup needed" framing for Linux isn't read as still-unconditionally-true.

## What I could and couldn't verify from this environment

Being upfront about verification limits (no real Linux desktop available here):

- ✅ Full crate compiles clean, all 267 existing tests still pass.
- ✅ 13 new unit tests cover the OS-agnostic pure logic (AppArmor profile text generation, zenity/kdialog choice-parsing) — these run on any host by design, not gated behind `target_os = "linux"`.
- ✅ The core detection mechanism (`unshare(CLONE_NEWUSER)` failing in a restricted environment) was verified directly in a real Ubuntu 22.04 container.
- ✅ A true `x86_64-unknown-linux-gnu` target build was attempted; it got as far as compiling this crate's own code before failing on a missing cross-compilation toolchain for an unrelated dependency (`wayland-sys` needs `gcc`/`pkg-config` for the target, not installed here) — not blocked by anything in this change.
- ⚠️ **The platform-gated code itself (spawning `unshare`/`pkexec`/`zenity`/`kdialog`, the actual dialog UX/wording, the AppArmor profile syntax) has not run on real Linux.** It's written by close analogy to already-proven patterns in this exact codebase (the Windows `MessageBoxW` precedent, the existing `--type=` subprocess-detection pattern, this codebase's `// SAFETY:` comment convention) but genuinely needs real Ubuntu testing — this repo's own Linux CI (`build-linux.yml`) has the toolchain this dev environment lacks. Flagged as rollout step 4 in the spec.

## Test plan

- [x] `cargo test -p agentmux-cef --lib` — 267 passed (0 failed), including 13 new
- [x] Full crate builds clean with zero new warnings (`#![cfg_attr(not(target_os = "linux"), allow(dead_code))]` scoped narrowly to the expected cross-platform-unused pure items)
- [x] Both new shell scripts pass `bash -n` syntax validation
- [x] Reproduced the `unshare(CLONE_NEWUSER)` failure signature directly in a real Ubuntu 22.04 container
- [ ] Real end-to-end test on an actual Ubuntu box with the AppArmor restriction active — not possible from this environment, needed before considering this fully verified (see above)


a5af:
@codex review

Agent1-asaf:
Both P0s and the P1 confirmed and fixed in ced8dd70d:

**P0 (wrong probe target):** \`probe_userns_available()\` now re-execs \`current_exe()\` with a new internal \`--internal-probe-userns\` flag (intercepted as the very first check in \`run()\`) instead of spawning the generic \`unshare\` utility — the probe subprocess is now matched by the exact same AppArmor profile rule \`build_apparmor_profile()\` grants to AgentMux's own binary path, not an unrelated, permanently-unconfined one. Confirmed this bug would have made the probe report "blocked" forever on every future launch too, not just the same session — good catch.

**P0 (no relaunch):** after a successful \`pkexec\` fix, the process now relaunches in place via \`exec()\` (replaces the process image, preserves PID/PPID/env) so the entire process tree — including every CEF-spawned renderer/GPU subprocess — starts fresh under the newly-installed AppArmor profile, per §5.3.

**P1 (silent downgrade):** restructured the whole flow as a loop — a failed fix attempt now loops back to re-show the dialog instead of silently falling through to \`AGENTMUX_UNSAFE_NOSANDBOX=1\`. Only an explicit "Continue without sandbox" or "Cancel" click ends it.

**P2 (persistent indicator, inline comment):** also addressed — added \`linux_sandbox::RUNNING_UNSANDBOXED\`, checked by \`on_title_change\` to append " — Sandbox Disabled" to the window title for the rest of the session. Lighter-weight than a full flash-notification/banner plumb-through, but persistent and visible without new frontend/IPC work.

Full crate still builds clean (0 new warnings) and all 267 tests pass.

Agent1-asaf:
Both confirmed and fixed in 4ec59ad87:

**P1 (TOCTOU/predictable temp path — apologies for missing this in the first fix pass, you were right to re-flag it):** \`run_pkexec_fix()\` now creates the temp profile via a new \`create_secure_temp_profile()\` with three independent mitigations: a random (\`/dev/urandom\`-sourced) filename suffix so the path isn't predictable in advance, \`O_CREAT|O_EXCL\` so the write fails rather than following a pre-existing path/symlink, and mode \`0600\` so no other unprivileged user can read/substitute it even within the pre-pkexec window.

**P1 (browser-pane title leak, new finding):** confirmed — the suffix was appended to the shared \`title_str\` before the \`self.is_browser_pane\` branch, leaking into the \`browser-pane-title-change\` event for arbitrary loaded web pages. Fixed by introducing a separate \`display_title_str\`, gated on \`!self.is_browser_pane\`, used only for the actual window-title-setting paths — \`title_str\` (used for the pane event) now stays exactly what the page reported.

Full crate still builds clean (0 new warnings) and all 267 tests pass.

PR #2784, opened 2026-08-24
fix(ci): move nightly release auto-tagging to gh-reporter (workflow_run never fired)

AgentX-asaf:
## Summary

Root-caused and remediated why the nightly release automation never advanced past v0.55.21 despite v0.55.22 and v0.55.23 merging as `chore: release` commits. Full investigation: `docs/reports/REPORT_NIGHTLY_RELEASE_STUCK_AT_0_55_21_2026_08_24.md` (added in this PR).

**Root cause:** `.github/workflows/nightly-release.yml`'s `workflow_run` trigger (chained off `ci-nightly-build.yml`'s completion) never fired automatically, even once — confirmed via direct GitHub REST API queries (zero automatic runs, ever, despite every checkable precondition being satisfied: byte-identical name match, listener present on `main` for 8+ commits/many hours before the upstream run, upstream run itself succeeding). GitHub's API doesn't expose "why didn't this event fire" telemetry, so the exact platform mechanism can't be proven — but the practical effect was clear and repeatable.

**Fix, confirmed directly with the repo owner:** move the actual auto-tag-and-publish responsibility to `a5af/shared-infrastructure`'s `gh-reporter` Lambda, which already runs reliably on a real AWS EventBridge schedule (unlike the GitHub Actions `workflow_run` chain) and already has the GitHub credentials this needs. Companion PR there: implements the detect-and-tag logic, a new frequent (every 30 min) EventBridge schedule so a release merged mid-day gets tagged same-day rather than waiting for a fixed nightly slot, and a new email-section line reporting what it did. Design doc: `GH_REPORTER_RELEASE_AUTO_TAG_SPEC_2026_08_24.md` in that repo.

**This PR's own change:** `nightly-release.yml` is kept (not deleted) as a `workflow_dispatch`-only manual fallback — the unreliable `workflow_run` trigger is removed, since leaving it in place alongside gh-reporter's now-working automation would just be redundant, not additionally safe.

## Test plan

- [x] YAML validated with PyYAML (`workflow_dispatch` only, no syntax errors).
- [x] v0.55.23 already manually tagged as part of this investigation (before this PR) — confirmed `release.yml`/`container-image.yml` fired correctly off that tag push, proving the downstream pipeline itself needs no changes.
- [ ] Once the companion `a5af/shared-infrastructure` PR deploys, confirm its frequent schedule picks up a future release commit without needing this workflow's manual fallback at all.

### DIFF
--- a/.changesets/1787583263-fix-ci-move-nightly-release-auto-tagging-to-gh-reporter-work-ezqe.md
+++ b/.changesets/1787583263-fix-ci-move-nightly-release-auto-tagging-to-gh-reporter-work-ezqe.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): move nightly release auto-tagging to gh-reporter (workflow_run trigger never fired)
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -49,10 +49,9 @@ permissions:
 # just fail (tag already exists) rather than corrupt anything, but it's an
 # easily-avoided race for a job with a push side effect. Queue rather than
 # cancel — killing a run mid-push could leave a tag pushed with no
-# corresponding downstream build ever having started. Kept even though the
-# automatic workflow_run trigger (this guard's original reason for being)
-# is gone — reagent's review of PR #2784: cheap insurance regardless of how
-# many trigger types remain.
+# corresponding downstream build ever having started. This guard predates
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentY-asaf:
Fixed in 83107996c — reworded the concurrency-guard comment to state the rationale (cheap insurance regardless of trigger count) without naming a specific review round.

PR #2785, opened 2026-08-24
fix(agent-pane): render tool-preview tabs at 2 spaces

Agent2-asaf:
## Summary

Small follow-up to the tool-preview dedent work (`dedent.ts`, PR #2780, `docs/specs/SPEC_TOOL_PREVIEW_DEDENT_2026_08_08.md`). That utility strips only the leading indentation *common* to every displayed line and deliberately never conflates tabs with spaces (there's no reliable way to know how wide a tab renders). The *relative* indentation left over in a tab-indented preview was rendering at the app's global `tab-size: 4` (`reset.scss`, inherited everywhere including the Editor pane), making a tab-indented Read/Write/Edit preview read noticeably wider than the same file's space-indented sibling.

This is a CSS-only fix, not a `dedent.ts` change — `tab-size: 2` scoped to `.agent-tool-panel` (`frontend/app/view/agent/styles/_document-nodes.scss`), the shared ancestor of every Read/Write/Edit/Diff preview path (Shiki-highlighted and plain-fallback alike). The Editor pane's CodeMirror instance is untouched — it keeps the global 4-wide default, which is a separate concern from tool-preview rendering.

## Test plan
- [x] `npx sass` compile of the agent-view SCSS entrypoint — confirms the rule lands in the output CSS with no syntax errors.
- [x] `npx tsc --noEmit -p .` — clean (CSS-only change, no TS touched).
- [ ] Manual visual check in a live pane — not done as part of this PR; no live dev instance was available.

### DIFF
--- a/.changesets/1787583377-fix-agent-pane-render-tool-preview-tabs-at-2-spaces-instead--i3yd.md
+++ b/.changesets/1787583377-fix-agent-pane-render-tool-preview-tabs-at-2-spaces-instead--i3yd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): render tool-preview tabs at 2 spaces instead of the inherited 4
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2786, opened 2026-08-24
feat(agent-picker): on-demand Haiku activity summary fallback when no conversation snapshot exists

AgentY-asaf:
## Summary

Follow-up to the audit in `docs/reports/REPORT_AGENT_PICKER_FIELD_ORDER_SORT_AND_DATA_GAPS_AUDIT_2026_08_24.md` §5 — the AgentPicker's "My Agents" rows show `"(no conversation snapshot)"` for legacy rows that predate structured `output.state.json` snapshot persistence, even when the instance genuinely has real conversation history in its raw terminal capture.

- **New table** `db_agent_activity_summaries` (`OBJECT_SCHEMA_VERSION` v27): one row per `definition_id`, generated at most once, cached forever.
- **No new Haiku-calling machinery.** `generate_definition_activity_summary` (`agentmux-srv/src/server/app_api/session.rs`) reuses the exact same shared Ambient Model Call gateway (`invoke_ambient_haiku_call`, `crate::ambient::gateway()`) that `dispatch_name`/`subagent_name`/`term:ambient_summary` already share — just a new purpose tag (`"definition_summary"`) and a new persistence layer.
- **Source data**: the instance's raw `"output"` filestore file (written unconditionally by the CLI capture pipeline) via `read_recent_activity_digest` — the same extraction `generate_pushed_activity_summary` already uses. This is why it works even when the structured snapshot doesn't exist.
- **On-demand, fire-and-forget**: `listrecentsessions` (`agentmux-srv/src/server/agent_handlers/session.rs`) checks the persisted table synchronously (cheap indexed read); if nothing's there yet, it spawns generation in the background and returns immediately — the row keeps showing today's fallback text on that first load. On success, generation broadcasts `agents:changed`, which `MyAgentsList.tsx` already refetches on.
- **Zero frontend changes.** `<Show when={row.preview} fallback={...}>` already prefers a non-empty `preview` over the fallback text — populating `preview` from the persisted summary is enough.
- **Scoped to local rows only** — a cross-channel row's synthetic empty `block_id` has no local raw output to summarize either, so it's left alone (the existing fallback text is accurate for that case: this channel genuinely has nothing).

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2767 passed (one `sysinfo` handle-leak test flaked under full-suite parallel load, confirmed passing in isolation, unrelated Windows process-handle counting code untouched by this PR).
- [x] New tests: `agent_activity_summaries::tests` (4, storage get/set/upsert roundtrip), `agent_handlers::session::tests::snapshot_less_row_uses_a_persisted_definition_summary_as_preview` + `..._with_no_persisted_summary_leaves_preview_empty` (end-to-end through the real RPC dispatch).
- [x] `npx tsc --noEmit` — clean (no frontend changes).
- [ ] Not manually verified in a running instance — the Haiku CLI round-trip itself isn't exercised by these tests (would require a real CLI binary + subprocess spawn). The persistence/wiring logic is fully covered; the actual generation path is the same, already-production, `invoke_ambient_haiku_call` code every other ambient-call feature already relies on.

### DIFF
--- a/.changesets/1787584272-feat-agent-picker-on-demand-haiku-activity-summary-fallback--t0va.md
+++ b/.changesets/1787584272-feat-agent-picker-on-demand-haiku-activity-summary-fallback--t0va.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-picker): on-demand Haiku activity summary fallback when no conversation snapshot exists
--- a/agentmux-srv/src/backend/storage/agent_activity_summaries.rs
+++ b/agentmux-srv/src/backend/storage/agent_activity_summaries.rs
@@ -1,0 +1,116 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! On-demand, once-per-definition activity summary — see
+//! `db_agent_activity_summaries` in migrations.rs (`OBJECT_SCHEMA_VERSION`
+//! v27) and
+//! docs/reports/REPORT_AGENT_PICKER_FIELD_ORDER_SORT_AND_DATA_GAPS_AUDIT_2026_08_24.md §5a.
+//!
+//! Fallback preview text for the AgentPicker's "My Agents" rows whose
+//! instance has no structured `output.state.json` conversation snapshot —
+//! generated once, lazily, via `app_api::session::generate_definition_activity_summary`,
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2787, opened 2026-08-24
Manoz@Area54: Pause AskUserQuestion auto-timeout on keyboard activity, not just mouse hover

GenericAgentX-asaf:
## Summary
- Extends the AskUserQuestion 30s auto-timeout's hover-pause mechanism to also fire on keyboard activity, not just mouse hover.
- Reuses the existing `onPanelPointerEnter` pause function directly (same `hidden` signal, same flat `HOVER_HIDE_GRACE_MS` window, same resume behavior) — no new timer/signal/mechanism.
- Scoped to `inPanel` (keydown must land inside the panel's own DOM), mirroring the mouse trigger's scope exactly. Keystrokes elsewhere in the pane (chat composer, Ctrl+F) don't count as engagement with this question.
- No behavior change for Enter/Escape — both already tear down or reset this same pause state via their own existing paths, so firing the new trigger first for them is a harmless no-op.

See `docs/specs/SPEC_ASK_USER_QUESTION_TIMEOUT_KEYBOARD_PAUSE_2026_08_20.md` for full design rationale.

## Test plan
- [x] New unit tests in `AgentQuestionPanel.test.tsx` covering: hides on qualifying keydown, resumes at fresh 30s after 15s, second keydown restarts the window, keydown outside panel but inside pane doesn't trigger, keydown while minimized is a no-op, Enter/Escape still work, composes correctly with mouse-triggered pause.
- [x] Full existing suite still passes (29/29, no regressions to hover-pause or Enter/Escape tests).

### DIFF
--- a/.changesets/1787584354-feat-agent-view-pause-askuserquestion-auto-timeout-on-keyboa-iwv0.md
+++ b/.changesets/1787584354-feat-agent-view-pause-askuserquestion-auto-timeout-on-keyboa-iwv0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-view): pause AskUserQuestion auto-timeout on keyboard activity, not just mouse hover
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed all three findings in 6bd2c0205:
- reagentx P1 (unbounded pause via key-repeat/continuous typing): gated the trigger on `!hidden()`.
- codex P2 (minimized button falsely counted as "inside the panel"): gated on `!minimized()`.
- codex P2 (Tab-into-panel from outside doesn't pause, since keydown target is the element losing focus, not gaining it): added a `focusin` listener reusing the same gate.

Added 8 more tests (33 total, all passing) covering the fixed behavior, including a direct regression test for the key-repeat/continuous-typing case. Spec updated with a §8 post-review-revisions section.

@codex review

GenericAgentX-asaf:

PR #2788, opened 2026-08-24
Manoz@Area54: Provider-specific startup instructions filenames (AGENTS.md, GEMINI.md, ...) instead of always CLAUDE.md

GenericAgentX-asaf:
## Summary
- Every provider's agent (Codex, Gemini, Qwen, Copilot, OpenClaw, Pi, Antigravity, Mux Code, Claude) now gets its startup instructions written under its own confirmed native filename — `AGENTS.md`, `GEMINI.md`, `QWEN.md`, `.pi/APPEND_SYSTEM.md` — instead of always `CLAUDE.md`, which none of them actually read by any confirmed convention. See `docs/specs/SPEC_PROVIDER_AWARE_STARTUP_INSTRUCTIONS_2026_08_24.md` §2 for the full per-provider research (with citations) behind each filename choice.
- Kimi gets **no** instructions file written at all — it has no confirmed native file-based discovery mechanism today (re-confirmed against `docs/specs/KIMI_PROVIDER_INTEGRATION_SPEC.md`'s prior research), so writing one would just be inert output nobody reads.
- An unrecognized `provider_id` also gets no instructions file — never a silent `CLAUDE.md` fallback that could mask a typo.
- **Found and fixed a second hardcode while implementing** (not caught in the original research pass): the "click Launch" RPC write path (`editor_handlers.rs`) gated global-memory-bundle injection behind a literal `file.path == "CLAUDE.md"` check. Left as-is, a Codex agent's now-correctly-named `AGENTS.md` would have silently received **no** Global Memory content at all. Fixed via a new `providers::is_known_startup_instructions_filename()` membership check (single source of truth, no parallel list to drift).
- Armory → Memory (Global Memory) now shows an "Applies to" summary above the preview, grouping providers by resolved filename, plus an explicit "not yet applied to: Kimi Code CLI" callout — so operators can see where their content actually lands instead of assuming it's always `CLAUDE.md`.

## Explicitly out of scope (see spec §5)
- Per-provider *content* divergence (the existing `instructions_by_provider` DB column stays unused at runtime, same as today — this PR fixes filename only).
- Renaming `.claude/commands/`, `.claude/skills/`, `.claude/settings.json`, `.mcp.json` per provider.
- Extending CLAUDE.md's ownership-protection (`write_claude_md_respecting_ownership`, deeply coupled to Claude Code's own `@import` syntax) to non-Claude filenames — flagged as a real follow-up, not attempted here.
- OpenClaw's exact ACP-Gateway-sandboxed working-directory mapping, and Antigravity's `GEMINI.md` inference (shared `~/.gemini/` config lineage with Gemini CLI, not independently doc-confirmed) — both flagged in spec §6 as known gaps; today's behavior for both is unconditionally wrong (`CLAUDE.md`, confirmed unread), so a best-effort correct filename is strictly an improvement even where the exact path isn't 100% nailed down.

## Test plan
- [x] Rust: 46 new/updated tests across `providers.rs` and `agent_config.rs` (per-provider filename table, kimi/unrecognized-provider no-file behavior) — full suite (2773 tests) green.
- [x] Frontend: matching tests in `providers/index.test.ts`, `agent-config-builder.test.ts`, `global-brain-model.test.ts` — full suite (3097 tests) green.
- [ ] Manual (`task dev`): launch a Codex-provider agent, confirm `AGENTS.md` (not `CLAUDE.md`) appears in its working directory with the expected content; confirm Armory → Memory's new "Applies to" chips render.


a5af:
@codex review

GenericAgentX-asaf:
Addressed the P1 in c0dc00fd6: added `write_startup_instructions_respecting_existing`, a marker-based ownership guard for every non-CLAUDE.md startup-instructions filename, wired into both write paths. A pre-existing foreign `AGENTS.md`/`GEMINI.md`/`QWEN.md`/etc. is now never overwritten — mirrors `write_claude_md_respecting_ownership`'s OWNED-vs-foreign check via a new generic marker comment, without the `@import`-syntax-specific side-file fallback (flagged as a real follow-up in the spec's §5, not attempted here since it needs per-provider syntax verification).

6 new Rust tests cover it directly (fresh dir, regenerate-if-owned, foreign file untouched byte-for-byte, foreign empty file, non-UTF-8 foreign file, nested-path parent-dir creation). Full suite (2779 tests) green.

@codex review

GenericAgentX-asaf:

PR #2789, opened 2026-08-24
feat(agent-picker): sort control, reordered fields, distinct account-failure text

AgentY-asaf:
## Summary

Implements the remaining recommendations from `docs/reports/REPORT_AGENT_PICKER_FIELD_ORDER_SORT_AND_DATA_GAPS_AUDIT_2026_08_24.md` (§2, §3, §4), prompted by direct user feedback: "a lot of the fields are out of order," plus a request for name/launch-date/type sort controls at the right edge of the filter bar. (§5/§5a of that same report already shipped separately as #2786.)

- **§3 — sort control**: a right-aligned "Sort" select in `AgentPickerFilterBar` — **Recently launched** (default, matches today's implicit backend ordering exactly, so omitting the prop is a no-op behavior change) / **Name (A–Z)** / **Type** (Host before Sandbox, matching `RuntimeBadge`'s own HOST/SANDBOX vocabulary, not alphabetical). Applied client-side over `MyAgentsList`'s already-fetched row set — no backend RPC change. Persisted to `localStorage` (this-machine-only; no existing precedent in this component tree to extend, so this is a new small pattern).
- **§2 — field reorder**: the account/identity name was sharing line 1 with the name via `flex: 1`, squeezed after two badges — moved to its own full-width line below, since it's arguably as important as the name for a multi-account user and previously had the least reliable space on the card. Timestamps reordered to lead with the most relevant fact (Last Active, or Last Launch if never active) instead of the least-actionable one (Created), which was previously listed first.
- **§4 — backend fix**: when `agent_identity_list_all()` itself fails, every row used to fall back to the exact same `"(ambient creds)"` text a genuinely-unbound agent gets — a backend failure reading identically to a normal, healthy state. Now shows `"(unknown account)"` instead, gated on whether `identity_links` is present in the response's own `degraded` list (computed once, not per-row).

## Test plan

- [x] New tests: 6 sort-behavior tests (default/name/recent/type ordering, reactivity on `sortBy` change, in-place-mutation safety) in `MyAgentsList.test.tsx`; 1 backend test (`identity_links_source_failure_uses_distinct_fallback_text_not_ambient_creds`, forces the failure via `DROP TABLE db_agent_identity_links`) in `agent_handlers/session.rs`.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 2777 passed.
- [x] `npx vitest run` — 3016 passed (1 unrelated pre-existing flaky test — `tool-renderers/registry.test.ts`, timeout under full-suite parallel load — confirmed passing in isolation).
- [x] `npx stylelint` on both touched SCSS files — clean (9 pre-existing raw-hex-color errors at untouched line numbers, confirmed present on `main` before this change too via `git stash`).
- [ ] Not manually verified in a running instance — flagging explicitly rather than claiming visual confirmation I don't have.

### DIFF
--- a/.changesets/1787599247-feat-agent-picker-sort-control-reordered-fields-distinct-acc-im7g.md
+++ b/.changesets/1787599247-feat-agent-picker-sort-control-reordered-fields-distinct-acc-im7g.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-picker): sort control, reordered fields, distinct account-failure text
--- a/agentmux-srv/src/server/agent_handlers/session.rs
+++ b/agentmux-srv/src/server/agent_handlers/session.rs
@@ -334,6 +334,18 @@ pub fn register(engine: &Arc<WshRpcEngine>, state: &AppState) {
                     Vec::new()
                 });
 
+                // docs/reports/REPORT_AGENT_PICKER_FIELD_ORDER_SORT_AND_DATA_GAPS_AUDIT_2026_08_24.md
+                // §4: when `agent_identity_list_all()` itself failed above,
+                // `links_by_agent` is empty for EVERY definition_id — not
+                // just the ones that genuinely have no bound identity —
+                // so every row would otherwise show the exact same
+                // "(ambient creds)" text a real unbound agent gets,
+                // making a source failure indistinguishable from a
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #2790, opened 2026-08-24
feat(toolchain): one-click install of git/Node/npm/Python across Windows/macOS/Linux

Agent2-asaf:
## Summary

Implements `docs/specs/SPEC_SYSTEM_TOOLCHAIN_INSTALLER_2026_08_24.md` (Phases 1+2). Users hitting the "git/npm not found" blocker now get a real **Install** action instead of just a download link, in both places that already knew about the problem but never acted on it:

- **`AgentPrereqModal`** — the pre-launch blocker.
- **Toolchain modal** — this was already scaffolded (`brewFormula` sitting unused in the catalog since `SPEC_TOOLCHAIN_MANAGER_2026-06-15.md`'s deferred "P3") and is now wired for real, generalized beyond brew-only to winget and Linux's own package managers.

**Backend — the security-critical part.** New `system_install_handlers.rs`: every command comes from a fixed, backend-owned catalog keyed only by a `tool_id` (`git`/`node`/`npm`/`python`) — the RPC boundary never accepts a raw command or package name from the frontend, and every catalog entry is a `Vec<String>` argv passed straight to `Command::new(program).args(args)`, never a shell string. Windows uses `winget` (own UAC prompt per package); macOS uses `brew` only when already present (never bootstraps Homebrew); Linux detects the system package manager (`apt-get`/`dnf`/`yum`/`pacman`/`zypper`/`apk`) and elevates via `pkexec`, which raises the desktop's native polkit dialog — this app never renders a password field itself. Reuses the exact `install_chunk` streaming shape and `InstallSessionRegistry` the existing `install.start` npm-install flow already uses — `install.cancel` works unchanged for these sessions too.

**Frontend.** `SystemToolInstallInline.tsx` — a shared inline consent+streaming panel: shows the exact resolved command before anything runs, explains when it needs elevation, streams output, and (deliberately) never offers a cancel button once the privileged command has actually started — interrupting a package-manager transaction mid-write can leave broken system state. Renders nothing when unavailable, so callers' existing link+copy-command fallback stays intact.

**Explicitly not included** (flagged in the spec, not silently decided): bootstrapping a package manager that's itself missing (installing Homebrew from scratch, guiding a pre-App-Installer Windows), the Node/Debian-Ubuntu NodeSource-vs-bare-package data decision (Debian/Ubuntu's default `nodejs` package is well-known to be stale — noted, not fixed here), and `uv`/Docker (script-based / interactive-GUI installers, a different risk profile than a package-manager install).

**One open item, stated plainly:** whether a UAC prompt raised by `winget` reliably attaches when spawned from this app's non-console GUI-launched process is **not yet verified on real Windows hardware** — the command construction is implemented and unit-tested, but this is a real-environment question no amount of unit testing answers. Flagged in the spec's status line.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` (isolated `AGENTMUX_DATA_DIR`/`CONFIG_DIR`/`SHARED_DIR`/`INSTANCE_DIR`) — 2774 passed, 0 failed (6 new: catalog resolution per platform/manager, argv correctness per Linux package manager's genuinely different flag syntax, and a defense-in-depth check that no catalog entry ever routes through a shell interpreter).
- [x] `npx tsc --noEmit -p .` — clean.
- [x] `npx vitest run` — 3092 passed, 0 failed (5 new: unavailable/consent/elevation-note/success/failure states).
- [ ] Manual, per real OS/distro — not done as part of this PR (no live multi-OS test environment available): Windows winget + UAC origin question above, macOS brew (confirm no sudo prompt), 2-3 real Linux distros (confirm correct package manager detected, not a blanket apt), and a machine with no polkit agent (confirm graceful fallback).

### DIFF
--- a/.changesets/1787599434-feat-toolchain-one-click-install-of-git-node-npm-python-via--42on.md
+++ b/.changesets/1787599434-feat-toolchain-one-click-install-of-git-node-npm-python-via--42on.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(toolchain): one-click install of git/Node/npm/Python via winget/brew/system package manager
--- a/agentmux-srv/src/server/install_handlers.rs
+++ b/agentmux-srv/src/server/install_handlers.rs
@@ -70,6 +70,12 @@ struct InstallCheckReq {
 pub struct InstallSessionRegistry {
     sessions: Mutex<std::collections::HashMap<String, tokio::sync::oneshot::Sender<()>>>,
     active_providers: Mutex<std::collections::HashSet<String>>,
+    /// Same idea as `active_providers`, but a separate set keyed by
+    /// system-tool id (`"git"`, `"node"`, …) rather than provider id — a
+    /// distinct namespace so a future provider literally named e.g. "git"
+    /// can never collide with a system-tool claim. See
+    /// `system_install_handlers.rs` / `SPEC_SYSTEM_TOOLCHAIN_INSTALLER_2026_08_24.md`.
+    active_system_tools: Mutex<std::collections::HashSet<String>>,
 }
 
 impl InstallSessionRegistry {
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

PR #2792, opened 2026-08-25
fix(tabs): remove the active-tab color line spanning the whole tab strip

AgentY-asaf:
## Summary

Repo-owner-confirmed request: "we want to eliminate those lines that span under the tab bar." This is the window-level tab bar's colored underline (`frontend/app/tab/tab-color-line.ts` + `tabbar.tsx`'s `.active-tab-color-line`), **not** the per-agent pane border color feature (`SPEC_AGENT_COLOR_2026_08_08.md`) — confirmed that one is unaffected (colors a single pane's own block frame, unrelated code path).

The line intentionally spanned the FULL tab strip (not just the active tab) since PR #2135 (2026-07-13), per an explicit request at the time — see `specs/SPEC_ACTIVE_TAB_COLOR_LINE_STOP_AT_TAB_STRIP_2026_07_13.md`, marked `Status: Superseded` rather than deleted (kept as historical record of that geometry decision).

- Deleted `tab-color-line.ts` outright — the measurement/geometry hook (`useActiveTabColorLine`) had no other callers.
- Removed its usage, the `<Portal>`-rendered line `<div>`, and the now-unused `Portal` import from `tabbar.tsx`.
- The per-tab `tab:color` feature itself — the whole tab's own background tint (`.tab-colored` in `tab.scss`, set via the tab's right-click color picker) — is completely untouched; only the additional full-strip underline beneath it is gone.

## Test plan

- [x] `npx vitest run app/tab` — 38 passed.
- [x] `npx vitest run` (full suite) — 3033 passed (1 unrelated pre-existing flaky test, `tool-renderers/registry.test.ts`, confirmed passing in isolation elsewhere this session).
- [ ] Not manually verified in a running instance — flagging explicitly rather than claiming visual confirmation I don't have.


a5af:
@codex review

PR #2794, opened 2026-08-25
Manoz@Area54: Surface ~/.claude/CLAUDE.md (read-only) in Global Memory

GenericAgentX-asaf:
## Summary
- Adds a new read-only block at the top of Armory -> Memory (Global Memory) showing the real path and content of `~/.claude/CLAUDE.md` -- Claude Code's own global user-level config file, hand-maintained and independent of AgentMux, shared by every agent on this machine.
- Follows up on this conversation's earlier finding: the system-tier spec (`SPEC_GLOBAL_MEMORY_SYSTEM_TIER_2026_08_24.md` §1) claims a "highest priority" file exists at `~/.agentmux/agents/CLAUDE.md` -- that path doesn't actually exist on this host. The real file matching that description is `~/.claude/CLAUDE.md`, confirmed by directly reading my own startup files earlier in the session.
- New `getclaudeglobalconfig` RPC command: no parameters (fixed, hardcoded path -- no path-traversal surface), no write counterpart. `content: None, exists: false` for a missing file is a normal state, not an error; genuine I/O errors (permission denied, etc.) still propagate.
- Read-only by construction in the UI -- no textarea, no save button, no edit affordance. A tooltip on the badge explains why (hand-maintained, not AgentMux's, shared machine-wide).

See `docs/specs/SPEC_SURFACE_CLAUDE_GLOBAL_CONFIG_2026_08_24.md` for the full design (including the explicit scoping decision to keep this to just this one file, not per-provider example paths or a live per-agent path listing -- both considered and declined for this pass).

## Test plan
- [x] Rust: 3 new tests on the extracted `read_claude_global_config` helper (present, missing-not-error, empty-file-not-missing) -- full suite (2800 tests) green.
- [x] Frontend: model-level tests (`claudeGlobalConfigAtom` populates, fetch failure stays silent and doesn't pollute `errorAtom`) + new component-level tests (renders path/content, empty-state fallback, nothing before fetch resolves, no edit affordances) -- full suite (3120 tests) green.
- [ ] Manual (`task dev`): confirm the block shows this machine's real `~/.claude/CLAUDE.md` path and content above the system tier.

### DIFF
--- a/.changesets/1787629229-feat-armory-surface-claude-claude-md-read-only-in-global-mem-ommd.md
+++ b/.changesets/1787629229-feat-armory-surface-claude-claude-md-read-only-in-global-mem-ommd.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): surface ~/.claude/CLAUDE.md (read-only) in Global Memory
--- a/agentmux-srv/src/backend/rpc_types/commands.rs
+++ b/agentmux-srv/src/backend/rpc_types/commands.rs
@@ -230,13 +230,14 @@ pub const COMMAND_REORDER_GLOBAL_BRAIN: &str = "reorderglobalbrain";
 /// moment they see an existing `is_system=1` row.
 pub const COMMAND_UPSERT_SYSTEM_MEMORY: &str = "upsertsystemmemory";
 pub const COMMAND_DELETE_SYSTEM_MEMORY: &str = "deletesystemmemory";
-/// Read-only. Returns `~/.claude/CLAUDE.md` (Claude Code's own global
-/// user-level config file — independent of AgentMux, independent of any
-/// per-project CLAUDE.md) so the Armory Memory tab can show the operator
-/// where their real "highest priority" instructions actually live. No
-/// parameters (the path is fixed, not caller-supplied — no path-traversal
-/// surface) and no write counterpart. See
-/// docs/specs/SPEC_SURFACE_CLAUDE_GLOBAL_CONFIG_2026_08_24.md.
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Good catch — you're right, and this went deeper than a wording fix. Fixed in 5d8cbe961: resolves the path via `DataPaths::provider_auth_dir("claude")` (agent_open.rs's own default `auth_dir` fallback logic, mirrored exactly), which is the actual `CLAUDE_CONFIG_DIR` a non-identity-bound spawned agent gets. Also re-verified empirically: neither the default shared provider dir nor my own (identity-bound) CLAUDE_CONFIG_DIR has a populated CLAUDE.md on this host today, so the block will correctly show "No file at this path yet" until someone puts one there — accurate, not previously-misleading content. Badge copy updated to match ("Claude Code — shared provider config"). Identity-bound agents (separate per-identity dir) are still not covered — flagged as a known gap in the spec, not fixed here.

Full detail in docs/specs/SPEC_SURFACE_CLAUDE_GLOBAL_CONFIG_2026_08_24.md §5.

@codex review

GenericAgentX-asaf:

GenericAgentX-asaf:
Fixed the two remaining stale doc comments in b72ceb6f5 (commands.rs's COMMAND_GET_CLAUDE_GLOBAL_CONFIG constant, global-brain-model.ts's fetchClaudeGlobalConfig) — both still described the ambient path after the P1 fix.

@codex review

GenericAgentX-asaf:

PR #2796, opened 2026-08-25
Manoz@Area54: feat(widgets): reorder default pinned widgets to Agent, Swarm, Armory, Sysinfo

GenericAgentX-asaf:
## Summary
- Fresh installs now bootstrap into a 4-pane layout — agent on the left, swarm/armory/sysinfo stacked on the right (was agent/sysinfo/swarm, 3-pane) — via first launch, "Open another window", and new-tab creation (three independent mechanisms, all updated: backend `default_four_pane_tree`, frontend `DEFAULT_TAB_PRESET`, `widgets.json`'s `display:pinned`/`display:order`).
- Widget bar's default pinned set is now exactly `agent`, `swarm`, `armory`, `sysinfo` in that order. Drone/Warden/Media move to unpinned (More dropdown) by default, since the ask named exactly these 4 as *the* default set.
- Updated every doc referencing the old default: `docs/specs/widget-pinning.md`, `SPEC_SESSION_RESTORE_AND_SAVED_LAYOUTS_2026_08_13.md`, `SPEC_EDITOR_FILE_TREE_OPEN_ACTIONS_2026_07_12.md`, and root `CLAUDE.md`'s widget tier table (already flagged stale in `SPEC_WIDGET_BAR_PARENT_SUBMENUS_2026_08_12.md`'s own audit note — fixed here).
- New spec: `docs/specs/SPEC_DEFAULT_WIDGETS_REORDER_2026_08_25.md` documents the design/rationale and the exact mechanisms touched.

## Test plan
- [x] `cargo test --bin agentmux-srv` — 2797 passed, 0 failed
- [x] `npx vitest run` — 3120 passed, 1 pre-existing flake confirmed unrelated (passes in isolation, `tool-renderer registry` timeout under full-suite resource contention — nothing touched by this PR)
- [x] Manual: verified `widgets.json`'s pinned/order fields resolve to `agent, swarm, armory, sysinfo` via a Node one-liner


a5af:
@codex review

GenericAgentX-asaf:
Fixed the P2 sibling-ordering bug in `frontend/app/tab/tab-presets.ts:59` — good catch. `applyNode` was splitting every child after the first off the FIRST child instead of the PREVIOUS one, which is harmless for a 2-child split but broke the new 3-child swarm/armory/sysinfo split (sysinfo landed ahead of armory instead of after it).

Fix: track `previousId` alongside `firstId` and use it as the split target for i>0. Added `frontend/app/tab/tab-presets.test.ts` — traces the dispatched split targets for both a 3-child and 2-child preset; confirmed it fails against the pre-fix code and passes against the fix (`849fea77b`..`b1581fbea`).

GenericAgentX-asaf:
Fixed both P2 notes from the latest re-review — `tests.rs:1890` and `obj.rs:653` still said "three-pane" after the rename to `default_four_pane_tree`. Missed them in the original doc sweep since they're inline code comments, not docs. Verified no other stale "three-pane"/"3-pane" references remain in either file, ran the touched tests (`test_layout_node_serde_compat_with_frontend_shapes`, `layout_seeders_route_through_reducer_coherently`, `layout_stays_coherent_across_full_mutation_lifecycle`, `layout_seed_unknown_tab_errors`) — all pass (`7dfd85e5c`).

GenericAgentX-asaf:
The Ubuntu CI job's failure was real (unlike an earlier stale jekt notification) — `restore_last_session_survives_a_real_process_restart` in `agentmux-srv/tests/integration_test.rs` hardcoded the old 3-block default-seed assumption (asserted 4 total = 3 seed + 1 marker block). Fixed to expect 5 (4 seed + 1 marker), matching this PR's new 4-block default seed.

This lives in a separate `cargo test` target (`--test integration_test`) from `cargo test --bin agentmux-srv`, which is what I'd run earlier in this branch's local verification — that's why it slipped through locally. Verified directly: `cargo test -p agentmux-srv --test integration_test` — 5/5 pass (`8d41ca675`).

PR #2797, opened 2026-08-25
Manoz@Area54: feat(armory): show ambient ~/.claude/CLAUDE.md as a second Global Memory read-only block

GenericAgentX-asaf:
## Summary
- Adds a second, independently-fetched read-only block to Armory -> Memory (Global Memory), alongside the already-shipped shared-provider-config block (PR #2794): the ambient `~/.claude/CLAUDE.md`, badged "Claude Code — host CLI config (ambient)".
- The two blocks are clearly, distinctly labeled by audience to avoid repeating PR #2794's original mistake (implying one file covers more than it does): the shared-provider-config block is what a default spawned-in-app Claude agent reads; the new ambient block is what a host-level Claude Code CLI reads outside AgentMux's `CLAUDE_CONFIG_DIR` isolation.
- Scoped by explicit choice: only the ambient file was requested. Per-identity `CLAUDE_CONFIG_DIR` dirs and an external coding-agent harness's own working-directory files were both explicitly declined — the latter isn't part of AgentMux's spawned-agent memory pipeline at all and would misrepresent what the feature shows.
- Updated `docs/specs/SPEC_SURFACE_CLAUDE_GLOBAL_CONFIG_2026_08_24.md` (§6) with the full design/rationale.

## Test plan
- [x] `cargo test --bin agentmux-srv` — 2797 passed, 0 failed (new handler reuses the already-tested `read_claude_global_config` helper against a different directory — no new Rust test needed, existing coverage applies to both call sites, noted inline)
- [x] `npx vitest run frontend/app/view/brain/` — 25/25 pass, including new tests for atom independence (a rejected ambient fetch doesn't affect the shared-provider-config atom and vice versa) and both blocks rendering with distinct badges
- [x] `npx vitest run` (full suite) — 3130 passed, 1 pre-existing flake confirmed unrelated (passes in isolation, `tool-renderer registry` timeout under full-suite resource contention)

### DIFF
--- a/.changesets/1787663234-feat-armory-show-ambient-claude-claude-md-as-a-second-global-xzwd.md
+++ b/.changesets/1787663234-feat-armory-show-ambient-claude-claude-md-as-a-second-global-xzwd.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): show ambient ~/.claude/CLAUDE.md as a second Global Memory read-only block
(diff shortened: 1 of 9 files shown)


a5af:
@codex review

PR #2798, opened 2026-08-25
docs(retro): confirm cross-agent GitHub identity binding in AgentY's PR #5 retro

AgentX-asaf:
## Summary
AgentY's retro (`retro-agentmux-corp-pr5-shared-gh-cli-session-misattribution-2026-08-24.md`) root-caused a PR misattribution to a possible cross-agent identity binding, but needed AgentX's own `IdentityAccounts` output to confirm §2a's hypothesis — that's me, so I ran it.

Confirmed: my own `IdentityAccounts` contains the exact same account (`"AgentX GitHub"`, `account_id: 15f7fe0a-7827-4c27-8456-f08da8df9ae5`) that showed up injected into AgentY's shell as their ambient `GH_TOKEN`. Not a reciprocal pair of mis-bindings as §2a's framing assumed — it's a single identity-provider account object bound to both agent contexts simultaneously.

This confirms it's a real bug in the identity-resolver's account-binding logic (`identity/resolver/{inject.rs,provider.rs}`), not a workflow-discipline issue on either agent's part. Full detail in the new §7 of the retro.

## Test plan
- [x] Doc-only change, no code touched
- [x] Ran the exact verification step the retro's own §6 asked for (`IdentityAccounts` on my own agent identity)

### DIFF
--- a/docs/retro/retro-agentmux-corp-pr5-shared-gh-cli-session-misattribution-2026-08-24.md
+++ b/docs/retro/retro-agentmux-corp-pr5-shared-gh-cli-session-misattribution-2026-08-24.md
@@ -256,21 +256,38 @@ side.** §2a's framing ("if AgentX's own identity bindings reciprocally
 include an account bound to AgentY's GitHub credential") assumed a pair of
 mirrored mis-bindings; what's actually there is a **single identity-provider
 account object, named "AgentX GitHub," visible/bound in both agents'
-`IdentityAccounts` lists simultaneously** — arguably a cleaner confirmation
-of a real resolver bug than a symmetric pair would have been, since there's
-no ambiguity about which agent "owns" it by name — it's mine by name, and
-AgentY's environment got it anyway.
+`IdentityAccounts` lists simultaneously**.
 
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentX-asaf:
Good catch, and correct. Fixed in c00ed3e1c.

Checked the code you cited directly: `IdentityAccount` is explicitly documented as a reusable credential, `agent_identity_link`'s uniqueness constraint is scoped to `(agent_id, provider)` only (nothing stops one `account_id` linking to multiple agents), and `identity_delete_captures_all_affected_agents` is a dedicated test proving one-account-multiple-agents is deliberate, tested behavior — not an edge case that slipped through.

Corrected §7 to walk back "confirmed resolver bug." The shared-account observation itself still stands (AgentY really does have a link to the same account I do), but what's actually open is how/why that specific AgentY→"AgentX GitHub" link was created — deliberate shared-credential setup vs. accidental binding — not whether the resolver malfunctioned. It didn't; it injected exactly what an existing link said to.

PR #2800, opened 2026-08-25
feat(agent-pane): hover-to-peek fires on every transcript node kind, 50ms delay [Naki@claudius]

GenericAgentX-asaf:
## Summary
- Widens the Aug-3 hover-peek feature (`SPEC_TRANSCRIPT_NODE_HOVER_PEEK_2026_08_03.md`, previously scoped to just tool calls, thinking clumps, and user input) to cover every remaining transcript node kind — `agent_message`, `jekt_message`, `shell`, `section`, `agent_error`, `context_compacted`, `compaction_started`, `day_divider`, `session_outcome`.
- Lowers the hover enter-delay from 150ms to 50ms, centralized into one shared constant (`hover-anchor.ts`) instead of three independently-declared copies.
- Factors the duplicated isPeeking/timer boilerplate into a shared `useNodePeek()` hook.
- Widens ToolBlock's and MarkdownBlock's existing anchors so the peek fires from anywhere in the row, not just a narrow inner span/thinking-only block.
- One deliberate exception: `history_link` (a synthetic CTA row with no timestamp/content field) gets no peek — there's no data to show.

See `docs/specs/SPEC_TRANSCRIPT_NODE_HOVER_PEEK_ALL_KINDS_2026_08_25.md` for full design rationale, including why this doesn't repeat the two prior "generic hover strip" attempts that got ripped out for UX complaints (this is a read-only floating overlay, not a layout-affecting expand/collapse).

## Test plan
- [x] `tsc --noEmit` clean
- [x] `vitest run frontend/app/view/agent`: 1426/1426 passing (115 files) — includes new tests for the hook, 3 new component test files, and 7 new DocumentRow cases
- [x] `vitest run frontend/app/view/swarm`: 90/90 passing
- [x] `task build:frontend` production build succeeds
- [ ] Interactive/visual verification not done — native desktop app, this machine runs shared live instances; see spec for details

### DIFF
--- a/.changesets/1787690356-feat-agent-pane-hover-to-peek-now-fires-on-every-transcript--zhb1.md
+++ b/.changesets/1787690356-feat-agent-pane-hover-to-peek-now-fires-on-every-transcript--zhb1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): hover-to-peek now fires on every transcript node kind, 50ms delay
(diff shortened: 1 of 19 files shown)


a5af:
@codex review

PR #2804, opened 2026-08-26
fix(splash,agent-pane): darkened splash border + agent-pane tab-strip gap/centering

AgentY-asaf:
## Summary

Two independent, repo-owner-requested UI tweaks, each with its own spec doc:

**1. Splash screen — 2px darkened border, all 3 platforms**
(`docs/specs/SPEC_SPLASH_SCREEN_BORDER_2026_08_25.md`)

No shared implementation — each platform draws the splash natively, so the border is implemented three times:
- `splash.rs` (Windows): new border pass in the GDI DIB compositor, drawn last so nothing overdraws it.
- `splash_mac.rs`: native `CALayer` `setBorderWidth:`/`setBorderColor:`, trivial alongside the existing `setCornerRadius:`.
- `splash_linux/mod.rs`: a second, antialiased coverage band (inner vs. outer rounded-rect) in the existing per-pixel `render_frame` compositor.

Color: `#0D0D10` (roughly half the `#1A1A1F` backdrop) — a starting proposal per the spec, not yet confirmed on macOS/Linux displays (only Windows is verifiable from this machine).

**2. Agent pane tab strip — remove the reserved gap; fix "+" centering**
(`docs/specs/SPEC_AGENT_PANE_PROGRESS_BAR_OVERLAY_NO_GAP_2026_08_25.md`)

The progress bar's always-present 3px flex row (from `SPEC_AGENT_PANE_PROGRESS_BAR_ABOVE_TAB_STRIP_2026_08_10.md`) pushed the tab strip 3px below the pane header. Took the slot out of flex flow entirely (`position: absolute`, overlaying the tab strip's top edge) instead of reserving space — preserves that spec's original no-layout-shift guarantee via a different mechanism, not a blind reversal (see the new spec's §2.1).

Fixing the gap surfaced two smaller, previously-masked "+"-centering issues, both confirmed via live pixel-sampling of a running `task dev` build:
- `.pane-tab-strip`'s `border-bottom` was eating 1px out of its own declared height under this repo's global `box-sizing: border-box`, clipping `.pane-tab-strip-inner`'s bottom edge. Height now explicitly accounts for the border.
- The "+" glyph (literal text, not an SVG icon) sits ~1-2px low within its own flex-centered line box — a font-metrics artifact. Wrapped in its own `<span>` and nudged via `transform: translateY`, independent of the button's own box/border/hover-background.

## Test plan
- [x] `cargo check -p agentmux-launcher` — Windows path only; `splash_mac.rs`/`splash_linux` are `cfg`-gated to their own OS and can't be typechecked from this machine (flagged in both specs and the commit message)
- [x] `PaneTabStrip.test.tsx` — 17/17 passing
- [x] Full `frontend/app/view/agent/` suite — 1403/1404 (the one failure is a pre-existing flake, confirmed passing in isolation, unrelated to this change)
- [x] Live visual verification via `task dev` — repo-owner confirmed the tab-strip gap is gone and the "+" centering looks right after the fixes; splash border not yet visually confirmed on a real display (flagged as an open item in the spec)

### DIFF
--- a/.changesets/1787711826-fix-splash-agent-pane-darkened-splash-border-agent-pane-tab--l5mx.md
+++ b/.changesets/1787711826-fix-splash-agent-pane-darkened-splash-border-agent-pane-tab--l5mx.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(splash,agent-pane): darkened splash border + agent-pane tab-strip gap/centering
--- a/agentmux-launcher/src/splash.rs
+++ b/agentmux-launcher/src/splash.rs
@@ -99,6 +99,14 @@ const BG_B: u8 = 0x1F;
 const BG_G: u8 = 0x1A;
 const BG_R: u8 = 0x1A;
 
+// Darkened window-edge border — SPEC_SPLASH_SCREEN_BORDER_2026_08_25.md.
+// Roughly half BG's RGB values, a straightforward "darker than the
+// backdrop" reading; not yet confirmed against a real display.
+const BORDER_B: u8 = 0x10;
+const BORDER_G: u8 = 0x0D;
+const BORDER_R: u8 = 0x0D;
+const BORDER_WIDTH: i32 = 2;
+
 static BRAIN_BGRA: &[u8] =
     include_bytes!(concat!(env!("OUT_DIR"), "/brain_bgra.bin"));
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #2805, opened 2026-08-26
fix(agent): scroll support for AskUserQuestion panel

AgentX-asaf:
## Summary
Implements `docs/specs/SPEC_ASK_USER_QUESTION_PANEL_SCROLL_2026_08_25.md`.

`AgentQuestionPanel` had no `max-height`/`overflow` anywhere — it grows unbounded inside `.agent-view`'s `overflow: hidden` flex column. A multi-question or many-option `AskUserQuestion` call could push the Submit/Answer-later buttons (and the queue/composer below the panel) out of view with no way to scroll to them. Not just cosmetic: the 30s auto-timeout keeps running regardless, so an unreachable Submit button means the user can't act early or defer — only wait on the timer.

- Wraps the question content in a new `.agent-question-panel-scroll` region (`overflow-y: auto`, `min-height: 0` — the classic flexbox scroll gotcha, called out in a code comment so it survives a later refactor).
- Header and actions bar become `position: sticky` top/bottom, following the same pattern already used by `.agent-search-header` (`_search.scss`).
- Fixed `max-height: 420px`, not `vh`/`%` — `.agent-view` only supports width-based container queries, and `vh` is relative to the OS window, not this specific split pane (could be much shorter in a multi-pane layout).

## Test plan
- [x] 35/35 tests passing (`AgentQuestionPanel.test.tsx`) — 33 existing unchanged, 2 new asserting the scroll region is a sibling of the sticky header/actions (not nested under either) and that Submit/Answer-later are never inside the scroll region.
- [ ] Live visual verification via `task dev` was attempted but the build stalled on a CEF/Rust link step (~25 min, no progress) and was killed rather than block further. Relying on the passing structural tests plus code review for this narrowly-scoped, well-established CSS pattern — flagging explicitly rather than claiming a check that didn't happen.

### DIFF
--- a/.changesets/1787712548-fix-agent-scroll-support-for-askuserquestion-panel-keep-butt-slgv.md
+++ b/.changesets/1787712548-fix-agent-scroll-support-for-askuserquestion-panel-keep-butt-slgv.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): scroll support for AskUserQuestion panel, keep buttons reachable
--- a/docs/specs/SPEC_ASK_USER_QUESTION_PANEL_SCROLL_2026_08_25.md
+++ b/docs/specs/SPEC_ASK_USER_QUESTION_PANEL_SCROLL_2026_08_25.md
@@ -1,7 +1,8 @@
 # SPEC: Scrolling for the AskUserQuestion panel
 
 **Date:** 2026-08-25
-**Status:** design only, not yet implemented
+**Status:** implemented, PR #2805 — see §3.1's two corrections (Codex P1/P2)
+made during that PR's review, before merge.
 **Builds on:** `docs/specs/SPEC_ASK_USER_QUESTION_2026_06_15.md` (original
 design), `SPEC_ASK_USER_QUESTION_AUTO_TIMEOUT_2026_08_06.md` /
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentX-asaf:
Both real, fixed in bff91ff9b.

1. (Codex, this thread) `max-height: 420px` on `.agent-question-panel` was ineffective once content exceeded it — as a flex item of `.agent-view`, the panel's automatic minimum height (content's min-content size, since it has no `overflow` set) wins over `max-height` per the flex sizing algorithm. Added `min-height: 0` on the panel itself.
2. (reagent, this thread) `position: sticky` on the header/actions was inert — both are siblings of `.agent-question-panel-scroll`, not nested inside it, and `.agent-view` never scrolls, so there's no scroll container for sticky to attach to. The `_search.scss` comparison in my own comment was wrong (that header genuinely is nested in its scroll container). Removed the dead sticky/z-index declarations — `flex-shrink: 0` on the header/actions was already doing the real work.

Spec updated with both corrections rather than silently rewritten. 35/35 tests passing.

AgentX-asaf:
Good catch — fixed in 1e9b47233, missed this one when correcting the SCSS comments. 35/35 tests still passing.

PR #2808, opened 2026-08-26
fix(agent-pane): composer strip zone balance — real DOM measurement (resolves #2809)

### LINKED ISSUE #2809: Composer strip: left zone still needs 2 lines while right fits in 1 (zone assignment confirmed correct)
## Summary

Follow-up work needed on the agent composer strip's responsive left/center/right zone layout. Two real bugs were already found and fixed in PR #2808 (draft) this session; this issue tracks the **third, still-open** problem that PR explicitly does not resolve.

Full history: `docs/specs/SPEC_COMPOSER_STRIP_DYNAMIC_BALANCE_2026_08_24.md` (Rev 1-5). Full diagnostic writeup: `docs/status/STATUS_COMPOSER_STRIP_ZONE_BALANCE_HANDOFF_2026_08_25.md` (both on branch `fix/composer-strip-left-right-balance`, PR #2808).

## The problem

In the common case (Claude agent, context tracked, HOST mode, logged in), the strip renders as 2 uneven lines instead of 1:

```
┌──────────────────────────────────────────────────┐
│ Bypass · Sonnet 5 · high     181k / 200k ⚠ ~0 to  │
│ auto-compact                                       │
│ [Compact]                        ● Logged in HOST [Shell] │
└──────────────────────────────────────────────────┘
```

Real screenshot:

lopsided

## Diagnostic finding (important — confirms this is NOT a zone-assignment bug)

Temporary debug outlines were added to all 3 zones (red = controls/left, blue = stats/center, green = right — still in the code as of PR #2808, marked `TEMPORARY DEBUG — outline1s0824x`) to definitively confirm which container each element renders in:

outlined

This confirms: the context group (ctx text + countdown + Compact) is entirely inside the left/red zone, and auth + HOST/Shell are entirely inside the right/green zone — nothing is torn across zones. The actual problem is that the **left zone's total content is wider than one line can hold** at common pane widths (runtime trigger + all 3 context-group sub-elements together don't fit, so Compact wraps to the left zone's own second internal line), while the right zone's content fits on one line. Net: 2 visual lines, left zone occupying both, right zone only the second — reads as imbalanced even though neither zone is empty and nothing is clipped.

## Why this has resisted fixing (read before trying another variant)

Every fix so far (this session had 5 revisions) targeted **which slot goes in which zone** or **how wide each zone's box is** as a fixed, hand-authored decision. The actual constraint — total rendered width of the left zone's content should be close to the right zone's — can only be known from real content, not a semantic pairing chosen at design time. Two earlier attempts at computing balance algorithmically (count-based split, weight-balanced subset-partition search over a hand-guessed integer weight) were themselves buggy and got reverted — the idea of computed/measured balance may still be right, it just needs correct execution.

## Options for the next attempt (see status doc for full detail)

1. **Real DOM measurement** — measure each slot's actual rendered width post-paint, redistribute based on real widths instead of a fixed pairing or guessed weight. Most likely to work; adds a measure-then-layout pass.
2. **Relax the context-group adjacency constraint** — currently ctx text + countdown + Compact must ALL stay together as one atomic unit; confirm whether that's a hard requirement or just how it was built, since splitting it could open more balancing options.
3. **Re-confirm the actual bar with the user** — a consistent 2-line layout might be acceptable if "up to 3 lines, never empty" was the real requirement; much of this session's difficulty came from each fix changing the *shape* of the imbalance in a way that read as "still broken."

## How to reproduce / test

```
git checkout fix/composer-strip-left-right-balance
npm run dev
```

Launches a `task dev` build with the debug outlines already in place. See `docs/reports/REPORT_AGENT_SCREENSHOT_WINDOW_CONTROL_BLOCKERS_2026_08_24.md` for known rough edges in this dev-build flow (unrelated but discovered same session) — notably, `mcp__agentmux__Shell` doesn't reliably work for launching this specific process, and HMR should not be trusted alone to confirm a fix landed (verify via a full kill+relaunch + direct `curl` of the dev server's served output, not just the HMR log).

## Related

- PR #2808 (draft) — the 2 bugs already fixed this session, plus the debug outlines and full diagnostic docs this issue references.

AgentY-asaf:
## Summary

Continuation of same-day composer-strip layout work (see `docs/specs/SPEC_COMPOSER_STRIP_DYNAMIC_BALANCE_2026_08_24.md` for full Rev 1-6 history). Originally opened as a draft handoff with 2 bugs fixed and a 3rd tracked as issue #2809 — **this PR now also fixes #2809**, closing it.

**Fixed (Rev 4/5, unchanged from the original draft):**
- **Rev 4:** `_composer-strip.scss`'s widest tier forced both edge zones to exactly equal width regardless of content, causing real dead space in whichever zone had less. Removed — the first attempt at removing it was itself incomplete (a leftover `flex-basis: 100%` base rule kept forcing the right zone onto its own full line even at the widest tier). Both are fixed.
- **Rev 5:** the left/right slot grouping was lopsided by content weight even with the CSS bug fixed. Rebalanced to a fixed pairing (context group left, badge/auth/HOST+Shell right) — still not enough on its own (see Rev 6).

**Fixed (Rev 6, new in this update — resolves #2809):**
- Even with Rev 4/5's fixes, the strip still needed 2 lines in the single most common case (Claude agent, context tracked, HOST mode) — no *fixed* pairing can be right for every combination of which slots are present and how wide their content is.
- Replaced the fixed `side` pairing with `computeBalancedLeftKeys` — a real DOM-measurement-based search (each slot's rendered width, measured via a `display: contents` ref wrapper with zero layout impact) that picks whichever left/right split minimizes width difference. `hostShell` (the Shell toggle) stays pinned right — a stable, predictable position for the strip's one real action. Falls back to the original fixed pairing until the first measurement lands or in any environment with no real layout engine (this file's own JSDOM-based unit tests) — the existing 8 tests needed no changes.
- **Visually verified** in a real `task dev` build via direct Win32 screenshot capture (not `CaptureWindow` — bypasses its own-instance/ambiguous-title hazards for a same-machine capture; not assumed from HMR logs). At the ≥482px tier the strip now renders on one line with a genuinely rebalanced grouping; narrower tiers still degrade to 2-3 readable lines with nothing clipped.
- Removed the temporary debug outlines (`outline1s0824x`) left in place for this handoff.

See `docs/specs/SPEC_COMPOSER_STRIP_DYNAMIC_BALANCE_2026_08_24.md`'s Rev 6 section and `docs/status/STATUS_COMPOSER_STRIP_ZONE_BALANCE_HANDOFF_2026_08_25.md` (now marked resolved) for full detail.

Closes #2809.

## Screenshots

Real captures from `task dev`, cropped to the composer strip only:

**Before this PR's fixes (Rev 3 state — 2 lines, right zone overflowing):**

before

**Rev 4/5 state (2 lines, left zone overflowing instead — see the handoff status doc):**

lopsided

**After Rev 6 (this update) — single line, real screenshot, outlines removed:** confirmed locally via a `task dev` build; not re-committed as a file since it was captured against this session's own live (and privacy-sensitive) transcript content — see the status doc's "How the fix was verified" section for the exact repro method if this needs re-confirming.

## Test plan

- [x] `npx vitest run AgentComposerStrip.test.tsx` — 12/12 pass (8 original + 4 new `computeBalancedLeftKeys` unit tests)
- [x] `npx stylelint _composer-strip.scss` — clean (2 pre-existing unrelated warnings only)
- [x] Full frontend suite (`npm test`) — 213 files / 3139 tests passed, 2 pre-existing skips
- [x] Visual fix for the remaining 2-line imbalance — confirmed via real screenshot (see above)
- [x] Removed temporary debug outlines


a5af:
@codex review

PR #2810, opened 2026-08-26
feat(agentmux-mcp): DiscoverWindows tool + pid-based CaptureWindow targeting

AgentY-asaf:
## Summary

Adds robustness to the cross-instance window discovery/capture MCP tools, grounded in real blockers hit during a live session (`docs/reports/REPORT_AGENT_SCREENSHOT_WINDOW_CONTROL_BLOCKERS_2026_08_24.md`). Implements Phase 1 of `docs/specs/SPEC_AGENT_APP_API_WINDOW_CONTROL_ROBUSTNESS_2026_08_24.md`.

- **New `DiscoverWindows` tool** — read-only enumeration of every AgentMux-owned window (pid, title, exe_path, is_self), no screenshot taken, nothing written to disk, nothing audit-logged. Lets a caller see real candidates before calling `CaptureWindow`, instead of guessing `index` values.
- **`CaptureWindow` accepts `pid`** as an alternative to `title_contains`. Titles on this app are actively rewritten by its own frontend (tab switches, workspace renames, even unrelated windows opening/closing elsewhere trigger a re-render) — a pid is stable for a process's whole lifetime and doesn't have that problem.
- **Fixed a real bug:** an ambiguous `title_contains` match with no explicit `index` used to silently capture index 0 instead of surfacing the candidate list — despite the tool's own docstring already claiming it would list matches when ambiguous. Now it actually does.
- **Blank-frame retry:** a short bounded retry (up to 2x, ~800ms) when a capture looks like a solid/near-solid-color frame, plus a `likely_unrendered` hint in the result either way.

Phase 2 (a real fix for cross-instance window renaming, via a local instance-discovery broker) and an open Shell-tool reliability mystery (exit code 201, root cause not found) are documented as follow-ups in the spec — not attempted here.

## Test plan

- [x] `cargo build -p agentmux-mcp` clean
- [x] `cargo test -p agentmux-mcp` — 16/16 pass (3 new tests for the blank-frame heuristic, 2 existing tests updated for the new audit-log field/tool count)
- [x] `cargo clippy -p agentmux-mcp --all-targets` — zero new warnings (2 pre-existing warnings in unrelated code, confirmed unchanged by this diff)
- [x] Manually verified via direct `curl` against a running dev server that the served code matches source (used repeatedly during the session this was built for)

### DIFF
--- a/.changesets/1787718976-feat-agentmux-mcp-discoverwindows-tool-pid-based-capturewind-k9ka.md
+++ b/.changesets/1787718976-feat-agentmux-mcp-discoverwindows-tool-pid-based-capturewind-k9ka.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agentmux-mcp): DiscoverWindows tool + pid-based CaptureWindow targeting
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -263,20 +263,20 @@ const UI_QUERY_TOOL: &str = r#"{
 // something to back into accidentally via a narrower tool's scope creep.
 const CAPTURE_WINDOW_TOOL: &str = r#"{
   "name": "CaptureWindow",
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2811, opened 2026-08-26
fix(tab): eliminate tab-close flash — click containment, single-RPC close, batched WaveObj updates (WS + response paths)

AgentY-asaf:
## Summary
- Clicking a background tab's ✕ close button let the `click` event bubble past the button into the parent tab's `onClick={onSelect}` (only `mousedown` propagation was stopped, not `click`), firing a spurious `SetActiveTab` for the tab that was simultaneously being closed.
- That raced an unordered `SetActiveTab` request against the `CloseTab` request over independent `fetch()` calls; depending on which the backend serviced first, the tab bar would flash-highlight the closing tab and then jump again as `DeleteTab`'s own neighbor-promotion reassigned active — the reported "flash of another tab selected and then deselected." In the worst case it could also silently leave the wrong tab active.
- Fix: stop `click` propagation on the close button the same way `mousedown` already is, so the spurious select never fires and no race is created.
- Full root-cause writeup: `docs/specs/SPEC_TAB_CLOSE_BUTTON_SELECT_FLASH_2026_08_25.md`.

## Test plan
- [x] `npx vitest run` — 3132 passed, 3 skipped (full suite, no regressions)
- [ ] Manual: open 3+ tabs, click a background tab's ✕ repeatedly and confirm the active tab's highlight never flashes/jumps

### DIFF
--- a/.changesets/1787786849-fix-tab-eliminate-tab-close-flash-batch-multi-object-ws-upda-c877.md
+++ b/.changesets/1787786849-fix-tab-eliminate-tab-close-flash-batch-multi-object-ws-upda-c877.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tab): eliminate tab-close flash — batch multi-object WS updates into one frame, emit parent-before-child on deletes
--- a/agentmux-srv/src/backend/eventbus.rs
+++ b/agentmux-srv/src/backend/eventbus.rs
@@ -16,6 +16,12 @@ use super::wps::{WaveEvent, WpsClient, EVENT_SYS_INFO, EVENT_BLOCK_STATS, EVENT_
 
 pub const WS_EVENT_RPC: &str = "rpc";
 
+/// One WS frame carrying an ARRAY of `WaveObjUpdate`s from a single atomic
+/// backend transition. Mirrored in `frontend/app/store/wps-events.ts`
+/// (`WpsEvent.WaveObjBatchedUpdates`) — the frontend applies the whole array
+/// in one Solid `batch()` flush. See `broadcast_wave_obj_updates` below.
+pub const WS_EVENT_WAVE_OBJ_BATCHED_UPDATES: &str = "waveobj:batchedupdates";
+
 /// Egress priority lane for a server→client event.
 ///
 /// `Background` is reserved for droppable perf telemetry (sysinfo + per-block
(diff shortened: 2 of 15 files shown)


a5af:
@codex review

AgentY-asaf:
**Why the flash survived the §6 batch fix — and the real fix (b7cde2a13):**

The repo owner reported the flash still visible right after the close-confirm modal dismisses. Root cause: a `CloseTab`'s `[delete tab, update workspace]` pair reaches the calling renderer through **three** paths, and §6 only batched the slowest one:

1. **Wave-obj bridge** — two separate WS frames, *tab delete first*, fired while the HTTP handler is still running.
2. **`run_service_call` response broadcast** — one WS frame *per update* (delete first), fired before the response body returns. It reaches the calling renderer too, not just "everybody else."
3. **HTTP response body** — the only path §6 batched. Arrives last; its updates are version-guarded to a no-op because paths 1-2 already applied.

`forward_event` (websocket.rs) wraps each raw frame as `eventrecv`, so each lands in `global.ts`'s scope-less handler as a bare, per-frame `updateWaveObject()` — no `batch()`. The delete-first frame blanks the still-mounted `<Tab>` a full paint before the workspace update unmounts it. Deterministic, not a race.

Fixes in this commit:
- `EventBus::broadcast_wave_obj_updates` — a response's whole updates array goes out as ONE `waveobj:batchedupdates` frame (all five multi-update broadcast sites switched); frontend applies it via the already-batched `updateWaveObjects`.
- Bridge delete ordering — `TabDeleted`/`BlockDeleted`/`SrvWindowClosed` now emit the **parent update before the child delete** (create child-first, delete parent-first), so the two genuinely-separate bridge frames can't paint a blank child.

Review feedback addressed: changeset added (reagent P1), §4 close-click component test added — `frontend/app/tab/tab.test.tsx` (reagent P2), spec status header corrected (codex P3). Spec §7 documents the full trace. Tests: 2 new Rust broadcast-order tests + 1 batched-frame test + 2 component tests; full `cargo test -p agentmux-srv` (2816 passed) and store/tab vitest suites (790 passed) green, `tsc --noEmit` clean.

Note for verification: no build the owner currently has contains ANY of this PR's fixes — v0.55.25 was cut from main before merge. A fresh portable off this branch (or post-merge main) is needed to confirm visually.

PR #2812, opened 2026-08-26
Manoz@Area54: fix(agent-pane): row-based composer strip layout so every line has both a left and right occupant

GenericAgentX-asaf:
## Summary

Follow-up to #2808 — Rev 6's zone-based composer strip fix shipped with a real structural flaw: `.agent-composer-strip-controls` (left) and `.agent-composer-strip-right` were two **independently-wrapping** flex containers, each free to claim a full dedicated line on its own. That can produce "2 lines, 2 filled sections" (one line entirely left-content, one entirely right-content) instead of the actual requirement — every rendered line should have both a left-anchored and right-anchored occupant (2 lines → 4 sections).

This PR replaces the zone model with a **row model**:
- `computeComposerRows(slots, hostShellKey, availableWidth, gapPx)` — a new pure function. If everything fits on one line, delegates to the existing `computeBalancedLeftKeys` unchanged. Otherwise sorts all slots descending by width and pairs them widest-with-narrowest via a two-pointer walk (one row per pair), reorienting `hostShell`'s pair so it's always the right occupant of the last row.
- A new `ResizeObserver`-backed `stripWidth` signal — every prior revision relied purely on CSS `@container` queries; this is the first real pixel measurement needed to decide single-row vs. multi-row in JS.
- Render output rewired to iterate `rowIndices()` (plain number primitives) rather than `rows()` directly, extending the reagent-P1 identity-stability fix from #2808 one level up (rows, not just slots).

**A real integration bug found during verification, not just the algorithm:** narrow-tier screenshot testing showed the row-decision logic still going wrong under a `zoom: 0.8` ancestor — `ResizeObserver`'s `entry.contentRect.width` reports local (pre-zoom) CSS pixels, while per-slot widths use `getBoundingClientRect()` (post-zoom, viewport-relative). That unit mismatch made the "does everything fit on one line" check systematically over-generous, silently deciding single-row and letting the row's own `flex-wrap` safety net reproduce the one-sided-lines pattern via a different mechanism. Fixed by reading `entry.target.getBoundingClientRect().width` instead, so both sides of the comparison share a coordinate space regardless of ancestor zoom.

Full analysis of how the original bug went undetected across 6 revisions: `docs/retro/retro-composer-strip-one-sided-lines-misdiagnosis-2026-08-26.md`. Full design: `docs/specs/SPEC_COMPOSER_STRIP_ROW_BASED_LAYOUT_2026_08_26.md` (§10 has a phase-by-phase progress log, including the zoom-mismatch finding).

## Test plan
- [x] 26 unit tests for `computeComposerRows` (single-row delegation, multi-row even/odd, `hostShell` reorientation, degenerate 1-slot case, empty pool, and a parameterized invariant test across 6 width combinations)
- [x] Existing component regression tests (identity stability, predictive countdown, etc.) all still pass
- [x] `npx tsc --noEmit`, `npx vitest run` (3153 tests), `npx stylelint` all clean (only 2 pre-existing grandfathered hex-color warnings, unrelated to this change)
- [x] Real screenshot verification, wide tier (≥482px) — single row, matches Rev 6's already-correct behavior
- [x] Real screenshot verification, narrow tier (<482px), via a live dev build inspected over CDP — confirmed 2 correctly-paired rows after the zoom-mismatch fix (the tier Rev 6 never re-checked, and the exact gap this whole redesign exists to close)


a5af:
@codex review

GenericAgentX-asaf:
Pushed 37acbe2d addressing all review findings:

**ReAgent P1** (`stripWidth` includes padding) and **P2** (wrong gap CSS var) — fixed by observing `.agent-composer-strip-rows` (no padding of its own) instead of the outer padded strip, and reading `--space-2` (the row's real `column-gap`) instead of `--space-1-5`.

**Codex P1 (shed slots)** — zero-width slots (hidden via this file's own SCSS shed-content queries) are now excluded from `computeComposerRows`'s pairing input and folded back into the last row afterward via their fallback `side`, so they can no longer consume a pairing partner while rendering nothing. Still mounted for continuous re-measurement.

**Codex P1 (stats zone in fit check)** — `computeComposerRows` now takes a `reservedWidth` param, fed by a real measurement of the stats zone, added to the single-row fit total.

**Codex P1 (per-pair capacity)** — the two-pointer pairing walk now checks `sorted[i].width + sorted[j].width + gapPx <= availableWidth` before pairing; when it doesn't fit, the wider slot falls back to its own one-sided row instead of an overflowing pair. Documented as a third "physical-capacity exception" in the spec (`docs/specs/SPEC_COMPOSER_STRIP_ROW_BASED_LAYOUT_2026_08_26.md` §1, §3.2).

2 new pure-function tests added for these cases; 5 existing tests recalibrated from an artificial `availableWidth=10` to realistic values (that constant was only ever meant to force the multi-row path, not exercise per-pair capacity — it now does both correctly).

Re-verified live via CDP against a running dev build: still 2 correctly-paired rows at the narrow tier after all fixes, no regression from the original task #48 verification.

One honest caveat: the shed-slot fix (auth/badge hiding below 220px/180px container width) couldn't be re-verified via a real narrow screenshot in this session — the dev-build window hit an app-enforced ~380px minimum width before reaching that threshold. The fix mirrors this file's own established measurement pattern (real DOM widths, not guesses), and all other integration points were re-verified live, but flagging this specific gap rather than overclaiming full coverage.

GenericAgentX-asaf:
Pushed fee2e0a7 addressing the fresh finding:

**ReAgent P1** (`statsWidth` stale across the single/multi-row transition) — the measurement effect now also depends on `stripWidth()`/`slotWidths()`, the actual root inputs that decide `rows().length`'s branch, so it re-measures whichever `statsRefs` variant is currently connected whenever a resize could plausibly flip that decision. Deliberately does NOT depend on `rows()` or `statsWidth()` itself — that would be circular, since `rows()` reads `statsWidth()` via `reservedWidth`.

Verified via `tsc --noEmit` and the full test suite (28/28 composer-strip tests, no hang/timeout that a reactive cycle would produce). One honest caveat, same spirit as the shed-slot one earlier: reproducing the exact failure condition live (a populated stats zone from an active turn, crossing the resize threshold) needs a live agent turn in progress, which I didn't set up given the fix itself is a straightforward, clearly-correct dependency widening rather than a new mechanism — flagging instead of overclaiming a live repro I didn't do.

Also re-verified the earlier shed-slot fix (Codex P1) live via a CDP `Emulation.setDeviceMetricsOverride` viewport override per your own suggestion, since the dev-build window's ~380px app-enforced minimum blocked a real OS-resize test: at 150px emulated width, `.agent-composer-strip-auth` correctly goes `display:none`, and folds into the last row's right side contributing 0 visible width rather than an empty row — exactly 1 one-sided row remained (the mathematically-allowed singleton for the 3 real pairable slots at that width), confirmed via `getBoundingClientRect()` on each row's left/right span, not just `textContent` (which reads hidden elements' text too and would have given a false negative).

PR #2813, opened 2026-08-26
Manoz@Area54: fix(agent-pane): composer strip zoom/gap unit mismatch + stale shed measurement

GenericAgentX-asaf:
Follow-up to #2812 — a real regression found via live user testing within minutes of that PR merging: the widest tier started splitting into 2 lines, and narrower tiers grew genuinely empty (both-sides-zero-width) rows. Both are the exact one-sided-lines pattern the whole row-based rewrite exists to prevent, reintroduced by two bugs in #2812's own review-response fixes:

**1. Zoom/gap unit mismatch.** #2812's ReAgent P2 fix read `--space-2` via `getComputedStyle()`, which returns the LOCAL (unzoomed) CSS pixel value — but every width it's compared against (`stripWidth`, slot widths) is measured via `getBoundingClientRect()`, which IS zoom-scaled. Under the agent view's `zoom: 0.8`, an "8px" gap actually renders as 6.4px — using 8 systematically overestimated how much space was needed. This explains both symptoms: the widest tier's single-row fit check failing when it shouldn't, and the per-pair capacity check (also from #2812) rejecting more pairs than necessary as the pane narrowed. Added a `zoomRatio()` helper and applied it everywhere a `getComputedStyle` length mixes with a `getBoundingClientRect` one — including a second, pre-existing instance of the same bug class in the per-slot internal-gap measurement (from #2808, never previously flagged).

**2. Stale shed-slot measurement on resize.** The per-slot width measurement effect only depended on `slots()` (prop-driven: `processCount`, `authStatus`, etc.), never on `stripWidth()` (resize-driven). A slot hidden by this file's own CSS shed-content queries (`.agent-composer-strip-auth`/`-process-badge` going `display:none` below a container-width threshold) is a case where a slot's rendered width IS a function of container width, not just content — a pure resize crossing that threshold never re-triggered remeasurement, so `slotWidths` kept the slot's pre-shed (nonzero) width indefinitely. #2812's own Codex P1 shed-slot fix (excluding zero-width slots from pairing) never actually ran for a slot that measurement never told it was zero — it kept consuming a pairing partner as if it were still real content, landing in its own empty row.

Both root-caused and verified via live CDP inspection against a fresh dev build off the just-merged `main` (`Emulation.setDeviceMetricsOverride` to test multiple widths without fighting the app's own enforced window minimum), not just re-reading the diff — confirmed: wide tier back to a single correctly-populated row, and narrow tiers (300px, 150px) show only real, non-empty content on any filled side.

## Test plan
- [x] `npx tsc --noEmit`, `npx vitest run` (3154 passing, 1 pre-existing unrelated flaky timeout in `tool-renderers/registry.test.ts` confirmed present on `main` before this change too)
- [x] Live verification via CDP against a fresh `task dev` build off updated `main`: wide tier (630px window) → 1 row, full real content both sides; medium (300px viewport override) → 2 correctly-paired rows, no empty sides; narrow (150px viewport override) → 3 rows, all real visible content on filled sides, `auth` slot correctly folds into the `hostShell` row contributing 0 visible width instead of its own empty row


a5af:
@codex review

PR #2815, opened 2026-08-26
Manoz@Area54: feat(agent-pane): composer strip edge priority — interactive elements flush against row edges

GenericAgentX-asaf:
User-directed follow-up to the Rev 7 row model (#2812, #2813, #2814): on every rendered line, interactive elements (buttons/dropdowns) sit flush against the strip's outer edges, with passive informational content (auth status, ctx text) placed inward — e.g. when the ctx group is the left edge slot, its **Compact** button now sits flush-left, before the token text.

## Mechanisms (both ordering-only)

1. **`orderKeysForEdgePriority`** — a stable partition per row side: interactive slots first on the left, last on the right. Applied as the final step of `rows()` on every return path (fallback included), after all row-membership and shed decisions, so it **cannot affect the spec §1 invariant or any fit/pairing math**. Stability preserves \"Shell always outermost\" for free (hostShell is interactive AND last in pool order).
2. **Side-mirrored internal order** for the composite ctx/hostShell slots: `render()` now receives the row side it mounts into and places Compact/Shell on the outer end. Plain parameter, not a reactive read — a slot changing sides crosses `<For>` lists and re-renders with the new side.

Slots gain an `interactive` flag: runtime dropup / process badge / hostShell true, auth false, ctx true exactly when its Compact button actually renders.

## Tests & verification

- 8 new tests: 4 pure-function (partition + stability), 4 DOM-order (Compact-before-ctx on left, auth inward of badge on right, Shell outermost-right, and the degenerate hostShell-lands-left flip).
- All 36 `AgentComposerStrip.test.tsx` tests pass; `tsc --noEmit` clean.
- Verified live via CDP against the running dev build: single-row tier reads `[Compact | ctx | auth] — [runtime | HOST | Shell]`; narrow 2-row tier keeps every row two-sided with Compact flush-left; wide → narrow → wide round-trips return to the original layout (the hysteresis check #2814 made mandatory).
- Spec progress log updated (`SPEC_COMPOSER_STRIP_ROW_BASED_LAYOUT_2026_08_26.md` rows 17–18).


a5af:
@codex review

PR #2816, opened 2026-08-27
Manoz@Area54: feat(agent-pane): composer strip element-level edge priority + conservative fit rounding

GenericAgentX-asaf:
Two user-directed follow-ups to #2815 (which was itself a follow-up to #2812/#2813/#2814).

## 1. Element-level edge priority

#2815's slot-level ordering could not place a passive element from one slot inward past a *neighboring* slot's interactive element — the single-row right side read `ctx · Compact · HOST · Shell`, passive HOST closer to the edge than interactive Compact. Per the user's clarified rule (\"sometimes there will be multiple interactive elements mixed with non-interactive elements in a section — the interactive elements should be placed closest to the edges\"):

- CSS `order` scoped to each slot's **top-level** elements (`.agent-composer-strip-slot-measure > :is(...)`): interactive elements order toward the outer edge (`-1` left / `1` right) **across slot boundaries**. Visual-only — DOM order, slot measurement, pairing math, and focus order untouched; #2815's JS layers keep DOM aligned with visuals at slot granularity.
- The `.agent-composer-strip-host-shell` wrapper is dissolved (HOST + Shell now direct siblings, same shape as ctx's elements) so each is independently orderable. That wrapper implemented an older direct user request (\"HOST just left of Shell\") — **the user explicitly chose strict edge priority over it** when the two directives collided (recorded in the SCSS removal note). Row-level `flex-wrap` replaces the wrapper's internal-wrap fallback; the 260px shed query is rescoped.

## 2. Conservative fit rounding (user-reported \"2 breakpoints within a couple px around 292px\")

Slot widths fed `computeComposerRows` rounded to the **nearest** 8px — up to 4px/slot under-reported, making the fit check optimistic: the row's CSS `flex-wrap` safety net could overflow a few px *before* the JS split threshold, producing two layout breaks within a couple px on a resize. Not reproducible with today's content (rounding direction is content-dependent; 1px-granular CDP sweeps showed a single clean transition), but it is the only mechanism consistent with the report. Both measurement effects now round **up** (`Math.ceil`): anti-jitter bucketing preserved, and the JS split is guaranteed to fire at-or-before real-layout overflow — one breakpoint, any content.

## Verification

- Live CDP (visual x-position order, since JSDOM can't see stylesheets): single-row right reads `ctx · HOST · Compact · Shell`; left edge = dropup trigger; narrow 2-row keeps Compact flush-left / Shell edge-right; 1300 → 906 → 700 → 1300 round-trip restores identical layouts with a single split transition.
- All 36 unit tests pass (wrapper-selector test updated); `tsc --noEmit` clean; stylelint clean apart from 2 pre-existing grandfathered hex-fallback hits.
- Spec progress log updated (row 19).


a5af:
@codex review

PR #2817, opened 2026-08-27
Manoz@Area54: fix(agent-pane): composer strip stats eviction middle tier — no more 1-to-3-line jump

GenericAgentX-asaf:
Follow-up to #2816, user-reported live: with turn stats ticking, the strip jumped from 1 visual line **directly to 3**.

## Root cause

`statsWidth` fed `computeComposerRows`' single-row fit check as `reservedWidth` (Codex P1, #2812). The moment slots-plus-stats stopped fitting, the SLOTS split into 2 rows — and the stats simultaneously relocated to their own dedicated line below: 3 visual lines at once. The strictly-better middle tier — slots stay on one line, only the stats move below (2 lines) — was unreachable, because the fit check conflated \"do the slots fit\" with \"do the stats fit next to them.\"

## Fix

- `computeComposerRows` no longer takes `reservedWidth`: row membership is decided from slot widths alone.
- Whether the stats zone shares the single row is the component's separate `statsInline` decision (`slotsTotal + stats + gap ≤ width`), now controlling the single/multi mount position that used to be keyed on `rows().length`.
- Codex P1's original overflow concern cannot recur: when slots-plus-stats don't fit, the stats leave the row entirely instead of overflowing it.

Tier progression is now **1 line → 2 (slot row + stats row) → 3 (2 slot rows + stats row)** — each transition moves exactly one thing.

## Verification

- Contract test rewritten (`reservedWidth` no longer splits slot rows); all 36 unit tests pass; `tsc --noEmit` clean.
- **Live CDP tier verification pending**: the dev instance's agent pane was closed mid-session while this was implemented. Will verify the 1→2→3 progression live (with ticking stats) and confirm here before merging, per this file's own verification discipline.
- Spec progress log updated (row 20).


a5af:
@codex review

GenericAgentX-asaf:
Live CDP tier verification complete (dev build, ticking turn stats `↑244k ↓17 · 1s` in the zone):

| strip width (zoomed px) | slot rows | stats | visual lines |
|---|---|---|---|
| ≥ ~295 | 1 | inline on the row | **1** |
| ~243–295 | 1 | evicted to own line | **2** |
| < ~243 | 2 | own line | **3** |

The middle tier exists (~50px wide band) and each transition moves exactly one thing — the 1→3 jump is gone. Emulation override cleared after the sweep. ReAgent's 4 P2 stale-comment notes addressed in `1ca49113`.

GenericAgentX-asaf:

PR #2801, opened 2026-08-25
fix(agent): Activity Dock backfill suppression + background-task registry reconciliation

Agent1-asaf:
## Summary

Implements Tier 1 of `docs/reports/REPORT_AGENT_PANE_ACTIVITY_DOCK_ARCHITECTURE_ANALYSIS_2026_08_25.md` (Tier 0 — cutting a release — turned out to already be done by `gh-reporter`'s automated tagging before this PR started).

- **Backfill-aware refresh suppression** (`activity/backfill-tracker.ts`): PR #2773's debounce coalesces refresh *volume* during a subagent backfill burst, but each surviving call is still a genuinely different, still-converging snapshot — dock rows kept visibly appearing/vanishing (`docs/retro/retro-activity-dock-flicker-survives-debounce-fix-2026-08-24.md`). This reuses the already-shipped `subagent:backfill_status` event to suppress `dispatch-source.ts`/`subagent-source.ts` refreshes entirely while any block's backfill is in flight, firing exactly one refresh once it settles.
- **Background-task registry reconciliation** (`activity/background-adapter.ts`): a background task that survived a session restart has no transcript history, so it was invisible as a dock ROW even though `useBackgroundTaskRegistry.ts` already tracked it for the `attachedTask` boolean axis. The hook now also returns its raw task list; `ActivityDock.tsx` renders registry rows not already covered by `tool-adapter.ts`'s transcript-derived classification, deduped by `BackgroundTaskView.id` (which mirrors the originating `tool_use_id` — confirmed via `background_tasks.rs`'s own doc comment).

Both are additive/suppression-only changes — no backend wire protocol changes, no touch to Tier 2/3 (explicitly out of scope per this session's scoping decision).

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] `npx vitest run frontend/app/view/agent/` — 1434/1435 pass (1 pre-existing flaky timeout in `tool-renderers/registry.test.ts`, confirmed unrelated — passes in isolation)
- [x] New unit tests: `backfill-tracker.test.ts` (20 tests), `background-adapter.test.ts` (7 tests), plus new coverage in `dispatch-source.test.ts`/`subagent-source.test.ts` for backfill-aware suppression wiring

PR #2818, opened 2026-08-27
fix(tab): optimistic tab removal on close — flash structurally impossible

AgentY-asaf:
## Summary
- The close flash survived all four ordering fixes in #2811 (click containment, single-RPC close, batched response application, WS parent-first ordering). Root problem: every fix in that PR tries to **win an ordering race** between the HTTP response, individual WS push frames, and Solid flush boundaries — any new transport/event arm can reopen the class.
- This PR flips to the repo owner's directed UX: **the tab leaves the strip the moment the close-confirm modal opens** (or immediately on the skip-confirm path); **cancel or RPC failure puts it back**. A hidden tab is unmounted, so no backend update ordering can ever paint it — the flash becomes structurally impossible rather than a race we keep winning.
- While the real active tab is mid-close, `displayActiveTabId()` highlights the neighbor the backend is about to promote (mirrors `handle_delete_tab`'s promotion), so the strip shows the final post-close state from the very first frame and the backend's update changes nothing visibly.
- Guards (`length <= 1`) deliberately stay on the raw list — a filtered guard would make a 2-tab workspace refuse every confirmed close.
- Full analysis in `docs/specs/SPEC_TAB_CLOSE_BUTTON_SELECT_FLASH_2026_08_25.md` §8, including why §§3-7 remain load-bearing for non-tab update paths.

## Test plan
- [x] `npx vitest run` — 3164 passed / 1 failed: `tool-renderers/registry.test.ts` timeout, pre-existing flake (passes in isolation, also flagged in #2801, untouched by this diff)
- [ ] Manual (needs a dev build of THIS branch — note v0.55.25 does not contain #2811 either): ✕ on active tab → tab gone from strip instantly, neighbor highlighted, modal open → Cancel → tab returns in place. Confirm → strip doesn't change (already showed final state); content switches atomically under the reveal gate.

### DIFF
--- a/.changesets/1787814801-fix-tab-optimistic-tab-removal-on-close-close-flash-structur-ocnk.md
+++ b/.changesets/1787814801-fix-tab-optimistic-tab-removal-on-close-close-flash-structur-ocnk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tab): optimistic tab removal on close — close flash structurally impossible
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #2819, opened 2026-08-27
Manoz@Area54: fix(agent-pane): evicted composer strip stats float above the rows

GenericAgentX-asaf:
User-directed follow-up to #2817: \"the center token count should float up (not down) as the width is made narrow.\"

Same `statsInline` tier model (1 line → 2 → 3), placement flipped: the evicted stats zone renders **above** `.agent-composer-strip-rows` — a DOM-order move within the column-flex strip — so narrowing floats the centered token stats up while the interactive slot rows stay anchored at the bottom next to the composer.

CDP-verified live with ticking stats (`↑248k ↓29 · 1s`): inline at wide; above the single slot row at the 2-line tier (`statsTop 849 < rowsTop 857`); above both rows at the 3-line tier; back inline on widening. 41 unit tests pass; `tsc --noEmit` clean. Spec progress log row 21.


a5af:
@codex review

PR #2822, opened 2026-08-27
feat(layout): flip splitter-drag defaults — plain drag group-resizes, Shift resizes a single border

AgentY-asaf:
## Summary

Implements **Change 1** of `docs/specs/SPEC_RESIZE_DEFAULT_FLIP_AND_WINDOW_EDGE_SHIFT_2026_08_26.md` (spec included in this PR — it was not yet on main):

| Gesture | Old | New |
|---|---|---|
| Plain drag | 2-node direct transfer | **group resize (two-block)** |
| Shift + drag | group resize (two-block) | **2-node direct transfer** |

- `TileLayout.{win32,darwin,linux}.tsx` now pass `!shiftKey` as `onResizeMove`'s `groupResize` argument (the only place mode selection happens).
- `layoutResize.ts`: comment on the `if (groupResize)` branch updated to describe the flipped mapping — **no math changes**.
- `quicktips.tsx`: entry flipped to "Resize Single Border — Shift + Drag".
- Changeset added (`minor` — behavior change).

## Why

Owner-directed: the group (two-block relative) behavior is the better everyday default — one drag adjusts the whole row/column coherently; the surgical "move only this border" case is the rarer, deliberate action a modifier is for. Per spec §4, after this flip **Shift uniformly means "surgical/direct"** across both splitter drags and the planned Shift+window-edge resize (Change 2, separate task) — before it, Shift would have meant "bigger effect" on splitters but "smaller effect" on window edges, an incoherent mental model.

Mid-drag Shift toggling keeps working unchanged (both formulas share the same snapshot context), just flipped in direction.

## Test plan

- [x] `npx vitest run frontend/layout/` — 9 files, 148 tests passed, untouched (pure-function tests do not encode the modifier mapping)
- [ ] Manual: plain drag moves the whole row/column; Shift+drag moves one border; toggling Shift mid-drag flips live in both directions

### DIFF
--- a/.changesets/1787814920-feat-layout-plain-drag-now-group-resizes-the-row-column-shif-pl9l.md
+++ b/.changesets/1787814920-feat-layout-plain-drag-now-group-resizes-the-row-column-shif-pl9l.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(layout): plain drag now group-resizes the row/column; Shift+drag moves a single border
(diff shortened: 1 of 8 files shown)


a5af:
@codex review

PR #2823, opened 2026-08-27
feat(layout): Shift+window-edge resize — delta feeds only the edge panes (Windows-first)

AgentY-asaf:
## Summary

Implements **Change 2** of `docs/specs/SPEC_RESIZE_DEFAULT_FLIP_AND_WINDOW_EDGE_SHIFT_2026_08_26.md` (§3 — the spec doc itself lands with the flip agent's PR and is not duplicated here):

- **Plain window resize: unchanged.** Proportional scaling, zero weight writes, layout file untouched — a session where Shift is never held stages nothing and commits nothing.
- **Shift held while dragging an OS window edge:** the entire pixel delta goes to the pane(s) abutting the dragged edge; every other pane keeps its exact **pixel** size (flex weights rewritten as `w_i' = W·p_i'/P'` under the new container size). Shift can be pressed/released mid-resize and the mode follows live; releasing Shift restages the session-start weights (pure proportional), so the gesture is reversible within the session.
- **Windows-only for now** — spec §3.4 phase 2 covers mac/linux (no `WM_SIZING` equivalent); the renderer listener is simply inert where the host never emits `windowresize:*`.

## Plumbing

**Host (`agentmux-cef`, Windows):**
- New pure-observer wndproc subclass `install_window_edge_resize_hook` in `client/wndproc.rs`, pattern-copied from the existing close-routing subclass (per-HWND original-proc map, `WM_NCDESTROY` self-prune, AppState `OnceLock`). Installed from `on_after_created` (`client/lifecycle.rs`) on every top-level non-popup window, `main` included — it never short-circuits a message.
- `WM_SIZING` → first tick of a loop emits `windowresize:begin {}` (begin is **lazy** here, not on `WM_ENTERSIZEMOVE`, which also fires for plain window *moves*); every tick samples `GetAsyncKeyState(VK_SHIFT)` (precedent: `commands/drag.rs`) and emits `windowresize:tick { edge, shiftHeld }` with the edge verbatim from `wParam` (`WMSZ_*` → `"left" | ... | "bottomright"`). `WM_EXITSIZEMOVE` → `windowresize:end {}` iff a begin was sent.
- Events route via the existing `emit_event_to_window` channel; the label is resolved fresh per emit via `AppState::label_for_hwnd` (install-time label as fallback), so promoted pool windows route correctly.

**Frontend (`frontend/layout/lib/windowEdgeResize.ts`, installed from `app-init.ts` alongside the other host-event listeners):**
- On `begin`: snapshot the active tab's tree (weights + `additionalProps` rects + container size; minimize-locked nodes excluded — they're fixed chips outside the flex pool and `resizeNode` rejects actions touching them).
- Per `tick` with Shift: cumulative axis delta since begin → recursive edge-chain rule (spec §3.2): in a Row container only the edge-most child's width changes (recursing into it if it's a container); siblings keep pixels with weights recomputed (parent weight **sum** preserved); in a Column container every child spans the axis, so recurse into each. Corners decompose per-axis — a node's weight lives on exactly one parent axis, so the two passes can't conflict.
- Floors at `MinNodeSizePx = 128` via a **new pure primitive `spillInwardBy`** (edge-first inward spill, then proportional fallback when the whole container is at the floor — converging with plain-resize behavior instead of wedging against an OS resize it can't refuse; spec §3.3). `shrinkBlockBy` untouched (its distribution is proportional-within-pool, not edge-first).
- Stages via `SetPendingAction`/`ResizeNode` + `updateTree(false)` exactly like a splitter-drag tick (so the WxH badge engages via `isSplitterDragging`); commits once on `end` (single history/persist entry). An end with Shift released clears the pending action instead of committing a no-op.
- **Units:** all deltas are computed renderer-side from the layout container's own `getBoundingClientRect` — the same source the `additionalProps` rects come from — so everything stays in one CSS-px space and no physical-px/DPR/`--zoomfactor` conversion exists in this path at all. The host forwards no coordinates, only `edge` + `shiftHeld`.

## Tested

- `npx vitest run frontend/layout/` — **164/164 pass** (10 files), including the new `frontend/layout/tests/windowEdgeResize.test.ts` (16 tests): right-edge grow routes all delta to the last child while siblings keep pixels; left-edge (first-child) variant; nested column-in-row recursion (width delta never touches column-child height weights); floor + inward spill; partial-headroom + proportional fallback; all-at-floor fallback producing **zero** weight ops (proportional = weight-neutral by construction); corner two-axis decomposition; below-floor members skipped, never enlarged; sub-pixel deltas are no-ops.
- `npx tsc --noEmit -p tsconfig.json` — clean.
- `cargo check -p agentmux-cef` — clean (no new warnings).

## Needs live manual verification (Windows)

The host→renderer event flow has **not** been exercised in a live build (cargo *check* only, per task scope):

1. 3-pane row, Shift-drag right window edge wider → only the rightmost pane grows; others' pixel widths hold (WxH badge should engage).
2. Shrink until the edge pane floors → confirm inward spill, then proportional convergence.
3. Release/press Shift mid-gesture → live flip between edge-fed and proportional staging, reversible.
4. Nested column inside the row → recursion touches the correct leaves; corner drags affect both axes independently.
5. Persistence: Shift-resize weights survive reload; plain resize leaves the layout file byte-unchanged.
6. WM_SIZING timing: renderer rect reads lag the in-flight tick by one message; the `end` pass restages at the final size before committing — verify no visible snap at release.

## Spec assumptions that didn't survive contact with the code

- **§3.3 suggested reusing `shrinkBlockBy` "as the right primitive"** then corrected itself — confirmed: its proportional-within-pool distribution is wrong for accordion order; `spillInwardBy` is new and standalone as the spec's own follow-up sentence suggests.
- **§3.5's `--zoomfactor` caution turned out to be moot**: chrome zoom never mixes units here because both the session-start rects and the per-tick container size come from the same `getBoundingClientRect` space, and the host forwards no pixel quantities.
- **`WM_ENTERSIZEMOVE` cannot open the session** (spec §3.4 sketch): it fires for window *moves* too; begin had to be lazy on the first `WM_SIZING`.
- **`resizeNode` rejects any op set touching a minimize-locked node and any size outside 0..100** (`layoutTree.ts:431`) — snapshot excludes locked nodes entirely, and the weight rewrite preserves each parent's weight *sum*, keeping values in the same scale the tree already uses.

### DIFF
--- a/.changesets/1787815686-feat-layout-shift-window-edge-resize-feeds-only-the-edge-pan-p6co.md
+++ b/.changesets/1787815686-feat-layout-shift-window-edge-resize-feeds-only-the-edge-pan-p6co.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(layout): Shift+window-edge resize feeds only the edge panes (Windows)
--- a/agentmux-cef/src/client/lifecycle.rs
+++ b/agentmux-cef/src/client/lifecycle.rs
@@ -294,6 +294,19 @@ impl AgentMuxHandler {
                 // Install on every top-level — both `main` and Subwindow.
                 if is_top_level_window && !is_popup {
                     unsafe { install_top_level_focus_restore_hook(hwnd); }
+
+                    // Shift+window-edge resize (spec SPEC_RESIZE_DEFAULT_FLIP_
+                    // AND_WINDOW_EDGE_SHIFT_2026_08_26.md §3.4): observe the
+                    // native size loop (WM_SIZING/WM_EXITSIZEMOVE) and forward
+                    // windowresize:* events to this window's renderer. Pure
+                    // observer-passthrough — safe on main and Subwindows alike.
+                    unsafe {
+                        super::wndproc::install_window_edge_resize_hook(
+                            &self.state,
+                            hwnd,
+                            &label,
+                        );
+                    }
                 }
 
                 // OS-close routing (task #30): Alt+F4 / taskbar-close
--- a/agentmux-cef/src/client/wndproc.rs
+++ b/agentmux-cef/src/client/wndproc.rs
@@ -226,6 +226,230 @@ pub(crate) unsafe fn install_main_window_floater_cascade_hook(
     }
 }
 
+/// Per-HWND entry for the window-edge-resize hook: original WndProc,
+/// install-time window label (fallback for event routing), and whether a
+/// `windowresize:begin` has been emitted for the in-progress size loop.
+#[cfg(target_os = "windows")]
+struct WindowEdgeResizeHookEntry {
+    original: isize,
+    install_label: String,
+    session_began: bool,
+}
+
+/// Map of top-level HWND -> hook entry for the Shift+window-edge-resize
+/// subclass (`install_window_edge_resize_hook`). Self-prunes on
+/// WM_NCDESTROY (HWND-reuse safety, same discipline as
+/// `CLOSE_ROUTING_WNDPROCS`).
+#[cfg(target_os = "windows")]
+static WINDOW_EDGE_RESIZE_WNDPROCS: std::sync::LazyLock<
+    std::sync::Mutex<std::collections::HashMap<usize, WindowEdgeResizeHookEntry>>,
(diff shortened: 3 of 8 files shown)


a5af:
@codex review

PR #2824, opened 2026-08-27
feat(browser-pane): credential-isolated HTTP auth auto-fill

a5af:
## Summary

Adds credential-isolated auto-fill for browser-pane HTTP Basic/Digest auth — inspired by Codex's credential-isolated browsing feature (see `docs/research/RESEARCH_CODEX_CREDENTIAL_ISOLATED_BROWSING_2026_08_26.md`). On a repeat auth challenge with a saved credential, a human approves via a genuinely separate top-level window before the password ever reaches CEF — the agent controlling the pane never sees it.

- **srv**: OS-keychain-backed credential store (identity-scoped, `agentmux-srv/src/identity/browser_credential_store.rs`) + a new `credential` RPC service (`ResolveIdentity`/`Lookup`/`Save`/`Delete`/`Fill` — `Fill` is the only method that ever returns a plaintext password, called exactly once per approved use).
- **agentmux-cef**: `credential_broker` orchestrates the `on_auth_credentials` flow (identity resolve → keychain lookup → approval window or fall-through) and owns a parked-approval registry that coalesces simultaneous challenges for the same protection space into one approval window. A 15s "recently auto-filled" marker auto-invalidates a wrong stored credential instead of looping it.
- **frontend**: `CredentialApprovalWindow` renders as the *entire content* of a dedicated subwindow (`open_subwindow`, never a `<Modal>`) — this is the key security property: AgentMux's `UIQuery`/`UIClick` treat anything outside a pane's own `[data-blockid]` subtree as reachable shared chrome, so a `<Modal>`-based approval UI would let *any* agent click Approve, not just the one whose pane raised the challenge. `BrowserAuthModal` gains an opt-in "Save this credential" checkbox.
- Every failure mode (no bound identity, no stored credential, srv unreachable, keychain error, window-creation failure) falls through to the existing, unmodified manual-entry prompt — this can only make auth *more* automatic, never break the baseline.

## Test plan

- [x] `cargo test -p agentmux-srv` — 2823 passed (incl. 7 new `browser_credential_store` tests + 3 new `credential::tests`)
- [x] `cargo test -p agentmux-cef` — 278 passed (incl. 11 new `credential_broker`/`approval` tests: coalescing, TTL, window/block cleanup)
- [x] `npx stylelint` on touched SCSS — clean
- [x] Live `task dev` verification: manual save → close/reopen pane → approval subwindow appears (not the old modal) → Approve loads the page; Deny shows the normal 401 body
- [x] Adversarial check: an agent's own `UIQuery`/`UIClick` cannot reach the approval window's Approve button
- [x] Coalescing: two simultaneous challenges to the same origin produce exactly one approval window

### DIFF
--- a/.changesets/1788031768-feat-browser-pane-credential-isolated-http-auth-auto-fill-hato.md
+++ b/.changesets/1788031768-feat-browser-pane-credential-isolated-http-auth-auto-fill-hato.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(browser-pane): credential-isolated HTTP auth auto-fill
--- a/agentmux-cef/src/client/crash_recovery.rs
+++ b/agentmux-cef/src/client/crash_recovery.rs
@@ -515,28 +515,23 @@ impl AgentMuxHandler {
             cb.clone(),
         );
 
-        // Broadcast to every TOP-LEVEL window — `emit_event_from_state`
-        // only dispatches to the "main" browser, so panes hosted in a
-        // tear-off / secondary window would never see the event and
-        // the CEF callback would wait until TTL/cancel. Filtering to
-        // top-level is critical: the payload carries origin/host/realm
-        // plus block_id correlation, which must not be visible to
-        // remote content loaded inside a sibling pane (whose main
-        // frame is an arbitrary URL). The host renderer filters on
-        // `payload.block_id` so only the window owning this pane
-        // surfaces the prompt.
-        crate::events::emit_event_to_top_level_windows(
-            &self.state,
-            "browser-pane-auth-required",
-            &serde_json::json!({
-                "block_id": block_id,
-                "request_id": request_id,
-                "origin": origin,
-                "host": host_str,
-                "port": port,
-                "realm": realm_str,
-                "is_proxy": is_proxy_bool,
-            }),
+        // Hand off to the credential broker: identity-resolve → stored-
+        // credential lookup → either a human-approval subwindow (never a
+        // `<Modal>` — see `credential_broker`'s own doc comment for why)
+        // or, on any friction at all (no bound identity, no stored
+        // credential, srv unreachable, etc.), the exact same
+        // `browser-pane-auth-required` broadcast this function emitted
+        // unconditionally before this feature existed. See
+        // `docs/status/majestic-painting-minsky` plan.
+        crate::credential_broker::on_auth_challenge(
+            self.state.clone(),
+            request_id,
+            block_id,
+            origin,
+            host_str,
+            port,
+            realm_str,
+            is_proxy_bool,
         );
 
         1
--- a/agentmux-cef/src/client/lifecycle.rs
+++ b/agentmux-cef/src/client/lifecycle.rs
@@ -1020,6 +1020,33 @@ impl AgentMuxHandler {
             }
         }
 
+        // A credential-approval subwindow (opened via `open_subwindow`,
+        // `initial_view=credential-approval`) closing before the human
+        // decided — parent window closing cascades down to it same as any
+        // other subwindow. No-op (empty Vec) for every ordinary window
+        // close; the registry only ever has entries keyed by an actual
+        // approval subwindow's label. See `credential_broker::approval`'s
+        // own doc comment — this is the mirror of `browser_pane::auth`'s
+        // pane-close cleanup below, just window-close instead of
+        // pane-close.
+        if let Some(ref lbl) = label {
+            let cancelled = crate::credential_broker::approval::cancel_for_window(lbl);
+            if !cancelled.is_empty() {
+                use cef::ImplAuthCallback;
+                tracing::info!(
+                    "[credential-broker] approval window {} closed before a decision — \
+                     cancelling {} parked auth request(s)",
+                    lbl,
+                    cancelled.len(),
+                );
+                for request_id in &cancelled {
+                    if let Some(cb) = crate::browser_pane::auth::take(request_id) {
+                        cb.cancel();
+                    }
+                }
+            }
+        }
+
         // Pane-specific on_before_close work (drain lifecycle entry) lives
         // in `crate::browser_pane::callbacks` after Phase 4.
         if let Some(ref lbl) = label {
--- a/agentmux-cef/src/commands/window/creation.rs
+++ b/agentmux-cef/src/commands/window/creation.rs
@@ -294,6 +294,8 @@ pub fn open_new_window(state: &Arc<AppState>, args: &serde_json::Value) -> Resul
 pub fn open_subwindow(
     state: &Arc<AppState>,
     parent_instance_id: String,
+    initial_view: Option<&str>,
+    initial_meta: Option<&str>,
 ) -> Result<serde_json::Value, String> {
     // Reject if the parent isn't a known live FullInstance — prevents
     // orphan sub-windows and enforces the lifecycle rule in the spec.
@@ -328,8 +330,8 @@ pub fn open_subwindow(
         state,
         crate::state::WindowKind::Subwindow,
         Some(parent_instance_id),
-        None,
-        None,
+        initial_view,
+        initial_meta,
         None,
         false,
     )
--- a/agentmux-cef/src/credential_broker/approval.rs
+++ b/agentmux-cef/src/credential_broker/approval.rs
@@ -121,17 +121,29 @@ pub fn reserve_or_join(
 /// Record the approval subwindow's `window_id` once it has actually been
 /// created, so `cancel_for_window` can later find this entry when the
 /// window (or its parent, cascading down) closes.
-pub fn finalize_window(approval_id: &str, window_id: String) {
+/// Returns `false` when the approval was already gone — the reservation was
+/// cancelled (pane closed) or timed out during the window-creation gap, so
+/// the window the caller just opened belongs to nothing and **the caller
+/// must close it**. `#[must_use]` because ignoring that answer is exactly
+/// the bug this return value exists to prevent (reagent P1 on PR #2824):
+/// the entry is gone, so `cancel_for_window` will never match it, the IPC
+/// decide-handler's `take` returns `None` and closes nothing, and the TTL
+/// timer has no entry left to fire on — leaving a dead, unresponsive
+/// subwindow open until the human closes it by hand.
+#[must_use]
+pub fn finalize_window(approval_id: &str, window_id: String) -> bool {
     let mut g = pending().lock();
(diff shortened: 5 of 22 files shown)


a5af:
@codex review

a5af:
## Review findings summary + handoff notes

Consolidating feedback from both automated reviewers (`reagentx-workflow` + `chatgpt-codex-connector`) before handing this off. Fix order below is priority order — the P0 must land before anything else matters, since it defeats the PR's entire stated goal.

### P0 — BLOCKING: agent can read the plaintext password directly, no approval involved
`agentmux-srv/src/server/service/credential.rs` — `Fill` (and `Lookup`/`ResolveIdentity`/`Save`/`Delete`) is gated only by the instance-wide `X-AuthKey` on `POST /agentmux/service`. That same key is injected into every spawned agent's environment as `AGENTMUX_AUTH_KEY` (see `server/agent_handlers/input.rs:306`), and `AGENTMUX_BLOCKID` is injected too. So the agent controlling the authenticated pane can:
1. `curl` `credential.ResolveIdentity` with its own known `block_id` → gets `identity_id`
2. `curl` `credential.Fill` with that `identity_id` + the origin/realm it already knows (it's the site it's browsing) → gets the raw plaintext password back

No human approval subwindow is ever involved in this path. This is a complete bypass of the security property the PR claims to deliver ("the agent controlling the pane never sees it").

**Fix direction**: `Fill` must require something an agent cannot forge from ambient env vars — e.g. a single-use, host-minted approval capability/token returned only after `credential_approval_decide` records a human Approve, passed to `Fill` instead of (or in addition to) the ambient `X-AuthKey`. `ResolveIdentity`/`Lookup` (masked-only) are lower risk but worth re-examining once the capability design is settled — an agent using `ResolveIdentity` to enumerate identity_ids is itself a smaller information leak.

### P1 — orphaned approval subwindow (race)
`agentmux-cef/src/credential_broker/mod.rs` (`run_challenge`) — `approval::reserve_or_join()` and `approval::finalize_window()` are two separate lock-guarded ops with no lock held across the synchronous `open_approval_window()` call in between. If `browser_pane_close` fires concurrently (another tokio worker) and calls `cancel_for_block` for the same block_id (its only pending request), the entry is removed before `finalize_window` runs. `finalize_window` then no-ops with a warning ("window is now orphaned and should be closed by the caller") — but nothing does. The window is never registered in `cancel_for_window`'s map, `credential_approval_decide` returns `Ok(false)` for the now-missing approval_id without closing anything, and the window sits open indefinitely.

### P2 — `cancel_for_block`'s early `break` drops cleanup (three related bugs)
`agentmux-cef/src/credential_broker/approval.rs:206` — loop breaks as soon as it finds the first entry that becomes empty after removing this block_id's request, skipping any other entries for the same block_id:
- A pane with two simultaneous challenges for different protection spaces (e.g. proxy-auth + origin-auth) → the second approval's stale request lingers until the 60s TTL instead of being cleaned up immediately.
- Its window also never gets closed early — same TTL-until-cleanup gap.
- Related, same root cause class: in `mod.rs`, when a *joined* (coalesced) request's window creation fails, `abandon()` correctly returns every stranded request id, but the caller only falls through the *original* request — the joined one hangs parked until TTL instead of getting the manual prompt.

**Fix direction**: collect *all* matching empty entries (don't break on first), return every associated window_id for closure, and iterate `abandon()`'s full return list when falling through.

### P2 — Fill failure post-approval violates the "always fall through" design
`agentmux-cef/src/ipc.rs` (`credential_approval_decide` handler) — every other failure path in this feature falls through to the existing manual-entry prompt (`browser-pane-auth-required`) by design ("this feature can only ever make auth more automatic, never break the baseline"). But if `Fill` fails *after* the human already approved (transient keychain/backend hiccup), the current code cancels every parked callback instead — the page renders its 401 body and the user has no way to type credentials manually for that challenge. Need to retain enough per-request challenge metadata (origin/host/port/realm/is_proxy) through to this point so it can emit `browser-pane-auth-required` per request instead of cancelling.

### P2 — TTL-expired approvals never close their window
`agentmux-cef/src/credential_broker/approval.rs` TTL timer removes the registry entry but doesn't close the associated subwindow. A late Approve/Deign click after TTL then hits `credential_approval_decide`'s "not found" branch and returns without closing anything — the window is permanently inert until the user manually closes it. Either close the window from the TTL callback, or preserve enough info for the expired-decision IPC path to close it.

### P2 — missing changeset
Per this repo's CLAUDE.md, feature PRs need a `.changesets/<ts>-<slug>.md` entry; none was added for this PR's 20 changed files.

---

**Status**: implementation (all 11 build/test tasks) and live manual verification of the happy path + adversarial UIQuery/UIClick unreachability were done and passed *before* these reviews landed — but the P0 above means the live adversarial test didn't cover the actual bypass vector (a direct RPC call from the agent's own shell, not through UIQuery/UIClick). That needs to be added to the verification checklist once P0 is fixed: confirm an agent's `Shell` tool cannot obtain the plaintext password via a direct `curl` to `/agentmux/service`, not just that it can't click the Approve button.

Agent2-asaf:
Picked up the handoff. **P0, P1 and three of the P2s are fixed** in `5111597dc` + `bd88b4fde`. Two items remain — detail at the bottom.

## P0 — agent can read the plaintext password. Fixed (`5111597dc`)

Every method on the `credential` service now requires `args[1]` to be `AGENTMUX_HOST_REG_SECRET`, reusing `host_ipc`'s existing constant-time `secret_matches` rather than growing a second comparison. An agent has `X-AuthKey` from its own env but can never have this, so the `curl` path is closed.

**I went with the host-secret rather than the single-use approval capability you sketched**, and want to be explicit about why so you can overrule it. Your P0 said *"something an agent cannot forge from ambient env vars — e.g. a single-use, host-minted approval token"*. The host secret satisfies that requirement, already exists, and is already reviewed under an identical threat model (#2662). The capability token would be **minted by the host and consumed by the host** — so against the attacker in question (an agent) it adds nothing the host-secret doesn't already give; it's an internal consistency check within the host, not a second independent gate. If you want that check too it's worth doing, but as defence against a broker bug, not against an agent — different goal, and I didn't want to smuggle it in under the P0 label. Recorded in the module doc so the reasoning isn't re-derived later.

Gated at the dispatcher rather than on `Fill` alone: a uniform gate can't be defeated by finding the one method someone forgot, and the others aren't harmless either — `ResolveIdentity` enumerates identity_ids, `Lookup` confirms which sites have saved credentials, `Save`/`Delete` let an agent plant or destroy them. You'd flagged `ResolveIdentity`/`Lookup` as "lower risk but worth re-examining"; covering them was free once the gate moved up.

Fails closed when srv has no secret. Writing the tests turned up one extra hole: `secret_matches("", "")` compares **equal**, so an empty configured secret would have admitted exactly the empty `args[1]` an agent sends. `Config` already filters empty (`config.rs:130`) so it's unreachable today — but the gate no longer depends on a filter two modules away. 8 tests.

## P1 + P2s — window lifetime. Fixed (`bd88b4fde`)

- **P1 orphaned window**: `finalize_window` now returns whether the approval survived the lock gap, and is `#[must_use]`; `run_challenge` closes the orphan and falls through to the manual prompt.
- **P2 `cancel_for_block` early break**: returns *every* emptied approval's window now. Worth noting this one was intermittent, not absent — which entry the `break` landed on depended on HashMap iteration order.
- **P2 TTL never closes its window**: closed from the TTL timer via a `set_window_closer` hook installed in `lib.rs`, so `approval` keeps its design property of holding `window_id` strings with no window-management dependency.
- **P2 missing changeset**: added, typed `minor`.

3 new approval tests. **281 cef / 2828 srv, all passing.**

## Still open — two items I did NOT do

1. **P2: `Fill` failure post-approval should fall through, not cancel.** and
2. **P2: `abandon()`'s joined requests only fall through the original.**

Both need the same thing: `AuthRequest` currently stores `request_id` + `block_id`, but falling a request through needs its `host`/`port` too. That's a real refactor of the registry's stored shape (`ResolvedApproval` and `abandon`'s return type both change), and I'd rather it be its own commit with your eyes on it than tacked onto a security fix. Happy to do it next if you want it in this PR.

## Verification gap — please read

Everything above is code-verified and unit-tested. **I have not run this live.** In particular your own note still stands and is now the important one:

> confirm an agent's `Shell` tool cannot obtain the plaintext password via a direct `curl` to `/agentmux/service`, not just that it can't click the Approve button

That adversarial check is exactly what the P0 fix claims to close, and it is precisely the claim a unit test can't make. It needs a `task dev` run and a real `curl` from an agent pane before this merges.

AgentX-asaf:
Both outstanding findings fixed in `7f2b1a463`.

**reagent P1 — `ipc.rs`, cancel-on-Fill-failure.** `credential_approval_decide` was cancelling every parked `AuthCallback` when `Fill` failed *after* the human had already approved. A credential deleted between approval and Fill, a transient keychain error, or an srv restart therefore hard-failed the page load to a bare 401 with no way to type credentials — strictly worse than pre-feature behaviour, and a direct violation of this feature's own "can only make auth more automatic, never break the baseline" guarantee.

The callbacks are now left **parked** in `browser_pane::auth` and the ordinary `browser-pane-auth-required` prompt is emitted for each instead; the manual prompt resolves those same parked callbacks, exactly as when no stored credential exists at all. Cancelling is still correct on **Deny** — that's the human declining, not a failure.

**reagent P2 — `credential_broker/mod.rs`, discarded `abandon` result.** Fixed; every returned request now falls through, not just the caller's own.

**Both needed something the registry was discarding**, so the shape changed rather than being worked around:

- `abandon` and `ResolvedApproval` now return `ParkedRequest { request_id, block_id }` instead of bare ids. This matters more than it looks: coalesced challenges can come from **different panes**, so each fall-through prompt has to be addressed to its own `block_id` — emitting them all against the original caller's would render prompts in a pane that never raised the challenge.
- `Entry`/`ResolvedApproval` now carry `host`/`port`, since the post-approval Fill-failure path reproduces the full `browser-pane-auth-required` payload long after the original challenge's locals are gone.
- The free `fall_through` fn is renamed `fall_through_to` and made `pub(crate)` so `ipc.rs` can reach it.

**Tests:** `cargo test -p agentmux-cef` — **283 passed, 0 failed**. Two added:
- `abandon_returns_every_coalesced_request_with_its_own_block_id` — pins the P2 fix *including* per-pane `block_id` routing
- `take_carries_host_port_and_per_request_block_ids` — pins the metadata the P1 fall-through depends on

Not done here: live `task dev` re-verification of the new fall-through path. The existing test plan's live checks were run against the original commit, and I can't re-run them from this session — worth a manual pass on the Fill-failure path specifically (approve, then make the keychain read fail) before merge.

PR #2825, opened 2026-08-27
fix(agent): remove false-positive-prone "agent unresponsive" detection

Agent2-asaf:
## Summary

`HealthMonitor`'s silence watchdog, error-rate tracker, and compaction-awareness
override were a persistent source of false positives — three separate rounds of
reagent-flagged bugs on the original implementation, a documented 231.6s
compaction false-positive (patched in #2754), an admitted no-hysteresis flicker
on recovery, and a confirmed structural exposure: any long-running tool call
(a multi-minute `Bash`, a `Task`/subagent dispatch) is judged purely by
wall-clock output silence, identical to the compaction case before its
dedicated patch.

This removes the detector, the "Agent unresponsive" banner + "Restart" action,
and the `FailureClass::Unresponsive` taxonomy entry entirely — resolving this
whole class of false positive categorically rather than patching each instance
as it's discovered.

**Kept, unchanged:** turn-active tracking (`is_active_turn`/`set_active_turn`/
`mark_turn_active_returning_was_active`), renamed `TurnActivityTracker` since
it's independently load-bearing for the Swarm pane's running/idle badge,
`process.rs::lifecycle_from()`, subagent-watcher reconciliation, and
`muxspect describe` — none of that is about detecting unresponsiveness, it's a
general "is this controller mid-turn" signal that happened to live on the same
struct. Also kept: process-exit classification and every other `FailureClass`
variant (auth, rate-limit, network, etc.) — entirely unaffected.

Full design + addenda: `docs/specs/SPEC_REMOVE_AGENT_UNRESPONSIVE_DETECTION_2026_08_25.md`
(Part 1). Part 2 of that spec (consolidating the separately-triggered
"Reconnecting…"/"Compacting…" status displays into one location) is scoped but
intentionally out of scope for this PR — tracked in the same doc as its own
follow-up.

Superseded docs archived: `agent-health-design.md`,
`SPEC_UNRESPONSIVE_FALSE_POSITIVE_DURING_COMPACTION_2026_08_22.md` → both moved
to `docs/specs/archive/` with a note pointing at this spec.

## Test plan

- [x] `cargo test` (isolated `AGENTMUX_DATA_DIR`/`CONFIG_DIR`/`SHARED_DIR`) —
      2786 passed, 0 failed, 6 ignored, plus `integration_test` (5/5) and
      `subprocess_io` (7/7) all green.
- [x] `tsc --noEmit` — clean.
- [x] `vitest run` — 218 test files, 3166 tests, 0 failed.
[...]

PR #2826, opened 2026-08-27
fix(agent-pane): consolidate Reconnecting…/Compacting… status into the Working row

Agent2-asaf:
## Summary

Follow-up to #2825 (Part 1 of the same spec, merged). "Reconnecting…" and
"Compacting…" used to render inside `AgentComposerStrip`'s stats zone — a
different DOM region than `AgentWorkingRow`'s "Working…", directly above the
textarea rather than in the conversation area above it. There was no visual
relationship between the two, so a user had no way to know there were two
separate status slots, or why "Reconnecting…" could appear while "Working…"
was simultaneously absent (correctly so — the underlying process had already
crashed and AgentMux was transparently retrying with a fresh `--resume`
session id; nothing was "working" in that moment, but nothing communicated
that distinction).

This moves both states into `AgentWorkingRow` as two additional
mutually-exclusive sub-states alongside `Working`/idle, so there's exactly
**one** place to check "is something happening to this agent right now" —
the same shimmer/type-out treatment `Working` already has applies
automatically. Wording stays distinct per state (`Reconnecting…` vs.
`Compacting…` vs. `Working…`) — this is a **placement** fix, not a wording
simplification. Collapsing to one generic label was considered and rejected
(spec §10, "Alternative considered, not recommended") — these states exist
specifically because a prior generic "is it hanging?" ambiguity was a real,
reported problem; collapsing them back would reintroduce it.

`AgentWorkingRow` now renders whenever `loading`, `compacting`, or
`reconnecting` is set (previously gated on `loading` alone) — `reconnecting`
in particular fires only *after* the underlying process has already exited,
so it's orthogonal to `TurnPhase`/`loading` by design. `AgentComposerStrip`
keeps the `compacting` prop (still gates the manual Compact button + its
tooltip) but loses `reconnecting` entirely and the two rightText() display
branches — it falls back to its plain token/elapsed readout, unchanged
otherwise. `AgentComposerStrip.tsx`'s layout/row-balancing algorithm
(computeBalancedLeftKeys/computeComposerRows/measurement effects) — recently
and heavily revised across several other PRs — was deliberately NOT touched
beyond these two prop/branch removals.

Full design + the rejected alternative: `docs/specs/SPEC_REMOVE_AGENT_UNRESPONSIVE_DETECTION_2026_08_25.md`
§9-§12 (Part 2).

## Test plan

- [x] `tsc --noEmit` — clean.
- [x] `vitest run` (full suite) — 218 test files, 3170 tests, 0 failed.
- [x] New `AgentFooter.test.tsx` cases: Reconnecting renders with
      `loading=false`/no `sessionStats`; Compacting renders similarly;
      Reconnecting takes priority over Compacting when both are set; the row
      stays absent when none of loading/compacting/reconnecting/sessionStats
      are set.
- [ ] Manual: trigger a real compaction and a real stale-resume retry and
      confirm the relocated status reads correctly in the single new
      location, with no flash of the old composer-strip text.

### DIFF
--- a/.changesets/1787864262-fix-agent-pane-consolidate-reconnecting-compacting-status-in-z97j.md
+++ b/.changesets/1787864262-fix-agent-pane-consolidate-reconnecting-compacting-status-in-z97j.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): consolidate Reconnecting…/Compacting… status into the Working row
--- a/docs/specs/SPEC_REMOVE_AGENT_UNRESPONSIVE_DETECTION_2026_08_25.md
+++ b/docs/specs/SPEC_REMOVE_AGENT_UNRESPONSIVE_DETECTION_2026_08_25.md
@@ -1,11 +1,13 @@
 # SPEC: Agent-pane status cleanup — remove "unresponsive" detection, consolidate Reconnecting/Compacting/Working
 
 **Date:** 2026-08-25
-**Status:** Part 1 (removal) implemented, verified (cargo test + vitest +
-tsc all green), pending PR/merge. Part 2 (status consolidation) scoped with
-an explicit design recommendation, not yet implemented. Two
-separately-implementable pieces in one doc — either can ship as its own PR
-(per repo-owner direction when this spec was extended).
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2830, opened 2026-08-28
Camper@claudius: fix(swarm): bounded backfill-naming backlog pass for historical subagents/dispatches

### LINKED ISSUE #2829: Camper@claudius: Swarm pane still shows raw slugs instead of named rows — backfilled/abandoned subagents have no bulk naming path (4th report)
## Summary

User-reported (verbally, not previously filed as an issue) for the 4th time: the Swarm pane shows long lists of raw, unnamed subagent slugs instead of clean, individually-named/expandable tool-call rows.

## Prior fix attempts (each closed one leak, none closed the root gap)

1. **#2226** (~mid-July) — `NameGroup` falls back to raw `slug` when `display_name` hasn't resolved yet.
2. **#2231 / #2232** (2026-07-19, `SPEC_SWARM_DISPATCH_NAMING_AND_ROW_MODEL_2026_07_19.md`) — superseded #2226: eager per-dispatch Haiku naming + two-bucket (Agent Tool / Workflow) row model, explicitly meant to end raw-slug rows.
3. **#2340** (2026-07-28) — found raw slugs leaking through 3 *other* entry points (agent-pane transcript chip, standalone subagent pane, ToolBlock button) never wired to the naming system; removed them.
4. **#2677** (2026-08-20, `SPEC_SWARM_DISPATCH_ATTRIBUTION_AND_LIFECYCLE_2026_08_19.md`) — found a lagging Workflow dispatch's members were spilling into the flat Agent-Tool bucket individually instead of one grouped placeholder row; fixed.

## Live evidence this is still happening (2026-08-28, v0.55.27 / bc3c0a3b1)

Via `muxlog swarm` against the currently-running instance:

```
01:36:17  backfilling session subagents on pane (re)open
01:36:17–18  subagent spawned  x48   (all within ~1s)
01:36:18  subagent reconciled: active -> abandoned (parent turn ended)  x48
01:36:18  reconcile_stale_subagents: pass complete
```

Searching the **entire day's** srv log for any naming activity (`naming`, `display_name`, `dispatch_name`, `haiku`, `ambient`) returns **zero matches**, despite these 50 spawn/reconcile events.

## Root cause hypothesis

Eager naming (`SPEC_SWARM_DISPATCH_NAMING_AND_ROW_MODEL_2026_07_19.md` §3) is deliberately guarded to **never fire during backfill** — this guard exists on purpose, to avoid repeating the 2026-07-17 OOM-storm incident (`docs/retro/retro-subagent-backfill-storm-oom-2026-07-17.md`) where naming every backfilled subagent on every restart caused runaway Haiku-call storms.

The only remaining naming path is the on-demand fallback still present at `frontend/app/view/swarm/swarm-view.tsx:804` (`subagent.GenerateName`), which fires **per-row, only on individual click-to-expand**.

Net effect: any subagent discovered via backfill (i.e. **every pane reopen with prior history** — confirmed live above, 48 subagents in under 2 seconds) shows as a raw, unnamed slug row by default and stays that way unless a human clicks each row individually. There is currently no bounded/rate-limited bulk-naming path for backfilled data — only "never" (eager, blocked) or "one at a time by hand" (on click).

## Ask

Don't patch another individual leak of this. Design (or confirm rejecting) a **bounded, rate-limited bulk-naming pass** for backfilled/abandoned subagents — e.g. fired once when a human actually opens the Swarm pane (not on every headless srv restart/backfill), reusing the existing `pull_call_semaphore`-capped Haiku call path (cap 2 concurrent) so it can't reproduce the 07-17 OOM storm, but actually clears the backlog instead of leaving it permanently unnamed.

GenericAgentX-asaf:
## Summary

Fixes #2829 — the Swarm pane's raw-slug problem has been patched narrowly four times before (#2226, #2231/#2232, #2340, #2677), each closing one leak without addressing the structural gap: **eager Haiku naming only ever fires for live spawns.** Backfilled/historical subagents (discovered on every agent-pane reopen) have no bulk naming path — only "never" (eager naming is deliberately blocked during backfill, to avoid repeating the 2026-07-17 OOM-storm incident) or "one at a time by hand" (the per-row on-click fallback, which never covers the default collapsed view).

Confirmed live via `muxlog swarm` before this fix: 48 subagents backfilled + abandoned in under 2 seconds on a real pane reopen, and **zero** naming activity logged for the entire day despite that.

## What changed

- `select_unnamed_backlog` (new, `subagent_watcher/query.rs`) — finds currently-unnamed solo subagents + Workflow dispatches, most-recently-active first, and atomically claims each in the existing `naming_triggered` dedup set (same structure the live eager path already uses — can never double-name or race a live spawn of the same dispatch).
- `resolve_unnamed_backlog` — bounded burst (`BACKLOG_NAMING_BATCH_LIMIT = 20` per call), gated through a **new, dedicated `backlog_naming_semaphore()` (cap 1)** — never the shared `pull_call_semaphore` live user-facing calls use.
- New RPC `("subagent", "ResolveUnnamedBacklog")`, fire-and-forget.
- Frontend: `SwarmViewModel`'s constructor fires this RPC once — this only runs when a human actually opens the Swarm pane (never from the headless per-agent-pane backfill scan). No other frontend changes needed — the existing `subagent:named`/`dispatch:updated` broadcasts already patch state in place.
- The original backfill-skip guard (`live` gate in `process_jsonl_change`) is untouched — this adds a parallel, bounded path rather than weakening it.

## Test plan

- [x] 5 new Rust tests (`select_unnamed_backlog`'s ordering, dedup-on-reselect, limit, already-claimed exclusion, no-visible-member edge case) — all passing, plus the existing 75 `subagent_watcher` tests unaffected.
- [x] 2 new frontend tests (RPC fires exactly once on `SwarmViewModel` construction; existing `subagent:named` handler still patches state) — all passing, plus the existing 92 swarm frontend tests unaffected.
- [x] `cargo check` (whole crate) and `npx tsc --noEmit` both clean.
- [x] Built the backend, launched an isolated `task dev` instance on this branch, called the new RPC directly via the dev `authkey.dev` file against the live srv — `{"success":true}`, zero errors/panics logged. Torn down cleanly afterward; never touched any other running instance.
- [ ] Not yet reproduced: the exact "N backfilled rows visually resolve from raw slug to a real name within seconds of opening the Swarm pane" scenario end-to-end with real subagent history — the unit tests deterministically cover the underlying logic, but a live visual confirmation would need either a throwaway dev instance seeded with real Task-tool history, or checking this on a real build. Flagging so reviewers can decide if that's required before merge.

### DIFF
--- a/.changesets/1787939804-fix-swarm-bounded-backfill-naming-backlog-pass-for-historica-27qm.md
+++ b/.changesets/1787939804-fix-swarm-bounded-backfill-naming-backlog-pass-for-historica-27qm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(swarm): bounded backfill-naming backlog pass for historical subagents/dispatches
--- a/agentmux-srv/src/backend/subagent_watcher/query.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/query.rs
@@ -246,14 +246,136 @@ impl SubagentWatcher {
                     &watcher,
                     &dispatch_id,
                     &first_member_agent_id,
+                    crate::server::app_api::session::pull_call_semaphore(),
                 ).await;
             } else {
                 crate::server::app_api::session::generate_subagent_name(
                     &watcher.wstore,
                     &watcher,
                     &first_member_agent_id,
+                    crate::server::app_api::session::pull_call_semaphore(),
                 ).await;
             }
         });
+    }
+
+    /// Select up to `limit` currently-unnamed subagents/dispatches for the
+    /// bounded backfill-naming pass (`resolve_unnamed_backlog`), most
+    /// recently-active first, and atomically claim each selected item in
+    /// `naming_triggered` before returning it — reusing the exact dedup
+    /// structure the live eager-naming path already uses
+    /// (`jsonl::process_jsonl_change`), so this can never double-name
+    /// something, race with a live spawn of the same dispatch, or need a
+    /// second `HashSet`. Because the claim is permanent, calling this again
+    /// (a second Swarm-pane-open before an earlier batch finishes, or after
+    /// it drains) never re-selects anything already claimed — this is what
+    /// makes `resolve_unnamed_backlog` safe to fire on every pane open with
+    /// no extra debounce.
+    ///
+    /// Deliberately never holds `sessions`/`dispatches` and `naming_triggered`
+    /// locked at the same time — mirrors `process_jsonl_change`'s own
+    /// convention (see its "Mutex released here" comment).
+    pub(super) fn select_unnamed_backlog(&self, limit: usize) -> Vec<BacklogNamingItem> {
+        let mut candidates: Vec<(u64, BacklogNamingItem)> = {
+            let sessions = self.sessions.lock().unwrap();
+            sessions
+                .values()
+                .flat_map(|s| s.subagents.values())
+                .filter(|state| {
+                    state.info.display_name.is_none() && state.info.dispatch_id.starts_with("solo:")
+                })
+                .map(|state| {
+                    (
+                        state.info.last_event_at,
+                        BacklogNamingItem::Solo { agent_id: state.info.agent_id.clone() },
+                    )
+                })
+                .collect()
+        };
+
+        let unnamed_workflows: Vec<(String, u64)> = {
+            let dispatches = self.dispatches.lock().unwrap();
+            dispatches
+                .values()
+                .map(|state| &state.info)
+                .filter(|info| info.kind == DispatchKind::Workflow && info.dispatch_name.is_none())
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Good catch — fixed in c41ef736d. The constructor's one-shot `ResolveUnnamedBacklog` call only covered what was already backfilled by pane-open time; a backfill landing while the Swarm pane is already open now also re-fires it, at the end of `scheduleLoadSubagents`'s existing debounced reload. Safe to call repeatedly since `naming_triggered` claims make an already-drained backlog a cheap no-op. Added a regression test (`re-fires ResolveUnnamedBacklog when a later reload discovers newly backfilled rows on an already-open pane`).

PR #2831, opened 2026-08-29
fix(statusbar): backend uptime from a monotonic clock, not two wall-clock stamps

AgentA-asaf:
## The bug

The status bar's backend uptime renders as a padded negative that appears to count down — e.g. `-59:0-14`, ticking `0-14` → `0-13` → `0-12`.

Reproduced live on the released `0.55.26`, not synthetically: this machine's clock read **2081-02-05** when the app started and was later corrected to **2026-08-29**, with no restart in between — a single `agentmuxsrv starting` line spans both eras in one log file (`agentmuxsrv-v0.55.26.log.2081-02-05`, 410 lines stamped 2081 followed by 54 stamped 2026). `uptimeSecs` sat at about `-1.7e9`.

## Root cause

Uptime was never reported by the backend at all. The frontend derived it by subtracting **two independent wall-clock stamps** taken at opposite ends of the backend's life:

- `backend_started_at` — `chrono::Utc::now()`, stamped once by the host in `agentmux-cef/src/sidecar.rs` when it spawns the sidecar, never re-stamped.
- the sysinfo tick's `ts` — `SystemTime::now()`, fresh every tick.

That subtraction is only correct while the system clock stays monotonic for the backend's whole lifetime. Any **backwards** step — NTP correction, manual set, VM resume — makes it negative and leaves it that way until the next restart.

Two unguarded formatters then compounded it. `formatUptime` had no sign guard, so `secs % 60` stayed negative, and `pad2`'s `n < 10 ? "0" + n : n` prepended a zero to the **minus sign** rather than to a digit. The countdown appearance is real: as `ts` advances the seconds field genuinely moves toward zero, but with the minus inside the padded field the visible digits shrink.

Running the real code against the real timestamps:

```
2026-08-29T06:32:55Z -> uptimeSecs = -1717984694 -> "-59:0-14"
2026-08-29T06:32:56Z -> uptimeSecs = -1717984693 -> "-59:0-13"
2026-08-29T06:32:57Z -> uptimeSecs = -1717984692 -> "-59:0-12"
```

## The fix

Stop subtracting wall-clock stamps. `srv` now reports `uptime_secs` from a **suspend-aware monotonic** counter anchored at process start (`sysinfo::mark_process_start`, called from `main.rs` before anything else can take time), and the frontend displays that value directly.

Uptime needs two properties at once and no single obvious clock has both — wall clock counts suspend but moves backwards on a clock step; `std::time::Instant` cannot move backwards but **stops while the machine is asleep**. This PR's first version used `Instant`, which fixed the reported bug but silently under-reported by the whole suspend interval on every lid-close — a regression against the behaviour it replaced. That was codex's P2, fixed in `98023edbf`. Each platform's suspend-aware counter is now used directly:

| Platform | Source | Why not the obvious one |
|---|---|---|
| Linux/Android | `CLOCK_BOOTTIME` | `CLOCK_MONOTONIC` does not count suspend |
| macOS/BSD | `CLOCK_MONOTONIC` | Darwin's continues across sleep, unlike `CLOCK_UPTIME_RAW` and unlike the `mach_absolute_time` behind `Instant` |
| Windows | `GetTickCount64` | `QueryUnbiasedInterruptTime` does not count suspend |

All three count from boot rather than process start, so the baseline is captured once and subtracted. `saturating_sub` makes a decrease structurally impossible even if a platform counter misbehaved. No Cargo.toml change: `libc` is already a dependency and `windows-sys` already enables `Win32_System_SystemInformation`.

Kept deliberately narrow otherwise:

- **Wire compatibility** — `uptime_secs` is `Option<u64>` with `skip_serializing_if`, so per-block stats and the CPU-stream RPCs keep their exact prior payload shape. A payload predating the field still deserializes (`None`).
- **Fallback, not a second source of truth** — the old wall-clock path survives only for a payload without the field (e.g. one replayed from the sysinfo persist ring across an upgrade), and it is now clamped at zero.
- **Defence in depth** — `formatUptime` clamps negative and non-finite input to zero. That guard is in the formatter rather than only at the call site because the crash popover formats a *host*-supplied `backendDeathInfo.uptime_secs` with exactly the same exposure.

Pure logic moved to `frontend/app/statusbar/backend-uptime.ts` for direct unit coverage, following `disk-volumes.ts` next to it.

## What isn't claimed

Neither property that matters — immunity to a clock step, and counting across suspend — can be asserted in a unit test: the platform counters are deliberately unsettable, and a test cannot suspend the machine. The tests cover what is assertable: the start marker is idempotent, the value never runs backwards, the counter is non-zero and in plausible millisecond units (a `clock_gettime` failure or a ns/us mix-up fails there), the serialization contract holds both ways, and the formatter/resolver handle every negative and non-finite case including the exact live `-1717984694`.

The Linux path is compiled and exercised by CI's ubuntu job. **macOS is not in this repo's PR checks**, so that path is written against documented Darwin semantics and is not machine-verified here.

Also unchanged: the uptime display is still gated on `startedAt() != null`, so a failed `getBackendInfo()` hides uptime even though it is now known independently. Pre-existing behaviour, left alone to keep this diff scoped.

## Verification

- `cargo test -p agentmux-srv` — 2790 passed, 0 failed (Windows)
- `npx tsc --noEmit` — clean
- `npx vitest run` — 3260 passed, 3 skipped, 0 failed

New: 13 frontend tests, 4 backend tests.

### DIFF
--- a/.changesets/1787986039-fix-statusbar-report-backend-uptime-from-a-monotonic-clock-s-7mig.md
+++ b/.changesets/1787986039-fix-statusbar-report-backend-uptime-from-a-monotonic-clock-s-7mig.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(statusbar): report backend uptime from a monotonic clock so a system clock step can't make it negative
--- a/agentmux-srv/src/backend/rpc_types/misc.rs
+++ b/agentmux-srv/src/backend/rpc_types/misc.rs
@@ -254,6 +254,18 @@ pub struct CommandVarResponseData {
 pub struct TimeSeriesData {
     pub ts: i64,
     pub values: HashMap<String, f64>,
+    /// Backend uptime in seconds, from a MONOTONIC clock (`sysinfo::uptime_secs`)
+    /// — NOT `ts` minus some earlier wall-clock stamp. Only the top-level
+    /// sysinfo tick carries it; per-block stats and the CPU-stream RPCs leave
+    /// it `None`, hence `skip_serializing_if` so their payload shape is
+    /// byte-for-byte unchanged.
+    ///
+    /// Exists because the frontend previously derived uptime by subtracting
+    /// the host's wall-clock `backend_started_at` from this struct's own
+    /// wall-clock `ts`, which goes permanently negative after any backwards
+    /// clock step. See `frontend/app/statusbar/backend-uptime.ts`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub uptime_secs: Option<u64>,
 }
 
 /// Matches Go's `RemoteInfo`
--- a/agentmux-srv/src/backend/sysinfo.rs
+++ b/agentmux-srv/src/backend/sysinfo.rs
@@ -37,32 +37,74 @@ const MAX_INTERVAL_SECS: f64 = 2.0;
 // name-guessing), and "other" (everything else, top-N by commit kept for
 // visibility — Chrome, VS Code, Docker, Windows itself, ...).
 
-/// Monotonic reference point for this backend process's uptime, set once at
-/// startup by `mark_process_start` (`bootstrap.rs`).
+/// Milliseconds from a SUSPEND-AWARE monotonic clock — one that keeps counting
+/// while the machine is asleep, and can never move backwards.
 ///
-/// `Instant` and NOT `SystemTime`/`chrono::Utc::now()` deliberately: uptime
-/// must survive a system clock step. The previous scheme subtracted two
-/// independent wall-clock reads taken at opposite ends of the backend's life
-/// (the host's `backend_started_at` stamp vs. each sysinfo tick's `ts`), so an
-/// NTP correction, a manual clock set, or a VM resume made the difference
-/// negative for the rest of that backend's life. Observed live on 0.55.26: the
-/// app started while the machine's clock read 2081-02-05, the clock was later
-/// corrected to 2026-08-29 with no restart, and the status bar rendered
-/// `-59:0-14`. An `Instant` cannot move backwards, so no clock event can
-/// reproduce that.
-static PROCESS_START: std::sync::OnceLock<Instant> = std::sync::OnceLock::new();
+/// Uptime needs both properties, and no single obvious clock has them:
+///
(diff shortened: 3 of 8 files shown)


a5af:
@codex review

AgentA-asaf:
Addressed in 98023edbf — this was a real regression, thanks.

`Instant` gave the property this PR needed (can't move backwards) but silently dropped one the wall-clock version had (counts suspend). I traded one inaccuracy for another without noticing. Uptime needs both at once, and no single obvious clock has them:

- wall clock — counts suspend, but moves backwards on a clock step (the original bug)
- `Instant` — can't move backwards, but stops while suspended

So each platform's suspend-aware counter is now used directly:

| Platform | Source | Why not the obvious one |
|---|---|---|
| Linux/Android | `CLOCK_BOOTTIME` | `CLOCK_MONOTONIC` doesn't count suspend |
| macOS/BSD | `CLOCK_MONOTONIC` | Darwin's continues across sleep, unlike `CLOCK_UPTIME_RAW` and unlike the `mach_absolute_time` behind `Instant` |
| Windows | `GetTickCount64` | `QueryUnbiasedInterruptTime` doesn't count suspend |

All three count from boot rather than process start, so `mark_process_start` takes a baseline and `uptime_secs` subtracts it. `saturating_sub` makes a decrease structurally impossible even if a platform counter misbehaved. No Cargo.toml change needed — `libc` is already a dependency and `windows-sys` already enables `Win32_System_SystemInformation`.

On your conditional framing ("if this value represents elapsed process lifetime") — agreed that's the design question underneath, and I think wall-elapsed is what a status bar labelled "Uptime" implies. A user who closes the lid at 10:00 and opens it at 14:00 expects uptime to have advanced.

**What I can't claim:** neither property is unit-testable — the platform counters are deliberately unsettable and a test can't suspend the machine. The new test covers what is assertable: non-zero, non-decreasing, and plausible millisecond units, so a `clock_gettime` failure or a ns/µs mix-up fails on whichever platform runs it. The Linux path is compiled and exercised by CI's ubuntu job; **macOS is not in this repo's PR checks**, so that path is written against documented Darwin semantics and is not machine-verified.

Verified on Windows: `cargo test -p agentmux-srv` — 2790 passed, 0 failed.

PR #2832, opened 2026-08-29
fix(sysinfo): keep the chart correct across a backwards system-clock step

AgentA-asaf:
Follow-up to #2831 — same root cause (wall-clock timestamps assumed monotonic), different subsystem. Found while tracing that bug: this machine ran under a **2081** clock, then it was corrected back to **2026** with no restart in between.

## The bug

The sysinfo chart is a time series keyed on each sample's wall-clock `ts` (`SystemTime::now()`, `sysinfo.rs`). `sampleReducer` had two forward-only assumptions, and a backwards clock step corrupted the series permanently:

**1. Pre-step points never age out.**

```js
const cutoffTs = item.ts - intervalMs * (numPoints + 1);
const trimmed = state.filter((d) => d.ts >= cutoffTs);
```

The cutoff is derived from the *new*, earlier `ts`. A point stamped before the step carries a timestamp 55 years larger, so it satisfies `>= cutoffTs` and survives every subsequent trim — for the life of the pane.

**2. Gap detection can't see a backwards jump.**

```js
const gap = item.ts - last.ts;                          // negative
if (gap > intervalMs * 1.5 && gap <= intervalMs * 3.5) { ... }
else if (gap > intervalMs * 3.5) { ... }
```

Both guards are `gap >`. A negative gap falls through to a plain append, so no NaN sentinel marks the discontinuity and the line mark connects straight across it.

`RESET` had both holes too, so a history read spanning the step reproduced them.

## The fix

Both paths now drop points stamped implausibly far ahead of the newest sample, and mark the seam so the line doesn't connect across it.

The threshold reuses the existing `getGapThresholdMs` rather than introducing a new constant, which keeps the "don't overreact to jitter" posture the reducer already has: ordinary slew (a sub-tick backwards nudge) keeps the history, while a real step clears it. There's a test for exactly that boundary — a small backwards jitter must not discard the buffer.

## Scope — what was *not* wrong

Worth stating because it's the natural assumption and it's incorrect: **the x-domain was never the problem.** `sysinfo-plot.tsx` anchors it on the newest point with a fixed window width, so the axis stayed a correct ~2-minute window and the stale points were simply off-screen. The chart never "zoomed out to 55 years."

The visible damage was the unmarked seam, plus — for auto-scaled metrics only (memory, network, disk) — `computeAutoMaxY` iterating the whole array and scaling the y-axis to invisible pre-step samples. **CPU was exempt from that second effect** (`maxy: 100`, `autoMaxY` unset).

I did not change `computeAutoMaxY`; it needs no fix once the stale points are gone.

## What I could not verify

I never observed the rendered chart. `UIScreenshot` is scoped to the calling agent's own pane and the Sysinfo widget is a different pane, so the visual symptom is inferred from the code and the confirmed clock step, not seen. The reducer-level defects are directly demonstrated by the tests.

Also worth knowing for anyone reproducing: the sysinfo persist ring holds 1024 events (~17 min at the default 1 Hz), so a pane reopened long enough after a step rebuilds cleanly from history and looks fine — the damage is most visible in a pane that was open *across* the step. That's a property of the ring, not of this fix; the fix makes both paths correct regardless of timing.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run` — 3259 passed, 3 skipped, 0 failed

12 new tests. Four of them cover the *pre-existing* forward-time behaviour (contiguous append, window trim, zero-order hold across 1-3 missed ticks, forward-break sentinel) specifically so this change can't silently alter it — that path had no test coverage at all before.

`sampleReducer` is exported for direct unit coverage, matching the pattern used elsewhere for pure reducer/decision functions.

### DIFF
--- a/.changesets/1788004345-fix-sysinfo-keep-the-chart-correct-across-a-backwards-system-6oga.md
+++ b/.changesets/1788004345-fix-sysinfo-keep-the-chart-correct-across-a-backwards-system-6oga.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(sysinfo): keep the chart correct across a backwards system-clock step
--- a/frontend/app/view/sysinfo/sysinfo-model.test.ts
+++ b/frontend/app/view/sysinfo/sysinfo-model.test.ts
@@ -1,0 +1,154 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from "vitest";
+import { sampleReducer } from "./sysinfo-model";
+import type { DataItem } from "./sysinfo-types";
+
+const INTERVAL = 1;
+const INTERVAL_MS = 1000;
+const NUM_POINTS = 120;
+
+function sample(ts: number, cpu = 10): DataItem {
+    return { ts, cpu };
+}
+
+function append(state: DataItem[], item: DataItem): DataItem[] {
+    return sampleReducer(state, { type: "APPEND", item, intervalSecs: INTERVAL, numPoints: NUM_POINTS });
+}
+
+function reset(items: DataItem[]): DataItem[] {
+    return sampleReducer([], { type: "RESET", items, intervalSecs: INTERVAL, numPoints: NUM_POINTS });
+}
+
+/** The exact live incident: srv ran under a 2081 clock, then the clock was
+ *  corrected back to 2026 with no restart. */
+const PRE_STEP = new Date("2081-02-05T09:58:22.373Z").getTime();
+const POST_STEP = new Date("2026-08-29T06:28:33.964Z").getTime();
+
+describe("sampleReducer — APPEND, ordinary forward time", () => {
+    it("appends a contiguous sample", () => {
+        const state = [sample(1000), sample(2000)];
+        expect(append(state, sample(3000)).map((d) => d.ts)).toEqual([1000, 2000, 3000]);
+    });
+
+    it("trims samples older than the visible window", () => {
+        const old = sample(1000);
+        const recent = sample(1000 + INTERVAL_MS * (NUM_POINTS + 1));
+        const next = sample(recent.ts + INTERVAL_MS);
+        const out = append([old, recent], next);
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentA-asaf:
Both findings were real and are fixed in 2ea4b2a08. Thanks — the P1 in particular was the fix reintroducing the exact defect it set out to remove.

**reagentx P1 — confirmed, and broader than reported.** Reproduced standalone first: `state=[8000,9000,10000]`, `item=5000` gave `8000 -> 8001 -> 5000`. Root cause was that `staleAheadCutoff` was doing two different jobs with one threshold — deciding whether a step happened, *and* deciding which points may remain. Those need different answers, so they're now separate rules:

- **Retention** is strict (`d.ts < item.ts`, no tolerance). Monotonicity is guaranteed by construction.
- **The threshold** now only decides whether a visible break is drawn, so slew still doesn't scar the chart.

Your point about magnitude was the useful part: I added a property test sweeping ten backwards-step magnitudes across the threshold, and it showed the original was non-monotonic even at **100 ms** jitter, not only in the partial-step band. So the reported case was one instance of a wider hole.

**codex P2 — confirmed, and the root is deeper than the example.** Bracketing a backwards seam with `prev.ts + 1` / `cur.ts - 1` can *never* yield a monotonic result, because the blanks inherit the inversion. And since ring order is insertion order, this isn't limited to a clean two-sample correction — any out-of-order content reproduces it. Took the suggested direction (drop the pre-step segment) and generalised it: `keepStrictlyIncreasing` walks back from the newest sample and drops superseded entries, anchoring on newest-by-arrival because those are the live samples. The `cur.ts < prev.ts` branch is now unreachable, so it's removed rather than left as dead code.

**Two more found while proving those out**, both putting two points on the same x: the seam sentinel could collide with the new sample's own ts when `last.ts + 1 === item.ts`, and a buffered sample sharing the new sample's exact ts was kept alongside it.

Every case above has a test that fails against the previous commit. 18 tests in this file now, up from 12.

Verified: `npx tsc --noEmit` clean, `npx vitest run` 3265 passed / 0 failed.

PR #2833, opened 2026-08-29
feat(agent): know at pane open whether a conversation will resume or start new

Agent2-asaf:
## Summary

Every continuity signal in the agent pane was **retrospective**. The
`agentmux_session_outcome` divider, the `session:resume_failed` banner and the
"Reconnecting…" readout all describe a resume that has *already been attempted* —
and the persistent controller spawns lazily, on the first message. So the only way
to learn a conversation was gone was to type into it and watch the transcript clear.

Two commits: close the detection gap, then move the answer to pane-open time.

## 1. `fix(agent)` — detect when the agent has no access to the prior conversation

`SPEC_AGENT_PANE_HISTORY_ALIGNMENT_2026_08_05.md` §2.1 exempted the
"no `--resume` attempted at all" spawn, reasoning it has *"nothing to lose"*. True
for a brand-new agent; false for the case
`STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md` recorded — a
long-lived named agent opened in a fresh channel whose registry pointer is empty
gets no `--resume`, so the CLI never errors and no outcome is ever emitted, while
`blockfile.rs`'s cross-channel fallback renders the whole prior conversation.

`spawn_process` now emits `Fresh` when no `--resume` was attached, it's the
controller's first generation, and a transcript already exists locally or in the
agent's global zone. `persistent_resume::update`'s `SpawnedFresh` arm stays
effect-free and its regression guards still hold — "does prior history exist" is a
FileStore question the pure state machine can't answer, so the gate lives in
`persistent.rs`.

Also retracts `session:resume_failed` when a recovery scan succeeds.
`mark_resume_failed` fires from the stderr reader *before* recovery runs, so a
successful recovery left the banner claiming the conversation was lost while the
transcript's own divider said `resumed`.

## 2. `feat(agent)` — pane-open resume pre-flight

`--resume` fails for exactly one reason in practice: the session's `.jsonl` isn't
under the `CLAUDE_CONFIG_DIR` the CLI will use. That's a file-existence question, so
`backend/resume_preflight.rs` evaluates the same condition the CLI will, just
earlier — reusing `session_backfill`'s path scheme and home-dir expansion so the two
can't drift. It mirrors `persistent.rs`'s real decision sequence, including
`find_recovery_session_id`'s refusal to recover the id that just failed:

| Verdict | Meaning |
|---|---|
| `resume` | exact session present — silent |
| `recover` | held id dead, retry path will reach a real one → "will reconnect, may pause" |
| `fresh` | next turn starts with none of the prior conversation → warns + Open Agent History |
| `unknown` | no resume flag / no config dir — stays silent rather than guessing |

The no-session-id case is deliberately `fresh` even when a session sits on disk,
because that spawn path never scans for one; the orphaned id rides along as
`recoverable_session_id` — evidence, not a promise. Calling it `recover` would
describe the rehydrate step Part B hasn't built.

Surfaced as a render-time synthetic document node, the shape
`inject-history-link.ts` established. It can't be an `AgentControlBar` banner —
that region only renders while the details panel is open, and a notice you can miss
is worthless. The node stands down the moment a real `session_outcome` exists (a
prediction must never sit beside the retrospective fact) and says nothing on an
empty document, so a new agent's first open stays clean. The progress list only
appears if the check outruns 200ms.

## Spec corrections

§2.1's exemption is marked superseded, and §6.2's claim that Part A already made
this fresh start *"an honest, visibly-labeled event"* is corrected — it did not, and
`spawned_fresh_and_never_confirmed_paths_never_emit_a_session_outcome` pinned the
opposite.

## Testing

- `cargo test -p agentmux-srv` — 2802 passed, 0 failed
- `npx vitest run frontend/app/view/agent` — 118 files, 1507 passed
- `npx tsc --noEmit` — clean

New coverage: 8 tests over the pre-flight verdict table (real tempdir transcript
layouts), 12 over the injection/suppression rules, 4 over the disclosure gate, 3
over the `resume_failed` retraction.

**Not verified in a live pane.** The verdict logic is unit-tested against real
on-disk layouts, but the RPC round-trip and node rendering are code-verified only —
worth a `task dev` pass against an agent known to have lost its pointer.

### DIFF
--- a/.changesets/1787881995-fix-agent-disclose-a-fresh-start-when-a-pane-spawns-with-no--8wep.md
+++ b/.changesets/1787881995-fix-agent-disclose-a-fresh-start-when-a-pane-spawns-with-no--8wep.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): disclose a fresh start when a pane spawns with no --resume but prior history exists
--- a/.changesets/1787897455-feat-agent-tell-the-pane-at-open-whether-its-conversation-wi-oct6.md
+++ b/.changesets/1787897455-feat-agent-tell-the-pane-at-open-whether-its-conversation-wi-oct6.md
@@ -1,5 +1,5 @@
 ---
-type: patch
+type: minor
 ---
 
 feat(agent): tell the pane at open whether its conversation will resume or start new
--- a/.changesets/1787897455-feat-agent-tell-the-pane-at-open-whether-its-conversation-wi-oct6.md
+++ b/.changesets/1787897455-feat-agent-tell-the-pane-at-open-whether-its-conversation-wi-oct6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): tell the pane at open whether its conversation will resume or start new
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -83,6 +83,77 @@ fn session_outcome_line(
     )
 }
 
+/// Should a spawn that attached NO `--resume` disclose itself as a fresh
+/// start (subject to the caller also confirming prior history actually
+/// exists — see `has_prior_transcript`)?
+///
+/// SPEC_AGENT_PANE_HISTORY_ALIGNMENT_2026_08_05.md §2.1 deliberately left
+/// this case unreported, reasoning that a spawn with no session id has
+/// "nothing to lose". That's true for a brand-new agent and false for the
+/// case `docs/status/STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md`
+/// recorded: a long-lived named agent opened in a fresh channel whose shared
+/// registry pointer is empty gets no `--resume` at all, so the CLI never
+/// errors, `persistent_resume` never tracks anything, and no outcome is ever
+/// emitted — while the pane renders the entire prior conversation through
+/// `blockfile.rs`'s cross-channel read fallback. That is exactly the silent
+/// disagreement between "what the pane shows" and "what the model has" that
+/// `EmitSessionOutcome` exists to prevent.
+///
+/// `generation == 1` restricts this to a controller's FIRST spawn. Later
+/// generations also spawn without `--resume`, but each already has its own
+/// disclosure or deliberately has none: `retry_after_resume_failure`'s
+/// no-recovery-candidate path emits `Fresh` itself, and
+/// `respawn_once_for_leftover_queue` restarts a session whose fresh-vs-resumed
+/// status was decided on an earlier generation. Only a first spawn can be the
+/// "pane just opened onto history this process never had" case.
+///
+/// Kept a pure free function (the FileStore lookup stays at the call site) so
+/// the gate is unit-testable without a controller or a spawned process.
+fn fresh_start_needs_disclosure(attempted_resume_sid: Option<&str>, generation: u64) -> bool {
+    attempted_resume_sid.is_none() && generation == 1
+}
+
+#[cfg(test)]
+mod fresh_start_disclosure_tests {
+    use super::fresh_start_needs_disclosure;
+
+    /// The case STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md
+    /// recorded: a pane's first spawn, no registry pointer to resume, prior
+    /// history on disk. Nothing else in the resume machinery reports this.
+    #[test]
+    fn a_first_spawn_with_no_resume_is_disclosed() {
+        assert!(fresh_start_needs_disclosure(None, 1));
+    }
+
+    /// A resume WAS attempted — `persistent_resume`'s own tracking owns the
+    /// outcome from here (Resumed, or Fresh via the retry/recovery cascade).
+    /// Disclosing here too would double-report and could contradict it.
+    #[test]
+    fn a_spawn_that_attempted_a_resume_is_never_disclosed_here() {
+        assert!(!fresh_start_needs_disclosure(Some("some-sid"), 1));
+        assert!(!fresh_start_needs_disclosure(Some("some-sid"), 4));
+    }
+
+    /// Later generations spawn without `--resume` too, but each already has
+    /// its own disclosure or deliberately has none —
+    /// `retry_after_resume_failure` emits `Fresh` itself when no recovery
+    /// candidate exists, and `respawn_once_for_leftover_queue` restarts a
+    /// session already decided on an earlier generation. Re-disclosing would
+    /// stack a second divider onto an unchanged conversation.
+    #[test]
+    fn a_later_generation_respawn_is_not_re_disclosed() {
+        assert!(!fresh_start_needs_disclosure(None, 2));
+        assert!(!fresh_start_needs_disclosure(None, 17));
+    }
+
+    /// Generation 0 never reaches a spawn (`spawn_process` bumps before use),
+    /// but the gate must not treat the sentinel as a first spawn.
+    #[test]
+    fn generation_zero_is_not_treated_as_a_first_spawn() {
+        assert!(!fresh_start_needs_disclosure(None, 0));
+    }
+}
+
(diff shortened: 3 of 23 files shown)


a5af:
@codex review

Agent2-asaf:
> **[P2] Make the preflight provider-aware** — for every non-Claude provider this reads the wrong env var or can't model its resume strategy, so the feature never works for them.

Acknowledged, and the observation is accurate — but I'm going to leave it as-is, because making *only* the preflight provider-aware would make this worse, not better.

The preflight's one job is to predict what the real spawn path will do. That path is Claude-only today:

- `persistent.rs:1611` — `find_recovery_session_id` does `config.env_vars.get("CLAUDE_CONFIG_DIR")?`, hardcoded, returning `None` for anything else.
- `session_backfill.rs` — `find_largest_session_for_working_dir` hardcodes Claude's `projects/<slug>/<sid>.jsonl` layout, and the identity-bundle scan literally `.join("claude")`.

So for Gemini today: `resume_flag` is `-r`, but no `CLAUDE_CONFIG_DIR` is in `cmd:env`, so `find_recovery_session_id` returns `None` and a stale resume falls through to a blank conversation with no recovery attempt. If I taught the preflight about `GEMINI_CLI_HOME` and Gemini's transcript layout, it would find a session on disk and confidently report `recover` — while the actual spawn still went `fresh`. A confident wrong prediction is strictly worse than the current `unknown`, which suppresses the notice and leaves behavior exactly as it is today.

The `unknown` verdict is deliberate for precisely this reason (`resume_preflight.rs`, `Verdict::Unknown`: *"Not determinable… Callers should stay silent rather than guess"*), and it's covered by `an_unknowable_pane_reports_unknown_rather_than_guessing`.

The right shape for this is one change, not two: make `find_recovery_session_id` + `session_backfill` provider-aware off the provider registry, and the preflight inherits it for free — it already routes every path lookup through `session_backfill` specifically so the two can't drift. That's a bigger piece of work than this PR (it changes real resume behavior for three providers, needs its own transcript-layout research per provider, and wants live verification on each), so I'd rather it be its own PR than ride along here.

Filing it as follow-up rather than fixing it here. Happy to be overruled if you'd rather block on it.

Agent2-asaf:
Both review points addressed in `1dcbf2ee7`.

**[P1] `agent:resume_flag` default divergence — fixed, and it was a real bug.**

Verified the claim before fixing: all four real spawn-path readers default to `"--resume"`, and `agent_handlers/input.rs:400` is specifically the line that builds `PersistentSpawnConfig` for the persistent controller — the exact path this preflight predicts. Mine defaulted to `""`.

The consequence is worse than "no verdict": it silently suppressed the notice for any Claude pane created before `agent_open.rs` started writing that key and not respawned since — i.e. precisely the long-lived, opened-across-many-channels panes the feature exists to serve. That's this module's own stated invariant broken (mirror the real spawn path, or predict nothing).

Also correct that nothing could have caught it: the existing tests exercise `preflight` against an explicit `PreflightInput`, which is the part that skips extraction entirely. So the fix is the default *plus* somewhere to assert it — extraction moved into `preflight_input_from_meta`, with six tests pinning each key's default, including that an explicitly-empty flag still survives as empty rather than being back-filled.

**[P2] changeset type — fixed**, `patch` → `minor`. New RPC command + new UI surface is a feature by CLAUDE.md's own table.

**Not addressed: Codex's P2 (provider-awareness).** Reasoning in [my earlier comment](https://github.com/agentmuxai/agentmux/pull/2833#issuecomment-5464324898) — short version: `find_recovery_session_id` and `session_backfill` are themselves Claude-only today, so teaching *just* the preflight about other providers would make it predict `recover` where the real spawn goes `fresh`. Confident and wrong beats silent only if it's right. Wants to be its own PR that makes the recovery path provider-aware first; the preflight inherits it for free, since it already routes every path lookup through `session_backfill`.

`cargo test -p agentmux-srv` — 2808 passed, 0 failed. `npx tsc --noEmit` clean.

Still worth flagging from the original PR description: **this hasn't been verified in a live pane.** The verdict logic is tested against real on-disk transcript layouts, but the RPC round-trip and node rendering are code-verified only.

Agent2-asaf:
Fixed in `7a1c306d6` — and this was a good catch, materially worse than the `resume_flag` one.

Verified before fixing: `resolve_bound_oauth_config_dir` (`inject.rs:275`) is exactly the purpose-built read-only pre-spawn lookup you describe, `subagent_watcher/parse.rs:478` already takes its output for the same question, and it returns `None` for the not-bound / non-OAuth cases — so it drops straight in with `cmd:env` as the fallback.

Why this one bites harder than the previous P1: `resume_flag` being wrong made the preflight fall silent (`Unknown`), which is merely the feature not working. `config_dir` being wrong makes it **answer confidently and inversely** — the entire verdict rests on a transcript-reachability check against that directory, so an identity-bound pane could get a spurious "will start new" warning plus an Open Agent History prompt while the real spawn resumes perfectly well, or get silence while it's about to lose its conversation. A wrong prediction is worse than no prediction, which is the principle the module is built on.

Three more tests pin the precedence: a resolved dir beats a *disagreeing* `cmd:env`, a resolved dir works with no `cmd:env` at all, and an unbound pane still falls back to `cmd:env`.

That's now two review passes finding the same class of defect — the preflight silently disagreeing with the real spawn path about one input. Worth saying plainly: the invariant is easy to state and easy to violate, because each input is read in a different place in the spawn code. If a third turns up I'd rather restructure so the spawn path and the preflight read their inputs through one shared function than keep patching them one key at a time. Flagging it now rather than after the next one.

`cargo test -p agentmux-srv` — 2811 passed, 0 failed.

Agent2-asaf:
## Live verification — done, and it caught the real thing

Built and ran this branch (`task dev`), then drove the running instance over its own WebSocket RPC. Closing out the "code-verified only" caveat I've been carrying on this PR.

**The RPC is registered and executing in a running binary.** Against a nonexistent block it returns this handler's own error, `session:resume_preflight: not found` — versus `unknown command: agent.list_definitions` for a genuinely unregistered command. So the constant, registration, dispatch, arg deserialization and error path all work.

**Real verdict against a real agent.** Opened AgentX (`8e5f7b6d…`) in the fresh dev channel:

```
verdict: fresh
  provider    ok    --resume
  session-id  ok    672f0ee0-0372-4951-8b5f-17538fc78aec
  transcript  FAIL  not under this config dir
  recovery    FAIL  nothing to recover
```

The registry handed over a live, current session id (that's the same id AgentX captured in its own srv log earlier today — the #2755 live-pointer fix working), but that transcript isn't reachable under the config dir this channel would spawn with. So the pane genuinely would start a new conversation.

**And the pane is showing 2,315,257 lines of that conversation** — `blockfile:line_count` on the same block, served by `blockfile.rs`'s cross-channel global-transcript fallback.

That is exactly `STATUS_CROSS_CHANNEL_RESUME_STALE_SESSION_ID_2026_08_20.md` reproduced live: 2.3M lines of history on screen, an agent that has none of it, and — before this PR — no signal at all until you type and watch it clear.

**The hook fires on pane mount.** Two `session:resume_preflight` log lines 18ms apart: my explicit call, and `useResumePreflight`'s own, unprompted. `duration_ms: 0`, so the progress list correctly never appears.

## What I still did not see

**The rendered pixels.** `DiscoverWindows`/`CaptureWindow` only enumerate packaged/portable windows — a `task dev` window is invisible to them, so I couldn't screenshot it. What I verified instead is that all three of `buildResumePreflightNode`'s preconditions hold on real data: verdict `fresh`, document non-empty (2.3M lines), no `session_outcome` node yet. The node construction from that state is covered by the 12 injection tests. So the notice rendering is *inferred from confirmed preconditions*, not observed — and I'd rather say that plainly than call it verified.

Two incidental findings, neither related to this PR:
- The dev build looks for the provider CLI at `~/.agentmux/<version>/cli/` while it's actually provisioned at `~/.agentmux/instances/v<version>/cli/`, so `agent.open` fails with `CLI_NOT_AVAILABLE` on a fresh dev channel. I worked around it locally with a junction (since removed).
- `DiscoverWindows` not seeing dev-layout windows makes cross-instance UI verification impossible for `task dev` builds.

Merging on this.

PR #2834, opened 2026-08-29
lark@claudius: fix(agent-pane): force stick-to-bottom on a pane's first overflow

GenericAgentX-asaf:
## Summary

- Rare bug: a pane that loads with no content, then finally grows past its viewport for the first time, could get stuck at the top instead of following the tail — the user had to scroll manually or start typing to catch back up.
- Spec: `docs/specs/SPEC_AGENT_PANE_FIRST_OVERFLOW_SCROLL_PIN_FIX_2026_08_29.md`. Root cause isn't fully pinned down (this ties into the still-open `FINDINGS_TOOL_CALL_SCROLL_OSCILLATION_LIVE_INSTANCE_DATA_2026_08_22.md` investigation's unexplained whole-pane 0px collapse — a pane recovering from that collapse re-lives exactly this transition), but the fix doesn't need to wait on that: a pane that has never overflowed can't have a legitimate "user scrolled away" state, since there's nowhere to have scrolled away to yet.
- Fix: latch a `hasOverflowedOnce` flag on `AgentViewState`. The first time a pane's `scrollHeight` exceeds `clientHeight`, force stick-to-bottom regardless of what geometry the triggering scroll event reports — closing off the race where a stray/misleading scroll event on that exact transition could disengage before any of the existing re-pin mechanisms (ResizeObserver, itemized signal effect) get a chance to correct it.
- Added a `[wave-scroll-first-overflow]` diagnostic line (same convention as the existing `[wave-scroll]`/`[wave-scroll-shrink]` tracing) so a future live-instance telemetry pull can confirm whether user reports of this correlate with the known 0px collapse or are a distinct mechanism.
- Closed a real test-coverage gap: neither existing scroll test exercised content growing past the viewport for the first time via content-only resize (as opposed to container resize). Added 3 tests to `AgentDocumentVirtualList.resize.test.tsx`; verified the two behavior-asserting tests actually fail without the fix (stashed the fix, confirmed 2/8 failures, restored it).

## Test plan

- [x] `npx vitest run frontend/app/view/agent/virtualization/` — 131/131 passing
- [x] Manually verified the two new regression tests fail without the fix (temporarily stashed `state.ts`/`AgentDocumentVirtualList.tsx`, reran, confirmed 2 failures, restored)

### DIFF
--- a/.changesets/1788034240-fix-agent-pane-force-stick-to-bottom-on-a-pane-s-first-overf-h92j.md
+++ b/.changesets/1788034240-fix-agent-pane-force-stick-to-bottom-on-a-pane-s-first-overf-h92j.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): force stick-to-bottom on a pane's first overflow
--- a/docs/specs/SPEC_AGENT_PANE_FIRST_OVERFLOW_SCROLL_PIN_FIX_2026_08_29.md
+++ b/docs/specs/SPEC_AGENT_PANE_FIRST_OVERFLOW_SCROLL_PIN_FIX_2026_08_29.md
@@ -63,6 +63,34 @@ actually needed — is also fixed (log/act only when the fix changes the
 outcome), but doesn't affect §4's reasoning or invalidate anything above
 the addendum; noted for completeness since it's the same review round.
 
+## Addendum 2 2026-08-29 (same day) — point 3 above was itself incomplete
+
+Reagent's re-review (after the first addendum's fixes shipped) found that
+point 3's own fix didn't actually deliver what it claimed. `markNotOverflowing`
+was reachable ONLY through `scrollToTrueBottom()`, and every content-resize
+call site of `scrollToTrueBottom()` (both `ResizeObserver`s, the itemized
+signal effect) is itself gated on `stickToBottom()` already being `true`.
+So a pane that collapses to non-overflowing WHILE scrolled away — the
+`/clear`-while-reading-history case the doc comment explicitly claimed to
+cover, or the whole-pane 0px collapse happening mid-history-read — never
+re-armed `isOverflowing` at all, because nothing that could observe the
+collapse ever ran while disengaged. The re-arm only actually worked for a
+pane that stayed pinned throughout, which is the one case that's *least*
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed all four review findings in 3d7785ed6:

- **codex P1** (missing `return`, pagination clobbers the pin) — fixed, added a regression test that exercises the exact `captureHeadAnchor`-in-effect scenario.
- **reagent P1** (force-pin didn't check for an already-legitimate disengage, e.g. from older-history pagination) — fixed by gating the force-pin on `stickToBottom()` already being `true`. This also directly resolves codex's P1 by construction: a pagination-disengaged pane now reads `stickToBottom() === false` and the whole branch is skipped, so there's nothing to `return` early from in that path — the explicit `return` still guards the other (still-following) path.
- **codex P2** (latch never re-arms after a collapse) — fixed: `hasOverflowedOnce` is now a bidirectional `isOverflowing` signal that re-arms via `scrollToTrueBottom`'s own shrink handling.
- **reagent P2** (diagnostic logs unconditionally) — fixed: only logs/acts when the raw geometry would otherwise have disengaged (a genuine save).

Spec updated with an addendum documenting all three corrections. 3 new tests added (11 total in the first-overflow suite), all passing; `tsc --noEmit` clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
Good catch — the P1 finding in the follow-up review was right: `markNotOverflowing` was only reachable through `scrollToTrueBottom()`, and every content-resize call site of that function is itself gated on `stickToBottom()` already being `true`. So a pane collapsing to non-overflowing *while already scrolled away* (the `/clear`-while-reading-history case my own doc comment claimed to cover) never actually re-armed `isOverflowing`.

Fixed in afeb50ff7: extracted `syncOverflowState()`, called unconditionally from every geometry-observation point (every scroll event, both ResizeObservers, the itemized effect) — independent of each site's own `stickToBottom()` gate on the actual re-pin action. Added a test for the collapse-while-disengaged case specifically; confirmed it fails without the fix (stashed, reran, restored). Spec addendum updated. 135/135 tests passing, `tsc --noEmit` clean.

PR #2835, opened 2026-08-30
docs: Phases 3 + 4 of SPEC_DOCS_CLEANUP_AUDIT — docs/status sweep + docs/analysis spot-check

AgentX-asaf:
## Summary

Continues the docs-cleanup audit after Smike's Phase 1 (#2766). Executes **§5.3 (Phase 3)** in full and **§5.4 (Phase 4)** as the spot-check that section specifies. Phase 2 is withdrawn per the spec itself; **§5.5's ~847-file audit is deliberately NOT touched** — it's explicitly gated on a decision about appetite/budget.

## Phase 3 — `docs/status/` resolution sweep (§5.3)

**15 files, not the 13 predicted** — two were added after the spec was written.

| Outcome | n | Detail |
|---|---|---|
| **RESOLVED** + PR citations | 4 | srv section handle leak (08-08) & its 08-19 recurrence → **#2666**/**#2673**; cross-channel stale `session_id` (08-20) → **#2755**/**#2693**/**#2776**; stale-resume fix plan (08-23) → §6.1→#2755, §6.2→#2776 |
| **SUPERSEDED** | 1 | attached-task ladder (08-09) → its 08-15 continuation |
| Partially resolved, left open | 1 | attached-task axis (08-15) — #2491/#2492 closed via #2589/#2590/#2681/#2683, but its architectural proposal is unbuilt |
| Content corrected, still open | 1 | messenger integrations (07-07) |
| **HISTORICAL** | 1 | issue/discussion cleanup (07-24) — its one open item, PR #2291, merged the next day |
| Left explicitly open + dated note | 3 | app-freeze (08-22), lifecycle+crash (07-16), lifecycle/OOM (06-30) |
| Already resolved & already citing PRs | 4 | untouched |

**Two findings worth carrying forward:**

1. **Docs went stale *internally*, not just against code.** `STATUS_MESSENGER_APP_INTEGRATIONS` §4 listed Slack/Telegram/WhatsApp as "Not started" while its own §1 table and update block recorded all three as shipped (#2026/#2022/#2028). A sweep that only diffs docs against *code* would miss this entirely.
2. **"RESOLVED" was under-propagated across linked docs.** The 08-19 recurrence doc opened by calling the 08-08 doc "still open, unpatched" long after #2666 fixed it.

**Deliberately not done:** per-item verification of the two in-flight architecture programs. A docs-status sweep is the wrong instrument for auditing an active program, and guessing is this audit's own §2.2 failure mode. Both carry an explicit note saying what was and wasn't checked.

## Phase 4 — `docs/analysis/` spot-check (§5.4)

Finding 2 above is exactly the trigger §5.4 was gated on, so it ran. **A spot-check, not a sweep** — the directory holds **132 files**.

**Sample of 8**, relevance-weighted rather than random: the scroll-oscillation chain, plus four unrelated bug analyses included specifically to test whether the pattern generalises beyond the cluster already known to be active. **4 hits. It generalises.**

- `ANALYSIS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_17` — still said its §6 question "needs a live repro this agent cannot drive"; that repro has since run **twice**, unlinked. Forward pointers added — while being explicit the doc is **not** resolved (root cause still open, #2648/#2718); only the blocker moved.
- `FINDINGS_..._2026_08_21` — dead-ended at 8 events from one pane, no pointer to the ~100× larger two-instance dataset a day later. Marked superseded *in scale*; conclusions not retracted.
- `ANALYSIS_BROWSER_PANE_BLACK_FREEZE_MACOS_2026_06_24` — **no resolution note at all**, read as an open bug. Fixed by **#1769** the same day it was written, plus **#1778**. A retro on that fix has been sitting in the same directory the whole time, unlinked.
- `ANALYSIS_MUXBUS_WINDOWS_SIGNIN_URL_TRUNCATION_2026_07_03` — cited an in-flight **branch** instead of a merged PR. Shipped as **#1938**; verified live in `util.rs`'s `open_browser`.

**The finding that matters for §5.5:** the doc that owns a bug and the doc that records its fix can both be individually correct while nothing links them. **Three of these four hits were invisible to a file-vs-code check** — any batch process for the full audit has to check cross-document linkage too.

4-of-8 on a small relevance-weighted sample is **not** a clean estimate for all 132 files and shouldn't be quoted as one; that caveat is written into the spec.

## Notes

- Docs-only. No changeset — a status/resolution correction carries no release note.
- Touches `SPEC_DOCS_CLEANUP_AUDIT_2026_08_22.md` (§5.3/§5.4 completion records). #2766 touches §2.1 of the same file; different sections, should merge cleanly either order.

### DIFF
--- a/docs/analysis/ANALYSIS_BROWSER_PANE_BLACK_FREEZE_MACOS_2026_06_24.md
+++ b/docs/analysis/ANALYSIS_BROWSER_PANE_BLACK_FREEZE_MACOS_2026_06_24.md
@@ -1,13 +1,23 @@
 # Browser Pane: Black Appearance + UI Freeze on macOS/Linux
 
-> **RESOLVED — fixed the same day this was written** (resolution note added
-> 2026-08-29, docs-cleanup Phase 4; the document previously carried no
-> indication it had been fixed at all):
-> - **#1769** `fix(browser-pane): black flash + UI freeze on macOS/Linux
->   when opening a browser pane` — 2026-06-24, hours after this analysis.
-> - **#1778** (`fix/browser-pane-deferred-bounds-macos`, merged 2026-06-27)
->   — follow-ups: deferred overlay bounds, and resizing
->   `NativeWidgetMacNSWindow` via ObjC for the residual macOS black screen.
+> **RESOLVED — but on two different dates for the two platforms**
+> (resolution note added 2026-08-29, docs-cleanup Phase 4; the document
+> previously carried no indication it had been fixed at all):
+> - **Linux — partially fixed 2026-06-24** by **#1769**
+>   (`fix(browser-pane): black flash + UI freeze on macOS/Linux when opening
+>   a browser pane`), hours after this analysis was written. Per
+>   `RETRO_BROWSER_PANE_MACOS_FIX_2026_06_26.md`, #1769 fixed an ordering
+>   bug, which helps only where `set_size`/`set_position` actually work.
+> - **macOS — fixed 2026-06-27** by **#1778**
+>   (`fix/browser-pane-deferred-bounds-macos`). On macOS
+>   `set_size`/`set_position` are **no-ops**, so after #1769 the bounds were
+>   still wrong, the black screen persisted, and CEF's Views mouse-event
+>   routing still used uncommitted full-window bounds — i.e. the
+>   full-window input interception this document reports. #1778 added
+>   deferred overlay bounds plus an ObjC `NativeWidgetMacNSWindow` resize.
+>
+> Despite the filename, **do not read #1769 as having closed the macOS
+> half** — it did not.
 >
 > A retrospective on the fix exists alongside this file:
 > `docs/analysis/RETRO_BROWSER_PANE_MACOS_FIX_2026_06_26.md`. It was never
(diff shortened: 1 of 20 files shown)


a5af:
@codex review

AgentX-asaf:
All three fixed in `0706607b1` — every one correct, and all three the same class of mistake this audit is *about*: a resolution note more confident than its evidence.

**1. macOS resolution date.** You're right, and the retro says so plainly (`RETRO_BROWSER_PANE_MACOS_FIX_2026_06_26.md` L45): #1769 only *partially* fixed Linux, because on macOS `set_size`/`set_position` are no-ops — so bounds stayed wrong and CEF's Views mouse-event routing kept using uncommitted full-window bounds, which **is** the full-window input interception that document reports. Split into two dated entries (Linux 06-24 / macOS 06-27 via #1778), with an explicit warning not to read #1769 as closing the macOS half despite its title naming both platforms.

**2. "whole-window collapse".** Correct, and worse than a wording slip — I checked the source and the 08-22 doc's own §3 heading reads "whole-**pane** collapse to 0px, multiple panes at once", explicitly hypothesising an app-wide event without establishing one. I promoted panes to the window *and* a hypothesis to a finding. As you say, that would have aimed the next investigation at window-lifecycle code instead of pane measurement state. Now describes what was actually logged: multiple panes' `scrollHeight` hitting exactly 0px within milliseconds (3 panes in one incident, 2 in another), mechanism explicitly unknown. Fixed in both banners.

**3. Telegram row.** Correct, and the second half of your point is the sharper one: my banner asserted "only the Bridge status column was wrong," which made the rest of the row read as *verified* when I hadn't checked it. The banner now says the right-hand column is a mix — genuine operational risks of shipped bridges alongside pre-implementation planning language — and the Telegram cell carries an inline superseded marker rather than being deleted, so the original planning record survives.

Worth noting for the audit's own record: three of the four Phase 4 hits were docs whose *fix existed but wasn't linked*. These three findings are the mirror image — my *links* were right and my *claims* about them were overconfident. Both failure modes belong in the §5.5 guidance.

PR #2836, opened 2026-08-30
docs(report): why agent pane load is still glitchy on 0.55.26

AgentA-asaf:
Live-traced analysis of why agent-pane load is still visibly glitchy on `0.55.26` — the first build containing **all four** prior fixes in this family (#2770, #2773, #2781, #2801). Docs only, no code.

## Why this exists

The 08-25 architecture analysis concluded the flicker was most likely a *release gap* — fixes merged but untagged. `v0.55.26` shipped with all four and the symptom survived, so that explanation is now dead and this is the replacement.

Method is a live trace of a real pane reopen (host `[fe]` console bridge correlated against the `agentmux-srv` log), not a synthetic repro and not doc archaeology.

## What it finds

**The performance half is genuinely fixed.** Same pane, same measurement as the 08-23 report:

| Metric | 0.55.21 | 0.55.26 |
|---|---|---|
| `ListActive` + `ListDispatches` calls | ~310 | **7** |
| Worst RPC round-trip | 7201 ms | **56 ms** |
| Launch → settled | ~14 s | **~1.9 s** |

**The visual half is untouched, for a reason none of the four fixes addressed.** On every reopen the backend inserts ~200 replayed subagents into its own live `Active` set, then `reconcile_stale_subagents` retracts all 200 about 314 ms later — while the same load is still in flight. So `subagent.ListActive`, the exact RPC the dock's rows are built from, genuinely returns up to 200 rows that the backend itself is about to disown.

The dock isn't glitching. Every component is behaving correctly; the composition is what's broken. Every fix so far has changed *when the frontend samples* that state — debounce, then an app-wide gate, then a spinner gate — rather than the fact that **the state itself is wrong during the sampling window.** That's why each fix almost works and each one needs a new timeout: §3 counts four independent mechanisms now approximating the one fact the backend never sends.

**A separate measured finding (§5):** `output.idx` is fully rebuilt every 30 seconds, indefinitely — 37 rebuilds in 19 minutes, mean 3168 ms, a **10.1% duty cycle** of full-file rescan on a 759 MB transcript, driven by `useSnapshotPersistence`'s line-count poll whose RPC times out at 3000 ms, so most of that work is discarded and repeated 30 s later.

## Verification

All three findings re-verified against `main` at `f2aca1f09` immediately before opening this PR — `jsonl.rs:125` still inserts `SubAgentStatus::Active` for a replayed spawn, `scan.rs:265` still retracts it, `blockfile.rs` still has no `spawn_blocking`, and `useSnapshotPersistence` still polls on a 30 s interval.

## What it deliberately does not claim

§6 is explicit, because the two prior reports in this family were both corrected for over-attribution:

- **Why three refresh rounds got through rather than one.** The gate logic reads correctly and event delivery was verified to reach it. The logs cannot distinguish "gate never engaged" from "gate engaged and two unrelated post-settle events fired." Not on the critical path — §2 makes all three rounds return the same correct answer.
- Whether the long-tasks come from dock churn or history replay — not isolated.
- No claim that any prior fix should be reverted. §1's numbers show they work; the argument is that three become *unnecessary* after §2.

## Follow-ups

The §2 and §5-caller fixes are being done now as separate PRs. §4 (a participant-registered paint gate replacing the `ready()` boolean that 26 independent subscriber modules must each be taught about) and the deeper index work are tracked rather than bundled.

Related: discussion #2375 (process/turn-liveness consolidation) — §2 is a concrete instance of exactly the pattern that thread describes, two mechanisms disagreeing about whether something is alive.

### DIFF
--- a/docs/reports/REPORT_AGENT_PANE_LOAD_RENDER_ARCHITECTURE_2026_08_27.md
+++ b/docs/reports/REPORT_AGENT_PANE_LOAD_RENDER_ARCHITECTURE_2026_08_27.md
@@ -13,12 +13,20 @@ reopening at `19:47:33Z` today on `0.55.26`, block
 correlating the host `[fe]` console bridge against the `agentmux-srv` log, plus
 a direct read of the current gate/transport code on `main` (`84cc072d4`). Not a
 synthetic repro, and not doc archaeology.
-**Status:** Analysis only when written. §2 and §5 are established by direct
-evidence; §3 and §4 are structural findings from the code; §6 lists what this
-report deliberately does **not** claim.
+**Status:** Analysis only when written; §2 and §5 have since been fixed and
+released (see the 2026-08-30 update below), §4 remains open. §2 and §5 are
+established by direct evidence; §3 and §4 are structural findings from the
+code; §6 lists what this report deliberately does **not** claim.
 
 **Update 2026-08-29** — all three findings were re-verified against `main` at
 `f2aca1f09` and still hold.
+
+**Update 2026-08-30 — two of the three findings are now FIXED and RELEASED.**
+Re-checked against `main` at `5936f8c98`; §2 and §5 are no longer open work,
+§4 is untouched. Read the per-finding status list below rather than the body
+prose, which still describes the 0.55.26 symptom in the present tense
+throughout — the diagnosis is preserved as written (it is the record of why
+these fixes were made), not rewritten to past tense.
 
 **This report has been substantively corrected since it was written**, unlike
 its predecessors in this family where only a status line changed. Codex's review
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentA-asaf:
Both correct, both fixed in `29376ef71`. The first one is the more serious of the two and I want to name it plainly.

**P2 (1) — unsupported attribution.** I checked your arithmetic against the trace and it holds: reconciliation finished at `33.954`, the earliest `ListActive` was reported at `34.070` with a 51 ms round-trip so it cannot have been issued before ~`34.019`, and every subsequent RPC is later still. Every fetch in this trace sampled post-settlement. Your point about the `subagent spawned` lines is also right — `scan_subagents_dir` processes files synchronously before calling `reconcile_stale_subagents`, so those later lines aren't insertion timestamps and can come from the 500 ms workflow-activity flush.

So the report claimed a causal link it did not demonstrate. What survives is the transient itself, which is code-verified (`jsonl.rs:125` inserts `Active`, `scan.rs:265` retracts it) — but "this is the flicker the user sees" is now stated as a hypothesis, with the specific evidence that would settle it: a trace catching a fetch or render inside the pre-reconciliation window.

Worth stating that this is the *same* over-attribution error the two prior reports in this family were corrected for — a fact this report cites in its own §6 as the reason to be disciplined, and then didn't live up to. The header now says the report was substantively corrected rather than claiming only a status line changed, so a future reader isn't misled about which parts were reviewed.

**P2 (2) — the deletable count.** Also right, and this one would have caused harm if acted on. Being born `Abandoned` doesn't suppress the `is_new` path, so the backend still emits up to 200 `subagent:spawned` events per reopen. Corrected to **two** of the four: the backfill gates (`useSubagentBackfillGate`, `backfill-tracker`) lose their job; `createDebouncedRefresh` is what keeps those 200 events from becoming the §1 request storm and must stay; `shellStatusCorrection` addresses an unrelated shell-status race that §2 doesn't touch. Suppressing the replay broadcasts themselves is a separate, larger change and is explicitly not proposed here.

Corrections are marked inline at each site (§TL;DR, §1, §2, §4, §6, §7) rather than silently rewritten, so the original claim and its retraction are both visible.

**Scope note:** the §2 fix is already in progress on `agenta/fix-backfill-false-active-subagents`, and it's justified on the transient itself — a state model asserting something it must immediately retract, plus 200 spurious abandon broadcasts per reopen — not on the unproven link to the visible symptom. Your first finding narrows the *claim*, not the value of removing it.

AgentA-asaf:
The P0 was **correct about what it saw**, and is already resolved — flagging rather than dismissing, since the review still shows CHANGES_REQUESTED and blocks this PR.

What happened: a `git add -A` on my side swept `agentmux-srv/src/backend/subagent_watcher/tests.rs` (131 lines of backend tests asserting insert-time abandon behaviour, which belongs to #2837) into this docs-only branch. Those tests genuinely did assert unimplemented behaviour and would have broken `cargo test`, exactly as described.

I caught it independently and force-pushed the cleaned branch before this review landed. The current head is the same sha the review is attributed to, but the PR's diff no longer contains that file:

```
$ gh api repos/agentmuxai/agentmux/pulls/2836 --jq '.changed_files'
1
$ gh api repos/agentmuxai/agentmux/pulls/2836/files --jq '.[].filename'
docs/reports/REPORT_AGENT_PANE_LOAD_RENDER_ARCHITECTURE_2026_08_27.md
```

The two reviews on this sha are 10 seconds apart and contradictory — APPROVED at 02:07:55, then CHANGES_REQUESTED at 02:08:05 — which is consistent with a job queued against the pre-force-push tree being attributed to the new head.

CI on the current head is green (vitest, ubuntu, windows all passing), which wouldn't be the case if the offending tests were still here.

The tests themselves are not lost — they're in #2837 where they belong, alongside the implementation that makes them pass.

Not merging past this myself. If a re-review doesn't clear it automatically, someone with the right permissions will need to decide whether to re-run or override.

GenericAgentX-asaf:
Pushed a status update to this branch (`2f41472b6`) — status overlay only, no finding rewritten. Re-checked against `main` `5936f8c98`.

**Two of the three findings have shipped since this was written**, both in v0.55.28:

| Finding | Status |
|---|---|
| **§2** — replay asserts ~200 false-`Active` subagents, then retracts them | ✅ **Fixed**, PR #2837 (`c58ce9f1a`) |
| **§5(a)** — `output.idx` fully rebuilt every 30 s | ✅ **Fixed**, PR #2838 (`693e6df60`) |
| **§4** — participant-registered paint gate | ⬜ **Still open, not started** |

Verified §2 in current code rather than taking the changelog's word: `process_jsonl_change` no longer inserts `SubAgentStatus::Active` unconditionally — a non-`live` replay reads the parent block's controller status and is born `Abandoned` when the turn is *confirmed* idle, with two deliberate `Active` fallbacks (live spawns always win; an unregistered controller is UNKNOWN, not idle, leaving reconcile's retry path as the authority). The shipped code cites this report by path at that call site.

**§4 is now the report's only remaining ask** — and its stated precondition ("do it after 1 so it is hiding nothing") is met, so it's unblocked. That's the finding carrying the actual architectural argument, so this report is still worth landing as the record for it.

Also flagged one thing worth a second opinion: §6 described the §2 transient as "code-verified" in the present tense and noted the causal link to the visible flicker was unproven, needing a trace inside the pre-reconciliation window. That transient no longer exists — so that confirming trace **can never be captured**. I marked the link as permanently unproven rather than disproven, so nobody goes hunting for evidence that's now unobtainable. Push back if you'd rather word that differently.

Body prose still describes the 0.55.26 symptom in present tense throughout — left deliberately as the record of why these fixes were made, not rewritten to past tense.

One process note: the `CHANGES_REQUESTED` on this PR is **stale** — its P0 was about a Rust test asserting unimplemented behavior, and that test is no longer in the diff (single markdown file now). Needs a re-review to clear.

PR #2837, opened 2026-08-30
fix(subagent-watcher): don't assert false-Active state on backfill replay

AgentA-asaf:
Implements §2 of `REPORT_AGENT_PANE_LOAD_RENDER_ARCHITECTURE_2026_08_27.md` (PR #2836).

## The problem

A cold backfill replays up to `BACKFILL_MAX_FILES` (200) historical transcripts through `process_jsonl_change(..., live: false)`, and every one was inserted into the watcher's live `sessions` map with `SubAgentStatus::Active` — because a spawn record, replayed, looks exactly like a spawn. `reconcile_stale_subagents` then retracted all 200 a few hundred ms later, broadcasting an abandon event for each.

So on every pane reopen the backend asserted state it was about to disown, and for that window `subagent.ListActive` would report rows that were not active.

## The fix

Decide it at insert time, using the **same predicate `reconcile_stale_subagents` already uses** rather than a second independently-invented one — a confirmed-idle parent turn means nothing on disk can still be running.

Two deliberate fallbacks to `Active`, both covered by tests:

- **`live` always wins.** `turn_active` can lag a genuine spawn, and mislabelling a real subagent would break live rows.
- **An unregistered controller is UNKNOWN, not idle.** `scan_session_subagents` can run before the controller registers, so reconcile's existing retry path stays the authority rather than inferring `Abandoned` from absent information.

A replayed transcript containing its own `result` event still lands `Completed` — the file's own evidence outranks the turn-idle inference.

## Scope — what this does NOT claim

Stated up front because the report originally overreached on both points; codex caught them on #2836 and they're corrected there.

- **This is not claimed as the fix for the visible dock flicker.** The trace behind the report does not establish that the frontend ever sampled inside the false-`Active` window — every `ListActive` in it post-dates the reconciliation. The transient is real and code-verified; its link to the observed symptom is a hypothesis.
- **This does not suppress the `is_new` broadcast path.** The backend still emits up to 200 `subagent:spawned` events per reopen, so `createDebouncedRefresh` is still what keeps those from becoming an RPC storm and must stay. Of §3's four mechanisms only the two backfill gates lose their job, not three.

What it does buy, on its own terms: a state model that no longer asserts something it must immediately retract, and 200 fewer spurious abandon broadcasts per reopen.

## Tests

Five new, all in `subagent_watcher/tests.rs`, using the existing `register_stub_controller` helper:

| Test | Pins |
|---|---|
| `backfill_replay_is_born_abandoned_when_the_parent_turn_is_confirmed_idle` | the fix (fails on `main`) |
| `live_spawn_stays_active_even_when_the_parent_turn_reads_idle` | `live` wins — the regression that would break real-time rows |
| `backfill_replay_stays_active_when_the_parent_turn_is_running` | a live turn can own still-running subagents already on disk |
| `backfill_replay_stays_active_when_no_controller_is_registered` | unknown ≠ idle; reconcile stays the authority |
| `a_finished_replayed_transcript_is_completed_not_abandoned` | `result` outranks the turn-idle inference |

Four of the five pass against `main` unchanged — they pin behaviour the fix must *preserve*, so a future change that over-applies the abandon inference fails here.

## Verification

`cargo test -p agentmux-srv` — **2839 passed, 0 failed**, 6 ignored.

One flake observed once during this work, disclosed rather than papered over: `read_task_prompt_extracts_plain_string_content_from_first_line` failed on `create_dir_all` under parallel execution, then passed in isolation and on every rerun. Unrelated to this change — my tests use distinct temp-dir prefixes — and that test's own comment documents earlier flakiness of exactly this kind. Not fixed here; flagging so it isn't mistaken for a new regression.

Relates to discussion #2375 (process/turn-liveness consolidation) — a concrete instance of two mechanisms disagreeing about what is alive.

### DIFF
--- a/.changesets/1788055801-fix-subagent-watcher-replayed-subagents-are-born-abandoned-w-q9py.md
+++ b/.changesets/1788055801-fix-subagent-watcher-replayed-subagents-are-born-abandoned-w-q9py.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(subagent-watcher): replayed subagents are born Abandoned when the parent turn is confirmed idle
--- a/agentmux-srv/src/backend/subagent_watcher/jsonl.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/jsonl.rs
@@ -111,6 +111,36 @@ impl SubagentWatcher {
                     );
                 }
             }
+            // A replayed spawn (`live == false`) describes a subagent whose
+            // transcript already existed on disk when this pane opened.
+            // Inserting it as `Active` asserts something the replay cannot
+            // know, and `reconcile_stale_subagents` (scan.rs) then retracts it
+            // moments later — up to 200 spurious abandon broadcasts per reopen,
+            // and a window where `subagent.ListActive` reports rows the backend
+            // is about to disown. See
+            // docs/reports/REPORT_AGENT_PANE_LOAD_RENDER_ARCHITECTURE_2026_08_27.md §2.
+            //
+            // Decided here using the SAME predicate reconcile uses, not a
+            // second independently-invented one — a confirmed-idle parent turn
+            // means nothing on disk can still be running. Two deliberate
+            // fallbacks to `Active`:
+            //   - `live` always wins. `turn_active` can lag a genuine spawn,
+            //     and mislabelling a real subagent would break live rows.
+            //   - An unregistered controller is UNKNOWN, not idle.
+            //     `scan_session_subagents` can run before the controller
+            //     registers, so this keeps reconcile's own retry path as the
+            //     authority rather than inferring `Abandoned` from absent
+            //     information.
+            let initial_status = if live {
+                SubAgentStatus::Active
+            } else {
+                match crate::backend::blockcontroller::get_block_controller_status(parent_block_id)
+                    .map(|s| s.turn_active)
+                {
+                    Some(false) => SubAgentStatus::Abandoned,
+                    _ => SubAgentStatus::Active,
+                }
+            };
             let state = session.subagents.entry(agent_id.clone()).or_insert_with(|| {
                 SubagentState {
                     info: SubAgent {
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentA-asaf:
Both fixed in `a861d1e59`.

**codex P1 — a real race, and one I introduced.** Walking it through: the backfill scan reads the lagging `turn_active == false` and inserts `Abandoned`; the `live == true` call that follows finds an existing entry so `or_insert_with` never runs; ordinary events never revive a status; and `reconcile_stale_subagents` only ever downgrades `Active` → `Abandoned`. A genuinely running subagent would have read as abandoned until a `result` line happened to arrive. That's strictly worse than the behaviour I was replacing, for exactly the case the previous commit's own comment warned about (`turn_active` lagging a genuine spawn) — I guarded that for the insert path and missed it for the existing-entry path.

Fixed on the principle that decides it: **`Abandoned` is always an inference; watching the file change is a direct observation, so the observation wins.** A live call now promotes an `Abandoned` entry back to `Active`.

That can't strand a wrong answer in the other direction either — if the subagent really has finished, its own `result` event sets `Completed`; if the parent turn really is idle, the next reconcile pass re-abandons it. Convergence, not a standoff.

Deliberately scoped to `Abandoned` only. `Completed` is established by the transcript itself, so a trailing live write must not flip a finished subagent back to running — there's a test for that direction too (`a_live_observation_does_not_resurrect_a_completed_subagent`), because "promote on live" is the obvious over-application of this fix.

**reagentx P2 — correct, and my omission.** The doc comment on `SubAgentStatus::Abandoned` still said `reconcile_stale_subagents` was its only writer, which my own previous commit had falsified. Rewritten to name both writers explicitly and to state the inference-vs-observation precedence rule, since that's what the promotion above turns on — a future reader needs the rule, not just the list, to know which writer wins.

Two new tests (7 total in this PR now): the race itself, and the Completed-must-stick guard.

Verified: `cargo test -p agentmux-srv` — 2841 passed, 0 failed.

PR #2838, opened 2026-08-30
fix(blockfile): stop a 30s line-count poll forcing a full output.idx rebuild

AgentA-asaf:
Implements §5 of `REPORT_AGENT_PANE_LOAD_RENDER_ARCHITECTURE_2026_08_27.md` (PR #2836).

## The problem

`global_zone_line_count` returned the cached entry count only when `output.idx`'s `covered_size` header matched the current `output` size, and **rebuilt from a full scan otherwise**. For a live agent `output` grows continuously, so the header effectively never matches — every call was an O(file-size) rescan.

Its one production caller is the `blockfile:line_count` handler, which `useSnapshotPersistence` polls **every 30 seconds** purely to stamp a snapshot high-water mark.

Measured on a 759 MB / 823k-line transcript:

| | |
|---|---|
| Rebuilds in 19 minutes | **37** |
| Mean duration | **3168 ms** |
| Total scanning | **117 s** |
| Duty cycle | **~10%**, permanent, on the shared srv runtime |
| Caller's RPC timeout | **3000 ms** — *below the mean* |

So most of that work was discarded by a caller that had already given up, then repeated 30 seconds later, forever.

## The fix

A **stale** index is now reused as-is for a count. It trails the true total by whatever was appended since it was built, which is immaterial for a high-water mark and vastly cheaper than the alternative. The `read_range` path builds its own index when it needs a *current* one; this counter never does.

A **missing** index still builds, deliberately. Without it the handler falls through to `session:line_count` (absent for a zone this channel has never written) and then the WPS ring (capped at 4096), so a fresh cross-channel open would under-report and render a near-empty pane — the exact failure the global path was added to prevent (`ANALYSIS_CROSS_CHANNEL_CONVERSATION_HISTORY_2026_06_14.md`). Now paid once per zone instead of once per poll.

**No frontend change.** With the backend cheap, the 30 s poll costs a `stat` plus an 8-byte read, so there's no reason to touch the poller or the snapshot's high-water mark semantics.

## Scope

This does **not** address the two deeper items from §5 — making the index append-only (the header already records covered size, so an append is `scan from covered_size` rather than from 0), and giving `rebuild_output_idx` a `spawn_blocking` offload so a build can't occupy a Tokio worker. Both remain real; the second has been open since `STATUS_CROSS_CHANNEL_AGENT_OPEN_FULL_APP_FREEZE_2026_08_22.md` §7.1. Tracked separately rather than bundled here, since this change removes the pathological *frequency* without touching the build itself.

## Tests

Two new, using the same seeded-wrong-index technique the existing fresh-index test established — assert the cached (deliberately wrong) count comes back, which proves no rescan happened:

- `global_zone_line_count_does_not_rebuild_a_merely_stale_index` — the fix. The old code returns the rescanned `3`; this asserts the cached `2`.
- `global_zone_line_count_still_builds_when_no_index_exists` — the case that must keep paying, plus that the build leaves an index behind.

The two pre-existing tests are unchanged and still pass, pinning the fresh-index reuse and blank-line counting.

## Verification

`cargo test -p agentmux-srv` — **2836 passed, 0 failed**, 6 ignored.

### DIFF
--- a/.changesets/1788056049-fix-blockfile-stop-a-30s-line-count-poll-forcing-a-full-outp-6ogw.md
+++ b/.changesets/1788056049-fix-blockfile-stop-a-30s-line-count-poll-forcing-a-full-outp-6ogw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(blockfile): stop a 30s line-count poll forcing a full output.idx rebuild
--- a/agentmux-srv/src/backend/blockcontroller/shell/indexing.rs
+++ b/agentmux-srv/src/backend/blockcontroller/shell/indexing.rs
@@ -30,6 +30,95 @@ pub(crate) fn rebuild_output_idx(
     block_id: &str,
     output_size: u64,
 ) -> Option<u64> {
+    build_output_idx_from(fs, block_id, output_size, 0, Vec::new(), 0)
+}
+
+/// Bring an existing `output.idx` up to `output`'s current size by scanning
+/// ONLY the appended bytes, instead of rescanning from byte zero.
+///
+/// Falls back to a full rebuild whenever the existing index can't be trusted
+/// as a base: missing, header unreadable, no entries yet, or `output` shrank
+/// below what the index already covers (rotation/truncation).
+///
+/// Why this exists: the index is a pure cache with no incremental mutation, so
+/// a stale one used to mean a full O(file-size) rescan. On a live agent
+/// `output` grows continuously, so "stale" was the steady state — a 30-second
+/// line-count poll drove 37 full rebuilds in 19 minutes on a 759 MB
+/// transcript, mean 3168 ms, ~10% duty cycle
+/// (docs/reports/REPORT_AGENT_PANE_LOAD_RENDER_ARCHITECTURE_2026_08_27.md §5).
+///
+/// Returning the stale count instead was tried and is WRONG: `line_count`
+/// feeds `useHistoryPagination`'s tail window (`offset = total - PAGE_SIZE`),
+/// so an undercount silently drops the most recent history on reopen — codex
+/// P1 on PR #2838. The count has to stay exact; only the cost of keeping it
+/// exact is negotiable.
+pub(crate) fn extend_output_idx(
+    fs: &FileStore,
+    block_id: &str,
+    output_size: u64,
+) -> Option<u64> {
+    const IDX: &str = "output.idx";
+
+    let full = || rebuild_output_idx(fs, block_id, output_size);
+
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentA-asaf:
**codex P1 — you're right, and the approach was wrong, not just the edge case.** Replaced wholesale in `059f90a6a` rather than patched.

I verified your claim before acting on it: `useHistoryPagination.ts:470` takes `total` from this RPC and then computes `offset = Math.max(0, total - PAGE_SIZE)` at line 486. So an index covering 1,000 lines of a now-1,200-line output loads lines 800-999 — the newest history silently missing on reopen. Your note that the stale count also lands in the snapshot high-water mark, which same-block restores never widen, holds too.

**My analysis error was specific and worth naming:** I checked callers of the *Rust function* (one — the `blockfile:line_count` handler) and concluded "one production caller, staleness is immaterial." I never checked callers of the *RPC that handler backs*, which include pagination. The PR body asserted that conclusion confidently. The count has to stay exact; only the **cost** of exactness was ever negotiable.

**What it does now.** The freshness check is restored verbatim, and staleness is answered by a new `extend_output_idx`: anchor on the existing index, scan only the appended bytes, reuse the settled offsets before them. Full rebuild whenever there's no trustworthy base — missing, header unreadable, zero entries, or `output` shrank (rotation/truncation).

**The boundary case that makes the naive version wrong**, since it's the part worth reviewing closely: the extend re-scans from the **start of the last indexed line**, not from `covered_size`. When the previous build ran, the file may have ended mid-line with no newline yet — that partial line already has an index entry, and bytes appended since *continue* it rather than starting a new one. Scanning from `covered_size` would count the straddling line twice. There's a direct test (`..._handles_a_line_completed_after_the_previous_build`): build at `{"a":1}\n{"b":2}\n{"par`, then append `tial":3}\n{"d":4}\n`, assert 4 rather than 5.

Both entry points now share one scanner (`build_output_idx_from`), so the blank-line and CRLF rules can't drift between the full and incremental paths.

Six tests over `global_zone_line_count`: the three pre-existing ones unchanged, plus exact-after-append, the partial-line boundary, and shrink-triggers-full-rebuild.

**reagentx P2** (stale doc comment claiming parity with `blockfile.rs`'s exact-match check) is resolved by the same change — the exact-match check is back, so the parity claim is true again.

Verified: `cargo test -p agentmux-srv` — 2838 passed, 0 failed.

Worth noting the perf result survives: the pathological case was a 30-second poll driving a full O(file-size) rescan because "stale" is the steady state for a live agent. That's now a scan of the appended bytes (~13 KB per 30 s in the measured trace) instead of 759 MB — without trading away correctness to get it.

PR #2839, opened 2026-08-30
Camper@claudius: fix(agent-pane): anchor the model selector and Shell to the composer strip's bottom row

GenericAgentX-asaf:
## Summary

User-directed constraint (Rev 8): two composer-strip slots must **never travel** as the pane resizes — the **model selector** (`runtime`) and the **Shell toggle** (`hostShell`). Both anchor to the **bottom**, nearest the composer input: model selector on its left, Shell on its right.

Confirmed the problem live on the release build (CDP, v0.55.27) **before** changing anything — at 722px the strip really rendered `left: [ctx, auth]` / `right: [model selector, HOST+Shell]`, i.e. the selector had travelled to the right edge. This deliberately reverses the *"the model selector moving sides is acceptable"* call recorded in `docs/retro/retro-composer-strip-one-sided-lines-misdiagnosis-2026-08-26.md` step 3.

## Design

- **Multi-row:** both anchors are reserved **out** of the widest-with-narrowest pairing pool and emitted as the final row. Reserving exactly **two** slots preserves the remainder's parity, so spec §1's *"at most one singleton row"* invariant is untouched.
- **Single row:** no "bottom" exists, so the constraint degrades to its positional meaning — selector outermost-**left**, Shell outermost-**right**.
- **Anchors that can't share a line** fall to the same physical-capacity exception (spec §1) as any other pair: two adjacent one-sided rows, still bottom-most.
- `computeBalancedLeftKeys` gains an optional `fixedLeftWidth` (defaults to `0`, so every existing caller/test is byte-identical).
- **No anchor present** (controls hidden) falls back to the exact pre-anchor behavior, asserted by a test comparing both call shapes.
- **Anchors' row is rendered outside the `<For>`** (Codex P2) — see below.

## Live verification ✅

Built a dev instance off this branch, brought up a real agent pane, and drove it over CDP.

**Width sweep — wide → narrow → wide round-trip** (a one-way sweep structurally cannot catch the "stuck multi-row" class of bug that PR #2814 hit):

| width | rows | layout |
|---|---|---|
| 1071 / 700 / 520 / 420 | 1 | `selector ··· auth, hostShell` |
| 320 | 2 | `auth ··· –` / **`selector ··· hostShell`** |
| 240 | 3 | `auth ··· –` / `selector ··· –` / `– ··· hostShell` (capacity exception firing for real) |
| 1071 (return) | 1 | identical to start — no stuck multi-row |

Selector is the first element of the last row's left at every width; `hostShell` the last of its right; neither ever appears in a non-last row.

**Component identity (Codex P2), A/B on a real pane** — open the model menu, resize across the 1→2 row boundary:

| build | menu stayed open | trigger DOM node survived |
|---|---|---|
| `origin/main` | ✅ | ✅ |
| anchor constraint alone | ❌ | ❌ |
| + fix (`3bca64e9a`) | ✅ | ✅ |

The constraint forced `runtime` into a brand-new `<For>` subtree on every 1→2 crossing, destroying `AgentRuntimeDropup` and its local open/selection state. Fixed by rendering the anchors' row from its own fixed element. Layout after the restructure is byte-identical to before it (sweep re-run).

## Test plan

- [x] 8 new pure-function tests, incl. one pinning the actual regression (my first attempt at it was a false positive — the widths I picked didn't exercise it, so I re-picked) and a wide→narrow→wide round-trip.
- [x] 49/49 in `AgentComposerStrip.test.tsx` (41 pre-existing unmodified + 8 new).
- [x] 1522/1522 in `frontend/app/view/agent/`. (One `tool-renderers/registry.test.ts` failure on the first run passes in isolation both with and without this change and on a clean re-run — parallel-load flake.)
- [x] Live multi-width + round-trip + component-identity verification (above).

## ⚠️ Cleanup note for the repo owner

Live verification needed a real agent pane, so I created a throwaway agent named **`ZZ-Camper-Layout-Probe`** (named to sort last and be obviously disposable). Agent definitions are **global**, so it will show in your agent list. I closed its pane and stopped the dev instance, but couldn't find a delete path from the picker UI — **please delete it when convenient.** No other state was touched; agent count was otherwise unchanged.

### DIFF
--- a/.changesets/1788056148-fix-agent-pane-anchor-the-model-selector-and-shell-to-the-co-yn23.md
+++ b/.changesets/1788056148-fix-agent-pane-anchor-the-model-selector-and-shell-to-the-co-yn23.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): anchor the model selector and Shell to the composer strip's bottom row
--- a/frontend/app/view/agent/components/AgentComposerStrip.test.tsx
+++ b/frontend/app/view/agent/components/AgentComposerStrip.test.tsx
@@ -512,10 +512,15 @@ describe("computeComposerRows — anchored model selector + Shell (Rev 8)", () =
         expect(oneSided.length).toBeLessThanOrEqual(1);
     });
 
-    it("anchors that cannot share a line degrade to two ADJACENT one-sided rows, still bottom-most", () => {
+    it("keeps the anchors on ONE row even when they cannot fit side by side (reagent P1, PR #2839)", () => {
         // runtime(120) + hostShell(120) + gap(5) = 245 > availableWidth
-        // 200 — the physical-capacity exception (spec §1). They must not
-        // be forced onto one overflowing row, but must still not travel.
+        // 200. An earlier revision split these into two adjacent one-sided
+        // rows via the generic physical-capacity exception — but only the
+        // LAST row is hoisted out of the render's <For>, so that put
+        // `runtime` back inside it and destroyed AgentRuntimeDropup on any
+        // resize across the split boundary. The row's own flex-wrap
+        // renders the same two visual lines without moving either anchor
+        // between DOM subtrees, so the row must stay singular here.
         const slots = [
             { key: "runtime", width: 120 },
             { key: "w1", width: 100 },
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Confirmed and fixed in `3bca64e9a` — and you were right that it's **not** pre-existing: this PR introduced it.

I tested it empirically on a real agent pane over CDP (open the model menu, resize across the 1→2 row boundary, check both the menu and a tagged DOM node):

| build | menu stayed open | trigger DOM node survived |
|---|---|---|
| `origin/main` (pre-change) | ✅ True | ✅ True |
| anchor constraint alone | ❌ **False** | ❌ **False** |
| + this fix | ✅ True | ✅ True |

Root cause is exactly as you described: `<For each={rowIndices()}>` only preserves a row subtree while the row *count* is unchanged, so crossing 1→2 rows makes index 1 a brand-new subtree, and the anchor constraint forces `runtime` into it every time.

**Fix:** render the anchors' row from its own fixed element instead of as the `<For>`'s last entry, so it's one stable subtree at every row count — the DOM expression of the same "these two never travel" requirement the constraint states at the layout level. The inline stats zone moves with it (`statsInline()` is only ever true at one row, and at one row that *is* this row).

Also re-ran the full width sweep after the restructure (1071/700/520/420/320/240 → back to 1071): byte-identical row contents to before it, anchor constraint holding at every width, clean round-trip.

---

Separately — the "live verification outstanding" caveat in the PR description is now **resolved**. I got a real agent pane up in a dev build and verified the constraint end-to-end; details in the updated description below.

GenericAgentX-asaf:

New commit `3bca64e9a` since the earlier approval — the Codex P2 fix (anchors' row hoisted out of the `<For>` for component-identity stability). Layout output is byte-identical to the approved commit; the change is confined to the rendering layer.

GenericAgentX-asaf:
Sharp catch — you're right, and it was a real hole. Fixed in `27a7b5502`.

You correctly identified that my CDP pass only covered the 1→2 row crossing, so the split boundary at ~240px went unverified. That "verified one tier, not the other" gap is precisely what `retro-composer-strip-one-sided-lines-misdiagnosis-2026-08-26.md` is about, and I repeated it.

**Fix:** the anchor pair is now **never split**. It's exempt from the generic physical-capacity exception because it's degenerate — exactly two items, one per side, nothing else competing — so the row's own long-standing `flex-wrap` (reagent P1, PR #2393) breaks them onto two *visual* lines at narrow widths, rendering identically to the split while keeping both anchors in one stable DOM subtree forever. The generic exception still governs every other pair, where the wrap outcome genuinely is unpredictable.

**Live re-verified** over CDP on a real agent pane, with the model menu held **open** across the whole sweep — this time covering every boundary in both directions:

| width | DOM rows | anchors together | selector strayed | trigger node alive | menu open |
|---|---|---|---|---|---|
| 420 | 1 | ✅ | ❌ | ✅ | ✅ |
| 320 | 2 | ✅ | ❌ | ✅ | ✅ |
| 240 | 2 | ✅ | ❌ | ✅ | ✅ |
| 180 | 2 | ✅ | ❌ | ✅ | ✅ |
| 240 → 320 → 420 → natural | 2→1 | ✅ | ❌ | ✅ | ✅ |

At 240/180 the strip is now 2 DOM rows (anchor row wrapping) where it was 3 split rows — same rendering, one fewer identity boundary.

Also added a sweep-style unit test asserting the anchors share the final row at every `availableWidth` from 1000 down to 50 — the property that actually makes the identity guarantee hold, rather than just the single 200px case the old test pinned.

50/50 tests, `tsc` clean.

GenericAgentX-asaf:

Your P1 is fixed in `27a7b5502` (anchor row never split + sweep-style unit test). Details in the comment above.

PR #2709, opened 2026-08-21
feat(mcp): add CaptureWindow tool for cross-instance/OS window screenshots

AgentY-asaf:
## Summary

- `UIScreenshot` is (deliberately) scoped to only the caller's own pane — a real credential/content-isolation boundary within one AgentMux instance (see `docs/analysis/ANALYSIS_AGENT_UI_AUTOMATION_CROSS_PANE_AND_CROSS_INSTANCE_TARGETING_2026_08_21.md` §2 for the full history, including a real P0 that boundary was closing).
- No agent-facing tool exists for screenshotting a genuinely *different* OS window — a separate AgentMux instance, or any other application. That crosses no agent-to-agent trust boundary at all (no shared srv/auth-key domain), so it's a much lower-risk, separately-scoped gap (see the same analysis doc, §1).
- New `CaptureWindow` MCP tool fills that gap: matches a top-level window by partial title, captures via `xcap` (cross-platform), independent of `ui_handlers.rs`'s signed-identity scheme entirely.

## Verification

Built + tested on Windows: `cargo build -p agentmux-mcp` / `cargo test -p agentmux-mcp` clean, 4/4 tests pass. Drove the compiled binary directly over JSON-RPC (no env vars set — confirms it needs none) and captured a real, separate AgentMux instance's window; verified the resulting PNG visually.

**macOS/Linux unverified from this machine** — tracked in #2708.

## Test plan
- [x] `cargo build -p agentmux-mcp`
- [x] `cargo test -p agentmux-mcp`
- [x] Manual JSON-RPC `tools/call` against the built binary, verified output PNG
- [ ] macOS build/run (tracked separately, #2708)
- [ ] Linux build/run (tracked separately, #2708)

PR #2845, opened 2026-08-30
feat(mcp): capture-tier model — agents can screenshot any window except another OS user's

AgentA-asaf:
Phase 1 of `SPEC_AGENT_UNRESTRICTED_CAPTURE_WITH_ACCOUNTABILITY_2026_08_30.md` (added in this PR), per repo-owner direction that **agents need to be able to screenshot anything, including other agents.**

## What changes

`CaptureWindow`'s binary `!is_self` exclusion becomes a tier model keyed on *what* is being captured:

| Tier | Target | Phase 1 |
|---|---|---|
| **T1** | another pane, same instance | **allow** ← was blocked; this is the change |
| **T2** | another AgentMux instance, same OS user | allow |
| **T3** | a **different OS user's** window | **withheld** |
| **T4** | non-AgentMux window, same OS user | allow |

Only T3 stays closed, and only because it's a *human* boundary rather than an agent one — that person never consented to this app's agents, and no in-app notification can reach them. Overridable later by a host setting, never by an agent.

## Why the old rule goes

It came from `SPEC_AGENT_UI_AUTOMATION_CLICK_SCREENSHOT_2026_08_18.md` §6, whose own text says cross-agent targeting *"if it's ever wanted at all"* should be a separate capability — i.e. it **defaulted closed because no mechanism existed to be selective**, and it was never ratified. The one repo-owner confirmation recorded in that area is a *loosening* (the shared-chrome exception for `UIClick`/`UIQuery`), not a ratification of the principle. `CaptureWindow`'s exclusion was purely derivative: reagent P0 on #2709 round 3 added it so `CaptureWindow` couldn't end-run the `UIScreenshot` boundary.

**The two P0s from #2662 don't argue against this.** Both were *bypasses of the chosen policy* — client-supplied `block_id`, and a DOM-ownership check missing descendants. They show the boundary was implemented sloppily at first, not that it should exist. Under this policy neither would have been a vulnerability.

## What is deliberately unchanged

**Identity proof.** `sign_ui_automation_auth`'s HMAC + server-derived resolution is untouched. This changes what a *proven* identity may reach, not how it's proven. Any follow-up that starts trusting a client-supplied target id has misread the spec.

**`UIClick`/`UIQuery`.** Reading and acting are different risks — §6's strongest concrete objection was clicking "Confirm" on another agent's destructive dialog, which no screenshot can do. §6 is marked superseded **for capture only**, with a reviewer note so it isn't reinstated against capture code.

## Fail-closed

Matching `own_instance_pids()`'s existing discipline: the OS-user check runs **first** and outranks every other signal, and an unresolvable owner on *either* side resolves to T3 (deny). Guessing "probably mine" is precisely how a cross-user capture would slip through. There's a test asserting no window with an unresolvable owning process is ever allowed.

## Audit

The existing NDJSON trail (shared by `CaptureWindow`/`DiscoverWindows`, best-effort so it can never break the tool) gains:

- `tier`
- `target` — the **resolved** window, not the caller's query string; a `title_contains` substring doesn't tell a reviewer what was actually captured
- `image_sha256` — traces a leaked screenshot back to the call that produced it
- `redacted: false` — explicit, so Phase 3 entries are distinguishable rather than the field's absence being ambiguous between "no" and "old entry"

## DiscoverWindows

Now enumerates non-AgentMux windows (they're T4 targets) but keeps them out of the default listing behind `include_foreign` — ordinary discovery shouldn't disclose the titles of a user's unrelated apps as a side effect of looking for AgentMux windows. Entries report `tier` and `capturable` so a caller sees *why* a window is out of reach before trying.

## Spec corrections made while implementing

Flagged rather than silently fixed: an earlier draft of §6 said `UIScreenshot` must be audited as "the most privileged capture path." Wrong twice — it's own-pane-only, the *least* privileged, and it's T0, which §3 deliberately leaves unaudited to keep the entries that matter from being buried. It also contradicted §3's own table. Dropped from Phase 1, with the correction recorded in the spec.

## Not in this PR

Phases 2–4 — notification (the non-silence property), `data-amx-redact` masking, rate limiting and kill switch. **Phase 1 makes capture permitted and accountable; Phases 2–4 are what make it *safe*.** Worth not stopping here.

Also honest about a limit stated in the spec: redaction will only ever be possible at T1, where AgentMux controls the renderer. T2/T4 are OS pixel grabs of processes we don't control — a plaintext secret on screen will be captured, and audit is the entire control. That's an accepted residual risk of the directive, not something this design quietly solves.

## Verification

`cargo test -p agentmux-mcp` — **20 passed, 0 failed**, including 3 new: allow defaults, stable audit tier labels (they land in logs, so a silent rename breaks analysis), and the fail-closed check.


a5af:
@codex review

AgentA-asaf:
All three fixed in `7c9186ba8`. Both reviewers landed on the same P1/P2 independently, and it was a real leak I introduced.

**reagentx P1 / codex P2 — foreign title disclosure.** Correct, and worse than a near-miss: I added `include_foreign` to `DiscoverWindows` *specifically* to avoid disclosing the titles of a user's unrelated apps, then left the `title_contains` miss branch building its list from a set that now includes them. A caller could submit a nonsense title and enumerate the desktop — bypassing the opt-in I'd just built, and reopening the exact KeePass-title disclosure reagent P2 closed on #2709 round 2.

Fixed at both sites, not just the one reported:
- the miss list filters to `is_agentmux`
- a new `candidate_label()` withholds a foreign window's title from **every** candidate list (the pid-ambiguous branch had the same exposure), showing only its pid — enough to disambiguate, and what the caller needs anyway

Verified functionally rather than only by unit test — a deliberate miss now returns one AgentMux title and nothing else.

**codex P2 (second, independent) — resolved metadata lost on failure.** Also right, and it undermined the feature's own justification. Any failure after target resolution logged `tier`/`target` as null: a denied T3 pid, a failed `capture_image`, a failed PNG write. So the trail couldn't distinguish "no such window" from "blocked cross-user attempt" — which is the single entry a reviewer most wants to find, and the tool's description promises to record the tier of every call.

`capture_window_impl` now fills an out-param the instant a target resolves, and the audit falls back to it when the outcome is `Err`. A T3 pid additionally gets a specific error naming the tier ("`window pid=N is T3-other-user — capture is withheld…`") instead of the generic "no window found", so the caller learns it was *withheld* rather than *absent*.

**reagentx P2 — stale doc comment.** The comment on `enumerate_agentmux_windows` still asserted `is_self` windows are "hard-excluded … the actual isolation boundary, not a nicety". Rewritten to record that this was **superseded, not merely relaxed**, with the reasoning — the boundary it protected was an unratified recommendation that defaulted closed because no mechanism existed to be selective. Written that way so the next reader doesn't reinstate it as a regression.

Two new tests: the T3-denial audit entry, and that no foreign title reaches a candidate label.

`cargo test -p agentmux-mcp` — 22 passed, 0 failed.

AgentA-asaf:
Both fixed in `7c314ad80` — and the P2 is a correction of a false claim I made last round, which I want to state plainly rather than bury.

**P2 — I reported a fix that had not landed.** My previous comment said the T3-denial audit gap was fixed. It wasn't. I applied that edit with a scripted string replace that **silently no-ops when its anchor doesn't match**, and I reported it as done without re-reading the file. Two of the three edits in that batch did apply — which is why it compiled and 22 tests passed — but the third, the only one addressing the case codex actually named, did not. The build passing gave me false confidence in a change that was never there.

Now really fixed, using exact-match edits that fail loudly instead of silently. Both branches: a pid *or* a title resolving only to a withheld window is looked up in the **unfiltered** set (`foreign` is tier-filtered, so it structurally cannot see T3), populates `resolved`, and audits as `T3-other-user` with its target rather than `tier: null`. The caller also gets a specific error naming the tier instead of the generic "no window found", so *withheld* and *absent* are distinguishable to them too.

One deliberate asymmetry: the **title** branch's error does not echo the matched window's title. Naming it would confirm a cross-user window's title to the caller — precisely the disclosure this tier exists to prevent. The pid branch has no such problem, since the caller already supplied the pid.

**P1 — correct, and it was new exposure this PR created.** `DiscoverWindows`' filters never consulted `tier`, so `include_foreign: true` still disclosed a T3 window's `title` and `exe_path` — and `exe_path` embeds the owning OS username. Withholding the *capture* while still listing the window defeats the point: the disclosure across that human boundary is the whole reason the tier is withheld. Before this PR foreign windows weren't enumerated at all, so the filter belongs at the listing too, not only at the capture gate. T3 is now omitted from `DiscoverWindows` entirely, with no opt-in.

**Test added that pins the property, not the diff:** enumeration must retain withheld windows (classified, never silently dropped) so the audit can name them. A test of that shape would have failed loudly last round instead of letting my claim pass unchecked — which is the actual lesson here, not the individual bug.

`cargo test -p agentmux-mcp` — 23 passed, 0 failed. Verified against the file, not against my intent.

AgentA-asaf:
Correct, subtle, and **introduced by my own previous commit** — fixed in `86a29faf3`.

Withholding the literal title wasn't sufficient, because the *responses* differed: `"a window matching X exists but is T3-other-user"` when a withheld window's title contained the needle, versus a generic message otherwise. That's an existence oracle — probe substrings, binary-search over characters, and reconstruct another OS user's window title without ever being granted a capture. Exactly the boundary this tier exists to hold.

The causal chain is worth naming, because it's the second time in this PR the same shape has bitten: I created this **while fixing the audit gap** in the commit before it. Making withheld-vs-absent legible to the *auditor* made it legible to the *attacker* in the same edit. Accountability and confidentiality wanted the same distinction exposed to different audiences, and I exposed it to both.

**The split that resolves it:** the audit is server-side and may know; the caller may not. The T3 title match still populates `resolved` so the trail records it, then falls through to the **identical** generic-miss message — the same code path emitting the same bytes either way, rather than two branches that happen to agree today and could drift under a later edit. Structural rather than coincidental.

**The pid branch stays asymmetric deliberately**, and now documents why rather than leaving it to be "fixed" later: a caller must already hold the pid to ask, `DiscoverWindows` never hands out T3 pids, and process ownership is enumerable by anything with shell access. A pid's existence isn't a secret this tool keeps; a window title is precisely the content it does.

Test pins the observable property, not the diff: no tier label may appear in a title-miss error. A reintroduced distinguishing branch would almost certainly name the tier — that's what the reverted version did — and fail there.

`cargo test -p agentmux-mcp` — 24 passed, 0 failed.

---

**On the CI failure reported against `7c9186ba8`:** it's `agentmux-bashwrap`'s `a1_e2e_static_label_then_delayed_cr_overwrite_collapses_to_one_line` ("command should exit cleanly", 93 passed / 1 failed). This PR touches only `agentmux-mcp`, `docs`, `.changesets` and `Cargo.lock` — no bashwrap code at all. The same test passed on this branch's earlier commit `5b5a6aea4` and on every other recent run across `main` and other branches. Timing-sensitive e2e flake, not a regression from this change. Flagging rather than silently re-running.

AgentA-asaf:
Fixed in `647f23484` — correct, and I introduced the gap in the same commit that added the flag.

`include_foreign` exists *for* a disclosure reason, and this audit trail exists *for* the same disclosure reason — and I updated one of them. A reviewer scanning query strings couldn't tell whether foreign-window disclosure had even been requested for a given call, only infer it by picking through each entry's `is_agentmux` tag.

Both flags are now in the query string. The existing test asserts the exact format rather than just presence, so a future flag added to this tool without a matching audit field breaks there instead of shipping silently.

`cargo test -p agentmux-mcp` — 24 passed, 0 failed.

AgentA-asaf:
Both fixed in `b7bf6524c`. The first is straightforward; the second needs a correction to its premise, because following it literally would have reopened something.

**(a) Stale doc block — correct, and my miss.** The block above `CAPTURE_WINDOW_TOOL` still asserted three invariants this PR falsifies: capture "crosses no agent-to-agent trust boundary within one instance", "the caller's OWN instance is always excluded from the candidate set", and "scoped to AgentMux's own windows only". Same stale-invariant class I fixed on `enumerate_agentmux_windows` and missed one file-position away.

Rewritten to name each superseded claim and what replaced it, and to preserve the two round-2 constraints that **do** still hold: `is_agentmux` no longer gates *inclusion* but still gates *disclosure*, which is what keeps the KeePass-title leak closed now that foreign windows are enumerated at all.

**(b) Audit detail — right about the defect, wrong about one premise.**

The defect is real: reusing `candidate_label` for `resolved` lost a T4 window's title on a post-resolution failure, so foreign-app capture failures carried less audit detail than AgentMux ones for no reason.

But the stated premise — *"the audit is server-side only"* — doesn't hold. The trail is an NDJSON file in this instance's own data dir, and agents have shell access. I read it with `tail` while verifying this PR earlier today. It is not a confidential sink.

So implementing the suggestion literally (always record the full title) would have **reopened the T3 existence oracle closed two commits ago** — through the log file instead of the error message. Same leak, different channel.

Fixed with the rule that actually follows: **audit detail tracks the ALLOW decision, not AgentMux-ness.**

| tier | audit records | why |
|---|---|---|
| allowed (T1/T2/T4) | full title | the agent could capture the window and read the title off the pixels anyway — recording it discloses nothing new |
| withheld (T3) | pid + tier only | recording the title hands back exactly what the tier denied, in a file the agent can read |

`candidate_label` keeps its own stricter caller-facing rule. The two audiences genuinely differ — which is the same split the oracle fix turned on, now applied consistently to both channels rather than just the one that was reported.

Test asserts both directions against the real enumeration. `cargo test -p agentmux-mcp` — 25 passed, 0 failed.

AgentA-asaf:
Both fixed in `12fb79888`.

**(a) Merged doc blocks — a botched insertion, entirely mine.** I added `audit_target_label` directly above `candidate_label`, which left `candidate_label`'s original doc block sitting above the *new* function: `audit_target_label` ended up with two doc intros run together, `candidate_label` with none.

Each function now documents itself, and both say explicitly that they differ **because their audiences differ** — audit trail vs caller. That's the distinction the last three commits turned on, and it's the thing most likely to be "simplified" away later by someone reading them as near-duplicates. Worth the words.

**(b) `title_contains` schema — correct.** It still told callers it matches "other AgentMux windows' titles" while `capture_window_impl` now matches the whole tier-allowed set, T4 included. So the field description contradicted the tool's own top-level description.

Corrected, and it now also states the non-obvious consequence: an ambiguous or failed match identifies a non-AgentMux window **by pid only, never by title**. A caller who doesn't know that would reasonably read an all-pids candidate list as a bug rather than as the disclosure rule working.

`cargo test -p agentmux-mcp` — 25 passed, 0 failed.

AgentA-asaf:
Fixed in `496bff153` — the CI failure plus both notes, which turned out to share a root cause.

## The CI failure was mine, and it was a *test* bug

`audit_target_label_withholds_only_for_withheld_tiers` asserted `label.contains(&w.title)` — comparing an **unescaped** title against a label built with `{:?}`, which doubles backslashes. Green locally because no window title on this machine contains one; red on the CI runner, which has a window titled `C:\ProgramData\GitHub\HostedComputeAgent\...`.

Demonstrated directly rather than assumed:

```
raw contains    : false
escaped contains: true
```

Worse, the **sibling** `candidate_label` test had the same flaw in the negative direction — `assert!(!label.contains(&w.title))` was passing **vacuously** for any Windows path, i.e. asserting nothing precisely where the leak would matter. Both now compare the escaped form.

## Both P2s were one bad choice

I made T3 **invisible** rather than **visible-but-withheld**, and the two findings are what that broke:

- the `is_self` fail-safe promises an unresolvable-pid window "still surfaces … instead of hiding it" — but an unresolvable *owner* resolves to T3, and the listing dropped non-allowed tiers outright
- `DiscoverWindows` advertises `capturable` so a caller can "see why a window is out of reach before trying" — unreachable when every listed entry is capturable by construction

Withheld windows are now listed with `title` and `exe_path` **null** plus a `withheld_reason`. That keeps the disclosure closed — those two fields are exactly what must not cross the human boundary, and `exe_path` embeds the OS username — while restoring both promises. A pid and "someone else owns this" are already available to anything with shell access, so surfacing them costs nothing.

**A third thing this fixes that nobody reported:** if `current_user_id()` ever fails wholesale, every window resolves to T3 — and the previous code returned a **silently empty list**. It now returns the windows marked not-capturable with a reason, which is at least diagnosable.

Verified functionally against the running instance, not only by test: the listing returns capturable entries with full detail alongside withheld ones with `exe_path: null`. The mapping is extracted to `window_listing_entry` so it's unit-testable at all — it previously lived inline in the handler, which is why none of this had coverage.

`cargo test -p agentmux-mcp` — 26 passed, 0 failed.

AgentA-asaf:
Fixed in `6b2738d34`.

The comment above the filter chain still described a `.filter(|w| w.tier.allowed())` that the previous commit removed, and still claimed "T3 windows are omitted entirely, with no opt-in" when they're now listed and redacted. It contradicted the code directly beneath it *and* the accurate rationale already sitting on `window_listing_entry`.

Replaced with a short note on what those filters actually decide (which windows get listed) plus a pointer to `window_listing_entry` for what a listed entry may say about itself — rather than restating the rationale in two places where it can drift apart again. Duplicating it is what created this.

**This is the third stale-comment finding on this PR, and the pattern is mine:** I move logic into a new function and leave its justification attached to the old site. So I swept the whole file for the same class rather than just patching the reported line. The only remaining matches are inside the deliberate historical-note blocks — the ones that quote superseded claims in order to explain what replaced them, which is their purpose.

`cargo test -p agentmux-mcp` — 26 passed, 0 failed.

---

On the CI failures reported since: both were against pre-fix commits (`b7bf6524c` and `12fb79888`) — the escaped-title test bug fixed in `496bff153`. The fix commit's own run is green on vitest with the Rust jobs still going.

AgentA-asaf:
Fixed in `4b301e246`.

Correct: I asserted unconditionally that "A window owned by a different OS user IS listed but withheld", but the `include_foreign || w.is_agentmux` filter runs **before** the redaction logic. So under the tool's own documented default, another user's *non-AgentMux* window is omitted entirely — the claim only holds for a T3 AgentMux window, or with `include_foreign: true`.

Reworded to separate the two questions a caller actually has, which the original conflated:

- **is it capturable** — never, for T3
- **is it listed at all** — the `include_foreign` gate, same as any other window

Both stated, so the description no longer promises visibility the default doesn't provide.

I introduced this in the same commit that changed withheld windows from omitted to listed-and-redacted: I documented the new redaction behaviour and didn't re-check it against the filter that runs ahead of it. Same shape as the stale-comment findings — a local change described locally, without walking the path that leads into it.

`cargo test -p agentmux-mcp` — 26 passed, 0 failed.

PR #2846, opened 2026-08-30
docs(analysis): pane-minimize distortions for a Row branch between two panes

AgentX-asaf:
## Summary

Analysis only — **no code changed**. Investigates the reported distortion when minimizing the middle panes of:

```
root(Column) → [ TOP, MIDDLE(Row) → [A, B, C], BOTTOM ]
```

**Four defects, all measured against the real exported geometry functions** (`computeMainAxisAllocation`, `resolveRowSlipTargets`, `minimizedCrossAxisPx`) in a scratch vitest file — not inferred from reading the code.

## A — 72px dead space (root cause)

`minimizedFixedPx` computes a fully-minimized subtree's extent as `countLeafPanes × (header+gap)` — "one header height per leaf, stacked." **That holds only when the subtree's own `flexDirection` matches the axis being measured.** A fully-minimized **Row** branch lays its chips out *side by side*, so its true vertical extent is one header height, not N.

| | |
|---|---|
| Root Column allocates to MIDDLE | **108px** |
| MIDDLE actually fills | **36px** |
| **Dead space** | **72px** |
| Same subtree as a *Column* branch | 108 / 108 — exact |

The function's own doc comment asserts the axis-independence that's wrong here: *"a stack's required extent doesn't depend on which axis it's being measured for."* A stack's doesn't — but a Row branch is a **strip**, not a stack.

Correct formula recurses on direction (Σ along the matching axis, `max` across the perpendicular one); `countLeafPanes` is the Σ-always special case.

## B — minimizing a *middle* pane deletes the divider between its two expanded neighbours

Phase C skips a handle when *either* flanking child is minimized, so a minimized B kills both `A|B` and `B|C` — leaving the row with **zero** handles. Meanwhile B contributes zero width, so A and C render flush with no divider and can't be resized against each other.

Measured: **0 handles, widths 600 / 0 / 600**, against a 2-handle baseline.

## D — enough docked chips crush their host to zero height

Phase B scales the *chips* to fit but computes the host's height independently as `rect.height − clampedSlipHeight`, which is exactly 0 when the stack fills the row. Measured: 5 chips onto one survivor in a 120px row → **host height 0**, the row's only expanded pane disappears.

## C — chip layout flips discontinuously on the last minimize

One toggle takes a chip from **1200px wide (stacked)** to **183px wide (side-by-side)**, because slip resolution returns empty once no sibling is expanded and everything falls back to the fixed-chip path. Flagged as needing a **product decision** rather than asserted as a bug — the all-minimized strip may be intended.

## Why this survived

The existing suite's only branch-level minimize regression test uses a **Column** branch (`layoutMinimize.test.ts` — `rightCol = newLayoutNode(FlexDirection.Column, …)`). There is no Row-branch case, and the bug is direction-specific.

## Explicitly not established

Recorded in §6 of the doc rather than glossed:

- **Which finding produced the visual the operator actually saw.** A is the best match for a visible empty band; B for something feeling subtly broken. I did not reproduce it visually.
- **Whether any of it reproduces in the running app** — all measurements are pure-function level. A live check is the honest next step.
- Whether C is a defect at all.
- The mirrored Column-under-Row case, which I did not test.

## Suggested order

1. **A** — direction-aware extent (root cause, self-contained, easy test to add)
2. **D** — host-height floor (small, adjacent to A)
3. **B** — rendered-slot adjacency; needs care around `parentIndex` / `onResizeMove`
4. **C** — product call first

### DIFF
--- a/docs/analysis/ANALYSIS_PANE_MINIMIZE_ROW_BRANCH_DISTORTIONS_2026_08_30.md
+++ b/docs/analysis/ANALYSIS_PANE_MINIMIZE_ROW_BRANCH_DISTORTIONS_2026_08_30.md
@@ -1,19 +1,32 @@
 # Analysis: pane-minimize distortions when a Row of panes sits between a top and a bottom pane
 
 **Date:** 2026-08-30
-**Status:** Analysis only — **no code changed**. Four distortions identified and
-reproduced against the real geometry functions; fix directions proposed but not
-implemented or tested.
+**Status:** Analysis only — **no code changed**. Two defects (§2, §3) reproduced
+end-to-end through the real `LayoutModel.updateTree()`; one seam flagged for a
+product decision (§4); one item investigated and **withdrawn as a defect** — it
+turned out to be intentional, regression-locked behaviour (§5).
 **Area:** `frontend/layout/lib/layoutGeometry.ts`, `frontend/layout/lib/layoutMinimize.ts`
 **Trigger:** operator report — "multiple panes open side-by-side with 1 pane atop
 and 1 below; if I start minimizing the middle panes, we get distorted behavior."
 
-**Confidence, stated up front:** every number below was produced by calling the
-actual exported functions (`computeMainAxisAllocation`, `resolveRowSlipTargets`,
-`minimizedCrossAxisPx`) in a scratch vitest file, not by reading the code and
-reasoning. What I have **not** done is run the app and watch it happen. So these
-are demonstrated defects in the geometry layer; which one(s) produced the visual
-the operator actually saw is not established — see §6.
+**Confidence, per finding — these are not all the same strength:**
+
+| § | Finding | How it was measured |
+|---|---|---|
+| §2 | Dead space under a minimized Row branch | **End-to-end**: built the tree, called `model.updateTree()`, read `model.additionalProps()` |
+| §3 | Resize handles vanish | **End-to-end**: same, read `resizeHandles` off the parent's props |
+| §4 | Discontinuous chip reflow | Exported pure functions (`resolveRowSlipTargets`, `computeMainAxisAllocation`) |
+| §5 | Zero-height host | **Not a defect** — existing intended behaviour, see §5 |
+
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentX-asaf:
Both fixed in `affbb1c89` — both correct, and the first one materially changes the analysis.

**Finding D withdrawn — it was intended behaviour, not a defect.** You're right, and I should have found this: `layoutModel.test.ts`'s saturated-slip case (added for a reagent P1 on **#2211**) builds exactly that scenario and asserts `expect(props[anchor.id].rect.height).toBeCloseTo(0, 0)` with the comment *"the whole container is chips when they overflow this badly."* The host-height floor I proposed would have reverted that test by design. §5 is now a withdrawal that cites the test and the PR, and reframes what's left as a product question — *is saturate-to-all-chips what we want* — with an explicit note that changing it means deliberately reversing #2211. Kept rather than deleted so the next person who spots a zero-height host finds the reasoning instead of reopening it.

**Confidence claim — correct, and worse than a wording slip.** It contradicted the document's own §7, which already admitted the Phase C predicate and Phase B arithmetic were copied rather than called. That's the exact internal-contradiction failure mode this repo's own docs-cleanup audit exists to catch, in a document I wrote days after running that audit.

Rather than relabel them as simulations, I took the second option you offered: **§2 and §3 are now genuinely re-measured end-to-end.** `layoutModel.test.ts` shows `updateTree()` is reachable in tests via `createLayoutModel()`, so I rebuilt both through it and read the results back from `model.additionalProps()`. The numbers held:

- **§2** — MIDDLE's rect: top 246, height 108. Chip strip ends at y=**282**. MIDDLE's rect ends at y=**354**. BOTTOM starts at y=**354**. → **72px dead band**, real pipeline.
- **§3** — **0** resize handles vs. a baseline of **2**; A at (0, 400), C at (400, 400), flush at x=400, B's chip docked on C.

The header now carries a per-finding confidence table instead of one blanket claim, and §7 documents the `updateTree` harness so this is reproducible rather than asserted.

PR #2847, opened 2026-08-30
fix(layout): drag ghost matches the dragged pane's aspect ratio

AgentA-asaf:
The ghost you hold while dragging a pane out of the tiled layout was a fixed `300 × 300` square, so **every pane was held as a square regardless of its actual shape** — a wide terminal and a tall sidebar looked identical.

It now takes the pane's aspect ratio, scaled to fit a 360 px box.

## Why not the pane's literal size

That was the first thing considered and it's the wrong answer: a ghost at true pane size **occludes the drop targets it's being aimed at**. Drag a half-screen pane and you're holding a half-screen image over the split indicators and tab strip you're trying to hit — harder to use, not more informative. `toPng` cost also scales with area, on a path that runs on every header hover.

The rationale is recorded in `drag-preview-size.ts` so the question doesn't get re-litigated by someone reading "match the pane" literally later.

## Sizing rules

| Rule | Why |
|---|---|
| Never scales **up** | A pane smaller than the box shows at its own size — enlarging misrepresents it and blurs the raster |
| 96 px floor on the short edge | `4000 × 50` would otherwise scale to a ~4 px sliver that reads as a line; the ratio is deliberately broken in that extreme |
| Falls back to `300 × 300` | The honest answer when the rect is unmeasurable |

## Two things beyond the resize

**The cursor grab-point.** The rasterised image is now stored **together with the size it was rendered at**, and the offset derives from that record rather than a nominal constant. This is structural, not tidiness: a separate constant is exactly how the two drift and the ghost visibly detaches from the cursor. A test pins the offset formula **unchanged at 300 × 300**, so the pre-existing grab behaviour is provably preserved.

**Cache invalidation.** Reuse now also requires the stored size to match the current measurement. The generation counters track *content* changes and know nothing about geometry — so without this, resizing a pane and then dragging it handed you the **pre-resize ghost shape**. Caught while reading the generated result back, not by a failing test; flagging because it's the kind of thing that ships looking almost-right.

## Consolidation

`DragPreviewWidth/Height` were triplicated identically across `TileLayout.{win32,darwin,linux}.tsx`. Now one shared module rather than three copies that agree today and drift later.

## What I can't verify

**360 px is a judgement call, not a measured value** — the point where a ghost stays recognisable without covering the drop targets. I can't feel the drag from here. If it reads too small or too large in use, it's a one-line change to `DRAG_PREVIEW_MAX_PX`.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run` — **3322 passed**, 3 skipped, 0 failed (226 files)
- 10 new tests on the pure sizing/offset functions

### DIFF
--- a/.changesets/1788127480-fix-layout-drag-ghost-matches-the-dragged-pane-s-aspect-rati-gvbp.md
+++ b/.changesets/1788127480-fix-layout-drag-ghost-matches-the-dragged-pane-s-aspect-rati-gvbp.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): drag ghost matches the dragged pane's aspect ratio instead of a fixed square
--- a/frontend/layout/lib/TileLayout.darwin.tsx
+++ b/frontend/layout/lib/TileLayout.darwin.tsx
@@ -25,6 +25,12 @@ import { setTileDragInFlight } from "./dragInFlight";
 import { clearCrossTabDrop } from "./crossTabDrag";
 import { dragState } from "./tilelayout-drag-state";
 import {
+    computeDragPreviewSize,
+    dragPreviewCursorOffset,
+    DRAG_PREVIEW_FALLBACK,
+    type DragPreviewSize,
+} from "./drag-preview-size";
+import {
     DisplayNodesWrapper,
     MagnifiedPaneOverlay,
     NodeBackdrops,
@@ -56,8 +62,6 @@ export interface TileLayoutProps {
     getCursorPoint?: () => { x: number; y: number };
 }
 
-const DragPreviewWidth = 300;
-const DragPreviewHeight = 300;
 
 function TileLayoutComponent(props: TileLayoutProps) {
     const layoutModel = useTileLayout(props.tabAtom, props.contents);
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

AgentA-asaf:
Both findings addressed in `8b67399e2`.

**ReAgent P1 — unconditional floor scales small panes up.** Confirmed by running the old function directly: `{50, 30}` returned `96x96`. The fix clamps the floor to the pane's own dimension rather than gating it on a `scale < 1` branch:

```ts
const floor = Math.min(DRAG_PREVIEW_MIN_PX, DRAG_PREVIEW_MAX_PX);
return {
    width: Math.min(w, Math.max(scaled.width, floor)),
    height: Math.min(h, Math.max(scaled.height, floor)),
};
```

The clamp form also fixes something the conditional form would not: a discontinuity at the cap. `360x40` and `361x40` previously returned `360x40` and `360x96` — one px of pane width flipped the ghost's height by 56px. They now agree.

The floor still does its actual job. The existing sliver test moved from `4000x50` to `4000x128` — 128 is `layoutResize.ts`'s `MinNodeSizePx`, i.e. the narrowest pane that can really exist — and still asserts `360x96`. You're right that the function must not depend on that external invariant, so the test now names it explicitly instead of relying on it silently.

**Codex P2 — obsolete rasterizations.** Real: `toPng` is async, two hovers across a resize leave two in flight, and the older resolving last wins. Added a monotonic `previewRequestSeq`; a superseded result is dropped rather than published. Applied to all three `TileLayout.*` files, as you noted.

Three new boundary tests (small-on-both-axes, floor-vs-short-edge, cap discontinuity). 13/13 pass, `tsc --noEmit` clean.

PR #2848, opened 2026-08-30
fix(layout): derive a minimized branch's extent from its own direction, not its leaf count

AgentX-asaf:
Fixes the pane-minimize distortion reported for a **cross-split** layout — a block above, a block below, several side by side in the middle. Analysis: `docs/analysis/ANALYSIS_PANE_MINIMIZE_ROW_BRANCH_DISTORTIONS_2026_08_30.md` §2 (#2846).

## The bug

A fully-minimized subtree's extent was `countLeafPanes(node) * (HeaderHeightPx + gap)` — **sum-always**. That assumes every branch on the path stacks along the axis being measured, which holds only for an all-Column subtree.

In `root(Column)[top, Row[A,B,C], bottom]`, minimizing all the middle panes makes the Row *effectively minimized*, so the parent Column allocates it `3 × (header+gap)` of **height**. But a Row lays its chips out **side by side** — it renders one header-high strip and leaves **2 header-heights of dead space** beneath it. A visible empty band between the chips and the pane below.

The mirror case **under**-allocated for the same reason: `minimizedFixedPx` returned a flat `MinimizedRowSlotWidthPx` for *any* minimized child of a Row parent, so a fully-minimized Row branch got one chip's width for N side-by-side chips.

## The fix

`collapsedExtentPx(node, axisIsVertical, gap)` recurses on each branch's **own** `flexDirection`:

| | |
|---|---|
| leaf | one chip on the measured axis |
| branch laid out **along** that axis | **sum** of children — a stack |
| branch laid out **across** it | **max** of children — a strip |

`countLeafPanes` was the sum-always special case, correct only when direction and axis happened to agree. Both `minimizedFixedPx` and `minimizedCrossAxisPx` delegate to it, so the main-axis allocation and Phase A's cross-axis clamp stay consistent by construction.

Also corrects the doc comment that *asserted* the bug: *"a stack's required extent doesn't depend on which axis it's being measured for."* True of a stack, false of a strip — which is exactly the case that broke.

## Verification

**Unit:** 174/174 in `frontend/layout`. Seven added for the cross-split shape — both axes for leaf/Column/Row, the reported `Column[top, Row, bottom]` layout asserting **zero** dead space, the Row-in-Row mirror, and two-level nesting that resolves per level rather than by leaf count.

**End-to-end:** measured through the real `LayoutModel.updateTree()` before the fix — the Row's rect ran to y=354 while its chip strip ended at y=282, with the bottom pane starting at y=354. **72px dead band.**

**Live:** reproduced in a running `task dev` instance in the reported broken state (`Column[top, Row[min,min], bottom]`), confirmed visually, then confirmed again after Vite hot-reloaded the fix — the band closes and the freed space returns to the flex children.

## Why it survived

The suite's prior branch-level minimize coverage was **all Column-inside-Column** — the one arrangement the old formula got right.

## Not included

The second defect from the analysis (§3 — minimizing a *middle* pane deletes the resize handle between its two expanded neighbours, leaving them flush with nothing to drag) is untouched here. It needs care around `parentIndex` and `onResizeMove`'s flanking lookup, so it's better as its own change.

### DIFF
--- a/.changesets/1788101900-fix-layout-cross-split-minimize-extent-x9k2.md
+++ b/.changesets/1788101900-fix-layout-cross-split-minimize-extent-x9k2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): derive a minimized branch's extent from its own direction, not its leaf count
--- a/frontend/layout/lib/layoutGeometry.ts
+++ b/frontend/layout/lib/layoutGeometry.ts
@@ -28,16 +28,58 @@ export interface MainAxisAllocation {
     pixelToSizeRatio: number;
 }
 
-/** Leaf count of a subtree — one header chip per leaf when fully minimized. */
-function countLeafPanes(node: LayoutNode): number {
-    if (!node.children?.length) return 1;
-    return node.children.reduce((s, c) => s + countLeafPanes(c), 0);
+/**
+ * Extent (in px) a FULLY-MINIMIZED subtree needs along one axis, once every
+ * leaf in it has become a header chip.
+ *
+ * This must recurse on each branch's own `flexDirection`, not just count
+ * leaves: chips only accumulate along the axis their container lays out on,
+ * and span (max) across the perpendicular one.
+ *
+ * - leaf → one chip: a header-height along the vertical axis, a fixed
+ *   compact width along the horizontal one.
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2849, opened 2026-08-30
Manoz@Area54: feat(statusbar): break out token usage panel by agent, with cost + turn count

GenericAgentX-asaf:
## Summary
- Regroups the status bar's token breakdown popover from per-provider to per-agent: each row now shows tokens, turn count, and cost ($, from `SessionStats.cost_usd` — already tracked per-pane but never surfaced in the status bar until now).
- Background/ambient usage (next-prompt-suggestion, activity summaries, subagent naming) collapses into a single collapsible "AgentMux internal" row instead of appearing as confusing peer rows next to real agents.
- Per-agent cache hit rate shown inline (same computation as the existing global rate, scoped to one agent).
- Clicking a real agent row focuses that agent's pane — extracted swarm-view's `focusBlock` into `frontend/app/util/focus-block.ts` so both call sites share it instead of duplicating the cross-window search.
- No backend changes — `cost_usd`/`blockId`/`agentName` were all already available frontend-side; this is aggregation + display only.

See `docs/specs/SPEC_STATUSBAR_TOKEN_PANEL_BY_AGENT_2026_08_30.md` for the full design, including the value-add ideas considered and why (cost, turn count, ambient split, click-to-focus) vs. deferred (active/idle grouping, burn rate).

## Test plan
- [x] `npx vitest run frontend/app/store/token-usage.test.ts` — 13/13 passing, including new by-agent aggregation/sort/reset coverage
- [x] `npx vitest run frontend/app/statusbar/TokenBreakdownPopover.test.tsx` — 5/5 passing (new), covers empty state, per-agent rows, ambient collapse/expand, focus-on-click
- [x] `npx vitest run frontend/app/view/swarm` — 93/93 passing (regression check on the `focusBlock` extraction)
- [x] Full suite: `npx vitest run` — 226 files / 3325 tests passing, 2 files/3 tests skipped (unrelated e2e, gated by `AGENTMUX_E2E_BUILD_DIR`)

### DIFF
--- a/.changesets/1788129105-feat-statusbar-break-out-token-usage-panel-by-agent-with-cos-1rcn.md
+++ b/.changesets/1788129105-feat-statusbar-break-out-token-usage-panel-by-agent-with-cos-1rcn.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(statusbar): break out token usage panel by agent, with cost + turn count
--- a/docs/specs/SPEC_STATUSBAR_TOKEN_PANEL_BY_AGENT_2026_08_30.md
+++ b/docs/specs/SPEC_STATUSBAR_TOKEN_PANEL_BY_AGENT_2026_08_30.md
@@ -1,0 +1,202 @@
+# Spec: Token stats panel — break out by agent + value-add details
+
+**Date:** 2026-08-30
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #2850, opened 2026-08-30
fix(layout): an all-minimized row's chips share the full width

AgentX-asaf:
Follow-up to #2848, straight from the operator's own check of it:

> the space above the 2 collapsed panes is gone, but there is still blank space to the right. The collapsed panes should equally fill up the empty space.

## The remaining half

#2848 fixed the row's **height**. Its **width** was still wrong: every minimized child of a Row took a flat `MinimizedRowSlotWidthPx` (180px + gap), so two chips occupied ~366px of a full-width row and left the rest empty.

That fixed width is correct when a chip sits beside an expanded pane. But this path is **only reachable when no sibling is expanded** — an expanded one would have absorbed the chips via `resolveRowSlipTargets` instead. With no flex child in the row, the chips are its only occupants, and there is nobody else the leftover space could belong to.

## The change

When a Row has fixed children and **no** flex children, the chips scale to fill `nodePixels`.

Distributed **proportionally to each child's own collapsed extent**, not strictly equally — an all-leaf row divides evenly (the reported case), while a nested fully-minimized Row branch keeps the larger share it genuinely needs for its N side-by-side chips.

**Row-only, deliberately.** A fully-minimized Column is already handed exactly its collapsed extent by its parent (#2848), so it has no leftover to fill, and stretching it would render chips *taller than a header*. There's a test pinning that.

Existing shrink-to-fit is untouched and still covered: six chips in a 300px row still scale below one chip-width and sum to the container.

## Tests

**193/193**, six added:

| | |
|---|---|
| two / three minimized leaves | split evenly, sum to the full width (was 366 of 1200) |
| nested fully-minimized Row branch | proportional — 300 / 600 of 900, not 450 / 450 |
| expanded sibling present | **no** stretch — still the fixed chip width (the slip path) |
| Column parent | **no** stretch — chips stay header-height |
| narrow row | still shrinks below one chip-width, sums to container |

Every pre-existing test pinning `MinimizedRowSlotWidthPx` has an expanded sibling, so none of them exercise this path — I checked that before changing it.

## Not verified visually

The live `task dev` panes had been restored to expanded by the time I captured, so the fill state wasn't on screen. The unit coverage is exact (600/600 in a 1200px row), but the visual check is still worth doing — the fix is hot-loaded in that instance if you re-minimize them.

### DIFF
--- a/.changesets/1788130400-fix-layout-minimized-row-fill-width-q4m8.md
+++ b/.changesets/1788130400-fix-layout-minimized-row-fill-width-q4m8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): an all-minimized row of panes fills the width instead of leaving dead space to the right
--- a/frontend/layout/lib/layoutGeometry.ts
+++ b/frontend/layout/lib/layoutGeometry.ts
@@ -140,9 +140,31 @@ export function computeMainAxisAllocation(
         isSlip(c) ? 0 : isEffectivelyMinimized(c) ? minimizedFixedPx(c, nodeIsRow, gapPx) : 0
     );
     const fixedTotal = fixed.reduce((a, b) => a + b, 0);
-    const scale = fixedTotal > nodePixels && fixedTotal > 0 ? nodePixels / fixedTotal : 1;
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2851, opened 2026-08-30
feat(agent-pane): dock a bare sleep immediately with a live countdown

Agent2-asaf:
## Summary

A `sleep 300` sat in "Working…" for 30s before the dock showed anything — promotion is duration-based and can't tell a wait from a build. Median for these in real transcripts is **61s**, so the old behaviour spent half of every wait saying the wrong thing.

`sleep-detect.ts` recognizes a command that is a wait **and nothing else**. Such a call promotes at `t=0` — it has already declared itself, and the threshold exists to *detect undeclared* work — and carries `sleepMs`, so the row renders **"~Ns left"** instead of a blind elapsed timer.

This is `REPORT_LONGRUNNING_TOOLCALL_AUTODETECT_STATUS_2026_07_26.md` §4.2's "phase 1a, cheap UX polish" — layered on duration promotion, never replacing it.

## The measurement §4.2 asked for

That section rejected text matching as the classifier and required *"real-transcript validation, not just Agent1's one heartbeat example, before it ships."* Done — 7,761 foreground Bash calls across 12 production transcripts:

| Rule | Matches | Verdict |
|---|---|---|
| Naive: "command starts with `sleep`" | 270 | **204 (76%) are not waits** — `sleep 90; tail -30 <log>`, `sleep 60; ls <dir>` |
| Narrow: whole command **is** a sleep | 66 | all genuine, median 61s, **zero false positives** |

§4.2's rejection was correct — the naive rule is wrong three times in four. What it over-shot is the case with no second clause to be wrong about. Two facts make the narrow rule free:

- All 204 compound sleeps ran **28–100s**, so duration promotion already catches every one. Ignoring them here costs nothing.
- **18 of the 66** finish under 30s — invisible in the dock today.

Validated the shipped matcher against the 299 distinct real sleep-containing commands: 15 matched (all bare `sleep N`, 5–400s), 284 rejected (all compound).

## Also measured — and rejected

Lowering `TOOL_PROMOTION_MS` to 10s. **Median foreground Bash is 7.5s**, so 10s lands on the steepest part of the distribution:

```
  5-10s : 3,372  43.4%   <-- largest bucket
 10-30s : 1,636  21.1%   <-- would be newly promoted
promote @30s :   647 ( 8.3%)
promote @10s : 2,283 (29.4%)   3.5x the dock rows
```

It converts the dock from "something unusual is happening" into a command log, and makes the same command dock or not depending on cache state. **30s stays.**

## Details

- `SLEEP_IMMEDIATE_MIN_MS = 5s` floor keeps micro-delays out — `sleep 1` between two commands would otherwise hold a row for its own duration plus the full retention window.
- A finished pure sleep **resolves in place** (done → retention) rather than vanishing, since it was docked from `t=0`.
- `nextToolPromotionAt` skips pure sleeps — no timer for a promotion that already happened.
- Countdown clamps at 0 (the process is reaped just after its deadline) and disappears once terminal.

**Fixture change worth a look:** `tool-adapter.test.ts` and `attached-task.test.ts` both used `sleep 300` as their generic long-running command, which now takes the fast path and made every threshold assertion vacuously pass. Their defaults are now a non-sleep command, with a comment saying why it must stay that way.

## Testing

- 24 matcher tests, including the eight real compound-sleep shapes it must refuse
- 7 adapter tests for immediate promotion, countdown data, retention, and the compound-sleep fallthrough
- `npx vitest run frontend/app/view/agent` — **1554 passed** (up from 1523)
- `npx tsc --noEmit` clean; new SCSS block stylelint-clean (the 8 `color-no-hex` errors in that file are pre-existing and identical on `main`)

**Not verified in a live pane** — the matcher is validated against production command text and the adapter against unit tests, but the rendered countdown is code-verified only.

Also refreshes that report's §3 status table: steps 1–3 shipped, step 4 partial (Swarm has `shellRows`/`cronRows` but not dock-promoted Bash calls), step 5 not started.

### DIFF
--- a/.changesets/1788130657-feat-agent-pane-dock-a-bare-sleep-immediately-with-a-live-co-gl5d.md
+++ b/.changesets/1788130657-feat-agent-pane-dock-a-bare-sleep-immediately-with-a-live-co-gl5d.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent-pane): dock a bare sleep immediately with a live countdown
--- a/docs/specs/REPORT_LONGRUNNING_TOOLCALL_AUTODETECT_STATUS_2026_07_26.md
+++ b/docs/specs/REPORT_LONGRUNNING_TOOLCALL_AUTODETECT_STATUS_2026_07_26.md
@@ -1,6 +1,39 @@
 # Report: auto-detecting long-running tool calls (sleep and beyond) and docking them — status refresh, 2026-07-26
 
-**Status:** Report — audit + design synthesis, all open questions resolved (§4), not yet implemented. This is **not** a from-scratch analysis: it verifies and consolidates two existing same-topic reports against `main` as of today, and updates them with what has (and hasn't) shipped in the 10 days since.
+**Status:** Report — audit + design synthesis, all open questions resolved (§4). **Largely implemented since; see the status table below (refreshed 2026-08-30).** This is **not** a from-scratch analysis: it verifies and consolidates two existing same-topic reports against `main` as of today, and updates them with what has (and hasn't) shipped in the 10 days since.
+
+### §3 implementation status (verified against `main`, 2026-08-30)
+
+| Step | State | Where |
+|---|---|---|
+| 1. Duration-threshold promotion | **shipped** | `activity/tool-adapter.ts`, `TOOL_PROMOTION_MS = 30_000` |
+| 1a. `sleep <N>` → immediate dock + "~Ns left" | **shipped 2026-08-30** | `activity/sleep-detect.ts` — see the note below |
+| 2. `run_in_background` end-to-end | **shipped** | issues #2490, #2518, #2491, #2492 |
+| 3. Two-axis pane status | **shipped** | `activity/attached-task.ts`, `SPEC_ATTACHED_TASK_STATUS_AXIS_2026_08_02.md` |
+| 4. Feed the signal to Swarm | **partial** | `shellRows`/`cronRows` exist; dock-promoted Bash calls are still absent |
+| 5. Generalize beyond Bash | **not started** | `isBashToolNode` still gates on `tool === "Bash"` |
+
+**On 1a (§4.2's "cheap UX polish"), with the measurement §4.2 called for.**
+That section asked for "real-transcript validation, not just Agent1's one
+heartbeat example, before it ships." Done — 7,761 foreground Bash calls across
+12 production transcripts:
+
+- The **naive** rule §4.2 rejected ("command starts with `sleep`") matches 270
+  calls, of which **204 (76%) are not waits** — `sleep 90; tail -30 <log>`,
+  `sleep 60; ls <dir>`. The rejection was correct.
+- Restricting to commands that are a wait **and nothing else** matches 66, every
+  one genuine, median 61s. No false-positive surface: there is no second clause
+  to be wrong about. 18 of the 66 finish under 30s, i.e. the duration rule
+  misses them entirely.
+- All 204 compound sleeps ran 28–100s, so duration already catches every one —
+  ignoring them here costs nothing.
+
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

Agent2-asaf:
Both P2s fixed in `fe0dbd5eb` — both were real bugs, and I verified each against the actual tools rather than taking the report at face value.

**`timeout N sleep M` — correct, and the more serious of the two.** The matcher returned the sleep's own argument, so `timeout 5 sleep 300` would have counted down from five minutes for a process that dies in five seconds. That's exactly the confidently-wrong reading this module exists to prevent: the countdown is only justified *because* the remaining time is known, so getting it wrong is worse than not showing one.

Now takes `min(timeout, sleep)`. Two details your comment pointed at that I handled explicitly:
- **Zero-duration**: confirmed from `timeout --help` — *"A duration of 0 disables the associated timeout."* So `timeout 0 sleep 300` correctly yields 300s, not 0.
- **The floor now applies to the effective wait**, not the sleep argument — `timeout 2 sleep 300` is a 2s wait, below `SLEEP_IMMEDIATE_MIN_MS`, so it gets no row at all.

**Case sensitivity — correct, with one nuance worth recording.** Verified `sleep 5M` → `sleep: invalid time interval '5M'`; coreutils accepts only lowercase `s`/`m`/`h`/`d`, exactly as you said. Dropped `/i` and the suffix lowercasing.

The nuance: `SLEEP 1` actually *succeeds* on this Windows box, because the filesystem is case-insensitive and resolves it to `sleep.exe`. So the uppercase-command case is platform-dependent rather than universally broken. Case-sensitive is still the right call — matching the shell's real behaviour on the platforms where it differs is the point, and a false negative just falls through to duration promotion, which costs nothing. Recording it so the next person doesn't "fix" it back after testing only on Windows.

Both would have produced the same failure shape: promoting a call that fails instantly, leaving a phantom dock row, a suppressed working row, and a bogus attached-task watchdog exemption for a command that never waited.

7 new tests (12 assertions) covering timeout precedence, zero-duration disabling, the floor on effective wait, and each rejected case form. **1561 passing**, up from 1554. `tsc` clean.

Agent2-asaf:
## Verification closed out — the full chain, without a live run

I'd flagged this as "rendered countdown is code-verified only." That gap is now closed, by a route I think is stronger than a one-off live check for this particular change. Every link:

1. **Wire shape** — `claude-translator.ts:244` assigns `params: block.input` verbatim, no transformation. So `ToolNode.params.command` *is* the wire `tool_use.input.command`, byte for byte. (Confirmed in code, not just from `BashParams`'s doc comment claiming it.)
2. **Real command text** — the matcher was validated against 299 distinct sleep-containing commands extracted from real `tool_use.input.command` in production transcripts: 15 matched (all bare `sleep N`, 5–400s), 284 rejected (all compound). Same strings the running app would hand it.
3. **Adapter** — 7 tests: immediate promotion at `t=0`, `sleepMs` propagation, retention after a short sleep finishes, compound-sleep fallthrough to the duration rule, no redundant promotion timer.
4. **Render** — 8 component tests asserting the countdown reaches the DOM: mounted value, ticking via `useTick`, clamping at zero past the deadline, absence once terminal, absence when `sleepMs` is unset (while the elapsed clock still renders), and coexistence with that clock.

Because the wire→params hop is an identity assignment, step 2 tests the matcher against exactly what production feeds it — which is what a live run would have demonstrated once, except this re-runs on every commit.

**What a live run would still add,** honestly: confidence that nothing *else* in the running app interferes between the tool call and the dock. That pipeline was already exercised end-to-end during PR #2833's live verification.

Approved by ReAgent (three passes), both Codex P2s fixed, all three CI checks green. **1569 tests passing**, up from the 1523 baseline before this work.

Merging.

PR #2853, opened 2026-08-31
fix(agent-pane): peek panel shrink-wraps, floats right, 12-hour AM/PM time

AgentX-asaf:
Three operator-requested refinements to the hover-to-peek panel (`SPEC_TRANSCRIPT_NODE_HOVER_PEEK_ALL_KINDS_2026_08_25`).

## 1. Shrink-wrap instead of full row width

The panel stretched the **entire width of the hovered row** to hold two short metadata lines — covering the content you're trying to inspect. Now sizes to its content (`width: max-content`), with the inline `max-width` still clamping it to the row so a long tool command can't escape the pane.

## 2. Float right

Pinned to the row's **right edge**, growing leftward, so it sits clear of the text you're reading on the left.

Implemented as `left: rect.right` + `transform: translateX(-100%)`, **not** a `right:` offset — `right` would need `window.innerWidth`, a different coordinate space from the `getBoundingClientRect()` values the rest of the component uses, which would break the CSS-`zoom` safety its header documents.

## 3. Not a flat black slab

Translucent + blurred (`color-mix` 88% + `backdrop-filter`), so it reads as floating *above* the transcript rather than punching a hole in it. Same treatment as `pane-size-badge.scss` — this codebase's other small floating overlay, which uses it for "legibility over busy content". Hard corners kept per `SPEC_HARD_CORNERS_2026_05_26`.

## 4. 12-hour local time with AM/PM

`formatExactTime` was **already local** — `getHours()` always was — so only the presentation changed. Hour is unpadded per the usual convention (`3:46:12 PM`, not `03:46:12 PM`); minutes and seconds still padded.

## On the shared component

Alignment is a new `align` prop defaulting to **`"end"`**, because six of the seven call sites are the metadata peek and want it. `UserMessageBlock`'s "Session context" body preview explicitly opts into `"stretch"` and keeps full width — it renders a real message body, sometimes kilobytes of startup payload, where full width is the point.

Defaulting the other way would have meant editing six call sites and leaving the odd one out silently correct.

## Verification

**1678 tests passed across 129 files** (`frontend/app/view/agent` + `frontend/util`).

Four cases added for the time format, covering the two boundaries 12-hour conversion usually gets wrong — **midnight as 12 AM, not 0 AM**, and **noon as 12 PM** — plus the unpadded hour. Eight peek tests asserted the old two-digit 24-hour pattern and were updated.

`tsc --noEmit` clean. Stylelint on the touched SCSS reports **the same 36 pre-existing errors before and after** — verified by running it against the stashed file, not assumed.

## Not verified

The panel only renders on hover, which I can't drive from here, so this is unverified visually. It's hot-loaded in the running `task dev` instance.

### DIFF
--- a/.changesets/1788132600-fix-agent-pane-peek-panel-shrinkwrap-right-ampm-w7t3.md
+++ b/.changesets/1788132600-fix-agent-pane-peek-panel-shrinkwrap-right-ampm-w7t3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): peek panel shrink-wraps, floats right, and shows 12-hour AM/PM time
--- a/frontend/app/view/agent/components/AgentMessageBlock.test.tsx
+++ b/frontend/app/view/agent/components/AgentMessageBlock.test.tsx
@@ -44,7 +44,7 @@ describe("AgentMessageBlock — peek tooltip", () => {
             hover(container);
             const metaLines = document.body.querySelectorAll(".agent-node-peek-tooltip-meta");
             expect(metaLines.length).toBe(2);
-            expect(metaLines[0].textContent).toMatch(/\d{2}:\d{2}:\d{2} · 1m ago/);
+            expect(metaLines[0].textContent).toMatch(/\d{1,2}:\d{2}:\d{2} (?:AM|PM) · 1m ago/);
             expect(metaLines[1].textContent).toMatch(/~\d+ tok \(est\.\)/);
         } finally {
             vi.useRealTimers();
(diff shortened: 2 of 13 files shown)


a5af:
@codex review

AgentX-asaf:
Fixed in `1b575538a` — correct, and the more annoying half of the two: the JSDoc header is what shows on hover in an editor, so a stale one there outlives the inline note nobody scrolls to.

Header now reads `"3:46:12 PM"` with a dated pointer to the inline note for what did and didn't change (local time was never the thing that changed — `getHours()` always was local).

Also swept `frontend/` for other stale 24-hour references while I was in there. The three remaining hits are all deliberate historical mentions in the new comments describing the change itself, not surviving claims about current behaviour.

16/16 in `format-time.test.ts`.

PR #2854, opened 2026-08-31
Manoz@Area54: fix(identity): stop isolated Claude Code config dirs leaking the host's ~/.claude/CLAUDE.md

GenericAgentX-asaf:
## Summary

Verified (at direct request) whether AgentMux agents actually avoid the operator's global Claude Code config — found a real, live leak, not just a doc claim.

- `CLAUDE_CONFIG_DIR` is correctly set to an isolated dir for every spawned Claude Code agent (default and identity-bound), and correctly relocates credential/session/project storage.
- **But AgentMux never wrote a `CLAUDE.md` into that relocated dir.** Claude Code CLI's own user-level `CLAUDE.md` discovery treats an isolated dir with none of its own as "fall through to the real `$HOME/.claude/CLAUDE.md`," not "no memory file."
- Verified live on this machine: **18+ isolated `claude` config dirs, none with a `CLAUDE.md`**, and a live agent session whose `CLAUDE_CONFIG_DIR` was a genuinely separate (non-ambient) directory nonetheless received the real host file's contents in its system prompt.
- This directly contradicts the Armory "Global Memory" pane's own claim that the host CLI config is "Not read by spawned in-app agents."
- Distinct from `SPEC_BLOCK_AMBIENT_HOME_DIR_IDENTITY_BINDING_2026_08_25.md` (already-fixed case: config dir literally *is* the ambient home) — this leak reproduces even when the dir is genuinely separate and correctly isolated, just empty.

## Fix

Seeds an empty, self-explanatory placeholder `CLAUDE.md` into a `claude`-provider isolated config dir the first time it's used — idempotent, never overwrites real content (so Global Memory or anything else already there is untouched):

- `agentmux-srv/src/backend/providers.rs` — new `seed_claude_md_placeholder_if_missing(provider, config_dir)`, scoped to `auth_dir_name == "claude"` only (unverified for other providers).
- `agentmux-srv/src/server/app_api/agent_open.rs` — wired into the default-agent spawn path, right where `CLAUDE_CONFIG_DIR` already gets set.
- `agentmux-srv/src/identity/resolver/inject.rs` — wired into the identity-bound spawn path, right after the existing ambient-home-dir block check passes.

See `docs/specs/SPEC_ISOLATE_HOST_CLAUDE_MD_2026_08_31.md` for the full writeup, including non-goals (no attempt to import/compose the host's real `CLAUDE.md` content — that would just reintroduce the leak in a different form; Global Memory stays the correct mechanism for shared instructions).

## Test plan
- [x] New unit tests in `providers.rs` (`claude_md_placeholder_tests`): writes when missing, creates the dir first if needed, never overwrites existing content, no-ops for non-`claude` providers — all passing
- [x] `cargo test -p agentmux-srv --bin agentmux-srv identity::resolver::inject` — 28/28 passing (confirms the ambient-home-dir gate is unaffected)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv server::app_api::agent_open` — 7/7 passing
- [x] Full suite: `cargo test -p agentmux-srv --bin agentmux-srv` — 2849 passed, 0 failed, 6 ignored
- [x] `cargo check -p agentmux-srv` — clean (no new warnings)

### DIFF
--- a/.changesets/1788151399-fix-identity-stop-isolated-claude-code-config-dirs-leaking-t-32yz.md
+++ b/.changesets/1788151399-fix-identity-stop-isolated-claude-code-config-dirs-leaking-t-32yz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): stop isolated Claude Code config dirs leaking the host's ~/.claude/CLAUDE.md
--- a/agentmux-srv/src/agents/failure.rs
+++ b/agentmux-srv/src/agents/failure.rs
@@ -238,6 +238,31 @@ pub fn classify(
             &tail,
         );
     }
+    // The identity spawn gate's CLAUDE.md isolation-seed refusal
+    // (identity/resolver/errors.rs's `SpawnGateError::ClaudeMdSeedFailed`
+    // Display impl — same reasoning as the two branches above: this never
+    // reaches the API, so it needs its own title/detail instead of falling
+    // through to the generic UnknownNonZero fallback. ReAgent P2 on PR
+    // #2854 — added alongside AmbientHomeDirNotAllowed's sibling branch
+    // above but initially missed for this newer variant.
+    if hay.contains("refusing to spawn with an unprotected config dir") {
+        let provider_phrase = extract_claude_md_seed_provider(&combined)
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2855, opened 2026-08-31
fix(layout): restore the resize handle between panes flanking a minimized one

AgentX-asaf:
Third and last defect from `ANALYSIS_PANE_MINIMIZE_ROW_BRANCH_DISTORTIONS_2026_08_30` (§3, #2846), completing the operator's original minimize report alongside #2848 and #2850.

## The bug

Minimizing a **middle** pane deleted the divider between its two expanded neighbours.

Phase C paired handles on adjacent child-array indices and skipped a pair if either side was minimized — so `A | B(min) | C` skipped both `A|B` and `B|C` and emitted **zero** handles. Meanwhile B, being a slip child, contributes **zero main-axis extent**, so A and C render flush with nothing to drag between them. Measured end-to-end: **0 handles against a baseline of 2**.

## The fix

Phase C now pairs consecutive **rendered** slots. A slip child is skipped over rather than breaking the chain, because it has no extent and the panes either side share one real boundary.

**An ordinary minimized child is deliberately NOT skipped.** In a Column — or in a Row where every child is minimized — the chip occupies a real slot, so its neighbours genuinely aren't adjacent and a spanning handle would float on top of the chip. That case resets the chain. This distinction is the whole reason the fix keys on *slip-ness* rather than *minimized-ness*.

## Why this needed a contract change

A spanning handle's two sides aren't adjacent in the child array, which the old contract couldn't express:

```ts
const beforeNode = parentNode.children![resizeHandle.parentIndex];
const afterNode  = parentNode.children![resizeHandle.parentIndex + 1];  // ← assumes adjacency
```

A spanning handle would have resolved the **minimized** pane as its after-node, and the locked-edge guard immediately below would then have rejected the drag — so it'd have looked fixed while still doing nothing.

`ResizeHandleProps` gains an explicit `afterIndex`; `onResizeMove` reads it, with a `??` fallback for any stale handle from a pre-upgrade frame.

## Verification

**3317 tests across 225 files, typecheck clean.** Seven added:

- six enumerate the pairing rule — single slip spanned, after-index ≠ before+1, two adjacent slips spanned by one handle, unchanged when nothing is minimized, **a Column chip NOT spanned**, an all-minimized row yielding none
- one goes **end-to-end through the real `updateTree`** and asserts the shipped path agrees: one handle, `parentIndex` 0, `afterIndex` 2, and A and C actually flush

## One thing worth recording

That end-to-end test first asserted the slip child's rect width was `0` and **failed at 400**. The code was right and my assumption wrong: Phase A allocates a slip child 0px, then **Phase B replaces that rect** with the docked chip, which takes its host's geometry. Asserting 0 would have pinned a pre-Phase-B intermediate rather than what actually renders. It now asserts the chip sits on its host, with a comment explaining why.

### DIFF
--- a/.changesets/1788132600-fix-agent-pane-peek-panel-shrinkwrap-right-ampm-w7t3.md
+++ b/.changesets/1788132600-fix-agent-pane-peek-panel-shrinkwrap-right-ampm-w7t3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): peek panel shrink-wraps, floats right, and shows 12-hour AM/PM time
--- a/.changesets/1788153000-fix-layout-handle-spans-slipped-pane-r2v9.md
+++ b/.changesets/1788153000-fix-layout-handle-spans-slipped-pane-r2v9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): restore the resize handle between two panes flanking a minimized one
--- a/frontend/app/view/agent/components/AgentMessageBlock.test.tsx
+++ b/frontend/app/view/agent/components/AgentMessageBlock.test.tsx
@@ -44,7 +44,7 @@ describe("AgentMessageBlock — peek tooltip", () => {
             hover(container);
             const metaLines = document.body.querySelectorAll(".agent-node-peek-tooltip-meta");
             expect(metaLines.length).toBe(2);
-            expect(metaLines[0].textContent).toMatch(/\d{2}:\d{2}:\d{2} · 1m ago/);
+            expect(metaLines[0].textContent).toMatch(/\d{1,2}:\d{2}:\d{2} (?:AM|PM) · 1m ago/);
             expect(metaLines[1].textContent).toMatch(/~\d+ tok \(est\.\)/);
         } finally {
             vi.useRealTimers();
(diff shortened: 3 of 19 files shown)


a5af:
@codex review

AgentX-asaf:
Fixed in `8f9dac6d1` — correct, and a real interaction bug, not just geometry tidiness.

You're right about the mechanism: in the right-preferred `A | B(min) | C` case Phase B docks B onto C and pushes **C's top down** by a chip height while A keeps full height. Deriving the span from A alone put the handle's upper segment **alongside B's locked chip** — intercepting pointer events there, and offering to resize A against C from an edge those two panes don't share. So the fix I shipped would have looked right and behaved wrong at the top of the boundary.

Perpendicular span is now the **intersection** of both flanking panes' final rects — max of the starts, min of the ends, on whichever axis is perpendicular to the parent's direction. For two ordinary siblings the rects cover that axis identically, so the intersection is exactly the old value and nothing changes.

Your suggestion also surfaced a degenerate case worth guarding: when the two rects don't overlap **at all** — a saturated slip group that consumed its whole host, which `layoutModel.test.ts` already pins as intended behaviour from #2211 — there's no shared edge to drag, so no handle is emitted. The after-pane still becomes the next pair's before-side.

**202 passing**, two added:
- the handle starts at C's top rather than A's and ends at the shorter pane's bottom — and it first asserts C really *was* pushed down, so it can't pass vacuously
- two ordinary siblings are unaffected

Third PR in a row where your catch was a behaviour I'd have shipped as "fixed."

PR #2380, opened 2026-08-01
feat(muxspect): new live process/turn-state introspection tool (Phase 1)

AgentA-asaf:
## Summary

New tool: `muxspect` — `muxlog`'s live-state sibling. `muxlog` answers "what happened" (log files on disk); `muxspect` answers "what is this instance doing right now." Grew directly out of a live debugging session where neither tool nor any existing RPC could answer "which instance is even hosting this conversation" or "what's this block's current state."

**Phase 1 scope only** (current instance, no cross-instance query yet — see the spec's §7 for Phase 2/3):

- Two new authenticated HTTP routes, `/api/v1/muxspect/list` and `/api/v1/muxspect/describe`, composing the existing `ProcessBroker::status()`/`list()` + `BlockControllerRuntimeStatus` + OS-process tree into the single "describe everything about block X" query `REPORT_PROCESS_ARCHITECTURE_STATE_AND_RETHINK_2026_07_22.md` §5.4 already named as missing.
- `muxspect.mjs`, deployed exactly like `muxlog.mjs` (same `include_str!` embed, same per-shell deploy, same `bash`/`zsh`/`pwsh`/`fish` function wiring). Connects via `$AGENTMUX_LOCAL_URL`/`$AGENTMUX_AUTH_KEY` inherited from its own environment — verified directly that this is the same path `agentmux-mcp` already uses for every other `/api/v1/*` route, not new IPC.

## Design doc

`docs/specs/SPEC_MUXSPECT_LIVE_INTROSPECTION_TOOL_2026_08_01.md` — external best-practices research (kubectl/docker/OTP observer/Chrome DevTools Protocol/Temporal CLI) + a full internal audit of what already existed, both via background research agents, informed every design decision below. Also `docs/MUXSPECT.md` (user-facing reference, mirrors `docs/MUXLOG.md`'s structure).

Notable decision worth flagging: the Agent pane's Activity Dock is **explicitly out of scope for every phase** — it's pure in-renderer SolidJS state, never persisted or RPC-exposed, so no backend tool can ever see it. AgentMux already ships CEF, which supports Chrome's remote-debugging protocol — that's the right tool for that specific gap, not something to build into `muxspect`.

## Test plan

- [x] `cargo test -p agentmux-srv` — 1860 passed, 0 failed (4 new: 2 route tests + auth-gate test + bad-request test; 1 new: `deploy_scripts` writes `muxspect.mjs` alongside `muxlog.mjs`)
[...]

PR #2856, opened 2026-08-31
feat(muxspect): `layout` — inspect the persisted pane tree, run the doctor

AgentX-asaf:
Adds `GET /api/v1/muxspect/layout[?tab_id=X]` and the `muxspect layout` command.

```
tab tab1  (4 pane(s), 2 minimized)  healthy
  column (3)  size=10
    leaf dbe65c4c  size=10
    row (2)  size=10  [all-minimized]
      leaf 44807f01  size=10  [minimized]
      leaf 04cc1750  size=10  [minimized]
    leaf dd598cf3  size=10
```

## Why

Layout state was previously only inspectable by copying `db_layout` out of SQLite and hand-walking the JSON — which is what diagnosing the cross-split minimize bugs (#2848, #2850, #2855) actually took, twice in one session. This applies muxspect's existing *"what is this instance doing right now"* contract to layout.

## Scope note — stated plainly, because it changed mid-flight

The original pitch for this was **detecting the cross-split extent mismatch**. Those fixes landed first, so that detection would now find nothing, and this is deliberately **not** built around it.

What it's actually for:
1. **Inspectability** — the outline above, instead of hand-rolled SQLite walking.
2. **Running the doctor against the tree ON DISK.** `validate_layout_invariants` normally only runs at reducer write-time, so on-demand checking catches corruption *already persisted* — including by older builds that never ran it. That's the case nothing else covers.

## Two things the implementation had to get right rather than assume

**`minimized` isn't a typed field** on the Rust `LayoutNode` — it round-trips through the untyped `extra` catch-all. Reported as `locked` via the reducer's own `is_node_locked`, which also recognises the legacy pre-display-mode markers (`minimizedSize`/`slipMinimize`/`columnDissolve`). An unmigrated tree on disk is exactly what this route should surface; reading the raw flag would have missed it.

**`effectively_minimized` is surfaced per node** because a *branch* never carries a minimized marker of its own — it's derived from every leaf beneath it, and that derived state is what drives chip geometry. It's the distinction all three minimize bugs turned on, and it's invisible from the stored flags.

## Error handling

Store errors return **200 with an `error` field**, not a 4xx: `apiGet()` in `muxspect.mjs` treats non-2xx as transport failure and would report a store-read problem as *"srv is unreachable."* A tab pointing at a missing `layoutstate` is likewise reported as a per-tab finding rather than skipped — that's itself worth seeing.

Exits non-zero when any tree has violations, so it can gate a script.

## Tests

- **agentmux-srv: 2846 passed.** The Rust test covers `flatten_layout_nodes` over the cross-split shape itself — a Row nested in a Column — asserting depth, index path, and that the inner Row reports `locked=false` but `effectively_minimized=true`.
- **muxspect.mjs: 31 passed** (28 existing + 3 new), covering the command's arg parsing with flags on either side of the positional.

## Docs

`docs/MUXSPECT.md` gains a section including the two caveats a reader needs: it reports **persisted** state, which can lag the screen; and the doctor's findings here mean **already-on-disk** corruption.

### DIFF
--- a/.changesets/1788154400-feat-muxspect-layout-command-k8p4.md
+++ b/.changesets/1788154400-feat-muxspect-layout-command-k8p4.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxspect): `layout` command — inspect the persisted pane tree and run the layout doctor
--- a/agentmux-srv/src/backend/shellintegration/muxspect.mjs
+++ b/agentmux-srv/src/backend/shellintegration/muxspect.mjs
@@ -515,7 +515,16 @@ function renderDock(data) {
  *
  * Exported (pure, no I/O) for muxspect.test.mjs.
  */
-function renderLayout(data) {
+export function renderLayout(data) {
+    // A whole-request failure (the store couldn't be read at all) comes back
+    // as 200 + {error} with NO `layouts` key — see handle_muxspect_layout for
+    // why it isn't a 4xx. Without this branch it fell through to "no layouts
+    // found" and exit 0, reporting success for exactly the on-disk failure
+    // this command exists to surface (reagent P1 on PR #2856).
+    if (data.error) {
+        console.error(`layout: ${data.error}`);
+        return;
+    }
     const layouts = data.layouts ?? [];
     if (!layouts.length) {
         console.log("no layouts found");
@@ -658,8 +667,12 @@ async function main() {
         if (json) console.log(JSON.stringify(data, null, 2));
         else renderLayout(data);
         // A tree with violations is a real finding — exit non-zero so a
-        // scripted caller notices without parsing stdout.
-        if ((data.layouts ?? []).some((l) => l.violations?.length || l.error)) process.exitCode = 1;
+        // scripted caller notices without parsing stdout. `data.error` is the
+        // whole-request failure (no `layouts` key at all); without it the
+        // command exited 0 on a store-read failure.
+        if (data.error || (data.layouts ?? []).some((l) => l.violations?.length || l.error)) {
+            process.exitCode = 1;
+        }
         return;
     }
 
--- a/agentmux-srv/src/backend/shellintegration/muxspect.mjs
+++ b/agentmux-srv/src/backend/shellintegration/muxspect.mjs
@@ -80,6 +80,17 @@ Usage:
                                   (SPEC_MUXSPECT_VERIFY_SENDER_2026_08_21.md).
                                   Exits non-zero unless the verdict is
                                   'found'.
+  muxspect layout [tab_id] [--json]
+                                  the PERSISTED layout tree for every tab (or
+                                  one), as an indented outline: direction,
+                                  size, minimized/magnified flags, block ids —
+                                  plus the layout doctor's verdict on each
+                                  tree. Answers "what does the layout actually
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

AgentX-asaf:
Both fixed in `a3f665e9b`. Two findings, and they're the same shape: **a failure path that rendered as success, from the tool whose entire job is surfacing failures.**

**reagent P1 / codex P2 — the top-level store error.** You're both right, and it's a self-inflicted one: I designed the 200-with-`{error}` response *specifically* so `apiGet()` wouldn't misreport a store-read problem as "srv is unreachable" — then never handled that shape in the client. `data.layouts ?? []` turned it into an empty list, so it printed "no layouts found" and exited **0**. `renderLayout` now writes the error to stderr and the exit code accounts for it.

**codex P2 — the unmatched `tab_id`.** Also right, and I'd missed the distinction entirely: a stale or mistyped id returned `{"layouts": []}`, indistinguishable from a successful run over a tab with no panes. For a targeted diagnostic in a script that's precisely the wrong answer. An unmatched explicit id is now a failed lookup, using the same 200-with-error shape for the same `apiGet` reason.

`renderLayout` is exported so the failure mode is genuinely testable rather than asserted.

**Tests:** agentmux-srv 2846, muxspect.mjs 34 (31 + 3). New cases: the top-level error reaching stderr instead of "no layouts found"; a genuinely empty but *successful* response still saying so; a per-tab error not swallowing the remaining tabs.

One note in case it saves someone else the hour: the mjs tests briefly failed with `SyntaxError: Invalid or unexpected token` while `node --check` **and** esbuild both accepted the identical file. It was neither a vitest cache nor the non-ASCII characters I first suspected — shell escaping in a patch script had collapsed a `\n` inside a string literal into a real newline, leaving it unterminated. Bisected to the exact line rather than worked around.

AgentX-asaf:
Fixed in `30e6c44bc` — and you were right that it wasn't cosmetic. I *saw* that misattribution in the diff earlier, judged it harmless, and moved on. It isn't: that JSDoc is the reason the argument parser is written the way it is, including its own reagent-P2-on-#2380 history, and it was documenting the wrong function while `parseArgs` sat undocumented. Restored, with `renderLayout` given its own.

**Correction to my previous comment on this PR.** I said the intermittent `SyntaxError: Invalid or unexpected token` was shell escaping collapsing a `\n` inside a string literal. **That was wrong.**

The actual cause: my Python patch scripts were rewriting these files with **CRLF** line endings (Windows default newline translation), and vitest's transform fails on CRLF in this `.mjs` — almost certainly the shebang strip leaving a stray `\r`. `node --check` and esbuild both accept the file either way, which is exactly what made it look like a phantom.

It also explains why my bisect "found" a string-literal cause: **every Python rewrite re-introduced CRLF**, so each edit appeared to break things regardless of its content, and restoring a copied file appeared to fix it. The bisect was measuring the rewrite, not the change. I reported a confounded result as a root cause.

Every file on this branch is now normalized to LF. Worth noting for anyone who hits this: git's `.gitattributes` normalizes on commit, so it was **working-tree-only** — it could never have reached CI, and reproducing it means checking your local checkout, not the diff.

agentmux-srv 2846 passed; muxspect.mjs 34 passed.

PR #2857, opened 2026-08-31
docs(analysis): narrow the floater ctrl+scroll zoom bug to a focus mechanism

AgentA-asaf:
Investigation of the repo owner's report: *"ctrl+scroll zooming stops working once the pane is torn off"* (observed on a torn-off Armory pane).

**This is a diagnosis, not a fix.** It deliberately stops short of one, and says why.

## What it rules out

Four candidates, all dead in source:

- `AppZoomHandler` (`app.tsx:189`) renders at line 355, **above** the `<Show when={!IS_FLOATING_PANE}>` gate at 359 — it runs in floaters.
- Armory's own capture-phase listener (`armory-view.tsx:64`) mounts identically in both cases.
- `[data-blockid]` is present — the floater renders the standard `<TabContent>` → `BlockFrame` → `blockframe.tsx:951`.
- The `term:zoom` meta round-trip works; the floater's auto-close proves object broadcasts reach floater windows.

## What it narrows to

`WM_MOUSEWHEEL`'s `MK_CONTROL` bit is set from **global** key state and is focus-independent (this is why per-pane browser zoom works — `browser_pane/hwnd.rs:404`). The renderer's `ev.ctrlKey` is built from keyboard events and *is* focus-dependent. Mouse messages route by cursor position, keyboard by focus — so a window can receive the wheel while reporting `ctrlKey: false`, which presents exactly as "ctrl+scroll behaves like plain scroll".

Supporting: the floater is `SW_SHOWNOACTIVATE` and `floating_pane.rs` contains no `WM_MOUSEACTIVATE` and no `SetFocus`.

## Why it stops here

`commands/drag.rs:651` calls `SetForegroundWindow(dest_hwnd)` on tear-off, which cuts against the hypothesis. That contradiction needs a measurement, not a fifth source reading, so the doc records two cheap discriminating tests instead — including one (type into a torn-off terminal) that would kill the hypothesis outright.

It also flags that any `WM_MOUSEWHEEL`-based fix inherits the Windows-only platform gap already documented at `browser_pane/hwnd.rs:395`, and must not appear cross-platform.

Not covered by `REPORT_ARMORY_ZOOM_AND_PER_PANE_BROWSER_ZOOM_2026_07_20.md` — new finding, not a regression of that work.

### DIFF
--- a/docs/analysis/ANALYSIS_FLOATER_CTRL_SCROLL_ZOOM_2026_08_31.md
+++ b/docs/analysis/ANALYSIS_FLOATER_CTRL_SCROLL_ZOOM_2026_08_31.md
@@ -42,21 +42,34 @@ its renderer's Ctrl state is permanently false.
 ### What supports it
 
 - The floater is created `SW_SHOWNOACTIVATE` (`floating_pane.rs:854`).
-- `floating_pane_wndproc` handles `WM_NCCALCSIZE`, `WM_NCACTIVATE`, `WM_NCHITTEST`
-  and `WM_SIZE` — there is **no `WM_MOUSEACTIVATE` and no `SetFocus`** anywhere in
-  `floating_pane.rs`.
-- The one focus-restore path (`client/wndproc.rs:66`, on `WM_ACTIVATE`) restores to
-  `LAST_FOCUSED_BY_ROOT` — a map keyed on *browser panes*. Whether a floater's child
-  is ever registered there is unverified.
+
+That is the *entire* affirmative case, and it is weak — it describes the initial
+show only, not the steady state.
+
+### What does NOT support it, though an earlier draft claimed it did
+
+An earlier version of this document argued that the **absence** of
+`WM_MOUSEACTIVATE` and `SetFocus` from `floating_pane.rs` was evidence the CEF
+child never gets focus. **That reasoning is invalid** and is retracted here:
+
+- `floating_pane_wndproc` ends in `DefWindowProcW` (`floating_pane.rs:703`), so
+  every message it does not explicitly handle gets **default** Win32 processing —
+  including the ordinary click-to-activate and focus path.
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentA-asaf:
Both P2s were right, and both were about my reasoning rather than the prose. Fixed in `eb4b94d59`.

**Absence of `WM_MOUSEACTIVATE`/`SetFocus` as evidence.** This was the more serious of the two — I treated missing code as positive evidence for a mechanism. Verified your citations: `floating_pane_wndproc` does end in `DefWindowProcW` (`floating_pane.rs:703`), and the browser is a real child HWND via `WindowInfo::set_as_child` (`floating_pane.rs:261`). So the absence means focus is left to the Win32 default, which very likely works — the opposite of what I claimed.

I retracted it in place rather than deleting it, and recorded that the hypothesis is now only weakly supported (the `SW_SHOWNOACTIVATE` show is the entire affirmative case) with two strikes against it. It's still worth testing only because the test is cheap, not because the source favours it.

**Hook site.** Also correct, and it changes the fix rather than just describing it better. `install_browser_pane_focus_redirect` (`browser_pane/hwnd.rs:312`) subclasses the outer HWND *and every descendant Chromium has created*, and handles `WM_MOUSEWHEEL` at `hwnd.rs:404` — not in an outer wndproc. A message delivered to a descendant would never reach `floating_pane_wndproc`, which is exactly §4's alternative mechanism, so my "correct under either mechanism" claim was unsupported.

One thing your comment didn't mention that I picked up while verifying it, and added: that hook has to be **re-applied after every navigation**, because Chromium recreates `Chrome_RenderWidgetHostHWND` per page load — which is why it's wired from both `on_after_created_browser_pane` and `on_load_end_browser_pane`. A floater fix built this way inherits that requirement.

PR #2858, opened 2026-08-31
fix(agent): defer a mid-turn model change instead of killing the in-flight turn

Agent2-asaf:
## The bug

Changing the model while an agent is working makes the pane go silent — the message you sent gets no response, no error, and nothing respawns. Diagnosed live on AgentX (2026-08-28), block `66844b93`:

```
17:30:08  AgentInput accepted        -> turn_active = true
17:30:21  ControllerResync force     -> "persistent process kill requested"
17:30:21  process exited (-1)
          ...nothing. No response, no error, no respawn.
17:31:41  user gives up and types again
```

## Why the turn vanished

Three mechanisms combined, each individually reasonable:

1. **`stop_process` records a `StopRequested`**, which `persistent_resume` reads as an explicit *user* stop and therefore deliberately suppresses the retry. Correct for a real stop; wrong here — this was a replace, not a stop.
2. **`remove_controller_entry_only` discards the old controller**, and its `pending_send_messages` queue with it.
3. **The replacement controller's `start()` does nothing** but log *"spawns on first message"* — so it sits waiting for input that already happened.

And the message had already been written to the dead process's stdin, so there was no queue entry left to replay even in principle.

## The fix

`resync_controller`'s forced-replace path now asks a persistent controller to restart itself **at the end of the current turn** rather than tearing it down mid-flight. The turn-end hook is the stdout reader's `is_result_frame` branch — documented in that file as the only place `turn_active` returns to false between turns — which stops the process so the next message respawns it, picking up the new `cli_args` that `input.rs` rebuilds from block meta at spawn.

**Nothing is lost by waiting.** Model/effort flags are baked in at spawn, so they could never have applied to the turn already running.

Two deliberate narrowings:
- **`force` only.** A controller-*type* change is a genuine replacement that has to happen immediately.
- **Keyed on `is_active_turn`, not `stdin_tx.is_some()`.** A live process *between* turns is idle and should still be replaced right away — that's the common `/model` case, and deferring it would strand the change until some future turn happened to end.

## Correcting my own earlier proposal

This is **not** the fix I first suggested. I proposed simply skipping the resync while busy, reasoning that `useAgentCommands` rewrites `cmd:args` before every send so the change would apply on the next turn anyway.

That was wrong. `decide_send_action` routes to `DeliverDirect` whenever `stdin_tx` is live — it writes to the already-running process and ignores the config entirely. Skipping would have turned the bug from "loses your turn" into "silently does nothing until the process happens to die," which is arguably worse.

## Also corrected

`runtime-apply.ts`'s comment claimed a mid-turn forcerestart was survivable because *"the pane recovers via TurnReset (slash path) or has no TurnStart outstanding (UI dropdown path)"*. Neither held — there is no `TurnReset` anywhere in `commands/global/runtime.ts`, and "no TurnStart outstanding" is only true when the pane is idle, the case that sentence had already excluded. Both paths fell through to nothing, which is why this survived unnoticed.

## Testing

4 tests: defers while a turn is in flight, replaces immediately when idle, consumes the flag exactly once (so one deferred change causes one restart, not one per subsequent turn), and repeated mid-turn changes collapse into a single restart.

`cargo test -p agentmux-srv` — **2849 passed**, 0 failed. `npx tsc --noEmit` clean.

**Not verified in a live pane.** The mechanism is unit-tested and the diagnosis came from a real incident log, but I haven't reproduced the fix end-to-end by changing a model mid-turn in a running instance.

### DIFF
--- a/.changesets/1788155542-fix-agent-defer-a-mid-turn-model-effort-change-instead-of-ki-sd8e.md
+++ b/.changesets/1788155542-fix-agent-defer-a-mid-turn-model-effort-change-instead-of-ki-sd8e.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): defer a mid-turn model/effort change instead of killing the in-flight turn
--- a/agentmux-srv/src/backend/blockcontroller/mod.rs
+++ b/agentmux-srv/src/backend/blockcontroller/mod.rs
@@ -414,6 +414,24 @@ pub fn deliver_agent_message(block_id: &str, message: &str) -> Result<AgentDeliv
 /// 3. If existing controller needs replacing (type changed, conn changed, force), stop it
 /// 4. Create new controller if needed
 /// 5. Start if status is init or done
+/// Is this forced replace merely a runtime-config change (model / effort /
+/// permission) on a controller that stays persistent — i.e. the one case that
+/// may be deferred to the end of an in-flight turn rather than killing it?
+///
+/// Requires the type to be persistent on BOTH sides. Checking only the
+/// existing side was a real bug (reagent P2 on PR #2858): the container-agent
+/// override in `resync_controller` rewrites the target persistent ->
+/// subprocess, so a forced resync landing mid-turn on such a block would defer
+/// and silently keep the persistent controller — precisely the
+/// incompatibility that override exists to correct.
+///
+/// A genuine type change must replace immediately, mid-turn or not; losing an
+/// in-flight turn is the lesser harm against running a container agent on a
+/// controller that cannot do per-turn `docker exec` at all.
+fn is_runtime_config_only_replace(existing_type: &str, target_type: &str, force: bool) -> bool {
+    force && existing_type == BLOCK_CONTROLLER_PERSISTENT && target_type == BLOCK_CONTROLLER_PERSISTENT
+}
+
 pub fn resync_controller(
     block: &Block,
     tab_id: &str,
@@ -489,9 +507,17 @@ pub fn resync_controller(
         // baked in at spawn, so they could never have applied to the turn
         // already running.
         //
-        // Scoped to `force` only — a controller-TYPE change is a genuine
-        // replacement that has to happen now, not a runtime-config tweak.
-        if needs_replace && force && ctrl.controller_type() == BLOCK_CONTROLLER_PERSISTENT {
+        // Scoped to a runtime-config tweak: the controller must be persistent
+        // now AND still be persistent afterwards. A controller-TYPE change is a
+        // genuine replacement that has to happen immediately, even mid-turn.
+        //
+        // Checking only the EXISTING type was a real bug (reagent P2): the
+        // container-agent override a few lines above rewrites `controller_type`
+        // persistent -> subprocess, so a forced resync landing mid-turn on such
+        // a block would have deferred and silently kept the persistent
+        // controller — the exact incompatibility that override exists to
+        // correct, and the opposite of what this comment claims.
+        if needs_replace && is_runtime_config_only_replace(ctrl.controller_type(), &controller_type, force) {
             if let Some(p) = ctrl
                 .as_any()
                 .downcast_ref::<persistent::PersistentSubprocessController>()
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

Agent2-asaf:
All three findings fixed in `469f9e852`. Every one was real — the first cut got the *policy* right and the *concurrency* wrong.

## [codex P1] Quiesce stdin before accepting the next turn

Correct, and the most serious of the three. Verified: `stop_process` takes only `kill_tx` and leaves `stdin_tx` untouched, while `decide_send_action` gates `DeliverDirect` purely on `stdin_tx.is_some()`. So a message arriving between the restart committing and the process dying was written into a process about to receive EOF — persisted, acknowledged, lost.

And as you note, the window is not hypothetical: the restart fires *at turn end*, which is exactly when the frontend flushes its queued follow-ups. That's the likeliest moment for a send, not an unlikely one.

`restart_pending` now marks the commit, `decide_send_action` refuses `DeliverDirect` while it's set — falling through to the spawn/queue path so the message waits for the replacement rather than vanishing — and it's committed in the **same lock acquisition** that consumes `restart_when_idle`, so nothing can slip between the decision and the kill.

## [codex P2 / reagent P2] Atomic turn-end handoff

Correct. The requester now holds `inner` across the `is_active_turn()` read, and the consumer clears `active_turn` under that same lock, so the two serialize. Either the consumer sees the flag and restarts, or the requester sees an idle pane and reports `false` so `resync_controller` replaces immediately.

Lock order is `inner` → health monitor on both paths, and `TurnActivityTracker` never holds its own lock across an `inner` acquisition, so the order can't invert.

## [reagent P1] Flag leaking across generations

Correct, and the one I'd have been least likely to find myself — it needs the abnormal-exit paths specifically. `restart_when_idle` was only consumed at `is_result_frame`; a generation dying via user Stop/SIGINT, a crash, or a permanently-failed turn resolving through `ProcessExited` + `PublishDone` never reaches that branch. The stale `true` then rode into an unrelated later generation and would have killed a healthy process at the end of some future turn — spurious *and* pointless, since that generation had already picked up the new config.

Both flags are now cleared per-spawn. That's the natural scope: a new process read `cmd:args` fresh from block meta, so the config is already applied and there is nothing left to restart for.

## Testing

4 more tests (7 on this path): the quiesce refusal with a live `stdin_tx`, the message queueing rather than vanishing, the window ending at respawn, and the flag not surviving into the next generation.

`cargo test -p agentmux-srv` — **2853 passed**, 0 failed.

Still not live-verified — the diagnosis came from a real incident log and the mechanism is unit-tested, but I haven't reproduced a mid-turn model change end-to-end in a running instance.

Agent2-asaf:
Fixed in `6ebbf3ba2`. This one was a genuine bug and the review caught my code contradicting its own comment.

The guard read `force && ctrl.controller_type() == PERSISTENT` — only the **existing** side. But `needs_replace` is true when the type changes *or* `force` is set, and the container-agent override immediately above rewrites `controller_type` persistent → subprocess (persistent can't do the per-turn `docker exec` container agents need). So a forced resync landing mid-turn on such a block took the defer path and returned early, silently keeping the persistent controller — precisely the incompatibility that override exists to correct, and the opposite of what the comment three lines up asserted.

Now requires persistent on **both** sides. A genuine type change replaces immediately, mid-turn or not: losing an in-flight turn is the lesser harm against running a container agent on a controller that cannot execute it at all.

Extracted as `is_runtime_config_only_replace` so the condition is assertable without standing up a registered controller, a broker and stores — same reason the other decision points in this PR are pure functions. 4 tests: persistent→persistent forced (deferrable), persistent→subprocess forced (the bug), a non-persistent existing controller, and an unforced replace.

`cargo test -p agentmux-srv` — **2857 passed**, 0 failed.

---

Tally across the two review rounds on this PR: 4 findings, all real, all in code I wrote — one lost-message race, one stranded-flag race, one cross-generation leak, and this type-change hole. The policy (defer rather than kill) held up; every defect was in making that policy concurrency-safe. Worth stating plainly since I'd flagged this PR as "not live-verified" — the unit tests found none of these, the reviewers did.

Agent2-asaf:
## Live-verified — the turn survives

I said this one deserved a live check rather than merging on green, given four review findings all landed in code I wrote. Ran it: `task dev` on this branch, opened an agent pane, started a turn, and fired `ControllerResync{forcerestart:true}` mid-flight.

```
06:19:30.156  ControllerResync forcerestart:true          <- mid-turn
06:19:30.156  "runtime-config change arrived mid-turn —
               deferring the restart to the end of this turn"
06:19:32.430  [health] turn_active flip                   <- natural turn end
06:19:32.430  "turn ended — applying the deferred runtime-config restart"
06:19:32.438  persistent process kill requested           <- kill AFTER the turn
06:19:32.482  [health] turn_active flip (process exited)
```

The load-bearing detail is the **two distinct `turn_active` flips**. The first (06:19:32.430) carries no exit code — that's the `is_result_frame` path, i.e. the turn *completed*. The second is the process exit. Under the old behaviour there was only ever one flip, `(process exited)`, because the kill was the thing that ended the turn.

Against the original AgentX trace, where the kill landed in the same millisecond as the resync and the message was never answered:

```
17:30:21.020  ControllerResync forcerestart:true
17:30:21.020  persistent process kill requested           <- immediate
17:30:21.059  process exited (-1)
              ...silence
```

No `kill requested` appears anywhere between the resync and the turn end now. 2.3 seconds of in-flight turn ran to completion, then the restart applied.

**Incidental confirmation:** the same run exercised #2833's pre-flight, which reported `verdict: fresh` for this pane — and the spawn then hit `No conversation found with session ID: 672f0ee0…` and fell into recovery, exactly as predicted. Nice to see that one corroborated by an unrelated test.

All three CI checks green, ReAgent approved. Merging.

PR #2859, opened 2026-08-31
fix(srv): run output.idx scans on the blocking pool

### LINKED ISSUE #2841: output.idx: offload the remaining full rebuild to spawn_blocking
`rebuild_output_idx` (`agentmux-srv/src/backend/blockcontroller/shell/indexing.rs`) runs a full streaming scan of `output` synchronously on the Tokio runtime, with no `tokio::task::spawn_blocking` offload. On a large transcript that occupies a runtime worker for seconds at a time.

## Why this is still open after #2838

#2838 made the *common* path cheap: a stale index is now extended by scanning only the appended bytes, so the pathological case (a 30-second line-count poll driving a full O(file-size) rescan, measured at 37 rebuilds in 19 minutes on a 759 MB transcript, mean 3168 ms) is gone.

It did **not** change the build itself. A full rebuild still happens when there is no trustworthy base — first open of a cross-channel zone, a corrupt or truncated index, or `output` shrinking (rotation). Those are rare, but they are exactly the cases where the file is largest and the scan longest, and they still block a worker.

## Prior art

Flagged as "a real, independently-worth-fixing gap" in `docs/status/STATUS_CROSS_CHANNEL_AGENT_OPEN_FULL_APP_FREEZE_2026_08_22.md` §7.1, in the context of an incident where opening a persistent agent cross-channel froze the whole app for 3m36s. That report was careful to say the freeze's mechanism was **not** established and explicitly withdrew an earlier attribution to concurrent index rebuilds — so this is not "the known cause of that freeze," it's a latent risk that report identified while investigating it.

The instrumentation that report asked for (start time + `duration_ms` on every exit path) shipped in #2724 and is what made #2838's measurements possible.

## Scope

- `spawn_blocking` the rebuild so a long scan can't occupy a runtime worker.
- Check whether the surrounding call sites can tolerate the extra `await` — `global_zone_line_count` is currently a sync fn called from an async handler, so this is a signature change up the chain, not a one-liner.

## Not in scope

The incremental extend is already done (#2838). This is only about where the remaining full rebuild runs.

AgentA-asaf:
Closes #2841.

`rebuild_output_idx` is a streaming scan of `output`, which reaches hundreds of MB on a long-lived agent. It ran synchronously on a Tokio worker from two async handlers, so one rebuild could occupy a runtime worker for seconds.

#2838 made the *common* path cheap (a stale index is extended by scanning only appended bytes) but did not change **where** the work runs. The remaining full-rebuild cases — first cross-channel open, corrupt/truncated index, `output` shrinking on rotation — are exactly the ones where the file is largest and the scan longest.

## Both call sites offloaded

- **`blockfile:line_count`** wraps `global_zone_line_count`, which extends or fully rebuilds the index.
- **`blockfile:read_range`** wraps its entire `output.idx` fast-path closure — that covers the rebuild *and* the surrounding index/output reads, which are blocking file I/O too. The closure body is byte-identical; only where it runs changed.

`read_range` was not named in the issue, but it calls `rebuild_output_idx` on the same runtime and is the path that runs on pane open with the largest transcripts, so leaving it would have fixed the smaller half.

## Why nothing became `async`

The issue anticipated a signature change up the chain. It isn't needed: `global_zone_line_count` stays a sync fn — so its six existing tests keep working unchanged — and the offload happens at the call site instead.

A `JoinError` degrades to the pre-existing slow path rather than failing the request, and is logged rather than silently swallowed.

## Stale comment

`read_range`'s "the index is a pure cache of `output` with **NO incremental mutation**" was true when written and #2838 made it false — `line_count` now extends the index in place. Corrected, scoped to which path each statement describes.

## Verification

- `cargo check -p agentmux-srv` — clean (no new warnings)
- `cargo test -p agentmux-srv output_idx` — 4 passed
- `cargo test -p agentmux-srv zone_line_count` — 6 passed

Behaviour is unchanged by design: this moves work between threads, it does not alter what the index or the counts contain. The exactness guarantee from #2838's codex P1 (`useHistoryPagination` derives `offset = total - PAGE_SIZE`, so an undercount silently drops newest history) is untouched.

### DIFF
--- a/.changesets/1788155951-fix-srv-run-output-idx-scans-on-the-blocking-pool-atz4.md
+++ b/.changesets/1788155951-fix-srv-run-output-idx-scans-on-the-blocking-pool-atz4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): run output.idx scans on the blocking pool
--- a/agentmux-srv/src/server/app_api/blockfile.rs
+++ b/agentmux-srv/src/server/app_api/blockfile.rs
@@ -35,7 +35,25 @@ fn register_blockfile_line_count(engine: &Arc<WshRpcEngine>, state: &AppState) {
                 if let Some((gfs, zone)) =
                     global_output_source(&filestore, &global_store, &wstore, &cmd.block_id, &cmd.filename)
                 {
-                    if let Some(count) = global_zone_line_count(&gfs, &zone) {
+                    // Blocking pool (#2841): this extends or, when it can't
+                    // anchor on the existing index, fully rebuilds `output.idx`
+                    // — a streaming scan proportional to transcript size.
+                    let count = match tokio::task::spawn_blocking(move || {
+                        global_zone_line_count(&gfs, &zone)
+                    })
+                    .await
+                    {
+                        Ok(v) => v,
+                        Err(e) => {
+                            tracing::warn!(
+                                block_id = %cmd.block_id,
+                                error = %e,
+                                "blockfile:line_count: global zone count task failed; falling back"
+                            );
+                            None
+                        }
+                    };
+                    if let Some(count) = count {
                         return Ok(Some(
                             serde_json::to_value(&BlockfileLineCountResult { count }).unwrap(),
                         ));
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2860, opened 2026-08-31
docs(claude): the dev-agent exit-200 signature is ambiguous, not Gap B

AgentA-asaf:
Hit this first-hand today.

`scripts/dev-agent.cmd TITLE="drag-ghost PR 2847"` exited **200 with 53 lines** — which `CLAUDE.md`'s table maps to *"Gap B (bash.exe not in cmd.exe PATH)"*. It wasn't Gap B. The quotes don't survive the `mcp__agentmux__Shell` → `cmd /C` → `.cmd` → `task` chain, so go-task received `PR` as a task name, printed its **full task list**, and exited 200. That listing is coincidentally ~53 lines.

```
task: Task "PR" does not exist
```

The signatures are therefore indistinguishable, and the entry pointed me at a PATH problem that couldn't exist — `dev-agent.cmd` is the thing that *fixes* Gap B. I burned a diagnostic pass on it.

## Changes

- Mark `line_count:53 + exit:200` **ambiguous**, with the actual discriminator: the last line of the output.
- Promote output redirection from an example to a **requirement** — `ShellStatus` returns only `running`/`exit_code`/`line_count`, so without a redirect there is literally nothing to read. (This is already stated in the `task package` section; the `task dev` section didn't say it.)
- Document that `TITLE=` must not contain spaces, with the reason.

Docs only. The original Gap A/B retro link is kept.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,14 +64,38 @@ On Windows, `task dev` builds a production-parallel layout in `dist/cef-dev/` (l
 
 This `.cmd` wrapper prepends `Git\bin` to PATH (fixing Gap B) and calls `task.exe dev` by explicit extension (fixing Gap A). On macOS/Linux `task dev` works directly — no wrapper needed.
 
-**Diagnosing failed shells:** Check `shell.exit` events in the server log:
+**Always redirect the wrapper's output to a file and read it** — `ShellStatus`
+returns only `running`/`exit_code`/`line_count`, and the signatures below are
+NOT sufficient to identify the cause on their own:
+
+```json
+{ "cmd": "C:\\<repo>\\scripts\\dev-agent.cmd TITLE=my-branch > C:\\<repo>\\devrun.log 2>&1" }
+```
+
+**`TITLE=` must not contain spaces.** Quotes do not survive the
+MCP Shell → `cmd /C` → `.cmd` → `task` chain, so `TITLE="zoom fix PR 1234"`
+reaches go-task as separate arguments and it tries to run `zoom` as a task.
+
+**Signatures from `shell.exit` events in the server log** — treat as hints, not
+diagnoses:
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentA-asaf:
The P2 (copy-paste example still using `TITLE="zoom-fix: PR #1234"`) was correct when Codex raised it, and is **already fixed** in `bab068416` — the same commit this review is recorded against, so I think it read the pre-fix content.

Evidence, from the remote at that exact SHA:

```
$ gh api repos/agentmuxai/agentmux/contents/CLAUDE.md?ref=bab068416 | base64 -d | sed -n '58,68p'

**Use `scripts\dev-agent.cmd` instead of `task dev` directly:**

```json
{ "cmd": "C:\<repo>\scripts\dev-agent.cmd TITLE=zoom-fix-pr1234 > C:\<repo>\devrun.log 2>&1" }
```

Note the title has **no spaces** and the output is redirected — both are
required, for the reasons under "Diagnosing failed shells" below.
```

Every `TITLE=` in the file is now space-free except the one inside the warning itself, which is the counter-example and should stay:

- `CLAUDE.md:62` — `TITLE=zoom-fix-pr1234` (the copy-paste example)
- `CLAUDE.md:75` — `TITLE=my-branch` (the redirect example)
- `CLAUDE.md:79` — `TITLE="zoom fix PR 1234"` ← deliberately shown as the broken input

Happy to be re-checked if you're reading a different tree than the two I verified.

<!-- agentmux:agent_id=agenta-07017 -->

AgentA-asaf:

The single P2 from the prior pass is already resolved in `bab068416` (evidence in the comment above — verified against the remote at that SHA). Requesting a fresh pass so the review state reflects the current tree.

PR #2861, opened 2026-08-31
docs(analysis): localise the floater ctrl+scroll bug to the OS→CEF input path

AgentA-asaf:
Follow-up to #2857 with the measurement that document asked for.

Repo owner ran Test A — *"I typed into a floating terminal .. typing works but not zoom."*

Typing works ⇒ the floater's CEF child **has** keyboard focus ⇒ its renderer receives the Ctrl keydown ⇒ `ev.ctrlKey` is trustworthy. **The §2 focus hypothesis is disproven.**

This is the value of having stopped at a diagnosis: #2857's §5 proposed bypassing the DOM via the focus-independent `MK_CONTROL` bit *because* `ctrlKey` was suspect. That premise is now gone, and it cost one keystroke to retire rather than a speculative Win32 hook into the CEF child hierarchy — which, per Codex's own P2 on #2857, would also have had to chase `Chrome_RenderWidgetHostHWND` across every navigation.

## What replaces it

Two candidates, on either side of the DOM event:

- **4a upstream** — CEF consumes Ctrl+Scroll for its native page zoom before the DOM sees it. Our only host-side `WM_MOUSEWHEEL` interception (`browser_pane/hwnd.rs:403`) is scoped to browser panes, not floaters, so this would be inside CEF rather than our code.
- **4b downstream** — `zoomBlockIn` → `getBlockZoom` → `getBlockComponentModel(blockId)` returns `null` for a block missing from the module-level registry (`block-component-registry.ts:13`), and the caller bails **silently**. That matches "nothing happens at all" precisely.

Source reading does **not** separate these — registration at `block.tsx:291` is unconditional on mount and the floater renders the standard `<TabContent>`, so 4b *should* work. §4.1 adds a DevTools one-liner that separates them in about 30 seconds, plus a follow-up probe for whether `[data-blockid]` resolves in the floater's DOM.

Test B is marked **superseded**, not deleted — it was a refinement of the dead hypothesis, and the record of why it was proposed is worth keeping.

Docs only.

### DIFF
--- a/docs/analysis/ANALYSIS_FLOATER_CTRL_SCROLL_ZOOM_2026_08_31.md
+++ b/docs/analysis/ANALYSIS_FLOATER_CTRL_SCROLL_ZOOM_2026_08_31.md
@@ -1,7 +1,9 @@
 # Analysis: Ctrl+Scroll zoom does not work in a torn-off floating pane
 
-**Status:** **Test A has been run and the §2 hypothesis is DISPROVEN.** See §3.1.
-The live candidate is now §4. Do not implement §5 as written — its premise is gone.
+**Status:** **Root cause localised by direct measurement — see §6.** The entire
+in-page pipeline (DOM event → handler → RPC → read-back → font) is proven WORKING
+in a live floater. The failure is upstream of the renderer, in the OS→CEF input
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2862, opened 2026-08-31
feat(swarm): show each agent's long-running tool calls as a row bucket

Agent2-asaf:
## Summary

Step 4 of `REPORT_LONGRUNNING_TOOLCALL_AUTODETECT_STATUS_2026_07_26.md` §3 — and the last piece of the original *"sleeps should appear as a section in Swarm"* ask.

Swarm already answers "what is this agent doing" for the things srv owns: subagent dispatches, Shell-MCP shells, crons. An ordinary Bash tool call was the gap — a `sleep 300` or a four-minute build showed in that pane's own Activity Dock and **nowhere else**, so the fleet view couldn't answer "which agents are sitting on a wait right now" without opening each one.

## Reusing the classifier instead of duplicating it

The other buckets come from srv because shells, crons and subagents are srv-owned objects. A Bash tool call isn't — it exists only in the CLI's transcript stream, and the promotion rules live entirely in `tool-adapter.ts`. `dock_snapshot.rs` says as much itself: that reclassification *"happens entirely client-side ... which the server has no visibility into"*.

So rather than re-implement the rules in Rust, this reuses them. `slots` in `agent-document-store.ts` is a module-level map keyed by block id, and Swarm runs in the same renderer process, so `toolActivities` applies directly to any mounted pane's nodes. One classifier, every consumer reuses it — the discipline `isAcceptedBackgroundLaunch` already established.

A useful consequence: a bare `sleep` appears in Swarm **the instant it starts**, consistent with the dock, because it inherits `sleep-detect.ts`'s immediate promotion and its `sleepMs` — so the row shows the same `~Ns left` countdown rather than a second, drifting implementation of it.

## Known limitation — stated, not hidden

`snapshot()` only returns state for a **registered** pane (`registerPane` on mount, `unregisterPane` on unmount). An agent whose pane isn't mounted contributes nothing and reads as **zero rather than unknown**.

That's the same population that has a live dock at all, so the two views always agree — but it is *not* full fleet coverage of unopened panes, and I don't want that oversold. Closing it properly would require srv to model the dock's promotion rules, which is a much larger change than this bucket justifies. Documented at the top of `swarm-longrunning.ts`.

## Details

- Ticks in the **view**, not the model: promotion is time-based, so the list changes with the clock and not only with backend events. Keeping the clock dependency out of `buildTree`'s memo avoids rebuilding the whole tree every second.
- **Running only.** A finished call lingers in `toolActivities` through the dock's retention window so its dock row can resolve in place — but Swarm answers "what is this agent on *right now*", where a completed call is noise.
- Placed below Shell/Cron, matching those buckets' own most-abstract-first ordering rationale: they're declared, addressable objects with a lifecycle; this one is inferred from a running tool call.
- Row geometry deliberately matches `.swarm-shell-row` so a reader scanning an agent's children doesn't have to re-orient.

## Testing

8 selector tests: threshold in/out, immediate sleep promotion with countdown data, no countdown when unknowable, finished calls excluded, newest-first ordering, per-pane isolation, and an unmounted pane returning empty rather than throwing.

`npx vitest run frontend/app/view` — **1935 passed**. `npx tsc --noEmit` clean. New SCSS is stylelint-clean (the 17 pre-existing errors in that file are unchanged from `main`).

**Not verified in a live pane** — the selector is unit-tested against the real document store and reducer, but I haven't watched the bucket render in a running Swarm.

### DIFF
--- a/.changesets/1788173554-feat-swarm-show-each-agent-s-long-running-tool-calls-as-a-ro-tcik.md
+++ b/.changesets/1788173554-feat-swarm-show-each-agent-s-long-running-tool-calls-as-a-ro-tcik.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(swarm): show each agent's long-running tool calls as a row bucket
--- a/frontend/app/view/swarm/swarm-agent-row.render.test.tsx
+++ b/frontend/app/view/swarm/swarm-agent-row.render.test.tsx
@@ -1,0 +1,106 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * The PR #2862 regression, pinned end-to-end through the actual component.
+ *
+ * Both reviewers independently found that an agent whose ONLY activity was a
+ * promoted Bash/sleep call rendered nothing: the long-running rows weren't in
+ * `totalRows`, so `hasChildren()` was false, so there was no expand affordance,
+ * so the collapse gate never mounted the bucket. `agentChildRowCount`'s unit
+ * tests cover the sum; these cover the wiring the sum feeds.
+ */
+
+import { cleanup, render } from "@solidjs/testing-library";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("@/app/store/rpc-api", () => ({
+    RpcApi: { DockNodeStatusCommand: () => Promise.resolve() },
+}));
+vi.mock("@/app/store/rpc-util", () => ({ TabRpcClient: {} }));
+
+import { __resetAllSlots, dispatch, registerPane } from "@/app/store/agent-document-store";
+import { AgentRow } from "./swarm-view";
+import type { AgentTreeNode, SwarmViewModel } from "./swarm-model";
+import type { DocumentNode } from "@/app/view/agent/types";
+
+afterEach(() => {
+    cleanup();
+    __resetAllSlots();
+});
+
+const BLOCK = "b1";
+
+/** A pane whose only activity is a bare sleep — promoted immediately. */
+function seedSleepingPane() {
+    registerPane(BLOCK, () => {});
+    const node: DocumentNode = {
+        type: "tool",
+        id: "t1",
+        tool: "Bash",
+        toolName: "Bash",
+        status: "running",
+        timestamp: Date.now(),
+        params: { command: "sleep 300" },
+        collapsed: false,
+        summary: "",
+    } as DocumentNode;
+    dispatch(BLOCK, { type: "StreamFlush", newNodes: [node], updatedNodes: [] }, "system");
+}
+
+function treeNode(): AgentTreeNode {
+    return {
+        blockId: BLOCK,
+        agentName: "AgentX",
+        agentProvider: "claude",
+        activitySummary: null,
+        contextTokens: null,
+        agentStatus: "running",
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

Agent2-asaf:
Fixed in `41c8b694d`. Both of you found the same P1 independently, which is about as strong a signal as it gets — and it was worse than a miscount: it made the feature invisible in precisely its primary case.

## [reagent P1 / codex P1] Long-running rows didn't count toward the expand affordance

`totalRows`/`hasChildren` summed only the four model-owned buckets. Because I computed the long-running rows *inside* `LongRunningBucket`, an agent whose only activity was a promoted Bash/sleep call reported `hasChildren() === false` → no chevron, no collapsed count, collapsed by default → the children container's `Show when={!collapsed()}` meant the bucket never mounted. The one thing this PR exists to surface was the one thing it couldn't show.

Both reviews prescribed the same fix and it's the right one: computation hoisted to `AgentGroup` and fed into the sum. The bucket now takes rows as a prop.

One addition beyond the fix. This failure mode is nastier than an ordinary off-by-one — a bucket missing from the sum reads as *"this agent has no children"* rather than *"wrong number"*, so it fails silently and invisibly. I've extracted the sum as `agentChildRowCount` and enumerated every bucket in tests, one per case, so a future sixth bucket omitted from it fails a test instead of shipping invisible.

## [reagent P2] Bucket label colour

Added — but deliberately **not** a fifth saturated hue, and I'd rather flag the reasoning than have it read as an oversight.

The other four labels mark declared, addressable objects: a dispatch, a workflow, a shell, a cron. This bucket is *inferred* from a tool call still running past a threshold. `--secondary-text-color` carries that distinction while still reading apart at a glance from blue/amber/green/purple. Minting a fifth theme token — as `--info-color` was minted for cron, per that rule's own comment — would assert a semantic peer relationship this bucket doesn't have.

Happy to switch to a dedicated token if you'd rather the five read as equals.

`npx vitest run frontend/app/view` — **1942 passed**. `tsc` clean. SCSS unchanged at 16 pre-existing `color-no-hex` errors, same as `main`.

Agent2-asaf:
Follow-up in `bf9adfca1` — closing the verification gap I'd flagged, on the specific thing that broke.

The P1 wasn't a wrong number, it was **the bucket not existing on screen**. `agentChildRowCount`'s unit tests cover the sum, but nothing covered the chain that sum feeds — sum → `hasChildren` → chevron → collapse gate → bucket — which is exactly where the bug lived. A green suite with that chain untested is how it shipped in the first place.

Four tests now drive the real `AgentRow` against a real seeded document store:

- the **collapsed count** appears — i.e. the work is discoverable *without* expanding, which is the actual fleet-view claim
- the bucket and row render once expanded, with the right title
- the `~Ns left` countdown renders for a whole-command sleep
- nothing renders for an idle agent

I chose this over a live screenshot on purpose: the rendered output is a function of the node and store state, so it re-asserts on every run rather than being observed once, and it fails precisely if the regression returns. `AgentRow` is exported for it — same test-only precedent as `DispatchActivityFeedEntry` in the sibling render test.

**1946 view tests passing**, tsc clean.

That retires the "not verified in a live pane" caveat for the part that mattered. What's still unobserved is only cosmetic: how the grey label and row spacing actually look next to the four coloured buckets. Worth a glance whenever someone has Swarm open, but it isn't load-bearing.

PR #2865, opened 2026-08-31
fix(agent-picker): finish the snapshot-preview data-gap fixes from the 08-24 audit

AgentX-asaf:
## Summary

User-reported symptom: a My Agents row still shows "(no conversation
snapshot)" when it shouldn't. Traced to `docs/reports/REPORT_AGENT_PICKER_FIELD_ORDER_SORT_AND_DATA_GAPS_AUDIT_2026_08_24.md`
§5 — the report's own status line claimed "§5/§5a shipped ... PR #2786",
but that's only true for §5a (the Haiku on-demand summary fallback). Both
§5 bullets were never actually implemented:

- `filestore.stat()`'s `Err` (I/O error, lock contention, DB read failure)
  collapsed into the same `_ =>` arm as a genuine `Ok(None)` — a transient
  storage error rendered identically to "never had a snapshot," silently.
- A cross-channel row (synthetic instance, no local `block_id`) also read
  as "no conversation snapshot," when its real history may exist in a
  different channel entirely.

## Changes

- `session.rs`: splits the `stat()` match into `Ok(Some)` / `Ok(None)` /
  `Err(e)`. The error case logs a warning, marks a new `"snapshot_stat"`
  degraded source (same mechanism §4's `identity_links` fix already uses),
  and sets a new `RecentSessionRow.snapshot_check_failed` field.
- `MyAgentsList.tsx`: new `noSnapshotText()` picks between three distinct
  messages instead of one generic string — `"(couldn't check for
  history)"` (stat() errored), `"(history may exist in another
  version)"` (cross-channel, `block_id_hint` empty — exact wording the
  report proposed), or the original `"(no conversation snapshot)"` (a
  real, confirmed empty state).
- `FileStore` gained a `conn()` accessor (mirrors `Store::conn()`) so a
  test can force the `stat()` error deterministically via `DROP TABLE` —
  same technique the existing `identity_links` failure test already uses.
- Corrected the report's own status line/§5/§7 to reflect what's actually
  shipped now.

## Test plan

- [x] `cargo test -p agentmux-srv` — 2871 passed (+2 new)
- [x] `npx vitest run` — 3439 passed (+2 new)
- [x] New tests written first, confirmed failing before the fix (backend:
      compile error; frontend: 2 assertion failures) per TDD

### DIFF
--- a/.changesets/1788205309-fix-agent-picker-distinguish-snapshot-check-failures-and-cro-8csk.md
+++ b/.changesets/1788205309-fix-agent-picker-distinguish-snapshot-check-failures-and-cro-8csk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-picker): distinguish snapshot-check failures and cross-channel rows from a genuinely empty conversation
--- a/agentmux-srv/src/backend/rpc_types/instance.rs
+++ b/agentmux-srv/src/backend/rpc_types/instance.rs
@@ -174,6 +174,17 @@ pub struct RecentSessionRow {
     /// block. False when the snapshot doesn't exist yet (no preview)
     /// or the block_id_hint was empty.
     pub has_snapshot: bool,
+    /// True when `has_snapshot` is false because the filestore lookup
+    /// itself ERRORED (I/O error, lock contention, DB read failure) —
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2866, opened 2026-08-31
docs(flicker): systemic analysis of the flash class + render-architecture spec

AgentY-asaf:
## Summary

The tab-close flash took five rounds to fix (`SPEC_TAB_CLOSE_BUTTON_SELECT_FLASH_2026_08_25.md` §§2-9). §§2-7 each closed a **real** ordering hole — click bubbling, a double round trip, unbatched response application, unbatched WS-push application — and not one of them moved the symptom. §8's *optimistic removal* fixed it by making the tab strip stop depending on backend update ordering at all.

That outcome is worth recording, because a sweep of `docs/` turns up **~40 documents across 11 families** describing the same class (tab strip, pane/block mount, activity dock, browser-pane indicators, startup/splash, terminal streaming, menus, modals, transparency/compositor, scroll-pin, working-row).

## New docs

**`docs/reports/REPORT_TAB_FLASH_SYSTEMIC_ANALYSIS_2026_08_31.md`**
Why suppressors keep failing on this class, the four unsynchronized render pipelines behind it, and — most usefully — a **scorecard** treating #2818 as a controlled test of the report's own predictions:

- ✅ Confirmed: the flash was architectural; only making the bad state unrepresentable closes it.
- ✅ Confirmed and fixed: the reveal gate was keyed on `tid === tabId()`, one reactive step behind what it gates (repaired by §9's targeted gate).
- ❌ **Refuted:** my leading hypothesis (the native pane compositor's `sendClip` → rAF → async HTTP seam) was *not* the cause. Real seam, documented twice elsewhere as a root cause, but I ranked an inferred hypothesis above a defect I had already verified by reading code.

It also records the finding that outranks any code defect here: **none of §§5-7 was ever tested against a build containing all of them** — some "still broken" reports were of incomplete builds.

**`docs/specs/SPEC_TAB_WINDOW_RENDER_ARCHITECTURE_2026_08_31.md`**
The tab/window render architecture as built, plus four candidate abstractions (`WorkspaceEpoch`, transport collapse, `PaneSurfaceSync`, `LayoutReadiness`). Its §0 deliberately **argues down its own scope**: this spec proposed a global epoch/transaction system; #2818 got the same guarantee for one surface in ~150 lines and one file. Prefer local decoupling; fund the epoch work only for surfaces that genuinely cannot predict their own outcome. It also reconciles with the already-committed direction in `SPEC_REDUCER_SSOT_CONSOLIDATION_2026_06_22.md` and `SPEC_STRONG_REDUCER_AUTHORITY_LAYOUT_2026_06_30.md` rather than starting fresh.

## Docs cleanup

Accuracy fixes surfaced while writing the above:

- **`SPEC_TAB_CLOSE_BUTTON_SELECT_FLASH_2026_08_25.md`** — status claimed PR #2811 was open and unmerged (a stale line I wrote pre-merge). Both #2811 and #2818 have landed; marked RESOLVED with what actually fixed it.
- **`SPEC_TAB_CONTENT_REVEAL_GATE.md`** — was `Proposed (implemented)` and still described the original one-global-boolean design. Two later specs changed its behaviour (leaf-scoped gate #2761; targeted gate #2818 §9). Amended with both, and flags the timer-based lift (80ms/800ms) as the remaining weak point.
- **8 dangling source references** in `navigation.rs`, `backend.rs`, `state/mod.rs`, `bootstrap.ts` pointed at `docs/specs/SPEC_LINUX_STARTUP_PAINT_GATING_2026_07_13.md`, which was **never committed** (a gap the 07-14 report had already flagged). Redirected to `REPORT_NEW_WINDOW_STARTUP_COLOR_FLASH_2026_07_14.md`, which holds the diagnosis.

**Left alone deliberately:** the browser-pane loading-indicator spec (Draft) and working-row tool-burst spec (proposed) — verified against git, no fix has shipped for either, so both statuses are accurate.

## Takeaway the docs propose keeping

> If a proposed fix is a gate, a debounce, a `batch()`, a delay constant, or an emission reorder, it is a **suppressor**. Suppressors are legitimate only when the intermediate state is genuinely unavoidable. First ask: can this surface stop depending on the ordering altogether?

## Test plan

Docs + comment-only changes; no runtime behaviour touched.

- `npx tsc --noEmit` — clean
- `cargo check -p agentmux-cef` — clean (the two `unused doc comment` warnings are pre-existing, in files this PR doesn't touch)

### DIFF
--- a/.changesets/1788205456-docs-flicker-systemic-analysis-of-the-flash-class-render-arc-9r7f.md
+++ b/.changesets/1788205456-docs-flicker-systemic-analysis-of-the-flash-class-render-arc-9r7f.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(flicker): systemic analysis of the flash class + render-architecture spec, fix stale statuses and dangling spec refs
--- a/agentmux-cef/src/client/navigation.rs
+++ b/agentmux-cef/src/client/navigation.rs
@@ -14,7 +14,7 @@ use std::sync::Arc;
 #[cfg(target_os = "linux")]
 use crate::state::AppState;
 
-/// Linux startup white-flash fix (docs/specs/SPEC_LINUX_STARTUP_PAINT_GATING_2026_07_13.md).
+/// Linux startup white-flash fix (docs/specs/REPORT_NEW_WINDOW_STARTUP_COLOR_FLASH_2026_07_14.md).
 ///
 /// Bound on how long the real window + native splash can stay hidden/up
 /// waiting for the frontend's first-paint signal before we show anyway.
@@ -54,7 +54,7 @@ static PAINT_GATE_NEXT_EPOCH: std::sync::atomic::AtomicU64 = std::sync::atomic::
 /// still the one this specific timeout was scheduled for; if `on_load_end`
 /// re-armed the gate since (reload/retry mid-startup), this stale timeout is
 /// a no-op instead of revealing the window ahead of the *current*
-/// navigation's real paint (docs/specs/SPEC_LINUX_STARTUP_PAINT_GATING_2026_07_13.md,
+/// navigation's real paint (docs/specs/REPORT_NEW_WINDOW_STARTUP_COLOR_FLASH_2026_07_14.md,
 /// reagent PR #2151 review).
 ///
 /// Must run on the CEF UI thread.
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #2867, opened 2026-08-31
fix(container): container agents never launch with --input-format stream-json

Agent2-asaf:
## Container agents have never once started

Confirmed on this machine: across **every srv log ever written**, the only container-related line is the startup Docker probe. No exec, no create, ever.

Docker was never the problem — daemon healthy, image present locally, probe succeeds every boot. Reproduced live with @a5af driving the real UI (2026-08-31): the container starts, the exec launches, and the CLI **inside** it dies on its first line of stdin in under a second.

```
19:36:20.647  restarted stopped container   agentmux-b64b63c1-…
19:36:20.647  container agent turn: bollard exec
19:36:21.439  container exec stderr:
              Error parsing streaming input line: # Session Context:
              SyntaxError: JSON Parse error: Unrecognized token '#'
19:36:21.446  container exec output EOF
```

The block's own meta shows why:

```
agentMode  = 'container'
controller = 'persistent'                          ← wrong
cmd:args   = ['--input-format', 'stream-json', …]  ← fatal
```

## The mismatch

`--input-format stream-json` tells the CLI that **every stdin line is a JSON envelope**. That's true for the **persistent** controller, which owns a long-lived stdin and writes real envelopes.

A container agent runs one `docker exec` per turn, and `container_spawn.rs` writes the raw message text (`format!("{}\n", message)`). So the CLI meets the startup markdown as its first line and dies.

**The container agent was getting the persistent controller's arguments with the subprocess controller's stdin behaviour.** Those can never work together.

## Root cause — one missing condition, in one of two copies of the same rule

`agent-model.ts` derived `isPersistent` from `provider.controllerType` alone, ignoring agent type — so a container agent on Claude got `persistentLaunchArgs`.

srv's own equivalent (`agent_open.rs:267`) **already applies the container override before deriving `is_persistent`**, and is correct. The same rule lived in two places and only one was right.

srv also overrides the *controller* at resync — the log even says *"container agent has persistent controller in meta — overriding to subprocess"* — but never re-derives `cmd:args`, so the bad argv survived the override that was supposed to fix exactly this.

## The fix

**Root cause** — extracted to `launch-args.ts` so the rule has one home, with `agentMode` computed once and reused for both `isPersistent` **and** the meta write. The mode a block advertises and the args it launches with can no longer disagree.

**Self-heal** — `cmd:args` lives in block meta, so every pane created before this fix would keep failing *forever*, with no route back short of recreating the agent. `strip_stream_json_input_format` drops the flag at the point of use in the container branch, where a block that never re-runs `resync_controller` can't bypass it.

## Found en route, deliberately not fixed here

`agent.open`'s CLI-presence gate (`agent_open.rs:238`) looks in `~/.agentmux/<version>/cli/`, but the CLI actually lives in `~/.agentmux/instances/v<version>/cli/`. The RPC open path therefore returns `CLI_NOT_AVAILABLE` for **host** agents too — I hit it twice this session and worked around it with a junction.

Unrelated to containers, and it *masked* this bug during investigation — I initially mis-diagnosed it as the container root cause until I noticed the identical error had appeared for a `host` agent earlier. Worth its own change.

## Testing

**9 frontend** — container never gets `--input-format`, host still does, `--output-format` preserved for both, returns a fresh array (callers push `provider_flags` onto it).

**5 srv** — strips the flag *and its value* from the real broken argv captured from the live block, leaves `--output-format` alone, no-ops on clean argv, tolerates a malformed trailing flag, removes repeated occurrences.

`npx vitest run frontend/app/view` — **1955 passed**. `cargo test -p agentmux-srv` — **2874 passed**. `tsc` clean.

**Not yet verified end-to-end.** The diagnosis came from a live reproduction and the fix is unit-tested, but I haven't re-run a container agent through the UI with these changes applied — that needs a fresh `task dev`. Worth doing before merge given this path has never worked.

### DIFF
--- a/.changesets/1788205718-fix-container-container-agents-never-launch-with-input-forma-0g2y.md
+++ b/.changesets/1788205718-fix-container-container-agents-never-launch-with-input-forma-0g2y.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(container): container agents never launch with --input-format stream-json
--- a/agentmux-srv/src/server/agent_handlers/input.rs
+++ b/agentmux-srv/src/server/agent_handlers/input.rs
@@ -15,6 +15,44 @@ use crate::backend::blockcontroller;
 
 use super::super::AppState;
 
+/// Drop `--input-format <value>` from a container agent's argv.
+///
+/// `--input-format stream-json` tells the CLI that every stdin line is a JSON
+/// envelope. That is true for the PERSISTENT controller, which owns a
+/// long-lived stdin and writes real envelopes — and false for a container
+/// agent, which runs one `docker exec` per turn and whose stdin is written by
+/// `container_spawn.rs` as the raw message text (`format!("{}\n", message)`).
+///
+/// Mixing the two is fatal rather than untidy: the first line the CLI reads is
+/// the startup markdown, and it dies with
+/// `Error parsing streaming input line: # Session Context: JSON Parse error:
+/// Unrecognized token '#'`, EOF'ing the exec in under a second. That is the
+/// whole reason container agents have never started (verified live,
+/// 2026-08-31).
+///
+/// The root cause is fixed at the source in `agent-model.ts`, which no longer
+/// picks `persistentLaunchArgs` for a container agent. This is the self-heal
+/// for blocks ALREADY persisted with the bad argv: `cmd:args` lives in block
+/// meta, so without this, every pane created before that fix keeps failing
+/// forever with no way back short of recreating the agent. Applied at the
+/// point of use so it cannot be bypassed by a block that never re-runs
+/// `resync_controller`.
+///
+/// Removes the flag and its value; tolerates a trailing `--input-format` with
+/// no value rather than panicking on a malformed argv.
+fn strip_stream_json_input_format(args: Vec<String>) -> Vec<String> {
+    let mut out = Vec::with_capacity(args.len());
+    let mut it = args.into_iter();
+    while let Some(a) = it.next() {
+        if a == "--input-format" {
+            let _ = it.next(); // discard its value
+            continue;
+        }
+        out.push(a);
+    }
+    out
+}
+
 /// Per-agent git commit identity env vars for a spawned agent process
 /// (2026-08-22, docs/retro/retro-shared-git-identity-committer-misattribution-2026-08-22.md).
 ///
@@ -524,7 +562,7 @@ pub fn register_agent_input_handlers(engine: &Arc<WshRpcEngine>, state: &AppStat
                             &block.meta, "agent:container_command", "claude",
                         );
                         let mut base_cmd = vec![container_command];
-                        base_cmd.extend(cli_args);
+                        base_cmd.extend(strip_stream_json_input_format(cli_args));
 
                         let config = blockcontroller::subprocess::SubprocessSpawnConfig {
                             cli_command: String::new(), // unused by spawn_container_turn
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

Agent2-asaf:
All three findings were real and are fixed in 5120483. Thanks — the reagent one would have broken the live verification I'm about to run.

**reagent P1 — `runtime-apply.ts` (third unmigrated copy).** Correct, and this was the worst of the three: the launch-time fix would have held only until the user's first `/model`, `/effort` or `/mode` change, at which point `buildRuntimeArgs(provider.persistentLaunchArgs, …)` wrote `--input-format stream-json` straight back into persisted `cmd:args` and the agent died on its next turn — with the fix apparently in place. `applyRuntimeChange` now takes `agentMode` and goes through `isPersistentLaunch`/`selectLaunchArgs`; both callers (`components/AgentRuntimeDropup.tsx`, `commands/global/runtime.ts`) pass `block.meta["agentMode"]`. Container agents also skip the `forcerestart` entirely now — `input.rs` re-reads `cmd:args` per turn, so there is nothing to restart.

**codex P1 — `CLAUDE_CODE_EXIT_AFTER_STOP_DELAY`.** Also correct: gated on `provider.controllerType !== "persistent"` rather than the effective launch shape, so a container Claude was the one case that got neither the persistent stdin nor the exit-delay env var. Now gated on `!isPersistent`. I changed only the `launchAgentDefinition` copy; the `launchAgent` copy at ~267 has no container path to reach it.

**codex P1 — partial argv repair.** Right, and this was the one I got lazy on. Stripping the single fatal flag left a stale block launching without `-p` and still carrying `--permission-prompt-tool stdio`, which a per-turn `docker exec` has no control channel to answer. `strip_stream_json_input_format` is replaced by `container_argv(stale, provider_id)`, which rebuilds from the provider's own `launch_args` and carries the user's `--model`/`--effort` across so a runtime change isn't silently reset. `strip_flag_with_value` survives as the fallback when the provider id isn't in the catalog — there's no baseline to rebuild from there, so removing only the fatal flag is the honest option.

Tests: the five old `strip_stream_json_input_format` tests are replaced by five for `container_argv`, built on the real broken argv captured from the live Moras block. srv suite green (9/9 in that module), frontend 1578/1578, `tsc --noEmit` clean.

Note the green CI on this PR is from 639c316 — pre-fix. Re-running now.

Still un-fixed and out of scope here: the `CLI_NOT_AVAILABLE` path-convention bug in `agent_open.rs:238` (`~/.agentmux/<version>/cli/` vs the actual `~/.agentmux/instances/v<version>/cli/`). It is not container-specific — it hit a `host` agent in the same session — and it was masking this bug. Tracked separately.

Agent2-asaf:
Right again, and this one was worse than the bug it replaced. Fixed in ec2b5b3.

The thing I missed is exactly what you led with: `container_argv` is on the **per-turn** path, not a one-time migration. So rebuilding from `provider.launch_args` didn't just heal stale blocks — it re-flattened the argv on every single turn, deleting `agent.provider_flags` and `--fork-session` from blocks that this PR's own frontend change had already assembled correctly. I'd have shipped a permanent regression for container agents in the same PR that claimed to consolidate the rule. Confirmed `resolveForkSessionArgs` is gated on `providerId === "claude"` only (`agent-model.ts:456`), with no `agentMode` check, so container agents genuinely reach it.

**Subtract, don't rebuild.** `container_argv` now:
1. Diffs the provider catalog's `persistent_launch_args` against `launch_args` to get the persistent-only flag set — not a hardcoded list, so a flag moving between the two lists can't silently strand this.
2. Returns the argv **untouched** if it carries none of them. That's the common case now: the frontend built it, hands off.
3. Otherwise strips just those flags (with their values), restores any one-shot flag that's missing (`-p` above all), and leaves every other token where it was — `--model`/`--effort`, `provider_flags`, `--fork-session` alike.

Arity comes from `flags_with_arity`, read off the catalog list itself (a flag whose next token isn't another flag takes a value), which is what lets it drop `--permission-prompt-tool stdio` as two tokens and `--verbose` as one without a per-provider table.

Two new tests target your finding directly: `leaves_an_already_correct_one_shot_argv_completely_untouched` (an argv with `provider_flags` + `--fork-session` must come back byte-identical) and `a_heal_preserves_provider_flags_and_fork_session_too` (they survive even when a heal does fire). 12/12 in the module.

A dev instance is up on this branch for live container-agent verification.

Agent2-asaf:
Confirmed and fixed in ca193b3. Good catch — a fourth copy, and the worst-placed one.

`useAgentCommands.ts:1267` runs before **every** send, so it rewrote `--input-format stream-json` into `cmd:args` on every turn regardless of what launch time chose. The three copies fixed earlier are all launch-or-config-change paths; this one re-broke the pane continuously. It now uses `selectLaunchArgs(prov, agentMode)` with `opts.block()?.meta?.["agentMode"]`, same as the other three.

Four tests pin the send path directly rather than the helper: `--input-format` absent and `-p` present for a container agent, `--input-format` still written for a host agent, and a block with no `agentMode` meta treated as host. The host-mode assertion is deliberate — it proves the harness actually reaches this code, so the container assertion isn't vacuously passing.

**On the `provider_flags` / `--fork-session` half, I'm not doing what you suggested, and want to be explicit about why rather than quietly skip it.**

You're right that both are discarded here. But that is not something this PR introduced, and it is not container-specific: this rebuild has always derived `baseArgs` from the provider catalog, so `agent.provider_flags` is dropped on the first send for **host** agents too, and has been for as long as this code has existed. Reproducing it needs only a host agent with `provider_flags` set.

More importantly, "preserve the existing `cmd:args` instead" is not obviously the correct fix, and I'd rather not guess at it inside a fix for a different bug. `buildRuntimeArgs` is idempotent (it strips `--model`/`--effort`/permission flags from its base before re-adding them), so feeding it the live `cmd:args` would work mechanically. The problem is `--fork-session` specifically: it is a *launch-time, one-shot* intent from `resolveForkSessionArgs`. Persisting it into `cmd:args` for every subsequent turn means every turn after the fork also carries it, alongside the `--resume <sid>` srv adds — i.e. plausibly forking the session on each turn instead of once. Preserving it could be worse than dropping it, and deciding that needs a look at the fork/resume interaction, not a one-line change here.

So: this is a real, pre-existing, provider-wide bug that deserves its own PR with its own reasoning about fork semantics. I'll open an issue for it and reference this thread. If you think it belongs in this PR anyway, say so and I'll take it — but I'd want to fix it deliberately rather than fold it in.

Also still outstanding and separate: the `CLI_NOT_AVAILABLE` path bug in `agent_open.rs:238` noted earlier.

PR #2868, opened 2026-08-31
docs(agent-pane): spec a single content-resize contract for the agent pane

Agent1-asaf:
## Summary

Docs-only. Proposes the structural fix `ANALYSIS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_17.md` §7 called for and deferred, and records why three plausible narrow fixes are dead ends.

Nine documented passes have now been made at one bug class — "content in the agent pane changes height and the scroll position visibly jumps". Most produced a confident conclusion a later pass had to withdraw (08-05 disproved its own leading hypothesis; 08-17 retracted two claims mid-document; 08-21 and 08-22 were each corrected twice on review). The problem isn't any individual analysis — it's that six independent mechanisms each decide on their own how to react to a height change, and nothing can observe the interaction end to end.

**Proposes:** one module (`resize-contract.ts`) owning "a subtree is about to change height", migrated one call site per PR, **gated on component-level instrumentation first** — the 08-17 doc's own recommendation #1, still unactioned, and the reason every conclusion in the 08-21/08-22 findings was granularity-limited.

**Rules out, with reasons** (§3 — the most reusable part; these keep getting rediscovered):

- **Shrink-aware eased `scrollToTrueBottom()`** — physically impossible. The browser clamps `scrollTop` synchronously in the same layout pass that produces the shrink, before any effect or RO callback runs. Nothing is left to animate.
- **FLIP on `ToolBlock`'s summary row** — `.agent-tool-summary` is `white-space: nowrap; overflow: hidden`; its height cannot change. The live-tail ↔ result-pill swap is horizontal only.
- **The `heightStale` FLIP bypass for `denied`/`canceled` tools** — mechanism newly traced (a `MutationObserver` on the `--hidden` *class*, which flips instantly, vs. the CSS `content-visibility 120ms allow-discrete`, which flips at the *end* of the collapse). Real, but fixing it buys nothing visible: that panel is already easing to `max-height: 0` over the same 120ms. Documented, not fixed — it disappears for free if the unified contract subsumes it.

**Closes the 251–252px lead** from `FINDINGS_..._LIVE_INSTANCE_DATA_2026_08_22.md` §5: no such height constant exists anywhere under `frontend/app/view/agent/`. Consistent with that doc's own caveat that a repeated *net* delta between two pin-checks can't establish a single element.

## Test plan

- [x] Docs-only — no code, no runtime surface. All six mechanisms in §1's table re-verified present against `216c593c4` before writing (file:line cited for each).
- [x] `heightStale` and 251px conclusions traced from the current source/CSS directly, not from the prior docs.

## Note

§6 is explicit that this does **not** claim the refactor fixes the reported symptom — it removes the structural reason the symptom keeps returning with a different face. Step 1 (instrumentation) exists because nobody has yet confirmed which mechanism dominates in a real session.

### DIFF
--- a/.changesets/1788205838-docs-agent-pane-spec-a-single-content-resize-contract-close--nttn.md
+++ b/.changesets/1788205838-docs-agent-pane-spec-a-single-content-resize-contract-close--nttn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(agent-pane): spec a single content-resize contract; close three ruled-out fix leads
--- a/docs/specs/SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31.md
+++ b/docs/specs/SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31.md
@@ -22,7 +22,7 @@ agent pane changes height and the scroll position visibly jumps":
 | 08-17 | `ANALYSIS_TOOL_CALL_SCROLL_OSCILLATION` | Retracted two of its own claims mid-document; shipped diagnostics only |
 | 08-21 | `FINDINGS_TOOL_CALL_SCROLL_OSCILLATION` | Corrected twice by PR review |
 | 08-22 | `FINDINGS_..._LIVE_INSTANCE_DATA` | Corrected twice by PR review |
-| 08-31 | this doc | Three further candidate fixes ruled out — see §3 |
+| 08-31 | this doc | Two further candidate fixes ruled out (§3); a third ruled out in draft, then withdrawn on review (§3a) |
 
 Every pass was made by someone reasoning carefully from the source, and most
 produced at least one confident conclusion that a later pass had to withdraw.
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

Agent1-asaf:
Both P2s accepted and pushed in `ba5b43fd3`.

**1. `heightStale` ruling withdrawn — you're right, and the error is instructive.** I claimed the fix buys nothing visible because the panel is "already eased" by its 120ms `max-height` collapse. `max-height` is a *constraint*, not a height: while the interpolated maximum is still above the new intrinsic height it isn't binding, so the panel drops to the new intrinsic height in one frame and only the *tail* of the collapse is eased. Moved from §3 (ruled out) to a new §3a (open — settle by measurement, not a third static argument).

I kept it on the record as a worked example rather than quietly rewriting it. The document's whole thesis is that careful reasoning about this subsystem keeps yielding confident conclusions that later have to be withdrawn — and it then did exactly that, in the section listing prior instances of the same failure. That's better evidence for the thesis than anything I argued in §0.

**2. Spinner migration call site corrected.** Right that `createSpinnerCollapser` is a synchronous DOM-independent transform with no element or mutation callback to hand `withHeightContinuity`. Step 5 now names `ChunkList`'s memo in `ToolOverlayLog.tsx` as the site, and states explicitly why the utility itself is the wrong place (couples a pure function to the DOM) and why skipping the distinction is also wrong (leaves the height change unmitigated).

Agent1-asaf:
P2 accepted — accurate, including the part about the test plan. Fixed in `23c88c57f`.

Cause: I read `AgentDocumentVirtualList.tsx` while on a branch 11 commits behind `main`, then checked out `main` and wrote the spec without re-reading that file. The three ranges were stale by the intervening diff and landed in the preceding comment blocks. My test plan's claim that all six were "re-verified... file:line cited for each" against `216c593c4` was false for those three.

Re-derived every row from the current worktree (grep a unique line inside each mechanism, then read its actual boundaries):

| # | was | now |
|---|---|---|
| 1 pin effect | 494-513 | **521-544** |
| 2 RO #1 viewport | 535-555 | **566-588** |
| 3 RO #2 content | 586-595 | **619-631** |
| 4 ToolOverlayLog FLIP | 246-297 | **246-298** |
| 6 MarkdownBlock | 62-78 | **62-81** |

Row 5 and `output-cap.ts:369-440` were already correct — those files weren't touched by the intervening commits, which is exactly why the error was invisible on a spot-check.

Worth noting this is the second review-caught error in a document whose thesis is that this subsystem defeats careful static reasoning. Different failure mode from the §3a one — that was faulty reasoning about correct data, this was correct reasoning about stale data — but it points the same direction, and reinforces why §5 gates the refactor on instrumentation rather than on more reading.

PR #2869, opened 2026-08-31
docs(spec): define the epoch completeness contract, drop stale compositor guidance

AgentY-asaf:
## Summary

Follow-up to #2866, addressing both of Codex's P2 findings (posted just after that PR merged). Both were correct.

### 1. The epoch staging algorithm was unimplementable as written

`{ epoch, updates: [...] }` carries neither an expected update count nor an end-of-epoch marker — so a client genuinely cannot distinguish a partial epoch from a complete one. It would have to either commit early (violating the spec's own F1 invariant) or wait out the timeout on *every* epoch. Codex caught a real hole.

**Resolved by fiat rather than by adding a marker:**

> **One epoch is delivered in exactly one frame, and that frame is always complete.**

That turns it into a producer-side requirement — §3.2's fetch-then-emit restructuring is what makes it satisfiable — and deletes the staging buffer, its bounded timeout, and its stall risk outright. If a future transition genuinely can't fit one frame, the contract must be reopened deliberately with an explicit `{ epoch, part, final }` marker, **not** papered over with a timeout.

This also closes the spec's own open question 2, which had already suspected the staging buffer was dead weight — Codex sharpened a hole I'd half-noticed into one I had to actually resolve.

### 2. Stale guidance pointed at a disproven investigation

§7's open question 1 still asked whether §3.3 (the native compositor seam) is the residual cause, and Phase 0/Phase 1 still instructed an implementer to go confirm it — directly contradicting §0, which records that PR #2818 fixed the symptom **without touching that pipeline**. Left alone, it would have sent someone down a path this very document proves is a dead end.

- Question 1 struck through and reframed around a *future, independently observed* overlay symptom, with §3.3's real precedent cited as prior art rather than as evidence.
- Phases 1-3 are now explicitly **unordered** — the tab-close case is closed, so there's no live symptom to sequence them against; ordering should come from Phase 0's trace of whatever comes next.
- Phase 0's deliverable now says to **rank hypotheses on the captured trace, not on narrative fit** — which is precisely the error the companion report's §0 scorecard records me making.

## Test plan

Docs-only, no runtime surface. Consistency re-checked: no remaining references to the removed staging buffer or to §3.3 as the residual cause.

### DIFF
--- a/docs/specs/SPEC_TAB_WINDOW_RENDER_ARCHITECTURE_2026_08_31.md
+++ b/docs/specs/SPEC_TAB_WINDOW_RENDER_ARCHITECTURE_2026_08_31.md
@@ -150,23 +150,42 @@ transition. Give that transition an identity and carry it to the client.
 - Backend: every reducer dispatch stamps a monotonic `epoch: u64` on **all**
   `WaveObjUpdate`s it produces. Same transition → same epoch.
 - Wire: updates travel as `{ epoch, updates: [...] }`, never as bare objects.
-- Frontend: `wos.ts` gains an epoch-aware apply:
-  - Updates from epoch **N+1** apply as one `batch()`.
-  - An update whose epoch is **older** than the cell's current epoch is
-    dropped (this is today's version guard, generalized from per-object to
-    per-transition).
-  - A **partial** epoch — some objects arrived, some haven't — is held in a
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2870, opened 2026-08-31
fix(agent): widen the transient-failure auto-retry ladder, add jitter

AgentY-asaf:
## Where we actually are

Retry for 429 / 529 / network **is implemented**, at two independent layers, and both work:

1. **The CLI's own backoff.** Claude CLI retries a 429 internally and emits `rate_limit_event` with `retry_after_ms`. We translate that to `provider_waiting` and use the hint to size the stall watchdog, so a legitimate backoff isn't misread as a hang. (The fix from `retro-busy-animation-stuck-on-429-2026-06-24.md` — shipped.) No AgentMux retry happens here, correctly: the CLI is alive and retrying.
2. **Our post-failure auto-retry.** When the CLI gives up, `agents/failure.rs` classifies the exit — `RateLimited` (429), `Overloaded` (529), `Network` are `retryable: true` — and `useAgentFailure` arms a countdown that re-sends the last user message.

So the plumbing was fine. **The weakness was policy.**

## The problem

`AUTO_RETRY_BACKOFF_S` was `[5, 10]` — two attempts, **~15 seconds** of coverage, then manual-only. Anthropic overload episodes routinely outlast that. The common case was: classifier correctly says "retryable" → UI correctly offers auto-retry → budget correctly exhausts → **user hand-clicks *Retry now* against a failure the system already knew was transient.**

## Changes

- **Ladder → `[5, 15, 30, 60, 120]`** — five attempts, ~3.9 minutes. Still bounded on purpose: a turn still failing after ~4 minutes is more likely a sustained outage or an account-level limit than a blip, and a human should decide rather than have the app hammer the API indefinitely. The existing episode/cap semantics (budget resets only on genuine turn success or a genuinely fresh user message) are untouched.
- **±20% jitter** via `jitteredBackoffSeconds`, floored at 1s so the countdown never starts at zero. AgentMux drives many agents concurrently — Fleet broadcast, cron sweeps, loops — so a provider-wide 529 fails them together, and unjittered rungs had every one retry on the *same second*, plausibly re-triggering the overload they were backing off from. Pure with an injectable rand, so the ladder stays deterministic under test.

Tests now derive expectations from a single `LADDER_MS` constant rather than hardcoding counts, so the next policy change updates one place.

## Gap this does NOT fix (documented, not hidden)

**Auto-retry is pane-scoped.** `useAgentFailure` is a Solid hook mounted in exactly one place (`agent-view.tsx:1815`). If no agent pane is rendered, *nothing retries* — a turn driven by cron, a loop, a Fleet broadcast, or an MCP send to an unrendered pane will classify the 429 correctly, persist it correctly, surface it correctly when a human next opens the pane, and never retry.

Worth being precise: this is **not** "the tab is backgrounded" — inactive workspace tabs stay mounted (`workspace.tsx` hides them with `display:none`), so those panes do retry. The exposure is genuinely unrendered panes.

Fixing it means moving the retry decision behind the pane into the persistent controller — which already *has* the classification (`classify_exit_line`) and today explicitly ignores it, its own doc comment reading *"just without auto-retry."* That's a real design change (who owns retry, how it reconciles with the pane's budget, what a headless agent does at cap) and wants its own spec rather than being smuggled in here.

Two smaller gaps also recorded in the report: the ladder ignores server-provided `Retry-After` hints (no field on `AgentFailure` today), and `SPEC_AGENT_FAILURE_RECOVERY_UI_2026_06_16.md` is still marked Draft though largely implemented.

## Test plan

- `useAgentFailure.test.ts` — 8/8, covering the full ladder, the cap, the jitter band, and all three budget-reset paths
- `frontend/app/view/agent` — 1572/1572 across 120 files
- `tsc --noEmit` — clean

**Not verified by observation:** I did not reproduce a live 429/529 against the provider. This is a policy constant plus a pure helper, both unit-tested, but end-to-end behaviour under a real sustained overload is unobserved.

### DIFF
--- a/.changesets/1788206573-fix-agent-widen-the-transient-failure-auto-retry-ladder-to-4-nazl.md
+++ b/.changesets/1788206573-fix-agent-widen-the-transient-failure-auto-retry-ladder-to-4-nazl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): widen the transient-failure auto-retry ladder to ~4min with jitter
--- a/docs/reports/REPORT_TRANSIENT_API_FAILURE_RETRY_STATE_2026_08_31.md
+++ b/docs/reports/REPORT_TRANSIENT_API_FAILURE_RETRY_STATE_2026_08_31.md
@@ -66,6 +66,27 @@ The budget is deliberately capped: an auto-retry stays inside the same
 resets only on a genuine turn success or a genuinely fresh user message.
 That design is sound and is preserved.
 
+## 3.1 This was already diagnosed, a month earlier
+
+Both gaps fixed below were identified in
+`REPORT_LOGIN_PERSIST_FAILURE_AND_STUCK_WORKING_2026_07_27.md`'s "Gaps vs.
+best practice" section on **2026-07-27**, verbatim:
+
+> 1. **Not exponential.** 5s → 10s is a fixed doubling for exactly two steps,
+>    not real exponential growth — a THIRD consecutive 429 has no automatic
+>    retry left at all and drops straight to manual, even though 429 bursts
+>    are frequently short and self-clearing over more than two attempts.
+> 2. **No jitter.** `armAutoRetry` fires at exactly T+5s / T+10s,
+>    deterministically. Multiple agent panes sharing one account/quota (a real
+>    scenario in this app — several agents, one Anthropic subscription) would
+>    all retry in lockstep and collide on the SAME rate limit again.
+
+The analysis was correct and nothing acted on it for five weeks. Worth noting
+as a process observation, not just a code one: the finding was filed inside a
+report about a *different* headline bug (login persistence + stuck working
+state), so it never became a tracked item of its own. A correct diagnosis
+buried in a larger document is, in practice, close to no diagnosis at all.
+
 ## 4. What changed in this PR
 
 ### 4.1 The ladder was too short (fixed)
--- a/docs/reports/REPORT_TRANSIENT_API_FAILURE_RETRY_STATE_2026_08_31.md
+++ b/docs/reports/REPORT_TRANSIENT_API_FAILURE_RETRY_STATE_2026_08_31.md
@@ -1,0 +1,151 @@
+# Transient API failure (429 / 529 / network) retry — where we are
+
+**Date:** 2026-08-31
+**Author:** AgentY
+**Status:** Assessment. One gap fixed in the same PR (§4.1); the rest recorded,
+not fixed.
+**Scope:** `agentmux-srv/src/agents/failure.rs`,
+`frontend/app/view/agent/hooks/useAgentFailure.ts`,
+`frontend/app/view/agent/failure/failure-accessory.ts`,
+`frontend/app/view/agent/providers/claude-translator.ts`,
+`frontend/app/store/agent-pane-state/reducer.ts`
+**Related:** `SPEC_AGENT_FAILURE_RECOVERY_UI_2026_06_16.md` (the design; still
+marked Draft/proposed though largely implemented),
+`SPEC_AGENT_ERROR_FRAMEWORK_2026_06_20.md`,
+`SPEC_PERSISTENT_CONTROLLER_FAILURE_CLASSIFICATION_2026_08_04.md`,
+`docs/retro/retro-busy-animation-stuck-on-429-2026-06-24.md`
+
+---
+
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed the P2 in `15d704cf5` — good catch, and it was a real miss: I updated the *test* file's header for the new ladder but left the source module's doc comment describing the old policy.

Three stale spots, all now fixed:
- module doc: *"the 5s/10s auto-retry countdown/budget"*
- module doc: *"a 5 s countdown ... capped at 2 auto-retries (5 s → 10 s)"*
- an inline comment near the budget-reset effect: *"still capped at 2"*

Rather than restate the new rungs in three places (which is how this drifted in the first place), the doc comments now describe the *shape* of the policy and point at `AUTO_RETRY_BACKOFF_S`, whose own doc comment owns the rationale. Also dropped the "5 s countdown" phrasing since the first rung is jittered now.

Re-verified: `useAgentFailure.test.ts` 8/8, `tsc --noEmit` clean.

AgentY-asaf:
Fixed in `07071b55b`. Fair catch again — and the reason it was missed is worth naming: my sweep after the first note grepped `capped at 2`, which didn't match `its own 2 auto-retries`. Same defect, different phrasing.

So this time I swept for the *shape* of the problem rather than one phrasing (`two|2 auto|of 2|twice|10 s|10s|5 s|5s`) and went through every hit:

- **Fixed** — the budget-restore comment now reads "its own full ladder", with no number left to drift.
- **Fixed** — `SPEC_AGENT_FAILURE_RECOVERY_UI_2026_06_16.md` §6's *Bounded* clause still specified the as-designed `2 auto-retries, 5s → 10s`. It now records the original design *and* the as-shipped ladder with the rationale, and names `AUTO_RETRY_BACKOFF_S` as the source of truth over the prose — so the next policy change doesn't have to chase the doc.
- **Deliberately left** — `AUTO_RETRY_BACKOFF_S`'s own comment says *"Was `[5, 10]` — two attempts, ~15s"*: an intentional historical note explaining why the ladder changed.
- **Deliberately left** — §1's "5-second auto-retry" and §8's "5 s auto-retry": both still accurate (the first rung is 5s), and §8 is a historical record of what phase P1 shipped.

The underlying cause of three rounds of this is that the policy was restated in prose in four places. The fix direction is the one applied above: describe the shape, point at the constant, keep exactly one source of truth.

Re-verified: `useAgentFailure.test.ts` 8/8, `tsc --noEmit` clean.

AgentY-asaf:
Fixed in `99ecca591` — and this time repo-wide, which is what I should have done two rounds ago.

**Why this kept recurring:** each of my sweeps was scoped to the file I'd just edited. Round 1 fixed `useAgentFailure.ts`, round 2 swept `useAgentFailure.ts` + one spec, and you found `agent-view.tsx` — a file I never looked at. The defect wasn't the individual misses, it was searching a scope defined by what I'd touched rather than by where the fact could live.

This sweep covered `frontend/` and `docs/specs/` and turned up your `agent-view.tsx:1800` plus **three more** you hadn't reached:
- `SPEC_AGENT_PANE_UNIFIED_FAILURE_REDUCER` §57 and its proposed-code snippet (`"same budget as today"` — no longer true)
- `SPEC_PERSISTENT_CONTROLLER_FAILURE_CLASSIFICATION` §53 and its Non-goals clause

Every remaining hit is now explicitly dated (*"when this spec was written"*, *"As designed (2026-06)"*, *"Was `[5, 10]`"*) or lives in a historical report/retro, where stating the then-current policy is correct.

**One thing worth surfacing that the sweep turned up:** `REPORT_LOGIN_PERSIST_FAILURE_AND_STUCK_WORKING_2026_07_27.md` already identified **both** gaps this PR fixes, on 2026-07-27:

> 1. **Not exponential.** 5s → 10s is a fixed doubling for exactly two steps ... a THIRD consecutive 429 has no automatic retry left at all
> 2. **No jitter.** ... Multiple agent panes sharing one account/quota ... would all retry in lockstep and collide on the SAME rate limit again

Correct diagnosis, five weeks unactioned — because it was filed inside a report about a *different* headline bug and never became a tracked item. I've credited it in the report and noted the process observation: a correct finding buried in a larger document is, in practice, close to no finding at all.

Re-verified: `useAgentFailure.test.ts` 8/8, `tsc --noEmit` clean.

PR #2874, opened 2026-09-01
Manoz@Area54: fix(agent-pane): drop composer strip's duplicate centered token/elapsed stats

GenericAgentX-asaf:
## Summary

The agent pane's composer strip (above the textarea) showed a centered token/elapsed readout that duplicated `AgentWorkingRow`'s "Worked" row (below the transcript), confusingly:

- **While loading**, both showed the identical live `turnTokens` + elapsed, ticking every second, in two places at once.
- **Once idle**, they diverged silently: the composer strip showed *cumulative* session totals (all turns since launch); the Worked row shows *per-turn* stats for just the last turn — same `↑X ↓Y · Ns` visual shape, no label saying which is which.

`AgentWorkingRow` is now the sole surface for these stats — it already covers both cases, plus `$cost · N turns` the composer strip never showed at all.

## Change

Removes the stats *content* only (`sessionTotals`/`turnTokens` props, `rightText`, `elapsedMs`/`loadStartMs`) from `AgentComposerStrip`, and the two prop wirings from `agent-view.tsx`. Deliberately leaves the row-layout scaffolding (`statsZone`/`statsInline`/`computeStatsInline`/the width-measurement effect) in place — that's this file's most heavily revised, most fragile logic (7 revisions in 2 days per its own module doc comment), it already treats "empty stats zone" as a normal existing case, and touching it would be risk for no functional gain. Full reasoning in the spec.

`turnTokensAtom` state/projection is untouched — it still feeds `AgentWorkingRow`'s own `turnTokens` prop elsewhere in `agent-view.tsx`. `sessionTotalsAtom` becomes fully unused after this (grep-confirmed `AgentComposerStrip` was its only consumer) — flagged as an optional follow-up cleanup in the spec's Non-goals, not bundled into this PR (reducer-level accumulator removal is separate, larger-blast-radius surgery).

See `docs/specs/SPEC_COMPOSER_STRIP_DROP_CENTER_STATS_2026_08_31.md` for the full investigation and design.

## Test plan
- [x] New regression test: `.agent-composer-strip-stats` never renders while loading or idle
- [x] `npx vitest run frontend/app/view/agent/components/AgentComposerStrip.test.tsx` — 52/52 passing (50 existing unaffected + 2 new)
- [x] `npx vitest run frontend/app/view/agent` — 120 files / 1571 tests passing (no regressions from the wiring removal)

### DIFF
--- a/.changesets/1788221656-fix-agent-pane-drop-composer-strip-s-duplicate-centered-toke-7dig.md
+++ b/.changesets/1788221656-fix-agent-pane-drop-composer-strip-s-duplicate-centered-toke-7dig.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): drop composer strip's duplicate centered token/elapsed stats
--- a/docs/specs/SPEC_COMPOSER_STRIP_DROP_CENTER_STATS_2026_08_31.md
+++ b/docs/specs/SPEC_COMPOSER_STRIP_DROP_CENTER_STATS_2026_08_31.md
@@ -1,0 +1,133 @@
+# Spec: Drop the composer strip's centered token/elapsed stats
+
+**Date:** 2026-08-31
+**Status:** Proposed
+**Motivated by:** direct report — the centered stats readout that
+sometimes appears in `AgentComposerStrip` (above the textarea) is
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2875, opened 2026-09-01
docs(spec): fix three unimplementable holes in the epoch design

AgentY-asaf:
## Summary

Codex reviewed #2869 just after it merged and found that the epoch contract **cannot be built as written**. All three findings are correct; this fixes them.

### 1. [P1] The unit must be the *dispatch*, not the event

`dispatch_to_reducer` returns a `Vec<Event>`, `publish_events` sends each element independently, and the bridge consumes one event at a time. But one *command* routinely emits several *events* — `DeleteWorkspace` emits `WorkspaceDeleted` plus one `SrvWindowClosed` per affected window.

So stamping per-event splits a single logical transition across several frames and reopens exactly the intermediate paint the epoch exists to prevent. The producer change isn't "stamp each event" — it's **publish a dispatch envelope** through `publish_events`, with the bridge subscribing to envelopes rather than individual events. Any design that keeps the bridge on individual events cannot satisfy the contract, however the stamping is done.

### 2. [P1] Gaps, not just ordering — the client needs a watermark

These frames are **deltas, not snapshots**. `try_send_lane` drops priority events when a lane fills, and `run_wave_obj_bridge` continues past a `Lagged` without resyncing. So if epoch N changed object A and the client next receives N+1 changing B, applying N+1 paints old A beside new B — a state that was never committed, an F1 violation by the same definition the tab flash was.

Added a **contiguous epoch watermark** plus a **resync arm**: a frame newer than `watermark + 1` must not apply; the client re-reads an authoritative snapshot and re-anchors. Without that arm the design is *strictly worse* than today's per-object version guard under packet loss, because it would confidently apply a delta onto a base it never received.

### 3. [P2] "Fail the transition" was unimplementable where it was written

Service handlers apply reducer events to SQLite and publish them *before* the async bridge performs its reads — so by the time an `emit_fetched` read fails, the mutation is already committed. Fetch-before-emit prevents a *partial* frame but converts that failure into a *missing* epoch behind a committed mutation.

Now names the two admissible resolutions — assemble-before-commit (strongest, puts object reads on the commit path) or heal-by-watermark — and recommends the latter, since the resync arm is non-optional anyway.

## Also

- §3.2 corrected: per-dispatch not per-event; the HTTP response body can't fill a gap either (a body at `watermark + 2` is as unapplicable as the WS frame); and a scope note that resync is deliberately a *fourth* path, admissible only because it carries a snapshot rather than a delta.
- Open questions now carry the concrete follow-ups this raises — including **"is the resync arm worth building at all?"**, which should be answered *before* §3.1 is funded rather than during.

## Net effect

This makes the spec **more expensive and less attractive**, which is the honest outcome. It's now three distinct pieces of work — dispatch envelope, client watermark, authoritative resync — where the earlier draft implied one. §0 already says the tab flash didn't justify funding it; this reinforces that rather than softening it.

A gap-tolerant delta protocol is a real distributed-systems problem, and the spec now says so plainly instead of hand-waving a completeness contract that the transport can't honour.

## Test plan

Docs-only, no runtime surface. Open-question list renumbered after insertion.

### DIFF
--- a/docs/specs/SPEC_TAB_WINDOW_RENDER_ARCHITECTURE_2026_08_31.md
+++ b/docs/specs/SPEC_TAB_WINDOW_RENDER_ARCHITECTURE_2026_08_31.md
@@ -147,45 +147,99 @@ those rather than starting from this document. Note also
 The reducer already produces every multi-object change as one atomic
 transition. Give that transition an identity and carry it to the client.
 
-- Backend: every reducer dispatch stamps a monotonic `epoch: u64` on **all**
-  `WaveObjUpdate`s it produces. Same transition → same epoch.
+- Backend: every reducer **dispatch** stamps a monotonic `epoch: u64` on **all**
+  `WaveObjUpdate`s it produces. Same dispatch → same epoch.
 - Wire: updates travel as `{ epoch, updates: [...] }`, never as bare objects.
 
-**Completeness contract (normative — an earlier draft left this undefined):**
+**The unit is the DISPATCH, not the event.** This is the subtlety an earlier
+draft got wrong. `dispatch_to_reducer` returns a `Vec<Event>`; `publish_events`
+sends each element independently; the bridge consumes one event at a time. One
+*command* routinely produces several *events* — `DeleteWorkspace` emits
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2877, opened 2026-09-01
docs(agent): spec retry for transient failures on turns with no rendered pane

AgentY-asaf:
## The gap

PR #2870 widened the auto-retry ladder, which fixed the *policy* half of the 429/529 problem. This specs the structural half it deliberately left alone:

> **Detecting a transient failure is server-side and works. Acting on it is a SolidJS hook mounted only by `agent-view.tsx:1815`. If no pane is rendered, nothing retries.**

`persistent.rs` already classifies on exit and persists the result — its own doc comment says the banner is surfaced *"just without auto-retry."* That was the right scope call for `SPEC_PERSISTENT_CONTROLLER_FAILURE_CLASSIFICATION`, but it leaves the decision structurally unreachable for headless turns.

### Turns genuinely run without a pane

Cron is the proof: `backend/cron/mod.rs:225` POSTs to `/agentmux/reactive/inject`, which delivers straight to `blockcontroller::get_controller` (`server/reactive.rs:1152`) — a server-side registry, no frontend participation at any point. Same for jekt delivery, `FleetBroadcast`, `Loop`, and an MCP `SendMessage` to an agent whose pane is down.

So a cron-driven turn can hit a 429, be classified correctly, be persisted correctly, show you a banner correctly the next time you open the pane — and never have retried. The scheduled work just didn't happen.

### What is *not* affected

Worth stating because the obvious guess is wrong: **a background tab is fine.** `workspace.tsx` keeps every tab mounted and hides inactive ones with `display:none`, so those panes still retry. The exposure is genuinely unrendered panes — a closed pane, a non-active block inside a block stack, a closed window.

## Why this isn't a one-liner

Most of the spec is the three decisions the naive fix ("just call retry from the controller too") would otherwise make by accident:

1. **Two owners, one budget.** If the controller retries *and* a pane is open, both fire — two re-sends, two independent countdowns. Recommends the **server owning the budget with the pane rendering server state**. Explicitly rejects the tempting alternative of branching on "is a pane mounted": the server can't observe that reliably (a renderer can die without unsubscribing), and making retry depend on it recreates precisely the ordering-race class this repo has already spent ~40 documents on.
2. **What "retry" means without a document.** The pane re-sends the last user message read from its own document. The controller has no document, and whatever it uses must not collide with `persistent_resume`'s separate stale-`--resume` retry.
3. **Cap behaviour with nobody watching.** A pane shows a banner; a headless agent has no one to show. Recommends persist-and-stop plus a notification, with "mark the agent degraded" left to its own design.

## Phase 0 is "quantify", not "build"

Instrument how often a transient failure is classified for a block with no mounted pane first. **If that number is near-zero in practice, the honest answer is to document the limitation and not build this** — the spec says so explicitly. That's the same measure-before-structural-work discipline `SPEC_TAB_WINDOW_RENDER_ARCHITECTURE_2026_08_31.md` §5 arrived at the hard way, applied up front this time.

## Open questions worth flagging

- **Fleet-wide backoff** becomes *possible* once the decision is centralized — the server could see all agents failing on one account and back off globally rather than per-agent. Strictly better than #2870's per-agent jitter, but only reachable after Phase 2.
- **`restart_on_crash`** is carried on the agent definition and consumed by *nothing*. Either dead config to remove, or the natural home for some of this policy — resolve before adding a parallel mechanism next to it.

## Test plan

### DIFF
--- a/.changesets/1788222751-docs-agent-spec-retry-for-transient-failures-on-turns-with-n-8iq3.md
+++ b/.changesets/1788222751-docs-agent-spec-retry-for-transient-failures-on-turns-with-n-8iq3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(agent): spec retry for transient failures on turns with no rendered pane
--- a/docs/specs/SPEC_HEADLESS_TRANSIENT_RETRY_2026_08_31.md
+++ b/docs/specs/SPEC_HEADLESS_TRANSIENT_RETRY_2026_08_31.md
@@ -1,0 +1,170 @@
+# Transient-failure retry for turns with no rendered pane
+
+**Date:** 2026-08-31
+**Status:** Proposal. Not implemented.
+**Owner:** unassigned
+**Scope:** `agentmux-srv/src/backend/blockcontroller/persistent.rs`,
+`agentmux-srv/src/backend/blockcontroller/subprocess/host_spawn.rs`,
+`agentmux-srv/src/agents/failure.rs`,
+`frontend/app/view/agent/hooks/useAgentFailure.ts`,
+`frontend/app/view/agent/agent-view.tsx`
+**Related:** `SPEC_AGENT_FAILURE_RECOVERY_UI_2026_06_16.md` (the retry policy
+and budget semantics this extends),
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2878, opened 2026-09-01
fix(auth): enforce per-channel login — close five paths to Claude without one

AgentX-asaf:
## The report

> how are all my agents able to operate without any auth inside the armory?

and then, decisively:

> we want per-channel isolation. a user needs a login anytime in the channel. the problem is agents are able to login in a channel without a claude auth.

Per-channel isolation is a **wanted feature**. Four separate mechanisms were defeating it. Full investigation: `docs/analysis/ANALYSIS_PER_CHANNEL_AUTH_BYPASSES_2026_08_31.md`.

## The four bypasses

| # | Path | Silent? | Status |
|---|---|---|---|
| 1 | `resolve_account()` global fallback | silent, spawn-time | **closed** |
| 2 | accounts per-channel vs links global | structural | out of scope (see below) |
| 3 | "Use existing login" | explicit click | **removed** |
| 4 | `checkcliauth` first-run bootstrap | **fully silent** | **removed** |

**#4 is the likely direct answer to the report.** During a routine *auth check* — no login, no user action, no Armory account — it copied `~/.claude/.credentials.json` into the isolated `CLAUDE_CONFIG_DIR`, gated only on a sentinel file. It also explains why the symptom varied by machine (whether `~/.claude` held a valid credential). Its justifying retro (2026-06-05) predates the "agents never use `~/.claude`" requirement and was never revisited against it.

**#3** read the user's **personal** `~/.claude` and copied it — refresh token included — into the agent's dir. Its `INV-R` guard only ever constrained the *destination*; nothing constrained the source. That put it in direct tension with `SPEC_BLOCK_AMBIENT_HOME_DIR_IDENTITY_BINDING_2026_08_25`: that spec blocks **binding** a `~/.claude` dir, while this copied the credential **out of** it into a compliant dir — same end state, passing the check, because the check tested a path rather than the credential's provenance.

## Claude's terminal login is now unified with every other provider

Claude was the sole provider whose config-dir env var was **stripped** for terminal login, so the login landed in global `~/.claude` and was copied back. Every other oauth provider already pointed the var **at** the isolated dir and detected completion via `pollForCliAuthReady`. The only stated reason for the special case was that "only Claude has a seed-from-global capability to fall back on" — a convenience, not a constraint.

`pollForCliAuthReady` now checks **before** its first sleep, removing a guaranteed 5s of dead latency on the fast path.

## Deliberate behavior reversals — not regressions

These tests asserted the *opposite* before, and are inverted on operator instruction:

- Two `inject.rs` tests now assert the spawn **refuses** an account existing only in the global mirror (they enforced the fallback for reagentx's P0 on #2632).
- A tier-3 test now asserts the config-dir var is **kept**, not stripped.
- **Binding continuity across a channel switch is given up.** Re-login per channel is the accepted cost.

I also removed three `useAgentControllerStatus` describes covering `useGlobalLogin`'s `loginWaiting` semantics — the function is gone, and the third's premise (TWO independent in-flight guards) no longer exists. The counter invariants stay covered by the relogin/useTerminalInstead suites; I left an inline note for whoever adds a second concurrent recovery flow.

`"seeded"` is removed from `ProviderLoginOutcome` — unreachable once tier 2 is gone, so leaving it would be a lie in the type.

## Not in scope

The `disposable_test`/Store B half of `SPEC_IDENTITY_STORE_SPLIT_2026_08_17` (#2). **Issue #2627 was closed with 5 of its 6 steps unbuilt** — which is exactly why accounts stayed per-channel while links went global, the mismatch #1 was papering over. That deserves its own reopened issue rather than riding along here.

## Expected behavior change (deliberate)

Opening an agent in a **fresh channel with no login in that channel** now fails with an actionable auth card instead of silently inheriting a credential. **This will be visible on every new local/dev/portable build.**

## Test plan

- [x] `cargo test -p agentmux-srv` — 2879 passed
- [x] `npx vitest run` — 3444 passed
- [x] `cargo build -p agentmux-cef` — 0 errors


a5af:
@codex review

AgentX-asaf:
All four findings fixed in `5cf65bf46`. **Codex's P1 was a genuine miss on my part, and the most important comment on this PR.**

## Codex P1 — a fifth bypass

Correct, and I should have found it: `checkcliauth`'s self-heal-from-global reads the **same** `~/.claude` source as the first-run bootstrap I *did* remove, in the **same file**, just on a different trigger. Removing only one left the bypass fully open — and this one is arguably worse, because it isn't gated by the first-run sentinel, so it could fire on **any** failed validation, repeatedly. A channel could report `authenticated: true` having never been logged into.

I deleted the machinery outright rather than leaving it unused: `claude_access_token`, `refresh_claude_dir_from_global_if_stale`, `refresh_dir_from_global`, and the `selfheal_tests` module. A helper whose entire job is copying the operator's personal credential into an isolated dir is exactly the thing that gets quietly rewired back in later; now there is nothing to call.

**Note the srv test count drops 2879 → 2875.** That is 4 tests removed *because the code under them is gone*, not coverage regressing. I considered a replacement regression test asserting the auth check never imports from global, but it would have to mutate `HOME`/`USERPROFILE` — process-global state that this codebase already guards with module-local locks elsewhere precisely because it races. Deleting the callable surface is the stronger guarantee here; I'd rather say that plainly than add a brittle test and call it covered.

The real condition it papered over (the "Pozl 401" — a global re-login rotating the access token and invalidating the isolated copy's refresh token) now surfaces honestly: failed validation → actionable auth card → log in for this channel.

## Codex P2 — dead-end recovery notices

Also correct, and a bad failure mode: telling a stuck user to click a button that no longer exists. Fixed at all three sites (`terminal-timeout` on both the relogin and `loginViaTerminal` paths, plus Claude's `terminal-unavailable` hint) — they now point at "Login Again" / Armory → Accounts. I dropped the Claude-specific *"if you're already signed in elsewhere"* hint entirely: being signed in elsewhere is precisely what can no longer satisfy this channel, by design.

## reagent P2 ×2

Both real leftovers from my removal — the dead `LogFn` import, and a `vi.mock` for a module the hook no longer imports. Fixed.

## One thing neither of you flagged

Sweeping for related debris, I found **two orphaned doc blocks** the removals left behind: the `useGlobalLogin` member doc (now misattributing to `loginViaTerminal`) and the `canSeed` prop doc in `PreLaunchAuthPanel`. Same defect class reagent caught on #2856 earlier this session — I evidently still don't check for it by reflex when deleting a member. Both removed, along with stale `useGlobalLogin()` / "Use existing login" prose across `agent-view`, `login.ts`, `commands/types.ts`, `useAgentCommands`, and `auth-state`.

The analysis doc is updated to five bypasses, with #5 explicitly credited to Codex and marked as missed in my original pass.

**Tests:** agentmux-srv 2875 passed; frontend 3444 passed; `tsc --noEmit` clean; `cargo build -p agentmux-cef` 0 errors.

AgentX-asaf:
All three fixed in `3088242c6` — comment-only, no behavior change.

- **`register-seeded-account.ts:21`** — the module header still described the deleted `registerSeededAccount` and called the deleted `seedGlobalLogin` "unchanged". Rewritten to describe what the module actually is now (mint dir + persist account), with an explicit note that the middle "seed from `~/.claude`" step is gone and that the `seeded` naming survives only for continuity with the RPC names.
- **`bind-to-agent-menu.ts:131`** — referenced `useAgentControllerStatus.useGlobalLogin()` as the thing whose live-apply pair it mirrors.
- **`useAgentCommands.ts:78`** — six present-tense references treating `useGlobalLogin()` as a still-live recovery path.

**On that third one, I want to be straight about what happened:** those exact lines appeared in my own sweep output last round. I applied the fix to `agent-view` / `login.ts` / `commands/types.ts` and skipped this file. Reading the grep and not acting on all of it is the real mistake — and "I already swept for that" is precisely what stopped me re-checking. That's the same shape as the orphaned-doc-block miss earlier in this PR, and the misattributed JSDoc reagent caught me on in #2856.

While in there I also rewrote `useAgentControllerStatus.test.ts`'s file header (it explained the suite's entire purpose in terms of the removed function) and converted the Codex-P1-on-#2338 note in `useAgentCommands.ts` into an explicit historical one — the gap it describes is now structurally impossible rather than a convention every future recovery path has to remember.

Verified zero remaining live references: every surviving mention of `useGlobalLogin` / `seedGlobalLogin` / `seed-from-global` is now inside an explicit "removed 2026-08-31" note.

**Tests:** agentmux-srv 2875 passed; frontend 3444 passed; `tsc --noEmit` clean.

AgentX-asaf:
Both fixed in `2f877dbfc`.

- **`cli_handlers.rs:416`** — correct, and a real compiler warning I introduced: removing the self-heal block left all four bindings never reassigned. `mut` dropped from each. The only `unused_mut` left in the crate is pre-existing in `process_tracker/windows.rs`, a file this branch never touches.
- **`run-provider-login.ts:580`** — correct; there is one poll here now, not two.

**On your observation that this file was skipped by the previous purge commit — that's the useful part of the review, and worth naming precisely.** The root cause is that my sweep grepped `seedGlobalLogin` but not its sibling `pollForGlobalLoginSeed`. The pattern silently under-matched, so the file genuinely didn't appear in my results. That's the same incomplete-sweep failure as the round before, one level down: last time I read the output and didn't act on all of it; this time I wrote a query that couldn't have found it.

So this round I swept on the **full list of removed symbols** rather than the ones I happened to remember. That turned up three more nobody had flagged:

- **`identity_handlers.rs:49`** named the deleted `seed_provider_auth_from_global` as the example caller justifying why `identity.ensureaccountdir` exists at all — rewritten around the caller that actually still needs it (the terminal-login tier, which must know where to point a login before opening a terminal).
- **`agentmux-cef/src/commands/cli_login.rs:1269`** documented `open_login_terminal` as polling via `seed_provider_auth_from_global` — describing a completion mechanism that no longer exists. Now describes the real one.
- Two `run-provider-login` **test names** still contrasted against `seedGlobalLogin` ("not seedGlobalLogin", "used to be called unconditionally"), which reads as a live comparison to something deleted.

Every surviving mention of a removed symbol across `frontend/`, `agentmux-srv/`, and `agentmux-cef/` is now inside an explicit "removed 2026-08-31" note. I verified that by grepping the complete symbol list, not a sample.

**Tests:** agentmux-srv 2875 passed; frontend 3444 passed; `tsc --noEmit` clean; `cargo build -p agentmux-cef` 0 errors.

AgentX-asaf:
Fixed in `f3fcaff9b`, plus nine more of the same class that your finding led me to.

`markSeeded()`'s doc was exactly as you describe. I kept the name and marked it **HISTORICAL** rather than renaming — a rename would churn the reducer's action union (`Seeded`) and its tests for no behavioral gain, and the comment now states plainly that nothing is seeded from anywhere.

**Why this one escaped the last sweep, specifically:** my previous grep matched the symbol `seedGlobalLogin` and the phrase `seed-from-global`. This comment said **`seed-global-login`** — the hyphenated *file* name. A third spelling neither pattern covered. So the sweep didn't under-match by carelessness this time; it under-matched because I was still searching for **identifiers** when the stale references are **prose**.

Re-sweeping on conceptual phrases instead found nine more the symbol greps structurally could not:

- `rpc-api/identity.ts` justified `identity.ensureaccountdir`'s existence by the removed seed path — now justified by the terminal-login tier that actually still needs it
- `auth-state.ts`'s `Seeded` action doc — same treatment as `markSeeded()`
- `identities.rs` ×2 — "every seed-from-global / terminal login's persist call" in the `serde(default)` rationale; that bug was universal, not seed-specific, so the qualifier was actively misleading
- `auth-state.test.ts`'s describe name
- `login.ts`, `useAgentControllerStatus.ts` — "tier-2/3 success" → tier-3
- `agent-view.tsx` — "(seed-from-global / terminal login)" → "(relogin / terminal login)"
- `run-provider-login.test.ts` header + three test names still framed around the deleted tier

Plus the **tier numbering prose** across `run-provider-login.ts`, `useAgentControllerStatus.ts` and two test files: "tier 2 or 3", "tier 2/3", "straight to tier 2", "trying tier 2 (fast)". I kept the numbering at 1/3 deliberately so existing spec and review references stay valid — but every sentence implying a *live* tier 2 was simply wrong. Unrelated "Tier 2" usages (MuxBus delivery tiers, widget-bar tiers, window-title tiers) are untouched.

Comment and test-name only; no behavior change. **Tests:** agentmux-srv 2875 passed; frontend 3444 passed; `tsc --noEmit` clean.

That's five rounds of stale-reference findings on this PR. The pattern across all of them is the same and worth stating once: **I kept verifying deletions by searching for what I deleted, when the residue is prose describing it.** Symbol greps were never going to converge here.

AgentX-asaf:
Addressed in `7cb4eeb61` — and these two were the most substantive notes of the review rounds, because they name a real behavioral divergence **this PR created**, not just stale wording.

You're right on both counts. Both comments claimed `resolve_account`'s global-mirror fallback kept the display paths "consistent with the spawn path". That stopped being true the moment `inject.rs` switched to `resolve_account_for_spawn`. Concretely, today: an oauth-class account resolvable only via the mirror is **listed** by `identity.self.accounts` and can return **`valid: true`** from `identity.account.validate`, while a real spawn in this channel refuses it with `MissingCredentials`. (An api-key-class one is silently skipped at spawn rather than blocking it — a third behaviour, worth knowing.)

**I did not change the display behaviour, and I want to be explicit that this is a judgement call rather than an oversight:**

- A listing should describe what exists. Dropping the fallback makes the account **vanish from Armory**, which reproduces the "my account disappeared after a version bump" confusion that reagentx's P0 on #2632 was written to fix. Swapping a misleading-but-informative row for no row at all isn't obviously an improvement.
- `validate`'s answer is correct on its own terms — it reports whether the *credential* is live at the provider, not whether a spawn would accept it. Its callers use it to check a key, not to predict a launch.

**What's genuinely missing** is that neither payload tells the caller the account can't satisfy a spawn in *this* channel. The honest fix is to surface that — a `usable_in_channel` flag plus an Armory "needs login in this channel" state — not to hide the row and not to widen the spawn gate back out. That needs a payload field, a UI affordance, and a decision about how it interacts with the unbuilt `disposable_test`/Store B split (same area, §1 of the analysis), so I recorded it as **§6 of the analysis doc** rather than bolting it on here.

**Interim mitigation so it isn't invisible:** `identity.self.accounts` now emits a `tracing::warn!` whenever it lists an account resolved only via the mirror (`Arc::ptr_eq` against `id_store`), visible under `muxlog auth`.

Both comments rewritten to state exactly what is and isn't guaranteed — in particular that `valid` must not be read as "this agent can launch".

**Tests:** agentmux-srv 2875 passed; frontend 3444 passed.

AgentX-asaf:
Fixed in `8627c1796`. Good catch, and a slightly embarrassing one: **a log line whose entire purpose is making a silent divergence visible would itself have been unreadable.**

Cause: I authored that Rust string through a Python heredoc and never read the emitted bytes back. The intended line-continuation didn't survive the round-trip, leaving the source indentation inside the literal. Rewritten as a single line, and this time verified with `cat -A` against the actual file rather than trusting the generator.

I also swept every added line on this branch for the same defect (a run of 3+ spaces inside a string literal). This was the only occurrence — confirmed post-commit, so the check ran against what's actually on the branch rather than my uncommitted tree.

**Tests:** agentmux-srv 2875 passed.

PR #2879, opened 2026-09-01
docs(spec): codex round-3 corrections — epoch resync + container retry scope

AgentY-asaf:
Four findings from Codex's post-merge reviews of #2875 and #2877. All correct.

## `SPEC_TAB_WINDOW_RENDER_ARCHITECTURE` — the resync arm was underspecified three ways

The watermark + resync arm I added in round 2 turns out to have three independent holes:

- **[P1] Reordering is not loss.** `dispatch_to_reducer` releases the reducer mutex *before* caller I/O and publication, so two concurrent requests can allocate `E` and `E+1` and publish `E+1` first. Treating every `> watermark + 1` frame as loss makes ordinary concurrent RPCs trigger authoritative reloads — potentially in a loop. Either the producer serializes publication by epoch, or the client needs a reorder window. The producer-side total order is preferable to a heuristic timeout.
- **[P1] The client can't know what a lost epoch touched.** Seeing only `N+1`, it has no record of which objects `N` changed — so a scoped resync can silently *omit* the stale object and then advance the watermark past `N`, permanently blessing a torn cache. Cross-workspace dispatches rule out a single-workspace reload too. Needs a retained per-epoch mutation manifest or a whole-renderer snapshot; nothing narrower is sound.
- **[P1] A resync races the live stream.** Apply `E+1`, then receive a snapshot captured at `E` — unconditional adoption rolls the cache *backwards*, and if nothing later arrives the renderer stays stale indefinitely. Snapshot-epoch adoption must be **monotonic**: discard and retry anything behind the current watermark.

## `SPEC_HEADLESS_TRANSIENT_RETRY` — there are three spawn paths, not two

- **[P2]** Container agents (`agentMode=container`) are forced onto the subprocess controller by `blockcontroller/mod.rs` and run each turn through `container_spawn.rs`, which classifies (`:410`), persists (`:446`) and publishes (`:453`) **independently**. My Scope listed only `persistent.rs` and `host_spawn.rs`, so implementing as written would leave container turns with the same zero-retry behaviour the spec claims to fix — and a containerised agent is headless by default, i.e. the *most* exposed case, not an edge one.

This strengthens rather than complicates the existing recommendation: three copies of a retry budget is three chances to diverge, and the divergence would stay invisible until someone's container agent silently stopped retrying. Shared controller-side ownership (§2.1) is the answer.

## The more important change: §0 is strengthened

Three review passes on §3.1, three new classes of correctness hole, no sign of convergence:

| Pass | Found |
|---|---|
| 1 | partial epochs undetectable without a count or terminator |
| 2 | unit is the dispatch not the event; delta frames need gap detection; "fail the transition" unimplementable post-commit |
| 3 | reordering indistinguishable from loss; client can't know what a lost epoch touched; stale resync rolls the watermark back |

**That iteration count is itself the finding.** This is a replicated-state protocol being designed incidentally as a cache refactor. §3.1/§3.2 are now explicitly marked **not implementable as specified** and should be read as *"why this is harder than it looks"* rather than a plan of record — if ever funded, it needs a dedicated design with an explicit consistency model, not a WaveObject-cache refactor.

Worth restating the contrast: the tab flash that motivated this entire design was closed in **~150 lines** by decoupling one surface from backend ordering. §3.3 and §3.4 are unaffected and remain independently shippable.

## Test plan

### DIFF
--- a/.changesets/1788223210-docs-spec-codex-round-3-corrections-to-the-epoch-resync-desi-wsmo.md
+++ b/.changesets/1788223210-docs-spec-codex-round-3-corrections-to-the-epoch-resync-desi-wsmo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(spec): codex round-3 corrections to the epoch resync design + container scope
--- a/docs/specs/SPEC_HEADLESS_TRANSIENT_RETRY_2026_08_31.md
+++ b/docs/specs/SPEC_HEADLESS_TRANSIENT_RETRY_2026_08_31.md
@@ -5,6 +5,7 @@
 **Owner:** unassigned
 **Scope:** `agentmux-srv/src/backend/blockcontroller/persistent.rs`,
 `agentmux-srv/src/backend/blockcontroller/subprocess/host_spawn.rs`,
+`agentmux-srv/src/backend/blockcontroller/subprocess/container_spawn.rs`,
 `agentmux-srv/src/agents/failure.rs`,
 `frontend/app/view/agent/hooks/useAgentFailure.ts`,
 `frontend/app/view/agent/agent-view.tsx`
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

Issue #2871, opened 2026-08-31
Browser pane: camera (getUserMedia video) access is unconditionally denied


AgentX-asaf:
## Summary

Browser panes cannot grant camera access to any page, by design — not an
OS-permission gap, a bug, or a TODO. Confirmed while checking camera-access
status on the Browser tab:

- `permission_handler()` (`agentmux-cef/src/client/handlers.rs:74`)
  explicitly returns `None` for any browser pane (`self.is_browser_pane`),
  falling through to CEF's default Alloy-runtime behavior, which denies
  media access outright.
- Even on the main app client, where a `PermissionHandler` IS installed
  (for mic/voice input, #1602), it only ever grants the audio-capture bits
  (`CEF_MEDIA_PERMISSION_DEVICE_AUDIO_CAPTURE` /
  `_DESKTOP_AUDIO_CAPTURE`) — never video. Per the CEF contract,
  `allowed_permissions` must match `required_permissions` exactly for a
  combined request, so a page asking for `{audio: true, video: true}` gets
  the *whole* request denied, not just the video half stripped. The inline
  comment at `handlers.rs:669` is explicit: "we never want camera."

Net effect: any site loaded in a Browser pane that calls
`getUserMedia({video: true})` gets `NotAllowedError` unconditionally, on
every page, every time.

## Why this might be worth revisiting

The mic precedent (`SPEC_VOICE_INPUT_PER_PANE_2026_05_19.md` §Phase 4,
#1591/#1602) already solved the harder half of this problem for audio:
per-pane origin-scoped grants, plus OS-level permission (Windows Privacy /
macOS TCC) handled separately in the frontend via `getUserMedia` error
classification + a settings deep-link. Camera is the same shape of
problem, just for video — video conferencing sites, QR/barcode scanners,
or any other camera-dependent page a user might load in a Browser pane
currently have no path to work at all, regardless of OS permission state.

## Open design questions (not yet investigated)

- Should this be a per-pane, per-origin opt-in (mirroring the mic model),
  or a blanket pane-level toggle?
- Where does the grant live — Settings, a per-tab permission prompt, an
  Armory-level setting?
- Any different threat-model considerations vs. mic — camera capture is
  arguably a more sensitive default-deny than audio, so an explicit,
  visible user opt-in (not a silent grant) seems like the right shape,
  but this hasn't been scoped.
- Desktop capture (screen sharing) is already partially allowed via the
[...]

PR #2880, opened 2026-09-01
docs(browser-pane): spec camera (getUserMedia video) access (#2871)

AgentY-asaf:
Closes the scope-and-design request in #2871 (AgentX-asaf), which explicitly asked for a spec rather than an implementation.

## Correcting the issue's central premise

#2871 says the mic work *"already solved the harder half of this problem for audio: per-pane origin-scoped grants."*

**It did not.** There is no origin scoping and no grant store anywhere in the codebase. The handler captures `requesting_origin` **only to log it** (`tracing::info!(target: "voice", %origin, …)`) and then grants the audio bits unconditionally for any origin. What Phase 4a delivered is *handler plumbing* — a `wrap_permission_handler!` block wired to the main client — not a policy layer.

That changes the scoping materially: the grant model, its storage, its prompt UI, and its per-pane threading all have to be built from scratch. The genuinely reusable part is roughly twenty lines of CEF wrapper.

## The constraint that shapes everything

Per the CEF contract already quoted in `handlers.rs:669`, `allowed_permissions` must **match** `required_permissions` — granting a subset is a denial of the whole request, not a partial grant.

So **audio and video cannot be decided independently for one request.** A page asking `{audio: true, video: true}` gets an all-or-nothing answer. Any permission UI presenting them as separate toggles would be lying to the user.

The decision unit is therefore the whole request, and grants are keyed by `(pane, origin, requested-bits)` — a later subset request reuses the grant, a superset re-prompts.

## Work the issue didn't identify

The browser-pane client is constructed as `AgentMuxClient::new(handler, /* is_browser_pane */ true)` — a **boolean**. The handler knows it's *a* browser pane but not *which* one, so it cannot look up a per-pane grant. `block_id` has to be threaded through three construction sites (`creation.rs:164`, `creation_views.rs:124`, `client/mod.rs`) before any of this works.

## Phase 0 is blocking and cheap

Verify that CEF permits retaining the `MediaAccessCallback` across the handler's return. **The entire prompt-based design depends on it.** If CEF requires a synchronous decision, the design changes shape completely — the options collapse to a pre-navigation allowlist or an in-page interstitial. A throwaway build settles it in an afternoon; nothing else should start first.

I've deliberately *not* asserted that `DEVICE_VIDEO_CAPTURE` is `1 << 1` (open question 2). The existing code hardcodes `1 << 0` and `1 << 2` with an ABI-stability comment, and the video bits almost certainly follow the pattern — but "almost certainly" isn't good enough for a permission bitmask, so it's flagged for confirmation against the CEF headers.

## Threat model

Camera is treated as strictly more conservative than the existing audio grant, for two reasons: passive video capture is more revealing than audio, and — more importantly — the existing grant is on the **first-party app client**, whereas this grants to **arbitrary browsed pages**. That's a categorically different trust boundary, and it's the exact reason `permission_handler()` returns `None` for panes today.

Non-negotiable for v1: no silent grants, visible-while-live indication, one-click revoke that actually stops the stream, and deny as both the default and the failure mode.

`DESKTOP_VIDEO_CAPTURE` (screen sharing) is explicitly **out of scope** and must not ride along just because the bitmask is adjacent — it has a different consent model (the user picks a surface as part of consenting).

## One thing worth your judgement

Open question 5: #2871 motivates this generically (video conferencing, QR scanners) rather than from a specific blocked workflow. Given the threat-model cost, the spec asks for one concrete use case before Phase 2 is funded. Phases 0 and 1 are cheap enough to do regardless.

## Test plan

### DIFF
--- a/.changesets/1788223483-docs-browser-pane-spec-camera-getusermedia-video-access-per--jl5i.md
+++ b/.changesets/1788223483-docs-browser-pane-spec-camera-getusermedia-video-access-per--jl5i.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(browser-pane): spec camera (getUserMedia video) access, per issue #2871
--- a/docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md
+++ b/docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md
@@ -1,0 +1,236 @@
+# Browser pane — camera (getUserMedia video) access
+
+**Date:** 2026-09-01
+**Status:** Proposal. Not implemented.
+**Owner:** unassigned
+**Issue:** #2871 (AgentX-asaf) — "Browser pane: camera (getUserMedia video)
+access is unconditionally denied"
+**Scope:** `agentmux-cef/src/client/handlers.rs` (permission handler +
+`permission_handler()` getter), `agentmux-cef/src/client/mod.rs`
+(`new_with_browser_pane`), `agentmux-cef/src/browser_pane/creation.rs`,
+`agentmux-cef/src/browser_pane/creation_views.rs`, a new grant store in
+`agentmux-srv`, and a prompt surface in the frontend.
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2881, opened 2026-09-01
docs(analysis): confirm floater ctrl+wheel is consumed by CEF, not stripped

AgentA-asaf:
Closes out the investigation from #2857 / #2861 with the measurement that identifies the cause.

## The measurement

A `wheel` recorder installed over CDP in each window, driven by the repo owner's **real mouse** — which is what §6 could not do, since `Input.dispatchMouseEvent` enters below the OS layer:

| window | input | wheel events reaching the DOM |
|---|---|---|
| main (docked terminal) | Ctrl+Scroll | **22 / 22, all `ctrl: true`** — zoom works |
| floater (same terminal) | Ctrl+Scroll | **0** |
| floater (same terminal) | plain scroll | **49, all `ctrl: false`** |

The recorder was proven live in the floater by injecting one CDP wheel and watching it land, so the zero is a genuine absence and not a dead probe.

**The floater's renderer receives wheel input normally; Ctrl+Wheel specifically is consumed before it.** A suppressed message, not a stripped modifier. §2 (focus) and §4b (zoom pipeline) are both dead; §4a is confirmed.

## Why floaters differ

Not runtime style — **both** windows are `RuntimeStyle::ALLOY`. The difference is hosting:

- **Main window** — CEF **Views** (`app/mod.rs:1057`)
- **Floater** — a **child HWND** (`floating_pane.rs:261`, `WindowInfo::set_as_child`)

The precedent is already in the tree: **browser panes are also child-HWND browsers** (`creation.rs:184`) and needed a manual `WM_MOUSEWHEEL`/`MK_CONTROL` subclass (`hwnd.rs:403`) for exactly this — its comment describes working around *"CEF's native, HostZoomMap-shared zoom"*. Browser panes got the workaround; **floaters never did.**

## Ruled out by measurement, not argument

`hwnd.rs:403`'s `return 0` sits **outside** its `if let Some(ctx)`, so a subclassed HWND that can't resolve a context swallows Ctrl+Wheel and does nothing — a perfect fit for the symptom. It is not what is happening: the same hook logs `[pane-wndproc] mouse-wheel` for non-ctrl wheels (`hwnd.rs:424`), and the host log has **zero** such entries across the 49 recorded plain scrolls.

Noted separately: that unconditional `return 0` is a latent bug regardless — it silently discards input whenever `find_context` misses.

## The fix (§8, not implemented here)

§5's shape with its premise now measured rather than assumed, keeping both constraints from Codex's review of #2857: hook the **CEF child hierarchy** (the outer popup WndProc never sees a message handled by a Chromium descendant), and **re-apply after every navigation** (Chromium recreates `Chrome_RenderWidgetHostHWND` per load). A floater can't reuse `browser_panes::zoom_in`, so the host must signal the frontend to run its normal `term:zoom` write — the path §6 already proved works end to end. Inherits the Windows-only platform gap documented at `hwnd.rs:395`.

Docs only.

### DIFF
--- a/docs/analysis/ANALYSIS_FLOATER_CTRL_SCROLL_ZOOM_2026_08_31.md
+++ b/docs/analysis/ANALYSIS_FLOATER_CTRL_SCROLL_ZOOM_2026_08_31.md
@@ -268,3 +268,77 @@ buffer scroll when you Ctrl+Scroll?**
 | nothing happens at all | **the wheel message itself** is consumed before the renderer |
 
 These need different fixes, so this should be answered before code is written.
+
+---
+
+## 7. ROOT CAUSE CONFIRMED (2026-08-31, second session)
+
+Measured with a `wheel` recorder installed over CDP in each window, driven by the
+repo owner's **real mouse** (not injected input — that was §6's blind spot):
+
+| window | input | wheel events reaching the DOM |
+|---|---|---|
+| main (docked terminal) | Ctrl+Scroll | **22 / 22, all `ctrl: true`** — zoom works |
+| floater (same terminal) | Ctrl+Scroll | **0** |
+| floater (same terminal) | plain scroll | **49, all `ctrl: false`** |
+
+The recorder was proven live in the floater by injecting one CDP wheel and
+watching it land, so the zero is a real absence, not a broken probe.
+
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2882, opened 2026-09-01
docs(spec): reject the epoch design after a fourth unsoundness finding

AgentY-asaf:
## The finding (Codex round 4, correct)

A retained manifest scoped to *"the objects the missing epoch touched"* is **still unsound**:

> Client is at `N-1`. It loses `N` (which changed A) and receives `N+1` (which changed B). It resyncs A — but the read lands *after* `N+1` committed, so it returns A-at-`N+1`, and the snapshot's epoch is `N+1`. Adopting that watermark marks `N+1` as applied **when B was never applied at all.** The cache is permanently torn on B, with the watermark asserting it's clean.

A scoped resync must therefore read the union of everything changed from the missing epoch *through the snapshot's own epoch*, or read a historical snapshot exactly at `N` — which requires point-in-time reads the store has no reason to support today.

## Why this changes the verdict rather than just the text

That's **four consecutive review passes, four new classes of unsoundness**, every fix looking complete until the next pass — and none of them found by implementation:

| Pass | Found |
|---|---|
| 1 | partial epochs undetectable without a count or terminator |
| 2 | unit is the dispatch not the event; delta frames need gap detection; "fail the transition" unimplementable post-commit |
| 3 | reordering indistinguishable from loss; client can't know what a lost epoch touched; stale resync rolls the watermark back |
| 4 | a manifest scoped to the lost epoch is *still* unsound — must span the interval, or read point-in-time |

The common thread is finally clear enough to name: **the design ships deltas over a lossy, reorderable channel and then tries to bolt exactly-once semantics on afterwards.** A snapshot-shaped design has none of these problems. So does simply keeping today's per-object versioning and accepting that multi-object coherence isn't guaranteed.

So §3.1/§3.2 move from *"not implementable as specified"* to **rejected**. They aren't unfinished — they're the wrong shape. Their remaining value is the enumerated failure list, for whoever proposes something similar later.

Worth restating the contrast one last time: the tab flash that motivated this entire design was closed in **~150 lines** by decoupling one surface from backend ordering.

**§3.3 and §3.4 are unaffected** and remain independently shippable — §3.4 (causal reveal, replacing the 80ms/800ms timers) is still the highest-value item in the document.

## Test plan

### DIFF
--- a/.changesets/1788223581-docs-spec-reject-the-epoch-design-after-a-fourth-unsoundness-mqnf.md
+++ b/.changesets/1788223581-docs-spec-reject-the-epoch-design-after-a-fourth-unsoundness-mqnf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(spec): reject the epoch design after a fourth unsoundness finding
--- a/docs/specs/SPEC_TAB_WINDOW_RENDER_ARCHITECTURE_2026_08_31.md
+++ b/docs/specs/SPEC_TAB_WINDOW_RENDER_ARCHITECTURE_2026_08_31.md
@@ -55,13 +55,17 @@ predict its outcome, and (b) demonstrably tears. §3.4's causal-reveal work is
 the highest-value remaining item; §3.3 is a real seam awaiting its own
 evidence (it was **not** the tab-close cause — see the report's §0).
 
-**Strengthened 2026-09-01, after three review passes on §3.1.** Each pass found
-a new class of correctness hole in the epoch design (see §3.1's cost/risk table),
-with no sign of converging. §3.1/§3.2 should now be read as *"here is why this
-is harder than it looks,"* not as a plan of record. Treat §3.1 as **not
-implementable as specified** and requiring a dedicated replicated-state design
-with an explicit consistency model before anyone writes code against it. §3.3
-and §3.4 are unaffected by this and remain independently shippable.
+**§3.1/§3.2 are REJECTED as of 2026-09-01, after four review passes.** Each pass
+found a new class of unsoundness in the epoch design and each fix looked
+complete until the next pass (see §3.1's cost/risk table). The design is not
+"unfinished" — it is the wrong shape: it ships deltas over a lossy, reorderable
+channel and then tries to bolt exactly-once semantics on afterwards.
+
+Read §3.1/§3.2 as **an enumerated list of failure modes**, valuable for anyone
+who later proposes something similar, and not as a plan of record. If the
+underlying need becomes real, start from a consistency model and the
+transport's actual guarantees. §3.3 and §3.4 are unaffected by this and remain
+independently shippable — §3.4 is still the highest-value item in the document.
 
 ## 1. Problem statement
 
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentY-asaf:
Fixed in the latest push — and the *method* was the problem, so that's worth naming.

My previous commit said it had "swept the whole file". It hadn't: it grepped for phrasings I anticipated (`fund` / `plan of record` / `ship behind a flag` / `shippable`), so it missed every passage expressing the same idea in different words — including "continue work the repo already scoped" and "anyone picking this up". **Third time this session I've made the same mistake** (the retry-ladder prose, the §3.3 stale guidance, now this). This time I read all ~18 `§3.1`/`§3.2` references in context instead of pattern-matching.

That turned up ReAgent's finding plus three more:

- **§3 "Prior committed direction"** (ReAgent's P1) — presented §3.1/§3.2 as continuing the SSOT-consolidation and layout-single-writer specs and told readers to pick it up there. Withdrawn. Those specs' direction is sound and *unaffected*; §3.1/§3.2 simply aren't a valid way to extend it clientward. Kept the one durable observation and re-scoped it: `UpdateObjectMeta` bypassing the event bus is a real hole in the **current** transport, worth closing on its own merits.
- **§4's deletion list** still promised §7's parent-before-child bridge ordering would be "subsumed by one-frame-per-epoch". It won't be — so that ordering and its tests are now **permanent**, not transitional. This is the one that actually mattered: a future reader could have deleted working code expecting a replacement that is never coming.
- **Open questions 2 and 3** still posed epoch-envelope and cross-renderer-epoch design questions as open. Both moot, struck through, retaining only the part that generalises — the persist subscriber and disk writer consume individual events, which constrains any future scheme that wants to group them.

@codex — your P2 cited §3 (134-144), §5 (457-466) and §7 (507-512). §5 and §7 were already fixed in `b15568290`, which landed while your review was in flight; §3 was the genuine remaining one and is now fixed. Net effect is the same as what you asked for.

AgentY-asaf:
Fixed — and you've identified the actual pattern behind three rounds of this.

I kept fixing passages that **reference** §3.1/§3.2 from elsewhere (§0, §3 intro, §4, §5, §7), while the **bodies of §3.1/§3.2 themselves** still spoke in the voice of a live proposal. Every forward-looking sentence inside them is a latent contradiction with the rejection, so fixing them one at a time was always going to leak.

So this is fixed as a class: both sections now open with a banner stating they're rejected, that the body is preserved **as originally written** for its failure-mode value, and that any forward-looking statement inside ("ship", "keep X until Y ships") is **historical, not operative**. A reader landing mid-section can no longer mistake it for a plan regardless of the surrounding prose.

Plus the specific sentence you flagged, because it was actively harmful rather than merely stale: *"Keep the §7 ordering and its tests until the epoch frame ships"* told a reader that **working tests are a stopgap awaiting a replacement that will never arrive.** Now states plainly that the ordering and its tests are permanent.

Verified by reading §3.1 and §3.2 end-to-end for operative-voice statements rather than grepping for phrasings I expected — which is the method that caused the previous misses.

Issue #2872, opened 2026-08-31
provider_flags and --fork-session are discarded from cmd:args on every send (all agents, not just container)


Agent2-asaf:
## What

`sendMessage` in `frontend/app/view/agent/hooks/useAgentCommands.ts` rebuilds `cmd:args` before every turn from the **provider catalog's** base args:

```ts
const baseArgs = selectLaunchArgs(prov, agentMode);
const updatedArgs = buildRuntimeArgs(baseArgs, runtimeConfig, prov.id);
```

`buildRuntimeArgs` re-applies the runtime config (`--model`, `--effort`, permission mode) but knows nothing about anything else the launch path appended. So both of these are dropped on the **first send** and never come back:

- `agent.provider_flags` — a real, user-configurable field, split and appended at `agent-model.ts:446`
- `--fork-session` — appended at `agent-model.ts:457` when `resolveForkSessionArgs` says so

## Scope

This is **not** container-specific and predates PR #2867. `baseArgs` has always come from the catalog, so a **host** agent with `provider_flags` set loses them on its first send too. Repro needs only a host agent with a non-empty `provider_flags`.

Surfaced by ReAgent while reviewing #2867 (https://github.com/agentmuxai/agentmux/pull/2867#pullrequestreview-5070765193), deliberately not fixed there — see the reasoning below.

## Why it isn't a one-liner

The obvious fix is to use the block's live `cmd:args` as the base instead of the catalog's. That works mechanically: `buildRuntimeArgs` strips `--model`/`--effort`/permission flags from its base before re-adding them, so it's idempotent under repeated application.

The problem is `--fork-session` specifically. It encodes a **launch-time, one-shot** intent. Persisting it into `cmd:args` means every subsequent turn also carries it, alongside the `--resume <sid>` srv adds — plausibly forking the session on every turn rather than once. Preserving it may be worse than dropping it.

So a fix needs to decide, per flag, whether it is a durable part of the argv (`provider_flags`: yes) or a one-shot launch intent (`--fork-session`: probably no), rather than blanket-preserving whatever is in `cmd:args`.

## Suggested direction

Separate durable user args from one-shot launch args at the point they're appended, so the per-turn rebuild can reapply the former and correctly drop the latter — rather than having the rebuild guess from the flattened argv.

Agent5-asaf:
[...]

PR #2883, opened 2026-09-01
fix(container): sandbox agent turns hung forever — a container turn can't use stdin

Agent2-asaf:
Two independent bugs, both found while verifying #2867 live. A sandbox agent opened fine, took a message, and then never answered — no output, no error, no exit.

## 1. A container turn can never receive stdin EOF

`spawn_container_turn` wrote the message to the exec's stdin and relied on `drop(input)` to signal EOF. That has never worked:

- bollard's hijacked exec stream can't half-close the write side, and
- on Windows the Docker transport is a **named pipe**, which has no half-close at all.

An explicit `input.shutdown().await` doesn't help either. Both were tried against a live container:

| stdin | result |
|---|---|
| closed (`docker exec -i` + pipe) | turn completes in ~50ms |
| held open (`drop`, or `shutdown().await`) | zero output, still running after 12 minutes |

**This repo already knew.** `container.rs`'s `exec (io)` integration test says so outright — *"bollard's hijacked exec stream does not reliably half-close stdin on `drop(input)`, so an EOF-dependent reader would hang"* — and works around it with `read line` instead of `cat`.

The production path survived because container agents carried `--input-format stream-json`, which is newline-delimited and never waits for EOF. #2867 correctly moved them to the one-shot `-p` argv (the persistent flags crashed them outright), and `claude -p` reads its prompt **to EOF** — which made a latent limitation load-bearing and converted a crash into a silent hang.

### Fix: don't use stdin

The turn message goes as a trailing argv token; the exec is created with `attach_stdin: false`. Docker takes argv as a JSON array, so there is no shell quoting to get wrong and a multi-line `# Session Context` body is a single token.

A trailing `-` (codex's read-prompt-from-stdin positional) is **replaced**, not appended after — it designates the same slot, and appending would leave the CLI waiting on a stdin that can never EOF.

`ContainerManager::exec` now takes `attach_stdin` explicitly, documented as a one-way door. The `!cmd` shell exec and the env integration test pass `false`; the test that specifically covers the stdin contract passes `true`.

## 2. `agent.open` looked for the CLI at a path that never existed

`agent_open.rs` hand-rolled `$HOME/.agentmux/<version>/cli/<provider>` — no `instances/`, no `v` prefix, and reading `$HOME`/`$USERPROFILE` directly instead of `DataPaths` (so it ignored `AGENTMUX_HOME_OVERRIDE` and portable layouts). The canonical path, used by `resolvecli`, `install.start`/`install.check` and the frontend's `resolveCliDir`, is `<agentmux_home>/instances/v<version>/cli/<provider>`.

Result: `agent.open` failed with `CLI_NOT_AVAILABLE: ... Open an agent pane in the UI to trigger installation` for **every** agent, even with the CLI installed — and since that message blames a missing install, it masked the real caller behaviour. Verified live: the binary was at `instances/v0.55.29/cli/claude/...` while this reported it missing at `.agentmux/0.55.29/cli/claude/...`. I hit this trying to script a repro of bug 1.

## Verification

Live, through `agent.open` + `agent.send` against a real container, not just unit tests:

- **Before:** turn stuck `running`; in-container `claude` alive 12+ min with zero output.
- **After:** `agent.send` acks in 20ms, turn completes in ~2s with a captured `session_id`, srv logs `output EOF -> reader exiting -> turn_active flip (process exited)`, **zero** lingering in-container processes, and the CLI's real error text reaches the pane's blockfile.

The error it now surfaces is `Not logged in · Please run /login` — that's a **separate**, pre-existing gap (this dev channel has no Armory account bound; per `CLAUDE.md`, accounts default to isolated on non-`stable` channels). Not addressed here. It's the right illustration though: the same turn that used to hang silently forever now fails visibly in ~2s with something the user can act on.

6 unit tests cover the argv assembly (append, codex `-` replacement, non-trailing `-` left alone, multi-line single token, metacharacters as data, empty message still a token). Full srv suite green: 2885 passed.

## Not fixed here

- The credential gap above.
- #2872 (`provider_flags`/`--fork-session` dropped from `cmd:args` on every send).
- Frontend/srv provider-catalog drift: `--exclude-dynamic-system-prompt-sections` exists only in the srv catalog, so UI-launched agents never get it.

### DIFF
--- a/.changesets/1788224235-fix-agent-agent-open-looked-for-the-cli-at-a-path-that-never-363l.md
+++ b/.changesets/1788224235-fix-agent-agent-open-looked-for-the-cli-at-a-path-that-never-363l.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): agent.open looked for the CLI at a path that never existed
--- a/.changesets/1788224235-fix-container-container-turns-hung-forever-stdin-can-never-r-ievy.md
+++ b/.changesets/1788224235-fix-container-container-turns-hung-forever-stdin-can-never-r-ievy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(container): container turns hung forever — stdin can never reach EOF
--- a/agentmux-srv/src/backend/blockcontroller/subprocess/container_spawn.rs
+++ b/agentmux-srv/src/backend/blockcontroller/subprocess/container_spawn.rs
@@ -16,7 +16,6 @@ use std::sync::atomic::Ordering;
 use std::sync::Arc;
 
 use futures_util::StreamExt as _;
-use tokio::io::AsyncWriteExt;
 
 use crate::backend::blockcontroller::{
     core, publish_controller_status, session_stats, shell, STATUS_DONE, STATUS_RUNNING,
@@ -25,6 +24,95 @@ use crate::backend::wps;
 
 use super::{argv::build_turn_argv, SubprocessController, SubprocessControllerInner, SubprocessSpawnConfig, SUBPROCESS_OUTPUT_SUBJECT};
 
+/// Put the turn message into the argv, where a container turn's prompt has to go.
+///
+/// A container turn cannot use stdin at all: bollard's hijacked exec stream
+/// can't half-close the write side, and on Windows the Docker transport is a
+/// named pipe, which has no half-close — so the in-container process's stdin
+/// NEVER reaches EOF. Verified live 2026-09-01: both `drop(input)` and an
+/// explicit `input.shutdown().await` left `claude -p` hung with zero output,
+/// no error and no exit, which is exactly what a sandbox pane showed.
+///
+/// This was survivable while container agents carried
+/// `--input-format stream-json`: that protocol is newline-delimited and never
+/// waits for EOF — the same reason `container.rs`'s `exec (io)` test uses
+/// `read line` rather than `cat`. Correcting container agents to the one-shot
+/// `-p` argv (the persistent flags crashed them outright) is what made the EOF
+/// dependency load-bearing, turning a crash into a silent hang.
+///
+/// Docker takes argv as a JSON array, so there is no shell quoting to get
+/// wrong: a multi-line `# Session Context` body is a single token.
+fn turn_argv_with_message(mut cmd: Vec<String>, message: &str) -> Vec<String> {
+    // A trailing `-` is the "read the prompt from stdin" positional (codex).
+    // It designates the same slot, so replace it rather than append after it —
+    // appending would leave the CLI still trying to read a stdin that can never
+    // EOF, i.e. the exact hang this function exists to avoid.
+    if cmd.last().map(|a| a == "-").unwrap_or(false) {
+        cmd.pop();
+    }
+    cmd.push(message.to_string());
+    cmd
+}
+
+#[cfg(test)]
+mod turn_argv_tests {
+    use super::turn_argv_with_message;
+
+    fn v(xs: &[&str]) -> Vec<String> {
+        xs.iter().map(|s| s.to_string()).collect()
+    }
+
+    /// The prompt must reach the CLI in argv — the whole point.
+    #[test]
+    fn appends_the_message_as_the_final_token() {
+        let got = turn_argv_with_message(v(&["claude", "-p", "--verbose"]), "say hi");
+        assert_eq!(got, v(&["claude", "-p", "--verbose", "say hi"]));
+    }
+
+    /// codex's trailing `-` MEANS "read the prompt from stdin". Appending after
+    /// it would leave the CLI waiting on a stdin that can never reach EOF.
+    #[test]
+    fn replaces_a_trailing_stdin_positional_rather_than_appending_after_it() {
+        let got = turn_argv_with_message(v(&["codex", "exec", "--json", "-"]), "say hi");
+        assert_eq!(got, v(&["codex", "exec", "--json", "say hi"]));
+        assert!(!got.iter().any(|a| a == "-"), "the stdin positional must be gone");
+    }
+
+    /// Only a TRAILING `-` is the stdin positional; one earlier in the argv is
+    /// some other flag's value and must survive untouched.
+    #[test]
+    fn leaves_a_non_trailing_dash_alone() {
+        let got = turn_argv_with_message(v(&["cli", "--file", "-", "--json"]), "hi");
+        assert_eq!(got, v(&["cli", "--file", "-", "--json", "hi"]));
+    }
+
+    /// Multi-line prompts (the `# Session Context` body) stay ONE argv token —
+    /// Docker takes argv as a JSON array, so there is no quoting to get wrong.
+    #[test]
+    fn keeps_a_multiline_message_as_a_single_token() {
+        let msg = "# Session Context\n\nline two\nline three";
+        let got = turn_argv_with_message(v(&["claude", "-p"]), msg);
+        assert_eq!(got.len(), 3, "exactly one token was added");
+        assert_eq!(got[2], msg);
+    }
+
+    /// Quotes, spaces and metacharacters are data here, never shell syntax.
+    #[test]
+    fn does_not_mangle_shell_metacharacters() {
+        let msg = "say \"hi\" & echo $HOME; rm -rf /tmp";
+        let got = turn_argv_with_message(v(&["claude", "-p"]), msg);
+        assert_eq!(got.last().unwrap(), msg);
+    }
+
+    /// An empty message is still its own token — dropping it would shift a
+    /// positional-taking CLI onto the wrong argument.
+    #[test]
+    fn an_empty_message_is_still_a_token() {
+        let got = turn_argv_with_message(v(&["claude", "-p"]), "");
+        assert_eq!(got, v(&["claude", "-p", ""]));
+    }
+}
+
 impl SubprocessController {
     /// Spawn a container agent turn via Docker socket (P1a: no secrets in argv).
     ///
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

Agent2-asaf:
All three findings were right, and together they ruled out two transports before the third one held. Fixed in b7579d0.

**reagent P1 — secrets in argv.** Correct, and it was a regression I introduced against this module's own stated invariant. The host path (`host_spawn.rs`) deliberately keeps message content off argv; putting it there for containers meant any secret pasted into a chat message was readable via `docker top`, host `ps`, or `/proc/<pid>/cmdline`. The one thing I'd add: the exposure is host-side, not in-container — the agent legitimately has its own prompt — which is what makes a *file inside the container* an acceptable home for it while argv is not.

**codex P1 — `-p ""` / trailing `-` placeholders.** Also correct, and worse than the empty-prompt symptom suggests: `providers.rs` encodes "the prompt arrives on stdin" for codex (`-`) and gemini/qwen/kimi/antigravity (`-p ""`), and my argv version silently made those a lie for five providers. This one **dissolves** rather than getting a targeted fix: with a real stdin restored, the CLI argv is passed through completely unchanged and both placeholders are correct again. Two tests pin exactly that (`keeps_codexs_trailing_stdin_positional`, `keeps_an_empty_prompt_placeholder`).

**codex P2 — unbounded prompt in one argv entry.** This is the one that changed the design. My first response to reagent's P1 was to move the message into an env var, which looked like it fixed both — it doesn't. `MAX_ARG_STRLEN` is per *string* and applies to `envp` as much as `argv`. Measured inside the actual agent image:

| size | env var |
|---|---|
| 100,000 B | ok |
| 130,000 B | ok |
| 200,000 B | `Argument list too long` |

So env was a bounded transport wearing a fix's clothing. Thanks for pushing on it.

### What it does now

The prompt is uploaded into the container as a file (`upload_turn_prompt`, using the `tar` crate already in the tree) and the CLI runs as `"$@" < "$F"`, then `rm -f "$F"` while preserving the CLI's exit status for `inspect_exec`. Only the *path* is in argv. That satisfies all three constraints at once: a real stdin that reaches a real EOF, nothing message-shaped in argv or env, and no size ceiling.

### One more bug, found only by running it

The first cut of the file transport passed every unit test and both live turns — and leaked a prompt file per turn, including a 300 KB one. A tar entry defaults to uid 0, and `/tmp` is sticky (`1777`), under which a non-root process cannot unlink a root-owned file no matter what its mode is. `rm -f` failed silently, exactly as `-f` is designed to. Caught by listing `/tmp` after the turns rather than trusting that the cleanup ran.

Fixed by owning the file as the exec user, resolved per container via `id -u; id -g` and cached rather than assuming a uid, so it holds for any image whatever user it runs as (falls back to `(0,0)` with a warning). Mode tightened to 0600.

### Verification

Live through `agent.open` + `agent.send` against a real container:

- normal turn — settles in **2.0s**, `status=done`, `session_id` captured
- **293 KiB** prompt — settles in **2.0s**, same result; this size fails outright under both the argv and env transports
- `/tmp` afterwards: **empty**
- no lingering in-container processes, no upload/exec errors

7 unit tests on the exec shape (path-not-message in argv, the redirect, the cleanup, exit-status preservation, argv passed through verbatim, and both stdin placeholders). Full srv suite: 2886 passed.

Note the green CI on this PR is from `ac63d97`, the argv version — it predates all of the above.

a5af:
@codex review

Agent2-asaf:
Both correct. Fixed in 8435c72.

**P1 — the poisoned `(0,0)` cache.** Right, and sharper than "no retry": the *previous* commit is what made it fatal. Caching the fallback was survivable while the mode was 0644 — the file was at least readable, just not removable. Tightening to 0600 in the same change turned it into a total break, since a root-owned 0600 prompt is unreadable by the exec user AND un-unlinkable under `/tmp`'s sticky bit. One transient Docker hiccup would then fail every subsequent turn on that container and orphan a file each time. Two changes that were each fine in isolation, harmful together, which is exactly the kind of thing I wouldn't have caught from the diff alone.

Now only a *successful* probe is cached; failure returns `None` and is retried next turn. `upload_turn_prompt` also fails with a clear, retryable error instead of writing a file the CLI can't open — if the uid can't be resolved there is no usable file to write, so a named error beats permission-denied-plus-a-leak.

I split the parse out as `parse_id_output` so it's testable without a daemon. The load-bearing test is `a_missing_or_malformed_probe_is_none_not_root`: empty output, a truncated single number, `id: command not found`, the wrong `uid=1000(agent)` form, and a negative value all yield `None` — never a silent root.

**P2 — orphan when the exec fails after a successful upload.** Also correct: the wrapper carries the `rm -f`, so if `cm.exec` never starts the wrapper, nothing cleans up. Added `remove_turn_prompt` on exactly that path — best-effort, run as the user that owns the file, and it drives the output stream to completion since Docker starts an exec lazily on the attached connection.

### Verification

Live through `agent.open` + `agent.send`, five consecutive turns on one container:

| turn | settled | status |
|---|---|---|
| normal | 2.0s | done, session_id |
| 293 KiB | 2.0s | done, session_id |
| burst 0/1/2 | 2.0s each | done, session_id |

`/tmp` afterwards: **empty**. Container exec identity confirmed as `1000:1000`, i.e. genuinely not root — which is what makes both the ownership fix and this cache fix load-bearing rather than theoretical.

Full srv suite: 2890 passed.

I also corrected a doc comment on `spawn_container_turn` still describing the argv transport from two iterations ago.

a5af:
@codex review

Agent2-asaf:
Both correct, both fixed in 9ea0ec7. Same root cause seen from two angles: the prompt file is mode 0600 and can hold a pasted secret, and it lives in a container that outlives the turn — so any path that skips the cleanup is a real leak, not just litter.

**Interrupted turns.** Right, and the mechanism is exactly as you describe: the `rm -f` lives *inside* the wrapper, and `pkill -f <cli>` matches full command lines. `container_turn_exec` passes the CLI and its args as `"$@"`, so the wrapper's own command line contains the CLI name and pkill takes the shell down with it before the cleanup line is ever reached. The interrupt path now calls `remove_turn_prompt` itself, from outside the process being killed; the path is retained in `prompt_path_for_kill` for that.

I'll be straight about the evidence here: I did not manage to stage a clean end-to-end interrupt repro, because an unauthenticated turn finishes in ~50ms and there's nothing to interrupt. I did confirm the underlying behaviour the hard way — an exploratory `pkill -f sleep` inside the container matched and killed its `sleep infinity` init, which is the same full-command-line matching that makes the wrapper a target. The rest follows from the argv `container_turn_exec` demonstrably builds, which is pinned by unit tests. Treating it as read rather than claiming a live verification I don't have.

**Stale cached identity.** Also right, and it's the same class of bug as the `(0,0)` poisoning from the last round: the cache is keyed by name, `ensure_running` deliberately recreates a removed container under that same name, and a replacement execing as a different uid — changed custom image, or the same mutable tag repulled — would get the old identity and chown every prompt to a user that can't read it at 0600. Permanent until restart, same failure shape as before. Evicted in `create_and_start`, which is the single point where a name starts pointing at a new container. I preferred that to keying by container ID because it needs no extra inspect round-trip per turn and there is exactly one place to get it wrong.

Full srv suite: 2890 passed. Prior live results (five consecutive turns including a 293 KiB prompt, each settling in 2.0s, `/tmp` empty afterwards) still stand — this change only adds cleanup on two paths that weren't being exercised there.

Agent2-asaf:
Correct, and it's a window this PR widened — worth being precise about that. Fixed in aef42f6.

`stop_subprocess` reports success when `inner.kill_tx` is `None`, reading "nothing to interrupt" as "already stopped". `kill_tx` was installed only after the exec had started, so every await before that point was a hole where a stop was silently discarded and the turn then ran on and flipped the status back to running.

That hole already existed for the exec round-trip. What makes it worth closing now rather than noting is that this PR put an identity probe and a prompt upload of potentially hundreds of KB inside it — I turned a narrow race into a routinely-hit one.

The kill channel, `STATUS_RUNNING` and `set_active_turn(true)` now all happen before any await. A stop arriving during startup isn't lost: `kill_rx` holds it, and the reader loop selects on it `biased`, so it's observed on the first poll and interrupted immediately — including the prompt-file cleanup from the previous round. The exec-failure early return also clears the active-turn flag now, which it didn't need to do while the flag was only set after a successful start.

I dropped one thing from my first draft: a second `remove_turn_prompt` call on the upload/exec failure path. The `cm.exec` error branch already does it, so it would have been a redundant double-`rm`.

### Verification — this one did reproduce

Unlike the interrupt finding, this window is wide enough to hit deliberately. Sending a 200 KB message and issuing `agent.stop` immediately after puts the stop squarely in the upload window:

| turn | settled | status |
|---|---|---|
| normal | 2.0s | done, exit 1 |
| 293 KiB | 1.0s | done, exit 1 |
| **stop during startup** | **1.0s** | **done, exit -1 (interrupted)** |

`exit -1` is the interrupted path, not a completed turn — i.e. the stop was honoured rather than swallowed. `/tmp` empty afterwards, so the interrupted turn also cleaned up its prompt. Full srv suite: 2890 passed.

Agent2-asaf:
Verified and fixed in dda5b15 — you're right on both counts.

`set_exited` clears `active_turn` itself (`health.rs:75`, before recording the exit code), and `build_status_snapshot` takes `turn_active` as an explicit parameter rather than reading `is_active_turn()`, so my call had no observable effect. Worse, the comment asserted something false — that the hoisted `set_active_turn(true)` "has to be undone" here — which is exactly the kind of comment that survives into a future change and misleads whoever reads it next.

Removed the call and kept the useful half of the observation on `set_exited`, where it now explains why no undo is needed rather than claiming one is.

I checked this rather than taking it as read, since "dead code" and "subtly load-bearing" look identical from a diff: `health.rs:73-79` sets `active_turn = false` unconditionally, and `mod.rs:270` confirms `turn_active` is a parameter, not a read of the monitor.

Suite still green: 2890 passed.

Agent2-asaf:
Both correct, fixed in 73e05f8 — plus a third instance of the same error you didn't flag.

You're right that these are worse than ordinary staleness. `ExecSession`'s doc said *"Turn input belongs in argv"* and the `drop(input)` comment said *"the turn message is in argv"* — both describing the intermediate design that was replaced **specifically because** argv leaks a pasted secret via `docker top` / host `ps` / `/proc/<pid>/cmdline`. Left in, they read as license to put message content back in argv, directly against the invariant the rest of the change exists to defend.

Grepping for the pattern turned up a third: `ContainerManager::exec`'s own doc ended with *"Pass its input in argv instead."* Same error, same file, and arguably the most likely to be followed since it sits on the function whose `attach_stdin` flag is the one-way door. All three now point at `upload_turn_prompt`, which explains why argv **and** env are both wrong.

The remaining `argv` mentions in these two files are accurate and intentional — the module-level CWE-214 note about env vars not travelling in argv, and `upload_turn_prompt`'s own rationale for rejecting argv and env in turn.

Suite still green: 2890 passed.

Issue #2863, opened 2026-08-31
Flaky test: subprocess_io::exit_code_nonzero times out at 5s on CI


AgentA-asaf:
`agentmux-srv/tests/subprocess_io.rs::exit_code_nonzero` failed on a **docs-only** PR (#2861, single markdown file), so it cannot have been caused by the change under test.

```
thread 'exit_code_nonzero' (9123) panicked at agentmux-srv/tests/subprocess_io.rs:123:10:
wait timed out: Elapsed(())
test result: FAILED. 5 passed; 1 failed
```

Run: https://github.com/agentmuxai/agentmux/actions/runs/33384071333/job/99462615878 (`check --tests + test (ubuntu-latest)`)

## Cause

`subprocess_io.rs:121` gives a spawned `node` process **5 seconds** to exit:

```rust
let status = timeout(Duration::from_secs(5), child.wait())
    .await
    .expect("wait timed out")
```

That is a wall-clock budget covering node process startup + exit on a shared GitHub runner under a `cargo test --workspace` load. It is tight enough to lose the race intermittently. The other five tests in the file passed, so this is not a broken toolchain — just this budget.

## Suggested fix

Raise the timeout substantially (30s+). The timeout exists to stop the suite hanging forever if `child.wait()` never returns; it is not asserting anything about *how fast* the process exits, so a larger value costs nothing on the happy path and removes the flake. The sibling `spawn_node` tests in the same file are worth a look for the same pattern.

## Why file rather than just re-run

The run could not be re-run (`cannot be rerun; its workflow file may be broken`), so the recovery was an empty commit to re-trigger — which means this failure mode costs a full CI cycle and looks like a real regression to whoever hits it next.

PR #2884, opened 2026-09-01
fix(cef): restore ctrl+wheel zoom in floating panes

AgentA-asaf:
Fixes the repo owner's report that Ctrl+Scroll zoom stops working once a pane is torn off. Implements §8 of the analysis merged in #2881.

## Cause

A floater hosts its browser as a **child HWND** (`floating_pane.rs`, `WindowInfo::set_as_child`); the main window is CEF **Views**-hosted. Child-HWND CEF browsers never dispatch Ctrl+Wheel to the renderer — CEF consumes it for its own native `HostZoomMap`-shared page zoom first.

Measured with a `wheel` recorder in each window, driven by a real mouse:

| window | input | events reaching the DOM |
|---|---|---|
| main, docked terminal | Ctrl+Scroll | **22 / 22, `ctrl: true`** |
| floater, same terminal | Ctrl+Scroll | **0** |
| floater, same terminal | plain scroll | **49, `ctrl: false`** |

The floater receives wheel input normally; only Ctrl+Wheel is suppressed. Everything in-page was already proven working end to end — per-view handlers, the `term:zoom` RPC, read-back, font application. The single missing link was the DOM event.

**Browser panes are also child-HWND browsers and already needed this exact interception** (`browser_pane/hwnd.rs:403`, whose comment names CEF's native shared zoom as the reason). Floaters never got one.

## Approach

**Host** — `floater_wheel.rs` subclasses the floater's browser HWND and every Chromium descendant (mouse input goes to the deepest descendant, not the ancestor), intercepts `WM_MOUSEWHEEL` with `MK_CONTROL`, and emits `floater:ctrl-wheel` to *that window's* renderer via `emit_event_to_window` — not `emit_event_from_state`, whose "main"/first-available fallback is the bug reagentx caught for `browser-pane-clicked` on #2597.

Installed from **`on_load_end`, not `on_after_created`**: Chromium recreates `Chrome_RenderWidgetHostHWND` on every page load, so a once-installed subclass strands on a destroyed HWND. This is why `install_browser_pane_focus_redirect` is wired from both. Idempotent, and it also covers a pane-pool window being promoted to a real floater (which re-navigates). Un-subclassed on close, since Windows reuses HWND values.

**Frontend** — `floater-ctrl-wheel.ts` **re-dispatches the swallowed DOM event** rather than calling a zoom function. Ctrl+Wheel zoom is implemented five times over (`term`, `armory`, `editor`, `swarm`, `warden`), each on its own view root writing `term:zoom` for its own block. Synthesising the event lets all of them run unchanged, so this file needs no knowledge of view types and won't need editing when a sixth is added. Untrusted events suffice — none of those handlers check `isTrusted`.

## One deliberate divergence from the precedent

`browser_pane/hwnd.rs:403` puts its `return 0` **outside** the context lookup, so an HWND it can't resolve silently eats Ctrl+Wheel and does nothing. This hook falls through to the original WndProc instead, leaving CEF's native zoom — degraded but not dead. That is a latent bug in the browser-pane hook; it's noted in the analysis and **not** fixed here.

## Known limitation, inherited

**Windows only.** The browser-pane equivalent is `cfg(windows)`-gated with no macOS/Linux counterpart either (`hwnd.rs:395`). Stated rather than hidden.

## Verification

- `cargo test -p agentmux-cef floater_wheel` — **3 passed**
- `tsc --noEmit` — clean

The 3 tests cover the wheel-delta sign conversion: `WM_MOUSEWHEEL`'s convention is the **opposite** of DOM `deltaY` (positive = away from user = zoom in, vs positive `deltaY` = scroll down = zoom out), and silently inverting it would make zoom run backwards.

**Not yet verified against a live floater** — that needs a host rebuild. Worth a manual check that Ctrl+Scroll zooms a torn-off terminal *and* in the correct direction.

### DIFF
--- a/.changesets/1788224397-fix-cef-restore-ctrl-wheel-zoom-in-floating-panes-utat.md
+++ b/.changesets/1788224397-fix-cef-restore-ctrl-wheel-zoom-in-floating-panes-utat.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): restore ctrl+wheel zoom in floating panes
--- a/agentmux-cef/src/client/lifecycle.rs
+++ b/agentmux-cef/src/client/lifecycle.rs
@@ -1030,6 +1030,15 @@ impl AgentMuxHandler {
                         label: lbl.to_string(),
                     },
                 );
+
+                // Restore the WndProcs the Ctrl+Wheel hook subclassed and drop
+                // its bookkeeping. Windows reuses HWND values, so leaving stale
+                // entries behind would let a future window inherit a hook whose
+                // context points at a dead floater. See floater_wheel.rs.
+                #[cfg(target_os = "windows")]
+                unsafe {
+                    crate::floater_wheel::remove_floater_ctrl_wheel_hook(lbl);
+                }
             }
         }
 
--- a/agentmux-cef/src/client/navigation.rs
+++ b/agentmux-cef/src/client/navigation.rs
@@ -360,7 +360,7 @@ impl AgentMuxHandler {
 
     pub(crate) fn on_load_end(
         &mut self,
-        browser: Option<&mut Browser>,
+        mut browser: Option<&mut Browser>,
         frame: Option<&mut Frame>,
         _http_status_code: i32,
     ) {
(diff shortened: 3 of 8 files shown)


a5af:
@codex review

AgentA-asaf:
All four findings were correct and are fixed in `614c24f85`. Two of them would have shipped a fix that only worked sometimes, which is worse than one that visibly doesn't.

**reagentx P1 — `window_handle()` is the wrong HWND source.** Verified: it returns null after page load for `set_as_child`-embedded browsers, which is exactly what floaters use, and `SPEC_POOL_WINDOW_HWND_NULL_2026_05_06.md` already documents this with the `window_hwnds` cache as the established workaround. Switched to `commands::window::resolve_window_hwnd`. A miss now **logs loudly** rather than silently no-opping — a hook that failed to install presents as "Ctrl+Wheel does nothing", indistinguishable from the original bug, and you were right that the PR admitting it was untested against a live floater made that worse.

**codex P1 — pool-promoted floaters kept a dead label.** Confirmed at `pane_pool.rs:497`: promotion does `RelabelBrowser` then emits `pool:pane-promote`, with **no navigation**, so `on_load_end` never re-fires. Since the floater I measured this bug on was itself pool-promoted, the fix would have failed on the most common path while working elsewhere.

Fixed with `relabel_floater_ctrl_wheel_hook`. Worth noting why it rewrites contexts instead of installing a second hook on the outer popup, which was my first attempt: contexts are keyed by HWND and `find_context` walks **up** from the deepest descendant, so a fresh entry above would still lose to the stale one on the browser HWND beneath it.

**codex P2 + reagentx P2 — stale descendant HWNDs.** Both of you found this from different angles (Windows reusing a handle so the "already hooked" check skips a live widget; entries leaking for the process lifetime). Handled at the source rather than at either call site: `WM_NCDESTROY` now purges both maps for that HWND, which covers reinstall-after-navigation and close-time cleanup at once.

**codex P2 — the synthesized event ignored where the user scrolled.** Correct, and it matters more than "position fidelity": `AgentShellSubblock.tsx` and `ToolBlock.tsx` register their own scoped Ctrl+Wheel handlers, so centre-aiming would zoom the whole block regardless of cursor. `lParam`'s screen coords are now converted to client space (GA_ROOT anchor) and forwarded, with the frontend dividing by `devicePixelRatio`. Centre remains the fallback only when conversion fails.

**On the CI failure:** unrelated flake — `agentmux-bashwrap::bash_wrap::tests::a1_e2e_static_label_then_delayed_cr_overwrite_collapses_to_one_line` (93 passed, 1 failed), a different crate and another timing-sensitive test. Second such flake today; the first is filed as #2863.

Verification: `cargo test -p agentmux-cef floater_wheel` 3 passed, `cargo check` and `tsc --noEmit` clean. Still **not** verified against a live floater — that needs a host rebuild.

PR #2885, opened 2026-09-01
docs(browser-pane): resolve the camera spec's blocking Phase 0 from CEF source

AgentY-asaf:
The camera spec (#2880) shipped with Phase 0 marked **blocking** — *"do not proceed without this"* — to verify CEF permits retaining `MediaAccessCallback` across the handler's return. It turned out not to need a throwaway build: the CEF source is on this machine and the headers answer it outright.

## Resolved: the async design is viable

`cef/include/cef_permission_handler.h`:

> *"Return true and call `CefMediaAccessCallback` methods **either in this method or at a later time** to continue or cancel the request."*

And `CefMediaAccessCallback : public virtual CefBaseRefCounted`, documented as *"used for asynchronous continuation of media access permission requests."* Retaining the `CefRefPtr` and continuing after a prompt resolves is **intended usage, not a workaround** — so the fallback designs the spec contemplated (pre-navigation allowlist, in-page interstitial) aren't needed.

## Resolved: the bitmask

From `internal/cef_types.h` — `DEVICE_AUDIO_CAPTURE = 1<<0`, `DEVICE_VIDEO_CAPTURE = 1<<1`, `DESKTOP_AUDIO_CAPTURE = 1<<2`, `DESKTOP_VIDEO_CAPTURE = 1<<3`. The inferred pattern was right, but it's now *checked* rather than assumed — which is the standard a permission bitmask deserves, and why the spec deliberately refused to assert it.

## Two findings the same paragraph gives for free

- **Returning `false` IS the Alloy deny path.** So today's no-handler behaviour for browser panes is the header's documented default, not an accident — worth knowing before anyone "fixes" it.
- **`--enable-media-stream` bypasses the handler entirely** and grants everything: *"This method will not be called if the `--enable-media-stream` command-line switch is used."* AgentMux does **not** pass it (verified across `agentmux-cef` and `agentmux-launcher`). Now recorded as a standing constraint — adding that switch for any reason would silently void **every** guarantee in §4.

## And one that hardens §3.7

CEF's public headers expose **no media-capture termination API** — no `MediaStream`/`StopMediaCapture` surface in `cef_browser.h` or `cef_permission_handler.h`. So "reload the pane" isn't the best of several options for revoke, it's the *only* one CEF offers. Worth stating plainly given §4's revoke guarantee rests entirely on it.

Added Q6 for whether CDP could do better, with the caveat that the codebase has **no `ExecuteDevToolsMethod` call sites at all** — so that path introduces a new dependency surface for one feature, which should be weighed against simply accepting the reload.

**Phase 1 (threading pane identity into the client) is now the entry point.**

## Test plan

Docs-only, no runtime surface. All claims verified against the CEF source at `~/cef-build/chromium_git/cef/include/`.

### DIFF
--- a/.changesets/1788224716-docs-browser-pane-resolve-the-camera-spec-s-blocking-phase-0-tr7w.md
+++ b/.changesets/1788224716-docs-browser-pane-resolve-the-camera-spec-s-blocking-phase-0-tr7w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(browser-pane): resolve the camera spec's blocking Phase 0 from the CEF headers
--- a/docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md
+++ b/docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md
@@ -112,11 +112,34 @@ three places.
 permitting the `MediaAccessCallback` to be **retained and called later**, off
 the initial callback return.
 
-**This must be verified before implementation** (see §7 Q1). If CEF requires a
-synchronous decision, the whole design changes shape: the only options become a
-pre-granted allowlist configured *before* navigation, or an in-page interstitial
-that re-triggers `getUserMedia` after the user opts in. Do not begin building
-§3.5 until this is settled — it is the load-bearing assumption.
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2886, opened 2026-09-01
feat(agent): log every classified agent failure (Phase 0(a))

AgentY-asaf:
I tried to actually answer the headless-retry spec's own Phase 0 question — *how often is a transient failure classified for a block with no mounted pane?* — and found the data **cannot exist yet**.

## The finding

None of the three classify/persist/publish sites logs anything. There is no `tracing::` call adjacent to any `persist_last_failure` in:

- `persistent.rs` (`:1672`, `:3042`, `:3136`)
- `host_spawn.rs` (`:653`)
- `container_spawn.rs` (`:446`)

The only classification that *is* logged is `agents/runner.rs:328` — `"agent run failed"` with `code` and `retryable` — which is a different path, and not one a controller-driven turn takes.

## The trap this avoids

Searching the retained srv logs for `rate_limited` / `overloaded_error` / `agent_failure` returns **nothing**.

It would be very easy to read that as *"transient failures on unrendered panes are rare, so this spec isn't worth building"* — and close the whole thing on that basis. But the silence means the events are **unlogged**, not infrequent. Absence of evidence, and not even the kind that could be evidence of absence.

I've recorded that explicitly in the spec so the next person doesn't mistake the silence for a measurement.

## What Phase 0 actually is

Two steps, not one:

- **(a)** add a log line or counter at each of the three sites, recording the failure class and whether a renderer is currently subscribed for that block;
- **(b)** let it run, then read it.

Step (a) is small and **independently worthwhile regardless of what this spec decides** — a classified agent failure leaving no trace in the logs is a diagnosability gap on its own. If someone reports "my cron agent silently did nothing overnight" today, there is nothing in the logs to confirm or refute a 429.

## Test plan

Docs-only, no runtime surface. All file:line claims verified against current `main`.

### DIFF
--- a/.changesets/1788224826-docs-agent-headless-retry-phase-0-needs-instrumentation-adde-2kki.md
+++ b/.changesets/1788224826-docs-agent-headless-retry-phase-0-needs-instrumentation-adde-2kki.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(agent): headless-retry Phase 0 needs instrumentation added first
--- a/agentmux-srv/src/backend/blockcontroller/core.rs
+++ b/agentmux-srv/src/backend/blockcontroller/core.rs
@@ -228,6 +228,33 @@ pub(crate) fn persist_last_failure(
             return;
         }
     }
+    // Every classified agent failure gets exactly one log line, here at the
+    // shared choke point rather than at the five call sites — `persistent.rs`
+    // (x3), `host_spawn.rs` and `container_spawn.rs` all funnel through this
+    // function, and a future controller path would too.
+    //
+    // Before this, a classified failure left NO trace in the srv log at all:
+    // none of those sites logged, and the only logged classification was
+    // `agents/runner.rs`'s "agent run failed", which is a different path that a
+    // controller-driven turn never takes. That made a whole class of question
+    // unanswerable from logs — "did my cron agent hit a 429 overnight?" had no
+    // answer, and an empty grep looked like "it never happens" rather than
+    // "it is never recorded". See
+    // docs/specs/SPEC_HEADLESS_TRANSIENT_RETRY_2026_08_31.md §Phase 0.
+    //
+    // Logged only for `Some` — this is also called with `None` after every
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentY-asaf:
Good catch, and it turned out to be worse than reported — which changes what this PR is.

**You found two more sites in `persistent.rs`. Counting properly, there are nine**, across five files: `persistent.rs` (six calls), `container_spawn.rs`, `host_spawn.rs`, `agent_handlers/input.rs`, and `app_api/agent_io.rs`. So my "three" was wrong, your "five" was closer, and neither of us had it — I'd written the number off a `grep | head -3` and read the truncation as the answer.

**Rather than correct the enumeration, I've removed the need for one.** This PR now *implements* Phase 0(a) instead of instructing someone to: a single `tracing::warn!(target: "agent-failure", …)` inside `core::persist_last_failure` — the choke point all nine calls funnel through — recording `block_id`, `code` and `retryable`, firing only for `Some` (the same function is called with `None` after every clean turn).

That's immune to the miscount, and to whatever the tenth site turns out to be. The enumeration went stale *before the PR merged*, which is about as clear a signal as you get that a spec shouldn't hold that list.

**Why this is worth shipping on its own merits**, independent of the retry spec: a classified agent failure previously left no trace in the srv log at all. The only logged classification is `agents/runner.rs:328`, a path controller-driven turns never take. So *"did my cron agent hit a 429 overnight?"* had no answer — and an empty grep looked like **"never happens"** rather than **"never recorded"**, which is precisely the wrong conclusion to draw before deciding whether to build the rest of the spec.

Phase 0(b) still needs a running build, and I've noted it only answers half the original question: *"was a renderer subscribed for that block"* has no query behind it today (`Broker` exposes no per-scope subscriber lookup; `WaveEvent::has_scope` is a different thing). Adding one is a prerequisite for the full measurement and deliberately out of scope for a logging change.

Verified: `cargo check -p agentmux-srv` clean, `blockcontroller::core` tests pass.

PR #2887, opened 2026-09-01
feat(agent-pane): attribute scroll shrinks to the rows that caused them

Agent1-asaf:
## Summary

Step 1 of `SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31.md` (#2868), and the gate on steps 2-5. **Diagnostic only — no behavior change.**

The existing `[wave-scroll-shrink]` line reports that the pane's `scrollHeight` got smaller between two pin checks, but not *which* content shrank. That's why every conclusion in the 08-21 and 08-22 findings docs had to be hedged, and why the 08-22 doc's recurring ~251–252px lead was closed on 08-31 with no culprit: no component of that height exists, so the number was almost certainly a **sum** of smaller shrinks — which a net delta cannot decompose.

**`shrink-trace.ts`** — records observed row heights, keeps a bounded log of shrinks only, attributes a pane delta to the rows that shrank in the preceding 250ms window, and reports the **unattributed remainder** explicitly. That remainder is the point: it says whether the observed rows explain the shrink or whether something unobserved did. Drawing a conclusion without it would repeat the exact error those two docs were corrected for.

**A second, diagnostic-only `ResizeObserver` over streaming-buffer rows** — which nothing measured before. Kept separate from `measureRO` deliberately: that one dispatches `RowMeasured` into the agent-pane-layout slice, and the slice models the *virtualized* partition only, so feeding it streaming rows would be a behavior change rather than instrumentation. The streaming buffer is also where the interesting shrinks are — a completing tool call is in the trailing N nodes by definition.

Sample output shape:

```
[wave-scroll-shrink] pane=43c1dea scrollHeight 14525px -> 1023px (delta=13502px)
  attributed: tc-9abc(tool) 13400->120px (sum=13280px, unattributed=222px)
```

**Not gated on `import.meta.env.DEV`**, unlike `perf-probe.ts` — the most useful data so far came from `task package` builds, where a DEV gate compiles it out entirely. Cost is bounded by recording only numbers on resize, and formatting a string only once a pane shrink has already been detected.

## Test plan

- [x] `shrink-trace.test.ts` — 10 new unit tests, incl. the multi-row-sum case that the 251px lead turned out to be, the negative-remainder case (rows shrank more than the pane), window expiry, double-crediting, and the unmount/remount fake-shrink guard.
- [x] `npx vitest run frontend/app/view/agent/virtualization/` — 146/146 pass.
- [x] Verified `DocumentRow`'s `ref` sits on the single outer `.agent-document-row` wrapper shared by every node type, so all streaming rows are observed uniformly rather than a subset (which would silently skew every attribution).
- [x] `task dev` running from this branch; instrumentation loaded, `[fe]` console bridge confirmed flowing.
- [ ] **Not yet observed live.** No `[wave-scroll-shrink]` line has been captured from a real streaming session with this build — that needs agent activity with chained tool calls in that instance. The attribution *logic* is unit-tested; the *emission* is not yet confirmed end to end. Flagging rather than claiming it works.

### DIFF
--- a/.changesets/1788223484-feat-agent-pane-per-node-shrink-attribution-in-the-wave-scro-swuq.md
+++ b/.changesets/1788223484-feat-agent-pane-per-node-shrink-attribution-in-the-wave-scro-swuq.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): per-node shrink attribution in the wave-scroll-shrink diagnostic
--- a/frontend/app/view/agent/virtualization/AgentDocumentVirtualList.shrink-attribution.test.tsx
+++ b/frontend/app/view/agent/virtualization/AgentDocumentVirtualList.shrink-attribution.test.tsx
@@ -1,0 +1,250 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * End-to-end wiring for per-node shrink attribution
+ * (SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31.md step 1).
+ *
+ * `shrink-trace.test.ts` covers the bookkeeping in isolation. What THIS file
+ * covers is the part that unit tests can't: that the streaming-buffer
+ * ResizeObserver is actually attached to real rendered rows, that what it
+ * records reaches `ShrinkTrace`, and that the result comes back out on the
+ * `[wave-scroll-shrink]` console line. Every one of those is a wiring step
+ * that can be individually correct and still not connect.
+ *
+ * jsdom has no layout engine and no ResizeObserver, so the fakes here mirror
+ * `AgentDocumentVirtualList.resize.test.tsx`'s: `triggerResize(el)` invokes
+ * whichever observer callbacks currently watch `el`, and geometry is stubbed
+ * explicitly. This proves the wiring, not the browser's real resize timing.
+ */
+
+import { cleanup, render } from "@solidjs/testing-library";
+import { createSignal } from "solid-js";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { AgentDocumentVirtualList } from "./AgentDocumentVirtualList";
+import { createAgentViewState } from "./state";
+import type { DocumentNode, DocumentState } from "../types";
+
+afterEach(() => cleanup());
+
+type ROCallback = (entries: ResizeObserverEntry[]) => void;
+let roInstances: { callback: ROCallback; targets: Set<Element> }[] = [];
+
+class FakeResizeObserver {
+    private callback: ROCallback;
+    private targets = new Set<Element>();
+    constructor(callback: ROCallback) {
+        this.callback = callback;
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

Agent1-asaf:
P2 accepted — it was dead code. Removed in `8c0a3df4d`.

I wrote `reset()` as a "test/teardown hook" and then never wired it to anything. Nothing needs it: the `ShrinkTrace` instance is created per-component so it's collected along with the component, and the per-node case that actually matters — a cap-advance retiring a tall row that later re-renders short — is already covered by `forget()`, which *is* called from the `<Key>` slot's `onCleanup`.

`tsc` clean, 146/146 in the virtualization suite still pass.

Still open on this PR (unchanged from the description): the emission path has not been observed live yet. The attribution logic is unit-tested, but no real `[wave-scroll-shrink]` line with an `attributed:` suffix has been captured from a streaming session. Holding the merge on that rather than on review state.

Agent1-asaf:
P1 accepted — real defect, and in the primary case. Fixed in `46d4504a7`.

Your read of the ordering is exactly right: the pin effect defers `scrollToTrueBottom()` with `queueMicrotask`, and reading `scrollHeight` there forces a layout flush, so the **pane** shrink is detected at the microtask checkpoint while RO callbacks aren't delivered until later in the rendering steps. The running→terminal case logged as wholly unattributed, and the row shrink then sat in the ring to be miscredited to a later unrelated delta — wrong on the one case this exists for, and wrong in the direction that *invents* attributions rather than missing them.

**I removed the ordering dependency instead of deferring the drain.** Your second suggestion (capture the row resize before the microtask path) is the one I took, generalized: row heights are now snapshotted synchronously inside `scrollToTrueBottom`, immediately after its own `scrollHeight` read. Both numbers come from the same layout-clean instant, and the row window is now *exactly* the pane window rather than a 250ms approximation. I avoided the defer-until-after-RO option deliberately — it would have required me to be right about rAF-vs-RO delivery order, and this subsystem has a long record of punishing that kind of confidence.

Most of the machinery fell out as a result: the RO, WeakMap, ref wiring, per-slot `onCleanup`, ring buffer, timestamps and `ATTRIBUTION_WINDOW_MS` are all gone. Pruning absent ids replaces `forget()`; diffing consecutive samples replaces the double-crediting guard.

Two things I found while rewriting:
- Switched to `offsetHeight` from `getBoundingClientRect().height`. The rect is scaled by an ancestor CSS `zoom` while `scrollHeight` is not, so the old code divided by zoom to compensate — not mixing the units is simpler and removes the chance of applying that correction backwards.
- Added `data-node-type` to the row so the sampler reads the node kind from the DOM rather than doing a reactive read from inside a microtask.

**On the test that missed this:** my original wiring test triggered the row resize *before* the pane pin — it encoded my assumed ordering rather than the real one, which is precisely how it passed over a broken design. The replacement fires **no row-level resize at all**: just a DOM height change and a pin check. I confirmed it fails against the old design and against a stubbed-out sampler.

`tsc` clean, 153/153 in the virtualization suite.

PR #2889, opened 2026-09-01
fix(cef): reject media-permission switches at both ingress paths

AgentY-asaf:
Closes the ingress documented in `SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md` §3.8 (found by Codex on #2885, spec'd in #2888).

## The problem

`AGENTMUX_CEF_EXTRA_FLAGS` splits an env var on whitespace and appends **every token verbatim** as a Chromium switch, with no validation. It exists to A/B GPU flags without a recompile.

But `--enable-media-stream` makes CEF **skip `OnRequestMediaAccessPermission` entirely** and grant all media permissions — per its own header: *"this method will not be called if the `--enable-media-stream` command-line switch is used."*

That's a **bypass, not a grant.** Browser panes deliberately install *no* permission handler so CEF's Alloy default denies. This switch turns that deliberate deny into a blanket allow for every origin with no prompt, and does it somewhere no handler-based guard can observe. So an env var — from a shell profile, a CI config, or **an agent with shell access** — silently gives every browser pane camera and microphone.

## Why this ships now rather than with the camera work

It's a **present-day exposure for microphone**, not a camera-feature prerequisite: it applies today, with none of the camera spec built. Waiting for Phase 2 of a feature that may never be funded would be the wrong call.

## The fix

Rejects `enable-media-stream` and `use-fake-ui-for-media-stream` (same outcome by a different route, and the flag most likely reached for while testing). Matching is on the switch **name**, so `=value` spellings are caught, and **case-insensitive** since Chromium's own parsing is — an uppercase spelling must not be a way around the guard. Rejection is logged loudly: someone who set it deliberately deserves to know it didn't apply.

Tests cover the bypass spellings, case-insensitivity, that the GPU diagnostics flags this variable exists for still pass unaffected, and that matching doesn't over-reach to unrelated switches containing "media" (`autoplay-policy`, `enable-media-session-service`).

## One deliberate limitation

This is a **denylist, not an allowlist**. An allowlist would be stricter but would break the variable's diagnostic purpose. §3.8 records the uncomfortable part honestly: this leaves a diagnostics escape hatch serving as a security boundary, which it was never designed to be. If that tradeoff is wrong, the alternative is dropping `AGENTMUX_CEF_EXTRA_FLAGS` to an allowlist entirely — a bigger call than I'd make unilaterally.

## Test plan

- `cargo test -p agentmux-cef media_switch_guard` — 4/4 pass

### DIFF
--- a/.changesets/1788225394-fix-cef-reject-media-permission-switches-from-agentmux-cef-e-68xy.md
+++ b/.changesets/1788225394-fix-cef-reject-media-permission-switches-from-agentmux-cef-e-68xy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): reject media-permission switches from AGENTMUX_CEF_EXTRA_FLAGS
--- a/agentmux-cef/src/app/mod.rs
+++ b/agentmux-cef/src/app/mod.rs
@@ -514,6 +514,34 @@ wrap_app! {
             command_line: Option<&mut CommandLine>,
         ) {
             if let Some(cmd) = command_line {
+                // Strip media-permission switches that arrived on the process
+                // command line, before anything else looks at it.
+                //
+                // The AGENTMUX_CEF_EXTRA_FLAGS guard further down covers only
+                // that one ingress. This covers the other: the launcher
+                // collects its own argv verbatim
+                // (`agentmux-launcher/src/main.rs:309`) and passes it to the
+                // host via `.args(args)` (`host_spawn.rs:41`/`:138`), so
+                // `agentmux.exe --enable-media-stream` reaches CEF without ever
+                // touching the env-var path.
+                //
+                // Both must be closed for the guarantee to hold — see
+                // docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md §3.8
+                // for why this switch is a handler BYPASS rather than a grant,
+                // and therefore invisible to every handler-based guard.
+                for name in MEDIA_PERMISSION_SWITCHES {
+                    let sw = CefString::from(*name);
+                    if cmd.has_switch(Some(&sw)) != 0 {
+                        cmd.remove_switch(Some(&sw));
+                        tracing::warn!(
+                            switch = %name,
+                            "removed media-permission switch from the command line — \
+                             it would bypass the CEF permission handler and grant \
+                             camera/mic to every browser pane with no prompt"
+                        );
+                    }
+                }
+
                 // Proactive memory-pressure guard (SPEC_MEMORY_PRESSURE_
                 // SUPERVISION_2026_06_16 §5.H): if the system is already
                 // critically low on commit at launch, start in software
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentY-asaf:
**Scope expanded since the approval — please re-review before this merges.** Holding rather than landing on an approval of a materially different diff, since this is security code.

## What changed after the first review

My original guard covered **only** `AGENTMUX_CEF_EXTRA_FLAGS`. That left the *other* documented ingress wide open: the launcher collects its own argv verbatim (`main.rs:309`) and passes it to the host via `.args(args)` (`host_spawn.rs:41`/`:138`), so **`agentmux.exe --enable-media-stream` reached CEF without ever touching the env-var path.**

Half a guard on a bypass is no guard — arguably worse, because it reads as protected.

Now also removed in `on_before_command_line_processing` via `CommandLine::remove_switch`, before anything else reads the command line, with the same loud warning. Both paths share one `MEDIA_PERMISSION_SWITCHES` list so they can't drift — a switch blocked on one route and allowed on the other would be the worst outcome.

## Also fixes a bad citation (codex P2 on #2888)

§3.8 cited `host_spawn.rs:79` as demonstrating unfiltered launcher forwarding. **That's wrong** — `:79` is a conditional, internally-controlled `--disable-gpu` built from retry state, not arbitrary pass-through. Anyone auditing this security-sensitive claim would have gone there and concluded it was unfounded.

Corrected to `main.rs:309` → `.args(args)`, and called out explicitly in the spec rather than silently swapped, since the wrong citation was already published.

## Verification

- `cargo test -p agentmux-cef media_switch_guard` — 4/4 pass

The tests cover the env-var helper (bypass spellings, case-insensitivity, GPU diagnostics flags still passing, no over-matching on `autoplay-policy`/`enable-media-session-service`). The command-line path is exercised through the shared constant rather than directly — testing `on_before_command_line_processing` needs a `CommandLine` instance, which isn't constructible outside a CEF process. Worth flagging as a coverage gap rather than implying it's tested.

AgentY-asaf:
@codex — correct diagnosis, and already fixed in `3ab946aaa` (pushed while your review was in flight; your comment anchors to the env-var block from the earlier commit).

The fix is exactly what you prescribed: `on_before_command_line_processing` now checks `has_switch` and calls `CommandLine::remove_switch` on the assembled command line before anything else reads it, with the same loud warning. Verified on merged `main` — `app/mod.rs:532-535`.

Both ingresses share one `MEDIA_PERMISSION_SWITCHES` constant (`:1170`) specifically so they can't drift; a switch blocked on one route and allowed on the other would read as protected while not being, which is worse than an obvious hole.

Worth recording that you and I converged on this independently: I caught the same gap while correcting the launcher citation in §3.8 — noticing that `main.rs:309` → `.args(args)` meant my env-var-only guard covered one of two documented paths. Half a guard on a *bypass* is arguably worse than none, since the spec would have claimed a boundary that didn't exist.

Remaining known gap, stated in the PR body rather than implied away: the command-line path is exercised only through the shared constant, not directly — testing `on_before_command_line_processing` needs a `CommandLine`, which isn't constructible outside a CEF process.

AgentY-asaf:
## End-to-end verification — the guard works

I flagged in this PR that the command-line path had no coverage, and on #2890 recorded that it can't be unit-tested (`cef::command_line_create()` aborts the test process rather than returning `None`). So I built a portable off merged `main` and exercised it for real.

**Build:** `0.55.29+g0e2d9a611` — `0e2d9a611` is the merge commit containing both guards.

**Test:** launched `agentmux.exe --enable-media-stream`, i.e. the exact ingress this PR closes and the one the env-var guard alone would have missed.

**Result** — from that instance's host log:

```json
{"level":"WARN","target":"agentmux_cef::app",
 "message":"removed media-permission switch from the command line — it would bypass
  the CEF permission handler and grant camera/mic to every browser pane with no prompt",
 "switch":"enable-media-stream"}
```

The switch is detected and removed at `on_before_command_line_processing`, before anything downstream reads the command line, and the warning fires as designed. The path that had zero automated coverage is now confirmed working against a real CEF runtime.

**Scope of what this proves:** the command-line ingress and the warning. The env-var ingress remains covered by the unit tests, and the two share one `MEDIA_PERMISSION_SWITCHES` constant so they can't diverge. What is *not* proven here is the downstream consequence — that a browser pane subsequently denies `getUserMedia` — since that needs a page actually calling it. Given the switch is gone before CEF parses it, that follows from CEF's documented Alloy default, but I'd rather say what was observed than imply more.

No cleanup needed: the test instance exited on its own, and no other running instance was touched.

PR #2891, opened 2026-09-01
chore: release v0.55.30

AgentY-asaf:
Forced patch release: `task release -- --as patch`, 0.55.29 → **0.55.30**.

## The override, stated plainly

26 changesets were pending: **23 patch and 3 minor**. The computed bump would have been **minor (0.56.0)**; `--as patch` forces 0.55.30, and the script printed its intended warning:

```
WARNING: forcing a LOWER bump than the changesets request — some
         minor-level change(s) will ship under a patch version.
```

That's the requested behaviour, not an accident. The three minor-typed changesets still ship and are still changelogged — they just land under a patch version. The feature work riding along:

- `feat(muxspect): layout command` — inspect the persisted pane tree, run the layout doctor
- `feat(swarm)`: each agent's long-running tool calls as a row bucket
- `feat(term)`: configurable scroll-wheel sensitivity
- `feat(agent-pane)`: per-node shrink attribution in the scroll model

## Release-consistency invariant

Verified independently of the script's own check, since this is the reagent gate:

| Location | Version |
|---|---|
| `package.json` | 0.55.30 |
| `Cargo.toml` `[workspace.package]` | 0.55.30 |
| `Cargo.lock` (`agentmux-cef`) | 0.55.30 |
| `package-lock.json` root | 0.55.30 |
| `VERSION_HISTORY.md` head | `## 0.55.30 — 2026-08-31` |

All five agree.

## Changesets consumed

26 changeset files deleted, 26 entries added to `VERSION_HISTORY.md` — counted, not assumed. `.changesets/README.md` correctly remains (it's the directory's own documentation, not a changeset).

## What's in it

Mostly this session's work — the tab-close flash resolution and its systemic analysis, the transient-failure retry ladder widening, the camera-access spec, the `--enable-media-stream` ingress fix (verified end-to-end on a real build), agent-failure logging — plus layout/pane fixes, per-channel login enforcement, and the container-agent stream-json fix from other agents.

### DIFF
--- a/.changesets/1788132600-fix-agent-pane-peek-panel-shrinkwrap-right-ampm-w7t3.md
+++ b/.changesets/1788132600-fix-agent-pane-peek-panel-shrinkwrap-right-ampm-w7t3.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(agent-pane): peek panel shrink-wraps, floats right, and shows 12-hour AM/PM time
--- a/.changesets/1788151399-fix-identity-stop-isolated-claude-code-config-dirs-leaking-t-32yz.md
+++ b/.changesets/1788151399-fix-identity-stop-isolated-claude-code-config-dirs-leaking-t-32yz.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(identity): stop isolated Claude Code config dirs leaking the host's ~/.claude/CLAUDE.md
--- a/.changesets/1788153000-fix-layout-handle-spans-slipped-pane-r2v9.md
+++ b/.changesets/1788153000-fix-layout-handle-spans-slipped-pane-r2v9.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(layout): restore the resize handle between two panes flanking a minimized one
--- a/.changesets/1788154400-feat-muxspect-layout-command-k8p4.md
+++ b/.changesets/1788154400-feat-muxspect-layout-command-k8p4.md
@@ -1,5 +1,0 @@
----
-type: minor
----
-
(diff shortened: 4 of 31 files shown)


a5af:
@codex review

AgentY-asaf:
Good catch — fixed in `e5e77686c`, and the history shows exactly how it drifted:

```
cd079181a  fix(auth): ... close FOUR paths to Claude without one   (original)
5cf65bf46  fix(auth): close a fifth bypass (Codex P1)              (found in review)
f0125cd28  fix(auth): ... close FIVE paths to Claude without one   (merged, #2878)
```

The changeset file was authored at the four-path stage and never revisited when review found the fifth, so the stale count carried through verbatim into `VERSION_HISTORY.md`. Worth correcting before it ships rather than after — a release bakes the changelog permanently, and this is a **security-relevant entry undercounting what was actually fixed**.

Re-verified after the edit: all five version locations still agree at 0.55.30, and the section still holds 26 entries for 26 consumed changesets.

**A general hazard this exposes**, beyond this one line: changesets are authored at PR-open time and are never revisited when review changes the PR's scope. Nothing catches that drift except a human (or reviewer) reading the release diff — which is exactly what happened here. Every PR whose scope grows during review is a candidate for the same silent staleness. Might be worth a release-time check that each consumed changeset's summary still matches its merged commit subject, since both are available at that point.

---

Separately, for the record: the `[CI] PR #2891 CI PASSED` notification that arrived before your review was **premature** — actual state at that moment was 1 of 4 checks complete (`verify` only; both Rust matrix legs and `vitest` still running), and `reviewDecision: REVIEW_REQUIRED`. That's the second false CI-passed signal on a release PR in this session. Not merging on it.

PR #2892, opened 2026-09-01
docs(report): where a universal agent work queue belongs, and what to call it

AgentX-asaf:
## Ask

> agentmux needs a universal queue of stuff that any agent can set, and any agent can pick up when the time comes ... considering all our current pane types, where would this be? ... would this be something managed through the drone pane? swarm? a new one?

Analysis only — no code.

## Three findings

**1. It's a new backend primitive, not an existing pane's feature.** Nothing today models *unassigned* work. Drone is edge-driven (the taker is decided at authoring time), Swarm is live-only, jekt is addressed push. The ask is pull, unaddressed, deferred — a different primitive.

**2. Not Drone, not Swarm — but both should integrate.** Swarm's dispatch state is already documented as dying with its block (`REPORT_AGENT_PICKER_..._2026_08_24.md` §5a). Hosting a durable backlog there repeats a bug class this repo has paid for twice. Drone fits as a **producer** (a node that enqueues); Swarm fits as a **viewer of in-flight claims**. Neither should own the store.

**3. The build is far smaller than the ask implies.** `db_cron_jobs` already *is* this row — prompt + target agent + created_by + fire tracking — with a **time** trigger. Swap the trigger for **readiness** and reuse cron's own delivery (POST `/agentmux/reactive/inject`), which inherits Tier 2a/2b/3 routing including same-host cross-channel. `db_agent_groups` gives "any agent of this kind" routing for free.

## The two things that can actually go wrong

- **Leases, not locks.** Copy `db_background_tasks`' heartbeat + a reaper. A bare `claimed` boolean reproduces the stuck-`running`-forever incident class (#2518).
- **Scope.** The queue **must** be global. A per-channel queue means "any agent *in this channel*" — nearly useless when every local build is its own channel. This is the same seam where `SPEC_IDENTITY_STORE_SPLIT_2026_08_17` left 5 of 6 steps unbuilt (#2627 closed the day step 1 landed) and where today's five auth bypasses originated. `db_cron_jobs` is named in that spec's own audit as having no legitimate reason to be per-channel, still unmigrated.

## Recommendation

Ship **backend + MCP tools + `muxspect work`** first, no pane. Cloud sync explicitly out of v1 — cross-host *claiming* is much harder than cross-host *delivery*, and the local semantics should be proven first.

**Names:** **Muxqueue** as the system name (fits muxbus/muxlog/muxspect), **Docket** as the pane label if one ships — same split precedent as `db_bundles` vs "Armory Bundle Format". Avoid **Relay**; it collides with muxbus/jekt.

Four open questions for the operator in §9, incl. whether claiming an item conveys authority (recommend no) and whether cross-channel claiming should be allowed given isolation was just deliberately tightened.

### DIFF
--- a/docs/reports/REPORT_UNIVERSAL_AGENT_WORK_QUEUE_2026_09_01.md
+++ b/docs/reports/REPORT_UNIVERSAL_AGENT_WORK_QUEUE_2026_09_01.md
@@ -1,0 +1,351 @@
+# Report — a universal agent work queue: where it belongs, and what to call it
+
+**Date:** 2026-09-01
+**Author:** AgentX
+**Status:** Analysis + design proposal. Not a commitment to build.
+**Prompted by:** operator ask —
+
+> agentmux needs a universal queue of stuff that any agent can set, and any
+> agent can pick up when the time comes. we need to build internal infra for
+> that, potentially with cloud sync, but not necessarily. considering all our
+> current pane types, where would this be? consider a couple names. would this
+> be something managed through the drone pane? swarm? a new one?
+
+---
+
+## 1. Bottom line up front
+
+**Three answers, in order of confidence:**
+
+1. **This is a new backend primitive, not a feature of an existing pane.** The
+   queue is a durable, globally-scoped store plus a claim protocol. Nothing that
+   exists today is shaped like it (§3).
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2893, opened 2026-09-01
feat(cef): resolve which pane a CEF Browser belongs to (camera Phase 1)

AgentY-asaf:
Phase 1 of `SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md` — but **not** the Phase 1 the spec described, and the difference matters.

## The spec's design was wrong

§3.3 called for threading `block_id` into the browser-pane client at construction, so the permission handler could tell which pane is asking. Two things surfaced while implementing it:

1. **There are five construction sites, not three.** `app/mod.rs`, `browser_pane/creation.rs`, `creation_views.rs`, and `floating_pane.rs` **twice**.

2. **One of them creates a pane with no identity.** `CreatePanePoolWindowWin32Task` builds a *prewarmed pool pane* that only becomes a specific block's pane later, at promote.

That second point makes the spec's approach actively harmful. Changing `is_browser_pane: bool` into an `Option` carrying the id would have made pooled panes resolve as **not browser panes** — silently losing their focus handler, context menu, and permission-deny. And a field captured at construction stays `None` for exactly the panes that later matter.

I nearly shipped that before checking each site.

## What this does instead

Resolve identity **at request time**: `AppState::block_id_for_browser` walks the `browser_panes` registry, maps each block id to its label via `live_browser_pane_label`, fetches that label's `Browser`, and compares with `is_same` — the reverse of the existing `block_id → label → Browser` chain.

- No construction site changes
- Cannot regress the pooled-pane case
- Starts working the moment a pane is registered, **including a pool pane after promote**

**Only `Live` panes match.** A `Closing` pane must not resolve, or a permission decision could be attributed to a pane that's going away. Returns `None` for the main client, an unpromoted pool pane, a pane mid-teardown, and any non-pane browser — all of which callers must treat as *"no pane-scoped grant applies"*, never as *"allow"*.

## Scope

**No behaviour change — nothing calls this yet.** Phase 2 (handler + grant store + prompt) is the consumer. Landing it separately keeps the pooled-pane reasoning reviewable on its own rather than buried in the feature diff.

## Test plan

- `cargo test -p agentmux-cef` — 290/290 pass

The `unused imports` warning in `state/mod.rs` is **pre-existing on `main`** — confirmed by stashing and re-checking, not assumed.

Not unit-tested directly: the lookup needs live CEF `Browser` handles, which aren't constructible under `cargo test` (same constraint recorded in #2890). Its correctness rests on the registry chain it mirrors, which is exercised throughout the pane code.

### DIFF
--- a/.changesets/1788271781-feat-cef-resolve-which-browser-pane-a-cef-browser-belongs-to-ok7h.md
+++ b/.changesets/1788271781-feat-cef-resolve-which-browser-pane-a-cef-browser-belongs-to-ok7h.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(cef): resolve which browser pane a CEF Browser belongs to (camera Phase 1)
--- a/agentmux-cef/src/state/mod.rs
+++ b/agentmux-cef/src/state/mod.rs
@@ -6,7 +6,7 @@
 use std::collections::HashMap;
 use parking_lot::Mutex;
 
-use cef::{Browser, Frame};
+use cef::{Browser, Frame, ImplBrowser};
 
 mod drag;
 mod window_meta;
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2894, opened 2026-09-01
Manoz@Area54: fix(identity): make provider auth-dir isolation correct by construction, with measured evidence

GenericAgentX-asaf:
## Summary

Follow-up to #2854, prompted by the operator asking the sharper question that spec couldn't answer: *"is the global claude completely isolated?"*

#2854 shipped on an **inferred** mechanism. This PR measures it, and measures the surfaces #2854 never checked.

### What was measured

Three-arm experiment per surface (control / treatment / **sensitivity**), from a cwd with a verified-clean `CLAUDE.md` ancestry:

| Surface | Isolated by `CLAUDE_CONFIG_DIR` alone? | Evidence |
|---|---|---|
| `CLAUDE.md` | **NO — leaks** | control returned `# Global Claude Code Rules` (host file); treatment returned `NONE` |
| `settings.json` | **YES** | host `env.BASH_ENV` absent in control; config-dir sentinel applied |
| `skills/` | **YES** | host `build`/`deploy`/`pr` absent; config-dir `zorbtest` appeared |
| `.claude.json` / `plugins/` | no exposure on this machine | zero MCP servers defined; sole plugin project-scoped elsewhere |
| Other providers | **untested, not currently exposed** | no host `AGENTS.md`/`GEMINI.md`/`QWEN.md` exists to leak |

**#2854's premise was correct, and `CLAUDE.md` is the only confirmed leak** — which is exactly why it was easy to miss: the boundary already works for credentials, settings and skills, so nothing else pointed at it.

Worth noting `settings.json` was the highest-stakes unknown going in — on this machine it carries `env.BASH_ENV`, a `PostToolUse` hook shelling into an unrelated repo, and `skipDangerousModePermissionPrompt: true`. Confirmed isolated.

### Hardening

`agent_open.rs` did `create_dir_all` and the isolation seed as two independently-deletable statements, and **the seed call had no test at all** — deleting that one line would have silently reopened the leak against a fully green suite. Both are now fused into `providers::prepare_provider_auth_dir()`: you cannot obtain a usable auth dir without its isolation guarantees, and removing the isolation now means removing the directory creation too, which fails loudly.

### Deliberately NOT done

Extending the seed generically to all providers. `ProviderConfig::startup_instructions_filename` describes the file written into the agent's *working directory* (project level), which isn't necessarily the provider's *user-config-level* filename — reusing it for the config-dir placeholder would conflate two concepts on a guess. Closing that properly needs the same three-arm method per provider, which requires each CLI installed and authed (`codex`/`qwen`/`openclaw` aren't). Documented as open in the report.

### One methodological note

My first pass ran only arms 1–2, phrased *"do your system instructions contain X?"*, and got `NO` from every arm — which reads as "no leak, fix unnecessary." That would have been the wrong conclusion: memory arrives as a user-turn reminder, not literal system instructions, so the model answered truthfully and uselessly. Rephrasing separated the arms immediately. Recorded in the report so a re-test doesn't repeat it.

Full evidence: `docs/reports/REPORT_CLAUDE_CONFIG_DIR_ISOLATION_EVIDENCE_2026_09_01.md`

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — **2890 passed, 0 failed**
- [x] 25 isolation-specific tests pass (21 existing + 4 new for `prepare_provider_auth_dir`: creation+seed together, non-claude no-op, idempotence/never-clobber, empty-path rejection)
- [x] Live experiments run against real isolated dirs; all temp dirs and copied credentials cleaned up afterward

### DIFF
--- a/.changesets/1788272085-fix-identity-make-provider-auth-dir-isolation-correct-by-con-8lg7.md
+++ b/.changesets/1788272085-fix-identity-make-provider-auth-dir-isolation-correct-by-con-8lg7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(identity): make provider auth-dir isolation correct by construction
--- a/agentmux-srv/src/backend/providers.rs
+++ b/agentmux-srv/src/backend/providers.rs
@@ -700,6 +700,35 @@ pub fn seed_claude_md_placeholder_if_missing(
     Ok(true)
 }
 
+/// Prepare an isolated provider auth/config directory for use as
+/// `CLAUDE_CONFIG_DIR` (or a provider's equivalent): create it, then apply
+/// every isolation guarantee that directory needs before a CLI is pointed
+/// at it.
+///
+/// Exists so callers cannot obtain a usable auth dir WITHOUT its isolation
+/// guarantees — the spawn paths call this instead of `create_dir_all` +
+/// a separately-skippable seed step. Deleting the isolation from a spawn
+/// path now means deleting the directory preparation too, which fails
+/// loudly instead of silently reopening the leak. See
+/// `docs/specs/SPEC_ISOLATE_HOST_CLAUDE_MD_2026_08_31.md` and
+/// `docs/reports/REPORT_CLAUDE_CONFIG_DIR_ISOLATION_EVIDENCE_2026_09_01.md`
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #2895, opened 2026-09-01
feat(cef): per-pane per-origin media grant store (camera Phase 2a)

AgentY-asaf:
Phase 2a of `SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md` — pure state, no CEF, no policy. It records what was decided and is deliberately incapable of inventing a grant.

## The reuse rule is the whole design

It follows directly from CEF's exact-match requirement: `allowed_permissions` must **match** `required_permissions`, so audio and video cannot be decided independently for one request. A grant is therefore a statement about a specific **combination** of devices, not about "camera" as a standalone permission.

So `covers()` = every requested bit already granted:

| Granted | Requested | Result |
|---|---|---|
| `{audio, video}` | `{video}` | **covered** — already answered, no prompt (subset) |
| `{audio}` | `{audio, video}` | **not covered** — re-prompt (superset) |
| `{audio}` | `{desktop_audio}` | **not covered** (disjoint) |

The superset case is the one that matters: audio was allowed, now the page wants the camera too. That's a new question, and a naive "is there a grant for this origin?" check would silently allow it.

## Scope of a grant

Keyed by `(block_id, origin)` — **pane-scoped** because two panes on the same site are two independent trust decisions, **origin-scoped** because anything finer re-prompts on every in-site navigation, and **session-only** because persisting a camera grant is a materially different threat model that wants its own decision.

## Details that earned their tests

- **An empty request is never covered.** Returning `true` for `0` would let a malformed request through the gate that's supposed to be the narrow one.
- **`grant()` replaces rather than accumulates**, so a user re-prompted after a revoke can't end up holding more than they just agreed to.
- **Granting `0` removes the entry** — "allowed nothing" and "no grant" must not be two spellings of the same state.
- **`clear_pane` on close**, so grants can't outlive their pane or be inherited if a block id were ever reused.
- **Not `Clone`, deliberately** — a snapshot could answer "granted?" after a revoke invalidated it, and a stale *allow* is the one error this type must not make possible.

`revoke()` is documented as **insufficient on its own**: it stops future requests but cannot stop a capture already in flight, which needs a pane reload (§3.7). Naming that at the definition, since "revoke" reads like it should be complete.

## Scope

**No behaviour change — nothing constructs this yet.** Phase 2b (the permission handler) is the consumer. Landing the policy logic separately keeps the subset/superset reasoning reviewable on its own.

## Test plan

- 13 new unit tests covering subset/superset/disjoint, cross-pane and cross-origin isolation (including that origin matching isn't a prefix match), replace-not-accumulate, zero handling, and both clear paths
- `cargo test -p agentmux-cef` — **303/303 pass**

### DIFF
--- a/.changesets/1788272119-feat-cef-per-pane-per-origin-media-grant-store-camera-phase--6vk9.md
+++ b/.changesets/1788272119-feat-cef-per-pane-per-origin-media-grant-store-camera-phase--6vk9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(cef): per-pane per-origin media grant store (camera Phase 2a)
--- a/agentmux-cef/src/browser_panes/media_grants.rs
+++ b/agentmux-cef/src/browser_panes/media_grants.rs
@@ -1,0 +1,268 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Per-pane, per-origin media capture grants.
+//!
+//! Spec: `docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md` §3.1-3.2.
+//!
+//! # Why the grant records a bitmask rather than a set of booleans
+//!
+//! CEF requires `allowed_permissions` to **match** `required_permissions` for a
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

Issue #1591, opened 2026-06-20
Tracking: server-side / high-accuracy speech-to-text (voice input Phase 4)


a5af:
## Tracking: server-side / high-accuracy speech-to-text (voice input Phase 4)

**Consolidates** the scattered voice-input tracking: closed issue #740 (and its never-filed Phase 2/3 follow-up), the deferred Phase 4 of `docs/specs/SPEC_VOICE_INPUT_PER_PANE_2026_05_19.md`, and the superseded `specs/SPEC_VOICE_INPUT_2026_05_08.md`. Use **this** issue as the single source of truth going forward.

---

### What's already shipped (Web Speech API)

Per-pane voice input is **live** end-to-end on the browser-native engine — no further UI work needed:

- `frontend/app/hook/useVoiceInput.ts` — singleton `webkitSpeechRecognition` session, `PaneVoiceHandle` registry, per-pane retargeting
- `frontend/app/element/MicButton.tsx` — mic button in the shared block frame header (`blockframe.tsx`), so it works for **both** terminal and agent panes; gated on `voice:enabled`
- `agent-model.ts` + `termViewModel.ts` expose `voiceHandle`; agent appends to the textarea with interim preview, terminal streams chars to the PTY via `blockinput`
- `Ctrl+Shift+V` toggles the focused pane; permission-error toast in `app-init.ts`
- Shipped across PRs #920 (building blocks), #930 (frame-header integration), #931 (polish)

**Strengths:** free, streaming with interim results, zero infra, no API key, Chromium-native (and AgentMux *is* Chromium/CEF). **The only weakness:** browser-engine accuracy is weak on code identifiers, technical vocabulary, jargon, and proper nouns. That weakness is the sole motivation for everything below.

---

### What's still open: server-side STT (this issue)

The remaining work is a **pluggable, higher-accuracy / offline transcription backend** routed through `agentmux-srv`, selectable in settings, leaving Web Speech API as the default. This is #740's Phase 2 (provider STT) + Phase 3 (local whisper.cpp), now merged into the per-pane spec's Phase 4.

---

#### There is no "Anthropic / Claude" transcription path
Verified against the Claude API reference: the Messages API input modalities are **text, images, and PDF/documents only**. There is **no audio content block and no transcription endpoint**, and the code-execution sandbox is network-isolated (no Whisper). Claude cannot turn speech into text. The `~~Claude multimodal~~` row in #740's table stays ruled out.

[...]

PR #2897, opened 2026-09-01
feat(cef): grant-store-driven permission handler for browser panes (camera Phase 2b)

AgentY-asaf:
Browser panes get a media permission handler for the first time. It consults the Phase 2a grant store and denies when nothing covers the request.

## Behaviour is unchanged — verified, not asserted

- **Nothing anywhere creates a grant yet** — `grep` finds no `.grant()` call sites outside the store's own tests. So `covers()` is always `false`.
- Therefore every request takes the deny path, which is exactly what CEF's Alloy default did when `permission_handler()` returned `None` for panes.

What changes is that the denial is now **explicit, attributed to a pane, logged**, and has somewhere for the Phase 2c prompt to go.

## Separate handler, deliberately

Kept as its own type rather than branching inside the existing one:

- **Main client** — AgentMux's own first-party UI, unconditionally grants the audio bits for voice input (#1591/#1602). **That code is untouched by this diff** (confirmed: the change is additive, no removed lines in the audio path).
- **Browser pane** — arbitrary web content, so the answer depends on what the user allowed for that pane and origin.

Branching inside one handler would have put the shipped voice path one editing mistake away from this work.

## Decision order — every step a denial, never a fallthrough

1. **No callback** → return `0` (CEF default deny). Returning `1` would leave the request dangling forever.
2. **Browser doesn't resolve to a pane** → deny. Covers the main client, an unpromoted pool pane, a pane mid-teardown, and any non-pane browser. All mean *"no pane-scoped grant applies"*, which is never an allow.
3. **Pane no longer `Live`** → deny. This discharges the obligation `block_id_for_browser`'s docs impose after #2896: it snapshots under a lock and drops it before its FFI comparison, so callers must re-check liveness at point of use rather than treat the returned id as proof.
4. **Grant covers the request** → echo `requested_permissions` **exactly**. Narrowing would deny the whole request under CEF's match rule, not grant a subset.

## Grants die with their pane

Cleanup on close, placed *after* the `Live→Closing` transition fires so it runs exactly once per real close (a `None` there means another close already owns the teardown). Grants are pane-scoped precisely so one pane's decision can't apply to another; block ids aren't reused in practice, but a stale entry would silently hand a new pane the previous occupant's camera grant.

## Test plan

- `cargo test -p agentmux-cef` — 303/303 pass

The handler itself isn't unit-testable (needs live CEF `Browser` handles — same constraint as #2890); its policy logic lives in the grant store, which has 13 tests. End-to-end verification comes with Phase 2c, when a grant can actually exist.

### DIFF
--- a/.changesets/1788272736-feat-cef-grant-store-driven-permission-handler-for-browser-p-neyb.md
+++ b/.changesets/1788272736-feat-cef-grant-store-driven-permission-handler-for-browser-p-neyb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(cef): grant-store-driven permission handler for browser panes (camera Phase 2b)
--- a/agentmux-cef/src/browser_panes/close.rs
+++ b/agentmux-cef/src/browser_panes/close.rs
@@ -65,6 +65,18 @@ impl BrowserPaneManager {
             Some(l) => l,
             None => return,
         };
+
+        // Drop this pane's media grants. Placed after the Live→Closing
+        // transition fired (a None above means "missing or already closing" —
+        // another close is mid-teardown and owns the cleanup), so it runs
+        // exactly once per real close.
+        //
+        // Grants must not outlive their pane: they are pane-scoped precisely so
+        // that a decision made for one pane cannot apply to another, and block
+        // ids are not reused in practice — but if one ever were, a stale entry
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2898, opened 2026-09-01
feat(muxqueue): universal agent work queue — store + claim protocol (slice 1)

AgentX-asaf:
Slice 1 of the design in `docs/reports/REPORT_UNIVERSAL_AGENT_WORK_QUEUE_2026_09_01.md` (#2892, merged): **the durable store and the claim protocol only.** No RPC, no MCP tools, no pane — those are slices 2/3, per that report's own sequencing.

## What this is

Muxqueue is the **pull / unaddressed / deferred** counterpart to jekt's **push / addressed / immediate** delivery. One row = one unit of work *any* agent may claim, rather than a message aimed at a named recipient right now.

Shaped deliberately as `db_cron_jobs`' sibling — cron is this row with a **time** trigger, this is the same row with a **readiness** trigger, and `not_before` makes that relationship literal. It sits directly below cron in the schema for that reason.

## Where it lives, and why that mattered

**The always-global identity store, not the per-channel one.** A per-channel queue only ever means "any agent *in this channel* can pick it up", which is close to useless when every local/dev/portable build is its own channel (report §6.2).

Worth flagging: my first pass put the DDL in `run_shared_store_schema` by mistake — the exact wrong-store error §6.2 warns about. Caught by the tests failing with `no such table`, then verified by asserting which schema function the DDL actually lands in. `IDENTITY_STORE_SCHEMA_VERSION` 4 → 5.

## Claiming is a lease, not a lock

- The claim is **one conditional `UPDATE ... RETURNING`**, so SQLite's writer serialization makes it exactly-once. A SELECT-then-UPDATE refactor would reintroduce a read-modify-write race that single-threaded tests cannot see — a test pins the behaviour and the doc comment says explicitly not to do it.
- `claim_expires` + heartbeat, copying `db_background_tasks`' `last_seen_ms` pattern. A bare `claimed` boolean reproduces the stuck-`running`-forever class (#2518), where a row whose owner vanished is indistinguishable from one being actively worked.
- `work_queue_reap` parks poison items as `failed` **before** reopening expired ones. Reversed, an already-exhausted row gets handed out for one more doomed round-trip — a test pins that ordering specifically.
- A voluntary `release` still consumes an attempt, so a hot-potato item can't cycle forever.

## Deliberate boundaries

- **Group eligibility is resolved by the caller** and passed in. `db_agent_groups` lives in the per-channel store; this module does not reach across stores. Group ids are **bound as parameters**, never interpolated.
- Every SQL state literal is built from the `work_state` constants via `format!`, so the constants can't drift from the queries using them.

## Tests — 16, all on the claim protocol

Claim exclusivity, priority/FIFO ordering, `not_before` boundaries, agent- and group-targeting (including the distinct no-groups SQL path, which builds different SQL), kind filtering, holder-only heartbeat/complete/release, lease expiry at the exact boundary, poison-item parking, cancel from either state, duplicate-id rejection.

`cargo test -p agentmux-srv` — **2902 passed**.

## Known, intentional

Two `never constructed` warnings for `WorkItem`/`ClaimFilter` — nothing calls the store until the RPC layer lands in slice 2. Not silenced with `#[allow(dead_code)]`, which would then need removing.

### DIFF
--- a/.changesets/1788272778-feat-muxqueue-universal-agent-work-queue-store-claim-protoco-mqe4.md
+++ b/.changesets/1788272778-feat-muxqueue-universal-agent-work-queue-store-claim-protoco-mqe4.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxqueue): universal agent work queue — store + claim protocol
--- a/agentmux-srv/src/backend/storage/migrations.rs
+++ b/agentmux-srv/src/backend/storage/migrations.rs
@@ -1325,7 +1325,20 @@ pub fn run_shared_store_schema(conn: &Connection) -> Result<(), StoreError> {
 ///        forever, silently defeating check_schema_compat/stamp_version's
 ///        forward-compat lock for this one store (reagent P1, PR #2782).
 ///        See docs/specs/SPEC_GLOBAL_MEMORY_SYSTEM_TIER_2026_08_24.md.
-pub const IDENTITY_STORE_SCHEMA_VERSION: i64 = 4;
+/// v5: `db_work_queue` — the Muxqueue backlog
+///        (`docs/reports/REPORT_UNIVERSAL_AGENT_WORK_QUEUE_2026_09_01.md`).
+///        A NEW table, so no ALTER is involved, but the counter still bumps:
+///        an identity-store.db stamped v4 that a v5 binary opens must be
+///        recognized as older, not "already current" (same reagent P1 on
+///        PR #2782 reasoning recorded for v4 above).
+///
+///        Deliberately in THIS store, not the per-channel one. A queue whose
+///        rows are per-channel means "any agent IN THIS CHANNEL can pick it
+///        up", which is close to useless given every local/dev/portable build
+///        is its own channel — see the report's §6.2, and
+///        `docs/analysis/ANALYSIS_PER_CHANNEL_AUTH_BYPASSES_2026_08_31.md`
+///        for what the per-channel/global seam has already cost once.
+pub const IDENTITY_STORE_SCHEMA_VERSION: i64 = 5;
 
 /// Initialize (or re-validate) the `~/.agentmux/shared/identity-store.db`
 /// schema — the permanently-global store introduced by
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentX-asaf:
Both Codex findings fixed in `5f7546406`, plus reagent's. **Both were real correctness bugs, and the P1 is one I would not have found by inspection** — thank you.

## Codex P1 — the ABA

Exactly right, and worth spelling out because the shape is subtle: `agent_id` is *stable across claims*, so my holder-and-state predicate could not tell two claims by the same agent apart. After `expire → reap → the same agent reclaims`, a delayed call from the **first** claim would happily complete the **second** one with a stale result, extend its lease, or release it out from under the live worker.

Fixed with an `attempt` fence on every holder mutation. `attempts` already increments on each claim, so it separates claim instances with **no extra column and no new concept** — the generation number was already sitting there.

The regression test drives the full sequence (claim → expire → reap → *same agent* reclaims) and asserts all three stale operations are refused while the current fence still works.

## Codex P2 — release on the final attempt

Also right, and it defeated the attempt bound *specifically for the case the bound exists for*. `release` returned to `open` unconditionally; the claim query had no attempts predicate; and the reaper can't park a released row because it's no longer `claimed`. So a hot-potato item cycles forever. Release now parks an exhausted item as `failed`.

I also added `attempts < max_attempts` to the claim query itself. Release and reap both park exhausted rows, so that's defence in depth — a future path that leaves a spent row `open` now fails closed instead of becoming an infinite loop. Separate test for it.

## reagent P2 — the false invariant

Correct: I wrote "no apostrophes or quotes in this block" and the block is full of apostrophes. Worth restating *precisely* rather than just deleting, because the failure mode is a build error that points at prose instead of code: an **unescaped double quote** terminates the Rust string literal the whole schema lives in. Escaped ones are fine and already appear above (`db_drone_definitions`' default JSON). Apostrophes are irrelevant to both Rust and SQLite — SQLite doesn't lex quotes inside a `--` comment. I verified the escaped-quote claim by counting them in the schema functions rather than asserting it.

## Still expected at this slice

The `never used` warnings (`WorkItem`, `ClaimFilter`, and the `work_state` constants) are transitive dead-code — the store's methods have no non-test callers until the RPC layer lands in slice 2. Deliberately not silenced with `#[allow(dead_code)]`, which would then need removing.

**Tests:** agentmux-srv **2905 passed**, 19 in `work_queue`.

PR #2899, opened 2026-09-01
feat(browser-pane): camera/mic prompt for browser panes (camera Phase 2c)

AgentY-asaf:
**The first phase that changes behaviour.** A browser pane can now obtain camera or microphone access — but only after the user explicitly allows it. Before this, `getUserMedia` in a pane failed with `NotAllowedError` unconditionally (#2871).

Flow: pane requests → no grant covers it → **the callback is parked** → the main window prompts → user answers → grant recorded, callback continued with exactly the bits requested.

Retaining the callback is what Phase 0 (#2885) verified CEF permits: *"call `CefMediaAccessCallback` methods either in this method **or at a later time**."*

## The threading constraint drove the design

`MediaAccessCallback` is a `RefGuard` with **no `Send`/`Sync`**, and must be used on the CEF UI thread. So it cannot live in `AppState`. Pending requests live in a **UI-thread-local registry**, and every mutation routes through a posted task; `debug_assert`s pin that.

## The invariant: every parked request resolves exactly once, and always eventually

- **`resolve()` removes the entry *before* continuing the callback.** A user answer racing the timeout therefore can't continue the same callback twice — that would be a use-after-continue into CEF.
- **Every park arms a 60s timeout.** A prompt that never renders (window closed, renderer crashed, event dropped) degrades to *denial* rather than hanging the page's `getUserMedia` forever.
- **Both teardown paths cancel a pane's pending prompts.**

## Also closes the drain-path gap Codex found on #2897

`drain_closed_label` is a **second** teardown path — CEF-initiated, on crash or a pane closing itself — that removes the reducer entry without going through `close()`. So `close()`'s cleanup never runs for it, and a later `close()` returns early because the entry is gone.

Grants left behind there would be inherited by a recreated pane with the same block id — and **`replay_pending_create` runs in that very block**, so the recreation is immediate. Cleared *before* the replay, not after.

## Prompt design — deny-biased throughout

- **Rendered in the main window's DOM, never the pane's.** The requesting page is arbitrary web content on a native surface; a prompt inside it could be drawn or clicked by the page itself.
- **Titled with the site and the exact devices** — *"example.com wants to use your camera and microphone"* — attributed to the page, not to AgentMux. A prompt that reads like an app prompt trains people to accept them.
- **`destructive`**: initial focus on "Don't allow"; Escape and backdrop deny.
- **Unknown permission bits surface as "other media devices"** rather than being dropped. If CEF adds a bit we don't know, under-reporting what a page asked for is the worst possible failure here.
- **Origin shown as host only**, so a crafted long path can't push the real origin out of view (tested with `https://evil.test/https://bank.example.com/login` → `evil.test`).
- **A second request while one is open denies the displaced one** rather than silently swapping the question under the user's cursor.

## Test plan

- `cargo test -p agentmux-cef` — 303/303
- `npx vitest run frontend/app/window frontend/app/store` — 799/799 (10 new, covering the device-naming and origin-display security surface)
- `tsc --noEmit` — clean

**Not yet verified end-to-end.** The parked-callback path needs a real `getUserMedia` page in a running build. I'm building a portable to exercise it and will report the result here before this merges — reviewer approval alone shouldn't land an async CEF-callback lifetime change unobserved.

### DIFF
--- a/.changesets/1788273361-feat-browser-pane-camera-mic-access-with-a-per-pane-permissi-ou8x.md
+++ b/.changesets/1788273361-feat-browser-pane-camera-mic-access-with-a-per-pane-permissi-ou8x.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(browser-pane): camera/mic access with a per-pane permission prompt (camera Phase 2c)
--- a/agentmux-cef/src/browser_panes/close.rs
+++ b/agentmux-cef/src/browser_panes/close.rs
@@ -77,6 +77,7 @@ impl BrowserPaneManager {
         // would silently hand a new pane the previous occupant's camera grant.
         // SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md §3.2.
         state.media_grants.lock().clear_pane(block_id);
+        crate::browser_panes::media_prompt::cancel_pane_any_thread(block_id);
         #[cfg(target_os = "windows")]
         {
             let ops = AppStateCloseOps(state);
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

AgentY-asaf:
## Verification: partial, and I want to be precise about the boundary

Built a portable off this branch (`0.55.30+ga8cdb5522`) and ran it.

**Verified:**
- Builds clean
- Launches and runs stably — 445 log lines, **zero `ERROR` entries, zero panics**, healthy memory heartbeat
- No regression from installing a permission handler on **every** browser pane, which is the change with the broadest blast radius here (previously panes had none at all)

**Not verified — and this needs a human:**
- The prompt actually appearing, being answered, and the grant taking effect
- The capture indicator lighting up and "Stop" reloading the pane

That flow needs a browser pane opened, navigated to a `getUserMedia` page, and a person clicking Allow. I can't drive another instance's UI — my MCP tools are scoped to the instance I'm running inside, which is on older code. So the interactive path is **reasoned-about and unit-tested, not observed.**

What backs it in the meantime: 8 tests on the exactly-once park/resolve/cancel invariant (including the answer-races-timeout case), 13 on the grant store's subset/superset semantics, 13 on the prompt/indicator copy, and Phase 0's verification that CEF permits late continuation.

Merging on approval per standing instruction — but flagging plainly that **first real use is the first real test** of the prompt round-trip. If a pane's `getUserMedia` hangs for 60s and then denies, the likely cause is the prompt event not reaching its window; the `pane-media` log target traces every step of the decision.

PR #2900, opened 2026-09-01
Manoz@Area54: feat(armory): drop the host CLI config block from Global Memory

GenericAgentX-asaf:
## Summary

Armory → Memory → Global surfaced `~/.claude/CLAUDE.md` as a read-only "Claude Code — host CLI config" block. Removing it.

That block existed because it was once genuinely unclear which file a spawned agent read. **That ambiguity is now settled** — the three-arm experiment in `REPORT_CLAUDE_CONFIG_DIR_ISOLATION_EVIDENCE_2026_09_01.md` (#2894) proved a spawned agent reads `$CLAUDE_CONFIG_DIR/CLAUDE.md`, and — once seeded, which `prepare_provider_auth_dir()` now guarantees at every spawn — **never** the host file.

So the block showed a file with no bearing on any AgentMux agent. Its own tooltip already conceded this ("Not read by spawned in-app agents"). Surfacing a file in Armory whose entire description is *"this doesn't affect anything here"* is noise, and worse, invites the misreading that editing it would change agent behaviour.

## What's removed

The whole chain, not just the JSX: model signal + fetch, `GetClaudeHostConfigCommand` RPC, the backend `getclaudehostconfig` handler, and the `COMMAND_GET_CLAUDE_HOST_CONFIG` constant.

## What's kept

The **shared-provider-config block** — it shows the file agents genuinely do read, and is now *more* useful than before: post-isolation-fix it displays the AgentMux placeholder, which is a direct, legible confirmation that isolation is in effect.

Section heading reworded: "External Claude Code files" was a plural covering both blocks, and the survivor isn't "external" at all — it's AgentMux's own isolated provider dir. Now "Claude Code provider config". The load-bearing *"reference only, not part of Global Memory"* clause stays — that's what stops the block being mistaken for an editable Global Memory section.

`read_claude_global_config()` / `resolve_shared_claude_provider_dir()` and their tests stay — still serving the surviving handler.

See `docs/specs/SPEC_ARMORY_DROP_HOST_CLI_CONFIG_BLOCK_2026_09_01.md`.

## Test plan
- [x] `npx vitest run frontend/app/view/brain` — 20/20 passing
- [x] Replaced the 5 now-obsolete host-block tests with 2 regression tests asserting the block is *gone* (no host badge, host path absent from the pane, heading no longer claims multiple external files)

### DIFF
--- a/.changesets/1788273379-feat-armory-drop-the-host-cli-config-block-from-global-memor-mgtl.md
+++ b/.changesets/1788273379-feat-armory-drop-the-host-cli-config-block-from-global-memor-mgtl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(armory): drop the host CLI config block from Global Memory
--- a/agentmux-srv/src/backend/rpc_types/commands.rs
+++ b/agentmux-srv/src/backend/rpc_types/commands.rs
@@ -239,8 +239,8 @@ pub const COMMAND_DELETE_SYSTEM_MEMORY: &str = "deletesystemmemory";
 /// companion when foreign), written by `agent_config.rs`'s
 /// `write_claude_md_respecting_ownership`, a per-agent path this command
 /// does not cover. This command exists purely to surface a
-/// Claude-Code-managed reference file for operator visibility ("External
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

PR #2901, opened 2026-09-01
Manoz@Area54: fix(memory): Personal Memory is empty for any agent with a blank working_directory

GenericAgentX-asaf:
## Summary

Found while verifying (at operator request) that an agent's written memories actually show up in **Armory → Memory → Personal**. They don't — and it's not an edge case.

`MemoryList` (MCP) and the Personal Memory tab both resolve through `memory_dir_for_agent()`. For this agent it fails outright:

```
memory/list failed: HTTP 500 — {"error":"memory: agent manoz has no working directory"}
```

…while the memory files sit on disk, intact:

```
.../claude/projects/C--Users-area54--agentmux-agents-manoz-0803a/memory/
    MEMORY.md  feedback_agentmux_cdp_live_debugging.md
    feedback_merge_on_approval.md  project_composer_strip_layout_history.md
    reference_secrets_cli.md
```

## Root cause

`memory_dir_for_agent()` resolves in two stages — instance row, then registry — but the **first stage errors out instead of falling through** when the row's `working_directory` is blank, so the registry fallback directly below is never reached.

A blank `working_directory` is not an error condition. `agent_open.rs` substitutes a default (`~/.agentmux/agents/<slug>`) whenever it's blank, so such a row still describes an agent that is running and writing memories in a real directory — one the **registry** knows (`source_agents_base` + `working_dir`) and the instance row doesn't.

**Since `working_directory` is blank by default, this breaks the common case, not an edge case.**

Worth noting the failure *shape*: a hard HTTP 500, not an empty list. So Armory can't distinguish "no memories yet" from "resolution blew up" — it just looks broken.

## Fix

Treat a blank `working_directory` as "this row can't answer" rather than "the answer is no": use the instance row only when it carries a working dir, otherwise continue to the registry fallback that already exists. The genuine error is preserved — an agent resolvable by neither still gets the registry's own `not found`.

## Test plan
- [x] `blank_working_directory_falls_through_to_the_registry_rather_than_erroring` — asserts on *which* error, proving control reached the fallback rather than stopping at the guard (the exact regression)
- [x] `non_blank_working_directory_still_resolves_from_the_instance_row` — unaffected path, asserted on path components so Windows separators don't matter
- [x] `cargo test -p agentmux-srv --bin agentmux-srv -- working_directory memory_dir_for_cwd` — 6/6 passing
- [x] Originally reproduced live via the `MemoryList` MCP tool

See `docs/specs/SPEC_FIX_PERSONAL_MEMORY_EMPTY_WORKDIR_2026_09_01.md`.

### DIFF
--- a/.changesets/1788273607-fix-memory-personal-memory-empty-for-agents-with-a-blank-wor-a5qi.md
+++ b/.changesets/1788273607-fix-memory-personal-memory-empty-for-agents-with-a-blank-wor-a5qi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(memory): Personal Memory empty for agents with a blank working_directory
--- a/agentmux-srv/src/backend/storage/agents.rs
+++ b/agentmux-srv/src/backend/storage/agents.rs
@@ -192,6 +192,34 @@ pub fn default_conversation_visibility() -> String {
     "private".to_string()
 }
 
+/// The working directory `agent.open` substitutes when an
+/// `AgentDefinition`'s own `working_directory` is blank —
+/// `~/.agentmux/agents/<name-derived-slug>`.
+///
+/// Blank is the DEFAULT for a newly defined agent, so this is the path most
+/// agents actually run in. Extracted so `agent.open` (which creates and uses
+/// it) and native-memory resolution (which must find the memories written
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #2902, opened 2026-09-01
feat(muxqueue): HTTP routes + six MCP tools (slice 2)

AgentX-asaf:
Slice 2 of `REPORT_UNIVERSAL_AGENT_WORK_QUEUE_2026_09_01.md` — agents can now actually **use** the queue built in slice 1 (#2898, merged). This is what clears slice 1's dead-code warnings.

## Surface

Seven routes, auth-gated by the same middleware as cron/reactive:

```
POST   /agentmux/work                 enqueue
POST   /agentmux/work/claim           claim next eligible
POST   /agentmux/work/:id/heartbeat   extend lease
POST   /agentmux/work/:id/complete    finish
POST   /agentmux/work/:id/release     hand back
GET    /agentmux/work                 list
DELETE /agentmux/work/:id             cancel
```

Six MCP tools: `WorkEnqueue`, `WorkClaim`, `WorkHeartbeat`, `WorkComplete`, `WorkRelease`, `WorkList`.

## Decisions worth reviewing

**No background reaper task.** The claim handler reaps expired leases *first*, so a dead claimant's item is recovered by the very next agent that asks for work. The only moment a stale claim matters is when someone wants the item — that makes it self-healing with no scheduler to own, and the reap predicate is indexed. A reap failure is logged and the claim proceeds: a not-yet-reaped queue is still correct, just possibly emptier than it could be.

**`identity_store`, not `shared_store`.** Every handler uses the always-global store, unlike the cron handlers sitting directly next to them. That inconsistency is deliberate and documented at the top of the module — cron's placement is the known-unmigrated step 1b of `SPEC_IDENTITY_STORE_SPLIT_2026_08_17.md`, not a pattern to copy.

**The claim response echoes `attempt` at the top level**, not only nested inside `item`. It's a required fence on every later call (slice 1's ABA fix), and burying it is the difference between a caller that fences correctly and one that discovers the requirement from a rejected write.

**A rejected fence is 409, not 404.** The row usually still exists — it just moved on without this caller. The MCP layer translates that into the one instruction that recovers it: claim again, don't retry with the old attempt.

**Enqueue rejects `target_agent` + `target_group` together** rather than letting one silently win.

## One deliberate limitation

`WorkClaim` sends an **empty groups list**. The MCP process has no cheap way to resolve this agent's group memberships yet, so it claims only untargeted and self-targeted work. Group-targeted claiming arrives with the group lookup, not before — silently ignoring a group restriction would be worse than under-claiming.

## Tests

- `cargo test -p agentmux-srv` — **2909 passed**
- `cargo test -p agentmux-mcp` — **26 passed**
- Both crates build clean

The tool-count test caught the six additions exactly as designed (36 → 42).

Next slice: `muxspect work` for CLI observability, then a pane if the shape holds.

### DIFF
--- a/.changesets/1788274136-feat-muxqueue-http-routes-six-mcp-tools-slice-2-83ba.md
+++ b/.changesets/1788274136-feat-muxqueue-http-routes-six-mcp-tools-slice-2-83ba.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxqueue): HTTP routes + six MCP tools (slice 2)
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -464,6 +464,99 @@ const LOOP_LIST_TOOL: &str = r#"{
   }
 }"#;
 
+// ── Muxqueue — the universal agent work queue ───────────────────────────────
+// docs/reports/REPORT_UNIVERSAL_AGENT_WORK_QUEUE_2026_09_01.md
+//
+// The pull/unaddressed/deferred counterpart to SendMessage's push/addressed/
+// immediate delivery: work goes in without naming a recipient, and whichever
+// agent asks next takes it. Deliberately NOT modelled on Cron (time-triggered)
+// or Loop (repeating) — the trigger here is an agent being ready.
+
+const WORK_ENQUEUE_TOOL: &str = r#"{
+  "name": "WorkEnqueue",
+  "description": "Put a unit of work on the shared Muxqueue for ANY agent to pick up later. Use this instead of SendMessage when you do NOT need a specific agent, or need it done eventually rather than now — 'someone should repro this', 'this PR needs review when a reviewer frees up'. The item persists across pane closes, app restarts, and version/channel changes, and is visible to every agent on this machine. If you know exactly who should do it and it should happen immediately, use SendMessage instead; if it should happen on a schedule, use CronCreate. Returns the item id.",
+  "inputSchema": {
+    "type": "object",
+    "properties": {
+      "title":        { "type": "string",  "description": "Short human-scannable summary (e.g. 'repro the minimize distortion on a 3-pane cross-split')" },
+      "payload":      { "type": "string",  "description": "The full instruction injected into whichever agent claims this. Write it as a standalone prompt: the claimant has none of your conversation context." },
+      "kind":         { "type": "string",  "description": "Optional free-form tag used to filter claims (e.g. 'review', 'repro', 'triage'). Agents can claim only a kind they handle. Omit for untyped work anyone may take." },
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

AgentX-asaf:
All six fixed in `86ac28173`. Both reviewers landed on the blocking one independently, and Codex's four MCP findings share a theme worth naming: **the agent-facing contract was claiming things that weren't true.**

## reagent P1 / Codex — bodyless cancel (blocking)

Correct, and the giveaway is that the sibling route sits a few lines away in the same router: `DELETE /agentmux/cron/:id` takes no body. A caller following that adjacent convention got a 400/415 from axum's `Json` extractor *before my handler ever ran* — even though `reason` was already `#[serde(default)]`, so the default it was written to support was unreachable. Now `Option<Json<CancelRequest>>`, with a router-level test issuing the exact bodyless DELETE.

## Codex ×4 — promises the API didn't keep

- **`WorkComplete` accepted a missing `result`** while its own description calls that field the only record the work happened. Now required in the schema *and* at runtime — better to reject the call than accept a silent hole in the audit trail.
- **`WorkClaim` discarded `item.result`**, so a claimant never saw why a previous agent handed the item back — despite `WorkRelease` explicitly promising they would. Now surfaced when `attempt > 1`, framed as "read this before repeating their approach."
- **`WorkRelease` said "released back to the queue" unconditionally**, but a release on the final allowed attempt parks the item as `failed`. It was telling the caller another agent could pick up something nobody ever will. The handler now returns the resulting state — **read back from the store rather than re-deriving the attempts rule**, since duplicating that rule in two places is exactly how they drift — and the tool phrases it accordingly.
- **`WorkList` omitted `result`**, leaving that same "only record" unreachable through the only agent-facing read tool.

## Codex P2 — query encoding

Right: `state` is free-form in the schema, not a validated enum, so `&` or `#` corrupted the query. Now built via reqwest's query serializer rather than string concatenation.

## A harness gap this exposed

Writing the router tests surfaced something worth flagging on its own: `test_state` aliases `wstore`/`id_store`/`identity_store` to **one** in-memory store carrying only the *object* schema. So any handler touching an identity-store-only table (`db_work_queue`) failed with a bare "no such table" 500 **in tests while being perfectly correct in production** — the harness was quietly answering a different question than the one under test.

My first fix gave `identity_store` its own properly-schema'd store. That broke six existing tests that seed through one alias and read through another, so I backed it out for the narrower change: the single store now carries both schemas (additive — every statement is `CREATE ... IF NOT EXISTS`). Worth knowing for anyone who adds identity-store-backed handlers later.

**Tests:** agentmux-srv **2916 passed** (7 new router-level, through the real axum router), agentmux-mcp **26 passed**.

PR #2903, opened 2026-09-01
feat(muxspect): work — inspect the Muxqueue backlog from the CLI (slice 3)

AgentX-asaf:
Slice 3 of `REPORT_UNIVERSAL_AGENT_WORK_QUEUE_2026_09_01.md`, and the last piece before deciding whether a pane is warranted: **make the queue observable without one.**

```bash
muxspect work            # every item, any state
muxspect work open       # just the claimable backlog
muxspect work claimed    # what is being worked, and by whom
muxspect work --json
```

No new srv route — it reads `GET /agentmux/work` through muxspect's existing authed `apiGet`, the same way `conversation` already calls a non-muxspect route.

## Why this matters before a pane exists

**A claimed item whose lease has already lapsed still reads as `claimed`.** Nobody is working it, and nothing surfaces that on its own — reaping happens on the next claim rather than in a background sweeper (a deliberate slice-2 decision). So a queue can look busy while being entirely idle.

This flags those rows inline with `LEASE EXPIRED` and totals them at the end, instead of making the reader compare epoch timestamps by eye. That single signal is most of the value of the command.

## Two error-shape behaviours copied from `renderLayout`

Both earned the hard way by reagent's P1 on #2856:

- A whole-request failure arrives as `200 + {error}` with no `items` key. Handled explicitly — falling through to `data.items ?? []` would print "queue is empty" and exit 0, **reporting success for exactly the store failure this command exists to surface.**
- An empty result under a state filter says `no <state> items`, not "queue is empty". Conflating those is the same error class one level down: a filtered query matching nothing is a different statement from an empty queue.

The state argument is validated against the five real states rather than passed through, so `muxspect work opne` fails loudly instead of returning an empty list that reads exactly like a healthy empty queue.

## Scope caveat, documented

Unlike every other muxspect command, this one shows items from **any channel on the host**, because the queue deliberately lives in the always-global identity store. That is the Phase-1 single-instance limitation *not applying*, not a bug — called out explicitly in `MUXSPECT.md` so it doesn't read as one.

Strictly read-only, like everything in muxspect except `dock clear`.

## Tests

**40 passed** in `muxspect.test.mjs` (6 new): whole-request error vs empty, filtered-empty vs empty, field rendering, expired-lease flagging, and a live lease *not* being flagged.

### DIFF
--- a/.changesets/1788275734-feat-muxspect-work-inspect-the-muxqueue-backlog-from-the-cli-mw5v.md
+++ b/.changesets/1788275734-feat-muxspect-work-inspect-the-muxqueue-backlog-from-the-cli-mw5v.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(muxspect): work — inspect the Muxqueue backlog from the CLI
--- a/agentmux-srv/src/backend/shellintegration/muxspect.mjs
+++ b/agentmux-srv/src/backend/shellintegration/muxspect.mjs
@@ -100,7 +100,11 @@ Usage:
                                   result/reason recorded on it. Answers "what is
                                   outstanding, and is anything stuck" without
                                   claiming anything yourself — this command is
-                                  strictly read-only.
+                                  strictly read-only. Shows the 500
+                                  most-recently-updated items and says so
+                                  explicitly when there may be more, rather
+                                  than presenting a truncated view as the
+                                  whole backlog.
   muxspect help                   this message
 
 Requires $AGENTMUX_LOCAL_URL and $AGENTMUX_AUTH_KEY in the environment.
@@ -118,6 +122,12 @@ itself documents for tool-spawned subshells):
   node ~/.agentmux/shell/muxspect.mjs list
 See docs/specs/SPEC_MUXSPECT_LIVE_INTROSPECTION_TOOL_2026_08_01.md for the
 full story and the planned fix.`;
+
+/// How many queue items `muxspect work` asks for. The server clamps to 500, so
+/// this is the practical ceiling. Threaded into renderWork rather than being
+/// compared against a second hardcoded literal there — that duplication is how
+/// a truncation warning silently stops matching the request that caused it.
+const WORK_LIST_LIMIT = 500;
 
 function fail(msg) {
     console.error(`muxspect: ${msg}`);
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentX-asaf:
All four fixed in `a1b810c52`. Both reviewers landed on truncation independently; reagent's dead-branch finding is the one I'd most want on the record.

## Truncation (Codex + reagent)

Right, and the direction of the loss is what makes it matter: `work_queue_list` orders by `updated_at DESC`, so the rows that fall off the end are precisely the **oldest open work and oldest expired claims** — exactly what this command exists to surface. It now requests the server's own ceiling (500) and states plainly when a full page comes back that there may be more and what is being cut. The limit is threaded into `renderWork` rather than compared against a second literal there, since that duplication is how a truncation warning quietly stops matching the request that produced it.

## Expired ≠ coming back (Codex)

Also right. The reaper parks an item whose attempts are spent as `failed` instead of reopening it, so my blanket "returns to the pool on the next claim" was false for **the most stuck items** — the ones the command is for. Now two distinct renderings with separate counts.

**Worth naming: this is the same over-promise you caught in `WorkRelease`'s acknowledgement on #2902, which I then reproduced in the renderer one slice later.** The rule "attempts spent means parked, not reoffered" lives in the store, and every surface that narrates queue state has to *ask* it rather than assume the happy path. I've now made that mistake twice in two days on the same rule.

## Dead error branch (reagent)

This is the sharpest note on the PR. My `if (data.error)` branch was copied from `renderLayout`, whose handler deliberately returns 200-with-`{error}`. `/agentmux/work` does not — `handle_work_list` returns HTTP 500, which `apiGet` already turns into `fail()` + exit(1) while printing the server's own message. So the branch was unreachable via the real endpoint, **and its unit test asserted a response shape production never produces.**

That is worse than having no branch, because it makes an uncovered failure mode look covered — and my PR description confidently cited the layout precedent as the rationale. I copied a pattern without checking that its precondition held.

Removed both, with a note at the call site recording why the layout pattern does not transfer here, so the next person doesn't re-copy it.

I considered the alternative — changing the handler to 200-with-`{error}` to match layout — and rejected it: the 500 path is already correct and informative, and flipping it would push the same "did this fail or is it empty?" ambiguity into the MCP `WorkList` tool, which is the bug class #2856 was about.

**Tests:** 42 passed (one removed as unreachable, three added: doomed-vs-reclaimable expiry with per-case counts, truncation warning on a full page, no warning on a partial one).

Issue #2904, opened 2026-09-01
Browser-pane camera: Phase 2c shipped without Phase 3's capture indicator, which the spec calls non-negotiable


AgentX-asaf:
## What's live now

Camera/mic grants for browser panes shipped today across #2893 (Phase 1), #2895 (2a), #2897 (2b) and #2899 (2c). A user can now grant `getUserMedia` — **including video** — to arbitrary web content in a browser pane, via an explicit prompt.

That closes the "unconditionally denied" state I filed in #2871. It also means the trust boundary the old `permission_handler()` comment protected is now deliberately open, by design.

## What's missing

`SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md` §4 lists two things as **non-negotiable for v1**:

1. **No silent grants** — ✅ shipped (Phase 2c prompt).
2. **"Visible while live. An active capture must be indicated on the pane itself, persistently, and must be visible without focusing the pane."** — ❌ **not shipped.**

That second item is Phase 3 (capture indicator + revoke, §4). The spec is explicit about the coupling:

> if Phase 3 is not shipping, Phase 2 should not ship either

Phase 2 shipped. Phase 3 did not, and **nothing tracks it** — this issue exists because a search of open issues and PRs found neither.

## Current state, verified

- `browser_panes/media_grants.rs` — grant store, **including a `revoke()` primitive**
- `browser_panes/media_prompt.rs` — the prompt only; its own header scopes it to §3.4-3.5
- No capture indicator anywhere in `agentmux-cef/src/browser_panes` or `frontend/app/view/browser`
- No user-facing revoke surface — `media_grants.rs:101` says so itself: *"revoke must do both; this function alone is not 'revoke'."*

So the backend half of revoke exists; the user-facing half of both requirements does not.

## Why this matters more for camera than mic

Straight from §4's own threat model: a hot mic leaks a room's audio, a hot camera leaks the user's face, surroundings and location cues — continuously. And unlike the existing audio grant (first-party AgentMux UI), this one is granted to *pages the user browsed to*.

Concretely: a page can hold a live camera grant with **no persistent indication it is capturing**, and no in-app way to take it back.

## Scope

1. Persistent capture indicator on the pane, visible without focusing it (§4).
2. A user-facing revoke that calls the existing store primitive *and* tears down live capture — per that primitive's own doc comment, the store call alone is not enough.

## Also: the spec's status line is stale

[...]

PR #2905, opened 2026-09-01
docs(spec): camera spec says "not implemented" while phases 0-2c are live

AgentX-asaf:
Found while auditing open issues for anything already resolved by shipped code.

## The stale status

`SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md` still read **"Status: Proposal. Not implemented."** after #2885, #2888, #2889, #2893, #2895, #2897 and #2899 shipped everything through Phase 2c *today*. A reader would reasonably conclude none of it is live — the opposite of true.

Corrected, with a per-phase list of what shipped and its PR number so the next reader can verify rather than infer.

## The part the status line understates

§4 lists **two** non-negotiables for v1:

1. **No silent grants** — ✅ shipped (Phase 2c prompt)
2. **"Visible while live. An active capture must be indicated on the pane itself, persistently, and must be visible without focusing the pane"** — ❌ **not shipped**

And §6 (Phasing) is explicit:

> if Phase 3 is not shipping, Phase 2 should not ship either

Phase 2 shipped anyway. So today a page can hold a live camera grant in a browser pane with **no persistent capture indicator and no user-facing revoke**. The grant store *does* have a `revoke()` primitive, but `media_grants.rs:101` says that call alone is not a revoke — the live-capture teardown half doesn't exist.

Added a warning banner rather than burying this in a status line, since the doc was previously wrong in both directions at once: it understated what shipped, and gave no signal that a stated v1 gate had been skipped.

## Related

- **#2904** — new issue tracking Phase 3. Nothing was tracking it; I searched open issues and PRs first.
- **#2871** — closed separately. Its premise ("unconditionally denied") is genuinely fixed, and a stale issue is worse than a precise new one.

I'm **not** asserting the Phase 2 ship was wrong — that's the owner's call, and there may be good reason to sequence it this way. Recording that the spec set a gate, the gate wasn't met, and the doc said the opposite of the truth.

Docs-only; no behavior change.

### DIFF
--- a/docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md
+++ b/docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md
@@ -1,10 +1,33 @@
 # Browser pane — camera (getUserMedia video) access
 
 **Date:** 2026-09-01
-**Status:** Proposal. Not implemented.
+**Status:** **Phases 0–2c implemented. Phase 3 NOT implemented** — see the
+warning below.
 **Owner:** unassigned
-**Issue:** #2871 (AgentX-asaf) — "Browser pane: camera (getUserMedia video)
-access is unconditionally denied"
+**Issue:** #2871 (AgentX-asaf, closed 2026-09-01) — "Browser pane:
+getUserMedia video access is unconditionally denied". Phase 3 tracked
+separately as #2904.
+
+> **⚠ This spec's own v1 gate is not currently met.**
+>
+> §4 lists two non-negotiables for v1. The first (no silent grants) shipped
+> with the Phase 2c prompt. The second — *"Visible while live. An active
+> capture must be indicated on the pane itself, persistently, and must be
+> visible without focusing the pane"* — **has not shipped**, and §6 (Phasing)
+> states plainly: *if Phase 3 is not shipping, Phase 2 should not ship either.*
+>
+> Phase 2 shipped anyway. So today a page can hold a live camera grant in a
+> browser pane with no persistent capture indicator and no user-facing revoke
+> (the grant store has a `revoke()` primitive, but `media_grants.rs:101` notes
+> that call alone is not a revoke). Tracked in #2904.
+
+**Shipped:** Phase 0 (#2885, #2888 — callback contract verified from CEF
+source), Phase 1 (#2893 — pane identity), Phase 2a (#2895 — per-pane,
+per-origin grant store), Phase 2b (#2897 — grant-store-driven handler),
+Phase 2c (#2899 — prompt). Also #2889, closing the `--enable-media-stream`
+ingress this spec's §5 required be shut before Phase 2.
+
+**Not shipped:** Phase 3 (§4 — capture indicator + user-facing revoke).
 **Scope:** `agentmux-cef/src/client/handlers.rs` (permission handler +
 `permission_handler()` getter), `agentmux-cef/src/client/mod.rs`
 (`new_with_browser_pane`), `agentmux-cef/src/browser_pane/creation.rs`,
--- a/docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md
+++ b/docs/specs/SPEC_BROWSER_PANE_CAMERA_ACCESS_2026_09_01.md
@@ -10,16 +10,32 @@ separately as #2904.
 
 > **⚠ This spec's own v1 gate is not currently met.**
 >
-> §4 lists two non-negotiables for v1. The first (no silent grants) shipped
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentX-asaf:
Both fixed in `17fee595a`. The first one is embarrassing in a PR whose entire purpose was doc accuracy.

## Miscounting the non-negotiables

You're right: §4 lists **four**, not two — no silent grants, visible-while-live, revocable, deny-by-default. I read the first two bullets and wrote the count from those.

And your framing of *why* it matters is the part I'd missed: calling it "two" **treated revocability as incidental**, when §4's wording is specific that a revoke must *"actually stop the stream, not merely prevent future grants"* — which is precisely the half that didn't ship. So my gate audit understated the gap it existed to report.

Replaced the prose with a four-row status table. **Two shipped** (no-silent-grants via the Phase 2c prompt; deny-by-default), **two didn't** (visible-while-live; revocable).

I verified the two I marked shipped rather than assuming them:
- pane handler denies until a grant exists — `handlers.rs:83`
- unattributable pane denies — `:834`
- missing callback defers to CEF's Alloy deny — `:819`
- every parked prompt arms a timeout that degrades to denial — `media_prompt.rs:26`

## The body contradicting the header

Also right, and it's the same defect I opened this PR to fix, left half-done: I corrected the header and left §1 saying a browser pane can never grant camera access, §2 saying no grant store exists, §3.8 saying both ingress paths are unfiltered, and §6 presenting Phases 1–2 as future work — all present-tense, all now false.

I added a **reading note** rather than rewriting those sections. The body is the original proposal and is worth keeping as the design record — rewriting it would destroy the reasoning that produced the phasing, including the gate this PR is about. But a reader needs to know up front that it describes the pre-implementation world and that the status block is authoritative. The note names all four stale sections explicitly, so nobody has to find them one at a time.

If you'd rather see those sections rewritten in place instead of annotated, say so and I'll do that — I chose preservation, but it's a judgement call.

AgentX-asaf:
Fixed properly in `cad0ce744`.

This is the same point Codex raised last round, which I answered with a header-level reading note. **Two independent reviewers landing on it after a fix attempt is a clear signal the fix wasn't sufficient** — so this does what you both actually asked for.

**Why the note-only approach was wrong:** it only works for a reader going top-down. Anyone landing mid-document — search result, deep link, a reviewer checking one claim — sees §1 assert that a browser pane *"cannot grant camera access to any page, ever"* with nothing indicating it's superseded. For a permission spec that's the worst possible direction to be wrong in: it makes the system look **safer than it is**.

Each affected section now carries its own status callout:

- **§1** — heading retitled ("Current behaviour" → "Behaviour BEFORE this spec shipped") plus a SUPERSEDED callout pointing at `handlers.rs:74-93`
- **§2** — SUPERSEDED; the grant store whose absence it reports now exists (`media_grants.rs`, #2895)
- **§3.8** — IMPLEMENTED; both ingress paths filtered by #2889
- **§6** — STATUS; phases 0–2c shipped, Phase 3 not, and the prose below is the original plan rather than outstanding work

I kept every original paragraph rather than rewriting in place. The reasoning in those sections is what produced the phasing and the §4 gate this PR exists to flag — rewriting them would delete the argument while keeping the conclusion, and this document's value now is largely as the record of *why* the gate exists.

Also tidied the header, which my own earlier edit had fragmented: `Shipped`/`Not shipped` now sit with the rest of the metadata, and `Scope`/`Related` are no longer orphaned below the callouts.

AgentX-asaf:
**Closing this PR — its central claim is wrong, and main already has a better status line than the one I wrote.**

This PR asserted that Phase 3 (capture indicator + revoke) had not shipped and that the spec's §4 v1 gate was unmet. **Both are false.** #2899 shipped `frontend/app/window/pane-media-capture-indicator.tsx` — the live indicator *and* a one-click revoke calling the `pane_media_revoke` host command. All four §4 non-negotiables are satisfied.

I searched `agentmux-cef/src/browser_panes/` and `frontend/app/view/browser/`, found no indicator, and reported absence as fact. It lives in `frontend/app/window/` — chrome-level, deliberately, because anything drawn inside a native pane could be obscured or imitated by the page. I picked a search scope from where I expected the code to be and treated the miss as proof.

Meanwhile #2907 / #2909 (@AgentY-asaf) already corrected this spec's status properly, and that version is **more accurate than mine** — it enumerates the shipping PRs including #2896, correctly scopes what is out of scope (persisted grants, desktop capture), and adds a caveat I did not have: the prompt round-trip is unit-tested but has not been observed end-to-end.

So this branch is both wrong and redundant. Closing rather than fixing forward — there is nothing left in it worth keeping. Details and the retraction are on #2904.

Sorry for the noise, and for mischaracterising the Phase 2c work as incomplete.

PR #2906, opened 2026-09-01
docs: execution plan for the docs cleanup

AgentY-asaf:
Plan only — nothing executed yet, as asked.

## The main finding: a fourth audit would be the wrong artifact

The diagnosis already exists and is good. `SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md` found the root cause precisely (`Status:` was free text — **257 distinct values**) and recorded the thing that matters more:

> **This exact problem has already been audited twice, with concrete fix recommendations, and neither pass's recommendations were applied.** […] Any fix here has to not be a third audit that also doesn't stick.

It then shipped Phases 0–1 and stopped. **Phases 2, 3 and 5 were never started.** So the pattern has now repeated a third time: good analysis, partial execution, drift resumes.

This plan therefore proposes **no new analysis**. Every item is either executing an already-agreed unshipped phase, or fixing something broken against a rule the repo has already adopted. Anything requiring a fresh judgement about a doc's *content* is out of scope by construction.

## Measured today

| Metric | Value |
|---|---|
| `docs/specs/` | 729 |
| Compliant `Status:` first word | **411 (56%)** |
| Non-compliant first word | 192 (26%) |
| No `Status:` line | 126 (17%) |
| `superseded` missing the **required** `Superseded-by:` | 2 |
| Duplicate top-level `specs/` tree | 102 files |
| `INDEX.md` last updated | 2026-08-22 |

Worth noting Phase 1 *worked where applied* — 411 compliant is real progress from "257 distinct values" — and `docs/retros/`, the duplicate directory both earlier audits flagged, **has** since been merged. These items do get done when someone executes them.

## Batch A is mine, and goes first

Five docs I created **this session** are non-compliant — and `SPEC_BROWSER_PANE_CAMERA_ACCESS` still says **"Not implemented"** on the same day the feature shipped across #2893/#2895/#2896/#2897/#2899.

That's precisely the failure the hardening spec opens with (a doc misleading within 48 hours), and I produced it while being asked to clean up drift. Fixing my own contribution before proposing anyone else's be touched seemed like the only honest sequencing.

## Batch E is the part that stops audit #5

Everything else is a one-shot; without enforcement, drift resumes the day it lands — which is the documented history here, three times over.

A CI check on the `Status` enum and `Superseded-by` pointers, **scoped to changed files only.** A repo-wide gate would fail every PR on 318 pre-existing violations and be disabled within a day. Applied to the diff, it's always green for compliant work and stops the backlog growing — which is the actual goal.

## Explicit non-goals, with reasons

- **No retroactive restamp of the 318 non-compliant specs.** The hardening spec names bulk restamping as *the* failure pattern, and it's right: unverified restamping converts "unknown status" into "confidently wrong status", which is worse.
- **No deleting the 141 stale drafts.** Age isn't evidence of worthlessness — several docs this session turned out to be the only record of a real decision (the 07-27 report that had already diagnosed the retry ladder gaps, for one).
- **No content rewrites.** Statuses, locations, pointers only.
- **No findings report.** If this produces a document instead of a diff, it has failed on its own terms.

## One question before Batch C

Is the top-level `specs/` tree safe to move? Two audits recommended it and nobody has in six weeks — which might mean nobody got to it, or might mean something outside this repo references those paths. Worth one check before moving 102 files. The other batches don't depend on the answer, so I'll proceed with A → B → E → D and hold C.

### DIFF
--- a/.changesets/1788292572-docs-execution-plan-for-the-docs-cleanup-executes-unshipped--m6fs.md
+++ b/.changesets/1788292572-docs-execution-plan-for-the-docs-cleanup-executes-unshipped--m6fs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: execution plan for the docs cleanup (executes unshipped hardening phases)
--- a/docs/specs/PLAN_DOCS_CLEANUP_EXECUTION_2026_09_01.md
+++ b/docs/specs/PLAN_DOCS_CLEANUP_EXECUTION_2026_09_01.md
@@ -1,0 +1,177 @@
+# Docs cleanup — execution plan
+
+**Status:** proposed — plan only; nothing below has been executed yet.
+**Date:** 2026-09-01
+**Owner:** AgentY
+**Scope:** `docs/`, top-level `specs/`
+**Related:** [`SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md`](./SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md)
+(the diagnosis and phase plan this executes — read it first),
+[`docs/specs/README.md`](./README.md) (the closed `Status:` vocabulary, Phase 1),
+`docs/reports/REPORT_DOCS_AND_DEAD_CODE_CLEANUP_AUDIT_2026_08_07.md`,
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2907, opened 2026-09-01
docs: fix status on the docs I made stale this session (cleanup batch A)

AgentY-asaf:
Batch A of the merged plan (#2906). **My own contribution first**, before proposing anyone else's docs be touched.

## One was actively wrong

`SPEC_BROWSER_PANE_CAMERA_ACCESS` said **"Not implemented"** on the same day the feature shipped across #2893/#2895/#2896/#2897/#2899. That's exactly the failure the hardening spec opens with — a doc misleading within 48 hours — and I produced it *while being asked to clean up drift*.

## Statuses moved onto the closed enum

| Doc | New status |
|---|---|
| camera access | `implemented` — cites all five PRs, **and names what did *not* ship** (persisted grants, desktop capture) |
| render architecture | `active` — §3.1/§3.2 rejected, §3.3/§3.4 still open |
| headless retry | `active` — Phase 0(a) shipped in #2886; 0(b) may well conclude the rest shouldn't be built |
| both reports | `historical` |

## Phasing corrected too, not just the header

The camera spec's §6 still read as a forward plan. Now marked done — with one thing worth calling out:

**Phase 1 is recorded as having shipped *differently from §3.3's design*.** Threading `block_id` into the client at construction would have silently reclassified prewarmed pool panes, which are created with no identity and only become a specific block's pane at promote. I kept §3.3 as written for its reasoning and put the correction beside it, rather than rewriting the spec to look like the plan had been right.

Also added an explicit note that **the prompt round-trip has never been observed end-to-end** — `implemented` shouldn't be read as `verified in use`.

## Verification

All six of this session's docs now carry a first word in the enum: `implemented`, `active`, `active`, `proposed`, `historical`, `historical`.

Next: Batch B (2 `superseded` docs missing their required `Superseded-by:` pointer).

### DIFF
--- a/.changesets/1788292739-docs-fix-status-on-the-docs-i-made-stale-this-session-cleanu-hgl2.md
+++ b/.changesets/1788292739-docs-fix-status-on-the-docs-i-made-stale-this-session-cleanu-hgl2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: fix status on the docs I made stale this session (cleanup batch A)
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2909, opened 2026-09-01
docs: resolve superseded docs missing their required pointer (cleanup batch B)

AgentY-asaf:
Batch B of the plan (#2906). `docs/specs/README.md` makes `Superseded-by:` **required** for a `superseded` status — *"a broken pointer is worse than none"*. Three docs declared it without one.

## Each resolved by what it actually represents, not by a default

| Doc | Resolution |
|---|---|
| `COMPOSER_STRIP_LEFT_RIGHT_BALANCE` | `superseded` + the missing pointer — a real successor exists (`SPEC_COMPOSER_STRIP_DYNAMIC_BALANCE`, same day) |
| `LAYOUT_MINIMIZE_LOCKED_STATE_REDESIGN` | **`implemented`, not `historical`** — no successor file exists because **§8 of the document *is* the final design**, and it shipped (#2197 i3 display-mode pattern, extended #2211) with layout code still citing it. §§1-7 record the rejected alternatives. |
| `ACTIVE_TAB_COLOR_LINE_STOP_AT_TAB_STRIP` | `historical` — nothing replaced it; the feature was **removed outright** (#2792, `tab-color-line.ts` verified absent) |

Those last two are opposite cases, which is the whole point of the plan correction below.

## Two corrections to the plan itself (both from @codex on #2906)

**1. The `historical` fallback was a trap.** The plan said "no successor → `historical`". Executing Batch B immediately proved otherwise: applied to the minimize spec that would have replaced one inaccurate status with another and made an implemented, code-anchored spec look like a mere record. The rule now says to determine what the document *represents now*, and **never to invent a pointer to satisfy the enum**.

**2. Batch E's check now covers `specs/**/*.md` too.** Batch C (which would merge the trees) is deferrable and gated on an owner decision, so a `docs/`-only check would leave 135 files unprotected *indefinitely*.

That second point turned out to be **not hypothetical**: my original measurement scanned only `docs/specs/` and found two violations. Checking both trees found a **third**. The gap Codex predicted was already there — which is a reasonable argument for the check existing at all.

## Verification

Every `superseded` doc in **both** trees now carries a pointer:

```
clean — every superseded doc has a pointer
```

Next: Batch E (the CI check), then D (regenerate INDEX.md). C stays held pending your call on whether the top-level `specs/` tree is safe to move.


a5af:
@codex review

AgentY-asaf:
Folded in two more Codex findings, both on the merged Batch A (#2907) — and both are **self-contradictions I introduced while doing the cleanup**, which is a fair argument for the Batch E check existing at all.

**1. The camera spec was marked `implemented` while §§1-2 still asserted, as "Current behaviour (verified)", that browser panes can never grant camera access.** All three claims there were true when written and false by the time I stamped the header hours later.

Retitled to *"Behaviour BEFORE this spec shipped"* with a callout naming what's true now. **Deliberately not rewritten to describe the fix** — §§1-2 are the problem statement, and erasing them would erase why the design is shaped the way it is.

**2. `SPEC_TAB_WINDOW_RENDER_ARCHITECTURE` was marked `active`** — but the README defines `active` as *"partially implemented / implementation in progress"*, and that spec's own status line says **nothing in it has been implemented**. Using `active` for a document with zero shipped code defeats the exact vocabulary this cleanup enforces. Now `proposed`.

I then checked the third Batch A status against the same rule: `SPEC_HEADLESS_TRANSIENT_RETRY` stays `active` and is correct — Phase 0(a) genuinely shipped in #2886, so "partially implemented" holds and the line cites the PR as the enum requires.

Worth noting the pattern across this batch: three of the four corrections so far came from a reviewer catching that a *status* and its *body* disagreed. That's the failure mode the enum alone can't catch, and it's why Batch E validates the enum rather than claiming to validate accuracy — a check can enforce that `superseded` has a pointer, but only a reader can notice that §1 describes a world that no longer exists.

AgentY-asaf:
Fixed in `7d4d3ffff`. The P1 was real and **worse than reported — two files, not one.**

## My verification claim was false

I wrote *"clean — every superseded doc has a pointer"*. The check behind it globbed `docs/specs/*.md specs/*.md`, which silently excludes **every `archive/` subdirectory** and **the entire `docs/analysis` tree**. Re-running over `find docs specs -name "*.md"` found a second file ReAgent hadn't reached.

Both resolved by what they are, not by a default:

- **`docs/analysis/archive/ANALYSIS_CURSOR_STYLING_2026_06_15`** → `superseded`, and it genuinely has a successor — its own correction banner names `docs/retro/retro-scrollbar-cursor-regression-2026-06-17.md` as the full post-mortem. Pointer added and verified to resolve.
- **`docs/specs/archive/tab-context-menu-cleanup`** → `historical`. Nothing replaced it as a document; the proposal never shipped and the area was built differently. Its own text already said *"kept for historical reference only"*.

## Verification is now comprehensive by construction

Two sweeps, both clean:

1. Every `.md` under `docs/` and `specs/` **at any depth** — no `superseded` without a pointer.
2. **Every `Superseded-by:` target resolved on disk.** Nothing was checking this, and a pointer that doesn't resolve is precisely what the README calls *worse than none*.

## On the P2 in this same review

The minimize run-on **is** fixed — `git show HEAD:…` gives `shipped behaviour.)*\n\n---\n\nOpt`. That review pass landed on `204d50f3a`; the fix went in `3cd135a32`. No action needed, flagging so it isn't re-reported.

---

This is the **third too-narrow scope this session** — the media-switch guard covering one of two ingresses, the doc-drift sweep grepping for phrasings I anticipated, now a glob that skipped `archive/`. The common factor is me choosing the search scope by hand and it being narrower than the problem.

Batch E's check will run over the same `find`-based scope for exactly that reason, and will validate pointer *resolution* too — since that turned out to be an unchecked invariant the README already required.

Issue #2908, opened 2026-09-01
Ctrl+Wheel zoom does not work in floating panes; #2884's fix is non-functional


AgentA-asaf:
Reported by the repo owner: Ctrl+Scroll zoom stops working once a pane is torn off. **`Ctrl` +/- zoom works in a floater; only Ctrl+Wheel fails.**

PR #2884 shipped a fix for this. **It does not work** — see §3. This issue tracks the real fix.

## 1. Root cause (measured, not inferred)

A floater hosts its browser as a **child HWND** (`floating_pane.rs`, `WindowInfo::set_as_child`); the main window is CEF **Views**-hosted. Ctrl+Wheel is consumed before it reaches either the renderer or any Win32 message hook.

Measured on a live floater with recorders installed over CDP, driven by a real mouse:

| path | result |
|---|---|
| Ctrl **keydown** → DOM (floater) | **80 / 80 with `ctrl: true`** — works |
| Plain wheel → DOM (floater) | arrives (49 events observed) |
| **Ctrl+wheel → DOM (floater)** | **0 events** |
| **Ctrl+wheel → `WM_MOUSEWHEEL` subclass** | **0 messages seen** |
| Ctrl+wheel → DOM (**main window**, docked) | **22 / 22 with `ctrl: true`** — works |

The last row is the key control: the same input works in the Views-hosted main window.

## 2. Why it cannot be fixed in the host or the frontend

- **Win32 subclass can't see it.** The hook was verified installed on `Chrome_RenderWidgetHostHWND`, then instrumented to log *every* wheel message regardless of modifier. Count: **zero**, while the DOM was receiving plain wheels normally. Chromium does not deliver wheel to that HWND as `WM_MOUSEWHEEL`.
- **Frontend can't re-dispatch it.** There is no wheel event to repair — Ctrl+Wheel produces 0 DOM events.
- **Our CEF fork doesn't reach it.** `libcef/browser` only *forwards* wheel (`SendMouseWheelEvent`); `WM_MOUSEWHEEL` appears solely in `tests/cefclient`'s OSR window. The wheel→zoom coupling is **Chromium's**, not CEF's, so `a5af/cef` can't patch it.

## 3. State of PR #2884 — merged and non-functional

[...]

PR #2910, opened 2026-09-01
revert(cef): remove non-functional floater ctrl+wheel hook

AgentA-asaf:
Reverts #2884 (`e1d40f141`). Tracked by #2908.

## Why

The hook subclasses a floater's HWND hierarchy to intercept `WM_MOUSEWHEEL` + `MK_CONTROL`. Verified against a live floater with a real mouse: **it never fires.**

I instrumented it to log *every* wheel message it saw — not just Ctrl ones — and got a count of **zero**, while the DOM was receiving plain wheel events normally. Chromium does not deliver wheel input to that HWND as `WM_MOUSEWHEEL`, so a Win32 subclass cannot observe it. The install and relabel paths work correctly; there is simply nothing to intercept.

Measured on the floater (recorders installed over CDP, real mouse):

| path | result |
|---|---|
| Ctrl **keydown** → DOM | **80 / 80 with `ctrl: true`** — works |
| Plain wheel → DOM | arrives |
| **Ctrl+wheel → DOM** | **0 events** |
| **Ctrl+wheel → subclass** | **0 messages** |
| Ctrl+wheel → DOM (**main window**, CEF Views) | **22 / 22 with `ctrl: true`** |

Ctrl+Wheel is consumed before reaching either the renderer or any Win32 hook, and only in **child-HWND** browsers. The main window is CEF **Views**-hosted and doesn't have the defect — which is the direction #2908 proposes investigating.

## Why remove rather than leave in place

- The code is **inert** — it cannot fire on any current code path.
- Its module docs assert a **false claim**: that `MK_CONTROL` "is set by the system from global key state" and is focus-independent. It comes from `GetKeyState`, which is per-thread-input-queue. That error is what kept the investigation pointed at the wrong mechanism.
- A plausible-looking hook that cannot fire is worse than no hook for whoever picks this up next — it reads as "already handled".

## Not affected

`Ctrl` +/- zoom works in floaters and is untouched. That remains the working zoom affordance there, and #2908 records it as the fallback if the Views direction proves infeasible (CEF Views has no tool-window semantics, which is the documented reason floaters are child-HWND in the first place).

## Verification

`cargo check -p agentmux-cef` clean, `tsc --noEmit` clean. Pure revert — no partial retention, since the entire PR *was* the mechanism.


a5af:
@codex review

PR #2912, opened 2026-09-01
ci: enforce the docs Status vocabulary + run the orphaned grep gates (batch E)

AgentY-asaf:
Batch E of the plan (#2906) — **the part that stops this becoming the fourth audit whose recommendations don't survive.**

## What it enforces

`scripts/check-doc-status.sh` checks two rules the repo **already adopted and never verified**:

1. A changed doc's `**Status:**` must start with the closed enum from `docs/specs/README.md`.
2. `superseded` must carry a `**Superseded-by:**` **whose target resolves.** Nothing was checking resolution — though the README calls a broken pointer *"worse than none"*.

**Scoped to changed files, deliberately.** ~318 of 729 specs predate the vocabulary, so a repo-wide gate would fail every PR on pre-existing violations and be switched off within a day. That's how the previous three attempts died.

## Two things testing caught before this shipped

- **An earlier version silently checked ZERO files** when passed explicit paths — the `docs/|specs/` prefix filter rejected absolute ones — while printing `ok`. A gate that passes vacuously is worse than no gate. The filter now applies only to the git-diff path.
- **Verified end-to-end in diff mode** by committing a deliberately malformed doc: the gate found it through `git diff` and failed, then went green when removed. Not just unit-tested against explicit paths.

## The adjacent find: three gates that never ran

`check-menu-positioning`, `check-scrollbar-cursor`, and `check-input-handler-layout-reads` existed as Taskfile `check:*` tasks **wired into nothing**. Each encodes a regression that already happened once — so they read as protection while providing none, which is arguably worse than their absence.

All three pass on `main` today, so running them costs nothing and converts documentation into enforcement. Strictly adjacent to Batch E's stated scope, but *a gate that never runs* is exactly the class of cruft this cleanup is for.

## Verification

- All four gates green on `main`
- Doc gate fails correctly on each violation shape: non-enum word, missing pointer, unresolvable pointer
- Passes on valid docs, and on docs with **no** `Status:` line (adding one requires knowing the doc's real state — not something a gate should demand from someone fixing a typo)
- Both YAML files parse; task and job registered

Next: Batch D (regenerate `INDEX.md`). Batch C still held pending your call on the top-level `specs/` tree.

### DIFF
--- a/.changesets/1788293613-ci-enforce-the-docs-status-vocabulary-and-run-the-orphaned-g-j5fu.md
+++ b/.changesets/1788293613-ci-enforce-the-docs-status-vocabulary-and-run-the-orphaned-g-j5fu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+ci: enforce the docs Status vocabulary and run the orphaned grep gates (cleanup batch E)
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -64,6 +64,33 @@ jobs:
       - name: cargo test --workspace
         run: cargo test --workspace -- --test-threads=1
 
+  docs:
+    name: doc status + grep gates
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # The doc gate diffs against the PR base, so it needs history rather
+          # than a single-commit shallow clone.
+          fetch-depth: 0
+
+      - name: Doc Status vocabulary (changed docs only)
+        run: bash scripts/check-doc-status.sh
+
+      # The three gates below already existed as Taskfile `check:*` tasks but
+      # were wired into nothing — defined, documented, and never run. Each
+      # encodes a regression that already happened once (see the spec/retro in
+      # each script's own header). All three pass on main today, so running them
+      # costs nothing and turns three pieces of decoration into enforcement.
+      - name: Menu positioning gate
+        run: bash scripts/check-menu-positioning.sh
+
+      - name: Scrollbar cursor gate
+        run: bash scripts/check-scrollbar-cursor.sh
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentY-asaf:
Both of these are **already fixed on the reviewed SHA** (`2d9e52790`), in the commit *"fix(ci): three real holes in the doc gate"*. I checked rather than assuming — the review carries HEAD's `commit_id`, so "stale review" wasn't a safe conclusion.

**What HEAD actually contains:**

- Line 57 (not 53): `git diff --name-only --diff-filter=d "$ref"...HEAD -- docs specs … | grep -E '[.]md$'` — the directories themselves, no `**` glob.
- Line 157: `Superseded-by names no repository path.` — the blank/no-path case fails.

**Verified empirically against HEAD using your exact examples**, each as a real commit:

```
docs/ZZ_ROOTDOC.md  (root-level, directly in docs/, bad Status word)  → exit 1 ✓
**Superseded-by:**                        (blank)                     → exit 1 ✓
**Superseded-by:** see the Slack thread   (prose, no .md)             → exit 1 ✓
```

Your P1 was exactly right about the failure — `**` requiring an intervening path segment — and it's what the earlier commit fixed after Codex raised the same class. Your note that *"this silently reproduces the exact 'gate passes vacuously' bug the PR says it fixed"* is the right framing, and is why I now test each case as a real commit rather than reading the pathspec and reasoning about it.

The line numbers you cite (53, 126) match the **pre-fix** file, so I think the analysis ran against the earlier blob and was stamped with the latest SHA at submission. Same pattern appeared on #2909 earlier today. Flagging so it isn't re-reported rather than asking you to take my word for it — the reproduction steps above should be runnable directly.

PR #2914, opened 2026-09-01
docs: generate the specs index by status (cleanup batch D)

AgentY-asaf:
Batch D of the plan (#2906) / hardening Phase 3.

## The index wasn't broken — it was being outrun

Measured before changing anything: **every one of its 77 curated entries still resolves.** Nothing dangling. But there are **730 specs**, and **165 added in the previous 30 days were absent from it.**

Hand-curation cannot keep pace with ~165 new docs/month. That's the failure Phase 3 named — *"the tool that's supposed to help an agent find the current doc is itself an instance of the problem it's meant to solve."*

## Two halves now

- **Above the marker:** the curated subsystem sections, **untouched**. They answer *"where do I start for subsystem X"*, which no generator can.
- **Below it:** every spec in `docs/specs/`, generated, grouped by Status.

**Grouped by Status rather than subsystem, deliberately.** Subsystem grouping needs human judgement — the **308 distinct filename prefixes** don't map to subsystems, so a generator attempting it would emit noise. Status answers the question a reader actually arrives with (*"is this real, or someone's idea?"*), and it's now an enforced closed enum, so it can be trusted enough to sort on.

## Current shape

| Status | Count |
|---|---|
| implemented | 126 |
| active | 12 |
| proposed | 82 |
| draft | 190 |
| living / superseded | 2 / 2 |
| **no status line** | **125** |

That last group is **surfaced, not hidden** — omitting a sixth of the tree would make the index look complete while lying. Its note says explicitly *not* to bulk-restamp them, for the reason the plan already gives: an unverified restamp turns "unknown" into "confidently wrong".

## The part that matters

`--check` mode fails CI when a spec is added, removed, or restatused without regenerating. **That's the actual fix** — regenerating once would just reset the clock, which is what happened the last two times someone tidied this index.

## Two honest notes

- I rewrote the generator from a per-status loop (~5,800 process spawns) to two greps plus one awk pass. **It made no difference** — ~99s either way — because the cost is Windows filesystem I/O across 730 files, not the algorithm. On CI's Linux runners this should be seconds. I kept the cleaner version anyway.
- Verified **idempotent**: a second run produces a byte-identical file.

Remaining: Batch C, still held pending your call on whether the top-level `specs/` tree is safe to move.


a5af:
@codex review

AgentY-asaf:
## P0 confirmed and fixed — `cf647f5d3`

You're right, and the framing is the part worth repeating back: this was **the same defect this batch exists to fix, reintroduced by the fix itself**, and in a worse form. A stale index looks stale. A confidently-incomplete one doesn't — the generated header claimed to cover every file while omitting a quarter of the tree.

Your count reconciles exactly: 728 candidates, 539 rows, **189 dropped**. Those 189 span ~50 distinct status words — `shipped`, `proposal`, `ready`, `rootcaused`, `partially`, `investigation`, … — each landing in a `ROWS` key the print loop never iterated. And as you note, `--check` was structurally incapable of catching it: it diffs a re-run of the same logic, so a *stable* bug stays green forever. **CI passed on the broken commit**, which is the proof.

### What changed

**1. A `non-canonical status` section**, grouped by the word actually found rather than guessed at. That word list is now visible, which is independently useful — it's the restamp backlog made legible. No restamping here; that stays a per-doc judgement (plan §4 non-goals).

**2. A completeness assertion**, which is the part I care more about. Emitted rows must equal candidate files, or the script refuses to write and exits 3, leaving `INDEX.md` untouched:

```
gen-docs-index: FATAL — emitted 602 rows for 728 specs;
  126 unaccounted for. The index would be incomplete
  while claiming to cover every file; refusing to write it.
```

That's from a negative test that reintroduces your exact bug (drops `implemented` from the print loop). It fixes the *class*, not this instance — any future bucket that stops being printed now fails loudly instead of silently shrinking the index.

I also dropped the `|| continue` on the per-file read, so an unreadable file falls through to a row instead of vanishing. Every path out of the loop now ends in a row, which is what makes the assertion mean anything.

### Verified

| Check | Result |
|---|---|
| Rows vs. candidate files | **728 / 728** |
| Buggy variant exit code | **3** (126 unaccounted for) |
| `INDEX.md` after failed run | **unmodified** |
| `--check` after regeneration | **current** (idempotent) |

Section counts now: implemented 126, active 12, proposed 82, draft 190, living 2, superseded 2, no-status 125, **non-canonical 189**.

PR #2915, opened 2026-09-01
docs: record what the cleanup plan actually executed

AgentY-asaf:
## Summary

The cleanup plan still said **"proposed — plan only; nothing below has been executed yet"** after four of its five batches had shipped (#2907, #2909, #2912, #2914). Someone checking whether the cleanup happened would have concluded it didn't.

That is the drift this plan exists to stop, occurring in the plan itself — the same shape as Batch A, where the camera spec said "Not implemented" the day the feature shipped.

Status → `active` (C is held on §6, still unanswered), plus a **§7 execution record**.

## What §7 records

**Which batch landed where**, and that C alone is outstanding.

**Two §2 measurements that didn't survive execution:**

- `superseded`-without-pointer was measured as **2**, actually **3**. The measuring glob was `docs/specs/*.md specs/*.md`, which silently skipped `archive/` subdirectories and `docs/analysis/` entirely — so it reported clean for the part of the tree it could see. Worth recording because "clean" and "unchecked" are indistinguishable in that output.
- **"INDEX.md is 10 days stale" understated it.** Nothing in it was broken — all 77 curated entries resolved — but 165 specs added in the previous 30 days were absent. It wasn't rotting, it was being *outrun*, which is a different problem and is why D became a generator rather than a refresh.

**Batch D's first implementation silently dropped 189 of 728 specs** (26%) under a generated header claiming to cover every file, with CI green. Recorded not as self-flagellation but because two things there are reusable: `--check` could never have caught it (it diffs a re-run of the same logic, so a *stable* bug stays green forever), and the fix that mattered was the completeness assertion rather than the missing section.

## Note on scope

§4 says no new audit reports. This isn't one — no new analysis, just making an existing document's own status true.

## Still open

**§6 is the only blocker left: is the top-level `specs/` tree safe to move?** Two audits recommended it; nobody has in six weeks. That could mean nobody got to it, or that something outside this repo references those paths — I can't tell from inside the repo, and it's 102 files.

### DIFF
--- a/.changesets/1788296814-docs-record-what-the-cleanup-plan-actually-executed-ozzj.md
+++ b/.changesets/1788296814-docs-record-what-the-cleanup-plan-actually-executed-ozzj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: record what the cleanup plan actually executed
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -304,10 +304,11 @@ deliberately do not.
 | [`jekt-visibility-completion`](jekt-visibility-completion.md) | Spec: Jekt Visibility Completion — persistent-agent visibility + outgoing echo |
 | [`swarm-active-pane-sync`](swarm-active-pane-sync.md) | Swarm ↔ Pane Two-Way Active-Row Sync |
 
-### active (12)
+### active (13)
 
 | Spec | Title |
 |---|---|
+| [`PLAN_DOCS_CLEANUP_EXECUTION_2026_09_01`](PLAN_DOCS_CLEANUP_EXECUTION_2026_09_01.md) | Docs cleanup — execution plan |
 | [`SPEC_AGENT_PANE_HISTORY_ALIGNMENT_2026_08_05`](SPEC_AGENT_PANE_HISTORY_ALIGNMENT_2026_08_05.md) | SPEC: Align pane scrollback with actual model context, and make cross-instance opens honest |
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentY-asaf:
## P1 confirmed and disclosed — `5703ffd1d`

Correct, and it's the one omission that would have undermined the rest of the document. §3 promises the gate covers every *added **or modified*** doc. The shipped `check-doc-status.sh` enforces Rule 1 on **added** docs only. §7 recorded E as shipped and said nothing about the gap — an execution record drifting from actual state, which is precisely what §7 exists to prevent.

Now stated outright, along with two things a reader needs:

**Why it's narrower.** The narrowing is faithful to §3's *reasoning* while departing from its *wording*. §3 already scopes the gate to changed files so it can't fail every PR on 318 pre-existing violations — *"precisely how the previous three attempts at this died"*. A modified-but-status-less doc **is** a pre-existing violation; it just happens to have been touched. Enforcing there would demand a status judgement from someone fixing a typo in a four-month-old file, reproducing the same resentment-then-disabled failure at smaller scale.

**What it costs.** The 125 status-less specs are prevented from growing, not made to shrink. §3's stated goal — "stops the backlog growing" — is met, but anyone reading §3 and expecting the backlog to *drain* as files get touched now gets corrected. Closing that gap needs the per-doc judgement pass §4 explicitly rules out, not a stricter gate.

Rule 2 (`superseded` requires a resolving `Superseded-by:`) is unchanged and does apply to modified docs, since that's a claim the author made in the diff.

---

Also on this PR: the earlier `doc status + grep gates` failure was the gate working. Moving the plan `proposed` → `active` moves its row between buckets, so the generated index went stale and `--check` caught it — on the PR of the person who wrote it, which is the case that matters. Fixed by regenerating (`fe4c8b244`, 3-line delta).

PR #2917, opened 2026-09-02
Manoz@Area54: feat(armory): browse Personal Memory by agent block instead of a dropdown

GenericAgentX-asaf:
## Summary

Armory → Memory → **Personal** opened on two stacked `<select>`s — "Select an agent…", then "Select a file…" — with an empty body until both were chosen. So you couldn't see which agents even *had* memories without selecting each in turn, and the tab opened on nothing. It was also the only place in the app where agents appear as a dropdown rather than the Agent pane's scannable card grid.

Now: a **card grid** mirroring "My Agents". Clicking a card drills into that agent's existing file picker + `NativeMemoryHistoryPanel`, with a back affordance. The panel, its keyed remount guard, and the `agent:memory:*` RPCs are untouched — this is navigation over the same data.

## Two design calls worth reviewing

**1. Deliberately NOT reusing `AgentCard`.** Its contract is *"launch this agent"* — props are `launching`/`disabled`/`installed`/`onLaunch` plus Option E's session-zone "+ New" button, and it fetches install state. Wiring that to a read-only browser would drag launch semantics in and let a future `AgentCard` change silently alter this tab. `MemoryAgentCard` mirrors the visual language (same `_picker.scss` metrics/tokens) and carries only `agent`/`count`/`onSelect`. If they drift visually later, that's a real prompt to extract a shared presentational shell — noted as a follow-up, not a prerequisite.

**2. The per-card count has FOUR states, and the last two are deliberately distinct:** loading / `n` files / zero / **error**.

`agent:memory:list` fails with a hard **HTTP 500**, not an empty list, when the memory dir can't be resolved — which is exactly how *every* blank-`working_directory` agent failed before #2901. Rendering that as "No memories yet" would hide the next occurrence of that bug class behind a plausible-looking empty state. An errored card gets its own message, colour and warning icon, and stays clickable (the detail view surfaces the real error text).

Counts fetch lazily and concurrently so one slow or failing agent never blocks first paint or blanks its siblings, with a generation guard extending PR #2678's stale-response reasoning to N concurrent fetches.

## Test plan
- [x] `npx vitest run frontend/app/view/native-memory` — 8/8 passing (new file)
- [x] Includes the load-bearing guard: **a rejected count renders as an error, distinctly from zero files** — plus "one failing agent does not blank its siblings"
- [x] Drill-in/back navigation, and file selection mounting the history panel for the right agent+file
- [x] `npx vitest run frontend/app/view/{native-memory,armory,agent/components}` — 43 files / 541 tests passing (no consumers broken; the component's props are unchanged — it takes none)
- [x] `npx stylelint` on the touched SCSS — clean

See `docs/specs/SPEC_ARMORY_PERSONAL_MEMORY_AGENT_BLOCKS_2026_09_01.md`.


a5af:
@codex review

PR #2920, opened 2026-09-02
docs: merge the top-level specs/ tree into docs/specs/ (cleanup batch C)

AgentY-asaf:
## Why this wasn't just a file move

Two audits recommended this and neither executed it, so the first job was working out whether that six-week gap meant *"nobody got to it"* or *"something depends on the split"*. Neither, it turns out — it was a **deliberate design that had quietly inverted itself.**

The READMEs described a promotion workflow: `docs/specs/` = drafts, `specs/` = approved and in-flight, `specs/archive/` = done. Measured against the `Status:` field, every tree said the opposite of the truth:

| Tree | Documented as | Actually held |
|---|---|---|
| `specs/` | active and approved | 51 `draft`, 2 `implemented` |
| `docs/specs/` | drafts, not approved | **126 `implemented`**, 14 `active` |
| `specs/archive/` | completed or superseded | 7 `draft`, 3 `ready` |

An agent trusting those descriptions looked in exactly the wrong tree.

## It was costing correctness, not just navigation

Promoting a file between trees broke every code comment citing it, **silently**. Measured before the merge: **32 of 165 spec citations in source were already dangling** — 9 of them pointing at `specs/X` for a file that had long since moved to `docs/specs/X`. Nothing ever checked, so nobody knew.

The root cause: directory-as-lifecycle and the `Status:` field are two answers to the same question, and only one is enforced (`check-doc-status.sh`, batch E) or kept current. So the directories stop being an answer.

## What landed

- **134 files moved** (101 + 33 archive) — git detected all as renames, history preserved.
- **165 citations rewritten** across 128 files. 11 `SPEC_RETIRE_WSH` refs repointed into `archive/` where the file actually lives. Net dangling **48 → 46**.
- `specs/README.md` deleted; `docs/specs/README.md` now documents the single tree *and why*, so the split isn't recreated from the very instructions that created it.
- `check-doc-status.sh` drops its dead `specs` pathspec and gains a **hard guard** failing if a top-level `specs/` reappears — placed *before* the "no docs changed" early exit, since after it the guard was unreachable in exactly the case it exists for. Verified in both directions.
- **`INDEX.md`: 829 specs, up from 728.** Those 101 were invisible to the generated index entirely — it only ever scanned `docs/specs/`.

I also caught the generator's header claiming to cover *"every file in `docs/specs/`"* while excluding `archive/` — the same overclaim as the #2914 bug, found by re-reading rather than by a test. Now it states what it actually covers.

## Not fixed here, deliberately

**46 dangling `docs/specs/` citations remain, 43 pre-existing and untouched by this change.** Three point at files that exist nowhere in git history (`backend-lifecycle.md`, `SPEC_EDITOR_TABS`, `SPEC_BROWSER_PANE_TABS`) — inventing targets isn't relocation, and plan §3 scopes batch C to "relocated, not reviewed or rewritten".

That gap wants a **diff-scoped citation gate** as a follow-up, which I'll open next: a stale citation sends an agent hunting a file that doesn't exist, which is the same cost as a wrong Status.


a5af:
@codex review

AgentY-asaf:
## Both failures were real — fixed in `fbdb4c151`

### P1: the guard was bypassable, and my verification claim was wrong

Correct on both counts. The guard sets `fail=1`, but the next line exited `0` without consulting `$fail` whenever no `docs/*.md` were in the diff. Since the guard is **filesystem**-scoped and `files` is **diff**-scoped, a change that resurrects `specs/` *without touching a doc* — a bad rebase, precisely the case it exists for — sailed through green.

Worth naming how "verified in both directions" survived being false: my negative test passed an **explicit filename**, and the explicit-file branch never leaves `files` empty, so the test could not reach that line. The claim was true of the *check* and false of the *path*. Now tested on all three:

| Scenario | Expected | Got |
|---|---|---|
| `specs/` resurrected, **no** docs changed | fail | **exit 1** |
| `specs/` resurrected, docs changed | fail | **exit 1** |
| no `specs/`, no docs changed | pass | **exit 0** |

### CI: the move demanded a bulk restamp

The gate failed on ~100 relocated specs. Moving a file makes it "changed", so a diff-scoped gate demanded the Status vocabulary from four-month-old docs nobody had edited — the **exact bulk restamp plan §4 forbids**, since it replaces "unknown status" with "confidently wrong status".

Pure renames (`R100`) are now exempt: the gate enforces claims the author made in this diff, and relocating a file makes no claim about its state. A rename *with* edits (`R0xx`) is still checked, because then content did change.

**My first attempt at that exemption silently did nothing**, which is worth recording: `-- docs` hid the `specs/X.md` deletions, so git couldn't pair them and reported all 134 relocations as plain adds — `R100` never appeared, and the exemption was a no-op. Rename detection needs to see both sides. The `docs/` filter now runs in awk on the destination path, *after* pairing. Same class as globbing `specs/**/*.md` and missing root-level files: the filter ran before the thing it filtered was fully known.

### Six docs restatused, each read first

My citation rewrite touched them, so their non-canonical statuses became claims in this diff — the gate was right to catch them. Per §4, each was read rather than guessed:

| Doc | → | Basis |
|---|---|---|
| `ANALYSIS_WARDEN_AUTO_CONTROLLER` | `proposed` | says "Nothing in this document has been implemented" |
| `cross-window-drag-wiring` | `historical` | dated 2026-03-30 snapshot, not re-verified against current code |
| `REPORT_TRUST_CENTER_TO_ARMORY` | `historical` | completed sweep |
| `b5-migration-architecture` | `historical` | **not** `superseded` — it says "*partly*", and that status needs one successor that fully replaces it |
| `next-steps-2026-04-29` | `historical` | April planning doc |
| `retro-md-drop-window-hijack` | `historical` | a retro records what happened; it isn't a plan |

Gate now green: **22 docs checked**, renames excluded. Index confirmed current.

AgentY-asaf:
## P2 fixed — `da28c157d`. And it exposed a worse problem than the citations.

You're right on all three sites. `tear_off_hook.rs:38,182` and `drag.rs:725` name `specs/SPEC_CROSS_WINDOW_TAB_REMOUNT_2026_07_11` **without a `.md` extension**, and my rewrite regex required one — so they were never candidates, and a previously-valid path now resolved nowhere. Fixed.

**The reason I didn't catch it myself matters more than the miss.** My verification greps used a `(?<!...)` lookbehind. ripgrep's default engine doesn't support lookaround — it exits with a parse error. I had redirected stderr to `/dev/null`, so `wc -l` faithfully reported `0`.

> `=== any surviving top-level specs/ path refs? ===`
> `(count: 0)`

That "0" was never a measurement. It was an error message I had silenced. Exactly the shape of the vacuous-pass bug fixed earlier in `check-doc-status.sh`, where the gate printed `ok` after checking zero files — and I reintroduced it in my own verification one PR later.

**Re-verified without lookaround: 38 real hits**, triaged rather than bulk-rewritten:

| Category | Count | Action |
|---|---|---|
| Code citations missing `.md` | 3 | **Fixed** |
| `docs/README.md` describing a two-tier layout | 2 lines | **Fixed** — it's the entry point agents read first, and it told them to treat the top tree as "legacy/lower-traffic" |
| `../specs/` links from `docs/` subdirs | 17 | **Left alone, deliberately** — `docs/retro/../specs` *is* `docs/specs`, so these always pointed here and the move can't have broken them. 15 resolve; the 2 that don't were already in the pre-existing dangling set. |
| Prose in historical audits/retros describing the old tree | ~13 | **Left alone** — accurate records of a past state; rewriting them would make them wrong |
| `check-doc-status.sh` comments naming the old tree | 3 | Deliberate |

---

### Also resolved: the CI-only `STALE` failure was never a generator bug

CI tests the **PR merge commit**, not the branch head. `main` had gained `SPEC_ARMORY_PERSONAL_MEMORY_AGENT_BLOCKS_2026_09_01.md` plus its `INDEX.md` row. Git auto-merged that row into my regenerated index — it landed in a region my branch hadn't touched — while keeping my `proposed (82)` header. So the merged file had **83 rows under an 82 header**, and regeneration produced a diff of exactly one line.

It reproduced only on CI because the branch head is internally consistent at 829/82; the inconsistency existed *only in the merge*, the one state I never built. Merged `main` and regenerated.

Two diagnostics added while chasing it are staying, since a header-only diff is otherwise unreadable:

- **`--check` now prints the actual diff.** Previously it said only "stale", and re-deriving what changed was the entire cost of the failure.
- **Every run prints the candidate count** (`830 specs under docs/specs/`). Against a local 829, that identifies a stale branch instantly; inferring it by summing section headers is the guesswork this removes.

`LC_ALL=C` stays too — it wasn't this bug, but `--check` compares a committed file against a fresh run and has no business depending on collation.

PR #2921, opened 2026-09-02
fix(agent): stand down the Working row on promotion, move it above the composer

AgentX-asaf:
## Summary

Two changes to the agent pane's bottom stack, both about saying each fact once, in the place the eye already is.

### 1. The `Working…` row hides once a tool is auto-backgrounded

Promotion into the ActivityDock — immediately for a declared `sleep`, otherwise after `TOOL_PROMOTION_MS` — is the point at which work stops being *"the pane is blocked on this"* and becomes *"this runs in the background, tracked over there, with its own live countdown."*

The row kept rendering straight through. Two rows for one fact, and the louder one asserted the opposite of what had just happened: `Working…` says the pane is busy waiting, when backgrounding the task is exactly what made it not busy.

`toolPromoted` already reached `AgentWorkingRow`, but it only suppressed the **label** (`AgentFooter.tsx`), never the row — the spinner and elapsed timer stayed. With the new gate and `sessionStats` cleared at turn submit, the row now hides outright rather than falling back to a stale `✓ Worked · Ns`.

**Deliberately narrow.** Only the plain "a promoted tool is running and nothing else is happening" case is suppressed. Everything the dock has no vocabulary for keeps the row:

| State | Why the dock can't express it |
|---|---|
| launch activity | the pane isn't up yet |
| `Interrupting` | "Stopping…" is about the turn, not the tool |
| `waitingReason` (rate limited) | no dock representation — and the one the user most needs to see |
| compacting / reconnecting | not tools |

Extracted as a pure predicate with its own tests rather than a closure inside `agent-view.tsx`, because that exception list *is* the interesting part: the failure mode here is silently swallowing one of those five states.

### 2. Order is now dock → working row → composer

Bottom-up, scope narrows: background tasks → what this turn is doing right now → where you type. Previously the turn's own status sat *furthest* from the composer.

## This one deletes rather than adds

The row used to float over `.agent-document-scroll-region` as an absolutely-positioned overlay (`SPEC_..._SCROLL_FOLLOW_AND_STATUS_OVERLAY_2026_07_24` §3.2) so the message list's scrollbar could run the box's full height. That needed a surprising amount of scaffolding — **all of which goes**, because a row that isn't inside the scroll region can't overlap its scrollbar or its content in the first place:

- `workingRowHeight` signal + ref-callback `ResizeObserver`
- the `--agent-working-row-height` custom property bridging two component boundaries
- `.agent-document`'s matching `padding-bottom` reservation
- that prop threaded through `AgentDocumentView` → `AgentDocumentVirtualList`
- …and its use as a third stick-to-bottom dependency (reagent P1 on #2292)
- `.agent-working-row-backdrop`, an element existing solely to color the scrollbar gutter the overlay had to stay inset from
- the anchor's scrollbar-width inset, `z-index: 2`, and the `pointer-events: none`/`auto` dance
- `.agent-view--working-row-visible` and its margin rule

`SPEC_AGENT_WORKING_ROW_SCROLLBAR_GAP_2026_08_06.md` is superseded **in full**; the 07-24 spec only at §3.2 (its §2 scroll-follow analysis is untouched and still load-bearing).

Net **−315/+182**.

## Reviewer's real question: is scroll-follow still correct?

Yes, and it needs no replacement signal. As a normal-flow sibling the row changes the scroll region's **`clientHeight`**, not its content height — and `AgentDocumentVirtualList`'s `clientHeight` `ResizeObserver` already re-pins on exactly that. Its own comment names this family: *"the retry bar, AgentDecisionPanel, AgentQuestionPanel, or PendingMessagesPanel appearing/growing mid-turn — all normal-flow, flex-shrink:0 rows."* The working row joins it.

That observer was added specifically to end the per-panel-signal whack-a-mole; the overlay's tracked dependency was an instance of it. Removing it is a return to that design, not a regression waiting to happen.

## Verification

- `working-row-supersession.test.ts` — **9 tests**: the promotion case, the no-promotion case, all five keep-the-row exceptions, and two guards on reading optional `waitingReason` off a non-`Streaming` variant
- full `frontend/app/view/agent` suite — **1604 passing / 123 files**
- `npx tsc --noEmit` clean
- `vite build` clean (confirms the SCSS removals compile)

Spec: `docs/specs/SPEC_AGENT_WORKING_ROW_ABOVE_COMPOSER_2026_09_01.md`

### DIFF
--- a/.changesets/1788314332-fix-agent-hide-working-row-once-a-tool-is-promoted-to-the-do-6i96.md
+++ b/.changesets/1788314332-fix-agent-hide-working-row-once-a-tool-is-promoted-to-the-do-6i96.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): hide Working row once a tool is promoted to the dock; move it above the composer
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -201,7 +201,7 @@ deliberately do not.
 | [`SPEC_AGENT_TOOL_CALL_TONES_2026_06_05`](SPEC_AGENT_TOOL_CALL_TONES_2026_06_05.md) | SPEC — Agent tool-call tones (subliminal "talking" voice) |
 | [`SPEC_AGENT_TURN_PHASE_TIMELINE_LOGGING_2026_08_18`](SPEC_AGENT_TURN_PHASE_TIMELINE_LOGGING_2026_08_18.md) | SPEC: Agent turn-phase timeline — unified, replayable phase-history logging + `muxlog phases` |
 | [`SPEC_AGENT_VIEW_SCSS_SPLIT_2026_04_24`](SPEC_AGENT_VIEW_SCSS_SPLIT_2026_04_24.md) | Spec: agent-view.scss Decomposition |
-| [`SPEC_AGENT_WORKING_ROW_SCROLLBAR_GAP_2026_08_06`](SPEC_AGENT_WORKING_ROW_SCROLLBAR_GAP_2026_08_06.md) | Spec: Continuous AgentWorkingRow background through the scrollbar gutter |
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

Issue #2916, opened 2026-09-01
flaky: bashwrap a1_e2e test retries only exit 124, hard-fails on other nonzero exits


AgentY-asaf:
## Symptom

`agentmux-bashwrap::bash_wrap::tests::a1_e2e_static_label_then_delayed_cr_overwrite_collapses_to_one_line` failed on `check --tests + test (windows-latest)` for a **docs-only PR** (#2915 — three markdown files, zero Rust):

```
thread '...a1_e2e_static_label_then_delayed_cr_overwrite_collapses_to_one_line'
  panicked at agentmux-bashwrap\src\bash_wrap.rs:3588:13:
assertion `left == right` failed: command should exit cleanly
  left: 1
 right: 0
```

## Evidence it's a flake, not a regression

Two commits on the same branch, **identical Rust code**, differing by one markdown file:

| Commit | Windows job |
|---|---|
| `71346fb18` | **SUCCESS** |
| `5703ffd1d` | **FAILURE** (above) |

Re-running the failed job passed. So the same binary both passes and fails on the same runner image.

## Root cause

The test already has a 5-attempt retry loop built for loaded CI runners — but it only handles **two** of the three flaky outcomes:

- `result == 124` (idle-kill) → retried, with a comment naming "a loaded CI runner delaying process spawn/scheduling" as the cause
- PTY unavailable → skipped
- **any other nonzero exit → `assert_eq!(result, 0, ...)` panics immediately on attempt 1**, bypassing `MAX_ATTEMPTS` entirely

Exit `1` here is almost certainly the same loaded-runner class the `124` branch was written for (bash failing to start cleanly under contention), but it takes the un-retried path. The retry loop exists and does not cover it.

## Suggested fix

Treat any nonzero exit as a retryable attempt rather than special-casing `124` — record it in `last_outcome` (which already exists for exactly this reporting purpose) and `continue`, so the final panic reports all N attempts. That keeps a genuinely broken build failing (all 5 attempts fail) while not failing on one unlucky spawn.

Deliberately not fixing this in #2915 — that PR is a docs change and this is unrelated Rust. Filing so the diagnosis isn't lost.

PR #2922, opened 2026-09-02
docs: close out the cleanup plan — batch C shipped, and what it changed

AgentY-asaf:
## Summary

All five batches are on `main` (#2907, #2909, #2912, #2914, #2920). Status → `implemented`. Leaving C recorded as **"Held"** would be the same drift this document exists to stop — a correction it already carries once, from when it still said "nothing below has been executed yet".

Two additions to §7, both because **the plan turned out to be wrong in ways worth keeping**.

## Batch C was not the mechanical move §3 described

§3 scoped C to *"relocated, not reviewed or rewritten"*, treating the two trees as an accidental duplicate. They were a deliberate promotion workflow that had inverted itself:

| Tree | Documented as | Actually held |
|---|---|---|
| `specs/` | active and approved | 51 `draft`, 2 `implemented` |
| `docs/specs/` | drafts, not approved | **126 `implemented`**, 14 `active` |

An agent trusting those READMEs looked in exactly the wrong tree. And promoting a file between trees silently broke every code comment citing it — **32 of 165 citations were already dangling**.

So **§6's open question had a third answer** neither option anticipated. Not "nobody got to it", not "something outside depends on those paths", but that directory-as-lifecycle and the `Status:` field are two answers to one question, and only the enforced one stayed current.

### The generalisable bit

Relocating 134 files made them "changed", so batch E's gate demanded the `Status:` vocabulary from ~100 untouched four-month-old specs — the bulk restamp §4 forbids, **triggered by the very batch §5 sequenced *after* the gate so the gate would make it safe.** A cheap gate plus a mechanical change can still combine into a demand for unverifiable judgement, and the fix belongs in the gate. Pure renames are now exempt: relocating a file makes no claim about its state.

## Verification that was not verification

Three claims in this cleanup were false when made, all the same shape — **a check reporting success without checking anything**:

- *"No dangling links."* The glob was `docs/specs/*.md specs/*.md`, which cannot see `archive/` subdirectories or `docs/analysis/`. It reported clean for the part of the tree it could see.
- *"0 surviving references to the old tree."* The search used a `(?<!…)` lookbehind. ripgrep's default engine rejects lookaround, stderr was redirected to `/dev/null`, and `wc -l` faithfully reported `0`. **That zero was a silenced parse error**, and it hid three real citations found in review.
- *"The `specs/` guard is verified in both directions."* The negative test passed an explicit filename — a branch that cannot reach the line the guard was bypassing. True of the check, false of the path.

The batch-D generator failed the same way: 539 rows emitted for 728 specs under a header claiming to cover every file. **Its fix is the pattern worth copying** — an assertion the tool cannot pass while broken (emitted rows must equal candidate files, or it refuses to write), rather than a test that happens to exercise the working path.

## Remaining follow-ups

- A **diff-scoped citation gate** — 46 dangling `docs/specs/` citations remain, 43 pre-existing. A stale citation costs an agent the same as a wrong Status.
- **#2916** — the `bashwrap` flake found while landing these.

### DIFF
--- a/.changesets/1788317206-docs-close-out-the-docs-cleanup-plan-with-batch-c-1d4m.md
+++ b/.changesets/1788317206-docs-close-out-the-docs-cleanup-plan-with-batch-c-1d4m.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: close out the docs cleanup plan with batch C
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -178,10 +178,11 @@ are doing history". Everything else is here; the completeness
 assertion in the generator fails the build rather than emit a
 partial list.
 
-### implemented (128)
+### implemented (129)
 
 | Spec | Title |
 |---|---|
+| [`PLAN_DOCS_CLEANUP_EXECUTION_2026_09_01`](PLAN_DOCS_CLEANUP_EXECUTION_2026_09_01.md) | Docs cleanup — execution plan |
 | [`PLAN_MUXBUS_KEYCHAIN_WINDOWS_BLOB_LIMIT_2026_08_03`](PLAN_MUXBUS_KEYCHAIN_WINDOWS_BLOB_LIMIT_2026_08_03.md) | Plan — fix MuxBus token persistence on Windows (Credential Manager 2560-byte cap) |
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2923, opened 2026-09-02
ci: gate that a cited docs/specs path actually resolves

AgentY-asaf:
## Summary

The follow-up promised on #2920. This repo cites specs from code comments constantly — often the only link between a design and its implementation — and **nothing has ever checked those links.**

The cost surfaced when the two spec trees were merged: **32 of 165 spec citations in source were already dangling**, 9 of them naming a path for a file that had moved months earlier. Nobody knew, because nothing looked. Each one sends the next reader hunting for a document that isn't there — the same cost as a wrong `Status:` line, and harder to notice.

## Scoped to changed files, and the measurement is why

Measured with this script over every tracked `.md/.rs/.ts/.tsx/.sh/.yml`:

> **90 dangling citations, across 78 files, naming 49 distinct missing specs**

A repo-wide gate would fail every PR on somebody else's rot and be switched off within a day — the documented fate of the three previous attempts at docs enforcement here. Applied to the diff it's green for anyone who hasn't touched a citation, and it stops 90 becoming 91.

Same philosophy as `check-doc-status.sh`: enforce claims made *in this diff*, don't demand a cleanup nobody signed up for.

*(An earlier revision of this PR quoted 46 and 89 in different places. Both were wrong — see the review thread; the three numbers above count different things and the script header now says which is which.)*

## Two deliberate refusals to be clever

Both aimed squarely at not getting disabled:

- **Glob patterns are skipped.** This script and its sibling both discuss `docs/specs/*.md` as a pattern. Failing on something never meant to name one file is exactly the false positive that kills a gate.
- **Pure renames are skipped**, same as `check-doc-status.sh` — relocating a file asserts nothing. The diff runs **without a pathspec**, because rename detection needs both sides of a move to pair them; limiting it hides the deletions and silently turns the exemption into a no-op. That's a bug this repo has now shipped once (#2920), so it's called out in the code.

The failure message says what to do, including the case people get wrong:

> If the spec never existed or is long gone, delete the path from the comment rather than leaving it… Do not invent a plausible-looking replacement path to satisfy this check.

## Verified

| Test | Expected | Got |
|---|---|---|
| File with a dangling citation | fail | **exit 1**, names the path |
| File with a resolving citation | pass | **exit 0** |
| File containing `docs/specs/*.md` glob | pass | **exit 0** |
| No files changed | pass | `nothing changed` |
| The gate running on its own PR | pass | **green** |

Wired into `ci-pr.yml` beside the existing doc gates, and exposed as `task check:spec-citations`.

### DIFF
--- a/.changesets/1788318108-ci-gate-that-a-cited-docs-specs-path-actually-resolves-pbp0.md
+++ b/.changesets/1788318108-ci-gate-that-a-cited-docs-specs-path-actually-resolves-pbp0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+ci: gate that a cited docs/specs path actually resolves
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -80,6 +80,9 @@ jobs:
       - name: Specs index is current
         run: bash scripts/gen-docs-index.sh --check
 
+      - name: Spec citations resolve (changed files only)
+        run: bash scripts/check-spec-citations.sh
+
       # The three gates below already existed as Taskfile `check:*` tasks but
       # were wired into nothing — defined, documented, and never run. Each
       # encodes a regression that already happened once (see the spec/retro in
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -201,6 +201,15 @@ tasks:
             See docs/specs/PLAN_DOCS_CLEANUP_EXECUTION_2026_09_01.md batch E.
         cmd: bash scripts/check-doc-status.sh
 
+    check:spec-citations:
+        desc: >-
+            Gate — fails if a CHANGED file cites a `docs/specs/*.md` path that
+            does not exist. Scoped to the diff for the same reason as
+            check:doc-status: 89 files carry a dangling citation today, so a
+            repo-wide gate would fail every PR on somebody else's rot and be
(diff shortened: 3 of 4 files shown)


a5af:
@codex review

AgentY-asaf:
## P2 fixed — and it's the right thing to be pedantic about here

A gate for catching inaccurate references shipped with **three inconsistent counts of its own**:

| Where | Claimed |
|---|---|
| script header | 46 dangling citations |
| Taskfile desc | 89 files |
| PR body | 89 files |

All three wrong, in different ways. **46** was a stale count of distinct missing *paths*. **89** came from counting `FAIL` lines — which the script emits once per *citation*, not per file — so it was never a file count, and I labelled it one.

Measured properly with the script itself, over every tracked `.md/.rs/.ts/.tsx/.sh/.yml`:

> **90 dangling citations, across 78 files, naming 49 distinct missing specs**

Which matches your independent ~78 / ~90 estimate.

The header now carries all three with what each one counts, plus a note that they're easy to conflate — one file can hold several dangling citations, and one missing spec can be cited from many files. The PR body above is corrected too.

The underlying mistake is the one this whole cleanup keeps turning up: **a number quoted from a command whose output means something slightly different from the label put on it.** Same shape as `grep -c $'\r'` reporting every line because the pattern collapsed to empty, and `wc -l` reporting a silenced parse error as a confident zero. Three instances in two days, all caught in review rather than by me — which is why the batch-D fix was an assertion the tool can't pass while broken, rather than another test.

PR #2924, opened 2026-09-02
test(bashwrap): retry any nonzero exit in a1_e2e, not just idle-kill

### LINKED ISSUE #2916: flaky: bashwrap a1_e2e test retries only exit 124, hard-fails on other nonzero exits

AgentY-asaf:
Fixes #2916.

## The failure

`a1_e2e_static_label_then_delayed_cr_overwrite_collapses_to_one_line` failed on a **docs-only PR** (#2915 — three markdown files, zero Rust):

```
assertion `left == right` failed: command should exit cleanly
  left: 1
 right: 0
```

## The cause

The test already has a 5-attempt retry loop built for loaded CI runners. It covered **two of the three** flaky outcomes:

| Outcome | Handling |
|---|---|
| `result == 124` (idle-kill) | retried — the comment names *"a loaded CI runner delaying process spawn/scheduling"* |
| PTY unavailable | skipped |
| **any other nonzero** | **`assert_eq!` panics on attempt 1, bypassing `MAX_ATTEMPTS` entirely** |

Exit `1` is the same loaded-runner class the `124` branch exists for — from inside the loop the two are indistinguishable — but it took the un-retried path. **A retry loop that doesn't cover the case that actually happens is decoration.**

## Evidence it's a flake, not a regression

Two commits on the same branch, **byte-identical Rust**, differing by one markdown file:

| Commit | Windows job |
|---|---|
| `71346fb18` | SUCCESS |
| `5703ffd1d` | FAILURE |

Re-running the failed job passed.

## The fix, and what it doesn't weaken

Any nonzero exit is now retried, matching the `124` branch. **A genuinely broken build still fails** — all 5 attempts must exit nonzero — and the panic now reports `last_outcome` rather than a bare assertion.

That reporting gap is worth noting: it was already fixed once for the idle-kill path (reagentx P2 on #2559, whose comment says `last_outcome` "must be updated on EVERY retry path"). This branch reintroduced it by not existing.

Verified: compiles, test passes locally.

### DIFF
--- a/.changesets/1788319484-test-bashwrap-retry-any-nonzero-exit-in-the-a1-e2e-test-not--wy11.md
+++ b/.changesets/1788319484-test-bashwrap-retry-any-nonzero-exit-in-the-a1-e2e-test-not--wy11.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+test(bashwrap): retry any nonzero exit in the a1 e2e test, not just idle-kill
--- a/agentmux-bashwrap/src/bash_wrap.rs
+++ b/agentmux-bashwrap/src/bash_wrap.rs
@@ -3585,7 +3585,29 @@ mod tests {
                 );
                 continue;
             }
-            assert_eq!(result, 0, "command should exit cleanly");
+            if result != 0 {
+                // Retried for the same reason 124 is, immediately above: on a
+                // loaded runner bash can fail to start cleanly, and from here
+                // that is indistinguishable from being idle-killed.
+                //
+                // This was `assert_eq!(result, 0, "command should exit cleanly")`,
+                // which failed the whole test on attempt 1 and bypassed
+                // MAX_ATTEMPTS entirely. The retry loop existed but covered only
+                // two of the three flaky outcomes, and the one it missed is the
+                // one that fired — exit 1, failing a docs-only PR that touched
+                // no Rust at all (issue #2916). A retry loop that does not cover
+                // the case that actually happens is decoration.
+                //
+                // A genuinely broken build still fails: every one of
+                // MAX_ATTEMPTS must exit nonzero, and the panic below reports
+                // the last outcome, which a bare assert_eq! never did.
+                last_outcome = format!(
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #1588, opened 2026-06-19
feat(native-memory): backend RPCs for agent memory file browser (Phase 2)

AgentA-asaf:
## Summary

Phase 2 of the agent native memory browser (spec §7). Adds 3 backend RPC commands that Phase 3 (full UI modal) will call:

- **`agent:memory:list`** — list `*.md` files in `~/.claude/projects/<sanitized>/memory/`, sorted MEMORY.md first then alpha; returns `NativeMemoryFileMeta[]` with filename, `is_index`, frontmatter `metadata_type`, size, mtime
- **`agent:memory:read_file`** — read one file by name; blocks path traversal (rejects `/`, `\`, `..`)
- **`agent:memory:write_file`** — atomic write via `.tmp`→rename; validates filename (`[a-zA-Z0-9\-_]+.md`) and creates dir if needed

Path algorithm mirrors Claude Code's `sessionStoragePortable.ts`: replace non-alphanumeric chars with `-`; if >200 chars, truncate at 200 + base-36 djb2 hash suffix.

## Files changed

| File | Change |
|------|--------|
| `agentmux-srv/src/server/native_memory_handlers.rs` | NEW — 3 handler implementations |
| `agentmux-srv/src/backend/rpc_types.rs` | 3 command constants + request/response structs |
| `agentmux-srv/src/server/mod.rs` | `mod native_memory_handlers` |
| `agentmux-srv/src/server/websocket.rs` | `register_native_memory_handlers` call |
| `frontend/types/gotypes.d.ts` | `NativeMemoryFileMeta`, `NativeMemoryListResult`, `NativeMemoryReadFileResult` |
| `frontend/app/store/rpc-api.ts` | `NativeMemoryListCommand`, `NativeMemoryReadFileCommand`, `NativeMemoryWriteFileCommand` |

## Test plan

- [ ] `cargo check -p agentmux-srv` clean ✅
- [ ] `npx tsc --noEmit` clean ✅
- [ ] Phase 3 UI will exercise all 3 RPCs end-to-end

Follows: #1587 (Phase 1 — brain+id-card icons)
Next: Phase 3 — `AgentNativeMemoryModel` + full file browser UI

Spec: `docs/specs/SPEC_AGENT_PANE_MEMORY_IDENTITY_MODALS_2026_06_19.md` §7

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #2926, opened 2026-09-02
Manoz@Area54: fix(memory): agent:memory RPC handlers never adopted #2901's blank-workdir fix

GenericAgentX-asaf:
## Summary

Live-tested Personal Memory end to end against a running v0.55.31 build (asked to verify a written memory actually surfaces in the Armory UI): a `MemoryWrite` MCP call succeeded and the file existed on disk, but the Armory grid showed "No memories yet" for that exact agent, and calling `agent:memory:write_file` directly for the same agent failed outright with *"has no configured working directory."*

**Root cause:** `SPEC_FIX_PERSONAL_MEMORY_EMPTY_WORKDIR_2026_09_01.md` (#2901) fixed blank-`working_directory` resolution inside `memory_dir_for_agent`/`memory_dir_for_agent_by_id` — but only the App API (what `MemoryWrite`/`MemoryList` MCP tools actually call) and `bundle.rs` use those resolvers. The four WebSocket RPC handlers backing the Armory UI itself — `list`/`read_file`/`write_file`/`revert` — had their own separate, never-updated inline `agent.working_directory.is_empty()` check: the exact pre-#2901 behavior (`list`'s dates to PR #1588), sitting right next to the fixed code without ever calling it. `memory_dir_for_agent_by_id`'s own doc comment already (incorrectly) claimed these handlers used it.

Since a blank `working_directory` is the *default* state (`agent.open` substitutes a derived directory at spawn time rather than persisting it back to the definition row), this meant Personal Memory silently showed "no memories" for any such agent's real files — indistinguishable from genuinely empty, since `list` never errored. Exactly the trap #2917's four-state card design exists to catch, via a different root cause than #2901 was built against.

## Fix

Repoint all four handlers at `memory_dir_for_agent_by_id`, the same resolver `bundle.rs` and `native_memory_drift.rs` already use. `history`/`diff` are unaffected — they resolve purely through `id_store`/`agent.id`, never touch `working_directory`.

## How this was found

Live CDP testing against the running build (`Runtime.evaluate` against the app's own exposed `window.RpcApi`), not inferred from reading the code — reading the code alone plausibly would have missed it, since `memory_dir_for_agent_by_id`'s own doc comment claims the handlers already use it.

## Test plan
- [x] Three new handler-level regression tests, driving the real `WshRpcEngine` dispatch (not the resolver function directly — that path was already covered elsewhere):
  - `write_file_resolves_a_blank_working_directory_instead_of_erroring`
  - `list_finds_files_already_on_disk_at_the_derived_default_dir_for_a_blank_workdir_agent` — proves `list`'s own resolution is fixed, not just piggybacking on `write_file` now sharing the same path
  - `revert_resolves_a_blank_working_directory_instead_of_erroring`
- [x] `cargo test -p agentmux-srv --bin agentmux-srv native_memory_handlers::tests` — 33/33 passing
- [x] `cargo test -p agentmux-srv --bin agentmux-srv server::app_api::bundle` — 55/55 passing (no regression in the callers that already used the fixed resolver)
- [x] `cargo build --release -p agentmux-srv` — clean, zero new warnings
- [x] Live re-verified: swapped the fixed binary into the actual v0.55.31 portable build and relaunched — no code-path regressions surfaced (the file-lock/relaunch friction encountered was OS/environment noise, not related to this diff)

See `docs/specs/SPEC_MEMORY_RPC_HANDLERS_BLANK_WORKDIR_2026_09_02.md`.

### DIFF
--- a/.changesets/1788357580-fix-memory-agent-memory-rpc-handlers-never-adopted-2901-s-bl-51oi.md
+++ b/.changesets/1788357580-fix-memory-agent-memory-rpc-handlers-never-adopted-2901-s-bl-51oi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(memory): agent:memory RPC handlers never adopted #2901's blank-workdir fix
--- a/agentmux-srv/src/server/native_memory_handlers.rs
+++ b/agentmux-srv/src/server/native_memory_handlers.rs
@@ -785,18 +785,22 @@ pub fn register_native_memory_handlers(engine: &Arc<WshRpcEngine>, state: &AppSt
                     .map_err(|e| format!("agent:memory:list: store: {e}"))?
                     .ok_or_else(|| format!("agent:memory:list: agent {} not found", cmd.agent_id))?;
 
-                // No working directory → no memory path (avoid mapping to the shared
-                // ~/.claude/projects/memory/ directory — reagent P1 on PR #1588).
-                if agent.working_directory.is_empty() {
-                    return Ok(Some(serde_json::to_value(NativeMemoryListResult { files: vec![] }).map_err(|e| e.to_string())?));
-                }
-
-                let config_dir = wstore
-                    .agent_content_get(&agent.id, "env")
-                    .ok().flatten()
-                    .map(|c| parse_claude_config_dir(&c.content))
-                    .unwrap_or_default();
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2927, opened 2026-09-02
Manoz@Area54: feat(tabs): default tab names are "Tab 1", "Tab 2", not "tab1", "tab2"

GenericAgentX-asaf:
## Summary

Direct request: change the default names for new tabs from `tab1`, `tab2`, … to `Tab 1`, `Tab 2`, ….

`SPEC_TAB_GAPS_AND_NAMING_2026_04_25.md` deliberately unified two prior inconsistent default names (`Untitled1`, `T1`) onto plain `tab1`/`tab2` — a considered choice at the time, not an oversight. This PR revises that choice on direct request.

## What changed

Four independently-maintained copies of the same one-line `format!("tab{}", …)` string, updated to `format!("Tab {}", …)`:

- `wcore::create_tab` — the "+" button path
- `wcore/window.rs` — new-window tab bootstrap
- `reducer/tab.rs` — the `CreateTab` reducer (TearOffBlock / fresh-workspace paths)
- `tab_lifecycle.rs::handle_create_tab` RPC handler — both its computed-count branch and its degenerate `"tab1"` fallback

(These four sites already existed as separate copies before this change — noted in the spec as a candidate for a shared helper if a fifth copy ever needs syncing with these four again, same shape as the duplication #2926 just fixed for memory-dir resolution. Not addressed here — kept this PR scoped to the rename that was actually asked for.)

## Non-goals
- **No migration** of already-existing tabs named `tab1`/`tab2` (or the even older `Untitled1`/`T1`) — same policy the original spec used for its own predecessors. Only newly-created tabs (empty name passed to `CreateTab`) get the new format.
- **No change to user-renamed tabs** — the empty-name branch is untouched.
- **No frontend change** — naming is entirely backend-side; no `defaultTabName()`-equivalent exists in the frontend.

## Test plan
- [x] Five test assertions across three files updated (`store/tests.rs`, `wcore/mod.rs`, `reducer/tab.rs` ×2) to expect `"Tab 1"`/`"Tab 2"`
- [x] Full `cargo test -p agentmux-srv --bin agentmux-srv` — **2920 passed, 0 failed** (confirms no other fixture depended on the literal string; several other `"tab1"` literals in the codebase are `tab_id` identifiers, unrelated to the display name, and correctly left untouched)
- [x] `cargo build -p agentmux-srv` — clean, no new warnings

See `docs/specs/SPEC_DEFAULT_TAB_NAME_TAB_N_2026_09_02.md`.

### DIFF
--- a/.changesets/1788358164-feat-tabs-default-tab-names-are-tab-1-tab-2-not-tab1-tab2-qpuo.md
+++ b/.changesets/1788358164-feat-tabs-default-tab-names-are-tab-1-tab-2-not-tab1-tab2-qpuo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(tabs): default tab names are "Tab 1", "Tab 2", not "tab1", "tab2"
--- a/agentmux-srv/src/backend/storage/store/tests.rs
+++ b/agentmux-srv/src/backend/storage/store/tests.rs
@@ -234,8 +234,9 @@
 
                 let mut tab = Tab {
                     oid: "tab-tx".to_string(),
-                    // tabN naming convention per SPEC_TAB_GAPS_AND_NAMING_2026_04_25.
-                    name: "tab1".to_string(),
+                    // "Tab N" naming convention per SPEC_TAB_GAPS_AND_NAMING_2026_04_25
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #2928, opened 2026-09-02
Posa@Area54: fix(agent): retry a compaction_started ping missed before turn-phase reconciliation

GenericAgentX-asaf:
## Summary
- After loading/resuming an agent, the "Working…" status row could disappear mid-compaction: a `compaction_started` WPS ping can arrive before `ReconcileTurnActive` promotes a freshly-mounted pane's `turnPhase` out of the mount-default `Idle`. The reducer's `workingFromPhase` gate correctly rejects that early ping (round 5's orphan-state guard, PR #2378), but since this transport is `persist: 0` (never replayed), the drop was permanent — `compacting` stayed `null` for the whole compaction and the liveness watchdog eventually force-demoted the phase, hiding the busy indicator well before the real `compact_boundary` arrived.
- `useCompactionStream.ts` now buffers a rejected ping locally and retries the exact same dispatch once `turnPhase` is next confirmed working, reusing the reducer's existing accept/reject gates unchanged (no reducer changes — the 11 prior race-hardening rounds on `CompactionStarted` are untouched).
- See `docs/specs/SPEC_COMPACTION_STARTED_RECONCILIATION_RACE_2026_09_02.md` for the full writeup.

## Test plan
- [x] New tests in `useCompactionStream.test.ts` (TDD: written first, confirmed red, then green): buffers-and-retries, retries-at-most-once, no-op-when-nothing-missed, no-spurious-retry-while-already-working.
- [x] `frontend/app/view/agent` + `frontend/app/store/agent-pane-state` suites: 126 files / 1850 tests pass.
- [x] `tsc --noEmit` clean.
- [x] Changeset added (`patch`).


a5af:
@codex review

GenericAgentX-asaf:
Addressed both findings in e615c302e:

- **reagent P1** (missedPing re-armed on a failed retry, causing unbounded re-dispatch on every subsequent turnPhase change): `attemptStart` now only buffers on the live-delivery path (`bufferOnReject=true`); a failed retry always clears `missedPing` regardless of outcome. New test: "gives up after one failed retry — does not keep re-dispatching on every later turnPhase change".

- **codex P2** (a genuinely-stale buffered ping could survive to replay against a later unrelated turn if its confirming `ReconcileTurnActive(active: false)` was a same-ref no-op): the retry is now bounded to `MISSED_PING_RETRY_WINDOW_MS` (15s) from the original rejection — an expired buffered ping is dropped silently instead of retried. New test: "drops an expired buffered ping instead of replaying it into a later, unrelated turn".

Both new tests were confirmed to fail against the prior commit before the fix, then pass after. Full `frontend/app/view/agent` + `frontend/app/store/agent-pane-state` suites (126 files / 1852 tests) and `tsc --noEmit` are clean.

GenericAgentX-asaf:

a5af:
@codex review

GenericAgentX-asaf:
Both of you (reagent P1 + codex P2, independently, across two rounds) found the same root problem in the hook-only fix: `useCompactionStream.ts`'s retry effect could only observe the *resulting* `turnPhase` value, so it couldn't distinguish "this promotion is the same turn the buffered ping was about" from "this is a later, unrelated turn" — and a bounded time window only narrows that gap in both directions without closing it (too short regresses the original bug, too long still allows the false replay).

Redesigned in 66f97292: the fix now lives in the reducer (`state.pendingCompactionPing`), not the hook, because the reducer sees every dispatched command as a discrete event regardless of whether it changes the resulting state — it can react to `ReconcileTurnActive(active: false)` as a real signal even when `turnPhase` stays `Idle` either way (the exact case that broke the hook-only version). `useCompactionStream.ts` reverts to its original plain dispatch-and-check form.

Summary of the new design (full writeup in the rewritten spec §3-5):
- `CompactionStarted` buffers onto `pendingCompactionPing` instead of dropping, but only when the phase is one a later authoritative signal could still legitimately promote (`Idle`/`Disconnected`/`Done.completed` — the same set `StreamFlushObserved`'s own promotion arm already uses).
- Promoted into `compacting` by `ReconcileTurnActive(active: true)` and `StreamFlushObserved`'s existing promotion arm — both authoritative "this turn is genuinely active" signals.
- Discarded (not promoted) by `TurnStart` — closes codex's exact scenario, since without this the very next `StreamFlushObserved` (which fires for practically every `TurnStart`) would otherwise inherit the stale ping.
- Cleared unconditionally by `ReconcileTurnActive(active: false)`, and by every other existing "compaction now moot" terminal transition.

New reducer-level test coverage for all of the above (buffer/promote/discard/supersede) in `reducer.test.ts`. Full `frontend/app/view/agent` + `frontend/app/store/agent-pane-state` suites (126 files / 1859 tests) and `tsc --noEmit` are clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
Fixed all three findings from the latest round (97c29178):

1. **StreamUnsubscribe / TurnEnd early-branch gap** — both early same-ref no-op branches now clear `pendingCompactionPing` when set, not just the main working-phase paths.
2. **Transcript-node wiring** — moved the "Compacting conversation…" push in `useCompactionStream.ts` from inspecting individual dispatch return values (which neither `agent-view.tsx`'s `ReconcileTurnActive` nor `stream-flush-queue.ts`'s `StreamFlushObserved` call site actually captured) to watching the reactive `compacting` signal itself, which `registerAgentPane`'s generic projection already keeps in sync regardless of which dispatch set it. One `createEffect`, correct for every promotion path, no per-call-site wiring needed. `wasCompactionStartedAccepted` removed as now-unreachable.

New test coverage for both. Full `frontend/app/view/agent` + `frontend/app/store/agent-pane-state` suites (126 files / 1864 tests) and `tsc --noEmit` clean. Spec §4 rewritten with the final design.

GenericAgentX-asaf:

PR #2929, opened 2026-09-02
Manoz@Area54: feat(armory): find/filter and sort for Personal Memory, mirroring My Agents

GenericAgentX-asaf:
## Summary

Direct request: add find/filter to Armory → Memory → Personal, "similar to My Agents in the agent pane." The grid PR #2917 added routinely renders 30+ cards in practice — past the point that spec explicitly flagged as the trigger to revisit its own deferred filter/sort non-goal.

## What's new

New `MemoryAgentFilterBar` mirrors `AgentPickerFilterBar`'s interaction pattern (magnifying-glass input, clear button, Escape-to-clear, sort `<select>` pinned right) **without reusing that component** — its `AgentSortOption` (`"recent"/"name"/"type"`) is launch-oriented and reads session-launch data this grid doesn't have. New `MemoryAgentSortOption` (`"name"/"count"/"provider"`) instead.

Filtering is pure client-side over the already-fully-loaded agent list — unlike `MyAgentsList`, this view has no RPC pagination to mirror, so no debounce needed.

Also added a **"Has memories" toggle** beyond what My Agents has: on this machine's live grid (confirmed during #2917's own testing), most of the 33 rendered cards read "No memories yet" — a real difference from My Agents, where every listed agent is inherently launchable.

## The one subtle piece — and a bug I caught while writing its test

**"Most files" sort ranks `loading`/`error` cards after every resolved count, never mixed in.** A numeric sort has no correct answer for a count that isn't known yet or failed, so burying those at the bottom is intentional, not a fallback bug.

The "Has memories" toggle follows the same principle from the other direction: it should only ever hide a resolved `count: 0` card, **never** `loading` or `error` — a slow or failing fetch shouldn't make a card vanish out from under the user. **My first-pass implementation had this condition inverted** (`if (!(c?.kind === "count" && c.files > 0))` — which also excludes `loading`/`error`, not just zero) and a dedicated test caught it immediately. Fixed; the regression test now covers exactly this case.

## Persistence

Sort choice persists to `localStorage` (`nativeMemory:sortBy`), matching `AgentPickerFilterBar`'s own precedent — **the operative default is "last used"**, not the hardcoded `"name"` fallback (which only applies once, before any preference has ever been set). Filter text and the "Has memories" toggle are session-only, same convention as `AgentPicker`'s own `filterQuery`.

## Test plan
- [x] `npx vitest run frontend/app/view/native-memory` — **20/20 passing** (12 new: name filter, clear (button + Escape), zero-match empty state distinct from zero-agents-total, all three sort modes including the loading/error-last rule, the toggle bug fix + its regression test, sort persistence across remount while filter/toggle don't persist, bar absent from detail view)
- [x] `npx vitest run frontend/app/view/{native-memory,armory,agent/components}` — 43 files / 553 tests passing, no regressions
- [x] `npx stylelint` on the touched SCSS — clean

See `docs/specs/SPEC_ARMORY_PERSONAL_MEMORY_FILTER_AND_SORT_2026_09_02.md`.


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #2931, opened 2026-09-02
Manoz@Area54: fix(armory): remove a NUL byte hiding native-memory-manager.tsx's diff since #2917

GenericAgentX-asaf:
## Summary

`agentIdsKey`'s `.join()` separator in `native-memory-manager.tsx` has been a literal NUL byte (`0x00`), not the space character it visually renders as, since this file was **first introduced by #2917** — a tool-call artifact, not anything deliberate.

`git diff` / GitHub both treat a file as binary once it contains a NUL byte. That means **every review of this file since #2917 merged** — #2917 itself, and every commit in #2929 that touched it — saw `Binary files ... differ` / `+0 -0` and had nothing to comment on, regardless of what the actual change was.

## How this was caught

ReAgent's P1 finding on #2929 (which found the byte reintroduced/carried-forward on that PR's own branch — fixing it there only fixed that branch's copy). `main`'s own copy needed the same fix independently, since #2929's branch and `main` are different blobs at this path. This PR is that fix, targeted at `main` directly.

## Verification

Verified byte-precisely, not via `grep` (which mishandles line-counting on NUL content in exactly the way that made this bug so persistent):

```
NUL byte count before fix: 1   (at the exact line ReAgent's finding pointed at)
NUL byte count after fix:  0
```

Swept the **entire tracked source tree** (1619 `.ts`/`.tsx`/`.rs`/`.scss` files) for the same pattern — no other file affected.

## A note for whoever reviews this

**This PR's own diff will also render as `Binary files differ` on GitHub.** That's inherent to fixing a file where one side of the diff (`main`, pre-fix) still contains the NUL byte — it is not a sign the fix didn't work. Verify via raw byte content instead (`git show <sha>:<path> | perl -ne 'print if /\x00/'` should print nothing), or trust `tsc`/`vitest` below, which exercise the real code on both sides regardless of what the diff viewer can render.

## Test plan
- [x] Byte-precise verification (above): 1 → 0
- [x] Repo-wide sweep, 1619 files, no other occurrences
- [x] `npx vitest run frontend/app/view/native-memory` — 10/10 passing (this branch is `main`'s current shape, pre-#2929's filter/sort additions)


a5af:
@codex review

GenericAgentX-asaf:
Superseded — #2929 merged first, and since that branch already carried its own fix for this same byte, merging it fixed `main`'s copy as a side effect (verified: `main`'s current copy of this file has 0 NUL bytes). This PR's own diff would now be a no-op. Closing rather than merging redundant content.

PR #2932, opened 2026-09-02
Manoz@Area54: feat(armory): reactive updates across the Armory

GenericAgentX-asaf:
## Summary

Direct question, then a direct request to broaden it: *"if you write a memory, will I see the update in Manoz in the armory? we want it reactive"* → *"build the reactive updates, across the entire armory in fact."*

**Audited first rather than assuming a from-scratch build was needed.** Grepped every `WaveEvent` the backend publishes, then checked which frontend view-models actually subscribe:

| Area | Backend event | Already wired? |
|---|---|---|
| Accounts | `identityaccounts:changed` | **Yes** — shared `IdentityViewModel`, nothing to do |
| Per-agent MCP/Skill bindings | `mcp:changed` / `skills:changed` | **Yes** — `bundle-mcp-model.ts`/`bundle-skill-model.ts` already subscribe |
| Global Memory | `memories:changed` | No — event existed, no subscriber |
| ABF (bundle list) | `memories:changed` | No — same gap |
| MCP Servers (standalone tab) | `mcp:changed` | No — same gap |
| Skills (standalone tab) | `skills:changed` | No — same gap |
| **Personal (native) Memory** | *(none existed)* | **No — the one real gap** |

Four of five gaps turned out to be a 3-line fix each (wire an already-correct event to a `waveEventSubscribe` call, same pattern their sibling models already use). Personal Memory needed actual backend work.

## A mid-implementation discovery that changed the design

`git log -S` on the planned event name turned up that **two of the five Personal Memory call sites already published something** — `app_api::mod`'s `memory_write_impl`/`memory_revert_impl` (commit `bcc478dc`, 2026-08-20, a different agent's work) — just under a different convention (the agent id baked into the event **name**, `agent:memory:changed:{id}`, not `WaveEvent.scopes`, my original plan) and with **zero subscribers anywhere**, so it was a silent no-op in production.

Adopted the existing convention everywhere instead of introducing a second one. And fixed a real bug found in the process: **those two existing calls keyed the event by the raw `agent_id` slug parameter, not the resolved canonical UUID** — the exact slug/UUID keyspace split this same file's own `memory_write_impl` doc comment already warns about for version storage, just never applied to this event. A frontend subscriber only ever has `AgentDefinition.id` (the UUID), so the pre-existing calls would have gone unheard even once something finally subscribed. Regression test: `write_publishes_using_the_resolved_uuid_not_the_raw_slug`.

## What changed

**Backend** — publish added/fixed at all five real call sites (`native_memory_handlers.rs` write_file + revert, `native_memory_drift.rs` fast-path + reconciliation sweep, `app_api::mod` write + revert re-keyed). Drift detector only publishes on an *actual* recorded change — never on a no-op sweep tick (would otherwise fire for every watched agent every 30s regardless of whether anything happened). Fixed a small pre-existing log inaccuracy in the same branch I had to touch anyway (fast path logged "recorded" even on the no-op case).

**Frontend** — `NativeMemoryManager` subscribes to `agent:memory:changed:{id}` once per agent currently in the grid, re-registering when the agent *set* changes (same `agentIdsKey` dependency the existing count-fetch effect uses). On an event: refetches just that one card's count (never a full-grid refresh), and refreshes the open detail view's file list in place if it's showing the changed agent — without clearing the selected filename unless that exact file is gone. Debounced 250ms per agent id.

The other four: `global-brain-model.ts`, `memory-model.ts`, `mcp-model.ts`, `skill-model.ts` each gained a 3-line `waveEventSubscribe` + `dispose()` unsubscribe, mirroring their already-working sibling models.

## Test plan
- [x] 5 new backend publish-verification tests (write_file, revert, drift sweep ×2, App API slug/UUID regression) — `cargo test -p agentmux-srv --bin agentmux-srv publish` → 27/27 passing
- [x] Full backend suite: **2930/2930 passing**, 0 failures
- [x] 8 new frontend tests in `native-memory-manager.test.tsx` (per-agent refetch, error-state recovery, debounce, detail-view in-place refresh, cross-agent isolation, selected-file-removed handling, resubscribe-on-agent-set-change)
- [x] 2 new tests each in `global-brain-model.test.ts`/`memory-model.test.ts`; 2 new focused test files for `mcp-model.ts`/`skill-model.ts` (neither had prior coverage)
- [x] Full frontend suite: **3521/3521 passing**, 0 failures

See `docs/specs/SPEC_ARMORY_REACTIVE_UPDATES_2026_09_02.md`.

### DIFF
--- a/.changesets/1788364392-feat-armory-reactive-updates-across-the-armory-live-refreshi-37l2.md
+++ b/.changesets/1788364392-feat-armory-reactive-updates-across-the-armory-live-refreshi-37l2.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(armory): reactive updates across the Armory (live-refreshing Personal Memory, Global Memory, ABF, MCP Servers, Skills)
--- a/agentmux-srv/src/backend/native_memory_drift.rs
+++ b/agentmux-srv/src/backend/native_memory_drift.rs
@@ -116,10 +116,68 @@ fn read_memory_file_lossy(path: &Path) -> std::io::Result<String> {
 /// tests assert on it directly). Errors from an individual agent/file are
 /// logged and swallowed — one agent's bad state (e.g. a permissions issue)
 /// must not stop the sweep from covering every other agent.
-pub(crate) fn reconciliation_sweep_once(wstore: &Store, id_store: &Store, broker: &crate::backend::wps::Broker) -> usize {
+///
+/// Also detects and publishes OUT-OF-BAND DELETIONS (Codex P2, PR #2932):
+/// content-drift detection alone only ever compares files that still exist
+/// against their last recorded version — a `.md` file removed directly (not
+/// through `agent:memory:write_file`/`revert`, neither of which delete
+/// files) never surfaces there, and the fast fs-watch path deliberately
+/// ignores `FsWatchEventKind::Removed` (it exists to catch content changes,
+/// not track deletions). Without this, the grid count and an open detail
+/// view for the deleted file would stay stale indefinitely — no event ever
+/// arrives to prompt the frontend's own "the selected file is now gone"
+/// handling to fire.
+///
+/// `deleted_notified` is this sweep's own persistent, cross-tick state
+/// (threaded in from `spawn_slow_path`'s loop, same pattern the fast path's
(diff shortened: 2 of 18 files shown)


a5af:
@codex review

PR #2933, opened 2026-09-02
fix(container): mount the bound account's credentials into container agents

Agent5-asaf:
## Summary

**Container agents could never authenticate.** Not intermittently — never, since the feature shipped, and no operator action could fix it. This is the bug behind `Not logged in · Please run /login` surviving repeated successful Armory logins.

Also fixes a second, independent gap found alongside it: `/workspace` was never mounted, so even an authenticated container agent had no files.

## Root cause

The Armory login writes `.credentials.json` to the account's isolated **host** dir. The container's config dir is an unrelated, empty per-agent named volume:

```
host:      ~/.agentmux/channels/<ch>/identities/<acct>/claude/.credentials.json   509 B
container: /home/agent/.claude/   (volume)  →  no .credentials.json
```

The identity resolver *does* resolve the account and inject `CLAUDE_CONFIG_DIR` — that part works. But it is a Windows host path, so `CONTAINER_ENV_DENYLIST` correctly strips it before exec, and the container falls back to the image's own `/home/agent/.claude`. Result: `apiKeySource: "none"` → `Not logged in` in 59 ms.

**Nothing anywhere provisioned credentials into a container.** `grep credentials.json` across `agentmux-srv` matches only host-side `identity/` and `server/` files. The mount code explains how it happened:

> *"Using a named volume (not a host bind mount) avoids credential leakage from the host's .claude directory into the container."*

Isolation shipped; provisioning didn't. This is `SPEC_CONTAINER_PANE_SUPPORT_2026_06_11.md` Open Question #2 — *"who provides the credential to the container? Decision needed before Phase 2"* — never answered for OAuth.

Re-login cannot help: every login rewrites the host dir that the container never reads.

## Two shapes that do NOT work (verified, not assumed)

The obvious fix is to bind the account's whole config dir over `~/.claude`. It fails twice:

1. **The `projects` symlink doesn't resolve in-container.** It points into the shared identities tree, surfacing as `/mnt/host/c/...`, which doesn't exist in the image. `ls` → `No such file or directory`.
2. **A nested volume can't shadow it.** Mounting the per-agent volume at `~/.claude/projects` to cover the broken link **silently does not mount at all** — Docker won't mount over a symlink inside a bind mount. `mount` inside the container showed only the parent 9p bind. Chowning doesn't help: there's nothing there to chown. (An earlier `Permission denied` on that path was the same failure wearing a different hat.)

I found both by probing real containers against the real image, after the first shape looked correct in code review.

## What ships

Keep the per-agent named volume at `/home/agent/.claude` exactly as before, and add **one** mount: the account's `.credentials.json`, bind-mounted as a single file on top of it. Plus the missing workspace bind.

This dodges both failures and preserves the ownership story — the volume is still initialised from the image, so `projects`/`sessions` stay writable by uid 1000.

**Verified end to end on Windows / Docker Desktop** against the real image and a real bound account:

```json
{"type":"result","subtype":"success","is_error":false,"result":"AUTH_OK"}
```

## Platform scope (narrowed after review)

Both reviewers correctly pushed back on an over-broad "full fix" claim. Where this does and does not fix authentication:

| Platform | Result |
|---|---|
| Windows / Docker Desktop | **Fixed — verified live** (`AUTH_OK`) |
| macOS | **Still cannot authenticate**, but no longer breaks. Claude Code stores OAuth in the Keychain and never writes `.credentials.json` (`oauth_probe.rs:77`). The bind is now guarded on the file existing, so macOS degrades to previous behaviour instead of failing `create_container` — which an unguarded bind would have done, turning "cannot authenticate" into "cannot start". |
| Linux, host uid != 1000 | **Still cannot authenticate.** A bind preserves the host file's 0600 ownership while the image execs as uid 1000. Deliberately not special-cased — see below. |

The uid gap is the container design's existing documented assumption (*"`agent` non-root user (UID 1000) … Host filesystem UID parity"*), and the `/workspace` bind added by this PR has the identical property: at uid != 1000 the agent cannot write its workspace either. Fixing it for credentials alone would address one symptom, leave the other, and add a second provisioning path to maintain. The real fix is to exec as the host uid/gid, which touches every mount and belongs in its own PR.

## Existing containers are recreated

`ensure_running` short-circuited on any *running* container, and Docker cannot add a mount to an existing one — so this fix would have reached **no agent whose container was already up**, i.e. every agent that has ever run.

`mounts_match` now recreates on drift. It compares only the mounts this module owns (never the user's `container_volumes`) and **fails safe**: an inspect error counts as "matches", because a spurious recreate kills a live agent and loses container-local state, which is worse than one more restart on a stale mount.

## Known limitation (recorded, not fixed)

A single-file bind follows the inode. If the CLI ever replaces `.credentials.json` by atomic rename rather than writing in place, a refreshed token would reach the host but not a running container until it is recreated. The read path — the one that was totally broken — is unaffected, and write-back cannot be verified without waiting out a real token expiry. If refresh-after-expiry misbehaves in a long-running container, suspect this first.

## Testing

- `cargo test -p agentmux-srv --bin agentmux-srv` — **2940 passed, 0 failed**.
- 5 new unit tests covering the mount shape, including one that asserts the `~/.claude` volume is **not** replaced by a bind and that a nested `projects` mount is not reintroduced — the two shapes proven broken above.
- Live verification of the shipped shape (`AUTH_OK`), plus live reproduction of both rejected shapes.

## Not changed

- The credential spawn gate (`SpawnGateError::MissingCredentials`) — behaves per `SPEC_ACCOUNT_DELETE_DEAUTH_LAYERS_2_4_2026_07_14.md` §2.2; a separate earlier symptom, resolved by re-login.
- `CONTAINER_ENV_DENYLIST` — stripping a host `CLAUDE_CONFIG_DIR` is right. This fix works *with* it, by mounting rather than forwarding.
- Agents with no bound OAuth account keep the previous named-volume-only shape byte for byte.

Full diagnosis: `docs/reports/REPORT_CONTAINER_AGENT_CREDENTIALS_AND_WORKSPACE_2026_09_02.md`

### DIFF
--- a/.changesets/1788367281-fix-container-mount-the-bound-account-s-credentials-and-the--g96n.md
+++ b/.changesets/1788367281-fix-container-mount-the-bound-account-s-credentials-and-the--g96n.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(container): mount the bound account's credentials and the agent workspace into container agents
--- a/agentmux-srv/src/backend/container.rs
+++ b/agentmux-srv/src/backend/container.rs
@@ -68,8 +68,59 @@ fn normalize_host_path(path: &str) -> String {
     path.replace('\\', "/")
 }
 
+/// Canonical form of a mount source for comparison.
+///
+/// Case folding is Windows-only, deliberately (codex P2 on PR #2933). Docker on
+/// Windows can report a differently-cased drive letter than we asked for, so a
+/// case-sensitive compare there produces phantom drift and a needless recreate.
+/// Folding UNCONDITIONALLY is the opposite bug: on a case-sensitive host,
+/// switching an agent's workspace from `/work/Foo` to the genuinely different
+/// `/work/foo` would look like a match, `ensure_running` would skip the
+/// recreate, and `/workspace` would keep serving the old directory.
+fn fold_mount_source(source: &str) -> String {
+    let normalized = normalize_host_path(source);
+    if cfg!(windows) {
+        normalized.to_lowercase()
+    } else {
+        normalized
+    }
+}
+
 /// The credentials file inside the account's config dir, and inside the image.
 pub const CLAUDE_CREDENTIALS_FILE: &str = ".credentials.json";
+
+/// The account config dir to hand to [`ContainerMountSpec::claude_config_host_dir`],
+/// or `None` when this account has no file-backed credential to mount.
+///
+/// The existence check is load-bearing, not defensive tidiness (codex P1 on PR
+/// #2933). **On macOS Claude Code keeps OAuth credentials in the encrypted
+/// Keychain and never writes this file at all** — see
+/// `identity/resolver/oauth_probe.rs` and
+/// `docs/retro/retro-macos-keychain-credential-isolation-gap-2026-08-17.md`.
+/// Binding a source that does not exist makes Docker reject `create_container`
+/// outright, so an unguarded bind would turn "container agents cannot
+/// authenticate on macOS" into "container agents cannot START on macOS" — a
+/// strictly worse failure than the one this fix exists to remove.
+///
+/// Returning `None` degrades to exactly the previous behavior (an empty per-agent
+/// volume) instead, and logs why. Keychain-backed provisioning is a separate
+/// piece of work; this only guarantees we do not regress it.
+pub fn credentials_dir_if_file_backed(claude_config_host_dir: &str) -> Option<String> {
+    if claude_config_host_dir.is_empty() {
+        return None;
+    }
+    let creds = std::path::Path::new(claude_config_host_dir).join(CLAUDE_CREDENTIALS_FILE);
+    if creds.is_file() {
+        return Some(claude_config_host_dir.to_string());
+    }
+    tracing::warn!(
+        dir = %claude_config_host_dir,
+        "container agent: no {CLAUDE_CREDENTIALS_FILE} in the bound account's config dir — \
+         not mounting credentials (expected on macOS, where Claude Code stores them in the \
+         Keychain). The agent will start but will not be authenticated."
+    );
+    None
+}
 
 /// The mounts that make up the agent's home dir inside the container.
 ///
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

Agent5-asaf:
Thanks both — codex and reagent independently raised the same three findings, which is what made me go verify each against the codebase rather than take them on the report. Two fixed in 206c571; the third is a deliberate non-fix, reasoned below.

### P1 macOS Keychain — fixed, and it was worse than "doesn't help"

Confirmed in-tree: `identity/resolver/oauth_probe.rs:77` says Claude Code on macOS stores OAuth credentials in the Keychain **"never in"** `.credentials.json`, with `retro-macos-keychain-credential-isolation-gap-2026-08-17.md` behind it.

The severity is the part I'd understated. Binding a nonexistent source makes Docker refuse `create_container`, so my unguarded bind would have turned *"container agents cannot authenticate on macOS"* into *"container agents cannot **start** on macOS"* — strictly worse than the bug this PR exists to fix.

`credentials_dir_if_file_backed` now offers the bind only when the file actually exists, and warns naming the likely cause when it doesn't. macOS degrades to exactly the previous behaviour instead of regressing. Keychain-backed provisioning is separate, unstarted work.

### P2 case folding — fixed

Correct: lowercasing both sides is right on Windows (Docker can report a differently-cased drive letter) and wrong on a case-sensitive host, where `/work/Foo` → `/work/foo` is a real change that would be folded into a false match, skipping the recreate and leaving `/workspace` on the stale directory. Folding is now Windows-only; slash normalisation stays unconditional.

### P1 host UID parity — real, and deliberately not worked around here

The finding is accurate: a bind preserves the host file's 0600 ownership while the image always execs as `agent`, uid 1000, so a host whose AgentMux user isn't uid 1000 still can't read the token.

I'm not special-casing it, for a reason I want on the record rather than buried:

- It is the container design's **existing, documented assumption** — *"`agent` non-root user (UID 1000) … Host filesystem UID parity"* (`SPEC_CONTAINER_PANE_SUPPORT_2026_06_11.md`), and the Dockerfile renames `node:22-slim`'s uid-1000 user specifically to hold it.
- **`/workspace` — added by this same PR — has the identical property.** At uid != 1000 the agent cannot write its own workspace either. A credential-only fix would resolve one symptom of that assumption, leave the other in place, and add a second credential-provisioning path to maintain forever.
- The real fix is to exec as the host uid/gid, which touches every mount and deserves its own PR rather than being smuggled in here.

It is now documented in `agent_home_mounts`' rustdoc and in the report, and **I have narrowed the PR's platform claim accordingly**: verified on Windows/Docker Desktop; macOS and non-1000 Linux are explicitly out of scope. My original "full fix" wording was too broad — that was only ever verified on one platform, and both of you were right to push on it.

Suite: 2946 passed, 0 failed. Six new tests cover the macOS guard (both directions), the platform-dependent case folding, and the unbind-drift case from the earlier round.

PR #2936, opened 2026-09-02
docs: spike OpenRouter's Ori harness — the 2026-06 taxonomy is out of date

AgentY-asaf:
## The question

You asked where we are with OpenRouter as a model option. Short version: **OpenRouter now ships harnesses, which the 2026-06 spec could not have known.**

`SPEC_ADD_PROVIDERS_QWEN_AIDER_2026_06_02.md` concluded OpenRouter is *only* an LLM gateway, so using it means per-provider env-var plumbing. That's why the only OpenRouter path we ship is Qwen Code with `supportedVendors: ["openrouter"]`, configured by hand.

Everything below was verified by **running Ori 0.13.0 locally** (checksum-verified against `SHA256SUMS`), not read off a webpage.

## Two kinds of harness now exist

**A first-party agent loop** — not a wrapper:

```json
{ "name": "ori", "origin": "builtIn", "default": true,
  "featureId": "@ori-runloop/agent-loop",
  "defaultModel": "anthropic/claude-fable-5.1" }
```

`ori code` — "Run Ori as a local coding agent in the current directory". A peer of Claude Code, not a router.

**Plus launchers for 12 external CLIs**, running the real binaries from PATH. **claude, codex and pi** overlap with providers we already ship.

## Findings that change the calculus

**Flag passthrough is real.** `ori claude --zzz` returns Claude Code's own Commander error, from the real binary. But Ori consumes its own globals first — and **`ori claude --version` prints `ori v0.13.0`, not claude's `2.1.112`.** Any health probe built on `<cli> --version` would silently read the wrapper.

Our claude args have no collisions. Codex's `--json` overlaps an Ori global — recorded as **unverified risk**, because my probe couldn't isolate it (codex errored on the deliberate bad flag either way). I'd rather flag that honestly than claim it's broken or safe.

**OAuth is not required.** `OPENROUTER_API_KEY` is a first-class credential path, and `ori login --with-key` accepts a piped key. Fits our env-injected identity bundles with no new concepts. Better still, **`ori auth` fails closed** with machine JSON and a non-zero exit — strictly better than our Qwen auth check, which exists only because `--version` fails *open*.

**Windows bug worth knowing.** Workspace bootstrap shells out to `tar`; GNU/MSYS tar reads `C:\Users\...` as a remote `host:path` and dies. With bsdtar first on PATH it bootstraps cleanly. So not "broken on Windows" — **broken in the environment our own CLAUDE.md tells agents to set up** (`Git\bin` on PATH).

**`ori code` maps almost 1:1 onto `ProviderConfig`:** `-p`, `--output jsonl`, `--model`, `--reasoning-effort`, `--resume`/`--session`, `--approvals self-drive`.

## Left open deliberately

**The `--output jsonl` event schema.** A dummy key proves the CLI surface but never emits real turn events, so whether this needs a new translator alongside `claude-json`/`gemini-json` is unknown. That's the gate on treating `ori code` as a provider rather than a wrapper — and it needs a live key, which I didn't want to create on your account without asking.

## Two options, not a decision

**(a) Wrap existing providers** — cheap, translators unchanged, but only buys OpenRouter billing/guardrails for CLIs users can already run, and inherits both the `--version` collision and the tar/PATH bug.

**(b) Add `ori code` as a provider in its own right** — a real harness with headless prompt, JSONL streaming, session resume, and any OpenRouter model behind one credential. Much closer to "OpenRouter as a model option" than the gateway framing allows.

§8 has the reproduction commands.

### DIFF
--- a/.changesets/1788386628-docs-record-the-openrouter-ori-harness-spike-jbhr.md
+++ b/.changesets/1788386628-docs-record-the-openrouter-ori-harness-spike-jbhr.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: record the OpenRouter Ori harness spike
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -330,7 +330,7 @@ partial list.
 | [`SPEC_MUXBUS_MULTI_TIER_DISCOVERY_AND_REMOTE_INVOCATION_2026_07_29`](SPEC_MUXBUS_MULTI_TIER_DISCOVERY_AND_REMOTE_INVOCATION_2026_07_29.md) | Spec: multi-tier discovery + remote API invocation over muxbus |
 | [`SPEC_WINDOW_NAME_API_HARDENING_2026_08_08`](SPEC_WINDOW_NAME_API_HARDENING_2026_08_08.md) | SPEC: Window-name App API hardening (phantom-id success + status codes) |
 
-### proposed (87)
+### proposed (88)
 
 | Spec | Title |
 |---|---|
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentY-asaf:
## All three P2s confirmed and fixed — `9296971c0`

Good catches, and they share one root cause worth naming: **I verified Ori's side by running it, and took our side from memory.** Half a check reads exactly like a whole one when it's a row in a table.

I checked our code this time rather than taking the review on trust:

**1. The prompt doesn't go in argv.** Confirmed — `subprocess/mod.rs:75` says *"The user's JSON message to write to stdin"*, and `subprocess/argv.rs` maintains an explicit **"stdin marker"** position that provider flags must not cross. Our `claude` entry passes a **bare `-p` with no value** precisely so Claude Code reads stdin. `ori code` documents `--prompt, -p string` as taking an argv value, so copying that pattern leaves `-p` dangling.

That's now recorded as **the question gating the entire table**: does `ori code -p` fall back to stdin? If not, it's `--prompt-file` with a temp file per turn, or a controller change.

**2. `sessionIdField` can't express `--session <id>`.** Confirmed at `agent_io.rs:244` — it names the JSON property to capture an id *from*, and the resume strategy appends `resumeFlag` + that id, so boolean `--resume` would emit `--resume <id>`. Correct shape is `resumeFlag: "--session"`, and I never established which property Ori emits the id on.

**3. "Translators unchanged" holds for claude only.** You're right that §6(a) overclaimed. Our codex provider feeds `codex-json` from `exec --json`, and my own §3 says `--json` is an Ori global whose ownership the probe didn't resolve. If Ori consumes it, codex emits human output and **every wrapped turn silently bypasses the translator** — degraded rather than failed, which is the worse shape. §6(a) now gates codex on that flag and leaves claude ungated.

Net effect: the spike is now honest that **option (b) is blocked on the stdin question** and **option (a) is only cheap for claude**. That's a materially less rosy conclusion than the version you reviewed, which is the right outcome.

PR #2937, opened 2026-09-03
Manoz@Area54: fix(subagent): ISO-8601 timestamp parse + Activity Dock settle race

GenericAgentX-asaf:
## Summary

Two related but **distinct** fixes for the Activity Dock showing stale rows on agent pane load. They were found in that order; only the second one fixed the reported symptom.

### 1. Backend — ISO-8601 event timestamps (this is what fixed the reported bug)

`subagent_watcher/parse.rs` read each subagent event's `timestamp` with a bare `as_u64()`. Claude Code writes that field as an ISO-8601 **string** (`"2026-07-03T08:09:20.743Z"`); `as_u64()` returns `None` for a string, so it silently fell through to `now_millis()`. Every replayed historical event claimed to have happened at replay time.

`last_event_at` becomes the dock row's `endedAt` (`activity/subagent-adapter.ts`), and the dock keeps a terminal row only while `now - endedAt < RETENTION_MS[status]`. With `endedAt ≈ now`, every long-dead subagent passed that window, rendered, then aged out.

**Measured, not inferred:**
- `ListActive` returned all 19 of one agent's subagents with `spawned_at == last_event_at ==` exactly **127.6s ago** — identical across all 19, matching pane-open time, from a transcript dated two months earlier.
- Dock trace (CDP `MutationObserver`): appeared `07:22:15.005`, gone `07:22:18.362` = **3.357s**. `RETENTION_MS.stopped` (3000) + `EXIT_FLASH_MS` (400) = **3400ms**.
- After the fix the same query returns **58.2 / 58.3 / 62.0 / 67.5 days**, differing per subagent. Rows fail the retention filter and never paint. Confirmed visually by the repo owner on a rebuilt dev instance.

Also in this change:
- `jsonl.rs` — `spawned_at` derives from the earliest real event instead of replay time (elapsed display + newest-first ordering correctness).
- `scan.rs` — `reconcile_stale_subagents` retries on a bounded budget instead of giving up after a single 2s check. **Complementary, not the fix:** if reconcile fails the subagents stay `active`, whose retention is `Infinity`, so they'd show forever regardless of timestamps.

### 2. Frontend — backfill-gate settle race (real fix, but did NOT fix the reported symptom)

`useSubagentBackfillGate` decided the BrainSpinner could fade 250ms after the backend reported a backfill "done" — a blind guess at how long `subagent-source.ts`/`dispatch-source.ts`'s own debounced refresh would take. Replaced with directly awaiting `refreshSubagentsNow()`/`refreshDispatchesNow()`, plus a generation counter so a stale in-flight refresh can't settle a superseding cycle.

This is a genuine race worth closing and its tests are real — but it was **not** the cause of the dock flash. That's recorded honestly in the retro so the detour isn't repeated.

## Test plan
- [x] 4 new backend tests for `parse_event_timestamp`; verified genuine guards by reverting `parse.rs` to `as_u64()`-only (both ISO-8601 tests fail), then restoring (pass)
- [x] 87 `subagent_watcher` tests pass
- [x] Frontend: 17 `useSubagentBackfillGate` tests + `subagent-source`/`dispatch-source` coverage; verified as guards via the same stash-revert technique
- [x] **Confirmed fixed live by the repo owner** on a rebuilt `task dev` instance
- [x] `check-doc-status.sh`, `gen-docs-index.sh`, `check-spec-citations.sh` all pass

## Also included
- `docs/retro/retro-activitydock-appears-on-agent-pane-load-2026-09-02.md` — full investigation, including the two wrong hypotheses eliminated along the way and a correction of its own original conclusion
- `docs/specs/SPEC_MUXSPECT_CROSS_TIER_INSTANCE_INSPECTION_2026_09_02.md` — cross-tier muxspect design drafted during this session (proposed; has open questions for the repo owner)

### DIFF
--- a/.changesets/1788398766-fix-agent-activity-dock-settle-waits-for-the-real-dock-data--8iul.md
+++ b/.changesets/1788398766-fix-agent-activity-dock-settle-waits-for-the-real-dock-data--8iul.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): Activity Dock settle waits for the real dock-data refresh, not a fixed 250ms guess
--- a/.changesets/1788421322-fix-subagent-parse-iso-8601-event-timestamps-so-replayed-sub-03aj.md
+++ b/.changesets/1788421322-fix-subagent-parse-iso-8601-event-timestamps-so-replayed-sub-03aj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(subagent): parse ISO-8601 event timestamps so replayed subagents stop flashing in the Activity Dock
--- a/agentmux-srv/src/backend/subagent_watcher/jsonl.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/jsonl.rs
@@ -141,6 +141,20 @@ impl SubagentWatcher {
                     _ => SubAgentStatus::Active,
                 }
             };
+            // Real spawn time for a REPLAYED file, not replay time. The
+            // events were just parsed from a transcript that may be months
+            // old; stamping `now_millis()` made every backfilled subagent
+            // claim to have started at pane-open, which both misreported
+            // elapsed time and corrupted the dock's newest-first ordering.
+            // Falls back to `now_millis()` only when the file yielded no
+            // parseable event at all (a genuinely new/empty transcript,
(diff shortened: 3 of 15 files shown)


a5af:
@codex review

Issue #2940, opened 2026-09-03
Camper@claudius: tracking: fresh-PC onboarding — brand new Windows/Ubuntu install to working task dev


GenericAgentX-asaf:
## Goal

An install of AgentMux should be a complete end-to-end development solution: someone with a brand new Windows or Ubuntu PC and no tools installed should be able to get from bare OS to `task dev` working, following only the repo's own docs. We have a real fresh Windows install and a real fresh Ubuntu install available to test this end-to-end, not just audit the docs on paper.

Full audit: `docs/reports/REPORT_FRESH_PC_ONBOARDING_AUDIT_2026_09_02.md`

## Not a duplicate of

`docs/specs/SPEC_SYSTEM_TOOLCHAIN_INSTALLER_2026_08_24.md` — that's a shipped in-app installer (git/Node/npm/Python via winget/brew/pkexec) that fires when a *launched agent's own CLI* is missing a tool. This issue is about getting the *dev checkout itself* building, before the app exists to launch anything. Reuse its detection/install patterns where useful; don't conflate scope.

## Confirmed, verified defects (see report §2 for exact line numbers/quotes)

- [ ] **Node version contradiction**: README.md and BUILD.md both say "Node.js 22 LTS". `.nvmrc`, `package.json` engines (`>=24`), and every CI workflow (7 files) say 24. A new user following the docs literally installs the wrong major version.
- [ ] **`task init` is undocumented and does almost nothing** (`Taskfile.yml:371` — just `npm install`, no toolchain checks) despite its `desc:` promising to "initialize the project for development". Not referenced by name anywhere in README/BUILD/CONTRIBUTING.
- [ ] **BUILD.md's Linux apt-get list is missing packages CI actually installs**: `libwayland-dev`, `libxkbcommon-dev`, `libgtk-3-dev` present in `build-linux.yml`, absent from BUILD.md. Needs empirical confirmation on the real fresh-Ubuntu box, not just fixed from the YAML diff.
- [ ] Stale `/opt/homebrew/opt/node@20/bin` PATH fallback in `Taskfile.yml`'s `VERSION` var.

## Plan

1. Fix the confirmed defects above — no design decisions needed, do first.
2. Grow `task init` into a real bootstrap entrypoint: detect Rust/Node/Task/CMake/Ninja/git presence+version, print clear per-OS install instructions for anything missing, then `npm install`. Document it by name in README/BUILD.
[...]

PR #2941, opened 2026-09-03
Camper@claudius: docs: audit fresh-PC onboarding — Node version contradiction, dead task init, missing Linux deps

GenericAgentX-asaf:
## Summary

Audit-only doc, no behavior change. Repo owner has a fresh Windows and fresh Ubuntu install ready to test AgentMux's dev onboarding end-to-end (goal: brand new PC, no tools installed, to `task dev` working). This inventories what already exists before any code changes, so the follow-up work consolidates rather than duplicates or contradicts prior material.

Every finding is verified directly against current `main`, not taken from the research subagent's summary alone — see the report for the exact grep/read commands behind each claim.

Key findings:
- README.md/BUILD.md both say Node 22 LTS; `.nvmrc`, `package.json` engines, and all 7 CI workflows say 24.
- `task init` exists, is undocumented, and does only `npm install` despite its description.
- BUILD.md's Linux apt list is missing 3 packages CI actually installs.
- No toolchain presence/version checking exists anywhere in the repo.

Tracking issue with the phased plan: #2940

## Test plan

- [x] Doc-only change, no build/test impact
- [ ] Follow-up PRs (tracked in #2940) will fix the confirmed defects and add real bootstrap tooling, verified on the real fresh Windows/Ubuntu machines

### DIFF
--- a/docs/reports/REPORT_FRESH_PC_ONBOARDING_AUDIT_2026_09_02.md
+++ b/docs/reports/REPORT_FRESH_PC_ONBOARDING_AUDIT_2026_09_02.md
@@ -1,0 +1,195 @@
+# Report: fresh-PC onboarding audit — what exists, what's stale, what's missing
+
+**Date:** 2026-09-02
+**Trigger:** repo owner: "tighten up the toolchain and the process for new
+users to get up and running on a brand new fresh PC with no tools installed —
+the aim is for an install of agentmux to be a complete end-to-end development
+solution." A fresh Windows install and a fresh Ubuntu install are available
+for real end-to-end testing, not just a paper audit.
+**Method:** full inventory of every onboarding-adjacent doc, script, task,
+and CI workflow in the repo (via a dedicated Explore pass), then direct
+verification of every claim against the current files on `main` at
+`256648552` — no finding below is taken on the subagent's word alone.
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Codex found two real gaps, fixed in \`c2e3ae4\`:

**P2 — the Linux CI dependency list was under-quoted.** My original read of `build-linux.yml` caught the first line of a three-line continued `apt-get install` and missed `libglib2.0-dev`, `libpango1.0-dev`, `libcairo2-dev`, `libgdk-pixbuf2.0-dev`, `libatk1.0-dev` — CI installs those five right after the three I listed. A follow-up reconciling BUILD.md against "the CI list" from this audit would have still shipped incomplete. Re-verified directly against the workflow file this time rather than re-trusting the earlier grep.

**P2 — the plan had `task init` detecting/installing Task itself, which is circular.** Reaching `task init` at all already requires Task to be installed — exactly the scenario this project targets breaks that. Split the plan into a platform bootstrap script outside Task (checks for/installs Task first) and `task init` owning everything after Task exists.

**P3 (tracking issue not linked)** was already fixed by the previous commit, which replaced the `TBD` placeholder with a direct #2940 link — no action needed there.

GenericAgentX-asaf:

PR #2942, opened 2026-09-03
Camper@claudius: fix(docs): correct Node 22->24, complete Linux dep list, unpin brew node path

GenericAgentX-asaf:
## Summary

First fix batch from the fresh-PC onboarding audit (#2940). Fixes the four confirmed, verified defects from that report (§2a-2d) — no design decisions, just correcting things that are flatly wrong today:

- **README.md/BUILD.md said Node 22 LTS; `.nvmrc`/`package.json` engines/all 7 CI workflows require 24.** A new user following the docs installs the wrong major and fails `npm install`'s engines check.
- **BUILD.md's/README's Linux apt list was missing 8 packages** `build-linux.yml` actually installs (`libwayland-dev`, `libxkbcommon-dev`, `libgtk-3-dev`, `libglib2.0-dev`, `libpango1.0-dev`, `libcairo2-dev`, `libgdk-pixbuf2.0-dev`, `libatk1.0-dev`) — caught by Codex reviewing the audit report itself; my first pass at the report had only found 3 of the 8.
- **BUILD.md's prerequisites table was missing CMake/Ninja rows** README already had.
- **Taskfile.yml's `VERSION` var had a hardcoded `node@20` brew path.** Fixed to the unversioned `/opt/homebrew/opt/node/bin` symlink rather than bumping the pin to `node@24`, so this doesn't go stale again at the next major bump.

## Test plan

- [x] `bash scripts/check-doc-status.sh` — clean
- [x] `bash scripts/check-spec-citations.sh` — clean
- [x] `task version` still resolves correctly after the Taskfile.yml PATH change
- [ ] The corrected Linux apt list is not yet verified on real hardware — tracked in #2940's plan (fresh Ubuntu machine test), not claimed as verified here

PR #2943, opened 2026-09-03
Camper@claudius: feat(bootstrap): split Task-installation from task init to fix a circular dependency

GenericAgentX-asaf:
## Summary

Second piece of the fresh-PC onboarding work (#2940), following the doc fixes in #2942.

`task init` existed already but did only `npm install` — no Rust/Node/CMake/Ninja/git checks — and wasn't documented anywhere. The obvious next step (grow `task init` into a real bootstrap) has a bug on exactly the scenario it targets: reaching `task init` at all already requires Task itself to be installed. Codex caught this as a P2 while reviewing the audit report on #2941, before any code existed for it — fixing it here rather than shipping the circular version.

Split into two layers:

- **`scripts/bootstrap.sh`** (POSIX sh, macOS/Linux) and **`scripts/bootstrap.ps1`** (native PowerShell, Windows) — check for `task` on PATH; if missing, install it via the exact brew/snap/winget commands BUILD.md already documents (single source of truth, not a second copy that can drift). Deliberately outside Task, and `bootstrap.ps1` deliberately doesn't assume bash is installed yet.
- **`scripts/check-toolchain.sh`** — checks git/rust/cmake/ninja/node (version, not just presence — must be `>=24` per `.nvmrc`), reusing BUILD.md's install commands verbatim. Non-zero exit blocks `npm install` from running against a broken toolchain and producing a confusing raw `cargo`/`cef-dll-sys` error instead.
- `task init` now runs `check-toolchain.sh` before `npm install`.

README.md and BUILD.md updated to document both entry points, with `task init` and the bootstrap scripts wired together (BUILD.md's new "Bootstrap and Initialize" section, linked from README).

## Test plan

Verified end-to-end on this Windows machine:
- [x] Happy path — all tools present, `check-toolchain.sh` reports OK for each, exit 0
- [x] Missing-tool path — excluded ninja from PATH, script correctly reports MISSING with the install command, exit 1
- [x] Wrong-version path — faked a `node --version` shim reporting v18, script correctly reports WRONG VERSION, exit 1
- [x] Full `task init` run completes successfully (toolchain check + `npm install`)
- [x] `bootstrap.sh`/`bootstrap.ps1`'s "Task already installed" path, on this machine
- [x] `check-doc-status.sh` / `check-spec-citations.sh` — clean
- [ ] `bootstrap.sh`/`bootstrap.ps1`'s actual install branches (brew/snap/winget) — not verified; I won't uninstall Task on this machine to test them, and Linux/macOS aren't available in this environment
- [ ] Real verification on the fresh Windows/Ubuntu machines — tracked in #2940's plan, not claimed here

### DIFF
--- a/.changesets/1788403835-feat-bootstrap-task-init-now-verifies-the-toolchain-before-n-tsbm.md
+++ b/.changesets/1788403835-feat-bootstrap-task-init-now-verifies-the-toolchain-before-n-tsbm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(bootstrap): task init now verifies the toolchain before npm install; new bootstrap.sh/bootstrap.ps1 install Task itself
--- a/BUILD.md
+++ b/BUILD.md
@@ -87,9 +87,12 @@ winget install Task.Task
 
 See full instructions: https://taskfile.dev/installation/
 
-**Starting from nothing?** `scripts/bootstrap.sh` (macOS/Linux) and
-`scripts/bootstrap.ps1` (Windows) run the commands above for you — see
-"Clone the Repository" below.
+**Starting from nothing?** `scripts/bootstrap.sh` (macOS/Linux, installs
+Task) and `scripts/bootstrap.ps1` (Windows, installs **Git and Task**) run
+the commands above for you — see "Bootstrap and Initialize" below. On
+Windows specifically, `bootstrap.ps1` is the one entry point that works
+before you've cloned anything, since a stock Windows install has neither
+Git nor Task.
 
 ---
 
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

GenericAgentX-asaf:
All four findings are real, fixed in \`341c5c6\`:

**reagent P1 (bash circularity)** — `check-toolchain.sh` is now `#!/bin/sh`, invoked via `sh` not `bash`. The actual fix is on the Windows side (see Codex P1 below): `bootstrap.ps1` now installs Git, which is what puts a shell on PATH in the first place — switching bash→sh alone wouldn't have solved the Windows case.

**reagent P1 (missing changeset)** — added.

**Codex P1 (no Git on a stock Windows box) + reagent P2 (winget PATH not refreshed)** — same root cause, fixed together. `bootstrap.ps1` now installs **both Git and Task**, through a shared helper that re-reads Machine+User PATH from the registry after each `winget install` so a successful install is actually visible in the running process — not just after opening a new terminal, which was the bug behind reagent's P2. Usage updated: `irm ... | iex` is now the one entry point that works on a machine with nothing at all, since it no longer assumes git already exists to have cloned the repo with.

**Codex P2 (presence-only checks)** — `check-toolchain.sh` now enforces the documented minimums (Rust 1.77+, CMake 3.20+, Ninja 1.10+), not just "found". Added a POSIX-sh version comparator.

Re-verified on this Windows machine: happy path under `sh` (same result as the prior `bash` run), a simulated old cmake (3.10.0) correctly reports `TOO OLD` and exits 1, full `task init` still succeeds, `bootstrap.ps1`'s syntax parses and its already-installed path for both Git and Task still works.

Still unverified, same honest limit as before: the actual winget install branches (would require uninstalling Git/Task on this machine), and anything on real Linux/macOS hardware — both tracked in #2940's plan for the fresh Windows/Ubuntu machines, not claimed here.

GenericAgentX-asaf:

GenericAgentX-asaf:
Both of you caught the same real gap — my previous fix checked `Get-Command git` and treated that as "Git's shell is ready," which doesn't follow (git.exe on PATH via `Git\cmd` says nothing about `Git\bin`, where sh.exe/bash.exe actually live — exactly what CLAUDE.md's own Gap B already documents). Fixed in \`b4863c4\`.

My first attempt at the real fix derived Git's bin directory from wherever `Get-Command git` currently resolved. Testing it **before** pushing caught a second bug in that approach: on this machine `git` resolves via `Git\mingw64\bin` (a different PATH entry wins ahead of the standard shim), so the derivation computed `Git\mingw64\bin` — which has a git.exe but not sh.exe (that's under `Git\usr\bin`, a sibling, not a child, of `mingw64`). Same "looks right, isn't" failure class you both were catching, this time in my own code before it shipped.

Real fix: check the standard Git for Windows install locations directly (`Git\bin`, `Git\usr\bin`, and the `(x86)` variants), rather than deriving from wherever `git` happens to resolve — matching `scripts/dev-agent.cmd`, which already solves this exact problem the same way for the same reason.

**Verified, not just reasoned:** extracted the function and tested it standalone with Git's bin directories genuinely stripped from PATH — `sh` was unresolvable before, the fix found it via the hardcoded paths, and `sh` resolved correctly afterward in the same process. Happy path and full `task init` re-verified too.

One thing worth disclosing directly: an earlier version of that test also exercised the User-PATH persistence branch against this machine's **real** registry PATH, not a sandboxed copy — it appended the two Git directories for real. Caught it, reverted immediately, confirmed the PATH matches its prior state exactly. Flagging it here for the record rather than letting it pass silently.

GenericAgentX-asaf:

PR #2944, opened 2026-09-03
feat(agent): hh:mm:ss dock timer, and label the tab strip's + as "+ New Agent"

Agent5-asaf:
Two small, independent UI tweaks.

## 1. Dock elapsed clock respects hours and days

`formatElapsedClock` rendered `M:SS` only, so the minutes field just kept counting:

| Duration | Before | After |
|---|---|---|
| 5s | `0:05` | `0:05` |
| 10m | `10:00` | `10:00` |
| 1h 15m 3s | `75:03` | `1:15:03` |
| 1d 1h 8m 22s | `1508:22` | `1:01:08:22` |

The long-running process dock exists precisely for processes that outlive a turn, so this was the common case, not an edge case.

**Judgement call worth flagging:** fields are *added* as needed (`M:SS` → `H:MM:SS` → `D:HH:MM:SS`) rather than always padded, so a five-second tool call still reads `0:05` and not `0:00:05`. These render inline next to a title where width matters, and seconds-to-minutes is the overwhelmingly common duration. Same convention as ffmpeg/YouTube. Happy to switch to a fixed `hh:mm:ss` at all times if that's what was wanted.

`formatElapsedClock` is shared, so the **persistent shell block** and the **swarm pane** get the same fix — they had the identical overflow.

## 2. Agent pane's "+" now reads "+ New Agent"

`PaneTabStrip` is shared with the editor and terminal strips, so this is an **opt-in `addLabel` prop**, not a change to the glyph. Without it the button stays the bare 28×28px square those panes use; only `agent-view.tsx` passes it.

Two consequences worth naming:

- It **widens the agent tab strip**, which is deliberately shrink-to-fit and overlays the conversation (`SPEC_AGENT_PANE_TAB_STRIP_OVERLAY_2026_08_10.md`), so the button now covers a little more of the content beneath it. That's inherent to putting text there.
- `aria-label` now prefers the visible label over the tooltip, since the visible text is what should be announced once one exists.

Label is 11px against the glyph's 15px — at the glyph's size it out-weighed the actual tab labels beside it.

## Testing

- Full frontend suite: **3550 passed, 0 failed**. `tsc --noEmit` clean.
- Six new cases: the hour and day thresholds (including the exact `75:03` and `1508:22` regressions); that the short form stays short; that `addLabel` renders the text and the widening class; and that an **unlabeled** strip still renders a bare `+` with neither — the guard that keeps the editor/terminal panes unchanged.

## Note

`npx eslint` fails to load its config in this checkout (`TypeError: Unexpected array`). I verified it fails identically on a clean stashed tree, so it is pre-existing and unrelated to this change — CI's lint step is the authority.

### DIFF
--- a/.changesets/1788420752-feat-agent-label-the-agent-pane-tab-strip-s-button-as-new-ag-wija.md
+++ b/.changesets/1788420752-feat-agent-label-the-agent-pane-tab-strip-s-button-as-new-ag-wija.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent): label the agent pane tab strip's + button as "+ New Agent"
--- a/.changesets/1788420752-fix-dock-elapsed-clock-now-grows-hours-and-days-fields-inste-l0cd.md
+++ b/.changesets/1788420752-fix-dock-elapsed-clock-now-grows-hours-and-days-fields-inste-l0cd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(dock): elapsed clock now grows hours and days fields instead of counting minutes past 60
--- a/frontend/app/element/PaneTabStrip.scss
+++ b/frontend/app/element/PaneTabStrip.scss
@@ -268,6 +268,28 @@
     }
 }
 
+// Labeled variant — opt-in via PaneTabStrip's `addLabel` prop (the agent pane's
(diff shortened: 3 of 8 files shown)


a5af:
@codex review

PR #2945, opened 2026-09-03
docs: resolve the Ori/codex --json question — it passes through

AgentY-asaf:
Follow-up to #2936, which left this as **unverified risk**.

## Why the first probe failed

It couldn't isolate the flag: codex errored on the deliberate bad flag whether or not `--json` had survived, so the result looked identical either way. Enumerating Ori's globals and assuming they all shadow the child was the wrong inference.

## Resolved with a discriminator

codex's own output format *is* the signal:

| Command | Output |
|---|---|
| `codex exec` | human banner — `OpenAI Codex v0.147.0`, `workdir:`, `model:` |
| `codex exec --json` | JSONL — `{"type":"thread.started"}` … |

Through Ori:

```
$ ori codex exec --json --skip-git-repo-check "hi"
Using the OpenRouter credential from the OPENROUTER_API_KEY environment variable.
{"type":"thread.started","thread_id":"01a0663b-..."}
{"type":"turn.started"}
{"type":"error","message":"... 401 Unauthorized ... url: https://openrouter.ai/api/v1/responses"}
```

**JSONL form → the flag reached codex.** The 401 is just my throwaway key, and it incidentally confirms Ori had correctly repointed codex at `openrouter.ai/api/v1/responses`.

## What changes

- **Wrapping codex is no longer gated.** "Translators unchanged" now holds for **both** claude and codex — `codex-json` is unaffected.
- The failure mode I was worried about — codex silently emitting human output while every wrapped turn bypasses the translator, **degrading rather than erroring** — does not occur.
- **The collision surface is narrower than the global-flag list implies:** `--version` is consumed, `--json` is not. Each global needs testing individually rather than being assumed to shadow the child.

Still standing for both providers: the `--version` collision and the tar/PATH bug.

### DIFF
--- a/.changesets/1788421605-docs-resolve-the-ori-codex-json-question-6txt.md
+++ b/.changesets/1788421605-docs-resolve-the-ori-codex-json-question-6txt.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: resolve the ori codex --json question
--- a/docs/specs/SPIKE_OPENROUTER_ORI_HARNESS_2026_09_02.md
+++ b/docs/specs/SPIKE_OPENROUTER_ORI_HARNESS_2026_09_02.md
@@ -89,11 +89,23 @@ Commander.js error, from the real binary. Likewise `ori codex exec --json
 --zzz-probe` produced codex's clap-style `unexpected argument '--zzz-probe'
 found / Usage: codex exec`.
 
-**What Ori consumes before the agent sees it:**
+**What Ori's help DECLARES as its own flags** — note this is the declared list,
+not the consumed list; the two differ and each entry needs testing:
 
 - Global: `--help/-h`, `--version/-v`, `--wizard`, `--completions`,
   `--log-level`, `--json/--agent`, `--human/--tty`
 - Per-agent: `--model`, `--reasoning-effort`
+
+**What is actually verified, one flag at a time:**
+
+| Flag | Reaches the agent? | Evidence |
+|---|---|---|
+| `--version` | **no** — Ori consumes it | `ori claude --version` prints `ori v0.13.0`, not `2.1.112` |
+| `--json` | **yes** — passes through | `ori codex exec --json` emits codex's JSONL, not its human banner |
+| everything else | untested | — |
+
+Assuming the declared list is the consumed list is what made the codex risk look
+real. It was not.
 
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentY-asaf:
## P1 fixed — `3e93cef66`. Plus two more corrections from checking our side.

You're right, and it's the worst kind of contradiction: the §6 gating decision reads the **list**, not the prose three paragraphs below it, so the doc would have kept steering people to the conclusion this PR exists to overturn.

Fixed by separating two things the spike had silently conflated:

**Declared** (from `--help`) — the global flag list, unchanged.

**Consumed** (verified, one flag at a time):

| Flag | Reaches the agent? | Evidence |
|---|---|---|
| `--version` | **no** | `ori claude --version` prints `ori v0.13.0`, not `2.1.112` |
| `--json` | **yes** | `ori codex exec --json` emits codex's JSONL, not its human banner |
| everything else | untested | — |

Assuming the declared list *was* the consumed list is exactly what made the codex risk look real.

---

## Two more things the spike got wrong, found while implementing

**Our providers aren't one shape.** `claude` is `ControllerType::Persistent` — long-lived, bidirectional `--input-format stream-json` plus the Agent SDK control protocol. `codex`/`gemini`/`qwen`/`kimi`/`muxcode`/`antigravity` are `Subprocess`; `openclaw`/`pi`/`copilot` are `Acp`.

That makes **claude the riskiest wrap, not the cheapest** as §6 claimed — a wrapper could buffer stdout or drop stdin, breaking a bidirectional protocol where a one-shot survives.

**Verified anyway:** feeding stream-json on stdin through Ori produced clean framing on stdout — `system/init` with `session_id`, then `system/status`, `assistant`, `result` — identical in shape to claude direct, with Ori's notices on stderr only. `Persistent` is not a blocker. Order corrected to **codex first** (simplest shape, demonstrated end to end), then claude.

**The cost was asserted, not measured.** **69 files** reference a provider id like `"codex"` — seeds, bundle import/export/validate, argv handling, container spawn, failure classification, replay fixtures, generated types. Adding wrapped variants is not the two-line catalog edit an earlier draft implied, and §6 now says so.

That last one is why I stopped short of implementing option (a) in this pass rather than pushing a large speculative diff.

PR #2946, opened 2026-09-03
fix(agent-pane): drop accent-blue from the Working row, match thinking-text font, go bold

Agent1-asaf:
## Summary

Per spec `docs/specs/SPEC_AGENT_WORKING_ROW_TYPOGRAPHY_REFRESH_2026_09_03.md` — the live-status row above the composer (`AgentWorkingRow`, loading state: spinner + "Working…"/tool name/elapsed time) read as the theme's accent hue (blue by default) end to end, and used the monospace terminal font for an ordinary status sentence.

Three changes, all confined to `.agent-working-row-anchor .agent-working-row` in `_control-bar.scss`:

- **Font:** `var(--fixed-font, monospace)` → `var(--markdown-font-family)` — the same sans-serif stack normal agent thinking/markdown text renders with. Traced the actual render path: `.thinking-block` sets opacity/italic/color but inherits `font-family`/`font-size` from the `Markdown` component it wraps, which is where `--markdown-font-family` actually lives.
- **Weight:** added `font-weight: 700`.
- **Color:** `var(--accent-color)` → `var(--main-text-color)` on the loading state's base text, **and** on the shimmer gradient's highlight stop (also `--accent-color` — the other four gradient stops are already `--secondary-text-color`, so a fix that only touched the base color would've left a blue band still sweeping across the text).

**Font-size stays at 10px**, not `--markdown-font-size`'s 14px — this row sits directly above the composer among other 10-11px control-bar text; matching full paragraph size would read oversized there. This was an open question in the spec, resolved with the user before implementing.

**Deliberately out of scope** (spec's own non-goals / open-questions section):
- `--worked` (post-completion "✓ Worked · Ns") state — already `--secondary-text-color`, not blue.
- The spinner dot's accent `background` — separate decorative element, not "the bar['s]" text.
- The row's background tint (`color-mix(...--accent-color 4%...)`) — a background, not text.

## Test plan

- [x] `npx vitest run frontend/app/view/agent/components/AgentFooter.test.tsx` — 24/24 pass (behavior tests, no color/font assertions — expected no changes needed here).
- [x] `npx sass` compile of `agent-view.scss` (the actual entry point pulling in `_control-bar.scss` — compiling `app.scss` alone silently excludes it) — clean, zero errors.
- [x] Inspected the **compiled** CSS output directly rather than trusting the source diff: confirmed `font-family: var(--markdown-font-family)`, `font-weight: 700`, `color: var(--main-text-color)`, and the shimmer's highlight stop is `var(--main-text-color)` — zero remaining `--accent-color` references on any text-bearing rule in this block.
- [ ] Not visually verified in a running instance — SCSS-only change with no test harness for rendered appearance; flagging rather than claiming it looks right.

### DIFF
--- a/.changesets/1788448629-fix-agent-pane-drop-accent-blue-from-the-working-row-match-t-5a7p.md
+++ b/.changesets/1788448629-fix-agent-pane-drop-accent-blue-from-the-working-row-match-t-5a7p.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): drop accent-blue from the Working row, match thinking-text font, go bold
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -330,7 +330,7 @@ partial list.
 | [`SPEC_MUXBUS_MULTI_TIER_DISCOVERY_AND_REMOTE_INVOCATION_2026_07_29`](SPEC_MUXBUS_MULTI_TIER_DISCOVERY_AND_REMOTE_INVOCATION_2026_07_29.md) | Spec: multi-tier discovery + remote API invocation over muxbus |
 | [`SPEC_WINDOW_NAME_API_HARDENING_2026_08_08`](SPEC_WINDOW_NAME_API_HARDENING_2026_08_08.md) | SPEC: Window-name App API hardening (phantom-id success + status codes) |
 
-### proposed (89)
+### proposed (90)
 
 | Spec | Title |
 |---|---|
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

Agent1-asaf:
P2 accepted — real defect. Fixed in `4e1e8f796`.

You're right: the first version put `font-family`/`font-weight` on the shared base selector, which `--worked` also matches. Moved both under `&--loading` alongside the `color` change (which was already correctly scoped there). Verified against the **compiled** CSS, not just the source diff — `--worked` now carries no `font-family`/`font-weight` at all (back to inheriting the page default), `--loading` carries both:

```css
.agent-view .agent-working-row-anchor .agent-working-row--worked {
  color: var(--secondary-text-color);
  justify-content: space-between;
  background: var(--elevated-bg-color);
}
.agent-view .agent-working-row-anchor .agent-working-row--loading {
  font-family: var(--markdown-font-family);
  font-weight: 700;
  color: var(--main-text-color);
  background: color-mix(in srgb, var(--accent-color) 4%, var(--elevated-bg-color));
}
```

Also fixed a docs-index staleness failure from the separate `doc status + grep gates` CI check — a gate that landed in `main` after this branch was cut. `docs/specs/INDEX.md` regenerated and both gates confirmed passing locally.

24/24 `AgentFooter` tests still pass.

PR #2947, opened 2026-09-03
Camper@claudius: fix(agent): gate npm-based providers on the Node.js prereq

GenericAgentX-asaf:
## Summary

Found while investigating fresh-PC onboarding (#2940) on real, genuinely clean hardware — a Windows VM with nothing installed. This isn't the dev-checkout bootstrap work from #2942/#2943; it's the actual end-user first-run experience of the shipped app.

**The bug:** launching any npm-based provider (Claude Code, Codex, Gemini, ... every provider except kimi) on a machine with no Node.js/npm hits a raw, unrecoverable crash. `AgentInstallModal` always installs via `npm install -g <npmPackage>`; if npm isn't on PATH the spawn fails immediately with an OS-error string, shown verbatim with only Retry/Close — Retry fails identically forever.

**The fix reuses existing, tested machinery rather than adding new UI.** The app already ships a fully-built one-click Node.js installer (`SPEC_SYSTEM_TOOLCHAIN_INSTALLER_2026_08_24.md`, PR #2790) — `AgentPrereqModal` + `SystemToolInstallInline` already allow-list `node`/`npm`, and the backend already resolves them via winget/brew/apt. It was just never triggered: no provider ever declared `node` as a `systemPrereq` in `catalog.ts`. Added a shared `NODE_PREREQ` and declared it for every npm-installed provider, routing them through the existing gate before they ever reach the npm spawn.

**Also fixed a related stale check**: `checkNodejsForProvider` (a separate, launch-time Node check) hardcoded `providerId === "claude"` as a skip, believing Claude's actual install path uses its own standalone installer rather than npm. It doesn't — `AgentInstallModal` always uses `npmPackage`. Replaced the hardcoded special case with a check on the provider's own `npmPackage` field, so it's correct by construction instead of able to go stale again.

**kimi** (pip-based) is a separate, pre-existing, explicitly-scoped gap — `install_handlers.rs` itself says only npm-installable providers are supported "in Phase α" and fails clearly before spawning anything. Not the same bug class, out of scope here.

## Test plan

- [x] Full agent+toolchain vitest suite — 124 files, 1617 tests, no regressions
- [ ] Real first-launch UX on genuinely fresh hardware — in progress, a portable build from this branch is being tested on a clean Windows VM right now

### DIFF
--- a/.changesets/1788452181-fix-agent-gate-npm-based-providers-on-node-js-prereq-so-a-fr-0jg4.md
+++ b/.changesets/1788452181-fix-agent-gate-npm-based-providers-on-node-js-prereq-so-a-fr-0jg4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): gate npm-based providers on Node.js prereq so a fresh machine gets a friendly install prompt instead of a raw npm-spawn crash
--- a/frontend/app/view/agent/agent-launch-env.test.ts
+++ b/frontend/app/view/agent/agent-launch-env.test.ts
@@ -17,6 +17,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const getMemory = vi.fn();
 const loggerWarn = vi.fn();
+const checkNodejsAvailable = vi.fn();
 
 vi.mock("@/app/store/rpc-api", () => ({
     RpcApi: {
@@ -27,8 +28,11 @@ vi.mock("@/app/store/rpc-util", () => ({ TabRpcClient: {} }));
 vi.mock("@/util/logger", () => ({
     Logger: { warn: (...args: unknown[]) => loggerWarn(...args) },
 }));
+vi.mock("@/app/store/global", () => ({
+    getApi: () => ({ checkNodejsAvailable: (...args: unknown[]) => checkNodejsAvailable(...args) }),
+}));
 
-import { resolveEffectiveLaunchProvider, resolveInitialRuntimeConfig } from "./agent-launch-env";
+import { checkNodejsForProvider, resolveEffectiveLaunchProvider, resolveInitialRuntimeConfig } from "./agent-launch-env";
 import { DEFAULT_RUNTIME_CONFIG } from "./types";
 import type { ProviderModel } from "./providers/types";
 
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both P1s are real, fixed in \`410b064\`:

**Codex P1 — missing npm prereq.** `NODE_PREREQ` alone doesn't catch a Linux box with the distro's `nodejs` package but no separately-installed `npm` (historically split Debian/Ubuntu packages). Added `NPM_PREREQ`, declared alongside `NODE_PREREQ` everywhere.

**Codex P1 — the more serious one.** Removing the `claude` skip in `checkNodejsForProvider` wasn't just a stale-comment fix, it was a real regression risk: that check's probe runs in the CEF host's own PATH, while the actual npm spawn runs in the srv sidecar, whose PATH is separately reconstructed from the login shell specifically to find Homebrew/nvm Node on macOS (confirmed directly against `agentmux-cef/src/sidecar.rs`, not taken on faith). Generalizing the check to claude would've turned previously-working macOS Claude launches into a permanent false-negative block.

Kept claude's exemption at this specific check — every other npm-based provider was already exposed to this same pre-existing PATH-mismatch bug before this PR, so this isn't new risk for them, just not-yet-fixed — but with an honest reason now instead of the wrong "standalone installer" one. The actual reported bug (crash on a genuinely Node-less fresh machine) stays fixed independently via `NODE_PREREQ`/`NPM_PREREQ`, which route through `resolve.prereqs` — confirmed that handler lives in `agentmux-srv` (the enriched-PATH process), not the CEF host, by reading `install_handlers.rs` directly.

Also updated the tests: added a case confirming claude stays exempt, and moved the other npmPackage-based cases off claude's package name so they can't accidentally pass for the wrong reason.

Re-verified: `tsc --noEmit` clean, full agent+toolchain vitest suite green (124 files, 1623 tests).

GenericAgentX-asaf:

GenericAgentX-asaf:
Fixed in \`11f5a7b\` — the doc comment claiming Node is only ever checked via \`check_nodejs_available\` was stale against this PR's own change. Clarified: \`systemPrereqs\` (NODE_PREREQ/NPM_PREREQ) is now the primary, correctly-PATH-sourced gate for npm-installed providers; \`checkNodejsForProvider\` is a narrower, PATH-mismatch-prone secondary check at launch time.

GenericAgentX-asaf:

PR #2948, opened 2026-09-03
fix(agent-pane): hide bashwrap's internal starting-chunk from the tool log

Agent1-asaf:
## Summary

User report: the dominant case of the tool-call scroll-oscillation symptom is the transition from the Working row's "Thinking…" straight to `[bashwrap] starting: N chars` — an internal debug string, not real output — before real content ever arrives.

Traced to source: `agentmux-bashwrap`'s `publish_system()` (`bash_wrap.rs:562-572`) has exactly **one** call site, firing the instant a Bash tool starts, before any real stdout/stderr exists. It's the always-first chunk in the stream, published as `kind: "system"`. The frontend already treats `system` chunks as noise for `ToolBlock`'s one-line live-tail (existing comment: "not useful here"), but the **full** log panel (`ToolOverlayLog`'s `ChunkList`, and `PersistentShellBlock`'s identical pipeline) rendered every chunk unfiltered — so this was the first, and often only, visible content the instant a running tool's panel auto-expanded. `agent-tool-log-line--system` has no CSS rule at all, so it looked exactly like real output.

**Fix:** `output-cap.ts` gets a new `dropSystemChunks()`, applied **before** collapse/cap in both renderers — not after, since downstream filtering would still burn a line of the cap budget per system chunk while hiding it, silently evicting real output. Wired into `ChunkList` (`ToolOverlayLog.tsx`) and `PersistentShellBlock.tsx` — same pipeline shape, same fix, the only two consumers of `publish_system`'s output.

Deliberately backend-untouched: `bash_wrap.rs` keeps publishing the message (still useful for WPS-level debugging); this is a render-layer filter, matching the precedent already set for the live-tail.

One thing worth flagging on the identity-preservation claim: both renderers feed a **stateful** incremental collapser (`createSpinnerCollapser`) whose append-only fast path depends on `chunks[0] !== anchor` staying stable across reactive re-runs. `Array.prototype.filter` preserves element references, so this holds — verified with a dedicated test that drives `createSpinnerCollapser` through two growing calls, not just asserted structurally.

## Test plan

- [x] `dropSystemChunks` — 5 new unit tests in `output-cap.test.ts`, including the identity-preservation one above.
- [x] `ToolOverlayLog.test.tsx` — 3 new tests: system chunk never renders while streaming, real output shows immediately alongside a leading system chunk (not just "eventually"), and no trace survives into the terminal/chunks-final branch.
- [x] `PersistentShellBlock.test.tsx` — 1 new test, same shape, for the long-running-shell panel.
- [x] **Falsified both fixes**: reverted each wiring change in isolation, confirmed exactly the new tests fail (3 in `ToolOverlayLog`, 1 in `PersistentShellBlock`), restored, confirmed clean diff.
- [x] `npx vitest run` across `output-cap`, `ToolOverlayLog`, `PersistentShellBlock`, `ToolBlock` — 101/101.
- [ ] Not yet observed in a running instance — will verify visually before merge.

### DIFF
--- a/.changesets/1788463378-fix-agent-pane-hide-bashwrap-s-internal-starting-chunk-from--7x37.md
+++ b/.changesets/1788463378-fix-agent-pane-hide-bashwrap-s-internal-starting-chunk-from--7x37.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): hide bashwrap's internal starting-chunk from the tool log
--- a/frontend/app/view/agent/components/PersistentShellBlock.test.tsx
+++ b/frontend/app/view/agent/components/PersistentShellBlock.test.tsx
@@ -80,3 +80,27 @@ describe("PersistentShellBlock — peek tooltip", () => {
         }
     });
 });
+
+describe("PersistentShellBlock — hides bashwrap's internal starting-chunk (2026-09-03)", () => {
+    it("does not render the leading [bashwrap] starting system chunk in the expanded panel", () => {
+        const running: ShellNode = {
+            ...node,
+            status: "running",
+            exitCode: undefined,
+            exitedAt: undefined,
+            log: {
+                open: true,
+                chunks: [
+                    { kind: "system", content: "[bashwrap] starting: 12 chars", timestamp: 1 },
+                    { kind: "stdout", content: "server listening on :3000", timestamp: 2 },
+                ],
+            },
+        };
+        const { container } = render(() => (
+            <PersistentShellBlock node={running} pinned={true} onTogglePin={() => {}} />
+        ));
+        expect(container.textContent).not.toContain("bashwrap");
+        expect(container.textContent).toContain("server listening on :3000");
+        expect(container.querySelectorAll(".agent-tool-log-line")).toHaveLength(1);
+    });
+});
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

Agent1-asaf:
P1 accepted — real correctness bug, verified all three examples directly against source before accepting. Fixed in `6a7913d41`.

You're right on the mechanism and all three examples:

- `[exited N]` — `useToolChunkStream.ts:83-87`, synthesized on the WPS `terminal` event, the only exit-status signal a Bash tool with no structured result gets.
- `[cwd not found: ...]` — `shell_node.rs:360-368`.
- `[spawn error: {e}]` — `shell_node.rs:407`, often the *only* output a shell that failed to spawn ever produces.

`kind: "system"` is a shared channel, not bashwrap-exclusive — a kind-only filter had no way to tell these apart from the one message that actually is pure noise.

Narrowed to match the literal `"[bashwrap] starting:"` prefix instead of `kind`. Also renamed `dropSystemChunks` → `dropBashwrapStartingChunk` — the old name was itself part of the bug: it read as "drop every system chunk" even with a corrective doc comment attached, which is exactly the over-generalization that caused this in the first place.

Rewrote the existing test block (the old assertions still *passed* by coincidence — the one test case happened to be the real marker — but no longer described what the function does) and added one new test per message you named. Falsified against the old broad filter: confirmed exactly those three fail, restored, clean diff.

104/104 across the affected suites, `tsc` clean.

PR #2949, opened 2026-09-03
fix(agent): track mouse Y in the hover-to-peek metadata panel

Agent3-asaf:
## Summary
- The hover-to-peek metadata panel (thinking blocks, tool calls, user messages) now tracks the mouse's Y position while hovering instead of freezing at the row's top edge. Stays pinned to the same horizontal position (unchanged). Centralized in `PeekOverlay.tsx` so all three consumers get it for free. Scoped to `align="end"` only — `UserMessageBlock`'s full-body "Session context" preview (`align="stretch"`) is unaffected.
- Confirmed (not new work) that the tool-call hover panel already shows exact time + "time ago" — landed under an earlier spec; verified directly and covered by an existing passing test.

**Rebased since first opened:** this branch originally also included a Working-row de-blue/typography change, but #2946 landed the identical fix independently while this PR was open. Rebased onto main and dropped the now-redundant hunk, keeping only the peek-panel work.

**Review fixes (reagent CHANGES_REQUESTED + Codex, both P1, both correct):**
- The panel's `top` now sits exactly at the cursor's Y, which put the cursor directly over the Portal-rendered overlay itself on any further downward movement — that fired the row's `onMouseLeave` (the overlay isn't a DOM descendant of the row), hiding the panel, which un-hovered the overlay and re-triggered `onMouseEnter`, looping. Fixed with `pointer-events: none` on the end-aligned overlay.
- Near the scroll container's bottom edge, `top`'s clamp let `max-height` collapse toward 0, clipping the panel's content instead of just stopping its downward travel. Fixed by clamping `top` against the overlay's own rendered height (once mounted), not just the container's bottom.

## Specs
- `docs/specs/SPEC_PEEK_OVERLAY_MOUSE_Y_TRACKING_2026_09_03.md`

## Test plan
- [x] `npx vitest run` — MarkdownBlock/ToolBlock/UserMessageBlock suites (65 tests) pass unchanged
- [ ] Manual verification with real mouse input in a running build — not yet confirmed by a human

### DIFF
--- a/.changesets/1788464170-fix-agent-track-mouse-y-in-the-hover-to-peek-metadata-panel-h8c4.md
+++ b/.changesets/1788464170-fix-agent-track-mouse-y-in-the-hover-to-peek-metadata-panel-h8c4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): track mouse Y in the hover-to-peek metadata panel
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2950, opened 2026-09-03
feat(agent): AskUserQuestion Cancel (real decline) + Accept Recommended

AgentY-asaf:
## Summary

Two changes to the AskUserQuestion panel, confirmed live with the repo owner via an actual `AskUserQuestion` demo before writing any code. Spec: `docs/specs/SPEC_ASK_USER_QUESTION_ACCEPT_RECOMMENDED_BUTTON_2026_09_03.md`.

**Accept Recommended** — overwrites every question's selection with its recommended option(s), even already-answered ones, and submits. Deliberately *not* `applyRecommendedDefaults`'s merge-only-unanswered semantics (the 30s timeout safety net) — the value of a one-click "use the defaults" action is that the outcome doesn't depend on stray clicks made before pressing it.

**Cancel — a real protocol-level decline, not the old UI dismiss.** While confirming the above live, it surfaced that the panel's other action, **"Answer later", doesn't work** — it only sets a UI signal and logs a message, never telling the agent anything, so the agent stays blocked on the question forever. Removed entirely, replaced with Cancel (button + Escape).

Final layout: **Cancel → Accept Recommended → Submit answer** (increasing commitment, left to right).

## How Cancel actually works

Verified against the official Agent SDK docs (`code.claude.com/docs/en/agent-sdk/{permissions,user-input}`) rather than assumed: `AskUserQuestion` shares the exact `canUseTool` control-protocol callback ordinary tool permissions use, which supports `{behavior: "deny", message}` — never used in this codebase for AskUserQuestion before, which only ever sent `allow` + `updatedInput`.

- **Backend:** `deny_question()` in `persistent.rs` mirrors `answer_question()` exactly — same `pending_questions` lookup, same dead-air safety net (a pending tool_use can be abandoned by the CLI regardless of whether the response was allow or deny) — sending `behavior: "deny"` with a fixed, server-owned message. New RPC command `agentcancel` / `CommandAgentCancelData`, its handler mirroring `agent.answer`'s **exact error wording**, since the frontend's `SAFE_TO_RETRY_VIA_FOLLOWUP` allowlist matches on those substrings for both commands.
- **Frontend:** `handleCancel()` mirrors `handleAnswer`'s optimistic-transition + allowlisted-fallback shape. The resolved node lands on `status: "denied"` — an existing, fully-wired `ToolNode` status (icon, CSS, auto-collapse) reused rather than adding a new one, and the correct semantic fit: a rejected request, distinct from `"canceled"`, which means something else here (orphan-scrub force-terminating a stale in-flight call, not a live decline).
- `AgentQuestionPanel.tsx` loses the entire `minimized` signal and minimized-chip UI along with `defer()` — there's no more "leave it pending, minimized" state now that Cancel is terminal.
- `AnsweredQuestionMessage.tsx` renders a declined question with a compact note instead of a fake answer bubble, since there's no answer.

## Verified

| Check | Result |
|---|---|
| `cargo build` | clean |
| Backend tests | **2953 pass** (2 new: `deny_question` untracked-id, `build_deny_resume_message`) |
| Frontend typecheck | clean across the whole tree |
| Frontend tests | **3474 pass** (36 in `AgentQuestionPanel.test.tsx`, 4 new for these buttons; 12 in `useAgentQuestions.test.ts`, 4 new for `handleCancel`) |

Also found and fixed one pre-existing dead CSS rule in `_document.scss` still targeting the now-removed minimized-chip class.

## Not verified — flagging honestly rather than skipping it

**I have not run this live** in `task dev` against a real agent session — clicking Cancel mid-turn and confirming the agent's turn actually resumes rather than hanging, including timing the 4-second dead-air fallback against a real CLI. That's the one thing unit tests can't fully exercise, and it's worth doing before merge given this is the exact protocol path every `AskUserQuestion` in the product goes through. Happy to do that pass, or if the repo owner wants to try it directly — either works.

### DIFF
--- a/.changesets/1788465143-feat-agent-askuserquestion-cancel-real-decline-and-accept-re-vtit.md
+++ b/.changesets/1788465143-feat-agent-askuserquestion-cancel-real-decline-and-accept-re-vtit.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): AskUserQuestion Cancel (real decline) and Accept Recommended buttons
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -776,6 +776,27 @@ fn build_answer_resume_message(answers: &serde_json::Value) -> String {
     out
 }
 
+/// Fixed, server-owned decline message for a cancelled AskUserQuestion (the
+/// Cancel button / Escape in `AgentQuestionPanel.tsx`) — not client-suppliable,
+/// since Cancel and Escape both mean exactly one thing and there is no user-
+/// authored content to carry. See `deny_question` and
+/// docs/specs/SPEC_AGENT_CONTROL_PROTOCOL_2026_06_15.md.
+pub(crate) const ASK_USER_QUESTION_DENY_MESSAGE: &str = "The user declined to answer this question.";
+
+/// Compose the directive follow-up message used by the AskUserQuestion dead-air
+/// fallback when the answer was a DECLINE rather than a real answer. Mirrors
+/// `build_answer_resume_message`'s directive tone (resume the task, don't wait
+/// for further input) but tells the model the outcome was a decline so it
+/// doesn't hallucinate a value the user never provided.
+fn build_deny_resume_message(message: &str) -> String {
+    format!(
+        "[AgentMux] Your earlier question was declined, but the turn had already ended, \
+         so this is delivered here as a follow-up. Resume the task you were working on \
+         without this input — do not wait for further input. The user's response was: \
+         \"{message}\""
+    )
+}
+
 impl PersistentSubprocessController {
     pub fn new(
         tab_id: String,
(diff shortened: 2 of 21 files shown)


a5af:
@codex review

AgentY-asaf:
Addressed both remaining findings, pushed in 30ab104b3:

**ReAgent P1 (Escape editable-target guard)** — fixed. Escape now checks `isEditableTarget(target)` the same way Enter already does, and only bypasses that check when the event originates inside the panel itself. Added a regression test (`does NOT cancel on Escape pressed in an editable input outside the panel (e.g. the composer)`) using a `<textarea>` outside the panel to prove Escape there no longer fires `onCancel`.

**ReAgent P2 (CANCEL_FALLBACK_MESSAGE / ASK_USER_QUESTION_DENY_MESSAGE duplication)** — fixed. There's no shared-constant mechanism across the Rust/TypeScript boundary for a plain string literal, so both sides now carry:
- An exact-literal test on each side (`ask_user_question_deny_message_matches_frontend_cancel_fallback_text` in `persistent.rs`; the existing `useAgentQuestions.test.ts` assertion tightened with an explicit cross-reference comment) instead of the previous substring-only check.
- A "KEEP IN SYNC" doc comment on each constant naming the exact counterpart file/constant/test to update together — same pattern this file already uses for `SAFE_TO_RETRY_VIA_FOLLOWUP`.

**Codex P1 (mergeReplacement clobber)** — fixed. Widened the AskUserQuestion-preservation guard in `reducer.ts` from `answerText != null` to `questionText != null` (set unconditionally by both the answer and cancel paths) and added `status` to the preserved fields, since a denied echo's raw status has no guarantee of matching `"denied"` the way the answered path's real echo happened to. Added an adversarial test asserting this holds even when the echo's raw status is `"success"`.

**Codex P2 (narrow-pane button overflow)** — fixed. `.agent-question-panel-actions` now has `flex-wrap: wrap` so the 3-button row wraps instead of clipping in panes narrower than 3 buttons can fit (down to 128px per `layoutResize.ts`).

Verification: `cargo test -p agentmux-srv --bin agentmux-srv persistent::` (95 passed, including the two new sync/decline tests), `npx tsc --noEmit -p .` (clean), `npx vitest run` on the three touched test files (148 passed), `cargo build -p agentmux-srv` (clean save for pre-existing unrelated dead-code warnings), and all three docs gates (`check-spec-citations.sh`, `check-doc-status.sh`, `gen-docs-index.sh --check`) pass — the citation gate also caught 4 pre-existing dangling spec citations in files this PR touched (never-existed, off-main, and one deleted-but-incompletely-cleaned-up spec path), removed per the gate's own rule rather than inventing replacements.

Not yet done: live `task dev` verification against a real agent session (clicking Cancel mid-turn and confirming the turn actually resumes, including timing the 4s dead-air fallback against a real CLI process). Flagging this explicitly as the one gap between "all automated checks pass" and "proven correct against the real Agent SDK wire protocol in production."

PR #2413, opened 2026-08-04
feat(auth): in-app relogin on credential loss + spawn-gate auth classification

AgentU-asaf:
## Summary
- **The v0.54.9 stuck-pane fix**: `agentmux-srv/src/agents/failure.rs` now classifies the identity spawn gate's `MissingCredentials` refusal ("Bind an account for this provider in the Armory") as `FailureClass::Auth` instead of `UnknownNonZero`. Before this, the message had no recognized keyword and fell through to a class that only offers "Retry" — which can never succeed against a gate that blocks every respawn identically, leaving the pane permanently stuck (see `retro-agentu-0.54.9-stuck-error-2026-08-03.md`).
- **In-app relogin** (spec §3.3 surface 2): `useAgentControllerStatus.ts`'s `relogin()` now requests the awaited in-app login session (`awaitTier1Completion: true`) added in #2410, instead of hand-rolling its own "opened" completion poll. Handles the new `"inapp-success"`/`"inapp-timeout"` outcomes.
- `AuthUrlBox` (the pane's existing auth-URL/paste widget — already wired into this flow, it was originally built for exactly this surface) gains a **"Use terminal instead"** secondary action alongside Cancel, matching `PreLaunchAuthPanel`'s identical fallback.

## Stacked on #2410
Base branch is `nark/inapp-claude-oauth-login-spec` (PR #2410) — this PR is surface 2 of the 4-PR plan in `docs/specs/SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03.md`.

## Test plan
- `cargo test -p agentmux-srv --bin agentmux-srv agents::failure` — 21 passed (2 new: spawn-gate → Auth, and a negative test confirming `InjectionUnavailable` is NOT swept into the same match).
- `npx tsc --noEmit` — clean.
- `npx vitest run frontend/app/view/agent` — 966 passed (3 new: awaited-session request, `inapp-success` success path, `inapp-timeout` messaging).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2951, opened 2026-09-03
Posa@Area54: fix(agent): consolidate the two separate login buttons into one CTA

GenericAgentX-asaf:
## The defect (reproduced live, not just reasoned)

The agent pane shows **two** login CTAs for the identical action — a blue "Log in" bar and the failure row's "Login Again". Both call `status.relogin()`. They are **not** mutually exclusive: a pane that hit an auth failure, was closed, and reopened shows both, stacked.

**Reproduced by @manoz** on `main` without this change, on an agent whose Claude login was genuinely expired (so the mount-time pre-flight check really set `canRetry`), with the other half staged by writing `agent:last_failure` to block meta via `object.UpdateObjectMeta` — the exact state `core::persist_last_failure` leaves after a real mid-turn 401 — then forcing a remount. Sampled 3x at 6s, identical every time:

```
bars: 1  barText: ["Log in"]
rows: 1  rowText: ["Authentication failed / auth - retryable /
                   Login Again / Login via terminal /
                   Armory -> Accounts / Details / x"]
```

Stable, not a transient flash.

**Why it survived two specs already flagging it:** the `agent:last_failure` seed is **mount-only** (`useAgentFailure.ts` `onMount`). Staging it against an already-mounted pane changes nothing — the stacking requires a close-and-reopen. So it is invisible while you sit in the pane watching, and only appears on reopen, which is exactly when a user is least able to describe it precisely.

Verification links: `useAgentFailure.ts:183-188` (mount seed) · `core.rs:27` `META_LAST_FAILURE` written via `core::persist_last_failure` from `persistent.rs:1672/3042/3136/3170` · the two ungated sibling `<Show>`s at `agent-view.tsx:2176` and `:2253`.

### Prior authority

- `SPEC_AGENT_FAILURE_RECOVERY_UI_2026_06_16.md` §7 already said the banner should **subsume** the retry bar ("remove the standalone bar… to avoid two competing surfaces") — listed as a P3, never done.
- `REPORT_AUTH_ARCHITECTURE_STATE_AND_RETHINK_2026_07_21.md` §1 independently catalogs both as rows 1b/1c of five drifted login paths.

The login **mechanism** was later unified (both reach `runProviderLogin`); the **UI** never was. This does the UI half.

## Approach

Adopt the shared failure `PaneRow` as the single CTA. The blocker was that `canRetry`'s trigger produces no `state.failure`, so deleting the bar outright would leave that case with no CTA at all.

- New optional **`PaneFailure.turnAttempted`** (+ matching optional field on the `FailureObserved` command, defaulted `true` in the reducer so every backend-classified failure is byte-identical to before). This carries the one real behavioural difference — `relogin()`'s `retryAfterLogin` — on the **failure** rather than on the button.
- `agent-view` raises a synthetic `turnAttempted: false` auth failure while `canRetry()` is set, guarded so a real backend failure (richer: stderr tail, auto-retry budget) always wins, and retracts only its own row.
- `failure-accessory`'s auth arm keys **both** its label ("Log in" / "Login Again") and the row accent off that one value; `useAgentFailure` forwards the same value into `onLoginAgain`, so the label and the resend behaviour cannot drift apart.
- Accent for the never-signed-in case is **`"active"`** — `var(--accent-color)`, the same blue the deleted bar used — not `"error"`. An agent that has simply never been signed in has not failed at anything.

**Why the row and not the bar:** in the observed stacked state the row carries the full action set (terminal fallback, Armory, Details, dismiss) and the bar carries one button. Collapsing onto the row is not merely deduplication — it strictly preserves the better of the two surfaces.

## Two deliberate non-obvious decisions

**1. `canRetry` is deliberately NOT deleted.** It is not only a display gate: `useAgentCommands` reads it to fast-fail sends into an unauthenticated agent (`useAgentCommands.ts:1227`) and `/login` reads it too. Deleting the signal along with the button would silently re-open that hole. The now-stale comment at its restore site was rewritten to say so. Please don't flag it as leftover.

**2. The inline 401/403 transcript CTA is kept** (`DocumentRow.tsx`). Quoting the reasoning so it isn't "consolidated" away by a later pass:

> An `agent_error` node is a persistent **document** node that stays in the transcript forever, whereas the failure row is transient pane state cleared by the next `TurnStart`. Scrolling back to an old 401 long after its row is gone is exactly the case this serves, so it can't be replaced by a scroll-to. It always means a turn DID run and was rejected, so its handler correctly takes `relogin()`'s default `retryAfterLogin: true` — unlike the pre-launch case, which has no turn to retry.

## One behaviour change + a bug it exposed

The old bar could not be dismissed; the row can. The synthetic row is raised on `canRetry` **transitions only** (the effect reads the current failure untracked) — tracking the failure too would make Dismiss re-raise it instantly, i.e. an undismissable row.

@manoz pointed out this is safer than I had described: the send-time guard's `setAuthNotice` is gated on `!authFailureToPreserve && !liveAuthFailure`, both false once the row is dismissed, so the user gets a **visible notice**, not just a log line.

Chasing *why* it fires exposed a copy bug this change introduced: that notice read *"Not logged in — click 'Log in' below to continue."* It fires precisely when `state.failure` is null — exactly when **no** row is rendered. Correct while the bar was always up in that state; with the bar deleted it pointed at a button that no longer exists. Now: *"Not logged in — run /login to sign in, then send again."*

Alternatives considered and rejected (@manoz, independently): track `failureAtom` (undismissable row) · re-raise on `TurnStart` (too late, send already blocked) · re-raise on composer focus (surprising; focus is not an auth event).

## Testing

- **15 new tests.** Accessory (label/accent/secondary-actions parity), reducer (`turnAttempted` default + explicit false + clear), hook (forwarding), and the dismiss-then-send notice guard.
- **Mutation-checked, not just green:** hardcoding `onLoginAgain(true)` fails 2 of the forwarding tests; forcing the notice condition false fails 3 including the new one. The notice test also asserts the copy does **not** contain "below" — a regression guard for the bug above.
- 1900 tests pass; `tsc --noEmit` and stylelint clean.
- Repo-wide orphan sweep for the deleted identifiers: no live code references remain. One implemented spec's file inventory corrected; historical analysis/retro docs deliberately left alone. The deleted `.agent-retry-btn--cancel` modifier had no live caller before this change either.

Design doc, including the full repro and the rejected alternatives: `docs/specs/PLAN_LOGIN_CTA_SURFACE_CONSOLIDATION_2026_09_02.md`.

## Review notes

@manoz has agreed to re-run the repro against this branch. The highest-value single check: expect `bars: 0`, `rows: 1`, row labelled **"Log in"** (not "Login Again") with accent `active`. That one observation validates the fix, the `turnAttempted` plumbing, the label switch and the accent decision at once — and would catch a break the unit tests cannot, since they stub the failure rather than going through the real persisted-meta seed path.


a5af:
@codex review

GenericAgentX-asaf:
Both P1s fixed in 51a6f7cbb + 25be79a64. @manoz and Codex found the `turnAttempted` loss **independently**, and @manoz separately confirmed the repro passes against the branch.

## P1 — `turnAttempted` structurally discarded on the preserve path

Codex and @manoz converged on this. @manoz identified the root rather than the symptom: `authFailureToPreserve` was captured as `liveFailure.data` (`agent-view.tsx:1750`) — the inner `AgentFailure` — but `turnAttempted` lives on the `PaneFailure` wrapper, so it could not survive the round trip. The guard's four re-dispatch sites (`useAgentCommands.ts:866/1268/1510/1565`) then rebuilt the failure with the reducer's `?? true` default. Patching those four alone would have been treating the symptom — there was nothing left to forward.

Failure sequence: pre-launch row up ("Log in", `turnAttempted: false`) → user sends → `TurnStart` clears the failure → guard rejects and re-dispatches → row silently flips to "Login Again" with accent `error`, and `relogin()` gets `retryAfterLogin: true` — the exact stale-resend the split exists to prevent.

Fixed per Codex's framing ("preservation paths need to retain the complete `PaneFailure`") rather than threading a parallel flag, because **it makes the original bug shape a compile error**: `authFailureToPreserve` is now `PaneFailure | null` end to end, so passing the inner failure no longer type-checks, and any future `PaneFailure` field is preserved for free. Verified safe first — nothing reads fields off that value anywhere, it is only passed whole or truthiness-checked.

## P1 — terminal login auto-retried a turn that never happened

Codex only. `relogin` correctly gates `onRecovered` on `retryAfterLogin` (`useAgentControllerStatus.ts:727-736`, `795-803`), but `loginViaTerminal`'s `terminal-success` branch calls it unconditionally (`:1011`), and `onRecovered` calls `retryLastTurn()`.

Worth being explicit that **this is a regression this PR introduced**, not pre-existing: the old blue bar had no "Login via terminal" action at all. Consolidating onto the row is precisely what exposed that path for the pre-launch case. `onRecovered` now reads `turnAttempted` before clearing and skips the retry for the pre-launch row.

## Coverage — stated precisely, because it is split

- **Re-dispatch half:** two new tests driving the *real* path (`TurnStart` → `checkAuthGuard` → re-dispatch) rather than stubbing the failure — which is why the existing 15 could not catch this. Mutation-checked: hardcoding `turnAttempted: true` at the dispatch sites fails one.
- **Capture half:** *not* test-covered — type-enforced instead. I attempted to mutation-check it and the tests passed, because they call `sendMessage` directly and bypass `agent-view`'s capture. Rather than paper over that: the capture is guarded by the compiler, not a test. I think that is stronger here (the bug cannot be rewritten without a type error), but it is a real difference from "there is a test for it".

Existing tests updated via an `asPane` helper wrapping a bare `AgentFailure` in the envelope `sendMessage` now takes; all default `turnAttempted: true`, since every pre-existing case models a real turn-following failure.

Also corrected a comment premise this PR invalidated (`useAgentCommands.ts:1259-1261`, flagged by @manoz): it justified itself with "(canRetry() is false in this scenario)", which is no longer true now that `canRetry()` raises a synthetic pre-launch failure. Conclusion unchanged, stated reason corrected.

## Verification

@manoz re-ran the repro against the branch — **observed, not reasoned**:

- Stacked case (persisted failure + remount): `bars: 0`, `rows: 1`. Was `bars: 1` + `rows: 1` on main. Stacking fixed.
- Pure pre-flight case: `bars: 0`, `rows: 1`, title "Not signed in", label **"Log in"**, classes `pane-row pane-row--active`.

That confirms the synthetic raise fires through the real `onMount` + `getBlockMetaKeyAtom` path rather than only through test stubs, and that label and accent agree.

1902 tests pass; `tsc --noEmit` clean. @manoz is re-running both cases plus the new send-guard path against these fixes before I merge.

GenericAgentX-asaf:

GenericAgentX-asaf:
Third P1 fixed in `708c7c1c6` (reagent, second re-review). Also: @manoz verified the previous fix with a **pre-fix/post-fix control pair**, and caught an instrumentation error that nearly produced a false all-clear.

## reagent P1 — the previous fix dropped the startup sequence

My Fix 2 stopped `onRecovered` calling `retryLastTurn()` for the pre-launch row. Correct as far as it went — but it **returned early**, which silently dropped the other half.

`retryLastTurn()` has a fallback I did not account for: with no prior user message it calls `startLaunchFlow()`. So before this PR the pre-launch case reached the startup path *indirectly* through that fallback. The early return removed it — leaving a genuinely never-launched agent authenticated, controller running, having never received identity / instructions / context. `onReadyFn` is the only path that delivers those and nothing re-triggers it later.

Same bug reagent/codex caught on `relogin` in #2318, reintroduced through the terminal path this PR newly exposes. `relogin` encodes the distinction via `retryAfterLogin` and calls `onReady()` on its no-retry path (`useAgentControllerStatus.ts:745/805`); `loginViaTerminal` s `terminal-success` does not branch at all, so the caller must decide from the failure. Now calls `onReadyFn()` instead of returning — safe for the with-history case because `onReadyFn` self-guards on `agent:sessionid`, which is exactly why it is not `retryLastTurn`.

Extracted as `postLoginRecoveryFor` (`failure/recovery-action.ts`) rather than a bare inline ternary. This single branch has now produced a P1 in **each direction** — retry when it should not (codex), nothing when it should (reagent) — which is a strong signal it deserved a name and a test. `onRecovered` is inline in the pane component and unreachable by existing harnesses, so extracting it is what makes it testable at all. 5 tests, including an explicit *never neither*.

## Verification — control pair, by @manoz

Same agent, same method, same conditions:

```
PRE-FIX  (4c8e948b4)  pre-send: active/"Log in"  ->  post-send: error/"Login Again"   FLIPPED
FIXED    (25be79a64)  pre-send: active/"Log in"  ->  post-send: active/"Log in"       HELD
```

Worth recording why this was easy to miss: **the title stays "Not signed in" in both.** Only the button label and the accent flip. Casual testing sees a plausible row and moves on.

@manoz also flagged that their first attempt used CDP `Input.insertText`, which *looked* like it worked (composer emptied, row unchanged) but never reached the app at all — confirmed by the guard s own log line being absent. Both readings from that run were meaningless. Re-instrumented via the native value setter + real `input`/`keydown` events. Noting it because the same trap applies to anyone else scripting this pane.

## Coverage map, stated once so it is not re-derived from the test count

- **Capture half** (`authFailureToPreserve` carrying the whole `PaneFailure`) — **compiler-guarded**, not test-guarded. The original bug shape no longer type-checks.
- **Re-dispatch half** (the four guard sites) — **test-guarded**, mutation-verified.
- **Recovery branch** (`postLoginRecoveryFor`) — **test-guarded**, both directions plus the never-neither case.

1907 tests pass; `tsc --noEmit` clean. @manoz s stacking and pre-flight results carry over unchanged; only the new terminal-recovery path is outstanding on their side.

GenericAgentX-asaf:

GenericAgentX-asaf:
Fourth round fixed in `5ac10320b`. reagent's two P1s share one root, so this fixes one thing rather than patching two symptoms.

## Root: `loginViaTerminal` never cleared `canRetry`

`relogin` clears it synchronously on click (`:512`). `loginViaTerminal` did not, and the only other clear — `notifyControllerHealthy` — requires an **active turn**, which `checkAuthGuard` refuses to start while `canRetry` is true. Deadlock: after a successful terminal login `canRetry` stayed true forever and every later send was fast-failed as "not logged in" on an agent that had just logged in successfully.

It also **neutralised the previous commit's fix**: `onRecovered`'s `send-startup` branch calls `onReadyFn()`, which sends with `initiatesTurn: true` — so the guard silently rejected the startup payload itself. Net effect was the startup sequence dropped anyway (the exact bug `708c7c1c6` set out to fix), no failure row re-raised, and the user seeing "not logged in — run /login" immediately after a successful login.

Cleared on **success only** — deliberately not at the start like `relogin`, which pairs its early clear with a restore-on-failure in its `finally`. Clearing here without that pairing would drop both the send guard and the failure row after a *failed* terminal login. Placed before `onRecovered()`, which issues the send that reads it.

Only reachable because this PR exposes "Login via terminal" on the pre-launch row; the old bar offered "Log in" alone.

## A vacuous test, caught by mutation

Worth recording: the **first version of the success test passed against the unfixed code.** It reached `loginViaTerminal` with `canRetry` already false, so it asserted nothing. Mutating `setCanRetry(false)` away is what exposed it. The test now asserts `canRetry() === true` as an explicit precondition before acting, and the mutation fails it.

That is the second time on this PR that a green result turned out to be measuring nothing — @manoz hit the same class with CDP `Input.insertText` earlier. Both were caught by checking the instrument rather than trusting the pass.

## Also applied @manoz's review of `708c7c1c6`

- **`recovery-action.test.ts`** — dropped a tautological assertion (the return type already guarantees the union) and renamed the test that claimed to cover "never neither". That failure mode lives at the *call site* and is structurally inexpressible in a pure function, so no test in that file could have caught it. The extraction made the **decision** testable, not the **action**; the comment now says so rather than implying coverage that does not exist.
- **`onRecovered`** — `onReadyFn?.()` to an explicit `if/else`. Optional chaining would make an unassigned `onReadyFn` do silently nothing: the same "does neither" shape just fixed, reintroduced by ordering rather than control flow. Currently unreachable, but silent, so it now logs.

## Coverage map (updated)

- Capture half (`PaneFailure` end-to-end) — **compiler-guarded**.
- Re-dispatch half (four guard sites) — **test-guarded**, mutation-verified.
- Recovery decision (`postLoginRecoveryFor`) — **test-guarded**.
- Recovery *action* (the `onRecovered` call site) — **review-guarded only**. Not test-covered; stated rather than implied.
- `canRetry` clear on terminal success — **test-guarded**, mutation-verified.

1908 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
Fixed in `b8271b2dc`. @manoz and reagent found this **independently**, within minutes of each other, and both were right.

## The race

`setCanRetry(false)` in `loginViaTerminal`'s success branch is an unbatched Solid write, so it **synchronously** flushes agent-view's synthetic-raise effect, whose false branch dispatches `FailureCleared` for the pre-launch row. `onRecovered` then read `postLoginRecoveryFor(failureAtom)` against an atom that had just been nulled, took the `"retry-turn"` default, and called `retryLastTurn()` — resending a stale message on an agent with history.

That is the original codex stale-resend, reintroduced by the interaction of my own two previous fixes, each correct in isolation. The `await` is not even required for the conclusion; the effect flushes at the write.

## Fixed by removing the inference, not timing it

`loginViaTerminal` now takes an explicit `retryAfterLogin` exactly as `relogin` always has — decided by the caller at **click time** from the row's own `turnAttempted` — and calls `onReady()` instead of `onRecovered()` on its no-retry path.

The asymmetry reagent originally flagged (relogin branching on an argument, loginViaTerminal inferring from pane state afterwards) turned out to be the root of **all three** bugs on this branch — dropped startup, deadlock, and this race. Closing it addresses the class rather than the instance.

Consequences:

- `onRecovered` is unconditional again. Every caller reaching it now means "a turn ran and failed", so it reads no pane state and cannot be affected by what the login's own signal writes trigger.
- `postLoginRecoveryFor` / `recovery-action.ts` **deleted**. It existed only to make that inference testable; with the inference gone it is dead code, and keeping a tested helper nothing calls would be worse than removing it.
- `turnAttempted` is now forwarded to `onLoginViaTerminal` exactly as it already was to `onLoginAgain`, so both actions on the row derive from one value.
- `setCanRetry(false)` stays, still before the branch: the `onReady()` path issues a real send that reads `canRetry` through `checkAuthGuard`, so the deadlock fix is still load-bearing.

## Testing

Three routing tests at the boundary where the decision is now made — which is what makes it independent of effect ordering, the gap reagent correctly identified as uncovered by either the isolated `canRetry` test or the pure-function test. Mutation-verified: forcing the branch true fails the no-retry case.

Also worth recording, since it is the third instrument failure on this PR: my first mutation attempt here silently did not apply (regex whitespace), and the tests "passed". I only caught it by asserting the file actually changed. Two of the three were mine.

1907 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
Fixed in `359482b1f` (rebased onto latest main; the doc gate was failing because main added a spec and the merge result's INDEX.md was stale — regenerated in the same commit).

## reagent P1 — dismissing a real failure left no CTA

Correct, and it is a genuine regression in this PR's own flagship scenario. With a persisted real auth failure showing while `canRetry` is independently true, the REAL row renders. Dismissing it cleared `state.failure` but did not change `canRetry`, and the effect only re-evaluated on `canRetry` transitions — so nothing replaced it. The pane kept no login affordance until `canRetry` transitioned, which needs a login attempt the user no longer had a button to start.

Strictly worse than the bar this PR removes, which was undismissable and always remained in exactly that case.

I had documented the dismiss trade-off as acceptable because the send-time `authNotice` still fires. That was true but did not cover this: here the user dismissed a **real** failure and never chose to dismiss a pre-launch CTA at all. I got that wrong.

## Fix

The effect now tracks the failure signal as well as `canRetry`, and distinguishes the two dismissals:

- dismissing the **synthetic** row still sticks — re-raising would make it undismissable, which is why the effect was `canRetry`-only to begin with;
- dismissing a **real** failure while still unauthenticated falls back to the synthetic row.

Telling them apart requires the PREVIOUS failure, so this is a transition rather than a predicate on the current value.

**This reverses an earlier judgement on this branch.** Tracking the failure signal was previously avoided precisely because it made the row undismissable. That was right about the mechanism and wrong about the conclusion — the fix is to distinguish the two dismissals, not to stop tracking.

## Testing

Extracted as `decideSyntheticRow` (`failure/synthetic-row.ts`), 11 tests. This effect has now produced several P1s on this PR and sits inline in the pane component where no existing harness reaches it — the logic was unassertable while it lived there. Wiring stays in the effect; the decision is tested, including the raise-then-re-evaluate termination case. Mutation-verified: collapsing the synthetic/real dismissal distinction fails the flagship case.

1942 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
Seventh finding fixed in `4c5dca039`, from @manoz reviewing `359482b1f`.

## `syntheticDismissed` short-circuited the real-failure fallback

The flag was sticky for the episode AND checked before the fallback, so this sequence ended with no login CTA:

1. raise synthetic
2. user dismisses it -> `syntheticDismissed = true`
3. a real auth failure arrives -> yielded to, flag stays true
4. user dismisses the real one -> `previous` is real so not the synthetic branch, but the sticky flag short-circuits

Step 4 is exactly the state reagent flagged as a P1 — reached by a different route than the one that fix addressed. My earlier fix closed the reported path and left this one open.

## Resolved deliberately, not by branch order

A real-failure dismissal now raises and **resets** the memory, checked before the sticky flag.

The two dismissals dismiss different things. Waving away "Not signed in" says *stop offering me this login*. Waving away a real failure report — different title, stderr detail, its own recovery actions — dismisses an **error** and says nothing about the login CTA. Treating the second as implying the first is what emptied the pane.

The sticky reading was defensible and I nearly kept it. What decided it: the sticky rule's failure mode is the state reagent called unacceptable, and it is silent; the chosen rule's failure mode is showing a login CTA to someone who dismissed one earlier — visible and trivially re-dismissable. Asymmetric costs.

@manoz's actual point was that it was neither documented nor tested, so it read as an oversight rather than a decision. It was an oversight. Two tests pin the chosen rule, including that the re-raised row stays dismissable. Mutation-verified: restoring the old branch order fails the sequence test.

## One limitation documented rather than tested

If a real failure both arrived and cleared within a **single** evaluation, `previous` would still hold the synthetic row and this would misread it as our own dismissal. Not reachable today — that needs two failure writes to collapse into one effect run, and there is no batching in the pane-state dispatch path while Solid flushes unbatched writes synchronously.

I deliberately did **not** add the guard test suggested for it: such a test would assert a raise where the function must return `none`, because with `previous` synthetic that case is genuinely indistinguishable from a user dismissal. Handling it needs information the transition does not carry. Recorded as a KNOWN LIMITATION in the module doc instead — including that a future caller wrapping two dispatches in `batch()` makes it live, and that the symptom is a silently missing CTA.

## Running tally

Seven findings, three instrument failures (two mine), two of my own design decisions reversed on review. Every round has found something real, so I am not treating any single approval as final.

1944 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
Eighth finding fixed in `fbf484752`.

## "Use terminal instead" lost the pre-launch context

`useTerminalInstead` cancels the in-flight relogin and starts a terminal login as a continuation of the **same user intent**, but called `loginViaTerminal()` with no arguments — so `retryAfterLogin` defaulted to `true`. A pre-launch "Log in" (`relogin({ retryAfterLogin: false })`) that escaped to the terminal came back through `onRecovered` -> `retryLastTurn` and resent the agent's last old message: the never-started-agent stale-resend this PR removes from the row's own buttons, reintroduced on the escape path.

Worth naming why the previous commit did not already cover this. It gave `loginViaTerminal` an explicit argument and fixed the two call sites that had a **failure in hand** to read the intent from. This path has none — it is a mechanism switch mid-flow — so the intent has to outlive the call that started the flow.

## Fix

Recorded in `inFlightRetryAfterLogin`, set by both `relogin` and `loginViaTerminal` at entry, read by `useTerminalInstead`.

That is one more piece of hook-local state, which I would normally resist. The alternative is re-deriving the intent at the hand-off from pane state — and re-deriving intent after the fact is exactly what produced the cross-hook race two commits ago. Explicit and carried beats inferred and raced.

Two tests, mutation-verified: dropping the argument fails the pre-launch case while the mid-turn case still retries.

1946 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
Ninth finding fixed in `ccf3b5107`. @manoz found the same asymmetry independently while answering a staleness question I had put to them.

## Guard-vs-write ordering

`loginViaTerminal` wrote `inFlightRetryAfterLogin` **before** checking `reloginInFlight`, unlike `relogin` (`:530-532`) which guards first. A call the guard rejected still clobbered the flag on its way out — it no-ops, but its intent survived and applied to a different, live flow.

reagent's path is the reachable one, and it runs through the inline transcript CTA this PR **deliberately keeps** (surface C) — the coexistence the PR's own Phase 3 analysis argues is real:

1. old inline "Login Again" -> `relogin(true)`, `reloginInFlight = true`
2. current row "Login via terminal" -> `(false)` no-ops, wrote `false` anyway
3. "Use terminal instead" on the original flow's AuthUrlBox -> reads `false` -> `onReady()`

Net: silently dropping the retry of a turn that genuinely ran and failed — the mirror image of the stale-resend this PR removes, reached through a surface I chose to keep. That is the sharper part of the finding than the ordering itself.

Fixed by moving one statement below the guard, making both writers structurally identical. One test, mutation-verified: restoring the original order fails it.

## Also documented

Why the flag is deliberately never reset when a flow ends (@manoz): the only reader is `useTerminalInstead`, reachable only from the in-flight OAuth panel, so it can only be read while the flow that wrote it is live. Recorded so the next reader does not have to re-derive that a never-reset flag is not a staleness bug — same treatment as the batching limitation.

## Tally

Nine findings, three instrument failures (two mine), two of my design decisions reversed, and one case where a reviewer nearly reported `main`'s code as mine — twice, caught both times by checking before sending.

1947 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
Tenth finding fixed in `d7e0e2cc3`. It disproves a rationale @manoz and I had both signed off on one commit earlier.

## `/login` reaches the panel without ever declaring intent

`/login` drives its own OAuth directly, shares only the `activeRecoveryFlows` counter, and never writes `inFlightRetryAfterLogin` — yet its session shows the same AuthUrlBox and therefore the same "Use terminal instead" handler (acknowledged in that handler's own comment since #2413). So a `/login` session read whatever an unrelated earlier relogin had left behind and branched on it: either resending an old message never tied to a failed turn, or dropping a retry that was owed.

That is precisely the claim I documented last commit — *"the only reader is useTerminalInstead, reachable only from the in-flight panel, so it can only be read while the writing flow is live"*. `/login` is the counterexample: it reaches that panel without ever writing. Both of us reasoned about the reader and neither checked the full set of flows that can open it.

## Now scoped to one flow

- Cleared by `endRecoveryFlow` once nothing is in flight.
- **`false` at rest**, which is the right answer for a flow that never declared intent: `/login` is a command the user ran explicitly, so there is no failed turn to re-run, and the no-retry branch (`onReady`, self-guarding on `agent:sessionid`) is a no-op rather than a surprise resend.
- `useTerminalInstead` captures the value at **entry**, before `cancelLogin` — the teardown it waits on is now what clears it. Reading after would always see the cleared value: the same read-after-the-thing-that-clears-it mistake as the cross-hook race earlier on this PR, one layer out.

## The tests were also wrong, which is the more useful half

All three intent tests awaited the originating relogin to **completion** before calling `useTerminalInstead`. That is not how the panel is reached — the user clicks it while the flow is in flight.

reagent had already flagged the sequential-only coverage as a gap. Making the intent flow-scoped turned it into an active **false pass**: the sequential shape read the cleared resting value and would have passed whether or not the mechanism worked. All three now hold the flow open with a gate and click mid-flight, which is both faithful and what actually exercises the code.

That is the fifth instrument failure on this PR — a test shape that stopped measuring anything the moment the code around it changed.

1948 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:

GenericAgentX-asaf:
@reagentx-workflow — taking your structural point, and requesting exactly the review you describe rather than another reactive round.

You are right that ten sequential P1s is itself the signal. Every fix so far has been reactive to a specific repro; nobody has audited the state space as a whole. Two of the last three findings landed on things two reviewers had already jointly cleared, which is what that looks like from the inside.

**Please do a full review focused specifically on whether the remaining state can go stale together under a plausible sequence**, rather than on the diff. The state, with its intended scope:

- `PaneFailure.turnAttempted` — per-FAILURE. Set once when the failure is recorded; read by the row (label, accent) and forwarded to both row actions.
- `syntheticDismissed` — per-EPISODE, hook-local to the pane. Reset when `canRetry` goes false.
- `inFlightRetryAfterLogin` — per-FLOW. Set at entry by `relogin`/`loginViaTerminal`, cleared by `endRecoveryFlow`, `false` at rest.

Plus the pre-existing state they interact with, none of which this PR introduced but all of which it now reads or is read beside: `canRetry`, `loginWaiting`/`activeRecoveryFlows`, `reloginInFlight`, `authFailureToPreserve`, `loginCancelled`.

The specific question I cannot answer myself with confidence: **the three scopes are different (per-failure, per-episode, per-flow), and every bug so far has come from a value crossing one of those boundaries.** Is there a sequence where two of them disagree about the same episode — e.g. a failure whose `turnAttempted` says one thing while the flow recovering it carries the other?

Two things I would flag as most likely to hide something, from having been wrong nearby:

1. Overlapping flows. `activeRecoveryFlows` is a counter, so `endRecoveryFlow` only clears `inFlightRetryAfterLogin` when it reaches zero. Two concurrent flows with different intents were not something I traced.
2. The pane surviving across episodes. `syntheticDismissed` resets on `canRetry` false, but a real failure dismissed after that reset is a different path than the one I tested.

I would rather have this answered as an audit now than discover an eleventh reactively. If the answer is that the design cannot be made safe by patching and wants restructuring, say so — that is a legitimate outcome and better learned before merge than after.

Latest is `c2098103e`, which also records the coexistence constraint and the four rules the fixes generalise to, so the next entry point does not repeat them.

GenericAgentX-asaf:

GenericAgentX-asaf:
Fixed in `9ab35e06e` — and I want to flag what it is rather than just that it is fixed.

## This finding was created by the previous fix

Letting `/login` declare an intent (`2bdb2aad`) closed a divergence and opened a clobber. `beginRecoveryFlow` overwrote the shared intent unconditionally whenever a value was passed, with no check that a flow was already running — and `/login` is never gated by `reloginInFlight`, so it can begin mid-flow and silently redefine what the running flow is recovering.

Now **first-flow-wins**: an intent is recorded only when `activeRecoveryFlows` is zero. Guarded inside `beginRecoveryFlow` so it closes for every caller rather than for `/login` specifically. One test, mutation-verified.

## The pattern is the finding

This is the **second time a fix on this branch created the next finding** — the first was the early return that closed the stale-resend and dropped the startup sequence. Both were narrow fixes to a specific repro that moved the problem rather than removing it.

Thirteen findings, and the recent ones increasingly land on state introduced by earlier fixes rather than on the original change. That is the shape @reagentx-workflow named several rounds ago: *"each fix so far has been reactive to a specific repro rather than an audit of the whole state space."* I think that assessment has now been confirmed twice over by my own commits.

So the [state-space audit request](https://github.com/agentmuxai/agentmux/pull/2951#issuecomment-5532332878) still stands and is the thing I actually want answered before merge — specifically whether `turnAttempted` (per-failure), `syntheticDismissed` (per-episode) and `inFlightRetryAfterLogin` (per-flow) can disagree about the same episode, given they have three different lifetimes and every bug so far has come from a value crossing one of those boundaries.

**If the answer is that this wants restructuring rather than patching, that is a legitimate outcome and I would rather hear it now.** The alternative reading — that the remaining risk is bounded and the last two rounds were genuinely the tail — is also possible, but I do not think I can establish it from inside the change.

1952 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:
That is the answer I was looking for, and it is more useful than a full re-review would have been — it bounds the remaining risk instead of restating it.

Taking the three parts as given, with one correction to my own framing:

- **`turnAttempted` cannot diverge.** Your reasoning is the part I could not construct myself: it is scoped to one `PaneFailure` object and always read from wherever that object currently lives — row, preserve capture, or reducer default — and never cached separately. So there is no window. My worry was that three different *lifetimes* implied three chances to disagree; the actual property is that only two of them are independently stored, and the third is always derived. That retires the question I asked rather than deferring it.
- **Surface C is out of scope.** Reads no shared mutable flag, hardcodes `retryAfterLogin: true`. Agreed — and worth noting it means the coexistence constraint is about *entry points sharing state*, not about C specifically.
- **`syntheticDismissed`** — fixed against the one sequence found (`4c5dca039`), not audited exhaustively. Noted as a known limit rather than a clean bill.

## Requesting the scoped review you suggest

@reagentx-workflow — please re-review scoped specifically to **concurrent-flow interleaving on `activeRecoveryFlows` / `inFlightRetryAfterLogin`**, not the diff and not the full three scopes.

What I know and what I do not:

`9ab35e06e` made recording first-flow-wins, which closes the clobber where a joining flow redefines a running one. What it does **not** address is the teardown side: `endRecoveryFlow` only clears the intent when the counter reaches **zero**, so with two overlapping flows the first flow's intent survives the first flow ending and remains live for the second — which by first-flow-wins never got to declare its own. Whether that is correct or a leak depends on which flow's `AuthUrlBox` "Use terminal instead" actually belongs to when two are in flight, and I have not traced that.

Concretely: A begins declaring `true`, B begins declaring `false` (ignored), A ends (counter 1, no clear), B still running, user escapes to terminal -> reads A's `true`. Right or wrong? I genuinely do not know, and it is the last thing on this branch I cannot answer from inside.

GenericAgentX-asaf:
This finding is correct and already fixed — it was reviewed against `9ab35e06e`, and `6b1c17bb4` (pushed a few minutes ago) is exactly its fix. Same diagnosis, independently: the guard lived inside `beginRecoveryFlow` while `relogin`/`loginViaTerminal` wrote directly and bypassed it, so only one ordering was closed.

Verified on the current head — the only remaining assignments to `inFlightRetryAfterLogin` are inside the guarded chokepoint and the teardown clear:

```
312  let inFlightRetryAfterLogin = false;                                   // init
358  if (activeRecoveryFlows === 0) inFlightRetryAfterLogin = retryAfterLogin; // recordRecoveryIntent
384  inFlightRetryAfterLogin = false;                                        // endRecoveryFlow
```

and all three writers now route through it:

```
369  beginRecoveryFlow    -> recordRecoveryIntent(...)
608  relogin              -> recordRecoveryIntent(retryAfterLogin)
1047 loginViaTerminal     -> recordRecoveryIntent(retryAfterLogin)
```

Deliberately a chokepoint rather than the same condition repeated at three sites — repeating it is precisely what produced this, since one writer got the guard and two did not.

Also fixed the specific coverage gap you named: there is now a test for the `/login`-first, row-button-second ordering. Mutation-verified — removing the guard from the chokepoint fails **both** ordering tests, where previously only one existed to fail.

And your correction to `9ab35e0`'s commit message stands: *"closes this for every caller"* was wrong, and I have said so in `6b1c17bb4`'s message rather than leaving the overclaim in the history.

Please re-review against `6b1c17bb4`.

1953 tests pass; `tsc --noEmit` clean.

GenericAgentX-asaf:

PR #2952, opened 2026-09-03
feat(swarm): list each agent's todo checklist and in-flight tool under its name

Agent5-asaf:
Adds two surfaces under each agent's name in the Swarm pane: its **todo checklist**, and the **tool it is running right now**.

## Why the backend

Nothing captured todos before this — there is no `TodoWrite`/`Task*` handling anywhere in `agentmux-srv`, and the frontend only holds tool calls as generic `ToolNode`s in `agent-document-store`, which is populated by **mounted panes only**.

Since the Swarm's whole value is seeing agents you are *not* looking at, the data has to come from where the shell/cron/summary rows already come from.

New `reactive::progress_watcher` mirrors `activity_watcher`'s sweep (same registry, same running/agent-pane gate, no `AppState`) but is a pure tail-read + JSON parse with **no model call**, so it sweeps every **3s** rather than 20s — a checklist that lags is worse than none. It publishes only when the payload actually changes, plus once when a previously-active agent goes quiet so the UI can clear.

## Matching todo tools by name, not by assuming `TodoWrite`

Worth flagging, because it surprised me: **`TodoWrite` is not what AgentMux currently ships against.** Claude Code 2.1.177 advertises `TaskCreate`/`TaskGet`/`TaskList`/`TaskUpdate` and no `TodoWrite` at all, and I found zero `TodoWrite` calls in any local transcript. Other providers differ again.

So the watcher matches a known name set plus a `Todo` substring fallback, and handles both shapes:

- a full `input.todos` array **replaces** the list outright (the tool's own semantics)
- Task-style single-item calls **accumulate** into an ordered map keyed by task id, so an update revises its own row instead of appending a duplicate

An unrecognized status is passed through verbatim and rendered with the pending treatment, rather than coerced to `pending` and silently misreported.

## Rendering

`todoRows` / `todosTruncated` / `currentTool` on `AgentTreeNode`, fed by a `waveEventSubscribe`.

- **Todos render first** among the buckets. The other four label *mechanisms* the agent is using (dispatch, workflow, shell, cron); a checklist is its stated *intent*, and intent reads before mechanism.
- **Rows keep the agent's own order**, never re-sorted by status — a checklist is an ordered artifact, and re-ordering it would be editing the agent's meaning rather than displaying it.
- Bucket count is **done/total**, not a bare row count: for a checklist the useful number is progress.
- The **current-tool line sits outside the collapse guard**, since a collapsed agent is exactly when you want to know what it's doing without expanding it.
- No sixth bucket hue. The four existing hues label mechanisms; the SCSS explains why a checklist gets plain primary text instead of competing for a fifth saturated colour.

## Two traps handled explicitly

1. **`agentChildRowCount` now counts `todoRows`.** That sum gates both the expand affordance *and* the collapse guard that mounts buckets at all — so omitting it would make an agent whose only activity is a checklist render **nothing**, with no way to expand to it. That is the exact shape of the PR #2862 regression `swarm-agent-row.render.test.tsx` exists to pin, and there's a test for it here.
2. **`currentTool` is suppressed for non-running agents.** The payload is push-only, so a stopped pane's last in-flight tool would otherwise linger and read as "still doing this".

Backend rows are capped at 24 with the dropped count reported (`+N more`), so a pathological checklist can't flood the tree and a partial list is never shown as if it were whole.

## Testing

- **Rust: 2963 passed, 0 failed.** 12 parser cases — full-list supersession, Task\* upsert-in-place, explicitly cleared list, truncation accounting, malformed lines, a truncated leading line (tail windows start mid-conversation), and current-tool resolution via unmatched `tool_use` ids.
- **Frontend: 3585 passed, 0 failed.** 13 render cases — order preservation, done/total, per-status treatments, unknown-status fallback, truncation disclosure, the child-count trap, and collapsed visibility.
- `tsc --noEmit` clean.

## Note

Not yet observed against a live agent emitting todos — I couldn't find a transcript containing any todo-tool call to replay, which is itself the finding in §2. The parser is tested against both documented shapes; first real-world exercise will be worth watching.

### DIFF
--- a/.changesets/1788465581-feat-swarm-list-each-agent-s-todo-checklist-and-in-flight-to-h3vd.md
+++ b/.changesets/1788465581-feat-swarm-list-each-agent-s-todo-checklist-and-in-flight-to-h3vd.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(swarm): list each agent's todo checklist and in-flight tool under its name
--- a/agentmux-srv/src/backend/reactive/mod.rs
+++ b/agentmux-srv/src/backend/reactive/mod.rs
@@ -10,6 +10,7 @@
 
 pub mod activity_watcher;
 pub mod handler;
+pub mod progress_watcher;
 pub mod poller;
 pub mod registry;
 pub mod sanitize;
--- a/agentmux-srv/src/backend/reactive/progress_watcher.rs
+++ b/agentmux-srv/src/backend/reactive/progress_watcher.rs
@@ -210,7 +210,7 @@ fn apply_lines(state: &mut BlockState, lines: &[&str]) {
             // item-at-a-time call after it would be describing a different
             // vocabulary than the one currently in force.
             if state.list.is_none() {
-                merge_single_task(&mut state.tasks, input);
+                merge_single_task(&mut state.tasks, input, name);
             }
         }
     }
@@ -314,15 +314,47 @@ fn parse_todo_entry(entry: &serde_json::Value) -> Option<TodoItem> {
 /// reason to resend text that did not change. Requiring a complete item meant
 /// such a call parsed to nothing and the row sat at `pending` forever, which is
 /// precisely the case a checklist exists to show moving.
-fn merge_single_task(items: &mut Vec<(String, TodoItem)>, input: &serde_json::Value) {
+/// Does this tool name describe CREATING an item, as opposed to revising one?
+///
+/// Only consulted for calls that carry no id (reagent P2 on PR #2952). With an
+/// id the key is unambiguous; without one, text is the only identity available,
+/// and keying on it alone silently collapses two genuinely distinct in-flight
+/// tasks that happen to share a title. Two visible duplicate rows are a much
+/// smaller harm than one invisibly missing row, so an idless CREATE always
+/// appends and only an idless UPDATE matches by text.
+fn is_create_call(tool_name: &str) -> bool {
+    tool_name.contains("Create") || tool_name.contains("Add") || tool_name.contains("New")
+}
+
+fn merge_single_task(items: &mut Vec<(String, TodoItem)>, input: &serde_json::Value, tool_name: &str) {
     let text = first_string(input, &["content", "title", "text", "activeForm", "description"]);
     let status = first_string(input, &["status", "state"]);
-    let Some(key) = first_string(input, &["task_id", "taskId", "id"]).or_else(|| text.clone()) else {
-        // Neither an id nor any text — nothing identifies a row to touch.
-        return;
-    };
+    let id = first_string(input, &["task_id", "taskId", "id"]);
 
-    if let Some(slot) = items.iter_mut().find(|(k, _)| *k == key) {
+    // An idless create is its own row, always — see `is_create_call`.
+    if id.is_none() && is_create_call(tool_name) {
+        let Some(text) = text else { return };
+        items.push((
+            // Unique so no later call can match it by key; an idless tool
+            // gives us nothing to correlate a future update against anyway.
+            format!("{}#{}", text, items.len()),
+            TodoItem { text, status: status.unwrap_or_else(|| "pending".to_string()) },
+        ));
+        return;
+    }
+
+    // Find the row this call refers to. An id matches the key it was stored
+    // under; an idless update has only its text, and must compare against each
+    // row's TEXT rather than its key — a row created WITH an id is keyed by
+    // that id, so a key comparison would miss it and append a duplicate.
+    let existing = match &id {
+        Some(id) => items.iter_mut().find(|(k, _)| k == id),
+        None => match &text {
+            Some(text) => items.iter_mut().find(|(_, item)| item.text == *text),
+            None => None,
+        },
+    };
+    if let Some(slot) = existing {
         if let Some(text) = text {
             slot.1.text = text;
         }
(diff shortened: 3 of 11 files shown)


a5af:
@codex review

Agent5-asaf:
Four real defects between you, all fixed in ebdf2d3. Verified each against the code before touching it.

### P1 — checklist couldn't survive the tail window (reagent + codex, same bug)

Correct, and it undercut the exact thing the PR claimed. `TodoWrite` is immune (one call carries the whole list), but `TaskCreate`/`TaskUpdate` — the vocabulary AgentMux actually ships against — describes **one item per call**, so a still-open task vanished the moment its creating line scrolled out of 32KB. Silently, with no indicator, since `todos_truncated` only ever reflected the `MAX_TODOS` cap.

The watcher now keeps per-block state and consumes only bytes appended since the last tick, reading to a **line boundary** so a split JSON object isn't dropped and then never seen in complete form. State resets when the file shrinks (a new session reusing the block). A block first seen already larger than the read cap is seeded from its tail and reports `todosPartial` — deliberately a *different* field from `todosTruncated`, because "history we could not see" and "a cap we chose" are different claims. A whole-list call clears it, being self-contained.

Regression test: `a_task_created_in_an_earlier_tick_survives_later_unrelated_output`.

### P1 — late subscribers got nothing (codex)

Also correct, and the worse failure mode of the two, since it needs no unusual conditions at all: open the Swarm after a checklist has settled and change-suppression means there's nothing left to deliver, so an active agent shows blank indefinitely.

Now `persist: 1` so the broker replays the latest to a freshly-subscribed route, **plus** an unconditional republish every 5th tick (~15s). I did both rather than pick one — the replay path is the documented mechanism, the republish covers any subscriber it doesn't reach, and the cost is one small event per running agent per 15s.

### P1 — status-only `TaskUpdate` never moved its row (codex)

Right, and embarrassing: `{task_id, status}` is exactly what an update with no text change looks like, `parse_todo_entry` required display text, so it parsed to nothing and the row sat at `pending` forever — the one transition a checklist exists to show. Fields are now merged into the existing row. An update for a task we never saw created still creates no row, since there'd be nothing to display.

### P2 — non-Claude providers were invisible (codex)

Confirmed against `gemini-translator.ts`: those transcripts put the frame at the **top level** with `tool_name`/`tool_id`/`parameters`, so my `message.content` walk yielded nothing and Gemini agents showed no in-flight tool and no todos at all. Both shapes are read now, with a guard so an ordinary line carrying a `name` field isn't mistaken for a bare frame.

### P2 — `TaskList`/`TodoRead` dropped (reagent)

Correct that they were dead. Their task data lives in the tool **result**, which this parser never reads, so listing them advertised support that contributed nothing. Removed rather than half-supported, with a comment saying what supporting them would actually take.

### P2 — unbounded progress Map (reagent)

Pruned in `subscribeToBlockStatuses` against the live block list, mirroring the backend watcher's own per-sweep `retain`. That's the general case and the same place the statuses Map is already rebuilt from the same authoritative list — better than keying off one event's payload.

---

**2975 Rust / 3585 frontend, 0 failed.** 12 new parser cases, one per fix.

One unrelated note, since it cost me time and may cost the next agent some: `backend::sysinfo::handle_anomaly_tests::refresh_processes_specifics_does_not_leak_a_handle_per_call` failed 3/3 for me at one point and passed 4/4 immediately after, in interleaved A/B runs with and without these changes. It is sensitive to host process churn, not to this diff — worth knowing before anyone bisects it.

Agent5-asaf:
Both non-blocking P2s fixed in f1196be — they were fair, so I took them rather than deferring.

### Blocking SQLite reads in the async sweep

Correct, and the 3s cadence is exactly what makes it matter: `FileStore::stat`/`read_at` are synchronous, mutex-guarded SQLite reads, run inline for every live agent, parking a Tokio worker on disk I/O several times a second. Fine at `activity_watcher`'s 20s, not here.

The file-reading half now runs in `spawn_blocking`. `states` is moved in and back out rather than shared behind a lock — only this loop ever touches it, and one owner is simpler to reason about than one mutex. A `JoinError` resets state and continues rather than returning: a lost tick re-seeds from the tail (flagged `todosPartial`), which is degraded but recoverable, whereas ending the loop would stop progress reporting for the life of the process.

### Text-keyed dedup collapsing distinct tasks

Also correct. Two genuinely distinct in-flight tasks sharing a title merged into one row, silently. Weighing the two failure modes: one invisibly missing row is worse than two visible duplicates, so an idless **create** now always appends under a unique key.

**Writing the test for that turned up a second bug in the same function**, which I'd otherwise have shipped: an idless **update** looked its row up by *key*, but a row created *with* an id is keyed by that id — so the lookup missed and appended a duplicate instead of updating. Lookup is now by id against the key, or by text against each row's **text**, which is the only correlation an idless call can offer. That one was mine, not in either review.

---

**2979 Rust / 125 swarm, 0 failed.** Four new cases: two idless creates staying two rows, an idless update finding its row, ids taking precedence over text for identity, and create-vs-update tool-name classification.

Agent5-asaf:
Fixed in b16f0a6. This one was my own inconsistency, and a fair catch.

I added `todosPartial` to the backend payload and argued in this very thread that it is a **different claim** from `todosTruncated` — one is a cap we chose and can count, the other is history the backend could not see — and then never wired it to the frontend. So the distinction I made the case for existed only in Rust. A checklist genuinely missing entries (srv restarted mid-session, so `TaskCreate` calls from before that point were never observed) rendered identically to a complete one.

Now carried through `AgentProgress` → the event handler → `AgentTreeNode` → `TodoBucket`, and rendered as its own line — *"earlier items may be missing"* — in the warning hue rather than the neutral one `+N more` uses. That difference is deliberate rather than decorative: a cap working as designed isn't something the reader can act on, a gap in what we could observe is (restart the pane to re-seed cleanly). The `title` attribute explains the cause.

Three render tests: partial disclosed on its own, silence when the list is complete, and **both disclosed together** — they're independent facts, so one must not mask the other.

**3588 frontend tests pass**, `tsc` clean, citations gate green.

PR #2953, opened 2026-09-03
fix(agent-pane): Edit diff preview vanished when syntax highlighting resolved

Agent2-asaf:
The Edit tool's diff showed for a moment and was then replaced by just the file path.

## Root cause

`DiffViewer`'s Shiki-highlighted `<pre>` lives **inside** the `highlightedHtml() !== null` `<Show>`, so the same signal that carries the HTML also creates the element that receives it:

```tsx
createEffect(() => {
    const h = highlightedHtml();
    if (preEl && h !== null) preEl.innerHTML = h;   // preEl may not exist yet
});
...
<Show when={highlightedHtml() !== null} fallback={<PlainDiff />}>
    <div class="agent-diff agent-diff--highlighted">
        <div class="agent-diff-header">{filePath()}</div>
        <pre ref={preEl} class="agent-diff-highlighted-body" />
    </div>
</Show>
```

Applying the HTML from an effect alone loses that race. The effect can run while `preEl` is still `undefined`; the `if (preEl && ...)` guard silently does nothing; and because the signal never changes again, nothing ever fills the freshly created `<pre>`. What survives is the header — which renders the file path and nothing else.

The moment of visible diff is the plain fallback: it renders the real diff while Shiki loads, then the branch swaps to the empty highlighted one.

`HighlightedCode` does the same Shiki dance and is fine, because its `<pre ref>` is unconditionally mounted — that contrast is what identified this.

## Evidence

Confirmed in a live pane, not inferred. For a real Edit, with the row still on screen:

| element | height | content |
|---|---|---|
| `.agent-diff--highlighted` | 20.77 | — |
| `.agent-diff-header` | 20.77 | `…\repro-target.txt` |
| `.agent-diff-highlighted-body` (`<pre>`) | **0** | **`""`** |

The positive `y` matters: the row had not scrolled off, which rules out the scroll-driven `expandedTools` release — the other candidate, since a collapsed tool row also renders as `✏️ Edit <file_path>`. I chased that one first and it was wrong.

## Fix

Apply the HTML from the `<pre>`'s `ref` callback as well as the effect. The ref runs exactly when the element exists, so whichever happens last wins and both orderings end with a filled `<pre>`.

## Why it "used to work"

The `<Show>`-wrapped `<pre>` dates to #388. This is a timing-sensitive race, not a structural break, so it can start showing up without the component changing — Shiki resolve latency vs. the branch swap depends on machine speed, diff size and cache state. Nothing in the last 48 commits touches this path, and `solid-js`/`shiki` have not moved in 60 days.

## Test

A remount with a warm highlight cache reproduces it deterministically — the HTML is resolved synchronously from `diffCache`, so it is known *before* the `<Show>` creates the `<pre>`. On the old code it fails with exactly the live signature:

```
AssertionError: a warm-cache remount must still fill the <pre>: expected '' not to be ''
```

4 tests total (plain fallback, post-Shiki, warm-cache remount, post-highlight node update). Full `app/view/agent` suite: 1620 passing. `tsc --noEmit` clean.

### DIFF
--- a/.changesets/1788465685-fix-agent-pane-edit-diff-preview-vanished-when-syntax-highli-qkcd.md
+++ b/.changesets/1788465685-fix-agent-pane-edit-diff-preview-vanished-when-syntax-highli-qkcd.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): Edit diff preview vanished when syntax highlighting resolved
--- a/frontend/app/view/agent/components/DiffViewer.test.tsx
+++ b/frontend/app/view/agent/components/DiffViewer.test.tsx
@@ -1,0 +1,109 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * The Edit tool's diff preview must survive Shiki resolving, and must survive
+ * the node being updated afterwards.
+ *
+ * User-reported regression: "I see the edited preview for a moment, but then it
+ * disappears and is replaced by a single path to the file."
+ */
+
+import { render, cleanup, waitFor } from "@solidjs/testing-library";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createSignal } from "solid-js";
+
+// Shiki is a heavy async ESM import; stub it with a synchronous-resolving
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #2954, opened 2026-09-03
feat(agent-pane): land the shared content-resize contract (step 2, no call sites)

Agent1-asaf:
## Summary

Step 2 of `SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31.md`. **No call sites — zero runtime effect.** Migration (step 3: `ToolOverlayLog.tsx`'s `flipHeight()`) is a separate, reviewable-in-isolation change, per the spec's own sequencing.

Gated on step 1's data landing first — it did (#2887, and real attribution finally came through today: 284/351 captured shrinks in a live session are `tool`-kind nodes, 0 `markdown`, confirming Source A dominates and de-prioritizing the `MarkdownBlock` migration the spec had queued after `ToolOverlayLog`).

`withHeightContinuity(el, mutate)` and `beginHeightContinuity(el)` — the two call shapes a caller actually has: a mutation you trigger synchronously (a signal write), or one that lands later (a throttled re-render's trailing commit, or a reactive effect reacting to a prop change from outside). Both do the same FLIP technique `ToolOverlayLog.tsx`'s `flipHeight()` already uses — moved here with a shared policy layer: reduced motion, a magnitude cap (skip the ease past 2000px — the 08-22 findings doc's whole-pane collapses were 21K-44K px, not the same phenomenon a content-layer ease was meant to smooth), per-element cancellation, and a content-visibility check.

The content-visibility check is written to directly fix the bug the spec's own §3a found in `ToolOverlayLog.tsx`'s current `heightStale`/`panelHidden`: it reads the **computed** `content-visibility`, not a class-attribute `MutationObserver` signal — the panel's close transition uses `content-visibility ... allow-discrete`, which doesn't actually apply until the END of the transition, so class-based detection reports "hidden" for the whole window it's still visibly collapsing. This PR doesn't fix that bug (no call sites yet) — it falls out for free once step 3 migrates.

## On the test process (worth flagging honestly)

The first pass — 13 tests, all green on the first run — turned out to be a weak signal by itself. Falsifying 5 of the assertions (deleting the exact behavior each claims to test) left the **same 13 tests green**, for three different reasons:
- A no-op fake `cancelAnimationFrame` that couldn't observe cancellation either way (final DOM state ended up correct regardless, since the later mutation's frame just overwrote the stale one's — order, not correctness, was doing the work).
- Two tests whose `mutate()` callback never actually changed the element's height, so "correctly skipped" and "measured, but nothing to animate" were indistinguishable.
- A genuinely redundant `isMeasurable` branch (`offsetHeight > 0 || offsetWidth > 0`) that `shouldAnimate`'s own `fromPx <= 0` guard already covered — confirmed by deleting it and finding nothing broke, then actually removed from the source rather than kept "for safety."

Rewrote the rAF fake with real ids and real cancellation-removal, rewrote every test whose `mutate()` needs to produce an observable height difference, and deleted the dead branch. Re-falsified all 5 — each now fails when its source line is removed, restored, clean diff.

## Test plan

- [x] 13 tests in `resize-contract.test.ts`, covering both entry points, reduced motion, content-visibility, the magnitude cap, cancellation, and independence across elements.
- [x] Every non-trivial assertion falsified individually (delete the behavior, confirm the specific test fails, restore, confirm clean diff) — not just run once and trusted.
- [x] No existing call sites touched — nothing else in the suite can regress from this PR.

### DIFF
--- a/.changesets/1788465977-feat-agent-pane-land-the-shared-content-resize-contract-no-c-ewre.md
+++ b/.changesets/1788465977-feat-agent-pane-land-the-shared-content-resize-contract-no-c-ewre.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(agent-pane): land the shared content-resize contract (no call sites yet)
--- a/frontend/app/view/agent/resize-contract.test.ts
+++ b/frontend/app/view/agent/resize-contract.test.ts
@@ -1,0 +1,263 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * resize-contract.ts — step 2 of SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31.md.
+ *
+ * No component under test here — this file has no call sites yet (that's
+ * step 3). These tests exercise the two entry points directly against a
+ * plain DOM element, mirroring the FLIP-mechanics assertions
+ * ToolOverlayLog.test.tsx already makes for the (soon to be migrated)
+ * `flipHeight()` this replaces, plus the policy layer that function didn't
+ * have to reason about in isolation: reduced motion, unmeasurable
+ * (content-visibility/zero-size) elements, the magnitude cap, and
+ * per-element cancellation.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+let reducedMotion = false;
+vi.mock("@/app/store/global", () => ({
+    atoms: {
+        prefersReducedMotionAtom: () => reducedMotion,
+    },
+}));
+
+import { beginHeightContinuity, withHeightContinuity } from "./resize-contract";
+
+/** Elements whose computed content-visibility should read "hidden". The
+ *  module under test only ever reads THIS one property off
+ *  getComputedStyle's return value, so replacing the global wholesale for
+ *  this file is safe and far simpler than partially mocking jsdom's CSSOM
+ *  (which doesn't recognize content-visibility as a settable inline style
+ *  property at all). */
+const hiddenEls = new Set<Element>();
+function setHidden(el: Element, hidden: boolean): void {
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

Agent1-asaf:
Both P1s accepted — real, verified against source before accepting, not just trusted from the description. Fixed in `cd3d590e0`.

**#1 (ancestor content-visibility):** Confirmed the nesting directly — `ToolOverlayLog.tsx`'s own current code walks `scrollRef.closest(".agent-tool-panel")` to reach the panel, which is proof it's an ancestor of `.agent-tool-overlay-log`, not the same element. `content-visibility: hidden` is set on `.agent-tool-panel--hidden` in `_document-nodes.scss`. Since the property is non-inherited, checking only `el`'s own computed style would never see it. Fixed by walking the ancestor chain instead of checking `el` alone.

**#2 (cancel before measuring a re-entrant mutation):** You're right about the mechanism, and right about why the original tests missed it — an explicit inline `height` (the FLIP's own pin) overrides content-based sizing entirely in a real browser, so `offsetHeight` reports the pinned value regardless of what `mutate()` actually changes, for both the "from" and "to" reads. Fixed by cancelling any in-flight transition on `el` **before** every measurement — at the top of `withHeightContinuity`, and at both capture *and* commit time in `beginHeightContinuity` (a different flip can start on `el` during the gap between those two calls, so commit needed its own cancel too, not just capture).

On the test gap specifically: you're right that `setOffset`'s fixed-value fake couldn't have caught this even with the right intent, since it's decoupled from `style.height`. Added `makeHeightAware()` — a fake where an explicit inline height genuinely overrides a tracked "natural" value, matching the real box model — and wrote the re-entrancy tests against that.

Falsified all three new protections (the ancestor walk, and cancel-before-measure in each entry point) by removing each independently and confirming exactly its own test fails; restored, clean diff. 16/16 total, `tsc` clean.

PR #2957, opened 2026-09-04
Camper@claudius: docs: retro — VM suspend leaves a running-but-frontend-less AgentMux instance

GenericAgentX-asaf:
## Summary

Found while testing the fresh-PC onboarding fix (#2942/#2943/#2947) live on a Windows VM. Docs-only — no code change here, though the retro proposes one (not yet implemented).

The VM slept for ~4h15m (confirmed via the Windows System event log). The running instance's WebSocket connection died silently at that moment — VM suspend delivers no graceful TCP close. When later double-clicked, the launcher correctly detected the existing instance and forwarded `open_new_window` (working exactly as designed), but the resulting window was a pre-warmed pool window whose own connection was equally dead from before the sleep. The only liveness check before promoting a pool window (`agentmux-cef`'s `promote_pool_window`) checks that the CEF browser handle still exists in host state -- not that its connection or rendering is actually functional. Four overlapping top-level windows now exist on the VM, none confirmed working.

Root cause chain, timeline, and code citations are documented directly against real log timestamps and the actual `promote_pool_window`/`pool_respawn_on_promote` code -- not inferred. One data point recorded honestly as inconclusive rather than over-claimed: a `vmrun captureScreen` came back fully black, which could mean a genuinely blank display or could just be VMware's console not getting real framebuffer content while Parsec drives the display -- flagged as ambiguous, not treated as proof.

**Recommendation** (matches what was explicitly asked for): verify -- not just dispatch -- that a promoted/forwarded window is actually alive before considering `open_new_window` satisfied, falling back to a genuinely fresh window if it isn't within a bounded timeout. Also flags invalidating the pool on system-resume to close the root cause rather than only the symptom, and a separate ops-layer fix (the VM's DC/battery sleep timer, not the AC one, is what actually fired).

### DIFF
--- a/docs/retro/retro-fresh-vm-suspend-orphaned-frontend-2026-09-03.md
+++ b/docs/retro/retro-fresh-vm-suspend-orphaned-frontend-2026-09-03.md
@@ -4,6 +4,7 @@
 **Author:** Camper
 **Severity:** Medium — no data lost, the instance's backend stayed healthy throughout, but the user's only recovery action (double-click the exe) produced no working window. On a real user's machine (not an agent-operated VM), this reads as "AgentMux is broken, I have to kill it in Task Manager."
 **Area:** `agentmux-launcher` single-instance forwarding (`other_instances.rs`), the `pool_respawn_on_promote` saga (`saga/pool_respawn.rs`), `agentmux-cef`'s `promote_pool_window` (`commands/window_pool.rs`), Windows/VM power management.
+**Status:** active — root cause identified and evidenced; the recommended fix (verify promoted/forwarded windows are actually alive, fall back to a fresh window otherwise) is not yet implemented.
 
 ---
 
--- a/docs/retro/retro-fresh-vm-suspend-orphaned-frontend-2026-09-03.md
+++ b/docs/retro/retro-fresh-vm-suspend-orphaned-frontend-2026-09-03.md
@@ -1,0 +1,67 @@
+# RETRO — VM suspend leaves a "running but frontend-less" AgentMux instance; double-click on the exe recovers nothing visible
+
+**Date:** 2026-09-03
+**Author:** Camper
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both real, fixed in `483b7e3`.

**P2 #1 (disconnect classification)** — you're right, and checking directly confirmed it: `handle_ws_connection` logs `"WebSocket client disconnected"` and runs `unregister_ws`/`unsubscribe_all` right after its read loop exits, and `ws.ts`'s `onclose` handler drives a real reconnect loop (up to 20 backoff-timed attempts). Neither claim in the original draft survived checking the actual code. Corrected to: the disconnect was detected and handled correctly at both ends. What's genuinely unproven is *why* reconnection didn't restore a working state across the multi-hour suspend -- added a mechanism-grounded hypothesis (browser `setTimeout` doesn't fire while suspended, so the reconnect budget could exhaust right at wake) without claiming it as confirmed, and flagged it explicitly in Open Questions as the now-load-bearing unresolved point.

**P2 #2 (wrong platform branch)** -- confirmed, and it was a real mistake, not just an omission: I quoted the `#[cfg(not(target_os = "windows"))]` fallback for a Windows incident. Replaced with the actual Windows path (`window_pool.rs:1009`), which resolves the CEF browser/host, the reported-or-cached HWND, and validates it via Win32 `IsWindow()` -- a genuinely more thorough check than what I'd originally quoted. Corrected the "why insufficient" reasoning to match: `IsWindow()` confirms the OS hasn't destroyed the handle, which a VM suspend never does regardless of renderer/connection health -- not "there's no check at all," which was the prior, wrong framing.

The core finding survives both corrections (checking handle validity isn't the same as checking functional liveness, and `open_new_window` should verify rather than just dispatch) -- what changed is how confidently the retro claims to have already proven the specific failure mechanism, which it hadn't.

GenericAgentX-asaf:

PR #2958, opened 2026-09-04
fix(agent-pane): tool-preview code wastes half its width on indentation

Agent4-asaf:
## Summary

The 2026-08-24 `tab-size: 2` fix (#2785) could only ever move **tab**-indented previews. Measured against a real stored `Read` of this repo's own source: **0 of 127 indented lines used tabs**, so that rule was inert — and no CSS property narrows a space. Meanwhile 58 of those 200 lines sit at column 0, which forces `dedent.ts`'s common prefix to empty. Neither lever fired; the deepest lines rendered at **column 24**.

#2785 also introduced a regression. `Read` results are `<N>\t<code>` with the number **left-aligned** (`9\t` then `10\t`, confirmed verbatim from a stored `tool_result`). That gutter tab lands on a tab stop whose column depends on the digit count, so at `tab-size: 2` the code's left edge stepped sideways at the 9→10 boundary — i.e. in nearly every preview. At `tab-size: 4` the only step was at line 1000.

## What changed

Three changes, in the text rather than the stylesheet:

- **`normalizeIndentWidth`** rescales leading runs to 2 columns per level, with the unit inferred as the **GCD** of observed widths. That makes it self-limiting in the right way: clean 4/8-space files rescale, while a file containing continuation-alignment lines yields a GCD of 1 and is left alone rather than having aligned code shifted out of alignment. Same decline-rather-than-guess posture the module already takes for tabs vs spaces.
- **`splitNumberedGutter` / `renderNumberedGutter`** re-emit the `Read` gutter right-aligned at fixed width with no tab, removing the raggedness.
- **The markdown `Read` path** now gets a gutter-free body, fixing the corruption `SPEC_TOOL_PREVIEW_DEDENT_2026_08_08.md` §2.1 flagged in August and deferred (a `1\t# Title` line is not a heading).

Edit hunks narrow with **one unit shared across both sides**, for the same reason the dedent prefix is shared — a per-side unit could rescale one side only and manufacture a phantom indentation diff.

**Wrapping**, the other half of the report: the surfaces that wrap (Bash output, Grep/JSON) are exactly the ones that get no dedent, and a wrapped continuation restarted at column 0 — indistinguishable from a real new line. They now carry a hanging indent. Streaming chunks were `pre` while the final render was `pre-wrap`, so a Bash block visibly reflowed the moment it finished; both now agree. `.agent-diff` gets an explicit `white-space` because its two render paths use different elements (`<pre>` plain, `<div>` highlighted), which flipped the header's layout when Shiki resolved.

## Measured effect

On the real transcript sample: deepest line **column 20 → 12**, gutter column constant, all four nesting levels preserved.

## Testing

- 3588 frontend tests pass (240 files), `tsc --noEmit` clean, `vite build` clean.
- `dedent.test.ts` includes 4 tests asserting against a **verbatim real transcript sample**, including one that fails if the gutter ever goes ragged again.

## Notes for review

- Analysis with the measurement method: `docs/analysis/tool-preview-indentation-and-wrapping-2026-09-02.md`
- **Not** visually verified in a live pane yet — the logic is unit-tested against real captured data, but no screenshot. Flagged honestly.
- The gutter-as-a-separate-column treatment was considered and **not** built: aligning a parallel column depends on Shiki's line boxes (it separates `<span class="line">` with literal `\n` inside an inline `<code>`), which I couldn't verify without a live pane. Recorded as a follow-up; the split/render helpers are factored so it becomes a rendering-only change.

### DIFF
--- a/.changesets/1788367009-fix-agent-pane-tool-preview-code-previews-waste-half-their-w-h8u5.md
+++ b/.changesets/1788367009-fix-agent-pane-tool-preview-code-previews-waste-half-their-w-h8u5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): tool-preview code previews waste half their width on indentation
--- a/docs/analysis/tool-preview-indentation-and-wrapping-2026-09-02.md
+++ b/docs/analysis/tool-preview-indentation-and-wrapping-2026-09-02.md
@@ -1,0 +1,386 @@
+# Analysis: tool-preview code still reads over-indented, and some lines wrap
+
+**Date:** 2026-09-02
+**Author:** Agent4
+**Repo state:** `agentmuxai/agentmux` main @ `01cd708e`
+**Method:** static read of the render path + CSS cascade, arithmetic on the
+tab-stop behaviour, and **measurement against real stored transcripts**
+(`~/.agentmux/shared/agents/transcripts/filestore.db`, opened read-only) to
+confirm the actual `Read` result format and the actual indentation mix. Not
+verified by eye in a live pane — see §7.
+
+**Prior work being followed up:** PR #2785 / commit `aa1d0a14`,
+*"fix(agent-pane): render tool-preview tabs at 2 spaces instead of the inherited
+4"* — a one-line CSS change adding `tab-size: 2` to `.agent-tool-panel`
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

PR #2959, opened 2026-09-04
feat(jekt): cross-channel trust — spec + publish per-agent public keys (phase A)

Agent4-asaf:
## Summary

Host, LAN and WAN-reagent all have working cryptographic sender verification. **Cross-channel — same machine, different AgentMux instance — does not**, and it is the tier that renders as `DELIVERY=host`, the label that reads as most trustworthy. Every cross-channel jekt on this machine currently arrives `TRUST=self-declared`.

Found live: a message between two instances on this machine was correctly refused by the receiving agent because it had no way to tell who sent it.

## Root cause

A gap between two correct subsystems. Agent identity went host-global in #1387–#1396 — definitions, instances, workspaces, auth and transcripts all live under `~/.agentmux/shared/agents/` now. **The jekt signing keys were not part of that sweep**; they are still minted per-channel in each channel's own `objects.db`. Measured on this machine:

| store | `db_agent_jekt_keys` |
|---|---|
| host channel | `agent1…agent5, agentx, agenty` |
| dev channel | `lark` |

So an agent's *identity* is global but the means of *proving* it is not, and `verify_jekt_signature`'s local-store lookup returns `None` every time.

## What this PR contains (Phase A only — strictly additive)

Nothing reads `jekt_public_key` yet, so **no message's `TRUST=` can change until Phase B**.

- **D1** — `AgentEntry` gains `jekt_public_key`: the base64 Ed25519 **public** half of the agent's existing LAN keypair, published to the host-global shared registry on every registration. Reuses `AGENTMUX_LAN_KEY`; `agent_lan_key_ensure` already runs unconditionally for every agent at config-injection time, so coverage is complete on day one (measured: 7/7 host-channel agents, 1/1 dev-channel).
- **D5** — `sign_channel_jekt` / `verify_channel_jekt`, deliberately **not** reusing the LAN signed material.

## Two design points worth reviewing closely

**Why not just globalize `db_agent_jekt_keys`?** That is the obvious move and is very likely what was assumed to have happened already. §5 rejects it: that key is a **symmetric** HMAC secret, so a shared copy would let every srv build on the machine *mint* signatures for every agent in every channel. This machine routinely runs many per-build channels off arbitrary branches, so that is strictly worse than the gap it closes. Asymmetric publication has no such property. **Corollary: `db_agent_jekt_keys` must stay per-channel.**

**Why a separate signing scheme from LAN?** Two reasons, both with tests:
- A **domain separator**. Same key, same tuple — without it a LAN signature and a cross-channel signature would be byte-identical and each would verify as the other, collapsing two different trust claims verified against two different key sources.
- **`source_channel` in the signed material.** One agent name can legitimately be live in several channels at once, so without binding, a signature minted in channel A replays as that agent speaking from channel B.

## The backward-compat edge

An entry written before this shipped has no `jekt_public_key` member. It must deserialize (serde default) **and an empty value must mean "cannot check", never "failed the check"** — otherwise every pre-upgrade sender on a mixed-version machine gets escalated. Both halves are tested. Verified `AgentEntry` has no `deny_unknown_fields`, so older readers tolerate the new field.

## Testing

168 `agentmux-common` + 2987 `agentmux-srv` tests pass. The registry test is **Store-backed with the same closure shape bootstrap installs**, so it exercises the real resolver→Store→entry wiring rather than a stub — the last time a jekt key feature shipped, the code was correct and the real call path simply never ran it (`REPORT_JEKT_SIGNING_KEY_INJECTION_GAP_2026_08_16.md`).

Verified end-to-end against a running instance: after a restart on this branch, a registry entry is written carrying a populated public key.

## Not in this PR

Phase B (verification), Phase C (enforcement — a resolvable sender who doesn't sign becomes a red flag), and the escalation-chaining rule (§7.3). Phase C must not land until Phase A has propagated.

Spec: `docs/specs/SPEC_JEKT_CROSS_CHANNEL_TRUST_2026_09_02.md`

### DIFF
--- a/.changesets/1788386725-feat-jekt-publish-per-agent-public-keys-for-cross-channel-tr-21oh.md
+++ b/.changesets/1788386725-feat-jekt-publish-per-agent-public-keys-for-cross-channel-tr-21oh.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(jekt): publish per-agent public keys for cross-channel trust (phase A)
--- a/agentmux-common/src/jekt_sign.rs
+++ b/agentmux-common/src/jekt_sign.rs
@@ -282,6 +282,109 @@ pub fn verify_lan_jekt(
     verifying_key.verify(material.as_bytes(), &signature).is_ok()
 }
 
+// ── Cross-channel per-agent Ed25519 signing (SPEC_JEKT_CROSS_CHANNEL_TRUST_2026_09_02.md §D5) ──
+//
+// Same machine, different AgentMux instance. Asymmetric for the same reason
+// LAN is: the receiving instance must be able to verify without being able to
+// forge — and unlike host-tier's HMAC, the key can therefore be published in
+// the host-global shared registry, which is what makes cross-channel
+// verification possible at all (§5 of the spec explains why globalizing the
+// symmetric host key instead would be strictly worse).
+//
+// Reuses the agent's EXISTING LAN keypair — no new key material, no new
+// provisioning step. What it does NOT reuse is the signed material.
+
+/// Domain separator for cross-channel signatures.
+///
+/// Without this, a cross-channel signature and a LAN signature over the same
+/// (msgid, source, target, ts, message) tuple would be byte-identical, signed
+/// by the byte-identical key — so either could be replayed as the other. The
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #2960, opened 2026-09-04
fix(reactive): deliver to persistent agents that have not spawned yet

Agent4-asaf:
## Summary

After **any** srv restart, a persistent (stream-json) agent could not receive agent-to-agent messages until a human sent it one from the UI. Every jekt failed with `persistent process not running`.

Found live, not by inspection: three sends to an agent in a freshly restarted instance all failed, while the forward itself was fine — it reached the target srv, returned 200, and the target logged `reactive inject request received` before refusing to deliver.

## Mechanism

Three individually-correct pieces with a gap between them:

1. Controllers register lazily — `persistent controller registered (spawns on first message)`.
2. `inject_message` writes to a live `stdin_tx` or fails — deliberately; its own comment says *"this function has no spawn config."* It is a mid-turn **steer**, not a way to start a process.
3. The reactive path rightly refuses to fall back to PTY keystrokes, which a stream-json agent rejects outright.

The third option nobody took was to **start a turn**. #2930 built exactly that machinery — `run_agent_turn` re-reads spawn config from block meta and already dispatches to persistent controllers via the spawn-capable `send_message` — but gated its caller to subprocess controllers (*"Only the subprocess case changes"*). A registered-but-unspawned persistent controller fell into the other branch. It is **not** in #2930's "what this PR does not change" list, so it reads as unnoticed rather than deferred.

## Fix

Widen that path to the one other controller kind that needs it, and only in the state that needs it:

```rust
pub fn needs_spawn(&self) -> bool {
    let inner = self.inner.lock().unwrap();
    inner.stdin_tx.is_none() && !inner.spawning_in_progress
}
```

**The second half is load-bearing.** A caller that has already claimed the spawn owns that round; reporting "needs spawn" inside that window would invite a second racing spawn and orphan a child — the bug `spawning_in_progress` exists to prevent (reagentx P1 on #2360). That window stays retryable (`"still starting up"`) rather than becoming spawnable. There is a test for exactly this.

No double-delivery: `inject_message` returns its error *before* the blockfile append, so nothing was written on the failing path.

#2930's three constraints are inherited unchanged — no optimistic `Ok(true)` before the fallible work runs, `block_in_place` under the reactive `Handler` mutex, and `TurnRegistration::Skip`.

## Verified end-to-end, not inferred

Precondition established and checked rather than assumed — instance restarted, controller registered with no process, pane not opened by anyone. Then a cross-channel jekt:

```
07:52:27  persistent controller registered (spawns on first message)
07:52:49.089  reactive inject request received
07:52:49.090  reactive delivery: persistent controller not yet spawned — starting a turn instead
07:52:49.258  persistent process spawned
07:52:56      blockfile:line_count            ← message landed in the agent's conversation
```

169 ms, no `structured delivery failed`. The message that previously would have been dropped is what started the agent.

## Known caveat this PR does NOT fix

The sending client can see a **transport timeout even though delivery succeeded** — the spawn is synchronous under `block_in_place` (#2930's deliberate "bounded stalling beats silent loss" trade) and a cold CLI spawn can outlast the caller's HTTP timeout. Strictly better than losing the message, but a retry-on-timeout would now **duplicate** it rather than error, and a transport timeout is not the `success: false` that `cloud_subscriber`'s retry path keys on.

Recorded in §6.1 with two candidate designs (async accept, or longer timeout + idempotency key). Left unfixed deliberately — picking wrong turns a lost message into a doubled one.

## Testing

3 unit tests on the predicate (including the racing-spawn window); 2987 `agentmux-srv` tests pass. The pre-existing test asserting `persistent process not running` is surfaced rather than swallowed is unaffected and still correct — it mocks the message sender, and surfacing a genuine failure is still required; this only adds a recovery ahead of it.

Report: `docs/reports/REPORT_JEKT_DELIVERY_DROPS_UNSPAWNED_PERSISTENT_AGENTS_2026_09_03.md`

### DIFF
--- a/.changesets/1788421232-fix-reactive-start-a-turn-for-persistent-agents-that-have-no-ir1w.md
+++ b/.changesets/1788421232-fix-reactive-start-a-turn-for-persistent-agents-that-have-no-ir1w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(reactive): start a turn for persistent agents that have not spawned yet
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -4106,20 +4106,43 @@ impl PersistentSubprocessController {
     /// tell those two situations apart. See
     /// `docs/reports/REPORT_JEKT_DELIVERY_DROPS_UNSPAWNED_PERSISTENT_AGENTS_2026_09_03.md`.
     ///
-    /// Deliberately `&& !spawning_in_progress`: a caller that has already
-    /// committed to spawning owns this round (see `spawning_in_progress`'s doc
-    /// comment on the concurrent-spawn TOCTOU race). Reporting "needs spawn"
-    /// during that window would invite a second, racing spawn — the exact
-    /// orphaned-child bug that field exists to prevent. That window instead
-    /// surfaces as "still starting up — try again shortly", which is retryable
-    /// by design.
+    /// **The invariant this must uphold:** `needs_spawn() == true` implies a
+    /// subsequent `send_message` takes `decide_send_action`'s `BecomeSpawner`
+    /// branch — i.e. it really spawns and delivers. It must never be true in a
+    /// state where `send_message` would instead return `SendAction::Queued`,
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2961, opened 2026-09-04
Camper@claudius: fix(launcher): widen the Windows splash window (280px -> 460px)

GenericAgentX-asaf:
## Summary

Follow-up to the fresh-PC onboarding investigation (#2940). The first launch on a genuinely clean Windows machine takes ~36s, ~30s of which is Windows Defender's cloud reputation check on newly-extracted, unsigned binaries -- confirmed empirically (closed and relaunched on the same machine: 23s -> 1s for the identical gap). Surfacing that wait as a staged status message in the splash instead of it just sitting there is a small, contained follow-up -- except the splash window was only 280px / ~28 monospace characters wide, with no word-wrap (`splash_text.rs` clips silently past the DIB edge), too narrow to fit a one-line reassurance message.

`SPLASH_SIZE` previously did double duty as both the window's WIDTH (`SPLASH_W = SPLASH_SIZE`) and the brain-logo region's HEIGHT contribution to `SPLASH_H`. Decoupled the two:
- `BRAIN_AREA_H` keeps the old value (280px) for everything that used it as a height (`SPLASH_H`, the stage-area Y offset, the separator line's Y position).
- `SPLASH_W` is now independent at 460px (~43 chars/row), sized to comfortably fit a full status line.
- The brain logo is now centered horizontally in the wider window (`BRAIN_X = (SPLASH_W - BRAIN_W) / 2`) rather than pinned at a fixed left padding, so it doesn't visually drift to one side.

No other layout constant changed -- stage rows, the separator line, the footer, and the border all already read `SPLASH_W`/`SPLASH_H` from the shared constants, so they scale automatically.

## Test plan

- [x] `cargo build -p agentmux-launcher` -- compiles clean, no new warnings
- [ ] **The actual rendered window on screen -- not verified, disclosed honestly rather than claimed.** The splash is a native `WS_EX_TOOLWINDOW` Win32 popup created with no window title text -- by design, excluded from normal window enumeration (so it's not alt-tabbable) -- which is outside what this session's available window-capture tooling can reach. Tried both `DiscoverWindows` and `CaptureWindow` against a live `task dev` launch and neither found it, before or after the change. The constant-math change is small and mechanical enough to review by hand (and the derivation is documented inline), but someone with a normal desktop session should confirm the actual rendered appearance before this ships in a release.

### DIFF
--- a/.changesets/1788484329-fix-launcher-widen-the-windows-splash-window-to-fit-staged-s-ic96.md
+++ b/.changesets/1788484329-fix-launcher-widen-the-windows-splash-window-to-fit-staged-s-ic96.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(launcher): widen the Windows splash window to fit staged status messages
--- a/agentmux-launcher/src/splash.rs
+++ b/agentmux-launcher/src/splash.rs
@@ -49,8 +49,22 @@ use crate::startup_events::{StartupEvent, StartupStatus};
 include!(concat!(env!("OUT_DIR"), "/brain_dims.rs"));
 
 const SPLASH_PADDING: i32 = 12;
-const SPLASH_SIZE: i32 = BRAIN_W + SPLASH_PADDING * 2;
-const BRAIN_X: i32 = SPLASH_PADDING;
+// Vertical size of the brain-logo region only (280px) — was also the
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #2962, opened 2026-09-04
fix(agent-pane): migrate ToolOverlayLog's height FLIP to the shared resize contract (step 3)

Agent1-asaf:
## Summary

Step 3 of `SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31.md`. Gated on step 1's real data (#2887): 284/351 captured shrinks in a live session are `tool`-kind nodes, 0 `markdown` — confirms this migration target (Source A) dominates.

**`resize-contract.ts`** gets an optional `measure` parameter on both entry points — discovered necessary, not designed in advance. The migration target, `.agent-tool-overlay-log`, scrolls its own overflow (`overflow-y: auto`) inside `.agent-tool-panel`'s `max-height: 50vh` cap, so the module's default `offsetHeight` read clamps at whatever's left of that budget and stops changing once content exceeds it, while `scrollHeight` keeps reflecting the true content height — exactly the large-shrink case (long raw chunk log → short compact result) this whole effort exists to fix. `offsetHeight` alone would have silently missed it. Defaults unchanged for existing/future non-scrolling consumers.

**`ToolOverlayLog.tsx`** — deleted `flipHeight()`/`HEIGHT_FLIP_MS` and the `lastMeasuredHeight`/`heightStale`/`cancelHeightFlip` state they needed. Replaced with `beginHeightContinuity(el, measureHeight)`, captured fresh every effect invocation (so the eventual FLIP's "from" is always the most recent height, not whatever it was when the branch started) and committed only when the branch actually changed. Re-baselines for the *next* invocation **before** committing the current one — `beginHeightContinuity` cancels any in-flight transition as part of capturing a trustworthy height, and doing that after committing would cancel the flip just started in the same tick. Kept the `nodeId` slot-reuse guard (simplified — just discards a pending commit now, no branch/height bookkeeping to carry).

This also resolves the spec's own §3a (`heightStale`/`panelHidden` allow-discrete lag) — not a targeted fix, a side effect of `isMeasurable()` reading computed `content-visibility` on the real ancestor chain instead of a class-based `MutationObserver` signal, exactly as that section predicted it might.

## On the tests (worth reading, not just the summary)

Three of the existing FLIP tests were weaker than they looked — same root cause each time: a constant `scrollHeight` stub across the compared reads meant `shouldAnimate`'s own zero-delta check masked whether the thing actually under test (branch-change gating, hidden-detection) was doing anything. Rewrote each so the height genuinely differs across the reads being compared.

The hidden-panel test's **first** fix was itself wrong: I mocked `getComputedStyle` using `el.closest(".agent-tool-panel--hidden")`, which climbs ancestors *itself* — so the test stayed green even after I deliberately broke `resize-contract.ts`'s own ancestor walk to check. Fixed to check only the exact queried element's own class (`el.classList.contains(...)`), matching real non-inherited `getComputedStyle` semantics, then reconfirmed the break is now actually caught.

Also falsified the branch-changed gate, the `nodeId` guard, and the capture-before-commit ordering (the specific self-cancellation bug this ordering avoids) by removing each independently and confirming the right test fails.

Spec doc updated: §3a marked resolved, §4's proposed signatures given the `measure` param neither original draft anticipated, §5 steps 1-3 marked done, mechanism table's `ToolOverlayLog.tsx` line ranges re-verified against the current file.

## Test plan

- [x] `npx vitest run` across `resize-contract`, `ToolOverlayLog`, `PersistentShellBlock`, `output-cap`, `ToolBlock` — 123/123.
- [x] Falsified: `measure` param usage in both entry points, the `nodeId` guard, the `branchChanged` gate, the capture-before-commit ordering, and the ancestor-walk fix at the consumer/integration level (not just resize-contract.ts's own unit tests) — each confirmed to fail when its corresponding source line is removed, restored, clean diff.
- [ ] Not yet observed in a running instance.

### DIFF
--- a/.changesets/1788484414-fix-agent-pane-migrate-tooloverlaylog-s-height-flip-to-the-s-ka2u.md
+++ b/.changesets/1788484414-fix-agent-pane-migrate-tooloverlaylog-s-height-flip-to-the-s-ka2u.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): migrate ToolOverlayLog's height FLIP to the shared resize contract
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -315,7 +315,7 @@ partial list.
 | [`jekt-visibility-completion`](jekt-visibility-completion.md) | Spec: Jekt Visibility Completion — persistent-agent visibility + outgoing echo |
 | [`swarm-active-pane-sync`](swarm-active-pane-sync.md) | Swarm ↔ Pane Two-Way Active-Row Sync |
 
-### active (12)
+### active (13)
 
 | Spec | Title |
 |---|---|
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

Agent1-asaf:
P2 accepted — real regression, verified against `MAX_TOOL_OUTPUT_LINES` (1000) before accepting: any raw chunk log anywhere near that cap has a `scrollHeight` delta easily in the tens of thousands of px against a short terminal result, comfortably past `MAX_ANIMATED_DELTA_PX` (2000), while the box's actual RENDERED shrink stays bounded by the ancestor panel's `max-height: 50vh` to a few hundred px. Gating on the unclamped number meant exactly the long-output transitions this migration exists to smooth would jump-cut instead. Fixed in `77915cb55`.

Implemented your suggestion directly: `resize-contract.ts`'s two entry points gain a second optional `measureForGating` parameter (defaulting to `measure` — a no-op for every caller without this divergence). `shouldAnimate`'s checks now run against `measureForGating`'s values; `flip()` still eases between `measure`'s true endpoints. `ToolOverlayLog.tsx` passes `scrollHeight` for the FLIP and `offsetHeight` for gating.

One thing worth flagging from getting there: fixing this exposed a real gap in two PRE-EXISTING tests. `offsetHeight` is 0 by default in jsdom, and it now genuinely matters for `shouldAnimate`'s `fromPx<=0` guard — two tests that only ever stubbed `scrollHeight` broke the moment gating started reading a real (zero) `offsetHeight`. Fixed by giving those two tests a nonzero, genuinely-changing `offsetHeight` stub rather than loosening the new gate to tolerate an untested value — the tests were relying on an implicit assumption (offsetHeight doesn't matter) that stopped being true.

Added 4 new `resize-contract.test.ts` unit tests plus one `ToolOverlayLog.tsx` integration test reproducing your exact scenario end to end. Falsified all of it — broke `measureForGating` in each entry point, and broke `ToolOverlayLog`'s own wiring of it — confirmed each fails correctly, restored.

34/34 across the two directly-touched suites, 128/128 including the three unaffected related ones, `tsc` clean.

PR #2963, opened 2026-09-04
Camper@claudius: docs: correct the VM-suspend retro — real root cause was vmrun/Session 0

GenericAgentX-asaf:
## Summary

Corrects a real, significant mistake in the merged VM-suspend retro (#2957). Discovered live, with the repo owner watching their own screen at the same time as the investigation: every AgentMux instance this whole investigation launched via `vmrun runProgramInGuest` -- including the very first one, whose sleep/wake cycle the original retro describes -- ran in Windows **Session 0** (the non-interactive services session), not the real interactive **Session 1** the repo owner actually sees via Parsec. Confirmed directly via `query session`:

```
agentmux pid=11052 SessionId=0
explorer pid=5596 SessionId=1
winlogon pid=732 SessionId=1
```

This is a known VMware/VIX limitation -- guest automation lands in Session 0 regardless of `-activeWindow` or real guest credentials -- not an AgentMux bug. Since AgentMux's single-instance pipe is global across sessions, the Session-0 instance correctly absorbed every subsequent launch attempt (including the repo owner's own genuine double-clicks), doing everything right internally while being structurally incapable of ever showing a window to anyone. Every log signal the original retro checked was genuinely healthy -- for an instance nobody could ever see.

Resolution: killed the Session-0 instance, confirmed zero AgentMux processes remained, had the repo owner double-click it themselves. Worked immediately.

Added an Update section rather than rewriting the original -- it was an honest record of what the evidence available at the time supported, and the code-level observations in it remain accurate readings of the actual code. What changed is that they're no longer proven to be what caused this specific incident. Downgraded the fix recommendation from "confirmed necessary" to "reasonable general hardening." The ops-layer power-timer finding is unrelated and unaffected.

**Lesson recorded for future agent-driven VM testing**: don't use `vmrun`/VIX guest automation to launch or interact with anything meant to be visible to a human -- it's fine for file transfer/scripts/state queries, but any GUI process it spawns lands in an unwatchable session.

### DIFF
--- a/docs/retro/retro-fresh-vm-suspend-orphaned-frontend-2026-09-03.md
+++ b/docs/retro/retro-fresh-vm-suspend-orphaned-frontend-2026-09-03.md
@@ -4,7 +4,7 @@
 **Author:** Camper
 **Severity:** Medium — no data lost, the instance's backend stayed healthy throughout, but the user's only recovery action (double-click the exe) produced no working window. On a real user's machine (not an agent-operated VM), this reads as "AgentMux is broken, I have to kill it in Task Manager."
 **Area:** `agentmux-launcher` single-instance forwarding (`other_instances.rs`), the `pool_respawn_on_promote` saga (`saga/pool_respawn.rs`), `agentmux-cef`'s `promote_pool_window` (`commands/window_pool.rs`), Windows/VM power management.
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2965, opened 2026-09-04
fix(agent): provider_flags were dropped from cmd:args on every send

### LINKED ISSUE #2872: provider_flags and --fork-session are discarded from cmd:args on every send (all agents, not just container)

Agent2-asaf:
Closes #2872.

## The bug

The per-turn `cmd:args` rebuild derives its base from the provider **catalog**, so anything the launch path appended afterwards is dropped:

```ts
const baseArgs = selectLaunchArgs(prov, agentMode);      // catalog
const updatedArgs = buildRuntimeArgs(baseArgs, runtimeConfig, prov.id);
```

`agent.provider_flags` is a real, user-configurable field appended at `agent-model.ts:446`. It was gone from the **first send** onward, for the life of the pane.

Not container-specific, and not new: host agents lost it identically, ever since the runtime controls landed in `f702ceae4` — which predates `provider_flags` itself. The two features were built independently and the rebuild simply never knew about anything appended after the base args.

## The part that isn't mechanical

The launch path appends **two** things, and they are not the same kind of thing:

| | kind | per-turn |
|---|---|---|
| `provider_flags` | durable — how this agent always runs | reapply |
| `--fork-session` | one-shot — fork the session being resumed **at launch** | do **not** reapply |

`--fork-session` is paired with a specific `continueSessionId` (`fork-session-args.ts`). Once forked, the agent has its own session and later turns resume it normally — so reapplying the flag would pair it with the `--resume <sid>` srv adds on every subsequent turn, forking again each time instead of once.

That is why this is fixed **where the args are appended**, rather than by having the rebuild preserve whatever it finds in the live argv. "Just use the existing `cmd:args` as the base" works mechanically — `buildRuntimeArgs` is idempotent — but it would carry the one-shot intent forward too, which is plausibly worse than the bug. That reasoning is why I filed #2872 instead of folding this into #2867 at the time.

## Change

- `provider_flags` is persisted at launch under `agent:provider_flags` and reapplied by both rebuild sites: the pre-send rebuild in `useAgentCommands`, and `applyRuntimeChange` (`/model`, `/effort`, `/mode`).
- `withProviderFlags` / `parseProviderFlags` live in `launch-args.ts`, next to the rule they belong with. `withProviderFlags` only ever knows about provider_flags, so a future caller can't route a one-shot intent through it by accident.
- `applyRuntimeChange` now takes the block's `meta` rather than one extracted field — it needs two of them now, and a third would otherwise mean a third positional parameter.
- srv's `container_argv` already preserves unknown tokens (it subtracts persistent-only flags rather than rebuilding), so a container agent's flags survive its self-heal too.

## Limits, stated plainly

Panes launched **before** this meta key exists have no flags to restore — theirs were already dropped and are not recoverable from the argv. They pick it up on next launch.

## Tests

The send-path test fails on the old code with `expected [ '--input-format', …(9) ] to include '--my-flag'`. Alongside it: the whitespace split, absent/non-string meta (a pane predating the key), argv not mutated, and that `--fork-session` is never reintroduced.

1705 passing in `app/view/agent`, `tsc --noEmit` clean.

### DIFF
--- a/.changesets/1788496373-fix-agent-provider-flags-were-dropped-from-cmd-args-on-every-zk5t.md
+++ b/.changesets/1788496373-fix-agent-provider-flags-were-dropped-from-cmd-args-on-every-zk5t.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): provider_flags were dropped from cmd:args on every send
--- a/frontend/app/view/agent/agent-model.ts
+++ b/frontend/app/view/agent/agent-model.ts
@@ -25,7 +25,7 @@ import { parseSeedZoom } from "./agent-zoom-seed";
 import { resolveForkSessionArgs } from "./fork-session-args";
 import { HISTORY_TAB_FOR_META_KEY, openOrFocusHistoryTab } from "./open-history-tab";
 import { quickForkAgent } from "./quick-fork";
-import { isPersistentLaunch, selectLaunchArgs } from "./launch-args";
+import { isPersistentLaunch, PROVIDER_FLAGS_META_KEY, selectLaunchArgs } from "./launch-args";
 
 export class AgentViewModel implements ViewModel {
     viewType = "agent";
@@ -688,6 +688,13 @@ export class AgentViewModel implements ViewModel {
                 // expression, so the controller/args choice and the mode this
                 // block advertises can never disagree.
                 agentMode,
+                // Persisted so the per-turn cmd:args rebuild can reapply it.
+                // That rebuild starts from the provider catalog, which knows
+                // nothing about what the launch path appended, so without this
+                // the user's flags are dropped on the first send and never come
+                // back (#2872). `--fork-session` is deliberately NOT persisted
+                // here — it is a one-shot launch intent, not a durable arg.
+                [PROVIDER_FLAGS_META_KEY]: agent.provider_flags ?? "",
                 ...(overrides?.containerImage || agent.container_image ? { "agent:container_image": overrides?.containerImage || agent.container_image } : {}),
                 controller: isPersistent ? "persistent" : "subprocess",
                 cmd: cliBin,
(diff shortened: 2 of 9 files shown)


a5af:
@codex review

PR #2966, opened 2026-09-04
Camper@claudius: Persist system-tool install success, show live version, collapse install log

GenericAgentX-asaf:
## Summary
- Fixes a real bug: after successfully installing a system tool (git/node/npm/python) from the AgentPrereqModal, the row reverted to looking like nothing happened. The `onInstalled` handler collapsed the panel and immediately re-probed, but srv's own PATH is captured at process startup, so a freshly-installed binary routinely still probes as "missing" right after a genuinely successful install. The parent modal now tracks installed-pending-restart state itself and shows a persistent "installed, restart AgentMux to use it" row instead of reverting.
- Backend now queries the live version a package manager would actually install (`winget show`, `brew info --json=v2`, `apt-cache policy`) and returns it as `resolvedVersion` — never hardcoded (see #2942 for why a hardcoded version string is a real staleness risk this repo already hit once).
- The install toggle/button now reads "Install v\<version\> now" once resolved, instead of a static "or install it now".
- Replaced the static hourglass emoji with the existing `fa-spinner fa-spin` pattern (already used in recording-section.tsx/swarm-view.tsx) plus a lightweight indeterminate progress bar during install.
- Moved the raw streamed install log behind a collapsed `<details>/<summary>Details</summary>` so the default view stays clean; the status header stays always visible.

## Test plan
- [x] `cargo build --release -p agentmux-srv` (via `task build:backend`) — clean, no errors from `system_install_handlers.rs`
- [x] `npx vitest run frontend/app/view/agent/components/AgentPrereqModal.test.tsx frontend/app/view/toolchain/SystemToolInstallInline.test.tsx` — 11/11 passing, including new coverage for the version-labeled button, the `resolvedInfo` pass-through prop, and the persisted post-install success state
- [ ] Manual verification in a running app (not done by the agent — see PR discussion)

### DIFF
--- a/.changesets/1788496460-fix-toolchain-persist-system-tool-install-success-across-sta-59u0.md
+++ b/.changesets/1788496460-fix-toolchain-persist-system-tool-install-success-across-sta-59u0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(toolchain): persist system-tool install success across stale-PATH refresh, show live version and collapsible install details
--- a/agentmux-srv/src/server/system_install_handlers.rs
+++ b/agentmux-srv/src/server/system_install_handlers.rs
@@ -72,6 +72,16 @@ struct SystemInstallStep {
     /// before the caller's own immediate re-probe shows "still not
     /// found" — Codex P1, PR #2790.
     verify_bin: String,
+    /// Version this exact command would install, queried live from the
+    /// package manager's own catalog at resolve time — never hardcoded.
+    /// A hardcoded "v24" here would repeat the exact staleness bug this
+    /// repo already hit once (Taskfile.yml's VERSION var pinned to a
+    /// specific `node@20` brew formula that outlived that major version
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in 95617ad3a — both reagent and Codex caught the same real bug: `installedPendingRestart` was local state inside `AgentPrereqModalPanel`, which `ModalLayer.replace()` remounts on every post-install refresh, wiping it right when it mattered.

Moved the set up into `AgentPicker.tsx`'s `openPrereqModal` closure (survives the recursive `replace()` calls in its own refresh loop) and threaded it through `AgentPrereqRequest` as `installedPendingRestart`/`onToolInstalled`, so the panel only ever reads it from props instead of owning it. Added a test that simulates the actual remount (full unmount + fresh mount with updated props) rather than only rendering the panel in isolation.

@codex review

PR #2967, opened 2026-09-04
Camper@claudius: Show a reassurance message in the splash screen when host startup runs long

GenericAgentX-asaf:
## Summary
Follow-up to the earlier splash-width fix (#2961) and the first-launch-latency findings on issue #2940.

Windows Defender's on-execution cloud reputation check for newly-extracted binaries can hold the splash screen open 18-36s on a genuinely fresh machine, with nothing telling the user anything is happening during that gap — confirmed against the Defender operational event log (Event ID 2010) on a real VM, both on the original 36s first-launch investigation and again just now on a fresh portable build.

- Adds a launcher-side 5s timer (well above a normal ~1s host-registration time) that posts a "First run can take longer" sub-row under the existing "Host startup" splash stage, reusing the splash's existing `SubBegin`/live-elapsed-counter rendering — no new rendering code.
- Deliberately does **not** claim "Windows Defender is scanning" specifically — that can't actually be known from the launcher (no event-log/process-state access in this code path), so a generic, honest message is used instead of a guess that also wouldn't generalize to macOS's own first-run Gatekeeper delay.
- Bumps `SUB_LABEL_MAX_CHARS` 14 → 28 to fit the message at full length without truncation, safe under the `SPLASH_W` widening already shipped in #2961 — only raises the truncation threshold for every sub-row, doesn't change layout.
- Fire-and-forget by design: if the host registers before the 5s threshold, the message is sent into either an already-torn-down channel or one whose consumer stopped draining — both harmless no-ops per `StartupEventSink`'s existing doc comment, never a visible glitch on a normal launch.

**Honest disclosure, same as #2961:** this has not been visually verified on a real Windows machine — the native `WS_EX_TOOLWINDOW` splash has no title text and is excluded from window-enumeration tooling I have access to. The release build compiles clean; the actual rendered result is unconfirmed.

## Test plan
- [x] `cargo build --release -p agentmux-launcher` — clean, no errors
- [ ] Visual confirmation on a real Windows machine (not done by the agent)

### DIFF
--- a/.changesets/1788536076-feat-launcher-show-a-reassurance-message-in-the-splash-scree-qyah.md
+++ b/.changesets/1788536076-feat-launcher-show-a-reassurance-message-in-the-splash-scree-qyah.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(launcher): show a reassurance message in the splash screen when host startup runs long (first-run Defender scan, issue #2940)
--- a/agentmux-launcher/src/host_pipe/mod.rs
+++ b/agentmux-launcher/src/host_pipe/mod.rs
@@ -611,6 +611,19 @@ impl HostPipe {
         self.inner.lock().await.writer.is_some()
     }
 
+    /// Production counterpart of `is_connected` above (which is test-only):
+    /// has a CEF host ever called `set_writer` — i.e. registered as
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2968, opened 2026-09-04
fix(startup): port the Linux paint gate to Windows so the window never shows blank

Agent5-asaf:
Fixes a live user report: the window loads blank for a long time behind the splash screen, and the splash says things are done well before the window is actually usable.

Full diagnosis: `docs/reports/REPORT_SPLASH_TO_FIRST_PAINT_BLANK_WINDOW_GAP_2026_09_03.md`.

## Root cause

The window was shown at CEF's `on_load_end` ("main-frame HTML finished loading") rather than at real first paint, on every platform **except Linux**. `on_load_end` can fire before anything has visually painted — the code's own doc comments already said so — and a July investigation measured that gap at ~2.08s on a comparable machine.

**Live-verified on this exact fix**, not just compiled: a fresh cold start logged

```json
{"message":"[startup-paint] frontend reported first paint","label":"main"}
{"message":"[startup-paint] revealing gated window","label":"main","reason":"signal","elapsed_ms":1177}
```

`elapsed_ms=1177` between `on_load_end` and the real double-rAF paint signal — over a full second the old code would have shown as a blank window. `reason="signal"` confirms the real signal won the race, not the 4s safety timeout.

## What changed

`reveal_top_level_window`'s existing Linux-only paint gate (`reveal_gated_window`, `handle_first_paint_signal`, `PaintGateRevealTask`, `FirstPaintSignalTask`, `PAINT_GATE_NEXT_EPOCH`) is now `#[cfg(any(linux, windows))]` — **Windows shares the exact state machine, not a parallel reimplementation.** Only the two calling-convention pieces differ per platform, both already flagged as the seam in that function's own doc comment:

- The `AGENTMUX_SPLASH_EVENT` `SetEvent` dismiss signal moved from an unconditional call at `on_load_end` into `reveal_gated_window`, firing only once the gate actually resolves.
- The `4000ms` safety-timeout constant is reused verbatim from Linux rather than picking a separate number — it's a backstop against a stalled renderer, not a target, and the live run confirms the real signal wins in practice.

### A latent bug fixed as a side effect of the move

The old unconditional `SetEvent` call had **no `is_pool_window` guard** (that check happens later in `on_load_end`), so a hidden pool-window prewarm racing ahead of the real main window during a cold start could dismiss the splash before the real window had even loaded — the same class of bug Linux's ready-file placement was already immune to, for the identical reason its own comment cites. Verified live: pool-window first-paint signals (`floating-pool-...`, `window-pool-...`) logged with **no** matching `revealing gated window` line, confirming they still skip the reveal path entirely.

## Splash telemetry (defect B)

Adds one new stage, **`paint`** — begun where `on_load_end` used to just call `window.show()`, ended where the gate resolves — so the gap this fix closes is visible in the splash's own timeline instead of the splash declaring "done" before it actually was.

This **diverges from the report's original two-change proposal** (a separate `frontend` stage plus extending `host` to span to first-paint): implementing the gate first made a cleaner shape visible, since both of those would have mostly duplicated one span already available at the same two call sites. No launcher-side changes needed — stage rows are fully key-driven, the same way `dlopen`/`cef_init` already prove. The report's §5.2 explains the reasoning in full.

## Explicitly out of scope

- **macOS is untouched.** The cfg gate stays `any(linux, windows)`, not widened further. Its applicability is read from the code (identical fallback path) but not separately measured — matching the original investigation's stated scope and the "don't touch what you can't verify" discipline this repo already established for the reverse case.
- **The frontend bootstrap / tab-reveal settle window is deliberately not counted** in splash telemetry — the in-page brain overlay already covers it visually; folding it into the native splash's lifetime would be an architectural change, not a telemetry one. See report §5.3.
- **The `paint` row's on-screen rendering wasn't visually confirmed** — the live run used `task dev`, which never connects to the launcher IPC socket the splash reads from, so those calls fired as designed no-ops there. The underlying reveal mechanism they report on **is** confirmed live; a packaged-build re-run to watch the row draw is the natural follow-up. (A packaged-build smoke test in this pass hit a single-instance collision with another already-running same-day build on this shared, multi-agent dev machine, and wasn't pursued further to avoid touching another agent's running instance.)
- **Windows-specific safety-timeout tuning wasn't done** — `4000ms` is reused from Linux, deliberately, per the reasoning above. If it's ever observed firing on real Windows hardware outside a genuinely crashed renderer, that's the signal a dedicated measurement pass is overdue.

## Testing

`cargo test -p agentmux-cef -p agentmux-launcher` — **521 passed, 0 failed** (308 + 213). Neither suite had prior unit coverage of this OS-coupled logic to extend on any platform — it's tightly bound to real CEF/Win32 state, the same reason Linux's original version shipped with none either. Verification here is the live cold-start run documented above, the same standard the original fix shipped under.

### DIFF
--- a/.changesets/1788540062-fix-startup-port-the-linux-paint-gate-to-windows-so-the-wind-8qrl.md
+++ b/.changesets/1788540062-fix-startup-port-the-linux-paint-gate-to-windows-so-the-wind-8qrl.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(startup): port the Linux paint gate to Windows so the window never shows blank, and report the gap in splash telemetry
--- a/agentmux-cef/src/client/navigation.rs
+++ b/agentmux-cef/src/client/navigation.rs
@@ -9,12 +9,14 @@ use cef::*;
 
 use super::AgentMuxHandler;
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "windows"))]
 use std::sync::Arc;
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "windows"))]
 use crate::state::AppState;
 
-/// Linux startup white-flash fix (docs/specs/REPORT_NEW_WINDOW_STARTUP_COLOR_FLASH_2026_07_14.md).
+/// Startup white-flash fix
+/// (docs/specs/REPORT_NEW_WINDOW_STARTUP_COLOR_FLASH_2026_07_14.md,
+/// docs/reports/REPORT_SPLASH_TO_FIRST_PAINT_BLANK_WINDOW_GAP_2026_09_03.md).
 ///
 /// Bound on how long the real window + native splash can stay hidden/up
 /// waiting for the frontend's first-paint signal before we show anyway.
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

Agent5-asaf:
Fixed both findings from the ReAgent re-review (pullrequestreview-5115776262) and Codex's matching line comments — thanks, these were real.

**P1 (navigation.rs:149, reveal_gated_window's fresh lookup)**: confirmed. The splash-dismiss signals (ready-file write, Windows `SetEvent`) fired unconditionally before the fallible `state.get_browser`/`browser_view_get_for_browser`/`bv.window()` lookup below them, and unlike the immediate `on_load_end` path, this delayed reveal (fired from the first-paint signal or the 4s safety timeout) had no retry if that lookup transiently failed. Fixed in `17aa11e` by splitting `reveal_gated_window` into `try_show_resolved_window` (the raw resolve+show) and `finish_gated_reveal`, which now retries up to `SHOW_WINDOW_MAX_RETRIES` times via a new `GatedRevealRetryTask` (mirroring the existing `ShowWindowRetryTask`), and only fires the paint-stage telemetry + dismiss signals once the window is confirmed shown — or retries are genuinely exhausted, in which case it still fires them (with an error logged) so `run_splash`'s untimed wait can't hang.

**P2 (navigation.rs:115, stage-end vs SetEvent ordering)**: confirmed — `report_startup_stage_end` only enqueues an async command drained over the launcher IPC pipe, so the synchronous `SetEvent` can win the race and a single non-blocking `try_recv()` drain in `run_splash` could miss a trailing stage-end, freezing the splash summary with that row shown as still in progress. Fixed in `splash.rs`: the post-dismiss drain now polls briefly (up to 150ms, 10ms steps) instead of once, but only while some `StageRow` is still open (`done.is_none()`) — the common case (everything already resolved) exits immediately, same as before, so this adds no latency on the normal path.

Also already fixed in the prior commit (`c1e8fa4`): the `label: None` fallback in `reveal_top_level_window` now also fires the Windows dismiss signal via a shared `signal_windows_splash_dismiss()` helper, closing the earlier P1 about that path never dismissing the splash.

Verification: `cargo check`/`cargo clippy` clean on both `agentmux-cef` and `agentmux-launcher` (no new warnings in either edited file — diffed against pre-change warning output to confirm). Full test suite: 521/521 passing (308 agentmux-cef + 213 agentmux-launcher).

PR #2969, opened 2026-09-04
feat(agent): auto-unblock on external auth bind, one-click Bind account

a5af:
## Summary

Two related fixes to the agent-pane login flow, from a live usage report — see `docs/specs/SPEC_AGENT_LOGIN_FLOW_TIGHTENING_2026_09_04.md` for the full analysis and design.

**1. Auto-unblock on an external bind.** If an agent pane is stuck behind a login prompt and the user binds an account for that provider elsewhere (Armory's Bind-to-Agent menu, the per-agent Identity tab, or this PR's new button), the pane now notices via the existing per-agent `agentidentities:changed:<agentId>` event, re-verifies with `CheckCliAuth`, and clears the block — without ever auto-retrying a turn (a bind event is not proof the credential works, and auto-resending an old message is the exact bug class `PLAN_LOGIN_CTA_SURFACE_CONSOLIDATION_2026_09_02.md` already paid down once).

**2. One-click "Bind account".** The auth failure row's "Armory → Accounts" action is replaced by "Bind: `<name>`" (one candidate) / "Bind account" (2+, opens a picker) whenever an already-authenticated account for this agent's provider exists in this channel — reusing the Armory's own `bindAccountToAgent` link+live-apply sequence directly, not reimplemented. "Login via terminal" is unaffected either way (kept unconditionally — it's the only working recovery for providers whose in-app OAuth can't complete in-app).

Explicitly NOT the removed "Use existing login" (`ANALYSIS_PER_CHANNEL_AUTH_BYPASSES_2026_08_31.md`) — that copied the user's personal `~/.claude` file across channel boundaries. This binds to a proper per-channel Armory account someone already authenticated in this channel; candidates are sourced from the channel-safe `id_store`-backed listing, not the flagged cross-channel-leaking one.

No backend changes — everything reuses existing RPCs/events (`LinkAgentIdentityCommand`, `CheckCliAuthCommand`, `agentidentities:changed`).

## Test plan

- [x] 44 new/updated unit tests (candidate-set filter, bind/Armory row-swap logic) — all passing
- [x] Full frontend suite: 3616 tests / 242 files, 0 failures
- [ ] Live: bind an account from the Armory while a different pane sits on the login prompt for that provider — confirm it clears without switching back to it
- [ ] Live: "Bind: `<name>`" / "Bind account" appears and works end-to-end when a candidate account exists; "Login via terminal" + "Armory → Accounts" still work unchanged when none exists

### DIFF
--- a/.changesets/1788556302-feat-agent-auto-unblock-on-external-auth-bind-one-click-bind-qsh9.md
+++ b/.changesets/1788556302-feat-agent-auto-unblock-on-external-auth-bind-one-click-bind-qsh9.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(agent): auto-unblock on external auth bind, one-click Bind account
--- a/docs/specs/SPEC_AGENT_LOGIN_FLOW_TIGHTENING_2026_09_04.md
+++ b/docs/specs/SPEC_AGENT_LOGIN_FLOW_TIGHTENING_2026_09_04.md
@@ -1,0 +1,311 @@
+# SPEC — Tighten the agent-pane login flow: auto-unblock on external bind, "Bind account" button
+
+**Date:** 2026-09-04
+**Status:** proposed — analysis complete, not yet implemented.
+**Repo:** agentmuxai/agentmux
+**Trigger:** Operator report, two related pieces of friction in the same flow:
+1. Starting an agent shows the login-required prompt. The user switches to the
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Reopening under the correct agent identity (GenericAgentX-asaf) with the agent@host title prefix — see the new PR.

PR #2970, opened 2026-09-04
fix(agent-pane): close the Worked/Queued-message desync; log the attached-task axis

Agent5-asaf:
## Summary

Investigation + first fixes from a refactor/cleanup pass on the pane's Working/Worked indicator, per `docs/specs/SPEC_AGENT_WORKING_STATE_UNIFICATION_2026_09_04.md`. Two user-reported issues:

1. **"Queued message" visible while the pane already reads "Worked".** `PendingMessagesPanel` gates purely on `state.pending`; a pending entry is only cleared by an async backend ack (`agent-message-accepted`), but the label flips to "Worked" the instant `turnPhase` reaches `Done` — synchronously, well before that ack lands. Nothing coupled the two, so the panel could show stale "Queued — sends at the agent's next step" copy for the entire ack round trip.

   Fixed at the reducer level: `TurnEnd` now marks any still-`enqueuedWhileBusy` pending entries `flushing: true` in the **same atomic transition** that flips `turnPhase` to `Done`. The panel renders honest "Sending — reaching the agent any moment" copy for flushing entries instead of claiming they're still waiting on a next step. The entry still only disappears once the real ack lands (`PendingMessageAccepted`/`Rejected`) — only the copy changes, not the timing.

2. **"Worked" showing while a long-running process is clearly still active.** Investigated in depth (see the spec doc for the full account). Short version: the `attachedTask`/`registryAttachedTaskSince` data layer and the dock's backgrounded-Bash wiring (`isAcceptedBackgroundLaunch`, `<task-notification>` completion tracking) are already fully built — more complete than the specs I initially found suggested (caught two independently-stale "still open" status lines in the process, corrected in the spec doc rather than left to mislead the next reader). PR #2489's footer render of this axis was deliberately reverted in PR #2495 as redundant with the dock row's own affordance; I found no live evidence this pass to justify re-litigating that call a third time, especially per the project's own "stop and ask before anything needing live-human UI testing" rule.

   Instead of guessing again: the attached-task axis had **zero** diagnostic logging of its own — every other axis this file tracks (`turnPhase` transitions, watchdog reasoning) gets a `[wave-turn]` line via the already-shipped `SPEC_AGENT_TURN_PHASE_TIMELINE_LOGGING_2026_08_18.md`, this one never did. Added it, so the next report of this can be diagnosed from real `muxlog phases` evidence instead of another guess.

## Verification

- `tsc --noEmit` clean.
- Full frontend suite: 3618/3618 passing across 242 files, zero regressions.
- 4 new reducer tests (atomic `flushing` transition), 5 new `PendingMessagesPanel` component tests (copy-switching behavior), 4 new `[wave-turn]` attached-task logging tests.
- `check-doc-status.sh` / `check-spec-citations.sh` both pass on the new spec doc.


a5af:
@codex review

Agent5-asaf:
Fixed the P2 finding from Codex (`3937939764`) — confirmed real. `flushHeldMessages` (`useAgentCommands.ts:1433-1455`) can genuinely bail without draining `heldQueue` when a controller refresh is still deferred, and the original version marked `flushing` unconditionally on `TurnEnd`, well before that. Fixed in `8372b34`: replaced the `TurnEnd`-based marking with a new `PendingMessageFlushStarted` command dispatched right before the actual `deliverToBackend` call — the moment delivery genuinely starts, not when the preceding turn merely ended. `TurnEnd` no longer touches `pending` at all.

On eventual delivery failure, `deliverToBackend`'s own catch block dispatches `PendingMessageRejected`, which removes the entry outright regardless of `flushing` — so there's no path to a stale "Sending" state.

Verification: reducer tests rewritten for the corrected design (4 tests — flush-started marking, id-scoping, unknown-id no-op, idempotency, plus a regression test confirming `TurnEnd` no longer touches `pending`). The full pre-existing `useAgentCommands.test.ts` suite (53 tests covering exactly this deferred-refresh/auth-ordering territory) still passes unchanged — confirms the one-line additive dispatch doesn't disturb its assertions. Full frontend suite: 3619/3619 passing, `tsc --noEmit` clean. Also fixed the doc-status CI failure from the prior push (stale `docs/specs/INDEX.md` after adding the new spec doc — regenerated via `gen-docs-index.sh`).

Agent5-asaf:
Fixed the P2 from ReAgent's re-review (`pullrequestreview-5117962165`) — `state.ts:173`'s doc comment still described `flushing` as set by the `TurnEnd` reducer arm, a leftover from before the previous commit moved that logic to `PendingMessageFlushStarted` (dispatched from `flushHeldMessages` at actual delivery start). Fixed in `24f7c8b`, along with the same stale reference in `PendingMessagesPanel.tsx`'s comment. Comment-only change — verified with the full touched-area test suite (328 tests) + `tsc --noEmit`, both clean.

PR #2972, opened 2026-09-04
fix(agent): show hover-to-peek time/token panel on expanded tool calls

Agent3-asaf:
## Summary
`ToolBlock.tsx` was the only one of the three document-node kinds (thinking/message blocks, user messages, tool calls) sharing `PeekOverlay`/`useNodePeek()` that fully suppressed its hover panel whenever the row was expanded (pinned, running, pending-approval, or held-open post-completion). The stated rationale — "command and time are visible in context there" — doesn't hold: the expanded panel body (`ToolBlockOverlay`) shows the command/output, but never a timestamp or token count. So a tool call's time/token info became completely unavailable for as long as its panel stayed open, while thinking text and user messages never lost that surface regardless of their own expand state.

## Fix
Only the redundant command line (`cmdText`) now suppresses on expand; the overlay itself — and its time/token-estimate lines — show regardless of `expanded()`, matching `MarkdownBlock.tsx`/`UserMessageBlock.tsx`'s existing (correct) behavior.

## Report
`docs/reports/REPORT_TOOL_CALL_PEEK_SUPPRESSED_WHEN_EXPANDED_2026_09_04.md` — full investigation, including a side-by-side of all three components' `show=` conditions confirming this was the one outlier.

## Test plan
- [x] `npx vitest run` — ToolBlock/MarkdownBlock/UserMessageBlock/PeekOverlay suites (73 tests) pass; 3 existing tests updated to assert the corrected behavior (overlay + time/estimate show while expanded, only the command line stays gated)
- [ ] Manual verification with real mouse input — not yet confirmed by a human

### DIFF
--- a/.changesets/1788563253-fix-agent-show-the-hover-to-peek-time-token-panel-on-expande-0ezs.md
+++ b/.changesets/1788563253-fix-agent-show-the-hover-to-peek-time-token-panel-on-expande-0ezs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): show the hover-to-peek time/token panel on expanded tool calls too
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #2973, opened 2026-09-04
fix(term): round-robin priority-lane egress across panes to stop cross-pane input delay

Agent2-asaf:
## Summary

- Typing into one pane feels delayed while a different pane is producing heavy output. Root cause: every pane on a connection shares a single FIFO WebSocket egress channel (`EventBus`'s priority lane) — a noisy pane's backlog can queue ahead of a different pane's own keystroke-echo frame indefinitely, and under sustained flood the lane can fill and silently drop the quiet pane's frame.
- Fix: when draining the priority lane, pull everything currently queued and round-robin it by pane (`fair_drain_priority` / `priority_pane_key` in `agentmux-srv/src/server/websocket.rs`) instead of forwarding in raw arrival order. Same-pane ordering is preserved; only cross-pane interleaving changes. No protocol change, no new WS message type.
- Full investigation: `docs/analysis/ANALYSIS_CROSS_PANE_INPUT_DELAY_UNDER_OUTPUT_LOAD_2026_09_04.md`.

## Validation

- TDD: 6 new unit tests written first (confirmed failing to compile before the fix existed), all green after. Existing `eventbus` tests unaffected.
- `cargo clippy` clean on the new code.
- Live-validated against a real running dev build with a new benchmark (`tools/tests/bench-term-cross-pane.mjs`): two terminal panes, pane A sustained a genuine ~54 KB/s / ~387 events/sec flood, pane B's keystroke-echo latency measured concurrently — stayed flat (p50 ≈9ms, p95 ≈11-14ms), no cross-pane regression.
- A direct unpatched-vs-patched side-by-side run was attempted for a concrete "before" number but got stuck behind build contention on a shared machine and was abandoned; the analysis doc's code-level reasoning for the prior regression (mirrors the precedent bug fixed in `0f34704a8`) stands in its place.

## Test plan

- [x] `cargo test -p agentmux-srv websocket::tests`
- [x] `cargo test -p agentmux-srv eventbus::tests` (no regression)
- [x] `cargo clippy -p agentmux-srv --bin agentmux-srv`
- [x] Live 2-pane benchmark against a real dev build under sustained flood
- [ ] Reviewer: sanity-check round-robin fairness doesn't reorder bytes within a single pane (see `fair_drain_priority`'s doc comment / ordering guarantee)

### DIFF
--- a/.changesets/1788554426-fix-term-round-robin-priority-lane-egress-across-panes-to-st-mlld.md
+++ b/.changesets/1788554426-fix-term-round-robin-priority-lane-egress-across-panes-to-st-mlld.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): round-robin priority-lane egress across panes to stop cross-pane input delay
--- a/agentmux-srv/src/server/websocket.rs
+++ b/agentmux-srv/src/server/websocket.rs
@@ -201,9 +201,21 @@ async fn handle_ws_connection(mut socket: WebSocket, state: AppState) {
             //   2. Direct broadcasts (e.g., SetMeta's obj:update) — raw
             //      { eventtype: "waveobj:update", oref: "block:xxx", data: ... }
             // This carries terminal echo output and all interactive events.
+            //
+            // Fairness: this lane is shared by EVERY pane on the connection, not
+            // just one. A single FIFO here means a noisy pane (heavy PTY output)
+            // can queue an unbounded number of frames ahead of a different pane's
+            // own keystroke echo, which arrives in the same channel. Draining all
+            // immediately-available events and round-robining them by pane (see
+            // `fair_drain_priority`) bounds that delay to one "round" per distinct
+            // active pane instead of the length of the noisy pane's backlog. See
+            // docs/analysis/ANALYSIS_CROSS_PANE_INPUT_DELAY_UNDER_OUTPUT_LOAD_2026_09_04.md.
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

Agent2-asaf:
Addressed all three findings in eeff67bd3:

1. **Head-of-line blocking on the read path** (Codex + ReAgent, independently flagged) — forwarding the whole fairly-ordered batch inside one `select!` branch body starved `socket.recv()` for the batch's duration. Fixed by trickling one event per `'ws` loop iteration via a `pending_priority` queue, so `socket.recv()` is re-polled between every single send — same cadence as before this PR.
2. **Object-update ordering** (Codex) — `priority_pane_key` no longer keys `waveobj:update`/`waveobj:batchedupdates` by scope/oref at all; fairness reordering is now scoped to `blockfile` (terminal output) events only, so object-model events always stay in pure FIFO arrival order relative to each other. Added `test_fair_drain_priority_preserves_object_update_order`, which reproduces the exact `[update1, update2, batch-delete]` scenario from the review comment.
3. **Non-portable flood command** (Codex) — `bench-term-cross-pane.mjs` now branches the flood generator on `process.platform` (`yes` on POSIX, a PowerShell loop on Windows) instead of assuming PowerShell unconditionally.

All three are covered by new/updated unit tests (`cargo test -p agentmux-srv server::websocket::tests`, 7/7 passing).

PR #2974, opened 2026-09-04
fix(agent-pane): skip the blank gap between Thinking… and tool output

Agent5-asaf:
## Summary

Follow-up to PR #2948 (`fix(agent-pane): hide bashwrap's internal starting-chunk from the tool log`). User report: after that fix, the tool-call panel still visually goes "Thinking… → blank → output" instead of a continuous flow — #2948 only fixed *what* renders once the tool-call panel's `ChunkList` branch is chosen, not *which branch gets chosen*.

**Root cause:** `ToolOverlayLog.tsx`'s `hasChunks()` counted the raw `log.chunks.length`, before `ChunkList`'s own `dropBashwrapStartingChunk()` filter. A Bash tool call with only its always-first `[bashwrap] starting: N chars` chunk received still satisfied `isStreaming() && hasChunks()`, routing to `ChunkList` — which then rendered nothing (its one chunk was filtered) instead of falling through to `ToolOverlayResult`'s own "nothing to show yet" branch. Combined with `ToolBlock.tsx`'s `autoExpanded()` (expands the panel the instant `status === "running"`), the visible sequence was: "Thinking…" (Working row) → panel auto-expands into an empty box → real output replaces it.

**Fix:** one line — `hasChunks()` now applies the same `dropBashwrapStartingChunk` filter before counting. The fallback branch it now correctly reaches renders `"⏳ Thinking..."` — the literal same text the Working row already shows — so the visible transition becomes "Thinking…" → "⏳ Thinking..." (tool panel, same words) → real output, with no distinguishable blank state.

Scoped to `ToolOverlayLog.tsx` only — `PersistentShellBlock.tsx` has the identical filter but only expands on manual pin, never automatically on tool start, so it can't produce this sequence.

Full root-cause writeup: `docs/analysis/ANALYSIS_TOOL_OVERLAY_BASHWRAP_BLANK_GAP_2026_09_04.md` (also notes this is a distinct, narrower bug from the still-open broader "scroll oscillation" investigation in `docs/analysis/ANALYSIS_TOOL_CALL_SCROLL_OSCILLATION_2026_08_17.md` and its follow-ups — not addressed here).

## Verification

- New test specifically distinguishes "correctly blank" from "wrongly blank instead of the Thinking placeholder" — the pre-existing bashwrap tests don't (both the bug and the fix produce zero `.agent-tool-log-line` rows).
- Falsified by reverting the fix in isolation and confirming exactly that one new test fails (11/12 pass); restored, diff clean.
- Full frontend suite: 3620/3620 passing across 242 files.
- `tsc --noEmit` clean.
- `check-doc-status.sh` / `check-spec-citations.sh` both pass on the new doc.

### DIFF
--- a/.changesets/1788563593-fix-agent-pane-skip-the-blank-gap-between-thinking-and-tool--lmw4.md
+++ b/.changesets/1788563593-fix-agent-pane-skip-the-blank-gap-between-thinking-and-tool--lmw4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): skip the blank gap between Thinking… and tool output
--- a/docs/analysis/ANALYSIS_TOOL_OVERLAY_BASHWRAP_BLANK_GAP_2026_09_04.md
+++ b/docs/analysis/ANALYSIS_TOOL_OVERLAY_BASHWRAP_BLANK_GAP_2026_09_04.md
@@ -1,0 +1,117 @@
+# Analysis: "Thinking… → blank → output" gap survived the bashwrap-hiding fix
+
+**Date:** 2026-09-04
+**Status:** implemented
+**Author:** Agent5
+
+## User's request (verbatim, for traceability)
+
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2975, opened 2026-09-05
fix(agent): revert tab strip + button to bare glyph, keep tooltip

Agent3-asaf:
## Summary
- #2944 added a visible "New Agent" text label beside the agent pane's `+` button (`PaneTabStrip`'s opt-in `addLabel` prop). Per request, reverted the visible label back to a bare `+`.
- The hover tooltip is preserved, and now shows **instantly**: switched from native `title` (OS-controlled delay, already documented elsewhere in this file as "slow/inconsistent in CEF") to the same Portal-based `Tooltip` component the per-tab labels already use, with `delayMs={0}`.
- `PaneTabStrip`'s `addLabel` prop itself is left in place (shared, tested, opt-in, no other call site) — this reverts agent-view's usage of it, not the underlying component feature.

## Test plan
- [x] `npx vitest run frontend/app/element/PaneTabStrip.test.tsx` — 20 tests pass (1 new: confirms the tooltip shows well under the component's 300ms default; 1 updated: "pinned last" check now accounts for the Tooltip's wrapper div)

### DIFF
--- a/.changesets/1788581353-fix-agent-revert-the-tab-strip-s-button-to-a-bare-glyph-keep-72wy.md
+++ b/.changesets/1788581353-fix-agent-revert-the-tab-strip-s-button-to-a-bare-glyph-keep-72wy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): revert the tab strip's + button to a bare glyph, keep the tooltip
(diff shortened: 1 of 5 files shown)


a5af:
@codex review

PR #2976, opened 2026-09-05
feat(swarm): larger select checkboxes, select-all, hide Stop during broadcast, fix groups dropdown overflow

Agent2-asaf:
## Summary

Four Swarm fleet-toolbar UX tweaks requested against `FleetToolbar`/`swarm-view`:

1. **Larger select checkboxes** — `.swarm-agent-select-checkbox` 13px → 18px, matching spacer updated too.
2. **"Select all" at the top** — new toggle in the fleet toolbar (`Select all` / `Select none`), reachable whenever at least one agent is listed, not gated on an existing selection (the toolbar itself now shows whenever there's at least one selectable agent, a live selection, or a saved group).
3. **Cancel vs. Stop conflict** — the broadcast composer already had a Cancel button; the "Stop N" button is now hidden while the composer is open so the two don't compete for the same row.
4. **Groups dropdown fix (best-effort)** — the dropdown was anchored `left: 0` with `min-width: 200px`. Swarm is typically docked as a narrow column and `.swarm-view` has `overflow: hidden`, so a left-anchored 200px flyout from a button positioned anywhere past a narrow pane's width can render past the visible edge and get invisibly clipped — which looks exactly like "clicking it does nothing." Re-anchored to `right: 0` plus a `max-width` safety net.

## Caveat on item 4

I audited the backend (`fleet.group.*` RPCs, `db_agent_groups` storage) and frontend wiring end-to-end and found no functional bug — registration, serialization, and the store layer all check out and are covered by existing tests. The overflow-clipping theory above is the most plausible concrete bug I could find by code review, but **I could not reproduce or verify it live** — cross-pane UI automation isn't available to this agent (my UI tools only reach my own pane, not the Swarm pane), and the two isolated dev-instance rebuild attempts in the previous PR's review cycle showed how slow that path is on this shared build machine, so I didn't retry it here. Please confirm this actually fixes what you saw — if the dropdown still doesn't work, it'd help to know exactly what "doesn't work" means (doesn't open at all? opens but empty? opens but clicking a group does nothing?) so I can dig further with a more targeted fix.

## Test plan

- [x] `npx vitest run frontend/app/view/swarm` — 133/133 passing, including 5 new tests in `swarm-fleet-toolbar.render.test.tsx` covering select-all reachability/toggle and the Stop-button hide-while-broadcasting behavior
- [x] `npx stylelint frontend/app/view/swarm/swarm-view.scss` — no new violations (18 pre-existing, unrelated)
- [ ] Manual: confirm checkbox size feels right, select-all works, Stop hides during broadcast, and groups dropdown is now visible/functional

### DIFF
--- a/.changesets/1788581665-feat-swarm-larger-select-checkboxes-select-all-hide-stop-dur-wutc.md
+++ b/.changesets/1788581665-feat-swarm-larger-select-checkboxes-select-all-hide-stop-dur-wutc.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(swarm): larger select checkboxes, select-all, hide Stop during broadcast, fix groups dropdown overflow
--- a/frontend/app/view/swarm/swarm-fleet-toolbar.render.test.tsx
+++ b/frontend/app/view/swarm/swarm-fleet-toolbar.render.test.tsx
@@ -1,0 +1,104 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * "Select all" reachability and the broadcast/stop button conflict, pinned
+ * through the actual `FleetToolbar` component.
+ */
+
+import { cleanup, fireEvent, render } from "@solidjs/testing-library";
+import { createSignal } from "solid-js";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("@/app/element/confirm-modal", () => ({
+    ConfirmModal: () => null,
+}));
+
+import { FleetToolbar } from "./swarm-fleet-toolbar";
+import type { FleetGroup, SwarmViewModel } from "./swarm-model";
+
+afterEach(() => cleanup());
+
+function modelStub(initialSelected: string[] = []) {
+    const [selected, setSelected] = createSignal<Set<string>>(new Set(initialSelected));
+    const [groups] = createSignal<FleetGroup[]>([]);
+    const [inFlight] = createSignal(false);
+
+    const model = {
+        selectedBlockIdsAtom: selected,
+        fleetGroupsAtom: groups,
+        fleetActionInFlightAtom: inFlight,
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

Agent2-asaf:
Good catch — you're right that `right: 0` just relocated the failure mode rather than fixing it. Fixed properly in cf67834ab: the groups dropdown now portals to `document.body` and is positioned via `computeMenuPosition` (the same floating-ui-backed primitive `flyoutmenu.tsx`/`popover-menu.tsx` already use, per `scripts/check-menu-positioning.sh`), kept live via floating-ui's `autoUpdate`. Since it's portaled entirely out of `.swarm-view`, it can't be clipped by that pane's `overflow: hidden` regardless of where the trigger button lands after wrapping — no fixed CSS anchor side needed. Also picked up click-outside/Escape-to-close, which the hand-rolled version never had.

Covered by a new test (`FleetToolbar — groups dropdown > opens without throwing`) that exercises the portal + async positioning path.

Agent2-asaf:
Good catch, and a subtle one — you're right that calling \`usePaneOverlay\` in \`FleetToolbar\`'s own body meant its \`onMount\` only ever ran once, for the pane's whole lifetime, before the dropdown ever existed. Fixed in a1b282b13 by extracting the dropdown into its own \`GroupsDropdown\` component, so Solid's \`<Show>\` genuinely unmounts/remounts it (re-running \`onMount\`, and with it the pane-overlay registration) on every open — matching \`flyoutmenu.tsx\`'s own documented reason for doing the same split.

Agent2-asaf:
Right again — missed carrying over \`avoidNativePanes: false\` when I copied the pattern. Fixed in 17203782e. Since this menu already punches a clip hole via \`usePaneOverlay\`/\`data-pane-overlay\`, it should open in place at its anchor rather than being pushed away from a Browser pane elsewhere in the layout — same reasoning \`flyoutmenu.tsx\` documents for the identical combination.

Issue #2977, opened 2026-09-05
Agent5: tracking: optional cross-platform system tray + persistent background service


Agent5-asaf:
## Overview

We want AgentMux to optionally behave like an OS-resident "always available" AI companion (the user's own reference point: Grok's desktop bot) — a system tray/menu-bar icon across Windows, macOS, and Linux, with the backend able to keep doing agent work even when no main window is open, plus a simplified panel reachable from the tray icon.

This issue is the **work tracker** for that feature. The design doc is `docs/specs/SPEC_TRAY_OPTIONAL_BACKGROUND_SERVICE_2026_09_04.md` — **read it before starting any workstream below**, it has the actual architecture research (current process model, crate choices, per-platform auto-start mechanisms, security defaults) this issue only summarizes.

**This is genuinely new surface** — confirmed via full-repo search that no tray icon, no app-self-install, and no auto-start mechanism exists anywhere in the codebase today. This is not a "wire up an existing thing" issue.

---

## How to use this issue

- Each workstream is independently pickable except where "Depends on" says otherwise.
- Tick the box and edit this issue with what you learned when you finish one — especially anything that turned out false. Same discipline as #2939.
- Don't mark a box done on "it compiles" — each item's acceptance criterion names an observable result.

---

## Status (2026-09-05)

Workstreams 0-4 have all been implemented at least in part; **7 PRs merged**
(#2983, #2987, #2989, #2992, #2996, #2999, #3002) with #3001 open.

**Three things are deliberately NOT done, and each blocks a box below:**

1. **Windows code signing is absent** — a hard prerequisite for shipping
   auto-start (Smart App Control blocks unsigned auto-starting apps). See WS2's
   last box.
2. **macOS auto-start uses a hand-written LaunchAgent, not `SMAppService`** —
   a documented deviation from the design doc with a user-visible consequence
   (Login Items). See WS2's macOS box.
3. **Nothing visual has been verified.** Every acceptance criterion phrased as
   "renders", "responds to clicks", or "verified via tasklist/ps" was written
   for a human with a display; this work was done in a headless agent
   environment. The logic behind each is unit-tested, but no icon, panel, or
   window has been observed. Treat all of WS1/WS3 as code-complete-but-unseen.

**2026-09-06 (AgentO):** macOS menu-bar backend in PR #3037 (open); the §7.5.1 splash-disabled gap below is addressed in the same PR. Linux is the only platform still without a backend.

[...]

PR #2978, opened 2026-09-05
docs: propose an optional cross-platform tray + persistent background service

Agent5-asaf:
## Summary

Design proposal for making AgentMux optionally behave like an OS-resident companion app (user's reference point: Grok's desktop bot) — a system tray/menu-bar icon across Windows/macOS/Linux, a backend that can keep working with no window open, and a simplified tray-triggered panel. Docs-only, nothing implemented.

Grounded in two research passes:
- **Codebase**: confirmed no tray/install/auto-start surface exists anywhere today; `srv` already survives a `host` *crash* but not a clean exit (that's the one load-bearing behavior change this whole feature needs); the window-pool hide/show/promote mechanism and the existing single-instance pipe-forward-and-reattach path are directly reusable; a known, still-open liveness gap in the reattach path (`docs/retro/retro-fresh-vm-suspend-orphaned-frontend-2026-09-03.md`) becomes a routine path this feature would hit constantly rather than a rare edge case.
- **External best practices**: `tray-icon`+`muda` for Windows/macOS, `ksni` directly for Linux (not `tray-icon`'s GTK path — a second toolkit's event loop alongside CEF's own, which multiple sources describe as fragile); Raycast's ephemeral-tray/on-demand-window/persistent-backend three-tier architecture as the closest analog; current (2025-2026) per-platform auto-start mechanisms and their real constraints (Scheduled Task not Windows Service — Session-0 isolation disqualifies a Service; `SMAppService` not legacy `launchctl`; XDG autostart plus optional `systemd --user`); security defaults informed by the Zoom 2019 / Cluely 2025 / Microsoft Recall 2024 cautionary precedents.

Full doc: `docs/specs/SPEC_TRAY_OPTIONAL_BACKGROUND_SERVICE_2026_09_04.md`
Tracking issue: agentmuxai/agentmux#2977

## Verification

- `check-doc-status.sh` / `check-spec-citations.sh` both pass.
- `docs/specs/INDEX.md` regenerated.


a5af:
@codex review

Agent5-asaf:
Both fixed in `b518579` — thanks, both were real:

1. **Changeset type**: `minor` → `patch`. Correct — this is docs-only, nothing implemented; `release.sh`'s highest-pending-type selection would have bumped the product version for a proposal.
2. **Prototype-target mismatch**: confirmed — §2 places tray ownership in `agentmux-launcher` (so the icon survives `host` crashes/restarts), but §7.2's prototype named CEF's event loop, which runs in `host`, not `launcher`. Fixed §7.2 to name the actual coexistence risk (launcher's own event loop), and applied the same fix to the tracking issue (#2977)'s matching workstream text.

PR #2980, opened 2026-09-05
fix(swarm): re-resolve and re-point subagent-watcher config dir after identity rebind

Agent5-asaf:
## Summary

Self-investigation, per user request ("why are your agent tool calls not showing in the swarm?"): my own subagent dispatches never appeared in AgentMux's Swarm view, for the entire lifetime of my agent pane. Confirmed the underlying dock/transcript data was fine (`muxspect dock` showed the ToolNodes correctly) — the backend simply never watched the right directory.

**Root cause**: `handle_reactive_register` resolves an agent's Claude config directory exactly **once**, at pane-registration time, via `watch_agent`. For an agent whose identity binding isn't committed to the DB yet at that exact moment (a real race against the recently-shipped "auto-unblock on external auth, one-click Bind account" flow), `resolve_bound_oauth_config_dir` returns `None` and the fallback lands on a stale `cmd:env` snapshot captured at launch time — not the agent's real, currently-live identity-bound directory. `watch_agent`'s own dedup logic means nothing ever re-triggers this resolution afterward: an already-watched `agent_id` just adds to `parent_block_ids` and returns immediately. The watcher is stuck on the wrong directory, permanently, for the rest of the pane's life — confirmed directly against my own live srv log (the register+watch pair fires exactly once across my instance's entire history, despite the underlying CLI process restarting several times since).

**Fix**: extracted `watch_agent`'s filesystem-watch setup into a `start_watch` helper (pure relocation), added `recheck_config_dir` (tears down and re-points a stale watch, preserving all dependent blocks, backfilling each one's own persisted session) and `recheck_all_watched_agents` (re-resolves every currently-watched agent — cheap, since it only runs on the rare identity-bind path, never per-turn). Wired into both production call sites that write an `agent_identity_link` row, right after the write succeeds, via the existing `subagent_watcher::global()` singleton — no `AppState` plumbing changes needed.

Full root-cause writeup: `docs/reports/REPORT_SWARM_SUBAGENT_WATCHER_ONE_SHOT_RESOLUTION_RACE_2026_09_04.md`

## Verification

- 5 new tests: no-op-when-unwatched, no-op-when-unchanged, repoint replaces the entry, repoint preserves multiple dependent blocks, repoint backfills a missed session.
- Falsified by neutering `recheck_config_dir` to an early return and confirming exactly the 3 behavioral tests fail (the 2 no-op tests correctly still pass); restored, diff clean.
- Full `subagent_watcher` suite: 92/92 (87 pre-existing + 5 new), confirming the `start_watch` extraction is a pure relocation.
- Full `agentmux-srv` suite: 3014/3014 passing, zero regressions.
- `cargo clippy -p agentmux-srv`: no new warnings at any touched line.
- `check-doc-status.sh` / `check-spec-citations.sh` pass (also fixed one pre-existing dangling citation in `mod.rs` — `REPORT_SWARM_SUBAGENT_HISTORY_FLOOD_2026_07_07.md` had moved to `docs/specs/archive/`).

## What this does not fix

- The pre-existing "events carry the first registrant's block id" limitation when multiple blocks share one `agent_id` (already documented as rare/out-of-scope in `watch_agent`'s own comment).
- My own subagent dispatches from before this fix shipped remain unrecovered for this exact session — the fix corrects the watch going forward from the next identity-bind event, not historical dispatches from a session that already ended without ever triggering one.
- Why my instance ended up with a blank `identity_id` in the first place (logged as `"Legacy row or UI regression?"`) — a separate, narrower launch-flow/DB-consistency question, out of scope here.


a5af:
@codex review

Agent5-asaf:
Fixed all three in `e43b2de` — all confirmed real:

**P1 (registration-before-insertion race)**: confirmed — `handle_reactive_register`'s own resolve-then-install window wasn't covered, and since `watch_agent` only ever runs once per pane, a bind landing in that exact window had no later trigger. Fixed by re-resolving and repointing immediately after `handle_reactive_register`'s own `watch_agent` call succeeds.

This took two attempts, worth being upfront about: my first fix added this self-check *inside* `watch_agent` itself (post-insert, using stores now held on `SubagentWatcher`). Running the full pre-existing test suite immediately caught that this was wrong — `watch_agent` has another caller (`server/service/misc.rs`'s legacy manual RPC, which deliberately passes `parent_block_id: ""` for panes with nothing to scope to) whose `config_dir` isn't backed by a resolvable instance/binding row at all. The self-check's blind re-resolution fell through `resolve_claude_config_dir`'s last-resort `derive_claude_config_dir` fallback and silently repointed the watch away from that caller's explicit choice, breaking `live_fs_event_with_empty_block_id_bypasses_the_ownership_check` and `live_fs_event_is_not_misattributed_to_a_block_that_does_not_own_the_session`. Relocated the self-check to `handle_reactive_register` instead — the one call site that actually derives `config_dir` from identity/binding resolution in the first place, so a fresh re-resolve there can only ever mean "the DB state changed," never "this caller never wanted identity resolution applied here."

**P2 (keep the old watch until replacement succeeds)**: confirmed — split watch setup into `build_watch` (pure construction, zero side effect on `watched_agents`) + `start_watch`/`spawn_consumer_loop`, so `recheck_config_dir` now builds the replacement *first* and only swaps the old entry out on success. New test reproduces the exact failure mode (a config dir with no resolvable ancestor) and confirms the old watch survives.

**P3 (preserve the original primary block)**: confirmed — added an explicit `primary_block_id` field to `WatchedAgent`, set once at first registration and preserved verbatim across every repoint, never re-derived from the unordered `parent_block_ids` set. New test registers 6 dependent blocks and confirms the *original* one survives a repoint.

Verification: 7 new tests total (up from 5) — the two new ones for P2/P3, both falsified against a live reproduction of their respective bugs before being confirmed correct, same discipline as the original 5. Full `subagent_watcher` suite: 94/94. Full `agentmux-srv` suite: 3016/3016, zero regressions. `cargo clippy`: no new warnings.

Agent5-asaf:
Fixed both in `a302431` — thanks for catching this, it's a legitimate, slightly embarrassing repeat of the exact bug the P1 fix's first attempt had already been caught and corrected for, one layer up.

**P1 (round 2)**: confirmed — `recheck_all_watched_agents()` runs on every `agent_identity_link` write anywhere in the app with no scoping, and it iterated every `WatchedAgent` regardless of whether that entry's `config_dir` was ever derived from identity resolution at all. The legacy `subagent.WatchAgent` RPC's `parent_block_id: ""` entries are exactly the shape `watch_agent`'s own (already-fixed) self-check had to be scoped away from — I just hadn't applied the same scoping to this sweep. Fixed by skipping any candidate with an empty `primary_block_id`, or one that doesn't resolve to a real `Block` — only ever recheck an agent whose config dir could legitimately have come from identity resolution in the first place.

**P2**: confirmed — `start_watch`'s doc comment described consuming its bool return for a post-insert self-check that no longer exists (leftover from the reverted first-attempt design). Simplified `start_watch` back to returning `()` since nothing uses the result anymore, rather than just patching the comment to describe an unused value.

Verification: 2 new tests (skip on empty block_id, skip on nonexistent block), both falsified against a live reproduction of the exact bug (temporarily disabled the new guard, confirmed both tests fail with precisely the wrong derived-fallback directory) before being confirmed correct. Full `subagent_watcher` suite: 96/96 (up from 94). Full `agentmux-srv` suite: 3018/3018, zero regressions. `cargo clippy`: no new warnings.

Issue #2979, opened 2026-09-05
Windows block/agent stop has no graceful-signal path, only hard TerminateJobObject


GenericAgentX-asaf:
## Summary

`BlockController::stop()` on Unix sends `SIGTERM` to the whole process group
(reaching the agent CLI's subtree, e.g. `claude`/`codex` and whatever they
spawn) and schedules `SIGKILL` after a `KILL_GRACE_SECS = 5` backstop —
`agentmux-srv/src/backend/blockcontroller/shell/lifecycle.rs:886-894`. That
entire block is `#[cfg(unix)]`-only. **On Windows, nothing analogous exists at
all.**

Killing a block on Windows goes through `JobObjectTracker::kill_tree()` →
`TerminateJobObject` (`agentmux-srv/src/backend/process_tracker/windows.rs:203-221`)
— an immediate, unconditional hard kill, zero grace period, every time. There
is no Windows equivalent of "ask nicely first" wired in anywhere (no
`GenerateConsoleCtrlEvent`/`CTRL_BREAK_EVENT`, no message to the CLI, nothing).

## Why this matters

This isn't scoped to app shutdown — it's the same path used for:
- closing a single pane
- closing a tab
- a workspace delete (`delete_workspace` saga)
- app-level quit

So on Windows, an agent CLI process never gets a chance to do anything on its
own way out — flush a partial response, persist its own session file, whatever
it might otherwise do on `SIGTERM` — regardless of *why* the block is being
torn down. On Unix, at least a well-behaved CLI gets a real 5-second window.

Note: this is not a data-corruption risk on its own — AgentMux's own storage is
SQLite in WAL mode (safe against a hard process kill, as opposed to a power
loss) and already has restart-time reconciliation for processes that died
uncleanly (`reconcile_stale_subagents`). The gap is specifically that the killed
process itself gets no graceful signal, not that AgentMux's own state gets
corrupted by it.

## Suggested fix shape

A Windows counterpart to the Unix SIGTERM/grace block, mirroring its shape:
send `CTRL_BREAK_EVENT` to the process's own console process group (requires
the child to have been created with `CREATE_NEW_PROCESS_GROUP`, which needs
checking against however the PTY/console is currently allocated for spawned
agent CLIs), wait the same grace window, then fall back to the existing
`TerminateJobObject` as the hard backstop — unchanged from today.

## References

[...]

PR #2981, opened 2026-09-05
Manoz@Area54: docs: propose a formal shutdown countdown modal + splash-style progress

GenericAgentX-asaf:
## Summary

Design proposal for closing a gap the repo owner flagged directly: AgentMux's
shutdown today "does a couple things" with zero user-visible explanation,
unlike startup, which has a splash screen narrating its own stages.

Proposes a launcher-owned, native modal (mirroring the startup splash
architecture, not a frontend/Solid modal) with:
- a 5-second countdown, "Close now" (skip it) / "Cancel" (abort the quit)
- once confirmed, the same modal switches to progress narration of the real
  teardown stages, reusing the `StartupEvent`/`spawn_splash` re-invocation
  mechanism already proven by crash-restart and PR #2967's reassurance timer

Grounded directly in the existing shutdown trace (`on_before_close`, the
pool-window close cascade, `backend_close_window`'s 2s timeout, srv's
`delete_workspace` saga, the 30s `TEARDOWN_GRACE` backstop and ~12s quit
watchdog) rather than inventing new numbers or a new mechanism.

Docs-only, nothing implemented.

## Cross-references

- PR #2978 (Agent5's tray/background-service proposal): flags and resolves a
  real interaction — that proposal's own premise requires "closing the last
  window" to stop meaning "quit," which is this spec's current trigger
  condition. §7 works out the split (window-close vs. an explicit "Quit
  AgentMux" action) so neither design blocks the other.
- Issue #2979 (opened alongside this PR): the same "no graceful stop on
  Windows" gap exists one level down and more generally — every block/agent-
  process stop, not just app shutdown, hard-kills via `TerminateJobObject`
  with zero grace on Windows. Tracked separately since it's independently
  actionable and out of this spec's scope.

## Verification

- `bash scripts/check-doc-status.sh docs/specs/SPEC_SHUTDOWN_COUNTDOWN_MODAL_2026_09_04.md` — ok
- `bash scripts/check-spec-citations.sh docs/specs/SPEC_SHUTDOWN_COUNTDOWN_MODAL_2026_09_04.md` — ok
- `docs/specs/INDEX.md` regenerated via `scripts/gen-docs-index.sh`

### DIFF
--- a/.changesets/1788586424-docs-propose-a-formal-shutdown-countdown-modal-splash-style--h74s.md
+++ b/.changesets/1788586424-docs-propose-a-formal-shutdown-countdown-modal-splash-style--h74s.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+docs: propose a formal shutdown countdown modal + splash-style progress narration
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #2984, opened 2026-09-05
Manoz@Area54: fix(agent-pane): hide the tab strip entirely on a fresh pane

GenericAgentX-asaf:
## Summary

A fresh agent pane — picker showing ("My Agents"), no agent launched yet — rendered the tab strip anyway. With a single conversation the strip collapses to just the "+" button, but since `SPEC_PANE_TAB_STRIP_TRAILING_BLUR_2026_08_12.md` it also spans the full pane width as a frosted-glass panel, so a fresh pane got a glass bar and a lone "+" floating over the top of the picker.

The strip is now gated on a `shouldShowTabStrip()` predicate:

| Pane state | Strip |
|---|---|
| Fresh — picker, no agent, lone tab | **hidden entirely** (new) |
| Lone tab, agent launched | shown (just the "+") |
| Lone read-only history tab | shown |
| 2+ tabs — including a blank 2nd tab opened via "+" | shown, with pills |

That last row is the case worth calling out: the "+" on a launched pane opens a second, still-blank picker tab with no `agentId` of its own. Keying purely off "does this tab have an agent" would hide the strip exactly there and strand the user on a blank pane with no way back to the first tab — so the predicate checks visible tab count first.

The strip is `position: absolute` and reserves no space (`agent-view.scss`), so hiding it shifts no layout — the picker just gets its top edge back.

## On the second half of the request (double-click to rename)

**Already implemented — no code change needed here.** `agent-view.tsx:572` wires `onTabDoubleClick` → `setRenamingBlockId`, rendering `PaneTabRenameInput` inline, and `handleTabRenameConfirm` renames the authoritative `AgentDefinition` with an optimistic local override plus rollback-on-failure. `PaneTabStrip.test.tsx` covers the dispatch ("fires onTabDoubleClick with the tab item").

It's just been hard to *reach*: a lone conversation renders no tab pill at all, so there's nothing to double-click until a 2nd tab exists. That's pre-existing behavior and unchanged by this PR.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run frontend/app/view/agent/tab-strip-visibility.test.ts frontend/app/element/PaneTabStrip.test.tsx` — 25/25 pass
- New `tab-strip-visibility.test.ts` covers all five states in the table above
- Extracted as a pure predicate rather than an inline JSX condition, matching how the rest of this directory is tested (`agent-view.tsx` has no render tests)

### DIFF
--- a/.changesets/1788623152-fix-agent-pane-hide-the-tab-strip-entirely-on-a-fresh-pane-s-cu3l.md
+++ b/.changesets/1788623152-fix-agent-pane-hide-the-tab-strip-entirely-on-a-fresh-pane-s-cu3l.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): hide the tab strip entirely on a fresh pane, so the picker isn't overlaid by a lone +
--- a/frontend/app/view/agent/agent-view.tsx
+++ b/frontend/app/view/agent/agent-view.tsx
@@ -119,6 +119,7 @@ import { lastLinkedAccountId } from "./providers/provider-id-aliases";
 import { buildStartupPayload, resolveAccounts } from "./startup/buildStartupPayload";
 import type { SignalPair } from "./state";
 import { createAgentAtoms } from "./state";
+import { shouldShowTabStrip } from "./tab-strip-visibility";
 import type { DocumentNode } from "./types";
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #2985, opened 2026-09-05
docs(retro): task dev killed 4x by Bash-tool SIGINT + mid-build worktree mutation

Agent2-asaf:
## Summary

`task dev` failed to launch four times during the previous session. I diagnosed it as external resource contention, told the user the build system was at fault, and gave up. **That was wrong** — running `cargo build --release -p agentmux-srv` directly succeeded first try (exit 0, 4m39s). Both real causes were self-inflicted:

1. **Launching via the Bash tool with `&`** instead of `mcp__agentmux__Shell`. A Bash-tool background job belongs to that tool call's process group, which is terminated when the call returns (seconds) — while `task dev` needs minutes. Signature: a literal `^C` in the log, a bare `error: could not compile` with **no `error[EXXXX]` diagnostic above it**, and go-task `exit status 58`. It fails *intermittently* (an orphaned child sometimes survives), which is what made one early success look like the method worked.
2. **Running `gh pr merge --delete-branch` while a build was live against the same worktree**, swapping tracked files out from under `cargo build` / `npm install`.

`task dev`'s own isolation/robustness work needed no changes — it worked correctly every time it was allowed to run to completion.

## The main finding isn't either bug

CLAUDE.md already had a section literally titled *"Launching `task dev` from an agent / MCP Shell (Windows)"* with the correct invocation, and two adjacent retros already existed. They weren't consulted. The retro records that plainly, along with the other process failures (adopting an unverified theory early, reading past the `^C` twice, running the obvious control experiment last, and handing the user a wrong root cause to work around).

## Changes

- **New retro:** `docs/retro/retro-task-dev-bash-tool-sigint-and-worktree-mutation-2026-09-05.md`
- **CLAUDE.md:**
  - States outright that the Bash tool *cannot* launch `task dev` (with the intermittency warning — one success doesn't validate the method)
  - Adds the `^C` + `exit:58` + *minutes* signature to the failed-shell diagnostic table, including the "bare `error:` with no `error[EXXXX]` = signalled, not broken" tell and the direct-`cargo build` control experiment
  - Adds the don't-mutate-the-worktree-during-a-build rule

## Verification

- `cargo build --release -p agentmux-srv` standalone → exit 0, 4m39s (proves the build is healthy)
- Bash tool + `&` → `^C`, exit 58, 3/3 reproductions, always at the point the tool call returned
- `mcp__agentmux__Shell` + `scripts\dev-agent.cmd` → survived multiple tool-call boundaries, **0 occurrences of `^C`** in an 84KB log, build completed, Vite ready on `localhost:5297`, host PID 64584 confirmed alive in `tasklist`
- `bash scripts/check-doc-status.sh` on the new retro → ok

Note: a dev instance from this verification is currently running against this worktree, so **don't merge this until the user is done testing** — that's the exact mutation cause #2 describes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -51,6 +51,24 @@ On Windows, `task dev` builds a production-parallel layout in `dist/cef-dev/` (l
 
 #### Launching `task dev` from an agent / MCP Shell (Windows)
 
+**Use `mcp__agentmux__Shell`. The Bash tool CANNOT launch `task dev`** — not
+with `&`, not with `run_in_background`, not with `nohup`/`disown`. A Bash-tool
+background job belongs to that tool call's process group, and the group is
+terminated when the call returns (seconds), while `task dev` needs minutes.
+`mcp__agentmux__Shell` runs server-side via `ShellNodeRunner` with an
+independent lifetime, which is the whole reason it exists.
+
+This fails **intermittently**, which is the trap: an orphaned child sometimes
+survives long enough to come up, so one success does not mean the method
+works. See the `^C` + `exit status 58` signature under "Diagnosing failed
+shells" below and
+`docs/retro/retro-task-dev-bash-tool-sigint-and-worktree-mutation-2026-09-05.md`.
+
+**Also: never `git checkout` / `git merge` / `gh pr merge` while a build is
+running against the same worktree.** Merging swaps tracked files out from under
+a live `cargo build` / `npm install` and corrupts it (same retro). Stop the
+build first, or use a separate `git worktree`.
+
 `task dev` requires `bash.exe` to be on the Windows PATH (go-task's Taskfile calls `bash -c '...'` for build steps via cmd.exe). The registry PATH has `Git\cmd` (shims) but not `Git\bin` (bash.exe). Two additional traps exist when launching from an agent's MCP Shell:
 
 - **Gap A:** MSYS2 bash won't resolve `.cmd` files from bare command names — `bash -c "task dev"` exits with "command not found".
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

Agent2-asaf:
Both P2s were correct — thanks, the first one in particular changed the conclusion of the retro rather than just its wording. Fixed in 29fad0558.

**P2 #1 — "the Bash tool cannot launch `task dev`" was wrong.** I verified your citations in the code rather than taking them on faith, and they hold: `hook.rs:95,107` threads `--declared-background`, `bash_wrap.rs:322-328` exempts it from the idle kill entirely (`Duration::from_secs(u64::MAX)`), and `bash_wrap.rs:330+` `setsid()`s into a brand-new session with no controlling terminal — with a comment saying that exists precisely so the wrapped command survives its invoker being signalled. Plus the recorded Windows success you pointed at.

The actual common factor across all six failures was an appended shell `&`, which I never removed — so I generalised from six runs of one uncontrolled variable and indicted the wrong component. `cmd &` returns instantly, so the `setsid()` protection lands on a launcher shell that exits in milliseconds and the real build gets orphaned. The rule is now **"never append `&`"**, with both working paths named (`mcp__agentmux__Shell`, or `run_in_background: true` with no `&`) and the mechanism explained. I've also marked that path as *asserted from code, not tested here*, so the next reader knows which claims are observed.

**P2 #2 — correct, a checkout/merge cannot inject `^C`.** I'd merged two distinct hazards into one diagnostic entry, which is exactly how a future interrupted build gets misdiagnosed. The signature is now scoped to a genuine signal source, with an explicit note that mid-build worktree mutation produces inconsistent-input errors and **no** `^C`. Stated separately in both CLAUDE.md and the retro.

The retro also picks up a "repetition is not isolation" lesson, since that's the underlying mistake both findings share.

Still blocked on the stale-branch P0 from ReAgent (v0.55.35 landed after I branched). Rebase is deliberately deferred — a dev instance is running against this worktree for the user to test in, and rebasing would be the exact mid-build mutation this PR documents. Will rebase and merge once they're done.

PR #2986, opened 2026-09-05
Loap #2@claudius: feat(window): Chrome-style drag-to-top maximize and border-drag vertical snap

GenericAgentX-asaf:
## Summary

Operator report: dragging the window to the top of the screen doesn't offer to maximize, and dragging a window border to the screen top/bottom doesn't vertically snap — both work in Chrome. Full analysis + design: `docs/specs/SPEC_WINDOW_SNAP_MAXIMIZE_2026_09_04.md`.

**Root cause (shared by both):** Windows' own Snap Assist never fires for AgentMux's main window. It's a CEF Views *frameless* window (`AgentMuxWindowDelegate::is_frameless`), which hands Win32 non-client behavior to Chromium's internal `HWNDMessageHandler` — code this repo doesn't own. The move gesture additionally can't use the native OS move loop at all (documented at `ui_tasks/drag.rs:314-323`: Chromium swallows `WM_NCLBUTTONDOWN` before `DefWindowProc` can start it), which is why a manual move loop already exists. So both gestures are re-implemented manually rather than chased via style flags.

**Border-drag vertical snap** — extends the *existing* `WM_SIZING` subclass (`window_edge_resize_wndproc`, already installed on every top-level window) to clamp the proposed rect to the monitor work area when a top/bottom drag lands near it. Dragging either horizontal border to its screen edge fills the window vertically (both edges), matching Windows/Chrome. Notable:
- The clamp runs **after** the passthrough to Chromium's WndProc — Chromium's own `WM_SIZING` handling can rewrite the rect (min-size/aspect constraints), so clamping first would let it silently discard the snap.
- Vertical axis only; width/x untouched by construction — that's what keeps it a vertical snap rather than an accidental maximize.
- Monitor resolved from the rect's **center**, not a corner: during a top-edge drag the top corner is the point crossing the screen edge, so a corner lookup would flip to the monitor above exactly when the snap should engage.
- The hook's "pure observer, every message always passes through" doc comment was rewritten rather than left stale — it still always passes through, but `WM_SIZING`'s rect is no longer untouched.

**Drag-to-top maximize** — the manual move loop now tracks a top-of-work-area **cursor** zone (not the window's own edge; the window follows the cursor at whatever grab offset, so its edge carries no intent), shows a translucent click-through preview overlay while in the zone, and maximizes on release. Preview teardown *and* the commit both sit at the single post-loop exit point every `break` funnels through, so the overlay can't outlive the drag on any path (release, Esc-then-release, `WM_QUIT`, stolen-capture safety net). Floaters excluded — their top-of-screen drag already means redock/tear-back-in.

`chrome.rs` gained a shared `toggle_maximize_hwnd` plus a separate **non-toggling** `maximize_hwnd`; the gesture uses the latter deliberately, since toggling would restore-*down* an already-maximized window being dragged — the opposite of what the gesture means.

All the actual decisions live in a new pure `client/window_snap.rs` — plain integer geometry, no Win32 types, deliberately not `cfg(windows)`-gated — so its tests run on every CI target instead of only `windows-latest`. Both native call sites do nothing but read the answer.

## Test plan

- [x] `cargo test -p agentmux-cef` — 322 passed / 0 failed (14 new: WMSZ edge mapping incl. corners + pinned against the existing tick-event map, snap-fill decision, opposite-edge non-arming, already-filled no-op, threshold boundary, negative/non-zero work-area origins, cursor-zone predicate)
- [x] `cargo build -p agentmux-cef` — clean, no new warnings
- [ ] **Live (Windows, `task dev`)** — native message-loop code, no harness can reach it:
  - Drag title bar to screen top → preview appears; release → maximizes; drag back down before releasing → no maximize, preview gone
  - Esc mid-drag in the zone → returns to pre-drag position, no maximize, preview cleared
  - Multi-monitor: snaps to the monitor actually under the cursor
  - Drag top border to screen top → fills vertically, **width and x unchanged**; same for bottom border
  - Corner drag → only the dragged axis clamps
  - Non-edge resizes → `windowresize:tick` pane redistribution unregressed
  - Floaters unaffected (top-drag still does redock, not maximize)

Deliberately left open (§3.3): whether floaters/subwindows want the same `WM_SIZING` clamp — floaters are `WS_POPUP` and don't currently get the hook at all; left to the live pass rather than guessed.

### DIFF
--- a/.changesets/1788623510-feat-window-chrome-style-drag-to-top-maximize-and-border-dra-r417.md
+++ b/.changesets/1788623510-feat-window-chrome-style-drag-to-top-maximize-and-border-dra-r417.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(window): Chrome-style drag-to-top maximize and border-drag vertical snap
--- a/agentmux-cef/src/client/mod.rs
+++ b/agentmux-cef/src/client/mod.rs
@@ -58,6 +58,10 @@ mod crash_recovery;
 mod recovery_pages;
 pub(crate) mod error_catalog;
 mod context_menu;
+// Deliberately NOT windows-gated even though both call sites are: it is pure
+// integer geometry with no Win32 types, so its tests run on every CI target
+// instead of only windows-latest. See the module's own doc comment.
+pub(crate) mod window_snap;
 #[cfg(target_os = "windows")]
 mod wndproc;
 #[cfg(target_os = "windows")]
--- a/agentmux-cef/src/client/window_snap.rs
+++ b/agentmux-cef/src/client/window_snap.rs
@@ -1,0 +1,248 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Pure geometry for Chrome-style window snapping —
+//! `docs/specs/SPEC_WINDOW_SNAP_MAXIMIZE_2026_09_04.md`.
+//!
+//! Deliberately **not** `#[cfg(target_os = "windows")]` and deliberately free
+//! of any Win32 type: the callers (`client::wndproc`'s `WM_SIZING` arm for
+//! border-drag vertical snap, `ui_tasks::drag`'s move loop for
+//! drag-to-top maximize) are both native message-loop code that no test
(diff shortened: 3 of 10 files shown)


a5af:
@codex review

PR #2988, opened 2026-09-05
Manoz@Area54: feat(layout): drop Armory from the default startup panes, put CPU above Swarm

GenericAgentX-asaf:
## Summary

The starter layout seeded on first launch (and for each new window) was four panes: agent on the left, then swarm / armory / sysinfo stacked down the right. It's now three — Armory is gone, and sysinfo (CPU) moved to the **top** of the right column:

```
┌────────────────┬──────────────┐
│                │   sysinfo    │  2 of 10 ≈ 20%
│     agent      ├──────────────┤
│    (tall)      │              │
│                │    swarm     │  8 of 10 ≈ 80%
└────────────────┴──────────────┘
     50% width        50% width
```

sysinfo keeps the same 20% share of the right column it already had, so swarm simply absorbs Armory's space. **Armory itself is unchanged** and still one click away in the widget bar — only the *starter* set changed.

## Notes for review

- **No migration needed.** This only affects newly-seeded tabs (first launch + newly-created windows). An existing persisted layout is untouched.
- **Both seeding paths were updated**, and they share the same tree builder so they can't drift: the pre-bootstrap store-direct first-launch seed (`wcore::seed_default_layout`) and the post-bootstrap reducer-routed `CreateWindow` seed (`window_create.rs`, which also had its `seeded_ids.len() == 4` guard updated — that would have silently skipped seeding otherwise).
- **Renamed** `*_four_pane_*` → `*_three_pane_*` across the three functions rather than leaving the arity wrong in the name.
- **One sharp edge, now documented in both places:** `default_three_pane_tree`'s parameter order is semantic (agent, swarm, sysinfo), *not* top-to-bottom display order, since sysinfo renders above swarm. `window_create` builds its block list positionally against that order.
- `obj.rs`'s `four_pane_json` serde fixture is deliberately left as-is — it exercises round-tripping a nested `LayoutNode`, not the current default; its comment now says so instead of claiming it mirrors wcore.

## Verification

- `cargo build -p agentmux-srv` — clean (only pre-existing warnings)
- `cargo test -p agentmux-srv --bins -- pane layout` — 205 passed, 0 failed
- Updated `seed_default_layout_creates_three_pane_layout` to assert 3 blocks and explicitly that armory is **not** seeded
- Added `default_three_pane_tree_puts_sysinfo_above_swarm`, asserting on the tree itself (child order + 2.0/8.0 sizes) — `leaforder` alone wouldn't catch the two being swapped in `children`
- Fixed `layout_seeders_route_through_reducer_coherently`, which had been registering a stale `b-armory` block and asserting 4 leaves

### DIFF
--- a/.changesets/1788624586-feat-layout-drop-armory-from-the-default-startup-panes-cpu-n-iktr.md
+++ b/.changesets/1788624586-feat-layout-drop-armory-from-the-default-startup-panes-cpu-n-iktr.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(layout): drop Armory from the default startup panes; CPU now sits above Swarm
--- a/agentmux-srv/src/backend/obj.rs
+++ b/agentmux-srv/src/backend/obj.rs
@@ -650,8 +650,8 @@ mod tests {
     /// the frontend produces today MUST yield the typed LayoutNode, and
     /// reserializing MUST produce equivalent JSON. The shapes here cover:
     ///   1. Single-leaf root (dnd / tear-off shape)
-    ///   2. The first-launch four-pane shape (deep nesting + mixed
-    ///      group/leaf nodes)
+    ///   2. A multi-pane shape (deep nesting + mixed group/leaf nodes) —
+    ///      the former first-launch four-pane layout, see below
     ///   3. Edge: missing `children` array for leaves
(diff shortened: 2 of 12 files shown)


a5af:
@codex review

PR #2990, opened 2026-09-05
Camper@claudius: fix(agent): bump stale fable pin, stop the picker advertising a model it doesn't select

GenericAgentX-asaf:
## Summary

Reported as **"why does the panel say Fable 5.1 and Fable 5?"** — the runtime dropup was rendering two different versions in a single row. Three distinct causes, all fixed.

### 1. The fable pin was stale (the real defect)

The curated row pinned `claude-fable-5`, which is now a **legacy** model — `claude-fable-5-1` is current.

This matters more for fable than it would for the others, because of a distinction that wasn't previously explicit anywhere:

| Row | Value | Behaviour |
|---|---|---|
| `opus` / `sonnet` / `haiku` | family **alias** | CLI resolves it at call time → always current, only the label can drift |
| `claude-fable-5` | **concrete id** | Does **not** self-resolve → a stale value actually selects the older model |

### 2. The overlay refreshed `label` but never `value`

`setProviderModels` folded the live catalog in by updating only the label. That's correct for alias rows — self-resolution is the whole point — but wrong for concrete ones: the row displayed **"Fable 5.1"** while still passing **`claude-fable-5`** to `--model`. Advertising one model, selecting another, with nothing in the UI showing the mismatch.

The overlay now refreshes `value` alongside `label` for concrete rows and leaves alias values untouched, so this **cannot silently recur** when the next version ships.

### 3. The visible contradiction: version numbers in `description`

The static `description` also named a version ("Claude Fable 5") and was never refreshed, so it disagreed with the refreshed label — that's literally the "5.1 over 5" that was reported. Descriptions are now **version-free**; the label is the single source of version truth. That removes the whole class rather than patching this instance.

## On the concrete id

`claude --help` documents `--model` as taking *"an alias for the latest model (e.g. 'sonnet' or 'opus') or a model's full name"* and lists no `fable` shorthand — so a full model name is the correct form here, not an invented alias.

## Tests

Adds `model-overlay.test.ts` (9 tests). The overlay had **no direct coverage**, which is why the alias/concrete distinction was never pinned down — the tests now assert both halves explicitly, plus newest-wins ordering, no-duplicate-family, unseen-family passthrough, empty-API no-op, `default` marker survival, and a guard that descriptions never contain digits.

## Test plan

- [x] `npx vitest run frontend` — **3658 tests, 247 files, all passing**
- [ ] Visual confirmation in the running app (not done by the agent)

## Related

Found during the audit that also produced `a5af/shared-infrastructure#458` (provider-reporter), whose drift engine independently flagged this same row as `pinned-stale` at `error` severity while correctly leaving the three alias rows alone.

### DIFF
--- a/.changesets/1788632478-fix-agent-bump-the-stale-fable-model-pin-and-stop-the-picker-q5f6.md
+++ b/.changesets/1788632478-fix-agent-bump-the-stale-fable-model-pin-and-stop-the-picker-q5f6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): bump the stale fable model pin and stop the picker advertising a model it doesn't select
--- a/frontend/app/view/agent/components/AgentCreateFromTemplateModal.tsx
+++ b/frontend/app/view/agent/components/AgentCreateFromTemplateModal.tsx
@@ -22,14 +22,14 @@
  * universal modal system per `feedback_use_universal_modal_system`.
  */
 
-import { createEffect, createMemo, createResource, createSignal, For, onCleanup, onMount, Show, type JSX } from "solid-js";
+import { createEffect, createMemo, createResource, createSignal, For, onCleanup, onMount, Show, untrack, type JSX } from "solid-js";
 
 import { Button } from "@/element/button";
 import { RpcApi } from "@/app/store/rpc-api";
 import { TabRpcClient } from "@/app/store/rpc-util";
 import { getCliCatalogEntry } from "../defaults/cli-catalog";
 import { PROVIDERS } from "../providers/catalog";
-import { getProvider } from "../providers";
+import { familyKey, getProvider } from "../providers";
 import { resolveEffectiveLaunchProvider } from "../agent-launch-env";
 import { providerSupportsModelFlag } from "../buildRuntimeArgs";
 import { isAvailable, watchCapability } from "@/app/store/toolchain-capabilities";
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in `a53331c5b`. Legitimate, and it's **a gap this PR's own change opened** — worth being explicit about.

Refreshing a concrete row's `value` means an option's value can now change underneath a user who has already picked it. That state didn't exist before: values were immutable, so `modelTouched` short-circuiting the re-pick effect was safe. Now a selection made before `providers.models` resolves leaves the list showing the refreshed row while `model()` still holds the old id — and submit sends `model()`. Same advertise-one/select-another gap this PR set out to close, just relocated from the picker to the create flow.

Verified before fixing: the effect returns early on `modelTouched` (`AgentCreateFromTemplateModal.tsx:147`) and submit reads the signal (`:281`), exactly as described.

The effect now still leaves a touched selection alone in the normal case, but if the selected value has vanished from the options **and** a same-family row replaced it, the selection migrates. Alias rows are unaffected — their values never change, which is the whole point of the alias/concrete split.

Two details worth noting:
- `familyKey` is now exported from `model-overlay` rather than reimplemented in the modal, so the grouping rule has a single definition.
- `model()` is read through `untrack` so the effect doesn't re-trigger on its own write.

247 files / 3659 tests green, `tsc --noEmit` clean.

GenericAgentX-asaf:

The prior approval was reset by `a53331c5b` (the Codex P2 follow-up you just verified). Requesting a fresh review state so this can merge.

GenericAgentX-asaf:
Fixed in `44fedf6b9`.

**P1 — persisted model id.** Correct, and it's the third place this PR's value-refresh could strand a selection: the picker's own option list (fixed in the original commit), the create-from-template modal's local signal (Codex, `a53331c5b`), and now block meta — which is the component the bug was originally reported against.

An agent configured before the bump holds `claude-fable-5`; once the overlay refreshes the row, `modelLabel()` falls back to the raw id and `build()` marks nothing `current`.

I migrated the **persisted** value rather than making the lookups family-tolerant. A display-only fix would have shown the new row as selected while meta still held the old id — the same advertise-one/select-another mismatch, one layer down. That's the trap this PR has now hit three times, so it seemed worth closing at the state layer rather than the render layer.

Worth noting: my first attempt at the test failed, and the failure was **the test's fault, not the fix's** — I'd written the block meta as dotted keys when `getRuntimeConfig` reads a nested object under `agent:runtime` (`buildRuntimeArgs.ts:153`). Corrected, and three cases now cover it: concrete id migrates, alias is left alone, unrecognised id is left alone rather than guessing a family.

**P2 ×2 — both correct.**
- The `row` helper was written and never used; removed.
- The shared-signal issue was a genuine defect in my tests, not just fragility: `setProviderModels` writes a module-level signal, so the "curated catalog" assertions were reading state left by earlier tests rather than the curated defaults they claim to check. Now re-imports the module per test via `vi.resetModules()`; verified order-independent with `--sequence.shuffle`.

247 files / 3663 tests green, `tsc --noEmit` clean.

GenericAgentX-asaf:

Thanks for verifying — but the review *state* is still `CHANGES_REQUESTED` from review 5122507959, which leaves the PR `BLOCKED`. Your confirmation came as a reply, which doesn't clear it. Requesting a formal pass so the state reflects the verified fixes in `44fedf6b9`.

PR #2991, opened 2026-09-05
docs(report): shared provider config — current state, agent read path, buildout

Agent2-asaf:
## Summary

Assessment of the **"Claude Code — shared provider config"** surface in Armory: what exists today, how a spawned agent actually consumes it, and what "building it out" would involve. Report only — **no code changes**.

## Headline findings

**1. It's thinner than the name suggests.** "Shared provider config" is currently *one hand-maintained file* — `~/.agentmux/shared/providers/claude/CLAUDE.md` — surfaced read-only under Armory → Memory → Global. No table, no write path, no versioning, no shared settings/hooks/model-defaults, and no non-Claude provider support (the seeding logic early-returns for any provider whose `auth_dir_name != "claude"`, explicitly flagged as uninvestigated).

**2. It's the one shared resource that isn't materialized per-agent.** Global Memory, skills and MCP servers are all *composed into* the agent's working directory at launch. The provider config isn't injected at all — AgentMux points `CLAUDE_CONFIG_DIR` at a directory and the Claude CLI reads the file itself. Legitimate design, but load-bearing: if the file is missing, the CLI falls through to the operator's personal `~/.claude/CLAUDE.md`.

**3. 🔴 Verified latent defect — the UI launch path skips isolation seeding.**
`agentmux-cef/src/commands/platform.rs:113-120` (`ensure_auth_dir`) only `create_dir_all`s. `seed_claude_md_placeholder_if_missing` has exactly **two** production callers — `agent_open.rs:360` and `identity/resolver/inject.rs:648` — and the UI path (`agent-model.ts:520` → `WriteAgentConfigCommand` → controller spawn) hits neither, confirmed by exhaustive grep and by tracing the launch flow.

Consequence: on a host with an unseeded shared dir **and** a personal `~/.claude/CLAUDE.md`, a UI-launched agent silently inherits the operator's global instructions — the exact failure `SPEC_ISOLATE_HOST_CLAUDE_MD_2026_08_31.md` exists to close, and which that spec's own doc comment records as having been observed live.

**Not firing on this host** — I checked: the shared dir *is* seeded (placeholder, mtime 2026-08-31) and there's no ambient file to inherit. Latent, not an active leak. Exposure window is fresh install / new provider dir / deleted `CLAUDE.md`, first launch via UI, operator has a personal `~/.claude/CLAUDE.md`. Once any App-API spawn happens the dir is seeded permanently, which likely explains why it's gone unnoticed.

Root cause is **two independently-maintained dir-provisioning paths**, not a missing call — so the recommended fix is to unify provisioning rather than add a second seeding call and preserve the divergence.

## Buildout proposal

Ranked, each step independently shippable. Recommends **splitting "shared provider settings" (new, properly modelled) from the CLAUDE.md preview (keep read-only)** rather than having AgentMux take ownership of a file whose documented purpose is "managed by Claude Code itself, not AgentMux" — that would pick an ownership fight with the vendor CLI. Step 1 (close the seeding divergence) is worth doing regardless of which model is chosen.

## Provenance

The report has an explicit §6 marking which claims I read and verified myself (the seeding gap, the call-site grep, the UI launch trace, `SPEC_SURFACE_CLAUDE_GLOBAL_CONFIG_2026_08_24.md` in full, live disk state) versus which are relayed from a codebase survey and unverified (specific line numbers in several frontend/handler files, the "no provider-config DB table" claim). Reviewers should weight those differently, and confirm line numbers before editing.

### DIFF
--- a/docs/reports/REPORT_SHARED_PROVIDER_CONFIG_STATE_AND_BUILDOUT_2026_09_05.md
+++ b/docs/reports/REPORT_SHARED_PROVIDER_CONFIG_STATE_AND_BUILDOUT_2026_09_05.md
@@ -1,0 +1,296 @@
+# Report: Shared provider config — current state, how agents read it, and buildout
+
+**Date:** 2026-09-05
+**Status:** Draft — assessment + proposal. Nothing here is implemented. Contains
+one verified latent defect (§3.1) that is independently actionable.
+**Author:** Agent2
+**Scope:** The "Claude Code — shared provider config" surface in Armory, the
+on-disk config it points at, and the path by which a spawned agent actually
+consumes it.
+
+---
+
+## 0. Short answer to "how are we doing"
+
+**Thinner than the name suggests.** "Shared provider config" today is *one
+hand-maintained file* — `~/.agentmux/shared/providers/claude/CLAUDE.md` —
+surfaced in Armory as a **read-only preview**. There is no table, no schema, no
+write path, no versioning, and no concept of shared provider *settings* beyond
+that single Markdown file.
+
+How agents read it is also different from every other shared Armory resource,
+and that asymmetry is the main structural finding:
+
+| Shared resource | How it reaches the agent |
+|---|---|
+| Global Memory (bundles) | **Composed** into `<work_dir>/CLAUDE.md` at launch |
+| Skills | **Copied** into `<work_dir>/.claude/skills/…` at launch |
+| MCP servers | **Composed** into `<work_dir>/.mcp.json` at launch |
+| **Shared provider config** | **Not injected at all** — the Claude CLI reads it itself, off disk, via `CLAUDE_CONFIG_DIR` |
+
+Everything else is materialized per-agent by AgentMux. The provider config is
+the one case where AgentMux points an env var at a directory and trusts the
+vendor CLI's own discovery. That's a legitimate design, but it has a
+consequence: **if the file isn't there, the CLI silently falls back to the
+operator's personal `~/.claude/CLAUDE.md`** — which is the exact failure
+`SPEC_ISOLATE_HOST_CLAUDE_MD_2026_08_31.md` exists to prevent. §3.1 documents a
+live code path where that seeding is skipped.
+
+---
+
+## 1. What exists today
+
+### 1.1 The artifact
+
+One file, no database:
+
+- **Path:** `DataPaths::provider_auth_dir("claude")/CLAUDE.md` →
+  `~/.agentmux/shared/providers/claude/CLAUDE.md`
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

Agent2-asaf:
Both correct, and the first one retracts this report's headline finding — thanks for catching it. Fixed in 393198099.

**P2 #1 — the "verified defect" was wrong for the normal flow.** You're right that my trace stopped at controller creation. I re-traced and confirmed your mechanism: `input.rs:337` calls `inject_identity_env_async` on the first turn, and for an agent with an `AgentInstance` + OAuth bindings, `inject.rs:617-667` blocks ambient-home dirs, seeds the placeholder fail-closed, and **overwrites `CLAUDE_CONFIG_DIR`** with the per-identity dir (`inject.rs:667`) before the CLI is spawned. Since the launch modal requires an identity at submit time, my "fresh install + UI launch" repro does not reproduce. Claim retracted.

**Re-scoped to what actually survives verification:** `ensure_auth_dir` still doesn't seed, and `inject_identity_env_with_broker` still early-returns with env untouched when a block has no `AgentInstance` row — `inject.rs:361-368`, whose own comment names the case as *"quick-launch panes that never went through the launch modal."* But I did **not** verify whether such a pane sets `CLAUDE_CONFIG_DIR` at all, and the two possible answers point in different directions (a leak, versus a known-and-accepted un-isolated route that seeding wouldn't fix). So §3.1 is now an open question with a decisive experiment attached, and buildout "Step 1 — do this regardless" is demoted to a conditional step behind a Step 0 that settles it. Shipping a provisioning refactor on the strength of the retracted claim would have been change without a demonstrated defect.

While re-checking I also picked up `REPORT_CLAUDE_CONFIG_DIR_ISOLATION_EVIDENCE_2026_09_01.md`, which *measures* the fallback mechanism directly (§2) — so the mechanism half is now on firmer ground than my original inference. Its §1 warning is folded into my §5 method: that report's own first pass ran arms 1–2 only, got a null from every arm, and would have concluded "no leak, fix unnecessary" — wrongly. Any follow-up needs the calibrated sentinel arm.

**P2 #2 — correct, one block not two.** The host-CLI block was removed 2026-09-01. I took the two-block description from the 08-24 spec chain (which predates the removal) and contradicted my own §3.5 in the process. Verified against `global-brain-manager.tsx:148-195` and corrected, with a note about why that spec chain — three self-corrections deep — shouldn't be read as current state without checking code.

The retraction is recorded in §6 rather than edited away silently: the survey I commissioned and my own reading were incomplete in the same direction, which is exactly why the claim looked solid from inside it.

PR #2993, opened 2026-09-05
Posa@Area54: fix(tabs): highlight the clicked tab immediately, not after the destination paints

GenericAgentX-asaf:
Switching to a large tab made the **tab bar itself** feel slow: the pill took a visible moment to look selected, and then the heavy pane appeared all at once. The ask was for the opposite ordering — switch now, paint after, with the pane's load time visually cordoned off rather than folded into the switch.

## Root cause — one signal, two consumers, no paint boundary

`activeTabId` (`store/window-identity.ts`) is **backend-authoritative**: it reads `ws.activetabid`, so it cannot move until `SetActiveTab` round-trips. Two unrelated pieces of UI read that same memo:

1. `tabbar.tsx`'s pill — `isActive={tabId === displayActiveTabId()}`, which fell straight through to raw `activeTabId()` on the select path.
2. `workspace.tsx`'s per-tab `display: none → flex` reveal of the destination.

Both land in **one synchronous Solid flush**. The browser then owes layout for a whole newly-displayed subtree before it can paint anything — including the pill. `SPEC_AGENT_PANE_TAB_SWITCH_PERF_2026_05_27.md` had already measured this:

> All five JS-side hypotheses fell. The 500ms long-task is **browser-side layout + paint** on `display:none → block`, opaque to JS-level perf hooks.

That cost is content-proportional, which is exactly why it's worse for large tabs. **The existing reveal gate cannot fix it** — `visibility: hidden` still participates in layout, so it hides the cascade's ugliness without shortening the frame it blocks. The pill sat computed-but-unpainted until that frame drained.

## The fix — a pattern this file already had

`displayActiveTabId`/`pendingHiddenTabIds` (`SPEC_TAB_CLOSE_BUTTON_SELECT_FLASH_2026_08_25.md` §8-9) is *already* an optimistic client-side override of `activeTabId()` — for the close flow, where the strip promotes the neighbor before `CloseTab` resolves. The plain click-to-select path never got one.

Now it has one: `pendingSelectedTabId` is written **before** the RPC is issued, so the pill commits and paints on its own cheap schedule.

`workspace.tsx` is **deliberately untouched** — the content reveal still reads the raw backend-authoritative atom, so only the highlight is optimistic. The destination keeps taking exactly as long as it takes, behind the existing gate. This PR does not try to make the reveal cascade faster; that stays `SPEC_AGENT_PANE_TAB_SWITCH_PERF_2026_05_27.md`'s job.

## Changes

| File | |
|---|---|
| `frontend/app/tab/active-tab-display.ts` | **new** — pure `resolveDisplayActiveTabId()`: optimistic select, then the pre-existing close-flow promotion verbatim |
| `frontend/app/tab/active-tab-display.test.ts` | **new** — 10 tests |
| `frontend/app/tab/tabbar.tsx` | `pendingSelectedTabId` signal; `handleSelect` writes before the RPC, clears in `finally` |
| `docs/specs/SPEC_TAB_SWITCH_DECOUPLE_SELECT_FROM_PAINT_2026_09_04.md` | **new** |

The decision moved into a pure module rather than staying inline so the precedence between the two optimistic overrides is assertable — same reasoning as `view/agent/failure/synthetic-row.ts`.

## Correctness details worth reviewing

- **Stale-click guard.** `handleSelect` clears the pending value only if it's still this call's own target (`cur === tabId ? null : cur`), so a newer click during an in-flight RPC isn't clobbered back to the stale committed id. Same stale-generation discipline as `tab-reveal.ts`'s gates.
- **Failure path.** The clear runs in `finally`, so a select whose RPC throws can't leave a phantom highlight on a tab the backend never activated — mirrors `handleClose`'s own `unhideTab`.
- **Guarded fallthrough.** A pending id that no longer exists, or is itself mid-close, falls through rather than highlighting a pill the strip doesn't render.
- **`handleClose` interaction.** It reuses `displayActiveTabId()` for gate targeting, so a pending select can now win there. Identical to old behaviour whenever no select is pending (the common case); when one is, the two candidates are already racing two concurrent RPCs, so neither is reliably "the" next active tab. Commented at the call site.

## Testing

**Mutation-checked, not just green:** with the optimistic-select branch removed, exactly the 2 fix-specific tests fail while all 8 close-flow tests still pass — confirming the close-flow tests guard *preserved* behaviour rather than the new behaviour.

- 52/52 across all four `frontend/app/tab/` suites (42 pre-existing + 10 new)
- `tsc --noEmit` clean
- Re-verified after rebasing onto latest `main`

**Verified live** by the repo owner in a `task dev` build — "much better".

**Not measured:** a before/after profile on a large tab. The mechanism is verified by construction (the pill no longer reads a value that requires the RPC), and the cost it sidesteps is the already-measured 500-600ms above, but I'm not claiming a profiled number.

## Scope note

This is the **window-level** tab bar. The in-pane block-stack tab strip (Agent History / Quick Fork / "+") has a *different* root cause — there the pill and the destination's full remount are driven off the same `localTreeStateAtom` write, and `NodeModel.blockId` immutability forces a real remount rather than a CSS reveal. Not addressed here; §2 of the spec records it as a separate follow-up.

### DIFF
--- a/.changesets/1788624955-fix-tabs-tab-bar-highlights-the-clicked-tab-immediately-inst-3twe.md
+++ b/.changesets/1788624955-fix-tabs-tab-bar-highlights-the-clicked-tab-immediately-inst-3twe.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tabs): tab bar highlights the clicked tab immediately instead of waiting on the destination pane's reveal
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -295,6 +295,7 @@ partial list.
 | [`SPEC_SYSTEM_TOOLCHAIN_INSTALLER_2026_08_24`](SPEC_SYSTEM_TOOLCHAIN_INSTALLER_2026_08_24.md) | SPEC: One-click system-toolchain installer (git, Node/npm, and friends) across Windows/macOS/Linux |
 | [`SPEC_TAB_COLOR_DESATURATION_2026_08_13`](SPEC_TAB_COLOR_DESATURATION_2026_08_13.md) | Spec: Desaturate tab colors, keep agent pane border colors as-is |
 | [`SPEC_TAB_CONTENT_REVEAL_GATE`](SPEC_TAB_CONTENT_REVEAL_GATE.md) | Tab content reveal gate |
+| [`SPEC_TAB_SWITCH_DECOUPLE_SELECT_FROM_PAINT_2026_09_04`](SPEC_TAB_SWITCH_DECOUPLE_SELECT_FROM_PAINT_2026_09_04.md) | Instant tab-bar selection, decoupled from destination-pane reveal cost (window-level tabs) |
 | [`SPEC_TERMINAL_LATENCY_BENCHMARK_2026_05_19`](SPEC_TERMINAL_LATENCY_BENCHMARK_2026_05_19.md) | SPEC: Terminal Input Echo-Latency Benchmark |
 | [`SPEC_TERMINAL_SCROLLBACK_PERSISTENCE_2026_07_23`](SPEC_TERMINAL_SCROLLBACK_PERSISTENCE_2026_07_23.md) | SPEC: Terminal scrollback doesn't survive reconnect (all `view:"term"` panes) |
 | [`SPEC_TERM_DOUBLE_RAF_TEAROUT_2026_05_30`](SPEC_TERM_DOUBLE_RAF_TEAROUT_2026_05_30.md) | SPEC: Remove the terminal Stage-1 RAF write-coalescer (double-rAF) |
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -178,7 +178,7 @@ are doing history". Everything else is here; the completeness
 assertion in the generator fails the build rather than emit a
 partial list.
 
-### implemented (133)
+### implemented (134)
 
 | Spec | Title |
 |---|---|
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both addressed in `48fda495c`.

**Codex P2 — confirmed real, and it had a second layer.** Fixing `handleSelect`'s guard alone is not sufficient: `setActiveTab` has its own `if (fromTabId === tabId) return` (`store/tab-actions.ts`), so re-issuing for `A` while committed is still `A` returns without an RPC — the in-flight `B` call would still land and win. The user's last click loses either way.

So both guards are addressed:
- `handleSelect` now compares against `displayActiveTabId()` rather than the committed `activeTabId()`.
- A switch is now a **loop** (`driveTabSelection`), not a single call: after each `setActive` resolves it re-reads the intent, and if a newer click arrived it goes again — by then committed has moved off it, so the next call is a real RPC rather than an early return. It terminates because each iteration either returns or observes a new intent, which only a fresh user click can produce.

6 new tests, **mutation-checked** against a naive single-call implementation: exactly the two mid-flight-click cases fail there — including your scenario verbatim (`click B`, then `click A` before B resolves, assert committed ends on `A` and both RPCs issued) — while the other 14 pass.

**CI failure was unrelated to the code:** `docs/specs/INDEX.md` was stale (`implemented 133 → 134`). Rebasing onto main picked up a commit adding an implemented spec; git merged my own count edit cleanly on top, leaving the file syntactically fine but semantically wrong — precisely what `gen-docs-index.sh --check` exists to catch. Regenerated. All six gates in that job now pass locally.

58/58 across `frontend/app/tab/`, `tsc` clean.

PR #2994, opened 2026-09-05
docs(spec): Armory Shared Provider Setup — provider-agnostic, versioned, dual-authored

Agent2-asaf:
## Summary

Design spec for the direction given: **Armory shouldn't mention Claude Code or any `.md` file.** One concept — *"shared provider setup"* — global, applies to all agents, with AgentMux resolving the right provider file behind the scenes, plus change management and writes from both the user and the agent.

Supersedes the buildout options in `REPORT_SHARED_PROVIDER_CONFIG_STATE_AND_BUILDOUT_2026_09_05.md` §4, which recommended keeping the surface read-only. **Design only — no implementation.**

## The good news: this is mostly composition, not invention

Three pieces already exist and the spec reuses rather than reinvents them:

- **Provider→filename resolution is already built.** `ProviderConfig::startup_instructions_filename` (`providers.rs:120`) maps every provider to `CLAUDE.md` / `AGENTS.md` / `GEMINI.md` / `QWEN.md` / `None`. The UI never names a file; the registry answers "which file." `None` is meaningful — those agents get nothing, and the spec requires disclosing that rather than silently no-op'ing.

- **Change management has a working precedent.** `db_agent_native_memory` + `db_agent_native_memory_versions` (`migrations.rs:760-798`) already models `content_hash` / `parent_version_id` / `source` / `source_detail` / `session_id` — i.e. *who wrote this, user or agent* is an existing column, not something to invent. Its history/diff/revert impls also carry reagent-P1 fixes (`app_api/mod.rs:988`, `:1151`) that a fresh implementation would rediscover the hard way. The spec says explicitly: don't design a second versioning scheme.

- **Ownership protection exists** via `CLAUDE_MD_MANAGED_MARKER`.

## The interaction worth reviewing carefully

Writing real content to `<provider_auth_dir>/<instructions_filename>` **composes correctly** with the isolation seeding — real content satisfies the same no-fallback invariant the placeholder does, and `seed_claude_md_placeholder_if_missing` already no-ops when a file exists (`providers.rs:697`).

But the spec calls out that the placeholder must **stay**: with shared setup empty, it's still the only thing between an agent and the operator's personal `~/.claude/CLAUDE.md` — a leak measured directly in `REPORT_CLAUDE_CONFIG_DIR_ISOLATION_EVIDENCE_2026_09_01.md` §2. Removing it as "now redundant" would be a regression.

## Operator decisions recorded

- **Any agent may write, fully audited** (no gating). Because auditability is then the *only* control on a global surface, the spec promotes the audit properties from nice-to-haves to requirements: unattributable writes rejected, history strictly append-only so an agent can't cover its tracks via the same API, and agent-authored versions visibly distinguished in the UI. Also states plainly that content is instructions read by a model, not executed — bounding this to influence, not arbitrary execution.
- **One shared body materialized everywhere.** Per-provider content deferred; schema is keyed per provider anyway so that stays a UI change rather than a migration.

## Left open deliberately

**Q3** materialize-on-write vs. at-launch (recommends both: write-through + launch-time reconcile so a hand-deleted file self-heals), and **Q4** the Global Memory overlap — both are global, both become instructions, differing only in destination. Flagged rather than resolved because the predecessor chain (`SPEC_SURFACE_CLAUDE_GLOBAL_CONFIG_2026_08_24.md`) needed three self-corrections largely from conflating exactly these two concepts.

### DIFF
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -434,7 +434,7 @@ partial list.
 | [`uptime-adaptive-width-help-zoom`](uptime-adaptive-width-help-zoom.md) | Spec: Adaptive Uptime Width + Help View Zoom |
 | [`widget-visibility-rearchitecture`](widget-visibility-rearchitecture.md) | Widget Visibility Re-Architecture |
 
-### draft (241)
+### draft (242)
 
 | Spec | Title |
 |---|---|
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

Agent2-asaf:
All four accepted and fixed in c6f585e40 (plus the stale `INDEX.md` that failed CI — adding a spec without regenerating the index).

**P1 #1 — you're right, and this was the serious one.** `startup_instructions_filename`'s contract says *"relative to the agent's working directory"*, and it carries an explicit *"set only where independently verified against the provider's own docs, not guessed"* discipline. I inferred a second, provider-home meaning for it — exactly what that discipline forbids — and OpenClaw's own entry flags `AGENTS.md` as an *unverified* workspace bootstrap. As specced, saves would have succeeded while non-Claude agents silently never received the setup.

Destinations are now an explicit verified mapping: **Kind A** (working-directory file, verified for all providers via the field's real contract) and **Kind B** (provider home — verified for **Claude only**, by measurement in `REPORT_CLAUDE_CONFIG_DIR_ISOLATION_EVIDENCE_2026_09_01.md` §2). Added a test asserting no Kind B destination exists for a non-Claude provider, so this can't be reintroduced quietly.

**P1 #2 — correct, and I should have caught it**, having documented `inject.rs:667`'s env overwrite in the companion report two hours earlier. Kind B is now reconciled into the resolved identity dir at first-turn injection, alongside the existing seed call.

**P2 #3 — correct and self-contradictory as written.** Per-provider keys with a provider-less API and a test plan demanding one version row. Now a singleton content/version stream with providers as materialization targets only; also documented that file writes are a best-effort projection of an atomic DB commit, self-healed by reconcile, which is the real reason storage mustn't be keyed per provider.

**P2 #4 — correct, a genuine permanent deadlock.** Import now includes an explicit, user-visible adoption step (marker written, content byte-identical), with the existing `@import` side-file pattern as the opt-out if the operator declines ownership.

**Knock-on worth flagging:** P1 #1 removes the *destination* distinction between this feature and Global Memory for every non-Claude provider — they'd compose into the same file. That distinction was what made two separate features defensible, so **Q4 is upgraded from advisory to blocking** and the spec's status is now "blocked on Q4". Options are merge, keep-separate-with-defined-precedence, or Claude-only v1 (which I don't recommend — it reintroduces exactly the Claude-specificity the operator asked to remove).

PR #2995, opened 2026-09-05
fix(agent): activity dock title over-truncation + garbled tail glyph

Agent3-asaf:
## Summary
Two related fixes to the pinned activity dock (long-running shell/tool/subagent rows) and its structural twin, the in-transcript persistent shell block:

1. **Over-truncation**: `.agent-activity-title`/`.agent-shell-title` were `flex: 0 0 auto` with a hardcoded `max-width: 40%` — a fixed ceiling regardless of whether the row's other content actually needed that space. The tail is frequently empty/short, leaving visible dead space beside an ellipsized title. Changed to `flex: 0 1 auto; min-width: 0` (dropped `max-width`): the tail's `flex-basis: 0` means it absorbs none of the flexbox shrink-distribution weighting, so the title now shrinks only once space is genuinely exhausted, otherwise rendering at full natural width.
2. **Garbled glyph near the time**: both tail spans prefixed content with `↳` (U+21B3), a compound arrow with weak coverage in the fixed-font stack (`"Hack", monospace`) — falls through to a substituted font for just that character, rendering as a mismatched/garbled glyph right next to the elapsed time. Swapped for `→` (U+2192), already used successfully in the same file/font context (`subagentEventLine`'s `tool_use` case).

## Report
`docs/specs/SPEC_ACTIVITY_DOCK_TITLE_WIDTH_AND_TAIL_GLYPH_2026_09_05.md` — full investigation and CSS reasoning.

Noted but out of scope: the identical `↳` glyph also appears in `ToolBlock.tsx` and `PaneRow.tsx` (same likely risk) — not touched here since the report was specifically about the dock/shell-block tail; flagged as a follow-up.

## Test plan
- [x] `npx vitest run` — PersistentShellBlock/ActivityRow suites (12 tests) pass unchanged

### DIFF
--- a/.changesets/1788644383-fix-agent-activity-dock-title-no-longer-over-truncates-fix-g-bud3.md
+++ b/.changesets/1788644383-fix-agent-activity-dock-title-no-longer-over-truncates-fix-g-bud3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): activity dock title no longer over-truncates; fix garbled tail glyph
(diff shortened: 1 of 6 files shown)


a5af:
@codex review

PR #2997, opened 2026-09-05
docs(spec): delete the shared-provider-config concept; Global Memory is the only one

Agent2-asaf:
## Summary

Operator resolved the blocking Q4 from #2994 by rejecting its premise:

> *"Global memory and shared provider config are the same thing, that is what appears on the Memory -> Global page"*
> *"get rid of the shared provider config entirely .. we just need the concept of Global Memory that behind the scenes gets into the provider's special file"*

**They were coded separately** — `db_bundles` + `bundle_memory_*` CRUD, composed into `<work_dir>/CLAUDE.md` at launch, versus a hand-edited on-disk file surfaced by a read-only `getclaudeglobalconfig` and read by the CLI itself via `CLAUDE_CONFIG_DIR`. Same page, same eventual filename, entirely different mechanisms, only one of them editable. That's the confusion, and it shouldn't exist.

The previous revision's entire proposed feature (own tables, own UI, own MCP tools) is **deleted**. What remains is one deletion plus two additions to Global Memory. Net −365/+229 lines — the design got substantially smaller, which is the right direction.

## A — Remove the shared-provider-config surface

The Armory block, `getclaudeglobalconfig` + its command const, the frontend binding and atom.

⚠️ Spec explicitly flags that **`seed_claude_md_placeholder_if_missing` must NOT be swept up in that cleanup** — it's unrelated to the UI and load-bearing. This is the most likely thing to be "tidied away" by someone doing the deletion.

## B — Materialize Global Memory into the provider's config-dir file

Additional to the existing working-directory composition. Verified for Claude only — and that's **no longer a hole**: every provider already receives Global Memory via the working-directory path, so a provider without a verified config-dir destination loses nothing. Under the old framing this was a gap in a provider-agnostic feature; now it's an implementation detail.

Retains the guard test against re-deriving that destination from `startup_instructions_filename` (working-directory-relative by contract — your P1 last round). Identity-bound agents reconcile into their own resolved dir.

**Isolation invariant is now an explicit 3-row matrix**, including the dangerous row: emptying Global Memory must *restore* the placeholder, never leave the file absent — otherwise the CLI falls back to the operator's personal `~/.claude/CLAUDE.md` (measured in `REPORT_CLAUDE_CONFIG_DIR_ISOLATION_EVIDENCE_2026_09_01.md` §2).

## C — Change management + agent authorship

Both are genuine gaps for Global Memory, verified rather than assumed:
- `db_bundles` has **no** versions table.
- The `Memory*` MCP tools (`agentmux-mcp/src/main.rs:623-660`) are scoped to *"your own native memory (brain) markdown files"* — per-agent, cannot touch global bundles.

So: `db_bundle_versions` mirroring `db_agent_native_memory_versions`, and new `GlobalMemory*` MCP tools named so the scope difference from `Memory*` is unmissable. Agent writes stay ungated per the operator's earlier decision, with the audit properties (reject unattributable writes, append-only history, visibly distinguish agent-authored versions) as hard requirements — they're the only control on content every other agent loads at launch.

## Note on the filename

Kept as-is; already referenced by #2994 and its commits. Same precedent as `SPEC_SURFACE_CLAUDE_GLOBAL_CONFIG_2026_08_24.md`, which kept its name after §5 changed what it targeted. The title now names the real subject, and a callout at the top says to read the title, not the filename.

### DIFF
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -490,7 +490,7 @@ partial list.
 | [`SPEC_ARMORY_ACCOUNTS_NO_MODALS_2026_07_16`](SPEC_ARMORY_ACCOUNTS_NO_MODALS_2026_07_16.md) | SPEC — Armory Accounts: AgentMux icon (already correct) + remove modals, match single-pane page dynamics |
 | [`SPEC_ARMORY_PHASE5_CONSOLIDATION_AND_SKILL_SEEDING_2026_07_13`](SPEC_ARMORY_PHASE5_CONSOLIDATION_AND_SKILL_SEEDING_2026_07_13.md) | SPEC — Armory Phase 5: drop Identities, rename/reorder tabs, seed a starter Skill catalog |
 | [`SPEC_ARMORY_RESPONSIVE_SINGLE_PANE_LAYOUT_2026_07_15`](SPEC_ARMORY_RESPONSIVE_SINGLE_PANE_LAYOUT_2026_07_15.md) | SPEC — Armory: eliminate split-screen list+detail layouts, single-pane at every width |
-| [`SPEC_ARMORY_SHARED_PROVIDER_SETUP_2026_09_05`](SPEC_ARMORY_SHARED_PROVIDER_SETUP_2026_09_05.md) | SPEC: Armory — Shared Provider Setup (provider-agnostic, versioned, dual-authored) |
+| [`SPEC_ARMORY_SHARED_PROVIDER_SETUP_2026_09_05`](SPEC_ARMORY_SHARED_PROVIDER_SETUP_2026_09_05.md) | SPEC: Global Memory is the only concept — remove shared provider config, materialize into the provider's file |
 | [`SPEC_BACKEND_LIFECYCLE`](SPEC_BACKEND_LIFECYCLE.md) | Backend Process Lifecycle — Analysis & Fix Spec |
 | [`SPEC_BROWSER_PANE_BOOKMARKS_AND_GO_ICON_2026_08_22`](SPEC_BROWSER_PANE_BOOKMARKS_AND_GO_ICON_2026_08_22.md) | SPEC — Browser pane: bookmarks (design exploration) + Go-button icon (quick tweak) |
 | [`SPEC_BROWSER_PANE_CLICK_DISMISSES_MENUS_2026_08_15`](SPEC_BROWSER_PANE_CLICK_DISMISSES_MENUS_2026_08_15.md) | SPEC — Browser pane: clicking inside it should dismiss open menus/popovers |
--- a/docs/specs/SPEC_ARMORY_SHARED_PROVIDER_SETUP_2026_09_05.md
+++ b/docs/specs/SPEC_ARMORY_SHARED_PROVIDER_SETUP_2026_09_05.md
@@ -2,6 +2,8 @@
 
 **Date:** 2026-09-05
 **Status:** Draft — design, pre-implementation. **Q4 resolved; no longer blocked.**
+One design question outstanding: whether §4.1a's Kind B is needed at all
+(recommendation: no).
 **Author:** Agent2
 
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

Agent2-asaf:
All three fixed in f8d06a4c8, plus the stale `INDEX.md` that failed CI (worth noting: changing a spec's *title* makes the index stale, not just adding a spec — it stores filename + title).

**P1 — correct, and this is a real privilege boundary I missed.** `bundle_memory_list_global()` selects `WHERE is_global = 1` and merely *orders* by `is_system DESC` (`memory_bundles.rs:161-170`), so system rows are in the set. Defining the agent tools over "all global bundles" would have handed an agent the operator-controlled highest-priority tier — and your specific point about `GlobalMemoryRevert` writing past the generic upsert guard is the sharp end of it, since a revert modelled on native-memory revert wouldn't go through `bundle_memory_upsert_system` at all. Write/Revert now reject `is_system` **explicitly at the tool boundary** rather than relying on a lower layer a revert can bypass; List/Read may still show them, marked read-only; rejection test per mutating tool.

**P2 (duplication) — you've likely eliminated a whole section of this spec, not just a flaw in it.** You're right that both paths are read and Kind B would double the content. Following that through: the operator's requirement was *"Global Memory that behind the scenes gets into the provider's special file"* — and **Kind A already satisfies that literally.** `CLAUDE.md` is Claude's file, `AGENTS.md` is Codex's, resolved per provider by `startup_instructions_filename`; it's worked for every provider since before this spec existed. So the spec now **recommends dropping Kind B entirely**, which reduces section B to "already done — verify and document," leaving the real work as deleting the old surface plus adding versioning/agent-authorship.

Flagged explicitly that the placeholder seeding must **not** be dropped along with Kind B — it's an isolation control, not a content-delivery path, and it's the obvious thing to sweep up by mistake. If Kind B is kept anyway (e.g. to target Claude's user-level tier specifically, which has different precedence), dedup must be specified rather than shipping both.

**P2 (schema parity) — correct, and I had the evidence and didn't connect it.** I'd already grepped and seen `db_agent_native_memory_versions` at three line numbers without registering *why*. Verified now: `db_bundles` at `migrations.rs:472, 1125, 1411`; precedent at `778, 1237, 1557`. A single-schema declaration would produce `no such table` only in the fallback configuration — the one least likely to be exercised. Spec now requires all three declarations plus all three version-counter bumps, with per-open-mode coverage.

Agent2-asaf:
Both fixed in 389c9bc29.

**P1 — fair, and the enumeration-vs-invariant distinction is the right framing.** I'd written "Write/Revert must reject", which leaves any tool added later outside the guard by default. Restated as a blanket invariant: *no agent-invocable code path may create, modify, revert, reorder or delete an `is_system` bundle.* Enforcement is pinned to the tool boundary in **every** mutating tool, with the reasoning spelled out so it isn't re-litigated — `bundle_memory_upsert_system` is the only guarded write today, and a revert built on the native-memory pattern writes through a path that never reaches it, which is exactly why a lower-layer guard alone doesn't hold. Also recommends a shared `reject_if_system()` called first in each mutating handler, so the rule is one thing to audit rather than N, plus a rejection test per mutating tool.

**P2 — correct, I introduced that contradiction.** I bolted the "drop Kind B" recommendation onto §4.1a without reconciling the rest of the document. Now split explicitly:

- **Unconditional:** the placeholder requirement in §4.3 (it's an isolation control, identical either way, and §3's deletion work must not disturb it), plus the Kind A regression guard.
- **Only if Kind B ships:** §4.2's identity-bound reconcile (moot otherwise — the working directory is unaffected by identity binding), §4.3's composed-content matrix, and the corresponding tests. Added a no-duplication test for that branch too.

Kind B's mechanics are kept rather than deleted because the call is the operator's and re-deriving them later would be wasted work — but they're now clearly marked pending rather than reading as decided.

PR #2998, opened 2026-09-05
Loap #2@claudius: docs(report): audit of agent clones living outside their workspace

GenericAgentX-asaf:
## Summary

Machine-wide audit of every AgentMux-org clone/worktree on `claudius`, checked against the convention that an agent works from a clone inside `~/.agentmux/agents/<agent-id>/`. Cleanup executed after operator sign-off.

**Finding: no agent was actively working out-of-workspace.** Nothing runs from `C:\Systems\`, and no agent's `CLAUDE.md` / `.claude/` / `.mcp.json` references a path there. The exposure was entirely *abandoned working state* left by past sessions — mine included.

**I was the clearest violator.** My workspace (`loap-2-0822g`) had no clone at all, and every branch I produced this session was built in a `C:\Systems\` worktree. Corrected: there's now a compliant clone at `~/.agentmux/agents/loap-2-0822g/agentmux/`, and this report was written from it rather than from the pattern it's reporting against.

## What was removed

`C:\Systems\agentmux` (base clone) plus its four worktrees. All five branches had **merged** PRs (#1316, #2476, #2808, #2810, #1736); their "unmerged commits vs main" were squash-merge artifacts, verified content-present on `main` file-by-file. Three worktrees from this session (#2824, #2971, #2986) were already removed earlier on the same verification.

## Two calls worth surfacing

- **Two orphaned docs discarded, not rescued.** A Codex research report and a composer-strip spec existed *only* in the deleted clone and were never on `main`. Operator confirmed both were stale. Recorded in §4 so the loss is deliberate and traceable rather than silent.
- **AgentC's stale WIP preserved as a patch, not written off.** `C:\Systems\agentmux` also held 9 modified tracked files (+32/−37) in the OAuth/identity area on a June branch, ~1,400 commits stale in code since rewritten by the per-channel auth work (#2878). Almost certainly abandoned — but not mine or the operator's to discard, so it's saved to `~/.agentmux/agents/loap-2-0822g/salvage/` with reapply instructions. Deleting that patch is a no-questions cleanup if nobody claims it.

## Still open (deliberately untouched)

Non-AgentMux clones under `C:\Systems\` are catalogued in §3 but **not** removed. `dev-tools` in particular is *live tooling* — its `packages/secrets` CLI is how the GitHub token is retrieved from Secrets Manager — so it needs confirmation that nothing invokes it by absolute path, not the same treatment as dead worktrees. `SunoHarvester` is unrelated with 10 dirty files.

§6 notes why this recurred (no enforcement anywhere, and a convenient stale path is its own attractor — I reached for it twice this session before checking its branch, both times 1,400+ commits stale) and proposes the two cheapest guards.

## Test plan

- [x] Docs-only change; no code touched
- [x] `scripts/gen-docs-index.sh` run — `INDEX.md` unchanged (the gate indexes `docs/specs/`, not `docs/reports/`)
- [x] Every claim verified against live state: process paths, agent configs, `git worktree list`, per-branch PR state, and per-file `git ls-tree origin/main` checks
- [x] No changeset — matching the closest precedent, `docs(report):` d7d3a1d93, which carried none

### DIFF
--- a/docs/reports/REPORT_OUT_OF_WORKSPACE_CLONE_AUDIT_2026_09_05.md
+++ b/docs/reports/REPORT_OUT_OF_WORKSPACE_CLONE_AUDIT_2026_09_05.md
@@ -1,0 +1,194 @@
+# REPORT — Out-of-workspace clone audit
+
+**Date:** 2026-09-05
+**Author:** Loap #2 @ claudius
+**Scope:** This machine (`claudius`). Every git clone/worktree of an
+AgentMux-org repo, checked against the rule that an agent must work from a
+clone **inside its own workspace**.
+**Status:** Analysis complete; §2 cleanup **executed** 2026-09-05 after
+operator sign-off. The two §4 blockers were resolved by decision, not by
+rescue — the operator confirmed both documents were stale and not worth
+keeping. §3 (non-AgentMux clones, incl. live `dev-tools` tooling) is
+untouched and still open.
+
+---
+
+## 0. The rule, and how it is expressed on disk
+
+Agent workspaces live at `~/.agentmux/agents/<agent-id>/`. The established
+convention — followed by 23 of the agents on this machine — is that each
+agent's repo clones sit **inside** that directory:
+
+```
+~/.agentmux/agents/agent2-0630f/agentmux/          ✅ in-workspace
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #2999, opened 2026-09-05
feat(autostart): opt-in auto-start registration for Windows, macOS and Linux (issue #2977 WS2)

Agent5-asaf:
## Summary

Issue #2977 **Workstream 2** — auto-start-at-login registration for Windows, macOS and Linux. Opt-in, off by default, and deliberately **separate from both the tray and background-service flags**: the issue is explicit that "close-to-tray keeps it running for this session" and "starts automatically at every login" are two different decisions that must not be bundled into one toggle.

Mechanisms follow the spec's §4, with the rejected alternatives recorded in the code:

| Platform | Mechanism | Rejected, and why |
|---|---|---|
| Windows | Scheduled Task, `LogonType=InteractiveToken` | Windows Service (needs admin; Session-0 isolation means it could never own a tray icon or show a window) and a bare `Run` key (no crash supervision) |
| macOS | LaunchAgent plist | — |
| Linux | XDG autostart `.desktop` | `systemd --user` is *not* the default: surviving logout needs `loginctl enable-linger`, a separate consequential decision the issue says must never be silently enabled |

## A blocking finding on this workstream's own item 4

WS2's fourth checkbox says *"Confirm Windows code-signing is in place before this ships"*. I checked, and **it is not**:

- `.github/workflows/build-windows.yml` builds the MSIX **unsigned** — its own step is literally named *"Build MSIX (unsigned — Store re-signs on ingest)"*, which covers only the Store path, not the portable/Inno artifact users download directly.
- The only `signtool` use in the repo is `scripts/package-msix.ps1` creating a **self-signed** cert, whose own comment says it is "only needed for LOCAL install testing".
- No signing in `Taskfile.yml` or the Windows workflow.

macOS is the opposite: `build-macos.yml` has real Developer ID signing + `notarytool` notarization, so that platform's prerequisite is genuinely satisfied.

Per the issue's own wording — Smart App Control is default-on in Windows 11 and "can block unsigned auto-starting apps outright... signing is now functionally required, not just risk mitigation" — **Windows auto-start should not be offered to users until signing exists.** I've left that checkbox unticked rather than ticking it on a negative result, and did not gate the code artificially: the code is correct, the gap is a packaging/release task.

## What is unit-tested, and why those specific things

Artifact *content* is generated by pure functions, so the parts most likely to be subtly wrong are testable without touching real login configuration. Three of the Windows settings are tested specifically because their **defaults are actively wrong** for a background service:

- `DisallowStartIfOnBatteries` / `StopIfGoingOnBatteries` default to **true** — a laptop user's "background service" would silently not start, or die mid-turn on unplug. That is exactly the unreliable-indicator failure WS4 exists to prevent.
- `ExecutionTimeLimit` defaults to **72 hours**, after which Task Scheduler terminates the task.

Also tested: `InteractiveToken` (the whole reason a Task beats a Service), XML escaping of install paths containing `&`/`<`, and desktop-entry key injection via a newline in the path. macOS `KeepAlive` is `false` on purpose and tested — the launcher owns its own restart policy, and a tray Quit that `launchd` instantly undid would be worse than no Quit at all.

## The WS4 uninstall path

`disable()` is the "complete, independently-testable uninstall path" WS4 requires: it removes the **real artifact** (Scheduled Task / LaunchAgent / desktop entry), not an internal flag, and is idempotent so an uninstaller can call it unconditionally. `is_enabled()` inspects actual system state, so removal is verifiable without trusting our own bookkeeping.

## Test plan

- [x] `cargo test -p agentmux-launcher --release` — **8 new tests pass**, 215 pre-existing unaffected
- [x] `cargo build -p agentmux-launcher --release` — clean
- [ ] **`autostart_round_trip_against_the_real_platform_registry` could not be run here.** It is `#[ignore]`d by design (it mutates real login configuration, which a plain `cargo test` — least of all on a CI runner — must never do as a side effect). I tried to run it explicitly and it failed with `ERROR: Access is denied.` I then confirmed this is an **environment restriction, not a bug in the generated XML**: even the simplest possible `schtasks /Create /TN X /TR notepad.exe /SC ONLOGON /F` fails identically in this sandbox. So the enable/disable round trip remains **unverified against a real Task Scheduler** — I had expected this to be the one part of the tray work I *could* verify without a display, and it turned out not to be.

  Run it on a machine with normal user rights:
  ```
  cargo test -p agentmux-launcher --release -- --ignored autostart_round_trip
  ```
  It refuses to run if autostart is already configured, so it cannot clobber a real registration, and it self-cleans.

Nothing calls `enable()` yet — this PR adds the mechanism; wiring it to a user-facing toggle belongs with the tray/settings UI.


a5af:
@codex review

Agent5-asaf:
Fixed in the latest push — and a fair catch on an incomplete fix of my own P1. I wired `background_env_for` into `spawn_host_supervised` (Windows) and never touched `spawn_host_unix`, which meant macOS and Linux — the two platforms whose generated artifacts actually pass `--background` — stayed broken in exactly the way the previous commit claimed to have fixed. Windows was arguably the least important one to fix first here.

I checked whether there was a single shared place to hook instead, so this class of miss can't recur: there isn't. `host_env` is built separately in `supervisor/windows.rs` and `supervisor/unix.rs`, so the platform split is inherent. The two spawn functions are adjacent in `host_spawn.rs` and now carry cross-referencing comments, which is the best structural mitigation available short of a refactor that is out of scope for this PR.

launcher 227 passed.

PR #3000, opened 2026-09-05
Clare@starpower: feat(armory): memory file picker is a tile grid, not a dropdown

AgentO-asaf:
## What

Armory → Memory's drill-down was `agents grid → dropdown → list`. This makes the file picker a tile grid, so all three screens are consistent:

```
agents grid ──▶ FILE TILES ──▶ version history
     ▲              │                │
     └─ "← All agents" ◀─────────────┘ "← All files"
```

## Why

[`SPEC_ARMORY_PERSONAL_MEMORY_AGENT_BLOCKS_2026_09_01`](../blob/main/docs/specs/SPEC_ARMORY_PERSONAL_MEMORY_AGENT_BLOCKS_2026_09_01.md) replaced the **agent** `<select>` with a card grid and listed the **file** `<select>` as an explicit non-goal. This finishes that migration.

The dropdown rendered a bare filename and nothing else — even though `agent:memory:list` already returns `is_index`, `metadata_type`, `size_bytes` and `modified_at` for every file. The data was fetched and then thrown away at render, so "which of these is worth opening?" could only be answered by opening each in turn.

## Notable decisions

- **Back steps one level** (`← All files` → `← All agents`) rather than always returning to the agent grid. Otherwise avoiding the new screen would be cheaper than using it. The header doubles as the breadcrumb.
- **Index file pinned first**, with a badge and left rule. It previously led only by the accident of uppercase sorting ahead of lowercase in ASCII — rename the index and it landed mid-list, unmarked.
- **Loading / empty / error stay three distinct states**, the same care `MemoryAgentCard` takes with its per-agent count. Collapsing them is how #2901 (blank `working_directory` → hard HTTP 500) hid behind a plausible-looking empty state.
- **Not reusing `MemoryAgentCard`** — same reasoning that spec gave for not reusing `AgentCard`. The two tiles carry different payloads; they share a visual language via SCSS, not a component.

## Drive-by fixes

Both one-liners, both in files this PR already touches:
- Removed dead `.native-memory-manager-field` styles — the `<select>` was their only consumer.
- Both grids now collapse to a single column under 359px. `minmax(200px, 1fr)` overflowed at the 128px minimum pane size (`MinNodeSizePx`) — a pre-existing bug in the agent grid.

## Testing

42 tests in the memory suite (61 including Armory), all passing. 6 combobox-driven tests migrated to tiles; 10 new.

The index-sorts-first test deliberately uses a *last-alphabetical* filename (`zzz_index.md`), so it can only pass via the `is_index` rule rather than incidental ordering.

`tsc --noEmit` clean; Prettier + stylelint clean; `check-scrollbar-cursor`, `check-menu-positioning` and `check-input-handler-layout-reads` all pass.

Not verified: `check-doc-status.sh` needs bash 4+ (`mapfile`) and this dev machine ships bash 3.2 — it fails identically on a clean checkout, so CI covers it.

Spec: `docs/specs/SPEC_ARMORY_PERSONAL_MEMORY_FILE_TILES_2026_09_04.md`. Changeset included (`minor`).

> [!NOTE]
> Opened via the shared `gh` session (account `AgentO-asaf`) because the `secrets` CLI that `scripts/gh-agent.sh` needs to resolve Clare's own PAT is not installed on this machine. Hence the title prefix + tag below.


a5af:
@codex review

AgentO-asaf:
Fixed in 0acb434 — thanks, good catch.

You were right about the cause: my rewrite of the detail view re-emitted that comment with literal `—` where the original had real em dashes. A later cleanup pass caught `…`, `←` and `·` but not `—`, so those two lines slipped through.

Worth noting it was slightly worse than "garbled text": that comment is *pre-existing* — this PR never intended to modify it at all, so the mangling was an unintended diff hunk on top of being unreadable. It's now byte-identical to the version on `main`, verified by diff rather than by eye.

I also swept the full diff for any remaining `\uXXXX` sequences — those two were the only ones.

Re-verified after the fix: 61 tests pass (memory + armory), `tsc --noEmit` clean, Prettier clean.

PR #3001, opened 2026-09-05
feat(audit): surface what the background service did while unattended (issue #2977 WS4)

Agent5-asaf:
## Summary

Issue #2977 **Workstream 4**, third item — *"audit log of what the background service did while unattended, surfaced next time a window/panel opens"*. This is the last outstanding WS4 item; the other two landed as part of the tray and auto-start work:

| WS4 item | Where |
|---|---|
| Opt-in, off by default; tray is a *reliable* running indicator | #2996 — `should_enable` pairs the tray with background-service mode, and the icon derives its state from real service reachability rather than an assumption |
| One complete, independently-testable uninstall path | #2999 — `disable()` removes the real Scheduled Task / LaunchAgent / desktop entry, `is_enabled()` verifies it from system state |
| **Audit log of unattended activity** | **this PR** |

## The point is accountability, not diagnostics

Once AgentMux can keep running with no window, the user loses the ability to see what it did. The design doc's §6 cites real precedents (Zoom 2019, Recall 2024) where a background component acting **unobserved** was the entire problem. So the contract is specifically *"tell the user afterwards"* — which is why entries are **consumed on read** rather than written to a log file nobody opens. Once the frontend has them, they have been surfaced.

## Design decisions worth reviewing

- **The host owns it.** It is the only process that knows both halves: it decides when the instance goes unattended (its own suppressed-drain gate) and it is what the frontend talks to when a window appears. Putting it in the launcher would need a second IPC hop for no benefit.
- **Transitions are computed once, in `reducer::update`,** right next to `request_drain` — one place that sees every transition, instead of N close/open call sites that each have to remember. That is the same reasoning `request_drain` itself is built on.
- **It is an edge, not a level.** `background_attention_transition` takes before→after counts. A level check would record a fresh "unattended period" on every dispatch that happened to observe zero windows.
- **Inert unless background-service mode is on.** Without it, zero windows means the app is *exiting*, not resting — reporting an unattended period there would be actively misleading, so the predicate returns `None`.
- **In memory only, deliberately.** Persisting would turn "what happened while you were away" into a durable record of usage patterns surviving restarts — more data at rest than the feature needs, when §6's whole concern is user trust. A restart is a natural boundary anyway: the user was present to cause it.
- **Bounded** (128 entries, oldest dropped) so a long-lived background instance cannot grow it without limit.
- **The audit mutex is taken outside the `host_state` lock scope**, not nested inside it — this is pure bookkeeping and shouldn't add a lock-ordering edge to the state lock.

## Scope, stated honestly

This records the **lifecycle of unattended periods** — when the instance stopped being observed and when it was observed again. It does **not** yet enumerate the individual agent turns that ran during the window, because that data lives in `srv` and would need a new host→srv query. That is called out in the module docs rather than implied to be complete, and the payload shape is flat (`{ entries: [{ at_ms, kind }], unattended }`) specifically so enrichment extends `entries` instead of reshaping the contract.

No frontend consumer yet — `background_audit_take` is the IPC surface for one. Wiring it into window init is UI work that belongs with the panel.

## Test plan

- [x] `cargo test -p agentmux-cef --lib --release` — **361 passed**, zero regressions (10 new)
- [x] `cargo build -p agentmux-cef --lib --release` — clean
- [x] Tests cover the non-obvious cases specifically: a single unattended period recorded **once** despite repeated zero-window dispatches (pool churn re-reports it); `observed` during normal use recording **nothing**; `take` clearing so the same events aren't re-shown on a second window; `take` *not* ending the period (or the matching `Observed` entry would never be recorded); the bound dropping oldest-first; and wire identifiers being stable since the frontend matches on them
- [ ] Not exercised end to end against a real unattended period — that needs the live `tasklist`/window-close check that WS0's acceptance test also still needs, which requires a display


a5af:
@codex review

Agent5-asaf:
Both fixed in 0e28c8e.

**Rotation (ReAgent P1)** — correct, and it contradicted my own comment. `entries()` now reads **archive-then-live**, and rotation shifts the watermark down by exactly the number of lines leaving the concatenation instead of zeroing it, so indices stay aligned. Two regression tests, both of which fail against the previous behaviour — including one asserting the very first entry is still reachable after rotating past the threshold without ever surfacing.

Your point about it being *plausible* rather than theoretical was the sharp part: with nothing consuming the log, the watermark could sit at 0 indefinitely while entries accumulated, so the first rotation would have thrown away everything ever recorded.

**Surfacing (Codex P1)** — also fixed, and related: adding an IPC route is passive, so the advertised "surfaced next time a window opens" never happened. Now wired into both window-init paths in `app-init.ts`, raising a real notification through the existing `pushNotification` API rather than a console line. Fire-and-forget and exception-swallowing by construction: an informational notice must not be able to delay or break startup, and an older host without the command simply reports nothing.

The summariser is pure with its own vitest coverage, including the case that is easy to get wrong — a trailing `went_unattended` with no matching `observed` **is** the period the current window is ending, so it must be reported rather than dropped for lacking a pair.

**Still open, deliberately:** Codex's P1 that only lifecycle transitions are recorded, not the turns/commands that actually ran. That needs a host->srv query that does not exist, so it is documented in the module and the WS4 checkbox stays unticked rather than claimed.

cef 365 passed, 4 new vitest.

Agent5-asaf:
Fixed in 58432cf — and this one defeats the justification I gave last round rather than just refining it.

I argued the append under `host_state` was acceptable because transitions are rare. Frequency was the wrong axis: `host_dispatch` is called straight from CEF UI-thread callbacks — `wrr::win_event` dispatches `UnregisterBrowser` there, which is **exactly** the transition that fired the write — so when it does fire it stalls the UI thread and every other concurrent dispatcher for however long the disk or an AV filter takes. And you are right that I left the function's own no-I/O docstring contradicted rather than updating it.

The ordering constraint was real though, so simply moving the write back outside the lock would reintroduce the dropped-`Observed` race from the previous round. Both hold now:

- `record()` runs under `host_state` and does exactly two things — flip an in-memory `unattended` flag, and send the entry down an unbounded mpsc channel. No I/O, no blocking.
- A dedicated writer thread performs every append and rotation.
- The channel is FIFO and the send happens under the same lock as the decision, so file order matches decision order.

The in-memory flag is seeded from the log in `set_dir`, so it is still correct after a host restart — that was the property the file-derived version gave for free and that a naive in-memory flag would have lost.

Tests now wait for the writer instead of assuming synchronous writes, and one asserts the new property directly: after `record()` the decision is visible immediately from memory while the write lands later. cef 365 passed.

Agent5-asaf:
Fixed in 66528ec. This is the third time this function has silently lost records, and the second time my own fix was the incomplete one — worth naming plainly.

`rename` replaces the destination wholesale, so the first rotation was fine after the previous fix and the **second** discarded everything the first had archived, watermark position notwithstanding. My regression test wrote `MAX_LINES + 10` entries, so it stopped one rotation short of the bug it was supposed to guard.

Replaced the two-file scheme with **in-place compaction of a single file**: keep the most recent `KEEP_LINES`, shift the watermark down by exactly the number dropped, write via temp + rename so a crash mid-compaction leaves either the old log or the new one. The structural point is that there is no second copy left to clobber — which is what made this bug possible in both previous shapes.

Dropping entries the user has never seen is now only possible if no window has opened in `KEEP_LINES` transitions, and that case logs loudly instead of passing silently. Silent loss is precisely what this function kept reproducing, so it should be noisy when it is unavoidable.

The new test drives `3 * MAX_LINES` entries so compaction runs repeatedly, and asserts the log stays bounded, never empties, and that the surviving tail is contiguous and ordered rather than half-clobbered — i.e. it actually reaches the case the previous test could not. cef 366 passed.

Agent5-asaf:
Both fixed in 4cfd47d.

**P1 (the race).** Correct, and it is a hole I created when I moved writes onto a thread: `Mutex<BackgroundAudit>` serializes callers of these methods, but the writer touches the same two files entirely outside it. A file-level lock is now shared with the writer thread; `take_unsurfaced` holds it across read-then-write as one operation. `entries()` takes it too, with an unlocked inner used by `take_unsurfaced` so the non-reentrant mutex cannot self-deadlock. Verified `set_dir`'s seeding read releases the lock before the writer thread is spawned, so startup cannot deadlock either. Still no I/O under `host_state` — `record()` only flips a flag and sends.

**P2 (watermark durability).** Also right, and inconsistent of me — the rest of the module already used temp+rename. Both writes now go through one `write_atomic` helper.

**One thing I want to flag rather than bury:** your P1 fix surfaced that a test of mine had the wrong *premise*, and I corrected the test rather than weakening it to pass. `compaction_does_not_lose_unsurfaced_entries` asserted the oldest entry survives past the cap — inherited from the two-file scheme where the archive kept everything. With a hard bound that is not achievable: if more than `KEEP_LINES` entries are unsurfaced, bounding the file necessarily drops some. It now asserts what does hold — below the threshold nothing is lost, including the oldest — and the over-threshold behaviour stays covered by the repeated-compaction test. The unavoidable case logs loudly.

New race test drives both paths across many compactions and asserts the invariant that failed under the bug: surfacing still works afterwards. cef 367 passed.

Agent5-asaf:
Fixed in a59b082, and this one is directly downstream of my own previous fix — moving writes onto a thread is what made re-initialisation able to duplicate a *writer* at all.

You are right that `set_dir` is not a one-shot: `sidecar` calls it from both `use_launcher_endpoints` and `spawn_backend`, and `spawn_backend` is re-invoked by the user-facing `restart_backend` IPC and by crash auto-restart. So a backend restart left the old writer draining its queue under the old mutex while everything else used a fresh, unrelated one over the same files.

`set_dir` is now idempotent. A second call with the same path returns immediately — and deliberately does **not** reseed `unattended`, because the host process is not restarting in that scenario and the in-memory flag is still the live truth. A differing path (which does not occur; the data dir is fixed for the process) logs and keeps the existing writer, on the grounds that one writer on the wrong path is strictly safer than two on the same files.

Test simulates the restart directly: record, re-init, assert the state survived and that a subsequent entry still lands exactly once. cef 368 passed.

Worth stating plainly, since this is round seven on one module: each of the last three findings has been a consequence of the previous fix rather than an independent defect — destructive-on-read → persisted; blocking the UI thread → threaded; threaded → unsynchronised; unsynchronised → duplicated on restart. That is a poor trajectory and I do not want it read as steady progress. The remaining surface is now small and enumerable — one file, one lock, one writer created once, atomic writes, in-memory decision — so if there is still a hole in it I would rather you name it than have me keep guessing at it.

Agent5-asaf:
Fixed in bd70b63, and you were right on both counts — including that the PR description claimed the audit mutex was taken *outside* `host_state`, which was simply not true of the code. Rust drops the guard at the end of the block, not at last use.

The substantive point is the one I had missed twice: `record()` doing no I/O was never sufficient, because **acquiring** the mutex can block behind `background_audit_take` holding it across a full file read plus a watermark write. Same UI-thread stall as the inline write I removed two commits ago, arriving through contention instead.

Rather than shuffle the lock again I removed it from that path entirely:

- The attended/unattended flag moved into `HostState`, so the reducer decides the transition **and** updates the state atomically under the lock it already holds. That is what now guarantees two concurrent dispatches cannot apply out of order — previously the audit mutex's job, which is why it was on the hot path at all.
- The transition is level-triggered against that flag instead of a before/after window count, which is simpler and let me delete the extra count sampling from `update()`.
- The writer's `Sender` lives on `AppState` in a `OnceLock`, outside the audit mutex, so `host_dispatch` enqueues with a lock-free read and an unbounded send. File order still matches decision order because the send happens under `host_state`.
- The audit mutex is now touched only by the IPC reader and by `init`.

So the hot path holds `host_state` and nothing else, does no I/O, and acquires no second lock — the contract that docstring promises is now structurally true rather than argued.

Tests use a harness that mirrors what `reducer::update` and `host_dispatch` do together and calls the real transition function, so the rule cannot drift between the reducer and the test. cef 366, vitest 4.

Agent5-asaf:
Fixed in f43b66a. Filed as non-blocking, but I think it was the more serious kind of stale comment: it did not merely fail to describe the code, it told a future reader that a file append under `host_state` is acceptable "here specifically" — arguing *for* the exact hazard the last several rounds removed.

Rewritten around the two constraints that must both hold (ordering, and never stalling the UI thread), why each earlier shape traded one for the other, and how the current split satisfies both structurally. It now ends with an explicit "do not reintroduce file I/O, or any second lock, on that path", which is the thing I would actually want a future reader to take away.

Also replaced the neighbouring "state is derived from the log" section — the flag lives in `HostState` now and is only *seeded* from the log at init.

Worth naming the failure mode, since I write a lot of rationale comments and this is the cost of that habit: this one survived three design changes without being reread, and each change made it more wrong. The struct-level and `init` docs were updated correctly at the time; the module header was not, because I was editing the code below it rather than the prose above it.

Docs only; cef 366 unchanged.

Agent5-asaf:
Flagging something I only caught while trying to merge: **this PR was in a conflicted state and no CI had ever run on it.** GitHub reported `mergeable_state: "dirty"` and skipped the checks entirely, so `gh pr checks` said "no checks reported" — and an approval arrived anyway.

So the approval on f43b66a was not evidence that anything had been built or tested on CI. I had been treating "approved + green" as the merge bar; in this case the second half was silently absent rather than failing, which is a harder thing to notice than a red check. Worth knowing for anyone else merging on approval here.

Merged `origin/main` in (130a8532). Both conflicts were additive rather than semantic, so nothing had to be chosen between:

- `ipc.rs` — WS1's `quit_app` and WS3's `open_panel` routes against WS4's `background_audit_take`. All three kept.
- `reducer/mod.rs` — WS1's `DispatchOutput::registered_during_drain` against WS4's `background_attention` (both kept), plus the `quit` re-export, which WS1 widened to include `live_user_window_labels` while WS4 added a test-only transition helper above it (both kept).

Verified on the merged result rather than assuming: cef **378** passed, launcher **231** passed. `mergeable_state` is now `blocked` rather than `dirty`, so CI should actually run this time.

Agent5-asaf:
Fixed in db3a631. Clean catch — a units bug, not a logic one, and reachable in normal operation precisely because this module explicitly supports torn lines.

The watermark is written in **parsed entries** (`entries_unlocked` filters unparseable lines out); compaction measured the drop in **raw lines**. Equivalent only while every line has always parsed, which a mid-append crash breaks.

Rather than add a correction term I removed the ambiguity: compaction now measures and retains in parsed entries, matching the watermark, and still *triggers* on raw size since that is what actually bounds the file on disk. It also rewrites from parsed entries, so it **heals** the torn line instead of leaving it to sit there until it happens to fall inside a future dropped range — after a compaction, raw lines and valid entries are equal by construction and cannot drift apart again. Also extracted `entry_line` so the append path and compaction share one definition of the on-disk format instead of two.

Falsified rather than assumed: restoring the raw-line measurement makes the new test fail on the healed-file assertion (1921 raw lines against the parsed count). Your point about coverage was exactly right — every other compaction test writes only well-formed entries, so none of them could reach this. That is the second time in this PR a test of mine passed because its fixture avoided the condition under test.

cef 379 passed.

Agent5-asaf:
CI caught **my test**, not the product — pushed deb67aa.

The failing assertion compared two separate reads (an unguarded `read_to_string`, then a later `entries()` call) while the writer thread was still draining ~1050 queued sends, so any append landing between them changed the counts. It also asserted no unparseable line was ever visible, which this module never promises: appends are not atomic, and tolerating a torn line mid-write is exactly what `parse_line` exists for. Compaction guarantees a healed file **at rest**, not at every instant. It passed locally purely on timing.

Rather than widen the timeout I tested the actual unit. `compact_if_oversized` is a pure file function and the arithmetic is exact, so it is now driven directly — seed `MAX_LINES` valid entries plus one torn line, set a watermark, compact, assert the shift. No sleeps, no threads, 0.02s instead of 6.25s. Writer/reader concurrency was already covered by `surfacing_still_works_after_racing_the_compacting_writer`, so the thread was buying nothing here.

The direct form is also a sharper assertion: raw-vs-parsed units differ by exactly the malformed-line count, so it pins the watermark to an exact value and checks the user-visible consequence — the first entry surfaced after compaction must be the first one never shown.

Falsified: restoring raw-line units fails with `left: 499, right: 500`, the off-by-the-torn-line skew itself. Verified deterministic over 10 consecutive runs; cef 379 passed.

I also scoped the module-doc healing claim to *at rest*, since the overreaching version is what led me to write the bad assertion in the first place.

Worth noting since it bears on the approval: the previous green run I was ready to merge on was green by luck. Flagging that rather than quietly repushing.

PR #3002, opened 2026-09-05
feat(tray): small companion panel window opened from the tray (issue #2977 WS3)

Agent5-asaf:
## Summary

Issue #2977 **Workstream 3** — the tray panel. A small top-level window opened from the tray, reusing the ordinary window machinery: **pool promote first** (so it opens instantly from a pre-warmed window), cold-path fallback when the pool is empty. That is WS3's stated requirement, and the spec rejects both alternatives explicitly — a native OS menu is "too limited for agent chat", and a separate process would need its own srv connection, update path and crash supervision.

Adds `open_panel` (host IPC) and a **Show Panel** tray menu item wired to it.

## Design notes worth reviewing

- **Geometry is a pure function**, so the parts that actually break without a display are tested: anchoring to the **work area** rather than `(0,0)` — a top/left taskbar or a secondary monitor otherwise puts the panel under the taskbar or on the wrong screen — and clamping, so a small or heavily scaled display gets a shrunk panel rather than one whose bottom half is off-screen, and a mis-reported monitor cannot produce a zero-size window the user can neither see nor close.
- **The panel is a normal top-level window** to the rest of the host: it registers, counts toward `count_live_user_windows`, and keeps the instance attended. Deliberate — a panel the user has open is a window they are looking at, so it should satisfy the last-window quit gate and suppress the unattended audit period exactly like any other window.
- **Anchored bottom-right of the work area**, where the tray lives on the default Windows layout, so it appears next to the thing the user just clicked. A more precise answer would be "the monitor containing the clicked icon", but the click arrives in the launcher and the position is not plumbed through; the primary work area is correct on the common single-monitor case and always yields an on-screen window otherwise, which is the property that matters.

## Scope, stated honestly

This opens the **normal UI at panel dimensions**. A visually *simplified* panel view is frontend work (a new React surface) that cannot be designed or verified without a display, so it is deliberately not attempted here. The plumbing already accepts an `initial_view`, so a dedicated view drops in later without touching this file.

## Test plan

- [x] `cargo test -p agentmux-cef --lib --release` — **360 passed** (5 new panel-geometry tests)
- [x] `cargo test -p agentmux-launcher --release` — **231 passed** (menu-model test updated for the third entry)
- [x] Both crates build clean
- [ ] Not visually verified that the panel renders or is correctly placed — same display constraint as the rest of WS1/WS3. The geometry logic is unit-tested; the rendering is not.

### DIFF
--- a/.changesets/1788648231-feat-tray-small-companion-panel-window-opened-from-the-tray--uk9v.md
+++ b/.changesets/1788648231-feat-tray-small-companion-panel-window-opened-from-the-tray--uk9v.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(tray): small companion panel window opened from the tray (issue #2977 WS3)
--- a/agentmux-cef/src/commands/drag.rs
+++ b/agentmux-cef/src/commands/drag.rs
@@ -391,6 +391,7 @@ pub fn tear_off_pool_promote(
         tab_anchor_y,
         None,
         None,
+        false, // tab tear-off, not a tray panel
     ) {
         Some(label) => Ok(serde_json::json!(label)),
         None => {
--- a/agentmux-cef/src/commands/window/mod.rs
+++ b/agentmux-cef/src/commands/window/mod.rs
@@ -13,8 +13,10 @@
 // keep resolving `commands::window::<name>` unchanged.
 
 mod lifecycle;
+mod panel;
 // Cross-platform command handlers dispatched by ipc.rs.
 pub use lifecycle::{close_window, close_window_by_label, quit_app};
+pub use panel::open_panel;
 #[cfg(target_os = "windows")]
 pub use lifecycle::find_main_window;
 // Windows-only helpers other modules resolve as `commands::window::<name>`
--- a/agentmux-cef/src/commands/window/panel.rs
+++ b/agentmux-cef/src/commands/window/panel.rs
@@ -1,0 +1,193 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Tray panel — issue #2977 Workstream 3,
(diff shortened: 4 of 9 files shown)


a5af:
@codex review

Agent5-asaf:
All four fixed in 4180a0b. The headline P1 was squarely mine — I had read the `let _ = (width, height);` comment in `promote_pool_window_for_new_window` and called it anyway, so the feature produced a normal 1200x800 window in exactly the warm-pool case it is designed around.

Now calls `promote_pool_window` directly, which honors `Some(w)/Some(h)` via `to_physical()`. That means the values must be handed over as **logical** pixels, so `panel_rect` now takes the DPI scale, returns **physical** pixels (a scaled display gets a proportionally larger panel rather than one rendered at half size), and the logical size is derived back from the clamped physical rect. There is a round-trip test across 1.0/1.25/1.5/2.0 to keep that honest, plus one for a zero scale since `GetDpiForMonitor` can fail.

H.7: guard added at the top of `open_panel`. You are right that only the cold path was protected — it inherited the check from `open_window_with_kind` while the promote ran unguarded.

Always-on-top: also correct, and I missed it in the spec — §2 says "small, separate, **always-on-top** CEF window" verbatim. Implemented with CEF Views `Window::set_always_on_top` rather than a Win32 `HWND_TOPMOST` call, which is cross-platform and needs no HWND resolution — that also makes it work on the cold path, where no HWND exists at the moment the request is made.

On the remaining P2 (the promote-liveness fallback recovering as a normal window rather than a panel): confirmed real. It is not fixed here because the fix belongs in `PromoteFallback` in `window_pool.rs` — the generic new-window recovery path shipped in #2987 — rather than in the panel, and threading panel geometry through it is a change to that mechanism that deserves its own review rather than riding along. Flagging it explicitly rather than leaving it silent; happy to take it in a follow-up.

cef 363 passed.

PR #3004, opened 2026-09-06
fix(agent): swap remaining tail glyphs from weak-coverage arrow

Agent3-asaf:
## Summary
Follow-up to #2995 (`SPEC_ACTIVITY_DOCK_TITLE_WIDTH_AND_TAIL_GLYPH_2026_09_05.md` §3.1's flagged-but-untouched cases): the same `↳` (U+21B3) glyph, with the same weak font-coverage risk, also appeared in:
- `ToolBlock.tsx`'s tool-call live-tail
- `PaneRow.tsx`'s tail

Swapped both for `→` (U+2192), matching the dock/shell-block fix already shipped in #2995.

## Test plan
- [x] `npx vitest run` — PaneRow/ToolBlock suites (40 tests) pass unchanged (no test asserted the glyph itself; only a stale comment needed updating)

### DIFF
--- a/.changesets/1788664032-fix-agent-swap-the-remaining-tail-glyphs-from-the-weak-cover-bt9e.md
+++ b/.changesets/1788664032-fix-agent-swap-the-remaining-tail-glyphs-from-the-weak-cover-bt9e.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #3005, opened 2026-09-06
Loap #2@claudius: docs(report): record the second-group cleanup in the workspace audit

GenericAgentX-asaf:
## Summary

Follow-up to #2998. That PR auto-merged **before** the second half of the report was pushed, so `main` landed with the report still saying the non-AgentMux clones were "untouched and still open" — after they had in fact been cleaned. This corrects the record.

Adds §3.1 (per-repo findings) and flips §5 from "recommended sequence" to "what was executed".

Two findings worth having on the record:

- **`reagent` was 33 commits ahead of its own remote branch** — work that existed only on that disk. A stale `origin/…` ref reported it as pushed; it only surfaced after an explicit `git fetch`. Preserved to `salvage/agenty-docs-provider-update-2026-09-05` on `a5af/reagent` and SHA-verified against the remote before anything was deleted. That branch is AgentY's and needs triage by whoever owns it.
- **`dev-tools` is live tooling** — its `packages/secrets` CLI is how the GitHub token is read from Secrets Manager. Removing it was safe only because the `secrets` command on `PATH` turned out to be an independently-installed published package rather than an `npm link` into `C:\Systems`; verified by running it before *and* after removal. Had it been a link, that deletion would have broken credential retrieval machine-wide.

`SunoHarvester` was deliberately kept (third-party `hartmark/` repo with real untracked work — out of scope for a rule about where *agents* work).

## Test plan

- [x] Docs-only; single file, no code touched
- [x] Diff is purely additive/corrective against `main` (+75/−33) — the first half of the report already landed via #2998
- [x] Every claim re-verified against live state before writing (fetch-then-compare for reagent, shim resolution + before/after functional test for dev-tools)

### DIFF
--- a/docs/reports/REPORT_OUT_OF_WORKSPACE_CLONE_AUDIT_2026_09_05.md
+++ b/docs/reports/REPORT_OUT_OF_WORKSPACE_CLONE_AUDIT_2026_09_05.md
@@ -5,11 +5,12 @@
 **Scope:** This machine (`claudius`). Every git clone/worktree of an
 AgentMux-org repo, checked against the rule that an agent must work from a
 clone **inside its own workspace**.
-**Status:** Analysis complete; §2 cleanup **executed** 2026-09-05 after
-operator sign-off. The two §4 blockers were resolved by decision, not by
-rescue — the operator confirmed both documents were stale and not worth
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #3006, opened 2026-09-06
fix(host): publish ipc-port where the launcher reads it under task dev

Agent5-asaf:
## Summary

Found while running `task dev` to visually verify the tray from #2996 (issue #2977 WS1). The icon rendered fine, but its running-indicator was stuck at "not running" for the entire session — even though `srv` came up one second later.

The host picked the port-file directory by asking *"am I a dev build"*, on a premise stated in the comment itself:

> In dev mode there is no launcher so port forwarding isn't wired anyway

That premise is false. **`task dev` does run the launcher** — only `task dev:standalone` bypasses it. So under `task dev` the host wrote `<instance>/cef-cache/ipc-port-<hash>` while the launcher read `<instance>/data/ipc-port-<hash>`.

Both consumers of that file failed silently:
- `tray::windows::service_reachable` → the icon claims the service is down, forever. Which is a bad failure for this particular icon: WS4's whole justification for it is that it must be a *reliable* "is it actually running" indicator, and it was reliably wrong.
- `second_instance::forward_host_cmd` → the tray's **Open Window** / **Show Panel** actions cannot reach the host.

## The fix

The deciding factor is whether a **launcher** spawned us, not whether this is a dev build. `AGENTMUX_IPC_HASH` is set by the launcher and by nothing else (`supervisor/windows.rs`, `supervisor/unix.rs`), and whenever it is set, `AGENTMUX_DATA_DIR` came from that same launcher via `paths.common.to_env_vars()` — so it is that instance's own data dir, not another instance's.

**The hazard the old branch guarded against is preserved**, and covered by its own test: a dev build with *no* launcher still writes to its host-local dir, so it cannot clobber the port file of the parent AgentMux instance whose env it inherited (`dev:standalone`).

Extracted as a pure function so all five cases are covered without mutating process env in tests.

## Verification

Falsified rather than assumed — restoring the `is_dev_build`-only logic fails exactly one test:

```
a_dev_build_under_a_launcher_writes_where_the_launcher_reads
  left: "/instance/cef-cache"
 right: "/instance/data"
```

Those are the same two paths I observed on disk. The parent-clobber test still passes under the broken version, which confirms the two concerns are independently covered rather than one test doing both jobs.

**Also verified live, which is how it was found in the first place.** Before:

```
tray: icon created (service_reachable=false); entering message pump
   (…nothing further, for the whole session)
```

After:

```
tray: icon created (service_reachable=false); entering message pump
tray: service_reachable -> true
```

— one `STATUS_POLL` interval later, and the port file now lands in `<instance>/data/`.

`cef 384 passed`, `launcher 231 passed`.

## Note for #2977

This is a prerequisite for anyone verifying WS1's remaining (visual) acceptance criteria, since `task dev` is the natural way to do it and the indicator was misleading there. Separately, that same run produced the first runtime confirmation that the tray works at all on Windows: `tray: icon created (…); entering message pump` — the launcher has no Win32 message pump of its own, so this was the one genuinely unverified architectural assumption in the design.

### DIFF
--- a/.changesets/1788668658-fix-host-publish-the-ipc-port-file-where-the-launcher-reads--dtpk.md
+++ b/.changesets/1788668658-fix-host-publish-the-ipc-port-file-where-the-launcher-reads--dtpk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(host): publish the ipc-port file where the launcher reads it in dev mode
--- a/agentmux-cef/src/lib.rs
+++ b/agentmux-cef/src/lib.rs
@@ -1324,7 +1324,7 @@ pub unsafe extern "system" fn RunWinMain(
 ///
 /// This MUST agree with where the launcher reads it — `paths.data_dir`, used by
 /// `second_instance::forward_host_cmd` (single-instance window forwarding) and
-/// by `tray::windows::service_reachable` (the tray running-indicator). When the
+/// by `tray::windows::service_reachable` (the tray's running-indicator). When the
 /// two diverge the failure is silent: forwarding reports a transient read error
 /// and the tray icon simply claims the service is down forever.
 ///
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

Agent5-asaf:
Fixed in 7ea175d — comment-only, no behaviour change. 384 cef tests still pass.

Fair catch. The apostrophes went missing because my patch script stripped them to avoid escaping inside a Python string literal — a shortcut that saved nothing and cost readability in the one place where the reasoning most needs to be clear. Fixed all five occurrences, not just the line you flagged.

PR #2283, opened 2026-07-23
refactor: implement Tier 2 large-file modularization items

AgentX-asaf:
## Summary

Implements the "Tier 2 — real value, moderate care" items from `docs/reports/REPORT_LARGE_FILE_MODULARIZATION_SCAN_2026_07_22.md`. 12 commits, all pure structural extractions — no behavior changes, no external call sites broken (re-exports used wherever a file has real outside consumers).

| # | Item | Before → After |
|---|---|---|
| 1 | `subagent_watcher.rs` → `subagent_watcher/*` | 3717 → 7 files (largest split in the batch) |
| 2 | `subprocess.rs` → `subprocess/*` | 1770 → 5 files, plus a bonus dedup of a 4x-duplicated status-snapshot helper |
| 3 | `identity_handlers.rs` → `identity_auth_{spawn,dirs,persist}.rs` | 1711 → 4 files |
| 4 | `window_pool.rs` → `+ pane_pool.rs` | 2256 → 1688 + 599 |
| 5 | `browser_panes.rs` → `browser_panes/*` | 1480 → 5 files (zero-field struct, zero shared-state risk) |
| 6 | `launcher_ipc.rs` + `app.rs` → submodules | both split, ~40 + several external call sites needed zero changes |
| 7 | `commands/platform.rs` → `+ cli_login.rs` | 1946 → 672 + 1287 |
| 8 | `agent-model.ts` + `AgentLaunchModal.tsx` | 974→680, 987→844 |
| 9 | `identity-view.tsx` + `keymodel.ts` | split by real consumer boundaries, confirmed via grep |
| 10 | `action-widgets.tsx` + `markdown.tsx` | 687→363, 620→351 |
| 11 | `tabbar.tsx` → focused `tab-*.ts` modules | 1366 → 263 |
| 12 | `settings-view.tsx` + `browser-view.tsx` | 812→129, 797→197 |

## Test plan

- [x] `cargo build --workspace` — clean, only pre-existing warnings
- [x] `cargo test --workspace` — all pass (0 failed across every crate)
- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx vitest run` — 124/125 test files, 1904/1908 tests passed, 3 skipped; the 1 failure is a pre-existing flaky timeout (`tool-renderers/registry.test.ts`, resource contention under full-suite parallel load) confirmed passing in isolation both before and after this change

[...]

PR #3007, opened 2026-09-06
Posa@Area54: fix(swarm): subagents that returned are no longer reported as interrupted

GenericAgentX-asaf:
Every Agent-tool dispatch, for every agent, was displayed as **"interrupted"** in the Swarm pane — including ones that returned full results the parent went on to use.

Found while auditing a real session: the pane showed 4 Agent-tool rows for one agent, all "interrupted." All four had succeeded, days apart.

## Root cause

`SubAgentStatus::Completed` had exactly **one** writer (`jsonl.rs:243`), gated on a subagent's last event being a `Result`-typed JSONL line:

```rust
if matches!(&last.event_type, SubagentEventType::Result { .. }) {
    completed = true;
    state.info.status = SubAgentStatus::Completed;
}
```

**AgentMux transcripts contain no `"type":"result"` line at all** — not in subagent sidechain files, not in parent session files. So `Completed` was unreachable. Members stayed `Active`, and `reconcile_stale_subagents` then downgraded them to `Abandoned` on the documented reasoning that a file *"lacking a terminal `Result` line was interrupted"*. The file lacks one because **nothing writes one**.

`types.rs` already flags `Abandoned` as *"always an inference, never an observation"* — the observation it inferred from was unobtainable.

## Evidence

| Check | Result |
|---|---|
| Subagent transcripts with a `result` line | **0 of 11** (2 agents, portables 0.55.31→0.55.36) |
| Parent transcripts with a `result` line | **0** (15k-line census: `queue-operation/user/attachment/atis-latch/assistant/ai-title/last-prompt/mode/system`) |
| The 4 dispatches' actual outcome | `is_error=false` on all four; two returned 26–31 KB |
| CC 2.1.198 (Aug) vs 2.1.247 (now) | **both 0** — not a regression |

Every completed subagent transcript ends on an `assistant` message.

**How it slipped in:** the detector arrived in #2283, replacing a placeholder-text match its own comment says *"almost never fired."* The replacement fires **never** — and because both fail identically (rows that never complete), it read as the pre-existing flakiness rather than a new absolute.

## The fix

Completion is now an **observation**: the parent's own `tool_result` for the dispatch's `tool_use_id`. No new plumbing was needed for correlation — Claude Code already writes the key into the sidecar beside each transcript:

```json
{"agentType":"Explore","description":"Research AgentMux shutdown sequence…",
 "toolUseId":"toolu_01RY4RsXyFZAQJP82G7Zw4YX","spawnDepth":1,"model":"opus"}
```

New `completion.rs` resolves it; `reconcile_stale_subagents` consults that instead of unconditionally writing `Abandoned`.

### Three properties worth reviewing — each was a way to get this wrong

- **Fails closed.** No sidecar, unreadable/absent parent, uncorrelatable entry → falls back to the historical `Abandoned`. It can never *optimistically* complete something it couldn't verify.
- **The `sessions` lock is not held across the file IO.** Parent transcripts run to tens of MB, and this function is explicitly written to avoid holding that mutex across slow work. Resolution is two-phase: snapshot candidates under a short lock, read unlocked, apply. The apply pass **re-checks `Active`**, so a member whose status moved in between is left alone rather than clobbered by a stale decision.
- **One transcript read per pass, not per member.** All of a session's members share a parent.

Completed members emit the same `subagent:completed` event `jsonl.rs` already sends live, so an open Swarm pane gets the correction immediately rather than at the next unrelated reload.

### One deliberate scoping decision

**An errored result still counts as `Completed`.** The distinction being drawn is *returned* vs *cut off*, and a dispatch that reported an error returned. Surfacing failure as its own state needs a status the enum doesn't have plus a matching frontend union — left as explicit follow-up rather than smuggled into this fix. `is_error` is parsed and available for whoever picks it up.

## Testing

- **10 new unit tests**, fixtures taken from the real transcripts above rather than invented.
- **Mutation-checked:** reverting `terminal_status` to the old always-`Abandoned` behaviour fails exactly the two completion assertions, while the genuinely-abandoned and fallback cases keep passing — confirming they test the fix rather than restate it.
- Full `agentmux-srv` suite: **3017 passed, 0 failed**. Re-verified after rebasing onto latest `main`.

## Not verified end-to-end

The mechanism is proven at the unit level and targets the exact line that produced the wrong status, but **I have not watched the four rows flip from "interrupted" to completed in a live Swarm pane.** That needs a build and a pane reopen (reconcile runs on backfill). Flagging it rather than implying otherwise.

Reports with the full evidence chain:
- `docs/reports/REPORT_SUBAGENT_COMPLETION_NEVER_DETECTED_2026_09_05.md`
- `docs/reports/REPORT_SWARM_PANE_VS_ACTUAL_AGENT_CALLS_2026_09_05.md` (the audit that surfaced it)

### DIFF
--- a/.changesets/1788681796-fix-swarm-subagents-that-returned-are-no-longer-reported-as--4c4g.md
+++ b/.changesets/1788681796-fix-swarm-subagents-that-returned-are-no-longer-reported-as--4c4g.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(swarm): subagents that returned are no longer reported as interrupted
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -3027,10 +3027,31 @@ impl PersistentSubprocessController {
                         // `global()` is `None` in tests that don't call
                         // `subagent_watcher::set_global` — a safe no-op, same
                         // pattern `process_tracker::registry` already uses.
+                        //
+                        // On a BLOCKING thread, not this one. Reconciliation
+                        // used to be a pure in-memory status flip under a
+                        // mutex (O(1)), but as of the completion-detection
+                        // fix (#3007) it reads the parent transcript from
+                        // disk — routinely tens of MB, plus per-line JSON
+                        // parsing — to decide whether each member actually
+                        // returned. Doing that inline would block a tokio
+                        // worker thread for the whole read, and this fires on
+                        // EVERY turn-end, so several blocks finishing at once
+                        // could stall the runtime (reagent P1 on #3007).
+                        //
+                        // Fire-and-forget: the result is a status correction
+                        // broadcast to whoever's listening, not something this
+                        // stdout-reader task consumes, so there is nothing to
+                        // await. The backfill call site (`scan_session_
+                        // subagents`) needs no equivalent — it already runs in
+                        // a blocking context that walks up to 200 files.
                         let session_id_snapshot = inner_read.lock().unwrap().session_id.clone();
                         if let Some(sid) = session_id_snapshot {
                             if let Some(watcher) = subagent_watcher::global() {
-                                watcher.reconcile_stale_subagents(&block_id_read, &sid);
+                                let block_id = block_id_read.clone();
+                                tokio::task::spawn_blocking(move || {
+                                    watcher.reconcile_stale_subagents(&block_id, &sid);
+                                });
                             }
                         }
                     }
--- a/agentmux-srv/src/backend/subagent_watcher/completion.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/completion.rs
@@ -1,0 +1,260 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Did this subagent actually finish, or was it cut off?
+//!
+//! `SubAgentStatus::Abandoned` is — as its own doc comment says —  "always an
+//! inference, never an observation". Until this module existed, the inference
+//! was "the subagent transcript has no terminal `Result` line, therefore it was
+//! interrupted" (`scan.rs::reconcile_stale_subagents`). That premise does not
+//! hold: **no AgentMux transcript contains a `"type":"result"` line at all** —
+//! not the subagent sidechain files, not the parent session file. The line type
+//! the old check waited for is absent from the format this deployment writes
+//! (`entrypoint: sdk-cli`), so `SubAgentStatus::Completed` was unreachable and
+//! every Agent-tool dispatch ended up displayed as "interrupted", including the
+//! ones that returned full results the parent went on to use.
+//!
+//! Evidence and the full chain:
+//! `docs/reports/REPORT_SUBAGENT_COMPLETION_NEVER_DETECTED_2026_09_05.md`.
+//!
+//! The replacement uses the strongest evidence available, and it is an
+//! *observation* rather than an inference: the parent's own `tool_result` for
+//! the dispatch's `tool_use_id`. If the parent recorded a result, the subagent
+//! demonstrably returned — whatever its transcript does or doesn't end with.
+//! The correlation key is already on disk: Claude Code writes an
+//! `agent-<id>.meta.json` sidecar next to each subagent transcript carrying the
+//! parent-side `toolUseId`.
+//!
+//! Deliberately kept free of watcher state so the decision is unit-testable on
+//! its own — the same reasoning that split `parse.rs` out.
+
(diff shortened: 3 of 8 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in `d0fd533ea` — good catch, and it was worse than a stray comment.

The pre-existing doc block (batch-not-spam rationale + the #2235 frontend-listener history) had silently become documentation for `broadcast_subagent_completed`, while `broadcast_subagents_abandoned` — the function it was actually written about — was left with none. Moved the new function below it so the original comment reattaches, and gave the new one its own doc.

Worth noting what the accidental placement had obscured: **the new emitter is deliberately per-agent, not batched like its sibling.** Sitting under the batching rationale, it read as though it followed the same rule and simply ignored it. It doesn't — the payload carries per-member fields (`totalEvents`, `dispatchId`) a batched shape has nowhere to put, and matching `jsonl.rs`'s existing `subagent:completed` event exactly is worth more than saving a broadcast, since a reconcile pass resolves a handful of completions rather than the dozens that rationale was written for. That reasoning is now stated in the doc instead of being implied by position.

No behaviour change. 106/106 `subagent_watcher` tests pass; no new warnings in the touched files.

GenericAgentX-asaf:
Both fixed in `ebfc12507`. The P1 was correct on both counts, and the second half was mine and worse than the bug this PR set out to fix.

**(a) Workflow-tool members could never resolve a completion.** `parent_transcript_for` required `subagents` to be the *immediate* parent; members sit one level deeper at `subagents/workflows/<run-id>/`. Every one returned `None` and stayed permanently "interrupted" — the exact defect this PR exists to remove, left in place for that entire dispatch kind. Now walks up to the `subagents` ancestor rather than assuming depth.

**(b) The `first()` non-determinism.** Deriving one shared parent path from `candidates.first()` over a std `HashMap` meant that, combined with (a), a Workflow member landing first zeroed the result map and dragged every Solo member in that pass into `Abandoned` — including ones whose parent genuinely recorded a `tool_result`. Same input, different outcome per run. Each candidate now resolves its own parent, with reads cached per distinct path, so the optimisation the shared read was reaching for (one read per transcript, not per member) survives without the false assumption.

**Why the original tests missed it:** there is no workflow run anywhere on this dev machine — every fixture I had was a flat Task-tool shape, so the nested case was invisible to both the code and the tests. Added 2 tests for it: the nested `workflows/<run-id>/` path, and solo+workflow members of one session agreeing on the same parent. Mutation-checked — reverting to the immediate-parent check fails exactly those two.

**P2** — `SubAgentStatus::Abandoned`'s doc still called `reconcile_stale_subagents` a blanket downgrade that "never promotes", stale since this PR made that function write `Completed`. Updated to describe actual behaviour (it still never promotes back to `Active`). I also corrected one of my own comments the same change invalidated: parent transcripts are read once per *distinct* parent, no longer once per pass.

Full `agentmux-srv` suite: **3019 passed, 0 failed**.

GenericAgentX-asaf:

Both P1 halves and the P2 are addressed in `ebfc12507` (details in the comment above). Requesting a fresh review to clear the standing CHANGES_REQUESTED on the current head.

GenericAgentX-asaf:
Both fixed in `6be4b8743`.

**P1 — correct, and a real regression this PR introduced.** Reconciliation was a pure in-memory status flip under a mutex (O(1)); the completion fix made it read a multi-MB transcript from disk, and that same function fires live on *every* turn-end inside a `tokio::spawn` stdout-reader task. Several blocks' turns ending together could stall the runtime.

The live call site now goes through `tokio::task::spawn_blocking` — fire-and-forget, since the result is a status-correction broadcast rather than something that task consumes, so there's nothing to await.

I deliberately did **not** give the backfill call site (`scan_session_subagents`) the same treatment: it already runs in a blocking context that walks up to 200 files, so the transcript read is consistent with what's already happening there. Flagging that as a decision rather than an oversight.

**P2 — also correct, and worth more than a comment tweak.** The claim that `is_error` "is logged at the call site" was simply false: `terminal_status` only asks `contains_key`, and nothing reads the flag anywhere. That's exactly the kind of inaccuracy that sends whoever picks up the follow-up (surfacing failed dispatches) hunting for data that isn't exposed. Corrected to state what's true — it's parsed and carried in the map so the follow-up has it without re-reading every transcript, and is not surfaced or logged.

`subagent_watcher` 108/0, `blockcontroller` 253/0.

**One transparency note on the full-suite run:** `sysinfo::…::refresh_processes_specifics_does_not_leak_a_handle_per_call` fails on my machine. I checked before attributing it — it fails **3/3 with these changes stashed**, so it's a pre-existing environment-sensitive handle-count check (this box currently has 14 AgentMux processes up), not a regression from this branch. CI is the authority on it.

PR #3008, opened 2026-09-06
fix(tray): use the AgentMux brand icon instead of the placeholder mark

Agent5-asaf:
## Summary

The tray shipped a generated **orange rounded square** placeholder. Its comment said swapping in the real brand icon was *"a packaging task, not a code one"*, because *"the launcher has no icon asset it can rely on at runtime"*.

That premise was wrong, and wrong in a way worth recording: the launcher's **own `build.rs` already embeds** `agentmux-cef/resources/win/agentmux.ico` via winres, which emits `1 ICON "..."`. The brand icon has been in the exe's resource table all along — the same one Explorer and Alt-Tab show. Nothing needed packaging; it needed asking the exe for its own icon.

## What changed

- Load ordinal `1` with `Icon::from_resource` instead of generating a square. No new asset, no PNG decoder, no build-time RGBA blob — and the tray can't drift from the app icon, since both sides now read the same `.ico`.
- **Request the small-icon size explicitly.** Passing `None` maps to `LR_DEFAULTSIZE`, which resolves to `SM_CXICON` — the 32×32 *large* icon metric — leaving Windows to shrink a 32px frame into a 16px tray slot. `SM_CXSMICON`/`SM_CYSMICON` selects the frame the `.ico` already carries at that size and follows DPI scaling on its own.
- Keep the fallback square (a cosmetic icon must never take down the tray), but **log it loudly** and leave it deliberately off-brand. A silent fallback is indistinguishable from "the brand icon doesn't work", which is exactly the kind of thing that quietly eats an afternoon.

## Test

`build.rs` embeds the icon only `if icon_path.exists()` — a silent skip. If that asset ever moves, the exe loses its icon resource, `from_resource` fails, and the tray degrades to the fallback with nothing else in the build complaining. The new test asserts that path exists.

I want to be straight about its limits: that's the *only* half of the build.rs coupling a unit test can reach. The ordinal itself is decided by winres at build time, so the real check is the live one below.

## Verification

Ran on Windows via `task dev` with `AGENTMUX_TRAY=1 AGENTMUX_BACKGROUND_SERVICE=1`:

```
tray: started (windows)
tray: icon created (service_reachable=false); entering message pump
tray: service_reachable -> true
```

Zero `brand icon` fallback lines, and the icon renders as the blue pane mark rather than the orange square — confirmed visually by the repo owner, who reported the orange placeholder in the first place.

`launcher 232 passed`.

## Note for #2977

Between this and #3006, WS1's acceptance criterion ("a tray icon renders and responds to clicks") now has real evidence on Windows: the icon renders with the correct brand mark, and the launcher log recorded `tray: open_new_window forwarded (attempt 1)` from an actual click. Still Windows-only — macOS/Linux backends don't exist yet — so the box stays unticked until those land.

### DIFF
--- a/.changesets/1788682482-fix-tray-use-the-agentmux-brand-icon-instead-of-the-placehol-em0o.md
+++ b/.changesets/1788682482-fix-tray-use-the-agentmux-brand-icon-instead-of-the-placehol-em0o.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tray): use the AgentMux brand icon instead of the placeholder mark
--- a/agentmux-launcher/src/tray/windows.rs
+++ b/agentmux-launcher/src/tray/windows.rs
@@ -246,21 +246,83 @@ fn run(
     Ok(())
 }
 
-/// The icon bitmap.
+/// Resource ordinal of the brand icon inside this exe.
 ///
-/// Generated rather than loaded from disk: the launcher has no icon asset it
-/// can rely on at runtime (the exe's own icon is injected post-build by
-/// `scripts/inject-exe-icon.sh`, and is not readable as a file next to the
-/// binary in a portable install). A flat, recognizable mark avoids adding a
-/// packaging dependency for the prototype; replacing it with the real brand
-/// icon is a packaging task, not a code one.
+/// `agentmux-launcher/build.rs` hands `agentmux-cef/resources/win/agentmux.ico`
+/// to winres, which emits `1 ICON "..."` — ordinal 1. Kept as a named constant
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

Agent5-asaf:
Fixed in 918cd08 — comment-only, 232 launcher tests still pass.

Real catch. `brand_icon_is_embedded` was the name in an earlier draft; I renamed the test in the code and left the prose pointing at the old one. A doc pointer to a symbol that does not exist is worse than no pointer at all — it sends the next reader hunting for a guard that was never written, which is the opposite of what naming it was for.

PR #3009, opened 2026-09-06
Loap #2@claudius: fix(agent): hover peek panel now respects the agent pane zoom level

GenericAgentX-asaf:
## Summary

The hover peek panel (timestamp / token estimate / path) rendered at 100% regardless of the agent pane's zoom, while everything around it scaled.

**Cause:** the panel is `Portal`-rendered to `document.body` so it can escape the virtualized transcript rows' stacking contexts (each row carries `contain: layout` + a `transform`, so no `z-index` inside one row can out-rank a later sibling row). Escaping the subtree also escapes the pane's `zoom`, which `agent-view.tsx` sets on the pane root.

**Fix:** read the factor off the anchor row's inherited `--agent-pane-zoom` — already published by `agent-view.tsx` right next to the `zoom` it sets — and apply it to the panel. Reading from the row rather than threading a prop means no changes at any of the six call sites, and it automatically resolves to whichever pane the hovered entry belongs to (including the history view, which sets the same pair).

**The non-obvious half is the division.** CSS `zoom` also multiplies the element's *own* inset lengths, and every input here is an already-post-zoom `getBoundingClientRect()` value. So each length is pre-divided by the factor — otherwise the panel would fly off-screen at high zoom rather than merely being the wrong size. `transform: translateX(-100%)` is deliberately untouched: it resolves against the element's own border-box width, which scales with it.

A factor of exactly 1 emits no `zoom` property and performs no division, so the overwhelmingly common case is byte-identical to before.

Also corrects this file's header comment, which asserted the component was "already zoom-safe". That was true only of its **position** (`getBoundingClientRect` reports post-zoom pixels, so it always landed in the right place — at the wrong size). That half-truth is why the gap survived.

## Test plan

- [x] `npx vitest run PeekOverlay.test.tsx` — 14 passed (6 new), including the 8 pre-existing autoUpdate-lifecycle and mouse-Y-tracking tests unchanged
- [x] New coverage: zoom applied; geometry pre-divided so the panel still lands on the row's edge; the `stretch` variant de-scaled too; **no** `zoom` emitted and geometry untouched at 100%; graceful fallback when the variable is absent, malformed, or non-positive
- [ ] Live: zoom an agent pane in/out (Ctrl+wheel), hover a transcript entry — the peek panel should scale with the pane and stay pinned to the entry's right edge

Reported alongside a broader complaint that pane-level and preview-level zoom conflict; that's a separate, larger piece of work and I'm auditing it separately. This fix is self-contained.

### DIFF
--- a/.changesets/1788683074-fix-agent-hover-peek-panel-now-respects-the-agent-pane-zoom--cp99.md
+++ b/.changesets/1788683074-fix-agent-hover-peek-panel-now-respects-the-agent-pane-zoom--cp99.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): hover peek panel now respects the agent pane zoom level
--- a/frontend/app/view/agent/components/PeekOverlay.test.tsx
+++ b/frontend/app/view/agent/components/PeekOverlay.test.tsx
@@ -233,4 +233,121 @@ describe("PeekOverlay", () => {
             vi.useRealTimers();
         }
     });
+
+    // The panel is Portal-rendered to document.body, which escapes the agent
+    // pane's `zoom` — so it used to paint at 100% while the pane around it
+    // scaled. It now reads `--agent-pane-zoom` off the anchor row (the var
+    // inherits down from the pane root) and applies it to itself.
+    describe("agent pane zoom", () => {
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #3010, opened 2026-09-06
Loap #2@claudius: docs(report): zoom bindings audit — 13 binding sites across 5 systems

GenericAgentX-asaf:
## Summary

Audit of every zoom binding in the frontend, prompted by: *previews are janky to zoom, pane-level and preview-level are separate systems, and sometimes the preview level doesn't even work.*

**Direct answer: 13 binding sites a user gesture can reach, across 5 independent systems** (plus 3 duplicated platform copies of the core module and 3 compensation sites that exist purely to undo zoom applied elsewhere). 85 files in `frontend/` mention zoom at all.

## The headline finding

The two systems aren't two implementations of the same idea — they use **different CSS mechanisms with different capabilities**:

| | Mechanism | Scales hardcoded `px`? |
|---|---|---|
| Pane zoom | CSS `zoom` on the pane root | **Yes** — scales the whole box tree |
| Preview zoom | `font-size: N%` on one wrapper | **No** — only `em`/`%` descendants |

`_document-nodes.scss` carries **66 hardcoded `font-size: …px` rules against 8 relative ones**, so preview zoom can only ever move ~11% of the type in the panel. `DiffViewer`, `BashOutputViewer`, `HighlightedCode` and `CompactResult` never reference the scale at all.

That fully explains *"sometimes it doesn't even work"* — whether it appears to do anything depends on which renderer you're looking at.

**So the proposed simplification isn't a feature trade — it replaces a mechanism that can't work with one that already does.** CSS `zoom` on the pane root scales previews correctly today, hardcoded pixels included.

## Also documented

Three conflicts reachable with one gesture: the same Ctrl+wheel does different things a few pixels apart (an explicit `overPreviewBody` hit-test, with no visual cue where the boundary is); the two scales compose multiplicatively with different clamps (0.5–2.0 vs 0.7–2.0); and the preview scale lives in a component-local signal inside a **virtualized** transcript, so it silently resets on scroll — and per `ToolBlock`'s own comments, `<Index>` recycles instances across different nodes, so a stale scale can land on an unrelated tool block.

The recommendation names exactly what to delete, states plainly **what is genuinely lost** (independently enlarging a dense diff without the surrounding transcript — a capability that today works on ~11% of the text and resets on scroll), and flags that previews will *start* obeying pane zoom in places they didn't, which is the fix but is also a visible change worth calling out in a changeset rather than letting it read as a regression.

§4 lists optional consolidation follow-ups (six views re-implementing the same `term:zoom` logic; three near-identical platform copies; `help:zoom` as a working parallel copy) — deliberately **not** bundled with the main fix.

## Test plan

- [x] Docs-only; no code touched
- [x] Every count verified against the working tree at `f946cb730`, not inferred — the 66/8 split by `grep -c`, the `font-size: N%` application read at `ToolOverlayLog.tsx:344` and its prop chain traced back to `ToolBlock.tsx`
- [x] An initial survey mislabeled swarm as using the shared helpers; corrected by reading `swarm-view.tsx:30-71` directly. §5 records what was checked and how.

Related: #3009 fixes the hover peek panel's own zoom escape (a different Portal-escapes-`zoom` bug in the same area), and is independent of this proposal.

### DIFF
--- a/docs/reports/REPORT_OUT_OF_WORKSPACE_CLONE_AUDIT_2026_09_05.md
+++ b/docs/reports/REPORT_OUT_OF_WORKSPACE_CLONE_AUDIT_2026_09_05.md
@@ -5,7 +5,7 @@
 **Scope:** This machine (`claudius`). Every git clone/worktree of an
 AgentMux-org repo, checked against the rule that an agent must work from a
 clone **inside its own workspace**.
-**Status:** Complete. Both §2 and §3 cleanup **executed** 2026-09-05 after
+**Status:** implemented — both §2 and §3 cleanup **executed** 2026-09-05 after
 operator sign-off. `C:\Systems\` now holds exactly one git repo — a
 third-party project unrelated to AgentMux (§3.2) — and zero
 out-of-workspace clones of any AgentMux-org repo. The two §4 blocker
--- a/docs/reports/REPORT_ZOOM_BINDINGS_AUDIT_2026_09_06.md
+++ b/docs/reports/REPORT_ZOOM_BINDINGS_AUDIT_2026_09_06.md
@@ -1,0 +1,213 @@
+# REPORT — Zoom bindings audit, and the case for collapsing preview zoom into pane zoom
+
+**Date:** 2026-09-06
+**Author:** Loap #2 @ claudius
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #3011, opened 2026-09-06
Manoz@Area54: docs: memory carry-over analysis across new/existing/compacted agent sessions

GenericAgentX-asaf:
## Summary

Traces how memory actually reaches an agent, for the three cases the repo owner named: **creating a new agent**, **loading an existing one**, and **after compaction**.

Measured against live data on this machine rather than reasoned from the code alone. **Two findings invert the assumptions the investigation started from**, so the design section was rewritten around them.

## Key findings

**1. Compaction is not where memory is lost.** Real Claude Code transcripts show the `compact_boundary` record followed by a `type:"user"` message carrying the continuation summary — compaction rewrites the **message list**. `CLAUDE.md` occurs **zero times in those transcripts, before or after**: it is delivered in the system prompt, which is rebuilt per request, so compaction structurally cannot remove it. Both memory paths survive. Independently corroborated by `TOKEN_TAX_ANALYSIS_2026_06_19.md:111`, whose table records CLI auto-memory loading at "Fresh session + post-compact".

**2. The memory AgentMux composes is empty — for every agent.** `db_agent_content` holds only `env`, `startup`, `ui:color`, `ui:zoom` across all 8 agents; there are **no `soul`, `agentmd`, or `memory` rows at all**, and zero bundles have `is_global=1`. End-to-end proof: `.claude/AGENTMUX_MEMORY.md` is **byte-identical at 1089 bytes across three separate agents** (`manoz`, `agenta`, `posa`), skills index only, no `# Memory` section. The pipeline works; it has nothing to carry.

**3. Storage scope, measured because it is easy to get backwards.** Transcripts, native memory, and agent definitions are all **global** — the per-channel memory path shares an **inode** with the shared one (same file through a link, not a copy that drifts). Per-channel `db_bundles` is the lone exception, which makes it the direct root cause of "memory doesn't carry over" on a fresh portable build.

**4. Two mechanism constraints resolved.** The `PreCompact` hook **cannot inject** — its own contract is exit 0 with no stdout at all. And `compact_boundary` exists only in `translator/claude.rs`, so anything keyed on it is Claude-only.

## Correcting a stale decision record

`docs/analysis/TOKEN_TAX_FOLLOWUP_2026_07_04.md:16` marks this question **"Already resolved"** on the reasoning that "AgentMux owns memory itself… the two-memory-systems risk doesn't exist today." Both halves are now false: the `autoMemoryEnabled` guard was narrowed to shared workdirs only (`agent_open.rs:731`), so CLI auto-memory runs for normal agents, while AgentMux's own memory is the empty one. The roles are inverted relative to the written decision. §4.2 asks for that record to be corrected.

## On the post-compaction digest

The repo owner selected "re-inject a memory digest turn" for case 3 — before findings 1 and 2 existed. §4.5 records that choice, the evidence that it would now be **redundant** (it would re-deliver what the model already has in its system prompt), and three options with a recommendation, rather than silently building or silently dropping it.

## Verification

- `scripts/check-doc-status.sh` / `check-spec-citations.sh` — both pass
- `docs/specs/INDEX.md` regenerated

Docs-only, nothing implemented. §7 carries three **decisions**, not unknowns — the highest-leverage being whether `db_bundles` should be globalized, which would fix cases 1 and 2 outright.

### DIFF
--- a/.changesets/1788683053-docs-memory-carry-over-analysis-across-new-existing-compacte-86rf.md
+++ b/.changesets/1788683053-docs-memory-carry-over-analysis-across-new-existing-compacte-86rf.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: memory carry-over analysis across new/existing/compacted agent sessions
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -337,7 +337,7 @@ partial list.
 | [`SPEC_MUXBUS_MULTI_TIER_DISCOVERY_AND_REMOTE_INVOCATION_2026_07_29`](SPEC_MUXBUS_MULTI_TIER_DISCOVERY_AND_REMOTE_INVOCATION_2026_07_29.md) | Spec: multi-tier discovery + remote API invocation over muxbus |
 | [`SPEC_WINDOW_NAME_API_HARDENING_2026_08_08`](SPEC_WINDOW_NAME_API_HARDENING_2026_08_08.md) | SPEC: Window-name App API hardening (phantom-id success + status codes) |
 
-### proposed (95)
+### proposed (96)
 
 | Spec | Title |
 |---|---|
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #3012, opened 2026-09-06
Loap #2@claudius: fix(agent): previews zoom with the pane instead of their own broken font-scale

GenericAgentX-asaf:
## Summary

Deletes the tool preview's separate Ctrl+wheel zoom. Ctrl+wheel over a preview now zooms the pane, like everywhere else.

Implements §3 of the audit in #3010.

**This makes previews zoom *more*, not less** — which is the counterintuitive part worth reading before reviewing. The two systems used different CSS mechanisms:

| | Mechanism | Scales hardcoded `px`? |
|---|---|---|
| Pane zoom | CSS `zoom` on the pane root | **Yes** — scales the whole box tree |
| Preview zoom (deleted) | `font-size: N%` on one wrapper | **No** — only `em`/`%` descendants |

`_document-nodes.scss` carries **66 hardcoded `font-size: …px` rules against 8 relative ones**, so the preview gesture could only ever move ~11% of the type in the panel — and `DiffViewer`, `BashOutputViewer`, `HighlightedCode` and `CompactResult` never referenced the scale at all. That's the reported *"sometimes it doesn't even work"*: whether it did anything depended on which renderer you happened to be looking at.

It also fought the pane three ways:
1. The same Ctrl+wheel did different things a few pixels apart (an `overPreviewBody` hit-test, no visual cue where the boundary was).
2. The two scales composed multiplicatively under different clamps (0.5–2.0 vs 0.7–2.0).
3. The scale lived in a component-local signal inside a **virtualized** list — so it silently reset on scroll, and per `ToolBlock`'s own comments `<Index>` recycles instances across nodes, so a stale scale could land on an unrelated tool block.

## Deliberately kept

`ToolOverlayLog`'s `if (e.ctrlKey) return` scroll-chain guard. Its old rationale ("leave the preview zoom alone") is gone, but the guard is still load-bearing: without it, a Ctrl+wheel at that box's scroll boundary would now zoom the pane **and** scroll it at the same time. Comment rewritten rather than deleted.

## What is genuinely lost

Enlarging a dense diff without enlarging the surrounding transcript. Worth naming plainly rather than pretending the feature was pure cost — though in practice it worked on ~11% of the panel's text and reset on scroll. If it's wanted back, the correct build is CSS `zoom` on the preview container plus block-meta persistence, mirroring the pane system.

**Behavior change to expect:** previews will *start* obeying pane zoom in places they previously didn't. That's the fix, but it's visible on first launch — flagged here and in the changeset so it doesn't read as a regression.

## Test plan

- [x] `npx vitest run frontend/app/view/agent/` — 1787 passed / 132 files, 0 failures
- [x] `ToolBlock.test.tsx` + `ToolOverlayLog.test.tsx` specifically — 44 passed
- [x] `npx tsc --noEmit` — clean (also removed `onMount`/`onCleanup` from `ToolBlock`'s imports, now unused)
- [x] No `previewFontScale` / `PREVIEW_ZOOM_*` / `fontScale` references remain anywhere in `frontend/`
- [ ] Live: Ctrl+wheel over a tool preview body, over its filename header, and over plain transcript text — all three should now zoom the pane identically, and the preview content should scale along with it

### DIFF
--- a/.changesets/1788686287-fix-agent-previews-now-zoom-with-the-pane-instead-of-their-o-apy2.md
+++ b/.changesets/1788686287-fix-agent-previews-now-zoom-with-the-pane-instead-of-their-o-apy2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): previews now zoom with the pane instead of their own broken font-scale
--- a/frontend/app/view/agent/components/ToolBlock.tsx
+++ b/frontend/app/view/agent/components/ToolBlock.tsx
@@ -42,7 +42,7 @@ import { useTick } from "@/app/hook/useTick";
 import { estimateTokenCount, formatCompactNumber } from "@/util/format-count";
 import { formatExactTime, formatTimeAgo } from "@/util/format-time";
 import clsx from "clsx";
-import { Show, createEffect, createMemo, createSignal, onCleanup, onMount, type JSX } from "solid-js";
+import { Show, createEffect, createMemo, createSignal, type JSX } from "solid-js";
 import { extractToolDetail } from "../stream-parser";
 import { useNodePeek } from "../hooks/useNodePeek";
 import type { AgentDispatch } from "../../swarm/swarm-model";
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #3013, opened 2026-09-06
fix(tray): rename Open AgentMux to New Window, drop the panel entry

Agent5-asaf:
## Summary

Repo owner's call after using the tray for real.

**"Open AgentMux" → "New Window".** The label was wrong precisely in the state where you read it: the icon is only in the tray when AgentMux is *up*, so the menu was offering to open something you can plainly see is already open. It reads as a no-op. What the item actually does is make a new window, so it now says that.

**"Show Panel" removed.** Not reachable from the tray for now.

**"Quit AgentMux" unchanged.**

## Details worth flagging

**The not-running label still differs.** `OpenWindow`'s label is state-dependent, and "New Window" would promise something that cannot happen when the service is unreachable — so that case keeps **"Start AgentMux"**. The menu still changes with running state, which is the WS4 requirement that the icon *and* its menu be honest about whether the service is up. Only the running-state wording changed.

**Nothing is deleted from the panel feature.** `open_panel` remains a host IPC command with its geometry tests (issue #2977 WS3, #3002) — it is simply not wired to a tray entry. I removed the `TrayAction::ShowPanel` variant and its dispatch arm along with the menu item rather than leaving them behind: unused, they would draw a dead-code warning and, worse, read to the next person as a live path that just doesn't work. Restoring the entry is a three-line change against this diff.

## Tests

Updated, not deleted:
- the ordering test now asserts **two** items, so an accidentally reintroduced entry still fails rather than silently passing a `len()`-free check
- the label test pins the new wording

`launcher 232 passed`.

### DIFF
--- a/.changesets/1788686547-fix-tray-rename-open-agentmux-to-new-window-and-drop-the-pan-wien.md
+++ b/.changesets/1788686547-fix-tray-rename-open-agentmux-to-new-window-and-drop-the-pan-wien.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(tray): rename Open AgentMux to New Window and drop the panel entry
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #3016, opened 2026-09-06
docs(settings): surface network:lan_discovery in the template + report on toggling without restart

Agent2-asaf:
## 1. Template change (the small half)

`settings-template.jsonc` gains a **Network** section for `network:lan_discovery`, **commented out so the default stays `false`**.

Rationale (operator's): enabling it makes the app advertise/browse on the local network, which triggers the macOS local-network prompt / Windows firewall prompt — not worth bothering someone who only runs agents on one machine.

The key previously appeared **nowhere** in the template, so there was nothing to uncomment and no way to discover it existed. The comment states why it's off by default and that it must be enabled on **both** machines.

## 2. Report: making the toggle work without a restart

`docs/reports/REPORT_LAN_TOGGLE_WITHOUT_RESTART_2026_09_06.md`

**Today the toggle is half-live — the worst possible shape.** The mDNS daemon starts immediately (`LanDiscoveryController::apply`, live from `setconfig` — verified in a running instance's log), but `bind_addr` is resolved **once at startup** (`bootstrap.rs:1321`). So the instance advertises itself and gets discovered while its listeners are still on `127.0.0.1`. Peers discover it, then fail to deliver.

Confirmed live: LAN discovery on since 08:16, `netstat` still shows `127.0.0.1`-only.

`ON→OFF` is wrong too, and arguably worse: mDNS stops but listeners **stay on `0.0.0.0`**, still LAN-reachable until restart.

### Why cloud login differs (the stated reference point)

Cloud is an **outbound dial** — enabling it just starts a task. LAN is **inbound** and needs a listening socket whose address is fixed at `bind()` time. Structural asymmetry, not an oversight.

### The key insight

The in-code deferral says *"rebinding an active axum server is non-trivial."* True — but **rebinding isn't required.** `0.0.0.0:PORT` conflicts with `127.0.0.1:PORT`, but `<lan-ip>:PORT` does not. And `Router` is already `Clone`, served on two listeners today.

So: **add** listeners on the LAN addresses at the same port on toggle-on, drop them on toggle-off. No rebind, no port change (ports are ephemeral `:0` and already published in the mDNS TXT record and `authkey.dev`), loopback never interrupted — and it fixes both toggle directions, including the ON→OFF exposure bug, for free.

### Explicitly rejected

Always binding `0.0.0.0` and gating in middleware is the least code, but it would trigger the OS permission prompt for **every** user regardless of the setting — directly contradicting the decision to default it off precisely to avoid that. Rejected on that basis.

### Flagged for verification before building

The load-bearing assumption — simultaneous `127.0.0.1:PORT` + `<lan-ip>:PORT` binding — is asserted from general socket behaviour and **not tested here**. The report says to prove it with a throwaway binary on all three platforms first, Windows `SO_EXCLUSIVEADDRUSE` especially. I'd rather flag that than have someone build on it.

## Stopgap worth noting

Until this ships, the honest statement is *"enabling LAN discovery requires restarting the instance before other machines can reach it."* That's currently documented **only in a Rust source comment**, where no user will ever see it.

Design/docs only — no behaviour change. All 45 `wconfig` tests pass; template still parses.

### DIFF
--- a/.changesets/1788711237-docs-settings-surface-network-lan-discovery-in-the-settings--8r1j.md
+++ b/.changesets/1788711237-docs-settings-surface-network-lan-discovery-in-the-settings--8r1j.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(settings): surface network:lan_discovery in the settings template (default off)
--- a/docs/reports/REPORT_LAN_TOGGLE_WITHOUT_RESTART_2026_09_06.md
+++ b/docs/reports/REPORT_LAN_TOGGLE_WITHOUT_RESTART_2026_09_06.md
@@ -1,0 +1,228 @@
+# Report: Making LAN discovery work on toggle, without a sidecar restart
+
+**Date:** 2026-09-06
+**Status:** Draft — options analysis + recommendation. Not implemented.
+**Author:** Agent2
+**Goal (operator):** *"we need a way around that. figure out how we can get the
+network working without needing a restart. we want it to work similar to muxbus
+cloud login, it works right after enabling it."*
+
+---
+
+## 1. What actually happens today
+
+Toggling `network:lan_discovery` is **half-live**. Verified on a running
+instance, not inferred:
+
+| Component | Live on toggle? | Evidence |
+|---|---|---|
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #3017, opened 2026-09-06
perf(lan): query LAN peers concurrently instead of sequentially

Agent2-asaf:
**Step 1 of the network work.** Also lands the design report the rest of the series follows.

## The problem

Both LAN lookups — `find_agent` ("which peer hosts this agent") and `find_agent_lan_pubkey` ("that agent's Ed25519 key") — walked peers one at a time:

```rust
for peer in &peers { ... .timeout(LAN_PEER_QUERY_TIMEOUT_SECS).send().await; }
```

With a 2s timeout that's **2 × peer_count seconds worst case** before a message is delivered or falls through — 10s across five peers. And the pathological ordering (a dead or slow peer ahead of the right one) is the *common* case on a laptop network, where stale peers linger until their mDNS TTL expires.

## The change

Both now fan out concurrently through one shared `query_peers_concurrently`. Worst case becomes ~a single timeout regardless of peer count.

It returns a `FuturesUnordered` **stream** rather than a first-success-wins helper, because the two callers have genuinely different acceptance rules: `find_agent` takes any 2xx, while `find_agent_lan_pubkey` must keep going past a 2xx whose body has no `lan_public_key` (that peer hosts the agent but hasn't minted a key). Collapsing them into one "first success" helper would have silently broken the pubkey case — worth flagging since that's the obvious refactor and it's wrong.

Dropping the stream on the first accepted answer **cancels** the in-flight requests, so a win stops the remaining work rather than just ignoring it.

**Deliberately preserved:** the peer-eligibility rule (skip peers with no address or no scoped `lan_key`), and both positive *and* negative caching at the same TTL — the negative entry is what keeps cloud-only agents from re-fanning-out on every inject.

Side benefit: removes the duplicated query construction between the two lookups — the same copy-paste shape as the tier-forward blocks the report documents.

## Tests

- `a_slow_peer_does_not_delay_a_fast_one` — puts a 7s peer **first** and a fast peer second, asserts the fast one wins in <1500ms. That ordering is precisely what the sequential loop failed. The bound sits clear of both the ~0ms concurrent path and the ≥2000ms sequential one, so it discriminates without being timing-fragile.
- `ineligible_peers_are_skipped` — pins the eligibility rule.
- Mock peers reuse the raw-TCP pattern already in `muxbus::pkce`'s tests; no new dev-dependency.

## Also included

`docs/reports/REPORT_NETWORK_ARCHITECTURE_DRYNESS_AND_ROBUST_LAN_2026_09_06.md` — the design doc for the series:

- **Part I:** the robust LAN-toggle design — an interface-aware `LanListenerSupervisor` that reconciles listeners against *both* the setting and live interface-change events, rather than snapshotting interfaces at toggle time (DHCP/Wi-Fi↔Ethernet/VPN churn makes a snapshot stale immediately). Preserves the ephemeral port, never interrupts loopback, fixes the ON→OFF exposure bug.
- **Part II:** DRYness — three copy-pasted tier-forward blocks (`reactive.rs:747/:876/:982`), `echo_jekt_to_sender` called with 12 positional args including three consecutive `Option<bool>`s, tier 4 existing only in a comment, and discovery reporting local truth as global.

The report flags one untested load-bearing assumption (simultaneous `127.0.0.1:PORT` + `<lan-ip>:PORT` bind) that gates Part I — the spike for that is the next PR.

## Verification

`cargo check` clean · `cargo clippy` clean on the changed file · 28 existing `lan_discovery` tests pass (1 pre-existing ignored) · 2 new tests pass.

### DIFF
--- a/.changesets/1788712277-perf-lan-query-lan-peers-concurrently-instead-of-sequentiall-a696.md
+++ b/.changesets/1788712277-perf-lan-query-lan-peers-concurrently-instead-of-sequentiall-a696.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+perf(lan): query LAN peers concurrently instead of sequentially
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -147,6 +147,56 @@ struct LanCacheEntry {
     expires: std::time::Instant,
 }
 
+/// One in-flight `GET /agentmux/reactive/agent?id=…` against a single peer.
+type PeerQueryOutcome = (String, String, Result<reqwest::Response, reqwest::Error>);
+
+/// Query EVERY eligible LAN peer concurrently for `agent_id`, yielding results
+/// as they arrive.
+///
+/// Replaces the sequential `for peer in &peers { … .await }` both lookups used
+/// to run. That loop cost up to `LAN_PEER_QUERY_TIMEOUT_SECS × peer_count`
+/// before a message could be delivered — 10s across five peers — and the
+/// pathological ordering (a dead or slow peer ahead of the right one) is the
+/// common case on a laptop network where stale peers linger until their mDNS
+/// TTL expires. Fanning out makes the worst case ~one timeout regardless of
+/// peer count.
+///
+/// Returns a stream so each caller keeps its OWN acceptance rule: `find_agent`
+/// accepts any 2xx, while `find_agent_lan_pubkey` must keep trying after a 2xx
+/// whose body carries no `lan_public_key` (that peer hosts the agent but hasn't
+/// minted a key). A shared "first success wins" helper could not express both.
+///
+/// **Dropping the returned stream cancels every still-in-flight request** —
+/// so a caller that `break`s on the first acceptable answer stops paying for
+/// the rest, rather than merely ignoring them.
+fn query_peers_concurrently<'a>(
+    peers: &'a [LanInstance],
+    agent_id: &'a str,
+    http: &'a reqwest::Client,
+) -> futures_util::stream::FuturesUnordered<impl std::future::Future<Output = PeerQueryOutcome> + 'a> {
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

Agent2-asaf:
**Added step 2 (the bind spike) to this PR** — because it disproved a claim the report in this same PR states as fact, and shipping that uncorrected would be worse than resetting the approval. Sorry for the churn.

**The load-bearing assumption holds:** `127.0.0.1:PORT` and `<lan-ip>:PORT` bind simultaneously — verified on Windows, asserted in CI on Linux. So the "add a listener rather than rebind" design is sound.

**But one claim was wrong.** The report said `0.0.0.0:PORT` universally conflicts with an existing loopback bind. True on Linux/macOS; **false on Windows**, where absent `SO_EXCLUSIVEADDRUSE` the wildcard binds happily alongside it. I wrote the test asserting the conflict, it failed on Windows, and that's how I caught it — before anything was built on it.

The correction *strengthens* the chosen design: "just bind the wildcard on toggle" is now ruled out as a portable shortcut, since on Windows it would leave two sockets on one port with ambiguous accept behaviour — the exact hijack hazard `SO_EXCLUSIVEADDRUSE` exists to prevent. Binding **specific** addresses has no such ambiguity anywhere.

`wildcard_vs_loopback_conflict_is_platform_dependent` now pins the observed behaviour per-platform and asserts nothing universal, so it fails loudly if either side changes. The report's §2 and provenance both record this as measured-and-disproved rather than quietly fixed — the "standard socket behaviour" reasoning that produced the bad claim is worth leaving visible.

`primary_lan_ipv4` carries an `#[allow(dead_code)]` with a note: it's the primitive the supervisor will call, landed alongside the spike so the assumption and its dependent helper get reviewed together. Happy to split it out if you'd rather.

Verification unchanged otherwise: `cargo check` clean, clippy clean on both changed files, 3 new tests + 30 `lan_discovery` tests passing.

PR #3018, opened 2026-09-06
fix(wrr): stop the quit watchdog killing a resting background service

Agent5-asaf:
## Summary

Found by finally running WS0's deferred acceptance test — *"closing the last window with tray mode on leaves `srv` and `launcher` running, verified via `tasklist`"*. They did not. The whole instance exited ~9s after the last window closed.

```
[wrr] quit watchdog fired: 0 visible for 3000ms (retries=3) but reducer
disagrees (registered=21 draining=false) — quitting on OS signal alone
```

`is_background_service_resting` required `registered == 0`, so the stand-down only applied once the reducer's count had caught up with the OS. Closing 21 windows at once, it had not; the three 3s retries expired; and the watchdog force-quit exactly the process this mode exists to keep alive.

Tying a "should we quit" decision to whether a count has settled makes the guarantee **timing-dependent**, which is not a guarantee.

## The fix

`registered` is no longer considered. The old doc argued that `registered > 0` *"is still a real desync regardless of this mode"* — which is true, but the disagreement is about the **remedy**, not the diagnosis. Killing the process is only a sane remedy when the process is supposed to be exiting. Nobody asked it to (`!draining`), and the user is running a background service precisely so it survives having no windows. So the watchdog stands down and logs the desync at `warn` instead of acting on it.

`!draining` is still required, so an explicit quit (tray **Quit AgentMux** → `quit_app` → `LauncherRequested`) that wedges stays bounded — that is the one case where forcing the loop closed is right.

## About the test that should have caught this

I had asserted the buggy behaviour:

```rust
assert!(is_reducer_lagging_os(true, false, 0, 1, true));
```

It passed for exactly as long as the bug existed, because it encoded my assumption rather than the requirement. Replaced with `a_lagging_reducer_does_not_quit_a_background_service_nobody_asked_to_stop`, which pins the observed `registered=21` case and asserts count-independence.

## Verification

- **Falsified**: restoring the old form fails all three background-service tests.
- **Live** — the only reason this was found at all. Closing every window now leaves host, launcher and srv running (checked at 25s; the old build was gone by 9s):

```
closed: 1
visible windows left: 0
host STILL ALIVE: 29
launcher STILL ALIVE: 1
srv STILL ALIVE: 4
```

`cef 389 passed`.

## Note for #2977

WS0's second checkbox claims this behaviour and is ticked, with a note that the live `tasklist` check was deferred in favour of unit tests. That deferral is what let this ship: the unit tests all passed, including the one asserting the wrong thing. I'll update the issue after this merges.

### DIFF
--- a/.changesets/1788712485-fix-wrr-stop-the-quit-watchdog-killing-a-resting-background--ufss.md
+++ b/.changesets/1788712485-fix-wrr-stop-the-quit-watchdog-killing-a-resting-background--ufss.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(wrr): stop the quit watchdog killing a resting background service
--- a/agentmux-cef/src/wrr/win_event.rs
+++ b/agentmux-cef/src/wrr/win_event.rs
@@ -386,17 +386,35 @@ fn should_quit_on_last_window(armed: bool, draining: bool, visible: usize, regis
 }
 
 /// Workstream 0 Phase 1 (`SPEC_TRAY_OPTIONAL_BACKGROUND_SERVICE_2026_09_04.md`
-/// §7): true iff `registered == 0 && !draining` is the INTENTIONAL
-/// background-service resting state, not a desync. With background-service
-/// mode on, `should_begin_drain` (reducer/quit.rs) permanently declines to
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #3019, opened 2026-09-06
fix(window): centre a new window on the main display when none is open

Agent5-asaf:
## Summary

Reported by the repo owner: the tray's **New Window** opened flush in the top-left corner.

`get_offset_position` cascaded `+30,+30` from an existing window, and when there was none it returned `CW_USEDEFAULT`. That *is* the correct sentinel for `CreateWindow`, which reads it as "you pick" — but **nothing on this path hands it to `CreateWindow`**. It travels on as a literal coordinate: `promote_pool_window` takes it as a physical-pixel `SetWindowPos` anchor, and `clamp_rect_within` then pulls `0x80000000` back inside the work area, landing the window flush in the corner.

Harmless for as long as some window was always already open, because the cascade branch ran instead. Background-service mode made "no window open" the normal state and the tray's New Window routinely the first one — so a latent sentinel-misuse became the common path.

## The fix

`new_window_origin(win_w, win_h)`: cascade when there is a window, otherwise centre on the primary monitor's work area.

The size passed in has to be the size the window will **really** be. The Windows pool path promotes at `POOL_WIDTH`/`POOL_HEIGHT` and discards the size its caller computed, so passing the caller's 70%-of-work-area figure would offset the window by half the difference. The two call sites deliberately pass different dimensions; the comments say why.

`center_in_work_area` is kept pure and platform-independent so the arithmetic is tested on CI's Linux runners too, not just Windows.

## Verification

**Falsified** — reverting to the corner behaviour fails three of the four new tests, the first with `left: (0, 0)` / `right: (360, 120)`, the reported bug exactly. The oversized-window test still passes when broken, which confirms it guards a genuinely different property (clamping at the work origin rather than going negative) instead of duplicating the others.

**Live**, at 3840×2100 physical / 1.25 scale:

| path | rect | expected centre |
|---|---|---|
| cold (startup window) | `x=576 y=315 w=2688 h=1470` | (3840−2688)/2=576, (2100−1470)/2=315 ✓ |
| pool (tray New Window) | `x=1320 y=650 w=1200 h=800` | (3840−1200)/2=1320, (2100−800)/2=650 ✓ |

And with a window already open, the pool path still lands at the cascade offset (+30,+30 of the existing window) — unchanged, which is the behaviour I did *not* want to disturb.

> Note: my first live measurement looked wrong until I realised PowerShell was reading DPI-virtualised coordinates. The numbers above are physical, read from a DPI-aware process.

`cef 389 passed`.

### DIFF
--- a/.changesets/1788712534-fix-window-centre-a-new-window-on-the-main-display-when-none-txyw.md
+++ b/.changesets/1788712534-fix-window-centre-a-new-window-on-the-main-display-when-none-txyw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(window): centre a new window on the main display when none is open
--- a/agentmux-cef/src/commands/window/creation.rs
+++ b/agentmux-cef/src/commands/window/creation.rs
@@ -261,8 +261,22 @@ pub fn open_new_window(state: &Arc<AppState>, args: &serde_json::Value) -> Resul
     // On macOS/Linux, emits pool:new-window (no workspaceId → fresh workspace).
     // On Windows, delegates to promote_pool_window with workspace_id="" and physical-pixel
     // anchor (bypassing DIP→physical DPI double-conversion; see promote_pool_window_for_new_window).
-    let (pos_x, pos_y) = get_offset_position();
-    let (win_w, win_h) = get_secondary_window_size(pos_x, pos_y);
+    // Probe the monitor from the cascade anchor when we have one, else the
+    // primary — `get_secondary_window_size` only uses the point to pick a
+    // monitor.
+    let (probe_x, probe_y) = cascade_anchor().unwrap_or((0, 0));
+    let (win_w, win_h) = get_secondary_window_size(probe_x, probe_y);
+    // Centre the dimensions the window will ACTUALLY have. On Windows the pool
+    // promote below discards `win_w`/`win_h` and uses POOL_WIDTH/POOL_HEIGHT
+    // (see `promote_pool_window_for_new_window`), so centring `win_w`/`win_h`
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

Agent5-asaf:
Fixed in 517e843 — comment-only, 388 cef tests still pass.

You flagged the one in `creation.rs`; the same rename left a second stale reference in `window_pool.rs:1889` that the diff does not touch, so I fixed both.

This is the second time this session I have shipped a doc pointer to a symbol I had just renamed (the other was #3008). A `grep` for the old name after a rename is the obvious guard and costs nothing — I did not run it either time. Doing that from now on rather than relying on review to catch it.

Agent5-asaf:
Fixed in a419efd — comment-only, 388 cef tests pass.

You are right that it is the same class I already fixed once in this PR, and the detail is worth recording: I *did* grep after the rename, but only for `get_offset_position` inside the function bodies I had edited. This leftover is the doc line one above where I was looking, so it came back clean. The patch anchored on the `fn` line and replaced downward, which is exactly how a summary line survives its own function being rewritten.

Widened the check — the crate is now clean of both the old name and its old summary text, verified by grep across `agentmux-cef` and `agentmux-launcher` rather than by reading the diff.

Also flagging a correction to this PR body: I claimed the cold path was verified live at `x=576 y=315`. Per your P1, a correctly DIP-centred window cannot come out of a physical/DIP mix, so that window was almost certainly restored via `explicit_rect` rather than centred by this code. I credited my fix with behaviour it did not produce. The pool-path measurement (`x=1320 y=650`) stands — physical on both sides.

PR #3020, opened 2026-09-06
Posa@Area54: fix(swarm): tool rows show the command or file path, not a raw JSON dump

GenericAgentX-asaf:
Expanding an Agent-tool row in Swarm showed this, verbatim:

```
{"command":"ls /c/Users/area54/ 2>/dev/null | head -50; echo \"---\"; ls /c/User…","description":"List home directory contents"}...
```

Escaped quotes, raw braces, cut mid-token. It now shows what the Agent pane has always shown for the identical call:

```
ls /c/Users/area54/ 2>/dev/null | head -50
```

Phase 1 of `docs/reports/REPORT_TOOL_SUMMARY_FORMATTING_ARCHITECTURE_2026_09_06.md` (included).

## The fault is architectural, not cosmetic

Two implementations of "summarise a tool call" exist:

| | |
|---|---|
| **A** — `stream-parser.ts:117` `extractToolDetail` | typed, per-tool, 10 kinds, 3 consumers (`ToolBlock`, `tool-adapter`, `stream-parser`) |
| **B** — `parse.rs:261` | `serde_json::Value::to_string()`, truncated at a bare `200` |

**The swarm could only ever reach B.** Summarising happens at *parse* time on the backend and the structured `input` is discarded — `SubagentEvent` carries only `input_summary: String`. So the frontend cannot apply the good extractor even though it's sitting right there, and the entire tool-renderer registry (`RecordTable`/`SearchResults`/`WebFetchResult`, `SPEC_TOOL_RESULT_RENDERER_REGISTRY_2026_06_17`) is unreachable from that feed for the same reason.

**This PR does not fix that.** Phase 2 does, by carrying the structured input across the boundary. This ports the field selection into Rust as a deliberate, documented down-payment that Phase 2 deletes. I'd rather say that plainly than present a duplicated mapping as the destination.

## Also fixed: a byte/char mismatch

```rust
if s.len() > 200 {                                            // BYTES
    let end = s.char_indices().nth(200).map_or(s.len(), …);   // CHARS
```

A string over 200 **bytes** but under 200 **chars** took the truncation branch, found no 200th char, and fell through to no truncation at all. Now named constants, chars throughout.

## One deliberate divergence from the frontend original

An **unknown tool falls back to the raw JSON object** rather than `""`. `extractToolDetail` can return empty because its callers still hold the full node to fall back on; here there is nothing else to show, and `mcp__*`/provider-specific names are open-ended by design. A known tool whose expected field is missing or empty takes the same fallback.

## Phase 4 — not done as the report proposed

Worth stating rather than quietly narrowing scope: **the report said to delete the dead `Result`-discriminant completion check. I didn't.**

Deleting it would strand `PendingCompletion`, the coalescing buffer's `completed` vector, and two broadcast shapes — all *correct* code. Its only fault is being unreachable in a transcript format we don't control and which could change back; if `"type":"result"` lines ever appear it resumes being the **faster** completion signal (immediate, versus waiting for the parent turn to end). Deleting correct handling for an external format we don't own trades a real capability for tidiness.

What actually caused harm was the comment asserting it works. Both sites now state plainly that it never fires, cite the evidence, point at `completion.rs` as the real path, and record the #2283 history — a placeholder match that "almost never fired" replaced by a discriminant check that fires *never*, indistinguishable from outside, which is exactly how the original bug survived a refactor.

If a reviewer disagrees, §4.4 and §8 of the report lay out the argument both ways. The trap is gone either way.

## Testing

- **11 new tests**, fixture taken from the transcript that produced the reported symptom.
- **Mutation-checked:** forcing the JSON-dump fallback fails 5 of them, including the flagship Bash case; the fallback and truncation tests keep passing, confirming they guard preserved behaviour.
- Full `agentmux-srv` suite: **3030 passed, 0 failed**.

## Not verified

The produced string is unit-tested, but **I have not watched a Swarm row actually render it.** Flagging rather than implying otherwise.

Phases 2 and 3 (move formatting to the edge; one binding surface) remain open in the report.

### DIFF
--- a/.changesets/1788714253-fix-swarm-tool-rows-show-the-command-or-file-path-instead-of-1blg.md
+++ b/.changesets/1788714253-fix-swarm-tool-rows-show-the-command-or-file-path-instead-of-1blg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(swarm): tool rows show the command or file path instead of a raw JSON dump
--- a/agentmux-srv/src/backend/subagent_watcher/jsonl.rs
+++ b/agentmux-srv/src/backend/subagent_watcher/jsonl.rs
@@ -234,12 +234,34 @@ impl SubagentWatcher {
                 state.events.drain(..excess);
             }
 
-            // Check last event for result type (completion). Keyed off the
-            // `Result` discriminant itself (a real `"result"`-typed JSONL
-            // line), not derived text content — real Claude Code result
-            // events populate `result`/`content`, so matching against the
-            // "Subagent completed" placeholder (only ever produced when
-            // both are absent) almost never fired.
+            // Completion by terminal `Result` line.
+            //
+            // ⚠ THIS NEVER FIRES ON ANY TRANSCRIPT AGENTMUX CURRENTLY WRITES.
+            // No `"type":"result"` line exists in the `entrypoint: sdk-cli`
+            // format — 0 of 11 subagent transcripts on the investigated
+            // machine, across CLI 2.1.198 and 2.1.247, and none in the parent
+            // session files either. Every completed subagent transcript ends
+            // on an `assistant` message. Evidence:
+            // docs/reports/REPORT_SUBAGENT_COMPLETION_NEVER_DETECTED_2026_09_05.md.
+            //
+            // Real completion is established by `completion.rs` instead,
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #3021, opened 2026-09-06
fix(lan): LAN discovery toggle takes effect live, no restart

Agent2-asaf:
**Step 3 of the network series.** This is the one that actually fixes the reported problem: enabling LAN discovery now works immediately, like cloud login.

## The bug

The toggle was **half-live**: mDNS started/stopped immediately, but `bind_addr` stayed whatever it was at startup. So enabling made the instance **advertise itself while its listeners were still on `127.0.0.1`** — peers discovered it and couldn't reach it. Disabling was wrong the other way: mDNS stopped but listeners stayed on `0.0.0.0`, still LAN-reachable until restart.

## The design

`LanListenerSupervisor` owns the LAN-facing listeners and reconciles them against **two** inputs: the setting, *and* the host's current interface set.

**Additive, never rebinding.** Startup's loopback listeners are untouched, so the frontend's own WebSocket is never dropped by a toggle. LAN reachability is added by binding the *same ports* on specific non-loopback addresses (the behaviour the spike in #3017 verified) and removed by cancelling them. Ports stay stable — they're already published in the mDNS TXT record and `authkey.dev`. This is precisely why rebinding, which the in-code comment called "non-trivial" and deferred, turns out not to be needed.

**Self-healing.** A 20s sweep re-reconciles against the live interface list, so DHCP renewal / Wi-Fi↔Ethernet handoff / VPN up is picked up with no toggle and no restart. Polling over OS interface-change events is deliberate: no new platform-specific dependency, and worst case is 20s of staleness on a path that's already best-effort discovery.

**Partial failure survivable.** One address failing to bind is logged and skipped; the rest come up. Both ports for an address bind as a unit with rollback, so a half-usable address is never registered active.

## Advertising is now gated on reachability

The supervisor owns the advertise/reachable pairing and re-gates mDNS on every reconcile. We never advertise an address nothing is listening on — that state is *strictly worse* than not advertising, because the peer can't distinguish it from a transient failure and keeps retrying.

**That's why `setconfig` now drives only the supervisor.** Calling `lan.apply` there too would advertise before any listener bound, re-creating the exact bug. Worth flagging for review since it reads like a deletion.

## Multi-homed hosts

Binds **every** reachable address, not just the primary outbound one — a docked laptop on Ethernet *and* Wi-Fi is reachable on either, and binding one would make it invisible to half the network. `if-addrs` was already in the tree via `mdns-sd`, so declaring it directly adds no new supply-chain surface. v6 link-local excluded (needs a scope id to be dialable); loopback excluded (already bound).

## Tests (9)

Binds nothing until enabled · enabling before the router exists is safe and recovers on the sweep · **on→off binds then releases** (the exposure half) · `apply` idempotent under repeats, which the fs watcher does cause · **a bound listener actually SERVES the router over the LAN address** · bind addresses exclude loopback and v6 link-local.

## Two notes for reviewers

- Also corrects the module doc, which still carried the universal *"0.0.0.0 conflicts with 127.0.0.1"* claim the spike disproved on Windows, plus a doc comment an earlier edit had orphaned onto the wrong function.
- **`cargo check` does not catch `AppState` field additions in test fixtures.** Three constructors needed the new field and only `cargo test` surfaced it — worth knowing if you add state fields.

## Verification

Full `agentmux-srv` suite: **3030 passed / 0 failed** · clippy clean on all changed files.


a5af:
@codex review

Agent2-asaf:
Confirmed and fixed — good catch, this was worse than the bug the PR set out to fix.

I verified the mechanism rather than taking it on faith, and the platform split is worth recording: I measured wildcard-then-specific on **Windows** and it **succeeds** (no `SO_EXCLUSIVEADDRUSE` by default). So the `EADDRINUSE` → `active` empty → `sync_advertising(false)` → **silent LAN-disable on every restart** chain is real on Linux/macOS exactly as described; on Windows it instead produces two sockets on one port with ambiguous accept. Neither is acceptable, so the fix is the same either way.

Also worth noting on how this got past me: my original spike tested loopback-then-wildcard, but the shipping boot sequence is wildcard-then-specific. A spike only covers the ordering it actually runs.

Fix (2f4b4a1):
- Startup binds `lan_listeners::STARTUP_BIND_ADDR` (`127.0.0.1:0`) **unconditionally** — the conditional is gone, not relocated. The supervisor is now the sole owner of every LAN socket, which is what the PR's "startup listeners are never touched" invariant actually requires.
- Removed `bootstrap.rs`'s eager `lan_discovery.apply(...)` too. The supervisor is the single driver, so mDNS can never advertise an endpoint before a socket is listening on it — this also closes the smaller advertise-before-bind window that existed independently of your finding. Drops the now-unused `config_watcher` param.

Tests, addressing your point that the existing ones all used fresh non-conflicting ports:
- `wildcard_then_specific_conflict_is_platform_dependent` — reproduces the exact boot sequence against a pre-existing wildcard bind, asserting the measured behaviour per platform so a change in either direction is visible.
- `startup_bind_is_loopback_only` — cheap guard against reintroducing the conditional.

3032 tests pass, clippy clean. Corrections section of `REPORT_NETWORK_ARCHITECTURE_DRYNESS_AND_ROBUST_LAN_2026_09_06.md` updated with both the finding and the spike-coverage lesson.

PR #3022, opened 2026-09-06
refactor(reactive): unify the three jekt forward tiers behind one helper

Agent2-asaf:
Item 4 of `REPORT_NETWORK_ARCHITECTURE_DRYNESS_AND_ROBUST_LAN_2026_09_06.md` §4. Follows #3021.

## The problem

`handle_reactive_inject` walks local → same-host same-channel → same-host cross-channel → LAN peer. The last three each carried their own copy of the same control flow:

```
build forward_url → POST with X-AuthKey → check body.success → echo_jekt_to_sender(…) → log + evict + fall through
```

Three hand-maintained copies of one flow is the shape that has already produced drift bugs in this repo. A tier that forgets the echo, or passes `delivery_tier` wrong, fails silently and differently from its siblings.

## The change

One `forward_inject_to_peer(state, req, forwarded_req, ForwardPeer)` returning:

| Outcome | Meaning | May evict? |
|---|---|---|
| `Delivered(body)` | peer confirmed delivery; echo already emitted | — |
| `Stale` | `success` wasn't `true`, or transport failed | yes, per tier policy |
| `Inconclusive` | non-2xx, or unparseable body | **never** |

Each tier keeps its own eviction policy — registry `should_evict_on_forward_failure` (2a/2b) vs. LAN discovery-cache evict (3) — because that part genuinely differs. Nothing else does. The `Stale`/`Inconclusive` split makes explicit what was previously implicit in the match arms: only one of them is evidence about the *candidate*.

`echo_jekt_to_sender` drops from twelve positionals to nine, with `sig_verified`/`reagent_verified`/`lan_verified` — three consecutive `Option<bool>`s, transposable with no compiler help, on a **security marker** — now named fields on `EchoTrust`.

## ⚠ One behaviour unified rather than preserved

Reading the three copies side by side surfaced a real divergence, which is the argument for the refactor rather than a footnote to it:

**The LAN tier delivered on "anything but `success: false`"; both host tiers required `success: true`.** So a 200 response omitting `success` entirely counted as a successful LAN delivery — echoing a delivery confirmation to the sender for a message that may never have arrived.

All three now require `success: true`. No real peer is affected: `InjectionResponse::success` is a plain `bool` with no `skip_serializing_if`, so a genuine AgentMux srv always sends it. Covered by `a_body_without_success_is_not_delivered`.

## Testing

Six new cases driving the real HTTP path against a throwaway loopback peer (mocking the client would skip the response-classification that *is* the behaviour under test) — one per outcome, including the changed one. `source_agent` is `None` throughout so the echo short-circuits and these stay off the global singleton `reactive_handler`.

**3049 tests pass, clippy clean, doc gates pass.**

### DIFF
--- a/.changesets/1788716157-refactor-reactive-unify-the-three-jekt-forward-tiers-behind--8j4u.md
+++ b/.changesets/1788716157-refactor-reactive-unify-the-three-jekt-forward-tiers-behind--8j4u.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(reactive): unify the three jekt forward tiers behind one helper
--- a/agentmux-srv/src/server/reactive.rs
+++ b/agentmux-srv/src/server/reactive.rs
@@ -46,6 +46,22 @@ const MAX_FORWARD_HOPS: u8 = 3;
 /// `requires_stop` from (not re-derived) — see the call site's own doc
 /// comment inline below for why passing anything else produces a
 /// self-contradictory echoed marker (reagentx P1 on PR #2623).
+/// The delivery-tier + verification signals an echoed marker must carry.
+///
+/// Grouped into a named struct rather than left as positionals because
+/// `sig_verified`, `reagent_verified` and `lan_verified` are three consecutive
+/// `Option<bool>`s: any two of them can be transposed at a call site and the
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

PR #3023, opened 2026-09-06
feat(muxbus): jekt delivery tier 4 — cloud relay performed by srv

Agent2-asaf:
Item 5 of `REPORT_NETWORK_ARCHITECTURE_DRYNESS_AND_ROBUST_LAN_2026_09_06.md` §5. Follows #3021 and #3022.

## The gap

Tier 4 was a comment in `reactive.rs`:

```rust
// 4. Return original error (muxbus-client will fall back to cloud relay).
```

But the callers agents actually use are not that client. The MCP `SendMessage` tool POSTs to `/agentmux/reactive/inject` and bails on `success != true` — so the documented four-tier model was **three tiers plus a comment**, while `SendMessage`'s own description promised "local → LAN → cloud".

## The change

New `crate::muxbus::relay` implements the outbound half (`POST /reactive/inject`); `handle_reactive_inject` calls it as tier 4 once tiers 1–3 decline, so **every** caller inherits it instead of re-implementing it.

## Three things worth review attention

**1. The contract is not what the specs implied.** I originally inferred the request body from `SPEC_JEKT_LAN_WAN_TRUST_HARDENING`'s description of `createInjection`'s stored row — and that would have shipped broken. Reading `agentmux-cloud/muxbus/server/src/index.ts` directly gives the real contract:

- `source_agent` travels in the **`X-Agent-ID` header, not the body** — the route 400s without it and derives the sender from it alone.
- **`X-Client-Wrapped: true` is required.** Without it the cloud wraps the message in its own `[JEKT:...]` marker *and* the receiving srv wraps it again in `Handler::inject_message` → the recipient sees a doubled marker. Sending raw + declining server-side wrapping yields exactly one marker, applied by the receiver, identical to tiers 2 and 3.

Both are pinned by `the_sender_travels_as_a_header_and_wrapping_is_declined`.

**2. Tier 4 needs its own loop guard.** `forward_hops` cannot stop a relay↔subscriber ping-pong: the cloud delivers each message as a *fresh* inbound request with the count reset. An agent unknown to every subscribed instance would bounce forever, re-queueing each hop. A message that arrived with `delivery_tier == "wan"` is therefore never re-relayed (`a_message_that_arrived_over_wan_is_never_re_relayed`).

**3. Queued is not delivered.** The relay persists the message and wakes subscribers; the recipient's srv collects it on its next sync — or never, if that instance stays offline. So the outcome type is `Queued`, not `Delivered`; the echoed marker carries `DELIVERY=wan`; and `SendMessage`'s description now says this plainly instead of implying confirmed delivery. The tier-3 path had exactly this conflation and it was a real bug (#3022).

## Also

Dedupes `relay_base_url` (was private to `pkce.rs`) into `muxbus::relay::rest_base_url` — one definition of "where is the cloud".

## Testing

9 new tests. The wire contract runs against a **capturing** stub relay so the assertions are on the actual headers and body, not just the outcome: queued-with-id, header/wrapping contract, 402 quota passthrough, 2xx-without-`success:true`, oversize pre-check, unreachable. Plus the tier-4 gates: wan loop guard, missing/empty `source_agent`, and that host/lan/unset tiers stay eligible.

`relay_inject` takes `base_url` and `token` as parameters rather than resolving them internally, specifically so tests need no `set_var` (which would race the rest of the test binary).

**3058 tests pass, clippy clean.**

## Not covered

No live call against production muxbus was made from this branch — the contract is verified against the cloud server source and a stub, not the deployed service.

### DIFF
--- a/.changesets/1788717537-feat-muxbus-jekt-delivery-tier-4-cloud-relay-is-now-performe-8gjt.md
+++ b/.changesets/1788717537-feat-muxbus-jekt-delivery-tier-4-cloud-relay-is-now-performe-8gjt.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(muxbus): jekt delivery tier 4 — cloud relay is now performed by srv
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -97,6 +97,14 @@ jobs:
       - name: Input-handler layout-read gate
         run: bash tools/lint/check-input-handler-layout-reads.sh
 
+      # Same rationale as the three above — encodes a regression that already
+      # happened (PR #3023 P0: tier-4 cloud relay read credentials from the
+      # wrong store and silently never fired). Not unit-testable: test_state()
+      # backs wstore/id_store/identity_store with one Store, so a runtime test
+      # cannot tell them apart.
+      - name: muxbus credential store gate
+        run: bash scripts/check-muxbus-credential-store.sh
+
   frontend:
     name: vitest
     runs-on: ubuntu-latest
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -110,7 +110,7 @@ const SHELL_STATUS_TOOL: &str = r#"{
 
 const SEND_MESSAGE_TOOL: &str = r#"{
   "name": "SendMessage",
-  "description": "Send a message to another agent by name. The message is injected as input into the target agent's active conversation. Use for agent-to-agent coordination — handoff, task delegation, status notifications. Delivery is best-effort and tries local → LAN → cloud in order. Returns once delivery is confirmed or all tiers have failed.",
+  "description": "Send a message to another agent by name. The message is injected as input into the target agent's active conversation. Use for agent-to-agent coordination — handoff, task delegation, status notifications. Delivery is best-effort and tries local → same-host → LAN → cloud in order. The first three tiers return only once the message has actually been injected; the cloud tier is store-and-forward, so success there means the relay accepted it and the recipient's AgentMux will pick it up when it next syncs (which never happens if that instance stays offline). Returns once one tier has taken the message or all have failed.",
   "inputSchema": {
     "type": "object",
     "properties": {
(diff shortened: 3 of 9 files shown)


a5af:
@codex review

Agent2-asaf:
Confirmed and fixed — this was the whole feature silently not working, and the failure mode was worse than a crash: `relay_token` returned `None`, tier 4 logged "cloud relay skipped: not logged in to muxbus" and returned. That reads like correct behaviour for a logged-out user, so nothing would ever have surfaced it.

I verified the chain rather than taking it: `CloudSubscriber::init_global(id_store.clone())` (`bootstrap.rs:1056`) and all three `muxbus.login`/`status`/`disconnect` handlers (`muxbus_handlers.rs:61,150,191`) write to `state.id_store`. Fixed to match.

**One deliberate divergence from your citation:** I used `id_store`, *not* `identity_store`, even though `AppState`'s doc comment steers new muxbus call sites toward the latter. The credentials are **written** to `id_store`; reading from a store the writer doesn't use would reintroduce this exact bug whenever `isolated_auth_enabled()` redirects one but not the other. Following the doc's steer here would have swapped one silent-miss for another. That reasoning is recorded on `relay_token` and in the gate's header so a future migration moves writers and readers together.

**A second bug fell out of investigating this.** `ensure_agent_credential` attempts a cloud provisioning round trip (`POST /agents/provision`) when unprovisioned. My original ordering called it *before* the shared-token check — so a logged-out instance made a network call on every failed local inject, i.e. on the hot path of every message to an unknown agent. Reordered: the shared token is a purely local read and is checked first; no credential means no login means nothing to provision against.

## On testing this

**It is not unit-testable, and I want to be explicit rather than quietly skipping it.** `server::tests::test_state()` backs `wstore`, `id_store` *and* `identity_store` with a single in-memory `Store` — so seeding a credential into one makes it visible through all three, and any such test passes whichever field the code reads. I checked before writing one; it would have been theatre.

So I added a CI grep gate instead (`scripts/check-muxbus-credential-store.sh`), matching the three existing gates that each encode a regression that already happened once. It fails the build if any muxbus credential call is handed `state.wstore`. Verified both directions — clean on the fix, and exit 1 with the bug deliberately reintroduced:

```
ERROR: muxbus credential lookup against state.wstore.
  - agentmux-srv/src/server/reactive.rs:1125: ... relay_token(source, &state.wstore, ...)
```

It matches `state.wstore` specifically, not a bare `wstore` identifier — several functions inside `muxbus/` name their own parameter `wstore` while callers correctly pass the id_store. That's a naming wart (item 7 of the network report) rather than a bug, and flagging it would make the gate noise.

3058 tests pass, clippy clean.

PR #3024, opened 2026-09-06
refactor(network): fix three misleading names + write the credential map

Agent2-asaf:
Item 7 of `REPORT_NETWORK_ARCHITECTURE_DRYNESS_AND_ROBUST_LAN_2026_09_06.md` — §6, §7 and §8, the "small, do opportunistically" tail of the network series. Completes it (after #3021, #3022, #3023).

No behaviour changes. One wire-field rename, one type rename, one new doc.

---

### §6 — `wan.subscribed_agents` → `wan.local_agents_subscribed`

The field lists **this host's own** cloud-subscribed agents, not remote agents reachable over WAN. Under a `wan` heading, in a response whose entire purpose is reachability, the old name reads as the opposite of what it means — and it directly caused a wrong diagnosis while I was debugging cloud delivery earlier in this series.

The report offered an alternative — make it answer the question its name implies — but that isn't available: the relay exposes no directory, so nothing can list who is reachable via cloud. So: rename, **and** make `DiscoverAgents`' description say outright that a cloud-only target will not appear anywhere in its output. That last part is what actually misled; the field name alone was only half the problem.

`CloudSubscriber::subscribed_agents()` is deliberately *not* renamed — it's accurate on the subscriber itself. Only the wire field was ambiguous.

### §7 — `wconfig::watcher::ConfigWatcher` → `wconfig::state::ConfigState`

It watches nothing. The name came from a comment promising that "the actual file system watching will be integrated with the event loop in a later phase" — that never happened; watching was built separately in `config_watcher_fs.rs` on top of `fs_watch::pool`, and the placeholder name outlived the plan. Two similarly-named modules where only one watches cost a wrong turn tracing the settings-reload path.

Pure rename, compiler-checked. Both the module doc and the type doc now state plainly that this holds config and watches nothing, with a pointer to where watching actually lives.

### §8 — `docs/specs/ARCHITECTURE_NETWORK_CREDENTIAL_MAP_2026_09_06.md`

Writing it surfaced a distinction the report hadn't separated cleanly, and it's the easiest thing in this area to get wrong:

- **Route gates** (`auth_key`, `lan_key`, `host_reg_secret`, `ipc_token`, the muxbus tokens) decide whether a request reaches a handler at all.
- **Message signatures** (`jekt_sig`, `lan_sig`, `reagent_sig`) are evaluated *after* it already has, and never grant or deny access — they only set `TRUST=`.

A valid `lan_key` gets a peer through the door and says nothing about who sent the message it carries; that's what `lan_sig` is for. The doc leads with that, then maps each credential to the exact routes it gates, and records where muxbus credentials live (`id_store`, per #3023) and why the tier-4 relay sets no verification fields.

Explicitly **not** an argument to consolidate any of these — scope separation is right for a security boundary. The gap was that no single place said which gates what.

---

## Testing

3057 pass, clippy clean, doc gates pass (index regenerated for the new spec).

**One pre-existing failure, not from this PR:** `backend::sysinfo::handle_anomaly_tests::refresh_processes_specifics_does_not_leak_a_handle_per_call`. I verified it fails identically on unmodified `main` on this machine (grew by 468 there vs 446 here), and my only change to `sysinfo.rs` is three type annotations from the rename. It passes in CI. The threshold appears calibrated for a quieter host than this one, which is currently running several AgentMux instances and builds — flagging it as worth a look separately rather than silently ignoring it.

### DIFF
--- a/.changesets/1788719271-refactor-network-rename-configwatcher-and-wan-subscribed-age-zwhg.md
+++ b/.changesets/1788719271-refactor-network-rename-configwatcher-and-wan-subscribed-age-zwhg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(network): rename ConfigWatcher and wan.subscribed_agents; add credential map
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -123,7 +123,7 @@ const SEND_MESSAGE_TOOL: &str = r#"{
 
 const DISCOVER_AGENTS_TOOL: &str = r#"{
   "name": "DiscoverAgents",
-  "description": "List the agents and instances reachable from here across the muxbus delivery tiers (host, LAN, cloud), so you can pick a valid target before SendMessage. Returns JSON: host.addressable (agents reachable right now via local delivery), host.agents (this host's agent directory, each with an `addressable` flag), lan (LAN peers and the agents on them), and wan.subscribed_agents (cloud-subscribed agents). Addressing is by agent name, case-insensitive. Takes no arguments.",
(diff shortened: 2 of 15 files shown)


a5af:
@codex review

PR #3025, opened 2026-09-06
fix(lan): stop this instance discovering itself as a LAN peer

Agent2-asaf:
Found by running `DiscoverAgents` against a live instance after the network series merged — not by reading code.

## Evidence

`DiscoverAgents` reported exactly one LAN "peer": `172.23.176.1:55019`, with empty `agents`, `hostname` and `instance_id`. The srv log shows what it really is:

```
LAN peer discovered  peer_id=""  address=fe80::5c1e:c2bb:b5bc:9655  port=55019
LAN peer discovered  peer_id=""  address=fe80::5c1e:c2bb:b5bc:9655  port=55019
LAN peer discovered  peer_id=""  address=192.168.1.230              port=55019
LAN peer discovered  peer_id=""  address=172.23.176.1               port=55019
```

Port 55019 is **this instance's own port**. All three addresses are **this host's** (confirmed against `ipconfig`: `192.168.1.230`, `192.168.153.1`, `192.168.116.1`, `172.23.176.1`). The instances map is keyed by fullname, so the four events collapse into one entry — us, wearing a peer's clothes.

## Why the existing fix didn't catch it

There is already a self-skip for exactly this (#2598, 2026-08-15) comparing the resolved fullname to our registered one, and it **is** in the running 0.55.34 build. It isn't firing. Why the fullnames differ isn't established — plausibly mdns-sd conflict-renaming when several AgentMux instances share this host, or a re-registration under `enable_addr_auto()`.

I chose not to guess at that. This adds a second guard on what *can* be stated with certainty: **a service on our port, at an address belonging to this machine, is us.** A genuine remote peer cannot satisfy both conditions.

## The part worth your attention

**The existing `handle_event_tests` for the self-skip pass, and have passed all along.** They build `service_fullname` and the resolved event from the *same* `instance_id`, so the fullnames match **by construction**. The test cannot express the production case where they differ.

That is how this survived a fix, a test, and three weeks — and it's the reason I went to the live logs rather than trusting a green suite.

## Why it matters more now

`find_agent` fans out concurrently to every known peer (#3016), so every LAN lookup was spending a request asking *ourselves* a question we'd already answered locally.

## Implementation

- `all_local_addresses()` in `lan_listeners` — deliberately **broader** than `lan_bind_addresses()`: loopback and IPv6 link-local included, because the exclusions that make sense for *binding* are exactly wrong for *"is this us?"*. A `fe80::` address is useless as a bind target and very much still us.
- `is_self_resolution()` split out as a pure predicate, testable without standing up a `ServiceDaemon`.
- `any` not `all` on the address match: interface enumeration and mDNS resolution race, and one stray address must not resurrect the phantom. `any` cannot false-positive — a remote peer would have to be listening on our port at one of our own addresses.
- Enumeration failure yields an empty set, which makes the guard inert — degrading to today's phantom rather than to something worse like discarding a real peer.

## Testing

6 tests: the measured case (one per observed address), a real remote peer on a coincidentally equal port, a sibling AgentMux instance on this host (same addresses, different port), the any-vs-all race, and both degenerate inputs. **3064 pass, clippy clean.**

## Not included

`lan_bind_addresses()` also binds this host's three *virtual* adapters (WSL/Hyper-V), which no real peer can reach — unnecessary sockets and mDNS advertising. Filtering those is a genuine judgment call (a heuristic risks excluding a VPN someone legitimately wants LAN over), so I've left it for a decision rather than guessing.

### DIFF
--- a/.changesets/1788720284-fix-lan-stop-this-instance-discovering-itself-as-a-lan-peer-58f0.md
+++ b/.changesets/1788720284-fix-lan-stop-this-instance-discovering-itself-as-a-lan-peer-58f0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(lan): stop this instance discovering itself as a LAN peer
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -474,12 +474,21 @@ impl LanDiscovery {
     /// while several AgentMux instances share a machine — but no remote peer
     /// can be listening on OUR port at an address that is OURS.
     fn resolves_to_this_instance(&self, info: &ServiceInfo) -> bool {
+        // Port first, and BEFORE touching the address set. A resolution on a
+        // different port cannot be us, and that is the overwhelmingly common
+        // case (every genuine remote peer), so it must not pay for an address
+        // lookup. Passing the set as an argument would defeat this — arguments
+        // are evaluated eagerly, so the cheap check inside `is_self_resolution`
+        // would come too late [reagent #3025 P2].
+        if info.get_port() != self.port {
+            return false;
+        }
         let addrs: Vec<IpAddr> = info.get_addresses().iter().copied().collect();
         is_self_resolution(
             info.get_port(),
             &addrs,
             self.port,
-            &crate::backend::lan_listeners::all_local_addresses(),
+            &crate::backend::lan_listeners::cached_local_addresses(),
         )
     }
 
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -197,6 +197,29 @@ fn query_peers_concurrently<'a>(
     futures
 }
 
+/// Pure predicate behind `LanDiscovery::resolves_to_this_instance`, split out
+/// so it can be tested without standing up an mDNS `ServiceDaemon`.
+///
+/// `any` rather than `all` on the address match: a resolution for our own
+/// service can legitimately carry an address that interface enumeration missed
+/// (they race), and requiring every address to be ours would let that one
+/// stray entry resurrect the phantom. `any` cannot produce a false positive —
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

Agent2-asaf:
Valid, and the flaw was slightly worse than described — worth spelling out because it changes the fix.

`all_local_addresses()` was passed as an **argument** to `is_self_resolution`. Arguments evaluate eagerly, so the cheap `resolved_port != own_port` check *inside* the predicate never got the chance to short-circuit it. The enumeration happened unconditionally on every resolution that missed the fullname fast-path — so a TTL cache alone would have papered over it while leaving every genuine remote peer paying for a lookup it can never need.

Two changes for the two halves:

**1. Port checked in the caller, before the address set is touched.** A resolution on a different port cannot be us, and that covers every real peer — so the common case now costs zero syscalls, cached or not.

**2. `cached_local_addresses()` — 5s TTL.** For the case the port check can't help with: our own high-frequency self-resolutions, which by definition match our port. That's exactly the churn you identified.

On the staleness trade, since it's the reason to be careful with a cache here: a brand-new interface isn't recognised as ours for up to 5s, so one resolution on that address can still land as a phantom — which clears on the next resolution after a refresh. Bounded and self-healing, which the pre-fix behaviour was not. A poisoned lock falls back to enumerating directly rather than failing the self-check, because failing it would resurrect the very phantom this PR removes.

The test asserts the contract rather than the timing — same contents as a fresh enumeration, and `Arc::ptr_eq` on a second call to prove the snapshot is reused. A TTL test that sleeps is a flaky test.

**Also fixed, unprompted:** three `///` comments where `[reagent #NNNN Pn]` was being parsed by rustdoc as an intra-doc link (`unresolved link to reagent`). Now parenthesised. One is in `muxbus/relay.rs` from #3023 and already on main — same class, so I took it here rather than leaving a known warning.

3065 pass, clippy clean, no new doc warnings.

PR #3027, opened 2026-09-06
chore: release v0.55.37

Agent2-asaf:
Release PR consuming the 20 changesets queued since v0.55.36.

## Bump: patch (forced)

Run as `task release -- --as patch`, per the patch override. One queued changeset is **minor** — `feat(muxbus): jekt delivery tier 4` — so this deliberately ships a minor-level change under a patch version. The script's own warning fired and is expected here, not an oversight:

```
NOTE: --as overrides the computed bump 'minor' with 'patch'.
WARNING: forcing a LOWER bump than the changesets request — some
         minor-level change(s) will ship under a patch version.
```

`0.55.36 -> 0.55.37`

## Version consistency invariant

Verified independently of the script's own self-check — all five agree at `0.55.37`:

| Location | Value |
|---|---|
| `VERSION_HISTORY.md` head | `## 0.55.37 — 2026-09-06` |
| `package.json` | `0.55.37` |
| `Cargo.toml [workspace.package]` | `0.55.37` |
| `Cargo.lock` (`agentmux-cef`) | `0.55.37` |
| `package-lock.json` root | `0.55.37` |

20 changesets consumed and deleted. (`.changesets/README.md` remains by design — it's the directory's documentation, not a changeset.)

## What's in it

Highlights of the 20 entries — the LAN/muxbus delivery work is the bulk:

- **`feat(muxbus)`** — jekt delivery tier 4: cloud relay is now performed by srv, so every caller inherits it instead of the tier chain silently stopping at three
- **`fix(lan)`** — LAN discovery toggle takes effect live, no restart
- **`fix(lan)`** — stop this instance discovering itself as a LAN peer
- **`perf(lan)`** — query LAN peers concurrently instead of sequentially
- **`refactor(reactive)`** — unify the three jekt forward tiers behind one helper
- **`refactor(network)`** — rename `ConfigWatcher`/`wan.subscribed_agents`, add the credential→route map
- **`fix(settings)`** — seed `settings.json` at startup instead of only when the Settings pane opens

Plus tray, swarm, window, agent-preview and provider-pin fixes from other agents. Full list in the `VERSION_HISTORY.md` diff.

## Note for anyone pulling this to test LAN

`network:lan_discovery` still defaults to **off** — that's intentional (don't trigger the OS firewall prompt unless the user needs it). Enabling it now applies live, no restart, which is new in this release.

### DIFF
--- a/.changesets/1788664032-fix-agent-swap-the-remaining-tail-glyphs-from-the-weak-cover-bt9e.md
+++ b/.changesets/1788664032-fix-agent-swap-the-remaining-tail-glyphs-from-the-weak-cover-bt9e.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(agent): swap the remaining tail glyphs from the weak-coverage arrow
--- a/.changesets/1788668658-fix-host-publish-the-ipc-port-file-where-the-launcher-reads--dtpk.md
+++ b/.changesets/1788668658-fix-host-publish-the-ipc-port-file-where-the-launcher-reads--dtpk.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(host): publish the ipc-port file where the launcher reads it in dev mode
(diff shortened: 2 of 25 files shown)


a5af:
@codex review

PR #3030, opened 2026-09-06
fix(window): a cold-path window inherits is_browser_pane from a cloned floater client and is never revealed

### LINKED ISSUE #3028: Cold-path open_new_window creates a correctly-placed window that is never shown
## Observation

While verifying #3019 on Windows I bypassed the window pool locally (made `promote_pool_window_for_new_window` return `None`) so that `open_new_window` would take the cold path. The result:

- the IPC call succeeded and returned a non-pool label (`window-7f73cb1b…`, i.e. not `window-pool-…`)
- the window was created at the correct rect (`x=576 y=315 w=2688 h=1470` physical, correctly centred)
- the frontend loaded and round-tripped IPC — `[window] register_backend_window received` appears in the host log
- **the window remained `IsWindowVisible == false` for the entire observation window (~60s)**

So a cold-path `open_new_window` produced a fully-initialised window that was never shown. Repeated across the whole run; the window sat hidden until the instance was shut down.

## Why this might matter in production

The pool bypass was artificial, but returning `None` is exactly what `promote_pool_window_for_new_window` does naturally when the pool is empty. If the pool can be empty at the moment a user asks for a window — e.g. immediately after startup before the pool has warmed, after pool-spawn failure, or under churn — then "New Window" would appear to do nothing.

This is more reachable now that background-service mode (#2977 WS0) makes "no window open, service resting" a normal state, and the tray's **New Window** the routine way back in.

## What I did NOT establish

- Whether the pool is ever genuinely empty at New-Window time in production. If it never is, this is unreachable and low priority.
- Whether the missing show is caused by my local patch rather than the cold path itself. The patch only short-circuits the pool promote, but I did not prove that has no other effect.
- Where the show is *supposed* to happen for a cold-path window. `WindowDelegate::on_window_created` adds the browser view and sets bounds; I did not find the corresponding show/visible step, and did not chase it further.

Deliberately filed as an observation rather than a diagnosis — I induced the condition artificially and stopped short of a root cause, so someone should confirm it reproduces naturally (empty pool, no local patch) before treating it as a defect.

## Repro used

1. Windows, `task dev` with `AGENTMUX_TRAY=1 AGENTMUX_BACKGROUND_SERVICE=1`
2. locally patch `promote_pool_window_for_new_window` to `return None` first thing
3. close every window (background-service mode keeps the service alive)
4. `POST /ipc {"cmd":"open_new_window"}` using `<data-dir>/ipc-port-<hash>`
5. enumerate top-level windows: the new one exists, is correctly placed, and is hidden

Agent5-asaf:
## Summary

Fixes #3028: a cold-path `open_new_window` (pool empty) creates a window that
loads, round-trips IPC (`register_backend_window received`), reports first
paint — and **stays hidden forever**.

> **Note:** the root cause below replaces the one this PR originally shipped.
> The first diagnosis (CEF Views misreporting `is_visible()`) was wrong; it is
> documented in the comments as a corrected mis-diagnosis, not quietly dropped.

## Root cause — proven from the repro's own host log

`CreateWindowTask::execute` clones its CEF `Client` from an arbitrary browser
returned by `list_top_level_browsers()`. That helper **deliberately includes
`Floater`s** (its other callers emit host JS events, and floaters are trusted
renderers — see `BrowserKind`'s own doc comment), and **every floater client is
built with `is_browser_pane = true`** (`floating_pane.rs`).

A cold-path top-level window that clones a floater's client inherits that flag.
`on_load_end` then returns at:

```rust
if self.is_browser_pane { /* pane work */ return; }
```

which sits **above the entire reveal block**. So the paint gate is never armed —
and every backstop hangs off arming:

- `report_first_paint` arrives with the correct label and no-ops
  (`reveal_gated_window` returns when nothing is pending);
- the 4s safety timeout is never *scheduled*, because scheduling happens inside
  the block that never runs.

No log is emitted from any reveal branch, which is exactly what made this
undiagnosable from the field.

### The proof

Host log of the original #3028 repro, at `20:13:19.309` — the **same
`on_load_end` invocation** as the cred injection, for the top-level
`window-7f73cb1b…`:

```
Injected IPC port 61928 into page: …windowLabel=window-7f73cb1b…
[pane-load-end] pane page loaded; reinstalling focus subclass
[pane-subclass] subclassed child HWND 0xa10cae class=Chrome_RenderWidgetHostHWND
[pane-load-end] couldn't resolve block_id for nav-state emit         (WARN)
[pane-loading-state] couldn't resolve block_id for nav-state emit    (WARN)
```

A browser registered as `TopLevel { is_pool: false }` ran the **browser-pane
path**, and the pane code could not resolve a `block_id` *precisely because it
is not a pane*. The only floater alive at creation time was
`floating-pool-55a5994…`.

### Why it looked intermittent

`browsers` is a `HashMap`: which client gets cloned is arbitrary, but **stable
within a process**. That is why it reproduced on every New Window in the
issue's run (all windows closed under background-service — only pool +
`floating-pool-*` browsers alive) and not at all in a later three-rapid-clicks
run with `main` open, where the clone happened to pick a non-floater client and
the window showed normally (`EVENT_SYSTEM_FOREGROUND` in the host log).

### Not a new class of bug

The same "the inherited flag lies, the registered kind doesn't" hazard is
already documented at `client/lifecycle.rs:189`:

> *…`CreateWindowTask::execute` reuses an existing browser's CEF Client via
> `first_browser()` — if the iteration happens to pick a pane, the new window
> inherits `is_browser_pane=true`… LABEL is the source of truth.*

That fix corrected the downstream **`BrowserKind` classification**. The flag
itself was left wrong, and `on_load_end` still branched on it.

## The fix — both ends

1. **`on_load_end` stops trusting the flag** (`takes_pane_path`): the pane
   branch is decided by this browser's own registered `BrowserKind`. Panes and
   floaters still take the pane path (floaters host panes and need the
   focus-subclass re-install — that's why their clients set the flag at all);
   a `TopLevel` or `Popup` never does, no matter what client it cloned. An
   **unresolvable** kind falls back to the flag — the pre-fix behaviour — so
   nothing is reclassified on missing evidence. Logs at `warn` when the flag
   and the kind disagree, since the silent version is what hid this.
2. **`CreateWindowTask` picks a better client**: prefers a genuine
   `BrowserKind::TopLevel` browser (`list_full_top_level_browsers`), falling
   back to the floater-inclusive list rather than aborting creation. This keeps
   the flag honest at the source and removes the `HashMap`-order flakiness.

## Also included (secondary hardening, NOT the fix)

The three-site `should_run_reveal(views_visible, native_visible)` change from
the first round. It came from the wrong diagnosis, but stands on its own terms:
the retry path was inconsistent with the initial pass, and
`try_show_resolved_window` could resolve a gate as "done" without showing.
**Happy to split this out** if a reviewer would rather this PR do exactly one
thing.

## Verification

- **Falsified**: reverting `takes_pane_path` to the flag-only behaviour fails
  `a_top_level_window_with_an_inherited_pane_flag_is_not_a_pane` (and the
  pane/floater test); restoring it passes.
- `cef 396 passed` (392 + 4 new).
- **Root cause is established from the recorded host log of the original
  repro**, not from a hypothesis — see the proof above.
- **Live re-verification of the issue's exact repro (background-service mode,
  all windows closed, empty pool) has NOT been run on this build.** Last round
  I let an approval plus four green checks stand in for that and nearly merged
  a non-fix, so I am stating the gap plainly rather than implying coverage I
  don't have.

### DIFF
--- a/.changesets/1788729056-fix-window-reveal-a-cold-path-window-even-when-cef-views-mis-coej.md
+++ b/.changesets/1788729056-fix-window-reveal-a-cold-path-window-even-when-cef-views-mis-coej.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(window): reveal a cold-path window even when CEF Views misreports it visible
--- a/agentmux-cef/src/client/navigation.rs
+++ b/agentmux-cef/src/client/navigation.rs
@@ -330,6 +330,51 @@ const SHOW_WINDOW_MAX_RETRIES: u32 = 10;
 /// two later, long enough not to busy-loop the UI thread.
 const SHOW_WINDOW_RETRY_DELAY_MS: i64 = 50;
 
+/// Should `on_load_end` run the reveal path for a resolved top-level window?
+///
+/// The Views-layer answer alone is NOT trustworthy on Windows. Observed live
+/// (issue #3028, 2026-09-06, cold-path `open_new_window` with the pool
+/// bypassed): `Window::is_visible()` returned nonzero for a freshly created
+/// cold-path window whose native HWND was `IsWindowVisible == FALSE` — so this
+/// branch silently skipped, the paint gate was never armed, and the window
+/// stayed hidden forever. The frontend's `report_first_paint` arrived with the
+/// correct label and no-opped (`reveal_gated_window` returns when nothing is
+/// pending), and the 4s safety timeout was never scheduled: every backstop
+/// hangs off arming, so "skip arming" defeated all of them at once. Same
+/// Views-vs-HWND mismatch class as `POOL_HWND_CACHE`
+/// (`SPEC_POOL_WINDOW_HWND_NULL_2026_05_06.md`) and cef#3638.
+///
+/// `views_visible` is the Views-layer claim; `native_visible` is
+/// `IsWindowVisible` on the HWND when we have one (None on non-Windows, or
+/// when no HWND is bound for the label). Reveal when EITHER layer says the
+/// window is not visible: revealing an actually-visible window is harmless
+/// (`try_show_resolved_window` treats already-visible as resolved), while
+/// skipping a hidden one strands it permanently.
+fn should_run_reveal(views_visible: bool, native_visible: Option<bool>) -> bool {
+    !views_visible || native_visible == Some(false)
+}
+
+/// Native-layer visibility for a label's bound HWND. `None` when no HWND is
+/// bound (or off Windows) — callers must treat that as "no extra information",
+/// not as hidden.
+#[cfg(target_os = "windows")]
+fn native_window_visible(state: &Arc<AppState>, label: Option<&str>) -> Option<bool> {
+    let label = label?;
+    let hwnd = *state.window_hwnds.lock().get(label)?;
+    // SAFETY: IsWindowVisible tolerates stale/invalid handles (returns FALSE).
+    let vis = unsafe {
+        windows_sys::Win32::UI::WindowsAndMessaging::IsWindowVisible(
+            hwnd as windows_sys::Win32::Foundation::HWND,
+        )
+    };
+    Some(vis != 0)
+}
+
+#[cfg(not(target_os = "windows"))]
+fn native_window_visible(_state: &Arc<AppState>, _label: Option<&str>) -> Option<bool> {
+    None
+}
+
 /// Show (or, on Linux, arm the startup-paint gate for) a top-level window's
 /// Views `Window`, once confirmed not yet visible. Shared by `on_load_end`'s
 /// primary resolution path and `ShowWindowRetryTask`'s retry path so BOTH
@@ -653,7 +698,30 @@ impl AgentMuxHandler {
             if let Some(bv) = browser_view_get_for_browser(browser_cloned.as_mut()) {
                 if let Some(window) = bv.window() {
                     resolved = true;
-                    if window.is_visible() == 0 {
+                    let views_visible = window.is_visible() != 0;
+                    let native_visible =
+                        native_window_visible(&self.state, browser_label.as_deref());
+                    if views_visible && native_visible != Some(false) {
+                        // Genuinely visible (or no native evidence otherwise).
+                        // Log it: the silent version of this skip is what made
+                        // issue #3028 undiagnosable from the field.
+                        tracing::debug!(
+                            target: "startup-paint",
+                            label = ?browser_label,
+                            "[on_load_end] window already visible — reveal skipped"
+                        );
+                    } else if views_visible {
+                        // The mismatch case from issue #3028: Views claims
+                        // visible, the native HWND is hidden. Loud, because it
+                        // means the Views layer cannot be trusted for this
+                        // window and only this check saved it.
+                        tracing::warn!(
+                            target: "startup-paint",
+                            label = ?browser_label,
+                            "[on_load_end] Views reports visible but native HWND is hidden — revealing anyway (issue #3028)"
+                        );
+                    }
+                    if should_run_reveal(views_visible, native_visible) {
                         // Linux: defer the actual show()/focus (and the splash
                         // dismiss above) until the frontend confirms a real
                         // compositor paint, instead of doing it here on mere
@@ -954,4 +1022,35 @@ pub(crate) fn show_load_error_page(
     let data_uri = format!("data:text/html;base64,{}", b64_str);
     let uri = CefString::from(data_uri.as_str());
     frame.load_url(Some(&uri));
+}
+
+#[cfg(test)]
+mod reveal_decision_tests {
+    use super::should_run_reveal;
+
+    /// The normal deferred-show case: nothing claims the window is visible.
+    #[test]
+    fn a_hidden_window_is_revealed() {
+        assert!(should_run_reveal(false, None));
+        assert!(should_run_reveal(false, Some(false)));
+    }
+
+    /// Issue #3028: Views said visible, the native HWND said hidden, and the
+    /// old `window.is_visible() == 0` gate silently skipped — never arming the
+    /// paint gate, so the paint signal AND the 4s timeout both no-opped and
+    /// the window stayed hidden forever. Native evidence must win.
+    #[test]
+    fn views_visible_but_native_hidden_still_reveals() {
+        assert!(should_run_reveal(true, Some(false)));
+    }
+
+    /// A genuinely visible window must not be re-revealed: the reveal path
+    /// also focuses, and stealing focus on every load-end would be a
+    /// regression for real visible windows.
+    #[test]
+    fn a_genuinely_visible_window_is_left_alone() {
+        assert!(!should_run_reveal(true, Some(true)));
+        // No native evidence => trust the Views layer, as before this fix.
+        assert!(!should_run_reveal(true, None));
+    }
 }
--- a/agentmux-cef/src/client/navigation.rs
+++ b/agentmux-cef/src/client/navigation.rs
@@ -126,7 +126,21 @@ fn try_show_resolved_window(state: &Arc<AppState>, label: &str) -> bool {
     let Some(mut browser) = state.get_browser(label) else { return false };
     let Some(bv) = browser_view_get_for_browser(Some(&mut browser)) else { return false };
     let Some(window) = bv.window() else { return false };
-    if window.is_visible() == 0 {
+    // Same native-vs-Views check as `on_load_end`'s arming site (issue #3028):
+    // trusting Views alone here would resolve the gate as "done" while the
+    // native HWND stays hidden — the reveal's very last step silently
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

Agent5-asaf:
Fixed in be2ee4e — and your framing was verbatim true: the retry path exists precisely to ride out CEF Views timing quirks, so gating it on the Views layer alone reproduces #3028 through the one path meant to rescue it.

The same sweep found a third site you did not flag, and it is arguably worse: `try_show_resolved_window` (the gate-fire show) returns `true` — resolved — without showing on a misreport, so even a correctly armed gate resolves as done while the window stays natively hidden. All three sites now share `should_run_reveal` + `native_window_visible`, each logging the mismatch at warn. Swept the crate: no `is_visible() == 0` reveal gate remains outside doc text.

The irony is not lost on me: my fix for a single-call-site bug had the single-call-site bug. The predicate existed; I wired it into the site I had traced live and stopped looking.

cef 392 passed. Live verification of the natural 3-rapid-clicks repro is running on this exact build; result to follow before merge.

Agent5-asaf:
## Do not merge — this does not fix #3028. Converting to draft.

I promised the live result before merge. It is negative, and I would have merged a non-fix on the strength of an approval and four green checks.

**The natural repro works exactly as predicted from code**: with the pool 2 deep, three rapid `open_new_window` calls gave `window-pool-e28a2d44`, `window-pool-0f4cbca6`, then `window-d3d6dfe0` — the third with no `pool-` prefix, i.e. genuinely the cold path. No local patch needed, so the reachability argument in the PR body stands.

**But the cold-path window is still hidden.** `IsWindowVisible(0x2e20b14) == False` for `window-d3d6dfe0` (title "Window 2"), ~40s after it reported first paint. The other two were visible until the pool demoted them back, which is by design.

**Why the fix missed:** neither of my new log lines appears in the host log — not the `warn` for the Views/native mismatch, not the `debug` for the already-visible skip. So `on_load_end`'s `if !is_pool_window` block never executed for this window at all. Neither did its `scheduling show() retry` warn, nor the `window will stay hidden` error. I fixed a predicate inside a branch that does not run on this path; the real gap is upstream of every site I touched.

The three-site change is still correct on its own terms (the retry path really was inconsistent, and `try_show_resolved_window` really can resolve-without-showing), so I would rather keep it than drop it — but it is not the fix for #3028, and shipping it under that title would close the issue on a false premise.

Next step is to find why `on_load_end` does not reach the reveal block for a cold-path window in background-service mode with zero visible windows. I will re-open with an actual root cause.

GenericAgentX-asaf:
## Root cause found — and my first diagnosis was wrong

Reopening with an actual root cause, as promised. The short version: **the bug has nothing to do with CEF Views misreporting visibility.** A cold-path window never reaches the reveal block at all, because it inherits `is_browser_pane = true` from a cloned floater client and returns out of `on_load_end` above it.

### How I got here

I had claimed "`on_load_end`'s `if !is_pool_window` block never executed — none of my log lines appear." Two things were wrong with that reasoning:

1. **DEBUG is filtered out of the host log** (only INFO/WARN/ERROR present). My "already visible — reveal skipped" line is `tracing::debug!`, so its absence proved nothing.
2. **The `views_visible == false` branch logs nothing at all**, so absence of logs was consistent with several different paths.

The absence of logs was real, but it didn't identify the branch. What does identify it is a line I hadn't looked for — `[startup-paint] revealing gated window` fires for `main` twice in that log and never for the cold-path window, which rules out the gate ever being armed.

### The proof

Host log of the **original** #3028 repro (`~/.agentmux/dev/main/.../agentmux-host-v0.55.37.log.2026-09-06`), at `20:13:19.309` — the same `on_load_end` invocation as the cred injection, for `window-7f73cb1b…`, registered as `TopLevel { is_pool: false }`:

```
Injected IPC port 61928 into page: …windowLabel=window-7f73cb1b…
[pane-load-end] pane page loaded; reinstalling focus subclass
[pane-subclass] subclassed child HWND 0xa10cae class=Chrome_RenderWidgetHostHWND
[pane-load-end] couldn't resolve block_id for nav-state emit         (WARN)
[pane-loading-state] couldn't resolve block_id for nav-state emit    (WARN)
```

A top-level window ran the **browser-pane path**, and the pane code could not resolve a `block_id` *because it is not a pane*. That is the `is_browser_pane` early-return firing, observed rather than inferred.

Source of the flag: `CreateWindowTask::execute` clones its CEF `Client` from an arbitrary browser in `list_top_level_browsers()`, which deliberately includes `Floater`s — and every floater client is built with `is_browser_pane = true`. The only floater alive at creation time was `floating-pool-55a5994…`.

### Why the three-rapid-clicks run "failed to reproduce"

It didn't reproduce because **the bug wasn't present in that run**. `browsers` is a `HashMap`, so which client gets cloned is arbitrary but stable within a process. In that session the clone picked a non-floater client and the window was shown normally — `[wrr] callback event=0x3` (`EVENT_SYSTEM_FOREGROUND`) on its HWND at creation, and both visibility layers read visible at load-end.

So my "the fix didn't work" conclusion was drawn from a run where there was nothing to fix. That was the right call for the wrong reason: the fix genuinely wasn't the fix, but the evidence I used to reject it didn't show that.

### What changed

- `on_load_end` decides the pane branch from the browser's own registered `BrowserKind` (`takes_pane_path`), not the shared client flag. Unresolvable kind still falls back to the flag. Warns when they disagree.
- `CreateWindowTask` prefers a genuine `BrowserKind::TopLevel` browser to clone from, falling back to the floater-inclusive list rather than aborting.

Falsified: reverting `takes_pane_path` to flag-only behaviour fails `a_top_level_window_with_an_inherited_pane_flag_is_not_a_pane`. `cef 396 passed`.

Also fixed a pre-existing dangling spec citation in `ui_tasks/window.rs` that the citation gate surfaced once my edit made that file "changed".

### What I have NOT done

**I have not run the issue's exact repro (background-service mode, all windows closed, empty pool) against this build.** The root cause is established from the recorded log of the run where the bug actually fired, which is stronger than the hypothesis I shipped last time — but it is not the same thing as watching the window appear. Last round I let an approval and four green checks substitute for that, so I'd rather name the gap than let it pass quietly. Happy to run it before merge if you want that gate.

GenericAgentX-asaf:
## Live verification: PASS — the gap I flagged is now closed

Ran against this build (`c8c91cd`), `task dev` with `AGENTMUX_TRAY=1 AGENTMUX_BACKGROUND_SERVICE=1`, no local patch of any kind. Pool drained naturally by concurrent `open_new_window` IPC bursts, so every cold-path window below is genuine (non-`window-pool-` label returned by the IPC call itself).

**4 of 4 cold-path windows revealed and are natively visible.**

Batch 1 — `{"cmd":"open_new_window"}` ×4 concurrent → 2 pool promotes + 2 cold path:

```
13:32:28.785 window-b64f63b9…  [startup-paint] frontend reported first paint
13:32:28.785 window-b64f63b9…  [startup-paint] revealing gated window
13:32:28.883 window-54f68047…  [startup-paint] frontend reported first paint
13:32:28.883 window-54f68047…  [startup-paint] revealing gated window
```

Batch 2, after closing every window (background-service kept the service alive) and re-draining the refilled pool — ×6 concurrent → 4 pool promotes + 2 cold path:

```
13:34:22.671 window-29e99fa7…  [startup-paint] revealing gated window
13:34:22.703 window-aff1fc12…  [startup-paint] revealing gated window
```

`revealing gated window` is precisely the line that never appeared for `window-7f73cb1b…` in the broken run — the paint gate is now armed on the cold path, so the first-paint signal has something to fire against.

Native check via `IsWindowVisible` (the assertion that failed in the issue, where it stayed `false` for ~60s):

```
window-54f68047  IsWindow=True  IsWindowVisible=True  rect=(818,491) 2150x1176
window-b64f63b9  IsWindow=True  IsWindowVisible=True  rect=(818,491) 2150x1176
window-29e99fa7  IsWindow=True  IsWindowVisible=True  rect=(90,90)   2240x1266
window-aff1fc12  IsWindow=True  IsWindowVisible=True  rect=(90,90)   2240x1266
```

The poison client was present throughout — `floating-pool-03f989e0…` (`Floater { is_pool: true }`) was registered for the whole session, so `list_top_level_browsers()` included it on every one of these calls. The new `list_full_top_level_browsers()` preference is what deterministically avoided it.

**Note on the new warn line:** `[on_load_end] client claims browser-pane but this browser is a top-level window` fired **0 times**, which is the expected result rather than a miss. With the client-selection half of the fix in place, the flag can no longer be wrong while any genuine `TopLevel` browser exists — and pool windows always qualify, so that holds even in the resting background-service state. The `takes_pane_path` half is the backstop for a floater-only population; it's covered by the falsified unit test (`a_top_level_window_with_an_inherited_pane_flag_is_not_a_pane`) rather than by this run.

CI green on all four checks. Taking this out of draft.

PR #3031, opened 2026-09-06
Manoz@Area54: docs: DRY and modularity audit — findings and a phased slimming plan

GenericAgentX-asaf:
## Summary

A measured audit of duplication and modularity across the six Rust crates and the frontend, with a phased plan ordered by payoff ÷ risk. Docs-only.

**Method:** `jscpd` over `frontend/` (2.8% duplicated, 122 clones) and all Rust crates (3.2%, 361 clones), with **every headline pair independently re-verified by `diff`** — and, for renamed twins, by identifier-normalized diff, since plain `diff` hides copy-rename duplication. Plus symbol-collision scans, a dependency-graph check, and a grep of the codebase's own "keep in sync" comments.

## Headline

The codebase is **not** copy-paste heavy, and frontend helpers are genuinely centralized. What it has is *structural* duplication with five root causes — and it annotates them itself: **334 "keep in sync" / "mirror" comments across 210 distinct files** (about one in six).

| # | Root cause | Evidence |
|---|---|---|
| 1 | No codegen across the Rust↔TS boundary | 322 hand-maintained RPC stubs; 2,819-line hand-maintained `gotypes.d.ts`; no generator script |
| 2 | `agentmux-common` isn't used as a common | 8K lines shared vs 300K Rust. `CREATE_NO_WINDOW` in 21 files — declared *privately, twice, inside common itself*. `now_ms` ×20, absent from common. `event_log.rs` copied whole between launcher and srv (415 lines, 56 comment-only diffs). Process-kill in 5 places / 3 mechanisms. Dep graph is a clean star → zero cycle risk |
| 3 | Whole-file platform forks | `.platform` resolver swaps entire files; `zoom.{win32,linux,darwin}.ts` differ by **9 comment lines**; `TileLayout` carries ~3,100 dup lines |
| 4 | Twin primitives by copy-rename | `mcp` ↔ `skill` are 58–71% structurally identical across all four layers |
| 5 | Parallel frameworks sharing only vocabulary | saga + reducer exist in both launcher and srv; the saga `mod.rs` files share exactly one method name |

## What's deliberately *not* flagged

Measured and excluded as not-duplication: `host_spawn`/`container_spawn` (1239/1485 lines differ), `useWindowDrag` variants (genuinely different), trait-impl fan-out. Migrations copying live logic is flagged as **probably correct but undocumented** — recommendation is to write the policy down, not remove the copies.

## Plan

Five phases, each independently shippable. Phase 1 is six mechanical lifts into `agentmux-common` removing ~1,200 lines at zero behavior change; Phase 2 (codegen) is the highest-leverage; Phase 5 (saga/reducer) is a decision, not a refactor. Implementation PRs to follow.

## Verification

- `scripts/check-doc-status.sh` / `check-spec-citations.sh` — both pass
- Two from-memory claims verified against source before filing (`effective_mcp_servers` doc says it "mirrors `effective_skills`"; `persistent_resume.rs` is a companion of `persistent.rs`)

### DIFF
--- a/.changesets/1788729452-docs-dry-and-modularity-audit-findings-and-a-phased-slimming-xg3c.md
+++ b/.changesets/1788729452-docs-dry-and-modularity-audit-findings-and-a-phased-slimming-xg3c.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: DRY and modularity audit — findings and a phased slimming plan
--- a/docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md
+++ b/docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md
@@ -1,0 +1,215 @@
+# DRY and modularity audit — findings and a slimming plan
+
+**Status:** proposed
+**Date:** 2026-09-06
+**Author:** Manoz@Area54
+**Baseline:** `main` @ `734214c7f` (v0.55.37)
+
+**Ask (repo owner):** *"a rigorous, deep analysis of the code base for DRYness and clean modularity. Identify clear places where we can slim things down with DRY and, as an extension, how that may play into better modularization of the architecture."*
+
+---
+
+## 0. The short version
+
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

Issue #2939, opened 2026-09-03
Container / sandbox agents — umbrella tracking issue (multi-generation)


Agent5-asaf:
## Overview

Container ("sandbox") agents run the provider CLI inside a per-agent Docker container instead of as a host subprocess. The lifecycle and turn execution have worked for months; **everything around them did not**, and the gaps were found one at a time because nothing tracked them together.

This issue is the **work tracker** for finishing container agents. Discussion #2938 is the narrative hub. This issue is what you pick up work from.

**Read this whole issue before starting.** It is written for an agent with no prior context, and §"Approaches already rejected" in particular exists so you do not spend a day re-discovering a dead end empirically, as I did.

---

## How to use this issue

- Each workstream below is independently pickable. They do not have to be done in order, except where "Depends on" says so.
- When you finish one, **tick the box and edit this issue** with what you learned — especially anything that turned out to be false. This issue is the handoff between generations; a stale tracker is worse than none.
- Do not mark a box done on "it compiles" or "tests pass". Every item has an acceptance criterion that names an observable result. Meet it, or say plainly which part you left.

---

## What a container agent is today

- Agent definition with `agent_type: "container"`; block meta carries `agentMode: "container"`.
- `app_api/agent_open.rs` **forces `controller_type = "subprocess"`** regardless of provider — a container turn is one `docker exec`, which a long-lived persistent stdin cannot express.
- One persistent container per agent, named `agentmux-<agentId>`, image `ghcr.io/agentmuxai/agent-claude`. Turns are `docker exec`; env goes through the Docker socket, never argv (CWE-214).
- The image execs as `agent`, **uid 1000**, by design — "Host filesystem UID parity" (`SPEC_CONTAINER_PANE_SUPPORT_2026_06_11.md`).
- Credentials: the bound Armory account's `.credentials.json` is bind-mounted as a **single file** over the per-agent named volume at `/home/agent/.claude`. `/workspace` is a bind of the agent's `cmd:cwd`.

Key files: `agentmux-srv/src/backend/container.rs`, `backend/blockcontroller/subprocess/container_spawn.rs`, `server/agent_handlers/input.rs` (container branch), `server/app_api/agent_io.rs` (parallel branch — **keep the two in sync**).

---

## Verified vs unverified

Be careful here. Not everything below has equal evidence behind it.

| Claim | Evidence |
|---|---|
[...]

PR #3032, opened 2026-09-06
Camper@claudius: feat(app-api): agent.open over HTTP + OpenAgent MCP tool + muxopen — automation can launch agents, not just stop them

GenericAgentX-asaf:
## Summary

Automation could **stop** an agent (`FleetBulkStop` reaches cross-channel targets) but could not **start** one anywhere — `agent.open` existed only on the frontend's WebSocket RPC engine, with no HTTP mapping and no MCP verb. A destructive verb crossed the instance boundary before the constructive one existed at all.

Found while working the container-agent effort (#2939/#1400) under a hard no-GUI constraint: `Scouto`'s container had been **Up 5 days**, visible in `FleetList`, and `SendMessage` still failed with `agent not found` — an unspawned agent isn't in the delivery registry, and nothing an agent could reach would open it. Full diagnosis + live probe transcript in the included `docs/reports/REPORT_AGENT_OPEN_API_GAP_2026_09_06.md`.

## What this adds (same-instance scope)

**1. `POST /api/v1/agent/open`** — the RPC closure's ~550-line body is extracted into `open_agent_impl`, and *both* surfaces call it. That keeps the `AGENT_OPEN_LOCKS` TOCTOU serialization (reagent P1, PR #2059), ABF provider shadowing, resume-session seeding guard, and CLI resolution in exactly one place — same extraction shape `open_pane`/`handle_pane_open` already uses. Caller-fault errors (`AGENT_NOT_FOUND`/`INVALID_PROVIDER`/`CLI_NOT_AVAILABLE`) map to 400, everything else 500.

**2. `OpenAgent` MCP tool** — `agent_id` (name or definition id), optional `tab_id`/`focus`. Wired into `tools/list` and the drift-check array, following `FleetBulkStop`'s request pattern exactly. Idempotent like the RPC path (`created: false` on an already-open agent).

**3. `muxopen`** — third sibling to `muxlog`/`muxspect`: deployed to `~/.agentmux/shell/muxopen.mjs` with bash/zsh/fish/pwsh functions. `parseArgs`/`renderResult` are exported pure and vitest-covered (same arrangement as `muxspect.test.mjs`, per reagent P1 on PR #2380). A 404 from an older srv gets an explicit *"this instance may predate /api/v1/agent/open"* hint.

## Deliberately deferred (report §3.3/§4.1)

Cross-channel, LAN and WAN reach. Opening is **not** the mirror image of stopping — a remote open spawns a process, consumes provider tokens, and can bind credentials, and `agent_open`'s own concurrency guard documents that it cannot see a different instance racing the same registry entry. Cross-channel should follow `FleetBulkStop`'s resolution path as a follow-up; LAN is gated on the `ENFORCE_AGENT_BINDING` decision.

## Verification

- [x] `cargo test -p agentmux-srv` — **3021 passed** (2 new: route is auth-gated; unknown agent → 400 carrying the impl's `AGENT_NOT_FOUND` vocabulary, not a 500)
- [x] `cargo check -p agentmux-mcp` — clean
- [x] `npx vitest run …/muxopen.test.mjs` — 9 passed
- [x] Deploy test extended: `muxopen.mjs` lands beside `muxlog.mjs`/`muxspect.mjs`
- [x] **Live smoke against the running v0.55.34**: `node muxopen.mjs Scouto` → the 404 version-hint path, exit 2 — which is also independent confirmation the route genuinely didn't exist before this PR
- [ ] Live successful open — requires a restarted srv carrying this route; will be exercised as the first step of the #2939/#1400 container work this unblocks

### DIFF
--- a/.changesets/1788729917-feat-app-api-agent-open-over-http-openagent-mcp-tool-muxopen-xszb.md
+++ b/.changesets/1788729917-feat-app-api-agent-open-over-http-openagent-mcp-tool-muxopen-xszb.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(app-api): agent.open over HTTP + OpenAgent MCP tool + muxopen — automation can now launch agents, not just stop them
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -332,6 +332,20 @@ const FLEET_BROADCAST_TOOL: &str = r#"{
   }
 }"#;
 
+const OPEN_AGENT_TOOL: &str = r#"{
+  "name": "OpenAgent",
+  "description": "Launch an agent into a pane by name or definition id (get these from FleetList/DiscoverAgents). Idempotent: if the agent is already open in the target tab, returns its existing pane with created:false instead of opening a second one. This spawns a real provider process (and, for container agents, execs into their container) — it consumes provider tokens like any human-opened pane. Returns JSON {block_id, tab_id, agent_id, provider, controller_type, status, created}. After a successful open the agent becomes addressable for SendMessage.",
+  "inputSchema": {
+    "type": "object",
+    "properties": {
+      "agent_id": { "type": "string", "description": "Agent definition id or name (case-insensitive), e.g. from FleetList's host.agents[]" },
+      "tab_id": { "type": "string", "description": "Optional target tab id; defaults to the active tab" },
+      "focus": { "type": "boolean", "description": "Optional: focus the new pane (default true)" }
+    },
+    "required": ["agent_id"]
+  }
+}"#;
+
 const FLEET_BULK_STOP_TOOL: &str = r#"{
   "name": "FleetBulkStop",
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in `13e89c4cb` — legitimate, and worth spelling out why it mattered more than a cosmetic miss.

`focused: true` was hardcoded and `cmd.focus` never read anywhere. Harmless while the RPC surface was the only caller (nothing set the field), but this PR *documented* `focus`/`--no-focus` on OpenAgent and muxopen — so `--no-focus` had zero effect and the pane always stole focus. That's precisely the disruption a headless open exists to avoid on a machine a human is actively using, which is this PR's whole motivating constraint.

Now `focused: cmd.focus.unwrap_or(true)` — default unchanged (matches the UI flow), the flag honored. 3021 tests still green.

GenericAgentX-asaf:

Same state mismatch as PR #2990: your reply confirms the P1 is fixed in `13e89c4cb`, but the review *state* is still `CHANGES_REQUESTED`, leaving the PR `BLOCKED`. Requesting a formal pass so the state matches the verified outcome.

PR #610, opened 2026-04-30
Phase E.1b: srv reducer skeleton + pipe IPC

AgentA-asaf:
## Summary

Second PR of Phase E. Adds the second reducer in the multi-reducer architecture: srv now hosts its own pipe IPC server with a Redux-style reducer over its canonical state. **Plumbing-only** — domain commands (workspace, tab, block, layout) land in E.2+.

### Wire protocol

- New `Command::GetSrvSnapshot` (no args)
- New `Event::SrvSnapshot { version, lifecycle }` reply (sparse for E.1b; E.2+ adds workspaces/tabs/blocks/layouts)

### Srv-side (new modules)

- **`agentmux-srv/src/state.rs`** — `State`, `ProcessRecord`, `ProcessState` + `bump_version` / `alloc_client_id` helpers (mirrors launcher's pattern)
- **`agentmux-srv/src/reducer.rs`** — pure `update(state, cmd, ctx)`. E.1b arms: `Register` / `Goodbye` / `Ping` / `GetSrvSnapshot` / `GetEvents`. **9 unit tests** (lifecycle transition, dup-register, recycled-pid, snapshot, monotonic versions, etc.)
- **`agentmux-srv/src/event_log.rs`** — copy of launcher's. Disk path: `<data-dir>/srv-events.log`. **6 unit tests pass**
- **`agentmux-srv/src/srv_ipc/{mod.rs,server.rs}`** — pipe server adapted from launcher's. Differences: no host-only enforcement (srv accepts commands from any registered client per Phase E §5); no WRR logging; identity patching via `patch_srv_identity`
- **`agentmux-srv/src/main.rs`** — binds + spawns srv-ipc when launcher passes `AGENTMUX_SRV_PIPE_PATH`. Non-fatal on bind failure

### Launcher side

- **`agentmux-launcher/src/ipc/mod.rs`** — new `srv_pipe_name(hash)` helper
- **`agentmux-launcher/src/srv_spawner.rs`** — `spawn_srv` takes `srv_pipe_path: &str` and forwards via `AGENTMUX_SRV_PIPE_PATH` env. Launcher is sole authority for the data-dir hash
- **`agentmux-launcher/src/main.rs`** — pre-computes srv pipe path, passes to `spawn_srv`

### Bonus fix

- **`agentmux-srv/src/server/tests.rs:30`** — pre-existing test compile error (missing `process_tracker` field on `AppState`) fixed alongside, so `cargo test` works for E.1b's new tests

## What this does NOT do

[...]

PR #3033, opened 2026-09-06
Manoz@Area54: refactor(common): lift CREATE_NO_WINDOW, now_ms, event_log, process-kill, WindowKind into agentmux-common

GenericAgentX-asaf:
## Summary

Phase 1 of `docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md` (merged as #3031) — the mechanical lifts into `agentmux-common`. **Zero behavior change**, net **−1,028 duplicated lines** across 56 source files. The dependency graph is a clean star (everything → `common`, `common` → nothing), so every lift is cycle-free by construction.

| Lift | Before | After |
|---|---|---|
| `common::win32` | `CREATE_NO_WINDOW` re-declared privately in **23 files across 5 crates** — twice inside function bodies in common itself, exported to nobody | one `pub const`; 21 sites `use` it; the `#[cfg(windows)]` blocks are unchanged |
| `common::time` | the same 3-line `SystemTime` body in **26 files** (16 `i64`, 8 `u64`, one `chrono` variant) | four fns; each local `now_ms` kept as a one-line delegate so `Self::now_ms`, `pub(super)`, and free-fn call sites are all untouched |
| `common::event_log` | **415 lines, byte-for-byte in both launcher and srv**, differing only in the logging sink. srv's header literally carried the to-do: *"Phase E.7 cleanup: lift the shared parts into agentmux-common… (reagent P2 #610)"* | one module; `run_disk_writer` takes an injectable `WarnSink` (launcher has no `tracing` subscriber); each crate keeps a ~15-line shim so **no call site changed** |
| `common::process` | five independent kill impls across four crates; one described itself as a *"mirror of the cef-side helper"*; only one suppressed the console flash | `kill_pid` + `kill_process_group`; three sites replaced; every Windows path now suppresses the flash |
| `WindowKind` | cef held a **byte-identical** private copy of `agentmux_common::ipc::WindowKind` and mapped between them variant-by-variant | `pub use`; mapping deleted |

## Deliberately scoped out (reasons recorded in the report)

- **ObjC extern consolidation** — on inspection the externs are *not* uniform (`objc_getClass` returns `Class`/`Id`/`*mut c_void` at different sites with per-file aliases), and it's macOS-only code that can't be compiled or tested on this Windows machine. Needs a macOS build.
- **`Rect` / `DataPaths`** — the audit listed them as shadows; they're **name collisions** (cef's `Rect` is `{x,y,width,height: f64}`, common's is a Win32 `RECT`). The report's §2.2 row and §5 plan are corrected in this PR.
- **bashwrap's kill helper** — its Unix branch is an intentional no-op the code says not to change without a repro. Left alone.

## Verification

- `cargo check -p agentmux-srv -p agentmux-cef -p agentmux-launcher -p agentmux-bashwrap --tests` — clean
- **`cargo test --workspace -- --test-threads=1`** (the exact CI invocation) — **3,987 passed, 0 failed**
- `agentmux-common` gains tests for all four modules, including one proving `event_log` warnings route through the injected sink rather than a hard-coded logger
- Imports orphaned by the `now_ms` sweep were pre-detected by grep and removed — seven in the end. An earlier version of this description claimed the one the grep kept (`session_archive.rs`) was "confirmed still live"; that was wrong (ReAgent P2, fixed in `7ca8c4120`). `SystemTime` there is live, `UNIX_EPOCH` was not — my check matched the import line against both names as one pattern, so the live one masked the dead one.

## Review history (for the record)

Three real findings were caught and fixed during review, all in code this PR introduced or touched:
1. **P1 ×2** — `kill_pid(u32::MAX - 1)` in my own tests cast to `pid_t` and wrapped to `-2`, which `kill(2)` reads as *process group 2*; the group variant negated it to signal PID 2. Fixed at the source with a shared `checked_pid()` guard, not by picking a safer test input.
2. **P1** — the guard didn't reject `0`, so `kill_pid(0)` still reached `kill(0, …)` = the caller's own group. `checked_pid` now refuses 0 too.
3. **P2** — two more variant-by-variant `WindowKind` mappings remained after I claimed "the mapping is gone" (`creation.rs`, `launcher_ipc/mod.rs`). Removed; a sweep confirms none are left.
4. **P2** — the orphaned `UNIX_EPOCH` import above.
5. The spec-citations CI gate surfaced two citations that were **already dangling on main** in files the sweeps touched; one moved to `archive/` (repointed), one exists nowhere (path dropped per the gate's own rule).

## Review notes

- The `now_ms` delegates are intentionally *kept* as thin local fns rather than replaced with `use` lines — it's the only shape that leaves `Self::now_ms()` associated-fn call sites and `pub(super)` cross-module refs untouched. The duplicated *body* is what's removed.
- The two `event_log` shims exist so that `crate::event_log::EventLog` and the 2-arg `run_disk_writer` keep resolving; they're where each crate's warn sink is bound, once.


a5af:
@codex review

GenericAgentX-asaf:
Both P1s addressed in `1e4e34b20` — and they were worse than stated: the wrap wasn't confined to the tests. Any caller passing a PID above `i32::MAX` would have hit the same `as pid_t` wrap in production and signalled a process *group*.

Fixed at the source rather than by picking a safer test input:

- `checked_pid()` — `i32::try_from`, shared by both functions so the guard can't drift. `kill_pid` returns `InvalidInput`; `kill_process_group` is a no-op.
- `kill_process_group` additionally refuses `pgid <= 0` — `kill(-0, …)` is `kill(0, …)` is *the caller's own group*.
- The Unix arms now take the checked `i32`, so the `SAFETY` comments are actually true.

Tests rewritten to exercise the guard on every platform: the exact overflow inputs (`u32::MAX - 1`) must be rejected *before* any syscall; `pgid 0` must be a no-op; and the boundary `i32::MAX` (fits `pid_t`, cannot be a live process on Linux or Windows) is the real "nonexistent PID fails cleanly" case — asserting the error is **not** the guard's, i.e. the OS was actually asked and said no.

`cargo test -p agentmux-common`: 183 passed.

GenericAgentX-asaf:

GenericAgentX-asaf:

GenericAgentX-asaf:

GenericAgentX-asaf:

PR #3034, opened 2026-09-06
Manoz@Area54: refactor(frontend): collapse the three byte-identical zoom platform forks into one zoom.ts

GenericAgentX-asaf:
## Summary

Phase 3 of `docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md` (§2.3), smallest item first. `zoom.{win32,linux,darwin}.ts` were selected per-platform by the `.platform` import resolver — but measured, they differ by **nine lines, every one a comment**. The JavaScript is identical on all three platforms. **Zero behavior change.**

## What changed

- **`frontend/app/store/zoom.ts`** — the win32 body verbatim, with a merged header. The three headers each carried a *different* platform rationale for the same rule, and that knowledge is real, so it's kept rather than lost:
  - Windows: `calc(100vw / var(--zoomfactor))` in `window-header.scss`; setting width from JS breaks it
  - macOS: `width: 100%` in `window-header.darwin.scss`; do **not** use the `calc` form — it double-divides on WebKit and the window buttons drift
  - Linux: WebKitGTK doesn't divide flex space by zoom, so no compensation at all
  - Shared rule on all three: JS sets only `--zoomfactor`; width compensation stays in CSS
- **Deleted** `zoom.win32.ts`, `zoom.linux.ts`, `zoom.darwin.ts`, and `zoom.platform.ts` (the `tsc` typecheck stub that re-exported `./zoom.win32` — dead once nothing imports through the resolver)
- **Rewrote** the six `zoom.platform` import sites (five static + the dynamic `import()` in `app-init.ts`) to `zoom`, plus two comment references

The `.platform` resolver itself is untouched and still serves the four families that genuinely differ (`TileLayout` ~50% shared, `CrossWindowDragMonitor`, `useWindowDrag`, `window-controls`) — those are a different refactor shape and not in this PR.

## Verification

- `tsc --noEmit` — clean
- `vitest run frontend/app/store frontend/app/element frontend/app/view/helpview` — **823 passed** (29 files)
- Grep confirms zero remaining references to `zoom.platform` / `zoom.{win32,linux,darwin}` outside `zoom.ts`'s own header


a5af:
@codex review

PR #667, opened 2026-05-03
refactor(cef): rename Pane → BrowserPane in reducer/state types (task #181)

AgentA-asaf:
## Summary

Resolves the long-standing ambiguity where reducer.rs / state.rs used bare \`Pane\` to mean specifically embedded-browser child windows (\`defwidget@browser\`). The IPC commands (\`browser_pane_*\`) and \`BrowserPaneManager\` were already explicit; only the host-internal types were abbreviated. A first-time reader of the reducer code would reasonably wonder if \`Pane\` meant terminals, agent panes, or browser panes — see the PR #664 retro session, 2026-05-02:

> \"is the pane terminology PaneCreate .. is that a terminal, agent pane, created by widgets?\"

Pure mechanical rename. Zero behavior changes. All 63 host tests pass.

### Host-internal (renamed)

| Category | Before | After |
|---|---|---|
| HostCommand variants | \`TryRegisterPaneLive\`, \`EnqueuePaneCreate\`, etc. (7 total) | \`TryRegisterBrowserPaneLive\`, etc. |
| HostEvent variants | \`PaneCreateRequested\`, \`PaneLive\`, etc. (5 total) | \`BrowserPaneCreateRequested\`, etc. |
| State types | \`PaneEntry\`, \`PaneLifecycle\` | \`BrowserPaneEntry\`, \`BrowserPaneLifecycle\` |
| HostState field | \`panes\` | \`browser_panes\` |
| DispatchOutput fields | \`pane_register_result\`, \`closed_pane_label\`, \`drained_block_id\` | \`browser_pane_register_result\`, etc. |
| AppState helpers | \`live_pane_label\`, \`live_pane_labels\`, \`any_pane_closing\` | \`live_browser_pane_label\`, etc. |
| Reducer handler fns | \`handle_enqueue_pane_create\`, etc. (8 total) | \`handle_enqueue_browser_pane_create\`, etc. |
| Pane callbacks | \`on_before_close_pane\`, \`on_after_created_pane\` | \`on_before_close_browser_pane\`, etc. |
| Module dir | \`src/pane/\` (4 files) | \`src/browser_pane/\` (git mv preserves history) |

### NOT renamed (still bare \"pane\" — intentional)

- \`BrowserPaneManager\` (already explicit)
- IPC command names: \`browser_pane_create\` / \`navigate\` / \`focus\` (already explicit)
- Label format: \`browser-pane-<id>-<seq>\` (already explicit)
[...]

PR #3035, opened 2026-09-06
fix(browser-pane): route pane nav-state/title/favicon events to the pane's owning window, not main

AgentX-asaf:
## Summary

A browser pane torn off into a second window (or opened in any window other than `main`) gets stuck on the loading brain overlay forever: its address bar and tab title never update, while navigation itself still works. Root cause: the five host→page push sites for browser panes (`browser-pane-nav-state` ×3, `browser-pane-title-change`, `browser-pane-favicon-urls`) all used `emit_event_from_state`, which always delivers into the window labelled `main`. After a tear-off the pane's view model lives in the new window's renderer, so every event was dropped silently — `main` resolves fine, nobody there listens for that `block_id`.

Full diagnosis with the host/frontend log timeline, bisection, and an open-issue canvas: `docs/reports/REPORT_BROWSER_PANE_EVENTS_MISROUTED_TO_MAIN_WINDOW_2026_09_06.md` (in this PR).

## Fix

- New `events::emit_browser_pane_event(state, block_id, event, payload)`: resolves the pane's owning window from its reducer entry (`HostState.browser_panes[block_id].window_label`, written by `TryRegisterBrowserPaneLive` before the CEF browser exists) and delivers via `emit_event_to_window`. Falls back to the legacy `main` routing, with a warning, only when the pane has no entry or an empty `window_label` (the legacy enqueue path) — nothing that worked before regresses. The reducer lock is dropped before emitting since `emit_event_to_window` re-locks.
- The routing decision is a pure `browser_pane_event_target(&HostState, block_id)` so reducer tests can pin it without CEF handles.
- All five sites rerouted. Same shape PR #2597 used for `browser-pane-clicked` / shortcut / context-menu.
- Targeted delivery rather than broadcast to all top-level windows, on purpose: `state/mod.rs` documents per-window routing so a hostile page in one pane cannot observe another pane's traffic.

No frontend change — the frontend already filters by `block_id`.

## Not a 0.55.37 regression

`git log -S` dates the nav-state site's `emit_event_from_state` call to 2026-05-02 (#667); #2642 re-touched it without changing routing. Latent since May; surfaces whenever a browser pane lives outside `main` while `main` is still open.

## Tests

- `reducer::tests::browser_pane_event_target_is_the_owning_window_not_main` — pane registered in `window-pool-…` resolves to that label.
- `reducer::tests::browser_pane_event_target_is_none_for_unknown_or_windowless_pane` — unknown block and legacy empty-label entry both fall back.
- `cargo test -p agentmux-cef`: 391 passed.

## Follow-ups

- Live-verify in a rebuilt host (tear a browser-pane tab into a second window; overlay must clear, URL/title must update), then re-run #768's tear-off repro — its "black browsers in the new window" symptom was never re-reproduced on pool-served tear-off and may be this.
- `emit_event_from_state` has other callers outside the pane paths; any with per-pane/per-window payloads are candidates for the same bug. Out of scope here.

### DIFF
--- a/.changesets/1788731291-fix-browser-pane-route-pane-nav-state-title-favicon-events-t-upti.md
+++ b/.changesets/1788731291-fix-browser-pane-route-pane-nav-state-title-favicon-events-t-upti.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(browser-pane): route pane nav-state/title/favicon events to the pane's owning window, not main
--- a/agentmux-cef/src/browser_pane/callbacks.rs
+++ b/agentmux-cef/src/browser_pane/callbacks.rs
@@ -212,8 +212,9 @@ pub(crate) fn set_pane_main_frame_loading(
         loading,
         url,
     );
-    crate::events::emit_event_from_state(
+    crate::events::emit_browser_pane_event(
         state,
+        block_id,
         "browser-pane-nav-state",
         &serde_json::json!({
             "block_id": block_id,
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #3036, opened 2026-09-06
Manoz@Area54: refactor(cef): one raw-TCP service_call transport for the ten backend_* helpers

GenericAgentX-asaf:
## Summary

`agentmux-cef/src/client/helpers.rs` carried **eleven** hand-rolled copies of the same raw-TCP `POST /agentmux/service` sequence — build the JSON body, format the HTTP/1.1 request with `X-AuthKey`, `connect_timeout` 2000 ms, write, read — one ~25-line block per function. It was the largest single-file duplication cluster in the Rust workspace (`docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md` §3.2). **Zero behavior change**; `helpers.rs` goes from 895 to 583 lines.

## What changed

- **New `client/service_call.rs`** — the transport only: `service_body()`, `service_call()` returning a `ServiceResponse` with the two views every caller wanted (`first_line`/`status_ok`, `data_if_success`), and a granular `ServiceCallError { Connect, Write, Read }`.
- **It does no logging on purpose.** Every wrapper keeps its own `tracing::warn!`/`error!` lines byte-for-byte, because several are operator-facing and cited by specs and retros. The error enum is granular precisely so callers can keep distinguishing connect vs. write vs. read failures the way they already did — the prior copies were *not* uniform on that (setters ignored read errors and reported the empty status line as "did not succeed"; getters returned `None` silently), and that asymmetry is preserved by the callers rather than flattened.
- **Ten wrappers rewritten** — five setters (`set_window_opacity`, `set_window_pos_and_size`, `set_window_topology`, `update_block_meta`, `register_ipc_with_backend`) and five getters (`get_window_opacity`, `get_window_pos_and_size`, `get_client_window_ids`, `find_window_by_label`, `get_window_topology`). Signatures unchanged; the `client/mod.rs` re-exports are untouched.
- **`backend_close_window` is deliberately left as-is.** It interleaves `dlog` tracing between the write and the read, escalates to `error!`, and gates a user-visible close path with `SPEC_WINDOW_LIFECYCLE_CLOSE_RELIABILITY_2026_07_04.md` behind it. Keeping it self-contained keeps that reliability story reviewable on its own.

## One documented near-equivalence

On a *mid-response* read error, the setters used to log whatever partial status line had arrived before the error (the `read_to_string` buffer contents are unspecified after an error). They now log the `(empty)` placeholder. Log text only, in a corner case; getters returned `None` silently on read error before and still do.

## Verification

- `cargo check -p agentmux-cef --tests` — clean (all warnings pre-existing)
- `cargo test -p agentmux-cef service_call` — **5/5**: body shape, the 200-check, `data_if_success` on success/failure/malformed, a **real `TcpListener` round trip** asserting the auth header and body on the wire, and distinct connect-failure reporting. The eleven prior copies had no transport test at all.
- The rewrite was generated from a table + two templates keyed on each exact signature, so a mismatch would have failed loudly rather than silently skipping a function (10/10 matched).

Independent of #3033 (Phase 1) — disjoint files.

## Review history (for the record)

ReAgent's first pass caught three things, all fixed in the follow-up commit:

1. **P1 — no changeset.** Added. My omission.
2. **P2 — one log line was not byte-for-byte.** The setter template built the "did not succeed" message from the RPC *method* name, so `register_ipc_with_backend`'s `host_ipc.Register did not succeed` briefly became `Register did not succeed`. Restored. A mechanical diff of every `"[backend_…]"` / `"[register_…]"` string literal on `main` vs. this branch confirms this was the **only** divergence across all ten functions — the other nine match exactly.
3. **P2 — a dropped invariant comment.** The getter template emitted `if width <= 0 || height <= 0 { return None; }` bare, losing the five-line explanation of *why* (never-written `Window.pos`/`winsize` default to zero via `#[derive(Default)]`, indistinguishable from "row predates the write-through"). Restored inside the block where it lived.

### DIFF
--- a/.changesets/1788732819-refactor-cef-one-raw-tcp-service-call-transport-for-the-ten--4osj.md
+++ b/.changesets/1788732819-refactor-cef-one-raw-tcp-service-call-transport-for-the-ten--4osj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(cef): one raw-TCP service_call transport for the ten backend_* helpers in client/helpers.rs
--- a/agentmux-cef/src/client/helpers.rs
+++ b/agentmux-cef/src/client/helpers.rs
@@ -384,6 +384,11 @@ pub(crate) fn backend_get_window_pos_and_size(web_endpoint: &str, auth_key: &str
     let width = data.get("winsize")?.get("width")?.as_i64()?;
     let height = data.get("winsize")?.get("height")?.as_i64()?;
     if width <= 0 || height <= 0 {
+        // Never-written Window.pos/winsize default to zero (Point/WinSize's
+        // #[derive(Default)]), which is indistinguishable from "this row
+        // predates the write-through" — treat a zero-or-negative size as
+        // "no real geometry persisted" rather than recreating a
+        // zero-sized window.
         return None;
     }
     Some(agentmux_common::ipc::Rect {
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

GenericAgentX-asaf:

PR #3039, opened 2026-09-07
Manoz@Area54: refactor: fold same-file clones (flyoutmenu, file-tree, PreLaunchAuthPanel), migrations freeze-copy policy, DRY audit corrections

GenericAgentX-asaf:
## Summary

The "anytime — contained cleanups" items from `docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md` §5, the two same-file frontend clones from §3.2, and one correction to the report itself. **No behavior change.** Net −93 lines.

| File | What was duplicated | Now |
|---|---|---|
| `frontend/app/element/flyoutmenu.tsx` | `MenuBody` and `SubMenu` each carried the same 85-line `<For>` row renderer — same hover-intent wiring, same item markup, same nested `<SubMenu>` hand-off — differing only in the row-key scheme (`${i}` vs `${parentKey}-${i}`) and the `parentKey` passed to `handleMouseEnterItem`. The 13-line pass-through prop list was declared twice as well. | one `MenuRows` component renders every level; `MenuLevelProps` is the one prop list. 589 → 509 lines |
| `frontend/app/view/editor/file-tree.tsx` | the root `<TreeNode>` and the recursive one re-listed the same twelve pass-through props | both spread `{...props}` and set only the five per-node values. 366 → 350 |
| `frontend/app/view/agent/components/PreLaunchAuthPanel.tsx` | the first and last `<Match>` arms rendered an identical `<ConnectCta …/>` | one arm keyed on `showConnectCta()`; the JSX comment explaining the stale-`ready` case moved next to the predicate. 833 → 822 |
| `agentmux-srv/src/migrations/mod.rs` | — (report step 18) | the module doc now states that migrations deliberately freeze copies of live helpers and must not be "deduplicated", and names the pairs |

## The report correction

The audit called `PreLaunchAuthPanel.tsx` "a 378-line render tree appearing twice … largest single clone in the frontend." Re-measuring with the right tokenizer shows that was an artifact: `jscpd`'s `javascript` mode mis-parses TSX and matched a 377-line span against a 12-line one, while `--format tsx` finds no clone in the file at all. The real duplication was the two 7-line `<ConnectCta>` invocations above. §3.2 and step 15 are corrected in this PR rather than left standing, and a progress note records what #3033 / #3034 / #3036 landed. The clone pairs named in the new policy paragraph (report §3.4) were re-verified with `jscpd --format rust` before being written down.

## Why the folds are safe

- **flyoutmenu** — the transform script diffed the two `<For>` blocks after normalising names and refused to run unless zero non-comment lines differed (zero did). The per-level `createPeerRegistry()` moves into `MenuRows`, so there is still exactly one registry per level. Row keys are unchanged; they are what `hoveredItems` / `visibleSubMenus` / `subMenuPosition` are keyed on.
- **PreLaunchAuthPanel** — `<Switch>` renders the first matching arm. The folded arm matches `unauthenticated`, `expired`, and stale-`ready`; the remaining arms match generic `ready`, `waiting`, `authenticated|saving`, `failed`. The sets are disjoint, so every controller state renders the arm it did before. The script asserted both `<ConnectCta>` elements were prop-for-prop identical before folding.
- **file-tree** — Solid's `{...props}` keeps the getters reactive and the five explicit props win. The two extra `FileTreeProps` callbacks (`onToggleHidden`, `onStartRename`) ride along unused at the root, as any spread would.

## Verification

- `tsc --noEmit -p tsconfig.json` — clean
- `vitest run` on the launch-modal integration test and the auth-flow / auth-state suites — **3 files, 77/77 passed**
- Prettier: all three files were already non-conforming on `main`; the edited lines follow the surrounding style and no unrelated hunks were reformatted (same rule as #2984).

Follow-up to #3031 (the audit), #3033, #3034, #3036.

### DIFF
--- a/.changesets/1788750245-dry-contained-cleanups-one-menurows-renderer-in-flyoutmenu-t-plab.md
+++ b/.changesets/1788750245-dry-contained-cleanups-one-menurows-renderer-in-flyoutmenu-t-plab.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+DRY contained cleanups: one MenuRows renderer in flyoutmenu, TreeNode prop pass-through in file-tree, single Connect-CTA arm in PreLaunchAuthPanel, migrations freeze-copy policy, audit report corrections
--- a/agentmux-srv/src/migrations/mod.rs
+++ b/agentmux-srv/src/migrations/mod.rs
@@ -14,6 +14,21 @@
 //! (use the `NNNN_slug` convention). Retiring old migrations: remove from
 //! [`REGISTRY`] once the minimum supported upgrade path passes the migration's
 //! origin version — the `db_migrations` row stays as a permanent record.
+//!
+//! # Migrations freeze copies of live logic on purpose
+//!
+//! Some migrations carry verbatim copies of helpers that also exist in live
+//! code: `m0017_ambient_login_grandfather` mirrors ~35 lines of
+//! `backend/blockcontroller/core.rs`; `m0020_agent_color_backfill` and
+//! `m0021_backfill_agent_bundles` mirror 25–30 lines each of
+//! `backend/storage/mcp_servers.rs` and `def_registry_mirror.rs`. That is
+//! deliberate, not accidental duplication. A migration must produce the same
+//! rows on every machine it ever runs on, so it cannot call a live helper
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #3037, opened 2026-09-07
feat(tray): macOS menu-bar backend + headless AppKit pump (issue #2977 WS1)

AgentO-asaf:
Issue #2977 Workstream 1, macOS half. Design doc: `docs/specs/SPEC_TRAY_OPTIONAL_BACKGROUND_SERVICE_2026_09_04.md` (§2, §7.5, §7.5.1).

## What

Two commits, deliberately separable so the Linux agent can cherry-pick the first:

1. **`refactor(tray)` — shared wiring.** The retrying host forwarder, the reachability probe and the actions thread were private to the Windows supervisor/backend. Hoisted into `tray/mod.rs` (`forward_tray_cmd`, `service_reachable`, `spawn_action_loop`, `TrayAction::host_cmd`) and the Unix supervisor now calls `spawn_action_loop` right after the single-instance socket bind, same placement as Windows. No behaviour change on Windows. **A Linux backend now only needs `tray/linux.rs` (ksni), the dep, and one `cfg` arm in `start_if_enabled`.**
2. **`feat(tray)` — macOS backend.** `tray/macos.rs`: an `NSStatusItem` with the shared New Window / Quit menu via tray-icon + muda, template (alpha-mask) glyph derived from the brand icon.

## Why the macOS backend owns no thread

AppKit requires the status item to be created *and* driven on the main thread (tray-icon enforces this with `MainThreadMarker`). The launcher's supervisor runs on a worker thread on macOS precisely so the main thread belongs to AppKit, and that thread already pumps `NSApplication` for the process lifetime (`splash_mac::pump_app_events`). So: `spawn` (supervisor thread) queues a request; the pump calls `tray::macos::pump_tick()` each iteration, which creates the item, then drains muda/tray-icon events and reachability changes. Inverse of the Windows backend, which had to *introduce* a pump on its own thread.

## Closes design doc §7.5.1 (macOS gap)

With the splash disabled, background-service mode had no `NSApplication` at all: no reopen delegate, nowhere for a menu-bar item to live. `main` now routes splash-disabled + `AGENTMUX_BACKGROUND_SERVICE` through `splash_mac::prepare_headless_app()` + `pump_forever()` on the main thread with the supervisor on a worker thread (`spawn_supervisor_thread`, factored out of the existing splash path). Ordinary splash-disabled mode is unchanged. The backend refuses to start if no pump was declared, so the log can't claim "started" for an icon nothing would ever service.

## Verified

[...]

PR #3040, opened 2026-09-07
docs(tray): implementation status in the design doc (issue #2977)

AgentO-asaf:
Docs only. `SPEC_TRAY_OPTIONAL_BACKGROUND_SERVICE_2026_09_04.md` still said "design only, nothing implemented" and described the §7.5.1 macOS gap as open. Adds a §7.6 status table pointing at the PRs, marks §7.5.1 closed by #3037, and records the one macOS event-loop finding that belongs next to §7.5 (status item must be created on the main thread, so the macOS backend owns no thread; `rect()` is meaningless on the first tick). Issue #2977 keeps the box-level detail.


a5af:
@codex review

Issue #774, opened 2026-05-10
Tab content reveal gate — eliminate piecemeal switch jank


AgentA-asaf:
## Symptom

After clicking a tab (or Ctrl+Tab), the new tab's content appears in 3-5 visible stages over ~150-300ms — title bar updates first, then pane frames, then block content streams in pane-by-pane, then late-arriving content (markdown, syntax highlight, browser HWNDs) finishes. User perceives this as visual jank.

This is a general issue across all pane types — not specific to the agent pane.

## Root cause

SolidJS reactive components mount independently as their data resolves. There's no orchestration that says "delay paint until everything is ready." Right default for streaming, wrong default for atomic tab switch.

## Proposed fix

Frame-budget gate using Long Tasks observer. Apply `visibility: hidden` to the tab container during a settling period (~80ms of clean frames or 800ms hard cap). Reveals atomically once content has stabilized.

**Why this approach:** No per-pane coordination needed. Reuses existing Long Tasks observer in `frontend/perf/observers.ts`. Self-tuning — fast tabs reveal quickly, slow tabs naturally take longer.

## Spec

`docs/specs/SPEC_TAB_CONTENT_REVEAL_GATE.md` on branch `agenta/spec-tab-content-reveal-gate`.

Covers: 4 design options (A frame-budget, B promise-based, C skeletons, D fixed-duration), recommendation, implementation outline (~115 LOC), edge cases (rapid switches, first-paint, background activity), and effort estimate (~0.75 day).

## Cross-references

- PR #772 — tab-switch perf marks (measures total switch duration; this gate addresses the *visual quality* during that duration)
- PR #773 — agent pane virtualization (reduces per-tab Long Tasks the gate is waiting on)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3041, opened 2026-09-07
Manoz@Area54: refactor(layout): one TileLayout core with per-platform hook objects instead of three near-identical copies

GenericAgentX-asaf:
## Summary

Phase 3 step 11 of `docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md`: `TileLayout.{win32,linux,darwin}.tsx` were three copies of the same component — **437 of 598 lines byte-identical across all three** — and the largest duplication cluster in the frontend. The shared body now lives once in `TileLayout.core.tsx` as `createTileLayout(platform)`; each platform file keeps only a `TileLayoutPlatform` object and its own `ResizeHandle`. **Zero behavior change per platform**; 1,652 → 957 lines (−695).

## What each platform still owns

Every member of `TileLayoutPlatform` is exactly one thing that differed between the old copies, and it is called exactly where that code was:

| Hook | win32 | linux | darwin |
|---|---|---|---|
| `animateDelayMs` | 150 (Issue #774 / reveal gate) | 50 | 50 |
| `onRootMount` | Win11 `dragend` safety net | — | — |
| `boundsCheckRequiresTileDrag` | — | ✓ (tab DnD crashed `persistToBackend`) | — |
| `onLeafGeometryChange` | `notifyPaneReflow` (native panes re-`SetWindowPos`) | — | — |
| `clearDraggingWhenInactive` | ✓ (after the safety net) | — | — |
| `rejectDragAt` | resize-handle-zone guard | — | — |
| `beforeDragStart` / `beforeDrop` | — | `preventUnhandled.start()` / `.stop()` | same as linux |
| `afterDragStart` / `afterDrop` | — | `setJsDragActive(true)` / `(false)` | — |
| `ResizeHandle` | pointer capture | `window` listeners (pointerId is unstable on CEF/Wayland) | capture + `body.cursor` workaround |

The drag **registration** (`draggable()` on the live header found by `querySelector`, 100 ms re-register poll, never torn down mid-drag) was already identical on all three engines — the three header comments each explained the same `dragHandle` breakage for WebView2 / WebKitGTK / WKWebView — so it moves to the core as is, with one comment covering all three.

## How I checked equivalence

- The split was designed from the three pairwise `diff`s, not from memory: every hunk in win32↔linux and linux↔darwin maps to one row above or to a comment-only difference. The mapping is in the commit message.
- Hook call sites preserve order: `beforeDragStart` is the first statement of `onDragStart` and `afterDragStart` the last, matching where linux/darwin had `preventUnhandled.start()` and `setJsDragActive(true)`; likewise for `onDrop`.
- Optional hooks default to "off", which is what the platforms that lacked the code had (`rejectDragAt` absent ⇒ `canDrag` is the old `!isEphemeral() && !isMagnified()`).
- The two conditional effects (geometry → `notifyPaneReflow`, clear-`.dragging`) are created only when the platform sets them, so linux/darwin get no new effects.
- Each platform's `ResizeHandle` is its original code verbatim; they genuinely differ (three pointer models) and stay separate.

## Verification

- `tsc --noEmit -p tsconfig.json` — clean (all three platform files are type-checked; the `.platform` stub is unchanged)
- `vitest run` — **253 files, 3,808 tests passed** (2 files / 3 tests skipped, pre-existing)
- **Not exercised here:** Linux and macOS builds — this machine is Windows. The equivalence argument above is structural; a Linux/macOS smoke test of pane drag and divider resize before the next release would be the belt to these braces.

Follow-up to #3031 (the audit); same family as #3033, #3034, #3036, #3039.

### DIFF
--- a/.changesets/1788751173-tilelayout-one-shared-core-createtilelayout-with-per-platfor-jnel.md
+++ b/.changesets/1788751173-tilelayout-one-shared-core-createtilelayout-with-per-platfor-jnel.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+TileLayout: one shared core (createTileLayout) with per-platform hook objects for win32/linux/darwin instead of three near-identical copies; no behavior change
--- a/docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md
+++ b/docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md
@@ -169,7 +169,7 @@ Two teams (or one team at two times) built the same *idea* twice with the same *
 
 Ordered by payoff ÷ risk. Each step is independently shippable.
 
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

PR #3043, opened 2026-09-07
Korp@claudius: fix(srv): failed data migration is fatal and surfaced — migration hardening Phase 1a

GenericAgentX-asaf:
## Summary

`SPEC_MIGRATION_SYSTEM_HARDENING_2026_08_03` **Phase 1a** (failure mode F4). Closes the gap `REPORT_LARGE_MIGRATIONS_COMPLETION_AUDIT_2026_09_06.md` §3.2 verified was still live: `bootstrap.rs` ran startup migrations behind a `warn!` and booted anyway against a store in an unknown state. Failing to *open* the store always exited 1; failing to *migrate* it did not.

## What changes

| Layer | Before | After |
|---|---|---|
| **srv** `bootstrap.rs` | `Err(e) => warn!(... continuing)` | error log → `AGENTMUXSRV-MIGRATION-FAILED error:<reason>` on stderr → `exit(1)` before ESTART |
| **common** (new `srv_stderr.rs`) | — | prefix + formatter (flattens multi-line errors) + parser (never empty). One definition instead of a literal per crate. |
| **launcher** `srv_spawner.rs` | srv exit before ESTART → 30 s silence → `EstartTimeout` | third channel; waiter breaks **immediately** on the line → `SrvSpawnError::MigrationFailed(reason)`; splash stage ends "migration failed" |
| **launcher** supervisors (win + unix) | `eprintln!` + `exit(1)` (invisible on Windows — `windows` subsystem) | additionally `show_fatal_dialog` with the reason and restart guidance |
| **CEF host** `sidecar.rs` (`task dev:standalone`) | same 30 s timeout | `"agentmux-srv database migration failed: <reason>"` immediately |
| `docs/exe-return-codes.md` | already claimed srv exits 1 on migration failure | now true; line protocol documented |

Restart-to-retry guidance in the dialog is accurate: srv re-runs every unapplied migration next launch, and `maybe_snapshot_pre_migration` already wrote a snapshot before the attempt.

## Tests
- **common:** 6 new — format strips newlines, round-trip, tolerant parse (missing `error:`), non-empty guarantee, other protocol lines → `None`, and an explicit check that `MIGRATION-FAILED` isn't a `starts_with` prefix of `MIGRATING` (both consumers branch on `starts_with`).
- **launcher:** 3 new — `Display`, dialog body content, end-to-end srv-format → launcher-parse over the shared module.
- **srv:** `migrations::` slice 64/64.
- `cargo check` clean: common, srv, launcher, cef.

## Not in this PR
Phase 1b (`agentmux-srv migrate --verify` / doctor pass, muxspect wiring) and Phases 2–6. The spec's Status line now says exactly that.

### DIFF
--- a/.changesets/1788751693-fix-srv-a-failed-data-migration-is-now-fatal-srv-emits-agent-ttg2.md
+++ b/.changesets/1788751693-fix-srv-a-failed-data-migration-is-now-fatal-srv-emits-agent-ttg2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): a failed data migration is now fatal — srv emits AGENTMUXSRV-MIGRATION-FAILED and exits 1 before ESTART instead of booting against a half-migrated store; launcher and CEF host surface the reason immediately (migration hardening Phase 1a)
--- a/agentmux-cef/src/sidecar.rs
+++ b/agentmux-cef/src/sidecar.rs
@@ -366,6 +366,10 @@ pub async fn spawn_backend(state: &Arc<AppState>) -> Result<BackendSpawnResult, 
     // running (same pattern as agentmux-launcher/src/srv_spawner.rs).
     let (tx, mut rx) = tokio::sync::mpsc::channel::<BackendSpawnResult>(1);
     let (migration_tx, mut migration_rx) = tokio::sync::mpsc::channel::<()>(4);
+    // Third channel: a fatal migration failure. srv reports it on stderr and
+    // exits 1 without ESTART (SPEC_MIGRATION_SYSTEM_HARDENING Phase 1); the
+    // waiter fails immediately with the reason instead of after the timeout.
+    let (failure_tx, mut failure_rx) = tokio::sync::mpsc::channel::<String>(1);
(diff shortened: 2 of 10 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed the unix.rs comment (reagent P2) in 522bf3a02: it was a windows.rs copy-paste and wrong for that file — on Unix `show_fatal_dialog` is itself an `eprintln`, so the comment now says the call is kept for cross-platform symmetry and a future native dialog, not because stderr is invisible. No code change.

GenericAgentX-asaf:
Addressed reagent P1 in 563fe35e7: the dialog no longer claims a snapshot/backup exists. It now states only the ordering fact the launcher knows — the runner writes a copy to `~/.agentmux/shared/backups/pre-migration-*/` before applying anything, so if the reason names that copy step, nothing was changed — and points at `muxlog srv`. Added `dialog_body_never_asserts_a_backup_exists` pinning the absence of the unconditional claim (4/4 launcher tests green).

a5af:
@codex review

Issue #1549, opened 2026-06-18
Architecture refactor tracking (A1–A15)


a5af:
Tracking issue for the codebase architecture-refactor program. Pick any 🟢 item, follow the workflow, check it off.

**Source of truth (full detail — entry points \`file:line\`, approach, acceptance criteria, gotchas):**
- Audit: \`docs/analysis/ANALYSIS_CODEBASE_ARCHITECTURE_AUDIT_2026_06_18.md\`
- Living handoff board: \`docs/analysis/TRACKING_ARCHITECTURE_REFACTOR_A1_A15_2026_06_18.md\`

## Board

- [x] **A1** — RPC contract enforcement test — ✅ #1544
- [x] **A2** — Extract srv↔mcp↔bashwrap HTTP DTOs into `agentmux-common` — ✅ #1575
- [x] **A3** — Break `global.ts` god-module + `global.ts ⇄ wos.ts` cycle — ✅ #1566
- [x] **A4** — Split `service.rs::dispatch_service` (2272-line match) — ✅ #1552
- [x] **A5** — Extract `BlockControllerCore` (3 near-clone controllers) — ✅ #1564
- [ ] **A6** — Collapse agent-pane 4 parallel state systems / kill the mirror — 🟡 mirror KILLED in #3054 (2026-09-07: AgentAtoms mirror removed, model.state/model.document are the single reactive source, acceptance criterion met + pinned by agent-pane-view.test.ts). Remaining half — scroll/expansion unification — is SPEC_AGENT_PANE_LAYOUT_REDUCER render-path wiring; track it there.
- [x] **A7** — Shared `ToolCorrelator` for translator tool-call/result — ✅ #1545
- [x] **A8** — Split `websocket.rs` by command family — ✅ #1554
- [x] **A9** — De-dup agent-pane "is busy?" selector (4×); route raw dispatches via `paneModel` — ✅ #3044 (busy predicate was already unified by the state-machine work; #3044 routed the 11 raw dispatches + guard test)
- [ ] **A10** — Consolidate data-dir resolution onto `DataPaths` — 🟡 smaller than estimated; current impl works; safe to skip or tackle incrementally
- [x] **A11** — Real `BlockRegistry` + registry-driven `ModalLayer` — ✅ #1562
- [x] **A12** — Dead-code sweep — ✅ #1542, #1565; submit/interrupt watchdogs confirmed live
- [x] **A13** — Spec/doc hygiene — ✅ #1558
[...]

PR #3044, opened 2026-09-07
Korp@claudius: refactor(agent-pane): route agent-view's 11 raw dispatches through AgentPaneModel (A9 of #1549)

GenericAgentX-asaf:
## Summary

**A9** of the architecture-refactor program (#1549). Filed 2026-06-18, blocked then on #1543 (long since merged), untouched since. Still exactly **11 raw store dispatches** in `agent-view.tsx` today — a file that has grown from 1,282 to 2,730 lines in the meantime (`REPORT_LARGE_MIGRATIONS_COMPLETION_AUDIT_2026_09_06.md` §3.1).

## Why it matters

`AgentPaneModel` (per its own contract in `agent-pane-model.ts`) is the one place that:
- **drops** a dispatch after `unregisterPane` instead of throwing — the hard `dispatch` throws on an unregistered pane, which is exactly the post-unmount race a view-level async callback hits
- emits the `agent:dispatchPane` **render-trail** point the crash boundary dumps

All 11 raw calls bypassed both.

## Change

Mechanical, one file:

| Before | After |
|---|---|
| `dispatchPane(model.blockId, cmd, src)` ×7 | `paneModel.dispatchPane(cmd, src)` |
| `dispatchPaneIfRegistered(model.blockId, …)` ×2 | `paneModel.dispatchPane(…)` |
| `dispatchDocIfRegistered(model.blockId, …)` ×2 | `paneModel.dispatchDoc(…)` |

`paneModel` is assigned synchronously in the component body ahead of every site; return values preserved. Three now-unused imports removed.

**The other half of A9** ("one `isPaneBusy()` selector; the `isLoading() || workingFromPhase()` combo duplicated 17×") was already resolved by the state-machine work — `isWorking`/`workingFromPhase` in `agent-pane-state/types.ts` is the single predicate ("since PR G") and agent-view has one remaining use. Nothing left to do there.

## Test

`agent-view-dispatch-via-pane-model.test.ts` pins the acceptance criterion grep-shaped (same pattern as `flows/run-cli-login-single-caller.test.ts`): no store-dispatch imports, no raw `dispatch*(` call, **and** a non-zero `paneModel.dispatch*(` count so the first two can't be satisfied by deleting call sites.

## Verification
- `tsc --noEmit`: **0 errors** (baseline on `main` is also 0 — the board's "~31 pre-existing" note is stale)
- vitest: **135 files / 1,822 tests** green across `frontend/app/view/agent` + agent-pane store suites
- ESLint: could not run — fails on config load (`TypeError: Unexpected array`) for *any* file on this checkout, including untouched ones. Pre-existing.

Board + #1549 checkbox will be updated with this PR number in a follow-up commit on this branch.

### DIFF
--- a/.changesets/1788752197-refactor-agent-pane-route-agent-view-s-11-raw-store-dispatch-xkvo.md
+++ b/.changesets/1788752197-refactor-agent-pane-route-agent-view-s-11-raw-store-dispatch-xkvo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(agent-pane): route agent-view's 11 raw store dispatches through its AgentPaneModel handle (A9 of #1549) — post-unmount dispatches now drop safely instead of throwing, and all pane commands hit the crash-trail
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

PR #3045, opened 2026-09-07
Manoz@Area54: refactor(storage): one ManagedResource implementation behind the skill_* and mcp_server_* store methods

GenericAgentX-asaf:
## Summary

Phase 4 (storage side) of `docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md`: `skills.rs` and `mcp_servers.rs` each carried the same **eighteen** store methods — agent- and bundle-level list / get / delete / bind / unbind / upsert-unique / is-accessible / is-bound plus the two catalog lists — identical except for table and column names and the noun in four error messages (§2.4 measured 58–71% duplication). One generic implementation now lives in `storage/managed.rs` behind a `ManagedResource` trait; `Skill` and `McpServer` implement it in ~45 lines each, and every public method keeps its name, signature, doc comment and response struct and delegates in one line. **No behavior change**; `skills.rs` 1,353 → 1,082 lines, `mcp_servers.rs` 939 → 677, `managed.rs` +~440 including its own tests.

## Shape

```rust
pub trait ManagedResource: Sized {
    const TABLE, REF_COL, AGENT_REF_TABLE, BUNDLE_REF_TABLE: &'static str;
    const COLUMNS, UPDATE_COLUMNS: &'static [&'static str];
    const NAME_NOUN, ARTICLE_NOUN, KIND: &'static str;   // the four error messages' nouns
    fn from_row(&Row) -> rusqlite::Result<Self>;
    fn values(&self) -> Vec<Value>;
    fn id(&self) -> &str;  fn name(&self) -> &str;
}
```

`Store` gains `managed_list / _list_global / _list_global_for_agent / _get / _delete / _bind_agent / _bind_bundle / _unbind / _upsert_unique / _upsert_unique_for_bundle / _upsert_unique_global / _is_accessible_to / _is_bound_to / _union_bundle_refs`, all `pub(super)`, parameterised on `Owner::{Agent, Bundle}` where the original had an agent-level and a bundle-level twin. `effective_skills` / `effective_mcp_servers` share the bundle-union step; the skills-only legacy `db_agent_skills` fallback stays exactly where it was (including the reagentx P1 ordering constraint from #2639 that `has_own_skill_refs` is decided *before* the union — the comment and the code order are untouched).

## Why it is safe

- **Every SQL statement is the hand-written one with names substituted.** The substitutions come from trait `const`s only, never runtime input, so the statements are exactly as fixed as the literals they replace; values still go through `?N` parameters. `is_global` is never in `UPDATE_COLUMNS`, and the two places that hard-coded it (`1` in the global upsert, `0` in the bundle-scoped upsert) are `force_global: Some(true|false)` — unit-tested in `managed.rs`.
- **Error strings are byte-for-byte the originals** (e.g. `cross-channel MCP server binding is not supported`, `skill name 'x' already bound to this bundle`), via the three noun consts. Nothing outside the two modules matches on them (grepped).
- **The rewrite was scripted, not hand-edited:** a table of 18 method names per module, bodies replaced between the signature's `{` and the closing `}`, each name asserted to occur exactly once; doc comments and the test modules were never touched.
- The `id_store`-not-`self` bundle-existence check (reagentx P0 on #2639) and the cross-channel agent check (reagentx P1 on #2315) are preserved in `managed_bind_bundle` / `managed_bind_agent`, with their rationale moved into those docs.

## Verification

- `cargo test -p agentmux-srv storage::` — **396 passed, 0 failed**; specifically `storage::managed` 3, `storage::skills` 17, `storage::mcp_servers` 14 (the 31 pre-existing primitive tests run unchanged against the new implementation — they cover list ordering, bind/unbind idempotency, cross-store bundle checks, name-conflict errors, delete purging both ref tables, and the effective-set union/dedup).
- `cargo test --workspace -- --test-threads=1` (the CI invocation) — **4,005 passed, 0 failed**
- One unused `rusqlite::params` import in `mcp_servers.rs` was removed; no other warnings in the three files.
- Also removes the `use std::time::SystemTime;` in `session_archive.rs` that #3033 left behind — its one use is fully qualified, so the import was dead and warned. A leftover of mine, not of this refactor.

## Not in this PR

The audit's step 13 also names the frontend half (a generic manager pane + agent modal for `McpManager` / `SkillManager`) and the "does memory bundles fit as a third?" test of the abstraction. Storage first: it is the half with an existing test suite to prove equivalence, and the trait is what a third primitive would implement.

Follow-up to #3031 (the audit); same family as #3033, #3034, #3036, #3039, #3041.

### DIFF
--- a/.changesets/1788751924-storage-one-generic-managedresource-implementation-behind-th-jmrp.md
+++ b/.changesets/1788751924-storage-one-generic-managedresource-implementation-behind-th-jmrp.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+storage: one generic ManagedResource implementation behind the skill_* and mcp_server_* store methods (agent- and bundle-level list/get/bind/unbind/upsert/access checks); public API, messages and tests unchanged
--- a/agentmux-srv/src/backend/session_archive.rs
+++ b/agentmux-srv/src/backend/session_archive.rs
@@ -17,7 +17,6 @@
 use std::io::{Read, Write};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use std::time::SystemTime;
 
 use flate2::Compression;
 use flate2::read::GzDecoder;
--- a/agentmux-srv/src/backend/storage/managed.rs
+++ b/agentmux-srv/src/backend/storage/managed.rs
@@ -1,0 +1,580 @@
+// Copyright 2025-2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+//! The "managed primitive" storage shape shared by standalone Skills and MCP
+//! Servers (v1 composable model, SPEC_V1_MCP_SKILLS_PRIMITIVES_2026_06_30.md;
+//! bundle-level refs per SPEC_BUNDLE_AS_CONTAINER_V2_2026_08_17.md).
+//!
+//! Both primitives are the same pattern: one catalog table whose rows are
+//! either global (visible to everyone) or private, an agent-level ref table
+//! binding rows to agents, and a bundle-level ref table binding rows to
+//! bundles. Before this module `skills.rs` and `mcp_servers.rs` each carried
+//! their own copy of the eighteen list / get / delete / bind / unbind /
+//! upsert / access-check methods — identical except for table and column
+//! names and the noun in error messages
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

PR #3046, opened 2026-09-07
Manoz@Area54: docs: Phase 2 codegen design spec and Phase 5 saga/reducer decision for the DRY audit

GenericAgentX-asaf:
## Summary

The two audit phases that need a decision rather than a refactor, written up so the decision can be made — plus the report's progress note brought current. Docs only.

| Doc | Status | What it settles |
|---|---|---|
| `docs/specs/SPEC_RPC_BINDINGS_CODEGEN_2026_09_07.md` | proposed | Phase 2 (steps 7–9). The obvious tool — `ts-rs` on the 190 `rpc_types` structs — produces types but not bindings: which command takes which request and returns which response is not data anywhere in srv (each of the 237 handlers picks its request type inside its closure and builds its response ad hoc). So the design is registry-first: a `register_typed<Req, Resp>` that records the mapping as it registers, a `--dump-rpc-schema` mode that emits `rpc.gen.d.ts` + generated stubs + shared constants, a `diff -r` CI gate, and a domain-by-domain migration order that starts with zero migrated handlers so the first PR is behaviour-free. |
| `docs/architecture/DECISION_SAGA_REDUCER_TWO_FRAMEWORKS_2026_09_07.md` | proposed | Phase 5 (step 14). The launcher's `Saga` is a trait-based state machine driven from a bus subscription (one consumer: `pool_respawn`); srv's sagas are async functions with in-process oneshot dispatch (every multi-step workspace command). The srv header and `docs/retro/saga-coordinator-location-analysis-2026-04-30.md` already record that this is deliberate. Three options are laid out; the recommendation is to keep both and cross-reference the split in the four module headers — not a shared crate (would undo a deliberate choice to remove a duplication that is one shared method name), not a rename (wire-format and spec churn for clarity a page provides). **Awaits the repo owner's call; no code moves until then.** |

The report's §5 progress note now records #3041 (TileLayout) and #3045 (Phase 4 storage half), and steps 7, 13 and 14 point at the documents above.

## Verification

- `scripts/check-doc-status.sh` and `scripts/check-spec-citations.sh` pass on all three files (every cited spec/retro path exists).
- Every count in the codegen spec was re-measured on `main` today (284 stubs by `grep -c 'Command('`, 237 `register_handler(` sites, 190 `pub struct Command*`, the `Promise<…>` census).

Follow-up to #3031 (the audit); same family as #3033, #3034, #3036, #3039, #3041, #3045.


a5af:
@codex review

GenericAgentX-asaf:
Fixed the P2 (the duplicated '#3041' sentence) and regenerated docs/specs/INDEX.md for the new spec — that was the failing 'Specs index is current' gate. Rebased onto current main. @reagentx-workflow review again

a5af:
@codex review

a5af:
@codex review

GenericAgentX-asaf:
P2 addressed: step 7 now says the 322 are the 284 command stubs plus the 38 service methods from §2.1, which is the split the spec uses. @reagentx-workflow review again

PR #3047, opened 2026-09-07
Korp@claudius: fix(launcher,cef): migration-failure reason can no longer lose the ESTART select! race

GenericAgentX-asaf:
## Summary

Follow-up to #3043, fixing the P1 ReAgent filed on it post-merge. Real bug.

**The race:** the stderr reader sends the migration-failure reason and then hits EOF within microseconds (srv `exit(1)`s right after the line), dropping every sender before the waiter's `tokio::select!` is next polled. Both "failure buffered" and "ESTART channel closed" are then ready. `select!` without `biased;` picks pseudo-randomly, so `rx.recv() → None` could win, degrading the outcome to `EstartChannelClosed` — which the supervisors' fatal-dialog branch doesn't match. On Windows that is precisely the silent failure #3043 was meant to end.

## Fix (identical in launcher and CEF host)

1. **`biased;`** with the failure arm first — when both are visible in one poll, the reason wins.
2. **Post-loop buffered check** — if the outcome is "channel closed", `try_recv` the failure channel and prefer a buffered reason. Covers the ordering `biased;` alone can't: closed channel observed on an earlier poll than the reason's arrival.

Launcher's guard is `prefer_buffered_failure()` on a now-module-level `EstartWait`, unit-tested (3 new tests: buffered+closed → `MigrationFailed`; closed alone stays `Ready(None)`; `Timeout`/`MigrationFailed` pass through without draining). CEF mirrors it inline on its `Result<_, String>`.

## Verification
- launcher: **7/7** `srv_spawner` tests green
- cef: `cargo check` clean

### DIFF
--- a/.changesets/1788753142-fix-launcher-cef-migration-failure-reason-can-no-longer-lose-bnj2.md
+++ b/.changesets/1788753142-fix-launcher-cef-migration-failure-reason-can-no-longer-lose-bnj2.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(launcher,cef): migration-failure reason can no longer lose a select! race to the closed ESTART channel — biased poll order plus a post-loop buffered-reason check, so a failed migration always surfaces as MigrationFailed instead of degrading to the generic channel-closed error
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #3048, opened 2026-09-07
Camper@claudius: fix(lan): make the mDNS instance label unique and dot-free so same-version hosts discover each other

GenericAgentX-asaf:
## Summary

Two hosts running the same AgentMux build could not discover each other over LAN — **in one direction only**. Reproduced live between `claudius` (192.168.1.230) and `gamerlove` (192.168.1.68), both on 0.55.37. Diagnosed jointly with **Manpo@gamerlove**, who independently found the same root cause and contributed the dot-escaping and stale-`service_fullname` halves.

The mDNS instance label was `format!("agentmux-{instance_id}")`, and `instance_id` is the **version** (`config.rs`'s `args.instance`). Two defects:

1. **Not unique** — every host on the same release registered an identical fullname, so `handle_event`'s self-skip (`if fullname == self.service_fullname`) discarded genuine peers as if they were us.
2. **Not a single DNS-SD label** — `v0.55.37` contains dots, nothing escaped them, so mdns-sd split at the first dot and treated `agentmux-v0` as the label. Caught verbatim in the peer's log as `agentmux-v0 (2).55.37._agentmux._tcp.local.` — note the conflict suffix landing *mid-version-string*.

They compound into the asymmetry: the collision triggers conflict-renaming, and because `service_fullname` is captured **before** `register()`, the renamed host keeps a stale self-name that exactly matches the *other* host's real name. That host then drops every announcement from its peer, permanently, with no self-heal — reporting `lan: []` forever while the peer sees it fine. The `resolves_to_this_instance` guard from #3025 cannot help: it bails early when the port differs, and a genuine peer never shares our port, so it never runs.

This also answers the question the source comment at the old `:528` admitted it could not: *"Why the fullname check misses them is not established — plausibly mdns-sd conflict-renaming."* It is conflict-renaming, and the trigger is the unescaped dots plus the non-unique name.

## Changes

- `mdns_instance_label(hostname, port)` — unique per instance, dot-free, and deliberately **version-free**. A version does not belong in an identity key, and it was the version being in there that created the cross-host collision. The version still travels in the TXT record, which is where peers actually read it.
- A fullname match is now a **hint, not proof**: it may only skip when the port+address test agrees this really is us. The blank-TXT self-resolution storm the check was originally added for is still caught — those are on our own port at our own address, so the port+address test returns true on its own.
- Log the registered fullname at INFO. Previously the only site logging a fullname was the `ServiceRemoved` branch, so an affected host was **mute** — this had to be diagnosed from the peer's log, because the broken host could not self-report its own service name.

**Backward compatible.** An un-upgraded peer sees a label different from its own, so it stops self-skipping us, and we stop self-skipping it. Mixed-version pairs recover without upgrading both sides simultaneously — which matters, because that is the state any rollout is in.

## Test plan

- [x] `cargo test -p agentmux-srv` — **3064 passed, 0 failed**
- [x] New: a peer whose fullname collides with ours is still discovered (the regression itself)
- [x] New: the label is dot-free, unique across hosts and across instances on one host, and carries no version
- [x] New: an all-punctuation hostname degrades safely rather than producing a malformed label
- [x] Existing self-skip fixtures updated — they set `port: 0` while minting self-events on a *different* port, which no longer represents a real self-resolution (they passed for the wrong reason)
- [x] Changeset added

## Not covered here

The acceptance criterion I actually care about is a LAN jekt arriving as `TRUST=lan-verified` rather than `network-claimed`. Both jekts we exchanged during this investigation landed `network-claimed`, which I believe is downstream of this bug (with no peer, srv cannot fetch the sender's pubkey to verify against) — but that is **inferred, not proven**. Manpo is taking that verification plus the `find_agent_lan_pubkey` path. If it still reports `network-claimed` after this lands, there is a second independent bug and discovery was only half of it.

### DIFF
--- a/.changesets/1788753575-fix-lan-make-the-mdns-instance-label-unique-and-dot-free-so--4y72.md
+++ b/.changesets/1788753575-fix-lan-make-the-mdns-instance-label-unique-and-dot-free-so--4y72.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(lan): make the mDNS instance label unique and dot-free so two hosts on the same version discover each other
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -588,52 +588,53 @@ impl LanDiscovery {
                 // blank-TXT self-resolution was never recognized as self and
                 // was inserted as a phantom peer that could never self-heal
                 // (see the instance_id-preservation fix below for why).
-                // A fullname match is now a HINT, never proof. It may only
-                // skip when the port+address test agrees that this really is
-                // us. Treating equality alone as proof is what broke
-                // cross-host discovery before 2026-09-06: with a
-                // version-derived (non-unique) name, a genuine peer's fullname
-                // could equal ours exactly — and on a conflict-renamed host,
-                // `service_fullname` is stale and matches the peer that WON
-                // the name. Either way the instance dropped every
-                // announcement from its peer, permanently, and reported
-                // `lan: []` forever. `mdns_instance_label` makes the collision
-                // itself impossible; this makes the consequence impossible
-                // even if some future name scheme regresses.
-                //
-                // The blank-TXT self-resolution storm this check was
-                // originally added for (~96% of own-address events resolving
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #3049, opened 2026-09-07
feat(settings): toggle to disable the startup splash screen

AgentO-asaf:
Adds the Settings toggle the user asked for. **Default is unchanged: the splash still shows.**

## The setting already half-existed

`agentmux-launcher` has read `"splash:disabled"` from `settings.json` since `SPEC_SPLASH_USERINFO_AND_DISABLE_2026_06_21.md` §6 (`splash_config.rs`). What was missing is everything that makes it *findable* — the key appeared in **no** schema, template, type or UI:

| Place | Before |
|---|---|
| `schema/settings.json` | absent — and `SettingsType` is `additionalProperties: false`, so the key was undeclared against its own schema |
| `settings-template.jsonc` | absent |
| `frontend/types/gotypes.d.ts` | absent |
| Settings pane | absent |

So the only way to turn the splash off was to already know the key existed and hand-edit the file. This declares it in all four.

## The one subtle bit: inverted polarity

The row reads positively — **"Startup splash screen"**, on = splash shows — while the stored key stays negative (`splash:disabled`, default `false`).

That is deliberate, not an oversight. The launcher reads the raw file *before* the host, the backend, or any of this code exists, and its read is fail-safe by design: a missing key, an unreadable file or a parse error must all resolve to **show the splash**. `disabled` is the form that gets that for free; a positive `splash:enabled` would make "file unreadable" mean "hide the splash".

Inverting in one place is cheap; getting it backwards would silently hide the splash for everyone who ever opened Settings. So `appearance-section.test.tsx` (new) pins the mapping in both directions — absent → ON, `disabled:true` → OFF, turning it off writes `true`, turning it back on writes `false`. **I mutation-checked it**: dropping the inversion fails 2 of the 4.

## Applies at the next launch

The launcher resolves this before startup, so toggling it does not affect the running app. The row's description says so. `AGENTMUX_SPLASH=0` / `AGENTMUX_NO_SPLASH=1` still override it for a single run.

## Verification

- `npx vitest run frontend/app/view/settings` — 17 passed (3 files), including the 4 new.
- `npx tsc --noEmit` — clean.
- `schema/settings.json` re-parsed as valid JSON.
- **Not verified:** the toggle was not clicked in a running app, and no splash was observed appearing/disappearing across a relaunch — this shell has no screen access on this Mac. The logic is unit-tested and the launcher side is pre-existing, unchanged code.

## Related

Came out of the §7.5.1 work in #3037: background-service mode with the splash disabled is exactly the configuration this toggle makes reachable without hand-editing.

### DIFF
--- a/.changesets/1788753709-feat-settings-toggle-to-disable-the-startup-splash-screen-xoi3.md
+++ b/.changesets/1788753709-feat-settings-toggle-to-disable-the-startup-splash-screen-xoi3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(settings): toggle to disable the startup splash screen
--- a/frontend/app/view/settings/sections/appearance-section.test.tsx
+++ b/frontend/app/view/settings/sections/appearance-section.test.tsx
@@ -1,0 +1,77 @@
+// Copyright 2026, AgentMux Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Tests for the startup-splash toggle in the Appearance section.
+ *
+ * This row is the one control in Settings whose stored value is INVERTED
+ * relative to its label: the key is `splash:disabled` (false by default)
+ * because `agentmux-launcher` reads settings.json directly, before the
+ * frontend exists, and "disabled" is the form that fail-safe read wants —
+ * an unreadable or absent value must mean "show the splash"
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #3052, opened 2026-09-07
fix(agent-pane,statusbar): reclaim the fresh-pane picker gap, restore a LAN signal

AgentY-asaf:
Two small UI fixes. Both are downstream of a recent change that silently removed something the UI had been relying on — neither was caught by review, because in both cases the diff that broke them contained no frontend files.

## 1. Dead gap above "My Agents"

`.agent-picker` reserved the pane tab strip's height unconditionally (`space-4 + 28px`). #2984 stopped rendering that strip **at all** on a fresh pane, so the reservation became ~28px of dead space above the header — reported as "an unnatural looking margin at the top."

The clearance is now a `--pane-tab-strip-reserve` custom property set on `.agent-picker-host` from the **same** `shouldShowTabStrip` predicate that decides whether the strip mounts, so the space reserved and the thing it's reserved for can't disagree. `0px` on a fresh pane; unchanged (`28px`) whenever the strip actually renders.

Note the old comment claimed parity with `.agent-document`'s rule while actually using a different base value — that's corrected too.

## 2. LAN diamond had no "on, but no peers" state

The diamond was gated purely on peer count (`lanCount() > 0`), which matches `docs/specs/hostname-popover.md`. But until #3025, every instance discovered **itself** as a phantom peer, so the count was never 0 while discovery was on — the glyph effectively behaved as an "enabled" lamp, and that's how it was learned. #3025 correctly removed the phantom; the consequence is that a lone instance with discovery genuinely enabled now renders **identically** to one with it switched off.

Three states now:

| State | Glyph | Color |
|---|---|---|
| Peers found | `◆` filled | accent |
| On, no peers | `◇` hollow | muted |
| Off (default) | `◇` hollow | warning |

Filled-vs-hollow carries the peers/no-peers distinction on its own, so state isn't conveyed by color alone. Glyph, color-class and tooltip all come from one pure `resolveLanIndicator()` so the label can't drift from what's drawn.

## Verification

- `npx tsc --noEmit` clean.
- New `lan-indicator.test.ts` (6 cases) pins the state table, including the case that had no representation before (`enabled + 0 peers` must differ from `off`) and a stale-peer-count-after-toggle-off case. `tab-strip-visibility.test.ts` still passes — 10 total.
- The LAN behavior was root-caused against the **live** instance's srv log, not inferred: discovery started cleanly (`LAN discovery started (mDNS)`, port 59859), no error, and **zero** `LAN peer discovered` events — while an older pre-#3025 instance on the same machine still logged the phantom signature verbatim (`peer_id:""`, its own port, its own link-local addresses). Written up in `docs/retro/retro-lan-diamond-vanished-after-self-peer-fix-2026-09-06.md`.
- **Not yet visually confirmed in a running build** — a `task dev` instance is being brought up on this branch for that. I'll confirm here once the picker gap and all three diamond states have been eyeballed.

### DIFF
--- a/.changesets/1788764361-fix-agent-pane-drop-the-dead-tab-strip-gap-above-my-agents-o-5r6w.md
+++ b/.changesets/1788764361-fix-agent-pane-drop-the-dead-tab-strip-gap-above-my-agents-o-5r6w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): drop the dead tab-strip gap above My Agents on a fresh pane
--- a/.changesets/1788764363-fix-statusbar-lan-diamond-now-shows-all-three-states-peers-o-ajtk.md
+++ b/.changesets/1788764363-fix-statusbar-lan-diamond-now-shows-all-three-states-peers-o-ajtk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(statusbar): LAN diamond now shows all three states (peers / on-but-idle / off)
--- a/docs/retro/retro-lan-diamond-vanished-after-self-peer-fix-2026-09-06.md
+++ b/docs/retro/retro-lan-diamond-vanished-after-self-peer-fix-2026-09-06.md
@@ -1,0 +1,144 @@
+# Retro: the status-bar LAN diamond vanished — because #3025 fixed the bug that was lighting it
+
+**Date:** 2026-09-06
+**Status:** Root-caused, no code change made yet. The current behavior matches
+the written spec; what changed is the behavior users had actually learned. A
+follow-up UI decision is proposed below but deliberately not implemented in
+this retro.
+**Severity:** Low functionally (LAN discovery is working correctly), Medium for
+trust — the only status-bar signal that discovery is alive silently disappeared
+for every single-instance user, with nothing to distinguish it from "off".
+**Observed by:** repo owner, on a running v0.55.37 portable
+**Related:** PR #3025 (`b904ae00d`, "fix(lan): stop this instance discovering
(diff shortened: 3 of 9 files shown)


a5af:
@codex review

AgentY-asaf:
**Codex P2 — "Handle discovery errors before showing LAN as idle": fixed in `846790936`.** Good catch, and it's the same bug class this PR exists to fix — a UI signal quietly meaning something other than it appears to. Rendering a failed daemon as the muted state whose tooltip reads "no peers found" would have re-created the exact problem in a new place.

`lanDiscoveryError` is now an input to `resolveLanIndicator`, with a fourth `error` state (`◇` in `--error-color`, tooltip `LAN discovery failed: ‹reason›`).

Ranking is **off → peers → error → idle**, which mirrors the popover rather than inventing a new precedent:
- **off wins outright** — the user turned it off, so a stale error from a previous attempt isn't worth nagging about.
- **peers outrank the error** — the popover's peer rows are *not* gated on the error, while its "no peers" row *is* (`!props.lanDiscoveryError()`). Reaching a peer is direct proof discovery works, which makes a lingering error stale rather than current.
- **error outranks idle** — the reason there are no peers is that discovery never started.

Four new tests cover it: failure surfaces as `error` and explicitly `not idle`; disabled-with-stale-error still reports `off`; peers win over a lingering error; `null`/absent error behaves as no error. 9 tests in `lan-indicator.test.ts`, all passing, `tsc` clean.

Also in that commit: the doc-status gate wanted the retro's `Status:` to *start* with a recognized keyword, and the retro still described the three-state indicator as "not implemented here" — it ships in this PR, so it now documents what actually landed (including this error state).

PR #3053, opened 2026-09-07
Manoz@Area54: docs: regenerate the specs index (unblocks every PR — 67% of recent CI failures)

GenericAgentX-asaf:
## Summary

One line. `docs/specs/INDEX.md` said `### proposed (95)` while that section actually contained **96 rows** — the file was internally inconsistent, so `gen-docs-index.sh --check` failed for **every** PR regardless of what it changed, including PRs touching no docs at all, and including runs on `main` itself.

```
-### proposed (95)
+### proposed (96)
```

## Why this happened

The generator's own header comment (`scripts/gen-docs-index.sh:42-47`) predicted it exactly:

> *"CI tests the PR MERGE commit, not your branch head. … git will happily merge a new spec's ROW in from main while keeping YOUR section header count, so the merged file is internally inconsistent in a way neither parent was."*

Two PRs each add a spec. Git text-merges both **rows** cleanly, but the header **count** line is a one-line conflict resolved in favour of one side. The result has N+1 rows and a count of N, and neither parent was wrong.

## Why it matters more than it looks

Measured across the last 90 PR CI runs: **24 failed, and 16 of those (67%) were this gate** — 5 of them on `main`. It is the single largest source of CI failure in the repo right now, and it fails PRs that cannot possibly have caused it (#3045 is a pure Rust storage refactor touching zero docs).

This PR is the immediate unblock only. The structural fix — regenerating the index automatically after merge so parallel spec PRs stop producing this artifact — follows separately.

### DIFF
--- a/.changesets/1788764607-docs-regenerate-the-specs-index-the-committed-header-count-9-9hmj.md
+++ b/.changesets/1788764607-docs-regenerate-the-specs-index-the-committed-header-count-9-9hmj.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: regenerate the specs index — the committed header count (95) disagreed with its own 96 rows, a merge artifact that failed the 'Specs index is current' gate on every PR including main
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #3054, opened 2026-09-07
Korp@claudius: refactor(agent-pane): drop the AgentAtoms mirror — render from the reducer stores (A6 of #1549)

GenericAgentX-asaf:
## Summary

**A6** of the architecture-refactor program (#1549), open since 2026-06-18. Spec rides with the code: `docs/specs/SPEC_A6_AGENT_PANE_MIRROR_REMOVAL_2026_09_06.md`.

**Before:** `agent-view.tsx` created **19 Solid signals**, handed **17 setters** to the pane-state store through a hand-maintained `AgentPaneProjections` interface (+ a `documentSetter` for the document store), and the stores wrote changed fields back through them. Adding a reducer field touched four files plus a 17-line wiring block; forgetting one was silent (11 setters were `?`-optional "for back-compat"); and the view could write a mirrored signal directly, bypassing the reducer that owned the field — which it **did**, for `detailsOpen`, at `agent-view.tsx:1024` and `:1764`.

**After:** the stores own their reactive read side.

| Piece | Change |
|---|---|
| `agent-pane-state-store.ts` | one Solid signal per `AgentPaneState` field, **generated** from `Object.keys(initialState())`; frozen `AgentPaneView` of getters; post-dispatch writes only identity-changed fields inside one `batch()`. `AgentPaneProjections` deleted. DEV guard if the reducer produces an undeclared key. |
| `agent-document-store.ts` | owns the `nodes` signal; exports `documentNodes(blockId)` |
| `AgentPaneModel` | `+ state: AgentPaneView`, `+ document: Accessor<DocumentNode[]>`; `PaneRegistration` is `{ agentId }` |
| 10 consumers | `SignalPair<X>` → `Accessor<X>` (`documentNodes`, `turnPhase`, `compacting`, `pendingMessages`). None ever used the setter half. |
| `agent-view.tsx` | reads `paneModel.state.<field>` / `paneModel.document()`; the two `detailsOpen` bypasses → `paneModel.dispatchPane({ type: "DetailsExpand" \| "DetailsCollapse" })`; stale A9 comment from #3044's post-merge P2 fixed |
| `state.ts` | `AgentAtoms` keeps exactly one field — the view-local `documentStateAtom` no reducer owns |

**Why not `createStore` + `reconcile`** (the `launch-flow-store` pattern, considered first): `reconcile` diffs *into* the store's existing tree in place, which would alias the reducer's own state objects and mutate the `prev` that `dispatch()`'s `[wave-turn]` diagnostics compare against (and any held `snapshot()`). Per-field signals never mutate anything the reducer produced.

**Cascade contract preserved** (`LIFECYCLE_DISPATCH_LEAK_2026_05_15`): `CASCADE_DETECTED` still fires naming the changed field(s); hard `dispatch` still throws after; `dispatchIfRegistered` still returns `[]`. Store tests express the disposing subscriber as a `createComputed` over `paneView(id).streaming`.

## Out of scope, deliberately
The board's "unify the dual scroll/expansion" half of A6 is `SPEC_AGENT_PANE_LAYOUT_REDUCER_2026_06_02`'s render-path wiring and is tracked there. A6's **acceptance criterion** — "adding a pane state field touches one place; no `AgentAtoms ⇄ AgentPaneState` copy" — is met here. Board row updated to 🟡 with that note.

## Tests
- **`agent-pane-view.test.ts` (new)** pins the property *generically* — no field is named: the view exposes exactly `initialState()`'s keys; every field is identity-equal to `snapshot()` after a run of dispatches; a dispatch notifies readers of exactly the fields it changed and a no-op notifies none; the view outlives its slot. Plus grep-shaped guards so the projections interface and any `agentAtoms().<reducerField>Atom[0|1]` in `agent-view.tsx` cannot creep back.
- 16 existing test files converted (setter-call assertions → `model.state` / `model.document()` reads).

## Verification
- `tsc --noEmit`: **0 errors** (baseline 0)
- Full frontend suite: **255 files / 3,816 tests green**
- ESLint: cannot run on this checkout (pre-existing config-load failure, see #3044)


a5af:
@codex review

GenericAgentX-asaf:
Addressed reagent P2 in ebcda3763: the cascade-detection comment now states that with batched writes only the set of changed fields is knowable (one of their readers disposed the pane), not the single triggering field — which is what the warning already reports. Comment-only change.

GenericAgentX-asaf:

PR #3055, opened 2026-09-07
fix(memory): backfill native-memory versions from disk, repair rows mislabeled "detected outside AgentMux"

AgentO-asaf:
Pre-existing memory files show up in Armory's Native Memory history tagged **"Detected outside AgentMux — provenance unknown."** They are ordinary files. Nothing wrote them out of band.

## What actually happened

`SPEC_MEMORY_VERSION_CONTROL_AND_ARMORY_AUDIT_2026_08_19.md` §9 asks for a rollout backfill of "one `source: 'agent_inferred'` version per existing `(agent_id, filename)` **row**". `m0023` implements that literally, iterating `db_agent_native_memory` — the **mirror** table.

That mirror is populated lazily, only once something reads or exports an agent's memory. **On a typical install it is empty.** So `m0023` iterates nothing and backfills nothing, in 0 ms.

The drift sweep (§4.5) doesn't use the mirror — it reads the real filesystem, which is authoritative here per `SPEC_NATIVE_MEMORY_DURABLE_SYNC_2026_08_07.md`. Moments after migrations finish it finds those same files, sees no recorded version, and records each as `external_fs_write`.

Measured on a real affected install:

| | |
|---|---|
| `m0023` applied | `…285933`, **0 ms, 0 rows** (mirror empty, still 0 rows today) |
| sweep wrote | `…285994` — **61 ms later** |
| result | **21 rows**, 6 agents, 16 files, all `reconciliation_sweep`, all `parent_version_id IS NULL`, all inside one second |
| since then | **0** further versions recorded in 2.5 weeks |

**This is not a race.** `bootstrap` runs migrations to completion before the drift tasks spawn — that's why the timestamps are 61 ms apart rather than interleaved. Ordering was never the problem; reading a lazily-populated mirror instead of the authoritative filesystem was. Hence a migration, not a change to startup sequencing.

## The fix (`m0024`)

1. **Backfill from the filesystem**, via the same `list_all_memory_targets` enumeration the detector uses — so the two agree by construction rather than by coincidence. Same idempotency guard `m0023` documents: only a pair with no version yet is touched, so nothing lands out of chronological order.
2. **Repair the rows already written.** Runs *first* — relabeling gives the pair the `agent_inferred` version it should have had, which then makes step 1 correctly skip it.

### The repair predicate is deliberately narrow

Relabeling a provenance claim is not something to do speculatively. All four must hold:

- `source = 'external_fs_write'` **and** `detected_via = 'reconciliation_sweep'` — the slow path is the only one that can observe a never-before-seen file at startup;
- `parent_version_id IS NULL` — genuine first sight, not an observed change to something already tracked;
- `created_at` inside a **5-minute window anchored to `m0023`'s own `applied_at`**.

That last clause is load-bearing: **a memory file created out-of-band next week is also first-sight-via-sweep, and that one is a real external write whose label must stand.** Only the burst that raced this specific migration is in scope. No `m0023` row (fresh install) → the subquery is NULL → clean no-op.

`source_detail` keeps the original `detected_via` and records that the row was relabeled, so the audit trail says what happened instead of presenting the new claim as if it were the original observation:

```json
{"relabeled_from":"external_fs_write","detected_via":"reconciliation_sweep","reason":"rollout_first_sight"}
```

## Verification

- **Against a copy of a real affected store**: all 21 rows relabel to `agent_inferred`, nothing else matches. Original untouched — the run was on a copy.
- **6 new store tests** pin the predicate, including the two negatives that matter: a later genuine external write, and an observed change to an already-tracked file. Plus fs_watch-detected rows, the no-m0023 case, and idempotency.
- **3 new migration tests**: repair end to end, idempotent on re-run, no-op with no shared store.
- `cargo test -p agentmux-srv`: **3051 passed**. Three failures are pre-existing on clean `main` and unrelated — two `lan_listeners` socket-binding tests whose own names say `is_platform_dependent` (macOS-only; CI's legs are ubuntu/windows), and one that passes single-threaded as CI runs it. I verified all three fail identically on `origin/main` with my changes stashed.

## Not covered

Phase 1's filesystem enumeration isn't unit-tested end to end — `list_all_memory_targets` needs agent definitions, working directories and the shared registry, which is heavy env-dependent scaffolding. It's the same call the drift sweep already relies on; the phase-1 insert path is otherwise identical to `m0023`'s, which has its own tests.

### DIFF
--- a/.changesets/1788764784-fix-memory-backfill-native-memory-versions-from-disk-and-rep-btae.md
+++ b/.changesets/1788764784-fix-memory-backfill-native-memory-versions-from-disk-and-rep-btae.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(memory): backfill native-memory versions from disk and repair rows mislabeled 'detected outside AgentMux'
--- a/agentmux-srv/src/backend/native_memory_drift.rs
+++ b/agentmux-srv/src/backend/native_memory_drift.rs
@@ -95,7 +95,7 @@ pub(crate) fn check_and_record_drift(
 /// InvalidData`, distinguishable from a transient I/O error) rather than
 /// read partially; callers log and skip it, same treatment as any other
 /// per-file read failure.
-fn read_memory_file_lossy(path: &Path) -> std::io::Result<String> {
+pub(crate) fn read_memory_file_lossy(path: &Path) -> std::io::Result<String> {
     use std::io::Read;
     let mut buf = Vec::new();
     let read = std::fs::File::open(path)?
--- a/agentmux-srv/src/backend/storage/agent_native_memory_versions.rs
+++ b/agentmux-srv/src/backend/storage/agent_native_memory_versions.rs
@@ -354,6 +354,69 @@ impl Store {
     /// happens to be old, and a hyperactive file never grows unbounded
     /// purely because every version happens to be recent. Returns the
     /// number of rows deleted.
+    /// Repair the mislabeled rollout backfill — see
+    /// `migrations::m0024_native_memory_backfill_from_fs` for the full
+    /// story. In short: `m0023` was supposed to tag every pre-existing
+    /// memory file `agent_inferred`, but it iterated the
+    /// `db_agent_native_memory` MIRROR, which is populated lazily and is
+    /// empty on a typical install — so it backfilled nothing, and the
+    /// drift sweep (which reads the real filesystem) claimed those same
+    /// files milliseconds later as `external_fs_write`, i.e. "detected
+    /// outside AgentMux — provenance unknown". Innocent files, alarming
+    /// label.
+    ///
+    /// The predicate is deliberately narrow, because relabeling a
+    /// provenance claim is not something to do speculatively. All four
+    /// must hold:
+    ///
+    /// - `source = 'external_fs_write'` and `detected_via =
+    ///   'reconciliation_sweep'` — the slow path, which is the only one
+    ///   that can observe a file it has never seen before at startup;
+    /// - `parent_version_id IS NULL` — a FIRST-EVER version for that
+    ///   `(agent_id, filename)`, i.e. genuine first sight rather than an
+    ///   observed change to something already tracked;
+    /// - `created_at` inside a 5-minute window starting at `m0023`'s own
+    ///   `applied_at`. This is the load-bearing one: a memory file created
+    ///   out-of-band NEXT WEEK is also first-sight-via-sweep, and that one
+    ///   is a real external write whose label must stand. Only the burst
+    ///   that raced this specific migration is in scope.
+    ///
+    /// `source_detail` keeps the original `detected_via` and records that
+    /// the row was relabeled, so the audit trail says what happened rather
+    /// than quietly presenting the new claim as if it were the original
+    /// observation.
+    ///
+    /// Returns the number of rows relabeled. A no-op (returns 0) when
+    /// `m0023` never ran — nothing to repair on a fresh install.
+    pub fn agent_native_memory_version_relabel_rollout_first_sight(
+        &self,
+    ) -> Result<usize, StoreError> {
+        const ROLLOUT_WINDOW_MS: i64 = 5 * 60 * 1000;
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

AgentO-asaf:
**P1 confirmed and fixed in `dcf993a` — the review was exactly right, and I verified it before changing anything.**

Empirically, not just by reading:

- `~/.agentmux/shared/store.db` on a real install has **no `db_agents` table**.
- `CREATE TABLE ... db_agents` is at `migrations.rs:535`, inside `run_object_schema` — not in `run_shared_store_schema`.
- `agent_def_list()` prepares that statement on the local connection **before** reaching its global cross-channel overlay, so on the shared store it returns `Err` immediately and the overlay never runs.
- `list_all_memory_targets` swallows it via `if let Ok(agents)`.

So the backfill half would have silently skipped every persisted-but-idle agent and covered only currently-active registry entries — "read the wrong source, get an empty list, look successful", which is `m0023`'s failure reproduced inside its own fix. Thanks for catching it; that's the second time this class of bug slipped through in this area.

## The fix

The migration is now **`Channel`-scoped** and enumerates with that channel's own store, matching what the drift sweep actually gets in production (`state.wstore` is the channel store). Versions are still read and written in the shared store, since that's where `db_agent_native_memory_versions` lives.

Channel scope also fixes a second gap the P1 implies: a single `Global` pass could only ever see one channel's `db_agents`. Per-channel, each channel backfills its own agents as its srv starts. The repair half is idempotent, so running once per channel is safe.

## The missing test, now present

`backfills_an_on_disk_file_for_an_agent_only_the_channel_store_knows` — an agent whose memory dir resolves through `working_directory` + `CLAUDE_CONFIG_DIR`, i.e. two per-channel reads, with a real file on disk.

**Mutation-checked**: reverted the one-line enumeration change and it fails; restored and it passes. It would have caught this.

Also made `test_agent_def` `pub(crate)` instead of duplicating its 30-field literal, which its own comment asks callers not to do.

`cargo test -p agentmux-srv -- --test-threads=1`: **3053 passed**, 2 failed — both `lan_listeners` socket tests whose names say `is_platform_dependent`, failing identically on clean `main` with my changes stashed (macOS-only; CI runs ubuntu/windows).

AgentO-asaf:
**P2 addressed in `679a5f9` — the review is right, and the honest fix is partly a documentation fix.**

The comment promised "a real external write's label must stand", full stop. It cannot deliver that *inside* the window: a file created out-of-band there is also first-sight-via-sweep with `parent_version_id IS NULL`, so it gets relabeled too. The code contradicted its own stated guarantee, and that overclaim was mine.

There is no clean discriminator to recover. `mtime` looks like one and isn't: a burst row's file may have been legitimately edited since, so an mtime cutoff would wrongly *exclude* genuine burst rows while still not identifying new ones.

So, two changes — neither pretending the ambiguity is gone:

**1. Window 5 min → 60s.** The sweep's `tokio::time::interval` fires immediately on its first tick, which is why the burst landed **61 ms** after migrations on the diagnosed install. A minute is ~1000× that slack for a loaded machine's first sweep while keeping the ambiguous region short. Re-verified against a copy of the affected store: all 21 rows still match at 60s.

**2. The comment now states the limitation plainly**, and justifies the trade by the *shape* of the two errors rather than by likelihood:

- Acting wrongly costs **one just-created file** a benign "Agent" label instead of "detected outside AgentMux", on an install the user is actively upgrading.
- Not acting costs **every pre-existing memory file on every affected install** a permanent security-flavored warning it did not earn — the bug this PR exists to repair.

**3. A test that pins the accepted false positive** (`relabels_a_genuinely_new_file_inside_the_window_too`) rather than a property I wish were true — so the guarantee can't be read as absolute, and narrowing it later is a deliberate change with a failing test rather than a silent one. Renamed the sibling to `leaves_a_genuine_external_write_outside_the_window_alone`, which is what it actually proves.

`cargo test -p agentmux-srv -- --test-threads=1`: **3057 passed**, 2 failed — the same two `lan_listeners` `is_platform_dependent` socket tests that fail on clean `main` (macOS-only; CI runs ubuntu/windows).

PR #3056, opened 2026-09-07
Manoz@Area54: ci: make the specs index merge-safe (67% of CI failures) + fix the node cold-start flake at its cause

GenericAgentX-asaf:
## Summary

CI was failing ~27% of the time (24 of 90 PR runs). Measured, not guessed:

| Cause | Runs | Real defect? |
|---|---|---|
| `Specs index is current` | **16 (67%)** | **No** — an artifact nobody authored |
| `Doc Status` / `Spec citations` | 6 | Yes |
| `all_tool_defs_are_valid_json_with_names` | 2 | **Yes** — real, reproduced every run |
| `create_no_window_flag_set` | 1 | Yes — genuinely flaky |

**Flaky tests are 1 of 24 failures (4%).** This fixes the 67% and the 4% at their causes, and deliberately adds **no blanket retries** — the two `all_tool_defs` failures were a real bug a retry wrapper would have hidden.

## 1. The specs index (67%)

`INDEX.md` is generated output that is also committed and gated. The section headers carried a row count (`### proposed (95)`). When two spec-adding PRs merged, git merged both **rows** cleanly but resolved the one-line **count** to a single side — leaving N+1 rows under a count of N, **a state neither parent was in and neither author could prevent**. It then failed for every open PR, including ones touching no docs, and for `main` itself five times.

**The fix is to stop committing the conflicting scalar.** With counts gone from the three section headers, rows from different branches three-way-merge to the correct union, and the only thing left to conflict is two rows landing adjacently — an honest conflict a human resolves by keeping both. The generator's header records this so the counts don't come back.

The gate **stays blocking**: a failure now means genuine drift (a spec added without regenerating), which is the author's to fix in one command.

### What changed since the first push of this PR

The original version added a workflow to regenerate post-merge and open a PR. **Both reviewers were right that it could not work, and I dropped it.**

- **Codex (P1), confirmed against the API:** `default_workflow_permissions` is `read` at **both** repo and org level, and `docs/retro/RETRO_GHCR_PUSH_403_2026_06_18.md` records that the org has *locked* it so repo admins cannot override. A workflow-level `permissions:` block can only restrict, never elevate — so `git push` and `gh pr create` would both have 403'd. The workflow would have been dead on arrival.
- **ReAgent (P1):** PRs opened with the default `GITHUB_TOKEN` do not trigger `pull_request` workflows, so that auto-PR would also have been the one PR in the repo with no checks.

Removing the count solves the same problem with no token, no bot, and no advisory downgrade — strictly better than what I first proposed. If auto-regeneration is ever wanted for another reason, it needs a provisioned PAT or App token; reusing an existing unrelated secret (e.g. the nightly tagging PAT) is not something this job should assume the scopes of.

**Trade-off:** the per-status counts are gone from the index. The rows are still there to count, and `--check` still catches real drift.

## 2. The one real flake

`create_no_window_flag_set` has failed three times (2026-06-28 at 5s, 08-13 at 15s, 09-07 at 15s×2). The first two were "fixed" by widening the timeout, which the test's own comment calls moving the goalpost. **This does not widen it a third time.**

The diagnostics the 08-13 fix added finally paid off:

- **AV scanning is falsified for this run** — `RealTimeProtectionEnabled: False`, `OnAccessProtectionEnabled: False`, `BehaviorMonitorEnabled: False`.
- **It is the first spawn specifically** — both attempts burned their full 15s (~49s total), then the *other six* node-spawning tests in the same binary finished in **~0.8s combined**.

So `warm_node()` pays node.exe's cold start outside the measured window. It is unasserted and time-boxed — if warm-up fails or hangs, the real attempts still run and still produce the genuine diagnostic, so it cannot relocate the flake. `ATTEMPTS` 2 → 3 as cheap insurance (only paid on failure). It is `#[cfg(windows)]`, matching its only caller, so the ubuntu/macos legs get no dead-code warning (ReAgent P2).

`PLAN_WINDOWS_CI_SUBPROCESS_IO_FLAKE_FIX_2026_08_13.md` records the falsified hypothesis and moves to `active`.

## Verification

- `cargo test -p agentmux-srv --test subprocess_io -- --test-threads=1` — **7 passed, 0 failed** in 0.53s
- `gen-docs-index.sh` then `--check` round-trips clean on the regenerated index
- `ci-pr.yml` parses (`js-yaml`) and the edited step was inspected in the parsed output

## What I did not do

- **No blanket retry wrapper** — it would fix 4% of failures and mask real ones.
- **Nothing deleted.** `create_no_window_flag_set` guards a real regression (stdout reaching a console instead of the pipe), which reproduces on every attempt, so the test still catches it.


a5af:
@codex review

GenericAgentX-asaf:
Both P1s were right and the auto-regen approach is gone.

**Codex's P1, verified against the API:** `default_workflow_permissions` is `read` at **both** repo and org level, and `RETRO_GHCR_PUSH_403_2026_06_18.md` records the org has *locked* it — a workflow's `permissions:` block can only restrict, never elevate. The `git push` and `gh pr create` would both have 403'd. Dead on arrival, and I'd have shipped it.

**ReAgent's P1** compounds it: a `GITHUB_TOKEN`-created PR doesn't trigger `pull_request` workflows, so that auto-PR would also have had no checks.

**Replacement needs no token at all.** The section headers carried a row count (`### proposed (95)`) — that single derived scalar is the only thing that merged *wrongly*. Rows from two branches three-way-merge to the correct union; the count silently resolved to one side, leaving N+1 rows under a count of N. Removing it kills the failure class, so the gate goes back to **blocking** (a failure now means real drift). While this PR was open, #3057 landed — another agent hand-fixing the same stale index, which is the failure mode itself.

**ReAgent P2 fixed:** `warm_node()` is `#[cfg(windows)]` like its only caller, so no dead-code warning on the ubuntu/macos legs.

Verified: `gen-docs-index.sh --check` → `INDEX.md is current` on the rebased tree; `subprocess_io` 7/7 in 0.53s.

GenericAgentX-asaf:
Both fixed in `f554971a2`.

**P1 (changeset)** — real, and the consequence you named is the important part: `task release` would have written a VERSION_HISTORY entry describing an auto-regen workflow that isn't in the diff. Replaced with one describing the shipped behaviour (count removed from the committed headers; check stays blocking).

**P2 (redundant cfg)** — flattened. It also turned up a second thing: the `CommandExt` import went with it, because on `tokio::process::Command` (unlike `std`'s) `creation_flags` is inherent, so the trait import was only earning an unused-import warning. `subprocess_io.rs` now compiles warning-free; 7/7 in 0.53s.

a5af:
@codex review

GenericAgentX-asaf:
Rebased onto current main — the conflict was `INDEX.md`, again, which is the failure this PR removes. Third time today: #3053 fixed it by hand, #3057 fixed it by hand a second time, and it just conflicted a third time on the branch that stops it happening.

No content change from the approved head; the two commits replay identically and the index was regenerated to resolve. @reagentx-workflow review again

PR #3058, opened 2026-09-07
Korp@claudius: feat(srv): migrate --verify doctor pass over applied migrations — migration hardening Phase 1b

GenericAgentX-asaf:
## Summary

`SPEC_MIGRATION_SYSTEM_HARDENING_2026_08_03` **Phase 1b**. Phase 1a (#3043, #3047) made a *failed* migration fatal. This closes the other half of F4/F5: `db_migrations` records only that `up()` returned `Ok` — never that it wrote what it was meant to — and nothing re-checked that. That is the F1/F2 "marker present, wrote nothing" shape the spec was written about.

## What ships

- **`Migration::verify(&ctx) -> VerifyOutcome`** — optional, read-only, cheap post-condition on the trait. Default `NotVerifiable`; the pass reports that honestly rather than guessing. Outcomes: `NotVerifiable` / `Ok(evidence)` / `Mismatch(what is missing)` / `Error(check could not run)`.
- **`agentmux-srv migrate --verify`** — runs `verify()` for every migration its tracking store records as **applied** (unapplied are skipped; `--list` covers them), prints one line per migration + a summary, emits matching `migration_verify` / `verify_complete` NDJSON, and exits **2** (`VERIFY_FAILED_EXIT_CODE`) on any `Mismatch` or `Error`. Distinct from 1 so a caller can tell "the data is inconsistent" from "the migrate command broke". `docs/exe-return-codes.md` updated.

## Real checks (the spec's own priority list)

| Migration | Post-condition |
|---|---|
| **0007_agents_consolidate** (the incident) | if `db_agent_definitions` + `db_agent_instances` have rows, `db_agents` must not be empty. Pure `consolidate_verdict()` with its own tests; counts via a shared `table_count()` that treats a missing table as 0 rows. |
| **0002_block_zones_v1** (legacy marker) | channel store present but `migration_agent_zones_v1.flag` absent ⇒ the helper bailed before doing work (it deliberately skips the marker on that path) while `up()` still returned `Ok`. |
| **0003_template_sessions_v1** | **deliberately not verifiable**, reason in the file: its helper explicitly *ignores* its marker and re-asserts the data invariant on every startup, so a "mismatch" could not persist and would mean nothing. |

## Tests

The harness is generic over the registry slice, so it is tested with fake migrations against real temp stores: only applied ones are checked; Channel-scoped migrations consult the channel store (unapplied when absent); a migration with no `verify` impl reads `NotVerifiable` and passes; mismatch/error → exit 2 (`≠ 1` asserted); `table_count` on a missing table is 0. `migrations::` slice **72/72** (was 64). `cargo check` clean.

## Remaining in Phase 1 → "1c"
Wiring the verify report into `muxspect`, which the spec wants coordinated with PR #2390's owner rather than duplicated. Spec Status line says exactly that.


a5af:
@codex review

GenericAgentX-asaf:
All six review points addressed in 48899a87f:

- **codex P2 (read-only):** `--verify` dispatches before any `Store::open*`; the whole path uses `SQLITE_OPEN_READ_ONLY` rusqlite connections (including m0007's check). A typoed data dir reads as "no store here" and creates nothing — tested.
- **codex P1 (swallowed tracking errors):** applied state now comes from a fallible `read_applied_ids`; an unreadable store makes every migration of that scope `VerifyOutcome::Error` (exit non-zero) instead of being skipped. Missing file / pre-`db_migrations` store is `Ok(empty)`. Tested both at the harness and the reader.
- **codex P1 (data dir):** bootstrap's migrate dispatch falls back to the canonical `AGENTMUX_DATA_DIR` first, same precedence as `Config::from_env_and_args`.
- **reagent P2 (duplicate rule):** m0007 now calls Phase 0b's `consolidate_looks_incomplete(&conn)`; `consolidate_report` is formatting only.
- **reagent P2 (exit code):** `VERIFY_FAILED_EXIT_CODE = 3` (clap owns 2); exe-return-codes lists both; test asserts `!= 1 && != 2`.
- **reagent P2 (NDJSON):** detail is serialized via `serde_json::to_string`; test round-trips `\n`/`\t`/quotes/backslashes through `serde_json::from_str`.

Also merged main + regenerated INDEX.md for the doc gate. migrations:: 76/76.

a5af:
@codex review

a5af:
@codex review

GenericAgentX-asaf:
Fixed the `--help` text in 238d0511a — it now says exit 3 and explains why 2 is reserved for clap. Nothing else changed.

a5af:
@codex review

PR #3062, opened 2026-09-07
Korp@claudius: fix(srv): migrations run under a cross-process lock — migration hardening Phase 3

GenericAgentX-asaf:
## Summary

`SPEC_MIGRATION_SYSTEM_HARDENING_2026_08_03` **Phase 3** (failure mode F6). No cross-process locking existed at the migration-runner level: two `agentmux-srv` processes against the same data dir (crash-restart race, misconfigured channel) could both see a migration as pending and both run `up()`. SQLite serializes individual statements, not the runner's check → run → mark-applied sequence — a classic TOCTOU.

## Mechanism

A dedicated one-table SQLite file beside the shared store (`migrations.lock.db`) with **`BEGIN IMMEDIATE` held for the whole batch**. That *is* a cross-process advisory lock — SQLite takes an OS-level file lock — and it gets three things a hand-rolled lockfile does not:

- it **blocks** (`busy_timeout`, 30 min — matching the launcher's extended ESTART deadline) instead of failing
- it is **released by the OS when the holder dies** — no stale-lockfile cleanup
- it is **portable to Windows**

It is deliberately *not* a transaction on the real stores: migrations open their own connections to those (m0007 opens the channel store itself), so a lock held there would deadlock them. No new dependencies.

The lock is taken **before** the pending check on both writers — the in-process runner (`apply_pending`) and the CLI apply path (`run_migrate_command`) — so check → run → mark is one critical section. `--verify`, `--list` and `--dry-run` do not take it.

## Tests (the spec's own acceptance)

- `two_concurrent_runners_on_one_data_dir_apply_a_migration_exactly_once` — two threads run `apply_pending` against the same fresh dir with a deliberately slow fake migration; asserts `up()` ran once, **both** runners returned `Ok` (the loser sees the winner's row and applies nothing), exactly one reported `applied=1`, and a third run is a no-op. This is also Phase 5's "concurrent-boot test".
- `lock_is_exclusive_across_connections_and_released_on_drop`.
- Runner factored as `apply_pending(regs, home, shared_store_path, data_dir)` so the test needs no env vars; `run_pending_migrations` resolves paths and delegates with `REGISTRY`.

`migrations::` slice **82/82**.


a5af:
@codex review

GenericAgentX-asaf:
Added Phase 6 (docs consolidation) to this PR in ae4fbb40f rather than racing the specs-index gate with a second PR: a **What actually ships** table at the top of `SPEC_MIGRATION_FRAMEWORK_2026_06_24.md` (in-process, fatal, backups, `--verify`, this PR's lock, two tracking tables, remaining markers; its Failure Handling / Launcher Integration sections annotated as the superseded subprocess design and why it still exists), plus `docs/recovery/RUNBOOK_MIGRATION_RECOVERY_2026_09_07.md` (living), which supersedes the hardening spec's Part 4. Docs only.

GenericAgentX-asaf:
Addressed all four findings in f77a8d2f5:

- **reagent P1 / codex P2 (dry-run under the lock):** `--list` and `--dry-run` never touch the lock now; they use the same `open_stores_for_preview` / `pending_of` helpers as the batch. New test: a preview completes while another holder has the lock.
- **reagent P1 (stores opened before the lock on the CLI path):** the CLI apply path *is* `apply_pending` now — the identical locked batch the daemon runs, with an `ApplyProgress` sink for its NDJSON lines — so the lock precedes `Store::open*` on both paths.
- **codex P1 (lock keyed only to the shared store):** the batch now takes two locks in fixed order — beside the shared store, then beside the channel store (`<data_dir>/db/migrations.lock.db`) — so isolated-auth instances that share a data dir but resolve different shared stores still serialize. New test drives exactly that shape (two shared paths, one data dir, channel-scoped fake) and asserts `up()` ran once.

Runbook + both specs updated for the two-lock design.

PR #3063, opened 2026-09-07
Camper@claudius: feat(abf): authoring UI for per-provider instruction overrides — closes the last ABF v0.2 gap

GenericAgentX-asaf:
## Summary

Closes the last open gap in **ABF v0.2**. Storage, export (`instructions/<provider>/AGENTS.md`) and import all shipped in #2521/#2523/#2527, and `SPEC_ABF_V0_2_PROVIDER_AWARE_COMPONENTS_AND_NATIVE_MEMORY_2026_08_10.md` §2.2 committed to shipping *"the storage shape, the authoring UI, and export/import"* — only the authoring half was ever built.

Until now an imported bundle's variants round-tripped through the form untouched but could not be **created or edited in the app at all**. The field's own tooltip admitted it: *"there's no authoring UI for those yet."*

## The one subtle thing: wipe-safety

The draft carries this field as a **raw JSON string**, not a parsed structure. That is deliberate and load-bearing — reagent P1 on #2523 found that editing *any* field of an imported bundle silently wiped its provider variants, because `bundle_memory_upsert`'s `ON CONFLICT UPDATE` overwrites the column unconditionally. Round-tripping the string verbatim is what fixed it.

An authoring UI has to turn that string into a list, which is exactly where that fix could be undone. So the seam enforces it: **malformed input is reported as malformed and yields no entries**, rather than yielding an empty list the UI would then serialize back over the top of. When the stored value is unreadable the editor refuses to edit and says the raw value is preserved, instead of rendering an empty editor that destroys it on the next save.

## Changes

**Model (`memory-model.ts`)** — where the correctness lives, and what the tests target:
- `parseInstructionsByProvider` / `serializeInstructionsByProvider`
- `providerKeyProblem` / `duplicateProviderKeys` — advisory checks mirroring the export-side rules

**UI (`BundleProviderInstructionsSection.tsx`)**, following the existing `BundleMcpSection`/`BundleSkillsSection` split:
- Add/edit/remove `provider → instructions` rows; known providers offered via `datalist`, free text still allowed (import accepts unknown keys, so the editor must too).
- **Flags keys that would not survive export.** `bundle_export.rs` runs each key through `sanitize_context_relative_path` and *skips* unsafe ones with a warning, and keeps only one of a colliding pair. Both are silent data loss at export time, so they are surfaced while the user is typing. Advisory only, never blocking — same posture as the Validate button.
- Controlled/uncontrolled hybrid, reasoned in the file header: rows are local state serialized out per edit, because deriving them from the JSON string on every keystroke would make a row vanish the instant its key went blank (a blank key cannot exist in a JSON object). An effect re-seeds when the incoming string changes for any reason other than our own write.

Also corrects the Instructions tooltip, which still claimed no authoring UI existed, and now states the **override-not-append** semantics.

## Test plan

- [x] 11 new model tests: sorted parse; **empty vs malformed** (the wipe-safety property, including an object with a non-string value, which is rejected rather than coerced); key trimming without content trimming; `parse → serialize` round-trip fidelity; both advisory checks
- [x] Written test-first — confirmed all 11 failed before the implementation existed
- [x] Pre-existing `instructions_by_provider` round-trip suite from #2523 still green
- [x] `npx vitest run` — **3826 passed**, 3 skipped
- [x] Changeset added (minor)

## Not covered

No screenshot — this is an editor form section, and I have no way to drive the GUI headlessly here. Manual check of the actual rendering would be worth doing before release.

### DIFF
--- a/.changesets/1788768360-feat-abf-authoring-ui-for-per-provider-instruction-overrides-ir4m.md
+++ b/.changesets/1788768360-feat-abf-authoring-ui-for-per-provider-instruction-overrides-ir4m.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(abf): authoring UI for per-provider instruction overrides (instructions_by_provider)
--- a/frontend/app/view/memory/BundleProviderInstructionsSection.tsx
+++ b/frontend/app/view/memory/BundleProviderInstructionsSection.tsx
@@ -25,7 +25,7 @@
  * our own last write (switching bundles, a fresh load).
  */
 
-import { createEffect, createMemo, createSignal, For, Show, type JSX } from "solid-js";
+import { createEffect, createMemo, createSignal, For, Index, Show, type JSX } from "solid-js";
 import { showTextInputContextMenu } from "@/app/store/contextmenu";
 import { PROVIDERS } from "@/app/view/agent/providers/catalog";
 import {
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

PR #3064, opened 2026-09-07
Korp@claudius: feat(jekt): cross-channel sender verification — DELIVERY=channel, TRUST=channel-verified (spec Phase B)

GenericAgentX-asaf:
## Summary

Phase B of `docs/specs/SPEC_JEKT_CROSS_CHANNEL_TRUST_2026_09_02.md` — the last same-machine jekt tier that had no sender verification. Until now a jekt forwarded between two AgentMux instances on one machine (e.g. a `task dev` build ↔ the stable install) rendered `DELIVERY=host TRUST=self-declared`: the strongest-sounding label with nothing behind it, because the receiving instance holds no HMAC key for an agent it didn't spawn (spec §1–§2). Phase A (#2959) published each agent's Ed25519 public key in the shared registry; this PR is the half that verifies against it.

**Strictly additive, per spec §10:** a message can move from `self-declared` to `channel-verified`, never the reverse. Phase C (forcing `sensitive` on a *failed* cross-channel signature) is deliberately **not** here — the spec holds it until published keys have propagated (§6). The forcing line and the test that flips are called out in §10.2 so it's a one-line follow-up.

### What changes

| Piece | Where |
|---|---|
| `source_channel` + `channel_sig` on the wire; server-only `channel_verified` (`skip_deserializing`, same guarantee as `sig_verified`/`lan_verified`) | `agentmux-common/src/api_types.rs`, `agentmux-srv/src/backend/reactive/types.rs` |
| Sender signs every outgoing jekt with `sign_channel_jekt` (LAN key + `AGENTMUX_CHANNEL`), alongside `jekt_sig`/`lan_sig`. Returns a named struct with `into_request` used by all three send paths, so no site can drop a field or transpose the three `Option<String>`s | `agentmux-mcp/src/main.rs` |
| srv injects `AGENTMUX_CHANNEL` into the MCP env from the same `local_channel_id()` the shared-registry writer uses (new helper; four duplicated env reads in `registry.rs` now go through it) | `backend/agent_config.rs`, `backend/reactive/registry.rs` |
| **D2** `verify_cross_channel_signature` (+ `_in` testable core): gated on `delivery_tier == "channel"`; same-instance senders left to the HMAC path (§D2 step 2); pubkeys from every shared entry for the claimed agent with a non-empty key (empty = pre-Phase-A = `None`, §6); 300 s window (§D6); `Some(false)` is `warn!`-logged, not escalated | `server/reactive.rs`, wired into `handle_reactive_inject` |
| **D4** Tier 2a/2b forwards label the hop `channel` via `same_host_forward_tier` (host/unlabelled → `channel`; `lan`/`wan`/`channel` preserved so the peer re-derives its own tier's verification). The receiver's `resolve_delivery_tier` honours a full-key caller's claim exactly as it already does for `lan`/`wan` — claiming `channel` only *adds* a check | `server/reactive.rs` |
| `TRUST=channel-verified` rendering; on the channel tier without a verified channel signature the marker falls back to the host-tier labels, not `network-claimed` (no network boundary was crossed) | `backend/reactive/sanitize.rs` |
| `channel-verified` joins the 2026-08-17 verified-sender list (`ESCALATE=none`) | `backend/reactive/handler.rs` |
| `EchoTrust` gains `channel_verified`; every literal updated | `server/reactive.rs`, `server/websocket.rs` |

### One deviation from the spec text, documented in D4

The spec says the receiver can identify a forward because "the forwarding peer authenticates with an `auth_key` belonging to a different channel's registry entry". It can't: the forward authenticates with the **receiver's own** full key (that's what the shared entry publishes, §2.4), so a forward is indistinguishable from a local caller by auth alone. The forwarding side labels the hop instead. Safe on the same grounds `resolve_delivery_tier` already rests on for `lan`/`wan` claims. Spec §D4 now carries an "as built" note explaining this.

### Tests (spec §9 items 1–11)

- `verify_cross_channel_signature_tests` (11): valid; unsigned with key on file → `Some(false)`; wrong key; pre-upgrade empty key → `None`; unknown sender → `None`; same-instance sender is a no-op; freshness boundary both sides of 300 s; agent live in two channels verifies against either key; **a LAN signature replayed as a cross-channel one fails** (D5 domain separator); **a chan-a signature replayed as chan-b fails** (D5 channel binding); tier scoping. Plus the forward-tier mapping.
- Handler (4): `TRUST=channel-verified` rendering; keyword match + verified → `ESCALATE=none` (§9 item 11 / §8.2, the escalation-fatigue fix); `Some(false)` **not** forced sensitive in Phase B (annotated to flip in Phase C); `Some(false)` never relaxes a STOP.
- Marker rendering, `resolve_delivery_tier` channel claim, MCP env carries `AGENTMUX_CHANNEL`.
- Entries are written with `write_shared_entry_for_test` (explicit pubkey) so the tests don't depend on the process-global `OnceLock` pubkey resolver another test owns.

```
cargo test -p agentmux-srv -- reactive inject_jekt_signing   → 251 passed
cargo test -p agentmux-common jekt_sign                      → 43 passed
cargo build -p agentmux-mcp                                  → ok
check-doc-status / gen-docs-index --check / check-spec-citations → ok
```

### Docs

Spec Status + §D4 as-built note + new §10.2 "Phase B as built"; `CLAUDE.md` jekt section gains the `DELIVERY=channel` paragraph and lists `channel-verified` in the `ESCALATE=none` list (repo-owner-directed unfinished-work item, real code + tests behind it — not a prose-only policy change); audit report row 8 updated.

### Not in this PR

- Phase C enforcement (one line + one test flip, after propagation).
- Phase D escalation chaining (its own spec, §7.3).
- The §9 integration item — two live instances messaging across channels. Same caveat §10.1 carries for Phase A; worth actually doing given `REPORT_JEKT_SIGNING_KEY_INJECTION_GAP_2026_08_16.md`.

### DIFF
--- a/.changesets/1788769847-feat-jekt-cross-channel-sender-verification-delivery-channel-l48g.md
+++ b/.changesets/1788769847-feat-jekt-cross-channel-sender-verification-delivery-channel-l48g.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(jekt): cross-channel sender verification — DELIVERY=channel, TRUST=channel-verified (spec Phase B)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -701,13 +701,34 @@ signing (issue #2586's other half) does not exist yet — an arbitrary
 non-reagent WAN jekt's `source_agent` remains exactly as forgeable as
 before, still forced sensitive per the reagent-only exception above.
 
+**`DELIVERY=channel` (same machine, DIFFERENT AgentMux instance — 2026-09-07,
+`docs/specs/SPEC_JEKT_CROSS_CHANNEL_TRUST_2026_09_02.md` Phase B):** a jekt
+forwarded between two instances on this machine (e.g. a `task dev` build and
+the stable install) used to render as `DELIVERY=host` with
+`TRUST=self-declared` — the strongest-sounding label with no verification
+behind it, because the receiving instance holds no HMAC key for an agent it
+didn't spawn. It now carries its own label. `TRUST=channel-verified` means
+the sender's Ed25519 signature verified against the public key that agent's
+own instance published in the host-global shared registry — the same
+strength of claim as `host-verified`/`lan-verified`, so it is in the
+`ESCALATE=none` verified-sender list below. Without a verified channel
+signature the marker falls back to the host-tier labels (`self-declared`
+for an agent that never registered or was registered before this shipped;
+`host-verified`/`unverified` if the receiver happens to hold the sender's
+HMAC key), never `network-claimed` — no network boundary was crossed. **A
+present-but-failed cross-channel signature is logged but NOT yet forced
+sensitive** — that is the spec's Phase C, deliberately held until published
+keys have propagated (spec §6/§10); until then treat an unverified
+`DELIVERY=channel` sender exactly as you would `self-declared`.
+
 ### Tier rules
 
 - `TIER=info` / `TIER=coord` — routine work; you may act and the human sees
   the marker. As of the 2026-08-15 narrowing, this is now the DEFAULT
   outcome for clean-content jekts at every trust level — `TRUST=host-verified`,
   `TRUST=self-declared`, `TRUST=network-claimed` (LAN or WAN), WAN
-  `SIG=verified`, and LAN `TRUST=lan-verified` all land here unless one of
+  `SIG=verified`, LAN `TRUST=lan-verified`, and cross-channel
+  `TRUST=channel-verified` all land here unless one of
   the forced-sensitive cases below applies. `sensitive` is meant to be the
   rare case, not the default.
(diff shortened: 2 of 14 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed in the follow-up commit — thanks, this was a real gap.

`InjectionResponse` now carries `channel_verified` (set from the request at every success site, same thread-it-through pattern as `effective_tier`/`requires_stop`), and `forward_inject_to_peer` builds the echo's `EchoTrust` through a new `echo_trust_from_peer_body` that takes the receiver's verdict from the response body and leaves the rest untouched (an older peer that omits the field keeps the caller's value). Two unit tests on the helper plus an assertion on the handler response; spec §10.2 records the decision.

PR #3065, opened 2026-09-07
Korp@claudius: feat(muxspect): migrations command — the migrate --verify doctor report from inside a live instance (hardening Phase 1c)

GenericAgentX-asaf:
## Summary

Migration hardening **Phase 1c** — the last item in Phase 1 of `docs/specs/SPEC_MIGRATION_SYSTEM_HARDENING_2026_08_03.md`: the `migrate --verify` doctor report, available from inside a running instance as `muxspect migrations`.

The spec's own diagnosis (§1.6) is the motivation: `muxspect` exists to answer "why did this agent fail to start", and a migration *recorded as applied that wrote nothing* is a legitimate answer it could not give. Phase 1b (#3058) built the check as a CLI you run against a stopped install after locating the srv binary and the channel data dir; this makes the same answer one command away from any agent or shell pane.

### Design: one producer, two consumers

| Piece | Where |
|---|---|
| `migrations::doctor_report` → `DoctorReport { migrations: [DoctorRow], summary, tracking_errors, exit_code, … }` — every registry migration's `applied` / `pending` / `unknown` state and, for applied ones, the `Migration::verify` verdict (`ok` / `MISMATCH` / `error` / `n/a`) with its evidence. Read-only end to end: same `open_readonly` path as `--verify`, never opens a `Store`, never creates a file | `agentmux-srv/src/migrations/runner.rs` |
| `cmd_verify` (`agentmux-srv migrate --verify`) rebuilt on the report — identical printed lines and NDJSON events, summary and exit code taken from the report, so the CLI and the live route cannot drift | same |
| `GET /api/v1/muxspect/migrations` → `doctor_report_for_instance(get_wave_data_dir())` on the blocking pool; 500 with the reason when the shared root is unresolvable | `server/muxspect_handlers.rs`, `server/mod.rs` |
| `muxspect migrations [--json]` — rows in registry order, a `MISMATCH` shouted as "recorded applied, effect is NOT there" with a pointer to the runbook, the `--verify` summary line, and the same exit code (3 on any mismatch or error) so a script can gate on either | `backend/shellintegration/muxspect.mjs` |

`unknown` (tracking store unreadable) is reported per scope with the cause named once under `tracking_errors`, not once per migration — the codex P1 fallible-read rule from #3058 carried through.

### Tests

- `doctor_report` ×3: fresh data dir (all pending, `checked=0`, exit 0, **creates nothing**, serialises with `verify` omitted on pending rows); an applied `0007` on a fresh store (`ok`, counts in the detail); an unreadable shared store (global rows `unknown`/`error`, channel rows still `pending`, cause named once, exit 3).
- vitest ×4 on `renderMigrations` (rows + summary, MISMATCH + runbook pointer + "read-only", unreadable tracking store) and `parseArgs`.

```
cargo test -p agentmux-srv -- migrations::                       → 87 passed
npx vitest run agentmux-srv/src/backend/shellintegration/muxspect.test.mjs → all passing
check-doc-status / check-spec-citations                          → ok
```

### Docs

`docs/MUXSPECT.md` gains "Checking migration state (`migrations`)"; the runbook's §4 now leads with `muxspect migrations` for a running app and keeps the CLI for a stopped one or another channel; hardening spec Status marks Phase 1 complete (1a #3043/#3047, 1b #3058, 1c here); audit report row 6 brought up to date — it still read "Phase 0 of 6".

### Not in this PR

Phases 2 (marker unification), 4 (needs a product decision), and the rest of 5 of the hardening spec remain, per its Status line.


a5af:
@codex review

PR #3066, opened 2026-09-07
Korp@claudius: test(migrations): Phase 5 regression suite against the real registry — hardening spec Phase 5

GenericAgentX-asaf:
## Summary

Migration hardening **Phase 5** — the framework-level regression tests `docs/specs/SPEC_MIGRATION_SYSTEM_HARDENING_2026_08_03.md` lists, run against the **real registry** (all 25 migrations, not fakes) in a throwaway `~/.agentmux`. With this, every phase of the hardening spec except 2 (marker unification) and 4 (a product decision) has shipped.

New file: `agentmux-srv/src/migrations/phase5_tests.rs` (6 tests). Each test sets `AGENTMUX_HOME_OVERRIDE` under the crate-wide `ISOLATED_AUTH_ENV_LOCK` so every env-resolved path the migrations use lands inside the temp home — the same discipline `registry::paths` and `runner` already follow — and marks the temp home as an existing install so `m0001` never looks at the developer's real `~/.waveterm`.

### What each test pins

| Spec item | Test | What it proves |
|---|---|---|
| 1. The incident (§1, F1/F2) | `a_stale_consolidate_marker_beside_an_empty_target_is_not_stamped_and_the_next_boot_backfills` | A `migration_agents_consolidate_v1.flag` beside an **empty** `db_agents`: `m0000_bootstrap` refuses to stamp 0007 off the flag alone (Phase 0a's content check), the next boot's `apply_pending` actually runs the backfill, and `verify_applied` then reports `ok` with the counts. |
| 2. Crash between effect and mark | `a_crash_between_a_migrations_effect_and_its_mark_resumes_without_duplicating_rows` | After a full boot, delete 0007's `db_migrations` row (effect + marker file intact) and boot again: exactly that migration re-runs, its content-checked marker short-circuits, **no rows are duplicated**, it is marked again, and a third boot is a no-op. `agents_consolidate.rs` claimed this in a comment; now it is a test. |
| 3. Upgrade skipping many versions | `an_upgrade_that_skips_many_versions_applies_every_later_migration_once_in_registry_order` | Only `0000`–`0005` recorded (each in its own scope's store); one boot applies every later migration exactly once, **in registry order** (asserted through `ApplyProgress`), records each in its scope, and a further boot is a no-op. |
| 4. Concurrent boot | — | Shipped with Phase 3 (#3062) as `runner::lock_tests`. |
| 5. `m0000_bootstrap` stamping | `bootstrap_on_a_fresh_install_stamps_nothing_and_creates_no_channel_store`, `bootstrap_stamps_every_pre_framework_marker_it_recognises`, `bootstrap_is_idempotent_on_a_second_run` | `m0000` had **no tests at all**. A fresh install stamps nothing and conjures no channel store; every recognised pre-framework marker (zones flag, a consolidate flag backed by real rows, the four shared-registry dot-files, the `agents/` install proxy) stamps exactly its migration into the right scope — `0011` stays unstamped without identity accounts, `0008` is never stamped by design; a second run is idempotent. |

Legacy rows go in through SQL against the schema `Store::open` lays down, since no live API writes `db_agent_definitions` any more.

```
cargo test -p agentmux-srv -- migrations::phase5_tests   → 6 passed (0.9 s)
cargo test -p agentmux-srv -- migrations::               → 90 passed
check-doc-status / check-spec-citations                  → ok
```

### Docs

Hardening spec Status marks Phase 5 shipped (Phases 2 and 4 remain) and the Phase 5 section points at the file; audit report row 6 updated.


a5af:
@codex review

GenericAgentX-asaf:
Both addressed in the follow-up commit. (1) The crash test now removes the marker file as well as the db_migrations row, i.e. the transaction-committed-before-marker window the spec names, and proves the from-scratch retry does not duplicate rows; the marker-present window stays as its own test. (2) Fixture agents now carry a working directory inside the temp home plus an env content row pinning CLAUDE_CONFIG_DIR to the temp shared/providers/claude, and the crash test plants a memory file there and asserts m0024 imported it, so the redirect is proven rather than assumed. @reagentx-workflow review again

PR #3067, opened 2026-09-07
Korp@claudius: chore(deps): drop the unused react dependency left over from the React→SolidJS migration

GenericAgentX-asaf:
## Summary

Closes the last residue item on the React → SolidJS row of `docs/reports/REPORT_LARGE_MIGRATIONS_COMPLETION_AUDIT_2026_09_06.md` (scoreboard row 2): the `react` package was still a runtime dependency although no source file imports it (390 files import `solid-js`, none import `react`, per the audit's §2.2).

- `package.json`: `react` removed from `dependencies`.
- `package-lock.json`: only the `node_modules/react` and `node_modules/loose-envify` entries (react's sole transitive dep) and the root reference are removed — pruned surgically rather than via `npm uninstall`, because the local npm re-normalised ~50 unrelated `peer`/`devOptional` flags on the way, which would have made a 2-line change read as a 200-line one. A guard in the prune step asserted nothing else in the lock still depends on either package.
- Audit report row 2 marked done.

```
npx tsc --noEmit -p tsconfig.json   → 0 errors
npm ls react                        → (empty)
npm ls --depth=0                    → no invalid / missing / extraneous
```

The other two "done" rows' residue is deliberately not swept here: the Tauri comment residue is 44 mentions across 29 source files and belongs in its own commit; §5.2–5.4's doc restamps already landed in #3042.

### DIFF
--- a/.changesets/1788772081-chore-deps-drop-the-unused-react-dependency-left-over-from-t-lti0.md
+++ b/.changesets/1788772081-chore-deps-drop-the-unused-react-dependency-left-over-from-t-lti0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+chore(deps): drop the unused react dependency left over from the React→SolidJS migration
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in the follow-up commit: the root package descriptor in package-lock.json still mirrored react under devDependencies (which is where package.json had it, not dependencies). Removed there too; grep for react in the lockfile is now 0 and npm ci --dry-run reports no error. @reagentx-workflow review again

GenericAgentX-asaf:
Already fixed in 141af14b7 (the commit after the one this review ran on): the root package descriptor no longer lists react anywhere, grep for react in package-lock.json is 0, and npm ci --dry-run is clean.

PR #3068, opened 2026-09-07
Korp@claudius: feat(docs): weekly docs stale-sweep — recurring enforcement for docs-lifecycle Phase 5

GenericAgentX-asaf:
## Summary

Docs-lifecycle hardening **Phase 5, item 2** (`docs/specs/SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md`): the recurring sweep the spec calls "the phase that actually prevents this report from becoming doc #1,147 nobody acts on". Two manual audits (2026-08-07, 2026-09-06) found the same rot — Status lines and file citations that were true when written and silently stopped being. `check-doc-status.sh` (item 1, already on every PR) stops the backlog *growing*; nothing re-verified a doc once it was merged. Now something does, weekly, with the output kept on one standing issue instead of a third one-off report.

### `scripts/docs-stale-sweep.mjs`

Flags a doc when all three hold: it claims to be current (Status not `historical`/`superseded`), nobody has touched it in more than N weeks (default 8), and it cites a file that has changed since — or that no longer exists. Exactly the spec's rule: a cheap git-recency check on cited paths, not a re-read.

- One `git log --format=@%ct --name-only` pass over the whole history (0.7 s locally) gives every tracked path's last-commit time; no per-file git calls.
- Citations are backticked or bare slash-paths with a source/config extension (line refs stripped), resolved against `git ls-files` exactly, by unique suffix, or by unique basename — never guessed between ambiguous names (`main.rs`); a file counts once however many ways it is written. Example-shaped tokens (`/path/to/foo.ts`, `.ts/.tsx`) are filtered — a first run surfaced them.
- Output: markdown (sorted missing citations > most drift > oldest, `--limit` rows, counts always full) or `--json`. Exit 0 always — a triage list, not a gate.
- 13 vitest cases on the pure parts.

First real run against `main`:

| | |
|---|---|
| docs scanned | 1303 |
| claiming to be current | 896 |
| … untouched > 8 weeks | 444 |
| … **flagged** (cite a changed or missing file) | **421** |
| no Status line | 391 |

That number is the point: it is the measured size of the problem, and it will now be visible every week until it shrinks.

### `.github/workflows/docs-stale-sweep.yml`

Mondays 06:30 UTC plus `workflow_dispatch` (adjustable `weeks`). Full-history checkout (the report says so if it isn't), runs the script, publishes to the step summary and an artifact, then **upserts one open issue** labelled `docs-stale-sweep` (label created `--force`) — edits the body in place, never opens a second. Permissions: `contents: read`, `issues: write`.

### Docs

Spec Status brought up to date — Phase 3 (generated `INDEX.md`) had shipped without the Status line saying so; Phase 5 now shipped; Phase 2 remains. The specs README's reader guardrail points at the sweep and how to run it locally.

```
npx vitest run scripts/docs-stale-sweep.test.mjs   → 13 passed
node scripts/docs-stale-sweep.mjs --weeks 8       → 0.7 s, report above
check-doc-status / check-spec-citations           → ok
```


a5af:
@codex review

GenericAgentX-asaf:
Both Codex points addressed in the follow-up commit: resolveCitation now distinguishes resolved / ambiguous / absent and only an absent slash-path lands in missing (test added with a two-way reducer/tests.rs); shallow clones are detected with git rev-parse --is-shallow-repository instead of an empty-map inference. ReAgent errored twice before producing findings on this PR. @reagentx-workflow review again

GenericAgentX-asaf:
Done: the weeks input now reaches the script only via an env var (WEEKS), never interpolated into the run block; the script already rejects a non-numeric value. @reagentx-workflow review again

PR #3069, opened 2026-09-07
ci(docs): scope the specs-index gate to PRs that touch docs/specs (#3059)

### LINKED ISSUE #3059: docs/specs/INDEX.md staleness red-lights every open PR; the gate asserts instead of regenerating
## Problem

`docs/specs/INDEX.md` contains a committed, generated count header (`### proposed (N)`). The `doc status + grep gates` CI job asserts it is current via `scripts/gen-docs-index.sh --check`.

Because it is one shared generated file, **any** PR that adds a spec (or changes a spec's `Status`) without regenerating the index makes `main` stale — and from that moment the gate fails on **every open PR in the repo**, none of which touched docs.

## Evidence from a ~2 hour window today

Three separate regeneration PRs were needed: #3050, #3051, #3053.

My unrelated Rust PR (#3048, one `.rs` file + a changeset) failed this gate **twice, in opposite directions**:

1. Committed `### proposed (96)` vs regenerated `(95)` — with the gate's own log printing `check-doc-status: no docs changed` immediately before failing.
2. After I fixed that in #3051 and `SPEC_RPC_BINDINGS_CODEGEN_2026_09_07.md` landed (855 → 856 specs), the same gate failed again the other way: committed `(95)` vs regenerated `(96)`.

Reproduced on a clean `main` checkout both times, so it is not a feature-branch artifact.

## Why it is worth fixing rather than absorbing

- The failure is unattributable: the PR that *causes* it is not the PR that *fails*. Whoever is unlucky enough to open a PR next pays for it.
- The fix is mechanical and carries no review value — three PRs in two hours whose entire content was a one-line count change.
- It scales with contributor concurrency, which is the direction this repo is going.

## Options

1. **Have CI regenerate rather than assert** — run `gen-docs-index.sh` and fail only if the *non-generated* half of the file was hand-edited.
2. **Stop committing the count header.** The counts are derivable; the rows are the useful part.
3. Keep asserting, but only when the PR itself touches `docs/specs/` — this narrows the blast radius to the PRs that can actually cause it, though it still lets `main` drift.

(1) or (2) removes the class. (3) only stops it from spreading.

Filed after hitting it three times in one session while fixing an unrelated LAN bug.

AgentO-asaf:
Closes the half of #3059 that #3056 could not. Fixes #3059.

## What #3056 already fixed, and what it can't

**#3056 removed the section-header counts** — the right call, and it fixed the *concurrent* case. Those counts were the only part of the file whose correctness depended on what **other** branches did: two branches each adding a spec merged cleanly into an arithmetically wrong total that neither computed. Rows are per-line, so with the counts gone concurrent regenerations compose.

**What remains is the non-arithmetic case**: a branch adds a spec and never regenerates at all. The row is simply missing — no arithmetic involved, so no amount of making the counts merge-safe helps. Under an unscoped `--check` that lands on **whoever opens a PR next**.

That is the failure #3059 describes: *"the PR that causes it is not the PR that fails."* It cost five mechanical regeneration PRs in one session (#3050, #3051, #3053, #3057 + one more), each blocking work with no docs in it. My own #3055 was red-lighted by it twice, in opposite directions.

## The change

`--check` now asserts only when the branch actually touches `docs/specs/`:

- **Touches specs** → must regenerate. Caught on the branch that omitted it.
- **Touches no specs** → cannot have made the index stale, so not asked about it.
- `--check-all` → the old unconditional assertion, kept for local use and for auditing the index itself. CI deliberately does not use it.
- Unresolvable base ref → skips rather than fails, matching `check-doc-status.sh`.

This is also what the gates on **either side of it in the same job** already do — `check-doc-status.sh` and `check-spec-citations.sh` are both changed-files-only. The index check was the odd one out, and the workflow step is renamed to say so.

This is #3059's option 3, which its author correctly called incomplete *on its own*. Paired with #3056 it isn't: #3056 handles the concurrent-regeneration case, this handles the forgot-to-regenerate case, and between them the unattributable failure is gone. Neither would have been sufficient alone — worth stating since option 3 was the one the issue ranked lowest.

## Why not option 1 (CI regenerates and commits)

It removes the class too, but it needs CI to push to `main`, which collides with this repo's "never push directly to main" rule and with branch protection. Left for the docs owner to decide; this change doesn't preclude it.

## Verification

In a **scratch clone**, all three paths:

| Case | Expected | Result |
|---|---|---|
| Branch with no spec changes | skip, exit 0 | ✅ `no specs changed on this branch — index not asserted` |
| Branch adds a spec, no regen | fail, exit 1 | ✅ `INDEX.md is STALE` |
| Same branch, after regenerating | pass, exit 0 | ✅ `INDEX.md is current` |

`bash -n` clean. No test harness exists for these scripts, so this is behavioral verification rather than an added test — worth knowing rather than assuming.

### DIFF
--- a/.changesets/1788783591-ci-docs-scope-the-specs-index-gate-to-prs-that-touch-docs-sp-f7bb.md
+++ b/.changesets/1788783591-ci-docs-scope-the-specs-index-gate-to-prs-that-touch-docs-sp-f7bb.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+ci(docs): scope the specs-index gate to PRs that touch docs/specs
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -83,7 +83,11 @@ jobs:
       # means real drift — a spec added without regenerating — which is the
       # author's to fix and takes one command. See the history note at the top
       # of scripts/gen-docs-index.sh for why it used to fail for everyone.
-      - name: Specs index is current
+      # Scoped to PRs that actually touch docs/specs (issue #3059) — an
+      # unscoped assertion failed every open PR whenever main went stale,
+      # including PRs with no docs in them at all. Same posture as the two
+      # gates either side of it.
+      - name: Specs index is current (changed specs only)
         run: bash scripts/gen-docs-index.sh --check
 
       - name: Spec citations resolve (changed files only)
--- a/scripts/gen-docs-index.sh
+++ b/scripts/gen-docs-index.sh
@@ -121,8 +121,12 @@ should_check() {
     # spec out of docs/specs/ (say, reclassifying it under docs/reports/)
     # showed up as a non-spec path, skipped the check, and left the index
     # carrying a row for a file no longer there. Verified against git 2.55:
-    # `--name-only --find-renames` prints just `docs/reports/X.md`, while
-    # `--name-status` prints `R100  docs/specs/X.md  docs/reports/X.md`.
+    # for a spec moved out of the tree, `--name-only --find-renames` prints
+    # the destination path alone, while `--name-status` prints an `R100` row
+    # carrying the old path and the new one. (Spelled out rather than shown
+    # as a literal example path — check-spec-citations.sh reads any
+    # spec-shaped path in a comment as a citation, and a made-up one is a
+    # dangling citation by its definition. It caught exactly that here.)
     #
     # Deliberately UNLIKE check-doc-status.sh, which skips pure renames (R100)
     # because relocating a file makes no claim about its Status. The opposite
--- a/scripts/gen-docs-index.sh
+++ b/scripts/gen-docs-index.sh
@@ -116,9 +116,29 @@ should_check() {
         echo "gen-docs-index: cannot resolve base ref '$base' — skipping."
         return 1
     fi
-    # INDEX.md itself counts: a hand-edit to it must still be caught.
-    if git diff --name-only --find-renames "$ref"...HEAD 2>/dev/null \
-        | grep -qE '^docs/specs/.*[.]md$'; then
+    # --name-status, not --name-only, and BOTH sides of a rename (Codex P2 on
+    # PR #3069). --name-only reports only a rename's DESTINATION, so moving a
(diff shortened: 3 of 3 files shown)


a5af:
@codex review

AgentO-asaf:
**Both P2s were right and are fixed in `416762f`.** I verified each empirically rather than taking them on description.

## P2 — include the generator in the trigger

Correct, and **this PR was the proof**: it changes `gen-docs-index.sh`, and CI logged

```
gen-docs-index: no specs changed on this branch — index not asserted.
```

A generator change can alter the output with no spec touched, so `INDEX.md` would drift silently until some later spec PR failed for it — the unattributable failure this PR exists to remove, reintroduced through the generator instead of through `main`. `scripts/gen-docs-index.sh` is now in scope.

## P2 — inspect both sides of a rename

Also correct. Reproduced in a scratch clone on git 2.55, renaming a spec to `docs/reports/`:

```
$ git diff --name-only --find-renames $BASE...HEAD
docs/reports/SPEC_864_LAYOUT_SINGLE_WRITER_2026_06_30.md      <- destination only; matches ^docs/specs/ zero times

$ git diff --name-status --find-renames $BASE...HEAD
R100  docs/specs/SPEC_864_...md   docs/reports/SPEC_864_...md  <- both sides
```

So the check skipped while the index kept a row for a file no longer in the tree. Now uses `--name-status` and matches either side via `awk '{for (i=2; i<=NF; i++) print $i}'`.

One deliberate divergence worth flagging, since it looks like an inconsistency: **`check-doc-status.sh` skips pure renames (`R100`) on purpose** — relocating a file makes no claim about its `Status`. The opposite holds here. Index rows are keyed on filename, so a pure rename is precisely the kind of change that makes the index stale. Both behaviors are right for their own gate; the reason is now in a comment so the next reader doesn't "fix" one into the other.

## Verification

Scratch clone, three cases:

| Case | Expected | Result |
|---|---|---|
| Generator changed, no specs touched | assert | ✅ asserts |
| Spec renamed out of `docs/specs/` | assert, report stale | ✅ asserts, `INDEX.md is STALE` |
| Unrelated change (neither) | skip | ✅ skips |

PR #3070, opened 2026-09-07
Korp@claudius: fix(migrations): 0002 no longer trusts its marker file — content-idempotent re-run, content-checked stamping and verify (hardening Phase 2)

GenericAgentX-asaf:
## Summary

Migration hardening **Phase 2** — marker unification — from `docs/specs/SPEC_MIGRATION_SYSTEM_HARDENING_2026_08_03.md`. Acceptance: *no migration in the registry has an internal marker whose existence (rather than verified effect) is trusted as complete.* `0007` got there in Phase 0b (`consolidate_looks_incomplete`) and `0003` never trusted its marker. `0002_block_zones_v1` was the holdout, in three places, and it had the exact F1 shape the spec was written about: a restored `objects.db`/`filestore.db` beside a stale `migration_agent_zones_v1.flag` would skip every block's conversation for good, get recorded applied, and pass `--verify`.

With this, every phase of the hardening spec has shipped except Phase 4 (a product decision about the stalled SQL consolidation).

### What changes

| Before | After |
|---|---|
| `migrate_block_zones_v1` returned early when the flag existed. The flag was also load-bearing for idempotency: without it each re-run minted a **new archive zone** (timestamp bumped until unique). | **Content-idempotent.** An archive is written only when no archive for that agent already holds the same bytes (existing archives read once per agent and extended as the run writes, so two identical block snapshots in one run collapse too); `:current` is seeded only when empty. The marker short-circuit is gone; the marker is still written at the end as evidence. New stat `archives_already_present`. |
| `m0000_bootstrap` stamped `0002` applied off the flag alone. | Stamps only when the flag exists **and** `block_zones_look_incomplete()` is false — the same rule Phase 0a gave `0007`. A data dir with no filestore stamps on the flag as before (nothing to migrate). |
| `m0002::verify` reported "marker present" as the post-condition. | Content-based: Mismatch when an agent block still holds a snapshot whose agent `:current` zone is empty; the marker is reported as a note. |
| — | `migrations/mod.rs` framework doc gains the **"No private marker files"** rule the spec asked for: `db_migrations` + a `verify()` post-condition is the sole source of truth; the three legacy markers are evidence only. |

`block_zones_look_incomplete(wstore, filestore)` is read-only and lives beside the helper in `agent_session/migrations/v1_blocks.rs`.

### Tests

- `migration_is_idempotent` now proves idempotency by **content**: the second run scans the block, writes nothing, leaves exactly one archive — and deleting the marker between runs changes nothing.
- `a_stale_marker_no_longer_skips_unmigrated_blocks`: flag present, block snapshot orphaned → the helper migrates it anyway.
- `block_zones_look_incomplete_*`: no blocks / empty snapshot / populated `:current` are all "complete".
- Phase 5 suite: the `0002` incident shape end to end against the real registry — stale flag + orphaned snapshot → bootstrap leaves `0002` pending → the boot migrates it and `:current` holds the conversation.

```
cargo test -p agentmux-srv -- migrations:: agent_session::tests   → 133 passed
check-doc-status / check-spec-citations                            → ok
```

### Docs

Spec Status (Phase 2 shipped; only Phase 4 remains) and a note on the Phase 2 section; runbook §5 no longer tells operators to delete the `0002` flag; audit report row 6.

### Not in this PR

Removing the marker files themselves — they still serve `m0000`'s pre-framework stamping and the doctor's evidence line, and retire with the migrations under the framework's "Retiring Old Migrations" rule. Phase 4.


a5af:
@codex review

GenericAgentX-asaf:
Both addressed in the follow-up commit. P1: migrate_block_zones_v1 and block_zones_look_incomplete are now fallible — an unreadable blocks table fails m0002::up (MigrationError, fatal per Phase 1a), makes bootstrap fail rather than stamp, and makes verify report Error rather than Ok; a test drops db_block and asserts Error. P2: new Store::open_read_only and FileStore::open_read_only (SQLITE_OPEN_READ_ONLY, busy_timeout only, no WAL pragma, no schema or version stamp) and m0002::verify uses them; a test asserts verify creates neither store. @reagentx-workflow review again

PR #3071, opened 2026-09-07
Manoz@Area54: refactor(mcp): split main.rs (4,727 lines) into tool_schemas and window_capture modules

GenericAgentX-asaf:
## Summary

Audit step 17 (`docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md` §5): *"Split `agentmux-mcp/src/main.rs` into `tools/` by tool family."* It was **4,727 lines**. Two clean seams come out; **no behavior change**.

| File | Lines | What |
|---|---|---|
| `main.rs` | 4,727 → **3,239** | dispatcher, transport, everything else |
| `tool_schemas.rs` | **716** | the 50 `*_TOOL` JSON schema consts |
| `window_capture.rs` | **814** | window capture/discovery + the audit log both write |

## Why these two seams

- **`tool_schemas.rs`** is pure data — 50 consts, no logic. It was ~15% of the file and meant every schema wording tweak showed up as a diff against the file that also holds the dispatcher.
- **`window_capture.rs`** was the largest single cluster in the file (20 items: `CaptureTier`, `enumerate_agentmux_windows`, `capture_window_impl`, `own_instance_pids`, the audit-log writers…) and is one concern. Coupling was measured before moving it: it calls exactly **one** function defined outside itself (`agent_slug`), and uses **no** outside types.

`main.rs` glob-imports both modules, so `call_tool`'s call sites and `mod tests`'s `use super::*` resolve exactly as before — **the test module did not move and did not change**.

## Why this is safe

- **All 50 schemas are byte-identical to `main`**, verified mechanically: extract each const body from `git show main:…/main.rs` and from the new file, strip only the added `pub(crate) `, compare. 50/50 identical, and the **declaration order is unchanged**, so the advertised tool list is the same.
- The split was scripted with asserted anchors — the const-block extractor refuses to run unless it finds exactly 50 consts, no stray top-level items inside the range, and every const referenced somewhere outside its own definition (so nothing became dead code).
- Visibility changes are the minimum the compiler demanded: `pub(crate)` on the 19 moved items named from outside, their 11 struct fields, `CaptureTier::{allowed,label}`, and `agent_slug`. Nothing became `pub`.

## Verification

- `cargo build -p agentmux-mcp` — clean, **no warnings in the crate**
- `cargo test -p agentmux-mcp -- --test-threads=1` — **26 passed, 0 failed**. These are the tests that matter here: 16 of them reference moved items (`capture_window_dir`, `own_instance_pids`, `CaptureTier`, `looks_unrendered`, `window_listing_entry`, …), so they exercise the re-export path directly.
- `cargo test --workspace -- --test-threads=1` — **4,072 passed, 0 failed**

## Not in this PR

`persistent.rs` (7,151 lines) is the other half of step 17 and is left alone deliberately — it is live block-controller state, not a data/leaf-cluster split, and deserves its own PR.


a5af:
@codex review

GenericAgentX-asaf:
Fixed in `3d67b2e29` — good catch, and the interesting part is *which* doc block it was.

My extractor anchored on the `fn` line, so the doc comment above `capture_window_dir` stayed behind and landed directly above `agent_slug`. That block records why captures go to `AGENTMUX_DATA_HOME` rather than `std::env::temp_dir()` — world-readable on a multi-user host, and `CaptureWindow` can capture arbitrary OS windows — which is reagent's own P2 from #2709. Stranding that rationale away from the function it constrains is the real cost, not the misattribution.

I also swept all three files for the same class (doc block with no item to attach to): none left. 26/26 still pass, crate still warning-free.

GenericAgentX-asaf:
Two corrections, then a request.

**It's already fixed** — `3d67b2e2`, pushed before your reply. No follow-up or fast-follow needed; `capture_window_dir`'s doc block is on `capture_window_dir` in `window_capture.rs` now.

**On severity:** agreed it's a P3 and not structurally broken, but I'd push back on "cosmetic." The block is the record of *why* captures go to `AGENTMUX_DATA_HOME` instead of `std::env::temp_dir()` — world-readable on a multi-user host, and `CaptureWindow` can capture arbitrary OS windows. That's your own P2 from #2709. A security rationale sitting on the wrong function is how the next person "simplifies" it back to `temp_dir()`. Cheap to fix now, so I did.

For the record on the sweep: my "none left" was scoped to the doc-block-with-no-item class, which is what I checked and what I claimed — Codex raised this one, and it was a real gap in what I thought to assert.

**Re-review needed regardless:** `dismiss_stale_reviews` is on, so the push dropped the approval and the PR is `REVIEW_REQUIRED`/`BLOCKED`. It can't merge without one.

PR #3073, opened 2026-09-07
fix(ci): generate the specs index from tracked files, not a filesystem glob

AgentO-asaf:
Follow-up to #3069. The generator globbed `docs/specs/*.md` off disk, so an **uncommitted** spec in the working tree counted as part of the spec tree.

## Why this is more than a local annoyance

Regenerating with an untracked spec present **writes a row for it into `INDEX.md`**. Commit that `INDEX.md` without the spec and the index asserts a file nobody else has — a link that resolves for exactly one person.

It also made `--check` report `STALE` on an otherwise clean checkout, which means the gate's own advice — *"Run: bash scripts/gen-docs-index.sh"* — was the thing that introduced the bad row.

I hit this repeatedly while working on the tray PRs and worked around it by `mv`-ing the file aside before every regeneration. That only works while someone remembers to do it.

## The change

Enumerate via `git ls-files` rather than a glob:

- Reads the **index**, not HEAD, so a `git add`ed spec counts immediately — the ordinary "write it, add it, regenerate, commit both" flow needs no commit-then-regenerate dance.
- Deleted-but-still-tracked paths are filtered out, so a failed read can't become a row.
- Outside a git repo (a tarball) it falls back to the glob rather than emitting an empty index.

## One trap worth naming

A git pathspec's `*` matches `/` as well (fnmatch without `FNM_PATHNAME`), so `git ls-files -- 'docs/specs/*.md'` **also matches `docs/specs/archive/*.md`** — the one subtree this index excludes on purpose. My first attempt did exactly that and grew the index by **76 archive rows**.

What caught it was asserting the regenerated file was *byte-identical* on a clean checkout, rather than just eyeballing that it ran. Depth is now bounded by a `grep -E '^docs/specs/[^/]+\.md$'`, not by the pathspec, with the reason in a comment.

## Verification

Scratch clone throughout:

| Case | Expected | Result |
|---|---|---|
| Clean checkout | byte-identical output (CI unaffected) | ✅ identical |
| Untracked spec present | not in index; `--check` not stale | ✅ 0 rows, `INDEX.md is current` |
| Same spec after `git add` | appears | ✅ 1 row |
| Archive subtree | never in the generated section | ✅ 0 rows |

The clean-checkout case is the load-bearing one: in CI tracked and on-disk are the same set, so this must not change what CI sees — and it doesn't.

Also confirmed against a real working tree holding an uncommitted spec: `--check-all` now reports `INDEX.md is current` where it previously reported `STALE`.

### DIFF
--- a/.changesets/1788785889-fix-ci-generate-the-specs-index-from-tracked-files-not-a-fil-kc4u.md
+++ b/.changesets/1788785889-fix-ci-generate-the-specs-index-from-tracked-files-not-a-fil-kc4u.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): generate the specs index from tracked files, not a filesystem glob
--- a/scripts/gen-docs-index.sh
+++ b/scripts/gen-docs-index.sh
@@ -161,10 +161,54 @@ fi
 # per status word — 730 files x 8 statuses, ~5,800 process spawns — which took
 # over two minutes on Windows. Now each file is read once into a status->rows
 # map, which is ~730 spawns and seconds.
+# ── Which files ARE the spec tree ───────────────────────────────────────────
+#
+# Tracked files, not a filesystem glob. The index describes the COMMITTED
+# spec tree, and a bare `docs/specs/*.md` glob does not: it also picks up
+# work-in-progress specs that are not in the repo yet.
+#
+# That is not a cosmetic difference. Regenerating with an untracked spec
+# present writes a row for it into INDEX.md, and committing that INDEX.md
+# without the spec leaves a link to a file nobody else has — the index
+# asserting the existence of something that does not exist. It also made
+# `--check` report STALE for a perfectly clean checkout, so the honest
+# response to the gate ("run the generator") was the thing that introduced
+# the bad row.
+#
+# `git ls-files` reads the INDEX, not HEAD, so a `git add`ed spec counts
+# immediately — the normal "write the spec, add it, regenerate, commit both"
+# flow works without a commit-then-regenerate dance.
+#
+# Deleted-but-still-tracked paths are filtered by the -f test: `git rm`'d in
+# the working tree but not yet staged, the file is gone and must not get a
+# row built from a failed read.
+#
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentO-asaf:
**Both P2s confirmed and fixed in `aac507a`.** I reproduced each failure first, then re-ran with the fix removed to check the reproduction actually bites.

## Unmerged index stages

Right, and **more load-bearing than it first looks**: this script's own header documents regenerating *during* a conflict as the way to resolve one —

```
# MERGE CONFLICT IN INDEX.md? Do not hand-resolve it.
#   git checkout --ours docs/specs/INDEX.md && bash scripts/gen-docs-index.sh
```

— so being mid-merge here is a documented normal state, not an edge case. And as you note, the completeness assertion cannot catch it: both totals come from the same duplicated list.

| | rows for one conflicted spec |
|---|---|
| Without the fix | **2** |
| With the fix | **1** |

Used `awk '!seen[$0]++'` rather than `--deduplicate`: it preserves `ls-files`' sorted order and needs no git >= 2.31, and this script is run by contributors, not only CI. Same result, wider floor.

## Repository does not own this checkout

Also right, and the consequence is worse than a wrong answer — it's a **silent wipe**. Reproduced by unpacking a source tree inside an unrelated repo:

```
git sees a repo here? /private/tmp/outer/.git
tracked specs here:   0
rows after run:       0     <- INDEX.md emptied
```

`0 == 0`, so the completeness assertion had nothing to say here either. Both of these slipped past it for the same structural reason, which is worth noting: that assertion compares two numbers derived from the same enumeration, so it cannot detect a *wrong enumeration* — only an inconsistent one.

Fixed by treating an **empty** tracked list as "git did not answer for this tree" and falling through to the glob — one condition covering both this and the plain not-a-repo case, rather than a separate ownership probe. With it: 1 row, index intact.

## Regression guard

Clean-checkout output re-verified **byte-identical** after both changes, so CI still sees exactly what it saw before.

PR #3074, opened 2026-09-07
Manoz@Area54: refactor(srv): record each RPC command's request/response types at registration (Phase 2 step 1)

GenericAgentX-asaf:
## Summary

Phase 2 step 1 of `docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md`, per the design in `docs/specs/SPEC_RPC_BINDINGS_CODEGEN_2026_09_07.md` §3.4: **make the command → request/response mapping exist as data.** No generator, no new dependency, no behavior change.

## Why this comes first

The audit's headline number is 334 "keep in sync with the agentmux-srv RPC" comments across 284 hand-written stubs and a 2,819-line hand-maintained `gotypes.d.ts`. The obvious fix is to derive the TypeScript from the Rust — and the spec's §2 finding is that **that alone cannot work**:

> which command takes which request struct and returns which response is not data anywhere in srv

Each handler picks its request type inline (`serde_json::from_value::<CommandXData>(data)`) and builds its response ad hoc (`serde_json::to_value(&x)`, `json!({"ok": true})`). A type generator can emit all 190 structs and still not know how to pair them into a callable stub. So the pairing has to exist before any generator is worth writing.

## What's here

- **`backend/rpc/schema.rs`** — `RpcSchema` / `RpcBinding`: one row per typed command, holding the fully qualified request and response type names, keyed and returned in command order so the output is byte-stable across runs (what a future CI gate diffs).
- **`RpcEngine::register_typed<Req, Resp>`** — registers a handler *and* records the pair. It wraps the closure in exactly the deserialize → run → serialize steps every hand-written handler already performs, so migrating one is not supposed to change the wire at all.
- **Two handlers migrated as proof** — `agentdefhide` and `agentdefunhide` (`CommandAgentDefHideData` → `AgentDefHideResult`). Their bodies are unchanged apart from dropping the two lines `register_typed` now does for them.

Handlers still on `register_handler` contribute nothing to the schema and are simply absent. That absence is deliberate and is what makes a 284-stub migration possible one command at a time.

## Verification

- `cargo test -p agentmux-srv register_typed` — **2 passed**, both asserting properties rather than smoke:
  - the schema records the exact request/response type names for both commands, and the paths are fully qualified (a rename that would silently change generated bindings fails here); and `listagents`, still untyped, is asserted *absent*.
  - a migrated handler's wire behaviour is unchanged — same success body, and a malformed request still errors with the same `agentdefhide:` prefix the hand-written body used.
- `cargo test --workspace -- --test-threads=1` — **4,092 passed, 0 failed**
- No new dependency; `cargo check -p agentmux-srv` clean.

## What is deliberately NOT here

**The dump mode and CI gate** (spec §3.2/§3.3). Building them surfaced the exact risk the spec flagged in §5:

> The dump mode must register every handler without a live `AppState`. If registration turns out to need one, the registry moves to a `const` table next to `commands.rs` — less elegant, same gate.

Handler registration does need a live `AppState` (stores, broker, network). A CLI `--dump-rpc-schema` would have to stand up a throwaway data dir to emit anything, and the alternative — exporting from a test, the way `ts-rs` itself does — is a different enough shape to deserve its own review. Rather than pick one in passing at the end of this change, the registry lands first and proves it works; the dump and the gate follow against something already exercised.

**The TypeScript generator and `ts-rs`.** Adding a proc-macro dependency to `agentmux-srv` is its own decision with its own build-time cost, and the spec asks for that cost to be measured. It belongs on top of a registry that already exists, not bundled with the change that creates it.

So this PR does not yet remove a single sync comment. It makes removing them possible, which is the part the spec identified as load-bearing.

### DIFF
--- a/.changesets/1788786183-refactor-srv-record-each-rpc-command-s-request-response-type-88z4.md
+++ b/.changesets/1788786183-refactor-srv-record-each-rpc-command-s-request-response-type-88z4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(srv): record each RPC command's request/response types at registration (RpcSchema + register_typed) — the mapping a bindings generator needs, which did not previously exist as data
--- a/agentmux-srv/src/backend/rpc/engine.rs
+++ b/agentmux-srv/src/backend/rpc/engine.rs
@@ -16,6 +16,7 @@ use tokio::sync::mpsc;
 use uuid::Uuid;
 
 use super::super::rpc_types::{RpcContext, RpcMessage, RpcOpts, COMMAND_EVENT_RECV};
+use super::schema::RpcSchema;
 
 // ---- Constants (match Go) ----
 
@@ -180,6 +181,10 @@ struct EngineInner {
     #[allow(dead_code)]
     auth_token: String,
     rpc_context: Option<RpcContext>,
+    /// Request/response types of every command registered through
+    /// [`WshRpcEngine::register_typed`]. Lives beside `handlers` so the
+    /// mapping is a side effect of registration and cannot drift from it.
+    schema: RpcSchema,
 }
 
 /// Core RPC engine: handles incoming RPC requests, dispatches to registered
@@ -211,6 +216,7 @@ impl WshRpcEngine {
                 active_handlers: HashMap::new(),
                 auth_token: String::new(),
                 rpc_context: None,
+                schema: RpcSchema::new(),
             }),
             output_tx,
         });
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in `f902dfae`.

**Codex P2 — real, and mine.** `register_typed` recorded the pair, but replacing that command through `register_handler`/`register_stream_handler` overwrote the handler and left the binding behind, so `schema_json()` would advertise types the live handler no longer accepts. A generator reading that emits a stub that fails at runtime — a stale binding is worse than a missing one. Both untyped paths now evict, and `register_typed` installs through a private helper so it doesn't delete the entry it just wrote. Regression tests cover typed→untyped, typed→streaming, and that ordering.

**The vitest failure was also mine**, which I nearly wrote off as unrelated to a Rust-only change. `test/contract/rpc-contract.test.ts` derives the backend surface by resolving `register_handler(NAME, …)` to a const; inside `register_typed` that argument is the function's own parameter, so the extractor correctly refused to proceed with an incomplete picture rather than passing quietly.

The important part of that fix: it also had to learn `register_typed` as a registration form. Without it, every command migrated in later phases would silently leave the contract — the failure mode getting *worse* as Phase 2 progresses. The skip is scoped to the engine file **and** the identifier `command` together. I tried a whole-file exclusion first and it was wrong: `engine.rs` also holds `#[cfg(test)]` registrations the baseline counts, and dropping the file shrank the compared surface from 28 to 24. A bare identifier allow-list would have been worse — it would silently skip a real handler using `command` as a local.

Two smaller ones: the extractor is comment-blind, so my own prose `register_handler (which evicts)` parsed as a call (reworded); and the three `#[cfg(test)]` command names the new tests register are baselined next to `echo`/`failme`/`checkctx`.

Rust 4,095 passed; vitest 258 files / 3,865 tests passed.

GenericAgentX-asaf:
Fixed in `39e998af6` — correct on both counts, and it was a consequence of my own earlier fix in this PR.

When I addressed Codex's P2, `register_typed` stopped calling `register_handler(command, …)` and started using the private `install_call_handler`, which the extractor's regex deliberately doesn't match. That made the skip branch unreachable and left its comment asserting a forwarding call that no longer exists — worse than no comment, since it would have told the next person extending the extractor something false.

Removed the branch and the `enginePath` binding with it. I verified unreachability first (no `register_handler`/`register_typed` call in engine.rs takes a non-literal argument), and the contract test still passes 4/4 without the branch, which is the actual proof. Net −9 lines.

The engine-side comment about the extractor being comment-blind stays — that trap is still live and still real.

vitest: 258 files / 3,865 tests passed.

PR #3075, opened 2026-09-07
Korp@claudius: feat(storage): db_agents carries the latest launch state — schema v29 + m0025 backfill (agent consolidation Phase 3b, PR 1 of 4)

GenericAgentX-asaf:
## Summary

PR 1 of 4 finishing the agent-concept consolidation (`docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md` Phase 3b/3c), per the hardening spec's Phase 4 decision to **finish** rather than park it. The end state is one `db_agents` table; `db_agent_definitions` and `db_agent_instances` go away in PR 4.

### Why this PR comes first

The inventory (every remaining legacy read/write, FK and caller) showed that `db_agents` is a complete mirror of the definitions table but **not** of the instances table: it never got `session_id`, `status`, `started_at`, `ended_at`. Phase 3a left those "on the block", but in practice three things still read them from `db_agent_instances`, and that is exactly what kept the table alive:

- the recent-sessions picker (`instance_list_named`),
- the pane close/reopen continuity write-back (`persist_session_id` → `instance_get_by_block_id` + `instance_update_partial`, `SPEC_PANE_CLOSE_REOPEN_CONTINUITY_GUARANTEE_2026_07_27.md` §4.1),
- the status-filtered `instance_list`.

One row per agent means one launch state per agent; a continuation moves it to the new block/session rather than adding a row (the dual-write already folds continuations that way — only the state was missing).

### What changes

| Piece | Where |
|---|---|
| Schema **v29**: `session_id`, `status`, `started_at`, `ended_at` on `db_agents` (CREATE TABLE + additive ALTERs, doc entry). Deliberately no unique index on `db_agents(slug)` — template-launch projections copy their template's slug. | `backend/storage/migrations.rs` |
| Dual-write keeps them current: instance **create** (fold into a user clone, continuation onto the chain head, template launch as its own row) and instance **update** (session capture, status, ended_at; `last_block_id` moves only when the write carries a block) | `backend/storage/dual_write.rs` |
| `m0025_agents_launch_state_backfill` (channel scope): latest instance per projection key — a frozen copy of the v29 projection rule, per the framework's "freeze copies of live logic" doc — copied into any consolidated row whose launch state is still entirely default. Idempotent, never overwrites state the dual-write wrote since, no-op once the legacy table is gone. | `migrations/m0025_agents_launch_state_backfill.rs` |
| Tracking spec: resumption note + the four-PR plan | `docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md` |

Nothing reads the new columns yet — that is PR 2. After this PR every consolidated row is a complete mirror of its legacy chain, so PR 2 can flip the instance reads without a data gap.

### Tests

- `m0025` ×4: the template chain's head row takes the **child's** (latest) state and the clone row takes its own; idempotent and never clobbers live state; tolerates a store without `db_agent_instances`; `up()` end to end plus a missing channel store.
- `dual_write_instance_lifecycle_keeps_launch_state_current_on_db_agents`: create → session capture → status/ended → continuation moves the head's state → a capture on the continuation reaches the head.

```
cargo test -p agentmux-srv -- migrations:: storage::   → 489 passed
cargo test -p agentmux-srv (full)                       → see CI
```

### The plan (also in the spec)

1. **This PR** — launch state on `db_agents`.
2. Instance read flip + instance writes to `db_agents` only; delete the dead `instance_get_by_name` / `user_clone_defs_for_template`.
3. Definition flip + re-point the six FKs (`db_agent_content`, `_skills`, `_history`, `_identity_links`, `_skills_ref`, `_mcp_ref`) to `db_agents(id)`.
4. Drop both legacy tables, `dual_write.rs`, `agents_consolidate.rs`, `repair_agent_def_gaps`; retire `0007` and its marker; keep `m0004`/`m0006`/`transcript_backfill` tolerant of `db_agents`-only stores.

### DIFF
--- a/.changesets/1788786719-feat-storage-db-agents-carries-the-latest-launch-state-sessi-5yo1.md
+++ b/.changesets/1788786719-feat-storage-db-agents-carries-the-latest-launch-state-sessi-5yo1.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(storage): db_agents carries the latest launch state (session/status/started/ended) — schema v29 + backfill from db_agent_instances (agent consolidation Phase 3b, PR 1)
--- a/agentmux-srv/src/backend/storage/dual_write.rs
+++ b/agentmux-srv/src/backend/storage/dual_write.rs
@@ -302,7 +302,11 @@ impl Store {
                     instance_name = ?7,
                     updated_at = ?8,
                     user_hidden = ?9,
-                    last_block_id = ?10
+                    last_block_id = ?10,
+                    session_id = ?11,
+                    status = ?12,
+                    started_at = ?13,
+                    ended_at = ?14
                  WHERE id = ?1",
                 params![
                     def.id,
@@ -315,6 +319,10 @@ impl Store {
                     now_ms,
                     if inst.display_hidden { 1_i64 } else { 0_i64 },
                     inst.block_id,
+                    inst.session_id,
+                    inst.status,
+                    inst.started_at,
+                    inst.ended_at,
                 ],
             )
         } else if is_continuation {
@@ -345,7 +353,11 @@ impl Store {
                     instance_name = ?7,
                     updated_at = ?8,
                     user_hidden = ?9,
-                    last_block_id = ?10
+                    last_block_id = ?10,
+                    session_id = ?11,
+                    status = ?12,
+                    started_at = ?13,
+                    ended_at = ?14
                  WHERE id = ?1 AND is_template = 0",
                 params![
                     root_id,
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both Codex P1s and the ReAgent P2 addressed in the follow-up commit. (1) The dual-write update now takes the launch-state fields from the newest instance in the projection key's chain rather than from the writer, so a late write from a superseded launch cannot swap the continuation's session back; the lifecycle test asserts exactly that. (2) m0025 now runs repair_def_gaps first, projects a db_agents row for any template-launch chain head that has none (INSERT OR IGNORE, frozen copy of the pass-3 mapping), and only then copies state; tests cover a missing head row and a missing clone definition. (3) The last_block_id comment no longer contradicts the v29 columns. The doc-gate failure was the specs index needing regeneration after the Status restamp; regenerated. @reagentx-workflow review again

PR #3078, opened 2026-09-07
Manoz@Area54: ci(srv): generate the frontend RPC types from Rust and gate their freshness (Phase 2 step 2)

GenericAgentX-asaf:
## Summary

Phase 2 step 2, on top of the registry from #3074: **generate the frontend's RPC types from the Rust ones, and make staleness a build failure.** This is the first change in the series that turns a "keep in sync" comment into something that can actually fail.

## What's here

- **`ts-rs` on `agentmux-srv`**, derive-only, with `serde-compat` so the emitted type reflects the wire shape serde produces (honouring `rename_all` / `default` / `skip_serializing_if`) rather than the Rust field names.
- **`frontend/types/rpc/*.ts`** — generated from the two command types the typed registry already covers, committed.
- **`scripts/check-rpc-bindings.sh` + a CI step** — regenerate, then ask git whether anything changed. A dirty tree means the committed bindings are stale.

## The generated output matches the hand-written stub exactly

This is the part worth checking, because it is the claim the whole plan rests on. `frontend/app/store/rpc-api/agent.ts` declares by hand:

```ts
data: { definition_id: string },
): Promise<{ ok: boolean }>
```

and ts-rs emits, from the Rust:

```ts
export type CommandAgentDefHideData = { definition_id: string, };
export type AgentDefHideResult = { ok: boolean, };
```

Identical shapes — and the generated files additionally carry the Rust doc comments across, which the hand-written stubs do not.

## The gate was verified to actually fail

A gate that has never failed is not evidence of anything, so I broke it on purpose: adding a `drift_probe: bool` field to `CommandAgentDefHideData` and re-running produced

```
check-rpc-bindings: generated RPC bindings are STALE.
-export type CommandAgentDefHideData = {
+export type CommandAgentDefHideData = { drift_probe: boolean,
```

exit 1, with the diff. Reverted afterwards; the committed tree passes.

The script also refuses to run when the generated directory is *already* dirty, because a pre-existing edit there is indistinguishable from drift the run produced, and the diff would blame the generator for someone's uncommitted work.

## Placement and cost

- **The CI step is in the `rust` job, not `docs`.** It needs a Rust toolchain and a built `agentmux-srv`; the steps above it already provide both, so regenerating is nearly free. My first attempt put it in `docs`, which has no toolchain at all.
- **Safe on the Windows leg** because `.gitattributes` normalizes every text file to `eol=lf`, so generated files cannot differ by line ending across platforms.
- **Build cost, which the spec asked to measure:** incremental `cargo check -p agentmux-srv` was **11s before and 11s after** adding `ts-rs`. That measures the incremental path, not a cold build — `ts-rs` and its deps are compiled once, and only two types currently derive `TS`.

## Scope

Only types reachable from commands migrated to `register_typed` carry `#[derive(TS)]`, so only those are generated and gated. The set grows as the migration proceeds (spec §3.4 step 2) and the gate does not need to change as it does.

**Not here: generated call stubs.** A stub needs the frontend's method name (`AgentDefHideCommand`), which is not derivable from the wire command (`agentdefhide`) — it is a frontend naming convention, not data the backend holds. Recording it needs a deliberate decision (a third argument to `register_typed`, or a mapping file), and that is worth its own review rather than being settled in passing here.

So the hand-written stubs still exist; what changed is that the *types* behind two of them are now generated and can no longer drift silently.


a5af:
@codex review

GenericAgentX-asaf:
**P1 (both reviewers, independently) — real, fixed in `519979cf`.**

Confirmed exactly as described: `git diff --quiet -- <path>` exits 0 with a new untracked file in that path, while `git status --porcelain` reports it `??`. Both checks now use `git status --porcelain -uall`.

I verified it against the real scenario rather than a synthetic one — temporarily derived `TS` on a third type (standing in for a future migration whose `.ts` wasn't committed), and the gate now fails with `?? frontend/types/rpc/AgentDefCreateFromTemplateResult.ts`, exit 1, where the previous version reported `ok`. Reverted after; the committed tree passes.

---

**P0 — I believe this one is incorrect, and I'd rather show the measurement than argue it.**

The claim is that `export_to` resolves relative to the crate manifest dir, so `../../` climbs past the repo root and writes outside the checkout, making the gate a permanent no-op.

`ts-rs` resolves `export_to` relative to its **export directory** — `TS_RS_EXPORT_DIR`, defaulting to `<manifest_dir>/bindings/` — not to the manifest dir itself. So from `agentmux-srv/bindings/`, `../../frontend/types/rpc/` lands on `<repo>/frontend/types/rpc/`.

Tested directly by emptying **both** candidate locations and regenerating:

```
repo-internal  frontend/types/rpc     : 2 files   (AgentDefHideResult.ts, CommandAgentDefHideData.ts)
repo-EXTERNAL  ../frontend/types/rpc  : 0 files
agentmux-srv/bindings                 : 0 files
```

Independently corroborated by the drift test in the PR body: injecting a field into `CommandAgentDefHideData` made the gate fail with a diff **against the in-repo file**. That could not happen if the generator were writing outside the repo — the gate would have reported `ok`, which is the very failure mode the finding predicts.

Changing it to `../frontend/types/rpc/` as suggested would resolve to `agentmux-srv/frontend/types/rpc/` and genuinely break it.

Happy to be shown wrong if there's a case I'm missing — e.g. a CI environment that sets `TS_RS_EXPORT_DIR`. We don't set it anywhere, and CI runs the same command I ran.

GenericAgentX-asaf:
I tested the suggested fix rather than just re-asserting my position, and it would break the gate. Three measurements, all on this branch:

**1. Where the current path writes.** Wiped both candidate locations by absolute path, regenerated, listed with timestamps:

```
IN-REPO   .../agentmux/frontend/types/rpc/AgentDefHideResult.ts        Sep 7 09:53
IN-REPO   .../agentmux/frontend/types/rpc/CommandAgentDefHideData.ts   Sep 7 09:53
OUTSIDE   .../manoz-0803a/frontend/                                    (does not exist)
```

Fresh mtimes, in-repo, nothing outside the checkout.

**2. What the suggested `../frontend/types/rpc/` actually does.** Applied it, wiped, regenerated:

```
./agentmux-srv/frontend/types/rpc/CommandAgentDefHideData.ts   <- inside the crate
frontend/types/rpc:  0 files
```

That is the change that would make the gate a no-op — the generator would write to `agentmux-srv/frontend/` while the gate watches `frontend/types/rpc`. Reverted.

**3. Corroboration that doesn't depend on either of us reading docs correctly.** The drift test in the PR body: injecting a field into `CommandAgentDefHideData` made the gate fail *with a diff against the in-repo file*. And this round, temporarily deriving `TS` on a third type made it fail with `?? frontend/types/rpc/AgentDefCreateFromTemplateResult.ts`. Neither is possible if the generator writes outside the repo — both would have reported `ok`.

**On the mechanism:** `ts-rs` resolves `export_to` relative to `TS_RS_EXPORT_DIR`, which defaults to `<manifest_dir>/bindings/` — not to the test's cwd. So `../../` starts at `agentmux-srv/bindings/`, and the two climbs land on the repo root. The cwd-based reading predicts observation (2), which is what applying `../` produced.

I've left the path as-is. If there's an environment that sets `TS_RS_EXPORT_DIR` and changes this, I'd want to know — we don't set it anywhere, and CI runs the same command.

P1 from the same review stands fixed in `519979cf`; the gate passes clean on the restored tree.

GenericAgentX-asaf:
Agreed on process — a reply shouldn't auto-resolve a P0, so here's the explicit re-review request.

To make the mechanism check easy to verify independently, the claim to test is narrow: `ts-rs` resolves `export_to` against `TS_RS_EXPORT_DIR`, whose default is `<manifest_dir>/bindings/` — one level deeper than `<manifest_dir>`. Everything else follows from that one directory. The cheapest reproduction is the `../` experiment: if `export_to` were manifest-relative, `../frontend/types/rpc/` would land at repo root; it actually lands at `agentmux-srv/frontend/types/rpc/`, which is only consistent with starting from `agentmux-srv/bindings/`.

Worth stating plainly: the failure mode described in the P0 is a real and serious one — a freshness gate that silently always passes is worse than no gate, because it looks like coverage. That is exactly why I ran the drift probes rather than reasoning about paths. Both probes failed with diffs against the in-repo file, which a no-op gate could not produce.

PR #3079, opened 2026-09-07
feat(term): frosted-glass band for the terminal pane tab strip

AgentY-asaf:
Spec + implementation. Gives the terminal pane's tab strip the same tinted, full-width glass treatment the agent pane's has, so the two pane types stop looking like different products.

Spec: `docs/specs/SPEC_TERM_PANE_TAB_STRIP_TRAILING_BLUR_2026_09_07.md`

## The thing worth knowing before reviewing

`SPEC_PANE_TAB_STRIP_TRAILING_BLUR_2026_08_12.md` §0 scoped the original blur to agent panes **on purpose**, with a structural reason that still holds:

- **Agent pane:** the strip is `position: absolute` and floats over the live, scrolling conversation. There is real content behind it, which is what read as "purely transparent" and motivated the glass.
- **Terminal pane:** the strip is a normal flex child — a reserved row — with `.view-term`'s own opaque background behind it. `.term-connectelem` starts *below* it.

So `backdrop-filter` on the terminal strip has almost nothing to soften. **This PR ships the visual join-up (tint + full-width band), not a literal blur-over-content**, and the spec and code comments both say so plainly rather than implying more.

The filter is kept anyway because it is **not** always inert: with a terminal background image configured (`computeBgStyleFromMeta`), there *is* texture behind that row, and it blurs for real.

## What was rejected, and why

Making the terminal strip a true overlay so live terminal output blurs underneath (spec §2.2). It's the only way to get the real phenomenon, and the cost is specific to terminals:

- the **top terminal row would render under glass** — fine for scrollback, not fine for `vim`/`htop`/`less` on the alt screen, which use every row;
- `.term-connectelem` growing by the strip height means xterm fits ~1 more row and the **PTY row count changes** — a real `SIGWINCH` to every running shell, not a cosmetic diff.

Padding the terminal to compensate (the agent pane's approach) just collapses back into what this PR does, with extra steps.

## Two consequences handled here

- **`animateWidth` removed** from `term.tsx`'s `<PaneTabStrip>` (spec §3.1). It smooths the box growing/shrinking with tab count, which only happens while the box is shrink-to-fit; at `width: 100%` it would measure, hold and transition a value that never changes. Pills still resize inside the fixed box. The editor strip still uses the prop, so nothing is dead — and its stale "this strip stays shrink-to-fit" comment is replaced rather than left lying.
- **`position: relative; z-index: 1`** on the strip. `term.tsx` renders the `termBg()` background image as an `absolute inset-0 z-0` layer spanning the whole pane *including this row*; being positioned, it would paint over a non-positioned flex child. Spec §4 raised this as an open question; rather than assume a painting order, the fix makes it deterministic — above `termBg`, below the `z-index: 10` search/link overlays. If the bug was real it was pre-existing (tabs obscured whenever a terminal background image was set).

## Verification

- `npx tsc --noEmit` clean.
- `PaneTabStrip.test.tsx` + `frontend/app/view/term/` — 42 tests pass, including the `animateWidth` suite (still exercised via the editor strip).
- **Not yet visually confirmed in a running build.** Bringing a `task dev` instance up on this branch for that; will confirm here. The criteria that matter are spec §5: the band renders full-width and matches the agent pane, and — the one piece of hard evidence — `stty size` inside the pane is **unchanged**, which is what separates this from having accidentally shipped Option B.

Related follow-up: `SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md` (separate PR) proposes hoisting this strip out of the terminal view to stop it remounting on tab switches. That would move the `.view-term > .pane-tab-strip` selector this PR adds; noted in both specs.

### DIFF
--- a/.changesets/1788800443-feat-term-frosted-glass-band-for-the-terminal-pane-tab-strip-xsg9.md
+++ b/.changesets/1788800443-feat-term-frosted-glass-band-for-the-terminal-pane-tab-strip-xsg9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(term): frosted-glass band for the terminal pane tab strip
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -301,6 +301,7 @@ partial list.
 | [`SPEC_TERMINAL_LATENCY_BENCHMARK_2026_05_19`](SPEC_TERMINAL_LATENCY_BENCHMARK_2026_05_19.md) | SPEC: Terminal Input Echo-Latency Benchmark |
 | [`SPEC_TERMINAL_SCROLLBACK_PERSISTENCE_2026_07_23`](SPEC_TERMINAL_SCROLLBACK_PERSISTENCE_2026_07_23.md) | SPEC: Terminal scrollback doesn't survive reconnect (all `view:"term"` panes) |
 | [`SPEC_TERM_DOUBLE_RAF_TEAROUT_2026_05_30`](SPEC_TERM_DOUBLE_RAF_TEAROUT_2026_05_30.md) | SPEC: Remove the terminal Stage-1 RAF write-coalescer (double-rAF) |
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #1111, opened 2026-05-28
feat(agents): instance_list no-status case reads from db_agents (Phase 3b.3a)

AgentA-asaf:
## Summary

Phase 3b.3 — split into three sub-PRs because the existing `updateagentinstance` handler depends on `instance_get` returning transient runtime state (`block_id`, `session_id`, `status`, `started_at`, `ended_at`) that has no analog on the consolidated `db_agents` row. A naive full flip would clobber those fields on write-back.

This PR ships **3b.3a** — the safe slice of `instance_list`.

| Sub-PR | Scope | Status |
|---|---|---|
| 3b.3a | `instance_list` with no `status` filter → `db_agents` | ✅ this PR |
| 3b.3b | `instance_list` status-filter case → `db_agents` after handler refactor | ⏳ pending |
| 3b.3c | `instance_get` → `db_agents` after handler refactor | ⏳ pending |

Tracking discussion: **#1095**. Spec updated in `docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md`.

## Why split

Only the no-status case is safe to flip now:

- Production caller (`listagentinstances` RPC) is invoked with `{}` from `swarm-model.ts::refreshInstances`. The resulting `instancesAtom` has **zero readers** in the frontend — set but never consumed.
- The status-filter branch falls back to a private `instance_list_legacy` helper. Same SQL as today; preserved for any future caller that needs transient state until 3b.3b lands.

## Mapping highlights

- `is_template = 0 AND user_hidden = 0` — same shape as 3b.2.
- `definition_id` filter matches `(id = ? OR parent_template_id = ?)` so it honors both folded user-clones (where `db_agents.id == def.id`) and template-instance projections.
- Continuation chains pre-collapse — one row per logical agent.
- Transient fields come back as type defaults (empty / 0).

## Test plan

- [x] `cargo test -p agentmux-srv --release instance_list_phase` — 4/4 pass
- [x] `cargo test -p agentmux-srv --release instance` — 44/44 pass (no regression)
- [x] `cargo test -p agentmux-srv --release storage` — 147/147 pass
- [ ] reagent + codex pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3080, opened 2026-09-07
Korp@claudius: refactor(storage): the instance API reads and writes db_agents only (agent consolidation Phase 3b, PR 2 of 4)

GenericAgentX-asaf:
## Summary

PR 2 of 4 finishing the agent-concept consolidation (`docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md` Phase 3b/3c). **Every `Store` instance method now reads and writes `db_agents` only** — `db_agent_instances` has no live readers or writers left. Net −700 lines.

Builds on #3075 (schema v29 gave the row its launch state).

### What an "instance" is now

The launch-side view of a non-template `db_agents` row:

| field | value |
|---|---|
| `id`, `definition_id` | both the agent's id (the rule `instance_list` adopted in #1111) |
| `block_id`, `session_id`, `status`, `started_at`, `ended_at` | the row's **latest launch state** (v29) |
| `parent_instance_id` | always empty — chains are pre-collapsed |
| `display_hidden` | `user_hidden` (one flag, one row) |

Launching a **template** creates a new agent row keyed by the launch id; launching a **user agent**, or continuing one, folds into that agent's row.

### Removed as dead

The instance-side dual-write helpers (create / update / set_hidden / delete / repoint, the chain-walking projection-key resolver, its definition loader) — nothing mirrors a launch when there is no second table holding one — plus `instance_get_by_name` and `user_clone_defs_for_template`, neither of which had a production caller. `dual_write.rs` keeps only the definition-side helpers until PR 4.

### Three behaviour changes, each pinned by a test

1. **`instance_create` returns the row the launch landed on**, whose id can differ from the requested one. Callers persist the returned id; the `createagentinstance` handler now responds with it, which is what the frontend already stores as the block's `agentInstanceId`.
2. **A launch's resolved working directory becomes the agent's workspace.** Phase 3a wrote the *definition's* configured cwd; the cross-version registry mirror skips any row whose workdir is not under the agents root, so that became a live regression the moment readers flipped. One agent, one workspace.
3. **Reseeding templates no longer deletes agents launched from them**, and deleting a template no longer deletes its agents. Such a row is the user's agent with its own conversation, exactly as a user clone always was. (Previously the FK cascade removed them.)

### `m0026_registry_agent_id_rekey`

The named-agent registry was keyed by **launch id** and is now keyed by **agent id**. Without a re-key, an upgraded install's existing files become duplicates shown next to the real entry — and "Forget agent" retires the new key while the stale file stays active, resurrecting the agent.

The pass maps each record through the legacy launch rows (launch of a user agent → that agent; continuation → its chain root), leaves records this channel does not own for the channel that does (the registry is host-global), and is idempotent. Four tests including the not-ours and no-legacy-table cases.

### Known gaps, deliberately left to PR 3

Both because `db_agent_identity_links` still has its FK on `db_agent_definitions`:

- an agent that exists only as a `db_agents` row cannot hold account links of its own, so `listrecentsessions` falls back to its template's links (documented at the fallback);
- `agent_def_delete` still owns definition rows while `instance_delete` owns agent rows.

### Tests

The store suite is rewritten to the consolidated model rather than patched: chain-collapse picker tests assert the head row carrying the newest launch; continuation tests assert the agent's own registry record (cross-channel continuity is now structural — the agent id is globally stable — instead of a chain walk); tests pinning removed machinery are deleted with their reasoning recorded in the diff. Two tests captured genuine bugs found on the way: folding a session-less continuation was erasing the agent's live session pointer, and `instance_repoint_definition` re-parented the promote target to itself.

```
cargo test -p agentmux-srv   → 3134 passed, 1 failed
npx tsc --noEmit             → 0 errors
```

The single failure is `backend::sysinfo::handle_anomaly_tests::refresh_processes_specifics_does_not_leak_a_handle_per_call`, a local-environment condition unrelated to this change — the same code passed Windows CI on #3075, and it fails identically on this machine with the storage changes stashed.


a5af:
@codex review

GenericAgentX-asaf:
Review round 1 addressed in 4e54c811c — all five findings accepted, each with a test.

**reagent P1 (session.rs, template shadowing the agent's own config)** — correct and the worst of the five. `def` was redefined to prefer the parent template for what was only meant to be an identity-link fallback, so every agent launched from a template reported the template's `provider` / `model_vendor_base_url` / `agent_created_at` / `agent_type` / `definition_name` and the background summary's provider. The template is now bound separately (`template`) and used only where the FK forces it; every display field reads the agent's own row.

One knock-on worth flagging: `definition_name` now carries the agent's own name, so `handler_returns_sessions_with_previews_sorted_by_snapshot_first` moves from `"Claude Code"` to `"name-inst-recent"`. That is not a UI change — the picker renders `instance_name || definition_name` (`MyAgentsList.tsx`), so the field is only reached for a nameless row.

**reagent P1 (`ON CONFLICT` dropping `working_directory`)** — correct. The fold branch wrote it unconditionally and the insert branch didn't, so a retry of the same launch id kept the first attempt's path. Added, matching the fold branch. Pinned in `launching_a_template_creates_an_agent_row_for_the_launch`.

**codex P1 (workspace not backfilled on upgrade)** — correct, and the one finding that reaches existing installs rather than new writes. Fixed as **`m0027_agents_workspace_backfill`** rather than by widening `m0025`: 0025 shipped in v0.55.39 and is already applied on exactly the upgraded stores this repairs, so amending it would fix nobody who needs it. 0027 copies each agent's latest legacy launch's non-empty `working_directory` onto its row, using the same frozen projection rule and the same `created_at`-then-id tie-break as 0025 so both pick the same launch. Four tests (template launch, user-agent chain, idempotent + empty-value no-op, no legacy table).

**codex P1 (m0026 last-write-wins)** — correct. Records are now grouped by target agent before anything is written, and the greatest `last_launched_at_ms` wins (launch id breaks a tie, so the result no longer depends on `list_active()`'s filesystem order). Losing files are deleted after the winner is written, keeping the "an interrupted run leaves both, never neither" property.

**codex P2 (deleted agent resurrected by gap repair)** — correct. `instance_delete` removed only the `db_agents` row, so the startup gap repair recreated the agent from its surviving `db_agent_definitions` row with launch state reset. It now clears both sides, symmetric with `agent_def_delete`. Pinned by a test that runs `repair_agent_def_gaps()` after the delete and asserts the agent stays gone.

Also dropped a dangling `SPEC_CONTINUATION_SESSION_PERSISTENCE_2026_05_23.md` pointer in `instance.rs` — the spec does not exist, and the citations gate covers touched files. Three other files still cite it; left for whoever touches them, to keep this diff scoped.

```
cargo test -p agentmux-srv   → 3151 passed, 1 failed
```
The failure is `backend::sysinfo::handle_anomaly_tests::refresh_processes_specifics_does_not_leak_a_handle_per_call`, a local-environment condition unrelated to this change (same code passed Windows CI on #3075; it fails identically here with the storage changes stashed).

GenericAgentX-asaf:

All five findings from round 1 (two reagent P1s, two codex P1s, one codex P2) are fixed in 4e54c811c, each with a test. Details in the comment above.

GenericAgentX-asaf:
Round 2 P1 fixed in 09f456177 — the finding is correct, and it names a real gap in my round-1 fix rather than a variation on it.

Round 1 closed the local resurrection path (`repair_agent_def_gaps` rebuilding the agent from its surviving `db_agent_definitions` row) and stopped there. But `agent_def_list` overlays every ACTIVE cross-channel definition record onto the local list on every read, so an agent deleted through `instance_delete` reappeared on the very next My Agents fetch — a shorter loop than the one I fixed, no restart required. `instance_delete` now writes the same global tombstone `agent_def_delete` does.

The observation about the test is the part worth calling out: it ran against a store with no definition registry attached, so the overlay did not exist and the test could not have failed on this. It now attaches a `DefinitionStore`, asserts the record is created on insert (so the sanity precondition is real rather than vacuous), and asserts both that the record is retired and that `agent_def_list` no longer returns the agent.

Scope note: the tombstone is written only when a local row was actually deleted. An agent that exists *only* in the global registry has no row for `instance_delete` to remove, and `agent_def_delete` is already the path that deletes one of those (returning `rows > 0 || global_retired` for exactly that reason). I did not widen `instance_delete`'s return semantics to match.

```
cargo test -p agentmux-srv   → 3152 passed, 0 failed
```
Clean this run — the `refresh_processes_specifics_does_not_leak_a_handle_per_call` failure I flagged on earlier rounds passed here too, which confirms it was environmental rather than anything in this change.

Noting for the record that an APPROVED review landed on 6f208ca14 52 seconds after the CHANGES_REQUESTED one on the same commit. I did not merge on it — the P1 was real (I confirmed the unconditional overlay in `agent_def_list` before fixing), so shipping on the later approval would have knowingly landed the defect.

PR #3081, opened 2026-09-07
docs(cef): spec the CEF 148→152 milestone upgrade across all three platforms

AgentX-asaf:
## Summary

Spec for upgrading the bundled CEF/Chromium runtime across Windows, macOS, and Linux. **Spec only — no code changes.**

We are **four Chromium milestones behind**: we pin CEF branch `7778` (Chromium 148), upstream's newest stable branch is `7977` (Chromium 152). Verified 2026-09-07 by reading `CHROMIUM_BUILD_COMPATIBILITY.txt` from each upstream branch.

## Two gaps, deliberately separated

| Gap | From → To | Cost |
|---|---|---|
| Patch-level (within our milestone) | `148.0.7778.180` → `148.0.7778.218` | Rebase + rebuild, no porting — security fixes only |
| **Milestone** | CEF `7778` → `7977` | The real work — this spec |

Conflating these is the main way this gets mis-estimated, so §1 splits them explicitly. The patch-level rebase is worth doing on its own if the milestone jump is deferred.

## What the spec covers

- **Patch inventory** (§2) — the four patches we carry, which are mandatory, and which may already be upstream. Re-checking these against 152 is Phase A precisely because *the answer can delete work*: several AgentMux patches were upstreamed between 146 and 148 and stopped being ours to carry.
- **Target decision** (§3) — recommends going straight to 152 rather than stepping through 149/150/151, because cost is dominated by rebuild (3–6 h × 3 platforms), not porting. States the fallback honestly.
- **Phased breakdown** (§4) — recon → fork source → Rust binding → per-platform builds → distribution → verification → rollback.
- **Resource cost** (§4 Phase D) — per platform: 3–6 h wall-clock, ~99 GB working tree, ≥120 GB free disk, ≥32 GB RAM. **9–18 machine-hours across three machines**, which must be three separate machines (no cross-building).
- **Parallelization** (§6) — Phase D is embarrassingly parallel across three agents, one per platform; everything before it is single-owner.

## Two findings worth acting on regardless of the upgrade

1. **The fork's release tags use three different version schemes** (§7.1) — Chromium version, CEF framework version, and CEF version, plus ad-hoc `-codecs`/`-2`/`-3` suffixes. A human resolves these by context; an automated drift checker cannot. This directly blocks the companion spec in `a5af/shared-infrastructure` (`docs/specs/SPEC_VERSION_DRIFT_REPORTER_2026_09_07.md`), which proposes machine-comparing our shipped CEF against upstream so this drift gets caught automatically next time.
2. **An archived root-cause doc contradicts its own tracking issue** (§7.3) — `ANALYSIS_WINDOWS_GPU_DISABLED_ROOTCAUSE_2026_06_11.md` blames the fork's libcef build for a Windows GPU failure; issue #1345's closing comment retracts that and points to #1354 (missing `supportedOS` manifest) as the real cause. The archived doc could mislead someone scoping this upgrade.

## Open questions for the repo owner (§8)

Notably: **is native window drag still required on Linux/Windows?** If it were dropped, two of the four patches stop needing forward-porting and this upgrade gets materially cheaper — the `patched-libcef` feature is already default-off with a working fallback.


a5af:
@codex review

PR #3082, opened 2026-09-07
docs(layout): spec the pane tab-switch chrome repaint, fix the stale key-function pointer

AgentY-asaf:
Spec + one documentation fix. **The refactor it describes is deliberately not implemented here** — see "Why not implemented" below, which is the part worth reviewing.

Spec: `docs/specs/SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md`

## The report

> when you move between tabs, the entire pane repaints, including the pane header. We don't want that, instead we want only the actual terminal to repaint, the pane header and tab bar should not flinch at all.

Accurate, and the current behavior is not incidental — the whole pane is *deliberately* hidden and re-revealed on every in-pane tab switch.

## Chain, traced end to end

1. `term.tsx:134` `handleTermTabSwitch` → `holdLeafRevealGate` → `setActiveBlockInStack` → `scheduleLeafRevealLift`
2. `layoutStack.ts:88` `model.nodeModels.delete(nodeId)` — evicts the cached node model
3. `layoutNodeModels.ts:142` `activeKeyFor` returns `` `${node.id}:${activeBlockId}` `` — the key **changes**
4. `tilelayout-shared.tsx:171`/`:354` render leaves in `<Key … by={activeKeyFor}>` — a changed key tears down the whole subtree
5. That subtree is `<Block>` (`tabcontent.tsx:42`) — **pane header included** — wrapping the view, which for a terminal contains `<PaneTabStrip>`
6. Because that rebuild looks bad, `TileLayout.core.tsx:553-559` hides the **entire tile**: `visibility: hidden`, `opacity: 0`, `transition: opacity 120ms`

So the flinch *is* the hide-and-fade. It isn't paint performance to be tuned.

## The actual defect

The remount boundary is drawn around the whole leaf when it only needs to wrap the view. Everything else — the gate, the fade, the whole-tile `visibility: hidden` — is compensation for that boundary being too wide.

The view remount is **correct and stays**: ViewModels have a one-instance-one-immutable-blockId contract. The header and tab strip are properties of the *leaf*, not the block, and shouldn't be inside the same boundary.

## Why not implemented

Because it is not the size the ask implies, and I would rather say so than hand over an unverified refactor of the app's most central component.

Every consumer in the `<Block>`/`BlockFrame` tree assumes *"this instance owns this blockId for its lifetime"*. Concretely: `useWaveObjectValue(oref: string)` takes a **plain string** and refcounts it in `onCleanup` — it does not re-subscribe when a blockId changes. The pane header alone reads `props.viewModel` for name, icon, text, favicon, pre-icon button, connection state, end-icon buttons, voice handle and settings-menu items, all per-block.

Keeping the chrome mounted across a switch means converting block identity to be reactive through that whole tree, which lands on **every** pane type at once (terminal, agent, editor, browser, media…), plus focus, menus, connection handling, magnify and tearoff.

The spec stages it in three independently revertible steps (hoist the strip to leaf chrome → reactive header + stop evicting → delete the gate), terminal first.

## Acceptance criteria are measurable, not visual

Because "doesn't flinch" must not be judged by watching it, the spec requires: capture the header and strip elements before a switch and assert they are the **same node instances** after (`el === elAfter`), and that `gatingNodeIds()` never contains the node during a switch. Those two are what would distinguish a real fix from a cosmetic one.

## What ships here

`layoutStack.ts` pointed at `TileLayout.{win32,linux,darwin}.tsx` for the key function; it moved to `tilelayout-shared.tsx` when the per-platform copies were folded into one core in #3041. Corrected — and the cost of leaf-wide keying is now recorded at that comment, so the next reader finds it in place rather than reconstructing it from four files.

Comment-only change; `tsc --noEmit` clean.

Related: #3079 (terminal tab-strip blur) touches the same strip. If Step 1 of this spec is ever built, it moves that PR's `.view-term > .pane-tab-strip` selector; noted in both specs.

### DIFF
--- a/.changesets/1788801294-docs-layout-spec-the-pane-tab-switch-chrome-repaint-fix-the--wdpg.md
+++ b/.changesets/1788801294-docs-layout-spec-the-pane-tab-switch-chrome-repaint-fix-the--wdpg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(layout): spec the pane tab-switch chrome repaint, fix the stale key-function pointer
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -402,6 +402,7 @@ partial list.
 | [`SPEC_PANE_MINIMIZE_COLUMN_DISSOLVE_2026_06_27`](SPEC_PANE_MINIMIZE_COLUMN_DISSOLVE_2026_06_27.md) | SPEC — Pane Minimize: Column Dissolve on Full-Column Collapse |
 | [`SPEC_PANE_MINIMIZE_REFINEMENTS_2026_06_24`](SPEC_PANE_MINIMIZE_REFINEMENTS_2026_06_24.md) | SPEC — Pane Minimize Refinements |
 | [`SPEC_PANE_OVERLAY_AUTO_CLIP_2026_05_11`](SPEC_PANE_OVERLAY_AUTO_CLIP_2026_05_11.md) | Auto-discovery pane-overlay clipping (declarative `data-pane-overlay`) |
+| [`SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07`](SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md) | SPEC: Switching in-pane tabs must not repaint the pane header or tab bar |
 | [`SPEC_PANE_TEAROFF_MOTHER_RESIZE_2026_06_20`](SPEC_PANE_TEAROFF_MOTHER_RESIZE_2026_06_20.md) | Pane Tear-Off — Mother Window Resize |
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentY-asaf:
**CI red on the doc gate is an ordering dependency, not a defect — it will clear without a code change.**

```
FAIL docs/specs/SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md
     Cites a spec that does not exist: docs/specs/SPEC_TERM_PANE_TAB_STRIP_TRAILING_BLUR_2026_09_07.md
```

That spec is real, it just lives on #3079 and hasn't landed on `main` yet. The citation gate is diff-scoped against `main`, so it correctly reports the pointer as dangling *from this branch's point of view*.

Deliberately **not** "fixed" by deleting the reference: the two specs describe the same strip and the ordering constraint between them is exactly the thing a future reader needs (Step 1 of this spec would move the selector #3079 adds). `docs/specs/README.md`'s own rule is that a pointer to nothing is worse than no pointer — but the answer here is to land them in order, not to drop a pointer that is about to resolve.

Plan: merge #3079, rebase this branch on `main`, CI goes green on its own.

PR #3083, opened 2026-09-07
Manoz@Area54: docs(saga): cross-reference the two saga/reducer frameworks in their module headers (Phase 5 decided)

GenericAgentX-asaf:
## Summary

Phase 5 (step 14) of `docs/reports/REPORT_DRY_AND_MODULARITY_AUDIT_2026_09_06.md`, now decided: **keep both saga/reducer frameworks, document the split.** Comments only — 26 comment lines across four module headers, no code moved.

## What the audit actually found

Both `agentmux-launcher` and `agentmux-srv` have a module called `saga` and one called `reducer` — ~11K lines between them. They are **not** two copies of anything: the two saga modules share exactly **one** method name (`new`) and no types.

| | launcher | srv |
|---|---|---|
| saga shape | a `Saga` **trait** driven by a bus subscription, for the cross-process case | plain **async functions**, in-process dispatch |
| consumers | one (`pool_respawn`) | every multi-step workspace command |
| reducer | window/process state machine | workspaces/tabs/blocks |

The divergence was a deliberate call, already recorded in srv's own header and `docs/retro/saga-coordinator-location-analysis-2026-04-30.md`: the Phase E plan assumed sagas would fan out across processes over IPC, the implementation kept that fan-out in the frontend, and an in-process call beats an IPC round-trip on every step.

So there was no duplication to remove. What existed was a **naming collision that reads like one abstraction that drifted** — which is exactly how a future contributor talks themselves into "unifying" two things that were split on purpose.

## What changed

Each of the four headers now states which framework it is, which one it isn't, and where the reasoning lives. The launcher's saga header additionally says plainly not to unify them without reading the decision first, because that is the tempting wrong move.

`DECISION_SAGA_REDUCER_TWO_FRAMEWORKS_2026_09_07.md` moves `proposed` → `active`; audit step 14 is closed.

## Why not the alternatives

- **One shared crate** would undo a considered decision in order to remove a duplication that amounts to one shared method name, and would force srv's sagas into a bus-driven coordinator they do not need.
- **Renaming srv's types** would churn the `SagaStarted`/`SagaFailed` event names — which are on the wire to the frontend — plus the Phase E specs and retros, for clarity this page provides for free.

Both stay open if a genuinely cross-process saga ever appears; the decision doc records when to revisit.

## Verification

- **Comments-only, checked mechanically:** every changed line in the four `.rs` files is a comment (`git diff` filtered for non-comment additions returns nothing).
- `cargo check -p agentmux-srv -p agentmux-launcher` — clean.
- `scripts/check-doc-status.sh` passes on the decision doc's new `active` status.


a5af:
@codex review

GenericAgentX-asaf:
All three fixed in `7c272cc0`.

**ReAgent P2** — real. `sagas/mod.rs` claimed the launcher framework 'still does nothing — no consumers,' contradicting this PR's own DECISION doc, which says Phase F.5 gave it `pool_respawn`. Corrected.

**Codex P2** — real, and worth stating precisely: I checked the actual imports rather than taking the finding on faith. Both `agentmux-launcher/src/reducer/mod.rs:50` and `agentmux-srv/src/reducer.rs:42` import the SAME `agentmux_common::ipc::{Command, Event}`. My cross-reference claimed they differ. Corrected in both headers and the decision doc — only `State` (each crate's own) and which variants each `update` handles actually differ.

**Codex P3** — real. The audit's own progress summary still said Phase 5 was 'awaiting the repo owner' a few lines above where step 14 is marked done — contradicting itself. Updated.

Re-verified comments-only (no non-comment `.rs` line changed) and `cargo check` on both crates.

PR #3084, opened 2026-09-07
Claude@gamerlove: fix(reactive): stop a same-host inject to an unspawned persistent agent from deadlocking the reactive handler

a5af:
## Summary

A same-host SendMessage to a persistent agent with no live controller (the
common case: any idle agent, or any agent after a srv restart before a human
sends it something from the UI) permanently deadlocks the reactive handler.

The reactive-delivery spawn fallback (bootstrap.rs) runs a respawn
synchronously on the injecting thread while inject_message still holds
ReactiveHandler.inner's non-reentrant std::sync::Mutex.
TurnRegistration::Skip guards the one re-lock at that call site, but the
spawn path's own auto-registration (persistent.rs:2801,
PersistentSubprocessController::spawn_process) is a second, independent
re-lock two modules away that Skip never covered.

Reproduced independently on two hosts running 0.55.37/0.55.38 (identical code
on main, unfixed). Once wedged: every reactive endpoint hangs forever
(inject, discovery, agents list, fleet, cloud subscriber, both periodic
watchers), and the next UI submit strands that connection's WebSocket task in
its worker's LIFO slot, freezing the UI while every health signal still
reads "up". Only a process restart clears it.

Full incident writeup: docs/incident/INCIDENT_2026_09_07_BACKEND_UPTIME_TIMER_FROZEN.md

## Fix

persistent.rs now calls a new ReactiveHandler::try_register_agent_with_nonce
(try_lock-based) instead of the blocking register_agent_with_nonce. On the
reentrant path this returns an error immediately instead of hanging; the
existing error handling at that call site already logs and continues, so
nothing else needed to change. Skipping is safe there specifically:
inject_message only reaches the spawn fallback after resolving the target
through agent_to_block, so the agent/block pair is registered by
construction before this call is ever reached.

## Test plan

- [x] Two new unit tests added (backend/reactive/tests.rs): the reentrant
      call must fail fast under a 5s timeout (reproduces the incident, a
      regression here hangs the test instead of failing it), and an ordinary
      non-reentrant call must still register normally.
- [ ] Not compiled or run in this environment: no Rust toolchain
      available. Every signature and lock type was checked by hand against
      the surrounding code, but cargo test -p agentmux-srv reactive:: must
      pass before merge.
- [ ] Manual repro: same-host SendMessage to an idle persistent agent, then
      submit from the UI to a different pane, should no longer freeze.

### DIFF
--- a/.changesets/1788799617-fix-reactive-stop-a-same-host-inject-to-an-unspawned-persist-i9s4.md
+++ b/.changesets/1788799617-fix-reactive-stop-a-same-host-inject-to-an-unspawned-persist-i9s4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(reactive): stop a same-host inject to an unspawned persistent agent from deadlocking the reactive handler
--- a/agentmux-srv/src/backend/blockcontroller/persistent.rs
+++ b/agentmux-srv/src/backend/blockcontroller/persistent.rs
@@ -2791,6 +2791,11 @@ impl PersistentSubprocessController {
         // See SPEC_MUXBUS_AGENT_DISCOVERY_AND_PERSISTENT_DELIVERY_2026_06_16.
         let agent_id_for_muxbus = muxbus_agent_id_from_env(&config.env_vars);
         *self.agent_id.lock().unwrap() = agent_id_for_muxbus.clone();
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

PR #3085, opened 2026-09-07
fix(ci): centralize and cross-check the three pinned CEF runtime tags

GenericAgentX-asaf:
## This PR was reworked after #3086 landed

Original scope: `release.yml` passed `cef-runtime-tag: ""` to all three
platform build jobs, and each independently resolved "latest" at build time
— so two builds of the same commit (or the same tagged release, rebuilt
later) could silently ship different Chromium versions.

**#3086 (a different agent) landed on `main` first and fixed the
reproducibility half of this properly**: it pins each platform's tag to a
literal string instead of resolving "latest" dynamically. That's a better
fix for reproducibility than this PR's original dynamic-resolution approach
— a rebuild of the same tagged release now always resolves to the same CEF
version, not just "the three platforms agreed with each other for this one
run." Rather than merge a competing design on top (which would've silently
undone #3086's fix), this PR is rebased onto it and reworked to close the
one gap it left open.

## The gap #3086 left open

Its three pins live in three separate places — one literal string per
`build-*.yml` call. A human bumping one platform's pin (e.g. Windows to CEF
152) and forgetting the other two reproduces the exact failure #3086 exists
to prevent — via a stale manual edit instead of a live "latest" race.
Nothing catches that today.

## The fix now

Adds a `cef-runtime-pins` job that is the single place all three tags are
declared — `build-windows`/`build-linux`/`build-macos` reference its
outputs, not their own literal string — and fails the release if the three
don't agree on Chromium **milestone** before any build starts. Milestone,
not full version equality: the three platforms don't even share a
version-string format (Windows/Linux embed Chromium's own
`major.minor.branch.patch`; macOS uses an unrelated local build-counter
scheme — `148.23.23` isn't a Chromium build number). All three do keep the
milestone as the leading dot-separated segment, so that's the one component
a comparison can mean something for.

No GitHub API calls, no secrets — this only parses the three literal
strings declared in the job itself, so it's an offline, always-reproducible
check, not a live "resolve latest" query that would silently reintroduce
#3086's bug.

Also fixes `publish`'s `needs` list to include `cef-runtime-pins` directly:
`if: !failure() && !cancelled()` only inspects jobs *directly* listed in
`needs` (GitHub Actions' documented behavior for an `if:` that references
those functions), so without this, a milestone mismatch would cascade to
skip every build job — invisible to `failure()` — while `publish` ran
anyway, downloaded zero artifacts, and failed confusingly later instead of
being cleanly blocked.

## Verification

- Ran the milestone check against #3086's actual committed pins (windows
  `148.0.7778.180`, linux `148.0.7778.180-codecs`, macos `148.23.23-codecs`)
  — all resolve to milestone 148, passes.
- Ran the mismatch case (windows bumped to 152, linux/macos left at 148) —
  correctly fails with a clear error naming which platform disagrees.
- `.github/workflows/release.yml` parses as valid YAML; full `needs:` graph
  checked by hand — acyclic, `cef-runtime-pins` has no dependency on
  `check-macos` (unlike this PR's earlier, now-superseded commits — no live
  API call means no macOS-unavailable resolution failure to guard against
  anymore, so that gating isn't needed).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,6 +71,80 @@ jobs:
           echo "✓ Version $VER consistent across all locations"
           echo "version=$VER" >> "$GITHUB_OUTPUT"
 
+  # ── Phase 1b: Single source of truth for the three pinned CEF runtime tags ──
+  # PR #3086 pinned each build-*.yml call's cef-runtime-tag to a literal
+  # string, fixing release-to-release reproducibility: a blank tag used to
+  # resolve to "latest published release for that platform" at build time, so
+  # two builds of the identical commit — or the same tagged release rebuilt
+  # later — could silently ship different Chromium versions. That fix has one
+  # gap: the three pins lived in three separate places (one per build-* call
+  # below), so a human bumping one platform's pin (e.g. Windows to CEF 152)
+  # and forgetting the other two reproduces the exact failure #3086 exists to
+  # prevent — via a stale manual edit instead of a live "latest" race. Nothing
+  # caught that.
+  #
+  # This job is now the ONE place the three pins are declared — build-windows/
+  # linux/macos below reference its outputs, not a literal string of their
+  # own — and it fails the release if they don't agree on Chromium MILESTONE
+  # before any build starts. Milestone, not full version equality: the three
+  # platforms don't even share a version-string format. Confirmed against the
+  # tags PR #3086 pinned: windows=148.0.7778.180, linux=148.0.7778.180-codecs
+  # (both Chromium's own major.minor.branch.patch), macos=148.23.23-codecs (an
+  # unrelated local build-counter scheme — 148.23.23 is not a Chromium build
+  # number). All three do keep the milestone as the leading dot-separated
+  # segment regardless of scheme, so that's the one component a comparison
+  # can mean something for.
+  #
+  # No GitHub API calls, no secrets: this only parses the three literal
+  # strings declared right here, so it's a fast, offline, always-reproducible
+  # check — not a live "resolve latest" query, which would silently undo
+  # #3086's reproducibility fix.
+  #
+  # To bump a pin: update the matching value below (this is now the only
+  # place to do it) in the same PR as the consumer-side CEF upgrade change,
+  # exactly as #3086's original guidance said — just centralized to one site
+  # instead of three, so there's no longer a second and third place to forget.
+  cef-runtime-pins:
+    name: CEF runtime tags (single source of truth + milestone check)
+    runs-on: ubuntu-latest
+    outputs:
+      windows-tag: ${{ steps.pins.outputs.windows-tag }}
+      linux-tag: ${{ steps.pins.outputs.linux-tag }}
+      macos-tag: ${{ steps.pins.outputs.macos-tag }}
+    steps:
+      - name: Declare pins and check milestone agreement
+        id: pins
+        run: |
+          WIN_TAG="cef-windows-x86_64-148.0.7778.180"
+          LINUX_TAG="cef-linux-x86_64-148.0.7778.180-codecs"
+          MACOS_TAG="cef-macos-arm64-148.23.23-codecs"
+
+          # Chromium milestone = the leading dot-separated numeric segment
+          # after the cef-<platform>-<arch>- prefix. Works across both
+          # version schemes above because both keep the milestone number
+          # first — it's the one component the three schemes agree on the
+          # meaning of.
+          milestone() {
+            echo "$1" | sed -E 's/^cef-[a-z0-9_]+-[a-z0-9_]+-//' | cut -d. -f1
+          }
+          WIN_MS=$(milestone "$WIN_TAG")
+          LINUX_MS=$(milestone "$LINUX_TAG")
+          MACOS_MS=$(milestone "$MACOS_TAG")
+          echo "Pinned CEF runtime tags: windows=$WIN_TAG (m$WIN_MS) linux=$LINUX_TAG (m$LINUX_MS) macos=$MACOS_TAG (m$MACOS_MS)"
+
+          if [ "$WIN_MS" != "$LINUX_MS" ] || [ "$WIN_MS" != "$MACOS_MS" ]; then
+            echo "ERROR: CEF Chromium milestone mismatch across the pinned tags — refusing to build a release from mismatched runtimes."
+            echo "  windows: $WIN_TAG (milestone $WIN_MS)"
+            echo "  linux:   $LINUX_TAG (milestone $LINUX_MS)"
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

a5af:
## Addressed: Codex P1 (mismatched Chromium generations across platforms)

Fair finding, and correct: resolving all three tags in one job only narrowed the *scheduling* race — it didn't prove the three tags are the same Chromium generation. The three `cef-<platform>-<arch>-*` series in `agentmuxai/cef` are cut by independent build pipelines with no shared publish gate, so "latest" can point at mismatched generations with zero race involved at all (exactly the "Windows on 152, Linux/macOS still on 148" scenario raised).

Pushed a fix (`5f8...` — latest commit): the resolve job now extracts each tag's Chromium milestone and hard-fails if the three don't agree, instead of just handing mismatched tags downstream.

One thing this surfaced that's worth calling out on its own: the three platforms don't even share a version-string *format*. Checked the live repo —

```
cef-windows-x86_64-148.0.7778.180
cef-linux-x86_64-148.0.7778.180-codecs
cef-macos-arm64-148.23.23-codecs
```

Windows/Linux embed Chromium's real `major.minor.branch.patch`; macOS uses an unrelated local build-counter scheme (`148.23.23` isn't a Chromium build number). So a full version-equality check isn't even meaningful across all three — the milestone (leading dot-separated segment, `148`) is the one component all three schemes agree on the meaning of, and it's the exact granularity the mismatch scenario in the review needs. I didn't try to solve full cross-platform version normalization, since it isn't solvable with the current macOS tag scheme — a milestone-level gate is what's actually enforceable here.

Ran the full resolve → extract → compare logic against the live `agentmuxai/cef` repo end-to-end (not just the regex in isolation): passes today, all three currently at milestone 148.

@codex review

a5af:
## Addressed: ReAgent P1 (macOS unavailability shouldn't block Windows/Linux)

Correct — `resolve-cef-runtime` needed only `verify`, so it resolved and hard-failed on the macOS tag/milestone unconditionally. Since `build-windows`/`build-linux` now depend on it, a missing or mismatched macOS CEF release (or simply `check-macos.outputs.available == 'false'`, no Apple signing secrets) would fail the whole release, including the two platforms that would otherwise build fine — a regression from the pre-PR behavior where macOS unavailability was already handled gracefully via `build-macos`'s own `if:`.

Fixed: `resolve-cef-runtime` now also `needs: check-macos`, and skips macOS tag resolution/its milestone check entirely when unavailable — windows/linux still resolve and cross-check against each other regardless.

Verified end-to-end against the live `agentmuxai/cef` repo for both branches:
- `MACOS_AVAILABLE=true`: resolves and validates all three, matches previous behavior
- `MACOS_AVAILABLE=false`: skips the macOS lookup entirely, resolves+validates windows/linux only, no failure

a5af:
## Codex re-review (5136949187, against 33ecac1) — status

**P2 "Exclude skipped macOS builds from the milestone gate"** — already fixed one commit earlier (`15e116c`, pushed before this review ran, so it reviewed `33ecac1` rather than current HEAD). `resolve-cef-runtime` now `needs: check-macos` and skips macOS tag resolution/its milestone check entirely when unavailable. Verified in an earlier comment on this thread.

**P2 "Provide the pinning inputs before recommending them"** — real, fixed now. The mismatch error told operators to "pin all three cef-runtime-tag inputs... via workflow_dispatch," but `release.yml`'s own `workflow_dispatch` only exposes `tag` — the per-platform override only exists on the standalone `build-*.yml` workflows, whose runs don't go through this workflow's `publish` job. Replaced the remediation with the one that's actually followable: cut a matching release for the lagging platform in `agentmuxai/cef`, then re-run this release. Deliberately didn't add new release-level override inputs to fix this — that's a real feature, but more scope than a CI-consistency fix should carry.

**The P1 in this thread ("Resolve one coherent CEF release set")** is not from this review — checked `pull_request_review_id` on each inline comment via the API: it belongs to the *first* review (`5135186340`, before the milestone-mismatch gate existed), just surfaced by the flat comments endpoint alongside the two genuinely new P2s. Not a live finding against current code.

@codex review

a5af:
## Addressed: ReAgent P1 (publish doesn't see resolve-cef-runtime's failure)

Real, and worth being precise about why. Checked GitHub Actions' actual `if:` semantics rather than assuming: **an `if:` condition that references `success()`/`failure()`/`cancelled()`/`always()` anywhere replaces the implicit "all needs succeeded" check entirely** — GitHub only auto-ANDs `success()` onto a job's condition when the condition does *not* reference one of those functions.

`publish`'s `if: ${{ !failure() && !cancelled() }}` does reference them, so it gets no implicit check — it only inspects jobs *directly* in its own `needs:`, and `resolve-cef-runtime` wasn't in that list. (For contrast, `build-macos`'s `if: needs.check-macos.outputs.available == 'true'` references no status function, so GitHub *does* still implicitly require all of build-macos's needs — including `resolve-cef-runtime` — to have succeeded. That one was never actually at risk, only `publish` was.)

So the failure mode is exactly as described: `resolve-cef-runtime` fails on a milestone mismatch → `build-windows`/`build-linux`/`build-macos` all get skipped as a consequence (their default `if: success()` can't be satisfied) → a skipped need doesn't trip `failure()` → `publish`'s condition stays true → it runs anyway, downloads zero artifacts, fails confusingly in "Create GitHub Release" instead of being cleanly blocked at the real failure point.

Fixed: added `resolve-cef-runtime` to `publish`'s `needs:` list, and documented the actual if-condition semantics inline so this doesn't get silently reintroduced.

a5af:
## Addressed: ReAgent P1 (dormant macOS pin shouldn't block Windows/Linux) — and a correction

You're right that this wasn't present at `43fd1e3` despite my PR body claiming it carried forward from the earlier design — thanks for verifying against `pull/3085/head` rather than taking the claim at face value. The reasoning I gave for dropping it was wrong: I was thinking about *network-failure* risk (genuinely gone once resolution became static strings) and missed the separate risk that a dormant, unused pin can still be *stale* — cross-checking against a value that ships nothing this run is a pure false positive with no safety benefit.

Fixed: `cef-runtime-pins` now `needs: check-macos` and only cross-checks the macOS milestone when `check-macos.outputs.available == 'true'`. Windows/Linux are still cross-checked unconditionally (both always build). The check is deferred, not lost — the moment macOS becomes available again, the same stale pin trips the same failure.

Ran the actual shell logic for all four scenarios, not just reasoned about them:
- baseline (all m148, macOS available) → passes
- genuine Windows/Linux mismatch (macOS available) → fails
- Windows+Linux bumped together to m152, macOS pin stale at m148, macOS **unavailable** → passes (dormant, nothing ships from it)
- same as above but macOS becomes available → fails (deferred check fires correctly)

PR #3086, opened 2026-09-08
fix(ci): pin cef-runtime-tag explicitly in release.yml instead of floating to latest

AgentX-asaf:
## Summary

`release.yml` passes `cef-runtime-tag: ""` to all three `build-{windows,linux,macos}.yml` call sites. Every callee resolves a blank tag to "newest published release for that platform on `agentmuxai/cef`" — so a release build's CEF version is decided by **release timing**, not by anything in the commit. Two builds of the identical commit could ship different Chromium versions.

This also makes the CEF-upgrade spec's rollback story (`docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md` §4G) not actually work: reverting a consumer PR back to the 148 pins wouldn't restore CEF 148, because the next release would still resolve to whatever is newest on the fork.

Split out on its own rather than folded into the upgrade spec's Phase E — this is a reproducibility bug independent of whether that upgrade ever ships (credit to Korp for the framing).

## Change

Pin all three `cef-runtime-tag` inputs in `release.yml` to today's actually-resolved tags:

| Platform | Pinned tag |
|---|---|
| Windows | `cef-windows-x86_64-148.0.7778.180` |
| Linux | `cef-linux-x86_64-148.0.7778.180-codecs` |
| macOS | `cef-macos-arm64-148.23.23-codecs` |

Each was verified against that platform's own resolver logic in `build-{windows,linux,macos}.yml` (filter by tag prefix, sort by `publishedAt`, take newest) — **this changes nothing about what currently ships**, it only stops it from being able to silently drift between release runs.

Added a comment explaining why these are pinned and how to bump them (cut the new `agentmuxai/cef` release first, then update the tag here in the same PR as the consumer-side change — never blank them back out).

## Verification

- `python3 -c "import yaml; yaml.safe_load(...)"` — valid YAML.
- Each pinned tag confirmed to currently be exactly what `TAG=""` resolves to today, via the same platform-prefix-filter + `publishedAt`-sort query each `build-*.yml` runs.

### DIFF
--- a/.changesets/1788837169-fix-ci-pin-cef-runtime-tag-explicitly-in-release-yml-instead-nkvi.md
+++ b/.changesets/1788837169-fix-ci-pin-cef-runtime-tag-explicitly-in-release-yml-instead-nkvi.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): pin cef-runtime-tag explicitly in release.yml instead of floating to latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,6 +80,20 @@ jobs:
   # steps (windows-portable/windows-installer/windows-msix) are what
   # `publish` downloads below, exactly as when this was inlined.
   # See docs/specs/SPEC_CEF_PROPRIETARY_CODECS_ALL_PLATFORMS_2026_07_26.md.
+  #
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #3087, opened 2026-09-08
fix(ci): pin the CEF runtime tag for release builds instead of resolving latest at build time

AgentY-asaf:
Follow-up from the finding in the CEF milestone-upgrade spec (#3081, from a peer agent's recon): `release.yml` passes `cef-runtime-tag: ""` to all three platform build jobs, and blank resolves to "most recently published `agentmuxai/cef` release for this platform" — evaluated **fresh at every workflow run**.

## Why this is a real reproducibility gap, not a style nit

That default is correct for the standalone `build-*.yml` workflows — a manual or CI-triggered test build wanting "whatever's newest" is reasonable there. But in the **release** pipeline specifically, it means a release artifact's Chromium runtime depends on *when the workflow happened to run*, not on anything in the commit being released. Concretely: re-running the release workflow for the exact same git tag after a new `agentmuxai/cef` release lands would silently ship a different CEF/Chromium build under the same AgentMux version number — two builds of the same commit, different runtimes.

## The fix

Pin `cef-runtime-tag` explicitly for all three platforms in `release.yml`, to the current, verified-working trio per `docs/specs/SPEC_CEF_PROPRIETARY_CODECS_ALL_PLATFORMS_2026_07_26.md` ("H.264/AAC verified on all three platforms. Verified 2026-08-10."):

| Platform | Pinned tag |
|---|---|
| Windows | `cef-windows-x86_64-148.0.7778.180` |
| Linux | `cef-linux-x86_64-148.0.7778.180-codecs` |
| macOS | `cef-macos-arm64-148.23.23-codecs` |

Each was confirmed to actually exist in `agentmuxai/cef` via `gh release view` before committing (not assumed from `gh release list`'s publish-date sort alone) — all three are published, non-draft, non-prerelease.

A comment above `build-windows` explains the why (this file differs from the `build-*.yml` default on purpose) and the how — bump all three together, and only after re-verifying the new set the same way, when qualifying a new `agentmuxai/cef` release. Named the obvious next occasion: the 148→152 milestone jump spec'd in #3081.

## Verification

- YAML re-parses cleanly (`python -c "import yaml; yaml.safe_load(...)"`).
- `check-spec-citations.sh` / `check-doc-status.sh` pass.
- Diff is else-unchanged — no behavior touched besides the three tag values.
- **Not yet verified against a real release run** (this workflow only triggers on a version tag push or manual dispatch, and I'm not cutting a release to test a CI-pinning change). The `build-*.yml` resolver steps this reuses are already exercised by the ordinary CI-triggered/manual builds, so the resolution mechanism itself is proven; what's new here is only which three literal strings get passed in.

### DIFF
--- a/.changesets/1788837242-fix-ci-pin-the-cef-runtime-tag-for-release-builds-instead-of-gism.md
+++ b/.changesets/1788837242-fix-ci-pin-the-cef-runtime-tag-for-release-builds-instead-of-gism.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): pin the CEF runtime tag for release builds instead of resolving latest at build time
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,6 +80,23 @@ jobs:
   # steps (windows-portable/windows-installer/windows-msix) are what
   # `publish` downloads below, exactly as when this was inlined.
   # See docs/specs/SPEC_CEF_PROPRIETARY_CODECS_ALL_PLATFORMS_2026_07_26.md.
+  # ── CEF runtime pinning for release builds ────────────────────────────────
+  # `cef-runtime-tag` is left blank in the standalone build-*.yml workflows on
+  # purpose (manual/CI-triggered test builds want "whatever's newest"), but a
+  # blank value there resolves to "most recently PUBLISHED agentmuxai/cef
+  # release for this platform" (see each build-*.yml's "Resolve codec-enabled
+  # CEF runtime tag" step) — evaluated fresh at EVERY run. In this file, that
+  # means two release runs of the exact same git tag could pull different
+  # Chromium runtimes if a new agentmuxai/cef release landed between them:
+  # a release artifact's CEF version would depend on build TIMING, not on
+  # anything in the commit being released. Pinned explicitly below instead.
+  #
+  # This trio is the current, verified-working set per
+  # docs/specs/SPEC_CEF_PROPRIETARY_CODECS_ALL_PLATFORMS_2026_07_26.md
+  # ("H.264/AAC verified on all three platforms. Verified 2026-08-10.").
+  # Bump all three together — and only after verifying the new set the same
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentY-asaf:
**Codex P2 — "Bind CEF pins to the requested release tag": fixed in \`657fc8f88\`.** Correct and worth catching — I'd only closed the gap for the push-tag trigger.

The mechanism: `workflow_dispatch`'s `tag` input only controls what *source* gets checked out inside each job (the `ref:` lines). It does not change which copy of `release.yml` itself runs — that's always resolved from whatever ref the dispatch happened FROM (`github.ref_name`), which defaults to `main` for a manual dispatch. So an emergency rebuild of an old tag, dispatched from `main`, would rebuild that release's *source* with *today's* CEF pins from `main` — the exact "same tag, different runtime depending on timing" problem, just relocated from "any time" to "any time someone dispatches from the wrong ref" (which, for an emergency rebuild, is the likely default — nobody thinks to pass `--ref <tag>`).

Went with your second suggested option — require the dispatch ref to match the release tag — over storing the pins in the checked-out source: that would need a new file read out of the checkout in all three `build-*.yml` workflows too, for a fix that's supposed to be about `release.yml` specifically. The guard is a single new step, first in the `verify` job, gated on `workflow_dispatch` only (push-tag runs skip it — `github.ref_name` already equals the pushed tag there, and there's no `tag` input to compare against). It fails loudly with the exact `gh workflow run --ref <tag> ...` invocation needed to satisfy it, rather than silently doing the wrong thing.

AgentY-asaf:
Closing as superseded — **#3086** landed the exact same pin independently while this was in review (another agent, working the same reproducibility gap found while scoping the CEF 148→152 milestone upgrade). Same three tags, already on `main`. No need to merge a duplicate.

Codex's finding on this PR (the `workflow_dispatch` ref gap) was real and not covered by #3086 — carried forward as its own, correctly-scoped PR: **#3089**.

PR #3088, opened 2026-09-08
Korp@claudius: refactor(storage): re-point six agent-child tables' FK from db_agent_definitions to db_agents (agent consolidation Phase 3c, PR 3 of 4)

GenericAgentX-asaf:
## Summary

PR 3 of 4 continuing the agent-concept consolidation (`docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md` Phase 3c). Builds on #3080 (instance API reads/writes `db_agents` only).

**Six tables' `agent_id` foreign key now targets `db_agents(id)` instead of `db_agent_definitions(id)`**: `db_agent_content`, `db_agent_skills`, `db_agent_history`, `db_agent_identity_links`, `db_agent_skills_ref`, `db_agent_mcp_ref`.

### Why

Until now, `agent_id` in these six tables could only ever be a `db_agent_definitions` id — a template-launched agent (a `db_agents` row with no definition row of its own) had no way to hold content overrides, skills, history, or an identity link of its own. That FK is exactly why PR 2's `listrecentsessions` had to fall back to the launch's TEMPLATE for identity-link lookups instead of the agent's own row.

Every existing `db_agent_definitions` row already has a `db_agents` mirror at the same id (the dual-write PR 1 established and PR 2 kept current), so re-pointing the FK does not orphan a single existing row — every `agent_id` value already in these tables satisfies the new target. What it unlocks is new rows keyed by a launch id that has no `db_agent_definitions` counterpart at all.

### How

Schema v30. SQLite has no in-place FK redefinition, so `m0028_agent_child_tables_repoint_fk` rebuilds each of the six tables on an existing store: rename, recreate with the new FK, copy (via `PRAGMA table_info` so it's not order-dependent), drop the renamed original, recreate the two secondary indexes, `PRAGMA foreign_key_check` before committing rather than trusting the swap silently. Runs `repair_def_gaps` first (same gap `m0025` backfills) so no legitimately-mirrored definition's content is mistaken for an orphan. Rows whose `agent_id` matches neither a definition nor a `db_agents` row (a pre-existing inconsistency) are dropped and the count logged.

A fresh install gets the new FK directly from the updated `CREATE TABLE` statements in `run_object_schema` — nothing for the migration to do there.

### Scoped narrower than the doc's original PR-3 sketch

This does **not** flip `db_agent_definitions` reads/writes, and does **not** remove the identity-link template fallback PR 2 added to `listrecentsessions`. Nothing writes an identity link against a launch id today — only the Agent Setup modal writes links, and only for templates/user agents — so removing the fallback now, with no writer ever targeting a launch's own row, would be a regression dressed as a cleanup. That follower work (giving a launch its own identity override, then dropping the fallback) moves into PR 4, alongside the definition flip and the legacy-table drop. `docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md` is updated to reflect this.

### Tests

`m0028`'s own suite: every table ends up targeting `db_agents` with data intact and cascade-delete still working; a bare launch-only row can hold its own identity link post-repoint (the actual unblock); an orphaned row is dropped and counted; gap-repair prevents a legitimately-unmirrored definition's content from being misclassified as an orphan; a fresh schema is a no-op. Two `phase5_tests.rs` fixtures needed a `PRAGMA foreign_keys=OFF` around their legacy-shaped `db_agent_content` insert, since they deliberately simulate a pre-`0007` store where `db_agents` hasn't been populated yet.

```
cargo test -p agentmux-srv   → 3159 passed, 1 failed (local-only)
```
The failure is `backend::sysinfo::handle_anomaly_tests::refresh_processes_specifics_does_not_leak_a_handle_per_call`, flaky on this machine across the last several PRs in this series — passed on Windows CI for #3075 and #3080 with the same code.


a5af:
@codex review

GenericAgentX-asaf:
Codex's P2 fixed in 975bd7295 — correct finding, and it's a real violation of a documented invariant, not a style nit.

`repoint()` called `agents_consolidate::repair_def_gaps` live, which breaks the "migrations freeze copies of live logic" rule in `migrations/mod.rs`: a migration must produce the same rows on every machine it ever runs on, and cannot call a helper whose behavior a later release may change. It's doubly true here since PR 4 schedules `agents_consolidate.rs` itself for removal — that would have stranded this call the moment PR 4 landed, forcing either a block on the cleanup or an edit to an already-released migration.

`frozen_repair_def_gaps` is now a verbatim inline copy of the v30 behavior, with the same "frozen copy, do not deduplicate against live code" framing `m0025`/`m0026`/`m0027` already use for their own frozen rules.

```
cargo test -p agentmux-srv   → 3159 passed, 0 failed
```
Clean this run, including the previously-flaky `refresh_processes_specifics_does_not_leak_a_handle_per_call`.

GenericAgentX-asaf:
All three P2s fixed in 028698ca4 — correct, all three comments were left stale by this same PR's own FK change:

- `session.rs`'s fallback comment claimed "Phase 3c re-points that FK; this fallback goes with it" — but this PR (Phase 3c) explicitly keeps the fallback, per the PR description's own scoping note. Rewritten to say why it stays: the FK now permits a launch to hold its own identity link, but nothing writes one there yet (only the Agent Setup modal writes links, only against a template/user agent's own definition id), so removing the fallback today would make a launched agent's identity silently vanish rather than clean anything up.
- `migrations.rs`'s "unlike the agent-level ref tables' FK to db_agent_definitions" was describing the exact tables this PR retargeted, 13 lines above its own claim.
- `app_api/mod.rs`'s doc comment quoted the literal old FK clause. Updated to the current target, noting a definition id still always satisfies it (every definition has a `db_agents` mirror at the same id), since `resolve_agent_definition_id`'s actual behavior is unchanged — only the FK it was warning about moved.

```
cargo test -p agentmux-srv   → 3159 passed, 0 failed
```

GenericAgentX-asaf:

PR #3089, opened 2026-09-08
fix(ci): require an emergency release rebuild to dispatch from the tag it is rebuilding

AgentY-asaf:
Follow-up from #3087. That PR pinned `release.yml`'s `cef-runtime-tag` for all three platforms — but while it was in review, **#3086** landed the exact same pin independently (another agent, working the same gap from the CEF-upgrade spec scoping). Since the pin itself is already on `main`, closing #3087 as superseded rather than merging a duplicate.

**This PR is the one piece of #3087 that #3086 doesn't have**: Codex's review of #3087 caught a real gap in the pin itself, and I'm carrying that forward here rather than letting it get lost with the superseded PR.

## The gap

A `workflow_dispatch` run's own workflow *definition* — including the `cef-runtime-tag` pins #3086 added — is read from whatever ref GitHub resolved the run **from** (`github.ref_name`), which for a manual dispatch defaults to `main`. The `tag` input only controls what *source* gets checked out inside each job; it never changes which copy of `release.yml` actually runs.

So an emergency `workflow_dispatch` rebuild of an old tag, launched from `main` (the likely default — nobody thinks to pass `--ref <tag>` for an emergency rebuild), would rebuild that release's source with whatever CEF pins are on `main` **today**. That's the exact "same tag, different Chromium runtime depending on when you happen to run it" gap #3086 closed for the push-tag trigger — just relocated to the dispatch path.

## The fix

A guard as the first step of the `verify` job: for `workflow_dispatch` specifically, fail loudly unless `github.ref_name` (the ref this run's own workflow definition comes from) equals the `tag` input (what's being rebuilt). The error names the exact `gh workflow run --ref <tag> ...` invocation that satisfies it.

Push-tag runs are unaffected — the guard's own `if` is gated on `event_name == 'workflow_dispatch'`, so it never evaluates for that trigger; `github.ref_name` already equals the pushed tag there regardless.

## Verification

- Cherry-picked cleanly onto current `main` (post-#3086) — no conflict, since #3086 never touched the `verify` job.
- YAML re-parses (`python -c "import yaml; yaml.safe_load(...)"`).
- `check-doc-status.sh` / `check-spec-citations.sh` pass.
- **Not exercised against a real dispatch** — same caveat as #3086: this workflow only triggers on a version-tag push or manual dispatch, and I'm not cutting a release to test a guard. The logic is a plain shell string comparison with no external dependency, so the risk surface is small, but it's worth someone actually running an intentionally-mismatched dispatch once to see the guard fire before the first *real* emergency rebuild needs it.

### DIFF
--- a/.changesets/1788838567-fix-ci-require-an-emergency-release-rebuild-to-dispatch-from-c8wu.md
+++ b/.changesets/1788838567-fix-ci-require-an-emergency-release-rebuild-to-dispatch-from-c8wu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): require an emergency release rebuild to dispatch from the tag it is rebuilding
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,6 +38,36 @@ jobs:
     outputs:
       version: ${{ steps.check.outputs.version }}
     steps:
+      # Guards the CEF pinning fix below, per codex's review of that change
+      # (PR #3087): a workflow_dispatch run's OWN DEFINITION — including the
+      # cef-runtime-tag literals a few jobs down — is read from whatever ref
+      # GitHub resolved this run FROM (github.ref_name), which for a manual
+      # dispatch is normally main, the default branch. `inputs.tag` only
+      # controls what SOURCE gets checked out inside each job (the `ref:`
+      # lines below and in build-windows/linux/macos.yml) — it does NOT
+      # change which copy of this workflow file runs. So an emergency
+      # workflow_dispatch rebuild of an OLD tag, launched from main, would
+      # silently rebuild that release's SOURCE with WHATEVER CEF pins are
+      # on main today — reintroducing the exact "same tag, different
+      # Chromium runtime depending on when you happened to run it" gap the
+      # pinning fix exists to close, just moved from "any time" to "any time
+      # you forget to dispatch from the right ref."
+      #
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentY-asaf:
**ReAgent P2 — hardened raw \`inputs.tag\` interpolation: fixed in \`b65e2986f\`.** Moved to a single \`env: TAG: \${{ github.event.inputs.tag }}\` and referenced \`$TAG\` everywhere the guard step previously interpolated it raw (comparison, the \`::error::\` message, and twice in the suggested \`gh workflow run\` command).

AgentY-asaf:
**Codex P2 (round 2) / ReAgent P2 — \`github.ref_name\` also raw-interpolated: fixed in \`dde8c689c\`.** Both flagged the same remaining gap — fixing \`TAG\`'s interpolation left \`github.ref_name\` untouched in the same comparison and error message, and I'd wrongly treated it as inert because it's "GitHub-controlled." It isn't: \`git-check-ref-format\` doesn't exclude shell metacharacters — backticks, \`$()\`, quotes, and \`;\` are all valid in a ref/tag name — so raw interpolation of a maliciously-named ref was a real injection path into this step.

\`REF_NAME\` now goes through the same \`env:\` block as \`TAG\`, referenced only as a quoted shell variable. Re-scanned the whole step this time rather than fixing reactively again — the only \`${{ }}\` left are the two \`env:\` assignments themselves, which is exactly where interpolation is supposed to happen.

PR #3090, opened 2026-09-08
Clare@starpower: feat(zoom): Ctrl+Shift+Scroll zooms every pane in the window at once

AgentO-asaf:
## What

A second wheel gesture alongside the existing Ctrl+Scroll: holding **Shift too** steps every pane in the current window's zoom together, each relative to its own current level — instead of just the one pane under the cursor. Over chrome (title bar/status bar/pane header) it behaves exactly like plain Ctrl+Scroll there: one chrome-zoom step, not an all-panes zoom.

## Why / design

Spec: [`docs/specs/SPEC_CTRL_SHIFT_SCROLL_ZOOM_ALL_PANES_2026_09_07.md`](../blob/main/docs/specs/SPEC_CTRL_SHIFT_SCROLL_ZOOM_ALL_PANES_2026_09_07.md). No new zoom system — `AppAllPanesZoomHandler` (`app.tsx`) reuses the same chrome hit-test `AppZoomHandler` already has, and `zoom.ts`'s new `zoomAllPanesIn`/`zoomAllPanesOut` loop the existing per-block zoom stepping over every pane in the window's registry. `getBlockZoom`'s existing viewType allowlist does all the filtering, so a browser or warden pane is silently skipped exactly as it already is for a single Ctrl+Scroll today — **zero new filtering logic**.

`getAllBlockComponentModelEntries()` is a small new export from `block-component-registry.ts` — needed because `ViewModel`'s base interface doesn't declare `blockId` (each concrete view model carries it as its own implementation detail). The registry's map is already keyed on it; this just stops discarding that key.

`stepZoom`/`setBlockZoom` gained an optional `showIndicator` flag (default `true`, every existing caller unaffected) so the all-panes stepper shows **one** summary toast for the whole gesture instead of N per-pane toasts each overwriting the last before it can be read.

## A bug I found while implementing, not just designing

The spec (written before this PR) flagged that the six views with their own duplicated Ctrl+Wheel handlers (term, editor, armory, warden, swarm, AgentShellSubblock) needed a `shiftKey` guard each, or they'd double-step the hovered pane once this gesture existed. It missed that **`AppZoomHandler`'s own base gate needed the identical fix** — it's a separate `window` "wheel" listener with no `shiftKey` check either, so without excluding Shift there too, it would *also* fire on Ctrl+Shift+Scroll and double-step whichever single pane the cursor happened to be over. Fixed alongside the six, and the spec now documents it under its own heading rather than silently reflecting only the original plan.

## Scope decisions (confirmed with the repo owner before writing the spec)

- **Current window only.** Each AgentMux window is a separate renderer process with its own block registry — no existing mechanism broadcasts one gesture to every window. A torn-off pane in another window is unaffected until you repeat the gesture there.
- **Excludes a browser pane's live rendered page content.** That page is a separate native view the app's DOM/JS cannot see at all, on any platform. Catching it would need new Rust wheel interception on macOS/Linux, where none exists today (Windows has a Ctrl+Wheel hook for browser-pane zoom already, but no Shift check either) — real platform work, tracked as a candidate follow-up, not part of this change.

## Testing

- New `frontend/app/store/zoom.test.ts` (10 tests) covers `zoomAllPanesIn`/`Out` directly: stepping every zoomable pane, skipping browser/warden with zero new filtering, relative-not-shared-baseline stepping, `MAX_ZOOM` clamping, and the one-summary-toast behavior — verified by recording *every* value a Solid effect sees on the indicator signal (not just the final one), proving per-pane toasts were suppressed rather than merely overwritten an instant apart.
- `armory-view.test.tsx` / `warden-view.test.tsx` each gained a regression test proving Ctrl+Shift+Wheel no longer triggers that pane's own zoom RPC call.
- `term.tsx`/`editor-view.tsx`/`swarm-view.tsx`/`AgentShellSubblock.tsx` have no existing wheel-handler test harness to extend cheaply — their guard is covered by the fix itself plus `app.tsx`'s own dispatch logic, not a new per-view test. Flagging this rather than leaving it silently uncovered.
- `tsc --noEmit` clean. All three repo grep gates (menu-positioning, input-handler-layout-reads, scrollbar-cursor) pass. 208 tests pass across the touched view/store test files.

One process note: my first pass at this PR ran a blanket `prettier --write` across the touched files, which silently reformatted large unrelated chunks (import reordering via `prettier-plugin-organize-imports`, incidental line-wrapping) — confirmed those files are *already* non-compliant with the current Prettier config on `main` itself (pre-existing drift, no CI gate on it), so I reverted and reapplied only the intended edits by hand. Diff is proportional to what actually changed.

Changeset included (`minor`).

> [!NOTE]
> Opened via the shared `gh` session (account `AgentO-asaf`) — the `secrets` CLI `scripts/gh-agent.sh` needs to resolve Clare's own PAT isn't installed on this machine, and the GitHub MCP tools aren't connected in this session. Hence the title prefix + tag below.

### DIFF
--- a/.changesets/1788855717-feat-zoom-ctrl-shift-scroll-zooms-every-pane-in-the-window-a-79bh.md
+++ b/.changesets/1788855717-feat-zoom-ctrl-shift-scroll-zooms-every-pane-in-the-window-a-79bh.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(zoom): Ctrl+Shift+Scroll zooms every pane in the window at once
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -373,6 +373,7 @@ partial list.
 | [`SPEC_COMPOSER_STRIP_DROP_CENTER_STATS_2026_08_31`](SPEC_COMPOSER_STRIP_DROP_CENTER_STATS_2026_08_31.md) | Spec: Drop the composer strip's centered token/elapsed stats |
 | [`SPEC_COMPOSER_STRIP_ROW_BASED_LAYOUT_2026_08_26`](SPEC_COMPOSER_STRIP_ROW_BASED_LAYOUT_2026_08_26.md) | SPEC: Composer Strip — Row-Based Layout (Rev 7) |
 | [`SPEC_CONTINUOUS_SESSION_PERSISTENCE_2026_09_08`](SPEC_CONTINUOUS_SESSION_PERSISTENCE_2026_09_08.md) | SPEC: Continuous session persistence + trustworthy shutdown |
+| [`SPEC_CTRL_SHIFT_SCROLL_ZOOM_ALL_PANES_2026_09_07`](SPEC_CTRL_SHIFT_SCROLL_ZOOM_ALL_PANES_2026_09_07.md) | Spec: Ctrl+Shift+Scroll zooms every pane in the window at once |
 | [`SPEC_DEFAULT_TAB_NAME_TAB_N_2026_09_02`](SPEC_DEFAULT_TAB_NAME_TAB_N_2026_09_02.md) | Spec: default tab names — "Tab N", not "tabN" |
 | [`SPEC_DEFAULT_WIDGETS_REORDER_2026_08_25`](SPEC_DEFAULT_WIDGETS_REORDER_2026_08_25.md) | SPEC: Default fresh-start widgets — Agent, Swarm, Armory, Sysinfo |
 | [`SPEC_DEPENDENCY_UPGRADE_PROCESS_2026_08_27`](SPEC_DEPENDENCY_UPGRADE_PROCESS_2026_08_27.md) | SPEC — A repeatable process for Claude model catalog + CLI version upgrades |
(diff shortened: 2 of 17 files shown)


a5af:
@codex review

AgentO-asaf:
Fixed in 618468f — all four issues raised across both reviews.

**ReAgent's P1 / Codex's matching finding (stale WOS cache):** confirmed real. `getBlockZoom(blockId)` right after `stepZoom()` read the pre-write value — `RpcApi.SetMetaCommand` is fire-and-forget, and the local cache only updates when the backend pushes a `WaveObjUpdate` event back, not synchronously. `stepZoom`/`setBlockZoom` now return the value they just computed instead of anything being re-read. Added a regression test with a `SetMetaCommand` mock that deliberately never updates its backing map, reproducing the real timing — confirmed it fails against the old code and passes against the fix.

**Codex — warden missing from the allowlist:** correct, and a real inconsistency: `warden-view.tsx` already reads/writes `term:zoom` and applies it as CSS zoom exactly like `armory`/`swarm`, both of which were already included. Added `warden`.

**Codex — a 7th handler I missed entirely:** `helpview.tsx` has its own Ctrl+Wheel handler my original research never surfaced (it uses `onWheel={}` JSX rather than `addEventListener(..., {capture:true})` like the other six, which is exactly why the grep-based sweep missed it). Added the same `shiftKey` guard, plus a new test file — `helpview.tsx` had no test coverage before this. Its own zoom stays out of the all-panes batch (separate `help:zoom` key, not `term:zoom`), matching browser's exclusion; the fix only stops it from swallowing the gesture before every other pane can zoom.

**Also:** fixed a pre-existing dangling spec citation in `AgentShellSubblock.tsx` unrelated to this feature — the citation gate scopes to whole changed files, and my shiftKey fix there put a rot that already existed on `main` into scope. Removed per the gate's own guidance rather than inventing a replacement path.

Re-verified: `tsc --noEmit` clean, 213 tests pass (including 3 new regression tests), all doc/grep gates pass.

PR #3091, opened 2026-09-08
fix(layout): dispose evicted NodeModels' reactive roots instead of leaking them

AgentY-asaf:
Found while scoping `docs/specs/SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md`'s Option B (a follow-up to #3082) — tracing exactly what "evict the NodeModel" does, to design a safer, smaller first slice of that refactor, surfaced a genuine, currently-shipping leak. **Unrelated to whether that refactor ever gets built** — this is a real bug on its own.

## The bug

`getNodeModel` (`layoutNodeModels.ts`) builds each NodeModel's memos (`isFocused`, `isMagnified`, `isMinimized`, `innerRect`, `blockNum`, …) inside `model.runInModelRoot(...)`, which ties them to the **whole `LayoutModel`'s reactive root** — the tab's entire lifetime, not to that one `nodeModels` map entry.

Every eviction site — all three `layoutStack.ts` mutators (tab switch, tab close, tab push) plus `cleanupNodeModels`'s leaf-close path — was a bare `model.nodeModels.delete(nodeId)`. That removes the **map entry**, but disposes nothing: the memos it pointed to keep running, still subscribed to `model.localTreeStateAtom()`/`model.numLeafs()`/etc, for the rest of the tab's lifetime.

**Every in-pane tab switch and every pane close leaked one full set of ~10 live memos.** A session with many terminal-tab switches (or agent-pane fork switches, or several pane closes) accumulates that many orphaned, never-disposed reactive computations, none of which are ever cleaned up until the tab itself finally closes.

## The fix

Each NodeModel now gets its own nested `createRoot` (inside the existing `runInModelRoot` call), giving it a disposal boundary independent of every other node's. A new `LayoutModel.nodeModelDisposers: Map<string, () => void>` tracks the per-nodeid dispose function, consumed by a new `disposeNodeModel(model, nodeid)` helper. Every prior `.delete(nodeId)` caller (the 3 in `layoutStack.ts`, the 1 in `cleanupNodeModels`) now goes through that helper instead of touching the map directly.

## Verification — falsifiable, not just inspected

New test in `layoutStack.test.ts`: captures a live memo reference (`isFocused`) before eviction, evicts via the real `pushBlockOntoStack` mutator, mutates the *same dependency* that memo tracks, and asserts the old reference's value stays **frozen** at its pre-disposal state (proving actual disposal) rather than updating (what a merely-orphaned-but-still-alive memo would do).

I didn't just write this and assume it was checking the right thing — I temporarily reverted all three `layoutStack.ts` call sites back to a bare `.delete()` and re-ran the test to confirm it **fails** against the pre-fix code, then restored the fix and confirmed it passes again. That round-trip is what makes this a real regression test rather than a test that would pass regardless of whether the fix exists.

- `layoutStack.test.ts`: 19/19 pass (18 pre-existing + 1 new).
- `frontend/layout/` full suite: 203/203 pass.
- `tsc --noEmit`: clean.

## Scope note

This does **not** touch the `<Key>`/remount architecture, `NodeModel.blockId`, or anything else `SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md` §4.2 (Option B) would eventually need — that refactor is still not started, deliberately (see that spec's §2.2 for why). This is a self-contained fix to how eviction disposes, not when it fires.

### DIFF
--- a/.changesets/1788857974-fix-layout-dispose-evicted-nodemodels-reactive-roots-instead-ruwn.md
+++ b/.changesets/1788857974-fix-layout-dispose-evicted-nodemodels-reactive-roots-instead-ruwn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(layout): dispose evicted NodeModels' reactive roots instead of leaking them on every tab switch
--- a/docs/specs/SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md
+++ b/docs/specs/SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md
@@ -248,3 +248,46 @@ takes the pane header and tab strip with it, and that this is what the reveal
 gate exists to hide. That cost was previously only discoverable by reading
 four files in sequence; a future reader of `layoutStack.ts` now finds it in
 place, with a pointer here.
+
+---
+
+## 9. A real bug found while scoping Option B, fixed separately (PR TBD)
+
+Tracing exactly what "evict the NodeModel" (`layoutStack.ts:80,104,145`)
+does to design a safer, smaller first slice of Option B surfaced a
+genuine, currently-shipping leak, unrelated to whether Option B itself
+ever gets built:
+
+`getNodeModel` (`layoutNodeModels.ts`) builds each NodeModel's memos
+(`isFocused`, `isMagnified`, `isMinimized`, `innerRect`, `blockNum`, …)
+inside `model.runInModelRoot(...)`, which ties them to the WHOLE
+`LayoutModel`'s reactive root — i.e., the tab's entire lifetime, not to
+that one `nodeModels` map entry. Every eviction site (all three
+`layoutStack.ts` mutators, plus `cleanupNodeModels`'s leaf-close path) used
+to be a bare `model.nodeModels.delete(nodeId)`, which removes the MAP
+ENTRY but disposes nothing — the memos it pointed to keep running,
+still subscribed to `model.localTreeStateAtom()`/`model.numLeafs()`/etc,
+for the rest of the tab's lifetime. **Every in-pane tab switch and every
+pane close leaked one full set of ~10 live memos.** A long session with
+several terminal-tab switches would accumulate that many orphaned,
+never-disposed reactive computations.
+
+Fixed by giving each NodeModel its own nested `createRoot` (inside the
+existing `runInModelRoot` call), and a matching per-nodeid disposer map
+(`LayoutModel.nodeModelDisposers`) that a new `disposeNodeModel(model,
+nodeid)` helper consumes before removing the map entry. Every prior
+`model.nodeModels.delete(nodeId)` caller now goes through that helper
+instead. Verified falsifiably, not just by inspection: a new test
+(`layoutStack.test.ts`) captures a live memo reference before eviction,
+evicts, mutates the SAME dependency the memo tracks, and asserts the old
+reference's value stays frozen (proving disposal) rather than updating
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

AgentY-asaf:
**Both real, both fixed in \`513685b0c\`.**

**ReAgent P1 / Codex P2 (same finding)** — \`getNodeAdditionalPropertiesAtom\` wraps its own \`createMemo\` in \`model.runInModelRoot(...)\`, which calls \`runWithOwner(this._modelOwner, fn)\` — that **replaces** whatever ambient owner is active at the call site with \`_modelOwner\` directly. My round-1 fix moved the *call* inside the new nested \`createRoot\`, but that only changed *when* it ran, not *where* its memo ended up owned — it was still attaching to the model's root regardless. Caught this myself too, independently, before seeing either review: I wrote a test for exactly this field and it failed against my own "fixed" code. Fixed by inlining the memo's logic directly (no `runWithOwner` of its own) so it actually picks up the nested `createRoot` as its owner. The standalone `getNodeAdditionalPropertiesAtom` export had zero other callers once this was inlined, so removed rather than left as a landmine with the same owner-escaping behavior for a future caller.

**Codex P1** — real, and worse than a leak: the original unconditional `disposeNodeModel()` in `closeBlockInStack` would tear down a *still-mounted* component's memos when closing a **background** (non-active) tab, since that path never changes `activeBlockId` and so never remounts anything — `isFocused`/`isMagnified`/etc would freeze the instant an unrelated tab closed. Moved disposal inside the `activeBlockId`-changed branch, the only case that actually remounts.

Both verified the same way as the original fix: wrote the test, confirmed it fails against the pre-fix code (temporarily reverted, re-ran, restored), confirmed it passes against the fix. A third test confirms the active-member-close case still disposes correctly, so the P1 fix didn't overcorrect into never disposing. 206/206 `frontend/layout/` tests pass, `tsc` clean.

AgentY-asaf:
Addressed in 1f59bbe83, replying to ReAgent's round-2 P1:

> LayoutModel.dispose() ... only calls this._disposeRoot() and nulls _modelOwner; it never iterates nodeModelDisposers to dispose the nested createRoot scopes introduced at layoutNodeModels.ts:62 ... any leaf whose NodeModel is never explicitly evicted via disposeNodeModel before the whole tab closes ... has its ~10-memo reactive root leaked/undisposed on tab close ... No test exercises model.dispose() against a populated nodeModels map.

Confirmed correct. Given this PR had already gotten SolidJS ownership semantics wrong twice in earlier rounds, I didn't trust my own read of `createRoot`'s detachment behavior a third time — before implementing anything I wrote a standalone scratch test (`createRoot` nested inside `runWithOwner(parentOwner, () => createRoot(...))`, disposing only the outer/parent root) and confirmed empirically that the nested root survives. That matches exactly what you described.

Fix: `dispose()` now iterates a snapshot of `nodeModelDisposers.keys()` (spread first, since `disposeNodeModel` mutates that same map) and calls `disposeNodeModel` for each before disposing the model's own root.

New test (`layoutStack.test.ts`, `"LayoutModel.dispose()"` describe block): populates the NodeModel cache for a leaf, calls `model.dispose()`, and asserts both the map-level cleanup (`nodeModelDisposers`/`nodeModels` both empty) and a frozen-vs-live check on the previously-captured NodeModel's `isFocused` memo — mutating the same underlying signal post-dispose, a live memo would read the new value, a disposed one stays frozen. Verified falsifiable: reverting the new disposal loop makes the test fail (map-size assertions), then restored.

`npx tsc --noEmit -p .` clean; full `frontend/layout/` suite 207/207 passing (was 206).

PR #3092, opened 2026-09-08
Korp@claudius: refactor(storage): every agent_def_* method reads and writes db_agents only (agent consolidation Phase 3d, PR 4 of 5)

GenericAgentX-asaf:
## Summary

PR 4 of 5 continuing the agent-concept consolidation (`docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md` Phase 3d). Builds on #3088 (six agent-child tables' FK re-pointed to `db_agents`).

**Every `agent_def_*` `Store` method now reads and writes `db_agents` only** — mirrors PR 2's instance-side flip, for definitions. `agent_def_insert`/`find_or_insert`/`update`/`delete`/`delete_seeded`/`set_hidden`/`set_memory_id_if_empty` and `agents_grandfather_ambient_login` write `db_agents` directly; no more legacy INSERT/UPDATE/DELETE followed by a dual-write mirror call. `dual_write.rs` (definition-side helpers) is deleted — nothing calls it any more.

### The plan grew from 4 PRs to 5

Flipping definition writes forced removing the per-boot gap-repair pass (`Store::repair_agent_def_gaps`, `agents_consolidate::repair_def_gaps`, `bootstrap.rs`'s call site) in this SAME PR, not later: once `agent_def_delete`/`instance_delete` stopped also clearing `db_agent_definitions`, a repair pass still reading it every boot would have resurrected every deletion — the same class of bug PR 2's round-2 review caught, reopened by this PR's own change if left in place. That's more than "flip reads" alone, so the physical `DROP TABLE` is cleaner split into its own PR 5 rather than riding along. `docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md` updated.

`agents_consolidate::run_consolidate_migration`/`consolidate_looks_incomplete` (backing frozen `m0007`/`m0000`) are untouched — they run at migration time, before this PR's live code ever executes. `m0025`'s own frozen copy of `repair_def_gaps` needed inlining once the live function it called was deleted (`m0025` already froze its own copy of a *different* helper for the same "freeze copies of live logic" reason — this just extends that to the one it was still calling live).

### Three real bugs found and fixed along the way

1. **`agent_def_get` / `agent_def_list` disagreed about which table was authoritative — on purpose, per the module's own doc comment.** `agent_def_list` already read `db_agents` unfiltered; `agent_def_get` stayed scoped to `db_agent_definitions`, so an agent visible in "My Agents" could 404 the moment its detail view opened, if it existed only as a `db_agents` row (a template launch, which never had a `db_agent_definitions` counterpart). `agent_def_get` now reads via the same unfiltered accessor `agent_def_list` uses (`agent_row_get`, already existed privately, used internally since PR 2). The inline test documenting the old split by name is rewritten to pin the new, consistent behavior.

2. **`content.rs`/`managed.rs`'s "is this agent local" existence checks were querying the wrong table since PR 3.** `db_agent_content`/`db_agent_skills_ref`/`db_agent_mcp_ref`'s FK moved to `db_agents` in #3088, but these gates still checked `db_agent_definitions` — so a template launch's content/skill/MCP-bind writes were silently misrouted to the global-registry fallback path the moment that PR shipped. Flipped to `db_agents`.

3. **A real gate was structurally unreachable for every template launch.** `identity/resolver/inject.rs`'s Step 5 (`if let Some(p) = def_provider { ... }`) is the one check that blocks an oauth-class-provider agent with zero identity binding from spawning. `def_provider` came from `agent_def_get`, which was unconditionally `None` for a template launch — so this gate never fired for that entire agent class, regardless of whether an account was ever bound. A template-launched agent with an oauth-class CLI provider and no binding could spawn without ever hitting the check that exists specifically to block exactly that. Now that `agent_def_get` resolves the launch's own row, the gate applies like it does for every other agent. Regression test added (`use_ambient_login` confirmed inert for this gate either way, matching the existing `spawn_still_blocked_when_bound_oauth_account_missing_and_flag_true` test).

Also fixed in passing: `template_promote_resolves_provider_through_the_templates_bundle_not_the_drifted_column`'s own test was mutating `template.memory_id` then calling `agent_def_update`, relying on the old dual-write mirror's unconditional overwrite as a side channel rather than the documented "set once via `agent_def_set_memory_id_if_empty`" contract — switched to the documented path.

### Tests

```
cargo test -p agentmux-srv   → 3160 passed, 0 failed
```
(One run showed the known local-only flaky `refresh_processes_specifics_does_not_leak_a_handle_per_call` — unrelated, passes on Windows CI, flagged in every PR of this series.)

All three doc gates green.


a5af:
@codex review

GenericAgentX-asaf:
Fixed in 5297a6fe7 — correct, genuinely dead code. `stamped_updated_at` is already read at the `params![...]` call site directly above, so the trailing `let _ = stamped_updated_at;` wasn't suppressing anything. It's a leftover from the round that dropped the dual-write mirror call this line used to guard against — never re-examined once that call was gone.

```
cargo test -p agentmux-srv   → 3160 passed, 0 failed
```

GenericAgentX-asaf:

GenericAgentX-asaf:
Both P1s fixed in f3c3561df — both correct, and #2 was the more serious of the two.

**Schema version (P1 #1)** — correct. I'd been treating `OBJECT_SCHEMA_VERSION` as tracking DDL only, but `check_schema_compat` is really a downgrade guard: it refuses to open a store stamped by a newer binary. This PR changes what a newer binary's writes MEAN to an older one (new agents exist only in `db_agents`) without changing any column, so it needed the bump too. `OBJECT_SCHEMA_VERSION` → 31, no DDL change, documented as such.

**Template identity fallback (P1 #2)** — correct, and worth being explicit about why it's more serious than the P1 label alone suggests: this PR's own earlier fix (enabling Step 5's definition-provider gate for launch rows, closing the gap where a template-launched agent bypassed it entirely) would have converted every LEGITIMATELY-configured template launch into a hard spawn failure. Binding an identity to a template — the Agent Setup modal's supported target — has always written the link under the template's id, never a launch's, since nothing writes a launch-scoped link. `resolve_bindings_for_instance` now falls back to the launch's `parent_template_id`'s own links when the launch has none of its own, at both call sites (the spawn gate and `resolve_bound_oauth_config_dir`).

Two tests: the zero-binding-anywhere case still blocks (Step 5 still fires correctly on its own), and a template's real binding now resolves AND actually injects the config dir for its launches (not just passes the gate — the second test asserts the env var lands).

```
cargo test -p agentmux-srv   → 3161 passed, 0 failed
```
All doc gates green.

GenericAgentX-asaf:
P0 fixed in 414fc8d9a — correct, and genuinely the most serious finding in this whole PR series so far. Thank you for catching it.

Confirmed by reading `template.rs:443`: `forkagentdefinition` sets `fork.parent_id = source.id.clone()` unconditionally, `is_seeded: 0` — the exact same column my earlier fallback trusted as "which template was this launched from." A fork of any regular agent with a bound OAuth/API identity would have silently inherited and spawned on the SOURCE agent's credentials, since forks are deliberately created with `accounts: ""` and no link of their own (fail-closed by design — the comment right next to `parent_id` in `template.rs` documents exactly this convention for `conversation_visibility`).

`template_parent_id_if_seeded(wstore, parent_id)` now gates the fallback: it returns the parent id back only when that row genuinely has `is_seeded = 1`, so it can only ever pull a real template's links, never a fork source's. Both call sites route through it instead of trusting `AgentDefinition.parent_id` directly.

Added the exact regression test you flagged as missing: a fork of a regular agent with a bound identity is now blocked (`MissingCredentials`), not silently spawned on the source's credentials. The genuine-template fallback test from the previous commit is unaffected (a real template has `is_seeded = 1`, so it still passes the gate).

```
cargo test -p agentmux-srv   → 3162 passed, 0 failed
```
All doc gates green.

GenericAgentX-asaf:
P1 fixed in f01e4cecd — correct (both the original finding and the more detailed follow-up describe the same root cause and mechanism; addressing both here).

Confirmed: `registry_def_upsert`'s only exclusion is `def.is_seeded != 0`, and a template-launch row has `is_seeded = 0` — indistinguishable by column shape from a genuine user agent, since `instance_create`'s fresh-launch INSERT sets both `is_template`/`is_seeded` to 0 the same way `agent_def_insert` does for a real one. Before the definition flip, `agent_def_get` (which this function calls first) 404'd for a launch id, so the early return already fired before the `is_seeded` check was ever reached. Once `agent_def_get` started resolving any `db_agents` row, that protection was gone, and `content.rs`'s/`skills.rs`'s unconditional `registry_def_upsert` call after every local write (e.g. the routine `ui:zoom` pane-state write) would mirror a launch's full definition into the shared registry the first time anything touched its content — surfacing as a new, permanent "My Agents" entry in every other channel, one ghost per launch.

I looked for a column-level way to distinguish "template launch" from "genuine user-created/cloned agent" and don't think one exists: both have identical `is_seeded`/`is_template`, and `parent_id` points at a genuine template for BOTH a raw launch AND a promoted/cloned agent (the two-tier-picker's fork-into-a-real-definition flow) — so "parent is a template" doesn't separate them either.

The fix doesn't need that distinction. `content.rs`'s and `skills.rs`'s own existing comments already documented the real intent — "keep an EXISTING mirror in sync," never "mirror this agent globally for the first time." A genuine user agent already gets its first mirror at CREATE time from `agent_def_insert`'s own `registry_def_upsert` call, so by the time any content write happens its record already exists. `registry_def_refresh_if_mirrored` makes that explicit: identical mirror logic, but checks `reg.exists(def_id)` first and returns early if the record was never created. `agent_def_insert`/`update`/`find_or_insert` keep calling the original `registry_def_upsert` unchanged — creation legitimately belongs to them.

New regression test: a template launch's first content write must not create a global registry entry, and the local write itself still succeeds (confirming the fix doesn't break the actual feature, just the pollution).

```
cargo test -p agentmux-srv   → 3163 passed, 0 failed
```
All doc gates green.

PR #3093, opened 2026-09-08
docs(cef): Phase A recon output + fix the spec's open findings

a5af:
## Summary

Phase A of `SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md` is complete —
that spec gates Phases B–G on it, and says the phase "alone justifies its own
PR." **Verdict: go** on targeting 152 directly.

Full output: `docs/reports/REPORT_CEF_UPGRADE_PHASE_A_RECON_2026_09_08.md`.

Everything was verified against the actual repos via the GitHub API rather than
read off the spec — which turned out to contain errors this corrects.

## Phase A answers

| Question | Answer |
|---|---|
| Are any of the 4 patches now upstream? | **No.** `BeginWindowDrag` confirmed still absent at 152 (checked `include/views/cef_window.h` at upstream `7977`). Nothing was deleted. |
| Is there a 152-compatible `cef` crate? | **Yes** — `cef 152.0.0+152.0.5`, published 2026-09-07. |
| Is `begin_window_drag` in its bindings? | **No** — binding fork still required, as at 148. |
| Have toolchain requirements moved? | **Windows: no** (`TOOLCHAIN_HASH` + `MSVS_VERSIONS` byte-identical across the two Chromium tags). **Linux: no** (same sysroot set). **macOS: unconfirmed** — recorded as such, not assumed fine. |

## Spec corrections (both verified)

- **Patch #1 is macOS-only**, not "All platforms" — it patches
  `base/apple/mach_port_rendezvous_mac.cc`, and Chromium's `base/apple/` isn't
  compiled on Windows or Linux.
- **Patch #1 was not "registered in `patch/patch.cfg`"** — neither the entry nor
  the file existed on `7778`. It sat unmerged from 2026-06-02 until
  `agentmuxai/cef` PR #7, merged today. That PR also fixed a defect (found by
  Codex) that would have broken *every* source build from that branch: the
  patch's diff header used `a/`/`b/` prefixes, which CEF's `git apply -p0`
  patcher can't resolve. Verified by reconstructing the target and running the
  patcher's exact config against both versions.
- **Phase E's "release builds don't pin CEF at all" finding is now stale** —
  closed by #3086, #3085 and #3087. The callout describes what Phase E actually
  has to do now instead (update three literals in one job, keep them on one
  milestone or the new gate fails the release by design).

## Open findings fixed

- **§7.2** — `agentmux-cef/Cargo.toml`'s `patched-libcef` comment cited a CEF
  146-era branch on the old `a5af` org. Now points at `agentmuxai/cef` branch
  `7778`, where the drag work actually lives.
- **§7.3** — the archived Windows-GPU root-cause doc blamed the fork's libcef
  for being a DCHECK build. Its author retracted that in #1345 the same day
  (real cause: missing `supportedOS` manifest, #1354). Added a RETRACTED banner
  and corrected its Status. **Verified the retraction against #1345's own
  comments first** — that doc would otherwise mislead exactly the person
  scoping this upgrade.

**Not fixed: §7.1** (three tag version schemes). Release tags are immutable, so
it can only be normalized at read time. Noted in the report as a prerequisite
for any automated drift checker — §5 there shows it already misleading in
practice (a *newer* macOS release carries a *lower* version label than the tag
we pin).

## The finding that outranks the upgrade

All three pinned CEF release tags resolve to **the same commit, cut
2026-04-23** — which predates all four patches. **So none of the fork's
documented patches are in any CEF binary we ship**, including the macOS -67030
renderer crash fix that this repo's own retro calls "a *real* functional
code-sign failure... the only reason we went from-source at all."

Source side is now fixed (`agentmuxai/cef` PR #7 merged). **Not shipped** —
that needs an arm64 macOS build and a fresh release cut on a machine I don't
have. Detailed in §5 of the report.

## Verification

- `check-doc-status.sh`, `check-spec-citations.sh`, `gen-docs-index.sh --check`
  all pass locally. The doc-status gate caught two bad Status lines (mine, and a
  pre-existing one on the archived doc that claimed "Complete — empirically
  verified" for a retracted conclusion) — both fixed to the documented enum.
- Docs-only change; no code paths touched.

### DIFF
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -26,11 +26,18 @@ sandbox = ["cef/sandbox"]
 # Opt-in for dev_authfile unit tests that touch the filesystem and Win32 ACL APIs
 test-authfile = []
 # Enables call sites that depend on the AgentMux libcef fork
-# (https://github.com/agentmuxai/cef, branch `7778` — which is where the drag
-# work now lives, merged via that repo's PR #3; the old
-# `agentmux/7680-drag-rightclick-and-transparency` branch this used to cite is
-# CEF 146-era and no longer the current milestone. The `a5af` org also
-# redirects to `agentmuxai` now).
+# (https://github.com/agentmuxai/cef). Two branches there matter and they are
+# NOT interchangeable:
+#   - `agentmux/7778-drag-rightclick-and-transparency` — where this patch was
+#     authored, and where the shipped runtime binaries were actually built
+#     from (e.g. macOS `148.23.23-codecs` was built at `6c570e249` on it).
+#   - `7778` — the milestone integration branch the above was merged into via
+#     that repo's PR #3. Also carries `BeginWindowDrag`.
+# As of 2026-09-08 those two have DIVERGED (the build commit is 4 ahead / 2
+# behind `7778`), so "which branch" matters when reproducing a build — prefer
+# the recorded build commit over either branch tip. The old
+# `agentmux/7680-…` branch this used to cite is CEF 146-era; the `a5af` org
+# also redirects to `agentmuxai` now.
 # Specifically: `_cef_window_t::begin_window_drag` (used by StartWindowDragTask).
 # Requires a matching patched cef-dll-sys via [patch.crates-io] in the workspace
 # Cargo.toml. Default OFF so unpatched checkouts compile — native left-click
--- a/agentmux-cef/Cargo.toml
+++ b/agentmux-cef/Cargo.toml
@@ -26,7 +26,11 @@ sandbox = ["cef/sandbox"]
 # Opt-in for dev_authfile unit tests that touch the filesystem and Win32 ACL APIs
 test-authfile = []
 # Enables call sites that depend on the AgentMux libcef fork
-# (https://github.com/a5af/cef, branch agentmux/7680-drag-rightclick-and-transparency).
+# (https://github.com/agentmuxai/cef, branch `7778` — which is where the drag
+# work now lives, merged via that repo's PR #3; the old
+# `agentmux/7680-drag-rightclick-and-transparency` branch this used to cite is
+# CEF 146-era and no longer the current milestone. The `a5af` org also
+# redirects to `agentmuxai` now).
 # Specifically: `_cef_window_t::begin_window_drag` (used by StartWindowDragTask).
 # Requires a matching patched cef-dll-sys via [patch.crates-io] in the workspace
 # Cargo.toml. Default OFF so unpatched checkouts compile — native left-click
--- a/docs/analysis/archive/ANALYSIS_WINDOWS_GPU_DISABLED_ROOTCAUSE_2026_06_11.md
+++ b/docs/analysis/archive/ANALYSIS_WINDOWS_GPU_DISABLED_ROOTCAUSE_2026_06_11.md
@@ -1,9 +1,40 @@
-> **Archived 2026-07-17:** Resolved. Root-caused issue #1345, now closed. Kept for
-> historical reference only.
+> **⚠️ RETRACTED 2026-06-11 — the root cause below is WRONG. Do not cite it.**
+>
+> This document blames the `agentmuxai/cef` fork's Windows `libcef.dll` for
+> being a "non-official / DCHECK-enabled build." That attribution was
+> **retracted by its own author** the same day, in issue #1345:
+>
+> - The Windows `libcef` at the time was **not** a custom or DCHECK build — it
+>   was the stock official CEF release from `cef-builds.spotifycdn.com`. The
+>   "source paths in logs" evidence cited below is normal for any CEF app
+>   (CEF ships `is_official_build=false` by design) and does **not** indicate
+>   a DCHECK build.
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

a5af:
## Codex P1 — you were right, and it was my central claim. Corrected.

I inferred build provenance from `target_commitish`. That's invalid, and this repo's own docs say so. Verified the real picture directly rather than re-inferring:

- `STATUS_CEF_PROPRIETARY_CODECS_MACOS_2026_07_27.md` records the pinned macOS asset as built from fork commit `6c570e249` — **112 patches applied, 3 skipped, 0 failed** — with `BeginWindowDrag` re-verified *in the built framework* (`verify-cef-framework-darwin.sh` exit 0). The framework's embedded `CEF_COMMIT_HASH` is the real provenance.
- Checked that commit directly: `BeginWindowDrag` present in `cef_window.h`, `views_caption_rightclick_passthrough` registered in `patch.cfg`. **Patches #2/#3/#4 ship.**
- `docs/cef-build/build-patched-cef-windows.md` confirms Windows builds from the same fork branch ("present but inert on Windows"), so it carries them too.

**What survives, and is now better-supported:** patch #1 alone is missing — and the reason is sharper than "stale release." It was never in `patch/patch.cfg` on *any* branch, and `patcher.py` reads that file to decide what to apply. So no build on any platform could ever have applied it. That was established by direct file checks (404 on the file, 0 `patch.cfg` matches), not by the bad inference.

**Consequences of the error, stated plainly:**
- §5 rewritten with the retracted claim quoted and marked, not silently edited.
- The "re-cut all three platforms" recommendation is **retracted** — it was over-scoped work that this error would have caused. It's macOS-only, one rebuild.
- Added §5.3 recording the method mistake so the next person doesn't repeat it.
- I'd already told a peer agent the wrong version; I've sent them the correction.

## Codex P2 — fair, also fixed

Phase A was marked `implemented` with two of three checks open. Both the report and the spec are now `active`, the verdict is **conditional go**, and the open checks are listed as explicit conditions rather than footnotes.

One of them did close in the meantime: patch #1 is now verified to apply to **real Chromium 152 source** unchanged (fetched the target file at `152.0.7977.83`, applied with CEF's own `git apply -p0` config). New §2.5. Patches #3/#4 still need a real checkout — that remains condition #1.

## Also

Regenerated `docs/specs/INDEX.md` — it's generated from spec Status and this PR changes one. That was the doc-gate CI failure, unrelated to the above.

a5af:
## ReAgent P2 — correct, fixed

The dispatch-ref guard merged as **#3089** (`cfe0cdf`), not #3087. Verified against git history: #3087 is CLOSED — a superseded pinning attempt that never landed — and doesn't appear in the history at all.

Source of my error, for the record: `release.yml`'s own inline comment says *"codex's review of that change (PR #3087)"*. That cites where the **review** happened, not where the change **landed**. I copied the number without checking it against history.

Both occurrences fixed (lines 10 and 175).

Worth noting this is the third citation-accuracy issue caught on this PR (after the `target_commitish` provenance error and the Phase A status overclaim) — all the same failure mode: trusting a secondary source instead of checking the primary one. The report's §5.3 method note now covers the provenance case specifically; this one reinforces the general point.

a5af:
## ReAgent P2 — fixed

Right, and it was the worst place to leave it: the Status line is the line people actually read, and it said an unqualified "go" while §4 and the report both said conditional-with-two-open-checks. It contradicted the exact safeguard §4 spells out.

Status now states the conditional verdict, names both open checks inline (patches #3/#4 not test-applied against real 152 source; macOS Xcode pin unconfirmed), and separates Phase B (started opportunistically, patch #1 ported and verified) from C–G (not started).

Grepped for any other unconditional "go" — none left in either doc.

a5af:
## ReAgent P2 (both) — the stated reason is wrong, but checking it found something better

**The claim that `7778` is "the Chromium milestone number, not a git branch ref" is factually incorrect.** `7778` is a real branch in `agentmuxai/cef` — `gh api repos/agentmuxai/cef/branches/7778` returns it (HEAD `5387e497`), it carries `BeginWindowDrag` in `include/views/cef_window.h`, and this whole session has been reading files off it via `?ref=7778`. The fork names its milestone integration branch after the CEF branch number, so it's both.

So I'm not applying the suggested fix as-given — it would have replaced a correct reference with a differently-scoped one.

**But checking it surfaced a genuine problem worth more than the original nit.** The shipped binaries were *not* built from `7778`:

- `cef-macos-arm64-148.23.21`'s target is `agentmux/7778-drag-rightclick-and-transparency`
- the pinned `148.23.23-codecs` was built at `6c570e249`, on that same feature branch
- **that build commit is diverged from `7778` — 4 ahead, 2 behind**

So citing *either* branch alone is misleading, for a reason neither of us started with: they're not the same code, and the one builds come from isn't the one the patch was merged into.

Updated the comment to name both, explain the relationship, record the divergence, and advise preferring **the recorded build commit** over either branch tip when reproducing a build. That's consistent with root `Cargo.toml`, which already refers to the feature branch — and it's the guidance that would actually have prevented my earlier provenance error on this same PR.

Same treatment applied to §7.2's fix note in the spec.

PR #3094, opened 2026-09-08
Loap #2@claudius: feat(discovery): expose host hostname in /agentmux/discovery

GenericAgentX-asaf:
## Summary

`GET /agentmux/discovery`'s `host` object reported `version`, `local_url`,
`addressable`, `agents` and `cross_channel` — but no machine name. Since
`local_url` is always loopback (`http://127.0.0.1:<port>`, hardcoded in
`bootstrap.rs` regardless of bind address), a client has no way to tell
*which machine* it's talking to.

Found while working on `agentmux-mobile`: its Discovery screen can label
LAN-discovered instances correctly (the UDP broadcast responder already
sends `hostname`), but a directly-connected instance could only ever be
shown as a raw IP — `10.0.2.2` on the Android emulator.

## Approach

The value already exists — `whoami::fallible::hostname()` is resolved at boot
for LAN discovery's mDNS TXT records (`bootstrap.rs`) — it just never reached
`AppState`. This carries it on `NetworkBundle` (where it's already computed)
and hands it to `build_app_state`, rather than resolving it a second time.

## Changes

- `NetworkBundle` gains a `hostname` field; the existing `hostname` local is
  cloned into it instead of being moved straight into `LanDiscoveryController`.
- `AppState` gains `pub hostname: String`.
- `handle_discovery` emits `host.hostname`.
- Three test-helper `AppState` literals updated.

## Testing

- New `discovery_reports_host_hostname` test in `server/tests.rs` asserts the
  field is present in the response — passes.
- `cargo check -p agentmux-srv` and `--tests` both clean.

## Notes

This is one of two gaps logged in `agentmux-mobile`'s
`docs/retro/retro-mobile-dev-and-discovery-session-2026-09-08.md`. The
other — LAN peers always reporting `agents: []`, so there's no way to
discover *which* agents are on another host — is the larger one and is
not addressed here.


a5af:
@codex review

PR #3096, opened 2026-09-08
Korp@claudius: Phase 3e — drop db_agent_definitions and db_agent_instances (agent consolidation, PR 5 of 5)

GenericAgentX-asaf:
## Summary

Final PR in the agent-concept consolidation series
(`docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md`): drops
`db_agent_definitions` and `db_agent_instances` entirely. Nothing has read
or written either table since PR 2 (#3080, instance flip) and PR 4 (#3092,
definition flip); PR 3 (#3088) re-pointed the six agent-child tables' FK
onto `db_agents`. This is a straight `DROP TABLE IF EXISTS`, schema v32.

- `db_agent_definitions`, `db_agent_instances` — dropped by new migration
  `m0029_drop_legacy_agent_tables`.
- `run_object_schema` no longer declares either table (`CREATE TABLE IF
  NOT EXISTS` removed, along with 10 `ALTER TABLE ... ADD COLUMN` entries
  for `db_agent_definitions`). This had to ship in the **same** release as
  the drop migration — `CREATE TABLE IF NOT EXISTS` runs on every boot
  *before* migrations, so leaving it in would silently resurrect an empty
  table right after `m0029` drops it. The `ALTER` removal matters
  separately: that swallow-list only catches "duplicate column" errors,
  not "no such table" — a stale entry there would break every boot after
  the drop.
- `adopt_legacy_table_names`'s `db_forge_agents → db_agent_definitions`
  rename entry is deliberately left in place. Traced the ordering: any
  store old enough to still have `db_forge_agents` gets it renamed by an
  *earlier* migration's own `Store::open()` call (m0025–m0028 each open
  the store internally) before `m0029` runs later in the same pass — so
  the rename target existing only transiently is safe.
- `agents_consolidate::consolidate_looks_incomplete` /
  `run_consolidate_migration` (backing frozen `m0007`, itself backing
  `m0000`'s stamping) are made tolerant of either legacy table being
  physically absent. **This is load-bearing, not defensive-only:** a
  fresh v32+ channel gets `m0007` stamped "applied" via its own
  pre-existing no-op guard, and a later `muxspect migrations --verify`
  doctor pass calls `m0007::verify()` — which calls this — against that
  channel once it has ordinary `db_agents` data. Without the tolerance
  fix, that doctor pass errors "no such table" on every fresh install,
  not just an upgraded one.
- `agents_consolidate.rs` itself, `m0007`, and its on-disk marker are
  **deliberately not retired** in this PR. Removing an already-applied
  migration is a "minimum supported upgrade path" policy call for the
  repo owner, per `migrations/mod.rs`'s own doctrine — not something to
  decide unilaterally as a side effect of the drop. The module is made
  tolerant instead, so it keeps working indefinitely regardless of when
  (or whether) that call gets made.
- Five shared test fixtures (`m0025`, `m0026`, `m0028`,
  `agents_consolidate.rs`, `phase5_tests.rs`) relied on `Store::open()`'s
  base schema to provide the legacy tables for their own raw-SQL setup;
  each now stands up the pre-drop shape by hand (frozen, per-fixture,
  matching exactly what that fixture's consuming code reads).
- One fixture needed a genuine behavior update, not just a schema patch:
  `phase5_tests.rs`'s crash-recovery test
  (`a_crash_between_a_migrations_effect_and_its_mark_resumes_without_duplicating_rows`)
  forces a from-scratch retry of `m0007` in a *separate*, later
  `apply_all()` pass — against the real, full registry, `m0029` has
  already run and dropped the source tables by the time that retry
  happens. The retry is still safe (no duplicate `db_agents` rows), but
  can no longer rewrite a marker for a backfill with nothing left to
  read. Updated the test's comment and assertions to state that, rather
  than loosening them to hide it.

`docs/specs/SPEC_AGENT_ARCHITECTURE_2026_05_27.md` updated: status stays
`active` (not `implemented`) — the doc's acceptance criteria describe a
broader scope (JSON registry sunset, orphan working-dir cleanup, frontend
`AgentDefinition`/`AgentInstance` → `Agent` rename) the 5-PR plan never
covered. Only the 3 criteria this series actually delivered are checked
off; the other 3 are explicitly called out as unaddressed follow-up.

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 3166 passed, 0 failed
- [x] `scripts/check-doc-status.sh` — ok
- [x] `scripts/check-spec-citations.sh` — ok
- [ ] CI green on this PR


a5af:
@codex review

GenericAgentX-asaf:
Fixed in 4d980f86e — restored the 10 db_agent_definitions ALTER TABLE ADD COLUMN entries, gated on the table still existing (sqlite_master check), instead of removing them outright. Confirmed the exact failure path: m0007/m0025/m0028 each call Store::open() internally before m0029 drops the table in the same pass, and their frozen SELECTs (m0007's updated_at; m0025/m0028's user_hidden + the container/ambient-login/model-vendor/auto-continue/memory/conversation-visibility columns) would hit "no such column" on a real pre-v2/v3 store upgrading straight to v32. Added two tests locking in both halves of the gate (column added when the table exists; block skipped once it's gone). Full suite: 3168 passed, 0 failed.

PR #3097, opened 2026-09-08
Claude@gamerlove: fix(agent-launch): stop autofilling the agent name from the model

a5af:
## Summary

The New Agent launch modal pre-filled the name field with `<Provider> Agent`
(e.g. `Claude Agent`, `Claude Agent 2`) the moment it opened (#780), so
clicking Launch without typing anything produced an agent named after its
model rather than its purpose. A fleet of agents all named after their model
carries no information about which agent does what.

## Change

- The name field now opens empty. Placeholder is a static ghost-text hint,
  `Descriptive nickname`, instead of the model's display name.
- A real name must be typed before Launch is enabled (`hasName()` already
  gated on non-empty input; unaffected by this change).
- `AgentLaunchModal` is the single shared launch surface for every provider
  (`displayName()` already resolves generically per-provider) — removing
  the autofill effect in this one place is the complete fix, no
  per-provider changes needed or made.
- Deleted `default-agent-name.ts` + its test: the removed effect was their
  only caller.

Full rationale: `docs/specs/SPEC_AGENT_LAUNCH_NAME_NO_AUTOFILL_2026_09_08.md`
(supersedes `docs/specs/SPEC_DEFAULT_AGENT_NAME.md`, issue #780).

Continue mode is unaffected — it never autofilled (name comes from the
continued row).

## Test plan

- [x] `npx vitest run` — 576/576 passing (44 test files), including the
      full `frontend/app/view/agent/components` suite.
- [ ] Manual: open New Agent modal, confirm field is empty with
      "Descriptive nickname" ghost text, Launch stays disabled until a name
      is typed, across at least two different providers.


a5af:
@codex review

PR #3098, opened 2026-09-08
Camper@narko: docs: spec OS-shutdown handling + continuous session persistence

GenericAgentX-asaf:
## Summary

Proposes making AgentMux's session record survive the ways a process actually dies, and making the shutdown path that flushes it trustworthy. Docs-only.

## Provenance — read this first

**The incident that prompted this spec turned out not to be a defect.** A reboot appeared to lose a workspace, but the actual cause was that a *different per-build channel* was opened: two distinct 0.55.39 builds existed (`local-main-b28b7a-2cd6e3d7` and `local-lark-gated-migrations-no-au-…`), and per-build data isolation is documented, intended behavior (`CLAUDE.md`: *"each build is its own AgentMux instance: its own data dir… pane layout + memories start fresh"*). Same version string, different build, different data dir.

The spec says so plainly in §1 rather than burying it, because a reader who assumes the incident *was* the bug will misjudge the priorities. The gaps below were found **while** investigating it, not **because** of it — and they stand on their own evidence.

## The two real gaps

**Gap 1 — the OS cannot ask us to save.** `WM_QUERYENDSESSION` and `WM_ENDSESSION` appear **nowhere** in the Rust sources. On a Windows restart the process is terminated, `backend_close_window` is never called, and `save_last_session_snapshot` never runs. Same class of gap on macOS (`applicationShouldTerminate:`); Linux is partially covered only via SIGTERM.

**This one is confirmed against live data (§1.1), not just code.** The instance killed by the reboot shows the unambiguous signature of a process terminated with no graceful close:

```
session:last_topology  => ABSENT      (snapshot never written)
meta keys              => []
windowids              => ['d354d7fa…']  (never emptied)
db_window: 1   db_workspace: 1   db_tab: 3   db_block: 22
```

`windowids` never emptied ⇒ the `is_empty()` gate that triggers the snapshot never fired, and `delete_workspace` never ran. So the gap is real and would bite anyone who reboots and reopens the *same* build — it just wasn't the reported symptom.

**Gap 2 — even the wired-up path isn't durable.** One point-in-time write, outrunnable by process exit (host allows srv 2000ms; srv's teardown grace is 5s/shell; on Windows srv is `child.kill()`'d outright, #2979), and silent on failure (`warn!`'d, handler returns success regardless).

## Four phases, independently shippable

- **Phase 0 — handle OS shutdown notifications.** Smallest change, independently valuable. Flush on `WM_ENDSESSION`; return TRUE from `WM_QUERYENDSESSION`. Deliberately does *not* run the full `delete_workspace` cascade — its 5s-per-shell grace won't fit the OS's shutdown budget.
- **Phase A — continuous record.** Debounced, dirty-gated writes to the same `Client.meta["session:last_topology"]` slot (no migration). Mirrors patterns already proven here — `layoutPersistence.ts:396-420`'s 100ms debounce and `useSnapshotPersistence.ts:43,51,73`'s interval + dirty-check + serialized-writer chain, whose comments already name the clobber race a new implementation would rediscover. Also makes Phase 0's flush cheap.
- **Phase B — clean-exit marker**, so restore can distinguish a quit from a crash. Written *after* the work it attests to, per `SPEC_MIGRATION_SYSTEM_HARDENING_2026_08_03.md`'s "existence is not proof of effect."
- **Phase C — shutdown hardening.** Surface snapshot failure; replace the host's fixed 2000ms guess with a real srv completion signal; give Windows a graceful backend shutdown (#2979).

## Relationship to #2981

Complementary; sequencing matters. `SPEC_SHUTDOWN_COUNTDOWN_MODAL_2026_09_04.md` proposes a progress stage labeled *"Saving your session"* — today that narrates a fire-and-forget write on a platform about to hard-kill the backend, which that spec is candid about. Recommended order: **Phase 0 → A → C.1/C.2 → countdown modal → C.3**.

## Scope

Topology only. Explicitly **not** live process state, scrollback, or conversation content — and agent output must not dirty the session record. Named saved Layouts (Feature 2 of the 08-13 spec), snapshot history, and cross-machine sync are out of scope.

Also records an incidental finding worth its own follow-up: those 22 orphaned blocks are still intact but unreachable — leftover live rows are correctly no longer a restore mechanism, and no snapshot exists either. A "recover an orphaned workspace" affordance may be worth having.

Four open questions in §7, including whether a *crash*-recovered session should auto-relaunch agents (today's restore does), and whether `window:confirmclose` (exists in three places, read by zero call sites) should finally be wired up.

### DIFF
--- a/.changesets/1788888050-docs-spec-continuous-session-persistence-trustworthy-shutdow-yg0b.md
+++ b/.changesets/1788888050-docs-spec-continuous-session-persistence-trustworthy-shutdow-yg0b.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs: spec continuous session persistence + trustworthy shutdown
--- a/docs/specs/SPEC_CONTINUOUS_SESSION_PERSISTENCE_2026_09_08.md
+++ b/docs/specs/SPEC_CONTINUOUS_SESSION_PERSISTENCE_2026_09_08.md
@@ -8,17 +8,24 @@ Feature 1 (shipped) — see §2 for exactly what is already built and what is ne
 
 ## 0. TL;DR
 
-Session restore exists today but only writes **once**, at the moment the last
-window closes. That single write is the only thing standing between a user and
-a lost workspace, and three separate properties of the current shutdown path
-make it untrustworthy: it never runs on a crash, it can be outrun by the
-process exiting, and when it fails it fails silently.
+Session restore works — it shipped 2026-08-14 (PR #2560) and correctly
+restores a workspace when the user closes AgentMux and reopens it. But it
+writes **once**, on that graceful close, and **the OS is never able to ask for
+that write**: AgentMux does not handle `WM_QUERYENDSESSION`/`WM_ENDSESSION`
+at all, so a Windows restart terminates the process without the snapshot ever
+being taken. Three further properties make even the wired-up path
+untrustworthy: it never runs on a crash, it can be outrun by process exit, and
+when it fails it fails silently.
 
-This spec makes the session record **continuous** (bounded crash-loss instead
-of total crash-loss) and makes the shutdown path that flushes it **actually
-finish before the process dies**. It deliberately reuses two auto-save
-patterns already proven elsewhere in this codebase rather than inventing a
-third.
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #3099, opened 2026-09-08
Camper@narko: docs: regenerate specs index + fix section order in the session-persistence spec

GenericAgentX-asaf:
## Summary

Follow-up to #3098, fixing two defects that landed on `main` because that PR auto-merged before these commits made it in.

- **`docs/specs/INDEX.md` was stale on main** — the new spec was added without regenerating the index. This is the more urgent of the two: `gen-docs-index.sh --check` is what the `doc status + grep gates` job runs, so leaving it stale means that gate keeps failing.
- **Section 4.0 was physically out of order** — the design section read 4.1, 4.2, **4.0**, 4.3, while the surrounding text tells the reader to ship Phase 0 first. Also fixes a cross-reference pointing at §8 ("Testing") for "out of scope" when Non-goals is §6.2.

Both are `git cherry-pick`s of commits that were already pushed to #3098's branch (`a76108a7d`, `c27fe906c`) but hadn't been picked up when it merged.

## How this happened — worth recording

I armed `--auto` merge on #3098 while the `doc status + grep gates` check was already failing, assuming auto-merge would gate on it. It didn't, because that job **is not a required status check** — per `ci-pr.yml`'s own comment, only `check --tests + test (windows-latest)` and `vitest` are enforced in branch protection. So the PR merged the moment those two went green, carrying the stale index and the failing gate with it.

Compounding it, GitHub then stopped syncing #3098's head: the branch API showed `c27fe906c` while the PR's own commit list stopped at `0ed8bb46d`, with `mergeable: null` for 10+ minutes. That wasn't lag — the PR had already merged at the older SHA, which is why it never advanced.

**The generalizable lesson:** "approved + auto-merge armed" is not the same as "all checks will be respected." Before arming `--auto`, either confirm the failing check is non-required *and* that shipping it red is acceptable, or fix it first. Here it wasn't acceptable — a stale index breaks the gate for everyone else touching `docs/specs`.

## Verification

```
check-doc-status:     ok (2 doc(s) checked)
check-spec-citations: ok (2 file(s) checked)
gen-docs-index.sh:    single-row addition, no other drift
```


a5af:
@codex review

PR #3102, opened 2026-09-09
Loap #2@claudius: feat(lan): populate LAN peers' agent lists

GenericAgentX-asaf:
## Summary

Every LAN peer reported `agents: []` — always. `LanInstance::agents` existed
and was initialized empty at mDNS-resolve time, but nothing ever populated it.
Consequence: there is no way to discover *which* agents live on another
machine. Confirmed live on this network — `starpower` and `gamerlove` both
resolve with correct hostname/version/address and zero agents.

This is retro item A1 from `agentmux-mobile`'s
`docs/retro/retro-mobile-dev-and-discovery-session-2026-09-08.md`. I hit it
concretely: asked to message an agent on another host, the only option was to
address it blind by name and hope. (It worked — `SendMessage` resolves by name
across tiers — but discovery gave zero confirmation the target existed.)

## Approach

- New `GET /agentmux/reactive/agent-names` on the existing
  `lan_or_full_auth_middleware` group.
- A 30s refresh loop asks each known peer and fills in `LanInstance::agents`.
  Slower than `LAN_AGENT_CACHE_TTL_SECS` on purpose: this list is for
  *display*, not delivery routing — `find_agent` still does its own
  authoritative lookup.
- Refresh failures are **non-destructive**: a briefly-unreachable peer keeps
  its last-known list rather than flapping to empty (staleness is already
  tracked by `last_seen`). A peer on an older build answers 404, which is
  logged at debug and treated the same way — expected during a rolling upgrade.
- The snapshot of peers is taken under the read lock and released before any
  `await`, so the map isn't held across I/O.

## Security scoping — please review this part specifically

This **widens what a captured `lan_key` is worth**, so it was an explicit
decision (repo-owner-confirmed, 2026-09-08), not an incidental consequence.

The new route returns agent **names only**. `block_id`, `tab_id`,
`registered_at`, `last_seen` and `registration_nonce` stay behind full auth on
the existing `/agentmux/reactive/agents` — which
`lan_key_is_rejected_on_other_reactive_routes` still pins at 401, unchanged.

The tradeoff was accepted because a `lan_key` holder could **already** confirm
any *specific* agent name via `/agentmux/reactive/agent?id=…`, so this makes
enumeration cheap rather than newly possible. The alternative considered and
rejected for now was requiring a signed request (reusing
`SPEC_JEKT_LAN_TIER_SIGNING`'s per-agent Ed25519 keys) — stronger, meaningfully
more work, and can layer on later.

`SPEC_JEKT_LAN_WAN_TRUST_HARDENING_2026_08_13.md` gets a scope amendment
recording this, since that spec is what deliberately narrowed the credential
three weeks ago. The now-stale "two LAN-forwarding routes" comments across
`lan_discovery.rs`, `bootstrap.rs`, `server/mod.rs` and `tests.rs` are updated
to three.

## Testing

Two new tests, both passing:

- `lan_key_can_list_agent_names` — a registered agent's name is returned to a
  `lan_key` caller.
- `agent_names_exposes_names_only_not_registration_internals` — asserts the raw
  response body contains none of `block_id`, `tab_id`, the tab value itself,
  `registration_nonce`, `registered_at`, `last_seen`. This is the assertion
  that makes the scoping decision above enforceable rather than aspirational.

All 10 existing `lan_key_*` / `agent_names*` tests pass together, including the
pre-existing hardening tests that pin the general surface at 401.

### DIFF
--- a/.changesets/1788928077-feat-lan-peers-report-which-agents-they-host-s58f.md
+++ b/.changesets/1788928077-feat-lan-peers-report-which-agents-they-host-s58f.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(lan): peers report which agents they host
--- a/agentmux-srv/src/backend/lan_discovery.rs
+++ b/agentmux-srv/src/backend/lan_discovery.rs
@@ -242,6 +242,12 @@ pub struct LanDiscovery {
     /// `shutdown()` is `&self` and called from both an explicit live-toggle
     /// path and `Drop`.
     udp_cancel: Mutex<Option<oneshot::Sender<()>>>,
+    /// Cancels `agent_names_refresh_loop`. Needed for the same reason
+    /// `udp_cancel` is: that task holds its own `Arc<LanDiscovery>` clone, so
+    /// dropping the controller's Arc on `apply(false)` never reaches refcount
+    /// zero and the loop would otherwise keep issuing authenticated requests
+    /// to peers forever — leaking one orphaned task per disable/enable cycle.
+    agent_names_cancel: Mutex<Option<oneshot::Sender<()>>>,
 }
 
 /// Normalize an OS hostname into a valid mDNS host name by appending the
@@ -406,6 +412,7 @@ impl LanDiscovery {
             version,
             port,
             udp_cancel: Mutex::new(None),
+            agent_names_cancel: Mutex::new(None),
         });
 
         // Spawn event receiver on a blocking thread to avoid starving the tokio runtime
@@ -426,10 +433,16 @@ impl LanDiscovery {
         });
 
         // Keeps each peer's `agents` list populated (see
-        // `agent_names_refresh_loop`). Dropped along with the whole
-        // `LanDiscovery` when the setting is toggled off, same as the
-        // loops above.
-        tokio::spawn(discovery.clone().agent_names_refresh_loop());
+        // `agent_names_refresh_loop`). Cancelled explicitly by `shutdown()`
+        // via its own oneshot — NOT by Arc drop, since this task holds its
+        // own clone and would otherwise outlive `apply(false)` forever.
+        let (names_cancel_tx, names_cancel_rx) = oneshot::channel();
+        *discovery.agent_names_cancel.lock() = Some(names_cancel_tx);
+        tokio::spawn(
+            discovery
+                .clone()
+                .agent_names_refresh_loop(names_cancel_rx),
+        );
 
         tracing::info!(
             instance_id = %instance_id,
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

a5af:
@codex review

GenericAgentX-asaf:
Fixed in 2814b59 — this was a real P1, and the review was right on both counts.

**The bug:** the refresh task holds its own `Arc<LanDiscovery>` clone, so `apply(false)` dropping the controller's Arc never reaches refcount zero. `shutdown()` cancelled `udp_responder_loop` and the mDNS daemon but had no signal for this loop, so it kept issuing authenticated requests to peers indefinitely after LAN discovery was disabled — one orphaned task per toggle cycle.

**Worse, and worth calling out:** the doc comment I shipped claimed the task was "dropped along with the whole LanDiscovery when the setting is toggled off, same as the loops above." That described cancellation that did not exist, which is arguably the more damaging half — it would have told the next reader the question was already settled.

**Fix:** an `agent_names_cancel` oneshot mirroring `udp_cancel`, selected on *inside* the sleep (so cancellation is prompt rather than up to 30s late), fired from `shutdown()`.

**On the test** — since the original defect was precisely "a comment asserting behavior nothing enforced," I verified the new test in both directions rather than just adding it:

- With the fix: passes.
- With `shutdown()` reverted to not touching the field (the original shape): fails with `refresh loop did not exit within 5s of shutdown() - it would outlive apply(false)`.

One note from doing that, in case it matters to a future reader: my *first* attempt at the negative check only disabled the `tx.send(())` call, and the test still passed — dropping the sender also cancels, because the receiver resolves on channel close. So that check proved nothing. The test is only meaningful against the real bug shape (`shutdown()` ignoring the field entirely), which is what it's now pinned against.

GenericAgentX-asaf:
Fixed in 88bbbcc — all three, and the P1 was a real one.

**[P1] Unbounded response body.** Worth stating plainly since it's easy to miss: the `lan_key` a peer advertises authenticates *this client to that peer*, not the peer to us. Anything that can fake an mDNS advertisement for `_agentmux._tcp.local` gets into `self.instances` and is polled by this loop — so a malicious "peer" could return an arbitrarily large successful response and `resp.json()` would buffer all of it before any check ran. Fixed by reading via a hand-rolled chunked loop capped at 64 KiB, bailing the moment it's exceeded — not trusting `Content-Length` for this, since a malicious peer can omit or lie about it while still streaming an unbounded body. Added count (500) and per-name-length (256) caps past that as defense in depth, since a compact-but-adversarial response could still pack many short strings into 64 KiB.

**[P2] Sequential polling.** Fanned out with `FuturesUnordered`, same pattern `query_peers_concurrently` already uses for the sibling lookup path. Confirmed the fix actually matters, not just tidier: two peers each answering after 300ms took 608ms sequential vs staying under the 500ms budget concurrent.

**[P2] No broadcast on change.** Now compares old vs new per peer and calls `broadcast_instances()` only when something changed, instead of never — the loop was silently defeating its own display purpose.

**Testing approach worth flagging**, since it's the same discipline the cancellation fix needed: both new tests are checked in **both directions**, not just added.

- `oversized_response_is_rejected_not_buffered` — fails against the original unbounded `resp.json()` call when I temporarily reverted to it.
- `peers_are_polled_concurrently_not_sequentially` — fails with the two peers taking 608ms when I temporarily drove them with sequential `.await` instead of `FuturesUnordered`.

Both use a real TCP server (same style as `server::tests::spawn_fake_browser_api`), not an in-process `tower::ServiceExt::oneshot` mock — the behavior under test is bailing out *mid-body*, which an in-process call can't exercise.

Full `agentmux-srv` suite: 3179 passed, 0 failed. Doc gates clean.

PR #3103, opened 2026-09-09
docs(cef): decide on staged per-platform rollout for the 152 upgrade

AgentX-asaf:
## Summary

Decides how Phase C ships: **staged per-platform rollout instead of a single workspace-wide `cef` version bump.** Windows moves to CEF 152 first; macOS and Linux stay on 148 in source *and* runtime until their own Phase D builds land. Spec-only — no code changes yet, this is the design decision Phase C's implementation should follow.

## Why this over the simpler bump

Repo owner's explicit requirement: all three platforms must keep building and running throughout the upgrade — for end users, and for anyone developing on macOS/Linux locally. A plain workspace-wide `cef` version bump doesn't satisfy that: the moment `agentmux-cef/Cargo.toml`'s single pin changes, `task dev` on a macOS/Linux machine fails immediately (the crate-major/runtime-major assertion in `Taskfile.yml` correctly rejects a still-148 runtime against a 152-linked binary) until that platform's own 152 runtime exists — which could be days or weeks away depending on build-machine availability.

## Why this is now known to be low-risk, not just appealing

Phase C's item 3 ("fix compile fallout from CEF API changes... the least predictable item in this spec") is de-risked, not assumed away. Enumerated all 13 distinct `cef::Impl*` traits `agentmux-cef`'s Rust source actually calls, mapped each to its upstream C++ header, and diffed all 12 of those files (some traits share a header) between CEF `7778` (148) and `7977` (152) via full-text comparison against the real upstream files — not sampled, not inferred from a changelog.

**Result: all 12 are byte-for-byte identical.** Zero `#[cfg(...)]`-branching is needed in the Rust source to support two CEF milestones side by side, because nothing this codebase's current call surface touches has moved between these milestones.

One methodological note worth flagging since it changed my own conclusion mid-investigation: my first attempt used GitHub's Contents API against `include/capi/...` paths, which don't exist in `chromiumembedded/cef`'s actual repo layout (the C API headers are generated at build time, not checked in) — that silently returned empty content on both sides, producing a false "no diff" from comparing nothing against nothing. Caught it via a sanity check (non-zero byte-length verification) before trusting the result, then redid it correctly against the real `include/*.h` / `include/views/*.h` paths via `raw.githubusercontent.com`.

## What changes mechanically

`[patch.crates-io]` is a single global override keyed by crate name — Cargo can't hold two different pinned revs of `cef-dll-sys` active for different targets through that mechanism. Phase C's steps 1-2 are replaced with:

1. Remove `cef` from `agentmux-cef/Cargo.toml`'s plain `[dependencies]`; remove the root `[patch.crates-io]` block.
2. `[target.'cfg(target_os = "windows")'.dependencies]`: `cef = "152"`, patched to `AgentU-asaf/cef-rs@agentmux/152-begin-window-drag` once published (`begin_window_drag` confirmed absent from the public 152 crate — still needed).
3. `[target.'cfg(any(target_os = "macos", target_os = "linux"))'.dependencies]`: `cef = "148"`, unchanged rev (`515b3ac53c`).
4. Validate with `cargo check --target x86_64-pc-windows-msvc` and `--target x86_64-unknown-linux-gnu` before merging — this workspace already uses target-gated dependencies extensively (`agentmux-cef`, `agentmux-launcher`, `agentmux-srv`, `agentmux-common` all do it for platform-exclusive crates), but never for the *same* crate at two different, semver-incompatible versions. Should be proven, not assumed.
5. No Phase E changes needed — the per-platform runtime pinning already landed via #3085/#3086/#3089 independently.

**Exit condition, stated explicitly so this doesn't become permanent scaffolding:** once Phase D lands working 152 builds for macOS and Linux too, collapse back into a single `[dependencies]` entry.

### DIFF
--- a/.changesets/1788928478-docs-cef-decide-on-staged-per-platform-rollout-for-the-152-u-0f5l.md
+++ b/.changesets/1788928478-docs-cef-decide-on-staged-per-platform-rollout-for-the-152-u-0f5l.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(cef): decide on staged per-platform rollout for the 152 upgrade
--- a/docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md
+++ b/docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md
@@ -160,6 +160,71 @@ Note: `agentmuxai/cef` has **zero CI** (verified: `actions/workflows` → `total
 2. Rebase the `cef-dll-sys` binding patch (`begin_window_drag` slot) onto the 152 binding; publish as `AgentU-asaf/cef-rs@agentmux/152-begin-window-drag`; update the `[patch.crates-io]` rev in root `Cargo.toml`.
 3. Fix compile fallout from CEF API changes across four milestones — **the least predictable item in this spec**; no way to size it before Phase A.
 
+> **Decision (2026-09-08): stage the rollout per platform instead of bumping the
+> workspace-wide `cef` version all at once — Windows moves to 152 first;
+> macOS/Linux stay on 148, in source and runtime both, until their own Phase D
+> builds land.** Repo owner's explicit requirement: all three platforms must
+> keep building and running throughout the upgrade, for end users *and* for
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #3104, opened 2026-09-09
Claude@gamerlove: chore: release v0.55.40

GenericAgentX-asaf:
## Release v0.55.40

Forced patch bump (`task release -- --as patch`), consuming 18 pending
changesets. All five version locations verified to agree at `0.55.40`
per the release-consistency invariant in `CLAUDE.md`:

- `package.json` → `0.55.40`
- `Cargo.toml [workspace.package].version` → `0.55.40`
- `Cargo.lock`'s `agentmux-cef` entry → `0.55.40`
- `package-lock.json` → `0.55.40`
- `VERSION_HISTORY.md` top section → `0.55.40`

Verified using the exact same shell checks `scripts/release.sh` itself
runs (not just eyeballed), after installing the Rust toolchain and
`@a5af/bump-cli` on this machine — neither was available here before,
which is why the lockfile initially failed to regenerate. `cargo
generate-lockfile` doesn't need a linker, so this didn't require the
MSVC build tools that are still missing here.

### DIFF
--- a/.changesets/1788729056-fix-window-reveal-a-cold-path-window-even-when-cef-views-mis-coej.md
+++ b/.changesets/1788729056-fix-window-reveal-a-cold-path-window-even-when-cef-views-mis-coej.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(window): reveal a cold-path window even when CEF Views misreports it visible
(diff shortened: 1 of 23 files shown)


a5af:
@codex review

PR #3106, opened 2026-09-09
Camper@narko: fix(session): flush the session snapshot on OS shutdown

GenericAgentX-asaf:
## Summary

Phase 0 of `SPEC_CONTINUOUS_SESSION_PERSISTENCE_2026_09_08.md` (merged in #3098). Closes the gap where **restarting the machine silently loses your workspace, while closing the app preserves it.**

`WM_QUERYENDSESSION` and `WM_ENDSESSION` appeared **nowhere** in the Rust sources. Those are the Windows notifications that give an application its one chance to persist before the OS terminates it. Without a handler, Windows just kills the process — `backend_close_window` is never called, so `save_last_session_snapshot` never runs.

**Confirmed against a real incident, not inferred.** The instance killed by a reboot had:

```
session:last_topology  => ABSENT      (snapshot never written)
windowids              => ['d354d7fa…']  (never emptied)
db_window: 1  db_workspace: 1  db_tab: 3  db_block: 22
```

`windowids` never emptied ⇒ the `is_empty()` gate that triggers the snapshot never fired, and `delete_workspace` never ran — the workspace was still sitting there frozen mid-session.

## Changes

**srv — `window.SaveSessionSnapshot`** (`session_restore.rs`, dispatched in `window.rs`, metadata in `backend/service.rs`)

Snapshot-only; **never** runs `delete_workspace`. Tearing down during an OS shutdown is both pointless (the OS reclaims everything) and actively harmful — that saga's 5s-per-shell `KILL_GRACE_SECS` does not fit the OS's shutdown budget. Takes no arguments and resolves the workspace from its own state, because the caller is a native wndproc with no window-id/label mapping in hand.

**cef — session-end wndproc hook** (`wndproc.rs`, installed in `lifecycle.rs`, helper in `helpers.rs`)

Mirrors the two hooks already in that file (`install_top_level_focus_restore_hook`, `close_routing_wndproc`): observer-passthrough, AppState via the same `OnceLock` static handoff, pruned on `WM_NCDESTROY` *after* passthrough so the original proc still sees it.

- Flushes on **`WM_QUERYENDSESSION`** — the roomier of the two windows, since `WM_ENDSESSION` can be followed by termination almost immediately.
- Flushes again on **`WM_ENDSESSION`** (wParam TRUE) for paths that skip a query round. Free once the flag is set.
- **Never returns 0** from the query arm — that vetoes the shutdown, which isn't ours to decide. `ShutdownBlockReasonCreate` is the sanctioned way to ask for more time.
- Deduplicated process-wide via `compare_exchange`, since session-end messages arrive per top-level window and several windows must not burn a bounded budget.
- Runs **inline on the wndproc thread**, not a background one — the process may die the moment it returns, so handing the work off would race the kill. 1500ms timeout against loopback (shorter than the close path's 2000ms) to stay inside the OS budget.

## Tests

Four new, covering the properties that matter — all 14 `session_restore` tests pass:

| Test | Asserts |
|---|---|
| `shutdown_flush_writes_a_snapshot_without_a_close` | the core fix — a snapshot exists with no `CloseWindow` |
| `shutdown_flush_leaves_the_workspace_intact` | no teardown; tabs survive untouched |
| `shutdown_flush_is_idempotent` | a second flush cannot corrupt the snapshot |
| `shutdown_flush_survives_a_missing_workspace` | quiet success, not an error, on a path that runs while the machine is going down |

## What is NOT verified

**The actual OS-shutdown path cannot be tested here** — it requires a real reboot, and the wndproc code is `#[cfg(target_os = "windows")]` unsafe subclassing that only executes during session-end. `cargo check` passes for both crates and the srv logic is unit-tested, but the end-to-end behaviour (Windows actually delivering the message, the flush completing inside the OS budget) needs a manual reboot-and-relaunch to confirm. Worth doing before relying on it.

Scope is Phase 0 only — continuous auto-save (Phase A), the clean-exit marker (Phase B), and shutdown hardening (Phase C) are separate.

### DIFF
--- a/.changesets/1788929003-fix-session-flush-the-session-snapshot-on-os-shutdown-wm-que-qda9.md
+++ b/.changesets/1788929003-fix-session-flush-the-session-snapshot-on-os-shutdown-wm-que-qda9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(session): flush the session snapshot on OS shutdown (WM_QUERYENDSESSION/WM_ENDSESSION)
--- a/agentmux-cef/src/client/helpers.rs
+++ b/agentmux-cef/src/client/helpers.rs
@@ -589,3 +589,106 @@ pub(crate) fn register_ipc_with_backend(
         ),
     }
 }
+
+/// Flush the session snapshot to srv, synchronously, for an OS shutdown.
+///
+/// `SPEC_CONTINUOUS_SESSION_PERSISTENCE_2026_09_08.md` Phase 0. Windows sends
+/// `WM_QUERYENDSESSION`/`WM_ENDSESSION` to give an application its one chance
+/// to persist before the OS terminates it; AgentMux handled neither, so a
+/// restart never reached `CloseWindow` and the session snapshot was never
+/// taken. This is that missing write.
+///
+/// Deliberately different from `backend_close_window` in three ways:
+///
+/// 1. **Snapshot-only.** Calls `window.SaveSessionSnapshot`, which never runs
+///    `delete_workspace`. Tearing down during an OS shutdown would be both
+///    pointless (the OS reclaims everything) and harmful (that saga's
+///    5s-per-shell grace does not fit the shutdown budget).
+/// 2. **Runs inline on the wndproc thread, not a background thread.** The
+///    caller is inside `WM_ENDSESSION`, and the process may be terminated the
+///    moment that returns — handing the work to another thread would race the
+///    kill and usually lose. Blocking here is the point.
+/// 3. **Shorter timeout.** Windows gives a shutting-down app a bounded window
+///    (historically ~5s) and kills apps that overstay it. 1500ms leaves room
+///    for the rest of teardown while still being ample for one local
+///    loopback write — this is srv on 127.0.0.1, not a network call.
+///
+/// Best-effort by design: every failure path logs and returns rather than
+/// propagating. A snapshot that cannot be written must never be able to wedge
+/// the shutdown it exists to protect.
+pub(crate) fn backend_save_session_snapshot(web_endpoint: &str, auth_key: &str) {
+    use std::io::{Read, Write};
+
+    let addr_str = web_endpoint
+        .trim_start_matches("http://")
+        .trim_start_matches("https://");
+    let addr: std::net::SocketAddr = match addr_str.parse() {
+        Ok(a) => a,
+        Err(e) => {
+            tracing::warn!(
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

a5af:
@codex review

GenericAgentX-asaf:
All three findings addressed in 6f06580f5.

**P1 — you were right, and it was a real bug.** `SESSION_END_FLUSHED` latched for the process lifetime, so a vetoed-then-retried shutdown (normal Windows behaviour) would silently skip the flush on the *real* shutdown, losing everything changed since the cancelled attempt. `WM_ENDSESSION` with wParam FALSE now resets the latch, making it per-*attempt* rather than per-process. My original comment literally named the cancellation case and waved it off as harmless without considering state drift — that was the error.

**P2 (arg_names)** — now `vec![]`, matching every other zero-arg method in that file.

**P2 (parse_web_endpoint)** — now reuses the existing helper, dropping ~12 duplicated lines.

**One thing worth flagging honestly:** the P1 edit in `wndproc.rs` appeared in my worktree from outside this session — my own scripted edit asserted and wrote nothing, and the committed wording differs from what I had drafted. Most likely another agent acting on this same review (today's SSH mesh work gave peers write access to this host). Rather than merge something I didn't write unexamined, I verified it independently: the logic is correct, both crates compile, and all 14 `session_restore` tests pass. Calling it out so the provenance is on record rather than silently absorbed.

**Still not verified, unchanged from the PR description:** the reset path lives in an unsafe Windows wndproc with no test harness, and the end-to-end OS-shutdown behaviour needs a real reboot to confirm.

a5af:
@codex review

GenericAgentX-asaf:
@codex review

All three findings addressed in 9eec3d6a5:

- **P1 (wndproc.rs)** — `backend_save_session_snapshot` now returns whether the save actually persisted; `flush_session_snapshot_once` re-arms `SESSION_END_FLUSHED` on any failure path (unreachable srv, write failure, missing AppState), not just the cancellation case the prior commit fixed.
- **P2 (helpers.rs)** — now parses the response body's `success` field instead of trusting the HTTP status line, since `handle_service` returns 200 with `success: false` for handler-level failures.
- **P2 (session_restore.rs)** — `current_workspace_id` now walks `windowids` instead of only checking the first entry. Added a regression test (`shutdown_flush_skips_a_stale_leading_window_id`) exercising a stale leading id with a live workspace behind it.

All 15 session_restore tests pass (14 existing + 1 new), both crates compile clean.

a5af:
@codex review

PR #3107, opened 2026-09-09
Manoz@Area54: refactor(srv): migrate agentdefcreatefromtemplate + forkagentdefinitionsuggest to register_typed (Phase 2 batch 2)

GenericAgentX-asaf:
## Summary

Second batch of Phase 2's domain-by-domain migration, on the machinery from #3074/#3078: **`agentdefcreatefromtemplate` and `forkagentdefinitionsuggest` migrated to `register_typed`**, 4 more RPC types generated and gated. Follow-up to #3074, #3078.

## Why these two, and not more from the same file

`template.rs` has five named-request handlers. Two are migrated here; three are deliberately left alone, and the reason is a wire-format constraint worth stating precisely, because it's what determines every future candidate:

**`register_typed` always sends a response** (`Ok(Some(value))`). The engine's `RpcMessage.data` field is `#[serde(default, skip_serializing_if = "Option::is_none")]` — confirmed by reading `rpc_types/mod.rs` — so `Ok(None)` (no response frame's data field at all) and `Ok(Some(Value::Null))` (`"data": null` present) are **different bytes on the wire**. That rules out:

- **`agentdefhide`/`agentdefunhide`** — already migrated in #3074, always return `Some`. ✓ (unaffected by this PR, listed for context)
- **`deleteagent`** — returns `Ok(None)` on every path. Left on `register_handler`.
- **`getagentcontent`** — returns `Ok(content.map(...))`, an `Option<Value>` that is conditionally `None` depending on business logic (content not found), not a fixed shape. Left on `register_handler`.
- **`renameagentdefinitiontitle`** — always returns `Some`, so it *could* migrate, but its response is `AgentDefinition`, the storage-level entity used throughout the codebase, not a small purpose-built `rpc_types` struct. Deriving `TS` on it is a bigger, more consequential change than this batch's scope; left for its own PR.

`agentdefcreatefromtemplate` and `forkagentdefinitionsuggest` were verified individually to always return `Ok(Some(fixed_shape))` on every success path (traced every `return Err(...)` and the one success tail in each) before migrating.

## What changed

- `ts_rs::TS` derived on `CommandAgentDefCreateFromTemplateData`, `AgentDefCreateFromTemplateResult`, `CommandForkAgentDefinitionSuggestData`, `ForkAgentDefinitionSuggestResult`.
- Both handlers converted from `register_handler` + manual `serde_json::from_value`/`to_value` to `register_typed`. Bodies otherwise unchanged.
- 4 new files under `frontend/types/rpc/`. The two files from #3074/#3078 are untouched — confirmed via `git status --porcelain -uall`, which showed only the 4 new files.
- A schema regression test asserting both new commands are recorded with their exact type names, and that `renameagentdefinitiontitle` (deliberately not migrated) stays absent.

## Verification

- Existing tests exercise both migrated handlers already: `agentdefcreatefromtemplate_resolves_provider_through_the_templates_bundle` asserts `resp.data` deserializes to `AgentDefCreateFromTemplateResult` and `resp.error.is_empty()` — this is exactly the wire-shape assertion that would catch a `register_typed` regression, and it passes unchanged.
- New test: `register_typed_records_the_second_batch_of_migrated_commands` — 1 passed.
- `cargo test --workspace -- --test-threads=1` — **4,139 passed, 0 failed**


a5af:
@codex review

a5af:
@codex review

GenericAgentX-asaf:
Fixed in `ff3b3ed2`, and it's a real one — thanks for catching it.

Confirmed against the hand-written stub it replaces: `identity_id?`, `memory_id?`, `agent_type?`, `model_vendor_base_url?` are all optional there, matching the four `#[serde(default)]` fields on the Rust side. The generated type made all four required.

The fix isn't just an attribute, though — worth flagging why. `ts-rs`'s `#[ts(optional)]` only accepts `Option<T>` (checked against the installed `ts-rs-macros-10.1.0` source rather than assumed). `model_vendor_base_url` was already `Option<String>`, so the attribute alone fixed it. `identity_id`/`memory_id`/`agent_type` were plain `String` — the correct fix there is changing them to `Option<String>`, which is what 'optional' actually means, not a workaround around the macro's restriction. `#[serde(default)]` already made a missing key deserialize to `String::default()` (`""`); `Option<String>`'s default deserializes it to `None` instead — a strict behavioral match, read back with `.unwrap_or_default()` at both call sites to preserve the exact prior `String` semantics. I confirmed this request type is constructed nowhere else in the codebase and no test sends these fields explicitly, so the change surface is exactly the one handler.

Regenerated: `identity_id?: string, memory_id?: string, agent_type?: string, model_vendor_base_url?: string` — byte-matching the hand-written stub's optionality on every field now.

`cargo test -p agentmux-srv agentdefcreatefromtemplate`: 3/3. Workspace: 4,139 passed, 0 failed.

PR #3113, opened 2026-09-09
docs(cef-build): fix a patcher.py invocation that silently applies zero patches

a5af:
## The bug

Both CEF build docs instruct:

```
python3 cef/tools/patcher.py --root-dir=cef
```

**`patcher.py` has no `--root-dir` flag.** Verified against `tools/patcher.py`'s own `OptionParser` in a real 7977 checkout — it accepts exactly `--patch-file` and `--patch-dir`, and in normal mode takes **no arguments** (it reads `cef/patch/patch.cfg` itself).

## Why this matters more than a typo

`STATUS_CEF_PROPRIETARY_CODECS_MACOS_2026_07_27.md` records `rebuild-mac-cef.sh` making this exact call, the failure being **swallowed as a non-fatal `WARN:`**, and the script continuing with **zero of 112 patches applied**. It was caught only because someone manually noticed `chromium/src` was a clean tree before starting the long build. Had it not been, that would have been an unpatched framework shipped as patched.

The Windows doc was already self-contradictory: its gotchas section (line 33) says the flag doesn't exist, while line 144 still told you to use it.

## Fix

Both docs now run it from inside `cef/` with no arguments, and follow with:

```
git status --porcelain | head -5
```

because **a clean tree at that point is the silent-failure signature, not success.** That check is the thing that catches this class of failure before six hours of build time.

## What I deliberately did NOT change

Three other `--root-dir` occurrences invoke **`translator.py`**, which *does* accept `--root-dir` — also verified against its `OptionParser`. Blanket-replacing every `--root-dir` in these docs would have broken working commands. Only the two `patcher.py` calls were wrong.

## Also in here

Spec status update: **Phase B's port is complete — 18 of 18 files**, done as real per-file 3-way merges (base = upstream 7778, ours = fork, theirs = upstream 7977), all clean with zero conflicts. Both previously-drifted files merged cleanly because our changes and upstream's occupy different regions.

That also closes patch #4's "partially verified" note — its five-commit CEF-side transparency cascade is ported, including `render_manager.cc`'s renderer half. **Ported is not built:** nothing here is compiled, and `agentmuxai/cef` still has no CI.

Context: tracking issue #3108. A Windows 152 build is running on claudius now, which is what surfaced the doc bug.

### DIFF
--- a/docs/cef-build/build-patched-cef-windows.md
+++ b/docs/cef-build/build-patched-cef-windows.md
@@ -141,7 +141,20 @@ Set-Location "$HOME\cef-build\chromium_git\chromium\src"
 # Reset to clean state -- patcher is NOT idempotent; double-apply produces .rej files
 git reset --hard HEAD
 Get-ChildItem -Recurse -Filter "*.rej" | Remove-Item
-python3 cef\tools\patcher.py --root-dir=cef
+
+# patcher.py takes NO arguments in its normal mode — it reads
+# cef/patch/patch.cfg itself, and accepts ONLY --patch-file / --patch-dir
+# (verified against tools/patcher.py's own OptionParser). Passing --root-dir
+# makes it fail, and a wrapper script has swallowed exactly that as a
+# non-fatal WARN and continued — applying ZERO of 112 patches and nearly
+# shipping an unpatched framework. Run it from inside cef/ with no args:
+Push-Location cef
+python3 tools\patcher.py
+Pop-Location
+
+# Verify it actually did something. A CLEAN tree here means zero patches
+# applied — that is the silent-failure signature, not success.
+git status --porcelain | Select-Object -First 5
 ```
 
 If you see "class member cannot be redeclared" compile errors later, the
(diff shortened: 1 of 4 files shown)


a5af:
@codex review

a5af:
This P2 is already fixed on the commit it reviews (`b3104727`).

Verified against the pushed blob, not my working copy:

```
$ gh api .../SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md?ref=b3104727     | base64 -d | grep -cE "3 of 18|3/18|must finish before Phase D"
0
```

Line ~138 there reads **"Phase B's port is COMPLETE — 18 of 18 fork-modified CEF files"**.

Codex raised the same issue validly against the *previous* commit; `b3104727` is the fix for it, and the audit I did then found 14 such stale claims across both docs rather than the one reported. No change needed here.

**Build status update, since it's the real test of these docs:** the corrected `patcher.py` invocation from this PR just ran against real Chromium 152 on claudius — **117 patches total, 117 applied, 0 skipped, 0 failed**, including both AgentMux patches. The `git status` check this PR adds confirms 441 modified tracked files. For contrast, the drifted-tree state the Windows doc warns about produced *55 failures* — so both the fix and the verification step are doing their job.

a5af:
All four checks green. The last review flagged a P2 that was already fixed on the commit it reviewed (evidence above), and the subsequent reconciliation push reset the approval state, so this now needs a fresh look at `b3104727`.

PR #3114, opened 2026-09-09
Manoz@Area54: refactor(srv): migrate deleteagentinstance + hidenamedagent to register_typed (Phase 2 batch 3)

GenericAgentX-asaf:
## Summary

Third batch of Phase 2's domain-by-domain migration, on the machinery from #3074/#3078/#3107. **`deleteagentinstance`** (`instance.rs`) and **`hidenamedagent`** (`identity.rs`) migrated to `register_typed`, 4 more RPC types generated and gated.

## What's new about this batch

The prior two batches migrated commands whose response was already a named `rpc_types` struct. These two weren't — they answered with an anonymous `json!({"deleted": deleted})` / `json!({"hidden": hidden})`. Two small named response types were added (`DeleteAgentInstanceResult`, `HideNamedAgentResult`) so the generator has something to name; this is the same `{flag: bool}` shape as the already-generated `AgentDefHideResult` from #3074.

Both commands were verified individually before migrating, using the same checklist as batch 2:

- **Always `Ok(Some(fixed_shape))`** on success — no `Ok(None)` path, no `Option<Value>` conditional on business logic.
- **Requests are the simplest possible shape** — both take `{ id: String }`.

## What's deliberately left in `instance.rs` / `identity.rs`

`listagentinstances`, `getagentinstance`, `createagentinstance`, `updateagentinstance` all still return the storage-level `AgentInstance` entity (or a `Vec<AgentInstance>`), same reasoning as `renameagentdefinitiontitle` in batch 2 — bigger blast radius than this batch's scope, left for their own PR. `listnamedagents` is untouched too (returns `Vec<NamedAgentRow>`, a join/view type, not migrated here). A new schema test in each file asserts the migrated command is present with its exact type names **and** that its non-migrated siblings stay absent from the schema.

## Wire-shape verification

Both generated response types are `{ deleted: boolean, }` / `{ hidden: boolean, }` — checked directly against the hand-written stubs they replace (`DeleteAgentInstanceCommand`/`HideNamedAgentCommand` in `frontend/app/store/rpc-api/agent.ts`), which already declare exactly `Promise<{ deleted: boolean }>` / `Promise<{ hidden: boolean }>`. Requests match too (`{ id: string }` both sides).

## Verification

- `git status --porcelain -uall -- frontend/types/rpc` after regenerating showed exactly the 4 new files — the six files from prior batches untouched.
- Existing tests exercising both handlers pass unchanged (they assert on `resp.data`, which is exactly what would break if `register_typed`'s wrapper changed the wire shape).
- Two new schema tests: `register_typed_records_deleteagentinstance`, `register_typed_records_hidenamedagent` — both assert exact type names and that the non-migrated sibling commands in the same file are absent.
- `cargo test --workspace -- --test-threads=1` — **4,145 passed, 0 failed**


a5af:
@codex review

PR #3115, opened 2026-09-09
Manoz@Area54: refactor(srv): migrate deletememory + deletesystemmemory to register_typed (Phase 2 batch 4)

GenericAgentX-asaf:
## Summary

Fourth batch of Phase 2's domain-by-domain migration. **`deletememory`** and **`deletesystemmemory`** migrated to `register_typed`, 2 more RPC types generated and gated (a request and a response, each shared by both commands).

## Why one response type for two commands

Both handlers already shared `CommandDeleteMemoryData` as their request (pre-existing, not introduced by this PR) — same `{ id: String }` shape, different store method underneath (`bundle_memory_delete` vs `bundle_memory_delete_system`). Their responses were two separate anonymous `json!({"deleted": deleted})` calls with the identical shape. Following the existing request-sharing precedent, one new `DeleteMemoryResult { deleted: bool }` replaces both — matching how `agentdefhide`/`agentdefunhide` already share `AgentDefHideResult` in #3074.

## What's deliberately left in `memory.rs`

`listmemories`, `getmemory`, `upsertmemory`, `reorderglobalbrain`, `upsertsystemmemory` are untouched — all either answer with the `Memory` storage entity or (`reorderglobalbrain`) a shape not yet worth a named type on its own. Same reasoning as prior batches. A schema test asserts both migrated commands are present with exact type names and that the five siblings stay absent.

## Test coverage: this batch had none before

Unlike the first three batches, `memory.rs` had **zero** existing tests for either command — nothing indirectly validated the wire shape the way `agentdefcreatefromtemplate`'s test did in batch 2. So this PR adds real coverage, not just a schema assertion:

- `deletememory_deletes_a_real_row_and_reports_it` — seeds a `Memory` row via `id_store.bundle_memory_upsert`, sends the RPC through `engine.handle_message`, asserts `resp.data` deserializes to `DeleteMemoryResult { deleted: true }`, and that the row is actually gone from the store afterward.
- `deletememory_reports_false_for_an_unknown_id` — same path, no such row, asserts `{ deleted: false }` (not an error — matches the pre-migration behavior of `bundle_memory_delete` returning `Ok(false)`).
- `deletesystemmemory_deletes_a_real_system_row_and_reports_it` — same shape via `bundle_memory_upsert_system` / `bundle_memory_delete_system`.
- `register_typed_records_both_delete_commands_sharing_one_response_type` — schema assertion.

## Wire-shape verification

`{ id: string }` request and `{ deleted: boolean }` response, checked directly against `frontend/app/store/rpc-api/memory.ts`'s `DeleteMemoryCommand`/`DeleteSystemMemoryCommand`, which already declare exactly these shapes.

## Verification

- `git status --porcelain -uall -- frontend/types/rpc` after regenerating showed exactly 2 new files.
- `cargo test -p agentmux-srv delete_memory_tests` — 4 passed.
- `cargo test --workspace -- --test-threads=1` — **4,151 passed, 0 failed**


a5af:
@codex review

PR #3117, opened 2026-09-09
fix(cef): linux userns-sandbox probe false-positives on AppArmor hosts

a5af:
## Summary
- `probe_userns_available()` reports "sandbox available" on Ubuntu hosts with the AppArmor unprivileged-userns restriction, because `unshare(CLONE_NEWUSER)` alone always succeeds under that policy -- the restriction denies `CAP_SYS_ADMIN` inside the new namespace instead, confirmed via `auditd` on the affected VM.
- Result: the existing sandbox-recovery dialog (PR #2783) never fires; CEF falls through into its real (blocked) sandbox init and crash-loops on Chromium's own `FATAL: No usable sandbox!` check.
- Fix: the probe now also writes `uid_map` after `unshare()` and only reports success if that succeeds too -- the actual capability AppArmor's policy gates, and the one CEF's own sandbox setup needs.
- Root-caused and verified live on an Ubuntu 24.04 VM (kernel restriction on, no AppArmor profile installed yet) via `auditd` audit records and a standalone reproduction (`unshare --user --map-root-user` fails on the `uid_map` write; the shipped binary's own `--internal-probe-userns` mode returned success despite that). See `docs/incident/INCIDENT_2026_09_09_LINUX_SANDBOX_PROBE_FALSE_POSITIVE.md` for full evidence.
- No change to the AppArmor profile text or the dialog/pkexec fix flow -- both were already correct; only detection was wrong.

## Test plan
- [ ] CI build (Linux)
- [ ] Deploy resulting AppImage to the affected VM and confirm the recovery dialog now fires on first launch, and that accepting "Fix it now" installs the profile and the app launches sandboxed afterward

### DIFF
--- a/.changesets/1788960467-fix-cef-linux-userns-sandbox-probe-false-positives-on-apparm-o68w.md
+++ b/.changesets/1788960467-fix-cef-linux-userns-sandbox-probe-false-positives-on-apparm-o68w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): linux userns-sandbox probe false-positives on AppArmor hosts
--- a/agentmux-cef/src/linux_sandbox.rs
+++ b/agentmux-cef/src/linux_sandbox.rs
@@ -138,6 +138,17 @@ mod imp {
     /// — it never spawns the async runtime, CEF, or any additional thread;
     /// this function is the entire lifetime of the process.
     pub fn run_internal_userns_probe_and_exit() -> ! {
+        // Must be read BEFORE unshare(): once inside the new namespace,
+        // getuid() returns the overflow UID (65534, "nobody") until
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
Addressed both findings in 108dd5eb9:
- codex P1 (uid captured after unshare -> overflow uid, always-fails write): fixed by capturing `getuid()` before `unshare()`. Verified on the affected VM: post-unshare `getuid()` does return 65534, and with the correct real uid the write still (correctly) fails while AppArmor blocks CAP_SYS_ADMIN.
- reagent P1 (missing changeset): added via `task changeset -- patch`.
- doc-status gate: added the required Status: line to the incident doc.

PR #3119, opened 2026-09-09
fix(build): fail bundle:* loudly on stub ANGLE GL libraries

Agent3-asaf:
## Summary

- `bundle:windows`/`bundle:darwin`/`bundle:linux` only checked that `libEGL`/`libGLESv2` *exist* before copying them into the shipped runtime — never that they're genuine ANGLE builds.
- A locally-rebuilt CEF tree (`~/cef-build`) can silently produce tiny placeholder libs that pass that check but crash the GPU process at init (`eglGetProcAddress not found`, `ui/gl/init/gl_initializer_win.cc:101`), fail the SwiftShader software fallback too, and leave GFX permanently **"off"** in the status bar — with zero build-time signal.
- Reproduced live: a running instance's bundled `libEGL.dll`/`libGLESv2.dll` were both exactly 470016 bytes, imported only `KERNEL32.dll`, and contained none of ANGLE's EGL/GLES export symbols (real ANGLE builds are multi-MB). Traced to the source CEF tree at `~/cef-build/.../Release_GN_x64`, where those two files were touched by a recent (incomplete?) rebuild while `libcef.dll` itself was untouched.

## Change

- Add `scripts/verify-angle-libs.sh`: checks a GL lib (when present) is >1MB and exports `eglGetProcAddress`/`glGetString`; skips silently if the file is simply absent (a separate, already-handled case). Same fail-loud posture as the existing `verify-cef-version.sh` guard.
- Wire it into all three `bundle:*` tasks, right where each resolves its CEF source directory, before the libs get copied into `dist/cef/`.

## Test plan

- [x] Ran the new script directly against the actual broken DLLs in `~/cef-build/chromium_git/chromium/src/out/Release_GN_x64` — fails loudly with the expected diagnostic (confirmed the exact real-world regression this guards against).
- [x] Ran it against a missing dir/files — passes (skip, not this script's concern).
- [x] Ran it against synthetic >1MB files with/without the expected symbol strings — passes / fails as expected.
- [x] `task --list` still parses `Taskfile.yml` cleanly after the edits.
- [ ] Not run end-to-end through a full `task bundle:windows`/`task package` (would require rebuilding `agentmux-cef` from a clean `target/release`, which wasn't necessary to validate the guard logic itself — the script is invoked with the exact same args used in the task).

### DIFF
--- a/.changesets/1788960693-fix-build-fail-bundle-loudly-on-stub-angle-gl-libraries-inst-noly.md
+++ b/.changesets/1788960693-fix-build-fail-bundle-loudly-on-stub-angle-gl-libraries-inst-noly.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(build): fail bundle:* loudly on stub ANGLE GL libraries instead of shipping a broken GFX-off build
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -706,6 +706,12 @@ tasks:
                 # window). Fail loudly here instead of shipping a broken build.
                 # See docs/specs/SPEC_WINDOWS_CEF_BUNDLE_VERSION_INTEGRITY_2026_06_03.md.
                 bash scripts/verify-cef-version.sh "$cefDir" || exit 1
+                # Guard: a locally-rebuilt CEF tree can silently produce stub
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #3120, opened 2026-09-09
fix(srv): jekt notifications addressed to the stable AGENTMUX_AGENT_ID now resolve

a5af:
## Summary
- PR #3100 documented but never fixed jekt-delivery failure: `ReactiveHandler`'s registry holds one key per block, and `input.rs`'s Register-tail re-keys it to the live, renameable display name on every turn -- evicting the stable `AGENTMUX_AGENT_ID` a spawn registered it under moments earlier.
- GitHub PR-body tags embed that stable ID specifically because it survives renames -- a jekt tagged with it had nothing left to resolve to after the agent's first turn.
- Reproduced live this session: `AGENTMUX_AGENT_ID=agentg` in env, but the live WAN subscription is `claude` (the current display name).
- Fix: a second, independent alias registry (`Handler::alias_to_block`) that `spawn_process` also populates with `AGENTMUX_AGENT_ID`, never evicted by the ordinary per-turn re-key. `inject_message_inner` falls back to it on a primary-registry miss. The existing #2695 recipient-identity check gets a matching `stable_agent_identity_confirmer` (backed by a new, write-once `Controller::stable_agent_id()`) so an alias-resolved delivery isn't rejected for legitimately not matching the live display name.
- Scoped to `PersistentSubprocessController` only (the stream-json/ACP agent path). No change to `ShellController` or the ordinary rename/#2697 behavior.
- See `docs/incident/INCIDENT_2026_09_09_JEKT_STABLE_ID_ALIAS.md` for full root cause. Also includes `INCIDENT_2026_09_09_JEKT_CASE_MISMATCH_SUBSCRIPTION.md`, a research-only doc from checking (and disproving) a case-sensitivity hypothesis along the way.

## Test plan
- [x] New unit tests: alias resolves via `inject_message`, alias survives a primary-key change, `unregister_block` cleans up the alias, dual-confirmer accepts a stable-only match, dual-confirmer still rejects a target matching neither identity.
- [ ] CI build
- [ ] Verify against a live GitHub PR-review jekt end-to-end once merged and deployed

### DIFF
--- a/.changesets/1788961597-fix-srv-jekt-notifications-for-a-renamed-agent-s-stable-agen-lmfg.md
+++ b/.changesets/1788961597-fix-srv-jekt-notifications-for-a-renamed-agent-s-stable-agen-lmfg.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): jekt notifications for a renamed agent's stable AGENTMUX_AGENT_ID now resolve
--- a/agentmux-srv/src/backend/blockcontroller/mod.rs
+++ b/agentmux-srv/src/backend/blockcontroller/mod.rs
@@ -249,6 +249,25 @@ pub trait Controller: Send + Sync {
     /// override [`agent_id`](Controller::agent_id) either.
     fn set_agent_id(&self, _id: Option<String>) {}
 
+    /// This block's STABLE jekt identity — `AGENTMUX_AGENT_ID` as captured
+    /// once at spawn, never refreshed. Distinct from [`agent_id`](Controller::agent_id),
+    /// which tracks the LIVE, renameable display name and is deliberately
+    /// refreshed every turn (see that method's doc comment on #2697).
+    /// AGENTMUX_AGENT_ID is embedded verbatim in PR-body tags specifically
+    /// because it does NOT change if the agent is later renamed — but
+    /// `ReactiveHandler`'s registry only tracks one key per block, and
+    /// every turn's `agent_id()` refresh evicts whatever this block was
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

a5af:
Fixed the confirmed leak in de207163e: `unregister_agent` now also removes the block's alias entry (mirroring `unregister_block`'s cleanup, which this PR already had). Addresses both the reagent P1 (dangling alias / unbounded growth) and codex's matching P2 finding on `handler.rs:277`.

On codex's other P1 (`persistent.rs:2848` — alias only populated in `spawn_process`, so a stable-ID jekt still fails in the post-restart, pre-lazy-spawn window): confirmed this is real and distinct from the primary bug this PR fixes. Deliberately NOT fixing it here — closing it means reading `AGENTMUX_AGENT_ID` from the block's *persisted* config at controller-construction time (before any process spawns), which is genuinely different code (controller lifecycle, not the registration/delivery path this PR touches) that this investigation didn't reach. Documented as an explicit known follow-up in `INCIDENT_2026_09_09_JEKT_STABLE_ID_ALIAS.md`'s new section rather than silently dropped. Given this file's incident history, I'd rather ship the confirmed, always-reproducing fix now and track the narrower restart-window gap separately than rush a second, riskier change into the same PR.

a5af:
Added test_handler_unregister_agent_also_removes_alias in c5304c2cd, mirroring the unregister_block coverage. Test-only change, doesn't affect the running Windows build's binary output.

PR #3121, opened 2026-09-09
Clare@starpower: docs(cef): branch model + upgrade practice for the CEF fork, all three platforms

AgentO-asaf:
## Why

We carry five AgentMux-specific changes on top of CEF and have **silently** lost pieces of that set twice. Neither failure produced an error, a conflict, or a failed build — you find out when a user reports a feature that used to work.

**Incident 1 (2026-07-01 → 2026-09-09).** The drag branch was merged into `fork/7778` via PR #3 on Jun 24, then kept receiving commits (PRs #4 and #5, Jul 1) that were never merged forward:

```
2720ba103 ──┬──► merged into fork/7778 as 2d7833dec (PR #3, Jun 24)
            │         └──► 5387e4974 (PR #7, Sep 8) ── fork/7778 tip
            │                                          has process_requirement, MISSING #4/#5
            └──► 7ba6b59f4 (PR #4, Jul 1) ──► 6c570e249 (PR #5, Jul 1) ── what SHIPPED
```

It cost nothing for two months because the July macOS framework was built from the drag branch tip directly. It bit when someone went to `fork/7778` for the P0 `process_requirement` fix — the obvious move, since that branch is the integration branch, is named for the Chromium version, and carried the newest commit. Building from it would have fixed the renderer crash-loop and regressed macOS window transparency at the same time.

**Incident 2 — found while writing this doc, still live.** The Chromium 152 port has already dropped the same work. `fork/agentmux/7977-drag-rightclick-and-transparency` ports `BeginWindowDrag`, `rwhv_background_opaque_check` and `views_caption_rightclick_passthrough`, but not the `blink_glue`/`mojom` half of transparency. Verified by identifier probe with the 7778 branch as positive control. **The 152 upgrade is currently set up to ship without transparency.**

## Root cause the doc names

Our changes live in **two layers with different mechanics**:

| | Layer A — Chromium-side patches | Layer B — libcef-side commits |
|---|---|---|
| Applied by | `patcher.py` at build time | nothing — just commits on a branch |
| Fails | loudly (`.rej`, non-zero exit) | **silently** — the tree still compiles |

Both incidents were Layer B. Nothing checks it, and the tree compiles fine without those commits because our additions are additive.

The 2026-07 gap is also a violation of a rule the repo already has in `CLAUDE.md` — *"Never reuse a branch after its PR is merged"* — which nobody had applied to the CEF fork.

## What's here

- `docs/cef-build/CEF_FORK_MAINTENANCE.md` — the standing practice: the two-layer model, the verified carry-set inventory (5 items, with per-platform scope), branch rules R1–R5, upgrade runbook, per-platform artifact verification, release-pinning practice P1–P3, checklists.
- `README.md` — the load-bearing half, so agents hit it without going looking.

## Verification

The doc's §5 block is **tested, not just written**. It reproduces both gaps independently and reports all-OK against a branch known to be complete:

| Branch | Result |
|---|---|
| `rebuild-7778-procreq-plus-transparency` (complete) | all 6 probes `OK` |
| `fork/7778` | `MISS rwhv_background_opaque_check`, `MISS transparency (blink_glue)`, `MISS transparency (mojom)` |
| `fork/agentmux/7977-drag-...` | `MISS transparency (blink_glue)` |

Testing it caught a real bug in my own script: written as `"$BR:libcef/..."` the shell eats the `:l` as a modifier and resolves `fork/7778ibcef/...`, producing a false MISS on the two probes that matter most. Fixed to `${BR}:` with a note explaining why, so it doesn't get "simplified" back.

Gates: `check-doc-status` ok (2 docs), `gen-docs-index --check` ok, `check-spec-citations` ok.

## Notes for the reviewer

- Docs only — no code paths touched.
- Platform scoping in the carry-set table was verified from the patch bodies, not assumed from names: `views_caption_rightclick_passthrough` is **Linux-only** (`window_event_filter_linux.cc`) despite the generic name.
- The `patcher.py --root-dir` trap is restated here rather than only in #3113, because the failure mode (applies **zero of 112** patches and builds anyway) is what makes the whole class of bug invisible.

## Follow-ups this does NOT do

1. **`fork/7778` is still incomplete upstream** — it has `process_requirement` but not the transparency commits. I merged them locally for the in-flight build; the fork itself still needs the merge. Wanted a decision before pushing to a shared repo.
2. **The 152 port needs the transparency half added** before that upgrade ships (#3103).
3. Suggested: have the release job assert all three pins resolve to the same `agentmuxai/cef` commit, which would make P1–P3 mechanical instead of convention.

### DIFF
--- a/.changesets/1788964010-docs-cef-branch-model-upgrade-practice-for-the-cef-fork-all--iqa5.md
+++ b/.changesets/1788964010-docs-cef-branch-model-upgrade-practice-for-the-cef-fork-all--iqa5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(cef): branch model + upgrade practice for the CEF fork, all three platforms
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -393,6 +393,14 @@ History: `docs/retro/retro-release-version-desync-2026-05-22.md` — PR #964 sil
 
 ## Git Workflow
 
+**Never reuse a branch after its PR is merged.** Once merged, that branch is
+dead — cut a new one from the target for follow-up work. Continuing to commit to
+an already-merged branch is how a change that *was* merged quietly stops being in
+the integration branch: the branch keeps moving, nothing merges it forward, and
+no build ever fails. This is not hypothetical — it cost two months and nearly
+shipped a macOS transparency regression in the CEF fork. See
+[docs/cef-build/CEF_FORK_MAINTENANCE.md](./docs/cef-build/CEF_FORK_MAINTENANCE.md) §1.1.
+
 ```bash
 # Create feature branch
 git checkout -b feature-name
--- a/README.md
+++ b/README.md
@@ -230,6 +230,56 @@ Local build outputs from `task package` on the host platform:
 
 Release artifacts (macOS DMG, Windows installer, Linux AppImage) are built by CI workflows in this repo. See [§Releases](#releases).
 
+## CEF Fork — read before touching the Chromium layer
+
+AgentMux ships a **fork of CEF** (`agentmuxai/cef`) carrying five AgentMux-specific
+changes. Every Chromium milestone upgrade must carry all five forward, for Windows,
+Linux and macOS together. We have lost pieces twice, both times **silently** — no
+error, no conflict, no failed build. Full practice:
+[docs/cef-build/CEF_FORK_MAINTENANCE.md](./docs/cef-build/CEF_FORK_MAINTENANCE.md).
+
+**Our changes live in two layers, and they fail differently:**
+
+| | Layer A — Chromium-side patches | Layer B — libcef-side commits |
+|---|---|---|
+| What | `cef/patch/patches/*.patch` + `patch.cfg` | edits to `cef/libcef/**`, `cef/include/**` |
+| Applied by | `patcher.py` at build time | nothing — they are just commits on the branch |
+| Fails | loudly (`.rej`, non-zero exit) | **silently** — the tree still compiles |
+
+Layer B is where both incidents happened. Nothing checks it.
+
+**The rules:**
+
+- **Build and release only from `fork/<milestone>`** (`fork/7778`, `fork/7977`) —
+  never from a feature branch, even if the feature branch looks newer.
+- **Never reuse a feature branch after its PR is merged** — start a new one off
+  `fork/<milestone>`. This is the existing workspace rule; it applies to the CEF
+  fork too, and violating it is exactly what caused the 2026-07 gap.
+- **Merge the whole carry-set before cutting a release.** A milestone's work is
+  split across several branches; merging one and not the others silently drops
+  features.
+- **Newer-looking is not superset.** Before building, this must print nothing:
+  ```bash
+  git log --oneline <last-shipped-commit> --not fork/<milestone>
+  ```
+  Anything it prints is in the shipped artifact but not in what you are about to
+  build — a regression you are about to ship.
+
+**Three traps that produce a wrong binary with no error:**
+
+1. `patcher.py` takes **no arguments** in normal mode (only `--patch-file` /
+   `--patch-dir`, and those take a patch *name*, not a path). Passing `--root-dir`
+   fails, and a wrapper has swallowed that as a `WARN` and built anyway — applying
+   **zero of 112 patches**. After running it, `git status --porcelain` in the
+   Chromium tree must show *hundreds* of modified files; a clean tree means zero
+   patches applied.
+2. Verify patch symbols with **full `nm`**. They are local, not exported —
+   `nm -gU` misses every one and reports a false negative.
+3. **All three platform pins live in one place** (`release.yml`'s
+   `cef-runtime-pins`) and must be bumped together, from **one** fork commit.
+   The job only cross-checks the leading milestone, so two tags can agree on
+   `148` and still come from different fork commits with different carry-sets.
+
 ## Version Management (Changesets)
 
 **Feature PRs add a changeset, not a version bump.** RFC #857 Phase 2.
--- a/README.md
+++ b/README.md
@@ -260,7 +260,9 @@ Layer B is where both incidents happened. Nothing checks it.
 - **Merge the whole carry-set before cutting a release.** A milestone's work is
   split across several branches; merging one and not the others silently drops
   features.
-- **Newer-looking is not superset.** Before building, this must print nothing:
+- **Newer-looking is not superset.** Before building, this must print nothing
+  (**same milestone only** — across an upgrade the carry-set is re-ported as new
+  commits, so use the carry-set gate instead):
   ```bash
   git log --oneline <last-shipped-commit> --not <remote>/<milestone>
   ```
--- a/README.md
+++ b/README.md
@@ -232,8 +232,9 @@ Release artifacts (macOS DMG, Windows installer, Linux AppImage) are built by CI
 
 ## CEF Fork — read before touching the Chromium layer
 
-AgentMux ships a **fork of CEF** (`agentmuxai/cef`) carrying five AgentMux-specific
-changes. Every Chromium milestone upgrade must carry all five forward, for Windows,
+AgentMux ships a **fork of CEF** (`agentmuxai/cef`) carrying four AgentMux-specific
+changes — 18 CEF source files plus 3 Chromium-side patches. Every Chromium
+milestone upgrade must carry all of them forward, for Windows,
 Linux and macOS together. We have lost pieces twice, both times **silently** — no
 error, no conflict, no failed build. Full practice:
 [docs/cef-build/CEF_FORK_MAINTENANCE.md](./docs/cef-build/CEF_FORK_MAINTENANCE.md).
--- a/README.md
+++ b/README.md
@@ -250,17 +250,18 @@ Layer B is where both incidents happened. Nothing checks it.
 
 **The rules:**
 
-- **Build and release only from `fork/<milestone>`** (`fork/7778`, `fork/7977`) —
-  never from a feature branch, even if the feature branch looks newer.
+- **Build and release only from the integration branch** — named for the
+  milestone alone (`7778`, `7977`), never from a feature branch, even if the
+  feature branch looks newer.
 - **Never reuse a feature branch after its PR is merged** — start a new one off
-  `fork/<milestone>`. See the rule under [Git Workflow](./CLAUDE.md#git-workflow);
+  `<milestone>`. See the rule under [Git Workflow](./CLAUDE.md#git-workflow);
   violating it is exactly what caused the 2026-07 gap.
 - **Merge the whole carry-set before cutting a release.** A milestone's work is
   split across several branches; merging one and not the others silently drops
   features.
 - **Newer-looking is not superset.** Before building, this must print nothing:
   ```bash
-  git log --oneline <last-shipped-commit> --not fork/<milestone>
+  git log --oneline <last-shipped-commit> --not <remote>/<milestone>
   ```
   Anything it prints is in the shipped artifact but not in what you are about to
   build — a regression you are about to ship.
--- a/README.md
+++ b/README.md
@@ -253,8 +253,8 @@ Layer B is where both incidents happened. Nothing checks it.
 - **Build and release only from `fork/<milestone>`** (`fork/7778`, `fork/7977`) —
   never from a feature branch, even if the feature branch looks newer.
 - **Never reuse a feature branch after its PR is merged** — start a new one off
-  `fork/<milestone>`. This is the existing workspace rule; it applies to the CEF
-  fork too, and violating it is exactly what caused the 2026-07 gap.
+  `fork/<milestone>`. See the rule under [Git Workflow](./CLAUDE.md#git-workflow);
+  violating it is exactly what caused the 2026-07 gap.
 - **Merge the whole carry-set before cutting a release.** A milestone's work is
   split across several branches; merging one and not the others silently drops
   features.
--- a/docs/cef-build/CEF_FORK_MAINTENANCE.md
+++ b/docs/cef-build/CEF_FORK_MAINTENANCE.md
@@ -6,7 +6,8 @@
 [build-patched-cef-windows.md](./build-patched-cef-windows.md),
 [build-patched-framework-macos.md](./build-patched-framework-macos.md)
 
-We maintain a fork of CEF carrying five AgentMux-specific changes. Every Chromium
+We maintain a fork of CEF carrying **four** AgentMux-specific changes — spanning
+18 CEF source files and 3 Chromium-side patches (§3). Every Chromium
 milestone upgrade has to carry that set forward, for Windows, Linux and macOS
 together. This doc is the standing practice for doing that without losing pieces.
 
@@ -112,8 +113,19 @@ diverge in behaviour.
 
 ## 3. The carry-set (canonical inventory)
 
-**18 hand-written CEF source files** differ between upstream 7778 and the fork,
-plus **3 Chromium-side patches**. Measured, not estimated:
+**Four features.** Two of them are Layer B only, one is Layer A only, and one
+spans both — which is why a feature count and a file count are different
+questions and both get stated here:
+
+| Feature | Layer | Files |
+|---|---|---|
+| Drag (`BeginWindowDrag`) | B | 3 |
+| Transparency | B + A | 15 + 1 patch |
+| Right-click passthrough | A | 1 patch (Linux) |
+| Process requirement | A | 1 patch (macOS 26) |
+
+That totals **18 hand-written CEF source files** differing between upstream 7778
+and the fork, plus **3 Chromium-side patches**. Measured, not estimated:
 
 ```bash
 git diff --name-only <upstream-base> <branch> -- \
@@ -458,9 +470,9 @@ distinguish "patched build" from "pristine Chromium build" on any platform.
 
 | Platform | Check |
 |---|---|
-| **macOS** | Frameless window drags from an HTCLIENT region (#1); window background is actually transparent, not white (#2+#3); renderer does not crash-loop on macOS 26 (#5); H.264/HEVC playback works (codec flags) |
-| **Linux** | Right-click in the caption area passes through (#4); drag (#1); H.264 playback (codec build) |
-| **Windows** | Drag (#1); transparency (#2) |
+| **macOS** | **Drag**: frameless window drags from an HTCLIENT region. **Transparency**: window background is actually transparent, not white — needs *both* layers, see §3. **Process requirement**: renderer does not crash-loop on macOS 26. Plus H.264/HEVC playback (codec flags). |
+| **Linux** | **Right-click passthrough**: right-click in the caption area passes through. **Drag**. Plus H.264 playback (codec build). |
+| **Windows** | **Drag**. **Transparency**. |
 
 ---
 
--- a/docs/cef-build/CEF_FORK_MAINTENANCE.md
+++ b/docs/cef-build/CEF_FORK_MAINTENANCE.md
@@ -22,7 +22,7 @@ reports a feature that used to work.
 
 ### 1.1 The forward-merge gap (2026-07-01 → 2026-09-09, found in production path)
 
-`agentmux/7778-drag-rightclick-and-transparency` was merged into `fork/7778` via
+`agentmux/7778-drag-rightclick-and-transparency` was merged into `agentmuxai/7778` via
 PR #3 on Jun 24. Then PRs #4 and #5 added two **more** commits to that same
 branch on Jul 1 — the renderer-side transparency work and the macOS
 white-substitution fix. Nobody merged the branch forward a second time.
@@ -30,9 +30,9 @@ white-substitution fix. Nobody merged the branch forward a second time.
 ```
 2720ba103  (Jun 24)  drag branch
     │
-    ├──► merged into fork/7778 as 2d7833dec   (PR #3, Jun 24 18:57)
+    ├──► merged into agentmuxai/7778 as 2d7833dec   (PR #3, Jun 24 18:57)
     │         │
-    │         └──► 5387e4974  (PR #7, Sep 8)  ── fork/7778 tip
+    │         └──► 5387e4974  (PR #7, Sep 8)  ── agentmuxai/7778 tip
     │                                             has process_requirement
     │                                             MISSING #4 and #5
     │
@@ -44,12 +44,12 @@ white-substitution fix. Nobody merged the branch forward a second time.
 
 For two months this cost nothing, because the July macOS framework was built from
 the drag branch tip directly (`cef HEAD: 6c570e249` in the build log), not from
-`fork/7778`. It only bit when someone went to `fork/7778` to pick up the
+`agentmuxai/7778`. It only bit when someone went to `agentmuxai/7778` to pick up the
 `process_requirement` P0 fix — the reasonable, obvious move. Every signal pointed
-that way: `fork/7778` is the integration branch, it is named for the Chromium
+that way: `agentmuxai/7778` is the integration branch, it is named for the Chromium
 version, and it carried the newest commit (Sep 8 vs Jul 1).
 
-**Newer-looking is not the same as superset.** Building from `fork/7778` would
+**Newer-looking is not the same as superset.** Building from `agentmuxai/7778` would
 have fixed the renderer crash-loop and simultaneously regressed macOS window
 transparency, with nothing anywhere reporting a problem.
 
@@ -82,7 +82,7 @@ So, as a rule:
 
 One genuine asymmetry does remain, and it is the R4 case rather than a defect:
 `agentmux_process_requirement` is absent from the drag branch because it lives on
-`agentmux/7977-process-requirement`. Both must be merged into `fork/7977`. That
+`agentmux/7977-process-requirement`. Both must be merged into `agentmuxai/7977`. That
 split is precisely what produced §1.1 on 7778.
 
 ---
@@ -138,6 +138,15 @@ Notes that have already caused confusion:
 
 ## 4. Branch model — the rules
 
+> **Notation.** The integration branch is named for the milestone alone --
+> `7778`, `7977` (`refs/heads/7778` on `agentmuxai/cef`). Where this doc writes
+> `agentmuxai/7778` that is `<remote>/<branch>`, not a branch called
+> `agentmuxai/7778`. An earlier revision wrote `agentmuxai/7778`, using one checkout's
+> local remote name; a reviewer reasonably read that as the branch name and
+> concluded the checkout commands were wrong. If the notation can mislead a
+> careful reader it can mislead an operator, so it is spelled out here.
+
+
 The root cause of §1.1 is branch reuse after merge. The repo-level rule for
 this lives in [`CLAUDE.md`](../../CLAUDE.md) under Git Workflow:
 
@@ -147,18 +156,18 @@ It was added by the same PR as this document — it was *not* previously written
 down anywhere in the repo, which is part of why §1.1 happened. It applies to
 `agentmuxai/cef` as much as to this repo.
 
-**R1 — One integration branch per Chromium milestone.** `fork/<milestone>`
-(`fork/7778`, `fork/7977`). This is the *only* branch anyone builds or releases
+**R1 — One integration branch per Chromium milestone.** `agentmuxai/<milestone>`
+(`agentmuxai/7778`, `agentmuxai/7977`). This is the *only* branch anyone builds or releases
 from. It is upstream CEF's branch plus our carry-set, nothing else.
 
 **R2 — Feature branches are single-use.** Once `agentmux/<ms>-<topic>` is merged
-into `fork/<ms>`, it is dead. Follow-up work starts a new branch off
-`fork/<ms>`, never off the merged branch. This alone would have prevented §1.1.
+into `agentmuxai/<ms>`, it is dead. Follow-up work starts a new branch off
+`agentmuxai/<ms>`, never off the merged branch. This alone would have prevented §1.1.
 
 **R3 — Never build or release from a feature branch.** The July build used
 `agentmux/7778-drag-rightclick-and-transparency` directly. It produced a correct
 artifact *by luck* — that branch happened to be ahead. Building only from
-`fork/<ms>` makes R1 self-enforcing: if the integration branch is incomplete,
+`agentmuxai/<ms>` makes R1 self-enforcing: if the integration branch is incomplete,
 you find out immediately rather than two months later.
 
 **R4 — Merge the whole carry-set before cutting a release, not piecemeal.** The
@@ -166,7 +175,7 @@ you find out immediately rather than two months later.
 `7977-process-requirement`. Merging one without the other reproduces §1.1
 exactly. Merge both, then verify §5.
 
-**R5 — `fork/<ms>` must never lose a carry-set item.** It is append-only with
+**R5 — `agentmuxai/<ms>` must never lose a carry-set item.** It is append-only with
 respect to §3. If an upgrade legitimately drops one (upstream absorbed it), that
 requires an explicit commit saying so, referencing the upstream change.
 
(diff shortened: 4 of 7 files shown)


a5af:
@codex review

AgentO-asaf:
Thanks — this was half right, and chasing it produced a better doc than I had.

**You're correct about bash.** `"$BR:libcef/..."` and `"${BR}:libcef/..."` do expand identically there, so "the shell eats `:l`" was wrong as written.

**But the behaviour is real in zsh**, which is the macOS default and what produced the original false MISS:

```
$ bash -c 'BR=fork/7778; printf "%s\n" "$BR:libcef/renderer/blink_glue.h"'
fork/7778:libcef/renderer/blink_glue.h        # identical to braced, as you said

$ zsh  -c 'BR=fork/7778; printf "%s\n" "$BR:libcef/renderer/blink_glue.h"'
fork/7778ibcef/renderer/blink_glue.h          # :l applied as a history modifier
```

zsh applies history-style modifiers to `$VAR:<letter>` for literal text. Verified with `BR=fork/7778`:

```
$BR:libcef/x   -> fork/7778ibcef/x     ( :l  lowercase )
$BR:head/x     -> forkead/x            ( :h  dirname   )
$BR:tail/x     -> 7778ail/x            ( :t  basename  )
$BR:upper/x    -> FORK/7778pper/x      ( :u  uppercase )
$BR:include/x  -> fork/7778:include/x  ( :i  not a modifier - safe )
$BR:patch/x    -> fork/7778:patch/x    ( :p  not a modifier - safe )
```

That bash is unaffected makes this **worse**, not benign: the snippet works for some readers and silently misreports for others, on the two probes that matter most. `:include` and `:patch` not being modifiers is why only the two `libcef` probes broke — which is exactly why it survived a first reading.

One thing worth recording: an early probe of mine appeared to show no modifier at all, because it built the suffix from a variable (`$BR:$w/x`). The modifier only applies to literal text, so the indirection masked it. If you reproduced via a variable rather than a literal, that would explain a clean bash-and-zsh result.

Fixed in 2c3c0a155 — the note now says zsh specifically, states plainly that bash is unaffected, and carries the verified table. The recommended code is unchanged.

AgentO-asaf:
All four findings were valid. Fixed in `773124d84`.

**ReAgent P2 — fabricated citation. Correct.** I cited "a rule this repo already has, in CLAUDE.md under Workspace Rules." No such rule or section exists here. The text I was recalling lives in the *agent workspace* `CLAUDE.md`, which repo contributors never see — I conflated the two files.

Rather than delete the citation I've added the rule to the repo's own `CLAUDE.md` under Git Workflow, where it belongs and where agents already look. The doc now cites that, and says plainly that it arrived with this PR and was not previously written down — which is part of why §1.1 happened at all.

**Codex P2 — stale 152 status. Correct, and my claim was wrong.** The probe was real but ran against a stale local ref: `agentmux/7977-drag-rightclick-and-transparency` had been **force-updated** (`d6fe3d449` → `55edc030a`). After `git fetch --prune`:

```
OK   libcef/renderer/blink_glue.h          OK   libcef/common/mojom/cef.mojom
OK   libcef/renderer/blink_glue.cc         OK   libcef/browser/browser_platform_delegate.cc
OK   libcef/renderer/browser_config.h      OK   libcef/browser/browser_info_manager.cc
OK   libcef/renderer/render_manager.cc
```

The 152 port is complete. You were also right that `REPORT_CEF_UPGRADE_PHASE_A_RECON_2026_09_08.md` and `SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md` already recorded this — I didn't consult them, and trusted my own probe because I'd run it myself.

§1.2 is now about that failure rather than the imaginary gap: re-fetch before judging, cross-check the milestone's own port records, and treat `MISS` as "absent from the ref I probed" — a weaker claim than "absent from the branch."

**Codex P1 — under-specified gate. Correct.** Probing 2 of 7 transparency files meant a partial port could pass while `render_manager.cc` or browser-side propagation was missing — the gate approving the exact regression it exists to catch. Now probes all seven as a unit plus a mandatory `git fetch --prune`. Expect **11 `OK`**; verified against a known-complete branch.

**Codex P1 — runbooks contradict R3. Correct.** Both `build-patched-libcef.md` and `build-patched-cef-windows.md` checked out the feature branch, reproducing the 2026-07 failure for anyone following them. Both now use the integration branch.

One deliberate choice worth surfacing: `fork/7778` is **currently incomplete** — it has `process_requirement` but not the transparency commits. So switching the runbooks to it without a gate would have been worse than leaving them alone. The verification step is the load-bearing half: it fails loudly and says not to build around it. The upstream fork merge is still pending a decision from the repo owner.

AgentO-asaf:
All three correct. Fixed in `7eb7a0c27`.

**P1 ×2 — probe count contradiction.** §6 step 4 and the pre-release checklist both still said "eight probes" after this PR expanded §5 from 2 named transparency probes to the full 7-file cascade. Both now say **11**, matching §5. The doc's gate criteria no longer disagree with the doc's own gate.

**P1 — macOS not brought into line.** Correct, and the sharpest of the three: a PR titled "all three platforms" that fixes two of them is worse than one that fixes none, because it looks done. `build-patched-framework-macos.md` named the dead feature branch in both the header and the release-notes template.

Worth noting Codex had reported macOS as delegating checkout to the Linux guide — it doesn't; it names the branch directly at line 28. Your read was the accurate one.

**Fixing macOS surfaced three more instances of the same class that neither review flagged:**

- `build-patched-libcef.md:18` — branch header
- `build-patched-libcef.md:237` — release-notes template
- `build-patched-cef-windows.md:246` — release-notes template

All corrected. Every remaining mention of the feature branch in `docs/cef-build/` is now a DO-NOT-USE warning or §1.1 history — none are instructions:

```
build-patched-framework-macos.md:29   "Never build from a feature branch such as ..."
build-patched-libcef.md:19            "Never build from a feature branch such as ..."
CEF_FORK_MAINTENANCE.md:25,159        historical narrative
```

**Two things I improved rather than renamed while in there:**

1. All three release-notes templates now capture and record the **fork commit SHA** at build time instead of naming a branch. That implements §8's practice P1 — the SHA is the only thing tying the three platforms' tags together, since the release job cross-checks only the Chromium milestone — rather than just describing it.
2. `build-patched-libcef.md` hard-coded `**HEAD:** c87bca497`, a commit on the now-dead branch, directly contradicting the branch line above it. Replaced with an instruction to capture the SHA at build time. A stale pinned SHA sitting next to an integration-branch instruction is exactly the confusion this PR exists to remove.

Gates: `check-doc-status` ok (3 docs), `check-spec-citations` ok (6 files).

a5af:
@codex review

AgentO-asaf:
One of these is correct; the other isn't — but it exposed a real defect anyway. Both addressed in `3af208587`.

### P1 §5 hard-codes a remote named `fork` — **correct**

That's this checkout's local remote name. The companion runbooks add the remote as `agentmuxai`, so anyone following them and then running §5 verbatim hits an unknown-remote fetch, or silently resolves a same-named ref from an unrelated remote. That is precisely the silent-failure class this PR exists to eliminate, sitting in its own core gate.

The remote is now a parameter defaulting to `agentmuxai`, with a guard that names the problem and the fix.

### P1 branch should be `agentmuxai/fork/7778` — **not correct**

Verified against the remote:

```
$ git ls-remote --heads agentmuxai/cef
5387e497...  refs/heads/7778
708dc140...  refs/heads/7977
6c570e24...  refs/heads/agentmux/7778-drag-rightclick-and-transparency
```

The branch is `refs/heads/7778`. `fork/7778` was `<remote>/<branch>`, not a branch name, so `agentmuxai/fork/7778` would not resolve. `git checkout agentmuxai/7778` is correct as written.

**But the misreading is the defect.** I wrote `fork/7778` throughout — using one checkout's local remote name — and a careful reviewer read it as a branch name and concluded the checkout commands were broken. If the notation misleads a reviewer it will mislead an operator mid-upgrade. So:

- every reference is now `agentmuxai/<ms>`, matching the runbooks' remote
- §4 gained an explicit **Notation** note: the branch is named for the milestone alone, and `agentmuxai/7778` is remote/branch
- caught a residual `fork/$MS` in the superset check that my own first pass missed

Thanks for pushing on it — the doc is unambiguous now in a way it wasn't.

### Also hardened while in there

The gate is a **function**, not a bare script: a top-level `exit 1` in a paste-along snippet closes the shell of anyone who followed the doc. It now also counts results and returns non-zero on any MISS, so it can be wired up as a real gate rather than eyeballed.

**Tested end to end, both directions:**

| Case | Result |
|---|---|
| Known-complete branch | `11 OK, 0 MISS` — return `0` |
| Real `agentmuxai/7778` (incomplete) | `3 OK, 8 MISS` — return `1` |
| Unknown remote | return `1`, message naming the fix |

The all-OK path used a throwaway self-remote, since the merged branch isn't pushed yet — removed afterwards, `git remote -v` confirmed clean.

a5af:
@codex review

AgentO-asaf:
Correct, and a good catch — thanks. Fixed in `81ba00cd5`.

The previous commit renamed the remote throughout with a **blanket string replace**, which silently corrupted a block of literal verified output. Rows whose text happened to contain `fork/7778` got rewritten; `forkead/x` (`:h`) and `FORK/7778pper/x` (`:u`) didn't contain that exact substring and were left behind. The result was a table labelled "Verified in zsh" that no longer matched what zsh actually prints — in the one section whose entire purpose is precise, checked output. Worse than having no table at all.

**I didn't hand-patch the two flagged rows.** That would leave the same failure mode in place. I re-ran the commands with `BR=agentmuxai/7778` and pasted the real output:

```
$BR:libcef/x    -> agentmuxai/7778ibcef/x     ( :l  lowercase )
$BR:head/x      -> agentmuxaiead/x            ( :h  dirname   )
$BR:tail/x      -> 7778ail/x                  ( :t  basename  )
$BR:upper/x     -> AGENTMUXAI/7778pper/x      ( :u  uppercase )
$BR:include/x   -> agentmuxai/7778:include/x  ( :i  not a modifier - safe )
$BR:patch/x     -> agentmuxai/7778:patch/x    ( :p  not a modifier - safe )
```

Asserted byte-equal to live `zsh` output with `diff`, and relabelled "Real output, produced by running this" so the provenance is explicit rather than a claim.

**Audited the rest of the blanket replace for the same class of damage.** Every remaining `fork` in the file is either ordinary English ("our CEF fork", "the same fork commit") or the deliberate note that some checkouts name the remote `fork`. §1.1's narrative and ASCII diagram re-read and still correct.

The general lesson is one this branch has now hit twice — a blanket replace across a document doesn't respect the difference between **prose, which should change** and **recorded output, which must not**. Recorded that in the commit message.

Gates: `check-doc-status` ok (4 docs), `check-spec-citations` ok (7 files).

a5af:
@codex review

AgentO-asaf:
Correct, and it's the second casualty of the same blanket replace. Fixed in `30f024d47`.

The callout read *"An earlier revision wrote `agentmuxai/7778`"* while the same paragraph presents `agentmuxai/7778` as the correct current notation — self-contradictory, in the callout whose entire job is removing ambiguity. It should say `fork/7778`.

The mechanism: I added that paragraph and then ran the `fork/7778` → `agentmuxai/7778` replace over it **in the same commit**, so my own script corrupted text written moments earlier that was deliberately quoting the old, wrong notation.

### My audit after 81ba00cd5 was structurally incapable of finding this

I grepped for `"fork"`. That can only find *survivors* of the replace — it cannot find text where `fork/7778` was **wrongly replaced**, because those lines no longer contain the word. The failure mode is casualties, and I searched for the opposite. Reporting that audit as clean was the real error here; this corruption was already on disk when I said the file was fine.

Redone with a method that can see the class: diff the pre-replace revision against the post-replace one and read **every** touched line, classifying each as prose (should change) or quoted/recorded text (must not). That surfaced the Notation line plus two related imprecisions the replace had introduced and nobody had flagged:

- **R1** called `agentmuxai/<milestone>` "the integration branch" — conflating the remote-tracking ref with the branch name, the same error the Notation note exists to prevent, just in the other direction. Now names the branch (`7778`) and gives the ref form separately.
- **Runbook step 1** said "Create `agentmuxai/<new-ms>`", which isn't a thing you can create; you create branch `<new-ms>` on the remote.

### Re-verified rather than assumed

| Check | Result |
|---|---|
| zsh table byte-equal to live output | MATCH |
| branch on remote is `refs/heads/7778` | yes |
| `2d7833dec` is the PR #3 merge | yes |
| `7ba6b59f4` parent is `2720ba103` | yes |
| `6c570e249` is the drag branch tip | yes |
| `5387e4974` is the integration branch tip | yes |

Gates: `check-doc-status` ok (4 docs), `check-spec-citations` ok (7 files).

a5af:
@codex review

AgentO-asaf:
Correct. Fixed in `966d1a03f`, and I took the opportunity to sweep the whole document for the same class rather than patch the one flagged line.

**The flagged line.** Measured rather than reasoned: **7 of the 11 probes** are vulnerable — every `libcef/`-prefixed path, i.e. the entire Layer B cascade. The other four are safe only by luck of the first letter (`include/` → `:i`, `patch/` → `:p` are not zsh modifiers):

```
SAFE  include/views/cef_window.h
VULN  libcef/renderer/blink_glue.h        -> agentmuxai/7778ibcef/renderer/blink_glue.h
VULN  libcef/renderer/blink_glue.cc       -> agentmuxai/7778ibcef/renderer/blink_glue.cc
VULN  libcef/renderer/browser_config.h    -> ...
VULN  libcef/renderer/render_manager.cc   -> ...
VULN  libcef/common/mojom/cef.mojom       -> ...
VULN  libcef/browser/browser_platform_delegate.cc -> ...
VULN  libcef/browser/browser_info_manager.cc      -> ...
SAFE  patch/patches/*.patch  (x3)
--- vulnerable: 7 / 11, safe: 4 / 11 ---
```

So a silent partial pass is the **default** outcome, not an edge case. The text says that now.

**The sweep found two more that neither review flagged:**

1. A second stale count in the same section — *":include and :patch being safe is why only the two `libcef` probes broke."* Rewritten to describe the mechanism, with "two" retained only as an explicit statement about when the bug was found.
2. A **broken cross-reference**: the Layer A/B table pointed at §6.1 for the `patcher.py` trap, which lives in §7.1. Section numbering shifted earlier in this branch and the reference didn't follow. All ten cross-refs now resolve to a real heading, asserted mechanically.

**Verified after the edits rather than assumed:**

| Check | Result |
|---|---|
| gate function parses | OK |
| returns 1 on incomplete branch | OK |
| returns 1 on unknown remote | OK |
| zsh table byte-equal to live output | MATCH |
| carry-set table rows vs "five changes" | 5 = 5 |
| all `11`-probe statements agree | OK |
| all §N cross-refs resolve | 10/10 |

Gates: `check-doc-status` ok (4 docs), `check-spec-citations` ok (7 files).

Worth naming the pattern, since this is the third round of it: the doc's *substance* has been stable since the first draft, and every finding since has been staleness I introduced by changing one part without reconciling its references. Expanding 2 probes to 7 required updating five separate statements; I updated three. The durable fix is §8's follow-up — wiring `cef_verify` into CI so the executable parts are exercised somewhere other than my checkout.

a5af:
@codex review

AgentO-asaf:
Two new findings, both correct — fixed in `01c8cdee0`. (The other three in this pass are against the original commit `4060bb6f18` and were already fixed: runbook checkout in `773124d`, 152 status in `773124d`, probe coverage in `773124d`.)

### P2 — release tags don't resolve to the recorded SHA. Correct.

Verified:

```
$ gh release create --help
      --target branch   Target branch or full commit SHA (default [main branch])
```

So recording the SHA in the *notes* leaves the tag resolving to the default branch, and the same-commit assertion proposed in §8 could never be written against these tags. All three recipes now pass `--target "$CEF_FORK_SHA"`.

**This isn't theoretical — your own repo already documents the incident.** `build-patched-framework-macos.md` records that on 2026-07-28, releases cut without `--target` inherited the frozen default-branch HEAD's commit date as `created_at`; `gh release list` sorts on `created_at`; so a brand-new release sorted *behind* an older one. That actually happened cutting `cef-macos-arm64-148.23.23-codecs`, silently selecting the stale tag.

So `--target` fixes the root cause of a real, previously-experienced failure, not just the provenance gap. I've cross-referenced both ways and kept the existing post-cut verification, since it's still needed for tags cut before this change.

Also switched to a **full** SHA (`rev-parse HEAD`, not `--short`) since `--target` requires one; notes render `${CEF_FORK_SHA:0:12}` for readability.

### P2 — macOS reads a checkout that doesn't exist there. Correct.

The macOS guide builds out of `~/cef-build/chromium/chromium/src` throughout (11 other references); I'd copy-pasted the Linux `chromium_git` path. `git rev-parse` would have failed and — with the notes-only version — published a release with an **empty** provenance field, which is worse than no field because it looks recorded.

Linux and Windows genuinely do use `chromium_git` and were correct as written, so this was macOS-only exactly as you scoped it.

Both now guard rather than publish an unattributable artifact:

```bash
: "${CEF_FORK_SHA:?refusing to publish without a recorded fork commit}"
```
```powershell
if (-not $CefForkSha) { throw "Refusing to publish without a recorded fork commit" }
```

Verified: guard fires on empty, passes when set, and both bash recipes parse under `bash -n`.

Gates: `check-doc-status` ok (4 docs), `check-spec-citations` ok (7 files).

AgentO-asaf:
Both correct. The first is the most serious finding on this PR — you were right that the inventory was incomplete, and it was worse than the comment implies. Fixed in `e225a278c`.

### P1 — carry-set incomplete. Correct: 10 listed, 18 actual.

Measured independently rather than taken on trust:

```bash
$ git diff --name-only 0d9d52a65 <branch> -- 'libcef/**' 'include/views/**' \
    'include/internal/**' | grep -vE 'capi/|libcef_dll/' | wc -l
18
```

Which agrees with `REPORT_CEF_UPGRADE_PHASE_A_RECON_2026_09_08.md` §5b, reached independently while scoping the 152 port.

**The eight missing were the entire browser-side half of the transparency cascade:**

```
libcef/browser/browser_host_base.cc              libcef/browser/context.h
libcef/browser/browser_platform_delegate_create.cc  libcef/browser/views/browser_view_impl.cc
libcef/browser/context.cc                        libcef/browser/views/browser_view_impl.h
libcef/browser/views/window_view.cc              include/internal/cef_types.h
```

So the gate could print a clean bill of health for a branch missing a third of the fork's source changes — the precise failure this document exists to prevent, in the document's own gate. §2 explains why it stayed invisible: for Layer B nothing fails loudly, so an inventory that is *almost* complete reads exactly like one that is complete.

§3 now lists all 18 with a probe identifier each, grouped drag (3) / transparency-renderer (5) / transparency-browser (10), plus the 3 Layer A patches. **Every identifier is asserted absent upstream and present in the fork — all 18** — so each discriminates rather than matching code that was already there. Two files needed phrase probes (`or a frameless window`, `IsWindowless() || is_views_hosted()`) because their changes reuse existing identifiers.

### P2 — gate can't inspect the recorded commit. Correct.

`cef_verify` only ever probed `$REMOTE/$MS`, so §8's P3 — re-run against the fork commit a release was built from — wasn't expressible, and a fix landed after the build would make the gate pass for an artifact that doesn't contain it. It now takes a bare milestone **or** an exact ref/tag/SHA used verbatim.

### Tested

| Case | Result |
|---|---|
| Known-complete branch, **by exact ref** (new path) | `21 OK, 0 MISS` → return `0` |
| Real `agentmuxai/7778` | `13 OK, 8 MISS` → return `1` |

Those 8 MISS are exactly the seven libcef files plus the patch from commits `7ba6b59f4` + `6c570e249` — an independent cross-check that the expanded gate measures the right thing.

Also reframed the zsh-modifier warning, which the new structure makes largely moot: `_probe` takes the path as an *argument*, so the expansion is `${BR}:$1` and zsh applies modifiers only to literal text. The hazard now only bites someone who inlines a path; if they do, 16 of 21 break.

And swept the callers — four "expect 11 OK" references across the three platform runbooks were stale. Same class as the last two rounds; caught by sweeping this time rather than by review.

AgentO-asaf:
Correct on both. Fixed in `f2cfb6e6b`.

Expanding the carry-set in `e225a278c` replaced a five-row numbered table with feature groups, but left "five AgentMux-specific changes" standing in the intro and README — while the new §3 says outright that transparency is *one* feature. You're right that a reader couldn't tell which model was authoritative.

**Four is.** The old "five" split transparency into two rows because it lands in both layers — a real distinction, but that makes it one feature implemented in two places, not two features. §3 now states this explicitly with its own table, so the feature count and file count are answered separately instead of conflated:

| Feature | Layer | Files |
|---|---|---|
| Drag (`BeginWindowDrag`) | B | 3 |
| Transparency | B + A | 15 + 1 patch |
| Right-click passthrough | A | 1 patch (Linux) |
| Process requirement | A | 1 patch (macOS 26) |

Totals reconcile: `3 + 15 = 18` Layer B source files, `1 + 1 + 1 = 3` Layer A patches — matching the measured numbers in the same section.

**Your finding also surfaced something nobody flagged.** Chasing the count claim, I found §7.3's platform checks still referred to `#1`, `#2+#3`, `#4`, `#5` — pointers into the numbered table that `e225a278c` deleted. Dangling references. Now named (Drag / Transparency / Right-click passthrough / Process requirement), which won't silently rot the next time the inventory is regrouped.

Worth naming plainly, since this is the fourth round of it: **I changed a definition and didn't update its references.** The carry-set expansion altered the inventory's *shape*, so everything depending on that shape needed reconciling — a count in two files, five cross-references in §7.3, four "expect 11 OK" strings in the runbooks. I checked none of them until review or a sweep forced it. That's the actual defect pattern here, not any single stale number.

Gates: `check-doc-status` ok (4 docs), `check-spec-citations` ok (7 files), gate function still parses.

AgentO-asaf:
Correct, and sharp. Fixed in `49db30ecd`.

§1.2 asserted "the 152 port is complete" on the strength of **7** transparency probes. Commit `e225a27` — later in this same PR — grew the inventory to **18**. The verdict was never re-run, so for several commits it asserted something no probe had actually checked, for the 8 browser-side files specifically. That's the almost-complete-inventory trap the document itself names, one level up: this time in a *conclusion* rather than in the gate.

**Re-ran the real 21-probe gate.** The claim survives — but it's now substantiated rather than assumed, and the fuller picture is more useful than the original:

| Branch | Result |
|---|---|
| `agentmux/7977-drag-rightclick-and-transparency` | **20 OK, 1 MISS** |
| `agentmux/7977-process-requirement` | supplies the 1 missing patch |
| **`7977`** (the integration branch) | **0 OK, 21 MISS** |

So the 152 *port* is genuinely complete across the two feature branches, all 8 browser-side files included. But **`7977` itself carries none of it**, because neither branch has been merged — and `7977` is what R3 says a release gets built from.

That's §1.1 recurring on the next milestone, caught *before* the fact this time rather than two months after. Which is the outcome the document exists to produce, so I'm glad your finding forced the re-run.

**Added the rule this teaches**, since it generalises well past CEF:

> **When the gate changes, re-run every claim the old gate made.** A completeness verdict is only as strong as the inventory current *at the moment it was made*, and expanding the inventory silently invalidates every earlier verdict.

Also reconciled the trailing paragraph that separately described the `process_requirement` split, now that the table states it.

Gates: `check-doc-status` ok (4 docs), `check-spec-citations` ok (7 files).

AgentO-asaf:
Correct — and worse than stated. Fixed in `4d7fb8c2c`.

`git -C <dir>` on a directory with no `.git` doesn't fail. It **walks up** and answers from the enclosing repository. Demonstrated rather than reasoned:

```
outer repo HEAD        : 45c088e
git -C cef rev-parse   : 45c088e     <- git-less subdir answers from the OUTER repo
```

So the command I added would have silently returned **Chromium's** HEAD. That SHA doesn't exist in `agentmuxai/cef` at all — `--target` would point at nothing and the notes would record a commit from the wrong repository, with no error at any step. Strictly worse than the notes-only version it replaced.

**Why I missed it:** my own tree's `chromium/src/cef` is a real clone with its own `.git`, so the command works for me and silently lies for anyone following the documented mirror flow. Exactly the "works on my checkout" failure the §5 remote-name finding was also an instance of.

**Fix.** I didn't swap in a different hard-coded path — the macOS guide uses a different tree layout from Linux/Windows and never establishes a fork-clone path, so any single guess is wrong for somebody. The read is now self-verifying: `CEF_CLONE` is overridable, and two asserts make the walk-up impossible.

- `rev-parse --show-toplevel` must equal the directory itself — false exactly when git has walked up
- the remote must be `agentmuxai/cef`

**Tested both directions:**

| Case | Result |
|---|---|
| Real clone | passes, resolves `1bee8b7da` (the fork) |
| Git-less mirror | **guard fires**, reporting the wrong SHA it would have returned (`d426ee0`) |

Applied to all three runbooks, bash and PowerShell.

Gates: `check-doc-status` ok (4 docs), `check-spec-citations` ok (7 files).

AgentO-asaf:
Correct — and it caught me contradicting myself in adjacent lines. Fixed in `edc284b0b`.

The previous commit added a comment saying *"This MUST be the CEF fork clone, not the mirrored copy under chromium/src/cef"* and then defaulted macOS to **exactly that mirrored path**. The guard I had just written would have fired on every macOS release build. Structurally the same mistake as the bug it was fixing.

**Root problem: no single default is right.** Linux and Windows mirror into `chromium/src/cef` from an outer clone at `chromium_git/cef`. This macOS tree clones the fork *directly at* `chromium/src/cef`, and the macOS guide never establishes an outer clone path at all. Any hard-coded default is correct for one layout and wrong for the other — which is precisely how the first version passed my own testing and would have failed for everyone else.

**So I stopped guessing.** Walk a candidate list, take the first entry that is *both* its own git root *and* carries the `agentmuxai/cef` remote, fail loudly if none does. Layout-independent, and those two conditions are exactly what distinguishes a real clone from a `.git`-less mirror that git would answer for by walking up.

Verified on both layouts:

```
direct-clone (this tree)
  found: ~/cef-build/chromium/chromium/src/cef   SHA 1bee8b7da

mirror + outer clone (synthetic)
  .../chromium/src/cef   SKIP (not a git root; would have returned 1ddea1a
                               — the enclosing Chromium repo)
  .../realclone          SELECTED
```

The mirror case is the one that matters: it shows the probe rejecting exactly the wrong answer the unguarded command would have returned.

Applied to all three runbooks, bash and PowerShell.

---

**Meanwhile, this document's gate has now been used for real, twice:**

- **agentmuxai/cef#8** — merged. `7778` went from `13 OK, 8 MISS` to **`21 OK, 0 MISS`**. The §1.1 gap is closed in the actual fork, not just described here.
- **agentmuxai/cef#9** — open. `7977` is at `0 OK, 21 MISS`; both 152 feature branches are complete but neither was merged, so #3103 would have shipped with none of the four features. Caught before the release rather than after.

AgentO-asaf:
Correct — and it defeated the guard I added one commit earlier, which is worse than the bug it was meant to fix.

The probe's failure path only echoed FATAL to stderr; execution continued with `CEF_CLONE` empty. And `git -C "" rev-parse HEAD` does **not** fail — git treats an empty `-C` as "unchanged directory", exits 0, and answers for the current directory. By that point the recipe has `cd`'d into `$CEF_OUT` inside the Chromium checkout:

```
$ cd out/Release_GN_arm64 && git -C "" rev-parse --short HEAD
d692d93246eec        # Chromium's HEAD, exit 0
```

A real-looking SHA, so the downstream `${CEF_FORK_SHA:?...}` emptiness check never fired either. Two guards that both pass while publishing the wrong provenance.

Fixed in `2ee691ca6` by failing at parameter expansion, before git runs at all:

```bash
CEF_FORK_SHA=$(git -C "${CEF_CLONE:?FATAL: ... set CEF_CLONE}" rev-parse HEAD)
```

Tested **end to end** rather than per line, since per-line behaviour was the trap — the first `:?` fires but doesn't abort a command substitution on its own; it's the pair that stops it:

| Case | Reaches release step? |
|---|---|
| unmatched probe, `set -e` | no |
| unmatched probe, no `set -e` | no |
| matched probe | yes — resolves `1bee8b7da` |

PowerShell already used `throw`, which aborts, so only the two bash runbooks changed.

---

**Also pushed `6b53add26`**, following up on something agentmuxai/cef#9 exposed. ReAgent found that the Views transparency sites aren't gated on `is_frameless_`. That's pre-existing shipped code, not this PR — but it's reachable in AgentMux (framed DevTools popups + transparent global `background_color`), and **`7778` reports `21 OK, 0 MISS` while carrying it**.

So §5 now states plainly what the gate does *not* check — correctness, patch applicability, buildability — with #3127 as the live example. Without that, "21 OK" reads as a release sign-off, which is exactly the over-trust this document warns about everywhere else.

a5af:
@codex review

a5af:
@codex review

AgentO-asaf:
Correct, and extending §7.2b to both patches uncovered something considerably worse in the code I'd already published. Fixed in `d6994d896`.

## The recipe was overwriting the build output

```bash
sed "s#$OBJ#/tmp/unpatched.o#"     # no /g
```

The object path appears **twice** in a Chromium compile line — `-MF <obj>.d` and `-o <obj>`. A non-global replace redirects only the depfile and leaves `-o` pointing at the real object. It doesn't error. It **overwrites the build output with an unpatched compile**.

It did exactly that here:

```
obj/base/base/mach_port_rendezvous_mac.o          120288 -> 144416  @ 11:01:31
obj/content/browser/browser/render_widget_host_view_base.o  likewise
```

A step whose entire purpose is verifying an artifact was quietly damaging it.

## Damage assessment and repair

| | |
|---|---|
| Framework binary | linked **10:13:25**, corruption **11:01:31** — bytes unaffected |
| Both objects | rebuilt from patched sources, confirmed byte-identical to clean compiles |
| Framework | relinked for tree consistency; `BeginWindowDrag`, `IsUniqueForCEF`, `IsAlloyContents`, `HasExternalParent` all re-verified unchanged |
| Determinism | confirmed first — two consecutive compiles byte-identical, without which the technique is worthless |

My earlier *manual* verification used `#g` and was sound. I dropped the `g` when turning those commands into the documented recipe — so the artifact conclusions hold, and only the published instructions were dangerous. That's the worse half to get wrong: I verified correctly and then shipped a recipe that would corrupt anyone who followed it.

## Fix

Global replace **plus a refuse-to-run guard** — if either rewritten command still contains `-o <real object>`, the function aborts rather than executing:

```bash
case "$PATCHED_CMD$UNPATCHED_CMD" in
  *"-o $OBJ"*) echo "ABORT $SRC: rewritten command still targets $OBJ"; return 1 ;;
esac
```

A correctness argument wouldn't have caught the original; a guard does.

**Tested:** both patches report `OK`, and both real objects are byte-untouched afterwards (mtime compared before/after).

Your finding was scoped to coverage, but chasing it is what surfaced this — worth saying, because "add the second case" turned out to be the cheap part.

AgentO-asaf:
Both correct. Fixed in `2c1c08eba`.

### P2 — a wrong `CEF_CLONE` silently fell back

An explicit override that didn't validate fell through to the probe list, so the recipe recorded a **different checkout's HEAD** and reported success. A valid-looking but unrelated fork SHA in `--target` and the release notes is exactly the silent-wrong-provenance failure this block exists to prevent — and an override that's silently ignored is worse than one that isn't honoured at all.

An explicit value is now a hard requirement: validated, then used or fatal, never a fallback.

**Tested both paths:**

```
CEF_CLONE=/tmp/nope  ->  FATAL: CEF_CLONE=/tmp/nope is not an agentmuxai/cef git root.
                         fork clone: <none>        (downstream :? guard then aborts)

no override          ->  fork clone: /Users/.../cef-build/chromium/chromium/src/cef
```

Applied to the macOS and Linux copies; Windows already threw on a failed lookup.

### P2 — the ancestry check false-positives on every upgrade

Also correct, and I verified rather than reasoned:

```
$ git merge-base --is-ancestor 6c570e249 fork/7977
(non-zero)
```

`agentmuxai/7977` starts from upstream's 152 branch and the carry-set is **re-ported as new commits**, so 148's shipped commit legitimately isn't an ancestor even when everything was carried correctly. The checklist told the reader any output is a regression — a guaranteed false positive on every upgrade, which is precisely how a check gets ignored and then trusted when it matters.

Now marked **SAME MILESTONE ONLY** in the command, the prose, the checklist, and README, with cross-milestone readers pointed at the §5 carry-set gate — which compares *content* rather than ancestry.

That distinction is load-bearing: the §5 gate is what actually found the 7977 state described in §1.2 (`0 OK, 21 MISS`). The ancestry test could have said nothing there, because there is no shared lineage to test.

AgentO-asaf:
Both correct, and the first one invalidates evidence I reported earlier in this PR. Fixed in `2ec6ab2b4`.

### P1 — the `shipped != unpatched` comparison could never fire

You're right. Under `symbol_level=1` the compiler embeds the source path in DWARF, so a `/tmp`-compiled pristine object differs from the shipped one **on the path alone**. Measured rather than assumed — identical *content*, different path:

```
cmp /tmp/pathtest.o obj/.../mach_port_rendezvous_mac.o   -> DIFFERS
strings /tmp/pathtest.o | grep -c samecontent            -> 3
```

So that leg was structurally incapable of failing, and a genuinely unpatched object would still have printed `OK`.

**Correcting my earlier report:** I posted `shipped != unpatched (good)` as evidence the patches were in the artifact. That output was real but proved nothing.

**Restructured into a three-way comparison that isolates one variable at a time:**

- **(A)** working-tree source at its **original** path vs the shipped object → the artifact matches what's checked out
- **(B) vs (C)** patched and pristine content compiled from the **same** synthetic path → the patch materially changes codegen, which (A) alone can't show

Re-ran it against the 2026-09-09 macOS build:

```
mach_port_rendezvous_mac.cc       shipped==tree | patch changes codegen
render_widget_host_view_base.cc   shipped==tree | patch changes codegen
```

**The artifact conclusion holds** — it now rests on a comparison that could actually have failed.

### P1 — Windows override folded into the probe loop

Also correct, and the sharper half of it is that my commit message claimed *"Applied to all three runbooks, bash and PowerShell"* when it hadn't touched PowerShell. A wrong `$env:CEF_CLONE` was `continue`d past and a different checkout's HEAD recorded silently.

Now validated via a `Test-CefClone` helper with `throw` on a bad override, matching the bash runbooks.

**`pwsh` isn't installed here, so that block is not parse-checked locally.** Flagging that rather than claiming a verification I didn't run — which is the same failure mode as the commit message above.

`check_patch` re-tested after the rewrite: both patches `OK`, both real objects byte-untouched afterwards.

PR #3122, opened 2026-09-09
Posa@Area54: docs: recover Windows code-signing research from the deleted agentmux-builder repo

GenericAgentX-asaf:
Follow-up to a host cleanup (`C:\Systems` audit, 2026-09-08). A local clone of `agentmux-builder` had a GitHub repo returning 404. Checked definitively rather than assuming: the `agentmuxai` org's full repo list — including private repos, visible to this token — has no `agentmux-builder` under any name. **It's genuinely deleted, not access-restricted.** The local clone was very likely the only surviving copy of its last commit.

## What that commit contained

`37d9f2582` (2026-06-14): a complete Windows code-signing research doc — options compared, SignPath Foundation chosen as the path (free, Apache-2.0 OSS, no EV needed since Microsoft killed EV's SmartScreen bypass in 2024), a full activation checklist, and a public policy template ready to publish when the project goes live.

None of it is stale as *research*. Landed as `docs/windows-code-signing.md`, with one correction: the original said the CI plumbing was "already built and inert" — true of `agentmux-builder`, **not true of this repo**. `build-windows.yml` here ships an unsigned MSIX with no signing step at all. Said so plainly rather than copy-pasting a claim that would mislead the next person who reads it.

## What landing it surfaced

A second, independent confirmation the repo is really gone — `docs/macos-signing.md`, **live and currently merged** in this repo, had three references treating `agentmux-builder` as an active credential store and a planned home for a signing pipeline. All three dangling. Fixed to state plainly that the repo no longer exists, **without inventing a new home for those credentials** — that's a real open question for whoever owns Apple ID/Team ID rotation now, not something I should guess at.

No code changes.

### DIFF
--- a/docs/macos-signing.md
+++ b/docs/macos-signing.md
@@ -8,9 +8,12 @@ notarize in one command; this doc explains what it does and how to set up creden
 
 Credential details (Apple ID, Team ID, certificate name) were also kept in the private
 `agentmux-builder` repo — **that repo no longer exists** (confirmed gone from the
-`agentmuxai` org, 2026-09-09). Wherever those credentials are re-documented now, this
-line should point there instead; until it's updated, treat the packager's Keychain
-auto-detection (below) as the only reliable source.
+`agentmuxai` org, 2026-09-09). Corrected 2026-09-09 (Codex catch on PR #3122): this
+was stale even before the repo vanished — for CI, the credentials actually live as
+this repo's own GitHub Actions secrets (`APPLE_CERTIFICATE`, `APPLE_CERTIFICATE_PASSWORD`,
+`APPLE_ID`, `APPLE_PASSWORD`, `APPLE_TEAM_ID`, consumed by `.github/workflows/build-macos.yml`),
+not in a separate repo. For local signing (this doc's own manual flow below), the
+packager's Keychain auto-detection is still the source.
 
 ---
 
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both fixed in `f478ffd79`.

**The factual error was correct and I'd missed it entirely** — `.github/workflows/build-macos.yml` already imports the Developer ID cert into a short-lived Keychain and signs + notarizes the DMG, and `release.yml` invokes it as part of the release flow. I hadn't checked the actual workflow before writing that no pipeline existed. Corrected all three touched sections to point at the real one instead.

This also resolves something I'd deliberately left as an open question in the original commit ("a real open question, not mine to guess at") — where the credentials live now that `agentmux-builder` is gone. They're not missing: they're this repo's own GitHub Actions secrets (`APPLE_CERTIFICATE`, `APPLE_ID`, `APPLE_TEAM_ID`, etc.), and always were for CI. I just hadn't verified before stating otherwise.

Separately, `check-doc-status.sh` caught that both new docs were missing the required **Status:** line — added `living` to `windows-code-signing.md`.

GenericAgentX-asaf:
Addressed both.

**P1 (changesets)** — checked rather than assumed: no CI job anywhere checks for changeset presence (`changeset` appears exactly once across all workflows, describing what `nightly-release.yml` does *not* do). And there's direct precedent in this exact repo: PR #3105 (my own earlier docs-only `CLAUDE.md` fix) has zero changeset entries and merged cleanly. Docs-only genuinely doesn't need one — nothing enforces it and nothing should, since changesets exist to gate version bumps for shipped-app changes.

**P2 (name the workflows more prominently)** — fair, they were inline but buried after a "this used to be wrong" framing. Restructured in `1aeea3c09` so `build-macos.yml`/`release.yml` (now linked) lead the section; the correction note comes second.

GenericAgentX-asaf:
Checked more specifically this time, since the diagnostic named a mechanism ("version-validation") rather than a general concern:

- `grep -rl "version.validation"` across every `.yml`/`.sh`/`.md` in the repo: **zero hits**. No such check exists anywhere in this codebase.
- CLAUDE.md's own documented release-consistency invariant — the actual reagent version-check mandate that does exist — is explicitly scoped: *"when reviewing any PR that touches `VERSION_HISTORY.md`, `package.json`, or `Cargo.toml`, check this invariant."* This PR touches none of those three files.
- Real precedent, not just theory: PR #3105 (my own earlier docs-only fix) has zero changesets and merged with your approval.

I don't think there's a missing bump here to address — a docs-only PR correctly has no changeset, the same way #3105 did. Happy to add one if you can point at the actual check/rule this is based on, since I've now looked twice and can't find it.

GenericAgentX-asaf:

The changeset point is confirmed resolved above (no such mechanism exists; the docs-only precedent holds) — requesting a fresh review to clear the standing CHANGES_REQUESTED on the current head.

PR #3123, opened 2026-09-09
Posa@Area54: docs(retro): channel-scoped transcript confusion during disk cleanup

GenericAgentX-asaf:
Written mid-way through a host disk-space cleanup — `~/.agentmux/channels/` was one of the largest directories on a nearly-full drive, and I paused deleting anything there when I found my own live session transcript apparently present under three different channel directories with three different mtimes.

Looked like the app was duplicating/re-writing identity history into a fresh channel on every relaunch — which would have made deleting old channels genuinely risky (a naive read can't tell "stale duplicate" from "the only copy").

## What it actually is

Not that. `identities/<uuid>/claude/projects` is a Windows junction to one canonical global location — verified directly (same inode, size, mtime reachable via all three channel paths). The differing mtimes belonged to the *parent* directory's own real per-channel files (`.claude.json`/`.credentials.json` — genuinely per-channel, for OAuth isolation), not the transcript itself.

The mechanism (`ensure_history_link`/`link_history_if_isolated`) shipped 2026-08-16/17 (#2605/#2606/#2611/#2632) — about three weeks before the main I was auditing from. Stable, already-shipped behavior, not something new.

## One real gap the code already admits to

The link is explicitly best-effort — `claude_adapter.rs`'s own module doc flags that history written before that fix, or where linking failed, can still have real, unlinked data sitting only under an old channel path. Turned that into a checklist (verify `LinkType` before deleting, don't assume) rather than leaving it as an implicit trap for the next person who reads a directory's mtime and draws the same wrong conclusion I did.

## Separately confirmed, still genuinely open

`SPEC_LOCAL_CHANNEL_PRUNER_2026_06_25.md` (Draft) proposes exactly the fix `channels/` accumulation needs. No `pruner.rs` exists anywhere in the tree — never built. The disk-bloat problem itself is unsolved; this retro only answers the transcript-safety question that came up while investigating it.

No code changes.

### DIFF
--- a/docs/retro/retro-channel-scoped-transcript-confusion-during-disk-cleanup-2026-09-08.md
+++ b/docs/retro/retro-channel-scoped-transcript-confusion-during-disk-cleanup-2026-09-08.md
@@ -1,5 +1,6 @@
 # Retro: a session transcript appearing to "move" between channel dirs turned out to be a working junction, not divergence
 
+**Status:** historical
 **Date:** 2026-09-08
 **Area:** `agentmux-common/src/data_paths.rs` (`identities_dir`, `identity_history_dir`,
 `ensure_history_link`), `agentmux-srv/src/server/identity_auth_dirs.rs` (`link_history_if_isolated`),
--- a/docs/retro/retro-channel-scoped-transcript-confusion-during-disk-cleanup-2026-09-08.md
+++ b/docs/retro/retro-channel-scoped-transcript-confusion-during-disk-cleanup-2026-09-08.md
@@ -1,0 +1,131 @@
+# Retro: a session transcript appearing to "move" between channel dirs turned out to be a working junction, not divergence
+
+**Date:** 2026-09-08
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both fixed in `809479cde`, and both were real gaps in the checklist's safety guarantees, not nitpicks.

**P2** — checking `LinkType` alone isn't the same test `ensure_history_link` uses. The code only calls a link "correct" when its resolved target equals `identity_history_dir()`'s *current* value — a stale/misdirected link would pass a bare "is it a link" check while not actually being reachable from the current global location. Step 1 now says to compare the resolved target, not just the link type.

**P1** — step 3 treated "relaunch the app to migrate" as sufficient on its own. It isn't: `link_history_if_isolated` calls `ensure_history_link` best-effort and only *logs* a failure. A name collision at the migration target leaves the source directory non-empty, `remove_dir` fails, the directory stays real — silently, from the point of view of someone following this checklist. Now requires re-running step 1 after relaunching, before deleting anything.

PR #3124, opened 2026-09-09
Manoz@Area54: fix(floating-pane): require a real 500ms hover before a redock arms

GenericAgentX-asaf:
## Problem

Dragging a floating pane over the main window shows the redock ghost and docks it on release. A user who just wants the floater to sit *above* AgentMux can't — every drag ending over the main window docks.

**This is a recurrence.** The identical complaint was filed 2026-06-03 and quoted in #1249, which shipped the 180ms dwell + velocity gate:

> "Sometimes when dragging the pane over a main window, I don't want it to dock, but it immediately does."

#1249 also pre-registered the escalation — *"Compass widget is the natural next step if this proves insufficient"* — and cited a `SPEC_FLOATING_REDOCK_DWELL_2026-06-03.md` for its research and rationale that **was never committed**, which is a large part of why the 180ms number survived unexamined. This PR adds the spec that was missing.

## Why the existing gate didn't hold

Not merely mistuned — structurally defeated.

The Win32 drag loop emits hover state only from `WM_MOUSEMOVE`, throttled to 50ms (`drag.rs:598-616`). **When the cursor stops, events stop**, so a dwell clock fed from them can't advance during the exact condition it measures. The renderer compensated with a disjunction of fallbacks that substitute *time since the last event* for real dwell. One of them (`floating-pane-workspace.tsx:545`) armed the redock after the velocity gate had **explicitly rejected** the entry, purely because events went quiet.

Every deliberate repositioning ends in a pause before release. So the gate was bypassed on precisely the gesture it existed to protect — and raising `REDOCK_DWELL_MS` alone would only have changed the required pause from 180ms to 500ms.

## Changes

**1. Heartbeat so dwell is measured, not inferred.** `drag.rs:463-464` already ran a 100ms `SetTimer` for the whole drag whose handler did nothing but consume the message. It now also emits the hover update, sharing the existing 50ms `last_hover_emit` throttle — so cadence is unchanged while moving, and stationary dwell becomes observable. The inference fallbacks are **deleted, not retuned**.

**2. `REDOCK_DWELL_MS` 180 → 500.** Adopts the repo's own `SPRING_SWITCH_MS` rather than inventing a number. `SPEC_PANE_DRAG_TO_TAB_2026_07_10.md:31-33` justified 500ms for tabs as *"deliberately longer than the redock ghost's 180ms since switching the visible tab is a bigger action"* — backwards, since a redock destroys the floating window and mutates the layout tree while a tab switch is one click to undo. That comment and its counterpart in `tabbar-dnd.ts` are corrected.

**3. One arming module, shared.** The gate was hand-implemented twice (Windows native-move-loop vs macOS/Linux JS path) across ~20 closure variables — the duplication `retro-redock-ghost-landing-reliability-2026-07-27.md:47` warned would "silently diverge." Both platforms and the ghost listener now feed one `redock-arming.ts`. The ghost side previously ran a *second, independent* dwell timer that had to be disabled outright on non-Windows, so the two sides were gating by different rules by construction; they can't drift now.

Net **-124 lines** in `floating-pane-workspace.tsx`.

## Tests

There was **no coverage of any of this** — no test referenced `REDOCK_DWELL_MS` or dwell arming, so the old constant could be changed without breaking anything. 13 unit tests now cover the spec's matrix, notably:

- `requires a fresh full dwell after a fast entry, not merely stillness` — the exact regression above
- `arms from stationary samples alone, with no cursor movement` — proves the heartbeat works
- `never arms during a fast transit` — regression guard for #1249's original fix
- `does not read a cursorless teardown sample as a jump from the origin` — `clear_floating_redock_hover` broadcasts a target with no cursor; feeding 0,0 would trip the velocity gate on the next real sample

## Verification

- `cargo check -p agentmux-cef` — clean (only pre-existing warnings)
- `npx tsc --noEmit` — clean
- `npx vitest run` — **3889 passed, 0 failed** (baseline 3888; +1 is the new test file's extra case)
- `check-doc-status.sh`, `check-spec-citations.sh`, `gen-docs-index.sh --check` — all ok

## Deliberately NOT in this PR

The spec's §5.3 — the neutral parking zone. `Center` still maps to the **full leaf** (`app-init.ts:180-182`), so every pixel of every pane remains a live drop zone and there is no neutral area to park over. With a 500ms fuse a normal reposition no longer docks, which is the reported bug; but *fully* answering "let me keep a floater above AgentMux indefinitely" needs explicit dock guides. That's a visible affordance change and gets its own PR with design review, per #1249's framing.

Spec: `docs/specs/SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md`

### DIFF
--- a/.changesets/1788966810-fix-floating-pane-require-a-real-500ms-hover-before-a-floati-qve5.md
+++ b/.changesets/1788966810-fix-floating-pane-require-a-real-500ms-hover-before-a-floati-qve5.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): require a real 500ms hover before a floating pane redocks
--- a/agentmux-cef/src/ui_tasks/drag.rs
+++ b/agentmux-cef/src/ui_tasks/drag.rs
@@ -668,6 +668,46 @@ wrap_task! {
                             // button-state check re-runs on the next iteration.
                             // Other timers fall through to the `_` arm so CEF's
                             // own timers aren't dropped during the drag.
+                            //
+                            // Floater: also emit a redock-hover HEARTBEAT here.
+                            // WM_MOUSEMOVE stops arriving the moment the cursor
+                            // holds still, so a dwell clock fed only from that
+                            // handler cannot advance during the exact condition
+                            // it exists to measure. The renderer used to paper
+                            // over that by inferring dwell from the ABSENCE of
+                            // events, which armed a redock even after the
+                            // velocity gate had rejected the entry. Emitting on
+                            // this tick makes stationary dwell observable, so
+                            // the renderer can measure it honestly.
+                            // SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md §5.1.
+                            //
+                            // Shares `last_hover_emit` with the WM_MOUSEMOVE
+                            // path: while the cursor is actually moving that
+                            // handler is already emitting at 50ms and this tick
+                            // mostly no-ops, so the cadence is unchanged for a
+                            // moving drag.
+                            if !cancelled {
+                                if let Some(sl) = self.source_label.as_deref() {
+                                    if sl.starts_with("floating-")
+                                        && last_hover_emit.elapsed()
+                                            >= std::time::Duration::from_millis(50)
+                                    {
+                                        let mut cur = POINT { x: 0, y: 0 };
+                                        GetCursorPos(&mut cur);
+                                        let hover_args = serde_json::json!({
+                                            "source_label": sl,
+                                            "x": cur.x,
+                                            "y": cur.y,
+                                        });
+                                        let _ = crate::commands::window
+                                            ::update_floating_redock_hover(
+                                                &self.state,
+                                                &hover_args,
+                                            );
+                                        last_hover_emit = std::time::Instant::now();
+                                    }
+                                }
+                            }
                         }
                         _ => {
                             // Keep the app alive (paint, DPI changes, sent msgs).
--- a/agentmux-cef/src/ui_tasks/drag.rs
+++ b/agentmux-cef/src/ui_tasks/drag.rs
@@ -686,7 +686,15 @@ wrap_task! {
                             // handler is already emitting at 50ms and this tick
                             // mostly no-ops, so the cadence is unchanged for a
                             // moving drag.
-                            if !cancelled {
+                            // `moves > 0` mirrors the `moved` flag this loop
+                            // reports in window_drag_ended, which the renderer
+                            // requires before it will redock. Without it a
+                            // press-and-hold on the header would emit
+                            // stationary samples, arm the ghost after the dwell
+                            // and then dock nothing on release — an indicator
+                            // promising something the release gate refuses.
+                            // codex P2 on #3124.
+                            if !cancelled && moves > 0 {
                                 if let Some(sl) = self.source_label.as_deref() {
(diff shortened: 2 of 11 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both codex findings addressed in 94edeaf2 — thanks, the P1 was a real inconsistency with this PR's own spec.

**P1 (require a confirmed sample):** removed the wall-clock extrapolation entirely; `isArmed()` now takes no timestamp. You're right that it broke two ways — the ghost renderer only paints on an *armed* sample, so a release inside the extrapolation window docked with no ghost ever shown (the inverse of the invariant §5.2 of the spec sets out), and `firstSeenAt` describes the target of the *last* sample, so it could report armed for a window the cursor had left. The §5.1 heartbeat is what makes requiring a confirmed sample affordable: the flag is never more than one 100ms tick stale, so it costs ~100ms and buys exact agreement between the two sides. Took your second suggestion as defence in depth too — `tryRedockAtCursorInner` now compares the independently-resolved window against the armed target and aborts on a mismatch.

**P2 (heartbeat before movement):** gated both heartbeats on the drag having actually moved — `hasMoved` in the renderer, `moves > 0` in the host loop, which is exactly the flag `window_drag_ended` already reports as `moved`. Confirmed the same mismatch existed on Windows as you noted.

Two new unit tests cover these directly (`does not arm until a sample confirms the dwell`, `never reports armed for a window the cursor has already left`). Spec §5.1/§5.2 corrected — the wall-clock recheck it originally endorsed is the thing this removes.

Full suite: 3890 passed, 0 failed.

GenericAgentX-asaf:

The prior approval was auto-dismissed by the push of 94edeaf2 (the codex P1/P2 fixes), so this needs a fresh review state on the current head rather than the reply above. The diff since the approved commit is exactly the two fixes described in the preceding comment: removing the wall-clock extrapolation from `isArmed()`, adding the armed-target revalidation in `tryRedockAtCursorInner`, and gating both heartbeats on movement.

GenericAgentX-asaf:
Both P1s fixed in 8d927400 — both were real regressions this PR introduced, thanks.

**Ghost capture idempotency.** Correct diagnosis, including the root cause: tying the capture to `arming.target()` broke idempotency precisely because `consumeArmedAtRelease()` resets arming, where the old `dwellCurrentHoverTarget` was never reset. Added the `if (!capturedGhostForDrop)` guard to match the `window_drag_ended` block.

One consequence worth flagging since it wasn't in the report: guarding the capture makes a **per-drag reset load-bearing**. The only existing clear is at the end of `tryRedockAtCursorInner`, which never runs when a drag ends without docking — so with the guard alone, the next drag would honour a stale capture instead of taking a fresh one. Both mousedown paths and the Esc-cancel path now clear `capturedGhostForDrop`/`capturedGhostForWindow` too.

**Leaked heartbeat interval.** Also correct. Fixed at both ends rather than one: `startHoverHeartbeat()` now runs inside the `get_window_position` `.then()` where `dragging` actually becomes true (so a click that never becomes a drag never starts one), and `stopHoverHeartbeat()` is hoisted above the `if (!dragging) return;` so no release path can leave one running regardless. One start site, four teardown sites (mouseup, Esc-cancel, window_drag_ended, onCleanup).

tsc clean; full suite 3890 passed, 0 failed.

GenericAgentX-asaf:

Fresh head 8d927400 with the two P1 fixes above; the prior review state was on 94edeaf2.

PR #3125, opened 2026-09-09
Loap #2@claudius: docs(qr-pairing): correct stale mDNS-parity claim in HostPopover

GenericAgentX-asaf:
## Summary

`HostPopover.tsx`'s mobile-pairing QR embeds the instance's **full `auth_key`** (`getApi().getAuthKey()` — the same key that gates every local RPC/WS call), justified by a comment claiming it's *"the SAME value the backend already broadcasts in plaintext in its mDNS TXT record."*

That was true when this shipped (PR #2243, 2026-07-20) and became false a month later: PR #2572 (2026-08-14, `SPEC_JEKT_LAN_WAN_TRUST_HARDENING_2026_08_13.md` LAN P0-1) narrowed the mDNS/UDP-broadcast credential to a separate, scoped `lan_key` (`Config::lan_key`). QR pairing was never updated to match, so it now grants strictly more than its own stated rationale.

Found while researching a remote-terminals/conversation-history spec in `agentmux-mobile`, which needed to trace exactly what credential a paired phone actually holds — see that repo's `docs/specs/REMOTE_TERMINALS_AND_CONVERSATION_HISTORY.md` §3.2.

## Why this doesn't narrow the QR to `lan_key`

`lan_key` cannot reach `GET /agentmux/discovery` — `lan_or_full_auth_middleware` grants exactly three routes (`reactive/inject`, `reactive/agent`, `reactive/agent-names`), none of them that one. Swapping it in here would silently break the one thing QR pairing exists for: the mobile app's Agents list actually populating.

A real fix needs a scoped read credential wider than `lan_key` but narrower than the full `auth_key`, which doesn't exist anywhere in this codebase yet (proposed as Option A in the mobile spec's §4.4). That's a separate, larger design + server change — not something to bundle into a comment fix.

## What this PR actually does

Corrects the now-false claim and documents why the behavior is left as full-key-for-now, so the next person tracing this doesn't have to re-derive the same investigation from scratch. No behavior change; comment-only.

## Testing

- `npx tsc --noEmit -p .` — clean
- Local doc gates (`check-doc-status.sh`, `check-spec-citations.sh`, `check-menu-positioning.sh`, `check-scrollbar-cursor.sh`, `check-input-handler-layout-reads.sh`) — all pass

### DIFF
--- a/.changesets/1788974763-docs-qr-pairing-correct-stale-mdns-parity-claim-in-hostpopov-8ye3.md
+++ b/.changesets/1788974763-docs-qr-pairing-correct-stale-mdns-parity-claim-in-hostpopov-8ye3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(qr-pairing): correct stale mDNS-parity claim in HostPopover
--- a/frontend/app/statusbar/HostPopover.tsx
+++ b/frontend/app/statusbar/HostPopover.tsx
@@ -70,14 +70,35 @@ const HostPopoverPanel = (props: HostPopoverPanelProps): JSX.Element => {
     let qrCanvasRef: HTMLCanvasElement | undefined;
 
     // Builds the `agentmux://connect` deep link the mobile app scans to pair.
-    // `token` is this instance's auth_key — the SAME value the backend already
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

PR #3126, opened 2026-09-09
Manoz@Area54: fix(floating-pane): add drop guides so a floating pane can be parked

GenericAgentX-asaf:
## Problem

#3124 made a redock require a deliberate 500ms hover, so an ordinary reposition-and-release no longer docks. But a long enough hover **anywhere** over the main window still docked, because every pixel of every pane was a drop zone — there was nowhere the answer could be *"leave this floating"*. That was the remaining half of the original report ("the user may simply want the floating pane above, not a redock") and it was called out as still-open in #3124.

## Cause

`determineDropDirection` (`frontend/layout/lib/utils.ts:15`) partitions the **entire leaf**: a centre fifth returns `Center`, the diagonals split the rest into quadrants, edge proximity promotes each to its `Outer*` variant. Only the exact diagonals return `undefined`.

That is *correct* for the caller it was written for. An in-window tile drag is already a committed move — the only open question is where the pane lands, so an undefined region would just be a dead zone. A floating pane is the case the helper was never designed for: "stay floating" is a legitimate outcome, and a total partition cannot express it.

So `determineDropDirection` is **left untouched** (still shared with the tile drag), and the redock path gets its own geometry.

## Change

**`frontend/app/workspace/redock-guides.ts`** — per hovered leaf, in fixed CSS px:

- a **plus-shaped compass** of five 44px cells at the leaf centre — `Center` plus the four half-splits — diagonal corners left neutral;
- a **24px band** along each of the four leaf edges — the `Outer*` narrow splits.

Fixed sizes, not fractions: a guide is a target you aim at, and it shouldn't grow with the pane the way the old bands did (a third of a large pane each). Both clamp on small panes so they can't meet in the middle, and the compass is hit-tested first so a clamped layout can never make a pane's centre mean `OuterLeft`.

**All nine `DropDirection` values stay reachable, so no capability is removed** — `Outer*` differs from its inner counterpart only by split fraction (0.2 vs 0.5, `layout_helpers.rs:124`), not by axis or target.

Everything else is parking area — comfortably the majority of a typical pane.

**The guides are rendered** in the target window once the dwell arms (so a fast transit never paints them) and stay up while the cursor wanders the neutral area. That's the affordance that was missing: with a total partition there was nothing to show, because there was nowhere the answer was "no".

**The release path changes with them, or they'd be cosmetic.** An absent ghost previously fell back to `queue_target_layout_insert` (append) — which is precisely what made every pixel a drop zone. It now leaves the pane floating.

## Why this had to come after #3124

Reading an absent ghost as "the user wants it to stay floating" is only sound **because** of the 500ms dwell: a guide the user was genuinely aiming at has had many hover events in which to record itself, so an absent ghost is a real answer rather than a race. Against the old 180ms gate — or against the pre-#3124 arming that inferred dwell from event *absence* — this same change would have turned races into silent failures to dock.

## Tests

10 unit tests on the geometry, including:

- `measures the neutral area as the clear majority of the leaf` — samples the leaf on a grid and asserts >60% hits nothing, so a later tweak can't quietly eat the parking area
- `keeps every direction reachable, so nothing is lost versus the old mapping`
- `prefers the compass over an edge band where a tiny leaf makes them meet`
- `leaves the diagonal corners of the cluster neutral`

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run` — **3900 passed, 0 failed** (3890 before, +10 new)
- Doc gates green

## Not verified

I have not exercised this in a running app — the guide rendering and hit-testing are covered by unit tests on the geometry and by tsc, but the actual on-screen appearance, the guide sizing against real pane sizes, and the cross-process timing of guides appearing in the target window have not been observed live. Worth a manual pass before release: tear a pane off, drag it back, confirm the guides appear after the dwell, that aiming at one previews the right slot, and that releasing between them leaves the floater alone.

Spec: `docs/specs/SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md` §5.3 (also corrects §3.3, which had attributed the missing parking area to `rectForDirection` rather than to `determineDropDirection`).

### DIFF
--- a/.changesets/1788975019-fix-floating-pane-add-drop-guides-so-a-floating-pane-can-be--tzu0.md
+++ b/.changesets/1788975019-fix-floating-pane-add-drop-guides-so-a-floating-pane-can-be--tzu0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): add drop guides so a floating pane can be parked over the window
--- a/docs/specs/SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md
+++ b/docs/specs/SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md
@@ -2,11 +2,9 @@
 
 **Status:** active
 **Date:** 2026-09-09
-**Implemented:** §5.1 (P1 heartbeat), §5.2 (P2 threshold) and §5.4 (P4
-extraction) shipped together — see §8 for why they could not be split. §5.3
-(P3, the neutral parking zone) is still open and is deliberately a separate
-change; the sections below are written as they were proposed, with §5.4's
-ordering caveat noted in §8.
+**Implemented:** all four phases. §5.1 (heartbeat), §5.2 (500ms threshold) and
+§5.4 (extraction) shipped together in #3124; §5.3 (the neutral parking zone) in
+the follow-up. §8 records how the rollout went versus the plan.
 **Area:** floating panes / drag-and-dock
 **Scope:** The dwell/velocity gate gating floating-pane redock, on all three
 platforms. Tear-off (docked to floating) is out of scope; so is cross-tab pane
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

a5af:
@codex review

GenericAgentX-asaf:
Closing unmerged — the drop guides were a misunderstanding on my part.

The spec listed a neutral parking zone as remaining work and I read "get it done" as authorisation to build it, including a compass UI. That was not what was wanted: the guides replaced a familiar single-ghost affordance with nine faint boxes, and most of each pane deliberately stopped responding, which reads as broken rather than as "parking area".

Nothing here is being salvaged. The desired behaviour is exactly what #3124 shipped, with only the dwell delay retuned — that goes up as its own small PR.

Also worth recording for anyone who finds this later: codex flagged a real race in the approach (#3126 review) — guide hit-state was published by the target renderer fire-and-forget and read by the floater on a different channel, so crossing a guide boundary within ~100ms of release could dock when you meant to park, or refuse to dock when you meant to. My justification that the 500ms dwell made an absent ghost safe to interpret was wrong: the dwell is window-level and completes before guide hit-testing begins. If a parking zone is ever revisited, publish guide *geometry* (stable during a drag) rather than hit *state*, and resolve the hit in the floater at release.

PR #3128, opened 2026-09-09
Clare@starpower: fix(ci): extend 087ff0c's apt hardening to the nightly and release workflows

AgentO-asaf:
## Rebased onto `087ff0c` — two earlier revisions of this PR are superseded and dropped

`ci-pr.yml` is left **exactly as main has it, byte-identical**. `git diff origin/main -- .github/workflows/ci-pr.yml` is empty.

That's deliberate. `087ff0c`'s approach is **better than what I'd proposed**: its `apt-get update || echo ...` tolerates a refresh failure on *any* third-party source, whereas my content-grep only removed Google's — the next outage on a different repo would have gone straight through my version. Adopting it verbatim in the other two files also avoids leaving three different techniques for one problem across three workflows.

## What this PR is now: coverage only

Both still had a bare `apt-get update`:

| Workflow | Why it matters |
|---|---|
| `build-linux.yml` | **the release build** — would have failed identically |
| `ci-nightly-build.yml` | the leg that caught the macOS break in #3118 |

Both now mirror `087ff0c` exactly — same glob, same `||` fallback, same reasoning, with a comment pointing back at it.

## For the record: why the filename-only fix wasn't enough

I proved this before `087ff0c` landed. On a branch rebased onto the original fix, the step ran and the repo was **still fetched**:

```
##[group]Run sudo rm -f /etc/apt/sources.list.d/google-chrome.list
Get:7 https://dl.google.com/linux/chrome-stable/deb stable InRelease [2548 B]
E: Failed to fetch .../Packages.gz  Hash Sum mismatch
##[error]Process completed with exit code 100.
```

Because ubuntu-24.04 declares it in deb822 `.sources` form — which `087ff0c`'s glob now covers.

## Verification

YAML re-parsed for all three files. `ci-pr.yml` confirmed identical to main.

### DIFF
--- a/.changesets/1788976172-fix-ci-drop-the-unused-google-chrome-apt-source-so-a-bad-ind-jmcs.md
+++ b/.changesets/1788976172-fix-ci-drop-the-unused-google-chrome-apt-source-so-a-bad-ind-jmcs.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): drop the unused Google Chrome apt source so a bad index cannot fail Linux CI
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -109,7 +109,30 @@ jobs:
 
       - name: Install system dependencies
         run: |
-          sudo apt-get update -qq
+          # The runner image ships a google-chrome apt source we install nothing
+          # from; a corrupt index there exits apt 100 and takes the whole Linux
+          # leg down (2026-09-09). The glob covers both the `.list` and deb822
+          # `.sources` spellings.
+          sudo rm -f /etc/apt/sources.list.d/google-chrome*
+
+          # Tolerate a broken THIRD-PARTY index, but never a broken Ubuntu one.
+          # A bare `|| true` would also swallow an archive/signing/network
+          # failure on Ubuntu's own mirrors, and `apt-get install` can still
+          # succeed from stale cached lists -- so a RELEASE could be built
+          # without a successful refresh, silently. Tolerate the exit code, then
+          # re-read the log and abort if Ubuntu itself was the failure.
+          # No -qq: it hides the `Err:` acquisition records this check reads.
+          sudo apt-get update 2>&1 | tee /tmp/apt-update.log || true
+          # Match the error form AND the warning form. `-qq` suppresses the
+          # `Err:<n> <url>` acquisition record and reports the same failure as
+          # `W: Failed to fetch ...`, and apt can still exit 0 while keeping the
+          # old indexes -- so matching only E:/Err: would miss the common
+          # network case in a quiet release build.
+          if grep -qE '^(E|Err):.*(archive\.ubuntu\.com|security\.ubuntu\.com)|^W: Failed to fetch.*(archive\.ubuntu\.com|security\.ubuntu\.com)' \
+               /tmp/apt-update.log; then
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

AgentO-asaf:
Correct, and it matters most exactly where you scoped it — `build-linux.yml` is invoked by `release.yml`, so a blanket `|| echo` means a **release** could be built from stale package lists after an Ubuntu-archive failure, with `apt-get install` still succeeding from cache. The comment claiming only third-party failures were tolerated made it worse, not better: it asserted a property the code didn't have.

Fixed in `46db5fa29`. Tolerate the exit code, then re-read the log and abort if Ubuntu itself was the failure:

```bash
<update> 2>&1 | tee /tmp/apt-update.log || true
if grep -qE '^(E|Err):.*(archive\.ubuntu\.com|security\.ubuntu\.com)' /tmp/apt-update.log; then
  echo "::error::apt-get update failed against the Ubuntu archive - refusing to build from stale package lists"
  exit 1
fi
```

**Discriminator tested against three real apt output shapes:**

| Input | Result |
|---|---|
| `dl.google.com` Hash Sum mismatch (the outage that started this) | continue |
| `azure.archive.ubuntu.com` connect failure | **abort** |
| clean run | continue |

`azure.archive.ubuntu.com` matches the `archive\.ubuntu\.com` alternative, which is why the pattern needs no separate case for the Azure mirror.

**One thing I deliberately did not do:** `ci-pr.yml` on main (from `087ff0c`) has the same blanket-tolerance property. Lower stakes there than in a release build, and I've already rewritten that file twice in this PR's history before reverting both times — so I'm flagging it for whoever owns that commit rather than touching it a third time. Happy to send a separate PR if wanted.

YAML re-parsed for both changed files.

AgentO-asaf:
Correct — and it's the same failure shape as the finding it follows: a check that reads for one form of an error while the build emits another.

`-qq` suppresses the `Err:<n> <url>` acquisition record and reports the failure as `W: Failed to fetch ...`, and apt can exit 0 while keeping the old indexes. So an `E:`/`Err:`-only match would miss the common network case in exactly the quiet release build that matters most.

**Two changes:**

1. Match both forms — `^(E|Err):.*ubuntu` **or** `^W: Failed to fetch.*ubuntu`
2. Dropped `-qq` in `build-linux.yml`, since it hides the records the check reads

Anchored on `W: Failed to fetch` rather than any `W:`, so an ordinary warning that happens to name a source doesn't abort a release.

**Discriminator tested against six real apt output shapes:**

| Input | Result |
|---|---|
| `dl.google.com` Hash Sum mismatch | continue |
| `E: ... azure.archive.ubuntu.com` | **ABORT** |
| `W: Failed to fetch ... archive.ubuntu.com` | **ABORT** |
| `W: Failed to fetch ... security.ubuntu.com` | **ABORT** |
| `W: Target Packages ... configured multiple times` | continue |
| clean run | continue |

The fifth row is the one worth calling out — it's why the pattern anchors on `Failed to fetch` instead of accepting any warning. A regex that aborted on benign `W:` output would fail releases for no reason and get reverted within a week, which is a worse outcome than the gap it closed.

YAML re-parsed for both files. `ci-pr.yml` still byte-identical to main.

PR #3129, opened 2026-09-09
Manoz@Area54: fix(floating-pane): retune the redock dwell to 300ms

GenericAgentX-asaf:
## What

`REDOCK_DWELL_MS: 500 -> 300`. Nothing else changes — no behavioural or UI change, same interaction as #3124 shipped.

## Why

500ms was noticeably sluggish in use. The case for it was desk research — the 400–700ms band comparable hover-intent interactions use, plus matching `SPRING_SWITCH_MS` — and actually using it beats that.

This is a deliberate step below the published band, and the band is a reason to be careful rather than a reason not to: those figures describe **passive** hover, where the user hasn't declared any intent yet. A floater drag is already a committed gesture and only the destination is open, so this threshold separates *"aiming here"* from *"passing through"*, not *"meant something"* from *"meant nothing"*.

## This is not a partial return to the 180ms bug

Worth stating explicitly, since 300 sits close to the value that failed. That failure was **mechanical, not numeric**: dwell was inferred from the *absence* of hover events, so any pause armed a redock even after the velocity gate had explicitly rejected the entry (spec §3.2). Since #3124 the host emits a heartbeat and dwell is measured from confirmed samples — so 300ms of genuine stillness is a real signal in a way 180ms never was. Raising the number was never the part of #3124 that did the work.

## Also in here

`tabbar-dnd.ts`'s comment asserted the two constants were equal, which this makes false — corrected, no code change. The spec records the retune and the reasoning so the value doesn't get "restored" later by someone reading only the original rationale.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run frontend/app/workspace/` — 14 passed
- No test pins the constant (the arming tests inject their own `dwellMs`), so nothing needed updating

## Context

Supersedes the approach in #3126, which I closed unmerged — the drop guides there were a misunderstanding of scope on my part, not something wanted. The parking-zone work in the spec (§5.3) stays open and unstarted.

Spec: `docs/specs/SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md` §4

### DIFF
--- a/.changesets/1788991931-fix-floating-pane-retune-the-redock-dwell-to-300ms-o36m.md
+++ b/.changesets/1788991931-fix-floating-pane-retune-the-redock-dwell-to-300ms-o36m.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(floating-pane): retune the redock dwell to 300ms
--- a/docs/specs/SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md
+++ b/docs/specs/SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md
@@ -146,7 +146,8 @@ the wrong direction.
 | macOS spring-loaded folders | ~500-750 ms (user-adjustable) |
 | Windows `MenuShowDelay` (submenu open) | 400 ms default |
 | AgentMux spring-loaded tabs (`SPRING_SWITCH_MS`) | 500 ms |
-| AgentMux redock (today) | **180 ms** |
+| AgentMux redock (before this spec) | **180 ms** |
+| AgentMux redock (now) | **300 ms** — see the retune note below |
 
 The consensus band for "hover became intent" is **400-700 ms**. AgentMux's
 redock is the outlier by a factor of roughly 2.5.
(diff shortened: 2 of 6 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in 659f400c — good catch, and a fair one to level at this PR specifically.

You're right that it was self-inconsistent: the whole argument here is that stale threshold references are how an obsolete value gets restored later, and I then left three of them behind. All three were written by me in #3124, which makes it worse rather than better.

Rather than swap 300 in for 500 and rebuild the same trap for whoever retunes this next, I went the other way:

- `app-init.ts:136` no longer names a figure at all — it only ever needed to say "the dwell", since the point of the sentence is the per-heartbeat IPC, not the threshold.
- `SPEC_PANE_DRAG_TO_TAB_2026_07_10.md:35` and `:207` now record the sequence (180 → 500 → 300) and state that the two constants move independently, which is the fact a future reader actually needs.
- `:169` had the same problem and wasn't in your list — also updated, and it now says to read the constant rather than any figure quoted in prose.

tsc clean.

PR #3130, opened 2026-09-09
docs(cef): correct the Windows toolchain-unchanged claim; record the SDK 152 blocker

a5af:
## Summary

Windows CEF 152 build on claudius (tracked on #3108) surfaced a real blocker
and, with it, a wrong claim in the Phase A recon report.

## The claim, and why it was wrong

The report said "Windows toolchain: unchanged between 148 and 152," based on
comparing `TOOLCHAIN_HASH`, `MSVS_VERSIONS`, and `SDK_VERSION` — all
identical across the two Chromium tags. **True about the pins, wrong about
buildability.**

A real build failed at 22,219/59,112 targets:
`ui/accessibility/platform/uia_client_info_source_win.cc` (new in Chromium
152, unconditional in the Windows `BUILD.gn`) needs
`IUIAutomationClientInfo{,Source}` interfaces **absent from Windows SDK
10.0.26100.0** — the exact SDK version the pins say is unchanged, and the
only one VS Build Tools 2022's installer catalog offers.

Root cause: Google's internal `DEPOT_TOOLS_WIN_TOOLCHAIN=1` packaged
toolchain bundles a different snapshot of "26100" than the public SDK
installer. External builders get the public one and hit this; Google's own
CI doesn't, so the pin never had to move.

## The fix

Install a newer public Windows SDK
(`10.0.28000.0` confirmed to contain the missing interfaces), then edit
**two** files — `SDK_VERSION` is hardcoded separately in
`build/vs_toolchain.py` and `build/toolchain/win/setup_toolchain.py`, and
the former's own comment says they must match. The GN arg
`windows_sdk_version` alone does **not** propagate — confirmed by checking
`toolchain.ninja` before and after.

## Also recorded

- An unrelated upstream CEF bug: `cef_installer_unittests` fails to compile
  in official+release builds (`SetInstallerE2EConfigForTesting` guarded by
  `#if !(OFFICIAL_BUILD && NDEBUG)` but called unconditionally by its own
  test). Workaround: build the `libcef.dll` ninja target directly — its
  dependency graph doesn't include the test binary.
- The correct way to verify `BeginWindowDrag` landed: **not** `strings` on
  the binary (struct/vtable function-pointer fields aren't named exports and
  never appear as bare strings — confirmed against a known-present sibling
  field, which also returned zero hits). Check the generated wrapper source
  instead.

## The pattern, named explicitly

This is the **third** time in this recon effort that comparing version pins
passed where actually compiling failed — after the API-annotation bug (fixed
in a prior commit) and the invalid `strings`-based symbol check just now.
Recorded in the report as a correction with the original claim quoted, not
silently edited, same as the two prior corrections on this effort.

## Verification

- Doc-status and spec-citation gates pass locally.
- The SDK fix itself is verified by an actual successful build:
  `libcef.dll` compiled clean (305,693,184 bytes) and boot-tested via
  `cefsimple.exe` (`MainWindowTitle = "Example Domain"` after navigating,
  confirming real render, not just process launch).

### DIFF
--- a/docs/cef-build/build-patched-cef-windows.md
+++ b/docs/cef-build/build-patched-cef-windows.md
@@ -49,6 +49,53 @@ again on a future build:
   (`Microsoft.VisualStudio.Component.VC.ATLMFC` via the VS Installer's
   `modify` command, needed for `atldef.h`).
 
+**Update (2026-09-09), building CEF 152 (branch `7977`) from this same doc:**
+the steps above still work, but a **152-specific** blocker appeared that
+didn't exist at 148 — read this before starting a 152 build on any machine:
+
+- `ui/accessibility/platform/uia_client_info_source_win.cc` (new in Chromium
+  152, unconditional in the Windows `BUILD.gn` — no flag skips it) needs
+  `IUIAutomationClientInfo`/`IUIAutomationClientInfoSource` interfaces that
+  are **absent from Windows SDK 10.0.26100.0** — the only SDK version VS
+  Build Tools 2022's installer catalog offers (checked directly: only
+  `22621` and `26100` exist there). Confusingly, `build/vs_toolchain.py`'s
+  `SDK_VERSION` pin is **identical** between Chromium 148 and 152
+  (`10.0.26100.0` both), so the pins say "unchanged" while the actual
+  requirement moved. Root cause: Google's internal
+  `DEPOT_TOOLS_WIN_TOOLCHAIN=1` packaged toolchain bundles a different
+  snapshot of "26100" than the public SDK installer — external builders get
+  the public one and hit this; Google's own CI doesn't.
+  - **Fix:** install a newer public Windows SDK
+    (https://developer.microsoft.com/windows/downloads/windows-sdk/ — the
+    version matching your OS build works; `10.0.28000.0` confirmed to
+    contain the missing interfaces). The VS Installer **cannot** supply
+    this — its catalog doesn't have it.
+  - Then **edit two files**, not one — `SDK_VERSION` is hardcoded
+    separately in both `build/vs_toolchain.py` and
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

a5af:
## All four correct — fixed, plus a real fork-topology gap found while fixing #1

**P1 (hardcoded 7778):** Introduced `$CefBranch`, threaded through every step. Set it once at the top.

**P1 (step 5 still said `cef`):** Fixed the actual command, not just the intro note.

**P2 (cefsimple not built by `libcef.dll` alone):** Added an explicit `cefsimple` build step. Also fixed step 6's `Set-Location` path while I was there — it pointed somewhere unrelated to the real build output (verified against my own live session).

**P2 (report/spec contradiction):** Swept exhaustively rather than fixing the one line — found **11** instances of "nothing built"/"not compile-verified" across both docs, including one that directly contradicted a sentence two lines above it. Rewrote §5b's status and history rather than patching around it again.

## The bigger thing #1 surfaced

Fixing the hardcoded-branch issue meant actually trying `$CefBranch = "7977"`, which exposed something the doc bug was hiding: **`7977` was still the bare upstream mirror.** Unlike `7778` — where patches are merged directly into the milestone branch — the two 152 patch branches existed but were never merged into `7977`. Following this doc's own "always build from the integration branch" instruction would have silently built **zero of the four patches** — none of them are load-bearing for compilation, only for behavior, so nothing would have errored.

Fixed by merging both feature branches into `7977` directly (one `patch.cfg` conflict, resolved). `7977` is now a genuine integration branch. Verified post-push via the API.

This also means patch #4's CEF-side cascade — "ported but not compile-verified" as of the last round — is honestly upgradeable to "compiled on Windows," since the Windows build used the completed branch. Kept separate from "functionally correct," which nothing has tested.

PR #3131, opened 2026-09-09
docs(armory): naming consolidation spec + consolidated memory status

AgentA-asaf:
Two docs, both docs-only. Consolidates six weeks of Armory memory work that #2024 never caught up with, and formalises the naming fix that has to land before the open architectural questions can be discussed precisely.

## The finding

"memory" currently denotes **six** different things: Memory, Bundle, ABF, Brain, Preset, Global Memory. `bundle_list_impl` (`app_api/mod.rs:757`) uses four of them in eight lines:

```rust
let memories = state.id_store.bundle_memory_list()...   // bundle_memory_*
let bundles: Vec<_> = memories.iter()...                // memories → bundles
Ok(json!({ "bundles": bundles, "presets": bundles }))   // both keys
```

`db_bundles` is surfaced under **two** Armory tabs (ABF, and Memory → Global). `bundle.*` and the `listmemories` family are duplicate RPC surfaces over one storage method.

## Target vocabulary (spec §3)

| Name | What | AgentMux posture |
|---|---|---|
| **Bundle** | curated instructions + context + refs (`db_bundles`); scope is a *flag* (global/system/agent) | owns |
| **Memory** | what the agent learns (`db_agent_native_memory`) | mirrors + versions |
| **Project Instructions** | repo `CLAUDE.md`/`AGENTS.md`/… | **reads only**, never writes |
| **ABF** | the export format | — |

Retired: Brain, Preset, Global Memory.

## The spec corrects its own earlier draft

The status doc's §4 proposed "Global Memory → Instructions". That was wrong: a `db_bundles` row carries seven payload fields, so naming the container after one of them is a new inaccuracy. Global bundles are bundles with a scope flag. Recorded as a correction (§2.4) rather than silently revised.

## What is explicitly NOT renamed (§4, normative)

The `MemoryWrite/Read/History/…` MCP tools, `agent:memory:*`, `db_agent_native_memory*`, and all `NativeMemory*` code are **already correct** and agent-facing — renaming breaks running agents for nothing. ABF `components.*` keys are unchanged; `components.memory` is already right, and import has no unknown-key warning, so any key rename would silently drop data from bundles already exported.

## Phasing (§5)

By blast radius. Phases 1–2 (internal names, retire "brain") carry no contract risk. Phase 3 (retire `*memory` RPCs for `bundle.*`) is wire-breaking and must first retire the still-agent-facing `PresetList`/`PresetGet` MCP tools, which read the `presets` key. Phase 4 folds the two bundle tabs into one. DB column renames are deferred with a reason: `migrations.rs` has no `RENAME COLUMN` precedent.

Two traps the inventory surfaced: `test/contract/rpc-contract.test.ts` does not baseline the seven retiring commands, so it won't catch their removal — add them first. And `check-rpc-bindings.sh` uses `git status --porcelain -uall`, so stale generated bindings must be `git rm`'d or they pass silently.

## Follow-on, recorded not dropped (§8)

Three portability requirements agreed this session: export carries Memory (`components.memory` exists, import honours it, exporter just never emits it), Project Instructions are tracked read-only, and the memory-location invariant from `SPEC_NATIVE_MEMORY_DURABLE_SYNC`. Own spec, after naming lands.

### DIFF
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -363,6 +363,7 @@ partial list.
 | [`SPEC_AGENT_ZOOM_PERSISTENCE_2026_06_22`](SPEC_AGENT_ZOOM_PERSISTENCE_2026_06_22.md) | Per-agent zoom persistence |
 | [`SPEC_ARMORY_DROP_HOST_CLI_CONFIG_BLOCK_2026_09_01`](SPEC_ARMORY_DROP_HOST_CLI_CONFIG_BLOCK_2026_09_01.md) | Spec: Drop the "Claude Code — host CLI config" block from Armory Global Memory |
 | [`SPEC_ARMORY_MEMORY_TAB_MERGE_2026_08_30`](SPEC_ARMORY_MEMORY_TAB_MERGE_2026_08_30.md) | Spec: Armory rail — merge "Global Memory" + "Personal Memory" into one "Memory" tab |
+| [`SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09`](SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md) | Spec: Armory Naming Consolidation — Bundle vs Memory |
 | [`SPEC_ARMORY_PERSONAL_MEMORY_AGENT_BLOCKS_2026_09_01`](SPEC_ARMORY_PERSONAL_MEMORY_AGENT_BLOCKS_2026_09_01.md) | Spec: Armory → Memory → Personal — browse by agent block, not a dropdown |
 | [`SPEC_ARMORY_PERSONAL_MEMORY_FILE_TILES_2026_09_04`](SPEC_ARMORY_PERSONAL_MEMORY_FILE_TILES_2026_09_04.md) | Spec: Armory → Memory → Personal — file tiles, not a dropdown |
 | [`SPEC_ARMORY_PERSONAL_MEMORY_FILTER_AND_SORT_2026_09_02`](SPEC_ARMORY_PERSONAL_MEMORY_FILTER_AND_SORT_2026_09_02.md) | Spec: find/filter and sort for Armory → Memory → Personal |
--- a/docs/specs/SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md
+++ b/docs/specs/SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md
@@ -12,8 +12,12 @@
 
 Two architectural questions are currently open and blocked on vocabulary:
 
-- ABF cannot round-trip memory (export never emits `components.memory`).
-- Curated memory is structurally empty for every agent and nothing surfaces it.
+- The agent-less `bundle.export` omits `components.memory` while the agent-scoped
+  `bundle.export_for_agent` includes it — whether that asymmetry is the intended
+  design or a gap is unclear from the code alone.
+- Curated memory is structurally empty for every agent, and while the Armory UI
+  shows the empty state, nothing at launch tells the operator the agent received
+  no `# Memory` section.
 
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentA-asaf:
All four P2s were correct — each verified against code before editing, fixed in `e019a4fbd`.

**Memory export (spec §1/§8, status §5)** — the important one. `bundle.export_for_agent` does splice `components.memory` (`splice_memory_component`, `bundle.rs:439-468`, tested `:2761`); only agent-less `bundle.export` omits it, by design. My "export never emits memory" claim conflated `bundle_export.rs` with the full export surface — my own inventory had said "agent-scoped path only" and I misread it. Narrowed to the real asymmetry: the agent-less path drops memory *silently* while the import side warns via `MEMORY_COMPONENT_IGNORED_WARNING`. This also means a requirement the repo owner raised this session ("export must carry personal memory") is already implemented.

**Empty-state visibility (status §6)** — correct, `GlobalBrainManager` renders `No global sections yet.` / `+ New section` / `(empty)` (`global-brain-manager.tsx:327-388`). Narrowed to the launch side: nothing at `agent.open` tells the operator the agent received no `# Memory` section.

**Schema v32 (spec §5 deferred)** — `db_agent_definitions`/`db_agent_instances` are retired; live columns are `db_agents.memory_id` (`:578`) and `db_agents.default_memory_id` (`:639`). The `:999` `ALTER TABLE db_agent_definitions` is a legacy adoption entry, not live schema. Fixed.

**`agent_handlers/memory.rs` (spec §5 Phase 1)** — zero native-memory references, confirmed; native-memory RPCs already live in `native_memory_handlers.rs`. Changed to a plain rename → `agent_handlers/bundle.rs`. The split still applies to `rpc_types/memory.rs` (19 `NativeMemory*` types alongside bundle types).

Also in that commit: regenerated `docs/specs/INDEX.md` — the docs gate diffs the committed index against `gen-docs-index.sh` output and the new spec's row was missing.

AgentA-asaf:

All four items in the `CHANGES_REQUESTED` review are the four Codex P2s, and all four were **already fixed in `e019a4fbd`** — the exact commit the review is attached to. The review appears to have restated Codex's findings without re-reading the head content.

Evidence, from the remote at that SHA (not my working tree):

```
$ gh api repos/agentmuxai/agentmux/contents/<file>?ref=e019a4fbd | base64 -d | grep -c "Codex P2 on #3131"
  docs/specs/SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md:   3
  docs/status/STATUS_ARMORY_MEMORY_CONSOLIDATION_2026_09_09.md: 2
```

Per item, at `e019a4fbd`:

- **[P0] status §6 "empty state invisible"** — rewritten: acknowledges `GlobalBrainManager` renders `No global sections yet.` / `+ New section` / `(empty)` (`global-brain-manager.tsx:327-388`), and narrows the gap to the launch side.
- **[P1] spec §8 "memory export deferred"** — rewritten to "already implemented for the agent-scoped path", citing `splice_memory_component` (`bundle.rs:439-468`) and the test at `:2761`. Spec §1 and status §5 updated to match.
- **[P1] spec §5 deferred "pre-v32 tables"** — rewritten to `db_agents.memory_id` (`:578`) / `db_agents.default_memory_id` (`:639`), noting `db_agent_definitions`/`db_agent_instances` are retired and `:999` is a legacy adoption entry.
- **[P2] spec §5 Phase 1 "split agent_handlers/memory.rs"** — changed to a plain rename → `agent_handlers/bundle.rs`, with the split retained only for `rpc_types/memory.rs` (19 `NativeMemory*` types).

Same pattern as #2860 (review recorded against the fix commit but reporting pre-fix content). Requesting a fresh pass so the review state reflects the actual head.

PR #3132, opened 2026-09-09
layout: stop evicting a leaf's NodeModel on an in-pane tab switch

GenericAgentX-asaf:
## Summary
- PR 1 of 3 for the agent-pane header/tab-strip flash fix
  (`docs/specs/SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md`).
- `activeKeyFor` now keys a leaf on its bare `node.id` instead of
  `nodeId:activeBlockId`, so an in-pane tab switch (e.g. Agent \u2194 Agent
  History) no longer remounts the leaf's whole subtree.
- `layoutStack.ts`'s three mutators (`pushBlockOntoStack`,
  `setActiveBlockInStack`, `closeBlockInStack`) no longer dispose the leaf's
  cached `NodeModel` on an active-member switch, since a still-mounted
  component now depends on it staying live.
- Added an additive, reactive `NodeModel.activeBlockId` field
  (`types.ts`/`layoutNodeModels.ts`) for leaf-level chrome that needs to
  track the active stack member without itself remounting \u2014 the existing
  frozen `blockId` field and all its consumers are untouched.
- This PR is layout-engine plumbing only; it does not yet move the pane
  header/tab strip into a shared chrome layer (PR 2/3).

## Test plan
- [x] `npx vitest run frontend/layout/tests/layoutStack.test.ts` \u2014 24/24
      passing, including inverted frozen-vs-live assertions for all three
      mutators and a new `activeBlockId` memo test.
- [x] `npx tsc -p tsconfig.json --noEmit` \u2014 clean.
- [ ] Manual visual verification is deferred to PR 3, where the actual
      chrome-hoisting lands \u2014 this PR alone doesn't change what's on
      screen yet, only what survives a switch underneath it.

### DIFF
--- a/.changesets/1788997152-refactor-layout-add-inert-nodemodel-activeblockid-groundwork-o2iz.md
+++ b/.changesets/1788997152-refactor-layout-add-inert-nodemodel-activeblockid-groundwork-o2iz.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(layout): add inert NodeModel.activeBlockId groundwork for pane chrome-stability fix
--- a/frontend/layout/lib/layoutNodeModels.ts
+++ b/frontend/layout/lib/layoutNodeModels.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { createSignalAtom, fireAndForget } from "@/util/util";
+import { findNode } from "./layoutNode";
 import type { Properties as CSSProperties } from "csstype";
 import { createMemo, createRoot } from "solid-js";
 import { LayoutNode, LayoutNodeAdditionalProps, NodeModel } from "./types";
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed in 2d198308a — thanks, both ReAgent and Codex caught a real regression.

You were right: `activeKeyFor` dropping `activeBlockId` and the three `layoutStack.ts` mutators no longer disposing the `NodeModel`, on their own, break the shipped in-pane tab-switch feature — `TabContent`/`Block` have no remount boundary of their own, so the outer whole-leaf remount this PR touched was the only thing updating displayed content on a switch.

Reverted `activeKeyFor` and the three mutators back to shipped behavior (leaf remounts on an active-member switch, `NodeModel` disposed, exactly as before). Kept only the additive, currently-inert `NodeModel.activeBlockId` field — real prep work with zero behavior change, no production consumer yet. The `activeKeyFor` change and the disposal removal will land together with the `pane-leaf-chrome.tsx` inner remount boundary in one deployable change, not split across PRs, per your finding.

Also added the missing changeset (P2).

GenericAgentX-asaf:
Good catch — 680fc8235 fixes this. My revert had reshaped the condition to \`node.data?.blockStack?.length ? ... : node.id\` instead of the original \`node.data?.activeBlockId ? ... : node.id\`.

Checked whether they can actually diverge: \`layoutStack.ts\`'s \`setActive()\` and \`closeBlockInStack\`'s neighbor-picking branch are the only writers of either field in the whole codebase, and both always set (and never clear) \`blockStack\`/\`activeBlockId\` together — so in every reachable state today they agree. But there's no reason to carry a differently-shaped-but-equivalent condition when the point of this commit is an exact revert to shipped behavior. Now matches main's expression verbatim (confirmed via \`git diff main\`).

PR #3134, opened 2026-09-09
refactor(block): BlockFrame_Header/EndIcons take a reactive blockId accessor

GenericAgentX-asaf:
## Summary
- PR 2 of 3 for the agent-pane header/tab-strip flash fix
  (`docs/specs/SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md`).
  Independent of PR #3132 (PR 1) — touches disjoint files, can merge in
  either order.
- `BlockFrame_Header` and `EndIcons` now take an explicit
  `blockId: Accessor<string>` prop alongside the existing `nodeModel`,
  instead of reading `nodeModel.blockId` directly. Every existing call
  site passes a trivial `() => nodeModel.blockId` wrapper —
  behavior-preserving everywhere already in use.
- `BlockFrame_Header`'s block-data lookup switches from
  `WOS.useWaveObjectValue` (ref-counted via `onCleanup` tied to this
  component's mount, never re-subscribes if the oref changes later) to a
  `getWaveObjectAtom` call inside a `createMemo` keyed on `blockId()` (no
  such lifecycle coupling — safe to call fresh with a changing oref).
- `nodeModel` stays unchanged for everything genuinely leaf-scoped
  (focus/magnify/minimize/close) — only block-identity reads move to the
  new accessor.
- This alone changes nothing on screen; it's mechanical prep so a future
  hoisted pane-chrome caller (PR 3) can pass `nodeModel.activeBlockId`
  instead of the trivial wrapper, to track an active-member switch
  without this header remounting.

## Test plan
- [x] `npx vitest run` — full suite, 3890 passing / 3 skipped, no
      regressions.
- [ ] No dedicated test file exists for `blockframe.tsx`; behavior
      preservation here relies on the mechanical nature of the change
      (every read-site swap is a like-for-like accessor call) plus the
      full suite passing. Manual visual verification is deferred to PR 3,
      where this plumbing actually gets used by a switch-surviving chrome
      layer.

### DIFF
--- a/.changesets/1788997373-refactor-block-blockframe-header-endicons-take-a-reactive-bl-iiw9.md
+++ b/.changesets/1788997373-refactor-block-blockframe-header-endicons-take-a-reactive-bl-iiw9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+refactor(block): BlockFrame_Header/EndIcons take a reactive blockId accessor
--- a/frontend/app/block/blockframe.tsx
+++ b/frontend/app/block/blockframe.tsx
@@ -27,7 +27,7 @@ import { NodeModel } from "@/layout/index";
 import * as util from "@/util/util";
 import { computeBgStyleFromMeta } from "@/util/waveutil";
 import clsx from "clsx";
-import type { JSX } from "solid-js";
+import type { Accessor, JSX } from "solid-js";
 import { createEffect, createMemo, createSignal, For, onCleanup, onMount, Show } from "solid-js";
 import { CopyButton } from "../element/copybutton";
 import { detectAgentColor, detectAgentFromEnv, detectAgentTextColor, getEffectiveTitle, isUsableFocusRingColor } from "./autotitle";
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Good catch, and correct: \`ConnectionButton\` (\`blockutil.tsx:89\`) destructures \`connection\` directly in its function signature, which in SolidJS freezes it at first-invocation value — it won't track a later change to the SAME mounted instance's \`connection\` prop.

This PR alone doesn't expose it yet: nothing in production passes anything other than the static \`() => nodeModel.blockId\` wrapper into the new accessor, so the header still remounts whenever the active block actually changes, same as before. It becomes live once PR 3 (the chrome-hoisting work this is prep for) actually keeps a header mounted across a real blockId switch.

Fixing \`ConnectionButton\` to read \`props.connection\` reactively (a memo instead of a destructured snapshot) as part of PR 3, where the switching behavior actually exists to test against, rather than as an untestable-in-isolation patch here.

PR #3136, opened 2026-09-10
fix(agent): stop the pane header/tab-strip flash on Agent <-> History switch

GenericAgentX-asaf:
## Summary

PR 3 of 3 for `docs/specs/SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md` (PR 1 = #3132, PR 2 = #3134, both merged). This lands the actual visible fix, deferred here because PR 1 alone (dropping `activeBlockId` from the leaf's remount key without a replacement mechanism) turned out to be a real regression — caught by ReAgent/Codex — that could only be fixed correctly by landing the replacement mechanism in the *same* change as the removal.

- New `frontend/app/tab/pane-leaf-chrome.tsx`: the router `tabcontent.tsx` now renders instead of `<Block>` directly. Always wraps a per-active-member, `<Key>`-scoped `<Block>` (via a block-scoped `NodeModel` wrapper, since the leaf's own `NodeModel.blockId` is frozen) — this inner remount is what updates displayed content on a switch now. When the active member's effective view type is `"agent"` and the leaf's stack has ever had 2+ members (`NodeModel.hasEverBeenMultiMember`, new, monotonic), wraps that in the active `ViewModel`'s own `renderPaneChrome` — a stable outer shell that survives every subsequent switch. Every other case is byte-for-byte today's passthrough.
- `layoutStack.ts`: the three block-stack mutators no longer dispose the leaf's `NodeModel` on an active-member change. `tilelayout-shared.tsx`: `DisplayNodesWrapper`'s outer `<Key>` now keys leaves on `node.id` alone (`OverlayNodeWrapper` deliberately unchanged — drag-overlay geometry, not content). Landing these together with the router above, per `layoutStack.ts`'s own header comment, is exactly what PR 1 got wrong.
- `NodeModel` gains `activeViewModel`/`setActiveViewModel` (`types.ts`, `layoutNodeModels.ts`), pushed by `block.tsx`'s existing `ViewModel` construct/cleanup effect (owner-checked the same way `unregisterBlockComponentModel` already is) — hoisted chrome's only live pointer to a callable `ViewModel`, since the global block-component registry is a plain `Map` (not reactive) and doesn't survive `Block`'s own dispose-on-unmount on every switch, including a repeat one.
- `agent-view.tsx` splits `AgentViewWrapper` into `AgentBlockContent` (becomes `AgentViewModel.viewComponent`, constructed fresh per stack member, unchanged internals) and `AgentPaneChrome` (the tab strip + progress-bar slot, mounted once per leaf via `renderPaneChrome`, bridges `progressBarMount`/`tabStripVisible` back to whichever `ViewModel` is currently active). `handleTabSwitch`/`handleTabClose`'s reveal gates are dropped on the branches only reachable once `hasEverBeenMultiMember()` is already true — chrome is hoisted and stable by construction by then.
- `blockutil.tsx`'s `ConnectionButton` fixed to read its `connection` prop reactively (a real latent bug Codex caught on PR #3134 — harmless there since nothing routed a changing blockId into a live header yet; this PR is what actually exercises it).

## Test plan

- [x] `npx vitest run` — full suite, 3900 passing / 3 skipped, no regressions.
- [x] New `frontend/app/tab/pane-leaf-chrome.test.tsx` — the spec's own acceptance criterion: mount, capture the hoisted chrome's DOM node reference, drive a switch, assert `===` on it afterward while the switch-scoped block content gets a genuinely fresh element.
- [x] `layoutStack.test.ts`'s frozen-vs-live assertions inverted (again — the first inversion on PR 1 was reverted when the regression was found) plus new coverage for `hasEverBeenMultiMember` and `activeViewModel`/`setActiveViewModel`.
- [ ] **Manual, in a live build — NOT done by me.** I don't have a way to visually verify a different pane's UI in this environment (AgentMux's own `UIScreenshot`/`UIClick`/`UIQuery` tools are scoped to my own pane only). The DOM/CSS containment chain changed meaningfully — `.agent-pane-stack` now sometimes sits OUTSIDE the leaf's own `BlockFrame` wrapper instead of always inside `.block-content` (see `agent-view.scss`'s updated comments for the full reasoning) — and genuinely needs a human or dev-server check before merge:
  - Switch Agent↔History several times — header/tab-strip must not flash on a repeat switch.
  - First History activation for a fresh agent — expected to still flash, per confirmed scope (not a bug).
  - "+" (new tab) / Quick Fork — existing gated settle must be unchanged.
  - Progress bar renders in the right place while an agent is busy, including after switching tabs.

### DIFF
--- a/.changesets/1789000090-fix-agent-stop-the-pane-header-tab-strip-flash-when-switchin-u0r3.md
+++ b/.changesets/1789000090-fix-agent-stop-the-pane-header-tab-strip-flash-when-switchin-u0r3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent): stop the pane header/tab-strip flash when switching Agent <-> History tabs
--- a/frontend/app/block/block.tsx
+++ b/frontend/app/block/block.tsx
@@ -39,7 +39,14 @@ import { useSubagentBackfillGate } from "@/app/view/agent/hooks/useSubagentBackf
 // overlay and picker-fade, AgentPicker's own overlay).
 const READY_GATE_FADE_MS = 200;
 
-function makeViewModel(blockId: string, blockView: string, nodeModel: NodeModel): ViewModel {
+/**
+ * Applies the view-type migration/rename redirects below. Exported so
+ * `pane-leaf-chrome.tsx` can apply the identical rules when deciding
+ * whether a stack member's EFFECTIVE view is "agent" — e.g. a still-live
+ * "forge" block must route through the same redirect `makeViewModel`
+ * itself applies, not just a raw `meta.view === "agent"` check.
+ */
+export function resolveEffectiveViewType(blockView: string): string {
     // Migration shims:
     //   * v0.33.197: forge was folded into the agent pane; redirect old
     //     "forge" blocks to "agent" so they keep rendering.
@@ -61,6 +68,11 @@ function makeViewModel(blockId: string, blockView: string, nodeModel: NodeModel)
     if (effectiveView === "forge") effectiveView = "agent";
     if (effectiveView === "workflows") effectiveView = "drone";
     if (effectiveView === "trust") effectiveView = "armory";
+    return effectiveView;
+}
+
+function makeViewModel(blockId: string, blockView: string, nodeModel: NodeModel): ViewModel {
+    const effectiveView = resolveEffectiveViewType(blockView);
     const ctor = getBlockViewClass(effectiveView);
     if (ctor != null) {
         return new ctor(blockId, nodeModel as any);
@@ -291,9 +303,23 @@ function Block(props: BlockProps): JSX.Element {
             registerBlockComponentModel(props.nodeModel.blockId, registeredBcm);
         }
         setViewModel(vm);
+        // See NodeModel.activeViewModel/setActiveViewModel's own doc
+        // comments (layout/lib/types.ts) — hoisted pane chrome's only live
+        // pointer to a callable ViewModel, since the global registry
+        // doesn't survive this component's own dispose-on-unmount. Owner
+        // matches the SAME object identity `unregisterBlockComponentModel`
+        // below would be called with, so an adopting (not creating) mount's
+        // eventual cleanup correctly no-ops instead of clobbering the
+        // creating mount's still-live registration.
+        props.nodeModel.setActiveViewModel?.(vm, registeredBcm ?? bcm);
     });
 
     onCleanup(() => {
+        // Owner is `registeredBcm` here too (not `registeredBcm ?? bcm`) —
+        // an adopting mount never created its own registration object, so
+        // this correctly no-ops for it instead of clearing the creating
+        // mount's still-live activeViewModel out from under it.
+        props.nodeModel.setActiveViewModel?.(null, registeredBcm);
         if (registeredBcm) {
             unregisterBlockComponentModel(props.nodeModel.blockId, registeredBcm);
         }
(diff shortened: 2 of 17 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Good catch, fixed in 73c20f59f. You were right — resolving the owning node via \`getNodeByBlockId(anchorBlockId)\` breaks permanently once the user closes that specific tab, and \`hasEverBeenMultiMember\` being monotonic means \`AgentPaneChrome\` never gets a chance to unmount and recover.

Switched all four call sites (\`stackTabs\`, \`handleTabSwitch\`, \`handleNewAgentTab\` x2) to resolve via \`findNode(layoutModel.treeState.rootNode, nodeModel.nodeId)\` instead — the leaf's own stable id, exactly as you suggested. \`handleTabClose\`'s existing \`getNodeByBlockId(targetBlockId)\` is untouched (correctly resolves the node owning whatever tab is being closed, which may be a different pane in the cross-pane-fork case — not affected by this bug).

GenericAgentX-asaf:
Both real, fixed in b8743501f.

1. **Missing replacement header**: \`AgentPaneChrome\` now reuses \`BlockFrame_Header\` directly (exported from \`blockframe.tsx\` for exactly this — PR #3134 built its reactive \`blockId\` accessor param specifically for a future hoisted-chrome caller and it went unused until now). Reads the same \`activeBlockId\` accessor the tab strip already uses; \`nodeModel\` stays the leaf-level one for magnify/close/drag-handle. \`manageConnection\` is never set by \`AgentViewModel\`, so the connection-button branch is inert either way — \`changeConnModalAtom\`/\`connBtnRef\` are unwired stand-ins for that reason, not an oversight.

2. **CEF \`[data-blockid]\` scoping**: added the same attribute (same value, when this pane is the active member) to \`AgentPaneChrome\`'s own persistent root. \`document.querySelector\` returns the first match in document order, and this outer element now precedes the nested \`<Block>\`'s own copy, so it resolves to the wider root that actually contains chrome.

GenericAgentX-asaf:
Fixed in 9c168e0d1. Moved the \`useAgentDefinitions()\` subscription into \`AgentViewModel\`'s own constructor (one per ViewModel instance) and exposed it as a plain \`agentDefinitions\` field — \`AgentBlockContent\` reads \`model.agentDefinitions\` directly, \`AgentPaneChrome\` reads \`nodeModel.activeViewModel()?.agentDefinitions\`, same bridging pattern already used for \`progressBarMount\`/\`tabStripVisible\`. No more second RPC/subscription.

GenericAgentX-asaf:
Real, and confirmed it empirically before trusting my own reasoning about \`<Show>\`'s edge semantics — wrote a minimal standalone repro of the sync-cleanup/deferred-effect timing, and the chrome-render counter went 1\u21922 on a single switch, exactly as you predicted. Fixed in c82d60e9a.

Latched the first non-null vm into a plain closure variable inside a memo — once captured, the memo's own return value stops changing (same reference), so no consumer (including \`<Show>\`) ever sees a falsy value again after the first real capture, regardless of how \`activeViewModel()\` itself blips underneath.

You were also right about my test's blind spot: the mocked \`Block\` never actually called \`setActiveViewModel\` at all, so \`activeViewModel()\` was a constant accessor that structurally couldn't exercise this transition. Rewrote it to drive the real sync-cleanup/deferred-effect timing, and added a direct regression test asserting \`renderPaneChrome\`'s call count stays at 1 across three switches \u2014 confirmed it fails at 2 without the fix, then restored the fix and confirmed green.

Issue #3140, opened 2026-09-10
tracking: make nightly CI actually robust — honest failure reporting, immediate alerting, pre-merge macOS coverage


AgentO-asaf:
## Why this exists

Two real macOS-only breaks reached `main` and sat there, unnoticed, for multiple nights each:

- **#3118** (fixed) — `agentmux-cef` failed to *compile* on macOS for 2 nights (9/8, 9/9). No PR-time leg ever builds `agentmux-cef` for macOS, so this was structurally uncatchable before merge — the nightly `sign+notarize` job was the only thing that ever tries.
- **#3137** (open, this issue's trigger) — two `lan_listeners` tests have been failing on macOS in `ci-nightly-build.yml`, and per that issue's own body, "this has been silently failing... without blocking anything." Found only because someone manually triggered a run and read the job log.

Both were real, both were silent. This issue tracks fixing the *process*, not just these two instances.

## #3137 — resolved empirically, included here as Workstream 0

Reproduced the exact bind sequence from both failing tests on a **physical macOS 26.5.2 Mac** (not a CI VM), 3x each, with a real LAN IP:

- `0.0.0.0` bound after an existing `127.0.0.1` on the same port → **no conflict**
- A specific LAN IP bound after an existing `0.0.0.0` → **no conflict**

Identical to what the GH Actions runner measured — **not** a CI-sandbox/virtual-NIC artifact (that was hypothesis 2 in #3137; ruled out). **macOS's bind semantics actually group with Windows here, not Linux** — the tests' `#[cfg(not(windows))]` grouping is factually wrong (hypothesis 1, confirmed).

One thing beyond what #3137 asked: when both sockets coexist on macOS, does it have Windows' "ambiguous accept" hazard? Tested directly — the kernel routes incoming connections by most-specific-match, deterministically. A connection to the LAN IP always reaches the LAN-IP listener, never the wildcard one. So the *design rationale* (`STARTUP_BIND_ADDR` = loopback-only, to stop a wildcard-then-specific boot race from silently disabling mDNS) doesn't fail the same way on macOS as on Linux — the specific bind **succeeds** instead of erroring, so the "disable mDNS on bind failure" path never fires. The actual defect on macOS is milder: a leaked, never-closed wildcard socket, not a silent discovery outage.

## Root causes (evidence-backed, not guesses)

[...]

PR #3141, opened 2026-09-10
fix(net): lan_listeners macOS bind-conflict assumption was wrong — groups with Windows, not Linux

### LINKED ISSUE #3137: macOS: lan_listeners platform-dependent bind-conflict tests fail on GH Actions macos-latest (behavior doesn't match assumed Linux/macOS parity)
## Summary

Manually triggered `ci-nightly-build.yml` on `main` (commit `a8e8ecce4`, after merging #3096) to get a live macOS confirmation. Windows and Linux legs are fully green (compile + full `cargo test --workspace`). The macOS leg compiled fine but **2 tests failed**:

- `backend::lan_listeners::tests::wildcard_vs_loopback_conflict_is_platform_dependent`
- `backend::lan_listeners::tests::wildcard_then_specific_conflict_is_platform_dependent`

Run: https://github.com/agentmuxai/agentmux/actions/runs/34415980065 (job `cargo build + test (macos-latest)`, ID 102680845849)

**Not related to #3096** (the agent-consolidation schema drop that prompted this check) — these tests are pure TCP socket bind-conflict probes in `agentmux-srv/src/backend/lan_listeners.rs`, unrelated to storage/migrations. The macOS leg only overall-reports "success" because it's `continue-on-error` in that workflow, so this has been silently failing (or never actually run to completion — see below) without blocking anything.

## Failure detail

```
---- backend::lan_listeners::tests::wildcard_then_specific_conflict_is_platform_dependent stdout ----
192.168.64.9:49263 vs existing 0.0.0.0:49263 — conflicts: false (expected: true on Linux/macOS, false on Windows)

thread '...' panicked at agentmux-srv/src/backend/lan_listeners.rs:551:9:
expected the per-address bind to be refused under a wildcard on this platform — this conflict is the whole reason startup binds loopback only; got None

---- backend::lan_listeners::tests::wildcard_vs_loopback_conflict_is_platform_dependent stdout ----
0.0.0.0:49264 vs existing 127.0.0.1:49264 — conflicts: false (expected: true on Linux/macOS, false on Windows)

thread '...' panicked at agentmux-srv/src/backend/lan_listeners.rs:504:9:
expected the wildcard bind to conflict with loopback on this platform; if it no longer does, the same report's platform note is stale

test result: FAILED. 3192 passed; 2 failed; 6 ignored
```

Both tests measure actual OS `bind()` behavior at runtime (not mocked) and assert on it — see the doc comments right above each test (`agentmux-srv/src/backend/lan_listeners.rs` ~L470-555) and `docs/reports/REPORT_NETWORK_ARCHITECTURE_DRYNESS_AND_ROBUST_LAN_2026_09_06.md` (search "wildcard"/"platform-dependent"). Both assert `true` (conflict) for "Linux/macOS" and `false` for Windows via `#[cfg(not(windows))]` / `#[cfg(windows)]`. On this GH Actions `macos-latest` runner, both measured `conflicts: false` — i.e. this runner's kernel is allowing a wildcard (`0.0.0.0`) socket and a specific-address/loopback socket to coexist on the same port, which the code comments say should only happen on Windows.

## Why this needs a real macOS investigation, not just a code fix

The report doc (`REPORT_NETWORK_ARCHITECTURE_DRYNESS_AND_ROBUST_LAN_2026_09_06.md` line ~334) says the Linux/macOS conflict assumption was "verified on Windows, asserted in CI on Linux" — it does not claim macOS was ever empirically verified, just assumed by BSD-socket-family analogy to Linux. `ci-pr.yml` (the PR gate) never runs the macOS leg at all — only the nightly workflow does, and it's possible this is the first time these specific tests actually executed to completion on a real macOS runner. Two non-exclusive possibilities:

1. **The runner's actual behavior is correct and the code comment's macOS assumption is wrong** (macOS's `bind()`/`SO_REUSEADDR` defaults differ from Linux here) — in which case the tests' `#[cfg(not(windows))]` branch needs to split macOS from Linux, and (more importantly) `STARTUP_BIND_ADDR`'s loopback-only design (whose whole rationale per the doc comments is this exact conflict) needs re-auditing for whether it still achieves its goal on macOS.
2. **Something about the GH Actions macos-latest sandboxed/virtualized network stack** produces different bind semantics than a real Mac (e.g. `192.168.64.9` looks like it could be an Actions-runner-internal virtual bridge address, not a real LAN NIC) — in which case the test's real-world validity for actual end-user Macs is unaffected, but the test itself needs to either skip or account for CI-specific network topology.

Either way this needs someone with real macOS access to reproduce locally and determine which. If it's (1), there may be a real functional bug in the LAN-discovery bind sequencing on macOS (the doc comments describe a boot-time mDNS-disable bug that loopback-only binding was supposed to prevent — worth confirming that fix still holds under macOS's actual measured behavior).

## Scope

Filed per repo owner's request after the nightly run surfaced this — not something to fix as part of #3096's line of work. Flagging for whoever picks up macOS-specific investigation.

AgentO-asaf:
Closes #3137 (WS0 of tracking issue #3140).

## What #3137 found

Two `lan_listeners` tests failing on the `macos-latest` nightly runner, both asserting platform-specific `bind()` behavior via `#[cfg(windows)]` / `#[cfg(not(windows))]` — expecting macOS to behave like Linux.

## What I verified

Reproduced the exact bind sequences on **physical macOS 26.5.2 hardware** (3x each, real LAN IP, not a CI VM), to settle the issue's two open hypotheses:

```
0.0.0.0 bound after existing 127.0.0.1, same port  -> no conflict
specific LAN IP bound after existing 0.0.0.0        -> no conflict
```

Identical to what the GH Actions runner measured — rules out the CI-sandbox/virtual-NIC theory. **macOS's bind semantics group with Windows here, not Linux.** The `#[cfg(not(windows))]` grouping was simply wrong from the day it was written — asserted by BSD-socket-family analogy to Linux, never actually measured on macOS.

## Went one step further than the issue asked

When both sockets coexist on macOS, does it have Windows' ambiguous-accept hazard? Tested directly: bound both, connected to each address, watched which listener answered. **The kernel routes by most-specific match, deterministically** — a connection to the LAN IP always reaches the LAN-IP listener, never the wildcard one.

So the actual consequence on macOS of the boot-race `STARTUP_BIND_ADDR` defends against is neither Linux's outcome (silent mDNS-disable — the specific bind doesn't fail) nor Windows' outcome (ambiguous accept — routing is deterministic). It's a **leaked, never-closed wildcard listener**. Milder than either, but still worth not racing against — `STARTUP_BIND_ADDR`'s loopback-only design is unaffected by this correction; it was already the right call for reasons independent of this specific bug.

## Changes

- Both tests: `#[cfg(windows)]`/`#[cfg(not(windows))]` → `#[cfg(target_os = "linux")]`/`#[cfg(any(windows, target_os = "macos"))]`, matching measured behavior.
- Corrected the same "Linux/macOS" grouping claim everywhere it appeared: the module doc comment, `STARTUP_BIND_ADDR`'s own doc comment, and **three separate places** in `REPORT_NETWORK_ARCHITECTURE_DRYNESS_AND_ROBUST_LAN_2026_09_06.md` — including one spot where the report's own "correction" section (written to fix an earlier wrong universal claim) repeated the *exact same* unmeasured-analogy mistake in the sentence recording the fix.

## Verification

- Both previously-failing tests pass on this Mac.
- `cargo test -p agentmux-srv -- --test-threads=1`: **3195 passed, 0 failed** — first fully-green run of the whole suite on this machine this session.
- Doc gates (status vocabulary, index, citations): all clean.

### DIFF
--- a/.changesets/1789015007-fix-net-lan-listeners-macos-bind-conflict-assumption-was-wro-heoo.md
+++ b/.changesets/1789015007-fix-net-lan-listeners-macos-bind-conflict-assumption-was-wro-heoo.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(net): lan_listeners macOS bind-conflict assumption was wrong — groups with Windows, not Linux
--- a/agentmux-srv/src/backend/lan_listeners.rs
+++ b/agentmux-srv/src/backend/lan_listeners.rs
@@ -29,14 +29,23 @@
 //! instead of shipping a silently broken toggle.
 //!
 //! Note what is deliberately NOT claimed: that `0.0.0.0:PORT` conflicts with an
-//! existing `127.0.0.1:PORT`. That is true on Linux/macOS but **false on
-//! Windows** (measured 2026-09-06 — absent `SO_EXCLUSIVEADDRUSE` the wildcard
-//! binds alongside loopback). An earlier draft asserted the universal version
-//! and the test failed on Windows, which is what caught it. The consequence is
-//! that "just bind the wildcard when enabled" is not a portable shortcut:
-//! on Windows it would leave two sockets on one port with ambiguous accept
-//! behaviour. Binding specific addresses has no such ambiguity anywhere. See
-//! `wildcard_vs_loopback_conflict_is_platform_dependent`.
+//! existing `127.0.0.1:PORT`. That is true on **Linux only** — **false on
+//! Windows AND macOS** (Windows measured 2026-09-06, absent
+//! `SO_EXCLUSIVEADDRUSE` the wildcard binds alongside loopback; macOS measured
+//! 2026-09-09 on real hardware — issue #3137 — after this module's own
+//! `#[cfg(not(windows))]` test grouping turned out to be wrong and failed
+//! nightly on a macOS runner for two nights before anyone looked). An earlier
+//! draft asserted the universal version and the test failed on Windows, which
+//! is what caught that half; macOS's own divergence from Linux went
+//! unnoticed for longer because nothing ran this test on macOS until the
+//! nightly build did. The consequence is that "just bind the wildcard when
+//! enabled" is not a portable shortcut: on Windows/macOS it would leave two
+//! sockets on one port. Unlike Windows, macOS's kernel routes an incoming
+//! connection to the more-specific-matching socket deterministically (also
+//! measured 2026-09-09) rather than leaving it ambiguous — so the practical
+//! risk there is a leaked, never-closed listener, not a hijack — but binding
+//! specific addresses (the chosen design) avoids needing to reason about
+//! either case. See `wildcard_vs_loopback_conflict_is_platform_dependent`.
 
 /// The address srv's own startup listeners bind, unconditionally.
 ///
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #3142, opened 2026-09-10
Clare@starpower: fix(cef): move the CEF verification gates out of markdown into tested scripts

AgentO-asaf:
## Why

The carry-set gate and the no-symbol patch check lived as code blocks in `docs/cef-build/CEF_FORK_MAINTENANCE.md`. **Four separate bugs shipped in them**, each making the check incapable of failing — or worse:

| # | Bug | Effect |
|---|---|---|
| 1 | `sed` without `/g` | left `-o` on the real object — the verification **overwrote the build output**. Two objects in the macOS tree had to be rebuilt and the framework relinked. |
| 2 | pristine build compiled from `/tmp` | differed on the embedded DWARF source path alone, so `shipped != unpatched` could never fire and an unpatched object would still report `OK` |
| 3 | `git -C ""` | does not fail — answers for the current directory, silently returning **Chromium's** HEAD instead of the fork's |
| 4 | `set -o pipefail` + `grep -q` | grep exits on first match, `git show` takes SIGPIPE, pipeline reports failure **despite** the match |

None were visible by reading. **Bug 4 was found by writing the test.**

## What's here

```
scripts/cef-verify.sh          carry-set gate, 21 probes, --ref / --repo / --remote
scripts/cef-verify-patches.sh  three-way differential compile for the two patches nm cannot see
scripts/cef-verify.test.sh     12 tests, run by ci-pr.yml on every PR
```

CI can't run the gates themselves (no CEF checkout), so what runs is the **test suite against synthetic fixtures** — which is precisely the run that would have caught all four bugs.

## The tests are checked to actually fail

| Mutation | Result |
|---|---|
| reintroduce bug 4 | `12 passed` → **`10 passed, 2 failed`** |
| reintroduce bug 1 | ABORT guard fires, refuses to run rather than corrupting |

Getting the fixture right took three attempts, and **both wrong ones are recorded in the test file**:

- 600 lines of padding — fits inside the ~64 KiB pipe buffer, no SIGPIPE
- 4000 lines with the identifier **last** — grep reads to EOF, never exits early, so again no SIGPIPE

The bug is about **where the match sits**, not file size. A regression test that cannot fail is the same defect as the code it guards, so I verified in both directions rather than trusting a green run.

## Verified against reality, not just fixtures

```
scripts/cef-verify-patches.sh --build-dir out/Release_GN_arm64
  OK   base/apple/mach_port_rendezvous_mac.cc
  OK   content/browser/renderer_host/render_widget_host_view_base.cc

scripts/cef-verify.sh --ref 7778   ->  21 OK, 0 MISS
scripts/cef-verify.sh --ref 7977   ->  21 OK, 0 MISS
```

**`7977` now passes too** — someone merged both 152 feature branches via `integration-7977` while this was being written, which closes the gap agentmuxai/cef#9 was opened for. I'll close that PR as superseded.

## Doc changes

`CEF_FORK_MAINTENANCE.md` keeps the reasoning, the inventory and the failure history; anything executable now points at the scripts. Zero inline shell functions remain.

### DIFF
--- a/.changesets/1789016663-fix-cef-move-the-cef-carry-set-gates-into-tested-scripts-2uu7.md
+++ b/.changesets/1789016663-fix-cef-move-the-cef-carry-set-gates-into-tested-scripts-2uu7.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): move the CEF carry-set gates into tested scripts
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -135,6 +135,17 @@ jobs:
       - name: Scrollbar cursor gate
         run: bash scripts/check-scrollbar-cursor.sh
 
+      # scripts/cef-verify.sh is the carry-set gate for the CEF fork
+      # (docs/cef-build/CEF_FORK_MAINTENANCE.md §5). It runs against a CEF
+      # checkout, which CI does not have -- so what runs here is its TEST SUITE,
+      # against synthetic fixtures. That is the point: four bugs in the previous
+      # markdown version of these checks made them incapable of failing (or, in
+      # one case, made them overwrite the build output they were verifying), and
+      # none were visible by reading. This is the run that would have caught all
+      # four.
+      - name: cef-verify self-test
+        run: bash scripts/cef-verify.test.sh
+
       - name: Input-handler layout-read gate
         run: bash tools/lint/check-input-handler-layout-reads.sh
 
--- a/docs/cef-build/CEF_FORK_MAINTENANCE.md
+++ b/docs/cef-build/CEF_FORK_MAINTENANCE.md
@@ -268,96 +268,31 @@ Run this against any `agentmuxai/<ms>` before building or releasing from it. It 
 cheap and it is the only thing that catches Layer B gaps.
 
 ```bash
-# Paste into a shell, then run:
-#   cef_verify                      # integration branch 7778 on remote 'agentmuxai'
-#   cef_verify 7977 fork            # milestone 7977, remote named 'fork'
-#   cef_verify a1b2c3d4...          # an exact SHA -- see section 8, P3
-#
-# A function, not a bare script: a top-level `exit 1` would close the shell of
-# anyone who pasted this, which is a poor reward for following the doc.
-cef_verify() {
-  local REF="${1:-7778}"
-  local REMOTE="${2:-${REMOTE:-agentmuxai}}"
-  local BR
-
-  if [[ "$REF" =~ ^[0-9]+$ ]]; then
-    # A bare milestone number means the integration branch on REMOTE.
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

AgentO-asaf:
Correct, and a regression I introduced while porting the check out of markdown. Fixed.

The doc version had `|| return 1` under a comment marked **MANDATORY**; the script version swallowed it with `|| true`. With a stale remote-tracking ref still resolving locally, the gate would report `21 OK` against yesterday's branch — a confident answer about state that has since moved, which is exactly how a wrong claim about the 152 port got into §1.2 in the first place. If the network or auth is down, the correct outcome is *"I cannot tell you"*, not `21 OK`.

You were also right that nothing exercised it: every other fixture passes `--ref <SHA>`, which skips the fetch branch entirely.

### The new test took two attempts, and the first one is worth recording

My first version set the remote URL to a path that no longer contained `agentmuxai/cef` — so the **repo validation rejected the fixture before the fetch ever ran**. The assertions held, the test passed, and it proved nothing about the code under test. It only reproduced once the URL still *looked* like the fork while being unreachable.

That's the same defect class as the four bugs this PR exists to fix, one level up: a check that passes for a reason unrelated to what it claims to verify.

### So it's confirmed in both directions

| Script state | Suite |
|---|---|
| unfixed (`\|\| true`) | **12 passed, 1 failed** |
| fixed (fatal) | **13 passed, 0 failed** |

Re-ran against the real fork afterwards — `agentmuxai/7778` still `21 OK, 0 MISS`, so the fatal fetch doesn't break the normal path.

AgentO-asaf:
Already fixed in `1e07c3d55` — this review is against `b6924e458`, the commit before it. ReAgent raised the same thing independently, which is a decent signal it was the real defect in this PR.

Current state (`scripts/cef-verify.sh:85`):

```bash
git -C "$REPO" fetch "$REMOTE" --prune >/dev/null 2>&1 \
  || { echo "cef-verify: fetch from '$REMOTE' failed; refusing to verify against a possibly stale ref" >&2; exit 1; }
```

Your reasoning about `--prune` is the right frame, and it's why I made it fatal rather than warn-and-continue: the refresh cannot be *assumed* when the command fails, so the honest outcome is "I cannot tell you", not `21 OK` against whatever was cached.

Covered by a new test — and worth noting my first attempt at that test **passed for the wrong reason**: it set the remote URL to something not containing `agentmuxai/cef`, so the repo validation rejected the fixture before the fetch ever ran. Green, and proving nothing. It only reproduced once the URL still looked like the fork while being unreachable.

Confirmed in both directions rather than trusted because it was green:

| Script state | Suite |
|---|---|
| unfixed (`\|\| true`) | 12 passed, **1 failed** |
| fixed (fatal) | **13 passed, 0 failed** |

Real fork re-checked after: `agentmuxai/7778` still `21 OK, 0 MISS`.

PR #3143, opened 2026-09-10
Manoz@Area54: fix(agent-pane): bind the working indicators to one predicate

GenericAgentX-asaf:
## Problem

The "Working…" row and the top progress bar could disagree — and at exactly 30 seconds into every long Bash call they reliably did: the row vanished while the bar kept marching.

Not a race. **Three hand-written copies of one idea that had drifted:**

| Consumer | Expression | Site |
|---|---|---|
| Top bar | `showingLaunchActivity() \|\| workingFromPhase(turnPhase)` | `agent-view.tsx:2296` |
| "Working…" row | the same, **minus** `supersededByDock()`, **plus** `compacting`/`reconnecting` | `agent-view.tsx:1630` |
| Composer strip | the bar's expression, copied again | `agent-view.tsx:2580` |

## The definition this settles on

Stated by the repo owner and now recorded in the report:

> Both indicators mean: **if you type now, your message will be queued until the next turn.** If neither is present, the agent responds immediately. Work you would not otherwise see is the **dock's** job, not theirs.

That makes the design fall out. The indicator predicate and the composer's own send gate (`wasAlreadyWorking`, `agent-view.tsx:1762`) are **one rule expressed twice** — once to decide, once to warn. Any state where they disagree is a bug by construction, not a judgement call about which indicator is more informative.

## Why `workingRowSupersededByDock` is deleted rather than retuned

It rested on a factual error, stated plainly in its own doc comment and in the spec that introduced it:

> Promotion is the moment the work stops being *"the pane is blocked on this"* … `Working…` says the pane is busy waiting, when **backgrounding the task is precisely what made it not busy**.

Nothing had been backgrounded. `TOOL_PROMOTION_MS` promotion is a **display** change that adds a dock row — the call stays in the foreground, the turn stays blocked, and a message typed at that moment still queues. The row was hiding a gate that was still closed, and the user learned the truth by typing and being told to wait.

**AgentMux does not decide backgrounding at all.** The Claude Code harness does, per call, and AgentMux only detects it *afterwards* by string-matching `"Command running in background with ID:"` (`tool-adapter.ts:57`). The two were conflated.

## Spec correction, not a silent workaround

`SPEC_AGENT_WORKING_ROW_ABOVE_COMPOSER_2026_09_01` §1 specifies the stand-down as intended behaviour and is marked Implemented. Routing around it in code would have left the spec standing to regenerate the bug, so it carries an explicit correction notice naming the wrong premise.

Its **§2 (row above composer) is untouched and still current**, as is the `toolPromoted` label suppression — not repeating the dock's tool name *inside* the row was never the problem. Hiding the whole row was.

## Tests

`working-indicator.ts` + 6 tests. Two worth calling out:

- `agrees with workingFromPhase for every phase when not launching` — pins the predicate against the send gate across all six phases, so they can't drift again.
- `stays busy for a Streaming turn regardless of any dock activity` — the predicate deliberately has **no dock input**. If someone adds one, this test is what should stop them.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run frontend/app/view/agent/` — **1774 passed, 0 failed** (132 files), including the 6 new
- Doc gates pass

## Not in this PR

The report's P2 (backgrounding should release the input gate; wiring the dead `ScrubOrphanedInProgress` into `finalizeTurn`) and P3 (ambient Haiku narration of a backgrounded task). Both independently shippable, both follow.

## Also included

`docs/reports/REPORT_AGENT_PANE_PROGRESS_INDICATORS_CONSOLIDATION_2026_09_09.md` — the analysis behind this, plus a consolidation index of ~20 fragmented tracking docs across specs/reports/retros. **No document previously treated the row and the bar as one system**, which is a large part of why this survived so long.


a5af:
@codex review

GenericAgentX-asaf:
Both addressed in 130515d5. reagent's P1 and codex's second comment are the same finding, and it was a fair hit — my own report named this as Divergence B (§3.2) and prescribed exactly this fix in §8.1, and I shipped the first revision without it while claiming the three consumers "cannot disagree". That claim was wrong.

**Divergence B.** `compacting` and `reconnecting` can be non-null while `turnPhase` is Idle. The row read them; the bar and composer strip did not. Both are now terms in `paneBusyForInput`.

Worth noting the direction: here the **row was right and the bar wrong** — the reverse of the dock-promotion case. `reconnecting` is set *only after the process has already crashed or exited*, so there is nothing alive to answer a message typed then, and `compacting` means the CLI is mid-compaction. Both are genuinely "not now", so the fix is to light the bar, not to darken the row.

Checked the label doesn't regress: `loadingLeftText` tests `reconnecting`/`compacting` before anything else, so "Reconnecting…"/"Compacting…" still win over "Working…", and `AgentFooter` already rendered for both.

**codex's first comment (P2) — the contract was overstated.** Correct, and it's the more interesting one. The module promised `true ⇔ queued`, which is false for two of its own inputs: during launch/relogin the message is rejected by the auth guard, and during a reconnect there's no process to take it. Only the turn-in-flight case actually queues. Rather than narrow the predicate I narrowed the claim — it now promises "will not be answered immediately", which is the common factor and the only thing the indicator can honestly assert. The three outcomes are enumerated in the doc so the next reader doesn't have to rediscover them.

Tests: 9 now (up from 6), including one pinning that a falsy-but-present state object (`0`, `""`) still reads busy — these are opaque payloads and only `null` means idle.

tsc clean; 2029 tests pass across `view/agent/` + `store/agent-pane-state/`.

GenericAgentX-asaf:

Fresh head 130515d5 with the Divergence B fix and the contract-wording correction; the standing review state is still CHANGES_REQUESTED from 1aa61fc8.

PR #3139, opened 2026-09-10
refactor: armory naming Phase 2 — retire 'brain' on the bundle side

AgentA-asaf:
Phase 2 of `docs/specs/SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md`: retire "brain" as a name for bundles. Follow-up to #3133 / #3135 (Phase 1). Tracking: #2024.

**Pure renames, no behaviour change**, one commit per side:

| Commit | Rename |
|---|---|
| `1ff84dc04` | Rust: `format_global_brain_block` → `format_global_bundle_block` (2 live callers + tests), `CommandReorderGlobalBrainData` → `CommandReorderGlobalBundlesData`, "global brain" comments |
| `999f7240e` | Frontend: `view/brain/` → `view/global-bundle/`, `GlobalBrainViewModel`/`GlobalBrainManager` → `GlobalBundle*`, `formatGlobalBrainBlock`, `ReorderGlobalBrainCommand` → `ReorderGlobalBundlesCommand`, the `.global-brain-*` SCSS selectors (47, used only inside the directory), test id, comments |
| `HEAD` | spec §9 rewritten from "decisions needed" to the decisions taken (Bundles tab; Phase 3 alias question deferred to Phase 3 with a default; `viewType` wire value frozen; Phase 2 now, re-evaluate before 3/4); changeset |

**Deliberately untouched**
- Wire command `reorderglobalbrain` and its `COMMAND_REORDER_GLOBAL_BRAIN` const (Phase 3, wire-breaking). `ReorderGlobalBundlesCommand` keeps that wire string.
- The MCP tool gloss "native memory (brain)": once bundles stop using the word it is unambiguous on its own (spec §5 Phase 2).
- `BrainSpinner` and the startup splash "brain" (a UI element), the `fa-brain` icon on the Memory tab and the native-memory modal (that brain *is* the genuine-memory one), and every "split-brain" comment.
- Folding `GlobalBundleManager` into the bundle view is Phase 4, not this PR.

**Verification**
```
cargo test -p agentmux-srv   → 3211 passed, 0 failed
npx tsc --noEmit             → exit 0
npx vitest run               → 260 files, 3892 passed, 0 failed
```
Per-commit rule from spec §4: every quoted line in the diff was reviewed; no string literal changed except the armory-view test id.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3138, opened 2026-09-10
ui(armory): rename the ABF tab to Bundles

AgentA-asaf:
Implements the §9.1 decision of `docs/specs/SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md`: the Armory tab that lists bundles is called **Bundles**. **ABF** stays the name of the Armory Bundle Format and remains on import/export affordances, the structural validator copy, and the bundle form's field help (that copy is reworked together with the form in Phase 4).

User-visible changes:
- Armory rail label `ABF` → `Bundles` (`ARMORY_SECTION_LABELS.bundles`); persisted section key `bundles` unchanged.
- Bundle pane title / back label `ABF` → `Bundles`.
- Startup modal and new-bundle modal pointers `Armory → ABF` → `Armory → Bundles`.
- Bundle-summary noun labels: "This agent's own bundle", "has no bundle of its own yet".

Tests updated to the new labels (`armory-view.test.tsx` rail/highlight assertions, `AgentStartupModal.test.tsx`, `bundle-summary.test.tsx`). Verified: `tsc --noEmit` exit 0; targeted vitest 52 files / 662 tests green. Tracking: #2024.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3135, opened 2026-09-10
refactor(frontend): armory naming Phase 1 — Memory→Bundle frontend renames

AgentA-asaf:
Frontend half of Phase 1 of `docs/specs/SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md`. Follow-up to #3133 (Rust half). Tracking: #2024.

**Pure renames, no behaviour change.** Each commit is a mechanical, independently revertible step:

| Commit | Rename |
|---|---|
| `520fdb4a7` | `frontend/app/view/memory/` → `view/bundle/`; `MemoryViewModel`→`BundleViewModel`, `MemoryManager`→`BundleManager`, `MemoryView`, `MemoryDraft`, `draftFromMemory`; `.memory-view-*` SCSS selectors (used only inside the directory) |
| `e27ba37ba` | ambient `type Memory` → `type Bundle` (`gotypes.d.ts`), type-position uses only |
| `1929b1a8d` | split `rpc-api/memory.ts` → `BundleApi` (into existing `bundle.ts`) + `NativeMemoryApi` (`native-memory.ts`); `ListMemoriesCommand`→`ListBundlesCommand`, `Get/Upsert/Delete[System]MemoryCommand`→`…BundleCommand` |
| `94d4d151b` | launch-flow state + modals: `memoryId`→`bundleId`, `MemoryChanged`/`MemoriesLoaded|Loading|Failed`→`Bundle*`, `realMemories`, `continueLocksMemory`, `onRequestNewMemory`, `AgentNewMemoryModalPanel` (+ file), `NewMemoryFormData`, `memoriesAtom` |
| `7b14153ac` | spec §4: three new do-not-rename rows (persisted markers, wire strings, user-visible text) recording the #3133 Codex P1 lesson; changeset |
| `ea45baae3` | Rust: `CommandDeleteMemoryData`/`DeleteMemoryResult` → `*Bundle*` (they carry `#[ts(export)]`, so their names are the generated filenames); bindings regenerated, orphaned old `.ts` files `git rm`'d, typed-registry assertions updated |

**Deliberately untouched** (spec §4):
- Wire RPC names `listmemories`/`getmemory`/`upsertmemory`/`deletememory`/`upsertsystemmemory`/`deletesystemmemory` (Phase 3 retires them), the `memories:changed` WPS event, the `memory_id` wire field.
- Persisted keys: `viewType: "memory"`, `armory:section` value `"memory"`, `armory:memory:subsection`.
[...]

PR #3133, opened 2026-09-09
refactor(srv): armory naming Phase 1 — Memory→Bundle internal renames

AgentA-asaf:
Phase 1 (Rust half) of `docs/specs/SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md` (landed in #3131). Tracking: #2024.

**Pure internal renames, no behaviour change.** Every commit is a mechanical, independently revertible step:

| Commit | Rename |
|---|---|
| `9f8e197de` | storage module `memory_bundles.rs` → `bundles.rs` |
| `dd3f1a6a6` | `Store::bundle_memory_*` → `Store::bundle_*` (8 methods: list, list_global, get, upsert, upsert_system, delete, delete_system, reorder) |
| `407ad5d45` | `pub struct Memory` → `pub struct Bundle` (the `db_bundles` row type) |
| `0c339e6fe` | `agent_handlers/memory.rs` → `agent_handlers/bundle.rs` (file was bundle-only; rename not split, per Codex finding on #3131) |
| `bc6d00a8f` | split `rpc_types/memory.rs` → `bundle.rs` (bundle command shapes) + `native_memory.rs` (the 19 `NativeMemory*` payloads). Glob re-exports preserved, so no importer changes. |
| `8bf71ff8e` | regenerated binding doc comment, CLAUDE.md line that said `bundle_memory_*` stays, changeset |

**Deliberately untouched** (spec §4 do-not-rename list):
- SQL literals in `migrations.rs` (`"db_memory_bundles"`, `idx_memory_bundles_is_blank`, `idx_ss_memory_bundles_is_blank`) — `LEGACY_TABLE_RENAMES` ordering is load-bearing. Verified intact.
- `db_agents.memory_id` / `default_memory_id` columns (deferred; no `RENAME COLUMN` precedent).
- MCP `Memory*` tools, `agent:memory:*` RPCs, `db_agent_native_memory*`, `NativeMemory*`, ABF `components.*` keys.
- `CommandDeleteMemoryData` / `DeleteMemoryResult` (ts_rs-generated bindings; renamed in the coordinated frontend PR so old generated files can be `git rm`'d together).
- `app_api/mod.rs` `mod memory;` — that one really is the native-memory App API.

**Verification**
```
cargo test -p agentmux-srv   → 3210 passed, 0 failed, 6 ignored
frontend/types/rpc diff      → one doc-comment line only (committed)
```

[...]

PR #3144, opened 2026-09-10
docs(specs): instruction and memory portability spec

AgentA-asaf:
The follow-on the naming spec's §8 deferred. Docs-only, no changeset.

**`docs/specs/SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md`** (Status: proposed) — the design for "tight control over anything the agent uses as instructions and memory, including portability". Verified against `a86cdee50`; every code citation in it was checked by hand.

What it establishes:

- **§2 baseline.** The three export paths and what each actually carries; the five `components.*` keys and where each is written; memory's on-disk-vs-mirror split; and what `write_claude_md_respecting_ownership` does with a foreign `CLAUDE.md`.
- **§3 gaps**, five of them, each with a citation:
  1. Agent-less export drops memory **silently** while import warns via `MEMORY_COMPONENT_IGNORED_WARNING`. The asymmetry is the announcement, not the omission — omitting memory from an agent-less bundle is correct.
  2. Project instructions are invisible and unportable. Nothing reads a foreign `CLAUDE.md`/`AGENTS.md` except the ownership branch test, and the only persisted state about one is a single boolean.
  3. Memory's location is derived from the working directory, so the durable mirror is a compensating control for a location AgentMux does not own.
  4. **Bundle components have two sources of truth.** `export_bundle` reads the inline JSON columns; the `db_bundle_skills_ref` / `db_bundle_mcp_ref` tables are read only by `bundle_skill_list` / `bundle_mcp_list`, whose callers are the App API. A skill or MCP server bound through the ref table and absent from the inline column **does not appear in the ABF**. Whether a sync path exists is unverified — Phase 0 is to find out.
  5. History files are written into the archive but never registered in the manifest.
- **§5 design.** Export-side symmetry warning; `components.projectInstructions` as **read-only tracking** (read, hash, export, surface on import — never written back, so the ownership protection stays intact); the memory-location invariant stated as an invariant with the migration deliberately left to its own spec; one source of truth for components before the format grows.
- **§7 phases**, ordered so the two cheap ones (symmetry, manifest history) can land together and the design-heavy one does not ride along.

**Three stale status lines fixed** while verifying the baseline, each with evidence:

| Doc | Was | Now |
|---|---|---|
| `SPEC_ARMORY_NAMING_CONSOLIDATION` | "proposed — needs the §9 decisions" | `active`, naming what landed (#3133, #3135, #3138, #3139) and what remains |
| `SPEC_ABF_V0_2` | "proposal — no implementation yet" | `implemented` — all four of its sections are in the code, citation given for each |
| `CLAUDE.md` | Armory tab branded "ABF" | "Bundles" since #3138; ABF now names only the file format |

`docs/specs/INDEX.md` regenerated for the new row and the ABF v0.2 status move.

**Verification**
```
task check:doc-status              → ok (3 docs checked)
scripts/check-spec-citations.sh    → ok (4 files checked)
scripts/gen-docs-index.sh          → committed, tree clean
```

Tracking: #2024.


a5af:
@codex review

AgentA-asaf:
All three Codex findings confirmed and fixed in `321a1e82b`.

**P1 — plural instruction sources.** Correct and it was the more serious of the two design holes: §5.2 resolved `startup_instructions_filename`, which is the file AgentMux *writes*, not the set the provider *reads*. The registry already documents the gap for Copilot — `.github/copilot-instructions.md`, `CLAUDE.md` and `GEMINI.md` alongside the chosen `AGENTS.md` (`providers.rs:501-507`). §5.2 now calls for a second plural native-sources field on the registry and resolves the union, and names populating it a Phase 3 prerequisite.

**P1 — wrong ownership marker.** Confirmed: `STARTUP_INSTRUCTIONS_MANAGED_MARKER` (`agent_config.rs:1040`, tested at `:1089`) is the marker for every non-`CLAUDE.md` startup file. Testing only `CLAUDE_MD_MANAGED_MARKER` would have reported every generated `AGENTS.md`/`GEMINI.md`/`QWEN.md`/`.pi/APPEND_SYSTEM.md` as `foreign` — inverting the single field that section exists to get right. Now specified as a shared two-marker helper reused by the resolver and both writers.

**P2 — ABF v0.2 status.** Also correct: §2.4 requires a `compatibility.agentmux` minimum and the manifest has no `compatibility` field at all. Status is now `active` with the shipped list and that one remaining item spelled out, rather than `implemented`. Index regenerated for the section move, and §5.5 of the new spec records it as a third piece of the same format-version debt.

Worth noting the original line said "proposal — no implementation yet" while four sections were shipped; the correction was right to make, just not all the way to "implemented".

Issue #3127, opened 2026-09-09
CEF fork: Views transparency not gated on is_frameless_ in window_impl.cc / browser_view_impl.cc (affects shipped 148 build)


AgentO-asaf:
Raised by ReAgent reviewing agentmuxai/cef#9. **Not introduced by that PR** — both files are byte-identical between `7778` and the 7977 branches, so this is pre-existing in the branch the shipped macOS framework was built from. Filed here because issues are disabled on `agentmuxai/cef`.

## The finding

**`libcef/browser/views/window_impl.cc` (`CefWindowImpl::CreateWidget`)** sets the compositor transparent whenever the *global* `CefSettings.background_color` is transparent, with no `is_frameless_` check:

```cpp
if (CefContext::Get() &&
    CefContext::Get()->GetBackgroundColor(nullptr, STATE_ENABLED) ==
        SK_ColorTRANSPARENT) {
  SetBackgroundColor(SK_ColorTRANSPARENT);
}
```

This contradicts the gating `CefWindowView::CreateWidget` performs immediately before (via `root_view()->CreateWidget`) — `window_view.cc:538`:

```cpp
bool is_translucent = is_frameless_ && color == SK_ColorTRANSPARENT;
```

`params.opacity` stays `kOpaque` for framed windows, so the compositor skips its screen-fill quad on an opaque native surface.

**`libcef/browser/views/browser_view_impl.cc` (`SetDefaults`)** resolves with `STATE_ENABLED` for every Views-hosted browser rather than frameless ones only, arming the renderer-side cascade from the global setting alone.

## Reachable here, not just latent

I checked our own consumer code rather than assuming:

- `agentmux-cef/src/lib.rs:1014` sets the **global** `CefSettings.background_color = 0x00000000` (transparent).
- `agentmux-cef/src/app/mod.rs:214-222` explicitly handles **non-frameless** windows: *"a non-frameless popup (DevTools) is meant to have a native title bar... The main window is frameless (custom HTML title bar) and is deliberately left untouched."*

So we do create framed Views windows (DevTools popups) while the global background colour is transparent — exactly the combination the ungated branch fires on. The main window is frameless, so the primary UI is unaffected; expected symptom is confined to DevTools popups on the Views path.

## Scope

- Present on **`7778` (shipped)** and on the 7977 feature branches.
- Not upstream CEF — introduced by our transparency work.
[...]

PR #3145, opened 2026-09-10
Clare@starpower: fix(cef): bump the macOS CEF pin to 148.23.25 — ships the macOS 26 renderer fix

AgentO-asaf:
```diff
-MACOS_TAG="cef-macos-arm64-148.23.23-codecs"
+MACOS_TAG="cef-macos-arm64-148.23.25-codecs"
```

Release side of #3108 **workstream A**. **Milestone stays 148** — this is *not* part of the 148→152 upgrade, and Windows/Linux need no rebuild, because the patch touches `base/apple/`, never compiled elsewhere.

## What 148.23.25 ships that 148.23.23 didn't

**`agentmux_process_requirement`** — the macOS 26 renderer fix (`-67030` / `errSecCSReqFailed`). Written in June and absent from **every shipped binary** until now: it was never registered in `patch.cfg`, and `patcher.py` reads that file to decide what to apply, so no build could ever have applied it.

**The renderer-side transparency work**, which `agentmuxai/7778` was itself missing until agentmuxai/cef#8. Building from `7778` before that merge would have fixed the crash-loop *and* regressed macOS window transparency at the same time, silently.

## Milestone gate

Re-ran the `cef-runtime-pins` check by hand:

```
windows m148   linux m148   macos m148   -> cross-check PASSES
```

## The release

[`cef-macos-arm64-148.23.25-codecs`](https://github.com/agentmuxai/cef/releases/tag/cef-macos-arm64-148.23.25-codecs), cut with `--target` at the **exact build commit** `1bee8b7da` (an ancestor of `7778`) rather than the branch tip, so the tag records what was actually built.

Confirmed *after* cutting that `targetCommitish` resolves to that SHA and not the default branch — that's the failure mode behind the 2026-07-28 tag-ordering incident documented in the macOS runbook.

## Verification — of the artifact, not the build tree

| Check | Result |
|---|---|
| carry-set at the build commit | `21 OK, 0 MISS` |
| `verify-cef-framework-darwin.sh` on the **extracted** tarball | exit 0 |
| `BeginWindowDrag` / `IsUniqueForCEF` / `IsAlloyContents` / `HasExternalParent` | 2 / 2 / 2 / 4 under full `nm` |
| the two patches that add **no symbol** | confirmed by differential compile |
| `Versions/Current` symlink chain | survives the tar round-trip |

## Not covered here

- **H.264 playback** — needs a packaged app, not a framework on disk. I'll fold it into a `task package:macos` run rather than claim it verified.
- **#3127** — Views transparency not gated on `is_frameless_`. Pre-existing and present in the previous release too; this build passes the carry-set gate at `21 OK` while carrying it, which is the concrete example of why a green gate isn't a release sign-off.

### DIFF
--- a/.changesets/1789018629-fix-cef-ship-the-macos-26-renderer-fix-bump-the-macos-cef-pi-b4pu.md
+++ b/.changesets/1789018629-fix-cef-ship-the-macos-26-renderer-fix-bump-the-macos-cef-pi-b4pu.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): ship the macOS 26 renderer fix — bump the macOS CEF pin to 148.23.25
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -163,7 +163,7 @@ jobs:
         run: |
           WIN_TAG="cef-windows-x86_64-148.0.7778.180"
           LINUX_TAG="cef-linux-x86_64-148.0.7778.180-codecs"
-          MACOS_TAG="cef-macos-arm64-148.23.23-codecs"
+          MACOS_TAG="cef-macos-arm64-148.23.25-codecs"
 
           # Chromium milestone = the leading dot-separated numeric segment
           # after the cef-<platform>-<arch>- prefix. Works across both
--- a/docs/cef-build/CEF_FORK_MAINTENANCE.md
+++ b/docs/cef-build/CEF_FORK_MAINTENANCE.md
@@ -517,17 +517,18 @@ All three pins live in one place — `release.yml`'s `cef-runtime-pins` job — 
 is correct and was a deliberate fix (PR #3086 and its follow-up). Keep it that
 way; do not reintroduce per-job literals.
 
-Current pins:
+**The live values are in `release.yml`, and are deliberately not repeated here.**
+This section previously pinned them inline and went stale the first time one was
+bumped — the same duplicated-value rot §5 and §7 keep running into. To read the
+current pins:
 
-```
-WIN_TAG="cef-windows-x86_64-148.0.7778.180"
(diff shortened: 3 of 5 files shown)


a5af:
@codex review

AgentO-asaf:
Correct. Fixed in `fa0dc9d05` — and rather than bump the number and leave it to rot again, I removed the duplication.

`release.yml` is the single source of truth; §8 now says so and gives a command to read the live values instead of restating them:

```bash
sed -n '/cef-runtime-pins/,/MISMATCH/p' .github/workflows/release.yml | grep '_TAG='
```

Verified that command actually prints the three tags. This is the same duplicated-value rot §5 and §7 of that document keep hitting, so it's worth deleting rather than repairing.

**Your finding pointed at one stale claim; checking for siblings found three more:**

- `SPEC_CEF_MILESTONE_UPGRADE_148_TO_152`'s "Our pins" list
- `build-patched-framework-macos.md`'s "what the verified artifact looks like" table — version `148.23.23.0` → `148.23.25.0`, released tag updated, and I swapped the stale `cef_version.h` row for the **build commit**, which is what the release's `--target` records and is the more useful provenance
- the `CEF_VERSION="148.23.23"` example in that same runbook, now an explicit *read-it-from-the-build* instruction with the PlistBuddy command — a hard-coded example version is exactly what went stale here

While in that table I also checked a claim nobody asked about: it said **547 MB unstripped**. Measured, this build's binary is **428 MB** (181 MB as the released tarball). Corrected.

**Deliberately not touched:** the many historical `148.23.23` references in retros, status docs, and the `release.yml` comment that says *"the tags PR #3086 pinned"*. Those are accurate records of what happened; rewriting them would be churn.

PR #3146, opened 2026-09-10
fix(srv): integration tests no longer trigger a real macOS Keychain prompt

AgentO-asaf:
Fixes the exact symptom hit this session: running `cargo test -p agentmux-srv` on a real Mac produced 4 near-simultaneous interactive Keychain prompts, with no indication of which process or why, requiring the person at the keyboard to deny each one blind.

## Root cause

`agentmux-srv/tests/integration_test.rs` spawns the real compiled `agentmux-srv` binary (`spawn_backend`/`spawn_backend_with_data_dir`) to test it end-to-end. That binary's normal startup unconditionally calls `CloudSubscriber::init_global`, which registers the muxbus credential with the broker scheduler and triggers `Store::muxbus_is_fresh()` — a real, synchronous OS-keychain read — almost immediately.

This repo already has a retro on the general phenomenon (`docs/retro/retro-macos-muxbus-keychain-prompt-storm-2026-08-19.md`), but it addressed the *shipped app's* prompt count (collapsing 12 keychain entries down, adding read timeouts). It didn't cover an *automated test run* reaching the real keychain at all.

The keychain entry's ACL only trusts specific notarized, Developer-ID-signed installs. An ad-hoc/dev-signed `target/debug` binary is never on that list, so macOS surfaces a fresh interactive consent prompt for **every spawned test process** — on whatever account happens to be running the suite, which on a shared dev machine may not even be the person who triggered the test run.

## Fix

`AGENTMUX_DISABLE_CLOUD_SUBSCRIBER` — presence-based, matching the `AGENTMUX_DEV`/`AGENTMUX_TRAY` idiom already used in this codebase — gates the `init_global` call in `bootstrap.rs`. **Off by default**: the shipped app and `task dev` keep the real muxbus reconnect-on-launch behavior exactly as before. Both spawn sites in `integration_test.rs` now set it.

## Verification

Two levels:

1. **Unit tests** on the gate function: defaults to enabled with no var set, disabled regardless of value once the var is present (empty string included — matching the presence-based idiom, not a value check).
2. **The actual regression, empirically** — not just testing the gate logic in isolation:
   - Captured wall-clock start time.
   - Ran `cargo test --test integration_test -p agentmux-srv` — **5 passed**, all the subprocess-spawning tests (`auth_rejects_missing_key`, `auth_accepts_valid_header`, `health_returns_200`, `restore_last_session_survives_a_real_process_restart`, `sigterm_exits_process`) still exercise real IPC/lifecycle behavior against the spawned binary.
   - Searched the macOS unified log (`log show --predicate 'eventMessage contains "keychain"'`) for any prompt event from `target/debug/agentmux-srv` since that timestamp.
   - **Zero.** The same class of run previously produced 4.

`cargo test -p agentmux-srv -- --test-threads=1`: **3197 passed** (2 more than before — the new gate tests), 0 failed.

### DIFF
--- a/.changesets/1789018616-fix-srv-integration-tests-no-longer-trigger-a-real-macos-key-bm6w.md
+++ b/.changesets/1789018616-fix-srv-integration-tests-no-longer-trigger-a-real-macos-key-bm6w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): integration tests no longer trigger a real macOS Keychain prompt
--- a/agentmux-srv/src/bootstrap.rs
+++ b/agentmux-srv/src/bootstrap.rs
@@ -1076,9 +1076,30 @@ pub fn spawn_background_subsystems(
     ));
 
     // Cloud push subscriber — single WS connection per sidecar that the cloud
-    // uses to push reactive injections instead of polling.
-    // No-op until the user connects via muxbus.login.
-    crate::muxbus::cloud_subscriber::CloudSubscriber::init_global(id_store.clone());
+    // uses to push reactive injections instead of polling. The WS connection
+    // itself is a no-op until the user connects via muxbus.login, but
+    // `init_global` registers the muxbus credential with the broker
+    // scheduler unconditionally, which calls `Store::muxbus_is_fresh()` —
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #3147, opened 2026-09-10
feat(srv): portability Phase 1+2 — export announces uncarried memory; history in the manifest

AgentA-asaf:
Phases 1 and 2 of `docs/specs/SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md` (#3144). The spec's §7 says these two are small enough to land together and that the design-heavy Phase 3 should not ride along; this is that pair.

**Phase 1 — `bundle.export` announces the memory it cannot carry (§3.1/§5.1).**

Import already announced its mirror case: an agent-less import that sees `components.memory` pushes `MEMORY_COMPONENT_IGNORED_WARNING` (`bundle_import.rs:432`). Export said nothing. Since every `components.*` key is optional, a missing `memory` key is indistinguishable from "this bundle had none", so an operator exporting a bound agent's bundle got a file that looked complete.

**What it emits is unchanged.** Omitting memory from an agent-less export is correct — a bundle detached from any agent has no memory to carry. Only the silence was the bug.

Two details worth review attention:

- **Both bindings count.** `AgentDefinition.memory_id` is the agent's own bundle (stored in `db_agents.default_memory_id`); `AgentInstance.memory_id` is one *launch* pointed at a different bundle (`storage/agents.rs:153-163`). Checking only the definition would miss exporting the bundle a running instance was launched with — likely the more common case. Memory is keyed on the definition id either way, so the instance path resolves through `definition_id`.
- **Reads the mirror, does not refresh it.** `build_export_for_agent` must refresh before copying files (`app_api/bundle.rs:558`); a warning has no business writing to the store as a side effect. The cost is under-warning for memory that exists on disk but was never mirrored — the same residual gap `SPEC_NATIVE_MEMORY_DURABLE_SYNC_2026_08_07.md` already accepts. Documented at the helper.

**Phase 2 — history is registered in the manifest (§3.5).**

`export_for_agent_with_history` wrote transcripts into the archive and counted them in the RPC response, but never touched the manifest after the memory splice — so anything reading `components.*` to learn what the archive holds could not see history at all. Now registered under `components.history`, and **absent rather than empty** when no session was included, since absence already means "none".

**Refactors, both in service of the above**

- `bundle.export`'s handler body extracted to `bundle_export_impl`, so the warning contract is asserted at the RPC boundary rather than only on its helper. This is the shape `bundle_export_for_agent_impl` and `bundle_export_for_agent_with_history_impl` already had.
- The find-parse-rewrite manifest dance is now one `set_manifest_component` helper shared by the memory and history splices, which had it byte for byte.

**Tests** (4 new, all failed before the change):

| Test | Pins |
|---|---|
| `agent_less_export_warns_when_a_bound_agent_has_memory` | helper **and** the RPC's `warnings` array; also that the archive still carries no `memory/` files |
| `agent_less_export_is_silent_without_memory_or_without_a_binding` | bound-but-empty, unbound bundle, and the blank-singleton sentinel |
| `agent_less_export_warns_for_an_instance_scoped_binding_too` | the `AgentInstance.memory_id` path |
| `omits_the_history_component_when_no_session_was_included` | absence, not `[]` |

plus a manifest assertion added to the existing `includes_the_agents_sessions_alongside_the_normal_bundle_files`.

One thing found while writing them, worth knowing independently of this PR: `agent_def_update` does not list `default_memory_id` among its columns (`storage/agents.rs:1188-1197`), so mutating `AgentDefinition.memory_id` and calling it silently no-ops. The test helper uses `agent_def_set_memory_id_if_empty` and says why.

**Verification**
```
cargo test -p agentmux-srv        → 3215 passed, 0 failed (4 new)
frontend/types/rpc                → clean, no binding change
```

Tracking: #2024.

### DIFF
--- a/.changesets/1789018698-feat-srv-armory-portability-phase-1-2-bundle-export-announce-m9er.md
+++ b/.changesets/1789018698-feat-srv-armory-portability-phase-1-2-bundle-export-announce-m9er.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): armory portability Phase 1+2 — bundle.export announces memory it cannot carry; history registered in the ABF manifest
--- a/agentmux-srv/src/server/app_api/bundle.rs
+++ b/agentmux-srv/src/server/app_api/bundle.rs
@@ -436,18 +436,6 @@ fn bundle_export_impl(
     Ok(result)
 }
 
-/// Splice native-memory files into an already-built bundle export's
-/// `armory.json` manifest and files list, adding `components.memory` (ABF
-/// v0.2 §2.3). Kept OUTSIDE `bundle_export.rs` deliberately — that
-/// module's `export_bundle()` is scoped to a bundle's own components
-/// (instructions/skills/MCP/accounts) with no concept of "agent" or
-/// native memory at all; memory is agent-scoped, not bundle-scoped, so
-/// splicing it in here (the RPC-handler layer, which already resolves
-/// other agent-scoped data like skill rows) keeps that module's
-/// documented scope intact rather than growing it a fifth, unrelated
-/// component category. A no-op when `memory_files` is empty — matches the
-/// existing omit-empty-components convention used elsewhere in the
-/// manifest.
 /// Warning the agent-less `bundle.export` pushes when the bundle it is
 /// exporting is bound to an agent that actually has native memory.
 ///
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentA-asaf:
Tracking issue for the full phase list: #3148. This PR is Phases 1 and 2.

AgentA-asaf:
Confirmed, and the analysis is exactly right — but it cannot be fixed the way the comment asks, so I documented it instead. Reasoning, in `209b8246f`:

There is **no globally durable source for the binding today.** `DefinitionRecordV1` does not carry `memory_id` at all (`storage/def_registry_mirror.rs:128-133`), which is why `agent_def_list`'s overlay preserves it only when a local row exists (`storage/agents.rs:462`) — a genuinely cross-channel agent reads back with `memory_id: ""`. The bundle row and the native-memory mirror are both global; the binding between them is the one channel-local piece. So "resolve the binding from globally durable data" has nothing to read.

I considered widening the warning to fire when the binding is merely *unknown*, and rejected it: it would fire on unrelated bundles whenever any cross-channel agent has memory, which trains operators to ignore the warning — worse than the gap for an advisory message that changes no bytes in the archive.

The actual fix is adding `memory_id` to the registry wire format. That is already a tracked gap for larger reasons than this warning — the same missing field makes a cross-channel reopen start unbound and confines m0021's backfill to local SQLite, both noted in `def_registry_mirror.rs`. Added to #3148 with your scenario spelled out, and the limitation is now documented at `bound_agent_has_native_memory` with the citation rather than left implicit.

Worth being explicit that this bounds a *warning*, not behaviour: what any export path emits is unchanged by this PR.

AgentA-asaf:
P2 fixed in `bf3fb4018` — comment-only move. My insertion split `splice_memory_component` from its doc comment; the doc is back on the function and the const keeps its own. Nothing else changed.

Issue #3148, opened 2026-09-10
Tracking: instruction and memory portability (ABF export/import, project instructions, memory location)


AgentA-asaf:
Tracking issue for `docs/specs/SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md` (merged in #3144).

**Goal:** tight control over everything an agent consumes as instructions or memory, including portability. For every byte reaching an agent, AgentMux should be able to answer where it came from, what is in it now, and whether it can move to another machine.

### Phases

- [x] **Phase 0 — DONE 2026-09-09. Answer: they are never synced, by design, and the divergence runs both ways.** No code path writes both stores. Bind reaches `Store::managed_bind_bundle` (`storage/managed.rs:285`) and issues one INSERT into the ref table; the handler never touches the bundle row. `bundle.upsert` and the ABF-import paths write the inline columns and never call `bundle_skill_bind`/`bundle_mcp_bind`. The v23 schema comment states the split (`migrations.rs:198-199`).

  **ABF export reads inline** (`app_api/bundle.rs:388`, `bundle_export.rs:526`); **agent launch reads the ref tables** (`app_api/agent_open.rs:793`, `:812`). Two concrete losses: (1) a skill/server bound in the Armory is live at launch but ships as an empty `skills/`+`mcp/` in the `.abf`, with no warning; (2) `bundle.import` writes inline `mcp_servers` but creates no server rows and no ref rows, so imported servers are inert at spawn. Also, deleting a bundle orphans its ref rows (no FK, deliberate — `migrations.rs:727-741`).

- [~] **Phase 0b — reconciliation landed, column retirement BLOCKED.** Prerequisite of Phase 3: extending the format on a known-lossy base would ship the same bug twice. `SPEC_BUNDLE_AS_CONTAINER_V2_2026_08_17.md:91-95` deferred this decision and named the trigger ("once the new ref tables have shipped and proven out") — which has happened.
- [x] **Phase 1 — export symmetry (§5.1).** `bundle.export` announces memory it cannot carry. — #3147
- [x] **Phase 2 — manifest history (§3.5).** Transcripts registered under `components.history`. — #3147
[...]

PR #3149, opened 2026-09-10
docs(spec): Phase 0 answered — bundle components have two unsynced stores

AgentA-asaf:
Records the answer to Phase 0 of `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md` (#3144). Docs-only, no changeset. Tracking: #3148.

The spec's §3.4 said *"whether a sync path exists elsewhere is unverified; §7 Phase 0 is to answer that."* It has been answered, and the answer is worse than the question assumed.

**There is no sync path, by design, and the divergence runs in both directions.**

No code path writes both stores:
- Binding a skill or server to a bundle (`skill.catalog.bind_to_bundle`, `mcp.catalog.bind_to_bundle`) reaches `Store::managed_bind_bundle` (`storage/managed.rs:285`) and issues exactly one INSERT into the ref table (`:298-305`). The handler (`app_api/skill.rs:513`) never reads or writes the bundle row.
- `bundle.upsert` and the three ABF-import paths write the inline columns (`storage/bundles.rs:245-246`) and never call `bundle_skill_bind` / `bundle_mcp_bind`.
- The v23 schema comment states the split outright (`migrations.rs:198-199`): inline columns are "still the .abf export/import" path, ref tables are the live-resolution one.

So the two consumers disagree, each authoritative for a different thing:

| | ABF export reads | Agent launch reads |
|---|---|---|
| Skills | `bundle.skills` inline (`app_api/bundle.rs:388`) | `effective_skills` → ref table (`app_api/agent_open.rs:793`) |
| MCP servers | `bundle.mcp_servers` inline (`bundle_export.rs:526`) | `effective_mcp_servers` → ref table (`app_api/agent_open.rs:812`) |

**Two concrete losses:**

1. **Bind, then export.** A skill or server added to a bundle in the Armory is live at launch and visible in the UI, but the inline column is still `"[]"` — so the `.abf` ships with empty `skills/` and `mcp/` directories and no warning. ABF is advertised as a backup format.
2. **Import, then launch.** `bundle.import` writes inline `mcp_servers` but creates no `db_mcp_servers` rows and no ref rows, so imported servers are never materialised at spawn. This is exactly the "inert at runtime" state `SPEC_BUNDLE_AS_CONTAINER_V2_2026_08_17.md:40-42` describes and the ref tables were meant to replace.

Secondary: deleting a bundle orphans its ref rows — no FK, deliberately (`migrations.rs:727-741`).

That spec kept the inline columns "as-is for this delivery" as an explicit non-goal (`:91-95`), to be revisited "once the new ref tables have shipped and proven out". They have. This PR therefore adds **Phase 0b — reconcile the two stores** as a prerequisite of Phase 3: adding `components.projectInstructions` on top of a base that already drops ref-table-bound skills would ship the same bug twice.

Spec status moves `proposed` → `active` (Phases 1-2 landed in #3147). Index regenerated.

No code changes. Every claim carries a `file:line` and was verified against `b8b37f13d`.

**Verification**
```
scripts/check-spec-citations.sh  → ok (2 files checked)
task check:doc-status            → ok (2 docs checked)
```

### DIFF
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -342,6 +342,7 @@ partial list.
 | [`SPEC_EDITOR_MCP_OPEN_BLANK_PREVIEW_AND_PANE_REUSE_2026_08_03`](SPEC_EDITOR_MCP_OPEN_BLANK_PREVIEW_AND_PANE_REUSE_2026_08_03.md) | Plan: MCP-opened markdown blank-preview investigation + Editor-pane reuse |
 | [`SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09`](SPEC_FLOATING_PANE_REDOCK_DWELL_2026_09_09.md) | SPEC: Floating-pane redock — hover-intent dwell and a neutral parking zone |
 | [`SPEC_HEADLESS_TRANSIENT_RETRY_2026_08_31`](SPEC_HEADLESS_TRANSIENT_RETRY_2026_08_31.md) | Transient-failure retry for turns with no rendered pane |
+| [`SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09`](SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md) | Spec: Instruction and Memory Portability |
 | [`SPEC_JEKT_CROSS_CHANNEL_TRUST_2026_09_02`](SPEC_JEKT_CROSS_CHANNEL_TRUST_2026_09_02.md) | SPEC: Cross-channel jekt trust — closing the last unverifiable same-machine tier |
 | [`SPEC_JEKT_LAN_WAN_TRUST_HARDENING_2026_08_13`](SPEC_JEKT_LAN_WAN_TRUST_HARDENING_2026_08_13.md) | Spec: Securing LAN and WAN tier jekt delivery — closing cross-tenant and cross-network trust gaps |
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

AgentA-asaf:
Both findings correct.

**P2 — Phase 0b's first option is incomplete.** Confirmed and fixed in `f81b5c644`. Pointing `export_bundle` at the ref tables closes bind-then-export and leaves import-then-launch exactly where it was, since all three import paths still write only inline `mcp_servers` while launch reads refs. §5.4 and the Phase 0b entry now both state that the ref-authoritative option is three changes — export reads refs, **imports create and bind the managed rows**, and existing inline-only data is migrated or it vanishes from every consumer at once — plus a recommendation and which part to size first. Good catch: as originally written that option read as the cheap one, which is precisely how the second bug would have survived the fix for the first.

**P1 — status claims work not in this ancestry.** Also correct. #3149 branched from main before #3147 merged, so the status line advertises Phases 1-2 as landed in a revision that does not contain them, and the generated index republishes that. I am holding this PR and will rebase it onto main once #3147 lands, which makes the claim true rather than softening it — the two are a set, and the finding here is what re-plans the phases that #3147 starts. Will re-request review after the rebase.

AgentA-asaf:
Rebased onto main now that #3147 is merged — the status line's claim that Phases 1-2 landed is true in this ancestry. @reagentx-workflow review again

PR #3150, opened 2026-09-10
Clare@starpower: fix(cef): stop verify-cef-framework-darwin reporting a stripped framework as UNPATCHED

AgentO-asaf:
## The bug

Running the script against the shipped `.app` prints:

```
✗ ... has a symbol table but NO BeginWindowDrag slot.
  This is the UNPATCHED upstream CEF — native window drag /
  floating-pane resize will silently no-op.
```

**against a framework that is correctly patched.** A confident, alarming, wrong answer about a good build — and pointing it at a `.app` is the obvious first instinct when you want to know what actually shipped.

## Cause

The stripped-vs-unpatched test was `read_syms | grep -q .` — *"any symbols at all"*. That assumes `strip` empties the symbol table. On macOS it doesn't: `strip` removes **local** symbols and keeps external/dynamic ones, so a stripped framework still prints thousands of `nm` lines and falls into the UNPATCHED branch.

Measured on the 148.23.25 build:

| | total symbols | local (lowercase type) |
|---|---|---|
| bundled `.app` | 4,354 | **0** |
| build tree | 1,232,824 | 1,228,470 |

The patch symbol is *local*, so the right question is whether any **local** symbols survive — not whether any symbols do.

## Verified all three verdicts

Both frameworks on this machine carry the patch, so the UNPATCHED branch couldn't be exercised with a real binary — I used an `nm` shim rather than leave it untested:

| Case | Result |
|---|---|
| patched + unstripped | `✓` exit **0** |
| unpatched + unstripped (shimmed) | `✗` exit **1** |
| stripped | `?` exit **2** |

## The exit-2 message now says what to do instead

"Cannot verify" is unsatisfying when the `.app` is the thing you care about. Stripping doesn't touch executable code, so:

```bash
otool -s __TEXT __text <binary> | tail -n +3 | shasum
```

Equal hashes prove identical code. That's how the shipped 148.23.25 DMG was confirmed — bundled and build-tree binaries hash the same (`8c3bc21a…`) despite differing by 195 MB of symbol table.

### DIFF
--- a/.changesets/1789021692-fix-cef-stop-verify-cef-framework-darwin-calling-a-stripped--7jex.md
+++ b/.changesets/1789021692-fix-cef-stop-verify-cef-framework-darwin-calling-a-stripped--7jex.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(cef): stop verify-cef-framework-darwin calling a stripped framework UNPATCHED
--- a/scripts/verify-cef-framework-darwin.sh
+++ b/scripts/verify-cef-framework-darwin.sh
@@ -82,8 +82,20 @@ if read_syms | grep -qE "$PATTERN"; then
     exit 0
 fi
 
-# No match. Distinguish "has symbols but unpatched" from "stripped (no symtab)".
-if read_syms | grep -q .; then
+# No match. Distinguish "has symbols but unpatched" from "stripped".
+#
+# `strip` on macOS removes LOCAL symbols and keeps the external/dynamic ones, so
+# a stripped framework still prints thousands of nm lines. Testing for "any
+# output at all" therefore misidentifies every stripped binary as UNPATCHED:
+# measured on a correctly-patched 148.23.25 build, the bundled framework has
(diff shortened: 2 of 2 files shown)


a5af:
@codex review

AgentO-asaf:
Correct — accidental, now dropped. The PR is 2 files: the script and its changeset.

Confirmed it was pure noise before removing it: **157 changed lines, all `dev` → `devOptional` flag flips, zero changes to any `version`, `resolved`, or `integrity` field.** That's a local npm version regenerating the lockfile — picked up by `task package:macos` running an install while I was building the DMG, then swept in by `git add -A`.

`git diff origin/main -- package-lock.json` is now empty, and I re-ran the script after amending to be sure the revert didn't disturb it: stripped → exit 2, patched → exit 0.

Worth noting this is the second time in this stretch of work that `git add -A` has pulled in something unrelated — the first was a `node_modules/.vite` cache artifact in agentmuxai/agentmux-cloud#71. Two for two on incidental files from build tooling running in the background. I should be staging explicit paths on any branch where I've also run a build, rather than relying on `.gitignore` to have anticipated every tool.

PR #3151, opened 2026-09-10
fix(agent): restore pane tab strip + header layout, and drop the remaining new-tab flash

GenericAgentX-asaf:
## Summary

Follow-up to #3136. Two of these are **regressions #3136 shipped** (it was merged before these fixes were pushed — my fault, they were sitting uncommitted locally); the third is the remaining flash the repo owner found once switching was working.

**1. The tab strip vanished from every agent pane (regression, severe).** Chrome was gated on `NodeModel.hasEverBeenMultiMember` — but `AgentPaneChrome` owns the tab strip, the tab strip owns the **"+"** button, and "+" is the only way to reach a 2nd stack member. So a single-member pane rendered no strip, no "+", and could never become multi-member: the whole feature was unreachable, and existing panes lost their strip. Chrome now hoists for **every** agent pane. The strip's own visibility rules (`shouldShowTabStrip`) are unchanged and still live inside chrome. `hasEverBeenMultiMember` is removed entirely (no other consumer). The `hoisted` memo is latched instead, since `effectiveViewType()` reads the active member's block meta and is briefly undefined while a newly-activated member loads.

**2. The header rendered as unstyled, vertically-stacked divs (regression).** `block.scss` only ever wrote `.block-frame-default-header`'s ruleset nested inside `.block.block-frame-default .block-frame-default-inner`, so reusing `BlockFrame_Header` outside that ancestor chain matched none of it. Extracted to a shared mixin (`frontend/app/mixins.scss`), applied in both places — the original nesting is preserved verbatim, so its compiled selector and specificity are unchanged.

**3. "+" still flashed.** `handleNewAgentTab` held `holdLeafRevealGate` for the whole `pane.open` round trip. That gate existed because `pushBlockOntoStack` used to force a whole-leaf remount; it doesn't anymore, so the gate had stopped protecting anything and started *causing* the flash — `gatingNodeIds()` hides the whole node, so chrome blinked out and back despite nothing about it rebuilding. Content settle is still covered by `Block`'s own ready-gate cross-fade.

## Why #3136's tests didn't catch 1 or 2

They cover the hoisted-vs-passthrough routing decision with a stubbed chrome renderer — they never assert the strip's "+" stays reachable, and they can't see CSS at all. Both were only findable by looking at the running app, which is what happened.

## Test plan

- [x] `npx vitest run` — 3900 passing / 3 skipped, against current main.
- [x] Tab **switching** confirmed fixed live by the repo owner (that part of #3136 works).
- [ ] Still needs a live check on this branch: strip + header render correctly, and **"+"** no longer flashes the header/strip while the new tab's own content still fades in.


a5af:
@codex review

GenericAgentX-asaf:
All three addressed in 14a9fb0e8.

1. **Dropped `--accent-color` fallback** — you're right that it contradicted the "zero behavior change" claim, so I've stopped claiming it. Kept the drop (block.scss is grandfathered in `.stylelintignore` for that pre-existing raw hex; `mixins.scss` isn't, so carrying it would newly fail `color-no-hex`) and documented it inline as the one deliberate deviation, with why it's safe: `--accent-color` is defined unconditionally in `theme.scss` and only ever overridden, never unset, so the fallback arm was unreachable.
2. **Stale `hasEverBeenMultiMember` comments** — updated all of them across both files. The two remaining mentions in `pane-leaf-chrome.tsx` are deliberate: they're in the `hoisted` memo explaining what the field was and why gating on it was a catch-22.
3. **Unused `createEffect` import** — removed.

GenericAgentX-asaf:
All three addressed in f5c141f7e — thanks both; the drag-preview one was caught independently by ReAgent (P1) and Codex (P2).

**1. Drag previews lost their header.** `noHeader` was unconditionally `true`, but `renderPreview` builds a plain `<Block preview>` with the raw leaf nodeModel and no chrome around it. `pane-leaf-chrome.tsx` now tags the wrapper it passes to hoisted content (`paneChromeHoisted`) and `noHeader` reads that — true only where something above really is rendering a replacement header. Fixed this way rather than making `blockframe.tsx` ignore `noHeader` in preview mode, since that would also have changed `LauncherViewModel`'s deliberate always-headerless behavior.

**2. Hoisted header didn't focus its pane.** Correct — it's a sibling above `.block`, so nothing reached `BlockFull`'s focus handlers. `AgentPaneChrome`'s root now calls `nodeModel.focusNode()` on click/focusin.

**3. Duplicate `@keyframes`.** Moved to `app.scss` (the one globally-imported stylesheet) rather than to `mixins.scss`'s top level — worth flagging that the obvious fix is actually **worse**: I compiled it both ways, and top-level placement leaks a copy into all 10 of that module's consumers, including unrelated ones like `button.scss`. Now: exactly 1 definition in `app.scss`, 0 in `block.scss`/`agent-view.scss`/`button.scss`, verified by compiling each with `sass`.

GenericAgentX-asaf:

PR #3152, opened 2026-09-10
fix(srv): Phase 0b — bundle ref tables become authoritative for export, with backfill

AgentA-asaf:
Phase 0b of `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md`, the reconciliation #3149 identified as a prerequisite for Phase 3. Tracking: #3148.

**The bug.** Bundle components lived in two stores that were never synced. Agent launch read `db_bundle_skills_ref` / `db_bundle_mcp_ref`; ABF export read the inline `db_bundles.skills` / `.mcp_servers` columns. Binding a skill or MCP server in the Armory writes only a ref row, so it ran at launch, showed in the UI, and **exported as an empty `skills/` and `mcp/` directory with no warning** — in a format this repo advertises as a backup.

**The fix: the ref tables become authoritative.** They are what the UI writes and what launch already trusts.

**Two commits, and they must ship together.** Export reading refs before the backfill has run would make every pre-existing bundle export empty.

| Commit | What |
|---|---|
| `f872be33b` | `m0030` carries every bundle's inline components into the ref tables. Additive only. |
| `46a0c0831` | `export_bundle` stops reading `bundle.mcp_servers`; both export paths resolve from the refs. |

**Migration notes** — the parts worth review attention:
- **Two stores.** `db_bundles` lives in the shared store; the catalog and ref tables live in the channel store, because the refs carry foreign keys to `db_skills`/`db_mcp_servers` and therefore cannot also FK to bundles (`migrations.rs:721-741`). Bundles are read from the former, refs written into the latter, resolving the shared store exactly the way `m0021` does with the same fall-back-and-warn posture.
- **Skills** are bare ids and bind directly. An id whose catalog row is gone is dropped with a warning — the ref tables cannot represent it and export could never render it either.
- **MCP servers** have no catalog row anywhere today, so each inline entry becomes a bundle-scoped `db_mcp_servers` row plus a ref, created in one transaction by `bundle_mcp_upsert_unique`. A nameless entry gets a stable synthetic name instead of being dropped.
- **Idempotent three ways**: `INSERT OR IGNORE` on refs, name uniqueness on the catalog, and a re-run that binds nothing new. Best-effort per bundle: one malformed blob must not abort a migration whose read switch ships in the same release.

**Export notes:**
- `export_bundle` is now a pure renderer — it reads no component column off `bundle` and takes both lists resolved. That split (skills resolved, MCP read inline) is exactly how the two paths drifted apart.
- One `resolve_bundle_components` serves both export paths, so they cannot disagree.
- **Scope is deliberately narrower than `effective_skills`/`effective_mcp_servers`**, which additionally union an agent's own binds, globals and the legacy blob. Those answer "what does this agent run with"; this answers "what is in this bundle", which is what an ABF describes. Documented at the resolver.
- `missing_skill_ids` is now always empty and kept for wire compatibility — a ref cannot dangle the way a bare inline id could.

**Tests** (8 new, all failing before the change):

| Test | Pins |
|---|---|
| `export_carries_ref_bound_components_the_inline_columns_never_had` | the regression itself, plus that redaction still applies when resolving from a different source |
| `a_bound_mcp_server_with_unparseable_config_warns_instead_of_vanishing` | the malformed-data guarantee at its new layer |
| `resolving_ignores_catalog_rows_this_bundle_is_not_bound_to` | `managed_list` returns globals too; only bound rows are the bundle's |
| `carries_inline_skills_and_mcp_servers_into_the_ref_tables` | the backfill, config surviving verbatim |
| `rerunning_binds_nothing_new` | idempotency |
| `a_malformed_blob_is_logged_and_the_rest_of_the_bundle_still_migrates` | partial failure |
| `an_inline_skill_id_with_no_catalog_row_is_dropped_not_fatal` | dangling id |
| `a_nameless_mcp_entry_gets_a_stable_synthetic_name` | two nameless entries must not collide |

The renderer's MCP tests now pass entries through a test shim. The malformed-JSON test in `bundle_export.rs` is **kept as a pointer** to where the guarantee moved rather than deleted, so it stays findable from the module that used to own it.

**Deliberately NOT in this PR**, each a follow-up on #3148:
1. **Imports still write only the inline columns**, so an imported MCP server remains inert at spawn — the other half of §3.4. Next PR.
2. `bundle_validate.rs` still parses the inline columns, so `bundle.validate` now reports on data nothing else reads. It is report-only and blocks nothing, but it is now inaccurate.
3. `bundle_delete` still orphans ref rows.
4. The inline columns themselves, and the frontend `draftToWire` that rewrites them on every save. Dropping them is a schema change and wants its own release once this proves out.

**Verification**
```
cargo test -p agentmux-srv   → 3225 passed, 0 failed
```

### DIFF
--- a/.changesets/1789022722-fix-srv-bundle-exports-now-carry-the-skills-and-mcp-servers--e638.md
+++ b/.changesets/1789022722-fix-srv-bundle-exports-now-carry-the-skills-and-mcp-servers--e638.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): bundle exports now carry the skills and MCP servers actually bound to the bundle, with a migration carrying existing inline data across
--- a/agentmux-srv/src/backend/bundle_export.rs
+++ b/agentmux-srv/src/backend/bundle_export.rs
@@ -360,14 +360,26 @@ pub(crate) fn sanitize_context_relative_path(path: &str) -> Option<String> {
     Some(parts.join("/"))
 }
 
-/// Export a bundle + its already-resolved skills into the ABF v0.1 layout.
+/// Export a bundle + its already-resolved components into the ABF layout.
 ///
-/// `skills` must already be resolved from `bundle.skills` (a JSON array of
-/// skill ids) via `Store::skill_get` per id — this function does no
-/// lookups. Callers should skip ids that failed to resolve (deleted skills)
-/// before calling this; it does not distinguish "missing" from "not
-/// passed."
-pub fn export_bundle(bundle: &Bundle, skills: &[Skill]) -> BundleExport {
+/// **This function does no store lookups and reads no component column off
+/// `bundle`.** Both `skills` and `mcp_servers` must be resolved by the caller
+/// from `db_bundle_skills_ref` / `db_bundle_mcp_ref`, which are authoritative
+/// for what a bundle contains (Phase 0b of
+/// SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md). It used to read
+/// `bundle.mcp_servers` inline here while skills came in resolved — that split
+/// is exactly how export and launch drifted apart, so the renderer now takes
+/// both the same way and has no opinion about where they came from.
+///
+/// `mcp_servers` entries are the exporter's own entry shape: the server's
+/// config object with its `name` alongside, which is what `redact_mcp_entry`
+/// reads. Callers skip ids that failed to resolve before calling; this does
+/// not distinguish "missing" from "not passed".
+pub fn export_bundle(
+    bundle: &Bundle,
+    skills: &[Skill],
+    mcp_servers: &[Value],
+) -> BundleExport {
     let root_slug = derive_slug(&bundle.name);
     let mut files = Vec::new();
     let mut skipped_skills = Vec::new();
@@ -522,8 +534,7 @@ pub fn export_bundle(bundle: &Bundle, skills: &[Skill]) -> BundleExport {
     // ------------------------------------------------------------------
     // mcp/<slug>.server.json + inferred accounts/requirements.json
     // ------------------------------------------------------------------
-    let mcp_entries: Vec<Value> =
-        parse_json_field_or_warn(&bundle.mcp_servers, "mcp_servers", &mut warnings);
+    let mcp_entries: Vec<Value> = mcp_servers.to_vec();
     let mut used_mcp_slugs: HashSet<String> = HashSet::new();
     let mut requirements: Vec<Value> = Vec::new();
     for (index, entry) in mcp_entries.iter().enumerate() {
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

AgentA-asaf:
All three findings confirmed and fixed in `498013a0d`. Two of them were the same class of bug this PR exists to remove, which is a good sign the reviews were reading it the right way.

**Codex P1 — imports must bind.** Correct, and it was a regression *this PR introduced*: with export reading only refs, an import writing only the inline columns produces a bundle that exports empty — and whose MCP servers never reach a spawned agent either, since launch reads refs too. All three import paths now bind after `bundle_upsert` (the bind checks bundle existence first, so the ordering is load-bearing). Failures are warnings, not errors: the bundle row is already committed by that point, so aborting would leave a half-imported bundle. Pinned end to end by `an_imported_bundle_exports_the_components_it_arrived_with`, which imports an ABF carrying a skill and a server and exports it back out.

**Codex P2 — duplicate names.** Also correct, and subtle: I was treating every `already exists` error as an idempotent re-run, which silently dropped the second of two same-named entries **permanently**, since export no longer falls back to the inline column. Idempotence is now decided by *what is already bound* rather than by an error string — those two conditions produce identical error text and mean opposite things. A colliding name gets a numeric suffix instead of being dropped, including when it collides with a visible global server. `two_entries_sharing_a_name_are_both_kept` covers both the first run and the re-run.

**ReAgent P2 — transport.** Correct: hardcoding `stdio` gave the row a transport column contradicting its own config JSON for http/sse entries. It now comes from the entry's `type`, with `stdio` as the fallback for an entry that says nothing. `an_entrys_own_type_becomes_the_transport`.

All three fixes land in one shared place — `Store::bundle_mcp_bind_inline_entries` — used by the m0030 backfill and all three import paths, so a server recovered from an old inline column and one that arrives in an import cannot end up shaped differently. That was the actual lesson of §3.4.

`cargo test -p agentmux-srv`: 3228 passed, 0 failed.

The PR body's "deliberately not in this PR" list is now shorter by one: imports are done here, because leaving them out would have shipped a regression. Still deferred: `bundle_validate.rs` reading the inline columns, `bundle_delete` orphaning refs, and removing the columns themselves.

AgentA-asaf:
P1 confirmed and fixed in `abe2465c5`. This one was the sharpest catch of the three rounds, because it was the bug this PR exists to fix, reintroduced inside the fix for it.

Skipping on a name match conflated two opposite conditions: *this exact entry was already migrated* and *a completely different server happens to share its name*. The second case dropped a real entry permanently and, as you noted, without the warning every other drop path in that function emits.

**Identity is now the config object**, matched as a multiset consume so N byte-identical entries stay N across re-runs. Name is used only for collision avoidance, and a disambiguated name now warns as well, so nothing in that function is silent any more.

Worth adding: matching on name was also wrong about its own output. An entry disambiguated to `foo-2` on the first run stores that name while still deriving `foo` from its config, so a re-run would have created a second copy — the opposite failure, in the same line.

New test `an_inline_entry_is_kept_when_a_different_server_already_holds_its_name` sets up exactly your scenario: an Armory-bound `github` pointing at one binary, an inline `github` pointing at another. Both survive, both configs intact, and a re-run still creates nothing.

`cargo test -p agentmux-srv`: 3229 passed, 0 failed.

PR #3153, opened 2026-09-10
fix(srv): Phase 0b cleanup — purge refs on bundle delete, validate the bound components

AgentA-asaf:
The two items #3152's body listed as deferred. Follow-up to #3152 (Phase 0b). Tracking: #3148.

**1. Deleting a bundle left its component refs behind.**

`bundle_delete` lives on the store that owns `db_bundles`, which does not have the ref tables at all — they sit beside the catalog tables they key into, in the other database. That is also why no foreign key reaches across (`migrations.rs:721-741`). So nothing removed a deleted bundle's refs, and a later bundle reusing that id would inherit components nobody chose.

The handler now purges them, since it is the only place both stores are in scope. Best-effort: the bundle is already gone by then, so a failure here must not turn a successful delete into an error. `managed_unbind_all_for_bundle` mirrors `managed_delete`'s existing purge from the other side — neither can rely on the declared FKs, which only cascade when `PRAGMA foreign_keys` is ON, and it is only ever set in tests.

**2. `bundle.validate` was reporting on retired data.**

It parsed the inline `mcp_servers`/`skills` columns, so after #3152 it validated data nothing else consumes. It now takes the resolved MCP entries the same way `export_bundle` does — the module keeps its no-Store-access rule, and the handler resolves them when the draft names a persisted bundle. An unsaved draft still validates without touching the store, which is what that RPC exists for.

**The skills checks are gone rather than moved**, because both became impossible to fail. `db_bundle_skills_ref` has a `PRIMARY KEY (bundle_id, skill_id)` so an id cannot repeat, and ids come out of a join against the catalog rather than out of parsed JSON, so nothing can be malformed. The two tests are replaced by one that records this, rather than deleted quietly — same treatment as the malformed-MCP test in #3152.

**Tests**

| Test | Pins |
|---|---|
| `deleting_a_bundle_takes_its_component_refs_with_it` | refs purged, and a bundle reusing the id starts empty |
| `skills_are_no_longer_validated_because_both_checks_became_impossible` | the inline column produces no issues, and why |

Validator tests now go through a `validate_with_inline_mcp` shim, the same pattern #3152 used for the renderer: these tests are about the checks, not about resolution.

**Verification**
```
cargo test -p agentmux-srv   → 3229 passed, 0 failed
```

Remaining on #3148 after this: removing the inline columns themselves and the frontend `draftToWire` that still rewrites them. That is a schema change and wants its own release now that the read paths are off them.

### DIFF
--- a/.changesets/1789024994-fix-srv-deleting-a-bundle-purges-its-component-refs-bundle-v-5z1d.md
+++ b/.changesets/1789024994-fix-srv-deleting-a-bundle-purges-its-component-refs-bundle-v-5z1d.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): deleting a bundle purges its component refs; bundle.validate checks the bound components rather than the retired inline columns
--- a/agentmux-srv/src/backend/bundle_validate.rs
+++ b/agentmux-srv/src/backend/bundle_validate.rs
@@ -47,13 +47,27 @@ pub struct ValidationReport {
     pub issues: Vec<ValidationIssue>,
 }
 
-/// Run every structural check against a bundle's JSON-encoded columns.
-pub fn validate_bundle(bundle: &Bundle) -> ValidationReport {
+/// Run every structural check against a bundle.
+///
+/// `mcp_entries` are the bundle's MCP servers **resolved from
+/// `db_bundle_mcp_ref`**, which is authoritative for what a bundle contains
+/// (Phase 0b, `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md`). They
+/// are passed in rather than read off `bundle` for the same reason
+/// `export_bundle` takes them: this module does no Store lookups, and reading
+/// the inline `bundle.mcp_servers` column would validate data that nothing
+/// else consumes any more.
+///
+/// Pass an empty slice for an unsaved draft — a bundle with no row yet has no
+/// bindings to check.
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

AgentA-asaf:
All three fixed in `41a3eaab1`.

**P1 — wrong delete path.** Correct, and the important one: there are two delete RPCs and I fixed the one the product does not use. `deletememory` in `agent_handlers/bundle.rs` is what `rpc-api/bundle.ts` calls, so the Armory kept orphaning refs exactly as before. Both paths now call one shared `purge_bundle_component_refs` — a second copy of the logic in the second handler would have been the same bug waiting to happen again.

**P2 — discarded warnings.** Also correct, and it defeated the point of the change: the resolver drops a bound server whose config will not parse and reports it in a warning; discarding that meant validate returned `is_valid: true` for precisely the malformed component the switch to ref-resolution was meant to let it see, and nothing downstream could notice because the entry was already gone from the list. Those warnings are now errors on the report — a config that will not load is not a stylistic warning like a duplicate name.

**P2 — swallowed store errors.** Correct: `unwrap_or_default()` turned a failed resolution into "this bundle has no components", so a check that never ran came back clean and trustworthy-looking. It now fails the RPC, which is the result the UI is already built to show.

New tests: `validate_reports_a_malformed_bound_component_instead_of_passing`, and the delete test now goes through the shared helper and asserts the purge already happened.

`cargo test -p agentmux-srv`: 3230 passed, 0 failed.

PR #3154, opened 2026-09-10
feat(srv): Phase 3 prerequisites — record what each provider reads, and one ownership test

AgentA-asaf:
The two prerequisites review identified for Phase 3 of `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md`. **Additive only** — nothing reads either yet and no behaviour changes. Tracking: #3148.

**1. `native_instruction_sources` — what a provider *reads*.**

`startup_instructions_filename` names the single file AgentMux **writes**: the canonical target it picked per provider. That is a different question from what the provider picks up on its own, and usually a smaller answer:

| Provider | Writes | Also reads |
|---|---|---|
| Copilot | `AGENTS.md` | `.github/copilot-instructions.md`, `CLAUDE.md`, `GEMINI.md` |
| pi | `.pi/APPEND_SYSTEM.md` | `.pi/SYSTEM.md` (which *replaces* pi's default system prompt) |
| everyone else | their target | nothing further confirmed |

Resolving only the write target would leave most of a Copilot project's instructions invisible and unportable — precisely the round-trip guarantee §4 claims (Codex, #3144).

Populated per provider from **each entry's own already-cited evidence**, under the same rule as the field above it: listed only where the provider's docs confirm it, never guessed. Where the write target is the only confirmed source, that is the whole list. An under-complete list shows the operator less than the truth, which is recoverable; a guessed one shows them something false, which is not.

Three invariants tested, the useful one being **every write target is also a read source** — otherwise AgentMux is writing a file nobody opens. That check passes for all ten providers, which is also what validates the population.

**2. `is_agentmux_managed_instructions` — one ownership test that knows both markers.**

`CLAUDE.md` carries `CLAUDE_MD_MANAGED_MARKER`; every other startup-instructions file carries `STARTUP_INSTRUCTIONS_MANAGED_MARKER`, because the two writers grew separately. A check against the first alone reports every AgentMux-generated `AGENTS.md`, `GEMINI.md`, `QWEN.md` and `.pi/APPEND_SYSTEM.md` as `foreign` — inverting the single field Phase 3 most needs to be right, since it decides whether a file is ours to touch.

**The two writers deliberately keep their own narrower checks**, and that is documented at the helper rather than left to look like an oversight: each asks "did I write this, in my format, such that I may rewrite it". A `CLAUDE.md`-marked file is ours but is not the other writer's to rewrite. This helper answers the broader question, which is the one a read-only tracker asks.

**Verification**
```
cargo test -p agentmux-srv   → 3234 passed, 0 failed (5 new)
```

Next on #3148: the resolver itself, which uses both of these to answer "what will this agent actually read" — useful on its own, since the Armory can then show an operator their agent's real instruction set for the first time.

### DIFF
--- a/.changesets/1789026221-feat-srv-provider-registry-records-every-instruction-file-a--a1fy.md
+++ b/.changesets/1789026221-feat-srv-provider-registry-records-every-instruction-file-a--a1fy.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): provider registry records every instruction file a provider reads, not just the one AgentMux writes
--- a/agentmux-srv/src/backend/agent_config.rs
+++ b/agentmux-srv/src/backend/agent_config.rs
@@ -1039,6 +1039,33 @@ fn resolve_claude_md_side_paths(
 /// see the doc comment below).
 const STARTUP_INSTRUCTIONS_MANAGED_MARKER: &str = "<!-- agentmux:managed-startup-instructions -->";
 
+/// Does this file's content say AgentMux wrote it?
+///
+/// There are two markers, not one: `CLAUDE.md` carries
+/// [`CLAUDE_MD_MANAGED_MARKER`] and every other startup-instructions file
+/// carries [`STARTUP_INSTRUCTIONS_MANAGED_MARKER`], because the two writers
+/// grew separately. Anything asking "is this ours" has to know both — a check
+/// against only the first reports every AgentMux-generated `AGENTS.md`,
+/// `GEMINI.md`, `QWEN.md` and `.pi/APPEND_SYSTEM.md` as foreign, inverting the
+/// answer for most providers (Codex, PR #3144).
+///
+/// Deliberately content-based rather than path-based: the marker is what makes
+/// the claim, and a file's name says nothing about who wrote it. Phase 3 of
+/// `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md` reuses this to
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentA-asaf:
Confirmed and fixed in `7a29ed799` — and the doc you cited says more than the finding did, so I took the rest of it too.

Claim 3 verified: `.github/instructions/**/*.instructions.md` are path-scoped instructions Copilot genuinely reads. A resolver driven by the literal list would have missed every one of them in any repository using them, which is the failure mode this field exists to prevent: an incomplete answer, presented confidently.

Added `native_instruction_dirs` as a structured directory scan (`dir`, `suffix`, `recursive`) rather than a glob string. There is no glob engine in the tree, one dependency for one pattern shape is a poor trade, and the struct forces the two things a glob leaves implicit — whether to descend, and what actually qualifies — to be written down.

Two more claims from that doc I did **not** implement, recorded inline as out of scope rather than missed: `$HOME/.copilot/copilot-instructions.md` (claim 4) and `$COPILOT_CUSTOM_INSTRUCTIONS_DIRS` (claim 6). Both are real. Both live outside the agent working directory that these two fields are relative to, and both are machine-global rather than project state — so they do not travel with a bundle, which makes them a genuinely different question from the one Phase 3 answers. Worth revisiting if the answer to "what is this agent reading" should include machine-level configuration; it is not a silent omission.

Two new tests: the Copilot scan is pinned by name, and every declared scan must be usable and stay inside the working directory.

PR #3155, opened 2026-09-10
Clare@starpower: docs(cef): document the Metal Toolchain requirement for from-scratch macOS CEF builds

AgentO-asaf:
## Found by running the Phase D macOS build

It dies ~13% in — roughly 7,500 of 58,156 targets:

```
FAILED: gen/angle/mtl_internal_shaders_autogen.air
error: cannot execute tool 'metal' due to missing Metal Toolchain;
       use: xcodebuild -downloadComponent MetalToolchain
```

This is the upgrade spec's *"confirm the 152 build toolchain requirements haven't moved (VS on Windows, Xcode on macOS, sysroot on Linux)"* item. **For macOS they moved.**

## Why a normal availability check won't catch it

Xcode 26 ships the `metal` **binary** but not the compiler behind it:

```bash
$ xcrun -f metal
/Applications/Xcode.app/.../usr/bin/metal      # resolves fine
```

So `command -v` / `xcrun -f` style checks all pass, and the tool still refuses to run. The runbook therefore documents a **compile** check, not a presence check:

```bash
echo 'kernel void k() {}' > /tmp/t.metal && xcrun metal -c /tmp/t.metal -o /tmp/t.air
```

That matters because `xcodebuild -downloadComponent` reports success regardless of whether the compiler ends up usable — and discovering the problem at target 7,500 costs an hour of build time.

## Confirmed genuinely new in 152

`angle_metal_internal_shaders_to_air` has **zero hits across both 148 build logs** — 148 never compiled that target at all. So this isn't a local misconfiguration on my machine.

## Two other Phase D questions answered by the same run

**The macOS hermetic Xcode pin** — flagged in the spec as *"unconfirmed — verify at Phase D build time"* — is **no change**. `mac_toolchain.py` prints `Skipping Mac toolchain installation for mac` on 152 exactly as on 148; the system Xcode is used.

**`patcher.py`: 118 patches, 118 applied, 0 failed** against real Chromium 152 source. The entire set forward-ports without a single conflict. Phase A predicted this from file-identity analysis (16 of 18 files byte-identical upstream 7778↔7977); this is the first confirmation by actually building. For comparison, 148 was 112 applied / 3 skipped.

Also: `translator.py` 955 files, `version_manager.py` **26/26 hashes match** — the 148 build logged `WARN: version_manager` and carried on, leaving those unverified.

## Status

Build resumed after installing the component and is running clean. This PR is the documentation; the build result will go to #3108 when it finishes.

### DIFF
--- a/.changesets/1789026385-docs-cef-chromium-152-needs-the-metal-toolchain-component-on-gqgp.md
+++ b/.changesets/1789026385-docs-cef-chromium-152-needs-the-metal-toolchain-component-on-gqgp.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+docs(cef): Chromium 152 needs the Metal Toolchain component on macOS
--- a/docs/cef-build/build-patched-framework-macos.md
+++ b/docs/cef-build/build-patched-framework-macos.md
@@ -67,6 +67,31 @@ not the `148.0.9` this table previously claimed):
 - macOS arm64 host (Apple Silicon) with Xcode + command-line tools.
 - ≥ 32 GB RAM, ≥ 120 GB free disk.
 - `depot_tools` on PATH; the chromium hooks pull the macOS toolchain automatically.
+- **Chromium 152+ only — the Metal Toolchain component:**
+
+  ```bash
+  xcodebuild -downloadComponent MetalToolchain     # ~688 MB
+  ```
+
+  Xcode 26 ships the `metal` *binary* but not the compiler behind it, so
+  `xcrun -f metal` resolves and the tool still refuses to run. Chromium 152 added
+  an ANGLE step that needs it and 148 did not — verified: zero hits for
+  `angle_metal_internal_shaders_to_air` across both 148 build logs. Without the
+  component the build dies ~13% in, at roughly 7,500 of 58,000 targets:
+
+  ```
+  FAILED: gen/angle/mtl_internal_shaders_autogen.air
+  error: cannot execute tool 'metal' due to missing Metal Toolchain;
+         use: xcodebuild -downloadComponent MetalToolchain
+  ```
+
+  Confirm it works before starting rather than after — the download itself
+  succeeds silently whether or not the compiler ends up usable:
+
+  ```bash
+  echo 'kernel void k() {}' > /tmp/t.metal && xcrun metal -c /tmp/t.metal -o /tmp/t.air \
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentO-asaf:
You're right and I was wrong. Fixed in `b7b3816f8`, and the PR title/description are now misleading — I'll correct those too.

`docs/cef-patches/README.md` §Metal already documented this **for the 148 build**:

> Getting Metal to build required a one-time Xcode-26 toolchain repair. Xcode 26 ships the Metal compiler as a *separate* downloadable component…

I hadn't read that file.

## My evidence didn't support the claim

Zero hits for `angle_metal_internal_shaders_to_air` in the 148 logs proves only that the target **didn't re-run** — both 148 builds were incremental against a tree whose shaders already existed. Measured:

```
148 tree:  gen/angle/mtl_internal_shaders_autogen.air   mtime 2026-06-02
my 148 rebuild:                                          2026-09-09
```

Built by the original from-scratch 148 build after the toolchain repair the README describes. My rebuild found it up to date and skipped it.

Also checked your `angle_enable_metal` point: **neither** milestone's `args.gn` sets it, so both default to `true` and both need the component from scratch. It surfaced on 152 only because that was the first *fresh* tree on this machine in months.

## What changed

Both documents now frame it as an **Xcode 26 requirement for any milestone**, with an explicit note that an incremental rebuild won't hit it — which is exactly the trap that misled me, and would mislead the next person the same way.

The runbook also now cross-references the README section, which carries a second failure mode I'd otherwise have rediscovered painfully: `-downloadComponent` can itself be broken by a stale `DVTDownloads.framework`, needing `XcodeSystemResources.pkg` installed first.

## The generalisable error

*"Absent from the logs"* is not *"was not required."* Same shape as two earlier confounds in this work — the `pipefail`/SIGPIPE race and the DWARF path embedding — evidence that looks conclusive but answers a narrower question than the one being asked. Worth me noticing that this is now three of the same kind.

AgentO-asaf:
Correct — the header said *"one new toolchain requirement found"* while the body two paragraphs down retracted precisely that. Worse than either version alone, since a status block at the top of a spec exists to be skimmed.

Fixed in `94ac471ae`. The header now states what the run actually established — clean forward-port, both Phase-A build-time questions answered — and flags the retraction inline rather than leaving it to be discovered further down.

Also swept both documents for other stale "new in 152" framing: none left.

PR #3156, opened 2026-09-10
feat(srv): Phase 3 — resolve what an agent actually reads as project instructions

AgentA-asaf:
Phase 3 of `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md`, the read half. Builds on #3154. Tracking: #3148.

**The gap.** AgentMux could say what it *wrote* into an agent's working directory and nothing about what the agent would *read* there. The only thing ever recorded about a file it found already present was a single boolean, `import_line_offered` — no path, no hash, no content (spec §2.4). A repository's own `CLAUDE.md` can be the majority of what an agent is told, and it was invisible.

`resolve_project_instructions` answers the question. For each file: path, whether it exists, size, a SHA-256 of its bytes, an `agentmux`/`foreign` owner, and its content.

**What is in the set.** The provider's `native_instruction_sources` — not the single file AgentMux writes — plus `.claude/AGENTMUX_MEMORY.md` where the `@`-import fallback applies, since that file is genuinely read and is AgentMux's own contribution. Copilot's `.github/instructions/**/*.instructions.md` is expanded by a bounded recursive scan.

**The states that are easy to get wrong**, each reported rather than collapsed into absence:

| Case | Reported as |
|---|---|
| Provider looks here, finds nothing | `exists: false`, kept in the list |
| A directory where a file should be | `exists: true` + `error` |
| Bytes that are not UTF-8 | hash present, `content` empty, `error` set |
| File over the 10 MB read cap | `truncated`, still listed |
| Scan matches more than 200 files | explicit overflow entry, not a silent prefix |

Hashing is over bytes rather than the decoded string, so it stays stable for content that cannot be decoded — which is exactly the case where `content` is unavailable. Scan results are sorted, so two calls agree; directory iteration order does not leak through. Symlinked directories are not followed, because a link out of the working directory would silently widen the answer beyond the project.

**`agent.project_instructions`** exposes it, agent-scoped behind `check_s1` — a working directory's contents are not window-scoped data the way a bundle is.

**Read-only, with no write counterpart, permanently.** A foreign instruction file belongs to the repository. `SPEC_CLAUDE_MD_OWNERSHIP_PROTECTION_2026_08_22.md` exists to keep AgentMux out of it, and a portability feature must not become the way around that (spec §5.2).

**Tests** — 11, covering the Copilot four-source case that motivated the field, the recursive scan and its suffix filter, overflow reporting, stable ordering, owner classification both ways, the hash moving when the file does, and each of the error states above.

**Verification**
```
cargo test -p agentmux-srv              → 3247 passed, 0 failed
npx vitest run test/contract/...        → 4 passed (new command added to the backend-only baseline)
```

Next: recording these observations so a foreign file changing under a running agent becomes visible, then carrying them in the export.

### DIFF
--- a/.changesets/1789027537-feat-srv-agent-project-instructions-reports-every-file-an-ag-jf0d.md
+++ b/.changesets/1789027537-feat-srv-agent-project-instructions-reports-every-file-an-ag-jf0d.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): agent.project_instructions reports every file an agent will read as instructions, with a content hash and an agentmux/foreign owner
--- a/agentmux-srv/src/backend/mod.rs
+++ b/agentmux-srv/src/backend/mod.rs
@@ -16,6 +16,7 @@ pub mod mcp_probe;
 pub mod mcp_seed;
 /// Phase E.4.B Phase 4 — pure layout-tree helpers (Rust port of layoutTree.ts).
 pub mod layout;
+pub mod project_instructions;
 pub mod providers;
 pub mod model_catalog;
 pub mod config_watcher_fs;
--- a/agentmux-srv/src/backend/project_instructions.rs
+++ b/agentmux-srv/src/backend/project_instructions.rs
@@ -262,43 +262,73 @@ pub const MAX_TOTAL_INSTRUCTION_BYTES: u64 = 8 * 1024 * 1024;
 /// order is not. Symlinked directories are not followed: a link out of the
 /// working directory would silently widen what this reports beyond the
 /// project, and `read_one` still reads a symlinked *file* normally.
+/// Ceiling on directories entered during one scan.
+///
+/// The file cap alone does not bound a deep or wide tree that contains few
+/// matches — the walk still has to enter every directory to find that out.
+const MAX_SCANNED_DIRECTORIES: usize = 500;
+
+/// Collect qualifying files under one declared scan directory.
+///
+/// **Bounded during traversal, not after it.** Collecting everything and then
+/// truncating leaves the advertised ceiling doing nothing about the cost it
+/// exists to bound: a repository-controlled directory with a very large number
+/// of matches still gets fully walked, allocated and sorted first (Codex, PR
+/// #3156). The walk now stops as soon as it has one file more than the cap,
+/// which is exactly enough to know the limit was hit.
+///
+/// Entries are sorted at each level before descending, so stopping early still
+/// yields the same set every time. Without that, an early stop would return
+/// whatever the filesystem happened to hand back first.
+///
+/// Symlinks are skipped outright: `file_type()` does not follow the link,
+/// unlike `metadata()`, which resolves the target and reports
+/// `is_symlink() == false` for every symlink — so an earlier guard written
+/// against `metadata()` never fired at all. A scanned directory is
+/// repository-controlled, and nothing about "follow this link" can be verified
+/// from here; the declared literal paths are still read through links, which
+/// is an ordinary layout.
+///
+/// Returns true when the cap was reached.
 fn scan_dir(
     working_dir: &Path,
     scan: &providers::InstructionDirScan,
     out: &mut Vec<String>,
-) {
+) -> bool {
     fn walk(
         root: &Path,
         dir: &Path,
         suffix: &str,
         recursive: bool,
         found: &mut Vec<String>,
+        dirs_visited: &mut usize,
     ) {
+        if found.len() > MAX_SCANNED_INSTRUCTION_FILES || *dirs_visited >= MAX_SCANNED_DIRECTORIES {
+            return;
+        }
+        *dirs_visited += 1;
+
         let Ok(entries) = std::fs::read_dir(dir) else {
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

AgentA-asaf:
P1 confirmed and fixed in `654848a5d`. Verified both citations: the side file is written with a bare `std::fs::write(&memory_path, generated_content)` at `agent_config.rs:1207`, and `agent_config.rs:990` calls it "100% AgentMux's own content".

So the marker test returned `foreign` for it every time — backwards for the one field whose stated job is deciding whether a file is ours to touch, and wrong for precisely the file this feature includes *because* it is AgentMux's contribution. The doc comment and the behaviour disagreed, and the doc comment was right.

Ownership for that exact path is now by construction rather than by marker, which is not a guess: AgentMux creates the path, regenerates it on every launch, and the ownership protection exists because the *other* file at that location is not ours. Everything else still goes through the marker test.

Added `only_the_side_file_gets_ownership_by_path` so the path rule cannot quietly widen — an unmarked `CLAUDE.md` must stay `foreign`, since that is the premise the whole ownership protection rests on.

17 project-instruction tests pass.

AgentA-asaf:
All five findings across both reviews are fixed; the last two commits are `001c49603` and `66e3407e7`.

**ReAgent P0 — slug vs UUID.** Correct, and it made the RPC return "no agent" for every real caller. `check_s1` authenticates a slug; `agent_def_get` queries by UUID. The existing test helper sets `id == slug`, so nothing in the suite could have caught it — the new test keeps them deliberately different, which is the actual lesson.

I delegated to the existing `resolve_agent_definition_id` rather than hand-rolling the two-tier lookup. It already carries the registry-only fallback a previous review round forced in, for a live agent that never created a local `db_agents` row; a fresh lookup here would have reopened that hole. `native_memory_handlers` documents the same trap at its own slug resolution, which is where I should have looked first.

**Codex, four findings, all real:**

1. **Working directory.** The resolver read the stored value raw while launch resolves a blank one to `default_agent_working_dir` and expands `~`. For default-workdir agents — the common case per `SPEC_FIX_PERSONAL_MEMORY_EMPTY_WORKDIR_2026_09_01.md` — it reported nothing at all. `effective_working_dir` reproduces both steps, and writing its test showed the blank case needs the expansion too, since the default is itself a tilde path.
2. **Side file gating.** Right that a `CLAUDE.md` write target does not mean the side file is read. Inclusion now tests for the actual `@` line, so a managed `CLAUDE.md` or a hand-removed import no longer reports stale content as instructions being consumed.
3. **Symlink guard.** The sharpest of the four: `metadata()` follows the link, so `is_symlink()` was always false and **the guard never fired at all**. Now `file_type()`, which does not follow, and symlinks are skipped outright inside a scan.
4. **Aggregate cap.** Correct that per-file and per-count ceilings do not bound the payload between them. Past an 8 MiB budget a file is still listed with hash and metadata; only content is dropped.

The side-file ownership finding you both raised was already fixed in `654848a5d`.

`cargo test -p agentmux-srv`: 3255 passed, 0 failed.

AgentA-asaf:
Fixed in `67a922386` — and thank you for catching that I missed it: this was Codex's sixth finding on this PR and I addressed only five, so the re-review is what actually caught the gap.

The substance is right. The ceiling was applied *after* `scan_dir` had recursively collected and sorted every match, so it did nothing about the CPU and memory cost it was introduced to bound. The walk now stops as soon as it holds one file more than the cap, which is exactly enough to know the limit was hit, plus a directory budget for a deep or wide tree containing few matches — the file cap alone cannot bound that, since the walk still has to enter every directory to find out.

One consequence worth naming: stopping early is only honest because entries are now sorted at each level *before* descending. Without that the result is whichever prefix the filesystem handed back first, and two calls would disagree. That is covered by `an_overflowing_scan_stops_early_and_still_returns_the_same_set`.

The overflow message also no longer claims an exact total, because counting the remainder would mean doing the traversal the cap exists to avoid.

AgentA-asaf:
P0 fixed in `48ea408a4`. You were right to raise it twice — I fixed the traversal bound and did not go back to the root check, which is a real escape and the more serious of the two.

The hole: guarding child entries does nothing when the scan directory *itself* is a link. `read_dir` follows it transparently, the paths still `strip_prefix` cleanly, and the outside file's content and hash go out over the authenticated RPC — contradicting this PR's own stated guarantee. My symlink test linked a *child*, so it could never have caught the root. That is the part worth keeping: the test was shaped like the fix I had written rather than like the property I claimed.

Fixed by canonicalizing both sides and requiring the scan root to live under the working directory. That closes the class rather than the reported shape — a symlinked final component, a symlinked ancestor such as `.github`, and `..` traversal all fail the same check.

A refused directory is now **reported** as an entry with an explicit error rather than silently omitted. A repository that has pointed its instruction directory out of the project is exactly what an operator should see.

New test `a_symlinked_scan_root_is_refused_and_reported` covers the root case specifically, alongside the existing nested one.

`cargo test -p agentmux-srv`: green, 20 project-instruction tests.

PR #3157, opened 2026-09-10
fix(term): stop the terminal pane header/tab-strip flash on tab switch

GenericAgentX-asaf:
## Summary

Applies the agent pane's chrome-hoisting fix (`SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md`, #3136/#3151) to the terminal pane, which was deliberately deferred there and has the same bug: switching shell tabs rebuilt the pane header and the in-pane tab strip along with the terminal.

- **`term.tsx`** — `TerminalView`'s tab-strip half is extracted into a new `TermPaneChrome`, rendered via `renderPaneChrome` so it lives outside the per-block remount boundary and stays mounted across switches. Registered through the same late-binding hook the view component already uses to break the `term.tsx` ↔ `termViewModel.ts` cycle.
- **Reveal gates dropped** from the three stack mutators, same reasoning as the agent pane: they existed because the whole leaf used to remount, and keeping them now *causes* a flash rather than hiding one.
- **Node resolution by `nodeModel.nodeId`**, not the anchor blockId — the correction ReAgent caught on the agent pane, applied up front here.
- **`pane-leaf-chrome.tsx`** — `"term"` joins `"agent"` in a named `HOISTS_OWN_CHROME` set, with the invariant (a listed type must drive `noHeader` off `paneChromeHoisted`) written down beside it.

## The one real difference from the agent pane

`TermViewModel` sets `manageConnection` to `!isCmd` — **true for every ordinary terminal** — so `BlockFrame_Header` really does render the connection button here. The inert stand-ins `AgentPaneChrome` gets away with (agent never sets `manageConnection`) would have left a dead button on every terminal.

Both the modal-open atom and the button ref now resolve the **same per-block objects** `BlockFrame_Default_Component` uses (block-atom cache, keyed on blockId), so the hoisted button still opens *and* anchors the `ChangeConnectionBlockModal` that stays inside `BlockFrame`. `blockframe.tsx`'s `connBtnRef` moves from a plain local to that shared holder for the same reason.

## Test plan

- [x] `npx vitest run` — 3901 passing / 3 skipped.
- [x] `npx stylelint` on `term.scss` — clean.
- [x] New `pane-leaf-chrome` test asserting `"term"` hoists; updated the passthrough test to use `"editor"` (it was using `"term"` as its example of a non-hoisting type, which this change invalidates).
- [ ] **Not visually verified.** Needs a live check. The terminal's connection button is the specific thing to exercise: open a remote terminal, click the connection button in the header, confirm the switcher modal both **opens** and **anchors to the button**. Also worth checking `term.scss`'s relocated strip band (glass/blur) still reads right in its new parent.

### DIFF
--- a/.changesets/1789029661-fix-term-stop-the-terminal-pane-header-tab-strip-flash-on-sh-23ty.md
+++ b/.changesets/1789029661-fix-term-stop-the-terminal-pane-header-tab-strip-flash-on-sh-23ty.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): stop the terminal pane header/tab-strip flash on shell-tab switch
--- a/frontend/app/block/blockframe.tsx
+++ b/frontend/app/block/blockframe.tsx
@@ -803,7 +803,18 @@ function BlockFrame_Default_Component(props: BlockFrameProps): JSX.Element {
     const magnifiedBlockBlur = () => magnifiedBlockBlurAtom();
     const magnifiedBlockOpacityAtom = getSettingsKeyAtom("window:magnifiedblockopacity");
     const magnifiedBlockOpacity = () => magnifiedBlockOpacityAtom();
-    let connBtnRef: { current: HTMLDivElement | null } = { current: null };
+    // Shared per-block holder rather than a plain local: when a pane's
+    // header is HOISTED out of this component (pane-leaf-chrome.tsx —
+    // `nodeModel.paneChromeHoisted`), the header, and therefore the
+    // connection button whose element this captures, renders in a different
+    // component tree, but the ChangeConnectionBlockModal it anchors stays
+    // HERE. A local ref would then never be populated and the modal would
+    // lose its anchor. Keyed on blockId through the block-atom cache, so
+    // both sides resolve the same object and it's cleaned up with the block.
+    const connBtnRef = useBlockAtom(nodeModel.blockId, "connBtnRef", () => {
+        const holder: { current: HTMLDivElement | null } = { current: null };
+        return () => holder;
+    })();
     const noHeader = util.useAtomValueSafe(props.viewModel?.noHeader);
     // Captured outer-frame ref for PaneSizeBadge. Live as long as the
     // frame is mounted; cleared on unmount via the callback ref.
--- a/frontend/app/block/blockframe.tsx
+++ b/frontend/app/block/blockframe.tsx
@@ -328,11 +328,20 @@ function BlockFrame_Header(
     // live data with no leak (frontend/app/store/wos.ts).
     const blockData = createMemo(() => WOS.getWaveObjectAtom<Block>(WOS.makeORef("block", props.blockId()))());
     const showBlockIds = getSettingsKeyAtom("blockheader:showblockids")();
-    const preIconButton = util.useAtomValueSafe(props.viewModel?.preIconButton);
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

GenericAgentX-asaf:
All three fixed in 05d1d74c8 — and the P1 was a genuine crash, good catch.

**1. Connection selector threw on tab switch.** Exactly right about the mechanism: `ConnectionButton` used a plain callback ref, which fires once at element creation. With a hoisted header the element is never recreated across a switch (the `<Show>` stays true when both tabs have connections), so the newly-active block's holder kept `current: null` and `TypeAheadModal`'s unconditional `anchorRef.current.getBoundingClientRect()` blew up. Replaced with an effect that re-runs when `props.ref` changes identity. Worth noting this was a latent bug in `ConnectionButton` generally — hoisting is just what made it reachable.

**2 + 3. Runtime badge and terminal background.** Same root cause: both are overlays that resolved against `.view-term` back when it contained the strip. Fixed together by giving chrome a `.term-pane-stack-body` wrapper spanning the strip **and** the terminal, and moving both into it — restoring the exact stacking they had before. It deliberately excludes the header row, since the background never painted behind the pane header pre-hoist either. Both now read the active member, so they follow tab switches.

This sharpens the live-check target I'd flagged: the specific path to exercise is **two remote terminal tabs, switch between them, then click the connection button** — that's what threw.

GenericAgentX-asaf:
Fixed in 6b073327e — and this one is a good illustration of why the terminal wasn't a copy-paste of the agent pane.

`preIconButton` and `manageConnection` were read at the top level of `BlockFrame_Header`, so both snapshotted whichever ViewModel was active at first mount. Invisible while `AgentPaneChrome` was the only hoisted caller (agent never sets `manageConnection`, and its `preIconButton` is a constant `null`), but `TermViewModel.manageConnection` is `!isCmdController()` — genuinely different between a cmd-controller terminal and an ordinary one in the same stack. The `<Show>` gate would have been pinned to whichever tab mounted the header first.

Both are memos now, along with `preIconButtonElem` since it consumed the snapshot.

Also dropped the adjacent `connName`/`connStatus` locals while in there — dead in this component (`ConnStatusOverlay` computes its own further down), and the dead `connStatus` read was subscribing every header to a connection-status atom for nothing.

Adds one more case to this PR's live check: a stack holding **both a cmd-controller terminal and an ordinary one** — the connection button should appear/disappear as you switch between them.

GenericAgentX-asaf:

GenericAgentX-asaf:
Both fixed in 18db18800 — both were leftovers from this PR's own refactoring.

- **Dead `termBg` memo**: right, `TerminalView` kept declaring it after the background layer moved to `TermPaneChrome`, so it recomputed and re-subscribed to `blockData()` on every terminal render with nothing reading it.
- **Stale stacking comment**: also right, and the reason is that the comment predates the fix that made it wrong — it was written when I first relocated the strip, and the `.term-pane-stack-body` wrapper only arrived a commit later with the badge/background fix. Corrected to describe the actual arrangement (siblings under the body wrapper) and why that's specifically what lets `backdrop-filter` still blur the image.

One I deliberately did **not** fix: `term.tsx`'s `invokeCommand` import is unused, but that predates this branch — confirmed with `tsc --noUnusedLocals` against a clean tree, so it isn't fallout from the extraction and doesn't belong in this diff.

GenericAgentX-asaf:

PR #3158, opened 2026-09-10
Manoz@Area54: fix(agent-pane): background shell output no longer pins a turn in Working

GenericAgentX-asaf:
## The stall, from a real session

Block `f861ac9c`, 2026-09-09. Backend `turn_active` flips:

```
23:11:38  active=true
23:12:30  active=false     ← turn genuinely ends
   ... 5h 00m of silence ...
04:13:03  active=true      ← next user message
```

The backend was right for five hours. The pane showed "Working…" throughout and gated input behind <kbd>Esc</kbd>.

Frontend `[wave-turn]` for the same window:

```
23:12:30  Idle → Done       cmd=TurnEnd
23:13:01  Done → Streaming  cmd=StreamFlushObserved toolsActive=0 currentTool=-
   ... no further transitions ...
watchdog: no recovery — idleSinceMs=48624 thresholdMs=45000     (every 2 min, for 5h)
```

Third recurrence of this shape — see `retro-persistent-agent-working-status-stuck-2026-07-16` (12h) and `REPORT_AGENTA_STUCK_WORKING_INVESTIGATION_2026_08_14`.

### 1. Background shell output counted as turn liveness — the load-bearing one

`flushPendingNodes` batches **six** independent queues and dispatches one `StreamFlushObserved` if *any* had content. The reducer bumps `lastEventMs` on every one.

So a live **background** shell writing to its dock row refreshed the *turn's* idle clock on every flush. That's why `idleSinceMs` reads 48624, 50003, 49374 at successive ticks instead of climbing — it was reset every ~50s by `task dev` output, never reached `LIVENESS_RECOVERY_MS` (180s), and the pane stayed pinned for exactly as long as that process lived.

Shell **chunks** and **exits** no longer mark a flush turn-relevant. Nodes, updates, tool chunks and shell **creates** still do — the agent opening a shell is turn activity; that shell's later output is not.

The module's own doc comment already asserted *"no empty StreamFlushObserved is ever dispatched"*. True only for the all-six-empty case. A shell-output-only flush is empty **as far as the turn is concerned** and still dispatched. That gap is the bug.

**This fix alone would have prevented the stall.**

### 2. A zero-node flush resurrected a completed turn

`Done{completed}` → `Streaming` is deliberate for multi-round continuations (#2420) — but those always open with a new document node. Promotion now requires `addedCount > 0`. In the trace above it fired 31s after `TurnEnd` with `addedCount=0`, `toolsActive=0`, no current tool.

### 3. The diagnostic reported the wrong bar

`stream-stuck` carried only `STUCK_THRESHOLD_MS` (45s — a cosmetic warning) where any reader assumes the recovery bar (180s). So the log reads *"idle exceeded the threshold and it still didn't recover"*.

**That misled me for most of a day** — I told the repo owner the watchdog had "agreed the pane was idle and refused to recover", when in fact it had never come within 130s of its bar. It now carries both, and logs `warnAtMs` / `recoverAtMs`.

## Compatibility

`turnRelevant` is optional and **absent means true**, so every other dispatcher and all pre-existing tests keep their exact behaviour. Only the flush queue — the one caller that can tell the difference — sets it.

## Tests

7 new, incl. `lets the watchdog recover a turn that only background output kept alive`, which replays the ~50s flush cadence across the full recovery window and asserts recovery now fires. Also pins that a real next round (which adds a node) still promotes, so #2420 isn't regressed.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run` over `store/agent-pane-state/`, `store/agent-document/`, `view/agent/` — **2135 passed, 0 failed** (136 files)

## Context

P2 of `docs/reports/REPORT_AGENT_PANE_PROGRESS_INDICATORS_CONSOLIDATION_2026_09_09.md`. P1 (one predicate for both indicators) merged in #3143. P3 (ambient Haiku narration of a backgrounded task) is separate and still to come.

Note the report's other P2 lead — `ScrubOrphanedInProgress` having zero dispatch sites — is **not** what bit here (`toolsActive=0` throughout). Still real debt, but a different bug; left alone deliberately.

### DIFF
--- a/.changesets/1789029739-fix-agent-pane-background-shell-output-no-longer-pins-a-fini-2je4.md
+++ b/.changesets/1789029739-fix-agent-pane-background-shell-output-no-longer-pins-a-fini-2je4.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): background shell output no longer pins a finished turn in Working
--- a/frontend/app/store/agent-pane-state-store.ts
+++ b/frontend/app/store/agent-pane-state-store.ts
@@ -351,7 +351,7 @@ export function dispatch(
                 console.info(
                     "[wave-turn]",
                     `pane=${blockId.slice(0, 7)}`,
-                    `watchdog: no recovery — idleSinceMs=${ev.idleSinceMs} thresholdMs=${ev.thresholdMs}`,
+                    `watchdog: no recovery — idleSinceMs=${ev.idleSinceMs} warnAtMs=${ev.thresholdMs} recoverAtMs=${ev.recoverThresholdMs}`,
                     exempt ? `EXEMPT toolsActive=${p.toolsActive} currentTool=${slot.state.currentTool ?? "?"}` : "",
                 );
             }
--- a/frontend/app/store/agent-pane-state/reducer.ts
+++ b/frontend/app/store/agent-pane-state/reducer.ts
@@ -228,6 +228,24 @@ export function update(
             if (state.lastEventMs == null) {
                 return { state, events: [] };
             }
+            // FIX 1 — shell OUTPUT alone is not evidence this turn is alive.
+            // The flush queue batches six independent queues and dispatches one
+            // StreamFlushObserved if ANY of them had content. A live background
+            // shell writing to its dock row therefore refreshed lastEventMs on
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

PR #3159, opened 2026-09-10
fix(ci): rename the WinGet identifier to AgentMux.AI

Agent3-asaf:
## Summary

- WinGet release automation has been **disabled since 2026-08-12** (#2549): `wingetcreate update` requires an existing manifest in `microsoft/winget-pkgs`, which was never bootstrapped — it failed identically on every release back through at least v0.54.14.
- Before doing that one-time bootstrap (tracked as the actual "deploy" step, done separately against `microsoft/winget-pkgs`, not this repo), rename the planned identifier per repo owner's decision: `AgentMux.AgentMux` → **`AgentMux.AI`**.
- Safe to rename freely: nothing has ever been published under the WinGet identifier. This is **not** the same identity as the already-live MSIX/Microsoft Store package (`AgentMux.AgentMux`, Store ID `9P9QCXNNCRK3`, PFN `AgentMux.AgentMux_vqr1k32tkfk4y`) — that one stays untouched, since changing it would break the update chain for existing Store installs. See the new spec for the full distinction.

## Changes

- `Taskfile.yml`: `WINGET_PACKAGE: AgentMux.AgentMux` → `AgentMux.AI`, with a comment pointing at the spec so the two identities don't get conflated later.
- `.github/workflows/release.yml`: corrected the identifier inside the **commented-out** `winget` job (still disabled — `wingetcreate update` needs the manifest to exist first; re-enabling is a follow-up once the bootstrap PR merges upstream).
- `docs/specs/SPEC_WINGET_PACKAGE_BOOTSTRAP_2026_09_10.md`: records the decision, the two-identity distinction, and the bootstrap procedure (installer is Inno Setup → native WinGet `InstallerType: inno`; license `Apache-2.0` from `package.json`; `WINGET_TOKEN` repo secret already exists).
- `docs/specs/INDEX.md`: regenerated (`bash scripts/gen-docs-index.sh`) to include the new spec.

## Test plan

- [x] `task --list` parses `Taskfile.yml` cleanly after the edit.
- [x] `grep -rn "AgentMux\.AgentMux"` confirms only the MSIX-specific files/comments still reference the old identity — nothing else needed updating.
- [x] `gh api repos/agentmuxai/agentmux/actions/secrets` confirms `WINGET_TOKEN` already exists (not re-verified for validity/scope — the actual bootstrap submission is the real test of that).
- [ ] Not yet done: the actual `wingetcreate new AgentMux.AI ...` bootstrap PR against `microsoft/winget-pkgs` — that's the follow-up "deploy" step once this merges, tracked in the spec's §4.

### DIFF
--- a/.changesets/1789029506-fix-ci-bootstrap-the-winget-identifier-as-agentmux-ai-distin-9jo9.md
+++ b/.changesets/1789029506-fix-ci-bootstrap-the-winget-identifier-as-agentmux-ai-distin-9jo9.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): bootstrap the WinGet identifier as AgentMux.AI, distinct from the published MSIX identity
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -379,6 +379,12 @@ jobs:
   # ── Phase 4: Distribution channels — DISABLED 2026-08-12 ──────────────────
   # MS Store is uploaded manually; WinGet automation never worked (see header
   # comment above). Both job bodies kept for reference, fully commented out.
+  # WinGet identifier corrected to AgentMux.AI 2026-09-10 (distinct from the
+  # already-published MSIX/Store identity AgentMux.AgentMux — see
+  # docs/specs/SPEC_WINGET_PACKAGE_BOOTSTRAP_2026_09_10.md) so this job is
+  # ready to uncomment once that spec's bootstrap PR merges into
+  # microsoft/winget-pkgs. Still commented out here: `wingetcreate update`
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

Agent3-asaf:
Good catch — confirmed against `wingetcreate.exe new --help` (v1.12.13.0) and a live dry run: `new` only takes installer URL(s) positionally, no `--version`/`--urls`/`--submit`. Fixed in 534834b — `new` now gets the installer URL positionally and the doc calls out that PackageIdentifier/Publisher/etc. are answered at the interactive prompts `new` throws afterward, not passed as flags.

PR #3162, opened 2026-09-10
feat(srv): Phase 3 — track project instructions across launches

AgentA-asaf:
Phase 3 of `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md`, the tracking half. Builds on #3156, which answered *what an agent reads right now*. This records those answers so the next one can be compared against the last — which is what makes a repository's own `CLAUDE.md` changing under an agent **visible** rather than merely present. Tracking: #3148.

**Schema v33** adds `db_agent_project_instructions`, keyed `(agent_id, path)`, holding a content hash, size, owner and observation time.

**Hash and metadata only, never content.** `db_agent_native_memory` mirrors content because it is a durable copy of files that can become unreachable. These files belong to the repository, are always readable from it, and are never written by AgentMux — a second copy of somebody else's file, with nothing keeping it honest, would be a liability rather than a feature. The distinction is stated in the schema doc so the next person does not "fix" the asymmetry.

**Observation happens at `agent.open`**, after AgentMux's own files reach their final state for that launch. Observing earlier would record the *previous* launch's content. Best-effort throughout: an agent must never fail to launch because a tracking row could not be written.

That is the spec's §8 decision 2 — it costs nothing, catches the case that matters (the repository changed since last launch), and avoids a watcher. **Drift during a session remains uncovered**, deliberately; the RPC therefore reports `modified` on every call until the agent is next opened, which is stated at the call site rather than left to surprise someone.

**`classify_change` is pure** — a function over `(previous, exists, hash)` — so every transition is unit-testable without a database.

| Transition | Result |
|---|---|
| No prior observation | `first_seen` |
| Same hash | `unchanged` |
| Different hash | `modified` |
| Absent → present | `added` |
| Present → absent | `removed` |
| Hash unavailable either side | `unchanged` |

That last row is the one worth arguing about. The resolver leaves the hash empty when it cannot read the bytes, so guessing `modified` would cry wolf on a permissions change, while calling it unchanged hides nothing — the resolver already reports the error alongside.

`agent.project_instructions` now returns each file's `change` and `last_observed_at` beside its current state, and **stays a pure read**: writes happen at launch.

**Tests** — 6 new. Store-level: replace-not-duplicate, per-agent scoping, forget-one-leaves-others. Pure: every transition, and the unreadable case. Plus an end-to-end test that writes a `CLAUDE.md`, observes, edits it, and asserts the edit is detectable and then settles.

**Verification**
```
cargo test -p agentmux-srv   → 3263 passed, 0 failed
```

Next: export carrying `components.projectInstructions` with the owner discriminator, surfaced on import and never applied.

### DIFF
--- a/.changesets/1789031080-feat-srv-track-what-each-agent-reads-as-project-instructions-rfl0.md
+++ b/.changesets/1789031080-feat-srv-track-what-each-agent-reads-as-project-instructions-rfl0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): track what each agent reads as project instructions, so a repository file changing between launches is detectable
--- a/agentmux-srv/src/backend/storage/agents.rs
+++ b/agentmux-srv/src/backend/storage/agents.rs
@@ -511,6 +511,11 @@ impl Store {
     }
 
     /// Delete all seeded agents (is_seeded=1). Used by reseed to clear built-in agents.
+    ///
+    /// No project-instruction cleanup here, unlike `agent_def_delete`: this
+    /// removes template rows, and a template is never launched — observations
+    /// are recorded by `agent.open`, so a template has none to forget (Codex,
+    /// PR #3162).
     pub fn agent_def_delete_seeded(&self) -> Result<usize, StoreError> {
         // Consolidation Phase 3b (PR 2): only the template rows go. Every
         // `is_template = 0` row — a user clone OR an agent launched from a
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

AgentA-asaf:
Both findings fixed in `e6a8ab399`.

**Deletion cleanup** (raised by both reviewers). Correct, and the kind of gap worth naming precisely: I wrote a function documented "for agent deletion" and then never called it from deletion, so the doc comment described an intention rather than the code. Now called from `agent_def_delete`, beside the row these observations belong to, rather than left to each caller to remember.

The consequence you both identified is the real one: with no foreign key — these record files an agent *reads*, which is not a relationship SQLite can express — an agent reusing a stable id inherits somebody else's baseline, with every file reporting `unchanged` against observations never made about it. That is worse than no tracking.

On `agent_def_delete_seeded`: deliberately no cleanup there, and now documented as such rather than left looking like the same oversight twice. It removes template rows, and a template is never launched, so it accumulates no observations — `observe_project_instructions` runs from `agent.open`.

**Scanned file removal** (Codex). Also correct. The response iterated current files only, so a scanned `.github/instructions/*.instructions.md` deleted since the last launch simply disappeared instead of reporting `removed`, and the stale row meant recreating it would then read `unchanged` — the drift hidden in exactly the direction tracking exists to expose. The response now covers the union of current files and prior observations. Declared paths always come back, absent ones included, so this only fires for dynamically scanned files.

`cargo test -p agentmux-srv`: 3265 passed, 0 failed.

PR #2333, opened 2026-07-28
feat(armory): add bundle.export RPC — ABF Phase 1 exporter

AgentX-asaf:
## Summary
Re-opens #2325, which GitHub auto-closed when its base branch
(`agentx/abf-phase0-agent-skills-format`) was deleted after Phase 0 (#2322)
merged. Same branch (`agentx/abf-phase1-exporter`), rebased cleanly onto
`main` (no conflicts — Phase 0's content is already there via #2322's squash
merge). No code changes beyond the rebase itself.

Implements **Phase 1** of the rollout plan in
`docs/specs/REPORT_ARMORY_BUNDLE_STANDARD_RESEARCH_2026_07_16.md`: serialize
a bundle + its referenced skills/MCP servers into the ABF v0.1 on-disk
layout via a new `bundle.export` RPC.

## Review history carried over from #2325
Both reagent and Codex review rounds already ran on this branch and their
findings were fixed:
- **Security**: `env`/`headers` secret values are redacted (`${VAR_NAME}`
  placeholder) before writing `mcp/*.server.json`.
- Manifest references the skill's directory (`skills/<slug>`), not the
  `SKILL.md` file, per the ABF spec's own example.
- Skill-lookup DB errors now fail the export loudly instead of being
  swallowed; genuinely-deleted skills are reported via `missing_skill_ids`.
- `bundle.export` registered in the RPC contract test's known-undeclared
  list.
- `accounts/requirements.json` now infers from `headers` as well as `env`,
  so header-only-authenticated servers get a matching requirement entry.
- Deliberately NOT attempted: converting the exported MCP config to the
  official MCP registry `server.json` schema — needs fields (registry type,
  package identifier) this data doesn't have for an arbitrary stdio
  command; documented as an accepted Phase 2 follow-up in the code.

## this push)
- Malformed `context_files`/`mcp_servers` JSON silently becomes an empty
  component via `unwrap_or_default()`, no error surfaced.
- Manifest `$schema` host mismatch (`docs.agentmux.ai` vs spec's
  `agentmux.ai`).
- Context-file entries whose sanitized paths normalize to the same value
  aren't collision-detected before being pushed into `files`.

## Test plan
- [x] `cargo test -p agentmux-srv`: 1711 passed on the rebased branch.
- [x] `cargo build -p agentmux-srv` clean.
- [ ] Manual: call `bundle.export` for a real bundle with skills + MCP
      servers bound, inspect the returned file list / decoded zip.

PR #3163, opened 2026-09-10
feat(srv): Phase 3 — carry project instructions in the export, never install them

AgentA-asaf:
The last piece of Phase 3 of `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md`, and where the original ask lands: an exported bundle now records what its agent was actually reading. Builds on #3156 and #3162. Tracking: #3148.

**Export.** Files land under `instructions/project/<sanitized-path>`; each `components.projectInstructions` entry carries the original `path`, a `contentHash`, and an `owner`.

**`owner` is the load-bearing field.** It is what stops this becoming a way to install one repository's instructions into another. An importer treating a `foreign` entry as content to write would be doing precisely what `SPEC_CLAUDE_MD_OWNERSHIP_PROTECTION_2026_08_22.md` exists to prevent — the whole reason a foreign `CLAUDE.md` is never overwritten in the first place.

**Import surfaces them and never applies them** (spec §5.2). An operator who wants them acts deliberately.

That warning has **no filter site anywhere**, unlike `MEMORY_COMPONENT_IGNORED_WARNING`. Memory warns only on the agent-less path because `bundle.import_for_agent` genuinely does install it; no import path installs these. So the warning is unconditional and permanent rather than a "use the other RPC" pointer — stated at the constant, because the asymmetry with memory looks like an oversight otherwise.

**Scope, stated so absence is not misread:** only files that exist and were readable are carried, since there is nothing to put in an archive otherwise. Their absence from a manifest is therefore *not* a claim the source agent had none. `agent.project_instructions` remains where the fuller picture lives, including absent, unreadable, oversized, and removed-since-last-launch files.

The working directory is resolved the way launch resolves it, so the export describes the directory the agent actually runs in — the same correction #3156 needed.

`set_manifest_component` now takes JSON values rather than paths. Most components are a list of archive paths; a bare path could not express the owner discriminator. The two existing callers map into it.

**Tests** — 2 new:

| Test | Pins |
|---|---|
| `export_carries_project_instructions_with_their_owner` | the repo's own `CLAUDE.md` is carried, with content, hash, and `owner: foreign` |
| `project_instructions_are_surfaced_and_never_installed` | the warning fires, and the content reaches neither the bundle's instructions nor its context files |

**Verification**
```
cargo test -p agentmux-srv   → 3267 passed, 0 failed
```

With this, Phase 3 is complete: AgentMux can say where every instruction an agent reads came from, what is in it now, whether it changed since last launch, and can move it to another machine without ever writing somebody else's file.

### DIFF
--- a/.changesets/1789033084-feat-srv-exported-bundles-carry-the-project-instructions-the-5y54.md
+++ b/.changesets/1789033084-feat-srv-exported-bundles-carry-the-project-instructions-the-5y54.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): exported bundles carry the project instructions their agent was reading, recorded read-only and never installed on import
--- a/agentmux-srv/src/backend/bundle_import.rs
+++ b/agentmux-srv/src/backend/bundle_import.rs
@@ -1007,29 +1007,47 @@ pub fn parse_bundle_import_with_budget(
                     let str_field = |k: &str| {
                         obj.get(k).and_then(|v| v.as_str()).unwrap_or_default().to_string()
                     };
-                    let file_path = str_field("file");
-                    if file_path.is_empty() {
+                    let raw_file = str_field("file");
+                    if raw_file.is_empty() {
                         warnings.push(
                             "components.projectInstructions: entry has no \"file\"; skipped"
                                 .to_string(),
                         );
                         continue;
                     }
+                    // Normalize the manifest's OWN reference the same way
+                    // `by_path`'s keys are normalized, exactly as every other
+                    // manifest-reference lookup in this file does (codex P2,
+                    // PR #2379 round 4; reagent P2, round 6; reagent P2 again
+                    // on PR #3163 — this is the third component to need it).
+                    // Without it a valid non-canonical spelling
+                    // (`./instructions/project/CLAUDE.md`, backslashes) misses
+                    // the lookup and gets reported as "not found" even though
+                    // the content is right there.
+                    let Some(file_path) = sanitize_context_relative_path(&raw_file) else {
+                        warnings.push(format!(
+                            "components.projectInstructions: \"{raw_file}\" is not a \
+                             safe path; skipped"
+                        ));
+                        continue;
+                    };
                     // The accounts/ allowlist applies here for the same reason
                     // it applies to components.instructions: a manifest is not
                     // a trustworthy inventory of its own bundle, and a
                     // projectInstructions entry must not become a second way to
                     // read accounts/requirements.json back out as content.
+                    // Checked on the NORMALIZED path so a non-canonical
+                    // spelling can't slip past it either.
                     if is_requirements_json(&file_path) {
                         warnings.push(format!(
-                            "components.projectInstructions: \"{file_path}\" is the \
+                            "components.projectInstructions: \"{raw_file}\" is the \
                              accounts/ requirements file; not readable as instructions"
                         ));
                         continue;
                     }
                     let Some(content) = by_path.get(file_path.as_str()) else {
                         warnings.push(format!(
-                            "components.projectInstructions: \"{file_path}\" not found \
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentA-asaf:
All three Codex findings addressed in `fb5e1a2ee`.

**P1 — `bundle_import.rs:934`, preserve imported project-instruction metadata.** Agreed, and this was the more serious of the three: the point you raise about `agent.project_instructions` is exactly right — it scans the *importing* agent's working directory, so it can never show what the source machine held. That made the warning the only signal, and "surfaced, never installed" only half true.

The parser now reads the entries back out into `ParsedBundleImport.project_instructions` (`path`, `file`, `contentHash`, `owner`, `content`), and `bundle.import.preview` returns them with a bounded per-entry preview (`MAX_PROJECT_INSTRUCTION_PREVIEW_CHARS = 4_000` — deliberately well under the 50k single-field instructions cap, since a preview can carry up to 200 of these).

Reading is not applying, and that separation is enforced structurally, not by convention: nothing writes them, and `bundle.import.commit` has no field that could select them. The `PROJECT_INSTRUCTIONS_NOT_APPLIED_WARNING` still rides alongside, still with no filter site. Hardening that came with it — entries capped at 200 (`MAX_IMPORTED_PROJECT_INSTRUCTIONS`, matching the exporter's own `MAX_SCANNED_INSTRUCTION_FILES` so it can only bite a hand-written bundle) with a warning on overflow; the accounts/ allowlist applied so this can't become a second route to read `accounts/requirements.json` back out as content; a dangling `file` reference warns instead of vanishing.

**P2 — `bundle.rs:927`, export readable empty instruction files.** Correct, and the diagnosis is precise: `content.is_empty()` was doing two jobs and only one of them right. A readable zero-byte `CLAUDE.md` is a fact about the source agent, and dropping it made the export indistinguishable from one where the file was absent — the exact ambiguity this component exists to remove. The predicate is now `!exists || error.is_some() || truncated`, which are the conditions that genuinely leave nothing to archive.

**P2 — `bundle.rs:934`, reject colliding project-instruction output paths.** Correct, and worth noting *why* it's newly reachable here: the fixed registry paths can't collide, but the scanned directories (Copilot's `.github/instructions/`, recursive) can. Now mirrors the case-folded `used_context_paths` check `bundle_export.rs` already applies to context files, warning on skip rather than overwriting silently. Sanitizer rejections now warn too — same reason, and the same finding you and reagent both raised on #2333.

Spec §5.2 updated to say "surfaced" means *readable*, not merely *announced*, with the reason.

7 new tests (3274 passed / 0 failed). The two collision/unreadable cases are direct unit tests on `splice_project_instructions_component` rather than filesystem-staged — a case-only collision can't be staged at all on this repo's own dev platform, so testing it through the filesystem would have meant not testing it.

AgentA-asaf:
P2 addressed in `01f330ad5`.

You're right, and the framing is the useful part: this is the **third** component in this module to need the same fix — `components.instructions` (codex P2, #2379 round 4), then `accounts/requirements.json` (your P2, #2379 round 6), now `projectInstructions`. I added a new manifest-reference lookup and didn't apply the rule the two neighbours 20 lines away already document in their own comments.

The reference now goes through `sanitize_context_relative_path` before anything touches it, and the accounts/ allowlist is checked on the **normalized** path — otherwise normalizing the reference would itself have opened `./accounts/./requirements.json` as a way around the allowlist. Warnings quote the manifest's original spelling, since that's what an operator would have to go fix.

Two tests: `a_non_canonical_project_instruction_reference_still_resolves` (backslashes + `./` segments resolve, and `file` comes back canonical) and `a_non_canonical_accounts_reference_is_still_rejected_from_project_instructions` (the allowlist survives the normalization).

Suite: 3276 passed / 0 failed.

PR #3164, opened 2026-09-10
Manoz@Area54: fix(agent-pane): fix a type error that a missing tsc gate let through

GenericAgentX-asaf:
## 1. Wire `ScrubOrphanedInProgress` — built, tested, dispatched from nowhere

`TurnEnd` cleans the **pane** reducer. The **document** reducer's `nodes[]` were left untouched, so a tool node that never received its `tool_result` stayed `status: "running"` forever — its dock row counting up with no process behind it.

`toolCallToNode` creates nodes optimistically at `running`, and the only exit is a matching `tool_result`. But a harness **pre-execution rejection** arrives as a turn-terminating top-level `result` frame instead, so the result never comes and the node never closes.

`scrubOrphanedInProgress` was written for exactly this (`SPEC_MUXSPECT_DOCK_DIAGNOSIS_AND_REMEDIATION_2026_08_06` §1.1) and then ran only on session/reload boundaries — never on a turn boundary. The standalone command had **zero production dispatch sites**; only tests referenced it.

Uses the documented `tools-only` scope. Thinking markdown, shell nodes and `awaiting_answer` legitimately outlive a turn; a full scrub at every turn boundary would cancel them.

### Why this can't kill a live background task

The obvious worry, so stating it explicitly: an **accepted** background launch is terminal `success` within ~a second — its result text *is* the acceptance message — so it is never `running` by the time a turn ends. The scrub only touches nodes with no result at all. `isAcceptedBackgroundLaunch` requires `status === "success"`, which the scrub's own `status === "running"` guard makes mutually exclusive; the document reducer's existing test suite already pins that (`reducer.test.ts:1032`).

Note this is **not** what caused the 5-hour hang fixed in #3158 — that had `toolsActive=0` throughout. Separate bug, same neighbourhood.

## 2. Fix a type error I merged in #3158

`turn-liveness-shell-output.test.ts` passed `{ type: "TurnEnd", at, outcome }`. `TurnEnd` carries only `stats`. The extra properties were ignored at runtime — the reducer derives the outcome itself — so the test was exercising the intended `Done{completed}` state and still passes. It simply did not compile.

**How it reached main:** CI has no `tsc` gate. The frontend jobs are `vitest` and the doc/grep gates, and vitest does not typecheck. I ran `tsc` before writing that test and not after, so nothing caught it.

That gap is worth closing on its own — a `tsc --noEmit` step would have caught this in seconds, and it is the second time this session that a type-level mistake survived a green build. Not doing it in this PR to keep the diff honest, but flagging it.

## Tests

Two new, on the wiring rather than the reducer (which was already covered):

- `scrubs orphaned in-progress tool nodes when the turn ends`
- `scrubs with the tools-only scope, sparing shells and awaiting_answer`

The existing `useTurnLifecycle` harness mocked only `dispatchPane`; extended with `dispatchDoc`.

## Verification

- `npx tsc --noEmit` — clean (including the fix above)
- `npx vitest run` over `view/agent/`, `store/agent-document/`, `store/agent-pane-state/` — **2137 passed, 0 failed** (136 files)

### DIFF
--- a/.changesets/1789033201-fix-agent-pane-scrub-orphaned-in-progress-tool-nodes-at-turn-nkas.md
+++ b/.changesets/1789033201-fix-agent-pane-scrub-orphaned-in-progress-tool-nodes-at-turn-nkas.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-fix(agent-pane): scrub orphaned in-progress tool nodes at turn end
--- a/.changesets/1789033201-fix-agent-pane-scrub-orphaned-in-progress-tool-nodes-at-turn-nkas.md
+++ b/.changesets/1789033201-fix-agent-pane-scrub-orphaned-in-progress-tool-nodes-at-turn-nkas.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): scrub orphaned in-progress tool nodes at turn end
--- a/.changesets/1789033627-fix-agent-pane-fix-a-type-error-that-a-missing-tsc-gate-let--d7v8.md
+++ b/.changesets/1789033627-fix-agent-pane-fix-a-type-error-that-a-missing-tsc-gate-let--d7v8.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(agent-pane): fix a type error that a missing tsc gate let through
--- a/docs/reports/REPORT_AGENT_PANE_PROGRESS_INDICATORS_CONSOLIDATION_2026_09_09.md
+++ b/docs/reports/REPORT_AGENT_PANE_PROGRESS_INDICATORS_CONSOLIDATION_2026_09_09.md
@@ -168,7 +168,11 @@ Distinct from the desync, and the strongest lead in the codebase.
 
 `finalizeTurn()` (`hooks/useTurnLifecycle.ts:73`) clears "Working…" in the *pane* reducer but never touches the *document* reducer's `nodes[]`, and never calls `scrubOrphanedInProgress`. That scrub (`frontend/app/store/agent-document/reducer.ts:53-148`) runs only on session/reload boundaries — never on a turn boundary.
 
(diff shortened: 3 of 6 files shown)


a5af:
@codex review

GenericAgentX-asaf:
You're right, and this one is entirely mine. Reverted in e9b39868.

**The scrub is not unwired.** `agent-view.tsx:1361-1372` already dispatches it on a `turnJustEnded` effect, with a 2s delay and a `workingFromPhase` guard, added 2026-08-10 for a user report about a ~1s `git status` stuck as a "running · 45m" dock row.

And my version wasn't merely redundant, it was worse: dispatching straight from `finalizeTurn` has neither protection. The 2s delay exists to absorb a tail flush still in flight; the guard exists to skip the pass when a new turn has already started. Scrubbing immediately can flip a node whose legitimate `tool_result` is still arriving to `canceled` and publish that terminal state to the dock cache — manufacturing the exact class of bug the original was written to fix. You called that precisely.

**Root cause of my error, since it's worth recording:** my grep for dispatch sites was piped through `head -10`, and ten matches in `reducer.test.ts` crowded out the single production hit. I read "zero production dispatch sites" off a truncated list. What made it stick is that `SPEC_MUXSPECT_DOCK_DIAGNOSIS_AND_REMEDIATION_2026_08_06` §1.1 genuinely does say the command is unwired — it predates the 08-10 fix — so the bad grep looked corroborated by a spec.

I've corrected the report's §5, which asserted the same false thing and is already merged to main, and repeated it in #3161.

**What survives** is the other half, and the PR is retitled for it: the type error from #3158, where a test passed fields `TurnEnd` does not declare. The props were ignored at runtime so the test was valid and passing — it just didn't compile, and CI has no `tsc` gate to notice. I have that gate staged as a follow-up, deliberately sequenced after this merges, since adding it while main still carries the error would red-line every PR.

PR #3166, opened 2026-09-10
Manoz@Area54: ci: add a tsc --noEmit gate to the frontend job

GenericAgentX-asaf:
## Why

**vitest transpiles per-file without typechecking**, so a type error compiles away and the suite goes green. Nothing in CI reads types today — the frontend jobs are `vitest` and the doc/grep gates.

Two got through that hole in a single day:

- **#3158** — a test passed `{ type: "TurnEnd", at, outcome }` when `TurnEnd` declares only `stats`. The extra properties were ignored at runtime, so the test was genuinely valid and genuinely passing *while not compiling*. Fixed in #3164.
- Earlier the same session, a `#[ts(optional)]`-shaped mistake with the identical signature: green suite, broken types.

Neither was catchable by any gate that existed. Both would have been caught here in seconds.

## Placement

Before `vitest`, in the same job — it needs the `npm ci` that step already runs, and failing on types first is a clearer signal than a wall of suite output. No build, just the checker.

## Sequencing

Deliberately landed **after #3164**. Adding this while main still carried that type error would have red-lined every open PR, including the one fixing it.

Verified against `origin/main` after #3164 merged: `npx tsc --noEmit` is clean, so this gate passes on the current tree rather than arriving already-red. That check is the whole reason it's a separate PR.

## Note on required checks

This adds a step to the existing `vitest` job rather than a new job, so it inherits that job's required-check status automatically — no branch-protection change needed. If you'd rather it be independently visible in the checks list, it should be its own job instead; say so and I'll split it.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -167,5 +167,16 @@ jobs:
           node-version: '24'
           cache: npm
       - run: npm ci
+      # Runs BEFORE vitest, and is not redundant with it: vitest transpiles
+      # per-file without typechecking, so a type error compiles away and the
+      # suite goes green. Two reached main this way (a `#[ts(optional)]`-shaped
+      # mistake, and a test passing fields the action type does not declare —
+      # ignored at runtime, so the test was still valid and still passed while
+      # not compiling). Neither was catchable by any gate that existed.
+      #
+      # Cheap: no build, just the checker, on an install the next step needs
+      # anyway.
+      - name: tsc --noEmit
+        run: npx tsc --noEmit -p tsconfig.json
       - name: vitest run
         run: npx vitest run
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -176,7 +176,12 @@ jobs:
       #
       # Cheap: no build, just the checker, on an install the next step needs
       # anyway.
+      #
+      # tsconfig.citypecheck.json, not tsconfig.json: the build config includes
+      # only frontend/**, while vitest below also runs test/contract/* and
+      # test/e2e/*. Checking the narrower set would have left the same hole in
+      # a different directory (reagent P2 on #3166).
       - name: tsc --noEmit
-        run: npx tsc --noEmit -p tsconfig.json
+        run: npx tsc --noEmit -p tsconfig.citypecheck.json
       - name: vitest run
         run: npx vitest run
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Good catch, and a slightly uncomfortable one — the gate as written reproduced the exact failure mode it exists to close, one directory over. Fixed in 3e025cd9.

`tsconfig.json`'s `include` is `["frontend/**/*"]`, but vitest also discovers and runs `test/contract/rpc-contract.test.ts` and `test/e2e/*.e2e.test.ts` — nothing excludes them and there's no separate job. A type error in any of those would still have compiled away under per-file transpilation and gone unseen. Verified both halves of your claim before changing anything: the include set really is frontend-only, and those three files really are picked up.

Fixed with `tsconfig.citypecheck.json`, which extends the build config and widens `include` to `frontend/**` + `test/**`. Deliberately a separate file rather than widening `tsconfig.json` — that one drives the build and the editor, and test sources belong in neither.

Also verified the widened set typechecks clean on main as it stands, so the gate still lands green rather than arriving already-red. That was the point of sequencing this PR after #3164 and it applies just as much to the wider net.

Worth noting the shape: this is the third time today a claim of mine has been too narrow because I checked a *subset* and reported it as the whole — a grep truncated by `head`, a log line showing one of two thresholds, and now a typecheck scoped to one of two directories vitest runs. Same failure, three costumes.

GenericAgentX-asaf:
Answered by enumeration rather than by "it passed locally" — `test/` contains exactly five files, all real TypeScript sources:

```
test/contract/rpc-contract.test.ts
test/e2e/crash-reproject.e2e.test.ts
test/e2e/harness.ts
test/e2e/window-close-baseline.e2e.test.ts
test/vitest-setup.ts
```

No fixtures, no `.d.ts` scratch, no vendored anything — a file-extension census over the whole tree returns `5 ts` and nothing else. Two of those (`harness.ts`, `vitest-setup.ts`) aren't test files but are imported by ones that are, so including them is correct rather than incidental: `vitest-setup.ts` is the configured `setupFiles` entry and runs on every vitest invocation.

`node_modules` can't leak in either — it's outside `test/`, and tsc excludes it by default regardless.

So there's no false-failure surface today. If someone later adds fixtures under `test/`, the failure mode is a loud CI error rather than a silent gap, which is the right direction for this to fail in.

Confirmed green locally on the widened set; CI on 3e025cd9 will confirm independently.

PR #3167, opened 2026-09-10
Manoz@Area54: docs(report): status-pass the nine false-Working paths

GenericAgentX-asaf:
Re-validated the 2026-07-27 catalog against `main` after #3143/#3158/#3164. **Two of nine closed, one reframed, six stand.**

## The notable one

**This report's own triggering incident is now fixed.** Its author hit *"a leftover `task dev` background bash process"* and had to reason it out by inspecting running processes — same root cause as the 5-hour stall of 2026-09-09. Background shell output was refreshing the *turn's* idle clock on every flush, so `idleSinceMs` never reached the recovery bar for as long as that process lived. Closed by #3158.

## Per-path

- **Path 7** (`StreamFlushObserved` broad re-promotion) — **closed**, verified in source at `reducer.ts:290-298`.
- **Path 9** (composer-perception mismatch) — **closed, in the opposite direction to the one proposed here.** The report read the banner as misleading *because the composer is not locked*. The indicator's actual meaning is precisely "type now and it queues", so the fix was to make the banner, the bar and the composer agree — not to soften the banner.
- **Path 1** — **reframed, not fixed.** Lit during a genuinely running tool is correct under that definition; the *false* half was the background-output case, now closed.
- Paths 2, 3, 4, 5, 6, 8 — stand.

## Scope, stated rather than implied

Paths 1, 2, 5, 7, 9 were checked against current source. Paths 3, 4, 6, 8 depend on runtime conditions (bashwrap idle-kill, a stalled rate-limit retry, an RPC/stream ordering window) and were **not reproduced** — they are marked *"open, unverified"*, which means untested here, **not** confirmed-present.

I'm being explicit about that because this session has already produced three wrong conclusions from exactly this: checking a subset and reporting it as the whole (a grep truncated by `head`, a log line showing one of two thresholds, a typecheck scoped to one of two directories). An audit is the easiest place of all to make that mistake quietly.

## Telemetry status

Sections 3.2 and 3.3 of the report were built in the interim (`SPEC_AGENT_TURN_PHASE_TIMELINE_LOGGING_2026_08_18`) and are what made the 09-09 diagnosis possible — the stall was read straight off the two logs. **3.1 (watchdog reasoning) is still not built**, and its absence cost real time: the `stream-stuck` line reported the 45s warning bar where a reader assumes the 180s recovery bar, sending the first diagnosis in the wrong direction. #3158 added `recoverAtMs`; the fuller reasoning 3.1 sketches is still worth building.

## Also

Fixes the `Status:` line, which predated the `check-doc-status` vocabulary and failed the gate the moment this file was next touched. Docs only.

### DIFF
--- a/docs/reports/REPORT_WORKING_STATE_TELEMETRY_AUDIT_2026_07_27.md
+++ b/docs/reports/REPORT_WORKING_STATE_TELEMETRY_AUDIT_2026_07_27.md
@@ -34,11 +34,19 @@ The telemetry section 3 recommends was also partly built in the interim —
 `[wave-turn]` transition lines (3.2) and the backend `[health] turn_active flip`
 (3.3) both exist now, via `SPEC_AGENT_TURN_PHASE_TIMELINE_LOGGING_2026_08_18`.
 They are what made the 09-09 diagnosis possible: the stall was read straight off
-the two logs rather than inferred. Section 3.1 (watchdog reasoning) is still not
-built, and its absence cost real time — the `stream-stuck` line reported the 45s
-warning bar where a reader assumes the 180s recovery bar, which sent the first
-diagnosis in the wrong direction entirely. #3158 added `recoverAtMs` alongside it;
-the fuller reasoning 3.1 sketches is still worth building.
+the two logs rather than inferred.
+
+**Section 3.1 is also implemented** — essentially verbatim, in
+`agent-pane-state-store.ts`: the `stream-stuck` branch, the
+`EXEMPT toolsActive=… currentTool=…` suffix, and the `watchdog: FIRED` line for
+`working-recovered`. What #3158 changed was the *threshold that line reports*, not
+its existence: it named only the 45s warning bar where a reader assumes the 180s
+recovery bar, and now carries both as `warnAtMs`/`recoverAtMs`.
+
+(An earlier revision of this section called 3.1 "still not built". That was
+wrong — codex P2 on #3167. It came from the telemetry having *misled* the 09-09
+diagnosis, and sliding from "this told me the wrong thing" to "this does not
+exist" without checking. The whole of section 3 is now built.)
 
 | # | Path | Status |
 |---|---|---|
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both correct, and I'll fix them shortly — a cargo build is running against this worktree so I'm not switching branches mid-build.

**3.1 is implemented, and I should have checked rather than inferred.** I compared the report's sketch against `agent-pane-state-store.ts` after your comment: it is implemented essentially verbatim, including the `EXEMPT toolsActive=${p.toolsActive} currentTool=${...}` branch and the `watchdog: FIRED` line for `working-recovered`. #3158 corrected the *threshold that line reports* (it named the 45s warning bar where a reader assumes the 180s recovery bar); it did not add previously-absent reasoning.

I reached "still not built" from the telemetry having *misled me* during the 09-09 diagnosis, and slid from "this told me the wrong thing" to "this doesn't exist" without looking. That's the same subset-for-whole error I've made three other times today, and leaving it in would have made a completed item falsely actionable — exactly as you say.

**The contract wording contradicts my own module doc, which is worse.** `working-indicator.ts` says the predicate is *"will not be answered immediately"* and enumerates why that is broader than queueing: launch/relogin input is rejected by the auth guard, and during a reconnect there is no process to receive it. Only the turn-in-flight case queues. I then wrote "precisely 'type now and it queues'" into this status pass — narrower than the implementation, and pointed at exactly the predicate someone might then narrow to match. You caught the same overstatement on #3143 and I fixed it there; I reintroduced it here in prose.

Both get corrected in the report, with the status-table row for path 9 restated against the real contract.

GenericAgentX-asaf:
Fixed in the push above.

**3.1** now reads as implemented, with the correction noted inline so the next reader sees why the claim was there and why it was wrong. Section 3 is complete in full — 3.1, 3.2 and 3.3 all built.

**Path 9** now states the real contract — *"a message typed now will not be answered immediately"* — spells out that only the turn-in-flight case actually queues, and carries an explicit warning against restating it as "it queues" and narrowing the predicate to match. That is the second time you have caught this exact overstatement from me, so it seemed worth writing the guard into the document rather than just fixing the sentence.

PR #3168, opened 2026-09-10
feat(srv)!: retire the inline db_bundles skills/mcp_servers columns

AgentA-asaf:
Closes out **Phase 0b** of `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md`. Follow-up to #3152 / #3153 / #3163; tracked on #3148.

## Why

`db_bundle_skills_ref` / `db_bundle_mcp_ref` and the inline `db_bundles.skills` / `.mcp_servers` JSON columns recorded the same fact in two places, and nothing kept them in sync — binding a skill in the Armory wrote a ref row while ABF export read the column, so a bundle could **run with components it did not export and export components it did not run with** (spec §3.4).

#3152/#3153 made the refs authoritative and `m0030` carried the inline data across, which left the columns written-but-never-read. Removing them is what actually prevents the divergence from recurring: a dormant column is just a writer nobody has added yet.

## Schema

`m0031_drop_bundle_inline_columns` drops from the shared and channel stores, guarded by `PRAGMA table_info`.

Two things about it are load-bearing rather than incidental:

- **It is registered immediately after `m0030`, and must stay there.** Dropping first would destroy the data `m0030` exists to carry.
- **It could not live in the schema functions' `ADD COLUMN` compat loops**, which is where the matching additions live. Those run on every `Store::open()` — *before* the migration chain reaches `m0030`. The identity store **is** handled that way, because its `db_bundles` is a never-written schema-parity copy (see `IDENTITY_STORE_SCHEMA_VERSION` v4's note) with no data to carry, so the ordering hazard doesn't exist there. That also sidesteps `MigrationContext` having no identity-store path.

The `table_info` guard is not belt-and-braces: a channel first booted *after* this ships runs the whole chain against a shared store some earlier channel already dropped, so no-op-is-success is the **normal** path. `m0030` now reads the columns through raw SQL rather than `Store::bundle_list()`, since `Bundle` no longer models them.

## ⚠️ Forward-compat consequence

All three schema versions bump — `OBJECT_SCHEMA_VERSION` 34, `SHARED_STORE_SCHEMA_VERSION` 10, `IDENTITY_STORE_SCHEMA_VERSION` 6.

`check_schema_compat` refuses to open a store stamped newer than the running binary, so **after this ships, older AgentMux builds on the same machine will refuse to open the shared store.** That is the intended trade: failing at open with a clear message beats failing later inside an `INSERT` that still names a column that is gone. Worth knowing if you keep old `local-*` per-build channels around.

## Frontend

`BundleDraft` and `gotypes.d.ts` drop the fields. Nothing rendered them — the Armory has managed both through the `skill.*` / `mcp.*` RPCs since #1943/#1946. `normalize_bundle_upsert_input` stops re-encoding the two keys; since `Bundle` has no `deny_unknown_fields`, an older client still sending them upserts successfully rather than erroring (covered by a test).

## Reviewer notes

- **The positional renumber is the risky part.** `bundle_list`, `bundle_list_global` and `bundle_get` share one `map_memory_row`, and a shift lands TEXT in a TEXT field — silent corruption, not a type error. Added `bundle_columns_map_to_the_fields_they_are_selected_as`, which gives every field a distinct value and checks all three readers.
- Neither dropped column is a PK, unique, indexed, generated, or referenced by a CHECK/view/trigger, so a plain `DROP COLUMN` is legal and no table rebuild (`m0028`'s pattern) is needed. `rusqlite` is pinned `bundled` (SQLite 3.45), so 3.35+ `DROP COLUMN` is available on every platform.
- The `bundle_export` / `bundle_validate` test shims that existed only to feed the inline column are gone; their fixtures now return the MCP entries alongside the bundle, which is the shape the real callers use.

## Tests

3284 Rust + 3907 frontend, 0 failures. `check-doc-status`, `check-spec-citations`, `check-rpc-bindings` all pass.


a5af:
@codex review

AgentA-asaf:
Codex is right, and the consequence is worse than the finding states. **Converting this to draft — it should not merge.**

## The finding, confirmed

`db_bundles` lives in `~/.agentmux/shared/store.db`, which is a **single host-global file**. `db_bundle_skills_ref` / `db_bundle_mcp_ref` are declared in `run_object_schema` — i.e. per-channel `objects.db`. `m0030` is channel-scoped: it reads the shared columns and writes refs into *that channel's* store, once per channel.

So the sequence Codex describes is real: first channel upgraded backfills its own refs, then `m0031` drops the shared columns, and every channel upgraded afterwards finds nothing to backfill. There are **26 channels** against one `store.db` on my machine alone, so this is not a corner case.

## The part that makes it structural

It isn't only *existing* channels. A **brand-new** channel starts with empty ref tables and is populated exactly one way: `m0030`, reading the shared inline columns. `bundle_skill_bind` / `bundle_mcp_bind_*` write into `self` (the channel store) and consult `id_store` only to check the bundle exists, so nothing else ever seeds them.

That means the inline columns are not a redundant second copy — **they are the only cross-channel representation of a bundle's components**, and the per-channel ref tables are a projection of them. Dropping them doesn't remove a stale duplicate; it removes the source, and permanently breaks the seeding path for every channel created from then on.

This invalidates the premise in the PR description. "The ref tables are the single source of truth" is true *within* a channel and false *across* channels — a distinction Phase 0b's §5.4 never drew.

## Related, and arguably worse

The same asymmetry means bundle components already diverge across channels today, independent of this PR: a skill bound in channel A writes a ref row only in A. Channel B ran `m0030` once at its own first boot and never sees that bind, so B launches and exports the same shared bundle without it. That is the *original* Phase 0b divergence (§3.4) reappearing along the channel axis rather than the store axis.

## What actually unblocks this

Codex named it: the component refs have to move to a shared authoritative store before the columns can go. That is not a small change — the ref tables carry foreign keys to `db_skills` / `db_mcp_servers`, which are themselves per-channel, so it reopens the "which store owns what" question rather than just relocating two tables.

Holding the PR as a draft rather than closing it: the mechanical parts (the schema/SQL/frontend removal, the `m0031` guard shape, the `map_memory_row` pin test) stay valid once the refs are shared, and the `m0030` raw-SQL change is independently correct. Raising the cross-channel divergence separately on #3148, since it is a live bug and not something this PR introduced.

Thanks — this was the right catch, and it was load-bearing.

AgentA-asaf:
Closing without merging — superseded by ongoing work, not by the review findings alone.

## Why

The review flagged a real P1 (multi-channel data loss: this migration is channel-scoped but drops columns from the *shared* store as a side effect, so the first channel to upgrade deletes data another not-yet-upgraded channel's own `m0030` backfill still needs to read). Investigating that finding further turned into `docs/specs/SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md` §3.4a, which found the problem is bigger than multi-channel: every AgentMux release gets a brand-new, empty per-version channel store (`channels/<ch>/versions/<v>/`, nothing carries `objects.db` forward), so the shared store's inline `mcp_servers`/`skills` columns are the *only* durable seed source `m0030` has, indefinitely — not redundant data safe to delete, the actual source of truth the per-channel-per-version ref tables are a projection of.

That in turn opened `docs/specs/SPEC_DURABLE_BINDINGS_2026_09_10.md` — giving bundle skill/MCP bindings a real durable home that survives both channel and version boundaries. It's not a simple table move (the ref tables carry FKs to `db_skills`/`db_mcp_servers`, which are scoped the same fragile way — "one shape, appearing six times"). That spec explicitly tracks this PR's status as **"#3168 (drafted, blocked by this)"**, and is already landing in phases (#3175, #3180, #3181 merged; Phase 2's application-layer read/write redirect is the remaining piece).

## What happens to this PR's goal

Once `SPEC_DURABLE_BINDINGS_2026_09_10.md` finishes relocating bundle components to a store that actually survives an upgrade, retiring these inline columns becomes safe and should be a small, low-risk follow-up at that point — this PR's approach and tests are a reasonable starting point to reference then, just premature today.

PR #3169, opened 2026-09-10
Manoz@Area54: feat(agent-pane): narrate backgrounded tasks into the conversation

GenericAgentX-asaf:
## What

When the harness detaches a long-running tool call, the pane goes quiet about a command that is **still running** and says nothing about why. This adds a short line explaining what was backgrounded.

P3 of `REPORT_AGENT_PANE_PROGRESS_INDICATORS_CONSOLIDATION_2026_09_09.md`, and the last open item from that report.

## A general facility, not a background-specific one

That was the ask, and it falls out cleanly: `kind` selects the prompt server-side, so narrating a *new* AgentMux action is a match arm in `narration_prompt` plus a call site — not new plumbing. Backgrounding is simply the first consumer.

## The trigger has to live in the renderer

The obvious backend hook is `COMMAND_DOCK_NODE_STATUS`, which already fires at exactly the moment a call is seen going background. It can't be used: `CommandDockNodeStatusData` carries `tool_name` but **not the command text**. The backend knows *a Bash call* was detached and cannot say *what* was — which is the entire content of the message. Only the frontend holds the `ToolNode` params.

## Correctness details worth reviewing

**Keyed on `isAcceptedBackgroundLaunch`, never the raw flag.** That's #2518's distinction: the raw `run_in_background` is true on every call that merely *requested* backgrounding, most of which the harness resolves synchronously — 11 of 17 in that issue's own session. Narrating those would be wrong *and* would buy a model call each. There's a test asserting the fast-finisher case narrates nothing.

**Deduped per `node_id`**, in the handler, before anything spawns — a node can be re-observed.

**Its own ambient purpose constant.** Two callers sharing one purpose cancel each other; the existing `_PUSHED` constant exists because of exactly that.

**Its own cross-block semaphore.** `AmbientGateway` dedupes only within a single `(entity, purpose)` key, so N panes backgrounding at once would otherwise mean N concurrent CLI spawns. The caller this is modelled on is bounded *externally*, by a semaphore in the sweep that drives it, which an RPC-driven path inherits nothing from. The 15s timeout bounds each call's duration, not how many run.

Both of those last two came out of codex's review on #3161 — this is the implementation of that plan.

**Best-effort throughout.** Every failure path is silence, and no UI state change is gated on a narration: the action already happened and the pane already reflects it.

## On presentation — the open decision

These render in the conversation flow, in the model's voice, but carry a small `ambient` tag.

The request was for them to "appear like they are coming from the model". They do — same voice, same place, no separate panel. What they aren't is *indistinguishable*, because the model didn't write them and its transcript doesn't contain them. Read back an hour later, by a user or by an agent debugging its own conversation, an unmarked line is something the model appears to have claimed with nothing to say otherwise.

The tag costs one word. If the smoother read is worth more, deleting one `<span>` removes it and nothing else changes — the comment above it says exactly that, so it stays a decision rather than a default. **Flagging it as the thing to look at first.**

Also worth your call: for *this particular* consumer a deterministic template ("Running `task dev` in the background — I'll report when it finishes") would work with no model call, no latency, and no concurrency cap at all. Haiku earns its place when the command is gnarly enough that summarising helps. The facility was asked for explicitly, so it's built; the tradeoff is real.

## Verification

- `cargo check -p agentmux-srv` — clean
- `npx tsc --noEmit -p tsconfig.citypecheck.json` — clean
- `npx vitest run frontend/app/view/agent/ frontend/app/store/` — **2601 passed, 0 failed** (162 files)
- 7 new tests: 5 on the hook (malformed payloads, whitespace-only lines, retention bound, unsubscribe), 2 on the trigger (accepted launch narrates; fast-finisher does not)

## Not verified

Not run in a live app — the on-screen appearance of the row and the actual quality of Haiku's sentence are unobserved. Worth a look before this is relied on.


a5af:
@codex review

GenericAgentX-asaf:
All three fixed in e4818263.

**Gateway key (reagent P1 / codex's first — same finding).** You're both right, and `useAmbientNarration` retaining several narrations should have tipped me off that coalescing was wrong. `admit()` cancels the prior in-flight call for a key the moment a newer generation lands — correct for the summary callers I modelled this on, where only the latest result is ever shown, and wrong here where each event is an independent fact. Two commands backgrounded moments apart in one block meant the second silently killed the first, whose Haiku call returned `cancelled` and got swallowed by the `.ok()?`. Now keyed `block:event`, with a test asserting a sibling admission leaves the first's token uncancelled.

**Stale narration (codex's second).** This one I hadn't considered at all. Fixed at both ends rather than one: the prompt now asks for a past-tense statement of fact with no promise to report back, *and* the publish path checks the registry and drops the message entirely if the task already ended. Deliberately only on a **definite** completion — an unknown id falls through and publishes, since absence of a registry row is not evidence of completion (a narration kind may have no row at all).

**The CI failure was a third thing, and it is the one worth recording.** It was not vitest — it was the `tsc` step inside that job, catching a top-level `await import` (TS1378) in my new test file. That is the gate added in #3166 finding its first real error, on the very next PR.

It reached CI because I reported that run as "tsc clean" after reading only the *tail* of the combined output, so the error scrolled past above what I looked at. Same subset-for-whole mistake as the `head -10` grep earlier today. Re-verified by exit code this time, on the full suite rather than a subset: `tsc` exit 0, `vitest` exit 0, 3914 passed / 265 files. Worth noting the pipeline trap too — `npx vitest run | tail -4` exits with *tail's* status, so a green "exit code 0" there proves nothing.

GenericAgentX-asaf:

Fresh head e4818263 with the gateway-key fix, the stale-narration guard, and the TS1378 fix. The standing state is CHANGES_REQUESTED from 5a3ff64a, and you're right that it shouldn't clear on my say-so — I haven't asked you to take the re-verification on trust.

Agreed on the follow-up too. The `tail`-based false-clean is worth writing down next to the gate rather than left as a lesson I happen to have learned, since the gate's whole purpose is catching what a contributor believes they already checked. Doing that as its own change instead of widening this one — it has nothing to do with narration.

GenericAgentX-asaf:
Both fixed in 42e02cb0 — and the P1 is a genuine duplicate-narration bug, not a tidiness point.

**Per-connection dedupe.** You're exactly right about the mechanism: `register_handlers` runs once per WebSocket *connection*, so my set reset on every reconnect — and a reconnect is precisely when duplicates arrive, because an honored truncate clears the frontend's `nodeIdSet` and re-parses the transcript, re-emitting every node as new. Nothing backstopped it: `AmbientGateway` only guards calls concurrently in flight for a key and releases as soon as the first completes. So one launch, two Haiku calls, two broadcasts — contradicting the invariant the code asserted.

Now an `AppState` field alongside `dock_snapshots` and `pending_background_pids`, which is the comparison that makes the error obvious in hindsight: those two are process-wide for the same reason and I put mine next to them without matching their lifetime.

**Unbounded growth (P2).** Also right. `NarratedEvents` keeps the most recent 512 and evicts oldest-first. A reconnect replays the current transcript, not the whole session, so forgetting the distant past costs nothing — whereas an unbounded set for a *recency* guard costs memory for the process lifetime. Three tests on the structure, including the eviction boundary and the fact that an evicted id reads as new again (the accepted trade).

Verification: `cargo check` clean, 6 Rust tests pass (3 narration + 3 narrated-events).

GenericAgentX-asaf:

Fresh head 42e02cb0 with the AppState-scoped, bounded `NarratedEvents`. The standing state is CHANGES_REQUESTED from e4818263, so it needs a review verdict rather than the reply above to clear.

PR #3170, opened 2026-09-10
docs(spec): correct Phase 0b's status — the column retirement is blocked

AgentA-asaf:
Follow-up to #3168 (drafted) and #3148. Docs only — no changeset.

I recorded Phase 0b as closed in the portability spec. It isn't, and the reason is worth writing down rather than leaving as a drafted PR nobody can explain in three months.

## New §3.4a

§3.4 found two *stores* disagreeing about a bundle's components; Phase 0b fixed that by making the ref tables authoritative. That holds **within** a channel and fails **across** them:

| Table | Store | Scope |
|---|---|---|
| `db_bundles` + inline `skills` / `mcp_servers` | `~/.agentmux/shared/store.db` | host-global, one file |
| `db_bundle_skills_ref`, `db_bundle_mcp_ref` | `<data_dir>/db/objects.db` | per-channel |

Bundles are shared. Their components are not. Two consequences:

1. **Binds don't propagate** — `bundle_skill_bind` writes into the channel store and consults the shared store only to check the bundle exists. Now that export reads refs, that is §3.4's divergence reappearing along a different axis. Live on `main` today.
2. **The inline columns are the cross-channel representation** — `m0030` seeds a channel's refs from them at that channel's first boot, and nothing else ever does. They aren't a redundant duplicate; the per-channel refs are a projection of them.

(2) is what blocks the drop: removing the columns removes the seeding source for every channel created afterwards.

Codex caught the multi-channel half on #3168; the new-channel half follows from the same asymmetry. Confirmed against 26 channels sharing one store on this machine.

The unblock reopens §5.4's "which store owns what" rather than settling it — the ref tables carry FKs to `db_skills` / `db_mcp_servers`, which are themselves per-channel — so it's flagged as needing its own spec, same as §5.3's memory-location invariant.

### DIFF
--- a/docs/specs/SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md
+++ b/docs/specs/SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md
@@ -1,6 +1,6 @@
 # Spec: Instruction and Memory Portability
 
-**Status:** active — Phases 1, 2 and 3 landed (#3147, #3162, #3163). Phase 0b landed its reconciliation (#3149, #3152, #3153) but is **not closed**: retiring the redundant inline columns turned out to be blocked by a cross-channel asymmetry (§3.4a), and Phase 4 is untouched. Tracking: #3148.
+**Status:** active — Phases 1, 2 and 3 landed (#3147, #3162, #3163). Phase 0b landed its reconciliation (#3149, #3152, #3153) but is **not closed**: retiring the redundant inline columns turned out to be blocked by a channel-and-version asymmetry (§3.4a), and Phase 4 is untouched. Tracking: #3148.
 **Date:** 2026-09-09
 **Verified against:** `a86cdee50` (code, not spec prose)
 **Follows:** `SPEC_ARMORY_NAMING_CONSOLIDATION_2026_09_09.md` §8, which deferred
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentA-asaf:
Both addressed in `59daa8ead`.

**Placement.** Correct and my mistake — I inserted §3.4a under `## 5. Design`, between 5.2 and 5.3, so the numbering ran 5.2 → 3.4a → 5.3 and the analysis read as part of the design section rather than the gaps section it belongs to. Moved to directly after §3.4, before §3.5.

**The version boundary — this one changes the conclusion, not just the framing.** I checked `DataPaths::resolve_internal` and you're right: `data_dir` sits under `channels/<ch>/versions/<v>/` for Installed and Portable runtimes (`SPEC_VERSION_ISOLATION_2026_06_01.md` §5 Phase 2), and nothing carries the previous version's `objects.db` forward. Verified on disk — `channels/stable/versions/*/data/db/objects.db` is a separate file per release, five of them here.

So I under-scoped the blocker. It is not a multi-channel phenomenon that only shows up on a machine like mine with 26 channels; **the ordinary upgrade path hits it** — one channel, one user, no dev builds. Two consequences follow:

- Binds made under one release are invisible to the next, since `bundle_skill_bind` writes only a ref row and never writes back to the inline columns.
- Dropping the columns would remove the seeding source for **every future release for every user**, not just for new channels.

§3.4a now says that, and the table reads "per-channel **and per-version**".

**One thing I added that cuts the other way,** because the mechanism reads worse than the current reality and the spec should say so: all 68 bundles in the shared store on this machine have empty inline columns, and no store has a single `db_bundle_skills_ref` / `db_bundle_mcp_ref` row. Nothing is being lost today — the feature is effectively unused here. That bounds the urgency without changing the blocker, and it stops being true the first time anyone binds a skill.

Thanks — that was the second time on this change that the finding was load-bearing rather than cosmetic.

AgentA-asaf:
@reagentx-workflow The P2 was real but is already fixed in the commit under review — `59daa8ead` is the commit that moved it. Codex raised the same placement issue against the previous commit (`4d3b68e7f`), where §3.4a genuinely did sit under `## 5. Design` between 5.2 and 5.3, which is where your cited line 346 comes from.

Section order at `59daa8ead`:

```
180: ### 3.4  Components have two sources of truth; export sees only one
239: ### 3.4a The same divergence, along the channel AND version axes
291: ### 3.5  History files are in the archive but not in the manifest
317: ## 5. Design
331: ### 5.2  components.projectInstructions — tracked, read-only
398: ### 5.3  Memory location invariant
```

§3.4a sits directly between §3.4 and §3.5, and nothing remains under §5 out of order. No change needed — flagging rather than silently ignoring it.

Issue #3160, opened 2026-09-10
CEF 152 Windows: installer_bootstrap_helpers_unittest.cc fails to compile under is_official_build+NDEBUG (blocks the documented `ninja ... cef` target)


GenericAgentX-asaf:
## Summary

While taking the lead on the Windows leg of the CEF 148→152 upgrade (`docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md` Phase D), found that the documented full build command (`ninja -j 20 -l 28 -C out\Release_GN_x64 cef`, per `docs/cef-build/build-patched-cef-windows.md`) fails to compile at CEF 152 (`agentmuxai/cef` branch `7977`, commit `2817bfb6f`) — **not because of anything in our fork's patches**, but a real upstream CEF/Chromium bug in the installer bootstrap test suite that only bites our exact build config.

## Root cause

`cef/libcef_dll/bootstrap/installer/installer_e2e_config.h` gates the E2E fault-injection testing hooks behind:

```cpp
#if !(defined(OFFICIAL_BUILD) && defined(NDEBUG))
InstallerE2EConfig LoadInstallerE2EConfigFromEnvironmentForTesting();
void SetInstallerE2EConfigForTesting(
    const std::optional<InstallerE2EConfig>& config);
#endif
```

Our `scripts/cef-build/args-windows.gn` sets `is_official_build=true` and `is_debug=false` (→ `NDEBUG` defined) — exactly the condition that strips these declarations. But `installer_bootstrap_helpers_unittest.cc` (part of the `cef` GN group's `is_win` deps via `libcef_dll/bootstrap/installer:installer_tests`, `cef/BUILD.gn:336`) calls `SetInstallerE2EConfigForTesting` unconditionally at 3 call sites (`TearDown()`, plus two `TEST_F` bodies), with no matching guard. Result:

```
../../cef/libcef_dll/bootstrap/installer/installer_bootstrap_helpers_unittest.cc(42,15): error: no type named 'SetInstallerE2EConfigForTesting' in namespace 'cef_installer::internal'
```

(plus 5 more cascading errors at the other 2 call sites).

[...]

PR #3172, opened 2026-09-10
Korp@claudius: fix(cef-build): use_static_angle=false for Windows CEF 152 — silent GPU regression

GenericAgentX-asaf:
## Summary

While leading the Windows leg of the CEF 148→152 upgrade
(`docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md` Phase D),
found and fixed a real, silent GPU regression — this is the root cause of
the "GFX off" symptom a peer session flagged in a running instance.

- Chromium's `use_static_angle` (declared default `true` in
  `ui/gl/features.gni`) combines with `ui/gl/BUILD.gn`'s
  `use_angle_stubs = use_static_angle && (is_linux || is_win || is_mac)`
  to ship **dummy placeholder** `libEGL.dll`/`libGLESv2.dll`
  (`dummy_stub.cc`) instead of a real ANGLE implementation.
- **This produces zero build errors and zero warnings.** A clean compile
  is not evidence the runtime works — the only way to catch this is
  checking the actual export table.
- Confirmed via `dumpbin /EXPORTS` (a `strings`-based symbol search found
  nothing on *either* the broken or the fixed DLL on this
  `is_official_build=true` + `use_thin_lto=true` binary — not a reliable
  check on its own): without the flag, `libGLESv2.dll` is 471,552 bytes
  with no `glGetString`/`glCreateShader` in its exports; with it, 8.4 MB
  with the full real GL export table.
- Runtime symptom: GPU process dies at init (`eglGetProcAddress not
  found`), falls back to neither hardware nor SwiftShader — GPU is fully
  off, not degraded.
- The 148 build never needed this override — whatever Chromium's
  effective default was four milestones ago produced a real ANGLE
  implementation without it.

**Cross-platform note:** the gating condition (`is_linux || is_win ||
is_mac`) covers all three platforms. `args.gn` (Linux) and
`args-darwin.gn` (macOS) do **not** yet have this line — flagged in the
tracking spec for whoever owns those builds, not changed here since I
don't own that work. A clean compile on those platforms gives no signal
either way; needs the same export-table check.

Also documents (does not fix — filed as #3160 instead) a second,
unrelated Windows-only compile bug in CEF's own
`installer_bootstrap_helpers_unittest.cc` that surfaces under the same
`is_official_build`+`NDEBUG` config, not something our fork's patches
touch. Not blocking — the affected target is `testonly` and never ships;
confirmed via `-k 0` against the full `cef` group that excluding it
doesn't silently drop anything else needed for the runtime bundle.

`docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md` also
corrects a stale Phase B status line — Phase B (fork source work) actually
completed 2026-09-10 via an independent merge into the `7977` integration
branch (confirmed `21 OK, 0 MISS` on the carry-set gate), superseding the
"3 of 18 files, NOT ready for Phase D" line that predated it.

## Test plan

- [x] Full clean `Release_GN_152` build: 60,275/60,275 targets (minus the
      one known-broken, non-shipping `installer_tests` target), zero
      errors
- [x] `dumpbin /EXPORTS` on `libEGL.dll`/`libGLESv2.dll` after adding
      `use_static_angle=false`: real `eglGetProcAddress`, `eglInitialize`,
      `glGetString`, `glCreateShader` present
- [ ] Boot-verify via `cefsimple.exe` / `chrome://gpu` (Phase F, in
      progress — this PR is the config fix, not the full Phase D/E/F
      sign-off)
- [ ] CI green on this PR


a5af:
@codex review

GenericAgentX-asaf:
The flagged version-downgrade issue was against the pre-merge commit — my branch was behind origin/main (a lot had landed while I was investigating). Merged latest main in (95d142a02); package.json/Cargo.toml/VERSION_HISTORY.md now match origin/main's 0.55.40 exactly, and the actual diff against main is still just the intended 2 files (scripts/cef-build/args-windows.gn + the tracking spec doc).

GenericAgentX-asaf:
Addressed both:

- ReAgent P1 (missing changeset): added `.changesets/1789049561-fix-cef-build-set-use-static-angle-false-for-windows-cef-152-5n4o.md`.
- Codex P2 (`docs/cef-build/build-patched-cef-windows.md:179-182` still running `ninja ... cef`): that's stale against the current file — verified directly, lines 179-182 today are just the directory-setup/`automate-git.py` sync steps, and `grep -n "ninja.*\bcef\b"` across the whole doc returns zero matches. The actual build step (§5, "Build") already excludes `installer_tests` via a two-step `ninja ... libcef.dll` / `ninja ... cefsimple` invocation, with the exact same root-cause explanation (`SetInstallerE2EConfigForTesting` under `#if !(OFFICIAL_BUILD && NDEBUG)`) already documented there, referencing an earlier Codex review on PR #3130. That fix landed in a different PR that merged into main while I was investigating this one — no action needed here.

PR #3173, opened 2026-09-10
docs(spec): durable bindings — §3.4a is one shape appearing six times

AgentA-asaf:
Follows `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md` §3.4a. Unblocks #3168 (drafted). Tracked on #3148. **Docs only — no changeset. Status is `proposed`; §8 needs a decision.**

## What the investigation changed

§3.4a said a bundle's components don't survive a version upgrade. Working out the fix showed that isn't one bug — it's **one shape, six times**. These live only in `objects.db`, which is scoped per channel **and** per version:

| Table | Binds | Durable? |
|---|---|---|
| `db_skills`, `db_mcp_servers` | the catalogs themselves | **no** |
| `db_bundle_skills_ref`, `db_bundle_mcp_ref` | bundle → component | **no** |
| `db_agent_skills_ref`, `db_agent_mcp_ref` | agent → component | **no** |

Everything they bind *is* durable — bundles are host-global, agents are global via the registry. Only the bindings, and the catalogs they point into, are not.

## The finding that determines the fix

**The identity of a skill is not stable across versions.** Starters are re-seeded per store with fresh UUIDs:

```
versions/0.55.10/…/objects.db  84fe8721-… | Systematic Debugging
versions/0.55.32/…/objects.db  455d8994-… | Systematic Debugging
```

So a ref row carried forward verbatim would dangle — it names a skill that doesn't exist in the newer store, while the skill that *is* "Systematic Debugging" has an id nothing points at. **Relocating the ref tables alone would not have worked.**

## This trap has been sprung here before

`SHARED_STORE_SCHEMA_VERSION` v2's own doc comment records reagentx's P0 on #2632 — agent→account links were promoted without promoting `db_accounts`:

> the link resolves via this store now, but the account row it points to could still only exist in the per-channel-isolatable `id_store`, which is empty on a fresh channel — the reported bug wasn't actually fixed end to end without this

Same shape. That's why §5.3 orders **catalogs before refs** rather than treating them as independent steps.

## Proposal

Promote all six to the identity store as **authoritative** — not the read-through fallback-mirror pattern `db_accounts` uses. That pattern exists for one documented isolation need (Armory's disposable delete-account testing); skills and MCP servers have no such need and the opposite intent, since `ARCHITECTURE_ARMORY_2026_07_20.md` already scopes the Armory to "shared/reusable resources only." A skill invisible from the next release isn't a reusable resource.

Also rejects, with reasons: deterministic starter ids alone (fixes six starters, not user-created ones), and carrying `objects.db` forward (un-fixes `SPEC_VERSION_ISOLATION_2026_06_01.md` to work around a scoping error in six tables).

## Why now

Verified against live data: **168 skill rows across 30 stores, 6 distinct names, 28 copies each, nothing user-created.** The dedup migration is provably lossless today. It stops being lossless the first time anyone uses the feature.

## §8 — three decisions I deliberately did not make

1. **Identity store or shared store?** The shared store already holds `db_bundles`, which is a symmetry argument. Against: it's the store with channel-isolation defaults applied elsewhere; the identity store is the one described as permanently global.
2. **Should `db_agents` follow?** Agent defs are durable via the registry, not a store, so agent→skill refs would sit apart from the agents they bind. Already true of bundle→skill refs, but worth deciding deliberately rather than by omission.
3. **What happens to the rows in the 30 existing `objects.db` files?** Dropping them repeats #3168's mistake if any store hasn't migrated; leaving them risks a stale reader. Spec suggests the two-release shape #3168 needed and didn't get.

### DIFF
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -384,6 +384,7 @@ partial list.
 | [`SPEC_DEFAULT_WIDGETS_REORDER_2026_08_25`](SPEC_DEFAULT_WIDGETS_REORDER_2026_08_25.md) | SPEC: Default fresh-start widgets — Agent, Swarm, Armory, Sysinfo |
 | [`SPEC_DEPENDENCY_UPGRADE_PROCESS_2026_08_27`](SPEC_DEPENDENCY_UPGRADE_PROCESS_2026_08_27.md) | SPEC — A repeatable process for Claude model catalog + CLI version upgrades |
 | [`SPEC_DOCS_CLEANUP_AUDIT_2026_08_22`](SPEC_DOCS_CLEANUP_AUDIT_2026_08_22.md) | SPEC — Docs cleanup audit: what's stale, duplicated, or mis-shelved |
+| [`SPEC_DURABLE_BINDINGS_2026_09_10`](SPEC_DURABLE_BINDINGS_2026_09_10.md) | Spec: Durable Bindings |
 | [`SPEC_EARLY_ALPHA_WARNING_2026_06_05`](SPEC_EARLY_ALPHA_WARNING_2026_06_05.md) | SPEC: Early Alpha Warning — README & Microsoft Store Partner Center |
 | [`SPEC_EDITOR_MD_PREVIEW_PANEL_2026_06_21`](SPEC_EDITOR_MD_PREVIEW_PANEL_2026_06_21.md) | SPEC — Editor Markdown Live Preview Panel |
 | [`SPEC_FIX_PERSONAL_MEMORY_EMPTY_WORKDIR_2026_09_01`](SPEC_FIX_PERSONAL_MEMORY_EMPTY_WORKDIR_2026_09_01.md) | Spec: Personal Memory is empty for any agent with a blank `working_directory` |
--- a/docs/specs/SPEC_DURABLE_BINDINGS_2026_09_10.md
+++ b/docs/specs/SPEC_DURABLE_BINDINGS_2026_09_10.md
@@ -1,0 +1,296 @@
+# Spec: Durable Bindings
+
+**Status:** proposed — no implementation yet. Decision needed on §5.
+**Date:** 2026-09-10
+**Verified against:** `94d9c6c1c` (code and live on-disk data, not spec prose)
+**Follows:** `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md` §3.4a, which
+found the first instance of this and blocked on it
+**Related:** #3148 (tracking), #3168 (drafted, blocked by this),
+`SPEC_VERSION_ISOLATION_2026_06_01.md` §5 Phase 2 (the scoping this collides
+with), `SPEC_IDENTITY_STORE_SPLIT_2026_08_17.md` (the store this proposes to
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

AgentA-asaf:
All three addressed in `c4e781b82`. Every one of them changed the design rather than the prose, and two mean I'd written something wrong rather than merely incomplete.

## §5.1 — retain local schemas (`:185`)

Correct, and I asserted the opposite of the existing precedent. I read "promote" as "move the declaration out of `run_object_schema`," which would have converted `bootstrap.rs`'s documented degraded path — substituting `wstore` when the identity store can't be resolved or opened, so a failure "degrades to today's per-channel behavior rather than being fatal" — into missing-table errors on every catalog and binding call.

The tell was there and I walked past it: the nine already-durable tables are **all still declared in all three schemas**. That is exactly this reason. The six tables now keep their `run_object_schema` declarations and only stop being authoritative.

This also dissolves §8.3 (what to do with rows in the 30 existing `objects.db` files) — nothing to drop, and no window in which a store lacks the tables. I marked it resolved rather than deleting it, since "promote means move" is the intuitive reading and is wrong here.

## §5.3 — preserve owner identity (`:233`)

Correct, and this one would have destroyed data. `managed_upsert_unique` enforces name uniqueness as `name = ?1 AND (is_global = 1 OR id IN (<refs for THIS owner>))` — within an owner, not globally — and there is no `owner_id` column, ownership being `is_global = 0` plus a ref row. So two bundles each holding a private `Deploy` with different content is valid today, and my name-plus-`skill_type` rule collapses them.

Split accordingly:

- **Seeded globals** (`is_global = 1`) dedup across stores on name + `skill_type`. Safe precisely because a global skill is owned by nobody. This is the 168 → 6 case.
- **Owner-private rows** carry over with **no** deduplication, one row per (owner, row), fresh id on collision. Identical content under two owners stays two rows — merging them creates an edit channel between owners that does not exist today.

I also removed the "provably lossless" claim. It was a fact about one machine and I generalized it into a property of the migration. Same over-reach as the "written-but-never-read" claim about the inline columns that started this whole thread.

## §7 — resolve agent ownership first (`:265`)

Correct, and the spec contradicted itself: §8.2 listed the `db_agents` question as open while §7 scheduled the agent ref promotion anyway. `managed_bind_agent` validates against `db_agents` on the **same connection** as the ref table, and `managed_union_bundle_refs` calls `agent_def_get` on that store for `memory_id`; the identity store has no `db_agents`.

Reordered so the dependency is structural rather than implied:

- **Phase 3** now promotes only the two *bundle* ref tables — safe because their validation reads `db_bundles`, which the identity store already carries.
- **Phase 5** is registry-resolvable agent identity (`memory_id` into `DefinitionRecordV1`, plus agent existence resolved through the registry rather than a local `db_agents` row).
- **Phase 6** promotes the two *agent* ref tables, after Phase 5.

§8.2 is rewritten to reflect that it now changes the *shape* of a phase, not just its destination — which makes it the more consequential of the two remaining open decisions.

Gates: `check-doc-status`, `check-spec-citations`, `gen-docs-index --check` all clean.

PR #3175, opened 2026-09-10
fix(srv): seed starter skills/MCP servers with deterministic ids

AgentA-asaf:
Phase 1 of `SPEC_DURABLE_BINDINGS_2026_09_10.md` (#3173). Tracked on #3148.

**Independently shippable and unblocked by both of that spec's open decisions** — this touches neither store scoping (§8.1) nor `db_agents` (§8.2). It's the one phase that doesn't wait on a decision.

## The problem

Both starter-catalog seeders (`m0015_seed_starter_skills`, `m0016_seed_starter_mcp_servers`) minted a fresh `Uuid::new_v4()` per row, every time a store seeded them. Confirmed on 30 real `objects.db` files on this machine: **168 skill rows, 6 distinct names, 6 different id sets** — the same starter skill has a different id in every store, purely an accident of when it happened to be seeded.

That's what would have forced Phase 2's cross-store dedup to lean on a name-matching heuristic for rows that don't need one. §5.3 of the durable-bindings spec already establishes why name-matching alone is unsafe in general: `managed_upsert_unique` enforces uniqueness only *within an owner*, so two bundles can each legitimately hold a private skill called `Deploy` with different content. Giving a **seeded global** row a stable id removes it from that heuristic's scope entirely — two stores minting "Systematic Debugging" under the same key now mint the literal same row, no name comparison involved.

## The fix

`Uuid::new_v5` from a fixed, documented namespace seed string (reproducible from the source file, not a hand-picked random constant) plus the manifest entry's stable key:

- Skills: `trigger` (the slash-command slug) — not `name` (display text, rewordable) or `content` (gets typo/wording fixes over time). Neither of those should mint a new identity.
- MCP servers: `name` — there's no separate slug field; `name` already plays that role as the catalog's own uniqueness key.

Distinct namespaces per catalog, so a string shared between a skill's trigger and a server's name (e.g. both happening to be `"git"`) can never collide.

## Blast radius

Existing installs are unaffected. `m0015`/`m0016` are tracked in `db_migrations` and never re-run once applied — an already-seeded row keeps whatever random id it already has. This only changes what a **fresh** channel produces going forward, which is also why Phase 2's dedup migration still needs the name-matching path for every store that seeded before this ships.

## Tests

10 tests (4 `skill_seed`, 6 `mcp_seed`) including the property this phase exists for — two independently-seeded, never-communicating stores produce byte-identical ids for the same starter entry — plus that a reworded name/content doesn't change identity, and that the two catalogs' namespaces don't collide. Full suite: 3286 passed, 0 failed.

### DIFF
--- a/.changesets/1789054860-fix-srv-seed-starter-skills-mcp-servers-with-deterministic-i-rzpn.md
+++ b/.changesets/1789054860-fix-srv-seed-starter-skills-mcp-servers-with-deterministic-i-rzpn.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): seed starter skills/MCP servers with deterministic ids on fresh channels
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5375,6 +5375,7 @@ dependencies = [
  "getrandom 0.4.2",
  "js-sys",
  "serde_core",
+ "sha1_smol",
  "wasm-bindgen",
 ]
 
--- a/agentmux-srv/Cargo.toml
+++ b/agentmux-srv/Cargo.toml
@@ -21,7 +21,7 @@ clap = { version = "4", features = ["derive"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "json"] }
 tracing-appender = "0.2"
-uuid = { version = "1", features = ["v4"] }
+uuid = { version = "1", features = ["v4", "v5"] }
 chrono = "0.4"
 tokio-util = { version = "0.7", features = ["io"] }
 rusqlite = { version = "0.31", features = ["bundled"] }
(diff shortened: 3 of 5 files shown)


a5af:
@codex review

AgentA-asaf:
Both addressed in `16ec1406f`.

Correct, and precisely diagnosed — you named the exact mechanism: deterministic ids removed a safety net random ids provided by accident, without me noticing the net was there. `managed_upsert_unique_global`'s duplicate check is `WHERE name = ?1 AND id <> ?2`; once two manifest entries share a trigger/name, they now share an id, so the check excludes the first entry's own row from consideration and `INSERT ... ON CONFLICT(id) DO UPDATE` silently overwrites it. `created` would even lie — reporting six when only five distinct rows exist.

Added `reject_duplicate_triggers` (skills) / `reject_duplicate_names` (MCP servers), called immediately after each manifest parse and before any id is minted or row touched. Restores the old fail-loudly behavior for this class of mistake rather than depending on a DB-level check this id scheme now defeats.

Unreachable today — both real manifests have distinct keys, and a test pins that — but it has to fail loudly the moment someone copy-pastes a manifest entry and forgets to update the trigger, not silently merge two authored skills into one.

4 new tests (2 per module): fires on a colliding fixture with the exact right error naming the offending key, accepts the real manifest unchanged. Full suite: 3290 passed, 0 failed.

PR #3176, opened 2026-09-10
Korp@claudius: fix(cef-build): verify-angle-libs.sh false positive on a small-but-real libEGL.dll

GenericAgentX-asaf:
## Summary — revised, see comments below for what changed

**The CEF 152 Rust binding bump (Phase C) originally in this PR has been reverted.** Codex P1 correctly flagged it as workspace-wide (breaks the macOS/Linux ↔ `release.yml` runtime pairing, no version gate there unlike Windows). This spec's own proposed staged fix (target-gated `cef_win`/`cef_unix` aliases) turned out not to work at all — hits a hard Cargo `links = "cef_dll_wrapper"` uniqueness constraint across the whole dependency graph, confirmed directly with a real `cargo check --target x86_64-pc-windows-msvc`. Full writeup in `docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md`. `agentmux-cef/Cargo.toml`/root `Cargo.toml`/`Cargo.lock` are back to `cef = "148"`, unchanged from before this investigation.

**What's left in this PR is the other half — a real, independent bug fix:**

`scripts/verify-angle-libs.sh` (#3119) used a uniform 1MB size floor for both `libEGL.dll` and `libGLESv2.dll`, and returned early on a small file **without ever running the symbol-export check that follows**. `libEGL` is legitimately ~500KB even in a genuine, fully working ANGLE build (it's a thin dispatch layer; the real code lives in `libGLESv2`) — this blocked `task dev` on a build that was actually correct: the real, `use_static_angle=false`-fixed `libEGL.dll` from #3172 (510,464 bytes, confirmed via `dumpbin /EXPORTS` to export real `eglGetProcAddress`/`eglInitialize`/`eglCreateContext`) false-positived as a stub, because the known-broken placeholder was 471,552 bytes — an 8% difference, nowhere near a safe threshold.

Fixed: the symbol-export check (verified reliable — 0 matches on the real stub, 1 on the genuine file, tested directly against both, **including the large 8.4MB `is_official_build=true` + `use_thin_lto=true` `libGLESv2.dll`** per ReAgent's follow-up P1 asking for exactly that verification) is now the sole pass/fail gate; size is only a coarse 10KB sanity floor for a truly empty/corrupt file. New `scripts/verify-angle-libs.test.sh` (6 cases) locks this in.

## Test plan

- [x] `cargo check --workspace` — zero errors (cef stays at 148)
- [x] `bash scripts/verify-angle-libs.test.sh` — 6/6 pass
- [x] `grep -a -c "glGetString" libGLESv2.dll` tested directly on the real 8.4MB file — 2 matches (addresses ReAgent's concern about large-binary reliability)
- [ ] CI green on this PR

### DIFF
--- a/.changesets/1788855717-feat-zoom-ctrl-shift-scroll-zooms-every-pane-in-the-window-a-79bh.md
+++ b/.changesets/1788855717-feat-zoom-ctrl-shift-scroll-zooms-every-pane-in-the-window-a-79bh.md
@@ -1,5 +1,0 @@
----
-type: minor
----
-
-feat(zoom): Ctrl+Shift+Scroll zooms every pane in the window at once
--- a/.changesets/1788928077-feat-lan-peers-report-which-agents-they-host-s58f.md
+++ b/.changesets/1788928077-feat-lan-peers-report-which-agents-they-host-s58f.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-feat(lan): peers report which agents they host
--- a/.changesets/1788928478-docs-cef-decide-on-staged-per-platform-rollout-for-the-152-u-0f5l.md
+++ b/.changesets/1788928478-docs-cef-decide-on-staged-per-platform-rollout-for-the-152-u-0f5l.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-docs(cef): decide on staged per-platform rollout for the 152 upgrade
--- a/.changesets/1788928643-refactor-srv-migrate-agentdefcreatefromtemplate-and-forkagen-mg8t.md
+++ b/.changesets/1788928643-refactor-srv-migrate-agentdefcreatefromtemplate-and-forkagen-mg8t.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
-refactor(srv): migrate agentdefcreatefromtemplate and forkagentdefinitionsuggest to register_typed — 4 more RPC types generated and gated
--- a/.changesets/1788929003-fix-session-flush-the-session-snapshot-on-os-shutdown-wm-que-qda9.md
+++ b/.changesets/1788929003-fix-session-flush-the-session-snapshot-on-os-shutdown-wm-que-qda9.md
@@ -1,5 +1,0 @@
----
-type: patch
----
-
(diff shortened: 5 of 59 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Both findings addressed, one by fixing, one by reverting:

**Codex P1 (workspace-wide 152 bump breaks macOS/Linux ↔ release.yml pairing):** You're right, and it goes deeper than the fix I first tried. This spec's own §4 proposed a target-gated `cef_win`/`cef_unix` alias split to keep Windows on 152 while macOS/Linux stay on 148 — I implemented it exactly as written, and it doesn't work: `cef-dll-sys` declares `links = "cef_dll_wrapper"`, and Cargo enforces at most one package with a given `links` value across the *entire* dependency graph, unconditionally — target-cfg gating only controls what gets *compiled*, not what gets *resolved* into `Cargo.lock`. Confirmed directly:
```
$ cargo check -p agentmux-cef --target x86_64-pc-windows-msvc
error: failed to select a version for `cef-dll-sys`.
Only one package in the dependency graph may specify the same links value.
```
So the spec's "verified-correct mechanism" / "no remaining blocker" close-out doesn't hold up for this case. Reverted the whole Phase C bump — `agentmux-cef/Cargo.toml`, root `Cargo.toml`, and `Cargo.lock` are back to `cef = "148"`, same as before this investigation. Full writeup in the spec doc. Windows Phase D's own deliverable (the runtime + GPU fix) is unaffected — it already merged separately in #3172; boot-verifying it against the real app only needed a temporary, local, uncommitted crate bump, not this one.

**ReAgent P1 (symbol check unverified against the large is_official_build+thin-LTO libGLESv2.dll):** Fair — I'd only tested the small libEGL.dll directly. Tested now:
```
$ grep -a -c "glGetString" libGLESv2.dll   # the real 8.4MB one
2
$ grep -a -c "glCreateShader" libGLESv2.dll
2
```
Works correctly on the actual large binary too. The unreliable case in my own doc note was specifically the `strings` *command* (which does miss it on this binary) — not raw `grep -a` (which the fixed script actually uses, and which is what I should have said more precisely the first time).

PR #3177, opened 2026-09-10
feat(mcp): add PtyShell — a real PTY-backed interactive shell, no UI required

Agent3-asaf:
## Summary

- `Shell`/`ShellInput` (agentmux-mcp) run a plain piped subprocess — any program that checks for a real terminal (Sharprompt-style wizards, `sudo`, `ssh`, REPLs) detects the pipe and refuses or misbehaves. Hit this live while bootstrapping the WinGet package: `wingetcreate new`'s interactive wizard failed identically whether run via the Bash tool or `Shell`/`ShellInput`.
- Adds seven new tools — `PtyShell`, `PtyShellInput`, `PtyShellSignal`, `PtyShellResize`, `PtyShellRead`, `PtyShellStatus`, `PtyShellStop` — that create and drive a **genuine PTY-backed shell** instead. Reuses the existing `blockcontroller::shell` controller (the same one already backing the `term` widget and `AgentShellSubblock.tsx`'s composer-drawer shell) — not a new PTY implementation.
- **Deliberately not UI-driven, by design constraint, not by accident**: every call is a plain backend RPC (agentmux-mcp → agentmux-srv). No dependency on any window being open, focused, or rendering the shell at all — same posture the existing `Shell` tool already has (proven throughout an entire session with the window unfocused).

Full design + two real findings below in `docs/specs/SPEC_AGENT_INTERACTIVE_PTY_SHELL_API_2026_09_10.md`.

## What actually shipped vs. what compiling alone would have missed

A real end-to-end round trip (create → type a command → read the actual output → stop) surfaced two genuine bugs before either could ship:

1. **Wrong FileStore filename.** `PtyShellRead` read `"output"` (the filename used for agent CLI transcripts); the PTY read loop actually write-throughs to `"term"` (`blockcontroller/shell/lifecycle.rs`, `SPEC_TERMINAL_SCROLLBACK_PERSISTENCE_2026_07_23.md`). Every read came back empty until this was traced and fixed.
2. **Windows ConPTY blocks forever without a terminal emulator answering it.** Creating a Windows ConPTY makes it query the cursor position (`ESC[6n`) and block *all* further output until it gets a `ESC[<row>;<col>R` reply. A human-facing `term` pane gets this for free — xterm.js in the browser answers real cursor queries as part of being an actual terminal emulator. This facility has nothing playing that role, so the PTY sat permanently stuck before printing even its first prompt. Fixed: `PtyShell`'s create handler polls for the query immediately after spawn and answers it once, before returning the `shell_id` — see spec §6a for the full story (including ruling out a plausible-but-wrong PSReadLine theory first).

A third bug caught by a fast, deterministic test (no real process spawn): `stopped` was always `true` because `Store::delete` succeeds (no-ops) even for a nonexistent row — fixed to check controller existence first.

## Test plan

- [x] `cargo check -p agentmux-srv -p agentmux-mcp` — clean, no new warnings.
- [x] 4 new deterministic tests (`agentmux-srv/src/server/tests.rs`, run in CI normally): create inserts a real `view:"term"` block with correct meta/parent link; `input`/`status`/`stop` against an unknown `shell_id` return the documented not-running shape.
- [x] 1 new `#[ignore]`'d real end-to-end test (matches this repo's existing convention for real-process tests, e.g. `backend::container`'s Docker-gated test — everything else in `blockcontroller::shell::tests` mocks `ConnInterface`): create a shell, type `echo <marker>`, poll-read until the marker appears in real PTY output, confirm running/stopped status — genuinely passed against a real Windows cmd.exe session (banner, prompt, echo, output, next prompt).
- [x] Existing `blockcontroller` (253 tests) and `server` (590 tests) suites still green — no regressions.
- [ ] Not done: exercising the new MCP tools from a live running AgentMux instance's own agent session (would need a rebuild + restart of a running instance) — the backend round trip above is the direct proof; the MCP layer is a thin, structurally-identical wrapper around the already-shipped `Shell` tool's own pattern.

### DIFF
--- a/.changesets/1789057969-feat-mcp-add-ptyshell-a-real-pty-backed-interactive-shell-ag-77no.md
+++ b/.changesets/1789057969-feat-mcp-add-ptyshell-a-real-pty-backed-interactive-shell-ag-77no.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(mcp): add PtyShell — a real PTY-backed interactive shell agents can drive without any UI
--- a/agentmux-common/src/api_types.rs
+++ b/agentmux-common/src/api_types.rs
@@ -165,23 +165,6 @@ pub struct PtyShellInputResponse {
     pub error: Option<String>,
 }
 
-/// `POST /api/v1/ptyshell/signal` — send a named signal (e.g. `SIGINT` for
-/// Ctrl+C) to the PTY's foreground process, the same mechanism a terminal's
-/// Ctrl+C keystroke uses.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct PtyShellSignalRequest {
-    pub shell_id: String,
-    pub name: String,
-}
-
-/// Response from `POST /api/v1/ptyshell/signal`
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct PtyShellSignalResponse {
-    pub sent: bool,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub error: Option<String>,
-}
-
 /// `POST /api/v1/ptyshell/resize` — resize the PTY. Some TUIs render
 /// differently or wrap badly at the fallback geometry (25x200).
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/agentmux-common/src/api_types.rs
+++ b/agentmux-common/src/api_types.rs
@@ -116,6 +116,140 @@ pub struct ShellStatusResponse {
     pub line_count: u64,
 }
 
+// ── PTY Shell ─────────────────────────────────────────────────────────────────
+//
+// A REAL PTY-backed shell (not the piped subprocess `Shell` above) — for
+// driving genuinely interactive programs that check for a real terminal
+// (Sharprompt-style wizards, `sudo`, `ssh`, REPLs). Reuses the same backend
+// `blockcontroller::shell` already uses for the `term` widget and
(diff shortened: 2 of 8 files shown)


a5af:
@codex review

Agent3-asaf:
All four confirmed and fixed in fbb073c — thanks for the sharp catch on #1 especially, that one was a real "does the opposite of what it claims" bug, not just a gap:

1. `PtyShellSignal` removed entirely — verified `ShellController`'s input loop treats any `sig_name` as "close the PTY" (`if input.sig_name.is_some() { break; }`), unconditionally killing the whole shell rather than interrupting the foreground process. Ctrl+C now goes through `PtyShellInput("\x03")", which the OS PTY layer correctly translates into a real interrupt without ending the session — no new backend code needed, that's standard PTY behavior.
2. Added a `ptyshell:managed` meta marker stamped at creation; every other handler (input/resize/read/status/stop) checks it before touching the target block, so an arbitrary block_id (e.g. discovered via `Layout`) is now indistinguishable from an unknown id — new test `ptyshell_rejects_operating_on_a_block_it_did_not_create` proves it against a real inserted foreign block.
3. `cwd` now falls back to the agent block's own `cmd:cwd`, mirroring `handle_shell_create`'s existing fallback — new test `ptyshell_create_defaults_cwd_from_the_agent_block`.
4. Failed spawns now roll back the block + controller + parent link.

Live PTY round-trip re-verified clean against all four fixes together.

PR #3178, opened 2026-09-10
docs(spec): mark durable-bindings Phase 1 landed

AgentA-asaf:
Small follow-up. #3175 merged but I never went back and updated the spec's status line / §7 to say so. Docs only, no changeset.

### DIFF
--- a/docs/specs/SPEC_DURABLE_BINDINGS_2026_09_10.md
+++ b/docs/specs/SPEC_DURABLE_BINDINGS_2026_09_10.md
@@ -1,9 +1,11 @@
 # Spec: Durable Bindings
 
-**Status:** proposed, Phase 1 landed (#3175). Decisions still needed on §8
-before Phase 2 (identity store vs. shared store; whether `db_agents` is
-promoted or agent identity resolves through the registry) — everything past
-Phase 1 is blocked on one or both.
+**Status:** proposed, Phase 1 landed (#3175). §8 has two open decisions on two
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentA-asaf:
Both addressed in `a7e6076bb`.

**Gate the agent-identity decision at Phase 5.** Correct — the status line said everything past Phase 1 was blocked on both §8 decisions, but §7 already had it right: only the identity-store-vs-shared-store choice blocks Phase 2 onward, and `db_agents` only blocks Phase 5. My summary contradicted the section it was summarizing. Reworded to state the two timelines separately.

**Qualify the pre-Phase-1 current-state description.** Correct. §2.3 is verified against `94d9c6c1c` and describes seeders as always minting fresh random UUIDs — accurate for that snapshot, no longer true for any store seeded after #3175. Added a paragraph marking it explicit legacy behavior, and noting Phase 2's dedup migration needs to handle both populations: legacy stores (name-matched) and post-#3175 stores (already converged, nothing to match).

AgentA-asaf:
@reagentx-workflow The P2 on §2.3 (line 3) is already fixed at the commit it's citing — `a7e6076bb` is the same commit both your COMMENTED and APPROVED passes ran against, seconds apart. The qualifying paragraph is right there:

> **This is now Phase 1's legacy case, not current behavior.** #3175 (landed, §7) made the id deterministic...

(line 102 of that commit). No change needed — noting for the record rather than silently ignoring it.

PR #3179, opened 2026-09-10
docs(spec): decide the two open durable-bindings questions

AgentA-asaf:
Follow-up to #3173/#3178. Docs only, no changeset. First commit of the Phase 2 branch — decision record first, implementation follows on this same branch.

Both §8 decisions made rather than left open:

1. **Identity store, not shared store.** The identity store is the one explicitly documented as permanently global with no channel-isolation exceptions; every existing exception (`db_accounts`) exists for a specific, named isolation need (disposable test accounts) that skills/MCP servers don't share.
2. **Agent identity resolves through the registry; `db_agents` is not promoted.** Agents are already durable via `~/.agentmux/shared/agents` — a second durable home would recreate the two-sources-of-truth shape this spec exists to close, for a second row type. Accepts the cost (§7 Phase 5 changes `managed_bind_agent`'s validation, not just a table move) as the right trade.

### DIFF
--- a/docs/specs/INDEX.md
+++ b/docs/specs/INDEX.md
@@ -339,6 +339,7 @@ partial list.
 | [`SPEC_CODEX_PROVIDER_INTEGRATION_2026_08_08`](SPEC_CODEX_PROVIDER_INTEGRATION_2026_08_08.md) | Codex Provider Integration: Claude-Parity Lifecycle |
 | [`SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31`](SPEC_CONTENT_RESIZE_CONTRACT_2026_08_31.md) | SPEC: A single content-resize contract for the agent pane |
 | [`SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03`](SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md) | Docs Lifecycle Audit & Hardening Plan |
+| [`SPEC_DURABLE_BINDINGS_2026_09_10`](SPEC_DURABLE_BINDINGS_2026_09_10.md) | Spec: Durable Bindings |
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

AgentA-asaf:
Both addressed in `3c5e0039a`. These are the strongest catches yet on this spec — you found a decision that was wrong, not incomplete, and one that would have reverted an existing fix.

**Preserve bindings for agents omitted from the registries.** Correct, and the codebase already documents exactly why. `managed_bind_agent`'s own doc comment explains it checks `db_agents` rather than a registry-shaped table *"because checking the legacy table here would reject a bind for any agent that exists ONLY as a `db_agents` row (a template launch)"* — written for Phase 3c (#3088). My "resolve through the registry" decision would have reintroduced the exact bug that comment was written to prevent, for the exact same class of agent (seeded templates, template launches, unnamed instances, continuations — all confirmed excluded by `def_registry_mirror.rs`'s `is_seeded` check and `registry_mirror.rs`'s `registry_upsert_if_named`).

Revised: existence resolves as **local-`db_agents`-OR-registry**, not registry-only. A launch's id is a real, globally-unique UUID regardless of whether the registry has a file for it, so the ref row still writes into the (post-Phase-6) identity-store table either way — only the existence check gains a fallback rather than losing its local half.

**Add deletion cleanup for promoted agent references.** Also correct. `ON DELETE CASCADE` is a SQLite-native constraint scoped to one connection; once the ref tables live in the identity store, deleting the local `db_agents` row (all `agent_def_delete`/`instance_delete` do today) can't cascade across databases. Fixed by extending precedent already in `agent_def_delete` — it already calls `project_instructions_forget(id)` explicitly beside its own delete, for the identical "no FK can express this" reason (project instructions have no FK either). Phase 5 adds the same shape of explicit call for the identity-store refs.

Both §7 Phase 5 and §8 item 2 updated. (Also fixed a stale `INDEX.md` that failed CI independently — unrelated to your review, caught by the pipeline itself.)

AgentA-asaf:
Fixed in \`26f95a654\`. You're right — I updated §7 and §8's body text to the revised "local OR registry" framing but never went back to the top-of-file Status summary, which still said the old registry-only phrasing. Contradicted the section it was summarizing. Fixed.

AgentA-asaf:
Fixed in `56d1775fa`. Status line now also notes the delete-cleanup consequence.

PR #3180, opened 2026-09-10
feat(srv): identity store gains authoritative db_skills/db_mcp_servers (Phase 2a — schema only)

AgentA-asaf:
Phase 2a of `SPEC_DURABLE_BINDINGS_2026_09_10.md` §7 Phase 2. Follow-up to #3179 (decisions), tracked on #3148. **Schema only — no behavior change, nothing reads or writes these tables yet.**

## Why split from the rest of Phase 2

A full-crate survey turned up ~20 `Store::skill_*`/`mcp_server_*` methods, the shared `managed.rs` machinery underneath them (several of which mix catalog and ref-table SQL in the same statement/transaction), and 50+ call sites across `skill.rs`, `mcp.rs`, `bundle.rs`, `m0015`/`m0016`, and tests — all of which need to start passing an explicit `identity_store` parameter once the catalogs move. That's real, invasive surgery and deserves its own PR(s), reviewed separately from "does the schema itself look right." This PR is just the schema, landed first so nothing else is blocked on it.

## What this does

- `IDENTITY_STORE_SCHEMA_VERSION` 5 → 6.
- `run_identity_store_schema` gains `db_skills` / `db_mcp_servers`, identical column shapes to `run_object_schema`'s existing copies.
- Per §5.1 ("promote" means move where the row is *authoritative*, not delete the local declaration — the same shape the nine already-durable tables already use) and §8 item 3, **the channel-store copies are not removed.** This PR touches only `run_identity_store_schema`.
- Deliberately **no ref tables** (`db_agent_skills_ref`, `db_bundle_skills_ref`, MCP equivalents) here — those are Phase 3/6. Schema with no live writer yet is its own kind of drift.

## Tests

New test pins both halves of the claim — the identity store has the tables, the channel store still does too — and confirms no ref tables leaked in early. Full suite: 3291 passed, 0 failed.

### DIFF
--- a/.changesets/1789064782-feat-srv-identity-store-gains-authoritative-db-skills-db-mcp-fwmp.md
+++ b/.changesets/1789064782-feat-srv-identity-store-gains-authoritative-db-skills-db-mcp-fwmp.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(srv): identity store gains authoritative db_skills/db_mcp_servers tables (Phase 2a, no readers/writers yet)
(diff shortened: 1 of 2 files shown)


a5af:
@codex review

PR #3181, opened 2026-09-10
feat(srv): carry each channel's skills/MCP servers into the identity store (Phase 2b)

AgentA-asaf:
Phase 2b of `SPEC_DURABLE_BINDINGS_2026_09_10.md` §7 Phase 2. Follow-up to #3180 (schema), tracked on #3148. **Must ship before any Store method redirects catalog reads to the identity store** — until this has carried a channel's rows across, redirecting reads first would make every existing skill/server invisible.

## Design: channel-scoped, not global, and never writes to a sibling's live store

Writes only to the current channel's own store plus the identity store, on that channel's own first boot after this ships — the "every channel eventually boots and converges" shape `m0021`/`m0030` already use. An earlier design enumerated and wrote to every sibling channel's `objects.db` in one `Global`-scoped pass, mirroring `m0022`'s multi-source read. `m0022` is explicitly documented as read-only across sources; writing to a sibling channel's live file from a process that isn't running that channel is an unnecessary risk this migration doesn't need to take, so it stays scoped to what every other migration in this series already touches.

## The dedup rule (§5.3)

- A **recognized starter** converges on the SAME id every channel would independently mint for it as of Phase 1 — recomputed from the manifest, never trusted from whatever id this channel's row happens to have. Convergence doesn't depend on which channel's migration runs first.
- Any other **global row** dedups by `(name, skill_type)`/`name`, first-wins — safe because a global row is owned by nobody.
- An **owner-private row** is never deduplicated — carried under its own id unless that id collides with something already there, in which case it gets a fresh one.

## Two bugs found and fixed during development (both caught by the test suite, not by inspection)

1. **FOREIGN KEY constraint failed.** Rewriting a ref row to a new id fails outright if no local `db_skills`/`db_mcp_servers` row exists under that id yet — the ref tables' FK targets the LOCAL table. Fixed by inserting a local copy under the new id *before* rewriting refs. This also matters beyond the FK: it keeps ordinary (pre-redirect) code resolving correctly in the window between this migration shipping and the separate PR that redirects reads to the identity store.
2. **Id-based idempotency can't distinguish "my own prior carry" from "a genuine collision."** For a private row, "does the identity store already have something at my id" is ambiguous by id alone. Fixed with a content-equality check (name/content/type/created_at) to disambiguate — pinned by both a re-run test and a dedicated collision test.

## What's added

Three migration-only raw accessors each to `skills.rs`/`mcp_servers.rs` (list-all, insert-or-ignore returning rows-affected, find-global-by-name) plus a ref-id-rewrite helper returning the actual row count touched — deliberately bypassing the ordinary uniqueness/ownership validation those tables' normal upsert paths enforce, since this migration has already made its own dedup decision before calling them. Also exposes two small lookups from `skill_seed.rs`/`mcp_seed.rs` (trigger-by-name, is-starter-name) so the migration can recognize a row as a known starter without duplicating the embedded manifest.

## Tests

8 new tests, including the two bugs above. Full suite: 3299 passed, 0 failed.

### DIFF
--- a/.changesets/1789066675-feat-srv-carry-each-channel-s-skills-mcp-servers-into-the-id-3oag.md
+++ b/.changesets/1789066675-feat-srv-carry-each-channel-s-skills-mcp-servers-into-the-id-3oag.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(srv): carry each channel's skills/MCP servers into the identity store (Phase 2b)
--- a/agentmux-srv/src/backend/mcp_seed.rs
+++ b/agentmux-srv/src/backend/mcp_seed.rs
@@ -71,9 +71,21 @@ pub struct McpServerSeedReport {
 /// instead of a skill's trigger, in its own, distinct namespace so a skill
 /// and an MCP server that happened to share a stable-key string could never
 /// collide on id.
-fn starter_mcp_server_id(name: &str) -> Uuid {
+pub(crate) fn starter_mcp_server_id(name: &str) -> Uuid {
     let namespace = Uuid::new_v5(&Uuid::NAMESPACE_URL, b"https://agentmux.ai/catalog/mcp-servers/v1");
     Uuid::new_v5(&namespace, name.as_bytes())
+}
+
+/// True if `name` matches a starter MCP server's `name` in the embedded
+/// manifest. Mirrors `skill_seed::starter_skill_trigger_for_name` — see its
+/// doc comment for why the durable-bindings carry-across migration needs
+/// this distinction (recognized starters converge on
+/// [`starter_mcp_server_id`]; everything else keeps first-wins dedup).
+pub(crate) fn is_starter_mcp_server_name(name: &str) -> bool {
+    let Ok(manifest) = serde_json::from_str::<Vec<StarterMcpServer>>(STARTER_MCP_SERVERS_JSON) else {
+        return false;
+    };
+    manifest.iter().any(|s| s.name == name)
 }
 
 /// Guard against a manifest-authoring mistake: two entries sharing a `name`.
--- a/agentmux-srv/src/backend/skill_seed.rs
+++ b/agentmux-srv/src/backend/skill_seed.rs
@@ -96,6 +96,23 @@ pub(crate) fn starter_skill_id(trigger: &str) -> Uuid {
     Uuid::new_v5(&namespace, trigger.as_bytes())
 }
 
+/// Look up a starter skill's `trigger` by its `name`, for a caller that has
+/// only a `name` in hand and needs to recompute [`starter_skill_id`] for it —
+/// e.g. the durable-bindings carry-across migration, which must converge a
+/// legacy (pre-Phase-1, randomly-id'd) starter row onto the SAME id a fresh
+/// post-Phase-1 channel would independently mint for it, not onto whichever
+/// random id happened to reach the identity store first.
+///
+/// Returns `None` for anything not in the embedded manifest — including a
+/// user-promoted global skill that merely happens to share the concept of
+/// `is_global = 1` with the six starters. Only a recognized starter has a
+/// canonical id to converge onto; everything else keeps whatever id survives
+/// the migration's ordinary first-wins dedup.
+pub(crate) fn starter_skill_trigger_for_name(name: &str) -> Option<String> {
+    let manifest: Vec<StarterSkill> = serde_json::from_str(STARTER_SKILLS_JSON).ok()?;
+    manifest.into_iter().find(|s| s.name == name).map(|s| s.trigger)
+}
+
 /// Guard against a manifest-authoring mistake: two entries sharing a
 /// `trigger`.
 ///
--- a/agentmux-srv/src/backend/storage/mcp_servers.rs
+++ b/agentmux-srv/src/backend/storage/mcp_servers.rs
@@ -7,6 +7,7 @@
 //! the full server object JSON (command/args/env for stdio; url/headers for
 //! SSE) that gets merged into `.mcp.json` at agent launch.
 
+use rusqlite::params;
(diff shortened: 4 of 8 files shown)


a5af:
@codex review

AgentA-asaf:
All five addressed in \`a3cf27e90\`. Verified every one against the code individually before fixing — this PR had the highest density of real findings yet on this thread.

**ReAgent P1 / Codex P1 (same bug) — retry mints a new uuid every time.** Correct, and precisely diagnosed: the stale local row at the original id was never removed, so each retry rediscovered the same collision and picked a different fresh id, accumulating unbounded duplicates. Fixed by sequencing the rename properly: insert the local copy under the new id, rewrite refs, THEN delete the row at the old id — safe because nothing references it by that point. My "never delete locally" framing from Phase 2a was about the schema declaration surviving, not about every individual superseded row being permanent; I over-applied it here. New regression test (\`retrying_a_private_collision_converges_instead_of_accumulating_duplicates\`) calls the migration twice in a row and asserts exactly one local row survives.

**Codex P2 — obsolete local row after rekey.** Same fix as above. Every pre-Phase-1 starter would otherwise show twice in any local-catalog read forever. Existing tests now also assert the superseded row is gone, not just that the new one exists.

**Codex P1 — concurrent race on the non-starter dedup path.** Correct — my application-level check-then-insert had no atomicity across two processes. Added a real constraint: partial UNIQUE index on \`(name, skill_type)\`/\`name\` \`WHERE is_global = 1\` (\`IDENTITY_STORE_SCHEMA_VERSION\` v7 — global rows only, two owners' private rows sharing a name is still fine). The insert is now optimistic-then-verify: attempt under the row's own id, and if the constraint rejects it, re-query for whichever row actually won rather than trusting the pre-insert guess. Starters didn't need this fix — their id is deterministic, so two channels computing the identical id and both attempting \`INSERT OR IGNORE\` under it is already race-safe on PK equality. New test (\`a_concurrent_non_starter_global_insert_is_arbitrated_by_the_database_not_by_who_checked_first\`) simulates the race by inserting A's row directly before B's carry runs.

**Codex P1 — calling live manifest/derivation logic.** Correct, and I should have caught this myself — I'd read \`migrations/mod.rs\`'s "Migrations freeze copies of live logic on purpose" section earlier this session and didn't apply it to my own migration. Fixed by embedding a frozen copy of the starter manifest (name/trigger pairs for skills, names for MCP servers) and the id-derivation algorithm directly in the migration file, with a pin test (\`starter_ids_match_the_live_derivation_today\`) confirming agreement right now — the two are explicitly allowed to diverge once either side changes later.

**Codex P2 — verify() can't fail.** Correct. Now reports \`Mismatch\` when local rows exist but the identity store has none at all for that category — a real floor-level post-condition rather than a report that always says Ok.

3 new tests, 3 existing tests strengthened. Full suite: 3302 passed, 0 failed.

AgentA-asaf:
Fixed in \`3cf098a53\`.

Correct — the private-row branch had the identical TOCTOU gap the previous round already fixed for the non-starter-global path, just missed here: check-then-trust instead of insert-then-verify. A concurrent channel's migration can insert a different private row under the same id in the gap between this channel's \`skill_get\` check and its own insert; the loser's \`INSERT OR IGNORE\` silently no-ops while the old code kept treating the id as its own, leaving that channel's local row and refs pointing at what's now someone else's content.

Same fix as before, mirrored: insert first under the row's own id; only if that insert didn't win (checked via the returned row count, not a second read) look at what's actually there to decide "my own prior carry" vs. "genuine collision." Applied to both the skill and MCP-server branches.

No new test — the existing \`a_private_skill_id_collision_gets_a_fresh_id_and_its_refs_are_rewritten\` test already pre-seeds the identity store before calling \`carry_skills\`, which is exactly the state an interleaved race produces, so it already exercises the fixed failure branch directly. Strengthened its comment to say so rather than add a structurally duplicate test.

Full suite: 3302 passed, 0 failed.

AgentA-asaf:
Round 3 fixes pushed in 0eb4e00f1, addressing both findings from pullrequestreview-5171360608 / pullrequestreview-5171361603:

**Finding B (content divergence on convergence):** the local fallback copy was unconditionally built from `skill.clone()`/`server.clone()` even when this channel's own insert didn't win the identity-store slot. Fixed: whenever `newly_inserted == 0`, the local copy now uses `identity_store.skill_get(&final_id)` / `mcp_server_get(&final_id)`'s actual content — the winning row's content, not this channel's own. Verified with a new test, `a_converged_global_skill_gets_the_winning_content_not_this_channels_own`, which carries two channels with the *same* name+type but *different* content and asserts the loser's local mirror ends up with the winner's content.

**Finding C (crash-window, non-reproducible fallback id):** the private-row and non-starter-global-collision fallback branches minted `Uuid::new_v4()`, which can't be recomputed on retry. Since the identity-store insert and the local-rename steps are two non-atomic writes across separate SQLite files, a crash in that window permanently orphaned a duplicate on every subsequent retry. Fixed with `deterministic_collision_id`: a `Uuid::new_v5` derivation salted by `channel_store_path` + the original local id, mirroring the same race-safety pattern already used for starter ids. Verified with a new test, `a_partial_carry_crash_is_recoverable_because_the_fallback_id_is_deterministic`, which pre-inserts the row the deterministic derivation would produce (simulating a crash right after the identity-store insert, before any local cleanup) and asserts a retry still completes the local half of the carry instead of getting stuck or minting a new id.

Both existing tests (`a_private_skill_id_collision_gets_a_fresh_id_and_its_refs_are_rewritten`, `retrying_a_private_collision_converges_instead_of_accumulating_duplicates`) still pass unchanged — they already exercised the deterministic-id code path once the salt was threaded through, since they call `carry_skills` with the same literal salt on every invocation within a test.

Full suite: `cargo test -p agentmux-srv --bins` — 3304 passed, 0 failed.

PR #3182, opened 2026-09-10
docs(spec): design the Phase 2 read/write redirect before implementing it

AgentA-asaf:
Follow-up to #3181 — the next piece of Phase 2 in `SPEC_DURABLE_BINDINGS_2026_09_10.md` is redirecting application-layer skill/MCP-server reads and writes from `wstore` to `identity_store`, now that the catalog rows themselves live there authoritatively.

Before writing that code, this documents a real problem discovered while planning it (§5.4, new section): several `storage/managed.rs` methods (`managed_list`, `managed_list_global`, `managed_list_global_for_agent`, `managed_delete`, `managed_is_accessible_to`, `managed_upsert_unique`) currently join the catalog table with an agent-/bundle-level ref table in a single SQL statement against one connection. Ref tables aren't promoted until Phase 3/6, so once the catalog moves to a different physical SQLite file, a naive `wstore.skill_get(...)` → `identity_store.skill_get(...)` call-site swap is correct for the genuinely single-table methods but silently wrong for every method that needs the join.

§5.4 documents the resolution (an explicit `catalog: &Store` parameter, composing two queries in Rust instead of one SQL join — the same shape `managed_bind_bundle`/`managed_upsert_unique_for_bundle` already use for bundle-existence checks) and two accepted interim gaps, both low-severity and with a documented path to closure once Phase 6 lands:

- Loss of single-transaction atomicity for owner-private skill/MCP-server name uniqueness (the global-uniqueness path is unaffected — Phase 2a's unique indexes arbitrate at the DB level regardless of transaction boundaries).
- Cross-channel dangling refs on catalog delete, degrading gracefully (a dangling ref reads as `None`, which every read path already handles) rather than corrupting anything.

No code changes in this PR — implementation follows in a separate PR against this design once it's reviewed.

### DIFF
--- a/docs/specs/SPEC_DURABLE_BINDINGS_2026_09_10.md
+++ b/docs/specs/SPEC_DURABLE_BINDINGS_2026_09_10.md
@@ -7,7 +7,9 @@ from an initial registry-only framing that Codex correctly flagged as
 rejecting real, currently-bindable agents — PR #3179), which in turn requires
 Phase 5 to add explicit agent-delete cleanup of identity-store refs (no
 cross-database `ON DELETE CASCADE`) — see §7 Phase 5 and §8 item 2 for the
-full mechanism. Phase 2 in progress.
+full mechanism. Phase 2 in progress: schema (#3180) and the carry migration
+(#3181) landed; the application-layer read/write redirect (§5.4) is the
+remaining piece.
 **Date:** 2026-09-10
 **Verified against:** `94d9c6c1c` (code and live on-disk data, not spec prose)
 **Follows:** `SPEC_INSTRUCTION_AND_MEMORY_PORTABILITY_2026_09_09.md` §3.4a, which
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

AgentA-asaf:
All three findings verified against the actual code before accepting — replies below, fix design pushed in ae066d6c1.

**P1 (local FK breaks binding new rows):** Confirmed. `configure_and_migrate` sets `PRAGMA foreign_keys=ON` in production (`storage/store.rs`), and `db_agent_skills_ref`/`db_agent_mcp_ref`/`db_bundle_skills_ref`/`db_bundle_mcp_ref` all FK `skill_id`/`mcp_id` to `wstore`'s own local `db_skills`/`db_mcp_servers` (`storage/migrations.rs` lines 717-770). This is a real, hard write-path bug my original §5.4 missed entirely — splitting reads doesn't touch it. Design added: a new migration drops the catalog-pointing FK from all four ref tables (SQLite has no `DROP CONSTRAINT`, so it's a rebuild-and-swap under `PRAGMA foreign_keys=OFF`), keeping the `agent_id → db_agents` FK where it exists. Existence check moves to the application layer via a new `catalog: &Store` param on `managed_bind_agent`, mirroring `managed_bind_bundle`'s existing `id_store` pattern — this is exactly the same shape `db_bundle_skills_ref`/`db_bundle_mcp_ref` already use for `db_bundles` (their own doc comment: "Existing is checked at the application layer instead"), just applied one layer over.

**P2 (unique index doesn't match the app's own invariant):** Confirmed, and confirmed this is a genuinely NEW exposure rather than a preexisting mislabeled one — before Phase 2, `db_skills` was per-channel, so the only concurrency `managed_upsert_unique_global` ever saw was same-process, already fully serialized by `Store`'s `Mutex<Connection>`; the index gap was inert. Once the catalog is shared cross-process, `conn.transaction()`'s default DEFERRED behavior (no `Immediate` override anywhere in `managed.rs`) genuinely allows two channels' dup-check SELECTs to both pass before either INSERT takes the write lock, for a same-name-different-`skill_type` pair — which `idx_ids_skills_global_name_type`'s `(name, skill_type)` key has nothing to say about, contrary to the app's own name-only dup-check. MCP servers don't have this mismatch (both sides are already name-only). Design added: narrow the index to `(name) WHERE is_global = 1`, with a defensive collapse-duplicates step (mirroring `m0031`'s own dedup-and-rewrite-refs shape, this time `Global`-scoped) for any store where `m0031`'s `(name, skill_type)`-keyed carry already produced a pair the tighter index would reject.

**P2 (insert-plus-bind atomicity):** Agreed the original "accepted gap" framing undersold it — a committed catalog insert followed by a failed ref-bind left a permanent, undiscoverable orphan, not just a race window. Design added: `managed_upsert_unique` now performs a best-effort compensating delete of the just-inserted catalog row if the ref-bind fails, so a failed RPC and durable state agree. The narrower same-instant same-owner race stays accepted as interim (low-severity, single-writer UI flow, full closure needs Phase 6).

Full revised design is in §5.4 of the spec, ae066d6c1. Implementation is a separate follow-up PR once this lands.

@codex review

PR #3183, opened 2026-09-10
feat(srv): redirect skill/MCP-server catalog reads and writes to identity_store

AgentA-asaf:
Follow-up to #3182 (design) — implements Phase 2's application-layer read/write redirect per `SPEC_DURABLE_BINDINGS_2026_09_10.md` §5.4, including the two additional migrations Codex's review of #3182 surfaced as hard prerequisites (both re-verified against the code independently before implementing):

- **`m0032_drop_catalog_fk_from_ref_tables`** — the four skill/MCP ref tables' `skill_id`/`mcp_id` FK targeted `wstore`'s own local catalog copy. Once that copy stops gaining new rows (writes redirect to `identity_store`), binding any post-redirect catalog row would hard-fail with `FOREIGN KEY constraint failed` under `PRAGMA foreign_keys=ON`. Rebuilds all four tables without that FK (`agent_id → db_agents` kept), mirroring `db_bundle_skills_ref`'s existing no-FK-to-`db_bundles` precedent. Existence is now checked explicitly in `managed_bind_agent`/`managed_bind_bundle`.
- **`m0033_narrow_skill_global_uniqueness_index`** — the Phase 2a unique index was `(name, skill_type)`, narrower than `skill_upsert_unique_global`'s own application-level invariant (`name` alone). Once the catalog is shared cross-process, a same-name-different-type race could slip past both layers. Narrows the index to `name`-only, with a defensive collapse-duplicates pass for any store where `m0031`'s own `(name, skill_type)`-keyed dedup already produced a colliding pair.

`storage/managed.rs`: every method that used to join the catalog table with a ref table in one SQL statement now takes an explicit `catalog: &Store` parameter and composes two queries in Rust instead (`managed_list`, `managed_list_global`, `managed_list_global_for_agent`, `managed_delete`, `managed_is_accessible_to`, `managed_upsert_unique`, `managed_union_bundle_refs`). `managed_upsert_unique` performs a best-effort compensating delete of the catalog row if the subsequent ref-bind fails, so a failed RPC and durable state agree instead of leaving a permanent orphan. Careful attention to `Mutex<Connection>` non-reentrancy where `self`/`catalog` may alias (test fixtures, degraded-mode fallback) — each cross-store helper acquires and releases its own lock rather than holding one while calling into the other.

`skills.rs`/`mcp_servers.rs` thread the new parameter through their public wrappers; `skill.rs`/`mcp.rs`/`bundle.rs`/`agent_open.rs`/`agent_handlers/skills.rs` redirect call sites to `state.identity_store`. `m0015`/`m0016`/`m0030`/`m0031` needed small compile-only fixes (passing `wstore` as its own `catalog` — they predate this redirect or specifically operate on `wstore`'s own local mirror, verified individually, not a semantic change).

3328 tests pass (up from 3304 baseline), including new cross-store composition tests in `managed.rs` that use two genuinely separate `Store` instances rather than the shared `test_state()` fixture, which would hide a wrong-store bug behind "querying the same store twice trivially works."

Known, documented nit (not fixed here, low risk): `bundle_export_impl`/`bundle_export_for_agent_impl`'s parameter order is `(id_store, wstore, identity_store)` while every other multi-store function in `bundle.rs` uses `(id_store, identity_store, wstore)` — cosmetic only, both are internally consistent with their own call sites and fully tested.

### DIFF
--- a/.changesets/1789075390-feat-srv-redirect-skill-mcp-server-catalog-reads-and-writes--haz0.md
+++ b/.changesets/1789075390-feat-srv-redirect-skill-mcp-server-catalog-reads-and-writes--haz0.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+feat(srv): redirect skill/MCP-server catalog reads and writes to the identity store
--- a/agentmux-srv/src/backend/mcp_seed.rs
+++ b/agentmux-srv/src/backend/mcp_seed.rs
@@ -123,7 +123,9 @@ fn reject_duplicate_names(manifest: &[StarterMcpServer]) -> Result<(), StoreErro
 pub(crate) fn any_starter_mcp_server_name_exists(wstore: &Arc<Store>) -> Result<bool, StoreError> {
     let manifest: Vec<StarterMcpServer> = serde_json::from_str(STARTER_MCP_SERVERS_JSON)
         .map_err(|e| StoreError::Other(format!("mcp server seed: parse manifest: {e}")))?;
-    let existing = wstore.mcp_server_list_global()?;
+    // Pre-dates Phase 2 of SPEC_DURABLE_BINDINGS_2026_09_10.md's catalog
+    // redirect — see skill_seed.rs's identical note.
+    let existing = wstore.mcp_server_list_global(wstore)?;
     Ok(manifest
         .iter()
         .any(|entry| existing.iter().any(|item| item.server.name == entry.name)))
@@ -170,7 +172,7 @@ pub(crate) fn seed_starter_mcp_servers(wstore: &Arc<Store>) -> Result<McpServerS
         };
         if let Err(e) = wstore.mcp_server_upsert_unique_global(&server) {
             for id in &inserted_ids {
-                if let Err(cleanup_err) = wstore.mcp_server_delete(id) {
+                if let Err(cleanup_err) = wstore.mcp_server_delete(wstore, id) {
                     tracing::error!(
                         "mcp server seed: cleanup after partial failure could not remove {id}: {cleanup_err}"
                     );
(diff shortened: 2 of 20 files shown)


a5af:
@codex review

AgentA-asaf:
All three findings verified against the actual code before accepting — fix pushed in 7946b37ef.

**P1 (m0031's dedup fallback stale under the name-only index):** Confirmed. `carry_skills`'s non-starter-global branch re-queried via `skill_find_global_by_name_and_type`, which m0033 (this same PR) makes semantically wrong — a same-name-different-`skill_type` pair now collides on the DB's actual constraint (name alone) but the re-query wouldn't find the winner, falling through to the unrelated-collision branch and minting a fresh id whose own insert then ALSO silently no-ops (`INSERT OR IGNORE`) against the same name constraint. Traced it through: this strands a local ref rewritten to an id the identity store never holds, exactly as you described — the skill disappears once reads redirect there. Fixed with a new name-only lookup (`skill_find_global_by_name`, mirroring the MCP-server equivalent, which never had this mismatch since its index was always name-only), plus a defensive re-check after the fresh-id insert for the residual TOCTOU case (a genuine concurrent race landing between the re-query and the fresh-id insert). New test: `two_channels_carrying_the_same_name_under_different_skill_types_still_converge`.

**P1 (managed_delete ordering can strand a row):** Confirmed — traced the failure mode exactly: ref-purge-then-catalog-delete meant a catalog-delete failure left the row's refs already gone, making a private row permanently inaccessible (not global, no ref) and un-retryable (the RPC's own pre-check needs the ref that was just destroyed). Reordered to catalog-delete-first: a failure there now returns before any ref mutation happens, so the row stays exactly as accessible/retryable as before the call. New test: `skill_delete_leaves_refs_intact_when_the_catalog_delete_itself_fails`, which forces a real SQL error (drops `db_skills`, the same technique the existing compensating-delete test uses) and asserts the ref row is still there afterward.

**P2 (rewrite_ref_id's UPDATE OR IGNORE leaves a dangling ref):** Confirmed — `skill_rewrite_ref_id`'s own doc comment had already accepted this as a "vanishingly unlikely, harmless no-op" case, but m0033 turns it harmful by deleting the loser catalog row right after the rewrite. Changed both `skill_rewrite_ref_id` and `mcp_server_rewrite_ref_id` (mirrored for consistency, even though m0033 is skills-only) from a blind `UPDATE` to insert-the-survivor-then-delete-the-loser, which resolves correctly whether or not the owner already held the survivor. Two new tests covering both the agent-level and bundle-level ref tables, plus one confirming the ordinary rename case is unchanged.

5 new regression tests, each reproducing the exact failure shape described rather than a weaker version of it. Full suite: 3333 passed, 0 failed (up from 3328).

@codex review

PR #3187, opened 2026-09-11
Camper@narko: fix(pane-tab-strip): stop terminal/agent tab pills flashing, keep terminal tabs mounted across a switch

GenericAgentX-asaf:
## Summary

Follow-up to the pane-tab-switch chrome-stability work (#3136/#3151/#3157). With header/tab-strip already stable, two smaller flickers remained:

- **Tab pill flash on switch**: `termTabs`/`stackTabs` rebuilt a brand-new object per tab pill on every switch (they rerun on any layout-tree bump, not just add/remove/rename), so `<For>` saw an entirely new list each time and tore down + recreated every pill's DOM node — invisible on plain inactive tabs, but visible as a flash on the active one's highlighted background. Fixed by caching tab objects per blockId and reusing them when nothing about that tab actually changed.
- **Terminal content flash on switch**: `pane-leaf-chrome.tsx`'s inner `<Key>` fully remounted `<Block>` (new `TermViewModel`, new xterm.js instance) on every terminal tab switch, showing `block.tsx`'s ready-gate `BrainSpinner` cross-fade briefly. Editor's own file tabs never remount at all (one persistent block, no remount), which is why editor felt smooth by comparison. Terminal tabs now all mount once and stay mounted — a switch just toggles `visibility` on an absolutely-positioned wrapper instead of tearing down and rebuilding. Each tab gets its own private ViewModel slot (since multiple `<Block>`s are now live simultaneously and a shared slot would otherwise track whichever tab mounted most recently, not whichever is active), and the tab-switch handler explicitly hands off keyboard focus to the target tab since its own mount-time focus grab no longer fires on a repeat switch. Scoped to terminal only; agent tabs are unchanged.

## Test plan

- [x] Full frontend test suite passes (153 files / 2130 tests)
- [x] New tests: tab-pill identity caching regression coverage; `pane-leaf-chrome.test.tsx` keep-alive coverage (both stack members stay mounted across a switch, a newly-pushed member mounts immediately)
- [x] Manually verified in `task dev`: tab-pill flash gone, terminal switch flicker gone
- [ ] Reviewer: please double check the per-tab ViewModel slot isolation (`pane-leaf-chrome.tsx`'s `viewModelSlotFor`) and the explicit focus hand-off in `term.tsx`'s `handleTermTabSwitch`

### DIFF
--- a/.changesets/1789087612-fix-term-keep-alive-terminal-tabs-no-longer-leak-input-zoom--uev6.md
+++ b/.changesets/1789087612-fix-term-keep-alive-terminal-tabs-no-longer-leak-input-zoom--uev6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(term): keep-alive terminal tabs no longer leak input/zoom into hidden dormant tabs
--- a/frontend/app/store/block-component-registry.test.ts
+++ b/frontend/app/store/block-component-registry.test.ts
@@ -2,71 +2,85 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /**
- * Regression coverage for Codex P1/P2 on PR #3187: once a pane can keep
- * multiple stack members' `<Block>`s mounted simultaneously (terminal
- * keep-alive, `pane-leaf-chrome.tsx`), every registered blockId is no
- * longer automatically "the one currently-visible pane" — a dormant,
+ * Regression coverage for Codex P1/P2 + ReAgent P1 on PR #3187: once a pane
+ * can keep multiple stack members' `<Block>`s mounted simultaneously
+ * (terminal keep-alive, `pane-leaf-chrome.tsx`), every registered blockId
+ * is no longer automatically "the one currently-visible pane" — a dormant,
  * hidden stack member registers here too. `getAllBlockComponentModels`/
- * `getAllBlockComponentModelEntries` must still expose only the ACTIVE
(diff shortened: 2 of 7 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Addressed all three findings:

- **Codex P1** (multi-input broadcast reaching hidden terminal tabs) and **Codex P2** (all-panes zoom stepping hidden tabs) — same root cause: the global block-component registry now holds dormant, kept-alive tabs that used to be unmounted and absent. Fixed once at the registry's `getAllBlockComponentModels`/`getAllBlockComponentModelEntries` accessors (filtered to each leaf's currently-active stack member) rather than patching each consumer separately — also fixes a related, unflagged instance of the same bug in `keymodel.ts`'s multi-input-eligible terminal count. New regression tests in `frontend/app/store/block-component-registry.test.ts`.
- **ReAgent P1** (missing changeset) — added.

GenericAgentX-asaf:
Fixed the ReAgent P1 correctness bug: replaced the tab-scoped \`getNodeByBlockId\` lookup with an explicit dormancy registry (\`setKeepAliveBlockDormant\`) that \`pane-leaf-chrome.tsx\` maintains directly as each stack member's visibility toggles — no layout/tab resolution involved, so a background tab's panes are never affected. Rewrote the registry tests to match; full suite (264 files / 3922 tests) and typecheck still pass.

PR #3188, opened 2026-09-11
Clare@starpower: fix(mcp): SendMessage reports whether a message was delivered or only queued

AgentO-asaf:
## The bug

`success: true` spans two very different outcomes, and the old return string conflated them — making agent-to-agent delivery **unfalsifiable from the sender's side**.

Verified against a running srv, not inferred. Sending to a name that exists on no machine anywhere:

```
SendMessage(to="definitely-not-a-real-agent-xyz123")
  -> "Message sent to definitely-not-a-real-agent-xyz123"
```

Server log for that same call:

```
"cloud relay: queued for WAN delivery"  target=definitely-not-a-real-agent-xyz123
```

**Byte-identical treatment to a real remote agent.** Meanwhile a same-host target logs:

```
"inject: structured delivery to non-PTY controller (mid-turn steer)"  target_agent=AgentO
```

…and AgentO confirmed receipt by replying (`DELIVERY=host TRUST=host-verified`).

So all three outcomes returned the same string:

| Outcome | Old return |
|---|---|
| injected into a live conversation | `Message sent to X` |
| queued for a real but unreachable agent | `Message sent to X` |
| queued for a name nobody has ever claimed | `Message sent to X` |

This is very likely why *"jekt messages not coming through"* has been so hard to pin down — the sender gets a confirmation, reasonably assumes delivery, and nothing contradicts it.

## The fix needs no srv change

The response **already carries the distinction**; the MCP layer was discarding it.

| Path | `block_id` |
|---|---|
| local delivery (`backend/reactive/handler.rs:947`) | `Some(block_id)` — the receiving block |
| cloud relay (`server/reactive.rs:1357`) | `None` — no receiver has seen it |

`try_cloud_relay`'s own doc comment already says **"Queued is not delivered."** The information was there and unused.

Now:

- `Delivered to X — injected into their conversation.`
- `QUEUED for X via the cloud relay — NOT yet delivered. …You get this same result for an agent name that does not exist anywhere, so check the spelling against DiscoverAgents if you expected local delivery.`

The nonexistent-name case is called out explicitly so it's actionable rather than mysterious. Tool description updated to match, since agents read that to decide whether to trust the result.

## Why this matters beyond cosmetics

`DiscoverAgents` cannot tell you whether a remote agent exists — the WAN entry lists only *this* host's subscribers, and the relay exposes no remote directory (documented behaviour). Combined with unconditional success reporting, there was no way for a sender to establish delivery at all. This restores that.

## Related but separate

agentmuxai/agentmux-cloud#71 (merged today) fixed `merge.ts` / `ci-failure.ts` silently dropping notifications for agents on shared GitHub accounts. That's the GitHub→jekt path; this is agent→agent.

### DIFF
--- a/.changesets/1789087893-fix-mcp-sendmessage-reports-whether-a-message-was-delivered--mzmm.md
+++ b/.changesets/1789087893-fix-mcp-sendmessage-reports-whether-a-message-was-delivered--mzmm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(mcp): SendMessage reports whether a message was delivered or only queued
--- a/agentmux-mcp/src/main.rs
+++ b/agentmux-mcp/src/main.rs
@@ -1153,7 +1153,34 @@ async fn call_tool(
                 .map_err(|e| anyhow::anyhow!("response parse failed: {e}"))?;
 
             if result.get("success").and_then(|v| v.as_bool()) == Some(true) {
-                Ok(format!("Message sent to {to}"))
+                // `success: true` spans two very different outcomes and the old
+                // message conflated them, which made agent-to-agent delivery
+                // unfalsifiable from the sender side: a name that exists on no
+                // machine anywhere reported exactly the same string as a
+                // message injected into a live conversation. Verified against a
+                // running srv — `SendMessage(to="definitely-not-a-real-agent-xyz123")`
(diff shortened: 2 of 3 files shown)


a5af:
@codex review

PR #3189, opened 2026-09-11
Korp@claudius: docs(cef): correct use_static_angle finding — platform-specific, not blanket

GenericAgentX-asaf:
## Summary

Small, doc-only correction to `docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md`, prompted by a real cross-host jekt exchange with Clare (macOS, host starpower) — first confirmed working end-to-end delivery for that path (issue #3108).

My earlier `use_static_angle=false` finding (from #3172) was written as if Linux/macOS need "the equivalent line added" to reach the same fix. Clare checked directly on macOS 152 and found the opposite: macOS already statically links the real ANGLE implementation into the framework binary (confirmed via export symbols), so `args-darwin.gn`'s existing config — no `use_static_angle` override at all — is already correct. Setting it to `false` there would undo a working config, not fix a broken one.

Corrected both the top status block and the Phase D callout box to say the value is platform-specific and must be verified per-platform (export-table check), not copied from Windows.

## Test plan

- [ ] CI green on this PR

### DIFF
--- a/docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md
+++ b/docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md
@@ -164,9 +164,18 @@ the old 148 build's separate `libGLESv2.dylib`, meaning macOS's existing
 already correct and working, not silently broken the way Windows was.
 **Setting `use_static_angle=false` on macOS would undo a working
 configuration, not fix a broken one — do not copy the Windows fix into
-`args-darwin.gn`.** Linux has not been checked yet as of this writing;
-verify by export table on the actual built `.so`, the same method both
-Windows and macOS used — a clean compile gives no signal either way on
+`args-darwin.gn`.** Linux has not been checked yet as of this writing.
+**Codex P2 on #3189: checking only `libGLESv2.so`/`libEGL.so` is not
+sufficient** — if Linux also statically links ANGLE (like macOS, into
+`libcef.so` rather than a separate library), those files legitimately
+won't export real symbols even in a correctly configured static build,
+and concluding "broken, needs `use_static_angle=false`" from that alone
+would repeat the macOS mistake in the opposite direction. Check both
+candidates — real ANGLE symbols in `libGLESv2.so`/`libEGL.so` (dynamic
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed — good catch. Both the top status block and the Phase D callout now tell whoever checks Linux to look for real ANGLE symbols in *either* libGLESv2.so/libEGL.so (dynamic case) *or* libcef.so itself (static case, same symbol families Clare used for macOS), and only call it broken if neither has them. Checking only libGLESv2.so would have reproduced the exact mistake this correction exists to fix, just pointed the other way.

GenericAgentX-asaf:
This finding doesn't match the actual reviewed commit's content. Checked directly:

\`\`\`
$ git show be4422547:docs/specs/SPEC_CEF_MILESTONE_UPGRADE_148_TO_152_2026_09_07.md | sed -n '551,567p'
> P2 on #3189, checking only `libGLESv2.so` is not sufficient on its own,
> for exactly the reason the macOS correction above describes.** If Linux
> also turns out to statically link ANGLE (like macOS, into `libcef.so`
> rather than a separate library), `libGLESv2.so` legitimately won't
> export `glGetString` even in a CORRECTLY configured static build — that
> file is a placeholder by design in that configuration, not evidence of
> brokenness. Checking only `libGLESv2.so` and concluding "must set
> `use_static_angle=false`" on a false negative would repeat the exact
> macOS mistake this correction exists to prevent, in the other direction.
> Check BOTH candidates before concluding anything: `glGetString`/
> `eglGetProcAddress` in `libGLESv2.so`/`libEGL.so` (the dynamic-ANGLE
> case, like Windows) OR the same real ANGLE symbols
> (`angle::`/`rx::`-prefixed, same families Clare checked on macOS) inside
> `libcef.so` itself (the static-ANGLE case, like macOS) — only treat it
> as actually broken if NEITHER binary has real ANGLE code...
\`\`\`

This is exactly the "check both candidates" language the review is asking for — it was already added in response to Codex's original comment (my earlier commit + reply on this PR), and it's present at the exact commit (be4422547) this review claims to be against. The same fix also exists in the Phase D callout box further down. Not making a redundant edit since there's nothing left to fix here — flagging as a false positive rather than silently re-applying.

PR #3190, opened 2026-09-11
fix(agent-pane): PageUp/PageDown in composer no longer scrolls the pane off screen

a5af:
## Summary
- Root cause: the composer `<textarea>` in `AgentFooter.tsx` never handled `PageUp`/`PageDown`. It isn't nested inside the scrollable chat history (`.agent-document`, `overflow-y: auto`) — it's a flex sibling of it, and the whole pane wrapper (`.agent-view`) is `overflow: hidden`. The browser's default Page Up/Down behavior walked past the pane looking for a scrollable ancestor and found one outside `.agent-view` instead, displacing the entire pane.
- Fix: swallow `PageUp`/`PageDown` in the composer's `handleKeyDown` with `preventDefault()`.
- Added a regression test asserting `defaultPrevented` for both keys.

## Test plan
- [x] `npx vitest run app/view/agent/components/AgentFooter.test.tsx` — 25/25 pass (including the new PageUp/PageDown test)

### DIFF
--- a/.changesets/1789091270-fix-agent-pane-pageup-pagedown-in-composer-no-longer-scrolls-zn34.md
+++ b/.changesets/1789091270-fix-agent-pane-pageup-pagedown-in-composer-no-longer-scrolls-zn34.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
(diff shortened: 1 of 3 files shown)


a5af:
@codex review

PR #3194, opened 2026-09-11
feat(mcp): PtyShell attaches to the pane's real visible shell, with a lease-based lock

Agent3-asaf:
## Summary

Follow-up to #3177 (repo owner's explicit request during this session's scoping conversation): agents need to drive the **same** shell a human sees in the composer drawer — not an independent, invisible one — with the human locked out of their own keyboard only while the agent is actively working, and bound "only when necessary," never held indefinitely.

- **`create()` now reuses "this pane's one shell"** (`term:shellsubblockid`, the same meta key `agent-view.tsx` already persists for the composer drawer) instead of always minting an independent block. Whichever side (agent or human-opened drawer) gets there first, the other attaches to it — live output and scrollback included.
- **Ownership check reframed**: from "did `PtyShell` create this block" to "is this block a sub-block of the calling agent's own pane" (`is_owned_by_agent`, via `parentoref` against a trusted `agent_block_id` now on every `PtyShell*` request). More correct than the original check, not just adapted for reuse — still blocks an arbitrary `block_id` discovered via `Layout`, while also correctly covering a shell the human created first.
- **Locking is a lease, not an explicit lock/unlock.** Every successful `PtyShellInput`/`PtyShellResize` stamps a ~4s expiry (`term:agentlockuntil`) the frontend gates human keyboard input on, refreshed on each write. An explicit release call is one the agent could simply never make (crash, error) — a self-expiring window needs no server-side timer to leak or get stuck. `AgentShellSubblock.tsx` drops human keystrokes while locked, shows a small "Agent is using this shell" badge, and re-evaluates on its own clock.
- **`PtyShellStop` no longer kills anything.** It now releases the lock immediately instead of waiting for it to lapse, but doesn't touch the process or the block — this shell is shared and pane-scoped now, not private to this tool.

## Test plan

- [x] All PR #3177 tests updated for the new request/response shape (`agent_block_id` added, `stopped` → `released`) and passing.
- [x] 3 new deterministic backend tests: reuse (attaches to an existing human-created shell instead of spawning a second one), cross-pane rejection (with a positive control proving the legitimate owner still succeeds), lock set-by-input + release-by-stop.
- [x] 3 new frontend tests (`AgentShellSubblock.test.tsx`): drops keystrokes while locked, forwards them once the lock has passed, badge mounts/unmounts with live lock-meta changes.
- [x] Real live-PTY round trip (`#[ignore]`'d, run manually) re-verified: lock gets set on a real successful write, `stop` releases it without killing the shell (`running: true` afterward, confirmed against real cmd.exe).
- [x] Full `agentmux-srv` suite (3336 tests) and frontend `agent-view` suite (1787 tests) green, no regressions.
- [x] `cargo check` (srv/mcp/common) and `tsc --noEmit` clean — zero new errors/warnings.

## Deferred (separate follow-up, not in this PR)

Moving `AgentControlBar`'s Session/History actions out of the shell-toggle composer-details panel into `AgentStashModal` — also raised in this session's scoping conversation, but an unrelated UI reorg not folded into an already-large change.

Full design: `docs/specs/SPEC_AGENT_INTERACTIVE_PTY_SHELL_API_2026_09_10.md` §10.

### DIFF
--- a/.changesets/1789092935-feat-mcp-ptyshell-attaches-to-the-pane-s-real-visible-shell--lxk0.md
+++ b/.changesets/1789092935-feat-mcp-ptyshell-attaches-to-the-pane-s-real-visible-shell--lxk0.md
@@ -1,0 +1,5 @@
+---
+type: minor
+---
+
+feat(mcp): PtyShell attaches to the pane's real visible shell, with a lease-based lock instead of destroying it on stop
--- a/.changesets/1789094219-fix-mcp-ptyshell-enforce-the-agent-lock-server-side-fix-stal-vda3.md
+++ b/.changesets/1789094219-fix-mcp-ptyshell-enforce-the-agent-lock-server-side-fix-stal-vda3.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(mcp): PtyShell — enforce the agent lock server-side, fix stale-handshake injection on reused shells, atomic first-create claim
--- a/.changesets/1789094899-fix-mcp-ptyshell-lock-the-shell-before-writing-not-after-per-4nqm.md
+++ b/.changesets/1789094899-fix-mcp-ptyshell-lock-the-shell-before-writing-not-after-per-4nqm.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(mcp): PtyShell — lock the shell before writing not after, persist the fallback block when the atomic claim's winner has vanished
--- a/.changesets/1789119727-fix-mcp-ptyshell-create-s-spawn-failure-rollback-now-also-cl-mwng.md
+++ b/.changesets/1789119727-fix-mcp-ptyshell-create-s-spawn-failure-rollback-now-also-cl-mwng.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(mcp): PtyShell create's spawn-failure rollback now also clears the parent's stale shellsubblockid pointer
--- a/agentmux-common/src/api_types.rs
+++ b/agentmux-common/src/api_types.rs
@@ -130,6 +130,15 @@ pub struct ShellStatusResponse {
 // (agent -> agentmux-mcp -> agentmux-srv), with no dependency on any window
 // being open, focused, or even rendering the shell's `term` sub-block — the
 // same posture the existing `Shell`/`ShellInput` family already has.
+//
+// `create` attaches to (creating if needed) the SAME shell a human's
+// composer-drawer session uses, not an independent, invisible one —
+// whichever side gets there first, the other joins. While the agent is
+// actively writing (`input`/`resize`), the human's own keyboard input to
+// that shell is locked out, for a short, self-expiring window
+// (`AGENT_LOCK_WINDOW_MS`) rather than an explicit lock/unlock the agent
+// could fail to release — "bound to the shell only when necessary," per
+// the repo owner's own framing of the requirement.
 
 /// `POST /api/v1/ptyshell/create`
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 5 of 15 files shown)


a5af:
@codex review

Agent3-asaf:
All three confirmed real and fixed in 4a3766e:

1. **Lock race** — confirmed the frontend's gate is eventually-consistent (WS push), not synchronous. Fixed at the actual authoritative point: `controllerinput` (server/websocket.rs) now checks `term:agentlockuntil` itself before forwarding to `send_input`. New test `controllerinput_is_dropped_while_an_agent_lock_is_active` proves it, with an unlocked-target control case proving the call path isn't just trivially always-dropped.
2. **Stale-handshake injection** — confirmed live: the `term` file is append-only for a block's whole lifetime (`FILE_OP_APPEND`), so my original whole-file scan absolutely would (and did, once traced through) match a reused shell's own old, already-answered query and inject `ESC[1;1R` into a live session. Fixed with a byte-offset baseline taken right before `resync_controller` runs — only bytes appended after it are ever inspected. Re-ran the live-PTY test against a genuinely reused shell to confirm no more spurious injection.
3. **First-create race** — added an atomic claim inside a `with_tx` transaction (re-check the pointer, insert+link+set only if still unset, loser pivots to the winner's shell). Documented explicitly that this does NOT close the equivalent race against the frontend's own `createsubblock` path — agreed that's a real, separate gap, not silently left unmentioned; full writeup in the spec's §10.6a.

Thanks for catching #1 and #2 especially — both were genuine "looks safe, isn't" bugs, not just gaps.

Agent3-asaf:
Both round-2 findings addressed in 928aeba4f:

- **P1 (lock ordering)** — `lock_shell_for_agent` now runs *before* `blockcontroller::send_input` in both `PtyShellInput` and `PtyShellResize`, unconditional on the write's outcome (previously only set on success, after the write). Shrinks the remaining race to the (much smaller, and left open on purpose — closing it fully needs a shared mutex, not a meta flag) window between `controllerinput`'s own lock check and the first lock-set for a concurrent message.
- **P2 (dangling-block fallback)** — the "winner's block vanished" branch in `handle_pty_shell_create`'s atomic claim now actually persists the locally-built block (`insert` + link into the parent's `subblockids` + set `term:shellsubblockid` + broadcast), instead of falling through with a never-inserted phantom `Block`.

Verified: full `agentmux-srv` suite (3337 tests) green, plus the `#[ignore]`'d live-PTY test re-run explicitly. `ptyshell_input_locks_the_shell_and_stop_releases_it_early` updated to assert the lock is set by the real HTTP call itself rather than a manual simulation, proving the new ordering. Spec updated at `docs/specs/SPEC_AGENT_INTERACTIVE_PTY_SHELL_API_2026_09_10.md` §10.6b.

Agent3-asaf:
Round-3 finding fixed in 5b5b5dbf1:

- **P1 (stale parent pointer on rollback)** — `handle_pty_shell_create`'s spawn-failure rollback now also clears `term:shellsubblockid` on the parent (in addition to unlinking `subblockids`) when it still points at the block being rolled back, and broadcasts the update so an already-open drawer sees the correction.

Also note on the CI hangs: the windows-latest leg genuinely hung 3x in a row on the previous commit (928aeba4f), but I traced it to a repo-wide CI infra issue unrelated to this PR's code — confirmed by pulling in main's `fix(ci): cap ci-pr.yml's required job at 20 minutes (#3201)` and the concurrency-group fix, both merged into this branch. My local Windows run of the full suite (both `--test-threads=8` and `--test-threads=1`, matching CI) passed cleanly in under a minute both times, and the diff in question doesn't touch the code path the stuck test exercises — so this wasn't a deadlock in the PtyShell changes.

Full `agentmux-srv` suite (3342 tests) green locally. No dedicated test for this round's fix — see spec §10.6c for why (tried triggering a real synchronous spawn failure via an invalid `cmd:cwd`; `resync_controller` doesn't actually fail synchronously for that case, so it didn't reproduce the branch under test).

PR #3193, opened 2026-09-11
feat(mcp): add ClosePane — close your own pane, or any pane by block_id (fleet tier)

a5af:
## Summary

Phase 1+2 of `docs/specs/SPEC_AGENT_PANE_LIFECYCLE_CONTROL_2026_09_10.md` — the capability this session's stuck-pane investigation needed and didn't have: there was no MCP tool that could close or reset a pane stuck in a broken render state, own or another agent's.

- **Own-pane** (`ClosePane` with no args): identity resolved via the same signed mechanism `UIClick`/`UIQuery`/`UIScreenshot` already use (`verified_block_id`, made `pub(crate)` for reuse) — never a client-supplied `block_id`, per the real vulnerability that mechanism exists to close (PR #2662).
- **Cross-pane** (`ClosePane(block_id: "...")`): fleet tier, same posture as `FleetBulkStop` — no ownership check on the target (that's the point: this needs to work on a pane whose own agent might not be able to cooperate). But unlike `FleetBulkStop` today, the caller is never anonymous: `log_fleet_action_audit` gains an optional `reason` param, and both new call sites pass the caller's *verified* agent_id as `source_agent` instead of `None` — closing a real gap Codex found in `FleetBulkStop`'s existing audit trail while adding this.
- New: `POST /api/v1/agent/pane/close`, `ClosePaneRequest`/`ClosePaneResponse` (`agentmux-common`), `handle_close_pane` (`agentmux-srv/server/app_api/pane.rs`, reuses `sagas::delete_block::run` — no saga changes), `ClosePane` MCP tool.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv close_pane_tests` — 3/3 pass (own-pane close, cross-pane close with verified audit attribution, rejects an invalid signature)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv fleet` / `delete_block` — no regressions (21 + 11 pass)
- [x] `cargo test -p agentmux-mcp` — 26/26 pass (tool count assertion bumped 43→44)
- [x] `cargo check -p agentmux-common -p agentmux-srv -p agentmux-mcp` — clean
- [ ] Manual verification against a real running instance (closing another agent's stuck pane) — not yet done in this PR; recommend before merge given the destructive nature of the cross-pane path

[...]

PR #3196, opened 2026-09-11
feat(mcp): add ClosePane — close your own pane, or any pane by block_id (fleet tier)

Agent5-asaf:
## Summary

Phase 1+2 of `docs/specs/SPEC_AGENT_PANE_LIFECYCLE_CONTROL_2026_09_10.md` (#3195) — the capability this session's stuck-pane investigation needed and didn't have: there was no MCP tool that could close or reset a pane stuck in a broken render state, own or another agent's.

- **Own-pane** (`ClosePane` with no args): identity resolved via the same signed mechanism `UIClick`/`UIQuery`/`UIScreenshot` already use (`verified_block_id`, made `pub(crate)` for reuse) — never a client-supplied `block_id`, per the real vulnerability that mechanism exists to close (PR #2662).
- **Cross-pane** (`ClosePane(block_id: "...")`): fleet tier, same posture as `FleetBulkStop` — no ownership check on the target (that's the point: this needs to work on a pane whose own agent might not be able to cooperate). But unlike `FleetBulkStop` today, the caller is never anonymous: `log_fleet_action_audit` gains an optional `reason` param, and both new call sites pass the caller's *verified* agent_id as `source_agent` instead of `None` — closing a real gap Codex found in `FleetBulkStop`'s existing audit trail while adding this.
- New: `POST /api/v1/agent/pane/close`, `ClosePaneRequest`/`ClosePaneResponse` (`agentmux-common`), `handle_close_pane` (`agentmux-srv/server/app_api/pane.rs`, reuses `sagas::delete_block::run` — no saga changes), `ClosePane` MCP tool.
- Codex P2 fix already included: target agent name is resolved *before* deleting the block, not after (a running agent's own exit-triggered `unregister_block` can otherwise race ahead of a post-delete lookup).
- Warden Audit UI fix included: the caller-supplied `reason` on a cross-pane close never rendered (the audit tab only showed `reason` when `outcome` was set, a condition written for Supervisor-nudge entries) — fixed and covered by a new test.

## Known limitation, not fixed in this PR

**`ClosePane` is not `blockStack`-aware** — found during spec review (codex P1 on #3195), confirmed to affect this PR's actual code, not just future work. `sagas::delete_block::run`'s layout-delete step resolves a leaf via `getNodeByBlockId`, which matches ANY member of a multi-block `blockStack` (`frontend/layout/lib/layoutStack.ts`), and deletes the whole leaf regardless of how many blocks it holds. Closing one member of a stack orphans its siblings (still in `tab.blockids`, no layout node, invisible). Tracked separately: [agentmuxai/agentmux#3202](https://github.com/agentmuxai/agentmux/issues/3202). Narrow edge case (stacked panes aren't the common single-block-per-leaf case this tool is mainly for), but real — flagging here rather than letting it hide in the spec doc's own follow-up list.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv close_pane_tests` — 3/3 pass (own-pane close, cross-pane close with verified audit attribution, rejects an invalid signature)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv fleet` / `delete_block` — no regressions (21 + 11 pass)
- [x] `cargo test -p agentmux-mcp` — 26/26 pass (tool count assertion bumped 43→44)
- [x] `cargo check -p agentmux-common -p agentmux-srv -p agentmux-mcp` — clean
- [ ] Manual verification against a real running instance (closing another agent's stuck pane) — recommend before merge given the destructive nature of the cross-pane path

One unrelated pre-existing failure, confirmed via `git stash`: `backend::sysinfo::handle_anomaly_tests::refresh_processes_specifics_does_not_leak_a_handle_per_call` fails identically with these changes fully stashed out — an environment-sensitive OS-handle-count test on this shared machine, not caused by this change.

Reopened from #3193, which was accidentally opened under the shared `a5af` account instead of this agent's own `Agent5-asaf` identity — same branch/content, ReAgent already approved and Codex's finding already fixed.

**Depends on #3195 merging first** (cites the spec doc — CI's citation gate will fail until then).


a5af:
@codex review

PR #3197, opened 2026-09-11
fix(srv): correct three review findings in the catalog redirect merged in #3183

AgentA-asaf:
## Why this PR exists

PR #3183 (Phase 2's application-layer catalog read/write redirect) got 3 real P1/P2 findings from Codex during review. I pushed a fix commit, but a background merge-watcher raced ahead and merged #3183's PRE-fix commit before the fix registered as the new head — so `main` currently has the version of #3183 with these 3 bugs still present. Not shipped to end users (pre-release), but it's broken code on `main` and needs correcting properly through review rather than left as-is.

This PR is the same fix commit (verified, tested, already reviewed-in-spirit by Codex's own findings) cherry-picked onto current `main`.

## The three findings, all verified against the actual code before being fixed

- **P1:** `m0031`'s non-starter-global dedup fallback re-queried by `(name, skill_type)`, which the new name-only unique index (`m0033`, part of the same original PR) makes stale. Two channels carrying the same skill name under different `skill_type`s would both fail the re-query, fall through to the "unrelated collision" branch, and the fresh-id insert would ALSO silently no-op against the name constraint (`INSERT OR IGNORE`) — stranding a local ref pointing at an id the identity store never actually holds, which vanishes the moment reads redirect there. Fixed by re-querying by name alone (`skill_find_global_by_name`, mirroring the MCP-server equivalent, which never had this mismatch), plus a defensive re-check after the fresh-id insert for a residual TOCTOU race.

- **P1:** `managed_delete` purged ref rows BEFORE deleting the catalog row. If the catalog delete then failed (identity store locked, I/O error), the row's refs were already gone — for a private row, this made it permanently inaccessible (no ref, not global) and un-retryable (the delete RPC's own pre-check depends on the ref that was just destroyed). Reordered: catalog-delete first, so a failure there returns before any ref is touched and the row stays exactly as accessible as before the call.

- **P2:** `skill_rewrite_ref_id`/`mcp_server_rewrite_ref_id` used `UPDATE OR IGNORE`, which silently no-ops when the owner already holds a ref to both the loser and survivor ids (a real scenario once `m0033` can collapse two independently-bound skills into one). The untouched loser ref became a permanent dangling ref once `m0033` deletes the loser catalog row right after. Changed to insert-the-survivor-then-delete-the-loser, which resolves the ref correctly either way.

5 new regression tests, each exercising the exact failure shape described rather than a loosened version of it.

## Tests

`cargo test -p agentmux-srv --bins` — 3337 passed, 0 failed (excluding two known-flaky PTY-shell tests from #3177 that hang in this sandboxed environment regardless of these changes, and one `sysinfo` handle-count test confirmed flaky under heavy concurrent build load, not a real regression — both unrelated to this diff, verified by `git diff main --stat` showing neither touched file in this PR's changes).

### DIFF
--- a/.changesets/1789110842-fix-srv-correct-three-review-findings-in-the-catalog-redirec-oolw.md
+++ b/.changesets/1789110842-fix-srv-correct-three-review-findings-in-the-catalog-redirec-oolw.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(srv): correct three review findings in the catalog redirect (#3183)
--- a/agentmux-srv/src/backend/storage/managed.rs
+++ b/agentmux-srv/src/backend/storage/managed.rs
@@ -325,17 +325,39 @@ impl Store {
         }
     }
 
-    /// Purge `id`'s ref rows on `self`, then delete the catalog row on
-    /// `catalog`. True if a catalog row was deleted.
+    /// Delete the catalog row on `catalog` FIRST, then purge `id`'s ref rows
+    /// on `self`. True if a catalog row was deleted.
+    ///
+    /// Order matters (Codex P1, PR #3183): with the ref-purge running first,
+    /// a catalog-delete failure AFTER the ref purge already committed (disk
+    /// full, the identity store locked, any I/O error) stranded the row —
+    /// still present in `catalog`, but no ref points at it and it is not
+    /// necessarily global, so no access-check path can find it again, and
+    /// the ordinary delete RPC's own pre-check (`skill_is_bound_to`) now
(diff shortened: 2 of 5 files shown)


a5af:
@codex review

AgentA-asaf:
Verified against the code — you're right, this was a real asymmetry, not a stylistic difference. Fixed in f60d5fbd7: mirrored the same defensive re-check (re-query by name if the fresh-id insert also silently no-ops via INSERT OR IGNORE, error out only if that finds nothing either) into \`carry_mcp_servers\`'s equivalent branch. Full suite: 3337 passed, 0 failed.

PR #3198, opened 2026-09-11
Korp@claudius: fix(ci): add concurrency groups to ci-pr.yml and nightly workflows

GenericAgentX-asaf:
## Summary

Investigated a 3+ hour hang on PR #3189's `windows-latest` CI leg (flagged directly by the repo owner). Root cause is **not** a code-level test hang — sampled job-level timing across recent `ci-pr.yml` runs and found two jobs that started **1.5 hours apart** both died at almost the exact same wall-clock moment (`04:08 UTC`), while a third job that started in between them succeeded normally in 15 minutes. That's the signature of GitHub-hosted-runner-side flakiness (a bad runner instance, eventually reaped), not a deterministic deadlock in this repo's own test suite — and it's not something a workflow YAML change can fix outright.

**What is fixable:** neither `ci-pr.yml` nor the nightly workflows had a `concurrency:` group at all. Every push to a PR spawned a full new run without cancelling the now-superseded previous one. Observed directly: several PRs had 3-4 near-duplicate runs in flight within minutes of each other from quick successive pushes. 88 `ci-pr.yml` runs fired in a sampled 21-hour window alone, across many concurrent agents — real, avoidable pressure on the same shared `windows-latest`/`ubuntu-latest` pool that makes exposure to the flaky-runner failure mode worse than it needs to be.

- **`ci-pr.yml`**: concurrency group keyed on workflow + branch/PR ref, `cancel-in-progress: true` — a new push cancels the stale run for the same PR instead of letting it run to completion pointlessly.
- **`ci-nightly-build.yml` / `ci-nightly-artifacts.yml`**: each gets its own group, `cancel-in-progress: false` (queue rather than cancel a real nightly run — unlike a superseded PR run, an in-progress nightly is still a signal worth finishing).

**Honest limitation, stated in the comments:** this does **not** give nightly separate runner capacity from PR CI. GitHub-hosted runners are one shared pool per repo/org regardless of concurrency group — true isolation would need self-hosted runners or GitHub Enterprise runner groups, neither of which this repo has. What this achieves is narrower but real: nightly stops competing with itself on re-trigger/schedule overlap, and PR CI stops piling up wasted, superseded runs — one less source of load on the shared pool, not a guarantee against GitHub's own runner flakiness.

## Test plan

- [x] YAML syntax validated (`python3 -c "import yaml; yaml.safe_load(...)"`) on all three files
- [ ] CI green on this PR
- [ ] Manual: push a second commit to this PR and confirm the first run gets cancelled

### DIFF
--- a/.changesets/1789111084-fix-ci-add-concurrency-groups-to-ci-pr-yml-and-nightly-workf-ct4w.md
+++ b/.changesets/1789111084-fix-ci-add-concurrency-groups-to-ci-pr-yml-and-nightly-workf-ct4w.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(ci): add concurrency groups to ci-pr.yml and nightly workflows — stop stale runs piling up on the shared runner pool
--- a/.github/workflows/ci-nightly-artifacts.yml
+++ b/.github/workflows/ci-nightly-artifacts.yml
@@ -26,6 +26,14 @@ on:
     - cron: '0 6 * * *' # 06:00 UTC daily (ahead of Phase A at 07:00)
   workflow_dispatch: {}
 
+# Own group, distinct from both ci-pr.yml's and ci-nightly-build.yml's — see
+# the matching comment in ci-nightly-build.yml for the full reasoning
+# (queue rather than cancel a real nightly run; this does not give artifact
+# builds separate runner capacity from PR CI, only prevents self-collision).
+concurrency:
+  group: ci-nightly-artifacts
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

GenericAgentX-asaf:
Fixed — good catch, that was a real cross-PR hazard. Keyed on github.event.pull_request.number instead, which is unique per PR regardless of branch naming, rather than github.head_ref (just the source branch name, no fork/owner qualifier).

PR #2774, opened 2026-08-23
Korp@claudius: fix(editor): use a real native scrollbar for markdown preview, matching source exactly

GenericAgentX-asaf:
## Summary

Follow-up to #2769 (already merged), which repositioned the markdown
preview's scrollbar to sit flush at the pane edge. That fixed *position*
but not *appearance*: Preview still used OverlayScrollbars (a JS-rendered
overlay track with `autoHide: "leave"` fade behavior), which looks visually
thinner/different than Source mode's plain native CodeMirror scrollbar even
at the same declared 7px width. Requested follow-up: make Preview use the
exact same scrollbar as Source, not just a same-sized one in the same spot.

## Changes

- `frontend/app/element/markdown.tsx` — new `nativeScrollbar` prop on
  `<Markdown>`. When set, `onMount` skips calling `OverlayScrollbars(...)`
  entirely, so `.content`'s plain CSS `overflow` renders a real native
  scrollbar styled by the same universal `*::-webkit-scrollbar` rule
  CodeMirror's `.cm-scroller` already uses. The heading-anchor-scroll effect
  (`focusedHeading`) now falls back to `contentsEl` directly as the scroll
  viewport when `contentsOs` is `null`, so in-page anchor links
  (`[text](#heading)`) still work in native mode.
- `frontend/app/view/editor/editor-view.tsx` — passes `nativeScrollbar` to
  `<Markdown>` at the editor preview's one call site.
- `frontend/app/view/editor/editor-view.scss` —
  `.content.editor-preview-markdown-content` switches `overflow-y` from the
  base `.content { overflow: scroll }` to `auto`, matching CodeMirror's
  `.cm-scroller` default (no reserved scrollbar gutter when content already
  fits).
- Spec updated: `docs/specs/SPEC_EDITOR_MARKDOWN_PREVIEW_SCROLLBAR_ALIGNMENT_2026_08_22.md` §6.

No other `<Markdown>` consumer is affected — `nativeScrollbar` defaults to
`false`, so every other call site (agent chat, tool output, etc.) keeps its
existing OverlayScrollbars behavior unchanged.

## Verification

- `npx tsc --noEmit` — clean.
- `npx stylelint` on both touched SCSS files — no new findings (13
  pre-existing issues elsewhere in `editor-view.scss`, none on touched
  lines).
- **Verified live via `task dev`**: opened a markdown file, compared Source
  and Preview scrollbars side by side — now visually identical (same width,
  same native rendering, no auto-hide fade difference).

PR #3199, opened 2026-09-11
fix(armory): render Global Memory's provider-config and combined preview as markdown, make both resizable

AgentA-asaf:
Part of the current push to get agent memory actually workable — this is the first, concrete UI fix flagged for the Armory → Memory → Global page.

## What changed

Two of the three content surfaces in `GlobalBundleManager` (Armory → Memory → Global) were plain `<pre>` dumps at a fixed `max-height: 240px`, no markdown rendering, no resize:

- The **"Claude Code provider config"** reference block (the one literally labeled "provider config" in the UI)
- The **"Combined preview"** toggle panel

The third surface (section content editors) was already a resizable `<textarea>` — untouched.

## How

Reuses two already-proven patterns already in this codebase rather than building anything new:
- `Markdown` (`frontend/app/element/markdown.tsx`) for rendering — the same component used in `AgentSkillsModal` and elsewhere.
- CSS `resize: vertical` for the drag handle, wrapped in a sized `<div>` rather than applied directly to `<Markdown>`'s own root — its root sets `height: 100%; overflow: hidden`, which would fight a resize/height override on the same element. `Markdown` fills 100% of whatever height the wrapper resizes to and handles its own scrolling internally (`nativeScrollbar={true}` — plain CSS scrollbar, appropriate for a read-only preview panel).

## Verification

- `npx tsc --noEmit` — clean, no type errors.
- **Automated test suite is broken repo-wide in this environment right now** (91 of 267 frontend test files fail identically with a `@solid-refresh`/vitest tooling error — confirmed on files I never touched, e.g. `global-bundle-model.test.ts`, so this is pre-existing and unrelated to this change, not something I introduced).
- Built and booted a full `task dev` instance with this change — compiles clean, app launches with no runtime errors.
- **Could not get a full click-through screenshot of the specific Armory → Memory → Global page**: the available UI-automation tools (`UIClick`/`UIQuery`) are scoped to my own pane and can't drive a separately-spawned dev instance's window, and I don't want to overstate what I verified. Saying so explicitly rather than claiming a visual check I didn't actually get. Worth a manual look before merge.

### DIFF
--- a/.changesets/1789113727-fix-armory-render-global-memory-s-provider-config-and-combin-15x6.md
+++ b/.changesets/1789113727-fix-armory-render-global-memory-s-provider-config-and-combin-15x6.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+fix(armory): render Global Memory's provider-config and combined preview as markdown, make both resizable
--- a/frontend/app/view/global-bundle/global-bundle-manager.test.tsx
+++ b/frontend/app/view/global-bundle/global-bundle-manager.test.tsx
@@ -54,7 +54,13 @@ describe("GlobalBundleManager shared-provider-config block", () => {
 
         expect(await screen.findByText("Claude Code — shared provider config")).toBeInTheDocument();
         expect(screen.getByText("/home/user/.agentmux/shared/providers/claude/CLAUDE.md")).toBeInTheDocument();
-        expect(screen.getByText("# Global rules")).toBeInTheDocument();
+        // Rendered as markdown now, not a raw <pre> dump (this test's own
+        // history predates that change) — "# Global rules" becomes an <h1>
+        // reading "Global rules", the "#" consumed by markdown parsing
+        // rather than appearing as literal text. Asserting on the rendered
+        // heading proves markdown rendering actually happened, not just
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentA-asaf:
Both verified against the code — thanks.

**Raw-text test assertion:** already fixed in 1d1f9b73f (pushed right as this comment landed) — updated to assert the rendered `<h1>` ("Global rules") instead of the raw markdown source, which also proves rendering actually happened rather than just that the string landed somewhere in the DOM. CI's own vitest run caught this independently and it's green now.

**overflow: scroll vs auto:** confirmed — `nativeScrollbar` only skips OverlayScrollbars, it doesn't touch `.content`'s own CSS `overflow: scroll` in markdown.scss (both axes). Fixed in 11dbd4918 with the exact same `contentClass` compound-selector override an existing call site already uses for this identical problem (`.content.editor-preview-markdown-content`, your own P2 on PR #2774) — `overflow: auto` so a scrollbar only shows when content genuinely overflows.

Full suite green in CI (this branch's earlier CI run already confirmed 3923/3927 tests passing before these two fixes, with only the one test needing an update).

@codex review

PR #3200, opened 2026-09-11
docs(incident): CI-PR windows-runner hang backlog, 2026-09-10

AgentY-asaf:
## Summary
Repo owner asked when the `ci-pr.yml` backlog started and for a written RCA, and asked me to check on Korp's in-progress fix (#3198, still open).

This independently re-derives Korp's root-cause finding — GitHub-hosted-runner-pool flakiness, not a code-level test deadlock — straight from the Actions API run/job history, rather than just citing it:

- Seven independently-started `windows-latest` jobs, with start times spread across more than 2.5 hours, **all died within the same 90-second wall-clock window** (04:08:05–04:09:35 UTC on 2026-09-11) — a signature inconsistent with a deterministic per-run timeout (which would expire at a fixed *offset* from each run's own start, not converge on one shared *absolute* moment).
- Pins the actual blockage start to **2026-09-10T23:01:40Z** (the first *required*-leg hang) — an earlier 16:37 UTC episode that day only ever affected the non-blocking `ubuntu-latest` leg and had no merge impact.
- Identifies a **second, still-unaddressed structural gap** alongside the missing concurrency groups Korp's PR fixes: no `timeout-minutes` anywhere in `ci-pr.yml`, so a hang occupies the required-check queue for GitHub's default 360 minutes instead of being killed early. Two runs in this incident hit that exact ceiling.
- Separately notes (own section, explicitly out of scope for this incident) that the repo owner's standing ask — get the required leg under 10 minutes — needs its own test-timing profiling to decide what moves to nightly; this doc doesn't attempt that.

## Test plan
- [x] Cross-checked every claimed timestamp against `repos/agentmuxai/agentmux/actions/workflows/ci-pr.yml/runs` and `.../runs/{id}/jobs`
- [x] Cross-checked the "not a specific commit" claim against `git log origin/main` for the incident window
- N/A — docs only, no code path affected

### DIFF
--- a/docs/incident/INCIDENT_2026_09_10_CI_PR_WINDOWS_RUNNER_HANG_BACKLOG.md
+++ b/docs/incident/INCIDENT_2026_09_10_CI_PR_WINDOWS_RUNNER_HANG_BACKLOG.md
@@ -1,0 +1,175 @@
+# INCIDENT 2026-09-10 — `ci-pr.yml`'s required Windows leg hangs for hours, backing up every open PR
+
+**Status:** partially addressed — Korp's PR #3198 (concurrency groups) is open, not yet
+merged. The two structural gaps that let a single hang become an hours-long backlog
+(no `timeout-minutes`, no concurrency groups) are analyzed below; only one is fixed so far.
+
+**Severity:** High — every open PR against `agentmuxai/agentmux` is gated on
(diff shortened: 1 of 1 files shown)


a5af:
@codex review

PR #3206, opened 2026-09-11
perf(srv): coalesce PTY output before broadcasting — cuts renderer/GPU CPU during active agent output

AgentO-asaf:
RCA + fix. Full investigation: `docs/reports/REPORT_RENDERER_CPU_UNBATCHED_PTY_OUTPUT_2026_09_11.md`.

Repo owner, live: *"figure out why the agentmux helper takes up so much CPU. we can feel the slowdown."*

## What I found (measured, not guessed)

- **Real, sustained CPU** confirmed via live `top -pid` reads: `AgentMux Helper (Renderer)` and `(GPU)` sustained 16–28% CPU in roughly-once-per-second bursts on a long-running production window.
- **Profiled** the renderer with macOS `sample` (15s @ 1ms resolution). The CPU time was overwhelmingly `mach_msg2_trap`/`kevent64` — 176,807 + 50,192 of ~253k samples — versus single/double-digit counts for every actual "doing work" frame. That's the signature of **many small, frequent wakeups**, not one hot function.
- **`muxspect`** showed two agent panes `turn_active: true` on that exact window at the time — one of them the investigating session itself.

## Root cause

`agentmux-srv/src/backend/blockcontroller/shell/lifecycle.rs`'s PTY output read loop reads in 4KB chunks with **zero batching**. Every single `read()` return fired its own file write, WebSocket broadcast, OSC extraction, and (for agent panes) JSON-stream translation pass. A fast producer — `cargo build`, `git log`, a verbose test run, all things this multi-agent session ran constantly — delivers output across many separate `read()` returns in quick succession, so what could have been one or two renderer updates became dozens of small ones, each its own IPC round-trip. Matches the profile exactly.

## Fix

The blocking `read()` loop's semantics are **completely untouched** — this file supports both Unix and Windows PTYs, and that boundary isn't something to risk getting subtly platform-wrong. Instead, reads now go through an `mpsc::unbounded_channel` to a new standalone async fn, `run_pty_output_flusher`, which:

- waits up to `PTY_COALESCE_WINDOW` (20ms) after the first chunk of a batch for more to arrive,
- or stops early at `PTY_COALESCE_MAX_BYTES` (256KiB) regardless of the window,
- does the file write + broadcast + OSC publish + translation **once per batch** instead of once per read,
- flushes any final partial batch when the channel closes (PTY EOF/error) rather than dropping it.

Content is byte-identical either way — same bytes, same order — only the call *frequency* changes.

## Verification

- `cargo test -p agentmux-srv`: **3280 passed, 0 failed** — no regression in existing PTY/lifecycle/scrollback/OSC/translation coverage.
- **5 new tests**, against a real in-memory `FileStore` + a real `wps::Broker` with a recording `WpsClient`, so broadcast **count** (not just final content) is directly observable — that's the property that actually matters here:
  - `rapid_chunks_coalesce_into_far_fewer_broadcasts_with_no_data_loss` — 50 back-to-back chunks collapse into **exactly 1 broadcast** (deterministic: every chunk is queued before the flusher is ever polled; verified stable across 8 repeated local runs before asserting the exact count rather than a loose bound), with content losslessly reconstructed both from the final file and from concatenating every broadcast payload in delivery order.
  - `byte_cap_forces_a_flush_before_the_channel_closes` — 80×4KB (over the cap) forces ≥2 broadcasts rather than growing one batch unboundedly.
  - Plus the single-batch flush primitive and the no-broker no-op path in isolation.

**Deliberately not done:** a live before/after CPU comparison against the actual production instance this investigation was run from. That would mean restarting the shared AgentMux window mid-investigation — tearing down every open pane, including ones that were actively `turn_active: true` at measurement time. That's a real, visible, disruptive action on shared state, not something to do unasked. Left for a deliberate follow-up (a `task package:macos` build + manual swap, or the next normal release) — see the report's §7 for exactly what would close this out.

### DIFF
--- a/.changesets/1789141357-perf-srv-coalesce-pty-output-before-broadcasting-cuts-render-oiuk.md
+++ b/.changesets/1789141357-perf-srv-coalesce-pty-output-before-broadcasting-cuts-render-oiuk.md
@@ -1,0 +1,5 @@
+---
+type: patch
+---
+
+perf(srv): coalesce PTY output before broadcasting — cuts renderer/GPU CPU during active agent output
--- a/agentmux-srv/src/backend/blockcontroller/shell/lifecycle.rs
+++ b/agentmux-srv/src/backend/blockcontroller/shell/lifecycle.rs
@@ -26,7 +26,8 @@ use super::controller::KILL_GRACE_SECS;
 use super::controller::{ShellController, SHELL_INPUT_CH_SIZE};
 use super::file_ops::handle_append_block_file;
 use super::pty::{
-    detect_local_shell_path_windows, PTY_COALESCE_MAX_BYTES, PTY_COALESCE_WINDOW, PTY_READ_BUF_SIZE,
+    detect_local_shell_path_windows, PTY_CHANNEL_CAPACITY, PTY_COALESCE_MAX_BYTES,
+    PTY_COALESCE_WINDOW, PTY_READ_BUF_SIZE,
 };
 use super::translation::accumulate_and_translate;
 use crate::backend::obj::{self, MetaMapType};
@@ -94,7 +95,7 @@ fn flush_pty_batch(
 /// it can be driven directly in a test with a synthetic channel and no real
 /// PTY — see `pty_output_flusher_tests` below.
 async fn run_pty_output_flusher(
-    mut rx: mpsc::UnboundedReceiver<PtyReadChunk>,
+    mut rx: mpsc::Receiver<PtyReadChunk>,
     broker: Option<Arc<wps::Broker>>,
     block_id: String,
     filestore: Option<Arc<crate::backend::storage::filestore::FileStore>>,
@@ -154,15 +155,33 @@ async fn run_pty_output_flusher(
             }
         };
 
-        flush_pty_batch(
-            &broker,
-            &block_id,
-            &batch,
-            &batch_osc,
-            filestore.as_ref(),
-            &mut line_buf,
-            translator.as_mut(),
-        );
+        // block_in_place, not a bare call (reagentx P1 on PR #3206):
+        // flush_pty_batch does blocking SQLite I/O (FileStore::stat /
+        // append_data_at) and takes std::sync::Mutex locks. Calling it
+        // directly here would run that on a Tokio async worker thread —
+        // this closure is inside a plain `tokio::spawn`, not
+        // `spawn_blocking` — and with several panes streaming output at
+        // once (exactly the scenario this whole fix targets), slow disk or
+        // lock contention could occupy every worker and delay WebSocket,
+        // input, and RPC tasks that have nothing to do with this pane.
+        // `block_in_place` runs the closure on this thread while telling
+        // the runtime to spin up a replacement worker if needed, so other
+        // tasks aren't starved — and unlike `spawn_blocking`, it needs no
+        // 'static/Send ownership transfer, so `&mut line_buf` and
+        // `translator.as_mut()` can stay borrowed exactly as they already
+        // are. Requires a multi-thread runtime; `agentmux-srv`'s
+        // `#[tokio::main]` (main.rs) is one by default.
+        tokio::task::block_in_place(|| {
+            flush_pty_batch(
+                &broker,
+                &block_id,
+                &batch,
+                &batch_osc,
+                filestore.as_ref(),
+                &mut line_buf,
+                translator.as_mut(),
+            );
+        });
 
         if closed_mid_batch {
             break;
@@ -798,7 +817,11 @@ impl Controller for ShellController {
         let filestore_flush = self.filestore.clone();
         let is_agent_read = is_agent;
         let is_agent_flush = is_agent;
-        let (pty_tx, mut pty_rx) = mpsc::unbounded_channel::<PtyReadChunk>();
+        // Bounded (reagentx P1 on PR #3206) — see PTY_CHANNEL_CAPACITY's own
+        // doc comment for why: an unbounded channel here would let a
+        // sustained fast producer grow queued memory without limit if the
+        // flusher ever falls behind the read rate.
+        let (pty_tx, pty_rx) = mpsc::channel::<PtyReadChunk>(PTY_CHANNEL_CAPACITY);
 
         tokio::task::spawn_blocking(move || {
             let mut reader = reader;
@@ -845,12 +868,20 @@ impl Controller for ShellController {
                                 raw
                             };
 
-                            // Best-effort: if the flusher task is gone
-                            // (block torn down mid-read), there is nothing
-                            // useful to do with this chunk — the PTY read
-                            // loop itself must keep draining so the child
-                            // process's stdout never backs up.
-                            let _ = pty_tx.send(PtyReadChunk {
+                            // `blocking_send`, not `send` — this closure runs
+                            // on a `spawn_blocking` thread, not async, and a
+                            // bounded channel's plain `send` is itself async.
+                            // Blocks THIS (already-dedicated-to-blocking)
+                            // thread once the channel is full, which is
+                            // exactly the backpressure PTY_CHANNEL_CAPACITY's
+                            // doc comment describes — never the async runtime.
+                            //
+                            // Best-effort beyond that: if the flusher task is
+                            // gone (block torn down mid-read), there is
+                            // nothing useful to do with this chunk — the PTY
+                            // read loop itself must keep draining so the
+                            // child process's stdout never backs up.
+                            let _ = pty_tx.blocking_send(PtyReadChunk {
                                 data: chunk.to_vec(),
                                 osc_events,
                             });
@@ -1414,7 +1445,7 @@ mod controller_agent_id_tests {
 
 #[cfg(test)]
 mod pty_output_flusher_tests {
-    use super::{flush_pty_batch, run_pty_output_flusher, PtyReadChunk};
+    use super::{flush_pty_batch, run_pty_output_flusher, PtyReadChunk, PTY_CHANNEL_CAPACITY};
     use crate::backend::storage::filestore::FileStore;
     use crate::backend::wps;
     use std::sync::{Arc, Mutex};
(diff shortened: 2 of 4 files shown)


a5af:
@codex review

AgentO-asaf:
**Both P1s confirmed and fixed in `f42b92c` — the review was exactly right, and both would have undermined the fix under the exact load scenario it targets.**

## Bound the PTY handoff queue

Correct. `PTY_COALESCE_MAX_BYTES` only bounded the batch currently being assembled — nothing bounded chunks still sitting in the channel if the flusher fell behind the read rate. A sustained fast producer could have grown queued memory without limit, which is the exact failure mode this whole fix exists to close, reintroduced one layer up.

Fixed with a **bounded** `mpsc::channel(PTY_CHANNEL_CAPACITY = 128)`. The read loop's send moved from `send` to `blocking_send` — required, since that closure runs on a `spawn_blocking` thread and a bounded channel's plain `send` is itself async. `blocking_send` blocks the calling thread once the channel is full, giving real backpressure all the way back to the PTY read rate, and from there (via the kernel's own PTY buffer) to the child process itself — the same mechanism a slow terminal naturally applies to a fast writer.

New test (`blocking_send_backpressures_once_the_channel_is_full`) proves this directly: a capacity-1 channel, a second `blocking_send` issued from a real background OS thread (mirroring the production call shape, not a simulation), confirmed still blocked after 50ms, then confirmed to complete once the receiver drains one message. Stable across 6 repeated local runs before trusting the timing-based negative assertion.

## Run synchronous flushes outside Tokio worker threads

Also correct. `flush_pty_batch` does blocking SQLite I/O and takes `std::sync::Mutex` locks — previously safe inside `spawn_blocking`, but my refactor moved the call into a plain `tokio::spawn` task, running it directly on an async worker. With several streaming panes at once (exactly this fix's target scenario), slow disk or lock contention could occupy every worker and delay unrelated WebSocket/input/RPC tasks.

Fixed with `tokio::task::block_in_place` around the flush call, not `spawn_blocking` — `spawn_blocking` needs `'static + Send` ownership of everything the closure touches, and `flush_pty_batch` borrows `&mut line_buf` / `translator.as_mut()` from the flusher's own stack. `block_in_place` runs inline on the current thread (no ownership transfer needed) while telling the runtime to spin up a replacement worker so other tasks aren't starved.

**Caught its own follow-on bug immediately**, which is worth surfacing: the two existing end-to-end coalescing tests started panicking — `block_in_place` requires a multi-thread runtime, and plain `#[tokio::test]` defaults to current-thread. This only affects the test harness (the real app's `#[tokio::main]` is already multi-thread by default); fixed by adding `#[tokio::test(flavor = "multi_thread")]` to both. Flagging it here mainly because it's a good example of the test suite doing exactly its job — I'd have shipped a runtime panic in a real multi-pane session otherwise.

`cargo test -p agentmux-srv -- --test-threads=1`: 3332 passed, 0 failed. Report updated with a new §8 covering this round.

AgentO-asaf:
@codex — these two are the same findings reagentx flagged in its review pass; both fixed in \`f42b92c\`, explained in my comment above (bounded \`mpsc::channel\` + \`blocking_send\` for the queue, \`block_in_place\` for the flush). Not ignoring these, just avoiding a duplicate reply — see the earlier comment for the detail and verification on both.

AgentO-asaf:
**Fixed in \`932e2bc\` — and worth explaining exactly why, since this touched tokio internals.**

I checked the claim against the actual mechanism rather than trusting either side's characterization. Read the vendored \`tokio\` 1.52.3 source directly (\`runtime/scheduler/multi_thread/worker.rs\`): \`block_in_place\`'s own replacement-worker path is \`runtime::spawn_blocking(move || run(worker))\` — it draws on the **same shared \`spawn_blocking\` pool** internally, so "unbounded growth" distinct from that pool doesn't hold literally.

What *does* hold, and is the real problem: using \`block_in_place\` here drew on that shared pool **twice per flush** — once for the replacement worker, once implicitly for the original thread continuing to run — for work \`spawn_blocking\` alone covers with a single call. Tokio's own docs name \`block_in_place\`'s one genuine advantage over \`spawn_blocking\` as protecting *other concurrent work in the same task* (e.g. across a \`join!\`). This flusher task does nothing else concurrently — so that advantage never applied here, and the extra indirection was pure overhead under exactly the long-running, many-panes, high-throughput scenario this whole fix targets. Good catch.

Switched to \`spawn_blocking\`, solving the original ownership problem properly instead of routing around it: \`batch\`/\`batch_osc\` are freshly built every iteration (cheap to move), and \`line_buf\`/\`translator\` — the only state the flush actually mutates — are moved into the closure and handed back out via the \`JoinHandle\`'s return value, rather than kept borrowed across the call.

One side effect: the two coalescing tests no longer need \`#[tokio::test(flavor = "multi_thread")]\` — that requirement came specifically from \`block_in_place\` (panics on a current-thread runtime); \`spawn_blocking\` has none. Reverted to plain \`#[tokio::test]\`.

\`cargo test -p agentmux-srv -- --test-threads=1\`: 3332 passed, 0 failed. Report's §9 has the full writeup.

a5af:
@codex review

AgentO-asaf:
Addressed reagentx's round-3 P1 (`lifecycle.rs:1016` — STATUS_DONE published with no join on the coalescing flusher) in a40cf5e5a.

**Fix:** captured the flusher's `JoinHandle` (`flusher_handle`) at its `tokio::spawn(run_pty_output_flusher(...))` call site (was dropped before). The wait task changed from a bare `tokio::task::spawn_blocking(move || { ... })` into an outer `tokio::spawn` async task that does `flusher_handle.await` first — logging and proceeding with teardown anyway on panic/cancellation rather than hanging the pane — then runs the entire original wait-task body (`child.wait()` through `run_lock.store(...)`), unchanged, inside its own inner `spawn_blocking`. That inner wrap isn't incidental: `child.wait()` and the cleanup that follows it (`registry::remove`, etc.) both do blocking/synchronous work, so none of it belongs on the outer async task either — a first draft of this fix that only wrapped `child.wait()` and left the rest inline would have reintroduced the exact bug class from an earlier round (flush running on an async worker) one call site over. Caught that myself on a second read before ever running it.

**New test:** `awaiting_the_flusher_join_handle_observes_every_broadcast_first` — spawns `run_pty_output_flusher` via `tokio::spawn` (same call shape as production), sends a chunk from a second task after a real delay so the flusher genuinely waits on the channel (not pre-queued like the existing coalescing tests), closes the channel, then asserts the trailing batch is already in FileStore and already broadcast by the time `flusher_handle.await` returns — no extra yield needed. Pins the exact ordering the wait task now depends on. The wait task itself spawns a real PTY + child process and isn't practical to drive end-to-end from a unit test, so this is the closest practical proxy.

**Verification:**
- `cargo check -p agentmux-srv --tests`: clean.
- `cargo test -p agentmux-srv pty_output_flusher_tests`: 7 passed.
- `cargo test -p agentmux-srv backend::blockcontroller::shell:: -- --test-threads=1`: 62 passed (includes `test_shell_controller_start_stop`/`test_shell_controller_force_start`, which exercise the real `start()` path this round touched).
- `cargo test -p agentmux-srv -- --test-threads=1`: 3333 passed, 0 failed, 7 ignored — full suite, no regressions.

Report updated: `docs/reports/REPORT_RENDERER_CPU_UNBATCHED_PTY_OUTPUT_2026_09_11.md` §10.

AgentO-asaf:
Addressed codex's P1 (\`lifecycle.rs:999\` — reap the child before awaiting PTY EOF) in 3228cbec2.

Real gap in the round-3 fix itself: gating child reaping on \`flusher_handle.await\` meant a background descendant that inherits the PTY slave fd and keeps it open (ignores SIGHUP) could prevent the read loop from ever observing EOF — \`pty_tx\` never drops, the flusher's channel never closes, its \`JoinHandle\` never resolves, and reaping/`STATUS_DONE`/deregistration/`run_lock` release would all hang for that descendant's entire remaining lifetime.

**Fix:** mirrors \`persistent.rs\`'s existing handling of the identical descendant-held-descriptor case for its stdout/stderr readers (thanks for the pointer). \`child.wait()\` now runs first, unconditionally, in its own \`spawn_blocking\` — not gated on the flusher at all. Only after it resolves does the wait task \`tokio::time::timeout(FLUSHER_DRAIN_TIMEOUT, flusher_handle)\`, and on expiry it \`abort()\`s the flusher (not just stops waiting on it) before proceeding with cleanup. \`FLUSHER_DRAIN_TIMEOUT = 10s\` (\`pty.rs\`), matching \`persistent.rs\`'s own stdout-reader bound for the same scenario.

No new dedicated test for the abort-on-timeout branch itself — same reasoning as \`persistent.rs\`'s own identical pattern (no test there either): reproducing it needs a real PTY and a real lingering descendant process, which a synthetic channel can't stand in for.

**Verification:**
- \`cargo check -p agentmux-srv --tests\`: clean (one compile error caught and fixed locally first — \`block_id_wait\` moved into the new reap closure, then reused after; fixed by cloning it in).
- \`cargo test -p agentmux-srv pty_output_flusher_tests\`: 7 passed (unchanged from the previous round — this fix doesn't touch \`run_pty_output_flusher\` itself).
- \`cargo test -p agentmux-srv backend::blockcontroller:: -- --test-threads=1\`: 261 passed, including \`persistent.rs\`'s own test modules and all of \`blockcontroller::shell::tests\`.
- \`cargo test -p agentmux-srv -- --test-threads=1\`: 3333 passed, 0 failed, 7 ignored — full suite, no regressions.

Report updated: \`docs/reports/REPORT_RENDERER_CPU_UNBATCHED_PTY_OUTPUT_2026_09_11.md\` §11.

AgentO-asaf:
Addressed reagentx's P1 (\`lifecycle.rs:1061\` — \`abort()\` only cancels the consumer, never the producer) in 0bf5c09af. You're right that the mirroring of \`persistent.rs\`'s pattern was only superficial — there, one combined async task both reads and processes, so aborting it genuinely stops the read. Here the read loop is a separate \`spawn_blocking\` doing a raw, blocking OS \`read()\` with no \`.await\` in it — a running \`spawn_blocking\` task can't be interrupted by \`abort()\` at all, so aborting the flusher reclaimed nothing, while making every later \`blocking_send\` find the receiver gone and silently discard the data, permanently losing any further descendant output with no log marking the loss. Net effect of the abort version: same leak, plus a new silent-data-loss regression.

**Fix:** removed the \`abort()\` call. On timeout, \`flusher_handle\` (owned by the \`tokio::time::timeout\` future) simply drops without being aborted — a bare drop detaches the task rather than cancelling it, so the flusher and its read loop keep running independently in the background, still persisting/broadcasting whatever the descendant produces, until the PTY genuinely reaches EOF on its own. This doesn't reclaim the OS thread either (nothing can, short of closing the underlying fd, which isn't safe to do from outside given it's a raw platform read) — but it stops pretending to, and in exchange nothing produced after the timeout is silently lost.

**Verification:**
- \`cargo check -p agentmux-srv --tests\`: clean.
- No new test — same reasoning as the prior round's note: this needs a real PTY and a real lingering descendant to exercise end-to-end.
- \`cargo test -p agentmux-srv backend::blockcontroller:: -- --test-threads=1\`: 261 passed.
- \`cargo test -p agentmux-srv -- --test-threads=1\`: 3333 passed, 0 failed, 7 ignored — full suite, no regressions.

Report updated: \`docs/reports/REPORT_RENDERER_CPU_UNBATCHED_PTY_OUTPUT_2026_09_11.md\` §12.

AgentO-asaf:
Fixed the stale comment (P2, `lifecycle.rs:1007`) in 939815e8e — it described the abort()-on-expiry design from the previous commit, which the following commit had already removed. Pointed it at the timeout's own (correct) comment instead of duplicating and contradicting it.

PR #3209, opened 2026-09-12
feat(codex): add disabled App Server stdio controller foundation

Agent5-asaf:
## Summary
- add provider-neutral ControllerType::AppServer and typed Codex App Server capability metadata without changing the active Codex controller
- add bounded JSONL framing, JSON-RPC request correlation, cancellation-safe pending slots, server-request/notification classification, and fail-closed protocol handling
- add exact initialize/initialized handshake metadata, bounded stderr tail, graceful close/kill timeout, and typed crash classification
- test a deterministic standalone fake App Server process plus fragmented/coalesced frames, CRLF, out-of-order responses, opaque IDs, malformed/oversized frames, remote errors, cancellation, stderr bounds, and lifecycle failures

## Verification
- cargo test -p agentmux-srv backend::blockcontroller::app_server::tests --no-fail-fast (9 passed)
- cargo test -p agentmux-srv backend::providers::tests::codex_advertises_app_server_without_enabling_it --no-fail-fast
- rustfmt/check and git diff --check
- manually verified the exact no-jsonrpc handshake against @openai/codex@0.154.0

Feature remains disabled by default: Codex controller_type is still Subprocess.


a5af:
@codex review

a5af:
@codex review

Agent5-asaf:
Superseded by #3228 (the squashed history of #3212–#3215), which carries its own independently-evolved and further-fixed copy of this same App Server foundation work — including concurrency fixes (deadlock in `AppServerProcess::shutdown`, three races in the turn/thread state machine) that ReAgent had already found on this PR's own branch but which were never actually applied here, since #3212's branch didn't build on top of it. Closing without merging to avoid duplicating/conflicting with the now-landed work.

PR #3211, opened 2026-09-12
feat(codex): add guarded App Server turn controls

Agent5-asaf:
## Summary
- add schema-shaped turn/steer and turn/interrupt request builders
- enforce active thread/turn identity and exact expectedTurnId preconditions
- keep local turn state active until authoritative turn/completed arrives
- add an explicit transport response primitive for server-initiated requests; it never auto-approves and is covered by a wire test

## Verification
- cargo test -p agentmux-srv backend::blockcontroller::app_server --no-fail-fast (15 lifecycle/protocol tests pass)
- rustfmt/check and git diff --check

The provider remains disabled by default and no approval policy is wired yet.


Agent5-asaf:
@codex review

a5af:
@codex review

Agent5-asaf:
Superseded by #3228 (the squashed history of #3212–#3215), which carries its own independently-evolved and further-fixed copy of this same App Server foundation work — including concurrency fixes (deadlock in `AppServerProcess::shutdown`, three races in the turn/thread state machine) that ReAgent had already found on this PR's own branch but which were never actually applied here, since #3212's branch didn't build on top of it. Closing without merging to avoid duplicating/conflicting with the now-landed work.

PR #3217, opened 2026-09-15
Camper@narko: feat(jekt): 24h TTL / lazy rotation for the host-tier signing key

GenericAgentX-asaf:
## Summary
- `db_agent_jekt_keys` (host-tier HMAC signing key) was minted once at spawn and never rotated — a documented, known gap in `SPEC_JEKT_TRUST_LAYER_COMPLETION_2026_08_13.md` and `SPEC_JEKT_LAN_TIER_SIGNING_2026_08_15.md`.
- `agent_jekt_key_ensure` now rotates a key past 24h old (`JEKT_KEY_TTL_SECS`) the next time it's called — i.e. at that agent's next spawn — instead of returning it forever. This applies for free to every already-existing agent (their `created_at` is already well past 24h, so their very next spawn rotates it) as well as every newly-spawned agent going forward.
- Verification (`agent_jekt_key_load`, read by `reactive.rs` sender verification and `ui_handlers.rs` UI-automation auth) is byte-for-byte unchanged — a live agent process is never forced into `TRUST=unverified` mid-session by its own key aging out from under it.
- No schema migration: TTL is computed from the existing `created_at` column at read time.

## Deliberately out of scope
LAN-tier keys (`db_agent_lan_keys`) do **not** get the same TTL in this PR. `db_lan_peer_pubkey_pins` is a trust-on-first-use pin of a peer's LAN public key with no expiry and no re-pin path — rotating the LAN key would make every peer that already pinned the old key see the new one as an active forgery (`lan_verified = Some(false)`), which CLAUDE.md's jekt rules treat as unconditionally forced `TIER=sensitive`/`ESCALATE=required`, with no way for that peer to recover short of a re-pin mechanism that doesn't exist yet. Full rationale in `docs/specs/SPEC_JEKT_HOST_KEY_TTL_ROTATION_2026_09_14.md` §4; closing that half is tracked as a follow-up.

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::storage::agent_jekt_keys` — 9/9 pass, including 5 new rotation tests (no-rotation-before-TTL, rotation-after-TTL, load-reflects-rotation, cross-agent isolation).
- [x] `cargo build -p agentmux-srv` — clean build, no new warnings.


a5af:
@codex review

PR #3218, opened 2026-09-15
feat(armory): show current file content in Personal Memory history panel

AgentA-asaf:
## Summary
- Armory -> Memory -> Personal never rendered a selected file's actual content — only version metadata and a two-version diff. This adds an always-visible, read-only "Current content" section to the shared `NativeMemoryHistoryPanel` (used by both Armory's Personal Memory tab and the Stash "Memory" History toggle).
- `NativeMemoryHistoryModel` gains `contentAtom`/`contentErrorAtom` + `loadContent()`, mirroring `AgentNativeMemoryModel.selectFile()`'s existing `agent:memory:read_file` call. Fetched on mount alongside `loadHistory()`. `null` = loading, `""` = genuinely empty file — same convention as the reference model, so loading/empty render differently.
- `revertTo()` now always refreshes `contentAtom` after a revert (previously only fetched content conditionally, only for the Stash caller's `onReverted` callback).
- Rendering reuses the resizable-panel + compound-selector CSS pattern from Global Memory's PR #3199 fix (`global-bundle.scss`).
- Editing stays exactly where it already lives (Stash's `AgentNativeMemoryModal` textarea) — this is a read-only preview only, per spec's non-goals.

Follows `docs/specs/SPEC_ARMORY_PERSONAL_MEMORY_CONTENT_VIEW_2026_09_11.md` (included in this PR), which documents in detail why this was a missing surface, not a bug, and what was reused vs. new.

## Test plan
- [x] New unit tests in `native-memory-history-model.test.ts` (written first, confirmed red, then green): initial content load, loading-vs-empty-string distinction, content-fetch error isolated from history error, content refresh on revert.
- [x] `npx vitest run frontend/app/view/agent/native-memory-history-model.test.ts` — 10/10 passing.
- [ ] Manual verification in `task dev` (Armory -> Memory -> Personal -> select a file).


a5af:
@codex review

PR #3219, opened 2026-09-15
Korp@narko: fix(swarm): couple tracked-blocks refresh to the registry it actually reads

GenericAgentX-asaf:
## Summary

`agent.tracked-blocks` (Swarm's only data source) reads `CONTROLLER_REGISTRY`
directly, but the frontend's `loadTrackedBlocks()` was only re-triggered by
four WPS events sourced from two *different*, independently-timed mechanisms
(`process_tracker`'s 2s poll-and-diff loop, and `reactive`'s own registration
map) — neither coupled to `register_controller`/`delete_controller`, the
functions that actually determine the answer.

**Live-reproduced, not theoretical:** a running instance's Swarm pane showed
1 of 5 currently-active, responding agent panes. The `agentmux-srv` log
proved all five had registered a controller correctly (`resync_controller`
firing with the right `controller_type` at boot and again on every message
send) — so the bug wasn't registration, it was refresh going stale for any
block whose proxy signal didn't happen to fire.

## Changes

- `broker/process.rs`: new `EVENT_TRACKED_BLOCKS_CHANGED` +
  `ProcessBroker::emit_tracked_blocks_changed()`, unscoped (list-membership
  signal, not a per-block status update).
- `blockcontroller/mod.rs`: `register_controller`/`delete_controller` now
  call it directly — the refresh signal is now sourced from the same code
  path that actually changes what the list contains.
- `swarm-model.ts`: subscribes to the new event (in addition to the
  existing four — they still carry other useful signal) and adds a bounded
  12s safety-net poll so the list can't go stale indefinitely from any
  *future* gap in event wiring either.

Full investigation, live evidence (log traces, a screenshot of the actual
reproduction, and what was ruled out along the way):
`docs/reports/REPORT_SWARM_MOUNT_DEPENDENT_TRACKING_GAP_2026_09_15.md`

## Verification

- `cargo test -p agentmux-srv --bin agentmux-srv blockcontroller::` — 253 passed
- `cargo test -p agentmux-srv --bin agentmux-srv broker::process::` — 15 passed
- `npx tsc --noEmit -p .` — clean (pre-existing unrelated `.bench.ts` errors only)
- `npx vitest run frontend/app/view/swarm/` — 134 passed


a5af:
@codex review

PR #3222, opened 2026-09-15
fix(launcher): unify splash-screen timeline logic across platforms

AgentY-asaf:
## Summary

The splash screen "looks stuck" in places because Windows (`splash.rs`), macOS (`splash_mac.rs`), and Linux (`splash_linux/`) each reimplemented the same row-timeline logic independently — so every fix that's landed so far only ever fixed one platform:

- The same-tick count-up deferral fix (a fast step's row snapping straight to its final value instead of visibly counting up) existed only on macOS.
- The frozen-row defect (a row whose `End` never arrives stays visually "running" forever once the render loop stops redrawing on a timer) was **never fixed on any platform, including macOS** — it's a separate defect from the one above.
- Total/other reconciliation ("does the tally add up") existed only on macOS, as a post-hoc bucket; Windows showed a bare unreconciled total, Linux showed no total at all.

Full diagnosis: `docs/reports/REPORT_SPLASH_SCREEN_ARCHITECTURE_RETHINK_2026_09_14.md` (written in a prior investigation this PR implements).

## What changed

- New `agentmux-launcher/src/splash_core.rs`: shared row model (`StageEntry`/`SubEntry`), event application with same-tick deferral (`StageTimeline::apply_tick`), a frozen-row finalizer (`finalize_running`), duration formatting, and total/other reconciliation (`reconcile`) — pure logic, no FFI, fully unit-tested (9 new tests).
- **Windows** (`splash.rs`): now uses `splash_core` instead of its own copy of the row/apply logic (fixes the count-up bug there for the first time); gains an `finalize_running` call at dismiss (fixes the frozen-row bug, including an already-shipping concrete instance — the `first-run-wait` sub-row, which has no `SubEnd` call anywhere in the codebase); gains a reconciled `other:` row alongside `total:`.
- **macOS** (`splash_mac.rs`): now delegates to `splash_core` instead of its own (already count-up-fixed) copy; gains the `finalize_running` call it was missing (the frozen-row bug was live here too, despite already having the count-up fix — the two are independent defects).
- **Linux** (`splash_linux/`): `StageList` now wraps `splash_core::StageTimeline` (fixes count-up + frozen-row there too). Deliberately does **not** get the total/other row or the freeze-then-hold-then-fade timing model Windows/macOS use — its 4-line budget (`STAGE_MAX_LINES`) makes an other/total pair a poor fit, and unifying the dismiss-timing model is flagged in the report as needing a product decision, not an engineering one. Left as an explicit follow-up.

## Verification

- `cargo check -p agentmux-launcher` / `cargo test -p agentmux-launcher` (244 tests) / `cargo clippy -p agentmux-launcher --no-deps` all clean on Windows, the buildable target on this machine.
- macOS and Linux platform files only compile under their own `cfg`, so I could not build or run either locally — same constraint the original macOS stage-panel PR documented. Both were read-verified line-by-line against the pre-refactor logic they were lifted from (Linux additionally gets a real compile check from CI's `ubuntu-latest` leg, which macOS's PR-time CI does not cover — see `build-macos.yml`, release-only).

## Test plan

- [ ] CI green on `windows-latest` (required) and `ubuntu-latest` (non-blocking but should still be checked for real Linux compile errors before merge)
- [ ] `task package` a Windows portable and eyeball the splash: fast steps should now show at least one "running" frame before completing, and a row that never gets a definite end (hard to trigger normally) should freeze amber/warn-colored rather than staying blue forever


a5af:
@codex review

PR #3223, opened 2026-09-15
feat(srv,launcher): show real per-migration progress on the splash instead of a silent clock

AgentY-asaf:
## Summary

The report's largest single "looks stuck" finding (`docs/reports/REPORT_SPLASH_SCREEN_ARCHITECTURE_RETHINK_2026_09_14.md` §2.1): real DB migrations run entirely inside srv's already-open `backend` splash stage, with the only cross-process signal being `AGENTMUXSRV-MIGRATING migrations:<n>` — a count emitted before the batch starts, used only to extend the launcher's ESTART timeout to 30 minutes. No per-migration progress ever crosses the process boundary, so a genuinely slow migration (the code's own timeout allows up to 30 minutes) shows only a bare `Backend startup ▶ N.Ns...` clock with zero indication anything unusual is happening.

The one place that *would* show real sub-rows — `run_migrate`, a subprocess path with its own stdout JSON progress protocol — is dead code, never called in real startup (its own doc comment says so; confirmed by a zero-hit grep for callers).

## What changed

- `agentmux-common/src/srv_stderr.rs`: two new stderr line formats/parsers, `AGENTMUXSRV-MIGRATION-BEGIN` / `-END`, alongside the existing `MIGRATING`/`-FAILED` lines in the same shared protocol module both real supervisors (the launcher, and the CEF host sidecar in `task dev:standalone`) already parse the rest of this protocol from.
- `agentmux-srv/src/migrations/runner.rs`: `run_pending_migrations` (the real, in-process boot path) now passes a real `ApplyProgress` — the hook already existed (built for the CLI/dead-subprocess path) but was previously passed `None` here.
- `agentmux-launcher/src/srv_spawner.rs`: parses the two new lines and turns them into `sub_begin`/`sub_end` calls against the already-open `"backend"` stage — a migration isn't a separate top-level stage, it happens entirely inside that one.

Net effect: a slow migration now shows real, individually-timed sub-rows on the splash instead of a silent clock.

Independent of #3222 (the splash_core DRY refactor from the same report) — this PR touches none of the same files and can land either order.

## Verification

- `cargo check`/`test` clean across all three touched crates (`agentmux-common`, `agentmux-srv`, `agentmux-launcher`) on Windows.
- New unit tests for the line format/parse round-trip, malformed-input safety, and prefix-collision safety against the other three protocol lines (`agentmux-common/src/srv_stderr.rs`).
- Not live-verified against a real slow migration (none are currently pending on this dev machine) — the wiring itself is exercised by the new unit tests plus the existing migration-runner test suite (157 tests, unaffected).

## Test plan

- [ ] CI green on `windows-latest` (required) and `ubuntu-latest`
- [ ] If a slow migration is ever pending during a packaged-build smoke test, confirm its sub-row appears and updates live


a5af:
@codex review

PR #3225, opened 2026-09-15
Loap #2@narko: fix(ptyshell): don't respawn a pane's shell that already exited

GenericAgentX-asaf:
## Summary

Fixes both halves of the original bug report: typing `exit` in a terminal pane used to (1) loop forever instead of closing, and (2) even after the loop fix, the pane just sat inert with no way to tell the shell had exited.

**Part 1 — the respawn loop (§4b/§7-9 in the spec):** `PtyShellCreate`'s "attach to this pane's existing shell" path called `resync_controller` unconditionally, which respawns any exited shell in place — correct for crash/backend-restart recovery, wrong for an intentional `exit`. Fixed so an exited shell is treated like a stale pointer (falls through to creating a fresh shell) instead of being silently resurrected. Two rounds of automated review (ReAgent, Codex) caught and fixed a stale comment plus two real race conditions in that fix (a TOCTOU between the status check and resync, and a concurrent-double-create race) — see spec §8-9.

**Part 2 — close-on-exit (§4a/§10):** the meta-driven close-on-exit knobs (`cmd:closeonexit` etc.) were fully implemented but never wired to anything. Now they are: `ShellController`'s wait/cleanup task triggers an actual pane close (delete the block, prune it from the tab's layout, notify the frontend) via a new process-wide callback slot (mirrors `crate::broker::process::set_global`'s shape) installed once `AppState` exists at boot. Defaults to `true` for interactive shell panes (Terminal widget, agent composer-drawer shells) and `false` for one-shot `cmd` panes (which need to keep showing their output). Live-testing on Windows during this PR also surfaced and fixed a real bug: gating the close on the PTY flusher having drained made it almost never fire, because `FLUSHER_DRAIN_TIMEOUT` (10s) gets hit on essentially every ordinary `cmd.exe` exit (ConPTY's console-host process routinely outlives the direct child). Reverted to closing unconditionally.

Full investigation + design writeup: `docs/specs/SPEC_TERM_EXIT_RESPAWN_LOOP_2026_09_15.md`.

## Test plan

- [x] `server::tests::ptyshell_create_does_not_respawn_a_shell_that_already_exited` — real-PTY regression test for the loop fix.
- [x] `server::tests::ptyshell_shell_pane_closes_itself_after_exit` — real-PTY regression test for close-on-exit (installs a local test hook, bypassing the saga's `AppState` requirement).
- [x] `shell::tests::test_effective_close_on_exit_*` — unit tests for the per-controller-type default.
- [x] Full `ptyshell_*` real-PTY suite (6 tests) passes together in one process.
- [x] Full non-ignored `agentmux-srv` suite: 3352 passed, 0 failed.
- [x] `cargo check -p agentmux-srv` clean.
- [x] Live-tested in a `task dev` build on Windows — confirmed the loop is gone in srv logs (one spawn/exit per `exit`, no runaway respawn), and iterated on close-on-exit against the real live instance.


a5af:
@codex review

PR #3226, opened 2026-09-15
fix(pane): restore full-perimeter selection ring on hoisted agent/terminal pane chrome

AgentA-asaf:
## Summary

- The focused-pane accent ring stopped wrapping an Agent or Terminal pane's header (and, for agents, the progress bar) once `AgentPaneChrome`/`TermPaneChrome` hoisted them outside the switch-scoped `<Block>` that `.block-mask` rings (PR #3136/#3157, `SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md`) — the ring visibly stopped partway down the pane instead of wrapping the whole thing, unlike every other pane type.
- Adds an equivalent ring directly on `.agent-pane-stack`/`.term-pane-stack`, mirroring `.block-mask`'s own three states (dim default / accent focused / transparent when the pane is alone in its tab).
- Suppresses the now-redundant inner ring so there's no double-line seam where the header used to end.
- Insets the agent pane's progress-bar overlay by the new ring's border width so it nests inside the ring instead of painting over it.

Full root-cause writeup + before/after ASCII diagrams: `docs/retro/RETRO_AGENT_PANE_SELECTED_BORDER_MISSES_HOISTED_HEADER_2026_09_15.md`.

## Test plan

- [x] Verified live in a `task dev` build: user confirmed the Agent pane fix visually, then reported the Terminal pane had the identical regression — fixed here using the same pattern.
- [ ] Would help to have a second visual pass on the Terminal pane specifically (fixed after the initial live check), across both a focused and an unfocused/single-tab pane.
- No automated test coverage added — this is a pure CSS/classList fix with no behavior change to verify beyond visual appearance; existing agent/term pane tests are unaffected (no assertions on `.block-mask`/ring classes existed for these components).


a5af:
@codex review

AgentA-asaf:
Addressed all four review findings in 508dac708:

- **Codex P2** (agent-view.scss:108) + the equivalent term.scss finding: a pane's configured border color (`frame:activebordercolor`/`frame:hue`/`bg:bordercolor`/`bg:activebordercolor`) now propagates to the outer ring instead of being hardcoded to `--accent-color`. Extracted `BlockMask`'s existing color resolution into an exported `computeFocusRingBorderColor` (blockframe.tsx) and reused it in both `AgentPaneChrome` and `TermPaneChrome`, wired through a `--pane-ring-color` CSS custom property.
- **ReAgent P1** (term.scss:303, z-index) + **P1** (term.scss:296, will-change): the new ring now matches `.block-mask`'s own treatment exactly — `will-change: transform` unconditionally, `z-index: 50` once focused — so it composites above `.term-stickers` and xterm's GPU-promoted scroll layer the same way `.block-mask` always did.
- **ReAgent P2** (agent-view.tsx:811): fixed the alone-pane suppression to require BOTH focused and alone (matching `.pane-alone .block-mask`'s own nesting under `.block-focused`) — an alone-but-unfocused pane now keeps its ordinary dim border instead of showing no border at all.

@codex review

PR #3227, opened 2026-09-15
refactor(armory): declutter Global Memory pane into a unified file list

AgentA-asaf:
## Summary

- The Global Memory pane (Armory → Memory → Global) stacked nine visually distinct blocks for what's fundamentally one thing: a list of entries composed into an agent's startup instructions file. Content was also hidden behind "Edit" for every section — you couldn't scan what Global Memory currently says without opening every row.
- Collapses this into: **one sentence** at the top, **one consistently-shaped list** — every entry (the read-only Claude Code reference file, system-tier entries, ordinary sections) now renders as a path/label row + an always-visible markdown preview (reusing the resizable `Markdown` pattern from PR #3199/#3218 a third time), a shrunk "+" add affordance, and the **Combined preview** kept at the very end.
- The "applies to" filename chips are removed entirely, per direct request.
- Inline add/edit/remove/reorder is unchanged in capability — confirmed with the user before implementing that this stays a decluttered inline editor, not a switch to a read-only/click-through redesign.

Full design + before/after ASCII diagrams: `docs/specs/SPEC_ARMORY_GLOBAL_MEMORY_DECLUTTER_2026_09_15.md`.

`filenameGroupsAtom`/`noFileProvidersAtom` (the provider→filename mapping that powered the removed chips) are kept in the model, just unused by this view now — they're real, tested domain logic that may be surfaced elsewhere later, not dead code from this change.

## Test plan

- [x] Type-checked (`tsc --noEmit`) — clean on all touched files.
- [x] Updated `global-bundle-manager.test.tsx` for the new DOM shape (badge text moved to a `title` attribute, so tests now scope off the visible path text instead); ran locally — hit the same pre-existing, environment-specific vitest failure (`file:///@solid-refresh` `ERR_INVALID_ARG_VALUE`) reproduced on an untouched, unrelated test file too, so not a regression from this change. CI runs the full suite fine.
- [x] Visually verified live in a running `task dev` instance — screenshot confirms the one-sentence intro, the reference file's path + resizable markdown preview rendering correctly, the shrunk add affordances, no "applies to" chips, and Combined preview at the bottom.
- [ ] Not independently re-verified: the per-section preview row (system/ordinary entries) — the running dev instance had zero sections at capture time, so only the reference-file row (structurally identical) was visually confirmed. Worth a quick look once a section exists.


a5af:
@codex review

PR #3229, opened 2026-09-15
Opaz@charlie: fix(cef-build): use_static_angle=false for Linux CEF 152 — silent GPU regression

GenericAgentX-asaf:
## What

`scripts/cef-build/args.gn` (Linux) never set `use_static_angle`. At CEF 152 the declared default (`ui/gl/features.gni: use_static_angle=true`) combines with `ui/gl/BUILD.gn`'s `use_angle_stubs = use_static_angle && (is_linux || is_win || is_mac)` to ship dummy placeholder `libEGL.so`/`libGLESv2.so` instead of a real ANGLE implementation — zero build errors, zero warnings, but GPU is fully disabled at runtime (`eglGetProcAddress` not found at init, no fallback to hardware or SwiftShader).

Same bug class as #3172 (Windows) — that fix didn't cover Linux (each platform's `args.gn`/`args-windows.gn`/`args-darwin.gn` needs its own line under the `is_linux || is_win || is_mac` gate), and nobody had checked Linux specifically until now (the spec explicitly flagged this as unchecked).

## Verification

- **Before fix:** `verify-angle-libs.sh` — both `libEGL.so` (963,712 B) and `libGLESv2.so` (963,912 B) exported neither `eglGetProcAddress` nor `glGetString`.
- Checked the macOS-style alternative too (real ANGLE statically linked into `libcef.so` itself, which is why macOS's `args-darwin.gn` correctly leaves this unset) — full `nm` symbol-table search of the unstripped `libcef.so` found **zero** `angle::`/`rx::` symbols. Not that case either — genuinely broken, not a false negative.
- **After fix:** `libGLESv2.so` grows to 27,086,504 bytes with a real GL export table; `verify-angle-libs.sh` passes clean; `verify-cef-patch.sh` (BeginWindowDrag) still passes; `cefsimple` boots clean.
- Rebuilt and republished [`cef-linux-x86_64-152.0.7977.83-codecs`](https://github.com/agentmuxai/cef/releases/tag/cef-linux-x86_64-152.0.7977.83-codecs) on `agentmuxai/cef` with the fix — this closes Linux's half of tracking issue #3108 (all three platforms now have a published 152 release).


a5af:
@codex review

GenericAgentX-asaf:
Fixed — good catch, the wording was genuinely contradictory (\`c3f1ad9\`). Reworded to make clear the gate condition means each platform must be *checked* independently, not that each needs an identical override — Linux/Windows need \`use_static_angle=false\`, macOS correctly leaves it unset.

PR #3230, opened 2026-09-15
Loap #2@narko: feat(term): close a shell pane when its process exits

GenericAgentX-asaf:
## Summary

Second half of the `exit` bug. The respawn-loop half shipped in #3225; this is close-on-exit, rebased onto main after that merge (fresh branch — #3225's branch is dead).

`should_close_on_exit` / `close_on_exit_delay_ms` were fully implemented in `shell/controller.rs` but had **no caller**. Now wired up: the wait/cleanup task triggers a real pane close (`sagas::delete_block::run`) via a process-wide callback slot installed at boot — mirroring `crate::broker::process::set_global`'s shape, because the saga needs a full `AppState` that `ShellController` has no handle to.

Defaults on for an interactive shell pane; off for a one-shot `cmd` pane (it exists to show its output), and off for any **sub-block** (`parentoref` starting with `block:` — an agent composer-drawer shell or a `PtyShellCreate` headless shell). Closing one of those leaves the parent's `term:shellsubblockid` dangling and the drawer silently recreates it, reproducing the respawn loop by another route.

Two further bugs surfaced only by live retesting, both fixed here:

- **`FLUSHER_DRAIN_TIMEOUT` 10s → 1s, and skipped entirely for a closing pane.** `STATUS_DONE` is published only after that wait, and on Windows it was hit on *every* ordinary exit (ConPTY's console host outlives the child, so the read loop never sees EOF) — leaving the pane unable to react for ten seconds. Pre-existing, and it affected every `STATUS_DONE`-driven affordance, not just this feature.
- **`wave_obj_bridge` had no arm for `LayoutBackendActionsQueued`**, so the saga's queued layout-delete never reached an already-open window. Block deleted, backend tree pruned — but the frontend kept its leaf, rendering a pane whose block resolved to `null`: a permanent loading spinner instead of a close. Invisible until now because the queue's only previous producer was tear-off, which creates a *new* window whose fresh `LayoutModel` reads the queue at construction.

`close_on_exit_delay_ms` default 2000 → 0 — the knob had no caller before this, and the pause reads as lag once it actually fires.

Measured end-to-end, same test and machine, across the three rounds: **13.09s → 4.06s → 1.56s** (and ~1s of that last figure is the test's own setup sleeps plus 500ms poll granularity).

Full design writeup and the diagnosis trail, including two tests that gave false confidence and why: `docs/specs/SPEC_TERM_EXIT_RESPAWN_LOOP_2026_09_15.md` §10-14.

## Test plan

- [x] `wave_obj_bridge::tests::test_backend_actions_queued_broadcasts_layout_update` — **verified to genuinely catch the bug**: with the new arm disabled it fails on the timeout, not vacuously. Checked deliberately, because two earlier tests in this effort passed under both broken and fixed logic.
- [x] `shell_pane_with_no_parent_closes_itself_after_exit` — real-PTY, now using a realistic `tab:<id>` parent (an earlier version used `parentoref: ""`, a value that never occurs in production, and so passed against broken logic).
- [x] `ptyshell_create_does_not_close_the_parented_shell_pane_after_exit` — the sub-block exclusion.
- [x] `shell::tests::test_effective_close_on_exit_*` — per-controller-type defaults.
- [x] Full real-PTY `ptyshell_*` suite (6) passes together in one process.
- [x] Full non-ignored suite: 3413 passed. One unrelated failure, `fs_watch::pool::tests::sweep_does_not_resurrect_a_target_unsubscribed_after_being_snapshotted`, reproduces identically on clean `main` — it counts process-global OS handles and is sensitive to parallel test scheduling.
- [x] Live-verified in a `task dev` build on Windows across several rounds.


a5af:
@codex review

PR #3232, opened 2026-09-15
refactor(armory): unify Global Memory into one list, drop Section/System Entry terminology

AgentA-asaf:
## Summary

Follow-up to #3227. Replaces the three visible "AgentMux" strings in `GlobalBundleManager` with "Global Memory" — the already-established name for this feature (`SPEC_ARMORY_MEMORY_GLOBAL_PERSONAL_RENAME_2026_08_22.md`), per direct request after reviewing the shipped #3227 UI:

- The system-tier inline tag (was `AgentMux`, now `Global Memory`)
- The "+ Add AgentMux system entry" button → "+ Add Global Memory system entry"
- The system-name input's example placeholder (`e.g. AgentMux Policy` → `e.g. Global Memory Policy`)

**Scoped to UI copy only.** `formatGlobalBundleBlock`'s composed-file heading (`# [AgentMux System] ...`, tested in `global-bundle-model.test.ts`) is left unchanged — that's content the agent itself reads at launch, a different and more consequential thing than UI chrome. Flagged to the user; happy to follow up separately if that should change too.

## Test plan

- [x] Type-checked (`tsc --noEmit`) — clean.
- [x] Confirmed via HMR log that the dev build reloaded these files with no compile errors.
- [ ] Not independently re-screenshotted — the dev window I'd been using went idle/unrendered (long gap since last capture) and I can't drive a foreign window to restore it. The change is a 3-string text swap in already-visually-verified JSX from #3227, so risk is low, but worth a quick look.


a5af:
@codex review

AgentA-asaf:
Follow-up commit (c00644790) per further review:

- Simplified action wording: "+ Add Global Memory system entry" → "+ Add Memory", "Add system entry" → "Add Memory", "Delete this system entry" → "Delete this Memory" — the "Global"/"system entry" phrasing was redundant once the action already reads as adding/deleting memory inside a pane already titled Global Memory.
- Removed the "Promote existing bundle" dropdown entirely (not just decluttered) — it was the only UI path to flip an existing bundle's `is_global` flag, but was asked to be removed outright. Deleted `GlobalBundleViewModel.promote()`/`candidatesAtom` alongside it (untested, no other caller, genuinely dead once the UI trigger is gone).

The system-tier inline tag still reads "Global Memory" (not just "Memory") — flagged as an open question, not yet changed, pending confirmation.

AgentA-asaf:
Second follow-up commit (301f02621 + 88361c251), a bigger one — per further discussion, the "system" vs "ordinary" split itself (not just its wording) is gone from this pane:

- One merged list (system rows still sort first, then ordinary — built explicitly from the two split atoms since `sectionsAtom`'s own order isn't documented as system-first).
- `SectionEditor`/`SystemSectionEditor` collapsed into one `MemoryEditor`.
- One "+ Add Memory" button, one "No memories yet." empty state. Creating a **new** system-tier entry has no UI trigger anymore — existing ones stay fully editable/removable via their own still-isolated RPCs (`upsertsystemmemory`/`deletesystemmemory`, unchanged, never wired to MCP tools).
- No backend/schema change — this is presentation-only. The backend write-path isolation from `SPEC_GLOBAL_MEMORY_SYSTEM_TIER_2026_08_24.md` is untouched and still real.
- Also fixed two remaining "Section" error strings the earlier wording passes missed.

Full postmortem (why the original frontend mirroring of the backend split was the actual mistake) + refactor plan: `docs/specs/SPEC_GLOBAL_MEMORY_UNIFY_SYSTEM_AND_ORDINARY_2026_09_15.md`.

**One resolved-by-judgment-call, not explicit confirmation:** pinned/override-priority entries now render identically to ordinary ones — no accent border or tag distinguishing them anymore. Flagged in the spec itself in case that's not what was wanted; happy to add back a minimal visual signal if so.

Visually verified live in the running dev instance (empty-state screenshot matches this design) — real per-item merged-row rendering with actual system+ordinary data mixed together hasn't been visually confirmed yet since the workspace currently has zero entries of either kind.

Retitled the PR to reflect the actual final scope.

PR #3233, opened 2026-09-15
Korp@narko: chore: release v0.56.0

GenericAgentX-asaf:
## Summary

Minor version bump, `0.55.43` → `0.56.0`, consuming all 12 pending
changesets — the CEF 152 milestone (target-gating the mechanism, the
Linux ANGLE fix, and today's full collapse to CEF 152 across all three
platforms) is the headline, riding alongside a dozen other queued
features/fixes.

Purpose: get a real, tagged build out so CEF 152 can actually be tested
on all three platforms, not just verified in CI/by the platform agents.

## Changes

- `VERSION_HISTORY.md`, `package.json`, `Cargo.toml`, `Cargo.lock`,
  `package-lock.json` — all five version locations bumped to `0.56.0`
  (the release-consistency invariant `reagent` checks).
- All 12 pending `.changesets/*.md` consumed and deleted.

## Verification

- `task release -- --as minor` — script-verified all 5 version locations
  agree at `0.56.0` before this diff was staged.
- No `major`-type changeset was pending, so `minor` is the release
  script's own naturally-computed type, not a forced override.


a5af:
@codex review

GenericAgentX-asaf:
Good catch — confirmed against both files directly:
\`nightly-release.yml\`'s subject filter and \`guard-stable-build.sh\`'s
case match both require the commit subject to literally start with
\`chore: release v\`, with nothing before it. A default squash-merge here
would have used this PR's title (\`Korp@narko: chore: release v0.56.0\`,
per this repo's shared-identity PR-naming convention) as the commit
subject, silently breaking both — v0.56.0 would never get tagged/
published, and stable packaging would refuse to build from it.

Will merge with an explicit subject override (\`chore: release v0.56.0\`,
no prefix) so the actual commit matches what the automation expects. The
PR title itself stays prefixed for the human-readable PR-list reason that
convention exists for — this is scoped to the commit subject specifically.

PR #3235, opened 2026-09-15
Loap #2@narko: fix(test): assert the fs_watch stale-rearm guard directly, not via handle counts

GenericAgentX-asaf:
## Summary

`sweep_does_not_resurrect_a_target_unsubscribed_after_being_snapshotted` fails intermittently on `main` — it passes in isolation but fails in a full `cargo test` run.

**Why:** it measured `process_handle_count()` (process-**global**) before/after 200 subscribe→unsubscribe→stale-rearm rounds and asserted growth under 100. In a full run the ~3400 other tests execute in parallel and contribute their own file/socket/thread churn to that same window. Observed failure: `grew_by=103` against a threshold of 100 — it was measuring the test suite, not the code under test.

It was also necessarily `#[cfg(target_os = "windows")]` (`process_handle_count` is Windows-only), so Linux and macOS got no protection from it at all.

**Fix:** assert the guard's own observable signature instead. `rearm_if_still_subscribed()` only touches degraded state on a path that actually reaches `watch()` — `clear_degraded` on success, `mark_degraded` on failure. Returning early, as it must for an unsubscribed target, leaves the flag untouched. So marking the target degraded *after* the unsubscribe and finding it still marked afterward proves `watch()` was never called.

The test's intent, setup and the invariant it protects are unchanged — only the instrument. It also now covers all three platforms, and runs in 0.00s instead of sampling handle counts across 200 rounds.

## Test plan

- [x] **Verified in both directions** — passes as written, and fails with the intended message when the re-check in `rearm_if_still_subscribed()` is removed to simulate the original bug. (Not vacuous: a test that passes under both broken and fixed logic is exactly what this PR is replacing.)
- [x] Added a precondition assert that the target really is absent from `targets`, so the test can't silently stop exercising the stale-rearm path.
- [x] Full non-ignored suite green across **three consecutive runs** (3418 passed, 0 failed) — previously failed intermittently on this test.


a5af:
@codex review

PR #3234, opened 2026-09-15
docs(specs): widening Linux package targets + a dedicated downloads page

Agent3-asaf:
## Summary

A fresh spec — no prior spec, issue, or discussion covering this was found anywhere (searched `docs/specs/`/`specs/` in `agentmux`, `agentmux-cloud`, `agentmux-landing`, `shared-infrastructure`, full git history/branches, and GitHub issues/discussions across the `agentmuxai` org). Written from a direct read of the current build/release pipeline and landing-page code, not a guess.

Covers two related asks:

1. **Widening Linux release/nightly package targets beyond AppImage-only.** Today Linux ships exactly one format (AppImage, x86_64) vs. 3 formats for Windows and 2 archs for macOS. Found that `.deb` support is already half-built — `agentmux-landing`'s `fetch-release.mjs`, `Download.tsx`'s types, and `docs/linux.md` all already expect a `.deb` asset, but nothing has ever produced one (`docs/linux.md` points at a nonexistent `agentmuxai/agentmux-builder` repo). Also found `ci-nightly-artifacts.yml` already implements "Phase B" of the nightly-builds spec (that spec's own doc is stale on this point) and shares `build-linux.yml` with real releases, so widening formats there widens both nightly and releases for free. Recommends repackaging the already-staged build tree with `fpm` into `.deb`/`.rpm`/`.pacman`/`.tar.gz` (no rebuild needed), phased by risk, with Snap/Flatpak/Linux-arm64 flagged as separate larger follow-ups.

2. **A dedicated `/download` page in `agentmux-landing`**, modeled on waveterm.dev/download (confirmed via direct fetch — three OS sections, every format per OS). Today's `#download` is a homepage section only, not a routable page. Spec keeps the homepage quick-CTA as-is and adds the new page as the full comparison grid, also surfacing SHA-256 checksums that are already computed/hosted (`fetch-release.mjs`) but never shown in the UI.

Includes phased rollout, per-repo files-to-touch tables, and 5 open questions (AUR vs. self-hosted pacman package, `fpm` as a new CI dependency, `.deb` dependency declarations, nav-link placement, and Linux-arm64 CEF-runtime access) that need a decision before implementation starts.

Also regenerates `docs/specs/INDEX.md` via `scripts/gen-docs-index.sh` to include the new spec.

## Test plan
- [x] `scripts/gen-docs-index.sh` run and its completeness assertion passed (871 specs accounted for)
[...]

PR #3236, opened 2026-09-15
feat(linux): widen release/nightly packaging to AppImage + .deb + .tar.gz

Agent3-asaf:
## Summary

Phase 1 of `docs/specs/SPEC_LINUX_DISTRO_TARGETS_AND_DOWNLOADS_PAGE_2026_09_15.md` (PR #3234). Linux release/nightly builds produced only an AppImage — Windows already ships 3 formats; this closes most of that gap.

- **`scripts/stage-linux-runtime.sh` (new)** — the shared "given `dist/`+`target/`, produce a runnable `usr/`-rooted tree" step, extracted out of `build-appimage-linux.sh`'s original steps 1-6b (binaries, CEF runtime, frontend, schema, VERSION marker, the BeginWindowDrag release gate) so every format ships the identical runtime instead of a hand-copied subset. This directly addresses Codex's P2 finding on PR #3234 — an earlier draft of the spec assumed raw `dist/` itself was a valid packaging input; it isn't, `build/AgentMux.AppDir` (now generalized) is.
- **`scripts/build-deb-linux.sh` (new)** — packages that staged tree into a `.deb` via `fpm`, installed to `/opt/agentmux` with a `/usr/bin/agentmux` wrapper and real desktop-integration files (distinct from the AppImage's self-registering `install-linux-desktop.sh`, which exists only because an AppImage has no privileged install step). Bundles everything statically, same as the AppImage — no `Depends:` declared (spec OQ4, still open).
- **`scripts/build-tarball-linux.sh` (new)** — the same staged tree as a portable `.tar.gz`, no packaging tool needed — Linux counterpart to Windows' portable ZIP.
- **`scripts/build-appimage-linux.sh`** — now calls `stage-linux-runtime.sh` instead of duplicating steps 1-6b; its own AppImage-specific steps (AppRun, `.desktop`, icons, `appimagetool` invocation) are unchanged.
- **`scripts/package-linux.sh`** — `--format=` now accepts a comma-separated list (e.g. `appimage,deb,tarball`) and runs the compile exactly once regardless of how many formats are requested — CI needs this to avoid tripling build time producing three formats from one commit.
- **`.github/workflows/build-linux.yml`** — installs `fpm`, builds all three formats from one compile, uploads each as both a workflow artifact and a release asset.
- **`.github/workflows/release.yml`** — extends the publish job's asset-flattening allowlist to include `*.deb`/`*.tar.gz`. **Without this the new formats would reach nightly artifacts but never the actual GitHub Release or the landing-page deploy that reads it** — this was Codex's P1 finding on PR #3234, caught before any code existed to be wrong about.
- **`Taskfile.yml`** — `task package:linux:deb` / `:tarball` (and `:release:` variants) for local builds, mirroring `package:linux`'s existing conventions.
- **`docs/linux.md`** — corrected the stale "Produced by CI builder (`agentmuxai/agentmux-builder`)" `.deb` claim (that repo doesn't exist in the org) and documented all three formats.

`ci-nightly-artifacts.yml` needs no change — it already calls `build-linux.yml` as a reusable workflow, so it picks up all three formats automatically.

Companion PR: agentmuxai/agentmux-landing#150 wires `fetch-release.mjs`/`Download.tsx` to recognize the new `.tar.gz` asset (`.deb` was already scaffolded there, unused until now).

## Test plan
- [x] `bash -n` on every changed/new shell script
- [x] YAML parses cleanly (`python3 -c "import yaml; yaml.safe_load(...)"`) for both changed workflow files
- [x] `Taskfile.yml` parses as valid YAML
- [ ] **Not runnable from this Windows dev machine** (no Linux, no `fpm`, no `appimagetool`) — needs a real CI run on `ubuntu-22.04` to confirm the `.deb`/`.tar.gz` builds actually succeed and the packages install/extract correctly. Recommend watching the first nightly run after merge (or a manual `workflow_dispatch` of `build-linux.yml`) before relying on this for a real release.


a5af:
@codex review

PR #3237, opened 2026-09-15
feat(mcp): agent-facing Global Memory API (GlobalMemoryWrite/List/Read/Remove) + audit trail

AgentA-asaf:
## Summary

Per direct request: "do you have an api to add a global memory entry?" — no, there wasn't one. This adds it, per `docs/specs/SPEC_AGENT_FACING_GLOBAL_MEMORY_API_2026_09_15.md`.

**Phase 0 — audit trail.** `db_bundle_versions`, an append-only version-history table for Global Memory content, added to all three schemas (objects.db v35, store.db v10, identity-store.db v9), mirroring `db_agent_native_memory_versions`' already-proven shape exactly (single `bundle_id` key instead of that table's composite `agent_id`+`filename`). Global Memory previously had **no audit trail at all** beyond `updated_at` — this was a real, confirmed gap (verified by a dedicated research pass, not assumed), and opening an agent-facing write path without it felt like the wrong order of operations.

**Phase 1 — REST routes.** `/api/v1/agent/globalmemory/{write,list,read,remove}`, same pattern and trust model as the existing `/api/v1/agent/memory/*` routes (agent identity from `agentmux-mcp`'s trusted `AGENTMUX_AGENT_ID` env, unforgeable from an agent's own PTY). Every route calls the **same** `Store::bundle_upsert`/`bundle_get`/`bundle_list_global` the human-facing Armory editor already uses — no new storage-layer write path.

**Phase 2 — MCP tools.** `GlobalMemoryWrite`/`GlobalMemoryList`/`GlobalMemoryRead`/`GlobalMemoryRemove`, same shape as the existing `Memory*` tool family. Tool descriptions explicitly distinguish this from `MemoryWrite` ("inherited by EVERY agent in this workspace... not your own private memory").

**Safety invariant, enforced twice independently:** nothing here ever reads or sets `is_system` from caller input, and `Store::bundle_upsert` (which every write goes through) separately still refuses to touch an existing `is_system=1` row regardless — the AgentMux-controlled system tier (`SPEC_GLOBAL_MEMORY_SYSTEM_TIER_2026_08_24.md`) stays completely unreachable through this new surface.

## Deliberately NOT done in this pass (flagged in the spec, not silently skipped)

- **Phase 3 (availability/gating) is left unresolved.** Every agent gets these tools unconditionally today — that's the default that falls out of shipping Phases 0-2 with no extra gating, matching native memory's own precedent, but it was never an explicit decision. The spec lays out three options (unconditional / notify-on-write / per-agent opt-in) without picking one.
- **`GlobalMemoryHistory`/`Diff`/`Revert` don't exist yet.** The read side of the audit trail (`bundle_version_list`/`bundle_version_get`) is implemented and unit-tested, just not exposed as a tool or in the Armory UI. Recording history with no way to see it yet is still strictly better than not recording it.

## Test plan

- [x] `cargo test -p agentmux-srv` — 3423 passed, 0 failed (includes 5 new `bundle_versions` tests: parent-chaining, per-bundle scoping, get-by-id, unknown-id handling, and a rename-changes-the-hash regression test).
- [x] `cargo test -p agentmux-mcp` — 26 passed, 0 failed (updated the tool-schema completeness test's hardcoded count from 44 to 48).
- [x] `cargo check --workspace` — clean, no errors.
- [ ] Not manually tested end-to-end against a live agent's actual MCP client (I don't have a way to drive that from here) — the REST→Store chain and the MCP tool JSON schemas are verified, but an actual `GlobalMemoryWrite` call from a running agent through the full stack hasn't been observed live.


a5af:
@codex review

AgentA-asaf:
Addressed all findings from both reviews in 664975c39:

**ReAgent P0 / Codex P1 (same bug, both caught it):** `global_memory_write_impl` accepted an arbitrary caller-supplied `id` and would overwrite that bundle's name/instructions + promote it to `is_global=true` with no check that the target was already global, or the blank singleton. Since bundle ids for *every* bundle are enumerable via `PresetList` (unfiltered), an agent could hijack another agent's private preset — or `'blank'` — into Global Memory. Fixed with both missing guards, mirroring `upsertmemory`'s own already-fixed blank-singleton guard.

**Codex P2 — trusted writer identity:** `db_bundle_versions` gained a `written_by` column, populated server-side from the trusted `AGENTMUX_AGENT_ID` (or `"armory-ui"`), independent of the caller-describable `source`/`source_detail` fields an agent could otherwise put anything in.

**Codex P2 — atomicity:** `Store::bundle_upsert_with_version` now does the `db_bundles` write and its `db_bundle_versions` row in one transaction instead of two separate lock acquisitions, closing the interleaving race under concurrent writers Codex described.

**Codex P2 — Armory UI writes weren't recorded:** the human-facing save path (`upsertmemory`, the actual handler behind the Global Memory editor) now goes through the same `bundle_upsert_with_version`, so operator edits are no longer invisible to the audit trail.

Since `db_bundle_versions` hasn't shipped yet (still unmerged in this same PR), `written_by` was added directly to the `CREATE TABLE` rather than a new migration/version bump.

10 new/updated regression tests (including one that reproduces the hijack bug and fails without the fix, and one confirming `written_by` can't be spoofed via `source`). Full suite: 3434 passed, 0 failed.

@codex review

AgentA-asaf:
Fixed in 59d580fbb: \`global_memory_remove_impl\` had the exact same missing-ownership-check bug already fixed for write, just not applied to remove — only checked \`is_system\`, never \`is_global\`/\`is_blank\`/\`id=="blank"\`. Same fix, same two guards, mirroring write's own. Good catch — I should have applied that fix to both functions in the same pass the first time instead of just write.

2 new regression tests (blank singleton, non-global bundle). Full suite: 3436 passed, 0 failed.

PR #3239, opened 2026-09-15
feat(agent-pane): make window-level tab switching near-instant

AgentY-asaf:
## Summary

Window-level tab switching had a noticeable delay even though panes host the same sort of DOM content Chrome switches between instantly. Full diagnosis in `docs/analysis/ANALYSIS_CHROME_PARITY_TAB_SWITCH_LATENCY_2026_09_15.md`, building on `SPEC_AGENT_PANE_TAB_SWITCH_PERF_2026_05_27.md`'s prior investigation, which had already measured the root cause: `display:none` removes an inactive tab's subtree from the layout tree entirely, so reactivating it forces the browser to lay out and paint it from zero — 500-600ms of browser-side work, confirmed by elimination to not be a JS-side cost.

## What changed

- **`display:none` → `content-visibility:hidden`** for inactive tabs — Chromium's actual documented primitive for "tab-like UI": skips rendering while inactive (same near-zero ongoing cost) but keeps cached layout state, so reactivating reuses it instead of starting over.
- **Two structural fixes** that swap required, both caught live while testing in a running `task dev` instance:
  - Every tab container is now permanently `display:flex` (content-visibility is what hides inactive ones now), which made all tabs flex siblings of the parent row simultaneously instead of only one at a time — every tab's content visibly **compressed horizontally** as more tabs opened. Fixed with `position:absolute` + `inset:0` on a `position:relative` parent, taking every tab out of flex flow entirely.
  - `content-visibility:hidden` doesn't imply `pointer-events:none` — an inactive tab's now-absolutely-positioned div still hit-tests, so whichever tab was **last in DOM order silently swallowed clicks** meant for whatever tab was actually showing underneath. Fixed with explicit `pointer-events` gating, active tab only.
- **A forced-synchronous-layout effect** (`getBoundingClientRect()` on activation), closing a reveal-gate timing gap found live-testing: `content-visibility`'s catch-up layout isn't guaranteed to happen synchronously when the style flips, so the existing settle detector (which only watches `PerformanceObserver` `longtask` entries) could decide a tab had "settled" before genuine layout/paint catch-up actually finished. The May 27 spec's own investigation log had already named this exact shape of gap under the *old* `display:none` mechanism ("a separate 500-600ms long-task fires AFTER the reveal gate lifts").
- **`document.startViewTransition()`** wrapping the reveal (feature-detected, scoped to just the tab-content region via `view-transition-name` so the window header/status bar never freeze into the snapshot), replacing the removed 120ms opacity fade with Chromium's own purpose-built cached-snapshot cross-fade.

## Verification

- `npx tsc --noEmit -p tsconfig.citypecheck.json` — clean
- `npx vitest run frontend/app/tab frontend/app/workspace frontend/app/store/tab-reveal.test.ts` — 91 pass, same 8 pre-existing environment-only failures present on unmodified `main` too (`pane-leaf-chrome.test.tsx`/`tab.test.tsx`, an unrelated `@solid-refresh` file-URL resolution issue)
- `npx stylelint frontend/app/app.scss` — clean on the added rule (one pre-existing, unrelated finding elsewhere in the file)
- Live-tested throughout against a running `task dev` instance with 4 real tabs open — this is how the two structural bugs above were actually caught

**Not yet done, deliberately:** a before/after benchmark against the recorded 500-600ms baseline (`SPEC_AGENT_PANE_TAB_SWITCH_PERF_2026_05_27.md`'s own `[perf] long-task` methodology). Repo owner's call — let this settle before re-approaching with numbers.

## Test plan

- [ ] CI green on `windows-latest` (required) and `ubuntu-latest`
- [ ] When benchmarking is picked back up: reuse the exact Phase 0 repro from `SPEC_AGENT_PANE_TAB_SWITCH_PERF_2026_05_27.md` (two tabs, ~50 and ~200 nodes, 5 switches each direction) so the result is directly comparable to the recorded baseline


a5af:
@codex review

PR #3241, opened 2026-09-15
Loap #2@narko: fix(test): compare the fs_watch stale-rearm probe's reason, not just its presence

GenericAgentX-asaf:
## Summary

Follow-up to **#3235** (already merged), addressing a **Codex P2** on it that is correct.

#3235 replaced a flaky process-global handle-count assertion with a degraded-flag probe — but asserted only `is_degraded()`, a boolean. `rearm_if_still_subscribed()` calls `clear_degraded` when `watch()` succeeds and `mark_degraded` when it **fails**, so the probe conflated two different outcomes:

| case | flag | old assertion |
|---|---|---|
| guard works (early return) | stays set | passes ✓ |
| guard bypassed, `watch()` succeeds | cleared | fails ✓ |
| **guard bypassed, `watch()` fails** | **stays set** | **passes ✗** |

That last row — reachable via permissions, resource exhaustion, or an unavailable backend — is exactly the passes-under-broken-logic failure mode #3235 existed to remove.

**Fix:** compare the degraded **reason** against a sentinel rather than testing for presence. That separates all three outcomes: sentinel intact = returned early (correct); entry gone = `watch()` succeeded; any other string = `watch()` failed. The last two both mean the guard was bypassed.

Also adds a `watcher.is_some()` precondition: `FsWatchPool::new()` documents `watcher: None` as reachable, and that makes `rearm_if_still_subscribed()` return early at a *different* guard — passing the probe for a reason unrelated to what this test checks.

No production code changes; `HealthState::snapshot()` already exposes the reason.

## Test plan

- [x] **Verified the newly-covered case directly.** With the re-check bypassed *and* `watch()` forced to fail (target directory removed first), the old boolean assertion passed while this one fails with:
  ```
  left:  Some("Input watch path is neither a file nor a directory.")
  right: Some("probe sentinel — must survive a correctly-guarded stale rearm")
  ```
- [x] Still catches the original bug (guard bypassed, `watch()` succeeds).
- [x] Passes unmodified; full non-ignored suite green across two consecutive runs (3418 passed, 0 failed).


a5af:
@codex review

PR #3245, opened 2026-09-16
Opaz@charlie: fix(lan-discovery): stop wiping peer identity on mDNS ServiceRemoved churn

GenericAgentX-asaf:
## Summary
- mDNS `ServiceRemoved` fires on ordinary TTL/interface churn, not just genuine peer departure. Live logs caught it firing on a real peer (`gamerlove`) seconds before it re-announced.
- The old handler deleted the peer's cache entry on `ServiceRemoved`. Since almost every re-resolution after that carries a blank TXT record (a known `mdns-sd` quirk this file already works around for other cases), the peer's hostname/version/auth_key could never be restored — it stayed blank forever, surfacing in the UI as a peer with no name and a bare "v" version tag.
- `ServiceRemoved` is now inert. `get_instances()` instead hides a peer once `last_seen` exceeds a new `LAN_PEER_STALE_TIMEOUT_SECS` (300s), and the existing 30s agent-name refresh loop garbage-collects those same stale entries so the underlying map doesn't grow unbounded or keep polling dead peers.
- `HostPopover.tsx` and `warden-lan-manager.tsx`: show "resolving…" instead of a blank peer name, and omit the version tag until a version is actually known (fixes the bare "v" display bug).

## Test plan
- [x] `cargo test -p agentmux-srv --bin agentmux-srv lan_discovery` — 45 passed (43 pre-existing + 2 new: `service_removed_does_not_delete_a_known_peer`, `get_instances_hides_a_peer_stale_past_the_timeout`)
- [ ] Manual: fresh build, confirm a real LAN peer (`gamerlove`) resolves to its real hostname/version in the status bar and Warden LAN panel instead of "v"


a5af:
@codex review

GenericAgentX-asaf:
Codex's P1 (missing broadcast after the stale-peer GC prunes something) is fixed in 9f71127 — tracked via `any_pruned` and calls `broadcast_instances()` when the retain actually removed an entry.

PR #3246, opened 2026-09-16
docs(retro): stale local CEF-148 cache under a CEF-152 binding crashed 0.56.0 on launch

AgentO-asaf:
## Summary

Retro for a P1 found while investigating why `AgentMux_0.56.0_arm64.dmg` showed only the splash screen on launch. Root cause: `task package:macos`'s tier-2 CEF resolution (`~/cef-build/darwin/arm64`) picked up a personal hermetic build cache that predates the CEF 148→152 migration (#3108/#3231) — the Rust glue compiled correctly against `main`'s CEF 152 dependency graph, but the bundled native framework was still 148, so `agentmux-cef` panics at startup with `dlsym(cef_version_full): symbol not found`.

**This is not a mistake in the cross-platform 152 migration itself** — Windows/macOS/Linux were all independently verified (real H.264 playback, ANGLE export-table checks, ABI struct-size guards) and the official `v0.56.0` GitHub release is correct end to end. The failure is scoped to one machine's stale local build cache, hit through the one packaging path (`task package:macos`) that doesn't share CI's hard `scripts/verify-cef-version.sh` gate — only an easy-to-miss `stderr` warning (`check_version()` in `resolve-cef-runtime-darwin.sh`, added on #3231 in direct response to Codex flagging this exact risk class, but deliberately warn-not-fail).

Full writeup: `docs/retro/retro-macos-0560-stale-cef-cache-launch-crash-2026-09-16.md`

## Test plan

- [x] Doc-only change, no code touched
- [x] Reproduced the crash directly (ran `agentmux-cef` by hand with reconstructed launcher env vars) and confirmed root cause via `ctypes`/`dlsym` against both the stale cache and the currently-running 0.55.43 framework
- [x] Cross-referenced against #3108's full thread to confirm this isn't a repeat of an already-known/already-fixed issue


a5af:
@codex review

AgentO-asaf:
Checked both findings directly against the raw file content at the exact reviewed commit (`d0a8d53e0`, via the GitHub contents API, not my local copy) — both are actually addressed in that commit, not "unaddressed":

**Finding 1 (arm64/aarch64):** current text (verified via API fetch of `d0a8d53e0`):
> tier 2 — `$HOME/cef-build/darwin/$ARCH`, where the script maps `uname -m`'s `arm64` to `ARCH=aarch64` (Rust's `target_arch` convention, not Apple's), so on this machine it probes `~/cef-build/darwin/aarch64` — a symlink to `~/cef-build/darwin/arm64`, the actual hermetic CEF build cache directory

**Finding 2 (verify-cef-version.sh scope):** current text:
> `verify-cef-version.sh` reads a bundled `libcef.dll`'s `ProductVersion` via `pwsh`... it is **Windows-only**... invoked from `build-windows.yml`/`bundle:windows`, and could never run against a macOS `.framework` at all.

Both were rewritten in the `d0a8d53e0` commit specifically in response to Codex's prior-round findings — see that commit's own message. My best guess is this re-review's line-number references (`:20`, `:132`) point at the *pre-fix* positions, since the fix added a `**Status:**` line near the top that shifted everything below it down by one line, and the 11s review time (`DOCS_ONLY`, low effort) suggests a fast pass rather than a full re-read of the new diff. Flagging in case that's a real gap in the review tool's line-tracking across a re-review, not asking anyone to trust me over it blindly — happy to make further edits if a closer look shows I'm actually still missing something, but I can't find it looking at the actual committed content.

PR #3247, opened 2026-09-16
fix(fonts): drop bundled JetBrains Mono — its calt ligatures blanked repeated punctuation on CEF 152

AgentY-asaf:
> **Note:** this PR was force-pushed with a different approach after review.
> Earlier comments target CSS rules that no longer exist on the branch — see
> the "Approach changed" comment below.

## Summary

Holding `.` (or `;` `:` `?` `!` `#` `&` `+` `_` `=` `-` `|`) renders a run of
blank cells with a single glyph at the end — `"            ."` — in **every**
AgentMux text surface: agent composer, editor, filter bars, swarm broadcast
field, browser address bar. The underlying value is unaffected (messages reach
agents with every character intact) — only the rendering is wrong.

Reproduces on CEF 152 (shipped in v0.56.0) on Windows **and** Ubuntu. CEF 148 is
clean.

## Root cause

JetBrains Mono implements programming ligatures the way monospace fonts must: to
ligate `..` without breaking the character grid, it maps the first period to
**`glyph00388` — a glyph with no contours** (invisible, still one cell wide) and
the second to a composed glyph whose outline extends *leftward* to draw both
dots. Same total advance, one symbol drawn.

The font **guards** this so it only applies to runs of 2–3. OpenType contextual
rules are tried in order and first match wins, so it places no-op rules ahead of
the substituting ones:

```
lookup 59 (the ".." ligature)
  rule 0: back=.  input=.  ahead=.   -> NO-OP GUARD  "mid-run: do nothing"
  rule 1:         input=.  ahead=..  -> NO-OP GUARD  "longer run ahead: do nothing"
  rule 4:         input=.  ahead=.   -> SUBSTITUTES (-> blank)
```

**CEF 148 honored those guards. CEF 152 does not**, so the substitution runs
across an entire run — many blanks, one composed glyph at the end that can only
draw 2–3 dots.

The font is correct and unchanged; this is an **upstream shaper regression**, and
not specific to JetBrains Mono — the guard technique is standard, so Fira Code /
Cascadia Code / Iosevka are affected identically.

## The fix: remove the font, not the feature

JetBrains Mono is the only font we bundled carrying these `calt` rules. Hack —
our actual mono default — has **no `calt` table at all**. And nothing ever
selected JetBrains Mono deliberately: it appears *only* inside the fallback of
`var(--termfontfamily, …)`, a variable that **is never defined anywhere in the
repo**. With the font unloaded, those rules fall through to system `monospace`
(Consolas / DejaVu Sans Mono), which have no programming ligatures.

### Why not disable `calt` in CSS

Tried twice on this branch and abandoned as structurally unsound. The `font`
shorthand resets `font-variant-ligatures`/`font-feature-settings`, and this
codebase has **19 `font:` shorthand declarations**, each re-enabling `calt` on
its element at a specificity the override cannot beat:

- attempt 1 (`html, body, input, …`) lost to `@layer base`'s own
  `input,button,textarea,select { font: inherit }` — caught by codex P1
- attempt 2 (`*, *::before, *::after` + `body` re-declaration) lost to
  `.pane-tab-rename-input`, `.block-frame-view-type--input`,
  `.agent-user-message-summary` — all `.class` (0,1,0) beating `*` (0,0,0),
  caught by ReAgent P1/P2×2

Reviewers had found 4 of 19. Winning needs `!important` on a universal selector
or 19 edits plus permanent vigilance. Deleting one font is smaller and total.

## Evidence

Verified offline with fontTools + HarfBuzz against the bundled woff2:

```
period      advance=600  bounds=(217,-10,383,156)   <- a real dot
glyph00388  advance=600  bounds=None                <- NO CONTOURS (blank)
glyph00194  advance=600  bounds=(-301,-10,303,156)  <- composed, draws leftward
```

Which characters break was **predicted from the `calt` tables before most were
tested**, matching observation **22 of 23**:

- broken (has a blank-glyph rule): `.` `;` `:` `?` `!` `#` `&` `+` `_` `=` `-` `|`
- fine (no such rule): `,` `@` `$` `%` `^` `(` `)` `\` `{` `}`

The comma is decisive — the only sentence punctuation with no blank-glyph rule,
and the only one that renders. Every broken character is half of a
programming-operator ligature.

## Also in this PR

Two references that advertised a font we no longer ship: the `term:fontfamily`
settings placeholder, and `AgentInstallModal`'s xterm fallback string.

## Testing

- Repo owner reproduced on CEF 152 (Win + Ubuntu) and confirmed CEF 148 is clean
- Character-level prediction from the font tables matched observation 22/23
- `tsc --noEmit` clean; no tests reference the removed font
- ⚠️ The earlier live confirmation covered form controls only — the **editor**
  needs re-checking against this branch (it was never fixed by the CSS attempt)

## Follow-up (PR 2)

Font-variable consolidation. **53 of ~108 font-variable usages reference
variables that are never defined** (seven names for "the mono font"), and **44
pass a `font` shorthand to `font-family`** — invalid CSS, silently dropped. Also
removes the 19 `font:` shorthands, after which a VS Code–style
`editor.fontLigatures` opt-in becomes implementable.

Retro: `docs/retro/retro-terminal-consecutive-period-input-loss-2026-09-15.md`


a5af:
@codex review

AgentY-asaf:
## Approach changed — force-pushed, prior reviews superseded

Thanks @codex and ReAgent — both P1s were correct, and together they showed the
original approach was structurally unsound rather than just incomplete.

**What the reviews found:**
- codex P1: `app.scss`'s `body { font: var(--base-font) }` resets the longhands
  after `@use "reset.scss"`, so the editor (CodeMirror's contenteditable
  `<div>`, not a form control) never got the fix.
- ReAgent P1/P2×2: `.pane-tab-rename-input`, `.block-frame-view-type--input` and
  `.agent-user-message-summary` all use `font: inherit` at `.class` specificity
  (0,1,0), beating even a universal-selector override (0,0,0).

**Why that's fatal to the CSS approach:** the `font` shorthand resets
`font-variant-ligatures`/`font-feature-settings`, and this codebase has **19
`font:` shorthand declarations**. Every one re-enables `calt` on its element,
and reviewers had found 4 of 19. Winning requires `!important` on a universal
selector, or edits to all 19 sites plus permanent vigilance against the 20th.

**New approach: delete the font instead.** JetBrains Mono is the only font we
bundled carrying these `calt` rules. Hack — our actual mono default — has no
`calt` table at all. And nothing ever selected JetBrains Mono deliberately: it
appears *only* inside the fallback of `var(--termfontfamily, …)`, a variable
that is never defined anywhere in the repo. With the font unloaded, those rules
fall through to system `monospace`, which has no programming ligatures.

One deleted font beats 19 CSS edits, and there is no specificity battle left to
lose. Removing the input to a bug rather than suppressing its output.

The diff is now: 3 woff2 files + their loader, two stale references that
advertised a font we no longer ship, the changeset, and the retro.

**Please re-review from scratch** — the CSS rules you commented on no longer
exist on this branch.

AgentY-asaf:
## Verified live, including the editor

Repo owner confirmed on a `task dev` build from this branch (CEF 152): repeated
punctuation now renders correctly in the **editor**, which was the one surface
still unverified.

Worth stating why that mattered: the earlier CSS-based attempt on this branch
appeared to work when tested, but only because the surfaces exercised were form
controls — codex's P1 showed it never reached CodeMirror's contenteditable
`<div>`. So "it works" had already been wrong once here. This confirmation is
against the font-removal approach that actually shipped, on the surface that
previously fell through the gap.

Unrelated issue seen in the same session and **not** addressed here: the editor's
first keystroke after clicking in is accepted but not rendered (the dirty marker
appears, the character doesn't) until you click a second time. Different cause —
specced separately in #3250.

PR #3248, opened 2026-09-16
fix(muxbus): skip automatic cloud-session reconnect on non-stable channels

AgentO-asaf:
## Summary

- User-reported: opening a freshly local-built `AgentMux_0.56.0_arm64.dmg` prompted for the macOS Keychain password, even though the underlying muxbus prompt-storm issue was already fixed on 2026-08-19.
- Root cause: TWO separate automatic (non-explicit-user-action) code paths do a real, synchronous `Store::muxbus_load()` OS-keychain read:
  1. `CloudSubscriber::init_global` — runs unconditionally on every app launch.
  2. `inject_muxbus_env` — runs unconditionally on **every agent spawn** (not just MuxBus-related ones).
- Every local `task package`/`task package:macos`/`task package:linux` build bakes a brand-new, randomized per-build channel (and, on macOS, a bundle identifier derived from it) into the binary — so every fresh local build looks like a never-before-seen app to Keychain, and the SAME already-`Always Allow`'d global `muxbus:global` credential (confirmed: MuxBus login is deliberately global/cross-channel by design, same as the identity store it lives in) prompts again on every single local rebuild, from either automatic path, indefinitely.
- This is the same class of gap Armory accounts and `settings.json` already closed (`SPEC_ISOLATED_AUTH_DEFAULT_BY_CHANNEL`, `SPEC_SETTINGS_ISOLATED_BY_CHANNEL`) — both default to isolated on any non-`stable` channel. Neither of MuxBus's two automatic reads got the same treatment before this PR.

## Change

- Adds `isolated_muxbus_reconnect_enabled()`/`isolated_muxbus_reconnect_reason()` to `agentmux-common/src/data_paths.rs`, mirroring `isolated_auth_enabled`/`isolated_settings_enabled` exactly: `AGENTMUX_ISOLATED_MUXBUS=1`/`=0` explicit override, channel-default-isolated-unless-`stable` otherwise, same fail-safe-on-malformed-value behavior, same independence-from-the-other-two-flags guarantee.
- Wires it into **both** automatic read sites:
  - `CloudSubscriber::init_global`'s call site in `bootstrap.rs`, alongside the existing `AGENTMUX_DISABLE_CLOUD_SUBSCRIBER` test-only flag.
  - `inject_muxbus_env` in `muxbus_handlers.rs` — early-returns before touching the store at all on an isolated channel.
- Real installs (`stable` channel) and `task dev` keep muxbus reconnect-on-launch AND spawn-time token injection working exactly as before — a real user's cloud session still resumes and flows into every spawned agent automatically. A local build now simply starts with no MuxBus session, and neither automatic path ever touches Keychain, until you explicitly `muxbus.login` inside that specific build.

Retro for the incident this closes: `docs/retro/retro-macos-0560-stale-cef-cache-launch-crash-2026-09-16.md` (separate PR #3246).

## Test plan

- [x] `cargo check -p agentmux-common -p agentmux-srv` — clean, no new warnings, twice (once per commit)
- [x] New tests (`isolated_muxbus_reconnect_reason_classifies_all_four_states`, `..._fails_safe_on_a_malformed_value_on_a_non_stable_channel`, `..._is_independent_of_auth_and_settings_isolation`) — all pass
- [x] Full `cargo test -p agentmux-common` — 194 passed, 0 failed
- [x] Changeset added (`patch`)
- [ ] `inject_muxbus_env` has no pre-existing test scaffolding in this file to extend (zero tests in `muxbus_handlers.rs`/`muxbus.rs` today) — covered instead by the underlying flag's exhaustive test suite; the call site itself is a one-line early-return


a5af:
@codex review

PR #3249, opened 2026-09-16
fix(term): take the per-keystroke SQLite read off the input path

Agent1-asaf:
## Summary

Cross-pane typing lag is back, but **not** because the September fix regressed — `fair_drain_priority`/`priority_pane_key`/the `pending_priority` trickle loop from #2973 are all intact (I diffed all 8 commits that have touched `websocket.rs` since it landed; none touch that machinery).

It's a new path. #3194's PtyShell lease check runs in `controllerinput` — the handler that processes **every keystroke in every pane on every connection** — and answered "is this block leased?" with `wstore.get::<Block>()`. `Store` is one process-wide SQLite connection behind one `Mutex<Connection>` ("matching Go's `MaxOpenConns(1)`", its own doc comment), and `Store::get` is a plain blocking call invoked inline on a Tokio worker thread with no `block_in_place`/`spawn_blocking`.

So each keystroke paid a disk query serialized against every other `Store` user in the process — other panes' `controllerinput` calls, `PtyShellInput`'s own lock write, background-task and turn bookkeeping. Same failure class as the sysinfo incident (`0f34704a8`, #1782), on a hot path that fix never touched.

Full diagnosis, including what was ruled out: `docs/analysis/ANALYSIS_CROSS_PANE_INPUT_DELAY_REGRESSION_2026_09_15.md`.

## What changed

New `backend/blockcontroller/agent_lock.rs` holds the lease in memory (`block_id -> expiry_ms` behind a mutex held for nanoseconds, no I/O inside, self-pruning on read). `controllerinput` reads that — for the common case (no agent driving any shell) it's a hash lookup that misses. `lock_shell_for_agent`/`PtyShellStop` write and clear memory first, then do the `term:agentlockuntil` meta write as before; meta is now explicitly the frontend-badge copy, which is what that handler's comment already claimed it was.

Chose this over wrapping the call in `block_in_place`: that only makes the blocking call *politely* blocking — still a DB round trip, still serialized on the global mutex, per keystroke, for a question that's almost always "no."

## The `exit` half

Enforcement in memory is also what makes the lease safe around a human typing `exit`, now the default for shell panes (`SPEC_TERM_EXIT_RESPAWN_LOOP_2026_09_15.md` §10). Three properties, each tested, each **falsified** (source deliberately broken, right test observed failing, restored):

1. **A shell that exits releases its lease immediately** (`shell/lifecycle.rs`, alongside publishing `STATUS_DONE`). Without it, a pane whose agent wrote within the last 4s keeps dropping the human's keystrokes after the PTY is gone — including `exit` itself. → `an_exited_shell_does_not_keep_the_agent_lease` (real-PTY, `#[ignore]`d like its siblings; the lease it seeds is deliberately longer than the real 4s window, or the test would be measuring the clock rather than the release wiring).
2. **A lease always lapses on its own**, with nobody calling `PtyShellStop` — an agent that crashes mid-write can't strand a human. → `controllerinput_flows_again_once_the_agent_lease_lapses`.
3. **A stale `term:agentlockuntil` in meta locks nothing by itself.** Deliberate fail-open: meta outlives the process that wrote it, memory doesn't, and a lease surviving a restart means a human who can't type `exit` in their own terminal.

**Not changed:** whether a *live* lease drops human input. It should — that's the write-collision guard, bounded at 4s, and the only panes that can be leased (an agent's own composer-drawer shell, per `is_owned_by_agent`) already show the "Agent is using this shell" badge. A standalone Terminal pane can never be leased.

## Test plan

- Full `agentmux-srv` suite: **3445 passed / 0 failed**
- `cargo clippy -p agentmux-srv --tests`: clean on the new file
- Real-PTY test run manually: `cargo test -p agentmux-srv --bin agentmux-srv an_exited_shell_does_not_keep_the_agent_lease -- --ignored` ✅
- Each of the three falsifications confirmed the expected test fails

## Not verified

**No live measurement.** No dev instance was available this session, so there's no before/after P95 number the way the 09-04 analysis has one, and the agent-drives-shell-then-human-types-`exit` sequence hasn't been exercised in a real pane. §6 of the analysis lists exactly what's left to measure — including that `tools/tests/bench-term-cross-pane.mjs` floods *output* only and cannot see this class of bug by construction.


a5af:
@codex review

PR #3250, opened 2026-09-16
docs(editor): spec the first-keystroke-not-rendered bug

AgentY-asaf:
## Summary

Spec only — no code change. Filed separately from the CEF 152 font work (#3247)
so the two aren't conflated; they have unrelated causes.

**Symptom:** click into an editor pane and type — nothing appears. But the
**dirty indicator does appear** next to the filename. Click in again and typing
works normally.

## Why that one detail settles a lot

The keystroke *is* reaching CodeMirror: the document changed, the model marked
the tab dirty, the tab re-rendered. Only the visible text is missing. So:

- **Not a focus/keyboard-routing bug** — an unfocused editor couldn't flip dirty.
- **Not the CEF 152 shaping bug** from #3247 — that renders blanks of the right
  width, for repeated punctuation only, across every text surface. This is the
  first keystroke of any character, editor only.

CodeMirror's **state updates; its view doesn't paint.**

## What the code shows

- `editor-view.tsx` never calls `cmView.focus()` — the only `.focus()` calls in
  `view/editor/` are the two rename inputs.
- The `createEffect` at ~466 restores snapshots via `cmView.setState(saved)`,
  which replaces the whole view state.
- `editor-model.ts:369` derives the **pane title** from dirty (`${fp} *`), so
  the first keystroke fires a reactive cascade through pane chrome and the tab
  strip — including the close/dirty affordance the reporter suspected.

## Three candidate mechanisms, none asserted

1. **C1** — the tab-switch effect re-runs on first edit and `setState`s a
   pre-keystroke snapshot (explains dirty-but-reverted precisely)
2. **C2** — stale measurement from a container that was zero-sized at
   construction (the effect's own comment mentions a first-open race)
3. **C3** — the dirty affordance appearing shifts layout on exactly that
   keystroke, invalidating cached geometry (reporter's hypothesis)

C1 and C3 may compound. Each has a concrete test in the spec.

## Deliberately instrument-before-fixing

The companion retro is a case study in what picking a mechanism on plausibility
costs: four wrong hypotheses and three reviewer P1s. This one has an easy
plausible story per mechanism, so the spec's first work item is instrumentation,
not a patch.

## Open questions in the spec

Notably **Q3: if you save without clicking a second time, does the first
character reach disk?** If not, this is data loss and severity rises.


a5af:
@codex review

PR #3252, opened 2026-09-16
refactor(fonts): consolidate the font variable system onto two canonical tokens

AgentY-asaf:
> **Stacked on #3247** — base is that branch, not `main`. Review #3247 first;
> this diff shows only the consolidation on top of it.

## Summary

The CEF 152 ligature investigation surfaced that the font system was largely
**inert**. Two separate classes of silent failure, both fixed here.

### 1. Fifty-three usages referenced variables that don't exist

Seven different spellings for "the monospace font", five never defined anywhere:

| Variable | Uses | Defined? |
|---|---|---|
| `--termfontfamily` | 26 | ❌ never |
| `--monospace-font` | 10 | ❌ never |
| `--mono-font` | 7 | ❌ never |
| `--mono-font-family` | 3 | ❌ never |
| `--monofontfamily` | 2 | ❌ never |
| `--termfontsize` | 2 | ❌ never |
| `--main-font-family` | 1 | ❌ never |

Every one silently fell through to a hardcoded fallback — and the fallbacks
**disagreed**: 14 fell back to bare `monospace` (Consolas/DejaVu), 11 to
JetBrains Mono. Different parts of the UI rendered in different fonts by
accident.

This is also why **`term:fontfamily` appeared to work in the terminal but
nowhere else**: `term.tsx:193` reads the setting directly in JS, while the 26
CSS rules expecting `--termfontfamily` got nothing, because nothing ever wrote it.

### 2. Forty-four rules passed a `font` shorthand to `font-family`

`--fixed-font` is a **shorthand** (`normal 12px / normal "Hack", monospace`), but
44 rules did `font-family: var(--fixed-font)`. That expands to
`font-family: normal 12px / normal "Hack", monospace` — **invalid CSS**. The
declaration is dropped at computed-value time and the element silently inherits.
Those 44 declarations had no effect at all.

## What this does

- **`--font-sans` / `--font-mono` are the only family tokens**, defined once in
  `tailwindsetup.css`'s `@theme`. Tailwind v4 needs the literal there to generate
  the `font-sans`/`font-mono` utilities (used by `quicktips.tsx`), and `@theme`
  emits into `:root` so the app's own SCSS resolves against the same values.
  They were previously declared in **both** `theme.scss` and `tailwindsetup.css`
  with **different values** — which won depended purely on emit order.
- **`--base-font` / `--fixed-font` stay** as `font` shorthand composites for the
  10 legitimate `font:` uses, now built from the canonical tokens plus new
  `--font-sans-size` / `--font-mono-size`, so family and size are defined once.
- Mono stack follows **VS Code's model**: system fonts plus Hack, which we bundle
  for Nerd Font glyph coverage in terminals.

## Also: fixes a false positive in the gate from #3247

`check-ligature-fonts.sh`'s comment exclusion only skipped lines *beginning* with
`//` or `/*`, so a continuation line inside a multi-line block comment tripped
it — caught when **the gate failed on this PR's own documentation** warning
people not to add ligature fonts. It now blanks comments out (preserving line
numbers) before matching. Re-verified both directions: passes clean, still fails
on a real regression.

## Verification — against the built CSS, not inferred

```
dist/frontend/assets/style-*.css:
  --font-mono:"Hack", Consolas, Menlo, monospace   <- defined
  references to deleted tokens:            0
  font-family:var(--font-mono):           55       <- now valid
  font-family:var(--fixed-font):           0       <- invalid ones gone
  font: shorthand (legitimate):           10
```

`vite build` exits 0; `tsc --noEmit` clean. Local vitest shows 29 failures from a
pre-existing `solid-refresh`/jsdom environment issue that **reproduce with these
changes stashed** — CI's vitest is authoritative.

## Not in this PR

Wiring `term:fontfamily` / `term:fontsize` to override `--font-mono` at runtime,
and the VS Code–style `editor.fontLigatures` opt-in. Both are now *possible*
(there is finally a single token to override), but they are behavior changes
rather than consolidation, and belong in their own PR.


a5af:
@codex review

AgentY-asaf:
## Both P1s fixed — and they were the same omission

@codex and ReAgent independently caught that the codemod collapsed
`font-size: var(--termfontsize, 15px)` onto `--font-mono-size` (12px) at
`.agent-view` and `_document-nodes.scss:72`. Since `--termfontsize` was never
defined, the 15px fallback was *always* what applied — so that was a universal
**20% shrink of agent-pane text**, and it silently reverted the explicit
readability default in `SPEC_PANE_ICON_AND_TEXT_VISIBILITY_2026_05_30.md`.

The framing in ReAgent's review is the part worth keeping: unlike the font-family
fallbacks, which genuinely **disagreed** across call sites and therefore had to be
reconciled, both size call sites already **agreed** on 15px. There was no
inconsistency to fix — only a name to give it. Consolidation isn't license to
change an agreed value.

**Fixed:** added `--font-agent-size: 15px` as its own token; both sites restored.
Verified in the built CSS — `--font-agent-size: 15px` defined, 2 occurrences of
`font-size:var(--font-agent-size)`, `--font-mono-size` still separately 12px.

## Behavior changes this PR *does* make (the omission you flagged)

You were right that the description hid a visible change behind "refactor". I
audited every fallback the codemod replaced:

| Old fallback | Sites | Becomes | Visible? |
|---|---|---|---|
| `monospace` | 30 | `"Hack", Consolas, Menlo, monospace` | **yes** — system mono (Consolas/DejaVu) → Hack |
| `"Hack", Consolas, Menlo, monospace` | 14 | identical | no |
| `15px` | 2 | `--font-agent-size` (15px) | no — was the regression above |
| `sans-serif` | 1 | `--font-sans` (Inter) | **yes** — generic sans → Inter |

So **31 sites change appearance.** That is the intended result of unifying seven
accidental spellings onto two tokens — previously, which font you got depended on
which dangling variable a rule happened to name — but it is a real visual change
and now says so in the PR description rather than being implied by "refactor".

AgentY-asaf:
Superseded by #3254 — same branch, rebased onto `main`.

GitHub auto-closed this PR when its base branch (`agenty/fix-calt-repeated-punctuation-cef152`) was deleted during the #3247 merge, and refuses to reopen a PR whose base no longer exists. My sequencing error: a stacked PR must be retargeted to `main` **before** merging its base with `--delete-branch`.

Both P1s raised here (@codex and ReAgent, on the agent-pane font-size shrink) are fixed in #3254, along with the behavior-change table you asked for.

PR #3253, opened 2026-09-16
feat(agent-pane): typing `exit` in the shell drawer closes the shell and collapses the drawer

Agent1-asaf:
## Summary

Open the shell drawer in an agent pane, type `exit`: the shell dies, and the drawer stays open around a dead terminal that renders its final scrollback and silently accepts keystrokes going nowhere. The only way out is closing the drawer by hand.

Spec (first commit) + implementation (second). Fixes it so the shell exits *and* the drawer collapses — matching what typing `exit` in a standalone Terminal pane now does (`SPEC_TERM_EXIT_RESPAWN_LOOP_2026_09_15.md` §10).

## Why the backend can't do this

`lifecycle.rs` is **untouched**. Close-on-exit deliberately excludes sub-blocks (`close_on_exit = … && !is_sub_block`) because its close action is `sagas::delete_block::run` — which deletes the block and prunes the owning tab's layout tree. A drawer shell has no layout entry to prune, and deleting it dangles the parent's `term:shellsubblockid`, whereupon the drawer's own attach-or-create path spawns a replacement moments later. That's the respawn loop §11 of that spec fixed, re-entered through a different door.

So the collapse is driven from the frontend, off the `controllerstatus` event the backend already publishes and the frontend already consumes elsewhere.

## What changed

- **`AgentShellSubblock.tsx`** subscribes to `controllerstatus` for its **own sub-block id** and fires a new `onShellExited` prop. Subscribed from a `createEffect`, not `onMount` — a freshly created shell has no id at mount (the async IIFE assigns it), so a one-shot mount subscription would bind an empty scope and never fire, i.e. the *common* first-open case would silently do nothing. Fires at most once per mount, since `STATUS_DONE` can be republished and the parent's handler tears down real state.
- **`shell-exit-collapse.ts`** (new) holds the parent-side response: drop the write closure, collapse, clear the pointer, delete the sub-block. Split out of `agent-view.tsx` purely to be testable — that file is ~3000 lines with no render harness, and each effect has its own failure mode (dangling pointer → respawn; undeleted block → leak; open drawer → the original bug), so they're asserted separately rather than as one "it collapsed".
- **`agent-view.tsx`** reads-and-clears `shellSubBlockIdRef` before delegating, so the pane-level `onCleanup` can't later delete a block this already removed.

## Clean exits only

Resolves the spec's §5 open question. A crash or backend restart publishes the same `STATUS_DONE`, and collapsing there would hide the output the human needs to read — so a non-zero exit leaves the drawer open with its scrollback intact. An **absent** exit code counts as clean: the field is `#[serde(default)]` on the wire, so an ordinary `exit` can arrive omitted, and reading that as "unknown, assume crash" would break the main case. Both branches tested.

## Test plan

12 tests, **every one falsified** (break the source, confirm that specific test fails, restore):

- `AgentShellSubblock.test.tsx` (7): clean exit fires once · omitted exit code counts as clean · non-zero exit does NOT collapse · `running` ignored · a different block's event ignored (the pane subscribes to its own block's status separately — a parent-scope event collapsing the drawer would fire every time the agent's process ended) · repeated publishes fire once · unsubscribes on unmount.
- `shell-exit-collapse.test.ts` (5): the three effects separately, plus the no-sub-block-id case.

Frontend suite **4047 passed / 0 failed**; `tsc` clean. (20 suite-level failures locally are all inside an untracked stray `amramebgi/` directory unrelated to this repo's tests.)

## Not verified

**No live run.** No dev instance was available this session, so this path is unit-tested end to end but has never met a real PTY in a real drawer.


a5af:
@codex review

PR #3254, opened 2026-09-16
refactor(fonts): consolidate the font variable system onto two canonical tokens

AgentY-asaf:
> **Supersedes #3252**, which GitHub auto-closed when its base branch was deleted
> during the #3247 merge (my sequencing error — it should have been retargeted to
> `main` first). Same branch, rebased onto `main`; #3252 carries the review
> history, including two P1s from @codex and ReAgent that are fixed here.

## Summary

Follow-up to #3247. That investigation surfaced that the font system was largely
**inert**, in two distinct ways.

### 1. Fifty-three usages referenced variables that don't exist

Seven spellings for "the monospace font", five never defined anywhere:
`--termfontfamily` (26 uses), `--monospace-font` (10), `--mono-font` (7),
`--mono-font-family` (3), `--monofontfamily` (2), `--termfontsize` (2),
`--main-font-family` (1).

Every one silently fell through to a hardcoded fallback — and the fallbacks
**disagreed**: 30 said bare `monospace`, 14 said the Hack stack. Which font a
given surface rendered in depended on which dangling variable its rule happened
to name.

This is also why **`term:fontfamily` worked in the terminal but nowhere else**:
`term.tsx:193` reads the setting directly in JS, while the 26 CSS rules waiting
on `--termfontfamily` got nothing, because nothing ever wrote it.

### 2. Forty-four rules passed a `font` shorthand to `font-family`

`--fixed-font` is a **shorthand** (`normal 12px / normal "Hack", monospace`), yet
44 rules did `font-family: var(--fixed-font)` — invalid CSS, dropped at
computed-value time, silently inheriting instead. Those declarations did nothing.

## What this does

- **`--font-sans` / `--font-mono`** are the only family tokens, defined once in
  `tailwindsetup.css`'s `@theme` (Tailwind v4 needs the literal there for the
  `font-sans`/`font-mono` utilities, and `@theme` emits into `:root`). They were
  previously declared in **both** `theme.scss` and `tailwindsetup.css` with
  **different values** — which won depended purely on emit order.
- **`--base-font` / `--fixed-font`** remain `font` shorthand composites for the
  10 legitimate `font:` uses, built from the canonical tokens plus
  `--font-sans-size` / `--font-mono-size`.
- **`--font-agent-size: 15px`** as its own token (see below).
- Mono stack follows VS Code's model: system fonts plus Hack, bundled for Nerd
  Font glyph coverage in terminals.

## Behavior changes (stated explicitly — this was the reviewers' point)

| Old fallback | Sites | Becomes | Visible? |
|---|---|---|---|
| `monospace` | 30 | `"Hack", Consolas, Menlo, monospace` | **yes** — system mono → Hack |
| `"Hack", Consolas, Menlo, monospace` | 14 | identical | no |
| `15px` | 2 | `--font-agent-size` (15px) | no — preserved |
| `sans-serif` | 1 | `--font-sans` (Inter) | **yes** — generic sans → Inter |

**31 sites change appearance.** That is the intended result of unifying seven
accidental spellings onto two tokens, but it is a real visual change and says so
here rather than hiding behind "refactor".

## The two P1s from #3252, fixed

Both reviewers caught that the codemod collapsed
`font-size: var(--termfontsize, 15px)` onto `--font-mono-size` (12px) at
`.agent-view` and `_document-nodes.scss:72`. Since `--termfontsize` was never
defined, 15px was *always* what applied — a universal **20% shrink** of
agent-pane text that silently reverted the readability default in
`SPEC_PANE_ICON_AND_TEXT_VISIBILITY_2026_05_30.md`.

ReAgent's framing is worth preserving: unlike the family fallbacks, which
genuinely **disagreed** and so had to be reconciled, both size call sites already
**agreed** on 15px. There was no inconsistency to fix — only a name to give it.
Consolidation is not license to change an agreed value.

## Verification — against the built CSS, not inferred

```
dist/frontend/assets/style-*.css:
  --font-mono:"Hack", Consolas, Menlo, monospace
  --font-agent-size: 15px          <- readability default preserved
  --font-mono-size: 12px           <- still separate
  references to deleted tokens:            0
  font-family:var(--font-mono):           55       <- now valid
  font-family:var(--fixed-font):           0       <- invalid ones gone
```

`vite build` exits 0 (re-run after the rebase onto `main`); `tsc --noEmit` clean;
`check-ligature-fonts.sh` passes. Local vitest has 29 pre-existing failures from
a `solid-refresh`/jsdom environment issue that reproduce with these changes
stashed — CI's vitest is authoritative.

## Also: fixes a false positive in #3247's gate

`check-ligature-fonts.sh`'s comment exclusion only skipped lines *beginning* with
`//` or `/*`, so a continuation line inside a block comment tripped it — caught
when the gate failed on this PR's own documentation warning people not to add
ligature fonts. Now blanks comments out (preserving line numbers). Verified both
directions.

## Not in this PR

Wiring `term:fontfamily` / `term:fontsize` to override `--font-mono` at runtime,
and a VS Code–style `editor.fontLigatures` opt-in. Both are finally *possible*
now that a single token exists, but they are behavior changes, not consolidation.


a5af:
@codex review

PR #3256, opened 2026-09-16
fix(term): close the blockinput agent-lock gap that corrupted a pane under concurrent PtyShell/human input

GenericAgentX-asaf:
## Summary

- `controllerinput` (Stop/SIGINT, resize) already checked `agent_lock::is_locked` before forwarding input to a block's PTY controller, but `blockinput` — the WS command real human keystrokes actually travel on (`termViewModel.ts`, used by both the standalone Terminal widget and an agent pane's composer-drawer shell) — had no server-side lease check at all. This was a documented, known gap in `agent_lock`'s own module doc.
- The only thing preventing a human keystroke and an agent's `PtyShellInput` write from landing in the same PTY simultaneously was the frontend's own eventually-consistent `agentLocked()` gate (reacts to a WS-pushed meta value, not synchronous with backend state).
- **Reproduced live:** an agent driving `vim` through `PtyShellInput` on its own composer-drawer shell had its writes interleave with other input reaching that same pane via `blockinput`, corrupting vim's cursor-addressed alternate-screen redraw into a visibly garbled pane — not a hang, byte interleaving mid-escape-sequence.
- Extracts the `blockinput` dispatch into `dispatch_blockinput()` and adds the same `is_locked` check `controllerinput` already had, closing the gap.

## Test plan

- [x] Added `blockinput_is_dropped_while_an_agent_lock_is_active` — written first against the old code, confirmed it **failed** (input reached `send_input` while locked), then applied the fix and confirmed it **passes**.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv server::websocket::` — 10 passed, 0 failed.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv backend::blockcontroller::` — 332 passed, 0 failed.
- [x] Verified without touching a live pane's own PTY (no `PtyShellInput`/`PtyShell` against this agent's own composer-drawer shell) — pure `cargo test`, to avoid reproducing the exact corruption this PR fixes.


a5af:
@codex review

PR #3257, opened 2026-09-16
fix(agent-shell): re-attach the composer-drawer terminal when its sub-block is respawned

GenericAgentX-asaf:
## Summary

- `AgentShellSubblock` reads `props.existingSubBlockId` exactly once, inside `onMount`'s one-shot IIFE, and binds `TermWrap` to that id for the component's entire lifetime.
- When a shell's process exits (e.g. typing `exit`) and the backend's "attach or create" logic spawns a replacement sub-block (`SPEC_TERM_EXIT_RESPAWN_LOOP_2026_09_15.md` §11's respawn path — deliberate for parented/composer-drawer shells, since the generic close-on-exit saga can't safely close them), the parent agent block's `term:shellsubblockid` meta gets repointed at the new id — but the already-mounted drawer never notices. It stays bound to the old, dead sub-block forever, rendering its last frame with no live process behind it.
- From the human's side, nothing on screen ever changes even though the backend already moved on — this is exactly the "typed `exit`, it hung" / "still looks offset" symptom.
- **Found live 2026-09-16** investigating a corrupted vim pane: confirmed via `PtyShellStatus` that the old shell had exited and a new one was spawned and healthy server-side, but the drawer stayed frozen on the old shell's last frame even after a tab switch (ruling out "just needs a visibility/focus event").

## Fix

Extracts the mount-time attach logic into a reusable `attachShell()` function and adds a `createEffect` that watches `existingSubBlockId` for an externally-driven change — distinct from the prop simply echoing back an id this component just created itself (which must NOT trigger a redundant re-attach) — tearing down the stale `TermWrap`/`ResizeObserver` and re-attaching to the new sub-block.

## Test plan

- [x] Added two regression tests: one proving the drawer disposes the old `TermWrap` and attaches a fresh one when `existingSubBlockId` changes externally (written first, confirmed failing against the old code with a clean assertion failure — `expected 2, got 1` — not a crash); one proving a self-created id echoing back through props does NOT trigger a spurious re-attach.
- [x] Full `AgentShellSubblock.test.tsx` suite: 11/11 passing.
- [x] Full `frontend/app/view/agent/` suite: 136 files / 1818 tests passing, no regressions.


a5af:
@codex review

PR #3259, opened 2026-09-16
fix(macos): task package:macos now bakes a per-build local channel

AgentO-asaf:
## Summary

`task package:macos` never baked a per-build channel at all — it always compiled against the `stable` default, via a Taskfile `deps:` ordering gap that ran the cargo builds *before* the channel could be set. Same class of bug `SPEC_LINUX_APPIMAGE_PER_BUILD_CHANNEL_2026_06_25.md` fixed for Linux weeks earlier, which explicitly flagged macOS as the unfixed platform. This closes that gap, mirroring `scripts/package.sh` (Windows) and `scripts/package-linux.sh` almost verbatim.

Motivating reproduction (see the spec, §0): a local macOS DMG build prompted macOS's "Allow AgentMux to find devices on local networks?" dialog on launch despite the LAN-discovery toggle appearing off in the UI. The gating code was correct — `settings.json` isolation (`SPEC_SETTINGS_ISOLATED_BY_CHANNEL_2026_08_19.md`) was already shipped and already correct — but `task package:macos` never baked a non-`stable` channel into the binary, so there was nothing for that isolation to isolate. Same root-cause shape independently surfaced again this week in a different incident (`docs/retro/retro-macos-0560-stale-cef-cache-launch-crash-2026-09-16.md` / PR #3246) and the MuxBus Keychain-prompt fix (PR #3248) — both of those fixes only actually take effect on a non-`stable` channel, which `task package:macos` couldn't produce until now.

## Change

- **`scripts/package-macos.sh`**: now a self-contained orchestrator that runs the full build pipeline itself (`build:frontend`, `build:backend`, `build:host`, `copy:schema`, `bundle`) rather than relying on Taskfile `deps:`, so it can compute and export `AGENTMUX_BUILD_CHANNEL_DEFAULT` *before* any cargo invocation. Bakes `local-<branch>-<hash>-<build-id>` by default; `RELEASE_CHANNEL=stable` overrides for release artifacts. DMG filename stays version-only (unlike the Windows ZIP / Linux AppImage) — it's the baked channel that makes each build its own instance, not the filename, and this preserves the existing Desktop keep-latest-by-version cleanup workflow.
- **`Taskfile.yml`**: `package:macos` no longer declares `deps:` (the script now owns the full pipeline); adds `package:release:macos` (bakes `stable`, guarded by the existing `scripts/guard-stable-build.sh` — refuses unless the tree is clean and HEAD is a merged release commit).
- **`.github/workflows/build-macos.yml`**: CI's release build now explicitly passes `RELEASE_CHANNEL=stable` to the script directly (matching `build-linux.yml`'s existing pattern for the same reason) — without this every official release DMG would ship isolated from the user's real data dir.
- **`CLAUDE.md`**: documents `task package:macos` in the isolation table alongside `task package`/`task package:linux`, and extends the data-isolation paragraph with the macOS-specific history.
- New: `docs/specs/SPEC_MACOS_DMG_PER_BUILD_CHANNEL_2026_08_24.md` — full spec, motivation, and design.

## Test plan

- [x] `bash -n scripts/package-macos.sh` — clean
- [x] `scripts/check-doc-status.sh` / `scripts/check-spec-citations.sh` — both pass
- [x] **Real end-to-end build + launch test**, not just a dry run: ran `task package:macos` for real. Build banner confirmed a genuine random channel (`local-agento-macos-dmg-per-build--378253-43402442`, bundle id `ai.agentmux.local-agento-macos-dmg-per-build--378253-43402442.0.56.1`), signed + notarized + stapled cleanly.
- [x] Mounted the resulting DMG and ran the real launcher with **no env overrides** (the exact conditions a normal Finder launch would see). Confirmed via the live srv log, all three isolation mechanisms fired correctly on the same run: `shared store: attached (ISOLATED — channel-scoped)`, `settings.json isolation... channel default — isolated`, and `cloud_subscriber: init_global skipped (ISOLATED...)`. `agentmux-cef` came up clean, no crash loop.
- [x] Changeset added (`patch`)


a5af:
@codex review

PR #3260, opened 2026-09-16
fix(editor): stop discarding the first character typed into a file

AgentY-asaf:
## Summary

Click into an editor pane and type — the character never appears, but the tab's
dirty marker does. Click again, and typing works normally.

**This was silent data loss**, not a rendering glitch. The character was
*destroyed*, so typing one character and hitting save saved the file without it.
Severity on the spec is raised from Medium to High accordingly.

Implements the fix for #3250's spec; C1 confirmed, C2 and C3 ruled out.

## Root cause

`loadingAtom` was a bare arrow function rather than a memo, so every caller that
read it subscribed to the whole tab **object** instead of the boolean:

```ts
this.loadingAtom = () => {
    const tab = this.activeTabAtom();
    return tab != null && !tab.contentLoaded && tab.loadError == null;
};
```

`editor-view.tsx`'s *"rebuild CodeMirror on tab change"* effect reads it — so any
mutation of the active tab re-ran that effect, even when `loading` was `false`
before and after. The first edit does exactly that:

1. keystroke → `onContentChange()`
2. → dispatches `MarkDirty`, guarded by `if (!this.dirtyAtom())` — **which is why
   only the first keystroke was affected**; later edits dispatch nothing
3. → tab object changes → un-memoized `loadingAtom` propagates
4. → effect re-runs → CodeMirror rebuilt from the **on-disk** content
5. → the typed character is destroyed with the old view

## Evidence — live trace, not inference

Instrumented the running CEF 152 build and reproduced:

```
15:12:59.776  setupEditor rebuild {gen:1, docLen:0}     <- file opened, view gkf9i
15:13:01.642  docChanged {viewId:"gkf9i", docLen:1}     <- FIRST KEYSTROKE lands
15:13:01.643  setupEditor rebuild {gen:2, docLen:0}     <- 1ms later, rebuilt from disk
15:13:03.482  docChanged {viewId:"gz0b4", docLen:1}     <- 2nd click, new view, works
```

## Also fixed: a separate `setupEditor` race

`setupEditor` is async and awaits `loadLanguage()` **between** destroying the old
view and constructing the new one. Two calls starting before either constructs
both find `cmView === null`, both skip the destroy, and both append an
`EditorView` to the same container — orphaning one: alive, in the DOM, able to
take focus, referenced by nothing, never destroyed. It fired on **every open**,
~2ms apart. Closed with a generation guard.

The trace shows a single view id throughout, so this was **not** the
first-keystroke bug — but it is a real defect worth closing.

## Two discarded hypotheses (recorded in the spec)

- **C2, stale measurement** — a `requestMeasure()` fix was written and tested on
  the premise that the container was zero-sized at construction. The trace shows
  `containerH: 1053`. Premise false, fix discarded rather than shipped as cargo
  cult.
- **The double-view race** — real, fixed above, but not this bug.

Both are documented so the next person doesn't retry them.

## Testing

Verified live by the repo owner on CEF 152: first keystroke now renders.
`tsc --noEmit` clean.

**Regression test deferred, deliberately.** There is no `EditorViewModel` test
harness — the model needs block atoms, the pane-state slice and RPC — and a test
that doesn't actually exercise the reactive path would be worse than none. Noted
as a follow-up rather than padding this PR with something unverifiable.


a5af:
@codex review

PR #3261, opened 2026-09-16
fix(term): flush barrier — drop the 1s exit delay without weakening the output-ordering guarantee

Agent1-asaf:
## Summary

Typing `exit` in the agent pane's shell drawer collapsed it ~1s later. Not UI lag — the wait/cleanup task waits for the PTY output flusher before publishing `STATUS_DONE`, and the drawer's collapse (#3253) is driven by that event.

**This PR changed approach mid-review.** The first version special-cased the wait away for cleanly-exited sub-blocks. Codex's P1 showed that's incorrect, so it's been replaced with a real flush barrier — which fixes the latency for *everyone* instead of carving out one case.

## Why the first approach was wrong (Codex P1)

It assumed a clean sub-block exit means no reader is left. False: `/api/v1/ptyshell/create` produces the same `block:`-parented shell and works with **no drawer mounted at all**. Agents poll `PtyShellStatus` until `running: false`, *then* read output via `PtyShellRead` — so letting `STATUS_DONE` outrun the final flush hands them **truncated command output**.

And it can't be narrowed: since #3194 the drawer's shell and the PtyShell API shell are the *same block*. The distinction I keyed on doesn't exist.

## The actual problem

Waiting on the flusher **task** was always a proxy for the real question. The task only ends when the channel closes — and on Windows that never happens: ConPTY's console host outlives the child, so the read loop stays blocked in a read that never returns and never drops its sender. Every ordinary exit burned the full `FLUSHER_DRAIN_TIMEOUT` while the flusher sat idle.

## The fix

`PtyReadChunk` gains an optional `ack`; `flush_barrier()` enqueues an empty chunk carrying one. Riding the same queue, it can't overtake the data ahead of it, and the flusher fires the ack **after** the batch commits. That asks the question actually being asked — *"is prior output committed?"* — and is answered in about one `PTY_COALESCE_WINDOW` instead of a second.

Strictly better than both the status quo and the rule it replaces:
- Every exit gets the fast path — top-level panes, the drawer, **and** API-driven shells
- The ordering guarantee is *preserved for all of them* rather than traded away for one
- The timeout stays as the fallback ceiling for a genuinely wedged flusher
- `close_on_exit`'s existing §13 skip is untouched

## Test plan

Three tests: the commit-ordering guarantee, a gone flusher resolving rather than hanging, and an unanswered barrier expiring rather than blocking exit forever.

**The first ordering test was worthless and falsification is how I found out.** It drained the channel by hand — proving the channel is FIFO and nothing about the flusher — and passed unchanged with the ack moved to *before* the flush, the exact bug it existed to catch. Rewritten against the real flusher with a real in-memory FileStore, reading the file at the moment the barrier resolves. That version does fail when the ack is moved early. (Same harness-gap shape as the two P0s on #3253.)

Latency was **verified live** by the repo owner on the previous approach (`exit` collapsed instantly); the barrier delivers the same speed by a sounder route, but has **not** itself been re-verified in a running build.

## Note

`backend::sysinfo::...refresh_processes_specifics_does_not_leak_a_handle_per_call` fails on my machine right now — but it fails identically on clean `main` (handle count grew 635 over 500 calls), and nothing here touches sysinfo. Pre-existing and unrelated; flagging rather than folding it in.


a5af:
@codex review

PR #3262, opened 2026-09-16
feat(agent-pane): My Agents row menu + make Delete actually remove the row

Agent2-asaf:
## Summary

Adds the per-row actions menu to the picker's **My Agents** list — Delete,
Duplicate, Rename, View History (`docs/specs/SPEC_AGENT_DELETE_2026_09_16.md`)
— and fixes the bug found while testing it: *"delete isn't completely working —
the icon on the card disappears, but the card stays."*

## The bug

The picker's rows don't come from `db_agents`. `listrecentsessions` builds them
from the host-global **instance registry** (`~/.agentmux/shared/agents/registry/`),
then resolves each row's display fields from `agent_def_list()`. When an
instance record outlives its definition, the row still renders — `provider: ""`
so `DualProviderLogo` draws nothing (the icon "disappears") and
`definition_name: "(missing definition)"` — while `instance_name` keeps the
card's title populated. Hence: icon gone, card stays.

Two independent causes, both confirmed against a real machine's on-disk state:

1. **Wrong key.** `agent_def_delete` called `reg.hard_delete(id)` with a
   *definition* id, but `Registry` is keyed by *instance* id
   (`<instance_id>.json`). Those coincide only for records
   `m0026_registry_agent_id_rekey` re-keyed — and that pass reads
   `db_agent_instances`, dropped in schema v32, so it no-ops. An install that
   consolidated before it could run still holds one launch-keyed file per agent
   (**30 of 44 records** on the reporting machine; 14 agents had a duplicate
   pair, one of which always survived the delete).
2. **Sweep skipped for cross-channel agents.** It sat inside `if rows > 0`. An
   agent seen only through the global overlay deletes with `rows == 0` — which
   is most rows a fresh channel shows (the reporting dev channel listed ~20 rows
   against **2** local `db_agents` rows).

## Fixes

- `Registry::hard_delete_for_agent` / `retire_for_agent` / `unretire_for_agent`
  match a record by file key **or** its own `definition_id`.
- Registry sweep gated on `rows > 0 || global_retired`.
- `instance_set_hidden` ("Forget agent") had cause 1 as well — literally the
  *"the forgotten agent reappears"* failure `m0026`'s own doc comment predicted.
- `instance_delete` — the same deletion under a second name — had **neither**
  the dependent-table purge nor the registry sweep. Both now come from shared
  helpers.
- `backend::registry_reconcile` runs a startup pass (beside the existing
  `backfill_session_ids`) for records already orphaned on disk, which none of
  the above can reach. Deliberately narrow: it requires a `retired/` tombstone,
  since "absent from the active definition tree" alone is not evidence of
  deletion — a definition may live only in some channel's local SQLite.

## Two further problems found on the way through

- **The dependent purge was aimed at the wrong database.**
  `db_agent_identity_links`, `db_agent_credentials` and both native-memory
  tables exist in *both* schemas, and per
  `SPEC_IDENTITY_STORE_SPLIT_2026_08_17.md` the **identity store** holds the
  live rows — `listrecentsessions` itself reads links from
  `identity_store.agent_identity_list_all()`. The purge only ever hit the object
  store's legacy copies, so the confirm dialog's *"deletes the agent and its
  credentials"* was false. `purge_agent_dependents` now skips tables absent from
  whichever connection it runs on (one list serves both schemas), and the
  `deleteagent` handler — the only layer holding both stores — purges the
  identity store too.
- **The spec's §5.1 premise was wrong.** `Store::configure_and_migrate` *does*
  set `PRAGMA foreign_keys=ON` on the production connection, so those six
  cascades are not inert. The explicit purge is still correct for the no-FK
  tables. Both corrections are recorded in the spec rather than silently
  rewritten. (`managed.rs`'s "only ever set in tests" comment is stale for this
  store; left untouched — it describes a different, genuinely cross-database
  case — but worth a follow-up.)

## Cleanup / DRY

Backend: one `purge_agent_dependents` table list instead of two open-coded
blocks; one `purge_agent_side_effects` shared by both delete paths.

Frontend (`MyAgentsList.tsx`): the duplicated fork-naming markup extracted into
a shared `NamePrompt` (Rename was a near-verbatim copy — autofocus effect and
key handling included); the exclusive row menu's `Set<string>` collapsed to a
single `openMenuId` signal; and the click-outside/Escape close the spec called
for but the code lacked — with the spotlight blur setting `pointer-events: none`,
the chevron was previously the only way back out.

## Testing

- `cargo test -p agentmux-srv`: **3494 passed, 0 failed.** New coverage for both
  delete gaps, the forget round-trip, the reconcile pass, and the identity-store
  purge.
- `tsc --noEmit`: clean. No new warnings in any touched file.
- **Not run:** the Solid component tests (`MyAgentsList.test.tsx`,
  `AgentPicker.test.tsx`). They fail in this environment with
  `ERR_INVALID_ARG_VALUE: file:///@solid-refresh` — pre-existing and unrelated;
  untouched `AgentCard.test.tsx` fails identically. The frontend changes need a
  real-app check via `task dev` (in progress).


a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

a5af:
@codex review

PR #3255, opened 2026-09-16
Vmer@asaf: feat(shell): add muxsh — open editor/browser panes from the terminal

a5af:
## Summary
- Adds `muxsh open <file>` / `muxsh web <url>` — open editor/browser panes from a terminal, no GUI required. AgentMux's own successor to Wave Terminal's `wsh` CLI (which AgentMux inherited and fully retired in 2026-04, `docs/specs/archive/SPEC_RETIRE_WSH_2026_04_12.md`), scoped narrowly to the one gap that's real today rather than reviving `wsh`'s full surface.
- Both subcommands are thin wrappers over the already-generic `POST /api/v1/pane/open` route — confirmed it already accepts `view=browser`+`url`, not just the editor-only subset the `OpenEditor` MCP tool exposes, so **no backend/API change was needed**, only a new terminal client.
- Same architecture and auth model as the existing `muxopen` (added 2026-09-06): Node core deployed to `~/.agentmux/shell/muxsh.mjs`, shell-function delegators added to all four shells (bash/zsh/fish/pwsh), `$AGENTMUX_LOCAL_URL` + `$AGENTMUX_AUTH_KEY` inherited from the pane environment — no new IPC, no new auth scheme.
- **Drive-by fix:** `shellintegration.rs`'s `version_marker()` hashed every embedded script except `MUXOPEN_JS`, so an edit to `muxopen.mjs` alone never invalidated the deployment marker and a running instance would never redeploy it. Fixed in the same lines while adding `MUXSH_JS` to the hash.
- New minor changeset, `docs/MUXSH.md` reference doc, `docs/specs/INDEX.md` regenerated.

## Design docs
- `docs/reports/REPORT_WSH_STYLE_CLI_FOR_AGENT_APP_API_2026_09_16.md` — the research: what `wsh` actually is (cloned `wavetermdev/waveterm` to verify from source), why AgentMux's 2026-04 retirement still holds, what's changed since (a real Agent App API + the `muxlog`/`muxspect`/`muxopen` precedent), and collision-checked naming research (`muxsh` vs `muxctl` vs `amux` — `amux` collides with several real, actively-maintained AI-agent-orchestration tools in the exact same space and was rejected).
- `docs/specs/SPEC_MUXSH_CLI_2026_09_16.md` — the implementation design this PR follows.

## Verification
- `cargo check -p agentmux-srv` clean.
[...]

PR #3263, opened 2026-09-16
Vmer@asaf: feat(shell): DRY foundations for the muxsh collection

a5af:
## Summary

Adds `docs/specs/SPEC_MUXSH_FULL_COLLECTION_2026_09_16.md` — the full plan for recreating the `wsh` command vocabulary as a consistent, noun-verb `muxsh` collection (house style, an honest per-`wsh`-command mapping onto real Agent App API routes, phased build list) — and implements its **Phase 2a prerequisite** before any new subcommand lands, per the spec's own §2.8.

**Why this comes first:** every route already had two independent Rust implementations (the REST handler, the MCP tool binding) before `muxsh` added a third by hand in Node (PR #3255). Scaling to more subcommands without addressing this compounds the exact bug class #3255 already hit once (a field silently not sent).

**Two tiers, per the spec:**

1. **`lib/muxclient.mjs`** — shared env-read + authenticated-fetch plumbing, now imported by `muxspect`/`muxopen`/`muxsh` instead of each hand-rolling its own ~20 lines of it (`muxlog` is unaffected — it makes no REST calls at all). Deliberately does **not** read response bodies — each tool's error-handling strategy (JSON-always-with-`{}`-fallback vs. JSON-on-success/text-on-failure) legitimately differs, and forcing that to converge risked silently changing behavior. All 131 pre-existing shellintegration tests pass unchanged, confirming this is a behavior-preserving refactor.
2. **`docs/specs/app-api-manifest.json`** — a checked-in source of truth for route field names, checked from *both* sides: a new Rust test module (`rpc_types/block.rs`'s `app_api_manifest_contract_tests`, using a new `Serialize` derive on `CommandPaneOpenData` added specifically for this) asserts the manifest matches the real struct's fields; a new Node test (`muxsh.contract.test.mjs`) asserts `muxsh` never emits a field the manifest doesn't list. A future field rename now fails CI on whichever side drifted, instead of shipping as a silent mismatch — which is exactly how #3255's real bug was actually found (after merge, by ReAgent, not before).

Covers only `pane.open` for now — proves the mechanism before Phase 2b extends it to each new subcommand's own PR (`view`/`edit`, `pane close/list`, `run`, `config edit/path`, `agent list/send`, `secret list` — see the spec §4).

## Verification
- `npx vitest run agentmux-srv/src/backend/shellintegration/` — 135/135 passing (131 pre-existing unchanged + 4 new contract-test cases).
- `cargo test -p agentmux-srv app_api_manifest_contract_tests` — 2/2 passing.
- `bash scripts/check-doc-status.sh` / `gen-docs-index.sh --check` clean.

## Test plan
- [ ] CI green
- [ ] Manual: `task dev`, confirm `muxsh open`/`muxsh web`/`muxopen` still work identically post-refactor (no behavior change intended)


a5af:
@codex review

Issue #3258, opened 2026-09-16
tracking: startup + splash — get migrations and updates off the boot path


AgentX-asaf:
## Why this exists

Umbrella tracker for making app startup fast and honest: the splash screen
itself, and everything the boot path does before the app is usable.

The immediate trigger: a slow migration batch on the first boot after an
upgrade currently blocks the entire app behind the splash â€” the launcher
silently widens its `AGENTMUXSRV-ESTART` deadline from 30s to **30 minutes**
while `agentmux-srv` applies migrations in-process, before it has even bound
its TCP listeners.

**Goal (repo owner):** no migrations and no update work at startup. Anything
upgrade-shaped runs through an Upgrade button in the status bar, which
migrates and restarts. Ordinary launches do only what that launch needs.

## Docs

| Doc | What it covers |
|---|---|
| `docs/specs/SPEC_FAST_STARTUP_UPGRADE_OWNS_MIGRATIONS_AND_UPDATES_2026_09_15.md` | The consolidated design â€” boot inventory, target Upgrade flow, update-check constraints, startup budget, invariants, phases |
| `docs/reports/REPORT_STARTUP_MIGRATION_AUDIT_AND_DEFERRAL_2026_09_15.md` | Why migrations sit on the boot path, the two prior attempts and why each was reverted |
| `docs/reports/REPORT_SPLASH_MIGRATION_ROW_OVERFLOW_AND_SUMMARY_2026_09_15.md` | Per-migration splash rows: too granular, and overflow drops the newest instead of scrolling to it |
| `docs/reports/REPORT_SPLASH_SCREEN_ARCHITECTURE_RETHINK_2026_09_14.md` | The earlier splash DRY/timing audit (Phase 1 shipped as #3222; phases 3-4 still open) |

## Hard constraints (both learned the hard way)

1. **The zero-pending boot must not gain a single new process spawn, file
   open, or network call.** `1052c985b` (repo owner, 2026-06-27) reverted an
   earlier deferral because its subprocess-per-boot design taxed every boot,
   not just the rare one with migrations pending.
2. **Migrations only ever run with no live srv holding that data dir.**
   `341faa981` (2026-06-26) had to remove an on-demand migrate path because
   running it against a live srv caused permanent data loss â€”
   `m0011_shared_store_backfill`'s skip-if-already-non-empty guards saw
   tables the running instance had touched and skipped the backfill for good.
   Enforce with a real process boundary and a wait-for-exit, not a sleep.

[...]

PR #3264, opened 2026-09-16
docs(startup): audit migrations + updates on the boot path, spec the Upgrade-owned model

AgentX-asaf:
## Summary

Docs only, no code. Tracked by #3258.

Startup currently blocks on DB migrations: `agentmux-srv` runs them in-process *before* it binds its TCP listeners (`bootstrap.rs:573`, ahead of `main.rs` step 5), so nothing serves until the batch finishes — and the launcher quietly widens its `AGENTMUXSRV-ESTART` deadline from 30s to **30 minutes** when it sees `AGENTMUXSRV-MIGRATING`. With 34 registered migrations, a machine that's behind can run most of them in one silent, splash-locked boot.

## What's here

| Doc | Content |
|---|---|
| `docs/reports/REPORT_SPLASH_MIGRATION_ROW_OVERFLOW_AND_SUMMARY_2026_09_15.md` | Per-migration splash rows (#3223) are one row each with no summarization, and all three renderers `break` at their row cap oldest-first (`splash.rs:515,545`, `splash_mac.rs:154,180`, `splash_linux/mod.rs:167,177`) — so a long batch fills the panel with the *earliest* migrations and never shows the running one, and can drop the trailing `total`/`other` rows entirely |
| `docs/reports/REPORT_STARTUP_MIGRATION_AUDIT_AND_DEFERRAL_2026_09_15.md` | Full boot-path audit + the prior history |
| `docs/specs/SPEC_FAST_STARTUP_UPGRADE_OWNS_MIGRATIONS_AND_UPDATES_2026_09_15.md` | The consolidated design, merging both reports with `SPEC_UPGRADE_PANEL_2026_06_27.md` and `app-update-check.md` |

## The two constraints the design is built around

Both come from prior attempts at exactly this, one day apart in June:

1. **No new per-boot cost on the zero-pending path.** `1052c985b` (repo owner) reverted an earlier deferral because its subprocess-per-boot design taxed every boot, not just the rare one with migrations pending.
2. **Migrations only run with no live srv on the data dir.** `341faa981` removed an on-demand migrate path because running it against a live srv caused permanent data loss — `m0011_shared_store_backfill`'s skip-if-already-non-empty guards saw tables the running instance had touched and skipped the backfill for good.

Constraint 2 is why "migrate, then restart" is the safety mechanism here, not a UX preference.

## Design, in short

- Zero-pending boots (virtually every boot) are unchanged — `count_pending_migrations()` already fast-paths.
- A pending batch is **not** applied at boot. The app renders a full-window blocking upgrade screen; clicking it quiesces srv, snapshots, runs the existing `agentmux-srv migrate` subcommand, and restarts. The next boot sees 0 pending.
- The button is **not** on the splash — that's a non-interactive native window on all three platforms (`splash.rs:237`). The splash instead dismisses *faster* than today, since `ESTART` no longer waits on migrations.
- Update checks, install-type detection, and downloads all stay off the boot path (§5). None of this is implemented today — `install_update` is still a stub and nothing emits `app-update-status`.
- Also inventories four **untimed** per-boot passes (registry `session_id` backfill, transcript backfill, gap-repair heal, agent auto-seed) and proposes measuring them before moving them, plus a `<300ms` warm-boot budget and five reviewer invariants.

## Notes for review

- `SPEC_UPGRADE_PANEL_2026_06_27.md`'s header still says "not yet implemented" though its components shipped — flagged in the spec's open items, deliberately not fixed here to keep this PR docs-additive.
- The full-window-blocking-screen decision is recorded on #3258; alternatives considered (advisory status-bar button, modal over a dimmed workspace) and why they were rejected are in §4.2.


a5af:
@codex review

AgentX-asaf:
Thanks — all six findings were real; I verified each against the code before acting on it, and all six are now fixed in the spec.

| Finding | Verified how | Fix |
|---|---|---|
| P1 pending detection must fail closed | `runner.rs:505-535`'s own doc comment says "Returns 0 on any error so startup is never blocked" | §4.1 now rejects `count_pending_migrations` as the gate outright; the check must be fallible and route errors to the blocking state. New invariant **S1b**: "couldn't tell" is not "nothing pending" |
| P1 gate the backend, not the frontend | `open_stores_and_migrate` → `Store::open` → attachment → repair/backfill/seed, then `main.rs:134` saga resume, all pre-`ESTART` | §4.2 now specifies an **upgrade-only srv mode** that skips store init, the passes, saga resume and background subsystems; the frontend gate is explicitly demoted to presentation/defense-in-depth. New invariant **S1c** |
| P1 suspend supervision | `supervisor/windows.rs:964` respawns on any srv exit; `supervisor/unix.rs:713` reads a clean exit as teardown | §4.3 gains a maintenance state as **step 1**, before the quiesce. Called out that without it the sequence re-creates the `341faa981` hazard it exists to prevent. Folded into **S2** and into Phase 1's scope |
| P1 snapshot ordering | `count_pending_migrations` opens read-write and may create `objects.db`; `bootstrap.rs:529-534` documents why the snapshot must precede any open | §4.3 step 3 + §4.1 point 2: either the check becomes non-mutating or the snapshot stays strictly ahead of it. The snapshot must be the first thing to touch those files |
| P1 install staged binary before migrating | Ordering error in my own §4.3/§5.4 | Swapped: install is now step 4, `migrate` step 5, with the reasoning inline in both places. New invariant **S2b** |
| P2 partial application | `apply_pending` commits and marks applied per migration; no batch transaction | Dropped the false "data untouched" claim. State H must report which applied, which failed, and where the backup is; the backup is a recovery source, not a rollback |

The doc-status CI failure is fixed too (Status lines on both reports, vocabulary-conforming Status on the spec, `INDEX.md` regenerated). All three doc gates pass locally.

PR #3265, opened 2026-09-16
docs(panes): spec routing typing into a pane the moment it opens

AgentY-asaf:
## Summary

Spec only, no implementation. Requested by the repo owner: open an editor or
terminal pane and **just type** — no mouse trip to click into the text area
first. The pane is already selected on creation; only the caret is missing.

## The finding: it's one shared contract, not a per-view bug

Every focus hand-off in the app runs the same three lines, duplicated across
`focusManager.refocusNode()`, `block-component-registry` and
`openOrFocusPaneByView()`:

```ts
const ok = bcm?.viewModel?.giveFocus?.();
if (!ok) document.getElementById(`${blockId}-dummy-focus`)?.focus();
```

Two properties of that contract produce the symptom:

- **Single synchronous attempt** — no retry, no notion of "the view isn't
  mounted yet".
- **Silent failure** — when `giveFocus()` returns false, focus lands on a hidden
  dummy input. The pane *looks* focused and the app is perfectly consistent,
  while keystrokes go nowhere.

## Actual state of `giveFocus()` across pane types

| View | `giveFocus()` | Status |
|---|---|---|
| Editor (`editor-model.ts:1210`) | `return false` | **stub** |
| Agent (`agent-model.ts:930`) | `return false` | **stub — identical** |
| Terminal (`termViewModel.ts:504`) | focuses xterm when it exists | implemented, **races mount** |
| Browser (`browser-model.ts:711`) | real implementation, with guards | works |

**Three of four pane types do not reliably take the caret, and two are outright
stubs.** Only the browser genuinely implements the contract.

> **Correction:** the first revision of this spec — and of this description —
> claimed the agent pane implemented `giveFocus()` and "works today", and leaned
> on that to argue the shared contract was sound. That was wrong (ReAgent P1,
> then P0 when the description lagged the corrected file). It changes the
> diagnosis rather than just a detail, which is why it is called out here rather
> than quietly edited.

That makes the shared-layer fix the clear choice rather than a judgement call:
patching per-view would mean writing the same readiness logic three times, and
the next view added would default to the same silent failure.

## What the spec proposes

Three options, recommending **making the shared contract retry until the view is
ready** (tri-state return or `focusWhenReady()`), with the editor's `giveFocus()`
implemented first since it's required under every option and is independently
correct.

It also records the constraints that make auto-focus actively hostile if
over-applied — never steal focus from a pane being typed in, from a modal or an
open find panel, or across a five-pane layout restore — plus acceptance criteria
and five open questions.

## Explicitly not the first-keystroke bug

§6 separates this from `SPEC_EDITOR_FIRST_KEYSTROKE_NOT_RENDERED_2026_09_15`
(fixed in #3260), which has a deceptively similar symptom. They're independent:
that one was **data loss**; this one loses nothing (keystrokes go to a dummy
input before any document exists). Worth noting that fixing *this* one first
would have **masked** that one — auto-focus would have hidden a silently dropped
first keystroke.


a5af:
@codex review

AgentY-asaf:
Fixed — and the P0 was pointing at something real that I'd half-corrected.

The **spec file** was corrected in `09261b651` (the commit this review ran against): §2c and its table now read `agent-model.ts:930` → `return false` → **stub — identical** to the editor's, and §4 names the agent pane as affected.

What I missed is that the **PR description** still carried the original wrong table ("Agent / Browser | implemented | Work today"). That's the first thing any reviewer reads, so the incorrect claim was still live even though the diff had fixed it. Description now rewritten to match the file, with the correction called out explicitly rather than quietly edited.

For the record, the corrected picture is worse than the original spec claimed, and that strengthens its recommendation: three of four pane types don't reliably take the caret, two are outright stubs, and only the browser implements the contract properly. Per-view patching would mean writing the same readiness logic three times.

PR #3266, opened 2026-09-16
fix(statusbar): stop metrics silently rendering the wrong font; add CEF version

AgentA-asaf:
## Summary

- `.stat-mono` (the status bar's CPU/GPU/RAM/disk/net/uptime readouts) had `font-family: var(--fixed-font)` — `--fixed-font` is a `font` *shorthand* string, invalid as a `font-family` value, so the declaration was silently dropped and these readouts rendered in the ordinary UI sans font (Inter) the entire time. A recent upstream font-variable consolidation (#3254, following #3247's JetBrains Mono removal) fixed the declaration to a valid `font-family: var(--font-mono)`, which then made the readouts visibly switch to a real monospace face for what may be the first time ever. Root-caused in `docs/retro/retro-statusbar-metrics-font-change-2026-09-16.md`.
- Decision: keep the pre-existing look on purpose now that it's a real, working declaration — `font-family: var(--font-sans)`, plus tighter `letter-spacing` per follow-up feedback.
- Instance panel: add a **CEF** row showing the linked `cef` crate's version (embedded at build time in `agentmux-cef/build.rs` by reading the resolved version straight out of the workspace `Cargo.lock`), and a timezone suffix on the **Time** row.
- Backend status popover: drop its **Version** row — it always duplicated the app version already shown in the Instance panel's version chip (same workspace release, always identical).

## Test plan

- [x] `task dev` — visually confirmed status bar metrics render in the sans font with tightened spacing (screenshot-verified)
- [x] `task dev` — visually confirmed CEF host rebuild compiles clean with the new `AGENTMUX_CEF_VERSION` build-time embed
- [ ] CI


a5af:
@codex review

PR #3267, opened 2026-09-16
feat(srv): time the boot path and log one summary line at ESTART

AgentX-asaf:
## Summary

Phase 4 of #3258 — measurement, so the rest of that work aims at real numbers.

`SPEC_FAST_STARTUP_UPGRADE_OWNS_MIGRATIONS_AND_UPDATES_2026_09_15.md` §3 inventories the boot path and finds several passes that run on **every** boot and have never been individually timed: the registry `session_id` catch-up backfill (`bootstrap.rs:606`), the definition-store listing the transcript backfill needs, the snapshot-source gap repair (`:797`), and the agent auto-seed (`:870`). The splash's `backend` stage reports one number covering all of them, so there's currently no way to say which — if any — is worth moving off the critical path. §6 asks for measurement before relocation.

## What this adds

`agentmux-srv/src/boot_timing.rs`: a `BootTimings` value type plus a process-global recorder. Wraps the four untimed passes, the saga-id-seed scan, and the four top-level phases in `main.rs`. One `info` line where the boot path ends (right after `emit_estart`), slowest step first:

```
boot_timing: total=412ms accounted=350ms | open_stores_and_migrate=210ms registry_session_backfill=88ms ...
```

Read it with `muxlog srv grep boot_timing`.

## Design notes

- **`tracing`, not the `srv_stderr` splash protocol** the per-migration rows use. The direction of travel is *fewer* splash rows, not more (`REPORT_SPLASH_MIGRATION_ROW_OVERFLOW_AND_SUMMARY_2026_09_15.md`), and these numbers are for whoever is profiling a boot, not the user watching one.
- **The lock is never held across the timed closure**, so a timed step may contain timed steps — `open_stores_and_migrate` wraps four of them — without deadlocking. Tested.
- **Uninitialized, it's a transparent pass-through** — that's every unit test in the crate and the `migrate` subcommand. A pass-through that silently skipped the work would be catastrophic, so that's tested too.
- `total` and `accounted` are reported separately and neither is derived from the other: `accounted` double-counts nested steps, and `total` includes boot time no step wraps.
- **Global state**, deliberately: the wrapped passes are scattered through `bootstrap.rs`'s long boot functions, and threading a `&mut BootTimings` through all of them would be a far larger diff than the measurement is worth. `BootTimings` itself has no global state, which is what the tests exercise.

## Behavior

No change — every wrapped call site computes exactly what it did before.

## Testing

- 7 new unit tests (value pass-through, call-order recording, real elapsed time, slowest-first summary, total-vs-accounted, uninitialized pass-through, nested-no-deadlock)
- `cargo test -p agentmux-srv`: **3455 passed, 0 failed**


a5af:
@codex review

AgentX-asaf:
**Stacked on #3264 — the doc-gate failure here is purely a sequencing artifact.**

`check-spec-citations.sh` fails with one dangling citation: `boot_timing.rs` cites `docs/specs/SPEC_FAST_STARTUP_UPGRADE_OWNS_MIGRATIONS_AND_UPDATES_2026_09_15.md`, which is the spec this instrumentation implements Phase 4 of — and it only exists in #3264, which is approved but not yet merged.

No action needed on the citation itself (it points where it should). I'll rebase this branch onto `main` once #3264 lands, which resolves it.

a5af:
@codex review

a5af:
@codex review

AgentX-asaf:
Good catch — verified and fixed in `6e55f70`.

`muxlog`'s grep tests `fields.message` only (`agentmux-srv/src/backend/shellintegration/muxlog.mjs:157`), and `boot_timing` was living in the tracing *target*, so `muxlog srv grep boot_timing` — the command this module's own doc comment tells you to run — would have filtered out the one line it exists to produce. The message is now prefixed `boot_timing: ...` (matching the commit's example output), with the target kept so `--target boot_timing` works too. Reasoning recorded at the call site so it doesn't get "tidied" back later.

PR #3268, opened 2026-09-16
fix(splash): summarize older sub-items, and keep the newest rows on overflow

AgentX-asaf:
## Summary

Part of #3258. Fixes the defect reported in `REPORT_SPLASH_MIGRATION_ROW_OVERFLOW_AND_SUMMARY_2026_09_15.md` (merging in #3264).

Per-migration sub-rows (#3223) give every migration a permanent row, and all three renderers then iterate oldest-first and stop dead at their row cap (`splash.rs:515,545`, `splash_mac.rs:154,180`, `splash_linux/mod.rs:167,177`). A long batch therefore fills the panel with migrations that have **already finished** and never shows the one actually running — and on Windows/macOS the trailing `other`/`total` rows, appended only if rows happened to be left over, get dropped too.

A progress display that hides the in-flight item and freezes on completed history is indistinguishable from a hung one, which is exactly how this was being read.

## Change

`splash_core::visible_rows` — one shared row-selection pass for all three platforms:

- **Summarize.** Older sub-items collapse into a single `4/6 done` tally; the newest two stay in detail. Two rather than one because the newest is usually still running, and the one just finished — with its real duration — is what shows the batch is moving.
- **Scroll, don't truncate.** If the list still exceeds the budget, the **newest** rows are kept and the oldest dropped — the inverse of the old `if row >= MAX_STAGE_ROWS { break }`.
- **Reserve the summary.** Windows/macOS reserve the `other`/`total` rows up front, so they survive a full panel and an old timeline row gives way instead.

Linux additionally stops showing only `subs.last()`: it gets the same tally + detail rows as the other two, so a migration batch is no longer one ever-changing line with no sense of progress.

Row *selection* is now shared; each platform still owns its own *formatting* (GDI text, `NSTextField` pairs, padded strings). That split is deliberate — `REPORT_SPLASH_SCREEN_ARCHITECTURE_RETHINK_2026_09_14.md` §1 traces every splash defect so far to the same fix having to land three times and never doing so.

## Testing

- 6 new `splash_core` tests: short list untouched, long list collapses with the right hidden-done count, **the running item always survives**, overflow keeps the newest, zero budget doesn't panic, indent shape.
- `splash_mac`'s `total_row_still_shown_when_panel_is_nearly_full` asserted the *old* append-if-room priority, so it's replaced by `a_full_panel_drops_the_oldest_row_not_the_summary` (both summary rows now survive; `rows[0]` is `s1`, not `s0`), plus a new tally test.
- `cargo test -p agentmux-launcher`: **251 passed, 0 failed**.

## Reviewer note — macOS is not compiled by PR CI

`splash_mac.rs` is `#![cfg(target_os = "macos")]` and this PR lane builds only ubuntu + windows, so the macOS edit and its two tests are **not** exercised by any check here. I verified it by inspection and parse-check only. Windows is covered locally + by CI, Linux by the ubuntu leg. A macOS eye on `flatten_rows` (and ideally a nightly build) would be worth having before this is relied on there.


a5af:
@codex review

AgentX-asaf:
Expect one CI failure here until #3264 merges: `check-spec-citations.sh` will flag `splash_core.rs`'s citation of `docs/reports/REPORT_SPLASH_MIGRATION_ROW_OVERFLOW_AND_SUMMARY_2026_09_15.md` as dangling, because that report is the one this PR implements and it lives in #3264 (approved, in CI).

Same sequencing artifact as #3267. I'll rebase onto `main` once #3264 lands, which resolves it. No change needed to the citation itself.

a5af:
@codex review

AgentX-asaf:
Both P1s fixed in `90d773d` — and both were found independently by ReAgent and Codex, which is a good sign they're real.

**Orphaned children on overflow.** `visible_rows` trimmed the flat list by count alone, so a cut landing inside a stage block dropped the stage row and kept its indented children — dangling under whatever stage happened to precede them. Exactly reachable on Linux's 4-line budget with an ordinary `prep → backend(migrations) → host` sequence, as Codex noted. Rows are now built per-stage as blocks and filled newest-block-first; a block that only partially fits keeps its stage row plus as many of its newest children as are left, never the children alone.

**Tally undercounted.** `done` was computed over the hidden slice while `total` was the whole stage, so once either always-shown detail row finished, the ratio stopped moving — systematically understating progress, which is the one thing that row exists to convey (and, as ReAgent points out, the module's own rationale guarantees a completed item is usually among the two shown). `done` now counts all of the stage's sub-items, matching the struct's doc comment.

Two new tests cover the gap that let this through — no test previously exercised multi-stage overflow *with* sub-items present:
- `overflow_never_orphans_a_sub_row_from_its_stage` — asserts the first row is a stage row and that no indented row precedes one.
- `a_partially_fitting_stage_keeps_its_header_over_its_oldest_child` — asserts which rows are given up when only part of a block fits.

Plus both tally assertions updated to the corrected ratio. `cargo test -p agentmux-launcher`: 253 passed.

PR #3269, opened 2026-09-16
Vmer@asaf: fix(widgets): Slack widget opens the get-started/create-workspace page

a5af:
## Summary
- `defwidget@slack`'s browser pane pointed at `https://app.slack.com/` — Slack's general sign-in/workspace-picker page, a dead end for a user who doesn't already have a Slack workspace.
- Changed to `https://slack.com/get-started?entry_point=home_page#/createnew` — Slack's own dedicated onboarding flow, which still offers sign-in for a returning user, so nothing is lost for the already-connected case.
- One-line config change — every messenger widget (Discord/Slack/Telegram/WhatsApp/Teams) is a static-URL CEF browser pane with no connection-state logic (`docs/status/STATUS_MESSENGER_APP_INTEGRATIONS_2026_07_07.md` §2), so this fits the existing architecture rather than needing new logic.

## Design doc
`docs/specs/SPEC_SLACK_WIDGET_GET_STARTED_URL_2026_09_16.md` — problem, fix, and explicitly why no dynamic "already connected" logic was added (would be new architecture for all 5 messenger widgets, not something this specific report asked for).

## Verification
- `jq . agentmux-srv/src/config/widgets.json` confirms the file is still valid JSON.
- Not verified: opening the widget in a live `task dev` instance to see the real page load — no GUI available in the environment this was prepared in. The change is a single string-literal URL in a static config file with no other logic touching it, so this is the practical extent of pre-merge verification available here.

## Test plan
- [ ] Open AgentMux, open the Slack widget (hamburger → More → Messengers → Slack, or pin it), confirm it loads Slack's get-started page instead of the generic sign-in page


a5af:
@codex review

PR #3272, opened 2026-09-16
feat(srv): a fallible pending-migration check that fails closed

AgentX-asaf:
## Summary

Groundwork for **P2** of #3258. **No behavior change** — this adds the primitive the boot gate needs and cannot be built without, so the gate itself lands as a reviewable change of its own rather than dragging this along with it.

## Why

`count_pending_migrations` returns `0` on every error, by its own contract:

> Returns 0 on any error so startup is never blocked

That's the right default for its current caller — `AGENTMUXSRV-ESTART`'s `pending_migrations:` field is reported *after* migrations run, so a wrong count costs nothing. It's the wrong default for a caller using the count to decide **whether the schema is current**, which is what the deferred-migration design in `SPEC_FAST_STARTUP_UPGRADE_OWNS_MIGRATIONS_AND_UPDATES_2026_09_15.md` needs: a transient store-open failure reading as "nothing pending" would take the fast path and run the new binary against a store that may be many migrations behind.

Codex raised exactly this as a P1 against the spec in #3264; invariant **S1b** ("couldn't tell" is not "nothing pending") is the rule that came out of it, and this is the primitive that lets a caller honor it.

## Change

- `try_count_pending_migrations(data_dir) -> Result<usize, PendingCountError>`. The error names *which* store and *why*, so the blocking upgrade screen can show a human something better than "could not count".
- `count_pending_migrations` becomes a thin wrapper over it with **exactly** its previous behavior — including the partial case: an unreadable **channel** store still reports the global-scope count rather than dropping to `0`. That's why `ChannelStoreUnreadable` carries `global_pending`.

## Testing

- `an_unreadable_shared_store_is_an_error_not_a_zero` — strict says `Err`, wrapper still says `0`.
- `the_lenient_wrapper_keeps_reporting_the_global_half_it_could_read` — same inputs to both functions; strict refuses to answer, lenient reports the readable half. That divergence is the whole point, so it's asserted directly rather than implied.
- `a_pending_count_error_says_which_store_and_why` — the message carries the path and the underlying SQLite error.
- `cargo test -p agentmux-srv`: **3509 passed, 0 failed**.

## Not in this PR

The boot gate itself (skip the inline migrate, upgrade-only srv mode, blocking screen) and the snapshot-ordering question — §4.1's other half — since today's check still opens stores read-write and may create `objects.db`. Both are P2 proper.


a5af:
@codex review

AgentX-asaf:
Fixed in `57bfdc3` — correct catch, and it exposed a second-order version of the same problem.

`mod runner` is private and only a handful of items are re-exported, so `try_count_pending_migrations` and `PendingCountError` were unreachable from `bootstrap.rs`, the module this primitive exists for. Both are now exported.

But exporting alone would have left them as `unused import` warnings, which is the same defect one layer up: an API nothing calls is an API nobody has checked. So the primitive now has a real caller — the `AGENTMUXSRV-MIGRATING` hint in `bootstrap.rs` goes through the strict check and reacts leniently via a new `PendingCountError::readable_pending()`.

**Behavior is still unchanged**, deliberately: `readable_pending()` reproduces exactly the number that call site produced before (the global-scope count when only the channel store failed, `0` otherwise), so the ESTART deadline is sized identically. What's new is that an unreadable store is now *visible in the log* instead of silently becoming a count — and the two cases are logged distinctly, because "one store was readable and this count is partial" and "nothing was readable and this is a placeholder" mean different things to whoever reads that line.

`cargo test -p agentmux-srv`: 3510 passed.

One unrelated note in case it shows up in CI: `backend::agent_session::tests::archive_moves_content_and_clears_current` failed once in a full-suite run here, then passed in isolation and on a clean re-run. It touches no migration code; looks like a parallel-execution flake rather than anything from this PR.

PR #3277, opened 2026-09-16
docs(specs): widget context-menu — pinned-bar parity + Open in New Tab (Phase 2)

AgentO-asaf:
## Summary
- Right-clicking a widget on the **pinned bar** currently shows a generic "New Window" item that opens a blank default-layout window — indistinguishable from right-clicking empty title-bar space — even though the handler already has the clicked widget's definition in scope and just doesn't use it.
- This is because Phase 1 (`SPEC_WIDGET_CONTEXT_MENU_OPEN_ACTIONS_2026_06_24.md`) only fixed the More-dropdown/flyout menu builder; its "no divergence needed" claim about the pinned-bar call site sharing the same menu did not hold in the shipped code.
- Phase 1 also explicitly deferred a third action, "Open in New Tab."
- This spec designs both fixes together: **"Open in New Window"**, **"Open in Floating Pane"**, and the new **"Open in New Tab"**, via one shared `buildWidgetOpenActions()` builder consumed by both the native (pinned-bar) and DOM (More-dropdown/flyout) menu paths, so they can't silently diverge again.
- Confirmed cross-platform by construction — every primitive it depends on (`open_new_window` view/meta plumbing, `pane.open{floating:true}`, `WorkspaceService.CreateTab`+`createBlockOnModel`) already has zero OS-conditional branches.
- **Design/spec only — no implementation in this PR.**

## Test plan
- [x] `bash scripts/check-doc-status.sh docs/specs/SPEC_WIDGET_CONTEXT_MENU_OPEN_ACTIONS_PHASE_2_2026_09_16.md` → ok
- [x] `bash scripts/gen-docs-index.sh` run, `docs/specs/INDEX.md` regenerated and committed


a5af:
@codex review

Issue #3242, opened 2026-09-15
tracking: widen Linux package formats + a dedicated downloads page


Agent3-asaf:
## Why this exists

`docs/specs/SPEC_LINUX_DISTRO_TARGETS_AND_DOWNLOADS_PAGE_2026_09_15.md` — Linux release/nightly builds shipped only an AppImage (x86_64), the narrowest of the three platforms by a wide margin (Windows: 3 formats; macOS: 1 format but was mistakenly believed to have 2 architectures until Codex corrected that during review). The spec also scoped a dedicated `/download` page for `agentmuxai/agentmux-landing`, modeled on waveterm.dev/download, since today's download UX is a homepage section only.

No prior spec, issue, or discussion covered either half of this before it was written fresh — confirmed by searching `docs/specs/`/`specs/` and GitHub issues/discussions across `agentmux`, `agentmux-cloud`, `agentmux-landing`, and `shared-infrastructure`.

### Spec
- [x] `docs/specs/SPEC_LINUX_DISTRO_TARGETS_AND_DOWNLOADS_PAGE_2026_09_15.md` — #3234 (merged). Codex caught 2 real inaccuracies before any code existed: macOS doesn't actually ship an x64 `.dmg` (arm64 only), and `release.yml`'s asset-publishing step needed a matching update for any new format — both fixed in the spec text.

### Phase 1 — `.deb` + `.tar.gz`
- [x] `stage-linux-runtime.sh` (new shared staging step), `build-deb-linux.sh`, `build-tarball-linux.sh`, CI/release wiring — #3236 (merged)
- [x] Landing-page plumbing (`fetch-release.mjs`, `Download.tsx`, `DownloadButton.tsx`) — agentmux-landing#150 (merged)
- [x] Per-asset S3 mirror-check correctness fix (found during #150's review, needed a second round) — agentmux-landing#151 (merged)

**Real bugs caught by ReAgent/Codex review and fixed before merge:**
- Missing `LD_LIBRARY_PATH` in both new launch entry points — would have made `.deb` and `.tar.gz` completely non-functional (`agentmux-cef` has no RPATH).
- A dropped `AppDir` wipe during the `stage-linux-runtime.sh` extraction — would have let stale files silently carry forward across local rebuilds.
[...]

PR #3238, opened 2026-09-15
feat(linux): add .rpm release/nightly packaging (Phase 2)

Agent3-asaf:
## Summary

Phase 2 of `docs/specs/SPEC_LINUX_DISTRO_TARGETS_AND_DOWNLOADS_PAGE_2026_09_15.md` (Phase 1: #3236, merged). Adds `.rpm` alongside AppImage/`.deb`/`.tar.gz`, using the same `fpm`-over-the-shared-staged-tree mechanism as Phase 1's `.deb`.

- **`scripts/build-rpm-linux.sh` (new)** — packages the shared staged tree into a `.rpm` via `fpm -t rpm`. Same `/opt/agentmux` install layout as `.deb`, same wrapper, same desktop integration. Written with both fixes Codex/ReAgent caught on Phase 1's `.deb` (`LD_LIBRARY_PATH`, the AppArmor userns-fix helper bundling) already applied from the start, not discovered independently. No `Requires:` declared, matching `.deb`'s static-bundling choice (spec OQ4, still open) — see the script's header comment on why declaring dependencies for one format but not the other would be an inconsistent policy with no visible reason.
- **`scripts/package-linux.sh`** — `--format=` now also accepts `rpm`.
- **`.github/workflows/build-linux.yml`** — installs `rpmbuild` (`fpm -t rpm` shells out to it, unlike `-t deb`'s pure-Ruby path), builds all four formats from one compile, uploads `.rpm` as both a workflow artifact and a release asset.
- **`.github/workflows/release.yml`** — extends the publish job's asset-flattening allowlist to include `*.rpm` — same gap Codex caught for `.deb`/`.tar.gz` on #3236, closed here before it could repeat for this format.
- **`Taskfile.yml`** — `task package:linux:rpm` / `package:release:linux:rpm`, same conventions as the `:deb`/`:tarball` variants.
- **`docs/linux.md`** — documents `.rpm` in the Package formats table.

Companion PR: agentmuxai/agentmux-landing wires `fetch-release.mjs`/`Download.tsx`/`DownloadButton.tsx` to recognize the new `.rpm` asset, same pattern as Phase 1's `.tar.gz`.

## Test plan
- [x] `bash -n` on `build-rpm-linux.sh` and the modified `package-linux.sh`
- [x] YAML parses cleanly for both changed workflow files and `Taskfile.yml`
[...]

PR #3278, opened 2026-09-16
docs(specs): commit the multi-format release matrix draft (never-committed analysis)

GenericAgentX-asaf:
## Summary

`docs/specs/SPEC_MULTI_FORMAT_RELEASE_MATRIX_2026_09_10.md` existed **only as an uncommitted file in a working tree** — never on any branch — alongside its `INDEX.md` registration. One `git clean` from being lost. This commits it.

It predates `SPEC_LINUX_DISTRO_TARGETS_AND_DOWNLOADS_PAGE_2026_09_15.md` (#3234) by five days and is broader. It remains the only written analysis of the parts that spec doesn't cover:

- **Windows `.msi`** (WiX) — argued as the cheapest item in the whole matrix; no new CEF binary needed
- **Linux portable archive** and **`.pacman`**
- **Snap confinement** — concludes AgentMux needs *classic* confinement (it spawns arbitrary agent CLIs), hence Canonical manual review, with `snap install --dangerous` direct download as the de-risking path
- **macOS Intel** — blocked by the same missing-CEF-binary class as Linux arm64, which #3242 doesn't currently track
- The **CEF-binary dependency** gating the arch-blocked rows

Its central framing — CEF binary cost is paid once per **arch**, not per **format**, so one arm64 binary unlocks every Linux format at once — is what makes the arm64 item in #3242 tractable to scope.

## Changes beyond committing it verbatim

Only the `**Status:**` header. It claimed *"Draft — proposal only, nothing in this doc is implemented yet"*, which is now stale: Phase 2 (`.deb`/`.rpm`) and part of Phase 1 (portable archive, shipped as `.tar.gz` rather than `.zip`) have since shipped for x86_64 via #3236 / #3238.

The header now records that supersession, points at #3234 as the spec to read first for Linux formats, and carries the verification I posted to #3242:

- `agentmuxai/cef` publishes exactly **three** binaries — `linux-x86_64`, `macos-arm64`, `windows-x86_64`, all at `152.0.7977.83`. **No `linux-aarch64` or `macos-x86_64` exists at any version** — checked across all 13 releases' *assets*, not just their titles.
- The fork is **public**, not private. #3242's *"needs someone with `agentmuxai/cef` access to check first"* premise was wrong, and blocked that item unnecessarily.
- The fork has **no CI workflows at all** — every binary is built by hand, so a new arch is a manual multi-hour Chromium build, not a CI matrix entry.
- Offsetting that: `patch.cfg` is shared across all platforms (`CEF_FORK_MAINTENANCE.md`), so an arm64 Linux build needs **no new patch work** — it reuses the existing `agentmux/7977-*` branches unchanged.

Status stays `draft` (not `superseded`): the doc never became the implementation spec for anything, and per `check-doc-status.sh`'s own guidance a `superseded` status would require a single `Superseded-by:` pointer, which would misrepresent a doc only *partially* overtaken.

## Test plan

- [x] `scripts/check-doc-status.sh` — ok (2 docs checked); header's first word parses to `draft`, a valid enum value
- [x] `scripts/gen-docs-index.sh --check` — INDEX.md is current (894 specs)
- [x] `scripts/check-spec-citations.sh` — ok (2 files checked); all added citations resolve
- [x] `INDEX.md` regenerated with `gen-docs-index.sh`, **not** hand-edited — the stray working-tree edit used a different row format than the generator emits and would have failed the specs-index gate

Docs-only; no changeset (no runtime change to release).

Refs #3242


a5af:
@codex review

GenericAgentX-asaf:
All three P2s verified against the tree and fixed in `73e1217bf` (batched into one commit to avoid cancelling in-flight runs).

**P2-1 — ARM64 needs a Rust binding, not just the CEF binary.** Confirmed: `Cargo.toml` documents that the pinned `cef-rs` fork appends `begin_window_drag` to the **`linux_x86_64`** binding specifically, and without that field `--features patched-libcef` fails with `error[E0609]`. My "no new patch work" line was true for the CEF-source patches and wrong to generalize. Scoped the claim and added §3.6a for the binding work (different repo, not covered by `patch.cfg`); phase 5 now cross-references it. I've also posted a correction on #3242, where I made the same claim publicly.

**P2-2 — MSI signing has nothing to reuse.** Confirmed: `build-windows.yml` has no signing step at all (its only signing-adjacent line is `Build MSIX (unsigned — Store re-signs on ingest)`), and `package-msix.ps1` only does `New-SelfSignedCertificate` for local `-Sign` testing. Rewrote §3.1's signing bullet and pulled certificate acquisition into phase 0 — an MSI has no Store-re-signing equivalent, so it's the first artifact that actually needs a real Authenticode cert. Also softened the "cheapest item in the whole matrix" claim, which only held while signing was assumed free.

**P2-3 — phase 6 dropped `.AppImage`.** Confirmed: the goal matrix promises `.AppImage` on ARM64, but phase 6 was scoped to "§3.2–3.4's formats", which covers zip/deb/rpm/pacman/snap and not AppImage. Added it explicitly.

PR #3279, opened 2026-09-16
feat(widgets): Open in New Window / Floating Pane / New Tab, on the pinned bar too

AgentO-asaf:
## Summary
Implements `docs/specs/SPEC_WIDGET_CONTEXT_MENU_OPEN_ACTIONS_PHASE_2_2026_09_16.md` (#3277).

- Right-clicking a widget on the **pinned bar** now shows the same three actions the More-dropdown already had (plus the new one): **Open in New Window**, **Open in Floating Pane**, **Open in New Tab** — instead of a generic "New Window" that opened a blank default-layout window.
- New `frontend/app/window/action-widgets-menu.ts`: `buildWidgetOpenActions()` is the single shared source of these three actions, consumed by both of `action-widgets.tsx`'s menu builders (native `ContextMenuModel` for the pinned bar, DOM `PopoverMenu` for the More dropdown / pinned-parent flyout) so they can't silently diverge again — that's exactly how the pinned bar fell out of sync the first time.
- `Open in New Tab` (`frontend/app/tab/tab-presets.ts`'s new `openViewInNewTab()`) reuses the same `CreateTab` + `waitForLayoutModel` + `createBlockOnModel` sequence `EditorViewModel.openInNewTab` already uses successfully — not the `pane.open` RPC, which silently fails to render against a brand-new `tab_id`.

**Backend fix found during manual verification (`task dev`), not in the original spec:** "Open in New Window" was seeding the new window with the default 3-pane layout and then `pane.open`-ing the widget as a 4th pane alongside it, instead of replacing it. Root cause: `CreateWindow`'s existing `initial_view`/`initial_meta` frontend mechanism only ever *added* a pane after the window was already seeded with the default layout. Fixed at the source — `CreateWindow` (`agentmux-srv/src/server/service/window_create.rs`) now takes optional `seedView`/`seedMeta` args (same additive pattern as the existing `restoreIfAvailable`) and, when present, seeds the fresh tab with ONLY that one block via a new `single_leaf_tree()` builder (`agentmux-srv/src/backend/wcore/mod.rs`) instead of the default agent/swarm/sysinfo seed. `frontend/app-init.ts`'s pool-mode and cold-path new-window bootstraps now pass `initialView`/`initialMeta` straight into `CreateWindow`'s seed instead of a separate post-hoc `pane.open` call.

Verified live end-to-end in `task dev` (all three actions, including the New Window fix) — see thread for details.

## Test plan
- [x] `cargo test -p agentmux-srv wcore::` — 23 passed, including new `single_leaf_tree_wraps_exactly_one_block_with_no_split`
- [x] `cargo test -p agentmux-srv new_window_with_seed_view` — new regression test passes (seeds exactly 1 block, correct meta, single-leaf layout)
- [x] `cargo test -p agentmux-srv new_window_seed_blocks` — pre-existing default-3-pane-seed regression test still passes unchanged
- [x] `npx tsc -p tsconfig.citypecheck.json --noEmit` — clean (3 pre-existing, unrelated `.bench.ts` errors only)
- [x] `npx vitest run frontend/app/window frontend/app/tab frontend/app/view/editor` — 112 passed, including new `action-widgets-menu.test.ts`
- [x] Manual: `task dev`, right-clicked several widgets on the pinned bar and in the More dropdown — all three actions work, "Open in New Window" opens with only the clicked widget


a5af:
@codex review

PR #3280, opened 2026-09-16
Manoz@Area54: feat(agent-picker): rubbery poof + reflow animation on My Agents row delete

GenericAgentX-asaf:
## Summary
- Deleting a row in the My Agents picker used to be visually instant — worse, every OTHER row was liable to tear down and rebuild on the same refetch, since `<For>` reconciles by reference and `createResource`'s refetch hands back a brand-new array of brand-new objects each time.
- Mirrors `rows()` into a `reconcile(..., { key: "definition_id" })`-keyed store (same idiom already used in `drone-model.ts`) so per-row identity survives a refetch — a prerequisite for any per-row animation, not just this one.
- Wraps the row list in `solid-transition-group`'s `<TransitionGroup>` for a FLIP-based exit ("poof" — scale/blur/fade, reusing `--motion-spring`, a previously-defined-but-unused overshoot easing token) and move (surviving cards glide into their new grid cell instead of snapping) animation.
- Optimistic local removal on successful delete, matching the existing "optimistic collapse" pattern already used elsewhere in this file, so the animation starts immediately rather than waiting on the `agents:changed` broadcast round-trip.
- Respects `prefersReducedMotionAtom` (JS gate, classes swapped to `""`) plus a `@media (prefers-reduced-motion: reduce)` CSS backstop, mirroring the existing two-layer pattern already used for view transitions (`app.scss:295-311`).
- New dependency: `solid-transition-group` — chosen over a hand-rolled FLIP implementation per repo-owner direction ("use the most robust-performance solution, engineering cost isn't important"); see the spec's §4.3/§5.1.

Full design writeup: `docs/specs/SPEC_AGENT_ROW_DELETE_ANIMATION_2026_09_16.md`. Companion to the already-shipped `docs/specs/SPEC_AGENT_DELETE_2026_09_16.md` (backend delete path, unchanged here).

## Test plan
- [x] `npx tsc --noEmit -p tsconfig.citypecheck.json` — clean
- [x] `npx stylelint frontend/app/view/agent/styles/_recent-sessions.scss` — no new violations (6 pre-existing hex-color violations, confirmed unchanged via stash diff)
- [x] `task dev` — app boots and renders cleanly, no console/build errors
- [x] Manually verified live in a `task dev` build: deleted a test agent from My Agents, confirmed the poof + reflow animation
- [ ] Automated test coverage for the new store-reconcile/optimistic-removal logic — not added in this pass; flag if wanted before merge


a5af:
@codex review

PR #3281, opened 2026-09-16
Vmer@asaf: feat(shell): muxsh Phase 2b — view/edit/pane-list/run/config/agent

a5af:
## Summary

Implements `docs/specs/SPEC_MUXSH_FULL_COLLECTION_2026_09_16.md`'s Phase 2b build list, restructuring `muxsh.mjs` into a noun-verb dispatcher (`open`/`web`/`view`/`edit` as bare top-level shortcuts, `pane`/`run`/`config`/`agent` as proper nouns) per the spec's own §2.1 grammar.

**New subcommands:**
- `muxsh view <path-or-url>` — auto-detects editor/browser/media
- `muxsh edit <file>` — always forces the editor view
- `muxsh pane list` (`--json`)
- `muxsh run <cmd...>` / `--status <id>` / `--stop <id>`
- `muxsh config edit` / `muxsh config path [data|config|logs|shared]`
- `muxsh agent list` (`--json`) / `muxsh agent send <name> <message>`

**Two real corrections found during implementation** (not assumed from the spec's own first draft — every route re-verified against source before writing code): `pane close` was originally scoped here too, but the real route (`ClosePane`) requires an HMAC signature from the calling **agent's own** `AGENTMUX_JEKT_KEY` — a credential only injected into `agentmux-mcp`'s spawn env for a registered agent, never into a terminal pane's. `muxsh` structurally cannot produce it. Similarly, `secret list`'s real route (`IdentityAccounts`) is scoped to a specific registered agent (`$AGENTMUX_AGENT_ID`), not to "whoever is running this terminal" — a conceptual mismatch, not a missing route. Both moved out of this pass; full account in the spec's §5.5/§5.6.

**DRY manifest extended** (`docs/specs/app-api-manifest.json`, Phase 2a's mechanism) to cover `shell.{create,status,stop}` and `agent.send`, with matching contract tests on both sides:
- Rust: `agentmux-common/src/api_types.rs` (shell.* — exact field-set match) and `agentmux-srv/src/backend/reactive/types.rs` (agent.send — subset match, since `InjectionRequest` carries many more jekt-signing fields `muxsh` never touches; enumerating all of them would add no real protection for muxsh's own risk surface).
- Node: `muxsh.contract.test.mjs`.

## Verification
- `npx vitest run agentmux-srv/src/backend/shellintegration/` — 180/180 passing. `muxsh.test.mjs` was substantially rewritten: `parseArgs`'s return shape now discriminates on a `command` field (noun-prefixed commands like `pane-list` need one; the old bare `subcommand` field alone was sufficient only for the flat open/web case).
- `cargo test -p agentmux-srv -p agentmux-common app_api_manifest_contract_tests` — 7/7 new tests passing.
- `cargo check -p agentmux-srv -p agentmux-common` clean.
- `bash scripts/check-doc-status.sh` / `gen-docs-index.sh --check` clean.
- Not verified: an actual `task dev` run exercising each new subcommand against a live instance — no GUI available in the environment this was prepared in.

## Test plan
- [ ] CI green
- [ ] Manual: `task dev`, exercise `muxsh view`/`edit`/`pane list`/`run`/`config edit`/`config path`/`agent list`/`agent send` against a live instance


a5af:
@codex review

PR #3282, opened 2026-09-16
fix(agent-pane): pin auth/Compact sides in the composer strip

AgentA-asaf:
## Summary

- "Logged in"/"Not logged in" (`auth`) and the Compact button (part of the `ctx` slot) swapped sides exactly at whatever width crossed the composer strip's single-row/multi-row boundary. Root cause: `computeBalancedLeftKeys` (single row) and `computeComposerRows`'s two-pointer pairing (multi-row) decide `auth`/`ctx`'s relative order independently, with no shared state, and disagree.
- Extends the existing anchor mechanism (Rev 8, `runtime`/`hostShell`) to a second pinned pair via a new `pinnedPairs` param on `computeComposerRows`: `auth`/`ctx` are excluded from both algorithms' free search and always resolve to `auth=left, ctx=right`, in both the single-row and multi-row regimes. Only engages when both are present in a given render — either alone falls through to the ordinary free pool, unchanged.
- Full design/rationale: `docs/specs/SPEC_COMPOSER_STRIP_AUTH_COMPACT_SIDE_STABILITY_2026_09_16.md`.

## Test plan

- [x] New test cases added (`AgentComposerStrip.test.tsx`) covering the single-row pin, the multi-row pin, a width sweep across the fit boundary (the actual reported defect), and the "only one of the pair present" fallback.
- [x] Local `vitest` is blocked by a pre-existing, unrelated `solid-refresh`/JSDOM environment issue (reproduces identically with this diff stashed) — the new cases were validated standalone against the extracted pure functions instead (all pass, including a direct before/after against the old unpinned code path proving the fix changes behavior).
- [x] Verified against the real compiled app via `task dev`: resized the actual window across the single-row/multi-row boundary — "Logged in" stayed left of "Compact" in both states, no swap.
- [ ] CI


a5af:
@codex review

PR #3286, opened 2026-09-17
test(tools): pane-load.mjs — a hand repro for cross-pane input lag

Agent1-asaf:
## Summary

The cross-pane typing lag reported earlier in this thread is still undiagnosed — `#3249` fixed a real defect on the `controllerinput` path but turned out to be far too small to explain the symptom, since keystrokes actually travel on `blockinput`, a different path entirely.

The reason it keeps slipping investigation is that nobody could summon it on demand. This is that missing first step: run it in one pane, type in another while it floods, see whether your keystrokes stutter. It measures nothing — a human's fingers are the instrument — it just produces load heavy and varied enough to provoke the symptom, for a window long enough to actually type in.

## Why not just `yes`

`bench-term-cross-pane.mjs` floods pane A with `yes`: a stream of `y\n`, maximal bytes and near-zero terminal work, which xterm parses almost for free. Output that makes a UI feel bad is escape-sequence heavy — cursor jumps, colour churn, in-place redraws, what a build log or TUI emits — and those cost parser/renderer time *per byte*, not just bandwidth. A repro built on `yes` can be perfectly green while the thing users complain about is untouched, which is roughly the story of this bug so far.

## The mode that reproduces is the finding

Three load shapes, each leaning on a different mechanism, so narrowing down points somewhere specific:

| mode | shape | implicates |
|---|---|---|
| `text` | bulk bytes, moderate writes | WS egress, FileStore write-through, xterm raw parse |
| `paint` | bulk escape sequences | xterm parser + renderer — real repaint work per frame |
| `spinner` | many tiny writes | per-write overhead: WS frames, event dispatch, `PTY_COALESCE_WINDOW` |
| `mixed` (default) | rotates all three | find out *if* it reproduces, then narrow with one mode |

Measured into a pipe, the three are genuinely distinct: spinner does 522k writes/s at 26 MB/s; paint does 54k writes/s at 394 MB/s. Same "flood," different load class.

## Two defaults exist because the naive version is a footgun

Both bit me while verifying this script:

- **`--max-mb 64`** is a safety valve, not a tuning knob. A pane's terminal output is written through to the FileStore for scrollback persistence, so an unbounded flood is an unbounded write to the user's disk — these modes sustain 250-400 MB/s uncapped.
- **Auto-pacing** spreads that budget across the requested duration. Without it, paint mode burned the whole 64 MB cap in 0.2s — the "10 second typing window" was over before a human could move their hands, which defeats the entire point. `--throttle-ms` or `--unpaced` opt out.

## Scope

This loads the **xterm path only** — a Terminal pane, or an agent pane's shell drawer. The agent pane's own conversation rendering (`MarkdownBlock`/`ChunkList`) is a different code path this can't touch, and the original report was about an *agent* pane producing output — so a clean run here doesn't clear that path. README documents how to load it instead (have an agent `cat` a large file).

## Usage

```bash
node tools/tests/pane-load.mjs --secs 10
```

Open two panes, run in A, move cursor to B during the 3s countdown, type continuously through the flood.

## Verification

Ran every mode and read the reported throughput rather than assuming the generator works: text 258 MB/s uncapped, spinner 26 MB/s / 522k writes/s, paint 394 MB/s. Verified the cap fires (paint hits 8 MB instantly when capped low) and that auto-pacing actually fills the requested window (5.0s at 12.8 MB/s vs. `--unpaced` burning the same budget in 0.2s).


a5af:
@codex review

PR #3285, opened 2026-09-16
Manoz@Area54: refactor: rename Wave* identifiers to Mux* (#851 phase 1)

GenericAgentX-asaf:
Closes nothing yet — **phase 1 of #851**, which has sat 125 days (Deferred the day it was filed, after a mid-implementation attempt died on "a long tail of cross-references").

It was also getting *worse*: `WaveEvent` 223 → 232 and wave-referencing source files 285 → 306 between the 2026-09-06 audit and today.

## Scope: identifiers only

This renames every Wave-derived **identifier** — the surface the compiler verifies end to end. It deliberately excludes the **string/contract** surface, because that half isn't compiler-checked and is where the real risk lives.

**Renamed** (466+ occurrences, 183 files):
- PascalCase types → `Mux*` (`MuxEvent`, `MuxObjUpdate`, `MuxObj`, `MuxKeyboardEvent`, `MuxFile`, `MuxWindow`, `MuxLock`, `MuxInfoData`, …)
- camelCase with `Wave` **embedded mid-identifier** — the ones a naive `\bWave` sweep misses: `getWaveObjectAtom` → `getMuxObjectAtom`, `useWaveObjectValue`, `reloadWaveObject`, `fetchWaveFile`, `loadAndPinWaveObject`, `initWave`/`initHostWave`/`initCefWave`, +20 more
- Rust snake_case: `wave_obj_*` → `mux_obj_*`, `get_wave_data_dir`, `get_wave_config_dir`, `get_wave_db_dir`, `get_wave_lock_file`, `build_wave_env`, `broadcast_wave_obj_updates`, …
- Files: `wave_obj_bridge.rs` → `mux_obj_bridge.rs`, `waveutil.ts` → `muxutil.ts`, `wave-file.ts` → `mux-file.ts`
- Prose in comments/log strings describing the app: "Wave" → "AgentMux"

**Deleted as dead code** (zero consumers outside the type file, each confirmed): `WaveAIOptsType`, `WaveAIPacketType`, `WaveAIPromptMessageType`, `WaveAIStreamRequest`, `WaveAIUsageType`, `WavePointerData`, `CommandGetWaveAIChatData`, `CommandWaveAIAddContextData`, `CommandWaveAIToolApproveData` — Wave Terminal leftovers.

## Deliberately NOT renamed — these are compatibility, not branding

A naive "finish the rename" sweep breaks all of these. Calling them out because that's the trap this issue has been sitting on:

| Kept | Why |
|---|---|
[...]

PR #3287, opened 2026-09-17
Manoz@Area54: refactor: wos/wps/wstore -> mos/mps/mstore (#851 phase 3)

GenericAgentX-asaf:
Phase 3 of #851, following #3285 (phase 1, merged). Renames the **Wave-derived abbreviations** — the last large compiler-verified chunk.

| Was | Now | Refs | Meaning |
|---|---|---|---|
| `WOS` / `wos` | `MOS` / `mos` | 416 | "**W**ave **O**bject **S**tore" |
| `WPS` / `wps` | `MPS` / `mps` | 660 | "**W**ave **P**ub/**S**ub" (per the module's own doc comment) |
| `wstore` | `mstore` | 2,567 | "**W**ave store" — the `Store` handle threaded through srv |
| `gotypes.d.ts` | `srv-types.d.ts` | 18 files | named for a Go backend removed long ago |

Module files moved: `wos.ts`→`mos.ts`, `wps.ts`→`mps.ts`, `wps-events.ts`→`mps-events.ts`, `backend/wps.rs`→`backend/mps.rs`, `wps_client.rs`→`mps_client.rs`.

Three commits, one per rename, so each is reviewable independently.

## Kept deliberately

- **`/agentmux/wps/publish`** — a real client/server contract, not intra-build lockstep: `agentmux-bashwrap` is a **separate binary** that POSTs to it, and shell-integration scripts on disk invoke that binary. Renaming needs a compat alias serving both paths → phase 2. Its handler keeps the name `handle_wps_publish` on purpose so it still visibly matches the route it serves.
- `scripts/benchmarks/measure-performance.{sh,ps1}` still say "WaveMux" and look for `src-tauri/target/.../WaveMux.app` — a path gone since the Tauri→CEF migration, and nothing references these scripts. Dead tooling; deleting it is a separate call, not something to smuggle into a rename.

## A regression this PR introduced, and the fix

Worth calling out because it shows where the sweep's blind spot was.

`frontend/test/fixtures/agent-sessions/bash-with-live-log.session.ndjson` carries `"src":"wps"` as a **data value**. Phase 3b renamed the validator in `replay/loader.ts` to expect `"mps"` but not the fixture, because my sweeps only covered `.rs`/`.ts`/`.tsx` — fixture data was invisible to them.

I caught it by diffing failing-test *files* between this branch and clean `main` (92 vs 91), not by trusting the totals. Confirmed nothing in production emits that value — it exists only in that one fixture — so the fixture moved to `"mps"` with its validator. Then swept `.scss/.json/.ndjson/.mjs/.ps1/.sh/.yml/.html` for the same class of miss: only three stale comment references, now fixed.

## Verification

- [x] `tsc --noEmit` — clean
- [x] `npm test` — **91 failed files / 29 failed tests / 2960 passed, exact parity with clean `main`** (those failures are the pre-existing `@solid-refresh` jsdom issue on Windows)
- [x] `check-spec-citations` — passes across all 373 changed files (run preemptively; it caught one pre-existing dangling citation to a spec that never existed, removed rather than replaced with a guess)
- [x] Audited `wstore`'s string occurrences before renaming: all log/error text and `.expect()` messages, no keys


a5af:
@codex review

PR #3288, opened 2026-09-17
feat(browser): "Set as Start Page" in the bookmarks menu

AgentX-asaf:
## Summary

Adds a configured start page for the browser pane, as the **first entry** in the bookmarks dropdown — `docs/specs/SPEC_BROWSER_PANE_START_PAGE_2026_09_16.md`.

The spec went through a real revision before implementation: the first draft mirrored bookmarks' own persistence shape (a dedicated `shared_dir` file + a lazy `.get`/`.set` RPC pair, with a new sequential RPC added to `app-init.ts`'s boot sequence to preload it). On "we want robust fast performance, engineering cost is not important," that design was replaced with one that adds **zero** boot latency and gets **true live cross-window sync for free**, by riding the app's existing `FullConfigType`/`GetFullConfig`/file-watcher infrastructure instead of building a parallel one — the same reasoning is in the spec's revision note.

## Design

The value still lives in its own `shared_dir` file (`~/.agentmux/shared/browser-start-page.json`, global across channels — same reasoning as bookmarks: `settings.json` isolation is whole-file only, no per-key tiering). What's different is delivery:

- It rides inside `FullConfigType` (new `browser_start_page` field, wire `browserstartpage`), which the frontend already awaits **before its very first render** (`app-init.ts`'s `GetFullConfig` call) — not a new fetch of any kind.
- A file watcher plugs it into `ConfigState`, mirroring the existing `settings.json` watcher exactly, on the **same** shared `FsWatchPool` instance — so an edit from any window (or an external process) reaches every connected client via the same live broadcast `settings.json` changes already use. This closes the exact gap bookmarks' own spec explicitly accepted ("a menu opened before another window's edit lands shows a stale snapshot until closed and reopened").
- The `.set` RPC writes to disk, updates `ConfigState`, and broadcasts **immediately** — the same write-then-update-then-broadcast-now shape `COMMAND_SET_CONFIG` already has, rather than waiting on the fs watcher's ~300ms debounce.
- Along the way, factored out a `broadcast_full_config` helper that was previously duplicated verbatim between the settings-reload path and the `setconfig` handler — now three call sites share it instead of a third copy.

Frontend: `browserStartPageAtom` is a `createMemo` off `fullConfigAtom`, identical shape to `settingsAtom` — no new subscription code, inherits boot-time population and live updates for free. `browser-model.ts`'s initial-URL fallback becomes `meta.url > browserStartPageAtom() > DEFAULT_BROWSER_URL`.

The new menu row reuses `MenuItem.checked` — a field bookmarks' own spec named for exactly this row and never used — to show a check when the active page already is the start page. That required teaching the nav bar's custom `renderMenuItem` override about `checked` (mirroring `FlyoutMenu`'s own default markup for it), since it previously only handled the icon slot.

## Testing

- 11 new backend tests (`browser_start_page.rs` mirrors `bookmarks_store.rs`'s existing suite, plus `ConfigState`/watcher coverage).
- 2 new frontend tests (start page vs. `DEFAULT_BROWSER_URL` fallback).
- `cargo test -p agentmux-srv`: **3550 passed, 0 failed**.
- `tsc --noEmit`: clean.
- `vitest`: 2962 passed, 29 failed — confirmed byte-identical to this repo's own documented pre-existing Windows-only `@solid-refresh`/jsdom baseline (91 failed files / 29 failed tests, noted in the recent `Wave*`→`Mux*` rename PR's own verification), and confirmed none of the failures touch any file this PR changes.

**Not yet manually verified in a running `task dev` instance** — typecheck/unit-test coverage only, same caveat the bookmarks feature's own spec carried at the same stage.


a5af:
@codex review

AgentX-asaf:
Fixed in `43ca53a`. Codex's P1 is the important one — verified and it was real:

**The feature had no effect for the actual common case.** `defwidget@browser` (the widget bar's own Browser entry, `widgets.json:42`) bundled `meta.url: "https://agentmux.ai"` directly on the widget definition — and the fallback chain checks `meta.url` first, so `browserStartPageAtom()` was dead code for the single most common way of opening a browser pane. Removed the hardcoded URL (it duplicated `DEFAULT_BROWSER_URL`'s own value, so nothing is lost); purpose-specific browser widgets (Discord/Slack/Telegram/WhatsApp/Teams) keep their own real `meta.url`, which still correctly wins.

**ReAgent's two catches, also both real:**
- `MenuItem.checked` replaces the icon slot entirely with a check-or-blank glyph (confirmed by reading `FlyoutMenu`'s own default markup, which this nav bar's custom `renderMenuItem` override replicates) — since the row's `checked` value is always a defined boolean, the house icon could never actually render. Switched to the same label-flip convention the bookmark-toggle row right below it already uses, and reverted the now-unused `checked` handling out of the override.
- `browser_start_page` was the only `Option<T>` field in `FullConfigType` without `skip_serializing_if = "Option::is_none"` — added it, matching every sibling field and the TS type (`browserstartpage?: string`, no `null`).

**Codex's second catch:** a failed `.set` left `startPageError` set permanently — a later successful retry never cleared it. Now cleared before each attempt, not only in the catch branch.

Also regenerated `docs/specs/INDEX.md` — that's what the doc-status CI failure on the prior push actually was.

`cargo test -p agentmux-srv`: 3552 passed, 1 unrelated pre-existing flake (`handle_anomaly_tests`, confirmed passing in isolation). `vitest`: 2962 passed, 29 failed — byte-identical to the pre-existing baseline, confirmed via `frontend/app/window`'s own widget tests that the `widgets.json` edit introduces nothing new.

PR #3289, opened 2026-09-17
Manoz@Area54: refactor: use the shared CREATE_NO_WINDOW constant instead of raw 0x08000000

GenericAgentX-asaf:
First slice of the DRY/modularity item (row 12 of `REPORT_MIGRATION_WRAPUP_STATUS_2026_09_16.md`). I started by checking whether its headline metric held up — it didn't, and **correcting that is half this PR**.

## The metric was misleading, including in my own report

Both the 09-06 DRY audit and my report §2.6 claimed the call-site sweep never followed the `agentmux-common` lift, citing *"29 files still name `CREATE_NO_WINDOW` locally and 23 still define their own `fn now_ms`"*. Verified directly rather than trusting it:

- **Exactly one** `const CREATE_NO_WINDOW` exists in the tree (`agentmux-common/src/win32.rs`). **Zero** files re-declare it; **24 import it correctly.** The "29 files" figure counts usages and comments.
- The 23 `fn now_ms` "duplicates" are three-line delegators:
  ```rust
  fn now_ms() -> i64 { agentmux_common::time::now_ms() }
  ```
  One implementation, a short local call site. That's a reasonable idiom, not duplication.

So that lift is essentially **complete**. §2.6 is corrected in this PR — I wrote the wrong version of it, and a stale metric steering work is exactly what that report is about.

## The one real residue, now fixed

**~20 call sites across 4 crates passed the raw magic number `0x08000000`** instead of the shared constant. That's *worse* than a duplicate `const`: a magic number is unsearchable, so it's invisible to the very grep that would find it — which is why the audit's own metric missed it while flagging harmless usages.

| Crate | File | Sites |
|---|---|---|
| agentmux-cef | `commands/providers.rs` | 10 |
| agentmux-cef | `commands/cli_login.rs`, `sidecar.rs` | 2 |
| agentmux-launcher | `srv_spawner.rs` | 1 |
| agentmux-srv | `backend/tool_store.rs`, `server/cli_handlers.rs`, `server/install_handlers.rs` | 7 |

Used the fully-qualified `agentmux_common::win32::CREATE_NO_WINDOW` rather than adding a function-local `use` at each site — they're scattered through `#[cfg(windows)]` blocks and the qualified form needs no structural edit.

`agentmux-common`'s own `assert_eq!(CREATE_NO_WINDOW, 0x08000000)` keeps its literal: that test is what pins the ABI value.

## Test plan

- [x] `cargo check --workspace` — clean. This is Windows-only code compiled on Windows, so the changed call sites are genuinely type-checked rather than `cfg`-skipped.
- [x] `check-doc-status` + `check-spec-citations` — pass
- [x] Confirmed no raw `0x08000000` remains outside `common`'s pinning test and two explanatory comments


a5af:
@codex review

PR #3291, opened 2026-09-17
Manoz@Area54: refactor(rpc): first consumer of the generated RPC bindings (voice.checkPath)

GenericAgentX-asaf:
DRY cause 1 (row 12 of `REPORT_MIGRATION_WRAPUP_STATUS_2026_09_16.md`).

## The audit's framing was out of date, and the real gap was sharper

It said *"no Rust↔TypeScript codegen"*. Not true any more — `SPEC_RPC_BINDINGS_CODEGEN_2026_09_07.md` §3.4 **step 1 shipped**: `register_typed`, the `RpcSchema` registry, ts-rs export and `scripts/check-rpc-bindings.sh` are all live, with 12 types generated into `frontend/types/rpc/`.

What hadn't happened is step 2 — and the gap was more specific than "partially migrated":

> **Nothing in the frontend imported a single generated type.**

All 12 were generated, gated against drift, and **unused**, while the hand-written stubs kept their own inline duplicates of the same shapes. The pipeline was built and never connected at the far end. (Also worth noting: the spec's own Status still reads "no generator exists yet" — same stale-status pattern this whole workstream keeps turning up.)

## This connects it, end to end, for one command

- `CommandVoiceCheckPathData` / `VoiceCheckPathResult` named in `rpc_types` with `#[derive(ts_rs::TS)]` — replacing a private `VoiceCheckPathReq` and an anonymous `json!({"exists": ..})`, **neither of which the generator could see**
- `voice.rs`: `register_handler` → `register_typed`
- `rpc-api/voice.ts` now **imports** those generated types instead of restating `{ path: string }` / `{ exists: boolean }` inline
- New `@/types/*` tsconfig path. Vite resolves via `vite-tsconfig-paths`, so tsconfig stays the single source of truth — no second alias list to drift

## Behaviour is provably unchanged

This is a live RPC, so that needs more than "it compiles". `register_typed` deserialises with the same `serde_json::from_value`, then `serde_json::to_value(&resp)` → `Ok(Some(value))`. `VoiceCheckPathResult` serialises to `{"exists": bool}` — byte-identical to the old `json!`. Even the error prefix is unchanged, since the `{command}` it interpolates *is* `"voice.checkPath"`.

## Why the smallest domain first

`voice.ts` is 19 lines and one command. The spec orders migration `agent.ts` first (51 stubs, richest types) → `file.ts` last, but that ordering is about **type richness**, not about where to de-risk. This slice proves the whole loop — Rust struct → generated `.ts` → consumed by the stub → gate catches drift — and gives the remaining 16 stub files a template to follow.

## Test plan

- [x] `tsc --noEmit` — clean (confirms the generated types actually resolve through the new alias)
- [x] `bash scripts/check-rpc-bindings.sh` — **ok, 14 generated types current** (was 12)
- [x] `npm test` — 91 failed files / 29 failed tests / 2960 passed, parity with `main` (pre-existing `@solid-refresh` jsdom issue)
- [x] Wire-shape equivalence traced through `register_typed` rather than assumed


a5af:
@codex review

PR #3292, opened 2026-09-17
Manoz@Area54: docs: update the migration wrap-up as its own items land

GenericAgentX-asaf:
Keeping `REPORT_MIGRATION_WRAPUP_STATUS_2026_09_16.md` current as the work it tracks completes. It's the artifact people plan from — letting it rot is precisely the failure it was written about.

## Changes

- **Status** `proposed`/handed-off → **`active`**, a living doc updated as §5 items land.
- **§5.4 Wave → Mux** — 🔴 *"decide, don't drift"* → ✅ **DONE, #851 closed.** Records both halves of the decision: identifiers shipped (#3285, #3287 — **0 Wave-derived identifiers left, 306 → 91 files**), and the string half declined *with the evidence that decided it* — phase 3's silently-404ing P1 that neither the compiler nor 2,960 tests caught. Also names the three references that must never change (`~/.waveterm`, `WAVEMUX_AGENT_ID`, `wavepwsh.ps1`) so a future sweep doesn't break them.
- **§5.4a — NEW, DRY/modularity.** Four of five causes closed or dismissed, and **three of the five weren't what the audit's metric implied**:
  - Cause 1 mis-framed — codegen *existed*; the real gap was that **nothing imported a generated type** (12 generated, gated, unused). Closed by #3291.
  - Cause 4 real duplication, but exactly two instances and already diverging → abstracting would be premature.
  - Cause 5 was never duplication — one shared generic type name, different domains.
- **§7 scoreboard** — rows 12 and 13 updated.
- **§8** — recommendations 1 and 3 resolved, plus a new *"What the day actually taught"*.

## The part worth reading

Five audit claims were checked against code today. **Four were wrong in the "actually done" or "not really a problem" direction** — mandatory ABF, migration-failure-is-silent, the `agentmux-common` sweep, and DRY causes 4 and 5. One was wrong the other way and nearly shipped a bad restamp: the jekt trust spec looked complete enough to mark `implemented` when only its host half had landed.

The common thread: **these reports' metrics are load-bearing and unverified.** Counts of occurrences, files or named causes get read as counts of problems. That's why §5.6 — automating the reverse staleness check — stays the top recommendation despite being the least glamorous item on the list.

## Test plan

- [x] `check-doc-status.sh` — ok
- [x] `check-spec-citations.sh` — ok
- [x] Docs-only; no code touched


a5af:
@codex review

GenericAgentX-asaf:
Merged `main` into this branch — that resolves the P1 and codex's earlier P2, which were both correct about the tree they reviewed.

The branch was cut from `main` *before* #3291 landed, so the voice wiring the doc describes genuinely wasn't present here: `voice.rs` still showed `register_handler`, `frontend/types/rpc/` had no Voice type, and `voice.ts` still hand-declared its shapes. The claim was true of `main` and unverifiable in the PR — which is the same category of error the report is about, so it deserved the P1 rather than a hand-wave.

Now present in this tree and checkable:
- `agentmux-srv/src/server/app_api/voice.rs:28` → `engine.register_typed(`
- `frontend/types/rpc/CommandVoiceCheckPathData.ts`, `frontend/types/rpc/VoiceCheckPathResult.ts`
- `frontend/app/store/rpc-api/voice.ts:10-11` → `import type { … } from "@/types/rpc/…"`

Verified after the merge: `cargo check --workspace --tests` clean, `tsc --noEmit` clean, `check-doc-status` ok.

PR #3293, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate bookmarks, delete a hand-maintained type (DRY cause 1)

GenericAgentX-asaf:
Second domain after #3291 (voice), and the first to deliver the **actual payoff**: a hand-maintained type declaration is *deleted*, not just duplicated into a generated file.

## Changes

- `bookmarks.list` / `bookmarks.set` → `register_typed`. `()` is a legal `Req` for list, which takes no arguments.
- New named `BookmarksResult` / `CommandBookmarksSetData`, replacing an anonymous `json!({"bookmarks": ..})` and a private `BookmarksSetReq`.
- `BrowserBookmark` derives `ts_rs::TS` and generates.
- **`frontend/types/srv-types.d.ts`'s hand-written `BrowserBookmark` is DELETED**; its three consumers import the generated type. That's spec step 3 in miniature — the 2,755-line hand-maintained file shrinks for the first time.

## A real blocker the spec didn't anticipate

Worth reading before anyone does the remaining 15 domains:

> **ts-rs maps `i64` → `bigint`, not `number`.**

`BrowserBookmark.created_at: i64` generated as `created_at: bigint`, which isn't assignable to `number`, and every consumer treats it as a JS millisecond timestamp (`ToggleBookmarkInput.now: () => number`). It failed to compile — *loudly*, which is the lucky version. The dangerous version is a field that happens to typecheck anyway.

Fixed with `#[ts(type = "number")]` plus a comment noting ms-since-epoch stays exactly representable in an f64 until year 275760. **Every future `rpc_types` field with an i64 timestamp needs the same annotation.** The spec's §3.2 covers `#[serde(default)]` optionality but says nothing about integer width — this is the repo's first generated binding with a numeric field at all, which is why it surfaced only now.

## One deliberate shape change, verified not assumed

The hand-written type said `favicon_url?: string` / `created_at?: number`; the Rust fields are `String`/`i64` with `#[serde(default)]` — **always present on output**. The generated type makes them required, which is more accurate. Checked the consumers rather than assuming: the only constructor (`toggleBookmark`) already supplies both from a required `ToggleBookmarkInput.faviconUrl: string`, and the one defensive `favicon_url ?? ""` read stays valid.

## Tests kept, not dropped

The two handler tests pin real deserialization semantics — `#[serde(default)]` filling missing fields, and a missing `bookmarks` key still erroring. They were repointed at `CommandBookmarksSetData` rather than deleted along with the struct they named.

## Test plan

- [x] `cargo check` + `cargo test -p agentmux-srv` — clean, 17 passed
- [x] `tsc --noEmit` — clean (proves the deleted global type really is fully replaced)
- [x] `check-rpc-bindings.sh` — **ok, 17 generated types** (was 14)
- [x] `npm test` — 91 failed files / 29 failed tests / 2960 passed, parity with `main`


a5af:
@codex review

GenericAgentX-asaf:
Fixed — and both of you found it independently (codex P1, reagent P0). reagent's trace was exactly right: `RpcMessage.data` carries `Some(Value::Object({}))`, so `handle_request`'s `unwrap_or(Value::Null)` never fires, and `from_value::<()>` at `engine.rs:294` rejects a map.

The root cause is worth naming for the remaining 15 domains: **the old untyped handler ignored `_data` entirely, so this migration introduced the first real deserialization of that payload.** `register_typed` converts a previously-ignored input into a parsed one, and the compiler is equally happy either way.

Fix: `CommandBookmarksListData {}` — an empty struct accepts `{}` and ignores unknown keys, matching the old ignore-everything behaviour. ts-rs generates it as `Record<string, never>`, which is *exactly* what the stub already declared, so the two sides now agree by construction rather than by comment.

Regression test encodes the failure rather than the fix:

```rust
list_req_accepts_the_empty_object_the_stub_sends()
    // asserts CommandBookmarksListData parses {}
    // AND asserts () does not — so the reason survives in the test
```

Swept for other instances: bookmarks.list was the only `register_typed` site using a unit Req, so nothing else is latent.

One gap this exposes, which I'll carry into the remaining domains: **neither `check-rpc-bindings.sh` nor `tsc` verifies runtime payload shape** — they only check generated-type drift. Request-shape correctness needs a per-command test fed what the stub actually sends, which is now the pattern.

Verified: `cargo test -p agentmux-srv bookmarks` 14 passed incl. the new guard; `check-rpc-bindings` ok (18 types); `tsc` clean; `npm test` at parity with main.

PR #3294, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate reactive.registrations, delete 4 mirror interfaces

GenericAgentX-asaf:
Third domain, after #3291 (voice) and #3293 (bookmarks). **Biggest duplication deleted so far** — four hand-written TypeScript interfaces, one of which documented itself as:

```ts
/** Mirrors agentmux-srv's `AgentRegistration` (backend/reactive/types.rs). */
```

That comment is precisely the promise `check-rpc-bindings.sh` now enforces instead of asking a reader to trust.

## Changes

- `reactive.registrations` → `register_typed`
- `AgentRegistration`, `RemoteRegistrationEntry`, `MismatchAuditSummary`, `ReactiveRegistrationsResult`, `ReactiveRegistrationsParams` derive `ts_rs::TS` — **23 generated bindings now, was 18**
- `rpc-api/reactive.ts`'s four interfaces deleted, replaced by re-exports of the generated types **under the same public names** — so `index.ts` and `AgentRegistrationPanel.tsx` are untouched and `@/app/store/rpc-api` is unchanged

## Lessons from the previous two slices, applied up front

**The i64 trap (#3293):** six `u64` fields carry `#[ts(type = "number")]` from the start rather than being rediscovered. Also confirmed the boundary while here: **`pid: u32` needs no annotation** — only 64-bit integers become `bigint`, since u32 fits an f64 exactly.

**The request-shape trap (#3293's P0):** added guards fed literally what the stub sends —

```rust
params_accept_the_payload_the_stub_sends()   // { agent_id: "..." }
params_reject_a_missing_agent_id()
```

`register_typed` parses the payload *before* the handler runs, and neither `tsc` nor `check-rpc-bindings.sh` verifies runtime payload shape — they check generated-type drift only. This command's Req is a real struct with a required field, so it was never at risk the way `bookmarks.list`'s unit Req was; the test is cheap insurance that stays honest if either side is edited.

## One deliberate shape change

`error_message` generates as `string | null`, not the hand-written `error_message?: string` — the Rust is a plain `Option<String>` with no `skip_serializing_if`, so it really is present-and-null on the wire. `tsc` confirms the existing consumer handles it.

## Test plan

- [x] `cargo check` + the two new request-shape tests pass
- [x] `tsc --noEmit` clean (proves the deleted interfaces are fully replaced, including the `null` vs `undefined` change)
- [x] `check-rpc-bindings.sh` — **ok, 23 generated types** (was 18)
- [x] `npm test` — 91 failed files / 29 failed tests / 2960 passed, parity with `main`


a5af:
@codex review

PR #3295, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate agent:memory:* to register_typed, delete 7 hand-written globals

GenericAgentX-asaf:
@
Fourth domain in the RPC codegen migration, after `voice`, `bookmarks` and
`reactive` (#3294). Generates 13 of the 14 native-memory wire types from their
Rust definitions with ts-rs and deletes the hand-written copies that previously
sat in the **global** namespace in `frontend/types/srv-types.d.ts`.

Unlike the reactive domain (four local interfaces in one stub file), these were
ambient globals, so eight consumer files gain an explicit
`import type { ... } from "@/app/store/rpc-api"`.

## Not a wire change

All 13 generated types were diffed field-by-field against the declarations they
replace before anything was deleted: identical, including the `u64 -> number`
fields (`size_bytes`, `modified_at`, `created_at`) which needed
`#[ts(type = "number")]` to avoid being emitted as `bigint`.

## Two deliberate exceptions, both worth a look

**1. `NativeMemoryWriteProvenance` is NOT generated.** Its `detail` field is a
`serde_json::Value` that the frontend has always treated as an optional
property (`detail?: unknown`), and ts-rs rejects `#[ts(optional)]` on anything
that is not `Option<T>`:

```
error: `optional` can only be used on an Option<T> type
```

The obvious workaround -- change the field to `Option<serde_json::Value>` --
would let the generator dictate wire behaviour: `default_detail()` exists
precisely so an omitted detail deserializes to `{}` rather than `null`. So this
one type stays hand-written, and both sides carry a note saying why and that
they must be kept in sync by hand. This is a real expressiveness gap in ts-rs,
not an oversight -- flagging it because it will recur in later domains.

**2. `agent:memory:write_file` now sends `"data": null`** where the old untyped
`Ok(None)` omitted the `data` key entirely (`RpcMessage.data` is
`skip_serializing_if = "Option::is_none"`), because `register_typed` always
sends a data field. Both call sites
(`agent-native-memory-model.ts:200`, `:240`) `await` and discard the result,
and a repo-wide sweep for the command string found no other caller in any
crate, so the difference is unobservable. Verified with:

```
grep -rn "agent:memory:write_file" . --exclude-dir=node_modules --exclude-dir=target
```

## Tests

Adds `req_shape_tests` pinning each of the six commands to the literal JSON its
call site actually sends. This layer is not redundant: `tsc` only checks the
frontend against the *generated* types and `check-rpc-bindings.sh` only checks
that a generated type *exists* per command -- neither ever deserializes a real
payload into the Rust struct. That gap is exactly how the `Req = ()` rejects
`{}` bug on `bookmarks.list` reached runtime. Two of the seven are negative
assertions: `diff` must reject a payload with no `agent_id` (the ownership
check from the earlier P1), and provenance must default `detail` to `{}` not
`null`.

## Verification

- `cargo test -p agentmux-srv req_shape_tests` -- 7 passed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 36 generated types current
- `scripts/check-spec-citations.sh` -- ok (26 files), `check-doc-status.sh` -- no docs changed
- `npm test` -- 91 failing suites / 29 failed tests, byte-identical to the
  baseline on `main`. `native-memory-manager.test.tsx` is in that list and I
  edited it, so I ran it alone to confirm the cause is the pre-existing
  Windows `file:///@solid-refresh` jsdom error, not the new import.

agent_id: manoz-0803a

@


a5af:
@codex review

PR #3296, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the 11 session commands, delete 23 hand-written globals

GenericAgentX-asaf:
@
Fifth domain in the RPC codegen migration (after `voice` #3291, `bookmarks`
#3293, `reactive` #3294, `native-memory` #3295), and the largest so far: 23
types, 11 commands. First domain where the deleted declarations were ambient
**globals** in bulk, so several consumer files gain explicit imports.

## All 23 diffed before anything was deleted

Nineteen matched the hand-written declarations exactly. The other four are each
a deliberate decision, not a slip:

| Field | Was | Now | Why |
|---|---|---|---|
| `word_target`, `user_message` | `T?` | `T \| null` → **fixed to `T?`** | ts-rs emitted `\| null`; serde treats a missing `Option<T>` as `None`, so optional is the accurate shape. Added `#[ts(optional)]`. |
| `SessionResumePreflightResult.verdict` | `"resume" \| "recover" \| "fresh" \| "unknown"` | `string` → **fixed back to the union** | ts-rs emits a bare `string` for a Rust `String`. `inject-resume-preflight.ts` branches on these values and the test uses `SessionResumePreflightResult["verdict"]` as a type, so this would have been a silent loss of real safety. Preserved with `#[ts(type = ...)]`. |
| `AgentSessionReadResult.content` / `.modts` | `T? \| null` | `T?` | **Deliberate narrowing.** Both are `skip_serializing_if = "Option::is_none"`, so the server omits the key and can never send `null` — the `\| null` was unreachable. |
| `CommandAgentSessionListArchivesData.limit` | `limit?` | `limit` | **Deliberate widening of the requirement.** `#[serde(default)]` on a non-`Option` is the shape ts-rs cannot express as an optional property (it rejects `#[ts(optional)]` on non-`Option<T>`). No caller exists, so nothing is forced to start passing it. A test pins the wire behaviour so the gap stays visible. |

The `verdict` one is worth dwelling on: it is the first case where *generating*
a type would have been **worse** than the hand-written one. Codegen is not
automatically more accurate — it is accurate about the things the generator can
see, and a Rust `String` does not carry its own domain.

## Two frontend test mocks corrected

`useHistoryPagination.test.ts` and `AgentPicker.test.tsx` mocked
`AgentSessionReadCommand` as `{ content: null, modts: null }` — a shape the
server cannot produce. Changed to `{}`, which is what it actually sends. Both
still pass; the "no content" path is still exercised, because the production
code does a truthy check.

## TokenCounts

Generated, but its hand-written global **stays**, because four other global
types reference it and a global type cannot import. It goes away when those
domains migrate. Flagging rather than hiding it: this is temporary duplication,
and the two shapes are identical (`rename_all = "camelCase"` →
`cacheCreation`/`cacheRead`, which the hand-written copy already said).

## Also

- Three commands (`append_output`, `agent:session:archive`, `list_archives`)
  have **no non-test frontend caller**. They are still registered and
  reachable, so their shapes are pinned rather than left unverified — but this
  is more evidence for step 3 (deleting `rpc-api/` wholesale).
- `list_archives` previously did `serde_json::from_value(data).unwrap_or_default()`,
  silently swallowing a malformed payload into an empty `definition_id`. Under
  `register_typed` it returns an error instead. No caller, so no impact.
- Dropped a dangling spec citation the changed-files-scoped citation gate
  surfaced. The cited doc was never committed; `backend/agent_session/mod.rs`
  already carries a note resolving the identical citation, so this mirrors it
  rather than repointing to the similarly-named but different
  `SPEC_CONTINUOUS_SESSION_PERSISTENCE_2026_09_08.md`.

## Verification

- `cargo test -p agentmux-srv req_shape_tests` -- 16 passed (9 new + 7 from #3295)
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 60 generated types current
- `scripts/check-spec-citations.sh` -- ok (37 files), `check-doc-status.sh` -- ok
- `npm test` -- 91 failing suites, identical to baseline. `AgentPicker.test.tsx`
  is in that list and I edited it, so I ran it alone: the cause is the
  pre-existing Windows `file:///@solid-refresh` jsdom error, not the mock change.

agent_id: manoz-0803a

@


a5af:
@codex review

PR #3297, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the 6 fleet commands, type two shapes that had none

GenericAgentX-asaf:
@
Sixth domain in the RPC codegen migration (after `voice` #3291, `bookmarks`
#3293, `reactive` #3294, `native-memory` #3295, `session` #3296). Replaces four
hand-written `export interface` declarations and five inline anonymous request
shapes with twelve generated types.

## Two shapes had no Rust type at all

This is the first domain where migrating meant *writing* types, not just
generating existing ones — and both gaps were the kind that no gate can see.

**`fleet.group.list` ignored its payload entirely**, so there was nothing to
generate. It needed a real struct rather than `()`, because the stub calls it
with `{}` (`swarm-model.ts:1588`) and serde deserializes `()` **only** from JSON
`null`. A unit `Req` would have rejected every real call at runtime while
compiling and passing every CI gate — the `bookmarks.list` P0 verbatim. An
empty struct generates `Record<string, never>`, which is exactly what the
hand-written stub already declared. The test pins both halves: that the struct
accepts `{}`, and that `()` would not have.

**`fleet.group.delete` returned an inline `json!({ "ok": deleted })`** built in
the handler. An untyped literal like that cannot be checked against the
frontend in either direction. It is now `FleetGroupDeleteResult`, with a test
pinning its serialized shape so it cannot drift from what the frontend reads.

## Everything else matched

All other generated shapes are field-for-field identical to the interfaces they
replace, including the optional fields (`signal`, `staged`, `name`,
`member_ids`), which needed `#[ts(optional)]` — serde treats a missing
`Option<T>` as `None`, so `T?` is accurate and the default `T | null` would
have forced callers to pass an explicit null.

`FleetStagePlan` is re-exported as an alias: the Rust type is `StagePlanInput`
(it is input-only). Renaming the server type to match a frontend spelling would
be the wrong direction, so the alias lives in the stub.

## Tests

Eight cases covering all six commands. Two are worth calling out:

- **bulk-stop parses both with and without its optional fields.** The call site
  is `{ targets, ...opts }` — a spread, so both shapes genuinely occur.
- **an out-of-range `max_fail_percentage` is rejected, not wrapped.** It is a
  `u8` that caps the blast radius of a bulk stop; silently wrapping `300` would
  be a bad failure mode for a destructive operation.

## Verification

- `cargo test -p agentmux-srv req_shape_tests` -- 24 passed (8 new + 16 existing)
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 72 generated types current
- `scripts/check-spec-citations.sh` -- ok (15 files), `check-doc-status.sh` -- ok
- `npm test` -- 91 failing suites, identical to the baseline count on `main`.
  Eight swarm suites are in that list and swarm is exactly what this PR
  touches, so I ran them directly: every one fails with the pre-existing
  Windows `file:///@solid-refresh` jsdom error and reports "no tests" — the
  files cannot load at all in this environment, independent of this change.

agent_id: manoz-0803a

@


a5af:
@codex review

PR #3298, opened 2026-09-17
Lark@narko: docs(spec): general agent-to-agent WAN-tier jekt signing (#2586's unbuilt half)

GenericAgentX-asaf:
## Summary

Specs the **WAN half of #2586** — the one delivery tier with no per-agent sender
verification. `SPEC_JEKT_LAN_TIER_SIGNING_2026_08_15.md` deferred it as blocked on
the account-scoped Cognito prerequisite; this re-checks that premise against
current code and finds it mostly stale.

Spec only. No behaviour change, no implementation.

## What changed since the 2026-08-13 audit (verified, not inferred from Status lines)

| August finding | Today |
|---|---|
| §5.1 per-agent Cognito clients don't scale (100/pool) — needs account-scoped redesign | **Resolved in code** — `agent-ownership.ts` + `AgentOwnershipTable`, binding is now an ownership lookup |
| §5.2 desktop treats only 401 as a credential problem; 403 silently stalls | **Resolved** — `cloud_subscriber.rs:607-671` |
| PLAN header: "desktop per-agent token fetch has not shipped" | **Stale — it shipped** (`agent_credentials.rs`) |
| §1.1 `agent_id` is a flat global namespace at the storage layer | **STILL TRUE** — `store.ts:284-285`, no account attribute in the key condition |

So this is no longer blocked on a redesign. It is blocked on **one migration**
(tenant-scoped injection storage) plus two flag flips (`ENFORCE_AGENT_BINDING`,
`DISABLE_LEGACY_AUTH`) that are set nowhere today.

## Design decisions worth reviewing

- **Supersedes `SPEC_JEKT_LAN_WAN_TRUST_HARDENING_2026_08_13.md` §6.2.** That proposed a
  server-minted shared secret per `(account, agent_id)`. This uses Ed25519 with the private
  half never leaving the sending srv: the cloud holds only public keys, so a muxbus
  compromise can drop or misroute messages but never impersonate an agent. It also deletes
  what would have been the dominant cost line (§5.2 — ~$200/mo in Secrets Manager at
  10 accounts × 50 agents).
- **The ownership registry is the PKI.** LAN needed TOFU pinning because mDNS is
  unauthenticated; WAN has a server-authoritative owner for every `agent_id`. No pin table,
  no pin-mismatch false positives, and keys *can* rotate — which
  `SPEC_JEKT_HOST_KEY_TTL_ROTATION_2026_09_14.md` §4 explicitly could not do for LAN.
- **Separate keypair from LAN, deliberately.** `signed_material` has no domain separator, so
  one shared key would make a captured LAN signature a valid WAN signature for the same
  message. WAN material gets a `wan|v1|` prefix; host/LAN material stays byte-identical, so
  no deployed verifier changes.
- **Tenant scoping is a hard dependency, not a nicety** (§2.1): with a flat namespace, two
  accounts can both own `atlas` and "verify atlas's signature" has no single answer.

## Resource/cost analysis (§5)

Grounded in `quota.ts` and the owner-gate spec's unit-cost derivation, not list pricing:
no new table, no new Cognito clients, no new secrets; incremental cost is sub-cent per
account per month against an existing $0.03–0.08 baseline, inside the $1/user cap. Calls
out the ceilings that *do* bite — Cognito's 100 app clients now bounding **accounts**
rather than agents, the 50/month `agent_provisions` rate limit that key publication rides
on, and the pre-existing O(N) broadcast `Scan`.

One decision flagged for sign-off: **pubkey lookups should not consume `jekt_messages`
quota** — charging them would bill a signed conversation at up to 2× an unsigned one.

## Phasing

W0 enforce binding → W1 scope/retire legacy auth → W2 tenant-scoped storage → **W3 the
signing mechanism itself** → W4 rotation. W0–W2 are pre-existing security debt with
independent value; W3 is the smallest phase. §5.4 says so plainly rather than letting
"WAN signing" absorb the cost of a live multi-tenant DynamoDB migration.

## Open questions (§9)

Scheduling for the W2 migration, legacy-auth retirement strategy, whether to raise the
Cognito ceiling proactively, pubkey-lookup billing, and whether the marker should name the
owning account (privacy question — recommended no).


a5af:
@codex review

GenericAgentX-asaf:
Thanks — verified all eight findings against code before accepting. **All eight hold**, including two that contradicted what I'd written. Pushed in `db59cc75d`.

| # | Finding | Verdict | Where |
|---|---|---|---|
| P1 | Verify on the primary cloud-subscriber path | **Confirmed** — `cloud_subscriber::sync_agent_reactive` builds the `InjectionRequest` and calls `inject_message` directly; reagent verification is duplicated there for exactly this reason. As specified, `TRUST=wan-verified` would never have rendered at all. | §3.4.2 |
| P1 | Preserve the complete signed tuple | **Confirmed** — `cloud_subscriber.rs:948` sets `request_id: Some(inj.id.clone())` (cloud injection id, not the signed msgid) and leaves `ts_secs` unset. Every signature would have failed → active-forgery escalation on legitimate traffic. Now carries `wan_msg_id`/`wan_ts_secs` explicitly, mirroring `reagent_msg_id`/`reagent_ts_secs`, plus a freshness window. | §3.4.1 |
| P1 | Republish independently of provisioning | **Confirmed** — `ensure_agent_credential` (`agent_credentials.rs:88-104`) provisions only when `agent_credential_load` returns nothing. Every already-provisioned agent would mint a key and never publish it. Replaced with `PUT /agents/:id/pubkey` + version sync. | §3.2 |
| P1 | Remove the cloud-forgery guarantee | **Confirmed overclaim, mine.** The cloud serves the directory and can overwrite `wan_public_key`. Replaced with the weaker true claim, plus a continuity chain (new key signed by the old) and last-known-good pinning to make substitution *detectable*. | §3.1, §3.6 |
| P1 | Throttled signed lookups are verification failures | **Confirmed** — shipped LAN maps `LanPubkeyLookup::RateLimited` → `Some(false)` (`reactive.rs:478-489`) with a comment naming this bypass. My "fail closed to unverifiable" was fail-*open*. Worth flagging: `SPEC_JEKT_LAN_TIER_SIGNING_2026_08_15.md` §2.2 still documents the rejected behaviour — I inherited the stale text. | §3.4 |
| P1 | Specify the recipient account in tenant-scoped addressing | **Confirmed** — I'd solved only the sender half (free from the authenticated caller). A bare target name can't reverse-resolve to an account. Added three addressing options; also noted `SPEC_AGENT_NAMING_AND_ADDRESSING_HOST_LAN_WAN_2026_08_22.md` does **not** cover this (Draft, display-naming only, explicitly not routing/security identity). | §2.1.1, Q6 |
| P1 | Scope trusted-peer grants to the sender account | **Confirmed** — `db_conversation_trust_grants` PK is `(agent_id, granted_peer_agent_id, tier)`, no account (`migrations.rs:1024-1030`). Sharpest finding here: signing makes this *worse*, since the attacker's same-named agent verifies under its own registered key and matches the grant, auto-approving a `transcript_request` that would otherwise stop. Moved into W3 rather than W4. | §3.5.1 |
| P2 | Sign before routing | **Confirmed** — MCP posts every send locally and srv resolves the tier afterwards; `sign_outgoing_jekt` already computes `lan_sig` unconditionally and says so. Now unconditional. | §3.3 |

Also revised the W3 effort estimate from "a few focused days" to one to two weeks — it had assumed LAN parity, which no longer holds.

PR #3299, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the 8 bundle CRUD commands, generate Bundle itself

GenericAgentX-asaf:
Seventh domain. **Split deliberately**: the three bundle import/validate
commands need their private `Req` structs promoted out of the handler file
first and will follow in a second PR. This one is already the largest of the
series and I did not want to make it unreviewable.

## `Bundle` was both the request and the response type, and could not correctly be either

This is the most interesting thing the migration has turned up so far.

**As a response, it is all-required.** Every field is `#[serde(default)]` but
none is `skip_serializing_if`, so the server always writes all sixteen. The
hand-written global marked twelve of them optional — a shape the wire never
produced.

**As a request, only `id` and `name` are required**; everything else defaults
server-side. The stub typed this `Partial<Bundle>`, which told callers `name`
was omittable. serde rejects a payload with no `name`, so any caller taking the
type at its word would fail at runtime. Same class as the `bookmarks.list` P0,
found the same way, and the new test pins it:

```rust
assert!(
    serde_json::from_value::<Bundle>(json!({"id": "b1", "description": "d"})).is_err(),
    "a payload with no name must be rejected -- Partial<Bundle> wrongly allowed it"
);
```

The fix is one generated type plus a **derivation**, not two hand-maintained
shapes:

```ts
export type BundleUpsertInput =
    Pick<Bundle, "id" | "name"> & Partial<Omit<Bundle, "id" | "name">>;
```

Derived from the generated `Bundle`, so a new Rust field flows through
automatically and this cannot drift. That felt worth more than hand-listing
fifteen optional fields.

## Reused rather than duplicated

`getclaudeglobalconfig` needed a named result type. I drafted one, then found
the handler file already had a private `ClaudeGlobalConfig` struct with exactly
that shape — so I deleted my draft and promoted the original instead. Adding a
second copy would have been the duplication this migration exists to remove.

Its `content` is generated as `string | null`, deliberately **not** optional: a
plain `Option<String>` with no `skip_serializing_if`, so the key is always
present and carries null. `exists: false` with `content: null` is the
meaningful "no file" answer, not an absent key.

Two inline `json!` literals became named result types with serialization tests.

## Fallout from making Bundle all-required: six sites, all of them wrong before

- **`draftToWire`** builds a *request* but was typed `Bundle`; it is now
  `BundleUpsertInput`. Its doc comment justified sending `0` for
  `created_at`/`updated_at` by claiming the Rust struct "has no serde defaults"
  for them. It does have them. The behaviour was right, the stated reason was
  not — corrected in place.
- **Five test fixtures** constructed partial bundles, i.e. asserted a response
  shape the server cannot produce. They now use a complete factory / a spread
  base (two edits, not five — both files already had a local factory).

## Verification

- `cargo test -p agentmux-srv req_shape_tests` -- 29 passed (5 new + 24 existing)
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 79 generated types current
- `scripts/check-spec-citations.sh` -- ok (24 files), `check-doc-status.sh` -- ok
- `npm test` -- 91 failing suites, identical to the baseline count on `main`.

**One caveat I want to be explicit about:** four of the suites I edited
(`bundle-model`, `global-bundle-model`, `AgentLaunchModal.integration`,
`AgentStartupModal`) are in that pre-existing failing list. They fail with the
Windows-only `file:///@solid-refresh` loader error and report "no tests" — they
cannot load at all on this machine, so **my local run did not exercise the
fixture changes**. The `vitest` CI job runs on `ubuntu-latest`, where that bug
does not occur, so CI is the real check on them here, not my local run.

agent_id: manoz-0803a


a5af:
@codex review

PR #3300, opened 2026-09-17
fix(agent-pane): stop window tabs from flashing on creation and close

AgentA-asaf:
## Summary

Two related fixes to the window-tab flashing reported during active work on the recent tab-switch perf rework (#3239) — both root-caused and documented in full before implementation:

- **Tab creation** (`docs/specs/SPEC_TAB_CREATION_REVEAL_ARCHITECTURE_2026_09_16.md`): `createTab()` used to activate the new tab as part of the `CreateTab` RPC itself, then race the reveal gate's frame-heuristic settle-detector against `applyTabPreset`'s own in-flight pane-creation RPCs — confirmed at the backend that activation happens eagerly, well before any panes exist. The gate reliably lifted early, revealing a near-empty tab that filled in piecemeal, and could re-trigger once the intentionally-deferred lift fired later. Now creates inactive, populates fully in the background (current tab untouched), and activates via the existing, unmodified `setActiveTab()` only once real content exists. Also retires the reveal gate's "untargeted hold" special case, which existed solely to serve the old flow.

- **Tab close timing** (`docs/specs/SPEC_TAB_CLOSE_BUTTON_SELECT_FLASH_2026_08_25.md` §10): the confirm modal hid the tab from the strip the instant it opened — a deliberate, explicit repo-owner directive from 2026-08-25 aimed at a different flash class entirely (see §8.1). In practice that read as its own flash: the tab vanishing before the user had even decided, then reappearing on Cancel. The tab now stays fully visible while the modal is open; confirming hides it and closes the modal in the same synchronous click, so both disappear together, in one frame. Cancel no longer needs to restore anything, since nothing was hidden.

## Known residual (not fixed by this PR)

A flash is still reported after *confirming* a close, distinct from the timing issue above — candidate mechanisms are named in §10.4 of the close-flash spec but not yet confirmed, and are being investigated as a tracked follow-up rather than guessed at here.

## Test plan

- [x] `tsc --noEmit -p tsconfig.citypecheck.json` clean
- [x] All 21 `tab-reveal.test.ts` cases pass (updated for the retired untargeted-gate case)
- [x] Live `task dev` verification of both fixes — confirmed real improvements
- [ ] CI


a5af:
@codex review

PR #3301, opened 2026-09-17
Nark@Area54: fix(lan-discovery): key peer staleness off each peer's own mDNS TTL

GenericAgentX-asaf:
## Summary

`GET /agentmux/discovery` (and therefore `DiscoverAgents`/`SendMessage` LAN
routing, the status-bar LAN indicator, and Warden's peer table) was silently
dropping healthy LAN peers for most of every ~40+ minute gap between their
mDNS re-announcements.

PR #3245 (`831a57149`, merged 2026-09-15) fixed a real bug — `ServiceRemoved`
firing on ordinary TTL churn was wiping peer identity — by adding a
`get_instances()` staleness cutoff. The cutoff it picked, `300s` fixed for
every peer, was tuned against the one peer visible in that debugging
session (`gamerlove`, which happens to re-announce every 1-2 minutes). Two
other real peers on the same LAN (`narko`, `starpower`) re-announce roughly
every 40-46 minutes instead — well within the mDNS-standard 75-minute
PTR/TXT TTL (RFC 6762 §10; `mdns-sd` 0.12 hardcodes this as `DNS_OTHER_TTL =
4500`), but 9x past the fixed cutoff, so they flickered in and out on every
cycle.

## Fix

`LanInstance` now carries `other_ttl_secs`, captured from
`ServiceInfo::get_other_ttl()` on every `ServiceResolved` (unconditionally —
it's a DNS record TTL, not TXT payload content, so it survives the
blank-TXT re-resolutions this file already works around elsewhere).
Staleness is now `peer_staleness_window_secs(other_ttl_secs)` — each peer's
own advertised TTL, floored at the old 300s value only as a guard against an
implausibly small advertised TTL — instead of one constant shared by every
peer regardless of its real re-announce cadence.

Full investigation, live-log evidence across three real hosts, and why a
simple "raise the constant" fix was rejected:
`docs/retro/retro-lan-peer-stale-timeout-too-short-for-slow-reannounce-peers-2026-09-17.md`

## Test plan

- [x] `cargo test -p agentmux-srv --bin agentmux-srv lan_discovery` — 46 passed, 0 failed, 1 pre-existing-ignored (environment-fragile mDNS test, unrelated)
- [x] New regression test `get_instances_does_not_prune_a_slow_reannouncing_peer_within_its_ttl` directly encodes the real-world bug (peer silent 46 minutes, must still be reported)
- [x] Updated `get_instances_hides_a_peer_stale_past_its_own_ttl` to assert staleness is keyed on the peer's own TTL, not the floor


a5af:
@codex review

PR #3302, opened 2026-09-17
Opaz@charlie: fix(shell): deploy shell integration per instance, not machine-global

GenericAgentX-asaf:
## Summary

Shell-integration scripts — the rcfiles **every terminal sources**, plus `muxlog/muxspect/muxopen/muxsh` — deployed to `~/.agentmux/shell/`: one directory shared by every instance of every version on the machine. Redeployment is gated on a `<version>-<content hash>` marker, so two concurrently-running versions each overwrite the other's scripts. And `deploy_scripts` runs not just at boot but on **every interactive shell spawn**, so it never settles.

Violates **I6** (*instances of different `(channel, version)` never share a directory*). Data dirs are carefully isolated per build; the scripts every terminal sources at startup then silently were not.

## Demonstrated live

Watching `~/.agentmux/shell/.version`:

| Time | Marker | Trigger |
|---|---|---|
| 20:37:20 | `0.56.3-94ef97c9…` | a 0.56.3 instance booted |
| 23:49:19 | `0.56.2-a8a36d03…` | a 0.56.2 instance booted **in a separate channel** |
| 23:49:46 | `0.56.3-94ef97c9…` | **one terminal opened** in the 0.56.3 instance flipped it back |

Also permits torn reads (`fs::write` truncates first) and protocol skew — the `.mjs` helpers talk to a backend over its API, so a 0.56.2 instance can run 0.56.3's `muxlog.mjs`.

## Why not just move it under the data dir

Two constraints, both preserved:

1. **The global path is deliberate.** MSIX virtualises writes to `%LocalAppData%`, so scripts written under the data dir are invisible to ConPTY-spawned children; `~` never is. Moving it back would reintroduce that bug.
2. **`~/.agentmux/shell/muxspect.mjs` is documented** (`docs/MUXSPECT.md`, `docs/MUXSH.md`) and must stay *predictable* — a user cannot know their instance's hash.

So: keep the home base, **key the subdirectory**, mirroring what I5 already requires of named OS objects. Shells source `~/.agentmux/shell/<key>/`, where `<key>` is a stable FNV-1a hash of the instance data dir (`<channel>/versions/<version>/data` — exactly the pair I6 names). The documented root is still deployed for direct invocation; nothing sources it at startup. `deploy_scripts` and `get_shell_startup` both take the shell root directly and both call sites get it from one helper, so they cannot drift.

## Recorded, not papered over

`integration_base()` resolves via `get_mux_data_dir()` → `AGENTMUX_DATA_HOME`, which bootstrap overwrites from the resolved config — but **before** that runs it can still hold a value *inherited from the instance that spawned this process*. Observed live: a dev srv launched from a terminal inside a 0.56.2 instance carried that instance's data dir. Both call sites run after bootstrap; the doc comment says not to move them earlier.

## Test plan

- [x] `two_instances_never_share_a_shell_integration_dir` — falsified: against the pre-fix path it fails with `left`/`right` both `"/home/u/.agentmux"`.
- [x] `the_deployed_rcfile_is_exactly_the_one_the_shell_is_told_to_load` — **the coupling that matters**: resolves the `--rcfile` arg to a real file on disk. If deploy and spawn drift, scripts land where nothing sources them and every terminal silently loses integration with no error anywhere. Falsified by introducing drift. Note the pre-existing `test_bash_startup_args` does **not** catch it (only asserts `contains("bash")`), so it passed both before and after the path-contract change.
- [x] `the_same_instance_always_resolves_to_the_same_dir`, `deployment_stays_under_the_never_virtualised_home_dir`
- [x] agentmux-srv shellintegration suite: **11 passed / 0 failed**
- [x] Live on a dev instance: shell spawns with `--rcfile ~/.agentmux/shell/31207cd1ecc7f419/bash/.bashrc` while unfixed instances keep using the global path; a terminal driven with **real synthetic keystrokes** ran vim end to end (insert → write → quit) with the file landing on disk. Repo owner independently confirmed vim works in that instance.

## Out of scope

- **Stale-directory sweep** — keyed dirs accumulate one per `(channel, version)` ever run (~100 KB each). Bounded but unbounded over time; needs a liveness rule, which is its own decision.
- **The documented root still flip-flops** between versions. It is only reached by explicit manual invocation, never by shell startup.
- **Does not fix the vim/terminal-lock report** it was found while investigating (included here as `docs/reports/REPORT_VIM_PANE_LOCK_0562_VS_0563_2026_09_16.md`). The deployed `.bashrc` defines on-demand functions and a prompt-time OSC 7 hook, so there is no mid-alternate-screen write path. Fixed on its own merits.


a5af:
@codex review

PR #3304, opened 2026-09-17
Lark@narko: feat(jekt): mint and sign per-agent WAN keys (W3a of #2586's WAN half)

GenericAgentX-asaf:
## Summary

First implementation increment of `SPEC_JEKT_WAN_TIER_SIGNING_2026_09_17.md` (merged in #3298) — issue **#2586's unbuilt WAN half**.

Ships **key material and client-side signing only**. No verifier, no `TRUST=wan-verified` marker value, no tier-escalation change, no cloud change, no schema migration on muxbus. Behaviour is unchanged for every delivery tier: an agent mints a WAN keypair at spawn and attaches a `wan_sig` that nothing currently reads.

## Why signing ships before verification

An agent only receives its signing key **when it is spawned**. A verifier landing before keys have propagated would apply to almost no live agent on day one — which is exactly what `REPORT_JEKT_SIGNING_KEY_INJECTION_GAP_2026_08_16.md` documents for host-tier signing, where ordinarily-launched agents rendered `TRUST=self-declared` for months after the feature shipped. Minting now starts that clock.

## Why verification cannot land yet

Two hard blockers, both verified against current code rather than assumed:

1. **Key resolution is ambiguous.** Verifying means resolving `(sender_account, source_agent)` to exactly one published key. `getPendingInjections` still queries a bare `target_agent` with no account scoping (`agentmux-cloud` `store.ts:284-285`), so one agent name can belong to several accounts — spec §2.1, phase W2.
2. **The signed tuple does not survive the cloud.** `cloud_subscriber` sets `request_id` to the cloud's own injection id and leaves `ts_secs` unset, so a signature would verify against different material than was signed and **fail** — rendering active-forgery escalation on legitimate traffic. Fixing that means relaying the signed msgid/timestamp explicitly, as the `reagent_*` fields already do (spec §3.4.1).

Deliberately **no `wan_verified` field yet**: a verification-outcome field that nothing ever sets reads as wired when it isn't.

## Design points worth review

- **Domain separation is load-bearing.** `WAN_DOMAIN` prefixes the signed material. Without it, a signature captured off the LAN would be a byte-identical valid WAN signature — and `TRUST=wan-verified` is a stronger claim, checked against a different key source (cloud registry vs. the peer's own answer). Tested in both directions.
- **A separate keypair from LAN, by necessity rather than style.** LAN public keys are permanently TOFU-pinned by peers (`db_lan_peer_pubkey_pins`), which is precisely why `SPEC_JEKT_HOST_KEY_TTL_ROTATION_2026_09_14.md` §4 declined to rotate them; WAN keys are registry-served and are meant to rotate. One keypair cannot have both lifecycles. `key_version` is stored now, unread, so adopting rotation needs no second schema change.
- **Signing is unconditional**, like `lan_sig`/`channel_sig`: the MCP process posts every send to the local reactive endpoint and srv picks the tier afterwards, so a tier-conditional signature has no implementable call site.
- **Key injection goes through the shared path** (`inject_jekt_signing_keys_into_mcp_json`), so `agent.open` and the Launch button can't drift apart the way they did before that function existed. Assertions added to both call sites' tests.

## Testing

- 5 new `jekt_sign` tests, including cross-tier replay rejection in both directions
- 8 new `agent_wan_keys` storage tests, including one asserting an agent's WAN and LAN keypairs are independent
- WAN assertions added to the two existing key-injection tests
- **Full `agentmux-srv` suite green: 3633 passed, 0 failed**

## Not in this PR

W0 (`ENFORCE_AGENT_BINDING`), W1 (legacy auth), W2 (tenant-scoped storage + the addressing decision), and the rest of W3 (publish/fetch, verification, marker, account-scoped trust grants). W0–W2 are live-production changes against paying customers and carry two open decisions that are the repo owner's: the addressing form (spec §9.6) and who runs the W2 migration (§9.1).


a5af:
@codex review

PR #3305, opened 2026-09-17
Manoz@Area54: refactor(rpc): generate the bundle validation report types

GenericAgentX-asaf:
Follow-up to #3299, which split the bundle domain. This covers `bundle.validate`;
the two import commands need response structs authored from scratch and follow
separately.

## The main thing to review: this deliberately does NOT migrate the handler

`bundle_validate_impl` runs its payload through `normalize_bundle_upsert_input`
**before** deserializing it. That step is load-bearing twice over:

- it fills a missing/null `id` with `""`, which is what lets an unsaved draft
  validate at all (the reagent P2 on #3299 was exactly this), and
- it coerces array-valued `context_files` / `mcp_servers` / `skills` into the
  JSON-encoded strings `Bundle` expects.

`register_typed` deserializes the payload into `Req` first and hands the handler
an already-typed value, so **there is no point at which normalize could run**.
Typing the request would mean re-implementing that coercion as custom serde
deserializers — duplicating security-relevant logic to satisfy the generator.
That is the wrong trade, so the request stays a `Value`, with a comment on
`register_bundle_validate` saying why.

The **response** is where the drift risk actually lived, and that is now
generated: `ValidationReport`, `ValidationIssue`, `IssueSeverity`.
`bundle_validate_impl` returns the concrete report instead of
`serde_json::Value`, so the handler is the only place that serializes.

`IssueSeverity` generates as `"error" | "warning"` — identical to the
hand-written union, but as a named type rather than a string literal repeated in
a comment.

I am flagging this rather than burying it because "migrate every handler to
`register_typed`" is the obvious reading of the spec's step 2, and this is the
first command where following that mechanically would have broken behaviour.
Not every handler should be typed.

## Noted for the import PR, not fixed here

While surveying, I found that `bundle.import.preview` sends a
`project_instructions` field that **neither the hand-written
`BundleImportPreviewResponse` declares nor any frontend code reads**. It is
shipped on the wire and invisible to the client. I have not touched it — that
belongs with the import work, where generating the response type will surface
it properly rather than having me quietly add or remove a field here.

Also for that PR: the stub types `bundle.import.preview` as
`{ file_path: string }`, but the Rust `PreviewReq` accepts three *optional*
alternatives (`file_path`, `zip_base64`, `files`). Two whole input modes are
currently unreachable through the typed client.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3641 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 82 generated types current
- `scripts/check-spec-citations.sh` -- ok (11 files), `check-doc-status.sh` -- ok

One existing test indexed into the returned `Value` (`report["is_valid"]`) and
now reads the typed fields directly.

agent_id: manoz-0803a


a5af:
@codex review

PR #3306, opened 2026-09-17
Lark@narko: feat(jekt): bind WAN signatures to the sending instance (host + channel)

GenericAgentX-asaf:
## Summary

Implements the decision recorded in `SPEC_JEKT_WAN_TIER_SIGNING_2026_09_17.md` §2.1.2 (spec change: `83f814283`): **an agent's WAN identity is `(account, host, channel, agent_id)`, not `(account, agent_id)`.**

This **fixes a defect in W3a (#3304)**, it isn't a refinement.

## The defect

A WAN public key was to be published per `(account_user_id, agent_id)`. But every srv instance has its own SQLite database, so each mints its **own** keypair for a given agent name.

One account running an agent named `lark` on two machines would publish two different public keys to **one row**. Last write wins — and the losing host's legitimate signatures then verify against the wrong key, producing `wan_verified = Some(false)`, which this spec defines as **active forgery**: forced `TIER=sensitive` / `ESCALATE=required`, permanently, on traffic that is exactly who it claims to be.

Not hypothetical for this deployment — agents already run under one account across `narko`, `Area54`, `starpower`, `gamerlove` and `claudius`. It's the same class of self-inflicted false positive that `SPEC_JEKT_HOST_KEY_TTL_ROTATION_2026_09_14.md` §4 refused to inflict on LAN by rotating pinned keys, arriving by a different route.

## Granularity: instance, not machine

One box can run a `stable` install alongside per-build `local-<branch>-<hash>-<build-id>` channels, each with its own data dir and database — so **hostname alone leaves the identical overwrite race between two channels on one machine.** The unit that owns a keypair is the instance: `(hostname, channel)`.

## Why the material changes, and why now

The signed material gains `source_host` and `source_channel`, following `sign_channel_jekt`'s existing precedent (it binds `source_channel` so a signature minted in channel A can't be replayed as channel B). The instance selects *which published key a receiver checks against*, so it has to be inside the signature — otherwise a mislabelled instance silently selects a different key instead of failing.

**Changing material already shipped in W3a is free right now, precisely because nothing verifies WAN signatures yet** — the property that shipping the mechanism inert was meant to buy. After a verifier exists this becomes a flag day.

The **account is deliberately not bound** into the material: it's a cloud-side fact the signing process has no trustworthy view of, and binding a client-asserted account would let a sender choose which account's key it's checked against. It stays server-resolved from the authenticated caller. Host and channel are different — both are injected at spawn, and a wrong value selects a key the signature cannot verify under, so it grants nothing.

## Changes

- `registry::local_host_label()` — lowercased/trimmed, defaults to `"unknown"` exactly as `bootstrap.rs` already does for the LAN advertisement, so sender and publisher can't disagree by case alone. A normalization mismatch here doesn't fail loudly; it selects a different key.
- `AGENTMUX_HOST_LABEL` injected at spawn beside `AGENTMUX_CHANNEL`, written explicitly rather than read from ambient env — for the same reason the channel is (a nested or scrubbed env would otherwise sign under a different instance identity than srv publishes under).
- `wan_source_host` on the wire; `source_channel` now ships when **either** `channel_sig` or `wan_sig` exists, since both now bind it.

## Testing

- New `a_wan_signature_is_bound_to_the_sending_instance` — covers both same-agent/different-machine and same-machine/different-channel
- Host-label assertions added to both key-injection tests (the path that silently drifted once before, per `REPORT_JEKT_SIGNING_KEY_INJECTION_GAP_2026_08_16.md`)
- `agentmux-srv` **3646 passed**, `agentmux-mcp` 26 passed, `jekt_sign` 49 passed


a5af:
@codex review

PR #3307, opened 2026-09-17
Manoz@Area54: refactor(rpc): type the bundle import preview/commit wire shapes

GenericAgentX-asaf:
Completes the bundle domain, after #3299 (CRUD) and #3305 (validation).

Unlike every prior domain, this was an **authoring** job rather than a
generation one. Both handlers built their responses as ad-hoc `json!` literals
and both request structs were `private` to the handler file — so there was
nothing for the bindings generator to read, and the frontend's matching
declarations were hand-maintained against nothing at all. Eleven types now
exist in `rpc_types/bundle_import.rs`.

## Two pieces of real drift fell out of that

**1. `project_instructions` has been shipping invisibly.** The preview response
has always included it, but the hand-written `BundleImportPreviewResponse` did
not declare it and no frontend code reads it. The handler's own comment says
those entries "exist to be looked at", so this is unfinished UI wiring, not
dead weight.

I surfaced it rather than silently dropping it — whether the UI should use it
is a product decision, not something to settle inside a refactor. The one
frontend fixture that omitted it now sets `project_instructions: []`, which is
the shape the server actually sends.

**2. Two of three input modes were unreachable.** The stub declared
`bundle.import.preview` as `{ file_path: string }`, but `resolve_import_input`
accepts three *alternatives*: a `file_path` on disk, a base64 zip, or an
already-unpacked `files` list. The typed client could only express one of them,
despite the server supporting all three since Phase 3. The generated request
type now has all three as optional alternatives, with a test asserting each
parses on its own.

## Deduplication, not just typing

`FileEntry` and `SkillSelection` were private structs in the handler file whose
doc comment already claimed `FileEntry` was "shared by `bundle.import`'s `files`
input and the Phase 3 handlers" — it just wasn't, it was a second copy. Both are
now the promoted types, so the handler file and the generated bindings cannot
disagree about them.

## Why I trust the conversion

The existing import tests asserted the response **field-by-field, by name**
(`resp["skills"][0]["slug"]`, `resp["content_digest"]`, and ~25 more). That made
them a genuine safety net for rewriting ~140 lines of `json!` into struct
construction: any field I renamed, dropped, or mistyped would have failed them.
They now read typed fields and all still pass, which is the actual evidence that
the authored structs match the old literals exactly.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3667 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 95 generated types current
- `scripts/check-spec-citations.sh` -- ok (24 files), `check-doc-status.sh` -- ok

Same caveat as #3299: the one suite I edited
(`BundleImportPreviewModal.test.tsx`) is in that pre-existing failing list — it
hits the Windows-only `file:///@solid-refresh` loader error and reports "no
tests", so it did not run locally. The `vitest` CI job on `ubuntu-latest` is the
real check on that fixture change.

agent_id: manoz-0803a


a5af:
@codex review

PR #3308, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the 4 blockfile commands, survey the rest of block

GenericAgentX-asaf:
First slice of the `block` domain, which is the largest so far: **26 commands
and 44 structs**. `app_api/blockfile.rs` is self-contained, so it goes first and
lets the approach be checked on something small.

All eight generated types match the declarations they replace **exactly** — no
drift in this slice.

## Two response fields, two different shapes, side by side

Worth calling out because it would be easy to assume they behave alike, and the
tests pin both:

| Field | Rust | Wire | Generated |
|---|---|---|---|
| `BlockfileReadRangeResult.stamps` | `Option<Vec<i64>>` **with** `skip_serializing_if` | key **omitted** when absent | `stamps?: number[]` |
| `BlockfileReadStateResult.content` | plain `Option<String>`, **no** `skip_serializing_if` | key always present, carries `null` | `content: string \| null` |

The `stamps` doc comment promises "old frontends ignore it, new frontends
tolerate absence" — a promise that only holds while it is genuinely omitted
rather than nulled, so there is now a test asserting the serialized form of
both.

## Survey note on the other 22 commands

I scanned all 44 structs before slicing. **40 are clean**; four are not, and
they are *not* migratable as-is:

`CommandCreateBlockData` is the clearest case, and it is the first type in this
whole migration where the **hand-written TypeScript is richer than the Rust**:

```
// frontend                          // rust
blockdef: BlockDef;                  blockdef: Option<serde_json::Value>
rtopts?: RuntimeOpts;                rtopts:   Option<serde_json::Value>
```

Generating from Rust would emit `unknown` for both and destroy real frontend
typing that has no server-side counterpart today. `BlockDef` and `RuntimeOpts`
*do* exist in `backend/obj.rs`, so the proper fix is to tighten the Rust field
types — but that is a **behaviour change** (stricter deserialization at the RPC
boundary; a payload that currently passes through opaquely would start being
rejected), not a refactor. It belongs in its own change rather than smuggled
into this one.

The other three (`CommandControllerResyncData`, `CommandBlockInputData`,
`CommandAgentAnswerData`) have `skip_serializing_if` on **non-`Option`** fields
(`bool`, `String`), which is the shape ts-rs cannot express as an optional
property. Generating them would mark genuinely-omittable fields **required** —
a strictly worse type than what is there now.

That is the first time in nine domains that migrating part of a domain would
make things worse, so I am flagging it rather than quietly doing 22 of 26 and
calling the domain done.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3681 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 103 generated types current
- `scripts/check-spec-citations.sh` -- ok (13 files), `check-doc-status.sh` -- ok
- `npm test` -- 91 failing suites, identical to the baseline on `main`, and no
  suite touched by this PR is among them

agent_id: manoz-0803a


a5af:
@codex review

PR #3310, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate 8 websocket-hosted block commands

GenericAgentX-asaf:
Second slice of the `block` domain, after #3308 (blockfile). Covers the eight
websocket-hosted commands: `deletesubblock`, `tooldecision`, `ambientnarrate`,
`docknodestatus`, `backgroundtaskcompletion`, `backgroundtaskpid`,
`listbackgroundtasks`, `agentcancel`.

## I rebuilt the expressibility scan, because the first one was wrong

The survey in #3308 said "40 of 44 structs are clean". That scan only looked at
fields declared **directly in `block.rs`** and never followed nested types into
other modules, so it was measuring the wrong thing.

The transitive version immediately found that **`CommandCreateSubBlockData` is
not migratable after all**: it embeds `BlockDef` from `obj.rs`, and both
`BlockDef.meta` and `FileDef.content` carry `skip_serializing_if` on
**non-`Option`** fields. Generating it would mark genuinely-omittable fields
required. So it joins the other four carve-outs and this slice is 8 commands,
not 9.

Flagging the bad scan explicitly because its number is quoted in #3308's
description and I would rather correct it here than leave it standing.

## Four of nine types needed correcting before deletion — two were my own errors

The diff-before-delete step earned its keep again:

- **I applied `#[ts(type = "number | null")]` to every `timestamp` with a blunt
  replace.** Wrong: these are *request* types, serde treats a missing `Option`
  field as `None`, and the hand-written declarations correctly said
  `timestamp?`. Corrected to `#[ts(optional, type = "number")]` — and note
  `optional` *alone* silently reverted them to `bigint`, which the re-diff
  caught.
- **`CommandToolDecisionData.outcome`/`scope` and `BackgroundTaskView.status`**
  are Rust strings carrying closed sets (`"allow" | "deny"`,
  `"once" | "session" | "project" | "global"`,
  `"running" | "done" | "error" | "stopped"`). ts-rs emits a bare `string`,
  which would have dropped unions the frontend actually branches on. Preserved
  explicitly — the same call already made for session `verdict` and
  bundle-import `collision`.

All nine types match their hand-written declarations exactly after those fixes.

## Wire note

The seven void handlers now put `"data": null` where the old untyped `Ok(None)`
omitted the key entirely. I checked all seven: **zero non-discarding call
sites**, so this is unobservable — same reasoning as the `native-memory`
`write_file` case in #3295.

One test deliberately pins something that might read as a gap:
`tool_decision_parses_any_outcome_string_and_leaves_validation_to_the_handler`.
An unknown `outcome` still *deserializes*; the handler rejects it with a domain
error. The TypeScript union is a caller-side aid, not the guard, and that
distinction is worth being explicit about.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3695 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 112 generated types current
- `scripts/check-spec-citations.sh` -- ok (21 files), `check-doc-status.sh` -- ok
- `npm test` -- 91 failing suites, identical to the baseline on `main`.
  `useBackgroundTaskRegistry.test.ts` is among them and I edited it, so I ran it
  alone: pre-existing Windows `file:///@solid-refresh` loader error, "no tests".
  CI's Linux `vitest` job is the real check on that file.

agent_id: manoz-0803a


a5af:
@codex review

PR #3311, opened 2026-09-17
Manoz@Area54: refactor(rpc): delete 10 dead block stubs; 25% of the stub surface is unregistered

GenericAgentX-asaf:
Third slice of the `block` domain — but this one **deletes** rather than
migrates, and the reason generalises well beyond `block`.

## 25% of the whole RPC stub surface is declared-but-unregistered

While mapping `block`'s remaining commands I found that ten of them have **no
server handler at all**. That is not a discovery — the repo already tracks it in
`test/contract/rpc-contract.test.ts`'s `KNOWN_DECLARED_UNREGISTERED` allowlist,
which the contract test asserts the diff matches exactly. I nearly reported it
as a find before reading that test; it is a documented, deliberately-tracked
condition.

What is worth reporting is the **scale**, which I measured across all 16 stub
files:

| stub | commands | declared-but-unregistered |
|---|---|---|
| `file.ts` | 42 | **23** |
| `workspace.ts` | 39 | **22** |
| `misc.ts` | 22 | **14** |
| `block.ts` | 26 | **10** |
| the other 12 | 148 | 0 |
| **total** | **277** | **69 (25%)** |

The dead surface is concentrated in exactly four domains, three of which I have
not migrated yet. `file.ts` and `workspace.ts` are the two largest remaining
stubs by line count and are **>50% dead**. That materially changes the shape of
the work left: for those domains the right first move is deletion, not codegen.

## What this PR does

Deletes the ten dead `block.ts` stub methods and removes them from the
allowlist. The contract test asserts the declared-vs-registered diff equals the
allowlist **exactly**, so it verifies this in both directions — deleting a stub
that *was* registered, or forgetting an allowlist entry, both fail it. It
passes.

Also drops five global types in `srv-types.d.ts` that the deletion orphaned
(`BlockInfoData`, `CommandDisposeData`, `CommandBlockSetViewData`,
`CommandControllerAppendOutputData`, `CommandCreateBlockData`) and corrects a
stub comment that pointed at one of them. `srv-types.d.ts` is now **2,223
lines**, down from 2,755 when this migration started.

## One thing I nearly got wrong

`captureblockscreenshot` looked identically dead, but it is a **reverse RPC**:
the frontend *implements* it (`handle_captureblockscreenshot` in
`tabrpcclient.ts`) and the backend calls it. The outbound stub genuinely is dead
and is deleted, but `CommandCaptureBlockScreenshotData` is **kept**, because the
inbound handler still needs it.

I checked the other nine against `tabrpcclient.ts` for the same pattern — none
of them match, so they are dead in both directions. "No server handler" is not
sufficient evidence that a command is dead; the reverse direction has to be
checked too.

## Knock-on: one carve-out evaporates

#3310 listed `CommandCreateBlockData` as a carve-out — the case where the
hand-written TypeScript was *richer* than the Rust (`blockdef: BlockDef` vs
`Option<serde_json::Value>`). With `createblock` deleted, that type has no live
consumer and the carve-out is moot. The remaining `block` carve-outs are four,
not five: `controllerinput`, `controllerresync`, `agentanswer`, `createsubblock`.

## Verification

- `npx tsc --noEmit` -- clean
- `test/contract/rpc-contract.test.ts` -- 4 passed (this is the real check here)

agent_id: manoz-0803a


a5af:
@codex review

PR #3312, opened 2026-09-17
Manoz@Area54: refactor(rpc): delete 55 dead stubs across misc, workspace and file

GenericAgentX-asaf:
Follow-up to #3311, which found that 25% of the RPC stub surface is
declared-but-unregistered and deleted `block.ts`'s share. This sweeps the other
three affected domains: **55 stub methods** across `misc.ts` (14),
`workspace.ts` (13) and `file.ts` (28).

One PR rather than three because it is a single mechanical change with a single
verification — the contract test asserts the declared-vs-registered diff equals
the allowlist **exactly**, so a wrongly-deleted stub or a forgotten allowlist
entry fails it in either direction. Splitting would add CI round-trips without
adding safety.

## Deleted only what is dead on every axis

A command was deleted only if all of these hold:

1. in `KNOWN_DECLARED_UNREGISTERED` (no backend handler),
2. **not** in `KNOWN_LIVE_UNREGISTERED` (the frontend does not call it),
3. no `RpcApi.<Method>(` call site anywhere in `frontend/`,
4. not implemented as a reverse RPC in `tabrpcclient.ts`.

Condition 4 exists because of the `captureblockscreenshot` near-miss in #3311.
Condition 2 matters here: **11 unregistered commands are still actively
called** — `connconnect`, `connlist`, `workspacelist`, `resolveids`,
`fileappend`, `filejoin` and friends. Those fail at run time today. They are
pre-existing, deliberately tracked, and explicitly **left alone** — fixing them
means either wiring a handler or removing a caller, which is product work, not
a refactor.

## A bug in my own scanner, caught before it deleted the wrong thing

My first pass paired methods to commands with a lazy
`([A-Z]\w*Command)[\s\S]*?rpcCall(?:Stream)?\("([^"]+)"` regex. That matches
`rpcCall` and `rpcCallStream` but **not** `rpcStream`, so stream-only methods
were skipped and silently paired with the *next* method's command. It reported
`StreamCpuDataCommand (termgetscrollbacklines)` — and had I trusted it, I would
have deleted a method that calls a live, registered command while believing it
was something else.

The extractor in `rpc-contract.test.ts` documents this exact trap at lines
136–144 ("a single lazy `…*?rpcCall` regex would run past a `rpcStream`-only
method into the next one… a guard that passes while drift slips through"). I hit
it independently before reading that comment.

## The stream canary had to be generalised, not deleted

`extractor sanity` pinned `declared.has("fileliststream")` as a canary that the
per-method extraction does not drop streams. It turned out **every** `rpcStream`
binding in the codebase was a dead stub, so after this sweep there are none left
and the canary had nothing to point at.

Deleting the assertion would have quietly removed a real guard. Instead it now
asserts over *whatever* stream bindings exist:

```ts
for (const cmd of streamCmds) {
    expect(declared.has(cmd), `stream binding ${cmd} missing…`).toBe(true);
}
```

That holds vacuously at zero and restores itself the moment a stream binding
comes back. The `declared.size` floor also moved 180 → 150, since it was
calibrated against the pre-deletion surface; it is a sanity floor against a
broken extractor, not a target.

## Result

| | before | after |
|---|---|---|
| declared commands | 277 | 221 |
| declared-but-unregistered | 69 | 14 |
| `srv-types.d.ts` | 2,755 (at migration start) | 2,222 |

The 14 remaining are the actively-called ones described above, plus
`captureblockscreenshot`.

## Verification

- `npx tsc --noEmit` -- clean
- `test/contract/rpc-contract.test.ts` -- 4 passed (the real check here)
- `cargo check --workspace --tests` -- clean (no Rust touched)

agent_id: manoz-0803a


a5af:
@codex review

PR #3313, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the muxbus and providers.models commands

GenericAgentX-asaf:
The `misc.ts` domain, after #3312 removed its 14 dead stubs. Promotes six
private handler-file structs (`MuxBusLoginReq`/`Resp`, `MuxBusLoginCancelResp`,
`MuxBusStatusResp`, `ProvidersModelsParams`/`Result`) plus `CatalogModel`, so
the frontend's inline copies are generated rather than hand-maintained against
types it could not see.

Adds four new types: empty request structs for
`muxbus.login.cancel`/`status`/`disconnect` — all three are called with `{}`,
and a unit `Req` would reject that (the `bookmarks.list` bug) — and
`MuxBusDisconnectResp` for what was an inline `json!({})`.

All generated shapes match the inline types they replace exactly.

## Two commands deliberately not migrated, for different reasons

**`activity` and `recordtevent` have no backend handler at all.** They are in
the contract test's `KNOWN_LIVE_UNREGISTERED` list — the frontend calls them and
they fail at run time today. There is nothing to generate *from*, and wiring a
handler or removing the callers is product work, not a refactor. Left exactly as
found.

**`widget.health` and `widget.api` read their fields straight off a
`serde_json::Value`** with lenient defaults: a missing `port` becomes `0`, which
the handler answers with `{healthy:false}` rather than an error. A typed `Req`
would convert those into deserialization failures — a behaviour change. Same
reasoning that kept `bundle.validate` untyped in #3305.

That is now three commands across the migration where the right answer was *not*
to type the request. The pattern is consistent: handlers that deliberately
accept loose input, or that transform the payload before deserializing, should
keep their `Value` and have only their response typed.

## One test worth keeping

These structs rely on `#[serde(rename_all = "camelCase")]` to match the wire
(`cognitoDomain`, `clientId`, `expiresAt`). Drop that attribute and the struct
still compiles, the binding still generates, and the gate still passes — it just
silently stops parsing real payloads. The test pins both directions: camelCase
parses, snake_case does not.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3710 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 123 generated types current
- `scripts/check-spec-citations.sh` -- ok (15 files), `check-doc-status.sh` -- ok
- `npm test` -- 91 failing suites, identical to the baseline on `main`, none of
  them touched by this PR

agent_id: manoz-0803a


a5af:
@codex review

PR #3314, opened 2026-09-17
Manoz@Area54: docs: refresh the migration wrap-up report (ten domains, 25% dead-code finding)

GenericAgentX-asaf:
`REPORT_MIGRATION_WRAPUP_STATUS_2026_09_16.md` is marked `active` — a living
status doc — and it still said "four domains in" with metrics from #3295. A
stale living doc is worse than no doc, so this brings it current and records
three things that are not written down anywhere else.

## Refreshed numbers (re-measured, same greps as the originals)

| metric | at report creation | now |
|---|---|---|
| `srv-types.d.ts` | 2,755 lines | **2,222** |
| declared commands | 277 | **221** |
| declared-but-unregistered | 69 | **14** |
| generated bindings | 14 | **123** |
| stub files importing generated types | 1 | **9 of 16** |
| `register_typed` / `register_handler` | 12 / 232 | **63 / 181** |

Fourteen PRs are now listed with what each one actually found, rather than a
count.

## §5.4b — a quarter of the stub surface was dead code

The single biggest change to the plan. 69 of 277 commands had no backend
handler, concentrated in exactly the domains that looked largest by line count
(`file` 23/42, `workspace` 22/39, `misc` 14/22, `block` 10/26). #3311 and #3312
deleted 65 of them.

The section also records two things that would have caused a bad deletion:

- **`captureblockscreenshot` is a reverse RPC** — the frontend implements it and
  the backend calls it. "No server handler" is *not* sufficient evidence a
  command is dead.
- **11 unregistered commands are still actively called** and fail at run time
  today. Pre-existing, tracked, and deliberately untouched across three PRs
  because fixing them is product work. It is the most actionable thing the
  contract data shows and it is currently nobody's task — which is exactly the
  kind of thing this report exists to surface.

## §5.4c — not every handler should be typed

The spec's step 2 reads as "migrate every handler to `register_typed`". Three
commands are deliberate exceptions, and rather than leave that reasoning
scattered across PR descriptions it is now a stated rule: a handler that
transforms its payload before deserializing (`bundle.validate`), or that
deliberately accepts loose input (`widget.health`/`widget.api`), should keep its
`Value` request and have only its **response** typed.

## §5.4d — where generating produces a *worse* type

Three distinct failure modes, all hit in practice:

1. ts-rs cannot express an optional property whose Rust type is not `Option<T>`,
   so `#[serde(default)]` on a `bool`/`String`/`Vec` generates as **required**.
2. A Rust `String` does not carry its domain — `verdict`, `collision`,
   `outcome`, `scope` and `status` all carry closed sets the frontend branches
   on, and each would have silently degraded to `string`.
3. The hand-written TypeScript is sometimes **richer** than the Rust
   (`blockdef: BlockDef` vs `Option<serde_json::Value>`).

## Verification

- `scripts/check-doc-status.sh` -- ok
- `scripts/check-spec-citations.sh` -- ok
- `scripts/gen-docs-index.sh --check` -- no specs changed, index not asserted
- Docs-only change; no code touched.

agent_id: manoz-0803a


a5af:
@codex review

PR #3315, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate all 16 skill commands to register_typed

GenericAgentX-asaf:
The `skill` domain — all 16 commands.

`skill.rs` declared its request shapes as **function-local anonymous
`struct Req` declarations inside each handler closure** — sixteen of them. The
bindings generator cannot see those, so the frontend's inline request types were
hand-maintained against types that had no addressable existence.

## Sixteen commands, ten types

Several commands genuinely share a shape, so they share a type rather than
getting sixteen near-identical ones:

| type | commands |
|---|---|
| `SkillAgentScopeData` | `skill.list`, `skill.catalog.list_for_agent` |
| `SkillAgentItemData` | `skill.get`, `skill.delete` |
| `SkillAgentBindingData` | `skill.bind`, `skill.unbind`, `skill.catalog.bind`, `skill.catalog.unbind` |
| `SkillBundleBindingData` | `skill.catalog.bind_to_bundle`, `…unbind_from_bundle` |

The names describe the **shape** (what is being addressed) rather than any one
command. That is what makes the sharing honest rather than incidental —
`skill.bind` and `skill.catalog.unbind` take `{agent_id, skill_id}` for the same
reason, so one type is correct, not merely convenient.

Also adds three result types for what were inline `json!` literals (`{deleted}`,
`{bound}`, `{unbound}`) and an empty request struct for `skill.catalog.list`,
which is called with `{}` — a unit `Req` would reject that.

## The three upserts: derived, not carved out

Every upsert field except the scope key and `name` is `#[serde(default)]`, so
all are omittable on the wire, and ts-rs generates them as **required** — it
cannot express an optional property for a non-`Option` field. In `block` I
carved this shape out. Here there is a better answer, because it was already
established for `Bundle`: generate the full type and **derive** the accurate
optionality from it in TypeScript.

```ts
export type SkillUpsertInput =
    Pick<CommandSkillUpsertData, "agent_id" | "name"> &
    Partial<Omit<CommandSkillUpsertData, "agent_id" | "name">>;
```

A field added in Rust flows through automatically, so this cannot drift — which
is the whole point, and is strictly better than hand-listing six optional
fields three times.

Making the fields `Option<String>` instead would change behaviour:
`default_skill_type` exists so an omitted `skill_type` becomes `"prompt"`, and
the rest treat `""` as "not provided". A test pins that, because it is the
reason the obvious simplification is wrong.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3728 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 136 generated types current
- `scripts/check-spec-citations.sh` -- ok (17 files), `check-doc-status.sh` -- ok
- `npm test` -- 91 failing suites, identical to the baseline on `main`, none of
  them touched by this PR

agent_id: manoz-0803a


a5af:
@codex review

PR #3316, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate all 18 mcp commands to register_typed

GenericAgentX-asaf:
The `mcp` domain — all 18 commands. Twin of #3315 (`skill`), same starting
point: seventeen function-local anonymous `struct Req` declarations inside the
handler closures, invisible to the bindings generator.

Eighteen commands now share ten named types, plus three result types for what
were inline `json!` literals (`{deleted}`, `{bound}`, `{unbound}`) and an empty
request struct for `mcp.catalog.list`.

## Deliberately NOT shared with `skill`

This is the one judgement call worth reviewing. The two primitives' shapes
rhyme:

```
McpAgentBindingData   { agent_id, mcp_id }
SkillAgentBindingData { agent_id, skill_id }
```

Collapsing them into one generic would mean **renaming a wire field to serve an
abstraction** — the payload really does say `mcp_id` for one and `skill_id` for
the other. That also matches the 09-06 DRY audit's own recommendation on these
twin primitives (cause 4), which this migration has not revisited and is not
trying to overturn.

There is a test pinning it rather than just a comment: a skill-shaped payload
must be **rejected** by the mcp binding type. If someone later decides to unify
them, that test is where the decision surfaces.

Sharing *within* each primitive is real (four commands genuinely take
`{agent_id, mcp_id}` for the same reason); sharing *across* them would not be.

## Upserts derived, not carved out

Same pattern as `BundleUpsertInput` and `SkillUpsertInput` — generate the full
type, derive the accurate optionality from it, so a field added in Rust flows
through automatically.

The defaults here are more load-bearing than `skill`'s: an omitted `transport`
is `"stdio"`, an omitted `config` is `"{}"`, and **`config` is parsed as JSON
downstream** — so making these `Option<String>` would turn an omission into a
different failure mode rather than the same one. Pinned by test, because that is
the reason not to take the obvious simplification.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3747 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 149 generated types current
- `scripts/check-spec-citations.sh` -- ok (17 files)
- `npm test` -- 91 failing suites, identical to the baseline on `main`, none
  touched by this PR

agent_id: manoz-0803a


a5af:
@codex review

PR #3317, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the 4 install commands, fix install.cancel error drift

GenericAgentX-asaf:
First slice of the `identity` domain, which spans **four** handler files
(`agent_handlers/identity.rs` 14 handlers, `identity_handlers.rs` 6,
`install_handlers.rs` 4, `app_api/identity.rs` 4). `install_handlers.rs` is
self-contained, so it goes first.

Promotes three private request structs, replaces a function-local anonymous one,
and adds five result types for what were inline `json!` literals.

## Real drift fixed: `install.cancel`

The handler writes an **explicit `serde_json::Value::Null`** for `error` on
success:

```rust
"error": if ok { serde_json::Value::Null } else { json!(format!(...)) }
```

So the key is *always* present. The hand-written stub declared
`error?: string` — i.e. the key may be absent, which it never is. The generated
type says `error: string | null`, and a test asserts the key is present rather
than just checking the value.

This is the same class as the `tab_id` P2 from #3294, in the opposite direction:
there a `skip_serializing_if` field was wrongly typed `| null`; here a
never-omitted field was wrongly typed optional.

## An asymmetry noted rather than hidden

`pinnedVersion` is `#[serde(default)]`, so the **server** accepts it missing —
but the hand-written stub declared it required, and the generated type keeps it
required (ts-rs cannot express an optional non-`Option` property anyway).

Callers being stricter than the wire is the safe direction, so I left it, but
both halves are pinned by test so the asymmetry is deliberate rather than
something a future reader has to rediscover.

## camelCase is load-bearing

These structs rely on `#[serde(rename_all = "camelCase")]` to match the wire.
Drop that attribute and the struct still compiles, the binding still generates,
and the gate still passes — it just silently stops parsing real payloads. The
tests assert camelCase parses and snake_case does not.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3759 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 158 generated types current
- `scripts/check-spec-citations.sh` -- ok (11 files)

agent_id: manoz-0803a


a5af:
@codex review

PR #3318, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the 6 auth/identity commands

GenericAgentX-asaf:
Second slice of the `identity` domain, covering `identity_handlers.rs`:
`auth.start`, `auth.poll`, `auth.cancel`, `auth.submitcallback`,
`auth.submitapikey`, `identity.ensureaccountdir`. Promotes eight private
structs.

`auth.start` uses the derive-from-generated pattern — `directAccount`,
`existingAccountId`, `authEnv` and `requiresTty` are all `#[serde(default)]` on
non-`Option` fields, so they are omittable on the wire but ts-rs generates them
as required.

## The contrast worth reviewing

Two `Option<String>` fields that look **identical in Rust** and have **opposite
wire contracts**:

| | Rust | wire | binding |
|---|---|---|---|
| `AckResp.error`, `EnsureAccountDirResp.dir` | `Option<String>` **with** `skip_serializing_if` | key **omitted** | `error?: string` |
| `InstallCancelResult.error` (#3317) | `Option<String>`, **no** `skip_serializing_if`, explicit `Value::Null` | key **always present** | `error: string \| null` |

Same field type, same apparent intent, different truth. This is the single most
recurring source of drift in the whole migration — it has now produced a P2
(`tab_id`, #3294), a corrected stub (`install.cancel`, #3317), and this pair,
which sit two files apart and would read as inconsistent without the reason.

Both are pinned by tests that assert the **serialized form** rather than the
value, because that is the only way to tell them apart:

```rust
assert_eq!(ok, json!({"success": true}));          // key absent
assert_eq!(ok, json!({"success": true, "error": null}));  // key present
```

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3770 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 166 generated types current

`identity` has two files left: `agent_handlers/identity.rs` (14 handlers) and
`app_api/identity.rs` (4).

agent_id: manoz-0803a


a5af:
@codex review

PR #3320, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate 7 identity commands; fix a generated tag the server never writes

GenericAgentX-asaf:
Third `identity` slice: `listidentityaccounts`, `getidentityaccount`,
`linkagentidentity`, `unlinkagentidentity`, `listagentidentities`,
`listallagentidentities`, `account.oauth.cancel`. Generates `IdentityAccount`,
`SecretRef` and `AgentIdentityLink`.

## The important one: ts-rs emitted a tag the server never writes

`SecretRef::OAuthConfigDir` carries
`#[serde(rename = "oauth_config_dir", alias = "o_auth_config_dir")]`.
**ts-rs does not honour that `rename`** — it emitted the derive-default
`o_auth_config_dir`, which is exactly the legacy tag the rename exists to stop
writing.

The variant's own doc comment records why that matters:

> the enum-wide `rename_all = "snake_case"` splits at every capital letter,
> including inside "OAuth" itself… Every real account on disk was written with
> `oauth_config_dir`… so the derived name silently stopped matching stored data
> … `identity_list()` aborts entirely on the first unparseable row

That is the "My Agents empty" incident. Shipping the generated binding would
have told the frontend to discriminate a union on a tag the server **never
serializes** — reintroducing that incident through the very codegen meant to
prevent drift.

**Nothing would have caught it.** `tsc` passes (the union simply never matches
at runtime), `check-rpc-bindings.sh` passes (the type exists and is current),
and no req-shape test covers a response discriminant. Fixed with
`#[ts(rename)]`, and pinned by a test asserting the *serialized* tag in Rust,
where it is checkable:

```rust
assert_eq!(v["backend"], "oauth_config_dir");
assert_ne!(v["backend"], "o_auth_config_dir",
    "the derive-default tag must never be written; it is read-only legacy");
```

This is the third distinct way generating produces a wrong type (after "cannot
express optional non-`Option`" and "a `String` does not carry its domain"), and
the first that would have caused a user-visible regression. Worth a line in the
spec.

## Three smaller corrections, in both directions

- `display_name` and `status` were marked **optional** in the hand-written
  declaration. Both are `#[serde(default)]` on non-`Option` fields with no
  `skip_serializing_if`, so the server *always* writes them — generated is
  right, hand-written was wrong.
- `context` is the opposite: ts-rs would emit bare `unknown`, but
  `default_context_json` is `{}` and every writer stores an object, so the
  hand-written `Record<string, unknown>` was **richer and accurate**. Preserved
  with `#[ts(type = ...)]`.
- `AgentDefinitionIdentity` is re-exported as an alias of the Rust
  `AgentIdentityLink` — same three fields, a naming difference only.

## Deferred, deliberately

`upsertidentityaccount`, `account.key.verify`, `account.oauth.start/poll` have
multi-path handlers returning **different shapes per path**, which needs
response types authored rather than promoted.

`deleteidentityaccount` is also held back: its `affectedAgents` field is
optional in the stub *"so older backends' `{ deleted }` shape stays
assignable"*. The current server always sends it, so generating it required
would be accurate-for-now but would drop a documented version-skew tolerance.
That is a product call, not a refactor one.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3775 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 171 generated types current

agent_id: manoz-0803a


a5af:
@codex review

PR #3321, opened 2026-09-17
Lark@narko: feat(history): SearchHistory — agents can search their own past conversations

GenericAgentX-asaf:
## Summary

Implements `SPEC_AGENT_HISTORY_SEARCH_2026_09_17.md` (merged to `main`).

**An agent's memory of what it did is bounded by its context window; its actual actions are not.** After a compaction or session reset it will answer questions about its own past confidently and wrongly, with nothing marking the boundary.

That isn't hypothetical — it happened three times in the session that prompted this spec:

1. Asked "why did you jekt Agent4?", an agent audited the only thing it could cheaply audit (the current session), found no `SendMessage` calls, and answered "I didn't." It had sent two, from a prior session whose history was lost.
2. Agent4 made the identical error — "no record in this session" — which briefly made two ordinary session resets look like an impersonation incident.
3. Neither could be settled, because no archived transcript was reachable.

## What this adds

Nothing new is discovered or parsed. `backend/history/` already indexes sessions per identity, and `bundle.export_for_agent_with_history` already uses `sessions_for_agent`. This adds the missing **verb**.

- **`SessionIndex::search_sessions`** — scans only the candidates it's handed, so the expensive `get_full` parse never touches a session the caller already ruled out on indexed metadata.
- **`HistoryService::search_for_agent`** — resolves linked identities, applies `since`/`until` against `modified_at` **before opening anything**, caps parsed candidates (newest first).
- **`GET /agentmux/reactive/history/search`** — full-auth route, deliberately *not* in the LAN-scoped set: conversation content is exactly what a captured LAN key must not reach. Parsing on `spawn_blocking`.
- **`SearchHistory` MCP tool** — exposes no `agent` parameter; always sends the caller's own `AGENTMUX_AGENT_ID` from trusted spawn-time env.

## Two design points worth review

**`tool` is its own filter, not part of `query`.** "Did I *call* `SendMessage`" is a different question from "did I write the word SendMessage", and only the structural answer settles an audit. That is exactly the question that was unanswerable this session.

**`truncated` is load-bearing.** A search that silently stops on its budget and returns nothing reproduces the precise failure this exists to prevent — concluding "I never did that" from an incomplete audit. An exhausted budget and a genuine absence are never conflated, and there's a test named for it.

## Honest limitation

Scope is own-history only, because reading another agent's conversation is already governed by `conversation_visibility` + the `transcript_request` tier rules, and a search verb reaching other agents' transcripts directly would be a second, ungoverned disclosure path around that machinery.

**But this is a client-side convention, not server enforcement.** The route takes an `agent` parameter (it must — the server has to know whose sessions to resolve) and is gated only by the instance-wide `auth_key` every agent shares, so the server cannot tell who is calling. Same property `/reactive/transcript` already has, so no new boundary is crossed — but I've documented it in the route, the spec's §5, and the credential map rather than implying enforcement I don't have. Real enforcement needs verifiable per-agent identity on local routes, which doesn't exist yet.

## Testing

6 new tests: content matches, structural tool matches, **budget-vs-absence**, candidate isolation, role filtering, and snippet bounding on a 100 KB message. Full `agentmux-srv` suite green — **3779 passed**.


a5af:
@codex review

PR #3322, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the 2 tool-status commands

GenericAgentX-asaf:
First `workspace` slice. `workspace` spans four handler files (`websocket.rs`,
`cli_handlers.rs` 7 handlers, `tool_handlers.rs` 2, `mod.rs`), so it is being
migrated one file at a time; `tool_handlers.rs` is self-contained and goes
first.

Generates `ToolStatus`, `ToolStatusEntry`, `GetToolStatusResult`,
`CommandInstallToolData`, `InstallFailure` and `InstallToolResult`, plus an
empty request struct for `gettoolstatus` (which ignores its payload but is
called with `{}`).

## One correction

`ToolStatusEntry.version` and `.path` were marked **optional** in the
hand-written declaration, but both are plain `Option`s with **no**
`skip_serializing_if` — the keys are always present and carry `null`.

That is the same class of error as `IdentityAccount.display_name` (#3320) and
the exact inverse of `AckResp.error` (#3318), which *is* `skip_serializing_if`
and genuinely omitted. Three occurrences across three PRs of the same
`Option<T>` ambiguity, so the test pins the **serialized form** rather than the
value — that is the only thing that distinguishes them.

## Pinning the enum tags

`ToolStatus` generates as a string union. After finding that ts-rs silently
ignored a `#[serde(rename)]` on `SecretRef::OAuthConfigDir` in #3320 — emitting
a tag the server never writes — a generated string union is only trustworthy
while its spellings are asserted somewhere checkable. So all five
`ToolStatus` tags are now pinned by a Rust test.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3782 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 173 generated types current

agent_id: manoz-0803a


a5af:
@codex review

PR #3323, opened 2026-09-17
Parlo@charlie: fix(linux): stop two isolated instances from sharing one dock icon

GenericAgentX-asaf:
## Summary

Two independently-running AgentMux instances on Linux (different channel and/or version, already fully isolated at the data-dir/socket level) showed up in the desktop dock/taskbar as a single icon with "2 windows" instead of two separate entries.

Root cause was two compounding bugs, both purely at the desktop-shell integration layer (the app's own multi-instance isolation was never at fault):

1. `window_settings.rs` wrote a static `b"agentmux"` into the Wayland app_id / X11 WM_CLASS for every window regardless of channel/version. GNOME/KWin/sway group dock entries purely by this string.
2. `install-linux-desktop.sh` always installed/overwrote the same shared `~/.local/share/applications/agentmux.desktop`, so even the launcher's click target would silently repoint to whichever instance started last.

Fix: both now key off `agentmux-<channel>-<version>`, mirroring the scheme already used for macOS's `CFBundleIdentifier` (`SPEC_MACOS_LAUNCH_COHERENCE_2026_06_18.md`). Full writeup in `docs/specs/SPEC_LINUX_DOCK_INSTANCE_GROUPING_2026_09_17.md`.

- Icons/labels are unchanged (`Icon=agentmux`, `Name=AgentMux`) — only the WM-matching key and desktop-file name vary per instance.
- `.deb`/`.rpm` builds and the AppImage's own internal top-level `.desktop` substitute the matching `StartupWMClass` at build time (they have no channel concept, so always `stable`).
- Scoped to Linux only. Windows has an analogous unfixed `AppUserModelID` gap (`agentmux-cef/src/lib.rs`) — tracked separately, not part of this PR.

## Test plan

- [x] `cargo check -p agentmux-cef` passes, no new warnings.
- [x] `bash -n` on every modified script passes.
- [ ] Build two local AppImages of different channels/versions and confirm they register as two dock entries (needs a Linux desktop session to verify visually — not done in this sandboxed environment).


a5af:
@codex review

PR #3324, opened 2026-09-17
Opaz@charlie: fix(appimage): key the extract-once cache per build, not per version

GenericAgentX-asaf:
## Summary

`linux-apprun.sh` cached its extraction under `~/.local/share/agentmux/extracted/$VERSION` and skipped extraction whenever a launcher already existed there. `task package` deliberately does **not** bump the version, so every local build of a version shared one directory — and the **first** one extracted won permanently. Later builds silently re-exec'd the older binary, carrying the per-build data-dir channel baked into it.

Reproduced with two local 0.56.3 builds: the second launch ran the first's binary in the first's channel, so a fix that had just been built was never exercised.

**This is a plausible explanation for the recurring "works in `task dev`, broken in the portable" reports** — `task dev` always runs freshly built code, while the portable re-runs whatever was extracted first, indefinitely, across rebuilds.

## The fix

`build-appimage-linux.sh` stamps `usr/share/agentmux/BUILD_ID` from the existing `AGENTMUX_BUILD_LABEL`; AppRun keys on it via `CACHE_KEY="${BUILD_ID:-$VERSION}"`. Release AppImages set no label and keep one cache per released version, exactly as before.

Two things this had to get right:

**1. The extracted-run marker was inheritable.** `AGENTMUX_EXTRACTED_RUN=1` is exported into every child process — including the shells in AgentMux's own panes. Launching any AppImage from inside a running AgentMux inherited it, took the "already extracted" branch, and ran from the FUSE mount without extracting or checking its cache. Observed live. It now carries the directory it refers to and is honoured only when it names *this* one — the self-describing shape tmux uses for `$TMUX` rather than a boolean.

**2. Per-build keying makes the prune dangerous.** These dirs now turn over on every package, so the pre-existing "keep the 2 most recent" rule would delete the ~384 MB tree a **running** instance is still lazily reading its binaries, `.pak` files and `libcef.so` from — killing a session that did nothing wrong. Pruning now skips any dir a live process is executing from. It still prunes: at 384 MB a build, unbounded growth isn't an option.

## Test plan

- [x] `scripts/linux-apprun.test.sh` — **8 passing**
- [x] Falsified against the pre-fix script: **3 fail**, including both builds resolving to `.../extracted/0.56.3` and the in-use dir being deleted
- [x] An earlier revision of this test was **vacuous** — two cases reimplemented the key resolution instead of exercising the script, and the in-use case passed only because the old prune block aborted on an unbound variable. It now sources the script's own lines and defines `VERSION` so that revision genuinely runs.
- [x] Verified in a real artifact: `BUILD_ID = 0.56.3+g59e9921.dirty.20260917T140207.431980` present inside the built AppImage

## Note

`AGENTMUX_EXTRACTED_RUN` is one instance of a broader defect — AgentMux exports its full instance identity into pane shells, and anything launched from a pane inherits it. That has its own RCA and follow-up PR; this change fixes only the launch-behaviour flag.


a5af:
@codex review

PR #21, opened 2026-03-04
fix(linux): fix window header drag on Wayland

AgentU-asaf:
## Problem

Window dragging via the title bar did not work on Linux/Wayland. Clicking and dragging the header had no effect.

**Root cause:** On Wayland, WebKitGTK intercepts `wl_pointer` events at the Wayland protocol level before GTK's signal system sees them. Tauri's `startDragging()` IPC path calls `begin_move_drag(1, x, y, 0)` (timestamp=0 = `GDK_CURRENT_TIME`), but Wayland's `xdg_toplevel_move` requires a valid compositor event serial — passing 0 is silently rejected by the compositor.

## Fix

Three changes working together:

1. **`src-tauri/src/main.rs`** — Force `GDK_BACKEND=x11` when `WAYLAND_DISPLAY` is set. This uses XWayland, where X11's `begin_move_drag(timestamp=0)` works correctly. Also sets `WEBKIT_DISABLE_DMABUF_RENDERER=1` to prevent the blank-screen issue caused by WebKit's DMA-BUF GPU renderer failing when GTK switches to X11 backend.

2. **`src-tauri/src/lib.rs`** — Attach a native GTK `button-press-event` handler directly to the WebView widget. On X11 (via XWayland), GTK signals now fire normally. The handler intercepts left-clicks in the top 40px (header region) and calls `begin_move_drag` with `event.time()` — the actual X11 event timestamp — for reliable dragging.

3. **`frontend/`** — Add JS-level `startDragging()` in `window-header.tsx` and `windowdrag.tsx` as a cross-platform fallback (handles macOS and Windows where the native GTK path doesn't apply).

## Test plan

- [x] Verified window dragging works on Linux/Wayland (XWayland) — click and drag the title bar moves the window
- [x] Verified no blank/white screen on startup
- [x] App renders correctly with full UI
- [ ] Verify drag still works on macOS (JS startDragging path)
- [ ] Verify drag still works on Windows (JS startDragging path)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3326, opened 2026-09-17
Opaz@charlie: fix(srv): stop panes inheriting this instance's identity

GenericAgentX-asaf:
## Summary

Both pane-spawn paths built their environment by inheriting srv's — and srv's environment is exactly where instance identity lives. The launcher injects `DataPaths::to_env_vars` (data dir, config dir, cache dir, channel, runtime mode); srv adds `AGENTMUX_DATA_HOME`, `AGENTMUX_CONFIG_HOME`, `AGENTMUX_LOCAL_URL`.

- **terminal panes** — `portable_pty::CommandBuilder` seeds from `std::env::vars_os()`, so `.env()` calls are overrides on a full copy
- **agent panes** — `tokio::process::Command` inherits by default

So every `AGENTMUX_*` variable reached every pane, and every process launched from one — **including another AgentMux build, which then adopted the launching instance's identity instead of the one baked into it at compile time.**

Observed live: a 0.56.3 build launched from a pane wrote into a 0.56.2 instance's channel; a dev srv resolved another instance's data dir through an inherited `AGENTMUX_DATA_HOME`.

This is **I6** breached through a channel I1–I6 never named — they enumerate named OS objects and directories, and their threat model is "a new launch must not crash a running one". This is neither: it's silent, and it corrupts the *new* instance from the *old* one.

`env_clear` appeared **zero times** in this repo. The entire pane-side sanitization surface was five `env_remove` calls in one branch.

## Design

`backend/pane_env.rs` owns the policy. An `AGENTMUX_*` key crosses into a pane only if it's on `PANE_ENV_KEEP`.

**Allowlist, not denylist** — and computed from the **live** environment rather than a hardcoded list, so a variable introduced later (by us or a dependency) is excluded by default instead of leaking until someone remembers to deny it. The previous behaviour was an implicit allow-everything; this inverts it.

The keep-set comes from an audit of every `process.env` read under `backend/shellintegration/` plus the shell rc scripts, and each entry carries its reason in the source. **Eleven high-risk vars turned out to have zero in-pane consumers** and are stripped with no loss at all.

Both call sites strip **before** applying their own explicit `.env()` values — the terminal path re-sets what a pane needs a few lines later, so stripping first costs nothing and keeps the allowlist honest rather than load-bearing.

## Also closes the sentinel gap

`agentmux-launcher/src/data_dir.rs` already ignores ambient `AGENTMUX_*` when the bare `AGENTMUX` sentinel is present — the right mitigation, and it predates this PR. But it was set only on the interactive shell-pane branch; the other setter (`shellexec.rs::build_mux_env`) **has no callers**. The agent path never set it, so a build launched from an agent's shell tool had `nested == false` while carrying the launching instance's channel.

**Every confirmed failure came through that gap**, and `retro-portable-0496-pool-window-false-close-2026-06-27.md:143-145` predicted it.

## Not fixed here — named rather than glossed

`AGENTMUX_LOCAL_URL` and `AGENTMUX_AUTH_KEY` **stay inheritable**. `lib/muxclient.mjs` reads exactly those two and every helper CLI goes through it. They are the largest residual coupling: a process inheriting both holds an authenticated handle to the instance that spawned it, and `AGENTMUX_AUTH_KEY` is the sole auth factor for the whole App API (a string equality against a per-launch UUID, no scoping, expiry or rate limit). Replacing them is a redesign of how helpers reach the server — a per-invocation handoff — not a strip. Tracked in the retro §9.4.

## Test plan

- [x] 6 `pane_env` unit tests
- [x] `the_identity_vars_that_caused_the_breach_are_stripped` pins the four variables behind the confirmed failures, so the breach can't be reopened by quietly adding one to the keep-set
- [x] `an_unknown_agentmux_var_is_stripped_by_default` — the allowlist property; what makes this survive variable #21
- [x] `agentmux-srv`: **337 blockcontroller tests** still green
- [x] doc gates: `check-doc-status`, `gen-docs-index --check`, `check-spec-citations` all pass

## Context

Full RCA, including the **five prior documents** that identified this class and left the scoping question explicitly open (`SPEC_DEV_BADGE_RUNTIME_MODE_2026_06_12.md:226` asks precisely "which `AGENTMUX_*` vars are instance identity vs legitimately inheritable?" and was never answered): `docs/retro/retro-env-inheritance-instance-isolation-breach-2026-09-17.md`

Spec: `docs/specs/SPEC_PANE_ENV_ISOLATION_2026_09_17.md`


a5af:
@codex review

GenericAgentX-asaf:
## Live verification

Not just unit tests asserting the policy — a real terminal pane spawned by a dev instance carrying this change, environment dumped from inside the PTY.

**Before** (same machine, instance without the fix) — 20 variables, the full identity set:

```
AGENTMUX_AGENT_ID, AGENTMUX_AGENTS_DIR, AGENTMUX_APP_PATH, AGENTMUX_AUTH_KEY,
AGENTMUX_BLOCKID, AGENTMUX_CEF_CACHE_DIR, AGENTMUX_CHANNEL, AGENTMUX_CONFIG_DIR,
AGENTMUX_CONFIG_HOME, AGENTMUX_DATA_DIR, AGENTMUX_DATA_HOME,
AGENTMUX_EXTRACTED_RUN, AGENTMUX_INSTANCE_DIR, AGENTMUX_INSTANCE_RUNTIME_DIR,
AGENTMUX_LOCAL_URL, AGENTMUX_LOG_DIR, AGENTMUX_RUNTIME_MODE,
AGENTMUX_SHARED_DIR, AGENTMUX_SPLASH_READY_FILE, AGENTMUX_SRV_PIPE_PATH
```

**After** (`env | grep '^AGENTMUX' | sort` inside a terminal pane on this branch):

```
AGENTMUX=1
AGENTMUX_BLOCKID=a632ae71-14f9-40a3-902e-aea081929fb6
AGENTMUX_LOCAL_URL=http://127.0.0.1:44855
AGENTMUX_LOG_DIR=/home/yas/.agentmux/logs
AGENTMUX_TABID=647dd912-3ab4-4783-b69e-7c0c0c07e31f
AGENTMUX_VERSION=0.56.3
```

20 → 6. Every variable that can redirect a data dir, config dir, cache dir, channel or runtime mode is gone, so a build launched from this pane resolves its own identity. `AGENTMUX=1` is the nesting sentinel, which the launcher wants present.

`AGENTMUX_AUTH_KEY` is absent from shell panes specifically because `config.rs:98` already scrubbed it from srv's own environment — it still reaches **agent** panes via the deliberate re-injection at `agent_handlers/input.rs:406`, which is the residual coupling called out in the PR description and retro §9.4.

Worth stating plainly: this is the check that matters. The unit tests assert what the policy *says*; only this shows the spawn path actually applies it.

a5af:
@codex review

PR #3327, opened 2026-09-17
Manoz@Area54: refactor(rpc): migrate the 12 agent core commands

GenericAgentX-asaf:
First `agent.ts` slice. `agent.ts` is the last and largest domain — **51
commands across twelve handler files** — so it is being migrated a file at a
time. `agent_handlers/core.rs` is the biggest single file at 12 commands.

Generates `AgentDefinition`, `AgentContent` and their command types, plus three
empty request structs for the payload-ignoring commands and two result types for
inline `json!` literals.

## Two real drifts in `AgentDefinition`, both from diffing before deleting

**Eleven fields were wrongly marked optional.** The hand-written declaration had
`accounts?`, `parent_id?`, `updated_at?`, `container_image?` and seven more.
`agents.rs` has **zero** `skip_serializing_if` attributes anywhere — the server
writes every field on every response, so none of those keys is ever absent.

**Two fields were missing entirely.** `container_volumes` and `container_name`
have always been on the wire and the frontend could not see them. Surfaced
rather than dropped — the same call as `project_instructions` in the bundle
import slice.

Also: `reseedagents` was typed `Promise<void>` in the stub while the handler
returns `{deleted, created, skipped}`. Those counts are now visible. Whether the
UI uses them is a product question; the binding should say what the wire
carries.

## create/update use the derived-input pattern

Sixteen of their fields are `#[serde(default)]` on non-`Option` types, so ts-rs
generates them required. `AgentDefinitionCreateInput` / `AgentDefinitionUpdateInput`
derive the accurate shape from the generated type rather than hand-listing —
same approach as `BundleUpsertInput` and `SkillUpsertInput`.

I checked the blast radius before committing to making the fields required: a
probe that flipped the ambient declaration produced **2 errors in one test
file**, so the change is cheap. That fixture now carries all fields.

## A test that encoded a temporary state

`register_typed_records_each_command_request_and_response_type` asserted
`listagents` was **absent** from the schema registry "because it is not migrated
yet" — a statement about a temporary state, not an invariant, so migrating it
correctly is what broke it. Rewritten to assert what the command now records.

Worth noting **the full `cargo test` caught this locally rather than in CI** —
that is the habit change from #3299, where the same class of failure reached CI
because I had run a filtered suite.

## Verification

- `cargo test -p agentmux-srv` (**full suite, unfiltered**) -- 3818 passed, 0 failed
- `npx tsc --noEmit` -- clean
- `scripts/check-rpc-bindings.sh` -- ok, 198 generated types current
- `scripts/check-spec-citations.sh` -- ok (62 files)
- dead-import scan (comment-stripped) -- clean

agent_id: manoz-0803a


a5af:
@codex review

PR #3328, opened 2026-09-17
Manoz@Area54: generate the template/fork RPC wire types and finish agent_handlers/template.rs

GenericAgentX-asaf:
Completes `agent_handlers/template.rs`. The last three handlers —
`forkagentdefinition`, `renameagentdefinitiontitle`, `agentdeflisthiddentemplates` —
move to `register_typed`, and their request shapes are generated from Rust by ts-rs
instead of retyped by hand in the stub. Their responses are the storage-level
`AgentDefinition`, which is why these three waited for #3327.

`template.rs` now has zero `register_handler` registrations.

## Two things the generator could not say on its own

**`branch_label` is optional and ts-rs cannot express that.** It is
`#[serde(default)]` on a `String`, so the server accepts it missing, but ts-rs only
marks a field optional when the Rust type is `Option<T>` — the generated type calls
it required. Rather than keeping the hand-written shape, the stub *derives* from the
generated one:

```ts
export type ForkAgentDefinitionInput = Omit<CommandForkAgentDefinitionData, "branch_label"> &
    Partial<Pick<CommandForkAgentDefinitionData, "branch_label">>;
```

Field names and types stay authoritative — a rename in Rust breaks that line — while
the one thing ts-rs cannot say is restored locally. Same pattern as `BundleUpsertInput`.
A Rust test pins the server half: drop the `serde(default)` and the derived TS type
silently becomes a lie.

**"No argument" has two encodings and serde accepts each from only one.** The stub
sends `{}`; a client that omits `data` arrives as `null`. `()` deserializes only from
`null`, a struct only from an object. So `agentdeflisthiddentemplates` registers
`Option<CommandListHiddenTemplatesData>`, which takes both. This was found by testing
it rather than by reasoning about it — the bare struct failed with
`invalid type: null, expected struct CommandListHiddenTemplatesData`.

## Test change worth reading

The existing schema test asserted `renameagentdefinitiontitle` was **absent** from the
registry (it was unmigrated). It is present now, so that assertion is inverted into its
recorded request/response type names rather than deleted — the absent-means-unmigrated
property is what makes migrating one command at a time safe, so it should keep being
asserted, just with the opposite sign.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3817 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 189 generated types current

Stacked on #3327 (`AgentDefinition`); merge that first.

agent_id: manoz-0803a


a5af:
@codex review

GenericAgentX-asaf:
Rebased past #3327 (merged) and added one commit.

#3327 shipped `containerruntimeavailable`/`reseedagents`/`exportagents` registering a bare empty struct; this PR registers `Option<CommandListHiddenTemplatesData>` for `agentdeflisthiddentemplates`. Two different answers to the same question in adjacent files, and only one is right — a bare struct rejects a `null` body with `invalid type: null, expected struct ...`, which is how this was found in the first place. The other three are wrapped to match, with one test covering the property for the registration shape rather than per command.

agent_id: manoz-0803a

PR #3329, opened 2026-09-17
Manoz@Area54: generate the Drone pane RPC wire types and migrate all six drone commands

GenericAgentX-asaf:
`drone_handlers.rs` had zero typed registrations. It has six now, and no
`register_handler` left. The Drone pane's whole wire surface — `DroneDefinition`,
`DroneGraph`, `DroneFlowNode`, `DroneFlowEdge`, `DroneViewport`, `DroneRun`,
`DroneBlockState`, `BlockKind`, `RunStatus`, plus the six request/response shapes —
is generated from Rust by ts-rs instead of retyped by hand.

Three copies of the block-state shape collapse into one: the Rust struct, the ambient
global, and a third hand-written mirror in `drone-run-state/types.ts` that carried a
comment saying it was a mirror. That one derives from the generated type now.

## Found by diffing generated output against the hand-written declarations

**ts-rs does not read `#[serde(rename)]` on a field.** `FlowNode.node_type` is
`#[serde(rename = "type")]` — the wire key, and the key xyflow reads, is `type`. The
first generated binding said `node_type`, which nothing sends and nothing reads. Same
failure mode as the identity `oauth_config_dir` tag: both halves compile, tsc passes,
and the field just silently stops resolving. Fixed with `#[ts(rename = "type")]`, and
a test asserts the serialized key and the generated text agree.

**`type` is omitted, not empty.** It was `String` + `skip_serializing_if = "String::is_empty"`,
so a required `type: string` in TS would be wrong in the *read* direction — which is the
direction the canvas uses. It is `Option<String>` now, with an `is_blank` skip predicate
rather than `Option::is_none`, so `Some("")` is still omitted and the wire bytes are
unchanged. Tested for both blank forms.

**Two fields are richer in TS than in Rust, and stay that way.** `FlowNode.data` is
`serde_json::Value` but every block has a `kind`; `BlockState.status` is a `String` but
the frontend branches on a closed set. Both keep their narrower hand-written TS via
`#[ts(type = ...)]` — generating `any`/`string` would be a downgrade dressed up as
codegen. `status` gets a test asserting the union still covers every literal the
executor writes, because the Rust compiler cannot.

## One deliberate widening

`block_states` generates as `{ [key in string]?: DroneBlockState }` rather than the
hand-written `Record<string, DroneBlockState>`. That is ts-rs's natural mapping and the
more honest one — indexing a run's block map by an arbitrary block id really can miss.
`tsc --noEmit` passes unchanged, so no consumer was relying on the stronger claim.

`listdrones` registers `Option<ListDronesReq>` for the same reason
`agentdeflisthiddentemplates` does in #3328: the stub sends `{}`, a client that omits
`data` sends `null`, and serde takes each of those from only one of `()` and a struct.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3856 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 222 generated types current

agent_id: manoz-0803a


a5af:
@codex review

PR #3330, opened 2026-09-17
Manoz@Area54: generate the LSP RPC wire types out of three closure-local structs

GenericAgentX-asaf:
Each of the three LSP handlers deserialized into a `Cmd` struct declared **inside its
own closure**:

```rust
Box::pin(async move {
    #[derive(serde::Deserialize)]
    struct Cmd { language: String, file_path: String }
    let cmd: Cmd = serde_json::from_value(data)...
```

That is the strongest form of the drift this migration exists to remove. A type nested
in a closure cannot be named, so it cannot be generated from and cannot be referenced —
the frontend had no choice but to restate all three shapes by hand, with nothing
connecting them to the Rust they describe.

The three requests, plus `lspstart`'s response (previously an inline `json!({..})`),
move to `rpc_types/lsp.rs` and are generated by ts-rs. **All four match the hand-written
declarations exactly**, so unlike the previous slices there is no shape drift to report —
this one is purely about provenance.

## Two deliberate choices

`lspsend.message` stays `serde_json::Value` with `#[ts(type = "unknown")]`. The backend
is a dumb proxy there — it re-serializes the value and writes it to the server's stdin
without inspecting it — so `unknown` is the honest type. A test covers the shapes a
narrower Rust type would have rejected: arrays, bare strings, `null`.

`lspsend` and `lspstop` answer `()`, and the registry records `()` rather than `Value`.
The stubs already said `Promise<void>`; now the two agree, and the test asserts it.

`lsp_handlers.rs` had no tests. It has three.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3838 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 209 generated types current

Note for the reviewer: `file.ts` also declares `fileappend` and `filejoin`, which have
**no backend handler at all** and fail at runtime today. They are two of the eleven
unregistered-but-called commands flagged in the wrap-up report — product bugs, out of
scope here.

agent_id: manoz-0803a


a5af:
@codex review

PR #3331, opened 2026-09-17
Manoz@Area54: generate the editor file-tree mutation wire types (seven commands, fourteen shapes)

GenericAgentX-asaf:
Seven file-tree mutation commands — `openinshell`, `renameeditorfile`,
`createeditorfile`, `createeditordir`, `deleteeditorfile`, `createscratchfile`,
`movescratchfile` — move to `register_typed`, and their fourteen request/response
shapes are generated from Rust by ts-rs.

Same root cause as #3330: each handler deserialized into a `struct Cmd` declared
**inside its own closure** and built its response from an inline `json!({..})`. Neither
form can be named, so neither could be generated from, and the frontend had no option
but to restate all fourteen by hand. All fourteen match the hand-written declarations
exactly — no shape drift to report, this one is purely about provenance.

`openinshell` and `deleteeditorfile` answer `()`, recorded as `()` rather than `Value`.
The stubs already said `Promise<void>`.

## Two properties worth asserting rather than assuming

**`DeleteEditorFileReq.recursive` has no `serde(default)`, deliberately.** Adding one
would compile, pass every gate, and quietly turn "a caller that forgot the field" into
"delete a directory tree it never asked to". The generated TS would not change either
way — ts-rs cannot express a default on a `bool` — so nothing downstream would flag it.
There is now a test that does.

**`CreateScratchFileReq`'s two fields are `Option<T>`**, which is the one optionality
case ts-rs handles unaided: `#[ts(optional)]` gives `display_name?: string` directly,
with none of the `Omit`/`Partial` derivation the `serde(default)`-on-`String` fields in
earlier slices needed.

## A false negative in my own tooling, worth flagging

I run a local scan for typed handlers that hand-serialize their response (it came out of
the `getagentcontent` finding on #3327). It matched `serde_json::to_value` only, so it
passed `movescratchfile` — which was still returning `Ok(Some(json!({..})))` — completely
clean. The schema-registry test is what caught it, with
`left: "core::option::Option<serde_json::value::Value>"`.

The scan now matches inline `json!` too, and I verified it isn't vacuous by re-injecting
the old line: it reports `server/editor_handlers.rs:996  "movescratchfile"`. Worth noting
because it's the second time a *test* caught something a *scan* missed, which is the right
way round but only because the test existed.

`readeditorfile`, `writeeditorfile`, `listeditordir`, `geteditorhome`, `geteditorroots`
and the four watcher commands are still untyped — next slices.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3871 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 234 generated types current
- ambient-duplicate + dead-import + (widened) hand-serialized-response scans: clean

agent_id: manoz-0803a


a5af:
@codex review

PR #3332, opened 2026-09-17
Manoz@Area54: generate the agent-instance wire types and finish agent_handlers/instance.rs

GenericAgentX-asaf:
`instance.rs` had one typed registration out of five. It has five now, and the
storage-level `AgentInstance` row — which is what the other four answer, and the reason
they waited — is generated rather than hand-written.

## The generated `AgentInstance` is stricter than the ambient one, and correctly so

Nine fields were typed `?:` by hand (`parent_instance_id`, `block_id`, `session_id`,
`github_context`, `identity_id`, `memory_id`, `instance_name`, `working_directory`,
`display_hidden`). The Rust struct's own doc comment already said why they are not
optional:

> `block_id` / `session_id` / `github_context` are modelled as empty strings on the wire
> rather than `Option<String>` to match the existing schema conventions
> (`NOT NULL DEFAULT ''`).

`#[serde(default)]` affects *deserialization* only — the server always serializes these
keys. So the hand-written `?` was wrong in the read direction, which is the direction
every consumer uses. `tsc --noEmit` passes with them required, so nothing was relying on
the weaker claim.

## One deliberate behaviour change

`listagentinstances` used to `serde_json::from_value(data).unwrap_or_default()`. That
bought two things and only one was wanted:

- **Wanted:** both encodings of "no filter" work — the `{}` the stub sends and the `null`
  a client that omits `data` sends. `Option<Req>` keeps that.
- **Not wanted:** a *malformed* filter silently became "no filter", so `{"status": 5}`
  returned every row. Answering an unfiltered list to a request whose filter failed to
  parse is the wrong answer to give quietly.

It errors now, and a test pins both halves so the change is visible rather than
incidental. Flagging it explicitly because it is the one place in this slice where the
wire behaviour changes — happy to revert to `unwrap_or_default()` if you'd rather keep
the tolerance.

`createagentinstance`'s six `#[serde(default)]` `String` fields are optional on the wire
and ts-rs cannot say so, so the stub derives `CreateAgentInstanceInput` from the generated
type rather than restating it — same pattern as `BundleUpsertInput`.

The existing schema test asserted these four were **absent** from the registry. Inverted
rather than deleted: absence-means-unmigrated is what makes migrating one command at a
time safe, so it should keep being asserted.

Worth noting for the dead-stub ledger: `ListAgentInstancesCommand` has **no frontend
caller**. The handler is live, so this is a stub-side cleanup candidate rather than
anything to remove here.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3869 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 230 generated types current

agent_id: manoz-0803a


a5af:
@codex review

PR #3333, opened 2026-09-17
Manoz@Area54: generate the editor read wire types (five commands, ten shapes)

GenericAgentX-asaf:
`readeditorfile`, `writeeditorfile`, `listeditordir`, `geteditorhome` and
`geteditorroots` move to `register_typed`. Same root cause as #3331 and #3330:
closure-local `struct Cmd` requests and inline `json!({..})` responses, neither of which
can be named, so the frontend restated both ends by hand.

After this, the only untyped things left in `editor_handlers.rs` are the four watchers.

## The generated read result is stricter than the ambient one

`CommandReadEditorFileResult` had four fields typed `?:` with the note *"Optional for
back-compat; absent ⇒ treat as UTF-8"*. But `decode_file` returns all four
unconditionally and the handler writes all four every time — **the server has never
omitted them.** The optionality described a client reading a response from a build
predating `SPEC_EDITOR_FILE_ENCODINGS`, which is not something a binding can express and
not something any current server does. A test pins that all six keys are always present,
since the Rust compiler cannot (they are plain `String`/`bool`, not `Option`).

By contrast, `CommandWriteEditorFileData`'s three encoding fields **stay** optional, for
a real reason rather than hedging: omitting them means "UTF-8 / no BOM / LF", which is
what a caller that never read the file back wants. They are `Option<T>` in Rust, so ts-rs
marks them optional unaided.

## A latent bug fixed on the way past

`listeditordir` built `Vec<Value>` and then sorted it by re-parsing its own output:

```rust
let a_dir = a["is_dir"].as_bool().unwrap_or(false);
```

Rename that key and every row silently sorts as a file — no error, no failing test, just
a file tree that stops putting folders first. It builds `Vec<DirEntry>` and sorts on the
field now.

## The ambient-duplicate scan had a blind spot

`DirEntry` lives in `custom.d.ts`, not `srv-types.d.ts`, and my standing scan only read
the latter — so it could not have seen the duplicate this slice would otherwise have
left behind, which is precisely what that check exists to catch. It reads both files now.

Two consumers (`editor-model.ts`, `file-tree-model.ts`) were resolving these types
through the globals; `tsc` caught that, not the scan. Right outcome, wrong order.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3891 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 247 generated types current
- ambient-duplicate (now 2 files) + dead-import + hand-serialized-response scans: clean

agent_id: manoz-0803a


a5af:
@codex review

PR #3334, opened 2026-09-17
Manoz@Area54: generate the agent-skill wire types, surfacing a create/update default asymmetry

GenericAgentX-asaf:
`skills.rs` had no typed registrations. It has four now and none left. The storage-level
`AgentSkill` row and all four command shapes are generated from Rust by ts-rs.

## A real asymmetry the hand-written TS was hiding

Writing the two request types down next to each other surfaced this: **`skill_type`
defaults differently on create and update.**

```rust
// CommandCreateAgentSkillData
#[serde(default = "default_skill_type")]  // -> "prompt"
pub skill_type: String,

// CommandUpdateAgentSkillData
#[serde(default)]                          // -> ""
pub skill_type: String,
```

The hand-written TS typed both `skill_type?: string`, so the difference was invisible.

It matters because **`updateagentskill` is a full replace, not a patch** — the handler
rebuilds the whole `AgentSkill` from the request and keeps only `agent_id` and
`created_at` from the existing row. So a create-then-update round trip that omits
`skill_type` silently blanks a value the create had filled in.

**I did not fix it.** Which default is right is a product decision, and nothing in the
frontend calls `updateagentskill` today, so it is latent rather than active. What this PR
does is stop hiding it:

- a test asserts both defaults explicitly, so "tidying" one without meaning to change the
  other fails loudly;
- the generated `CommandUpdateAgentSkillData` deliberately does **not** restate its fields
  as optional the way `CreateAgentSkillInput` does — advertising a patch API the server
  does not implement is how you end up here;
- the `serde(default)`s stay, since dropping them would reject payloads the server accepts
  today.

Happy to follow up with the actual fix (most likely: make update's default `"prompt"` too,
or make it a genuine patch) if you want it — just didn't want to decide that inside a
refactor.

## Dead-stub ledger

`CreateAgentSkillCommand`, `UpdateAgentSkillCommand` and `DeleteAgentSkillCommand` have
**no frontend callers**; only `ListAgentSkillsCommand` does. The handlers are live, so
these are stub-side cleanup candidates, not removals for this PR.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3884 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 242 generated types current

agent_id: manoz-0803a


a5af:
@codex review

Issue #3319, opened 2026-09-17
Universal Pane Tabs — finish cross-platform QA + remaining phases (handoff)


AgentO-asaf:
## What this is

Tracking issue for the "Universal Pane Tabs" redesign — every pane's header becomes a tab strip (cmux-style), so a Pane can hold multiple widget instances as tabs instead of each widget always getting its own split.

- **Design:** `docs/specs/SPEC_PANE_TABS_UNIVERSAL_CMUX_REDESIGN_2026_09_17.md`
- **Implementation plan / task breakdown:** `docs/specs/PLAN_PANE_TABS_UNIVERSAL_IMPLEMENTATION_2026_09_17.md`
- **PR (draft, not merged):** #3309 — branch `agento/pane-tabs-universal`

## Current state (as of this issue)

**Done and locally verified (typecheck, full test suite — 4098 passing, `cargo check`, real `task dev` launches):**
- Unified header/tab-strip row for **every** widget type: agent, terminal, browser, editor, sysinfo, swarm, armory, media, drone, help, warden.
- Agent/term keep their own richer Chrome components (fork lineage, rename-by-click, etc.); every other widget type shares one generic `GenericPaneChrome` component built from already-existing view-agnostic primitives.
- Went through 4 real rounds of ReAgent review on the agent/term slice — all legitimate findings, all fixed (see PR #3309's commit history for the specifics — a couple of them were genuine identity/regression bugs worth reading before assuming this is done).

**Explicitly NOT done — this is the actual work left:**
1. **Deep, per-widget-type manual QA.** We've confirmed nothing crashes with all 12 view types hoisted, but nobody has individually clicked through browser/sysinfo/swarm/armory/media/drone/help/warden's own "+"-opened widget picker, tab-switching, and tab-closing behavior one at a time.
2. **Cross-platform verification.** All development/testing so far has been on macOS only. This repo has real, documented platform-specific code paths adjacent to this feature (floating-pane tear-off has 3 separate platform implementations; window/title-bar handling differs by OS) — even though this specific feature's own code has no `target_os` branches (confirmed in the design spec), a real Windows (and ideally Linux) pass is the only way to be confident the visual result and interaction feel are actually right there, not just "didn't crash."
3. **`GenericPaneChrome`-specific automated tests** — unit tests exist for `PaneHeaderTabStrip` and the hoisting router, but not yet for `genericRenderPaneChrome`'s own tab-derivation/add/close logic.
[...]

PR #3309, opened 2026-09-17
feat(panes): universal pane tabs — unified header/tab-strip row (agent+term)

AgentO-asaf:
## Summary

Implements `docs/specs/SPEC_PANE_TABS_UNIVERSAL_CMUX_REDESIGN_2026_09_17.md` — collapsing every pane's old two-row chrome (a full header on top, tab pills in a separate row below it) into **one unified row**, matching cmux's model where the tab strip IS the pane's entire header.

**Scope, as it actually shipped** (extended live from the original agent/term-only plan after repo-owner confirmation — "don't all panes need to be generalized to this?"):
- `BlockFrame_Header` gained `leadingTabStrip`/`trailingAddButton` props: only overrides the iconview with pills when there are 2+ real tabs; for 0 or 1, the REAL iconview (name/icon/rename) shows, with "+" appended separately.
- `AgentPaneChrome`/`TermPaneChrome` migrated onto the new `PaneHeaderTabStrip` component.
- **Every other widget type** (browser, editor, sysinfo, swarm, armory, media, drone, help, warden) now ALSO gets the unified header, via one shared `GenericPaneChrome` component built entirely from already-existing, view-agnostic primitives (`blockStack`, `setActiveBlockInStack`/`closeBlockInStack`, and the widget bar's own picker for "+").
- Each tab pill now shows its widget-type icon to the left of the label (`PaneTabStrip`'s new optional `getIcon` accessor — omitted entirely for existing callers, so it's a pure addition).

**Handoff continuation (issue #3319, this session):**
- Windows boot verification — first time this branch has been tested outside macOS. Clean `task dev` boot, no runtime errors, no duplicate-header symptoms.
- Closed the explicit automated-test gap the implementation plan flagged: `GenericPaneChrome.test.tsx` covers tab-derivation, activate/close/add wiring, icon derivation, and focus-ring classList.
- Fixed a real, repo-owner-confirmed-live regression: `.generic-pane-stack` (GenericPaneChrome's wrapper) had no corresponding SCSS, so headers for all 9 newly-hoisted widget types rendered as unstyled, vertically-stacked block-level divs instead of a flex row — the exact failure mode agent/term already hit and fixed for themselves. (Found independently by both this session and a concurrent ReAgent P1 finding on the same PR — reconciled via rebase onto the already-landed fix rather than duplicating it.)

Went through several rounds of real ReAgent review across this PR's history — all legitimate findings, all fixed.

## Test plan
- [x] `npx tsc --noEmit` clean
[...]

PR #3335, opened 2026-09-17
feat(panes): make every pane generic — shared "+" picker, wheel-scrollable tab strip

Agent4-asaf:
## Summary

Follow-up to #3309 (issue #3319). Closes the gaps that still made **agent** a special-case pane type, per repo-owner directive: *"all panes should be generic … there should be no such thing as a non-generic pane."*

Four independent changes, one per commit:

1. **Agent's "+" is always shown.** `tab-strip-visibility.ts` hid it entirely on a fresh, unlaunched picker pane, so agent was the one type whose header had no "+" until an agent was running. Every `GenericPaneChrome`-driven type and term already showed it unconditionally. Removes the now-fully-unused `shouldShowTabStrip` predicate and its test.

2. **An overflowing tab strip is wheel-scrollable.** `.pane-tab-tip` is `flex: 0 0 auto` (tabs never shrink) and the strip was `overflow-x: hidden`, so once tabs stopped fitting the extras were clipped and *permanently unreachable*. Now `overflow-x: auto`, with the scrollbar track itself hidden — the same hidden-but-scrollable pattern `.tab-bar-scroll` already uses for the window tab bar.

   The wheel handler is a real `addEventListener("wheel", …, { passive: false })`, **not** the JSX `onWheel` prop: Solid delegates common events through one top-level listener registered passive, which silently no-ops `preventDefault()`. Confirmed live — the prop version ran but never scrolled.

3. **A fresh agent pane is titled "New Agent".** `AgentViewModel.viewName` fell back to `"Agent"` while a tab created via "+" in the identical state already fell back to `"New Agent"`. Only the fallback changes; a launched agent still shows its real name.

4. **Every pane's "+" opens the generic widget picker.** Agent and terminal were the only types you couldn't add a *different* widget type to — their "+" was hardwired to "fork another agent" / "open another shell". New shared `pane-tab-picker.ts` is used by all three chromes, and folds in the "New tab failed" toast only agent had. Tab labels/icons became type-aware, so a sysinfo tab in an agent pane reads "Sysinfo" with its own icon instead of "New Agent", and isn't numbered "Terminal 2" in a terminal pane.

## Known regression, deliberately called out

Terminal's old "+" started the new shell in the **current tab's cwd**. The generic picker path doesn't carry that through, so a new terminal tab now starts in the default directory. Restoring it means letting the picker pass inherited context from the active block — a type-specific branch inside the generic path — so it's left as a follow-up call rather than smuggled in here.

## Follow-up (next PR)

Collapsing `AgentPaneChrome`/`TermPaneChrome`/`GenericPaneChrome` into **one** chrome with a trait-like capability interface any ViewModel can implement (tab derivation, extra tabs, inline rename, a generic below-header slot for the progress bar, content wrapper, zoom). Then the `Generic*` → `Pane*` rename, as a pure mechanical diff.

## Test plan
- [x] `GenericPaneChrome.test.tsx` — 11 tests pass (mock made faithful to `addWidgetAsPaneTab`'s async signature, which the new picker chains `.catch()` on)
- [x] No new failures across `frontend/app/element`, `view/agent`, `view/term` — the failing files are the pre-existing Node 24/`solid-refresh` environment issue reproduced identically on `main`
- [x] Verified live in `task dev` on Windows by the repo owner: agent pane "+" present, tab icons, "New Agent" title
- [ ] Wheel-scroll over an overflowing strip — implemented after the last live check, not yet hands-on verified


a5af:
@codex review

PR #3337, opened 2026-09-17
Manoz@Area54: generate the editor watcher wire types, finishing editor_handlers.rs

GenericAgentX-asaf:
Follow-up to #3333 (merged) — finishes `editor_handlers.rs`. Originally opened as #3336, stacked on #3333; that PR was auto-closed when its base branch was deleted on merge, so this is the same commit re-opened against `main`.

The four watch/unwatch commands were the last closure-local `struct Cmd` requests in
`editor_handlers.rs`. After this the only untyped registration left in the file is
`writeagentconfig`, which is not an editor command — it's the launch path, and its stub
lives in `agent.ts`, so it migrates there rather than here. The schema test asserts it's
still absent, so that stays visible.

## One shared request type, one deliberately not shared

`watcheditorfile` and `unwatcheditorfile` share **one** `WatchEditorFileReq`. That is the
opposite of the call I made for `CreateEditorFileReq` / `CreateEditorDirReq` in #3331,
which are separate types despite having identical fields — so it's worth saying why the
two cases differ:

- create-file and create-dir are *merely same-shaped*. A future divergence between them
  is legitimate, and sharing would make that divergence look like a binding bug.
- watch and unwatch must address the **same thing**. An unwatch that cannot name exactly
  what the watch named leaks a watcher. Sharing makes a divergence a compile error.

`watchmediadir` / `unwatchmediadir` do **not** share, for the same reasoning read the
other way: the extension filter is genuinely not part of the registration key, so
unwatching really does take fewer fields.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3909 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 259 generated types current

agent_id: manoz-0803a


a5af:
@codex review

PR #3339, opened 2026-09-17
Manoz@Area54: generate the agent-history wire types and finish agent_handlers/history.rs

GenericAgentX-asaf:
`history.rs` had no typed registrations. It has three now and none left. The
storage-level `AgentHistory` row and all three command shapes are generated from Rust by
ts-rs.

**No shape drift** — all four generated types match the hand-written declarations
exactly. The only thing ts-rs cannot carry over is the two pagination defaults
(`listagenthistory.limit` → 50, `.offset` → 0, both `#[serde(default)]` on `i64`), so the
stubs derive `ListAgentHistoryInput` / `SearchAgentHistoryInput` from the generated types
rather than restating them. A test pins the server half — dropping a default would
silently make those derived TS types a lie, and nothing else would notice.

`AgentHistory.session_date` picked up a doc comment on the way past: it is set by the
server, not the caller, which isn't obvious from a request type that takes only
`agent_id` and `entry`.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3911 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 260 generated types current

agent_id: manoz-0803a


a5af:
@codex review

Issue #3338, opened 2026-09-17
Tracking: agent availability & automatic backgrounding (dock rows, the input gate, and who decides)


GenericAgentX-asaf:
Canonical doc: `docs/specs/TRACKING_AGENT_AVAILABILITY_AND_BACKGROUNDING_2026_09_17.md`

## The invariant

> If the agent has nothing in the foreground that blocks a new message, a message typed now is answered immediately. Long-running work does not count as blocking once it has been accepted as backgrounded; it stays visible as a dock row.

## Why this issue exists

This territory has been consolidated twice — `SPEC_PROCESS_AND_TURN_STATE_TRACKING_CONSOLIDATION_2026_07_31.md` and `REPORT_AGENT_PANE_ACTIVITY_DOCK_ARCHITECTURE_ANALYSIS_2026_08_25.md` §3.1 — and never got a tracking home. The latter's §5 Tier 3 named the failure mode: of 8 "fix this properly later" notes in this incident family, 1 was ever picked back up. This is that tracked item.

## The correction that keeps getting re-learned

**A dock row does not mean the call was backgrounded.** Promotion at `TOOL_PROMOTION_MS` (30s) is a *display* change — the call is still foreground and the turn is still blocked. AgentMux does not make the backgrounding decision at all; the Claude Code harness does, per call, and AgentMux detects it afterwards by string-matching the tool result (`activity/tool-adapter.ts:57`). `SPEC_AGENT_WORKING_ROW_ABOVE_COMPOSER_2026_09_01` §1 shipped against the wrong version of this and was reverted on 2026-09-09.

## Open items

- [ ] **1. Land the 2026-09-17 policy reversal — currently uncommitted working-tree WIP.** No branch, no PR, no commit. It's the change that makes the invariant true (composer reopens once only dock work remains), and it exists only as modified files in one clone. Highest priority; everything else assumes it.
- [ ] **2. Pre-dispatch auto-backgrounding.** The `PreToolUse` hook exists and `muxspect.mjs` reads `run_in_background` — but only to print a `bg` column. Nothing in the repo ever *writes* it. Making the hook force it for obviously-long commands is what makes the invariant true by construction. Needs a false-positive policy first (`SPEC_FOREGROUND_BACKGROUND_PROCESS_ABSTRACTION_2026_08_20.md` §5.1).
[...]

PR #3340, opened 2026-09-17
fix(agent-pane): backgrounding releases the foreground, and the send path honours it

AgentA-asaf:
Implements §2.3a of `REPORT_AGENT_PANE_PROGRESS_INDICATORS_CONSOLIDATION_2026_09_09.md` (repo-owner-confirmed 2026-09-17): a turn still `Streaming` **only** because work has been accepted as backgrounded no longer counts as busy. The indicator goes dark, the composer reopens, and the dock row remains as the live indicator of the still-running work.

Tracking: #3338 (item 1). The docs consolidation commits ride along here rather than as a standalone docs PR.

## Promotion is still not backgrounding

The carve-out keys off *accepted background launches* (`attachedTask` / `registryAttachedTaskSince`) **plus** the absence of any other blocking tool call. A merely dock-promoted call at `TOOL_PROMOTION_MS` is still synchronously running and correctly keeps the pane busy. That distinction is the one this doc family keeps re-learning — `SPEC_AGENT_WORKING_ROW_ABOVE_COMPOSER_2026_09_01` §1 shipped against the wrong version of it and was reverted. `hasBlockingForegroundToolCall`'s doc comment spells out why the two need separate checks.

## Two gaps in the in-progress state this picks up from

This work existed as uncommitted working-tree changes with no branch or PR. Both halves below were missing:

1. **It didn't typecheck.** `agent-view.tsx` called `hasBlockingForegroundToolCall` without importing it (`tsc` TS2304).
2. **The send path was never updated.** `wasAlreadyWorking` reads the raw turn phase, so a message typed while the indicator was dark still took the HOLD branch and sat in the "send now" queue until the next tool-call boundary — which, for a turn whose only remaining work is a detached dev server, may be minutes away or may never arrive. The indicator promised an immediate answer while the gate silently did the opposite: the exact indicator/gate divergence `working-indicator.ts` exists to prevent, just inverted.

The HOLD branch now flushes immediately when the carve-out applies, using the **same exported predicate** the indicator renders (`turnHeldOnlyByBackgroundWork`) rather than a second inline copy — two independent re-derivations of this predicate is how the row and the bar came to disagree in the first place.

## Design notes for review

- Delivery goes through `flushHeldMessages` rather than inline. The code below that branch is the "pane is idle, initiating a new turn" path and assumes exactly that; `flushHeldMessages` already delivers mid-turn with `initiatesTurn=false` and is documented safe to call during an active turn. It's awaited, not fire-and-forget, so `sendMessage` still resolves after the message has gone out.
- `wasAlreadyWorking` is deliberately unchanged — it still gates optimistic `TurnStart` and auth bookkeeping, and redefining it would regress `Streaming -> Submitting`.
- The carve-out is `Streaming`-only. `Submitting` has no tool-call bookkeeping yet to consult, and `Interrupting` is already mid-stop (steering a message in would race the interrupt).

## Verification

- `tsc --noEmit` clean (was TS2304 before).
- `working-indicator.test.ts` — 13/13 pass, including the four §2.3a cases.
- **The two new `useAgentCommands` tests could not be run locally.** That suite fails to *load* in this clone with `ERR_INVALID_ARG_VALUE: ... Received 'file:///@solid-refresh'`. This reproduces on clean `HEAD` with every change stashed, and hits 47 suites under `frontend/app/view/agent/` — it is pre-existing and unrelated to this change, but it does mean CI is the first real run of those two tests. Please don't read the local pass count as covering them.


a5af:
@codex review

AgentA-asaf:
Fixed in `465c360` — thanks, the P1 was real and I'd mis-described the original in the PR body.

**[P1] compacting/reconnecting dropped from the gate — fixed.** You're right that a subset of a predicate is a second definition of it. I'd claimed the send path used "the same exported predicate the indicator renders"; it used `turnHeldOnlyByBackgroundWork`, which is only the turn-busy half, so `compacting`/`reconnecting` (which `paneBusyForInput` ORs in on top) were silently dropped. Your reachability analysis is also right about why it matters here specifically: the flush is gated on there being no other blocking tool call, which is exactly when a compaction is the only thing running.

The flush now calls `paneBusyForInput` itself, so its condition is literally "the indicator is dark" rather than a re-derivation of part of it. `agent-view.tsx` threads `showingLaunchActivity` through so the hook has the whole predicate available.

On the new opt being optional-defaulting-to-`false`: `false` can only ever make the gate look **more** busy, never less, because the flush is additionally gated on `Streaming` with accepted background work — a state a launching pane is never in. Reasoning is in the opt's doc comment so it doesn't get "simplified" later.

Added `still holds while a compaction is in flight, even with nothing else blocking` to `useAgentCommands.test.ts` to pin it.

Also in this push, fixing the `doc status + grep gates` CI failure: bannering 22 existing docs pulled them into that gate's changed-files scope and exposed eight pre-existing free-text `Status:` lines predating the controlled vocabulary. Each now leads with a vocabulary term and keeps its original prose; the one marked `superseded` got the required `Superseded-by:` pointer. Specs index regenerated.

**One caveat worth your attention:** the three `useAgentCommands` tests in this PR — including the new compaction regression test — **have never run**. That suite fails to *load* in a clean checkout with `ERR_INVALID_ARG_VALUE: ... 'file:///@solid-refresh'`, reproducing on `main` with everything stashed, across 47 suites under `frontend/app/view/agent/`. Locally verified here is `tsc` clean, `working-indicator.test.ts` 13/13, and all three doc gates. CI is the first real execution of those three tests.

PR #3341, opened 2026-09-17
refactor(panes): one pane chrome — part 1 refinements

Agent4-asaf:
## Summary

Follow-up to #3335. There is now **one** pane chrome. `AgentPaneChrome` and `TermPaneChrome` are gone; agent and terminal become the first two implementers of a trait-like capability interface, alongside the nine widget types that already used the shared chrome with pure defaults.

Repo-owner framing: *"all panes should be generic … there should be no such thing as a non-generic pane"* — and the capabilities are genuinely generic, not agent escape hatches. `renderBelowHeader` is the progress-bar slot **any** view type can take (a browser page-load bar, a drone run-progress bar); nothing in the shared chrome is named `agent*`.

### 1. `PaneChromeModel` — the capability surface

Optional `ViewModel.paneChromeModel?: (nodeModel) => PaneChromeModel`. Every field is optional; omit it entirely and a pane gets exactly today's behavior.

- `tabs` / `getId` / `getLabel` / `getIcon` / `getTooltip` / `getAttention` / `getTabClass` — replaces the default `blockStack`-derived tab list, e.g. for a type whose tabs can live in **other** panes
- `onActivate` / `onClose` — return `true` to mean "handled", so the default same-pane stack switch/close is skipped
- `onTabDoubleClick` / `renderLabel` — inline rename
- `renderBelowHeader` — generic slot between header and content
- `wrapContent` — for a surface spanning more than the content box
- `zoomFactor`, `connBtnRef`, `changeConnModalAtom`, `rootClass`, `contentClass`

The tab fields mirror `PaneHeaderTabStripProps` deliberately: the chrome merges what a model supplies over its own defaults and hands the result straight to the strip, so there's one tab-rendering contract rather than a parallel one per pane type.

### 2. The collapse

Both former components turned out to be ~300–500 lines of **tab model**, not chrome — their actual render was the shared markup plus one slot each. So the render collapses and the model is just a returned object (`buildAgentPaneChromeModel` / `buildTermPaneChromeModel`).

- agent's turn-progress bar → `renderBelowHeader`
- terminal's background image + runtime badge → `wrapContent`
- both return `true` from `onActivate`/`onClose` — agent's tabs may live in another pane (cross-pane forks), terminal needs focus restored to the newly-active xterm

`rootClass`/`contentClass` let both keep the CSS hooks their stylesheets already target — notably `.term-pane-stack-content`, which is load-bearing for `pane-leaf-chrome`'s keep-alive positioning — so **this lands with zero stylesheet churn**. Their now-dead `-focused` variants are superseded by the shared chrome's and can be swept in the rename pass.

The `createComponent()` dance both ViewModels used existed to give the chrome component the right reactive owner; the shared chrome calls `paneChromeModel` in its own scope, which is that same ownership by construction.

### 3. "Agent", not "New Agent"

Repo-owner call. #3335 aligned pane title and tab label on "New Agent"; they should be aligned on "Agent" instead. Both fallbacks now read the same; a launched agent still shows its real name.

## Follow-up

The `Generic*` → `Pane*` rename (`GenericPaneChrome` → `PaneChrome`, `.generic-pane-stack` → `.pane-stack`) is now genuinely correct to do — with one chrome, "Generic" names a distinction that no longer exists. Held as a separate mechanical, zero-behavior-change diff so this one stays reviewable.

## Test plan
- [x] `GenericPaneChrome.test.tsx` — 19 tests, 8 new covering the capability surface (including that a view type opting out keeps today's behavior exactly)
- [x] Full sweep across `element`/`view/agent`/`view/term`/`tab`: exactly the 12 known pre-existing Node 24/`solid-refresh` failures, no new ones
- [x] `task dev` boots clean on Windows with zero frontend runtime errors
- [ ] **Needs hands-on verification** — I can screenshot but not click. The collapse touches agent's progress bar, cross-pane fork tabs, and rename; and terminal's keep-alive, background, and connection button. Those want a real click-through before merge.


a5af:
@codex review

Agent4-asaf:
Rebased onto latest `main` (picks up #3334 / #3337 / #3333), plus two additions since the PR was opened:

- **"Agent", not "New Agent"** — repo-owner call, aligning the fresh-pane title and the `+`-created tab label on `"Agent"`.
- **Dropped the tab strip's doubled bottom border inside the pane header.** `.pane-tab-strip` has always drawn its own `border-bottom`, correct when it was a separate row *below* the header. Now that the strip *is* the header, the header row already draws a full-width one — and since the strip is shrink-to-fit while the header is full-width, the strip's border rendered as a second, short line hanging under just the tabs and the `+`. Reported live; most visible on a single-tab pane, where the strip is only the `+` button. Scoped to `.block-frame-default-header .pane-tab-strip` rather than removed outright, since the editor's file-tab strip and the history strip still render `PaneTabStrip` as their own standalone row where that border is the only separator.

Verified after the rebase: typecheck clean, 41 tests green across the chrome + term suites, and `task dev` boots on Windows with zero frontend runtime errors. The SCSS fix was confirmed live via HMR.

Landing this as **part 1**; the `Generic*` → `Pane*` rename and further refinements follow in part 2.

Agent4-asaf:
Thanks — all four P2s addressed:

- **`createComponent` unused in `agent-model.ts` / `termViewModel.ts`** — fixed here in 6bd3676f. Correct catch: both used it solely to instantiate their own chrome component with the right reactive owner, and that call is gone now that they register the shared chrome. The import survived only because the comments explaining why the dance *used* to be necessary still name it.

- **Dead `.agent-pane-stack-focused*` / `.term-pane-stack-focused*` selectors** — also correct, and already fixed in the follow-up branch (`agent4/pane-tabs-part2`) rather than here, since I'd spotted the same thing while sweeping what this PR's collapse commit had deferred to the rename pass. Worth noting the finding slightly understates it: the *base* `.agent-pane-stack::after` / `.term-pane-stack::after` rules were dead weight too — not unmatchable like the `-focused` variants, but byte-identical duplicates of the shared `.pane-stack::after`, since the chrome puts `pane-stack` on the same root via `PaneChromeModel.rootClass`. The follow-up removes all six rules, having diffed them to confirm no visual change.

Agent4-asaf:
Both dead-selector P2s are now fixed **in this PR** (0fee40c4) rather than the follow-up branch — you're right that they belong here: this PR's collapse is what orphaned them, and its own commit message said they'd be "swept in the rename pass," which is also in this PR. Moved the sweep in.

It removes six rules, not four. Alongside the four dead `-focused`/`-focused-alone` selectors you flagged, the *base* `.agent-pane-stack::after` / `.term-pane-stack::after` rules went too: those weren't unmatchable, they were byte-identical duplicates of the shared `.pane-stack::after`, since the chrome puts `pane-stack` on the same root as `rootClass`. Diffed all three implementations before removing to confirm no visual change, and verified live via HMR in a running dev build.

PR #3342, opened 2026-09-17
Manoz@Area54: generate the shell/agent-input wire types, finishing three more handler files

GenericAgentX-asaf:
Seven commands: `shell_handlers.rs` (3), `agent_handlers/input.rs` (3), and
`writeagentconfig` — the last untyped registration in `editor_handlers.rs`, which lives
there but is stubbed in `agent.ts`, so it belongs with this batch rather than with the
editor slices. **All three files now have zero `register_handler` calls.**

## `exit_code` is `number | null`, not `number | undefined`

`shellstop` and `shellstatus` answered inline `json!({..})` literals, so the registry
would have recorded `Value` for both. Writing the result types down surfaced a real
drift in `shellstatus`:

The hand-written stub said `exit_code?: number`. But **both** arms of the handler write
the key unconditionally — the unknown arm literally wrote `"exit_code": null`. So the
honest type is `number | null`.

That distinction is load-bearing here specifically. `known: false` must stay
distinguishable from `running: false` — collapsing them is the reagent P1 on #2770, where
a live, still-registering shell got misreported as failed for its entire run. A caller
reaching for `exit_code === undefined` as a proxy for "unknown" would have got the wrong
answer, and the old type actively suggested that reading. There's a test pinning that the
key is present-and-null.

## Smaller finds

`CommandWriteAgentConfigResult` already existed in `rpc_types/block.rs` and was **unused**
— the handler built its own `json!({ "working_dir": .. })` instead. It's the response type
now.

Four derived input types (`SubprocessSpawnInput`, `AgentStopInput`, `ShellExecInput`,
`WriteAgentConfigInput`) restore optionality ts-rs can't express on `#[serde(default)]`
non-`Option` fields — between them a `Vec`, a `HashMap`, two `String`s and two `bool`s.

The watcher schema test from #3337 asserted `writeagentconfig` was *absent* from the
registry. Inverted to assert its recorded types, since `editor_handlers.rs` is finished
rather than one-command-pending.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3923 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 271 generated types current

One note on a flake I chased so you don't have to:
`backend::sysinfo::handle_anomaly_tests::refresh_processes_specifics_does_not_leak_a_handle_per_call`
failed on one local run. I re-ran it against a **stashed (clean) tree** and it failed
there too, then passed on the next full run — it's a pre-existing handle-count flake on
this machine, not anything in this branch.

agent_id: manoz-0803a


a5af:
@codex review

GenericAgentX-asaf:
@reagentx-workflow the last re-review errored before producing findings ("Review failed — no findings produced … Claude CLI exited 1"), so this PR is currently unreviewed on its latest commit. Requesting a retry.

The one open item from your previous pass — the stale doc comment at `editor_handlers.rs:962` claiming `writeagentconfig` is the only untyped registration in the file — is fixed in 1d2b931, along with a second instance of the same staleness a few lines down (the `register_typed_records_the_editor_read_commands` doc comment made the same claim about the four watchers). All checks are green.

GenericAgentX-asaf:
@reagentx-workflow the reviewer account should have quota again — requesting a re-review. The previous attempts on this PR failed with `Claude CLI exited 1` (no findings produced), which was an account quota exhaustion rather than anything in the diff. All checks are green and the earlier findings are addressed.

PR #3343, opened 2026-09-17
Opaz@charlie: fix(build): vim freezes the terminal — esbuild miscompiled xterm's DECRQM handler

GenericAgentX-asaf:
## Summary

Running `vim` in a terminal pane froze it — no rendering, no input. Reproduced reliably in packaged builds (0.56.3, 0.56.4) and **never** under `task dev`. That split is what let it survive several investigations.

**Root cause: a minifier miscompile, not a source bug.**

`vite.config.ts` targeted `safari13`. Safari 13 predates logical assignment, so esbuild lowers `a ||= {}` — and on xterm.js 6.0.0's `requestMode` it lowered it wrong:

```js
source:  requestMode(e,i){ let r; ((P)=>(…))(r ||= {});
shipped: requestMode(t,n){ ((p)=>(…))(void 0 || (r = {}));
```

It constant-folds the read of `r` (always `undefined` from `let r;`) to `void 0`, **drops the declaration as unused**, and **keeps the write**. ES modules are strict, so the orphaned assignment throws `ReferenceError: r is not defined`.

## Why that froze the pane

`requestMode` is the **DECRQM** handler — the bundle registers it as `registerCsiHandler({prefix:"?", intermediates:"$", final:"p"})`. So the throw fires inside xterm's escape-sequence parser. Captured live over CDP:

```
requestMode 189:103840   ← throws
(anon)      189:78048
parse       189:68196
parse       189:83539
```

The exception aborts `parse` mid-buffer, so the rest of the chunk is never processed and the terminal is left mid-state. A shell prompt never sends DECRQM; full-screen programs negotiating terminal modes do — hence "vim breaks the terminal".

## This is a known upstream bug — we were one patch behind

[evanw/esbuild#4508](https://github.com/evanw/esbuild/issues/4508), fixed in **0.28.2**. Upstream's explanation matches exactly: lowering duplicates the LHS, the duplicate wasn't counted as a usage, so the minifier inlined the initializer and dropped the declaration. Vite 6.4.3 resolves `esbuild ^0.25.0`; we had **0.28.1**.

```
esbuild 0.28.1 --target=safari13  →  void 0||(t={})       declaration dropped
esbuild 0.28.2 --target=safari13  →  let E; … E||(E={})   correct
```

## Changes

1. **Drop `safari13`** — a leftover from the Tauri/WebKitGTK era. The frontend has rendered in CEF on every platform since; every remaining WebKit reference in the tree is a historical comment.
2. **Pin `esbuild >=0.28.2`** — fixes the miscompile itself, covering any other lowering path that hits the same flaw. The two fixes are independent; either alone is sufficient for this bug.
3. **Harden the failure class** — `frontend/app/view/term/parser-wedge.ts`. Any future throw inside `parse()` wedges a pane the same silent way. Detection is by **liveness**, deliberately not by sniffing `error.stack` for xterm frames — stack sniffing is unreliable against minified vendor code, which is the exact condition being survived. `Terminal.write(data, cb)` fires `cb` once the chunk is parsed, so writes handed over and never acknowledged mean the parser stopped consuming, whatever the cause. On trip, `termwrap` resets the terminal and calls `resyncController()` — the same path a crashed controller already takes, so this is a new trigger, not a new mechanism. Rate-limited, and logged at error level: a pane that silently recovers still hides a bug.

## Test plan

- [x] Minimal esbuild repro isolating the miscompile; `safari13` broken, `es2021`/`chrome97`/`es2022` correct
- [x] Executed the miscompiled output — `ReferenceError: t is not defined`
- [x] Rebuilt bundle: `requestMode(t,n){let r;…` restored; orphaned `void 0||(x={})` absent from every chunk
- [x] Extracted the built **AppImage** and verified the same in the shipped artifact
- [x] Launched it: **0** `ReferenceError` vs **2** in the unfixed 0.56.4 running beside it
- [x] **Repo owner confirmed vim works in the fixed build**
- [x] 7 new `parser-wedge` tests; term module 29 passing; `tsc --noEmit` clean

## Why it hid so long

The bug is **not in the source** — two independent source-level audits of the 0.56.2→0.56.3 delta cleared it, correctly. The build introduces it. And packaging strips the very source maps `vite.config.ts` emits for the runtime resolver (`stage-linux-runtime.sh` deletes `*.map`), so all the renderer could report was `ReferenceError: r is not defined` at `index-<hash>.js (190)` with a 404 for the map. Diagnosis came from symbolicating against the map the build left in `dist/`.

Full write-up incl. follow-ups: `docs/retro/retro-xterm-requestmode-minify-freeze-2026-09-17.md`


a5af:
@codex review

PR #3344, opened 2026-09-17
Manoz@Area54: generate the cli/toolchain/widget wire types, typing two loose-input handlers

GenericAgentX-asaf:
`cli_handlers.rs` had no typed registrations. It has seven now and none left.

## This revises a rule the earlier slices set

Those concluded that a handler which *deliberately* accepts loose input had to stay on
`register_handler` entirely. `widget.health` and `widget.api` are exactly that case — they
read their payload field-by-field off a `serde_json::Value` and answer a bad port or path
with `ok: false` rather than an RPC error, so a struct request would turn those answers
into deserialize failures.

But that conclusion was stronger than necessary. **`Value` deserializes from anything**, so
registering it *as* the request type is exactly as tolerant as the untyped handler, while
still recording the response type — which is the half that was actually drifting. Both
widget commands do that now, and a test pins the property the trick rests on.

Same treatment is available for `bundle.validate` and `listagents`, which I left on
`register_handler` earlier for this reason. Happy to follow up if you'd like them
converted; I didn't want to touch them inside an unrelated slice.

## Third sighting of the `ts(rename)` trap

`ToolchainEnvResult.pathSource` is camelCase on the wire while every other field in that
response is lowercase. It needs **both** `#[serde(rename)]` and `#[ts(rename)]` — ts-rs
doesn't read the serde one, so without the second the generated key is `path_source`: a key
nothing sends and nothing reads, with every gate still green.

That's now three (`oauth_config_dir` on the identity slice, `FlowNode.type` on the drone
slice, this). It has a test here too.

Amusing small thing: my first version of that test asserted `!ts.contains("path_source")`
and failed on a *correct* binding — it was matching the doc comment that explains why the
rename exists. It matches the declaration now.

## A required field four callers were omitting

`CommandResolveCliData.block_id` is `#[serde(default)]`, which ts-rs can't mark optional,
and four call sites omit it. `tsc` caught all four the moment the ambient global was
deleted. The stub derives `ResolveCliInput` to restore the optionality.

The two install-command fields are *also* `serde(default)` but stay **required**, matching
the hand-written type: every caller passes them, and a resolve that silently falls back to
an empty install command is a worse failure than a compile error.

## The ambient-duplicate scan earned its keep again

Six ambient globals (`CommandResolveCliData`, `ResolveCliResult`, `CommandCheckCliAuthData`,
`CheckCliAuthResult`, `CommandRunCliLoginData`, `RunCliLoginResult`) would otherwise have
shipped alongside their generated replacements, with consumers silently resolving to the
stale ones.

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3931 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 276 generated types current

agent_id: manoz-0803a


a5af:
@codex review

PR #3345, opened 2026-09-17
Manoz@Area54: correct a wrong claim about ts-rs and serde(rename_all)

GenericAgentX-asaf:
Correcting something I got wrong and merged.

#3329 added `#[ts(rename_all = "snake_case")]` to `BlockKind` and `RunStatus` with a
comment asserting *"ts-rs does not read `serde(rename_all)` on variants"*. **That claim is
false**, and I never tested it — I generalised it from a case I *had* tested, where ts-rs
ignored a per-field `#[serde(rename = "type")]` on `FlowNode.node_type`.

ts-rs honours `serde(rename_all)` on both structs and enums. Two counter-examples were
already in the tree, neither carrying any `ts` attribute:

| type | attribute | generated |
|---|---|---|
| `ToolStatus` | `rename_all = "snake_case"` | `"installed_system" \| "installed_bundled" \| …` |
| `InstallStartReq` | `rename_all = "camelCase"` | `{ providerId: string, cliCommand: string, … }` |

So the three `ts(rename_all)` attributes were only ever agreeing with the serde ones.
Removed — and **the generated output is byte-identical**, which is the proof they were
inert. `git status frontend/types/rpc/` is empty after regenerating.

The existing test is what makes the removal safe: it asserts the generated union against
what serde *actually writes*, not against a second attribute that was duplicating the
first. Its doc comment is corrected to say so.

## What stays

The real limitation is narrower and genuinely load-bearing: **ts-rs ignores per-field and
per-variant `#[serde(rename = "...")]`.** That is the `oauth_config_dir` (#3320),
`FlowNode.type` (#3329) and `pathSource` (#3344) bug — three occurrences, all caught by
reading generated output rather than by any gate. Those three keep their `#[ts(rename)]`
and their tests.

I've also written a scan for it (`serde(rename =)` on a `ts_rs::TS` type without a matching
`ts(rename =)`), which is clean across the tree. It was matching `rename_all` in its first
version and reported 44 false positives — which is how I noticed the original claim was
wrong.

## Verification

- `cargo test -p agentmux-srv drone`: 72 passed, 0 failed
- generated bindings byte-identical before/after (`git status frontend/types/rpc/` empty)

agent_id: manoz-0803a


a5af:
@codex review

PR #3346, opened 2026-09-17
chore(providers): bump five CLI pins to current upstream

AgentA-asaf:
Drift check against the npm registry, 2026-09-17. Five of the ten pinned providers were behind.

| provider | was | now |
|---|---|---|
| claude | 2.1.263 | **2.1.274** |
| gemini | 0.58.0 | **0.60.0** |
| qwen | 0.23.0 | **0.24.0** |
| openclaw | 2026.9.2 | **2026.9.4** |
| copilot | 1.0.83 | **1.0.85** |

codex (`0.154.0`) and pi (`0.73.1`) were already current. muxcode, antigravity and kimi have no published package to compare against.

13 edits, all pin sites: `catalog.ts`, `agentmux-srv/providers.rs`, the cef installer constants, the `container-image.yml` workflow default, and the Dockerfile `ARG`.

## Deliberately not touched

**Model catalog labels.** Per `SPEC_DEPENDENCY_UPGRADE_PROCESS_2026_08_27` §1, those are a semantic claim about what a stable alias currently resolves to — not a version string, and no regex can verify them. They need their own verification pass rather than riding along with a mechanical pin bump.

## How this was found

While specing toolchain version-drift awareness (repo owner asked for the pane to surface being behind). The finding that prompted it: **nothing in the app compares an installed version against its pin.** `toolchain-capabilities.ts` captures `version` per tool and `catalog.ts` carries `pinnedVersion`, but grep for `outdated|behind|isStale|compareVersion|semver` across both returns zero hits — so this drift was invisible until checked by hand.

Machine-side on the author's box it was worse than the pins suggested: gemini was installed at `0.45.0` (15 minors behind even the old pin) and codex was not installed at all. Both fixed locally.

## Verification

- `pin-consistency.test.ts` — 6/6 green (this is the test that mechanically enforces all five locations agree)
- `tsc --noEmit` clean


a5af:
@codex review

PR #3347, opened 2026-09-17
feat(toolchain): surface version drift against the pin, automatically

AgentA-asaf:
Repo owner: *"the toolchain pane should let me know if I am behind on any versions."*

The pane could already fetch latest-from-npm and render an "X available" pill. Three things stopped it answering that question.

## What was wrong

1. **It was manual.** You had to click *Check latest versions*. Drift was only visible if you already suspected it — backwards for the one question this pane exists to answer.
2. **It compared by string equality** — `row.version === row.latestVersion`. That misorders `2.1.9` vs `2.1.10`, and `node --version` prints `v22.22.2` while `gemini --version` prints `0.60.0`, so a bare `v` prefix alone produced a permanent false "update available".
3. **It never consulted `pinnedVersion`.** It compared installed against *upstream* only. So the gap that actually breaks agents — running older than the version AgentMux validated — was invisible, while the pane actively nudged users **ahead** of the pin onto something we haven't tested.

## The design point

Two axes, kept separate on purpose:

| comparison | whose problem | rendering |
|---|---|---|
| installed vs **pinned** | the **user's** — one click to fix | loud |
| pinned vs **upstream** | **ours** — they can't fix it | quiet, never a call to action |

Collapsing these into one "update available" boolean is the bug being fixed. The header count counts `behind-pin` only, so it can never nag about something outside the user's control. No pin (every system tool) resolves to `unknown`, **not** `current` — the pane must not look reassuring about a comparison it never made.

## Concrete case

On the author's machine: claude `2.1.218` installed against a `2.1.274` pin, gemini `0.45.0` against `0.58.0`. Both silent before this.

## Notes for review

- Upstream lookup is cached 6h at module level and fired unawaited on mount. Rows render from the purely-local installed-vs-pin comparison immediately; the upstream column fills in later or never. Offline is a normal state, reported as `unknown` currency, never as an error — consistent with `SPEC_PROVIDER_CLI_VERSION_UPGRADE_2026_09_06` §3 on `lookup-failed`.
- The manual button still force-refetches; that's what it's for.
- No backend change. Upgrading already works — `install_cli` does `npm install --prefix <dir> <pkg>@<pinnedVersion>`, so re-running it against a newer pin *is* the upgrade. This PR makes the need for it visible.

## Verification

- 17 new unit tests in `version-drift.test.ts`, including the `2.1.9`/`2.1.10` ordering case and the `v`-prefix case
- providers suite 166/166 across 15 files
- `tsc --noEmit` clean

Refs #3338. Spec: `SPEC_TOOLCHAIN_VERSION_DRIFT_AWARENESS_2026_09_17` (held local per the no-specs-in-PRs norm).


a5af:
@codex review

PR #3348, opened 2026-09-17
Manoz@Area54: generate the websocket wire types (ten commands, the highest-traffic surface)

GenericAgentX-asaf:
`websocket.rs` holds the highest-traffic RPCs in the app — every pane subscribes, every
terminal keystroke is a `controllerinput` — and none of them were typed. **Ten are now.**
Seven remain (`setmeta`/`getmeta`/`createsubblock`/`controllerresync` and friends), which
drag in `ORef`, `MetaMapType` and `BlockDef` and want their own slice.

## Two drift findings

**`MuxInfoData` promised five fields; the handler sends one.** The hand-written type
declared `clientid`, `buildtime`, `configdir` and `datadir` alongside `version`.
`waveinfo` has only ever returned `{ version }`. Nothing reads the other four — the
command has no caller at all today — so pointing the stub at the generated `AppInfoResult`
is a correction, not a narrowing.

**`CommandBlockInputData` was missing `seq`.** The Rust has had `seq: Option<u64>` since
the seq-based input-ordering work; the hand-written type never listed it, so no TypeScript
caller could pass it without a cast.

## It also corrects an assumption the earlier slices were built on

Those added empty structs for no-argument commands on the reasoning that *"the RPC client
sends `{}`"*. On this connection every no-argument stub sends **`null`** instead:

```ts
return client.rpcCall("waveinfo", null, opts);
```

…and four more. `Option<NoArgsReq>` takes both encodings, and there's a test asserting the
bare struct really does reject `null` — so the `Option` is load-bearing rather than
decoration. Worth knowing before anyone reaches for the bare-struct pattern again.

`eventunsub` takes a **bare string** payload, the only command in the app shaped that way.
`Option<String>` types it while keeping the old `data.as_str().unwrap_or("")` tolerance.

## Three responses deliberately stay `Value`

- `getfullconfig` redacts on the *serialized* form, so what goes on the wire is `FullConfig`
  **minus** fields a generated type would still claim are there. Generating it would promise
  the renderer secrets it will never receive.
- `routeannounce` is fire-and-forget, may arrive with no reqid, and only logs its payload.
- `termsize` and `answers` keep the richer hand-written TS via `#[ts(type = ...)]` rather
  than degrading to `any` — spelled inline rather than as `TermSize`, because `#[ts(type)]`
  bypasses ts-rs dependency tracking and a named reference would emit an import resolving to
  nothing (the `DroneRun.block_states` trap from #3329).

## Verification

- `cargo test -p agentmux-srv` (unfiltered): 3931 + 5 + 7 passed, 0 failed
- `npx tsc --noEmit`: clean
- `scripts/check-rpc-bindings.sh`: ok, 275 generated types current
- all four standing scans clean: ambient-duplicate, dead-import, hand-serialized-response,
  and the new `serde(rename)`-without-`ts(rename)` scan from #3345

agent_id: manoz-0803a


a5af:
@codex review

PR #3349, opened 2026-09-17
Manoz@Area54: docs: close out the RPC codegen effort, with a correction and an honest ledger

GenericAgentX-asaf:
Closes out the RPC codegen effort (§5.4a) with its final state, and retracts a claim this
report itself repeated.

## Metrics

| metric | at report creation | now |
|---|---|---|
| `register_typed` vs `register_handler` | 12 / 232 | **170 / 66** |
| `srv-types.d.ts` | 2,755 lines | **1,698** |
| generated bindings | 14 | **280** |
| stub files importing generated types | 1 of 16 | **15 of 18** |

72% of registrations typed, from 5%. (#3342 and #3348 were still in review when measured,
so this understates by ~17 commands.)

## §5.4e — a correction

This report said ts-rs ignores `#[serde(rename_all)]`. **It does not.** That was
generalised, untested, from the per-field `#[serde(rename)]` case — which *is* real and has
now bitten three times (`oauth_config_dir`, `FlowNode.type`, `pathSource`), each time
producing a key nothing sends and nothing reads with every gate green.

#3345 removed the redundant `ts(rename_all)` attributes and the generated output was
**byte-identical** — the proof they were inert.

Method note worth keeping: the scan I wrote for this was too broad on its first run and
reported 44 false positives. Investigating them is what surfaced the error. A noisy check
that gets read beats a quiet one that gets trusted.

## §5.4c — a rule revised, not restated

It said a handler that deliberately accepts loose input must stay untyped entirely. That
was stronger than necessary: `Value` deserializes from anything, so registering it *as* the
request type is exactly as tolerant while still recording the response — the half that
actually drifts.

## §5.4f — "no argument" has two encodings and the stubs disagree

Most stubs send `{}`; every no-argument stub on the websocket connection sends `null`. An
empty-struct `Req` rejects the second, a unit `()` rejects the first. `Option<Req>` takes
both.

## §5.4g — honest ledger

Three of the most-cited findings from this effort were bugs **the migration itself nearly
introduced**, caught before merge. That is risk paid down, not value delivered, and the
report shouldn't claim otherwise.

What is unambiguously banked: 65 dead stubs deleted, 11 called-but-unregistered commands
surfaced (still the most actionable item in this document and still nobody's task), and
five type lies corrected. Most were wrong in the *safe* direction — optional where the
server always sends — which is why few produced live crashes.

## §5.4h — stopping point

Stopped at 72% deliberately. Records what remains, and why it's the lowest value per unit
of churn: every slice touches the same three shared files, so the rebase tax is constant
regardless of slice size.

agent_id: manoz-0803a


a5af:
@codex review

a5af:
@codex review

GenericAgentX-asaf:
@reagentx-workflow the re-review on the merge commit errored before producing findings ("Review failed — no findings produced … Claude CLI exited 1"), which dismissed the prior approval without replacing it. Requesting a retry.

Since your last successful pass (which was LGTM), the only change is dce1bb1 — a merge of `origin/main` to unstick the `gen-docs-index.sh --check` gate. Two conflicts, both docs:
- `SPEC_AGENT_PANE_LAYOUT_REDUCER_2026_06_02.md` — took main's Status line verbatim (the one carrying the `(#1270)`/`(#1281)` citations), which is what makes the spec identical to main again and takes it out of the changed-specs gate.
- `REPORT_MIGRATION_WRAPUP_STATUS_2026_09_16.md` — kept this branch's text, i.e. the `listagents` carve-out rationale that resolved your earlier P2.

No `INDEX.md` regeneration was committed: regenerating locally on Windows produces a structurally different file than CI does (differing status-bucket parsing and row ordering), so the merge was the correct fix rather than a hand-generated index. All checks are green.

GenericAgentX-asaf:
@reagentx-workflow the reviewer account should have quota again — requesting a re-review. The previous attempts on this PR failed with `Claude CLI exited 1` (no findings produced), which was an account quota exhaustion rather than anything in the diff. All checks are green and the earlier findings are addressed.

PR #3350, opened 2026-09-17
feat(agent-picker): show an upgrade hint on cards whose CLI is behind the pin

AgentA-asaf:
Follow-up to #3347, which surfaced drift in the Toolchain pane. The picker is where it actually matters — it's the moment you're about to launch an agent on that CLI.

Repo owner: *"we need to put an indicator somewhere, perhaps on the agent's card in the agent pane, letting them know an upgrade is available."*

## No new probe

`install.check` already ran per card to drive the "Click to install" ribbon, so this rides the same round-trip. `InstallCheckReq` gains an optional `npmPackage`; `InstallCheckResult` gains `version`.

**Read from `<install_dir>/node_modules/<pkg>/package.json`, not by running `<cli> --version`.** The picker calls this once per card — spawning a process per card to answer what a file read already answers isn't a trade worth making.

## The restraint is the design

The badge renders for `behind-pin` and nothing else:

| state | why silent |
|---|---|
| `ahead-of-pin` | untested but working — not the user's problem |
| `unknown` | we never made the comparison; must not look reassuring |
| `current` | fine |
| not installed | the install ribbon already owns this card |

It yields entirely to the install ribbon — "not installed" supersedes "out of date", and two CTAs on one card is what that ribbon's own spec warns against. A card is a launch affordance, not a version dashboard; the Toolchain pane is where the full picture lives.

Drift is resolved with the shared `version-drift` predicate from #3347, so the card can't develop its own opinion about what "behind" means.

## Security note

The npm package name becomes a path segment, so it's validated against `..`, absolute prefixes and NULs **before** any filesystem access. Scoped names (`@anthropic-ai/claude-code`) are legitimate and must *not* trip that check — both directions are pinned by tests. Version is only reported when the binary actually resolves, so a stale `package.json` beside a missing binary can't claim a version for a provider that cannot launch.

## Verification

- 2 new Rust tests — the validation branch returns before touching the filesystem, so they hold with no data dir present
- frontend providers suite 166/166
- `tsc --noEmit` clean (also caught two `InstallCheckCommand` mocks needing the new field)

**The 3 new `AgentCard` component tests did not run locally.** `AgentCard.test.tsx` is one of 47 suites under `frontend/app/view/agent/` that fail to *load* in a clean checkout with `ERR_INVALID_ARG_VALUE: ... 'file:///@solid-refresh'`, reproducing on `main` with no local changes. CI is their first execution — please don't read the local pass counts as covering them.

Refs #3338.


a5af:
@codex review

AgentA-asaf:
Fixed in `b7bad86` — the P1 was real and the diagnosis was exactly right.

**[P1] drive-letter escape — fixed by calling the existing helper.** `PathBuf::push` replaces the path when the pushed component carries a drive prefix, so `npm_package = "C:\Users\Victim\AppData"` walked straight out of `provider_install_dir`. My hand-rolled check only looked at `..`, a leading `/` and a leading `\`.

The sharper part of the finding is that `safe_join_within_base` already existed, and its own doc comment spells out why `starts_with('/')` / `starts_with('\')` is insufficient on Windows. I reimplemented it badly instead of calling it — the same write-a-second-copy failure this PR's frontend half was deliberately avoiding by reusing the shared `version-drift` predicate. It now calls the real one, which additionally covers UNC roots and the drive-*relative* form (`C:payload.txt`) my version never considered at all.

Tests grew the four missing cases — `C:\Users\...`, `C:/some/dir`, `C:payload.txt`, `\server\share\evil` — listed explicitly so a future refactor back to a hand-rolled check fails here rather than shipping.

**Also fixed the vitest failure** (unrelated to the P1): `AgentPicker.test.tsx:509` asserts the exact `install.check` argument object, which now carries `npmPackage`. The assertion's point is unchanged — every field must resolve from the effective launch provider, never the drifted column. That test lives in the suite I flagged as locally unrunnable, so CI caught it first, exactly as expected.

Verified: 2 Rust tests green including all four drive-letter forms; `tsc` clean; `cargo check` clean.

AgentA-asaf:
Fixed in `1894b98`. Correct catch, and I'd misread the evidence.

**[P1] the mock was wrong, not the assertion.** `ProviderDefinition.npmPackage` is `string` (`providers/types.ts:64`), so the mock's `["@openai/codex-cli"]` describes a shape production never produces. I saw that array in the first CI diff and mistook it for vitest's formatting, then asserted the string against it.

Why it survived this long: the mock's only consumer was `!prov.npmPackage || prov.npmPackage.length === 0`, which a 1-element array and a non-empty string satisfy *identically*. The moment the value was passed through to an RPC instead of just length-checked, the lie surfaced. Fixed the mock; the assertion stays a string, matching the real type.

**Also correcting something I claimed twice in this PR.** I said the 47 unloadable suites under `frontend/app/view/agent/` were a repo-wide problem reproducing on `main`. They are not — CI runs all of them (276 files pass). The cause is local: **CI uses Node 24 (`ci-pr.yml:267`), this machine runs Node 22**, and the `@solid-refresh` resolver error doesn't reproduce under 24. A clean `npm install` doesn't fix it; matching the Node version should.

You were right to flag that this test "was never actually executed before being pushed" — that's exactly what happened, twice. The root cause turns out to be a toolchain version gap silently costing local coverage, with nothing in the app surfacing it, which is precisely what this PR is about.

Verified: `tsc` clean. `AgentPicker.test.tsx` still can't execute locally for the Node reason above, so CI remains its first real run — but mock and assertion are now type-consistent with production, which is the part that was actually wrong.

PR #3351, opened 2026-09-17
fix(panes): a new terminal tab opens in the current tab's cwd again

Agent4-asaf:
## Summary

Part 2, first piece. Restores the regression I called out deliberately in #3335.

Terminal's old `"+"` handler passed the current shell's `cmd:cwd` to `pane.open`, so a new tab started in the same directory — the way a real terminal's "new tab" does. Making `"+"` the shared widget picker (#3335) dropped that, and new terminal tabs fell back to the default directory.

## How it's restored

Without putting a view-type branch back into the generic path. The destination pane contributes the params, and the picker just forwards whatever it gets:

- New optional `PaneChromeModel.newTabParams(view)` — "given the view type being added to me, what extra `pane.open` params do I contribute?"
- `openPaneTabWidgetPicker` passes the pane's contribution straight through to `addWidgetAsPaneTab`, which spreads it into the `pane.open` call. The picker itself still knows nothing about any view type.
- Terminal returns `{ cwd }` for a `term` tab and `undefined` for anything else — so adding a *browser* tab to a terminal pane is unaffected, and a pane with no `PaneChromeModel` contributes nothing, which is exactly today's behavior.

This is the same shape as the rest of the capability surface from #3341: generic hook, view type opts in.

Also drops the `openPaneTabWidgetPicker` imports `agent-view.tsx` and `term.tsx` kept after the collapse moved `"+"` into the shared chrome — neither calls it any more.

## Test plan
- [x] `PaneChrome.test.tsx` — 2 new tests: `newTabParams` reaches the add path, and contributes nothing for a view type the pane doesn't recognise. One existing assertion updated for the new 4-arg signature (it was pinning the old 3-arg call).
- [x] `layoutStack.test.ts` — 2 new tests for the params passthrough and its omission. Note these can't run on this machine: that file is one of the ~51 hit by the pre-existing Node 24/`solid-refresh` incompatibility (fails identically on a clean checkout), so CI's `vitest` leg is what actually exercises them.
- [ ] Hands-on: open a terminal, `cd` somewhere, add a terminal tab via `"+"` → should start in that directory.


a5af:
@codex review

Agent4-asaf:
Good catch — verified against the backend and it's exactly right. Fixed in 844a7918.

`open_pane` does `match cmd.meta { Some(m) => m, None => build_pane_meta(&cmd)? }` (`app_api/mod.rs`), and `cmd.cwd` is only consumed inside `build_pane_meta`'s `"term"` branch (`app_api/pane.rs`) — which never runs here, because `addWidgetAsPaneTab` always supplies `meta`. So the `cwd` the RPC carried was dropped server-side and the new shell still opened in the default directory. The old pre-#3335 `handleTermTabAdd` worked precisely *because* it sent `cwd` with no `meta` at all.

Since this path must send meta, the key now goes in there: the hook is `newTabMeta` (contributes META keys), `addWidgetAsPaneTab` merges them over the blockdef's own, and terminal writes `cmd:cwd` directly — the same key `build_pane_meta`'s `"term"` branch would have written.

Your point about the tests is the more useful half of the finding, so I rewrote them rather than just adapting them. They asserted the RPC payload *shape*, which is the wrong level — they passed happily while the feature did nothing. They now assert the key lands **inside** `meta`, plus an explicit `expect(sent).not.toHaveProperty("cwd")`, since a top-level arg is exactly the shape srv ignores.

Still flagging honestly: this is verified by construction against the srv source, not yet by hand. The end-to-end check (`cd` somewhere in a terminal, add a terminal tab via "+", confirm it starts there) is in the PR's test plan and remains unticked.

PR #3352, opened 2026-09-17
feat(launcher): quiesce+migrate primitive for the upgrade flow (P1a)

AgentX-asaf:
## Summary

Phase 1a of #3258 (issue tracking the fast-startup redesign). The self-contained, testable half of "quiesce srv, run pending migrations, come back" — **not yet wired into anything live**. See the module's own doc comment for the full reasoning; short version below.

## Why split like this

The spec's §4.3 sequence has two halves with very different risk profiles: the quiesce/migrate logic is a self-contained async function; "suspend the live supervisor loop's respawn/teardown handling" means editing `supervisor/windows.rs`'s and `supervisor/unix.rs`'s 1000+-line `tokio::select!` state machines — exactly the surface CLAUDE.md's I1–I6 isolation invariants exist to protect, and something this repo has no way to exercise end-to-end outside a real multi-process launcher run. Phase 2 (the boot gate + upgrade-only srv mode) hasn't landed yet either, and it very possibly decides whether srv ever becomes a respawn-on-exit child of the normal supervised loop for this path at all — wiring into today's loops now would be guessing at an integration shape phase 2 gets to actually decide. This PR is the callable, tested half; the calling half is explicit follow-up work.

## What's in it

`agentmux-launcher/src/upgrade.rs`:

- **`quiesce_srv(child, timeout)`** — the fix for the exact gap `agentmux-cef/src/commands/backend.rs:223`'s `sleep(300ms)` heuristic left open (that comment already says "so the OS releases file locks before we open the DB," which a sleep can only approximate). Returns only once `Child::wait()` confirms the OS has genuinely reaped the process. This is **Constraint 2** from the spec — migrations only run with no live srv holding the data dir, the exact thing `341faa981`'s data-loss incident violated.
  - Unix: SIGTERM via the already-existing (previously *uncalled*) `terminate_child_gracefully`, wait up to 10s, escalate to a hard kill if it doesn't exit.
  - Windows: `start_kill()` immediately — there's no graceful signal available for a console-less child in this launcher today (every existing srv-teardown call site already uses `start_kill`, `supervisor/windows.rs:574,741,1141`), so this stays consistent with that convention rather than inventing an untested new shutdown mechanism for one caller.
- **`run_migration_upgrade`** wraps `quiesce_srv` + the revived `srv_spawner::run_migrate` (was `#[allow(dead_code)]` with no callers — now has one).

**No snapshot step**, deliberately: `agentmux-srv migrate`'s own `apply_pending` already takes a backup under the migration lock before applying anything (`backup_stores` in `runner.rs`). The spec's original §4.3 step 3 assumed the launcher would re-implement what `bootstrap.rs` does for the in-process daemon path; going through the CLI subcommand instead means that safety property already exists on this path, so a launcher-side snapshot would be redundant, not additive.

## Testing

Real spawned child processes throughout (`sleep`/`ping`/`sh`), matching this crate's own existing process-lifecycle test convention in `host_spawn.rs` rather than mocking `Child`:

- Already-exited child → no-op, no signal sent.
- A live child is genuinely stopped **and reaped** (not just signaled) — the actual guarantee this function exists for.
- Unix-only (`#[cfg(not(windows))]`): a SIGTERM-ignoring child gets force-killed after the timeout; a normal child exits gracefully without ever reaching the force-kill branch.

`cargo check --workspace`: clean. `cargo test -p agentmux-launcher`: 255 passed + 2 new, Windows-verified here (this machine). **2 more tests are Unix-only and will only run on CI's `ubuntu-latest` leg — I could not verify them locally.**


a5af:
@codex review

AgentX-asaf:
Both fixed in \`87994fb\` — thanks, both real.

Also: CI's ubuntu-latest leg caught a genuine flake in my own test (\`a_sigterm_ignoring_child_is_force_killed_after_the_timeout\` — got \`ExitedGracefully\` instead of \`ExitedAfterForceKill\`), exactly the Unix-only coverage I'd flagged I couldn't verify locally. Root cause: I sent SIGTERM immediately after \`spawn()\`, with no guarantee the shell had actually run \`trap '' TERM\` yet — on a loaded runner, SIGTERM can arrive before the trap installs and hits its default disposition instead, killing the shell "gracefully" for the wrong reason. Fixed with a real synchronization point (the shell echoes "ready" right after installing the trap; the test blocks on reading that line before signaling) instead of a delay that would only have narrowed the race.

Still true that I can't verify the two \`#[cfg(not(windows))]\` tests compile/pass locally (no Linux toolchain on this machine) — relying on CI's ubuntu-latest leg again to confirm this fix actually lands the way I intend.

PR #3353, opened 2026-09-17
Clare@starpower: fix(term): flush CPU/mem badge to pane top-right, add toggle

AgentO-asaf:
The terminal pane's live CPU%/mem badge (`BlockStatsBadge`) reserved a fixed `--header-height` offset from the top of `.block-frame-default-inner`, assuming the pane's inline header always renders there. Terminal (and agent) panes render through the shared `pane-leaf-chrome` instead, which hoists the real header ABOVE this component and suppresses the inline one (`TermViewModel.noHeader`) — so that reserved space is phantom, and the badge lands well inside the terminal's content instead of flush against the pane's actual top-right corner.

Also adds a Settings → Terminal toggle ("Show CPU/mem badge", `term:showstatsbadge`, default on) to hide the badge — scoped specifically to terminal panes per explicit request; agent panes keep the prior always-on behavior.

Spec: docs/specs/SPEC_TERM_STATS_BADGE_TOP_OFFSET_AND_TOGGLE_2026_09_17.md

## Changes
- `frontend/app/element/blockstats.tsx` / `.scss` — `noHeader` prop, `.no-header` modifier zeroing the offset
- `frontend/app/block/blockframe.tsx` — threads `noHeader` through; new `showStatsBadge` memo gating the badge for `view === "term"` only
- `frontend/app/view/settings/sections/terminal-section.tsx` — new toggle row
- `frontend/types/srv-types.d.ts`, `schema/settings.json` — new `term:showstatsbadge` key (frontend-only, no Rust field — same precedent as `term:predictiveecho`)
- Changeset (patch)

## Test plan
- [x] `npx tsc -p tsconfig.citypecheck.json --noEmit` — clean
- [x] `npx stylelint frontend/app/element/blockstats.scss` — clean
- [x] `npx vitest run frontend/app/element frontend/app/block frontend/app/view/settings` — 120/120 passing
- [ ] Live `task dev` visual check — **not done this session** (no capturable GUI window available in this environment); the offset fix is traced precisely through the DOM/CSS structure (see spec §2) but a human should eyeball it on the next dev run before merge.


a5af:
@codex review

PR #3354, opened 2026-09-18
Opaz@charlie: feat(tooling): discover the CEF debug port instead of assuming 9222/9223

GenericAgentX-asaf:
## Summary

Agent and test tooling hardcodes the CEF debug port — "9223 dev / 9222 release" — in `bench-agent-keystroke.mjs`, `agent-layout-drift.mjs`, `bench-aggregate.mjs`'s documented invocation, and `docs/linux.md`.

Those are only **preferred** values. `agentmux-cef` binds the preferred port if free and otherwise takes an OS-assigned one — deliberately, because parallel instances are a designed feature (I1–I6) and a fixed port collides. The behaviour is correct; what was missing is any way for a caller to learn which port it got.

With more than one instance up — the normal case here — **the constant names whichever instance won the race**, not the one the caller meant.

## This failed silently, and it happened

Debugging a packaged build, I attached to 9222 believing it was my own test instance. 9222 belonged to the repo owner's live instance, which had started first. Mouse clicks and synthetic keystrokes went into it. **Nothing errored.** I caught it afterwards only by noticing the debugger's page targets were served from a different port than the build under test.

A tool that fails loudly costs a minute. A tool that silently targets the wrong instance corrupts someone else's session and produces results that look valid.

## Why it was undiscoverable

The port was never secret — it's logged at startup (`CEF remote-debugging port: 42149 (preferred 9222)`) and kept in `app_state.debug_port` for `browser_api`. It just wasn't published anywhere a tool reads, even though `authkey.dev` exists for exactly this purpose (`SPEC_TEST_API_ACCESS`, `SPEC_BENCHMARK_PORTABLE_DISCOVERY_2026_05_20`) and already carries the web/ws/ipc endpoints.

## Changes

- **`authkey.dev` publishes `debug_port`.** Port selection moves above the authkey write so the actual bound value is available to write.
- **`tools/tests/lib/instance-discovery.mjs`** enumerates live instances from their own `authkey.dev` (checking `host_pid`), and **refuses when ambiguous**, listing candidates. Guessing is the defect, so the API doesn't offer it.
- **`resolveCdpPort` verifies** the debugger's pages are served by that instance's own web endpoint before handing the port back — a port number alone doesn't prove whose it is.
- An instance predating the field reports `null` and throws with instructions **rather than falling back to 9222. The fallback is the bug.**
- `bench-agent-keystroke.mjs` / `agent-layout-drift.mjs` now discover by default (`--cdp-port` still overrides, `--instance <substring>` selects); `docs/linux.md` corrected.

## Test plan

- [x] 7 `node:test` cases with `root`/`alive` injected, so no running instance needed: dead-instance filtering, port read from file (asserts 42149 **not** 9222), `null` for instances predating the field, **ambiguity throws**, `match` selects, no-match errors with candidates listed, zero-live errors
- [x] `agentmux-cef`: `the_authfile_publishes_the_debug_port` asserts serialization with the real port
- [x] Run live against this machine: correctly enumerated all 4 running instances and reported `cdp=unknown` for each (they predate the field) instead of defaulting

## The pattern

Fourth instance of one defect class found in a single session — a **machine-global constant used as an identity**:

| Resource | Constant | Consequence |
|---|---|---|
| Shell integration | `~/.agentmux/shell/` | versions overwrote each other's rcfiles (#3302) |
| AppImage extraction | `extracted/$VERSION` | a new build silently ran an older binary (#3324) |
| Instance identity in env | inherited `AGENTMUX_*` | a build adopted the launching instance's channel (#3326) |
| CDP debug port | `9222`/`9223` | a tool drove a different instance than intended (this PR) |

**A well-known constant cannot identify an instance** when N are supported by design; it resolves to whoever got there first. Identity must be discovered from per-instance state, verified, and ambiguity must be an error.

Spec: `docs/specs/SPEC_INSTANCE_DISCOVERY_FOR_TOOLING_2026_09_17.md`


a5af:
@codex review

PR #3355, opened 2026-09-18
Opaz@charlie: fix(package): keep source maps in local Linux portables

GenericAgentX-asaf:
## Summary

`scripts/stage-linux-runtime.sh` deleted every `*.map` from the staged runtime **unconditionally** — contradicting both its own comment ("Mirrors the STRIP_MAPS policy for release builds") and `SPEC_PORTABLE_SOURCE_MAPS_2026_06_01.md`:

> - **`task package`** (local portable) — maps **included**
> - **`task package:release`** — maps **stripped**

Windows honours that via `${STRIP_MAPS:-0}` in `scripts/package.sh`. **Linux ignored the variable entirely**, so every local Linux portable shipped without maps.

## Why it matters

`vite.config.ts` sets `sourcemap: true` specifically so `frontend/log/source-map-resolver.ts` can rewrite raw `error.stack` positions into original-file frames before they reach the host log (`SPEC_FE_SOURCE_MAP_RESOLVER_2026_05_27.md` §7.1).

Stripping the maps leaves that resolver with nothing. A packaged build reports a crash as:

```
Uncaught ReferenceError: r is not defined   index-<hash>.js (190)
[source-map] missing or invalid for chunk index-<hash>.js: HTTP 404
```

That is exactly what the vim/DECRQM terminal freeze looked like for most of its investigation. The root cause was eventually found by symbolicating **by hand** against a map recovered from `dist/` — doing the resolver's job manually, because the artifact had been stripped of its inputs. We paid the build cost for maps and then discarded the benefit.

## Changes

- The strip is gated on `STRIP_MAPS`, defaulting to **keep** — matching Windows and the spec.
- **All four Linux release tasks** (AppImage / deb / rpm / tarball) now pass `STRIP_MAPS=1` explicitly. Without this, gating the strip would have silently started shipping ~28 MB of maps in every release — a regression in the opposite direction.

## Test plan

- [x] `scripts/stage-linux-runtime.test.sh` — 5 passing: default keeps, `STRIP_MAPS=0` keeps, `STRIP_MAPS=1` strips, and every Linux release task opts in
- [x] The test extracts the real block from the script so it can't drift from the implementation
- [x] **The first revision of this test was wrong** and I kept it honest: `STAGING_ROOT` was one directory level off, so the script globbed a non-existent path, deleted nothing, and the keep-cases passed for the wrong reason. It now asserts the fixture is findable *before* drawing any conclusion from a count.

## Noted, not changed

`scripts/package-macos.sh` uses `${STRIP_MAPS:-1}` — it strips local builds by default, contrary to the same spec. I've flagged it rather than changing it: it may be a deliberate DMG-size call and deserves its own decision.


a5af:
@codex review

PR #2506, opened 2026-08-10
docs(specs): adopt closed Status vocabulary (docs-lifecycle Phase 1) + restamp 44 code-verified specs

AgentX-asaf:
## Summary

Implements **Phase 1 of `SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md`**: `docs/specs/README.md` now states the closed Status enum (`draft | proposed | active | implemented | living | historical | superseded`) as the actual rule, with a required `**Superseded-by:**` field for superseded specs and a reader guardrail (spot-verify checkable claims — Phase 4's mechanism). `docs/README.md` references it.

Restamps **44 specs** whose real state was verified against code on main (`ad0dbc5a5`) in a 2026-08-10 three-agent sweep. The sweep found staleness in both directions: ~28 shipped specs still said Proposed/Draft, and several "done" specs have unshipped halves (now `active` with the remainder listed and PR evidence cited).

No spec body content changed — only the Status field (and one added Superseded-by). Deliberately not a big-bang restamp of all ~600 specs, per the docs-lifecycle spec's own scoping note: these 44 are the ones with fresh verification evidence.

## Verification

Docs-only. Post-edit script check: every changed spec's Status first word is in the closed enum; the one `superseded` spec carries a valid `Superseded-by:` path; per-file diffs are 2-6 lines.

PR #3356, opened 2026-09-18
Manoz@Area54: docs: cite the implementing PR on 99 specs (R3 burn-down, 105 -> 6)

GenericAgentX-asaf:
`docs/specs/README.md` requires an `implemented` Status to cite the implementing PR(s).
**105 specs did not. 99 do now** — R3 drops 105 → 6, and the lifecycle total goes
473 → 374 (52% → 41% of files).

## How each PR was found — because a wrong citation is worse than none

That's not my principle; it's the README's, about dangling `Superseded-by` pointers. It
applies identically here.

The naive question — *"at which commit is the Status `implemented`?"* — gives false answers
for any file whose git log doesn't start at its creation. It produced
`frontend-log-pipe.md → #128 "fix: focus border stops tracking after tab switch"`.

The right question is where the Status **changed** to `implemented`: compare the file at
each commit against the same file at that commit's parent. Two filters then removed the
citations that would *still* have been wrong:

| filter | rationale | caught |
|---|---|---|
| commit touched ≥5 spec files | that's restamping, not implementing | 5, incl. **#2506** (45 specs, 0 source — the vocabulary sweep) and **#2920** (136 specs) |
| commit changed 0 source files | cannot have implemented anything | 0 of the surviving 98 — the check passing, not the check being useless |

Spot-verified a sample against PR subjects. Three matched their spec exactly. The fourth,
`frontend-log-pipe.md → #128`, still looked wrong — so I opened #128, and it **does**
contain `frontend/log/log-pipe.ts`. The PR bundled two changes and its subject named only
one. Correct citation; my suspicion was not.

## A bug in my own tooling, worth recording

`execSync` shells through **cmd.exe** on Windows, where `^` is the escape character. So
`git rev-parse <sha>^` silently became `git rev-parse <sha>` — "the parent" was the commit
itself, no transition could ever be detected, and **every file reported "no transition"**.
Verified directly: `^` → `92d9ca2071d8` (self), `~1` → `3dfac3492af0` (parent). It cost a
full wrong run and would have cost a silent one if the output had been less obviously
broken.

## The 6 left are deliberate, and say so

One was recoverable: `SPEC_AGENT_PANE_LAYOUT_REDUCER` now cites #1270 (Phase 3) and #1281
(Phase 4), both verified against code today.

Five had candidate PR numbers in their own body. **All were checked and rejected** —
#1780 is "rename Shell→Log" on a spec about a dev window title arg; #2311 is docs-only;
#2024 has no commit at all. Each of the six now carries a note saying the PR is
unrecoverable, why, and that the candidates were already investigated — so the next reader
doesn't repeat the search.

## One dangling citation resolved on the way past

`SPEC_BENCHMARK_PORTABLE_DISCOVERY` cited `SPEC_DEAD_TERMINAL_PANE_2026_05_20.md` three
times. That spec was real (added in `62d50cb`) and is gone from the tree with no delete
commit for the path — nothing to repoint to. Per `check-spec-citations.sh`'s own
instruction the path comes out rather than dangling, and no plausible substitute was
invented. It surfaced only because the R3 edit pulled the file into the gate's changed-file
scope, which is that scoping working exactly as designed.

## Verification

- `scripts/check-docs-lifecycle.mjs`: 374, re-baselined
- `scripts/check-doc-status.sh`: ok (100 docs checked)
- `scripts/check-spec-citations.sh`: ok (105 files checked)

agent_id: manoz-0803a


a5af:
@codex review

PR #3357, opened 2026-09-18
Opaz@charlie: ci: compile the feature-gated authfile tests; document the probe and a gh breakage

GenericAgentX-asaf:
## Summary

Three small gaps this session exposed, all of them "nothing was actually checking this".

### 1. `agentmux-cef`'s `test-authfile` tests were compiled by **no lane**

PR CI deliberately excludes `agentmux-cef` — `ci-pr.yml`'s header explains it well (CEF compile ate ~19 of 20 minutes on the Windows leg), and nightly owns the full workspace instead. But nightly's `cargo test --workspace` doesn't enable non-default features, so that module was built nowhere.

It rotted exactly as you'd expect. PR #3354 shipped **a duplicate `mod tests` (E0428) and two stale call sites with the wrong argument count**, and every one of them passed the `windows-latest` leg, because the code was never compiled. Review caught all three; CI structurally could not have.

Added to **nightly**, which already owns `agentmux-cef` coverage and has already paid for the CEF build — near-zero marginal cost. Deliberately **not** added to PR CI: that would reintroduce precisely the 20-minute Windows problem that workflow header documents.

### 2. `tools/tests/vim-freeze-probe.mjs` was on main, undocumented

It landed via an overly broad `git add tools/tests` in #3354 — my mistake; it wasn't in that PR's stated scope. It's genuinely useful (it's what drove real keystrokes into vim to find the DECRQM miscompile), so it now has a README entry rather than being quietly deleted.

The entry leads with `--control` and why it isn't optional: **the first version of that probe reported "FREEZE REPRODUCED" against a pane it had never managed to focus** — a false positive indistinguishable from the real bug. It also records the two harness defects so nobody rediscovers them:

- `el.focus()` sets DOM focus but doesn't drive the app's focused-pane routing
- xterm.js takes printable characters from the textarea's `input` event, **not** `keydown` — so `keydown`+`char` per letter delivers real events xterm correctly ignores, and only Enter arrives

### 3. `gh pr edit` is broken on gh 2.46.0

It still queries the sunset `projectCards`, so editing a PR body or title always errors. That's exactly what you need when you forgot the `<!-- agentmux:agent_id=… -->` tag — which, on the shared `GenericAgentX` identity, is the **only** signal routing review notifications to you. CLAUDE.md now gives the REST workaround.

## Test plan

- [x] Workflow YAML parses
- [x] `cargo test -p agentmux-cef --features test-authfile` passes locally (379) — the step added to nightly is the one I ran by hand to catch #3354's breaks
- [x] Doc gates pass

## Note

Two larger follow-ups from this session are deliberately **not** here, since they're design decisions rather than fixes: replacing the ambient `AGENTMUX_AUTH_KEY` in panes with a per-invocation handoff (retro §9.4), and an **I7** invariant plus enforcement test for env inheritance. `package-macos.sh`'s `${STRIP_MAPS:-1}` also still contradicts `SPEC_PORTABLE_SOURCE_MAPS` — flagged in #3355, may be a deliberate DMG-size call.


a5af:
@codex review

PR #3358, opened 2026-09-18
Opaz@charlie: docs(spec): panes should not hold the instance's API credential

GenericAgentX-asaf:
## Summary

Design doc, no code. Explores the one hole `SPEC_PANE_ENV_ISOLATION_2026_09_17` deliberately left open.

`AGENTMUX_AUTH_KEY` is injected into every spawned agent's environment and is the **sole** auth factor for the entire App API — `auth_middleware` is a string compare against a per-launch UUID, with no scope, expiry, per-caller identity or rate limit. Living in the environment means **every descendant inherits it**: the agent CLI, its shell tool, an `npm install` postinstall, a language server. On Linux it's also readable by any same-uid process via `/proc/PID/environ`.

A holder can do anything the UI can, since the frontend authenticates the same way.

## Already understood for one surface

`service/credential.rs` documents this exact attack and closes it with a host-only second factor, explicitly refusing to fall back to `X-AuthKey` *"which every agent can read from its own environment."* Right instinct, applied once — everything else behind `auth_middleware` still trusts the ambient key.

## Threat model, stated honestly

An agent is semi-trusted and runs arbitrary code by design. **Containing a hostile agent is a non-goal.** The achievable goals:

1. **Blast radius** — a postinstall script shouldn't inherit full API authority as a side effect of where it was started
2. **Cross-instance** — a credential inherited by a *different* build is authority over the instance that spawned it; the confused-deputy case, with no legitimate use
3. **Auditability** — today the server can't say which pane called

## Options (five, with costs)

- **B. Second factor per sensitive route** — cheap, proven in-tree, but a denylist: each new route must be remembered, which is the failure mode that produced five review rounds on #3326
- **C. Per-pane scoped + expiring tokens** — keeps the env-var transport so `muxclient.mjs` is unchanged, turns the master key into a capability, delivers auditability
- **D. Peer-credential auth over the per-instance Unix socket** (`SO_PEERCRED`) — authority from *who is connecting*, not a string anyone downstream can read; an inherited environment grants nothing because there's nothing in it
- **E. FD passing** — stronger, rejected: `muxsh` is a shell function the user invokes directly, so we don't launch every helper

**Recommended: C then D**, phased, because D alone is a redesign of four helper CLIs.

Notes the costs rather than hiding them: `SO_PEERCRED` identifies a *process*, not a pane, so pid→pane mapping still needs a registry; Windows needs named-pipe peer identity and this shouldn't land Unix-only; and `muxspect`'s cross-channel lookup may need an explicit capability under (C).

## Also records why rotation isn't the fix

The leak is inheritance *at spawn*, so a child gets whatever is current. Scope is the property that matters; expiry only helps alongside it.


a5af:
@codex review

PR #3360, opened 2026-09-18
fix(build): remove the esbuild >=0.28.2 override — it broke task dev

Agent4-asaf:
## Summary

`task dev` on `main` comes up as a bare logo. Vite's dep optimizer fails before
serving anything, with 2,705 copies of:

```
Transforming destructuring to the configured target environment
("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
is not supported yet
```

Root cause is #3343's `"overrides": { "esbuild": ">=0.28.2" }`. An npm override is
**global across the tree**, so it forced esbuild 0.28.2 into `vite@6.4.3`, which
declares `^0.25.0`. Vite's optimizer passes a `supported` map alongside `target`;
between 0.25.x and 0.28.x esbuild changed how those two interact, so destructuring
gets queued for lowering — and destructuring lowering isn't implemented. Every dep
containing a destructuring pattern fails.

Reproduced with Vite's exact options against a single dependency:

```
esbuild 0.25.12 (vite's own)  -> OK - no error
esbuild 0.28.2  (forced)      -> ERRORS: 22
```

## The override also never protected what it was added to protect

The floor was added on the reading that "Vite 6.4.3 resolves esbuild ^0.25.0 and we
had 0.28.1, one patch short." That 0.28.1 was the **hoisted root copy, owned by
`tsx`** (`~0.28.0`), which doesn't touch the frontend bundle. Vite had its own
nested copy the whole time — before #3343 and after this revert:

```
node_modules/esbuild                   -> 0.28.1  (tsx)
node_modules/vite/node_modules/esbuild -> 0.25.12 (vite)   <- built the bundle
```

So the override was a no-op for `tsx` (`~0.28.0` resolves to 0.28.2 either way), and
for Vite it didn't *raise* a version — it *forced an unsupported one*. That was its
only effect on this repo. Meanwhile 0.25.12, the esbuild that actually built the
bundle, still miscompiles under `safari13`:

```
0.25.12 -> function o(r,t){return(e=>e[e.SET=1]="SET")(void 0||(n={})),r+t}   <- declaration dropped
0.28.2  -> function o(r,n){let e;return(t=>t[t.SET=1]="SET")(e||(e={})),r+n}  <- correct
```

## Why removing it is safe — the vim fix is the target change, not the floor

#3343's own commit message says so: *"removing safari13 and this floor are
independent fixes; either alone is sufficient for this bug."*

The miscompile only fires when esbuild **lowers** `a ||= {}`, and it only lowers when
the target lacks logical assignment. With `safari13` dropped, `vite.config.ts`'s
`target: ["es2021", "chrome97"]` supports it natively, so nothing lowers on any
esbuild version:

```
["es2021","chrome97"]            -> lowered: false | o.handlers ||= {};
["es2021","chrome97","safari13"] -> lowered: true  | o.handlers || (o.handlers = {});
```

The parser-wedge recovery from #3343 (`frontend/app/view/term/parser-wedge.ts`) is
untouched.

## Alternatives rejected

| Option | Why not |
|---|---|
| Upgrade Vite so 0.28.x is in range | 6.4.3 / 7.0 / 7.1 / 7.2 declare `^0.25.0`; 7.3.0 declares `^0.27.0`; 8.x drops esbuild for rolldown. No version in the line accepts 0.28.x. |
| Scope the override to Vite's nested copy | Hoists the pin to the root and invalidates `tsx`'s `~0.28.0` — trades one range violation for another. Tried; it breaks. |

## Changes

- `package.json` — drop `"esbuild": ">=0.28.2"` from `overrides`.
- `package-lock.json` — purely additive (486 insertions, 0 deletions): restores
  `vite/node_modules/esbuild@0.25.12` + platform binaries. Root stays 0.28.2.
- `docs/specs/SPEC_ESBUILD_OVERRIDE_VITE_INCOMPAT_2026_09_17.md` — full writeup.
- changeset (patch).

## Verification

- `npm ls esbuild` → `tsx -> 0.28.2`, `vite -> 0.25.12` (the historical layout).
- `npx vite optimize --force` completes — previously 2,705 errors, now none.
- Lowering check above confirms `||=` survives verbatim under the shipped target on
  Vite's own esbuild, so the DECRQM handler can't regress.
- `task dev` launch verified (see comment below).

## Follow-up (not in this PR)

The real gap #3343 exposed is dependency currency, and it's unchanged: Vite 6.4.3 is
pinned to an esbuild line that no longer gets the fix for evanw/esbuild#4508. We're
not exposed today because our target never triggers lowering — but that's a property
of one config line, not of the toolchain. Moving off 0.25.x means moving Vite forward
(7.3.0 for `^0.27.0`, or the rolldown migration), which deserves its own scoped work
rather than arriving via `overrides`.


a5af:
@codex review

Agent4-asaf:
## `task dev` verified end-to-end on Windows

Full `scripts\dev-agent.cmd TITLE=esbuild-override-fix` run against this branch:

```
Built dist/cef-dev-1789707168 with launcher layout
Auto-selected Vite port 5333
Waiting for Vite on port 5333...
[platformResolve] Target platform: win32
  VITE v6.4.3  ready in 749 ms
  ➜  Local:   http://localhost:5333/
Vite ready at http://localhost:5333.
```

Zero `Transforming destructuring ... is not supported yet` errors (was 2,705), and
the CEF window paints the full UI — agent list, sysinfo graph, swarm pane, widget
bar, `v0.56.4 DEV` in the status bar — rather than the bare logo `main` currently
shows.

The only remaining stderr in the run is the pre-existing Dart Sass `@import`
deprecation warning from `frontend/app/element/markdown.scss`, unrelated to this
change and present on `main`.

### Note on the earlier CI failure

Run 35308489141 failed the `Specs index is current (changed specs only)` gate —
adding a spec requires regenerating `docs/specs/INDEX.md`. Fixed in a995ea35 (a
one-line index addition). The `[CI] FAILED` notification that arrived after that
push refers to the same superseded run, which reports `"conclusion": "cancelled"` —
not a second, separate failure.

PR #3362, opened 2026-09-18
docs: cite the implementing PR on the esbuild spec, regenerate INDEX.md, add account-switch retro

Agent4-asaf:
## Summary

Three docs changes, two of them fixing gates that are currently red.

**1. `SPEC_ESBUILD_OVERRIDE_VITE_INCOMPAT_2026_09_17.md` — cite the implementing PR.**
The spec landed in #3360 with `**Status:** Implemented` and no PR citation, which trips
`check-docs-lifecycle.mjs` R3 (`'implemented' citing no PR`). It passed CI on #3360's first
run only because that branch's base predated #3356; merging main in pulled the stricter rule
and the gate went red. #3360 was merged manually before it settled, so the violation is on
main now. Now reads `**Status:** implemented — PR #3360 (404f8c9f).`

**2. Regenerate `docs/specs/INDEX.md`.**
#3356 restamped the Status line on 99 specs but never regenerated the index, so
`gen-docs-index.sh --check` fails for *any* PR touching `docs/specs` — including this one.
Pure regeneration, no hand edits. Verified: **915 unique spec links before and after, zero
added, zero dropped.** The net -105 lines are ~21 collapsed junk status sections
(`approved`, `audit`, `decision`, `ready`, `proposal`, ...) and their table headers, which
exist only because the index predates the restamp.

**3. New retro: provider account switches orphan an agent's history.**
`docs/retro/retro-provider-account-switch-loses-agent-history-2026-09-17.md`. Switching
Anthropic accounts mid-session creates a new identity UUID, and because transcripts and
agent memory live *inside* the credential-isolation boundary
(`~/.agentmux/shared/identities/<uuid>/`), the agent silently comes back with no history and
no signal that anything is missing. `SearchHistory` has the same blind spot — it resolves
under the current identity, so it returns a clean zero-hit result that reads as an
authoritative "you never did that". Recommendations in §6, ordered; the real fix is making
`SearchHistory`/`ListConversations` span sibling identities for the same `AGENTMUX_AGENT_ID`.

## Verification

All four doc gates, with real exit codes (not `$?` after a pipe, which reports `tail`):

| gate | rc |
|---|---|
| `check-doc-status.sh` | 0 |
| `check-docs-lifecycle.mjs` | 0 — `fixed: 1  newly violating: 0`, debt 183 -> 182 |
| `check-spec-citations.sh` | 0 |
| `gen-docs-index.sh --check` | 0 (was 1) |

Docs only — no code, no behaviour change.


a5af:
@codex review

Agent4-asaf:
@reagentx-workflow review

(Verification trigger for reagent v5.19.0, deployed to prod at 06:34:24Z. The account behind the reviewer is still over quota, so this is expected to fail — the point is *how* it fails: it should now name one cause and say an operator is needed, instead of listing content-filter/crash/timeout.)

Agent4-asaf:
@reagentx-workflow review

(Verification trigger for reagent v5.20.0 — account pool + secrets TTL, deployed to prod. Pool is still unpopulated (only the legacy claude-api-token is set), so this should behave identically to before: same quota-exhausted classification as the last run, going through the legacy fallback path in _resolve_account rather than erroring.)

Agent4-asaf:
@reagentx-workflow review

(Verification trigger post account-switch: claude-account now points at poppercornell in services/infra. Expecting an actual review this time, not a quota_exhausted skip.)

Agent4-asaf:

PR #3363, opened 2026-09-18
Opaz@charlie: feat(statusbar): the Disk pill is not Windows-only any more

GenericAgentX-asaf:
## Summary

The status bar has had **no Disk readout on Linux or macOS**. The pill is gated on `disk:pagefile_volume:free_pct`, produced only by `get_pagefile_volume_data` (`agentmux-srv/src/backend/sysinfo.rs`), whose entire body is `#[cfg(target_os = "windows")]`. Off Windows the key never appears, the `<Show>` never fires, and the click-open per-drive popover is unreachable because it renders inside that same `Show`.

**The data was already there.** `get_disk_data` runs on every platform and emits `disk:vol:<mount>:free_gb` / `:total_gb` per mount — so on Linux the popover's figures have been collected every tick and discarded. Only the *summary* was missing: a free-percentage for one designated volume, plus the markers saying which volume it is.

Watching a volume's free space is not a Windows idea. The *reason* we started was Windows-specific (`SPEC_WIN10_PAGEFILE_OOM_CRASH_2026_06_29`: a system-managed page file silently failing to grow, pinning the commit ceiling). The gauge is generic. Because the concept was named after the mechanism, the `cfg` gate travelled with the name.

## What changed

- **srv** — summary keys renamed `disk:pagefile_volume:*` → `disk:watch:*`, and a Unix collector added. Watch volume = the mount backing the **data dir** (where the store, logs and packaged builds land), chosen by longest-prefix match against the mounts `sysinfo` already enumerated. No new syscall: it reuses this tick's `Disks` list and the per-volume figures `get_disk_data` just wrote, so the pill and the popover row are arithmetically identical by construction.
- **frontend** — `pagefileSystemManaged` becomes **tri-state**, which is the part that actually matters. It was parsed `(vals[k] ?? 0) > 0`, which cannot distinguish an *absent* key (not Windows) from an explicit `false` (fixed-size page file, deliberately never coloured). Merely emitting the metric would have rendered a Linux pill that stays muted at 0.5% free.
- Colour fn moved out of `SystemStats.tsx` (private, untested) into `disk-volumes.ts` beside the popover's identical 8/15 brackets, and tested across all three states.
- `disk:pagefile_system_managed` deliberately keeps its Windows name: it *is* a page-file fact, and its **absence** is the signal the frontend reads.

## Why now

This machine's root filesystem hit **100% (1.9 GB free)** on 2026-09-17. Nothing in the UI said so — it surfaced as a failed packaging build, after an hour spent looking at the build instead of the disk. The one widget whose job is to warn about this is the one that does not render on this platform.

## Testing

- `cargo test --release -p agentmux-srv watch_volume` — 9 passed. Covers longest-prefix over `/`, root fallback, exact match, the sibling-mount trap (`/home2` must not prefix `/home/yas`), partial/zero-capacity ticks emitting no summary, and the summary matching its per-volume figures.
- `vitest frontend/app/statusbar/disk-volumes.test.ts` — 16 passed, including an explicit regression test that a **fixed-size page file is still never coloured** (the Windows behaviour this must not break).
- `tsc --noEmit` clean.
- Not verified on macOS hardware — same Unix code path, `/` fallback is the only platform assumption.

## Reviewer note

**461 lines of the `INDEX.md` diff are not mine.** Main's index is stale (#3356 rewrote 99 specs' status lines without regenerating), and the gate forces a full regeneration. My own contribution to that file is one row.

Spec: `docs/specs/SPEC_STATUSBAR_DISK_PILL_CROSS_PLATFORM_2026_09_18.md`


a5af:
@codex review

PR #3364, opened 2026-09-18
fix(composer): land caret at start after ArrowUp history recall

AgentO-asaf:
## Summary
- Pressing ArrowUp to recall a previously sent message only steps further back once the caret is at true position 0, but `setComposerValue` always parked the caret at the **end** of the recalled text — so for a single-line message, a second consecutive ArrowUp just moved the caret back to 0 natively instead of recalling the next-older entry, requiring two presses per history step.
- `setComposerValue` now takes a `caret: "start" | "end"` option (defaults to `"end"`, preserving all other call sites' behavior — Esc-restore, autocomplete accept, queued-message un-send, and ArrowDown-toward-newer). The ArrowUp sent-history recall path now passes `"start"`, so the caret lands at position 0 on the recalled message and repeated ArrowUp presses walk back through history continuously.

## Test plan
- [x] `npx vitest run frontend/app/view/agent/components/AgentFooter.test.tsx` — 27/27 pass, including a new regression test that presses ArrowUp three times in a row (no manual caret reset in between, unlike the file's other history tests) and asserts both the recalled value and `selectionStart === 0` at each step.
- [x] `npx tsc --noEmit -p .` — no errors.


a5af:
@codex review

PR #3365, opened 2026-09-18
Opaz@charlie: feat(isolation): add invariant I7, and close two spawn paths that broke it

GenericAgentX-asaf:
## Summary

`docs/retro/retro-env-inheritance-instance-isolation-breach-2026-09-17.md` §9.5 asked for an **I7 invariant plus an enforcement test**, and named the absence of that test as the reason five prior documents identified this class of bug and none closed it. This is that.

**I7 — an instance's identity must not be inheritable by the processes it spawns.**

I1–I6 did not catch the breach, and the reason is structural rather than an oversight: they enumerate **named OS objects and directories**, and their stated threat model is *"launching a new build must never crash a running one."* Environment inheritance is neither an OS object nor a crash, and it runs the opposite way — the **new** instance is silently corrupted by the **old** one.

## Two live violations on main

Auditing every spawn site in srv against I7 turned up two that were never covered:

| Site | What inherits |
|---|---|
| `util.rs::open_browser` | Every link AgentMux opens hands the user's **browser** this instance's channel, data/config/cache dirs, runtime mode and `AGENTMUX_LOCAL_URL`. The browser is long-lived, and on Linux that environment stays readable by any same-uid process via `/proc/PID/environ` for as long as it runs. |
| `editor_handlers.rs` reveal-in-file-manager | Same inheritance into the desktop file manager — from which a user can launch anything, including another AgentMux build. That is the retro's §2.2 failure exactly. |

Both now take the strict policy (`sanitize_external_std_command`), the same treatment `identity_auth_spawn.rs` already gives provider CLIs.

**The credential was not exposed.** `config.rs` strips `AGENTMUX_AUTH_KEY` from srv's own environment at startup, so what leaked here was identity, not authority. Worth stating plainly so this isn't mis-scoped in either direction.

## The enforcement test is the actual deliverable

The policy was never the hard part. Remembering to apply it at a **new** spawn site was — both P0s on the original fix were exactly that.

`spawn_site_coverage` (in `pane_env.rs`, beside the policy it enforces) walks srv's source and fails when a file constructs a process without either sanitizing or appearing in `EXEMPT` **with a stated reason**. A second test fails when an exemption goes stale, since an allowlist of dead entries stops meaning anything.

File-granularity substring matching is deliberate: a false positive costs one line and a sentence of justification; a false negative costs another cross-instance breach.

It earned its keep immediately — it rejected one of my own exemptions for having a reason too short to be a reason (`"Test-only."`). I wrote a real one.

Eight exemptions, each justified. The instructive one is `crash_monitor.rs`: it re-spawns our *own* exe, so inheritance is the requirement there, not the defect.

## Testing

- `cargo test --release -p agentmux-srv pane_env` — **9 passed**, the 7 pre-existing policy tests plus the 2 new coverage tests.
- Cross-checked the walker by reimplementing it independently in Python: 414 files walked, zero offenders, no stale exemptions — agreeing with the Rust result.
- Not covered: this checks *source* coverage, not runtime behaviour. An end-to-end test that launches a build from a pane and asserts it resolves its own channel and data dir is still the stronger check, and belongs in the nightly lane rather than per-PR.


a5af:
@codex review

GenericAgentX-asaf:
Both addressed in `f3d47e2`.

**P1 — you were right, and my justification was wrong.** I argued file-level granularity's failure mode was false *positives* ("costs a line and a sentence"). It's false *negatives*: a file that already sanitizes one command passed wholesale, so a new unsanitized spawn beside it was invisible. That's the expensive direction, and as you noted it describes most spawn files in the crate — `cli_handlers.rs` being the live example.

Per-occurrence matching doesn't work either, and `shell/lifecycle.rs` is why: it builds four `CommandBuilder`s as `c` and sanitizes once, later, at a choke point as `cmd`. That's the *correct* pattern, so any syntactic per-site rule flags it — proving it needs dataflow analysis.

So the surface is pinned instead. `SPAWN_INVENTORY` enumerates every `(file, program)` pair that constructs a process — **42 keys, 50 sites** — with an exact count and a classification (`sanitized:` / `probe:` / `test:` / `own-exe:` / `not a spawn:`). A new spawn anywhere changes the map and fails regardless of its neighbours; removing one fails too, so the table can't rot.

Verified both directions against your counterexample file specifically:

| injection into `cli_handlers.rs` (which already sanitizes twice) | result |
|---|---|
| `Command::new("curl")` — new program | new key → **fails** |
| `Command::new("which")` — program already present | count `2 → 3` → **fails** |

The second case is the one a key-only check would miss.

**Auditing all 50 sites to build the table found one more worth fixing:** `tool_store::probe_version` runs a third-party binary, so it now takes the strict policy for the same reason `open_browser` does. It also showed two of the grep matches aren't spawns at all — a module header comment in `system_install_handlers.rs` and a comment in `persistent.rs` — both now classified as such rather than silently passing.

**P2** — the mangled strings are gone with the old table. The new rows are checked for a recognised class prefix *and* a reason long enough to be one, which rejected ten of my own first-pass entries for being too terse; they now name the specific choke point or probe instead of "same as above".

Tests: `cargo test --release -p agentmux-srv pane_env` — 9 passed. Rebased onto `56832d4`; `check-doc-status`, `check-spec-citations` (the widened gate from #3369), `check-docs-lifecycle`, `gen-docs-index --check` and `check-doc-links` all clean.

PR #3366, opened 2026-09-18
Manoz@Area54: docs: gate relative doc links, and clear the 45 already broken

GenericAgentX-asaf:
## The gap

`check-spec-citations.sh` only recognises a citation that literally names `docs/specs/<x>.md`. A Markdown link written as `](specs/archive/Y.md)` inside `docs/reports/` contains no such substring — so the gate cannot see it, and that exact form has been **dead since the top-level `specs/` tree was folded into `docs/specs/` (#2920)**. It kept resolving only from files that happened to sit in `docs/`.

## The gate

`scripts/check-doc-links.mjs` resolves every relative Markdown link against the directory of the file containing it. First run over the whole tree:

| | |
|---|---|
| Links checked | 1,226 across 1,486 docs |
| **Broken** | **45 across 27 files → 0** |
| Runtime | ~0.5s |

Wired into the docs CI job, **scoped to changed files** like every gate beside it — a repo-wide gate fails every PR on somebody else's rot and gets switched off within a day, which is the documented fate of the three previous attempts at docs enforcement here. `--all` reports the backlog without failing.

## Three false-positive classes, each found the hard way

- **Prose in parens** — `](true)`, `](url)`, `](next)` all appear in real specs.
- **Fenced blocks** — several specs contain ASCII-art UI mockups that *depict* a rendered Markdown list. I de-linked one before adding this rule and it broke the box drawing; that's how the rule was found.
- **Inline code** — specs routinely *show* link syntax while explaining these gates. The spec note in this very PR did it and the gate caught me.

## The 45

11 were wrong relative paths (repointed). 23 pointed at things that no longer exist, or never did — de-linked to plain text, per this repo's own rule that a broken pointer is worse than none. The rest were the classes above.

Two were substantive rather than cosmetic:

- **`perf-baseline-2026-05-09.md`** cross-referenced `SPEC_APP_API_AUTOMATION_SURFACE.md` as *"pushed alongside this retro"*. No such file exists in the repo **or anywhere in its history** — it never landed. I said so rather than repointing at a plausible-looking substitute, which the citation gate explicitly warns against.
- **Six links** pointed into `C--Systems/memory/`, the machine-local agent-memory directory retired 2026-09-08. Dead by construction.

Touching two May retros pulled their free-text `Status:` lines into the doc-status gate's scope; both are now `historical`, original prose preserved.

## Phase 2 status, recorded rather than guessed

`SPEC_DOCS_LIFECYCLE_HARDENING_2026_08_03.md` listed Phase 2 (directory consolidation) as not started. Measuring it: the duplicate top-level `specs/` tree and `docs/retros/` are **already gone**, so the four archive directories the audit found are now three.

The spec now records why the last one is not simply moved: **`docs/specs/archive/` is load-bearing tooling, not just a folder.** `gen-docs-index.sh` excludes it by path (its generated preamble explains `archive/` to the reader), `check-docs-lifecycle.mjs` hardcodes the path, and two build scripts cite files inside it. Moving it forces an INDEX.md regeneration — and that is **not reproducible off-CI**: a Windows run of `gen-docs-index.sh` emits ~21 extra status buckets and a different row order than CI's Linux run. That is a real constraint the original audit did not surface, and the spec's own escape hatch ("or explicitly document why more than one is needed") applies.

## Verification

Run locally on the full tree and on the changed set:

- `check-doc-links.mjs --all` → 0 broken (from 45)
- `check-doc-links.mjs` non-vacuous: probed with a deliberate broken link, exits 1 and names it
- `check-doc-status.sh` → ok (17 docs)
- `check-spec-citations.sh` → ok (18 files)
- `check-docs-lifecycle.mjs` → ok, `newly violating: 0`

`gen-docs-index.sh --check` was **not** run locally, deliberately — see the platform-divergence note above. Nothing index-relevant changed (no spec added, removed, or re-Status-ed at the first word), so it should be a no-op; CI is the authority on that.


a5af:
@codex review

PR #3367, opened 2026-09-18
Manoz@Area54: docs: fold docs/analysis/archive into docs/archive, and write down the rule

GenericAgentX-asaf:
## What

Phase 2 of the docs-lifecycle hardening asked for **one** archive directory, not four — *"or explicitly document why more than one is needed if there's a real reason this audit didn't surface."* This does both halves.

| | Before | After |
|---|---|---|
| Archive directories | 3 | **2** |
| Documented rule | none (no README in any of them) | `docs/archive/README.md` |

Two of the original four were already gone (`specs/archive/` left with the top-level `specs/` tree). This folds `docs/analysis/archive/` (6 files) into `docs/archive/`.

## Why `docs/specs/archive/` stays

It is **load-bearing tooling, not merely a folder**:

- `gen-docs-index.sh` scans `docs/specs/` and excludes exactly that subtree — it is the mechanism by which a spec leaves the generated index while staying beside its siblings, and the generated preamble explains `archive/` to the reader in those terms.
- `check-docs-lifecycle.mjs` hardcodes the path in its `EXCLUDED` regex.
- `linux-apprun.sh` and `package-portable.sh` cite files inside it by path.

Moving it means editing the generated preamble, which means regenerating `INDEX.md` — and that is **not reproducible off-CI**: a Windows run emits ~21 extra status buckets and a different row order than CI's Linux run. So this takes the spec's own escape hatch, deliberately and in writing, rather than leaving the split unexplained.

## The link gate earned its keep immediately

The moved files went from `docs/analysis/archive/` (depth 3) to `docs/archive/` (depth 2), which silently invalidated three of their **outbound** links — `../../retro/x.md` had to become `../retro/x.md`. `scripts/check-doc-links.mjs` (added in #3366, merged yesterday) caught all three the moment the move landed. That is the first time it has caught something I would otherwise have shipped.

## Inbound references: split deliberately

- **Repointed (4)** — live pointers that name a moved file.
- **Left alone (2)** — `REPORT_DEAD_CODE_AND_DOCS_SWEEP_2026_08_03.md` and `REPORT_TRUST_CENTER_TO_ARMORY_TERMINOLOGY_SWEEP_2026_07_19.md` narrate a decision made at the time (*"→ `docs/analysis/archive/`"*). Rewriting those would falsify the record rather than fix a pointer.

## Incidental

Touching four docs pulled their free-text `Status:` lines into the doc-status gate's scope; all four now use the closed vocabulary with their original wording preserved. One citation in the moved `data-dir-status-2026-05-09.md` pointed at `docs/specs/SPEC_DATA_DIR_UNIFICATION_2026-05-05.md`, which has since been archived — repointed to where it actually lives.

## Verification

- `check-doc-status.sh` → ok (12 docs)
- `check-spec-citations.sh` → ok (12 files)
- `check-docs-lifecycle.mjs` → ok, `newly violating: 0`

`gen-docs-index.sh --check` not run locally (platform divergence above); nothing under `docs/specs/` was added, removed, or re-Status-ed at the first word, so it should be a no-op. CI is the authority.


a5af:
@codex review

PR #3368, opened 2026-09-18
Manoz@Area54: docs: gate the docs/ directory map against reality

GenericAgentX-asaf:
## Why

`docs/README.md` is **the map** — it is what tells a reader, or an agent, where a given kind of doc belongs. A wrong map is worse than no map, and this one has gone stale at least twice:

- The **2026-08-03 lifecycle audit** found *"this directory list and the specs-location claim above were themselves found stale."* The file still carries that note, about itself.
- **`docs/handoff/` was deleted by PR #2407** — a *docs cleanup sweep*. The sweep removed the directory and left its row behind, so for six weeks the map advertised a directory that did not exist. Over the same period **`docs/debug/` was added and never listed at all**.

Both drifts are mechanical. Both were invisible until something looked. That is precisely the failure mode `SPEC_DOCS_LIFECYCLE_HARDENING` exists to stop — *"one-shot fixes rot, self-verifying systems don't."*

## What

`scripts/check-docs-dir-map.mjs`: every `` `dir/` `` named in the table must contain tracked files, and every directory under `docs/` holding tracked files must be named in the table. Both existing drifts are fixed in this PR, so it lands green.

**Deliberately not changed-files-scoped**, unlike the four docs gates beside it. Those are scoped because their backlogs run to the hundreds and a repo-wide gate would fail every PR on somebody else's rot — the documented fate of three previous attempts at docs enforcement here. This one reads a single file and its backlog was exactly two, so it can be absolute without ever failing a PR that didn't cause the problem.

## Two traps, both hit while writing it

- **Table rows only.** The prose below the table names `docs-internal/` *precisely to say it does not exist*. Scanning the whole file turned that into a demand that the repo create it.
- **My own note failed the gate.** I first wrote the explanation of the removed handoff directory *inside a table cell* — and a backticked `` `name/` `` in a row is a claim, so the gate failed on its own fix. The note now lives below the table.

## Verification

- `check-docs-dir-map.mjs` → ok (19 directories, all listed and all real)
- Non-vacuous: created an unlisted `docs/probe-xyz/`, gate exits 1 and names it; removed, back to green
- `check-doc-status.sh`, `check-spec-citations.sh` → ok

## Context

Third in the docs-lifecycle sequence: #3366 added the relative-link gate and cleared 45 broken links; #3367 folds `docs/analysis/archive/` into `docs/archive/` and documents the archive rule. This one is independent of both — it branches from `main` and touches only `docs/README.md`, the new script, and the CI job.


a5af:
@codex review

PR #3369, opened 2026-09-18
Manoz@Area54: docs: widen the citation gate from docs/specs/ to all of docs/

GenericAgentX-asaf:
## The gap reagent found

Its P2 on #3367 flagged a source comment citing a moved doc that **no gate could see**:

- `check-doc-links.mjs` (#3366) only scans Markdown.
- `check-spec-citations.sh` only recognises paths naming `docs/specs/`.

A code comment citing `docs/analysis/...` or `docs/retro/...` falls straight between them.

## Measuring it

| | |
|---|---|
| Source citations to `docs/**/*.md` | 2,014 |
| Dangling | **54** |
| …outside `docs/specs/`, i.e. invisible to the existing gate | **44** |

## Ten are real — and all five targets never existed

No file, no deletion, **nothing in git history**:

| Document | Cited from |
|---|---|
| `ANALYSIS_IDLE_SEND_RACE_2026_06_11` | 4 frontend files |
| `ANALYSIS_ARMORY_STASH_CREDENTIAL_VISIBILITY_GAP_2026_08_04` | 2 |
| `retro-window-count-stale-post-1701-2026-06-27` | 2 |
| `retro-my-agents-fresh-channel-regression-2026_07_27` | 1 |
| `retro-dev-instance-vite-orphan-port-2026-06-16` | 1 (`Taskfile.yml`) |

Same shape as the `SPEC_APP_API_AUTOMATION_SURFACE.md` finding in #3366 — a doc referenced as if it existed, written in a working tree and never pushed.

These are the **most expensive** kind of dangling pointer. A comment citing design rationale is often the only link between a decision and the code implementing it; here it sends the reader after a file that never existed. Each now names the document *without a path* and says it was never committed, so the next reader stops instead of searching. Per the gate's own advice: don't invent a plausible-looking replacement.

## The other 34 are legitimately fake, and excluded

`bundle_export.rs` holds path-collision examples and fixture data (`docs/a.md`, `docs/./a.md`, `docs/readme.md`); `docs-stale-sweep.test.mjs` uses `docs/fixtures/*.md`; `check-doc-links.mjs` quotes `docs/reports/X.md` while explaining itself. Firing on those is precisely the false positive that gets a gate switched off — the documented fate of three previous attempts here.

**Correction (reagent P1, fixed in 1f0f777).** The first version of this PR excluded them with `*test*|*fixtures*|scripts/check-*`, and this description claimed that covered `bundle_export.rs`. It did not — that file is not test-named, so the glob never matched it, and the next PR touching it would have hit the exact false positive this was meant to prevent. reagent caught the contradiction between the claim and the code.

The rule now requires a **subdirectory**: `docs/<dir>/<file>.md`. That is the shape every real citation has (and what the original `docs/specs/` rule already required), while `docs/a.md` and `docs/readme.md` have no directory component and stop matching by construction — no special-casing. `docs/en/*` is also skipped: that is Anthropic published docs site, not this tree.

The two globs stay, still load-bearing: `docs/fixtures/A.md` and `docs/reports/X.md` *do* have a directory component, so only a file-level skip catches them.

## Scoping unchanged

Still changed-files-only, so the **14 pre-existing** dangling citations elsewhere don't fail anyone's PR — same posture as every docs gate beside it.

## Verification

- Widened gate on the 10 changed files → ok
- **Non-vacuous**: added `// see docs/retro/definitely-not-a-real-doc-xyz.md` to `state.ts` → exits 1 and names it (a path the *old* `docs/specs/`-only rule would have ignored); removed → green
- Repo-wide backlog under the new rule: 14 files, all pre-existing

Fourth in the docs-lifecycle sequence (#3366, #3367, #3368). Branches from `main`, independent of the other three.


a5af:
@codex review

PR #3370, opened 2026-09-18
Manoz@Area54: docs: refresh the wrap-up report, fold two singleton dirs, normalise 17 Status lines

GenericAgentX-asaf:
Two things, landed together because the second changes what the first documents.

## 1. The wrap-up report now carries current data

**§2.4's heading was wrong in both directions.** It read *"Phase 2 is all that's left ✅"* — Part 4 wasn't finished either, and Phase 2 was further along than the spec recorded. Rewritten with the four gates that landed 2026-09-18 (#3366–#3369) and their first-run backlogs.

**§5.5 only counted `docs/specs/`.** Re-measured across all 1,488 tracked `.md` under `docs/`:

| | count |
|---|---:|
| canonical `**Status:**` + vocabulary word | 856 |
| canonical line, non-vocabulary word | 208 |
| other format, word IS in the vocabulary | 17 |
| other format, word is not | 46 |
| no status declaration at all | 355 |
| **total non-compliant** | **626** |

**I have not claimed a burn-down against the 357 in that section.** The methods aren't comparable, and its named top offenders don't survive re-measurement (`ready` ×35 is now zero files). Quoting a delta before one method is written down is the error §2.6 already corrects once.

**Why most are non-compliant isn't neglect.** Top non-vocabulary words are `root` ×45 (*"root-caused; fix in …"*), `analysis` ×25, `shipped` ×11 — and they sit in `docs/retro/` and `docs/analysis/`. The closed vocabulary was designed for **specs**, which are 975 of 1,488. A retro isn't "implemented"; it's a record. That's a decision about whether to extend the enum or exempt those directories — **not** a restamping campaign, which `PLAN_DOCS_CLEANUP_EXECUTION` §4 already ruled against.

## 2. Organisation

- `docs/sessions/` (1 file, 0 inbound refs) → `docs/status/`
- `docs/recovery/` (3 files) → `docs/incident/` — `docs/README.md` already described these as one category

**18 directories → 17.** All four moves are `R100` pure renames, which `check-doc-status.sh` excludes by design, so nothing was restamped as a side effect of moving.

`docs/api/`, `docs/cef-patches/` and `docs/providers/` **stay** despite being small — each has a distinct audience or purpose (user-facing App API guide; the vendored CEF patch set with its own README; a cross-provider reference cited from five places).

**17 Status lines normalised** — the subset where the word was already in the vocabulary and only formatting hid it. The other 46 are deliberately untouched: reformatting a line reading `Root-caused` converts a file the gate *ignores* into one it *fails*, without anyone deciding the doc's state.

## What the gates caught, on me

- The moves invalidated **7 inbound references** to the migration runbook, across two specs, two docs and three source files. The widened citation gate (#3369) caught every one.
- A **sixth never-committed document**: `ANALYSIS_UX_LATENCY_THREE_FRONTS_2026_05_29.md`, cited as *"findings this spec implements"*. No file, no deletion, nothing in history.
- **reagent P1** — §5.5a's table said 355 and the paragraph below said 336. Same population, two numbers, both mine. Fixed.
- **reagent P2** — I wrote "four frontend files"; it's five. The fifth (`agent-view.tsx`) writes the name with no `docs/analysis/` prefix, so the path-shaped scan couldn't see it. Corrected there too.

That last one is worth noting: **the widened gate would not have caught it either.** It matches `docs/<dir>/<file>.md`, and a bare `ANALYSIS_X.md` has no directory component by design — the same property that keeps `docs/a.md` fixtures from matching. Catching bare names needs a different rule.

## Verification

- `check-doc-links.mjs --all` → 0 broken across 1,488 docs
- `check-docs-dir-map.mjs` → ok (17 directories)
- `check-spec-citations.sh` over the full diff → ok (32 files)
- `check-doc-status.sh` **invoked as CI invokes it** → ok (22 docs); renames correctly excluded


a5af:
@codex review

PR #3371, opened 2026-09-18
Lark@narko: fix(agent-pane): progress bar matches pane border color and sweeps the full perimeter

GenericAgentX-asaf:
## Summary

The agent pane's working/busy indicator was a fixed blue/white marching-ants strip across just the top edge. Requested: match the pane's own border color, and sweep the animation around all 4 sides continuously in one direction.

- **Color:** `--progress-bar-color` (new, set on `.agent-pane-stack`, mirroring `PaneChrome.scss`'s own `--pane-ring-color`/`--border-color`/`--accent-color` ring-color logic exactly — unfocused vs focused) replaces the hardcoded `--accent-color` the bar used directly. The bar now reads as the same color as the pane's selection ring, switching correctly when the pane gains/loses focus.
- **Full perimeter:** `.agent-pane-progress-bar-slot` now covers the whole pane rect (was a 3px top-only strip). `.agent-pane-progress-bar` itself is a static "picture frame" mask (`padding` = ring thickness, two masks XORed via `mask-composite: exclude` — the standard CSS animated-gradient-border technique), with an animated `::before` child underneath.
- **Continuous clockwise sweep:** the `::before` child uses a `repeating-conic-gradient` (not the original `repeating-linear-gradient`) animated via `transform: rotate()`. A linear diagonal hatch translated in one screen direction looks fine sliding along a single straight strip, but breaks down once revealed through a full-perimeter ring — same global slide direction reads as diagonal motion on top/bottom but the wrong axis of motion on the left/right edges, never reading as one continuous loop (caught in manual testing). A conic gradient's wedges are defined by angle-from-center, so a single `rotate()` sweeps every point on the ring in the same direction at once, including through all 4 corners. Positive `rotate()` is clockwise in CSS, matching the ask.
- Kept `transform`-only animation (rotate, like the original's translateX) rather than `background-position` — not reliably GPU-promoted on all Linux/Windows HW-accel paths, per the existing code's own documented reasoning.

Full design writeup: see the code comments in `agent-view.scss` above `.agent-pane-progress-bar`.

## Test plan
- [x] `task dev` on this branch — SCSS compiles cleanly, HMR updates apply with no errors
- [x] Visual check in a running dev instance — confirmed by the requester as matching the desired look
- [ ] No automated test coverage added (pure CSS/visual change, no new branching logic)


a5af:
@codex review

PR #3372, opened 2026-09-18
Manoz@Area54: refactor(paths): one resolver for the AgentMux root (A10)

GenericAgentX-asaf:
**A10** of `docs/analysis/TRACKING_ARCHITECTURE_REFACTOR_A1_A15_2026_06_18.md` — the last open item on that board.

## The defect

`agentmux-common/src/data_paths.rs` calls itself the *"single source of truth"* for where state lives. srv had its own copy. They disagreed **twice over**:

| | common crate | srv's `base.rs` |
|---|---|---|
| env var | `AGENTMUX_HOME_OVERRIDE` | `AGENTMUX_DATA_HOME` |
| `home_dir()` is `None` | returns `Err` | falls back to `PathBuf::from("/")` |

So setting either env var moved only half the process — and a host with no resolvable home directory **silently wrote to `/.agentmux`**, at the root of the disk.

## The change

`resolve_root` becomes `pub fn agentmux_root()` and honours **both** env vars in a documented order, so there is one resolver and **no install moves**. srv delegates to it. The `/` fallback is gone — replaced by a logged error and a relative `.agentmux`, because writing to the filesystem root is never the right answer.

Seven ad-hoc `dirs::home_dir().join(".agentmux")` sites now route through it: `config_watcher_fs`, `history/index`, `history/claude_adapter`, `session_archive`, `shellintegration`, `bootstrap`, `registry/paths`.

**The tracker's inventory was incomplete.** It named four sites plus `main.rs`; `main.rs` had already been fixed, and four sites it did not name turned up by grep.

**Deliberately unchanged:** `claude_adapter.rs:60` reads `~/.claude/projects` — that's *Claude's* directory, not AgentMux's. Routing it through this resolver would be a bug, not a fix.

## Acceptance: existing installs must still find their data

That's the criterion the board sets, so it's what the tests assert:

- an install that only ever set `AGENTMUX_DATA_HOME` still resolves to **exactly** that path
- with neither var set, still `~/.agentmux` (asserted against `dirs::home_dir()`, not a literal)
- precedence between the two is pinned, not accidental
- empty values are ignored — otherwise `AGENTMUX_DATA_HOME=` in a shell profile would redirect every install to `""`

**Verified non-vacuous:** removing `AGENTMUX_DATA_HOME` from the resolver fails the compatibility test *specifically* (`agentmux_root_honours_srvs_env_var_for_existing_installs`), and restoring it passes.

## Not to be confused with

`AGENTMUX_DATA_DIR` is untouched and means something else — the launcher exports it for the per-channel **data** directory (`<root>/channels/<channel>/data`), not the root. That near-identical name is documented at the resolver, since it's plausibly how srv grew a second one in the first place.

## Verification

- `cargo check -p agentmux-common` → clean
- `cargo check -p agentmux-srv` → exit 0, no new warnings in any touched file
- `cargo test -p agentmux-common data_paths` → 46 passed
- the 4 new tests → 4 passed, and proven to fail when the guarantee is broken


a5af:
@codex review

a5af:
@codex review

PR #3373, opened 2026-09-18
Clare@starpower: fix(panes): one shared pane-tab model — every tab is a pill with a stable icon

GenericAgentX-asaf:
## Problem

Adding a tab to a pane made icons disappear or change size. #3341 unified how the pane chrome is *drawn*, but left three separate *tab models* (generic in `PaneChrome.tsx`, `buildAgentPaneChromeModel`, `buildTermPaneChromeModel`) that had drifted:

- **No icon for agent/terminal tabs.** Both gave tabs of their own type no icon. After #3335 (any widget can be added to any pane), a mixed stack showed icons only on the added tabs.
- **Stale reads.** Agent/term read dormant and just-added tabs with a non-reactive snapshot, so a new tab could render as "Terminal 2" with no icon until the next layout change. PaneChrome already had the fix (ReAgent P1); it never propagated.
- **Size jumps.** The strip's zoom followed the *active* tab's content zoom (`term:zoom`), so adding a tab (which becomes active at 1.0) rescaled every pill and icon.
- **1 → 2 tab threshold.** A lone tab rendered as the header iconview (18px icon, 16px brand logo); at two tabs it switched to pills with 11px icons. The "hide a lone pill" rule was copied three times.

## Change

- **One tab model**, owned by `PaneChrome` (`frontend/app/element/pane-tab-model.tsx`). View types register a small `PaneTabDescriptor` for label/icon/rename:
  - agent (`agent-pane-tab.ts`): agent name + provider logo; rename writes the definition and `agentName`
  - terminal (`term-pane-tab.ts`): `pane-title` or "Terminal N"
  - everything else falls back to the **widget bar's own label and icon** (`widgets.json`), so a tab matches the button that opened it (Armory → vault instead of a square, Swarm → bee)
- Agent keeps only what's genuinely agent-specific via `PaneChromeModel.extraTabs` (cross-pane forks) plus `onActivate`/`onClose`; terminal keeps focus hand-off, connection button, new-tab cwd, and body wrapper. ~250 lines of duplicated tab-model code removed, plus dead focus-ring code in the agent builder.
- **Always a pill.** A lone tab is a pill too (the existing `pane:tabstrip: "multi-only"` setting still opts out).
- **Icons are data, rendered in one fixed 14px box** (font icon, provider logo, or favicon). Tabs are keyed by blockId so pills keep their DOM across recomputes.
- **Only chrome zoom scales the tabs.** `zoomFactor` removed from `PaneChromeModel`; the header row's `--zoomfactor` still applies. `PaneTabStrip.zoomFactor` remains for the editor's in-content file tabs.
- A browser tab keeps its last page title/favicon after another tab is added.
- The editor's file-tree toggle (an interactive view icon) stays in the header, in front of the pills.
- The first pill sits flush with the pane's left edge.

## Test plan

- [x] Full frontend suite: 4176 passed. New tests cover: lone pill, an icon surviving a tab add (DOM identity), descriptor labels + per-type numbering, remembered live names, extra (fork) tabs, rename + rollback on failure, widget-bar icon/label fallback, agent descriptor
- [x] Live `task dev` on macOS, checked via DevTools: every pane shows a pill with a 14×14 icon (sparkles / chart-line / bee / vault), first pill flush left
- [ ] Repo owner QA: add terminal/sysinfo tabs to a launched agent pane; zoom a terminal and add a tab; double-click rename; editor tree toggle


a5af:
@codex review

PR #3374, opened 2026-09-18
chore: release v0.56.6

AgentA-asaf:
Forced patch release (`task release -- --as patch`) consuming 15 pending changesets, all already typed `patch`, so the override doesn't change the computed bump — it just makes the release explicit per the standard release-PR workflow.

- `0.56.5` -> `0.56.6`
- All 5 release-consistency locations verified in sync (`scripts/verify-release-consistency.sh`)


a5af:
@codex review

PR #3376, opened 2026-09-18
Opaz@charlie: docs: correct two spec claims that were false

GenericAgentX-asaf:
## Summary

Two specs on main stated things that were not true. Both were found by reading, not by a gate, and the first one cost me real time yesterday.

### 1. `SPEC_AGENT_HISTORY_SEARCH_2026_09_17` said "nothing has shipped"

Phase 1 shipped the **same day** in #3321 and released in v0.56.4. I read that status line, believed it, and hand-parsed JSONL to search my own session history — while `SearchHistory` was already sitting in `agentmux-mcp/src/tool_schemas.rs`.

Now `implemented`, citing the PR (R3), re-verified against `90aa773` with the specific symbols named so the next reader can spot-check instead of trusting. Phase 2 (cross-agent search) is called out as still needing its own spec — `implemented` rather than `active` because §5 explicitly scopes this document to Phase 1 and defers Phase 2 elsewhere.

### 2. `SPEC_AGENT_APP_API_MCP_BINDINGS_2026_06_28` described the auth key backwards

It claimed `AGENTMUX_AUTH_KEY` is "*not* a PTY env var for trust reasons", citing `internals/env-vars.md` ("Available In Pane? No"), and called its presence in a shell a leak. Both halves were wrong:

- **No such file exists** in this repo, and never has. The only occurrence of the phrase "Available In Pane" anywhere under `docs/` is this spec quoting it.
- **The code does something more specific than either answer.** `config.rs` strips the key from srv's own environment, so it does *not* reach interactive shell panes — but `agent_handlers/input.rs` re-injects it for every agent spawn, and `pane_env.rs` keeps it on `PANE_ENV_KEEP` deliberately, because `lib/muxclient.mjs` (and therefore every in-pane helper CLI) reads it. An agent pane holds the key **by design**, not by leak.

The paragraph gets a **dated correction block** rather than a silent rewrite — a reader who remembers the old claim should be able to see that it changed and why.

This closes the last open item in the env-inheritance retro's §9. Worth noting only **one** of the three "contradictory" specs was actually wrong: `docs/MUXSH.md` and `docs/agent-identity-bootstrap.md` both describe agent panes and were right all along. The retro's ledger now records that.

## Two findings for whoever owns the docs gates

Neither blocks this PR; both came out of running the new gates against these edits.

**A blind spot the gates cannot see.** R3 requires an `implemented` spec to cite a PR, so the gates police specs claiming *too much*. A spec claiming *too little* — `proposed` when it actually shipped — is invisible to all six. That is exactly failure #1 above. I tried a detector (flag a `proposed` spec naming a symbol that now exists in source) and it is too noisy to ship: it fires on any spec that merely *mentions* an existing tool. Reporting rather than proposing a seventh gate.

**A false positive from widening the gate in #3369.** `check-spec-citations.sh` matches `docs/[A-Za-z0-9_-]+/…\.md` with no left boundary, so a path in the *sibling* `agentmux-docs` repo — `agentmux-docs/src/content/docs/internals/agent-app-api.md` — matches as `docs/src/content/docs/internals/agent-app-api.md` and fails as dangling. It is a pre-existing citation in §10 that I inherited by touching the file.

I fixed it in the prose rather than in your gate, since a bare `docs/src/content/docs/…` reads as *this* repo's tree and the ambiguity was worth removing regardless. But the matcher will hit this again for any external path ending in `…docs/<dir>/<file>.md`, and the obvious boundary fix (require the match not be preceded by `[A-Za-z0-9_.-]`) also drops coverage of repo-prefixed citations like `agentmux/docs/specs/X.md` — a real trade-off, and your call, not mine.

## Testing

All six doc gates green, run **after** committing (they diff against the merge-base, so running them pre-commit reports a false all-clear):

`check-doc-status` · `check-spec-citations` · `check-doc-links` (920 links) · `check-docs-dir-map` · `check-docs-lifecycle` (0 newly violating) · `gen-docs-index --check`

Docs only — no code changes.


a5af:
@codex review

PR #3378, opened 2026-09-18
Manoz@Area54: docs: add retro and analysis to the Status vocabulary, and stamp them

GenericAgentX-asaf:
## The finding

The closed enum — `draft | proposed | active | implemented | living | historical | superseded` — was designed for **specs**. Specs are ~974 of the 1,484 tracked docs. `docs/retro/` (180) and `docs/analysis/` (141) are not specs and **never reach any of those states**: a retrospective is never "implemented", and an audit of a question at a point in time is not "active".

So the backlog looked like neglect and wasn't. Of 610 non-compliant docs, **288 were in those two directories**, and their Status lines were free text *precisely because no enum word fit*:

| actual Status text | count |
|---|---:|
| `root-caused; fix in …` | 45 |
| `analysis, not implemented` | 25 |
| `shipped` | 11 |

## The change

The enum grows by two words, rather than ~290 documents being forced into a word that misdescribes them.

| | before | after |
|---|---:|---:|
| compliant | 873 | **1,163** |
| non-compliant | 610 | **321** |
| compliance | 58.8% | **78.4%** |

Stamped: **98** with a non-vocabulary word (original prose preserved after the enum word — `**Status:** retro — COMPLETE, all four phases shipped`), **8** in an unreadable format, **164** with none at all. **51 left alone** — already valid, or their status lives in YAML frontmatter, which isn't this gate's business.

## Why this isn't the bulk restamp §4 forbids

`PLAN_DOCS_CLEANUP_EXECUTION_2026_09_01.md` §4 exists to stop *"unknown status"* becoming *"confidently wrong status"* — and it's about **lifecycle claims**.

Stamping a spec `implemented` asserts something that must be verified against code. Stamping a file in `docs/retro/` as `retro` asserts only **what kind of document it is** — which its own directory already asserts. No claim is made that could turn out false, and none of the 164 insertions invents a state.

## What's deliberately untouched

**`docs/specs/`.** A spec's Status *is* a lifecycle claim, so its non-compliant files still need a human to read them. That's the irreducible remainder, and it doesn't shrink by adding words. Anyone who wants that number to fall further has to do the reading.

## Verification

- `check-doc-status.sh` over the changed set → ok
- `check-doc-links.mjs --all` → 0 broken across 1,490 docs
- `check-docs-dir-map.mjs` → ok (17 directories)
- No `INDEX.md` regeneration needed — the only `docs/specs/` file touched is `README.md`, which the generator excludes


a5af:
@codex review

PR #1498, opened 2026-06-16
feat(mcp): Loop + LoopStop tools — recurring prompt injection (Claude /loop analogue)

a5af:
## What

Adds AgentMux's equivalent of Claude's `/loop` as two new `agentmux-mcp` tools, so agents get them alongside `Shell`/`SendMessage`/`OpenEditor`:

- **`Loop(prompt, interval?, to?, immediate?)`** → `loop_id`. Re-injects `prompt` (a prompt or slash command) into a conversation on a recurring interval. `interval` accepts `30s`/`5m`/`1h` (bare number = minutes), default `10m`, clamped `[10s, 24h]`. `to` defaults to **self** (`AGENTMUX_AGENT_ID`) — a self-loop; pass it to nudge a peer agent. `immediate` runs once on start too.
- **`LoopStop(loop_id)`** → aborts the loop (idempotent).

## How

A loop is just *"re-inject this prompt every N"*. The delivery primitive already exists — `POST /agentmux/reactive/inject` (`reactive_handler.inject_message`), the same path `SendMessage` uses, with `wait_for_idle` so a re-prompt lands when the target finishes its turn.

Each loop is an **in-process `tokio` task in the agentmux-mcp server** (Claude's `/loop` is likewise client/tool-driven, not backend). A `Mutex<HashMap<loop_id, JoinHandle>>` registry tracks them; `LoopStop` aborts by id. The MCP process is long-lived for the agent session, so **loops stop automatically when the agent pane closes** — correct lifecycle, no orphans.

**No srv changes** — would mean new endpoints + an AppState registry + scheduler for a feature that's purely a timer over an existing endpoint. Trade-off: loops don't persist across an MCP-process restart (acceptable; a loop is session-scoped — srv-side persistence is a possible follow-up).

## Verification

- `cargo check -p agentmux-mcp` — green.
- ⚠️ End-to-end (the inject actually re-prompting the agent) needs an AgentMux build to smoke.

Spec: `docs/specs/SPEC_MCP_LOOP_TOOL_2026_06_16.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3379, opened 2026-09-18
Manoz@Area54: docs: correct the A1-A15 board — two shipped items were still marked open

GenericAgentX-asaf:
The board is the **handoff surface** for this refactor — its own instructions say *"Pick any 🟢 item."* It was wrong about two of fifteen, in the direction that wastes someone's time: telling them work is available that's already done.

| item | board said | reality |
|---|---|---|
| **A6** — collapse agent-pane's 4 parallel state systems | 🟡 "mirror killed" | ✅ shipped **2026-09-06**; its spec reads `implemented — this PR`. Stale for 12 days |
| **A10** — consolidate data-dir resolution onto `DataPaths` | 🟢 ready | ✅ shipped today in #3372 |

**Real state: 14 of 15 done.** Only **A2** is open, and it's blocked on *coordination*, not effort — the `agentmux-common` DTO extraction still collides with a5af #1498, and no DTO module exists in that crate, so nothing has moved since June.

A10's row also now records what it actually found, because the board undersold it as "cleanup": srv had its own root resolver falling back to `PathBuf::from("/")`, so a host with no resolvable home wrote to `/.agentmux`.

**The pattern worth naming:** this is the same failure the docs-lifecycle work has been chasing all week — a status surface trusted without being checked. The gates now catch it for `Status:` lines, but a markdown table of 🟡/🟢 markers is not something any gate reads. Worth considering whether boards like this should carry machine-checkable state at all.


a5af:
@codex review

PR #3380, opened 2026-09-18
Manoz@Area54: fix(rpc): make ORef generate the TypeScript type it actually sends

GenericAgentX-asaf:
## The defect

`ORef` is `{ otype: String, oid: String }` in Rust and a single `"otype:oid"` **string** on the wire — it has a hand-written `Serialize`/`Deserialize` pair, and its own doc comment already said *"Wire format: `block:550e8400-…`"*.

Everything agreed except what codegen would produce.

ts-rs derives from the **struct shape**. The `serde-compat` feature — which this crate's `Cargo.toml` says exists so *"the emitted type matches the wire shape serde actually produces"* — only reads serde **attributes** (`rename_all`, `default`, `skip_serializing_if`). It cannot see an `impl Serialize`.

So `ORef` is exactly the case serde-compat can't cover. Any command carrying one would have generated `{ otype: string, oid: string }` — a shape no client ever receives, that TypeScript would believe, and that no compiler on either side could catch.

## Two attributes that look right and aren't

Worth recording, because both are the obvious first try:

| attempt | result |
|---|---|
| `#[ts(as = "String")]` | consumer still generates `{ oref: ORef }` |
| `#[ts(type = "string")]` | consumer still generates `{ oref: ORef }` |

Both leave `name()` as `"ORef"`, and ts-rs emits a **reference to that name** at every use site. For a field to come out as `string`, `name()` itself has to be `"string"` — which needs a hand-written `impl TS`, mirroring what ts-rs does internally for foreign string-like types (`impl_primitives! { uuid::Uuid => "string" }`).

**I only know that because the test was written first and failed against both.** It would have been very easy to ship either attribute with a confident message and no working fix.

## The test

It asserts what a **consumer** generates, not `ORef::name()` — a field is what any real command actually has — and pins the serialized form alongside it, so if either half drifts the other catches it.

## No current caller

The three commands carrying an `ORef` (`CommandGetMetaData`, `CommandSetMetaData`, `CommandMessageData` in `rpc_types/block.rs`) derive `Serialize`/`Deserialize` but **not** `TS` — they're part of the `websocket.rs` cluster codegen hasn't reached. This is groundwork so the binding is right when it does, rather than confidently wrong.

That cluster is parked for a better reason than §5.4h records: not churn, but three unresolved modelling questions — this one, `MetaMapType` (an untyped `map[string]any` inherited from Go), and **two different `BlockDef` structs** (`backend/obj.rs:235` vs `backend/wconfig/types.rs:481`, differing in whether `files` is optional).

## Verification

`cargo test -p agentmux-srv oref` → 9 passed.


a5af:
@codex review

PR #3381, opened 2026-09-18
Manoz@Area54: test(obj): pin the wire-equivalence of the duplicated BlockDef/FileDef

GenericAgentX-asaf:
## What I found

This crate defines `BlockDef` and `FileDef` **twice**:

| | `backend/obj.rs` | `backend/wconfig/types.rs` |
|---|---|---|
| feeds | `rpc_types/block.rs` (RPC payloads) | `WidgetConfig` |
| `BlockDef.files` | `Option<HashMap<..>>` | `HashMap<..>` |
| `FileDef.meta` | `Option<HashMap<..>>` | `HashMap<..>` |

They look like a latent bug. **They aren't — today.** Every field on both carries `skip_serializing_if`, so an absent map and an empty one both vanish, and the two serialize identically. I verified that rather than assuming it, and this test is the check, kept.

## Why pin it

**The frontend has one hand-written contract for each.** `frontend/types/srv-types.d.ts` declares `type BlockDef = { files?, meta? }` — still tagged `// waveobj.BlockDef` from the Go original. Both Rust structs satisfy it now. If either drifts (someone drops a `skip_serializing_if`, or makes a field required) one starts lying to that contract, and **nothing else in the repo would say so**.

**And it's a codegen hazard.** When the `websocket.rs` commands are eventually typed, two structs sharing a name generate two `BlockDef.ts` files into the same export directory — one silently overwriting the other. Knowing the shapes agree is the precondition for merging them, or for picking either as canonical.

## Non-vacuity

Removing `skip_serializing_if` from wconfig's `files` fails the test with:

```
assertion `left == right` failed: the two BlockDef structs disagree on the empty case
  left: "{}"
 right: "{\"files\":{}}"
```

Restored, passes.

## What this deliberately does not do

**It does not fix the duplication, and shouldn't be read as doing so.** It stops the duplication becoming a *divergence* while nobody is watching. Actually merging the two pairs spans two domains — the RPC object model and the widget config format — and deserves its own PR and its own argument.

This is the third of three things blocking the `websocket.rs` codegen cluster. `ORef` is fixed (#3380). `MetaMapType` — an untyped `map[string]any` inherited from Go, which generates `Record<string, any>` — is the one still open.


a5af:
@codex review

PR #3382, opened 2026-09-18
Manoz@Area54: feat(docs): flag specs that claim to be unbuilt while the code cites them

GenericAgentX-asaf:
**§5.6** of `REPORT_MIGRATION_WRAPUP_STATUS_2026_09_16.md`, which calls it *"the highest-leverage item in this report — the one that stops the next audit from being necessary."*

## The check

The sweep already asks *"does this doc claim to be current while the code it cites moved?"* This adds the cheaper, opposite question: **does a doc say `draft`/`proposed` while source code names it?** Somebody built the thing and never came back to the Status line.

§3.1 records what that costs. A spec read *"Phase 0 implemented … no render-path wiring yet"* for **fifteen months** after Phase 3 landed. Two independent readers believed it — and the second produced an ROI argument and a phase-by-phase estimate for three phases that had **already shipped**. It was caught by `grep -n tanstack package.json` returning nothing.

## It found 279, and the top one is embarrassing

Ranked by how many files cite them. The sharpest:

> `SPEC_RPC_BINDINGS_CODEGEN_2026_09_07.md` — **Status: proposed** — *"No generator exists yet; this is the plan and the reasons for its shape."*

Meanwhile **187 of 235 RPC registrations are already generated** (~80%), and that spec is cited from **16 source files**, including `rpc/engine.rs` and `rpc/schema.rs`.

The codegen effort is most of the way done and its own spec says it hasn't started. I'd been quoting 79.6% from the wrap-up report all day while that spec sat there saying zero — exactly the §3.1 trap, live.

Others near the top: `SPEC_INAPP_CLAUDE_OAUTH_LOGIN` (23 files), `SPEC_AGENT_PICKER_TWO_TIER` (21), `SPEC_ACCOUNT_DELETE_DEAUTH_LAYERS_2_4` (17).

## Two scoping decisions, both to keep the signal usable

- **`active` is not flagged.** It means "partially implemented, here's what shipped" — being cited from source *confirms* it rather than contradicting it. Only `draft` and `proposed` assert the thing isn't built.
- **`.md` is excluded from the citation scan.** Specs discuss each other constantly; a spec referenced in another doc says nothing about whether it shipped. Only real source extensions count.

Without the first, ~50 `active` specs would flood the list with non-findings. Without the second, essentially every spec would flag.

## Signal, not gate

Nothing here fails a build, and the report is explicit about why: deciding whether a status is *true* means comparing a claim to shipped code — a judgement, not a grep. This produces a triage list for a human, in the weekly sweep that already exists and already posts to a standing issue.

## Tests

19 pass (14 existing + 5 new), covering both halves — including the `active` exclusion, that docs citing docs don't count, and that an unreadable source file is skipped rather than aborting the sweep.


a5af:
@codex review

PR #3386, opened 2026-09-18
Manoz@Area54: docs(specs): correct 17 more Status lines (batch 3)

GenericAgentX-asaf:
Batch 3 of the §5.6 reverse-check triage (batch 1 #3384, batch 2 #3385, both
merged). **17 more Status lines**, each read against the code first.

## Implemented (12)

| Doc | Said | PRs |
|---|---|---|
| `SPEC_STORE_MODULARIZATION_2026_05_27` | "Design — multi-PR refactor proposal. No tracking discussion yet." | #1100, #1116, #1126, #1127, #1128, #1129, #1135 |
| `SPEC_AGENT_DISPATCH_SUBAGENT_HIERARCHY_2026_07_17` | "design only, no implementation yet" | #2208 |
| `SPEC_AGENT_PICKER_FILTER_SEARCH_2026_08_17` | "needs answers to the two open questions before implementation" | #2621 |
| `SPEC_BROWSER_DOM_API` | `draft` | #453 |
| `SPEC_STATUSBAR_TOKEN_USAGE_2026_04_24` | "Draft, ready to implement" | #549 |
| `SPEC_AGENT_PANE_STATE_MACHINE_2026_05_23` | `Draft` | #996 |
| `command-palette.md` | `Draft` | #309 |
| `lan-discovery-toggle.md` | `Draft — design` | #1032 |
| `SPEC_WIDGET_BAR_PARENT_SUBMENUS_2026_08_12` | `proposed` | #2633 |
| `SPEC_BACKGROUND_TASK_TEARDOWN_SURVIVAL_2026_08_20` | `Proposed` | #2681 |
| `SPEC_RESPONSIVE_TAB_BAR_TOP_POSITION_2026_08_24` | `proposed` | #2793 |
| `SPEC_ISOLATE_HOST_CLAUDE_MD_2026_08_31` | `Proposed` | #2854 |

## Active — partially built, unbuilt half named (4)

| Doc | Shipped | Still open |
|---|---|---|
| `REPORT_MUXSPECT_MUXLOG_CROSS_CHANNEL_INSPECTION_2026_08_22` | §2.2–2.3 #2738, Ext 4 #2745 | **Ext 1 — named in the report as the actual root fix** — plus Ext 2, 3, 5, 6 |
| `REPORT_AGENT_PANE_STATE_RECONCILIATION_2026_07_07` | Finding 2 #2002, Finding 1 #2241 | Finding 3 — no concurrency cap in `agentmux-srv/src/ambient/` |
[...]

PR #3385, opened 2026-09-18
Manoz@Area54: docs(specs): correct 23 more Status lines, and reconcile the two Status gates (batch 2)

GenericAgentX-asaf:
Batch 2 of the §5.6 reverse-check triage. Batch 1 (#3384, merged) corrected 24
and added `scripts/docs-status-origin-survey.mjs`, the tool that ranks this
queue. This is **23 more**, each read against the code before being touched,
plus a fix for two gates that disagreed with each other.

## Implemented (16)

| Spec | Said | PRs |
|---|---|---|
| `SPEC_AGENT_LOGIN_FLOW_TIGHTENING_2026_09_04` | "analysis complete, not yet implemented" | #2971 |
| `SPEC_RESIZE_DEFAULT_FLIP_AND_WINDOW_EDGE_SHIFT_2026_08_26` | "analysis + design — not implemented" | #2822 |
| `SPEC_AMBIENT_GHOST_TEXT_NEXT_PROMPT_2026_07_03` | "not yet implemented" | #1961 |
| `SPEC_BROWSER_PANE_HTTP_BASIC_AUTH_2026_05_18` | `Draft` | #906, later #2824 |
| `SPEC_LAUNCH_MODAL_STATE_MACHINE_2026_05_19` | `Draft` | #916, #917 |
| `SPEC_BROWSER_PANE_BOOKMARKS_AND_GO_ICON_2026_08_22` | `Draft` | #2730 |
| `SPEC_BROWSER_PANE_LOADING_INDICATOR_FLICKER_2026_08_17` | `Draft` | #2642 |
| `SPEC_USER_INPUT_VISIBILITY_AND_STARTUP_COLLAPSE_2026_05_24` | `Draft` | #1020 |
| `SPEC_EDITOR_FILE_TREE_2026-05-26` | `Draft — design` | #1064 |
| `SPEC_PANE_FILE_DROP_2026_05_30` | `Draft` | #1201 |
| `SPEC_PROVIDER_SYSTEM_PREREQS_2026_05_18` | `Draft` | #908 |
| `SPEC_PANE_DRAG_TO_TAB_2026_07_10` | `Draft (v3 …)` | #2073, #2079, #2086, #2095, #2310 |
| `SPEC_PROVIDER_AWARE_STARTUP_INSTRUCTIONS_2026_08_24` | `proposed` | #2788 |
| `SPEC_TOOL_BLOCK_LIVE_LOG_2026_05_11` | `Proposed` | #800, #804 |
[...]

PR #3384, opened 2026-09-18
Manoz@Area54: docs(specs): correct 24 Status lines the code contradicts, and rank the rest

GenericAgentX-asaf:
Batch 1 of the §5.6 reverse-check triage that landed in #3382: a spec whose
`**Status:**` claims it was never built, cited by source files that build it.
279 candidates. This corrects **24** and adds a tool that orders the remaining 255.

Every one was read against the code before its Status line was touched. That
matters: **seven of the 24 came back `active`, not `implemented`** — a bulk
restamp would have gotten all seven wrong, trading an unknown status for a
confidently wrong one.

## Implemented (17)

| Spec | Said | PRs |
|---|---|---|
| `SPEC_INAPP_CLAUDE_OAUTH_LOGIN_2026_08_03` | `PROPOSED (implemented — see note below)` | #2410, #2414 |
| `SPEC_ACCOUNT_DELETE_DEAUTH_LAYERS_2_4_2026_07_14` | "implementation dispatched" | #2164, #2161 |
| `SPEC_CTRL_SHIFT_SCROLL_ZOOM_ALL_PANES_2026_09_07` | `Proposed` | #3090 |
| `SPEC_AGENT_PICKER_TWO_TIER_2026_05_24` | "needs answers … before implementation" | #1011, #1012 |
| `SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07` | "§5 is **not** implemented" | #3091, #3132, #3136, #3151 |
| `SPEC_ARMORY_REACTIVE_UPDATES_2026_09_02` | `Proposed` | #2932 |
| `SPEC_AGENT_HISTORY_AS_TAB_AND_DRAFT_PRESERVATION_2026_08_11` | `proposed` | #2539 |
| `SPEC_BACKGROUND_TASK_PID_CAPTURE_2026_08_20` | `Proposed` | #2681 |
| `SPEC_BACKGROUND_TASK_DASHBOARD_INTELLIGENCE_2026_08_20` | `Proposed` | #2681 |
| `SPEC_SOUND_NOTIFICATIONS_2026_06_05` | `Draft v1 — for review` | #1301 |
| `SPEC_SLASH_COMMAND_ARCHITECTURE_2026_04_14` | `Draft` | #379, #381, #382, #383 |
| `SPEC_PRE_LAUNCH_OAUTH_FLOW_2026_05_14` | `Draft` | #840 |
[...]

PR #3387, opened 2026-09-18
Manoz@Area54: ci(docs): stop specs shipping as "proposed" in the PR that implements them

GenericAgentX-asaf:
Prevention for the thing the §5.6 reverse check (#3382) keeps finding *after the
fact*.

Triaging 64 of its 279 candidates by hand across #3384, #3385 and #3386 turned
up one cause behind most of them: **the spec was added to the tree by its own
implementing PR**, carrying a `**Status:**` written before that PR merged, which
nobody re-read on the way past. That is `feedback_no_doc_only_prs` working
exactly as designed — the doc rides with the code — plus a missing last step.

What it leaves behind:

| Spec | Status it shipped with |
|---|---|
| `SPEC_CTRL_SHIFT_SCROLL_ZOOM_ALL_PANES` | `Proposed`, in a file containing a section titled *"found during implementation"* |
| `SPEC_AGENT_PICKER_TWO_TIER` | *"needs answers to the two decision points"* — in the same commit that added the section answering both |
| `SPEC_INAPP_CLAUDE_OAUTH_LOGIN` | `PROPOSED (implemented — see note below)`, plus its own audit note reading *"Status field was never updated"*, from six weeks earlier |

Each was cheap to prevent and expensive to find later. This is the prevention;
the sweep's reverse check stays as the backstop for specs that go stale some
other way.

## The signal, and why it isn't the obvious one

Fires only when this PR **adds** a doc **and** source files in the same PR cite
that doc **by name**.

Deliberately *not* the weaker "adds a doc and touches source" —
`docs-status-origin-survey.mjs` uses that to rank the existing backlog, and it
has real false positives. `SPEC_DECISION_PROMPT_2026_04_24` was introduced by
`feat(statusbar): always-show connection widget` (#553), a PR with nothing to do
with it, where `Draft` was the honest status. Failing that PR would teach people
to route around the check, which is worse than not having it.

A PR can legitimately add a proposal alongside unrelated code. It cannot
legitimately add a proposal that its own new code implements.

## Replayed against the real commits

Each run against its own parent:

| PR | Doc | Status | Citing sources | Result |
|---|---|---|---:|---|
| #3090 | zoom | `proposed` | 13 | **caught** |
| #2932 | armory reactive | `proposed` | 14 | **caught** |
| #1301 | sound notifications | `draft` | 12 | **caught** |
| #1301 | tool-call tones | `draft` | 4 | **caught** |
| #553 | decision prompt | — | — | added no doc in that commit; nothing to check |

## One known miss, and why it's acceptable

Replaying #1011 (the two-tier picker) shows 13 citing sources but **passes** —
its Status read `Design analysis — …`, and `design` is neither `draft` nor
`proposed`.

That's a pre-vocabulary doc; the closed `**Status:**` enum didn't exist in May.
`check-doc-status.sh` now runs on added files too and rejects anything outside
the enum, so a doc added *today* can't reach this gate saying `design`. The two
gates compose. Widening the unbuilt set to "anything not `implemented`" would
instead fire on `historical`, `superseded`, `living`, `analysis` and `retro` —
all legitimate things for a new doc to be.

## Scope

**Added docs only.** It can never fire on the backlog it exists to stop growing;
editing a stale spec stays the reverse check's job. A PR with no new docs exits
immediately.

## Tests

11 unit tests over the pure parts — `statusOf`, `docName`, `isSourcePath`,
`isViolation` — with no git and no filesystem, in the same `.test.mjs` +
vitest shape as `docs-stale-sweep.test.mjs`. They include the #553 shape
(proposal, zero citing sources → passes) and the real
`PROPOSED (implemented — see note below)` parse.


a5af:
@codex review

Issue #1469, opened 2026-06-16
Phase 2: route tool-permission can_use_tool to the decision-prompt UI (#551)


AgentU-asaf:
Follow-up to #1451 (Agent SDK control protocol — `docs/specs/SPEC_AGENT_CONTROL_PROTOCOL_2026_06_15.md`).

## Context
#1451 made the persistent controller speak the Agent SDK control protocol to answer AskUserQuestion. As part of that, Claude now launches with `--permission-prompt-tool stdio --permission-mode default`, so the CLI sends a `can_use_tool` `control_request` for **every** non-auto-allowed tool. Today the control handler in `persistent.rs` **auto-allows** all non-AskUserQuestion tools (to preserve the prior `--dangerously-skip-permissions` yolo UX).

The control channel is exactly the transport `SPEC_DECISION_PROMPT_2026_04_24.md` (#551) flagged as missing — "the agent CLI asks the subprocess's stdin for y/n ... but AgentMux isn't listening."

## Scope
- In `persistent.rs::handle_control_frame`, route non-auto-allowed `can_use_tool` requests to the existing `AgentDecisionPanel` instead of blanket-allowing them, honoring the user's `--permission-mode` / settings.
- Reply with the control_response `behavior: "allow" | "deny"` (+ `updatedInput` for approve-with-changes, `message` for deny — see the SDK `can_use_tool` shapes captured in the spec §2.3).
- Persist allow-rules (the "remember last choice" goal in `SPEC_DECISION_PROMPT` G4) so repeat calls skip the prompt.
- Keep auto-allow as the default until rules/mode say otherwise, so we don't regress to prompting on every tool.

## Notes
- Watch permission chatter in `default` mode; consider `acceptEdits` or allow-rules to cut volume.
- Subtypes beyond `can_use_tool` (`set_permission_mode`, hook callbacks, `control_cancel_request`) are currently logged + safe-defaulted; enumerate them here.

PR #3388, opened 2026-09-18
Manoz@Area54: fix(agent): a runtime change silently disabled AskUserQuestion on persistent agents

GenericAgentX-asaf:
Found while scoping the decision-prompt work (#551 / #1469 — the migration
wrap-up's §5.3, "the highest-value functional item left").

## The invariant, and the code that broke it

Both provider catalogs state this in capitals — `providers/catalog.ts`'s
`claude` entry and `static CLAUDE` in `agentmux-srv/src/backend/providers.rs`:

> **CRITICAL:** `--dangerously-skip-permissions` **DISABLES** that routing (it
> bypasses `canUseTool`), so it must **NOT** be in `persistent_launch_args`.

Nothing enforced it.

`buildRuntimeArgs` is what rebuilds a **running** agent's args on every
`/model`, `/effort` or permission-mode change (`runtime-apply.ts`,
`useAgentCommands.ts`). With the default runtime config —
`permissionMode: "bypass"` — it stripped `--permission-mode default` and
appended `--dangerously-skip-permissions`.

## The chain, verified end to end

1. User runs `/model opus` on a persistent Claude agent.
2. `runtime-apply.ts` rebuilds args via `buildRuntimeArgs`.
3. `bypass` → strips `--permission-mode`, appends the bypass flag.
4. Result is written to `cmd:args` block meta.
5. `persistent.rs`'s spawn reads `cmd:args` **fresh from meta on every spawn**.
6. The CLI stops sending `can_use_tool`.
7. `pending_questions` never fills — **AskUserQuestion is unanswerable for that
   agent from then on**, permanently, because the meta persists.

So the *first* runtime change on an agent quietly took it out of the control
protocol. It also escalated that agent to genuine `--dangerously-skip-permissions`
bypass — the opposite of what a user touching a permission-mode setting would
expect.

Measured directly before fixing, feeding Claude's real `persistentLaunchArgs`
through the function with the default config:

```
["-p", ..., "--permission-prompt-tool", "stdio",
 "--dangerously-skip-permissions", "--model", "sonnet", "--effort", "high"]
```

`--permission-mode default` gone, bypass flag added, `--permission-prompt-tool`
left behind pointing at a protocol that is now off.

## The fix

If the base args carry `--permission-prompt-tool`, this agent speaks the control
protocol: never hand it the bypass flag, and map `bypass` onto
`--permission-mode default` instead.

**The user-visible "yolo" behaviour is unchanged**, because it never came from
that flag — the persistent controller's `ControlChannel` auto-allows every tool
except AskUserQuestion (`persistent.rs::handle_control_frame`). An explicitly
chosen non-bypass mode is still honoured, and the non-persistent claude path is
untouched.

## Tests

Five new, **two of which fail without the source change** (verified by stashing
the fix and re-running):

- never hands a control-protocol agent the bypass flag ← fails without
- keeps the transport intact, `bypass` → `default` ← fails without
- still honours an explicitly chosen non-bypass mode
- **is idempotent** — `runtime-apply` writes its output to `cmd:args` and the
  *next* change rebuilds from **that**, not from the catalog, so a
  non-idempotent transform would drift a little further on every `/model`
- leaves the non-persistent claude path alone

The existing 10 tests are unchanged and still pass.

## Relationship to #551

This is a **prerequisite** for Phase 2 rather than part of it. Routing
`can_use_tool` to the decision panel is worth nothing on a transport that
switches itself off after the first `/model`.


a5af:
@codex review

PR #3389, opened 2026-09-18
Manoz@Area54: feat(agent): backend plumbing for per-tool-call permission decisions (Phase 2, gated off)

GenericAgentX-asaf:
Scopes #551 / `SPEC_DECISION_PROMPT_2026_04_24.md`'s Phase 2 and
`SPEC_AGENT_CONTROL_PROTOCOL_2026_06_15.md`'s §5 Phase 2 — the migration
wrap-up's §5.3, called "the clearest half-migration... the highest-value
functional item left" — down to the part that's unambiguous: the **backend
mechanism** to park an ordinary (non-AskUserQuestion) tool-permission
request and actually deliver a decision over the control protocol, mirroring
`answer_question`/`deny_question` exactly.

## What this does NOT do, deliberately

**Change any default behavior.** `should_route_to_decision_panel` is
hardcoded `false` — every tool call is still auto-allowed exactly as today.

That's not caution for its own sake. `SPEC_AGENT_CONTROL_PROTOCOL`'s own §7
risk section already names the danger of getting this wrong:

> **Permission chatter:** `--permission-mode default` may route many tools
> through `can_use_tool`. Mitigation: auto-allow in the handler (Phase 1)

`--permission-mode default` is the mode baked into every persistent Claude
agent's launch args today. Flipping the gate to "route everything" would turn
every agent into a prompt-per-tool-call experience the instant this shipped —
a real regression the spec's own author already flagged, not a hypothetical
I'm inventing caution about. Which tools/modes actually warrant a prompt is a
policy decision (a risk-tiering scheme, `acceptEdits` instead of `default`,
or persisted allow-rules per the spec's own unbuilt §6) — not something an
engineering PR should decide unilaterally.

## New, and independently tested

- `PersistentInner::pending_permissions` — the sibling of
  `pending_questions`, same in-memory-only, per-controller-instance
  lifetime.
- `park_tool_permission_request` — extracted as its own function
  specifically so it's unit-testable directly, independent of the
  (currently always-false) gate.
- `PersistentSubprocessController::decide_tool_permission` — allow echoes
  the original input verbatim (no edit-before-approve UI exists yet); deny
  carries user feedback verbatim into the CLI-facing message per the spec's
  G6 ("Denials carry user-typed feedback verbatim to the agent"). Same
  dead-air fallback as `answer_question`/`deny_question`.
- `tooldecision`'s RPC handler now actually calls it instead of logging and
  dropping the decision (previously: `"[tooldecision] received (delivery
  mechanism deferred to PR-3b/PR-4)"`), mirroring `agent.answer`'s existing
  controller-resolution pattern.

## Found and fixed along the way

`CommandToolDecisionData.request_id` was documented as correlating against a
`PermissionRequestEvent` (§4.1 of the spec) that was never built — superseded
by the control-protocol design entirely. It's repurposed here as the
`tool_use_id` key the real mechanism actually uses, matching how
AskUserQuestion already correlates. Kept the field **name** unchanged to
avoid unrelated frontend/TS-binding churn for a field whose producer — the
frontend event wiring below — doesn't exist yet.

## What's still missing, named rather than silently deferred

- **The frontend half.** Unlike AskUserQuestion, the frontend can't infer
  "this tool_use is pending" from tool identity alone — any tool could be
  auto-allowed or parked depending on the gate — so a real WPS/MPS push
  event is needed to flip a `ToolNode` to `pending_approval` at all. Not
  built here; this is the natural next PR once the policy question below is
  answered.
- **§6's rules-persistence layer** (`permissions.json`) — `scope` is
  validated but not acted on, same as before this PR.
- **The actual policy decision** behind `should_route_to_decision_panel` —
  which tools/modes should prompt, and what "gate it on `acceptEdits`
  instead of `default`" vs. "build allow-rules" costs each — is a product
  call, not mine to make.

## Testing

7 new tests in `persistent.rs`: park + decide allow, park + decide deny with
feedback, deny with no feedback still tells the model it was refused, the
untracked-id error path, the resume-message builder, and one pinning the
gate's current inertness across a representative tool set (`Bash`, `Edit`,
`Write`, `WebFetch`, `AskUserQuestion`, empty string) — the test that must
change, deliberately, the day this is flipped on for real.

Full `agentmux-srv` suite: **3979 passed, 0 failed, 15 pre-existing ignored.**
`check-rpc-bindings` regeneration on `CommandToolDecisionData.ts` is
doc-comment-only — no shape change, diffed and confirmed before committing.

Follows #3388 (the persistent-args bug that was silently disabling the
control protocol this all depends on).


a5af:
@codex review

PR #3391, opened 2026-09-18
perf(agent-pane): keep agent tabs mounted across a pane-tab-strip switch

AgentA-asaf:
## What

Same-pane agent-tab switching was slow — every switch fully unmounted the old agent's `<Block>` and mounted a fresh one, paying for `useHistoryPagination`'s cold-start RPC chain (session read -> line-count -> up to 5,000 lines of transcript) plus a full virtualizer rebuild, every single time instead of just on first visit. Terminal tabs in the same strip don't pay this cost because `"term"` is already in `pane-leaf-chrome.tsx`'s `KEEP_ALIVE_TYPES` (mount once, toggle visibility on switch).

This adds `"agent"` to that set.

## Why it was safe to do

`KEEP_ALIVE_TYPES`'s own comment previously warned agent's per-tab state (quick-fork, launch-in-place) was "not yet audited for keep-alive safety." Did that audit before flipping the flag — full writeup in `docs/specs/SPEC_AGENT_PANE_TAB_KEEPALIVE_2026_09_18.md`. Summary:

- **Quick-fork** and **launch-in-place** both reset via mechanisms already internal to a single Block (an in-place `<Show when={agentId()}>` swap, a `pushBlockOntoStack`) — neither relies on the leaf remounting the Block, so keep-alive doesn't disturb them.
- **Two real gaps found and fixed**: `AgentQuestionPanel`'s 30s auto-timeout and `useAgentFailure`'s auto-retry countdown both used to implicitly pause when a backgrounded tab unmounted. Under keep-alive they'd otherwise keep running invisibly — a hidden tab could auto-answer a question the user never saw, or auto-resend a failed turn nobody could cancel. Both are now explicitly gated on a new reactive `isBlockDormant()` (`block-component-registry.ts`), defaulting to "never dormant" so every pre-existing call site is unaffected.
- One pre-existing hazard (a global CLI-login-cancel slot, `useAgentControllerStatus.ts`) is documented but deliberately **not** fixed here — narrower/rarer, and touching that fragile auth surface is out of scope for this change. See the spec's own callout.

## Testing

- New unit tests: `block-component-registry.test.ts` (reactive dormancy), `AgentQuestionPanel.test.tsx` (dormancy pause/re-arm), `useAgentFailure.test.ts` (dormancy pause/resume), `pane-leaf-chrome.test.tsx` (agent keep-alive mount/visibility, mirroring the existing term suite).
- `tsc --noEmit` clean.
- Could not run the frontend test suite locally end-to-end (this machine's Node is v22.22.2; the repo pins >=24.11.0 via `.nvmrc`, and the installed vitest/solid-refresh combination fails to transform ANY test file under v22 — reproduced identically on unmodified `main` before this change, so it's a pre-existing local environment gap, not something this PR introduces). CI runs the pinned Node version.
- Manually verified in `task dev`: switching between two agent tabs in the same pane is now instant, no history-restore flash.


a5af:
@codex review

PR #3392, opened 2026-09-18
fix(pane-tabs): resolve pane chrome's LayoutModel from the node's own tab, not the globally-active one

AgentA-asaf:
## What

A newly-created tab's default agent pane silently ignored every "+"/close/switch action on its own tab strip — reported live: "opened a new window tab, loaded Posa in an agent pane, and try to add another pane tab to the pane posa is in, trying to add anything is a no-op."

## Root cause

`renderPaneChromeShell` (`PaneChrome.tsx`), `buildAgentPaneChromeModel` (`agent-view.tsx`), `buildTermPaneChromeModel` (`term.tsx`), and `PaneLeafChrome`'s own stack lookup (`pane-leaf-chrome.tsx`) all resolved their `LayoutModel` via `getLayoutModelForStaticTab()` (a non-reactive read of `atoms.activeTabId()`) **at chrome-construction time**. Chrome constructs exactly once per pane and never remounts (`SPEC_PANE_TAB_SWITCH_CHROME_STABILITY_2026_09_07.md`), so whichever tab happened to be globally active at that single moment is bound **forever**.

`createTab()` (`tab-actions.ts`) builds a new tab's entire default-preset layout — including its default agent pane — while the tab is still `activate: false`, only calling `setActiveTab` afterward (deliberate, per `SPEC_TAB_CREATION_REVEAL_ARCHITECTURE_2026_09_16.md`). Since the window-level tab bar keeps every tab's content mounted simultaneously (visibility-toggled, never unmounted), the default agent pane's chrome constructs **while its own tab is still inactive** — not a rare race, the deterministic order for every single new tab. Its chrome permanently latches the *previous* tab's `LayoutModel`.

Live log signature that gave this away (correlated in real time by watching both srv and host logs while reproducing on request):
```
INFO  pane.open: block created, placement skipped   block_id=<X>
ERROR [fe] Cannot apply eventbus layout action DeleteNode, could not find leaf node with blockId <X>
```
`pane.open` created a real block server-side every time; the pane's own `findNode` against the wrong tab's tree could never find it, so it fell straight into the "orphaned, clean it up" path — no exception, no notification, just silence.

## Fix

Gave `NodeModel` its own `layoutModel` field (`layout/lib/types.ts`), populated at construction from `getNodeModel`'s own `model` parameter — the correct, tab-scoped model for that node by construction, no lookup, no timing dependency. All four call sites now read `nodeModel.layoutModel` instead of calling `getLayoutModelForStaticTab()`.

Full writeup: `docs/specs/SPEC_PANE_CHROME_LAYOUT_MODEL_TAB_BINDING_2026_09_18.md`.

## Testing

- New regression suite in `PaneChrome.test.tsx` ("resolves the OWNING tab's LayoutModel, never the globally-active one"): mocks `getLayoutModelForStaticTab` as a spy returning a distinguishably-wrong model, and asserts every mutating handler (`onActivate`/`onClose`/`onAdd`) operates on `nodeModel.layoutModel` and never calls the spy at all.
- Updated `pane-leaf-chrome.test.tsx`'s fake `NodeModel`s to carry the new field.
- `tsc --noEmit` clean.
- Reproduced live twice: first on a shared multi-agent instance (confirmed via live log tail while the bug was retriggered on request), then again in an isolated `task dev` build with the fix applied — same "+" click on a freshly-created tab's default agent pane now correctly adds a new pane tab, with no `DeleteNode` error in either log.
- Could not run the frontend vitest suite locally end-to-end (this machine's Node is v22.22.2; the repo pins >=24.11.0 via `.nvmrc`, and the installed vitest/solid-refresh combination fails to transform any test file under v22 — a pre-existing local environment gap, unrelated to this change). CI runs the pinned Node version.


a5af:
@codex review

PR #3393, opened 2026-09-18
Lark@narko: fix(container-agents): inject AGENTMUX_LOCAL_URL so container agents can reach the sidecar

GenericAgentX-asaf:
## Summary

Part of agentmuxai/agentmux#2939 workstream 1 (host integration — the biggest remaining gap for container/"sandbox" agents).

Container agents run via `docker exec` over the Docker socket, which never inherits the host process's env the way a host subprocess does — unlike a host agent, which gets `AGENTMUX_LOCAL_URL` purely via OS env inheritance from `bootstrap.rs`. Without an explicit injection, a container agent has no sidecar URL at all.

- New `rewrite_local_url_for_container()` (`container.rs`): rewrites `127.0.0.1`/`localhost` to `host.docker.internal`, since the loopback value this process sees isn't reachable from the container's own network namespace. Shared by both spawn paths (`input.rs`'s `agentinput` RPC path and `agent_io.rs`'s App API path) rather than duplicated, per #2939's own "keep the two paths in sync" note (workstream 9).
- `create_and_start()` now sets `extra_hosts: ["host.docker.internal:host-gateway"]`, which is what makes `host.docker.internal` resolve on native Linux's default bridge network (Docker Desktop already resolves it automatically; this is a no-op there).
- 3 new unit tests for the rewrite helper.

**Scope note:** verified live against this exact bug (a real container agent's `init` frame showing `mcp_servers: [{"name":"agentmux","status":"failed"}]`) while working on this. This fix compiles and its own tests pass, but full end-to-end verification (`agentmux-mcp` actually connecting) is blocked on a separate, larger gap this PR does not attempt to close: `agentmux-mcp`/`agentmux-bashwrap` are not yet baked into the `ghcr.io/agentmuxai/agent-claude` image at all — that's a multi-stage Dockerfile change plus widening `.github/workflows/container-image.yml`'s build context, tracked as a follow-up on #2939 rather than bundled here.

## Test plan
- [x] `cargo build --release -p agentmux-srv` — clean
- [x] `cargo test --release -p agentmux-srv container::` — 24 passed, 0 failed (3 new)
- [ ] Full end-to-end (`agentmux-mcp` connecting inside a real container) — blocked on the image-build gap noted above, tracked separately on #2939


a5af:
@codex review

a5af:
@codex review

PR #3397, opened 2026-09-19
docs: redact credentials and infra identifiers from public repo

AgentY-asaf:
## Summary
`agentmuxai/agentmux` is public. Two committed files carried material that shouldn't be:

- **`docs/retro/retro-agentmux-corp-pr5-...-2026-08-24.md`** — pasted the literal output of `echo $GH_TOKEN`, leaving a real classic-PAT-format token in the tree since 2026-08-24. **Verified dead** (`401 Bad credentials`) — GitHub secret scanning almost certainly auto-revoked it on push, which is expected for a `ghp_` token hitting a public repo. Redacted regardless, since leaving real-format credentials in-tree is bad practice independent of liveness.
- **`docs/specs/archive/SPEC_TRUST_CENTER_MODALS_...md`** — mapped every agent to its Secrets Manager PAT key name and the AWS account ID. Identifiers rather than credentials, but together they describe the shape of the secret store. Now placeholders pointing at the private inventory, with an inline note explaining the redaction.

## Deliberate non-action
Git history still contains both values. Rewriting a public default branch's history is disruptive and buys little here: the token is dead, and the remaining values are identifiers that cannot authenticate. Flagging rather than hiding that tradeoff.

## Related
A companion change on `security/github-app-identity-migration` adds a "what must not be committed to this repo" rule to `CLAUDE.md`, so this class of leak has a stated rule going forward rather than relying on reviewer memory.

## Test plan
- [x] Confirmed the exposed token returns `401 Bad credentials` against `GET /user` before redacting
- [x] Grepped both files post-redaction for the account ID and any `ghp_` pattern — clean
- [x] Confirmed no other `AKIA`/account-ID/instance-ID hits remain on this branch outside the prohibition text itself


a5af:
@codex review

PR #3398, opened 2026-09-19
Opaz@charlie: fix(pane-chrome): agent/term pane double header when its ViewModel is adopted stale

GenericAgentX-asaf:
## Summary
- Fixes a real double-header bug: an agent (or term) pane could render two stacked headers — the hoisted `PaneChrome` tab-strip header, plus `BlockFrame`'s own inline header underneath (bare, no tab strip/`+`).
- Root cause: `block.tsx`'s ViewModel registry is keyed only by `blockId` and *adopts* an existing ViewModel whenever `viewType` matches, skipping `makeViewModel`. `AgentViewModel`/`TermViewModel.nodeModel` is a plain field frozen at construction, and `noHeader` reads `this.nodeModel.paneChromeHoisted` off that frozen reference forever. A pane's very first mount (before block meta loads) goes through a wrapper with `paneChromeHoisted: false`; once meta loads and `pane-leaf-chrome.tsx`'s `hoisted()`/`keepAlive()` latch true, a second mount for the same blockId goes through a *different* wrapper (`paneChromeHoisted: true`) — but `block.tsx` adopts the first ViewModel instead of creating fresh, so `noHeader()` stays pinned to the stale `false` forever.
- Fix: added `ViewModel.updateNodeModel?(nodeModel)`, called unconditionally by `block.tsx` on every mount attempt (create or adopt) — right next to the existing unconditional `setActiveViewModel` call. `AgentViewModel`/`TermViewModel` implement it to refresh their cached `nodeModel` field.

## Test plan
- [x] `npx vitest run` (block/agent/term/tab suites) — 148 files / 2034 tests passing
- [x] Reproduced live in a `task dev` instance, confirmed via CDP DOM query: 2 panes had `headerCount: 2` (`.pane-stack .block-frame-default-header`)
- [x] Applied the fix, reloaded the dev window via CDP `Page.reload`, re-queried the same panes — both now `headerCount: 1`, every other pane unaffected (still 1)


a5af:
@codex review

GenericAgentX-asaf:
Thanks for the catch — confirmed the P1: all 9 of those view types had the identical pattern.

Rather than adding `updateNodeModel` to all 9 (which I initially did, then found it was still racy — reloading surfaced the same double-header bug on Swarm/CPU, just relocated), I replaced that approach entirely: `BlockNodeModel.paneChromeHoisted` is now a live accessor (`() => boolean`, forwarding `pane-leaf-chrome.tsx`'s `hoisted` memo itself) instead of a frozen boolean snapshot, and `blockframe.tsx`'s `noHeader` is read inline in JSX instead of captured once. This fixes the root cause — a ViewModel-registry adoption can no longer matter, since every wrapper for a leaf shares the same live memo — for all 11 hoisted-chrome view types at once, with a smaller diff than the per-ViewModel patch would've needed (net -1 file vs the previous commit).

Verified live via CDP (`Page.reload` + DOM query for `.pane-stack .block-frame-default-header` counts) across 4 repeated reloads — no recurrence.

GenericAgentX-asaf:

PR #3402, opened 2026-09-19
fix(panes): closing a pane stops every agent tab in it, in order (#3202)

### LINKED ISSUE #3202: ClosePane (and FloatPane) must be blockStack-aware, not leaf-granular
Found during review of `docs/specs/SPEC_AGENT_PANE_LIFECYCLE_CONTROL_2026_09_10.md` (codex P1 on PR #3195), and confirmed to affect the already-implemented `ClosePane` (PR #3196), not just future work.

A layout leaf can host a `blockStack` of multiple blockIds (`frontend/layout/lib/layoutStack.ts`) with one `activeBlockId`. The frontend's own close path (`closeBlockInStack`) removes just the targeted member from the stack, only deleting the leaf's layout node once the last member is closed.

`ClosePane`'s server-side path (`sagas::delete_block::run`) has no such awareness: its `LayoutDeleteNodeByBlock` step resolves the leaf via `getNodeByBlockId`, which matches ANY block in a stack, and deletes the WHOLE leaf regardless of how many blocks it holds. Calling `ClosePane` on one member of a multi-block stack leaves the sibling blocks live in `tab.blockids` but with no layout node — invisible, effectively orphaned.

Needs a server-side stack-aware close: remove/deactivate only the requested member when others remain in the stack, falling back to the existing whole-leaf `delete_block::run` path only when it's the last (or only) member. Same gap applies to `FloatPane`'s planned source-delete step once that's built.

AgentA-asaf:
## What

Closing a pane that holds several tabs now stops **every** agent in it, not just the visible one. Each agent is stopped, and its final state saved, **before** its records are deleted.

Spec: `docs/specs/SPEC_AGENT_PANE_CLOSE_GRACEFUL_SHUTDOWN_2026_09_18.md`. This PR lands the spec plus Phase 0 (measurement) and Phase 1 (ordering + every member). Graceful stop is Phase 2.

Closes #3202 (`ClosePane` MCP tool closed only one block of a stacked pane).

## Why

Live incident, 2026-09-18 (spec §2). AgentA and Manoz were tabs in one pane, and the user closed the pane.
- Only AgentA's block got a `DeleteBlock`.
- Manoz's `claude.exe` kept running with no pane. It stayed listed in Swarm but was unreachable by messaging.
- AgentA itself was hard-killed *after* its block record was deleted, so its exit cleanup failed with `compare-and-clear active pid failed … not found`.

## Changes

- **New `close_pane` saga** (`agentmux-srv/src/sagas/close_pane.rs`), run on every block in the pane. For each block, in order:
  1. mark it closing, so `resync_controller` refuses to respawn it;
  2. unregister it from messaging;
  3. stop the controller and drop its process tracker;
  4. clear `session:active_pid` and mark the instance row `stopped` + `ended_at`, keeping the session id;
  5. only then, `DeleteBlock`.
  
  The layout leaf is removed, and one frontend delete queued, only when no member of it survives. A single id (a tab's own ×) therefore closes that tab and keeps its siblings.
- **`delete_block` saga** runs the same pre-delete shutdown before its reducer dispatch, so a tab's × and the pane's × stop an agent the same way.
- **Frontend:** `onNodeDelete` (`tabcontent.tsx`) calls a new `ObjectService.ClosePane(effectiveStack(data))` instead of `DeleteBlock(data.blockId)`.
- **`ClosePane` MCP** expands its target to every member of the target's pane (`find_leaf_containing_block`).

## Phase 0 findings (spec §5.1)

Measured on the pinned Claude CLI with production stream-json flags:

| Case | Result |
|---|---|
| EOF while idle | exit 0 in 0.4s |
| EOF mid-turn | the CLI **finishes the turn** first (22.8s) |
| interrupt, then EOF | exit 1 in about 2s |

Resume stayed intact in every case. Phase 2 will therefore send interrupt, wait for the result, then EOF, and must treat exit 1 after an interrupt as a requested stop, not a crash.

## Not in this PR

- Graceful stop and waiting for exit (Phase 2).
- `delete_tab` reordering: its last-tab guard can refuse *after* agents would have been stopped (spec §5.2).
- Confirmation, reopen guard, orphan reaper (Phase 3).

## Tests

- Rust (`close_pane::tests`):
  - a stacked pane `[A,B,C]` closes all three, removes the leaf and queues one prune;
  - closing one member keeps the pane and its sibling;
  - final state is saved while the block still exists;
  - blocks spanning tabs are rejected;
  - missing ids are skipped;
  - the closing guard refuses resync.
- Rust (`close_pane_tests`): the MCP close of a background tab closes the whole pane.
- Vitest (`layoutStack.test.ts`): the pane × hands `onNodeDelete` every member; a tab × hands it only that tab.


a5af:
@codex review

AgentA-asaf:
Addressed the reagent P1 (close_pane.rs:154, both reviews) in 296d40eb2: the leaf is removed only when every member's `DeleteBlock` succeeded (`leaf_to_delete.filter(|_| failures.is_empty())`). On partial failure the pane stays and holds the surviving block (its process already stopped); closing it again finishes the job. No unit test for the failure branch: `DeleteBlock` can't be made to fail from a test (the reducer no-ops on a missing block), so the guard is a one-line filter checked by review.

AgentA-asaf:
Re review 5254844070: that review ran on 296d40eb2, which already had the guard (`leaf_to_delete.filter(|_| failures.is_empty())`, just before the layout mutation). Line 154 is the comment above it now. To make the guard explicit, 6c7967597 restates it the way `delete_block` does: `run_inner` returns `Err` on any failed member delete **before** `LayoutDeleteNode` / `queue_source_layout_delete`. The trailing check is gone.

PR #3405, opened 2026-09-19
docs(spec): pane-close shutdown Phase 2 — graceful, one chain for every close

AgentA-asaf:
Follow-up to #3402. It specs Phase 2 of `docs/specs/SPEC_AGENT_PANE_CLOSE_GRACEFUL_SHUTDOWN_2026_09_18.md` (new §9). This PR changes the spec only; no code.

## What Phase 2 does

- **Graceful stop:** a new `Controller::shutdown(deadline)`.
  - An active Claude turn gets the control-protocol interrupt. Then comes EOF, then a force-kill at the deadline.
  - It resolves only once the process has actually exited, and the process tracker is dropped after that.
  - This follows the Phase 0 measurements (spec §5.1): EOF alone doesn't stop a running turn; the interrupt does, in about 2s.
- **One chain for every close:** `shutdown_agents()` runs concurrently with one 5s deadline per close. It is used by all four ways of closing an agent: a pane tab's ×, a pane's ×, closing a window tab, closing a window.
- **A requested stop isn't a failure:** an interrupted turn ends in an `is_error` result and exit 1, which today would raise a failure banner.
- **Closing a window tab:** the last-tab guard is checked before anything is stopped, under a per-workspace lock.
- **Reopen during the grace window** waits for the old process to exit instead of racing it with a second `--resume`.

## Corrections to the earlier design

- `stop(graceful)` stays an immediate stop. The watchdog, controller replacement and `delete_controller` all pass `true` and rely on the stop being immediate; §4.3 is marked superseded.
- The `agent.open` live-elsewhere guard keys on the controller registry, which Phase 2 empties before the process exits. §9.6 closes that gap.

## Also recorded

- §6: reducer-owned pane-tab commands (stack push/switch/remove) as a separate follow-up spec.
- §6: using the interrupt for the Stop button.
- §7: open questions on other providers' exit behavior and on permission prompts pending at close.


a5af:
@codex review

PR #3406, opened 2026-09-19
feat(agents): graceful shutdown for every agent close (pane-close Phase 2)

AgentA-asaf:
Implements Phase 2 of `docs/specs/SPEC_AGENT_PANE_CLOSE_GRACEFUL_SHUTDOWN_2026_09_18.md` (§9, spec merged in #3405). Follow-up to #3402.

## What changes

Closing a pane, a pane tab, a window tab or a window now stops each agent gracefully instead of hard-killing it.

- **New `Controller::shutdown(deadline) -> StopOutcome`.** The default is today's immediate stop. The persistent (Claude stream-json) controller:
  - interrupts an active turn with the control-protocol interrupt and waits for its result;
  - then sends EOF;
  - force-kills only at the deadline.
  
  This follows the Phase 0 measurement (spec §5.1): EOF alone lets a running turn finish first (22.8s), while the interrupt ends it in about 2s.
- **A requested stop is not a failure.** The interrupted turn's `is_error` / `error_during_execution` result no longer raises a failure banner or triggers a stale-resume retry (`shutdown_generation`).
- **`shutdown_agents()`** is used by `delete_block`, `close_pane`, `delete_tab` and `delete_workspace`.
  - Agents stop concurrently under one deadline, so ten agents take about 5s, not 50s.
  - The process tracker is dropped only after the process has exited.
  - One `agent_shutdown` log line per agent records the outcome and elapsed time.
- **`delete_tab`** checks the last-tab guard under a per-workspace lock *before* stopping anything. A refused close can't leave a tab of stopped agents.
- **`agent.open`** waits for this agent's closing blocks before its live-elsewhere check. A reopen during the grace window can't start a second `--resume` on the same session.
- `stop(true, …)` stays immediate. The watchdog, controller replacement and `delete_controller` all rely on that (spec §9.3).

## Not in this PR

- **ACP and app-server controllers keep the immediate stop.** Spec §9.3 requires measuring each provider's exit behavior first.
- **Picker reattach wait:** that is Phase 3.

## Tests

- `persistent::shutdown_tests` runs a real node stub that behaves as measured:
  - mid-turn → interrupted, `Exited`, no `EVENT_AGENT_FAILURE`;
  - idle → exits on EOF;
  - ignores both → `Killed` at the deadline;
  - never spawned → `NotRunning`.
  
  These tests skip if `node` isn't on PATH. The check uses a PATH lookup, not a spawn, so it doesn't trip the I7 spawn-site inventory.
- `close_pane::tests` covers:
  - three agents that take 1s each finish in about 1s;
  - last-tab refusal stops nothing;
  - a window-tab close stops its agents;
  - the closing-wait signal.
- Full `agentmux-srv` suite run locally with `--test-threads=1`: all passed after the spawn-inventory fix.


a5af:
@codex review

a5af:
@codex review

PR #3408, opened 2026-09-19
fix(shell): inject AGENTMUX_AUTH_KEY into plain Terminal-widget panes

AgentO-asaf:
## Summary

`muxsh` (and `muxspect`/`muxopen`/`muxlog`) is meant to be run by a human from inside any AgentMux-opened pane — agent or shell — the same role Wave Terminal's `wsh` played. In practice it only worked from an agent pane's own PTY; a plain Terminal-widget pane always failed with `"AGENTMUX_LOCAL_URL / AGENTMUX_AUTH_KEY not set"`, even though that's exactly the documented, intended use case (`SPEC_MUXSH_FULL_COLLECTION_2026_09_16.md`).

**Root cause:** `AGENTMUX_AUTH_KEY` is injected into a spawned process's environment in exactly one place in the whole backend — `agent_handlers/input.rs`, via `PersistentSpawnConfig`/`SubprocessSpawnConfig`'s `env_vars`, used only for agent panes. The plain-terminal `ShellController` never set it at all. This isn't fixable by just reading the var back the way `AGENTMUX_LOCAL_URL` already is (`std::env::var(...)`) — `config.rs` deliberately scrubs `AGENTMUX_AUTH_KEY` from `agentmux-srv`'s own process environment right after startup (PR #801), so it has to be threaded through explicitly: `AppState.auth_key` → `resync_controller` → `ShellController` → the spawned PTY's env.

**While here:** also promoted the `AGENTMUX_LOCAL_URL` injection (previously scoped to only the interactive-shell branch of `ShellController`'s three spawn branches) into the same "applied to ALL THREE branches" section as the new `AGENTMUX_AUTH_KEY` injection and the existing `sanitize_pty_command` call — the direct-spawn and shell-wrapped `cmd` branches had the exact same latent gap, the kind of half-fix this file's own ReAgent P0 precedent (PR #3326) already warns against.

## Test plan
- [x] New real-PTY regression test (`plain_terminal_pane_shell_receives_the_instance_auth_key`, `#[ignore]`-gated like its siblings): spawns an actual Terminal-widget shell, has it print its own `AGENTMUX_*` env, and asserts `AGENTMUX_AUTH_KEY` matches the instance's configured key. Verified **red** without the fix (temporarily disabled it — the shell showed a *leaked ambient* value from the outer dev environment, not the instance's own key, so the assertion still correctly failed) and **green** with it restored.
- [x] `cargo test -p agentmux-srv --bin agentmux-srv` — 3983 passed, 0 failed, 16 ignored (unchanged ignore count)
- [x] `cargo test -p agentmux-srv --bin agentmux-srv -- --ignored` (targeted): new test passes
- [x] `cargo clippy -p agentmux-srv --lib --bin agentmux-srv` — no new warnings in touched files
- [x] `pane_env::spawn_site_coverage` tests still pass — confirms no new unsanitized spawn path was introduced


a5af:
@codex review

PR #3410, opened 2026-09-19
Opaz@charlie: fix(pane-chrome): keep-alive pane can render permanently blank — a subscription race in chromeVm

GenericAgentX-asaf:
## Summary
- Fixes a real bug reported live: two panes in a shared dev window (one was another agent's actual, fully-populated, actively-running session) went **permanently blank** — no header, no content, no spinner, no error. Confirmed via `window.MOS.getObjectValue` that both blocks' data was fully intact, and a drag-preview thumbnail of the same block rendered fine — ruling out the old dangling-layout-node bug class (`docs/investigations/INVESTIGATION_LAYOUT_DEAD_SPACE_STALE_TREE_RESURRECTION_2026_07_08.md`), which requires the block to not exist at all. This is a pure frontend rendering race, not a data/persistence issue.
- **Root cause:** `pane-leaf-chrome.tsx`'s `chromeVm` memo subscribes to a per-blockId `viewModelSlots` entry via `viewModelSlots.get(id)?.get()`. That slot is a lazily-created `createSignal`, only instantiated inside `keepAliveNodeModelFor` — itself only called once `content`'s `<For>` loop mounts that blockId's `<Block>`. If `chromeVm` evaluates *before* that mount happens, `viewModelSlots.get(id)` is `undefined`, so `?.get()` short-circuits and reads nothing reactive — `chromeVm`'s latch permanently settles on `null`, and nothing can ever wake it up again, even once the real Block mounts moments later and populates the slot.
- **Fix:** `chromeNodeModel`'s memo now calls `viewModelSlotFor(activeBlockId())` eagerly before returning its `activeViewModel` accessor. Since `chromeNodeModel()` runs synchronously inside `chromeVm`'s own computation, this guarantees the slot exists before `chromeVm` ever reads it — this doesn't just win the race, it eliminates it structurally.
- The non-keep-alive path (browser/editor/sysinfo/etc.) was never vulnerable — it reads the leaf's own `activeViewModel` signal, created unconditionally at node construction (`layoutNodeModels.ts`), not lazily per-blockId.

## Test plan
- [x] `npx vitest run` — 281 files / 4243 tests passing
- [x] **Live-verified via CDP, repeatedly:** 4/4 reloads with the fix showed zero blank `.tile-leaf` elements; toggling the fix off and reloading 3x reproduced exactly 4 blank panes every time — confirms both that the bug is real and that the fix closes it
- [x] Added a regression test (`pane-leaf-chrome.test.tsx`) exercising the async-view-type-resolution ordering that exposes this bug class. **Being upfront:** this test currently passes both with and without the fix in the jsdom/vitest harness (verified both ways) — it does not itself prove the fix, the live CDP reproduction does. Kept it anyway as documentation of the exact mechanism and a smoke test against this specific transition shape, since every existing test in the file sets block view *before* `render()`, never exercising async resolution timing at all.


a5af:
@codex review

GenericAgentX-asaf:
Good catch, and confirmed real — but I don't think it's worth fixing, and fixing it the obvious way would be actively counterproductive.

The obvious guard (`if (stackBlockIds().includes(id)) viewModelSlotFor(id);`) would reintroduce exactly the race this PR closes, for the narrow transient window where `activeBlockId()` hasn't yet landed in `stackBlockIds()` (e.g. `node.data.blockStack` hasn't caught up to a just-switched active member yet). The whole point of creating the slot *eagerly*, unconditionally, is to not depend on that ordering — gating it on stack membership puts the ordering dependency right back.

The cost on the other side is a stray `createSignal(null)` + Map entry that sits until the next `stackBlockIds()` recompute (any layout tree mutation for the tab) reaps it via the existing cleanup loop — never read by anything, never rendered, no user-visible effect. Given the choice between "a permanently blank pane" and "a harmless orphaned signal for a while," I'll take the latter.

PR #3409, opened 2026-09-19
docs(spec): pane tabs as reducer commands

AgentA-asaf:
New spec: `docs/specs/SPEC_PANE_TABS_REDUCER_COMMANDS_2026_09_18.md`. It is the follow-up named in `SPEC_AGENT_PANE_CLOSE_GRACEFUL_SHUTDOWN_2026_09_18.md` §6 (#3402, #3405). Spec only; no code in this PR.

## Why

The reducer stores each pane's tab stack but never changes it. Every add, switch and close is a frontend edit followed by a whole-tree `LayoutSetTree` push, with no version check. As a result, "this block exists in the tab" and "this block is in a pane" come apart. Section 2 lists three live ways a block ends up running with no pane.

1. **Create-then-place is two steps:** `pane.open { skip_placement }`, then a frontend push.
2. **A backend `delete` action for one tab removes the whole stacked pane in the frontend,** because `getNodeByBlockId` also matches background members. Every `delete_block` queues one. So when a terminal tab's shell exits (close-on-exit), its sibling agent tabs are orphaned.
3. **A stale whole-tree push drops a stack member the backend placed.**

## Design

- Invariants the reducer enforces:
  - I1: every block in a tab is in exactly one leaf.
  - I2: the stack fields are consistent.
- Stack commands:
  - `LayoutStackPush`
  - `LayoutStackActivate`
  - `LayoutStackRemove`
  - `CreateBlockInStack`, which creates and places a block in one reducer step.
- The frontend applies each change optimistically and sends it as an intent, de-duplicated by `actionid`.
- `pane.open { stack_onto_block_id }`.
- A backend `delete` action means "remove this one block from the layout".
- Stale-push protection keyed on a per-tab `layout_revision`.
- One rule for which tab becomes visible, shared by backend and frontend.

**Phase 0** (the `delete` action fix, item 2 above) ships separately, as a small PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3411, opened 2026-09-19
fix(ci): claude-md-references gate fails on main via INDEX.md

AgentA-asaf:
Since #3403/#3404, `scripts/check-claude-md-references.sh` fails on **main**, and so on every PR's "doc status + grep gates" job:

```
FAIL docs/specs/INDEX.md:525
     | [`SPEC_CLAUDE_MD_CONTENT_PORT_2026_09_18`](SPEC_CLAUDE_MD_CONTENT_PORT_2026_09_18.md) | SPEC: Port `CLAUDE.m…
```

`INDEX.md` is generated. It repeats each spec's title, and this spec's title names `CLAUDE.md`. The spec itself is exempt from the check: it's dated, matching `EXCLUDE_PATH`, and it's `$SPEC`. Its index line wasn't.

**Fix:** for `docs/specs/INDEX.md`, judge an entry by its link target. If the target is an excluded dated doc, skip the entry the way the doc itself is skipped. Entries linking to undated (living) docs are still checked; verified by appending an undated test entry, which still fails.

This doesn't add `INDEX.md` to `EXCLUDE_PATH`, which the script asks you not to do. Seen failing on #3409.


a5af:
@codex review

PR #3414, opened 2026-09-19
fix(panes): closing one tab of a stacked pane no longer drops the whole pane

AgentA-asaf:
Phase 0 of `SPEC_PANE_TABS_REDUCER_COMMANDS_2026_09_18.md` (spec in #3409). It fixes a live way for agents to be orphaned (spec §2.2).

## The bug

- A backend `delete` layout action names **one** block.
- The frontend handler looked the leaf up with `getNodeByBlockId`, which also matches a background member of a stacked pane (`layoutNodeModels.ts`), and then **deleted the whole leaf**. Every sibling tab went with it, and their blocks, agents included, kept running with no pane.
- **Every** `delete_block` saga run queues this action (Step 2b). So when a terminal tab's shell exits and close-on-exit closes it, the agent tabs sharing its pane are orphaned: still in the tab, still running, visible only in Swarm. Tear-off, redock and promote queue the same action.
- The backend had the mirror bug. `LayoutDeleteNodeByBlock` on the **visible** member deleted the whole leaf.

## Fix

- **Frontend:**
  - New pure `frontend/layout/lib/stackMembers.ts` with `effectiveStack` and `removeMemberFromStack`.
  - The `delete` handler now removes just that member. The leaf goes only when it was the last member.
  - `closeBlockInStack` uses the same helper.
- **Backend:**
  - `backend::layout::remove_stack_member` applies the same right-neighbour rule.
  - `LayoutDeleteNodeByBlock` now removes the member and emits a tree-only `LayoutTreeReplaced` (`slices: None`, so focus, magnify and the pending queue are untouched). It deletes the leaf only for its last member.

## Tests

- **Reducer:** background member, visible member (right neighbour), rightmost visible member (new last), last member (leaf removed).
- **Saga:** `delete_block` of the terminal tab in `[agent, term]` keeps the agent block and its pane, and the agent becomes visible.
- **Vitest:** the `delete` action on a background member and on the visible member of a stacked pane.
- `tsc --noEmit` is clean. The related `agentmux-srv` tests pass locally (19). The vitest suite runs in CI only (local Node is below the pinned version).

The doc gate will be red until #3411 lands; it's the main-wide `check-claude-md-references` failure, unrelated to this change.


a5af:
@codex review

AgentA-asaf:
Addressed the P1 (layoutPersistence.ts:236) in b89e31740. The tree-walk fallback no longer deletes on its own: both lookups (`getNodeByBlockId`, then `findNodeByBlockId`) resolve the leaf, then one decision runs, which removes the member and deletes the leaf only for its last member. Added a vitest for the case you described: two sibling deletes in the same batch leave the pane with its remaining tab.

a5af:
@codex review

a5af:
@codex review

